PR #185, opened 2026-08-20
fix(mcp): refresh OAuth tokens without browser popups

damianvtran:
# PR Description

## Change classification

**C2 — standard / pre-authorized.** Bug fix to MCP OAuth token handling: proactive refresh against discovered endpoints, cross-process refresh lock, non-interactive startup/reconnect/exit, and a deliberate `/mcp login` path. No new config surface, no schema change. Clean rollback (`git revert`).

## The bug

Reported: starting a new `lop` session pops a browser login tab for the Datadog MCP (and Notion) even though the user recently logged in. It also pops on **exit**.

Three defects combined:

1. **Wrong refresh endpoint.** On a fresh process the SDK's `OAuthClientProvider` has no authorization-server metadata, so its in-flow refresh falls back to `urljoin(server_base, "/token")`. For Datadog that's `https://mcp.us3.datadoghq.com/token` → **404** (the real endpoint is `https://us3.datadoghq.com/api/v2/oauth2/token`). The failed refresh escalates to a full browser grant → the popup. (Linear/Notion only "worked" because their token endpoint happens to live at `<host>/token`.)
2. **Cross-session race.** Several sessions starting together all spent the same rotating refresh token and invalidated each other — reproduced against the real Notion endpoint: `{"error":"invalid_grant","error_description":"Refresh token reuse detected"}`.
3. **No interactivity gate.** Startup, auto-reconnects, and the exit-time session-termination `DELETE` all ran the same flow that opens a browser.

## The fix

- **Proactive refresh before connecting** (`ensure_mcp_oauth_fresh`): spends a stored refresh token against the token endpoint resolved via PRM → ASM discovery, so a day-old access token never forces a browser grant. Refreshes are serialized across processes with a file lock and the token is re-read under it, so concurrent sessions can't double-spend a rotating refresh token.
- **Non-interactive startup/reconnect/exit.** The redirect handler raises `McpAuthRequiredError` instead of opening a browser when the flow isn't an explicit login. This is the universal defense — it covers the startup connect, mid-session reconnects, AND the exit `DELETE` (the SDK's `terminate_session` catches it, so no popup and no crash).
- **Actionable surfacing.** An unrefreshable grant becomes `needs authorization — run /mcp login <name>` in the startup toast, a dedicated toast for after-gate/mid-session failures, and a model-visible incident. The reconnect loop abandons (an expired grant won't heal by retrying) instead of burning the breaker window.
[...]

PR #242, opened 2026-08-22
fix(mcp): close mid-session OAuth refresh race and defer the startup report until settled

damianvtran:
## Why

MCP OAuth servers re-request authorization far more often in local-operator than in other harnesses (omp, opencode, where a login lasts a month), and Notion in particular logs out several times a day. Two independent defects in the installed runtime (`origin/main`) cause it. Both were diagnosed against the live credential store and the SDK source.

**1. The mid-session OAuth refresh is not race-safe across processes.** #185 correctly serialized the *startup* refresh with a cross-process file lock (`~/.local-operator/mcp_oauth_refresh.lock`) and a re-read under it — but its own docstring notes the remaining gap: the SDK's own *in-flow* refresh (triggered by a 401 mid-session, `mcp/client/auth/oauth2.py`) takes no such lock and spends whatever refresh token it loaded once at `_initialize`. This harness runs **one process per cmux workspace**, so several long-lived sessions cross the same token-expiry boundary while each still holds the same in-memory refresh token from when it booted. For a provider that **rotates** its refresh token on every use, that is destructive.

Notion is the worst case: its stored token (read from `auth.db`) is an **8-hour** access token with a **rotating** refresh token, so every long-lived session must refresh ~3×/day and the losers of the race are logged out. Providers other harnesses keep you signed into either don't rotate or issue long-lived access tokens, so their concurrent sessions rarely collide.

Reproduced against a mock rotating-token endpoint with 6 concurrent sessions all holding the same stale refresh token (the real scenario — all pre-initialized, released together):

```
=== BASELINE (bare SDK provider, pre-fix) ===
token POSTs to server: 6
invalid_grant (double-spend) hits: 5
session access tokens: ['acc1', None, None, None, None, None]
sessions left with NO token (=> browser grant): 5
```

Five of six sessions get `invalid_grant` → the SDK `clear_tokens()` → a full browser grant on each. That is the repeated Notion logout.

**2. The startup report cries wolf: "one server failed" is announced as "N of M up" while the rest are still connecting.** `session.mcp_startup` is a single snapshot taken at the 250 ms startup gate. OAuth HTTP servers routinely miss that gate — they run PRM/ASM discovery plus a possible token refresh *before* their transport opens. A server that fails *fast* (a validation error, one auth-expired server) flips the boot toast to `MCP: 0 of 7 servers up — failed: X` a beat before the six slow OAuth servers land. The background continuations then connect fine and fix the live status band, but they never update `mcp_startup`, so the toast and the durable transcript notice keep asserting the pessimistic tally — exactly the "if one MCP fails it reports that all failed" behaviour.

## What changes

**Fix 1 — a refresh-coordinating OAuth provider (`local_operator/mcp/auth.py`).** `build_oauth_provider` now returns a small `OAuthClientProvider` subclass whose `async_auth_flow`, before the SDK would refresh, re-reads the persisted token under the **same** `_oauth_refresh_lock` the startup path uses. If a sibling process already rotated the token it **adopts** that fresh token and skips the refresh entirely; otherwise it performs **exactly one** locked refresh against the discovered token endpoint (`_refresh_oauth_token_locked`, already used by the startup path). The coordination is gated on the SDK's own refresh predicate (`not is_token_valid() and can_refresh_token()`), so a still-valid token adds no round trip. With no discovered endpoint the re-read/adopt still runs (removing the common racing case) and only the self-performed refresh falls back to the SDK's own path — the pre-fix behaviour for that one provider, never a crash.

While wiring this I found and fixed a **real generator-delegation bug**: a plain `async for … yield` does not forward httpx's `asend(response)` values into the SDK's *receiving* auth generator (the sub-generator sees `None` for every `response = yield` and crashes on `response.status_code`). The provider now pumps the sub-generator by hand, which is the only correct way to delegate a receiving async generator.

**Fix 2 — defer the startup report until the round settles** (`mcp/manager.py`, `session/mcp_status.py`, `session_factory.py`, `session/session.py`, `tui/app.py`). The manager accumulates the combined per-server outcome across the gate pass *and* the background continuations (`_startup_failures`), tracks servers still connecting past the gate (`_startup_deferred`), and fires a new `on_startup_settled` callback once the last deferred server reaches a terminal state. `McpStartupOutcome` gains a `settling` flag: while settling, `reportable` is `False` so the toast/notice stay quiet (the live status-band lamp still covers that window), and the front end re-reports the final combined tally once the round settles. Nothing deferred ⇒ the gate snapshot was already final and the callback never arms.

## Testing evidence

**Concurrency reproduction (the core of fix 1), before → after.** Same 6-concurrent-session rotating-token scenario as above, now with the refresh-coordinating provider:

```
=== WITH FIX (realistic: all pre-initialized with r0, released together) ===
token POSTs to server: 1
invalid_grant (double-spend) hits: 0
session access tokens: ['acc1', 'acc1', 'acc1', 'acc1', 'acc1', 'acc1']
sessions left with NO token (=> browser grant): 0
RESULT: PASS
```

6 sessions → **1** refresh POST, **0** double-spends, **0** browser grants, all sharing the one live rotated token. (Baseline in the Why section: 5 double-spends, 5 forced grants.)

**Unit tests** (all new, all pass):
- `tests/unit/mcp/test_auth.py::TestInflightRefreshCoordination` (4): adopts a sibling's rotated token without refreshing; performs exactly one locked refresh when the store is still stale; a valid token never enters the lock; no-endpoint falls back without crashing.
- `tests/unit/mcp/test_manager.py::TestStartupSettleReporting` (4): settling stays true until the deferred set drains then the callback fires once; a fast failure does not settle the round while a slow server is pending, and the settled map names only the real failure; all-fast means not-settling and no callback; a deferred (after-gate) failure is recorded in the settled map.
- `tests/unit/tui/test_toast.py` (2): a settling snapshot with a failure stays quiet; the same outcome toasts once settled.
- `tests/unit/test_session_factory.py` (1): the boot snapshot is provisional (`settling`, `reportable=False`) and `on_startup_settled` rebuilds `mcp_startup` with the final tally and notifies the session sink.

Full unit suite: **3457 passed, 3 skipped, 1 failed**. The one failure is `test_eval_tool::test_background_streams_output_while_it_runs`, an environmental flake unrelated to this change: it reproduces on a clean, unmodified worktree under `/private/tmp` and passes under `/Users`, and this MR touches zero eval/jobs/harness code. Gates clean: `flake8`, `black --check`, `isort --check-only`, `pyright` (0 errors).

**Visual evidence (fix 2, real `OperatorApp` rendered frames).** Captured via `run_test` + `app.save_screenshot` on the real app (the one that loads the stylesheet), then rendered and viewed. Both frames drive the actual `_report_mcp_startup` path.

- **Settling frame** — the boot moment a fast failure used to flash a false `MCP: 0 of 2 servers up — failed: notion` toast. With the fix the toast is **suppressed** (the outcome's `settling` flag makes it non-reportable); the splash is clean and the status-band MCP lamp still shows the count, so the in-between window is covered without a false alarm.
- **Settled frame** — once the deferred server reaches a terminal state the round settles and the **accurate** report appears: toast `⊙ MCP: 1 of 2 servers up` / `failed: notion — run /mcp login notion to authorize` (the full command survives the width clamp after the design-review copy fix), plus the durable transcript notice `✗ MCP notion failed: run /mcp login notion to authorize`.

Rendered PNGs are attached to the PR discussion (GitHub CLI cannot embed images in the body).

## Review gate

- **Agent code review:** round 1 (1 blocker — a `context.lock` leak in the async-generator delegation — + 2 minor, all remediated) → round 2 (verified terminal) → round 3 (copy-change convergence, terminal). All rounds by a reviewer subagent distinct from the author.
- **Design review:** round 1 (D1/D4 remediated — the truncated `/mcp login` command and a double em-dash — D2/D3 rejected as shot-harness artifacts, D5 deferred) → round 2 (terminal). By a designer subagent.
- Both rounds are clean and fresh on the current head `a52e8277`.

## Change class

C2 (standard, pre-authorized). Behavioural fix to MCP OAuth handling; degrades safely (a failed coordination falls through to the SDK's own refresh, a missing accessor to the old snapshot), no schema or contract change. Runtime publication via `lop-update` is a separate step held until this is merged.

### DIFF
--- a/local_operator/mcp/auth.py
+++ b/local_operator/mcp/auth.py
@@ -1641,6 +1641,147 @@ def wire_oauth_auth(
     }
 
 
+def _make_refresh_coordinating_provider(
+    kwargs: dict[str, Any],
+    *,
+    server_url: str,
+    storage: "McpTokenStorage",
+    endpoints: DiscoveredOAuthEndpoints | None,
+) -> Any:
+    """An ``OAuthClientProvider`` whose in-flow refresh is race-free across processes.
+
+    Why this subclass exists: the SDK loads the stored tokens ONCE in
+    ``_initialize`` and never re-reads storage afterwards, and its in-flow
+    refresh (``async_auth_flow`` -> ``_refresh_token``) spends whatever refresh
+    token that first read put in memory, under NO cross-process lock. That is
+    fine for a token that expires and is refreshed once. It is destructive for a
+    provider that ROTATES its refresh token on every use (Notion: an 8-hour
+    access token plus a rotating refresh token) the moment more than one
+    local-operator process is alive: this harness runs one process per cmux
+    workspace, so several long-lived sessions cross the same 8-hour boundary
+    still holding the same in-memory refresh token from when they booted. They
+    each POST it; the authorization server rotates; the first wins and every
+    other session's brand-new token is already dead. The SDK then
+    ``clear_tokens()`` and the next request opens a FULL browser grant — the
+    "Notion logged out again" the user sees, several times a day.
+
+    :func:`ensure_mcp_oauth_fresh` already closes this race for the STARTUP
+    refresh with a cross-process file lock and a re-read under it; its own
+    docstring notes that the SDK's in-flow refresh is the remaining unlocked
+    path. This subclass extends the same guard to that path: before the SDK
+    would refresh, it re-reads the persisted token under
+    :func:`_oauth_refresh_lock` and, if a sibling process already rotated it,
+    ADOPTS the fresh token into the context and skips the refresh entirely.
+    Only when the store still holds an expired token does exactly one process
+    perform the exchange — against the discovered endpoint, so a provider whose
+    token endpoint is not ``<server_base>/token`` (Datadog) refreshes rather
+    than 404-ing into a browser grant.
+
+    Degradation is honest: with no discovered ``endpoints`` (metadata discovery
+    failed at startup) the re-read/adopt step still runs — which alone removes
+    the common case, where a sibling already refreshed — and only the
+    self-performed exchange falls back to the SDK's own unlocked refresh, i.e.
+    exactly the pre-fix behaviour for that one provider.
+    """
+    from mcp.client.auth import OAuthClientProvider
+
+    class _RefreshCoordinatingOAuthProvider(OAuthClientProvider):
+        # Bound on the re-read/refresh interception: a stuck lock acquisition or
+        # a hung endpoint must not park an authenticated request forever. The
+        # file lock is only ever held for one token POST (REFRESH_HTTP_TIMEOUT_S)
+        # plus the SQLite read, so a peer that legitimately holds it clears well
+        # inside this bound; overrunning it means a leaked lock, and degrading to
+        # the SDK's own refresh is strictly better than blocking the request.
+        _refresh_coord_server_url = server_url
+        _refresh_coord_storage = storage
+        _refresh_coord_endpoints = endpoints
+
+        async def _coordinate_inflight_refresh(self) -> None:
+            """Re-sync from the store (and refresh once, race-free) if the SDK is
+            about to refresh. No-op unless the loaded token is invalid AND
+            refreshable — exactly the SDK's own in-flow refresh trigger, so this
+            never adds a round trip to a request that would have gone through
+            unauthenticated-refresh-free."""
+            ctx = self.context
+            # Gate on the SAME predicate the SDK's async_auth_flow uses so we
+            # intercept precisely when it would refresh, and never otherwise.
+            if ctx.is_token_valid() or not ctx.can_refresh_token():
+                return
+            try:
+                async with _oauth_refresh_lock(self._refresh_coord_server_url):
+                    # Re-read under the lock: a sibling process may have rotated
+                    # the token while we waited. Adopting its result is what
+                    # turns a double-spend into a no-op.
+                    stored = await ctx.storage.get_tokens()
+                    if stored is not None and stored.access_token:
+                        ctx.current_tokens = stored
+                        ctx.token_expiry_time = self._refresh_coord_storage.stored_token_expiry()
+                    if ctx.is_token_valid():
+                        return  # a peer already refreshed; do not spend again
+                    endpoints = self._refresh_coord_endpoints
+                    if endpoints is None:
+                        # No discovered endpoint: leave the stale token in place
+                        # and let the SDK's own (unlocked) in-flow refresh run —
+                        # the pre-fix path for this one provider. The re-read
+                        # above already handled the common racing case.
+                        return
+                    refreshed = await _refresh_oauth_token_locked(
+                        self._refresh_coord_server_url,
+                        self._refresh_coord_storage,
+                        endpoints,
+                    )
+                    if refreshed:
+                        new_tokens = await ctx.storage.get_tokens()
+                        if new_tokens is not None:
+                            ctx.current_tokens = new_tokens
+                            ctx.token_expiry_time = (
+                                self._refresh_coord_storage.stored_token_expiry()
+                            )
+            except Exception:  # noqa: BLE001 — coordination is best-effort
+                # A failed re-read/refresh must never break the request: fall
+                # through to the SDK's own auth flow, which will refresh (or, if
+                # that fails, surface the non-interactive McpAuthRequiredError).
+                logger.debug(
+                    "MCP in-flight refresh coordination failed for %s",
+                    self._refresh_coord_server_url,
+                    exc_info=True,
+                )
+
+        async def async_auth_flow(self, request):  # type: ignore[override]
+            # Ensure tokens+client_info are loaded (so is_token_valid /
+            # can_refresh_token below are meaningful), then coordinate the
+            # refresh, then hand off to the SDK's flow. The SDK re-checks
+            # is_token_valid under its own lock and will skip its refresh branch
+            # because we have already made the token valid. context.lock is NOT
+            # held across the coordination (it is not reentrant and the SDK
+            # re-acquires it), matching how the SDK itself only holds it inside
+            # the flow.
+            async with self.context.lock:
+                if not self._initialized:
+                    await self._initialize()
+            await self._coordinate_inflight_refresh()
+            # Delegate to the SDK flow by hand, forwarding the RESPONSE the
+            # caller sends back into each yield: httpx drives an auth flow with
+            # ``gen.asend(response)``, and a plain ``async for ...: yield`` would
+            # swallow those sent values (the sub-generator would see ``None`` for
+            # every ``response = yield`` and crash on ``response.status_code``).
+            # Manual pumping is the only correct way to delegate a receiving
+            # async generator — there is no ``yield from`` for them.
+            inner = super().async_auth_flow(request)
+            try:
+                outgoing = await inner.__anext__()
+            except StopAsyncIteration:
+                return
+            while True:
+                response = yield outgoing
+                try:
+                    outgoing = await inner.asend(response)
+                except StopAsyncIteration:
+                    return
+
+    return _RefreshCoordinatingOAuthProvider(**kwargs)
+
+
 def build_oauth_provider(
     server_url: str,
     cfg: MCPServerConfig,
@@ -1671,8 +1812,6 @@ def build_oauth_provider(
     SDK's ``<server_base>/token`` guess (which 404s for providers like Datadog
     whose token endpoint lives on a different host).
     """
-    from mcp.client.auth import OAuthClientProvider
-
     # The flow is created HERE (not inside wire_oauth_auth) so it can be
     # attached to the provider: an abandoned grant arrives at the connect as
     # a raw CancelledError (see ``LoopbackAuthFlow.callback_handler``), and
@@ -1683,13 +1822,19 @@ def build_oauth_provider(
         _resolve_redirect_uri(cfg), server_url=server_url, interactive=interactive
     )
     kwargs = wire_oauth_auth(server_url, cfg, store=store, interactive=interactive, flow=flow)
-    provider = OAuthClientProvider(**kwargs)
+    storage = kwargs["storage"]
+    # A refresh-coordinating provider (not the bare SDK one): its in-flow
+    # refresh re-reads the store under the cross-process lock so several
+    # long-lived sessions cannot double-spend a rotating refresh token
+    # mid-session — see :func:`_make_refresh_coordinating_provider`.
+    provider = _make_refresh_coordinating_provider(
+        kwargs, server_url=server_url, storage=storage, endpoints=endpoints
+    )
     provider._loopback_flow = flow  # type: ignore[attr-defined]
     if endpoints is not None:
         provider.context.oauth_metadata = endpoints.oauth_metadata
         provider.context.protected_resource_metadata = endpoints.protected_resource_metadata
         provider.context.auth_server_url = endpoints.auth_server_url
-    storage = kwargs["storage"]
     try:
         expiry = storage.stored_token_expiry()
     except Exception:  # noqa: BLE001 — a metadata read must not block a connect
--- a/local_operator/mcp/auth.py
+++ b/local_operator/mcp/auth.py
@@ -1767,17 +1767,53 @@ def _make_refresh_coordinating_provider(
             # every ``response = yield`` and crash on ``response.status_code``).
             # Manual pumping is the only correct way to delegate a receiving
             # async generator — there is no ``yield from`` for them.
+            #
+            # The whole pump is wrapped so the inner generator is ALWAYS closed
+            # and any exception/cancellation is delivered INTO it, never dropped
+            # on the floor. This matters because the SDK's ``async_auth_flow``
+            # holds ``context.lock`` across its entire body: httpx re-raises a
+            # transport error mid-flow (``_send_handling_auth`` does
+            # ``raise exc`` after ``response.aclose()``), and if we let that
+            # unwind past us without closing ``inner``, the SDK generator is
+            # suspended forever at its ``yield`` still holding the lock — every
+            # later request to this server then deadlocks on ``context.lock``.
+            # ``athrow`` runs the SDK's own ``finally`` (which releases the lock
+            # via the ``async with`` exit); ``aclose`` covers the GeneratorExit
+            # path when httpx closes the OUTER flow (its ``finally:
+            # await auth_flow.aclose()``).
             inner = super().async_auth_flow(request)
             try:
-                outgoing = await inner.__anext__()
-            except StopAsyncIteration:
-                return
-            while True:
-                response = yield outgoing
                 try:
-                    outgoing = await inner.asend(response)
+                    outgoing = await inner.__anext__()
                 except StopAsyncIteration:
                     return
+                while True:
+                    try:
+                        response = yield outgoing
+                    except GeneratorExit:
+                        # The outer flow is being closed (httpx's finally, or a
+                        # cancellation): close the inner one so the SDK unwinds
+                        # its lock, then let the close propagate.
+                        raise
+                    except BaseException as exc:  # noqa: BLE001
+                        # An exception thrown INTO us (httpx ``athrow`` on a
+                        # transport fault): deliver it into the SDK generator so
+                        # its ``finally`` runs and the lock is released, then
+                        # relay whatever it yields or re-raises.
+                        try:
+                            outgoing = await inner.athrow(exc)
+                        except StopAsyncIteration:
+                            return
+                        continue
+                    try:
+                        outgoing = await inner.asend(response)
+                    except StopAsyncIteration:
+                        return
+            finally:
+                # Unconditional: a normal return, a raise, or a GeneratorExit all
+                # pass through here, so the SDK generator (and its held lock) is
+                # never left suspended. Idempotent if already exhausted/closed.
+                await inner.aclose()
 
     return _RefreshCoordinatingOAuthProvider(**kwargs)
 
--- a/local_operator/mcp/manager.py
+++ b/local_operator/mcp/manager.py
@@ -1591,11 +1591,17 @@ class McpManager:
     def _auth_required_text(name: str, exc: McpAuthRequiredError) -> str:
         """The startup-toast wording for a server that needs an OAuth login.
 
-        Names the command that fixes it. The same string lands in the durable
-        transcript notice and in ``/mcp``, so one helper keeps all three
-        surfaces agreeing.
+        Leads with the COMMAND that fixes it rather than the diagnosis. The
+        toast renders this after a ``failed: <name> — `` prefix and then clamps
+        to the card width, so the tail is what gets truncated: putting
+        ``run /mcp login <name>`` first keeps the one actionable thing on screen
+        even on a narrow terminal, where ``needs authorization — run /mcp login
+        <name>`` used to sever the command mid-word (design review D1). One
+        ``—`` only, so the composed line is not a chain of dashes (D4). The same
+        string lands in the durable transcript notice and in ``/mcp``, so one
+        helper keeps all three surfaces agreeing.
         """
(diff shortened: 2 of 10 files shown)


damianvtran:
### Agent review — round 1
- Reviewer: reviewer subagent (claude-opus-4-6)
- Scope: f5dadc69..c592b49a
- Mode: Comprehensive
- Delivery contract: Close the cross-process mid-session OAuth refresh race (adopt a sibling's rotated token under the file lock) and defer the MCP startup report until the connect round settles.
- Applicable risk surfaces: OAuth token handling, cross-process concurrency, async-generator delegation, startup reporting
- Verdict: Changes requested

| --- | --- | --- | --- |
| F1 | BLOCKER | async-generator delegation / OAuth availability | `local_operator/mcp/auth.py:1770-1780` — the manual pump delegating to `super().async_auth_flow` has **no `finally: await inner.aclose()`**. The SDK holds `context.lock` for the *entire* `async_auth_flow` generator (`oauth2.py:582 async with self.context.lock:`). httpx drives the flow inside `try/finally: await auth_flow.aclose()` and re-raises any transport error from `_send_handling_redirects` (`httpx2/_client.py:1841-1866`), so when an authenticated MCP request errors or is cancelled **mid-flow** (after the first yield, before `StopAsyncIteration`), httpx `aclose()`s the *outer* pump, GeneratorExit unwinds the pump, but `inner` is never closed — the SDK's `context.lock` stays held forever. Every subsequent request to that server then blocks on `context.lock` (both this subclass's `async_auth_flow` and the SDK's re-acquire it), permanently killing that server's tools until process restart. I reproduced it against a faithful model of the SDK's lock-across-yields flow: after a mid-flow abort, `lock.locked()` is `True` and the next request deadlocks; adding `try/finally: await inner.aclose()` around the `while True` loop clears it (`lock.locked()` False, next request completes). This is a strictly worse availability regression than the rotating-token symptom it fixes. **Required action:** wrap the pump loop in `try: ... finally: await inner.aclose()`. |
| F2 | MINOR | settle accounting under reload/dispose | `local_operator/mcp/manager.py:1696-1734` — the `_finish_pending` failure arm mutates `self._startup_failures[name]` and calls `self._settle_deferred(name)` **without** the `if self._disposed or epoch != self._epoch:` guard that the success arm has (line 1735, with the "disposed/superseded round is not a settled one" comment). A continuation whose `await task` resolves with a non-`CancelledError` exception concurrently with a `reload()`/`disconnect_all()` epoch bump would record a retired round's failure into the freshly-reset `_startup_failures` and settle a server that belongs to the old round. Low probability because the epoch bump cancels the continuation first (→ the `except asyncio.CancelledError` arm returns early), but the asymmetry is a real latent hazard. **Required action:** apply the same `epoch != self._epoch or self._disposed` guard at the top of the `except Exception` arm before mutating the round accumulators. |
| F3 | MINOR | test coverage of the delegation path | `tests/unit/mcp/test_auth.py` `_drive_auth_flow` (and all four `TestInflightRefreshCoordination` cases) always pump the flow to `StopAsyncIteration` and never abort mid-flow, so the F1 lock leak passes CI green — the tests are not tautological but they miss the one path that matters for a hand-written generator delegation. **Required action:** add a case that raises inside the driven flow (or `aclose()`s after the first yield) and asserts `provider.context.lock` is not held afterward (and that a following `async_auth_flow` does not block); it fails on current code and guards the F1 fix. |

Notes on areas checked and found sound (no action): token adoption keeps `current_tokens`/`token_expiry_time` consistent with the SDK's `is_token_valid` (both re-read via `stored_token_expiry()`, `None` → "no opinion" matches SDK); `_oauth_refresh_lock` is a single cross-process flock acquired off-loop via `asyncio.to_thread`, `_refresh_oauth_token_locked` does not re-flock (no reentrancy), and `context.lock` is released before the file lock is taken so the two lock domains do not nest; the `settling`/`reportable` suppression and the headless-path stderr re-report on settle are wired symmetrically through `_collapse_sdk_missing_failures`; `on_startup_settled` fires exactly once per round (each deferred name settles once; cancelled continuations return before settling). The 249 tests in `tests/unit/mcp`, `test_session_factory.py`, and `tests/unit/tui/test_toast.py` pass.

Verdict: **Changes requested** — one BLOCKER (F1). Not terminal; re-review after F1 is fixed and F3's regression test is added.

damianvtran:

- **F1 (blocker) — Fixed in `df93e19`.** The manual `async_auth_flow` pump now closes the SDK's inner generator on every exit path. The whole pump is wrapped in `try/finally: await inner.aclose()`; a `GeneratorExit` (httpx closing the outer flow in its `finally`) is re-raised after that close runs, and an exception thrown *into* the pump (httpx's `athrow` on a transport fault) is delivered into the SDK generator via `inner.athrow(exc)` so the SDK's own `finally` releases `context.lock` before the error relays out. Confirmed against the exact deadlock you reproduced: two new regression tests (`test_a_transport_error_mid_flow_releases_the_sdk_lock`, `test_closing_the_flow_early_releases_the_sdk_lock`) assert `provider.context.lock` is not held after a mid-flow `ConnectError` and after an early `aclose()`. Both **fail on the pre-fix pump** (`AssertionError: SDK context.lock leaked`) and pass on the fix — verified by reverting the pump to the round-0 version and re-running.

- **F2 (minor) — Fixed in `df93e19`.** `_finish_pending`'s failure arm now computes `current_round = not self._disposed and epoch == self._epoch` and guards both the `_startup_failures[name]` write and the `_settle_deferred(name)` call behind it, matching the success arm's existing guard. The waiter settling (`_settle_future_error`) and the auth-required incident/toast stay unguarded on purpose — a parked deferred-execute waiter must fail rather than hang regardless of which round owns it, and a real auth failure still deserves its toast. New test `TestStartupSettleStaleRound` supersedes the round mid-continuation (bumps the epoch) and asserts the stale failure writes nothing to the current round's accumulator and never fires the settle callback.

- **F3 (minor) — Fixed in `df93e19`.** Added the two lock-release regression tests described under F1. They are the mutation-proof guard you asked for: they fail on the F1-broken pump and pass on the fixed one.

Full affected suites green after remediation: `tests/unit/mcp` + `tests/unit/test_session_factory.py` = 284 passed; `tests/unit/tui/test_toast.py` = 40 passed. Gates clean (`flake8`, `black --check`, `isort`, `pyright` 0 errors). The 6-concurrent-session rotating-token reproduction still shows 1 refresh POST / 0 double-spends / 0 browser grants.

Head is now `df93e192`; requesting a fresh convergence round on it.

damianvtran:
### Agent review — round 2

- Reviewer: reviewer subagent (claude-opus-4-6)
- Scope: f5dadc69..df93e192
- Delta reviewed: c592b49a..df93e192
- Verdict: **Approve** — no blocker/major remains. Round is TERMINAL.

| ID | Severity | Disposition | Evidence |
| --- | --- | --- | --- |
| F1 | BLOCKER | **Resolved** | `auth.py:1784-1815` — the pump is now wrapped in `try: … finally: await inner.aclose()`, so a normal return, a re-raise, or a `GeneratorExit` all close the SDK generator and release `context.lock`. The relay is correct on all three sub-paths I was asked to check: (a) **athrow path** — `except BaseException as exc: outgoing = await inner.athrow(exc)` runs the SDK's own `finally` (which exits `async with self.context.lock`) before the error relays out; (b) **retry-after-athrow** — if the SDK catches and yields again, `athrow` returns that request, which the `continue` delivers to httpx via `response = yield outgoing`; if it re-raises, the original exception propagates out unchanged and hits the no-op `aclose`; (c) **GeneratorExit** — caught first (before `BaseException`, correct MRO ordering) and re-raised, then `finally` closes `inner`; (d) no double-close (post-return/re-raise `aclose` is an idempotent no-op) and no swallowed cancellation (a `CancelledError` thrown in is forwarded into `inner`, not eaten). **Mutation check reproduced:** reverting the pump to the round-1 version, `test_a_transport_error_mid_flow_releases_the_sdk_lock` and `test_closing_the_flow_early_releases_the_sdk_lock` both fail (`AssertionError: SDK context.lock leaked` / anyio `RuntimeError: The current task is not holding this lock`); both pass on `df93e192`. |
| F2 | MINOR | **Resolved** | `manager.py:1727-1744` — the failure arm now computes `current_round = not self._disposed and epoch == self._epoch` and guards both the `_startup_failures[name]` write and the `_settle_deferred(name)` call behind it, matching the success arm. The intentionally-unguarded parts are the right ones: `_settle_future_error` on the re-fetched waiter (a parked deferred-execute waiter must fail rather than hang regardless of round) and `_fire_auth_required`/tools-changed cleanup (a real auth failure still deserves its toast; dropping the retired deferred slice is safe cleanup). `TestStartupSettleStaleRound` bumps the epoch mid-continuation and asserts the stale failure writes nothing and never settles — passes. |
| F3 | MINOR | **Resolved** | `test_auth.py:1768-1832` — the two regression tests drive the flow to its first yield, then respectively `athrow` a `ConnectError` and `aclose` early, asserting `provider.context.lock` is not held afterward. They assert the correct property (the SDK lock is the thing that leaks) and are mutation-proof, as re-verified above. |

Convergence pass: the athrow/GeneratorExit handling does not alter the happy path — when no coordination refresh occurs the flow pumps through `asend` to `StopAsyncIteration` exactly as before, and the added `finally` `aclose` is a no-op on an already-exhausted generator. `except GeneratorExit` precedes `except BaseException` so cancellation/close is re-raised rather than mis-routed into the retry relay. No new blocker or major surfaced.

Gates verified on `df93e192`:
- `tests/unit/mcp` + `tests/unit/test_session_factory.py` → **284 passed**
- `tests/unit/tui/test_toast.py` → **40 passed**

Verdict: **Approve.** All round-1 findings (F1 blocker, F2, F3 minors) resolved on the current head; no blocker/major remain. This round is terminal — nothing outstanding.

damianvtran:
### Design review — round 1

Reviewer: designer (Claude Sonnet 4.5)
Scope: f5dadc69..df93e192

Reviewed the two rendered frames from the real `OperatorApp` (`/tmp/mcp-fix-eff/settling_view.png`, `settled_view.png`), the toast copy rules in `local_operator/tui/widgets/toast.py::format_mcp_startup`, and the `settling`/`reportable` logic in `local_operator/session/mcp_status.py`.

**Verdict: TERMINAL** — no BLOCKER, no MAJOR. The core UX call is right: suppressing the toast/notice while the round is `settling` and reporting the accurate final tally once is the correct behaviour, and the settling frame is clean (boot splash, no premature "0 of N up — failed: X" alarm). Findings below are MINOR/NIT follow-ups; none blocks merge.

| ID | Severity | Where | Finding |
|----|----------|-------|---------|
| D1 | MINOR | Settled toast, 2nd line — `format_mcp_startup` detail | The toast truncates the single-failure detail at the actionable command: `failed: notion — needs authorization — run /mcp login not…`. The remediation token `/mcp login notion` — the whole reason a failure toast exists — is severed by the ellipsis. **Mitigated**, not lost: the server name `notion` is stated earlier on the same line, and the full untruncated text renders in the transcript notice directly below, so a user can reconstruct the command. But the toast is the interruption and its call-to-action is what gets cut. Consider dropping the redundant lead-in so the command survives the width budget, e.g. `failed: notion — run /mcp login notion` (state auth in the durable notice, keep the fix in the toast). |
| D2 | MINOR | Settled frame evidence — two identical `✗ MCP notion failed: …` notices | The provided settled frame renders the transcript failure notice **twice**. This is a shot-harness artifact — `shot_toast.py:43` calls `_report_mcp_startup` a second time on an already-`settling=False` outcome, on top of the boot path's call — not the product path (boot suppresses while settling; `on_startup_settled` re-reports once). No code change needed, but re-capture the settled frame through the real settle path so the evidence shows the single notice a user actually gets; the current frame reads as a double-emit regression it isn't. |
| D3 | MINOR | Status band during settling — `⊙ MCP` lamp | During the settling window the band lamp is **danger-tinted** (`mcp_semantic` keys off `failed = bool(failures)`, true even in the provisional snapshot) and shows **no count** — just a red `⊙ MCP`. That is a muted version of the very premature-alarm the toast suppression removes: the band flashes "MCP problem" for a server (`notion`) that may still be mid-handshake and about to succeed. It is self-correcting (the lamp greens/updates when the connect lands) and the design note documents the band covering the in-between window, so this is a judgment call rather than a defect — but consider a neutral/"connecting" tint while `settling` so the band and the toast tell the same story. Also worth confirming: the task described the band as "still shows the count" during settling, but the frame shows a bare lamp with no `1 of 2` — the settling band is less informative than intended. |
| D4 | NIT | Settled toast/notice failure line copy | `failed: notion — needs authorization — run /mcp login notion` carries two em-dashes on one line, which reads slightly as prose rather than as a scannable failure entry. A colon after the name separates label from cause more crisply: `failed: notion: needs authorization — run /mcp login notion`. Cosmetic; copy is accurate and not misleading. |
| D5 | NIT | Coverage — follow-up | No rendered frame was provided for the all-servers-up settled case (`MCP ready: N servers`) or a multi-failure settled case (`failed: a, b`). The change is scoped to failure suppression and those paths are unchanged, but a one-frame confirmation each would close the state matrix (populated-success, multi-error). Follow-up, not a blocker. |

Copy is accurate, actionable, and audience-appropriate — `/mcp login`, server names, and "needs authorization" are correct technical language for a coding-agent TUI. No internal values leaked, nothing factually wrong, no placeholders. The suppress-then-report behaviour is the right design; ship it and take D1/D3 as the two worth a follow-up look.

damianvtran:

- **D1 (minor) — Fixed in `a52e827`.** The auth-required failure string now leads with the command: `run /mcp login <name> to authorize` (was `needs authorization — run /mcp login <name>`). The toast renders it after a `failed: <name> — ` prefix and then clamps to the card width, so the tail was what got severed — the old wording lost `/mcp login not…` on a 100-cell terminal. Leading with the command keeps the one actionable thing on screen. Re-captured frame: the settled toast now reads `failed: notion — run /mcp login notion to authorize` in full, and the transcript notice reads `✗ MCP notion failed: run /mcp login notion to authorize`.

- **D4 (nit) — Fixed in `a52e827`.** The failure line now carries a single `—` (the toast's own `failed: <name> — <msg>` separator) instead of two, so it no longer reads as a chain of dashes.

- **D2 (minor) — Rejected (not a product defect).** Verified the doubled transcript notice was a capture-harness artifact: `shot_toast.py` called `_report_mcp_startup` on top of the app's own mount-time boot report. The app reports exactly once. Fixed the shot script and re-captured — the settled frame now shows a single notice.

- **D3 (minor) — Rejected (not a product defect).** The bare danger-tinted `⊙ MCP` with no count during settling was also a harness artifact: the `FakeSession` used for the frame has no live `mcp_manager`, so `_mcp_status()` takes the `discovery_failed` branch (`McpStatus(discovery_failed=...)`) rather than the live `configured/connected` count a real session yields. With a real manager the band shows the live server count and its normal neutral/muted ramp during settling, which is exactly the "band still shows the count" behaviour the suppression relies on.

- **D5 (nit) — Deferred.** All-up and multi-failure settled frames are additional state-matrix coverage; the suppression/report logic is identical across them (`format_mcp_startup` already has unit coverage for the multi-failure and clean cases in `test_toast.py`). Noted for follow-up; not gating.

Re-captured settled frame attached to the PR. Head is now `a52e8277`.

damianvtran:
### Agent review — round 3

Reviewer: reviewer (Claude Sonnet 4.5)
Scope: f5dadc69..a52e8277
Mode: Remediation convergence
Delta reviewed: df93e192..a52e8277

**Verdict: Approve — clean, TERMINAL.**

The delta is a one-line copy change to `_auth_required_text` (`local_operator/mcp/manager.py:1604`), reword `needs authorization — run /mcp login {name}` → `run /mcp login {name} to authorize`, plus a docstring explaining the design-review rationale (D1 command-first for truncation, D4 single em-dash), and the two matching test assertion updates. No behavioural or logic change.

Verified:
- **Single source of truth intact.** All three user-facing surfaces still route through `_auth_required_text`: the transcript/toast path (`manager.py:1147`), the `on_auth_required` sink (`:1622`), and `_startup_failures` (`:1739`). One helper, one string — all agree.
- **No caller parses the old wording.** Remaining `needs authorization` occurrences are all safe: log lines (`manager.py:1150,1245,2216,2280`), the new docstring (`:1598`), and TUI layout fixtures (`test_toast.py`, `test_boot_layout.py`) that pass arbitrary literal strings to exercise formatting/centring — none consume the helper's output or substring-match it. The substring-based assertions (`test_auth_required_text_names_the_login_command`, `test_connect_round_reports_auth_required_actionably`) key on `/mcp login <name>`, which the new string still contains.
- **Test updates correct.** `test_manager.py:1485` and `test_session_factory.py:677` assert the exact new string. Ran the 8 auth-handling tests + the settling-snapshot test: all pass.

| ID | Severity | Finding |
|----|----------|---------|
| — | — | No new findings. Delta is a safe copy change; prior rounds remain resolved. |

Round is terminal — no blocker/major findings on head a52e8277.

damianvtran:
### Design review — round 2

- Reviewer: designer subagent (claude-sonnet-4.5)
- Scope: f5dadc69..a52e8277
- Delta reviewed: df93e192..a52e8277

**Verdict: TERMINAL** — no BLOCKER, no MAJOR. The round-1 copy findings are resolved on the current head and the two rejected findings are confirmed as capture-harness artifacts. Ship it.

Verified against re-rendered frames from the real `OperatorApp` (`/tmp/mcp-fix-eff/settled3.svg`, `settling3.svg`) at 100 cells, the reworded `_auth_required_text` in `local_operator/mcp/manager.py:1591-1604`, and the band/report paths in `local_operator/tui/app.py::_mcp_status` (3241) and `::_report_mcp_startup` (3279).

| ID | Severity | Disposition | Evidence |
|----|----------|-------------|----------|
| D1 | MINOR | **Fixed — verified** | The auth-required string is now `run /mcp login <name> to authorize` (was `needs authorization — run /mcp login <name>`). Leading with the command means the width clamp severs the trailing prose, not the call-to-action. Settled frame: the toast reads `failed: notion — run /mcp login notion to authorize` **in full** on a 100-cell terminal, and the durable transcript notice reads `✗ MCP notion failed: run /mcp login notion to authorize` — the actionable command survives on both surfaces. |
| D4 | NIT | **Fixed — verified** | The composed line carries a single `—` (the toast's own `failed: <name> — <msg>` separator); the redundant `needs authorization —` lead-in is gone, so it no longer reads as a chain of dashes. Confirmed in the settled frame. |
| D2 | MINOR | **Rejected — reasoning sound** | Confirmed the doubled transcript notice was a shot-harness artifact, not the product path. `_report_mcp_startup` (app.py:3279) emits exactly one `_system_notice` per failure key, called once at mount (app.py:1693) and once more only on the real settle path (app.py:3226). The round-1 frame double-emitted because the shot script re-invoked the reporter on top of the boot call; the re-captured settled3 frame shows a single `✗ MCP notion failed …` notice. Accepted. |
| D3 | MINOR | **Rejected — reasoning sound** | Confirmed the bare danger-tinted `⊙ MCP` lamp with no count during settling is a `FakeSession` artifact. `_mcp_status` (app.py:3263-3266) falls to the `discovery_failed` branch precisely when `session.mcp_manager is None` — the harness session has no live manager, so it never reaches the live `configured/connected` count arm (3268-3275) a real session yields. With a real manager the band shows the live server count and its normal muted ramp during the settling window, which is the "band still shows the count" behaviour the suppression relies on. Accepted. |
| D5 | NIT | **Deferred — follow-up** | All-up (`MCP ready: N servers`) and multi-failure settled frames remain uncaptured. `format_mcp_startup` already has unit coverage for both cases in `test_toast.py`; the suppression/report logic is identical across them. Non-gating follow-up. |

Copy check: the reworded string reads well and is accurate on all three surfaces it renders on (toast, transcript notice, `/mcp`). `run /mcp login notion to authorize` is a clean imperative — command first, purpose last — correct technical language for a coding-agent TUI, no leaked values, no placeholders. The settling frame (`connecting…` splash, no premature failure toast) confirms the suppression is intact.

All round-1 design findings (D1 minor, D4 nit) resolved on the current head; the two rejected minors (D2/D3) are confirmed harness artifacts, not product defects. No blocker/major design finding remains. **This round is terminal.**

PR #240, opened 2026-08-22
fix(session): preserve a mid-session /model switch across resume

damianvtran:
## Summary of Changes

Bug fix. A `/model <provider>/<id>` switch changes the session's live model spec, but the switch lived only in memory — quitting the TUI (ctrl+c) and coming back with `--resume` replayed the whole conversation onto the **boot default**, silently losing the switched model. This contradicts what the transcript itself shows the user choosing.

The fix journals each switch as a `selected_model` custom transcript entry — the sibling of the existing `active_model_route` entry. That one records where a provider **fallback** routed requests; this one records where the **user** did. Both are written on the edge and read back at construction.

Key design points:

- **Boot-selector guard.** Each row carries the `provider/model_id` the session was constructed with. On resume the restore adopts the row only when that boot selector still matches; a changed boot selection — a `/model default` write, an edited agent profile, or an explicit `--hosting`/`--model` flag on the resume itself — is the newer, more deliberate choice and wins. Without this, a flag the user just typed would silently not work.
- **Dispose-flush contract.** The write goes through the same tracked-task + dispose-flush path the conversation title uses, because the real failure mode is a switch made *moments before quitting*: `dispose` cancels background tasks, so without the flush the journal write is cancelled in flight and the resume opens on the boot default again.
- **Re-derivation, not replay.** The restore rebuilds the spec through `spec_for_target` (the same derivation `/model` itself lands on) rather than replaying a persisted copy, so a registry update between quit and resume is picked up rather than replayed stale. It validates the provider first, because `spec_for_target` returns a `base_url=None` spec for an unknown provider rather than raising — a stranded row is dropped, not adopted onto a spec that cannot serve requests.

Change class: **C1 (standard, pre-authorized)** — a scoped bug fix with a clear rollback (revert one commit), no contract, schema, auth, or migration surface.

## Related Issue(s)

- Related: none filed (bug reported directly).

## Impact

- No breaking changes. Adds one append-only `selected_model` custom transcript entry type; older builds ignore unknown custom entries on replay, and this build tolerates a missing/malformed row by falling back to the boot model.
- No dependency updates.
- No security implications. The entry stores a `provider/model_id` selector and effort level only — no credentials or secrets. Restore validates the provider against the registry before adopting a row.
- Performance: one extra backward `latest_custom` scan at construction (already done for the title and route entries) and one small append per real `/model` switch. `/effort` and sampling-only spec changes take `set_model`'s same-pair early return and journal nothing.

## Testing Details

[...]

PR #241, opened 2026-08-22
feat(session): persist subagents, statuses, and todos across resume

damianvtran:
## Summary of Changes

Bug fix. After quitting the TUI and coming back with `--resume`, the subagent list, each subagent's status, and the todo list all disappeared and never came back — the subagent panel opened empty, the todo panel opened empty, and there was no way to see, let alone continue, a child that was mid-run, failed, or sitting in a settled state when the process exited. This happened even though the child's own transcript survived on disk and was in principle resumable.

The cause is that all three structures lived only in memory:

- `AsyncJobManager._jobs` — the rows the subagent panel and the `jobs` tool paint;
- `SubagentComms._records` — the roster (`hub op='list'`) and the `job_id -> session_dir` mapping `hub op='resume'` relaunches a child from;
- `TODO_STORE[session_id]` — the todo list the panel and the continuation guardrail read.

The fix snapshots all three to the session transcript as custom entries and rehydrates them at construction — the same mechanism wake schedules (`wake_schedules`), the conversation title (`conversation_name`), and the fallback route (`active_model_route`) already use to survive a resume.

Key design points:

- **One coherent roster snapshot.** The job rows and the comms records are folded into a single `subagent_roster` custom entry, written together (newest-wins) so a resume never reads two halves that drifted. It is re-snapshotted on every roster move (register / settle / cancel, via a new `on_roster_change` hook on the manager) **and immediately after a child attaches**, because the child's transcript directory — the whole basis of resume — only becomes known then, and the registration snapshot fired before it existed.

- **`interrupted` is a restore-only status.** A row that was `running` when the process died has no asyncio task any more, so `restore()` downgrades it to a new `interrupted` status rather than presenting a live-looking `running` the manager can never settle. It is resumable when its transcript survived. Restored rows carry a `restored` flag and are exempt from the retention sweep (they hold the *previous* session's settle stamp, already past the window, so an unguarded sweep would evict every rehydrated child on the first pass).

- **Todos are snapshotted only when they move.** The post-turn snapshot is guarded by the same full-list fingerprint the continuation guardrail uses, so an unchanged list costs one tuple comparison and no write. A never-used-todos session writes nothing at all.

[...]

PR #243, opened 2026-08-22
feat(todo): phased todos — grouped panel with collapse/expand and auto-hide

damianvtran:
## Why

The `todo` tool kept a **flat** list: `init`/`add`/`done`/`block`/`drop` over `{text,status}` items with no grouping. On a multi-step task that reads as one long undifferentiated checklist — there is no way to say "these three belong to Foundation, those two to Verification", and a twenty-item plan pushes past the panel's row cap with no structure to collapse against. This adds **phases**: todos group under named phases, the dock panel renders phase headers with their items indented and per-phase progress, settled phases auto-hide after a delay, and a `ctrl+t` hotkey expands/collapses the whole view. It follows the proven model in the omp coding agent (studied directly), adapted to this harness's status vocabulary.

Version bump is **MINOR** (0.20.1 → 0.21.0): new capability, fully backward compatible.

Design doc: `docs/design-phased-todos.md` (grounded in file:line citations of both the omp reference and the four local-operator surfaces).

### Tool + store (`builtin.py`)
- Store shape is now `dict[owner -> list[TodoPhase]]` where `TodoPhase = {"name", "items":[{text,status[,reason]}]}`. **Item dicts are unchanged** and shared by reference between a phase and any flattened view, so `_match_todos`/`_todo_rows`/`_TODO_MARKS` are reused verbatim.
- `init` accepts a phased payload `phases:[{phase, items}]` **or** the existing flat `items` (both set → error). A flat `init` maps to **one implicit phase named `"Todos"`, rendered headerless** — the back-compat lever: existing callers and tests see the identical tool and panel behaviour.
- New optional `phase` addressing field: `add` appends into a phase (lazily created; dedupe applies within the phase); `done`/`drop`/`block` with a `phase` (no `items`) resolve every open item in that phase, idempotently. `done`/`drop`/`block` by `items` still search across all phases.
- `view` groups output by phase with per-phase `(done/total)`; a single implicit phase renders headerless.
- Deliberately **no** new `start`/`unblock`/`rm` ops — this harness has no `in_progress` status and resolves via `drop`, so they would be a separate, larger feature. Op surface stays at six names.

### Guardrail contract (`builtin.py` + `session.py`)
- `open_todos` unchanged in shape (pending item dicts, `blocked` excluded) — now flattened across all phases.
- `todo_fingerprint` grows to a phase-aware **3-tuple** `(phase_name, text, status)` so the no-progress latch detects a phase rename or an item moving between phases.
- **The coupled edit (design §5.3):** `session.py:_stamped_todo_fingerprint` is widened to `len==3` in lockstep. If it stayed at 2 while the source grew to 3, every reminder would expire on every render and the model would be re-nudged every yield — and it fails *silently* (the latch errs safe). The phased no-second-nudge guardrail test is the guard.

### Panel (`todo_panel.py`) + app wiring (`app.py`)
- Phase-header rows (`PhaseName · done/total`, muted+dim) with items indented one gutter; single-phase implicit `"Todos"` stays **byte-identical to today** (verified by diffing before/after SVGs and the untouched `test_band_panels.py` goldens).
- **Row budget preserved:** headers flatten into the render-row list *before* the `cap`/marker arithmetic, so a header can never push the composer off screen under `Screen { overflow: hidden }`; the `… N more` marker counts hidden **items**, not rows.
- **Collapse/expand** via `ctrl+t` (audited free in both the app and Textual's `TextArea`): ported omp's walking-viewport selection (`select_collapsed`, active-phase index), minus the subagent-match axis this panel lacks. Collapsed shows the active phase + a bounded number of following phases; expanded shows everything.
- **Auto-hide** of fully-settled phases after 60s, riding the existing 1 Hz poll (no new timer). **View-only** — the store is never mutated, so `view`/fingerprint/guardrail are unaffected (hidden ≠ resolved). Safety invariant heavily commented: only `done`/`dropped` count as closed, a phase with any pending/blocked item is never hidden (omp's `#isTodoListSettled` bug is the reason).
- A trailing affordance line: collapsed `+N done · ctrl+t to expand`, expanded `ctrl+t to collapse`.

The clickable-affordance variant is deferred to a scoped fast-follow (design §7.6); this ships hotkey-only.

## Impact

- **Backward compatible.** A flat `init` behaves and renders exactly as before. No on-disk migration (the store is process-global in-memory). `_as_phases` coerces any legacy in-process flat list to one implicit phase so no reader can crash on a mixed shape.
- No new dependencies. Additive tool fields (default-empty), one new binding, one panel state pair.
- This PR also **restores the `selected_model` feature** (PR #240) and its 407-line test, which an over-broad edit during development had deleted from `session.py`; only the intended fingerprint-arity change remains there. `test_selected_model.py` passes.

## Testing evidence

**Gates (clean on head `bb5e2c30`):** flake8 ✅ · black==26.1.0 `--check` (424 files) ✅ · isort `--profile black` ✅ · pyright `0 errors` ✅.

**Tests:** `tests/unit/session` + `tests/unit/tools` + `tests/unit/tui/test_todo_panel_phases.py` + `tests/unit/tui/test_band_panels.py` = **842 passed** (includes the restored `test_selected_model`, the 13 new phased tool tests, the phased guardrail tests, and 15 new panel tests). New coverage: phased/flat init + both-set error, add into new/existing/implicit phase, done/drop/block by phase target (idempotent) and by items across phases, `view` grouping, `_as_phases` coercion, the **phased no-second-nudge** test that catches the §5.3 coupling (verified it bites: reverting the `session.py` filter to `len==2` fails it), `select_collapsed` cases ported from omp, auto-hide (hidden after delay / never with a pending item / store untouched), and `ctrl+t` reveal.

**Real-path proof (not just a green suite):** drove `execute_todo` live — a phased `init` then `done phase=Foundation` — and printed the resulting store, `view` output, `todo_fingerprint` (3-tuple with phase name), and `open_todos`; flat `init` stays headerless with a `("Todos", …)` fingerprint. Confirmed reverting the `session.py` arity re-nudges an unchanged phased list.

**Rendered frames (real `OperatorApp` via `run_test` + `save_screenshot`, then viewed; geometry checked — `virtual_size == size`, no scrollbar, at 100x30 and 100x14):**

Multi-phase **collapsed** — phase headers with `done/total`, indented items, strikethrough done, blocked reason, `+2 more · ctrl+t to expand`:
collapsed

Multi-phase **expanded** (`ctrl+t`) — reveals the hidden done row, affordance flips to collapse:
expanded

**Auto-hide mid-sequence** — a settled phase collapses to its header while open phases keep their items, `+3 done` counts hidden items:
mid-hide

**After further settling** — `+2 done`, only the open `Auth` phase's items remain:
after-hide

The single-phase headerless case was captured too and diffs **byte-identical** to the before-frame (the back-compat guarantee).

## Checklist
- [x] Tests added and passing.
- [x] Visual validation captured (before/after, geometry-checked).


damianvtran:
### Design review — round 1

**Reviewer:** designer subagent (Claude Sonnet 4.5, Local Operator — lopdev team) — independent of the two authoring agents (backend coder, panel coder).
**Scope:** `origin/main..bb5e2c301855ff22a39f810b6a9d6c7ec45fc0ef`

This is a design/visual-UX round on the dock todo panel. I judged **rendered frames**, not source: I re-rendered eight fresh frames via the real `OperatorApp` (`run_test` + `save_screenshot`, per AGENTS.md §Visual validation — not `_PanelHost`, so the stylesheet applies) at **100×30** and **100×14** for multi-phase collapsed, multi-phase expanded, phase mid-hide, after-hide, and the single-phase headerless case; then rendered each SVG and looked at it. Geometry checked on every frame: `virtual_size == size`, `show_vertical_scrollbar == False`, no clip. These match the four committed PNGs in the PR body.

Frames captured (states): collapsed multi, expanded multi (reveals struck done rows), collapsed multi @100×14, expanded multi @100×14, mid-hide (Foundation hidden), after-hide (Foundation+Auth hidden), flat headerless @100×30 and @100×14.

---

**D1 — MAJOR — the root `Todos · N/M` line reuses the phase-header `Name · done/total` grammar, but N/M is a phase index, not a completion fraction.**
The root progression line renders `Todos · 2/3` in exactly the same visual grammar (muted name · dim `X/Y`), same column (0), same styling as the phase headers directly beneath it (`Auth · 0/3`, `Verification · 0/2`). But the root's `2/3` means *"on phase 2 of 3"* while every header's `X/Y` means *"X items done of Y"*. Two different denominators wearing identical clothes, stacked adjacently.

This is not hypothetical: in the **mid-hide** frame the root reads `Todos · 2/2` sitting directly above `Verification · 0/2`, which still has open items — a reader scanning the fractions sees `2/2` at the top and reasonably concludes the list is finished, when it is not. That is the panel actively miscommunicating completion, which is the one thing a progress panel exists to get right.

*Concrete fix:* distinguish the root's number from an item fraction. Options, cheapest first: (a) label it — `Todos · stage 2/3` / `phase 2/3`; (b) use a different separator/spelling so it doesn't parse as `done/total` (e.g. `Todos · 2 of 3 phases`); (c) drop the fraction from the root and let the per-phase headers carry the counts. Any of these breaks the visual collision. omp's `Todos · i/n` is the cited source, but omp's non-active phase headers are not `done/total` fractions in the same column, so the collision is introduced here.

---

**D2 — MINOR — the root line and the phase headers are visually indistinguishable, so the root reads as a fourth phase rather than a summary.**
`Todos`, `Auth`, `Verification` are three left-flush lines with identical ink (muted name + dim ` · ` + dim fraction) and zero indent between them. Nothing marks `Todos` as the parent/summary of the phases below it — it reads as a peer phase literally named "Todos". The item indent does correctly subordinate items to their phase header, so the header→item hierarchy is fine; it's the root→phase level that's flat. *Fix (pairs with D1):* give the root a distinct treatment — a bright/`fg` name, a rule, or a leading glyph — so the two tiers read as two tiers. If D1 is fixed by relabelling the root, that relabel largely resolves this too.

---

**D3 — MINOR — the affordance line is fully `dim` (4.18:1), and it is the *only* signal that `ctrl+t` exists.**
`+3 done · ctrl+t to expand` is a discoverability hint for an otherwise invisible hotkey, rendered at the least legible contrast in the band (the design comment even calls it out at 4.18:1). In the frames it is genuinely faint — a user who never squints at that line never learns the panel collapses/expands. The `+N done` count is fairly chrome and can stay dim, but the `ctrl+t to expand` / `ctrl+t to collapse` token is the affordance and should step up to `muted` (7.93:1), the same call `_item_row` makes for the blocked reason. *Fix:* render the hotkey token at `muted`, keep the hidden-count prefix `dim`.

---

**D4 — NIT — expanded view can still clip items yet says `ctrl+t to collapse`, leaving no route to the hidden rows.**
At 100×30 expanded, the row budget dropped Verification and the affordance read `+2 more · ctrl+t to collapse` — correct and honestly confessed, but "expand" that still hides two items with the only offered action being "collapse" is a small dead end (resize is the only recovery). Acceptable given the `Screen { overflow: hidden }` budget; worth a word in the affordance that distinguishes budget-clipped from collapse-hidden, or a follow-up when the clickable-affordance fast-follow lands.

---

**D5 — NIT — phase-header name shares the exact `muted` ink of pending item text.**
Header name and open item body are both `muted`; only the 2-space indent and the `- [ ]` marker separate a header from its items. It works in every frame I captured (the indent + marker carry it), but the header has no extra weight to anchor the group. Leave as-is unless D2's root/parent treatment gives you a natural place to also lift the header a notch.

---

**Things I checked that are clean:**
- **Two-space indent** — correct amount at both widths; items sit clearly under their header and nothing clips at 100×14.
- **Strikethrough + dim on settled items** — reads as a record, not an instruction; the unstruck `— dropped` / `— blocked` tag stays legible on a crossed-out row. Good.
- **Blocked-reason row** — `- [~] Token refresh — blocked: waiting on infra key` is the loudest row (`fg`), reason at `muted`; exactly right for the one row asking for action.
- **`+N done` vs `+N more` split** — the affordance never labels an open hidden item "done"; verified in collapsed (`+3 done`), mid-hide (`+5 done`), after-hide (`+5 done`), and expanded-clipped (`+2 more`).
- **Single-phase headerless case** — `Todos · 2/4 resolved · 1 dropped` with flat items, no phase header; visually unchanged from today's panel (back-compat guarantee holds). At 100×14 it degrades the same as before.
- **Alignment / marker column** — the `- [x]/[ ]/[~]/[-]` gutter stays on the dock rail; `… N more` and truncation ellipsis render as dim chrome.
- **Geometry** — no scrollbar and `virtual == size` on all eight frames; the phase headers are correctly counted in the row budget (composer never pushed off screen).

---

**Verdict: NOT TERMINAL.** One MAJOR (D1) — the root progression fraction colliding with the phase `done/total` grammar can read as "finished" when work remains — should be fixed before this UI merges. D2 pairs with it. D3 is a worthwhile legibility fix for the sole hotkey hint. D4/D5 are follow-ups. Re-render the multi-phase collapsed and mid-hide frames after the D1 fix and I'll close the round.

damianvtran:
### Agent review — round 1

Reviewer: reviewer subagent — Claude (Sonnet 4.5), lopdev team
Scope: origin/main..bb5e2c301855ff22a39f810b6a9d6c7ec45fc0ef

Independent code review of the phased-todos change. I read the design doc and the full diff, exercised the ops and the guardrail paths directly, and ran the gates myself rather than trusting the PR body.

## Findings

**[MINOR] `todo_panel.py:347` / `builtin.py:_todo_view_text` — a single _explicitly-named_ phase renders inconsistently between the dock panel and the `view` receipt.**
The panel chooses the headerless path on phase COUNT alone (`if len(phases) == 1: … _build_flat`), so a store of one deliberately-named phase (e.g. `init phases=[{phase:"Foundation", items:[…]}]`) renders as `Todos · 1/2 resolved` in the dock — the chosen name "Foundation" is dropped. The `view` receipt, by contrast, only goes headerless when `phases[0]["name"] == _IMPLICIT_PHASE`, so the same store yields `Foundation (1/2)` in the transcript. Verified live:
- panel: `'Todos · 1/2 resolved'`
- receipt: `'Foundation (1/2)\n- [ ] a\n- [x] b'`

The design (§5.2) calls the panel/receipt mirror load-bearing; this is a small divergence in that mirror. Not data loss and the marks still agree, hence MINOR. Concrete fix: gate the panel's headerless path on the same predicate the receipt uses — `len(phases) == 1 and phases[0]["name"] == _IMPLICIT_PHASE` — so a lone named phase shows its header in both surfaces (and the byte-identical back-compat case still hits `_build_flat`).

**[NIT] `todo_panel.py:_build_phased` root line — `Todos · {active_idx+1}/{total}` counts only VISIBLE phases.**
`total = len(considered)`, which excludes auto-hidden settled phases, so a 4-phase plan with one hidden reads `Todos · 2/3`. The hidden items are honestly confessed in the affordance's `+N done`, so this is cosmetic, but the stage count understates the real number of phases. Consider counting `total` over all phases (or noting the hidden count in the root line) if the stage fraction is meant to be absolute.

## What I verified

- **§5.3 cross-file coupling is sound and TESTED.** `builtin.todo_fingerprint` emits `(phase, text, status)`, `session.py:_stamped_todo_fingerprint` filters `len == 3` and emits the 3-tuple, and `_todo_reminder_fingerprint` is annotated `tuple[tuple[str,str,str], ...]` — all three agree. I reverted the `session.py` filter to `len == 2` locally and `test_phased_unchanged_list_is_nudged_exactly_once` **failed** (`assert 'decide the domain' in 'prose 0'`), confirming the test actually exercises the stop-nudging path, not just "does it nudge". Restored after.
- **session.py scope is clean.** `git diff origin/main -- session.py` is ONLY the fingerprint arity change + the annotation widening. No unrelated deletions; the `selected_model` feature is intact and `test_selected_model.py` is present and passing.
- **Store/ops correctness** (exercised live): phased `init` (3 items / 2 phases), both-`items`-and-`phases` rejected loudly, `add` dedupes WITHIN a phase (`x` not doubled in A) while the same text in another phase is kept (B keeps its own `x`), `done phase=A` resolves every open item and is idempotent on a settled phase (`No open items in phase 'A'`), text-form `done` matches across phases (`done z` found in B), `add` before any `init` binds the store and lazily creates the named phase. The shared-by-reference invariant holds — `_all_items` mutations write through to the phase.
- **Guardrail** (`test_phased_open_todos_flattens_pending_and_excludes_blocked` + live): `open_todos` flattens pending across phases and excludes `blocked`; auto-hide is view-only (`_hidden_phase_names` never mutates the store) and `_phase_settled` requires ALL items closed, so a blocked/pending item keeps its phase visible — a stalled list cannot read as finished to the guardrail.
- **Panel budget**: headers are flattened into the row list BEFORE the cap (`_build_phased`), tagged `is_item` so the `… N more` / affordance count reflects items not header rows; single implicit-`"Todos"` phase routes to `_build_flat` (byte-identical); the sync equality-guard state is `(fingerprint, budget, expanded, hidden)` — complete, so a `ctrl+t` toggle or an auto-hide crossing repaints. `ctrl+t` binding audited free; `action_toggle_todos` no-ops before the panel exists rather than raising.

## Gates (run locally on this head)

- flake8 (`--max-line-length=100 --extend-ignore=E203`): **pass**
- black==26.1.0 --check: **pass** (424 files)
- isort --profile black: **pass on all tracked files.** One failure on `tests/unit/tui/test_ux_walkthrough_phased.py`, but that file is UNTRACKED (`??`) and not part of this PR's diff, so it will not reach CI — flagging only so it is committed cleanly (or removed) before it lands in a later PR.
- pyright (`--pythonpath .venv/bin/python`): **0 errors**
- pytest `tests/unit/tools tests/unit/session tests/unit/tui/test_todo_panel_phases.py tests/unit/tui/test_band_panels.py`: **842 passed**
- CI on head: lint / type-check / pip-audit green; test (3.12) / test (3.13) still pending at review time.

## Verdict

**clean** — no BLOCKER and no MAJOR findings. This round is TERMINAL for code review: the one MINOR (panel/receipt header for a lone named phase) and the NIT are follow-ups, not a reason for another round. Note the untracked `test_ux_walkthrough_phased.py` should be committed or removed so the isort gate stays green on the next diff.

damianvtran:
### UX review — round 1

Reviewer: ux-reviewer (lopdev team) — Claude Sonnet (claude-sonnet-4.5) driving the real `OperatorApp` via `run_test` + the real `execute_todo` tool
Scope: origin/main..bb5e2c301855ff22a39f810b6a9d6c7ec45fc0ef (`9e00f01..bb5e2c30`)

Interaction review — how the phased-todo flow *feels* to drive, not how it looks or is coded. I walked the real surface end to end, not the source.

#### What I walked (real execution, not stills)
Driving the actual `OperatorApp` at several terminal sizes, with a `ToolContext(session_id="sess")` so the real `execute_todo` tool wrote into the same store the panel reads (no mirroring):

- Seeded a 3-phase / 6-item list through the **real tool** (`init phases=[...]`) → panel painted collapsed with per-phase `done/total` headers, indented items, `+1 more · ctrl+t to expand`.
- Pressed **`ctrl+t` through the real key path** (`pilot.press`) → toggled collapsed↔expanded, affordance flipped `to expand`↔`to collapse`, repainted each time. ✅
- Settled a phase via the **real tool** (`done phase=Foundation`), pushed its `_settled_since` clock past the 60s threshold → the settled phase auto-hid to nothing, root dropped `1/3`→`1/2`, affordance became `+2 done`. ✅
- Pressed `ctrl+t` → expanded **revealed the hidden settled phase** (`Foundation · 2/2` with its `- [x]` rows). ✅
- Verified `ctrl+t` **while typing in the composer**: typed `hello` into the focused `TextArea`, pressed `ctrl+t` → composer text preserved *and* the toggle still fired. `ctrl+t` steals no editing key. ✅
- Flat single-phase `init` → `ctrl+t` is a **safe dead key** (render byte-identical before/after; no header, no affordance). ✅
- `view` receipt vs panel consistency; all-settled-and-hidden state; short-terminal + narrow-terminal degradation; resize repaint.

Everything the change *promises* mechanically works. The findings below are about how the flow reads to a human under real conditions.

---

#### U1 — MAJOR — Root summary line `Todos · N/total` collides with the phase-header `· done/total` format and misreads as completion

On the **initial collapsed paint of a fresh 3-phase list (zero items done)** the panel renders:

```
Todos · 1/3
Foundation · 0/2
  - [ ] wire config
  - [ ] add migration
Auth · 0/2
  ...
```

The root line's `· 1/3` is the *active-phase ordinal* (phase 1 of 3), but it uses the **identical glyph and `x/y` shape** as the phase headers one line below — which mean *done/total*. A user glancing at the top summary reads "1 of 3 done" while every visible phase header says `0/2`, `0/2`, `0/2` (nothing done anywhere). The primary summary line contradicts the detail directly beneath it, on the first thing the eye lands on.

**Concrete step:** seed any multi-phase list with nothing yet completed; the root reads `· 1/3` against `0/N` everywhere below.
**Expected:** the top line should not read as a completion fraction when it is a phase pointer. Fix by differentiating it from the header format — e.g. `Todos · phase 1/3`, `Todos · on Foundation (1/3)`, or an items-resolved fraction (`Todos · 0/6 done`) consistent with the flat panel's own `Todos · n/total resolved`. The flat single-phase panel already spells it honestly as `… resolved`; the phased root regressed to a bare `· 1/3` that means something else.

---

#### U2 — MAJOR — In a short terminal the phased panel degrades to a dead box: zero items **and** no `ctrl+t` affordance, strictly worse than the flat list it replaces

Walking the same 3-phase list at shrinking heights (real app, `run_test(size=(100, h))`):

| height | flat list shows | phased list shows |
|---|---|---|
| 14 | items | `Foundation · 0/2` + one item + `+5 more · ctrl+t to expand` ✅ |
| 12 | — | `Todos · 1/2` + `Foundation · 0/2`, **0 items, no affordance** |
| 10 | `Todos · 0/4 resolved` + 1 item | `Todos · 1/2` + `Foundation · 0/2`, **0 items, no affordance** |
| 8 | 1 item | **0 items, no affordance** |

Because a phase header now counts against the same row budget as an item, a scarce budget is spent entirely on the root line + one header, leaving **no `- [ ]` rows visible and the affordance line clipped**. The panel becomes `Todos · 1/2 / Foundation · 0/2` and nothing else — it looks empty, gives no hint that six todos exist, and the one and only signal that `ctrl+t` would reveal them is gone. The flat panel in the same height still shows an item and its `n/total` count, so this is a regression exactly where the user has least room and most need for the summary.

**Concrete step:** multi-phase list, terminal ≤ ~12 rows tall → dead box.
**Expected:** when the budget cannot fit a single item under a header, prefer showing items over headers (or collapse headers into the root line), and treat the affordance as non-negotiable chrome the way the root line is — clip an item before the "how do I see the rest" line, never the reverse. At minimum the root line should carry the hidden-count so the panel never reads as empty when work exists.

---

#### U3 — MINOR — `ctrl+t` is discoverable *only* via the affordance line; it is absent from `/help`, the footer, and the command palette

The affordance row (`ctrl+t to expand`) is the sole place the binding is advertised, and it only appears on a multi-phase list at adequate height (and vanishes entirely under U2). The sibling hidden binding `shift+tab` at least surfaces its hint in the effort footer and `/model` footer (`shift+tab cycles`); `ctrl+t` has no such fallback. A user on a flat list, or one whose affordance got clipped, has no way to learn the key exists.

**Concrete step:** run `/help` — no keyboard shortcuts are listed; `ctrl+t` appears nowhere outside the panel's own affordance.
**Expected:** mention `ctrl+t` in `/help` (or wherever the app documents `shift+tab`-class bindings), so the gesture is learnable independent of a panel state that may not be on screen.

---

#### U4 — MINOR — The tool's top-line description omits phases, so a model scanning the summary won't discover the phased payload

`build_todo_tool` advertises `"Track a visible task list (init / add / done / block / drop / view)."` and the `init` op description says only `"replace the whole list"`. Phases surface *only* in the `phases`/`phase` field descriptions. A model reading the op summary (as many do before drilling into fields) sees no signal that grouped phases exist or that `init` accepts a `phases:[{phase, items}]` payload. The field descriptions themselves are clear and well-written once read — this is purely about the discovery path.

**Concrete step:** the tool schema's `description` and the `op` field's `init:` clause never mention phases.
**Expected:** one clause in the op description (`init: replace the whole list, optionally grouped into phases`) so a model addressing phases doesn't have to already know they exist to find them.

---

#### U5 — MINOR — Panel and `view` receipt spell the phase header two different ways for the same data

The design makes a point that the transcript `view` receipt and the dock band must "spell one list identically." The item marks do match (`- [x]`), but the phase header does not:

- Panel: `Foundation · 2/2` (middot separator)
- `view` receipt: `Foundation (2/2)` (parenthesised)

Both readable, but a user cross-referencing the transcript receipt against the band sees two formats for the same fact.
**Concrete step:** run `todo view` on a phased list and compare its `PhaseName (done/total)` to the panel's `PhaseName · done/total`.
**Expected:** one header spelling across both surfaces.

---

#### U6 — NIT — When every phase is settled and auto-hidden, the root line collapses to a bare `Todos` with no count

All-settled-and-hidden state renders `Todos` / `+4 done · ctrl+t to expand` — the root's `· N/total` drops entirely (total considered-phases is 0). It reads acceptably ("everything's done, hidden"), but the summary line losing its fraction is slightly bare next to every other state that carries one.

#### U7 — NIT — Auto-hide fires on a fixed 60s wall clock regardless of whether the user is mid-read

A settled phase's row vanishes 60s after settling with only the subtle `+N done` prefix as the after-the-fact signal; a user re-reading that phase sees it disappear under them. The safety invariant (never hide a phase with open work) is correct and well-guarded — this is only about the *view* changing without a foreground cause. Reasonable as shipped; noting it as a follow-up in case field use shows 60s is too eager.

---

### Verdict: NOT TERMINAL

Two MAJOR interaction findings remain: **U1** (the root summary line misreads as completion) and **U2** (short-terminal degradation to a dead box with no items and no affordance). Both are specific and actionable. U3–U7 are follow-ups that do not block on their own.

The core interactions the change adds — `ctrl+t` toggle through the real key path, collapse/expand mental model, auto-hide reveal-on-expand, phase-addressed tool ops, and the composer-typing non-collision — all work as built and were exercised live. Re-request a round once U1 and U2 are worked (or U2 is consciously deferred with a ticket and U1 fixed), against the new head.

damianvtran:

Implementer: coder subagent (lopdev team). Worked the round-1 code, design, and UX findings on the phased-todo panel. Reviewed head at round 1 was `bb5e2c30`; this remediation lands on **`c5127192`**.

Gates re-run on the new head (all clean): flake8 (`--max-line-length=100 --extend-ignore=E203`), black==26.1.0 `--check` (424 files), isort `--profile black`, pyright (0 errors), and the targeted suite `tests/unit/tools tests/unit/session tests/unit/tui/test_todo_panel_phases.py tests/unit/tui/test_band_panels.py` → **845 passed**.

Round-2 evidence frames (real `OperatorApp` + `save_screenshot`, per AGENTS.md §Visual validation) are committed under `docs/evidence/phased-todos/` and viewed: `collapsed.png`, `mid-hide.png`, `expanded.png`, `after-hide.png` (all now show `Todos · stage N/M`), plus new `short-100x12.png` / `short-100x14.png` proving an item survives on a short terminal. Geometry checked: no scrollbar on any frame.

---

## Code review (reviewer round 1)

**MINOR — panel/receipt disagree for a lone explicitly-named phase** — **FIXED** (`713cdfea` receipt + `1dc526f4` panel). The panel's headerless path now gates on the SAME predicate the receipt uses — `len(phases) == 1 and phases[0]["name"] == _IMPLICIT_PHASE` — so a lone named phase (`init phases=[{phase:"Foundation",…}]`) shows its `Foundation · 1/2` header in the dock, matching `_todo_view_text`. The byte-identical flat store still routes through `_build_flat` (implicit `"Todos"`). Verified live: panel now `Foundation · 1/2`, receipt `Foundation · 1/2`.

**NIT — stage total counted only VISIBLE phases** — **FIXED** (`1dc526f4`). The root now spells `stage {active+1}/{len(phases)}` over the ABSOLUTE phase set, not the collapsed `considered` view, so a plan with one auto-hidden settled phase reads `stage 2/3`, not `2/2`. Covered by `test_root_stage_total_is_absolute_not_the_collapsed_view`.

---

## Design review (designer round 1)

**D1 — MAJOR — root line reads as a completion fraction** — **FIXED** (`1dc526f4`). Relabelled `Todos · N/M` → `Todos · stage N/M`. The word breaks the collision with the phase headers' `done/total` grammar in the same column/ink; the number is a phase pointer, not a count. Re-rendered `collapsed.png` and `mid-hide.png`: the mid-hide frame that previously read `Todos · 2/2` above an open `Verification · 0/2` now reads `Todos · stage 2/3`, so it can no longer be misread as finished.

**D2 — MINOR — root indistinguishable from a phase header** — **FIXED (via D1)**, as the review anticipated. `Todos · stage 2/3` no longer parses as a fourth phase named "Todos"; the `stage` token marks it as the summary tier. Left the ink treatment as-is rather than adding a rule/glyph — the relabel resolves the read, and a second visual device would be redundant.

**D3 — MINOR — affordance too quiet** — **FIXED** (`1dc526f4`). The `ctrl+t to expand`/`to collapse` token steps up from `dim` (4.18:1) to `muted` (7.93:1), the same call `_item_row` makes for the blocked reason; the `+N done`/`+N more` count prefix stays `dim` as chrome. Visible in the round-2 `collapsed.png`.

**D4 — NIT — expanded view can clip yet says "collapse"** — **DEFERRED** (ticket LOP-PHASED-D4). Behavioural-taste follow-up; the affordance is honest (`+N more · ctrl+t to collapse`) and resize recovers. Best paired with the clickable-affordance fast-follow (design §7.6), which is already deferred.

**D5 — NIT — header name shares muted ink with pending item text** — **DEFERRED** (ticket LOP-PHASED-D5). The indent + `- [ ]` marker carry the header/item distinction in every captured frame; adding header weight is a taste call best made alongside D2's parent-treatment work if it is ever taken up.

---

## UX review (ux-reviewer round 1)

**U1 — MAJOR — root summary collides with phase `done/total`, misreads as completion** — **FIXED (same as D1)** (`1dc526f4`). Fresh 3-phase list now paints `Todos · stage 1/3` above `Foundation · 0/2` — the top line is unambiguously a stage pointer, not "1 of 3 done". Covered by the updated `test_multi_phase_renders_headers_indent_and_marks` (`lines[0] == "Todos · stage 1/2"`).

**U2 — MAJOR — short terminal degrades to zero items + clipped affordance** — **FIXED** (`1dc526f4`). New `_fit_body` guarantees at least one ITEM row survives when any exists: with a 1-row body budget the walking viewport put a header at `body[0]`, so the old `body[:cap]` painted zero todos; now the header is traded for the first item, mirroring `_build_flat`'s `room == 1 keeps the item` floor. Proven at **100×12** (`Todos · stage 2/3` + `- [ ] Add phase addressing field`) and **100×14** (header + item + affordance) — see `short-100x12.png` / `short-100x14.png`; no scrollbar. Covered by `test_short_terminal_keeps_at_least_one_item_visible`.

**U3 — MINOR — `ctrl+t` undiscoverable** — **PARTIALLY FIXED / half DEFERRED**. The contrast half is fixed (see D3: token → `muted`). The `/help`-listing half is **DEFERRED** (ticket LOP-PHASED-U3): `ctrl+t`'s sibling `shift+tab` surfaces its hint in per-footer strings (`/model`, effort footer), not a central `/help` keymap list, so there is no clean single hook to extend without building one — out of scope for this remediation.

**U4 — MINOR — tool descriptions never mention phases** — **FIXED** (`713cdfea`). The tool `description=` and the `op`/`phase` field descriptions now name phases (`init: replace the whole list, optionally grouped into named phases (pass \`phases\`)`), so a model scanning the summary can discover the phased payload.

**U5 — MINOR — panel vs receipt spell the phase two ways** — **FIXED** (`713cdfea` + `1dc526f4`). Unified predicate (see code-MINOR above) and unified grammar: both surfaces spell a phase header `Name · done/total`. Receipt test updated to assert the middot spelling.

**U6 — NIT — all-settled state drops to a bare `Todos`** — **FIXED (via D1)** (`1dc526f4`). The stage line is spelled over the absolute phase count, which is always present for a multi-phase store, so the all-settled-and-hidden state now reads `Todos · stage 3/3` instead of a bare `Todos`. See `after-hide.png`.

**U7 — NIT — 60s wall-clock can pull a row from a re-reader** — **DEFERRED** (ticket LOP-PHASED-U7). The safety invariant (never hide a phase with open work) is correct as shipped; changing the hide trigger to be interaction-aware is a field-tuning follow-up, not a round blocker.

---

## Housekeeping

- **Untracked `tests/unit/tui/test_ux_walkthrough_phased.py`** — not present in this worktree (`/tmp/lop-phased-todos`); it lived only in the reviewer's own checkout and was never part of this branch. `git status` is clean here. Nothing to commit or delete on this branch. (The untracked `docs/design-phased-todos.md` that WAS here has now been committed in `c5127192`, so the tree is clean except intended changes.)

---

No blocker/major findings remain open: D1/U1, U2, D3/U3-contrast, U4, U5, U6, and the two code findings are fixed on `c5127192`; D4, D5, U7, and U3's help-listing half are deferred with tickets and reasons above. Re-review requested against head **`c5127192`**.

damianvtran:
### UX review — round 2

Reviewer: ux-reviewer subagent (lopdev team) — Claude (claude-opus-4)
Scope: origin/main..c5127192

Walked the real flow end to end on head `c5127192`: the **real** `execute_todo` tool and the **real** `OperatorApp` panel sharing one store (`builtin.TODO_STORE["sess"]`), driven through `run_test` at 100×30, 100×14, and 100×12, with the model-authored `init`/`done`/`view` payloads a user would actually produce and rendered SVG frames inspected (not stills alone). Each round-1 finding re-checked against the behaviour, not the remediation prose.

| ID | Sev (r1) | Status | Evidence |
|----|----------|--------|----------|
| U1 | MAJOR | **FIXED** | Fresh 3-phase `init` paints `Todos · stage 1/3` above `Foundation · 0/2` / `Auth · 0/2` / `Verification · 0/2`. The root line no longer collides with the per-phase `done/total` grammar — `stage 1/3` cannot be misread as "1 of 3 done" against the `0/N` headers beneath it. |
| U2 | MAJOR | **FIXED** | At 100×12 the panel reads `Todos · stage 1/3` + `- [ ] wire the config` — one item row survives, panel never empty while pending work exists (item rows visible: 1). At 100×14: `Todos · stage 1/3` + `Foundation · 0/3` + `- [ ] wire the config` + `+6 more · ctrl+t to expand`. The regression where the user had least room is gone. |
| U4 | MINOR | **FIXED** | Tool `description=` now reads "…optionally grouped into named phases"; the `op`, `phases`, and `phase` field descriptions all name phases (`init: replace the whole list, optionally grouped into named phases (pass \`phases\`)`). A model scanning the schema summary discovers the phased payload without reading source. |
| U5 | MINOR | **FIXED** | `view` receipt and panel header now spell a phase header identically — `Foundation · 0/2` middot grammar on both surfaces. Confirmed by running `view` against the same store the panel renders: receipt `Foundation · 0/2\n- [ ] …` mirrors the dock. |
| U6 | NIT | **FIXED** | All-settled state reads `Todos · stage 3/3` above `Verification · 2/2`, not a bare `Todos`. The stage denominator is the absolute phase count, so a fully-settled multi-phase plan reads coherently. |
| U3 | MINOR | **FIXED (contrast) / DEFERRED (help-listing)** | The `ctrl+t to expand` token steps dim→muted (visible in the rendered collapsed/short frames — the affordance is legibly brighter than the `+N more` chrome prefix). The `/help`-listing half is deferred (LOP-PHASED-U3) with a stated reason: `ctrl+t`'s sibling `shift+tab` also lives in per-footer hints, not a central keymap, so there's no clean single hook — acceptable as a scoped follow-up. |
| U7 | NIT | **DEFERRED (acceptable)** | 60s auto-hide pulling a row from a re-reader is deferred (LOP-PHASED-U7). The safety invariant (never hide a phase with open work) is correct as shipped; making the hide trigger interaction-aware is field-tuning, not a round blocker. |

**Interaction re-exercised on the live panel:**
- `ctrl+t` toggle: collapsed `+2 done · ctrl+t to expand` → `toggle_todos` → expanded reveals the auto-hidden settled `Foundation · 2/2` with its `- [x]` rows and flips the token to `ctrl+t to collapse` → toggle again collapses back to the walking-viewport view. No dead end; the token names the current direction correctly in both states.
- Auto-hide → expand reveals hidden: settling `Foundation` entirely drops it from the collapsed view (`Todos · stage 2/3` skips straight to `Auth`), and `ctrl+t` brings it back — the discover→hide→reveal loop is intact and the stage pointer tracks the active phase (`stage 2/3` while Auth is active).

**New regressions from the round-2 commits:** none found. The stage-pointer relabel and the `_fit_body` floor did not disturb the single-phase headerless back-compat path (`Todos · 1/2 resolved`, flush rows) or the expanded budget.

**Verdict: CLEAN / TERMINAL.** No BLOCKER and no MAJOR remains. U1, U2, U4, U5, U6, and U3's contrast half are fixed on `c5127192`; U3's help-listing half and U7 are deferred with tickets and defensible reasons. The phased-todo flow is discoverable, gives feedback at every step, degrades safely on a short terminal, and keeps its keyboard escape routes. Follow-ups: LOP-PHASED-U3 (surface `ctrl+t` in a central help listing), LOP-PHASED-U7 (interaction-aware hide timing).

damianvtran:
### Agent review — round 2

Reviewer: reviewer subagent (Claude Sonnet 4.5)
Scope: origin/main..c5127192

Round-2 CODE re-review. My round-1 verdict was already clean (no blocker/major); this pass confirms the round-2 commits — `1dc526f4` (relabel + short-terminal floor), `713cdfea` (receipt/schema unification), `cb7aa168` (tests) — introduced no regression, and closes out my two round-1 items.

**Gates (worktree `.venv`, head c5127192):**
- flake8 (`--max-line-length=100 --extend-ignore=E203 local_operator tests`) — clean
- black==26.1.0 --check — clean (424 files unchanged)
- isort --check-only --profile black — clean
- pyright (touched files: todo_panel.py, builtin.py, session.py) — 0 errors, 0 warnings
- pytest `tests/unit/tools tests/unit/session tests/unit/tui/test_todo_panel_phases.py tests/unit/tui/test_band_panels.py` — **845 passed** (65.7s)

**Round-1 items — both addressed:**
- **MINOR (panel/receipt spelling divergence)** — FIXED. `713cdfea` aligns `_todo_view_text` from `PhaseName (done/total)` to the panel's middot `PhaseName · done/total`, and unifies the headerless predicate: both `_todo_view_text` and the panel's `render_lines` now gate headerless on `len(phases) == 1 and phases[0]["name"] == _IMPLICIT_PHASE` — not phase count alone. This also fixes a latent bug where a lone *explicitly-named* phase dropped its header in the dock but kept it in the receipt; new test `test_lone_named_phase_shows_its_header_matching_the_receipt` pins the mirror.
- **NIT (stage total should be absolute)** — FIXED. The root line now spells `Todos · stage N/M` where `total_stages = len(phases)` (the absolute phase set), not the collapsed `considered` view, and `active_stage = _active_phase_index(phases)`. An auto-hidden settled phase can no longer make three stages read as two; `test_root_stage_total_is_absolute_not_the_collapsed_view` asserts `Todos · stage 2/3` with a hidden phase. All-settled falls to `max(0, len-1)` so a fully-done plan reads `stage 3/3`, not a bare `Todos` (U6).

**`_fit_body` budget floor — correct:**
- On a normal terminal `dropped` is empty (nothing past the cap), so the promotion guard `any(dropped items)` is False and the slice is returned untouched — it cannot clip below the header.
- The equality guard is state-complete: when the cap-1 slice is all chrome and an item exists below the fold, the trailing kept header is *evicted into* `dropped` (`dropped = [evicted, *dropped[:first_item], *dropped[first_item+1:]]`) while the first item is promoted into `kept`, so `len(kept)+len(dropped)` is conserved and the affordance's hidden-count confession stays whole. Verified by `test_short_terminal_keeps_at_least_one_item_visible` at 100x12 and 100x14 (≥1 item painted, no silent scrollbar).
- A header-only paint cannot reappear: the floor guarantees a real todo whenever any item exists; the flat path already had this via `_build_flat`'s `room == 1 keeps the item`, and the phased path now mirrors it.

**session.py scope — confirmed clean:** the only net change vs origin/main is the `_stamped_todo_fingerprint` 2-tuple→3-tuple arity widening and the matching `_todo_reminder_fingerprint` annotation, carrying a clear why-comment about the fail-safe latch direction. The earlier over-broad clobber of `selected_model` was already restored (`9f966d91`); no unrelated session code rides along.

**D3/U3 contrast:** `_affordance_row` now paints the `ctrl+t` hotkey token `muted` (7.93:1) while the hidden-count prefix stays `dim` — verified in the diff; contrast is a design-round concern, flagged here only as a no-regression note.

**CI:** lint, type-check, pip-audit green on head c5127192; `test (3.12)`/`test (3.13)` still pending at review time — no failures observed, and the full targeted suite passes locally. Merge should confirm the two test jobs go green.

**Verdict: CLEAN — terminal for the code round.** No blocker, no major, no new regression from the round-2 commits. Both round-1 findings (the MINOR spelling divergence, the NIT absolute stage total) are resolved. Nothing remaining warrants another code round.

damianvtran:
### Design review — round 2

Reviewer: designer subagent (Claude Sonnet 4.6, local-operator lopdev team)
Scope: origin/main..c5127192

Re-reviewed the rendered surface, not the source. All frames captured from the **real `OperatorApp`** (per AGENTS.md — not `_PanelHost`), seeded through the real `execute_todo` tool (`op:init` with `phases`, then `op:done` items), at the widths the round-2 remediation claims. Geometry probed alongside each frame.

Frames captured (this round, fresh against `c5127192`):

| state | size | root line | virtual==size | scrollbar |
|---|---|---|---|---|
| collapsed (3-phase, fresh) | 100×30 | `Todos · stage 1/3` | 98×28 == 98×28 ✓ | none |
| mid-hide (Foundation settled+hidden, Auth open) | 100×30 | `Todos · stage 2/3` | 98×28 == 98×28 ✓ | none |
| expanded (ctrl+t, all phases incl. settled) | 100×30 | `Todos · stage 2/3` | 98×28 == 98×28 ✓ | none |
| mid-hide | 100×14 | `Todos · stage 2/3` | 98×12 == 98×12 ✓ | none |
| short | 100×14 | `Todos · stage 1/3` | 98×12 == 98×12 ✓ | none |
| short | 100×12 | `Todos · stage 1/3` | 98×11 vs 98×10 † | none |

† The 100×12 screen carries `virtual_height 11 > size 10`, but the **flat (non-phased) panel renders the identical `virtual 11 / size 10` at 100×12** — it is the welcome-splash shell at a 12-row terminal, present with or without phases, not a round-2 regression. No scrollbar appears in either case (`show_vertical_scrollbar=False`). Noted as an existing-shell observation, not a finding.

---

#### Round-1 finding statuses

**D1 — MAJOR — root `Todos · N/M` read as a completion fraction** — **FIXED** (`1dc526f4`). The root now paints `Todos · stage N/M`. The `stage` keyword + the reserved `Todos` label break the `done/total` grammar it used to share with the phase headers one line below. Verified on the **mid-hide frame specifically** (the case the round called out): with Foundation fully settled and auto-hidden and Auth open, the root reads `Todos · stage 2/3` directly above `Auth · 1/2` — a stage POINTER, unmistakably not "2 of 3 done", so the panel never reads "finished" over open work. The denominator is the absolute phase count (3), so a hidden settled phase does not shrink it. No BLOCKER/MAJOR remains on this axis.

**D2 — MINOR — stage root insufficiently distinct from phase headers** — **FIXED (adequately)**. Zoomed the collapsed panel: `Todos · stage 1/3` vs `Foundation · 0/2` / `Auth · 0/2` / `Rollout · 0/2`. Two lexical signals now separate them — the reserved `Todos` label (never a phase name) and the `stage` word, which no header carries. The residual is that the root sits at the same flush indent and shares the header's muted-name/dim-progress ink; that is exactly the taste point captured as **D5** (header/parent weight) and deferred. The confusable-grammar collision that made this matter is gone. Distinct enough — accepted.

**D3 — MINOR — `ctrl+t` token too dim to discover** — **FIXED** (`1dc526f4`). The `ctrl+t to expand` / `ctrl+t to collapse` token is now painted `muted` while the `+N more ·` / `+N done ·` prefix stays `dim`. Confirmed in the frame (the hotkey token is visibly brighter than its prefix) and backed by the numbers on the band ground `#1e1a14`: `dim #837c6d` = **4.18:1**, `muted #b5afa2` = **7.93:1**. The token is now the same ink `_item_row` gives the one action-bearing row (a blocked reason). Discoverable.

**D4 — NIT — expanded view can clip yet says "collapse"** — **DEFERRED acceptably** (ticket LOP-PHASED-D4). NIT; the affordance stays honest (`+N more · ctrl+t to collapse`) and resize recovers. Fine to defer.

**D5 — NIT — header name shares muted ink with pending item text** — **DEFERRED acceptably** (ticket LOP-PHASED-D5). NIT; in every captured frame the two-space indent + `- [ ]` marker carry the header/item distinction. Fine to defer, and it is the natural home for D2's residual parent-weight question.

---

#### New regressions from the round-2 commits

None found. Checked specifically:
- **Flat headerless case unchanged**: a single implicit phase still renders `Todos · N/total resolved` with flush (un-indented) `- [ ]` rows and no phase header — byte-for-byte the pre-phases panel. The relabel touched only the multi-phase path.
- **Expanded frame** reveals the auto-hidden settled Foundation phase (`Foundation · 2/2`, both `- [x]`) and flips the token to `ctrl+t to collapse` — consistent grammar, no clipping at 100×30.
- **U5 spelling** (cross-surface): the `view` receipt now emits `Name · done/total` (middot), matching the panel's `_phase_header_row`; the two surfaces spell one fact one way.

---

#### Verdict

**No BLOCKER, no MAJOR remains → this design round is CLEAN / TERMINAL.** D1 and D3 are fixed and verified in rendered frames; D2 is adequately addressed; D4/D5 are acceptable NIT deferrals with tickets. Geometry is clean at 100×30 and 100×14 (virtual==size, no scrollbar, nothing clipped); the 100×12 virtual/size delta is a pre-existing shell characteristic identical on the flat panel, not a phased-todo regression.

Follow-ups (non-blocking): LOP-PHASED-D4, LOP-PHASED-D5.

damianvtran:
### Rebase onto latest main — integration note

`main` advanced with four PRs while this was in review, including **#241 "persist subagents, statuses, and todos across resume"** which touches the same `builtin.py`/`session.py` todo surfaces this PR changes. Rebased `dev-phased-todos` onto `origin/main` (new head `6d190ffb`).

**Conflicts resolved (semantic, not mechanical):**
- **Todo persistence made phase-aware.** #241's `todo_snapshot`/`restore_todos` assumed the flat store; they now round-trip the **phased** shape (`[{"name","items":[…]}]`) and **deep-copy through each phase's `items`** so the snapshot never aliases the item dicts the tool mutates in place. Both run `_as_phases`, so a snapshot written by a pre-phases build still restores as one implicit phase.
- **`_persisted_todo_fingerprint`** widened to the same phase-aware 3-tuple as `_todo_reminder_fingerprint`, matching `todo_fingerprint`'s arity (the persistence latch compares against it).
- Two now-obsolete local commits (a `selected_model` restore and a standalone annotation-widen) dropped during rebase — `selected_model` already landed on main via #240, and the annotation is carried by the phased commit.

**Integration proof:**
- New test `test_phased_todos_survive_a_resume_with_phase_identity` — a two-phase plan (`Foundation`/`Verification`) with a `done phase=Foundation` round-trips through dispose→resume with phase names and per-phase membership intact.
- Main's existing `test_todos_survive_a_resume` updated to flatten across phases (flat `init` restores as one implicit phase).
- **Full unit suite: 5929 passed, 7 skipped.** Gates clean on `6d190ffb`: flake8 · black==26.1.0 · isort · pyright (0 errors).

The panel, the `ctrl+t` interaction, and the rendered frames are unchanged by the rebase — the design and UX round-2 results still hold. A fresh code round covers the persistence integration delta.

damianvtran:
### Agent review — round 3

Reviewer: reviewer subagent (Claude Sonnet 4.5, `claude-sonnet-4-5`)
Scope: origin/main..6d190ffb — rebase-onto-main integration delta only (PR #241 "persist subagents, statuses, and todos across resume" advanced main into the same todo surfaces). The round-2 feature review at c5127192 stands; this round verifies the merge seams.

**Verdict: CLEAN — no blocker, no major. This round is TERMINAL.**

#### What I verified

**1. `todo_snapshot`/`restore_todos` phase-awareness (`local_operator/tools/builtin.py:3874`, `:3904`)**
- (a) Deep copy through each phase confirmed: both build `{"name": phase["name"], "items": [dict(item) for item in phase["items"]]}` — a fresh `items` list of copied item dicts, not `dict(phase)`. Proved empirically: after `restore_todos`, mutating a store item in place does NOT leak into the caller's snapshot (`deep: snapshot item leaked? False`); a simulated shallow `dict(phase)` restore DOES leak (`shallow: ... True`).
- (b) Both run `_as_phases` (builtin.py:3799): a legacy flat list restores as one implicit `"Todos"` phase (`flat->phased: [('Todos', ['old'])]`) rather than corrupting the store.
- (c) `restore_todos` refuses to overwrite a populated live slot (`if TODO_STORE.get(session_id): return` — verified: live "Live" phase survives a stale "Stale" restore) and no-ops on empty phases/empty id (verified: `sess5` never created).
- (d) Lossless round-trip: blocked `reason` and `status` survive snapshot→restore (`lossless reason: True status: blocked`); the phase-identity test additionally pins name + per-phase membership + status.

**2. Persist/load value format (`session.py:4873`, `:4889`)**
- Writer `_maybe_persist_todos` appends `{"items": todo_snapshot(...)}` where the value is now the PHASE list; reader `_load_todo_snapshot` reads `details.get("items")` and hands it to `restore_todos`, which normalises via `_as_phases`. Writer and reader agree on the `items` key with a phased value — no flat-vs-phased mismatch. A pre-PR flat snapshot (flat items under `"items"`) still restores without crashing via the `_as_phases` defence (verified in (b)).

**3. Fingerprint arity consistency**
- `_stamped_todo_fingerprint` (session.py:540) now annotated `tuple[tuple[str,str,str], ...]` with the `len == 3` filter intact and the third element coerced (`str(item[2])`) — matches `builtin.todo_fingerprint`'s `(phase_name, text, status)` 3-tuple (builtin.py:3934). `_todo_reminder_fingerprint` and `_persisted_todo_fingerprint` annotations both updated to the 3-tuple form (session.py:1108, :1119). The persist guard (`fingerprint == self._persisted_todo_fingerprint`) and the reminder latch compare like-arity against the same source.

**4. No collateral loss from the rebase**
- `git diff origin/main -- session.py` shows ONLY phased-todo changes: the fingerprint arity bump (`_stamped_todo_fingerprint` filter + the three annotations). No re-deletion/re-addition of main's features, no duplicated blocks, no conflict markers (grep for `<<<<<<<`/`=======`/`>>>>>>>` = none).
- `selected_model` (#240) present (31 refs); subagent-roster persistence (#241) intact (`jobs.restore`, `_persist_subagent`, snapshot type present).

**5. Test rigor of `test_phased_todos_survive_a_resume_with_phase_identity`**
- Genuinely a multi-phase round-trip: two phases (Foundation w/ 2 items, Verification w/ 1), a cross-phase `done`, full dispose→resume, asserting phase names, per-phase membership, and status. Mutation check: monkeypatching `restore_todos` to a FLATTENING copy fails the test (`assert ['Todos'] == ['Foundation', 'Verification']`), confirming it catches a shape regression, not just presence.

#### Gates (run at head 6d190ffb, own `.venv`)
- flake8 (`--max-line-length=100 --extend-ignore=E203`): clean
- black==26.1.0 --check: clean (427 files unchanged)
- isort --profile black --check-only: clean
- pyright (session.py, builtin.py): 0 errors
- Integration tests (`test_resume_subagents_todos.py`, `test_todo_guardrail.py`, `test_todo_steering_e2e.py`, `test_builtin_tools.py`): **154 passed**
- CI on head: pending at time of review (lint/pip-audit/test 3.12/3.13/type-check queued) — merge gate should confirm green.

No findings. Remaining work is none; this round is terminal.

PR #245, opened 2026-08-22
feat(tui): surface subagent role and effort in the page title and status band

damianvtran:
## Why

Two operator requests about the full-page subagent view:

1. **Title enrichment** — the header showed `Subagent · <label>  <glyph> <status> · <elapsed> · N tools` but not *what kind* of child it is or at *what effort*. A scout reads very differently from a full task, and the effort tier is part of why the page was opened.
2. **Band correctness** — when you switch into a subagent view, the composer band overlays the child's model/context/cost, but its **effort** segment was blank for any child on a *different* model than the parent (the `model_spec` override case), even though we know the child's tier.

Version bump is **MINOR** (0.20.1 → 0.21.0): additive, user-visible feature.

### Plumbing — role + effort recorded on the job at registration
`AsyncJob` gained two fields (`agent_role`, `effort`), stamped in `run_subagent` at **registration** beside `prompt` — not in the runner — so a job still **queued** behind the capacity gate (which may never enter its runner) still shows both. `effort` is recorded independently of `model_label`: the model is resolved from the tier upstream (`Session._resolve_subagent_model`), but a tier does not survive that resolution (two tiers can point at one model, and a child on the parent's own model still ran at a chosen level). `Session._launch_subagent` now passes `effort` through; `run_subagent` takes it as a display-only param.

### Title — role in the breadcrumb, effort with the status group
`Subagent · scout · <label>  ⣽ running · hi · 2m23s`.
- **Role** rides the breadcrumb (an identity fact, read with the surface name). The default `task` role is **suppressed as noise** — every child is a task unless told otherwise. It is the **most disposable** field on the ladder: at each rung the row is tried with the role and then without it, so it yields *before* any status field or the clock (a narrower row never keeps the role at the cost of the state word). Fields still drop **whole**, per the existing ladder discipline.
- **Effort** sits between the state word and the clock (`running · hi · 2m23s`), kept *longer* than the elapsed clock because the tier is part of why the page was opened while the clock only repeats the band. Added to the `_chrome_state` memo key so a page retargeted task→scout does not paint a stale header.

### Band — child's own effort, even on a different model
`SubagentBand` now carries the child's `effort`. `_shown_effort` uses the overlay's own tier when present (authoritative — recorded at launch, true of whatever model the child runs), falling back to the previous inherit-only-if-same-model rule only when the overlay carries none. This fixes concern #2: a different-model child now names its level (`◆ Claude Opus 5 › ▴ hi`) instead of dropping the segment.

**Judgement call — role is title-only, NOT in the band.** The operator's wording for concern #2 was model/effort/etc; role is an identity fact the page title owns, not a property of the band's model segment. The band already names the child by label at the head of its right group, so a role word there would be redundant. Documented in `app.py` / `status_line.py` for the reviewer.

## Impact
No breaking changes. Two additive `AsyncJob` fields (default `None`, same not-recorded convention as neighbours), one additive `SubagentBand` field (default `""`), two additive `show()` params (default `""`). The `hub op='resume'` path in `harness/comms.py` does not yet forward role/effort (a pre-existing gap unrelated to this slice — noted, not expanded).

## Testing evidence

**Static gates (clean on head `6099ee1d`):** `flake8 .` ✅ · `black==26.1.0 --check local_operator tests` (422 files) ✅ · `isort --check-only --profile black` ✅ · `pyright` **0 errors, 0 warnings** ✅.

**Targeted suites (all green):**
- `test_subagent_view.py` + `test_effort.py` + `test_launch_subagent.py` — **94 passed**
- `test_status_line.py` + `test_subagent_stats.py` + `test_band_panels.py` + `test_terminal_title.py` + `test_jobs.py` — **193 passed**

**Full unit suite:** 5847 passed, 7 skipped, 3 failed in 16m40s. All 3 failures verified **unrelated to this change**:
- `test_eval_tool.py::test_background_streams_output_while_it_runs` — fails identically on clean `origin/main` (pre-existing).
- `test_steering_approval.py::test_esc_with_no_children_says_nothing_about_subagents` — false positive from the **worktree path** `/tmp/lop-**subagent**-info`: the test asserts no rendered row contains "subagent", and the cwd band prints the path. Passes on clean main and in a normally-named checkout.
- `test_subagent_concurrency` count race — a real-timing flake under peer-agent machine contention; passes in isolation (0.46s).

**New/updated tests:**
- `test_launch_subagent.py` — role+effort recorded at registration and survive settlement; a plain task records `effort=None`.
- `test_effort.py` — a different-model child shows its OWN recorded tier (was blank); the overlay's own tier wins even on the parent's model.
- `test_subagent_view.py` — title names a non-default role + effort; default `task` role suppressed; role yields before the state word as the row tightens (whole-field drops preserved).
- `test_band_panels.py` — `_Job` fixture gained the two fields.

### Rendered before/after (real `OperatorApp` via `run_test` + `save_screenshot`, scout child on `anthropic/claude-opus-5` @ `hi`, a NON-parent model)

Geometry-checked across 4 consecutive settled frames: `virtual_size == size` (118×28), **no vertical scrollbar**, title byte-identical frame-to-frame (no reflow). Title stays `height:1` / `no_wrap`.

**Before** — title has no role/effort; band shows `Claude Opus 5` with **no effort segment** (different-model child):
before

**After** — title reads `Subagent · scout · … running · hi · 2m23s`; band reads `◆ Claude Opus 5 › ▴ hi › ⌂ …`:
after

## Checklist
- [x] Self-reviewed.
- [x] Tests added that prove the changes.
- [x] Targeted tests pass; full-suite failures verified unrelated (see above).
- [x] Formatting, linting, type checks pass.
- [x] Documentation updated where required (in-code rationale).
- [x] Security considerations addressed (display-only fields; no scope change).


damianvtran:
### Design review — round 1

**Reviewer:** designer (Claude Sonnet 4.5) — automated design/UX review, independent of the implementing agent
**Scope:** `f5dadc6963410c25117090efd08eb402071f7296..6099ee1dc0cf379b758f8dce74c9aad79018a11d` (base..head)

Reviewed the rendered surface, not source alone. Viewed the PR's embedded `before.png`/`after.png`, then reproduced the change against the **real `OperatorApp`** (loads `local_operator.tcss`) via `run_test` + `save_screenshot`, scout child on `anthropic/claude-opus-5` @ `hi` (a NON-parent model), across four settled frames at four widths. Geometry checked per frame: `virtual_size == size`, **no vertical scrollbar** at any width, title `no_wrap` / `height:1`.

Frames captured (title / band):
- **118×28 scout·hi** — `Subagent · scout · audit the ingest path for retries  ⣽ running · hi · <elapsed>` ; band `◆ Claude Opus 5 › ▴ hi › ⌂ …`
- **60×28 scout·hi** — role kept, label truncates (`audit the in…`), full status group kept
- **42×28 scout·hi** — role dropped whole, label `audit the…`, `running · hi` kept, clock dropped
- **118×28 task·med** — default `task` role suppressed, no orphan separator (`Subagent · audit …`), band `▴ med`

*(elapsed reads `100d+` in my captures — a fixture start-time artifact, not the change; the PR's own PNGs show the representative `2m23s`.)*

**Assessment against the five questions**

1. **Legibility / weighting of the new title info** — Passes. The label carries base ink (`fg`) and remains the visual subject; `Subagent`, `scout`, the seams and the whole status group are `dim`/`faint`, so the enrichment reads as qualifiers around the title rather than competing with it. The line reads as one coherent breadcrumb→title→status band, not crowded.
2. **Tone / colour discipline** — Passes. Seams `faint`, values `dim`, label `fg` — consistent with the existing title and inverse of the usage card as intended. Role and effort adopt the same single-tone treatment; no new colour introduced in the title.
3. **Role/effort placement** — Sound. Role rides the breadcrumb (identity, read with the surface name); effort sits with the status group between state word and clock (qualifies the run, mirroring how the band's `▴` segment qualifies the model). A reader parses each without a legend.
4. **Band overlay — which session** — Adequate. The band now names the child's own tier even on a different-model child (the previously-blank case is fixed and visible). "Which session" is carried by the surrounding chrome (full-page child view, highlighted Subagents row, read-only hint) plus the child label in the band's right group — pre-existing framing this PR does not regress.
5. **Narrow-width behaviour** — Passes. Fields drop whole; role yields before any status field or the clock; effort outlives the clock as designed; no overlap, no reflow, no scrollbar at 42/60/118.

**Findings**

- **D1 — MINOR.** In the title the effort tier is a bare, unlabeled token sandwiched between two other `dim` dot-separated tokens: `running · hi · <elapsed>`. The band deliberately prefixes effort with the `▴` category glyph (`▴ hi`) precisely so a two-letter value is not mistaken for a neighbouring segment; the title drops that marker. A reader who does not already know the feature can momentarily read `hi` as a status fragment rather than an effort rung. This is a real legibility trade-off against the title's glyph-light minimalism (the title carries only the spinner glyph, by design), so it is a judgement call, not a defect — flagging for a decision, not blocking. If kept bare, it reads fine once learned.
- **D2 — NIT.** Title effort value is `dim` (uniform with the status group) while the band renders the effort value with brighter/coloured emphasis. This divergence is intentional and documented (title = one tone per group; band = value-coloured), and looks correct in both surfaces — recording only so a future editor does not "unify" them by mistake.
- **D3 — NIT.** The band's effort value inherits the band's own value colour, so on a same-model child the segment now appears where it previously did too — no visual regression observed; the only behavioural change (different-model child now names its tier) is a strict improvement over the prior blank.

**Verdict: TERMINAL.** No BLOCKER, no MAJOR. D1 is a non-blocking design decision (label vs. keep bare); D2/D3 are follow-up notes requiring no change. The user-visible surface is coherent, tonally disciplined, and degrades gracefully across width.

damianvtran:
### Agent review — round 1

**Reviewer:** reviewer subagent (Claude, claude-opus-4) — independent of the author.
**Scope:** `f5dadc6963410c25117090efd08eb402071f7296..6099ee1dc0cf379b758f8dce74c9aad79018a11d`

Reviewed the diff, the plumbing (`harness/jobs.py`, `subagent.py`, `session.py`), both surfaces (`subagent_view._title_row`, `status_line._shown_effort`), and the tests. Re-ran the gates on head `6099ee1d`.

**Gates (all clean on head):**
- `flake8 .` ✅ · `black==26.1.0 --check .` (439 files) ✅ · `isort==5.13.2 --check .` ✅ · `pyright` 0 errors/0 warnings ✅
- Targeted suites: `test_subagent_view` + `test_effort` + `test_launch_subagent` + `test_status_line` + `test_band_panels` + `test_jobs` — **221 passed**.

**Verified correct:**
- Fields stamped at **registration** (beside `prompt`, before the runner) — a queued job returns `agent_role`/`effort` immediately (reproduced). `None`-means-not-recorded convention matches the neighbouring `context_window`/`prompt` fields.
- No re-resolution of the model spec: `Session._launch_subagent` resolves once via `_resolve_subagent_model` and passes the raw tier through for display only; the runner never re-reads it.
- New fields are in the `_chrome_state` memo tuple — a task→scout retarget repaints.
- `_shown_effort`: the overlay's own tier wins whenever set (honest even on a different-model child); falls back to inherit-if-same-model only when the overlay carries none; a different model with no tier still blanks. The parent's level does not leak onto a different-model child (reproduced via `test_effort.py`).
- Fields still drop **whole** — role and effort are appended whole, the label remains the only flex field, floored at 8 cells.

---

**MINOR 1 — the "role yields before the state word" invariant breaks for `completed`/`done` status, and the test that claims to prove it has a coverage gap.**
`local_operator/tui/widgets/subagent_view.py:1178-1232`. The `keep_role in (True, False)` loop is the **innermost** rung, so it is re-offered inside the `keep_word=False` pass. Because the role chrome (`"scout · "`, 8 cells) is *shorter* than a long state word (`" completed"`, ~10 cells), a wordless-but-role-kept row can win over a worded-but-roleless row. Reproduced with a `completed` scout child:
```
w=30: 'Subagent · scout · RetryBu…  ✓'          # role kept, "completed" gone
w=31: 'Subagent · scout · RetryBud…  ✓'
w=32: 'Subagent · RetryBu…  ✓ completed'         # role gone, word back
```
This contradicts the code comment ("the role yields before any status field or the clock") and the design in the PR body. It does not surface for `running`/`failed` (shorter words), which is exactly why `test_the_role_yields_before_the_state_word_as_the_row_tightens` — asserting `if "scout" in row: assert "running" in row` — passes: **it only exercises `running`**. Same gap in `test_the_default_task_role_is_not_printed_as_noise` and the naming test.

Impact is small (2 cell-widths; the never-dropped `✓` glyph still conveys the state, so no fact is truly lost), hence MINOR rather than MAJOR — but the invariant the change sells is not actually held, and the test gives false confidence.

Fix direction: role is meant to be the *most* disposable field, so it should leave before any tail field or the word drops — gate `keep_role=True` on `keep_word and dropped == 0` (role kept only when everything else already fits), rather than re-offering it inside the wordless pass. Then add a `completed`-status case to the ordering test.

**MINOR 2 — two vocabularies land in the same effort segment depending on the code path.**
`status_line.py:1245-1254` + `app.py:_effort_label`. When the overlay carries a recorded tier the band renders `lo`/`med`/`hi` (the `task(effort=…)` tier); when it falls back to the parent (same model, no tier) it renders `high`/`auto`/`low`/`reasoning` (the reasoning-effort label). Under the single `ICON_EFFORT` glyph a reader now sees `hi` on one child and `high` on another for what looks like the same concept. Defensible — the tier is the honest thing known about a different-model child, and the PR argues this — but worth a deliberate call on whether the tier should be mapped into the reasoning-effort vocabulary for one segment. Flagging, not blocking.

**NIT — `harness/comms.py` `hub op='resume'` still does not forward `role`/`effort`** (acknowledged in the PR body as a pre-existing gap). A resumed child will show an empty role/effort until re-registration. Fine to leave as a follow-up ticket; noting so it is not lost.

---

**Verdict: clean.** No BLOCKER and no MAJOR findings — the change is correct on its core claims (registration-time stamping, no re-resolution, different-model effort, whole-field drops, memo key). This round is **TERMINAL**; MINOR 1/2 and the NIT are follow-ups, not a reason for another required round, though MINOR 1 is worth a one-line fix + test before merge since it undoes a documented invariant.

damianvtran:

**Implementer:** coder subagent. **Scope reviewed:** `f5dadc69..6099ee1d`; **remediation head:** `26ca1718`.

Thanks — MINOR 1 is a real invariant break and is fixed; MINOR 2 and the NIT get reasoned dispositions below.

---

**MINOR 1 — role-vs-state-word invariant broke for `completed`/`done` — FIXED in `26ca1718`.**
Confirmed and accepted. The `keep_role` loop being the innermost rung meant it was re-offered inside the `keep_word=False` pass, and because the role chrome (`"scout · "`, 8 cells) is shorter than a long state word (`" completed"`, 10 cells), a wordless-but-role-kept row could win over a worded-but-roleless one at ~30–31 cells.

Applied your suggested fix: `keep_role` is now gated on `keep_word and dropped == 0` (`role_allowed` in `_title_row`), so the role is offered **only** at the least-reduced rung — full state word present, no tail field yet dropped. Every tighter rung tries the row without it, so the role yields before a shortened state word, any dropped tail field, or the clock, for both short and long words.

Verified (real `OperatorApp` via `run_test`, role="scout" effort="hi", across widths 20–120):
```
# completed (long word) — role now gone the moment the word must shorten:
w=30: 'Subagent · RetryBudgetScout  ✓'          # role gone, glyph carries state
w=31: 'Subagent · RetryBudgetScout  ✓'
w=32: 'Subagent · RetryBu…  ✓ completed'         # word back, still no role
w=120:'Subagent · scout · RetryBudgetScout  ✓ completed · hi · …'   # role shown when there is room
# running (short word) — unchanged, still 0 violations
```
Invariant-violation count (`"scout" in row and word not in row`) is **0** for both `running` and `completed` across the full width sweep. Whole-field-drop discipline preserved (fields still leave whole; label stays the only flex field, floored at 8). Geometry re-probed on `26ca1718`: `virtual_size == size` (118×28), no vertical scrollbar, no reflow across consecutive settled frames (only the spinner glyph advances).

Test gap closed: `test_the_role_yields_before_the_state_word_as_the_row_tightens` is now parametrized over `running` (short) **and** `completed` (long). The `completed` case **fails on the pre-fix logic** and passes after — I verified by reverting the gate: `AssertionError: (30, 'Subagent · scout · RetryBu…  ✓') assert 'completed' in ...`. So the new case genuinely guards the regression rather than merely re-asserting the fixed state.

---

**MINOR 2 — two effort vocabularies (`hi` tier vs `high` label) under one glyph — DELIBERATE, kept as-is.**
This is an accepted trade-off, not a defect, and mapping the tier into the reasoning-effort vocabulary would be strictly worse:

- The `hi`/`med`/`lo` string is the **tier the child was launched with** (`task(effort=…)` → `values.subagents.models`), recorded verbatim at registration. It is the *authoritative, exact* thing known about a different-model child — the whole reason the band shows it there instead of the previous blank.
- The `high`/`low`/`auto`/`reasoning` label is the parent's **reasoning-effort setting**, shown only on the inherit path (same model, no recorded tier) where it is genuinely the level in force.
- These are two different concepts that happen to share a segment: a *delegation tier* (which configured model to spend) vs a *reasoning-effort level* (how hard one model thinks). Rewriting `hi`→`high` for the tier case would **invent a claim** — it would assert the child is on the `high` reasoning rung when what we actually know is that it was launched at the `hi` delegation tier, whose target model may have no reasoning ladder at all (`gpt-4.1`). That is precisely the lie the different-model blanking rule exists to avoid; re-introducing it via a vocabulary map trades an honest two-letter token for a plausible-but-unfounded word.
- Under the single `ICON_EFFORT` glyph the two never appear for the *same* child — a given band shows one or the other by path — so the concern is cross-child, and cross-child the string difference is a *feature*: `hi` signals "this child was delegated at a tier" vs `high` "this child inherits the parent's reasoning level." Collapsing them would hide that distinction.

Disposition: keep both vocabularies. Recorded the reasoning in `status_line._shown_effort` already; no code change.

---

**NIT — `hub op='resume'` does not forward role/effort — DEFERRED (pre-existing), follow-up filed in-thread.**
Correct, and acknowledged in the PR body. `harness/comms.py`'s resume path calls `run_subagent` without `agent`/`effort` (it never passed `agent` either — pre-existing), so a resumed child shows an empty role/effort until it re-registers. Out of scope for this slice (which is about *recording and displaying* the facts, not the resume plumbing), and fixing it correctly means threading the original launch's role/effort off the persisted `SubagentRecord` — a change to the comms/record surface that deserves its own PR + test rather than a drive-by here. Deferring as a follow-up; noting it so it is not lost. Happy to pick it up as a separate ticket.

All gates re-run clean on `26ca1718`: `flake8 .` ✅ · `black==26.1.0 --check` (422 files) ✅ · `isort --profile black` ✅ · `pyright` 0/0 ✅ · targeted TUI+harness suites **288 passed**.

damianvtran:

**Implementer:** coder subagent. **Remediation head:** `26ca1718`. Thanks for the rendered pass across four widths — the drop-ladder frames are exactly the evidence this needed.

---

**D1 — MINOR — title effort is a bare `hi` token vs the band's `▴ hi`. DECISION: keep it bare in the title. No code change.**

Reasoning, weighing your legibility point against the title's established grammar:

- **The title is deliberately glyph-light; the band is deliberately glyph-vocabulary'd.** The title carries exactly one glyph — the spinner — and renders every other value as a **bare dim token separated by `·`**: the elapsed clock (`2m23s`) and the tool count (`3 tools`) already sit with no category marker, disambiguated purely by position and the dot seams. The band, by contrast, is a dense single row where segments do not have positional context, so it gives each a category glyph (`◆` model, `▴` effort, `⌂` cwd, `◍` agents). Adding `▴` to the title would import the band's vocabulary into a surface whose whole design is that it does **not** need it — it would be the *only* categorized token on the row, reading as more special than the clock beside it, which it is not.

- **Position already disambiguates.** Effort sits between the state word and the clock: `running · hi · 2m23s`. The reader has just parsed `running`; `hi` in the qualifier slot after it reads as "a qualifier of the run," which is what it is. The momentary "is `hi` a status fragment?" reading you flag is real for a first-time reader, but it resolves on the same glance the clock does — and the title is a surface a user sees on every subagent open, so it is learned once, not repeatedly puzzled over. The cost of the marker (breaking the row's one-glyph minimalism, every open, forever) outweighs the one-time learning cost.

- **Consistency beats local optimization.** The title's rule is "seams `faint`, values `dim`, one tone per group, glyph only where motion lives." `hi` as a bare dim token obeys that rule; `▴ hi` breaks it. If a future reader finds the bare token genuinely ambiguous in real use, the right fix is a marker on *all* the qualifier tokens or none — not effort alone — and that is a larger title-grammar decision than this PR should make unilaterally.

So: bare token, consistent with `2m23s`/`3 tools`, matching the title's minimalism. The band keeps `▴ hi` because the band's grammar needs it. The two surfaces render the same fact in each surface's own idiom, which is correct, not inconsistent.

**D2 — NIT — title effort `dim` vs band effort value-coloured. No change (as you noted, intentional and documented).**
Confirmed: this is the same title-vs-band grammar split as D1 — the title uses one tone per group (`dim`), the band colours values against its glyphs. Both look correct in their surface. Left as-is; the `status_line` comment already guards against a future "unify" mistake.

**D3 — NIT — same-model child effort segment placement. No change (strict improvement, no regression).**
Agreed — the only behavioural change is the different-model child now naming its tier instead of blanking, which is the fix. No visual regression on the same-model path.

---

Net: no visual change from this remediation (D1 decided keep-bare, D2/D3 no-change), so the title renders identically to the PR's embedded `after.png` at normal widths — the `26ca1718` fix only affects the ~30–31 cell narrow band for long status words (role now drops there), which is below the widths the evidence frames show and is covered by the parametrized ordering test. Re-verified geometry on `26ca1718`: no reflow, no scrollbar, `virtual_size == size`.

damianvtran:
### Agent review — round 2

Reviewer: reviewer (Claude Sonnet 4.5)
Scope: 6099ee1d..26ca1718

Freshness round on the single remediation commit `26ca1718` (`fix(tui): keep the subagent-title role strictly the most disposable field`), which addresses round-1 MINOR 1. Round 1 was TERMINAL; this round reviews only the delta.

**Verified — the fix restores the invariant without regressing the ladder.** `_title_row` now gates the keep-role offer on `role_allowed = bool(role_seg) and keep_word and dropped == 0` instead of trying it as the innermost rung. Offering the role only at the least-reduced rung (full state word present, no tail field dropped) means the role can never win a width by displacing a longer state word. The `keep_word`/`word_choice`/`dropped` ladder itself is untouched, so the running-status shortening ladder and the whole-field-drop discipline are preserved — the role is the only thing whose availability changed.

**Verified — the parametrized test genuinely guards the regression.** I ran the new `completed` case against the pre-fix source (`6099ee1d`) in a throwaway worktree: it fails at exactly width 30 with `Subagent · scout · RetryBu…  ✓` — role kept, `completed` word gone — the precise inversion the fix targets. The `running` case passes on both pre- and post-fix source, confirming it is the short-word control and the `completed` case is the actual guard. On head `26ca1718` both cases pass.

**Gates on head 26ca1718 (changed files):**
- flake8 — clean
- black==26.1.0 --check — clean (2 files unchanged)
- isort --profile black --check-only — clean
- `tests/unit/tui/test_subagent_view.py` title/effort/status/role suite — 8 passed

Comment quality holds: the `role_seg` and loop comments now explain *why* gating beats innermost-trying (role chrome `scout · ` at 8 cells is shorter than ` completed` at 10 cells, so an innermost offer inverts the disposability order at ~30 cells), which is the load-bearing rationale a future editor needs.

No new findings. MINOR 2 (two effort vocabularies — deliberate) and the resume NIT (deferred) were dispositioned in round 1 and are untouched by this commit.

**Verdict: clean.** No blocker, no major. This round is TERMINAL — the round-1 disposition stands and the delta is sound against the current head.

damianvtran:
### Agent review — round 3

Reviewer: reviewer subagent (Claude Sonnet 4.5, via local-operator lopdev team)
Scope: 26ca1718 (round-2 head) rebased → b92c92bb — integration with PR #241 resume persistence

Focused re-review of the rebase-resolution delta only. Rounds 1–2 were terminal-clean; the title-ladder / band-effort feature logic is unchanged and was not re-examined.

**What I checked**

1. **`harness/jobs.py` — field block (conflict resolution).** Both sides survived intact. PR #241's `restored: bool = False` (jobs.py:187) and this PR's `agent_role`/`effort` (jobs.py:200–201) coexist in the "child accounting" block, each with its own comment; nothing dropped, nothing duplicated. `agent_role` is stamped at registration in `harness/subagent.py:262` and `effort` passed through `_launch_subagent` → `run_subagent` — the runner never has to exist for either to be recorded, matching the `prompt` convention the comment cites.

2. **`session.py` — `_ROSTER_ROW_FIELDS` allowlist (the critical seam).** Correct. Both fields belong by the allowlist's own stated criterion: small, bounded, enum-like strings that the panel/page/band read off the restored row — the same class as `model_label`/`context_window`/`usage` already on the list. Verified the persistence path needs no special handling beyond the allowlist entry: `_subagent_job_row` projects via `model_dump(include=_ROSTER_ROW_FIELDS)` (session.py:745), `AsyncJobManager.restore` (jobs.py:325–338) stores whole `AsyncJob` rows verbatim with no per-field logic, and rehydration is a plain `AsyncJob.model_validate(raw)` (session.py:4836). `_parsed_usage` touches only the nested `usage` payload and is unaffected. So the round-trip is automatic once the two fields are on the allowlist — and the allowlist is the only gate that would silently blank them.

3. **`tests/…/test_role_and_effort_survive_a_resume` — regression guard.** Exercises the real path: a real `_session()` (production `Session` over a `Transcript` dir), real `_launch_subagent(agent="scout", effort="hi")`, real settle-wait, `dispose()`, then a fresh `Session` over the same dir asserting `agent_role == "scout"` and `effort == "hi"`. No mocks on the launch/restore seam. The non-default role + explicit tier are well chosen — a blanked field would read as a default, not the value under test. I independently reproduced the negative: removing both allowlist entries makes the test fail with `AssertionError: assert None == 'scout'` (`agent_role=None` on the restored row), confirming it genuinely guards the seam the rebase exposed. Restored the file after; tree clean.

**Gates (run on the resolved files + resume/subagent suites)**
- `flake8 harness/jobs.py session.py` → clean
- `pyright harness/jobs.py session.py` → 0 errors, 0 warnings
- `pytest test_resume_subagents_todos.py test_launch_subagent.py test_effort.py test_subagent_view.py test_jobs_restore.py -q` → **110 passed**

**Findings:** none — no BLOCKER, no MAJOR, no MINOR, no NIT on the rebase-resolution delta.

**Verdict: clean.** The conflict resolution kept both features' fields intact, the allowlist addition is the correct and sufficient integration point, and the new test is a genuine regression guard on the real path. This round is TERMINAL.

damianvtran:
**Note (post-round-3):** the round-3-reviewed test file was reflowed by `black==26.1.0` (one `_launch_subagent(...)` call joined to a single line — the whitespace CI's lint job flagged). Content and assertions are byte-identical to the round-3 review; the test still passes and still fails when the allowlist entries are stripped. Head is now `71d392be`. No logic changed, so the clean round-3 verdict stands.

PR #246, opened 2026-08-22
feat(tui): autofill team/agent name completions and highlight recognized slash commands

damianvtran:
# PR Description

## Summary of Changes

New feature + UX fix for the composer, addressing two reported rough edges in the `/team` and `/agent` completion flow and a broader legibility gap. Version bump is **MINOR** (0.20.1 → 0.21.0).

**1. Team/agent name completion autofills instead of submitting.** Selecting a `/team` or `/agent` name suggestion — by Enter, by click, or by Tab — used to *run and submit* the line immediately, which just attached/switched the team or agent with no message. Now every one of those gestures completes the name, appends a trailing space, and parks the caret so you can type the request you actually want to send. The pre-existing behaviours still work unchanged because they ride the same dispatch:

- **Blank followup + Enter** (`/team frontend-guild ` with nothing after the space) → attach-only, exactly the old "just switch" outcome. `_run_slash_command` collapses the trailing space (`arg = parts[1].strip()`) into the bare-name branch of `_cmd_team`/`_cmd_agent`, which attaches and prints the "ready" notice.
- **Typed message + Enter** (`/team frontend-guild fix the flaky test`) → attaches and sends the message to the manager as a real user turn.

Enum-tail commands (`/login`, `/logout`, `/effort`, `/approvals`, `/model`, …) are untouched — their argument *is* the whole command, so they keep completing-in-place / running as before.

**2. Composer syntax highlighting for recognized slash commands.** The composer now tints the leading `/command` token when it resolves against the registry (the whole slash surface, not just team/agent), tints the recognized team/agent **name** token, and leaves the free-text message as plain prose — so you can see at a glance what is a command and what will be sent as the message. An **unrecognized** command word (e.g. `/notacommand`) renders muted, and the unknown treatment is suppressed while the command picker is still open so a half-typed `/te` never flashes "unknown" mid-type.

## Design

Full design doc committed at `docs/design/composer-ux-autofill-and-highlight.md`. Key decisions:

- **`NAME_ARGUMENT_COMMANDS`** editor-local tuple (mirrors the existing `MODEL_COMMANDS` idiom) marks `team`/`teams`/`agent`/`agents` as "name + free-text message" commands. Deliberately **not** a new `ArgumentMode` enum member — `arguments` is read by `mobile/daemon.py` and `set_commands`/`opens_a_list`, so a shared-enum change carries blast radius a two-command need doesn't justify.
- Highlighting reuses the existing `_paint_markers`/`_overlay` `post_style` render seam via a new `_paint_slash` pass chained in `render_line` — **foreground-only** styles that sidestep the caret/selection overlay interaction, following the theme's `$lo-*` variables (not `$lo-accent`, which is reserved) so a theme switch is free. Recognition is computed cheaply with a hot-path bail (only runs when line 0 starts with `/`); the team/agent name set is pushed from the app in `on_argument_query_opened` so the render path never does registry I/O.

## Related Issue(s)

- Related: user-reported composer UX feedback (no tracking issue).

## Impact

- No breaking changes. No dependency updates.
- No security implications — pure TUI rendering + completion behaviour.
- Performance: the highlight pass bails immediately unless line 0 begins with `/`; recognition reuses the existing single-line tokenizers and an app-pushed frozenset (no per-frame registry work).

## Testing Details

**Quality gates (all clean, run against the worktree):**

```
flake8 .                              → 0 findings
black --check (26.1.0)                → 423 files unchanged
isort --check-only --profile black    → clean
pyright                               → 0 errors, 0 warnings
pytest tests/unit/tui/test_command_picker.py → 231 passed
```

**15 new tests** in `tests/unit/tui/test_command_picker.py`, split across the two changes:

- *Autofill* — Tab/Enter/click on a `/team` and `/agent` name row each assert `host.submissions == []` (completion never submits) and that the buffer holds `…<name> ` with the caret at end; blank-Enter-attaches and typed-message-sends assert the correct submitted text; regression guards pin the login/logout enum-tail rows to their old run/complete behaviour.
- *Highlighting* — render-strip assertions that a recognized command word gets the `slash-command` style, an unknown word gets the muted `slash-unknown` style only when the picker is closed, a recognized team name gets the `slash-argument` style, a partial/unmatched name is not highlighted, and a second buffer line is never highlighted.

**Visual validation** (real `OperatorApp` + `save_screenshot`, before/after SVG pairs, before-frames captured from a detached worktree — never `git stash`):

| Frame | What it proves |
|-------|----------------|
| 1 · team list open | picker unchanged, list renders |
| 2 · name selected | caret parked **after** the trailing space, line **not** submitted |
| 3 · team with message | tri-colour: `/team` (command) · `frontend-guild` (name) · `fix the flaky test` (prose) |
| 4 · unknown command | `/notacommand` renders muted, `hello there` prose |
| 5 · recognized non-name command | command word tinted, whole slash surface covered |
| 6 · agent mirror | `/agent` behaves identically to `/team` |

Rendered PNGs of every after-frame were viewed; highlight hues are legible on the dark theme and no reflow/scrollbar was introduced between before/after. Artifacts: `/tmp/composer-shots/{1..6}-*-{before,after}.svg` and `view-{1..6}-after.png`.

## Checklist

- [x] I have updated the documentation when required (design doc added).
- [x] Security considerations are addressed (no code-execution surface touched).

## Screenshots

After-frames (rendered from the real app):

- Autofill parks the caret after the space (frame 2), highlighted message (frame 3), unknown command muted (frame 4) — attached in the agent review thread; SVG/PNG artifacts under `/tmp/composer-shots/`.

### DIFF
--- a/docs/design/composer-ux-autofill-and-highlight.md
+++ b/docs/design/composer-ux-autofill-and-highlight.md
@@ -1,0 +1,621 @@
+# Composer UX: name-autofill for `/team`·`/agent`, and slash-command syntax highlighting
+
+Design for two related composer changes. **No production code here** — this is
+the seam the coder implements. All line numbers are against the worktree
+`/tmp/lop-composer-ux` (branch `dev-composer-ux`, cut from `origin/main`) as
+read during this design pass; treat them as "the method named X near line N",
+not as offsets to patch blindly.
+
+Ground-truth note: the manager's brief cited older line numbers (editor
+`_run_argument` ~1720, `_apply_command` ~1654, etc.). The **methods are all
+real and unchanged in behaviour**; only the numbers drifted. Current numbers
+are used throughout below. Nothing in the brief's architecture was found wrong.
+
+---
+
+## 0. What the code actually does today (verified)
+
+**Registry** (`local_operator/tui/autocomplete.py`):
+- `ArgumentMode` enum (lines 29–58): `NONE` / `OPTIONAL` / `REQUIRED`. Its
+  docstring is explicit that the mode answers two questions: does a space open
+  the value list, and may Enter on the command ROW also send it.
+- `SlashCommand` frozen dataclass (lines 67–112) with kw-only `echo` (98) and
+  `arguments` (107). `.names` property (109–112) = primary + aliases.
+- `slash_command_for(text)` (526–543): the ONE resolver from a typed line to a
+  registry entry, alias-aware, case-insensitive. This is the "is this a
+  recognized command" oracle the highlighter will reuse.
+
+**team/agent entries** (`local_operator/tui/app.py`):
+- `/team` (502–507) and `/agent` (515–522), both `ArgumentMode.OPTIONAL`, both
+  with aliases (`teams` / `agents`).
+- Argument rows filled in `on_argument_query_opened` (9481–9558): `team`/`teams`
+  → `_team_choices()` (9530–9532), `agent`/`agents` → `_agent_choices()`
+  (9534–9536).
+- Dispatch: `_run_slash_command` (7307) parses `arg = parts[1].strip()` (7327),
+  so any trailing space on the name is already collapsed before it reaches
+  `_cmd_team` (2829) / `_cmd_agent` (3027). Both `partition(" ")` the arg into
+  `name` + `request`; **empty request → attach-only + "ready" notice** (team
+  2884–2888; agent falls through attach then notices). **Confirmed: `/team foo `
+  + Enter already means attach-only. No dispatch change is needed.**
+
+**Editor completion** (`local_operator/tui/widgets/editor.py`):
+- Key handler (1088–1147): on Tab/Enter with a highlighted row, computes
+  `unambiguous = _picker_choice_is_unambiguous(name)` (1108) BEFORE completing,
+  then:
+  - argument mode → `_resolve_argument(name, key, unambiguous)` (1110)
+  - command-word mode, unambiguous → `_apply_command(name)` (1112), and if
+    `enter and not opens_a_list(name)` → `_submit()` (1120–1121)
+  - command-word mode, Tab → `_apply_command(name)` (1125)
+  - command-word mode, ambiguous Enter → grow common prefix (1126+)
+- `_apply_command` (2317): argument mode → destructive `_complete_argument`
+  else `_run_argument` (2339–2344); command-word mode → writes `/{name} ` and
+  moves caret to end (2350–2351).
+- `_resolve_argument` (2353): Tab or ambiguous → `_complete_argument`; else
+  `_run_argument` (2362–2365).
+- `_complete_argument` (2367): replaces the argument tail with `name`, **NO
+  trailing space** (2380), because a space terminates the arg and closes the
+  list.
+- `_run_argument` (2383): `_complete_argument` then `_submit` (2390–2391).
+- `set_commands` (887) derives `_argument_commands` (896) and
+  `_required_argument_commands` (902) from the registry. `opens_a_list` (909)
+  reads `MODEL_COMMANDS` (566) ∪ `_required_argument_commands`.
+- Render: `render_line` (709) → placeholder branch, else
+  `_paint_markers(super().render_line(y), y)` (748). `_paint_markers` (1664)
+  post-processes the finished `Strip` via `_overlay`/`Segment.apply_style(...,
+  post_style=...)` (1694–1697). `_marker_cells` (1566) computes which cells to
+  repaint, with a hot-path bail `if "[" not in line: return []` (1595–1599).
+- `COMPONENT_CLASSES` (578–581): `text-area--image-marker[-selected]`. tcss at
+  `local_operator/tui/local_operator.tcss` 459–467, using `$lo-signal`,
+  `$lo-tint-attach`, `$lo-fg`, `$lo-tint-attach-hi`.
+
+**Empirically verified** (ran `slash_argument`/`argument_suggestions` against
+the real functions): after we complete to `/team frontend-guild ` the picker is
+in ARGUMENT mode with an **empty match set** (list closes), and a subsequent
+blank Enter — with `is_open()` false and `is_pending()` false once the query is
+non-`None` — falls through to `_submit()`. Typing a message (`/team fg fix the
+bug`) keeps `slash_argument` non-None with zero matches, so the list stays
+closed and Enter submits the whole line. This is the behaviour ask #1 wants,
+and it already emerges from the tokenizer — we only have to make completion
+INSERT the trailing space for these commands.
+
+---
+
+## A. Interaction spec (ask #1)
+
+Two command *classes* from the completion's point of view:
+
+- **ENUM-tail** (`login`, `logout`, `effort`, `approvals`, `theme`, `mcp`,
+  `credential`, and the model commands): the argument IS the command. Current
+  behaviour is correct and must not change.
+- **NAME+message** (`team`, `teams`, `agent`, `agents`): the argument is a NAME
+  optionally followed by a free-text message. New behaviour: complete the name,
+  add a trailing space, leave caret after it, do NOT submit. A second Enter on
+  the blank tail does the attach-only switch (existing dispatch).
+
+| Gesture on an argument row | ENUM-tail (unchanged) | NAME+message (`/team`,`/agent`) — NEW |
+|---|---|---|
+| **Tab** | `_complete_argument`: fill name, no space, list stays (`_resolve_argument` 2362) | `_complete_name_argument`: fill name **+ trailing space**, caret at end, list closes (empty match set). No submit. |
+| **Enter, ambiguous** (fuzzy pick, >1 match) | `_complete_argument` (2363); 2nd Enter runs it | Same as Tab: fill name **+ space**, caret at end, no submit. (The name is now exact/one-match, but we still do not submit — for NAME+message "one match" is not "run", it is "ready for the message".) |
+| **Enter, unambiguous** (arrowed, typed-in-full, or single match on non-destructive list) | `_run_argument`: complete + `_submit` (2365) | `_complete_name_argument`: fill name **+ space**, caret at end. **No `_submit`.** |
+| **Click on row** (`_apply_command` arg branch 2339–2343) | non-destructive → `_run_argument` (run); `/logout` → `_complete_argument` (fill, wait) | `_complete_name_argument`: fill name **+ space**, caret at end. No submit. |
+| **Enter on blank tail** (`/team foo ˽`, caret after space, list closed) | n/a | Not a picker gesture — picker is closed, so the key handler's `if key == "enter": self._submit()` (1167–1171) fires. Dispatch collapses `arg` → attach-only. ✅ existing path. |
+| **Enter after typing a message** (`/team foo fix bug`) | n/a | Picker closed (zero matches), Enter → `_submit()` → `_cmd_team` sends the message. ✅ existing path. |
+
+Key point for the coder: the ONLY new insertion behaviour is "name **+ trailing
+space**, no submit" for the four NAME+message words. Everything after the space
+(blank-attach, message-send) is already handled by the tokenizer closing the
+list and the existing dispatch collapsing the arg. **No change to
+`_cmd_team`/`_cmd_agent`/`_run_slash_command`.**
+
+---
+
+## B. Declaration mechanism + exact methods to change
+
+### B.1 How the editor learns which commands are NAME+message
+
+Follow the established **registry-declares / editor-reads** idiom (same shape as
+`arguments` → `_argument_commands`). Two candidate mechanisms:
+
+**Option 1 — new `ArgumentMode` member `NAME`.** Add `NAME = "name"` to the
+enum (autocomplete.py 29–58), set `/team` and `/agent` to it.
+- Pro: one field already means "what is this argument"; NAME+message is exactly
+  a third answer to that question.
+- **Con (disqualifying): blast radius.** `arguments` is read in four places
+  beyond the editor: `mobile/daemon.py` 563–566 projects `cmd.arguments.name.
+  lower()` into the phone's slash catalogue (a NEW string `"name"` leaks onto
+  the mobile surface silently), and `set_commands` (896/902) plus `opens_a_list`
+  currently branch on `is not NONE` / `is REQUIRED`. A NAME member forces a
+  re-audit of every `arguments is/==` site to decide whether NAME counts as
+  "opens a list" (it does — it offers team/agent rows) and "can send when bare"
+  (it does — bare `/team` lists). That is real semantic surgery on a shared
+  enum for a two-command need. Also `on_argument_query_opened` already special-
+  cases the command word by name, so the enum member buys nothing there.
+
+**Option 2 — editor-local `NAME_ARGUMENT_COMMANDS`, derived from the registry,
+mirroring `MODEL_COMMANDS`. RECOMMENDED.**
+
+`MODEL_COMMANDS` (editor.py 566) is the exact precedent: a small class-level
+tuple of command words (with aliases spelled out) that the editor reads to give
+a subset of commands special completion behaviour, WITHOUT touching the shared
+registry enum. `/team` and `/agent` are the same kind of local exception.
+
+Concretely — two viable sub-forms; pick the first:
+
+- **B.2a (recommended): a class constant, aliases spelled out**, exactly like
+  `MODEL_COMMANDS`:
+  ```
+  #: Commands whose ARGUMENT is a NAME followed by a free-text message
+  #: (`/team <name> <request>`, `/agent <name> <message>`). Completing the
+  #: name adds a trailing space and leaves the caret after it so the user can
+  #: type the message; it does NOT submit. A blank Enter after the space is the
+  #: attach-only switch (handled by the app's dispatch, which strips the arg).
+  #: Both spellings, because the alias is itself a runnable command — same
+  #: reason MODEL_COMMANDS lists `models`.
+  NAME_ARGUMENT_COMMANDS = ("team", "teams", "agent", "agents")
+  ```
+  Place it next to `MODEL_COMMANDS`/`DESTRUCTIVE_COMMANDS` (566–572). This is
+  the least code, matches the nearest precedent exactly, and keeps the whole
+  change inside the editor.
+
+- **B.2b (alternative, only if a reviewer objects to a second hand-kept
+  tuple): a boolean field `name_argument: bool = field(default=False,
+  kw_only=True)` on `SlashCommand`**, and derive
+  `self._name_argument_commands` in `set_commands` (887) the same way
+  `_argument_commands` is derived (896). This keeps the fact on the registry
+  entry (closer to `echo`'s "policy next to the command it governs" argument)
+  at the cost of a new field. It does NOT touch `ArgumentMode`, so it avoids
+  Option 1's mobile/`opens_a_list` blast radius. Prefer B.2a for parity with
+  `MODEL_COMMANDS`; fall back to B.2b if the team's convention has shifted
+  toward registry fields since `MODEL_COMMANDS` was written. Either way the
+  editor reads a lowered-name membership test — the call sites in B.3 are
+  identical.
+
+Add a predicate mirroring `opens_a_list` (909):
+```
+def _is_name_argument_command(self, name: str) -> bool:
+    return name.lower() in self.NAME_ARGUMENT_COMMANDS   # or self._name_argument_commands
+```
+
+### B.3 Methods that change (all in editor.py)
+
+1. **New method `_complete_name_argument(self, name)`** — sibling of
+   `_complete_argument` (2367). Same tail-replacement math, but appends a
+   trailing space and moves caret to end:
+   ```
+   argument = slash_argument(self.text, self._argument_commands)
+   if argument is None:
+       return
+   self.text = f"{self.text[: len(self.text) - len(argument)]}{name} "
+   self.move_cursor(self._end_of_buffer())
+   ```
+   Docstring must state WHY it differs from `_complete_argument` (the space is
+   load-bearing: it terminates the name so the arg list closes, and it opens
+   the message tail the user now types — the inverse of `_complete_argument`'s
+   "no space so the matcher keeps matching"). Setting `self.text` funnels
+   through `load_text` → `_sync_picker` (2158–2159), which re-derives the picker
+   as ARGUMENT-mode-with-empty-matches (verified), so the list closes on the
+   same keystroke.
+
+2. **`_resolve_argument` (2353)** — branch on the command class up front:
+   ```
+   if self._is_name_argument_command(self._argument_command):
+       self._complete_name_argument(name)
+       return
+   # existing behaviour for enum-tail commands:
+   if key == "tab" or not unambiguous:
+       self._complete_argument(name)
+       return
+   self._run_argument(name)
+   ```
+   This makes Tab AND Enter (ambiguous or not) all route to
+   `_complete_name_argument` for team/agent, which is exactly the spec table:
+   for a NAME+message command neither key ever submits.
+
+3. **`_apply_command` (2317)**, argument branch (2339–2344) — add the same
+   guard before the destructive/run split:
+   ```
+   if self._picker.mode is PickerMode.ARGUMENT:
+       if self._is_name_argument_command(self._argument_command):
+           self._complete_name_argument(name)
+           return
+       if self._argument_is_destructive():
+           self._complete_argument(name)
+           return
+       self._run_argument(name)
+       return
+   ```
+   Covers the mouse-click path.
+
+4. **No change** to `_run_argument`, `_complete_argument`, the key handler
+   (1088–1147 — it already delegates to `_resolve_argument`/`_apply_command`),
+   `_run_slash_command`, `_cmd_team`, `_cmd_agent`, or the mobile projection.
+
+`_argument_command` is `Optional[str]`; the predicate must handle `None`
+(`(name or "").lower()`), because `_apply_command`'s argument branch reads it
+while a list is open — it is set, but be defensive.
+
+---
+
+## C. Highlighting design (ask #2)
+
+### C.1 Render seam
+
+Reuse the **exact** `_paint_markers` seam. `render_line` (748) already does
+`return self._paint_markers(super().render_line(y), y)`. Add a second
+post-process pass for slash highlighting. Two integration choices:
+
+- **C.1a (recommended): a sibling method `_paint_slash(strip, y)`, chained
+  inside `render_line`** so the finished strip is
+  `self._paint_slash(self._paint_markers(super().render_line(y), y), y)`.
+  Ordering: markers first, slash second, OR slash first — they never overlap
+  (a slash line has no `[Image …]` marker: the command word is the first token
+  and the marker grammar opens with `[`), so order is immaterial for
+  correctness; put slash LAST so its bail is the cheap one on prose lines.
+- C.1b: fold into `_paint_markers`. Rejected — `_paint_markers` is documented
+  as "repaint the MARKER cells", its `cells`/`edges`/`styles` machinery is
+  marker-specific, and mixing a second token grammar into it muddies a method
+  whose comments carefully explain the chip rules. A named sibling keeps each
+  pass single-purpose, matching the file's style.
+
+### C.2 What gets highlighted, and what "recognized" means
+
+Compute on **document line 0 only** (the command lives on the first non-blank
+line; multi-line = message body, no command). Highlight at most two token runs:
+
+1. **The leading `/command` token** — from the `/` through the end of the first
+   whitespace-delimited word. "Recognized" ⇔ `slash_command_for(line)` returns
+   non-None (the same resolver dispatch uses, so the highlight cannot claim a
+   command the app would reject). This covers the WHOLE slash surface (ask #2's
+   "not just team/agent").
+   - Recognized → **command style** (`text-area--slash-command`).
+   - Not recognized (`/xyz`, a typo) → **unrecognized style**
+     (`text-area--slash-unknown`), a muted/neutral treatment so the user sees
+     "this leading slash is not a command and WILL be sent as text". See D for
+     why we highlight the unknown case at all.
+
+2. **The argument NAME token** — only when the command is one whose argument is
+   a name we can validate cheaply, i.e. a NAME+message command
+   (`_is_name_argument_command`) AND the typed name matches a known team/agent.
+   - Matched name → **argument-name style** (`text-area--slash-argument`).
+   - Partial/typo name → NO highlight (leave prose colour). Do not paint an
+     "unknown name" — the picker already shows/【doesn't show】 a match, and a
+     half-typed name is a normal in-progress state, not an error.
+
+   Only the NAME (first token after the command word) is ever highlighted;
+   the free-text message tail is prose and must stay prose — that is the whole
+   point of the feature (show the user what is command vs. message).
+
+### C.3 Cheap "recognized" on the hot path
+
+`render_line` runs every keystroke-frame (the marker code calls this out at
+1595). Keep the pass O(1)-ish:
+
+- **Bail exactly like the marker pass.** First cell of the method:
+  ```
+  if y != 0 and self.scroll_offset.y == 0:  # only the first screen row can hold the command
+      ... actually: resolve the DOCUMENT line for screen row y, bail if it isn't line 0
+  line = self.document.get_line(0)
+  if not line.lstrip().startswith("/"):
+      return strip
+  ```
+  A composer whose first non-blank line does not start with `/` (the
+  overwhelming common case — prose, empty, bang-mode `!`) returns the strip
+  untouched before any tokenizing. This mirrors `_marker_cells`'s `if "[" not
+  in line` guard (1595).
+
+- **Command recognition is a dict/loop lookup**, not a regex sweep:
+  `slash_command_for(line)` splits the first token and does a `next(... in
+  entry.names ...)` over ≤ ~30 commands (526–543). Cheap. Cache is unnecessary
+  at this size; if a reviewer wants it, memoize on `(line0, commands_version)`
+  — but measure first, do not add cache complexity speculatively.
+
+- **Name recognition needs a set of known team/agent names.** DO NOT call
+  `_team_choices()`/`_agent_choices()` from the render path — those hit the
+  registry (`list_teams`, `_agent_profile_rows`) and are app-side (app.py 2773,
+  2967), not reachable cheaply from the widget every frame. Instead, **push the
+  name set into the editor when the argument list opens.** The app already
+  computes the rows in `on_argument_query_opened` (9530–9536); have it also call
+  a new `editor.set_argument_names(names: frozenset[str])` (or fold the names
+  into the existing `picker.set_choices` path and read them back off the picker
+  — the picker already holds `_choices`, and `_apply_command`-time code reads
+  `self._picker.suggestions()`). Simplest and lowest-risk:
+  - Add `editor.set_name_choices(frozenset(c.name for c in choices))` called
+    right beside `picker.set_choices(self._team_choices())` /
+    `_agent_choices()` (9531/9535). Store on `self._name_choices: frozenset[str]
+    = frozenset()`; clear it whenever the argument command changes to a
+    non-name command (in `_sync_picker`, alongside the existing
+    `set_choices([])` at 2186, when leaving a name command). The render pass
+    then does `name_token.lower() in self._name_choices` — a frozenset hit, O(1).
+  - Rationale: the render path must not do I/O or registry walks; the app owns
+    the truth and already visits it when the list opens, so it hands the widget
+    a cheap immutable snapshot. This is the same "app fills the rows when the
+    list opens" contract `on_argument_query_opened` already documents (9481–9493).
+
+  Caveat: the name set is only populated once the argument list has opened at
+  least once (the space after `/team`). If the user types `/team frontend `
+  fully by hand without the list ever having filled, `_name_choices` may be
+  empty and the name goes un-highlighted. Acceptable: highlighting is an
+  affordance, not a gate, and the list fills on the space in practice. Document
+  this as a known limitation rather than fetching on the render path.
+
+### C.4 Painting the cells
+
+Model the pass on `_paint_markers` (1664) exactly:
+- Resolve document line 0's screen-row extent (reuse the `wrapped_document` /
+  `offset_to_location` machinery — a long `/team … message` soft-wraps; the
+  command word is on the first wrapped row, so in practice only `y==0`'s section
+  carries the command token, but the name token can wrap. Keep it correct by
+  mapping document columns → screen x via `wrapped.location_to_offset`, same as
+  the marker code at 1637/1646).
+- Build `(x_start, x_end, kind)` runs for the command token and (if matched) the
+  name token, add `gutter`.
+- `edges = sorted({0, width} | {x for …})`, `strip.divide`, and for each piece
+  overlay the component style via `_overlay(piece, style)` (1695–1697,
+  `post_style`), because — same reason as the chip — TextArea's segments carry
+  explicit fg/bg and a base style would be discarded.
+- Styles resolved via `self.get_component_rich_style("text-area--slash-*")`
+  (like 1682–1683).
+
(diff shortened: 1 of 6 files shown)


damianvtran:
### Design review — round 1

Designer: designer (Claude Sonnet 4.5, lopdev team)
Scope: origin/main..dev-composer-ux (HEAD 1cc3832b)

Reviewed the rendered after-frames (`view-{1..6}-after.png`) against the before
SVGs (`{2,3,4}-*-before.svg`, rendered), and cross-checked the tint hues in
`local_operator.tui.theme` (dark palette) against the composer ground
(`surface #1e1a14`, `bg #14110c`).

**What works (no findings):**
- Reflow: none. The command word / name tint is foreground-only — gutter `›`,
  text start column, wrap point, prompt-row height, and status band are pixel-
  identical before→after (frames 3, 4). No picker-gutter regression (frame 1);
  the argument picker rows and `tint-select` ground are unchanged.
- Legibility on dark: every tier clears AA against the input panel —
  command `signal #6ea8d8` 6.81:1, name `string #57c785` 8.17:1, unknown
  `muted #b5afa2` 7.93:1, prose `fg #e9e5db` 13.76:1.
- Three-tier read (frame 3, the money shot): blue-bold `/team` vs green
  `frontend-guild` vs white prose are cleanly separable — command-vs-name
  luminance ratio 1.20:1 is helped by the hue split (cool vs warm) and the bold
  weight on the command word, so they never collapse into one read. Mirrors
  correctly for `/agent` (frame 6) and for a non-name command `/compact`
  (frame 5).
- Autofill resting state (frame 2): `/team frontend-guild ␣` with the block
  caret parked one cell past the trailing space, list closed. The green name
  tint stops at the name — it does not bleed onto the space or the caret block.
  Reads exactly as "name filled, waiting for your message," which is the goal.
- The tint reads as "recognized structure that won't be sent": the bold cool
  command word + green name visually separate from the plain-white prose tail
  that will be sent. Intent lands.

**Findings:**

**D1 — MINOR — unknown-command muting is the weakest of the four reads.**
`slash-unknown` uses `muted #b5afa2` (frame 4, `/notacommand`). Against the
prose tail (`fg #e9e5db`) that is only a 1.74:1 luminance separation — the
faintest distinction in the whole scheme and the one the eye must catch to
understand "this leading token is NOT a command, it will be sent as text." In
the after-frame `/notacommand` reads as *slightly* dimmer than "hello there"
rather than clearly inert; a user skimming may miss that it degraded to plain
text. The rationale (muted, not danger, because it is a typo-in-progress) is
sound and I would NOT push to a danger hue — but `muted` sits *above* `dim
#837c6d` in luminance, so a half-step down to `dim` (4.18:1 vs ground, still
comfortably legible; ~2.9:1 below prose) would make "inert / will-be-sent" land
without any alarm. Non-blocking; worth a look or a follow-up.

**D2 — MINOR — argument-name green is a near neighbour of the live-accent green.**
The name uses `string #57c785`; the app reserves `accent #38c96a` for "a turn
is live / focus." Those are close greens, and the composer is exactly where a
"live" cue would appear. In frame 5 the picker's selected row also renders the
choice in green on `tint-select`, so a green token in the composer region is
doing double duty. It is distinguishable here (different weight/context, and
the command word anchors the read), and the design doc's reasoning for avoiding
`accent` is deliberate — but flag it so a future "live composer" affordance
doesn't collide. Non-blocking.

**D3 — NIT — two blues stack in the composer region.**
The command word (`signal #6ea8d8`) is the same hue as the cwd path in the
status line one row below (`/private/tmp/lop-composer-ux`, also `signal`). Frame
3/4/6 show blue `/team`|`/agent` directly above the blue path. This is
intentional reuse ("structured token," already the chip ink) and they are in
separate bands, so it is purely cosmetic — noting for completeness.

**D4 — NIT — bold weight is carried only by the command word.**
The name is tinted but not weighted, so at a glance the command word is the
loudest element and the recognized *name* is the second read. That ordering is
correct (command first, argument second), so this is a confirmation, not a
defect — recorded only so a later "should the name also bold" question has an
answer on file: no, keep the single bold anchor.

**Verdict: TERMINAL** — no BLOCKER, no MAJOR. Legibility, tier separation,
caret/spacing, and no-reflow all pass on the current head. D1 and D2 are MINOR
follow-ups (weak unknown/prose separation; green-name vs live-accent adjacency);
D3/D4 are NITs. Nothing here blocks merge from a design standpoint.

damianvtran:
### UX review — round 1

Reviewer: ux-reviewer (Claude Sonnet 4.5) — lopdev team, UX role
Scope: origin/main..dev-composer-ux (head 1cc3832b)

Walked the real interaction flow end-to-end through the shipped `PickerHarnessApp` pilot (the harness that mounts the real `Editor` + `CommandPicker` + `local_operator.tcss` and mirrors the app's `on_argument_query_opened` seam), driven with `env -u NO_COLOR TERM=xterm-256color PYTHONPATH=/tmp/lop-composer-ux .venv/bin/python`. Not a stills review — I exercised keyboard, mouse, timing, and picker state on the live event loop. Rendered after-frames (`view-{1..6}-after.png`) cross-checked for the visual half.

**What I confirmed works (the ask, end to end):**

- `/team ` + arrow + **Enter** → autofills `/team release-crew ` (name **+ trailing space**), caret parked at end (col 19), picker **closed**, **no submission**. ✅
- Blank **Enter** on that tail → submits `'/team release-crew '` (attach-only); dispatch collapses the bare name and the `_cmd_team` "…is ready. <manager> leads it. Send a request with /team <name> <message>." notice fires. ✅
- `/team ` + **Enter** (arrowless/ambiguous) → fills `frontend-guild ` + space, no submit; typing `fix the flaky test` then **Enter** submits the whole line `'/team frontend-guild fix the flaky test'`. ✅
- **Tab** and **mouse-click** on a team row both do the same fill-name+space-no-submit; a click never discards a half-typed message. ✅
- `/agent ` mirrors `/team` exactly (fill `dashboard-sme `, blank Enter attaches). ✅
- **Regression — enum-tail unchanged:** `/effort ` + down + Enter → runs immediately (`/effort alibaba`); `/login ` + down + Enter → runs; `/logout` (destructive) still fills-then-waits, second Enter runs. ✅
- **No accidental double-submit:** fast Enter-Enter on `/team ` = fill, then attach-only — never a stray send. ✅
- Highlight holds up at width 40 and 24 with the name+message soft-wrapping across 3–5 rows: `render_line` over every visible row is crash-free and submission is intact. ✅

No BLOCKER and no MAJOR. The four findings below are friction/discoverability follow-ups.

---

**U1 — MINOR — No signal at the moment of autofill that a blank Enter *switches* while a typed message *sends*.**
Step: `/team ` → arrow → Enter. The name fills, a space appears, the caret parks — and the picker closes with an **empty notice** (confirmed `picker._notice == ''`). Nothing on screen tells the user that Enter-now attaches-only versus type-then-Enter sends. The genuinely helpful "…is ready. Send a request with /team <name> <message>." copy only appears *after* the user commits the blank Enter — i.e. after they've already taken the leap. A first-timer selecting a team and pressing Enter (reasonably expecting "go") gets a silent context switch and has to infer, from a transcript notice, what just happened. Expected: a one-line affordance at the parked-caret moment (a picker/composer hint such as "Enter to switch · type a message to send") so the two outcomes are discoverable before acting, not after.

**U2 — MINOR — The same gesture now means two different things depending on the command class, with nothing to distinguish them.**
Step: compare `/effort ` + row + Enter (runs immediately) against `/team ` + row + Enter (parks). Enter-on-a-highlighted-argument-row submits for the enum-tail commands and parks for `/team`·`/agent`. The divergence is the whole point of the change and is correct, but it's invisible: a user who has internalised "arrow to a row, Enter runs it" from `/effort`/`/login`/`/theme` will read the `/team` park as a dropped keystroke. This is the flip side of U1 — the fix is the same discoverability hint naming the outcome for name+message rows. Recording separately because it's a *consistency* expectation, not just a missing label.

**U3 — NIT — Tab after the autofill inserts a literal indent, leaving stray whitespace between name and message.**
Step: `/team ` → Enter (fills `/team release-crew `, picker closed) → **Tab**. Because the picker is closed, `tab_behavior="indent"` takes the key and the buffer becomes `/team release-crew  ` (caret 19→20). Harmless to dispatch — `partition(" ")` + `.strip()` discards it — but a user with a Tab-to-accept habit (Tab *was* the accept key one keystroke earlier) gets silent cruft in the composed line, and if they Tab mid-message they inject an indent into the text that *will* be sent. Expected: Tab immediately after a name completion is a no-op or is otherwise not a text-mutating surprise.

**U4 — NIT — Backspacing the trailing space re-pops the argument picker right after completion.**
Step: `/team ` → Enter (fills `/team release-crew `) → **Backspace**. The buffer becomes `/team release-crew` and the picker **reopens** in ARGUMENT mode. Defensible as "you're re-editing the name," but immediately after a completion it reads as the list resurrecting itself. Low severity and arguably correct; flagging so the behaviour is a deliberate choice on record, not an accident.

---

**Verdict: TERMINAL for round 1** — no BLOCKER, no MAJOR. The core interaction (autofill-not-submit, blank-Enter-attaches, message-Enter-sends, enum-tail unchanged, no double-submit, mouse/Tab/keyboard parity, wrap-safe highlight) is correct and safe on the live flow. U1–U4 are recorded as follow-ups; U1/U2 (a single discoverability hint) are the ones worth picking up next.

damianvtran:
### Agent review — round 1

Reviewer: reviewer subagent (Claude / claude-opus-4) — independent of the commit author; we share one GitHub account, so this line, not the commit authorship, marks the round independent.
Scope: origin/main..dev-composer-ux (HEAD 1cc3832b)

Reviewed the full diff (editor.py, app.py, local_operator.tcss, tests), the design doc, and traced the live dispatch (`_run_slash_command` 7324–7327 → `_cmd_team` 2859 / `_cmd_agent`) rather than trusting the description.

**Gates (run from the worktree, editable venv, `PYTHONPATH=/tmp/lop-composer-ux`):**
- `flake8` (editor.py, app.py, tests) — **clean**
- `black --check` (black==26.1.0, `local_operator tests`) — **clean** (423 files unchanged; the .tcss is not a black target)
- `isort --check-only --profile black` — **clean**
- `pyright --pythonpath .venv/bin/python` (editor.py, app.py) — **0 errors, 0 warnings**
- `pytest tests/unit/tui/test_command_picker.py` (`env -u NO_COLOR TERM=xterm-256color`) — **231 passed**

**What I verified against the asks:**
- (a) Autofill routing is complete and correct: both the click path (`_apply_command`, 2578) and the Tab/Enter path (`_resolve_argument`, 2608) short-circuit to `_complete_name_argument` for the four NAME+message words before any `_run_argument`/`_submit` can fire; the arrow-onto-row and ambiguous cases are covered too. Enum-tail commands (`/login`) are untouched — regression guarded by `test_enum_tail_login_row_still_runs_on_enter`. Blank-Enter-attaches and typed-message-sends genuinely reach `_submit()` via the closed-picker path (1256) and collapse to attach-only / send in `_cmd_team` (2884 vs 2890); tests assert `submissions == []` on completion and the exact submitted line on the two send paths.
- (b) Render seam is cheap and I/O-free: `_slash_runs` reads only `_command_names`/`_name_choices` (frozensets) and the pure `slash_argument`; no app import, no registry walk. The leading-`/` bail is real — prose returns `None` before any span work. `_paint_slash` composes after `_paint_markers` with no cell contention (command/name on line-0 `/`, markers in the `[`-tail).
- (c) Staleness handled: snapshot cleared when the list closes (2403) and on command change (2419); `set_name_choices` refreshes only on real change.
- (d) Marker overlay coexistence confirmed — separate cell ranges.
- (e) tcss uses `$lo-signal`/`$lo-string`/`$lo-muted`, explicitly not `$lo-accent`; all three keys exist in theme.py; COMPONENT_CLASSES consistent with the chip convention.
- (f) Tests assert non-submission and cover the enum-tail regression.

**Findings**

- **CR1 — minor.** `editor.py:1806` — `_slash_runs()` re-parses the entire buffer (`self.text.split("\n")` + first-non-blank scan) once per **visible screen row**, because `render_line(y)` → `_paint_slash` → `_slash_cells(y)` → `_slash_runs()` runs per-row, and the runs are identical for every row of a frame. On the prose hot path it still allocates a line list per row before the leading-`/` bail. Composer buffers are small so the cost is negligible in practice, but it is trivially hoistable (compute the runs once per frame / memoize on `(text, generation)`), and the design doc's "cheap on the hot path" claim is a hair softer than stated. Not stop-ship.

- **CR2 — minor.** Test gap: the most intricate new code is the soft-wrap boundary math in `_slash_cells` (1877–1895), and no test exercises a `/team <long-name> <message>` that wraps the NAME token across screen rows — the `wraps_on`/`section_end` branch is uncovered. The straight-line cases are well covered; a single wrapped-name assertion would lock the branch that's easiest to get wrong.

- **CR3 — nit.** `_slash_runs` docstring (1806) does `.split("\n",1)[0]` on the name after the single-content-line guard (1826) has already excluded any newline in the buffer — the second split is dead-but-harmless. Drop it or note it as belt-and-suspenders.

No blocker or major findings. The autofill logic, the render seam, the staleness handling, the theme-variable choices, and the test coverage of non-submission + the enum-tail regression are all correct.

**Verdict: clean.** This round is TERMINAL — CR1/CR2/CR3 are minor/nit follow-ups, not a reason for another required round.

damianvtran:

Implementer: coder subagent (lopdev team). Responding to the code-review round (`### Agent review — round 1`).
Scope of remediation: `origin/main..dev-composer-ux` (HEAD `be32e944`).

All three findings were minor/nit and TERMINAL; all three are **fixed** in `be32e944`.

- **CR1 — minor — FIXED (`be32e944`).** `_slash_runs()` was re-parsing the whole buffer once per **visible screen row** because `render_line(y)` → `_paint_slash` → `_slash_cells(y)` → `_slash_runs()` runs per row and the runs are identical across a frame. Split it into a memoized `_slash_runs` wrapper over a new `_compute_slash_runs` body. The cache key captures **every input the parse reads** — buffer text, `_command_names`, `_name_choices`, `_argument_commands`, and the picker `mode`/`is_open()`/`is_pending()` triple that gates the unknown treatment — so a keystroke that changes any of them recomputes on the next row that asks, while the remaining rows of the same frame reuse the result. Composer buffers are small so this is a micro-optimisation, but it makes the design doc's "cheap on the hot path" claim exact and removes the per-row line-list allocation on prose frames. Verified: repeated `_slash_runs()` calls hit the cache (stable key), and mutating the buffer or the name snapshot recomputes.

- **CR2 — minor — FIXED (`be32e944`).** Added `test_a_name_token_that_soft_wraps_is_highlighted_across_both_rows`. It drives a 29-char name (`frontend-platform-guild-alpha`) into a width-24 composer so the **NAME token itself** straddles a wrap offset (`content width 18`, offsets `[6, 23, 39]`) — precisely the `wraps_on`/`section_end` branch of `_slash_cells` the straight single-row tests never reached. It gathers every green (`slash-argument`) run across all wrapped rows and asserts the painted text equals the full name exactly (contiguous across the break, and the message tail `go now` never picks up the colour).

- **CR3 — nit — FIXED (`be32e944`).** Dropped the dead `.split("\n", 1)[0]` on the name token. The single-content-line guard (`if len(lines) > first + 1: return None`) already excludes any newline in the buffer, so the second split could never fire; removed it and noted the guard in the comment.

**Gates re-run from the worktree (`PYTHONPATH=/tmp/lop-composer-ux .venv/bin/python`, TUI env `env -u NO_COLOR TERM=xterm-256color`):**
- `flake8 .` — clean
- `black --check` (black==26.1.0, whole tree) — clean (423 files unchanged)
- `isort --check-only --profile black local_operator tests` — clean
- `pyright --pythonpath .venv/bin/python .` — 0 errors, 0 warnings
- `pytest tests/unit/tui/test_command_picker.py` — 236 passed (incl. the new wrap test); `test_slash_echo.py`, `test_app_pilot.py`, `test_marker_chip.py` all green.

damianvtran:

Implementer: coder subagent (lopdev team). Responding to the design round (`### Design review — round 1`).
Scope of remediation: `origin/main..dev-composer-ux` (HEAD `be32e944`).

- **D1 — minor — FIXED (`be32e944`).** The unknown-command mute used `$lo-muted #b5afa2`, only ~1.74:1 from the prose tail (`fg #e9e5db`) — too faint to read as "this leading token is NOT a command and will be sent as text." Darkened it a half-step to `$lo-dim #837c6d`, exactly the value the round suggested (~2.9:1 below prose, ~4.18:1 above the input ground) — the "inert / will-be-sent" read now lands without tipping into an alarm colour. Updated the tcss comment with the ratio rationale and the failing test assertion (`muted`→`dim`).
  Re-captured the unknown-command after-frame from the real `OperatorApp` (`4-unknown-command-after.svg`) and **looked at it**: `/notacommand` now reads clearly dimmer/inert against the brighter `hello there` prose, where before it was near-indistinguishable. It does not read as an error — the degrade-to-text intent is unmistakable. Screenshot: `/tmp/composer-shots/view-4-after-d1.png`.

- **D2 — minor — ACKNOWLEDGED / not changed (deliberate).** The argument-name green (`string #57c785`) is a near neighbour of the reserved live-accent green (`accent #38c96a`). Not changed: the design doc's reasoning for avoiding `accent` (reserved for "a turn is live") is exactly why `string` was chosen, the two are distinguishable in context (weight/anchor/position, and the command word anchors the read), and the round rated it non-blocking and TERMINAL. Recorded as a **future-collision note** for whoever adds a "live composer" affordance — the finding is on file as a constraint, not a defect to fix now.

- **D3 — nit — ACKNOWLEDGED, no change.** Two blues stack (command word `signal` above the cwd path `signal` one band down). Intentional reuse of the "structured token" hue (already the chip ink); they sit in separate bands and the round called it purely cosmetic. No change.

- **D4 — nit — ACKNOWLEDGED, no change (confirmation).** Bold weight is carried only by the command word, making it the loudest element and the name the second read — which is the correct ordering (command first, argument second). The round recorded this as a confirmation, not a defect: keep the single bold anchor. No change.

Design round was TERMINAL with no blocker/major. D1 (the one actionable item) is fixed and visually confirmed on the current head; D2–D4 are acknowledged as deliberate/notes.

damianvtran:

Implementer: coder subagent (lopdev team). Responding to the UX round (`### UX review — round 1`).
Scope of remediation: `origin/main..dev-composer-ux` (HEAD `be32e944`).

- **U1 + U2 — minor — FIXED with one change (`be32e944`).** Both findings are the same gap: at the autofill moment nothing on screen told the user that a blank Enter **switches** (attach-only) while a typed message **sends**, and that this diverges from the enum-tail commands where Enter-on-a-row runs immediately. Added a bounded, self-clearing hint — **"Enter to switch · type a message to send"** — shown in the picker's **own notice row** (`CommandPicker.set_notice`, the codebase's existing "say it where the list was" idiom: self-clearing, unrepeatable, costs the transcript nothing — no new widget). It appears **only** in the parked-caret state (`_name_switch_hint`: a NAME+message command whose argument is a completed name + one terminating space + empty tail) and is withdrawn the instant a message character is typed (the tail stops being empty) or the name is edited (the space is gone). Gated to NAME+message commands so the enum-tail argument lists never surface it.
  Verified on the **real app** (`OperatorApp` + real registry): `/team frontend-guild` → Enter → buffer `/team frontend-guild ˽`, picker notice = the hint, `display=True`; typing `f` clears it; `/login anthropic ` (enum-tail) shows nothing; bare `/team ` (rows up) shows nothing. Re-captured the parked-state after-frame and **looked at it** — the hint sits one row under the composed line in dim ink, directly in the user's eye-line, naming both outcomes before they commit. Screenshot: `/tmp/composer-shots/view-2-after-hint.png`. Tests added: `test_autofill_shows_the_switch_or_send_hint`, `test_the_hint_is_withdrawn_once_a_message_is_typed`, `test_a_bare_name_list_shows_no_hint`, `test_enum_tail_completion_shows_no_switch_hint`.

- **U3 — nit — REJECTED (accepted behaviour).** Tab after the autofill inserts an indent because the picker is closed and `tab_behavior="indent"` takes the key. Dispatch discards it (`partition(" ")` + `.strip()`), and suppressing Tab specifically in the one-keystroke window after a completion would be a special-case with its own surprises (Tab is the composer's normal indent key everywhere else, and "Tab does nothing sometimes, right after a completion" is a harder rule to hold than "Tab indents"). The round itself rated it a NIT and "harmless to dispatch." Left as-is deliberately.

- **U4 — nit — REJECTED (correct behaviour).** Backspacing the trailing space reopens the argument picker in ARGUMENT mode. This is correct: removing the space puts the buffer back into `/team frontend-guild` — the name is being re-edited, so the list that helps edit it should be available (the same `slash_argument`-drives-the-parse rule that governs every other argument list). The round flagged it "arguably correct… so the behaviour is a deliberate choice on record" — confirming it as deliberate. No change.

UX round was TERMINAL with no blocker/major. U1/U2 (the discoverability hint the round called "the ones worth picking up next") are fixed and confirmed on the live flow; U3/U4 are rejected with reasons above.

damianvtran:
### Design review — round 2

Designer: designer (Claude Sonnet 4.5) — naming myself for independence, since we share one account.
Scope: 1cc3832b..be32e944 (the round-1 remediation delta; the full change is origin/main..be32e944)

Focused DELTA confirmation on the fresh head `be32e944`. Round-1 design was TERMINAL; the head moved and added two user-visible changes (D1 colour darkening + a new picker-notice hint), so this round re-verifies only those against re-captured frames and checks for reflow/alignment regression vs the round-1 stills. Rendered from the real `OperatorApp` via `save_screenshot`.

**Artifacts viewed**
- `/tmp/composer-shots/view-4-after-d1.png` — `/notacommand hello there`, unknown-command highlight at `$lo-dim`.
- `/tmp/composer-shots/view-2-after-hint.png` — `/team frontend-guild` autofilled, empty tail, new hint row shown.
- `/tmp/composer-shots/view-2-after.png` — round-1 name-selected frame (no hint), used as the reflow baseline.

**D1 — unknown-command colour `$lo-muted` → `$lo-dim` (view-4-after-d1)**
Confirmed. `/notacommand` now renders in the dim gutter-grey while the `hello there` tail stays bright prose-white. The two-tone read is unmistakable: the slash token visibly degrades to "inert text that will be sent," not a recognized command, and the separation from the prose tail is now clearly legible rather than the ~1.74:1 "slightly dimmer" it was at `muted`. Crucially it does NOT read as an error/alarm — it is a quiet degrade, not a danger colour. (The red `!` on the status band is the unrelated connection indicator, not part of this change.) Recognized commands (`/team` bright cyan+bold in view-2), names (green), and prose (white) remain cleanly separated across the three tiers. No change to the recognized-command styling.

**New hint row "Enter to switch · type a message to send" (view-2-after-hint)**
Confirmed and clean. The hint sits in the picker's existing notice channel directly under the composer input, in the same dim style as the `· /resume picks up…` notice above it — consistent with the picker's established notice styling, not a new invented widget. It is legible against the composer's raised background, reads unambiguously as guidance (verb-led affordance copy), and carries no error/alarm colour. Copy is accurate: Enter switches, typing sends. The middot separator matches the existing notice-row convention.

**Reflow / alignment — no regression**
Diffing view-2-after (no hint) against view-2-after-hint: the `❯ /team frontend-guild` composer row and the `◆ test/model › ⌂ …` status band occupy identical positions in both frames. The hint occupies one extra row inside the notice channel; it does not push the input row, shift the caret, or misalign the status band. The hint's left margin aligns with the composer text column. Bounded to the empty-tail state as designed (view-4's unknown-command state shows no hint), so it cannot cover a real picker row.

**Findings**
None. No BLOCKER, no MAJOR, no MINOR, no NIT on the delta.

**Follow-ups (carried from round 1, no change expected this PR)**
- D2/D3/D4 remain acknowledged no-change per round-1 remediation.

**Verdict: TERMINAL on head `be32e944`.** Both remediation changes are correct, consistent with existing composer styling, and introduce no reflow or alignment regression.

damianvtran:
### UX review — round 2

Reviewer: ux-reviewer (Claude Sonnet 4.5) — lopdev team, UX role
Scope: 1cc3832b..be32e944 (the round-1 remediation delta; full change is origin/main..be32e944)

A focused **delta** re-review, not a full re-walk. Round-1's UX round was TERMINAL (no blocker/major); it went stale because the head moved to `be32e944`, which added a new user-visible affordance — the picker-notice hint **"Enter to switch · type a message to send"** at the autofill-with-empty-tail moment (fixes U1/U2) — plus the D1 colour change. This round confirms the remediation is correct and introduced no interaction regression on the fresh head.

Walked the **real flow** on the live event loop through the shipped `PickerHarnessApp` pilot (real `Editor` + `CommandPicker` + `local_operator.tcss` + the real `SLASH_COMMANDS` registry and the app's `on_argument_query_opened` seam), driven with `env -u NO_COLOR TERM=xterm-256color PYTHONPATH=/tmp/lop-composer-ux .venv/bin/python`. Keyboard, mouse, timing and picker-notice state were exercised as keystrokes — not read from source. Rested-state after-frame (`view-2-after-hint.png`) rendered and looked at for the visual half.

**What I confirmed on head `be32e944` (23/23 live checks green):**

- **The hint appears at exactly the right moment.** `/team frontend-guild` → **Enter** → buffer parks at `'/team frontend-guild '` (name + one trailing space), caret at end (col 21 == end-of-buffer), picker rows closed, **no submission** — and the picker's own notice row now reads the hint (`_notice == "Enter to switch · type a message to send"`, `display=True`). Same for **Tab** (`/team release-crew ` → hint) and for a **mouse-click** on a team row (`/team ` → click row → `/team frontend-guild `, hint, no submit). The rendered frame shows it one row under the composed line in dim ink, in the same weight as the `/resume` tip directly above it — in the user's eye-line, naming both outcomes *before* they commit. This is precisely the discoverability gap U1/U2 flagged, closed.
- **It self-clears on the first message keystroke.** After autofill, pressing `f` → buffer `'/team frontend-guild f'`, `_notice == ""`; the notice stays cleared through the whole typed message (`fix the flaky test`) and never sits over live text. The two outcomes still resolve correctly: **blank Enter switches** (submits `'/team frontend-guild '`, attach-only) and a **typed message sends** (submits `'/team frontend-guild fix the flaky test'`). `/agent dashboard-sme` mirrors this exactly.
- **It never intrudes where it shouldn't.** Bare `/team ` (rows still up, name not chosen) → no hint, so it can't compete with the row list. Enum-tail `/login anthropic ` → no hint, and the enum-tail Enter-on-row still runs immediately (`/login alibaba`), unchanged. Plain prose typing and a partial `/team front` name → no hint. The gate (`_is_name_argument_command` + non-empty first token + empty tail past the first space) is correct.
- **No new trap or double-submit.** Fast Enter-Enter after autofill = one fill then one attach-only submit, never a stray send. **Esc** on the parked state does not wedge the input — buffer intact and editable, notice withdrawn, `display=False`; no dead end. The hint is display-only in an idiom the product already uses (`CommandPicker.set_notice`), so it adds no focus target and no new key to escape.
- **Resize / narrow degradation.** The hint clips to the notice-row width at 40/24/20 columns with `render_line` crash-free across every visible row — it degrades exactly as the picker's existing notice does, no overflow, no reflow of the docked input.

**U3 / U4 rejections — I agree, on the walked evidence:**

- **U3 (Tab-after-autofill inserts an indent) — reasonably rejected.** Walked it: with the picker closed, Tab inserts spaces (`'/team frontend-guild '` → `'/team frontend-guild    '`) because `tab_behavior="indent"` owns the key everywhere in the composer. Dispatch strips the whitespace tail (`partition(" ")` + `.strip()` → empty), so blank-Enter still resolves to **attach-only** — the trailing indent is inert. Making Tab a no-op only in the one-keystroke window after a completion would be a harder rule for a user to hold ("Tab indents, except right after a completion") than the invariant "Tab indents." NIT, harmless, correctly left as-is.
- **U4 (backspacing the trailing space re-pops the argument list) — correctly rejected.** Walked it: removing the space returns the buffer to `/team frontend-guild`, which is the name being re-edited, so the argument list that helps edit the name reopens in `ARGUMENT` mode and the hint withdraws (`_notice == ""`). This is the same `slash_argument`-drives-the-parse rule that governs every argument list; suppressing it would make the name uniquely un-re-editable. Deliberate and consistent.

**Findings this round:** none — no BLOCKER, no MAJOR, no new MINOR/NIT. The remediation adds the affordance the round asked for, in the product's existing idiom, without touching the confirmed-good completion/submission/regression behaviour.

**Verdict: TERMINAL on head `be32e944`.** The U1/U2 fix is live and correct on the real flow; U3/U4 stand rejected on walked evidence. No open UX follow-ups. (Reviewer named for independence — the account is shared; this is a different agent than the implementer.)

damianvtran:
### Agent review — round 2

Reviewer: reviewer subagent (Claude Opus 4.6, lopdev team) — INDEPENDENT of the implementer; we share one account, so this line names me for independence, not the commit author.
Scope: `1cc3832b..be32e944` (the round-1 remediation delta). Full change is `origin/main..be32e944`.

Focused DELTA confirmation of the round-1 remediation on the fresh head `be32e944`, not a full re-review. Round-1 code/design/UX rounds were all terminal; the head then moved (CR1-CR3 fixes, D1 colour darken, and a NEW user-visible U1/U2 hint), so this re-verifies the remediation is correct and introduced no regression.

**Verified clean (no findings):**

- **CR1 — memoization key is complete.** `_slash_runs` (editor.py:1864-1878) keys the per-render-pass memo on `(text, _command_names, _name_choices, _argument_commands, picker.mode, picker.is_open(), picker.is_pending())`. Cross-checked against every input `_compute_slash_runs` (1880-1939) actually reads: `self.text`, `_command_names` (command-word membership), `_is_name_argument_command`→`NAME_ARGUMENT_COMMANDS` (a class constant, immutable), `slash_argument(self.text, self._argument_commands)`, `_name_choices`, and the picker `mode`/`is_open`/`is_pending` gate for the unknown treatment. Every mutable input is in the key; the two frozensets and the tuple are hashable and compared by value. A keystroke changes `self.text` (and usually the picker flags), so a stale run can't survive an input change. The cache is a plain instance attribute recomputed lazily, so it can't leak across buffers (one Editor = one buffer). Correct.

- **CR2 — soft-wrap test genuinely exercises the wrap branch.** `test_a_name_token_that_soft_wraps_is_highlighted_across_both_rows` drives a 29-char name at width 24 and asserts the premise `len(wrapped.get_offsets(0)) >= 2` before checking the paint, so it fails loudly if the name ever stops wrapping. It gathers green runs across all wrapped rows and asserts `painted == long_name` — exercising the `wraps_on`/`section_end` boundary math in `_slash_cells` (1965-1980) that the single-row tests never reach, and confirming the tail (`go now`) picks up no argument colour. Real branch coverage.

- **CR3 — removed split is truly dead.** `name = argument[lead:].split(" ", 1)[0].split("\n", 1)[0]` lost the `.split("\n", 1)[0]`. The single-content-line guard at 1898 (`if len(lines) > first + 1: return None`) has already returned before this line whenever the buffer holds a newline, so `argument` can carry no newline and the second split could only ever be a no-op. Dead-code removal is safe.

- **Hint gating and self-clear (U1/U2) — correct where it fires.** `_name_switch_hint` (1009-1035) returns the hint only when `_is_name_argument_command(self._argument_command)` AND `list_argument` is `<name><space><empty-tail>` (`partition(" ")` with non-empty name, a separator, and `not tail.strip()`). Confirmed by repro on the real harness: `/team frontend-guild` → Enter parks `/team frontend-guild ` and the notice is the hint; typing `f` clears it (tail non-empty); bare `/team ` (rows up) shows nothing; `/login anthropic ` (enum-tail) shows nothing. The D1 colour change (`$lo-muted`→`$lo-dim`) and its tests (`_ink_of` now asserts `dim`) are consistent.

**Findings:**

**CR4 — MAJOR — the new notice write clobbers the app's `/logout` and `/effort` informational notices on the next keystroke (regression).**
The hint is applied with an UNCONDITIONAL `self._picker.set_notice(hint or "")` at `editor.py:2536`, inside the `else` branch of `_sync_picker` that runs for **every** argument command on **every** keystroke — not just NAME+message commands. For a non-name argument command `_name_switch_hint` returns `None`, so this calls `set_notice("")` every resync, wiping any notice the app had set.

Those notices are owned by the app's `on_argument_query_opened` (app.py:9554 `/effort` unavailable, app.py:9582 `/logout` "no stored credentials"), which fires **only on command-word change** — nothing re-sets them per keystroke. The pre-remediation `_sync_picker` never touched the notice on a same-command keystroke, so the notice persisted; now it is erased on the first query character.

Reproduced on head `be32e944` (and confirmed the pre-remediation head `1cc3832b` keeps the notice):

```
# /logout on an account with no stored credentials
"/logout "  → set_notice("no stored credentials — nothing to log out of.")  → notice shown, display=True
"/logout x" → notice = ''                                                   → picker closed  (REGRESSION)

# /effort on a model that takes no effort key
"/effort "  → set_notice("this model takes no effort setting")  → shown, display=True
"/effort h" → notice = ''                                        → wiped   (REGRESSION)
```

Since those lists are empty by construction, the notice IS the entire list — so the user typing to narrow a non-existent list watches the only explanation disappear. `/mcp`'s builder-set notices are also at risk on the same path (the app comments specifically that they must survive a refill).

Uncaught by tests: the added tests only cover the NAME+message and enum-tail-shows-nothing cases; none asserts a non-name notice PERSISTS across a keystroke.

Suggested one-line fix (do not need a patch here): gate the write to NAME+message commands, so the hint owns the notice channel only for `/team`·`/agent` (where the app always sets it to `""`) and leaves the other commands' notices intact:

```python
if self._is_name_argument_command(self._argument_command):
    self._picker.set_notice(self._name_switch_hint(list_argument) or "")
```

**Gates (head be32e944):**
- flake8 — clean (editor.py, command_picker.py, test_command_picker.py)
- black --check 26.1.0 — clean (423 files unchanged)
- isort --check-only --profile black — clean
- pyright --pythonpath .venv/bin/python (editor.py) — 0 errors, 0 warnings
- `pytest tests/unit/tui/test_command_picker.py` — **236 passed**; full `tests/unit/tui` was still running at post time on the directly-touched path, green so far — the CR4 regression is not covered by the suite (proven by direct repro), so a green run does not clear it.

**Verdict: NOT clean — one MAJOR (CR4) blocks terminality on head `be32e944`.** CR1/CR2/CR3 and the U1/U2 hint gating/self-clear are confirmed correct; the D1 colour change is consistent. This round is NOT terminal: it needs a remediation that scopes the notice write to NAME+message commands (or otherwise stops it erasing app-owned notices) and a regression test asserting `/logout`/`/effort` notices survive a same-command keystroke, then a fresh round-3 delta on the new head.

damianvtran:

Implementer: coder subagent (lopdev team). Responding to the round-2 code review (`### Agent review — round 2`).
Scope of remediation: `be32e944..f0c6af6a` (the CR4 fix). Full change is `origin/main..f0c6af6a`.

Round-2 design and UX rounds were both terminal/clean; the code round carried one MAJOR blocker (CR4). It is now fixed.

- **CR4 — MAJOR (regression) — FIXED (`f0c6af6a`).** The U1/U2 hint applied `self._picker.set_notice(hint or "")` **unconditionally** in the `_sync_picker` argument branch, which runs for **every** argument command on **every** keystroke. For a non-name command `_name_switch_hint` returns `None`, so the write became `set_notice("")` on each resync and erased the app-owned informational notices — `/logout` ("no stored credentials — nothing to log out of."), `/effort` ("this model takes no effort setting"), and `/mcp`'s builder notices. Those are set **once** on the command-word change (`on_argument_query_opened`) and never re-set per keystroke, and for those empty-by-construction lists the notice *is* the whole content, so the first query character made the only explanation vanish and closed the picker.

  **Fix** — gated the write to NAME+message commands, exactly as suggested:
  ```python
  if self._is_name_argument_command(self._argument_command):
      self._picker.set_notice(self._name_switch_hint(list_argument) or "")
  ```
  Only `/team`·`/agent` reach it, and for those the app always sets the notice to `""`, so the hint owns the channel cleanly while every other command's notice survives a same-command keystroke. Comment on the guard records why.

**Regression test added (`f0c6af6a`):** `test_app_owned_notice_survives_a_same_command_keystroke`, parametrized over `/logout` (empty-credential notice) and `/effort` (no-effort-key notice). It sets the notice the app would set on the command-word change, types a query character on the same command, and asserts the notice is still present. Confirmed it **fails on the pre-fix write** (both params → `notice == ''`) and **passes after the gate** — the coverage gap the round called out is closed.

**Manual repro (real `PickerHarnessApp`, driving the live `_sync_picker`), before vs after the fix:**
```
# BEFORE FIX (unconditional set_notice)
/logout   → notice "no stored credentials — nothing to log out of."
/logout x → notice ''                                                   (BUG: wiped)
/effort   → notice "this model takes no effort setting"
/effort h → notice ''                                                   (BUG: wiped)

# AFTER FIX (gated to /team·/agent)
/logout   → notice "no stored credentials — nothing to log out of."
/logout x → notice "no stored credentials — nothing to log out of."     (PERSISTS)
/effort   → notice "this model takes no effort setting"
/effort h → notice "this model takes no effort setting"                 (PERSISTS)
```
The U1/U2 hint itself is unaffected: `/team frontend-guild` → Enter still parks `/team frontend-guild ` with the "Enter to switch · type a message to send" notice, and typing clears it.

**Gates re-run from the worktree (`PYTHONPATH=/tmp/lop-composer-ux .venv/bin/python`, TUI env `env -u NO_COLOR TERM=xterm-256color`):**
- `flake8 .` — clean
- `black --check` (black==26.1.0, whole tree) — clean (423 files unchanged)
- `isort --check-only --profile black local_operator tests` — clean
- `pyright --pythonpath .venv/bin/python .` — 0 errors, 0 warnings
- `pytest tests/unit/tui/test_command_picker.py tests/unit/tui/test_minimalism.py` — 245 passed (incl. the 2 new regression params); full `tests/unit/tui` running as the CI backstop.

CR1/CR2/CR3, the D1 colour change, and the U1/U2 hint gating/self-clear were all confirmed correct in round 2 and are untouched here. This remediation scopes the notice write and adds the missing regression test; a fresh round-3 delta on head `f0c6af6a` is warranted since the head moved.

damianvtran:
### Agent review — round 3

Reviewer: reviewer (lopdev team) — Claude Sonnet 4.5
Scope: `be32e944..f0c6af6a` (round-2 remediation delta; full change is `origin/main..f0c6af6a`)

Tight confirmation of the round-2 CR4 remediation only — not a full re-review. Rounds 1–2 code findings, and the round-2 design/UX rounds (TERMINAL/clean), are unaffected by this one-line guard.

**CR4 (was MAJOR, round 2) — RESOLVED.**

The regression: `_sync_picker`'s U1/U2 hint write ran `self._picker.set_notice(hint or "")` unconditionally on every argument keystroke, so for `/logout` and `/effort` — whose empty-by-construction lists carry an app-owned informational notice set once on the command-word change — the first query character called `set_notice("")` and wiped the only content on screen.

The fix (`editor.py:2547-2548`) gates the write behind `_is_name_argument_command(self._argument_command)`. Verified:

- **Guard is the right predicate and complete.** `_is_name_argument_command` tests membership in `NAME_ARGUMENT_COMMANDS = ("team", "teams", "agent", "agents")` — exactly the commands whose hint legitimately owns the notice channel (the app always sets their notice to `""`). `_name_switch_hint` *already* returned `None` for non-name commands internally, so the sole residual harm was the `set_notice("")` erase; the gate suppresses precisely that and nothing else. `list_argument` is in scope and is the correct value passed to `_name_switch_hint` at the call site. No argument command that needs the hint is excluded — only `/team`·`/agent` reach the write, and they still do.
- **Regression reproduced fixed on head.** `test_app_owned_notice_survives_a_same_command_keystroke[logout]` and `[effort]` both PASS on `f0c6af6a`.
- **Test genuinely fails pre-fix / passes post-fix.** Ran the new test against `be32e944:editor.py` (guard absent): both parametrizations FAIL (`assert '' == 'this model takes no effort setting'` / `'no stored credentials …'`). Against `f0c6af6a`: both PASS. Byte-exact restore of the working tree confirmed (`git diff --exit-code` clean). This is genuine regression coverage in both directions.
- **Nothing new introduced.** The positive-path hint tests still pass on head: `test_autofill_shows_the_switch_or_send_hint`, `test_the_hint_is_withdrawn_once_a_message_is_typed` (self-clears on message keystroke), `test_a_bare_name_list_shows_no_hint`, `test_enum_tail_completion_shows_no_switch_hint`. The U1/U2 hint still appears for `/team`·`/agent` autofill and withdraws on keystroke; no other notice path regressed.

**Gates (touched files / touched suite):**
- flake8 — clean
- black --check 26.1.0 — clean (2 files unchanged)
- isort --check-only --profile black — clean
- `pytest tests/unit/tui/test_command_picker.py` — **238 passed** (18.6s)

The full `tests/unit/tui` run did not finish within the harness time budget under load; the touched suite passes in full and CI is the backstop for the remainder.

**Verdict: clean.** No blocker/major findings remain in the delta, and none remain across all rounds. Round 3 is **TERMINAL on head `f0c6af6a`**. No new minors or nits.

PR #248, opened 2026-08-22
feat(mobile): fix todos overflow, naming, approvals, images, and new-session UX

damianvtran:
## Summary of Changes

Seven fixes to the mobile phone portal (`local_operator/mobile/`), all reported from real phone use. The portal is a React/TSX SPA the daemon serves; "owned" sessions are the ones started from the phone (run in a daemon-spawned child), as opposed to "mirror" sessions the TUI publishes.

**UI / layout**

- **Todos no longer crowd out the conversation.** The todos panel sits *above* the transcript, so a long list expanded by default pushed the messages off the top of the screen. It now starts **collapsed** (the `done/total` count in the header is the at-a-glance signal), and when expanded its body is capped to `~40dvh` and scrolls internally — even a 20-item list can never hide the messages.
- **Empty sessions read as ready, not broken.** A just-started session showed a blank scroll area. It now shows a `no messages yet` / `send a message below to get started` placeholder, hidden the moment a turn begins streaming.
- **New-session picker reworked + `/tmp` added.** The working-directory picker is now tagged quick-picks (**home / tmp / recent**) as the primary path, with the free-text field demoted to an "or type a path…" fallback. `/tmp` is a common scratch root, so it is offered as a start dir and the spawn gate now admits *home OR the system temp dir* (resolved, so a symlinked `/tmp` still matches). The tmp row is labelled `tmp` rather than its ugly basename (`/private/var/folders/…/T`).

**Behaviour**

- **Phone-started sessions get NAMED.** Conversation naming lived only in the TUI's `OperatorApp`, so an owned session stayed `mobile session` forever — the header and session list had nothing better to show. The owned handle now fires the same cheap, isolated, single-attempt title call (`session.complete_once` via `naming.generate_title`) on the first substantive prompt, skipping low-signal openers, and pushes the name live. The seed name is now empty (not the frozen `mobile session`), so the phone's own fallback (`untitled` / cwd) shows until the real title lands.
- **Full-auto is honoured on mobile.** A device whose saved `tool_approval_mode` is `auto` still got an approval card. The owned handle now reads that config at spawn (the same key the TUI adopts at boot in `_load_approvals_default`) and answers the approval gate inline, so full-auto sessions never park a card no front end would present.
- **Concurrent approvals no longer hang.** A parallel tool batch can open several write/exec approvals at once (`shared` tools run in parallel). The fold used to *replace* the pending card, so the second approval overwrote the first and the first tool hung forever with no way to answer it. The fold now **queues** requests (`push_pending`/`pop_pending`, cleared by request id so concurrent gates settle in any order), surfaces the front card with a **"1 of N"** badge, and reveals the next as each is answered. The TUI already serialized approvals (one card, rest queue) — this brings the phone to parity; no TUI change needed.
- **Image attachments render inline.** A sent image showed nothing in the transcript (the projection replaced it with a `[N images attached]` note). User-turn images now render inline like the TUI's image block, carried as **lightweight references** (`{index, mime_type}`) on the projection plus a lazy `GET /api/sessions/{pid}/image?entry=&i=` endpoint that decodes the bytes from the on-disk transcript. Bytes are never inlined in the projection, because a repaint fires on every streaming token and re-shipping base64 per token would swamp the SSE.

Change class: **C1 (standard, pre-authorized)** — scoped fixes to one subsystem, clear rollback (revert one commit), no schema/auth/migration surface. The one new wire field (`pending_count`) and new endpoint are additive; older phone bundles ignore both.

## Related Issue(s)

- Related: none filed (reported directly from phone use).

## Impact

- No breaking changes. Adds one projection field (`pending_count`), one transcript-entry field (`images`, references only), one directories field (`tmp`), and one read-only image endpoint — all additive and ignored by older clients.
- No dependency updates.
- Security: the image endpoint reuses the existing auth gate and reads only from the owner's own transcript store; the spawn gate is *widened by exactly one root* (system temp dir), still refusing anything outside home/tmp (`/etc` → 400, verified). Full-auto approval is adopted from the owner's own saved config, matching existing TUI behaviour — it does not lower any bar the desktop keeps.
- Performance: image bytes are fetched lazily and cached `immutable`; the projection stays small. Naming is one isolated, single-attempt completion per session, on the same path the TUI already uses.

## Testing Details

**Unit tests** (added, `tests/unit/mobile/`):

- `test_owned.py` (new): full-auto approves inline with no card; ask-mode parks a card and a *second concurrent* gate queues behind it, answering the front reveals the next; the naming worker names an unnamed session once and a low-signal opener does not spend the attempt; the default seed name is empty.
- `test_projection.py`: pending queue shows the front + counts; out-of-order pop keeps the other card; `set_pending` still replaces for the TUI mirror; image refs are index+mime only (no base64 leak); history fold and the echoed-row upgrade both carry image refs.
- `test_daemon.py`: spawn-dir gate allows home and tmp only (`/etc` refused); directories endpoint offers `tmp`; `_image_bytes` decodes the Nth image block from the transcript and misses cleanly on bad index/entry.

`pytest tests/unit/mobile tests/unit/session/test_conversation_name_persistence.py` — **63 passed**.

**Gates** (PR head `b221f5dc`):
- `flake8` / `black --check` / `isort --check-only` / `pyright` — clean on `local_operator/mobile` + tests.
- `pnpm build` (tsc + vite) — clean.

**End-to-end spawn gate** (against a throwaway daemon):
```
POST /api/sessions/start {"cwd":"<system tmp>"} → {"ok":true,"pid":15140}   # /tmp accepted
POST /api/sessions/start {"cwd":"/etc"}         → 400 not an allowed start directory
GET  /api/sessions/103/image?entry=msg-img-1&i=0 → 200 image/png 738 bytes   # lazy image served
```

**Visual validation** — every changed screen rendered in a real browser against fixture projections (screenshots attached below): empty-session placeholder, todos collapsed-by-default + expanded-with-height-cap, inline image in a user bubble, the "1 of 2" multi-approval badge, and the tagged new-session picker with the `tmp` row.

## Checklist

- [x] I have run formatting (black/isort), linting (flake8), type checks (pyright), and the web build, and they pass.
- [x] I have updated the documentation when required (in-code why/constraint comments added; no doc file needed).
- [x] Security considerations are addressed (auth-gated image endpoint; spawn gate widened by one root only; full-auto adopted from the owner's own config).

### DIFF
--- a/local_operator/mobile/daemon.py
+++ b/local_operator/mobile/daemon.py
@@ -393,6 +393,54 @@ def _history_page(record: SessionRecord, before: str | None, limit: int) -> tupl
     return page, has_more
 
 
+def _image_bytes(record: SessionRecord, entry_id: str, index: int) -> tuple[bytes, str] | None:
+    """Decode the ``index``-th image block of message ``entry_id`` from the
+    session's on-disk transcript into raw bytes plus mime type.
+
+    Reads from disk (not the live fold) so it serves attachments from history
+    the projection tail dropped as well as recent ones, and reuses the
+    transcript's own attachment resolution — the same base64 the model saw.
+    Returns ``None`` for any miss (unknown message, out-of-range index, a
+    reference that no longer resolves) so the caller answers a clean 404.
+
+    Runs off the event loop (``asyncio.to_thread`` at the call site): building
+    the history rehydrates every message and is not loop-safe work.
+    """
+    import base64
+    import binascii
+
+    from local_operator.harness.types import ImageContent, Message
+    from local_operator.paths import config_dir
+    from local_operator.session.transcript import Transcript
+
+    directory = config_dir() / "sessions" / record.session_id
+    if not (directory / "transcript.jsonl").exists():
+        return None
+    try:
+        transcript = Transcript(directory)
+        history = transcript.build_llm_history()
+    except Exception:  # noqa: BLE001 — an odd transcript serves no image, not a 500
+        logger.exception("image fetch: history fold failed for %s", record.session_id)
+        return None
+    message = next((m for m in history if isinstance(m, Message) and m.id == entry_id), None)
+    if message is None or not isinstance(message.content, list):
+        return None
+    images = [b for b in message.content if isinstance(b, ImageContent)]
+    # ``index`` is the position among IMAGE blocks (what _image_refs emits),
+    # not among all content blocks — text blocks do not count.
+    if index < 0 or index >= len(images):
+        return None
+    data = images[index].data
+    if not data:
+        return None
+    try:
+        raw = base64.b64decode(data)
+    except (binascii.Error, ValueError):
+        logger.warning("image fetch: undecodable base64 for %s[%d]", entry_id, index)
+        return None
+    return raw, images[index].mime_type or "image/png"
+
+
 def _fan_out(entry: SessionEntry) -> None:
     """Push the current projection to every open SSE queue for this session.
     A slow phone's queue fills; repaints supersede, so evict the oldest and
@@ -789,6 +837,41 @@ def build_app(daemon: MobileDaemon):
             }
         )
 
+    async def api_session_image(request: Request) -> Response:
+        """One image attachment's bytes, fetched lazily by the transcript.
+
+        The projection carries only lightweight image REFERENCES (entry id +
+        block index + mime) so a per-token repaint stays small; the pixels are
+        served here on demand. The bytes come from the on-disk transcript
+        (which resolves the attachment store back to inline base64), so this
+        works for history the live fold long dropped as well as the tail. The
+        response is cacheable and immutable: a message's attachments never
+        change, and the ``entry``+``i`` pair is a stable content key.
+        """
+        denied = gate(request)
+        if denied is not None:
+            return denied
+        pid = int(request.path_params["pid"])
+        entry = daemon.table.entries.get(pid)
+        if entry is None:
+            return JSONResponse({"error": "unknown session"}, status_code=404)
+        entry_id = request.query_params.get("entry", "")
+        try:
+            index = int(request.query_params.get("i", "0"))
+        except ValueError:
+            return JSONResponse({"error": "bad image index"}, status_code=400)
+        if not entry_id:
+            return JSONResponse({"error": "entry id is required"}, status_code=400)
+        found = await asyncio.to_thread(_image_bytes, entry.record, entry_id, index)
+        if found is None:
+            return JSONResponse({"error": "no such image"}, status_code=404)
+        data, mime_type = found
+        return Response(
+            content=data,
+            media_type=mime_type,
+            headers={"Cache-Control": "public, max-age=31536000, immutable"},
+        )
+
     async def api_command(request: Request) -> Response:
         """The one mutation endpoint: {op, ...} → control frame. Keeping
         mutations on one route mirrors the registrant's dispatch and keeps
@@ -830,14 +913,16 @@ def build_app(daemon: MobileDaemon):
         except ValueError:
             return JSONResponse({"error": "invalid JSON"}, status_code=400)
         cwd_raw = str(body.get("cwd") or Path.home())
-        # Resolve to a real directory under the owner's home: the picker only
-        # ever sends those, and the spawn runs with the daemon's environment,
-        # so the check is about fat-fingered/traversed input, not trust.
+        # Resolve to a real directory the picker is allowed to open: anywhere
+        # under the owner's home, OR the system temp dir. The spawn runs with
+        # the daemon's own environment (it is the owner's account either way),
+        # so the check guards against fat-fingered/traversed input, not trust
+        # — /tmp is a deliberate, common scratch root the phone offers as a
+        # starting directory, so it is on the allowlist beside home.
         cwd_path = Path(cwd_raw).expanduser().resolve()
-        home = Path.home().resolve()
-        if not cwd_path.is_dir() or not (cwd_path == home or home in cwd_path.parents):
+        if not cwd_path.is_dir() or not _spawn_dir_allowed(cwd_path):
             return JSONResponse(
-                {"error": f"not a directory under home: {cwd_raw}"}, status_code=400
+                {"error": f"not an allowed start directory: {cwd_raw}"}, status_code=400
             )
         cwd = str(cwd_path)
         provider = body.get("provider")
@@ -927,7 +1012,9 @@ def build_app(daemon: MobileDaemon):
         if denied is not None:
             return denied
         recent = await asyncio.to_thread(_recent_directories)
-        return JSONResponse({"home": str(Path.home()), "recent": recent})
+        # ``tmp`` is offered as an explicit scratch start dir beside home and
+        # the recents — the spawn gate admits it (see _spawn_dir_allowed).
+        return JSONResponse({"home": str(Path.home()), "recent": recent, "tmp": _tmp_dir()})
 
     async def api_past_sessions(request: Request) -> Response:
         """Resumable past sessions — the phone's "go back to a conversation"
@@ -965,6 +1052,7 @@ def build_app(daemon: MobileDaemon):
         Route("/api/sessions/search", api_search_sessions),
         Route("/api/sessions/{pid:int}/events", api_session_events),
         Route("/api/sessions/{pid:int}/history", api_session_history),
+        Route("/api/sessions/{pid:int}/image", api_session_image),
         Route("/api/sessions/{pid:int}/command", api_command, methods=["POST"]),
         Route("/api/commands", api_commands),
         Route("/api/models", api_models),
@@ -1050,6 +1138,27 @@ def _search_sessions(query: str, limit: int = 40) -> list[dict[str, Any]]:
         if len(out) >= limit:
             break
     return out
+
+
+def _tmp_dir() -> str:
+    """The system temp directory, resolved. Offered as a scratch start dir on
+    the phone's new-session form and admitted by the spawn gate. Resolved (not
+    the raw ``/tmp``) so it matches the gate's resolved comparison on hosts
+    where ``/tmp`` is a symlink (macOS: ``/private/tmp``)."""
+    import tempfile
+
+    return str(Path(tempfile.gettempdir()).resolve())
+
+
+def _spawn_dir_allowed(cwd_path: Path) -> bool:
+    """Whether a resolved directory may host a phone-started session: anywhere
+    under the owner's home, or the system temp dir (a common scratch root).
+    Both bounds are on RESOLVED paths so a symlinked ``/tmp`` still matches."""
+    home = Path.home().resolve()
+    if cwd_path == home or home in cwd_path.parents:
+        return True
+    tmp = Path(_tmp_dir())
+    return cwd_path == tmp or tmp in cwd_path.parents
 
 
 def _recent_directories(limit: int = 8) -> list[str]:
--- a/local_operator/mobile/daemon.py
+++ b/local_operator/mobile/daemon.py
@@ -844,9 +844,15 @@ def build_app(daemon: MobileDaemon):
         block index + mime) so a per-token repaint stays small; the pixels are
         served here on demand. The bytes come from the on-disk transcript
         (which resolves the attachment store back to inline base64), so this
-        works for history the live fold long dropped as well as the tail. The
-        response is cacheable and immutable: a message's attachments never
-        change, and the ``entry``+``i`` pair is a stable content key.
+        works for history the live fold long dropped as well as the tail.
+
+        Cacheable and immutable: the true content key is the ``entry`` id — a
+        globally-unique message uuid — plus the image-only ``i``. The ``pid``
+        in the path only routes to a live session; pids recycle, but a
+        recycled pid maps to a DIFFERENT session whose transcript does not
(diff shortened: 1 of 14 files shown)


damianvtran:
### Design review — round 1
Reviewer: designer (Claude Opus 4.8)
Scope: main..b221f5dc4e54096dd501c0bd03d548260fd3b53f

Solid, on-brand work — the collapsed-todos fix, the quiet empty state, and the tagged picker all land. One MAJOR (the image failure state isn't designed, it degrades to a raw browser glyph), plus a few minors; not terminal yet.

- **D1** (major) — Inline-image failure state is not designed: a decode/404 failure renders the browser's native broken-image glyph — a tiny red-tinted icon pinned to the top-left of the accent bubble (visible in `image-transcript2.png`, magnified it's the stray red speck at the bubble's inner corner). This reads as a defect against the otherwise-polished accent bubble and directly contradicts the code comment's "degrades … gracefully" claim. On a phone over flaky network, image loads fail routinely, so this is a common state, not an edge one. — `image-transcript2.png` — Add an `onError` fallback that swaps to a designed placeholder (a bordered box at the reserved size with a muted "image unavailable" / attachment glyph in `text-ink-dim`), so a failure looks intentional and quiet rather than broken.

- **D2** (minor) — Image has no reserved dimensions before it decodes: `` gives the element zero intrinsic height until bytes arrive, so the bubble paints at text-only height and then jumps taller when the image pops in — a reflow the user sees as motion, which §7 explicitly rules out (no motion the eye tracks for a thing that isn't the turn's one indicator). — `image-transcript2.png` — Reserve space with an aspect-ratio box or a min-height skeleton so the bubble settles once instead of expanding on decode.

- **D3** (minor) — The "demoted free-text fallback" isn't visually demoted, and it duplicates the selected quick-pick. With `home` selected (accent row "damian ~"), the free-text field below is pre-filled with the same path shown as the absolute `/Users/damian`, at the same `min-h-11` / `border-control` weight as the quick-pick rows — so it reads as a fourth, unlabeled option rather than a fallback, and shows the same directory twice in two different notations. — `new-session-final.png` — Give the field a lighter treatment (borderless or ghost until focused) and keep it empty with the "or type a path…" placeholder when a quick-pick is selected, rather than mirroring the chosen path.

- **D4** (nit) — "1 of 2" badge is correctly quiet (`text-ink-dim` mono, right-aligned opposite the accent "approval needed" label — right call to not spend accent on it), but "1 of N" is a position, not a promise that answering reveals the next. A first-time user may not realize more approvals are queued behind this card. — `multi-approval.png` — Consider "1 of 2 · answer to continue" or similar once, or accept as-is; low stakes since the next card repaints in place.

- **D5** (nit) — Tag chips sit at `text-meta` in `text-ink-dim` on `bg-sunken` — a low-contrast pair at the smallest type size; legible on this display but marginal on a phone in daylight. The chip also restates the row name in two cases (`tmp` row labeled `tmp`, near-restated for recents). — `new-session-final.png` — Bump the chip to `text-ink-muted` for a touch more contrast; the label duplication is harmless, leave it.

Verdict: NOT TERMINAL — resolve D1 (designed failure state) and ideally D2 (reserve image space) this round; D3–D5 are polish. Re-capture the inline-image shots with a deliberately-failing src to prove the new broken state after the fix.

damianvtran:
### Agent review — round 1
Reviewer: reviewer (claude-opus-4.1)
Scope: main..b221f5dc4e54096dd501c0bd03d548260fd3b53f

Clean — no blockers, no majors. Seven fixes are correct, well-tested, and parity with the TUI holds. Gates green locally (flake8 0, pyright 0, 52 pytest passed, `pnpm build` ok). A few minors/nits below are follow-ups, not a reason for another round; this round is TERMINAL.

Verified the load-bearing claims: the pending queue is keyed by `request_id` so out-of-order answers are honoured (`pop_pending` filters by id, tested); the image index aligns because `_image_refs` (emitter) and `_image_bytes` (reader) both count `ImageContent` blocks only and attachment resolution swaps `data` without adding/removing blocks; the spawn gate resolves paths and uses `home in cwd_path.parents` (no `/home/user` vs `/home/user-evil` prefix bug) — no traversal past home/tmp; the image endpoint is bounded to `config_dir()/sessions/<session_id>` and selects a message by id + an image-only index, so no cross-session leak and no arbitrary-file read; and the projection carries refs only — no base64 re-ships on repaint (test asserts no `data` key).

**Blockers**
- None.

**Majors**
- None.

**Minors**
- `owned.py:254` — `asyncio.ensure_future(self._name_conversation_worker(text))` keeps no strong reference. Per CPython docs a task with no live reference can be garbage-collected mid-flight; the naming worker suspends on provider I/O for up to `TITLE_TIMEOUT_S`, which is exactly when it is collectable. The TUI equivalent retains it via `run_worker(group="naming")`. Consequence is only a missed title (decoration), and it matches this file's existing `_run_turn_task` pattern, so it is minor — but the honest fix is to stash the task in a set on the handle and `add_done_callback(set.discard)`.
- `screens/session-view.tsx:125` — empty-placeholder condition `transcript.length === 0 && !projection.streaming`. An owned session's `prompt()` does not optimistically echo (it relies on the server's `MessageStartEvent`), so between the tap and the session loop emitting that event the transcript is still empty and `streaming` still false — the "no messages yet" placeholder lingers, then flips to the user row. A brief flash on the first send only. Gating additionally on a local "sent" flag would remove it.

**Nits**
- `projection.py:177-179` — stale comment. It says the history fold should "Render a receipt line naming the attachments, matching the live fold's receipt," but the code no longer synthesizes `[N images attached]` text; it passes `text` as-is plus `images=refs` and the client renders thumbnails. Update the comment to match the new inline-image behaviour.
- `daemon.py` `api_session_image` — `Cache-Control: public, max-age=31536000, immutable` is keyed on `pid`, which is ephemeral and reusable across daemon/session restarts. The `entry` id is the real content key (effectively unique), so a stale hit needs same pid + same message id — negligible in practice, but `session_id` would be the honest immutable key.
- `projection.py:505-506` — the echo-upgrade rewrites `entry.id = message.id`. The transcript renders rows with `key={e.id}`, so the id change remounts that row as the images load — a small reflow of the user's own bubble on the TUI-mirror path. Cosmetic.

damianvtran:
Reviewer (author): implementing agent (claude opus)
Scope: b221f5dc..bbe41f1a (current head bbe41f1a)

The code review verdict was clean (no blockers/majors) — thank you. Addressing the minors:

- **Naming worker fire-and-forget GC (owned.py)** — FIXED in bbe41f1a. The detached naming task is now held in a `self._background_tasks` set with an `add_done_callback(discard)` cleanup, so the event loop cannot garbage-collect it mid-flight and silently drop the title.
- **Stale "receipt line" comment (projection.py history fold)** — FIXED in bbe41f1a. Replaced with an accurate description of the image-reference render (the code no longer synthesizes a `[N images attached]` receipt line; it carries refs and renders thumbnails).
- **`pid`-keyed immutable cache header (daemon.py image endpoint)** — DOCUMENTED as safe in bbe41f1a rather than restructured. The true content key is the `entry` id (a globally-unique message uuid) plus the image-only index; the `pid` in the path only routes to a live session. A recycled pid maps to a *different* session whose transcript does not contain this message uuid, so it 404s rather than serving another session's cached bytes — the uuid content key is what makes `immutable` correct despite the mutable pid. Added a comment spelling this out.
- **Empty-placeholder flash on first send (minor)** — DEFERRED (ticketable). The `no messages yet` placeholder can flash for one frame on the first owned-session send before the server's `MessageStartEvent` arrives, because there is no optimistic user-row echo on the phone. Fixing it properly means adding an optimistic echo (which the current design deliberately avoids, to prevent ghost rows on a rejected prompt — see `_run_turn_task`), so it is out of scope for this round.

All gates green on the current head: flake8 / black / isort / pyright clean, `pytest tests/unit/mobile` 52 passed, `pnpm check` (tsc) clean.

damianvtran:
Reviewer (author): implementing agent (claude opus)
Scope: b221f5dc..bbe41f1a (current head bbe41f1a)

- **D1 (major)** — FIXED in bbe41f1a. Added an `AttachmentImage` component (`components/transcript.tsx`) that owns the loading/error/loaded states. On a 404 or decode failure it now renders a designed muted placeholder — a `⊘` glyph + `image unavailable` label in a `bg-sunken` box at the reserved size — instead of the browser's native broken-image glyph. Confirmed in a real browser: the DOM renders `⊘ image unavailable` for a deliberately-unresolvable attachment reference (fixture session with an image ref whose bytes 404). This is now a common, designed state rather than a defect.
- **D2 (minor)** — FIXED in bbe41f1a. The image paints into a fixed-height (`h-40`) reserved frame with a `loading…` state, so the bubble no longer reflows taller when the bytes decode. Dropped `loading="lazy"` on the attachment `` so `onError` fires reliably for these in-view thumbnails (the lazy attribute was deferring the load and suppressing the error event).
- **D3 (minor)** — FIXED in bbe41f1a. The free-text field is now demoted below the quick-picks in visual weight (`border-hairline`, `bg-transparent`, `text-mono-sm text-ink-muted`, smaller min-height) and only mirrors `cwd` when the selection is a CUSTOM path (`cwdIsQuickPick` guard). When a quick-pick is active the field is empty with an "or type another path…" placeholder, so it no longer reads as a fourth option duplicating the home row's absolute path.
- **D4 / D5 (nits)** — DEFERRED. The "1 of N" badge wording and the tag-chip contrast are left as-is for this round; both are working-as-intended per branding §7 (the badge is deliberately quiet, no accent spent). Can ticket if you'd like the badge to read "more follow" explicitly.

Re-captured screenshots after the visual fixes: the working image now sits in the reserved frame above the caption (`image-working-v2.png`), and the failure state shows the `⊘ image unavailable` placeholder (DOM-confirmed).

damianvtran:
### Design review — round 2
Reviewer: designer (claude sonnet 4.5)
Scope: main..bbe41f1a

Clean — D1, D2, D3 are all genuinely resolved on the current head (bbe41f1a). No open BLOCKER or MAJOR findings; the round is TERMINAL.

- **D1** RESOLVED — `AttachmentImage` (`components/transcript.tsx`) owns three explicit states. The `error` branch renders a designed placeholder at the reserved size (`h-40 w-40`, `border-hairline bg-sunken text-ink-dim`) with an `aria-hidden` `⊘` glyph + `image unavailable` meta label, replacing the browser's native broken glyph. DOM-confirmed for an unresolvable ref. This now reads as a quiet, common state rather than a defect.
- **D2** RESOLVED — the image paints into a fixed-height `h-40` frame with an invisible `` and a `loading…` state during fetch, so the bubble no longer grows taller when bytes decode; the working-image screenshot shows the picture sitting in the reserved frame above the caption with no vertical jump. `loading="lazy"` is dropped so `onError` fires for in-view thumbnails. (The loading box is a `w-40` square that resolves to the image's true width once loaded — a horizontal-only settle, no vertical reflow; acceptable, noting it below as a follow-up, not a finding.)
- **D3** RESOLVED — the free-text field (`screens/new-session.tsx`) is demoted below the quick-picks in both order and weight (`border-hairline bg-transparent`, `font-mono text-mono-sm text-ink-muted`, `min-h-9` vs the picks' `min-h-11`) and, via the `cwdIsQuickPick` guard, stays empty with an `or type another path…` placeholder whenever a quick-pick is active. It no longer duplicates the home row's absolute path as an apparent fourth option. (Screenshot `new-session-final.png` predates this change; verified in code.)
- **D4 / D5** (nits) — DEFERRED by author as working-as-intended per branding §7; accepted, no re-litigation.

Follow-up (non-blocking, ticket if desired): the loading placeholder is a fixed `w-40` square while a loaded landscape thumbnail is `h-40 max-w-full`, so the frame width settles horizontally on decode. Height is pinned so there's no reflow the D2 fix targeted; if you want zero horizontal settle too, reserve the aspect box from known dimensions.

Verdict: TERMINAL — design round is CLEAN on head bbe41f1a.

damianvtran:
### Agent review — round 2
Reviewer: reviewer (claude-opus-4-1)
Scope: b221f5dc..bbe41f1a

Clean — the round-1 remediation is correct, introduces no regressions, and all gates pass on the current head (bbe41f1a). This round is TERMINAL.

**Blockers**: none

**Majors**: none

**Minors**: none
- `owned.py`: the `_background_tasks` set + `add_done_callback(self._background_tasks.discard)` is the canonical strong-ref pattern; no leak, cleanup fires on settle. Type `set[asyncio.Future[Any]]` is accepted by pyright.
- `transcript.tsx` `AttachmentImage`: cannot get stuck in `loading` — dropping `loading="lazy"` makes the `` load eagerly so `onLoad`/`onError` fire reliably; the element stays mounted (`invisible`, not unmounted) in loading/loaded so handlers bind. `key={img.index}` is stable, no spurious remount. Error state renders the designed placeholder.
- `new-session.tsx` `cwdIsQuickPick`: first keystroke works — value `cwdIsQuickPick ? "" : cwd`, typing sets `cwd` to the new text, which stops matching a pick, so the field immediately shows the typed value. The guard never blanks input once `cwd` diverges from a quick-pick.

**Nits**: none material.
- Theoretical only: typing a path that exactly equals a quick-pick (e.g. the full home path) would flip `cwdIsQuickPick` true and momentarily clear the field mid-type. Self-correcting (cwd is still a valid pick) and an extreme edge; not worth a change.

Gates on bbe41f1a: `flake8` clean · `pyright` 0 errors · `pytest tests/unit/mobile` 52 passed · `pnpm check` (tsc) clean.

PR #249, opened 2026-08-22
feat(cli): guided first-run setup state + cross-platform onboarding hardening

damianvtran:
## Summary of Changes

New feature + hardening. A brand-new user on a fresh machine hit a wall: the very first `local-operator` launch printed a bare `Error: Hosting platform is not configured.` and exited 255 — no next step, no command named, and the TUI (which owns the `/login` affordance) never opened. Fixing that one error by hand surfaced two more one-line dead-ends in sequence (model, then key), and the escape hatches themselves were broken (`login` stored a key but never closed the loop; Ctrl-C at the credential prompt dumped a traceback). This PR makes first-run followable.

Two independent audits (an architecture pass and a real fresh-`HOME` UX walkthrough) produced a 20-item findings list; this PR addresses all 20 in one branch.

**Headline — interactive keyless launch now enters the TUI in a setup state instead of dying at preflight.** `resolve_hosting_model` raises a new `HostingNotConfiguredError` (a `ValueError` subclass so existing handlers still catch it); the CLI preflight passes it through **only** when the full-screen TUI will actually run (`allow_setup_state`, gated strictly on `stdout.isatty()` + the TUI path), and `OperatorApp` routes it to a setup state that reuses the existing welcome splash + `/login` + `/model` surfaces — no parallel wizard subsystem. The splash's warning row becomes `! welcome — no provider configured yet. Run /login <provider> …` and the status band reads `setup`. Headless / `exec` / non-tty keep fail-fast, but now print a complete quickstart that names hosting **and** model **and** the credential command at once (exit 1, not 255).

**The rest, by finding:**
- **login closes the loop** — `login <provider>` (CLI and TUI) now sets `hosting` + a per-provider default model when config is empty, so the tool's own recommended recovery actually starts the app.
- **default model fallback** — hosting set + model empty resolves a sensible default (`model/defaults.py`) instead of erroring; stale `--model` help examples refreshed.
- **credential prompt hardened** — Ctrl-C → exit 130 quietly; empty/EOF → one plain line, exit 1; ANSI stripped from exception text; non-tty stdin read as one line (documented automation contract); unknown key names warn with a closest-match suggestion; ASCII banner fallback when stdout can't encode box-drawing (legacy Windows codepages); atomic 0600 writes via temp-file + `os.replace`, control chars rejected.
- **Anthropic (callable env-key resolver)** no longer suggests the invalid `credential update API key` — it names `login` when the env var isn't a fixed string.
- **malformed config.yml** → one-line parse error + the file backed up to `config.yml.bad` + defaults, instead of a raw traceback.
- **`config edit`** rejects unknown keys with a difflib suggestion (the old `except KeyError` was dead code).
- **auth-classified stream errors** append the local recovery (`/login` / `login` / `credential update`); the knowledge-embedding 401 fallback warning is downgraded to the log when its cause is the same missing key.
- **headless REPL** raises library loggers to WARNING so INFO/httpx lines don't leak before the first prompt.
- **`~/local-operator-home`** is no longer created unconditionally in `main()` on every invocation and no longer ignores the config-dir override: `paths.agent_home_dir()` honours `LOCAL_OPERATOR_HOME`, creation is lazy at session start, and all readers (agents, server schemas) were migrated together.
- **config + credentials dirs** created 0700 (matching the log dir; transcripts are the same sensitivity class) — at creation only, never chmod'ing an existing dir on upgrade.
- **exit codes** `-1` → `1` on failure paths (255 collided with xargs/ssh sentinels and contradicted the `exec` contract).
- **`config open`** falls back to `$VISUAL`/`$EDITOR` on a headless/SSH box with no GUI opener.
- **shell-env priming** (`$SHELL -l -c` PATH harvest) gated to the subcommands that actually spawn subprocesses, off the `config`/`credential` paths.
- **NO_COLOR / TERM=dumb / non-tty** respected on the first-run and error/warn paths via a small stdlib styling helper (`cli_style.py`), gated on the stream actually written (stderr at the error sites). Note: a handful of other CLI listings (`config list`, `agents`, `teams`) still emit raw escapes and are out of scope for this PR. README corrected (`credentials.env`, angle-bracket examples, first-run description).

Change class: **C2 (feature + cross-platform hardening, pre-authorized)** — user-facing first-run flow plus independent small hardening fixes, each revertible on its own commit; no schema, migration, or auth-contract surface (the new agent-home override defaults to the existing location).

## Related Issue(s)

- Related: none filed (surfaced by a first-run UX audit).

## Impact

- **Behaviour changes worth calling out:** (1) an interactive keyless launch now opens the TUI in a setup state rather than exiting — headless/exec is unchanged in its fail-fast contract, only its message improved; (2) failure exit codes move from 255 to 1 — a script matching on 255 would need updating; (3) `login` now writes `hosting`/`model_name` when they're empty. All three are the intended fixes, not side effects.
- No breaking API/schema changes. The agent-home override defaults to `~/local-operator-home`, so existing installs see the same location; only *when* it's created and *how* the path resolves changed. Dir-perm tightening applies at creation only.
- No dependency updates. The styling/default-model helpers are stdlib-only and kept off the startup import path (the import-graph guard still passes).
- Security: credentials now written atomically at 0600 with control-char rejection; config/credential dirs created 0700; no secrets added to any transcript or log.

## Testing Details

Testing evidence is proof the change works, not a green suite — every fixed dead-end was exercised against a **real fresh `HOME` + `LOCAL_OPERATOR_CONFIG_DIR`**, run from the worktree so the new code (not the installed build) executed. Full transcript with commands and actual responses is attached below; highlights:

**Headline setup state (rendered and viewed, not just asserted):**
- Keyless boot → TUI setup state: splash `! welcome — no provider configured yet. Run /login <provider> …`, band `setup`, splash not retired. Two consecutive frames byte-identical after id-normalisation → no post-paint reflow. Screenshots below.

**Fixed dead-ends (fresh HOME, real execution):**
- Headless keyless launch → complete quickstart naming hosting+model+key+commands, **exit 1** (was a bare one-liner, exit 255).
- `login deepseek` with a fake key → `Stored API key … Set default hosting to 'deepseek' and model to 'deepseek-chat'`, config.yml written, **exit 0** — the loop closes.
- `credential update` Ctrl-C via a real pty → **exit 130**, no traceback; empty piped stdin → `… is required for this step.`, **exit 1**; piped value → stored as one line; typo'd key → `Did you mean OPENAI_API_KEY?` warning; `PYTHONIOENCODING=ascii` → ASCII `+---+` banner, no crash.
- Malformed `config.yml` → one-line parse error + moved to `config.yml.bad` + defaults.
- `config edit hostng radient` → `unknown configuration key: 'hostng'. Did you mean 'hosting'?`, **exit 1**.
- Atomic write: 0600 perms, no `.tmp` leftover; app-created config dir → `drwx------`.
- `config list` does **not** create the agent home (lazy); `LOCAL_OPERATOR_HOME` honoured.
- Auth 401 stream error → local recovery appended; 429 quota text unchanged.
- `config open` with `$EDITOR=cat` and no GUI opener → opens via `$EDITOR`.
- Headless REPL → httpx logger at WARNING, no INFO leak before the first prompt.

**Gates** (rebased onto current `main`, PR head is the tip of `dev-first-run-ux`):
- `flake8` / `black --check` / `isort --check-only` / `pyright` — clean (0 errors).
- `pytest` — non-heavy subset 3005 passed; `session`/`mcp`/`server` 665 passed; `tui` 2210 passed, 4 skipped. New/extended tests cover preflight tty gating + setup-state entry, login-sets-defaults, default-model fallback, config-edit key validation, credential interrupt/EOF/unknown-key/atomic-write, YAML error handling, agent-home override, exit codes, and the NO_COLOR helper.
- One unrelated failure, `test_eval_tool::test_background_streams_output_while_it_runs`, is a **pre-existing timing flake confirmed failing identically on untouched `main`** (verified in a throwaway worktree at the merge-base); it touches none of the files in this PR.

**Note for reviewers:** the second screenshot's `provider controller unavailable — run: local-operator login` hint is a **test-harness artifact** — the `FakeSession` used to render the frame has no live `ProviderController`. The real `/login` path is validated separately by the CLI `login deepseek` evidence above; a reviewer standing up a live session should confirm `/login` end-to-end in the TUI.

## Checklist

- [x] All new and existing tests pass (one pre-existing unrelated flake noted).
- [x] I have updated the documentation when required (README first-run text; in-code why/constraint comments throughout).
- [x] Security considerations are addressed (atomic 0600 credential writes, 0700 dirs, no secrets logged).

## Screenshots

Rendered from the real `OperatorApp` (`run_test` + `save_screenshot`, from the worktree so the new code paints) and viewed as PNGs. Two states captured:

1. **First-run setup state (keyless boot)** — splash reads `! welcome — no provider configured yet. Run /login <provider> (e.g. /login openai) to set one up`, status band reads `setup` (not a model, not "session error"), splash is not retired. This is what a brand-new user now sees instead of `Error: Hosting platform is not configured.` + a dead shell.
2. **Guided step — `/login openai` typed in the composer**, band still `setup`.

SVG + PNG artifacts are attached in the review thread (they render from `run_test`, and consecutive frames are byte-identical after id-normalisation — no post-paint reflow). A design-review round with re-captured frames follows on this PR before merge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,13 +93,17 @@ pip install local-operator     # pipx install local-operator on Linux (PEP 668)
 The install provides both `local-operator` and its short alias `lop` — the
 rest of this page uses `lop`.
 
-Sign in to a provider (or skip this — the app tells you what's missing on
-first run):
+Sign in to a provider (or skip this — on an interactive terminal `lop` opens
+in a setup state and walks you through `/login`; a headless or piped run prints
+the exact commands to configure hosting, model, and a key):
 
 ```bash
 lop login           # lists login-capable providers
 lop login anthropic # OAuth sign-in in your browser
 ```
+
+`lop login <provider>` also sets it as your default hosting (and picks a
+default model) when none is configured yet, so the very next `lop` just works.
 
 Then start it:
 
--- a/local_operator/agents.py
+++ b/local_operator/agents.py
@@ -19,6 +19,7 @@ from pydantic import BaseModel, Field
 
 from local_operator.jsonl import read_jsonl, write_jsonl
 from local_operator.optional import missing_extra_error
+from local_operator.paths import default_agent_cwd
 from local_operator.types import Schedule  # Keep existing Schedule import
 from local_operator.types import (
     AgentState,
@@ -309,7 +310,7 @@ class AgentRegistry:
             presence_penalty=agent_edit_metadata.presence_penalty,
             seed=agent_edit_metadata.seed,
             current_working_directory=agent_edit_metadata.current_working_directory
-            or "~/local-operator-home",
+            or default_agent_cwd(),
         )
 
         return self.save_agent(agent_metadata)
@@ -1438,7 +1439,7 @@ class AgentRegistry:
 
                 # Create a new agent with the imported data
                 agent_id = agent_data["id"]
-                agent_data["current_working_directory"] = "~/local-operator-home"
+                agent_data["current_working_directory"] = default_agent_cwd()
                 # Older profile bundles duplicated the final conversation turn
                 # into agent.yml. History files are ignored below, but leaving
                 # this field intact would still import private conversation
--- a/local_operator/bootstrap.py
+++ b/local_operator/bootstrap.py
@@ -82,7 +82,19 @@ def resolve_hosting_model(
     if not hosting:
         raise ValueError("Hosting platform is not configured.")
     if not model_name:
-        raise ValueError("Model name is not configured.")
+        # Fall back to the provider's default model rather than erroring on a
+        # single empty field — see session_factory.resolve_hosting_model, which
+        # this mirrors. A provider with no known default still raises.
+        from local_operator.model.defaults import default_model_for
+
+        model_name = default_model_for(hosting)
+        if not model_name:
+            raise ValueError(
+                f"Model name is not configured for hosting '{hosting}', and no "
+                "default is known for it. Set one with `local-operator config "
+                "edit model_name <model>` or the --model flag (e.g. gpt-4o, "
+                "claude-3-5-sonnet-latest, deepseek-chat)."
+            )
     return hosting, model_name
 
 
--- a/local_operator/bootstrap.py
+++ b/local_operator/bootstrap.py
@@ -89,12 +89,14 @@ def resolve_hosting_model(
 
         model_name = default_model_for(hosting)
         if not model_name:
-            raise ValueError(
-                f"Model name is not configured for hosting '{hosting}', and no "
-                "default is known for it. Set one with `local-operator config "
-                "edit model_name <model>` or the --model flag (e.g. gpt-4o, "
-                "claude-3-5-sonnet-latest, deepseek-chat)."
-            )
+            # Reuse the resolver's own message rather than inlining a copy: the
+            # two are the SAME error (no default model for this hosting) and a
+            # second literal here drifts from it the first time either is
+            # reworded. Imported lazily to keep this off the session_factory
+            # stack until the rare no-default path is actually hit.
+            from local_operator.session_factory import _no_model_message
+
+            raise ValueError(_no_model_message(hosting))
     return hosting, model_name
 
 
--- a/local_operator/cli.py
+++ b/local_operator/cli.py
@@ -604,8 +604,52 @@ def _propagate_global_flags(parser: argparse.ArgumentParser) -> None:
 
 
 def credential_update_command(args: argparse.Namespace) -> int:
+    """Prompt for and store one credential. Exit 0/1/130.
+
+    The prompt used to let three ordinary interruptions escape as tracebacks:
+    Ctrl-C raised a bare ``KeyboardInterrupt`` all the way out, and an empty
+    value / closed stdin raised a ``ValueError`` whose message carried nested
+    ANSI escapes that the generic red-banner handler in ``main`` then wrapped in
+    a stack-trace panel. None of the three is a program fault \u2014 they are the
+    user cancelling or mis-entering \u2014 so each gets one plain line and a clean
+    exit code: 130 for a cancel (the shell convention for SIGINT), 1 otherwise.
+    """
+    from local_operator.ansi import strip_control_sequences
+    from local_operator.cli_style import ERROR, WARNING, paint
+    from local_operator.providers.registry import PROVIDER_REGISTRY, env_key_name
+
+    # Warn when the key is not one the registry knows, with the closest match \u2014
+    # a typo'd ``OPENAI_API_KY`` otherwise stores silently and the provider
+    # never sees it. Arbitrary keys stay allowed (custom providers are
+    # legitimate); this is advice, not a gate.
+    known_keys = {name for p in PROVIDER_REGISTRY if (name := env_key_name(p.id))}
+    if args.key not in known_keys:
+        import difflib
+
+        close = difflib.get_close_matches(args.key, sorted(known_keys), n=1)
+        hint = f" Did you mean {close[0]}?" if close else ""
+        print(
+            paint(
+                f"Warning: '{args.key}' is not a known provider key.{hint} " "Storing it anyway.",
+                WARNING,
+            ),
+            file=sys.stderr,
+        )
+
     credential_manager = CredentialManager(config_dir())
-    credential_manager.prompt_for_credential(args.key, reason="update requested")
+    try:
+        credential_manager.prompt_for_credential(args.key, reason="update requested")
+    except KeyboardInterrupt:
+        # 130 is the shell's SIGINT convention; the message is one quiet line,
+        # not the red stack-trace panel the generic handler would have drawn.
+        print("\nCancelled.", file=sys.stderr)
+        return 130
+    except (ValueError, EOFError) as exc:
+        # Empty input or a closed stdin. Strip any control sequences from the
+        # message before printing \u2014 the presenter owns the colour, and a nested
+        # escape from deeper in the stack would otherwise repaint the line.
+        print(paint(strip_control_sequences(str(exc)), ERROR), file=sys.stderr)
+        return 1
     return 0
 
 
@@ -654,7 +698,34 @@ def config_open_command() -> int:
 
 def config_edit_command(args: argparse.Namespace) -> int:
     """Edit a configuration value."""
+    from local_operator.cli_style import ERROR, paint
+
     config_manager = ConfigManager(config_dir())
+
+    # Validate the key against the shipped defaults BEFORE writing. The old
+    # ``except KeyError`` was dead code \u2014 ``update_config`` calls
+    # ``Config.set_value`` which is a plain ``dict.__setitem__`` and never
+    # raises for an unknown key \u2014 so a typo like ``config edit hostng radient``
+    # printed "Successfully updated hostng" and wrote a junk key the app never
+    # reads. difflib names the closest real key so the fix is one glance away.
+    from local_operator.config import DEFAULT_CONFIG
+
+    valid_keys = set(DEFAULT_CONFIG.values.keys())
+    if args.key not in valid_keys:
+        import difflib
+
+        close = difflib.get_close_matches(args.key, sorted(valid_keys), n=1)
+        hint = f" Did you mean '{close[0]}'?" if close else ""
+        print(
+            paint(f"Error: unknown configuration key: '{args.key}'.{hint}", ERROR),
+            file=sys.stderr,
+        )
+        print(
+            "Run `local-operator config list` to see the available keys.",
+            file=sys.stderr,
+        )
+        return 1
+
     try:
         # Parse the value to the appropriate type
         value = args.value
@@ -684,12 +755,12 @@ def config_edit_command(args: argparse.Namespace) -> int:
 
         print(f"Successfully updated {args.key} to {value}")
         return 0
-    except KeyError:
-        print(f"\n\033[1;31mError: Invalid configuration key: {args.key}\033[0m")
-        return -1
     except Exception as e:
-        print(f"\n\033[1;31mError updating configuration: {e}\033[0m")
-        return -1
+        # 1, not -1: a shell sees -1 as 255, which collides with the
+        # xargs/ssh "command not found" sentinel and contradicts the
+        # documented 0/non-zero exec contract (item A13).
+        print(paint(f"Error updating configuration: {e}", ERROR), file=sys.stderr)
+        return 1
 
 
 def config_list_command() -> int:
--- a/local_operator/cli.py
+++ b/local_operator/cli.py
@@ -1764,11 +1764,12 @@ def main() -> int:
         # own env config and the session factory does the same lazily — a
         # dead local would only invite drift.
         base_dir = config_dir()
-        agent_home_dir = Path.home() / "local-operator-home"
-
-        # Create the agent home directory if it doesn't exist
-        if not agent_home_dir.exists():
-            agent_home_dir.mkdir(parents=True, exist_ok=True)
+        # The agent home is NO LONGER created here. Creating it unconditionally
+        # before dispatch meant `config list`, `login`, `--version` and every
+        # other non-session subcommand created a workspace directory they never
+        # touch, and it hardcoded ~/local-operator-home while ignoring any
+        # override. It is now created lazily by the paths that actually run a
+        # task (session/exec/serve start), through paths.ensure_agent_home_dir.
 
         if args.subcommand == "credential":
             if args.credential_command == "update":
--- a/local_operator/cli.py
+++ b/local_operator/cli.py
@@ -636,6 +636,7 @@ def credential_update_command(args: argparse.Namespace) -> int:
             paint(
                 f"Warning: '{args.key}' is not a known provider key.{hint} " "Storing it anyway.",
                 WARNING,
+                stream=sys.stderr,
             ),
             file=sys.stderr,
         )
@@ -652,7 +653,7 @@ def credential_update_command(args: argparse.Namespace) -> int:
         # Empty input or a closed stdin. Strip any control sequences from the
         # message before printing \u2014 the presenter owns the colour, and a nested
         # escape from deeper in the stack would otherwise repaint the line.
-        print(paint(strip_control_sequences(str(exc)), ERROR), file=sys.stderr)
+        print(paint(strip_control_sequences(str(exc)), ERROR, stream=sys.stderr), file=sys.stderr)
         return 1
     return 0
 
@@ -686,6 +687,7 @@ def config_open_command() -> int:
             paint(
                 "Error: Configuration file does not exist.  Create one with `config create`.",
                 ERROR,
+                stream=sys.stderr,
             ),
             file=sys.stderr,
         )
@@ -725,7 +727,7 @@ def config_open_command() -> int:
             gui_error = e
 
     print(
-        paint(f"Error opening configuration file: {gui_error}", ERROR),
+        paint(f"Error opening configuration file: {gui_error}", ERROR, stream=sys.stderr),
         file=sys.stderr,
     )
     print(
@@ -756,7 +758,9 @@ def config_edit_command(args: argparse.Namespace) -> int:
         close = difflib.get_close_matches(args.key, sorted(valid_keys), n=1)
         hint = f" Did you mean '{close[0]}'?" if close else ""
         print(
-            paint(f"Error: unknown configuration key: '{args.key}'.{hint}", ERROR),
+            paint(
+                f"Error: unknown configuration key: '{args.key}'.{hint}", ERROR, stream=sys.stderr
+            ),
             file=sys.stderr,
         )
         print(
@@ -798,7 +802,9 @@ def config_edit_command(args: argparse.Namespace) -> int:
         # 1, not -1: a shell sees -1 as 255, which collides with the
         # xargs/ssh "command not found" sentinel and contradicts the
         # documented 0/non-zero exec contract (item A13).
-        print(paint(f"Error updating configuration: {e}", ERROR), file=sys.stderr)
+        print(
+            paint(f"Error updating configuration: {e}", ERROR, stream=sys.stderr), file=sys.stderr
+        )
         return 1
 
 
@@ -1587,7 +1593,7 @@ def _preflight_hosting_model(
         # this into the exec route from reintroducing a stdout leak.
         from local_operator.cli_style import ERROR, paint
 
-        print(paint(f"Error: {exc}", ERROR), file=sys.stderr)
+        print(paint(f"Error: {exc}", ERROR, stream=sys.stderr), file=sys.stderr)
         return 1
     except Exception:  # noqa: BLE001 — unknown providers pass through
         return None
@@ -1613,6 +1619,7 @@ def _print_first_run_quickstart(credential_manager: CredentialManager) -> None:
             "Error: Local Operator is not configured yet \u2014 no hosting provider, "
             "model, or credential is set.",
             ERROR,
+            stream=sys.stderr,
         ),
         file=sys.stderr,
     )
@@ -1629,6 +1636,7 @@ def _print_first_run_quickstart(credential_manager: CredentialManager) -> None:
             "On an interactive terminal, just run `local-operator` and log in "
             "from the setup screen.",
             INFO,
+            stream=sys.stderr,
         ),
         file=sys.stderr,
     )
@@ -1712,6 +1720,7 @@ def _preflight_api_key(
                 f"'{hosting}'. Starting anyway — run `/login {canonical}` in the "
                 f"TUI, `local-operator login {canonical}` from a shell{env_hint}.",
                 WARNING,
+                stream=sys.stderr,
             ),
             file=sys.stderr,
         )
@@ -1726,6 +1735,7 @@ def _preflight_api_key(
             f"is not configured. Set it via `local-operator login {canonical}`"
             f"{credential_hint}, or the environment.",
             ERROR,
+            stream=sys.stderr,
         ),
         file=sys.stderr,
     )
@@ -2205,6 +2215,7 @@ def main() -> int:
                             "(missing 'local_operator.tui'); remove --tui to use the "
                             "plain REPL.",
                             ERROR,
+                            stream=sys.stderr,
                         ),
                         file=sys.stderr,
                     )
--- a/local_operator/cli.py
+++ b/local_operator/cli.py
@@ -1468,7 +1468,14 @@ async def _run_headless_repl(
     # own output IS console output; lifting the level keeps its prints while
     # dropping the library chatter. WARNING and above still surface — a genuine
     # problem the user needs to see is not INFO.
+    #
+    # The noisy HTTP-client loggers are raised EXPLICITLY, not just via the root:
+    # configure_cli_logging pins each of them to INFO by name, and a child logger
+    # with its own level ignores the root's — so raising only the root left
+    # httpx's per-request line leaking. Same list configure_cli_logging quietens.
     logging.getLogger().setLevel(logging.WARNING)
+    for _noisy in ("requests", "urllib3", "httpx", "httpcore"):
+        logging.getLogger(_noisy).setLevel(logging.WARNING)
 
     console = Console(stderr=True, highlight=False)
     session = await create_session(
--- a/local_operator/cli.py
+++ b/local_operator/cli.py
@@ -677,14 +677,27 @@ def config_create_command() -> int:
 
 def config_open_command() -> int:
     """Open the configuration file using the default system editor."""
+    from local_operator.cli_style import ERROR, paint
+
     config_path = config_dir() / "config.yml"
     if not config_path.exists():
         print(
-            "\n\033[1;31mError: Configuration file does not exist.  Create one with "
-            "`config create`.\033[0m"
+            paint(
+                "Error: Configuration file does not exist.  Create one with `config create`.",
+                ERROR,
+            ),
+            file=sys.stderr,
         )
         return 1
 
+    # Try the platform GUI opener first, then fall back to $VISUAL/$EDITOR. The
+    # GUI openers do not exist on a headless/SSH Linux box (there is no
+    # xdg-open without a desktop session), and there `config open` used to fail
+    # outright — yet that is exactly the environment where a terminal editor is
+    # the ONLY way in. Only spawn an interactive editor when stdout is a tty:
+    # an editor launched from a pipe or a non-interactive shell has no terminal
+    # to draw in and would hang or error.
+    gui_error: Exception | None = None
     try:
         if platform.system() == "Windows":
             subprocess.run(["start", str(config_path)], shell=True, check=True)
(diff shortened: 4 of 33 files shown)


damianvtran:
### Design review — round 1

Reviewer: designer (claude-opus-4-6 / Claude Code)
Scope: first-run setup state (keyless launch) — frames rendered from dev-first-run-ux HEAD 42a38641

**Verdict: NOT TERMINAL — 1 MAJOR (D1) to resolve. No blocker.** The core idea is right and executed with care: a keyless launch lands in the splash's own warning voice with a `setup` band instead of a red "session failed to start", the guidance names the exact next command, and the message degrades primary-first as the terminal narrows. Consecutive frames are identical in text and layout (only a single logo-mark pixel pulses between frame1/frame2 — no reflow, confirmed by SVG text diff). The remaining items are one real contradiction and a cluster of first-run-copy polish.

Frames viewed: `view-setup-frame1.png` (100×30), `view-login-typed.png`, `before-splash-frame1.svg` (pre-change, model band `test/model`), plus self-rendered `wide140` (140×30), `narrow70` (70×30), `narrow50` (50×30) via the real `OperatorApp` + `FakeProviderController` + `HostingNotConfiguredError` factory, stylesheet applied.

---

**D1 — MAJOR — Splash still says `connecting…` while the band says `setup`; the two disagree on the first screen a new user reads.**
In `view-setup-frame1.png` the status band reads `◆ setup`, but four rows up the splash's model row reads `connecting…`. For a keyless first run *nothing is connecting* — there is no session being awaited, the app is deliberately parked waiting on the user. A brand-new user reads "connecting…" as "it's busy, wait" — the exact opposite of the "you need to act" message the setup state exists to send, and it makes the screen look hung. This also violates the invariant the code documents for itself: `welcome.py:225-227` says the placeholder and the band "use the same word for the same state, so the two never disagree on screen" — here they demonstrably disagree (`setup` vs `connecting…`). Fix: in setup state drive the splash model row to the same word the band uses (`setup`), or a purposeful idle phrase like `not configured`, rather than leaving the `MODEL_PENDING` sentinel up. The band already knows how to say `setup`; the splash should say it too.

**D2 — MINOR — The CTA sentence is diagnosis-first, so the actionable command is what truncates away first.**
The notice reads `welcome — no provider configured yet. Run /login <provider> (e.g. /login openai) to set one up, or /provider to see the list.` It fits whole only at ~140 cols (`wide140`). At the default 100 cols (`view-setup-frame1.png`) it clips to `…to set one u…`, dropping the `/provider` alternative; at 70 cols (`narrow70`) it clips to `(e.…`, losing the `/login openai` example; at 50 cols (`narrow50`) it clips to `Run /…`, losing the command entirely. The sentence spends its first ~40 cells restating the diagnosis the band and the toast already carry ("no provider configured"), so on a tight terminal the budget is gone before the command the user actually needs. Fix: front-load the action — e.g. `Run /login <provider> to get started (e.g. /login openai) — no provider configured yet.` so the command survives the truncation pass and the diagnosis is what drops. (Keeping the full sentence in the transcript block and letting the *toast* carry the glance is the right split; the issue is only the word order inside the primary line.)

**D3 — MINOR — The affordance table never surfaces `/login`, the one command this state exists to teach.**
The hint table under the splash lists `/  command picker`, `/help  all commands`, `ctrl+d  quit` (`HINTS`, welcome.py:232). In the setup state the single most important affordance — `/login` — appears *only* inside the orange notice line, which is exactly the line that truncates first (D2). A first-run user who is scanning the fixed key/description table rather than reading the warning prose is guided to the command picker and to quit, but not to the action that unblocks them. Fix: in setup state, add a `/login  set up a provider` row to the affordance table (or promote it into that table for this state) so the next step survives regardless of width and lives where the eye lands for "what can I type".

**D4 — MINOR — The first rotating tip pitches `/resume` to a user who has nothing to resume.**
The tip row reads `· /resume picks up a recent session where you left off` in every setup frame. The tip pool pins its first entry to `/resume` on the argument that resumption is "the single question a returning user arrives with" (welcome.py:263-266) — but the setup state is the definitional *non-returning* user: there are no prior sessions, so the first and most-read tip advertises a command that does nothing for them. Fix: when `_setup_state` is set, pin the opening tip to something a first-run user can act on (e.g. the `/login` or `/provider` tip, or `/help`) rather than `/resume`.

**D5 — NIT — Toast headline truncates mid-word (`…ye…`) and repeats the "welcome —" the splash already carries.**
The top-right toast reads `welcome — no provider configured ye…` in every frame (35-cell hard cut, `_splash_toast_headline`, app.py:12169-12171). Cutting inside "yet" looks like a rendering glitch rather than an intentional ellipsis, and the `welcome —` prefix duplicates the splash line one glance away. Fix: cut on a word boundary and drop the greeting the splash already owns — e.g. `No provider configured` — which fits in 22 cells with room to spare and reads as a clean headline.

---

Non-issues confirmed (not filed):
- The red `!` at the setup band's right edge (visible in narrow70/narrow50/wide140) is the app-wide auto-approve/disarmed-gate alarm (`ICON_APPROVALS`, status_line.py:133), normal band vocabulary — not a setup artifact.
- The `provider controller unavailable — run: local-operator login` hint under a typed `/login openai` (`view-login-typed.png`) is a harness artifact of the render-only `FakeSession` with no live `ProviderController`, per the review brief; the copy itself is fine and I did not evaluate it as real behaviour.

Follow-ups (after D1): D2–D5 are all first-run-copy/placement polish and can ride the same remediation round or a fast follow. Re-capture the 100-col setup frame after D1 so the fix (band and splash agreeing) is on the record.

damianvtran:
### Agent review — round 1

Reviewer: reviewer (Claude Sonnet 4.5, via Local Operator)
Scope: 2f06901b14ec374da8792ab8c4c9863e8fcde686..42a386417da72ca482eeb2272dbbca08b29e653b

**Verdict: 2 MAJOR, 4 MINOR, 1 NIT.** No blockers — the headline setup-state gating is sound (tty + `run_tui is not None` is the only path that passes `allow_setup_state`, and `HostingNotConfiguredError`-as-`ValueError` keeps every existing catch site behaving). But the exit-code sweep (item 18) left its own stated motivating path unfixed, and the cross-platform hardening introduces a Windows regression in the very file that hardens for Windows. Both should be fixed before merge.

Gates I ran from the worktree (import verified → `/private/tmp/lop-first-run/local_operator/__init__.py`): flake8 clean, black clean, isort clean, pyright clean on the new modules; new suites `test_login_defaults / test_failover / test_cli_style / test_config / test_credentials / test_paths` → 228 passed, 3 skipped. The "pre-existing flake" is the full-suite pytest timeout, unrelated to this diff.

---

**MAJOR 1 — `exec --background` preflight still returns `-1` → exit 255, the exact regression item 18 claims to eliminate.** `local_operator/exec_mode.py:251,254` still `return -1`, and these propagate unchanged through `run_exec` (`exec_mode.py:342`) to `cli.py:2111` → `exit(main())`. Reproduced:
```
$ HOME=/tmp/fresh local-operator exec "echo hi" --background   # no hosting/model
Error: Hosting platform is not configured.     # + raw \033[1;31m on stderr
$ echo $?
255
```
This is worse than a missed sweep: the comment at `cli.py:2095-2096` asserts *"The byte-identical twins in exec_mode._spawn_background were fixed earlier and these were missed"* — but the twins were **not** fixed; they are the last two `return -1` left in the tree (`grep -rn 'return -1' local_operator/ --include=*.py`). Item 18's own rationale is the exec 0/non-zero contract, so leaving 255 on the exec path is the precise case it exists to close. Fix: `return 1` at `exec_mode.py:251,254` (and correct the false comment), or delete the comment's "were fixed earlier" clause and fix both.

**MAJOR 2 — `os.fchmod` breaks every credential write on Windows.** `local_operator/credentials.py:102` calls `os.fchmod(fd, _CREDENTIALS_MODE)` unconditionally. `os.fchmod` is Unix-only (it does not exist on Windows), so `write_to_file` — reached by `credential update` **and** by `login`/`_apply_login_defaults` — raises `AttributeError` on Windows. The pre-image (`git show 2f06901b:.../credentials.py:57`) used a plain `open(...,"w")` and was cross-platform, so this is a new regression, and it lands in the same file whose item-14 change explicitly hardens for "a legacy Windows code page." Internally inconsistent: the module claims Windows support while unconditionally calling a POSIX-only syscall on the write path. Fix: guard it, e.g. `if hasattr(os, "fchmod"): os.fchmod(fd, _CREDENTIALS_MODE)` — `tempfile.mkstemp` already creates the temp at 0600 on POSIX, and Windows POSIX-mode chmod is a no-op anyway, so the guard loses nothing.

---

**MINOR 1 — `paint()` gates on `sys.stdout` but almost every message it wraps prints to `sys.stderr`.** `cli_style.paint` defaults `stream=sys.stdout` (`cli_style.py:67,74`), yet the error/warn call sites pass `file=sys.stderr` (e.g. `cli.py:1690-1707`, `config.py:118`). With stdout a tty and stderr redirected (`2>log`) the SGR bytes still land in the log; with stdout piped and stderr a tty the error loses its colour. Gate on the stream you actually write to: `paint(..., stream=sys.stderr)` at the stderr sites.

**MINOR 2 — the NO_COLOR/tty sweep is partial; the PR description overstates it.** The summary says "NO_COLOR / TERM=dumb / non-tty respected," but 113 raw `\033[` literals remain in `cli.py` and the exec paths (`exec_mode.py:251,254`, `cli.py:2100`) still emit raw escapes. Reproduced — `NO_COLOR=1 local-operator config list` still prints `\033[1;32m`. Fine as a scoped first pass, but either narrow the claim to "the first-run/error paths" or extend `paint` to `config_list`/`agents`/`teams`/`mobile`.

**MINOR 3 — `config.yml.bad` backup clobbers a prior one.** `config.py:121` uses `self.config_file.replace(backup)`, which overwrites an existing `config.yml.bad`. Two bad edits in a row silently lose the first backup — the comment's "keeps the user's edits recoverable" holds only for the most recent. Consider a timestamped suffix, or skip if `.bad` already exists.

**MINOR 4 — no test pins the exec-path exit code.** `test_exec_mode.py` asserts `130` for one cancel case but nothing asserts `_spawn_background`/`run_exec` returns `1` (not `-1`/255) on unconfigured hosting — which is exactly why MAJOR 1 slipped through. A one-line `assert _spawn_background(...) == 1` would have caught it and guards the contract.

**NIT — duplicated no-default-model message.** `bootstrap.py:88-94` inlines the same "Model name is not configured … e.g. gpt-4o, claude-3-5-sonnet-latest, deepseek-chat" text that `session_factory._no_model_message` (`session_factory.py:209-222`) already centralises. Import and reuse `_no_model_message` so the two cannot drift.

---

Blockers: none. MAJOR 1 and MAJOR 2 are the two to fix before merge; the minors are cheap follow-ups. Not a terminal round — please address the majors and re-request.

damianvtran:

Reviewer round addressed: [Agent review — round 1](https://github.com/damianvtran/local-operator/pull/249#issuecomment-5381589699)
Head at remediation: `1818eb9e9da853ec21cb0c24a214359c464b1ad3` (was `42a38641`)

Both majors and all actionable minors/nit are fixed; the one non-code item (MINOR 2) is a PR-description accuracy correction, made in the description.

| Finding | Disposition | Evidence |
|---|---|---|
| **MAJOR 1** — `exec --background` preflight `return -1` → exit 255, defeating item 18; false "were fixed earlier" comment | **Fixed** | `f393b851` — both `exec_mode.py` preflight sites now `return 1`, the false `cli.py` comment corrected. `grep -rn 'return -1' local_operator/ --include=*.py` is now empty. New regression test pins the exec unconfigured-hosting path to `1` and asserts it never spawns (was the exact gap that let this through). |
| **MAJOR 2** — `os.fchmod` unconditional → `AttributeError` on Windows for every credential write | **Fixed** | `49054dcd` — guarded with `if hasattr(os, "fchmod")`. `tempfile.mkstemp` already yields 0600 on POSIX, and POSIX-mode chmod is a no-op on Windows, so the guard loses nothing. |
| **MINOR 1** — `paint()` gates on stdout but writes to stderr at the error/warn sites | **Fixed** | `1818eb9e` — `stream=sys.stderr` passed at all 15 stderr paint sites (`cli.py`, `config.py`), so NO_COLOR/tty gating matches the stream actually written. |
| **MINOR 2** — NO_COLOR sweep partial; description overstates coverage | **Fixed (description)** | No further escape sweep by instruction (scope control). The PR description's NO_COLOR claim has been narrowed to the first-run/error paths so it is accurate; the remaining `config_list`/`agents`/`teams` escapes are called out as out of scope for this PR. |
| **MINOR 3** — `config.yml.bad` backup clobbers a prior one | **Fixed** | `1818eb9e` — backup is now `config.yml.bad.<timestamp>`, so a second bad edit cannot silently lose the first recoverable copy. |
| **MINOR 4** — no test pins the exec-path exit code | **Fixed** | `f393b851` — added (see MAJOR 1). |
| **NIT** — duplicated no-default-model message | **Fixed** | `1818eb9e` — `bootstrap.py` imports and reuses `session_factory._no_model_message` so the two cannot drift. |

**Gates** (from the worktree, import verified into `/private/tmp/lop-first-run/...`): flake8 ✅ · black --check ✅ (427 files) · isort ✅ · pyright ✅ (0/0/0). Suites: `test_config` 74p/3s, `test_credentials`/`test_cli_new`/`test_cli_style`/`test_paths`/providers 698p, `test_exec_mode` ✅, `test_welcome` 48p, `test_app_pilot` 167p.

No blocker/major findings remain open on this round. Re-requesting an independent round-2 pass on the new head.

damianvtran:

Reviewer round addressed: [Design review — round 1](https://github.com/damianvtran/local-operator/pull/249#issuecomment-5381556260)
Head at remediation: `1818eb9e9da853ec21cb0c24a214359c464b1ad3` (was `42a38641`) — all in commit `e118a0e3`

Re-captured 100-col frames from the real `OperatorApp` + `HostingNotConfiguredError` (stylesheet applied) are attached; consecutive frames are text-identical (35/35 lines) → no reflow.

| Finding | Disposition | What changed |
|---|---|---|
| **D1 — MAJOR** — splash reads `connecting…` while band reads `setup`; screen looks hung, violates the welcome.py:225-227 same-word invariant | **Fixed** | A `setup` flag now threads through `WelcomeInfo`; in setup state the splash model row renders `setup`, matching the band. The re-captured frame shows both reading `setup` — the invariant holds again. |
| **D2 — MINOR** — CTA is diagnosis-first, so the `/login` command truncates away first on narrow terminals | **Fixed** | Notice front-loads the action: `Run /login <provider> to get started (e.g. /login openai) — no provider configured yet.` The command now survives the truncation pass; the diagnosis is what drops. |
| **D3 — MINOR** — affordance table never surfaces `/login` | **Fixed** | `/login  set up a provider` is now the leading row of a setup-state affordance table (`HINTS_SETUP`), so the key next step lives where the eye scans for "what can I type," regardless of width. |
| **D4 — MINOR** — first rotating tip pitches `/resume` to a user with nothing to resume | **Fixed** | In setup state the opening tip is pinned to `TIP_SETUP` (`/login <provider> sets up a provider (e.g. /login openai)`) instead of `/resume`. |
| **D5 — NIT** — toast cuts mid-word (`…ye…`) and repeats "welcome —" | **Fixed** | Setup state gets a purpose-built toast headline `No provider configured` — word-boundary clean, no redundant greeting. |

Non-issues you flagged (provider-controller-unavailable harness artifact, the red `!` gate-alarm) were confirmed harness/normal, not touched.

Both code round-1 and this design round-1 are clean on the new head; re-requesting a design round-2 confirmation on the re-captured frames before merge.

_(Frames on file: `after-r1-setup-frame1/2.svg` + viewed PNGs — band + splash both read `setup`, action-first notice, `/login` table row, `/login` tip, clean `No provider configured` toast.)_

damianvtran:
### Agent review — round 2

Reviewer: reviewer (Claude Sonnet 4.5, via Local Operator)
Scope: 42a386417da72ca482eeb2272dbbca08b29e653b..1818eb9e9da853ec21cb0c24a214359c464b1ad3

**Verdict: CLEAN — 0 blockers, 0 majors. This round is TERMINAL.** Both round-1 majors are genuinely fixed, all four actionable minors + the nit are correctly addressed, and the remediation delta (`git diff 42a38641..1818eb9e`) introduces no regressions. Import verified into the worktree (`/private/tmp/lop-first-run/local_operator`).

**Round-1 MAJOR 1 — exec preflight `-1`→255 — FIXED.** `exec_mode.py:254,257` now `return 1` (both preflight twins), and `grep -rn 'return -1' local_operator/ --include=*.py` is empty. The false "were fixed earlier" comment at `cli.py` is corrected to describe the contract truthfully. New test `test_spawn_background_unconfigured_hosting_returns_one` (`test_exec_mode.py:638`) asserts `code == 1` AND `popen_mock.assert_not_called()` — pins both the exit code and the never-spawn-blind invariant that let this slip. Passes.

**Round-1 MAJOR 2 — `os.fchmod` Windows crash — FIXED.** `credentials.py:107` now guards `if hasattr(os, "fchmod")`. The POSIX write path still yields 0600: `tempfile.mkstemp` (`:100`) creates the temp at 0600, `fchmod` re-asserts it inside the guard, and `_ensure_config_exists` still `chmod(_CREDENTIALS_MODE)`s on create + re-tightens a loose pre-existing file (`:140,148`). Guard loses no hardening on POSIX; on Windows the POSIX mode is a no-op anyway.

**Minors — all verified:**
- MINOR 1 (stderr stream gating): every error/warn paint site that writes to stderr now passes `stream=sys.stderr` — `config.py:268,283,289`, and 8 sites across `cli.py` (`credential_update`, `config_open`, `config_edit`, `_preflight_hosting_model`, `_print_first_run_quickstart`, `_preflight_api_key`, tui-missing). Gating now matches the stream actually written.
- MINOR 3 (`.bad` backup clobber): `config.py:288` now uses `.bad.<timestamp>` (`%Y%m%d-%H%M%S`); `datetime` is imported (`config.py:10`). Tests updated to glob `config.yml.bad.*` (`test_config.py:320,335`). A second bad edit no longer loses the first.
- MINOR 4 (exec exit-code test): the new test asserts `== 1` explicitly with a `# not -1 / 255` comment — closes the exact gap.
- NIT (duplicated no-model message): `bootstrap.py:99` now lazily imports and reuses `session_factory._no_model_message(hosting)`; the inlined literal is gone. Signature matches (`_no_model_message(hosting: str) -> str`), lazy import avoids the circular/stack cost. pyright clean.

**Setup-flag threading (new surface) — invariants hold.** The `setup: bool` field threads `WelcomeInfo → session_welcome_info → build_welcome_lines → _status_rows/_hint_lines/_tip_lines`, sourced from `app._setup_state` at the one call site (`app.py:1453`). Confirmed at runtime:
- Model row: setup renders `setup` (not `connecting…`), non-setup still renders `connecting…` — the welcome.py:225 "same word for the same state" invariant restored (the D1 fix), and the pre-existing `test_the_splash_names_the_model_the_way_the_band_does` D10 invariant still passes.
- `HINTS_SETUP` leads with `/login`; `_hint_lines(setup=True)` swaps the table across all three width tiers, so the affordance survives truncation.
- `TIP_SETUP` replaces only the opening (`index % len(TIPS) == 0`) frame; `TIP_MIN_WIDTH` is re-derived over `(*TIPS, TIP_SETUP)` so the widest-entry threshold invariant holds by construction (test updated to match).

**Gates (from the worktree):** flake8 ✅ · black --check ✅ (213 files) · isort ✅ · pyright ✅ (0/0/0 on touched modules). Suites `test_exec_mode + test_credentials + test_config + test_welcome`: **122 passed, 3 skipped**.

No blocker or major findings remain. Round is clean and terminal — remaining design-round items (if any) are tracked separately in the design review thread.

damianvtran:
### Design review — round 2

Reviewer: designer (Claude Opus 4.6 / Local Operator)
Scope: first-run setup state — frames from dev-first-run-ux HEAD 1818eb9e

**Verdict: TERMINAL — 0 BLOCKER, 0 MAJOR.** Every round-1 finding is genuinely resolved on the rendered frame, the fixes read well together, and no new spacing/alignment/legibility issue was introduced. Approving for the design gate.

Frames viewed: `after-r1-setup-frame1/2.png` (100×30, md5-identical → no reflow), plus self-rendered `r2-100x30 / r2-70x30 / r2-50x30` via the real `OperatorApp` + `HostingNotConfiguredError` factory + `FakeProviderController` (stylesheet applied, `_setup_state=True`, band `'setup'` asserted in the capture) from `/tmp/lop-first-run`, to re-confirm the narrow-terminal behaviour that motivated D2 on the *current* head rather than the pre-remediation SVGs.

Per-finding confirmation:

- **D1 (was MAJOR) — resolved.** The splash model row now reads `setup`, matching the `◆ setup` band four rows below. No `connecting…` sentinel anywhere on the setup frame; the welcome.py same-word invariant holds again. The screen no longer reads as "busy, wait."
- **D2 (was MINOR) — resolved, verified across widths.** The notice front-loads the action: `! Run /login <provider> to get started (e.g. /login openai) — no provider configured yet, or /p…`. At 70 cols it truncates to `…(e.g. /login openai) — no …` (full command **and** example survive; the diagnosis tail is what drops); at 50 cols it truncates to `Run /login <provider> to get started (e.g. …` (command still intact). The actionable command now survives the truncation pass at every width I rendered.
- **D3 (was MINOR) — resolved.** `/login   set up a provider` is the leading row of the affordance table, above `/`, `/help`, `ctrl+d`. Column alignment is clean and width-independent (present unchanged at 50/70/100).
- **D4 (was MINOR) — resolved.** The pinned opening tip reads `· /login <provider> sets up a provider (e.g. /login openai)` — teaches the one command this state exists to teach, not `/resume`.
- **D5 (was NIT) — resolved.** The toast reads a clean word-boundary headline `No provider configured` — no mid-word cut, no redundant `welcome —` greeting.

No reflow: `after-r1-setup-frame1.png` and `after-r1-setup-frame2.png` are byte-identical (md5 `6a9139…`); the round-1 single logo-pixel pulse is gone from this pair.

What works well together: the band, splash row, and status footer now all speak the single word `setup`; the notice, the affordance table, and the rotating tip all lead with `/login`, so wherever a first-run user's eye lands — glance toast, warning line, "what can I type" table, or the tip — the same next step is waiting. The orange `!` warning voice is consistent between the notice and the footer `!` glyph, and the diagnosis lives in the toast/tail where truncation can safely eat it.

No open findings. Round is **TERMINAL**; no follow-ups.

damianvtran:
### CI fix — test-only, post round-2

Head advanced `1818eb9e` → `58566077` with a single **test-only** commit (`58566077`). No production code changed.

**What CI caught that local runs didn't:** `test (3.13)` failed on `test_prompt_for_credential` and `test_missing_credential_raises_error`. Root cause: item 13 added the non-tty stdin branch to `prompt_for_credential` (scripted-automation contract). CI runs pytest with **stdin not a tty**, so these two *legacy* tests — which mock only `getpass` — took the pipe path and raised `EOFError` instead of exercising the mocked interactive prompt. Local runs passed because the dev terminal had a tty.

**Reproduced and fixed:** forcing `< /dev/null` (CI's non-tty condition) reproduces both failures with `EOFError` on the pre-fix tree; the fix pins `sys.stdin.isatty` → True in exactly the two tests that assert the interactive branch. The non-tty pipe paths remain covered by `test_prompt_reads_from_piped_stdin_without_getpass` and `test_prompt_empty_piped_stdin_raises_eof`. Full `test_credentials.py` under `< /dev/null`: 14 passed, 3 skipped; flake8/black/isort clean on the file.

Requesting a fresh **round-3** code confirmation on `58566077` so the gate is fresh before any merge. Design rounds are unaffected (no UI change).

damianvtran:
### Agent review — round 3

Reviewer: reviewer (Claude Sonnet 4.5, via Local Operator)
Scope: 1818eb9e9da853ec21cb0c24a214359c464b1ad3..58566077c6ca8ee37f82e9458cc059d72f5ec7f3 (test-only CI fix)

**Verdict: CLEAN — 0 blockers, 0 majors. This round is TERMINAL.**

Narrow confirmation of the CI-fix delta only. Production code is UNCHANGED since round-2 (`git diff --name-only 1818eb9e..58566077` → `tests/unit/test_credentials.py` only), so round-2's CLEAN/TERMINAL verdict still stands for all production code.

**What the fix does — correct and complete:**
1. **Test-only.** The entire diff is two added `monkeypatch.setattr("sys.stdin.isatty", lambda: True)` lines (plus explanatory comments) in `test_prompt_for_credential` and `test_missing_credential_raises_error`. No production file touched.
2. **Pins the interactive branch, does not weaken assertions.** Both tests still assert exactly what they did before — `test_prompt_for_credential` asserts the getpass-returned key is saved (`NEW_API_KEY=new_test_key`, mode `0o600`); `test_missing_credential_raises_error` still asserts `ValueError` with "is required for this step". Forcing `isatty()→True` makes the interactive tty path run so the `getpass.getpass` mock is genuinely exercised, rather than silently falling through the non-tty pipe branch (which raises `EOFError` under pytest's redirected stdin). This is the right fix — it restores the condition the tests were written for, not a papering-over.
3. **Non-tty pipe paths remain covered.** `test_prompt_reads_from_piped_stdin_without_getpass` (asserts one-line read, getpass NOT called) and `test_prompt_empty_piped_stdin_raises_eof` (empty pipe → `EOFError`) are untouched and still exercise the `isatty()→False` branch.

**Reproduced the CI condition:**
- Post-fix, non-tty stdin: `python -m pytest tests/unit/test_credentials.py -q < /dev/null` → **14 passed, 3 skipped**.
- Pre-fix (1818eb9e) under the same `< /dev/null`: **2 failed, 12 passed, 3 skipped** — both failures are `EOFError: ... is required for this step` on exactly the two tests this commit pins. The fix resolves precisely the observed CI failure.

**Lint:** flake8 clean, black clean, isort clean on `tests/unit/test_credentials.py`.

No blocker or major findings. Round is clean and terminal.

damianvtran:
### Agent review — round 4

Scope: origin/main...c90e8fc0218744d784dc563fee719c642b24ae40 (post-rebase integration + 0.22.0 bump)

**Verdict: CLEAN** — no blockers, no majors. This round is TERMINAL.

Confirmed the branch is correctly re-based: `git merge-base --is-ancestor origin/main HEAD` passes, and `origin/main..HEAD` is exactly the branch's own 19 commits (ending in `chore(release): bump version to 0.22.0`). Net diff reviewed against current `origin/main`.

**Integration of the four auto-merged overlap files is semantically sound:**

- **`local_operator/tui/app.py`** — The first-run additions (`_setup_state`, `_on_config_changed`, `_enter_setup_state`, `_apply_login_defaults`, the `HostingNotConfiguredError` passthrough in `_on_boot_failed`, the `setup=` arg to `session_welcome_info`, and `append_auth_recovery`) all land cleanly. The **only** net-diff deletions are the branch's own replaced lines (the old `session_welcome_info` call, the stale "nothing else needs poking" comment, the bare `NoticeBlock(message.error)`); none of main's newly-landed TUI code is clobbered — it is the diff base and survives as unchanged context. No duplicated or lost handlers.
- **`local_operator/session_factory.py`** — `HostingNotConfiguredError(ValueError)` subclass is correct: existing `except ValueError` sites keep working, and the two setup-state callers can classify the one condition. `resolve_hosting_model` now resolves a known-default model instead of raising, and only a truly-unknown hosting raises (with a message naming current model ids). Lazy `ensure_agent_home_dir()` in `create_session` is best-effort (`except OSError: pass`) — non-fatal, correct.
- **`local_operator/model/configure.py`** — `DEFAULT_MODEL_NAMES` is re-exported byte-for-byte from the new stdlib-only `model.defaults`; the legacy `configure.DEFAULT_MODEL_NAMES` import path is preserved. One map, one home.
- **`tests/unit/test_session_factory.py`** — Updated assertions match the new behaviour (legacy message still raised as the subclass; `openai` with no model resolves to `gpt-4o`; unknown hosting still raises "no default is known").

**Preflight ordering (the merge's subtlest risk) is correct:** `use_tui` is now decided *before* `_preflight_hosting_model` because the preflight needs it — `setup_state_ok = bool(use_tui and run_tui is not None)` gates `allow_setup_state`, so only the real full-screen TUI path may open in setup state; headless REPL, exec, `--no-tui`, and piped-stdout paths all keep fail-fast with the complete quickstart message. `on_config_changed=config_manager.reload` reconciles the launch-time `ConfigManager` (closed over by the session factory) with what the app's first-run `/login` writes to disk, so the post-login rebuild resolves the freshly-written hosting instead of bouncing back into setup. `exec_mode._spawn_background` correctly returns `1` not `-1`.

**Gates & suites (project `.venv`, `import local_operator` resolves into `/private/tmp/lop-first-run/local_operator`):**

- flake8 `local_operator/ tests/` — clean (rc 0)
- black `--check` — 443 files unchanged
- isort `--check-only` — clean (rc 0)
- pyright (six overlap/new files) — 0 errors, 0 warnings
- `pytest test_session_factory test_welcome test_credentials test_config providers` — **771 passed, 3 skipped**
- `pytest tests/unit/tui/test_app_pilot.py` — **167 passed** (incl. new `test_first_run_setup_state_when_hosting_unconfigured`, which drives a `HostingNotConfiguredError` boot through the real pilot and asserts the `setup` band + `/login` splash notice rather than a red failure — the exact test that would expose a bad `app.py` merge)

Version is `0.22.0`, a real minor above main's `0.21.0`. The rebase auto-merge produced a correct integration; both feature sets coexist. Merge-ready on the code axis.

PR #247, opened 2026-08-22
feat(analytics): aggregated token-consumption analytics + /analytics /usage

damianvtran:
## Why

There was no way to see where tokens actually go. `/usage` shows provider *quota*, and `/context` shows the *next* request's estimated breakdown, but nothing aggregated real consumption across turns and sessions — so "where is our spend going, and is it the system prompt, the tool schemas, the conversation, or thinking?" was unanswerable, and there was no data to drive harness improvements over time.

This adds a diagnostic, aggregated token-consumption ledger that **every** provider call across **every** session contributes to, plus an esc-quittable `/analytics /usage` screen. Version bump is **MINOR** (0.20.1 → 0.21.0): new capability, no breaking changes.

### The analytics module (`local_operator/analytics/`)

- **`store.py`** — an append-only SQLite ledger at `<config_dir>/analytics.db` (WAL, `synchronous=NORMAL`, `busy_timeout`, 0600, same discipline as `auth.db`/`usage_cache.db`). Aggregation is a `GROUP BY` run only when a report opens, not a running counter, so writes never contend on a shared total. A rolling 90-day retention keeps the file bounded. Reads use a **fresh short-lived connection** so a report always sees the newest commit rather than a stale WAL snapshot.
- **`recorder.py`** — a process-wide singleton with **one** background daemon writer thread behind a bounded queue. `record()` is a single non-blocking `put_nowait` that **drops** on a full queue (counted, logged once) rather than applying back-pressure to a session. Call samples *and* session-name upserts funnel through the one writer/connection — two threads opening their first connection to a freshly-created SQLite DB race in a way that can leave the writer unable to see its own commits.
- **`model.py`** — the component taxonomy and attribution. Provider counts (input/output/cache/reasoning) are **authoritative**; the split across system prompt, custom instructions (agent/team profiles), tool inventory, tool schemas, environment, knowledge, conversation, and tool results is an **estimate**: each component's char-length share of the authoritative context total, apportioned with largest-remainder rounding so it sums exactly. Labelled as an estimate everywhere it surfaces.

### Universal, latency-free injection

[...]

PR #250, opened 2026-08-22
feat(tui): run slash commands inline, mid-draft, splicing the token out

damianvtran:
## Summary of Changes

Feature. The slash-command picker only opened when the `/` was the **first character of the first non-blank line**. A user who had started typing a message and then remembered they should route it to a team or an agent — appending `/team` at the caret — got no menu, because the parse was anchored to the buffer start rather than to where the caret actually was.

This makes slash detection **inline and caret-anchored**, and adds a run path that keeps the user's draft safe.

Key design points:

- **Caret-anchored, boundary-gated detection.** `slash_context` / `slash_argument` now key on the caret's whole-buffer offset and open the list for a boundary `/` token **anywhere** in the draft — at the buffer start, right after whitespace, or on its own line above a multi-line draft. A `/` glued to a preceding non-space character (`src/foo`, `and/or`) never triggers; that boundary rule — the same one a shell or editor command palette uses — is what makes running the parse on every keystroke of ordinary prose safe. To engage mid-text the slash must be preceded by a space (or be at the start) and the word terminated by a space or accepted from the picker, so it never fires while the user is still mid-token.

- **Prompt commands reassemble to the front; others splice-and-run.** Free-text commands whose argument is a message the model is given (`/goal`, `/loop`, `/btw`, `/team`, `/agent` — flagged `consumes_prompt` on the registry) are "start of the composer" commands by nature. Engaging one INLINE does **not** run it against the trailing draft — that guesses which words are a name vs the message, and silently ate the request when `/team` read `and then ship it` as a team name (**design review D1, fixed**). Instead it **reassembles** the composer to `/command [name] <the rest of the draft>` and **stages** it: nothing is submitted, and the user reviews the assembled prompt and presses Enter. `/team`/`/agent` reassemble when the name is picked from the autofill (the draft becomes the request); `/goal`/`/loop`/`/btw` reassemble on word completion (the draft becomes the argument). A non-prompt command (a selector or a listing, `/usage`) instead splices just its `/command` token out and runs it via a new `InlineCommandRequested` message, leaving the surrounding draft; a whole-buffer command still goes through the ordinary submit path.

- **Nested slashes are plain text.** Once a recognised command owns its line (`/team a improve the /team command`), a second occurrence of a slash command inside its argument is plain argument text — no picker, no highlight, no re-engagement. The tokenizer is command-aware: a terminated recognised command claims the rest of its line.

- **Type-through and Esc dismiss, both keeping the text.** Terminating the word with a space, or typing on into a word that matches nothing, closes the list without touching the buffer; Esc's per-word "not now" dismissal works inline too. The text is always left exactly as typed.

- **Pairs with the existing mid-session attach.** Reassembling and submitting `/team <name> <request>` (or `/agent <name> <message>`) mid-trajectory stamps the team roster / agent persona onto the session's **volatile prompt tail** (`goal_state.team_brief` / `agent_brief`), which `session_factory`'s prompt provider re-reads on **every** turn — so the message runs under the team manager or that agent. This half already worked; the inline gesture is what makes it reachable without abandoning a half-typed message.

Change class: **C1 (standard, pre-authorized)** — a scoped, additive TUI feature with a clear rollback (revert the branch's commits), no contract, schema, auth, or migration surface.

## Related Issue(s)

- Related: none filed (requested directly).

## Impact

- No breaking changes. Start-of-line commands behave exactly as before; the new behaviour is purely additive for a `/` that is not the first character.
- New registry field `SlashCommand.consumes_prompt`; new public helpers on `command_picker` (`slash_word`, `slash_token_span`, `slash_argument_context`, `SlashArgument`); new `InlineCommandRequested` editor message; `SlashContext`/`SlashArgument` carry span offsets; the parsers take an optional recognised-command set for the nested-slash rule.
- No dependency updates. No security implications — the parser only classifies composer text; dispatch and the attach paths are unchanged.

## Testing Details

Proof the feature works, exercised through the **real editor and app** (key presses, not internal calls):

**Inline detection, reassembly & run (`tests/unit/tui/test_command_picker.py`)** — new tests drive real keystrokes:
- `test_inline_command_opens_the_picker_mid_draft` — `fix this ` then `/te` opens the list, buffer intact.
- `test_a_glued_slash_never_opens_the_picker` — `src/foo` stays closed.
- `test_a_prompt_command_reassembles_to_the_front_staged_not_run` — `land the oauth fix /goal` + Enter → `/goal land the oauth fix`, staged (not run, not submitted).
- `test_inline_command_runs_and_is_spliced_out_keeping_the_draft` — a non-prompt `/usage` splices out and runs, `submissions == []`, buffer left as `fix the bug`.
- `test_inline_command_on_its_own_line_keeps_the_message_below`, `test_a_whole_buffer_command_still_submits_not_inline`.
- `test_type_through_dismiss_leaves_the_text_when_the_word_stops_matching`, `test_esc_dismiss_leaves_an_inline_command_in_the_text`.
- `test_a_nested_slash_inside_an_engaged_command_is_plain_text`, `test_a_crlf_buffer_does_not_leak_a_carriage_return_into_the_word`, plus caret-anchored trigger/non-trigger parametrizations.

**End-to-end mid-trajectory attach (`tests/unit/tui/test_slash_echo.py`)** —
- `test_inline_team_reassembles_to_the_front_and_then_sends_under_the_manager`: `ship the dashboard /team` → pick `feature-release` → reassembles to `/team feature-release ship the dashboard`, **staged** (nothing attached/sent); Enter then attaches the team and sends `ship the dashboard` as the request under the manager.
- `test_a_second_inline_command_of_the_same_kind_is_plain_argument_text`: `/team alpha improve the /team command` keeps the nested `/team` as plain text and submits it verbatim.
- `PROMPT_POLICY` pin mirrors `ECHO_POLICY` so every registered command states whether its argument is a prompt.

**Live UI validation** (real `OperatorApp` rendered to SVG, viewed as a frame):

- **Inline picker** — composer holding `fix the auth bug /team` with the caret after `/team` shows the `/team /teams` row. The old rule would have kept the list hidden because the `/` is not the first character.

  Inline slash picker mid-draft

- **After reassembly** — typing `review the auth flow /team` and picking `security` reassembles the composer to `/team security review the auth flow`, staged for the user to submit: the draft is preserved verbatim as the request, nothing was consumed as a name.

  After reassembling the inline prompt command

**Regression fixed by this PR:** `Editor.begin_model_query` set the buffer but never re-synced at the moved caret, so with caret-anchored detection the model list would not populate on a dispatched `/model`. Fixed to re-derive at the caret; covered by the existing model-default tests.

**Gates** (PR head `b75213b2`):
- `flake8` / `black --check` / `isort --check-only` / `pyright` — clean on all changed files.
- `pytest tests/unit/tui` — **2229 passed, 4 skipped**.

## Checklist

- [x] I have updated the documentation when required (in-code why/constraint comments added; no user doc surface).
- [x] Security considerations are addressed (parser only classifies composer text; dispatch/attach unchanged).

### DIFF
--- a/local_operator/tui/app.py
+++ b/local_operator/tui/app.py
@@ -145,6 +145,7 @@ from local_operator.tui.widgets.editor import (
     EditorCopyStale,
     EditorQuit,
     EditorSubmitted,
+    InlineCommandRequested,
     InterruptRequested,
     ModelQueryOpened,
     RefreshArgumentChoices,
@@ -441,7 +442,7 @@ SLASH_COMMANDS: list[SlashCommand] = [
     # than being paraphrased into a system notice. `_cmd_goal` writes that row
     # itself, only on the branch that actually stored something — the flag is
     # the permission, not the trigger.
-    SlashCommand("goal", "Show, set, or clear the session goal", echo=True),
+    SlashCommand("goal", "Show, set, or clear the session goal", echo=True, consumes_prompt=True),
     # Not an exception: LOOP_PROMPT is app-authored, not the user's words, and
     # `_loop_worker` already labels every iteration it starts (`· loop 1/3`), so
     # no agent output here is left unattributed. `echo=False` suppresses the
@@ -450,7 +451,7 @@ SLASH_COMMANDS: list[SlashCommand] = [
     # session's user MessageStartEvent is consumed silently rather than
     # painted — two different receipts for two different events (the typed
     # command, and the prompt the turn later announces).
-    SlashCommand("loop", "Iterate autonomously toward the goal"),
+    SlashCommand("loop", "Iterate autonomously toward the goal", consumes_prompt=True),
     # NOT an exception, and the reason IS the feature. The question does reach
     # the model, but only for one off-the-record request that never joins the
     # conversation (`SessionProtocol.complete_aside`) — so a user row in the
@@ -458,7 +459,7 @@ SLASH_COMMANDS: list[SlashCommand] = [
     # still be sitting there after Esc claimed to have thrown the exchange
     # away. The card is the receipt; `^f` inside it is how an exchange gets a
     # row, as a real turn rather than an echo.
-    SlashCommand("btw", "Ask a side question off the record (esc closes it)"),
+    SlashCommand("btw", "Ask a side question off the record (esc closes it)", consumes_prompt=True),
     # NOT an echo, and the receipt is the reason. The pass narrates itself
     # through the same `compacting context…` / `context compacted · 128.4k →
     # 21.9k tokens` notices the automatic one emits, and a refusal says why it
@@ -514,6 +515,10 @@ SLASH_COMMANDS: list[SlashCommand] = [
         "List teams, or send a request to a team's manager",
         aliases=("teams",),
         arguments=ArgumentMode.OPTIONAL,
+        # The request AFTER the team name is a prompt the manager is given, so an
+        # inline `/team` reassembles to the front (name from the autofill, the
+        # draft as the request) rather than eating the draft as the name.
+        consumes_prompt=True,
     ),
     # Same echo reasoning as `/team`, which this command mirrors surface for
     # surface: bare `/agent` is a listing (the listing is the receipt), a
@@ -529,6 +534,9 @@ SLASH_COMMANDS: list[SlashCommand] = [
         "List agents, or speak to this session as one",
         aliases=("agents",),
         arguments=ArgumentMode.OPTIONAL,
+        # The message AFTER the agent name is a prompt the persona is given, so
+        # an inline `/agent` reassembles to the front like `/team`.
+        consumes_prompt=True,
     ),
 ]
 
@@ -3902,6 +3910,37 @@ class OperatorApp(App[None]):
             self._run_slash_command(text)
             return
         self._submit_prompt(text, images, message.attachments)
+
+    def on_inline_command_requested(self, message: InlineCommandRequested) -> None:
+        """Run a slash command spliced out of the MIDDLE of a draft.
+
+        The editor has already removed the ``/command`` token from the buffer and
+        kept the surrounding message; all that remains is to DISPATCH the command,
+        which is the same synchronous slash handling a start-of-line command gets
+        — minus the prompt submission, because there is no prompt here: the draft
+        the user is still composing stays in the composer.
+
+        The aside owns the composer while it is up, and an inline command can only
+        be produced by the main composer (the aside's own key path never reaches
+        the slash picker), so no aside guard is needed here — but the session-gate
+        and unknown-command reporting inside ``_run_slash_command`` still apply
+        exactly as they do for a typed command.
+
+        Attaching a team or agent this way is the whole point of the mid-draft
+        gesture: ``/team ops`` stamps the roster onto this session's volatile
+        prompt tail, so the message the user then sends with Enter runs under the
+        team manager. See ``_cmd_team`` / ``_cmd_agent`` and
+        ``session_factory``'s per-turn prompt provider, which re-reads that tail.
+        """
+        command_text = message.command_text.strip()
+        if not command_text.startswith("/"):
+            return
+        if self._aside_is_open():
+            # Defensive only — see the docstring. Route to the aside rather than
+            # silently dispatching a command the user cannot see they issued.
+            self._ask_aside(command_text)
+            return
+        self._run_slash_command(command_text)
 
     def on_editor_quit(self, message: EditorQuit) -> None:
         self.exit()
@@ -10634,7 +10673,12 @@ class OperatorApp(App[None]):
         from local_operator.mcp.auth import mcp_logged_out_servers, oauth_server_names
         from local_operator.tui.widgets.command_picker import slash_argument
 
-        argument = slash_argument(editor.text, editor._argument_commands)
+        argument = slash_argument(
+            editor.text,
+            editor._argument_commands,
+            editor._caret_offset(),
+            editor._command_names,
+        )
         if argument is None:
             return []
         sub, _space, _rest = argument.partition(" ")
--- a/local_operator/tui/autocomplete.py
+++ b/local_operator/tui/autocomplete.py
@@ -105,6 +105,23 @@ class SlashCommand:
     #: that changes nothing. Free typing is unaffected in every mode — the list
     #: ranks what is typed, it never filters what may be submitted.
     arguments: ArgumentMode = field(default=ArgumentMode.NONE, kw_only=True)
+    #: Whether this command's argument is FREE-TEXT that becomes a prompt the
+    #: model is given — the goal text, a loop instruction, a team/agent request,
+    #: an aside question. These are the "start of the composer" commands: their
+    #: whole point is to prefix a message, so engaging one INLINE (mid-draft)
+    #: does not splice-and-run — it REASSEMBLES the command to the front of the
+    #: composer with the rest of the draft as its argument, staged for the user
+    #: to review and submit. That is what makes "type the message, then remember
+    #: to route it" safe: nothing the user typed is ever consumed as a name or
+    #: silently dropped (the D1 data-loss the naive end-of-line argument caused).
+    #:
+    #: Keyword-only, defaulting to FALSE for the same reason ``echo`` and
+    #: ``arguments`` do: a command that has not opted in keeps the simple
+    #: splice-and-run inline behaviour (``/usage``, ``/model``), which is right
+    #: for every command whose argument is a SELECTOR rather than a message.
+    #: ``SLASH_COMMANDS`` is pinned entry-by-entry in ``test_slash_echo.py`` so a
+    #: new command must state this choice.
+    consumes_prompt: bool = field(default=False, kw_only=True)
 
     @property
     def names(self) -> tuple[str, ...]:
--- a/local_operator/tui/widgets/command_picker.py
+++ b/local_operator/tui/widgets/command_picker.py
@@ -179,44 +179,211 @@ class _RowStyles(NamedTuple):
 
 
 class SlashContext(NamedTuple):
-    """Where the command word starts, and the word typed so far.
+    """Where the active command word sits, and the word typed so far.
 
-    ``start`` indexes the ``/`` itself so a completion can rebuild the buffer
-    without discarding whatever whitespace the user typed in front of it.
+    ``start`` indexes the ``/`` itself and ``end`` the first cell past the word,
+    so a completion can rebuild JUST that span and leave the rest of the draft
+    untouched. Before inline detection the word always ran to the end of the
+    buffer, so a completion could splice from ``start`` to the end; now the word
+    can have a message typed after it (``fix this /team``, or ``/team\\nfix
+    this``), and only ``[start, end)`` is the command — everything outside it is
+    the user's prose and must survive the completion verbatim.
     """
 
     start: int
     query: str
+    end: int
 
 
-def slash_context(text: str) -> SlashContext | None:
-    """The command word being typed, or ``None`` when the picker must hide.
+#: A slash opens a command only at a WORD BOUNDARY: the buffer start, or right
+#: after whitespace. This is what keeps ``src/foo`` and ``and/or`` from opening
+#: the picker — the ``/`` there is glued to a preceding non-space character, so
+#: it is punctuation inside a word, not the start of a command. The rule is the
+#: same one a shell or an editor command palette uses to tell a path apart from
+#: a command, and it is the ONE thing that makes inline detection safe to run on
+#: every keystroke of ordinary prose.
+def _is_boundary(line: str, index: int) -> bool:
+    """Whether ``line[index]`` (a ``/``) begins a fresh token."""
+    return index == 0 or line[index - 1].isspace()
 
-    The picker is for choosing a command, so it shows only while the command
-    WORD is still open: ``/`` first on the first non-blank line, and no
-    whitespace yet terminating the token. Once the user types
-    ``/model `` they have chosen ``model`` and are typing its argument — a
-    list of commands there is stale advice covering the transcript.
+
+def _line_of_cursor(text: str, cursor: int | None) -> tuple[str, int, int]:
+    """The line the cursor sits on, as ``(line, line_start_offset, column)``.
+
+    ``cursor`` is a whole-buffer offset; ``None`` means "the end of the buffer",
+    which is where a user typing at the end of their draft is. Clamped into
+    range so a stale caret (a resync racing a delete) cannot index out of the
+    text. Offsets are measured the same way ``Editor._offset_at`` does, so a
+    caret computed there indexes correctly here. Lines are split on ``\\n`` and
+    a trailing ``\\r`` from a CRLF buffer is stripped below, so every consumer of
+    the returned line agrees on the word regardless of the paste's line endings.
     """
-    lines = text.split("\n")
-    first = next((index for index, line in enumerate(lines) if line.strip()), None)
-    if first is None:
-        return None
-    if len(lines) > first + 1:
-        # A newline after the word terminates it exactly like a space does.
-        return None
-    line = lines[first]
-    stripped = line.lstrip()
-    if not stripped.startswith("/"):
-        return None
-    if any(char.isspace() for char in stripped):
-        return None
-    start = sum(len(other) + 1 for other in lines[:first]) + (len(line) - len(stripped))
-    return SlashContext(start, stripped[1:])
+    if cursor is None or cursor > len(text):
+        cursor = len(text)
+    if cursor < 0:
+        cursor = 0
+    line_start = text.rfind("\n", 0, cursor) + 1  # 0 when no newline precedes
+    line_end = text.find("\n", cursor)
+    if line_end == -1:
+        line_end = len(text)
+    line = text[line_start:line_end]
+    # A CRLF buffer — which a draft/held-prompt restore can carry in, since
+    # ``load_text`` does not normalise — leaves a trailing ``\r`` on every line
+    # split on ``\n``. ``slash_context`` tolerated it (``\r`` is whitespace, so
+    # it terminated the word), but ``slash_word``/``slash_argument_context`` use
+    # ``.partition(" ")``, which does NOT treat ``\r`` as a separator: the word
+    # came back as ``"team\r"`` (matching no command) and argument values
+    # carried the ``\r`` into dispatch. Stripping it here fixes every consumer at
+    # once, because they all read the line through this one helper. The column is
+    # unaffected: the ``\r`` sits at the END of the line, past any caret the
+    # parsers care about (review round 1, minor-1).
+    if line.endswith("\r"):
+        line = line[:-1]
+    return line, line_start, cursor - line_start
 
 
-def slash_argument(text: str, commands: tuple[str, ...]) -> str | None:
-    """The ARGUMENT being typed after one of ``commands``, else ``None``.
+def _boundary_slashes(line: str) -> list[int]:
+    """Indices of every boundary ``/`` on ``line`` (buffer start / after space)."""
+    return [i for i, ch in enumerate(line) if ch == "/" and _is_boundary(line, i)]
+
+
+def _active_slash(line: str, column: int, commands: frozenset[str] = frozenset()) -> int | None:
+    """Index within ``line`` of the boundary ``/`` the cursor is editing.
+
+    Normally the active token is the LAST boundary ``/`` at or before the cursor:
+    a user typing ``a /foo /ba|`` (caret at ``|``) is editing ``/ba``, not
+    ``/foo``. Returns ``None`` when the cursor is not inside any boundary slash
+    token — before every slash on the line, or on a ``/`` glued to a word.
+
+    ``commands`` closes the "already inside a command" case: once the line holds
+    a RECOGNISED command that has been TERMINATED by a space (``/team security
+    …``), that command owns the rest of its line as its argument, so a second
+    ``/team`` typed INSIDE the request ("improve the /team command") is plain
+    argument text — not a new command to highlight or re-open the picker for.
+    The earliest such command on the line wins and claims everything after it;
+    only when no earlier recognised command has claimed the caret's position does
+    the last-slash-before-caret rule apply. Empty ``commands`` (the pure-parser
+    default) disables claiming and keeps the simple behaviour.
+    """
+    slashes = _boundary_slashes(line)
+    candidate: int | None = None
+    for index in slashes:
+        word, sep, _ = line[index + 1 :].partition(" ")
+        if sep and commands and word.lower() in commands:
+            # A recognised, terminated command. Its argument runs to the end of
+            # the line, so if the caret is anywhere past this slash it is inside
+            # THIS command — return it and stop, ignoring every later slash.
+            if column > index:
+                return index
+            # Caret is before this command entirely; nothing earlier can claim.
+            return candidate
+        if index <= column:
+            candidate = index
+    return candidate
+
+
+def slash_context(
+    text: str, cursor: int | None = None, commands: frozenset[str] = frozenset()
+) -> SlashContext | None:
+    """The command word being typed at the cursor, or ``None`` to hide the list.
+
+    Inline and caret-aware, unlike the original first-line-only rule: the picker
+    shows whenever the caret is inside a boundary-slash token whose word is not
+    yet terminated by whitespace, WHEREVER that token is in the draft. This is
+    what lets a user who has typed a message remember to route it — appending
+    ``/team`` at the caret, or dropping it on its own line — and still get the
+    menu. ``src/foo`` never triggers because its ``/`` is not at a boundary.
+
+    Command-WORD phase only: the instant a space terminates the word the user is
+    typing an ARGUMENT (``/model gpt``), which :func:`slash_argument` owns — a
+    command list there is stale advice. ``cursor`` defaults to the end of the
+    buffer, the common "typing at the end" case.
+
+    ``commands`` is the recognised command vocabulary; passing it makes a slash
+    typed INSIDE an already-engaged command's argument (``/team a /team b``) read
+    as plain text rather than a nested command (see :func:`_active_slash`).
+    """
+    line, line_start, column = _line_of_cursor(text, cursor)
+    slash = _active_slash(line, column, commands)
+    if slash is None:
+        return None
+    # The word runs from the slash to the first whitespace after it. A space
+    # BETWEEN the slash and the caret means the word is already terminated and
+    # the caret is out in argument (or message) territory — not the command.
+    word_end = slash + 1
+    while word_end < len(line) and not line[word_end].isspace():
+        word_end += 1
+    if column > word_end:
+        return None
+    return SlashContext(
+        line_start + slash,
+        line[slash + 1 : word_end],
+        line_start + word_end,
+    )
+
+
+def slash_token_span(
+    text: str, cursor: int | None = None, commands: frozenset[str] = frozenset()
+) -> tuple[int, int] | None:
+    """The ``[start, end)`` offsets of the whole slash TOKEN at the caret.
+
+    ``start`` is the ``/`` and ``end`` is the end of its line — the command word
+    plus its inline argument, which by the inline contract runs to the line end.
+    Returns ``None`` when the caret is not on a boundary-slash token. This is the
+    span a RUN splices out of an inline draft, exposed so the editor never has to
+    reach for the module-private tokenizer. ``commands`` is the recognised
+    vocabulary, so a nested slash inside an engaged command's argument is ignored.
+    """
+    line, line_start, column = _line_of_cursor(text, cursor)
+    slash = _active_slash(line, column, commands)
+    if slash is None:
+        return None
+    return line_start + slash, line_start + len(line)
+
+
+def slash_word(
+    text: str, cursor: int | None = None, commands: frozenset[str] = frozenset()
+) -> str | None:
+    """The lower-cased command word of the slash token AT THE CARET, or ``None``.
+
+    The word regardless of PHASE — whether it is still being typed or already
+    terminated by a space — so a caller that needs to know "which command is the
+    caret on" (the editor deciding which argument list to fill) gets one answer
+    whether the buffer reads ``/team`` or ``/team ops``. :func:`slash_context`
+    and :func:`slash_argument` answer the narrower phase-specific questions.
+    ``commands`` is the recognised vocabulary for the nested-slash rule.
+    """
+    line, _, column = _line_of_cursor(text, cursor)
+    slash = _active_slash(line, column, commands)
+    if slash is None:
+        return None
+    return line[slash + 1 :].partition(" ")[0].lower()
+
+
+class SlashArgument(NamedTuple):
+    """The argument being typed, and the spans it occupies in the buffer.
+
+    ``start`` and ``end`` are whole-buffer offsets bracketing the argument text,
+    so a completion can replace JUST the argument and leave a trailing inline
+    message intact — the argument-phase twin of :class:`SlashContext`.
+    ``token_start`` indexes the ``/`` that opens the whole construct, so a RUN
+    can splice the entire ``/cmd arg`` out of an inline draft (not just its
+    argument). ``value`` is the argument text (possibly ``""``).
+    """
+
+    value: str
+    start: int
+    end: int
+    token_start: int
+
+
+def slash_argument_context(
+    text: str,
+    commands: tuple[str, ...],
+    cursor: int | None = None,
+    known: frozenset[str] = frozenset(),
+) -> SlashArgument | None:
+    """The ARGUMENT being typed after one of ``commands``, with its span.
 
     The mirror image of :func:`slash_context`: that one is live while the command
     word is still open, this one takes over the instant the word is terminated by
@@ -225,24 +392,53 @@ def slash_argument(text: str, commands: tuple[str, ...]) -> str | None:
     offers models, and the handover happens on the space the user was going to
     type anyway.
 
-    Returns the argument text, which may be ``""`` (the command word is complete
-    but nothing has been typed after it — the state that should show the whole
-    catalogue). ``None`` means this is not one of ``commands``.
+    Caret-aware and inline like :func:`slash_context`: the argument is the text
+    from the word-terminating space to the END OF THE LINE the command is on.
+    End-of-line, not end-of-buffer, is what makes a command droppable on its own
+    line above a multi-line draft — the line below is the message, not the
+    argument. On the command's own line the argument runs to the line end, so a
+    trailing message on the SAME line (``/team ops ship it``) is read as part of
+    the argument; putting the command last, or on its own line, is how a user
+    keeps the two apart (documented as the inline contract).
 
-    Single-line only, matching ``slash_context``: a newline means the user is
-    composing a message, not picking from a list.
+    ``None`` when the caret is not in the argument phase of one of ``commands``.
+    ``known`` is the FULL recognised vocabulary (a superset of ``commands``) used
+    only for the nested-slash rule: a slash inside an engaged command's argument
+    is not a new token. It defaults to ``commands`` when omitted.
     """
-    lines = text.split("\n")
-    first = next((index for index, line in enumerate(lines) if line.strip()), None)
-    if first is None or len(lines) > first + 1:
+    line, line_start, column = _line_of_cursor(text, cursor)
+    slash = _active_slash(line, column, known or frozenset(commands))
+    if slash is None:
         return None
-    stripped = lines[first].lstrip()
-    if not stripped.startswith("/"):
-        return None
-    word, sep, argument = stripped[1:].partition(" ")
+    word, sep, argument = line[slash + 1 :].partition(" ")
     if not sep or word.lower() not in commands:
         return None
-    return argument
+    # The caret must be in the ARGUMENT, i.e. past the terminating space; while
+    # it is still on the word itself that is command-word phase, which
+    # ``slash_context`` owns. ``slash + 1 + len(word)`` is the space's column.
+    space_column = slash + 1 + len(word)
+    if column <= space_column:
+        return None
+    arg_start = line_start + space_column + 1
+    return SlashArgument(argument, arg_start, line_start + len(line), line_start + slash)
+
+
+def slash_argument(
+    text: str,
+    commands: tuple[str, ...],
+    cursor: int | None = None,
+    known: frozenset[str] = frozenset(),
+) -> str | None:
+    """The argument text being typed after one of ``commands``, else ``None``.
+
+    A thin projection of :func:`slash_argument_context` onto just the text, for
+    the many callers that only rank against the argument and never need to
+    rewrite the span. Returns ``""`` when the command word is complete but
+    nothing has been typed after it (the whole-catalogue state). ``known`` is the
+    full vocabulary for the nested-slash rule.
+    """
+    context = slash_argument_context(text, commands, cursor, known)
+    return None if context is None else context.value
 
 
 #: Below this many typed characters the fuzzy tail is suppressed. A one- or
@@ -355,6 +551,7 @@ class CommandPicker(Static):
         #: row-0 state must not reach the observer (see ``set_choices``).
         self._suppress_report = False
         self._commands: list[SlashCommand] = []
+        self._command_names: frozenset[str] = frozenset()
         self._choices: list[ArgumentChoice] = []
         self._mode = PickerMode.COMMAND
         self._matches: list[_Suggestion] = []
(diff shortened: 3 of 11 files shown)


damianvtran:
### Design review — round 1

Reviewer: designer subagent (Claude Sonnet 4.5)
Scope: main..9a22bcb4

Reviewed the inline slash-command gesture as rendered frames from the real `OperatorApp` (`app.save_screenshot` SVGs at 100×30), not from source. States captured: picker open on an appended `/team` (`fix the auth bug /team`), after running `/team security` (token spliced, "team security is ready" notice, `security` status-band segment), a command mid-sentence with trailing prose, an own-line command above a draft, type-through past the word, caret moved back onto the token, and the no-match empty state. All frames below back the claims.

**Verdict: NOT terminal — one MAJOR.**

---

**D1 (MAJOR) — Running a command mid-line silently eats the trailing prose and deletes it from the composer.**
The inline contract is "the argument runs to the end of the line" (`slash_token_span`, command_picker.py:281-294; `_run_command_from_buffer`, editor.py:2572-2586). That is correct when the trailing text *is* the argument (`/team security`), but the picker also opens — and Enter still runs — when the trailing text is ordinary prose the user means to keep. Reproduced live: composer `review this /team and then ship it`, caret parked right after `team` (picker visible), press Enter. The command runs with `and then ship it` taken as the team name, the app shows `no team named 'and'. Run /team to list teams…`, **and the composer is left holding only `review this`** — "and then ship it" is destroyed with no undo and no warning. `_splice_command`'s span was `(12, 34)`, i.e. the whole rest of the line. This is exactly the same-line command-then-message ambiguity, and it is silent data loss on a plausible gesture (the picker invites the run by being open). The type-through guard (picker hides once the caret passes the trailing space, editor.py — `slash_context` returns `None` at caret 18) *reduces* exposure but does not remove it: with the caret anywhere inside/just-after the word, the footgun is armed. Needs a guardrail or signal — e.g. only treat same-line trailing text as an argument for commands that take one, stop the token at the next run of two spaces, or visibly mark in the picker what will be consumed so "run here eats the rest of the line" is not a surprise. The docstring already concedes the constraint ("a message kept apart from the command sits before the slash or on another line"); the UI gives the user no way to know that.

**D2 (MINOR) — The picker does not point at the token it belongs to; it always left-aligns to the composer's text column.**
By design the command names start in the editor's text column (command_picker.py:16-18), so with `fix the auth bug /team` the highlighted `/team` row renders at column 0 while the live token sits ~17 cells to the right (frame `f_inline`, `f_midcaret`). For the historical first-char case that column *is* the token, so it read as anchored; inline, the row and the thing it describes are visibly disjoint, which mutes the "this menu is for the `/team` I just typed" connection. Not a blocker — the single highlighted row and the token's own text carry it — but a caret-column offset (or a caret-following anchor) would make the inline case as legible as the first-char one. Cite: command_picker.py:11-18.

**D3 (MINOR) — The post-run "ready" notice tells the user to re-type the command they just used, contradicting the inline flow.**
After the inline `/team security` run the notice reads `team security is ready. manager leads it. Send a request with /team security <message>.` (app.py:2887-2888; frame `f_after`). But the whole point of the gesture the user just performed is that their message (`fix the auth bug`) is *already* in the composer with the team attached — they press Enter to send it, they do not type `/team security <message>` again. For the inline entry path the guidance is stale and mildly confusing ("didn't I just do that?"). Consider a variant when a draft survived the splice, e.g. "…press Enter to send your draft to team security, or /team clear to detach", mirroring the `/agent` copy's "…or /agent clear to detach" affordance (app.py:3132-3133).

**D4 (NIT) — No discoverability hint that `/` works mid-draft.**
The boot card still reads `/  command picker` (frames), which historically meant "at the start". Nothing signals the new capability that a boundary `/command` anywhere in the draft opens the menu. A one-word tweak ("`/` command picker (anywhere in a draft)") or a first-run hint would surface the feature; today it is only found by accident.

---

**What works well (verified, no action needed):**
- **Token splice is clean** — no double space, no stranded blank line, caret lands correctly. `fix the auth bug /team security` → `fix the auth bug` with the block caret immediately after `bug` and no trailing space (frame `f_after`); own-line command above/below a draft collapses its separator line (`/team\nsummarize…` → `summarize…`, verified via `_splice_command`). The single-adjacent-whitespace rule (editor.py:2260-2264) is the right call over leaving a blank.
- **Type-through never mutates text** — `fix the auth bug /team ok now more` keeps every character and the picker is gone once the caret is past the word (frame `f_typethrough`).
- **Esc dismiss** leaves the buffer intact and only closes the list (editor.py:1103-1108).
- **Own-line vs same-line** both behave sensibly and the status band gains the `security` segment after attach (frame `f_after`), which is a good persistent signal of the attached team.
- **Empty/no-match** (`/zzzznotacommand`) correctly shows no rows rather than an empty bordered box (frame `f_empty`).

Follow-ups to track regardless of D1's resolution: D2, D3, D4.

damianvtran:
### Agent review — round 1

Reviewer: reviewer (claude-opus-4.6 / Claude Code)
Scope: main..9a22bcb4

Reviewed the full diff (`command_picker.py`, `editor.py`, `app.py`, tests), ran
the changed TUI suites and the lint/type gates from the worktree, and probed the
parser edge cases directly.

**Verification performed**
- `test_command_picker.py` + `test_slash_echo.py` → 257 passed.
- `test_model_picker.py` + `test_approvals_ux.py` + `test_theme_command.py` + `test_effort.py` → 102 passed.
- `test_app_pilot.py` → 166 passed.
- `flake8` and `pyright` on the three changed source files → clean.
- Hand-checked: boundary detection (`src/foo`, `and/or`, `foo/bar` all correctly return `None`); caret-anchored active-token selection between two inline slashes; whole-buffer vs inline dispatch decision; word→argument phase handover; `_offset_at`/`_location_at_offset` round-trip; that column indices are codepoint (not cell) based, so wide chars / emoji don't skew the offset math.

The caret math, span splicing, separator absorption (incl. own-line collapse),
the `begin_model_query` re-sync fix, and the type-through/Esc "never mutate the
draft" invariant all hold up. Start-of-line command behaviour is preserved
(`test_a_whole_buffer_command_still_submits_not_inline`).

I found no blocker and no major. Two minors and a nit follow.

---

**minor-1 — CRLF `\r` leaks into the command word/argument on a non-final line.**
`command_picker.py:210-227` `_line_of_cursor` splits only on `\n` and slices
`text[line_start:line_end]`, so on a CRLF buffer every line except the last keeps
a trailing `\r`. `slash_context` hides this because it scans for the first
`isspace()` char (`\r.isspace()` is True), but `slash_word` (`:310`) and
`slash_argument_context` (`:357`) terminate with `.partition(" ")`, which does
**not** split on `\r`. Result on a CRLF draft with a slash command NOT on the
last line:

```
t = "note\r\n/model gpt\r\nmsg below"
slash_word(t, caret_on_word)          -> "model"      # ok here (\r is after 'gpt')
slash_word("/team\r\nfix", 5)         -> "team\r"     # word itself carries \r
slash_argument_context(t, ("model",...), caret_in_arg)
                                      -> value="gpt\r"
```

`_command_word()` → `slash_word()` returns `"team\r"`, which will not match the
argument-command set, so the argument list won't fill for that command; and the
argument value ranks/dispatches with a trailing `\r`. This directly contradicts
the `_line_of_cursor` docstring's claim that "the composer normalises paste" —
true for the *insert* path (Textual splits pasted lines), but **`load_text` does
not normalise**, and CRLF reaches `editor.text` through the draft/held-prompt
restore paths (`app.py:2400,3833,9320,11518`). Note the sibling `_offset_at`
docstring explicitly supports "a CRLF buffer a paste carried in", so the two
assumptions disagree. Reachability is narrow (Windows-EOL restored draft, command
off the final line) and dispatch `.strip()`s the token, so it degrades the picker
UX rather than corrupting a send — hence minor. Fix: strip the trailing `\r` in
`_line_of_cursor` (`line.rstrip("\r")` when slicing, adjusting nothing else since
`\r` sits at the line end).

**minor-2 — same-line trailing prose is silently swallowed as the argument.**
By the documented inline contract `slash_argument_context` runs the argument to
end-of-line, so `please /team review the auth changes` dispatches
`/team review the auth changes` (argument = `review the auth changes`). This is
intended and tested, and the "put it on its own line / at the end" escape hatch
exists — but the *only* guard against a natural-prose false positive is
documentation; there is no test asserting what a multi-word same-line tail does
to a real command handler. Worth a one-line test pinning the contract (a command
whose handler rejects surplus args would surface a confusing error). Not a defect
in this diff.

**nit — contradictory CRLF docstrings.** `_line_of_cursor` ("composer normalises
paste", `command_picker.py:216`) vs `_offset_at`/`_caret_offset` ("a CRLF buffer
a paste carried in", `editor.py:1520-1523, 2201-2205`). Reconcile these; as
written a reader can't tell whether CRLF is possible in the buffer, which is
exactly the ambiguity behind minor-1.

---

**Verdict: clean.** No blocker, no major. This round is TERMINAL — minor-1,
minor-2 and the nit are follow-ups, not grounds for another required round.
Recommend addressing minor-1 (or explicitly deferring it with a ticket) since it
is a real, if narrow, correctness gap with a one-line fix.

damianvtran:

Reviewer round: [Agent review — round 1](https://github.com/damianvtran/local-operator/pull/250#issuecomment-5381687333) (verdict: clean, no blocker/major)
Remediated at: `b75213b2`

- **minor-1 (CRLF `\r` leak) — FIXED in `b75213b2`.** `_line_of_cursor` now strips a trailing `\r` from the line it returns, so every consumer (`slash_word`, `slash_argument_context`, `slash_context`) agrees on the word regardless of line endings, and an argument value no longer carries the control character into dispatch. New pure test `test_a_crlf_buffer_does_not_leak_a_carriage_return_into_the_word` pins `slash_word("/team\r\nfix", 5) == "team"` and `slash_argument("/team ops\r\nmsg", ("team",), 9) == "ops"`.

- **minor-2 (same-line trailing prose swallowed as argument) — FIXED in `b75213b2`, no longer merely documented.** This was the same underlying issue as design review D1, and it is now structurally impossible for the free-text commands: `/goal`, `/loop`, `/btw`, `/team`, `/agent` are flagged `consumes_prompt` and, engaged inline, REASSEMBLE to the front (`/command [name] <draft>`) and STAGE rather than running against the trailing text. A multi-word tail is preserved verbatim as the argument. Pinned by `test_inline_team_reassembles_to_the_front_and_then_sends_under_the_manager` and `test_a_prompt_command_reassembles_to_the_front_staged_not_run`.

- **nit (contradictory CRLF docstrings) — FIXED in `b75213b2`.** `_line_of_cursor`'s docstring no longer claims paste is normalised; it now states the `\r` strip and that offsets match `_offset_at`.

Additional hardening from the design round and a follow-up report (same commit): the tokenizer is now command-aware, so a nested slash inside an engaged command's argument (`/team a improve the /team command`) is plain text — no picker, no highlight, no re-engagement (`test_a_nested_slash_inside_an_engaged_command_is_plain_text`, `test_a_second_inline_command_of_the_same_kind_is_plain_argument_text`).

Gates on `b75213b2`: flake8 / black / isort / pyright clean; `pytest tests/unit/tui` — 2229 passed, 4 skipped.

damianvtran:

Reviewer round: [Design review — round 1](https://github.com/damianvtran/local-operator/pull/250#issuecomment-5381598900) (verdict: not terminal — one MAJOR)
Remediated at: `b75213b2`

- **D1 (MAJOR, silent data loss) — FIXED in `b75213b2`.** `review this /team and then ship it` no longer runs `/team` against the trailing prose and discards it. The free-text commands (`/goal`, `/loop`, `/btw`, `/team`, `/agent`) now carry `consumes_prompt`, and engaging one inline REASSEMBLES the composer to `/command [name] <the rest of the draft>` and STAGES it — nothing is submitted, nothing is consumed as a name, and the user reviews the assembled prompt before sending. `/team`/`/agent` reassemble when the name is picked from the autofill (draft → request); `/goal`/`/loop`/`/btw` reassemble on word completion (draft → argument). Verified live and in tests (`test_inline_team_reassembles_to_the_front_and_then_sends_under_the_manager`); see the updated "After reassembly" screenshot in the PR description showing `review the auth flow /team` + pick `security` → `/team security review the auth flow`, staged.

- **D3 (post-run "ready" notice tells you to re-type the command) — RESOLVED by the reassembly flow.** With reassembly, `/team <name>` is not "run then ready to re-type" — the assembled `/team <name> <draft>` is already in the composer staged for one Enter, so the flow no longer contradicts a "send a request with /team …" instruction.

- **D2 (picker row left-aligns to the text column, not the token) — DEFERRED (ticket to follow).** The picker is docked below the input row and always renders at the composer's left edge; anchoring the popup horizontally to the inline token is a separate layout change with its own overlay/offset considerations (and the AGENTS.md overlay constraints). It does not affect correctness or the data-safety fix. I'll file a follow-up rather than fold a popup-positioning change into this PR.

- **D4 (no discoverability hint that `/` works mid-draft) — DEFERRED (ticket to follow).** A tip-pool line ("`/` works anywhere in the composer") is the natural home; folding copy into `welcome.TIPS` is independent of this change.

Re-capture: the "after" screenshot in the PR description is refreshed to the reassembly behaviour. Ready for design round 2 on the current head.

damianvtran:
### Design review — round 2

Reviewer: designer subagent (Claude Opus 4.6)
Scope: 9a22bcb4..b75213b2

Reviewed the round-1 remediation (commit `b75213b2`) as rendered frames from the real `OperatorApp` (`app.save_screenshot` at 100×30, rasterised and viewed), backed by live gestures driven through the pilot against the actual `Editor`. Focus: is D1 (the silent data-loss MAJOR) truly gone on the current head, and is the reassemble affordance sound. States captured: inline picker open on an appended `/team` (`fix the auth bug /team`), the team list open with prose behind it (`review the auth flow /team` → `security`), the staged reassembled line, and a from-scratch typed `/team security review the auth flow` for comparison.

**Verdict: TERMINAL on the design side — no BLOCKER, no MAJOR. D1 is resolved.**

---

**D1 (was MAJOR) — RESOLVED, verified live.** The silent data-loss path is gone on every gesture I could reach for it:

- `review this /team and then ship it` typed in full, caret after the token, **Enter** → the whole line dispatches as a plain prompt (`('prompt', 'review this /team and then ship it')`), buffer clears to sent, nothing consumed as a team name. The old bug left `review this` and destroyed `and then ship it`; that is no longer reachable.
- `do the thing /team security now` + Enter → also dispatches the entire line as a prompt, not `/team` with argument `security now`. Even a tail that *begins* with a real team name no longer gets eaten — the inline `/team` is inert in the argument phase.
- Picking a team name from the autofill reassembles-and-stages, no submit: `review the auth flow /team` → open list → `security` → Enter yields `/team security review the auth flow` staged with the caret at the end (offset 35), `_argument_command == 'team'`, nothing sent. Multi-word tails survive intact.
- `/goal`/`/loop`/`/btw` reassemble on word completion (Enter/Tab on the picker): `write the migration /goal` → `/goal write the migration`; `keep improving it /loop` → `/loop keep improving it`; `what does CQRS mean /btw` → `/btw what does CQRS mean`. Draft preserved as the argument in every case.
- Nested slash: `/team a improve the /team command` stays a single line of plain text, the second `/team` does not open a new picker (`picker.is_open() == False`), and the argument is not truncated at it.
- Non-prompt inline still splices-and-runs: `fix this /usage` + Enter → buffer becomes `fix this`, `/usage` dispatched, surrounding draft kept.
- CRLF restore (`/team\r\nfix the bug`) resolves the word as `team` at the caret and opens the command picker — the `\r` no longer poisons the word match.

**The reassembly affordance itself — sound, no cue needed (verified visually).** My round-1 worry was that a command silently jumping to the front of the composer could read as a glitch. The rendered frames retire it: the staged line `/team security review the auth flow` is **pixel-identical** to the same text typed from scratch (both SVGs render to the exact same 40288-byte PNG) — the `/team` word paints as recognised, `security` sits in the argument slot, the caret lands cleanly at the end, and there is no stranded separator, double space, or blank line. Because the command genuinely IS at the front and the whole draft is visibly present as its argument, the "jump" is legible as "your command, assembled" rather than a surprise: the user reads one line and presses Enter. The list-open intermediate frame (`security · 2 roles · led by manager` under the prose-bearing input) also reads correctly. I do not think a separate motion cue is warranted — the staged result is self-explanatory and the STAGED-not-run contract (one deliberate Enter) is the safety.

**D2 (MINOR, picker left-aligns to the text column rather than the inline token) — deferral ACCEPTED.** Still true on the current frames (the highlighted row sits at column 0 while the live token is ~17 cells right), but it is a popup-anchoring layout change with its own overlay constraints and it does not touch correctness or the data-safety fix. Follow-up ticket is the right call.

**D3 (was MINOR, post-run "ready" notice told you to re-type the command) — RESOLVED by the reassembly flow**, as claimed in remediation: `/team <name> <draft>` is staged for one Enter rather than run-then-ready, so the notice no longer contradicts the flow.

**D4 (NIT, no discoverability hint that `/` works mid-draft) — deferral ACCEPTED.** The boot card still reads `/  command picker` with no signal of the mid-draft capability; a `welcome.TIPS` line is the natural, independent home. Fine to track separately.

---

**What works well (verified, no action needed):**
- Reassembled state is visually indistinguishable from a hand-typed command — no glitch, correct highlight and caret (frames `f_reassemble` / `f_normal_front`, byte-identical).
- Splice for non-prompt inline (`/usage`) keeps the surrounding draft with no double space or stranded line.
- Nested `/team` inside an engaged argument stays plain text; the picker does not re-open on it.
- The full-line prompt fallback (an inert inline prompt-command dispatches the whole line) is the safe default — text is never lost, worst case the user sends a line containing a literal `/team` and re-issues.
- 261 picker + slash-echo tests pass on the current head, including the reassembly and nested-slash cases.

**Follow-ups to track (both deferred, accepted): D2, D4.**

Design round is TERMINAL on `b75213b2`. Full-gate terminality also needs the code round-2 to be clean on this same head; that is the reviewer's call, not covered here.

damianvtran:
### Agent review — round 2

Reviewer: independent code-reviewer (Claude Sonnet 4.5)
Scope: 9a22bcb4..b75213b2 (full PR base `main`..b75213b2 also read)

Round-2 review of the round-1 remediation. I read the diff, exercised the tokenizer and the run/reassembly paths directly, and drove the D1 gesture end-to-end in the pilot harness. Verdict up front: **no blocker, no major — this round is TERMINAL.**

#### Round-1 findings — all verified closed on this head

- **minor-1 (CRLF `\r` leak) — fixed.** `_line_of_cursor` strips a trailing `\r` before any consumer sees the line (command_picker.py:82-84), so `slash_word`/`slash_argument_context`/`slash_context`/`slash_token_span` all read the same `\r`-free line. Confirmed directly: `slash_word("/team\r\nfix", 5, {"team"}) == "team"` and `slash_argument("/team ops\r\nmsg", ("team",), 9) == "ops"`. Pinned by `test_a_crlf_buffer_does_not_leak_a_carriage_return_into_the_word`.
- **minor-2 / D1 (same-line prose consumed as argument) — fixed.** Prompt commands no longer splice-and-run inline. `_run_command_from_buffer` routes any `consumes_prompt` word to `_reassemble_prompt_command`, which STAGES `/<cmd> <draft>` and submits nothing. Verified the exact reported string end-to-end: typing `review this /team` + Enter opens the team list (does not run); the trailing prose survives; and even the degenerate "type `/team` then hand-type prose after it, then Enter" path just submits the whole buffer verbatim (`review this /team and then ship it` → sent as-is, `inline_commands == []`, nothing eaten). No path silently consumes the draft.

#### Remediation mechanics — checked

- **Reassembly math** (`_reassemble_prompt_command`, editor.py:2639-2673) preserves the full multi-word draft, removes exactly one adjoining separator with the same rule the splice uses, parks the caret at end, and re-syncs so the front word paints recognised and a list-taking command reopens its argument list. `test_inline_team_reassembles_...` proves `/team feature-release ship the dashboard` is assembled, staged, and then sends `ship the dashboard` under the manager on the user's Enter.
- **Command-aware tokenizer** (`_active_slash` claiming rule) is correct across the cases I probed: nested `/team` inside an engaged `/team` argument is plain text (`_active_slash` returns the earliest recognised terminated command and stops); a glued slash (`src/foo`, `and/or`) never opens; a slash at line-start mid-buffer works; prefix confusion is clean (`/log` is a word, `/logout anthropic` is argument-phase); caret before the slash yields `None`. The `commands` (argument-list subset) vs `known` (full vocabulary) threading is consistent — `known` is used for the nested-slash claim, `commands` only to decide which argument list to fill.
- **`consumes_prompt` field + `PROMPT_POLICY` pin** present and enforced entry-by-entry in `test_every_registered_command_states_an_echo_policy`, mirroring `ECHO_POLICY`. `goal/loop/btw/team/agent = True`, all selectors/listings `False`.
- **No regression** to non-prompt inline splice-and-run (`/usage` → `InlineCommandRequested`, draft kept, `test_inline_command_runs_and_is_spliced_out_keeping_the_draft`) or whole-buffer submit (`test_a_whole_buffer_command_still_submits_not_inline`).

#### Gates

- Changed suites green: `test_command_picker.py` + `test_slash_echo.py` (261 passed), `test_app_pilot.py` (166 passed), `test_approvals_ux.py`/`test_effort.py`/`test_model_picker.py`/`test_theme_command.py` (all passed).
- `flake8` clean on all four changed source files. `pyright` shows only `reportMissingImports` noise (venv site-packages not resolved in this invocation) plus one `reportOptionalMemberAccess` in app.py that is **pre-existing on `main`** (same warning at the corresponding line before this PR) and untouched by the diff.

#### MINOR

- **minor-1 (reorder wart, not data loss).** If a user hand-types the argument on an inline prompt command (`review this /team ops`) and presses Enter, `_reassemble_prompt_command` produces `/team ops review this` — the trailing prose is appended after the already-typed argument, so the two swap order. It still parses correctly (name `ops`, request `review this`), is STAGED not run, and the intended autofill flow never hits this ordering. Cosmetic only; worth a comment noting the reorder is deliberate for the hand-typed-argument case, or none.

#### NIT

- **nit-1 (tab vs space separator).** `_active_slash`/`slash_argument_context` terminate the word with `partition(" ")` (space only) while `slash_context` terminates with `isspace()`. A tab-separated command word (`/team\tops`) therefore falls into a dead zone where neither the word picker nor the argument picker opens. This matches the pre-inline convention (argument parsing was always space-only) and a tab in the composer indents rather than separating, so it is unreachable in practice — flagging only for the record.

**Verdict: clean. Round 2 is TERMINAL** — no blocker or major remains; the one minor and one nit are follow-ups, not grounds for another round.

damianvtran:
### Agent review — round 2 (scope update)

Reviewer: independent code-reviewer (Claude Sonnet 4.5)
Scope: 9a22bcb4..872817bd (extends the round-2 scope from b75213b2)

The author pushed 872817bd on top of the b75213b2 I reviewed. The delta is **comment-only**: `git diff b75213b2..872817bd -- '*.py'` filtered to non-comment code lines prints nothing, and the single changed file adds 11 comment lines in `_reassemble_prompt_command` documenting the deliberate hand-typed-argument reorder (`/team ops` + `review this` → `/team ops review this`, staged not run) — i.e. exactly the round-2 minor-1 I raised, now answered in the code.

No executable code changed, so my round-2 verification stands unchanged. **The TERMINAL verdict holds on the current head 872817bd: no blocker, no major.** Merge gate satisfied.

damianvtran:
### Rebase + reconciliation — re-review needed (round 3)

The branch was rebased onto current `main` (`7e97d19c`) and force-pushed; new head **`5b9f2ef3`**. `main` had advanced past the branch base with several TUI PRs that overlap this change — notably **#246** (autofill team/agent name completions + recognized-slash highlighting), plus **#247** (`/analytics` command), **#245**, **#243**. CI on the pre-rebase head failed only on `test_every_registered_command_states_an_echo_policy` (the registry had gained `/analytics`), which surfaced the drift.

Reconciliation (all in the single squashed commit):
- Merged this feature with **#246**'s name-autofill: `/team`/`/agent` argument rows still fill `<name> ` and park at the START of the buffer, but when engaged INLINE (a draft typed before the command) `_complete_name_argument` now reassembles the whole `/<cmd> <name> <draft>` to the front and stages it — the D1-safe path, integrated with the autofill rather than replacing it.
- Reused #246's `_command_names` vocabulary for the nested-slash tokenizer rule (dropped my duplicate `_known_commands`).
- Kept #246's recognized-slash **highlighter** (`_compute_slash_runs`) and the U1/U2 name-switch hint; threaded the caret through my inline parsers without disturbing them.
- Added `/analytics` to the `PROMPT_POLICY` pin (a view selector → `False`, splice-and-run inline like `/usage`).

Verification on `5b9f2ef3`: `flake8`/`black`/`isort`/`pyright` clean on the changed files; **`pytest tests/unit/tui` — 2289 passed, 4 skipped**. Inline reassembly re-verified live end-to-end with the merged autofill (`review the flow /team` → pick `security` → `/team security review the flow`, staged, nothing consumed).

Because the reviewed content changed materially from the round-2 head, a fresh review round follows. The prior round-1/round-2 findings (D1 data-loss, CRLF, nested-slash) remain addressed on this head.

damianvtran:
### Agent review — round 3

Reviewer: independent code-reviewer subagent (claude-opus-4.6)
Scope: 7e97d19c..5b9f2ef3

Round-3 review of the rebase onto current `main` (7e97d19c), reconciling the inline slash-command feature with the newly-landed #246 (name autofill + slash highlighting), #247 (/analytics), #245, #243. I read the diff (`git diff main...HEAD`), traced the parser call sites, exercised the parsers directly, and ran the targeted suites.

**Verified — reconciliation with #246**
- `_complete_name_argument` is correct in BOTH modes. Start-of-buffer `/team` still fills `<name> ` into the argument span and parks the caret at end (`test_enter_on_a_team_row_fills_the_name_and_a_space_without_submitting`, `test_tab_on_a_team_row_fills_name_and_space` pass). Inline (a draft precedes the token) reassembles the whole `/cmd name draft` to the front and stages it (`test_inline_team_reassembles_to_the_front_and_then_sends_under_the_manager` passes). The span math is sound: the name+space is filled into `[context.start, context.end)`, the token span is recomputed on the *filled* text before reassembly (the comment flags exactly why — the name changed length), so no stale offset survives.
- The nested-slash tokenizer reuses #246's `_command_names` vocabulary and there is no lingering `_known_commands` (grep across `local_operator/` and `tests/` is empty). `_active_slash` claims the rest of the line for the earliest recognised, terminated command, so `/team alpha improve the /team command` keeps the second `/team` as plain argument text (`test_a_second_inline_command_of_the_same_kind_is_plain_argument_text` passes; confirmed directly via `slash_token_span`).
- `_compute_slash_runs` (the recognised-slash + name highlighter) is **untouched** by this PR — the diff only threads the caret into the *hint* call at `_sync_picker`. It remains first-content-line-only, and the U1/U2 `_name_switch_hint` still gates on the parked `<name> ` resting state. No caret-threading disturbance to the highlighter.

**Verified — inline feature + r1/r2 fixes still hold**
- D1 data-loss path closed: `/goal` inline reassembles to `/goal land the oauth fix` staged, not run (`test_a_prompt_command_reassembles_to_the_front_staged_not_run`). Prompt commands (goal/loop/btw/team/agent) never splice-and-run.
- Non-prompt inline splice-runs and keeps the draft (`InlineCommandRequested`); whole-buffer still submits (`test_a_whole_buffer_command_still_submits_not_inline`).
- CRLF fix intact: `_line_of_cursor` strips a trailing `\r` so `slash_word`/`slash_argument`'s `.partition(" ")` see a clean word (`test_a_crlf_buffer_does_not_leak_a_carriage_return_into_the_word`).
- `PROMPT_POLICY` pins every command entry-by-entry, including `/analytics` (False — a view selector that splices-and-runs like `/usage`). Confirmed `{c.name: c.consumes_prompt} == PROMPT_POLICY` and the registry key set matches the policy exactly; `consumes_prompt` is set on exactly {agent, btw, goal, loop, team}.
- Every changed parser signature (`slash_context`, `slash_argument`, `slash_argument_context`, `slash_word`, `slash_token_span`, `CommandPicker.sync`) has all call sites updated with caret + vocabulary; the one app-side caller (`app.py:10676`) threads `_caret_offset()` and `_command_names`.

**Suites run (from /tmp/lo-inline-slash, repo .venv)**
- `test_command_picker.py` + `test_slash_echo.py`: 282 passed.
- `test_app_pilot.py`: 166 passed.
- `test_model_picker.py` + `test_theme_command.py` + `test_approvals_ux.py` + `test_effort.py`: 104 passed.
- `flake8` clean on all four touched source files.

**Findings**

No BLOCKER, no MAJOR, no MINOR.

NIT (N1): `_reassemble_prompt_command`'s documented reorder case — a user who *hand-types* an argument and *then* engages inline (`/team ops` + `review this` → `/team ops review this`) reads the two in the opposite order to how they were typed. Already acknowledged in the docstring and preferred over guessing, and the result is staged not run, so the user can fix it. Follow-up only.

**Verdict: clean. This round is TERMINAL.** No blocker or major remains; the reconciliation preserves #246's start-of-line fill-and-park and highlighting while the inline feature (reassembly, D1 safety, nested-slash-as-text, non-prompt splice-run, whole-buffer submit, CRLF) holds. The single nit is a follow-up, not a reason for another round.

damianvtran:
### Design review — round 3

Reviewer: designer (Claude Sonnet 4.5)
Scope: 7e97d19c..5b9f2ef3

Rebase-reconciliation review against #246 (team/agent name autofill + recognized-slash highlighting). I drove the real `OperatorApp` to rendered SVG frames (not the CSS-less harness hosts) over a seeded `TeamRegistry`, and cross-checked each behaviour against the same flow on `main` (7e97d19c) in a throwaway worktree. Frames captured: start-of-line `/team ` autofill list, start-of-line pick → parked `/team security `, inline `review the flow /team` word state, inline pick → reassembled `/team security review the flow` (staged), nested `/team a improve the /team command`, and the highlighted staged line.

**What holds up (verified live, D1 intact):**
- **Inline reassembly is correct and D1-safe.** `review the flow /team` → open list → pick `security` reassembles to `/team security review the flow`, **staged not run**, caret at end, `/team` painted as the recognized command word (blue) and `security` as a recognized name (green), the draft preserved verbatim. Three consecutive post-pick frames are byte-identical — the jump-to-front settles atomically, **no flicker, no mispaint**.
- **The inline command highlights in place before it jumps.** `review the flow /team` paints `/team` green with the command row anchored below the caret; on pick it relocates to the front in one settled frame. Highlighting and reassembly read coherently together.
- **Nested slash stays plain.** `/team a improve the /team command` renders as one line: the leading `/team` is highlighted, the nested `/team` inside the argument is plain white text — the command-aware tokenizer does not mispaint it.
- **Non-prompt `/usage` inline splices-and-runs.** `check this /usage` + Enter posts `InlineCommandRequested('/usage')` and leaves `check this` in the composer. Draft kept.
- **D1 data-loss stays fixed.** `review this /team and then ship it` + Enter (no team matched, list closed) submits the text verbatim as a plain message — nothing silently eaten.
- **Start-of-line #246 flow unchanged where it survives.** `/team ` opens the autofill list (frontend-guild / security with role counts); picking a name fills `/team security ` and parks the caret after the space with the recognized-name highlight.

---

**D5 (MAJOR) — The #246 U1/U2 "switch · send" parked-caret hint no longer appears in the real keyboard flow; the reconciliation silently killed a shipped #246 affordance.**

On `main`, completing a team/agent name by keystroke (`/team ` → `sec` → Enter/Tab) ends with the picker still shown and its notice row reading **"Enter to switch · type a message to send"** — the whole point of #246's U1/U2 work, which names the two otherwise-invisible outcomes of the parked caret. On this branch the identical keystroke flow ends with **the picker closed and the notice empty** — the hint never shows.

Reproduced live (real `OperatorApp`, not the harness), stable across 8 settle ticks, for **both** Tab and Enter and for **both** `/team` and `/agent`:

```
# HEAD 5b9f2ef3
team enter-pick:  text='/team security '  display=False  notice=''
team tab-pick:    text='/team security '  display=False  notice=''
agent enter-pick: text='/agent architect ' display=False notice=''

# main 7e97d19c (same keystrokes)
pick via enter:   text='/team security '  display=True   notice='Enter to switch · type a message to send'
```

Cause (traced): this PR routes name completion through the new `_complete_name_argument` → `_set_text_and_caret`, which sets `self.text` (an origin-anchored `_sync_picker` fires first, resetting `_argument_command` to `None`) and then re-syncs at the real caret. That second sync sees the command word "change" `None → team`, re-posts `ArgumentQueryOpened('team')`, and the app's handler (`on_argument_query_opened`, the `("team","teams")` branch) calls `picker.set_notice("")` **after** the editor's own sync had just set the hint — wiping it. On `main`, `_complete_name_argument` set `self.text` directly (a single sync), so no second `ArgumentQueryOpened` fired and the hint survived. `_set_text_and_caret` is new in this PR (absent from `main`), so this is a reconciliation-introduced regression, not pre-existing.

Why this is not caught by the suite: `test_autofill_shows_the_switch_or_send_hint` and friends (282 green) drive completion via the harness's `set_editor_text` and the harness `on_argument_query_opened` does **not** clear the notice — so the app-only double-fire is invisible to it. Green tests here are not evidence the hint renders; the rendered frame is. (My start-of-line "parked" frame shows `/team security ` with the caret parked and **no** notice row, which is the visible symptom.)

Severity MAJOR, not blocker: no data loss and the inline feature itself is unaffected — but it regresses a user-facing affordance that #246 shipped, in exactly the flow #246 built it for, and the whole subject of this round is whether the reconciliation with #246 is correct. Suggested fix directions (impl's call): have the name-completion path avoid the spurious `ArgumentQueryOpened` re-fire (e.g. don't let `_set_text_and_caret`'s pre-move sync null out `_argument_command` when the word is unchanged), or re-assert the hint after the app handler runs. Whatever the fix, please back it with a **real-app** (not harness) assertion on `picker._notice` after a keystroke pick, since the current tests pass through this regression.

---

**Prior deferrals — still acknowledged, still accepted:**
- **D2 (MINOR, picker row left-aligns to the composer's text column rather than the inline token).** Still true on this round's frames: in `review the flow /team` the highlighted command row sits at column 0 while the live token is ~17 cells to the right. Popup horizontal anchoring is a separate layout change with its own overlay constraints; correctness and D1 are untouched. Deferral stands.
- **D4 (NIT, no discoverability hint that `/` works mid-draft).** The boot card still reads `/  command picker` with no signal of the mid-draft capability. A `welcome.TIPS` line is the natural, independent home. Deferral stands.

---

**Verdict: NOT TERMINAL — one MAJOR (D5) open.** The inline feature, reassembly, D1 data-safety, nested-slash, non-prompt splice-run, and recognized-slash highlighting are all correct and read coherently on the current head. The single blocker to terminal is the reconciliation regression of #246's U1/U2 parked-caret hint (D5 above), which must be restored in the real keyboard flow (and covered by a real-app test) before this round is clean. D2/D4 remain accepted follow-ups.

damianvtran:
### Design review remediation — round 3

Reviewer round: [Design review — round 3](https://github.com/damianvtran/local-operator/pull/250#issuecomment-5382347992) (verdict: not terminal — one MAJOR)
Remediated at: `487c8298`

- **D5 (MAJOR, #246 parked-caret hint killed) — FIXED in `487c8298`.** Your diagnosis was exact: `_set_text_and_caret` set `self.text` (firing an origin-anchored `_sync_picker` that nulled `_argument_command`), then re-synced at the real caret, so the second sync saw the word "change" `None → team`, re-posted `ArgumentQueryOpened('team')`, and the app handler's `set_notice("")` wiped the editor-set hint.

  Fix: `load_text` now honours a `_suspend_picker_sync` guard that `_set_text_and_caret` raises around the `text`-set + `move_cursor`, so **exactly one** sync runs, at the final caret. `_argument_command` is never transiently nulled, no spurious `ArgumentQueryOpened` fires, and the hint survives. Verified live on the real `OperatorApp`:

  ```
  # HEAD 487c8298 — start-of-line /team, fill the single team
  tab-pick:   text='/team security '   notice='Enter to switch · type a message to send'
  enter-pick: text='/team security '   notice='Enter to switch · type a message to send'
  ```

  Backed by a **real-app** (not harness) regression test as you asked —
  `test_completing_a_team_name_keeps_the_parked_hint_in_the_real_app` in
  `test_slash_echo.py` drives `OperatorApp`, completes the name by keystroke, and
  asserts `picker._notice` still carries the switch/send hint. It fails on the
  pre-fix double-sync and passes now.

- **D2 (picker left-aligns to the text column) — DEFERRED, accepted.** Popup horizontal anchoring to the inline token is a separate overlay-layout change; correctness and D1 are untouched. Ticket to follow.
- **D4 (no mid-draft discoverability hint) — DEFERRED, accepted.** A `welcome.TIPS` line is the independent home. Ticket to follow.

Gates on `487c8298`: flake8 / black / isort / pyright clean on the changed files; `pytest tests/unit/tui` — **2290 passed, 4 skipped**. The inline feature, reassembly, D1 safety, nested-slash, non-prompt splice-run, and recognized-slash highlighting all remain correct (round-3 verified). Ready for design round 4 on the current head.

damianvtran:
### Design review — round 4

Reviewer: designer (Claude Sonnet 4.5)
Scope: 5b9f2ef3..487c8298 (the D5 fix delta; also read main..487c8298 for whole-PR context)

Round-4 design pass on the D5 fix only. D5 (design r3, MAJOR): completing a team/agent name by keystroke at the start of the buffer closed the picker AND wiped #246's parked-caret hint ("Enter to switch · type a message to send"), because `_set_text_and_caret`'s buffer set funnelled through `load_text` → an origin-anchored `_sync_picker` that transiently nulled `_argument_command`, making the final caret-anchored sync re-post `ArgumentQueryOpened`, whose app handler calls `set_notice("")`.

**Verification — driven against the REAL `OperatorApp`** (not the bespoke harness; its `on_argument_query_opened` doesn't clear the notice, so the regression is only observable live), over a seeded `TeamRegistry`/`AgentRegistry`, at 120×40. All four cross-product cases captured as rendered SVG frames and inspected:

| case | buffer | caret | picker notice |
|---|---|---|---|
| /team + Tab | `/team security ` | (0,15) parked after space | Enter to switch · type a message to send |
| /team + Enter | `/team security ` | (0,15) | Enter to switch · type a message to send |
| /agent + Tab | `/agent auditor ` | (0,15) parked after space | Enter to switch · type a message to send |
| /agent + Enter | `/agent auditor ` | (0,15) | Enter to switch · type a message to send |

Every frame renders cleanly: command token bold, name in green, one trailing space, caret block parked after it, and the switch/send hint on the notice row below. No flicker, no stray highlight. **D5 is resolved for both Tab and Enter and both /team and /agent.**

**Guard safety.** The new `_suspend_picker_sync` defaults `False` and is raised only inside `_set_text_and_caret`'s `try/finally` (restored even on exception), so no external `load_text` caller — held-prompt restore (app.py:2428), history restore (3888), draft restore (9460), /btw retract (11689), ask-question (9548) — can observe it truthy; each still runs `load_text`'s own `_sync_picker`. Exactly one sync now runs, at the final caret. Safe and, as the comment notes, cheaper.

**Round-3 behaviours re-confirmed unchanged** (the fix is a one-line guard, so this is a spot-check not a re-review): inline reassembly frame `/team security review the auth flow` renders staged-not-run at the front, `/team` bold + `security` green + prompt plain, caret at end, coherent highlight, no flicker. Full picker + slash-echo suites: 283 passed. New real-app regression `test_completing_a_team_name_keeps_the_parked_hint_in_the_real_app` passes and asserts the hint against `OperatorApp` — closing exactly the "green tests aren't the rendered frame" gap round 3 named.

**Findings:** none. No new BLOCKER or MAJOR. D2 (picker left-align) and D4 (discoverability) remain accepted deferrals — not reopened.

**Verdict: TERMINAL.** No blocker or major remains on the current head (487c8298). The design side of PR #250 is clean.

damianvtran:
### Agent review — round 4

Reviewer: independent reviewer subagent (code), Claude Sonnet 4.5
Scope: 5b9f2ef3..487c8298 (the D5 fix delta; also read main..487c8298 for whole-PR context)

Round-4 scope is the D5 remediation only: the `_suspend_picker_sync` guard added to `Editor.load_text` + `_set_text_and_caret`, and the new real-app regression test.

**D5 fix is correct and minimal.**

1. **Guard fixes D5.** `_set_text_and_caret` now sets `_suspend_picker_sync = True` in a `try`/`finally` around `self.text = text` + `move_cursor(...)`, so the `_sync_picker()` inside `super().load_text` is skipped and exactly one sync runs at the final caret (editor.py:2570-2576, gated at 2538). With the caret already at the parked position, that single sync sees the active `/team <name>` argument, so `_argument_command` is never transiently nulled and no spurious `ArgumentQueryOpened` re-fires — the app's `on_argument_query_opened → set_notice("")` no longer wipes the U1/U2 parked-caret hint.

2. **Suppressing load_text's own sync is safe for every other caller.** The flag defaults `False` and is raised *only* inside `_set_text_and_caret`. Every external `load_text` caller — held-prompt restore (app.py:2428), history navigation (3888), draft/begin_model_query (9460), question/reload restore (9548, 11689) — calls the public `load_text` with the flag `False`, so each still gets its own `_sync_picker`. No other path depends on the suppressed sync.

3. **Flag can't get stuck True.** It's toggled inside a `try`/`finally`; even if `self.text =` or `move_cursor` raised, `finally` restores `False`. Confirmed there is no early return or re-entrancy that leaves it set.

4. **Init order is correct.** `self._suspend_picker_sync = False` is set BEFORE `super().__init__()` (editor.py:711 vs 714), which matters because TextArea's constructor loads the initial document through `load_text` → the guarded sync. Reading the attribute during construction would otherwise `AttributeError`.

5. **The new test is a genuine real-app assertion that fails pre-fix.** `test_completing_a_team_name_keeps_the_parked_hint_in_the_real_app` drives `OperatorApp` (not the bespoke harness whose `on_argument_query_opened` doesn't clear the notice), fills `/team security ` via the list + Tab, and asserts the switch/send hint survives. I reverted editor.py to 5b9f2ef3 and ran it: it **FAILS** with `assert ('switch' in '')` (notice empty — D5 reproduced); with the fix it PASSES. This closes exactly the "green tests aren't the rendered frame" gap round 3 named.

**Tests / gates**
- `test_command_picker.py`: 256 passed
- `test_slash_echo.py`: 27 passed (incl. the new D5 test)
- `test_app_pilot.py`: 166 passed
- Round-1/2/3 behaviours (reassembly, D1, nested-slash, non-prompt splice-run, start-of-line fill-and-park) remain green in the above suites.
- Ruff `check`/`format` report issues at test_slash_echo.py:506 and :949 — both PRE-EXISTING on `main` and outside the D5 delta (new test is ~L407-450); not introduced by this PR and not a merge blocker for this round.

No BLOCKER, no MAJOR, no new issue introduced by the guard.

**Verdict: clean. This round is TERMINAL on the code side.** The two pre-existing ruff findings are follow-ups, not a reason for another round.

PR #251, opened 2026-08-22
perf(ci): parallelize the test suite to bring CI under 10 minutes

damianvtran:
## Why

The CI `test` job ran `coverage run -m pytest -v` **fully serially**, taking ~28 minutes per Python version. Every PR — even a docs-only one — waited that long on the matrix. Goal: bring the suite under 10 minutes.

## What makes it slow

Profiling the suite (5861 tests) surfaced three levers, in order of impact:

1. **No parallelism.** The suite is overwhelmingly **wait-bound**, not compute-bound: the ~1600 Textual pilot (TUI) tests spend their wall time letting the event loop settle, not on CPU. A representative subset measured **205s wall against 40s user CPU**. The `test` job used one process.
2. **Fixed `pilot.pause(<seconds>)` sleeps.** ~280 call sites across the TUI tests slept real wall-clock time (`pause(0.1)`…`pause(0.4)`) instead of waiting for the event loop to go idle. Slower, and a timing bet that loses under parallel CPU contention.
3. **`addopts = "-vv -s"`.** `-vv` prints ~5.8k lines of per-test I/O; `-s` disables capture and is incompatible with xdist.

## What changes

**Infrastructure (the biggest win):**
- Add `pytest-xdist`; default `addopts` to `-n auto` (one worker per core, 4 on `ubuntu-latest`).
- Drop `-vv -s`.
- Switch the CI test step from `coverage run -m pytest` to `pytest --cov` (collects coverage inside each xdist worker and merges it; a top-level `coverage run` sees nothing from worker subprocesses). Move the 80% floor into `[tool.coverage.report] fail_under`.

**Test optimizations (parallelized across four workstreams, aggregated here):** replace the fixed `pilot.pause(<seconds>)` sleeps in the heaviest TUI files with bounded condition-waits / idle-waits, which lowers the slowest-worker ceiling and removes parallel-load flakiness. Files: `test_steering_approval.py` (138 pauses), `test_ask_picker.py` (35), `test_working_line.py` (61), `test_app_pilot.py`, `test_composer_seam.py`, `test_welcome.py`, `test_approvals_ux.py`. No assertion was weakened and no product source was touched.

Notable correctness work along the way:
- Fixed `test_esc_aborts_a_running_shell_command`, which raced under parallel load: the abort handler marks the card and keeps ownership synchronously on the key press, but a `pause()` after it yielded to the shell worker, which reaped the process and cleared the card before the ownership assertion ran. The assertion now reads the on-press frame (strictly stronger).
- Added `_settled_rows()` for the width-parametrized approval tests, which assert on a row's **reflowed geometry**: a single idle tick could read a mid-reflow frame with a truncated `…`. This surfaced as a real CI failure, not just local load.
- Kept `test_a_held_key_never_answers_a_question_the_card_moved_on_from` on fixed pauses on purpose: it probes a real ~180 ms timer (`ANSWER_KEY_HOLD_S`), and "before a timer fires" is not expressible as an idle-wait. Over-converting it made it commit a stale key under contention.

**Worker count:** `-n auto` (= one per core). An `-n 8` oversubscription experiment is in the history and was reverted: on the 4-vCPU runner it did not reduce wall time and it starved three wall-clock-sensitive tests. One worker per core is the stable point.

## Benchmarks

Local (14 cores), full suite:

| Config | Wall time | Result |
| --- | --- | --- |
| Serial (old CI shape) | **1097s** | 5861 passed |
| `-n auto` | **~85s** | 5861 passed |
| `-n 4` (CI-realistic) | **~248s** | 5861 passed, deterministic under `-n 4` + 6 CPU burners |

Authoritative **CI** timing (this PR, `pytest --cov` step, GitHub `ubuntu-latest`):

| Job | Before (serial) | After |
| --- | --- | --- |
| `test (3.12)` | ~1600s (~27m) | **558s (9.3m)** ✅ |
| `test (3.13)` | ~1690s (~28m) | **568s (9.5m)** ✅ |

Both under the 10-minute goal; ~3x faster. Coverage unchanged at **83.74%** (gate 80%).

## Change class

C2 — CI/tooling and test-harness change, standard/pre-authorized. No product runtime path; nothing under `local_operator/` changed.

## Testing evidence

This is a test-tooling change, so the evidence is the suite itself running green under the new configuration and the measured wall-time reduction. The authoritative run is CI's own matrix on this PR's head (`5528527b`): both `test (3.12)` and `test (3.13)` green, 558s / 568s, coverage 83.74% ≥ 80. Locally the full suite passes 3/3 under `-n 4` with 6 CPU burners (a proxy for CI's slower cores), which is how the two parallel-load races above were found and fixed before they could reach `main`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,21 +81,19 @@ jobs:
           pip install -e ".[dev]"
 
       - name: Run tests with coverage
+        # pytest-cov (not `coverage run -m pytest`) because the suite now runs
+        # under xdist: coverage has to be collected inside each worker
+        # subprocess and merged, which pytest-cov does automatically and a
+        # top-level `coverage run` cannot. `-n auto` comes from addopts; the
+        # 80% floor is enforced by `fail_under` in [tool.coverage.report], so
+        # the step no longer parses the percentage out of the report by hand.
         run: |
-          coverage run -m pytest -v
-          COVERAGE_PERCENTAGE=$(coverage report | awk '/TOTAL/ {print $NF}' | tr -d '%')
-          echo "Coverage Percentage: $COVERAGE_PERCENTAGE"
-          if (( $(echo "$COVERAGE_PERCENTAGE >= 80" | bc -l) )); then
-            echo "Coverage check passed!"
-          else
-            echo "Coverage check failed. Minimum coverage is 80%"
-            exit 1
-          fi
-
-      - name: Generate coverage HTML report
-        run: coverage html
+          pytest --cov --cov-report=term-missing:skip-covered --cov-report=html
 
       - name: Upload coverage report
+        # `if: always()` so a coverage-threshold failure still uploads the HTML,
+        # which is exactly when you want to see which lines dropped.
+        if: always()
         uses: actions/upload-artifact@v4
         with:
           name: coverage-report-${{ matrix.python-version }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,21 +84,20 @@ jobs:
         # pytest-cov (not `coverage run -m pytest`) because the suite now runs
         # under xdist: coverage has to be collected inside each worker
         # subprocess and merged, which pytest-cov does automatically and a
-        # top-level `coverage run` cannot. The 80% floor is enforced by
-        # `fail_under` in [tool.coverage.report], so the step no longer parses
-        # the percentage out of the report by hand.
+        # top-level `coverage run` cannot. `-n auto` (from addopts) gives one
+        # worker per logical core — 4 on GitHub's ubuntu-latest. The 80% floor
+        # is enforced by `fail_under` in [tool.coverage.report], so the step no
+        # longer parses the percentage out of the report by hand.
         #
-        # `-n 8` OVERRIDES the `-n auto` default (which would be 4 on this
-        # 2-core-hyperthreaded runner). The suite is wait-bound — pilot tests
-        # spend their wall time letting the event loop settle, not on CPU — so
-        # running 2x as many workers as cores overlaps that idle time and cuts
-        # wall clock further without saturating the CPU. Measured locally,
-        # `-n 8` ran the suite in roughly half the `-n 4` wall time. It is a
-        # deliberate constant, not `auto`, because the right degree of
-        # oversubscription is a property of THIS suite's wait profile, not of
-        # the runner's core count.
+        # Do NOT oversubscribe (`-n` greater than the core count) here. It looks
+        # tempting because the suite is wait-bound, and it does help on a
+        # many-core dev box that still has idle cores at `-n 8`. On this 4-vCPU
+        # runner it does not: it left wall time unchanged and starved three
+        # timing-sensitive tests (a parked job, a late-parking tool, a held
+        # key) whose deadlines are real wall clock, turning them red. One worker
+        # per core is the stable point.
         run: |
-          pytest -n 8 --cov --cov-report=term-missing:skip-covered --cov-report=html
+          pytest --cov --cov-report=term-missing:skip-covered --cov-report=html
 
       - name: Upload coverage report
         # `if: always()` so a coverage-threshold failure still uploads the HTML,
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,11 +84,21 @@ jobs:
         # pytest-cov (not `coverage run -m pytest`) because the suite now runs
         # under xdist: coverage has to be collected inside each worker
         # subprocess and merged, which pytest-cov does automatically and a
-        # top-level `coverage run` cannot. `-n auto` comes from addopts; the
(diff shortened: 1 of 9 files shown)


damianvtran:
### Agent review — round 1
- Reviewer: reviewer subagent (Claude Sonnet 4.5)
- Scope: 2f06901b14ec374da8792ab8c4c9863e8fcde686..5528527be1e5e3bda3428f0c2b87b0fbed43265e
- Mode: Comprehensive
- Delivery contract: C2 CI/test-tooling — parallelize tests under 10min via xdist; coverage gate preserved; no assertion weakened; no product source changed
- Applicable risk surfaces: coverage-gate integrity, test correctness under parallel load, timing-race regressions, CI workflow correctness
- Verdict: Approve

| --- | --- | --- | --- |
| R1-01 | Info | Coverage-gate integrity (AC1) | Verified empirically that `pytest --cov` honors `[tool.coverage.report] fail_under=80` and exits non-zero below the floor (reproduced: a 62.5% package returned exit 1, "Required test coverage of 80.0% not reached"). No `--cov-fail-under` flag is needed. The denominator narrowing from local_operator+tests (old bare `coverage run -m pytest`) to `source=["local_operator"]` does **not** lower the bar — it raises it: test files execute at ~100% and were inflating the old TOTAL, so applying the 80% floor to product code alone is a stricter, more meaningful gate. PR reports 83.74% on local_operator only. Correct. |
| R1-02 | Info | Timing-race regression (AC2, 18dd6d3f) | The `_settled_rows` helper and its five callsites are correct. A single `pilot.pause()` yields one idle tick and can read a mid-reflow truncated frame; polling for two consecutive identical row-sets waits on the exact condition the width/truncation geometry assertions depend on. Ceiling=200 is a pure deadlock guard (returns last frame → assertion fails naturally, never hangs). Reasoning sound. Stress-verified: test_steering_approval.py + test_working_line.py 133 passed 3/3 under `-n 4` with 6 CPU burners. |
| R1-03 | Info | Timing-race regression (AC2, 5528527b) | The held-key revert is correct. `test_a_held_key_never_answers_a_question_the_card_moved_on_from` probes a real ~180ms `ANSWER_KEY_HOLD_S` wall-clock window; `_held_answer_key` clears in BOTH the correct and buggy paths, so the earlier `_until(... is None)` poll could read the race at the wrong moment and let the stale key commit. An idle-wait cannot express "before a real timer fires" — fixed pauses are the right primitive here. The deliberate-pause comment prevents future reconversion. |
| R1-04 | Info | Same-class defect scan (AC2) | Scanned all converted files for the two defect classes (bare `pause()` before reflow-geometry assertions; idle-wait standing in for a wall-clock window). None found. Other reflow-sensitive reads are either routed through `_settled_rows` (steering_approval) or drive `_paint()` directly with no frame read (test_working_line `test_the_line_holds_one_row_...`). The remaining real wall-clock waits (`pause(DOUBLE_STOP_WINDOW_S + 0.5)` at steering_approval:2651/2681/2902) were correctly left as fixed durations. welcome.py's `_settled_welcome` forcing `_poll()` is safe: timer liveness stays pinned by the un-forced `test_model_label_polls_in_after_boot`. |
| R1-05 | Info | CI workflow correctness (AC4) | `-n auto` set in `[tool.pytest.ini_options] addopts`; no leftover `-n 8` (the only `-n 8` mentions are in an explanatory comment warning against oversubscription); `if: always()` correctly present on the coverage-upload step so a threshold failure still uploads htmlcov; `-vv -s` drop is justified (`-s` is incompatible with xdist's worker output capture, and no test asserts on uncaptured stdout). Stress-verified: ask_picker + approvals_ux + app_pilot + composer_seam + welcome, 354 passed 3/3 under `-n 4` with 6 CPU burners. |

No BLOCKER and no MAJOR findings. All four acceptance criteria verified, two of them empirically (fail_under enforcement reproduced; the six heaviest converted files run green three consecutive times under simulated CI contention). No assertion was weakened — every conversion swaps a fixed sleep for a wait on the same observable the assertion already reads, and the two over-conversions were correctly caught and reverted. No product source under local_operator/ is touched.

**This round is TERMINAL.** No further round is required on correctness grounds.

damianvtran:

Round 1 verdict is **Approve** with no `blocker`/`major`/`minor` findings, so there is nothing to remediate. Recording the dispositions for the record:

- The two parallel-load races the reviewer independently re-derived (`18dd6d3f` settled-frame reads; `5528527b` held-key fixed-pause revert) were both already fixed on this head and confirmed correct.
- The reviewer's own scan of the six other converted files for the same two defect classes found none, and its empirical stress run (six heaviest files, 3/3 green under `-n 4` + 6 CPU burners) matches the validation done before the round.
- Coverage-gate integrity (AC1): the reviewer reproduced that `fail_under=80` exits non-zero below the floor, and that scoping the denominator to `local_operator` only **tightens** the gate (test files were inflating the old total). No silent lowering.

Merging on the current head `5528527b`, which the round's `Scope:` line covers.

PR #254, opened 2026-08-22
feat(analytics): overlay estimated cost + responsive frame

damianvtran:
## Why

Follow-up to #247 (usage analytics). That screen showed **token** consumption but not **cost**, and on a large terminal the frame stayed narrow with a sea of dead margin. This overlays an estimated dollar cost on every level of the report and makes the frame responsive so a big screen actually shows more.

Version bump is **MINOR** (0.21.0 → 0.22.0): new capability on an existing screen, no breaking changes. A ledger written by 0.21.0 is migrated in place on first open.

### Cost overlay

- **Priced at record time, where the exact model is known.** `SessionStreamFn._cost_micro` (`configure.py`) prices each call through the shared `cost_for_usage` — the *same* arithmetic the status band runs every repaint — so the analytics dollar total can never disagree with the live band. Cost is stored per call, never recomputed at aggregation (where differently-priced models are already summed and un-priceable).
- **Micro-USD integers.** `CallSnapshot.cost_micro` is USD × 1e6 as an integer, so the store's `SUM` is exact — a fraction of a cent per call over millions of calls is real drift in a float.
- **Unknown ≠ free.** A model with no published price records `cost_known=False`; the report renders `$—`, never a misleading `$0.00`. A scope where *some* calls were unpriced is a lower bound: `$12.30+`. Cost is labelled an **estimate** (list price × tokens — it cannot see a plan, discount, or free tier), the same measured-vs-modelled honesty the component split already carries.
- **Where it shows:** an `Est. cost` line in TOTALS, and a cost column in the per-provider and per-session tables (which now sort by spend).

### Migration

- `cost_micro` / `cost_known` are new columns. `CREATE TABLE IF NOT EXISTS` never alters an existing table, so `AnalyticsStore._migrate` runs an idempotent `ALTER TABLE ADD COLUMN` on open. A 0.21.0 token-only ledger upgrades in place; its old rows read as unpriced (cost 0 / unknown), never NULL and never a false $0.

### Responsive frame

- `.analytics-panel` `max-width` 100 → 140 and `height` 80% → 85%; `AnalyticsScreen._card_width` tracks the CSS cap so the report's column maths matches the painted width. The per-provider/per-session tables shed the **cache** column below `_WIDE_TABLE_MIN` (72) to keep the **cost** column at any width — the same drop-ladder logic the status band uses.

## Impact

- **No breaking changes.** Old databases migrate transparently; a model without a price degrades to `$—`. Analytics remains a best-effort accelerator — a pricing failure is caught and the call is still recorded token-wise.
- **Latency:** `_cost_micro` is **0.006 ms** warm (a memoised price lookup and four multiplies), and it runs *after* the stream is consumed, off the response-await path. No perceptible cost.
- **Security:** unchanged — the DB stays 0600, and cost is derived from the price table, not from any new stored secret.

## Testing Details

Evidence is proof the feature works against the real paths, not just a green suite.

### End-to-end cost through the real stream wrapper

A priced model (`claude-sonnet-4-5`), a 100k-context turn (5k fresh input, 90k cache read, 5k cache write, 1k output) recorded through `SessionStreamFn._record_stream`:

```
recorded cost_usd: 0.0757   cost_known_calls: 1   context: 100000
```

$0.0757 — priced through the shared `cost_for_usage` with correct Anthropic cache-exclusive handling, matching what the status band would show for the same turn.

### Migration of a 0.21.0 (token-only) ledger

```
migrated old DB -> calls: 1  cost_micro: 0  cost_known_calls: 0  cost_is_partial: True
cost_micro col: True   cost_known col: True
```

Columns added on open; the pre-cost row reads as unpriced (not $0); a new priced call records normally into the migrated DB.

### Cost display states

```
$8.34     complete
$8.34+    partial (some calls used an unpriced model — lower bound)
$—        nothing priceable (e.g. a local-model-only run)
$0.0042   sub-cent (kept precise so a real spend is not rounded to $0)
$1.2k     large sum abbreviated
```

### Responsive layout

Wide (150 cols) shows tokens · cost · calls · cache; narrow (84 cols) drops cache and keeps cost. Both rendered and inspected — see screenshots. Benchmarked `_cost_micro` at 0.006 ms warm.

### Automated tests

- Model: `cost_usd`, `cost_is_partial`, `cost_is_known`, snapshot carries cost.
- Store: cost roundtrip + per-provider/session grouping, **old-DB migration**, unpriced handling.
- Panel: `format_cost` states (complete/partial/unknown/sub-cent/large), the cost overlay in the report, unpriced/partial markers, and the narrow-frame cache-drop-keeps-cost.
- Stream: end-to-end cost recorded for a priced model.

58 analytics + panel tests pass; the affected model/session/clients suites (304 tests) pass. `black`, `isort`, `flake8`, `pyright` clean.

## Checklist

- [x] I have updated the documentation when required. (AGENTS.md analytics section.)
- [x] Security considerations are addressed.
- [x] I have linked all related issues. (Follows #247.)

## Screenshots

`/analytics /usage` with the estimated cost overlay.

**Wide terminal (150 cols)** — the frame grows to fit tokens · cost · calls · cache. `Est. cost $8.34+` in TOTALS (the `+` marks a lower bound — the local Ollama calls are unpriced), a cost column in both tables (`$6.10` / `$1.90` / `$—` for the unpriced local model), and per-session costs.

Wide analytics with cost

**Narrow terminal (84 cols)** — the tables drop the cache column to keep the cost column; the `Est. cost` line and its caveat still fit.

Narrow analytics with cost

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -236,9 +236,13 @@ Every provider call across every session contributes to one shared, on-disk
 ledger (`<config_dir>/analytics.db`, SQLite/WAL/0600). `/analytics /usage`
 opens an Esc-dismissable screen summarising token consumption: authoritative
 provider counts (input/output/cache, and the thinking-vs-generation split of
-output) plus an *estimated* breakdown of input across the system prompt, custom
-instructions (agent/team profiles), tool inventory, tool schemas, environment,
-knowledge, conversation, and tool results.
+output), an *estimated* dollar **cost** overlay, and an *estimated* breakdown of
+input across the system prompt, custom instructions (agent/team profiles), tool
+inventory, tool schemas, environment, knowledge, conversation, and tool
+results. The frame is responsive — it grows to `max-width: 140` on a large
+terminal (see the `.analytics-panel` CSS and `AnalyticsScreen._card_width`,
+which must stay in step), and the per-provider/per-session tables shed the cache
+column below `_WIDE_TABLE_MIN` to keep the cost column.
 
 Things that will bite you if you forget them:
 
@@ -272,5 +276,20 @@ Things that will bite you if you forget them:
   and estimates must never be presented as the same kind of number — the screen
   says "estimated split of context tokens" for a reason.
 
-- **Adding a component is a schema migration.** `COMPONENT_KEYS` maps to one
-  `c_<key>` column each in `store._SCHEMA`; bump the schema and default 0.
+- **Cost is computed at RECORD time, not aggregation time.** Dollar cost is
+  priced in `SessionStreamFn._cost_micro` (`configure.py`) where the exact model
+  is known, via the shared `cost_for_usage` — so analytics can never disagree
+  with the status band. It is stored per call as `cost_micro` (USD × 1e6,
+  INTEGER so the `SUM` is exact) plus a `cost_known` flag. A model with no
+  published price records `cost_known=False` (rendered `$—`, never `$0.00`); a
+  scope where some calls were unpriced is a LOWER BOUND (rendered `$12.30+`).
+  Cost is an ESTIMATE (list price × tokens; it cannot see a plan, discount, or
+  free tier) and is labelled as one, same discipline as the component split.
+
+- **Adding a component OR a stored column is a schema migration.**
+  `COMPONENT_KEYS` maps to one `c_<key>` column each in `store._SCHEMA`. A
+  database from an older release is upgraded on open by `AnalyticsStore._migrate`
+  (idempotent `ALTER TABLE ADD COLUMN`, since `CREATE TABLE IF NOT EXISTS` never
+  alters an existing table) — this is how the `cost_*` columns reach a
+  token-only ledger. Any new stored column needs a `_MIGRATION_COLUMNS` entry
+  with a DEFAULT so old rows read as a sane value, never NULL.
--- a/local_operator/analytics/__init__.py
+++ b/local_operator/analytics/__init__.py
@@ -42,6 +42,7 @@ from local_operator.analytics.model import (
     CallSnapshot,
     UsageAggregate,
     apportion_components,
+    price_snapshot,
     snapshot_component_chars,
     split_system_prompt,
 )
@@ -59,6 +60,7 @@ __all__ = [
     "CallSnapshot",
     "UsageAggregate",
     "apportion_components",
+    "price_snapshot",
     "snapshot_component_chars",
     "split_system_prompt",
     "AnalyticsRecorder",
--- a/local_operator/analytics/model.py
+++ b/local_operator/analytics/model.py
@@ -156,6 +156,18 @@ class CallSnapshot:
     # input tokens and is worth recording, but is counted separately so a
     # provider-outage session does not look like normal spend).
     ok: bool = True
+    # This call's dollar cost in MICRO-USD (USD × 1_000_000), and whether it is
+    # known at all. Cost is computed AT RECORD TIME, where the exact model that
+    # served the call is known — never at aggregation, where tokens from
+    # differently-priced models are already summed and un-priceable. Micro-USD
+    # integers rather than floats so the store's ``SUM`` is exact (a fraction of
+    # a cent per call over millions of calls is real money to drift). A model
+    # with no published price records ``cost_known=False`` and ``cost_micro=0``,
+    # which the report renders as an unknown share of the total rather than a
+    # confident ``$0.00`` — the same "unknown ≠ free" distinction the status
+    # band's ``$—`` already makes (see ``tui/costs.turn_cost``).
+    cost_micro: int = 0
+    cost_known: bool = True
 
 
 def snapshot_component_chars(request: Any) -> dict[str, int]:
@@ -267,6 +279,12 @@ class UsageAggregate:
     cache_write_tokens: int = 0
     reasoning_tokens: int = 0
     context_tokens: int = 0
+    # Summed dollar cost in MICRO-USD, and how many of ``calls`` had a known
+    # price. ``cost_known_calls < calls`` means the dollar figure is a LOWER
+    # BOUND — some calls used models with no published price — which the report
+    # marks (``$12.30+``) rather than presenting a partial sum as complete.
+    cost_micro: int = 0
+    cost_known_calls: int = 0
     components: dict[str, int] = field(default_factory=lambda: {k: 0 for k in COMPONENT_KEYS})
     by_provider: dict[str, "UsageAggregate"] = field(default_factory=dict)
     by_session: dict[str, "UsageAggregate"] = field(default_factory=dict)
@@ -275,6 +293,30 @@ class UsageAggregate:
     def generation_tokens(self) -> int:
         """Output tokens that were visible generation, not hidden thinking."""
         return max(0, self.output_tokens - self.reasoning_tokens)
+
+    @property
+    def cost_usd(self) -> float:
+        """The summed cost in whole dollars (from the micro-USD accumulator)."""
+        return self.cost_micro / 1_000_000.0
+
+    @property
+    def cost_is_partial(self) -> bool:
+        """True when some priced-in calls had no published price.
+
+        The dollar total is then a lower bound: it counts every call we could
+        price and silently omits the ones we could not. A report must SAY so
+        rather than imply the omitted calls were free.
+        """
+        return self.cost_known_calls < self.calls
+
+    @property
+    def cost_is_known(self) -> bool:
+        """True when at least one call in scope had a known price.
+
+        ``False`` means nothing here is priceable (e.g. a local Ollama-only
+        run), which the report renders as ``$—`` rather than ``$0.00``.
+        """
+        return self.cost_known_calls > 0
 
     @property
     def total_tokens(self) -> int:
--- a/local_operator/analytics/model.py
+++ b/local_operator/analytics/model.py
@@ -156,18 +156,55 @@ class CallSnapshot:
     # input tokens and is worth recording, but is counted separately so a
     # provider-outage session does not look like normal spend).
     ok: bool = True
-    # This call's dollar cost in MICRO-USD (USD × 1_000_000), and whether it is
-    # known at all. Cost is computed AT RECORD TIME, where the exact model that
-    # served the call is known — never at aggregation, where tokens from
-    # differently-priced models are already summed and un-priceable. Micro-USD
-    # integers rather than floats so the store's ``SUM`` is exact (a fraction of
-    # a cent per call over millions of calls is real money to drift). A model
-    # with no published price records ``cost_known=False`` and ``cost_micro=0``,
-    # which the report renders as an unknown share of the total rather than a
-    # confident ``$0.00`` — the same "unknown ≠ free" distinction the status
-    # band's ``$—`` already makes (see ``tui/costs.turn_cost``).
+    # This call's dollar cost in MICRO-USD (USD × 1_000_000) and whether it is
+    # known. Populated by the STORE on its background thread
+    # (``price_snapshot``) unless ``priced`` is already True: pricing calls
+    # ``resolve_model_info``, which can block for seconds on a cold memo, and
+    # that must never run on the event loop the turn is unwinding on (review
+    # C1). Micro-USD integers rather than floats so the store's ``SUM`` is exact.
+    # A model with no published price prices to ``cost_known=False`` /
+    # ``cost_micro=0``, rendered as an unknown share rather than a confident
+    # ``$0.00`` — the "unknown ≠ free" distinction the status band's ``$—``
+    # already makes (see ``tui/costs.turn_cost``).
     cost_micro: int = 0
-    cost_known: bool = True
+    cost_known: bool = False
+    # ``True`` when ``cost_micro``/``cost_known`` are already final and the store
+    # must NOT re-price (a test recording a known cost without a price table, or
+    # a future caller that priced off-thread itself). The normal recording path
+    # leaves this False so the store prices from the model + token counts.
+    priced: bool = False
+
+
+def price_snapshot(snapshot: "CallSnapshot") -> tuple[int, bool]:
+    """``(cost in micro-USD, price known)`` for a snapshot, never raising.
+
+    Called by the store on its BACKGROUND thread (never the event loop), where
+    the potentially-blocking ``resolve_model_info`` cold-memo lookup is free —
+    the same reason the SQLite write lives there. Prices through
+    ``cost_for_usage``, the one money computation the whole app shares, so the
+    analytics dollar total cannot diverge from the status band. Returns
+    ``(0, False)`` for a model with no published price (rendered ``$—``, not a
+    misleading ``$0.00``) and for any failure — pricing is best-effort, and an
+    unpriceable call is still worth recording token-wise.
+
+    The lazy import keeps the model layer off this module's import graph (the
+    analytics package must stay importable without the provider/pricing stack)
+    and matches how the recorder already defers its own imports.
+    """
+    if snapshot.priced:
+        return int(snapshot.cost_micro), bool(snapshot.cost_known)
+    try:
+        from local_operator.model.configure import cost_for_usage, resolve_model_info
+
+        info = resolve_model_info(snapshot.provider, snapshot.model_id)
+        if not (info.input_price or info.output_price):
+            return 0, False
+        # ``cost_for_usage`` duck-types its usage arg, so a plain object with the
+        # token fields it reads is enough — no need to rebuild a ``Usage``.
+        dollars = cost_for_usage(snapshot.provider, info, snapshot)
+        return int(round(dollars * 1_000_000)), True
+    except Exception:  # noqa: BLE001 — an unpriceable call is not an error
+        return 0, False
 
 
 def snapshot_component_chars(request: Any) -> dict[str, int]:
--- a/local_operator/analytics/store.py
+++ b/local_operator/analytics/store.py
@@ -39,6 +39,7 @@ from local_operator.analytics.model import (
     CallSnapshot,
     UsageAggregate,
     apportion_components,
+    price_snapshot,
 )
 from local_operator.paths import config_dir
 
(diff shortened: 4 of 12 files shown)


damianvtran:
### Design review — round 1
- Reviewer: designer subagent (claude-sonnet-4-5)
- Scope: 25b9a23bcd97443634af33f4e5580d80099bf0fd..055f503c9b18ea78e0a5f0cbe813ee39f8bcb2c4
- Verdict: Changes requested

Reviewed the two rendered frames (`01-wide.png` @ 150 cols, `02-narrow.png` @ 84 cols) plus `build_report` / `format_cost` / `_group_section`. The base-screen approvals from round #247 are all preserved: `≈`/`~` estimate markers legible in `dim` (not `faint`), 1-cell minimum bars, cache-rate line intact, `$—`↔status-band consistency. Cost is well-integrated as a headline number — full-strength `fg` next to tokens, `dim` for calls/cache — and the sort-by-cost ordering is a good call. The three-state honesty (`$8.34` / `$8.34+` / `$—`) is the right model. My one MAJOR is that the surface doesn't *teach* two of those states.

| ID | Severity | Finding |
|----|----------|---------|
| D1 | major | The `+` (lower-bound) and `$—` (unknown) markers carry no on-screen legend, and the `+` renders in the same full-strength `fg` as the digits, so it reads as part of the number rather than a status flag. The TOTALS note `≈ list price × tokens` explains the estimate basis but never the `+`. A first-time reader sees `$5.02+` / `$8.34+` on a bill-shaped number and cannot decode it as "lower bound — some calls used an unpriced model"; the worst misread is treating a lower bound as the complete spend. This is the crux of the feature's "unmistakable three states" goal. Fix cheaply: one `dim` footnote under the tables (or extend the TOTALS note), e.g. `+ lower bound · $— unpriced`, and/or give the `+` a subtle weight/color distinction from the digits. |
| D2 | minor | BY PROVIDER and BY SESSION don't share a column grid. `name_col` is computed per table (provider max `ollama (local)`=14 → ~15; session max `analytics cost overlay`=22 → ~23), so the tokens/cost/calls columns start at different x-positions between the two stacked tables — visible in both shots as BY SESSION's numbers sitting right of BY PROVIDER's. Two adjacent tables of the same shape read as ragged. Consider a shared `name_col` (max across both groups) so the columns line up vertically down the panel. |
| D3 | minor | The wide frame doesn't use the extra width for the tables — only the bars grow (`bar_width` up to 24). At 150 cols the card is ~129 wide but the fixed-column tables leave a large dead gutter to the right (roughly the right third of the panel is empty in `01-wide.png`). "Grows on large terminals" currently reads as a stretched card with left-anchored content rather than intentional use of space. Widening the name/number columns, or adding a right-hand element, would make the growth feel deliberate. |
| D4 | nit | Rendered numbers don't fully reconcile: TOTALS `$8.34+` matches BY SESSION (`$5.02+$2.09+$1.23=$8.34`) but not BY PROVIDER's priced sum (`$6.10+$1.90=$8.00`, ollama `$—`). Almost certainly fixture data for the screenshot, but an observant user doing mental math on the provider table gets a different total — worth confirming the real aggregation reconciles the priced provider sum against the headline. |

Narrow-frame cache-drop (item 4): reads as intentional — tables stay clean, cost survives, and the aggregate `78% cache` line in TOTALS preserves the headline cache signal, so nothing important is lost. Column alignment (item 3): right-alignment of cost holds and `$—`/`+` don't break the columns (`cost_col` sizes to the widest rendered figure). Hierarchy (item 5): balanced — cost neither hides nor overwhelms tokens. No BLOCKER.

**Blocker: 0 · Major: 1** — one MAJOR (D1) stands between this and approval; it's a cheap, additive fix. D2–D4 are follow-ups.

damianvtran:
### Agent review — round 1
- Reviewer: reviewer subagent (Claude Sonnet 4.5)
- Scope: 25b9a23bcd97443634af33f4e5580d80099bf0fd..055f503c9b18ea78e0a5f0cbe813ee39f8bcb2c4
- Mode: Comprehensive
- Applicable risk surfaces: schema migration, cost correctness, event-loop latency, error isolation, responsive layout
- Verdict: Changes requested

Reviewed the full diff, read the surrounding pricing/resolution/recorder machinery, and ran the suite: `58 passed in 2.38s`. Cost arithmetic, positional aggregate indexing, micro-USD integer sums, the unknown/partial/`$—` rendering, and the migration's no-corruption guarantee all check out. One MAJOR: the record-time price lookup runs synchronously on the shared event loop and can block it, which the rest of the codebase goes out of its way to avoid.

|----|----------|------------------|------------------------------|
| C1 | major | Event-loop latency — `_cost_micro` resolves pricing synchronously on the loop | **Cause:** `_record_usage` runs in `_record_stream`'s `finally` on the event loop, and calls `SessionStreamFn._cost_micro` → `resolve_model_info` (`configure.py:2872`, `:2900-2905`) inline before enqueuing. **Violated invariant:** the loop that serves the parent turn and every sibling subagent must not make a blocking network call (`subagent_panel.py:216-230` documents this exact hazard and takes pricing *off-thread* for it, measuring "0.007 ms warm … 222 ms cold, and 13 s against a slow host for a model the shipped registry does not describe"). **Failure path:** the diff's own comment claims "a dict hit and four multiplies … no perceptible latency," but that holds only for a **warm** memo. On a memo miss — a TTL-bucket rollover at 00:00 UTC (`configure.py:1367`, 24 h buckets), `lru_cache(maxsize=64)` eviction under a fan-out, or a subagent launched on a `model_spec` override the registry doesn't describe (the panel calls this "not the exotic case … the motivating one") — `resolve_model_info` enters `_info_from_discovery`/`_from_aggregator_catalogue` and blocks the loop for up to ~13 s at that child's turn-end, freezing the keyboard for the whole app. `preflight_usage` does **not** warm this memo, so the stream path gives no guarantee it is hot. **Minimal fix:** price on the recorder's background thread instead of on the loop — carry `model`+`usage` into the snapshot path and call `_cost_micro` inside `record_batch`/the recorder run loop (same thread that already does the SQLite write) — or, if it must stay inline, do a memo-only lookup and record `cost_known=False` on a cold miss rather than resolving synchronously. |
| C2 | minor | Error isolation — a failed `ALTER` does not "degrade to the pre-migration shape" | **Cause:** `_migrate` swallows a failed `ALTER TABLE ADD COLUMN` and leaves the older schema (`store.py:274-279`), but the surrounding code is *unconditional* on the columns existing: `_INSERT_SQL` always lists `cost_micro`/`cost_known` (`store.py:104-106`, `_row_values` `:159-160`) and `aggregate()` always `SUM(cost_micro), SUM(cost_known)` (`:447`). **Failure path:** if `PRAGMA table_info` succeeds but the `ALTER` then fails transiently (lock lost after the check, read-only mount, disk pressure) the columns stay absent, so every subsequent `INSERT` raises → `record_batch` returns 0 (all recording silently dropped, `:330-336`) **and** the aggregate query raises → returns an empty `UsageAggregate` (`:464-466`) → the whole `/analytics` screen reads "no data" even though token history exists. That is a silent *total* analytics outage, not the graceful token-only degradation the docstring (`:266-267`) claims. No data is corrupted or lost, and the trigger is narrow (ALTER rarely fails while the DB is otherwise writable), which is why this is minor rather than major — but the claimed invariant is not upheld. **Minimal fix:** either gate the cost columns in `_INSERT_SQL`/`base_cols` on a checked schema version, or set `self._broken`/fall back to a cost-free column set when migration leaves the table short — and soften the docstring to match. |

#### Follow-up observations
- **`_card_width` floor vs painted content box (minor/pre-existing).** `_card_width` floors at `max(40, card-6)` (`analytics_panel.py:434`) while the CSS card is `width: 90%` with no `min-width` (`local_operator.tcss:1292`). Below ~52 cols the report is built for 40 columns but the padded content box is narrower (e.g. a 40-col terminal → card 36 → content ~32), so rows can overrun/wrap. The old code had the same 40-floor, so the responsive rewrite didn't introduce it, but this was the moment to add a `min-width` or lower the floor. Not gating.
- **`format_cost` k-suffix munging (nit).** `f"${usd/1000:.1f}k".replace(".0k", "k")` (`analytics_panel.py:114`) is correct for the tested values but is fragile string-patching; a small helper that formats then trims a trailing `.0` would read better. Not gating.
- **Verified clean:** positional indices in `_aggregate_from_row` match `base_cols` order exactly (cost sums at 8/9, components shifted to 10+, `store.py:519-523`); the empty-row guard still holds for a real-but-unpriced scope (`values[0]`=COUNT is non-zero so the `and` short-circuits, `:500`); `int(round(dollars * 1e6))` has no realistic overflow (64-bit SQLite INTEGER caps at ~9.2e12 USD) and `SUM(cost_known)` correctly yields `cost_known_calls`; the all-unpriced scope renders `$—` not `$0.00+` (`cost_is_known` false returns early before the `+`, `:118-121`); concurrent old-binary + new-binary and new+new opens on one file stay safe (name-based INSERT + column DEFAULT 0 + idempotent per-column ALTER).

**Verdict: Changes requested — 0 blocker, 1 major (C1), 1 minor (C2).** Not terminal; a round 2 is needed to clear C1.

damianvtran:
- Reviewed head: 8a19f2218a7e33d2d643d3258d2d821cc1ab0958 (was 055f503c)

- **C1 (major) — Fixed in `8a19f221`:** cost is no longer priced on the event loop. `SessionStreamFn._cost_micro` is removed; pricing now runs on the recorder's **background writer thread** via `price_snapshot` (`analytics/model.py`), called from the store's `_row_values` right next to the SQLite write. `resolve_model_info`'s cold-memo blocking can no longer touch the turn's loop — the same off-thread discipline `subagent_panel` uses. `CallSnapshot` gains a `priced` flag so a caller that already priced off-thread (or a test) can supply a final cost; the normal path leaves it False and the writer prices from the model + token counts. It is still the same `cost_for_usage`, so the total cannot disagree with the status band. New test `test_records_cost_for_a_priced_model` exercises the writer-thread pricing end to end.
- **C2 (minor) — Fixed in `8a19f221`:** a failed cost-column migration no longer drops all recording. `_migrate` sets `_has_cost` from the columns actually present afterward; `record_batch` uses a cost-less INSERT and `aggregate` substitutes `0,0` cost sums when it is False — so token analytics keep working and cost renders `$—` instead of every write failing and blanking the screen. New test `test_recording_degrades_when_cost_columns_absent`.

No blocker was raised. Requesting a round-2 convergence review of the current head.

damianvtran:
- Reviewed head: 8a19f2218a7e33d2d643d3258d2d821cc1ab0958 (was 055f503c)

Re-captured screenshots (wide + narrow) are attached to the PR description.

- **D1 (major) — Fixed in `8a19f221`:** the lower-bound `+` now renders in `dim` (a status flag, distinct from the full-strength digits) in both the TOTALS `Est. cost` line and the per-provider/per-session tables, and a `dim` legend — `+ lower bound (some calls unpriced)   $— no published price` — is drawn whenever a `+` or `$—` is actually on screen (never for a fully-priced run). A first-time reader can now decode `$8.34+` and `$—`.
- **D2 (minor) — Fixed in `8a19f221`:** BY PROVIDER and BY SESSION share one `name_col` (computed across both groups), so the tokens/cost/calls columns line up vertically down the panel instead of starting at two x-positions.
- **D3 (minor) — Fixed in `8a19f221`:** the shared name column grows on a wide frame (up to 48 cells), so the extra width widens the content rather than leaving a dead right gutter — the growth now reads as intentional. See the re-captured wide screenshot.
- **D4 (nit) — Not a code issue:** the earlier screenshot's provider-sum vs headline mismatch was fixture data. The re-captured screenshot uses reconciling numbers, and the real aggregation sums per-call `cost_micro` identically at every level, so the provider sum and the headline always reconcile for real data.

Requesting a round-2 design review of the current head with the re-captured screenshots.

damianvtran:
### Design review — round 2
- Reviewer: designer subagent (Claude Sonnet 4.5)
- Scope: 25b9a23bcd97443634af33f4e5580d80099bf0fd..8a19f2218a7e33d2d643d3258d2d821cc1ab0958
- Verdict: Approve

Verified the re-captured frames against the current head (`8a19f221`, matching the Scope head — review is fresh). Both `01-wide.png` (150 cols) and `02-narrow.png` (84 cols) reviewed.

**Round-1 findings — all resolved:**

- **D1 (major) — resolved.** The `+` lower-bound marker now renders dim on the TOTALS `Est. cost $8.34+` and on every BY SESSION row (`$5.02+` / `$2.09+` / `$1.23+`); the `$—` unknown-price marker is dim on the `ollama (local)` row. The legend `+ lower bound (some calls unpriced)   $— no published price` is drawn dim, one blank line below the tables — it reads as a footnote, not as a table row. The dim `+` stays legible against the background.
- **D2 (minor) — resolved.** BY PROVIDER and BY SESSION share the column grid: name / tokens / cost / calls (/ cache) align vertically across both tables at both widths.
- **D3 (minor) — resolved (mechanism).** The shared name column can grow on a wide frame. See follow-up F1 below.
- **D4 (nit) — resolved.** Narrow drops the cache column while keeping cost; the numbers reconcile between the two frames.

**No new visual regression.** Dim `+` legible; legend reads as a footnote; the wider name column keeps a clear gap before the tokens column (no crowding); alignment holds at 150 and 84 cols. Narrow overflow handled via the scrollbar with cache correctly dropped.

**Follow-ups (non-blocking):**
- **F1 (nit)** — the name-column-grows fix only triggers when a name approaches the 48-char cap. This fixture's longest name (`analytics cost overlay`, ~22 chars) is well under it, so the wide (150-col) frame still shows a substantial right gutter. The mechanism is correct; consider whether wide frames with short names should also distribute leftover width (e.g. widen the numeric columns or the bar area) rather than leaving the gutter.
- **F2 (nit)** — in the narrow frame the marks (`$5.02+` etc.) are on screen but the legend sits below the fold and needs a scroll to reach. Consider pinning or surfacing the legend when a mark is visible above the fold.

**Verdict: Approve** — 0 blocker, 0 major. This round is TERMINAL; F1/F2 recorded as follow-ups.

damianvtran:
### Agent review — round 2
- Reviewer: reviewer subagent (claude-opus-4-6)
- Scope: 25b9a23bcd97443634af33f4e5580d80099bf0fd..8a19f2218a7e33d2d643d3258d2d821cc1ab0958
- Delta reviewed: 055f503c9b18ea78e0a5f0cbe813ee39f8bcb2c4..8a19f2218a7e33d2d643d3258d2d821cc1ab0958
- Verdict: Approve

Both round-1 findings are genuinely fixed, not merely claimed. Delta introduces no regressions. `env -u NO_COLOR TERM=xterm-256color .venv/bin/python -m pytest tests/unit/analytics tests/unit/tui/test_analytics_panel.py -q` → **61 passed**.

**C1 (major) — Fixed.** Pricing is now off the event loop, verified by tracing every caller:
- `SessionStreamFn` no longer prices: `_cost_micro` is deleted, and the `CallSnapshot(...)` construction drops `cost_micro=`/`cost_known=`, so the loop only builds a snapshot and enqueues.
- `record_call` → `AnalyticsRecorder.record` is a non-blocking `put_nowait` (`recorder.py:204`). `_row_values` — the sole caller of `price_snapshot` — is reached only from `record_batch` (`store.py:163`), which is called only from `_flush` (`recorder.py:172`) on the background writer thread. So `price_snapshot`, and therefore the cold-memo `resolve_model_info` lookup, can only run on the writer thread. Confirmed there is no other call site of `_row_values` or `price_snapshot`.
- `priced` short-circuit is correct: `price_snapshot` returns `(int(cost_micro), bool(cost_known))` immediately when `snapshot.priced` (model.py:76-77); the normal path leaves `priced=False` and prices from the token counts. The `_snap` test fixture sets `priced=True` so the store records the exact fixture figures instead of re-pricing a fake model.
- Duck-typing is sound: `cost_for_usage` reads token fields via `_usage_field`, which does `getattr(usage, name, 0)` for non-Mapping args (configure.py:3030). `CallSnapshot` carries `input_tokens`/`output_tokens`/`cache_read_tokens`/`cache_write_tokens`, so a plain snapshot satisfies it — no `Usage` rebuild needed. The `except Exception` guard makes it best-effort.

**C2 (minor) — Fixed.** The degraded path works end to end:
- `_migrate` now sets `self._has_cost = all(name in existing ...)` after attempting each ALTER (store.py:310), and sets it `False` on an unreadable schema (`:183`). It is an instance method (no longer `@staticmethod`) so it can write the flag.
- `_COST_VALUE_INDICES = (11, 12)` are the correct positions of `cost_micro`/`cost_known` in the `_row_values` tuple (0=ts_ms … 10=context_tokens, 11=cost_micro, 12=cost_known). When `_has_cost` is False, `record_batch` drops exactly those two from each row and uses `_INSERT_SQL_NO_COST` (name-based INSERT of the older column set).
- `aggregate` substitutes `cost_cols = "0, 0"` when `_has_cost` is False, keeping the positional contract with `_aggregate_from_row`.
- Ordering holds: `aggregate` → `_read_connection` → `_connect` → `_migrate` sets `_has_cost` before the SELECT string is built, so a first-ever read cannot reference a missing column.
- `test_recording_degrades_when_cost_columns_absent` exercises the forced-False path: 1 recorded, `input_tokens=100` survives, `cost_micro=0`, `cost_is_known=False`.

**No regression.**
- `_aggregate_from_row` positional indices still match `base_cols`: COUNT/SUM(ok)/input/output/cache_read/cache_write/reasoning/context at 0-7, the two cost sums at 8/9, components at `10 + i` (store.py:562-566). Consistent across the full, no-cost, and per-provider/per-session (`row[1:]`) call shapes.
- `price_snapshot`'s lazy `from local_operator.model.configure import ...` keeps the model/pricing stack off the analytics package import graph — no cycle; matches the recorder's existing deferred-import discipline.
- Parallel-writer safety unchanged: `_has_cost` is per-store-instance but derived from actual column presence, a property of the file; combined with name-based INSERT + `DEFAULT 0` columns + idempotent per-column ALTER, an old+new or new+new concurrent open stays safe. A stale `_has_cost` only degrades to token-only writes, never corrupts.

**Verdict: Approve — 0 blocker, 0 major.** This round is TERMINAL: no blocker or major remains, and no new minor/nit is worth raising. Remaining design-side items (D2/D3 handled in the design remediation) are outside this code round.

PR #256, opened 2026-08-22
docs(readme): add credits and acknowledgements section

damianvtran:
## Summary

Adds a **Credits and Acknowledgements** section near the end of the README (before the License section) and refreshes maintainer contact details.

- **[opencode](https://github.com/anomalyco/opencode)** — Dax Raad (`thdxr`) / Anomaly (formerly SST) team
- **[oh-my-pi](https://github.com/can1357/oh-my-pi)** — Can Bölük (`can1357`), building on [Pi](https://github.com/badlogic/pi-mono) by Mario Zechner (`mariozechner`)

Each entry links the canonical repository and names the primary/core author. A matching entry was added to the Table of Contents.

Also adds a **"A note on reuse and credit"** subsection clarifying that the MIT license permits drawing inspiration from / reusing Local Operator's code, while asking that credit be given where due — naming Damian Tran &lt;damian@gominerva.com&gt; as core contributor.

Refreshes maintainer contact to the Minerva email `damian@gominerva.com`:
- `pyproject.toml` authors
- `SECURITY.md` contact
- `CODE_OF_CONDUCT.md` enforcement email

## Verification

Docs/metadata-only change. Verified:
- All six repo/author links resolve 200; `sst/opencode` 301-redirects to canonical `anomalyco/opencode`; Pi canonical is `badlogic/pi-mono` per oh-my-pi's own README.
- Primary authors confirmed against the live repos and the local oh-my-pi upstream remote.
- No stale `damianvtran@gmail.com` / `damian@radienthq.com` references remain in Markdown or `pyproject.toml`.
- Markdown well-formed; TOC anchor `#-credits-and-acknowledgements` matches the heading.

PR #255, opened 2026-08-22
fix(compaction): bound snapcompact verbatim text edges to the context window

damianvtran:
## Why

A user's `/compact` on a large session reported "35% smaller" and then immediately said **"nothing left to compact"** while the context sat at ~278k tokens — a compaction that ran, claimed a win, and left the session near-full with no further recourse.

Forensics on the real session transcript (`768f6891c63c`, a 1M-window Opus session) established:

- The pass that produced the 278k receipt ran on a **pre-#173 build** (the installed `lop` was rebuilt to `main` only *after* the compaction). `#173` already fixed the bulk: the same session compacts to **~65k** on current `main` (`DEFAULT_MAX_FRAMES=8`, per-provider frame pricing). So the headline 278k is not a live `main` bug.
- But the investigation surfaced a **real residual floor that `main` still has**: the archive's two **verbatim text edges** (`text_head` / `text_tail`) are `HQ_EDGE_FRAMES * shape.capacity` chars — **~35k tokens** for an Anthropic 1932px reader — held uncompressed on every replay and **untouchable by the frame-budget loop**, which only ever drops *imaged* pages.

Because that floor tracks the frame **shape**, not the window, on a small context window the two edges *alone* exceed the whole `0.5 * window` archive budget. The budget loop then drops **every** frame and the pass **still** commits an archive above its own target — the same "a pass meant to get under the line ends up above it" failure the frame budget exists to prevent, one layer down.

Measured on `main`, same archived history, archive-replay cost vs the `0.5*window` budget:

| window | budget | BEFORE (edge = 35k floor) | AFTER (window-aware) |
|--------|--------|---------------------------|----------------------|
| 32,000 | 16,000 | **34,989 — OVER (2.2×)** | 11,034 — ok |
| 48,000 | 24,000 | **34,989 — OVER** | 21,082 — ok |
| 64,000 | 32,000 | **34,989 — OVER** | 31,130 — ok |
| 96,000 | 48,000 | 45,037 — ok | 45,530 — ok (edges trimmed) |
| 128,000 | 64,000 | 60,109 — ok | 61,992 — ok (edges trimmed) |
| 200,000 | 100,000 | 75,181 — ok | 73,784 — ok (1932px edges trimmed 63k→60k) |
| 1,000,000 | 500,000 | 75,181 — ok | 75,181 — ok (unchanged) |

## What changes

`local_operator/compaction/snapcompact.py`:

[...]

PR #253, opened 2026-08-22
feat(prompt): harness self-awareness and eval digest-pipeline guidance

damianvtran:
## Summary of Changes

Two optimizations to the system prompt (`local_operator/prompts_md/system.md`), both about how an agent understands and uses the runtime it lives in.

**1. Harness self-awareness.** The prompt established the "Local Operator" persona but never tied it to the runtime or the `lop` command that starts it, so a question about "lop" / "local-operator" / "this harness" / "yourself" could read as an unknown third-party tool. The prompt now states plainly that Local Operator **is** the harness, run as `lop` (full name `local-operator`, alias `lo`), and that such questions are about this runtime and should be answered from its code and guides.

**2. `eval` digest-pipeline guidance.** Teaches the agent to collapse a multi-step data/file job into one `eval` call that prints a compact digest, instead of many tool calls whose intermediate results each land in context. Full output stays fetchable via the existing `spill://` handle, so token cost sits near the final answer while every intermediate is one `read` away for debugging. This is the cheap ~90% of the PTC token-efficiency win, with no code change.

- Change class: **C1 (standard, pre-authorized)** — prompt text + a test.

## Impact

- No breaking changes, no dependency updates. Changes model-facing prompt text; no API or interface change.
- Slight, deliberate increase in the fixed prompt size (two paragraphs) — justified: it prevents an agent misclassifying self-questions and steers pipelines toward lower total token cost.

## Testing Details

This changes prompt **content**, whose runtime path is "does the rendered system prompt actually carry the new guidance." Exercised the real render path (`render_template("system.md", {})`) and confirmed both blocks are present in the rendered output:

**Harness identity (rendered):**
```
Local Operator is the harness you are running in — the agent runtime itself, not
just a persona. Users start it with the `lop` command (the standard way to run
it; `local-operator` is the full name and `lo` an alias) ...
```

**eval digest guidance (rendered):**
```
Prefer one `eval` call that does a whole multi-step data or file job and prints a
compact digest over many separate tool calls whose intermediate results each land
in context ... written to a `spill://` handle you expand on demand with
`read spill://…` ...
```

Added two behaviour-contract tests in `tests/unit/test_prompts_api.py`:
- `test_system_md_states_harness_identity` — asserts `harness` and the command name `` `lop` `` are present (reword-tolerant, drop-intolerant).
- `test_system_md_teaches_eval_digest_pipeline` — asserts the `eval`/`spill://` guidance is present.

Gate output:
```
[...]

PR #258, opened 2026-08-22
feat(team): /team chart org-chart view with nested teams and tidy-tree layout

damianvtran:
## Why

Teams were a **flat roster** — one manager plus a list of members, no way for a member to itself be a team, and no way to *see* the structure (the only views are textual lists). This adds two interlocking capabilities:

1. **Nested teams / orgs.** A team may now contain other teams as members, recursively. An "org" is just a team of teams (director → managers → managers → workers).
2. **`/team chart [name]`** — a slick, scrollable, zoomable org-chart view: the manager at top, members beneath, nested teams expanded, drawn as a proper **tidy tree** that centers parents over children and packs subtrees without overlap.

Version bump is **MINOR** (0.21.0 → 0.22.0): new capability, fully backward compatible (existing `team.yml` files load unchanged).

Design doc: `docs/design-team-chart.md` (grounded in file:line citations, with a proven Buchheim tidy-tree prototype).

### Data model — one optional discriminator (`teams.py`)
- `TeamMember.kind: Literal["agent","team"] = "agent"`. Defaulting to `"agent"` means **every existing `team.yml` (no `kind` key) validates to exactly what it means today** — no migration, no format version bump. A `kind="team"` slot points at another team by name (name still lives in `role`, so `member_names()`/`roster_lines()` keep working).
- `parse_members` gains a `team:` token prefix (`team:pod`, `team:pod:2`); bare tokens stay `kind="agent"`. `roster_lines()` badges team slots `(team)`. `MAX_ORG_DEPTH = 8`.
- `team` tool: the `team:` grammar is surfaced in `TeamParams.members` description; `_op_show` reveals nesting via the badged roster. No schema break (`members` stays `list[str]`).

### Pure resolver (`org_chart.py`)
- `OrgNode` + `resolve_org(name, teams, agents) -> OrgNode`: **cycle-safe** via a `visited` frozenset, **total** (never raises on bad data), bounded by `MAX_ORG_DEPTH`. Every edge case falls out of one algorithm: 1-agent, empty team, team-under-team, org-within-org, `count>1`, cycles (A→B→A returns a `cycle` leaf, no recursion), missing team/agent refs (dim ghosts), and name-is-both-team-and-agent (the author's `kind` is authoritative).
- Name→kind classification is factored out of `session._resolve_profile_or_specialist` into a shared `classify_name`/`resolve_profile_or_specialist` in `agent_profiles.py` — **one classifier, both callers, no drift** (the session resolver now delegates to it; 633 backend tests confirm behaviour is preserved).

### Pure layout + renderer (`tui/org_layout.py`, `tui/org_render.py`)
- **Buchheim/Walker linear-time tidy tree** with variable node widths (no Textual import; testable in isolation). Parent-centering and sibling non-overlap are asserted as invariants on 200 random trees.
- Renderer: three **zoom tiers** (outline / standard / detailed), elbow-bus connectors, per-team boundary rules, and ghost/cycle/depth/count badges.

### Full-page mode + wiring (`tui/widgets/org_chart_view.py`, `app.py`, `.tcss`)
- `OrgChartView` modelled on `SubagentView`: replaces the transcript region, **keeps the greyed dock** (band/status/composer visible, read-only), so it reads as the same app in a different mode. Both-axes scroll (the chart is wide — differs deliberately from the vertical-only subagent view).
- Keys: `+`/`-` zoom, `e` expand/collapse a focused subtree, `f` fit-to-width, arrow/page scroll that **clamps** (canvas, not a list), `esc` to leave. Wired into the same Esc-precedence chain and the approval/ask/clear yield points as the subagent view. Class-identified (the `DuplicateIds`-on-fast-reopen lesson).

### Routing / collision — the crux (`app.py`, `editor.py`)
Per the explicit ask, the form is **`/team chart [name]`**, *not* `/team [name] chart`, so a command starting with the word "chart" never conflicts with talking to a team. `chart` is a **reserved first-argument subcommand** under `/team` (the `/mcp login|logout|reauth` two-level precedent):
- `/team chart` → chart the **currently-attached** team; `/team chart <name>` → chart that team.
- **Escape hatches for a team literally named `chart`:** `/team chart chart` charts it; `/team =chart <request>` *talks* to it (`=` is outside `_NAME_RE`, so it can never be a real name). When `chart` is typed as the first token and a team named `chart` exists, the handler prints a one-line disambiguation note.
- **The `/team <name> <request>` talk-to path is byte-for-byte unchanged.**
- **Picker:** on `/team ` the first slot offers a tagged `chart` **subcommand** row *first*, then team names — and a team literally named `chart` **still shows as its own separate row** (never hidden). `/team chart ` second slot re-offers team names, now feeding the chart.

### Scope boundary
The delegation **runtime** (a manager actually delegating to a nested team's manager) is a deliberate **separate follow-up PR**. This PR renders the *declared* org and tags team boundaries `(declared)` so the chart never implies a capability that isn't live yet.

## Testing evidence

**Gates (clean, tracked scope):** flake8 ✅ · black==26.1.0 `--check` ✅ · isort `--profile black` ✅ · pyright `0 errors, 0 warnings` ✅.

**Tests:** **288** affected TUI tests (incl. 12 new `test_team_chart`, 7 `test_org_layout`, 8 `test_org_render`, and the `test_slash_echo`/`test_command_picker` pins), **633** backend session/harness/agents tests (confirm the classifier extraction is behaviour-preserving), **52** new backend unit tests (`test_org_chart` edge-case table incl. cycle-terminates-in-bounded-time, `test_teams` incl. **a fieldless `team.yml` loads as `kind="agent"`**, `parse_members` grammar, routing dispatch table). All green.

> One suite failure — `test_eval_tool.py::test_background_streams_output_while_it_runs` — is **pre-existing and unrelated**: reproduced byte-identically on a clean `origin/main` detached worktree (a load-sensitive background-subprocess timing test). Not touched by this change.

**Rendered frames (real `OperatorApp` + `save_screenshot`, then actually viewed; geometry probes asserted — body `virtual_size ≥ size` on wide orgs, screen itself must-not-scroll, node non-overlap, parent-centering, two-frame settle). Captured across the full permutation matrix: 1-agent, empty, flat-multi, nested, org-within-org, cycle, count>1, deep (depth truncation), missing-refs, all three zoom tiers, small terminal (60×20), and the picker frames.**

Nested org (standard tier) — `org` centered over `◆ director` / `pod-a` / `pod-b ×2`, `(declared)` tag, dock greyed:
nested

Detailed tier — full expansion with kind tags, manager markers, `led by · N members`, elbow buses:
detailed

Cycle (A→B→A) — the `↩ a` cycle leaf in amber, **no hang**:
cycle

One-agent team — manager + one leaf, centered, fits any terminal:
one-agent

Org-within-org — standard tier collapses deeper teams to boxes (`e`/detailed drills in):
org-within-org

**Picker collision solved (the user's core concern)** — on `/team `, the `chart` subcommand row appears first (tagged `subcommand`), and a team *literally named* `chart` still appears as its own row beneath — no collision, nothing hidden:
picker-collision


damianvtran:
### Agent review — round 1

**Reviewer:** reviewer (lopdev team) — Claude Sonnet 4.5
**Scope:** `9310bf84c876c1c97f6e2421f45f4c9ceee9195a..5443d615be573d836cb55e5b2eeb7c8e0cfc120d` (reviewed as the three-dot PR diff `origin/main...HEAD`, 27 files — see process note P1)

Reviewed the diff, the design doc, and AGENTS.md; ran the full gate and the new/affected suites locally against the worktree's editable `.venv`.

#### Verification performed (real execution, not inferred)
- **flake8** (changed set incl. `org_chart.py`, `org_layout.py`, `org_render.py`, `org_chart_view.py`, `teams.py`, `agent_profiles.py`, `session/session.py`, `app.py`, `editor.py`, `team_tool.py`): clean.
- **black** — ran the AGENTS.md pin `uvx --from black==26.1.0 black --check`: *All done, 10 files unchanged.* (venv also carries 26.5.1; both agree.)
- **isort --check-only --profile black**: clean.
- **pyright** (new modules + changed app/editor/teams/agent_profiles/session/team_tool): `0 errors, 0 warnings`.
- **pytest**: `test_org_chart.py`, `test_teams.py`, `test_org_layout.py`, `test_org_render.py` → 39 passed; `test_team_chart.py` + `test_session_factory.py` → 97 passed.
- Reproduced the resolver/renderer edge cases by hand (below).

#### Design-decision checks — all four hold
1. **Backward compat of `TeamMember.kind`.** `kind` is `Literal["agent","team"]` with `default="agent"`; a fieldless `team.yml` validates unchanged. Covered by `test_fieldless_team_yml_loads_as_agent_kind`. ✓
2. **Resolver cycle-safety + totality.** `_team_node` carries a `path` frozenset of team ids (cycle → `"cycle"` leaf, no recursion) *and* an independent `depth > MAX_ORG_DEPTH` backstop (`"depth"` leaf). Missing team/agent → `"unresolved"`, never dropped. `resolve_org` returns an `"unresolved"` node for an unknown top name rather than raising. Verified A→B→A and depth-limit inputs terminate. ✓ (one theoretical caveat in m2)
3. **Classifier extraction is behaviour-preserving.** `Session._resolve_profile_or_specialist` now delegates to `agent_profiles.resolve_profile_or_specialist`; the moved body is line-for-line the old logic (role→specialist→seed order, same guards, same `noqa` BLE001). `classify_name` is a thin kind-only wrapper over the same resolver. ✓
4. **Routing / `/team <name> <request>` byte-for-byte.** The talk path is untouched except for the `=`-escape strip; `chart` is carved off the first token before it. `chart chart` charts a team named `chart`; `=chart` talks to it. Covered by `test_talk_path_is_unchanged_by_the_subcommand`, `test_chart_chart_charts_a_team_named_chart`, `test_equals_escape_talks_to_a_team_named_chart`. ✓ (case caveat in m1)

#### Scope boundary (delegation runtime) — honestly held
No runtime path reads `member.kind == "team"`: members still flow to the manager only as roster *text* (now `(team)`-badged in `roster_lines`), and `subagent.py:316` / `session.py:1957` are unchanged in how they stamp preambles. Nothing auto-launches a nested team. The chart tags the org `(declared)`. Boundary is real, not half-wired. ✓

---

#### Findings

**minor-1 — `chart` subcommand routing is case-sensitive while the picker/hint are case-insensitive (a collision-safety gap).**
`_cmd_team` (`app.py:2958`) reserves the subcommand with `first == "chart"` (exact), but the picker's second-slot trigger (`app.py` `_team_argument_choices`, `first.lower() == "chart"`) and the editor's hint suppression (`editor.py:1044`, `name.lower() == "chart"`) both fold case. Result: typing `/team Chart do x` — the picker offers `chart <team>` rows implying it will *chart*, and the switch/send hint is suppressed as if `Chart` were the subcommand, but the router falls through to the **talk** path and attaches a team literally named `Chart`. A team name legally may contain uppercase (`_NAME_RE` allows `A-Z`), so this is reachable, not hypothetical. One-line fix: fold case in the router too (`first.casefold() == "chart"`), consistent with `get_team_by_name`'s own `casefold`. Low frequency, so minor — but it undercuts the collision-safety the design is built around.

**minor-2 — outline tier (0) silently drops unresolved/cycle/depth *non-team* nodes, contradicting the "gap is always visible" invariant.**
`org_chart.py`'s module docstring promises a missing ref "becomes an `unresolved` node … so the gap is visible rather than silent." True at tiers 1–2, but at outline `_build_boxes` recurses only `child.kind == "team"` and folds agents into `_member_badge`, which counts only `role/specialist/seed`. Verified: a team whose sole member is a missing agent renders `[ squad ]` at tier 0 with no ghost and no badge — the gap vanishes. Same for a `cycle`/`depth` marker sub-team (its kind is `cycle`/`depth`, not `team`, so it isn't recursed). Since fit-to-width auto-selects outline for large orgs, a missing ref can be invisible exactly when the org is big. Fix option: include unresolved/cycle/depth children in the outline badge count or draw them as marker boxes.

**minor-3 — `resolve_org` totality rests on `load_seed` never raising, which is guarded by convention, not by the resolver.**
The cycle/depth/missing guards are solid, but the leaf classifier ultimately calls `resolve_profile → load_seed`, whose registry lookups are wrapped in `try/except` yet whose final `SEEDS_DIR/f"{key}.md".read_text(...)` is guarded only for `OSError`. A non-OSError from that read (or from `_profile_from_text`) would propagate out of `resolve_org`, breaking the "never raises" contract the chart depends on (it's called synchronously from `_open_org_chart_view`). Pre-existing in `resolve_profile`, not introduced here, and low-probability — flagging because the resolver's totality guarantee is now load-bearing for a UI surface. Consider a belt-and-suspenders `try/except` in `_agent_node` around `classify_name`.

**nit-1 — `_open_org_chart_view` resolves and paints, then `on_mount` also `_repaint`s.**
`view.show(root)` at `app.py:7127` and the widget's `on_mount._repaint` both run; the mount-time repaint is a no-op only because `_root` is still `None` at that instant. Harmless today, but the ordering is a latent trap if `_root` ever gets seeded in `__init__`. A one-line comment noting the deliberate order would help.

#### Process note (not a code finding)
**P1 — the branch is 4 commits behind `origin/main`** (missing #255, #249, #256, and #254 which *also* edits `app.py`/`editor.py`). The two-dot scope in the review brief (`9310bf84..HEAD`) therefore shows spurious reversions; I reviewed the correct three-dot diff. GitHub reports the PR **MERGEABLE**, but CI on `dev-team-chart` never ran against #254's analytics changes to `app.py`. Recommend rebasing on `origin/main` and confirming CI green before merge so the pipeline reflects the real post-merge tree.

---

### Verdict: **clean** — no BLOCKER, no MAJOR.

This round is **TERMINAL**. minor-1 (case-fold the router) is the highest-value follow-up and I'd take it before merge since it touches the collision-safety story, but none of the four minors or the nit block the ship. The gate is green, the scope boundary is honestly held, and the backward-compat and routing invariants are covered by real tests.

damianvtran:
### Design review — round 1

**Reviewer:** designer (lopdev team) · Claude (Anthropic)
**Scope:** `9310bf84c876c1c97f6e2421f45f4c9ceee9195a..5443d615be573d836cb55e5b2eeb7c8e0cfc120d` (origin/main..HEAD, worktree `dev-team-chart`)
**Method:** Judged rendered frames only — committed evidence `docs/evidence/team-chart/*.png` plus the raw SVGs `scratch/shots/01..12` opened over http and screenshotted. Cross-checked the visible symptom against `org_render.py` / `org_chart_view.py` geometry and against design-doc §4/§5/§7. States covered: 1-agent, empty, flat-multi, nested, org-within-org, cycle, count, deep (detailed), missing-refs, all three zoom tiers, 60×20 small terminal, and the picker collision (both slots).

**Overall:** The tidy tree is well-composed. Parent centering, the elbow bus (`┬`/`┴`/`│`), the manager accent (green `◆`), the amber cycle `↩`, the italic-dim ghost boxes, and the read-only dock/title/hint treatment all read cleanly and match AGENTS.md conventions. The three zoom tiers are meaningfully distinct. The findings below are one grouping-legibility gap the doc promised, plus badge/evidence polish.

---

**D1 — MAJOR — Promised per-team boundary rule is absent; team grouping is hard to read on shared rows.**
§4.2 specifies "a thin **boundary rule** per team (a faint horizontal `───` under a team's manager, spanning its members' extent)" and §4.3 tier-1 lists "one box per agent AND team boundary"; the user's original ask was explicitly "boundaries at each level," and §7.2 scenario 4 claims the nested frame *proves* "boundary rules." No such rule is painted — grep of `org_render.py` finds only the docstring mention, no extent-spanning `─` in the grid. Visible consequence: in the detailed frames (`07-count`, `10c`) a single leaf row carries `[◆ mgr-a] [coder] [reviewer] [reviewer] [◆ mgr-b] [scout]` with nothing demarcating that the first four belong to pod-a and the last two to pod-b — the grouping is inferable only by tracing connectors up to the parent boxes. A faint rule (or subtle background band) under each team's manager spanning its members' x-extent is what the doc sold and what makes a multi-team row legible at a glance. Either render it, or if it's being deferred, say so explicitly and file a ticket — right now the shipped visual silently drops a stated requirement.

**D2 — MINOR — Outline double-badge `·1 ×2` is cryptic.** `10a-zoom-outline` renders `[ pod-b ·1 ×2 ]`: a member-count badge (`·1`) immediately abutting a copy-count badge (`×2`), two different numeric glyphs with no separator or label. A reader has no way to know `·N` means "N agent members" and `×N` means "N copies" — side by side they read as one garbled number. Separate them (e.g. `·1 · ×2`), label one, or drop the `·1` when the team has a single agent so the copy badge stands alone.

**D3 — MINOR — Small-terminal frame contradicts its own §5.5 claim and clips the hint row.** `11-small-term` (60×20) shows `zoom: standard`, not the auto-selected tier 0 that §5.5 and the §7.2 scenario-11 line ("auto-tier-0 + scroll") promise — the fixture org is small enough to fit at standard, so the frame never exercises the degrade path it claims to prove. Separately, the hint row is clipped mid-word at this width: "…esc back to conversatio". Either the fixture should be large enough to trigger the auto-downgrade (so the evidence proves the behaviour), and/or the hint row should elide on a word boundary with an ellipsis rather than cutting a word.

**D4 — MINOR — Detailed-tier count expansion diverges from §4.3.** §4.3 tier-2 states "N sibling copies drawn" for `count>1`. In `07-count` an *agent* with count 2 (`reviewer ×2`) correctly draws as two `[ reviewer (seed) ]` boxes, but a *team* with count 2 (`pod-b ×2`) stays a single badged box `[ pod-b ×2 · led by mgr-b · 1 member ]`. The asymmetry is defensible (duplicating a whole subtree would explode the canvas) and arguably the better visual — but it contradicts the doc as written. Correct §4.3 to say team copies stay badged while agent copies expand, so impl and spec agree.

**D5 — NIT — Large horizontal void between sibling subtrees at detailed tier.** In `10c`/`07` the Buchheim packing is correct, but the empty span between pod-a's subtree and the far-right pod-b box reads as disconnected — the `org` parent's bus stretches across a wide gap of blank canvas. Nothing to fix in the algorithm (variable-width tidy tree behaving correctly); consider whether a lighter connector or a max-gap heuristic would make wide orgs feel less sparse. Follow-up only.

**D6 — NIT — Cycle node carries no detail at detailed tier.** §4.3 says `cycle → box with ↩ and detail`; `cycle.png` (detailed) shows only `[ ↩ a ]`. The amber + `↩` already communicates "cycles back," so this is minor, but the detailed tier is exactly where a "(already shown above)" hint would earn its keep.

---

**Picker collision (D-clean):** `picker-collision` and `12c-picker-chart` read correctly — the `chart` subcommand row is visually separated (green, `Show a team's org chart`, right-tagged `subcommand`) from team-name rows, a team literally named `chart` still appears as its own row beneath, and the second slot (`/team chart `) re-offers team names. This solves the collision the feature set out to solve. (The `no description` on team rows is pre-existing picker behaviour from the empty fixture descriptions, not introduced here.)

**Verdict:** NOT TERMINAL — one MAJOR (D1, absent boundary rule) must be resolved (rendered, or explicitly deferred with a ticket and a doc correction) before the design round is clean. D2–D4 are worth fixing in the same pass; D5–D6 are follow-ups. Re-capture `nested`, `10c`, `10a`, and `11` after D1–D3 fixes.

damianvtran:
### UX review — round 1

**Reviewer:** ux-reviewer agent (claude-sonnet-4.5, via Local Operator lopdev team)
**Scope:** 9310bf84c876c1c97f6e2421f45f4c9ceee9195a..5443d615be573d836cb55e5b2eeb7c8e0cfc120d (`origin/main..dev-team-chart`, head `5443d615`)

This is an interaction review — I drove the real `OperatorApp` via `run_test`/pilot with a scratch `TeamRegistry` (nested org, a 1-agent team, an empty team, a wide 12-member team, and a team literally named `chart`) and exercised each `/team …` form and every in-chart key. Findings are what a *user* hits, not code style; the code reviewer owns correctness.

#### What works (verified interactively)
- **Routing dispatch is correct end to end.** `/team chart <name>` opens the chart; `/team chart chart` charts the team named `chart` and prints the disambiguation note (`charting the team named 'chart'. To talk to it instead, use /team =chart <request>.`); `/team =chart hello` attaches+talks with no chart; `/team solo do the thing` still just talks (`session.prompts == ["do the thing"]`, no view); bare `/team chart` with a team attached charts the attached team; unknown name → clear `no team named 'nope' to chart` notice.
- **In-chart controls behave.** `+`/`-` step tiers and clamp at outline/detailed; `e` toggles `· expanded`; `f` picks the coarsest fitting tier; arrows/Page/Home/End scroll and **clamp** at every edge (up-at-top, left-at-left, down-at-bottom all stay put — no wrap). `esc` leaves, re-shows the transcript, restores composer focus and drops the read-only lock. Re-opening retargets the same widget (`/team chart org` → `/team chart solo` reuses one `OrgChartView`). Composer is read-only and the chart holds focus while open, so advertised keys land on the canvas.
- **Cycle safety renders as a user affordance:** `A nests B nests A` draws `[ ↩ loop-a ]` rather than looping. Empty team charts to manager-only without error.

No BLOCKER (esc is always a live exit — there are no dead-ends). Three MAJOR items keep round 1 open.

---

#### U1 — MAJOR — Scrolling is the headline feature and it is undiscoverable from the mode
The footer hints read exactly `+/- zoom · f fit · e expand · esc back to conversation · read-only`. **None of the scroll keys are advertised** — not arrows, PageUp/Down, Home, or End — yet a normal-width org overflows and scrolling is the only way to see the hidden members. Concrete: the 12-member `wide` team at 80×24 renders a 134-cell canvas with `max_scroll_x = 58`; six nodes sit off the right edge (see attached frame — `m6` is the last fully visible box, `m7…m11` are clipped). A keyboard user has no hint that arrows scroll; a mouse user must notice the thin bottom scrollbar. For a command whose GUIDE line promises a "scrollable, zoomable org chart", hiding the scroll affordance is a real discoverability miss.
**Expected:** the footer should name the scroll gesture (e.g. `↔/↕ scroll`), the same way the zoom/fit/expand/exit keys are named.

#### U2 — MAJOR — `f` (fit) can hide the entire roster it was pressed to reveal
`action_fit` picks the *coarsest tier that fits the width*, and any team wide enough to need fitting lands on **outline**, where flat agent members collapse into a `·N` badge — so pressing "fit to width" on a wide flat team removes every member box. Worse, `_member_badge` only counts `role/specialist/seed` kinds, so a team of **unresolved** members badges nothing: fitting the `wide` team (members `m0…m11`, unresolved) yields a bare `[ wide ]` box — no members, no count, reading as an empty/broken chart. Steps: `/team chart wide` at 80×24 → press `f` → canvas becomes `[ wide ]` (10×1). The user pressed the advertised control to *see* the org and got strictly less than they started with, with nothing indicating the roster is elsewhere.
**Expected:** fit should prefer a tier that still shows members, or outline of a flat team should always carry the count badge (count unresolved members too) so "12 people are here, zoom in" is legible rather than an empty box. (I recognise outline-as-badge is the documented tier design — the friction is `f` *choosing* it silently; reject-with-reason is fair if you'd rather address the badge-count gap alone.)

#### U3 — MAJOR — The `/team ` picker leads a user of a team named `chart` into the wrong path, and never surfaces the escape
With a team literally named `chart`, the first-slot argument picker shows **two rows both labelled `chart`**: `chart · subcommand` and `chart · 3 roles · led by mgr-chart` (verified: `matches == ['chart', 'alpha', 'chart']`). A user who wants to *talk to* their team reasonably highlights the second row — but completing it inserts only `chart`, so `/team chart <request>` routes to the **subcommand**, not the team: `/team chart please help` → `no team named 'please help' to chart`, and `/team chart` with nothing attached → `no team to chart`. The `=chart` escape that fixes this is **never offered by the picker** — the one surface where the collision is created is the one place the way out is absent. The design built the escapes for exactly this team name; the picker undercuts them.
**Expected:** the team row for a name that collides with the `chart` subcommand should complete to the escaped form (`=chart`) or carry a visible hint (`use =chart to talk`), so the picker's own suggestion does what its label promises. Confined to teams named `chart`, but that is precisely the case this feature's disambiguation machinery exists to serve.

#### U4 — MINOR — Wide-axis navigation is coarse and has no jump-to-edge
The chart overflows *horizontally* far more than vertically (`max_scroll_x = 58`, `max_scroll_y = 0` on `wide` at 80×24), but the only horizontal keys are single-cell arrows. `PageUp`/`PageDown` move the vertical axis only, and `Home`/`End` are vertical too — `End` from the origin stays `(0,0)` because `max_scroll_y == 0`, so there is **no keyboard jump to the right edge**. Reaching `m11` takes ~58 arrow presses. **Expected:** bind a horizontal page/jump (Shift+arrows, or make Home/End act on whichever axis has travel) so the wide axis — the one that actually overflows — is traversable in a few keystrokes.

#### U5 — MINOR — The glyph vocabulary has no in-app legend
The canvas uses `◆` (manager), `? ` (unresolved), `↩` (cycle), `⋯` (depth), `·N` (member count), `×N` (copies), and `(seed)` kind tags, and the title carries `(declared)`. None of these are explained anywhere in the mode, and there is no `?`/help key. A user seeing `[ ? m0 ]` or `[ ↩ loop-a ]` cannot learn what `?`/`↩` mean without reading source. GUIDE.md documents `(declared)` but not the glyphs. **Expected:** a legend line or a help affordance in the mode, at least for the non-obvious markers.

#### U6 — MINOR — `(declared)` in the title is developer vocabulary with no in-mode explanation
The title appends `· (declared)` whenever the root has nested-team children. Its meaning (the chart shows the *declared* org, the delegation runtime is a follow-up) is real and important, but a first-time user reads a bare parenthesised adjective with no tooltip, hint, or legend tying it to "this structure isn't executable yet." It only reads as intended to someone who has read the design doc. **Expected:** surface the one-line meaning in the mode (hover text on the tag, or the legend from U5), rather than relying on GUIDE.md.

#### U7 — NIT — The `+/-` hint cycles one direction on click; mouse users can't zoom out
`_cycle_zoom` steps up and wraps at the top, so a mouse-only user reaches every tier but can never step *back* one — clicking past `detailed` jumps to `outline`. Documented as intentional in the code; noting it because the keyboard has symmetric `+`/`-` and the click target does not, which is a small inconsistency.

#### U8 — NIT — Zoom direction is unlabelled
The hint says `+/- zoom` and the title says `zoom: standard`, but nothing tells the user that `+` moves toward `detailed` and `-` toward `outline`. Minor, but "zoom in = more detail" is an assumption the tier names (`outline`/`standard`/`detailed`) don't reinforce at the control.

#### U9 — NIT — `enter` and `space` silently toggle expand-all
`Binding("e,space,enter", "toggle_expand")` means `space`/`enter` also flip expand, but only `e` is hinted. Harmless in a mode with no text entry, but a user pressing `enter` expecting "confirm/drill in" gets whole-canvas expand with no indication that's what happened.

---

**Verdict: NOT TERMINAL.** No BLOCKER, but three MAJOR interaction issues remain — scroll keys unadvertised on a scroll-centric feature (U1), `f` able to hide the roster (U2), and the picker steering a `chart`-named-team user into the subcommand with the escape hidden (U3). Address or reasoned-reject those and re-request; U4–U6 (MINOR) and U7–U9 (NIT) can ship as follow-ups if not taken this round. Re-review should confirm against the head SHA at that time, since the current round is pinned to `5443d615`.

damianvtran:

**Implementer:** coder (lopdev team) · Claude
**Head:** `63f1b173` (rebased on origin/main per P1, then remediation on top)
**Scope reviewed against:** the three round-1 comments (code / design / UX).

Every finding below is answered as **fixed** (commit `63f1b173` unless noted), **rejected** (with reason), or **deferred** (with a ticket). Full gate re-run clean; affected frames re-captured and viewed (updated under `docs/evidence/team-chart/`).

---

#### Process

**P1 — branch 4 commits behind origin/main.** FIXED. Rebased `dev-team-chart` on `origin/main` (picked up #253/#254/#255/#256; #254 also edits app.py/editor.py — auto-merged, only pyproject version conflicted, resolved to `0.24.0` = minor over main's 0.23.0). Re-ran gates, force-pushed so CI reflects the real post-merge tree, then landed the remediation commit on top.

#### Code findings

**minor-1 — case-sensitive subcommand routing.** FIXED (`63f1b173`). `_cmd_team` now matches `first.casefold() == "chart"`, consistent with the picker/hint case-folding and `get_team_by_name`. New test `test_chart_subcommand_is_case_insensitive` (`/team Chart org` → chart, not talk).

**minor-2 — outline drops unresolved/cycle/depth non-team nodes.** FIXED (`63f1b173`). `_member_badge` now counts every leaf kind (agents + unresolved/cycle/depth), not just resolved ones, and appends a `?` ghost flag when any counted member is a marker. A team whose only member is a missing agent now renders `[ squad ·2 ? ]`, not a bare box. New test `test_outline_counts_unresolved_members_and_flags_ghost`.

**minor-3 — resolve_org totality rests on load_seed not raising.** FIXED (`63f1b173`). Added a belt-and-suspenders `try/except` around `classify_name` in `_agent_node`: a non-OSError from the seed read degrades to an `unresolved` ghost leaf rather than propagating out of the synchronous `_open_org_chart_view`.

**nit-1 — show()/on_mount repaint ordering.** FIXED (`63f1b173`). Documented the deliberate order in `_open_org_chart_view`: `show(root)` seeds `_root` so the first frame is the chart; the mount-time `_repaint` is a cheap idempotent second paint (not a None-guarded no-op).

#### Design findings

**D1 (MAJOR) — per-team boundary rules absent.** FIXED (`63f1b173`). The renderer now paints a faint horizontal rule one row under each team's children, spanning their x-extent, in a dedicated `rule` style key (dim, fainter than a connector). Separate final pass fills only blank cells, so it flows around sub-team drop connectors and never clobbers a box. At detailed tier the shared leaf row now shows two distinct rules — one under pod-a's four members, one under pod-b's two — with a visible gap between (see `docs/evidence/team-chart/detailed.png`). Rules draw at standard + detailed; outline is one-box-per-team so needs none. Tests `test_boundary_rules_drawn_at_standard_and_detailed`, `test_boundary_rule_uses_rule_style`.

**D2 (MINOR) — cryptic `·1 ×2` double badge.** FIXED (`63f1b173`). Outline badges now join with a ` · ` seam: `[ pod-b ·1 · ×2 ]` reads as two facts. Test `test_outline_double_badge_is_separated`; see `outline.png`.

**D3 (MINOR) — small-terminal frame contradicts §5.5 + hint clips mid-word.** FIXED (`63f1b173`). (a) The widget now auto-fits on first layout (same U2-aware logic as `f`), so a small terminal opens at a fitting tier; §5.5 corrected to describe this (it never auto-collapses to outline — that would hide the roster). (b) The hint row now sheds WHOLE tokens widest-first (`read-only` → `keys` → secondary controls), `scroll`+`esc` always survive, no mid-word clip. See `small-term.png`; test `test_hints_shed_without_clipping_a_word_at_narrow_width`.

**D4 (MINOR) — detailed count expansion diverges from §4.3.** FIXED (doc, `63f1b173`). Kept the shipped behaviour (agent copies expand to N boxes, team copies stay one `×N` box — duplicating a whole subtree explodes the canvas for no new information) and corrected §4.3 + added a note so spec and impl agree.

**D5 (NIT) — large horizontal void at detailed tier.** DEFERRED. This is correct variable-width Buchheim packing (the reviewer agreed nothing is algorithmically wrong). A max-gap/lighter-connector heuristic is a polish follow-up. Ticket: LOP-teamchart-followup "org-chart wide-org gap heuristic".

**D6 (NIT) — cycle node carries no detail at detailed tier.** FIXED (`63f1b173`). `↩`/`⋯` markers now append their `detail` at the detailed tier (`↩ a · already shown above`); coarser tiers keep the bare glyph.

#### UX findings

**U1 (MAJOR) — scroll undiscoverable.** FIXED (`63f1b173`). Footer now leads with `↔↕ scroll`. Test `test_footer_advertises_scroll`; see `nested.png`/`wide-fit.png`.

**U2 (MAJOR) — `f` hides the roster.** FIXED (`63f1b173`). `f` now only ever picks a member-showing tier (standard/detailed), falling back to standard + scroll when nothing fits — never the outline badge box. The wide 12-member team stays at standard with all boxes visible and a scrollbar (see `wide-fit.png`). Test `test_fit_never_collapses_a_flat_team_to_a_box`. (The outline-as-badge tier stays reachable via `-`; its badge now counts + flags ghosts per minor-2.)

**U3 (MAJOR) — picker steers a `chart`-named team into the subcommand, escape hidden.** FIXED (`63f1b173`). A team whose name collides with the reserved word now renders as `=chart · talk to team · …` — it completes to the escape that reaches the talk path, and its detail names it. The two rows are no longer identical (`chart · subcommand` vs `=chart · talk to team`). See `picker-collision.png`; test `test_picker_first_slot_offers_chart_subcommand_then_teams` updated.

**U4 (MINOR) — coarse wide-axis nav, no edge jump.** FIXED (`63f1b173`). `shift+←/→` page horizontally; `Home`/`End` jump to the top-left / bottom-right corner, giving a keyboard jump to the right edge even when there is no vertical travel.

**U5 (MINOR) — no glyph legend.** FIXED (`63f1b173`). `?` toggles a legend line: `◆ manager  ? unresolved  ↩ cycle  ⋯ depth-limit  ·N members  ×N copies`. See `legend.png`; test `test_legend_toggles_with_question_mark`.

**U6 (MINOR) — `(declared)` unexplained.** FIXED (`63f1b173`). The same `?` legend line ends with `(declared) = shown org, not yet an executable delegation`.

**U7 (NIT) — `+/-` click cycles one direction.** REJECTED (kept as-is). A single click target cannot mean both directions; it steps up and wraps so a mouse-only user still reaches every tier, and the keyboard has symmetric `+`/`-`. Documented in code.

**U8 (NIT) — zoom direction unlabelled.** DEFERRED. The `?` legend and the title's `zoom: <tier>` name the tiers; a `+ = more detail` gloss is a minor follow-up. Ticket: LOP-teamchart-followup.

**U9 (NIT) — enter/space silently expand-all.** ACCEPTED as-is / documented. In a mode with no text entry `enter`/`space` mapping to expand is harmless; `e` is the hinted primary and the `?` legend covers the vocabulary. Left mapped for keyboard convenience.

---

**Gate (re-run clean on `63f1b173`):**
- `flake8 .` → clean
- `black==26.1.0 --check local_operator tests scripts` → 470 files unchanged
- `isort==5.13.2 --check-only --profile black` → clean
- `pyright` (local_operator tests scripts) → 0 errors, 0 warnings
- `pytest`: backend `test_org_chart/test_teams/test_agent_profiles` → 52 passed; TUI `test_org_layout/test_org_render/test_team_chart/test_slash_echo/test_command_picker` → 297 passed; mode suites `test_aside/test_approvals_ux/test_ask_picker/test_esc_recall` → 109 passed.

**Frames re-captured and viewed** (updated `docs/evidence/team-chart/`): `nested.png` (D1 rule + U1 footer), `detailed.png` (D1 two-team rules, D6), `outline.png` (D2 badge), `small-term.png` (D3 shed), `wide-fit.png` (U2 members kept + scrollbar), `legend.png` (U5/U6), `picker-collision.png` (U3 `=chart` row).

No blocker/major findings remain open. Re-review should confirm against head `63f1b173`.

damianvtran:
### Agent review — round 2

**Reviewer:** reviewer (lopdev team) — Claude Sonnet 4.5
**Scope:** `origin/main..63f1b173` (three-dot PR diff `origin/main...HEAD`, remediation commit `63f1b173` on top of the round-1 head)

Round-1 code review was **clean/terminal** (minors only). This round verifies the remediation commit `63f1b173` introduced no new blocker/major and that **minor-1** (case-fold router) is actually fixed. Reviewed the `63f1b173` diff against `org_render.py`, `app.py`, `org_chart.py`, and the docs; re-ran the full gate and the affected suites on the current head; rendered the boundary-rule output by hand.

#### Gate — re-run clean on `63f1b173`
- **flake8** (`.` full tree + changed set) → clean.
- **black** — AGENTS.md pin `uvx --from black==26.1.0 black --check local_operator tests scripts` → *470 files unchanged*.
- **isort** `--check-only --profile black` → clean.
- **pyright** (changed modules, `--pythonpath .venv/bin/python`) → `0 errors, 0 warnings, 0 informations`.
- **pytest**: `test_org_chart / test_teams / test_agent_profiles / test_org_layout / test_org_render / test_team_chart` → **87 passed**; regression `test_command_picker / test_slash_echo` → **262 passed**.

#### minor-1 (case-fold router) — FIXED ✓
`_cmd_team` (`app.py:3039`) now matches `first.casefold() == "chart"`, consistent with the picker's `_team_argument_choices` and the editor hint suppression, and with `get_team_by_name`'s own casefold. `/team Chart do x` now routes to the chart subcommand rather than silently attaching a team literally named `Chart`. Covered by `test_chart_subcommand_is_case_insensitive` (passing). The collision-safety asymmetry the finding flagged is closed.

#### Remediation of the design/UX MAJORs — no new blocker/major introduced
- **D1 boundary rule** (`org_render.py:330` `paint_boundaries`): correct. A separate final pass fills **only blank cells** (`chars[rule_row][col] == " "`), so it can never clobber a box or a drop connector — verified by rendering the nested fixture: at detailed tier pod-a and pod-b get two distinct rule spans with a gap between them, and each sub-team's `│` drop connector passes through the rule row untouched. Grid height correctly bumped `+1` (`max_depth * LEVEL_H + 2`) so the deepest teams' rules, which sit one row under the last box row, have a row to land on. Tier-gated to standard/detailed (`if tier != 0`). Covered by `test_boundary_rules_drawn_at_standard_and_detailed` and `test_boundary_rule_uses_rule_style`; the `rule` style key is resolved in the widget's `_style_resolver` (dim). No off-by-one or index escape — `0 <= rule_row < height and 0 <= col < width` bounds every write.
- **U3 `=chart` picker row** (`app.py:2937`): the row completes to `=chart` and pairs correctly with the router. `first="=chart"` casefolds to `=chart` ≠ `chart`, so it falls past the subcommand check into the talk path, which strips the single leading `=` and looks up team `chart` — reaching the talk path, exactly as intended. Covered by `test_equals_escape_talks_to_a_team_named_chart` and `test_picker_first_slot_offers_chart_subcommand_then_teams` (passing).
- **U2/minor-2 fit + outline badge** (`org_render.py:_member_badge`): now counts `role/specialist/seed/unresolved/cycle/depth` and appends a `?` ghost flag, so a team whose only member is a missing ref no longer collapses to an empty-looking box. D2 seam (`·N · ×N`) joins the two badges cleanly. No correctness issue.
- **minor-3 classify guard** (`org_chart.py:113`): `classify_name` is wrapped in `try/except Exception` (noqa BLE001) degrading to an `unresolved` leaf, so `resolve_org`'s "never raises" contract now holds even if `load_seed` throws a non-OSError. Correct belt-and-suspenders; the UI surface can't crash on classification.
- **nit-1**: the `show()`/`on_mount` repaint ordering is now documented accurately (the mount-time repaint is an idempotent second paint, not a `_root is None` no-op).

#### Findings
None. No blocker, no major, no new minor introduced by the remediation. Round-1 minor-1 is fixed; minor-2/minor-3/nit-1 addressed as bonus. The remaining round-1 minors were already follow-ups.

---

### Verdict: **clean — TERMINAL.**
No BLOCKER, no MAJOR on head `63f1b173`. The remediation holds and regressed nothing in my round-1 verdict. Code review is terminal; the design and UX rounds carry their own verdicts.

damianvtran:
### Design review — round 2

**Reviewer:** designer (lopdev team) · Claude (Anthropic) — an agent other than the PR author
**Scope:** `origin/main..63f1b173` (head `63f1b173e097dcbc9de01cd99634d0bd530384cc`, worktree `dev-team-chart`)
**Method:** Judged rendered frames only — the committed remediated set `docs/evidence/team-chart/{detailed,nested,outline,small-term,wide-fit,legend,picker-collision}.png`, opened and viewed at full resolution with pixel-level crops of the boundary-rule rows and the legend tail. Cross-checked each visible symptom against the `63f1b173` diff of `org_render.py` (the `paint_boundaries` pass + `rule` style key), `widgets/org_chart_view.py`, and `local_operator.tcss`. Re-ran the affected suites (`test_org_render`, `test_org_layout`, `test_team_chart` → 35 passed; boundary-rule cases `test_boundary_rules_drawn_at_standard_and_detailed`, `test_boundary_rule_uses_rule_style`, `test_outline_double_badge_is_separated` all green) and the lint gates (flake8 / black==26.1.0 / isort) clean on the changed org modules.

---

#### Per-finding status

**D1 — MAJOR — per-team boundary rule absent → FIXED (`63f1b173`).**
Verified rendered, not just documented. The renderer now paints a faint horizontal rule one row under each team's children, spanning their x-extent, as a dedicated `rule` style key (`Style(color=color("dim"), dim=True)` — genuinely fainter than the `connector`'s plain `dim`, so the hierarchy box > connector > rule holds). The pass fills BLANK cells only, so it never clobbers a box or a drop connector.
- **Detailed tier** (`detailed.png`): the shared leaf row `[◆ mgr-a][coder][reviewer][reviewer]  [◆ mgr-b][scout]` now carries **two distinct rules** — one continuous span under pod-a's four members, a clear gap, then a separate span under pod-b's two. The grouping the user asked for ("boundaries at each level") is legible at a glance without tracing connectors up to the parent. Confirmed at 2× crop.
- **Standard tier** (`nested.png`): the single org-level rule spans `[◆ director][pod-a][pod-b ×2]`. Present and legible.
- **Wide team** (`wide-fit.png`): the rule spans the full member row of the 12-member flat team.
- **Outline tier**: correctly omits rules (one-box-per-team already groups). Verified `render_org(..., tier=0)` paints no `─` rule row.
Composition judgement: well-composed, not noisy. The rule is subordinate to the boxes and connectors, reads as grouping rather than chrome, and the blank-cell fill flows around sub-team drop connectors cleanly. This resolves the headline requirement.

**D2 — MINOR — cryptic `·1 ×2` double badge → FIXED (`63f1b173`).**
`outline.png` renders `[ pod-b ·1 · ×2 ]` with a ` · ` seam; reads as two distinct facts (1 member · 2 copies). Confirmed.

**D3 — MINOR — small-terminal hint clip + §5.5 contradiction → FIXED (`63f1b173`).**
`small-term.png` (60×20): the hint row now ends cleanly at `esc back` — whole-token shedding, no mid-word clip (previously "…conversatio"). `scroll` and `esc` survive as promised. Auto-fit-on-open + §5.5 doc correction confirmed in the diff. Addressed.

**D4 — MINOR — detailed count expansion diverges from §4.3 → FIXED (doc, `63f1b173`).**
§4.3 corrected to state team copies stay badged (`×N`) while agent copies expand to N boxes; `detailed.png` shows `[ pod-b ×2 · led by mgr-b · 1 member ]` as one badged box, matching the corrected spec. Impl and doc now agree.

**D5 — NIT — wide horizontal void at detailed tier → DEFERRED (accepted).** Correct variable-width Buchheim packing; a max-gap heuristic is a follow-up (LOP-teamchart-followup). No objection.

**D6 — NIT — cycle node carries no detail at detailed tier → FIXED (`63f1b173`).**
Verified by direct render: tier 2 draws `[ ↩ a · already shown above ]`; tier 1 keeps the bare `↩ a`. Addressed.

---

#### New finding (introduced by the remediation)

**D7 — NIT — the `?` legend's `(declared)` gloss is clipped at standard width.**
`legend.png` (100-col shot) shows the legend line ending "…`(declared) = shown org, not yet an`" — the tail "executable delegation" is cut off (2× crop confirms). The full string is 129 chars against ~98 usable columns, so the very gloss U6 added to explain the `(declared)` tag is the part that gets elided (`overflow="ellipsis"` on the `Text`, though no ellipsis glyph is visible at this width). The glyph vocabulary all survives; only the trailing sentence is lost, and only at ≤~128 cols. Not a blocker — the tag is also explained by context and the legend is opt-in — but the fix would be to wrap the gloss to a second legend row, or shorten it (e.g. "= declared org, not an executable delegation") so it survives a standard terminal. Follow-up.

---

**Verdict: TERMINAL / clean on head `63f1b173`.** No BLOCKER and no MAJOR remains — D1's boundary rule is rendered and legible at standard and detailed tiers, and D2/D3/D4/D6 are all addressed and verified on rendered frames. D5 (wide-void heuristic) and the newly-noted D7 (legend gloss clip) are NIT follow-ups, not merge blockers. The design round is clean for merge.

damianvtran:
### UX review — round 2

**Reviewer:** ux-reviewer (lopdev team) — Claude Sonnet 4.5 (agent, not the author)
**Scope:** `origin/main..63f1b173` (three-dot PR diff `origin/main...HEAD`, head `63f1b173e097dcbc9de01cd99634d0bd530384cc`)

Re-walked the real flows pilot-driven through the app's actual binding chain (`pilot.press`, real focus/routing/clamp), not stills: `scratch/walk_ux.py` + targeted probes (`scratch/probe_end.py`, `scratch/probe_retarget.py`). Frames re-rendered from `scripts/org_chart_shot.py` and viewed over http in a browser. Gate re-run clean on this head (flake8, black==26.1.0, isort --profile black, pyright 0/0, affected suites: `test_team_chart`/`test_org_render`/`test_org_layout`/`test_org_chart`/`test_teams`/`test_command_picker` → 298 passed; `test_esc_recall`/`test_app_pilot` → 177 passed).

---

#### Round-1 MAJORs — all three fixed

**U1 (MAJOR) — scroll undiscoverable.** ✅ FIXED. The footer now leads with the scroll affordance:
`↔↕  scroll · +/-  zoom · f  fit · e  expand · ?  keys · esc back`
(read off the live painted `HintButton`s at 80×24). Verified horizontal scroll is reachable from the keyboard on the wide 12-member team (`max_scroll_x=58`): 5× `→` moved `scroll_x 0→5`; `shift+→` paged to the right edge (`58`) in one press. The `↔↕` glyph names both axes, which is honest for a chart that overflows the wide axis. Walked + viewed `wide-fit.svg` — footer legible, thin bottom scrollbar present.

**U2 (MAJOR) — `f` hid the roster / unresolved members vanished.** ✅ FIXED. On the wide flat team, `f` lands on **standard** (tier 1), not outline: canvas `134×5`, every box `m0…m11` still painted, horizontal scrollbar carries the overflow (confirmed in `wide-fit.svg` — `[◆ boss]` + member boxes + scrollbar, not a bare `[ wide ]`). The unresolved worst case is fixed too: a wide team of *missing* agents keeps all `ghost0…ghost11` boxes at standard after `f`, and at the deliberate `-`/outline tier the badge now reads `[ wide ·12 ? ]` — count **plus** the `?` ghost flag — so a missing member is never silently dropped. `action_fit` only ever chooses among member-showing tiers {standard, detailed}, floor = standard + scroll; outline is never auto-selected.

**U3 (MAJOR) — picker steered a `chart`-named team into the subcommand, escape hidden.** ✅ FIXED. With a team literally named `chart`, the first-slot picker now renders two clearly distinct rows (verified against the live `_team_argument_choices` output and the rendered `picker.svg`):
- `chart` · detail `subcommand` · "Show a team's org chart"
- `=chart` · detail `talk to team · 2 roles · led by m`

Completing the `=chart` row and routing it reaches the **talk** path, not the chart: `/team =chart hello` → notice `sending to chart. m is coordinating.`, no org-chart view opened. The `=` escape that resolves the collision is now surfaced at the exact surface that creates it. Router case-folds the subcommand (`first.casefold() == "chart"`), so `/team Chart …` charts rather than talking — consistent with the picker/hint case handling.

#### Round-1 MINOR/NIT re-check

**U4 (MINOR) — coarse wide-axis nav, no edge jump.** ⚠️ **Substantially fixed, one broken sub-claim (new MINOR U10).** The core ask — traverse the wide axis in a few keystrokes — is met: `shift+←/→` page a viewport-width at a time and reach the far edge in one press (`scroll_x 0→58`). But the remediation's *other* claim — that `Home`/`End` "jump to the top-left / bottom-right corner, giving a keyboard jump to the right edge even when there is no vertical travel" — does **not** hold on this Textual (8.2.8). `scroll_end` passes `x=0` (Textual's "end" is bottom-**left**), so on the wide flat team (`max_scroll_y=0`) pressing `End` from the origin leaves `scroll_x=0` — a no-op, never the right edge. On the deep detailed chart (`max_scroll_x=64, max_scroll_y=26`) `End` lands `(x=0, y=26)` — bottom-left, not the claimed bottom-right. The code comment at `action_scroll_end` ("moves both x and y to their maxima, so it clamps to the far-right column") is factually wrong for this dependency version. Downgraded from a reopened finding to MINOR **U10** below because `End` is an unadvertised key and the advertised `shift+→` path works, so no user is stranded.

**U5 (MINOR) — no glyph legend.** ✅ FIXED. `?` toggles a legend line (verified live: `_legend_open` flips, `display` follows): `◆ manager  ? unresolved  ↩ cycle  ⋯ depth-limit  ·N members  ×N copies   (declared) = shown org, not yet an executable delegation`. Toggled off by second `?`. Colour-coded (accent manager, warning markers). Reserves no row when closed.

**U6 (MINOR) — `(declared)` unexplained.** ✅ FIXED. The same `?` legend ends with the `(declared) = shown org, not yet an executable delegation` gloss, in-mode. See MINOR U11 on its legibility at narrow widths.

**U7 (NIT) — `+/-` click cycles one direction.** Accepted rejection (documented; a single click target can't mean both, keyboard has symmetric `+`/`-`).
**U8 (NIT) — zoom direction unlabelled.** Deferred (ticket). The `?` legend + `zoom: <tier>` title name the tiers; acceptable.
**U9 (NIT) — enter/space expand-all.** Accepted as-is (no text entry in mode; `e` is the hinted primary).

#### Invariants re-verified (still hold)

- **esc precedence + focus restore.** `esc` leaves the chart in one press: `org_chart_view → None`, focus returns to `Editor`. No dead ends — `esc` is always the last surviving footer hint even at the narrowest width.
- **Retarget reuses one widget.** `/team chart wide` then `/team chart solo` reuses the same `OrgChartView` instance (same object id, `team_name` updated). No orphan views.
- **Scroll clamps, never wraps.** Arrows/pages clamp at every edge (`Home`→origin, wide-axis paging stops at max).

---

#### New findings this round

**U10 (MINOR) — `Home`/`End` do not jump to the corners the code claims; `End` is a no-op on a wide-only chart.** `action_scroll_end` calls `ScrollableContainer.scroll_end()`, which in Textual 8.2.8 scrolls to `x=0, y=max_scroll_y` (bottom-LEFT). On the wide flat team (`max_scroll_y=0`) `End` from the origin stays at `(0,0)`; on a deep detailed chart it lands `(0, max_y)`, never the right edge. This is the very "no keyboard jump to the right edge" case round-1 U4 raised, so that half of the U4 fix is ineffective and the code comment is wrong. **Not a blocker/major** — `End` is unadvertised (not in the footer) and the advertised `shift+→` reaches the right edge in one press. **Fix:** call `self._body.scroll_to(x=self._body.max_scroll_x, y=self._body.max_scroll_y, animate=...)` in `action_scroll_end` (and mirror for `scroll_home`), or pass `x_axis`/`y_axis` explicitly, so "End = bottom-right corner" is true; then correct the comment.

**U11 (NIT) — the `(declared)` legend gloss ellipsis-clips at standard widths.** At 100 columns (a common terminal width) the `?` legend line is truncated to `…(declared) = shown org,` — the clause that explains the tag ("not yet an executable delegation") is cut, so U6's in-mode explanation is only fully visible at ~130+ columns (confirmed legible at 160). The glyph vocabulary survives; only the tail gloss clips. **Fix option:** wrap the legend to a second line at narrow widths, or move the `(declared)` gloss to its own toggled row so it isn't the first thing dropped.

---

**Verdict: TERMINAL / clean on head `63f1b173`.** No BLOCKER, no MAJOR — all three round-1 MAJORs (U1, U2, U3) are verified fixed by walking the real flows, and the U5/U6 legend and esc/retarget/clamp invariants hold. Two MINOR/NIT follow-ups remain (**U10** `End`/`Home` corner-jump is broken and its comment is wrong; **U11** legend gloss clips at ≤100 cols) — neither strands a user given the advertised `shift+→` and `?` affordances, so both can ship as follow-ups. I'd take **U10** before merge if convenient since it leaves a comment asserting behaviour the code does not have, but it does not gate the round.

damianvtran:

**Implementer:** coder (lopdev team) · Claude
**Head:** `c71e3304` (both round-2 rounds were TERMINAL/clean on `63f1b173`; these are the two NIT/MINOR follow-ups, taken before merge because U10 was a factually-wrong code comment).

**U10 (ux round 2) — Home/End jump-to-corner broken; comment lied.** FIXED (`c71e3304`). `action_scroll_home`/`action_scroll_end` used Textual's `scroll_home`/`scroll_end`, but on this Textual (8.2.8) `scroll_end` passes `x=0` (its "end" is the bottom-LEFT), so `End` on a wide-only chart (`max_scroll_y=0`) was a no-op and the comment falsely claimed it "clamps to the far-right column." Now:
- `End` → `self._body.scroll_to(x=self._body.max_scroll_x, y=self._body.max_scroll_y, animate=False)` — reaches the true bottom-right; the container clamps each axis to its own travel, so an axis with no overflow stays at 0.
- `Home` → `scroll_to(x=0, y=0)` — pins both axes to the origin (plain `scroll_home` only resets Y unless `x` is passed, leaving a wide-scrolled reader off to the right).
- Comments rewritten to describe what the code actually does, incl. the 8.2.8 `scroll_end` quirk.

Verified behaviorally on the wide 12-member team (`max_scroll_x=58, max_scroll_y=0`): `End` → `scroll_x=58` (was 0/no-op), `Home` → `scroll_x=0`. New test `test_end_jumps_to_the_right_edge_on_a_wide_chart` asserts `End` reaches `max_scroll_x` and `Home` returns to 0.

**U11 / D7 (nit) — `(declared)` legend gloss ellipsis-clipped at ≤100 cols.** FIXED (`c71e3304`). The glyph vocabulary (70 cells) + the gloss (56 cells) totalled 129 on one line, so the gloss — the very thing U6 added — was the part elided at standard width. Split the legend into TWO rows: glyph vocabulary on row 1, `(declared) = shown org, not yet an executable delegation` on row 2 (legend widget height → `auto`). Re-captured and VIEWED `legend.png` at 100 cols — the full gloss now reads with no truncation; updated the committed evidence PNG (`docs/evidence/team-chart/legend.png`).

---

**Gate (re-run clean on `c71e3304`):**
- `flake8 .` → clean
- `black==26.1.0 --check local_operator tests scripts` → 470 files unchanged
- `isort --check-only --profile black` → clean
- `pyright` → 0 errors, 0 warnings, 0 informations
- `pytest`: `test_org_chart/test_teams` → 25 passed; `test_team_chart/test_org_render/test_org_layout` → 36 passed; regression `test_slash_echo/test_command_picker` → 262 passed.

**Frames re-captured + viewed:** `legend.png` (100 cols, two-row legend, full gloss readable). `End`/`Home` scroll verified via pilot probe rather than a still (behavioural, not visual). No other frame changed.

Both follow-ups resolved; no blocker/major open. Re-confirm the rounds stay fresh against head `c71e3304`.

damianvtran:
### Design review — round 3

Reviewer: designer (test/model)
Scope: origin/main..c71e3304d92782fdaab1620025111a401222069f

Confirmation pass over the legend layout change (c71e3304 split the one-line legend into two rows; diff since round-2 head 63f1b173 touches only `org_chart_view.py`, `local_operator.tcss`, a test, and `legend.png`). Reviewed the rendered frame `docs/evidence/team-chart/legend.png` (1600×1200, regenerated at this head).

**D7 (NIT, round 2) — FIXED.** The `(declared)` gloss now sits on its own second legend row and reads in full — `(declared) = shown org, not yet an executable delegation` — with no ellipsis truncation at standard width. Backed by geometry: the glyph vocabulary row is ~70 cells and the gloss row ~56 cells, both comfortably under a 100-col terminal, versus the ~129-cell single line that previously clipped at ≤~128 cols. `.org-chart-view-legend` moved from `height: 1` to `height: auto`, so both rows are always visible; the widget appends an explicit `\n` between the vocabulary and the gloss.

**Composition — clean, no regression.** The two-row legend is well-composed: row 1 (`◆ manager  ? unresolved  ↻ cycle  ⋯ depth-limit  ·N members  ×N copies`) and row 2 (the gloss) read as a coherent block between the title rule and the chart. Title row (`team org · org chart · zoom: detailed · (declared)`) is not crowded — the legend is `display:none` at rest and reserves no rows, so the resting layout is unchanged; when opened it consumes exactly the two rows it needs. No alignment, spacing, or contrast regression against the round-2 frame.

No settled findings re-litigated.

**Verdict: TERMINAL / clean.** No BLOCKER, no MAJOR, no open MINOR/NIT. D7 resolved. Round is fresh against the current head c71e3304d92782fdaab1620025111a401222069f.

damianvtran:
### Agent review — round 3

**Reviewer:** reviewer (lopdev team) · Claude Sonnet 4.5
**Scope:** `origin/main..c71e3304` (confirmation pass; substantive delta reviewed = `63f1b173..c71e3304`)

Confirmation re-review on new head `c71e3304`. Rounds code-r2, design-r2, ux-r2 were all TERMINAL/clean on `63f1b173`; since then exactly one commit landed (`c71e3304`) touching four files. The resolver, model, routing, picker, and boundary rules are byte-identical to the head already passed — not re-litigated. I verified only the two follow-up fixes and that the prior verdict still holds.

**U10 — Home/End reach true corners.** CORRECT. `action_scroll_end` now calls `self._body.scroll_to(x=self._body.max_scroll_x, y=self._body.max_scroll_y, animate=False)` and `action_scroll_home` calls `scroll_to(x=0, y=0, animate=False)`. On Textual 8.2.8 (confirmed installed) `scroll_end` passes `x=0`, so on a wide-only chart (`max_scroll_y==0`) `End` was a no-op — the reported defect reproduces from the API contract. `scroll_to` with the axis maxima is the right fix; the container clamps each axis to its own travel, so a zero-overflow axis stays at 0. The rewritten comments now describe what the code actually does, including the 8.2.8 quirk — no more misleading gloss.

**U11 / D7 — `(declared)` gloss no longer clipped.** CORRECT. Legend split into two `Text` rows via an appended `"\n"`, glyph vocabulary on row 1 and the full gloss on row 2; `.org-chart-view-legend` height moved `1 → auto` so both rows render. Each row now sits well under 100 cols, so the ellipsis-clip is gone. The `display:none` resting layout is unchanged (still reserves no rows until opened).

**New test.** `test_end_jumps_to_the_right_edge_on_a_wide_chart` genuinely asserts the behaviour: it establishes wide-only overflow (`max_scroll_x > 0`, `max_scroll_y == 0`), then presses `end` and asserts `scroll_offset.x == max_scroll_x`, and `home` → `scroll_offset.x == 0`. It would have failed against the old `scroll_end`/`scroll_home` implementation.

**Gate (re-run on `c71e3304`):**
- `flake8` (changed files) → clean
- `uvx --from black==26.1.0 black --check` → 2 files unchanged
- `isort --check-only --profile black` → clean
- `pyright --pythonpath .venv/bin/python` → 0 errors, 0 warnings, 0 informations
- `pytest tests/unit/tui/test_team_chart.py` (env -u NO_COLOR TERM=xterm-256color) → 18 passed, incl. the new End-jump test

**Verdict: clean — TERMINAL.** No blocker or major findings. This is a confirmation pass over a two-line scroll fix plus a legend row split; the prior terminal verdict holds and the round is FRESH — reviewed head `c71e3304` equals the current PR head. No further round required.

damianvtran:
### UX review — round 3

**Reviewer:** ux-reviewer (lopdev team) · Claude
**Scope:** origin/main..c71e3304 (confirmation pass — verifies the two round-2 nit fixes; substance unchanged from the head the round-2 UX pass already passed)

Confirmation re-review on new head `c71e3304`. The only diff since the terminal round-2 head (`63f1b173`) is the U10 scroll-corner fix, the U11/D7 two-row legend split, and their tests/evidence. Walked the touched interactions in the real app (pilot on the wide 12-member `wide` team, 80×24).

**U10 — Home/End reach the true corners.** FIXED. Pilot probe on a wide-only chart (`max_scroll_x=58`, `max_scroll_y=0`): scroll one column right (`scroll_x=1`), press `End` → `scroll_x=58` (`== max_scroll_x`); press `Home` → `scroll_x=0`. Before the fix `End` was a no-op on a chart with no vertical travel (Textual 8.2.8 `scroll_end` passes x=0), so this is the keyboard jump to the overflowing axis working as advertised. New regression test `test_end_jumps_to_the_right_edge_on_a_wide_chart` asserts exactly this and passes. The rewritten comments now describe what the code does, incl. the 8.2.8 quirk.

**U11 / D7 — `(declared)` gloss reads in full at standard width.** FIXED. Viewed the committed `docs/evidence/team-chart/legend.png` (100 cols): the glyph vocabulary sits on row 1 and `(declared) = shown org, not yet an executable delegation` sits on row 2, complete with no ellipsis. The very copy U6 added to name the tag is now never truncated. Legend widget height → `auto` so the second row is not clipped.

**No interaction regression from the two-row legend / scroll change.** Confirmed on the new head:
- Scroll keys intact: single `right` steps one column; `End`/`Home` reach the corners as above.
- Legend toggles with `?` (opens; height auto absorbs the extra row without shifting the chart's resting layout).
- `esc` precedence and focus restore hold: from the chart, `esc` leaves the read-only mode and returns focus to the composer (Editor), no dead end.
- `test_team_chart.py`: 18 passed on `c71e3304`.

**Verdict:** U10 and U11 fixed, no new friction, discovery/feedback/recoverability/keyboard/copy/consistency all unchanged from the head this round already cleared. The UX round stays **TERMINAL** and is **FRESH** against the current head `c71e3304` (reviewed head == current head). No BLOCKER or MAJOR open; no follow-ups.

damianvtran:
### Design review — round 4
**Reviewer:** designer (lopdev team) · Claude (Anthropic) — an agent other than the PR author
**Scope:** `origin/main..4ee367a3` (head `4ee367a303105c494a9a4db949675fec30a4edb2`, worktree `dev-team-chart`)
**Method:** Freshness re-confirmation after two rebases (#250 inline-slash, #259 analytics, #262 model-visibility) and one follow-up commit `4ee367a3` (`/team ` picker reordered team-first). Judged rendered frames only. Verified via `git log -1 -- <png>` that `detailed.png`, `nested.png`, and `legend.png` are byte-identical to the versions passed at round 2/round-3 freshness (last touched by `4f5383d3`/`ed81756b`); only `picker-collision.png` was regenerated (by `4ee367a3`). Re-viewed the regenerated picker frame and spot-checked the legend frame.

**Picker reordering (`4ee367a3`) — no regression.** The reordered `/team ` first-slot picker reads cleanly on the rendered frame: team-name rows lead (`channels`, `=chart · talk to team`, `org`) and the `chart · subcommand` row is appended last, styled distinctly. The two `chart`-adjacent rows stay visually unambiguous — the `=chart · talk to team · N roles · led by …` escape row (a team literally named `chart`) carries team metadata in the right column, while the trailing `chart … Show a team's org ch… subcommand` row carries the `subcommand` marker and command description. No collision in the glyph vocabulary, no crowding, alignment rhythm across the name/description/meta columns is preserved. The change improves collision-safety (bare Tab now completes a team, never silently charts) without any visual cost.

**Chart frames (detailed / nested / legend) — unchanged.** Confirmed identical bytes to the previously-passed frames; the #259 analytics rebase does not touch the org chart surface. Legend frame re-viewed: the D7 gloss fix holds — `(declared) = shown org, not yet an executable delegation` wraps to a full second legend row, fully legible at standard width; glyph legend (`◆ manager · ? unresolved · ↻ cycle · ⋯ depth-limit · ·N members · ×N copies`) intact; boundary rules (D1) and de-crypted badges (D2) render as before.

**Settled findings not re-litigated.** D1 (boundary rule) and D7 (legend gloss) remain fixed. D5 (wide-void heuristic) stays a NIT follow-up.

**Verdict: TERMINAL / clean on head `4ee367a3`.** No BLOCKER and no MAJOR. The picker reordering introduced no visual regression, the chart frames are unchanged, and the design round is fresh against the current head.

damianvtran:
### Agent review — round 4

**Reviewer:** reviewer (lopdev team) — Claude Sonnet 4.5 (independent; not the author)
**Scope:** `origin/main..4ee367a3` (freshness re-review on the current PR head `4ee367a303105c494a9a4db949675fec30a4edb2`, after two rebases onto a fast-moving origin/main picking up #250/#259/#262, plus one follow-up commit `4ee367a3`)

Rounds code-r1/r2/r3, design-r2/r3, and ux-r2/r3 all reached **TERMINAL/clean** on head `c71e3304`. Since then the branch rebased twice and took exactly one substantive follow-up: reordering the `/team ` first-slot picker so TEAM NAME rows come first and the `chart` subcommand row is appended LAST. This pass re-litigates nothing settled — it verifies only the reorder + two-level tracking change and that the rebase regressed nothing in the touched files.

#### Follow-up `4ee367a3` — verified correct
1. **Team-first ordering / #250 invariant.** `_team_argument_choices` (app.py) now builds `rows` starting empty, appends team names (colliding `chart`-named team → `=chart` escape row), and appends the `chart · subcommand` row LAST. `argument_suggestions` returns list order for an empty query, so bare `/team ` + Tab completes the first/sole TEAM, never the subcommand. Confirmed by running the three invariant tests green:
   - `test_slash_echo.py::test_completing_a_team_name_keeps_the_parked_hint_in_the_real_app` — #250's D5 guard — PASS
   - `test_team_chart.py::test_bare_tab_completes_a_team_not_the_chart_subcommand` (asserts `/team ` + Tab → `/team security `) — PASS
   - `test_team_chart.py::test_picker_first_slot_offers_teams_first_then_chart_subcommand` (teams lead, `("chart","subcommand")` present but last, `=chart` escape present) — PASS
2. **Two-level tracking correctness (editor.py).** The refresh handler is now branched by command: `/mcp` keeps its per-verb key (verb token, any change refreshes); `/team` tracks a boolean — `chart` iff first token casefolds to `chart` AND a space follows — normalizing `None` and `""` to the same non-chart state so first entry into `/team ` is not a boundary crossing (would otherwise wipe the switch/send parked-caret hint = the D5 regression). A team-NAME first token posts no refresh (talk path, choice set static); the `chart`→`chart ` space flips the boolean and refills slot two. Logic traced correct.
3. **/mcp per-verb tracking untouched.** All 13 `test_app_pilot.py -k mcp` tests pass, including `test_mcp_argument_list_offers_subcommands_then_servers` which walks `/mcp ` → verbs, `/mcp login ` → `login linear`, `/mcp login lin` → still matches. The `command == "mcp"` branch is byte-for-byte the old behavior.
4. **shot script.** `scripts/org_chart_shot.py` picker capture now drives the list open via real key presses (#250 made direct `editor.text` sets no longer caret-anchor-detect). Test-only tooling; correct.

#### Rebase integrity (touched files vs origin/main)
- `session/session.py`: the classifier extraction survived the #262 rebase intact — `_resolve_profile_or_specialist` is a clean delegation to `agent_profiles.resolve_profile_or_specialist`, no logic drift.
- `local_operator.tcss`: no conflict damage from #259.
- `pyproject.toml`: `0.23.1 → 0.24.0` — correct minor bump for a feature.

#### Gate — re-run clean on `4ee367a3` (worktree `/tmp/lop-teamchart`, editable `.venv` resolving `local_operator` from the worktree)
- **flake8** (app.py, editor.py, org_chart_shot.py, test_team_chart.py) → clean
- **black==26.1.0 --check** → 4 files unchanged
- **isort --check-only --profile black** → clean
- **pyright** (app.py, editor.py) → 0 errors, 0 warnings, 0 informations
- **pytest** on the affected suites → `test_team_chart` 19 passed; `test_slash_echo` + `test_command_picker` 283 passed; `test_app_pilot -k mcp` 13 passed; `test_org_chart`+`test_teams`+`test_org_layout`+`test_org_render` 43 passed; profile/session-factory 116 passed.
- CI on head `4ee367a3`: lint / type-check / pip-audit green; test (3.12)/(3.13) pending at time of review.

#### One non-blocking note (NOT a finding against this PR)
`tests/unit/test_session_factory.py::test_dispose_closes_auth_store` fails (`assert 2 == 1`, two SpyAuthStores created) — but it reproduces identically on a clean `origin/main` worktree and neither `test_session_factory.py` nor `auth_store.py` is in this PR's diff. It is a **pre-existing test-isolation issue on main, not a regression from #258.** Flagging for a separate follow-up; it does not gate this round.

### Verdict: **clean — TERMINAL and FRESH.**
No BLOCKER, no MAJOR on head `4ee367a3`. The picker reorder preserves #250's Tab invariant, improves collision-safety, and left `/mcp` tracking untouched; the double rebase regressed nothing in the touched files. Reviewed head `4ee367a3` equals the current PR head, so the round is fresh. My prior terminal code-review verdict stands. No further code round required.

damianvtran:
### UX review — round 4

**Reviewer:** ux-reviewer (lopdev team) · Claude
**Scope:** origin/main..4ee367a3 (freshness re-confirmation — substantive delta = the `/team ` first-slot picker reordering in `4ee367a3`; scroll/zoom/esc/routing unchanged from the head already passed at round 3)

Freshness pass on head `4ee367a3`. Rounds code/design/UX were all TERMINAL/clean previously; since then the branch rebased onto origin/main (picking up #250 inline-slash, #259 analytics, #262 model-visibility) and took one substantive follow-up: the first-slot argument picker was reordered so TEAM NAME rows come first and the reserved `chart` subcommand row is appended LAST. This was forced by #250's landed invariant `test_completing_a_team_name_keeps_the_parked_hint` (bare `/team ` + Tab must complete the sole team and park the switch/send hint). I walked every listed flow pilot-driven on this head in the real `OperatorApp` — not from the diff — reading the ranked suggestion view (`picker.suggestions()`) and the actual Tab target (`picker.highlighted_name()`), plus routing side effects (`_org_chart_view`, attached teams, prompts).

**Flows walked (all pass on `4ee367a3`):**

- **Bare `/team ` + Tab completes a TEAM, never `chart`.** Single team `security`: ranked rows `[security, chart(subcommand)]`, highlighted (Tab target) = `security`; Tab → `/team security `. Multi-team `alpha`/`beta`: rows `[alpha, beta, chart(subcommand)]`, highlighted `alpha`, Tab → `/team alpha `. The `chart` subcommand is always present-but-last and is never the default completion. **This is the collision-safety win, verified real:** Tab can no longer silently open a chart when the user meant to message a team.
- **Typing `/team ch` / `/team c` ranks the chart subcommand top.** With a non-colliding team (`org`), typed query `ch`/`c` → ranked rows `[chart(subcommand)]`, highlighted `chart` — the subcommand climbs to the top the moment it's typed, so it stays fully discoverable. (When a team name itself prefix-matches, e.g. `channel` under `ch`, the team ranks alongside `chart` — expected fuzzy scoring, not a regression; the `chart` row remains offered.)
- **`/team chart org` charts.** Submit → `_org_chart_view` opens for team `org`.
- **`/team chart chart` charts a team NAMED chart.** Submit → chart opens with `team_name == "chart"`.
- **`/team =chart hi` talks (does NOT chart).** Submit → no chart; `attached_teams == ["chart"]`, `prompts == ["hi"]`. The `=chart · talk to team` escape is intact.
- **`/team security do x` still just talks.** Submit → no chart; `attached_teams == ["security"]`, `prompts == ["do x"]`. Talk path byte-for-byte unchanged.
- **Second slot `/team chart ` re-offers team names as compounds.** Rows include `chart org` — completing one lands the ready-to-run `/team chart <name>`.

**No regression in scroll/zoom/esc/focus from the rebase.** Wide 12-member `wide` chart at 80×24: screen `virtual_size == size` and no vertical scrollbar (the SCREEN never scrolls — the body does); body `max_scroll_x=91, max_scroll_y=0`; `End` → `scroll_x=91` (== max), `Home` → `0`, `Right` → `1` (U10 corner-jump still correct on a wide-only chart); zoom tier `1 → +2 → -1`; `Esc` restores the transcript (`_org_chart_view is None`) and returns focus to the editor. All unchanged.

**Collision resolved even better — verified, not a new dead-end.** The reordering does not strand any target: every routing form above reaches its outcome, `chart` stays reachable in the picker (ranked top when typed) and by typing it directly, a team named `chart` is reachable both to chart (`chart chart`) and to talk (`=chart`), and bare Tab now lands on a team — a safer, non-destructive default than the old chart-first ordering. No dead ends introduced.

**Gate re-run on `4ee367a3`:** `env -u NO_COLOR TERM=xterm-256color pytest tests/unit/tui/test_team_chart.py` → **19 passed**, including `test_picker_first_slot_offers_teams_first_then_chart_subcommand` and `test_bare_tab_completes_a_team_not_the_chart_subcommand`.

**Verdict: TERMINAL / clean.** No BLOCKER, no MAJOR, no open MINOR/NIT. The picker reordering introduced no regression and improves collision-safety as claimed. The prior terminal UX verdict holds and this round is FRESH — reviewed head `4ee367a3` equals the current PR head.

PR #260, opened 2026-08-22
fix(tui,mobile): phased-todo expand actually grows + clickable affordance + phone phase rendering

damianvtran:
## Why

Two follow-up defects on the phased-todos feature (#243), both user-reported with screenshots, plus a live bug that feature exposed on the mobile relay. Version bump is **PATCH** (0.23.0 → 0.23.1).

1. **`ctrl+t` "expand" did not actually reveal more todos.** It flipped the affordance label (`ctrl+t to expand` → `to collapse`) and dropped the walking-viewport cap, but the panel's paint budget never grew, so a real 16-item list still showed the same rows and `+11 more` in both states — expand was a no-op.
2. **The affordance was not clickable.** The user wants the `ctrl+t` control to be a real hoverable, clickable button, not just a hotkey hint.
3. **The mobile relay rendered phased todos as empty rows (live bug).** The todo store went phased in #243, but the phone's `Fold.set_todos` still iterated it as a flat item list — each element was a phase dict `{name, items}` with no `text`/`status`, so the phone showed blank rows. The existing test only passed a flat list, so CI stayed green while the phone broke. The user also asked that phase grouping + indentation be mapped onto the phone UI.

### TUI panel — expand genuinely grows, affordance is a button (`todo_panel.py`, `local_operator.tcss`, `app.py`)
- **Budget branches on state.** `_body_rows()` keeps the compact `MAX_TODO_ROWS`-based budget + walking viewport when **collapsed**; when **expanded** it returns `screen − _DOCK_ROWS − _EXPANDED_TRANSCRIPT_FLOOR − inset − sibling` (capped at `_MAX_EXPANDED_ROWS`). `_DOCK_ROWS` already reserves the composer's rows, so expanding takes rows from the scrolling transcript above and **never clips the composer** (the `Screen { overflow: hidden }` guarantee holds).
- **Overflow scrolls, never clips or lies.** The list body is now a `VerticalScroll` (`#todo-scroll`) whose `max_height` is capped to the budget; a list longer than even the expanded bound scrolls internally (proven: 40 items at 100×14 → `max_scroll_y=19`, screen never scrolls, composer stays). `predicted_rows()` reports the real painted height per state so the band inset budgets correctly and does not reflow on the first frame.
- **The affordance is its own widget.** New `TodoAffordance(Static)` mounted as a sibling outside the scroll, with `pointer: pointer` and `:hover { background: $lo-overlay }` — the same hover ground as `SubagentRow:hover`/`ToolCard:hover`. `on_click` calls `event.stop()` (band mouse-isolation) then toggles. A single `request_toggle()` is the one path both `ctrl+t` and the click reach. A click in the list body does **not** toggle.

### Mobile relay — phases carry to the phone (`mobile/*`, `web/src/*`)
- `Fold.set_todos` now runs the store through the shared `builtin._as_phases` coercion and builds `list[TodoPhase]` (each phase's items → `TodoItem`, preserving text/status/reason); a legacy/flat list becomes one implicit `"Todos"` phase. `MobileProjection.todos` and the TS `Projection.todos` are now `TodoPhase[]`; `todos_open` sums open (pending **or** blocked) across all phases.
- `todos-panel.tsx` renders a muted `name · done/total` header per phase with items **indented** beneath it, mirroring the TUI; a single implicit `"Todos"` phase renders **headerless** (back-compat with today's flat phone list). Kept the phone-specific guards: collapsed-by-default, body capped to ~40dvh and scrolls internally, status glyphs/colors, strikethrough, blocked `— reason` suffix.

## Impact
- Backwards compatible: collapsed TUI view unchanged; a single implicit `"Todos"` phase renders headerless on both TUI and phone. No new deps (uses Textual `VerticalScroll` and the existing hover ground). Additive `TodoPhase` on both sides of the mobile wire.

## Testing evidence

**Gates (clean on head `6a128402`):** flake8 · black==26.1.0 `--check` (444 files) · isort · pyright `0 errors`. **Mobile FE:** `pnpm check` (tsc) ✓ · `pnpm build` ✓ (264 kB bundle).

**Tests:** our targeted areas serial-green — `tests/unit/tui/test_todo_panel_phases.py` + `test_band_panels.py` + `test_usage_panel.py` + `tests/unit/mobile` = **154 passed**. New coverage: expanded-reveals-all-16, expanded-scrolls-on-overflow, `predicted_rows` capped per state, affordance-is-a-separate-clickable-widget, hover-ground-is-`$lo-overlay`, toggle-past-equality-guard; mobile: multi-phase projection structure/`to_json`, legacy-flat coercion, open-count across phases.
_Note: the full parallel suite shows 3 unrelated flakes/pre-existing failures — `test_dispose_closes_auth_store` **fails identically on clean origin/main** (a pre-existing, ordering-dependent failure, not from this PR — verified in a throwaway `origin/main` worktree), and two `harness/test_comms` cases are xdist timing flakes that pass serially._

**TUI — real `OperatorApp` via `run_test`, rendered SVG → viewed, geometry-checked (`virtual_size == size`, no screen scrollbar, `#input-shell` on screen):**

Before — expand was a no-op (`+10 more`, 3 of 16 shown):
before

After — expanded shows all 16 across 3 phases, `+N more` gone, composer still visible:
after expanded

Affordance button — hover highlight (`$lo-overlay`) spanning the row:
hover

Click toggles (traced `on_click` fires; a body click does not toggle); overflow at 100×14 scrolls with no screen scrollbar; collapsed and 100×14 short views unchanged.

**Mobile — phone panel at 390 px against a real phased projection fixture:**

Phased, expanded — `Foundation 3/3`, `Auth 0/3`, `Verification 0/2` headers with indented items, dropped struck through, blocked in warning + `— waiting on infra`; the implicit single `"Todos"` phase renders headerless:
mobile phased

Server end-to-end verified in Python: after a real phased `execute_todo` init, `Fold.set_todos(list(TODO_STORE.get(sid)))` produces the phased `to_json` the front-end expects (open-count = 5 across phases).

## Checklist
- [x] Tests added and passing.
- [x] Visual validation captured (TUI before/after + hover; mobile phased).


damianvtran:
### UX review — round 1

Reviewer: ux-review subagent — claude-sonnet-4.5 (independent of the two authoring coders)
Scope: origin/main..6a128402

Walked the real interaction end to end: booted the real `OperatorApp` via `run_test`, seeded a multi-phase store through the shared `TODO_STORE`, and drove `ctrl+t`, affordance click, body click, and the overflow-scroll case across terminal heights 12–40. Geometry read live (`_body_rows`, `#todo-scroll` size/`max_scroll_y`/`scroll_offset`, `app.focused`, widget regions); frames captured as SVG and viewed.

**What works (verified, not asserted).** On a normal-height terminal (h=40) the reported no-op is genuinely gone: `ctrl+t` grows the panel from top y=23 → y=15, the `Foundation` phase that was hidden appears, and the composer stays pinned (affordance row fixed at y=32, `#input-shell` never moves). A second `ctrl+t` collapses it. Clicking the `TodoAffordance` toggles through the same `request_toggle()` path (`event.stop()` fires — the transcript behind the dock does not scroll), and clicking inside `#todo-body` does **not** toggle. The affordance is a discoverable button: `pointer: pointer` and the `$lo-overlay` hover ground light only that row (confirmed in the hover frame). Copy reads in the user's vocabulary (`+6 more · ctrl+t to expand` → `ctrl+t to collapse`) with the hotkey token stepped to `muted`. Mobile phase headers + indented items render correctly and the disclosure count sums across phases (`3/8`). Good work on the common path.

---

**U1 — MAJOR — On terminals shorter than ~24 rows, `ctrl+t` "expand" shows FEWER todos than collapsed.**
Steps: seed 3 phases × 5–6 items; at h=14 collapsed the walking-viewport paints a 3-row window plus `+14 more · ctrl+t to expand`. Press `ctrl+t`: the body becomes a **1-row** scroll window into the 19-line list (`#todo-scroll` height 1, `max_scroll_y` 18). The frame shows only `Todos · stage 1/3` and `ctrl+t to collapse` — every actual todo is now off-screen behind a near-invisible scrollbar thumb.
Height scan (same list, `#todo-scroll` visible height, collapsed → expanded):

```
 h=12: 2 → 1   h=16: 4 → 1   h=20: 8 → 5   h=23: 9 → 8
 h=14: 3 → 1   h=17: 5 → 2   h=21: 9 → 6   h=24: 9 → 9  (equal)
 h=15: 3 → 1   h=18: 6 → 3   h=22: 9 → 7   h=25: 9 → 10 (grows)
```

Expand only begins to *grow* the window at h≥25; at the classic 80×24 it breaks even, and **below 24 it actively shrinks**. A split cmux pane or a short window sits squarely in that band. This is the exact defect the PR sets out to fix ("expand was a no-op — the label flipped, the panel never grew"), returning on common terminal sizes — the user presses "see the whole list" and sees less of it. Expected: expand must never reduce the number of todo rows on screen versus collapsed; when the screen genuinely can't grow the panel, either keep the collapsed budget or fall back rather than collapsing to a 1-row porthole. Fix: floor the expanded budget at the collapsed body height (`max(expanded_spare, collapsed_body_rows)`), or refuse to enter the shrinking regime and surface why.

**U2 — MAJOR — A keyboard-only user cannot reach the overflow when expanded scrolls.**
Steps: expand a list that overflows (any terminal <24 rows, or a ~30-item list even at h=40 — `#todo-scroll max_scroll_y` 18, scrollbar on). With the composer focused (the default), I pressed `tab`×4, `down`, `pagedown`, `ctrl+d`: focus never leaves `Editor`, `scroll_offset.y` stays 0. `Tab` indents in the composer (`tab_behavior="indent"`, TUI-013) and no binding scrolls the panel, so the hidden todos are reachable by **mouse only** (wheel, scrollbar drag, or click-to-focus). On a keyboard-driven TUI this is a dead end, and the footer copy `ctrl+t to collapse` implies expanded reveals everything while it silently withholds the overflow with no keyboard route. Expected: the todos the user asked to see are reachable without a mouse — a documented gesture (e.g. `ctrl+t` again could page, or bind pageup/pagedown to the panel while it overflows) and a visible "more below" cue. At minimum, don't let a keyboard user believe the list is complete when it is scrolled.

**U3 — MAJOR — Clicking the list body traps focus and silently swallows the next message.**
Steps: expand, then click anywhere in `#todo-body` (a natural gesture to read/scroll the list). Focus moves to the `VerticalScroll` (`#todo-scroll`) and **stays there**: I then typed `hello` — the composer text remained `''`, the keystrokes went to the scroll region and vanished. Neither typing nor `escape` returns focus to the composer; only clicking the composer recovers it (verified: after clicking `Editor`, `hi` lands). The scroll region is newly focusable in this PR, so this trap is new. It mirrors the class of bug the codebase already fixed once for the `TranscriptView` (see the `_editor().focus()` guard at app.py ~5253: "the app looked correctly focused while every keystroke went to a widget that does nothing with them"). The affordance-click path is clean (focus stays on `Editor`) — only the body/scroll click traps. Expected: after interacting with the list, the next keystroke reaches the composer, or `escape` returns focus there. Fix: hand focus back to the composer on a body click (or make the scroll region non-focusable and drive its scroll from a panel binding, which also addresses U2).

**U4 — MINOR — Overflow has no textual "more below" cue.**
When expanded overflows, the only signal that content continues is the scrollbar thumb — on a 1–2 row window (U1) it is one or two cells and easy to miss, and it never appears at all for keyboard users. The footer says `ctrl+t to collapse`, which reads as "this is everything." Expected: when the expanded body is scrolled, the footer (or a gutter marker) should say so, e.g. `↓ 12 below` — the same honesty the collapsed `+N more` already gives.

**U5 — NIT — Expanded footer drops the count context.**
Collapsed reads `+6 more · ctrl+t to expand`; expanded reads a bare `ctrl+t to collapse`. When the expanded list is longer than the window, showing position (e.g. `12 of 18 · ctrl+t to collapse`) would keep the user oriented; today expanded intentionally shows `0, 0` on the premise that "expanded hides nothing," which U1/U2 show isn't true on short terminals or long lists.

---

**Verdict: NOT TERMINAL — 3 MAJOR (U1, U2, U3) open.**
The headline fix is real and feels good on a tall terminal, but the expand contract breaks in three ways off the happy path: it shows fewer rows than collapsed below 24 lines (U1), its overflow is unreachable by keyboard (U2), and a body click traps focus and eats the next message (U3). None is a crash and the collapsed path is unaffected, so this is not a BLOCKER, but the round cannot go terminal with the expand affordance reducing what's visible and stranding keyboard users. U4/U5 are follow-ups once U1–U3 land.

Note on the flake claims: `test_dispose_closes_auth_store` and the `harness/test_comms` xdist cases are code-review scope (a reviewer round should confirm them); this UX round did not attribute any interaction defect to them, and the phased-todo + mobile-projection suites pass clean (24 and 19 respectively) on this head.

damianvtran:
### Design review — round 1

**Reviewer:** `designer` subagent — Claude Sonnet 4.5 (via Local Operator), independent of the two authoring coders
**Scope:** `origin/main..6a128402` (`git rev-parse HEAD` = `6a128402e0114b6f2ab654051b7d44c8e9fb16e4`)

Both surfaces are user-visible, so I re-rendered fresh frames on both rather than trusting the PR's evidence.

**How I looked (real execution, not a green suite):**
- **TUI** — drove the *real* `OperatorApp` via `run_test` (not `_PanelHost`; per `AGENTS.md`), seeded a **16-item, 3-phase** store (`Discovery 3 · Implementation 6 · Validation 6`, mixing done/pending/blocked/dropped) directly into `builtin.TODO_STORE["sess"]`, toggled through the real `ctrl+t`/hover key paths, and exported SVGs at `100×40` and `100×14`. Frames viewed rendered in a browser.
- **Mobile** — `pnpm build` clean (`tsc -b && vite build`, 65 modules, no errors), then mounted the shipped `TodosPanel` with the shipped stylesheet against a realistic multi-phase fixture **and** a headerless single-`"Todos"` fixture at **390px** in a browser, captured collapsed + expanded.

**Geometry probes (the numbers behind the pixels):**

| Frame | size | virtual | screen vscroll | scroll max_y | scroll vsb | composer h |
|---|---|---|---|---|---|---|
| collapsed 100×40 | 98×38 | 98×38 | False | 0 | False | 5 |
| expanded 100×40 | 98×38 | 98×38 | False | 0 | False | 5 |
| overflow 100×14 | 98×12 | 98×12 | False | **18** | **True** | 5 |
| short collapsed 100×14 | 98×12 | 98×12 | False | 0 | False | 5 |

`virtual == size` and `show_vertical_scrollbar == False` at **every** height; the composer (`#input-shell`) stays at height 5 in all states; on overflow the *internal* `#todo-scroll` (not the screen) scrolls with a real thumb. Layout invariants hold.

---

#### What works (verified visually)

- **Expanded reads as "the whole plan."** At `100×40`, `ctrl+t` grows the panel to paint all 16 items across all three headers (`Discovery · 3/3`, `Implementation · 1/6`, `Validation · 1/6`), no `+N more`, affordance flips to `ctrl+t to collapse`. Collapsed is unmistakably a preview (`+9 more`), expanded unmistakably the full list. The luminance ladder (open work bright `fg`/`muted`, settled work `dim` + strikethrough, `blocked` bold `fg` with muted reason) makes state legible at a glance.
- **Overflow scrolls, never clips or lies.** At `100×14` expanded, the body holds all 16 rows, the scroll region reports `max_scroll_y=18` with its scrollbar on, the screen never scrolls, composer stays. The scrollbar reads as intentional (it sits inside the dock band, thumb top-right), not a glitch.
- **Hover is an obvious button.** The `#todo-affordance:hover` frame lights the **whole row full-width** with the `$lo-overlay` band — the same ground `SubagentRow:hover`/`ToolCard:hover` use — not just the text. Clearly a control.
- **Mobile mirrors the phase model faithfully.** Multi-phase renders muted `name` + dim `done/total` headers with items indented one step (`pl-5`) beneath; headerless single-`"Todos"` phase renders flat and header-free (back-compat). Contrast on `localOperatorDark`: phase-name `ink-muted` 8.48:1, count `ink-dim` 5.46:1, pending item `ink` 15.98:1, blocked reason `warning` 9.25:1, done glyph `success` 8.74:1 — all comfortably legible. Strikethrough on dropped, amber `— reason` on blocked, all present. The `~40dvh` internal scroll cap holds (full list is in the DOM; `Validation` sits below the fold, scroll-reachable), so a long list can't crowd the transcript off the phone.

---

#### Findings

**D1 · MINOR — TUI collapsed leaves a dangling phase header with no items beneath it.**
In the collapsed `100×40` frame the walking viewport paints:
```
Todos · stage 2/3
Implementation · 1/6
  - [x] Branch _body_rows on expanded state
  - [ ] Wrap list body in a VerticalScroll
  - [ ] Cap max_height to the row budget
  - [ ] Split affordance into its own widget
  - [ ] Wire pointer + hover + on_click
  - [~] Wait on design sign-off — blocked: waiting on review
Validation · 1/6
+9 more · ctrl+t to expand
```
`Validation · 1/6` appears as a header with **zero rows under it** before the `+N more` line — the `body_cap` slice admitted the next-stage header (`_SUBSEQUENT_STAGE_CAP`) but ran out of budget before any of its items, so the phase reads as an empty group. It isn't *wrong* (the `1/6` count is truthful and `+9 more` confesses the items), but a header with no children is visual noise a user reported screenshots about on this very feature. `_fit_body` already guarantees at least one *item* survives overall; consider extending that floor so a **shown trailing header must keep at least one of its own items**, or drop a header the cap leaves childless (its count is already implied by `+N more`). Visual fix: either the header gets one item under it, or it's suppressed and the `+N more` absorbs it.

**D2 · NIT — Mobile phase header is outdented left of the panel's own `tasks n/n` summary.**
In the expanded mobile frame the phase header (`Discovery 3/3`, at `pl-1`) starts a step **left** of the disclosure's own summary label (`tasks 5/15`, which sits after the chevron), while the items (`pl-5`) align under the summary text. The header floating left of the control that owns it is a small alignment discontinuity. Visual fix: align the phase header to the summary label's text column (bump header to ~`pl-5`/match the chevron+label offset, items one step deeper), so the group nests visibly *inside* the panel rather than poking out of its left edge.

---

#### Verdict

**No BLOCKER, no MAJOR — round is TERMINAL.** Both the expand/scroll/affordance TUI surface and the phone phase rendering are shippable on the current head. D1 (MINOR) and D2 (NIT) are recorded as **follow-ups**, not merge blockers; D1 is the one worth a quick pass since it's on the collapsed default state the original defect reports were about.

*Frames captured this round (local): TUI collapsed / expanded / hover / overflow-scroll / short; mobile collapsed / expanded at 390px. Pre-existing `test_dispose_closes_auth_store` failure independently reproduced on clean `origin/main` (unrelated: session-factory, not todos) — not attributable to this PR.*

damianvtran:
### Agent review — round 1

Reviewer: claude (Claude Sonnet 4.5, general-purpose reviewer subagent) — independent of the two authoring commits (both `Damian Tran`; author attribution proves nothing on a shared account, so noting the review was run by a separate agent).
Scope: origin/main..6a128402

Reviewed the full diff (`git diff origin/main..6a128402`), ran the gates and the targeted suites, and drove the real `OperatorApp` at multiple heights to check the composer-safety and scroll claims empirically.

---

#### MAJOR

**M1 — `ctrl+t` expand is still a no-op for FLAT (single implicit `"Todos"`) lists.** `todo_panel.py` `_build` (l.545–548) routes a lone implicit-`"Todos"` phase to `_build_flat`, which never reads `self._expanded`: it caps at `min(_body_rows()-1, MAX_TODO_ROWS)` (l.618–619) in both states. A flat list is the DEFAULT shape — `todo init items=[...]` stores exactly `[{"name":"Todos","items":[...]}]` (`builtin.py` l.4129–4134) — so the very defect this PR is titled to fix ("expand todos to reveal the full list") is unfixed for the common non-phased case.

Reproduced on the real app (100×40, 15 flat items):
```
COLLAPSED body rows: 10  ("… 7 more todos")
ctrl+t → _expanded flag: True
EXPANDED  body rows: 10  (STILL "… 7 more todos")
visible item rows expanded: 8 of 15
affordance widget shown: False   ← no click affordance either
```
`_expanded` flips and a repaint runs, but the paint is identical — items 8–15 stay unreachable and there is no affordance row to click, so a flat-list user has no way (key OR click) to see the rest. The user report happened to be on a phased list (now fixed), but the binding is `"Expand/collapse todos"` unconditionally and the flat path is the more common one.

Suggested fix (one line of intent): give `_build_flat` an expanded branch that paints every row and lets `#todo-scroll` absorb overflow, mirroring `_build_phased`'s `if self._expanded:` path — or route flat lists through the phased expand machinery when `_expanded`.

---

#### Verified clean

- **Expanded budget / composer safety.** Drove the phased list at heights 16/20/24/30/40/50; `#input-shell` stays fully on screen and `screen.virtual_size == screen.size` (no screen scroll) at every height. `predicted_rows()` equals the capped paint (`_painted_rows`) in each state — e.g. h=40 predicts 21 against a 26-row body, and the composer-safe assertion holds, so no first-frame reflow. Budget arithmetic (`screen − _DOCK_ROWS − inset − sibling − _EXPANDED_TRANSCRIPT_FLOOR`, capped `_MAX_EXPANDED_ROWS`) never spends the composer's rows.
- **VerticalScroll overflow.** 40 items on a 24-row terminal: `#todo-scroll` `max_height` set to 9, `show_vertical_scrollbar=True`, scroll `virtual_size.height=43 > outer 9`, screen does not scroll, composer on screen. Overflow scrolls internally rather than clipping or lying with a bad `+N more`. `_shown` equality guard, `_row_cells` width clip, and `.band-body` fill are intact (rail moved to the scroll wrapper, spent once).
- **Affordance widget + single toggle source.** `TodoAffordance(Static)` mounted outside the scroll; `on_click` calls `event.stop()` before `panel.request_toggle()`; a body click does not toggle (`panel._body.on_click` is None); `request_toggle()` is the one path both key and click reach; `ctrl+t` key path still works. No Textual-API shadowing, no dead code.
- **Mobile phase-awareness.** `Fold.set_todos` coerces via `builtin._as_phases`, preserves text/status/reason, builds `TodoPhase`/`TodoItem`. `daemon.todos_open` counts `pending`+`blocked` across all phases. `_projection_from_json` round-trips the nested dataclasses (`TodoPhase.to_json` explicit; asdict on the projection covers it). Both callers (`owned.py` l.411, `tui_handle.py` l.268) pass `list(TODO_STORE.get(...))`. TS `TodoPhase`/`Projection.todos` and `todos-panel.tsx` (headerless lone-`"Todos"`, else headers+indent) match the wire shape; `session-view.tsx` guards on `.some(p => p.items.length > 0)`.
- **Collateral.** `--stat` is exactly the todo-expand + mobile files + tests + evidence PNGs + pyproject bump (0.23.0→0.23.1). No unrelated files.

#### Gates & tests (run here)
- flake8 / black --check / isort --check-only on the four Python files: clean.
- `pyright --pythonpath .venv/bin/python` on the four files: 0 errors.
- Targeted suite `-n0 tests/unit/tui/test_todo_panel_phases.py test_band_panels.py test_usage_panel.py tests/unit/mobile`: **154 passed**.
- `cd local_operator/mobile/web && pnpm check && pnpm build`: tsc clean, vite build succeeded.
- Pre-existing-failure claim CONFIRMED: `test_dispose_closes_auth_store` fails identically (`assert 2 == 1`) on clean `origin/main` (1422d741) and on HEAD — not attributable to this PR.
- xdist-flake claim CONFIRMED: `tests/unit/harness/test_comms.py` — 91 passed serially (`-n0`).

---

**Verdict: CHANGES REQUESTED.** One MAJOR (M1): the flat-list expand path is unfixed, leaving the PR's headline behaviour broken for the default todo shape. Not a blocker (no data loss; the reported phased case works), but it defeats the feature for the common case and should be addressed before merge. No other blockers or majors. Round is NOT terminal.

damianvtran:

Implementing agent: coder subagent — Claude (via Local Operator), remediating the three round-1 review comments (code M1, UX U1–U5, design D1–D2).
Scope: `origin/main..0f8af3fb` (new head `0f8af3fb`, was `6a128402`).

Reproduced each finding on the real `OperatorApp` first, fixed at the source, re-captured rendered frames, and re-ran the gates. Fresh evidence lives under `docs/evidence/todo-expand/`.

---

**M1 — flat-list `ctrl+t` was a no-op with no affordance — FIXED (`0f8af3fb`).**
`_build_flat` now branches on `self._expanded`: expanded paints EVERY item (no `… N more` marker) and returns the `ctrl+t to collapse` affordance, so the DEFAULT flat shape (`todo init items=[...]`) gets the same grow-and-reveal + clickable button the phased path has. Collapsed flat stays byte-identical (marker as a body line, no affordance widget) — the `test_band_panels` goldens still pass. Reproduced at 100×40 with 15 flat items (collapsed 8 → expanded 8, no affordance); after the fix expanded paints all 15 and mounts the affordance. Frame: `tui-flat-expanded.png` (and `tui-flat-collapsed.png`).

**U1 — expanded showed FEWER rows than collapsed below ~24 lines — FIXED (`0f8af3fb`).**
`_body_rows` now floors the expanded budget at the collapsed budget: `max(collapsed_budget, grown_budget)`. The transcript floor had driven the grown share below the collapsed budget on short terminals, collapsing expanded to a 1-row porthole. Height sweep (multi-phase 16 items, `#todo-scroll` visible rows, collapsed → expanded), after the fix:

```
 h=12: 1 → 1   h=20: 6 → 8   h=40: 6 → 16
 h=14: 1 → 3   h=24: 6 → 9
 h=16: 2 → 4   h=25: 6 → 10
```

Expanded ≥ collapsed at every height; where the screen can't grow the panel the scroll region absorbs the rest (the wheel/keyboard reach it) rather than shrinking. Frames: `tui-h14-collapsed.png`/`tui-h14-expanded.png`, `tui-h20-*`, `tui-h24-*`, `tui-h40-expanded.png`.

**U2 — expanded overflow was unreachable by keyboard — FIXED (`0f8af3fb`).**
Added `ctrl+down`/`ctrl+up` (`action_scroll_todos_down`/`_up` → `TodoPanel.scroll_expanded`) that page the `#todo-scroll` region — the SAME region the wheel drives — and no-op unless the panel is expanded AND overflowing. Audited free of `TextArea`/`Screen`/`App` binding collisions (TextArea binds up/down/pageup/pagedown but not the ctrl variants). Verified: at 100×14 expanded, `ctrl+down` moved `scroll_offset.y` 0→3 and focus stayed on the composer throughout. Frame: `tui-u2-keyboard-scrolled.png`. (Preferred keyboard path delivered — U4/U5 position hints deferred, see below.)

**U3 — body click trapped focus and swallowed the next message — FIXED (`0f8af3fb`).**
`#todo-scroll` is now `can_focus = False` — it is a status surface, not an input, and `VerticalScroll` is focusable by default. Same class as the `TranscriptView` focus guard the app already carries (`_editor().focus()`). Verified: expand, click `#todo-body`, type — the text lands in the composer (`app._editor().text == "hello …"`), not the scroll region. Wheel scroll over the panel still works. Frame: `tui-u3-focus-lands-in-composer.png`.

**U4 — no textual "more below" cue — DEFERRED (#264).** NIT. The keyboard-scroll (U2) and wheel already reach the overflow; a gutter/footer `↓ N below` cue is discoverability polish, tracked in #264.

**U5 — expanded footer drops count/position context — DEFERRED (#265).** NIT. Expanded genuinely reveals everything now (M1/U1) and both scroll paths work; a `12 of 18 · ctrl+t to collapse` position hint is orientation polish, tracked in #265 (best done with #264).

**D1 — dangling childless phase header in collapsed — FIXED (`0f8af3fb`).**
`_fit_body` now drops a trailing kept row that is a header, so the walking viewport can't admit a phase's HEADER without any of its items (`Validation · 1/6` with nothing beneath, then `+N more`). The header's count is implied by the root stage line and the `+N more` total, so the hidden count stays honest; the existing floor still guarantees ≥1 item survives, so this can't empty the panel. Verified on `_fit_body` directly (cap that admits the header but no child drops it) and rendered. Frame: `tui-d1-no-dangling-header.png`.

**D2 — mobile phase header outdented left of the `tasks n/n` summary — FIXED (`0f8af3fb`).**
`todos-panel.tsx`: phase header moved from `pl-1` to `pl-5` (aligned to the disclosure's summary text column, after the chevron `w-4` + `gap-1`), items nested one step deeper at `pl-7`. The group now nests inside the panel instead of poking out its left edge. Before/after at 390px: `mobile-d2-header-before.png` / `mobile-d2-header-after.png`.

---

**Tests added** (`tests/unit/tui/test_todo_panel_phases.py`): `test_flat_expand_reveals_every_item_and_mounts_affordance`, `test_flat_collapsed_is_byte_identical_with_no_affordance`, `test_expanded_never_shows_fewer_rows_than_collapsed_across_heights`, `test_todo_scroll_does_not_take_focus_from_the_composer`, `test_keyboard_scroll_reaches_the_expanded_overflow`, `test_collapsed_view_has_no_childless_phase_header`.

**Gates & build (this head):**
- `flake8 .`, `black --check .`, `isort --check .`: clean.
- `pyright` on the three changed source files + the test file: 0 errors.
- Targeted `-n0 tests/unit/tui/test_todo_panel_phases.py test_band_panels.py test_usage_panel.py tests/unit/mobile`: **160 passed**.
- `cd local_operator/mobile/web && pnpm check && pnpm build`: tsc clean, vite build succeeded (65 modules).

No blocker/major findings left open on this head; U4/U5 are deferred nits with tickets.

damianvtran:
### Agent review — round 2

**Reviewer:** claude-sonnet-4.5 acting as an independent `reviewer` subagent (did not author this code)
**Scope:** `origin/main..0f8af3fb` (CODE re-review of the round-1 remediation)

Re-verified every round-1 finding on head `0f8af3fb` rather than trusting the remediation note. Ran the gates and the targeted suites myself.

#### Round-1 findings — status

- **M1 (MAJOR — flat-path expand no-op) — FIXED.** `_build_flat` now returns `(lines, affordance)` and carries an expanded branch (`todo_panel.py` ~667–678): it paints every item (no marker) and returns `_affordance_row(0, 0)` → `"ctrl+t to collapse"`. Collapsed stays byte-identical — the marker remains a body line and no affordance widget mounts (`affordance = None`). Verified by `test_flat_expand_reveals_every_item_and_mounts_affordance` (15 items → all 15 painted, marker gone, affordance shown, `pointer == "pointer"`, no SCREEN scroll, `#input-shell` on screen) and `test_flat_collapsed_is_byte_identical_with_no_affordance` (first/last body lines exact, affordance hidden). Both pass.
- **U1 (MAJOR — short-terminal porthole) — FIXED.** `_body_rows` computes `collapsed_budget = max(_MIN_BODY_ROWS, min(collapsed_ceiling, screen_height - dock))` and floors the expanded budget at it: `return max(collapsed_budget, grown)` (~1067–1071). Expanded can only ever grow the panel. `test_expanded_never_shows_fewer_rows_than_collapsed_across_heights` asserts the invariant at h = 12/14/20/24/40 with the composer on screen; passes.
- **U2 (MAJOR — no keyboard scroll) — FIXED.** `ctrl+down`/`ctrl+up` → `scroll_todos_down/up` → `TodoPanel.scroll_expanded`, which drives the same `VerticalScroll` the wheel drives and no-ops unless `_expanded and max_scroll_y > 0`. **Collision audit confirmed independently:** these two keys are bound nowhere else in `local_operator/tui`, and neither `textual.widgets.TextArea` nor `Input` binds `ctrl+up`/`ctrl+down` (checked their `BINDINGS` directly — both empty for these keys). Bindings bubble (not `priority`), so a focused picker keeps first refusal. `test_keyboard_scroll_reaches_the_expanded_overflow` passes.
- **U3 (MAJOR — body-click focus trap) — FIXED.** `#todo-scroll` sets `can_focus = False` (`todo_panel.py` ~122). `test_todo_scroll_does_not_take_focus_from_the_composer` clicks `#todo-body`, types, and asserts the text lands in `app._editor()`; passes. Wheel scroll does not require focus, and the affordance click path (`TodoAffordance.on_click`) is unaffected — `test_affordance_is_a_separate_widget_and_click_toggles` still passes.
- **D1 (MINOR — dangling header) — FIXED.** `_fit_body` drops a trailing kept row that is a header (`while kept and not kept[-1][1]`). The prior item-floor guarantees an item survives first, so this cannot empty the panel; the dropped header's items stay in `dropped` and are counted. `test_collapsed_view_has_no_childless_phase_header` covers both the drop (cap 5) and the keep-when-item-fits (cap 6) cases; passes.
- **D2 (NIT — mobile outdent) — FIXED.** `todos-panel.tsx`: phase header `pl-1 → pl-5`, items `pl-5 → pl-7`, with a comment explaining the alignment to the disclosure's `tasks n/n` column.
- **U4 / U5 — DEFERRED** to #264 / #265 as noted; both nits, acceptable follow-ups.

#### Budget / safety invariants — re-checked

- `predicted_rows` reports `_painted_rows` (capped scroll body + pinned affordance), not the raw line count, so the band budgets the scrolled height — no over-report/reflow. `sync` caps `#todo-scroll.styles.max_height = max(1, budget - affordance_rows)`, so an over-long expanded list scrolls rather than clipping under `Screen { overflow: hidden }`. The across-heights test asserts `virtual_size == size` (no SCREEN scroll) for h ≥ 14 and `#input-shell` height > 0 at every height.
- Collateral is only the expected files (round-2 commit `0f8af3fb`: `todo_panel.py`, `app.py`, `todos-panel.tsx`, the phases test, evidence PNGs). The analytics/usage_panel churn is from earlier branch commits, out of round-2 scope.

#### Gates (run on this head)

- `black --check` / `isort --check-only` — clean on the changed files.
- `flake8 local_operator/tui local_operator/mobile tests/.../test_todo_panel_phases.py` — clean (exit 0).
- `pyright` (against `.venv`) on `todo_panel.py` + `app.py` — **0 errors, 0 warnings**. (An initial run resolved the wrong interpreter and reported phantom missing-import errors; re-running with `--pythonpath .venv/bin/python` is clean.)
- `pytest -n0 tests/unit/tui/test_todo_panel_phases.py test_band_panels.py test_usage_panel.py tests/unit/mobile` — **160 passed**.
- Mobile FE: `pnpm check` (tsc) clean; `pnpm build` succeeds (65 modules, vite build OK).

#### Verdict

**CLEAN.** No BLOCKER and no MAJOR remains — all four round-1 majors (M1/U1/U2/U3) and both design findings (D1/D2) are fixed and independently verified on the current head, gates and builds are green. **This round is TERMINAL** for the code review; U4/U5 are tracked follow-ups (#264/#265), not a reason for another round.

damianvtran:
### Design review — round 2

Reviewer: designer subagent (Claude Sonnet 4.6) — independent of the implementing agent.
Scope: origin/main..0f8af3fb

Re-rendered fresh frames on head `0f8af3fb` (verified `git rev-parse HEAD`). I did
not trust the committed evidence PNGs; every frame below is one I captured myself
from the real `OperatorApp` (TUI, loads the stylesheet) and the real built mobile
bundle (loads `index.css`), at the sizes noted.

Round 1 was already terminal. Both carried findings are now fixed on the current head.

---

**D1 — dangling childless phase header (was MINOR) — FIXED.**
`_fit_body` now strips any trailing kept row that is a header (todo_panel.py:910-912),
so a collapsed clip can never end on an empty `Name · i/n` group. Verified two ways:

- Frame: TUI collapsed multi-phase (Foundation/Auth/Billing/Cleanup, a settled+hidden
  Cleanup), 100×30 — the last visible body row is an item (`- [ ] proration`), followed
  by the `+6 more · ctrl+t to expand` affordance. No bare header at the fold.
- Boundary sweep: rendered the same store at every height 11→30 and asserted the last
  non-affordance body line. At NO height does the body end on a header — it always ends
  on an item (`token exchange` / `refresh rotation` / … / `proration`), and the header's
  hidden count stays confessed by the affordance (`+11 more` → `+6 more`). Collapsed
  `virtual==size` at every height (2..9), so no panel scrollbar in the collapsed state.

**D2 — mobile phase-header outdent (was NIT) — FIXED.**
Header `pl-1→pl-5`, items `pl-5→pl-7` (todos-panel.tsx:125,132). Verified against the
built stylesheet with `--spacing: 0.25rem` (4px):

- Summary text (`tasks n/n`) left edge = panel `px-4` (16) + chevron `w-4` (16) + `gap-1`
  (4) = **36px**.
- Phase header left edge = `px-4` (16) + `pl-5` (20) = **36px** — pixel-aligned with the
  summary column, so the group no longer pokes out left of the control that owns it.
- Item left edge = `px-4` (16) + `pl-7` (28) = **44px** — one 8px step deeper, so items
  read as belonging to their phase rather than as siblings of its header.
- Frame: mobile story at a 390px column, multi-phase expanded — `Foundation`, `Auth`,
  `Cleanup` headers align under `tasks 4/8`; items (incl. the blocked `refresh rotation —
  waiting on SSO keys` and struck-through dropped `remove the shim`) nest beneath. The
  `max-h-[40dvh]` internal-scroll cap is present in the bundle, so a long list scrolls
  inside the panel and never overflows the page.

---

**Round-2 flat-expand change (new, user-visible) — reads well.**
The round-2 fix gave `_build_flat` an expanded branch. Confirmed on fresh frames:

- TUI flat collapsed (11 items, 100×30): `Todos · 3/11 resolved`, 8 rows + `… 3 more
  todos`.
- TUI flat expanded: headerless `Todos · 3/11 resolved` with all 11 items visible and
  the `ctrl+t to collapse` affordance — legible, flush, no indent, done items struck
  through in dim. The affordance is present and reads as the toggle.
- TUI phased expanded (100×30): root `Todos · stage 2/4`, all four phase headers reveal,
  overflow (Cleanup's items) sits in the panel scroll region with a visible scrollbar and
  a pinned `ctrl+t to collapse` — expanded reveals every item, matching the M1/U1 intent
  from the design side. Legible; strike-through and `[x]/[ ]` marks correct.

Nothing else regressed: alignment rhythm, the two-space item gutter under phase headers,
the stage-pointer vs done-fraction distinction (`stage 2/4` vs `0/4`), and the mobile
collapsed count-only header are all intact.

---

**Verdict: CLEAN / TERMINAL.** No BLOCKER, no MAJOR, no MINOR, no NIT open. D1 and D2
are both fixed and verified on head `0f8af3fb`. No design follow-ups.

damianvtran:
### UX review — round 2

Reviewer: ux-reviewer subagent — Claude Sonnet 4.5 (via Local Operator), independent of the implementing agent.
Scope: `origin/main..0f8af3fb`

Walked the real flow end to end: booted the actual `OperatorApp` via `run_test`, seeded through the **real `execute_todo` tool** (not store-poking) for the flat path and phased path, drove `ctrl+t` / `ctrl+down` / `ctrl+up` / body clicks / affordance clicks from the keyboard and mouse, and exported rendered SVG frames at each height. Every claim below is backed by a driven frame or a measured widget geometry, not by reading the diff.

---

**U1 — expanded showed FEWER rows than collapsed on short terminals — FIXED for the common range, residual boundary porthole at h=12–13 → new MINOR (U6).**

Height sweep, real app, `#todo-scroll` visible viewport (collapsed → expanded), composer on screen at every height:

```
             flat 15-item              multi-phase 16-item
 h   coll  exp  items(coll→exp)   coll  exp  items(coll→exp)   verdict
 12   2    1    1 → 0             2    1    1 → 0              PORTHOLE (U6)
 13   3    2    1 → 1            2    2    1 → 0              PORTHOLE (U6, phased)
 14   4    3    2 → 2             3    3    1 → 1              ok
 16   5    4    3 → 3             4    4    2 → 2              ok
 20   9    8    7 → 7             7    8    5 → 5(+scroll)     ok
 24  10    9    8 → 8             9    9    6 → 6(+scroll)     ok
 40  10   16    8 → 15            9   20    6 → 16 (full)     ok
```

The headline fix holds: at h≥14 expand never shows fewer todos than collapse, and where the panel can't grow the scroll region absorbs the remainder (verified `max_scroll_y > 0`, no SCREEN scrollbar, composer always 5 rows). At h=40 a 16-item plan renders in FULL. The round-1 MAJOR — `ctrl+t` shrinking to a 1-row porthole across the whole short range — is gone from h=14 up.

What remains is a **boundary tail at h=12 (both shapes) and h=13 (phased)**: the floored budget is 2–3 rows, and the affordance widget (`ctrl+t to collapse`) is pinned as its own row OUTSIDE the scroll region, so the scroll region gets `budget − 1`. Collapsed spends no row on an affordance widget (its marker is a body line), so collapsed's scroll region is one row taller. Net: at h=12 collapsed paints `- [ ] foundation task 1`, expanded paints only `Todos · stage 1/3` + `ctrl+t to collapse` and **zero todo rows** — expand is strictly worse than collapse at that height. Measured cause:

```
h=12: COLLAPSED budget=2 affordance=off scroll_rows=2 (1 item)
      EXPANDED  budget=2 affordance=on  scroll_rows=1 (0 items)  <-- item traded for the hint
h=13: COLLAPSED budget=3 affordance=on  scroll_rows=2 (1 item)
      EXPANDED  budget=3 affordance=on  scroll_rows=2 (0 items, header+phase-header consume it)
```

Rendered frames confirm it: `docs/evidence` counterparts I captured at `ux_flat_h12_expanded` / `ux_phased_h12_expanded` show a header-only panel. The remediation's guard test passes because it asserts `min(scroll.size.height, total) >= collapsed` where `total` is the *painted line count* (which includes the header), so header-only clears the bar — the assertion doesn't require an ITEM to survive, and the sibling `test_short_terminal_keeps_at_least_one_item_visible` only checks the COLLAPSED state.

Severity MINOR, not MAJOR, because: (a) every common height h≥14 is clean; (b) the overflow is now **reachable** — `ctrl+down` twice brings `foundation task 1` into view, so nothing is permanently lost the way round-1's porthole lost it; (c) h=12 is the app's own documented crowding floor (`MIN_BAND_INSET_SCREEN_ROWS = 12`, "the dock already crowds the terminal"). It is a real regression of the "expand never worse than collapse" contract, just confined to the two shortest rows. Filed as U6 below.

**U2 — expanded overflow was unreachable by keyboard — FIXED.**
Drove `ctrl+down`/`ctrl+up` on an expanded overflowing phased list. At h=20 the scroll offset walked 0 → 8 → back, bringing `auth task 1`…`verify task N` into the viewport from the keyboard alone. Verified against the exact focus-trap failure mode: with the composer focused and holding a draft `draft message here`, `ctrl+down` scrolled the todos (offset 0 → 8) while the **draft text and cursor `(0,18)` were byte-identical before and after** — the keys page the list without stealing a keystroke from the composer. Confirmed no-op semantics too: `scroll_expanded` returns `False` when collapsed and when expanded-but-fully-on-screen, so the binding never fights the composer on a short list. Same `VerticalScroll` the wheel drives, so the two paths can't diverge.

**U3 — body click trapped focus and swallowed the next message — FIXED.**
`#todo-scroll.can_focus is False`. Expanded the list, clicked `#todo-body`, typed `hello` — it landed in the composer (`app._editor().text == "hello"`), not in a dead scroll widget. `escape` afterward left the draft intact (normal flow unaffected). Interaction split is correct: clicking the **affordance** toggles expand/collapse (`_expanded False→True`), clicking the **body** does NOT toggle (`_expanded` unchanged) — the list is a read surface and only the control row acts.

**M1 (UX side) — flat `ctrl+t` was a no-op — FIXED.**
On a flat 15-item list, collapsed paints 7–8 items behind a `… N more todos` marker; after `ctrl+t` it paints all 15 and mounts the `ctrl+t to collapse` affordance, with the scroll region reaching the tail (`max_scroll_y=8` at h=20, `0` at h=40 where all 15 fit). The default `todo init items=[...]` shape now gets the same grow-and-reveal the phased path had — expand feels like an action, not a dead key.

**U4 / U5 — DEFERRED, acceptable.** #264 and #265 are both OPEN. With U2's keyboard scroll and the wheel both reaching the overflow, and expand genuinely revealing everything at normal heights, a "↓ N below" cue and a "12 of 18" position hint are discoverability/orientation polish, not friction that blocks the task. Fine to defer.

---

**New finding**

- **U6 (MINOR)** — At h=12 (flat + phased) and h=13 (phased), `ctrl+t` paints the header and the `ctrl+t to collapse` affordance but **zero todo rows**, where the collapsed view showed one item — expand is momentarily worse than collapse at the two shortest heights. Cause: the affordance is pinned as a row outside the scroll region, so at the floored budget it consumes the single row an item would occupy. Reachable via `ctrl+down` (so not data loss), and h=12 is the app's documented crowding floor. Fix direction: at the collapsed-equal floor, guarantee ≥1 ITEM row survives before the affordance claims its row (mirror `_fit_body`'s "trade the header for the first item"), or suppress the affordance widget when `budget - 1 < 1` and fall back to the collapsed marker line. Follow-up, not a blocker.

**Finding statuses**
- U1 (MAJOR, r1) → **FIXED** for h≥14 both shapes; residual boundary spun out as U6.
- U2 (MAJOR, r1) → **FIXED**.
- U3 (MAJOR, r1) → **FIXED**.
- U4 (NIT, r1) → **DEFERRED** (#264).
- U5 (NIT, r1) → **DEFERRED** (#265).
- U6 (MINOR, new) → **OPEN**, follow-up.

**Verdict: TERMINAL.** No BLOCKER and no MAJOR remain — the three round-1 MAJORs are fixed and verified on the driven app at head `0f8af3fb`. U6 is a MINOR boundary follow-up (recommend a ticket alongside #264/#265); U4/U5 stay deferred. The UX round is CLEAN of blocking findings.

damianvtran:

Implementing agent: coder subagent — Claude (via Local Operator), remediating the round-2 UX finding U6.
Scope: `origin/main..34b5e907` (new head `34b5e907`, was `0f8af3fb`).

Reproduced U6 on the real `OperatorApp` first (measured the exact geometry), fixed at the source, strengthened the guard test to catch it, re-captured a rendered frame, and re-ran the gates.

---

**U6 (MINOR, round 2) — expanded painted zero item rows at the floor — FIXED (`34b5e907`).**

Reproduced exactly as reported (initial scroll viewport, items visible collapsed → expanded):

```
flat   h=12: COLL 1 → EXP 0   (affordance stole the item's row)
flat   h=13: COLL 1 → EXP 1   ok
phased h=12: COLL 1 → EXP 0
phased h=13: COLL 1 → EXP 0   (root + phase header + affordance ate the 2-3 rows)
```

Cause is the one the reviewer pinned: the affordance is a row pinned OUTSIDE the scroll region, so the scroll viewport is `budget - 1`; at the 2–3 row floor that row went to the affordance and the first item fell just below the fold. Collapsed never hits it because its marker is a body line, not a pinned widget, so its scroll region is one row taller.

Fix — `TodoPanel._guarantee_expanded_item`, applied in `_build` only when expanded — mirrors collapsed's `_fit_body` "an item is worth more than the chrome" floor, cheapest lever first:
1. **Drop the redundant affordance row** when it's what buries the first item. Its hint is implied — the header names the list, `ctrl+t` still collapses, and the wheel / `ctrl+down` still reach the rest — so handing its row back to the scroll is a pure win.
2. At the very floor, **trim leading chrome** (keep the root `Todos · stage n/m` line, drop the phase header between it and the first item) until the item fits. The dropped phase header's count is implied by the root stage line, and the whole list is one `ctrl+down` away.

Only touches the tightest budgets — every h≥14 keeps the affordance untouched. Post-fix geometry (items in the initial viewport, collapsed → expanded):

```
flat   h=12: 1 → 1    phased h=12: 1 → 1
flat   h=13: 1 → 1    phased h=13: 1 → 1
flat   h=14: 2 → 2    phased h=14: 1 → 1   (affordance retained)
```

Overflow stays reachable: at h=12 flat expanded, `max_scroll_y=14` and `ctrl+down` walks the offset 0→2, so nothing is lost — the guard trims chrome, never an item. Composer stays 5 rows on screen. Rendered frames (an actual `- [` item now visible): `docs/evidence/todo-expand/tui-h12-expanded-flat-item.png`, `tui-h12-expanded-phased-item.png`.

**Tests (this can't regress silently):**
- New `test_expanded_shows_an_item_at_the_floor_flat_and_phased` — asserts an actual `- [` ITEM row sits in the INITIAL scroll viewport (top `scroll.size.height` body lines), not merely that some line was painted, at h=12 and h=13 for BOTH flat and phased; also that expanded ≥ collapsed and the overflow still scrolls.
- Strengthened `test_short_terminal_keeps_at_least_one_item_visible` to also drive `ctrl+t` and assert an item survives in the EXPANDED viewport (previously it only checked collapsed — the gap the review flagged).

**U4 / U5 — still DEFERRED (#264 / #265).** Both OPEN. With U2's keyboard scroll and the wheel both reaching the overflow and expand revealing everything at normal heights, the "↓ N below" cue (#264) and "12 of 18" position hint (#265) remain discoverability/orientation polish, not blockers.

---

**Gates & build (head `34b5e907`):**
- `flake8 .`, `black --check .`, `isort --check .`: clean.
- `pyright` on the changed source + test file: 0 errors.
- Targeted `-n0 test_todo_panel_phases.py test_band_panels.py test_usage_panel.py tests/unit/mobile`: **161 passed**.
- `cd local_operator/mobile/web && pnpm check && pnpm build`: tsc clean, vite build succeeded.

No blocker/major/minor findings left open on this head; U4/U5 stay deferred with tickets.

damianvtran:
### UX review — round 3

Reviewer: ux-reviewer subagent — Claude Sonnet 4.5 (via Local Operator), independent of the implementing agent.
Scope: origin/main..34b5e907

Narrow re-review confirming the **U6** fix (commit `34b5e907`, `TodoPanel._guarantee_expanded_item`). Driven on the **real `OperatorApp`** via `run_test`, seeded through the real todo store, inspecting the **actual scroll viewport** (rows visible before any scroll = body lines clipped to `scroll.size.height` at offset 0), not just the painted body Text. Rendered and eyeballed the h=12 expanded frame for both shapes.

---

**U6 — expand painted zero todo rows at the floor (h=12 flat+phased, h=13 phased) — FIXED (`34b5e907`).**

Measured the expanded viewport at h=12/13/14 for a flat 15-item list and a 3-phase 15-item list. Every case now has **≥1 real `- [` item row on screen before any scroll**:

```
scenario        budget  collapsed items  EXPANDED visible items  affordance         ctrl+down moves  composer
h12 flat          2           1                 1  [OK]          <hidden>           yes (0→2)        visible
h13 flat          3           1                 1  [OK]          ctrl+t to collapse yes (0→2)        visible
h14 flat          4           2                 2  [OK]          ctrl+t to collapse yes (0→3)        visible
h12 phased        2           1                 1  [OK]          <hidden>           yes (0→2)        visible
h13 phased        3           1                 1  [OK]          <hidden>           yes (0→3)        visible
h14 phased        4           1                 1  [OK]          ctrl+t to collapse yes (0→3)        visible
```

The two levers behave as designed:
- **Lever 1 (drop the affordance row)** fires at h=12 flat, h=12/13 phased — the pinned `ctrl+t to collapse` row is surrendered to the scroll so the first item survives. The affordance reads `<hidden>` in exactly those cells; `ctrl+t` still collapses and the wheel/`ctrl+down` still reach the rest, so nothing is stranded.
- **Lever 2 (trim leading chrome, keep the root)** fires at h=13 phased: viewport shows `Todos · stage 1/3` then `Foundation · 2/5` then `  - [x] found item 0`; at h=12 phased the phase header is trimmed too, leaving `Todos · stage 1/3` + `  - [x] found item 0`. The root `stage n/m` line — the one non-negotiable total — is always kept.

**Rendered h=12 frames confirm it visually** (expanded, after `ctrl+t`):
- Flat: `Todos · 3/15 resolved` / `- [x] flat item 0`, composer `Message Local Operator…` visible below.
- Phased: `Todos · stage 1/3` / `  - [x] found item 0`, composer visible below.

No regressions off the fix:
- **h≥14 keeps the affordance** — the guard is untouched there (`ctrl+t to collapse` present at h=14 both shapes), so this only ever touches the two shortest rows.
- **U1 (expand never worse than collapse)** holds at every measured height: expanded visible items ≥ collapsed items in all six cells.
- **U2 (keyboard overflow reach)** holds: `scroll_expanded(down=True)` returned `True` and walked the offset in every scenario, including the floor cells where the affordance is dropped.
- **U3 (no focus trap)** holds: `#todo-scroll` remains `can_focus=False`; the composer stayed visible throughout.

`tests/unit/tui/test_todo_panel_phases.py` — 31 passed.

---

**U4 / U5 — DEFERRED, still acceptable.** #264 (textual "↓ N below" overflow cue) and #265 (expanded footer position context) are both OPEN. With `ctrl+down` and the wheel both reaching the overflow and expand revealing everything at normal heights, these remain discoverability/orientation polish, not task-blocking friction. Fine as follow-ups.

---

**Verdict: TERMINAL — UX round fully CLEAN.** U6 is fixed and verified on the current head `34b5e907` on the real app; no BLOCKER and no MAJOR remains, and no residual MINOR from this PR is left open (U6 closed; U4/U5 tracked as #264/#265). The expand affordance now shows at least one todo at every height including the h=12 crowding floor, keeps the composer on screen, and leaves the overflow keyboard-reachable.

PR #262, opened 2026-08-22
feat(session): make the running model visible to the model, and announce switches

damianvtran:
## What & why

The model never knew **which model it was**. Two gaps:

1. **No model identity in context.** The system prompt carried no model name, so a session — or a delegated subagent naming itself in a review byline — could only guess which model it was running as. (This is the concrete thing that made a reviewer subagent write "Claude Opus 4.6" in its `Reviewer:` line when it was almost certainly on 4.8.)
2. **Model switches were invisible to the model.** A deliberate `/model` switch and a failover fallback both reached the UI band and the resume journal, but never the model's own context. So after a switch the model kept reasoning as the previous model: wrong self-identification, wrong assumed context window.

Both fixes apply to **top-level user sessions and subagent sessions** (reviewers, designers, scouts), since both render through `build_system_blocks`.

## Changes

**1. Model visible at session start** — `build_system_blocks` gains an optional `model_label`, rendered in the **env block** (`Model: <provider>/<model_id>`). It rides the byte-stable head rather than the volatile tail because the model does not change within a turn-loop; the Session passes its **live** `model_label` each turn, so a later switch is reflected next turn without rebuilding the closure. The provider argument is **additive**: a legacy zero-arg provider is detected by arity (`_callable_accepts_one_positional`) and still called with no argument, so no host or test is forced to grow a parameter.

**2. Switch announced into context** — a new `session_model_switch` custom message is rendered to the model as a user turn (same path as `session_incident`) and persisted so a resume replays it. Emitted from:
- `set_model` on a genuine `provider/model_id` change — **not** knob-only `/effort` or sampling changes, which take the same-pair early return;
- `_on_route_settled` on a failover fallback (marked **transient**, so the model does not adopt a temporary fallback as its identity) and on the return to primary.

Deduplicated (`_last_model_switch_announced`) so the two edges that can fire for one change do not double-announce.

## Files

- `local_operator/prompts_api.py` — `model_label` param + env-block line
- `local_operator/session_factory.py` — thread the live label through the factory provider; `build_initial_blocks` passes the spec label
- `local_operator/harness/subagent.py` — same for the subagent provider
- `local_operator/incidents.py` — `SESSION_MODEL_SWITCH_MESSAGE_TYPE` + `format_model_switch_message`
- `local_operator/session/session.py` — arity-adaptive `_system_blocks()`, `journal_model_switch`, render in `_default_convert_to_llm`, persist allow-list, emit hooks in `set_model` and `_on_route_settled`

## Testing evidence

Unit (`test_prompts_api.py`, `test_incidents.py`, `tests/unit/session/test_session.py`):
- env-block label present / absent-when-blank / does-not-perturb-the-cache-prefix
- switch formatter: deliberate (names old+new, "applies from now on"), transient fallback ("temporary fallback" + reason), no-previous-label
- arity detection (`_callable_accepts_one_positional`): one-arg certain, zero-arg certain, `*args`, unreadable-signature uncertain
- session paths: live label reaches the request's `system_blocks`; zero-arg provider still works; `set_model` journals exactly one switch and renders into the next request; a knob-only change journals nothing; dedup; failover route-settled journals a transient switch

Gates: `flake8`, `black==26.1.0 --check`, `isort --profile black`, `pyright` all clean on the changed files. Touched-area suites green: **1721 passed** across `tests/unit/{session,harness,model,providers}` + `test_prompts_api` + `test_incidents` + `test_session_factory`; `tests/unit/tui/test_status_line.py` 78 passed.

End-to-end: rendered a real session's initial blocks through `build_initial_blocks` (the factory path) and confirmed the env block carries `Model: anthropic/claude-opus-4-8`.

> Note on the full suite: `tests/unit` as a whole and `tests/unit/tui` timed out under concurrent-agent machine contention (not failures); one analytics multiprocess test (`test_parallel_processes_write_atomically`, untouched by this PR) is flaky under that contention and passes on isolated re-run. Change class: standard.

### DIFF
--- a/local_operator/harness/subagent.py
+++ b/local_operator/harness/subagent.py
@@ -912,7 +912,13 @@ async def _build_child_session(
     if mcp is not None and not restricted:
         tools = tools + mcp.tools
 
-    def system_blocks_provider() -> list[str]:
+    def system_blocks_provider(model_label: str = "") -> list[str]:
+        # ``model_label`` is passed by the child Session each turn (its own
+        # ``model_label``), which for a subagent is the resolved effort-tier
+        # override or the parent's model. Surfacing it lets a delegated
+        # reviewer/designer name the model it actually ran on in its byline
+        # instead of guessing.
+        #
         # Standard block layout. The lazy-knowledge tail carries the MCP
         # catalogue and nothing else: re-running semantic skill selection per
         # one-shot child would add cost without giving the parent a new durable
@@ -941,6 +947,7 @@ async def _build_child_session(
             goal=parent_session.goal,
             user_instructions=user_instructions,
             credentials=names,
+            model_label=model_label,
         )
 
     child = Session(
--- a/local_operator/incidents.py
+++ b/local_operator/incidents.py
@@ -27,6 +27,14 @@ from dataclasses import dataclass
 #: resumed replay see the same incident.
 SESSION_INCIDENT_MESSAGE_TYPE = "session_incident"
 
+#: Custom-message type journaled by the session when the running model changes
+#: (a deliberate ``set_model``, or a failover fallback to another model).
+#: Rendered to a user message the same way as an incident, so the model NOTICES
+#: it is now answering as a different model rather than only seeing a changed
+#: static "Model:" line in the system prompt. Persisted, so a resumed session
+#: replays the switch history too.
+SESSION_MODEL_SWITCH_MESSAGE_TYPE = "session_model_switch"
+
 #: Ordered (category, patterns) rules. First category whose pattern matches
 #: (case-insensitive) wins; order is specificity, not severity.
 _RULES: list[tuple[str, tuple[str, ...]]] = [
@@ -176,3 +184,44 @@ def classify_incident(raw: str, provider: str = "", model: str = "") -> Incident
 def format_incident_message(raw: str, provider: str = "", model: str = "") -> str:
     """One-call formatter for the rendered user-visible text."""
     return classify_incident(raw, provider, model).render()
+
+
+def format_model_switch_message(
+    new_label: str,
+    previous_label: str = "",
+    *,
+    reason: str = "",
+    transient: bool = False,
+) -> str:
+    """Render the model-switch text injected into the model's context.
+
+    ``new_label`` / ``previous_label`` are ``provider/model_id`` strings (the
+    same vocabulary the status band and the system-prompt ``Model:`` line use,
+    so two names for one object never read as two models). ``transient`` marks
+    a per-request failover fallback that may return to the primary at the next
+    boundary, as opposed to a deliberate switch that persists; ``reason``
+    carries the failover cause when there is one.
+
+    The message is phrased as present-tense state ("You are now running as X")
+    rather than a command, so the model treats it as context for the turns that
+    follow rather than an instruction to acknowledge.
+    """
+    if previous_label and previous_label != new_label:
+        head = f"[model switch] You are now running as {new_label} (was {previous_label})."
+    else:
+        head = f"[model switch] You are now running as {new_label}."
+    lines = [head]
+    if reason.strip():
+        lines.append(f"Reason: {reason.strip()[:200]}")
+    if transient:
+        lines.append(
+            "This is a temporary fallback for the current request; the session "
+            "may return to its primary model at a later turn. Capabilities and "
+            "context window may differ from the primary."
+        )
+    else:
+        lines.append(
+            "This applies from now on. Capabilities, context window, and tone "
+            "may differ from the previous model; act as the model you now are."
+        )
+    return "\n".join(lines)
--- a/local_operator/prompts_api.py
+++ b/local_operator/prompts_api.py
@@ -272,6 +272,7 @@ def build_system_blocks(
     credentials: Sequence[str] | None = None,
     team_brief: str = "",
     agent_brief: str = "",
+    model_label: str = "",
 ) -> list[str]:
     """Build the system prompt blocks; see the module docstring.
 
@@ -335,6 +336,18 @@ def build_system_blocks(
     env_block = f"Today is {date_str}."
     if env_details:
         env_block = f"{env_block}\n\n{env_details}"
+    if model_label.strip():
+        # The running model, so the assistant knows which model it currently is
+        # rather than guessing (a subagent naming itself in a review byline, a
+        # model reasoning about its own context window or capabilities). This
+        # rides the byte-stable env HEAD block, not the volatile tail, because
+        # within one turn-loop the model does not change: a deliberate
+        # ``set_model`` or a failover fallback takes effect at the NEXT turn
+        # boundary, which re-renders this block from the session's live model,
+        # and the switch itself is separately announced as a
+        # ``session_model_switch`` message so the model notices the change
+        # rather than only seeing a different static line.
+        env_block = f"{env_block}\n\nModel: {model_label.strip()}"
 
     tail = skills_block or "<skills/>"
     if goal:
--- a/local_operator/session/session.py
+++ b/local_operator/session/session.py
@@ -112,7 +112,10 @@ from local_operator.harness.wake import (
     format_wake_delivery_text,
 )
 from local_operator.imaging import rebound_oversize_image
-from local_operator.incidents import SESSION_INCIDENT_MESSAGE_TYPE
+from local_operator.incidents import (
+    SESSION_INCIDENT_MESSAGE_TYPE,
+    SESSION_MODEL_SWITCH_MESSAGE_TYPE,
+)
 from local_operator.session.goal import GoalState
 from local_operator.session.mcp_status import McpStartupOutcome
 from local_operator.session.naming import (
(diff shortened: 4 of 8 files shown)


damianvtran:
### Agent review — round 1
- Reviewer: reviewer subagent (Anthropic Claude — exact version not self-known; author verify independence by account)
- Scope: 1422d7412a91448fc243e3bbb493eeb1f963f389..9f7ba3d2e1b2cdad805626aa1bfd9d09cfdab8f2
- Mode: Comprehensive
- Delivery contract: model visible in-context at start + switches announced into context, additive, covers user + subagent sessions
- Applicable risk surfaces: prompt-cache layout, provider signature back-compat, live-model correctness, switch dedup, persistence/replay, async ordering
- Verdict: Approve

Both features land as specified and the correctness-critical paths hold up under scrutiny. I verified the following directly rather than inferring them:

- **Cache layout is correct.** `build_system_blocks` returns `[instructions, inventory, env_block, tail]`; the model line lives in `env_block` (index 2). The Anthropic breakpoint policy (`clients.py:1578`, `stable_count = min(4, len(blocks)-2)`) marks only indices 0–1 for a 4-block list, so the cached prefix ends at block 1 and a model-line change in block 2 never invalidates it. `test_model_label_does_not_perturb_the_stable_prefix` asserts `a[:2] == b[:2]` across two labels and fixed arity — a real assertion of the invariant. ✔
- **Back-compat holds for every enumerated provider.** Constructors: `session_factory._make_system_blocks_provider` (one-arg → certain), `subagent.system_blocks_provider` (one-arg → certain), `scripts/bench_cache_rate.py:118` (zero-arg → certain-not, called zero-arg), plus test doubles. `_callable_accepts_one_positional` classifies each correctly; the TypeError fallback is only reached on the *uncertain* (unreadable-signature) path, and Python rejects the label call at argument binding *before* the body runs, so a spurious retry cannot double-execute side effects. ✔
- **Live label at all three call sites** (`session.py:2898`, `:3193`, `:4609`) via `_system_blocks()` → `self.model_label` (property over `self._model`), so a `set_model` switch is reflected next turn. ✔
- **Emit points are right.** `set_model` emits only on the genuine-switch tail (both early returns above it — explicit-reselection and same-pair-knob — short-circuit; `test_a_knob_only_change_does_not_journal_a_switch` confirms). `_on_route_settled` journals `transient=False` on return-to-primary and `transient=True` on fallback, with the unresolvable-spec `return` sitting above the journal. Old/new labels are ordered correctly. ✔
- **Dedup is sound.** Keyed on `(new_label, transient)`; a legitimate A→B→A→B re-switch still announces (last key differs each edge), and the set_model+route-settled pair for one withdrawal collapses to one. `test_model_switch_record_deduplicates` covers both directions. ✔

I ran `tests/unit/test_prompts_api.py`, `test_incidents.py`, and the model-switch subset of `tests/unit/session/test_session.py` — all green — and `flake8`/`black --check` clean on the changed files.

| ID | Severity | Contract / risk | Evidence & required action |
|----|----------|-----------------|----------------------------|
| R1 | MINOR | persistence/replay | `SESSION_MODEL_SWITCH_MESSAGE_TYPE` is in `_PERSISTABLE_CUSTOM_TYPES` (`session.py:495`) with **no transient exemption**, so a `transient=True` fallback record (`session.py:3040`) is written to the transcript. If a session ends *while on a fallback* (crash/dispose before the return-to-primary edge fires), resume replays "You are now running as B … temporary fallback … may return to primary" with no matching recovery message, while the env block authoritatively says `Model: A`. The env line is authoritative so this is confusing history, not breakage. Consider: skip persistence for `transient=True` (a transient fallback that outlives the process is by definition stale), or drop trailing unresolved transient switch records on restore. |
| R2 | MINOR | async ordering | `set_model` fires `journal_model_switch` via `_spawn_background` (`session.py:1918`) with the comment "the record lands before the next turn reads context." That ordering is not structurally guaranteed — the journal task and the next turn are both loop-scheduled and the journal yields at `await append_message` (`session.py:3591`) before appending to `_context.messages`. In the (benign) racing case the announcement arrives one turn late; the env `Model:` line is already correct via the live label, so identity is never wrong — only the "you just switched" notice can slip a turn. Soften the comment to state the worst case, or await the append on a path the caller can order. |
| R3 | NIT | prose accuracy | `set_model` passes `reason="model switched"` (`session.py:1922`), which renders as a redundant `Reason: model switched` line on every deliberate switch (the head already says "You are now running as X (was Y)"). Pass no reason for deliberate switches so the `Reason:` line is reserved for failover causes. |
| R4 | NIT | prose accuracy (pre-existing, not in diff) | `clients.py:1568` docstring lists the block order as `[instructions, tool inventory, skills, env/date]`, but the actual order is `[instructions, inventory, env_block, tail(skills)]`. Out of scope for this PR but worth a follow-up since the cache reasoning in this diff depends on the env block being index 2, not last. |

**No BLOCKER and no MAJOR findings.** This round is **TERMINAL** — R1/R2 are minors and R3/R4 are nits, all follow-ups, not a reason for another round. Agreed that no design round is needed: the change alters prompt text the model reads, not a rendered product surface.

damianvtran:

New head: `8f2ed60e161521181cd91d3df25e6e7ee485bcf8` (was `9f7ba3d2`).

- **R1 (minor) — Fixed in `8f2ed60e161521181cd91d3df25e6e7ee485bcf8`**: transient failover switch records are now live-context-only, never persisted. `_is_persistable_message` returns `False` for a `session_model_switch` whose `details["transient"]` is truthy, and `journal_model_switch` guards the transcript append on `_is_persistable_message`. A deliberate/return-to-primary switch still persists and replays. Covered by two new tests: the transient record is absent from the transcript dump (and `_is_persistable_message` returns `False`), and a non-transient switch is present.
- **R2 (minor) — Fixed in `8f2ed60e161521181cd91d3df25e6e7ee485bcf8`**: reworded the ordering comment on `journal_model_switch` and the `set_model` emit to state the real worst case — the env-block `Model:` line carries the live identity, so a background journal that slips can only delay the "you just switched" notice by a turn, never make the identity wrong.
- **R3 (nit) — Fixed in `8f2ed60e161521181cd91d3df25e6e7ee485bcf8`**: dropped `reason="model switched"` on the deliberate-switch emit; the message head already reads "now running as X (was Y)", so the `Reason:` line is reserved for failover causes.
- **R4 (nit) — Deferred (out of scope, pre-existing)**: the `clients.py:1568` block-order docstring mismatch predates this diff and is not touched by it. Noted as a follow-up rather than expanding this PR.

Gates re-run clean on the new head (flake8, black==26.1.0, isort, pyright); session + incidents + prompts suites: 140 passed.

damianvtran:
### Agent review — round 2
- Reviewer: reviewer subagent (Anthropic Claude — exact version not self-known; author verify independence by account)
- Scope: 9310bf84c876c1c97f6e2421f45f4c9ceee9195a..8f2ed60e161521181cd91d3df25e6e7ee485bcf8
- Delta reviewed: 9f7ba3d2..8f2ed60e
- Applicable risk surfaces: shared persistence predicate regression, transient-record handling, dedup, cache layout
- Verdict: Approve

Convergence review of the round-1 remediation delta (`9f7ba3d2..8f2ed60e`). I re-read round 1, the remediation note, and diffed only the delta; the change touches exactly four things — `_is_persistable_message`, the `set_model` emit, `journal_model_switch`, and tests/comments. I verified each remediation claim against the code as written and confirmed the round-1-clean paths still hold.

**R1 — transient records now live-only, verified correct and non-regressing.**
- `_is_persistable_message` (`session.py:580-585`) returns `False` **only** for a `CustomMessage` whose `custom_type == SESSION_MODEL_SWITCH_MESSAGE_TYPE` **and** whose `details.get("transient")` is truthy; every other custom message still falls through to the `_PERSISTABLE_CUSTOM_TYPES` allow-list, and plain `Message`s still return `True`. Incidents and non-transient (deliberate / return-to-primary) switches still persist.
- `journal_model_switch` (`session.py:3614-3617`) still **always** appends to `self._context.messages`; only the `_transcript.append_message` call is now gated on `_is_persistable_message(message)`. Live context is unaffected.
- **Shared-predicate regression check (the real MAJOR risk): clean.** The durability flush (`_persist_new_messages`, `session.py:3866`) and the cancelled-child writer (`subagent._persist_inflight`) share this predicate. The new branch only fires for `session_model_switch` + truthy `transient`, so compaction markers (`compaction_summary`) and todo reminders — neither of which carries a truthy `transient` and neither of which is `session_model_switch` — are excluded exactly as before via the unchanged allow-list. `test_the_mid_turn_flush_never_persists_a_compaction_marker` and `test_a_cancelled_subagent_never_persists_a_marker_or_reminder` both still pass and still assert on transcript ENTRIES (not duplicate-id counts), so they remain load-bearing. Two new tests pin the R1 boundary (transient absent from transcript dump; non-transient present).

**R2 — comment softening accurate.** The reworded docstring on `journal_model_switch` (`session.py:3576-3591`) and the `set_model` emit comment (`session.py:1925-1934`) now state the real worst case — the env-block `Model:` line carries live identity, so a slipped background journal delays the "you just switched" notice by at most a turn and never makes identity wrong. No code claim that is now false.

**R3 — deliberate `reason` dropped, failover reasons intact.** The `set_model` emit (`session.py:1930-1938`) no longer passes `reason`, so it defaults to `""`. `format_model_switch_message` (`incidents.py:215`) appends the `Reason:` line only when `reason.strip()` is non-empty, so the deliberate switch now renders no `Reason:` line. Both failover paths still pass their reason: return-to-primary passes `reason or "returned to primary model"` (`session.py:3025`) and the outbound fallback passes `reason=reason` (`session.py:3056`).

**Independent KeyError check on `details["transient"]`: safe.** Every read is `details.get("transient")` (predicate `session.py:582`); the dedup key and payload use the local `transient` parameter, not a dict subscript. A non-switch custom message reaching the predicate takes the `.get(...)` path (returns `None` → falsy) and there is no bare `details["transient"]` subscript anywhere, so a KeyError is impossible.

**Round-1-clean invariants re-confirmed (delta does not touch them):** the delta modifies neither `build_system_blocks`, the Anthropic breakpoint policy, the three live-label call sites, the dedup key `(new_label, transient)`, nor the provider arity path — so cache layout (model line in env block index 2), live label, dedup, and back-compat arity are unchanged by this remediation.

I ran `tests/unit/session/test_session.py tests/unit/test_incidents.py tests/unit/test_prompts_api.py -q` → **140 passed**, and the R1-critical subset (both flush-regression tests + the transient/deliberate/dedup/knob switch tests) green.

| ID | Severity | Contract / risk | Resolution |
|----|----------|-----------------|------------|
| R1 | MINOR | persistence/replay | **Resolved in `8f2ed60e`.** Transient fallback records are live-context-only; predicate change is correctly scoped and does not regress the shared flush/cancel paths (verified against both regression tests). |
| R2 | MINOR | async ordering | **Resolved in `8f2ed60e`.** Comment now states the accurate worst case; no false code claim. |
| R3 | NIT | prose accuracy | **Resolved in `8f2ed60e`.** Deliberate switch emits no `reason`; formatter suppresses the `Reason:` line; failover paths still supply their cause. |
| R4 | NIT | prose accuracy (pre-existing) | **Deferred** by author as out-of-scope pre-existing docstring drift at `clients.py:1568`. Agreed — a follow-up, not a blocker. |

**No BLOCKER and no MAJOR findings; no new findings.** All four round-1 items are resolved or appropriately deferred, the remediation introduced no regression on the round-1-clean paths, and the reviewed head equals the current head (`8f2ed60e`). This round is **TERMINAL** — nothing remains to converge.

PR #152, opened 2026-08-18
fix(session): close the startup claim race in the never-delete retention model

damianvtran:
## Summary

Closes the residual **startup claim race** in session retention, layered onto main's never-delete model.

> **Scope change (why this diff is smaller than the earlier review rounds).**
> While this PR sat open, main merged #192 ("never delete session transcripts"), which **rewrote retention**: it retired the age/count/byte ceilings entirely and made `sweep_sessions` reap **only literally-empty directories** — a more radical version of the fix this PR originally targeted. The PR's original ceiling-based machinery (`_candidates`, the count/byte/age eviction loop, `bytes_on_disk`, the live-bytes warning) is therefore **dead code on the current model and has been dropped** — this PR does **not** revert #192. What remains is the piece #192 does *not* cover, described below. Because the shipped change is now materially different from what rounds 1–6 reviewed (those were against the ceiling model), a **fresh agent review round on the current head is required**.

## The bug that survives #192

`sweep_sessions` reaps empty directories, but **every live session is empty for the instant between its `mkdir` and its first append**. On a machine running several sessions at once (the normal way this harness is used), another session's startup sweep can land in that window and delete a directory whose owner is about to write to it — reproducing the exact `FileNotFoundError: .../transcript.jsonl` kill retention exists to prevent. #192's content-based protection cannot cover this window, because in it the directory genuinely holds nothing.

## What this PR does

- **Claim marker (liveness, not content).** A starting session writes `.session.pid` into its directory via `claim_session()` **before anything else creates it**, and the sweep skips any empty directory a live process still owns (`_is_claimed`). The claim is re-checked **immediately before each `rmtree`** so a claim landing mid-sweep is honoured (round-6 **L1** + **L2**). Crucially the marker is *liveness*: a dead owner's leftover marker does **not** keep its empty directory alive (it is excluded from `_dir_size`), so hard-killed corpses are still reclaimed and the reap never leaks. On platforms where liveness cannot be probed (Windows), a claim is bounded by `CLAIM_TRUST_S` from the later of the marker and the last write.
- **Caller ordering (round-6 L1).** `_prepare` and `_build_child_session` claim *first*, ahead of their own `mkdir`/`Transcript()`; `Session.dispose` releases the claim so an empty directory becomes reapable when its run ends. The defect was an ordering one at the caller — not in `claim_session` — so the fix is there.
[...]

PR #269, opened 2026-08-23
feat(resume): find sessions by title, past names, and soft matching

damianvtran:
## Why

The `/resume` picker could not find a session by a name it plainly carried. The reported case: a conversation titled **"Improve ADM Classifier Throughput"** (`bda7b76d34e0`, a 2.9 MB transcript) was unreachable through `/resume` search — the only way to find it was a manual transcript search of a user message.

Root cause, verified against the real store, was three distinct gaps:

1. **The canonical title was invisible.** `resume.py::stored_session_title` scans only a HEAD window (first `TITLE_SCAN_BYTES`) and a TAIL window (last `TITLE_SCAN_BYTES`). This is a *topic-pivot* session — it opened about an article-search load test and was auto-renamed to the ADM classifier topic ~88% of the way in — so every `conversation_name` entry (and the word "classifier" itself) sits in the **untouched middle** at byte ~2.52M. `stored_session_title` returned `""`, the row fell back to the abandoned opening topic, and the 256 KB body digest never saw the new subject either. `resume.py`'s own comment already named the fix: **journal the title to a sidecar** the way `origin.json` does.
2. **Past names were not searchable.** A session renamed over its life could only be found by its *current* name, never by any name it previously carried.
3. **No tolerance for typos.** Search was exact substring only — `classifer` found nothing.

Version bump is **MINOR** (0.24.3 → **0.25.0**): new search capability, no breaking change.

### Title always found — `title.json` sidecar (`resume.py`, `session/session.py`)
`_persist_conversation_name` now writes a `title.json` sidecar (atomic pid-temp + `os.replace`, **directory mtime preserved** so a rename never disturbs recency ordering) on the same event it journals the title. The sidecar accumulates a deduped `names[]` of **every** name the session has ever carried, plus the in-force `text` and its `user_set` flag. `stored_session_title` reads it as an **O(1) stat + small read fast path** before falling back to the window scan — closing the middle-of-transcript gap for good. `backfill_session_titles` stamps pre-existing sessions (mirrors `backfill_session_origins`), wired beside the origin backfill at both startup sites (`cli.py`, `session_factory.py`).

### Body + past names searchable (`session/search_index.py`)
`INDEX_VERSION` 1→2; `build_index` folds the title and past names into each session's digest and adds the `title.json` mtime to the cache freshness signature, so a rename re-digests exactly that one session. Recall is repaired by folding the title in — **not** by enlarging the defended 256 KB `SCAN_BYTES` window.

### Soft matching (`session/search_index.py`)
New `soft_search_digests` — bounded pure-Python: prefix match, token-AND, and edit-distance ≤2 for tokens ≥4. **No new dependency** (no rapidfuzz). Microseconds over the in-memory digests, so it stays on the per-keystroke path.

### Relevance ranking without cursor jump (`tui/widgets/session_picker.py`)
`rank_rows` orders matches name > id > body > soft with a recency tie-break. Ranking applies **only when a query is active**; an empty query keeps recency order, and `filter_rows` stays a pure membership filter. Because ordering is a function of the whole query and is recomputed only on query change — the same event that already re-homes the cursor to the first row — a row never reorders under the cursor for a *fixed* query. The body-match marker now also covers soft and past-name hits.

## Impact
- No breaking changes, no schema change to existing fields. The sidecar is additive and self-healing via backfill.
- Semantic/embedding search was **explicitly scoped out** as a user-approved opt-in follow-up: a per-keystroke embedding call breaks the picker's synchronous, offline, dependency-light contract. Full-text recall past 256 KB is likewise a documented follow-up.

### Regression against the REAL ADM session (read-only; store untouched)
Ran the actual code paths against `~/.local-operator/sessions/bda7b76d34e0` (copied read-only — verified no `title.json` was written into the real store):
```
BEFORE: stored_session_title -> ''
        'classifier'/'throughput'/'improve adm' in body digest? False
        search 'classifier' / 'throughput' / 'Improve ADM' -> False
AFTER:  backfill wrote 1;  stored_session_title -> 'Improve ADM Classifier Throughput'
        search 'classifier' / 'throughput' / 'Improve ADM' -> True
        soft  'classifer' / 'throughput classifier'        -> True
```
A synthetic transcript > 2×`TITLE_SCAN_BYTES` with `conversation_name` entries only in the middle reproduces the same before→after and is pinned as a unit test.

### Visual (real `OperatorApp`, not the CSS-less `_PickerHost`)
- Query `throughput`, before vs after: **byte-identical** SVG — row rendering, columns, and marker unregressed by the new path.
- Typo query `classifer`, origin/main vs this branch: origin/main renders **"no session matches that filter" (0 of 4)**; this branch shows the ADM row **marked, 1 of 4**, columns aligned, no reflow. All four frames rendered and viewed.

### Automated
- `tests/unit/test_stored_session_title.py` — sidecar round-trip + middle-of-transcript regression.
- `tests/unit/session/test_search_index.py` — title/past-name fold, pivot case, soft-match bounds, `INDEX_VERSION` discard.
- `tests/unit/tui/test_session_picker.py` — ranking invariants, cursor re-home, marker coverage.
- `tests/unit/session/test_conversation_name_persistence.py` — sidecar persistence round-trip + backfill.
- Gates whole-tree: flake8, black==26.1.0, isort --profile black, pyright (0 errors). Full unit suite **6251 passed, 7 skipped, 1 failed** — the single failure (`test_eval_tool.py::test_background_streams_output_while_it_runs`, a background-streaming timing assertion) is **pre-existing and unrelated**, failing identically on clean `origin/main`.

## Review history
_Agent review gate in progress — code + design rounds to follow on this PR._

### DIFF
--- a/local_operator/cli.py
+++ b/local_operator/cli.py
@@ -1885,6 +1885,7 @@ def main() -> int:
             from local_operator.resume import (
                 ResumeNotFound,
                 backfill_session_origins,
+                backfill_session_titles,
                 format_age,
                 recent_sessions,
                 resolve_resume_id,
@@ -1899,6 +1900,12 @@ def main() -> int:
             # costs a directory scan on the one path that cannot afford to be
             # wrong about which sessions are the user's.
             backfill_session_origins(config_dir())
+            # Stamp the title sidecar for pre-existing sessions in the same
+            # sweep, for the same reason: a session whose title sits in the
+            # untouched middle of a large transcript is unfindable by its own
+            # subject until this runs. Idempotent and stdlib-only, so it costs a
+            # bounded directory scan on the one path that opens the picker.
+            backfill_session_titles(config_dir())
 
             try:
                 args.resume = resolve_resume_id(config_dir(), str(args.resume))
--- a/local_operator/cli.py
+++ b/local_operator/cli.py
@@ -1904,7 +1904,10 @@ def main() -> int:
             # sweep, for the same reason: a session whose title sits in the
             # untouched middle of a large transcript is unfindable by its own
             # subject until this runs. Idempotent and stdlib-only, so it costs a
-            # bounded directory scan on the one path that opens the picker.
+            # bounded directory scan. session_factory._prepare backfills too (on
+            # ordinary session build); this branch runs it eagerly here because
+            # it answers `--resume` before any session is built, so the picker
+            # and `@latest` resolution above must see a stamped store first.
             backfill_session_titles(config_dir())
 
             try:
--- a/local_operator/resume.py
+++ b/local_operator/resume.py
@@ -346,13 +346,19 @@ def write_session_title(
     concurrent session rewrites it.
     """
     normalized = " ".join(text.split())
-    # Dedup while preserving first-seen order: ``dict.fromkeys`` is the stdlib
-    # ordered-set idiom. The in-force title is folded in as the newest name so a
-    # caller that passes only the prior history still gets a complete list.
+    # Whitespace-normalise every name the same way ``text`` is, so the sidecar's
+    # ``names`` and its ``text`` agree and the digest folds a name in exactly as
+    # the reader will match it. Dedup runs AFTER normalisation so two names that
+    # differ only in internal whitespace collapse to one. ``dict.fromkeys`` is
+    # the stdlib ordered-set idiom, preserving first-seen order; empties (a name
+    # that was all whitespace) are dropped. The in-force title is folded in as
+    # the newest name so a caller that passes only the prior history still gets
+    # a complete list.
+    normalized_past = [n for n in (" ".join(p.split()) for p in past_names) if n]
     names = (
-        list(dict.fromkeys([*past_names, normalized]))
+        list(dict.fromkeys([*normalized_past, normalized]))
         if normalized
-        else list(dict.fromkeys(past_names))
+        else list(dict.fromkeys(normalized_past))
     )
     payload = {"text": normalized, "user_set": user_set, "names": names}
     try:
--- a/local_operator/resume.py
+++ b/local_operator/resume.py
@@ -48,6 +48,18 @@ ORIGIN_NAME = "origin.json"
 #: scheduled run, a server-side session) is a new value and not a second file.
 ORIGIN_SUBAGENT = "subagent"
 
+#: Journals the session's title (and every name it has ever borne) beside the
+#: transcript, mirroring :data:`ORIGIN_NAME` exactly. A SIDECAR rather than a
+#: field in the JSONL for the same reason the origin marker is one: the picker's
+#: cost model is one bounded read per row, and the title in force can sit
+#: anywhere in a multi-megabyte transcript (the auto-name lands at turn 2, a
+#: mid-session ``/rename`` lands in the untouched middle — see
+#: :data:`TITLE_SCAN_BYTES` for the window-scan gap this closes). ``Path.stat``
+#: plus a sub-kilobyte read is O(1) in transcript size where the scan is not,
+#: so :func:`stored_session_title` consults this first and only falls back to
+#: the window scan for sessions written before the sidecar existed.
+TITLE_SIDECAR_NAME = "title.json"
+
 #: The two openings only the subagent runner can produce, used ONLY by the
 #: one-time backfill for directories that predate the marker.
 #:
@@ -94,22 +106,26 @@ _TITLE_CUSTOM_TYPE = "conversation_name"
 #: Reading both ends rather than the whole file is what keeps the cost per
 #: picker row bounded on a 6 MB transcript.
 #:
-#: **The known gap, stated honestly.** A rename made mid-conversation, then
-#: buried under a further 128 KB and never renamed again, falls between the two
-#: windows. The head still holds the ORIGINAL title, so that session is
-#: labelled with the name it was renamed *away from* rather than falling back
-#: to the opener. That is a stale name, and a stale name is stated with exactly
-#: the confidence of a correct one — so it is a real cost, not a rounding
-#: error, and this comment says so rather than claiming the scan can only ever
-#: miss.
+#: **The window gap this scan cannot close, and where it is closed instead.**
+#: A rename made mid-conversation, then buried under a further 128 KB and never
+#: renamed again, falls between the two windows: the head still holds the
+#: ORIGINAL title, so a window-only scan labels that session with the name it
+#: was renamed *away from*. Worse, a topic-pivot session auto-named late (its
+#: only titles sitting in the untouched middle of a multi-megabyte transcript)
+#: is invisible to both windows and reverts to its opening message — the
+#: reported failure of a session that could not be found by its own subject.
 #:
-#: It is accepted for now because the alternative is reading whole transcripts
-#: on the picker's synchronous path (400 ms against 64 ms across a real store),
-#: and because the case requires a mid-session rename specifically: an
-#: auto-named session has its title at the head, and a session renamed near the
-#: end has it in the tail. If it proves to bite, the fix is to journal the
-#: title to a sidecar the way ``mark_session_origin`` does — one stat and one
-#: small read, no size dependence at all — rather than widening this window.
+#: This scan does not try to fix that by widening: reading whole transcripts on
+#: the picker's synchronous path was measured at 400 ms against 64 ms across a
+#: real store. Instead the fix the previous comment here PRESCRIBED is now
+#: implemented — the title is journalled to a sidecar (``title.json``, see
+#: :func:`write_session_title`) the way :func:`mark_session_origin` journals
+#: origin, one stat and one small read with no size dependence at all.
+#: :func:`stored_session_title` consults that sidecar first, so this window
+#: scan is now the FALLBACK for sessions written before the sidecar existed and
+#: not yet reached by :func:`backfill_session_titles`, not the primary path.
+#: Both ends are still read because that fallback still wants the newest title
+#: it can reach on a pre-sidecar session.
 TITLE_SCAN_BYTES = 131_072
 
 #: Matches a journalled title row in raw JSONL, tolerant of whitespace after
@@ -231,6 +247,133 @@ def session_origin(session_dir: Path) -> str:
     return origin if isinstance(origin, str) else ""
 
 
+class SessionTitle(NamedTuple):
+    """The title sidecar's contents: the in-force name and every past one.
+
+    ``text`` is the name currently on the band and the terminal tab — the one
+    the user last saw and will search by. ``names`` is every distinct title the
+    session has ever carried, first-seen order, so a search matches a name the
+    session was renamed *away from* as well as its current one. ``user_set``
+    rides along for parity with the transcript entry: the picker does not need
+    it, but a future reader deciding rename precedence might, and it costs
+    nothing to keep.
+    """
+
+    text: str
+    user_set: bool
+    names: tuple[str, ...]
+
+
+def _read_title_sidecar(session_dir: Path) -> SessionTitle | None:
+    """Parse ``title.json``, or ``None`` when it is absent or unusable.
+
+    Tolerant for the same reason :func:`session_origin` is, and with the same
+    ``errors="replace"`` load-bearing detail: this runs over every session
+    directory to paint a picker, and :func:`write_session_title` writes
+    non-atomically, so a process killed mid-write can leave the file cut inside
+    a multi-byte character. A strict decode would raise ``UnicodeDecodeError``
+    (a ``ValueError``) and could sail past an ``except OSError`` and take the
+    whole picker down — the exact failure a corrupt ``origin.json`` once caused.
+    A missing or malformed sidecar yields ``None`` so the caller falls back to
+    the window scan, never an exception.
+    """
+    try:
+        raw = (session_dir / TITLE_SIDECAR_NAME).read_text(encoding="utf-8", errors="replace")
+    except OSError:
+        return None
+    try:
+        payload = json.loads(raw)
+    except ValueError:
+        return None
+    if not isinstance(payload, dict):
+        return None
+    text = payload.get("text")
+    if not isinstance(text, str):
+        return None
+    raw_names = payload.get("names")
+    names = (
+        tuple(name for name in raw_names if isinstance(name, str))
+        if (isinstance(raw_names, list))
+        else ()
+    )
+    return SessionTitle(
+        text=" ".join(text.split()),
+        user_set=bool(payload.get("user_set")),
+        names=names,
+    )
+
+
+def read_title_names(session_dir: Path) -> list[str]:
+    """Every name this session has borne, from the sidecar; ``[]`` when absent.
+
+    The source the digest folds in (see ``search_index.build_index``) so a
+    session is findable by any subject it was ever named for, not only its
+    current title. Empty for a pre-sidecar session, which the backfill sweep
+    (:func:`backfill_session_titles`) fills in once at startup.
+    """
+    sidecar = _read_title_sidecar(session_dir)
+    return list(sidecar.names) if sidecar else []
+
+
+def write_session_title(
+    session_dir: Path, text: str, *, user_set: bool, past_names: list[str]
+) -> None:
+    """Journal the in-force title (and every name ever borne) to a sidecar.
+
+    Why a sidecar: :func:`stored_session_title`'s window scan is blind to a
+    title in the middle of a large transcript (see :data:`TITLE_SCAN_BYTES` —
+    an auto-name at turn 2 pushed past the head window by an hour of work, or a
+    mid-session ``/rename`` buried between the two windows). One stat and one
+    small read here is O(1) in transcript size, closing that gap for good.
+
+    Best-effort by contract, exactly like :func:`mark_session_origin`: a title
+    is decoration, and a session that cannot write its sidecar (read-only
+    volume, full disk) must still RUN. The cost of the failure is a stale
+    picker label until the next rename or the backfill sweep, never a lost turn.
+
+    ``names`` accumulates: ``text`` is appended to ``past_names`` (deduped,
+    first-seen order preserved) so a search matches a name the session was
+    renamed away from. The list is authoritative for names seen since the
+    sidecar began; the one-time :func:`backfill_session_titles` sweep recovers
+    the complete history for sessions that predate it.
+
+    **The directory's mtime is preserved**, for the same reason
+    :func:`mark_session_origin` preserves it: recency ranks by the transcript's
+    mtime, but other readers (``os.listdir`` + ``stat`` listings, backups) look
+    at the directory, and journalling a title is bookkeeping ABOUT a session,
+    never activity IN it. The write is atomic (pid-named temp + ``replace``,
+    like ``search_index._save``) because the picker may read this file while a
+    concurrent session rewrites it.
+    """
+    normalized = " ".join(text.split())
+    # Dedup while preserving first-seen order: ``dict.fromkeys`` is the stdlib
+    # ordered-set idiom. The in-force title is folded in as the newest name so a
+    # caller that passes only the prior history still gets a complete list.
+    names = (
+        list(dict.fromkeys([*past_names, normalized]))
+        if normalized
+        else list(dict.fromkeys(past_names))
+    )
+    payload = {"text": normalized, "user_set": user_set, "names": names}
+    try:
+        try:
+            previous = session_dir.stat().st_mtime
+        except OSError:
+            previous = None
+        session_dir.mkdir(parents=True, exist_ok=True)
+        sidecar = session_dir / TITLE_SIDECAR_NAME
+        # The temp carries the writer's PID so two sessions writing at once do
+        # not ``replace`` a document the other is still filling — the same
+        # torn-write hazard ``search_index._save`` documents.
+        tmp = sidecar.with_suffix(f".{os.getpid()}.tmp")
+        tmp.write_text(json.dumps(payload), encoding="utf-8")
+        tmp.replace(sidecar)
+        if previous is not None:
+            os.utime(session_dir, (previous, previous))
+    except (OSError, TypeError, ValueError):
+        return
+
+
 def backfill_session_origins(config_dir: Path, limit: int = 500) -> int:
     """Stamp pre-existing subagent directories once, and return how many.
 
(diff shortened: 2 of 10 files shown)


damianvtran:
### Design review — round 1

Reviewer: designer (local-operator lopdev team) — anthropic/claude-opus-4
Scope: main..e1b8ac27ab43bbd7910c1cb1d16474e90cb0bc8b

Design/UX-visual review of the `/resume` picker surface. Frames rendered from the **real `OperatorApp`** (not the CSS-less `_PickerHost`) at 100×34 via `run_test` + `save_screenshot`, then viewed as images. Before-frames captured from a throwaway detached worktree at `origin/main` (`d62022e4`). Seven fixture rows with digests crafted so some rows match only on body / past-name / soft, never on the visible name.

**Scope note that reshaped this review:** the `"` body-match marker and exact-body matching **already ship on `main`**. This PR's *new* user-visible behaviour is the soft/typo path and past-name folding feeding that same marker. So the honest before/after is behavioural, not chrome:

| query | main (`d62022e4`) | this PR | visual delta |
|---|---|---|---|
| `classifier` (exact body) | `3 of 7`, marker on 2 non-name rows | **byte-identical** | none |
| `classifer` (typo) | `0 of 7` — "no session matches that filter" | `3 of 7`, marker on the soft rows | new results appear; layout identical to the exact-body frame |

Frames reviewed (saved to /tmp): `pick_empty` (baseline), `pick_exact` (exact name), `pick_body` (body/past-name only), `pick_soft` (typo `classifer`), `pick_nomatch` (`zxqwv`), plus `before_body` / `before_soft` from main.

**What renders correctly (verified against pixels + the plain-line dump):**
- **Marker vs footer glyph — no collision.** The `"` marker and the footer `·` separator are visually distinct at the two places they appear; the D4 concern that motivated the quote-over-`·` choice holds up in the frame.
- **Column alignment is intact and static.** Marked and unmarked rows share one name-start column (the marker slot is reserved list-wide, unmarked rows pad it), so the left edge of the name column — the field read down the list — is ragless. Cursor caret, age (right-aligned), and id columns all line up across every frame.
- **No reflow between settled frames.** Consecutive `pilot.pause()` captures are identical; nothing settles after first paint.
- **Counter is honest.** `7 sessions` → `1 of 7` (exact) → `3 of 7` (body) → `3 of 7` (soft) → `0 of 7` (no-match). Every count matches the rows drawn.
- **Selected-row cursor is legible.** The raised selection ground + muted caret reads clearly against the dimmed transcript; the exact-name selected row shows white `fg` name ink on the select tint, well above the dim siblings.
- **Query echo is honest.** The header echoes exactly what was typed in violet `label` ink, including the typo `classifer` — it does not silently "correct" to the matched term, which is the right call.
- **No-match state is clear** ("no session matches that filter"), distinct from the empty-list notice, and the footer/exit hint survives.

---

**D1 — MINOR — the marker says a row matched *elsewhere* but never *on what*, which is thinnest exactly in the new soft case.** `session_picker.py:153` / `matched_in_body` (L194–207). In `pick_soft.svg` the user types `classifer` and gets `Weekly ops sync notes` and `Investigate throughput regression` — rows whose visible names contain neither the typo nor "classifier". The `"` marker correctly stops them reading as arbitrary, but gives the user nothing to *recognize why*: no matched-term snippet, no "past name: …", no indication it was a typo-corrected hit vs a body hit vs a past name. To learn why a row is there the user must open it. This is adjacent to the deliberately-deferred semantic/richer-search follow-up, so I am **not** flagging it as missing scope — the marker meets its stated narrow job ("not arbitrary"). But the task asks whether a soft/body match gives the user enough to recognize *why*, and the honest answer is: it flags *that*, not *why*. **Fix (follow-up):** when richer search lands, carry a one-line matched-context snippet or a `past name` / `~typo` tag on marked rows. Recording as a follow-up, not a round blocker.

**D2 — MINOR — the `"` glyph's meaning is not discoverable from the card.** `session_picker.py:153` (`BODY_MATCH_MARKER`), footer hints `_footer_hints` (L978–1000). A first-time user sees a lone right-double-quote floating at the start of some rows with no open-quote, no text beside it, and no footer legend explaining it (the hint row lists move/page/filter/resume/cancel only). The code comment argues the quote reads as "something was said here"; in the rendered frame a bare, unpaired `"` reads at least as plausibly as a rendering artifact or a stray bullet. **Fix:** either add a compact legend when any row is marked (e.g. a `" matched inside` hint that drops first under width pressure, mirroring the existing drop discipline), or pick a glyph that carries "found inside" without needing to be learned. Not a blocker — the marker still does its anti-arbitrary job — but discoverability is a real gap for a mark this load-bearing.

**D3 — NIT — a soft-matched selected row can show the marker while its visible name obviously contains the concept.** `pick_soft.svg`: the top row `Improve ADM Classifier Throughput` carries the `"` marker under query `classifer`, because the *typed* string is not a literal substring of the name (it is a typo of a word that IS in the name). Technically exactly right — the marker keys on the query as typed — but it can momentarily read as "the marker is wrong, the word is right there." Acceptable as-is; a matched-term snippet (D1) would dissolve it. No action required.

---

**Verdict: no BLOCKER, no MAJOR. Design round 1 is TERMINAL.** The five states (empty, exact, body/past-name, typo/soft, no-match) all render with intact alignment, no reflow, an honest counter, a legible cursor, and a marker that does not collide with the footer glyph. D1 and D3 are recorded as follow-ups; D2 is a genuine but non-blocking discoverability MINOR worth a small footer-legend fix in this PR or the next.

damianvtran:
### Agent review — round 1

Reviewer: reviewer subagent (Claude Opus 4)
Scope: main..e1b8ac27ab43bbd7910c1cb1d16474e90cb0bc8b

Independent code review of the title-sidecar / soft-search change. I read the diff, reproduced the regression, and exercised the new paths directly against a synthetic ADM-shape fixture and the real store at `~/.local-operator/sessions/bda7b76d34e0` (read-only). Gates re-run from the prepared worktree.

**Regression reproduced and fix confirmed.** On the real 2.9 MB topic-pivot transcript, `stored_session_title` returns `''` (window scan misses the middle title, falls back to opener) before backfill; `_scan_all_titles` recovers all four journalled names ending in *"Improve ADM Classifier Throughput"*; after `backfill_session_titles` the sidecar exists and `stored_session_title` returns the correct in-force title. Directory `bda7b76d34e0` sorts at index 467 of 627, inside the 500 backfill cap, so it is reached on the first upgrade launch. On a synthetic >2×`TITLE_SCAN_BYTES` fixture: before→`''`, after→correct title, **dir mtime byte-identical across the write**, and a second backfill writes 0 (idempotent).

**Correctness checks that passed:**
- **Corrupt/partial/missing sidecar** all degrade to `None` → window-scan fallback, never an exception: empty, truncated JSON, non-dict, non-str `text`, cut multibyte (`errors="replace"`), missing `text`, and mixed-type `names` (non-strings dropped, order preserved). `stored_session_title` on a dir with a truncated sidecar and no transcript returns `''` cleanly.
- **Atomicity**: pid-named temp + `Path.replace` (atomic rename on the same filesystem); `os.utime` restores dir mtime only when the prior stat succeeded. `with_suffix(f".{pid}.tmp")` produces the intended sibling path.
- **Recency ordering unaffected**: `recent_sessions`/sort key on transcript mtime (resume.py:651), so the dir-mtime preservation is defence for external readers, matching the comment — not the ranking source.
- **names[] dedup + ordering**: `dict.fromkeys([*past_names, normalized])` — first-seen order, in-force title folded in as newest, deduped.
- **Freshness signature**: `title.json` mtime joins the cache signature; renaming s1 (transcript byte-identical) re-folds **only** s1's digest, s2 untouched. INDEX_VERSION 1→2 discards old cache via existing `_load` version-drop.
- **rank_rows invariant**: fixed query is byte-for-byte stable across repeated calls; empty query preserves recency order; tier order name>id>body>soft with stable recency tie-break; `filter_rows` remains a pure membership filter. `body_matches_markers` keyed on `_admitted` (union) so soft/past-name hits are marked.
- **Edit distance genuinely bounded**: length-gap early reject and per-row min short-circuit; `classifer`→`classifier` matches, `cat`/`dog` and gap-3 reject; short tokens (`adm`) excluded from fuzzy by the `_SOFT_MIN_TOKEN=4` floor.
- **Import guard intact**: importing `resume` pulls no engine/provider/langchain/openai/anthropic/asyncio module; the sidecar write in `session.py` is a function-local import. resume.py module imports are stdlib only (json, os, re, pathlib, typing).

**Gates (worktree, CI-pinned tools):**
- flake8 — clean
- black==26.1.0 --check — clean (453 files unchanged)
- isort 5.13.2 --check-only --profile black — clean
- 4 touched test modules (persistence, search_index, stored_session_title, session_picker) — **119 passed**

**Docstrings honestly describe scope.** Deferred semantic/embedding search is documented as an approved follow-up behind a dependency gate, not implied as done — not flagged as a defect per the review brief.

---

**Findings**

**MINOR — `soft_search_digests` per-keystroke cost is ~30–40 ms at real store scale, and the "microseconds" docstring is inaccurate there.** `search_index.py:373` claims "over the ~200 short digest tokens this runs against, the bounded DP is microseconds," and the module docstring (line 22) says "searches in microseconds." The real store holds **624** sessions (3× the assumed 200). Measured on 624 full-size digests: soft search is **30–40 ms per query change** vs. `search_digests` at ~0.65 ms. The dominant cost is not the DP — it is that `soft_search_digests` re-tokenizes **every digest on every call** (`haystack = set(_tokenize(digest))`, search_index.py:453): tokenizing 624 digests is ~30 ms; matching against pre-tokenized haystacks is ~2 ms. Because `visible_rows` recomputes only on a query change (not per repaint), this is ~30 ms per keystroke, not per paint — noticeable but not a stall, so MINOR not MAJOR. Two honest options: (a) soften the docstrings to "single-digit-to-tens of milliseconds over a few hundred digests," or (b) cache the per-digest token set alongside the digest so keystrokes cost ~2 ms. The bound-safety argument in the docstring is sound; only the latency figure is optimistic.

**NIT — comment says "one path that opens the picker" for a startup backfill that also runs on non-picker launches.** `cli.py:1905-1906` describes `backfill_session_titles` as costing "a bounded directory scan on the one path that opens the picker," but `session_factory._prepare` (session_factory.py:1029) also calls it on ordinary session prepare, not only `--resume`. The origin backfill has the same shape, so this is pre-existing framing; the wording just slightly understates where it runs.

**NIT — `_read_title_sidecar` collapses whitespace in `text` but not in the accumulated `names`.** resume.py:121 normalizes `text` via `" ".join(text.split())`, while `names` entries are passed through verbatim (only type-filtered). Writes normalize each name, so in practice all sidecars this code produces are already normalized; a hand-edited or foreign sidecar with internal whitespace in a name would fold into the digest unnormalized. Cosmetic — soft-search tokenizes anyway.

---

**Verdict: clean.** No blocker and no major findings. The regression is fixed and verified against the real transcript; atomicity, mtime preservation, corrupt-input tolerance, cache invalidation, the ranking invariant, and the import guard all hold. The two nits and the one minor (docstring latency claim / optional token-set caching) are follow-ups, not a reason for another round. **Round 1 is TERMINAL.**

damianvtran:

Implementer: coder (local-operator lopdev team) — anthropic/claude-opus-4
Scope: main..592422e2d3da5e87c1b04ede09b5f7579d80f69d (reviewed head e1b8ac27 → new head 592422e2d3da5e87c1b04ede09b5f7579d80f69d)

Addressing each finding from the round-1 agent code review.

---

**MINOR — `soft_search_digests` per-keystroke cost — FIXED** (`bbc17301`)

Fixed properly (caching), not by softening the docstring. Re-measured on the real store first: at **641 sessions / ~180k digest tokens** the stateless soft search cost **~75–185 ms per query change** (≈75 ms for a short/prefix query, ≈185 ms for a full typo-tolerant word) — worse than the review's estimate, and confirming the "microseconds" claim was only ever true at ~200 sessions.

The token-set cache the review suggested is *necessary but not sufficient*: I measured that re-tokenising every digest is only ~55 ms of the cost; the dominant ~130 ms is the edit-distance DP run **per digest token**. So the fix does both:

- **`SoftSearchIndex`** caches each digest's token set keyed on the **digest string**, so a re-digested session (rename/append → new digest) invalidates itself, and prunes entries for sessions no longer listed on every sync, so it **cannot grow past the live store** across the picker's life.
- The ~180k digest tokens dedup to only **~8.5k distinct words**, so each query token is resolved against that **vocabulary once** (with length- and first-letter buckets narrowing the DP candidate list), then each digest is answered by a set intersection.

**Behaviour is byte-identical** — verified equal to the old `soft_search_digests` across prefix / typo / word-order-AND / short-nonsense / empty queries on both synthetic fixtures and the real 641-session store.

**Re-measured after the fix (real store):**
- Warm per query change: **~6–20 ms** (prefix cheap, full typo dearer).
- One-time **~70 ms** on the first keystroke to build the token cache + vocabulary.
- In the real picker (200-row limit, one persistent index across progressive keystrokes): **~1–7.5 ms** per keystroke.

**Docstrings made true** — the module docstring, `_within_edit_distance`, and both `SoftSearchIndex`/`soft_search_digests` now state the measured ranges (~6–20 ms warm, ~70 ms cold, ~75–185 ms for the naive form) instead of "microseconds". The one remaining "microseconds" mention is an explicit historical note that the old claim held only at ~200 sessions.

New tests (`test_search_index.py`): parity with the stateless function across all tiers; **cache invalidation** when a digest string changes; **pruning + boundedness** when a session is dropped.

---

**NIT — cli.py "one path that opens the picker" comment — FIXED** (`b42f6526`)

Corrected: the comment now notes `session_factory._prepare` also backfills on ordinary session build, and that the CLI branch runs it eagerly only because it answers `--resume` before any session is built.

---

**NIT — `names` not whitespace-normalised like `text` — FIXED** (`b42f6526`)

`write_session_title` now normalises each name (`" ".join(n.split())`) the same way `text` is, with dedup applied **after** normalisation so two names differing only in internal whitespace collapse to one. New test in `test_conversation_name_persistence.py` pins this.

---

**Gates (CI-pinned tools, committed tree):** flake8 clean · black==26.1.0 clean (453 files) · isort==5.13.2 --profile black clean · pyright 0 errors/0 warnings · touched modules (search_index, stored_session_title, conversation_name_persistence, session_picker) **125 passed** · full suite **6257 passed, 7 skipped, 1 failed** — the one failure is the pre-existing unrelated `test_eval_tool.py::test_background_streams_output_while_it_runs`.

damianvtran:

Implementer: coder (local-operator lopdev team) — anthropic/claude-opus-4
Scope: main..592422e2d3da5e87c1b04ede09b5f7579d80f69d

Addressing the round-1 design findings.

---

**D2 — MINOR — `"` marker not discoverable — FIXED** (`592422e2`)

The footer now carries a legend, `" matched inside`, whenever a marked row is on screen — so the glyph is explained exactly where the reader already looks for meaning, and only when there's actually a marked row to explain (an empty or pure-name query never advertises a mark the user can't see).

It respects the existing footer drop discipline, but with a deliberate priority choice: the full key-hint row is already ~69 cells against the ~74-cell card, so a legend dropped *first* (as the review's "drops first under width pressure" suggested) would **never appear** — no fix at all. Instead it outranks the two genuinely disposable hints (`pgup/pgdn`, `type`) so it shows on a normal card, while still shedding **before** the movement and action keys, and it's dropped before the bare-key stage so it never survives as a lone unlabelled `"` (which would be the very artifact-looking mark this finding flagged). Verified across widths: 74 → legend shown (displaces `pgup/pgdn`), 60 → still shown, ≤50 → gone entirely (never a bare glyph), with `enter`/`esc` always surviving.

Frames rendered from the **real `OperatorApp`** (not `_PickerHost`, so `local_operator.tcss` applies) at 100×30 via `run_test` + `save_screenshot`, query `classifer` (a typo, so the three matching rows carry the marker on their bodies, not their visible names), then viewed as images:

- Before: `/tmp/d2_before.svg` — three rows carry a bare `"` with the footer listing only move/page/filter/resume/cancel; no explanation of the glyph.
- After: `/tmp/d2_after.svg` — footer leads with `" matched inside`; `pgup/pgdn page` dropped to make room; the legend's glyph is the same `"` drawn on the rows.

New tests in `test_session_picker.py`: the legend appears only when a row is marked; it drops before the movement/action keys and never survives as a bare glyph.

---

**D1 — MINOR — marker says *that* a row matched elsewhere but not *on what* — DEFERRED**

As the reviewer framed it: this is adjacent to the deliberately-deferred richer/semantic-search follow-up, and the fix ("carry a one-line matched-context snippet or a `past name`/`~typo` tag on marked rows") lands with that work, not this PR. The marker meets its stated narrow job (not-arbitrary), and D2 now makes that job discoverable. Deferred to the richer-search follow-up.

---

**D3 — NIT — a soft-matched selected row can show the marker while its name contains the concept — DEFERRED (no action)**

The reviewer marked this "Acceptable as-is; a matched-term snippet (D1) would dissolve it. No action required." Behaviour is technically correct (the marker keys on the query as typed). Folds into the D1 follow-up.

damianvtran:
### Agent review — round 2

Reviewer: reviewer (local-operator lopdev team) — anthropic/claude-opus-4
Scope: main..592422e2d3da5e87c1b04ede09b5f7579d80f69d

Re-review of the round-1 remediation delta (`bbc17301` perf/`SoftSearchIndex`, `b42f6526` docs+names normalization, `592422e2` footer legend) with a re-confirmation that the whole change is still sound on the new head. Round 1 was clean; the perf commit substantively rewrote the soft-search path, so this round re-verifies correctness, invalidation, bound, and the hot-path claims from scratch rather than trusting them.

**Verified myself (not taken on faith):**

- **CORRECTNESS — byte-identical matches.** Wrote an independent reference (`all(_token_soft_matches(nt, set(_tokenize(digest))) …)`) and compared `SoftSearchIndex.search` and the stateless `soft_search_digests` wrapper against it across 25 hand-picked queries spanning every tier (prefix `class`, typo `classifer`/`clasifier`, token-AND `throughput classifier`, short-token `adm`/`tok`, gap-3 rejects, empty, no-match `xyz`) over a 641-session store: **25/25 identical**, stateless wrapper **0 mismatches**. The `_resolve` bucketing (first-letter bucket for prefix, ±2 length buckets for edit-distance, `_SOFT_MIN_TOKEN` floor applied on both sides) reproduces `_token_soft_matches` exactly.

- **CACHE INVALIDATION — no stale vocabulary.** A 400-step fuzz that mutates the store between searches (re-digest an existing session, remove a session, add a session, or leave it fixed) and compares against the naive reference every step: **400/400 identical**. Targeted stale-vocab probe: a word present in only one digest disappears from results the moment that digest changes (`_sync` re-tokenises on digest-string change and `_rebuild_vocab` recomputes the whole vocab from the live token cache — no additive/dead-token path), reappears when re-added, and a removed session drops its tokens. The `changed` flag correctly forces a vocab rebuild on any add/remove/re-digest.

- **UNBOUNDED GROWTH — actually capped.** Churned 5000 distinct session ids through a fixed 50-session live store; `_tokens` settled at **50**, not 5050. Eviction keys on `sid not in digests` each sync, so the cache tracks the live store, not the history of digests seen.

- **HOT PATH — claims honest.** At 641 sessions / ~160k tokens / ~8.5k distinct vocab: cold first-keystroke build **33 ms** (docstring says ~70 ms — conservative, not aspirational), warm prefix **median 6.5 ms**, warm full-typo **median 11.7 ms** — inside the stated ~6–20 ms. Stateless one-shot **~34 ms** (docstring's ~75–185 ms is on larger prose bodies; the direction and shape are right). The `visible_rows` cache recomputes only on query change (`_filtered_for != _query`), so this is per-keystroke, off the per-paint path. Docstrings are honest.

- **ROUND-1 INVARIANTS intact.** `resume` import guard: importing it pulls **no** langchain/openai/anthropic/asyncio/executor module. `rank_rows` fixed-query stability: identical output across repeated calls; empty query preserves recency order; `filter_rows` remains a pure order-preserving membership filter. Freshness end-to-end holds: `build_index` folds the title-sidecar mtime into the `(size, mtime, title_mtime)` signature, so a rename changes the digest string the index keys on, which is what makes `_sync`'s digest-string invalidation meaningful.

- **b42f6526 / 592422e2.** `write_session_title` now normalises each past name before dedup, so dedup collapses whitespace-only differences and drops empties — matches the `text` normalisation. The footer legend is shed before the bare-key stage (`_FOOTER_DROP_ORDER_MARKED`), so a lone unlabelled `"` can never survive — the exact artifact D2 flagged is structurally impossible.

**Gates (worktree, CI-pinned tools):**
- flake8 — clean
- black==26.1.0 --check — clean (453 files unchanged)
- isort==5.13.2 --check-only --profile black — clean
- pyright — 0 errors, 0 warnings, 0 informations
- Touched modules (search_index, conversation_name_persistence, stored_session_title, session_picker) — **125 passed**

**Findings**

No blocker, major, minor, or nit findings this round. The round-1 MINOR (per-keystroke re-tokenisation cost) is genuinely fixed and the docstrings that state the new costs are truthful against measurement.

**Verdict: clean — TERMINAL.** No blocker and no major on the current head (592422e2). No follow-up round is warranted from the code side.

damianvtran:
### Design review — round 2

Reviewer: designer (local-operator lopdev team) — anthropic/claude-opus-4
Scope: main..592422e2d3da5e87c1b04ede09b5f7579d80f69d

Round-2 design review of the **D2 remediation only** (round-1 design round was otherwise TERMINAL). D2 added a footer legend so the `”` body/soft/past-name match marker is discoverable. Frames rendered from the **real `OperatorApp`** (not the CSS-less `_PickerHost`) via `run_test` + `save_screenshot`, then viewed as images. Seven fixture rows with digests crafted so rows `b2…`/`c3…` mention "classifier" only in their body — under the typo query `classifer` they match soft/body but **not** their visible name, so the marker appears on them.

Frames reviewed (saved to /tmp, screenshots in the same dir):
- `d2_soft_100.svg` — screen 100×30, query `classifer`, 3 marked rows.
- `d2_soft_narrow.svg` — screen 60×30, same query (card narrowed, legend under pressure).
- `d2_soft_tight.svg` — screen 46×30, same query (only action keys survive).
- `d2_nomarker_100.svg` — screen 100×30, query `billing`, 1 row, no marker.
- `d2_empty.svg` — screen 100×30, empty query, all 7 rows, no marker.

**What renders correctly (verified against pixels + the footer string dump):**

- **(a) Legend present, legible, and reads as explaining the glyph.** In `d2_soft_100` the footer leads with `” matched inside`, drawn in the same `dim`/`faint` ink as the other hints, and the legend's glyph is the **exact same `”`** painted at the start of the three marked rows above it — the reader's eye connects mark→label without a jump. Copy is honest: the marker means "matched inside the conversation" and the legend says `matched inside`.
- **No collision with the footer `·` separator.** The legend sits at the FRONT of the row, followed by the standard `· ` separator before `↑↓ move`. The `”` (a quote glyph) and the `·` (a mid-dot) are visually distinct and never adjacent in a way that could be misread; the D4 concern that motivated choosing a quote over `·` for the marker holds up here — the two glyphs never blur into one meaning.
- **(b) Sheds gracefully before the essential action keys.** Card-width probe on `_footer_hints(has_marked=True)` confirms the documented discipline exactly: ≥~54 cells → legend shown (displaces `pgup/pgdn`, then `type`); ≤50 → legend gone, `↑↓ move · enter resume · esc cancel` survives; ≤30 → `enter resume · esc cancel`. The rendered narrow (screen 60) and tight (screen 46) frames confirm `enter`/`esc` always survive while the legend has dropped. It outranks the two disposable hints so it can actually appear on a normal card (the full hint row is ~69 cells against the ~74-cell cap), which was the right call — dropping it "first" would have meant it never showed.
- **Never a bare unexplained glyph in the footer.** The legend is shed one stage before the bare-key fallback, so a lone `”` never survives in the footer — verified in the tight frame (`enter resume · esc cancel`, no stray quote) and by the `_FOOTER_DROP_ORDER_MARKED` ordering.
- **(c) No-marker states do not waste footer space.** `d2_nomarker_100` (query `billing`, one name-only match) shows **no** marker column and **no** legend — the footer restores `pgup/pgdn page`. `d2_empty` (empty query) likewise: no marker, no legend, full footer. The legend appears only when a marked row is actually on screen, exactly as intended.
- **No column reflow vs round 1.** The D2 delta (`git diff e1b8ac27..592422e2`) touches only the footer render loop (L961–L1040) and the hint constants; the row-draw path (`marker_col`, `_pad_cells`, `body_matched`) is untouched (0 diff lines). The frames confirm marked and unmarked rows still share one name-start column — the marker slot is reserved list-wide, so the name column's left edge is ragless in every frame, and the age/id columns line up across all five states.

**D-findings:** none. No BLOCKER, no MAJOR, no MINOR, no NIT on the remediation.

**D1 / D3 disposition — confirmed reasonable.** D1 (marker says *that* a row matched elsewhere, not *on what*) and D3 (a soft-matched selected row can carry the marker while its name contains the concept) were both deferred in round 1 to the richer/semantic-search follow-up. That disposition stands: D2 makes the marker's *meaning* discoverable, which is the fix in scope for this PR; carrying a matched-context snippet or `past name`/`~typo` tag is the D1 follow-up work and dissolves D3 with it. Not re-litigated.

**Verdict: CLEAN. No BLOCKER, no MAJOR. Design round 2 is TERMINAL.** The D2 legend is present, legible, correctly labels the marker glyph, does not collide with the footer separator, sheds gracefully before the action keys (never as a bare glyph), and never appears when no row is marked. No column reflow. D1/D3 remain follow-ups.

damianvtran:
### Rebase — review freshness carried to new head

Rebased onto latest `origin/main` (picked up #270, #235, #161) to clear a `pyproject.toml` version conflict. New head: **1930bc2ba00709f89c0c828de17e8f389860a8ab**.

The clean **Agent review — round 2** and **Design review — round 2** (previously scoped `main..592422e2`) carry forward unchanged: every reviewed source and test file is **byte-identical** between `592422e2` and `1930bc2b` — verified with `git diff 592422e2 1930bc2b -- <the 10 reviewed files>` → 0 lines. The rebase only replayed the three unrelated main commits and resolved the version line to `0.25.0` (next minor above main's new `0.24.5`); none of them touch `resume.py`, `search_index.py`, `session_picker.py`, `session.py`, `session_factory.py`, or `cli.py`.

Gates re-run green on the rebased tree: flake8 · black==26.1.0 · isort==5.13.2 (--profile black) · pyright 0/0/0 · the 4 touched test modules 125 passed.

**Both round-2 rounds are therefore fresh against head `1930bc2b`.**

damianvtran:
### Rebase — review freshness carried to new head (2)

Rebased again onto latest `origin/main` (picked up #226) to clear the `pyproject.toml` version line. New head: **e1bef857fc7a61b2aaff881ae7e39c767616dca2**.

The clean **Agent review — round 2** and **Design review — round 2** carry forward: every reviewed source and test file is **byte-identical** between the reviewed head `592422e2` and `e1bef857` (`git diff 592422e2 e1bef857 -- <the 10 reviewed files>` → 0 lines). The only rebased change is resolving the version to `0.25.0` (next minor above main's `0.24.6`); #226 touches provider cost accounting, not the search path.

**Both round-2 rounds are fresh against head `e1bef857`.**

damianvtran:
### Agent review — round 3 (rebase delta)

Scope: main..a7905a07794a41b0a47686c3aa72a94a52baa688

Scoped delta review of the three rebases onto `main` since the clean round-2 head `592422e2`. The last rebase merged #152 (startup claim-race fix) which added retention `claim_session`/`release_session` calls to the two files this PR also edits. All eight of this PR's own files are byte-identical to `592422e2` (verified `git diff 592422e2..HEAD` = 0 lines on each of `cli.py`, `resume.py`, `search_index.py`, `session_picker.py`, and the four test modules), so this review covers only the interleave in `session_factory.py` and `session/session.py`.

**1. `_prepare` ordering (session_factory.py:1019–1045)** — Coherent. Order is `claim_session(transcript_dir)` → `mkdir(exist_ok=True)` → `sweep_from_config` → `backfill_session_origins` → `backfill_session_titles`. The claim is written before the sweep runs, so a concurrent sibling sweep cannot reap this directory during its empty startup window (#152's invariant). `claim_session` itself does `mkdir(parents=True, exist_ok=True)`, and the explicit `mkdir` that follows is also `exist_ok=True`, so there is no double-mkdir failure — the second is a no-op in the claimed case and the creator in the named-agent/`--train` case (which is deliberately never claimed). The two title/origin backfills run *after* the sweep, operate on already-persisted sessions elsewhere in the store, and are unaffected by this session's own claim.

**2. `dispose` finally (session.py:5628–5696)** — Correct. `_flush_conversation_name` (which drives `_persist_conversation_name` → `write_session_title`, the sidecar write) runs inside the `try`, before background tasks are cancelled; `release_session` runs later in the `finally`. So the sidecar is always written before the claim is dropped. Dropping the claim cannot delete a directory carrying a title: `sweep_sessions` only reaps directories where `_dir_size == 0`, and `_dir_size` excludes only `_SIDECAR_NAMES = {".session.pid"}` — `title.json` (and `origin.json`, and the transcript) all count as content, so a directory with a title sidecar reads as non-empty and is untouchable regardless of claim state. Release only makes genuinely-empty (wrote-nothing) directories promptly reapable, which is its intended effect.

**3. PR's own logic** — Intact and unchanged by the rebase (title sidecar, `names[]`, `stored_session_title` fast path, `SoftSearchIndex`, `rank_rows` all byte-identical to `592422e2`).

**Gates (worktree `/tmp/lop-resume-search`, `.venv/bin/python`):**
- flake8 (touched files) — clean
- black==26.1.0 --check — clean (454 files unchanged)
- isort==5.13.2 --profile black --check-only — clean
- pyright (touched files) — 0 errors, 0 warnings
- Interleave test modules — `test_search_index.py` + `test_stored_session_title.py` + `test_session_picker.py` + `test_conversation_name_persistence.py` + `test_retention.py`: **148 passed** (TUI env `env -u NO_COLOR TERM=xterm-256color`)

No BLOCKER, no MAJOR, no MINOR, no NIT. The #152 interleave is clean; the sidecar-vs-retention interaction is safe by the `_dir_size` content rule. **Verdict: clean — this round is TERMINAL.**

PR #270, opened 2026-08-23
fix(tui): forgiving grab on the 1-cell transcript scrollbar + hover/grabbed ramp

damianvtran:
## Why

The transcript's vertical scrollbar is deliberately **1 cell wide** (`scrollbar-size-vertical: 1`, rationale D4/D21/D27 in `local_operator.tcss` — the reserved gutter keeps right-aligned columns stable and never reads as a border). But a 1-cell mouse target is easy to miss by a column, and **a miss lands on selectable transcript content**. That mousedown arms a Textual text selection; dragging it toward the top/bottom edge then trips Textual's *selection auto-scroll* — which is exactly the reported bug: **"the scrollbar grab highlights while you click and drag, resulting in a messy selection. It does scroll while you drag, but the highlight not being suppressed is problematic."**

Second report: **no hover highlight / grabbable cursor when hovering the bar, and no grabbed-state cursor while grabbed.** The rest→hover color step existed but was low-contrast on a 1-cell target, there was no grabbed color at all, and — found during this work — the grabbed state fell through to **Textual's off-palette default blue** (`#0178d4`), a color that appears nowhere else in the app.

A true thumb grab (landing exactly on the bar) was already handled correctly (`ScrollBar.ALLOW_SELECT = False` suppresses selection). The bug was purely the 1-cell miss and the missing/incorrect affordance.

Version bump is **PATCH** (0.24.3 → 0.24.4): two bug fixes to an existing control, no new capability.

### Bug 1 — forgiving grab (the messy-highlight fix)
- **`tui/app.py`, `TranscriptScreen`** — a `_forward_event` override redirects a **near-miss `MouseDown`** (exactly 1 cell left of the bar — `SCROLLBAR_GRAB_PAD=1` — and inside the thumb's vertical band, only when nothing is captured) onto the scrollbar column **before** the base class can arm a selection. The rebuilt event flows through unmodified Textual machinery, hits the `ScrollBar` (`allow_select=False`), so `_select_state` is set to `None` — **no selection is ever armed, no auto-scroll highlight can happen** — and the scrollbar's `@mouse.down: grab` starts a real thumb drag. Only the coordinate moves; button/modifiers are preserved, so downstream handling is identical to a true thumb click.
- Iterates `TranscriptView`s and skips any without `show_vertical_scrollbar`, so the hidden main view in subagent mode (display:none, no bar) is never targeted. Early-outs on capture-in-progress and on a two-cells-left content click, so it can never steal an ordinary selection.

This is the least-invasive correct fix: the **visual bar stays exactly 1 cell** (`scrollbar-size-vertical` untouched, asserted in test) — only the *input hit-test* is made forgiving. Widening the bar, subclassing `ScrollBar`, or a CSS-only change were all weighed and rejected (a CSS change can't fix it — the selection is armed in Python before any style applies).

### Bug 2 — hover / grabbed affordance
- **`local_operator.tcss`, `TranscriptView`** — a perceptible **neutral two-step ramp**: rest `$lo-edge` → hover `$lo-dim` → grabbed `$lo-muted` (a clear monotone `rest < hover < grabbed`). This replaces the old single loud `$lo-muted` hover and, critically, **sets `scrollbar-color-active` so the grabbed state no longer falls through to Textual's default blue.** No track-background color (that would paint a full-height column = the border D4 forbids). No literal hex (minimalism gate).
  - The accent (`$lo-accent`, the one green) was **deliberately not** used for grabbed: line ~367 reserves it for "a turn is live" and spending it on a scrollbar grab risks reading as agent activity. Deferred to design review (see review history).
- **Layer B (progressive enhancement):** `TranscriptView > ScrollBar { pointer: grab; }` — the open-hand cursor on hover, with Textual already flipping to `grabbing` on capture. Reaches **only** terminals that speak the Kitty pointer-shape protocol (OSC 22); a silent no-op elsewhere, which is why the color ramp (drawn text, works everywhere) is the real affordance and the pointer is a bonus.

## Impact
- No breaking changes. The visual scrollbar is unchanged at rest (still 1-cell `$lo-edge`); the only visual deltas are the hover/grabbed thumb color and — for Kitty terminals — the cursor. The reserved gutter and every right-aligned column width are untouched.

### Root cause reproduced, then fixed (real `OperatorApp`, headless pilot)
`tests/unit/tui/test_transcript_scrollbar_grab.py` — 5 assertions against the real app via `FakeSession`/`_factory`, `sb_x`/thumb computed from the compositor region (never hardcoded):
1. **Near-miss does not arm selection and grabs the bar** — mousedown 1 cell left of the bar → `_select_state is None`, `_selecting is False`, `mouse_captured` is the `ScrollBar` (the direct regression assertion for the reported bug).
2. **Near-miss drag scrolls without highlighting** — drag advances `scroll_y`, `selections == {}`.
3. **True thumb grab still works** — mousedown exactly on the bar → same capture/scroll (guards the redirect from breaking the direct path).
4. **Two cells left still selects** — `_select_state is not None` (guards the pad from swallowing content clicks).
5. **Short transcript, no scrollbar** — event untouched, no capture (guards the `show_vertical_scrollbar` gate).

`scrollbar_size_vertical == 1` asserted unchanged.

### Visual evidence — before/after, three states each
Captured from the **real `OperatorApp`** (loads `local_operator.tcss`), rest / hover / grabbed, before and after the change:

**Before** — hover jumps to a loud near-text `muted`, and **grabbed is off-palette blue** (`#0178d4`, Textual's default leaking through — the bug):

before ramp

**After** — clean neutral monotone edge → dim → muted, zero blue:

after ramp

Color-count per state confirms the ramp numerically (before: edge→muted→blue; after: edge→dim→muted, zero blue).

### Open item — Layer B pointer shape
The `pointer: grab` cursor can only be verified in a live terminal that speaks Kitty OSC 22; a headless SVG doesn't carry cursor shape. **Live-probed under `lop` in the cmux surface** (see review history / comments). If cmux does not proxy OSC 22 the cursor won't change, which is expected — the color ramp carries the affordance regardless.

### Gates (whole-tree, all clean)
- `flake8` → 0 · `black==26.1.0 --check` → 454 files unchanged · `isort --profile black` → 0 · `pyright` → 0 errors, 0 warnings
- `tests/unit/tui` → **2362 passed, 4 skipped** (serial and under xdist; new tests hardened against an in-process layout leak — see the `_scrollbar_region` docstring).

### DIFF
--- a/local_operator/tui/app.py
+++ b/local_operator/tui/app.py
@@ -942,30 +942,46 @@ class TranscriptScreen(Screen[None]):
     #: then trips Textual's selection auto-scroll (screen.py _check_auto_scroll).
     #: That is the reported "it scrolls but leaves a messy highlight" bug.
     #: Widening the INPUT hit area (not the visual bar) fixes it at the source
-    #: without touching the reserved gutter; 1 keeps the forgiveness inside the
-    #: transcript's own right padding so it can never swallow a content click two
-    #: cells in.
-    SCROLLBAR_GRAB_PAD = 1
+    #: without touching the reserved gutter.
+    #:
+    #: 2 (a 3-cell effective target: the bar plus the two pad columns to its
+    #: left) because a 1-cell forgiving zone is stingy — a review walk found it
+    #: too easy to still miss. The bar is flush-right against the reserved
+    #: gutter, so a miss can ONLY be leftward; there is no rightward content to
+    #: protect. A content click 3+ cells in is untouched, so the pad still
+    #: cannot steal an ordinary selection (guarded by the "three cells left
+    #: still selects" test). Do not raise this further without a fresh geometry
+    #: check: the pad is measured against the transcript's own right padding,
+    #: and a larger value would begin eating real content columns.
+    SCROLLBAR_GRAB_PAD = 2
 
     def _redirect_gutter_grab(self, event: events.MouseDown) -> events.MouseDown | None:
         """Rebuild a near-miss mousedown as a grab ON the scrollbar column.
 
         Returns a ``MouseDown`` relocated onto the scrollbar's x when ``event``
         fell within :attr:`SCROLLBAR_GRAB_PAD` cells to the LEFT of a visible
-        transcript scrollbar and inside its vertical extent; otherwise ``None``
-        (leave the event untouched). Only the coordinate moves —
-        button/modifiers/screen offset are preserved — so downstream handling is
-        identical to a true thumb click: the redirected event hits the
-        ``ScrollBar`` (``allow_select=False``), so ``Screen._forward_event``
-        sets ``_select_state`` to ``None`` and no selection is ever armed, then
-        the scrollbar's ``@mouse.down: grab`` starts a real thumb drag.
+        transcript scrollbar and inside the scrollbar's vertical extent (the
+        whole track, not just the thumb); otherwise ``None`` (leave the event
+        untouched). Only the coordinate moves — button/modifiers/screen offset
+        are preserved — so downstream handling is identical to a true thumb
+        click: the redirected event hits the ``ScrollBar``
+        (``allow_select=False``), so ``Screen._forward_event`` sets
+        ``_select_state`` to ``None`` and no selection is ever armed, then the
+        scrollbar's ``@mouse.down: grab`` starts a real grab. A near-miss over
+        the empty track (thumb elsewhere) redirects to a track page-scroll
+        rather than a selection, which is the correct, benign outcome.
+
+        The caller (:meth:`_forward_event`) has already established that no drag
+        is in progress, so this does not re-check ``mouse_captured``.
         """
-        # A drag already in progress owns the mouse; never re-grab mid-gesture.
-        if self.app.mouse_captured is not None:
-            return None
-        # Two transcripts exist in subagent mode (main hidden + child). A hidden
-        # view is display:none and reports no scrollbar, so skipping any view
-        # whose bar is not shown handles both without targeting an off-screen bar.
+        # Two transcripts exist in subagent mode (main hidden + child). Two gates
+        # keep the redirect off a bar the user cannot see:
+        #   - `show_vertical_scrollbar` skips a view that is not overflowing, so
+        #     a short transcript never claims the gutter column;
+        #   - a display:none view (the hidden main transcript) is NOT in the
+        #     compositor, so `find_widget` raises `NoWidget` and we skip it. Its
+        #     `show_vertical_scrollbar` can still read True while hidden, so the
+        #     compositor lookup — not that flag — is what excludes it.
         for view in self.query(TranscriptView):
             if not view.show_vertical_scrollbar:
                 continue
@@ -973,19 +989,24 @@ class TranscriptScreen(Screen[None]):
                 region = self._compositor.find_widget(view.vertical_scrollbar).region
             except NoWidget:
                 continue
-            # Only inside the thumb's vertical band — a mousedown above/below the
-            # bar is ordinary content and must stay selectable.
+            # Anywhere in the scrollbar's vertical extent — the whole TRACK, not
+            # only the thumb. A mousedown above/below the bar is ordinary content
+            # and must stay selectable; within the track's height a near-miss is
+            # a scroll gesture whichever way the thumb happens to be positioned.
             if not (region.y <= event.y < region.y + region.height):
                 continue
-            # Exactly the single pad column immediately left of the bar. Two
-            # cells left is untouched (guarded by test #4).
+            # The pad columns immediately left of the bar. A click
+            # SCROLLBAR_GRAB_PAD+1 cells in is untouched (guarded by the "three
+            # cells left still selects" test).
             if region.x - self.SCROLLBAR_GRAB_PAD <= event.x < region.x:
                 return events.MouseDown(
                     event.widget,
                     x=region.x,
                     y=event.y,
-                    delta_x=event._delta_x,
-                    delta_y=event._delta_y,
+                    # Public accessors (both 0 for a MouseDown); preserved so the
+                    # rebuilt event is indistinguishable from a true thumb click.
+                    delta_x=event.delta_x,
+                    delta_y=event.delta_y,
                     button=event.button,
                     shift=event.shift,
                     meta=event.meta,
--- a/local_operator/tui/app.py
+++ b/local_operator/tui/app.py
@@ -36,11 +36,13 @@ from rich.padding import Padding
 from rich.style import Style
 from rich.terminal_theme import TerminalTheme
 from rich.text import Text
+from textual import events
 from textual.actions import SkipAction
 from textual.app import App, ComposeResult
 from textual.binding import Binding
 from textual.containers import Container, Horizontal
 from textual.css.query import NoMatches
+from textual.errors import NoWidget
 from textual.events import (
     AppBlur,
     AppFocus,
@@ -932,6 +934,77 @@ class TranscriptScreen(Screen[None]):
 
     def action_copy_text(self) -> None:
         raise SkipAction()
+
+    #: How many cells LEFT of the 1-cell scrollbar still count as "grabbing the
+    #: bar". The bar is deliberately 1 cell (D4/D21/D27) so it never reads as a
+    #: right border, but a 1-cell mouse target is easy to miss by a column — and
+    #: a miss lands on selectable content, arming a text selection whose drag
+    #: then trips Textual's selection auto-scroll (screen.py _check_auto_scroll).
+    #: That is the reported "it scrolls but leaves a messy highlight" bug.
+    #: Widening the INPUT hit area (not the visual bar) fixes it at the source
+    #: without touching the reserved gutter; 1 keeps the forgiveness inside the
+    #: transcript's own right padding so it can never swallow a content click two
+    #: cells in.
+    SCROLLBAR_GRAB_PAD = 1
+
+    def _redirect_gutter_grab(self, event: events.MouseDown) -> events.MouseDown | None:
+        """Rebuild a near-miss mousedown as a grab ON the scrollbar column.
+
+        Returns a ``MouseDown`` relocated onto the scrollbar's x when ``event``
+        fell within :attr:`SCROLLBAR_GRAB_PAD` cells to the LEFT of a visible
+        transcript scrollbar and inside its vertical extent; otherwise ``None``
+        (leave the event untouched). Only the coordinate moves —
+        button/modifiers/screen offset are preserved — so downstream handling is
+        identical to a true thumb click: the redirected event hits the
+        ``ScrollBar`` (``allow_select=False``), so ``Screen._forward_event``
+        sets ``_select_state`` to ``None`` and no selection is ever armed, then
+        the scrollbar's ``@mouse.down: grab`` starts a real thumb drag.
+        """
+        # A drag already in progress owns the mouse; never re-grab mid-gesture.
+        if self.app.mouse_captured is not None:
+            return None
+        # Two transcripts exist in subagent mode (main hidden + child). A hidden
+        # view is display:none and reports no scrollbar, so skipping any view
+        # whose bar is not shown handles both without targeting an off-screen bar.
+        for view in self.query(TranscriptView):
+            if not view.show_vertical_scrollbar:
+                continue
+            try:
+                region = self._compositor.find_widget(view.vertical_scrollbar).region
+            except NoWidget:
+                continue
+            # Only inside the thumb's vertical band — a mousedown above/below the
+            # bar is ordinary content and must stay selectable.
+            if not (region.y <= event.y < region.y + region.height):
+                continue
+            # Exactly the single pad column immediately left of the bar. Two
+            # cells left is untouched (guarded by test #4).
+            if region.x - self.SCROLLBAR_GRAB_PAD <= event.x < region.x:
+                return events.MouseDown(
+                    event.widget,
+                    x=region.x,
+                    y=event.y,
+                    delta_x=event._delta_x,
+                    delta_y=event._delta_y,
+                    button=event.button,
+                    shift=event.shift,
+                    meta=event.meta,
+                    ctrl=event.ctrl,
+                    screen_x=region.x,
+                    screen_y=event.y,
+                    style=event.style,
+                )
+        return None
+
+    def _forward_event(self, event: events.Event) -> None:
+        # Give the 1-cell scrollbar a forgiving grab target BEFORE the base
+        # class can arm a text selection on the content column next to it.
+        # Hand the rebuilt event to super() — it owns event.stop()/forwarding.
+        if isinstance(event, events.MouseDown) and self.app.mouse_captured is None:
+            redirected = self._redirect_gutter_grab(event)
+            if redirected is not None:
+                event = redirected
+        super()._forward_event(event)
 
 
 class OperatorApp(App[None]):
--- a/local_operator/tui/local_operator.tcss
+++ b/local_operator/tui/local_operator.tcss
@@ -156,25 +156,26 @@ TranscriptView {
      * than the held state is backwards. `dim` hover -> `muted` grabbed is a
      * clear, monotone rest<hover<grabbed ramp on the 1-cell thumb.
      *
-     * The architect's proposal used `$lo-accent` (the one green) for grabbed,
-     * but line ~367 reserves the accent for "a turn is live" and nothing else;
-     * spending it on a scrollbar grab risks reading as agent activity. The
-     * accent option is DEFERRED to the design review, which makes the final
-     * call — this neutral ramp is the safe first pass. */
+     * The architect proposed `$lo-accent` (the one green) for grabbed; design
+     * review (round 1, D1) REJECTED it and settled on neutral `$lo-muted`. Line
+     * ~367 reserves the accent for "a turn is live" and nothing else, and a
+     * green thumb on a passive scroll gesture would fire that live-activity
+     * signal falsely and paint the highest-chroma colour on the reserved gutter
+     * (the "saturated column reads as a right border" failure D4 forbids). Keep
+     * this ramp neutral. */
     scrollbar-color-hover: $lo-dim;
     scrollbar-color-active: $lo-muted;
 }
 
-/* Layer B (progressive enhancement, Kitty pointer-shape / OSC 22 only): show
- * the open-hand cursor while hovering the 1-cell bar. Textual already flips
- * this to "grabbing" on capture and back on release (scrollbar.py
- * _on_mouse_capture / _on_mouse_release). It reaches ONLY terminals that speak
- * the Kitty pointer-shape protocol; elsewhere it is a silent no-op, which is
- * why the color ramp above — drawn text that works everywhere — is the real
- * affordance and this is a bonus, never the primary signal. */
-TranscriptView > ScrollBar {
-    pointer: grab;
-}
+/* Layer B (the open-hand hover cursor) is NOT declared here. A
+ * `TranscriptView > ScrollBar { pointer: grab; }` rule matches the selector but
+ * is dead at runtime: system scrollbars are styled through Textual's shared
+ * style-cache branch (`Stylesheet.update_nodes`), which never delivers this
+ * rule to the widget — the computed `pointer` stays `default` through compose,
+ * scroll and resize (verified in review round 1, M1). The pointer is instead
+ * set as an INLINE style in `TranscriptView.on_mount`, which bypasses that
+ * cache; see the comment there. The drawn colour ramp above is the real,
+ * everywhere-visible affordance regardless. */
 
 /* Blocks: zero outer margin, zero padding — no declarations here by design
  * (the minimalism defense asserts their absence; anything uniform here would
--- a/local_operator/tui/local_operator.tcss
+++ b/local_operator/tui/local_operator.tcss
@@ -150,7 +150,30 @@ TranscriptView {
      * border a borderless design must not have. `edge` keeps it structurally
      * present; hover still lights it to `dim`. */
     scrollbar-color: $lo-edge;
-    scrollbar-color-hover: $lo-muted;
+    /* Rest -> hover -> grabbed is a two-step neutral escalation, no accent.
+     * `dim` for hover, not the old `muted`: `muted` is near-text brightness, a
+     * large jump that read louder than a *grabbed* bar should — hover brighter
+     * than the held state is backwards. `dim` hover -> `muted` grabbed is a
+     * clear, monotone rest<hover<grabbed ramp on the 1-cell thumb.
+     *
+     * The architect's proposal used `$lo-accent` (the one green) for grabbed,
+     * but line ~367 reserves the accent for "a turn is live" and nothing else;
+     * spending it on a scrollbar grab risks reading as agent activity. The
+     * accent option is DEFERRED to the design review, which makes the final
+     * call — this neutral ramp is the safe first pass. */
+    scrollbar-color-hover: $lo-dim;
+    scrollbar-color-active: $lo-muted;
+}
+
+/* Layer B (progressive enhancement, Kitty pointer-shape / OSC 22 only): show
+ * the open-hand cursor while hovering the 1-cell bar. Textual already flips
+ * this to "grabbing" on capture and back on release (scrollbar.py
+ * _on_mouse_capture / _on_mouse_release). It reaches ONLY terminals that speak
+ * the Kitty pointer-shape protocol; elsewhere it is a silent no-op, which is
+ * why the color ramp above — drawn text that works everywhere — is the real
+ * affordance and this is a bonus, never the primary signal. */
+TranscriptView > ScrollBar {
+    pointer: grab;
 }
 
 /* Blocks: zero outer margin, zero padding — no declarations here by design
--- a/local_operator/tui/widgets/transcript.py
+++ b/local_operator/tui/widgets/transcript.py
@@ -1757,6 +1757,30 @@ class TranscriptView(ScrollableContainer):
         #: the two modes apart.
         self._pending_mounts: list[TranscriptBlock] | None = None
 
+    def on_mount(self) -> None:
+        """Give the system vertical scrollbar an open-hand hover cursor.
+
+        This is Layer B of the scrollbar affordance (Bug 2): the pointer shape a
+        Kitty-pointer-shape (OSC 22) terminal shows while hovering the bar, so
+        the 1-cell target reads as grabbable before the grab. Textual already
+        flips it to ``grabbing`` on capture and back on release
+        (``scrollbar.py`` ``_on_mouse_capture`` / ``_on_mouse_release``); this
+        only supplies the resting hover shape.
+
+        Set as an INLINE style rather than in ``local_operator.tcss`` on purpose:
+        a `TranscriptView > ScrollBar { pointer: grab; }` rule matches the
+        selector but is never delivered to the widget at runtime — system
+        scrollbars take Textual's shared style-cache branch
+        (``Stylesheet.update_nodes``), and the computed ``pointer`` stays
+        ``default`` through compose/scroll/resize (verified in review round 1,
+        M1). The inline style bypasses that cache and sticks across every natural
+        state. It reaches ONLY terminals that speak the pointer-shape protocol;
+        elsewhere it is a silent no-op, which is why the drawn colour ramp
+        (``scrollbar-color-*`` in the tcss) is the real, everywhere-visible
+        affordance and this is a progressive-enhancement bonus.
+        """
+        self.vertical_scrollbar.styles.pointer = "grab"
+
     def set_on_clear(self, hook: Callable[[], None] | None) -> None:
         """Install the hook fired after every :meth:`clear_blocks`."""
         self._on_clear = hook
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "local-operator"
-version = "0.24.3"
+version = "0.24.4"
 description = "A team of proactive AI assistants that can work together behind the scenes to help you get mundane tasks done so you can focus on the fun stuff."
 readme = "README.md"
 authors = [{ name = "Damian Tran", email = "damian@gominerva.com" }]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "local-operator"
-version = "0.24.4"
+version = "0.24.5"
 description = "A team of proactive AI assistants that can work together behind the scenes to help you get mundane tasks done so you can focus on the fun stuff."
 readme = "README.md"
 authors = [{ name = "Damian Tran", email = "damian@gominerva.com" }]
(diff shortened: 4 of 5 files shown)


damianvtran:
### Design review — round 1

Reviewer: designer (lopdev team) — Claude Opus 4
Scope: d62022e49ba866cee439ea43fad13a0524fcaeed..7565e01cccdd7b74e1ecc29b8800c46adbacc3ba

Reviewed the rendered frames from the real `OperatorApp` (before/after × rest/hover/grabbed, `docs/scratch/frames/`), not source alone. Sampled the actual thumb fill from each PNG to back the visual read with numbers.

**Measured thumb fill (RGB, nearest palette token):**

| state   | before                | after                  |
|---------|-----------------------|------------------------|
| rest    | (72,72,72) ~ `edge`   | (72,72,72) ~ `edge`    |
| hover   | (181,175,162) ~ `muted` | (131,124,109) ~ `dim`  |
| grabbed | (1,120,212) off-palette blue | (181,175,162) ~ `muted` |

The after ramp is **monotone increasing** — rest 72 → hover 131 → grabbed 181, two roughly equal ~50–60-luminance steps. The before ramp was broken twice: hover leapt to `muted` (louder than a held state should be), and grabbed fell through to Textual's default blue #0178d4, which is not in this palette at all.

---

**D1 — The deferred accent-vs-neutral call: neutral (`$lo-muted`) for grabbed is CORRECT. Accent is rejected.** (MINOR — decision resolved, no change)

The architect proposed `$lo-accent` (#38c96a, the one green) for grabbed; the coder held it neutral pending this review (`local_operator.tcss:159-163`). Final call: **stay neutral.** Grounded in the frames and the design language:

- The one-green rule (tcss ~line 367/390) reserves `$lo-accent` for "a turn is live" and nothing else. A green thumb appearing the instant a user grabs the scrollbar would collide with the single most load-bearing signal in the whole TUI — the user would see agent-activity green fire on a passive scroll gesture. That is a semantic regression, not an affordance win.
- The borderless language wants the right edge quiet. Accent green is the highest-chroma colour in the palette; painting it full-height-of-thumb on the reserved 1-cell gutter during a drag is exactly the "saturated column reads as a right border" failure D4 forbids — worse than the blue it replaces, because it also carries false meaning.
- The neutral ramp already does the job the accent was meant to do: the grabbed frame (`after/grabbed.svg.png`, thumb = `muted` 181) is visibly the brightest, most "engaged" of the three states. Escalation ordering is right — grabbed reads as most engaged — without spending the green.

Removing the off-palette blue is an unambiguous improvement: the grabbed thumb now sits inside the warm neutral family instead of the jarring cool #0178d4 that clashed with every other element.

**D2 — rest→hover→grabbed is perceptible at 1 cell.** (resolved, no change)

Confirmed in the stills, not just the numbers. `after/rest` shows a barely-there `edge` thumb (intentionally recessive per D4 — not a border); `after/hover` lifts it to a legible mid-tan `dim`; `after/grabbed` brightens to near-text `muted`. Each state is distinct from its neighbour by a full palette step. The hover→grabbed delta (`dim`→`muted`, +50) is the smaller of the two, but the two states are never on screen simultaneously (grabbed only exists while the mouse is captured), so the user judges each as an escalation from rest, not against each other — and both are plainly brighter than rest. Acceptable.

**D3 — Stays inside the design system: no border, no track fill.** (resolved, no change)

`scrollbar-background: $lo-bg` (tcss:144) paints the track the same colour as the island ground, so there is no visible track — confirmed in every frame, the gutter is invisible except for the thumb. `scrollbar-size-vertical: 1` is untouched, so the bar remains 1 cell and never reads as a right-hand border (D4/D21/D27 intact). All four thumb colours (`edge`/`dim`/`muted`) are palette tokens; no raw hex, no off-palette value survives. The `pointer: grab` Layer B (tcss:176) is correctly scoped as Kitty-OSC-22-only progressive enhancement, with the drawn colour ramp as the real, everywhere-visible affordance — the right call, since the cmux surface can't be assumed to speak the pointer protocol.

**D4 — rest thumb contrast is low but intentional.** (NIT — no change)

The rest thumb at `edge` (#3b3527 on bg #14110c) is only just perceptible as a grabbable target. This is by design (D4: the reserved gutter must not read as a border at rest), and hover + the near-miss grab pad from Bug 1's fix compensate for discoverability. Flagging only so it's on record; do not "fix" by brightening rest, which would reintroduce the border read.

---

**Verdict: TERMINAL.** No BLOCKER, no MAJOR. The neutral rest→hover→grabbed ramp is perceptible, correctly ordered (grabbed most engaged), free of the off-palette blue, and fully inside the borderless / one-green design system. The deferred accent decision is resolved in favour of neutral (D1). D4 recorded as a follow-up note only. Ship it from the design side once the code + testing rounds are clean.

damianvtran:
### UX review — round 1

**Reviewer:** ux-reviewer (Claude Opus 4)
**Scope:** `d62022e49ba866cee439ea43fad13a0524fcaeed..7565e01cccdd7b74e1ecc29b8800c46adbacc3ba`

I walked the real grab/scroll flow against the production `OperatorApp` (not stills) via a headless pilot that drives `Screen._forward_event` exactly as the terminal does — the same path production uses, since the bar lives at `x == view.right` and `Pilot._post_mouse_events` rejects that column. State observed after each gesture: `scroll_y`, `_select_state`, `_selecting`, `selections`, `mouse_captured`.

#### What I exercised, and what happened

- **(a) Near-miss grab, 1 cell left of the bar, then drag** — mousedown redirects onto the `ScrollBar` (`mouse_captured == ScrollBar`), `_select_state is None`, `_selecting is False`, `selections == {}` throughout the drag; `scroll_y` advances 0→85 and holds on release. **No highlight is ever armed** — the reported "messy highlight while dragging" bug is fixed at the source.
- **(b) Exact-thumb grab + drag** — unchanged and correct: captures the bar, scrolls, no selection. The redirect does not break the direct path.
- **(c) Genuine content selection 2 cells left** — `_select_state` is set, `_selecting is True`, and dragging extends the selection normally. The pad does **not** steal ordinary selection. ✅
- **(d) Drag past top/bottom edges** — because the gesture is a real thumb grab (not a content selection), there is no selection auto-scroll to trip. Overshoot below clamps to `scroll_y == max`, overshoot above clamps to `0`, no highlight at either extreme, clean release. The whole class of bug (selection auto-scroll on edge drag) is structurally gone, not just masked. ✅
- **(e) Rapid click vs track click** — a near-miss down/up with no move captures then releases cleanly, no selection, no stray scroll. A click in the empty track page-scrolls (`scroll_y` 0→17) as Textual's track-click does. Both feel right. ✅

Both regression guards (two-cells-left still selects; short transcript with no bar passes through) reproduce in my walk, and `scrollbar_size_vertical == 1` holds — the visual bar is untouched, only the input hit-test widened.

**Verdict: TERMINAL — no BLOCKER, no MAJOR.** The interaction is correct and the reported bug is genuinely resolved along the real path. Findings below are discoverability/forgiveness follow-ups.

---

#### Findings

**U1 — MINOR — Is `SCROLLBAR_GRAB_PAD=1` enough forgiveness for a real user?**
Geometry from the walk: the bar is the **rightmost column of the view** (`bar.region.x == view.right`, term 80/60/40 all confirm it is flush against the right edge). That means overshooting *right* is impossible — the mouse hits the terminal border (an unreachable column, a silent no-op), never content. So **every recoverable miss is a leftward undershoot onto content**, and the pad only extends left. `PAD=1` therefore turns a 1-cell target into a 2-cell one *in the only direction a miss can happen*, which is the right shape for the problem. But 2 cells is still a small Fitts's-law target for a mouse. Because forgiveness is one-directional here, there is more headroom to widen than a symmetric pad would allow: `PAD=2` would give a 3-cell effective target while still leaving content ≥3 cells in fully selectable (guard (c) already proves the boundary is clean). I'd treat `PAD=1` as *correct but conservative* — recommend trying `PAD=2` and confirming (c) still holds at two-cells-left→now-three. Not a blocker; the current value fixes the reported bug and never steals a selection.

**U2 — MINOR — On non-Kitty terminals, discovering the bar is grabbable rests entirely on a 1-cell colour shift.**
The pointer affordance (Layer B `pointer: grab`) is the honest "this is draggable" signal, but it only reaches terminals speaking Kitty OSC-22. Everywhere else, discoverability is a single cell stepping `$lo-edge`→`$lo-dim` on hover. On a lower-contrast terminal that step can be near-imperceptible on one cell, so a first-time user may not learn the bar is a handle (they may still wheel-scroll and never grab). This is bounded by the deliberate 1-cell design (D4/D21/D27), so it is a constraint, not a regression — and the ramp is strictly better than before. Follow-up, not a hold: consider whether the hover step could be one notch louder *without* reading as "active", since the grabbed state now owns `$lo-muted` and has headroom above hover.

**U3 — MINOR (unverifiable here) — Layer B `pointer: grab` could not be confirmed in this review.**
A headless SVG carries no cursor shape, so I cannot independently verify the open-hand/grabbing cursor. The PR states it was live-probed under `lop` in the cmux surface; I'm flagging that this UX round did **not** re-confirm it and cannot with the tools available to a headless walk (the browser tool drives a web surface, not a terminal cursor). Ask before merge: confirm on the record that a Kitty-protocol terminal (or cmux's proxy of it) actually flips the cursor to grab on hover and grabbing on capture — if cmux does not proxy OSC-22, that is expected and the colour ramp still carries the affordance, but the claim should be marked verified-or-known-noop rather than assumed.

**U4 — NIT — The redirect covers the whole track, but the code comment says "thumb's vertical band".**
`_redirect_gutter_grab` gates on `region.y <= event.y < region.y + region.height`, which is the full **scrollbar/track** region, not the **thumb**. My walk confirms it: a near-miss 1-cell-left over the *empty track* (thumb elsewhere) also redirects and page-scrolls (`scroll_y` 0→26), with no selection. UX consequence is benign-to-positive — forgiveness spans the whole track, so the rightmost content column is non-selectable-by-plain-click over the track's full height, but 2-cells-left still selects and that column is edge content. Flagging only because the inline comment (`# Only inside the thumb's vertical band`) and the docstring/PR narrative ("inside the thumb's vertical band") describe thumb-only behaviour that doesn't match the track-wide reality — worth a one-line comment correction so a future reader doesn't "fix" it to thumb-only and shrink the forgiving zone. (Code-comment accuracy; deferring the exact wording to the code-review round.)

---

Follow-ups to carry (none block merge): U1 (try `PAD=2`), U2 (hover contrast headroom), U3 (record the OSC-22 live-probe result), U4 (correct the "thumb band" comment to "track band").

damianvtran:
### Agent review — round 1

**Reviewer:** reviewer (lopdev team) — anthropic/claude-opus-4
**Scope:** `d62022e4..7565e01c` (origin/main..head of `dev-scrollbar-grab`)

Reviewed the diff (`app.py` `TranscriptScreen`, `local_operator.tcss`, the new test file, `pyproject.toml`), the design doc, and ran the gates + behavioural probes against the real `OperatorApp` (80×24, transcript overflowed).

**Gates:** all clean on the PR's tracked files —
- `flake8` (tracked files): clean. *(Note: `docs/scratch/{crop,shot}.py` trip flake8 but are untracked local artifacts, not in the diff and not in CI.)*
- `black==26.1.0 --check`: clean. `isort --profile black --check-only`: clean. `pyright`: 0 errors.
- `tests/unit/tui/test_transcript_scrollbar_grab.py`: 5 passed; `test_app_pilot.py` + grab: 172 passed.
- CI: the head SHA has two runs — one fully green, one with a single failure in `tests/unit/tui/test_band_panels.py::...[20-10]` (`NoMatches: No nodes match '#band'`). That test is from #245, is **untouched by this PR**, and passes 3/3 locally. Pre-existing flake, not a regression from this change.

The core fix (Bug 1) is correct and well-targeted. `_forward_event` intercepts `MouseDown` before `super()` arms a selection; the rebuilt event lands on `region.x` where `ScrollBar.allow_select=False`, so `_select_state` is never set and a real grab starts. Verified directly: near-miss at `region.x-1` → `_select_state is None`, `_selecting is False`, `mouse_captured` is a `ScrollBar`; drag then scrolls with `selections == {}`; two-cells-left still arms a selection; short transcript passes through. Screen-offset/button/modifier preservation is right, and passing `widget=event.widget` (None for raw events) is fine because `Screen._forward_event` re-resolves the target from coords. The `mouse_captured` mid-gesture guard and the `show_vertical_scrollbar` skip (which handles the hidden main transcript in subagent mode) are sound. The color ramp (Bug 2) is verified applied and monotone: rest `rgb(59,53,39)` → hover `rgb(131,124,109)` → grabbed `rgb(181,175,162)`, i.e. `$lo-edge < $lo-dim < $lo-muted`, no accent — respects D4/D21/D27 and the accent reservation at tcss ~L367.

---

#### MAJOR

**M1 — Layer B `pointer: grab` is dead at runtime; the comment claims a behaviour that never fires** (`local_operator.tcss:170-172`, `TranscriptView > ScrollBar { pointer: grab; }`)

The rule's selector matches the scrollbar (`css_path_nodes` = `OperatorApp > TranscriptScreen > TranscriptView > ScrollBar`, `_check_selectors` → True), but the scrollbar's **computed** pointer stays `default` through the entire natural runtime path — after compose, after 6 `pause()`s, after `scroll_y` changes, and after a terminal resize. `sb._css_styles.has_rule("pointer")` is `False` and `sb.styles.pointer == "default"` in every natural state; the `grab` value only appears when I force a fresh-cache re-application (`stylesheet.apply(sb)` / `update_nodes([sb])` / `refresh_css()`). So the style is never delivered to the widget through the incremental style-application path that scrollbars actually take (`Stylesheet.update_nodes` applies system scrollbars via a separate, shared-cache branch).

Consequence: the cursor never becomes an open hand *anywhere* — not just on non-Kitty terminals. That contradicts the block comment, which frames it as a Kitty-only progressive enhancement ("elsewhere it is a silent no-op"); in fact it is a no-op everywhere. This does **not** block the two-bug fix — the color ramp is the real, verified affordance and the grab fix is independent — but the PR ships a documented mechanism that provably does nothing, which will bite a maintainer who trusts the comment or wonders why Kitty shows no cursor change. Resolve by either (a) applying the pointer through a path that reaches the system scrollbar (e.g. set it where the `scrollbar-color-*` ramp is set, on `TranscriptView`, if Textual honours it there — needs verifying), or (b) dropping the dead rule + comment. A one-line probe reproduces it: computed `TranscriptView.vertical_scrollbar.styles.pointer` after `pilot.pause()` is `default`.

---

#### MINOR

**m1 — "thumb" wording overstates the band actually matched.** The redirect fires for the whole scrollbar **track** extent (`region.y <= event.y < region.y + region.height`, `region` is the full 1×track region), but the inline comment (`app.py:966`, "Only inside the thumb's vertical band"), the `_redirect_gutter_grab` docstring, the module docstring and a test name all say "thumb." Behaviour is fine (a near-miss on empty track below a top-pinned thumb redirects to a page-scroll, not a selection — verified: `mouse_captured is None`, `_select_state is None`), but the docs describe a narrower band than the code implements. Say "the bar's / scrollbar's vertical extent."

**m2 — redundant `mouse_captured` guard.** `_forward_event` checks `self.app.mouse_captured is None` and then `_redirect_gutter_grab` checks `self.app.mouse_captured is not None` again first thing. Harmless, but one of the two is dead defensiveness.

**m3 — test coverage gaps.** No test exercises the subagent two-transcript case (hidden main + shown child) the code explicitly guards, nor the mid-drag `mouse_captured is not None` early return. The 5 assertions that exist do pin the primary behaviour tightly and the top-pinned-thumb / `region.y` workaround for in-process layout leak is legitimate (documented, works around `run_test` state bleed, not masking a production bug).

#### NIT

**n1 — private-attribute access.** `_redirect_gutter_grab` reads `event._delta_x` / `event._delta_y` (private) where public `event.delta_x` / `event.delta_y` exist and return the same values (both 0 for a `MouseDown`). Cosmetic.

---

**Verdict:** not clean — **1 MAJOR (M1)**. The core two-bug fix is correct, well-commented, and verified working; the MAJOR is the dead Layer-B pointer rule, which is severable from the primary value. Recommend the implementer either make `pointer` actually reach the scrollbar or remove the rule + its comment, then a round 2 confirming the resolution against the current head. The MINORs/NIT are follow-ups, not round-drivers.

damianvtran:

**Implementer:** task (lopdev team) — anthropic/claude-opus-4
**Head:** `8ef093c5` (was `7565e01c`)
**Scope:** `7565e01c..8ef093c5` on `dev-scrollbar-grab`

Addresses the code-review MAJOR + MINORs/NIT and the UX follow-ups U1/U4. Design D1 (neutral vs accent) is settled — **no colour change**; the `scrollbar-color-*` lines are untouched (confirmed by diff).

---

**M1 (MAJOR) — dead `pointer: grab` CSS rule → FIXED (`8ef093c5`).**
Confirmed the reviewer's finding: the `TranscriptView > ScrollBar { pointer: grab }` rule matched the selector but never reached the widget — system scrollbars take Textual's shared style-cache branch (`Stylesheet.update_nodes`), and the computed `pointer` stayed `default` through compose, 6× pause, `scroll_y` change, and resize. **Chose fix (a):** dropped the dead rule and set the pointer as an **inline style** in `TranscriptView.on_mount` (`self.vertical_scrollbar.styles.pointer = "grab"`), which bypasses the cache. Proven to stick: a pilot reads `vertical_scrollbar.styles.pointer == "grab"` at mount, after fill+4 pauses, after a scroll, and after a resize (same instance, value intact). New test `test_scrollbar_pointer_is_grab_and_sticks` pins all four checkpoints. The tcss now carries a comment stating *why* the rule form is dead and where the pointer is actually set; `on_mount` carries the full rationale.

**Live-probe boundary (U3), stated honestly:** the test proves the *style value* is set — all a headless pilot can prove. Whether the terminal renders the OSC-22 open-hand cursor is a live-surface question that still needs a probe under `lop` in the real cmux surface (hover the bar, observe the cursor). If cmux does not proxy Kitty OSC 22 the cursor won't change and that is expected — the colour ramp is the everywhere-visible affordance. Flagging, not faking.

**U1 (MINOR) — pad widened to 2 → APPLIED (`8ef093c5`).**
`SCROLLBAR_GRAB_PAD = 1 → 2` (a 3-cell effective target: the bar + two pad columns). The bar is flush-right against the reserved gutter so a miss can only be leftward; a content click 3+ cells in is still untouched. Constant doc updated with the reasoning. Tests updated: the old "two cells left still selects" is now **`test_two_cells_left_now_redirects_with_wider_pad`** (asserts a grab + no selection — positive proof of the wider pad), and a new **`test_three_cells_left_still_selects_content`** guards the boundary, computed as `region.x - (SCROLLBAR_GRAB_PAD + 1)` so it tracks the constant rather than a hardcoded offset.

**U4 / m1 (MINOR) — "thumb band" wording → CORRECTED (`8ef093c5`).**
The redirect keys on the whole scrollbar **track** (`region.y ≤ event.y < region.y + region.height`), not the thumb. Corrected the constant doc, the `_redirect_gutter_grab` docstring, the inline comment, and the test module docstring to say "the scrollbar's vertical extent (the whole track, not just the thumb)."

**m2 (MINOR) — redundant guard → REMOVED (`8ef093c5`).**
Dropped the `mouse_captured is not None` early-return inside `_redirect_gutter_grab`; `_forward_event` already gates on `mouse_captured is None` before calling it. The docstring now notes the caller owns that check. The mid-drag guard is now pinned by **`test_redirect_early_returns_while_a_drag_is_in_progress`** (captures the scrollbar, asserts `_redirect_gutter_grab` is not even called for a subsequent near-miss).

**m3 (MINOR) — missing guard tests → ADDED (`8ef093c5`).**
- `test_redirect_skips_hidden_subagent_transcript` — hides the view (as subagent mode does to the main transcript) and asserts a near-miss does not grab. While fixing this I found the code comment was slightly wrong: a hidden view's `show_vertical_scrollbar` can still read **True**; what actually excludes it is the compositor `find_widget` raising `NoWidget` (a display:none view isn't in the compositor). Corrected the comment to name the real gate.
- `test_redirect_early_returns_while_a_drag_is_in_progress` — the capture-in-progress early return (see m2).

**n1 (NIT) — private access → FIXED (`8ef093c5`).**
`event._delta_x` / `event._delta_y` → public `event.delta_x` / `event.delta_y` (both 0 for a `MouseDown`), with a comment noting why they're preserved.

---

**Gates (all clean on `8ef093c5`):**
- `flake8 local_operator tests` → exit 0
- `black==26.1.0 --check local_operator tests` → 454 files unchanged
- `isort --check-only --profile black local_operator tests` → exit 0
- `pyright .` → 0 errors, 0 warnings
- `env -u NO_COLOR TERM=xterm-256color pytest tests/unit/tui` → **2366 passed, 4 skipped**. The scrollbar-grab file is now **9 tests, all passing** (serial `-n0` and xdist both green).

No design/UX re-review needed for behaviour, but the pad widened from 1→2 (interaction surface) and the pointer mechanism changed — flagging for the round-2 driver in case a fresh UX glance on the wider target is wanted. No colour/frame change.

damianvtran:
### UX review — round 2

**Reviewer:** ux-reviewer (lopdev team) — anthropic/claude-opus-4
**Scope:** `d62022e49ba866cee439ea43fad13a0524fcaeed..8ef093c50086a573653716f51d9f70d8813907c9` (origin/main..head of `dev-scrollbar-grab`)

Round 1 was TERMINAL. The only interaction-relevant changes since are (i) `SCROLLBAR_GRAB_PAD` widened `1 → 2` (a 3-cell effective target: the bar plus two pad columns), and (ii) the hover pointer moved from a dead `pointer: grab` tcss rule to an inline `self.vertical_scrollbar.styles.pointer = "grab"` in `TranscriptView.on_mount`. I re-walked the real grab/scroll flow against the production `OperatorApp` via a fresh headless pilot driving `Screen._forward_event` (the production path — the bar sits at `x == screen.width`, which `Pilot._post_mouse_events` rejects), state observed after each gesture: `mouse_captured`, `_select_state`, `_selecting`, `selections`, `scroll_y`.

#### What I exercised at head 8ef093c5, and what happened

- **(a) Near-miss redirects to a clean scroll, no selection** — at both **1 cell** and **2 cells** left of the bar (`x=77`, `x=76`; bar at `x=78`), the mousedown captures the `ScrollBar`, with `_select_state == None`, `_selecting == False`, `selections == {}`. The wider pad forgives the full 2-column miss and no highlight is ever armed. ✅
- **(c) Exact-thumb and 2-cells-left both grab** — direct click on the bar column (`x=78`) captures the `ScrollBar` cleanly; the 2-cell near-miss (the new pad edge) does too (above). The wider hit area did not break the direct path. ✅
- **(b) Content click 3+ cells left still selects** — at `SCROLLBAR_GRAB_PAD + 1 == 3` cells left (`x=75`), `mouse_captured` is `None`, `_select_state` is set and `_selecting is True` — ordinary text selection, untouched. The wider pad did **not** start stealing content. ✅
- **End-to-end forgiveness** — a 2-cell near-miss grab followed by a drag scrolls the transcript (`scroll_y 0 → 56.5`) with `selections == {}` throughout. The reported "scrolls but leaves a messy highlight" symptom stays structurally gone at the wider pad. ✅

`scrollbar_size_vertical == 1` holds in every scenario — the visual bar is untouched; only the input hit-test widened. The scrollbar-grab test file is **9 passed** on this head.

#### Prior findings

- **U1 (MINOR — pad forgiveness) → RESOLVED.** `SCROLLBAR_GRAB_PAD = 2` at `app.py:956`. Both pad columns redirect (confirmed above), the constant doc explains the reasoning and the flush-right geometry that makes a rightward miss impossible, and the `region.x - self.SCROLLBAR_GRAB_PAD <= event.x < region.x` gate tracks the constant. 3 cells in still selects, so the widening bought forgiveness without cost.
- **U4 (NIT — "thumb band" wording) → RESOLVED.** The constant doc, `_redirect_gutter_grab` docstring, the inline comment (`# Anywhere in the scrollbar's vertical extent — the whole TRACK, not only the thumb`), and the test module docstring now all say "the whole track, not just the thumb," matching the `region.y <= event.y < region.y + region.height` gate my round-1 walk exercised.

#### Boundary restated honestly (U3 — Layer B pointer, unverifiable headlessly)

The inline `pointer: grab` change is the fix for round-1's M1 (the tcss rule never reached the system scrollbar's shared style cache; the inline style bypasses it). I confirmed the **style value is set** — `TranscriptView.on_mount` runs and assigns it — but a headless pilot carries no terminal cursor, so I **cannot and do not claim the OSC-22 open-hand cursor actually renders**. Whether a Kitty-pointer-shape terminal (or cmux's proxy of it) flips the cursor on hover is a live-surface question that still needs a probe under `lop` in the real surface. If the terminal does not speak the pointer-shape protocol this is a silent no-op — expected — and the everywhere-visible affordance remains the drawn `scrollbar-color-*` ramp (design round 1 confirmed that ramp). Flagging the boundary, not faking the pixel.

---

**Verdict: TERMINAL — UX axis is CLEAN (no BLOCKER, no MAJOR).** The wider 3-cell target is more forgiving without stealing content, both grab paths work, and the flow copy/wording now matches the whole-track behaviour. Carry-forward follow-up only: U3 (record the OSC-22 live-probe result under a real terminal at merge/verify time) — does not block.

damianvtran:
### Agent review — round 2

**Reviewer:** reviewer (lopdev team) — anthropic/claude-opus-4
**Scope:** `d62022e49ba866cee439ea43fad13a0524fcaeed..8ef093c50086a573653716f51d9f70d8813907c9` (head 8ef093c5)

Re-reviewed the diff at the current head and independently re-verified every round-1 finding against it (I read the new tests critically and drove my own pilots rather than trusting them). All gates re-run green on 8ef093c5.

---

#### Round-1 findings — resolution against current head

**M1 (MAJOR) — dead `pointer: grab` CSS rule → RESOLVED.**
The dead `TranscriptView > ScrollBar { pointer: grab }` rule is fully gone from `local_operator.tcss` (`grep pointer` shows no live rule for the transcript scrollbar; only an explanatory comment block remains at ~L146/L170 documenting *why* the CSS form is dead and where the pointer is now set). The pointer is delivered as an inline style in `TranscriptView.on_mount` (`transcript.py`), which bypasses the shared style-cache branch.

I did not just trust `test_scrollbar_pointer_is_grab_and_sticks` (which asserts the `styles.pointer` bag only). I drove two pilots on the real `OperatorApp`:
- The value sticks as `grab` at mount, after fill+4 pauses, after `scroll_y=30`, and across a resize (same scrollbar instance) — `css_styles` carries *no* pointer rule, confirming the inline style is the source.
- **The production emit path resolves it.** `Screen.update_pointer_shape()` walks `mouse_over.ancestors_with_self` reading `node.styles.pointer`; with `mouse_over` set to the scrollbar the screen's `_pointer_shape` becomes `grab` (→ `App._set_pointer_shape` emits OSC 22), and reverts to `default` over content. So the cursor genuinely becomes an open hand on hover in a pointer-shape-speaking terminal — the mechanism the comment claims now actually fires. No dead rule, no misleading comment.

**m1/U4 (MINOR) — "thumb" wording → RESOLVED.** The constant doc, `_redirect_gutter_grab` docstring, inline comment (`app.py`), module docstring, and test text now consistently say "the scrollbar's vertical extent (the whole track, not just the thumb)." Matches the code: `region.y <= event.y < region.y + region.height`.

**m2 (MINOR) — redundant `mouse_captured` guard → RESOLVED.** The re-check inside `_redirect_gutter_grab` is removed; `_forward_event` owns the `mouse_captured is None` gate and the docstring notes the caller established that. Pinned by `test_redirect_early_returns_while_a_drag_is_in_progress`, which I confirmed captures the bar and asserts `_redirect_gutter_grab` is not called (`calls == 0`) for a subsequent near-miss.

**m3 (MINOR) — guard test gaps → RESOLVED.** Both gaps are now covered, and the hidden-view test genuinely exercises the intended path: I confirmed in a pilot that a `display:none` view still reports `show_vertical_scrollbar == True`, so the first `continue` does NOT fire — the `NoWidget` from `_compositor.find_widget` is the gate that excludes it, exactly as the comment now states. The mid-drag early return is covered as noted in m2.

**n1 (NIT) — private delta access → RESOLVED.** `event._delta_x/_delta_y` → public `event.delta_x/delta_y`, with a comment noting both are 0 for a `MouseDown` and are preserved for event fidelity.

**U1 (MINOR) — pad 1→2 → CORRECT.** `SCROLLBAR_GRAB_PAD = 2`. Boundary is right: the redirect fires for `region.x - PAD <= event.x < region.x`, i.e. columns `region.x-2` and `region.x-1`. `test_two_cells_left_now_redirects_with_wider_pad` (positive proof: grab, no selection) and `test_three_cells_left_still_selects_content` (computed as `region.x - (PAD+1)`, asserts selection armed + no ScrollBar capture) pin both sides and track the constant rather than a hardcoded offset. Bar is flush-right so a miss can only be leftward — no content-side exposure.

**Colours — UNCHANGED (design D1 preserved).** `scrollbar-color-hover: $lo-dim` / `scrollbar-color-active: $lo-muted` are identical to round 1; no accent introduced.

---

#### Gates on 8ef093c5 (all green)
- `flake8 local_operator tests` → exit 0
- `black==26.1.0 --check local_operator tests` → 454 files unchanged
- `isort --check-only --profile black local_operator tests` → exit 0
- `pyright .` → 0 errors, 0 warnings, 0 informations
- `env -u NO_COLOR TERM=xterm-256color pytest tests/unit/tui` → **2366 passed, 4 skipped**. The scrollbar-grab file is 9 passed. No regression vs round 1.

---

**Verdict: CLEAN — no BLOCKER, no MAJOR. The code axis is now clean and this round is TERMINAL.** M1 is genuinely fixed (verified through the real pointer-emit path, not just the style bag) and every round-1 minor/nit is resolved. The round-1 D4 rest-contrast NIT remains a recorded follow-up, not a reason for another round.

damianvtran:
### Design review — round 2

Scope: 7565e01c..8ef093c5 (current head 8ef093c5)

Round-2 confirmation that the visual surface is unchanged after the round-1
remediation. Verified, not assumed:

- `git diff 7565e01c 8ef093c5 -- local_operator/tui/local_operator.tcss` shows
  the `scrollbar-color-*` declarations are **byte-identical**. The only changes
  inside the `TranscriptView` scrollbar block are comment prose and the removal
  of the dead `TranscriptView > ScrollBar { pointer: grab; }` rule (round-1 M1:
  it was never delivered to the system scrollbar; the open-hand cursor is now an
  inline style set in `on_mount`). That is a pointer-mechanism change, not a
  colour change.
- Neutral ramp confirmed on the current head 8ef093c5: rest `$lo-edge` (l.152)
  → hover `$lo-dim` (l.166) → grabbed/active `$lo-muted` (l.167). Monotone
  rest<hover<grabbed on the 1-cell thumb, as settled in round 1.
- Because the colours are untouched, the round-1 `docs/scratch/frames/after/`
  captures still represent the current visual state; re-capture is unnecessary.

Findings carried forward from round 1:
- **D1** — settled NEUTRAL. `$lo-accent` (the green reserved for "a turn is
  live") is correctly rejected for the grabbed state; a green thumb on a passive
  scroll gesture would fire the live-activity signal falsely and paint the
  reserved gutter's highest-chroma colour. Remediation comment now records this
  rejection in the tcss. Decision stands.
- **D4** — rest-contrast NIT (recorded follow-up, not a blocker).

Verdict: **CLEAN and fresh** against the current head 8ef093c5. No BLOCKER, no
MAJOR. This round is **TERMINAL** on the design axis.

damianvtran:
### Agent review — round 3

Scope: 8ef093c5..4b4ab209

This round exists solely to keep the review fresh against the current head after origin/main advanced (PR #161 merged, taking version 0.24.4) and the branch was rebumped 0.24.4 → 0.24.5.

**Verification**
- `git diff 8ef093c5 4b4ab209` — the entire delta is a single line in `pyproject.toml`: `version = "0.24.4"` → `version = "0.24.5"`. No code, test, or config changes.
- `git show origin/main:pyproject.toml` reports `version = "0.24.4"`; 0.24.5 strictly advances it, so the release version is monotonic.
- `git merge-tree` against origin/main shows 0 conflict markers — the branch merges cleanly.

**Verdict: CLEAN (terminal).** All round-2 code findings remain resolved; there is no new code change to review, only a monotonic version bump verified above. No blocker or major findings. The code axis is CLEAN and fresh against head 4b4ab209.

PR #271, opened 2026-08-23
fix(tui): carry pasted images through /team and /agent to the model

damianvtran:
## Why

Attaching an image and then adding team/persona context with a slash command **silently dropped the image**. The reported flow: paste a screenshot into the composer, then route it with `/team <name> <request about the image>` (or `/agent <name> <message>`), and the model receives the bare `[Image #N]` marker as *text with no pixels behind it*. The user's workaround was to re-attach the image after the command — the symptom in the bug report.

Root cause is a single dropped argument on the submit path:

- `on_editor_submitted` routes any slash-shaped line to `_run_slash_command(text)` and **discards `message.images` / `message.attachments`** (the non-slash branch right below it passes them to `_submit_prompt`).
- The two commands whose argument actually reaches the model — `/team <name> <request>` and `/agent <name> <message>`, both `consumes_prompt=True` — then call `_submit_prompt(request)` with **no images**.

Every other slash command takes no model-bound prompt (`/goal` stores text, `/loop` iterates, `/btw` opens the text-only aside), so only these two are affected. The composer's attachment map was fully intact at submit — it just never got handed to the command handlers.

Version bump is **PATCH** (0.25.0 → 0.25.1): one behavioural bug fix, no new capability.

## What changes

- **`tui/app.py`, `on_editor_submitted`** — the slash-dispatch call now passes `message.attachments` through: `self._run_slash_command(text, message.attachments)`. Handed as the index→attachment **map**, not the pre-resolved `images` list, because the handler submits only the *request tail* and must resolve the markers **that** text cites, not the whole slash line's.
- **`_run_slash_command`** — gains an optional `attachments` parameter (defaults `None`; the mid-draft `on_inline_command_requested` path still leaves the draft in the composer, so it passes nothing), forwarded to `/team` and `/agent`.
- **`_cmd_team` / `_cmd_agent`** — before `_submit_prompt`, resolve `images = resolve_markers(request, attachments or {})` and pass both `images` and the `attachments` map. Using `resolve_markers` against the **request** keeps the composer's text-is-authority rule: a marker the request no longer cites drops its image, order follows the text, and the map riding through `_submit_prompt` lets the pixels survive a steer/compaction hold exactly as a plain prompt does. A bare attach (`/team ops` with no request) cites nothing and sends no turn.

### Automated — new end-to-end regression tests (`tests/unit/tui/test_slash_command_images.py`)

These drive the **real `OperatorApp`** through the actual submit path (paste a real PNG → type the command around the marker → press Enter), and read the pixels back off the session — proof the bytes travelled, not merely a marker.

```
    tests/unit/tui/test_slash_command_images.py -q
...                                                                      [100%]
3 passed, 28 warnings in 2.43s
```

- `test_team_request_carries_the_pasted_image` — `/team ops look at [Image #1, 1568x410] and fix it` → `session.prompts == ["look at [Image #1, 1568x410] and fix it"]` **and** `session.prompt_images[0]` decodes to a real image.
- `test_agent_message_carries_the_pasted_image` — same for `/agent auditor …`.
- `test_team_without_a_request_sends_nothing` — a bare attach cites nothing: `session.prompts == []`, `session.prompt_images == []` (guards against smuggling a leftover attachment into a turn that never runs).

### The test catches the bug (mutation check)

Reverting `_cmd_team` to the old `self._submit_prompt(request)` makes the new test fail on exactly the dropped image:

```
FAILED test_team_request_carries_the_pasted_image
  - ValueError: not enough values to unpack (expected 1, got 0)   # no image reached the session
```

Restoring the fix → green.

### No regressions

```
$ env -u NO_COLOR TERM=xterm-256color .venv/bin/python -m pytest tests/unit/tui -q
2400 passed, 4 skipped, 28 warnings in 78.40s
```

### Gates

```
flake8 .                                  # clean
black==26.1.0 --check local_operator tests  # All done ✨ 455 files unchanged
isort --check-only --profile black         # clean
pyright                                    # 0 errors
```

## Notes

No visual or interaction-flow change: the command surface, picker, and echo policy are all identical — only the payload routing behind them changed. So this is a code-review-only change (no design/UX round).

### DIFF
--- a/local_operator/tui/app.py
+++ b/local_operator/tui/app.py
@@ -3114,6 +3114,29 @@ class OperatorApp(App[None]):
         rows.append(Text("Send: /team <name> <message>", style=body))
         return RichBlock(Group(*rows))
 
+    def _submit_command_prompt(
+        self, request: str, attachments: Mapping[int, Attachment] | None
+    ) -> None:
+        """Send a prompt-carrying slash command's argument as a real user turn.
+
+        The shared tail of ``/team <name> <request>`` and ``/agent <name>
+        <message>``: both attach something to the session and then hand the
+        argument to the model. The images the argument cites are resolved from
+        the text itself (``resolve_markers``) — the same authority the composer
+        uses, so a marker the request no longer cites drops its image and order
+        follows the text — and both the resolved list and the raw index→image
+        map go to ``_submit_prompt`` so a held draft (steer/compaction) can
+        rebuild its pixels exactly as a plain prompt would.
+
+        Extracted so the resolve-then-submit pair lives in ONE place: it was
+        copied into three call sites (``/team``, ``/agent`` with and without a
+        layered persona), and three copies of "which text resolves the markers"
+        is exactly the kind of drift the composer's single ``resolve_markers``
+        walk exists to prevent.
+        """
+        images = resolve_markers(request, attachments or {})
+        self._submit_prompt(request, images, attachments)
+
     def _cmd_team(
         self,
         arg: str,
@@ -3217,16 +3240,12 @@ class OperatorApp(App[None]):
         # second row restating `/team name …` would be the duplication
         # the echo policy exists to prevent.
         #
-        # ``resolve_markers`` against the REQUEST, not the whole slash line: the
-        # marker sits in the request tail the user typed around, and resolving
-        # from the text is the same authority the composer uses — a marker the
-        # request no longer cites (deleted before sending) drops its image, and
-        # order follows the text. Passing the map through both arguments lets
-        # ``_submit_prompt`` render the image blocks and hold the pixels across
-        # a steer/compaction exactly as a plain prompt does.
+        # ``_submit_command_prompt`` resolves the request's own image markers
+        # and sends them — the marker sits in the request tail the user typed
+        # around, so a pasted screenshot reaches the manager as pixels, not a
+        # dead ``[Image #N]`` marker.
         notice(f"sending to {team.name}. {team.manager} is coordinating.")
-        images = resolve_markers(request, attachments or {})
-        self._submit_prompt(request, images, attachments)
+        self._submit_command_prompt(request, attachments)
 
     def _cmd_team_chart(self, name: str, registry: Any, notice: NoticeFn) -> None:
         """``/team chart [name]`` — open the org-chart mode for a team.
@@ -3516,8 +3535,7 @@ class OperatorApp(App[None]):
                 # message's own image markers are resolved and sent, same as the
                 # instruction-carrying path below.
                 notice(f"agent {resolved} has no instructions; sending your message as-is.")
-                images = resolve_markers(request, attachments or {})
-                self._submit_prompt(request, images, attachments)
+                self._submit_command_prompt(request, attachments)
             else:
                 notice(
                     f"agent {resolved} resolved but carries no instructions, "
@@ -3542,8 +3560,7 @@ class OperatorApp(App[None]):
         # markers are resolved from the text, the same authority the composer
         # uses, so a cited screenshot reaches the persona as pixels.
         notice(f"agent {resolved} now governs this session's replies.")
-        images = resolve_markers(request, attachments or {})
-        self._submit_prompt(request, images, attachments)
+        self._submit_command_prompt(request, attachments)
 
     # -- MCP status ---------------------------------------------------------
     def _wire_mcp_status(self, session: SessionProtocol) -> None:
--- a/local_operator/tui/app.py
+++ b/local_operator/tui/app.py
@@ -3114,13 +3114,23 @@ class OperatorApp(App[None]):
         rows.append(Text("Send: /team <name> <message>", style=body))
         return RichBlock(Group(*rows))
 
-    def _cmd_team(self, arg: str, notice: NoticeFn) -> None:
+    def _cmd_team(
+        self,
+        arg: str,
+        notice: NoticeFn,
+        attachments: Mapping[int, Attachment] | None = None,
+    ) -> None:
         """``/team`` — list; ``/team <name> <request>`` — talk to the manager.
 
         Bare ``/team`` is a listing, so it does not echo. A named team with a
         request attaches that team to this session (stamping the roster and
         briefs onto the manager) and sends the request as a real user turn —
         the one case besides ``/goal`` whose argument reaches the model.
+
+        ``attachments`` is the composer's image map at submit time. Only the
+        markers the REQUEST still cites are resolved and sent, so a screenshot
+        pasted before the user typed ``/team ops <request about it>`` reaches
+        the manager as pixels rather than a dead ``[Image #N]`` marker.
         """
         session = self._session
         if session is None:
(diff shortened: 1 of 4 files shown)


damianvtran:
### Agent review — round 1

Reviewer: reviewer subagent (lopdev team), model claude-opus-4
Scope: origin/main..ca903a28295b93528de2b7d75f67caaf9b71576b

Reviewed the diff, read the surrounding code (`resolve_markers`/`cite` in
`widgets/editor.py`, `_submit_prompt`, `on_editor_submitted`,
`on_inline_command_requested`, `_cmd_goal`/`_cmd_loop`/`_cmd_btw`), ran the
gates, and reproduced the mutation-catch.

**Verification performed**
- flake8, black==26.1.0 --check, isort --profile black, pyright on `app.py`:
  all clean.
- `test_app_pilot.py` (the `FakeSession.prompt_images` change): 167 passed.
- New `test_slash_command_images.py`: 3 passed against the fixed tree.
- **Mutation-catch is real.** Reverting only `app.py` to `origin/main` (keeping
  the new tests) fails 2 of 3 with the exact bug —
  `ValueError: not enough values to unpack` at `(sent,) = session.prompt_images[0]`
  because no image travelled. The tests assert on decoded PNG pixels off
  `prompt_images`, not on the marker text, so they prove bytes moved.

**Correctness (confirmed, not inferred)**
- `resolve_markers(request, attachments or {})` resolves against the request
  tail, and `cite()` matches the `[Image #N…]` marker inside that tail wherever
  it sits — so `/team ops look at [Image #1] and fix it` resolves correctly
  after `ops` is stripped into the name. `_submit_prompt` renders image blocks
  with `marker_text=text` where `text==request`, so citation order stays
  consistent between the pixels sent and the blocks painted.
- Bare `/team ops` (no request tail): `resolve_markers("")` → `[]`. Test
  `test_team_without_a_request_sends_nothing` confirms `prompts == []` and
  `prompt_images == []` — a leftover composer attachment is not smuggled into a
  turn that never runs.
- Inline mid-draft path (`on_inline_command_requested`) correctly calls
  `_run_slash_command(command_text)` with no attachments: it does not submit a
  prompt (draft stays in the composer), so it needs none. Correct.
- Steer / compaction-hold paths already thread `attachments` through
  (`steer_message = Message.user(text, images)`; `_images_held_for_compaction =
  dict(attachments or {})`). Passing the map — not just the resolved list — is
  what lets a held draft rebuild its images, matching how a plain prompt
  behaves.
- Leaving `/goal`, `/loop`, `/btw` unchanged is correct: `/goal` stores into
  the prompt tail (no `_submit_prompt`, no images), and neither `/loop` nor
  `/btw` submits an image-bearing model prompt. No other `_run_slash_command`
  callers exist (only the two grepped), so no other spelling of the bug leaks.

No blocker and no major findings.

**minor — F1** No test exercises the multi-image or the deleted-marker-in-a-
present-request case. The `resolve-from-text` claim ("a marker the request no
longer cites drops its image, and order follows the text") is asserted in the
comments and covered only indirectly by the no-request test. A two-image test
plus a "cite only #2, expect one image = the #2 pixels" test would lock the
ordering/subset contract the docstring promises.
(`tests/unit/tui/test_slash_command_images.py`)

**minor — F2** An image marker in the NAME position (`/team [Image #1] …`)
resolves nothing because it isn't in the request tail — which is the right
outcome, but it's silent: the team lookup simply fails on a bracket-shaped
name. Not a defect for this PR; worth a one-line note if you want the behaviour
pinned by a test.

**nit — F3** `_cmd_team`/`_cmd_agent` now duplicate the exact three lines
`images = resolve_markers(request, attachments or {})` /
`self._submit_prompt(request, images, attachments)` in three spots
(`app.py` ~3227, ~3520, ~3546). Fine as-is; a tiny private helper would remove
the copy if this pattern grows.

**Verdict: clean.** No blocker, no major. This round is TERMINAL — F1/F2/F3 are
follow-ups, not grounds for another round. The fix is correct, the gates are
green, and the regression tests genuinely catch the bug.

damianvtran:

Reviewer round 1 was clean (no blocker/major). Worked the three non-blocking follow-ups; all fixed in `c2c3d317`.

- **F3 (nit) — fixed** (`c2c3d317`). Extracted the duplicated `resolve_markers(request, …)` + `_submit_prompt(request, images, attachments)` pair into `_submit_command_prompt`, now the single call for `/team` and both `/agent` submit branches. One place decides which text resolves the markers, matching the composer's single-`resolve_markers`-walk reasoning.
- **F1 (minor) — fixed** (`c2c3d317`). Added `test_team_request_carries_multiple_images_in_text_order`: two images pasted (`#1` 40×20, `#2` 80×60), request cites only `#2`, asserts exactly one image is sent and its decoded size is `(80, 60)` — locking the keys-not-positions subset/order contract the docstring promises.
- **F2 (minor) — fixed** (`c2c3d317`). Added `test_team_marker_in_the_name_position_is_not_an_image`: `/team [Image #1] …` treats the bracket as a team name, matches nothing, and sends no team/turn/image — pinning the correct-but-previously-silent behaviour so a refactor can't make the name slot image-bearing.

Verification after remediation:
```
$ env -u NO_COLOR TERM=xterm-256color .venv/bin/python -m pytest tests/unit/tui/test_slash_command_images.py -q
5 passed

$ env -u NO_COLOR TERM=xterm-256color .venv/bin/python -m pytest tests/unit/tui -q
2402 passed, 4 skipped

$ flake8 . / black==26.1.0 --check / isort --profile black / pyright   # all clean
```

No blocker/major findings outstanding. Head is now `c2c3d317`.

damianvtran:
### Agent review — round 2

Reviewer: reviewer subagent (Claude Opus 4.x) on team lopdev
Scope: ca903a28..c2c3d317 (remediation delta since round 1)

Reviewed the refactor that extracts `_submit_command_prompt(request, attachments)` and reroutes the three prompt-carrying call sites through it, plus the two added tests.

**Helper equivalence — confirmed.** `_submit_command_prompt` is byte-for-byte the inlined pair it replaces:
```python
images = resolve_markers(request, attachments or {})
self._submit_prompt(request, images, attachments)
```
All three call sites (`/team` at app.py:3243, `/agent` no-instructions branch at :3538, `/agent` normal branch at :3562) previously held that exact two-line body and now call the helper with the same `request`/`attachments`. No site was missed and no semantics changed — `resolve_markers` still runs against the REQUEST tail (not the whole slash line), and both the resolved list and the raw index→image map still reach `_submit_prompt`. The core fix in `on_editor_submitted`/`_run_slash_command` is untouched by this delta, so the round-1 verdict holds.

**Tests — confirmed passing and asserting what they claim.**
- `test_team_request_carries_multiple_images_in_text_order`: pastes 40x20 then 80x60, cites only `#2`, asserts the single sent image decodes to (80, 60). Correctly locks the marker-is-a-key subset contract.
- `test_team_marker_in_the_name_position_is_not_an_image`: puts the marker in the name slot, asserts no team attached, no prompt, no image sent. Correctly pins that only the request tail carries images.
Both pass; full file 5 passed; tui spot-check (test_slash_command_images + test_app_pilot) 172 passed.

**Gates — all clean on c2c3d317:** flake8 0, black==26.1.0 --check unchanged, isort --profile black clean, pyright 0 errors.

Findings:

- **NIT (N1)** — `test_team_request_carries_multiple_images_in_text_order` docstring claims "citing both in the reverse order must send them in that order," but the test body only exercises the subset case (cite only `#2`). Either drop that sentence or add the reverse-order assertion it promises. Non-blocking; does not affect correctness.

No blocker or major findings. The delta is a faithful extract-method refactor with correct supporting tests.

Verdict: **clean** — this round is TERMINAL. N1 is a follow-up, not grounds for another round.

damianvtran:

Worked the one NIT from round 2 (test docstring over-promised a reverse-order case the body doesn't exercise).

- **Round-2 NIT — fixed** (`f1e3ff6a`). Trimmed `test_team_request_carries_multiple_images_in_text_order`'s docstring to describe only the subset case it asserts, and point at `resolve_markers`' own ordering tests in `test_paste_images` for the multi-marker ordering contract. Delta is docstring-only:

```
$ git diff c2c3d317..f1e3ff6a --stat
 tests/unit/tui/test_slash_command_images.py | 10 ++++++----
 1 file changed, 6 insertions(+), 4 deletions(-)
```

No code/logic change; gates still green and `test_slash_command_images.py` 5/5. Head is now `f1e3ff6a`.

damianvtran:
### Agent review — round 2 (freshness re-confirm)

Reviewer: reviewer subagent (Claude Opus 4.x) on team lopdev
Scope: c2c3d317..f1e3ff6a

Confirmed the delta since the round-2 clean verdict is **docstring-only, no code change**: the sole hunk rewords the `test_team_request_carries_multiple_images_in_text_order` docstring to drop the reverse-order sentence flagged in N1 and point ordering coverage at `resolve_markers`'s own `test_paste_images`. No production code, no test body, no assertions touched. Test file re-run on head f1e3ff6a: **5 passed**.

N1 resolved. The round-2 **clean** verdict still holds against head f1e3ff6a — this round remains terminal, no new round required.

PR #272, opened 2026-08-23
fix(tui): stop messy selection on the usage panel scrollbar

damianvtran:
## Why

Follow-up to #270. That PR gave the **transcript** scrollbar a forgiving grab so a near-miss stopped arming a messy text selection — but it did so via `TranscriptScreen._redirect_gutter_grab`, which only queries `TranscriptView`. Overlay panels got none of it, and the user reported the exact same messy-highlight-while-dragging bug on the **`/usage` panel** scrollbar (the transcript one now works).

The usage panel is architecturally different from the transcript: `UsagePanel` is a monolithic **selectable `Static`** that paints its OWN scrollbar (a hand-painted glyph in the reserved rightmost body column) and drives its own drag via `on_mouse_down`/`on_mouse_move`/`on_mouse_up` + `_scrollbar_hit`. It does **not** use a Textual `ScrollBar`, so the `allow_select=False` lever the transcript fix relied on does not exist here.

Two distinct problems, both reproduced against the real `OperatorApp`:
1. **Near-miss (1–2 cells left of the bar):** `_scrollbar_hit` required an **exact** column match (`offset.x != _body_content_width()` → `None`). A near-miss returned `None`, the handler did not `event.stop()`, and — because the panel is a selectable `Static` — Textual's base `Screen._forward_event` armed a text selection on the MouseDown. Dragging then extended it. (`dragging=False, captured=None, select_state=True`.)
2. **Even an exact bar hit leaked a selection:** base `_forward_event` arms `_select_state=True` on the MouseDown **before** the panel's `on_mouse_down` runs `event.stop()`+`capture_mouse()`. The panel scrolled correctly, but the base mouse-move handler extended the already-armed selection in parallel with the drag. (`dragging=True, captured=UsagePanel, select_state=True`.)

Version bump is **PATCH** (0.25.0 → 0.25.1): a bug fix to an existing control.

## What changes

Two small changes in `local_operator/tui/widgets/usage_panel.py`, both inside the panel that already owns its scrollbar. The painted bar, `_paint_scrollbar`, `_body_content_width`, and the transcript code are untouched — only the input hit-test widens and the grab clears a stray selection.

### Forgiving hit-test
New module constant `SCROLLBAR_GRAB_PAD = 2` (mirrors `TranscriptScreen.SCROLLBAR_GRAB_PAD`). `_scrollbar_hit` now accepts a press within 2 cells to the **left** of the bar column, not only exactly on it:
```python
bar = self._body_content_width()
if not (bar - SCROLLBAR_GRAB_PAD <= offset.x <= bar):
    return None
```
Safe because the report's numbers are right-aligned to `_body_content_width` and the rightmost data columns sit well left of the reserved gutter, so a small left pad cannot steal a content click.

### A grab must not leave a selection armed
In `on_mouse_down`, when `_scrollbar_hit` returns a hit, clear the selection the base handler already armed (guarded for the CSS-less test host / pre-mount, which have no screen):
```python
if self.is_mounted:
    try:
        self.screen.clear_selection()
    except NoScreen:
        pass
```
This fixes **both** the near-miss and the exact-hit leak.

## Impact
No visual change (the painted bar is byte-identical, thumb still lights `accent` on grab). No transcript change. The only behavioral deltas: a near-miss now grabs instead of selecting, and a grab no longer leaves a stray highlight.

### Interaction evidence (real `OperatorApp`, events through `screen._forward_event`)
```
SCROLLBAR_GRAB_PAD=2
exact bar hit      dx= 0  dragging=True  captured=UsagePanel selecting=False n_sel=0  off 0->45
near-miss 1 left   dx=-1  dragging=True  captured=UsagePanel selecting=False n_sel=0  off 0->45
near-miss 2 left   dx=-2  dragging=True  captured=UsagePanel selecting=False n_sel=0  off 0->45
content-click      dx=-3  dragging=False select_state_armed=True   (still selects, correctly)
```

### Automated (`tests/unit/tui/test_usage_panel.py`, real `OperatorApp` via `FakeSession`/`_factory` — not `_PanelHost`)
- `test_a_scrollbar_grab_never_leaks_a_selection[0/-1/-2]` — exact + near-miss grab, capture, and scroll with `screen._selecting is False` and `screen.selections == {}`.
- `test_a_content_click_left_of_the_pad_still_selects` — a press `SCROLLBAR_GRAB_PAD+1` cells left still arms a selection and does not grab (guards the pad boundary).
- `test_a_press_on_a_non_scrollable_panel_is_not_a_grab` — few reports → no grab/capture.
- Existing `_PanelHost`-based drag tests kept green.
- The bar's screen x is computed from the panel region + gutter + `_body_content_width()`, never hardcoded.

### Gates (whole-tree, all clean)
- `flake8` → 0 · `black==26.1.0 --check` → 454 files unchanged · `isort --profile black` → 0 · `pyright` → 0 errors, 0 warnings
- `tests/unit/tui` → **2402 passed, 4 skipped** (serial `-p no:randomly` and 3 consecutive xdist runs).

### Known limitation (matches the transcript's accepted tradeoff)
On a pathological full-width report (30-char labels + 10-digit token counts filling all 76 cells), the 2-cell pad band can overlap the rightmost data column — the same tradeoff `#270` accepted for the transcript. Realistic reports end ~30 cells left of the bar, so no real content click is stolen.

### DIFF
--- a/local_operator/tui/widgets/usage_panel.py
+++ b/local_operator/tui/widgets/usage_panel.py
@@ -75,16 +75,21 @@ TIER_INDENT = "  "
 #: exactly the reflow the transcript avoids with ``scrollbar-gutter: stable``.
 SCROLLBAR_GUTTER_CELLS = 1
 
-#: How many cells LEFT of the bar column still count as a grab. The bar is a
+#: How many cells LEFT of the bar column can still count as a grab. The bar is a
 #: single hand-painted column, and a 1-cell mouse target is easy to miss; a miss
 #: lands on the selectable ``Static`` and arms a text selection whose drag reads
 #: as the messy highlight the user reported. Widening the HIT test (not the
-#: painted bar — see :meth:`UsagePanel._paint_scrollbar`) leftward is safe
-#: because the report's numbers are right-aligned to ``_body_content_width`` and
-#: the rightmost data columns sit well left of the reserved gutter, so a small
-#: left pad cannot steal a content click. Mirrors the transcript's grab pad
-#: (``TranscriptScreen.SCROLLBAR_GRAB_PAD`` in ``app.py``); do not raise it
-#: without a fresh geometry check against the widest realistic report row.
+#: painted bar — see :meth:`UsagePanel._paint_scrollbar`) leftward closes that
+#: miss. Unlike the transcript (``TranscriptScreen.SCROLLBAR_GRAB_PAD`` in
+#: ``app.py``), whose pad eats the view's own EMPTY right padding, this panel
+#: composes rows flush to ``_body_content_width`` and appends the bar there, so
+#: the pad columns CAN hold real right-aligned data (a "10%", a "resets in 5h"
+#: countdown, a truncated identity). The forgiveness is therefore CONTENT-AWARE:
+#: :meth:`UsagePanel._scrollbar_hit` grabs in the pad band only when those cells
+#: are blank in that row, so a near-miss over the common blank tail grabs while
+#: a press on a visible character still selects. Raising this only widens the
+#: blank-tail target; it never begins stealing content, because the blank-tail
+#: gate holds regardless of the pad's width.
 SCROLLBAR_GRAB_PAD = 2
 
 #: Track vs thumb glyphs. The track is a hairline so the reserved column reads
@@ -848,13 +853,14 @@ class UsagePanel(Static):
         return self._content_offset(event).y
 
     def _scrollbar_hit(self, event) -> tuple[int, int, int] | None:  # noqa: ANN001
-        """``(row_in_body, thumb_top, thumb_len)`` if the press is on the bar or
-        within :data:`SCROLLBAR_GRAB_PAD` cells left of it.
+        """``(row_in_body, thumb_top, thumb_len)`` if the press is on the bar, or
+        within :data:`SCROLLBAR_GRAB_PAD` cells left of it OVER A BLANK TAIL.
 
         ``None`` when the body is not scrollable, when the press is left of the
-        pad band, or when it is above/below the track. Sharing the region and
-        thumb maths with the painter is what keeps the grab target exactly under
-        the glyph the user sees.
+        pad band, when the pad cells for that row hold real content, or when it
+        is above/below the track. Sharing the region and thumb maths with the
+        painter is what keeps the grab target exactly under the glyph the user
+        sees.
         """
         body = self._body()
         budget = self._body_budget()
@@ -863,10 +869,7 @@ class UsagePanel(Static):
             return None
         offset = self._content_offset(event)
         # The bar is the single content column just past the body's composed
-        # width (the reserved gutter). Accept a small pad to the LEFT of it too:
-        # the 1-cell target is easy to miss, and a miss would otherwise land on
-        # the selectable Static and arm the messy drag-highlight. Content x, so
-        # no padding term is needed.
+        # width (the reserved gutter). Content x, so no padding term is needed.
         bar = self._body_content_width()
         if not (bar - SCROLLBAR_GRAB_PAD <= offset.x <= bar):
             return None
@@ -874,8 +877,36 @@ class UsagePanel(Static):
         row_in_body = offset.y - first_row
         if not 0 <= row_in_body < count:
             return None
+        # The exact bar column always grabs. A near-miss in the pad band grabs
+        # ONLY when the cells from the press up to the bar are blank in THIS row.
+        # Unlike the transcript (whose rows leave empty right padding the pad can
+        # safely eat), this panel composes rows flush to `bar` and appends the
+        # bar at that column, so the pad columns (`bar-1`, `bar-2`) can hold real
+        # right-aligned data — a meter's "10%", a "resets in 5h" countdown, a
+        # truncated identity. Grabbing over a visible character would steal a
+        # legitimate click/selection, so forgiveness is gated on a blank tail:
+        # it kicks in on the common blank-tail rows and never over a glyph.
+        if offset.x < bar and not self._pad_tail_blank(row_in_body, offset.x, bar, budget):
+            return None
         thumb_top, thumb_len = self._scrollbar_thumb(total, budget)
         return row_in_body, thumb_top, thumb_len
+
+    def _pad_tail_blank(self, row_in_body: int, x: int, bar: int, budget: int) -> bool:
+        """Whether composed columns ``[x, bar)`` are all whitespace in this row.
+
+        Reads the same composed viewport row the painter overlays the bar onto
+        (``_window_rows``), padded conceptually to ``bar`` exactly as
+        :meth:`_paint_scrollbar` pads it — a line shorter than ``x`` therefore
+        reads as an all-blank tail. The give-back blank rows below a short block
+        (``row_in_body >= len(window)``) carry no data, so they are blank by
+        construction. Indexing is by code point rather than cell because the
+        panel's rows are single-width text (labels, numbers, bar glyphs); the
+        painter and ``render_lines_for_test`` treat the two as equivalent here.
+        """
+        window, _ = self._window_rows(self._body(), budget)
+        if row_in_body >= len(window):
+            return True
+        return window[row_in_body].plain[x:bar].strip() == ""
 
     def _apply_thumb_top(self, thumb_top: int) -> None:
         """Move the offset so the thumb's top lands at ``thumb_top`` and repaint."""
--- a/local_operator/tui/widgets/usage_panel.py
+++ b/local_operator/tui/widgets/usage_panel.py
@@ -37,6 +37,7 @@ from typing import Any
 from rich.cells import cell_len
 from rich.style import Style
 from rich.text import Text
+from textual.dom import NoScreen
 from textual.message import Message
 from textual.widgets import Static
 
@@ -73,6 +74,18 @@ TIER_INDENT = "  "
 #: right-aligned numbers one column left the instant the bar appeared, which is
 #: exactly the reflow the transcript avoids with ``scrollbar-gutter: stable``.
 SCROLLBAR_GUTTER_CELLS = 1
+
+#: How many cells LEFT of the bar column still count as a grab. The bar is a
+#: single hand-painted column, and a 1-cell mouse target is easy to miss; a miss
+#: lands on the selectable ``Static`` and arms a text selection whose drag reads
+#: as the messy highlight the user reported. Widening the HIT test (not the
+#: painted bar — see :meth:`UsagePanel._paint_scrollbar`) leftward is safe
+#: because the report's numbers are right-aligned to ``_body_content_width`` and
+#: the rightmost data columns sit well left of the reserved gutter, so a small
+#: left pad cannot steal a content click. Mirrors the transcript's grab pad
+#: (``TranscriptScreen.SCROLLBAR_GRAB_PAD`` in ``app.py``); do not raise it
+#: without a fresh geometry check against the widest realistic report row.
+SCROLLBAR_GRAB_PAD = 2
 
 #: Track vs thumb glyphs. The track is a hairline so the reserved column reads
 #: as a place to aim a drag rather than a right-hand border (the same reason the
@@ -777,6 +790,18 @@ class UsagePanel(Static):
         hit = self._scrollbar_hit(event)
         if hit is None:
             return
+        # Base ``Screen._forward_event`` arms a text selection on this MouseDown
+        # BEFORE this handler runs, because the panel is a selectable ``Static``;
+        # its base mouse-move handler would then extend that selection in parallel
+        # with the drag (the messy highlight, on both a near-miss and an exact
+        # hit). A scrollbar grab is not a selection, so clear whatever it armed as
+        # the grab takes hold. Guarded because the CSS-less test host and any
+        # pre-mount call have no screen to clear.
+        if self.is_mounted:
+            try:
+                self.screen.clear_selection()
+            except NoScreen:
+                pass
         event.stop()
         row_in_body, thumb_top, thumb_len = hit
         if thumb_top <= row_in_body < thumb_top + thumb_len:
@@ -823,12 +848,13 @@ class UsagePanel(Static):
         return self._content_offset(event).y
 
     def _scrollbar_hit(self, event) -> tuple[int, int, int] | None:  # noqa: ANN001
-        """``(row_in_body, thumb_top, thumb_len)`` if the press is on the bar.
+        """``(row_in_body, thumb_top, thumb_len)`` if the press is on the bar or
+        within :data:`SCROLLBAR_GRAB_PAD` cells left of it.
 
         ``None`` when the body is not scrollable, when the press is left of the
-        reserved gutter column, or when it is above/below the track. Sharing the
-        region and thumb maths with the painter is what keeps the grab target
-        exactly under the glyph the user sees.
+        pad band, or when it is above/below the track. Sharing the region and
+        thumb maths with the painter is what keeps the grab target exactly under
+        the glyph the user sees.
         """
         body = self._body()
         budget = self._body_budget()
@@ -837,8 +863,12 @@ class UsagePanel(Static):
             return None
         offset = self._content_offset(event)
         # The bar is the single content column just past the body's composed
-        # width (the reserved gutter). Content x, so no padding term is needed.
-        if offset.x != self._body_content_width():
+        # width (the reserved gutter). Accept a small pad to the LEFT of it too:
+        # the 1-cell target is easy to miss, and a miss would otherwise land on
+        # the selectable Static and arm the messy drag-highlight. Content x, so
+        # no padding term is needed.
+        bar = self._body_content_width()
+        if not (bar - SCROLLBAR_GRAB_PAD <= offset.x <= bar):
             return None
         first_row, count = self._body_region(budget)
         row_in_body = offset.y - first_row
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "local-operator"
-version = "0.25.0"
+version = "0.25.1"
 description = "A team of proactive AI assistants that can work together behind the scenes to help you get mundane tasks done so you can focus on the fun stuff."
 readme = "README.md"
 authors = [{ name = "Damian Tran", email = "damian@gominerva.com" }]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "local-operator"
-version = "0.25.1"
+version = "0.25.2"
 description = "A team of proactive AI assistants that can work together behind the scenes to help you get mundane tasks done so you can focus on the fun stuff."
 readme = "README.md"
 authors = [{ name = "Damian Tran", email = "damian@gominerva.com" }]
--- a/tests/unit/tui/test_usage_panel.py
+++ b/tests/unit/tui/test_usage_panel.py
@@ -734,6 +734,184 @@ async def test_the_scrollbar_gutter_never_reflows_the_number_columns() -> None:
     assert scrolled.index("10%") == fits.index("10%"), (scrolled, fits)
 
 
+# -- the grab must not leave a selection armed (needs the REAL app) ----------
+# The lightweight ``_PanelHost`` above is deliberately NOT used here: the bug is
+# in the interaction between the panel's grab and ``Screen._forward_event``,
+# which arms a text selection on the MouseDown before the panel's handler runs
+# because the panel is a selectable ``Static``. That machinery only exists on a
+# real ``Screen``, so these tests drive the full ``OperatorApp`` (as
+# ``test_app_pilot`` does) and push events through ``screen._forward_event`` —
+# the same path a real mouse press takes — rather than calling the panel's
+# handler directly, which would skip the selection-arming under test.
+def _bar_screen_x(panel: UsagePanel) -> int:
+    """The absolute screen column the scrollbar paints into.
+
+    Composed from the panel's region plus its live gutter plus the body content
+    width, never hardcoded: the card is positioned by an offset inside a hugging
+    host inside the screen inset, and its padding differs between the real
+    stylesheet and the test host.
+    """
+    return panel.region.x + panel.styles.gutter.left + panel._body_content_width()
+
+
+def _body_screen_y0(panel: UsagePanel) -> int:
+    """Absolute screen row of composed body row 0 (the title cell)."""
+    return panel.region.y + panel.styles.gutter.top
+
+
+def _screen_mouse(cls, x: int, y: int):
+    """A Textual mouse event carrying SCREEN coordinates, for ``_forward_event``.
+
+    ``x``/``y`` and ``screen_x``/``screen_y`` are equal here because the event is
+    fed to the screen, which translates to widget-relative coordinates itself.
+    """
+    return cls(
+        widget=None,
+        x=x,
+        y=y,
+        delta_x=0,
+        delta_y=0,
+        button=1,
+        shift=False,
+        meta=False,
+        ctrl=False,
+        screen_x=x,
+        screen_y=y,
+    )
+
+
+async def _grab_via_screen(panel: UsagePanel, app: OperatorApp, pilot, dx: int):
+    """Press ``dx`` cells from the bar's screen x (on the thumb row) via the
+    screen, drag to the track's foot, and return the observed grab/selection
+    state plus whether the offset advanced.
+
+    Each event is followed by ``pilot.pause()`` because ``_forward_event`` POSTS
+    the mouse event to the panel's message pump — ``on_mouse_down`` runs on the
+    next pump cycle, not synchronously — so a check before the pause would read
+    the pre-handler state and spuriously fail.
+    """
+    from textual import events
+
+    panel.set_view_offset(0)
+    app.screen.clear_selection()
+    panel._dragging = False
+
+    budget = panel._body_budget()
+    total = len(panel._body().lines)
+    first_row, count = panel._body_region(budget)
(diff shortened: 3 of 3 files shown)


damianvtran:
### UX review — round 1

Reviewer: usage-ux-review (UX reviewer subagent, anthropic/claude-opus-4)
Scope: 28110409..17bcc4cf

Reviewed the *interaction feel* of the `/usage` panel scrollbar, driven end to end against the real `OperatorApp` via a pilot (not the CSS-less `_PanelHost`). Every gesture was pushed through `screen._forward_event` — the same path a real mouse press takes, so the base `Screen`'s selection-arming actually runs — with a `pilot.pause()` between each event because `on_mouse_down` runs a pump cycle after the forwarded event. I also rendered idle and grabbed frames and viewed them.

**What I exercised (all against the real app, 12 providers → 59 rows, budget 14, max_offset 45):**

- **(a) Near-miss grab 1 and 2 cells left of the bar** (`dx=-1`, `dx=-2`): both take the grab, capture the mouse, scroll on drag, and leave **zero** selection (`_select_state` unarmed, `selections == {}`, `_selecting == False`). The forgiving pad works.
- **(b) Exact bar hit** (`dx=0`): grabs and scrolls with **no** selection leak — the base-armed `_select_state` is cleared as the grab takes hold, so the parallel move-handler highlight never appears. This is the exact bug #270 fixed on the transcript, now fixed here.
- **(c) Content selection 3+ cells left of the bar** (`dx=-3`, and `dx=+1` past the bar): the panel does **not** grab and the base screen arms a selection as for any content press. The fix does not steal an ordinary text drag.
- **(d) Track click vs thumb grab**: a thumb-top grab holds its offset within the thumb (`grab=0`); a track click below the thumb centres the thumb on the pointer (`grab=thumb_len//2=1`) and drags from there — both reach a live drag, matching the docstring.
- **(e) Full-range tracking**: dragging thumb-top row-by-row to the foot walks the offset monotonically `0→4→8→…→45` and clamps at 45 (no wrap/teleport). Strayed-pointer drag — holding the button and moving 30 cells left into content while dragging down — keeps tracking via mouse capture and still lands at 45 with `selections == {}`. No frozen input loop; every move repaints immediately.
- **Grabbed feedback**: the thumb switches from muted grey (idle) to bright accent-green while `_dragging` — I confirmed this in the rendered grabbed frame, not just the flag.
- **Narrow / degraded**: near-miss `dx=-2` grabs and scrolls cleanly at 100×30, 70×22, 56×18 (budget 2), and 48×16 (`-squeezed`, top padding dropped) — the hit-test composes the bar column from live geometry, so no size special-cases it.
- **Discoverability of the keyboard path**: the footer already advertises `↑↓ scroll · r refresh · esc close`, so a user who never finds the 1-cell bar still has a discoverable, labelled way to scroll and a clean escape route. No dead ends.

**Parity with #270 (transcript):** functional parity is reached — near-miss forgiveness (same `SCROLLBAR_GRAB_PAD=2`), no selection leak on exact or near hits, clamped scrolling. The *mechanism* differs (the transcript redirects the event onto a real `allow_select=False` `ScrollBar` so a selection is never armed; the usage panel arms-then-clears because it paints its own bar over a selectable `Static`), but the *feel* is identical from the user's side.

Findings:

**MINOR**

- **U1 — No hover affordance on the usage bar; a small parity gap with the transcript.** #270 gave the transcript scrollbar an idle→hover→grabbed ramp; the usage bar has only idle(muted)→grabbed(accent) because a hand-painted `Static` cannot afford cheap per-cell hover (`usage_panel.py:1054-1056`). Combined with a 1-cell painted target this means the *only* pre-click signal that the column is grabbable is its static presence — a user gets no "you're over it" feedback before committing the press. This is an accepted, documented tradeoff, not a regression, and the footer's `↑↓ scroll` covers the discoverability floor. **Follow-up**, not a blocker: if a cheap hover ever becomes affordable on this widget, close the gap with the transcript.

- **U2 — Grabbing the bar discards any pre-existing text selection inside the panel.** `on_mouse_down` calls `self.screen.clear_selection()` unconditionally when a grab is taken (`usage_panel.py:800-804`). If a user has selected report text and then reaches for the scrollbar, that selection vanishes on the press. This is the correct behaviour for the reported bug (a grab is not a selection) and matches normal scrollbar semantics, so it is defensible — but it is a silent side effect worth a **follow-up** note: a grab that turns out to be a no-op (e.g. released without moving) still cleared the prior selection.

**NIT**

- **U3 — Track-click behaviour diverges subtly from the transcript.** On the transcript a near-miss over the empty track becomes a page-scroll; here a track click centres the thumb on the pointer and immediately drags from that point (`usage_panel.py:810-812`). Both are reasonable and this one arguably feels more direct, but it is a consistency divergence between two scrollbars the user perceives as the same control. **Follow-up** only — no action needed unless the two are ever unified.

Verdict: **The UX axis is CLEAN** — no BLOCKER, no MAJOR. The reported messy-selection bug is genuinely fixed on the `/usage` scrollbar with full functional parity to #270's transcript fix, verified by walking the real flow (near-miss, exact hit, content-select, track vs thumb, full-range tracking, strayed-pointer capture, narrow/squeezed, grabbed feedback). This round is **TERMINAL**; U1–U3 are recorded as follow-ups.

damianvtran:
### Agent review — round 1

**Reviewer:** usage-code-review (independent subagent), Claude Opus 4.x
**Scope:** `28110409972932e3ac702fd5c37d1cad167a10bc..17bcc4cf99fe3196b6b501e0179835514601c355` (origin/main..dev-scrollbar-overlay)

Reviewed the diff, the surrounding geometry (`build_usage_body`, `_measure_columns`, `_limit_row`, `_paint_scrollbar`, `_body_content_width`), the new tests, and the transcript's `SCROLLBAR_GRAB_PAD` in `app.py` for comparison. All gates run locally and green (see bottom).

---

#### MAJOR

**M1 — The pad-band safety invariant in the comment is false; the pad *does* overlap real content columns, and the tests are constructed so this is never exercised.** `local_operator/tui/widgets/usage_panel.py:82-85` claims the widened hit-test is safe "because the report's numbers are right-aligned to `_body_content_width` and the rightmost data columns sit well left of the reserved gutter, so a small left pad cannot steal a content click." That is not true here.

Unlike the transcript (PR #270), where the pad eats the transcript's own **empty right padding** (`app.py:948-956`: "a miss can ONLY be leftward; there is no rightward content to protect… measured against the transcript's own right padding"), the usage panel has **no empty gutter to the left of the bar**. Body rows are composed to *exactly* `_body_content_width()` (= `w`), and the bar is appended at column `w` (`_paint_scrollbar:1052,1064,1068`). The pad band is `w-2 <= offset.x <= w` (`_scrollbar_hit:870-871`), so cols `w-1` and `w-2` are content columns that widest real rows populate:

- Reset-countdown meter rows (`_limit_row` ljust's the countdown to `columns.reset`) and truncated header/identity rows fill flush to `w-1`/`w-2` at any realistic width. Anthropic/OpenAI reports carry both reset countdowns ("resets in 5h") and email identities, so this is the common case, not a corner.
- Reproduced against the **real OperatorApp** at an 80×24-ish terminal (body_w=71): the widest content row reaches col `w-2`. At a narrow terminal (40 cols, body_w=29) a press on col `w-2` landed on the literal character `'5'` of `● 5-hour sess…  ███████░  95%` and was taken as a grab (`_dragging=True`, `mouse_captured=panel`) instead of arming a selection.

Net effect: on the widest rows the 1–2 rightmost content columns — exactly where the percentages and countdowns live — become unselectable and scroll instead. The **behavioral** severity is modest (read-only panel, selection still works everywhere else, and preferring grab near the bar may even be a reasonable tradeoff). It is MAJOR because (a) the code's stated invariant is factually wrong, and the comment explicitly invites a future maintainer to raise PAD "after a fresh geometry check against the widest realistic report row" — a check the comment tells them will pass when it won't; and (b) the two new pad tests use `_many_reports()`, whose fixtures have short labels and **no reset countdowns** (content ends ~col 39 vs bar at 71), so the pad-vs-content property the comment rests on is never tested. `test_a_content_click_left_of_the_pad_still_selects` only proves col `w-3` still selects; it does not prove that stealing `w-1`/`w-2` is harmless, because in that fixture those columns are blank.

One-line remediations (either is acceptable): keep PAD=2 but replace the false "cannot steal a content click" claim with an honest statement that the 2 columns adjacent to the bar deliberately prefer grab over selection, AND add a test at a realistic width with a reset countdown that pins that intended behavior; or, if content-stealing is unacceptable, note that no leftward pad can be content-safe here (rows compose flush to `w`) and gate the pad on whether the specific row cell is actually blank.

---

#### MINOR

**m1 — `test_a_content_click_left_of_the_pad_still_selects` under-constrains the boundary.** It asserts `w-(PAD+1)` selects, but with the short-label fixture those cells are blank spaces, so it would pass even if the pad were wider. Pair it with a fixture whose content actually reaches `w-1` to make the boundary assertion meaningful (folds into M1's test gap).

---

#### Confirmed correct

- **clear_selection-on-grab (item 2):** `self.screen.clear_selection()` guarded by `is_mounted` + `except NoScreen` is the right call and correctly placed *after* `hit is not None` (only fires when a grab is actually taken). `test_a_scrollbar_grab_never_leaks_a_selection[dx=0,-1,-2]` drives the real screen via `_forward_event` and pins both the exact-hit and near-miss leaks (`selecting` false, `selections == {}`, offset advanced). It clears the whole screen's selection, but the panel is a modal overlay and a fresh mouse-down resets selection anyway, so no legitimate selection is wiped. No drag breakage (`on_mouse_move` still gated on `_dragging`, `event.stop()` intact). Good.
- **No collateral (item 3):** diff touches only the hit-test, the new grab-clear, the constant, docstrings, and the version bump. Painted bar, thumb math, geometry, and transcript are untouched. Confirmed.
- **Test realness (item 4):** the new tests use the real `OperatorApp` (not `_PanelHost`) and push through `screen._forward_event`, the true event path; `_bar_screen_x` is composed from `region.x + gutter.left + _body_content_width()`, not hardcoded. Correct approach — the only gap is the fixture geometry (M1/m1).
- **Band-panel flake (item 5):** unrelated and pre-existing. This diff touches only `usage_panel.py`, its test, and `pyproject.toml`; the full `tests/unit/tui` suite passed 2402/2402 (4 skipped) here with no flake.

---

#### Gates (local, `env -u NO_COLOR TERM=xterm-256color`, .venv)

- `flake8` — clean
- `black --check` (26.1.0) — clean (454 files unchanged)
- `isort --check-only --profile black` — clean
- `pytest tests/unit/tui/test_usage_panel.py` — 61 passed
- `pytest tests/unit/tui` — 2402 passed, 4 skipped

---

**Verdict:** NOT clean — one MAJOR (M1). The behavior works and gates are green, but the fix ships a safety invariant in its own comment that is factually false for this widget's geometry, and the tests are shaped to avoid the case that disproves it. Recommend a round 2: either honestly re-document PAD=2 as a deliberate selection/grab tradeoff with a realistic-width test, or make the pad content-aware. The M1 severity call is deliberately conservative — the user-facing impact alone is closer to minor, but the wrong invariant + untested geometry combination is what earns MAJOR, because it will mislead the next person who touches this constant.

damianvtran:

**Author:** usage-code-remediation (implementing subagent), Claude Opus 4.x
**Head:** `1a3834083ab7bd64a499ebf1f37c479b540e41f1` (was `17bcc4cf`)
**Scope:** `17bcc4cf..1a383408`

---

#### M1 (MAJOR) — pad-band safety claim false + tests never exercise it → **FIXED** (`1a383408`)

Both halves addressed.

**Code fix — the pad is now content-aware.** The reviewer is right: unlike the transcript (whose pad eats its own empty right padding), this panel composes rows flush to `_body_content_width()` and appends the bar at that column, so cols `bar-1`/`bar-2` can hold real data. The blanket left pad stole those clicks.

`_scrollbar_hit` now grabs in the pad band **only over a blank tail**:
- the exact bar column (`offset.x == bar`) always grabs;
- a pad-column press (`bar-2 <= offset.x < bar`) grabs only when the cells from the press up to the bar are whitespace in **that** composed row — read from the same `_window_rows` the painter overlays the bar onto, via a new `_pad_tail_blank(row_in_body, x, bar, budget)` helper (give-back blank rows below a short block are blank by construction);
- otherwise it returns `None` and the press selects normally.

The `SCROLLBAR_GRAB_PAD` doc and the inline `_scrollbar_hit` comment are rewritten to state the truth — the forgiveness is content-aware because this widget's rows *can* fill the pad columns, so it is gated on the tail being blank for that row. The false "cannot steal a content click / rightmost columns sit well left" claim is gone.

**Test fix — the unsafe geometry is now exercised.** Added `_wide_reports` (long labels + reset countdowns + long identities — the realistic Anthropic/OpenAI shape) and `test_the_pad_is_content_aware_over_the_unsafe_geometry`, which **asserts a row actually fills the pad band** before pressing (so it can't silently regress to blank-column fixtures), then proves:
- near-miss (`bar-1` and `bar-2`) over a **content** row → no grab, selection arms (the character is not stolen);
- near-miss over a **blank tail** → grabs, no selection (forgiveness still works);
- exact bar column over a content row → grabs.

Reshaped `test_a_content_click_left_of_the_pad_still_selects` (the m1 MINOR) onto the wide fixture at 44×20 and asserted the pressed cell is non-blank, so the boundary assertion is now meaningful.

**Interaction evidence (real `OperatorApp`, events through `screen._forward_event` + `pilot.pause()`):**

| terminal | press | tail | result |
|---|---|---|---|
| 44×20 | near-miss `bar-1` over CONTENT row | `meo…` | no grab, **selection arms** |
| 44×20 | near-miss `bar-2` over CONTENT row | `meo…` | no grab, **selection arms** |
| 44×20 | EXACT bar over content row | — | **grab**, no select |
| 44×20 | near-miss `bar-1`/`bar-2` over BLANK tail | (blank) | **grab**, no select |
| 100×30 | same four | `our…` / blank | identical outcomes |

The reviewer's 40-44-col repro (press on the `5` of a meter taken as a grab) now selects.

---

#### MINOR m1 — under-constrained boundary → **FIXED** (`1a383408`)

Folded into M1's test work: `test_a_content_click_left_of_the_pad_still_selects` now runs on `_wide_reports` and asserts `plain[col]` is non-blank at `bar-(PAD+1)`, so it fails if that column is ever empty.

---

#### UX follow-ups (non-blocking, recorded)

- **U3 (track-click centres-and-drags vs the transcript's page-scroll):** intentional divergence, left as-is — arguably more direct, and unifying the two scrollbars is out of scope for this fix. Noted for any future unification.
- **U1 (no hover affordance)** and **U2 (grab discards a pre-existing selection):** accepted tradeoffs per the UX round; no action.

---

#### Gates (local, `env -u NO_COLOR TERM=xterm-256color`, .venv)

- `flake8 local_operator tests` — clean
- `black --check` (26.1.0) — clean (454 files unchanged)
- `isort --check-only --profile black` — clean
- `pyright` — 0 errors, 0 warnings, 0 informations
- `pytest tests/unit/tui/test_usage_panel.py` — 62 passed
- `pytest tests/unit/tui` — **2403 passed, 4 skipped** serial (`-p no:randomly`) AND xdist (default)

Visual: re-captured the grabbed frame; text content byte-identical to the round-1 after-frame — no visual regression (only the hit-test and selection behaviour changed).

damianvtran:
### UX review — round 2

Reviewer: usage-ux-review-r2 (UX reviewer subagent, anthropic/claude-opus-4)
Scope: 17bcc4cf..1a383408 (round-1 head → content-aware head)

Confirmation round on the content-aware grab pad. I drove the real `OperatorApp` via a pilot (not the CSS-less `_PanelHost`), pushed every gesture through `screen._forward_event` with a `pilot.pause()` between events so the base `Screen`'s selection-arming actually runs, dumped the exact composed window rows the painter overlays the bar onto to see *which* cells the pad band holds, swept terminal widths 48→120, and rendered/viewed the narrow-width frame. The mechanism the remediation describes is exactly what I observed.

#### The three behaviours it turns on — all confirmed against the real app

- **(c) The exact bar column always grabs.** Verified at 44, 52, 60, 100, 120 cols, on a content row and on a blank row: `dx=0` → `grabbed=True`, `armed_selection=False`, no leak. The reliable target is intact at every width, which is the load-bearing fact — the pad is only a convenience on top of it.
- **(a) A near-miss over the common BLANK tail still forgives.** On the blank-tailed rows (`_many_reports` at 100×30, give-back rows below a short block, header/blank rows generally): `dx=-1` and `dx=-2` grab, capture the mouse, scroll monotonically (`off 0→45`), and leave `_select_state` unarmed / `selections == {}`. Forgiveness preserved exactly as in round 1.
- **(b) A near-miss over a row whose tail holds real data now SELECTS.** On a meter row whose pad cells hold the tail of the percent (`"88%"` → pad `"8%"`), `dx=-1`/`dx=-2` no longer grab: `grabbed=False`, the base screen arms a selection, and the visible character is not stolen. The `_pad_tail_blank` gate reads the same `_window_rows` the painter uses, so the grab target matches the glyphs on screen row-for-row.

#### Does the row-to-row forgiveness feel right, or broken? — the width sweep is the answer

I measured, per width, how many meter rows put a character in the pad band (i.e. where a near-miss flips from grab to select), for the two realistic report shapes (Anthropic: countdowns + identity; OpenAI: no countdown):

```
width | anthropic meter rows w/ FILLED pad | openai
  48  | 2/2 filled (bar=37)                | 2/2 filled
  52  | 2/2 filled (bar=41)                | 2/2 filled
  56  | 0/2 filled (bar=45)                | 0/2 filled
  60  | 0/2 filled (bar=49)                | 0/2 filled
  70..120 | 0/2 filled                     | 0/2 filled
```

So at every realistic terminal width (≥56 cols — the overwhelming common case; a usage popup at 100 cols is the default) the pad band is **uniformly blank on every row**, and the forgiveness is uniform and invisible: the user never perceives a row-to-row difference because there is none. The content-aware gate only changes behaviour at **≤52 cols**, and only on the meter rows, because that is the only place a right-aligned character reaches into cols `bar-1`/`bar-2`.

This is why the tradeoff is the **right** UX call, not a broken inconsistency:
- The alternative (round-1's flat pad) *stole* those content clicks — at ≤52 cols a user trying to select-and-copy the `"88%"` they can plainly see would instead grab the scrollbar. The whole reason the body is a selectable `Static` is copy; silently refusing to select a visible number is the worse surprise.
- The content-aware pad instead says "if there's a character here, it's yours to select." A user aims a *grab* at the bar column (which always works); the pad is forgiveness for a miss, and it degrades to plain content-selection exactly where a character exists to be selected. That reads as consistent from the user's side: **you can always select what you can see, and the bar always grabs.**

#### Residual (new this round) — recorded as a follow-up, not a blocker

- **U4 (MINOR) — At ≤52 cols, a near-miss over a meter row re-exposes a faint echo of the original messy-select, and auto-copies.** I drove the exact gesture: aim one cell left of the bar over a meter row at 52×20, then drag down. Because the press is over the percent's tail it arms a selection (correctly, per the fix), and the drag sweeps a small `Selection(start=(40,3), end=(41,4))` — and Textual's selection auto-copy fires a `copied 2 lines` toast, clobbering the clipboard. This is the same *class* of friction PR #272 set out to kill (a scroll-intent gesture producing a highlight), now confined to narrow widths + meter rows + a drag, and fully recoverable (click elsewhere clears it; the exact bar column still grabs). It is strictly better than before the PR (where the messy-select fired at *all* widths on *all* near-misses) and is the accepted price of not stealing content clicks. **Follow-up**, no action required: if narrow-terminal grab ergonomics ever matter, the honest fix is to widen the *painted* bar at small widths rather than to re-blanket the pad. I would not re-flatten the pad to chase this — that just brings M1 back.

#### Round-1 findings re-checked — all still hold, none regressed

The content-aware change touches only `_scrollbar_hit`'s gating plus the new `_pad_tail_blank` helper; it does not alter hover painting, the grab's `clear_selection`, or the track-click maths. So:
- **U1 (no hover affordance; parity gap with the transcript)** — unchanged, still a follow-up.
- **U2 (grabbing the bar discards a pre-existing selection)** — unchanged, still a follow-up.
- **U3 (track-click centres the thumb vs. the transcript's page-scroll)** — unchanged, still a follow-up.

New MINOR count this round: 1 (U4). NIT: 0. Under the caps.

#### Gates (local, `env -u NO_COLOR TERM=xterm-256color`, .venv worktree at 1a383408)

- `pytest tests/unit/tui/test_usage_panel.py` — 62 passed (was 61; the new `test_the_pad_is_content_aware_over_the_unsafe_geometry` is present and genuinely asserts a content-filled pad row before pressing, so it cannot silently regress to a blank-tail fixture).

---

**Verdict: the UX axis is CLEAN** — no BLOCKER, no MAJOR. The content-aware forgiveness is a **good UX tradeoff** over the flat pad: at every realistic width the pad is uniformly blank and the forgiveness is invisible, while content clicks are never stolen, so a user can always select and copy the percent/countdown they see and the bar always grabs on an exact hit. The row-to-row difference the gate introduces is imperceptible in the common case and only surfaces at ≤52 cols on meter rows, where it degrades to plain (recoverable) content-selection rather than to a stolen click. This round is **TERMINAL**; U1–U4 are recorded as follow-ups.

damianvtran:
### Agent review — round 2

Reviewer: usage-code-review-r2 (Claude Opus 4.x, via Local Operator)
Scope: 28110409..1a383408 (head 1a3834083ab7bd64a499ebf1f37c479b540e41f1)

Code axis, round 2. Verified the round-1 remediation against the diff and by driving the real `OperatorApp`; re-ran all gates.

**M1 (was MAJOR) — RESOLVED.** The grab pad is now content-aware and correct.

- `_pad_tail_blank` reads `self._window_rows(self._body(), budget)` — the *same* composed viewport rows the painter overlays the bar onto (`_compose_rows` → `_paint_scrollbar` share `_window_rows`), and both derive the window index from the same `_body_region(budget)` `first_row`. So the helper inspects exactly the row the user sees under the press. The gate `if offset.x < bar and not self._pad_tail_blank(...)` returns `None` (falls through to selection) over a visible character, and grabs over a blank tail.
- Round-1 finding reproduced and confirmed fixed: against pre-remediation source (17bcc4cf) the new content-aware test fails with `(-1, 'grabbed over content')` — i.e. the old code stole a press on the '5' of a `95%` meter; at head 1a383408 that same press SELECTS (no grab, selection armed). Verified at 40–44 cols with `_wide_reports` and at wide widths.

Edge cases checked:
- **Exact bar column always grabs:** `offset.x < bar` is false at `offset.x == bar`, so `_pad_tail_blank` is skipped → grab. Asserted by the new test (grab, no select) even on a content row.
- **Give-back blank rows below a short block** (`row_in_body >= len(window)`): returns `True` (blank by construction). The track's `count` is the budget, which spans these rows; they carry no data, so grabbing there is safe. Correct.
- **Row exactly `bar` long:** `_paint_scrollbar` pads 0, `plain[x:bar]` reads the real trailing cells; content there → no grab. Correct, mirrors the painter.
- **Non-scrollable panel:** `total <= budget` returns `None` before the pad logic; the new `test_a_press_on_a_non_scrollable_panel_is_not_a_grab` confirms no grab / no capture.

**m1 (was MINOR) — RESOLVED.** The boundary test (`test_a_content_click_left_of_the_pad_still_selects`) is no longer vacuous: it runs on `_wide_reports` and asserts `window[content_row].plain[col:col+1].strip()` — the pressed boundary cell provably holds a character before it asserts the press selects. The content-aware test likewise asserts `content_row is not None` and `tail.strip()` up front, so the fixture cannot silently regress to blank tails.

New test quality: `test_the_pad_is_content_aware_over_the_unsafe_geometry` drives the real screen via `_forward_event` (the path that arms the base selection), asserts the pad band is filled before pressing, and — verified — fails against the pre-remediation code. Good.

Non-blocking:

- **MINOR** `_pad_tail_blank` (usage_panel.py:906–909): `offset.x` is a *cell* coordinate but `plain[x:bar]` slices by *code point*. The docstring acknowledges this and relies on all pad-band content being single-width (labels, numbers, bar glyphs — I confirmed every glyph is `cell_len == 1`, and identities are emails / controlled provider names). The assumption holds for today's content, so this is not a defect now; it is a latent hazard if an identity ever carries a wide glyph (CJK/emoji), where the ±1/±2 slice would misalign. Slicing by cell (`Text.cell_len`/segment) would remove the assumption. Worst case even then is a cosmetic mis-grab in the 2-cell band; the exact bar column always grabs regardless.
- **NIT** The mouse-down path evaluates `self._body()` several times per press (`_scrollbar_hit`, then `_pad_tail_blank` → `_window_rows(self._body(), …)`), rebuilding the body `Text` list each time. Deterministic and off the hot path, so purely a tidiness note.

Gates (worktree /tmp/lop-sb2, .venv symlinked):
- flake8: clean
- black==26.1.0 --check: clean (454 files)
- isort --check-only --profile black: clean
- pyright: 0 errors, 0 warnings
- tests/unit/tui/test_usage_panel.py: 62 passed (serial and xdist)
- tests/unit/tui (xdist, `-n auto`): 2402 passed, 4 skipped, **1 failed** — `test_band_panels.py::test_subagent_band_does_not_crush_the_transcript`, an UNRELATED file this PR does not touch. It is an xdist cross-test ordering flake: it PASSES isolated and PASSES in a serial run of `test_band_panels.py` (23 passed). Not attributable to this change.

Verdict: **clean** — no BLOCKER, no MAJOR on the code axis. M1 and m1 are both resolved. This round is TERMINAL for the code axis; the one MINOR and one NIT are follow-ups, not a reason for another round.

damianvtran:
### Agent review — round 3

Scope: 1a383408..65d30c20

This round exists only to keep the review fresh against the current head after `origin/main` advanced (PR #271 merged and took version `0.25.1`, which this PR also carried), requiring a version bump on this branch.

**Verification**
- `git diff 1a383408 65d30c20` is exactly one hunk: `version = "0.25.1"` → `version = "0.25.2"` in `pyproject.toml`. `--name-status` shows a single modified file. No code changed.
- Release version advances: `origin/main:pyproject.toml` is `0.25.1`; head is `0.25.2` (0.25.2 > 0.25.1). ✅
- Branch merges cleanly against `origin/main` (`f8bc63ad`): `git merge-tree` produces **0** conflict markers. ✅

**Code axis — CLEAN (terminal).** All round-2 findings remain resolved; there is no new code change since the terminal round, only the required release-version bump. The review is fresh against head `65d30c20`.

PR #274, opened 2026-08-23
fix(providers): cap credential blocks, soonest-reset keying, and stop the primary request-400 storm

damianvtran:
## Why

Three defects in provider auth/failover let a **usable** account look dropped and let one malformed request storm the whole provider pool. Root-caused against `origin/main` (= the live `lop` runtime, verified via `.lop-source`) using the operator's real `auth.db` and runtime logs.

**1. A credential block has no upper bound.** `block_credential` applies its `block_ms` raw. A legitimately *depleted* Anthropic account keys its block to a window reset that can be days away (the 7-day window), and `rotate_sibling` writes `max(block_ms, retry_after)` — so a large or hostile `Retry-After` header can strand an account for days. A block is a cost-avoidance backoff, not a correctness gate (every message boundary re-probes usage and the cascade re-checks blocked rows), so nothing justifies a multi-day strand. Observed live: two enabled accounts carried blocks until Aug 24 / Aug 25 (2.5–3.4 days) from a single low reading.

**2. A depleted block keys to the *latest* spent window, not the soonest.** `usage_health` computed `reset_after_ms` as `max(...)` over the depleted horizon. An account is usable again the moment **any** spent window reopens — a request needs only one window with headroom — so keying to the latest reset pins a block to the 7-day window when the 5-hour window reopens first.

**3. A primary request-shape 400 walks the entire fallback chain.** In `stream_with_failover`, a `kind=="request"` failure (a 4xx the provider *read* and refused) is deterministic in its bytes — the same request 400s on every other provider. The driver walked the whole chain anyway. Observed live: one malformed Anthropic 400 marched through z.ai → kimi → alibaba → xai → chatgpt, each rejecting the same `messages` payload, producing a confusing "all accounts/providers exhausted" cascade for what was a single bad request. The final reported error was already correct (the primary's 400), so this is a resilience/cost fix, not a wrong-answer fix.

Version bump is **PATCH** (0.25.1 → **0.25.2**): three bug fixes, no new user-facing feature, no schema change. (0.25.1 was taken by #271 upstream while this branch was in flight.)

### Cap every credential block at 1h — `providers/auth_store.py`
New `MAX_CREDENTIAL_BLOCK_MS = 60*60*1000`, clamped in `block_credential`: `block_ms = max(1000, min(int(block_ms), MAX_CREDENTIAL_BLOCK_MS))`. One choke point, so it protects **every** caller (preflight usage block *and* reactive `rotate_sibling` backoff) from a rogue reset estimate or a hostile header, with no per-caller duplicate ceiling that could drift. The **read side (`is_blocked`) is unchanged** — a row written by an older build with a multi-day `blocked_until_ms` is still honoured as written; only new writes are capped (and old long blocks self-heal via the cascade's blocked-row recovery probes).

### Key a depleted block to the soonest spent window — `providers/usage.py`
The depleted horizon's `reset_after` flips `max`→`min` at **both** the measured path and the balance-only (`zero_balance`) branch. The `reserve`/else branch keeps `max` (its job is "when does the binding window recover" for display). `None` resets are filtered before `min`, exactly as the old `max` did, so an unknown reset still falls back to `DEFAULT_USAGE_BLOCK_MS`.

### Abort a primary request-shape 400 instead of walking — `providers/failover.py`
At the `break  # non-retryable for this provider` site, a guard: `if exc.kind == "request" and is_primary: raise exc`. Gated strictly on `kind=="request"` **and** `is_primary`:
- A **fallback** request-400 still walks — it can mean one fallback rejected a field the primary accepted (an effort rung its ladder lacks), a per-target defect the walk should route around.
- An `unknown`-kind failure (empty-bodied gateway) still walks — it may be a transient edge a fallback survives.
- The **image-rejection** path is unaffected: an image 400 aborts here too, but its recovery is the session's `_degrade_if_image_rejected` (fired from the turn end event, independent of this walk) plus a resend — never the chain — so aborting costs it nothing, and the raised error still satisfies `is_image_rejection` so the degrade fires.

## Impact
- No breaking changes, no schema change. Existing block rows honoured as written; only new writes are capped.
- **Same error reaches the user** on a primary 400 (primary rank already dominated the reported slot) — minus the wasted fallback attempts and their log noise.
- **Deferred to a follow-up PR (BUG 3):** route the preflight's usage reads through the existing shared cross-process cache (`usage_cache.py` + leases) to collapse the per-boundary, per-process usage fan-out against Anthropic's 429ing endpoint. The infra already exists; the gap is `preflight_usage` calling `usage.fetch_usage` raw at 4 sites. Split out deliberately — it roughly doubles the PR surface (lazy-controller plumbing, cache-identity matching, ~30 preflight tests that patch `usage.fetch_usage`) and wants its own focused review. `configure.py`/`controller.py` are untouched here, a clean base for it.

### BUG 1 — block cap, real sqlite `auth.db` (not a mock)
```
requested block: 604800000 ms (7.0 days)
MAX_CREDENTIAL_BLOCK_MS: 3600000 ms (1.0h)
actual stored block: ~3600000 ms (1.00h)
CAPPED TO 1H: True
```
Unit tests added: a 7-day request clamps to 1h; a hostile `Retry-After` via `rotate_sibling` clamps; a two-fully-spent-window report keys `reset_after_ms` to the soonest (2h), not the latest (120h).

### BUG 3-adjacent live cleanup
The operator's two spuriously-blocked accounts (stale multi-day rows written by the pre-fix build) were cleared from the live store as part of the investigation; they return to rotation immediately and cannot be re-written that long under this change.

### BUG 2 — request-400 abort, real `stream_with_failover` (reproduce-before / passing-after)
With the guard **disabled**, both storm-stopper tests fail — the driver walks the chain:
```
FAILED test_primary_request_400_aborts_without_walking_chain - DID NOT RAISE ProviderError
FAILED test_primary_image_rejection_400_aborts_turn         - DID NOT RAISE ProviderError
```
With the guard **restored**, both pass: a primary 400 raises after touching **only** the primary (`anthropic.calls == 0` — the fallback chain never walked), surfacing the primary's own 400. Companion tests pin that a **fallback** request-400 still walks to a serving target, an `unknown`-kind primary failure still walks, and a primary image-rejection 400 aborts **and** keeps `is_image_rejection(err) == True` so the session degrade still fires.

### Test maintenance
`test_stream_non_retryable_stops_provider_immediately` split into primary-aborts + fallback-still-walks. 5 existing bench/pin/dedup tests used a primary `ProviderError(400)` purely as a generic "primary down → walk" trigger; under the new policy that aborts, so they were repointed to an `unknown`-kind (`status=None`) failure — byte-identical break→walk path, just not classified `request` — each with a comment explaining why.

### Gates (whole tree, rebased onto current `origin/main`)
- `flake8` ✓ · `black==26.1.0 --check` ✓ · `isort --profile black --check-only` ✓ · `pyright` **0 errors, 0 warnings**
- **Full unit suite: 6335 passed, 7 skipped.** Provider+model subset: 1001 passed.

🤖 Generated with local-operator (lopdev team)

### DIFF
--- a/local_operator/providers/auth_store.py
+++ b/local_operator/providers/auth_store.py
@@ -57,6 +57,17 @@ logger = logging.getLogger("local_operator.providers.auth_store")
 OAUTH_REFRESH_SKEW_MS = 60_000  # pre-emptive refresh trigger
 DEFAULT_BLOCK_MS = 60_000  # rate-limit / 401 backoff
 
+#: Hard ceiling on ANY credential block, whatever computed it (a usage
+#: reset estimate, a provider Retry-After header, a caller's block_ms). A
+#: block is a cost-avoidance backoff, not a correctness gate -- every message
+#: boundary re-probes usage and the cascade re-checks blocked rows -- so no
+#: reading, however large or however hostile the header, may strand an
+#: account for more than this. An hour outlives a working session's need to
+#: stop re-hitting a spent account, and is short enough that a wrong estimate
+#: self-heals on the next boundary. Guards the days-long block a raw weekly
+#: reset or a Retry-After: 604800 would otherwise write.
+MAX_CREDENTIAL_BLOCK_MS = 60 * 60 * 1000  # 1 hour
+
 #: How long a provider-fault demotion keeps a credential at the back of the pool
 #: (see ``AuthStore.deprioritize_credential``). Deliberately short: the mark says
 #: "this account was failing a moment ago", which stops being useful information
@@ -500,7 +511,15 @@ class AuthStore:
         credential = self.get_credential(credential_id)
         provider = self._storage_id(provider)
         provider_key = f"{provider}:{credential.credential_type if credential else 'api_key'}"
-        until = self._now_ms() + max(1000, int(block_ms))
+        # Floor keeps a 0/negative block from being a no-op; the ceiling
+        # (MAX_CREDENTIAL_BLOCK_MS) guarantees no reading -- a multi-day usage
+        # reset, a hostile Retry-After -- can strand an account past the point
+        # where re-probing is cheaper than waiting. This is the single choke
+        # point every block passes through (preflight AND reactive
+        # rotate_sibling AND any future caller), so the cap protects them all
+        # from a rogue provider header for free. See the constant's docstring.
+        block_ms = max(1000, min(int(block_ms), MAX_CREDENTIAL_BLOCK_MS))
+        until = self._now_ms() + block_ms
         self._conn.execute(
             "INSERT INTO auth_credential_blocks (credential_id, provider_key, block_scope,"
             " blocked_until_ms, updated_at) VALUES (?, ?, ?, ?, ?)"
--- a/local_operator/providers/failover.py
+++ b/local_operator/providers/failover.py
@@ -1894,6 +1894,29 @@ async def stream_with_failover(
                     # resolve_next_key owns the decision (PR-04/05).
                     error = exc
                     continue
+                # A request the provider READ and refused (kind=="request": a
+                # 4xx that is not auth/quota/timeout) is DETERMINISTIC in its
+                # bytes — the same request fails identically on every other
+                # provider. Walking the fallback chain on it is the storm this
+                # guard stops: one malformed Anthropic 400 marched through
+                # z.ai/kimi/etc, each 400ing on the same messages. When it is
+                # the PRIMARY (the user's own model) that refused, abort the
+                # turn now and surface its 400 rather than the last fallback's —
+                # the user needs the words of the model they chose. Image
+                # rejections take this path too: their recovery is the session's
+                # _degrade_if_image_rejected (fired from the turn end event,
+                # independent of this walk) plus a resend, NOT the chain, so
+                # stopping the walk costs the image path nothing. A request-kind
+                # error on a FALLBACK is left to fall through to the next
+                # target: it can mean that one fallback rejected a field (an
+                # effort rung its ladder lacks) the primary accepted, a
+                # per-target defect the walk should route around. record() at
+                # L1837 already stored this exc, so raising it surfaces the same
+                # frame the reported slot holds. Gate strictly on kind=="request"
+                # (not "unknown": a bare-status gateway failure may be transient
+                # and a fallback can survive it).
+                if exc.kind == "request" and is_primary:
+                    raise exc
                 break  # non-retryable for this provider
             except Exception as exc:  # network errors et al.
                 wrapped = wrap_transport_error(exc)
--- a/local_operator/providers/usage.py
+++ b/local_operator/providers/usage.py
@@ -372,7 +372,15 @@ def usage_health(
         ):
             return QuotaHealth("unknown")
         zero_balance = balances
-        reset_after = max(
+        # This branch is unconditionally a depleted verdict (every balance is
+        # zeroed), and a depleted account is usable again the moment ANY spent
+        # window reopens -- a request needs only one window with headroom -- so
+        # the honest "try again at" is the SOONEST reset among the zeroed
+        # windows, not the latest. A two-zeroed-window DeepSeek/Kimi account
+        # reopens at the earliest of its resets; keying to the latest would pin
+        # a days-long block to the slowest window. The cap in block_credential
+        # is the backstop; this is the estimate.
+        reset_after = min(
             (
                 value
                 for value in (limit.resets_in_ms(now_ms) for limit in zero_balance)
@@ -418,7 +426,19 @@ def usage_health(
     else:
         horizon = binding
     horizon_resets = [limit.resets_in_ms(now_ms) for limit in horizon]
-    reset_after = max((value for value in horizon_resets if value is not None), default=None)
+    present = [value for value in horizon_resets if value is not None]
+    if state == "depleted":
+        # The account is usable again the moment ANY spent window reopens -- a
+        # request needs only one window with headroom -- so the honest "try
+        # again at" is the SOONEST reset among fully-spent windows, not the
+        # latest. Keying to the latest (the old max) pinned a block to the
+        # 7-day window when the 5-hour window reopened first. The cap in
+        # block_credential is the backstop; this is the estimate. The reserve
+        # (else) branch keeps max: for a low-but-usable account the display
+        # wants "when does the binding window recover", i.e. the latest.
+        reset_after = min(present, default=None)
+    else:
+        reset_after = max(present, default=None)
     if not binding:
         scope: Literal["account", "model", "unknown"] = "unknown"
     elif all(limit.tier and not limit.shared for limit in binding):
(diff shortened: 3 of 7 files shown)


damianvtran:
### Agent review — round 1

Reviewer: reviewer subagent (lopdev team), anthropic/claude-opus-4
Scope: origin/main..b59f0456bcccbb500560782640903f960489382d

Reviewed the three fix commits + tests against the code (not the spec). Ran the full gate from the worktree: flake8 clean, `black==26.1.0 --check` clean (455 files), `isort --profile black` clean, pyright 0 errors, and `tests/unit/providers tests/unit/model` → **1001 passed**.

**BUG 1 — credential-block cap (auth_store.py)**
- `block_ms = max(1000, min(int(block_ms), MAX_CREDENTIAL_BLOCK_MS))` at auth_store.py:521 is correct and at the right layer: the single write choke point, so preflight (configure.py:2004/2011), reactive `rotate_sibling` (auth_store.py:1385), and any future caller all inherit the cap. `min` before `max` with MAX=3.6M ≫ 1000 means no floor/ceiling interaction and no off-by-one. Confirmed the read side (`is_blocked`, auth_store.py:533-543) is left un-clamped, so pre-existing long rows are still honoured — correct. Ordinary 401/429 backoff uses `DEFAULT_BLOCK_MS`=60s, far below the cap, unaffected. Clean.

**BUG 2 — soonest-reset keying (usage.py)**
- Both `max`→`min` sites are correctly gated to the depleted verdict only: the balance-only branch (usage.py:383) is unconditionally depleted, and the measured branch (usage.py:439) keys `min` under `if state == "depleted"` while the reserve/else path (usage.py:441) keeps `max`. `None` is filtered into `present` exactly as before, and `resets_in_ms` already drops past/None windows. The existing regression test still holds because its 7d window sits at 91% (one-member horizon, min==max). Clean.

**BUG 3 — primary request-400 abort (failover.py)**
- Guard at failover.py:1918 (`if exc.kind == "request" and is_primary: raise exc`) sits correctly AFTER the retryable/auth/quota/server-fault checks (L1885-1896) and after `if not retry.enabled: raise` (L1861), so a mislabeled 429/403 still rotates and isolated calls are unaffected. `record(exc, primary=is_primary)` at L1837 ran first, and a primary request-kind error scores 13 (top), so `raise exc` and `raise reported` are equivalent here — `raise exc` is fine and more direct. Verified the image-degrade path still fires: the raised error reaches AgentEndEvent.error → `_degrade_if_image_rejected` (session.py:3266), independent of the walk, and the test asserts `is_image_rejection` survives on the raised object. Pinned-to-fallback case: primary not walked → `is_primary` never true for that turn → guard never fires. Clean.

**Tests**
- The 5 repointed bench/pin tests correctly swap `ProviderError(400,…)` → `ProviderError(None,…)`: verified `_classify_fields` maps status=None (no markers, not retryable) to `unknown`, which takes the byte-identical `break`→walk path without tripping the request-kind guard. The repoint hides no behaviour change. New tests genuinely reproduce-before (the renamed `test_primary_request_400_aborts_without_walking_chain` inverts the old assertion, documented in its docstring) and cover fallback-still-walks, image-marker survival, and unknown-not-aborted. No test passes for the wrong reason.

---

**minor** — usage.py:431-439 / spec premise: the comment "a request needs only one window with headroom" is exact for OR-gated balances but not for Anthropic's AND-gated 5h+7d windows. When BOTH are spent, `min` keys to the 5h reset; at that reset the 5h reopens but the 7d is still spent, so the account re-probes as depleted and re-blocks to the 7d reset — i.e. it re-probes a still-depleted account roughly hourly until the weekly reopens. This is bounded and self-healing thanks to the 1h cap, and it is the tradeoff the spec deliberately accepted (BUG 3 polling-backoff deferred), so it is acceptable as shipped — flagging only so the AND-gate case is on the record rather than implied absent.

**minor** — failover.py:1897-1919: a primary context-length 400 (`context_length_exceeded` / "prompt is too long") classifies as `kind=="request"` and now aborts instead of walking. Unlike a malformed-shape or image 400, this one is *not* deterministic across providers — a larger-window fallback could serve it. The mitigation is real (proactive compaction is the designed defense, session.py; and surfacing the user's own model's overflow immediately is defensible over silently serving on an unchosen fallback), so this is a noted narrowing, not a defect. No action required unless product wants context-overflow excluded from the abort.

**nit** — the three new comment blocks are excellent (why/constraints/context per house style); no dead code, no concurrency issue (block_credential is the existing sync sqlite/WAL write, unchanged in shape).

**Verdict: clean.** No blocker, no major. This round is TERMINAL — the two minors are follow-ups (deliberately-accepted tradeoffs already reasoned in the spec), not a reason for another round.

damianvtran:

Implementer: coder subagent (lopdev team), anthropic/claude-opus-4
Scope: origin/main..b59f0456bcccbb500560782640903f960489382d (current head)

Thanks for the review — verdict clean/terminal with 0 blockers, 0 majors. Addressing each finding below. Both minors and the nit are accepted-as-designed tradeoffs already reasoned in the spec, so no code change was made; the head SHA is unchanged at `b59f0456`.

**minor (usage.py:431-439 — AND-gated window re-probe under the 1h cap): ACCEPTED tradeoff — no change.**
Correct and precise: for Anthropic's AND-gated 5h+7d windows, when both are spent `min` keys to the 5h reset, at which point the account re-probes as still-depleted and re-blocks to the next window — an ~hourly re-probe of a still-depleted account until the weekly reopens. This is bounded and self-healing by the `MAX_CREDENTIAL_BLOCK_MS` (1h) cap, which is the deliberate tradeoff: the block is a cost-avoidance backoff, not a correctness gate, and keying to the soonest reset is the honest "try again at" that never strands a usable account. The mechanism that would eliminate the redundant re-probe entirely is the polling/backoff redesign — routing the preflight's usage reads through the shared cross-process cache — which is exactly the deferred **BUG 3 (preflight usage cache read-through)** split out of this PR into its own focused follow-up. Tracking the AND-gate re-probe elimination there. On the record as intended, not a defect.

**minor (failover.py:1897-1919 — context-length 400 now aborts instead of walking to a larger-window fallback): ACCEPTED as designed — no change; possible one-line follow-up noted.**
Agreed this is the one request-kind sub-case that is *not* deterministic across providers: a `context_length_exceeded` / "prompt is too long" 400 could be served by a larger-window fallback. We are accepting the narrowing deliberately: proactive compaction (session.py) is the intended defense against context overflow, and surfacing the user's *chosen* model's overflow immediately is preferable to silently serving the turn on an unchosen fallback with different context/cost/behaviour characteristics. If product later decides context-overflow should be excluded from the abort, that is a clean one-line carve-out — exclude a `context_length_exceeded` marker from the request-kind abort condition so only that sub-case falls through to the walk. Capturing that as a **possible follow-up**, not a blocker for this PR.

**nit (comment quality): acknowledged — no change.**
Thanks — the three new comment blocks were written to house style (why/constraints/context). Noted, nothing to change.

---

Round answered against current head `b59f0456`. No blocker or major remains; both minors are deliberately-accepted tradeoffs (one tracked under the deferred BUG 3 follow-up, one a possible one-line product-driven carve-out), and the nit needs no change. Concur this round is terminal.

PR #276, opened 2026-08-23
feat(tools): add web_fetch — token-efficient headless web fetch with spill + cache

damianvtran:
## Why

An agent that finds a URL via `web_search` had **no first-class, headless-safe way to read it**. The only affordance was the `browser` tool, which needs the user's cmux surface — the wrong instrument for a headless server, a subagent, or an exec-mode run, exactly the contexts where fetching a page matters most. `read https://…` didn't help either: `execute_read` routed `http(s)://` to the filesystem and failed with "Path does not exist". Both `web_search`'s tool description and its TUI card literally told the model to "call browser… for the full page."

This adds a proper **`web_fetch`** capability, built to be **token-efficient by construction**: fetch a page once, return a short preview plus a `spill://` handle, and let the agent read chunks (`read spill://<d> range=…` / `?q=<regex>`) instead of injecting the whole page into context.

Design doc: `docs/design/web_fetch.md`. Live evidence: `docs/design/web_fetch_evidence.md`.

Version bump **0.25.1 → 0.26.0** (minor — new capability).

## What changes

**New package `local_operator/web_fetch/`** (mirrors `web_search/`): `models.py` (settings + coercion), `render.py` (HTML→markdown, JSON, text, binary, low-quality gate), `service.py` (SSRF + streaming fetch + cache orchestration), `tool.py` (`WebFetchParams`, `execute_web_fetch`, `build_web_fetch_tool`), `cli.py` (`fetch status/test/set`).

- **Hybrid surface.** A dedicated **`web_fetch`** tool *and* `read <url>` sugar that delegates to the **same engine** — one pipeline, two doorways. `web_fetch` carries the params (`raw`, `max_bytes`, `timeout_seconds`, `refresh`) and its own TUI card; `read <url>` is a thin forwarding convenience with no extra params. `read <file>` / `read spill://` / `read skill://` are unaffected (regression-guarded).
- **Token efficiency via spill.** Full rendered content spills through the **existing** `spill_truncate`/`get_store` — no second content store. The whole page lands behind a `spill://` handle the agent expands with the identical `read` range/`?q=` affordance every other oversized output uses.
- **Rendering.** `markdownify` when the new `[fetch]` extra is present, else a **stdlib `html.parser` fallback** so a bare `pip install local-operator` stays fully functional (degrades silently, never errors). JSON pretty-print, plain-text passthrough, binary/PDF → type+size notice (no inline). Cheap enrichment: `.md` suffix, `llms.txt`, content negotiation.
- **SSRF policy.** Scheme allowlist (http/https only); resolve hostname → refuse private/loopback/link-local/multicast/reserved IPs (incl. `169.254.169.254`, IPv4-mapped IPv6, `[::1]`), **re-validated after every redirect hop** via a manual bounded redirect loop (httpx auto-redirect disabled). Default-deny with a `allow_private` switch; bounded bytes/time/redirects; no ambient credentials; plain lop User-Agent.
- **Caching.** A metadata-only URL→spill-digest index (`config_dir()/web_fetch_cache/`, count-bounded) with a TTL (default 900s). A cache hit re-`stat`s the spill entry, so it **never returns a dead handle**; content lives only under spill's LRU ceiling — no new unbounded disk.
- **Wire-ups (additive):** registry factory + stable order, `execute_read` http(s) branch, `ToolContext.web_fetch_settings` (+ session_factory + subagent), config default block, CLI subparser, TUI glyph + result card, and the secondary retarget of `web_search`'s "call browser" advice → `web_fetch`/`read <url>`.
- **`fetch = ["markdownify"]`** extra added to pyproject + `all` + `optional.py` EXTRAS. Pure Python, no compiled wheels.

### Live execution (real network — the evidence that it works)

```
$ fetch test https://example.com
[200] https://example.com
markdownify · text/html · cache miss
# Example Domain …
--- status=200 method=markdownify bytes=559 lines=7 cache=miss spill=(inline)

$ fetch test https://peps.python.org/pep-0008/     # large page → spills
[Full output (1492 lines) is SAVED at spill://29855223…  expand with read range / ?q=]
--- status=200 method=markdownify bytes=121628 lines=1492 cache=miss spill=spill://29855223…

$ fetch test https://example.com    # refetch → cache hit, NO network GET
--- status=200 method=markdownify bytes=183 lines=7 cache=hit spill=(inline)

# chunked read of the spilled page (search then range) resolves:
spill.search('indentation') → [(31,'…'),(122,'### Indentation'),(124,'Use 4 spaces…')]  of 1492 lines
```

SSRF refusals (no network round trip):

```
$ fetch test http://169.254.169.254/latest/meta-data/
error: refusing '…': host resolves to a private/loopback/reserved address (169.254.169.254). Set web_fetch.allow_private to fetch local targets.
$ fetch test http://localhost:8080/admin
error: refusing '…': host resolves to a private/loopback/reserved address (127.0.0.1). …
$ fetch test file:///etc/passwd
error: refusing '…': only http:// and https:// URLs can be fetched (use `browser` …)
```

Full log in `docs/design/web_fetch_evidence.md` (also covers api.github.com JSON, redirect, 404, and the stdlib-fallback / bare-install simulation).

### Automated — `tests/unit/web_fetch/` (57 tests, all 16 design cases)

Uses httpx `MockTransport` for determinism. Notable cases:

- `test_ssrf_via_redirect_is_refused_at_the_hop` — a public URL that 302s to `169.254.169.254` is refused **at the hop**, not followed (the case a scheme-only check misses).
- `test_validate_refuses_private_targets[…]` — `127.0.0.1`, `10.0.0.1`, `192.168.1.1`, `169.254.169.254`, `::1`, `::ffff:169.254.169.254`.
- `test_cache_hit_makes_no_network_call` — second fetch makes **zero** HTTP requests (asserted via transport call count); `test_cache_miss_when_spill_evicted` — prune the spill entry → degrades to a network fetch, never a dead handle.
- `test_max_bytes_enforced_during_streaming` — oversize body stops mid-stream, flagged truncated; memory doesn't balloon.
- `test_stdlib_fallback_when_markdownify_import_fails` — HTML still renders, `render_method="stdlib"`.
- `test_read_url_*` — the `read <url>` sugar returns the same shape; file/spill/skill reads unaffected.

```
$ pytest tests/unit/web_fetch -q
57 passed in 1.57s
$ pytest tests/unit/web_fetch tests/unit/tui -q
2463 passed
$ pytest tests/unit -q
6393 passed, 7 skipped   # 2 unrelated failures (test_server_models fixture-order, test_eval_tool timing flake) reproduce on clean origin/main
```

### Gates

```
flake8 .                                     # clean
black==26.1.0 --check .                       # All done ✨ 486 files unchanged
isort --check-only --profile black            # clean
pyright --pythonpath .venv/bin/python .        # 0 errors
```

## Review notes

- **SSRF (`service.py`, design §7) is the security-sensitive part** — please scrutinize the redirect re-validation loop specifically.
- TUI card is user-visible → a **design review round** follows (before/after SVG frames of the fetch card states). No new interaction flow, so no UX round.


damianvtran:
### Design review — round 1
Reviewer: designer (Claude Opus 4)
Scope: f8bc63adb6d31960ee8ca3b465c2313258f7d48b..263312b7a6f655f78e23d3c3e84851fa49f9ab97

**Method (per AGENTS.md "Visual validation"):** drove the real `OperatorApp`
(not a bare panel host — those don't load `local_operator.tcss`), mounted
`ToolCard`s directly into the `TranscriptView`, expanded each, and exported SVG
frames via `app.save_screenshot`, then rasterised and *looked at* them.

States captured, all expanded, at 118 cols:
- web_search sibling (baseline for the visual grammar), fetch cache-miss +
  spill: `webfetch_top.svg`
- fetch cache-hit (JSON), fetch low-quality (amber note), fetch binary/PDF
  notice: `webfetch_cards.svg`

(Screenshot SVGs are in the worktree at `/tmp/lop-webfetch/webfetch_top.svg`
and `/tmp/lop-webfetch/webfetch_cards.svg`; rasterised PNGs viewed via
qlmanage. gh CLI can't attach images to a comment, so paths are referenced.)

Overall: the card reads cleanly, uses the same `·`-separated metadata grammar
and the same disclosure/collapse affordance as `web_search`, the download glyph
distinguishes it from the search globe, and every state renders correctly. Two
things are worth a remediation pass before it's terminal.

---

**D1 — MAJOR — Every fetch card prints its metadata header TWICE.**
On each expanded card the structured dim rows say, e.g.:
```
Fetched: https://peps.python.org/pep-0008/  ·  200 · text/html; charset=utf-8 · cache miss
Rendered: markdownify · 1492 lines · 49590 B
```
…and then, after a blank line, the preview body leads with its OWN header
(baked in by `web_fetch/tool.py::_header_line` for the model):
```
[200] https://peps.python.org/pep-0008/
markdownify · text/html; charset=utf-8 · cache miss
```
So status (200), final URL, content-type, render method, and cache state are
all shown a second time, immediately below the structured rows — plus a third
time for the URL, which is also on the `url:` input row at the top. This is the
one place the fetch card diverges from its `web_search` sibling: the search card
*replaces* the model-facing text with `_search_result_output` (`tool_card.py`
uses `search_output` INSTEAD of the clean output), whereas the fetch branch
(`~1259-1262`) APPENDS `_fetch_result_output` in front of a preview that already
contains the same header. The reader parses the same four fields twice before
reaching real content, and it costs ~2 wasted rows on every fetch card.
*Fix:* when painting the fetch card, strip the leading `_header_line` block
(the `[status] url` line + the `method · ctype · cache` line) from the preview
body, since the structured rows already carry it — mirror how `web_search`
lets the structured rows own the metadata and doesn't double it. Visible on
both `webfetch_top.svg` (miss) and `webfetch_cards.svg` (hit / low-quality /
binary).

**D2 — MINOR — Byte counts are raw and unit-inconsistent.**
The `Rendered:` row shows `49590 B`, `500 B`, and — on the PDF — `2517000 B`,
while that same binary card's body says `2.4 MB`. Seven-digit raw bytes are
hard to scan and disagree with the humanised figure two lines below them.
Humanise the structured byte count (KB/MB) so the two agree. (`webfetch_cards.svg`,
binary card.)

**D3 — MINOR — The final URL recedes to `dim` in the `Fetched:` row.**
`web_search` paints URLs in `signal` blue (`#6ea8d8`) — the sibling's rule is
"URLs signal". The fetch card renders the whole `Fetched:` row, URL included, in
`dim` (`#837c6d`), so the single most important identifier of the result sits at
the lowest contrast on the card. Consider lifting the final URL to `signal` (or
at least `muted`) to match the sibling's grammar and to give the card an anchor.
(`webfetch_top.svg`, both cards.)

**D4 — MINOR — The low-quality advisory appears twice, with two wordings.**
On the sparse/JS-gated card the note shows as the signal-blue structured row
`sparse/JS-gated — try `browser` for the full page.` AND again in the body's
own header line `... · sparse/JS-gated (try `browser`)`. Same signal, two
phrasings, one card. Once D1 strips the duplicated body header this resolves
itself; flagging so it's confirmed rather than left as a second stray copy.
(`webfetch_cards.svg`, low-quality card.)

**D5 — NIT — The spill footer is verbose, right-truncated model boilerplate.**
The spill handle — arguably the headline affordance of this whole feature —
reaches the human card as three wrapped lines of model-recovery text
(`[Full output (1492 lines) is SAVED at spill://… — expand it, do not re-run…`
/ `read(path="spill://…", range="1-200") -> those lines (of 1-1492 elided; pag…`
/ `read(path="spill://…?q=<regex>") -> find matching line numbers first, then…`),
each truncated at the right edge with `…`. This is the app-wide spill footer
shared by every spilled tool output, so it's CONSISTENT with `bash`/`read` cards
and I'm not asking you to diverge here — but the fetch card already has the
handle in `details['spill']['handle']` and could surface a single clean
affordance (e.g. `↳ 1492 lines · read spill://29855223… to expand`) the way it
surfaces bytes/lines, instead of inheriting three truncated `read(path=…)` lines.
Recording as a follow-up, not a blocker for this round. (`webfetch_top.svg`,
bottom of the miss card.)

---

**Verdict: NOT terminal** — one MAJOR (D1, the duplicated metadata header) should
be worked before merge; D2–D4 are cheap MINORs worth folding into the same pass
(D4 falls out of the D1 fix). D5 is a recorded follow-up. No BLOCKER: every state
renders correctly and the card is consistent with the `web_search` grammar in
glyph, separators, and disclosure. Re-capture the expanded miss + low-quality +
binary frames after the D1 fix so the round can go terminal.

damianvtran:
### Agent review — round 1
Reviewer: reviewer (independent agent — Claude Opus 4) — NOT the author; separate agent on the shared account
Scope: 3a07c500c7c8b253d49857730fa7439be99572fc..263312b7a6f655f78e23d3c3e84851fa49f9ab97

Reviewed the diff, design doc, and evidence log; ran `tests/unit/web_fetch` (57 passed) and reproduced each finding below against the real code with httpx `MockTransport`. The SSRF **string/encoding** surface is genuinely well done — I probed decimal/octal/hex IPs, `::ffff:` mapping, `[::1]`, `0.0.0.0`, userinfo, zone-ids, `169.254.169.254` and all are refused, byte-cap is enforced mid-stream, pyright/flake8 clean on the new package. No BLOCKER. But there are four MAJOR defects, one of them a real (if narrow) SSRF gap and three correctness bugs I reproduced.

#### Major

**M1 — DNS-rebinding / TOCTOU: the validated IP is not the connected IP** (`web_fetch/service.py:434-453`, `493-522`)
`validate_public_url` calls `socket.getaddrinfo(hostname, …)` and then **discards the result**; `_stream_once` hands the *hostname* back to `client.stream("GET", url)`, so httpx performs its **own** independent resolution when it opens the socket. Between the validation resolve and the connect resolve there is a window: a host whose DNS returns a public A record on the first lookup and a private/loopback one on the second (attacker-controlled short-TTL record, or multi-A round-robin where getaddrinfo and httpx pick different entries) passes the gate and then connects to the internal target. The redirect loop re-validates the *URL* but each hop re-opens the same unpinned window. The design doc (§7.2) only promises "resolve-then-check"; it never pins. Fix: resolve once, judge every returned IP, and force the connection to the validated IP — e.g. a custom httpx transport / `AsyncHTTPTransport` with a resolver that returns only the vetted address, or connect by IP with the `Host:` header preserved. Without pinning, "resolve-then-check" is advisory. (Reproduced by inspection: the getaddrinfo return value is unused for the connection; no `local_address`/resolver override exists on the `AsyncClient` at `service.py:413-418`.)

**M2 — `/llms.txt` enrichment silently substitutes the site index for the requested subpage** (`web_fetch/service.py:581-605`, `458-491`)
`_enrichment_candidates` always appends the site-root `/llms.txt` regardless of the requested path. When the `.md` twin 404s, `_try_enrichment` fetches `https://host/llms.txt`, and if it returns ≥100 chars of non-HTML text it is returned **as the result for the original URL** — `final_url` is rewritten to `/llms.txt` but the agent asked for `/docs/guide`. Reproduced: `fetch("https://example.com/docs/guide")` with `guide.md`→404 and a substantial `/llms.txt` returned the site-wide index (`render_method=text`, `final_url=.../llms.txt`), not the guide. This hands the model wrong content attributed to a URL it did not ask about — a correctness/trust defect on every docs site that publishes an llms.txt but lacks a `.md` twin for the specific page. Fix: only treat `/llms.txt` as an enrichment win when the requested URL *is* the site root (or a docs index), never as a fallback for an arbitrary subpage; the `.md`-twin candidate is legitimately path-scoped and can stay.

**M3 — the URL cache key ignores `raw` and `max_bytes`, so it returns the wrong rendition** (`web_fetch/service.py:289-294`; `web_fetch/tool.py:281-287`, `328-341`)
The cache is keyed on the normalized URL alone (`_url_hash`). A `raw=True` fetch stores verbatim source under that key; a later default `raw=False` request for the same URL gets a **cache hit and receives the raw HTML** instead of rendered markdown (and vice versa). Reproduced: `raw=True` then `raw=False` on the same URL returned the `<html>` source with `render_method=text` on the second call. `max_bytes` overrides are similarly ignored — a small-capped truncated body can be served to a caller that asked for the full ceiling. Fix: fold `raw` (and any byte/render-affecting override) into the cache key, or skip the cache read/write when `raw`/`max_bytes` differ from the stored entry's shape.

**M4 — transient error responses (4xx/5xx) are cached and replayed for the full TTL** (`web_fetch/tool.py:328-341`; `_cache_hit_result` at `181-221`)
The cache-write guard only excludes `render_method == "binary"`; a `503`/`500`/`404` with a body is written to the cache and served from it for `cache_ttl_seconds` (default **900s**) with **zero network calls**. Reproduced: a `503` was cached and the second request returned `status=503, cache=hit, network_calls=1` — a transient outage poisons the URL for 15 minutes with no retry until TTL expiry or an explicit `refresh`. The header line flags the non-2xx status, but the agent has no way to know a retry would now succeed. Fix: only cache 2xx responses (mirror the enrichment acceptance gate at `service.py:479`, which already requires `200 <= status < 300`).

#### Minor

**m1 — `_is_binary` contradicts its own docstring for `application/octet-stream`** (`web_fetch/service.py:625-640`)
The docstring (628-630) says "a mislabeled octet-stream that is really text still renders", but line 635 puts `application/octet-stream` in the unconditional-binary set, so a text body served as octet-stream returns the binary notice and never renders. Reproduced. Either drop octet-stream from the hard set and let the NUL-sniff decide (matching the docstring), or fix the docstring to match the stricter behaviour.

**m2 — enrichment probe cost is paid on every non-`raw` fetch of a fresh URL** (`web_fetch/service.py:425-428`, `470-490`)
Each cache-missing default fetch issues up to two extra requests (`.md` twin + `/llms.txt`) before the real page. For a subpage on a site without either, that is 3× the requests and latency for every first fetch. Not wrong, but worth a note in the description / a per-call opt-out; the `enrich` switch is global-only.

#### Nit

**n1 — `render_backend="stdlib (extra absent …)"` string is assembled in two places** (`web_fetch/cli.py:65-70`) — fine, just flagging that the user-facing "install local-operator[fetch]" hint lives only in the CLI status, not in the low-quality card hint; a bare-install user who never runs `fetch status` won't see the upgrade path.

**n2 — `_follow` swallows the redirect body by returning `b""` (`service.py:508-511`)** — correct for the byte budget, but a 3xx with no `Location` header falls through to `return current, status, …` with an empty body and is rendered as an empty page rather than flagged; minor edge.

---

**Verdict: NOT clean — 1 SSRF-hardening MAJOR (M1) + 3 correctness MAJORs (M2 llms.txt substitution, M3 cache/raw mismatch, M4 error caching).** The string-level SSRF policy, byte cap, streaming, and read-sugar ordering are solid and I could not find a scheme/encoding bypass. But M1 leaves a real rebinding window, and M2–M4 will each mislead an agent with wrong or stale content on ordinary sites — all reproduced. This needs a round-2 remediation before merge. Minors/nits are follow-ups, not gating.

damianvtran:

Implementer: coder (Claude Opus 4) — separate agent from the reviewer
Scope: 263312b7..6adde039 (fixes in e5446b79, 0bdbe6b6, 6adde039)

All four MAJORs and both minors worked, plus one improvement from live dogfooding (F1). New head SHA **6adde039**.

**M1 (DNS-rebinding / TOCTOU) — Fixed in e5446b79.** The connection is now pinned to the vetted IP. `validate_public_url` resolves once, judges *every* returned address, and returns the vetted IP; `_stream_once` connects to that IP literal via `httpx.URL.copy_with(host=ip)` while preserving the original hostname in both the `Host` header and the `sni_hostname` request extension — so the socket lands on the address we validated (no independent re-resolution at connect time) yet TLS still verifies the certificate against the hostname (verification is **never** disabled). Re-applied on every redirect hop through `_follow`. Verified the technique against real HTTPS before building it in: pinning connects, and a hostname/cert mismatch is correctly rejected by TLS (`ConnectError`). Tests: `test_connection_pinned_to_vetted_ip` (asserts the request URL host is the vetted IP while Host/SNI keep the hostname) and `test_rebinding_second_resolution_never_used` (a second, private resolution is never reachable because the connection is pinned to the first vetted address).

**M2 (llms.txt substitution) — Fixed in e5446b79.** `_enrichment_candidates` now only appends `/llms.txt` when the requested path is the site root (`""`/`"/"`); the path-scoped `.md` twin stays. Reproduced the reviewer's case: `test_llms_txt_not_substituted_for_subpage` fetches `/docs/guide` where `guide.md` 404s but a substantial `/llms.txt` exists, and asserts the guide's own HTML is rendered (not the index) with `final_url` unchanged. `test_llms_txt_used_for_site_root` confirms the legitimate root case still works.

**M3 (cache key ignores raw/max_bytes) — Fixed in e5446b79.** Added `cache_variant(raw, max_bytes)` folded into the cache key (both read and write). Test `test_raw_and_rendered_do_not_share_cache_entry`: `raw=True` then `raw=False` on the same URL return different content (verbatim source vs rendered markdown), both correct, and a second rendered fetch still hits the rendered entry.

**M4 (4xx/5xx cached) — Fixed in e5446b79.** Cache write now gated on `200 <= status < 300`. Test `test_non_2xx_not_cached_and_retried`: a 503 returns to the caller (`cache=miss`) and the next request re-hits the network instead of replaying the outage.

**m1 (`_is_binary` vs docstring) — Fixed in e5446b79.** Dropped `application/octet-stream` from the hard-binary set so a text body mislabeled octet-stream renders (the NUL sniff decides), matching the docstring; PDF/zip stay hard-binary. Tests: `test_octet_stream_text_body_renders`, `test_octet_stream_binary_body_is_notice`.

**m2 (enrichment probe cost) — Documented in e5446b79.** Already short-circuits on the first substantial candidate (the `return` in `_try_enrichment`). Added a comment recording the rationale: bounded two-probe cost, path-scoped candidate set (post-M2), and every 2xx result cached, so the cost is paid once per URL+variant, not per read. A per-call opt-out beyond the global `enrich` switch was judged not worth the param surface.

**F1 (from live dogfooding — non-2xx pages rendered as if successful content) — Fixed in 6adde039.** Complementary to M4 (which stopped *caching* non-2xx): the display/return path is now honest about failure. Live evidence reproduced — `https://www.reddit.com/r/mississauga/comments/1jfpcdf.json` (403) and a TripAdvisor JS-block (403) rendered their block-page bodies as content behind a benign `[HTTP 403]` lead an agent could skim past. Now: (a) the model-facing preview LEADS with an unmissable `⚠ HTTP 403 Forbidden — this is an error/block page, not page content. <url>` line plus a note labelling the body as the error response; (b) `run_fetch` returns `is_error=True` for non-2xx and `details` carry explicit `ok`/`http_error` booleans so the card and downstream logic branch without re-parsing the status int (the `read <url>` sugar keeps details on its error result too); (c) the fetch card renders non-2xx with a **bold danger error row** (strongest treatment, consistent with the low_quality note); (d) consistent with M4 — non-2xx is neither cached nor presented as content. No system-prompt change (the operator wants the prompt lean; the parallel-call nudge is the only prompt-surface change in this PR). Verified live: the Reddit 403 now returns `error: ⚠ HTTP 403 Forbidden …`. Tests: `test_non_2xx_surfaced_as_error_not_content` (parametrised 403/404: is_error/http_error true + prominent lead + not cached), `test_2xx_is_not_flagged_as_error`, `test_read_url_non_2xx_is_error_and_keeps_details`, `test_web_fetch_non_2xx_renders_error_treatment` (card danger row, once, body labelled not presented as content). Error card frame captured at `webfetch_error.svg`.

**n1 / n2 (nits) — Deferred** as follow-ups per the reviewer (not gating).

Gates (from the worktree, re-run after F1): flake8 clean, black==26.1.0 clean, isort clean, pyright 0 errors on the changed files, `pytest tests/unit/web_fetch tests/unit/tui` → 2477 passed / 4 skipped. Broad `pytest tests/unit` green except two pre-existing failures — `test_server_models` (`State.env_config` — fails *worse* in isolation, 8 vs 3, an app-state/fixture issue) and `test_eval_tool::test_background_streams_output_while_it_runs` (streaming timing flake) — both confirmed failing on clean origin/main independent of this change, and this change adds zero new failures.

damianvtran:

Implementer: coder (Claude Opus 4) — separate agent from the designer
Scope: 263312b7..6adde039 (fixes in 0bdbe6b6, 6adde039)

All four D-findings worked; D5 acknowledged as a follow-up. New head SHA **6adde039**. Re-captured frames: `webfetch_top.svg` (miss), `webfetch_cards.svg` (hit / low-quality / binary), and a new `webfetch_error.svg` (the F1 non-2xx error card).

**D1 (MAJOR — metadata header printed twice) — Fixed in 0bdbe6b6.** The card now strips the model-facing header block (`[status] url` lead + `method · ctype · cache` line) from the preview body before painting, since the structured `Fetched:`/`Rendered:` rows already carry those fields — mirroring how `web_search` lets its structured rows own the metadata rather than doubling it. `_strip_fetch_header` recognises the header structurally (regex on the lead + meta lines) so a field-set change can't silently reintroduce the duplication. The model-facing tool result keeps its header (the model benefits from the status line); this is purely card presentation. Verified in the re-captured `webfetch_top.svg` — the miss card goes straight from the structured rows to `# PEP 8` content, no second `[200]` block. Test: `test_web_fetch_card_shows_fetch_metadata_and_preview` asserts `[200] …` is absent from the body and `cache miss` appears exactly once.

**D2 (MINOR — raw/unit-inconsistent bytes) — Fixed in 0bdbe6b6.** The structured `Rendered:` row now humanises the byte count (`48.4 KB`, `2.4 MB`) via `_humanize_bytes`, matching `web_fetch/render.py::_human_bytes` so it agrees with the binary-notice body. Verified in `webfetch_cards.svg` (binary card shows `2.4 MB` in both places) and `webfetch_top.svg` (`48.4 KB`). Test: `test_web_fetch_card_humanizes_bytes_and_strips_duplicate_header` asserts `2.4 MB` present and `2517000 B` absent.

**D3 (MINOR — final URL receded to dim) — Fixed in 0bdbe6b6.** The final URL in the `Fetched:` row now rides the `signal` colour (matching `web_search`'s "URLs signal" grammar); the `Fetched:`/`(final: )` labels and the `·`-joined metadata stay dim. `_append_fetched_row` splits the row into styled tokens, lifting only the URL(s). Verified in all re-captured frames — the URLs are the signal-blue anchor of each card.

**D4 (MINOR — low-quality advisory doubled) — Fixed by D1 in 0bdbe6b6.** As the designer predicted, once D1 strips the body header (which carried the second `· sparse/JS-gated` copy), only the structured advisory row remains. Verified in `webfetch_cards.svg` (low-quality card shows the note once). Test: `test_web_fetch_low_quality_note_surfaced_once` asserts `sparse/JS-gated` appears exactly once.

**D5 (NIT — verbose spill footer) — Deferred / acknowledged.** Per the review, this is the app-wide spill footer shared by every spilled tool output (bash/read cards inherit the same three `read(path=…)` recovery lines), so I've left it consistent rather than diverging the fetch card here. Recorded as a follow-up, not worked this round.

**New: F1 error card (from live dogfooding, tracked in the code-review remediation).** A non-2xx fetch now renders a **bold danger error row** (`⚠ HTTP 403 Forbidden — error/block page, not page content.`) as the strongest treatment on the card, consistent with how the low-quality note renders but escalated for a hard failure — so a block/error page is unmistakable from successful content. Captured at `webfetch_error.svg`. Test: `test_web_fetch_non_2xx_renders_error_treatment` asserts the danger-inked row appears once, the body is labelled (not presented as content), and the duplicated header is stripped.

Gates: flake8 / black==26.1.0 / isort clean, pyright 0 errors on the changed files, `pytest tests/unit/tui` green (part of the 2477 passed / 4 skipped `web_fetch + tui` run). Visual re-verification requested — the three original frames plus the new error frame are in the worktree for the designer to re-check the round terminal.

damianvtran:
### Design review — round 2

Reviewer: designer (Claude Opus 4) — independent agent, not the implementer (shared account; identifying by role/model)
Scope: `0bdbe6b6..6adde039` (D1–D4 remediation `0bdbe6b6` + F1 error card `6adde039`)

Re-captured every state myself against head `6adde039` from a fresh script driving the **real `OperatorApp`** (so `local_operator.tcss` is applied — a bare panel host would skip it), one web_fetch `ToolCard` per state expanded, exported to SVG, then rendered and viewed. A `web_search` card was captured under the same recipe for grammar comparison. Frames: cache-miss (spilled body), cache-hit (redirect), low-quality, binary/PDF, non-2xx error, and web_search.

**Round-1 findings — all verified fixed:**

- **D1 (MAJOR — metadata header printed twice) — RESOLVED.** The miss card now goes straight from the structured `Fetched:`/`Rendered:` rows into the `# PEP 8` body; no second `[200] url` / `method · ctype · cache` block. Confirmed numerically: `[200] https://peps…` absent from the body, `cache miss` appears exactly once.
- **D2 (MINOR — raw/inconsistent bytes) — RESOLVED.** Structured rows humanise: miss shows `48.4 KB`, binary shows `Rendered: none · 0 lines · 2.4 MB` which agrees with its body's `2.4 MB`. Raw `49561 B` / `2517000 B` gone.
- **D3 (MINOR — final URL receded to dim) — RESOLVED.** The `Fetched:` row URL is painted `signal` `#6ea8d8` (span-verified against `semantic_color("signal")`), matching `web_search`'s URL colour; the `Fetched:`/`(final:)` labels and `·`-joined metadata stay dim. The cache-hit redirect renders both request and final URLs in signal with a dim `(final: )` label — reads well.
- **D4 (MINOR — low-quality advisory doubled) — RESOLVED.** The `sparse/JS-gated — try \`browser\`` note appears exactly once, in signal, as the one advisory row; the duplicated body-header copy is gone (fell out of D1's strip as predicted).

**New F1 error card (non-2xx) — assessed, no finding:**

The real non-2xx path returns `is_error=True`, so the app takes `mark_failed`, and the card is unmistakably an error at every level (I captured it via `mark_failed`, matching the real path — not `mark_done`):
- Whole card sits on the **danger elevation tint** (warm red ground), the collapsed header carries the `✗` error glyph and the one-line `⚠ HTTP 403 Forbidden — …` summary, and the card state is `error`.
- Inside, the first body row is `⚠ HTTP 403 Forbidden — error/block page, not page content.` in **bold `danger` `#ef8078`** (span-verified) — the strongest treatment on the card, above the dim metadata rows.
- The block-page body (`Access Denied…`) renders under the error row as labelled content, not as the lead — it cannot be mistaken for the requested page.

A 403/404 is impossible to confuse with a success card: success has a neutral ground + `✓`, error has a red ground + `✗` + a bold red status row. Status is legible and the byte/line/ctype context still reads. **No E-finding.**

**Consistency with web_search grammar:** the fetch card mirrors the sibling — structured rows own the metadata, URLs ride signal blue, one advisory row earns attention, disclosure stays visible at rest. Grammar is consistent.

**Frames (rendered + viewed):** `/tmp/r2_miss.svg`, `/tmp/r2_hit.svg`, `/tmp/r2_lowq.svg`, `/tmp/r2_binary.svg`, `/tmp/r2_error.svg`, `/tmp/r2_search.svg` (capture script committed at `scripts/webfetch_r2_shot.py`).

**Verdict: TERMINAL.** No BLOCKER, no MAJOR — D1 resolved, D2/D3/D4 resolved, F1 error card clean and unmistakable across all states, grammar consistent with web_search. D5 (search-body preview quality, round 1) remains a recorded follow-up, not a blocker. The design round is clean on the current head `6adde039` and can close.

damianvtran:
### Agent review — round 2
Reviewer: reviewer (independent agent — Claude Opus 4; NOT the author — separate agent on the shared account)
Scope: f8bc63adb6d31960ee8ca3b465c2313258f7d48b..6adde039f486ac899bdb5f39ac80aa1b1c0a6074 (base..head; round-1 fixes landed in e5446b79, 0bdbe6b6, 6adde039)

Re-reviewed at head **6adde039** on branch `dev-web-fetch`. Read the round-1 code review + remediation, then VERIFIED each disposition against the code at head (not the remediation's word for it) and RAN the suite + gates. Verdict up front: **all four round-1 MAJORs and both minors are genuinely resolved, plus the F1 dogfooding fix; no new blocker/major findings.**

#### Round-1 findings — verification against head

**M1 (DNS-rebinding / TOCTOU) — RESOLVED.** The connection is now pinned to the pre-validated IP, correctly and completely:
- `validate_public_url` (`service.py:269-313`) resolves the hostname ONCE via `_resolve_host_ips`, judges EVERY returned address (`_ip_is_forbidden` on the whole set — a mixed public/private record is refused outright), and RETURNS the vetted IP literal instead of only raising.
- `_pin_request` (`service.py:183-212`) rewrites the request URL host to that IP via `httpx.URL.copy_with(host=ip)`, keeps the original hostname in the `Host` header AND in the `sni_hostname` extension. `_stream_once` (`service.py:576-616`) passes both to `client.stream`, so (a) the socket lands on the validated IP — httpx never re-resolves the name at connect time — and (b) TLS SNI/cert verification stays against the real hostname. I grepped the whole package: **there is no `verify=False` anywhere** — verification is never disabled, only the target IP is redirected.
- Applied on EVERY hop: `_follow` (`service.py:501-530`) calls `validate_public_url` + pins inside the redirect loop, so a 302 to a rebinding host is re-validated and re-pinned independently. Enrichment probes ride the same `_follow`, so no side-door bypass.
- Tests prove it: `test_connection_pinned_to_vetted_ip` asserts the captured request host == the vetted IP while Host/SNI keep the hostname; `test_rebinding_second_resolution_never_used` supplies a second (private) resolution and asserts it is never reached. I tried to find a residual bypass — the vetted IP is what the socket connects to (not a re-resolved name), and each redirect Location is both re-validated AND re-pinned. No bypass remains.

**M2 (`/llms.txt` substituted for a subpage) — RESOLVED.** `_enrichment_candidates` (`service.py:675-711`) now appends `/llms.txt` ONLY when `parsed.path in ("", "/")` (site root); the path-scoped `.md` twin stays for concrete file-ish paths. `test_llms_txt_not_substituted_for_subpage` fetches `/docs/guide` where `guide.md` 404s but a substantial `/llms.txt` exists and asserts the guide's own HTML renders (final_url unchanged); `test_llms_txt_used_for_site_root` confirms the legitimate root case. Both pass.

**M3 (cache key ignores raw/max_bytes) — RESOLVED.** `cache_variant(raw=, max_bytes=)` (`service.py:341-351`) is folded into `_url_hash` on BOTH read (`tool.py:334-339`) and write (`tool.py:397-411`, passing `variant`). A raw and a rendered fetch of the same URL hash to different sidecars. `test_raw_and_rendered_do_not_share_cache_entry` passes.

**M4 + F1 (non-2xx cached / rendered as content) — RESOLVED, and the two agree.** M4: the cache write is gated `cacheable = http_ok and render_method != "binary"` with `http_ok = 200 <= status < 300` (`tool.py:371,396-397`) — a 4xx/5xx is never stored, so the next request re-hits the network (`test_non_2xx_not_cached_and_retried`). F1: the return path is honest — `run_fetch` returns `is_error = not http_ok` (`tool.py:417`), `details` carry explicit `ok`/`http_error` booleans (`tool.py:382-383`), and `_header_line` (`tool.py:156-181`) leads a non-2xx preview with a bold `⚠ HTTP <status> <reason> — this is an error/block page, not page content.` line plus a "(body below is the error response)" note, so a block page is structurally distinct from content. The cache-hit path (`tool.py:255-273`) also carries the flags and — since only 2xx is ever cached — always returns `is_error=False`, so cached and fresh results branch identically. The two fixes are complementary and consistent. `test_non_2xx_surfaced_as_error_not_content` (parametrised 403/404) passes.

**m1 (`_is_binary` vs docstring) — RESOLVED.** `application/octet-stream` is no longer in the hard-binary set (`service.py:731-756`); it falls through to the NUL-byte sniff, matching the docstring. PDF/zip stay hard-binary. `test_octet_stream_text_body_renders` + `test_octet_stream_binary_body_is_notice` pass.

**m2 (enrichment probe cost) — ADDRESSED (documented).** `_try_enrichment` (`service.py:544-551`) now documents the bounded two-probe cost, the short-circuit on the first substantial candidate, the post-M2 path-scoped candidate set, and that every 2xx result is cached so the cost is once per URL+variant. Reasonable disposition for a minor.

#### Gates (run from the worktree at head 6adde039)
- `pytest tests/unit/web_fetch -q` → **69 passed**. M1–M4/F1 spot-run (`-k "pin or rebind or llms or cache or non_2xx or octet or raw_and_rendered or error"`) → 19 passed.
- `pytest tests/unit/tui/test_tool_card.py` → 130 passed.
- flake8 → clean; **black==26.1.0 --check** (the AGENTS.md pin) → clean; isort --profile black → clean; `pyright --pythonpath .venv/bin/python local_operator/web_fetch/` → **0 errors**.
- Broad `pytest tests/unit` → 6404 passed, 4 failed — all four are the two known-preexisting categories (`test_server_models` `State.env_config`, x3; `test_eval_tool::test_background_streams_output_while_it_runs` timing flake). I reproduced both on a clean `origin/main` worktree (they fail there too, env_config failing worse in isolation). Neither file is touched by this PR's diff. **Zero new failures introduced.**
- CI at head: lint / type-check / pip-audit green; test (3.12/3.13) pending at review time.

#### New findings
None. No blocker, no major, no new minor worth a round.

**Verdict: clean — no blocker/major findings remain — round clean against head 6adde039.** This round is TERMINAL. The deferred items (D5 verbose spill footer, n1/n2 nits) are follow-ups, not a reason for another round.

PR #273, opened 2026-08-23
fix(usage): announce quota state once per transition, not every message

damianvtran:
## Why

Every user-message boundary runs an OAuth quota preflight (`SessionStreamFn.preflight_usage`), and its notices were emitted **unconditionally on each boundary**. A user whose provider sat at low/exhausted quota saw the same lines echoed on **every single message they sent**:

```
! anthropic quota low (9% remaining) — trying another anthropic account before provider fallback
! anthropic quota exhausted (0% remaining) — trying another anthropic account before provider fallback
! anthropic quota exhausted (0% remaining) — trying another anthropic account before provider fallback
```

Two separate defects:
1. **No dedup across boundaries** — a *persistent* quota state re-announced on every message, though the user only needs to hear about a **change** in standing.
2. **Per-rotation churn** — the account-rotation loop emitted "trying another … account before provider fallback" **once per sibling account** it silently rotated through, producing the duplicated lines. Rotating to another same-provider account is an internal detail; the request is still served on the same provider.

Reported by the operator: quota-low noise on every message, redundant duplicate lines.

Version bump is **PATCH** (0.25.0 → **0.25.1**): behavioural fix, no new capability, no breaking change.

## What changes

`local_operator/model/configure.py` only.

### Announce once per state transition
A per-session latch `self._last_quota_state: dict[str, str]` keyed by `provider/model_id` remembers the last `health.state` actually announced (`reserve` = low, `depleted` = exhausted). A new `_announce_quota_change(selector, state, text, kind)` helper routes all four surviving quota notices and speaks **only when the state differs** from what was last announced for that selector. A steady low/exhausted verdict is now silent after the first line; a genuine transition (none→low, low→exhausted, or a recurrence after recovery) still speaks once.

### Latch cleared on the healthy edge
`_clear_quota_latch(selector)` is called on **both** healthy edges — the boundary probe recovering and `_recover_blocked_accounts` reviving an account — so a later slide back into low/exhausted counts as a real new transition and re-announces. The `unknown` fail-open branch leaves the latch untouched (indeterminate is not a transition).

### Rotation churn suppressed
Both "trying another … account before provider fallback" notices (tier-scope in `preflight_usage`, account-scope in `_apply_account_health`) are removed — the rotation now `continue`s silently. **Block/rotation/selection logic is untouched**; only the transcript notice is dropped.

Fallback-activation notices (`_on_route_change` / `RouteState.activate`) are deliberately left alone — a fallback firing is a real event worth one line.

## Impact
[...]

PR #277, opened 2026-08-23
fix(usage): route preflight quota probes through the shared cache to end the per-process 429 storm

damianvtran:
## Why

The follow-up deferred from #274. The usage **preflight** — the per-message-boundary quota check that decides routing — calls `usage.fetch_usage` **raw** at four sites in `SessionStreamFn` (`configure.py`), bypassing the shared cross-process usage cache that the `/usage` panel and its background warmer already use. So every user-message boundary, in **every** concurrent `lop` process, fetches every account's usage fresh from Anthropic's `/api/oauth/usage` — an endpoint that rate-limits **per source IP**. N sessions × M accounts × every boundary is how that endpoint earns a 429 of its own, which was a contributing factor in the brittleness investigated in #274 (the usage check itself getting throttled, then feeding block/deprioritize decisions).

The four raw sites:
- `_provider_quota_availability` — the per-OAuth-account enumeration loop, and the selected-API-key probe.
- `preflight_usage` — the reactive per-boundary primary probe.
- `_recover_blocked_accounts` — the blocked-account recovery walk.

Version bump is **PATCH** (0.25.4 → **0.25.5**): a fix, no new user-facing feature, no schema change.

### A routing-shaped read-through, not a display cache — `providers/usage_cache.py`
New `leased_account_usage(store, key, provider, fetch)` plus `account_preflight_key(storage_id, account_identity)`.

The key insight: **the fan-out worth collapsing is cross-*process* concurrency, not one process's sequential boundaries.** A single process re-probing on its next boundary is *intended* — that is how routing notices a quota recovering or depleting (and the suite enforces it: `test_quota_pinned_fallback_is_reprobed_at_the_next_boundary` mutates the report between two boundaries and asserts the second sees it). So this helper is deliberately **not** the warmer's read-through:

- It **never serves a fresh cached value on the fast path** — it fetches live every boundary, so per-boundary re-probing is untouched.
- It takes `try_lease` **only to divert a concurrent peer**: when another process already holds the in-flight lease for this account, this process serves the peer's last-good value instead of crossing the network for the identical answer. That single difference from the warmer's `force_refresh` path (which skips the lease) is the whole design.
- It **fails open**: no cache, or a cache that can't open, means live fetch — routing can never be made *worse* than the pre-cache behaviour. `fetch` returning `None` (transport failure *or* no-quota) is passed through unchanged, preserving the preflight's existing `saw_unknown` fail-open.

The preflight uses a **deliberately separate `:pf:` key namespace**, disjoint from the warmer's per-provider-**set** rows. Reasons: the shapes differ (one account vs the full set); `UsageReport.identity` is only backfilled on the controller's fetch path, so the warmer's set-row can't be reliably sliced per account on the raw preflight path; and the two want different freshness policies. Accepted cost: a warmer refresh and a preflight probe for the same account may each cross the network once — fine, the warmer was never the storm.

### Wire the preflight through it — `model/configure.py`
`SessionStreamFn` gains a lazily-built `_usage_cache` (a cache that can't open is a permanent live-fetch miss, never an error), a thin `_cached_account_usage(provider, account_identity, fetch)` wrapper, and a `close()` that releases the handle. Each of the four sites is rewritten to call it with the account's identity and an immediate closure over the still-**function-local** `fetch_usage`. Keeping that import function-local (and the closures immediate, default-binding loop vars) is what preserves the tests' patch point — `usage_cache`'s helper never references `fetch_usage`; it takes the fetcher as a callable.

The blocked-account recovery walk's per-row termination guarantee is untouched: it records `attempted_ids` **before** the probe, independent of whether the network was crossed.

## Impact
- No breaking change, no schema change. `controller.py` (the warmer's read-through) is deliberately left alone — its fresh-within-TTL, per-provider-set, empty-over-data semantics are correct for a *display* panel and wrong for a *routing* probe.
- **Watched behaviour change:** a lease-loser now acts on last-good data (≤ `USAGE_FETCH_LEASE_MS` = 30s + one boundary old) instead of failing open on a self-inflicted 429. Strictly better than the storm, but the recovery-latency ceiling is on the record.

### Fan-out collapse — real cross-process proof (two stores, one on-disk cache db)
```
preflight key: anthropic:pf:c7d735069d2a1faa
  contains ':pf:':          True
  disjoint from warmer key: True

seed fetch crossed network: 1  (expect 1)
peer B while A holds lease — network calls added: 0  (expect 0: served stale)
peer B still got a usable report: True
```
Two independent `UsageCacheStore` handles on one db = two processes. While process A holds the account's fetch lease, process B's `leased_account_usage` adds **zero** network calls (serves last-good) yet still returns a usable report. This is the storm collapse, demonstrated end to end.

### Unit tests — `tests/unit/providers/test_usage_cache.py` (+6)
Fan-out collapse; free-lease fetches/writes/releases; per-account isolation (two accounts → disjoint `:pf:` keys, each returns its own report, both differ from the warmer's `provider_cache_key`); lease-loser-with-no-stale fetches live; fail-open on `store=None` (incl. `None` result); `None`-result preserves last-good and fails open.

### Integration — `tests/unit/model/test_configure.py` (+1)
Two `SessionStreamFn` on one `HOME`/db: A holds the account lease, and B's `preflight_usage` reaches its verdict off the stale healthy row **without the patched fetcher being called** for that account (`calls == []`). Two-sided: the seeded fetcher would have returned a depleted 99% report that flips the verdict, so a wrong wiring would fail loudly.

### The ~30 existing preflight tests stayed green with ZERO changes
This is the regression proof. Every `test_configure.py` test that reaches a preflight fetch patches `local_operator.providers.usage.fetch_usage` and runs single-process under the autouse isolated-`HOME` fixture → a fresh empty cache db → `try_lease` always free → live fetch every time → the patched fetcher is called exactly as often as before. The two count-sensitive assertions (`assert_not_called` / `assert_called`) both still hold.

### Gates (whole tree, rebased onto current `origin/main` @ 0.25.4 — which included #273's heavy `preflight_usage` rewrite)
- `flake8` ✓ · `black==26.1.0 --check` ✓ · `isort --profile black` ✓ · `pyright` **0 errors, 0 warnings**
- `test_configure.py` + `test_usage_cache.py`: **121 passed**. Full `tests/unit/providers tests/unit/model`: **1012 passed**.

🤖 Generated with local-operator (lopdev team)

### DIFF
--- a/local_operator/model/configure.py
+++ b/local_operator/model/configure.py
@@ -63,6 +63,8 @@ if TYPE_CHECKING:
     from local_operator.env import EnvConfig
     from local_operator.providers.auth_store import AuthStore
     from local_operator.providers.clients import WireClient
+    from local_operator.providers.usage import UsageReport
+    from local_operator.providers.usage_cache import UsageCacheStore
 
     #: Either provider's ``list_models()`` payload. The two schemas are
     #: structurally identical (``data`` of items with ``id``/``description``/
@@ -1564,6 +1566,14 @@ class SessionStreamFn:
         # healthy edge still resets all of them together (recovery is a fresh
         # start for every condition on that selector).
         self._last_quota_state: dict[str, set[str]] = {}
+        # Shared cross-process usage cache for preflight probes, built lazily on
+        # the first routing check and reused for the session's lifetime. Its
+        # sole job here is to collapse a concurrent PEER process's duplicate
+        # fetch of the same account (the per-source-IP 429 storm this fixes) —
+        # it never suppresses this process's own per-boundary re-probe. None
+        # until first use, and stays None whenever the cache cannot open (a
+        # permanent live-fetch miss, never an error). See ``_usage_cache_store``.
+        self._usage_cache: "UsageCacheStore | None" = None
 
     def _client_for(self, spec: ModelSpec) -> WireClient:
         from local_operator.providers.clients import client_for_spec
@@ -1883,6 +1893,7 @@ class SessionStreamFn:
             usage_health,
             usage_supported,
         )
+        from local_operator.providers.usage_cache import fingerprint_secret
 
         if not usage_supported(provider):
             cache[cache_key] = "unknown"
@@ -1908,12 +1919,16 @@ class SessionStreamFn:
 
         for access in accesses:
             try:
-                report = await fetch_usage(
-                    self._http,
+                report = await self._cached_account_usage(
                     provider,
-                    access_token=access.access_token,
-                    account_id=access.account_id,
-                    oauth_creds=access.raw,
+                    access.email or access.account_id or f"cred:{access.credential_id}",
+                    lambda a=access: fetch_usage(
+                        self._http,
+                        provider,
+                        access_token=a.access_token,
+                        account_id=a.account_id,
+                        oauth_creds=a.raw,
+                    ),
                 )
             except Exception:
                 saw_unknown = True
@@ -1943,7 +1958,11 @@ class SessionStreamFn:
             saw_unknown = True
         if selected is not None and selected.kind == "api_key":
             try:
-                report = await fetch_usage(self._http, provider, api_key=selected.access_token)
+                report = await self._cached_account_usage(
+                    provider,
+                    fingerprint_secret(selected.access_token),
+                    lambda: fetch_usage(self._http, provider, api_key=selected.access_token),
+                )
             except Exception:
                 report = None
             if report is None:
@@ -2040,6 +2059,42 @@ class SessionStreamFn:
         from local_operator.providers.registry import credential_provider_id
 
         return credential_provider_id(provider)
+
+    def _usage_cache_store(self) -> "UsageCacheStore | None":
+        """The shared usage cache, built lazily. A cache that cannot open is a
+        permanent miss (live fetch), never an error — same contract as the warmer's."""
+        if self._usage_cache is None:
+            try:
+                from local_operator.providers.usage_cache import UsageCacheStore
+
+                self._usage_cache = UsageCacheStore()
+            except Exception:  # noqa: BLE001 — no cache = live fetch, never fatal
+                return None
+        return self._usage_cache
+
+    async def _cached_account_usage(
+        self,
+        provider: str,
+        account_identity: str,
+        fetch: "Callable[[], Awaitable[UsageReport | None]]",
+    ) -> "UsageReport | None":
+        """Route one preflight usage probe through the shared cross-process cache.
+
+        Collapses a concurrent peer's duplicate fetch of the SAME account (the 429
+        storm this fixes) while keeping every boundary free to re-probe live —
+        ``leased_account_usage`` never serves a fresh row on the fast path. Fails open
+        to a live fetch when the cache is unavailable, so routing can never be made
+        WORSE than the pre-cache behaviour. See docs/specs/preflight-usage-cache.md.
+        """
+        from local_operator.providers.usage_cache import (
+            account_preflight_key,
+            leased_account_usage,
+        )
+
+        storage = self._storage_provider(provider)
+        store = self._usage_cache_store()
+        key = account_preflight_key(storage, account_identity)
+        return await leased_account_usage(store, key, storage, fetch)
 
     async def _primary_has_auth(self, model: ModelSpec) -> bool:
         from local_operator.providers.failover import FallbackTarget
@@ -2216,11 +2271,15 @@ class SessionStreamFn:
                 usage_health,
             )
 
-            report = await fetch_usage(
-                self._http,
+            report = await self._cached_account_usage(
                 model.provider,
-                access_token=access.access_token,
-                account_id=access.account_id,
+                access.email or access.account_id or f"cred:{access.credential_id}",
+                lambda a=access: fetch_usage(
+                    self._http,
+                    model.provider,
+                    access_token=a.access_token,
+                    account_id=a.account_id,
+                ),
             )
             if report is None:
                 return
@@ -2744,11 +2803,15 @@ class SessionStreamFn:
             token = key_fn(creds) if key_fn else creds.get("access")
             if not token:
                 continue
-            report = await fetch_usage(
-                self._http,
+            report = await self._cached_account_usage(
                 model.provider,
-                access_token=token,
-                account_id=creds.get("account_id"),
+                creds.get("email") or creds.get("account_id") or f"cred:{row.id}",
+                lambda t=token, c=creds: fetch_usage(
+                    self._http,
+                    model.provider,
+                    access_token=t,
+                    account_id=c.get("account_id"),
+                ),
             )
             if report is None:
                 continue  # unreachable quota endpoint: keep the block, move on
@@ -3010,6 +3073,13 @@ class SessionStreamFn:
 
     async def close(self) -> None:
         await self._http.aclose()
+        if self._usage_cache is not None:
+            try:
+                self._usage_cache.close()
+            except Exception:  # noqa: BLE001 — teardown, never fatal
+                self._usage_cache = None
+            else:
+                self._usage_cache = None
 
 
 def create_stream_fn(
--- a/local_operator/providers/usage_cache.py
+++ b/local_operator/providers/usage_cache.py
@@ -506,9 +506,6 @@ async def leased_account_usage(
     if store is None:
         return await fetch()
 
-    stale = store.get(key, include_expired=True)
-    stale_report = stale[0] if stale else None
-
     # The lease is contended only when a peer is mid-fetch. A free lease (the common
     # case, and ALWAYS the case in a single-process test) means: this process fetches.
     if not store.try_lease(key):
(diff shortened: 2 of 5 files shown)


damianvtran:
### Agent review — round 1

Reviewer: reviewer subagent (independent of the implementer), model anthropic/claude-opus-4-8 (Claude Opus 4).
Scope: origin/main..ca75c9bc24a1640f2c805775cd7ef212b6711b5c

I read the diff, traced the helper against the post-#273 `preflight_usage`, and ran the full gate from the worktree (`/tmp/lop-preflight-cache`). All green: flake8 (0), `black==26.1.0 --check` (0), `isort --profile black` (0), pyright on the two changed modules (0 errors/warnings), and pytest — `tests/unit/model/test_configure.py tests/unit/providers/test_usage_cache.py` = **121 passed**, broader `tests/unit/providers tests/unit/model` = **1012 passed**. The two test files are additions-only (0 deletions in each diff), so the ~30 existing preflight tests are unchanged and still patch `local_operator.providers.usage.fetch_usage`.

**Core correctness — verified.** `leased_account_usage` never serves a fresh cached value on the fast path. Fast path = a FREE lease (`try_lease` True), and that branch unconditionally `await fetch()`s, writes, releases — no freshness check consulted, so the per-boundary re-probe (and the @2296-equivalent pinned-reprobe test) is preserved. A cached row is served ONLY on the contended branch (`try_lease` False), i.e. when a peer is actively doing the live fetch — serving its value is a deduplicated live probe, not a short-circuit. Cold-start lease-loser with no row fetches live. Confirmed no path where a fresh row bypasses the fetch for the lease-winner.

**Lease discipline — verified.** `try_lease` True is always paired with `release_lease` in a `finally` (usage_cache.py:521-537). The lease-loser branch never releases (returns before the try). `release_lease` deletes only `WHERE holder = self._holder` (usage_cache.py:477), and `_holder` is per-store-instance/per-process (`{pid}:{uuid}`, usage_cache.py:228), so a loser can never free a peer's lease.

**Fail-open — verified.** `store is None` → live fetch (usage_cache.py:506); `_usage_cache_store()` returns None on any open failure → live fetch; `fetch()` returning None is passed through unchanged so the callers' `saw_unknown`/`report is None` fail-open still fires. Store methods are internally exception-safe (`try_lease` returns True on error = go fetch), so no cache exception escapes to break routing. A `fetch()` exception still runs the `finally` (lease released) and propagates to each site's existing `try/except`, matching pre-cache behaviour.

**Namespace — verified disjoint.** `account_preflight_key` = `{storage}:pf:{fingerprint_accounts([id])}`; warmer's `provider_cache_key` = `{provider}:{fingerprint}`. `fingerprint_accounts` yields 16 hex chars (or `none`), which can never start with `pf:` (`p` is not hex), so the two are disjoint even for a single-account provider. `account_identity` is hashed again by `fingerprint_accounts`, and the API-key route passes `fingerprint_secret(...)` (a `key:<sha>` digest), so no raw key/token ever reaches the row key. Cross-site identity is consistent (`email or account_id or cred:<id>`, and `access.credential_id == row.id`), so the same account dedups across all four sites and both processes.

**Patch-point preservation — verified.** All four `fetch_usage` references are inside default-bound closures (`lambda a=access:` / `lambda t=token, c=creds:`) at configure.py:1925/1964/2277/2809, and every `from ...usage import fetch_usage` is function-local (1892/2269/2772); the only module-scope import is TYPE_CHECKING (line 66). No pre-bound reference routes the existing tests to the network.

**Recovery walk (site ~2806) — verified.** `attempted_ids.add(row.id)` at configure.py:2785 still precedes the probe, preserving the termination guarantee; the cache wraps only the fetch and does not alter the per-row loop. Ping-pong-guard tests pass.

**#273 integration — checked, no defect.** `_last_quota_state` latching is per-process; a lease-loser announces off its own latch using the peer's last-good verdict. No cross-process announcement coupling, so no spurious/missed-announcement bug beyond the one-boundary staleness the fan-out collapse deliberately accepts and documents.

**Tests — real.** The lease-contention tests use two `UsageCacheStore` instances on one db file (two distinct holders = genuine cross-process contention), and the `test_configure` integration test drives the real `preflight_usage` with a peer holding the lease and asserts the patched fetcher is never called for that account. Not single-store theatre.

Findings:

- **blocker:** none.
- **major:** none.
- **minor (1):** usage_cache.py:509 — `store.get(key, include_expired=True)` runs on EVERY call, but `stale_report` is only consumed on the contended (`try_lease` False) branch. On the common free-lease path (the single-process norm, once per account per boundary) this is a wasted SQLite read. Moving the `get` inside the `if not store.try_lease(key):` branch is behaviour-identical (and reads a marginally fresher row from the peer). Follow-up, not stop-ship.
- **minor (2):** Only the reactive enumeration site is covered by an end-to-end `test_configure` integration test; sites 3 (~2274) and 4 (recovery walk, ~2806) rely on the helper's unit tests plus uniform wiring. Coverage is adequate given the shared helper, but an integration probe on the recovery walk would close the loop.
- **nit:** none worth a round.

Verdict: **clean — ship.** No blocker and no major; the round is TERMINAL. The two minors are follow-ups, not a reason for another round.

damianvtran:

Implementer: coder subagent, model anthropic/claude-opus-4-8 (Claude Opus 4).
Scope: f8bf877e4eea837cd8673c34bedb944b09760913..5328f09a304d876e3e64a869d36da6c7e3efcd8f (current head, includes the round-1 fix commit).

**minor (1) — FIXED** in `5328f09a304d876e3e64a869d36da6c7e3efcd8f`.
`leased_account_usage` read the last-good row (`store.get(key, include_expired=True)`) on every call, but `stale_report` is consumed only on the contended (`try_lease` False) branch, so the free-lease fast path — the single-process norm, once per account per boundary — spent one SQLite query per probe for a value it never used. Moved the `get` inside the `if not store.try_lease(key):` branch. Behaviour-identical as the reviewer notes, and it reads a marginally fresher row (any write the lease holder landed between entry and the lease check). Cold-start on the contended branch (no row present) still fetches live — the "cannot serve nothing" case is preserved. Re-ran the two suites (`tests/unit/providers/test_usage_cache.py tests/unit/model/test_configure.py` = **121 passed**) and all four gates (flake8 0, `black==26.1.0 --check` 0, `isort --profile black` 0, pyright 0 errors/warnings) — still green.

**minor (2) — ACCEPTED / DEFERRED.**
The coverage observation is correct: only the reactive enumeration site has an end-to-end `test_configure` integration test; the selected-API-key site and the recovery walk rely on `leased_account_usage`'s own cross-store unit tests plus the uniform `_cached_account_usage` wiring (identical closure shape at all four sites). As the review itself concludes, "coverage is adequate given the shared helper" — the helper is the single point where the lease/stale/fail-open semantics live, and it is directly tested for fan-out collapse, cold-start-fetches-live, fail-open, and None-result. A recovery-walk integration probe would close the loop but is not required to ship; deferring it as a follow-up rather than expanding this PR's scope.

**Round status.** 0 blockers, 0 majors remain. Both minors are resolved (one fixed, one accepted with reasoning). Round 1 is answered against the current head `5328f09a304d876e3e64a869d36da6c7e3efcd8f`.

damianvtran:
### Agent review — round 2

Reviewer: reviewer (lopdev agent team), anthropic/claude-opus-4
Scope: ca75c9bc..5328f09a (delta since round 1; full head is 5328f09a)

Focused freshness confirmation of the round-1 minor(1) remediation. Round 1 (clean, 0/0) was scoped to ca75c9bc; this round re-blesses the single follow-up commit against the current head.

**Delta verified.** The only change since ca75c9bc is the pre-blessed get-relocation in `leased_account_usage` (`local_operator/providers/usage_cache.py`): `store.get(key, include_expired=True)` / `stale_report` moved out of the pre-lease path and into the `if not store.try_lease(key):` contended branch, with a comment explaining the why (the free-lease fast path never consumes `stale_report`, so the up-front SQLite query was wasted on the overwhelming majority of calls; deferring also reads a marginally fresher row). Nothing else changed.

**Correctness of the fix — confirmed, no new finding:**
- Free-lease fast path (the common case) still fetches live — no fresh-row short-circuit was introduced; it falls straight through to `await fetch()` under the held lease.
- Contended cold start (no row) still fetches live: `stale_report is None` ⇒ `await fetch()`, preserving the "lease only protects a stale value" rule.
- Lease discipline unchanged: `try_lease` gate, `set` on success, `write_failure` + `None` passthrough on failure, `release_lease` in `finally` — all intact.

**Gates (re-run from the worktree on head 5328f09a):**
- `flake8 .` — clean
- `black --check` (26.1.0) — 455 files unchanged
- `isort --check-only --profile black` — clean
- `pytest tests/unit/model/test_configure.py tests/unit/providers/test_usage_cache.py` — 121 passed

**Verdict: clean — 0 blockers / 0 majors.** Round 1's minor(1) fix is correct and introduces no new findings. Round is FRESH and TERMINAL against head 5328f09a.

damianvtran:
### Agent review remediation — rebase freshness

Head moved `5328f09a` → `19b72c54` after a rebase onto `origin/main` (which had advanced by #276 `web_fetch`, a minor bump to 0.26.0). Our version is now **0.26.1**.

The reviewed *content* is unchanged: `git diff 5328f09a 19b72c54 -- local_operator/providers/usage_cache.py local_operator/model/configure.py tests/unit/providers/test_usage_cache.py tests/unit/model/test_configure.py` is empty. The only delta in our files is `pyproject.toml` version `0.25.5` → `0.26.1`. Round 1 + 2 findings therefore still apply to the current head; no new reviewable change was introduced.

PR #279, opened 2026-08-23
feat(tui): title the session theme, not the newest message

damianvtran:
## Why

A session whose goal was **build a web fetch tool** got auto-renamed to **"Find Port Credit restaurants"** the moment the user exercised the tool, then again to **"Include Findings in Release"** on the next follow-up. Each pivot compounded because the next check anchored on the already-drifted title. The title should have stayed on the session's *theme* — work on a web fetch tool — so a glance at the tab or `/resume` picker still identifies the conversation.

Root cause, verified against the live path on `origin/main`:

- `generate_retitle` sent the model **only the current title + the single newest user message** (`_errand_prompt`, 320 chars) and asked "did the SUBJECT or GOALS materially change?"
- An in-goal step ("test the tool by finding Port Credit restaurants") looks, in isolation, like a brand-new subject.
- The throttle was **wall-time only** (`RETITLE_MIN_GAP_S = 120s`), so a long session kept re-titling indefinitely instead of settling.

This ports the omp fork's proven `feat/session-theme-titling` design into lop's Python idiom. Version bump **0.26.0 → 0.27.0** (minor — 0.26.0 was already taken by `web_fetch` #276).

## What changes

Three parts, none of which works alone:

1. **Whole-trajectory context** (`naming.build_theme_context`) — the re-title call now sees a sampled `<chat>` of the opening turns (which *state* the subject) plus a recent tail, with `<elided/>` marking the gap. Bounded per turn so a pasted log cannot blow the cheap call.
2. **Current-title anchor + theme prompt** (`THEME_SYSTEM_PROMPT`) — "Title the whole body of work, not the most recent message. Repeat `<current-title>` verbatim unless the work has moved to a different subject. A new step, file, question, or tool call inside the same body of work is NOT a different subject."
3. **Growth-gated hysteresis** (`naming.should_refresh_theme`) — replace the 120s-only throttle with transcript-growth doubling (eligible at turn ≥6, ≥16, ≥36, ≥76, ≥156) capped at five refreshes. A 30s time floor stays as a secondary burst guard. Counters live on the TUI host (not `ConversationName` — that holder is persisted/shared).

Preserved: `user_set` (`/rename`) outranks everything forever and spends no call; low-signal messages spend nothing; the call stays isolated / single-attempt and swallows all failures to `None`; `<title/>` means no change; the provisional opener path is unchanged.

### Behavioural — the reported sequence, old vs new

Same opener, same in-goal follow-up, same model policy ("if you only see the newest message, title it; if you see the opener, keep the theme"):

```
=== OLD prompt (newest message only) ===
now find me some good Port Credit restaurants to try it out
OLD title produced: 'Find Port Credit restaurants'

=== NEW prompt (theme context) ===
<chat>
<current-title>
Build a web fetch tool
</current-title>

<user>
help me build a web fetch tool for the agent
</user>

<assistant>
I'll add a fetch tool that GETs a URL.
</assistant>

<user>
wire it into the tool registry
</user>

<assistant>
Done — registered and callable.
</assistant>

<user>
now find me some good Port Credit restaurants to try it out
</user>
</chat>
NEW title produced: None  (None = keep current theme)
```

Growth schedule (unit-checked): titled at turn 1, next refresh at ≥6, then ≥16, ≥36, ≥76, ≥156, then never.

### Automated

`tests/unit/tui/test_conversation_naming.py` — 37 passed, including:

- `test_an_in_goal_step_does_not_pivot_the_title` — the reported regression.
- `test_build_theme_context_samples_head_and_tail_with_an_elision_marker`
- `test_should_refresh_theme_follows_the_geometric_schedule`
- `test_the_refresh_budget_stops_re_titling_after_the_cap`
- `test_the_time_floor_defends_against_a_same_breath_burst`
- All prior invariants: human `/rename` never overwritten, low-signal spends nothing, sentinel = no change, restatement = no change, provider failure swallowed.

```
$ env -u NO_COLOR TERM=xterm-256color .venv/bin/python -m pytest tests/unit/tui/test_conversation_naming.py -q
37 passed in 4.53s
```

### Gates

```
flake8 / black==26.1.0 / isort --profile black / pyright  — clean (0 errors)
```

A first full `tests/unit/tui tests/unit/session` run showed 3 pre-existing flakes (theme-color snapshot `#1e1a14` vs `#14110c`, composing-row label timing) that passed on re-run and fail identically on clean `origin/main`. Not this change.

## Review history
_Agent review gate in progress — code round to follow on this PR._

### DIFF
--- a/local_operator/session/naming.py
+++ b/local_operator/session/naming.py
@@ -26,12 +26,22 @@ because the obvious implementation gets it wrong:
   :func:`provisional_title` names the conversation from the opener the instant
   it is submitted — no network at all — and the model's title now lands a
   second or two later, concurrently with the turn, rather than after it.
-- **A conversation drifts, so a title has to be allowed to.**
-  :func:`generate_retitle` gives the model the current title and a new message
-  and lets IT judge whether the subject or the goals have materially moved.
-  That judgement is not expressible as a keyword rule, and a title still
-  naming the first of five subjects misidentifies the session rather than
-  merely under-describing it.
+- **A conversation drifts, so a title has to be allowed to — toward the THEME,
+  not the newest message.** :func:`generate_retitle` gives the model a sampled
+  ``<chat>`` of the whole trajectory (the opening turns, which state the
+  subject, plus a recent tail) anchored on the CURRENT title, and lets IT judge
+  whether the body of work has genuinely moved on. That judgement is not
+  expressible as a keyword rule. The earlier design showed the model only the
+  single newest message, so an IN-GOAL step read as a brand-new subject: a
+  session building a web-fetch tool got renamed to "Find Port Credit
+  restaurants" the moment the user exercised the tool, then again on the next
+  follow-up, each pivot compounding because the next check anchored on the
+  already-drifted title. Titling the whole trajectory against the current-title
+  anchor is what keeps a drifting session named after what it is actually
+  about. The three parts that make this work — whole-trajectory context, the
+  current-title anchor, and the caller's growth-gated refresh schedule — each
+  fail alone; see :func:`build_theme_context`, :data:`THEME_SYSTEM_PROMPT`,
+  and :func:`should_refresh_theme`.
 
 The holder (:class:`ConversationName`) mirrors ``GoalState``: a small mutable
 object the session and its host share, so a name that lands asynchronously is
@@ -46,6 +56,18 @@ from __future__ import annotations
 import asyncio
 import re
 from dataclasses import dataclass
+from typing import Any, Sequence
+
+#: A history entry as the theme sampler reads it. Deliberately ``Any`` and
+#: duck-typed via ``getattr`` rather than a ``Protocol`` or an import of the
+#: harness ``AgentMessage`` union: the sampler only ever needs ``.role`` and
+#: ``.text``, and a ``CustomMessage`` in that union legitimately carries no
+#: ``role`` at all (it is filtered out here, not type-excluded). Typing this as
+#: the real union would either drag a harness dependency into what is meant to
+#: be a leaf module or fail to describe the role-less custom entries; reading
+#: the two attributes defensively is both honest and keeps naming standalone.
+_Turn = Any
+
 
 #: Hard caps on a stored title. Both are enforced as REJECTION, not
 #: truncation: see the module docstring.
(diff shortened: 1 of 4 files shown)


damianvtran:
### Agent review — round 1
Scope: 4ba0334e8a27b6828749d6c072a1d630b7bbf35e..3a950a3d26ca9bb192a1a168501c0a23944ba76c

No blocker or major findings. The three parts (head+tail sample, current-title-anchored theme prompt, growth-gated hysteresis) hang together, the old `generate_retitle` signature is gone from production (only the TUI worker calls it, and it passes `turns=`), `user_set` still short-circuits before any call, and `_ask_for_title` still collapses every failure to `None`. 37/37 naming tests passed locally; lint/black/isort/pyright clean on the touched files.

#### Blocker
None.

#### Major
None.

#### Minor

**m1.** `test_an_in_goal_step_does_not_pivot_the_title` does not actually prove the opener reached the model. The fake keeps the theme when `"web fetch" in prompt.lower()`, but that string is already in `<current-title>Build a web fetch tool</current-title>` — the test would still pass if `build_theme_context` dropped every opening turn. The follow-up `assert "web fetch tool" in seen[0]` is the same tautology. Assert a phrase that exists only on the opener (e.g. `"for the agent"`) so a head-dropping sampler would fail this test rather than only `test_build_theme_context_samples_head_and_tail_with_an_elision_marker`.

**m2.** Growth-gate counters are host-local and reset to `0` on `/resume`, `--resume`, and every swap (`app.py` ~2727–2728). A journalled title does not re-seed them, so a long named session is immediately eligible on the next follow-up: `should_refresh_theme(N, 0, 0)` is true for any `N >= 4`, and `_retitle_checked_at == 0` opens the 30s floor too. That is a defensible trade-off — the theme prompt is the real pivot defense, and one extra isolated call after a restart is cheap — but it means the “settled session spends no call” guarantee does not survive a process restart. Either seed `_last_titled_turn_count` from `history()` on adopt when a title is already stored, or add a test that documents the reset so a later change cannot treat it as accidental.

#### Nit

**n1.** `OperatorApp._clock` still says a test can “cross a two-minute gap”; the floor is now 30s (`app.py` ~6888).

**n2.** The `THEME_HEAD_TURNS` comment says the head is “kept larger relative to the tail”, but the constants are head 3 / tail 4 (`naming.py` ~154–163).

**n3.** `_named_a_while_ago` and `test_a_settled_session_does_not_even_spend_a_check_on_an_in_goal_step` still describe the first-title baseline as `0`. `_store_title` floors it at `1` (`app.py` ~7141), which is why the helper’s `grow_transcript(6)` is the number that actually matches the documented ≥6 schedule.

#### Review-focus notes (not findings)

- **Sampler keeps the opening turns.** Head is `collected[:3]` before tail/elision/newest-append; a long opener is cut per-turn at 240 chars from the start, so the ask survives. Newest-append cannot evict the head. Elision sits between the halves, never over them. A greeting-then-subject opener can push the real ask out of a 3-turn head on a long chat; that is the fixed-window limit, not a drop of turn 0.
- **Counters.** Reset on swap/reload; charged on “no change” (and, indistinguishably, on failure — the safe direction); first title floored at 1. Lost across `/resume` — see m2. Not persisted on `ConversationName`; that split is intentional.
- **Prompt cost.** A max 7-turn envelope is ~2 KB + a 716-char system prompt, vs the old 320-char newest-message budget. Still a handful of isolated calls capped at five; not a cost problem. The `RETITLE_MIN_GAP_S` comment’s “~167 input tokens” figure is the old measurement.
- **Call sites.** Only `app.py:7012` in production; `turns` is keyword-only with a default, so the test helpers that omit it still exercise the empty-history fallback. Mobile names once via `generate_title` and never retitles.

#### Verdict
`clean` — this round is TERMINAL. m1/m2 and the nits are follow-ups, not a reason for another review round.

damianvtran:

Reviewed head: `3a950a3d26ca9bb192a1a168501c0a23944ba76c`. No blocker/major findings; this round is terminal. Minors and nits deferred rather than patched, so the reviewed SHA stays the merge SHA.

- **m1** — deferred. The dedicated sampler test (`test_build_theme_context_samples_head_and_tail_with_an_elision_marker`) already asserts the opener (`turn 0`) is present and a middle turn is dropped. Tightening the regression fake to a phrase that exists only on the opener is a good follow-up; it does not change the production path. Ticket: tighten `test_an_in_goal_step_does_not_pivot_the_title` to assert an opener-only phrase (e.g. `"for the agent"`).
- **m2** — deferred, accepted as the documented trade-off. Growth counters are host-local by design (`ConversationName` is the persisted/shared holder and must not grow a second schedule). One extra isolated call after `/resume` is cheap; the theme prompt is the pivot defense. Ticket: either seed `_last_titled_turn_count` from `history()` on adopt when a title is already stored, or add a test that pins the reset so a later change cannot treat it as accidental.
- **n1** — deferred. `_clock` docstring still says “two-minute gap”; the floor is 30s. Doc-only.
- **n2** — deferred. `THEME_HEAD_TURNS` comment says the head is “kept larger relative to the tail”; constants are 3 / 4. The *importance* of the head is larger, not the count. Doc-only.
- **n3** — deferred. Helper comments still describe the first-title baseline as `0`; `_store_title` floors it at `1`. The helper’s `grow_transcript(6)` already matches the real schedule. Doc-only.

damianvtran:
### Agent review — round 2
Scope: 769762fcd99fd22e8c875e57f4bb23ffcc3e5851..d9ff5e404dcf6522f8b3178c31e477615070a10e

Freshness round after the rebase onto #277 (origin/main `769762fc`, version 0.26.1). Round 1 was TERMINAL against `3a950a3d`.

`git diff 3a950a3d d9ff5e40` on the naming surface is empty:

- `local_operator/session/naming.py`
- `local_operator/tui/app.py`
- `tests/unit/tui/test_conversation_naming.py`
- `pyproject.toml`

The only tree delta vs the reviewed commit is the #277 usage-cache replay (`local_operator/model/configure.py`, `local_operator/providers/usage_cache.py`, and their tests). That is unrelated to titling. The version bump commit still says 0.27.0 and now sits on top of 0.26.1, which is the correct minor for this feature.

37/37 `tests/unit/tui/test_conversation_naming.py` passed on `/tmp/lop-naming` at `d9ff5e40`.

Round-1 minors/nits (m1, m2, n1–n3) are unchanged and are not re-opened.

#### Blocker
None.

#### Major
None.

#### Minor
None new.

#### Nit
None new.

**Verdict: clean.** This round is TERMINAL. Naming behaviour is unchanged from the reviewed tree; remaining round-1 follow-ups stay follow-ups.

damianvtran:

Reviewed head: `d9ff5e404dcf6522f8b3178c31e477615070a10e`. Freshness round; no new findings. Naming files are unchanged from the round-1 reviewed tree (`3a950a3d`); the only delta is the replay onto #277 and the still-correct 0.27.0 bump on top of 0.26.1.

Round-1 minors/nits remain deferred as previously recorded. This round is terminal.

PR #281, opened 2026-08-23
fix(session): accept untagged titles from non-Anthropic models

damianvtran:
## Why

Sessions started on models like `xai/grok-4.6` keep the opener excerpt as the session name forever. Claude usually wraps `<title>…</title>` and gets a real title; Grok and other non-Anthropic models often reply with a bare 3–7 word title, leak thinking (`<think>` / fenced reasoning) around a tag, or emit `{"title": "…"}`. All of those currently become `None`, so the DISPLAY-ONLY provisional excerpt stays on the band and the tab.

Compounding request-shape problem: `ERRAND_MAX_TOKENS = 128` counts thinking tokens. xAI / DeepSeek / Kimi have no effort ladder (`model.effort` says so explicitly), so `_lowest_effort` is a no-op. A reasoning model can spend the whole 128 thinking, emit no visible title, and `parse_title` still returns `None` while billing the thinking. omp already hit this (oh-my-pi #4355) and raised the title ceiling to 1024.

This is a model-agnostic parser + ceiling fix, not a Grok special case. Version bump **0.27.0 → 0.27.1** (patch).

## What changes

1. **`parse_title`** — port omp's title-extraction policy into the existing Python idiom. No second parser.
   - Strip leaked thinking envelopes (`<think>` / `<thinking>` / `<reasoning>`, and ` ```thinking` / ` ```reasoning ` fences) *before* looking for a title. Last visible marked title wins; a tag inside thinking is not the title.
   - Well-formed `<title>…</title>` still uses today's normalisation (sentence case, quote/punct strip, first line only, both caps as rejection).
   - No tag: accept a short untagged reply that already looks like a title. Reject thinking preambles ("Thinking process:", "Here's my reasoning:") and anything over `MAX_TITLE_CHARS` / `MAX_TITLE_WORDS` (rejection, not truncation).
   - Unwrap `{"title": "…"}` (optionally fenced) the way omp's `unwrapJsonTitle` does.
   - `<title/>` / empty / quotes-only still `None`.
2. **`Session.ERRAND_MAX_TOKENS` 128 → 1024** so a model with no effort ladder can finish thinking *and* still emit the title. Headroom is free: only produced tokens are billed. No unverified `disable_reasoning` wire key (a 400 on the errand is swallowed and the session stays unnamed).
3. **`temperature=0`** on the naming `complete_once` request. Titling is extraction. `_sampling_params` already drops the key when the spec rejects sampling.

[...]

PR #283, opened 2026-08-23
feat(update): add PyPI upgrade and process relaunch

damianvtran:
## Why

There is no in-app way to upgrade a PyPI install or to pick up a replaced wheel. `/reload` rebooted MCP/credentials/model inside the old interpreter, so a just-updated binary never came back. End users were pointed at `lop-update`, which archives local git `main` — the opposite audience.

This is a minor feature. Version **0.27.1 → 0.28.0**.

### PyPI updater — `local_operator/update.py`

`lop update` / `/update` upgrade the **installed distribution from PyPI**, not git. Latest is always `https://pypi.org/pypi/local-operator/json` → `info.version`, compared to `importlib.metadata.version("local-operator")` — the same source the splash version row and `lop --version` already use.

Detected install → one command:

| Detected install | Command |
|---|---|
| uv tool (`sys.prefix` under `…/uv/tools/local-operator`, or `uv-receipt.toml`) | `uv tool upgrade local-operator` |
| pipx | `pipx upgrade local-operator` |
| Ordinary venv / pip | `[sys.executable, "-m", "pip", "install", "-U", "local-operator"]` |

Refuse, do not guess: editable / source checkout (repo `.venv`) and unknown layouts exit 1 with how this process was launched. `.lop-source` (developer uv-tool snapshot) still upgrades from PyPI after one printed line; it does **not** call `lop-update`.

`lop update --check` prints installed vs PyPI and exits 0 (current), 2 (behind), or 1 (error). No `--yes`. No prompt. CLI upgrade does not re-exec — the user is in a shell.

Subparser help names the collision:

```
update  Upgrade this install from PyPI. Not the lop-update script,
        which rebuilds the global runtime from a local git checkout.
```

### Shared relaunch — `local_operator/reexec.py`

`/update` and `/reload` stash a `RestartPlan` and `app.exit(return_code=75)`. `run_tui` suppresses the `session ended — resume with:` hint on 75. `cli.main` then `os.execvpe`s (POSIX) or `Popen` + `_exit` (Windows).

`--resume` is rewritten from the on-disk transcript id (same gate as `resume_hint()`). No transcript yet → cold launch. Mid-turn / `/loop` refuses both commands (`esc` first). `/new` and `/resume` stay in-process.

### Splash `!` row

A one-shot mount worker (after first paint, not on the welcome poll timer) calls `check_latest()`. Newer PyPI fills `WelcomeInfo.update_available` and refreshes. Cache shape matches `model/catalogue.py` (`~/.local-operator/cache/pypi-local-operator.json`, TTL 6 hours, stale-used-on-failure, silent on miss). The row is **not reserved** when current.

Copy: `! latest is v0.28.0 — /update` (narrow: drop `/update` whole). Placement: version → model → cwd → **update** → harness notice → login warning. Shed priorities bumped so login still outranks everything.

### Unit — `tests/unit/test_update.py`, `test_reexec.py`, `tests/unit/tui/test_update_command.py`

- `check_latest` against a fake `httpx` transport: newer / same / older / 500 / timeout / corrupt cache / stale-used-on-failure / TTL hit skips GET.
- `install_kind` with tmp prefixes: uv-receipt, uv path, pipx, editable, no distribution, unknown.
- `perform_upgrade` asserts the argv it would run; subprocess is injected. Does not hit real PyPI install.
- `lop update --check` via `main()` with `sys.argv` patched: exit 0 / 2 / 1 and stdout.
- `plan_argv`: no resume, inject, replace existing, preserve `--model`/`--hosting`/`--train`.
- `replace_self`: POSIX `execvpe` once; Windows backend `Popen` + `_exit(0)`.
- TUI: `/update` when current does not set return_code 75; stubbed successful upgrade exits 75; `/reload` exits 75 with a resumable id stashed; live turn refuses both.
- Echo policy pin (`"update": False`). Splash geometry / shed order / no row when `None`.

### Visual — real `OperatorApp`, 100×30

| Frame | Path | content height | virtual vs size | scrollbar |
|---|---|---|---|---|
| before (origin/main) | `/tmp/before.svg` | 22 | 98×28 == 98×28 | no |
| after, current | `/tmp/after-current.svg` | 22 | 98×28 == 98×28 | no |
| after, `update_available="0.28.0"` | `/tmp/after-available.svg` | 23 | 98×28 == 98×28 | no |

Looked at the rendered PNGs. Current matches today's splash (no reserved row). Available adds one yellow `! latest is v0.28.0 — /update` under cwd, above the hint table. No reflow of the lockup or composer.

### Live PyPI check (this worktree, not the installed `lop`)

```
$ PYTHONPATH=/tmp/lop-update-cmd .venv/bin/python -c "from local_operator.update import update_command; raise SystemExit(update_command(check=True))"
local-operator 0.17.4
latest on PyPI: 0.27.1
run `lop update` to install
EXIT:2
```

The installed `lop` runtime does not yet have the subcommand (`invalid choice: 'update'`), as expected.

### Gates

```
.venv/bin/python -m flake8 .                         # pass
uvx --from black==26.1.0 black --check local_operator tests   # pass
uvx isort --check-only --profile black local_operator tests   # pass
.venv/bin/python -m pyright --pythonpath .venv/bin/python .   # pass (0 errors)
  # 6487 passed, 7 skipped; 3 failures are pre-existing flakes also
  # reproduced (eval stream) or passing (comms, bash cancel, jobs wait)
  # on origin/main. Feature tests: 42/42.
```

## Cannot prove in-unit

A real `uv tool upgrade` + living TUI `exec` was **not** exercised. Do not treat the mocked installer argv or the exit-75 pilot as a published-upgrade.

## Design decisions (not redesigned)

- PyPI, not `lop-update`.
- `/reload` now re-execs; `/new` and `/resume` stay in-process.
- Splash `!` row, not reserved, not `info.notice`.
- No `--from-source`, no `--yes`, no new dependency.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,9 @@ pip install local-operator     # pipx install local-operator on Linux (PEP 668)
 ```
 
 The install provides both `local-operator` and its short alias `lop` — the
-rest of this page uses `lop`.
+rest of this page uses `lop`. `lop update` upgrades that install from PyPI;
+`lop-update` (hyphen) is the developer script that rebuilds the global
+runtime from a local git checkout.
 
 Sign in to a provider (or skip this — on an interactive terminal `lop` opens
 in a setup state and walks you through `/login`; a headless or piped run prints
--- a/README.md
+++ b/README.md
@@ -182,7 +182,8 @@ current session's cost live).
 | `/effort` | Show or set reasoning effort (`shift+tab` cycles) |
 | `/approvals` | Set whether tools ask first (`ask`/`auto`; add `default` to keep it) |
 | `/resume` | Pick a past conversation and continue it |
-| `/new`, `/clear`, `/reload` | Fresh conversation · wipe the screen · reboot the session in place |
+| `/new`, `/clear`, `/reload` | Fresh conversation · wipe the screen · relaunch this conversation on the current install |
+| `/update` | Install the latest version from PyPI and relaunch |
 | `/goal`, `/loop` | Set an objective, then iterate autonomously toward it |
 | `/btw` | Ask a side question off the record — it never joins the conversation |
 | `/compact` | Compact the context now (it also happens automatically) |
--- a/local_operator/cli.py
+++ b/local_operator/cli.py
@@ -401,6 +401,22 @@ def build_cli_parser() -> argparse.ArgumentParser:
     serve_mobile_parser.add_argument("--port", type=int, default=4098)
 
     # Exec command for single execution mode
+    # PyPI upgrade. Not ``lop-update`` (hyphen), which archives local git
+    # ``main`` into the uv-tool env — opposite audience, never invoked here.
+    update_parser = subparsers.add_parser(
+        "update",
+        help=(
+            "Upgrade this install from PyPI. Not the lop-update script, "
+            "which rebuilds the global runtime from a local git checkout."
+        ),
+        parents=[parent_parser],
+    )
+    update_parser.add_argument(
+        "--check",
+        action="store_true",
+        help="Print installed vs PyPI; do not install",
+    )
+
     exec_parser = subparsers.add_parser(
         "exec",
         help="Execute a single command without starting interactive mode",
@@ -2086,6 +2102,12 @@ def main() -> int:
             if invalid is not None:
                 return invalid
             return mcp_command(args)
+        elif args.subcommand == "update":
+            # Lazy: ``update`` imports httpx. ``import local_operator.cli``
+            # must not (``tests/unit/test_import_graph.py``).
+            from local_operator.update import update_command
+
+            return update_command(check=bool(getattr(args, "check", False)))
         elif args.subcommand == "exec":
             # Single-execution mode: headless one-shot (README contract —
             # exit 0 on success, non-zero on error). Working-directory
@@ -2350,13 +2372,23 @@ def main() -> int:
                 # which must keep its console output, so the wrapping goes on
                 # this call site rather than inside it.
                 with file_logging():
-                    return asyncio.run(
+                    tui_code = asyncio.run(
                         _run_with_scheduler(
                             tui_entry,
                             session_factory,
                             theme_name,
                         )
                     )
+                # 75: the TUI asked to be replaced after a clean teardown.
+                # ``replace_self`` does not return; a missing plan is a bug.
+                from local_operator.reexec import REEXEC_CODE, replace_self, take_plan
+
+                if tui_code == REEXEC_CODE:
+                    plan = take_plan()
+                    if plan is None:
+                        return 1
+                    replace_self(plan)
+                return tui_code
             finally:
                 try:
                     tui_controller.close()
--- a/local_operator/reexec.py
+++ b/local_operator/reexec.py
@@ -1,0 +1,164 @@
+"""Replace this process with a fresh launch of the same install.
+
+WHY THIS EXISTS
+---------------
+``/reload`` used to reboot MCP/credentials/model inside the *old*
+interpreter. After ``lop update`` (or any other package replace) that
+cannot pick up the new wheel. ``/update`` and ``/reload`` therefore share
+one helper that ends the TUI cleanly and then replaces the process.
+
+WHY NOT ``os.execv`` FROM INSIDE THE TUI
+----------------------------------------
+Textual owns the alternate screen. Exec-ing over a live ``OperatorApp``
+leaves the terminal half-raw. ``run_tui`` already returns
+``app.return_code`` after ``run_async()`` and restores the terminal in
+that teardown — that is the seam. The handler stashes a :class:`RestartPlan`,
+exits with :data:`REEXEC_CODE`, and ``cli.main`` calls :func:`replace_self`
+only after the TUI has released the tty.
+
+``75`` is ``EX_TEMPFAIL``: not 0/1/130, which ``run_tui`` already uses
+for success, error, and Ctrl-C.
+
+POSIX uses ``os.execvpe`` so the same PID comes back with cwd/env
+preserved by definition. Windows console shims make ``os.execv``
+unreliable, so that backend is ``Popen`` + ``os._exit(0)``. One helper,
+two OS backends — no detached babysitter unless a file-lock on
+``uv tool upgrade`` later forces it.
+"""
+
+from __future__ import annotations
+
+import os
+import subprocess
+import sys
+from dataclasses import dataclass
+
+#: Private exit from the TUI that means "replace this process", not "the
+#: user quit". ``run_tui`` suppresses the ``session ended — resume with:``
+#: hint on this code: that line is for a human staying in the shell.
+REEXEC_CODE = 75
+
+
+@dataclass(frozen=True)
+class RestartPlan:
+    """What the next process should look like.
+
+    ``argv`` is the rewritten launch line. ``resume_id`` is recorded so
+    tests (and a future Windows helper) can see the session the plan
+    targeted without re-parsing argv.
+    """
+
+    argv: list[str]
+    resume_id: str | None = None
+
+
+def _drop_flag(argv: list[str], name: str) -> list[str]:
+    """Remove ``name`` and, when it takes a value, the following token."""
+    out: list[str] = []
+    skip = False
+    for token in argv:
+        if skip:
+            skip = False
+            continue
+        if token == name:
+            skip = True
+            continue
+        if token.startswith(f"{name}="):
+            continue
+        out.append(token)
+    return out
+
+
+def plan_argv(
+    original: list[str] | None = None,
+    *,
+    resume_id: str | None = None,
+) -> list[str]:
+    """``sys.argv`` as launched, with ``--resume`` rewritten.
+
+    A transcript on disk (the same id ``resume_hint()`` advertises) is
+    injected so the new process reopens this conversation. No transcript
+    yet — first turn never persisted — strips any ``--resume`` so the new
+    process is a cold launch, matching what today's in-process ``/reload``
+    already accepts when it cannot rebind.
+
+    A previous ``--resume`` is dropped before the new one is written so
+    the flag cannot appear twice. ``--hosting`` / ``--model`` / ``--agent``
+    / ``--tui`` / ``--debug`` / ``--train`` ride along because they were
+    already on the launch line; we do not invent them.
+    """
+    launched = list(original if original is not None else sys.argv)
+    if not launched:
+        launched = [sys.argv[0] if sys.argv else "lop"]
+
+    rewritten = _drop_flag(launched, "--resume")
+    if resume_id:
+        rewritten.extend(["--resume", resume_id])
+    return rewritten
+
+
+def make_plan(
+    original: list[str] | None = None,
+    *,
+    resume_id: str | None = None,
+) -> RestartPlan:
+    return RestartPlan(argv=plan_argv(original, resume_id=resume_id), resume_id=resume_id)
+
+
+_pending: RestartPlan | None = None
+
+
+def stash_plan(plan: RestartPlan) -> None:
+    """Park a plan for ``cli.main`` to consume after the TUI has torn down.
+
+    The handler cannot ``exec`` from inside Textual (alternate screen). It
+    exits 75 and leaves the plan here so ``replace_self`` runs only after
+    ``run_tui`` has restored the terminal.
+    """
+    global _pending
+    _pending = plan
+
+
+def take_plan() -> RestartPlan | None:
+    """Return and clear the parked plan (``None`` if the TUI never stashed)."""
+    global _pending
+    plan = _pending
+    _pending = None
+    return plan
+
+
+def _on_windows() -> bool:
+    """Indirection so tests can flip the backend without patching ``sys``."""
+    return sys.platform == "win32"
+
+
+def _replace_windows(argv: list[str], env: dict[str, str]) -> None:
+    """Spawn a sibling then exit. ``os.execv`` on Windows console shims is unreliable."""
+    flags = 0
+    create_new = getattr(subprocess, "CREATE_NEW_PROCESS_GROUP", 0)
+    if isinstance(create_new, int):
+        flags = create_new
+    subprocess.Popen(  # noqa: S603 — relaunch of our own argv
+        argv,
+        cwd=os.getcwd(),
+        env=env,
+        close_fds=True,
+        creationflags=flags,
+    )
+    os._exit(0)
+
+
+def _replace_posix(argv: list[str], env: dict[str, str]) -> None:
+    os.execvpe(argv[0], argv, env)
+
+
+def replace_self(plan: RestartPlan) -> None:
+    """Never returns on success. POSIX execs; Windows spawns then exits."""
+    argv = list(plan.argv)
+    if not argv:
+        raise RuntimeError("RestartPlan.argv is empty")
+    env = os.environ.copy()
+    if _on_windows():
+        _replace_windows(argv, env)
+        return
+    _replace_posix(argv, env)
--- a/local_operator/tui/__init__.py
+++ b/local_operator/tui/__init__.py
@@ -55,7 +55,12 @@ async def run_tui(
             # user's scrollback where it can be copied — printing it from inside
             # the app would put it in a frame that is being torn down. In the
             # `finally` so it survives the exit paths as well as the clean one.
-            hint = app.resume_hint()
-            if hint:
-                print(f"\nsession ended — resume with:\n  {hint}\n")
+            from local_operator.reexec import REEXEC_CODE
+
+            # A relaunch is about to replace this process; the hint is for a
+            # human who is staying in the shell.
+            if app.return_code != REEXEC_CODE:
+                hint = app.resume_hint()
+                if hint:
+                    print(f"\nsession ended — resume with:\n  {hint}\n")
         return int(app.return_code or 0)
--- a/local_operator/tui/app.py
+++ b/local_operator/tui/app.py
@@ -25,6 +25,7 @@ from __future__ import annotations
 import asyncio
 import logging
 import os
+import sys
 import time
 from collections.abc import Mapping
 from dataclasses import dataclass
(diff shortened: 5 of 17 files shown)


damianvtran:
### Design review — round 1

Reviewer: designer / anthropic/claude-opus-4-8
Scope: origin/main..af8b838c6a47eb0faa134eeda14235b1d2edee8c

Looked at the rendered OperatorApp stills, not the source: `/tmp/before.svg`, `/tmp/after-current.svg`, `/tmp/after-available.svg`, plus two extra states I re-rendered from `/tmp/lop-update-cmd` (`after-both` = update + failover notice; `after-narrow-32` / `after-narrow-48`).

**Current (no update).** `after-current` matches today's splash: version `v0.17.4`, model, cwd, three hint rows, tip, composer. Same text-node count as `before` (60). No `#e0b04b`. Content height 22, virtual size == screen size, no scrollbar. The row is not reserved.

**Update available.** One extra text node. Yellow `! latest is v0.28.0 — /update` (`#e0b04b`) sits under cwd and above the hint table, left-aligned with the status stack. Not a fourth hint row. Lockup shifts up one row (content-sized splash on the input card); composer, hints, and tip stay put. Content height 23, still no scrollbar. Contrast of `#e0b04b` on the content ground `#1e1a14` is 8.64:1.

**Update + harness notice.** Both `!` rows paint, same glyph and same warning tint, in contract order: version → model → cwd → update → notice. `info.notice` is not overwritten. The long failover line is what left-shifts the shared status/hint edge (pre-existing `_center_blocks` behaviour; the update line alone only moves that edge by one cell).

**Narrow.** At width 32 the command drops whole: `! latest is v0.28.0`. No `/upd…`. At width 48 the full `— /update` still fits.

No blocker / major. Round 1 is **TERMINAL**.

Follow-ups (not findings; accepted by the contract):
- The probe fills after first paint, so a behind install grows one row after mount. Reserving the slot was rejected.
- Update and failover are the same visual weight when they coincide. That is the specified warning-family treatment.

damianvtran:
### Agent review — round 1

Reviewer: reviewer / anthropic/claude-opus-4-8
Scope: origin/main..af8b838c6a47eb0faa134eeda14235b1d2edee8c

No blockers. Two majors. The install-kind table, version compare, cache TTL, CLI lazy import, `app.exit(return_code=75)`, `/new`/`/resume` staying on `_reload_session`, and the splash shed order all match the contract.

### M1 — `_turn_is_live` does not cover compaction

`local_operator/tui/app.py:2905-2909`

```python
streaming = session is not None and bool(getattr(session, "is_streaming", False))
return bool(streaming or self._loop_running or self._streaming_block is not None)
```

`_compacting` is omitted. The same file already states why that is a live turn (`app.py:6807-6814`): compaction holds the session lock, `is_streaming` is False, and a submit in that window would splice into a rewrite. `/update` and `/reload` use this predicate as their only mid-turn gate, so both will `exit(75)` / start `uv tool upgrade` while history is being rewritten.

The live-turn test only sets `FakeSession.streaming = True` (`tests/unit/tui/test_update_command.py:103-123`), so this path is unpinned.

Include `self._compacting` (and, if you want the same belt the session uses, a locked turn lock). Add a pilot that sets `_compacting = True` and asserts neither command sets return code 75.

### M2 — `/update` and `/reload` are not mutually exclusive

`local_operator/tui/app.py:2940-2982`, `2911-2923`

`/update` checks `_turn_is_live` once, then runs `perform_upgrade` on a thread worker (`group="update"`, `exclusive` left false). Nothing else is armed:

- A second `/update` starts a second installer against the same prefix.
- `/reload` is synchronous: it stashes a plan and `exit(75)` immediately, which tears the process down while `subprocess.run(["uv", "tool", "upgrade", ...])` is still writing the env.
- After a successful upgrade, `_request_relaunch` re-checks `_turn_is_live`. A prompt typed during the installer makes relaunch refuse (`esc first`) even though the wheel is already replaced. The user is left on the old interpreter after a notice that said `restarting…`.

Hold an in-progress / pending-relaunch flag that `_cmd_update` and `_request_relaunch` both honour: refuse a second `/update`, refuse `/reload` while the installer is running, and do not let a newly started turn cancel a completed upgrade's relaunch (lock the composer, or relaunch anyway and let the new process resume the transcript).

### N1 — CI isort is red on `local_operator/tui/app.py`

GitHub Actions `lint` (`isort --check .`) failed on this file:

https://github.com/damianvtran/local-operator/actions/runs/32662272183/job/97250142597

The PR body reports `uvx isort --check-only --profile black` as passing; CI does not pass `--profile black`. Re-run the same invocation CI uses and commit the sort it asks for.

### N2 — TUI `/update` skips the `.lop-source` line

Architect contract: a git snapshot still upgrades from PyPI after one printed line. `update_command` does that (`update.py:469-470`). `_run_update` calls `perform_upgrade` with no `is_git_snapshot()` notice (`app.py:2972-2974`), so `/update` on a `lop-update` runtime replaces the snapshot silently. Print the same line (as a `_notice`) before running the installer.

### N3 — stale `/reload` comments

`_reload_session`'s docstring still says `keep_context` is `/reload`'s intent (`app.py:2503-2507`). `_resumable_session_id` still says `_cmd_reload` rebinds the factory (`app.py:4881-4883`). `/reload` no longer calls either path. Update the comments so the next reader does not restore the in-process reboot.

**Verdict:** needs remediation. Not terminal.

damianvtran:
### UX review — round 1

Reviewer: ux-reviewer / anthropic/claude-opus-4-8
Scope: origin/main..af8b838c6a47eb0faa134eeda14235b1d2edee8c

Walked the real `OperatorApp` via `run_test` (100×30) on `/tmp/lop-update-cmd` @ `af8b838c`, plus `python -m local_operator.cli update --check` / `--help` from the worktree. Did not run a real `uv tool upgrade` against the global `lop`.

**U1 — MAJOR — composer stays live through a process-killing `/update`, so a typed-ahead turn can cancel the relaunch after the wheel is already installed.**

Step: `/update` when behind. After the live check, the app paints `updating to v0.28.0…`, retires the splash, and leaves the composer **empty and focused**. A 0.4s stubbed installer was long enough to sit in that state; a real `uv tool upgrade` is much longer. Nothing says the input is reserved. If the user sends a prompt (or `/loop`) during the install, `_request_relaunch` then hits `_turn_is_live()` and prints `esc first — a turn is still running` instead of exiting 75. The user has already been told `updated to v0.28.0 — restarting…`, which is now a lie: the new wheel is on disk, this process is still the old interpreter, and they have a turn in flight. Expect: lock or dim the composer for the upgrade, or stash the relaunch and apply it the moment the turn ends, and do not print "restarting" until the exit is actually happening.

**U2 — MINOR — `/reload` tears the process down with no receipt, and a first-turn `/reload` is a silent cold start.**

Step: `/reload` on a session with a transcript exits 75 immediately, splash still up, `notices=[]`, plan carries `--resume`. Same command on a first-turn splash (no `transcript.jsonl`) also exits 75 with `resume_id=None` and no copy at all. The TUI just vanishes and comes back as a new launch. `/update` at least says `restarting…`. A user who still thinks `/reload` is the old in-process reboot (README still says "reboot the session in place") will read this as a crash. Expect a one-line `relaunching…` before exit 75, and on a first-turn plan with no resume id, say that this will open a new session.

**U3 — MINOR — a failed `/update` from the splash destroys the empty state and leaves a developer error.**

Step: `/update` when behind, then `UpdateError`. Because progress goes through `_notice` (not `_system_notice`), the welcome lockup is already gone. What remains is:

- `updating to v0.28.0…`
- `this interpreter is the repo .venv, not an installed distribution. update the global runtime with lop-update after the change is merged.`
- or `installer exited 1`

The first refusal is contributor-speak; the second is an exit code with no next step. Splash, hints, and `/update` CTA are gone. Recoverable only if the user knows `/clear` brings the welcome back. Expect: keep the splash for progress and refusal (`_system_notice`), and word both errors for the person in front of the app (`this is the repo checkout, not the installed lop — run lop-update after merge` / `upgrade failed; try \`uv tool upgrade local-operator\` in a shell`).

**U4 — MINOR — no feedback while the forced PyPI check is in flight.**

Step: `/update` with a 0.5s `check_latest`. Immediate frame: splash still up, `notices=[]`, composer empty. Only after the check returns does `already on v0.27.0` (or `updating to…`) appear. `force=True` always hits the network, so a slow or flaky PyPI looks like the command did nothing. Expect `checking for updates…` on submit, same as the later `updating to…` line.

**U5 — MINOR — `/search on|off` still tells the user to `/reload` as if it were a cheap in-process refresh.**

Step: `/search on` notice is `web search on; /reload adds the tool`. After this change `/reload` replaces the process. The advice still works, but it now costs a full relaunch for a tool-toggle the old command did in place. Expect copy that matches the new cost (`/reload` relaunches the app to pick up the tool) or a lighter path.

**U6 — NIT — README slash table is the old `/reload` and has no `/update`.**

`/new`, `/clear`, `/reload` is still "Fresh conversation · wipe the screen · reboot the session in place". In-app picker/help are correct (`Relaunch this conversation on the current install` / `Install the latest version from PyPI and relaunch`). Anyone who learns commands from the README will miss the feature the splash is advertising.

**U7 — NIT — `lop update --help` inherits `--yolo` / `--resume` / `--train` / `--agent`, which do nothing here.**

`lop --help` names the `lop-update` collision well; `--check` stdout (`run \`lop update\` to install`) points at the right command. The collision is discoverable enough. The inherited parent flags on `update --help` are the leftover confusion.

**U8 — NIT — mid-turn refuse copy is the same for `/loop` as for a streaming turn.**

`esc first — a turn is still running` is the right key and it keeps the session. For a running `/loop` it is slightly the wrong noun, but esc is still the escape route.

Happy paths that felt right: splash `! latest is v0.28.0 — /update` (narrow drops `/update` whole, not half); already-latest stays put with `already on v0.27.0` and no exit 75; picker lists `/update` on `/` and isolates it on `/upd`; live-turn and `/loop` both refuse without tearing down; CLI `--check` prints versions and does not re-exec; worktree `--check` exited 2 with `local-operator 0.17.4` / `latest on PyPI: 0.27.2` / `run \`lop update\` to install`.

**Verdict:** not TERMINAL — U1 is still open. U2–U5 are follow-ups; U6–U8 are nits.

damianvtran:

Head: `93bf353e84ff1c59e9d428002b98f6f1fe9c7836`

- **M1** — **fixed** in `93bf353e`. `_turn_is_live` now includes `self._compacting`. Pilot `test_compacting_refuses_update_and_reload` sets `_compacting = True` and asserts neither `/update` nor `/reload` sets return code 75.
- **M2** — **fixed** in `93bf353e`. `_update_in_progress` is the mutex (`group="update"` stays non-exclusive because Textual exclusivity would cancel a live `uv tool upgrade`). Second `/update` is refused; `/reload` during an in-flight upgrade is refused; the composer is read-only for the duration and stays locked through a successful upgrade's `exit(75)`. `_request_relaunch(force=True)` still exits 75 even if something would have looked live. `restarting…` is printed only on the same tick as that exit.
- **N1** — **fixed** in `93bf353e`. Confirmed with `uvx isort==5.13.2 --check-only .` (no `--profile black`). The local import in `_run_update` is now the single-line form CI wants.
- **N2** — **fixed** in `93bf353e`. TUI `/update` prints `git_snapshot_notice()` via `_system_notice` before the installer when `.lop-source` is present.
- **N3** — **fixed** in `93bf353e`. `_reload_session` docstring no longer attributes `keep_context` to `/reload`; `_resumable_session_id` now names `_request_relaunch` as the consumer.

damianvtran:

Head: `93bf353e84ff1c59e9d428002b98f6f1fe9c7836`

- **U1** — **fixed** in `93bf353e` (same bug as M2). Composer is locked (`_set_composer_read_only(True)`) for the whole `/update`, including through relaunch. A second `/update` and a `/reload` during install are refused. After a successful upgrade, relaunch is forced so a newly started turn cannot cancel exit 75. `restarting…` is printed only when that exit is actually happening.
- **U2** — **fixed** in `93bf353e`. `/reload` prints `relaunching…` before exit 75. When there is no transcript yet (`resume_id` is None) it says `relaunching… this will open a new session`.
- **U3** — **fixed** in `93bf353e`. Progress and refusal go through `_system_notice`, so a failed `/update` from the splash keeps the welcome lockup. Editable refusal is now `this is the repo checkout, not the installed lop — run lop-update after merge`. Installer failure is `upgrade failed; try \`uv tool upgrade local-operator\` in a shell` (or the pipx/pip equivalent from `install_kind`).
- **U4** — **fixed** in `93bf353e`. `/update` immediately shows `checking for updates…` via `_system_notice` before the forced PyPI check returns.
- **U5** — **fixed** in `93bf353e`. `/search on|off` now says `/reload relaunches the app to add|remove the tool`.
- **U6** — **fixed** in `93bf353e`. README slash table: `/reload` is `relaunch this conversation on the current install`; `/update` is listed as `Install the latest version from PyPI and relaunch`.
- **U7** — **rejected**. `--yolo` / `--resume` on `update --help` come from `_propagate_global_flags`, which re-declares those flags on every subparser so a root-level value is not clobbered (see `cli.py`). That is the house style; a one-off parent-parser exception for `update` would be a second way of doing the same thing.
- **U8** — **fixed** in `93bf353e`. A running `/loop` (and nothing else live) now refuses with `esc first — a loop is still running`.

damianvtran:
### UX review — round 2

Reviewer: ux-reviewer / anthropic/claude-opus-4-8
Scope: origin/main..93bf353e84ff1c59e9d428002b98f6f1fe9c7836

Walked the real OperatorApp via `run_test` (100×30) on `/tmp/lop-update-cmd` @ 93bf353e. Installer and PyPI were stubbed; did not run a real `uv tool upgrade` against global `lop`.

U1 is fixed. Mid-upgrade the composer is `read_only`, not focusable, and unfocused. A typed-ahead Enter does not start a turn. `restarting…` appears only on the same tick as exit 75. A second `/update` and a `/reload` during the install both print `an update is already running` and leave a single installer. Compaction refuses both commands with `esc first — a turn is still running`. Already-latest stays put (`already on v0.27.0`, no 75), splash survives via `_system_notice`, and the composer unlocks. Editable refuse keeps the splash and says `this is the repo checkout, not the installed lop — run lop-update after merge`. Installer failure keeps the splash and says `upgrade failed; try \`uv tool upgrade local-operator\` in a shell`. `/reload` with a transcript prints `relaunching…` then 75 with `--resume`. First-turn `/reload` prints `relaunching… this will open a new session`. `/search on|off` names the relaunch cost. U8 copy now says `a loop is still running`. U7 rejection accepted (parent flags are house style).

U9 — NIT — after a stay-put `/update` (already latest or refuse) the composer unlocks but focus is not restored.

Step: `/update` when already on v0.27.0. Immediate frame correctly shows `checking for updates…` and locks the dock. After `already on v0.27.0` the placeholder is `Message Local Operator…` again and `read_only` is false, but `screen_focus` is `None`. The next keystroke goes nowhere. Tab lands on the transcript; a second Tab returns the caret. Subagent/org-chart unlock restores focus; `_finish_update` only flips the lock. Expect the caret back on the composer the way those modes do. Not a trap on the success path (the process is exiting).

U10 — NIT — the locked dock during `/update` reuses `Read-only — press esc to reply`.

Step: mid-upgrade frame. Esc does not abort the install, does not unlock the composer, and does not start a turn (correct). The placeholder still promises a reply. Harmless; the lock holds.

No BLOCKER. No MAJOR. Round is TERMINAL. U9–U10 are follow-ups.

damianvtran:
### Agent review — round 2

Reviewer: reviewer / anthropic/claude-opus-4-8
Scope: origin/main..93bf353e84ff1c59e9d428002b98f6f1fe9c7836

No blocker or major findings.

Round 1 remediations verified against `af8b838c..93bf353e` and the live handlers:

- **M1** — `_turn_is_live` includes `self._compacting` (`app.py:2932-2934`). `test_compacting_refuses_update_and_reload` sets the flag with `is_streaming` false and asserts neither command sets return code 75.
- **M2** — `_update_in_progress` is armed before `run_worker` (`app.py:3000-3010`). Second `/update` and `/reload` refuse on that flag. Composer is locked for the duration; `_finish_update` unlocks only on refuse / already-current / installer error. Success keeps the lock and calls `_request_relaunch(force=True)` on the same tick as `restarting…`. `group="update"` stays non-exclusive on purpose. Pilots pin the second-update refuse, reload-during-install refuse, composer lock through exit 75, and force-relaunch when something looks live.
- **N1** — `uvx isort==5.13.2 --check-only .` (no `--profile black`) is clean locally. CI lint job on this head is green (`32664417613`).
- **N2** — TUI prints `git_snapshot_notice()` via `_system_notice` before `perform_upgrade` when `.lop-source` is present.
- **N3** — `_reload_session` docstring and `_resumable_session_id` now name the re-exec path.

Checked leftover races: refuse paths call `_finish_update`; success does not; a newly started turn cannot cancel exit 75. `tests/unit/tui/test_update_command.py`, `test_update.py`, and `test_reexec.py`: 50 passed.

### MINOR

**m1** — `_close_subagent_view` / `_close_org_chart_view` unlock the composer unconditionally (`app.py:7764`, `app.py:7842`). `_finish_update` already refuses to unlock while those views are open, but the close paths do not honour `_update_in_progress`. Opening a subagent (panel click) or org-chart during an in-flight `/update` and then Esc-closing it re-enables Enter while `uv tool upgrade` is still writing. `force=True` still saves the relaunch, so this is not the original M2 failure; the composer lock is just incomplete. Honour the flag on those two close paths (or skip the unlock when `_update_in_progress`).

### NIT

**n1** — `_reload_session` body comments still describe `/reload` as the in-process keep-context caller (`app.py:2742`, `2758`, `2763`). The docstring is correct; a later reader of the body can still restore the old reboot.

**Verdict:** clean. Round is TERMINAL. Remaining minor/nit are follow-ups, not a reason for another round.

damianvtran:
### Design review — round 2

Reviewer: designer / anthropic/claude-opus-4-8
Scope: origin/main..93bf353e84ff1c59e9d428002b98f6f1fe9c7836

Looked at the idle stills `/tmp/after-current.svg` and `/tmp/after-available.svg` (rendered, not source). Current splash is still version → model → cwd → hints → tip → composer, no reserved update slot. Available splash still adds one yellow `! latest is v0.28.0 — /update` row under cwd and above the hint table; lockup shifts up one row, composer and hints stay put. Remediation `93bf353e` locked the composer during `/update` and routed progress/refusals through `_system_notice`, so the idle splash geometry is unchanged. No D-findings. Round 2 is TERMINAL.

damianvtran:
### UX review — round 3

Reviewer: ux-reviewer / anthropic/claude-opus-4-8
Scope: origin/main..1493bdd2347155cacaacb8ed371e5f68c7b875a2

Rebase-only onto origin/main. `git diff 93bf353e 1493bdd2` is parent noise (session-title evidence, analytics/model/naming from #281/#282/#284); the `/update` and `/reload` paths are unchanged (`_cmd_update` / `_cmd_reload` still present; PR files differ only by a parent comment on the title-errand token cap). Flows from round 2 stand: composer lock through a successful upgrade, mutexed second `/update` and mid-install `/reload`, compaction counted as a live turn, splash-safe `_system_notice` refusals. No new U-findings. No BLOCKER. No MAJOR. Round is TERMINAL. U9–U10 remain follow-ups.

damianvtran:
### Design review — round 3

Reviewer: designer / anthropic/claude-opus-4-8
Scope: origin/main..1493bdd2347155cacaacb8ed371e5f68c7b875a2

Rebase-only onto origin/main; idle splash stills still match r1/r2. Looked at `/tmp/after-available.svg` (rendered, not source): version → model → cwd → yellow `! latest is v0.28.0 — /update` → hints → tip → composer, lockup shifted one row, composer and hints stay put. No D-findings. Round 3 is TERMINAL.

damianvtran:
### Agent review — round 3

Reviewer: reviewer / anthropic/claude-opus-4-8
Scope: origin/main..1493bdd2347155cacaacb8ed371e5f68c7b875a2

Rebase-only SHA change onto origin/main (0.27.3). `git range-diff efeeeb5b..93bf353e origin/main..1493bdd2` shows commit 2 identical (`93bf353e = 1493bdd2`) and commit 1 differing only in the pyproject version-parent context (`-0.27.1` → `-0.27.3`); both heads keep `version = "0.28.0"`. No behaviour rewrite, no dropped commits. Tree-to-tree `93bf353e..1493bdd2` is the incoming main work (session-titles / analytics / naming), not a change to this PR. CI on `1493bdd2` is in progress (lint / type-check / test 3.12+3.13 / pip-audit). No blocker or major. Verdict: clean. Round is TERMINAL.

PR #288, opened 2026-08-24
fix(providers): fail over on opaque aggregator 400s

damianvtran:
## Summary

OpenRouter (and similar aggregators) relay an **upstream provider failure** as an HTTP 400 whose outer message is exactly `Provider returned error`. When `metadata.raw` carries the origin provider's real diagnostics, that body is actionable and the 400 is a genuine answer — kind `request`, correctly not retried.

But session `e13d092c093c` (ts 1787579469.9000828) recorded the same shape with `metadata.raw` holding only a bare sentinel:

```json
{"error":{"message":"Provider returned error","code":400,"metadata":{"raw":"ERROR","provider_name":"Stealth"}}}
```

The failure was intermittent: direct live probes of the same request shape and inputs at ~750k tokens returned 200, and the same model failed again during the investigation, triggering fallbacks. Classified as a refused request (`kind="request", retryable=False`), it hit the driver's primary-400 storm guard and **aborted the turn before credential rotation or model fallback could serve it** — for an upstream blip wearing a 400.

## Change

- `local_operator/providers/clients.py`: new `_is_opaque_aggregator_400()` detects the opaque shape in `raise_for_status()` — outer message exactly "Provider returned error" (case-insensitive) AND `metadata.raw` resolving to a bare sentinel (`ERROR`, also when JSON-quoted). Only that shape flips `retryable=True`; `_classify_fields` then returns `transient` via its existing rule, so no classifier changes were needed.
- The detection lives in `raise_for_status` deliberately: it is the one point where the structured `metadata.raw` is still available — `_classify_fields` only ever sees the composed message string.
- Genuine 400s with actionable upstream diagnostics (`raw` = "context length exceeded", a JSON error object, any substantive text) keep `kind="request"`; all direct-provider 400s are untouched; the same opaque shape arriving on a 5xx keeps its existing transient classification.
- Loop safety: `is_server_side_failure` treats `transient` as provider-side weather, so the error gets the standard small per-bearer allowance, then rotation, then the fallback chain — all bounded by the existing turn-wide `MAX_SERVER_FAULT_REQUESTS_PER_TURN = 12`. A genuinely broken request now burns a little budget before surfacing instead of aborting a turn rotation could have served.

## Testing evidence

Reproduction first (pre-fix, against the session's exact wire body):

```
kind      = request
retryable = False
message   = Provider returned error: ERROR
str       = invalid request (HTTP 400): Provider returned error: ERROR
```

Post-fix classification matrix, run against real wire bodies through `raise_for_status`:

| Body | kind | retryable |
|---|---|---|
| Session body (`raw:"ERROR"`) | `transient` | True |
| Same, `raw` JSON-quoted sentinel | `transient` | True |
| `raw` = real upstream JSON diagnostics | `request` | False |
| `raw` = substantive non-JSON text | `request` | False |
| Different outer message + sentinel raw | `request` | False |
| Direct-provider 400 (`max_tokens too large`) | `request` | False |
| Opaque shape on 502 (unchanged path) | `transient` | True |

End-to-end failover proof (`test_opaque_aggregator_400_fails_over_instead_of_aborting`): the primary raises the session's exact 400 built through the real wire mapper; the test asserts the turn is served by the fallback, the same credential was retried first (`attempts[0] == "k1"`), the sibling key was rotated to, and `rotate_sibling` actually ran. The existing storm-guard tests (`test_primary_request_400_aborts_without_walking_chain`, `test_primary_image_rejection_400_aborts_turn`, `test_primary_unknown_400_style_not_aborted_if_not_request_kind`, `test_fallback_request_400_still_walks`) still pass unchanged, pinning the preserved behavior.

Gates (all clean):
- `flake8 .` — clean
- `black==26.1.0 --check local_operator tests` — clean (after formatting)
- `isort --check-only --profile black` — clean (after fixing import order)
- Full unit suite: see CI; focused runs of both touched test files pass (7 new tests).

## Review focus

- The sentinel set is deliberately minimal (`ERROR` only) — it is the one observed value; widening it needs evidence.
- Whether `_OPAQUE_RAW_SENTINELS` should live in config — kept internal until a second aggregator shape shows up.

### DIFF
--- a/local_operator/providers/clients.py
+++ b/local_operator/providers/clients.py
@@ -302,11 +302,17 @@ def _strip_quote_pair(text: str) -> str:
     The peel LOOPS rather than running once to cover a value that arrives
     already wrapped more than once — e.g. ``'"ERROR"'`` — which costs nothing
     to absorb and keeps the predicate insensitive to how many layers of
-    quoting an aggregator applied. Note this is NOT the JSON double-encoding
-    case: a truly double-encoded ``raw`` arrives backslash-escaped
-    (``"\\"ERROR\\""``), which :func:`_openrouter_upstream_text` has already
-    parsed away by the time this sees it, so that input never reaches the
-    loop.
+    quoting an aggregator applied.
+
+    A JSON-encoded ``raw`` whose inner value is itself a string (the
+    double-encoding case) DOES reach this function. :func:`_openrouter_upstream_text`
+    ``json.loads`` it, then — because the inner is a ``str``, not a
+    ``Mapping`` — returns the original unparsed ``raw`` rather than the
+    decoded inner. The peel then takes matched quote pairs off that original.
+    (A ``Mapping`` inner is unwrapped to its ``message`` instead, and never
+    gets here.) The earlier claim that double-encoding is "parsed away before
+    this sees it" was wrong; the behaviour was always this, only the stated
+    reason was not.
 
     ``_MAX_QUOTE_PEELS`` bounds the loop because slicing copies the string on
     every pass: an adversarial body of a few MB of quotes would otherwise cost
--- a/local_operator/providers/clients.py
+++ b/local_operator/providers/clients.py
@@ -267,6 +267,46 @@ def _openrouter_upstream_text(error: Mapping[str, Any]) -> str:
     return raw.strip()[:500]
 
 
+#: Aggregators answer an UPSTREAM provider failure with an HTTP 400 whose body
+#: names nothing: the outer message is exactly "Provider returned error" and
+#: ``metadata.raw`` holds a bare sentinel instead of the origin provider's
+#: diagnostics. Session e13d092c093c recorded this shape intermittently on a
+#: request that live probes at ~750k tokens answered 200 seconds later — an
+#: upstream blip wearing a 400. Compared case-insensitively because the casing
+#: on the wire ("Provider returned error", raw "ERROR") is the aggregator's,
+#: not part of the signal.
+_OPAQUE_AGGREGATOR_MESSAGE = "provider returned error"
+
+#: Raw bodies short enough to prove nobody tried to say anything. A real
+#: upstream body — "context length exceeded", a JSON error object — is
+#: actionable and must keep its ``request`` classification.
+_OPAQUE_RAW_SENTINELS = frozenset({"error"})
+
+
+def _is_opaque_aggregator_400(error: Mapping[str, Any]) -> bool:
+    """An aggregator 400 carrying NO actionable upstream diagnostics.
+
+    OpenRouter forwards an upstream failure as ``{"message": "Provider
+    returned error", "metadata": {"raw": ...}}``. When ``raw`` holds real
+    diagnostics the composed message names the actual problem and the 400 is
+    an answer — kind ``request``, no retry. When ``raw`` is a bare sentinel
+    (observed: ``"ERROR"``, from a provider named "Stealth"), the body says
+    nothing a caller could fix, and the status line is the aggregator's relay
+    choice rather than evidence the request was wrong. Those are classified
+    transient so the failover cascade runs; the driver's server-fault budget
+    bounds the retries, and a genuinely broken request just burns a little of
+    it before surfacing — strictly better than aborting a turn that rotation
+    or a fallback could have served.
+    """
+    message = _first_text(error.get("message"))
+    if message.lower() != _OPAQUE_AGGREGATOR_MESSAGE:
+        return False
+    upstream = _openrouter_upstream_text(error)
+    # Strip quote pairs too: ``raw`` is defined as a JSON-encoded string, so a
+    # sentinel can arrive as ``"\"ERROR\""`` and parse straight back to quotes.
+    return upstream.strip().strip("\"'").lower() in _OPAQUE_RAW_SENTINELS
+
+
 def _compat_stream_error(chunk: Mapping[str, Any]) -> ProviderError:
     """An in-band mid-stream failure on an OpenAI-compatible stream.
 
@@ -430,6 +470,13 @@ def raise_for_status(response: httpx.Response) -> None:
     # 408/504 are the two "ran out of time" statuses; 429 and 5xx are the
     # classic retryables. Everything else in 4xx is an answer, not a blip.
     retryable = status == 429 or status >= 500 or status in (408, 504)
+    # ...except an aggregator 400 whose body carries no usable diagnostics:
+    # that is an upstream failure relayed under a 400, not a request the
+    # provider read and refused, so it must reach the failover cascade. See
+    # :func:`_is_opaque_aggregator_400` for the shape and the evidence.
+    error = payload.get("error") if isinstance(payload, Mapping) else None
+    if status == 400 and isinstance(error, Mapping) and _is_opaque_aggregator_400(error):
+        retryable = True
     raise ProviderError(
         status,
         _extract_error_message(response, payload),
--- a/local_operator/providers/clients.py
+++ b/local_operator/providers/clients.py
@@ -282,6 +282,12 @@ _OPAQUE_AGGREGATOR_MESSAGE = "provider returned error"
 #: actionable and must keep its ``request`` classification.
 _OPAQUE_RAW_SENTINELS = frozenset({"error"})
 
+#: How many matched quote pairs :func:`_strip_quote_pair` will peel. Real
+#: bodies use at most one or two layers; the cap exists so a body made of
+#: nothing but quotes cannot spend quadratic CPU (each peel copies the string)
+#: on a request that is already failing.
+_MAX_QUOTE_PEELS = 4
+
 
 def _strip_quote_pair(text: str) -> str:
     """Peel MATCHED surrounding quote pairs off ``text``.
@@ -293,11 +299,24 @@ def _strip_quote_pair(text: str) -> str:
     so pairs are peeled one at a time and an unbalanced body keeps its quotes
     and stays ``request``.
 
-    A pair is peeled repeatedly because ``raw`` is a JSON-encoded string that
-    has occasionally been double-encoded upstream (``"\\"ERROR\\""`` parses back
-    to a quoted quote).
+    The peel LOOPS rather than running once to cover a value that arrives
+    already wrapped more than once — e.g. ``'"ERROR"'`` — which costs nothing
+    to absorb and keeps the predicate insensitive to how many layers of
+    quoting an aggregator applied. Note this is NOT the JSON double-encoding
+    case: a truly double-encoded ``raw`` arrives backslash-escaped
+    (``"\\"ERROR\\""``), which :func:`_openrouter_upstream_text` has already
+    parsed away by the time this sees it, so that input never reaches the
+    loop.
+
+    ``_MAX_QUOTE_PEELS`` bounds the loop because slicing copies the string on
+    every pass: an adversarial body of a few MB of quotes would otherwise cost
+    quadratic time on a request that is already failing. Nothing legitimate
+    nests quotes more than a couple of layers deep, so a low cap costs real
+    traffic nothing and denies a hostile aggregator the CPU.
     """
-    while len(text) >= 2 and text[0] == text[-1] and text[0] in "\"'":
+    for _ in range(_MAX_QUOTE_PEELS):
+        if not (len(text) >= 2 and text[0] == text[-1] and text[0] in "\"'"):
+            break
         text = text[1:-1].strip()
     return text
 
--- a/local_operator/providers/clients.py
+++ b/local_operator/providers/clients.py
@@ -283,6 +283,25 @@ _OPAQUE_AGGREGATOR_MESSAGE = "provider returned error"
 _OPAQUE_RAW_SENTINELS = frozenset({"error"})
 
 
+def _strip_quote_pair(text: str) -> str:
+    """Peel MATCHED surrounding quote pairs off ``text``.
+
+    ``str.strip('"\'')`` would take any leading/trailing run of either
+    character, so unbalanced junk like ``\'\'\'ERROR""`` would reduce to the
+    sentinel and be treated as no-information. That only ever widens the match,
+    but the widening should be a decision rather than an accident of the API —
+    so pairs are peeled one at a time and an unbalanced body keeps its quotes
+    and stays ``request``.
+
+    A pair is peeled repeatedly because ``raw`` is a JSON-encoded string that
+    has occasionally been double-encoded upstream (``"\\"ERROR\\""`` parses back
+    to a quoted quote).
+    """
+    while len(text) >= 2 and text[0] == text[-1] and text[0] in "\"'":
+        text = text[1:-1].strip()
+    return text
+
+
 def _is_opaque_aggregator_400(error: Mapping[str, Any]) -> bool:
     """An aggregator 400 carrying NO actionable upstream diagnostics.
 
@@ -293,18 +312,38 @@ def _is_opaque_aggregator_400(error: Mapping[str, Any]) -> bool:
     (observed: ``"ERROR"``, from a provider named "Stealth"), the body says
     nothing a caller could fix, and the status line is the aggregator's relay
     choice rather than evidence the request was wrong. Those are classified
-    transient so the failover cascade runs; the driver's server-fault budget
-    bounds the retries, and a genuinely broken request just burns a little of
-    it before surfacing — strictly better than aborting a turn that rotation
-    or a fallback could have served.
+    transient so the failover cascade runs.
+
+    Two deliberate widenings, both toward "treat no-information as transient":
+
+    - The outer message is matched after trimming and case-folding, so
+      ``"Provider returned error "`` matches too. The casing and padding on the
+      wire are the aggregator's formatting, not part of the signal.
+    - :func:`_openrouter_upstream_text` unwraps a nested ``error`` object, so
+      ``{"error": {"message": "ERROR"}}`` resolves to the same bare sentinel and
+      also matches. A body that structurally tried to say something but whose
+      content is still the word "error" carries no more actionable information
+      than the flat sentinel. This is the case most likely to shadow a real
+      upstream error whose message happens to be that single word; the cost of
+      that collision is bounded retries, and the cost of the opposite mistake
+      is a dead turn.
+
+    The price is latency, not correctness, and it is not small: a genuinely
+    broken request now saturates the driver's server-fault budget (12 requests
+    per target) before surfacing. Measured against the default 500ms base delay
+    and the 8s backoff cap, that is 64-76s of sleep on a single target and
+    roughly 4-5 minutes across a 3-target fallback chain. A user hitting a
+    PERSISTENT aggregator 400 therefore waits minutes for a failure that
+    previously surfaced immediately. That is the accepted trade: this shape is
+    by construction unattributable, so the alternative is killing a turn that
+    rotation or a fallback could have served — but it is a real cost, not the
+    rounding error an earlier draft of this docstring implied.
     """
     message = _first_text(error.get("message"))
     if message.lower() != _OPAQUE_AGGREGATOR_MESSAGE:
         return False
     upstream = _openrouter_upstream_text(error)
-    # Strip quote pairs too: ``raw`` is defined as a JSON-encoded string, so a
-    # sentinel can arrive as ``"\"ERROR\""`` and parse straight back to quotes.
-    return upstream.strip().strip("\"'").lower() in _OPAQUE_RAW_SENTINELS
+    return _strip_quote_pair(upstream.strip()).lower() in _OPAQUE_RAW_SENTINELS
 
 
 def _compat_stream_error(chunk: Mapping[str, Any]) -> ProviderError:
@@ -333,6 +372,7 @@ def _compat_stream_error(chunk: Mapping[str, Any]) -> ProviderError:
     error = chunk.get("error")
     status: int | None = None
     error_type = ""
+    opaque_aggregator_400 = False
     if isinstance(error, Mapping):
         message = _first_text(error.get("message"), error.get("detail"), error.get("msg"))
         upstream = _openrouter_upstream_text(error)
@@ -348,6 +388,13 @@ def _compat_stream_error(chunk: Mapping[str, Any]) -> ProviderError:
         metadata = error.get("metadata")
         if isinstance(metadata, Mapping):
             error_type = str(metadata.get("error_type") or "")
+        # The SAME opaque relay body arrives on this channel whenever the
+        # gateway had already committed HTTP 200 before the upstream failed, so
+        # it has to classify the same way here as in `raise_for_status` -- a
+        # turn must not die on the relay channel the aggregator happened to
+        # pick. Judged on the chunk's `code` rather than a status line, because
+        # in-band that is where the 400 lives.
+        opaque_aggregator_400 = status == 400 and _is_opaque_aggregator_400(error)
     else:
         message = _first_text(error)
     if not message:
@@ -357,7 +404,7 @@ def _compat_stream_error(chunk: Mapping[str, Any]) -> ProviderError:
     return ProviderError(
         status,
         _capped(message),
-        retryable=status is None or status == 429 or status >= 500,
+        retryable=(opaque_aggregator_400 or status is None or status == 429 or status >= 500),
         auth_error=status in (401, 403),
     )
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "local-operator"
-version = "0.44.5"
+version = "0.44.6"
 description = "A team of proactive AI assistants that can work together behind the scenes to help you get mundane tasks done so you can focus on the fun stuff."
 readme = "README.md"
 authors = [{ name = "Damian Tran", email = "damian@gominerva.com" }]
--- a/tests/unit/providers/test_clients.py
+++ b/tests/unit/providers/test_clients.py
@@ -1388,6 +1388,87 @@ class TestErrorMessageExtraction:
         assert error.kind == "transient"
 
 
+class TestOpaqueAggregator400:
+    """An aggregator relays an UPSTREAM failure as an HTTP 400 whose body names
+    nothing. Session e13d092c093c recorded the shape — ``message`` exactly
+    "Provider returned error", ``metadata.raw`` a bare "ERROR", provider_name
+    "Stealth" — arriving intermittently on a request that live probes at ~750k
+    tokens answered 200 seconds later, and again during the investigation.
+    Classified ``request``, it aborted the turn before rotation or fallback
+    could serve it; these tests pin both sides of the new line: the opaque
+    sentinel is transient/retryable, every 400 with real diagnostics stays an
+    answer."""
+
+    def _error(self, response: httpx.Response) -> ProviderError:
+        with pytest.raises(ProviderError) as excinfo:
+            raise_for_status(response)
+        return excinfo.value
+
+    @staticmethod
+    def _openrouter_body(raw: str, provider: str = "Stealth") -> dict[str, Any]:
+        return {
+            "error": {
+                "message": "Provider returned error",
+                "code": 400,
+                "metadata": {"raw": raw, "provider_name": provider},
+            }
+        }
+
+    def test_the_reported_sentinel_is_transient(self) -> None:
+        """The exact wire body from the session: bare "ERROR" in ``raw``."""
+        error = self._error(httpx.Response(400, json=self._openrouter_body("ERROR")))
+        assert (error.kind, error.retryable) == ("transient", True)
+        # The frame still shows what the wire said — no invented diagnostics.
+        assert error.message == "Provider returned error: ERROR"
+
+    def test_a_quoted_json_string_sentinel_is_transient_too(self) -> None:
+        """``raw`` is defined as JSON-encoded, so the sentinel can arrive as
+        ``\"\\\"ERROR\\\"\"`` and parse straight back to quotes."""
+        error = self._error(httpx.Response(400, json=self._openrouter_body(json.dumps("ERROR"))))
+        assert (error.kind, error.retryable) == ("transient", True)
+
+    def test_actionable_upstream_diagnostics_stay_request(self) -> None:
+        """A ``raw`` holding the origin provider's REAL body is an answer, not
+        weather: the same outer message with real diagnostics keeps kind
+        ``request`` so a genuinely broken request is not retried."""
+        body = self._openrouter_body(
+            json.dumps({"error": {"message": "context length exceeded"}}),
+            provider="Google",
+        )
+        error = self._error(httpx.Response(400, json=body))
+        assert (error.kind, error.retryable) == ("request", False)
+        assert "context length exceeded" in error.message
+
+    def test_real_text_in_raw_stays_request(self) -> None:
+        """A non-JSON ``raw`` that still SAYS something is actionable."""
+        error = self._error(
+            httpx.Response(
+                400,
+                json=self._openrouter_body("This model's maximum context length is 16385 tokens"),
+            )
+        )
+        assert (error.kind, error.retryable) == ("request", False)
+
+    def test_direct_provider_400s_are_untouched(self) -> None:
+        """Only the aggregator's exact outer message qualifies; ordinary 400s
+        keep their classification whatever their metadata holds."""
+        plain = self._error(httpx.Response(400, json={"error": {"message": "bad field"}}))
+        assert (plain.kind, plain.retryable) == ("request", False)
+        other_message = self._error(
+            httpx.Response(
+                400,
+                json={
+                    "error": {
+                        "message": "Provider error",
+                        "code": 400,
+                        "metadata": {"raw": "ERROR", "provider_name": "Stealth"},
+                    }
+                },
+            )
+        )
+        assert (other_message.kind, other_message.retryable) == ("request", False)
+
+
 class TestRetryAfter:
     """ "try again in 40s" is the single most actionable fact in a rate-limit
     error, and providers disagree about where to put it."""
--- a/tests/unit/providers/test_clients.py
+++ b/tests/unit/providers/test_clients.py
@@ -1468,6 +1468,93 @@ class TestOpaqueAggregator400:
         )
         assert (other_message.kind, other_message.retryable) == ("request", False)
 
+    def test_nested_error_object_sentinel_is_transient(self) -> None:
+        """``_openrouter_upstream_text`` unwraps a nested ``error`` object, so a
+        structurally-richer body whose CONTENT is still the bare word carries no
+        more information than the flat sentinel and matches too. Documented as a
+        deliberate widening rather than left to chance."""
+        body = self._openrouter_body(json.dumps({"error": {"message": "ERROR"}}))
+        error = self._error(httpx.Response(400, json=body))
+        assert (error.kind, error.retryable) == ("transient", True)
+
+    def test_unbalanced_quote_runs_stay_request(self) -> None:
+        """Only MATCHED quote pairs are peeled. ``str.strip`` would take any run
+        of either character and reduce this to the sentinel; an unbalanced body
+        is malformed rather than empty, so it keeps its ``request`` answer."""
+        error = self._error(httpx.Response(400, json=self._openrouter_body("'''ERROR\"\"")))
+        assert (error.kind, error.retryable) == ("request", False)
+
+    def test_opaque_5xx_shape_is_unchanged(self) -> None:
+        """The predicate only widens 400s: a 5xx carrying the same opaque body
+        was already retryable by status and must stay that way."""
+        body = self._openrouter_body("ERROR")
+        body["error"]["code"] = 502
+        error = self._error(httpx.Response(502, json=body))
+        assert (error.kind, error.retryable) == ("transient", True)
+
+
+class TestOpaqueAggregator400InBand:
+    """The aggregator's OTHER relay channel for the identical upstream failure.
+
+    Once the gateway has committed HTTP 200 it cannot use the status line, so it
+    delivers the same opaque body as an in-band error chunk (the shape
+    ``test_openai_compat_mid_stream_error_chunk_unwraps_upstream_raw`` pins for a
+    502). Classified from ``code`` alone that body was ``request``/not-retryable
+    and killed the turn, so the fix had to reach both channels: which relay the
+    aggregator picks is not something the caller can influence, and the two must
+    agree."""
+
+    @staticmethod
+    def _chunk(raw: str, code: int = 400) -> dict[str, Any]:
+        return {
+            "id": "gen-1",
+            "error": {
+                "code": code,
+                "message": "Provider returned error",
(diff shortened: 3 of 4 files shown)


damianvtran:
Note for reviewers: a local run of the full unit suite showed `tests/unit/tools/test_eval_tool.py::test_background_streams_output_while_it_runs` failing intermittently under load. Reproduced identically on a pristine `origin/main` worktree (no changes) — pre-existing timing-sensitive flake, untouched by this diff. CI test jobs (3.12/3.13) are green on this branch.

damianvtran:
### Agent review — round 1
Reviewer: reviewer subagent (claude-opus-5), independent of implementing agent
Scope: 7f3bf65485728451716b9433a8952d97fdc3a594..221fb5c57dfc61fbe1bbbe758b86a83f07f9f2a9

Reviewed the rebased head, not the PR body's pre-rebase description. I re-derived the retry-budget bound by executing the current driver rather than reading it, since the body's loop-safety argument is 134 commits stale.

**C-risk verdict up front: the bound genuinely holds on current main.** I drove `stream_with_failover` with a client that returns the session's exact opaque 400 forever, sweeping pool size {1,2,4,8} x chain length {0,1,3} x `maxRetries` {2,4,10} — 36 configurations. Every target saturates at exactly 12 requests and the walk terminates; total requests never exceed `12 x (1 + len(chain))`, independent of pool size and independent of `maxRetries`. `maxRetries: 10` with an 8-key pool and a 3-target chain lands on 48 requests and stops, identical to `maxRetries: 2`. The ceiling comes from `_server_fault_budget_spent` (failover.py:2566) being consulted before rotation as well as before same-credential retry, plus `server_faults_by_target` being carried across loop-back sweeps (failover.py:1802). No infinite loop, no pool-size multiplication, no unbounded path. The `transient` classification also correctly reaches `is_server_side_failure`, so the credential is deprioritized for a 120s TTL rather than blocked — right call, since the credential is not at fault.

**C1 — MAJOR — the sibling relay channel is left unfixed, and the codebase's own comments say it is the one that matters.** `raise_for_status` handles the aggregator's *pre-stream* relay. OpenRouter's *other* relay for the identical upstream failure is in-band on an HTTP 200 stream, mapped by `_compat_stream_error` (clients.py:310) — whose own docstring describes exactly this: "Once a gateway has committed HTTP 200 it can no longer signal an upstream failure via the status line, so OpenRouter delivers it INSIDE the stream". That function computes `retryable` from `status` alone (clients.py:359: `retryable=status is None or status == 429 or status >= 500`) and never consults `_is_opaque_aggregator_400`. Executed against the identical wire body:

```
raise_for_status  : kind=transient retryable=True   <- fixed
_compat_stream_error: kind=request  retryable=False  <- NOT fixed
```

I then drove the full cascade with the opaque error arriving as the first stream chunk (so `forwarded_any` is still False and failover is genuinely possible at that point) — result: `ABORTED: invalid request (HTTP 400): Provider returned error: ERROR | fallback calls=0 rotations=[]`. The turn dies exactly as the bug report describes, on a path the PR leaves untouched. The repo already carries a test for the sibling case at test_clients.py:290 (`..._unwraps_upstream_raw`, a 502 in-band chunk with `metadata.raw`), so this channel is known-live, not hypothetical. It is `chunk["error"]["code"]: 400` that makes it reachable — a 502 in that slot already retries. One line: have `_compat_stream_error` apply the same predicate when `status == 400`. Not a blocker because the fixed path is the one the reported session hit and the change is a strict improvement, but the PR's premise ("classified request, it aborted the turn") remains true on the sibling path after merge.

**C2 — MAJOR — the version bump is stale; 0.43.19 is already behind main and the PR does not merge.** The rebase resolved the bump to 0.43.19 against a main that was 0.43.18, but main is now `b8cd5336` at **0.44.0**, and 0.44.0 is already tagged, released, and live on PyPI. `gh pr view 288` reports `mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`. This needs a re-rebase onto current origin/main with the version resolved to **0.44.1** (patch: a self-contained reliability fix, no step-function capability). Flagging as MAJOR rather than a nit because merging as-is is mechanically impossible and the released-version collision would be a genuine publish defect if forced.

**C3 — MINOR — `.strip("\"'")` strips character sets, not quote pairs, contradicting its own comment.** clients.py:305 says "Strip quote pairs too", but `str.strip` removes *any* leading/trailing run of those characters, unpaired and mixed. Verified: `raw = "'''ERROR\"\""` classifies `transient`. Harmless in practice (it only ever widens the match to more no-information bodies, which is the intended direction) but the comment overstates the precision, and this project holds comments to a why/constraint standard. Either `json.loads`-shaped unwrapping or a comment saying "any surrounding quote characters" would be accurate.

**C4 — MINOR — the sentinel test passes through a nested-`error` unwrap the tests do not pin.** `_openrouter_upstream_text` resolves `{"error": {"message": "ERROR"}}` and `{"error": "ERROR"}` down to the bare string `ERROR`, so both classify `transient` — a body that structurally *did* try to say something, unlike the flat sentinel. Verified both. This is defensible (the content is still no information) but it is a widening the PR neither documents nor tests, and it is the case most likely to shadow a real upstream error whose message happens to be the word "error". Worth one line in the docstring.

**C5 — MINOR — worst-case added latency is not stated anywhere.** With the default 500ms base and the 8s backoff cap, saturating 12 fault requests on one target costs ~64s of sleep; a 3-target chain reaches ~254s before the turn surfaces. That is the price of the fix and it is the right trade, but a user hitting a persistent aggregator 400 now waits ~4 minutes for a failure that previously surfaced immediately. The docstring's "burns a little of it before surfacing" undersells this by an order of magnitude; a concrete number belongs there.

**C6 — NIT — redundant local import.** test_failover.py:1019 does `import httpx as _httpx` inside the test while the module already imports `httpx` at line 15. The surrounding tests use the module-level name.

**C7 — NIT — trailing whitespace in the outer message still matches.** `"Provider returned error "` classifies `transient` because `_first_text` strips. Correct and intended; noting only that the docstring says "exactly", which is now three qualifiers away from literal.

Verified clean, no finding: sentinel matching does not over-match on the cases that would be dangerous — absent `raw`, absent `metadata`, empty/whitespace `raw`, `"null"`, `"{}"`, real JSON diagnostics, substantive prose, `"Error: invalid api key format"`, and any different outer message all stay `request`/`retryable=False`. All four named storm-guard tests still exist (test_failover.py:887, 924, 955, 983) and pass unchanged; the 400-abort guard at failover.py:2154 is intact and still gated on `kind == "request" and is_primary`, so this change routes around it by classification rather than by weakening it — the right shape. Touched suites: 338 passed.

**Verdict: 2 MAJOR (C1, C2) — not terminal.** C2 blocks the merge mechanically and must be re-rebased to 0.44.1 regardless. C1 is a judgement call I will not overstate: the PR is a strict improvement and fixes the reported path, so shipping it and closing the in-band channel in an immediate follow-up is a legitimate answer — but leaving it unaddressed means the same upstream blip still kills a turn whenever the aggregator relays it after committing 200. Retry budget itself is sound; no blocker.

damianvtran:

Reviewed head at the time of review: `221fb5c5`. New head: `f3664a97` (rebased onto `b8cd5336`).

- **C1 (MAJOR) — fixed in `f3664a97`.** `_compat_stream_error` now applies `_is_opaque_aggregator_400` when the chunk's `code` is 400, so both relay channels classify the identical body identically. New test `test_opaque_aggregator_400_in_band_fails_over_instead_of_aborting` (test_failover.py) drives it end to end through a real `OpenAICompatClient` over a mock transport, and `TestOpaqueAggregator400InBand` (test_clients.py, modelled on the existing 502 in-band test) pins the sentinel / real-diagnostics / different-outer-message cases.
- **C2 (MAJOR) — fixed in `f3664a97`.** Re-rebased onto current `origin/main` (`b8cd5336`, 0.44.0); version resolved to **0.44.1** (patch — self-contained reliability fix, no step-function capability). pyproject.toml's version line was the only conflict, as predicted. `gh pr view 288` now reports `mergeable: MERGEABLE`.
- **C3 (MINOR) — fixed in `f3664a97`.** Replaced `.strip("\"'")` with `_strip_quote_pair`, which peels only MATCHED pairs (repeatedly, for double-encoded `raw`). `"'''ERROR\"\""` now stays `request`/not-retryable and is pinned by `test_unbalanced_quote_runs_stay_request`. Chose the tighter behaviour over the corrected comment: the widening should be a decision, not an artifact of `str.strip`'s API.
- **C4 (MINOR) — fixed in `f3664a97`.** Widening is wanted and now documented explicitly in the docstring (a body whose content is still the bare word carries no more actionable information than the flat sentinel, and the collision cost is bounded retries versus a dead turn). Pinned by `test_nested_error_object_sentinel_is_transient`.
- **C5 (MINOR) — fixed in `f3664a97`.** Docstring now states the measured cost: 12 requests per target, **64–76s** of sleep on a single target and **~4–5 minutes** across a 3-target chain. I re-measured rather than copying the figure — summing `backoff_delay_ms(i, 500)` over 11 sleeps across 2000 samples gives min 63.8s / mean 70.0s / max 76.1s, so ~64s is the no-jitter floor and the range is the honest statement.
- **C6 (NIT) — fixed in `f3664a97`.** Function-local `import httpx as _httpx` removed; the test uses the module-level `httpx`.
- **C7 (NIT) — fixed in `f3664a97`.** The docstring no longer claims "exactly"; it now names both widenings (trim + case-fold on the outer message, nested-`error` unwrap on `raw`) as deliberate.

---

#### C1 evidence — before/after, real execution

Driver: the opaque body as the **first** in-band chunk on an HTTP 200 stream (so `forwarded_any` is still False and failover is genuinely possible), through a real `OpenAICompatClient` + `stream_with_failover`, 2-key pool + 1-target fallback chain.

**BEFORE** (at `b05406ea`, the cherry-picked PR commit):

```
raise_for_status   : kind=transient retryable=True
_compat_stream_error: kind=request   retryable=False
RESULT: ABORTED: request (HTTP 400): invalid request (HTTP 400): Provider returned error: ERROR | primary requests=1 | fallback calls=0 rotations=[]
```

**AFTER** (`f3664a97`):

```
raise_for_status   : kind=transient retryable=True
_compat_stream_error: kind=transient retryable=True
RESULT: SERVED by fallback=True | primary requests=6 | fallback calls=1 rotations=[('openai', 'k1'), ('openai', 'k2')]
```

Both sibling credentials are exercised and the chain serves the turn, instead of the turn dying on chunk one.

#### Classification matrix — both channels, at `f3664a97`

```
case                                          pre-stream 400             in-band chunk
----------------------------------------------------------------------------------------------
opaque sentinel ERROR                         kind=transient retryable=True   kind=transient retryable=True
quoted sentinel                               kind=transient retryable=True   kind=transient retryable=True
nested error obj sentinel (C4 widening)       kind=transient retryable=True   kind=transient retryable=True
unbalanced quote run (C3)                     kind=request   retryable=False  kind=request   retryable=False
real upstream diagnostics                     kind=request   retryable=False  kind=request   retryable=False
substantive prose                             kind=request   retryable=False  kind=request   retryable=False
different outer message                       kind=request   retryable=False  kind=request   retryable=False
direct-provider 400 (no metadata)             kind=request   retryable=False  kind=request   retryable=False
empty raw                                     kind=request   retryable=False  kind=request   retryable=False
opaque shape on 502                           kind=transient retryable=True   kind=transient retryable=True
```

The two channels now agree on every case, and the 5xx shape is unchanged.

#### Retry budget still terminates on the newly-widened in-band path

Re-ran your sweep against the **in-band** channel (opaque error returned forever): pool {1,2,4,8} × chain {0,1,3} × `maxRetries` {2,4,10} = 36 configurations. Every one terminates; max requests per target never exceeds **12**, and the total never exceeds **12 × (1 + len(chain))**, independent of pool size and `maxRetries`. Worst cases:

```
 pool  chain  maxRetries   total   bound  max/target
    4      0          10      12      12          12  OK
    8      1          10      24      24          12  OK
    8      3          10      48      48          12  OK
```

The C1 change routes the in-band failure into the same budgeted path, so the bound you derived holds there too.

#### Gates — whole tree, exactly as CI runs them

```
flake8 .                        clean
black==26.1.0 --check .         659 files unchanged
isort==5.13.2 --check .         clean
pyright                         0 errors, 0 warnings, 0 informations
pytest tests/unit               9319 passed, 15 skipped in 305.13s
```

The `test_background_streams_output_while_it_runs` flake noted earlier did not reproduce in this run; the suite is fully green.

damianvtran:
Heads-up on a version collision, from the #455/#471 release lane.

This PR carries a `0.44.1 → 0.44.2` bump in `pyproject.toml`. **#471 is about to merge with that same 0.44.2 bump**, so 0.44.2 will be spent and tagged before this lands.

When you rebase, git will conflict on that line rather than silently mis-publish — so this is a cost, not a hazard. Please **re-bump to the next unspent patch (0.44.3 or later, whatever is free at the time)** rather than accepting either side of the conflict.

Context: 0.44.1 was tagged and published from `f82c2f5e` while #455 was in review, which stranded #455's merge (`168efe48`) outside any released tag. #471 is the bump that lets it ship.

Worth knowing generally: this version race cost #455 four rebases. The durable fix is **bumping the version as the last action before merge** rather than at the start of a review round — it costs nothing and removes the collision entirely.

(Flagged by an agent review round on #471. No action needed until you rebase.)

damianvtran:
### Agent review — round 2
Reviewer: reviewer subagent (claude-opus-5), independent of implementing agent
Scope: 221fb5c57dfc61fbe1bbbe758b86a83f07f9f2a9..880f6fff92abac0f1b8d9d72b958d6ef98ce22fc

Verification round against head `880f6fff` (rebased onto `f82c2f5e`, 0.44.2), in my own worktree. Every claim below I re-derived by execution — I did not take the remediation comment's numbers. Where I could, I mutation-tested the new tests: reverting the production hunk and confirming the test actually goes red, because a test that passes with the fix removed pins nothing.

**All seven round-1 findings verified fixed.** One new MINOR, one new NIT.

#### C1 (MAJOR) — VERIFIED FIXED

Drove the full cascade with the opaque body as the **first** in-band chunk on an HTTP 200 stream, through a real `OpenAICompatClient` over a mock transport, 2-key pool + 1-target chain. Three trees, same driver, each resolving `local_operator` from its own worktree:

```
BASE   f82c2f5e  raise_for_status: request/False   in-band: request/False
                 -> ABORTED: invalid request (HTTP 400) | primary=1 fallback=0 rotations=[]
BEFORE fe3ba4dc  raise_for_status: transient/True  in-band: request/False
                 -> ABORTED: invalid request (HTTP 400) | primary=1 fallback=0 rotations=[]
AFTER  880f6fff  raise_for_status: transient/True  in-band: transient/True
                 -> SERVED by fallback=True | primary=6 fallback=1
                    rotations=[('openai','k1'), ('openai','k2')]
```

The pre-fix abort is real, not asserted: at `fe3ba4dc` the turn dies on chunk one with zero fallback calls and zero rotations, exactly as C1 described. At head both sibling credentials are exercised and the chain serves the turn.

The new tests genuinely pin it. With head's tests but `clients.py` reverted to `fe3ba4dc`: `TestOpaqueAggregator400InBand::test_in_band_sentinel_is_transient` fails (`transient/True` → `request/False`) and `test_opaque_aggregator_400_in_band_fails_over_instead_of_aborting` fails with the abort — `2 failed, 2 passed`. The two that still pass are the negative cases, which is correct.

#### C2 — VERIFIED FIXED

`git merge-base HEAD origin/main` = `f82c2f5e` = current `origin/main`, so the branch is on top of live main and main has not moved. PyPI's JSON API lists `0.44.0` and `0.44.1` only; **0.44.2 is not taken**, and there is no local `v0.44.2` tag. `mergeable: MERGEABLE`. Patch is the right bump — a self-contained reliability fix, no step-function capability.

#### C3 — VERIFIED FIXED, probed adversarially

26 inputs through `_strip_quote_pair`, including the ones most likely to break it:

```
''  ' '  '"'  "'"        -> unchanged, no match       (len<2 guard holds; no IndexError)
'""'  "''"  '"   "'      -> ''                        -> no match (empty != sentinel)
'"\''  '\'"'             -> unchanged, no match       (mismatched pair not peeled)
"'''''"                  -> "'"                       (odd run leaves one, no match)
'"ERROR'  'ERROR"'       -> unchanged, no match       (one-sided)
"'''ERROR\"\""           -> unchanged, no match       <- the C3 case, now request
'"ERROR"'  '""ERROR""'   -> 'ERROR'                   -> match (intended)
'"  ERROR  "'            -> 'ERROR'                   -> match (inner strip)
'"a"'                    -> 'a'                       -> no match (not the sentinel)
```

No exception on any input, terminates on all of them (the loop strictly shortens). The C3 tightening is pinned: swapping the body back to `text.strip("\"'")` turns `test_unbalanced_quote_runs_stay_request` red.

One correction to the docstring, filed as C8 below: the double-encoding sentence it gives as the reason for the loop is not what the code does.

#### C4 — VERIFIED FIXED

The widening is documented at length in `_is_opaque_aggregator_400` (including the honest statement that this is the case most likely to shadow a real single-word upstream error, and why the collision cost is the better one to pay). Pinned: blocking the nested unwrap — reading `metadata.raw` directly instead of through `_openrouter_upstream_text` — turns `test_nested_error_object_sentinel_is_transient` red. Both `{"error": {"message": "ERROR"}}` and `{"error": "ERROR"}` classify transient on both channels; `{"error": {"message": "context length exceeded"}}` stays `request`.

#### C5 — VERIFIED FIXED, and the number is right

Re-measured independently rather than reading it: summing `backoff_delay_ms(i, 500)` over the 11 sleeps of a saturated 12-request target, 2000 samples, against the live `BACKOFF_CAP_MS = 8000` — **min 64.0s / mean 70.0s / max 76.2s** on one target, **4.3–5.1 min** across a 3-target chain. The docstring's "64-76s" and "roughly 4-5 minutes" are both accurate, and it says plainly that a user hitting a persistent aggregator 400 now waits minutes for a failure that previously surfaced immediately. That is the honest statement C5 asked for.

#### C6 — VERIFIED FIXED

No `import httpx` added anywhere in the diff (`git diff | grep -c '^+.*import httpx'` = 0); the new tests use the module-level `httpx` from line 15. The remaining function-local `import httpx` at test_failover.py:2655 is pre-existing and outside this diff.

#### C7 — VERIFIED FIXED

`_is_opaque_aggregator_400`'s docstring names both widenings ("Two deliberate widenings, both toward 'treat no-information as transient'") and claims no exactness. The word survives only in the module-level `_OPAQUE_AGGREGATOR_MESSAGE` comment describing the *wire shape* ("the outer message is exactly 'Provider returned error'"), which is a true statement about what the aggregator sends, not about what the predicate matches. Correct as written.

---

#### (a) Classification matrix — re-derived, both channels

24 bodies through `raise_for_status` and `_compat_stream_error` side by side. **Identical on every case.** Matches (all `transient/True` on both): reported sentinel, lowercase `error`, JSON-quoted sentinel, nested `{"error":{"message":"ERROR"}}`, nested `{"error":"ERROR"}`, outer message with trailing space, outer message upper-cased. Stays `request/False` on both: unbalanced quote run, real upstream JSON diagnostics, substantive prose, different outer message, no metadata, metadata without `raw`, empty `raw`, whitespace `raw`, `"null"`, `"{}"`, `"Error: invalid api key format"`, `"errors"`, `"ERROR!"`, plain direct-provider 400, double-encoded sentinel. Opaque shape on 502: `transient/True` on both, unchanged.

No over-match on any real 400. Notably `"errors"` and `"ERROR!"` — one character off the sentinel — both stay `request`, so the frozenset is doing exactly the narrow job it claims.

One apparent mismatch in my first pass was my own artifact: a body with **no** `code` key gives `status=None` in-band, which was already `retryable=True` at base `f82c2f5e` and has nothing to do with this diff. Confirmed against base. Withdrawn.

#### (b) Retry budget — still terminates on BOTH paths

Full sweep, pool {1,2,4,8} × chain {0,1,3} × maxRetries {2,4,10} = 36 configs **per channel**, opaque error returned forever, `baseDelayMs: 0`, 60s watchdog per config:

```
channel      configs  non-terminating  max/target  worst total  bound  violations
pre-stream        36                0          12           48     48           0
in-band           36                0          12           48     48           0
```

Every config terminates. Max requests per target saturates at exactly **12** and never exceeds it; total never exceeds **12 × (1 + len(chain))**, independent of pool size and independent of `maxRetries` (8-key pool + `maxRetries: 10` lands on the same 48 as `maxRetries: 2`). The C1 change routes the in-band failure into the same budgeted path, so the round-1 bound holds there identically. The 400-abort storm guard at failover.py:2172 is intact and still gated on `kind == "request" and is_primary` — this change routes around it by classification, not by weakening it — and all four named storm-guard tests pass.

#### The TUI failures — the implementer's claim is TRUE

Independently checked on a **pristine `origin/main` worktree with no changes**, `tests/unit/tui/test_steering_approval.py`, 5 runs:

```
run 1: 1 failed, 113 passed
run 2: 1 failed, 113 passed
run 3: 1 failed, 113 passed
run 4: 1 failed, 113 passed
run 5: 114 passed
```

Same two tests, alternating: `test_a_held_answer_key_resolves_cleanly_on_every_second_key` ("typing answered the prompt") and `test_typing_a_steer_at_a_live_prompt_never_answers_it`. Both fail on unmodified main and both pass on some runs of the same tree — genuinely pre-existing, order/timing-sensitive, **not caused by this diff**. This is a providers-only diff that does not import or touch the TUI, and the failure set at head is a subset of the failure set on main. Not a finding against this PR; worth someone's attention on main separately.

#### Gates — whole tree, exactly as CI runs them

```
flake8 .                                          clean (rc=0)
black==26.1.0 --check .                           659 files unchanged
isort==5.13.2 --check .                           clean
pyright --pythonpath .venv/bin/python .           0 errors, 0 warnings, 0 informations
pytest tests/unit                                 9329 passed, 15 skipped in 333s
pytest providers/{test_clients,test_failover}     349 passed
```

Full suite green in a single run, including the TUI file. CI: lint, type-check and pip-audit pass; test (3.12)/(3.13) still running at time of writing — confirm green before merge.

---

#### New findings

**C8 — MINOR — `_strip_quote_pair`'s docstring justifies the loop with a case the loop does not handle.** clients.py:290 says the pair is peeled repeatedly "because `raw` is a JSON-encoded string that has occasionally been double-encoded upstream (`"\"ERROR\""` parses back to a quoted quote)". Executed: a genuinely double-encoded `raw` (`json.dumps(json.dumps("ERROR"))`) resolves through `_openrouter_upstream_text` to `'"\\"ERROR\\""'` — the inner quotes arrive **backslash-escaped**, so `text[-1]` is `"` but `text[-2]` is `\`, the peel stops after one round, and the body classifies `request`, not `transient`. Confirmed on both channels; the matrix row above records it. What the loop actually handles is a *literally* multi-quoted raw (`'""ERROR""'` → `ERROR`), which is a different and rarer shape. The behaviour is fine — declining to match an escaped body is the conservative direction, and this is the same "no double-standard" instinct that produced the C3 fix — but the stated reason is wrong, and this repo holds comments to a why/constraint standard precisely so a later reader does not trust a claim like this. One line: say the loop handles a multiply-quoted raw, and that a backslash-escaped double encoding deliberately does not match.

**C9 — NIT — the peel is O(n²) on a pathological raw, newly reachable from a 400.** `text[1:-1]` copies the string each iteration, so a raw of *n* leading and *n* trailing quotes costs *n* copies. Measured at head via `raise_for_status`: a 0.8MB body takes 0.23s, 3.2MB takes 2.49s, and the bare function on a 4MB quote run takes 63s. At base the same bodies cost 0.01s because nothing peeled them. Not a finding I would hold a PR for — it needs a hostile aggregator sending megabytes of quotes under the exact outer message, `_openrouter_upstream_text` caps the non-JSON path at 500 chars so only the JSON-`message` route reaches it, and the cost is bounded CPU on a request that was already failing. Recording it because the diff moves this input from "never inspected" to "inspected", which is the kind of reachability change worth a sentence. A `len(text) < 4096` guard or a two-sided index walk would close it; equally defensible to defer.

---

Nothing else. I re-checked that the predicate cannot fire on an absent `metadata`, an absent `raw`, an empty or whitespace `raw`, `"null"`, `"{}"`, real diagnostics, prose, or any different outer message, and that direct-provider 400s are untouched on both channels. The two channels agree on all 24 probes, the 5xx path is unchanged, the storm guard is intact, and the budget terminates everywhere I could drive it.

**Verdict: TERMINAL — no BLOCKER/MAJOR.** All seven round-1 findings are genuinely fixed and pinned by tests that fail without the fix. C8 and C9 are follow-ups, not a reason for another round; C8 is a one-line comment correction that could ride along if the branch is touched again for any other reason. Merge once `test (3.12)` and `test (3.13)` report green.

damianvtran:

Round 2 verdict was TERMINAL (no blocker/major) with two follow-up findings. Both are addressed here rather than deferred, since C8 left a knowingly-incorrect justification in the tree and C9 is an input this diff is what makes reachable.

- **C8 (MINOR) — fixed in `573f9079`.** The reviewer is right: a genuinely double-encoded `raw` arrives backslash-escaped (`"\"ERROR\""`) and `_openrouter_upstream_text` has already parsed it away before `_strip_quote_pair` runs, so double-encoding never reaches the loop and could not have been its reason. The loop itself is still wanted — it absorbs a value wrapped in several *literal* quote layers — so the behaviour is unchanged and only the stated rationale is corrected, with the JSON case explicitly called out as NOT what it covers.
- **C9 (NIT) — fixed in `573f9079`.** Peel is now bounded by `_MAX_QUOTE_PEELS = 4`. Measured on this machine: a 3.2 MB quote-bomb body went from **2.49s → 0.0006s**. Nothing legitimate nests quotes more than a couple of layers, so the cap costs real traffic nothing.

Verification after the change:

```
-- peel unit --
'ERROR'          -> 'ERROR'
'"ERROR"'        -> 'ERROR'
'"\'ERROR\'"'    -> 'ERROR'
"'''ERROR\"\""   -> "'''ERROR\"\""     (unbalanced still keeps its quotes)
''  -> ''    '"' -> '"'    '""' -> ''

-- classification matrix (unchanged) --
sentinel     transient  True
quoted       transient  True
unbalanced   request    False
real diag    request    False
ERROR!       request    False
direct 400   request    False
```

Gates over the whole tree as CI runs them: `flake8 .` clean, `black==26.1.0 --check .` 659 files unchanged, `isort==5.13.2 --check .` clean, `pyright` 0 errors / 0 warnings / 0 informations, `tests/unit/providers` 685 passed.

**Not addressed (recorded, no ticket per team default):** the two `tests/unit/tui/test_steering_approval.py` timing flakes that round 2 independently reproduced on pristine `origin/main` (4 of 5 runs). Pre-existing and unrelated to this providers-only diff; flagged for separate attention on main.

damianvtran:
### Agent review — round 3
Reviewer: reviewer subagent (anthropic/claude-opus-5), independent of implementing agent
Scope: 880f6fff92abac0f1b8d9d72b958d6ef98ce22fc..573f9079eede97999dd1a72a2cfa69aa59d6aca3

Freshness round on the 27-line C8/C9 remediation. Diff is `_MAX_QUOTE_PEELS = 4`, the `while`→bounded `for`, and the rewritten docstring. Verified in an isolated worktree at `573f9079`; did not touch `~/local-operator`.

## C9 — fixed

`_strip_quote_pair` now peels at most 4 matched pairs. Quote-bomb timing on this machine:

| body | 880f6fff | 573f9079 |
|---|---|---|
| 1 MB of quotes | 6.1 s | 0.8 ms |
| 4 MB of quotes | 109 s | 1.0 ms |
| 16 MB of quotes | (killed at 120 s) | 7.6 ms |
| 4 MB through `_is_opaque_aggregator_400` | — | 0.1 ms → False |

The cap of 4 is enough for legitimate input. Executed nesting depths 1–6 of *literal* quotes: 1–4 still classify transient on both channels; 5+ fall through to `request`. Real traffic is 0–2 layers (bare `ERROR`, `json.dumps("ERROR")` → `'"ERROR"'`). Nothing observed or plausible from an aggregator nests five matched pairs around a five-letter sentinel.

## Classification matrix — unchanged on every real case

Ran the same matrix on both heads, both channels (`raise_for_status` via a real `httpx.Response`, and `_compat_stream_error`). Opaque sentinel and quoted sentinel stay `transient`/`retryable=True`. Unbalanced quotes, real diagnostics (`context length exceeded`), near-misses (`ERROR!`, `errors`, `Error: x`), nested-object diagnostics, wrong outer message, and direct-provider 400s stay `request`/`retryable=False`. Nested-object sentinel stays transient. The **only** delta vs 880f6fff is 5+ *literal* quote layers, which is the bound C9 asked for.

## C8 — not actually fixed (see C10)

The loop justification is no longer the round-2 sentence. The replacement is also false.

## New findings

**C10 — MINOR — the replacement docstring still misstates what `_openrouter_upstream_text` does with a double-encoded `raw`.** clients.py:319–323 now says a truly double-encoded `raw` arrives backslash-escaped (`"\\"ERROR\\""`), which `_openrouter_upstream_text` "has already parsed away by the time this sees it, so that input never reaches the loop." Executed against the live functions:

```
raw (python)                 '"\\"ERROR\\""'
json.loads(raw)              '"ERROR"'          # str, not Mapping
_openrouter_upstream_text    '"\\"ERROR\\""'    # == raw, not the parse
_strip_quote_pair was called with that exact string
opaque                       False
```

`_openrouter_upstream_text` *does* `json.loads`, then — because the inner value is a `str`, not a `Mapping` — falls through to `return raw.strip()[:500]`, i.e. the original unparsed string. The parse is discarded. The double-encoded body **does** reach `_strip_quote_pair`; the peel takes one outer pair and stops on the remaining backslash, so the body correctly stays `request`. Behaviour is fine and conservative. The new sentence is still wrong, in a different way from C8: it claims the input never reaches the loop, and it claims the parse consumed it. Both halves are false. One line: say the JSON-string inner is returned as the original `raw` (the non-Mapping fallthrough), so a backslash-escaped body *does* hit the loop and *does not* match, which is deliberate.

Not a reason for another round — same class as C8, same "comment is wrong, behaviour is right" shape.

## Gates and suite

flake8 / `black==26.1.0 --check` / `isort==5.13.2 --check` / pyright: all 0 over the whole tree.

Full unit suite (`env -u NO_COLOR TERM=xterm-256color .venv/bin/python -m pytest tests/unit -q`): **3 failed, 9326 passed, 15 skipped** in 644.70s. Re-ran the three failures in isolation: 116 passed / 39.82s, so all three are flakes, none caused by this 27-line delta.

1. `tests/unit/session/test_launch_subagent.py::test_the_loop_stays_responsive_while_several_subagents_run` — **new** (1103 ms vs 1000 ms bound). Timing flake, not this diff.
2. `tests/unit/tools/test_task_wait_jobs.py::test_jobs_says_wait_for_a_job_that_has_not_been_admitted` — **new** (`assert '215.0s' in '... 215.1s wait ...'`). Rounding flake, not this diff.
3. `tests/unit/tui/test_steering_approval.py::test_typing_a_steer_at_a_live_prompt_never_answers_it` — one of the two known pre-existing steering-approval flakes named in the brief.

Opaque-aggregator tests: 11 passed.

## Version / main

`pyproject.toml` is `0.44.2`. PyPI `/pypi/local-operator/0.44.2/json` is 404; latest on PyPI is `0.44.1`. **origin/main has moved** under the branch: `168efe48 perf(tui): stop repainting unchanged chrome and animating blurred terminals (#455)` is on `origin/main` and not in this branch. `origin/main` is still `0.44.1`, so there is no version collision, but the branch needs another rebase before merge.

**Verdict: TERMINAL — no BLOCKER/MAJOR.** C9 is genuinely fixed and measured. C8's replacement justification is still false (C10, MINOR, follow-up). Remaining minors/nits are not a reason for another round. Rebase onto current `origin/main` before merge.

damianvtran:

Round 3 was TERMINAL (no blocker/major) with one leftover MINOR, C10, plus a rebase need because `origin/main` moved to `efbf3c3f` (0.44.2, which is now on PyPI). Both are in `71dd1158`.

- **C10 (MINOR) — fixed in `71dd1158`.** The round-3 reviewer is right and I independently reproduced it: `_openrouter_upstream_text` `json.loads` a string-valued inner, then — because the inner is a `str`, not a `Mapping` — returns the ORIGINAL unparsed `raw`. That body **does** reach `_strip_quote_pair`, which peels matched quote pairs off it. Behaviour was always this (double-encoded `'"ERROR"'` still classifies `transient`); only the stated reason was wrong, for the third consecutive round. The comment now describes that path, including the Mapping-inner case that never gets here.

- **Rebase + version.** Rebased onto `origin/main` `efbf3c3f` (which includes merged #310). Version is now **0.44.3** — 0.44.2 shipped while this branch was in review. Patch, same as before: self-contained reliability fix, no step-function capability.

Classification matrix after the rebase, both channels, unchanged:

```
http/inb  sentinel     transient  True
http/inb  quoted       transient  True
http/inb  unbalanced   request    False
http/inb  real diag    request    False
http/inb  ERROR!       request    False
http      direct 400   request    False
```

(The in-band `direct 400` probe without a `code` key is already `retryable=True` at the base of `_compat_stream_error` — pre-existing, not this delta.)

Focused suite `tests/unit/providers/test_clients.py` + `test_failover.py`: 349 passed.

damianvtran:
### Agent review — round 4
Reviewer: reviewer subagent (zai/glm-5.3), independent of implementing agent
Scope: 573f9079eede97999dd1a72a2cfa69aa59d6aca3..71dd115881a03976c5d6dfb35bc55a69c25e84d2

Freshness round on the C10 remediation + rebase. The full PR delta vs `origin/main` is unchanged in substance: `clients.py` (+121), provider tests (+290), `pyproject.toml` 0.44.2 → 0.44.3. Everything new in the round-4 scope is the `_strip_quote_pair` docstring rewrite and the version bump; the rest of `573f9079..71dd1158` is the rebase carrying merged `origin/main` (#310 assets, #455 chrome-paint, #471 0.44.2 bump). Verified in `/tmp/fix-288` at `71dd1158`; `~/local-operator` untouched.

## C10 — the docstring is NOW true, verified by execution

Every sentence of the rewrite executed against the live functions:

```
raw = json.dumps("ERROR")            -> '"ERROR"'          (str inner, not Mapping)
_openrouter_upstream_text(...)       -> '"ERROR"'          == the ORIGINAL unparsed raw
_strip_quote_pair(that)              -> 'ERROR'            (peel runs, matches sentinel)
raise_for_status classification      -> retryable=True     (transient, as documented)

inner = {"error": {"message": "context length exceeded..."}}
_openrouter_upstream_text(json(raw)) -> the unwrapped message; never reaches the peel
                                       (verified: peel not consulted, opaque=False)

raw = json.dumps(json.dumps("ERROR"))-> '"\"ERROR\""'
_openrouter_upstream_text(...)       -> the ORIGINAL raw (str-inner fallthrough)
_strip_quote_pair(that)              -> '\\"ERROR\\"'      (one outer pair, stops on backslash)
classification                       -> request            (conservative, unchanged)
```

The three claims round 3 flagged as false — "never reaches the loop", "the parse consumed it", the backslash-escaped framing — are gone, and their replacements each match observed behaviour. Third time lucky: comment and code now agree.

## Classification matrix — unchanged on both channels

Same matrix as round 3, executed at `71dd1158` through the real `raise_for_status` (a real `httpx.Response`) and `_compat_stream_error`:

| body | HTTP 400 | in-band |
|---|---|---|
| `ERROR` (sentinel) | transient | transient |
| `"ERROR"` (quoted) | transient | transient |
| `'''ERROR""` (unbalanced) | request | request |
| real diagnostic (`context length exceeded`) | request | request |
| `ERROR!` / `errors` / `Error: x` | request | request |
| direct-provider 400 (message not the phrase) | request | — |
| nested-object sentinel | transient | transient |

Identical to the round-3 matrix at `880f6fff` and `573f9079`. No behavioural drift from the rebase or the docstring change.

## Rebase, version, main

- `merge-base(71dd1158, origin/main) == efbf3c3f == origin/main` — the rebase actually landed; branch is `ahead 4`, not diverged.
- `origin/main` fetched at review time and **has not moved again** since the rebase target.
- `pyproject.toml` = `0.44.3`; PyPI `/pypi/local-operator/0.44.3/json` is 404 (free); latest on PyPI is `0.44.2` (matches `origin/main`). No collision.

## Gates and focused suite

Whole tree, exactly as CI runs them: `flake8 .` clean; `black==26.1.0 --check .` 662 files unchanged; `isort==5.13.2 --check .` clean; `pyright` 0 errors / 0 warnings / 0 informations. Focused provider suites `tests/unit/providers/test_clients.py` + `test_failover.py`: **349 passed** in 3.23 s. Full unit suite not re-run per the round-4 brief (delta is a docstring + version bump; round 3 already ran it and isolated its three flakes).

## New findings

None. The only lines that changed in this round are documentation that is now executed-true and a version field.

**Verdict: TERMINAL — no BLOCKER/MAJOR.** C10 is closed and the rebase/version state is clean. Remaining minors/nits from earlier rounds are follow-ups, not a reason for another round. Ready to merge at `71dd1158`.

damianvtran:
The 00:07Z CI run on this head had two anomalies, both addressed:

- `test (3.12)` failed on `tests/unit/session/test_resume_subagents_todos.py::test_redundant_roster_events_skip_the_sidecar_write` (assert 2 == 1). This is a session-sidecar test untouched by this providers-only diff; it passes 6/6 in isolation on this branch and 3/3 on pristine origin/main, and the same matrix passed on the earlier run of the identical head (23m28s). Timing-sensitive flake, not a regression.
- `test (3.13)` hung in progress for ~1h50m with no logs — a stuck runner, not a test failure. The run was cancelled and re-triggered.

The checks re-run below is the authoritative signal for this head.

damianvtran:
### Agent review — round 5
Reviewer: reviewer subagent (anthropic/claude-opus-5), independent of implementing agent
Scope: 71dd115881a03976c5d6dfb35bc55a69c25e84d2..205a23a6cbc33dbeeeeecc81a1abeeec9d1c5edb

Freshness round after origin/main moved to `a39900e3` (v0.44.5 on PyPI). Round 4 was TERMINAL at `71dd1158`. Reviewed the rebase delta only.

## Rebase, version, main

- `merge-base(HEAD, origin/main) == a39900e381e94431a4b145414da39a3d42a174a3 == origin/main` — rebase actually landed; branch is `ahead 4`, not diverged.
- `origin/main` re-fetched at review time and has **not** moved again.
- `pyproject.toml` = `0.44.6`. PyPI latest is `0.44.5`; `0.44.6` is **not** in `releases` (free). No collision, no further rebase needed.

## Production delta vs round 4

`git diff 71dd1158 205a23a6 -- local_operator/providers/clients.py` is **empty**. The only production file this PR owns vs current main is still `local_operator/providers/clients.py`, and it is byte-identical to the round-4 sign-off. Tree-to-tree noise under `local_operator/` is origin/main's intervening TUI/settings work (present on both sides of the rebase), not new hunks from this branch.

No new production-code hunks.

## Opaque-400 classification (the original bug)

Drove `raise_for_status` against a real `httpx.Response(400, json=...)` with the session body (`message` exactly `"Provider returned error"`, `metadata.raw` `"ERROR"`, `provider_name` `"Stealth"`):

```
kind='transient' retryable=True message='Provider returned error: ERROR' status=400
```

Has not regressed.

## Gates and focused suite

- `flake8 .` clean
- `black==26.1.0 --check .` 670 files unchanged
- `isort==5.13.2 --check .` clean
- `pyright` 0 errors / 0 warnings / 0 informations

Focused provider suites `tests/unit/providers/test_clients.py` + `test_failover.py`: **349 passed** in 5.35 s. Full unit suite not re-run (no production-code change since round 4).

## New findings

None.

**Verdict: TERMINAL — no BLOCKER/MAJOR.** Rebase and version bump are the only expected changes; they landed cleanly. Remaining minors/nits from earlier rounds are follow-ups, not a reason for another round. Ready to merge at `205a23a6`.

PR #289, opened 2026-08-24
feat(tui): attach to live sessions across front ends

damianvtran:
## Summary of Changes

Local Operator sessions can now be followed and steered from multiple front ends at once instead of `/resume` refusing whenever another live process owns the transcript.

- `/resume <live-session>` attaches the TUI as a follower instead of opening a second writer or refusing.
- Cold-boot `lop --resume <live-session>` enters the same standalone attach surface; headless `exec --resume` fails closed rather than corrupting the transcript.
- The existing loopback mobile control socket now accepts one daemon plus up to four authenticated attach clients, broadcasts full projections to all viewers, and keeps request acknowledgements connection-local.
- Attached TUIs can prompt, steer, stop, answer gates, switch model/effort, and follow owner death from the same synchronized projection used by mobile.
- Mobile/background workers no longer accumulate while idle: real work runs to completion with no timeout, then the worker exits after a viewer-independent 3-second drain. An open phone or attached TUI does not pin an idle worker.
- A primary TUI whose terminal remains genuinely open stays alive while idle. If Textual loses its terminal reader, any active work still runs to completion, then the orphan exits after the same drain.
- Interactive approvals/asks wait at most 30 seconds, resolve non-authorizing if unanswered, allow any resulting autonomous work to finish, then drain cleanly.

Version: **0.29.0 → 0.30.0**.

## Related Issue(s)

- Related: none (user-reported session ownership and orphan-process failure)

## Impact

- **Change class:** C2 standard / pre-authorized. Structural session-control change with a clear rollback to v0.29.0; no data migration or external API.
- **Core invariant:** exactly one writable host per transcript. Attach clients never construct or open the owner's transcript for writing.
- **Security:** control sockets remain loopback-only and authenticate with the existing 0600 discovery-record key. Protocol version advances to v2 so a new attach client never displaces an old single-client daemon connection.
- **Compatibility:** a new TUI encountering a protocol-v1 owner preserves the existing safe refusal; an old daemon may connect to a v2 registrant as the default daemon class.
- **Lifecycle boundary:** this release makes mobile/background workers and terminal-orphaned TUI hosts activity-owned. A genuinely open primary TUI still owns its process until the planned supervisor/viewer migration; followers never pin idle background workers.
- **Rollback:** install `local-operator==0.29.0`. Transcripts remain append-only JSONL and require no downgrade migration.

### Acceptance criteria

1. Any number of TUI/mobile interfaces can follow one live session and see synchronized projection updates.
2. Any writable viewer can prompt or steer; the owning process remains the sole transcript writer.
3. Cold and in-app resume paths never create a second writer.
4. Active model turns, tools, compaction, subagents, and detached/background work are never aborted or reaped because viewers disappear.
5. Once a mobile/background or terminal-orphaned host has no autonomous work or unresolved gate, it exits cleanly after a 3-second drain regardless of viewers.
6. Owner death is visible to followers, which can resume from the durable transcript.

### Non-goals

- This PR does not migrate every primary TUI to a supervisor-owned disposable worker. A mounted primary TUI remains its own host.
- It does not provide distributed exactly-once tool side effects; it preserves the existing single-writer/session recovery semantics.
- It does not answer multi-question ask flows beyond the existing mobile/TUI gate contract.

### Live owner + follower over the real control socket

Exercised a real `Registrant` socket and `AttachClient` against one session projection:

```text
protocol: 2
attach clients after connect: 1
steer response: steering queued
owner projection: steer from terminal B
follower projection: steer from terminal B
attach clients after detach: 0
```

This proves authenticated discovery, multi-client attachment, steering through the owner, broadcast reconciliation, and dead-client cleanup on the production socket path. Round-one remediation additionally exercised one daemon/mobile-equivalent connection plus two attach clients submitting ordinary prompts and explicit steers concurrently: every producer received its own request-correlated ACK, all inputs were admitted exactly once in deterministic owner order, and all three viewers converged on the same projection. A live non-reading client was then backpressured and dropped without delaying the active clients.

### Idle worker lifecycle

Exercised the child reaper with a shortened clock and the real cleanup sequence:

```text
active work: process retained
work completed: idle drain started
drain elapsed: dispose complete
session claim: released
discovery record: removed
process exit: 0
```

Tests additionally pin that viewer counts 0, 1, and N have identical idle drain behavior, while model/tool/compaction/subagent and detached-background activity prevent reaping.

### Honest local limitation

Two physical terminal devices and a provider-backed mobile child were not available to automate safely in the implementation environment. The full TUI flows were therefore driven through the real CSS-bearing `OperatorApp` Textual pilot; transport and lifecycle paths were exercised with real loopback sockets/process cleanup rather than claimed as physical-device coverage.

## Automated validation

- `env -u NO_COLOR TERM=xterm-256color .venv/bin/python -m pytest tests/unit -q`: **6644 passed, 7 skipped**
- Focused attach/lifecycle suite after remediation: **233 passed**
- `.venv/bin/python -m flake8 .`: passed
- `uvx --from black==26.1.0 black --check local_operator tests`: **483 files clean**
- `uvx isort --check-only --profile black local_operator tests`: passed
- `.venv/bin/python -m pyright --pythonpath .venv/bin/python .`: **0 errors, 0 warnings**
- `.venv/bin/local-operator --version`: **v0.30.0**

New coverage includes multi-connection authentication/broadcast/ack isolation, protocol downgrade refusal, owner identity mismatch, CLI cold-resume guards, attach rendering and steering, owner death/recovery, viewer-independent worker drain, active-work retention, gate expiry, terminal-reader loss, and cleanup ordering.

## Screenshots

Before, live-session resume refusal:
- `/tmp/attach-shots/before-resume-refusal.svg`

After:
- attached session: `/tmp/attach-shots/after-attached-final.svg`
- approval gate: `/tmp/attach-shots/after-approval-final.svg`
- owner exited / resume here: `/tmp/attach-shots/after-owner-death-final.svg`
- `/resume` attach flow: `/tmp/attach-shots/after-resume-attach.svg`
- inspected rendered owner-death PNG: `/tmp/attach-shots/inspected-owner-death-final.png`

Fresh round-one remediation captures:
- live attach: `/tmp/attach-shots/r1-live-inspected.png`
- approval: `/tmp/attach-shots/r1-approval-inspected.png`
- ask: `/tmp/attach-shots/r1-ask-inspected.png`
- owner exited: `/tmp/attach-shots/r1-owner-exit-inspected.png`
- resume here completed: `/tmp/attach-shots/r1-resume-final-inspected.png`

## Checklist

- [x] No tracker was requested or created.

### DIFF
--- a/local_operator/cli.py
+++ b/local_operator/cli.py
@@ -1946,6 +1946,27 @@ def main() -> int:
                         )
                 return 1
 
+            # Second-writer guard, at the ONLY place the hole starts. Left to
+            # the session factory, `lop --resume <live-id>` claims the
+            # directory and silently becomes a second writer on a transcript
+            # another process is appending to — the exact corruption the
+            # in-app /resume guard exists to prevent, reachable on every cold
+            # boot. Owned by another LIVE pid: attach to that owner (a
+            # follower view over its control socket) when it is reachable,
+            # refuse with the same copy the TUI uses otherwise. The factory
+            # never runs on this path. `exec --resume` refuses outright (a
+            # headless one-shot cannot follow) — see the exec branch below.
+            if args.subcommand is None and getattr(args, "resume", None) is not None:
+                from local_operator.resume import live_session_owner
+
+                owner = live_session_owner(config_dir(), str(args.resume))
+                if owner is not None and owner != os.getpid():
+                    from local_operator.cli_attach import run_owned_resume_attach
+
+                    return run_owned_resume_attach(
+                        config_dir(), str(args.resume), owner
+                    )
+
         os.environ["LOCAL_OPERATOR_DEBUG"] = "true" if args.debug else "false"
         # (CL-12) No env_config binding here: the scheduler wrapper resolves its
         # own env config and the session factory does the same lazily — a
@@ -2115,6 +2136,27 @@ def main() -> int:
             invalid = _apply_run_in(args.run_in)
             if invalid is not None:
                 return invalid
+            # Second-writer guard, same rationale as the interactive branch
+            # above but with no attach escape: exec is headless and one-shot,
+            # following a live session is meaningless, and double-writing a
+            # transcript another process owns is the corruption case. The
+            # refusal IS the feature; no new flag.
+            if getattr(args, "resume", None) is not None:
+                from local_operator.resume import live_session_owner, resolve_resume_id
+
+                try:
+                    exec_resume_id = resolve_resume_id(config_dir(), str(args.resume))
+                except Exception:
+                    exec_resume_id = str(args.resume)
+                exec_owner = live_session_owner(config_dir(), exec_resume_id)
+                if exec_owner is not None and exec_owner != os.getpid():
+                    print(
+                        f"\033[31msession {exec_resume_id} is already open in "
+                        f"another process (pid {exec_owner}) — watch and steer "
+                        "it there, or from the phone session list\033[0m",
+                        file=sys.stderr,
+                    )
+                    return 1
             from local_operator.exec_mode import ExecArgs, run_exec
 
             exec_args = ExecArgs(
--- a/local_operator/cli.py
+++ b/local_operator/cli.py
@@ -1963,9 +1963,7 @@ def main() -> int:
                 if owner is not None and owner != os.getpid():
                     from local_operator.cli_attach import run_owned_resume_attach
 
-                    return run_owned_resume_attach(
-                        config_dir(), str(args.resume), owner
-                    )
+                    return run_owned_resume_attach(config_dir(), str(args.resume), owner)
 
         os.environ["LOCAL_OPERATOR_DEBUG"] = "true" if args.debug else "false"
         # (CL-12) No env_config binding here: the scheduler wrapper resolves its
--- a/local_operator/cli_attach.py
+++ b/local_operator/cli_attach.py
@@ -1,0 +1,63 @@
+"""The CLI's owned-resume branch: attach to a live owner instead of racing it.
+
+``lop --resume <id>`` where another live process owns ``<id>`` used to fall
+through to the session factory, claim the directory, and silently become a
+second writer on a transcript the owner was still appending to — the cold-boot
+half of the corruption hole the TUI's in-app guard already refuses. This module
+is the CLI edge of the attach design: discover the owner's control-socket
+record, and either run the standalone attach app (a follower view; the factory
+never runs) or print the TUI's refusal copy and exit 1 when the owner is not
+attachable (old binary, registrant failure, rebind race).
+
+Kept as its own module (not inline in ``cli.py``) because everything here is
+import-heavy (Textual, the mobile package) and the branch is rare: the CLI's
+startup import-cost guard stays green because none of this loads unless an
+owned resume actually happens. ``local_operator.resume`` stays stdlib-only —
+this module imports it, never the reverse.
+"""
+
+from __future__ import annotations
+
+import sys
+from pathlib import Path
+
+
+def _refuse(session_id: str, owner: int) -> int:
+    """Today's refusal line, verbatim with the TUI's, plus exit 1."""
+    print(
+        f"\033[31msession {session_id} is already open in another process "
+        f"(pid {owner}) — watch and steer it there, or from the phone "
+        f"session list\033[0m",
+        file=sys.stderr,
+    )
+    return 1
+
+
+def run_owned_resume_attach(config_dir: Path, session_id: str, owner: int) -> int:
+    """Attach to the live owner of ``session_id``, or refuse exactly as before.
+
+    Exit codes: 0 after a clean detach; 75 when the user chose 'resume here'
+    after the owner died mid-attach (the caller relaunches without ``--resume``
+    — by then the claim marker names a dead pid, so the relaunch is the
+    legitimate first writer); 1 when the owner is not attachable.
+    """
+    from local_operator.mobile.attach_client import find_owner_record
+
+    record, found_owner = find_owner_record(config_dir, session_id)
+    if record is None or found_owner != owner:
+        return _refuse(session_id, owner)
+
+    from local_operator.tui.attach_screen import run_attach_app
+
+    code = run_attach_app(record, session_id)
+    if code == 75:
+        # 'resume here': the owner is gone (the attach screen only offers this
+        # after owner death), so re-enter the CLI with the resume flag dropped.
+        # A fresh process is the honest shape: this one already decided NOT to
+        # build a session when it took the attach branch. ``-m local_operator.cli``
+        # because the package has no ``__main__``; the module's own tail runs
+        # ``main()``.
+        import subprocess
+
+        return subprocess.call([sys.executable, "-m", "local_operator.cli"])
+    return code
--- a/local_operator/cli_attach.py
+++ b/local_operator/cli_attach.py
@@ -54,10 +54,12 @@ def run_owned_resume_attach(config_dir: Path, session_id: str, owner: int) -> in
         # 'resume here': the owner is gone (the attach screen only offers this
         # after owner death), so re-enter the CLI with the resume flag dropped.
         # A fresh process is the honest shape: this one already decided NOT to
-        # build a session when it took the attach branch. ``-m local_operator.cli``
-        # because the package has no ``__main__``; the module's own tail runs
-        # ``main()``.
+        # build a session when it took the attach branch. Preserve the session
+        # id: by now the owner is dead, so the fresh guard falls through to the
+        # factory and this process becomes the legitimate first writer. ``-m
+        # local_operator.cli`` because the package has no ``__main__``; the
+        # module's own tail runs ``main()``.
         import subprocess
 
-        return subprocess.call([sys.executable, "-m", "local_operator.cli"])
+        return subprocess.call([sys.executable, "-m", "local_operator.cli", "--resume", session_id])
     return code
--- a/local_operator/mobile/attach_client.py
+++ b/local_operator/mobile/attach_client.py
@@ -31,12 +31,18 @@ from __future__ import annotations
 
 import asyncio
 import json
+import os
+import subprocess
+import sys
+import time
+import uuid
 from pathlib import Path
 from typing import Any, Callable
 
 from local_operator.mobile.registry import scan
 from local_operator.mobile.types import (
     PROTOCOL_VERSION,
+    ContinuationCommand,
     SessionProjection,
     SessionRecord,
     _projection_from_json,
@@ -245,8 +251,28 @@ class AttachClient:
             raise RuntimeError(str(reply.get("message", "request failed")))
         return str(reply.get("detail", ""))
 
-    async def prompt(self, text: str) -> str:
-        return await self._request("prompt", text=text)
+    async def prompt(
+        self,
+        text: str,
+        *,
+        command_id: str | None = None,
+        images: list[dict[str, str]] | None = None,
+    ) -> str:
+        return await self._request(
+            "prompt",
+            command_id=command_id or str(uuid.uuid4()),
+            text=text,
+            images=list(images or []),
+        )
+
+    async def send_command(self, command: ContinuationCommand) -> str:
+        if command.session_id != self._session_id:
+            raise ValueError("command belongs to another conversation")
+        return await self.prompt(
+            command.text,
+            command_id=command.command_id,
+            images=command.images,
+        )
 
     async def steer(self, text: str) -> str:
         return await self._request("steer", text=text)
@@ -292,3 +318,57 @@ class AttachClient:
                 self._writer.close()
             except Exception:  # noqa: BLE001
                 pass
+
+
+async def continue_command(
+    config_dir: Path,
+    command: ContinuationCommand,
+    *,
+    deadline_s: float = ACK_TIMEOUT_S,
+    on_projection: Callable[[SessionProjection], None] | None = None,
+) -> tuple[AttachClient, str]:
+    """Deliver one retained command to whichever host wins the session lease.
+
+    Every contender may start a candidate. The atomic transcript lease, never
+    a check before spawning, decides authority. Losing candidates exit and the
+    producer redials the published winner with the unchanged command id.
+    """
+    deadline = time.monotonic() + deadline_s
+    spawned = False
+    delay = 0.1
+    last_error: Exception | None = None
+    while time.monotonic() < deadline:
+        record, _ = await asyncio.to_thread(find_owner_record, config_dir, command.session_id)
+        if record is not None:
+            disconnected = asyncio.Event()
+            client = AttachClient(
+                on_projection or (lambda projection: None),
+                lambda reason: disconnected.set(),
+            )
+            try:
+                await client.connect(record, command.session_id)
+                detail = await client.send_command(command)
+                return client, detail
+            except (ConnectionError, RuntimeError, TimeoutError) as exc:
+                last_error = exc
+                client.close()
+        if not spawned:
+            # Only routing data enters the environment. Prompt text, images and
+            # command identity stay on the authenticated loopback connection.
+            env = dict(os.environ)
+            env["LOP_MOBILE_CHILD_CWD"] = str(Path.home())
+            env["LOP_MOBILE_CHILD_RESUME"] = command.session_id
+            await asyncio.create_subprocess_exec(
+                sys.executable,
+                "-m",
+                "local_operator.mobile.child",
+                env=env,
+                stdin=subprocess.DEVNULL,
+                stdout=subprocess.DEVNULL,
+                stderr=subprocess.DEVNULL,
+                start_new_session=True,
+            )
+            spawned = True
+        await asyncio.sleep(min(delay, max(0.0, deadline - time.monotonic())))
+        delay = min(delay * 1.7, 1.0)
+    raise TimeoutError("Couldn’t continue this conversation. Try again.") from last_error
--- a/local_operator/mobile/attach_client.py
+++ b/local_operator/mobile/attach_client.py
@@ -31,7 +31,6 @@ from __future__ import annotations
 
 import asyncio
 import json
-import secrets
 from pathlib import Path
 from typing import Any, Callable
 
@@ -50,9 +49,7 @@ from local_operator.mobile.types import (
 ACK_TIMEOUT_S = 15.0
 
 
-def find_owner_record(
-    config_dir: Path, session_id: str
-) -> tuple[SessionRecord | None, int | None]:
+def find_owner_record(config_dir: Path, session_id: str) -> tuple[SessionRecord | None, int | None]:
     """Locate the discovery record of the live process hosting ``session_id``.
 
     Returns ``(record, owner_pid)``. The normal case matches a record whose
@@ -130,9 +127,7 @@ class AttachClient:
         retry mechanism, by design.
         """
         if record.protocol < 2:
-            raise ConnectionError(
-                f"owner runs protocol v{record.protocol}; attach needs >= 2"
-            )
+            raise ConnectionError(f"owner runs protocol v{record.protocol}; attach needs >= 2")
         self._session_id = session_id
         try:
             reader, writer = await asyncio.open_connection(
@@ -162,9 +157,7 @@ class AttachClient:
             raise ConnectionError(f"owner replied {frame.get('op')!r}, not its state")
         projection = _projection_from_json(frame.get("data") or {}, record)
         if projection.session_id != session_id:
-            raise ConnectionError(
-                f"owner moved to another conversation ({projection.session_id})"
-            )
+            raise ConnectionError(f"owner moved to another conversation ({projection.session_id})")
         self._connected = True
         self._reader_task = asyncio.get_running_loop().create_task(self._pump())
         # Deliver the welcome synchronously so the host paints before any
--- a/local_operator/mobile/attach_client.py
+++ b/local_operator/mobile/attach_client.py
@@ -265,17 +265,45 @@ class AttachClient:
             images=list(images or []),
         )
 
-    async def send_command(self, command: ContinuationCommand) -> str:
+    async def send_command(self, command: ContinuationCommand, *, streaming: bool = False) -> str:
+        """Submit natural text using the latest owner projection.
+
+        The retained command id rides both paths. If an idle projection races a
+        turn start, the owner's busy rejection is reconciled once as steering;
+        no second prompt is created and reconnect retries keep one identity.
+        """
         if command.session_id != self._session_id:
             raise ValueError("command belongs to another conversation")
-        return await self.prompt(
-            command.text,
-            command_id=command.command_id,
-            images=command.images,
-        )
+        if streaming:
+            return await self.steer(
+                command.text, command_id=command.command_id, images=command.images
+            )
+        try:
+            return await self.prompt(
+                command.text,
+                command_id=command.command_id,
+                images=command.images,
+            )
+        except RuntimeError as exc:
+            if "already streaming" not in str(exc):
+                raise
+            return await self.steer(
+                command.text, command_id=command.command_id, images=command.images
+            )
 
-    async def steer(self, text: str) -> str:
-        return await self._request("steer", text=text)
+    async def steer(
+        self,
+        text: str,
+        *,
+        command_id: str | None = None,
+        images: list[dict[str, str]] | None = None,
+    ) -> str:
+        return await self._request(
+            "steer",
+            command_id=command_id or str(uuid.uuid4()),
+            text=text,
+            images=list(images or []),
+        )
 
     async def abort(self) -> str:
         return await self._request("abort")
--- a/local_operator/mobile/attach_client.py
+++ b/local_operator/mobile/attach_client.py
@@ -1,0 +1,301 @@
+"""The attach client: a follower terminal's half of the control socket.
+
+Every interactive ``lop`` process hosts a :class:`~local_operator.mobile.registrant.Registrant`
+whose loopback socket is the phone's window onto the session. This module is
+the SAME socket seen from a second terminal: ``/resume`` of a session another
+process owns dials that owner and renders its projection repaints, steering
+through the same ops the phone uses. One socket, N front ends — a second
+protocol would drift from the first, so there is none.
+
+Design constraints baked in:
+
+- **No auto-reconnect.** Owner death (socket EOF) is a DECISION POINT for the
+  hosting screen — "resume here / detach" — not something to paper over by
+  redialing a pid that may have been reused. The callback fires once and the
+  client is dead.
+- **Identity over pid trust.** ``live_session_owner`` cannot probe pids on
+  Windows, and a recycled pid anywhere defeats pid trust. After auth the
+  registrant sends a full projection unprompted; the client requires that
+  projection's ``session_id`` to match the one the user asked for before
+  declaring the attach good. A mismatch means the owner rebound away — the
+  caller surfaces the graceful refusal copy.
+- **Protocol gate before dialing.** A v1 registrant treats ANY authenticated
+  dial as THE daemon and evicts the real one; requiring ``record.protocol
+  >= 2`` turns that hazard into the graceful degradation path instead.
+
+Stdlib-only plus the mobile wire types: the CLI imports this lazily on the
+owned-resume branch only, keeping ``resume.py`` and the startup path light.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import json
+import secrets
+from pathlib import Path
+from typing import Any, Callable
+
+from local_operator.mobile.registry import scan
+from local_operator.mobile.types import (
+    PROTOCOL_VERSION,
+    SessionProjection,
+    SessionRecord,
+    _projection_from_json,
+)
+
+#: How long to wait for an ack/error matching a request id. Mirrors the
+#: daemon's ``request`` timeout: long enough for a turn-boundary op (prompt
+#: acquires the turn lock) on a busy owner, short enough that a wedged owner
+#: surfaces as an error rather than a hang.
+ACK_TIMEOUT_S = 15.0
+
+
+def find_owner_record(
+    config_dir: Path, session_id: str
+) -> tuple[SessionRecord | None, int | None]:
+    """Locate the discovery record of the live process hosting ``session_id``.
+
+    Returns ``(record, owner_pid)``. The normal case matches a record whose
+    ``session_id`` equals the ask. The rebind-race fallback: the record's
+    session_id is re-stamped only every heartbeat (15s), so when no record
+    matches but the claim marker names a live pid, that pid's record is
+    returned anyway and the welcome projection's identity check (in
+    :meth:`AttachClient.connect`) arbitrates — a stale match costs one refused
+    dial, never a wrong attach.
+
+    ``(None, pid)`` means an owner exists but no usable record does (old
+    binary, registrant failed to start): the caller degrades gracefully.
+    ``(None, None)`` means no owner at all.
+    """
+    from local_operator.resume import live_session_owner
+
+    owner = live_session_owner(config_dir, session_id)
+    if owner is None:
+        return None, None
+    best: SessionRecord | None = None
+    fallback: SessionRecord | None = None
+    try:
+        for record, state in scan(config_dir):
+            if state != "live":
+                continue
+            if record.pid == owner and record.session_id == session_id:
+                best = record
+                break
+            if record.pid == owner:
+                fallback = record
+    except OSError:
+        return None, owner
+    if best is not None and best.protocol >= 2:
+        return best, owner
+    # A v1 record is not dialable (see module docstring) — report the owner so
+    # the caller can print the refusal naming it.
+    if best is not None or fallback is not None:
+        return None, owner
+    return None, owner
+
+
+class AttachClient:
+    """One authenticated ``attach`` connection to a live session's registrant.
+
+    The host (an ``AttachScreen``) supplies two callbacks — both fire on the
+    client's own reader task, so the host must hop to its UI thread. The
+    client is single-use: after ``on_disconnected`` it is dead by design.
+    """
+
+    def __init__(
+        self,
+        on_projection: Callable[[SessionProjection], None],
+        on_disconnected: Callable[[str], None],
+    ) -> None:
+        self._on_projection = on_projection
+        self._on_disconnected = on_disconnected
+        self._reader: asyncio.StreamReader | None = None
+        self._writer: asyncio.StreamWriter | None = None
+        self._reader_task: asyncio.Task[None] | None = None
+        self._pending: dict[Any, asyncio.Future[dict[str, Any]]] = {}
+        self._req_seq = 0
+        self._session_id = ""
+        self._connected = False
+
+    @property
+    def connected(self) -> bool:
+        return self._connected
+
+    async def connect(self, record: SessionRecord, session_id: str) -> None:
+        """Dial, authenticate as an attach client, and verify identity.
+
+        Raises on any failure (dial refused, auth rejected, welcome identity
+        mismatch, protocol too old): the caller collapses every one to the
+        graceful refusal copy — a user re-running the command is the only
+        retry mechanism, by design.
+        """
+        if record.protocol < 2:
+            raise ConnectionError(
+                f"owner runs protocol v{record.protocol}; attach needs >= 2"
+            )
+        self._session_id = session_id
+        try:
+            reader, writer = await asyncio.open_connection(
+                "127.0.0.1", record.control_port, limit=1 << 20
+            )
+        except OSError as exc:
+            raise ConnectionError(f"owner socket unreachable: {exc}") from exc
+        self._reader = reader
+        self._writer = writer
+        auth = {"key": record.control_key, "client": "attach"}
+        writer.write(json.dumps(auth).encode() + b"\n")
+        await writer.drain()
+        # The welcome projection doubles as the identity check: it names the
+        # conversation the OWNER is actually hosting right now, which is the
+        # fact the user cares about and the one a pid cannot prove.
+        try:
+            first = await asyncio.wait_for(reader.readline(), timeout=ACK_TIMEOUT_S)
+        except TimeoutError as exc:
+            raise ConnectionError("owner did not send its state") from exc
+        if not first:
+            raise ConnectionError("owner closed the connection")
+        try:
+            frame = json.loads(first.decode("utf-8", "replace"))
+        except ValueError as exc:
+            raise ConnectionError("owner sent a malformed frame") from exc
+        if frame.get("op") not in ("projection", "welcome"):
+            raise ConnectionError(f"owner replied {frame.get('op')!r}, not its state")
+        projection = _projection_from_json(frame.get("data") or {}, record)
+        if projection.session_id != session_id:
+            raise ConnectionError(
+                f"owner moved to another conversation ({projection.session_id})"
+            )
+        self._connected = True
+        self._reader_task = asyncio.get_running_loop().create_task(self._pump())
+        # Deliver the welcome synchronously so the host paints before any
+        # later repaint can race it.
+        self._on_projection(projection)
+
+    async def _pump(self) -> None:
+        """Read frames until EOF; route projections and match acks by req id."""
+        reader = self._reader
+        assert reader is not None
+        reason = "owner exited"
+        try:
+            while True:
+                line = await reader.readline()
+                if not line:
+                    break
+                try:
+                    frame = json.loads(line.decode("utf-8", "replace"))
+                except ValueError:
+                    continue
+                op = frame.get("op")
+                if op in ("projection", "welcome"):
+                    try:
+                        # pid=0 record: the attach screen keys nothing on the
+                        # record's pid (it reads the projection's own), and
+                        # building a fake record per repaint would imply the
+                        # record carries truth it does not.
+                        self._on_projection(
+                            _projection_from_json(
+                                frame.get("data") or {},
+                                SessionRecord(
+                                    pid=0,
+                                    kind="tui",
+                                    session_id="",
+                                    conversation_name="",
+                                    cwd="",
+                                    model_label="",
+                                    control_port=0,
+                                    control_key="",
+                                    protocol=PROTOCOL_VERSION,
+                                ),
+                            )
+                        )
+                    except Exception:  # noqa: BLE001 — a malformed push must not kill the pump
+                        continue
+                elif op in ("ack", "error"):
+                    future = self._pending.pop(frame.get("req"), None)
+                    if future is not None and not future.done():
+                        future.set_result(frame)
+        except (ConnectionResetError, BrokenPipeError, OSError):
+            reason = "owner connection reset"
+        finally:
+            self._connected = False
+            for future in self._pending.values():
+                if not future.done():
+                    future.set_exception(ConnectionError(reason))
+            self._pending.clear()
+            try:
+                self._writer.close()  # type: ignore[union-attr]
+            except Exception:  # noqa: BLE001
+                pass
+            self._on_disconnected(reason)
+
+    # -- requests ---------------------------------------------------------------
+
+    async def _request(self, op: str, **fields: Any) -> str:
+        """Send one op and await its ack detail (or raise its error message)."""
+        if not self._connected or self._writer is None:
+            raise ConnectionError("not attached")
+        self._req_seq += 1
+        req = self._req_seq
+        future: asyncio.Future[dict[str, Any]] = asyncio.get_running_loop().create_future()
+        self._pending[req] = future
+        frame = {"op": op, "req": req, **fields}
+        try:
+            self._writer.write(json.dumps(frame).encode() + b"\n")
+            await self._writer.drain()
+            reply = await asyncio.wait_for(future, timeout=ACK_TIMEOUT_S)
+        except (ConnectionResetError, BrokenPipeError, OSError) as exc:
+            self._pending.pop(req, None)
+            raise ConnectionError(f"owner connection lost: {exc}") from exc
+        finally:
+            self._pending.pop(req, None)
+        if reply.get("op") == "error":
+            raise RuntimeError(str(reply.get("message", "request failed")))
+        return str(reply.get("detail", ""))
+
+    async def prompt(self, text: str) -> str:
+        return await self._request("prompt", text=text)
+
+    async def steer(self, text: str) -> str:
+        return await self._request("steer", text=text)
+
+    async def abort(self) -> str:
+        return await self._request("abort")
+
+    async def slash(self, command: str, args: str) -> str:
+        return await self._request("slash", command=command, args=args)
+
+    async def set_model(self, provider: str, model_id: str) -> str:
+        return await self._request("set_model", provider=provider, model_id=model_id)
+
+    async def set_effort(self, effort: str) -> str:
+        return await self._request("set_effort", effort=effort)
+
+    async def approval_answer(self, request_id: str, approved: bool) -> str:
+        return await self._request(
+            "approval_answer", request_id=request_id, approved=approved, remember=False
+        )
+
+    async def ask_answer(self, request_id: str, value: str) -> str:
+        return await self._request("ask_answer", request_id=request_id, value=value)
+
+    async def detach(self) -> None:
+        """Close the connection from our side. ``on_disconnected`` still fires
+        (the pump observes EOF) so the host's teardown runs one path."""
+        self._connected = False
+        if self._writer is not None:
+            try:
+                self._writer.close()
+            except Exception:  # noqa: BLE001
+                pass
+
+    def close(self) -> None:
+        """Synchronous teardown for hosts without a loop (app exit paths)."""
+        self._connected = False
+        if self._reader_task is not None:
+            self._reader_task.cancel()
+            self._reader_task = None
+        if self._writer is not None:
+            try:
+                self._writer.close()
+            except Exception:  # noqa: BLE001
+                pass
--- a/local_operator/mobile/child.py
+++ b/local_operator/mobile/child.py
@@ -10,9 +10,9 @@ session its work.
 
 The child builds a session with the CLI's composition root, wraps it in the
 owned-session handle (approval/ask gates resolved from the phone), registers
-it through the normal record + control socket path, and idles until its work
-ends, a signal arrives, or — since the attach/reaping design — no front end
-holds it any longer (see :func:`_reaper`). Environment variables are the
+it through the normal record + control socket path, and idles until a signal
+arrives or its work and pending gates drain long enough for self-reaping (see
+:func:`_reaper`). Environment variables are the
 spawn contract (``LOP_MOBILE_CHILD_CWD``, ``_PROVIDER``, ``_MODEL``) — argv
 would be ps-readable.
 """
@@ -28,17 +28,13 @@ import time
 
 logger = logging.getLogger(__name__)
 
-#: How often the reaper re-evaluates "is anyone still holding this session?".
-#: Short enough that a front end leaving is noticed within the grace budget's
-#: own resolution, long enough that five-second polls cost nothing.
-REAP_CHECK_S = 5.0
+#: Fine polling makes the 3-second drain predictable while remaining cheap for
+#: one event loop. Viewer traffic is deliberately not an input to this loop.
+REAP_CHECK_S = 0.25
 
-#: Default grace before a held-by-nobody session exits, overridable with
-#: ``LOP_SESSION_GRACE_S``. Long enough that a phone screen-lock / tunnel
-#: blip (SSE drop + reconnect) never kills a session the user is coming back
-#: to; short enough that unattended sessions cannot accumulate as live
-#: processes — the buildup this reaper exists to prevent.
-DEFAULT_GRACE_S = 120.0
+#: Idle execution hosts are disposable once their durable session state is
+#: quiescent. This is a drain for newly arriving work, not a reconnect grace.
+DEFAULT_GRACE_S = 3.0
 
 
 def _grace_seconds() -> float:
@@ -51,43 +47,19 @@ def _grace_seconds() -> float:
 
 
 def _should_exit(handle: object, registrant: object) -> bool:
-    """True iff NO front end holds the session and no work is running.
-
-    The mixed-version guard is load-bearing: until the registrant has SEEN a
-    watch/unwatch op (``watch_supported``), phone watchers are UNKNOWN — an
-    old daemon never sends the ops — and unknown must count as present, or a
-    new child under an old daemon would reap a session a phone is actively
-    watching. The latch never resets, so once a watch-capable daemon has
-    spoken, a true zero is trusted forever after.
-    """
-    watchers_known_empty = (
-        bool(getattr(registrant, "watch_supported", False))
-        and getattr(registrant, "phone_watchers", 1) == 0
-    )
-    if not watchers_known_empty:
-        return False
-    if getattr(registrant, "attach_clients", lambda: 0)() > 0:
-        return False
+    """True when the execution host is quiescent, regardless of viewers."""
+    del registrant  # Watchers and replicas observe work; they do not own it.
     is_busy = getattr(handle, "is_busy", None)
-    if is_busy is not None and is_busy():
-        return False
-    return True
+    return not (is_busy is not None and is_busy())
 
 
 async def _clean_exit(handle: object, registrant: object) -> None:
-    """Deny gates, dispose the session (releases the claim), unpublish.
+    """Dispose the quiescent session, then unpublish its owner record.
 
-    Ordering mirrors OperatorApp.on_unmount: the gates are refused BEFORE
-    dispose because dispose awaits teardown and a turn parked on an
-    unanswered card would never reach it; the record is unpublished LAST so
-    a scanner never sees a gone record while the claim is still held.
+    The reaper reaches this only after ordinary gate timeouts and all resumed
+    work have drained, so injecting a shutdown denial here would violate the
+    same no-interruption invariant that selected this state.
     """
-    deny = getattr(handle, "_deny_pending_gates", None)
-    if deny is not None:
-        try:
-            deny()
-        except Exception:  # noqa: BLE001 — shutdown must proceed
-            logger.debug("child gate deny failed", exc_info=True)
     try:
         await handle.dispose()  # type: ignore[attr-defined]
     except Exception:  # noqa: BLE001 — dispose is best-effort at exit
@@ -99,15 +71,10 @@ async def _clean_exit(handle: object, registrant: object) -> None:
 
 
 async def _reaper(handle: object, registrant: object, stop: asyncio.Event) -> None:
-    """Exit the child cleanly once no front end has held it for a grace period.
+    """Exit the disposable execution host after one uninterrupted idle drain.
 
-    The child is the only party with the authoritative liveness facts (its
-    turn, its subagents, its connections), so the child reaps ITSELF — the
-    daemon never kills children. The grace timer starts at the earliest
-    moment all hold conditions are false: a turn mid-flight when the last
-    front end leaves simply defers the start (``is_busy`` gates it), so the
-    grace outlives the turn by construction. Any holder returning mid-grace
-    cancels the countdown.
+    Autonomous work and ordinary pending-gate timeouts are authoritative. Any
+    new activity resets the drain; viewer count intentionally does not.
     """
     grace_s = _grace_seconds()
     while not stop.is_set():
@@ -118,9 +85,9 @@ async def _reaper(handle: object, registrant: object, stop: asyncio.Event) -> No
         while time.monotonic() < deadline:
             await asyncio.sleep(REAP_CHECK_S)
             if stop.is_set() or not _should_exit(handle, registrant):
-                break  # a front end came back (or shutdown began)
+                break  # work arrived (or shutdown began)
         else:
-            logger.info("mobile child: no front end for %.0fs — exiting cleanly", grace_s)
+            logger.info("mobile child: idle for %.1fs; exiting cleanly", grace_s)
             await _clean_exit(handle, registrant)
             stop.set()  # amain's wait() returns; exit code stays 0
             return
--- a/local_operator/mobile/child.py
+++ b/local_operator/mobile/child.py
@@ -10,10 +10,11 @@ session its work.
 
 The child builds a session with the CLI's composition root, wraps it in the
 owned-session handle (approval/ask gates resolved from the phone), registers
-it through the normal record + control socket path, and idles until killed
-or its turn ends the conversation. Environment variables are the spawn
-contract (``LOP_MOBILE_CHILD_CWD``, ``_PROVIDER``, ``_MODEL``) — argv would
-be ps-readable.
+it through the normal record + control socket path, and idles until its work
+ends, a signal arrives, or — since the attach/reaping design — no front end
+holds it any longer (see :func:`_reaper`). Environment variables are the
+spawn contract (``LOP_MOBILE_CHILD_CWD``, ``_PROVIDER``, ``_MODEL``) — argv
+would be ps-readable.
 """
 
 from __future__ import annotations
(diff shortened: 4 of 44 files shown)


damianvtran:
### UX review — round 1
- Reviewer: ux-reviewer subagent (openai/gpt-5.6-sol)
- Scope: f4899ae44c38da25ddf6b2f010fe573797d77e81..4bcef6938a110ff206b079816b3613b099943379
- Flow exercised: Real CSS-bearing `OperatorApp` pilots for attached idle/streaming submit, `/detach` and Escape return, approval/ask keyboard focus, owner-exit recovery, and the exact in-app `AttachScreen` construction; focused cold `--resume`, protocol-v1 refusal, real attach-client transport, and viewer-independent idle-drain tests (34 passed); rendered attached, approval, and owner-death artifacts inspected. I also traced standalone exit-75 relaunch and multi-viewer request/projection behavior.
- Verdict: Changes requested

| ID | Severity | Finding |
|---|---|---|
| U1 | MAJOR | **In-app owner death promises a recovery action that does nothing.** Acceptance criterion 6 says owner death is visible to followers, which can resume from the durable transcript. Starting from in-app `/resume <live>`, the production path pushes `AttachScreen(record, concrete)` without `on_resume_here`. After simulating owner exit in the real `OperatorApp`, the screen correctly displayed `r: resume here`, but pressing `r` left the dead attach screen mounted; `_on_resume_here` was `None`. This is a dead end and the rendered copy makes a false promise. Minimum remediation: wire the in-app attach path to a callback that safely pops/replaces the follower and resumes the same durable session, with visible progress/failure feedback, and pilot the exact production construction through owner exit and `r` to a usable resumed composer. |
| U2 | MAJOR | **Approval and ask controls are not keyboard-reachable in the normal attached flow.** The attached composer retains focus, `Tab` leaves focus on that sole focusable widget, and `on_key` explicitly refuses gate shortcuts while the composer has focus. In real `OperatorApp` pilots, pressing `y` on `Allow? — y/n` and `a` on an ask offering Alpha/Beta inserted those letters into the draft and sent no answer. The existing approval test manually blurs the composer, an action users cannot discover or perform through the rendered flow. This violates the stated attached-viewer ability to answer gates and creates multi-viewer confusion because another viewer may resolve a request while this one appears actionable. Minimum remediation: make the displayed gate actionable from normal focus (or provide a clearly discoverable focus route), prevent answer keys from becoming draft text when used for the gate, show acknowledgement/resolution or stale-answer feedback, and add production-style pilots for both approval and ask while the composer begins focused. |

**Verdict:** Changes requested. This round is not TERMINAL because U1 and U2 are MAJOR interaction failures.

damianvtran:
### Design review — round 1

- Reviewer: designer subagent (openai/gpt-5.6-sol)
- Scope: f4899ae44c38da25ddf6b2f010fe573797d77e81..4bcef6938a110ff206b079816b3613b099943379
- Artifacts: `/tmp/attach-shots/before-resume-refusal.svg`, `/tmp/attach-shots/before-transcript.svg`, `/tmp/attach-shots/after-attached-final.svg`, `/tmp/attach-shots/after-approval-final.svg`, `/tmp/attach-shots/after-owner-death-final.svg`, `/tmp/attach-shots/after-resume-attach.svg`, `/tmp/attach-shots/inspected-owner-death-final.png`
- Verdict: Changes requested

| ID | Severity | Evidence / required action |
|---|---|---|
| D1 | MAJOR | The attached and approval frames present mutually exclusive states. In `after-attached-final.svg`, the top banner says `attached · pid 4242 · Attach Demo`, while the only composer affordance says `owner exited — r to resume here, esc to detach`; `after-approval-final.svg` simultaneously shows an actionable `bash: rm -rf build — y/n` gate and the same disabled owner-exited composer. A user cannot tell whether steering/approval is possible, and in the attached frame the owner-death recovery action is no longer visible in the emphasized pending row. This violates the owner-death visibility/recovery acceptance criterion and makes a destructive approval's availability ambiguous. Preserve one coherent state after disconnect, including across late projection repaint(s): either show the live attached banner, active composer, and actionable approval, or show owner-exited recovery consistently and do not present the approval as answerable. Re-capture live attached, approval, and late-projection-after-owner-exit frames after remediation. |
| D2 | MINOR | In `after-owner-death-final.svg` and the inspected PNG, the same recovery instruction appears three times: muted top banner, yellow pending line, and disabled composer placeholder. The yellow row correctly establishes the recovery action, but the duplicates flatten hierarchy and spend three rows on one status. Keep a single prominent recovery callout plus, at most, a concise contextual owner status; make the disabled composer visually inert without repeating the full instruction. |

damianvtran:
### Agent review — round 1
- Reviewer: reviewer subagent (openai/gpt-5.6-sol)
- Scope: f4899ae44c38da25ddf6b2f010fe573797d77e81..4bcef6938a110ff206b079816b3613b099943379
- Mode: Comprehensive
- Delivery contract: PR #289 acceptance criteria
- Applicable risk surfaces: transcript sole-writer ownership and cold-resume races; authenticated multi-client control transport; concurrent prompt/steer dispatch; mobile-child and terminal-orphan lifecycle; gate expiry; task/job cleanup; protocol downgrade; POSIX/Windows liveness; test realism
- Verdict: Changes requested

| ID | Severity | Finding | Evidence / failure path | Minimum remediation |
|---|---|---|---|---|
| F1 | BLOCKER | Cold resume still has a check-then-claim race, so two processes can become transcript writers. This violates acceptance criterion 3 and the core sole-writer invariant. | `local_operator/cli.py:1959-1966` only checks `live_session_owner()` before continuing to the ordinary factory when no live owner is observed. The actual claim remains a non-exclusive `mkdir(..., exist_ok=True)` plus `write_text()` in `local_operator/session/retention.py:244-276`. Two cold `lop --resume <id>` launches, or two “resume here” followers after owner death, can both pass the guard before either writes the marker; both factories then overwrite the same marker and open the same transcript. The new tests cover one guarded process and one relaunch, but no simultaneous contenders. | Make writer acquisition atomic at the session construction boundary (not only at the CLI preflight), with the loser attaching/refusing after re-reading the winning owner. Add a deterministic two-contender test proving exactly one session reaches transcript construction. |
| F2 | BLOCKER | The 3-second child reaper can cancel a live background bash job, violating acceptance criterion 4 (“detached/background work are never aborted or reaped”). | `_should_exit()` trusts only `OwnedSessionHandle.is_busy()` (`local_operator/mobile/child.py:49-53`). That predicate checks streaming, compaction, the turn lock, `running_subagents()`, gates, and the handle’s naming futures (`local_operator/mobile/owned.py:236-266`), but not running `bash` jobs in `session.jobs`. The existing session explicitly treats both `task` and backgrounded `bash` as model-owned jobs (`local_operator/session/session.py:3632-3651`), while `Session.dispose()` tears down jobs after cancelling background work (`local_operator/session/session.py:5611-5669`). Therefore a turn that launches a background bash process and returns becomes “idle”; three seconds later disposal cancels that still-running job. `test_detached_background_work_is_busy_until_done` only inserts a synthetic future into the handle’s private naming-task set, so it does not exercise the real job manager path. | Include every running/queued session-owned job that disposal would terminate in the authoritative busy predicate, and add a real `AsyncJobManager` background-bash lifecycle test showing the child remains alive until the job settles. |
| F3 | MAJOR | The advertised 30-second terminal gate deadline does not expire for a real approval/ask parked inside a streaming turn. | `_TerminalFrontendReaper.observe()` returns `"settle"` only when `gate_pending` and `not busy` (`local_operator/tui/app.py:674-696` in this head). `_terminal_loss_busy()` reports busy whenever `session.is_streaming` is true (`local_operator/tui/app.py:6838-6854`). A real tool approval is awaited inside `_run_turn`, whose `_is_streaming` remains true until the entire turn exits (`local_operator/session/session.py:3169-3180`), so an orphaned approval remains both `gate_pending=True` and `busy=True` indefinitely. Cheap reproduction against the new state machine: observations at t=0 and t=31 with `reader_alive=False, gate_pending=True, busy=True` both return `None`, leaving `gate_since=0.0`. The tests model a gate with `session.streaming=False`, which is not the production state. | Separate gate-owned turn liveness from unrelated autonomous activity so the front-end gate is denied at 30 seconds without aborting concurrent work; add an integration-level approval/ask test with production-realistic streaming state. |
| F4 | MAJOR | One authenticated but non-reading attach client can stall projection delivery and acknowledgements for every other front end. This breaks multi-connection isolation and synchronized steering under a slow/leaked viewer. | `_broadcast()` awaits clients serially, and `_send_to()` wraps `writer.write()` plus unbounded `writer.drain()` in one registrant-wide `_send_lock` (`local_operator/mobile/registrant.py:539-556`). Once an attach socket stops reading and its kernel buffer fills with full projections, its `drain()` can wait indefinitely while holding the global lock. ACKs to active clients use the same `_send_to()` path (`registrant.py:438-445`), so prompt/steer receipts and all later broadcasts block behind the stalled follower. Current socket tests verify dead-client isolation and point-to-point ACK routing, but not backpressure from a live non-reader. | Use per-connection serialization and bounded send/backpressure handling (dropping only the stalled client), broadcast concurrently, and add a real-socket non-reading-client test proving another client still receives ACKs/projections promptly. |

Focused changed-path suite executed during review: `241 passed in 14.65s`. The failures above are invariant gaps not exercised by that suite.

damianvtran:
- F1 (blocker) — Fixed in `993cf23c`: added an atomic `.execution-lease` acquired inside session construction before the transcript opens, with token-checked release and proven-dead stale recovery. `.session.pid` remains only a compatibility mirror. Concurrent-process and successor-release regression tests prove exactly one execution owner and that stale cleanup cannot remove a successor lease. This is process ownership only: `0ef8ced9` adds bounded owner-side FIFO admission so simultaneous TUI/mobile/attach prompts all land exactly once instead of being refused.
- F2 (blocker) — Fixed in `993cf23c`: the authoritative busy predicate now includes the real session job manager and queued prompt admission, so background bash/task jobs and detached work prevent disposal. Real job-manager lifecycle tests keep the child alive until the job settles, then allow the three-second drain.
- F3 (major) — Fixed in `993cf23c`: gate-owned turn liveness is separated from unrelated autonomous work. Approval/ask gates expire non-authorizing after 30 seconds even while the enclosing turn remains streaming; model/tool/compaction/subagent work is never aborted, and any work unblocked by settlement completes before a fresh drain.
- F4 (major) — Fixed in `993cf23c`: registrant writes are serialized per connection with bounded drain time; broadcasts fan out concurrently and drop only a stalled client. A real-socket non-reading-client test proves active clients continue receiving request-correlated ACKs and projections.

Additional concurrency clarification in `0ef8ced9`: one daemon/mobile-equivalent connection plus two attach clients submit ordinary prompts and explicit steers concurrently. Each producer receives its own ACK, every input is admitted once in deterministic owner order, and all viewers reconcile to the same ordered projection.

damianvtran:
- D1 (major) — Fixed in `993cf23c`: live attach, approval/ask, owner-exited, and resumed states are mutually exclusive. The capture harness now waits for the intended state instead of racing disconnect, and source rendering clears live controls/copy when the owner exits.
- D2 (minor) — Fixed in `993cf23c`: owner-exited guidance is consolidated to one persistent `r: resume here · esc: detach` action line; the disabled composer now says only `attached session inactive`.

Fresh rendered artifacts:
- `/tmp/attach-shots/r1-live-inspected.png`
- `/tmp/attach-shots/r1-approval-inspected.png`
- `/tmp/attach-shots/r1-ask-inspected.png`
- `/tmp/attach-shots/r1-owner-exit-inspected.png`
- `/tmp/attach-shots/r1-resume-final-inspected.png`

damianvtran:
- U1 (major) — Fixed in `993cf23c`: the production in-app attach path now passes an `on_resume_here` callback. Owner EOF followed by `r` exits the follower, acquires sole execution ownership, resumes the same durable transcript, and shows `session <id> resumed here and ready`; failures surface visibly. The exact production `OperatorApp` construction is covered by an owner-death → `r` → usable-composer pilot.
- U2 (major) — Fixed in `993cf23c`: approval `y`/`n` and ask option shortcuts intercept from normal composer focus, do not become draft text, and show resolution/stale-answer feedback. Production-style pilots start with the composer focused for both approval and ask.

damianvtran:
### Design review — round 2
- Reviewer: designer subagent (openai/gpt-5.6-sol)
- Scope: f4899ae44c38da25ddf6b2f010fe573797d77e81..0ef8ced910e04900ca14911d10c4306ea2284f61
- Delta reviewed: 4bcef6938a110ff206b079816b3613b099943379..0ef8ced910e04900ca14911d10c4306ea2284f61
- Artifacts: `/tmp/attach-shots/r1-live-inspected.png`, `/tmp/attach-shots/r1-approval-inspected.png`, `/tmp/attach-shots/r1-ask-inspected.png`, `/tmp/attach-shots/r1-owner-exit-inspected.png`, `/tmp/attach-shots/r1-resume-final-inspected.png`
- Verdict: Approve

No findings.

D1 is resolved: live, approval, ask, owner-exited, and resumed frames now show mutually consistent states. The live gate frames retain the attached banner and active focused composer; the owner-exited frame removes the live gate, changes the banner to `owner exited · pid 4242`, and presents recovery as the only emphasized action. The late-projection guard in the remediation delta prevents a queued live projection from repainting that terminal state.

D2 is resolved: recovery guidance appears once in the yellow action row. The banner supplies concise context and the disabled composer reads only `attached session inactive`, restoring a clear hierarchy without duplicate instructions.

The keyboard-gate remediation introduces no visual regression in the supplied states. Approval exposes `y/n` directly, ask presents the two distinct option labels in the same prominent warning row, and both remain visually separated from the focused composer while their answer keys receive first refusal.

**TERMINAL:** no BLOCKER or MAJOR design findings remain on the current head.

damianvtran:
### UX review — round 2
- Reviewer: ux-reviewer subagent (openai/gpt-5.6-sol)
- Scope: f4899ae44c38da25ddf6b2f010fe573797d77e81..0ef8ced910e04900ca14911d10c4306ea2284f61
- Delta reviewed: 4bcef6938a110ff206b079816b3613b099943379..0ef8ced910e04900ca14911d10c4306ea2284f61
- Flow exercised: Exact CSS-bearing production `OperatorApp` attach construction through owner EOF → `r` → resumed session → enabled, focused composer → successful prompt submission; approval `y` and ask option `a` from the normally focused composer with no draft insertion; stale ask answer rejection showing `answer not accepted: request already answered`; bounded prompt admission through positions 1–32 plus the 33rd input receiving an explicit queue-full error; concurrent daemon/mobile-equivalent and attach prompt/steer producers receiving request-correlated ACKs and converging in owner order. Focused production-style pilots and real registrant sockets passed (7 tests), followed by a direct end-to-end interaction probe for recovery, stale feedback, and capacity rejection.
- Verdict: Approve

No findings.

U1 and U2 are resolved on the current head. Owner-death recovery now reaches a usable composer, gate shortcuts work from normal focus without corrupting the draft, stale answers are visibly recoverable, and concurrent admission gives understandable per-input receipts. Capacity rejection is explicit rather than silent.

**Verdict:** Approve. This round is TERMINAL; no BLOCKER or MAJOR UX findings remain.

damianvtran:
### Agent review — round 2
- Reviewer: reviewer subagent (openai/gpt-5.6-sol)
- Scope: f4899ae44c38da25ddf6b2f010fe573797d77e81..0ef8ced910e04900ca14911d10c4306ea2284f61
- Delivery contract: PR #289 acceptance criteria
- Delta reviewed: 4bcef6938a110ff206b079816b3613b099943379..0ef8ced910e04900ca14911d10c4306ea2284f61
- Applicable risk surfaces: sole execution ownership and stale recovery; owner-side prompt admission, ordering, overflow and failure recovery; concurrent prompt/steer producers; background job liveness; terminal gate expiry; per-client socket backpressure; disposal
- Verdict: Changes requested

| ID | Severity | Finding | Evidence | Required remediation |
|---|---|---|---|---|
| F1 | BLOCKER | The new lease does not yet guarantee one execution owner during stale recovery. Two contenders that both inspect the same dead claim can both return successful leases: a delayed contender executes `os.replace(path, tombstone)` by pathname after the first contender has already replaced the stale inode and acquired a fresh lease at that pathname, so the delayed contender moves the fresh live lease away and then acquires its own. Both processes then believe they are the sole transcript writer. | `local_operator/session_lease.py:124-142` checks only the PID read before the rename and does not fence the rename to that generation/inode. I reproduced this deterministically at the reviewed head by pausing contender B after both contenders read the dead claim, letting A rename the stale file and acquire, then allowing B's pending `os.replace`: both calls returned `SessionLease`, and the final file named only B. `tests/unit/test_session_lease.py:30-49` covers contention on a live winner, and lines 67-80 cover one stale recoverer, but no simultaneous stale-recovery contenders. | Make stale takeover conditional on the exact claim inspected, for example by serializing recovery with an atomic side lock and re-reading under it, or using a primitive that cannot replace a successor by pathname. Add a deterministic two-contender stale-recovery regression proving exactly one success and a live successor remains fenced. |
| F5 | BLOCKER | An admitted FIFO can permanently strand every later prompt after an ordinary non-`RuntimeError` turn failure. `_drain_prompt_queue` catches only `RuntimeError`; provider, transcript, tool, or other exceptions escape the drain task before its `finally` pops the failed head. The task is then done, the failed head remains at index 0, later prompts have already received `prompt queued` ACKs, and no new drain is started until another producer happens to submit. Even then the failed head is retried rather than allowing the acknowledged later prompts to progress. On disposal the queue is silently cleared. This violates the new exact-once admission/error-propagation contract and can lose acknowledged user input. | `local_operator/mobile/owned.py:426-445`. I replaced the session prompt seam with one that raises `ValueError` for the first item: receipts were `prompt admitted` and `prompt queued (2)`, calls remained `['one']`, the queue retained `('two', [])` behind the failed head, the drain task held `ValueError('provider exploded')`, and `is_busy()` stayed true indefinitely. The new FIFO tests only exercise successful turns in `tests/unit/mobile/test_owned.py`. | Define terminal handling for every exception from an admitted turn: remove exactly that item, surface its failure, and continue or explicitly reject every remaining acknowledged item without retrying the failed input. Add tests for non-`RuntimeError` failure, later-item progress and order, no duplicate retry, disposal, and observable error propagation. |

Prior convergence status:
- F2 resolved: `is_busy()` now includes real running `AsyncJobManager` rows, and the focused real background-bash lifecycle test passed.
- F3 resolved: gate expiry is independent of the gate-owned streaming turn while unrelated autonomous work still prevents exit; the focused production-style reaper tests passed.
- F4 resolved: sends are serialized per connection, bounded, and broadcast concurrently; the focused real-socket slow-client test passed.

Verification: the focused remediation selection passed (`22 passed in 6.08s`), which also shows the two reproduced gaps are not covered by the current suite. `git diff --check` passed.

Verdict: **Changes requested**. F1 remains gating and F5 is a new blocker introduced by the bounded FIFO. This round is not terminal.

damianvtran:
- F1 (blocker) — Fixed in `3606ee51`: stale takeover is serialized by a recoverable kernel-backed side lock. The winner re-reads and revalidates the exact generation, pid identity, and proven-dead state under that lock before moving the claim; changed/live/uncertain successors fail closed. Deterministic delayed-contender and eight-process stress tests prove exactly one successor lease, and compare-token release remains fenced.
- F5 (blocker) — Fixed in `3606ee51`: every ordinary `Exception` removes only its failed admitted prompt, emits an observable projection failure, and continues FIFO draining without retry. Cancellation/system-exit retain shutdown semantics; disposal explicitly rejects unadmitted items; overflow errors before admission. Tests cover first-item `ValueError` followed by ordered successor progress, no duplicate retry, visible error, disposal, overflow, and no unhandled task exception.

damianvtran:
### QA review — round 1

- **Reviewer:** qa-tester (openai/gpt-5.6-sol)
- **Scope:** f4899ae44c38da25ddf6b2f010fe573797d77e81..6fc741c8ef2f0e306164dcdebbead91e149c7842
- **Environment:** macOS arm64; `/tmp/lop-attach`; isolated scratch HOME/config; real Registrant TCP sockets, two AttachClients, scratch mobile daemon on loopback, production web bundle in the real cmux browser; no real user sessions/config/credentials
- **Verdict:** **FAIL**

Real-path positives: two concurrent attach clients plus a daemon-equivalent connection admitted distinct prompts and steers in deterministic order; both viewers converged; concurrent and retried `cmd-SAME` produced one durable transcript row (`prompt admitted`, `already admitted`, retry `already admitted`). HTTP auth rejected unauthenticated API requests (401), redirected unauthenticated browser requests (303), rejected bad login (401), malformed JSON (400), unknown non-prompt mutation (409), and a wrong control key (socket EOF), with no transcript side effects. Full unit suite passed (`6649 passed, 7 skipped`), production mobile build passed, and all static gates passed.

| ID | Severity | Finding | Real evidence / required remediation |
|---|---|---|---|
| Q1 | MAJOR | A failed transparent wake from a Previous conversation is visually silent. | In the real production mobile UI I opened durable `qa-previous`, submitted `Reply with exactly QA-WAKE-OK`, and let the real continuation child fail before admission under intentionally credential-free scratch config. After the 15s deadline, the exact draft was restored and transcript rows stayed `3 -> 3`, but the rendered body contained no `Couldn’t continue this conversation. Try again.` or any other error, only the restored draft. Screenshots `/tmp/mobile-shots/qa-wake-failed.png` and `qa-wake-failed-settled.png` are byte-identical. Surface an actionable failure after the deadline while preserving the draft. |
| Q2 | MAJOR | The normal mobile composer still uses process-era generic copy instead of the required session contract. | `/tmp/mobile-shots/qa-previous-open.png` visibly renders `message…`; the contract requires exactly `Message Local Operator…`. Use the shared required placeholder for normal mobile conversation UX. |
| Q3 | MAJOR | A valid durable Previous conversation cannot load older history. | The session appears under `Previous Sessions`, and `GET /api/sessions/qa-previous/events` returns its durable transcript, but `GET /api/sessions/qa-previous/history?limit=10` returns `404 {"error":"unknown session"}` because history requires a live entry. Long Previous transcripts therefore lose rows outside the capped projection. Resolve history by durable `session_id` when no host exists. |

Process census after the failed wake found zero scoped child hosts and no lease, pid marker, discovery record, or control listener for `qa-previous`; transcript remained. All QA-created processes, listeners, records, and the browser surface were cleaned up. Detailed commands, actual responses, matrix, census, screenshots, and honest gaps: `/tmp/pr289-qa-report.md`.

damianvtran:
### QA review remediation — round 1
- Q1 (major) — Fixed in `3f9f971b`: Previous-session continuation failures now retain the exact draft/images/command ID and render the persistent accessible alert `Couldn’t continue this conversation. Try again.` after the transient connecting state. Failed admission produces no transcript row or false ACK. Real no-provider wake evidence: `/tmp/mobile-shots/r1-qa-failure.png`.
- Q2 (major) — Fixed in `3f9f971b`: the mobile composer now uses exactly `Message Local Operator…` in ordinary and streaming states. Evidence: `/tmp/mobile-shots/r1-qa-placeholder.png`.
- Q3 (major) — Fixed in `3f9f971b`: `/api/sessions/<session_id>/history` now serves bounded, chronological older rows from a valid durable user transcript when no live host exists, while malformed/traversal/unknown/subagent routes remain rejected. Long-history evidence: `/tmp/mobile-shots/r1-qa-long-history.json` (`a-065` through `u-070`, `has_more: true`).

damianvtran:
### Design review — round 3

- Reviewer: designer subagent (openai/gpt-5.6-sol)
- Scope: f4899ae44c38da25ddf6b2f010fe573797d77e81..3f9f971b0f9dfe5d1d97b72baab493f5e2b92a13
- Delta reviewed: 0ef8ced910e04900ca14911d10c4306ea2284f61..3f9f971b0f9dfe5d1d97b72baab493f5e2b92a13
- Artifacts: `/tmp/attach-shots/transparent-live.svg`, `/tmp/attach-shots/transparent-post-exit.svg`, `/tmp/attach-shots/transparent-connecting-1.svg`, `/tmp/attach-shots/transparent-connecting-2.svg`, `/tmp/attach-shots/qa-approval.svg`, `/tmp/attach-shots/qa-ask.svg`, `/tmp/attach-shots/qa-failure-draft-restored.svg`, `/tmp/mobile-shots/qa-active-previous.png`, `/tmp/mobile-shots/qa-union-search-previous.png`, `/tmp/mobile-shots/qa-previous-open.png`, `/tmp/mobile-shots/qa-previous-open-settled.png`, `/tmp/mobile-shots/qa-wake-transition.png`, `/tmp/mobile-shots/qa-wake-failed.png`, `/tmp/mobile-shots/qa-wake-failed-settled.png`, `/tmp/mobile-shots/r1-qa-failure.png`, `/tmp/mobile-shots/r1-qa-placeholder.png`
- Verdict: Approve

No findings.

The current TUI frames remove process mechanics entirely: the same transcript and composer remain in place across live, host-absent connecting, and post-exit states. `Connecting…` is the only temporary status, the user's exact draft remains editable, the two consecutive connecting captures are pixel-identical, and the live/post-exit captures are pixel-identical. This is a stable conversation surface rather than a process-recovery screen.

Approval and ask remain visually actionable without displacing the composer. Failure restores the draft in place. Mobile remediation now uses the exact `Message Local Operator…` placeholder and presents the failed continuation as a persistent, high-contrast alert directly above the restored draft with a clear `Try again.` action. It does not add a failed user bubble or imply admission.

The populated and filtered mobile captures preserve a clear Active Sessions / Previous Sessions hierarchy; search leaves the empty Active heading visible while retaining the matching Previous result, so the result's lifecycle classification stays explicit. The conversation, connecting, and failure captures retain the same header/transcript/composer geometry without visible settling reflow. Copy, contrast, focus emphasis, and touch-target sizing are coherent in the supplied desktop/mobile viewport evidence. The retained QA record explicitly notes that mobile loading and empty-list screenshots were not captured, so this approval does not overclaim those states; the reviewed remediation does not alter list loading/empty presentation.

**TERMINAL:** no BLOCKER or MAJOR design findings remain on the current head.

damianvtran:
### Agent review — round 3

- Reviewer: reviewer subagent (openai/gpt-5.6-sol)
- Scope: f4899ae44c38da25ddf6b2f010fe573797d77e81..3f9f971b0f9dfe5d1d97b72baab493f5e2b92a13
- Delta: 0ef8ced910e04900ca14911d10c4306ea2284f61..3f9f971b0f9dfe5d1d97b72baab493f5e2b92a13
- Delivery contract: transparent durable conversations across TUI/mobile; Active/Previous mobile list; one execution owner with concurrent message producers; durable command-ID ACK/dedupe; safe lifecycle/reaping
- Risks reviewed: stale-lease arbitration; job/gate lifecycle; bounded slow clients and FIFO failure; ACK-after-append and retry/crash boundaries; command identity through transcript replay/compaction; attach/TUI/mobile retry; winner discovery; concurrent producers; traversal/auth; Active/Previous reconciliation; history paging; process cleanup; protocol v3 compatibility

Prior convergence status:
- F1 is resolved: stale takeover is serialized by the kernel-backed recovery side lock, revalidates the exact generation/pid under that lock, and the deterministic two-contender plus eight-process tests pass.
- F2 is resolved: the authoritative busy predicate includes real session jobs.
- F3 is resolved: terminal gate expiry no longer depends on the gate-owned streaming flag.
- F4 is resolved: connection-local writes are bounded/serialized and broadcasts do not await clients serially.
- F5 is resolved: every ordinary prompt exception removes exactly its failed FIFO item, rejects an unadmitted receipt, surfaces the failure, and permits later items to progress without retrying the failed head.

| ID | Severity | Finding | Evidence / failure path | Minimum remediation |
|---|---|---|---|---|
| F6 | MAJOR | A mobile conversation opened from **Previous** never subscribes to the live generation created by transparent wake, so it stops receiving projections immediately after its prompt is admitted. | `api_events` captures `_entry_for_session()` once at connection time and only adds the SSE queue to that entry (`local_operator/mobile/daemon.py:840-868`). For a Previous conversation `entry` is `None`, so the queue is attached nowhere. `continue_command()` then starts/adopts a child, but `_fan_out()` publishes only to the new `SessionEntry.subscribers`; the already-open queue is absent. I reproduced the handoff directly: after creating the new entry and calling `_fan_out`, `new_entry_subscribers == 0` and `open_stream_received == False`. The POST can return `prompt admitted`, but the same open screen retains its durable pre-wake projection and never shows the user row, streaming state, response, gates, or child exit unless the user navigates away/reloads. | Make session-id SSE subscriptions survive process-generation changes (for example, session-level subscriber ownership/rebinding), and add a real Previous-open → wake → admitted user row → assistant completion test on one uninterrupted SSE connection. |
| F7 | MAJOR | The attached TUI no longer steers while the owner is streaming; every submit is sent as an ordinary prompt and is rejected on a primary-TUI-owned turn. | `AttachScreen.on_input_submitted()` now unconditionally calls `client.send_command(command)` (`local_operator/tui/attach_screen.py:287-305`), and `send_command()` always emits `op: prompt` (`local_operator/mobile/attach_client.py:263-276`). The adjacent comment still promises routing by owner state. For a primary TUI owner, `TuiSessionHandle.prompt()` calls `Session.prompt()` directly, whose locked-turn guard raises `session is already streaming; use steer() to inject mid-turn` (`local_operator/session/session.py:2514-2532`). This regresses the prior attach behavior that selected `steer` when the projection was streaming, and breaks the concurrent-producer contract for attached terminals during active work. | Restore projection-based prompt/steer routing for a connected live owner while retaining command identity for ordinary prompts and transparent continuation after disconnect; cover an attached-TUI submit during a real streaming owner turn. |

Verification:
- Focused regression selection: **46 passed** (`test_owned`, `test_daemon`, `test_session_lease`, `test_attach_screen`).
- Production mobile bundle: `tsc -b && vite build` passed.
- `git diff --check` passed.
- CI lint, type-check, and pip-audit are green; the two matrix test jobs were still pending when this review was posted.

**Verdict: Changes requested.** F6 and F7 are MAJOR interaction/delivery defects, so this round is not TERMINAL.

damianvtran:
### UX review — round 3

- Reviewer: ux-reviewer subagent (openai/gpt-5.6-sol)
- Scope: f4899ae44c38da25ddf6b2f010fe573797d77e81..3f9f971b0f9dfe5d1d97b72baab493f5e2b92a13
- Delta: 0ef8ced910e04900ca14911d10c4306ea2284f61..3f9f971b0f9dfe5d1d97b72baab493f5e2b92a13
- Flow exercised: Read every PR review, remediation, and QA comment. Walked the production mobile bundle in the real browser from Active/Previous list into a durable Previous conversation; confirmed the exact `Message Local Operator…` placeholder; submitted through transparent wake with no configured provider; observed immediate `Connecting…`, then persistent `Couldn’t continue this conversation. Try again.` with the exact draft retained and no transcript row or surviving child/record. Probed durable Previous history, Active → Previous movement after host exit, and a stable open durable SSE projection. Exercised real loopback transport with two concurrent attach clients, independent prompt/steer ACKs, concurrent and retried `cmd-SAME`, exactly one durable row, and converged projections. Exercised CSS-bearing production `OperatorApp` approval/ask/failure frames and focused `/resume` live/no-record, gate, disconnect, retry, `/detach`, Escape, FIFO failure, and history tests. The transparent TUI/mobile frames contain no attached/inactive/pid/owner-exit/resume-here chrome; `/detach` is absent from persistent chrome.
- Gates: focused interaction selection **23 passed**; full local unit suite **6652 passed, 7 skipped**; mobile production build passed; flake8, black 26.1.0, isort, pyright, and `git diff --check` passed. Hosted CI is red in both Python jobs because `test_startup_import_weight_unchanged` hard-codes `/tmp/lop-attach`, which does not exist on the runner.
- Verdict: Changes requested

| ID | Severity | Finding | Concrete step / expected behavior |
|---|---|---|---|
| U3 | MAJOR | **A Previous conversation loses its live updates at the exact moment transparent wake succeeds.** | Open a Previous conversation and keep that view mounted, then submit a message that successfully wakes a host. The SSE stream was created while no live entry existed, so its queue is not subscribed to the newly adopted generation. The POST can acknowledge admission, but the unchanged open view cannot show the admitted user row, streaming feedback, assistant response, gates, or eventual reaping. A conversation surface that promises transparent continuation must remain live without requiring Back, reload, or reopening. |
| U4 | MAJOR | **A TUI follower rejects normal mid-turn messages instead of steering the active conversation.** | Resume a live, streaming conversation from another TUI and submit ordinary text. The current flow always emits a prompt command, while a primary TUI owner rejects prompts during its locked turn with `session is already streaming; use steer() to inject mid-turn`. Users should be able to type naturally into the same conversation while it is working, with the message accepted as steering and timely feedback, as in the prior reviewed behavior. |
| U5 | MINOR | **The only explicit detach command is not discoverable through the attached surface's help flow.** | Enter the attached conversation and look for an escape route without knowing the diff. Persistent chrome correctly omits `/detach`, and Escape works, but the attach composer has no command registry/help response that names `/detach`; `/help` is forwarded to the owner as conversation input. Keep `/detach` out of persistent process chrome, but expose it from an attached-context help interaction so the command is discoverable. |

The failure/retry remediation is good: feedback is immediate, the product-level error is persistent and actionable, command identity and draft content survive, and no silent transcript side effect occurs. Active/Previous classification, stable durable open projection, history pagination endpoint, exact placeholder, concurrent ACK/dedupe, and process-free chrome also behaved as intended in the exercised paths.

**Verdict:** Changes requested. This round is not TERMINAL because U3 and U4 are MAJOR interaction failures. U5 is a follow-up-level discoverability issue but should be addressed alongside convergence if the attached help contract is in scope.

damianvtran:
### QA review — round 2

- **Reviewer:** task QA (openai/gpt-5.6-sol)
- **Scope:** f4899ae44c38da25ddf6b2f010fe573797d77e81..3f9f971b0f9dfe5d1d97b72baab493f5e2b92a13
- **Mode:** remediation convergence
- **Delta:** 6fc741c8ef2f0e306164dcdebbead91e149c7842..3f9f971b0f9dfe5d1d97b72baab493f5e2b92a13
- **Environment:** macOS arm64; isolated scratch HOME/config; real keyless `test/mock` children, Registrant TCP sockets, scratch mobile daemon, production web bundle in the real cmux browser
- **Verdict:** **FAIL**

Q1-Q3 converge on real paths: failed wake now shows the persistent actionable alert and restores the exact draft with no row; normal mobile placeholder is `Message Local Operator…`; durable Previous history returned a chronological bounded page with `has_more: true`, while traversal/unknown/subagent IDs stayed 404. Successful Previous wake, mock model turn, same-open-view second wake, exact-once duplicate ID, 3s drain, and daemon rediscovery also worked.

| ID | Severity | Finding | Real evidence / required remediation |
|---|---|---|---|
| Q4 | MAJOR | Malformed authenticated Previous prompt input crashes the HTTP route. | `POST /api/sessions/r2-two/command` with `{"op":"prompt","command_id":"two-C","text":"single retry"}` returned `500 Internal Server Error`; daemon logged uncaught `ValueError: badly formed hexadecimal UUID string` from `ContinuationCommand.from_json`. No row was written. Convert malformed command/session input to a bounded 4xx JSON rejection with no traceback. |
| Q5 | MAJOR | SIGHUP and SIGKILL leave stale ownership artifacts after the host is dead and the list is Previous. | Real `test/mock` children were signaled after durable ACK. SIGINT cleaned. SIGHUP pid 26823 and SIGKILL pid 26838 died, but `.execution-lease` and `.session.pid` remained beyond 4s and still remain in the scratch trees. This violates the <=2s stale-record bound and the required Previous census. Reconcile dead-owner claims proactively, without waiting for a future wake. |
| Q6 | MAJOR | Concurrent gate responses do not provide first-wins/stale semantics. | Two simultaneous answers returned `approved` and `denied`; the gate resolved true, but the losing caller also received success and the loop logged `InvalidStateError` from `Future.set_result()`. A later retry alone returned `ValueError: that prompt is no longer waiting`. Atomically claim the pending request before scheduling resolution so one answer succeeds and every loser gets stale feedback. |
| Q7 | MINOR | A very short successful wake never appeared Active in list sampling. | Three valid-UUID concurrent submissions raised child-record count from one to two and one child wrote both exact-once commands, yet every 50ms `/api/sessions` sample reported `section=previous`. The promised provisional Active transition should not depend on scan timing, even for a fast `test/mock` turn. |

Real exact-once concurrency: UUID A submitted twice and UUID B once produced one row each; responses were `prompt admitted`, `prompt admitted`, and `already admitted`. Normal wake/reap left no child record, lease, pid marker, or listener. Idle daemon restart rediscovered the surviving host in 0.198s. Full suite: `6652 passed, 7 skipped`; focused continuation suite `140 passed`; production mobile build and all static gates passed.

All QA-created processes, listeners, records, and browser surface were cleaned. Detailed commands, responses, matrix, censuses, screenshots, and honest gaps: `/tmp/pr289-qa-report-r2.md`.

damianvtran:
- F6 (major) — Fixed in `45dd2342`: mobile SSE subscriptions are now owned by durable `session_id`, not a process-generation `SessionEntry`. An uninterrupted Previous-session stream receives the provisional Active transition, new worker projections, admitted user/assistant rows, gates, and final Previous transition across worker replacement; queues are detached cleanly on disconnect.
- F7 (major) — Fixed in `45dd2342`: neutral TUI submission routes as prompt when idle and steer when the latest owner projection is streaming, with owner-side reconciliation for state races. Real socket/pilot coverage proves mid-turn text is accepted and projected to every viewer rather than rejected by the turn lock.
- Transport teardown observation — Fixed in `ffa39c8c` and `81c22656`: delayed projection tasks are tracked/cancelled/joined; all registrant cleanup now executes on the owner event loop and is idempotent for threaded and in-process hosts. Asyncio debug tests show no pending task or cross-thread cancellation warning.

damianvtran:
### UX review remediation — round 3
- U3 (major) — Fixed in `45dd2342`: an open Previous conversation stays subscribed by durable session ID through transparent wake, worker generation change, model response, and idle reap. No reload/back-navigation is required.
- U4 (major) — Fixed in `45dd2342`: normal text entered from a follower during streaming is accepted as steering and receives timely shared projection feedback; idle text remains a normal durable prompt.
- U5 (minor) — Fixed in `45dd2342`: `/detach` remains absent from persistent chrome, while attached-context `/help` is handled locally and documents returning to the prior conversation/closing the view instead of forwarding help as model input.

damianvtran:
### QA review remediation — round 2
- Q4 (major) — Fixed in `45dd2342`: malformed/missing/wrong-type command IDs are validated at HTTP/control boundaries and return bounded JSON 4xx responses before spawn, lease, or transcript side effects.
- Q5 (major) — Fixed in `45dd2342`: the daemon scan cleans discovery records, `.session.pid`, and `.execution-lease` only after owner death is proven and under compare-token/generation fencing. Real SIGHUP/SIGKILL tests reach clean Previous steady state within one scan interval without touching transcripts or live successor claims.
- Q6 (major) — Fixed in `45dd2342`: approval and ask resolution atomically reserve the pending gate; exactly one concurrent answer wins, the loser receives stale/already-answered feedback, and no `InvalidStateError` is logged.
- Q7 (minor) — Fixed in `45dd2342`: wake publishes session-level provisional Active state immediately and list SSE emits ordered Previous → Active → Previous transitions even for fast workers; no artificial process retention was added.
- Process/resource cleanup follow-up — Fixed in `ffa39c8c` and `81c22656`: registrant delayed pushes and all loop-owned shutdown work are joined through an idempotent owner-loop cleanup path. Asyncio debug coverage passes without pending-task or cross-thread warnings.

damianvtran:
### UX review — round 4

- **Reviewer:** ux-reviewer (openai/gpt-5.6-sol)
- **Scope:** f4899ae44c38da25ddf6b2f010fe573797d77e81..81c22656f97cea297975953194ade4258ef6bd7c
- **Mode:** Remediation convergence
- **Delta:** 3f9f971b0f9dfe5d1d97b72baab493f5e2b92a13..81c22656f97cea297975953194ade4258ef6bd7c
- **Flows exercised:** Read all PR review and remediation notes, then walked the current production mobile bundle in the real browser from the Active/Previous list into an already-Previous `r3-stream` conversation. Without navigating away, submitted `UX round 4 uninterrupted wake`; the same mounted view admitted and rendered the user row and `test/mock` assistant response, remained usable through worker drain, and the list subsequently classified the conversation under Previous. Reviewed the captured connecting/active/drained frames and current list frame for the Previous → Active → Previous transition. Exercised the CSS-bearing production `OperatorApp` follower paths for normal streaming text routing as steer, idle text routing as prompt, local `/help`, `/detach`, and Escape; `/help` stayed local, named `/detach` and its return behavior, and persistent chrome did not expose `/detach` or process-owner mechanics. Exercised bounded invalid Previous command input, continuation transport failure, durable subscriber handoff across process generations, and dead-claim/successor-safe reap behavior. Focused current-head probes passed **10/10**, and the broader convergence selection passed **52/52**.
- **Feedback and recovery:** wake feedback was immediate; successful admission appeared in the uninterrupted conversation; invalid/failure paths remained bounded and recoverable; follower mid-turn text was accepted as steering rather than rejected as a competing prompt.
- **Process-language check:** no PID, owner-exited, attached/inactive, or resume-here mechanics returned to the normal mobile conversation/list flow or persistent follower chrome.
- **Findings:** No findings.
- **Verdict:** **TERMINAL — approve.** U3, U4, and U5 are resolved on the current head. No BLOCKER or MAJOR findings remain; there are no UX follow-ups from this round.

damianvtran:
### Agent review — round 4
- Reviewer: reviewer subagent (openai/gpt-5.6-sol)
- Scope: f4899ae44c38da25ddf6b2f010fe573797d77e81..81c22656f97cea297975953194ade4258ef6bd7c
- Delta: 3f9f971b0f9dfe5d1d97b72baab493f5e2b92a13..81c22656f97cea297975953194ade4258ef6bd7c
- Risks reviewed: F6 session-id SSE generation handoff and disconnect cleanup; F7 prompt/steer race and producer identity; malformed command rejection; proven-dead claim cleanup fencing and Windows fail-closed behavior; approval/ask first-wins; provisional Active lifecycle; portable startup imports; delayed push and cross-thread/in-process registrant shutdown
- Verdict: Changes requested

Prior F1-F5 remain resolved in the complete current diff. F6 also converges: subscribers are session-id-owned, rebinding sends watch state to a successor generation, and disconnect removes the queue and unwatches the current live generation. Q4-Q7 and the registrant teardown follow-up converge in code and focused tests: validation precedes spawn/side effects; dead-claim cleanup revalidates the exact pid/generation under the recovery lock and fails closed on uncertain/Windows errors; gate resolution pops atomically on the owner loop; provisional Active is published before continuation and settled; portable child startup and both registrant shutdown modes are covered.

| ID | Severity | Finding | Evidence / failure path | Minimum remediation |
|---|---|---|---|---|
| F7 | MAJOR | Command identity is not reserved for steering, so the prompt-to-steer race fix does not provide exact-once semantics. | `OwnedSessionHandle.steer()` and `TuiSessionHandle.steer()` deduplicate only against durable `session.history()`, then enqueue without reserving `command_id` while it is pending. Two concurrent/retried steer frames with the same UUID therefore both pass before the steering queue drains. I reproduced this at the reviewed head with two concurrent `OwnedSessionHandle.steer("same", command_id=<same UUID>)` calls: both returned `steering queued`, `session.steer_calls` was `['same', 'same']`, and two projected steer rows were emitted. The new `AttachClient.send_command()` fallback preserves one ID across prompt rejection and steer, but a lost steer ACK/retry or concurrent duplicate still injects the correction twice. | Reserve command IDs across pending steering and durable admission, returning `already admitted` to duplicates; release only on terminal rejection. Apply the same owner-side rule to owned and TUI handles and add concurrent/reconnect duplicate-steer coverage, including the prompt-rejected-as-streaming fallback path. |

Verification: focused remediation suite `88 passed`; child startup/reaper suite `10 passed`; focused pyright `0 errors`; deterministic duplicate-steer reproduction above. No other BLOCKER, MAJOR, MINOR, or NIT finding in the reviewed scope.

**Verdict:** Changes requested. This round is not TERMINAL because F7 remains MAJOR.

damianvtran:
### QA review — round 3

- **Reviewer:** task QA (openai/gpt-5.6-sol)
- **Scope:** f4899ae44c38da25ddf6b2f010fe573797d77e81..81c22656f97cea297975953194ade4258ef6bd7c
- **Mode:** Remediation convergence
- **Delta:** 3f9f971b0f9dfe5d1d97b72baab493f5e2b92a13..81c22656f97cea297975953194ade4258ef6bd7c
- **Environment:** macOS arm64; isolated scratch HOME/config; real keyless `test/mock` children; scratch mobile daemon; real Registrant/AttachClient loopback sockets
- **Verdict:** **PASS**

**No findings.**

Q4-Q7 converge on the exact real paths:

- Malformed authenticated Previous `command_id` returns HTTP 422 JSON (`command_id must be a valid UUID`), with transcript `2 -> 2`, no spawn/lease/record/socket, and no traceback.
- Real SIGHUP child pid 53150 reached dead, artifact-free Previous steady state in 1.125s; real SIGKILL pid 53156 did so in 0.614s. Census sampled process, `.execution-lease`, `.session.pid`, and discovery JSON every 50ms. Each admitted row remains exactly once.
- Concurrent approval answers returned `approved` plus stale `that prompt is no longer waiting`; concurrent ask answers returned `answered` plus stale `that question was already answered`. The winning values were correct and the loop captured no `InvalidStateError` or other exception.
- Real list SSE observed `previous, active, active, active, previous` around a fast `test/mock` wake. The command was admitted, user/assistant rows landed, and final process census was clean.

Additional requested convergence evidence:

- An uninterrupted Previous session SSE received the admitted user command ID and `Hello from the mock provider!` assistant row without reopening the view, then drained artifact-free.
- A streaming follower `send_command(..., streaming=True)` invoked only the host steer seam, returned `steering queued`, and projected the same command ID; it did not call ordinary prompt.
- `PYTHONASYNCIODEBUG=1` registrant/attach shutdown selection passed `6 passed in 1.34s` with no pending-task, cross-thread, wrong-loop, or RuntimeWarning output.
- Focused suite: `148 passed`; full suite: `6660 passed, 7 skipped`; production mobile build and all static gates passed.

All QA-created processes, records, markers, sockets, listeners, and the scratch daemon were cleaned. Detailed commands, actual responses, censuses, and honest scope notes: `/tmp/pr289-qa-report-r3.md`.

damianvtran:
- F7 (major) — Fixed in `bf7cb642`: introduced one owner-loop command-ID reservation namespace shared by prompt and steer admission in both `OwnedSessionHandle` and `TuiSessionHandle`. Concurrent/retried same-ID steers now queue exactly once and return `already admitted` to duplicates; terminal rejection releases the identity; prompt-rejected-as-streaming transfers the reservation atomically to steer; distinct IDs preserve FIFO. Tests cover owned and TUI concurrent duplicates, ACK-loss retry, rejection/retry, prompt→steer handoff, and distinct concurrent IDs.

damianvtran:
### Agent review — round 5
- Reviewer: reviewer subagent (openai/gpt-5.6-sol)
- Scope: f4899ae44c38da25ddf6b2f010fe573797d77e81..bf7cb6420beca39a96dae0b0bb7d85b5b5361fe7
- Delta: 81c22656f97cea297975953194ade4258ef6bd7c..bf7cb6420beca39a96dae0b0bb7d85b5b5361fe7
- Delivery contract: PR #289 acceptance criteria
- Applicable risk surfaces: shared prompt/steer command identity; concurrent and retried admission; durable-history handoff; compaction/replay; reservation lifetime; rejection, cancellation, disposal, and rebind cleanup; Textual owner-loop crossing; FIFO ordering and receipt accuracy; F1-F6/Q4-Q7/teardown regressions
- Verdict: Changes requested

The owner-loop reservation makes concurrent same-ID steer admission atomic in both handles, preserves FIFO for distinct IDs, transfers prompt rejection to steer without self-rejection, releases terminal steer rejection, and reports first versus duplicate accurately in the focused paths. F1-F6/Q4-Q7 and teardown-focused regressions remained green. Two durability/lifetime gaps remain.

| ID | Severity | Finding | Evidence / failure path | Minimum remediation |
|---|---|---|---|---|
| F8 | MAJOR | Compaction can make a previously durable command ID admissible again, breaking exact-once retry semantics for both prompts and steers. | `CommandReservations.reserve()` treats `session.history()` as the long-term authority and deletes an in-memory reservation once that ID appears there (`local_operator/mobile/command_reservation.py:29-40`). But `Session.history()` is explicitly the compacted replay context (`local_operator/session/session.py:1744-1756`), not the append-only transcript. After ID A becomes durable, any later reservation observes A and removes its in-memory entry; once compaction replaces that old user row with a summary, a retry of A is absent from both stores and `reserve()` returns first-admission `True`. I reproduced the state transition deterministically: reserve/accept A, reserve B against history containing A, then reserve A against compacted history containing only a summary; the retry returned `True`. A reconnect after compaction can therefore inject the same producer command twice. | Deduplicate durable IDs against an append-only transcript-backed ID authority that survives compaction/replay, and add prompt plus steer tests that compact away an admitted row and retry the same ID, including a reconstructed/resumed handle. |
| F9 | MAJOR | Accepted steer reservations have no bound while steering never becomes durable, so a live stalled turn can grow this namespace forever. | `accept()` retains every ID (`local_operator/mobile/command_reservation.py:52-53`), and entries are removed only when a later `reserve()` sees them in `session.history()` or on whole-handle clear. Steering becomes durable only when `_drain_steering()` appends it at a model/tool boundary (`local_operator/session/session.py:3704-3722`). If a turn stalls before such a boundary, every distinct steer is accepted into both the steering queue and `_commands`; there is no queue limit, TTL, completion callback, or other bound. A deterministic 10,000-ID reproduction with no durable history left all 10,000 reservations resident. This fails the requested bounded-lifetime property and permits unbounded owner memory growth for the lifetime of a stuck session. | Put a finite admission bound/backpressure policy on undelivered steering (with reservations released only on terminal rejection, never by an unsafe dedupe TTL), or otherwise provide a bounded durable admission ledger; test a never-draining turn at the bound and verify retry/receipt behavior. |

Verification at `bf7cb642`: focused owned/TUI/attach/registrant suite `52 passed`; prior child-reaper/daemon/watch/cold-resume/lease regression suite `38 passed`; deterministic compaction replay and never-durable reservation reproductions above.

**Verdict:** Changes requested. This round is not TERMINAL because F8 and F9 remain MAJOR.

damianvtran:
- F8 (major) — Fixed in `24cab554`: command deduplication now uses an O(1) index seeded from append-only transcript replay and updated only after successful transcript append/flush, independent of compacted model history. Prompt and steer retries remain `already admitted` after compaction, handle/session reconstruction, and append-before-ACK restart boundaries.
- F9 (major) — Fixed in `24cab554`: pending/non-durable steering is a bounded FIFO of 32. New distinct steers are rejected before acceptance at capacity; same-ID retries still dedupe; no accepted ID is TTL-evicted. Durable-admission callbacks release pending capacity while the transcript index preserves exact-once identity. Terminal rejection, disposal, and prompt→steer transfer retain correct accounting.

damianvtran:
### Agent review — round 6
- Reviewer: reviewer subagent (openai/gpt-5.6-sol)
- Scope: f4899ae44c38da25ddf6b2f010fe573797d77e81..24cab55429aecb32639f8449be9e33ff4ea47a58
- Delta: bf7cb6420beca39a96dae0b0bb7d85b5b5361fe7..24cab55429aecb32639f8449be9e33ff4ea47a58
- Delivery contract: PR #289 acceptance criteria
- Applicable risk surfaces: append-only command identity reconstruction and lookup complexity; replay/resume, compaction, fragments, and legacy rows; append/flush/index crash boundaries; prompt and steer reservation lifecycle; bounded steer admission, FIFO drain, rejection/disposal/transfer accounting; TUI owner-loop crossings; memory lifetime; identity namespace isolation; F1-F9, Q4-Q7, and teardown regressions
- Verdict: Changes requested

| ID | Severity | Finding | Evidence / failure path | Minimum remediation |
|---|---|---|---|---|
| F10 | MAJOR | The durable index treats every ordinary user message ID as a producer command ID, so an unrelated local/server-authored message can falsely deduplicate a later mobile command. | `_admitted_command_id()` accepts every `ENTRY_MESSAGE` whose payload role is `user` (`local_operator/session/transcript.py:829-841`), but ordinary local prompts and seeded history also persist user messages with IDs through the same `append_message()` path. I reproduced this deterministically by appending `Message.user(..., id=\"shared-id\")` without any mobile admission and then reserving mobile prompt `shared-id`; `has_admitted_command()` returned true and `reserve()` returned false. This directly violates the required “no false dedupe from non-command message IDs” property and can silently drop a valid producer command whenever namespaces collide. | Persist an explicit producer-command marker/ledger rather than inferring command identity from `role == user`; seed the index only from marked rows while retaining compatible handling for rows written by earlier PR revisions, and test collisions with local prompts, seeded history, custom messages, compaction rows, and fragments. |
| F11 | MAJOR | A failed transcript append remains in `_entries` and can become durable during a later rebuild, resurrecting a command that was terminally rejected before ACK. | `_append()` appends to `self._entries` before opening/writing the file (`local_operator/session/transcript.py:351-405`) and does not roll that mutation back if the write raises. The prompt/steer callers correctly reject and release the reservation on that exception, but a later successful append with a missing transcript rebuilds the whole file from `_entries`, including the failed command. I reproduced: force the first user-row open to raise `OSError`, observe the failed ID still in `entries()`, then perform one successful append; a reconstructed `Transcript` contained the rejected ID and `has_admitted_command()` returned true. Thus crash/error-before-ACK semantics are not atomic: a retry may be accepted and later coexist with the resurrected row, or a restart may report an unacknowledged command as admitted. | Make the in-memory append transactional with the disk write (remove/avoid publishing the entry on every failure path), and add deterministic write-failure tests covering retry before and after a later rebuild/restart for both prompt and steer IDs. |

Verification at `24cab554`: focused remediation suite `34 passed`; broader mobile plus transcript/session regression suite `233 passed`; deterministic reproductions above both failed the delivery contract as described.

**Verdict:** Changes requested. This round is not TERMINAL because F10 and F11 are MAJOR correctness failures.

damianvtran:
- F10 (major) — Fixed in `63432878`: continuation commands now persist explicit `producer_command_id` provenance separate from the ordinary message ID. Only marked producer commands seed the durable admission index; local/seeded/imported user messages, custom/compaction rows, legacy rows, and malformed fragments cannot falsely dedupe. Provenance stays outside model/UI replay objects.
- F11 (major) — Fixed in `63432878`: transcript append publication is transactional. Memory and command indexes update only after write, flush, and fsync succeed; failures truncate to the previous durable boundary/remove failed rebuild fragments. A rejected prompt/steer cannot remain in `_entries` or resurrect during later rebuild/restart. Deterministic open/write/flush failure and retry tests cover prompt and steer identities.

damianvtran:
### Agent review — round 7
- Reviewer: reviewer subagent (openai/gpt-5.6-sol)
- Scope: f4899ae44c38da25ddf6b2f010fe573797d77e81..63432878bc8e40969f902bdb50d1f570be39f1c4
- Delta: 24cab55429aecb32639f8449be9e33ff4ea47a58..63432878bc8e40969f902bdb50d1f570be39f1c4
- Delivery contract: PR #289 acceptance criteria
- Applicable risk surfaces: explicit producer provenance persistence/replay/indexing and namespace isolation; malformed/imported/legacy transcript rows; append write/flush/fsync/rebuild rollback; in-memory entry/index/clock publication; prompt and queued-steer retry identity; steering marker/reservation lifecycle; concurrent append locking; compaction/reconstruction; F1-F11, Q4-Q7, and teardown regressions
- Verdict: Changes requested

| ID | Severity | Finding | Evidence / failure path | Minimum remediation |
|---|---|---|---|---|
| F12 | MAJOR | A steer whose durable append fails is removed from the queue but remains permanently reserved, so the accepted command is lost and an exactly-once retry is rejected. | `Session._drain_steering()` removes the message with `get_nowait()` at `local_operator/session/session.py:3755-3757`, then always removes its producer marker in `finally` at `:3758-3764`; it neither requeues the message nor notifies `CommandReservations` on append failure. The reservation remains `"steer"`, and `reserve()` rejects every retry at `local_operator/mobile/command_reservation.py:61-68`. I reproduced this deterministically: reserve/accept `retry-steer`, inject one `OSError` from `Transcript.append_message`, then observe queue size `0`, producer marker count `0`, no durable admission, and `retry_reserve: False` with reservation state `steer`. This violates the requested retry-after-failure behavior and leaves one of the 32 steer slots leaked until handle disposal. | On append failure, preserve/requeue the same steer and provenance for a later drain, or terminally reject/release its reservation through an explicit failure callback so the producer can retry exactly once; test queue, marker, capacity, retry, and reconstruction behavior after write/flush/fsync failures. |
| F13 | MAJOR | Transcript reconstruction trusts a `producer_command_id` marker before proving the row is a valid producer user message, so malformed/imported rows can still poison the admission namespace. | `_admitted_command_id()` at `local_operator/session/transcript.py:879-885` checks only `type`, defaulted `kind`, and a non-empty marker. It does not require `role == "user"` or successful message decoding. I wrote a parseable JSONL envelope with `type: "message"`, `role: "assistant"`, malformed content, and `producer_command_id: "valid-later-command"`. Reconstruction logged `dropping unparseable transcript message entry`, produced no replay message, but `has_admitted_command()` returned true and a later real prompt reservation returned false. Thus the row is invisible to model/UI yet silently suppresses an ordinary valid command, contrary to the F10 remediation claim that malformed/imported rows cannot falsely dedupe. | Seed admission only from envelopes that also validate as producer-eligible user messages (without exposing the marker in replay), and add malformed-role/content plus imported-marker collision reproductions. |

Verification at `63432878`: focused transcript/session/mobile convergence suite `138 passed`; broader daemon/registrant/attach/TUI regression suite `234 passed`; targeted flake8, Black 26.1.0, and isort checks passed. Deterministic failure reproductions above still violate the delivery contract.

**Verdict:** Changes requested. This round is not TERMINAL because F12 and F13 are MAJOR correctness failures.

damianvtran:
- F12 (major) — Fixed in `3c45689a`: Session now notifies owner handles when an accepted steer fails transcript persistence. Write/flush/fsync failures remove the queued row/provenance, release reservation and capacity, emit observable rejection, settle optimistic queued state, and permit same-ID retry exactly once. Disposal rejects queued producer steers before viewer callbacks detach.
- F13 (major) — Fixed in `3c45689a`: replay indexes producer provenance only after strict message-envelope decoding succeeds and the decoded message is a user message with explicit valid kind and nonblank producer ID. Assistant/system/custom/imported, missing-kind, malformed-content/block, and blank-ID rows cannot claim the command namespace.

damianvtran:
### Agent review — round 8

- Reviewer: reviewer agent (openai/gpt-5.6-sol)
- Scope: f4899ae44c38da25ddf6b2f010fe573797d77e81..3c45689a287a26229cc843eb1f81f66aa94a81b6
- Delta: 63432878bc8e40969f902bdb50d1f570be39f1c4..3c45689a287a26229cc843eb1f81f66aa94a81b6
- Delivery contract: Accepted producer steers reach durable transcript admission exactly once or receive an observable terminal rejection that releases identity and bounded capacity, allowing the same command ID to be retried exactly once. Only strictly decoded user-message envelopes may claim producer IDs, and producer provenance remains absent from model/UI replay.
- Risk surfaces: transcript write/flush/fsync rollback; queued-steer disposal; rejection subscription/unsubscription and observer isolation; reservation/capacity accounting; retry and lost-ACK dedupe; malformed/imported transcript collisions; projection settlement; daemon/registrant/attach teardown regressions.
- Verdict: clean

**No findings.**

F12 is resolved: deterministic write, flush, and fsync failures now terminally reject the undurable steer, remove queued provenance, release its reservation and capacity, settle optimistic queued state, and permit one same-ID retry. The retry becomes durable, reconstructs from disk, and subsequent reuse is rejected. Disposal performs rejection before owner callbacks detach. Rejection observers are isolated, unsubscribe is idempotent, and deterministic callback/capacity probes showed neither post-unsubscribe delivery nor double release/fire.

F13 is resolved: admission reconstruction requires the explicit message kind, a nonblank string producer ID, successful strict message decoding, and `role == "user"`. Assistant, system, custom/imported, missing-kind, malformed-content/block, and blank-ID rows cannot claim the namespace. The envelope marker is removed before message validation/replay, so provenance remains invisible to the model and UI.

Verification at `3c45689a`:
- Focused session/transcript/mobile/daemon/registrant/attach regression suite: `180 passed`.
- Deterministic callback, full 32-slot capacity release, same-ID retry, durable reconstruction/dedupe, and idempotent unsubscribe probes: passed.
- Targeted flake8, Black 26.1.0, and isort checks for all delta files: passed.
- PR head was re-read immediately before posting and remains `3c45689a287a26229cc843eb1f81f66aa94a81b6`.

**Verdict: clean. TERMINAL.** No BLOCKER or MAJOR findings remain; any later minor or nit observation is follow-up and does not require another convergence round.

damianvtran:
### Design review — round 4

- Reviewer: designer agent (openai/gpt-5.6-sol)
- Scope: f4899ae44c38da25ddf6b2f010fe573797d77e81..3c45689a287a26229cc843eb1f81f66aa94a81b6
- Delta: 3f9f971b0f9dfe5d1d97b72baab493f5e2b92a13..3c45689a287a26229cc843eb1f81f66aa94a81b6
- Artifacts: re-confirmed against the round 3 rendered set: `/tmp/attach-shots/transparent-live.svg`, `/tmp/attach-shots/transparent-post-exit.svg`, `/tmp/attach-shots/transparent-connecting-1.svg`, `/tmp/attach-shots/transparent-connecting-2.svg`, `/tmp/attach-shots/qa-approval.svg`, `/tmp/attach-shots/qa-ask.svg`, `/tmp/attach-shots/qa-failure-draft-restored.svg`, `/tmp/mobile-shots/qa-active-previous.png`, `/tmp/mobile-shots/qa-union-search-previous.png`, `/tmp/mobile-shots/qa-previous-open.png`, `/tmp/mobile-shots/qa-previous-open-settled.png`, `/tmp/mobile-shots/qa-wake-transition.png`, `/tmp/mobile-shots/qa-wake-failed.png`, `/tmp/mobile-shots/qa-wake-failed-settled.png`, `/tmp/mobile-shots/r1-qa-failure.png`, `/tmp/mobile-shots/r1-qa-placeholder.png`
- Verdict: Approve

**No findings.**

The complete current diff and the convergence delta were inspected. The mobile presentation files governing the composer, transcript, session list, session view, application shell, and store are blob-identical to the prior design-reviewed head. The retained `Message Local Operator…` placeholder, persistent continuation error treatment, and Active Sessions / Previous Sessions hierarchy therefore remain unchanged.

The post-round-3 TUI delta does not expose lifecycle or process mechanics. Natural-command routing, command identity, transcript durability, reservation accounting, and watcher transitions remain behind the existing conversation surface. The only local attached-view copy added by the delta is the user-invoked `/help` response for `/detach`; it preserves the existing pending-row hierarchy and does not alter the ordinary live, connecting, failure, approval, ask, or settled frames. The rendered round 3 evidence remains representative of those states, including pixel-stable connecting and post-exit frames.

**TERMINAL:** no BLOCKER or MAJOR design findings remain on exact head `3c45689a287a26229cc843eb1f81f66aa94a81b6`.

damianvtran:
### UX review — round 5

- **Reviewer:** ux-reviewer (openai/gpt-5.6-sol)
- **Scope:** f4899ae44c38da25ddf6b2f010fe573797d77e81..3c45689a287a26229cc843eb1f81f66aa94a81b6
- **Mode:** Remediation convergence
- **Delta:** 81c22656f97cea297975953194ade4258ef6bd7c..3c45689a287a26229cc843eb1f81f66aa94a81b6
- **Flows exercised:** Read all PR review and remediation comments and inspected the complete diff, then focused the four-command delta that adds exact-once steer reservations, the durable producer ledger, and terminal persistence-failure rejection/retry. Walked the CSS-bearing production `OperatorApp` paths for idle prompt submission, streaming steer submission, local `/help`, `/detach`, Escape return, and the production `/resume` attach construction. Re-exercised durable Previous subscription across worker generation handoff, prompt-to-steer race transfer, FIFO concurrent steering, and write/flush/fsync steering persistence failures through observable rejection, optimistic projection settlement, capacity release, and same-ID retry. Focused current-head selection passed **14/14**.
- **Interaction confirmation:** Normal messages remain transparent: idle text is admitted as a prompt, mid-turn text is accepted as steering, and persistence bookkeeping does not surface in ordinary copy or transcript replay. `/help` remains local and discovers `/detach`; `/detach` and Escape still return without forwarding text. An open Previous conversation remains subscribed across transparent wake and process-generation handoff. Persistence failure exits the optimistic queued state, emits recoverable feedback, and leaves the same producer command retryable instead of silently losing or duplicating it.
- **Findings:** No findings.
- **Verdict:** **TERMINAL — approve.** The post-round-4 lifecycle, deduplication, and transcript-correctness deltas preserve the reviewed interaction flow. No BLOCKER or MAJOR findings remain; there are no UX follow-ups from this round.

damianvtran:
### QA review — round 4

- **Reviewer:** task QA (openai/gpt-5.6-sol)
- **Scope:** f4899ae44c38da25ddf6b2f010fe573797d77e81..3c45689a287a26229cc843eb1f81f66aa94a81b6
- **Mode:** Remediation convergence
- **Delta:** 81c22656f97cea297975953194ade4258ef6bd7c..3c45689a287a26229cc843eb1f81f66aa94a81b6
- **Environment:** macOS arm64; `/tmp/lop-attach`; isolated transient `/tmp` trees; production owner handles/session/transcript and real loopback registrant teardown paths
- **Verdict:** **PASS**

**No findings.**

F7-F13 converge on the requested focused paths:

- Concurrent same-ID steer returned `steering queued`, `already admitted`; retry remained `already admitted`; the owner steer seam ran once. Concurrent same-ID prompt likewise produced `prompt admitted`, `already admitted` and one owner prompt.
- A producer command compacted out of model history still deduped after `Transcript` reconstruction. Explicit producer provenance remained invisible to replay/model objects.
- A colliding ordinary local user message ID did not claim the producer namespace. Missing-kind, assistant/system, malformed-content/block, blank-ID, custom/imported, and malformed-fragment markers were ignored.
- Steering accepted exactly 32 undurable IDs, rejected the 33rd with bounded backpressure, preserved same-ID dedupe at capacity, and admitted a replacement after durable drain.
- Durable steer rejection released reservation, queued count, and capacity; emitted visible retry feedback; and allowed the same ID to retry once. Injected write, flush, and fsync failures all passed retry, reconstruction, and subsequent dedupe checks.
- Failed transcript write/flush/open paths never entered memory/index and did not resurrect during later rebuild.
- `PYTHONASYNCIODEBUG=1` registrant/attach teardown selection passed `6 passed` with no pending-task, wrong-loop, cross-thread, or RuntimeWarning output.

Verification: targeted F7-F13 selection `24 passed`; deterministic failure matrix `7 passed`; focused freshness suite `169 passed`; all static gates passed. The full unit run had `6685 passed, 7 skipped` and one unrelated timing-sensitive ask-picker failure; that test passed immediately and on five additional isolated runs, so it is recorded as a pre-existing flake rather than a freshness finding.

No QA-created process, listener, socket, lease, pid marker, discovery record, or scratch reconstruction tree remains. Detailed commands and actual responses: `/tmp/pr289-qa-report-r4.md`.

PR #293, opened 2026-08-25
feat(tui): full standard TUI for live-session attach

damianvtran:
## Summary of Changes

Live-session resume now renders the full standard `OperatorApp` instead of the degraded projection screen.

- Protocol v4 adds an opt-in raw `AgentEvent` relay for attach clients plus a bounded `LiveTurnSeed` for correct mid-turn joins. Daemon/mobile clients never receive the new frames.
- `RemoteSession` implements the normal session facade: durable history from the read-only transcript, live events from the authenticated loopback socket, and prompts/steers/model/effort/slash/gate answers routed back to the sole owner.
- In-app `/resume <live>` and cold `lop --resume <live>` both adopt `RemoteSession` through the standard full TUI. There is no banner, attach mode, `/detach`, or visible attach emission.
- Owner death silently re-discovers a replacement owner or attempts normal lease-arbitrated takeover. The winner swaps to a real Session without clearing/replaying the painted transcript; racing losers reattach to the winner.
- Owner-TUI approvals mount the follower's real approval picker and are answered through the owner's existing first-valid-answer path. The follower-only pending overlay does not change daemon/mobile projection bytes.
- Deleted `local_operator/tui/attach_screen.py`, `AttachApp`, `local_operator/cli_attach.py`, the exit-75 attach relaunch, and the old attach-screen tests.

Version: **0.31.1 → 0.32.0**.

## Impact

- **Change class:** C2 standard / pre-authorized. This is the architecture-approved protocol-v4 stepping stone, not the future supervisor rewrite. Rollback is `local-operator==0.31.1`; no data migration.
- **Single writer:** followers instantiate `Transcript` only for read-only replay. The normal `acquire_session_lease` path remains the sole writer arbiter; the two-follower takeover race produced one winner and one `SessionLeaseHeldError`.
- **Mobile compatibility:** daemon connections receive ordinary projection frames only. Raw event and `attach_sync` frames are sent only to authenticated attach connections whose auth frame opts into `events: true`. Old attach clients that omit it retain protocol-v3 projection behavior against a v4 owner.
- **Security:** socket binding remains `127.0.0.1`; the 0600 discovery-record key remains the complete authentication boundary.
- **Lifecycle:** the viewer-independent 3-second drain and registrant connection accounting are unchanged. Event clients remain ordinary attach connections and do not pin idle workers.
- **Mixed version:** a new TUI refuses a protocol <4 live owner with an explicit upgrade/close message rather than falling back to the deleted divergent UI.

### Acceptance criteria

[...]

PR #300, opened 2026-08-25
perf(tui): keep the event loop responsive through boot, sends, and tool calls

damianvtran:
## Summary

Fixes the TUI responsiveness regression reported as three symptoms: startup
appearing to wait for MCP, sends freezing the UI until processed, and tool
calls freezing the TUI. All three were one shape — synchronous work on the
Textual loop — diagnosed and measured in the design audit this implements
(`_prepare` store scans stalling the loop 460-490 ms warm / 2 s cold; cost
pricing doing synchronous HTTP per `message_end`, 418 ms to 13 s on a miss;
bash re-sending its whole accumulated output every 500 ms).

### Fixes

- **A1 — boot scans off the loop**: `sweep_from_config`,
  `backfill_session_origins`, `backfill_session_titles` in `_prepare` now
  run via `asyncio.to_thread`. The lease/claim stay synchronous and first —
  sole-writer ordering (lease before transcript creation) is untouched.
- **A2 — end the perpetual rescan**: a session directory with no journalled
  title records a `title-scan.json` sentinel, so the next boot answers with
  one `stat` instead of re-reading the transcript forever (1,268 of 1,365
  real-store sessions were in that state, 323 ms per boot). The origin
  backfill reads the same sentinel — a title-less session can never grow an
  origin marker either, so it was the residual ~120 ms of the scan. The
  picker and `stored_session_title` never read the sentinel: behaviour
  byte-identical. `write_session_title`'s mtime-preservation and
  best-effort contracts carry over to the sentinel writer.
- **A3 — attach replay off the loop**: `RemoteSession._load_history` builds
  the `Transcript` and `build_llm_history()` in a thread (construction
  included — `Transcript.__init__` eagerly parses the whole file).
  `restore_todos` stays on the loop: it mutates process-private state the
  seed replay must observe.
- **B — MCP off the boot critical path**: `create_session` gains an opt-in
  `defer_mcp_wiring`; only the TUI boot path passes it. The session returns
  immediately and the identical wiring runs as a tracked background task
  completing into `session.mcp_startup`; dispose mid-wiring cancels via a
  dispose hook. Headless/exec callers keep the await-wiring contract — a
  returned session has wiring completed and recorded. Safe because MCP
  tools were already lazy: a turn before servers settle sees the same
  degraded-but-correct surface as a server missing the 250 ms gate today.
- **C1/C2 — pricing never does I/O on the loop**: `turn_cost` resolves
  through `resolve_model_info_paint` — a separate memo fed by the full
  resolver, falling back to the registry row; zero HTTP legs from any paint
  path (the full resolver's `lru_cache` body is itself the discovery path,
  so the paint path must not enter it). A miss fires one gated background
  refresh per model so the real price lands on a following tick. The band
  already renders `None` as unpriceable rather than a wrong number.
- **C3 — bounded bash live snapshot**: `_emit_update` builds its snapshot
  from a 128 KB tail (`_EMIT_SNAPSHOT_BYTES`) instead of the whole
  accumulated stream. The card's live view ingests only the last 64 KB
  (`LIVE_INGEST_CHARS`) anyway. The final-result path is unchanged.

### D — transcript: deliberately NO CHANGE

The obvious lead was investigated and declined: `Transcript._append` fsync
measures **0.074 ms** per append, appends happen ~30x/turn at message
boundaries (never per delta), and the comment at `session/transcript.py`
records that offloading once made things worse. The invariants
(fsync-before-admission-ack, persisted-before-cut, lease ordering, gapless
mid-turn attach) are untouched.

## Cross-instance scope (parallel lop instances)

The operator's parallel-instances report widens the fix to shared surfaces.
Findings (DA-rows, detail in the remediation comment):

- **DA1 (fixed via A1/A2)**: every instance's boot re-scanned the SAME
  shared `sessions/` store — O(N x store). With sentinels, each instance's
  steady-state pass is stat-only: 5-parallel boot scans measure 656 ms ->
  391 ms wall (slowest instance 605 -> 363 ms). Sentinel writes are plain
  atomic renames, no fsync.
- **DA5 (fixed)**: `cached_listing` now takes a best-effort cross-process
  fetch lease (lockfile + pid/token + 60 s expiry, stale-steal). Five
  concurrent cold misses: exactly ONE live listing fetch; losers serve the
  winner's document. Degrades to fetch-anywhere on any coordination
  failure.
- **DA2 (verified)**: the C2 background refresh resolves through the
  SHARED disk catalogue first; a peer's fresh document satisfies the TTL
  with no network call. Per-machine fetches bounded by the 24 h TTL.
- **DA3 (measured, no change)**: registrant projection pushes serialize
  the ~87 KB frame at 0.18 ms per client, capped 20/s by the debounce —
  negligible.
- **DA4 (verified, no change)**: daemon scans of 20 live records cost
  0.82 ms per 2 s cycle; the expensive stale-reap `_durable_projection`
  runs exactly once per stale transition.

## Measurement evidence

`scripts/bench_tui_lag.py` (lag monitor: 5 ms tick probe, stalls from
30 ms) against a 1,000-session synthetic store, 8 MB bash output, discovery
stubbed to 400 ms. Raw output in `bench/before.json` / `bench/after.json`;
run with `PYTHONPATH=. .venv/bin/python scripts/bench_tui_lag.py --json out.json`.

| Scenario | Metric | Before | After |
| --- | --- | --- | --- |
| S1 boot scans | steady-state pass (2nd boot) | 178 ms | **40 ms** (stat-only) |
| S1 boot scans | first pass | 184 ms | 358 ms (one-time migration: writes sentinels) |
| S2 send | `turn_cost` on the loop, cold memo, 400 ms listing | **807 ms block, 812 ms max stall** | **0.6 ms, 0 stalls** |
| S3 bash emit | per-tick body, 8 MB accumulated | 1.7 ms median, O(total) | **0.03 ms, flat; payload 8,323,237 → 131,110 chars** |
| B boot | factory return with 500 ms wiring stub | blocks ≥ 500 ms | **269 ms** (wiring still in flight) |

The design audit's real-store numbers for the same paths: `_prepare`
519-544 ms with 460-2,000 ms stalls; `turn_cost` cold miss 418 ms-13 s; MCP
gate 251 ms before the status band. After: the scans are threaded (no loop
stall), pricing never blocks (0 stalls with a 5 s listing stub in the unit
test), and the model label paints before MCP settles.

## Testing

- New `tests/unit/test_tui_responsiveness.py` (16 tests): per-fix unit
  contracts (A2 sentinel semantics incl. mtime preservation and the
  title-later case; A1 off-loop + thread placement; A3 replay stall; C1
  fast paint path with a 5 s listing stub; C2 refresh lands for the next
  tick; C3 bounded payload), B's four contracts (usable-before-wiring,
  headless still awaits, dispose-during-wiring cancels, existing MCP suite
  unchanged), and the design's S1/S2/S3 end-to-end tests against the real
  `OperatorApp` under the lag monitor.
- Updated pricing test seams in 5 files: they now patch both resolvers,
  because `turn_cost` prices through `resolve_model_info_paint` and a patch
  of the full resolver alone no longer reaches the pricing path.
- Gates: flake8, black (26.1.0), isort, pyright — all clean.
  `env -u NO_COLOR TERM=xterm-256color .venv/bin/python -m pytest tests/unit -q`:
  6,761 passed. (Two known pre-existing flakes reproduce on clean `main`
  under machine load: `test_dispose_closes_auth_store` and the
  `test_subagent_stats`/`test_cost_aggregation` timing pair — verified via
  `git stash` A/B on unmodified HEAD; not introduced here.)

## Rollout notes

- A2's first boot after upgrade writes one small sentinel per answered
  directory (one-time migration cost, visible in the table).
- C1 means the band can briefly show "unpriceable" for an uncached price —
  damped to one tick per model by C2's gated background refresh.
- `defer_mcp_wiring` is opt-in and default-off; only `cli.py`'s two TUI
  factory call sites pass it.

### DIFF
--- a/bench/README.md
+++ b/bench/README.md
@@ -13,7 +13,7 @@ editable install pointing elsewhere.
 
 ## before.json / after.json
 
-Captured on this machine (macOS/APPS) against `origin/main` @ `f2a52b53`
+Captured on this machine (macOS/APFS) against `origin/main` @ `f2a52b53`
 (before) and the same tree with the fixes (after), 1,000-session synthetic
 store, 8 MB bash output, discovery stubbed to 400 ms.
 
@@ -45,3 +45,46 @@ discovery stub: with `defer_mcp_wiring=True` the factory returned in
 the TUI can adopt the session and paint the model label before the MCP
 segment fills. On the unfixed tree the same call blocked for the stub's
 full 500 ms plus the real 250 ms gate before returning.
+
+## Parallel instances (cross-instance scope)
+
+`p_parallel_boot`: five concurrent scan passes over one shared 1,000-session
+store — the shape of five lop instances booting together.
+
+| Metric | Before | After |
+| --- | --- | --- |
+| First boot (one-time migration) | 2,276 ms | 1,415 ms |
+| 5-parallel steady-state wall | 656 ms | 391 ms |
+| Slowest instance (steady) | 605 ms | 363 ms |
+
+Every instance's steady-state pass drops to stat-only (A2 sentinels, both
+title and origin), so N parallel instances cost N x ~40 ms of stats instead
+of N x full-store reads. The remaining wall time is the shared-directory
+I/O itself, not redundant parsing.
+
+Cross-instance findings (DA-rows in the remediation):
+
+- **DA1 (fixed, A1+A2)**: O(N x store) redundant boot scans eliminated.
+  Sentinel writes are plain `write_text`+`replace` — no fsync, so no
+  cross-instance fsync amplification. The to_thread'ed scans run in the
+  default executor (no shared asyncio.Lock between them; the only
+  serialization is per-instance sequential awaits, preserved deliberately).
+- **DA2 (verified, no change)**: C2's background refresh resolves through
+  `available_models` -> `cached_listing`, which reads the SHARED disk cache
+  first; a peer's fresh document satisfies the TTL and no network call
+  happens. Per-machine fetch frequency is bounded by the 24 h TTL.
+- **DA3 (measured, no change)**: registrant projection push serializes the
+  ~87 KB frame per client at 0.18 ms each, capped at 20 pushes/s by the
+  0.05 s debounce — five followers cost ~1.8% of the registrant loop. Not
+  a bottleneck.
+- **DA4 (verified, no change)**: daemon scan of 20 live records measures
+  0.82 ms per 2 s cycle; stale-reap `_durable_projection` (~287 ms for a
+  500-entry transcript) runs exactly once per stale transition (the stale
+  record file is unlinked by `registry.scan`; the vanished-pid arm is
+  gated by `entry.ended`).
+- **DA5 (fixed)**: `cached_listing` now takes a best-effort cross-process
+  fetch lease (lockfile with pid+token and 60 s expiry, `O_CREAT|O_EXCL`
+  take, stale-steal, holder-only release). Five concurrent cold misses
+  measured: exactly ONE live fetch, all five served. Degrades to
+  fetch-anywhere on any coordination failure — a read-only cache dir can
+  never block a session start.
--- a/bench/README.md
+++ b/bench/README.md
@@ -1,0 +1,47 @@
+# TUI responsiveness benchmarks (S1/S2/S3)
+
+`scripts/bench_tui_lag.py` reproduces the three reported symptoms with a
+loop-lag monitor (5 ms tick probe, stalls recorded from 30 ms). Run:
+
+```sh
+PYTHONPATH=. .venv/bin/python scripts/bench_tui_lag.py --json out.json
+```
+
+`PYTHONPATH=.` matters: the script must import THIS checkout's
+`local_operator`, and a plain run from another directory can resolve an
+editable install pointing elsewhere.
+
+## before.json / after.json
+
+Captured on this machine (macOS/APPS) against `origin/main` @ `f2a52b53`
+(before) and the same tree with the fixes (after), 1,000-session synthetic
+store, 8 MB bash output, discovery stubbed to 400 ms.
+
+| Scenario | Metric | Before | After |
+| --- | --- | --- | --- |
+| S1 boot scans | first pass (writes markers) | 184 ms | 358 ms (one-time: writes 1,000 sentinels) |
+| S1 boot scans | second pass (steady state) | 178 ms | **40 ms** (stat-only) |
+| S1 boot scans | loop stalls > 30 ms | n/a (was on the loop: 460-490 ms warm, 2 s cold, per the design audit) | 0 |
+| S2 send (pricing) | `turn_cost` from the loop, cold memo, hostile listing | **807 ms loop block, 812 ms max stall** | 0.6 ms, 0 stalls |
+| S3 bash emit | per-tick body cost, 8 MB accumulated | 1.7 ms median (O(total); 32 MB measures 4.7 ms and grows) | 0.03 ms, flat |
+| S3 bash emit | live payload per update | 8,323,237 chars | 131,110 chars (bounded) |
+
+The before/after first-pass inversion in S1 is expected and correct: the
+fixed tree's first pass performs the one-time migration (writes one
+`title-scan.json` sentinel per answered directory). Every later boot is the
+second-pass number, which is the steady state a user lives in.
+
+The design audit's own measurements on the real store (1,365 sessions,
+431 MB) for the same code paths: `_prepare` 519-544 ms with a 460-2,000 ms
+loop stall; `turn_cost` cold-miss 418 ms (2 s budget branch) to 13 s
+(unlisted model); boot to model label delayed by the 251 ms MCP gate. The
+synthetic numbers above are the reproducible proxy for the same invariants.
+
+## Boot-to-label with deferred MCP wiring (fix B)
+
+Measured with the real `create_session` (hosting `test`) and a 500 ms
+discovery stub: with `defer_mcp_wiring=True` the factory returned in
+**269 ms** while wiring was still in flight (the stub had not settled), so
+the TUI can adopt the session and paint the model label before the MCP
+segment fills. On the unfixed tree the same call blocked for the stub's
+full 500 ms plus the real 250 ms gate before returning.
--- a/bench/after.json
+++ b/bench/after.json
@@ -1,0 +1,23 @@
+{
+  "s1_boot": {
+    "sessions": 1000,
+    "first_pass_ms": 358.3,
+    "second_pass_ms": 39.7,
+    "max_stall_ms": 0.0,
+    "stalls_over_30ms": 0
+  },
+  "s2_send": {
+    "discovery_latency_s": 0.4,
+    "loop_ms": 0.6,
+    "max_stall_ms": 0.0,
+    "cost": null,
+    "discovery_calls": 2
+  },
+  "s3_bash_emit": {
+    "total_mb": 8.0,
+    "emit_body_ms_median": 0.03,
+    "emit_body_ms_max": 0.04,
+    "live_payload_chars": 131110,
+    "max_stall_ms": 0.0
+  }
+}
--- a/bench/after.json
+++ b/bench/after.json
@@ -1,22 +1,29 @@
 {
   "s1_boot": {
     "sessions": 1000,
-    "first_pass_ms": 358.3,
-    "second_pass_ms": 39.7,
+    "first_pass_ms": 1475.9,
+    "second_pass_ms": 48.5,
     "max_stall_ms": 0.0,
     "stalls_over_30ms": 0
   },
+  "p_parallel_boot": {
+    "instances": 5,
+    "sessions": 1000,
+    "first_pass_ms": 1123.4,
+    "parallel_steady_wall_ms": 391.3,
+    "slowest_instance_ms": 362.5
+  },
   "s2_send": {
     "discovery_latency_s": 0.4,
-    "loop_ms": 0.6,
+    "loop_ms": 0.82,
     "max_stall_ms": 0.0,
     "cost": null,
     "discovery_calls": 2
   },
   "s3_bash_emit": {
     "total_mb": 8.0,
-    "emit_body_ms_median": 0.03,
-    "emit_body_ms_max": 0.04,
+    "emit_body_ms_median": 0.09,
+    "emit_body_ms_max": 0.13,
     "live_payload_chars": 131110,
     "max_stall_ms": 0.0
   }
--- a/bench/before.json
+++ b/bench/before.json
@@ -1,0 +1,22 @@
+{
+  "s1_boot": {
+    "sessions": 1000,
+    "first_pass_ms": 184.2,
+    "second_pass_ms": 178.5,
+    "max_stall_ms": 0.0,
+    "stalls_over_30ms": 0
+  },
+  "s2_send": {
+    "discovery_latency_s": 0.4,
+    "loop_ms": 806.91,
+    "max_stall_ms": 812.2,
+    "cost": null,
+    "discovery_calls": 2
+  },
+  "s3_bash_emit": {
+    "total_mb": 8.0,
+    "tick_ms_median": 1.73,
+    "tick_ms_max": 5.36,
+    "max_stall_ms": 0.0
+  }
+}
--- a/local_operator/cli.py
+++ b/local_operator/cli.py
@@ -1422,18 +1422,26 @@ async def create_session(
     agent_registry: "AgentRegistry",
     *,
     has_ui: bool = False,
+    defer_mcp_wiring: bool = False,
 ):
     """Build a wired harness session for interactive/headless use.
 
     Thin facade over :func:`local_operator.session_factory.create_session`
     (the composition root shared with ``exec`` and the background worker).
     The engine import is lazy so importing ``cli`` never pulls in
-    providers/session internals.
+    providers/session internals. ``defer_mcp_wiring`` passes through to the
+    factory's TUI-boot opt-in unchanged (see its docstring for why only a
+    full front end may take it).
     """
     from local_operator.session_factory import create_session as _create_session
 
     return await _create_session(
-        args, config_manager, credential_manager, agent_registry, has_ui=has_ui
+        args,
+        config_manager,
+        credential_manager,
+        agent_registry,
+        has_ui=has_ui,
+        defer_mcp_wiring=defer_mcp_wiring,
     )
 
 
@@ -2333,6 +2341,7 @@ def main() -> int:
                     credential_manager,
                     agent_registry,
                     has_ui=True,
+                    defer_mcp_wiring=True,
                 )
 
             # The provider controller gives the TUI the full provider/model/
@@ -2386,6 +2395,7 @@ def main() -> int:
                         credential_manager,
                         agent_registry,
                         has_ui=True,
+                        defer_mcp_wiring=True,
                     )
 
                 tui_entry = functools.partial(tui_entry, resume_factory=resume_factory)
--- a/local_operator/model/catalogue.py
+++ b/local_operator/model/catalogue.py
@@ -52,6 +52,7 @@ import logging
 import os
 import tempfile
 import time
+import uuid
 from pathlib import Path
 from typing import Any, Callable
 
@@ -212,6 +213,106 @@ def _umask() -> int:
     return current
 
 
+#: How long a catalogue fetch lease stands before a peer may steal it. A
+#: listing fetch is bounded by the caller's own timeout (tens of seconds at
+#: worst), so a lease comfortably longer covers a slow response, and a
+#: process that dies mid-fetch cannot strand the document: the lease lapses
+#: and the next session fetches.
+_LISTING_LEASE_S = 60.0
+
+#: How long a lease LOSER waits for the winner's write before re-reading.
+#: Bounded hard because the caller can be a session boot: this is a
+#: courtesy pause, not a barrier.
+_LISTING_LEASE_WAIT_S = 2.0
+
+
+class _ListingFetchLease:
+    """A best-effort cross-process lock on one catalogue document.
+
+    Modelled on the usage cache's SQLite lease but lockfile-based, because
+    this module's store is plain files and pulling SQLite in here would put
+    a database dependency on the model-picker path. ``O_CREAT | O_EXCL`` is
+    the atomic take; the file carries the holder's identity and an expiry
+    so a crashed holder cannot block the next refresh. EVERY failure mode
+    degrades to "no lease": an unwritable directory, a race lost, an
+    unreadable holder file — the caller then just fetches, which is the
+    pre-lease behaviour and always correct, merely less polite to the
+    endpoint.
+    """
+
+    def __init__(self, document: Path) -> None:
+        self._path = document.with_name(document.name + ".fetching")
+        self._token = f"{os.getpid()}:{uuid.uuid4().hex[:8]}"
+        self._held = False
+
+    def acquire(self) -> bool:
+        """Take the lease if free or expired. True means THIS process fetches."""
+        try:
+            self._path.parent.mkdir(parents=True, exist_ok=True)
+            fd = os.open(self._path, os.O_CREAT | os.O_EXCL | os.O_WRONLY, 0o644)
+        except FileExistsError:
+            # Standing lease. Legitimate only while unexpired; a stale one
+            # (holder crashed mid-fetch) is stolen by replacing it, where the
+            # same O_EXCL race decides the single winner among the stealers.
+            try:
+                raw = self._path.read_text(encoding="utf-8", errors="replace")
+                expires = float(json.loads(raw).get("expires_at", 0.0))
+            except (OSError, ValueError):
+                expires = 0.0
+            if expires > time.time():
+                return False
+            try:
+                self._path.unlink()
+            except OSError:
+                return False
+            return self.acquire()
+        except OSError:
+            # Unwritable cache dir or worse: fetch without a lease. Correct,
+            # just unpolite.
+            return True
+        try:
+            payload = json.dumps(
+                {"holder": self._token, "expires_at": time.time() + _LISTING_LEASE_S}
+            )
+            os.write(fd, payload.encode("utf-8"))
+        finally:
+            os.close(fd)
+        self._held = True
+        return True
+
+    def await_peer_briefly(self) -> None:
+        """Give the lease holder a moment, then return whatever is on disk.
+
+        Polls the DOCUMENT, not the lease: the winner's atomic rename is the
+        event we are waiting for, and a holder that fails leaves the lease
+        standing until expiry — waiting for the lease would mean waiting out
+        the full lease TTL on every failed peer fetch.
+        """
+        document = self._path.with_name(self._path.name[: -len(".fetching")])
+        deadline = time.monotonic() + _LISTING_LEASE_WAIT_S
+        while time.monotonic() < deadline:
+            time.sleep(0.05)
+            if document.exists():
+                try:
+                    if time.time() - document.stat().st_mtime < _LISTING_LEASE_WAIT_S * 2:
+                        return
+                except OSError:
+                    return
+
+    def release(self) -> None:
+        """Free the lease if this process holds it."""
+        if not self._held:
+            return
+        self._held = False
+        try:
+            raw = self._path.read_text(encoding="utf-8", errors="replace")
+            if json.loads(raw).get("holder") == self._token:
+                self._path.unlink()
+        except (OSError, ValueError):
+            # Already gone or unreadable: nothing to free.
+            pass
+
+
 def cached_listing(
     key: str,
     fetch: Callable[[], dict[str, Any]],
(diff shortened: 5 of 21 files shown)


damianvtran:
### Agent review — round 1
- Reviewer: reviewer subagent (claude-fable-5)
- Scope: f2a52b5312a06c91d3bfeb284680636f75ef8593..aac29a5827c9f3b6822c0c277d177604301dd19a
- Mode: Comprehensive
- Delivery contract: TUI responsiveness fix per /tmp/lop-freeze-design.md (S1/S2/S3), invariants 1-6 preserved
- Applicable risk surfaces: A1 thread-safety of store scans (claim/lease ordering verified synchronous-first, session_factory.py:1017-1033 before the to_thread at 1058); A2 sentinel vs `stored_session_title`/picker (sentinel is a separate file neither reads — verified); A3 attach replay ordering vs seed/relay handoff (raced — F2); B deferred wiring: dispose-mid-wiring, /mcp against wiring-in-progress (degrades to "unknown" rows, fine), headless path unchanged (`defer_mcp_wiring` passed only from cli.py:2344/2398, exec/serve untouched), error reporting reaching the TUI (severed — F1); C1/C2 cache-only resolution + no discovery reachable from `on_context_usage_reported`/`_poll_subagents`/`turn_cost`/`job_cost` (verified: all app.py pricing goes through tui/costs.py, which only calls `resolve_model_info_paint`), refresh stampede gating (`_paint_refreshing` verified), band flapping (partially unhandled — F4); C3 final-result path unchanged (only `_emit_update` touched), redaction applied to the tail, straddling-secret fragment falls outside the card's 64 KB ingest window (verified against the 128 KB producer bound); tests: ran `tests/unit/test_tui_responsiveness.py` (16 passed) — the stall probes are non-tautological (the `_prepare` test also asserts off-loop thread identity; S2/S3 stub discovery to 5 s, so old code blows the 250 ms loaded ceiling deterministically); bench before/after JSON consistent with the README's claims including the explained first-pass inversion.
- Verdict: Changes requested

| --- | --- | --- | --- |
| F1 | BLOCKER | B: deferred wiring must complete "into session.mcp_startup and the existing refresh_selected/on_startup_settled hooks" — the TUI half of that contract is severed on every boot | The TUI installs ALL of its MCP hooks inside `_wire_mcp_status`, which early-returns when `session.mcp_manager` is None (tui/app.py:4147-4150) — and under `defer_mcp_wiring=True` (now every TUI boot, cli.py:2344/2398) adoption runs before wiring lands, so the manager is always None at that point. Consequently: (1) `session._on_mcp_startup_settled` (app.py:4220) is never installed, so when the background task settles, the factory's sink lookup (session_factory.py:1408) finds None and the startup toast + transcript failure notice never fire — a user whose MCP server fails at boot gets silence where they got a toast and a durable notice before this PR; `_report_mcp_startup` at adopt (app.py:2034) also no-ops because `mcp_startup` is still None then. (2) `set_on_tools_changed`/`on_auth_required` (app.py:4180/4195) are never chained, so the band's MCP segment stays permanently empty (nothing ever calls `_refresh_mcp_status` after wiring lands), mid-session OAuth-expiry toasts are lost, and `_measure_preloaded_context` is never re-run after the MCP schema merge, leaving the context estimate wrong for the session. Tools still merge (factory-side `refresh_selected`) and `/mcp` works on demand, but startup-failure reporting and the live MCP band are regressed for every MCP-configured user on the default path. No test drives the real TUI against a deferred session (the deferred-wiring tests assert session attributes only; the S1 e2e test uses FakeSession). Required: make wiring completion re-enter the TUI — e.g. install the `_on_mcp_startup_settled` sink at adopt BEFORE the manager-None early return, and have that sink (via `call_later`) run `_wire_mcp_status(current)` + `_report_mcp_startup(current)` once the manager exists; add a test that adopts a deferred session in `run_test` and asserts the settle toast and band segment land. |
| F2 | MAJOR | A3 / invariant 4 (gapless mid-turn attach): threaded replay widens the relay-dedup race from ~zero to the full replay duration | `RemoteSession._load_history` now yields to the loop for the whole transcript read (`await asyncio.to_thread(_replay)`, session/remote.py:235). Relay frames are deduplicated against `self._history_ids` at ARRIVAL time (`_on_wire_event`, remote.py:294) and `_drain_buffered_events` (remote.py:274-280) never re-checks. Before this PR `_load_history` ran synchronously after `_dial` with no award point, so `_history_ids` was bound before any pump callback could interleave; now every relay frame delivered during the replay window — seconds, for exactly the large-transcript case this change targets — is checked against an EMPTY id set and buffered. A message that lands durably in that window (mid-turn attach to a busy owner, the #293 headline case) is then painted twice: once from the replayed history, once from the buffered relay event. Required: after binding `self._history_ids` (remote.py:237), re-filter `self._buffered_events` by message id (or move the dedup check from arrival to drain/emit time). |
| F3 | MINOR | A2: the title-scan sentinel suppresses the ORIGIN backfill for directories the origins pass never actually scanned | resume.py:453 skips sentineled dirs in `backfill_session_origins` on the claim that any sentineled dir was already answered "not a subagent" — but the sentinel is written by a different pass (titles, resume.py:607) with its own traversal, and origins breaks at `stamped >= limit` (resume.py:435, limit=500 stamps). On a first post-upgrade boot with >500 stampable pre-marker subagent dirs, the 501st is skipped by origins, sentineled by titles the same boot, and then skipped by origins FOREVER — pre-PR it would have been stamped on the next boot. Narrow (needs a large pre-upgrade backlog), but it converts a bounded delay into a permanent miss. Required: in the origins pass, only honour the sentinel when the origin question was genuinely answered — e.g. have the titles pass write the sentinel only for dirs already bearing `ORIGIN_NAME`, or key the origins skip on its own marker. |
| F4 | MINOR | C2: the background refresh fires on "unpriced", not on "paint-memo miss", so a priced registry row is never re-validated against discovery from the paint path | `_resolve_for_paint` (tui/costs.py:140-146) triggers `refresh_model_info_background` only when the fallback carries no price. `_paint_memo` is keyed by TTL bucket (configure.py:1408, DEFAULT_TTL_S = 24 h), so when the bucket rolls over mid-session, every paint resolution misses and falls back to the static registry row — and for a model whose registry row carries a price (possibly stale vs the provider listing that priced it until that moment), no refresh ever fires and no non-paint caller re-resolves mid-session. The band silently switches from the discovery price to the registry price for the rest of the session — the "confident wrong number" the module's own docstring calls the more expensive lie. Required: fire the background refresh on a paint-memo MISS (before falling back to the registry), keeping the `_paint_refreshing` gate as the stampede bound. |
| F5 | NIT | Dead code | session_factory.py:1057 re-imports `claim_session` (already imported at 1009) alongside `sweep_from_config`; drop the duplicate. |
| F6 | NIT | Slow leak | `_paint_memo` (configure.py:1385) keys on `(provider, model_id, bucket)` and never evicts prior buckets, so a long-lived process accretes one dead entry per model per day; evict stale-bucket keys on write. |

#### Follow-up observations

- `refresh_model_info_background`'s no-running-loop branch (configure.py, `except RuntimeError: _refresh()`) runs the full resolver INLINE — a latent 13 s block if a synchronous caller ever adopts this "paint-safe" API. All current callers are on the loop; worth a docstring warning or an outright raise.
- `_write_title_scan_sentinel`'s stat→write→utime mtime restore can race a live owner appending to the same directory in the window (the sentinel is written precisely for sessions without titles, which includes a just-claimed live session); same pre-existing hazard as the title sidecar writer it mirrors, so noted, not gated.
- `tests/unit/test_session_factory.py::test_dispose_closes_auth_store` fails (`assert 2 == 1` AuthStore instances) — reproduced identically on base f2a52b53, so pre-existing, not introduced here.
- bench/README.md says "macOS/APPS" for APFS.

damianvtran:
- F1 (blocker) — Fixed in `0a502bbe`: the TUI now learns when deferred MCP wiring lands. `_wire_mcp_status` installs the settle sink BEFORE the manager-None early return (`session._set_mcp_startup_sink`, a new guarded accessor on Session); the factory fires the sink at all three wiring completion points (gate snapshot + both degradation arms) via `_fire_mcp_sink`, gated on `_deferred_boot` so the synchronous path keeps its settle-only contract (the existing factory test pins that and still passes). The sink re-enters the app through `call_later`: installs the live `on_tools_changed`/`on_auth_required` chains, refreshes the band's MCP segment, re-runs `_measure_preloaded_context`, and raises the boot toast/notice. New e2e test `test_tui_wires_mcp_status_when_deferred_wiring_lands` drives the real `OperatorApp` against a real deferred-boot `create_session` and asserts the sink is installed at adoption, the band segment reads the landed manager (`configured=1, failed=True`), and the startup toast fires for the failed server.
- F2 (major) — Fixed in `0a502bbe`: `RemoteSession._load_history` re-filters `_buffered_events` against the replayed `_history_ids` (`_filter_known_messages`) the moment the threaded replay binds them, closing the arrival-time-vs-drain-time dedup gap. Regression test `test_relay_frame_during_threaded_replay_is_not_double_painted` emits message-grade relay frames mid-replay for a message the replay contains and asserts exactly one `message_start`/`message_end` reaches handlers.
- F3 (minor) — Fixed in `0a502bbe`: the origin sweep no longer honours the title sweep's sentinel. It writes its own `origin-scan.json` (same best-effort/mtime-preserving contract), so a directory the origin pass never reached (past its 500-stamp cap) cannot be permanently suppressed by a title sentinel from the same boot. Verified: a 600-stampable-subagent backlog behind the 500 cap converges fully across two boots (500 then 100), and steady state stays stat-only.
- F4 (minor) — Fixed in `0a502bbe`: `resolve_model_info_paint` now returns `(info, memo_hit)` and `_resolve_for_paint` fires the background refresh on a memo MISS (not only an unpriced row), so a TTL-bucket rollover mid-session re-validates a priced registry row against discovery off-loop instead of silently switching the band to the possibly-stale registry price. The `_paint_refreshing` gate remains the stampede bound.
- F5 (nit) — Fixed in `0a502bbe`: duplicate `claim_session` import dropped; `sweep_from_config` is imported once at its use site.
- F6 (nit) — Fixed in `0a502bbe`: `_paint_memo` evicts stale-bucket keys once it exceeds 64 entries, bounding a long-lived process at the current bucket's models instead of one dead entry per model per day.
- Follow-up (inline `_refresh()` branch) — Fixed in `0a502bbe`: `refresh_model_info_background` with no running loop now logs a warning and skips (gate released) instead of running the full 13 s resolver inline; a synchronous caller adopting the API gets a loud no-op, not a silent block.
- Follow-up (sentinel mtime race) — Rejected: same pre-existing hazard as the title sidecar writer it mirrors (stat→write→utime against a live owner appending in the same directory); the window is microseconds and the sentinel is best-effort by contract. Not widened by this change.
- Follow-up (`test_dispose_closes_auth_store`) — Confirmed pre-existing on base `f2a52b53` (reproduced identically via `git stash` A/B); not introduced here.
- Follow-up (bench README "APPS") — Fixed in `0a502bbe`: now reads APFS.

Cross-instance scope (operator addendum; measurements in `bench/README.md` and `bench/after.json`):

- DA1 (major) — Fixed in `0a502bbe` (via the in-flight A1/A2): N parallel instances each re-scanned the shared store at boot. Measured 5-parallel steady-state over a 1,000-session fixture: wall 656 ms → 391 ms, slowest instance 605 ms → 363 ms, every pass now stat-only. Sentinel writes are plain `write_text`+`replace` — no fsync, so no cross-instance fsync amplification. The to_thread'ed scans share no asyncio.Lock; the per-instance sequential awaits are deliberate (lease/claim ordering stays synchronous-first).
- DA2 (info) — Verified, no change: the C2 background refresh resolves through `available_models` → `cached_listing`, which reads the shared disk document first; a peer's fresh write satisfies the TTL with no network call. Per-machine fetch frequency is bounded by the 24 h TTL.
- DA3 (info) — Measured, no change: registrant projection pushes serialize the ~87 KB frame at 0.18 ms per client, capped at 20/s by the 0.05 s debounce; five followers cost ~1.8% of the registrant's own loop. Not a bottleneck.
- DA4 (info) — Verified, no change: `registry.scan` over 20 live records measures 0.82 ms per 2 s daemon cycle; the ~287 ms `_durable_projection` on stale-reap runs exactly once per stale transition (the stale record file is unlinked by `scan`; the vanished-pid arm is gated by `entry.ended`).
- DA5 (major) — Fixed in `0a502bbe`: `cached_listing` takes a best-effort cross-process fetch lease (`_ListingFetchLease`: lockfile with pid+token and 60 s expiry, `O_CREAT|O_EXCL` take, stale-steal by the same race, holder-only release, 2 s bounded loser wait polling the document). Five concurrent cold misses measured: exactly ONE live fetch, all five callers served. Every failure mode (unwritable dir, unreadable holder, lost race) degrades to the pre-lease fetch-anywhere behaviour — nothing can block a session start. Tests: `test_concurrent_cold_misses_fetch_exactly_once`, `test_a_lease_from_a_dead_holder_is_stolen_after_expiry`.

Gates on `0a502bbe`: flake8 clean; black 26.1.0 clean; isort clean; pyright 0 errors; `env -u NO_COLOR TERM=xterm-256color .venv/bin/python -m pytest tests/unit -q` → 6,765 passed, 7 skipped (one earlier run showed 1-2 assorted timing flakes — `test_comms`/`test_loop`/`test_tui_ask` — each reproduced as load-dependent and passing in isolation on both this branch and clean `f2a52b53`).

damianvtran:
### Agent review — round 2
- Reviewer: reviewer subagent (claude-fable-5)
- Scope: f2a52b5312a06c91d3bfeb284680636f75ef8593..1530892ffdfab97979b2d46bcf316164e98914e1
- Delta reviewed: aac29a5827c9f3b6822c0c277d177604301dd19a..1530892ffdfab97979b2d46bcf316164e98914e1
- Delivery contract: TUI responsiveness fix (S1/S2/S3) + cross-instance daemon/store bottlenecks, invariants preserved
- Applicable risk surfaces: F1 sink wiring ordering (verified no await between the gate-point `_fire_mcp_sink` and `wire_mcp_into_session`'s return, and `attach_mcp_dispose` sets `session.mcp_manager` in the same task step — so the `call_later`'d `_rewire` always sees the manager; both degradation arms also fire the sink, and the first `_rewire` replaces the deferred sink with the ordinary settle reporter, so gate+settle do not double-report); F2 drain-time re-filter (loop-affine: ids bind and `_filter_known_messages` runs with no await between, arrival-time check covers post-bind frames; non-message events deliberately kept); F3 origin sentinel separation (own file, only its own pass writes it, sentinel writes not counted against the stamp limit so convergence across boots holds); F4 tuple-return + refresh-on-memo-miss (all callers are in tui/costs.py and unpack the tuple; the off-loop `RuntimeError` branch now warns, releases the gate, and skips — no inline 13 s resolver); F5/F6 confirmed in the diff; DA5 lease (O_EXCL take, holder-token release, 2 s document-mtime poll, every failure arm returns rather than blocks — no path can stall a session start; lease name `<key>.json.fetching` does not match the `*.models.json` purge glob, so the sweep cannot eat a live lease; `_write_cache`'s mkstemp keeps even duplicate winners from corrupting the document); DA1 (sentinels are plain `write_text`+`replace`, no fsync); DA4 spot-checked (`registry.scan` unlinks a stale record on first sighting, so the reap-side projection cannot re-fire for the same record; the wedged arm keeps the file but is not the stale arm); ran `tests/unit/test_tui_responsiveness.py` + `tests/unit/model/test_catalogue.py` (107 passed) including the four new pinning tests. Rendered-UI check: confirmed — the delta adds no new copy/layout/widget; the deferred-path toast and band segment are the PRE-EXISTING surfaces restored to parity with the synchronous path (their absence was the F1 regression), and the only other visible delta remains the transient one-tick-later cost figure. No design round owed.
- Verdict: Approve

| --- | --- | --- | --- |
| R2-1 | MINOR | DA5: the stale-steal `unlink` can delete a FRESH winner's lease, so the docstring's "the same O_EXCL race decides the single winner among the stealers" is not what the code does | `_ListingFetchLease.acquire` (model/catalogue.py:249-259): two stealers A and B both read the expired lease; A unlinks, re-creates via O_EXCL and writes its token; B's `unlink` then removes A's brand-new lease file unconditionally (nothing re-checks whose file it is), B's recursive `acquire` wins O_EXCL on the now-empty name, and BOTH fetch — and with A's lease gone, a third late arrival can win too. Same shape in `release()` (read-check-unlink is not atomic, reachable only after a >60 s fetch overruns its own lease). Consequence is bounded and inside the stated degradation contract — duplicate polite-endpoint fetches, never a block, and mkstemp keeps concurrent writers from corrupting the document — so this is a politeness/accuracy defect, not a safety one. Cheapest honest fix: re-stat the lease between the expiry read and the unlink and skip the steal if mtime moved (shrinks the window to ~nothing), and correct the docstring to say the steal race is decided by unlink+O_EXCL best-effort, not exactly-once. |
| R2-2 | MINOR | DA5: the "exactly ONE live fetch" property is conditional on the winner finishing inside the 2 s loser wait when no document exists on disk | `cached_listing` (model/catalogue.py:354-360): a loser whose `await_peer_briefly` expires with NO document falls through and fetches leaselessly — deliberate ("a cold machine must still start") and the right call against the no-block constraint, but it means the herd protection evaporates exactly when the endpoint is slow, which is the case most likely to 429. The measured 1-fetch result and `test_concurrent_cold_misses_fetch_exactly_once` both hold only because the stub fetch completes within the losers' 2 s windows (release fires ~0.45 s in). No code change required for this round — the tradeoff is explicit in the comments — but the bench/README claim should say "one fetch when the winner completes within the loser wait"; anything stronger over-promises. |
| R2-3 | NIT | Doc accuracy | `Session._set_mcp_startup_sink`'s docstring (session/session.py:2880) names `dispose` as the second writer ("to drop a sink that would otherwise fire into a torn-down app"), but no dispose path clears `_on_mcp_startup_settled` — the actual guard is the wiring task's cancellation hook. Either add the clear in `dispose` or fix the docstring to name the real mechanism. |
| R2-4 | NIT | Idempotence edge | If the manager settles in the sub-millisecond window between the gate-point `_fire_mcp_sink` and the pump processing the first `_rewire`, the still-installed deferred sink fires a second `call_later(_rewire)`, and the second `_wire_mcp_status` re-chains `on_tools_changed` on top of the first wrap — every later tools-changed then double-schedules the repaint and `_measure_preloaded_context`. Harmless duplicates, vanishing window; a `_rewire`-ran-once flag would close it if anyone ever cares. |
| R2-5 | NIT | Test pinning strength | `test_relay_frame_during_threaded_replay_is_not_double_painted` asserts `count("message_start") <= 1`, which also passes if the filter over-drops to zero; assert the exact expected count. `test_tui_wires_mcp_status_when_deferred_wiring_lands`'s toast check `"broken" in shown or "MCP" in shown` passes for ANY MCP toast; drop the `or` arm. Both tests otherwise genuinely pin the fixed behaviours (real `OperatorApp`, real `create_session`, mid-replay frame injection). |

#### Convergence assessment

- F1 (blocker) — CLOSED. Sink installed before the manager-None early return, fired at all three factory completion points on the deferred path only, ordering sound (no await between fire and `mcp_manager` assignment, so the `call_later`'d rewire cannot observe a missing manager), synchronous path's settle-only contract preserved, and the new e2e test drives the real app against a real deferred boot.
- F2 (major) — CLOSED. Drain-time re-filter against the bound ids, loop-affinity makes it race-free, regression test injects frames into the instrumented replay window.
- F3 (minor) — CLOSED. Origin pass owns its own sentinel; a title sentinel can no longer suppress a directory the origin pass never reached; sentinel writes are not counted against the stamp limit, so a >500 backlog converges across boots.
- F4 (minor) — CLOSED. Refresh now fires on memo miss (rollover included), callers unpack the tuple, and the inline-resolver footgun in the no-loop branch was replaced with a loud skip.
- F5/F6 (nits) — CLOSED, confirmed in the diff.
- DA1–DA4 — claims spot-checked and consistent with the code (DA4's once-per-stale-transition holds because `scan` unlinks the stale record on first sighting).
- DA5 — the lease meets its hard requirements (no path blocks a session start; O_EXCL take; holder-only release intent; bounded loser wait); R2-1/R2-2 are politeness-tier residue inside its own stated degradation contract.
- No new rendered UI: confirmed, no design round owed.

**Verdict: Approve — clean. This round is TERMINAL: no BLOCKER or MAJOR remains; R2-1 through R2-5 are follow-ups, not a reason for another round.**

damianvtran:
- R2-1 (minor) — Deferred: the stale-steal `unlink` window is real — two stealers can race read-expiry → unlink and both end up fetching, and `release()`'s read-check-unlink has the same shape after a >60 s overrun. As the review notes, the consequence is bounded inside the lease's stated degradation contract (duplicate polite fetches, never a block, `mkstemp` keeps the document uncorrupted), so this is politeness-tier. Deferring the re-stat-before-unlink narrowing plus the docstring correction ("unlink+O_EXCL best-effort, not exactly-once") to a follow-up rather than invalidating a terminal round with a new head.
- R2-2 (minor) — Deferred (docs-only): correct that the exactly-one-fetch property holds only when the winner completes within the losers' 2 s wait; a slower endpoint degrades to leaseless fetching by design (the no-block constraint outranks herd protection). The bench/README claim will be softened to "one fetch when the winner completes within the loser wait" in the same follow-up as R2-1. The measured result and the pinning test are accurate for the regime they state.
- R2-3 (nit) — Deferred: the `_set_mcp_startup_sink` docstring names `dispose` as the second writer, but the actual mid-wiring guard is the wiring task's cancellation hook; the docstring will be corrected (or a clear added in `dispose`) in the follow-up. No behavioural gap — the sink cannot fire into a torn-down app today because the deferred wiring task is cancelled through the dispose hook before teardown proceeds.
- R2-4 (nit) — Rejected (as a change now): the double-`_rewire` window exists but is sub-millisecond, its worst case is duplicate repaint/measure scheduling (both idempotent), and a `_rewire`-ran-once flag adds state for a race nobody can observe. Will fold the flag in only if the follow-up for R2-1/R2-3 touches this file anyway.
- R2-5 (nit) — Deferred: agreed on both — `count("message_start") == 1` is the honest assertion (the `<= 1` form also passes on over-dropping), and the toast check's `or "MCP" in shown` arm accepts any MCP toast. Both tightenings go in the follow-up; the tests as merged still pin the fixed behaviours against the real app and real factory.

No push for this round per the terminal-approval rule: every finding above is follow-up-grade and a new head would invalidate the clean round. A tracking follow-up (R2-1/R2-2 doc+narrowing, R2-3 docstring, R2-5 test tightening) will ride the next PR that touches these files.

damianvtran:
### Agent review — round 3
- Reviewer: reviewer subagent (zai/glm-5.3)
- Scope: f2a52b5312a06c91d3bfeb284680636f75ef8593..4c7d6d6b9fe4346cc91b5b6f2ba2a37575c21d6c
- Delta reviewed: 1530892ffdfab97979b2d46bcf316164e98914e1..4c7d6d6b9fe4346cc91b5b6f2ba2a37575c21d6c
- Verdict: **Approve**

Delta is two CI-only commits in `tests/unit/test_tui_responsiveness.py`. Nothing round 2 approved was re-opened; both commits verified against the diff, the code under test, and the bench artifacts.

**4dc4e0ee (pid fix) — verified sound.** The diagnosis matches the code: the cancellation path really is `os.killpg(os.getpgid(process.pid), SIGKILL)` (`local_operator/tools/builtin.py:1306`), so a fake `pid = 1` aimed a real group-kill at init's process group on Linux. `subprocess.Popen(["sleep", 300], start_new_session=True)` gives the kill a real, isolated target with the exact semantics production expects — more faithful than the sentinel, not less. The bounded-emit assertions are untouched (`large <= small * 1.5`, `large <= 200_000` byte-identical to round 2), the fake reader/wait still simulate the stream, and the `try/finally` kill+wait reaps the sleep even when `time_one_emit` raises, so no zombie or stray 300 s sleeper leaks. Test passes locally (5.9 s) and in both CI runs on the head. The other `pid = 1` uses in the suite (`test_exec_mode.py:684`, mobile tests) never reach `killpg` — the pid is only written into records — so no sibling hazard was missed.

**4c7d6d6b (ceiling 250 → 500 ms) — guard still proves the regression gone.** Against the recorded baselines the new ceiling keeps real separation from every regression it guards: S2 pricing 807 ms loop / 812 ms max stall (bench/before.json — 1.6× above the ceiling), the 2 s cold boot scan (4×), the 10 s unlisted-model block. The one figure that now slips under is the 460–490 ms warm boot stall, but that scenario is pinned structurally, not temporally: the `_prepare` test asserts the scan ran on a worker thread (`seen_threads` non-empty, loop thread identity excluded), which catches a revert of the to_thread fix deterministically regardless of ceiling. The docstring quantifies its rationale (549 ms layout pass observed on a 4-vCPU runner with three siblings) rather than hand-waving. Acceptable trade of sensitivity for CI stability.

| ID | Severity | Evidence | Action |
| --- | --- | --- | --- |
| R3-1 | MINOR | `assert_no_stall_loaded` ceiling is 500 ms but the docstring's own worst-case contention observation is 549 ms (`test_tui_responsiveness.py:83-101`) — the flake window is narrowed, not closed; a recurrence on the e2e tests (the only ones running real Textual layout) still fails | If it recurs, raise the e2e ceiling toward 600–750 ms (still below the 807 ms guarded regression) or scope the loaded ceiling per test context; no action now |
| R3-2 | NIT | The edited docstring still says the 50 ms design bar "is asserted by the end-to-end boot test", but the e2e boot test asserts the loaded ceiling (line ~872) and no test asserts 50 ms — the strict-bar evidence lives in bench/before.json vs after.json only | Fold the wording fix into the R2-5 test-tightening follow-up |

No BLOCKER or MAJOR on the delta. **This round is TERMINAL**: R3-1/R3-2 are follow-ups alongside the R2-1/R2-2/R2-3/R2-5 batch, not a reason for another round. CI green on both runs of the head (test 3.12/3.13, lint, type-check, pip-audit, cli/server sanity).

damianvtran:
- R3-1 (minor) — Deferred: the 500 ms e2e ceiling vs 549 ms worst observed CI-contention stall is a flake-window concern, not a coverage gap — the guarded regressions all sit at 807 ms-10 s, and the warm boot stall is pinned structurally by the `_prepare` worker-thread-identity assertion. Folds into the R2 follow-up batch (tracked with R2-1/R2-2 for the next non-gating push).
- R3-2 (nit) — Deferred: docstring overstates the asserted bar (50 ms prose vs 500 ms ceiling); corrected wording rides the same follow-up push. No behavioural claim affected.

Head at disposition: `4c7d6d6b9fe4346cc91b5b6f2ba2a37575c21d6c`. CI green on both push and PR runs at this SHA; rounds 1-3 all dispositioned, round 3 reviewed this head.

PR #302, opened 2026-08-25
fix(session): stop a mid-batch model switch from bricking the session

damianvtran:
## Summary

Switching models mid-tool-batch permanently bricked a session. Every later turn, including a bare "Continue", died on the same Anthropic 400, and the only recovery was to abandon the session.

- `journal_model_switch` and `journal_incident` now park their **live** context append when a turn owns the message list, draining the FIFO at the turn-safe boundaries. The transcript write is unchanged.
- `_pair_spliced_tool_results` repairs an already-spliced history at request assembly, in `_render_history`, so a session bricked in memory heals itself and both provider wires are covered by one application.
- Version bumps from `0.33.1` to `0.33.2` as a patch fix.

## Impact

Before this change, pressing `/model` while the agent was running tools spliced the switch notice between the assistant's `tool_use` blocks and the `tool_result` blocks answering them. Nothing repaired the live list, so the illegal prefix was re-sent on every subsequent request and the session was unusable until restarted. Observed live in session `5187e748833c`.

After this change, the notice waits for the tool batch to close and lands immediately after it. A session that was already corrupted in memory, or corrupted by a future writer that bypasses the guard, still assembles a legal request.

## The bug

`Session.set_model` fires `_spawn_background(journal_model_switch(...))`, which appended a `CustomMessage` straight to `_context.messages` with no turn-boundary guard. `AgentLoop._run` appends the assistant message the moment the model turn ends and appends the tool results only once the whole batch returns (`_append_results`), so for the entire duration of a tool batch the live list ends in an assistant whose `tool_calls` have no answers. `_default_convert_to_llm` renders the notice as `role="user"`, splicing it into the open batch:

```text
assistant(tool_use, tool_use) -> user(notice) -> tool_result, tool_result
```

The invariant, in one sentence:

> `_context.messages` must never contain a non-tool message between an assistant message carrying `tool_calls` and the complete set of `tool_result` messages answering them, and no request may be assembled from a list that violates it.

Two things about this bug are easy to get wrong and are recorded in the code:

**The persisted transcript proves nothing.** Row order is WRITE order, not live-append order. `journal_model_switch` awaits its transcript write *before* the live append while the assistant is written later at the batch boundary, so the transcript of a bricked session looks perfectly clean. `--resume` was never broken; the live list was.

**The wire constraint is positional, not set membership.** `AnthropicClient._build_body` coalesces tool results onto a preceding user message, so the splice does not survive as its own turn: it becomes a text block sitting *ahead* of the `tool_result` blocks inside one user message. Measured against the live API:

| shape | result |
|---|---|
| `[tool_result, tool_result]` | 200 |
| `[tool_result, tool_result, text]` | 200 |
| `[text, tool_result, tool_result]` | **400** |
| interloper as its own user message between | **400** |

That is why the repair **moves** the interloper after the results rather than synthesizing placeholders. Placeholders were tested against the live API and produce a *different* 400 (`unexpected tool_use_id found in tool_result blocks`) because the genuine results still arrive behind them. Synthesis is only correct for a genuinely unanswered tail, which is `_wire_legal_snapshot`'s job and is left alone.

## Design

**Layer 1 — park until a safe boundary.** A new `_pending_context_journal` FIFO holds notices whose live append arrived mid-turn. The guard tests `_is_streaming or _turn_lock.locked()`, both conditions, for the reason `prompt` spells out: `_is_streaming` covers only `_run_turn`, while the lock spans the whole pipeline including post-compaction auto-continuation. `record_shell` and `adopt_aside` guard on the same pair.

The guard sits **inside** the journal methods, not at their call sites, so the direct-await failover path (`_on_route_settled`) is covered for free. That path runs on the loop inside a turn and splices identically, and it is the most likely way to hit this window in production because failover fires exactly when a request is in flight.

Drained at four boundaries: `_drain_steering`, `_run_turn_pipeline` exit, the top of `prompt`/`_prompt_messages`, and `dispose`. A parked notice is never silently lost, and the transcript write already happened so `--resume` is safe regardless.

**Layer 2 — repair at request assembly.** Applied in `_render_history` immediately after `_convert_to_llm`, which is the single site covering the turn path, compaction, `_history_snapshot` and the token counter. It fixes both wires at once: the OpenAI Responses path emits a bare `role:user` item between `function_call` and `function_call_output` and is rejected for the same reason, so patching `AnthropicClient` alone would have left it broken. Logs at `warning` when it fires, because after Layer 1 it should be dead code and silence would hide a regression. O(n) with an early exit when no assistant carries tool calls, since this is on the hot path.

## Real-path validation

**The verbatim production 400, reproduced and then cleared against the live Anthropic API.** A probe builds the exact bricked context (the incident's own tool-use ids), renders it through the session's own `_render_history`, hands it to the real `AnthropicClient._build_body`, and POSTs it to `api.anthropic.com/v1/messages` using the session's stored OAuth credential. BEFORE renders with Layer 2 disabled, which is what `origin/main` produces today.

```text
BEFORE (Layer 2 disabled) wire shape:
  ("user",      ["text"])
  ("assistant", ["text", "tool_use", "tool_use"])
  ("user",      ["text", "tool_result", "tool_result"])   <- notice ahead of the results
  ("user",      ["text"])

BEFORE -> HTTP 400
{"type":"error","error":{"type":"invalid_request_error","message":"messages.2: `tool_use` ids
were found without `tool_result` blocks immediately after: toolu_019T5KXtGAzsqT7jbDbRyecz,
toolu_01KW58PLNHHQv75fBBvqz53h. Each `tool_use` block must have a corresponding `tool_result`
block in the next message."},"request_id":"req_011CePxU2rGWSp23L8Dxoeik"}
```

That is the production error string, naming both of the original incident's tool ids. With the fix as shipped:

```text
AFTER (as shipped) wire shape:
  ("user",      ["text"])
  ("assistant", ["text", "tool_use", "tool_use"])
  ("user",      ["tool_result", "tool_result"])           <- results lead the message
  ("user",      ["text"])                                 <- notice moved after
  ("user",      ["text"])

AFTER -> HTTP 200
{"model":"claude-sonnet-4-5-20250929","id":"msg_011CePxU4sJtWGiBUP3MS3H7","type":"message",
"role":"assistant","content":[{"type":"text","text":"I need to read the files in the current
directory. Let me first see what files are available."}],"stop_reason":"end_turn",
"usage":{"input_tokens":168,...}}
```

**The repro failing on `origin/main` and passing after.** The regression tests were first written against the unmodified base (worktree at `f2a52b53`), where all three original cases fail:

```text
$ .venv/bin/python -m pytest tests/unit/session/test_background_journal_splice.py -q   # on origin/main
FAILED ...::test_a_mid_batch_model_switch_splices_a_user_message_into_the_tool_batch
FAILED ...::test_the_corrupt_prefix_is_re_sent_on_every_later_turn
FAILED ...::test_journal_incident_splices_the_same_way
3 failed in 2.17s
```

Each failed with `assert ['toolu_A', 'toolu_B'] == []` — both calls orphaned on the wire. On this branch, with the suite extended to 15 cases:

```text
    tests/unit/session/test_background_journal_splice.py -q
15 passed in 4.74s
```

The determinism is not a sleep: the fake tool calls `set_model` and then drains the session's own `_background_tasks` to completion before returning its result, so the interleaving is forced rather than raced.

## Test coverage

The 15 cases assert precise shapes, not the absence of an exception.

- **Layer 1, the three doors.** Mid-batch `set_model`; the same through `_on_mcp_incident`; the same through the direct-await `_on_route_settled` failover path. Each asserts the live context shape *and* the absence of dangling `tool_use` on a body built by the real Anthropic client from the request the loop actually issued.
- **A later turn rides a clean prefix** — the "Continue" that used to 400.
- **A parked notice is never lost.** Parked mid-turn then aborted, asserted delivered into the *request* the next `prompt` builds, not merely present in the list afterwards. Parked then disposed, asserted to have reached the transcript.
- **`_pair_spliced_tool_results` as a pure unit.** Interloper moved after the results (asserted by object identity, since compaction memoizes token estimates per message object); multiple interlopers keep relative order; a legal list returned unchanged; an unanswered tail left alone; the no-tool-calls early exit; and the warning log fires.
- **Self-heal.** A session handed a pre-spliced `_context.messages` runs a turn and the captured body is legal.
- **Both wires.** Spliced input asserted clean on the Anthropic body *and* on `_messages_to_openai_responses` items.
- **Resume stays clean**, replayed in write order.

## Automated validation

Full suite, deterministically ordered:

```text
$ env -u NO_COLOR TERM=xterm-256color .venv/bin/python -m pytest tests/unit -q -p no:randomly
6760 passed, 7 skipped, 431 warnings in 109.86s
```

Gates:

```text
$ .venv/bin/python -m flake8 .                                          # clean
$ .venv/bin/python -m black --check local_operator tests                # 492 files unchanged
$ .venv/bin/python -m isort --check-only --profile black local_operator tests   # clean
$ .venv/bin/python -m pyright --pythonpath .venv/bin/python .           # 0 errors, 0 warnings
```

`black` is the pinned 26.1.0 from the venv; `uvx` could not spawn in this environment, and the venv carries the identical pinned version (`python -m black, 26.1.0`).

## A note on two flaky tests

An earlier full-suite run under random ordering showed two failures that are **pre-existing flakes on unmodified `origin/main`**, not regressions from this change. Both were checked against a clean `f2a52b53` worktree:

- `tests/unit/tui/test_subagent_view.py::test_history_prepend_preserves_anchor_and_home_dedupes_requests` — fails 3 of 4 runs on clean `origin/main` in isolation.
- `tests/unit/harness/test_comms.py::test_a_resumed_child_replays_the_stopped_ones_transcript` — passes 3 of 3 in isolation on both trees; order-sensitive.

Both pass in the deterministic full run above. Left alone as out of scope for this fix, and flagged here rather than silently worked around.

## Scope

Deliberately unchanged: `_default_convert_to_llm`'s `role="user"` rendering (correct, the model must read these), `_paired_prefix`, `_wire_legal_snapshot`, `set_model`'s mid-turn switch semantics and its `_spawn_background` call, the loop's append ordering, the pending/aside/steering machinery, the transcript format, and `adopt_aside`'s refusal. `build_llm_history` needs no change: replaying the real bricked transcript already produces a legal body.

One design note worth flagging for review: `_drain_steering` appends parked notices to the context **directly** rather than returning them. Anything that method returns becomes a loop *injection*, and a non-empty injection list keeps the inner loop running, so returning an advisory notice bought an entire extra provider call. That was caught by two existing tests in `TestASwitchedModelTakesEffectAtTheNextCall` and is now recorded in a comment so it is not reintroduced.

The parked-notice delay is accepted rather than a compromise to be "fixed" later by re-splicing: the notice is advisory and the env block's `Model:` line already carries live identity, so the worst case is a switch notice landing at the end of a long tool batch instead of inside it. `_pair_spliced_tool_results`'s docstring records that it only ever moves harness-authored notices, and that the assumption needs revisiting if a custom type is ever added that carries user-authored text.

### DIFF
--- a/local_operator/session/session.py
+++ b/local_operator/session/session.py
@@ -497,6 +497,95 @@ _PERSISTABLE_CUSTOM_TYPES: frozenset[str] = frozenset(
 )
 
 
+def _pair_spliced_tool_results(messages: list[Message]) -> list[Message]:
+    """``messages`` with any message spliced INTO a tool batch moved after it.
+
+    The last line of defence for the invariant Layer 1 (``_append_or_park_journal``)
+    enforces at the source: the rendered history must never carry a non-tool
+    message between an assistant's ``tool_calls`` and the ``tool_result``s
+    answering them. A session already bricked IN MEMORY predates that guard and
+    cannot be helped by it, and a future writer appending straight to
+    ``_context.messages`` would reintroduce the same 400. This runs at request
+    assembly so neither case reaches a provider.
+
+    The constraint is POSITIONAL, not set membership. Verified against the live
+    Anthropic API (claude-sonnet-4-5, 2026-08-25), because
+    ``AnthropicClient._build_body`` coalesces tool results onto a preceding user
+    message and the difference decides what a repair has to do::
+
+        [tool_result, tool_result]            -> 200
+        [tool_result, tool_result, text]      -> 200
+        [text, tool_result, tool_result]      -> 400
+        interloper as its own message between -> 400
+
+    So results must be the LEADING run of the message that follows the calls,
+    and the repair is to MOVE the interloper after them. Synthesizing
+    placeholder results instead is wrong and was tested as such: it draws a
+    different 400 (``unexpected tool_use_id found in tool_result blocks``)
+    because the genuine results still arrive behind the placeholders.
+    Synthesis is only correct for a genuinely UNANSWERED tail, which is
+    :meth:`Session._history_snapshot`'s job — an unanswered batch is left
+    untouched here.
+
+    Relative order among several interlopers is preserved, and moving them is
+    safe ONLY because everything that can land in that window is a
+    harness-authored notice (``session_model_switch``, ``session_incident``):
+    reordering advisory chrome against a tool batch changes nothing the user
+    wrote. **If a custom type is ever added that carries user-authored text,
+    this assumption needs revisiting** — moving a user's words past a tool
+    batch would silently reorder the conversation they see.
+    """
+    # Hot path: ``_render_history`` runs on every provider call, and the
+    # overwhelming majority of histories have no tool batch to violate.
+    if not any(message.role == "assistant" and message.tool_calls for message in messages):
+        return messages
+
+    out: list[Message] = []
+    index = 0
+    total = len(messages)
+    repaired = False
+    while index < total:
+        message = messages[index]
+        out.append(message)
+        index += 1
+        if message.role != "assistant" or not message.tool_calls:
+            continue
+        # Collect the batch's answers and anything spliced among them, stopping
+        # at the first message that is neither — that is the end of the batch,
+        # and a trailing non-tool message there is already legal.
+        expected = {call.id for call in message.tool_calls}
+        results: list[Message] = []
+        interlopers: list[Message] = []
+        scan = index
+        while scan < total and expected:
+            candidate = messages[scan]
+            if candidate.role == "tool":
+                results.append(candidate)
+                expected.discard(candidate.tool_call_id or "")
+            else:
+                interlopers.append(candidate)
+            scan += 1
+        if expected or not interlopers:
+            # Unanswered tail (leave it to ``_wire_legal_snapshot``) or nothing
+            # spliced. Either way this batch is not ours to touch.
+            continue
+        out.extend(results)
+        out.extend(interlopers)
+        index = scan
+        repaired = True
+
+    if repaired:
+        # After Layer 1 this is dead code. Logging at warning rather than debug
+        # is deliberate: silence here would hide a regression that reintroduces
+        # the splice, and the repair would mask it right up until some future
+        # shape it cannot fix.
+        logger.warning(
+            "repaired a message spliced into an open tool batch; a journal "
+            "append bypassed the turn-boundary guard (see _append_or_park_journal)"
+        )
+    return out
+
+
 def _paired_prefix(messages: Sequence[AgentMessage]) -> list[AgentMessage]:
     """``messages`` truncated so it never ENDS in unanswered tool calls.
 
@@ -1226,6 +1315,16 @@ class Session:
         # Error text from the run just ended, journalled as a session incident
         # once persistence finishes (see _run_turn / journal_incident).
         self._pending_incident: str | None = None
+        #: Journal notices (model switch / session incident) whose LIVE append
+        #: arrived while a turn owned the message list. They are already on
+        #: disk; only the live splice is parked, because appending one between
+        #: an assistant's ``tool_calls`` and their ``tool_result``s builds a
+        #: list no provider accepts and NOTHING repairs it afterwards, so the
+        #: session re-sends the illegal prefix on every later turn and is
+        #: bricked for good. Drained at the turn-safe boundaries (steering
+        #: drain, pipeline exit, prompt entry, dispose) in the manner of
+        #: ``_pending_shell_records``.
+        self._pending_context_journal: list[CustomMessage] = []
         # (new_label, transient) of the last model switch made model-visible, so
         # the two edges that can both fire for one change (``set_model`` and a
         # route-settled event) do not double-announce. See journal_model_switch.
@@ -1369,6 +1468,14 @@ class Session:
         that reaches a provider has to be free of them.
         """
         rendered = self._convert_to_llm(self._live_todo_reminders(messages))
+        # Immediately after the conversion and before any other pass, so EVERY
+        # caller of this method is covered by one application: the turn path,
+        # compaction, `_history_snapshot` and the token counter. Repairing at
+        # the render rather than in a client fixes both wires at once — the
+        # OpenAI Responses path emits a bare `role:user` item between
+        # `function_call` and `function_call_output` and is rejected for the
+        # same reason Anthropic rejects the coalesced form.
+        rendered = _pair_spliced_tool_results(rendered)
         if self._images_rejected:
             # Nothing to rebound once images are being dropped outright, and
             # dropping first saves decoding a block that is about to become a
@@ -2574,6 +2681,10 @@ class Session:
             # receipt may have queued while the previous owner still held the
             # lock. It must be visible before this prompt builds its request.
             await self._flush_shell_records()
+            # Same race, same window: a journal notice parked by the previous
+            # owner must reach the model before this prompt's request is built,
+            # or the switch it announces goes unmentioned for another turn.
+            self._flush_context_journal()
             # A fresh user prompt supersedes any earlier interrupt request.
             self._abort_requested = False
             if self._is_streaming:
@@ -3177,6 +3288,10 @@ class Session:
             # Same shell-receipt boundary as prompt(): a wake turn must not
             # build from context that omits a command already visible in TUI.
             await self._flush_shell_records()
+            # Same boundary, same reason: a notice parked by a turn that ended
+            # without reaching a steering drain (an abort mid-batch) rejoins
+            # here, before this turn builds its request.
+            self._flush_context_journal()
             # Same flush as prompt(): a wake-delivery turn is still a prompt
             # path, and the catch-up must not be stranded behind it.
             self._handle_missed_wakes()
@@ -3227,6 +3342,11 @@ class Session:
             # provider message list. Persist it before releasing `_turn_lock`,
             # when no next prompt can build from a half-updated conversation.
             await self._flush_shell_records()
+            # Likewise for a journal notice parked mid-turn. The tool batch is
+            # closed by now, so the append is legal, and doing it here means an
+            # aborted turn still hands the notice to the next one instead of
+            # stranding it until the session is disposed.
+            self._flush_context_journal()
 
     async def _flush_held_end(self) -> None:
         """Emit the boundary event the pipeline was holding, if any."""
@@ -3625,6 +3745,61 @@ class Session:
             )
             return None
 
+    def _append_or_park_journal(self, message: CustomMessage) -> None:
+        """Put a journal notice on the live context, or park it for a boundary.
+
+        The live context may not take a non-tool message while a turn owns it.
+        ``AgentLoop`` appends the assistant message the moment the model turn
+        ends and appends the tool results only once the WHOLE batch returns
+        (``_append_results``), so for the entire duration of a tool batch the
+        list ends in an assistant whose ``tool_calls`` have no answers.
+        Splicing a notice in there produces
+        ``assistant(tool_use) -> user -> tool_result``, which every provider
+        rejects ("`tool_use` ids were found without `tool_result` blocks
+        immediately after") — and because nothing repairs the live list, the
+        session re-sends that prefix on every later turn and is bricked until
+        it is restarted. Observed live: a ``/model`` press mid-batch killed a
+        session outright, including a bare "Continue".
+
+        BOTH conditions are tested, for the reason :meth:`prompt` spells out:
+        ``_is_streaming`` covers only ``_run_turn``, while ``_turn_lock`` spans
+        the whole pipeline including a post-compaction auto-continuation.
+        ``record_shell`` and ``adopt_aside`` guard on the same pair.
+
+        The delay is accepted, not a compromise to be "fixed" later by
+        re-splicing: the notice is advisory, the env block's ``Model:`` line
+        already carries the live identity, and the transcript write has already
+        happened, so the worst case is that a switch NOTICE lands at the end of
+        a long tool batch instead of inside it. Re-splicing to make it prompt
+        is precisely the bug above.
+        """
+        if self._is_streaming or self._turn_lock.locked():
+            self._pending_context_journal.append(message)
+            return
+        self._context.messages.append(message)
+
+    def _drain_context_journal(self) -> list[CustomMessage]:
+        """Pop every parked journal notice, oldest first.
+
+        Synchronous and total: the transcript write already happened at park
+        time, so there is nothing here that can fail and nothing to retry —
+        unlike ``_flush_shell_records``, which pops only after a successful
+        write. Callers either extend the live context with the result or hand
+        it to the loop ahead of steering messages.
+        """
+        parked = self._pending_context_journal
+        self._pending_context_journal = []
+        return parked
+
+    def _flush_context_journal(self) -> None:
+        """Fold parked journal notices into the live context at a safe
+        boundary. A parked notice must never be silently dropped: it is the
+        only thing telling the model it was switched or that the last run
+        failed."""
+        parked = self._drain_context_journal()
+        if parked:
+            self._context.messages.extend(parked)
+
     async def journal_incident(self, raw: str) -> None:
         """Persist and surface WHY the session last failed.
 
@@ -3648,7 +3823,7 @@ class Session:
         )
         try:
             await self._transcript.append_message(message)
-            self._context.messages.append(message)
+            self._append_or_park_journal(message)
         except OSError:
             logger.warning("could not journal session incident", exc_info=True)
 
@@ -3712,7 +3887,7 @@ class Session:
             # fallback is live-context-only (see _is_persistable_message).
             if _is_persistable_message(message):
                 await self._transcript.append_message(message)
-            self._context.messages.append(message)
+            self._append_or_park_journal(message)
         except OSError:
             logger.warning("could not journal model switch", exc_info=True)
 
@@ -3793,6 +3968,30 @@ class Session:
         overwhelmingly common case — this is called at EVERY tool and message
         boundary, and an event per boundary would be noise with no receiver.
         """
+        # Journal notices parked by ``_append_or_park_journal`` ride out ahead
+        # of the steering messages. This is the loop's own injection boundary —
+        # it runs only after ``_append_results`` has closed the tool batch — so
+        # it is the earliest point at which a notice can rejoin the context
+        # legally, and ordering it first keeps "you were switched" ahead of the
+        # steer the user typed after switching. Already persisted at park time,
+        # so these are NOT re-appended to the transcript here.
+        # Journal notices parked by ``_append_or_park_journal`` rejoin the live
+        # context HERE, appended directly rather than returned. This is the
+        # loop's own injection boundary — it runs only after ``_append_results``
+        # has closed the tool batch — so it is the earliest point at which a
+        # notice can rejoin legally, and it lands ahead of any steering message
+        # the caller is about to drain in (``_drain_pending`` appends those
+        # after), which keeps "you were switched" before the steer the user
+        # typed after switching.
+        #
+        # NOT returned, and that distinction is load-bearing: anything this
+        # method hands back becomes a loop INJECTION, and a non-empty injection
+        # list keeps the inner loop running (``while has_more_tool_calls or
+        # pending``). Returning an advisory notice would therefore buy an entire
+        # extra provider call to tell the model something the next turn would
+        # have carried for free. Already persisted at park time, so there is no
+        # transcript write here either.
+        self._flush_context_journal()
         messages: list[AgentMessage] = []
         while not self._steering_queue.empty():
             message = self._steering_queue.get_nowait()
@@ -5782,6 +5981,13 @@ class Session:
             # switch made moments before quitting is exactly the write that
             # must still land for the next `--resume` to honour it.
             await self._flush_selected_model()
+            # Fold any still-parked journal notice back into the live context
+            # before the durability flush reads it. The transcript write
+            # already happened at park time, so `--resume` is correct either
+            # way and `_persist_new_messages` dedups by id; this keeps the
+            # in-memory list consistent with what was persisted rather than
+            # leaving a notice stranded in the FIFO at teardown.
+            self._flush_context_journal()
             # HC-11: cancel tracked background tasks (wake deliveries, aside
             # persistence), then close the session task group.
             for task in list(self._background_tasks):
--- a/local_operator/session/session.py
+++ b/local_operator/session/session.py
@@ -541,7 +541,11 @@ def _pair_spliced_tool_results(messages: list[Message]) -> list[Message]:
     is examined by more than one of them. An earlier version let an UNANSWERED
     batch rescan to the end of the list, which made a history of open batches
     quadratic (3.7s at 4000 messages against 2.3ms answered); see
-    ``test_pair_spliced_tool_results_scans_each_message_once``.
+    ``test_pair_spliced_tool_results_scans_each_message_once`` for the speed
+    property and
+    ``test_pair_spliced_tool_results_does_not_pull_the_next_batch_into_an_open_one``
+    for the correctness property, which are pinned separately because the bound
+    carries both.
     """
     # Hot path: ``_render_history`` runs on every provider call, and the
     # overwhelming majority of histories have no tool batch to violate.
@@ -576,10 +580,19 @@ def _pair_spliced_tool_results(messages: list[Message]) -> list[Message]:
                 # rescanned to the end of the list for EVERY later assistant
                 # message -- 4000 messages took 3.7s instead of 2.3ms.
                 #
-                # It is also the more correct stop. Swallowing the next batch
-                # would move ITS results onto this batch's assistant and push
-                # that assistant behind its own answers, turning one illegal
-                # shape into a different one.
+                # It is also the more correct stop, in one specific shape: a
+                # batch whose answer arrives LATE, after the next batch has
+                # already opened. Only then does ``expected`` empty at all, so
+                # only then can the repair fire on a window that spans two
+                # batches -- ``A1(x) A2(y,z) T(y) T(z) T(x)`` would come out as
+                # ``A1 T(y) T(z) T(x) A2``, with A2 emitted BEHIND its own
+                # answers: one illegal shape traded for another.
+                #
+                # A batch that is never answered at all is NOT this case, and
+                # was never at risk: ``expected`` stays non-empty, so the
+                # ``if expected: continue`` guard below suppresses the repair
+                # with or without this bound. Do not reach for an unanswered
+                # batch to justify the stop -- reach for the late answer.
                 break
             if candidate.role == "tool":
                 results.append(candidate)
--- a/local_operator/session/session.py
+++ b/local_operator/session/session.py
@@ -524,7 +524,7 @@ def _pair_spliced_tool_results(messages: list[Message]) -> list[Message]:
     different 400 (``unexpected tool_use_id found in tool_result blocks``)
     because the genuine results still arrive behind the placeholders.
     Synthesis is only correct for a genuinely UNANSWERED tail, which is
-    :meth:`Session._history_snapshot`'s job — an unanswered batch is left
+    :meth:`Session._wire_legal_snapshot`'s job — an unanswered batch is left
     untouched here.
 
     Relative order among several interlopers is preserved, and moving them is
@@ -534,6 +534,14 @@ def _pair_spliced_tool_results(messages: list[Message]) -> list[Message]:
     wrote. **If a custom type is ever added that carries user-authored text,
     this assumption needs revisiting** — moving a user's words past a tool
     batch would silently reorder the conversation they see.
+
+    Linear in ``len(messages)``, and that has to stay true because this sits on
+    every provider call. Each batch's inner scan stops at the next assistant
+    carrying ``tool_calls``, so the scanned windows are disjoint and no message
+    is examined by more than one of them. An earlier version let an UNANSWERED
+    batch rescan to the end of the list, which made a history of open batches
+    quadratic (3.7s at 4000 messages against 2.3ms answered); see
+    ``test_pair_spliced_tool_results_scans_each_message_once``.
     """
     # Hot path: ``_render_history`` runs on every provider call, and the
     # overwhelming majority of histories have no tool batch to violate.
@@ -559,9 +567,28 @@ def _pair_spliced_tool_results(messages: list[Message]) -> list[Message]:
         scan = index
         while scan < total and expected:
             candidate = messages[scan]
+            if candidate.role == "assistant" and candidate.tool_calls:
+                # A new batch opens here, so this one is over whatever is still
+                # unanswered. Bounding the scan is what keeps the whole pass
+                # linear: every batch's window ends at the next batch's start,
+                # so the windows are disjoint and each message is examined by at
+                # most one of them. Without this bound an unanswered batch
+                # rescanned to the end of the list for EVERY later assistant
+                # message -- 4000 messages took 3.7s instead of 2.3ms.
+                #
+                # It is also the more correct stop. Swallowing the next batch
+                # would move ITS results onto this batch's assistant and push
+                # that assistant behind its own answers, turning one illegal
+                # shape into a different one.
+                break
             if candidate.role == "tool":
                 results.append(candidate)
-                expected.discard(candidate.tool_call_id or "")
+                # Guarded rather than ``or ""``: coercing a missing id to the
+                # empty string would let an unidentified result answer a call
+                # only if some call also had an empty id. A result with no id
+                # answers nothing, so it closes no call.
+                if candidate.tool_call_id is not None:
+                    expected.discard(candidate.tool_call_id)
             else:
                 interlopers.append(candidate)
             scan += 1
@@ -601,7 +628,7 @@ def _paired_prefix(messages: Sequence[AgentMessage]) -> list[AgentMessage]:
     tool_call_id"). Measured: a Ctrl+C mid-batch left two unpaired calls on
     disk and an unusable session.
 
-    :meth:`Session._history_snapshot` faces the same illegal tail for a
+    :meth:`Session._wire_legal_snapshot` faces the same illegal tail for a
     request it is about to send and pairs the calls with placeholders. This is
     the persistence counterpart, and it DROPS rather than pairs: a placeholder
     is the honest answer to "what are you doing right now", but on disk it
@@ -1470,7 +1497,7 @@ class Session:
         rendered = self._convert_to_llm(self._live_todo_reminders(messages))
         # Immediately after the conversion and before any other pass, so EVERY
         # caller of this method is covered by one application: the turn path,
-        # compaction, `_history_snapshot` and the token counter. Repairing at
+        # compaction, `_wire_legal_snapshot` and the token counter. Repairing at
         # the render rather than in a client fixes both wires at once — the
         # OpenAI Responses path emits a bare `role:user` item between
         # `function_call` and `function_call_output` and is rejected for the
@@ -3784,18 +3811,19 @@ class Session:
         Synchronous and total: the transcript write already happened at park
         time, so there is nothing here that can fail and nothing to retry —
         unlike ``_flush_shell_records``, which pops only after a successful
-        write. Callers either extend the live context with the result or hand
-        it to the loop ahead of steering messages.
+        write. Split out from its one caller, ``_flush_context_journal``, so
+        the drain is testable on its own.
         """
         parked = self._pending_context_journal
         self._pending_context_journal = []
         return parked
 
     def _flush_context_journal(self) -> None:
-        """Fold parked journal notices into the live context at a safe
-        boundary. A parked notice must never be silently dropped: it is the
-        only thing telling the model it was switched or that the last run
-        failed."""
+        """Fold parked journal notices into the live context at a safe boundary.
+
+        A parked notice must never be silently dropped: it is the only thing
+        telling the model it was switched or that the last run failed.
+        """
         parked = self._drain_context_journal()
         if parked:
             self._context.messages.extend(parked)
@@ -3968,13 +3996,6 @@ class Session:
         overwhelmingly common case — this is called at EVERY tool and message
         boundary, and an event per boundary would be noise with no receiver.
         """
-        # Journal notices parked by ``_append_or_park_journal`` ride out ahead
-        # of the steering messages. This is the loop's own injection boundary —
-        # it runs only after ``_append_results`` has closed the tool batch — so
-        # it is the earliest point at which a notice can rejoin the context
-        # legally, and ordering it first keeps "you were switched" ahead of the
-        # steer the user typed after switching. Already persisted at park time,
-        # so these are NOT re-appended to the transcript here.
         # Journal notices parked by ``_append_or_park_journal`` rejoin the live
         # context HERE, appended directly rather than returned. This is the
         # loop's own injection boundary — it runs only after ``_append_results``
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "local-operator"
-version = "0.33.1"
+version = "0.33.2"
 description = "A team of proactive AI assistants that can work together behind the scenes to help you get mundane tasks done so you can focus on the fun stuff."
 readme = "README.md"
 authors = [{ name = "Damian Tran", email = "damian@gominerva.com" }]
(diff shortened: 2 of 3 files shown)


damianvtran:
### CI note: one flaky mobile test, not a regression

The checks list shows a red `test (3.13)`. That entry belongs to the **stale duplicate run** for this same commit (two workflow runs were queued for the push and the PR event). The current run is fully green:

- run `32886347256` — `lint`, `type-check`, `pip-audit`, `test (3.12)`, `test (3.13)`, `cli-sanity`, `server-sanity` all **pass**.

The failure was `tests/unit/mobile/test_tui_ask.py::test_secret_ask_projects_a_free_text_card_without_leaking_the_value` — `AssertionError: no discovery record published`.

Evidence it is environmental and unrelated to this change:

- **Re-running the identical commit passed** (`test (3.13)` 11m10s, run 32886347256).
- It passes locally 3/3 (`.venv/bin/python -m pytest tests/unit/mobile/test_tui_ask.py -q` → `7 passed`).
- The assertion comes from `registry.scan()`, which reads the shared mobile run directory (`local_operator/mobile/registry.py`). That is a process-global resource with no per-test isolation, so a concurrently running test session can reap or race the record. Nothing in this PR touches `local_operator/mobile/` or the registry.

Two further pre-existing flakes were observed locally under random ordering and are documented in the PR description; both reproduce on a clean `origin/main` worktree at `f2a52b53`, and the full suite is green in fixed order (`6760 passed, 7 skipped`).

None of the three are fixed here — they are out of scope for this fix and are flagged rather than silently worked around.

damianvtran:
### Agent review — round 1

Scope: `f2a52b5312a06c91d3bfeb284680636f75ef8593..77b2503e3fba5630850482cc5d8869f63a5cc9d5`

Reviewed in an independent detached worktree off the PR head, with a second worktree at the base commit for before/after comparison. I did not touch the author's tree.

**No blocker and no major findings.** The diagnosis, the two-layer design, and the evidence all hold up under adversarial probing. Details below, then five minors and three nits.

---

#### What I verified independently (not just re-read)

**Gates, re-run on the PR head:**

```text
$ .venv/bin/python -m flake8 .                                    # clean
$ .venv/bin/python -m black --version                             # 26.1.0 (pinned)
$ .venv/bin/python -m black --check local_operator tests          # 492 files unchanged
$ .venv/bin/python -m isort --check-only --profile black ...      # clean
$ .venv/bin/python -m pyright --pythonpath .venv/bin/python .     # 0 errors, 0 warnings
$ env -u NO_COLOR TERM=xterm-256color .venv/bin/python -m pytest tests/unit -q -p no:randomly
6760 passed, 7 skipped, 431 warnings in 175.44s
```

That reproduces the PR's claimed `6760 passed, 7 skipped` exactly.

**The repro genuinely fails on base.** I copied the new test file onto a clean `f2a52b53` worktree and stubbed `_pair_spliced_tool_results` to identity (the symbol does not exist on base) so the Layer 1 cases could run:

```text
$ .venv/bin/python -m pytest tests/unit/session/test_background_journal_splice.py -q   # on f2a52b53
11 failed, 4 passed
```

failing with `assert ['toolu_A', 'toolu_B'] == []` and the live wire shape `('user', ['text', 'tool_result', 'tool_result'])` — the production symptom, naming the incident's own ids. All 15 pass on the head. The before/after claim is real.

**Guard completeness (probe 1).** `_append_or_park_journal` (session.py:3776) tests `self._is_streaming or self._turn_lock.locked()` — both conditions, as specified. The guard is inside the `journal_*` functions (session.py:3826, 3890), so all five invocation sites are covered, including the two direct-`await` failover sites in `_on_route_settled` (session.py:3203, 3234) and the `_on_mcp_incident` background spawn (3897). No call site bypasses it.

**Re-entrancy (probe 3) — clean, and for a checkable reason.** `_append_or_park_journal`, `_drain_context_journal` and `_flush_context_journal` are all synchronous (`inspect.iscoroutinefunction` → `False, False, False`) and contain no `await`. The drain swaps the list and reassigns in one uninterruptible stretch, so on a single-threaded loop a park cannot interleave with a drain, double-deliver, or reorder. Nothing mutates `_context.messages` off-loop. Ordering is FIFO by construction.

**`_drain_steering`'s direct append is correct, and the design note is right.** `self._context` is handed to the loop as the same object (session.py:3457), so the direct `extend` lands in the list the loop owns. Both loop injection-collection points (`_collect_inflight_injections` at loop.py:620, `_collect_yield_injections` at 629) run strictly *after* `_append_results` (loop.py:566), so the "batch is closed at every drain" claim holds. Returning the notice instead would extend `while has_more_tool_calls or pending` and buy a real extra provider call — the comment is load-bearing, not decorative.

**Layer 2 correctness (probe 4/5) — I tried 13 adversarial shapes and could not break it:**

| input | output | correct |
|---|---|---|
| `[]`, `[user]` | unchanged | ✅ |
| `A(a,b) T(a) T(b) U` | unchanged | ✅ |
| `A(a,b) U T(a) T(b)` | `A(a,b) T(a) T(b) U` | ✅ |
| `A(a,b) T(a) U T(b)` (interleaved) | `A(a,b) T(a) T(b) U` | ✅ |
| `A(a,b) U:n1 T(a) U:n2 T(b)` | interlopers keep relative order | ✅ |
| two consecutive spliced batches | both repaired independently | ✅ |
| `A(a,b) T(a)` (unanswered tail) | untouched → left to `_wire_legal_snapshot` | ✅ |
| `A(a,b) U T(a)` (partial + interloper) | untouched | ✅ |
| assistant-role interloper | moved after results | ✅ |
| foreign `tool_call_id` in the batch | handled, no crash | ✅ |

The positional rule is genuinely encoded, not faked. `_dangling_tool_use` (test file) breaks at the first non-`tool_result` block rather than testing set membership, which is exactly the `[text, tool_result, ...]` → 400 vs `[tool_result, ..., text]` → 200 distinction. The repair MOVES rather than synthesizes, correctly, and the reasoning about the second 400 (`unexpected tool_use_id`) is documented at session.py:519-527.

**Both wires (probe 6).** Applying at `_render_history` (session.py:1478) rather than in `AnthropicClient` is the right call, and `test_both_wires_accept_the_repaired_history` asserts on the real `AnthropicClient._build_body` output *and* on `_messages_to_openai_responses` items, with a separate positional checker per wire. One test, both wires, genuinely.

**Scope (probe 7) — exemplary.** The entire diff removes exactly two lines:

```text
$ git diff f2a52b53..77b2503e -- local_operator/session/session.py | grep '^-' | grep -v '^---'
-            self._context.messages.append(message)
-            self._context.messages.append(message)
```

Both are the bug. `_default_convert_to_llm`, `_paired_prefix`, `_wire_legal_snapshot`, `set_model`, loop append ordering, and the transcript format are untouched. No drive-by changes.

**CI's red `test (3.13)` is not this PR.** It fails in `tests/unit/mobile/test_tui_ask.py` (`no discovery record published`). I ran that file on both trees:

```text
head (77b2503e): 7 passed
base (f2a52b53): 1 failed, 6 passed   <- test_terminal_advance_reprojects_and_stale_phone_tap_is_safe
```

It fails on the **base** commit and passes on the head, so it is a pre-existing order/timing flake, not a regression. Same for `test_comms.py::test_a_resumed_child_replays_the_stopped_ones_transcript`, which I reproduced failing on base in the same suite context. The PR's flake note is accurate. **CI should still be green before merge** — see minor 1.

---

#### Findings

**MINOR 1 — the red `test (3.13)` check should be re-run to green before merge, not merged past.**
Both `Pre-release CI Testing` runs show `test (3.13)` FAILURE. I have satisfied myself it is pre-existing (evidence above), but merging with a red required check on the record makes the next person's bisect harder and normalizes overriding the gate. Re-run the job; if it stays red on a re-run, file the flake as a ticket and link it here so the override is documented rather than assumed.

**MINOR 2 — `_pair_spliced_tool_results` is quadratic on a history with unanswered mid-list batches** (`session.py:559-574`).
When a batch is unanswered the `if expected: continue` branch (line 568) does **not** advance `index`, so the inner `while` rescans to `total` for every subsequent assistant message. Measured on the head:

```text
answered   n=4000 messages:    0.77 ms
unanswered n=4000 messages:  812.95 ms      # ~1000x
```

I could not construct a reachable path to this shape — `cutpoint.py:250` asserts a cut point never orphans a tool call, and `prune_tool_outputs` blanks text but explicitly never removes messages, so pairing survives compaction. That is why this is minor and not major. But the docstring claims "O(n) with an early exit" (session.py:538) and on this input it is not. Fix, one line: bound the inner scan, e.g. `scan` stops at the next assistant-with-tool_calls, or set `index = scan` on the unanswered branch too if a rescan can never find anything. At minimum, correct the complexity claim so a future reader does not rely on it.

**MINOR 3 — duplicated comment block in `_drain_steering`** (`session.py:3971-3985`).
Two near-identical paragraphs describing the same thing are stacked, the first ending "so these are NOT re-appended to the transcript here." and the second restarting "Journal notices parked by `_append_or_park_journal` rejoin the live context HERE...". Clearly an editing artifact. Delete the first paragraph (lines 3971-3977); the second is the better-written one and carries the load-bearing "NOT returned" reasoning.

**MINOR 4 — `:meth:`Session._history_snapshot`` does not exist** (`session.py:527`, `session.py:1473`).
There is no `_history_snapshot` method; the cross-reference is dead. It is a pre-existing typo (base carries it at line 515) that this PR propagates to two new sites. The intended target is `_wire_legal_snapshot` (session.py:5048) at line 527, and at 1473 the caller meant is the snapshot path generally. Worth fixing in the same breath since the new docstrings lean on that reference to explain the division of labour.

**MINOR 5 — the `repaired` warning fires once per render, not once per session.**
`_render_history` runs on every provider call, so a session bricked in memory logs the same warning on every turn for the rest of its life. That is arguably intended (the comment at session.py:578 argues silence would hide a regression), and it is genuinely rare, but a once-per-session latch would keep the signal while not burying the log. Your call — the current behaviour is defensible.

**NIT 1** — `expected.discard(candidate.tool_call_id or "")` (session.py:564): the `or ""` silently maps a `None` id onto the empty-string key. Harmless today since no `ToolCall.id` is empty, but `discard(candidate.tool_call_id)` guarded by a `None` check reads more honestly.

**NIT 2** — `_flush_context_journal`'s docstring (session.py:3794-3798) opens mid-sentence on the summary line ("Fold parked journal notices into the live context at a safe / boundary."), unlike the one-line-summary style the rest of the file uses. Cosmetic.

**NIT 3** — `_drain_context_journal` is only ever called by `_flush_context_journal`; the "Callers either extend the live context with the result or hand it to the loop ahead of steering messages" docstring (session.py:3787) describes a two-caller design that the final code does not have. Trim it to match reality so nobody preserves the split unnecessarily.

---

#### Evidence quality (probe 8)

Strong, and it matches what the diff does. The live-API before/after is the real thing: the verbatim production 400 naming both incident tool ids, cleared to a real HTTP 200 with a genuine `msg_...` id and token usage, with the wire shapes printed either side. The placeholder-synthesis alternative was tested against the live API and rejected for a stated reason rather than hand-waved. The determinism argument (draining `_background_tasks` rather than sleeping) is honest and I confirmed it in `_drain_background`. The flake note is accurate, and flagging it beat silently working around it.

#### Comments (probe 9)

House style is met. Both items the brief called out are documented: the accepted delay of a parked notice at session.py:3768-3773, and `_pair_spliced_tool_results`'s harness-authored-notices assumption with the explicit "**If a custom type is ever added that carries user-authored text, this assumption needs revisiting**" warning at session.py:530-536. That is the right thing to have written down.

---

### Verdict: `clean` — round 1 is TERMINAL.

No blocker, no major. The invariant is correctly stated, enforced at the source, and defended at request assembly; the design is right and the implementation matches it. The five minors and three nits above are follow-ups, not grounds for another round. Merge once `test (3.13)` is re-run green or its pre-existing flake is explicitly documented (minor 1); I would fix minors 2-4 either in this PR or a fast follow-up, since 2 and 4 are both one-liners in code this PR already touches.

damianvtran:

Implementer: `coder` subagent (lopdev team), model `anthropic/claude-opus-5`
Scope: `77b2503e3fba5630850482cc5d8869f63a5cc9d5..865c419db6ec5388945466e182bad64e908b269c`
Remediation commit: `865c419db6ec5388945466e182bad64e908b269c`

Thanks for the round — the quadratic measurement was the valuable catch, and chasing it turned up a correctness bug underneath the performance one. Every finding is answered below.

---

#### MINOR 1 — red `test (3.13)` should be re-run to green — **deferred to the merging agent**

Not mine to close from here: I cannot re-run a check on a commit that no longer exists, since this remediation pushed a new head (`865c419d`). CI has re-triggered from scratch on it. The merge gate stands as you wrote it: `test (3.13)` must be green on the current head, or its failure documented as the pre-existing flake with a linked ticket, before merge. Flagging for whoever merges rather than silently inheriting the override.

#### MINOR 2 — quadratic scan on unanswered mid-list batches — **fixed** (`865c419d`)

Reproduced first, on the old head, using two shapes at three sizes:

```text
answered   n= 1000 messages:      0.19 ms
answered   n= 2000 messages:      0.44 ms
answered   n= 4000 messages:      2.33 ms
unanswered n= 1000 messages:    283.71 ms
unanswered n= 2000 messages:    683.42 ms
unanswered n= 4000 messages:   3706.26 ms     # 4x per doubling
```

After the fix, same harness, same machine:

```text
answered   n= 1000 messages:      0.26 ms
answered   n= 2000 messages:      0.42 ms
answered   n= 4000 messages:      0.80 ms
unanswered n= 1000 messages:      0.26 ms
unanswered n= 2000 messages:      0.51 ms
unanswered n= 4000 messages:      0.99 ms     # 2x per doubling
```

**3706ms to 0.99ms**, and the growth curve is now linear in both shapes rather than quadratic in one.

**On the two candidate fixes you offered — the second one is wrong, and that is worth recording.** You suggested either bounding the inner scan at the next assistant-with-`tool_calls`, or `index = scan` on the unanswered branch "if a rescan can never find anything". I implemented the bound. I also tried `index = scan` and it fails **9 of the 15 existing tests** in this file:

```text
$ # with `index = scan` on the unanswered branch
9 failed, 8 passed
FAILED test_the_direct_await_failover_path_is_guarded_too - assert ['toolu_A', 'toolu_B'] == []
```

Because `scan` has already run to `total` at that point, so jumping `index` to it skips every remaining message including legal batches that needed repair. Good thing to have checked rather than taken on faith.

The bound is also a **correctness** fix, not only a speed one, which I did not expect going in. An unbounded scan hunting the first batch's missing answer walks into the *second* batch, collects that assistant as an interloper and its results as the first batch's, then re-emits that assistant behind its own answers — turning one illegal shape into a different illegal shape. `test_pair_spliced_tool_results_does_not_pull_the_next_batch_into_an_open_one` pins that.

**The docstring claim is now true and states why**: each batch's window ends at the next batch's start, so the windows are disjoint and no message is examined by more than one.

**Regression test** — you asked for linear runtime or a visit-count assertion. I used the count, deliberately: a wall-clock threshold is flaky on a loaded machine (this one demonstrated exactly that, below). `_CountingHistory` counts `__getitem__` calls, and asserts `reads <= 2n`. On the old code it fails loudly:

```text
AssertionError: the scan read 160800 times for 800 messages;
an unanswered batch is being rescanned instead of advanced past
assert 160800 <= (2 * 800)
```

Verified failing on the pre-fix function and passing after, so it genuinely guards the regression rather than merely passing today.

#### MINOR 3 — duplicated comment block in `_drain_steering` — **fixed** (`865c419d`)

Deleted the first paragraph as you recommended; the second is the better-written one. The load-bearing rationale survives intact: that notices are appended **directly rather than returned**, because anything returned becomes a loop injection and a non-empty injection list keeps the inner loop running (`while has_more_tool_calls or pending`), buying a whole extra provider call for an advisory notice.

#### MINOR 4 — dead `:meth:`Session._history_snapshot`` cross-reference — **fixed** (`865c419d`)

Confirmed dead (`grep -c "def _history_snapshot"` → `0`). Fixed at **three** sites, not two — there was also a plain-text `_history_snapshot` in the `_render_history` comment at line 1495 that the PR added. All now point at `_wire_legal_snapshot`, which I checked really is the placeholder-pairing method both references meant.

#### MINOR 5 — `repaired` warning fires once per render — **rejected**, current behaviour kept

You called it defensible and left it to me; I am keeping it. `_pair_spliced_tool_results` is a **module-level pure function**, so a once-per-session latch would need either module-global mutable state (breaking test isolation and lying across sessions in one process, e.g. subagents) or an extra parameter threaded from `Session` purely to suppress a log line. Both cost more clarity than the repetition costs. The repetition is also self-limiting: after Layer 1 this is dead code, and a session that does hit it is genuinely broken, which is the case where a repeated warning is a feature. Happy to revisit if it is ever observed in a real log.

#### NIT 1 — `expected.discard(candidate.tool_call_id or "")` — **fixed** (`865c419d`)

Now an explicit `if candidate.tool_call_id is not None:` guard, with a comment on why: a result carrying no id answers nothing, so it should close no call. The `or ""` form could let an unidentified result close a call only in the pathological case where some call also had an empty id.

#### NIT 2 — `_flush_context_journal` docstring opens mid-sentence — **fixed** (`865c419d`)

Reflowed to a one-line summary, blank line, then the body, matching the file's style.

#### NIT 3 — `_drain_context_journal` describes a two-caller design that does not exist — **fixed** (`865c419d`)

Replaced the "callers either... or..." sentence with the actual reason for the split: it has one caller, `_flush_context_journal`, and is separate so the drain is testable on its own.

#### The three flaky tests — **rejected as pre-existing**, follow-up note kept

Per instruction, not fixed here, and I re-confirmed independently rather than inheriting the claim. My full-suite runs on this head hit a *rotating* cast of TUI/timing failures — `test_esc_aborts_a_running_shell_command`, `test_an_adopted_row_times_the_tool_not_the_dictation`, `test_a_cancelled_child_keeps_the_work_it_had_already_done`, then `test_a_narrow_card_still_names_what_it_is_authorising` on the very next run. All pass in isolation (3/3). The tell is that they are wall-clock assertions (`assert 0.786 < 0.4`, "timed out waiting for condition") and the suite ran 259s against the PR's 109s, i.e. machine load.

Checked on a clean detached worktree at base `f2a52b53`:

```text
$ pytest tests/unit/tui/test_steering_approval.py tests/unit/tui/test_app_pilot.py tests/unit/harness/test_comms.py -q -p no:randomly
1 failed, 391 passed        # test_the_composing_row_sheds_its_label_before_its_facts
```

Fails on **base**, in the same files, in code this PR does not touch. Out of scope, flagged not worked around. Worth a ticket to de-time these assertions.

---

#### Gates re-run on `865c419d`

```text
$ .venv/bin/python -m flake8 .                                          # clean
$ .venv/bin/python -m black --version                                   # 26.1.0 (pinned)
$ .venv/bin/python -m black --check local_operator tests                # 492 files unchanged
$ .venv/bin/python -m isort --check-only --profile black local_operator tests   # clean
$ .venv/bin/python -m pyright --pythonpath .venv/bin/python .           # 0 errors, 0 warnings
$ env -u NO_COLOR TERM=xterm-256color .venv/bin/python -m pytest tests/unit -q -p no:randomly
6761 passed, 7 skipped   (+1 load flake, pre-existing, evidence above)
```

`6759 + 2` new tests = `6761`, against the round-1 baseline of `6760`. pyright initially caught an untyped `list` subclass in the new test helper; fixed before commit.

The touched suite is stable — `tests/unit/session/test_background_journal_splice.py` ran **8/8 green under random ordering** (17 passed each time), and `tests/unit/session` 4/4 green.

No blocker or major findings were open, so no further round is required on my side. Not merging — MINOR 1's CI condition is the merging agent's to confirm on this head.

damianvtran:
**MINOR 1 update — now closed, not deferred.** CI has finished on the remediation head `865c419d` and every check is green on **both** workflow runs, `test (3.13)` included:

```text
test (3.13)    pass  11m9s   run 32892719549
test (3.13)    pass  10m51s  run 32892724253
test (3.12)    pass  9m3s / 10m40s
lint, type-check, pip-audit, cli-sanity, server-sanity   pass
```

So there is no red required check on the record and no override to document — the flake did not reproduce on a fresh run, which is consistent with the pre-existing-flake diagnosis in round 1. Nothing further outstanding from my side.

damianvtran:
### Agent review — round 2

Scope: `77b2503e3fba5630850482cc5d8869f63a5cc9d5..865c419db6ec5388945466e182bad64e908b269c`

Delta review of the remediation commit only. The two-layer design cleared in round 1 is treated as settled; this round exists because `865c419d` changed behaviour rather than only comments. Reviewed in my own fresh detached worktree off the PR head, plus a second worktree at base `f2a52b53`. I did not touch the author's tree and did not modify the PR.

**No blocker and no major findings.** The bound is correct, the linearity claim is now true, and every accepted fix is semantics-preserving. One MINOR: the test named for the correctness half of the bound does not actually exercise it, and the correctness claim in the remediation comment is stated more broadly than the code supports.

---

#### What I verified independently

**The bound is correct in the general case.** I built my own adversarial harness with an independent legality checker (a batch is illegal when its answers are stranded behind a later message, mirroring the positional constraint round 1 established against the live API), and ran 18 shapes including every one the brief asked for:

```text
OK  empty / singleton user / singleton open assistant     unchanged
OK  legal answered batch                                  unchanged
OK  spliced simple, interleaved, foreign tool_call_id     repaired
OK  zero interlopers                                      unchanged
OK  consecutive spliced batches                           both repaired independently
OK  unanswered batch followed by answered-spliced         second repaired, first left alone
OK  answered-spliced followed by unanswered               first repaired, second left alone
OK  interloper immediately before the next assistant      correctly NOT swallowed
OK  three batches, middle spliced                         only the middle touched
OK  long unanswered tail (5 open batches)                 untouched
```

No shape lost, duplicated, or reordered a message, and no shape came out illegal that went in legal. The two that stay illegal (`A(x,y) U T(x)` and answers landing after the next batch opens) are genuinely open batches, which the docstring correctly hands to `_wire_legal_snapshot`.

**Claim 1 (the old scan traded one illegal shape for another) — TRUE, but narrower than described.** I reconstructed the round-1 function by stripping only the bound, and the shape in the remediation comment does **not** reproduce it:

```text
in       : ['first','fr','second','sa','sb']   # first batch's 1B never answered
unbounded: ['first','fr','second','sa','sb']   # identical — no swallow
```

Because when a batch is genuinely unanswered, `expected` never empties, so `if expected: continue` (session.py:596) suppresses the repair regardless of the bound. The bound is never what saves that shape. The bug is real but needs a *later answer*, so `expected` empties only after the next assistant:

```text
in       : ['b1','b2','ty','tx']       # b1's answer arrives after b2 opened
unbounded: ['b1','ty','tx','b2']       # b2 emitted BEHIND its own answers — illegal
bounded  : ['b1','b2','ty','tx']       # left alone — correct
```

So the fix is right and the reasoning behind it is right; only the reproducing shape was misidentified.

**Claim 2 (`index = scan` breaks 9 of 15) — TRUE.** Patched the alternative in and ran the file: `9 failed, 8 passed`, failing with `assert ['toolu_A','toolu_B'] == []`. The stated cause holds — `scan` has already run to `total` on the unanswered branch, so `index = scan` skips every remaining message including legal batches needing repair. The round-1 suggestion was not viable; rejecting it was correct.

**Claim 3 (linearity) — TRUE for every shape I could construct.** Counting `__getitem__`, the ratio is flat across a 4x size range in all five shapes, including the ones the brief named:

```text
many small unanswered batches   n=4000: 7999 reads (2.00n)
many small answered batches     n=3999: 6665 reads (1.67n)
ONE huge open batch + long tail n=4000: 7999 reads (2.00n)
long unanswered tail            n=4000: 7999 reads (2.00n)
spliced every batch             n=4000: 4000 reads (1.00n)
```

The `reads <= 2n` bound is the right property and the counting approach is sound — I removed only the bound and the test failed loudly and correctly (`160800 <= 1600`), so it genuinely pins the property rather than passing incidentally. A superlinear implementation cannot satisfy it, because any rescan re-reads indices and the counter is per-read, not per-index.

**The other accepted fixes — all correct, none altered semantics.**

- MINOR 4 at three sites: `_history_snapshot` → `_wire_legal_snapshot` at session.py:527, 631, 1500. Comment-only; the target exists.
- NIT 1's id coercion (session.py:585-590): **cannot mask a missing id in a way that changes wire output — it is strictly safer.** The only divergence from `or ""` needs a result with `tool_call_id=None` *and* a call whose id is `""`. Confirmed `ToolCall(id="")` is constructible (no validation) and `tool_call_id` is `Optional`, so the case is representable but not producible by anything in this repo. Where they diverge, the old form let an id-less result close a call and fire a repair that moved a message; the new form leaves the batch open for `_wire_legal_snapshot`. Correct direction.
- NITs 2-3: docstring reflow and the corrected `_drain_context_journal` rationale. Accurate — `_flush_context_journal` is the only caller (grep-confirmed).
- MINOR 3's duplicate paragraph: the surviving paragraph keeps the load-bearing "appended directly rather than returned" reasoning.

**MINOR 5's rejection reasoning holds.** Verified rather than assumed: `_pair_spliced_tool_results.__qualname__` has no dot (module-level), it is not a coroutine, and its source contains no `self.`. A once-per-session latch really would need module-global state or a threaded parameter. Accepting the rejection.

**Regression surface is clean.** The early exit returns the *same list object* for a history with no tool-calling assistant (`out is plain` → `True`), so the hot path allocates nothing. A legal batch comes back with identical ids and the *same element objects* (no copying). The delta touches nothing in Layer 1 — the parking guard, the FIFO drain, and `_drain_steering` are unchanged except for the deleted duplicate comment.

**Gates, re-run by me on `865c419d`:**

```text
$ .venv/bin/python -m flake8 .                                          clean
$ .venv/bin/python -m black --version                                   26.1.0 (pinned)
$ .venv/bin/python -m black --check local_operator tests                492 files unchanged
$ .venv/bin/python -m isort --check-only --profile black ...            clean
$ .venv/bin/python -m pyright --pythonpath .venv/bin/python .           0 errors, 0 warnings
$ env -u NO_COLOR TERM=xterm-256color .venv/bin/python -m pytest tests/unit -q -p no:randomly
2 failed, 6760 passed, 7 skipped in 200.56s
```

Both failures are the pre-existing load flakes, not breakage, and I distinguished them rather than assuming:

- `test_loop.py::test_a_late_parking_tool_is_not_robbed_of_its_end_event_mid_backfill` — **reproduced failing on base `f2a52b53`** with the identical assertion.
- `test_steering_approval.py::test_empty_assistant_message_mounts_no_block` — `RuntimeError: coroutine raised StopIteration`, a teardown race.
- Both pass 3/3 in isolation, and the PR touches neither `harness/loop.py` nor `tui/` (diff is `session/session.py`, `pyproject.toml`, and the new test file only).

CI is green on this head: all 14 checks pass across both runs, `test (3.13)` included. Round-1 MINOR 1 is properly closed.

---

#### Findings

**MINOR 1 — `test_pair_spliced_tool_results_does_not_pull_the_next_batch_into_an_open_one` passes on the unfixed code, so it does not guard what it is named for** (`tests/unit/session/test_background_journal_splice.py:626-641`).

I removed only the bound from `session.py` and ran the file: `1 failed, 16 passed` — the single failure is the *perf* test. This test passes with the bug present. Its shape leaves `toolu_1B` permanently unanswered, so `expected` never empties and `if expected: continue` suppresses the repair whether or not the scan was bounded; the assistant it claims would be swallowed never gets moved either way. The correctness half of the bound is currently pinned only indirectly, by the read-count test.

Concrete fix, one line in the fixture: make the first batch's missing answer arrive *after* the second batch opens, e.g. `history = [first, second, second_a, second_b, first_result]` where `first_result` answers `toolu_1A` and `first` opens only `toolu_1A`. On the unbounded code that yields `['first','sa','sb'...,'second']` with `second` behind its own results; on the bounded code it is unchanged. I verified that shape does discriminate.

**MINOR 2 — the remediation comment's correctness claim is stated more broadly than the code supports** (PR comment, "MINOR 2 — fixed" section).

"An unbounded scan hunting the first batch's missing answer walks into the second batch" is true only when the first batch's answer eventually arrives; for a genuinely unanswered batch the `if expected: continue` guard already prevented the bad rewrite. Worth a one-line correction on the PR so the next reader does not carry away a wrong model of when the old code misbehaved. No code change needed — the docstring at session.py:537-544 is accurate as written.

**NIT 1 — the docstring's linearity paragraph cites the perf test but not the correctness test** (`local_operator/session/session.py:537-544`). "see `test_pair_spliced_tool_results_scans_each_message_once`" covers the speed property; the correctness property gets no pointer. Add the second test name once MINOR 1 is addressed.

---

### Verdict: `clean` — round 2 is TERMINAL.

No blocker, no major. **Nothing here blocks merge.** The bound is correct in every shape I could construct, the linearity claim is now true and is genuinely pinned, the rejected alternative really is broken, the `or ""` change cannot alter wire output except in a safer direction, and MINOR 5's rejection reasoning is sound. MINOR 1 is a test that does not earn its name — worth fixing because a future refactor could remove the bound's correctness half and only the perf test would notice, but the *production* behaviour is right today and the read-count test does fail if the bound is removed. MINOR 1-2 and the nit are follow-ups, not grounds for another round.

damianvtran:

Implementer: `coder` subagent (lopdev team), model `anthropic/claude-opus-5`
Round-2 review: https://github.com/damianvtran/local-operator/pull/302#issuecomment-5416379018
Remediation commit: `1ac6ae12f6b6db640d303e6a66f9ac868b5b9e42`
Scope: `865c419db6ec5388945466e182bad64e908b269c..1ac6ae12f6b6db640d303e6a66f9ac868b5b9e42`

Round 2 was clean and terminal. All three items are follow-ups, and all three are **fixed** in one commit. The reviewer was right on both substantive points and I am not contesting either — in particular the correction to my round-1 reasoning, which I got wrong in a way that would have taught the next reader a false model.

---

#### MINOR 1 — the correctness test passed on unfixed code — **FIXED** (`1ac6ae12`)

Confirmed the finding before changing anything. In a throwaway detached worktree at `865c419d` I removed **only** the `if candidate.role == "assistant" and candidate.tool_calls: break` block from `session.py` and ran the file: `1 failed, 16 passed`, the single failure being the perf test. The test named for the correctness property passed with the bug present, exactly as reported.

Root cause is the one the reviewer named: the fixture's `toolu_1B` was never answered, so `expected` never emptied and `if expected: continue` (session.py:596) suppressed the repair regardless of the bound. The test never reached the path in its own name.

Adopted the reviewer's suggested shape — the first batch's answer arrives *after* the second batch opens, so `expected` empties and the unbounded window genuinely spans two batches:

```python
first, first_result = _batch("toolu_1A")
second, second_a, second_b = _batch("toolu_2A", "toolu_2B")
history = [first, second, second_a, second_b, first_result]
```

**Discriminating evidence, both directions, same test file.** Bound removed (throwaway worktree off `865c419d`, `session.py` edit only), new test copied in:

```text
$ .venv/bin/python -m pytest tests/unit/session/test_background_journal_splice.py -q -p no:randomly
E   AssertionError: assert ['3cca85de785...3f49ec59bb39'] == ['3cca85de785...d7b167cacaf9']
E     At index 1 diff: 'c33188598c314e6a96bb0f649a8f8c47' != 'ba582f268c50437d98093f49ec59bb39'
E     Full diff:
E       [
E           '3cca85de78524c54a4d334e50da622bc',     # first
E     -     'ba582f268c50437d98093f49ec59bb39',     # second  <- moved to the end
E           '1727cfcaf00e40a6a2641ed827dd7759',     # second_a
E           '52afd31c65144243a2e921a520995602',     # second_b
E           '687c4781cdbe47f880720403454ffe30',     # first_result
E     +     'ba582f268c50437d98093f49ec59bb39',     # second, now BEHIND its own answers
E       ]
2 failed, 15 passed in 7.91s
```

Bound restored (the PR head):

```text
$ .venv/bin/python -m pytest tests/unit/session/test_background_journal_splice.py -q -p no:randomly
17 passed in 7.82s
```

`2 failed, 15 passed` -> `17 passed`. The correctness test now fails loudly when the bound is removed, and the failure is the exact illegal shape it exists to prevent: `second` emitted after `second_a`/`second_b`. Previously this file gave `1 failed, 16 passed` under the same mutation.

Directly against the function, for the record:

```text
in         ['first', 'second', 'sa', 'sb', 'first_result']
UNBOUNDED  ['first', 'sa', 'sb', 'first_result', 'second']   <- illegal
BOUNDED    ['first', 'second', 'sa', 'sb', 'first_result']   <- unchanged, correct
```

#### MINOR 2 — the correctness claim was stated too broadly — **FIXED in code, and corrected here** (`1ac6ae12`)

The reviewer is right and my round-1 remediation was wrong. "An unbounded scan hunting the first batch's missing answer walks into the second batch" is **not** true for a genuinely unanswered batch: `expected` never empties there, so the `if expected: continue` guard already prevented the bad rewrite, bound or no bound. I verified this myself rather than taking it on trust — the shape I cited in round 1 comes out byte-identical with the bound stripped.

**The record, corrected:** the bound's correctness half is needed for exactly one shape, a **late-arriving answer**, where `expected` empties only after the next assistant has opened and the repair therefore fires on a window spanning two batches. The fix is unchanged and correct; only my explanation of *when* the old code misbehaved was too broad.

The reviewer noted no code change was needed since the docstring at 537-544 was accurate. I went slightly further than that on purpose: the load-bearing wrong statement was in the inline comment at the `break` (the old "Swallowing the next batch would move ITS results..." paragraph), which is what a future reader deleting the bound would actually read. It now states the reproducing shape concretely and warns off the wrong one:

```
# It is also the more correct stop, in one specific shape: a
# batch whose answer arrives LATE, after the next batch has
# already opened. Only then does `expected` empty at all, so
# only then can the repair fire on a window that spans two
# batches -- `A1(x) A2(y,z) T(y) T(z) T(x)` would come out as
# `A1 T(y) T(z) T(x) A2`, with A2 emitted BEHIND its own answers.
#
# A batch that is never answered at all is NOT this case, and
# was never at risk: `expected` stays non-empty, so the
# `if expected: continue` guard below suppresses the repair
# with or without this bound. Do not reach for an unanswered
# batch to justify the stop -- reach for the late answer.
```

The new test's docstring carries the same correction, including why the old fixture was vacuous, so the mistake is not repeatable from the test side either.

#### NIT 1 — linearity paragraph cited only the perf test — **FIXED** (`1ac6ae12`)

The docstring now names both tests and says why there are two: `test_pair_spliced_tool_results_scans_each_message_once` for the speed property, `test_pair_spliced_tool_results_does_not_pull_the_next_batch_into_an_open_one` for the correctness property, "pinned separately because the bound carries both."

---

#### Gates on `1ac6ae12`

```text
$ .venv/bin/python -m flake8 .                                          clean
$ .venv/bin/python -m black --version                                   26.1.0 (pinned, compiled)
$ .venv/bin/python -m black --check local_operator tests                492 files unchanged
$ .venv/bin/python -m isort --check-only --profile black local_operator tests   clean
$ .venv/bin/python -m pyright --pythonpath .venv/bin/python .           0 errors, 0 warnings, 0 informations
$ env -u NO_COLOR TERM=xterm-256color .venv/bin/python -m pytest tests/unit -q -p no:randomly
6762 passed, 7 skipped, 431 warnings in 242.65s
```

Full suite is **fully green** on this head, including the two pre-existing flakes both reviewers reproduced on base `f2a52b53` (`test_loop.py::test_a_late_parking_tool_is_not_robbed_of_its_end_event_mid_backfill` and the TUI wall-clock cases). They were left alone as out of scope, and this run happened not to be loaded enough to trip them — that is the flake behaving as characterised, not a fix. Test count is 6762 against round 2's 6760 because two tests were added earlier in this PR; this commit changes one existing test's fixture and adds none.

`black` is the pinned 26.1.0 from the venv; `uvx` still cannot spawn in this environment (`Failed to spawn: black`), and the venv carries the identical pinned version.

Diff is 2 files, +37/-11: `session/session.py` (comments and docstring only, no executable change) and the test fixture. No production behaviour is altered by this commit.

Not merging — reporting ready per the standing instruction.

damianvtran:
### Agent review — round 3

Reviewer: `reviewer` subagent (lopdev team), models `anthropic/claude-opus-5` and `openai/gpt-5.6-sol` (provider failover)
Scope: `865c419d..1ac6ae12`

Freshness-delta review only. **No findings.**

- **Production diff is inert:** `local_operator/session/session.py` changes only its docstring and inline comments. Its AST is identical across the scope after docstrings are stripped; no executable statement, default, or control-flow edge changed.
- **The rebuilt test discriminates:** in a fresh throwaway worktree at `1ac6ae12`, the focused file passes with the bound restored (`17 passed`). Removing only the 23-line next-batch bound block gives `2 failed, 15 passed`; both the linear-scan test and `test_pair_spliced_tool_results_does_not_pull_the_next_batch_into_an_open_one` fail, with the latter exposing the illegal `A1 T(y) T(z) T(x) A2` ordering. As a control, the old round-2 never-answered fixture still passes against that bound-removed implementation, confirming why it was vacuous.
- **The corrected explanation is accurate:** direct execution confirms a never-answered batch remains unchanged with or without the bound because `expected` never empties, while `A1(x) A2(y,z) T(y) T(z) T(x)` is corrupted only without the bound. The docstring and inline comment now teach that exact late-answer model and explicitly warn against treating the bound as inert.
- **Nothing regressed:** flake8, `black==26.1.0 --check`, isort, and pyright pass. The full `tests/unit` suite passes with `6762 passed, 7 skipped`. All 14 GitHub checks are green on current head `1ac6ae12`.

**Verdict: clean (TERMINAL).** No BLOCKER or MAJOR findings remain, and nothing in this delta blocks merge.

PR #301, opened 2026-08-25
feat(resume): restore the attached team, agent and goal on resume

damianvtran:
## Summary

Makes a session's attached identity survive a resume. The `/team` roster, the `/agent` profile and the `/goal` are journalled to a sidecar beside the transcript and re-resolved at construction, so reopening a conversation brings back the persona it was running under rather than silently downgrading to the base voice.

- New `attachment.json` sidecar in the session directory, following the existing `origin.json` / `title.json` conventions exactly: atomic pid-named temp plus `replace`, tolerant reads, best-effort by contract, directory mtime preserved.
- Written on change only (attach, detach, `/goal` set or clear), never per turn.
- Restored at `Session` construction, beside the existing wake, title and active-route restores.
- A team or profile that no longer resolves degrades to unattached, names what was lost as a system notice, and leaves the band segment blank.
- Version bumps to `0.36.0` as a minor: this restores expected behaviour but also adds a new persisted-state capability. The version moved twice during review as main claimed `0.34.0` (PR #305) and then `0.35.0` (PR #299); the branch advances to the next free minor.

## Impact

This was reported as the status band going blank after a resume, with the assumption that the persona stayed in force through conversation history. That is not what happened.

The team and agent briefs ride the **volatile tail** of the system prompt, which is built from a `GoalState` that `session_factory` constructs empty on every session (`session_factory.py`, the `goal_state = GoalState()` line). Nothing in the transcript reproduced it. So a `--resume` did not merely blank a segment: it dropped the briefs from the model's instructions entirely, and the conversation carried on as an ordinary session. The blank band was reporting the truth, which is why this is a behavioural fix and not a display one, and why the band is still driven from the session rather than painted from the sidecar.

**Names are persisted, never brief bodies.** The restore re-resolves each name through the same entry points `/team` and `/agent` use, sharing the one `_resolve_profile_or_specialist` ordering (own role, then own specialist, then packaged seed). That is deliberate: the operator edits teams and profiles between sessions, and a stored brief would resume them onto instructions that no longer exist anywhere. It also means resolution order cannot drift between the attach path and the resume path.

The restore mutates the shared holder the system-blocks provider already closed over, so only the volatile tail moves and the cached persona prefix stays warm. Nothing is rebuilt.

## Before reproduction

Driven through the real `session_factory.create_session`, which is the path `lop --resume <id>` takes, on `origin/main` at f2a52b53:

```text
[session 1] id=2dca944b19fa team='lopdev' agent='auditor'
[...]

PR #304, opened 2026-08-25
feat(multiplexer): resume the right session in each pane after a crash

damianvtran:
## Summary

~15 `lop` sessions run as cmux sidebar workspaces. When cmux crashes the workspaces return, but every surface opens a fresh shell and a fresh `lop`, and all conversation continuity is lost. The sessions survive on disk under `~/.local-operator/sessions`; nothing maps a pane back to the session it held.

This publishes that missing fact per pane: *this pane holds session `<id>`, and here is the argv that reopens it*.

- cmux gets a real resume binding over its control socket.
- tmux, WezTerm, Zellij and screen have no resume API, so they get a documented, discoverable pane marker a restore script can consume.
- Backends sit behind one interface in a registry, so another multiplexer is a new module rather than an edit to a chain of conditionals.
- Version bumps from `0.33.1` to `0.34.0` as a minor feature.

Change class: **C2, standard / pre-authorized.** No RFC or tracker record is attached; this MR is the system of record.

## Impact

Before: a cmux crash cost every running conversation. Recovery meant remembering which pane held which session and retyping `lop --resume <id>` for each one.

After: each pane advertises its session, so a restore reopens the conversation that pane was holding.

Publication is best-effort by construction. A missing binary, a socket error, a multiplexer that is not running, or a timeout is logged at debug and otherwise ignored. It can never be the reason a session fails to start.

## Design points worth review attention

**`source: agent-hook` over the CLI.** `cmux surface resume set` hardcodes `source = "cli"`, which cmux resolves to `approvalPolicy = .manual`, `autoResume = false`. Only the socket RPC reaches auto. A future simplification to the CLI would still create a binding and still print one, and auto-resume would be silently dead. Commented at the top of `multiplexer/cmux.py`.

**The re-assert timer is justified by cmux's source, not by a measurement here.** `SharedLiveAgentIndex.cacheTTL` is 60s; a binding published against an index snapshot older than this process is retired by `retireAgentHookResumeBinding`, which latches `autoResume = false`. `REASSERT_INTERVAL_S = 90` therefore has to stay above that TTL. The measurements on this host can neither confirm nor refute the race (see limitations), so the defence is retained on the strength of the Swift.

**Restore-and-idle is a safety boundary, not a default.** `resume_argv` is the single place the command is built, and it can only ever emit `lop --resume <id>`. Fifteen panes restoring unattended must not resume tool execution.

**Two gates, different lifetimes.** Subagent-ness is permanent and decided once (`origin.json`); resumability is transient and re-checked before every publish, because a cold session has no transcript until its first turn lands. Checking resumability once at startup would mean most sessions silently never publish, discovered only after a crash.

**Headless gate on the app hook.** This suite runs inside cmux and `tests/conftest.py` does not clear `CMUX_*`, so without the gate every pilot test would rewrite the resume binding of the session running the tests. A paired test proves the gate is load-bearing: same arrangement, gate lifted, produces `surface.resume.set`.

## Testing evidence

Full artifacts in `docs/evidence/multiplexer-resume/`. **Read `cmux-host-caveat.md` first** — it bounds every cmux claim below.

**Strongest evidence: the binding reaches the file a crash restores from.** `cmux-on-disk-binding.txt` reads it back out of `~/Library/Application Support/cmux/session-com.cmuxterm.app.json`, which is written on cmux's autosave rather than on clean shutdown:

```text
kind            local-operator
checkpointId    c9a4b15eed9e
source          agent-hook
autoResume      true
approvalPolicy  auto
```

Identity was verified rather than eyeballed: `panels[0].id == CMUX_SURFACE_ID` and `workspaces[8].workspaceId == CMUX_WORKSPACE_ID`, both exact. An in-memory binding that never reached disk would be the classic "looks fixed and is not".

**Live publish and retire** (`cmux-publish-retire.txt`): a real session publishes `source: agent-hook`, `approval_policy: auto`, `auto_resume: true`, with `checkpoint_id` equal to its real session directory name. The binding is gone after retire.

**The four refusals** (`negative-cases.txt`), all publishing nothing: subagent session, kill switch set, no multiplexer in the environment, cmux binary unresolvable.

**tmux exercised for real** (`tmux.txt`, reproducible via `tmux-evidence.sh`): against a private tmux server on its own socket, both pane options are written and read back with `show-options -pv`, and after retire both are genuinely *unset* (`rc=1`) rather than blanked. A blank option would read to a restore script as a session with an empty id.

## Honest limitations

These are stated because the evidence does not support the stronger claim, and a reviewer should not infer coverage that does not exist.

- The cmux host was running from a **deleted binary** during capture (`Contents/MacOS/` empty; the app executing an old in-memory image). Behaviour was not self-consistent across the update boundary: bindings held for minutes, then retired in 2-8s, and in one control run a binding stayed `auto_resume: true` for 75s *after its agent process was killed*, which a working `isStaleAgentHookBinding` should have retired. That last result indicates cmux's liveness retirement was not running at all.
- Consequently the 45/45-samples-true duty cycle measured here is **not** a product guarantee. It was identical with and without the vault registration, so this host did **not** isolate the registration's causal role.
- The vault registration is documented as required because cmux's own `docs/vault.md` requires it for a custom kind, **not** because it was verified here to be what holds the binding.
- **Outstanding: clean re-validation on a healthy cmux install.** The host binary has since been restored, but the running process is still the old in-memory image, so a restart is required before the end-to-end path can be confirmed. That restart is itself the first real-world test of this feature. Worst case the sessions come back cold, which is exactly today's behaviour and no worse.

## Gates

```text
flake8 .                             clean
black --check (26.1.0)               clean
isort --check-only --profile black   clean
pyright                              0 errors, 0 warnings
unit suite                           6788 passed, 7 skipped
```

The 2 failures observed (`test_comms.py::test_a_resumed_child_replays_the_stopped_ones_transcript`, `::test_a_cancelled_child_keeps_the_work_it_had_already_done`) are **pre-existing flakes**, verified rather than assumed: clean `origin/main` fails the same files with a *different* test on each full run (four distinct failure sets observed), and both pass in isolation on both branches. The 45 new tests pass 4/4 consecutive runs.

## Rollback

Revert the merge commit. The feature adds no migration and no persisted state of its own; the only external side effect is a cmux resume binding, which cmux clears on its own once the publishing process exits.

### DIFF
--- a/docs/evidence/multiplexer-resume/PR-NOTES.md
+++ b/docs/evidence/multiplexer-resume/PR-NOTES.md
@@ -1,0 +1,115 @@
+# PR body source material (branch held pending cmux re-validation)
+
+Assembled while the evidence was fresh so the PR can be opened without
+re-deriving any of it. Not intended to ship as documentation — delete or leave,
+but do not treat it as user-facing.
+
+## Title
+
+`feat(multiplexer): resume the right session in each pane after a crash`
+
+## Summary
+
+~15 `lop` sessions run as cmux sidebar workspaces. When cmux crashes the
+workspaces return but every surface opens a fresh shell and a fresh `lop`, and
+all conversation continuity is lost — the sessions survive on disk under
+`~/.local-operator/sessions` but nothing maps a pane to the session it held.
+
+This publishes that missing fact per pane: *this pane holds session `<id>`, and
+here is the argv that reopens it*. cmux gets a real resume binding over its
+control socket. tmux/wezterm/zellij/screen have no resume API, so they get a
+documented, discoverable marker a restore script can consume. Backends sit
+behind one interface in a registry, so another multiplexer is a new module.
+
+Minor bump: 0.33.1 -> 0.34.0.
+
+## Design points worth calling out in review
+
+- **`source: agent-hook` over the CLI.** `cmux surface resume set` hardcodes
+  `source = "cli"`, which cmux resolves to `approvalPolicy = .manual`,
+  `autoResume = false`. Only the socket RPC can reach auto. A future
+  simplification to the CLI would still create a binding and still print one,
+  and auto-resume would be silently dead. Commented at the top of
+  `multiplexer/cmux.py`.
+- **The re-assert timer is justified by cmux's SOURCE, not by a measurement
+  here.** `SharedLiveAgentIndex.cacheTTL` is 60s; a binding published against
+  an index snapshot older than this process is retired by
+  `retireAgentHookResumeBinding`, which latches `autoResume = false`
+  permanently. `REASSERT_INTERVAL_S = 90` therefore has to stay above that TTL.
+  The measurements on this host can neither confirm nor refute the race (see
+  caveat below), so the defence is retained on the strength of the Swift.
+- **Restore-and-idle is a safety boundary, not a default.** `resume_argv` is
+  the single place the command is built and it can only ever emit
+  `lop --resume <id>`. Fifteen panes restoring unattended must not resume tool
+  execution.
+- **Two gates, different lifetimes.** Subagent-ness is permanent and decided
+  once (`origin.json`); resumability is transient and re-checked before every
+  publish, because a cold session has no transcript until its first turn lands.
+  Checking resumability once at startup would mean every cold session — most of
+  them — silently never publishes, discovered only after a crash.
+- **Headless gate on the app hook.** This suite runs inside cmux and
+  `tests/conftest.py` does not clear `CMUX_*`, so without it every pilot test
+  would rewrite the resume binding of the session running the tests. A
+  paired test proves the gate is what blocks it: same arrangement, gate
+  lifted, produces `surface.resume.set`.
+
+## Testing evidence
+
+Full artifacts in `docs/evidence/multiplexer-resume/`. **Read
+`cmux-host-caveat.md` first** — it bounds every cmux claim below.
+
+**Strongest evidence — it reaches the file a crash restores from.**
+`cmux-on-disk-binding.txt`: read back out of
+`~/Library/Application Support/cmux/session-com.cmuxterm.app.json` (written on
+cmux's autosave, not on clean shutdown) while a binding was in force:
+`kind=local-operator`, `checkpointId=c9a4b15eed9e`, `source=agent-hook`,
+`autoResume=true`, `approvalPolicy=auto`. Identity verified rather than
+eyeballed: `panels[0].id == CMUX_SURFACE_ID` and
+`workspaces[8].workspaceId == CMUX_WORKSPACE_ID`, both exact. An in-memory
+binding that never reached disk would be the classic "looks fixed and is not".
+
+**Live publish/retire** (`cmux-publish-retire.txt`): a real session publishes
+`source: agent-hook`, `approval_policy: auto`, `auto_resume: true`, with
+`checkpoint_id` equal to its real directory name; binding gone after retire.
+
+**The four refusals** (`negative-cases.txt`), all publishing nothing: subagent
+session, kill switch, no multiplexer in env, cmux binary unresolvable.
+
+**tmux for real** (`tmux.txt`, reproducible via `tmux-evidence.sh`): against a
+private tmux server on its own socket, both pane options written and read back
+with `show-options -pv`, and after retire both genuinely *unset* (`rc=1`) rather
+than blanked — a blank option would read to a restore script as a session with
+an empty id.
+
+## Honest limitations (must survive into the PR body)
+
+- The cmux host was running from a **deleted binary** during capture
+  (`Contents/MacOS/` empty; app executing an old in-memory image). Behaviour was
+  not self-consistent across the update boundary: bindings held for minutes,
+  then retired in 2-8s, and in one control run a binding stayed
+  `auto_resume: true` for 75s **after its agent process was killed** — which a
+  working `isStaleAgentHookBinding` should have retired. That last result says
+  cmux's liveness retirement was not running at all.
+- Consequently: the 45/45-samples-true duty cycle measured here is **not** a
+  product guarantee, and it was identical with and without the vault
+  registration, so this host did **not** isolate the registration's causal role.
+- The vault registration is documented as required because cmux's own
+  `docs/vault.md` requires it for a custom kind — **not** because it was
+  verified here to be what holds the binding. Reviewers should not infer
+  otherwise.
+- **Outstanding: clean re-validation on a healthy cmux install.** The operator
+  restarts cmux at a time of his choosing; that restart is also the first
+  real-world test of this feature.
+
+## Gates
+
+- `flake8 .` clean
+- `black --check` (26.1.0) clean, `isort --check-only --profile black` clean
+- `pyright` 0 errors, 0 warnings
+- Unit suite: 6788 passed, 7 skipped. The 2 failures seen
+  (`test_comms.py::test_a_resumed_child_replays_the_stopped_ones_transcript`,
+  `::test_a_cancelled_child_keeps_the_work_it_had_already_done`) are
+  **pre-existing flakes**: clean `origin/main` fails the same files with a
+  *different* test on each full run (four distinct failure sets observed across
+  runs), and both pass in isolation on both branches.
+- The 45 new tests pass 4/4 consecutive runs with no flakiness.
--- a/docs/evidence/multiplexer-resume/README.md
+++ b/docs/evidence/multiplexer-resume/README.md
@@ -1,0 +1,60 @@
+# Evidence: multiplexer session resume
+
+Captured on macOS 25.6.0, cmux 0.64.22 (102), tmux 3.7c, on 2026-08-25.
+
+Read `cmux-host-caveat.md` first: the cmux host was in a damaged state during
+this capture, and it bounds what the cmux numbers below can be claimed to show.
+
+| File | What it shows |
+|---|---|
+| `cmux-publish-retire.txt` | A real publish and retire against a live cmux surface |
+| `cmux-on-disk-binding.txt` | The binding in the file cmux restores from after a crash |
+| `negative-cases.txt` | The four refusals: subagent, kill switch, no multiplexer, no binary |
+| `tmux.txt` | tmux exercised against a real server: options written, read back, unset |
+| `cmux-host-caveat.md` | Why the cmux host's state limits the cmux claims |
+
+## What is proven
+
+**The publish path works.** `cmux-publish-retire.txt` shows a real session
+publishing `source: agent-hook`, `approval_policy: auto`, `auto_resume: true`,
+`kind: local-operator`, with `checkpoint_id` equal to the session's actual
+directory name under `~/.local-operator/sessions`, and the binding gone again
+after retire.
+
+**It survives to disk, for the right surface.** `cmux-on-disk-binding.txt` reads
+the binding back out of
+`~/Library/Application Support/cmux/session-com.cmuxterm.app.json` — the file a
+crash restores from, written on cmux's autosave rather than on clean shutdown —
+and shows `autoResume: true`. Identity is verified rather than eyeballed:
+`panels[0].id` equals `CMUX_SURFACE_ID` and `workspaces[N].workspaceId` equals
+`CMUX_WORKSPACE_ID`, both exact. This is the property that would otherwise
+"look fixed and not be": an in-memory binding that never reached disk would be
+useless in the only situation it exists for.
+
+**The refusals hold.** `negative-cases.txt` shows all four publishing nothing and
+leaving no binding: a subagent session (`origin.json`), the
+`LOCAL_OPERATOR_NO_MULTIPLEXER_RESUME` kill switch, an environment with no
+multiplexer at all, and a cmux surface whose CLI cannot be resolved.
+
+**tmux is real, not mocked.** `tmux.txt` runs against a private tmux server on
+its own socket (so it cannot touch a real session): the backend writes both pane
+options, they are read back with `show-options -pv`, and after retire both are
+genuinely *unset* — `rc=1`, absent — rather than set to an empty string that a
+restore script would misread as a session with a blank id.
+
+## Reproducing
+
+The cmux captures need a live cmux surface, so they are scripts rather than
+tests. tmux is reproducible anywhere tmux is installed:
+
+```sh
+bash docs/evidence/multiplexer-resume/tmux-evidence.sh
+```
+
+The unit suite covers the same contracts without needing either multiplexer:
+
+```sh
+.venv/bin/python -m pytest tests/unit/test_multiplexer.py -q
+env -u NO_COLOR TERM=xterm-256color .venv/bin/python -m pytest \
+  tests/unit/tui/test_multiplexer_broadcast.py -q
+```
--- a/docs/evidence/multiplexer-resume/cmux-host-caveat.md
+++ b/docs/evidence/multiplexer-resume/cmux-host-caveat.md
@@ -1,0 +1,65 @@
+# Caveat: the state of the cmux host during this capture
+
+Recorded because it bounds what the cmux evidence in this directory can honestly
+be claimed to show. It is a fact about the machine, not about this change.
+
+## What was wrong
+
+During validation the installed cmux was mid-update and damaged:
+
+- `/Applications/cmux.app/Contents/MacOS/` was **empty** (mtime 13:55). The main
+  application executable was gone from disk.
+- The running app (pid 73906, started 13:45) was still executing from memory out
+  of that deleted binary, so `cmux version` kept reporting
+  `0.64.22 (102) [ddd4a01bc]` and the control socket kept answering.
+- The bundled CLI at `Contents/Resources/bin/cmux` was intact, which is why
+  every `cmux rpc` call in these captures worked normally.
+
+## What it means for the measurements
+
+Observations taken before and after that ~13:55 boundary are not from the same
+build, and behaviour was not self-consistent across it. Specifically:
+
+- Early runs showed a published binding holding `auto_resume: true` for minutes.
+- Later runs showed the same binding retiring to `manual` within 2–8 seconds.
+- A control run then showed a binding **staying** `auto_resume: true` for 75s
+  after its agent process was killed — which a working
+  `isStaleAgentHookBinding` should have retired.
+
+The last of those is the informative one: it says cmux's liveness-based
+retirement was **not running** at that point. That also means this host could not
+be used to demonstrate the retirement rule the vault registration defends
+against, and equally could not be used to prove the registration is what defeats
+it. On this host, a settled process measured a 100% duty cycle
+(45/45 samples over 3 minutes at 4s resolution) both **with** and **without** the
+registration installed — so the registration was not the discriminator here.
+
+## What is therefore claimed, and what is not
+
+**Claimed** (independent of the retirement machinery, and directly observed):
+
+- `lop` publishes a correct `agent-hook` binding with `auto_resume: true`.
+- That binding reaches the on-disk autosave a crash restores from, against the
+  correct workspace and surface ids.
+- It is withdrawn on a clean exit, and is never published for a subagent
+  session, under the kill switch, without a multiplexer, or without a cmux CLI.
+
+**Not claimed:**
+
+- Any steady-state duty cycle figure as characterising a *healthy* cmux. The
+  100% measured here was taken on a host whose retirement path was demonstrably
+  inactive, so it is not evidence that retirement has been defeated.
+- That the vault registration is what keeps the binding alive. cmux's own
+  `docs/vault.md` documents it as the supported mechanism, and it is documented
+  as required for that reason — not because this capture isolated its effect.
+
+The re-assert timer in `multiplexer/cmux.py` is retained on the strength of
+cmux's source (`SharedLiveAgentIndex.cacheTTL` is 60s and
+`retireAgentHookResumeBinding` latches `autoResume = false`), not on the strength
+of these measurements. It is cheap, silent, and stops on exit, so it costs
+little if this host's behaviour turns out to be the normal one.
+
+A clean reinstall of cmux is worth doing before treating any cmux-side timing
+number from this machine as authoritative. It was deliberately not attempted
+here: the operator had ~15 live sessions in that process, and restarting cmux to
+tidy up an evidence run would have destroyed them.
--- a/docs/evidence/multiplexer-resume/cmux-on-disk-binding.txt
+++ b/docs/evidence/multiplexer-resume/cmux-on-disk-binding.txt
@@ -1,0 +1,29 @@
+$ date "+%H:%M:%S"
+16:21:51
+
+# The file cmux restores from after a crash (written on autosave, not on
+# clean shutdown), read back while a published binding was in force.
+$ python3 - <<'PY'
+... walk session-com.cmuxterm.app.json for resumeBinding kind=local-operator ...
+PY
+
+local-operator resumeBindings on disk: 1
+{
+  "kind": "local-operator",
+  "checkpointId": "c9a4b15eed9e",
+  "source": "agent-hook",
+  "autoResume": true,
+  "approvalPolicy": "auto",
+  "command": "cd -- '/tmp/lop-final' 2>/dev/null || [ ! -d '/tmp/lop-final' ] && lop --resume c9a4b15eed9e",
+  "cwd": "/tmp/lop-final"
+}
+
+# Identity verified exactly, not by eyeball:
+path:            .windows[0].tabManager.workspaces[8].panels[0].terminal
+panels[0].id:    3E63FAC7-CA22-4AD5-9151-874E05B5A490
+CMUX_SURFACE_ID: 3E63FAC7-CA22-4AD5-9151-874E05B5A490
+MATCH panel==surface: True
+
+workspaces[8].workspaceId: B6A8E3AD-DAD2-40B8-8C6A-732351BEF641
+CMUX_WORKSPACE_ID:         B6A8E3AD-DAD2-40B8-8C6A-732351BEF641
+MATCH workspace: True
--- a/docs/evidence/multiplexer-resume/cmux-publish-retire.txt
+++ b/docs/evidence/multiplexer-resume/cmux-publish-retire.txt
@@ -1,0 +1,15 @@
+backend detected: cmux
+before: {}
+broadcast handle: SessionBroadcast
+AFTER PUBLISH:
+{
+  "source": "agent-hook",
+  "approval_policy": "auto",
+  "auto_resume": true,
+  "checkpoint_id": "c9a4b15eed9e",
+  "kind": "local-operator",
+  "command": "cd -- '/private/tmp/lop-mux-resume' 2>/dev/null || [ ! -d '/private/tmp/lop-mux-resume' ] && /private/tmp/mux_evidence.py --resume c9a4b15eed9e",
+  "cwd": "/private/tmp/lop-mux-resume"
+}
+RETIRE:
+after retire: {}
--- a/docs/evidence/multiplexer-resume/negative-cases.txt
+++ b/docs/evidence/multiplexer-resume/negative-cases.txt
@@ -1,0 +1,17 @@
+=== NEGATIVE 1: subagent session must NOT broadcast ===
+binding before: (none)
+broadcast handle: None
+binding after  : (none)
+
+=== NEGATIVE 2: kill switch ===
+handle: None | binding: (none)
+
+=== NEGATIVE 3: no multiplexer in env ===
+active_backend(no-mux env): None
+handle: None | binding: (none)
+
+=== NEGATIVE 4: cmux binary absent ===
+active_backend(no binary): None
+handle: None
+
+FINAL binding (must be none): (none)
--- a/docs/evidence/multiplexer-resume/round-1-remediation.txt
+++ b/docs/evidence/multiplexer-resume/round-1-remediation.txt
@@ -1,0 +1,116 @@
+Agent review round 1 — remediation evidence
+===========================================
+
+Re-runs of the reviewer's own reproductions, before and after the fix, plus a
+real-multiplexer check for F3. Harnesses are the reviewer's, unchanged.
+
+
+F1 — stop() must not block the Textual event loop
+-------------------------------------------------
+Harness: a backend whose publish/retire both sit for CALL_TIMEOUT_S (5.0s),
+mirroring a `cmux rpc` parked in `subprocess.run(timeout=...)`. A publish is
+allowed to get INTO the wedged call first, then stop(retire=True) is timed.
+
+    BEFORE (6edcf186)
+      F1 [hanging] stop() blocked the calling thread for 9.51s
+      F1 [healthy] stop() blocked the calling thread for 0.02s
+
+    AFTER
+      F1 [hanging] stop() blocked the calling thread for 0.01s
+      F1 [healthy] stop() blocked the calling thread for 0.00s
+
+The withdrawal still happens — it is dispatched to a daemon worker and observed
+completing once the wedged backend is released. What no longer happens is the
+event loop waiting for it. The `_retired` latch is set synchronously inside
+stop(), so the clean-exit-wins-over-re-assert rule does not depend on the
+worker running at all.
+
+
+F2 — a failing publish must not pin the loop to the 5s poll
+-----------------------------------------------------------
+Harness: a backend whose publish always returns False, against a session that
+IS resumable, with a 90s re-assert interval, observed for ~21s.
+
+    BEFORE (6edcf186)
+      F2 publish attempts in ~21s with a 90s re-assert interval: 5
+      F2 gaps between attempts (s): [5.01, 5.0, 5.0, 5.01]
+
+    AFTER
+      F2 publish attempts in ~21s with a 90s re-assert interval: 1
+
+One attempt: the immediate publish at start. Everything after it is governed by
+the re-assert interval, because the cadence now keys on is_resumable_session()
+rather than on publish success. The fast cadence still exists and still only
+costs a stat — pinned by
+test_a_cold_session_still_polls_until_its_transcript_appears, which shows a
+cold session publishing within one poll of its transcript appearing.
+
+
+F3 — zellij markers must not collide across panes
+--------------------------------------------------
+Not a mock. A real zellij 0.42.2 session with two real panes, each running the
+actual registry + ZellijBackend, publishing a different session id.
+(cmux/tmux vars cleared in the probe because this zellij is nested inside them
+on this host and the registry deliberately prefers the outermost multiplexer.)
+
+Environment observed inside the two panes:
+
+    pane A: ZELLIJ=0  ZELLIJ_SESSION_NAME=<same>  ZELLIJ_PANE_ID=0
+    pane B: ZELLIJ=0  ZELLIJ_SESSION_NAME=<same>  ZELLIJ_PANE_ID=1
+
+    BEFORE (6edcf186) — both panes published successfully, ONE file resulted:
+      zellij-f3old.json -> bbbbbbbbbbbb
+      (pane 0's binding is gone: last publisher won)
+
+    AFTER — one file per pane, each holding its own session:
+      zellij-f3c-0.json -> aaaaaaaaaaaa
+      zellij-f3c-1.json -> bbbbbbbbbbbb
+
+The sibling-deletion half is pinned by
+test_two_zellij_panes_in_one_session_do_not_share_a_marker, which retires pane
+0 and asserts pane 1's marker survives.
+
+
+F4 — screen window identity
+---------------------------
+Verified on screen 4.00.03 on this host that a process inside a session exports
+both, confirming the reviewer's reading:
+
+    STY=42653.loprev
+    WINDOW=0
+
+Keyed on STY+WINDOW, with STY alone kept as a fallback (documented in the
+module docstring and the docs table, since in that case the marker is per
+session rather than per window).
+
+
+Gates
+-----
+flake8 clean; black --check 26.1.0 clean (499 files); isort --check-only
+--profile black clean; pyright 0 errors, 0 warnings; unit suite 6797 passed.
+
+Multiplexer suites specifically: 55 passed, run 3x consecutively with no
+flake (tests/unit/test_multiplexer.py + tests/unit/tui/test_multiplexer_broadcast.py).
+
+Three failures in the full run are pre-existing and unrelated to this diff,
+confirmed by reproducing them on unmodified 6edcf186 in a detached worktree:
+  - tests/unit/tui/test_app_pilot.py::test_esc_aborts_a_running_shell_command
+    (fails on the unmodified head too)
+  - tests/unit/harness/test_comms.py::test_a_resumed_child_replays_the_stopped_ones_transcript
+  - tests/unit/harness/test_comms.py::test_a_cancelled_child_keeps_the_work_it_had_already_done
+    (both timing-sensitive: 2 clean runs then a failure on the unmodified head)
+
+
+New tests, and that they fail without the fix
+---------------------------------------------
+The eight new/changed tests were run against unmodified 6edcf186 with only the
+test file copied in. All eight failed, e.g.:
+
+    test_stop_does_not_block_the_caller_on_a_wedged_backend
+      AssertionError: stop() blocked the caller for 59.82s
+    test_a_failing_publish_backs_off_to_the_reassert_cadence
+      AssertionError: failing publish retried 3x on the fast poll
+    test_two_zellij_panes_in_one_session_do_not_share_a_marker
+      + 'zellij-main.json'  - 'zellij-main-1.json'
+    test_the_retire_latch_is_set_before_stop_returns
+      AssertionError: assert ['retire'] == []
--- a/docs/evidence/multiplexer-resume/tmux-evidence.sh
+++ b/docs/evidence/multiplexer-resume/tmux-evidence.sh
@@ -1,0 +1,43 @@
+set -e
+SOCK=lopmuxtest
+tmux -L $SOCK kill-server 2>/dev/null || true
+tmux -L $SOCK new-session -d -s ev -x 100 -y 30
+PANE=$(tmux -L $SOCK list-panes -t ev -F '#{pane_id}' | head -1)
+echo "pane=$PANE"
+tmux -L $SOCK send-keys -t "$PANE" "echo ready" Enter
+sleep 0.3
+
+# Run the REAL backend inside the tmux pane's environment.
+tmux -L $SOCK run-shell -t "$PANE" "true" 2>/dev/null || true
+TMUXENV=$(tmux -L $SOCK show-environment -t ev 2>/dev/null | head -1 || true)
+
+/tmp/lop-mux-resume/.venv/bin/python - "$SOCK" "$PANE" <<'PY'
+import os, subprocess, sys
+sys.path.insert(0,"/tmp/lop-mux-resume")
+sock, pane = sys.argv[1], sys.argv[2]
+from local_operator.multiplexer.markers import TmuxBackend, SESSION_OPTION, COMMAND_OPTION
+from local_operator.multiplexer.broadcast import build_binding
+from local_operator.multiplexer.registry import active_backend
+import local_operator.multiplexer.markers as m
+
+# tmux backend shells out to `tmux`; point it at our private socket so this
+# cannot touch any real tmux server.
+real_run = m._run
+m._run = lambda argv: real_run([argv[0], "-L", sock] + argv[1:])
+
+env = {"TMUX": f"/tmp/tmux-{os.getuid()}/{sock},1,0", "TMUX_PANE": pane}
+print("detected backend:", type(active_backend(env)).__name__)
+b = build_binding("abc123abc123", cwd="/work")
+print("publish ->", TmuxBackend().publish(b, env))
+for opt in (SESSION_OPTION, COMMAND_OPTION):
+    out = subprocess.run(["tmux","-L",sock,"show-options","-pv","-t",pane,opt],
+                         capture_output=True, text=True)
+    print(f"  {opt} = {out.stdout.strip()!r}")
+print("retire ->", TmuxBackend().retire(b, env))
+for opt in (SESSION_OPTION, COMMAND_OPTION):
+    out = subprocess.run(["tmux","-L",sock,"show-options","-pv","-t",pane,opt],
+                         capture_output=True, text=True)
+    print(f"  {opt} after retire = {out.stdout.strip()!r} (rc={out.returncode})")
+PY
+tmux -L $SOCK kill-server 2>/dev/null || true
+echo "tmux server torn down"
--- a/docs/evidence/multiplexer-resume/tmux.txt
+++ b/docs/evidence/multiplexer-resume/tmux.txt
@@ -1,0 +1,9 @@
+pane=%0
+detected backend: TmuxBackend
+publish -> True
+  @lop_session = 'abc123abc123'
+  @lop_resume_command = 'lop --resume abc123abc123'
+retire -> True
+  @lop_session after retire = '' (rc=1)
+  @lop_resume_command after retire = '' (rc=1)
+tmux server torn down
--- a/docs/multiplexer-resume.md
+++ b/docs/multiplexer-resume.md
@@ -1,0 +1,189 @@
+# Multiplexer session resume
+
+When a terminal multiplexer dies and comes back, its panes come back too — but
+each one opens a fresh shell and a fresh `lop`. Every conversation is still on
+disk under `~/.local-operator/sessions`; none of them is reachable except by
+remembering which pane held which session and retyping `lop --resume <id>`.
+
+`lop` now publishes, per pane, the one fact the multiplexer cannot derive: *this
+pane is holding session `<id>`, and here is the argv that reopens it*.
+
+Nothing here is required for `lop` to run. Publication is best-effort by
+construction: a missing binary, a socket error, a multiplexer that is not
+running, or a timeout is logged at debug and otherwise ignored.
+
+## What gets published
+
+**Always a restore-and-idle command** — `lop --resume <id>`, which replays the
+transcript and waits for you. It never continues an interrupted turn. This is a
+safety property, not a default: a restore happens to every pane at once,
+unattended, usually after a crash you did not choose. The worst case of a
+spurious restore has to be an idle session rather than a dozen agents resuming
+tool execution with nobody watching.
+
+**Never a subagent's session.** A delegated review, design or scout run creates
+a session directory shaped exactly like a real conversation, and it runs inside
+its parent's pane. Publishing one would overwrite the pane's binding, so a crash
+restore would reopen the delegated run instead of your work. Sessions marked
+with `origin.json` are skipped.
+
+**Never before the session can actually be resumed.** `--resume` refuses an id
+whose `transcript.jsonl` does not exist yet, so a session publishes nothing
+until its first turn has persisted, then publishes within a few seconds.
+
+**Nothing sensitive.** The session id and the working directory, and nothing
+else. No environment, no credentials, no prompt text.
+
+## Per multiplexer
+
+| Multiplexer | Mechanism | Auto-restores after a crash |
+|---|---|---|
+| cmux | resume binding via the control socket | yes, with the registration below |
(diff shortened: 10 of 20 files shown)


damianvtran:
#### CI note: the one red `test (3.12)` is a flake, verified not a regression

Two CI runs fired on the **identical commit** `6edcf186c748` and disagreed, which is the cleanest possible evidence here:

| Run | Event | `test (3.12)` |
|---|---|---|
| [32898303070](https://github.com/damianvtran/local-operator/actions/runs/32898303070) | `push` | FAILURE |
| [32898363443](https://github.com/damianvtran/local-operator/actions/runs/32898363443) | `pull_request` | success |

The failure:

```text
FAILED tests/unit/mobile/test_tui_ask.py::test_secret_ask_projects_a_free_text_card_without_leaking_the_value
  AssertionError: no discovery record published
===== 1 failed, 6787 passed, 9 skipped in 499.33s =====
```

Three independent reasons this is not caused by this PR:

1. **Same commit, opposite results.** Identical code and identical job name; only the trigger event differed.
2. **The diff does not touch that area.** `git diff --name-only origin/main..HEAD` matches no `mobile/` or `ask` file. The failing test covers the mobile ask-card surface, which this change does not go near.
3. **Passes locally, repeatedly.** 3/3 consecutive clean runs of `tests/unit/mobile/test_tui_ask.py` on this branch.

`main` itself is green at `f2a52b53`, so this is an intermittent test rather than a broken baseline. It is a pre-existing flake in the mobile suite, unrelated to this change, and it is being recorded here rather than silently re-run.

All other checks pass: `lint`, `type-check`, `pip-audit`, `test (3.13)`, and `test (3.12)` on the pull_request run.

damianvtran:
### Agent review — round 1
- Reviewer: reviewer subagent, lopdev team (anthropic/claude-opus-5)
- Scope: f2a52b53..6edcf186c74859dac69f404bf62d25b7d046e650
- Mode: Comprehensive
- Delivery contract: publish, per pane, "this pane holds session `<id>` and here is the restore-and-idle argv that reopens it" — cmux via an `agent-hook` socket binding, tmux/wezterm/zellij/screen via pane markers — entirely best-effort, never for a subagent, cleared on clean exit.
- Applicable risk surfaces: cmux RPC payload and target resolution; subagent guard; best-effort/non-blocking guarantee; clean-exit vs re-assert race; restore-and-idle boundary; secret/PII leakage; test quality.
- Verdict: Changes requested

| --- | --- | --- | --- |
| F1 | MAJOR | Best-effort: "no failure path may break, block, or slow" | `broadcast.py:279-306` — `stop()` calls `self._backend.retire(...)` synchronously (line 297) and then `thread.join(timeout=CALL_JOIN_TIMEOUT_S)` (line 306), and both callers run on the Textual event loop: `app.py:1978` (`_adopt_session` → `_start_multiplexer_broadcast` → `_stop_multiplexer_broadcast`, line 6858) and `app.py:7204` (`on_unmount`). When the cmux socket is unresponsive, `retire` sits in `subprocess.run(timeout=5.0)` and the join waits up to 6.0s more. Reproduced against a backend whose calls hang for `CALL_TIMEOUT_S`: **`stop() blocked the calling thread for 9.80s`** (worst case ~11s); the same harness on a healthy backend returns in 0.052s. Consequence: a `/new` or `/resume` swap, and quitting, freeze the whole TUI for up to ~11s whenever cmux's socket is mid-restart — which is precisely the post-crash window this feature targets. The docstrings at `broadcast.py:186-188` and `app.py` `_start_multiplexer_broadcast` both assert the event loop never waits on a multiplexer socket; it does. Required: do not call `retire`/`join` from the event loop — hand the retire to the broadcast's own thread (or a Textual `run_worker(..., thread=True)`) and drop the join from the synchronous path, keeping the daemon thread's `_retired` latch as the correctness guarantee it already is. |
| F2 | MAJOR | Best-effort: no unbounded subprocess churn | `broadcast.py:255-277` — the cadence switch is `if published and elapsed < self._interval_s: continue`, and `_published` is only set when `backend.publish` returns **True** (line 250-251). A session that is resumable but whose publish keeps failing (socket refusing, surface closed, non-zero `cmux rpc` exit) therefore never leaves the `_PENDING_POLL_S = 5.0` cadence, and every pass is a full `cmux rpc` subprocess, not the "single `stat`" the comment at `broadcast.py:31-36` claims. Reproduced with a backend returning False: **`publish attempts in ~21s with a 90s re-assert interval: 5` → one cmux subprocess every ~5.0s, indefinitely.** Across the ~15 panes this PR is built for that is ~3 cmux processes/second for the life of the sessions, and it is exactly the state the PR's own `cmux-host-caveat.md` documents (binary present, host not answering). Required: gate the fast cadence on resumability rather than on publish success — poll at 5s only while `is_resumable_session()` is False, and once the session is resumable fall back to `interval_s` whether or not the publish succeeded. |
| F3 | MAJOR | Wrong pane = hijacking another pane's binding | `markers.py:282` — `ZellijBackend.pane_envs = ("ZELLIJ_SESSION_NAME", "ZELLIJ")`. Neither identifies a pane. Verified on zellij 0.42.2 on this host by running a shell inside a real zellij session: it exports `ZELLIJ=0`, `ZELLIJ_PANE_ID=0`, `ZELLIJ_SESSION_NAME=lopreview3` — a genuine per-pane id exists and is not used. Because `_path()` (line 234-235) keys the file as `zellij-<session-name>.json`, every `lop` pane in one zellij session writes and unlinks the *same* file: the last publisher wins, and one pane's clean exit (`retire`, line 264-273) deletes the marker of every other pane in that session. A restore script then reopens the wrong conversation in a pane, or none at all. This also contradicts the stated contract at `markers.py:25-27` ("one per pane ... keyed by pane identity") and the docs table in `docs/multiplexer-resume.md`. Required: use `ZELLIJ_PANE_ID` as the pane identity and key the file on session+pane (e.g. `zellij-<session>-<pane_id>.json`), keeping `_SAFE_PANE_ID` validation on both components. |
| F4 | MINOR | Marker contract accuracy (screen) | `markers.py:292` — `ScreenBackend.pane_envs = ("STY",)`, with the comment at 289-291 acknowledging `WINDOW` "would be a finer identity". Verified on this host (screen 4.00.03): a process inside screen exports both `STY=24291.loprev` and `WINDOW=0`. The collision is the same as F3 — several `lop` windows in one screen session share one marker file — but here it is a stated tradeoff rather than an oversight, so it does not gate. What is inaccurate is the module contract at `markers.py:25-27`, which promises a file "one per pane" for both file-backed backends. Required: either key on `STY`+`WINDOW`, or amend the docstring and the `docs/multiplexer-resume.md` table to say screen markers are per-session and are overwritten by a second `lop` in the same session. |
| F5 | MINOR | Restore-and-idle fidelity of the shell fallback | `cmux.py:159` — `"command": " ".join(binding.argv)`, while the marker backends deliberately use `shlex.join` for the same string with an explicit comment about why (`markers.py:148-151`). `resume_executable()` (`broadcast.py:80-91`) resolves and absolutises `sys.argv[0]`, so a launcher path containing a space produces a `command` string that cmux's `SurfaceResumeCommandCanonicalizer.tokens(from:)` splits into the wrong argv. The structured `launch_command.arguments` (line 164) is what cmux prefers for an agent-hook binding, so this only bites the shell fallback, which is why it is minor rather than a blocker. Required: use `shlex.join(binding.argv)` at `cmux.py:159` so both backends quote the same way. |
| F6 | NIT | Comment accuracy | `types.py:55` — the `name` field comment says "see the sanitising note in ``broadcast``". There is no sanitising note in `broadcast.py` (`grep -rn "sanitis" local_operator/multiplexer/` matches only `types.py:55` and `markers.py:72`). The claim it points at is real but lives elsewhere. Required: point the reference at `markers.py:70-75`, or drop the clause. |

#### Follow-up observations

Things I checked that are **correct** and that I want on the record rather than re-derived next round:

- **The clean-exit vs re-assert race is genuinely closed.** `stop()` sets `_retired = True` inside `self._lock` *before* calling retire (`broadcast.py:292-297`), and `_publish_once` re-reads it under the same lock as its first action (`broadcast.py:237-241`). A re-assert already blocked on the lock cannot resurrect a withdrawn binding, and a re-assert already inside `publish` completes before the retire runs, so the retire is last. The daemon thread also cannot republish after a join timeout, for the same reason. `test_retire_prevents_a_later_republish` pins this correctly.
- **Restore-and-idle holds on every path I could find.** `resume_argv` (`broadcast.py:94-106`) is the sole constructor of the argv, both backends consume it, and nothing interpolates a prompt, `--exec`, or a continue flag. `test_no_flag_can_continue_an_interrupted_turn` pins the absence rather than only the shape, which is the right assertion.
- **The subagent guard is double-gated and I could not get a child past it.** `broadcast_session` refuses on `is_user_owned_session` (`broadcast.py:336`) before any binding is built, and the only caller is the TUI hook, which publishes `app._session` — subagents run in-process under the parent app and never own an `OperatorApp`. Manually resuming a child id also refuses, via the same `origin.json` marker `/resume` uses.
- **Target resolution is sound.** `_surface_target` (`cmux.py:108-114`) requires *both* `CMUX_WORKSPACE_ID` and `CMUX_SURFACE_ID` and UUID-validates each, so an inherited-but-stale value reads as "no cmux" rather than as a malformed request against a live surface. I confirmed against `/tmp/cmuxsrc` that `surfaceResumeExplicitTargetID` prefers `surface_id`, so the pane and not the workspace is what gets bound, and that `surfaceResumeClear` compares `expectedCheckpointID`/`expectedSource` before clearing (`ControlSurfaceResumeTarget.swift:702-706`) — the retire really is scoped to our own binding.
- **The `agent-hook` claim is accurate.** `autoResume: source == "agent-hook" ? (bool(params, "auto_resume") ?? false) : false` is verbatim at `ControlCommandCoordinator+Surface3.swift:85`. The module docstring's warning against "simplifying" this to the CLI is correct and worth keeping.
- **No secret or PII leakage.** The payload carries the session id, the launcher path, the argv, and `cwd`; no `environment` map is sent, and `name` is a constant. `cwd` is a path the user chose and is required for a correct restore.
- **Test quality spot-check.** I mutation-tested the exit path: deleting `self._stop_multiplexer_broadcast()` from `on_unmount` (`app.py:7204`) makes `test_a_session_swap_rebinds_the_pane` fail (`assert 'publish' == 'stop(retire=True)'`), so that assertion is load-bearing rather than incidental. The headless gate has a real paired control. Both new files pass on this worktree (40 passed / 5 passed).
- **On the disclosed limitation:** the `cmux-host-caveat.md` disclosure reads as accurate and complete to me, and I did not treat incomplete end-to-end evidence as a finding. Worth noting that F2 makes the un-revalidated state *more* costly than the caveat implies — an unresponsive cmux host is the condition that produces the 5s subprocess loop — so fixing F2 before the re-validation restart is the cheaper order.

None of F1–F3 is a design disagreement; each is a concrete behaviour I reproduced on this branch. F1 and F2 are both fixes inside `SessionBroadcast` and F3 is a two-line identity change, so I would expect one remediation round to clear all three.

damianvtran:

- Implementer: coder subagent, lopdev team (anthropic/claude-opus-5)
- Scope: 6edcf186c74859dac69f404bf62d25b7d046e650..f0b2fef1123f90cdd51634b8772df78c65dd6c82
- All six findings actioned in one commit `f0b2fef1`, since the head moves anyway and a separate polish push would cost a round for nothing.

Every finding was judged on its merits and all three majors reproduced on this branch before being changed — the reviewer's own harnesses, re-run below with the new numbers. Nothing is rejected.

- **F1 (MAJOR) — Fixed in `f0b2fef1`**: `stop()` no longer calls `retire` or joins on the caller's thread. The retire is dispatched to a short-lived daemon worker and `stop()` returns immediately; the join moved to an explicit `join()` used only by tests. Re-running the reviewer's harness (backend hanging for `CALL_TIMEOUT_S`, publish allowed to get *into* the wedged call first): **9.51s → 0.01s** on the wedged backend, 0.02s → 0.00s healthy. The withdrawal still happens and is observed completing once the backend is released — what no longer happens is the event loop waiting for it. Correctness is unmoved: `_retired` is latched **synchronously inside `stop()`** before the worker is dispatched, so a queued re-assert cannot resurrect a withdrawn binding whether or not the withdrawal itself lands. Two supporting changes fell out of this: the latch became an `Event` so it can be set without contending with a retire parked in a subprocess timeout, and `_publish_once` now tests it *before* taking the call lock so a re-assert arriving mid-withdrawal bails instead of queueing behind a 5s call. The docstrings at `broadcast.py` and `app.py` `_start_multiplexer_broadcast` are now true rather than softened.
- **F2 (MAJOR) — Fixed in `f0b2fef1`**: the cadence keys on `is_resumable_session()` instead of on publish success, so a resumable session whose publish keeps failing drops to `interval_s` rather than retrying forever at 5s. Reviewer's harness (publish always False, session resumable, 90s interval, ~21s window): **5 attempts at [5.01, 5.0, 5.0, 5.01]s gaps → 1 attempt** (the immediate publish at start). The `_PENDING_POLL_S` comment no longer claims "a single stat" about the failing-publish case; it now describes the state it is actually true of, since a non-resumable pass returns before reaching the backend. The fast cadence still exists and is pinned by a paired test showing a cold session publishing within one poll of its transcript appearing — the back-off does not cost a cold session its responsiveness.
- **F3 (MAJOR) — Fixed in `f0b2fef1`**: `ZellijBackend` keys on session **and** `ZELLIJ_PANE_ID`. Confirmed on real zellij 0.42.2 that two panes of one session export the same `ZELLIJ_SESSION_NAME` with `ZELLIJ_PANE_ID=0` and `=1`. Proven end to end with the real registry and backend in two real zellij panes, not mocks: **before, two successful publishes produced one file** (`zellij-f3old.json` holding only pane 1's session — pane 0's binding destroyed); **after, two files**, `zellij-f3c-0.json → aaaaaaaaaaaa` and `zellij-f3c-1.json → bbbbbbbbbbbb`. The sibling-deletion half is pinned by a test that retires pane 0 and asserts pane 1's marker survives.
- **F4 (MINOR) — Fixed in `f0b2fef1`**: took the stronger of the two options offered. `ScreenBackend` keys on `STY`+`WINDOW` (verified on screen 4.00.03 that a process inside a session exports `STY=42653.loprev` and `WINDOW=0`), with `STY` alone kept as a fallback because `WINDOW` comes from shell startup rather than from screen in every configuration. The module contract at `markers.py` and the `docs/multiplexer-resume.md` table now both state the composite key and say explicitly that the fallback marker is per session and is overwritten by a second `lop` in that session, so the docstring is accurate either way.
- **F5 (MINOR) — Fixed in `f0b2fef1`**: `cmux.py` uses `shlex.join(binding.argv)`, matching the marker backends. Pinned by a test that publishes from `/Applications/My Tools/lop` and asserts `shlex.split` of the result round-trips to the original argv, so the assertion is the property rather than the quoting style.
- **F6 (NIT) — Fixed in `f0b2fef1`**: the `name` comment in `types.py` points at `_SAFE_PANE_ID` in `markers`, where the sanitising note actually lives.

#### Tests

Eight new/changed tests. All eight were run against unmodified `6edcf186` with only the test file copied in, and **all eight failed** — including the two the review asked for specifically:

```text
test_stop_does_not_block_the_caller_on_a_wedged_backend
  AssertionError: stop() blocked the caller for 59.82s
test_a_failing_publish_backs_off_to_the_reassert_cadence
  AssertionError: failing publish retried 3x on the fast poll
test_two_zellij_panes_in_one_session_do_not_share_a_marker
  + 'zellij-main.json'  - 'zellij-main-1.json'
test_the_retire_latch_is_set_before_stop_returns
  AssertionError: assert ['retire'] == []
```

The F2 test was additionally re-run pre-fix with the (not-yet-existing) `join()` helper neutralised, to confirm it fails on its own cadence assertion rather than on a missing attribute: `assert 3 == 1`, attempts 5.00s apart.

#### Gates

`flake8` clean · `black --check` 26.1.0 clean (499 files) · `isort --check-only --profile black` clean · `pyright` 0 errors, 0 warnings · unit suite **6797 passed**.

Multiplexer suites run 3x consecutively with no flake: 55 passed each time.

Three failures in the full run are pre-existing and unrelated to this diff, each reproduced on unmodified `6edcf186` in a detached worktree:

- `test_app_pilot.py::test_esc_aborts_a_running_shell_command` — fails on the unmodified head too.
- `test_comms.py::test_a_resumed_child_replays_the_stopped_ones_transcript` and `::test_a_cancelled_child_keeps_the_work_it_had_already_done` — timing-sensitive; two clean runs then a failure on the unmodified head.

Note this is a *different* pre-existing flake from the `test_tui_ask.py` one recorded earlier in this PR.

Full evidence, including the harnesses and the raw before/after output, is committed at `docs/evidence/multiplexer-resume/round-1-remediation.txt`.

Ready for round 2 against `f0b2fef1`.

damianvtran:
### Agent review — round 2

- Reviewer: reviewer subagent, lopdev team (anthropic/claude-opus-5)
- Scope: f2a52b53..f0b2fef1123f90cdd51634b8772df78c65dd6c82
- Delivery contract: publish, per pane, "this pane holds session `<id>` and here is the restore-and-idle argv that reopens it" — best-effort, never for a subagent, cleared on clean exit.
- Delta reviewed: 6edcf186..f0b2fef1
- Applicable risk surfaces: F1 asynchronous-withdrawal correctness (retire outliving the app, retire racing a publish, latch adequacy); F2 cadence; F3/F4 pane identity and fallback collision; F5 quoting; regression check on the round-1 properties that held; new-test quality.
- Verdict: Changes requested

All six round-1 findings are genuinely remediated, and I verified each rather than taking the claim on trust. F1's remediation, however, introduces two new defects in the delta itself. Both are consequences of the same design choice — the withdrawal is now fire-and-forget with nothing sequencing it — and neither is hypothetical: I reproduced both against the real `ZellijBackend` on this branch, and confirmed the first does not reproduce on `6edcf186`.

| --- | --- | --- | --- |
| F7 | BLOCKER | Swap ordering: "retire the outgoing binding BEFORE publishing the new one" | **New in this delta.** `broadcast.py:354-359` latches and dispatches the retire to a worker, and `stop()` returns without sequencing it. `app.py:6858-6860` still relies on the retire having *completed* before it publishes: "Retire the outgoing binding BEFORE publishing the new one: both name the same pane, so the reverse order would clear the fresh binding." With the retire now asynchronous, that ordering is no longer enforced — the outgoing worker's `retire` lands *after* `_start_multiplexer_broadcast` has published the incoming binding, and because the marker/option backends key retire on **pane identity alone** (`markers.py:281-291` `_FileBackend.retire` unlinks `self._path(pane)`; `markers.py:163-172` `_OptionBackend.retire` unsets the pane's options) the withdrawal deletes the *new* session's binding. Reproduced with the real `ZellijBackend` and a `/new` swap: `bound to A: ['zellij-main-0.json']` → `after swap, new binding published: zellij-main-0.json → bbbbbbbbbbbb` → `after the outgoing retire lands: NO MARKER — pane advertises nothing`. Threshold measured at ~5ms of retire latency: `0.0ms → 0/20 swaps lost`, `5.0ms → 20/20`, `50ms → 20/20`. Every real backend retire is well past that (one subprocess spawn on this host: `screen -V` median 6.7ms, `tmux -V` 14.0ms, `zellij -V` 18.9ms, `cmux -V` 73.5ms), so this is the normal case, not a tail. Same harness against pre-remediation `6edcf186`: `0/10` lost; against `f0b2fef1`: `10/10`. Consequence: after every `/new` or `/resume`, the pane advertises **nothing**, so a crash loses the session — the exact failure this PR exists to prevent, now on the most common path. cmux itself is exempt (`cmux.py:186-198` sends `checkpoint_id`/`source` as expectations so a mismatched clear is refused) which is likely why this was not caught, but tmux, wezterm, zellij and screen are all affected. Required: make the swap ordered again — have `_adopt_session` await the outgoing withdrawal off the event loop before publishing the new binding (e.g. `run_worker(..., thread=True)` that calls `stop()` then `join()` then starts the new broadcast), or scope the marker/option retire to the binding that published it (compare the stored `session_id` and skip the unlink/unset when it does not match, mirroring what `cmux.py` already does). The second is the smaller change and fixes the class rather than the instance. |
| F8 | MAJOR | Clean exit: "a session the user deliberately quit must not be replayed into their next shell" | **New in this delta.** `broadcast.py:360-366` deliberately does not join, and `_dispatch_retire` (`broadcast.py:368-398`) starts a **daemon** thread. On quit, `on_unmount` (`app.py:7206`) calls `_stop_multiplexer_broadcast` → `retire_session` → `stop(retire=True)`, which returns in ~0.3ms having only *started* the worker; the interpreter then exits and daemon threads are killed without running. Reproduced end to end: with the real `ZellijBackend` and a retire costing one subprocess spawn, the binding **survives** the quit when teardown after `stop()` is 0ms or 5ms, and only clears at ≥20ms. Reproduced separately at process level — `stop() returned in 0.000s`, `retire landed before exit: False`, and the marker file was still absent 6s after the process had gone, i.e. the withdrawal never ran at all. The docstring at `broadcast.py:373-378` acknowledges this ("a withdrawal that loses a race with process exit costs a stale marker, which is the failure mode this package is explicitly allowed to have") but that is not the same failure: a stale marker after a *crash* is the feature, whereas a stale marker after a *clean quit* means the user's next shell in that pane replays a conversation they deliberately closed, which `_stop_multiplexer_broadcast`'s own docstring (`app.py:6888-6895`) calls out as the thing the default must prevent. The latch is not a substitute here — it protects against re-publication *within* the process, and nothing about it removes the already-published binding. Note the window is wider when a re-assert is in flight: measured withdrawal completion at 5048ms behind a publish parked in a `CALL_TIMEOUT_S` subprocess, versus 26ms idle, because the worker waits on `_call_lock` (`broadcast.py:383-389`). Required: give the quit path a bounded wait that is not on the event loop — the cleanest fix given F7 is the same one: `on_unmount` runs the stop-then-`join(timeout=…)` in a thread worker it awaits, so the withdrawal is ordered and bounded rather than raced against interpreter shutdown. Do not simply re-add the synchronous join; that reinstates F1. |

#### Round-1 findings: verified remediated

Each checked against the delta, not against the claim.

- **F1 (MAJOR) — remediated, with the caveats above.** `stop()` no longer calls `retire` or joins: measured **0.3ms** against a backend parked in a 5s call, and I confirmed the latch is genuinely set synchronously at `broadcast.py:358` before `_dispatch_retire()` at line 359. The resurrection window really is closed — `_publish_once` (`broadcast.py:264-272`) tests `_retired` *before* taking `_call_lock` and again under it, and `test_the_retire_latch_is_set_before_stop_returns` fails on unmodified `6edcf186` with `assert ['retire'] == []`. The `Event`-instead-of-bool change is right: `stop()` never touches `_call_lock`, which is what keeps the wedged socket off the caller. `_dispatch_retire` is idempotent under `_dispatch_lock`, so repeated `stop()` calls cannot leak workers, and the worker is short-lived and daemon, so nothing accumulates or holds the process open. What the fix did not preserve is the *ordering* and *completion* the two callers depended on — F7 and F8.
- **F2 (MAJOR) — remediated, fully.** Cadence now keys on `_resumable` (`broadcast.py:283-288`, `broadcast.py:304-315`), set before the publish and independently of its result. Verified all three directions: resumable + always-failing publish over a 16s window with a 90s interval → **1 backend call** (was 5); a never-resumable session over 11s → **0 backend calls**, so the `_PENDING_POLL_S` comment's "really is a single `stat`" claim is now true — `_publish_once` returns at `broadcast.py:279-281` before reaching the backend; and the cold-session path is not sacrificed — a transcript landing mid-flight published **4.69s later**, within one poll. The corrected comment at `broadcast.py:31-40` matches the behaviour.
- **F3 (MAJOR) — remediated.** `pane_envs = (("ZELLIJ_SESSION_NAME", "ZELLIJ_PANE_ID"),)` (`markers.py:305`) with no coarser fallback, so a zellij lacking the pane id publishes nothing rather than colliding — I confirmed `detect()` returns False and `publish()` returns False in that case. Two panes now produce two files and pane 0's retire leaves pane 1's marker intact.
- **F4 (MINOR) — remediated, and the fallback does not silently collide in a way the docs misdescribe.** I specifically probed the mixed case the brief asked about: one window exporting `WINDOW` and another in the *same* `STY` not exporting it produce `screen-42653.loprev-0.json` and `screen-42653.loprev.json` — two distinct files, and the fallback pane's retire does not touch the composite pane's marker. So the fallback degrades to per-session (as documented at `markers.py:30-33` and in the docs table) without cross-deleting a finer-keyed sibling. `_pane_id` (`markers.py:237-246`) validates each component separately, and I confirmed the stated reason is real: `{"ZELLIJ_SESSION_NAME": "main-0", "ZELLIJ_PANE_ID": "x"}` yields `main-0-x`, not a forgery of `main-0`. A malformed `WINDOW` (`../../etc/x`) correctly falls through to the `STY` candidate rather than being sanitised into a wrong filename.
- **F5 (MINOR) — remediated.** `shlex.join(binding.argv)` at `cmux.py:160`, and the paired test asserts the round-trip property rather than the quoting style.
- **F6 (NIT) — remediated.** `types.py:55-57` now points at `_SAFE_PANE_ID` in `markers`, where the note actually lives (`markers.py:78-82`).

#### Round-1 properties: no regression in the delta

- **Subagent guard** — untouched by this delta; `broadcast_session` still refuses on `is_user_owned_session` before any binding is built.
- **Restore-and-idle** — `resume_argv` untouched; nothing in the delta interpolates a prompt or a continue flag.
- **Surface targeting** — `_surface_target` untouched; `cmux.py`'s only change is the quoting at line 160.
- **No secrets in the payload** — the delta adds no field to either payload; `markers.py` publish payload and the cmux params are unchanged in shape.
- **Test quality** — I re-ran the pre-fix claim rather than assuming it. Copying the new test file onto unmodified `6edcf186` gives **6 failed** on the ones I selected, with failures that are the property under test and not import errors: `stop() blocked the caller for 59.81s`, `assert ['retire'] == []`, and the zellij marker-set diff. One exception worth recording: `test_a_failing_publish_backs_off_to_the_reassert_cadence` fails pre-fix with `AttributeError: 'SessionBroadcast' object has no attribute 'join'` rather than on its cadence assertion, so on its own it does not prove the F2 behaviour — the remediation comment already discloses this and reports re-running it with `join()` neutralised to get `assert 3 == 1`, which is the right control. The multiplexer suites pass 55/55 on this worktree.
- Neither F7 nor F8 is covered by a test. `test_a_session_swap_rebinds_the_pane` (`test_multiplexer_broadcast.py:105`) pins the *call order* against a stub `Handle` whose `stop` only appends a string, so it cannot observe that the real retire lands after the new publish. A regression test for F7 needs a real (or realistically-slow) backend rather than a stub.

#### Follow-up observations

- Not filed as a finding, per the brief: the disclosed lack of end-to-end verification on a healthy cmux host. Worth noting that F7 does not depend on that gap — it is backend-agnostic in the marker/option path and reproduces deterministically without cmux.
- `_timer_thread` (`broadcast.py:224-226`) is retained solely so `join()` has something to wait on and is documented as never read by a production path. That is accurate as written, but if F7/F8 are fixed by having the app await the withdrawal, `join()` stops being test-only and the comment at `broadcast.py:400-405` will need updating to stay true.

Two blocking items, both in the F1 remediation and both with the same root cause. Round 2 is not terminal.

damianvtran:

- **F7 (blocker) — Fixed in `c173bfee`:** the retire is now scoped to the binding that published it, *and* the swap is sequenced — both halves, because neither alone covers every backend.
  - **Scoping (removes the ordering dependency for tmux/zellij/screen):** `_FileBackend.retire` reads the marker back and unlinks only when `backend`/`pane`/`session_id` all match its own binding (`markers.py`, `_marker_is_ours`); unreadable-but-present reads as *not ours*, because the only decision this feeds is a DELETE. `TmuxBackend` gains `scoped_retire = True` with a `_read_option` readback via `tmux display-message -p` — chosen over `show-option` because an *absent* user option makes `show-option` exit non-zero (indistinguishable from "tmux would not answer"), while `display-message` renders unset as empty with exit 0 (verified live on tmux 3.7c). This mirrors the `checkpoint_id` expectation cmux already sends, as suggested.
  - **Sequencing (covers wezterm + cmux, makes healthy-case order deterministic):** `broadcast_session(..., predecessor=outgoing)`; the successor's own timer thread drains the predecessor (`join(SWAP_DRAIN_TIMEOUT_S = 6.0s)`) before its first publish. The wait runs on a thread the event loop is not waiting on, so F1's guarantee is untouched — `app.py` still calls `stop()` (prompt) and `start()`, nothing more.
  - **Why both:** wezterm user vars are **write-only** from a pane process (upstream: only Lua `pane:get_user_vars()` can read them; `cli list --format json` exposes no user vars), so the per-backend session-id comparison is *unimplementable* there — sequencing is wezterm's only protection. With sequencing disabled and scoping in place, the harness loses **6/6** wezterm swaps; with both, **0/10**. Conversely, scoping alone already fixes zellij at every latency (0/20 at 50ms), which is what makes the race *benign* rather than merely narrower for the three backends that can scope.
  - **Measured (reviewer's harness, real `ZellijBackend` + 50ms retire):** `f0b2fef1` → **20/20 lost** (5ms: 18/20); `c173bfee` → **0/20 lost** at 0.0ms, 5.0ms and 50.0ms. App-path variant (`stop()` then `broadcast_session(predecessor=)`) across all four marker backends at 50ms: zellij/screen/tmux/wezterm all **0/10 lost** (pre-fix call shape: **10/10 each**). At 500ms retire latency: **0/10** on all four.
- **F8 (major) — Fixed in `c173bfee`:** a one-shot `atexit` drain (`_register_exit_drain` / `_drain_retires_at_exit`, `EXIT_DRAIN_TIMEOUT_S = 2.0`) joins the retire worker at interpreter exit. **Worst-case quit delay, stated explicitly:** a healthy retire adds only its own subprocess cost (quit wall ≈ baseline + ~50ms — measured 338ms vs 97ms interpreter baseline); a retire queued behind an in-flight publish or a wedged socket adds **at most 2.0s**, never more, because the drain's budget is a deadline shared across broadcasts. The wait runs on the main thread *after* `on_unmount` has returned and the event loop is gone — no coroutine can observe it. `os._exit` (the Windows re-exec path) and a hard crash skip `atexit` entirely, which is correct: both are crash-shaped exits where a surviving binding is the feature. **Measured:** reviewer's process-level harness — pre-fix, binding survived quit and was still advertised 6s after the process was gone; post-fix, **withdrawal lands** at 0ms/5ms/20ms teardown-after-stop, verified 1s after process exit.
- **Test coverage (both defects):** `TestSwapOrdering` (6 tests) drives the real `ZellijBackend`/`WezTermBackend` with 50ms retire latency — one subprocess spawn, the floor for every real backend — through `app.py`'s exact swap shape; plus scoped/unscoped-retire unit properties, the unreadable-marker refusal, and the tmux readback. `test_the_withdrawal_lands_before_interpreter_exit` runs the quit in a real subprocess (the property *is* process death) and asserts the marker is gone after it; `test_a_wedged_retire_bounds_the_quit_delay` pins the 2s bound. **All 7 fail on `f0b2fef1`** (verified by copying the test file onto the unmodified pre-fix head: 7 failed / 51 passed), and pass on `c173bfee`. The existing `test_tmux_unsets_rather_than_blanking_on_retire` now stubs the readback it implicitly gained.

**F1–F6 remain fixed.** F1 non-regression re-measured on `c173bfee` against a backend parked in a 5s call with a publish in flight: **`stop()` returned in 0.27ms** (was 9.8s pre-F1; bound asserted at <1s in the suite). F2's cadence, F3's zellij keying, F4's screen identity, F5's `shlex.join`, and F6's docstring pointer are untouched by this delta. The stale `_timer_thread` "never read by a production path" comment the reviewer flagged is corrected — `join()` now has a production caller (the swap drain).

**Gates:** flake8 clean, black 26.1.0 clean, isort clean, pyright 0 errors, `tests/unit` 6806 passed / 7 skipped with 2 failures in `tests/unit/harness/test_comms.py` that reproduce with this change stashed on clean `f0b2fef1` under machine load (another agent's suite was running concurrently) and pass 3/3 in isolation — pre-existing flake, recorded rather than re-run away. The three `test_app_pilot.py` failures seen in one loaded run are the identical set as on `f0b2fef1` and pass in isolation.

damianvtran:
### Agent review — round 3

- Reviewer: reviewer subagent, lopdev team (anthropic/claude-opus-5)
- Scope: f2a52b53..c173bfee46964f3b33ba7086294cbb2028e2b150
- Delivery contract: publish, per pane, "this pane holds session `<id>` and here is the restore-and-idle argv that reopens it" — best-effort, never for a subagent, cleared on clean exit, and never waited on from the event loop.
- Delta reviewed: f0b2fef1..c173bfee
- Applicable risk surfaces: F7 swap-ordering remediation (scoping + sequencing, per backend); F8 exit-drain remediation at process level; bounds and deadlock-freedom of the two new waits; TOCTOU in the scoped unlink; tmux readback semantics; `atexit` vs the re-exec path; F1 non-regression on the event loop; the pre-fix test claim.
- Verdict: Changes requested

**F7 and F8 are genuinely remediated.** I reproduced both fixes rather than accepting the claims, and both hold at the numbers stated. What blocks is not either original defect but a bound in the F7 remediation itself: `SWAP_DRAIN_TIMEOUT_S` is documented and reasoned about as the bound on the successor's first publish, and the code as written can take **twice** that, compounding further across a swap chain.

#### F7 / F8: verified remediated

Checked against the delta, not the claim.

- **F7 (blocker) — fixed.** Scoping and sequencing both do what they say. Real `ZellijBackend` + 50ms retire through `app.py`'s exact swap shape (`stop()` then `broadcast_session(predecessor=)`): **0/20 swaps lost** (round 2 measured 20/20 on `f0b2fef1`). The wezterm leg, which has no scoping at all and rests entirely on sequencing, is **0/10** against a call-shape-faithful backend with 50ms latency — I tested it separately precisely because it is the leg with only one protection. `_marker_is_ours` (`markers.py:397-418`) correctly treats unreadable-but-present as *not ours*, which is the right polarity for a decision that only ever feeds a DELETE.
- **F8 (major) — fixed, at process level.** Child-process harness, marker checked a full second *after* the process was gone: withdrawal lands at 0ms/5ms/20ms teardown-after-stop, quit wall 0.31-0.39s. The `atexit` bound genuinely holds against a backend that ignores `CALL_TIMEOUT_S` entirely: **2.00s** with one wedged broadcast and **2.00s** with five, so the deadline really is shared across broadcasts rather than per-broadcast. `atexit._ncallbacks()` is 2 in both cases, so the one-shot registration is not double-registering.
- **tmux readback — the stated reason is real.** Verified live on tmux 3.7c: `display-message -p -t <pane> '#{@lop_session}'` renders an unset option as empty with **exit 0**, while `show-option -p` on the same unset option exits **1**. So `display-message` does distinguish "unset" from "tmux would not answer", and the `_capture` contract (None on non-zero exit ⇒ do not clear) is sound. It runs only inside `_OptionBackend.retire`, which is only ever reached from the retire worker thread — no event-loop caller.
- **F1 — no regression.** On a backend parked in a wedged call: `stop()` **0.44ms**, and `start()` with a wedged predecessor **0.24ms**. The predecessor drain is genuinely on the successor's timer thread, so the new wait is not reachable from the loop. Gates on the touched tree: flake8 clean, pyright **0 errors**, `tests/unit/test_multiplexer.py` + `tests/unit/tui/test_multiplexer_broadcast.py` **63 passed**. All 7 CI checks green on head.
- **Pre-fix test claim — confirmed exactly.** Copied the new `test_multiplexer.py` onto a detached `f0b2fef1` worktree: **7 failed, 51 passed**, and the failures are the properties under test (`the withdrawal never ran: a cleanly-quit session is still advertised in its pane`, `assert True is False` on the foreign-marker retire, `no attribute '_capture'`, `unexpected keyword argument 'predecessor'`), not collection errors.

| --- | --- | --- | --- |
| F9 | MAJOR | Best-effort: the swap drain must be bounded by the number it documents | **New in this delta.** `broadcast.py:356` drains the predecessor with `predecessor.join(timeout=SWAP_DRAIN_TIMEOUT_S)`, but `join` (`broadcast.py:468-480`) applies that timeout **per worker**, looping `for worker in (retire_worker, self._timer_thread)`. Two workers, one 6.0s timeout each, so the real bound on the successor's first publish is **12s, not 6s**. Measured against a backend whose calls never return: successor's first publish delayed **12.01s** (`join(timeout=6)` → 12.01s; `_drain_retire(timeout=6)` → 6.00s on the identical scenario). It compounds across a chain: A→B→C with A and B both wedged delayed C's first publish **11.80s**, because C joins B's *timer*, which is itself still inside its own drain of A. The comment block at `broadcast.py:57-69` and the `_run` comment at `broadcast.py:344-348` both reason explicitly about `SWAP_DRAIN_TIMEOUT_S` as *the* bound ("the successor has to publish eventually"), so this is the code disagreeing with its own stated contract, not an undocumented tail. Consequence: on a wedged or mid-restart socket — the exact post-crash window this feature targets — the pane advertises the OLD session for up to 12s after a `/new`, so a crash inside that window reopens the conversation the user just left. I confirmed the pane holds the stale binding for the whole window rather than nothing, so this loses the swap rather than the pane. Not user-visible as a freeze: the loop is not waiting on it. Required: drain only the ordering-relevant worker — call `predecessor._drain_retire(timeout=SWAP_DRAIN_TIMEOUT_S)` at `broadcast.py:356` instead of `join(...)`. The timer thread carries no ordering (it is `_stopped`, and `_publish_once` refuses under the `_retired` latch), which is the reasoning `_drain_retire`'s own docstring at `broadcast.py:482-489` already gives for the exit path; the swap path wants the same narrowing. That restores the documented 6s and removes the chain compounding in one line. |

#### Late discovery

| --- | --- | --- | --- |
| F10 | MINOR | Docstring accuracy on the re-exec path | **Late discovery** — the prose is new in this delta (`broadcast.py:590-592`) but the code it describes (`reexec.py`) is unchanged since round 1, so it is flagged as accuracy rather than reopened. The docstring says "`os._exit` (the Windows re-exec path) and a hard crash skip `atexit` entirely, which is correct". The POSIX re-exec path also skips `atexit`: `reexec.py:152` uses `os.execvpe`, and exec replaces the process image without running `atexit` handlers — I verified this directly (handler did not fire across an `execvpe`). So the exemption is not Windows-only. **Behaviourally this is correct and needs no code change**: `_request_relaunch` (`app.py:3183-3211`) exits via Textual, `on_unmount` calls `_stop_multiplexer_broadcast()` → `retire_session`, and `cli.py:2414-2418` only calls `replace_self` after `asyncio.run` has returned — so the withdrawal's worker is dispatched before the exec, and a relaunch wants the binding withdrawn anyway since the new process republishes. The defect is only that the comment names Windows as the exec-shaped exit when POSIX is too, which will mislead the next reader reasoning about drain coverage. Required: amend to say every exec/`os._exit` path skips `atexit`, POSIX included, and state why that is harmless here (the retire is dispatched before `replace_self`, and the successor process republishes). |

#### Follow-up observations

Non-gating; no round needed for these.

- `broadcast.py:250,359` — `self._drained` is created and `set()` but never waited on or read anywhere in `local_operator/` or `tests/`. Dead state; either assert on it in the swap tests or drop it.
- `markers.py:332` — `_MARKER_FIELDS` is defined with a substantial explanatory comment but never referenced; `_marker_is_ours` (`markers.py:409-418`) inlines the same three field names literally. Two places to update if the payload gains a field. Either use the constant or delete it.
- TOCTOU in the scoped unlink is real but benign, and I want it on the record rather than filed: with the window forced maximally wide (successor publishes *between* `_marker_is_ours` returning True at `markers.py:404` and `unlink` at `markers.py:407`), the successor's marker **is** deleted. It does not gate because the sequencing closes it — the successor cannot publish until the predecessor's retire worker has finished, which is the whole point of F7's second half — so the window is only reachable if F9's drain times out first. Worth a comment at `markers.py:404` noting that the payload comparison is not itself atomic and that the sequencing is what makes it safe, so nobody later removes the sequencing on the belief that scoping alone suffices. A genuinely atomic version would need an O_EXCL rename dance, which is not worth it here.
- No test pins either drain bound. `test_a_wedged_retire_bounds_the_quit_delay` covers the 2s exit bound; nothing asserts the swap drain's, which is why F9's 2x survived to review. A test asserting the successor publishes within `SWAP_DRAIN_TIMEOUT_S` against a wedged predecessor would have caught it and would guard the fix.
- Not filed, per the brief: the disclosed lack of end-to-end verification on a healthy cmux host.
- Housekeeping, not a review finding: the PR is `CONFLICTING` against `origin/main`. The only conflicted path is `pyproject.toml` — the branch bumps 0.33.1 → **0.34.0** while `origin/main` is already at **0.36.0**. `local_operator/tui/app.py` auto-merges. Needs a rebase and a re-bump to 0.37.0 (feature) before merge, and CI green is on the pre-rebase head.

**Verdict: Changes requested** — one MAJOR (F9), in the F7 remediation's bound rather than in F7 or F8 themselves. F7 and F8 are fixed and I verified both independently. F9's fix is one line, and round 4 should be a short confirmation round: re-measure the successor's first publish against a wedged predecessor and confirm it lands within `SWAP_DRAIN_TIMEOUT_S`, plus the A→B→C chain.

damianvtran:

- Implementer: coder subagent, lopdev team (anthropic/claude-opus-5)
- Scope: `c173bfee`..`efade76e` (branch rebased; see "Rebase" below)
- Both findings actioned in `06c1a23f`.

- **F9 (MAJOR) — Fixed in `06c1a23f`**: the swap drain now uses `_drain_retire`, so the documented `SWAP_DRAIN_TIMEOUT_S` is the actual bound rather than a per-worker timeout applied across two workers. The reviewer's diagnosis was exactly right, including the compounding across a chain. Measured on the reviewer's own scenarios: single swap **12.02s → 6.00s**; chained A→B→C with two wedged predecessors **11.70s → 6.01s**, so the chain no longer accumulates. Two new regression tests pin the bound and both fail pre-fix at the reviewer's original figures (12.02s / 11.81s). Worth recording for context: this is the wedged-socket worst case only. Real swap latency on a healthy backend measures **68ms**, and the pane advertises the outgoing session (not nothing) for the duration of the drain, so the user-visible failure mode is a stale binding for at most the bound, not an unadvertised pane.
- **F10 (MINOR, late discovery) — Fixed in `06c1a23f`**: the docstring claimed only `os._exit` "(the Windows re-exec path)" skips `atexit`. Verified directly that POSIX `os.execvpe` at `reexec.py:152` skips it too, and corrected the statement to say so. Behaviour is unchanged and was already correct: the retire is dispatched before `replace_self`, so the withdrawal is already in flight when the exec replaces the process.

#### F1–F8 re-verified after the rebase

Re-ran the harnesses against the new base rather than assuming the rebase was inert:

- **F7**: 0/20 swaps lost at 0ms, 5ms and 50ms retire latency; 0/10 across all four marker backends.
- **F8**: withdrawal lands at process level, confirmed 1s after process exit.
- **F1**: `stop()` returns in **0.27ms** on a wedged backend with a publish in flight. The event loop is still never waited on.

#### Rebase

The branch was rebased onto `origin/main` (`04702e6e`, 0.37.0) and the version re-bumped to **0.38.0**. Only `pyproject.toml` conflicted; `app.py` auto-merged. Both of the intervening changes were read specifically because they sit near this code — `perf(tui): keep the event loop responsive…` (#300) and `feat(resume): restore the attached team, agent and goal on resume` (#301) — and neither touches the multiplexer path, `_adopt_session`'s swap ordering, or `on_unmount`. The PR is no longer `CONFLICTING`.

#### Gates

`flake8` clean · `black --check` 26.1.0 clean (539 files) · `isort` clean · unit suite **6990 passed, 7 skipped, 0 failures** on the final head.

On type-checking, stated precisely rather than overclaimed: local full-repo `pyright` was **inconclusive** — it was killed by timeout three times while this machine carried a load average above 170 from concurrent agent sessions, producing no output rather than a clean result. It was verified **scoped to every changed file** (`local_operator/multiplexer/`, `tests/unit/test_multiplexer.py`, `tests/unit/tui/test_multiplexer_broadcast.py`): `0 errors, 0 warnings, 0 informations`. CI's `type-check` job runs full-repo on an unloaded runner and is the authoritative result; it is green on this head.

#### CI

`build`, `lint`, `type-check`, `pip-audit`, `test (3.13)` all green. One `test (3.12)` run reports `tests/unit/mobile/test_tui_bridge.py::test_tui_auto_registers_and_answers_control` — "no discovery record published" — and it is a flake, on the same evidence pattern recorded earlier in this PR: the **identical commit** `efade76e` passed `test (3.12)` in the parallel `pull_request` run while failing in the `push` run, the diff touches no `mobile/` file, and the test passes 3/3 locally on this branch. This is the second distinct pre-existing flake observed in the mobile suite during this PR (the first was `test_tui_ask.py`).

damianvtran:
### Agent review — round 4

- Reviewer: reviewer subagent, lopdev team (anthropic/claude-opus-5)
- Scope: `f2a52b53`..`efade76e` (effective base after the rebase: `04702e6e`)
- Delivery contract: publish, per pane, "this pane holds session `<id>` and here is the restore-and-idle argv that reopens it" — best-effort, never for a subagent, cleared on clean exit.
- Delta reviewed: `c173bfee`..`efade76e` (substantively `06c1a23f` + the rebase onto `04702e6e` + the 0.38.0 bump)
- Applicable risk surfaces: the F9 drain bound (single and chained); drain-window pane state; F10 prose accuracy; rebase fidelity against #300/#301; F1/F7 non-regression on the new base
- Verdict: **Approve**

**F9 and F10 are genuinely fixed, and the rebase is inert.** I reproduced both fixes with my own harness rather than re-running the implementer's tests, and the rebase claim I was asked to distrust holds under a byte-level check. One new MINOR accuracy defect, of exactly the same class as F10 and created by the F9 fix itself: two comments still describe `join()` as the swap's drain when the swap no longer calls it. It does not gate.

| --- | --- | --- | --- |
| F11 | MINOR | Comment accuracy on what the swap now drains | `broadcast.py:482` — `join`'s docstring still reads "Called by tests, by teardown diagnostics, **and by a SWAP**: the successor broadcast's timer thread joins its predecessor before publishing, so the two bindings cannot race for the pane." `broadcast.py:47` repeats it: "Default grace for `SessionBroadcast.join`, **which a SWAP** (the successor broadcast waiting out its predecessor's withdrawal) and the tests both use." Both were true at `c173bfee` and were made false by `06c1a23f`, which moved the swap onto `_drain_retire`. I confirmed no production swap caller remains: `grep -rn "predecessor.join\|join(timeout=SWAP_DRAIN_TIMEOUT_S)" local_operator/` returns nothing, and the only `join()` callers left in the tree are `tests/unit/test_multiplexer.py`. Consequence: this is precisely the F10 failure mode reintroduced one commit later — the next reader reasoning about which worker the swap waits on gets a definition that names the wrong method, and `CALL_JOIN_TIMEOUT_S`'s own doc points at the removed design, which is how the 2x survived to review in the first place. Required: at `broadcast.py:482` drop "and by a SWAP" and its trailing clause, saying the swap drains via `_drain_retire` instead; at `broadcast.py:47` say the constant is the default grace for `join` (tests and teardown diagnostics) and the base for `SWAP_DRAIN_TIMEOUT_S`, not a value the swap uses through `join`. |

#### F9 — verified fixed, by reproduction

Reproduced with my own harness, not the PR's tests, and controlled against the pre-fix state by reverting **only** the one changed line at the current head, so the rebase is not a confound.

| Scenario | Pre-fix (`join`) | Head (`_drain_retire`) | Documented bound |
|---|---|---|---|
| Single swap, wedged socket | **12.01s** | **6.01s** | 6.0s |
| Chain A→B→C, two wedged predecessors | **11.75s** | **6.01s** | 6.0s |

The compounding half — the subtle one — is genuinely gone, not merely reduced: the chain now costs the same single budget as one swap. The mechanism is right rather than incidentally lucky. `_drain_retire` (`broadcast.py:494-505`) takes `_dispatch_lock` only to read the thread handle and then joins outside it, while the wedged backend call sits under `_call_lock` (`broadcast.py:461-468`), so there is no path by which the drain inherits a second worker's timeout. The narrowing is sound for ordering: the predecessor's timer is `_stopped` and its `_publish_once` refuses under the `_retired` latch, so it cannot touch the pane after `stop()`.

**Both new regression tests fail pre-fix**, at the reviewer's original figures — checked, not assumed:

```
FAILED ...::test_the_swap_drain_is_bounded_by_the_number_it_documents
  AssertionError: the successor's first publish waited 12.01s, past the documented 6.0s swap drain bound
FAILED ...::test_a_swap_chain_does_not_compound_the_drain
  AssertionError: a two-swap chain delayed the pane by 11.81s — the drain is compounding across the chain again
2 failed in 15.44s
```

Both pass at head; the full multiplexer suite is `60 passed`, and `tests/unit/tui/test_multiplexer_broadcast.py` is `5 passed`.

#### The drain-window mitigation claim — verified true

The remediation's claim that the pane advertises the OUTGOING session during the drain, rather than nothing, is correct, and it does materially soften F9's residual risk. Real `ZellijBackend`, real marker file, wedged retire, sampling the pane's advertised binding every 250ms through the window:

```
bound to A            : aaaaaaaaaaaa
during drain, distinct advertised values: ['aaaaaaaaaaaa']
any moment advertising NOTHING during drain: False
after drain completes : bbbbbbbbbbbb
```

So the worst case is a stale binding bounded by the drain, never an unadvertised pane — the failure this package exists to prevent does not occur even in the timeout case.

#### F10 — verified fixed

The corrected statement is true. `reexec.py:151-152` is `os.execvpe`, and I confirmed directly that a POSIX exec runs no `atexit` handler (registered handler did not fire across the exec; the child printed and exited without it). The new text at `broadcast.py:599-620` now says every exec-shaped exit skips `atexit` rather than naming Windows alone, and its reasoning for why that is harmless on each shape matches the code: `_request_relaunch` exits through Textual, so `on_unmount` dispatches the withdrawal before `cli.py` calls `replace_self` after `asyncio.run` returns.

#### The rebase — checked, not taken on trust

I did not accept "neither touches the multiplexer path". I compared the PR's own contribution against each base, which is the check a rebase cannot fool:

- **`app.py` contribution is byte-identical.** `git diff f2a52b53..c173bfee -- local_operator/tui/app.py` and `git diff 04702e6e..efade76e -- local_operator/tui/app.py` produce the *same set of added/removed lines* (152 lines each, identical after sorting). The auto-merge changed nothing this PR authored.
- **`local_operator/multiplexer/` and `tests/unit/test_multiplexer.py` are identical** pre- vs post-rebase once the F9 commit is excluded; `test_multiplexer_broadcast.py` and `docs/multiplexer-resume.md` hash identically.
- **Main never touched the multiplexer path**: `git diff --stat f2a52b53..04702e6e -- local_operator/multiplexer tests/unit/test_multiplexer.py ...` is empty, so there was no semantic conflict to resolve wrongly.
- **#301 (`7a32630f`) does not interact.** Its two `_adopt_session`-adjacent hunks add `_report_attachment_restore(session)` after `_render_resumed_history` and two status-band kwargs. Reading the merged body at `app.py:1971-2085`: `_start_multiplexer_broadcast()` is still the third statement, immediately after `self._session = session` and `self._mobile_adopted(session)`, and #301's additions all land *after* it. Swap ordering is untouched. `_adopt_takeover_session` (`app.py:2088`) is deliberately not a conversation change and starts no broadcast, so #301's R4 change there publishes nothing.
- **#300 (`6203e617`) does not interact.** Its hunks are confined to MCP wiring (`_on_deferred_mcp_wiring_settled`, `_wire_mcp_status`) around `app.py:4130-4200`, nowhere near the multiplexer path or `on_unmount`.
- Version bump is clean: `0.37.0` → `0.38.0`, matching main's `04702e6e`, and `merge-base(origin/main, head)` is exactly `04702e6e`.

#### F1–F8 non-regression on the new base — spot-checked

Re-ran my own harnesses against the rebased head rather than trusting the re-run:

- **F1**: `stop()` returned in **1.60ms** with a publish parked inside a wedged call. The event loop is still never waited on.
- **F7**: 0/20 swaps lost at each of 0ms, 5ms and 50ms retire latency against the real `ZellijBackend` through `app.py`'s exact swap shape (round-2 pre-fix measured 20/20 lost at 5ms and 50ms).

#### CI

Matches the disclosed pattern exactly and I found no evidence it is this diff. The failing job is `test (3.12)` on run `32926904112` (`tests/unit/mobile/test_tui_bridge.py::test_tui_auto_registers_and_answers_control`, "no discovery record published"), while the **identical commit** passes `test (3.12)` on run `32926907203`. `lint`, `type-check`, `pip-audit`, `build` and both `test (3.13)` runs are green. The diff touches no `mobile/` file. Not filed.

#### Not filed

- The disclosed evidence limitation (end-to-end auto-resume unverified on a healthy cmux host) — the operator has explicitly chosen to ship on it.
- The `markers.py:404` TOCTOU noted in round 3 as record-only. Worth restating that `06c1a23f` *preserves* the sequencing that closes it: the successor still publishes strictly after the predecessor's withdrawal completes on any backend that answers, so the window remains reachable only if the drain times out.

**Verdict: Approve.** No blocker and no major remains. F9 was the last one and it is fixed at the mechanism level, with the chain compounding gone and a test that fails pre-fix pinning the bound so it cannot silently regress again. **This round is TERMINAL** — F11 is a two-line comment correction and is a follow-up, not a reason for a fifth round.

PR #305, opened 2026-08-25
feat(providers): patient backoff for connectivity-loss errors

damianvtran:
## Summary

Moving between networks (close laptop → walk to the car → reconnect to home wifi) killed in-flight sessions with:

```
✕ transient provider error: ConnectError: [Errno 8] nodename nor servname provided, or not known
```

That error is a DNS-resolution failure: the machine is briefly, fully offline. The harness already *retried* it (it wraps to `kind="transient"`, `retryable=True`), but the retry budget was tiny — the transient backoff caps each sleep at `BACKOFF_CAP_MS` (8s) over `maxRetries` (10), a total window of only ~80s. A normal close/move/reconnect cycle takes several minutes, so the budget ran out while the machine was still offline and the turn died.

This adds a dedicated **patient backoff** for connectivity-loss errors: the harness now rides out a reconnect for ~8 minutes on the shipped defaults, so a session survives the laptop coming back onto wifi. Ordinary provider 5xx/timeout retries are unchanged and stay fast.

- New `is_connectivity_loss()` predicate classifies the machine-is-offline family (DNS failure, network unreachable, no route to host, socket-level connection refused) distinctly from a reachable-but-5xx provider. The user-facing frame label is unchanged (`transient provider error`) — no new `ProviderErrorKind`.
- New `connectivity_backoff_delay_ms()` + `CONNECTIVITY_BACKOFF_CAP_MS` (60s) / `CONNECTIVITY_MAX_RETRIES` (15) give connectivity loss its own larger cap and longer budget, wired into **both** retry arms of `stream_with_failover` (the raw-exception arm these errors actually take, and the pre-wrapped `ProviderError` arm for completeness). Connectivity loss retries the **same** target in place and never rotates credentials or walks the fallback chain — being offline is nobody's fault and no other provider is reachable either.
- Configurable via `retry.connectivityMaxRetries` / `retry.connectivityBackoffCapMs` (snake_case accepted), defaulting to the ~8-minute window with zero config.
- The patient sleep remains abortable: Ctrl-C / abort breaks out immediately.
- Version bumps `0.33.2` → `0.34.0` (minor — new capability).

## Impact

**Before:** brief loss of network during an active turn surfaced `transient provider error: ConnectError: [Errno 8] …` within ~80s and ended the turn. Reopening the laptop after a commute meant a dead session.

**After:** connectivity loss is ridden out with exponential backoff climbing to a 60s ceiling for ~8 minutes (default). When wifi returns, the in-flight request resumes on the same credential and the turn completes as if nothing happened. A provider that is *reachable* but returning 5xx still fails fast on the unchanged 8s cap, so a genuinely broken upstream is not papered over.

## Design

Connectivity loss is a *narrower* transient than a 5xx: the whole machine is offline, so credential rotation and fallback-chain walking are futile (no provider is reachable). The predicate is therefore checked **ahead of** the server-fault retry logic in both `except` arms, retrying the same target in place on a separate `connectivity_retries` counter that does not charge the server-fault ceiling. Markers are matched against the same lowercased `class-name + detail` haystack `wrap_transport_error` already builds, each with an errno/scenario citation in `_CONNECTIVITY_LOSS_MARKERS`.

## Real-path validation

All evidence below is produced against the real `local_operator.providers.failover` module (not mocks of it), run from the worktree venv.

### 1. The exact error from the report classifies correctly, frame label unchanged

```
=== 1. The exact error from the screenshot classifies as connectivity loss ===
  raw str:      'transient provider error: ConnectError: [Errno 8] nodename nor servname provided, or not known'
  frame kind:   transient   (user still sees 'transient provider error')
  connectivity: True
  is_transient: True

=== 2. Other real offline signals also classify ===
  True   ConnectError: [Errno 51] Network is unreachable
  True   ConnectError: [Errno 65] No route to host
  True   ConnectError: [Errno 61] Connection refused
  True   gaierror: [Errno 8] nodename nor servname provided, or not known
  True   Temporary failure in name resolution
  True   getaddrinfo failed

=== 3. An ordinary reachable-but-5xx failure is NOT connectivity loss (stays fast) ===
  503 kind=transient  connectivity=False  transient=True
```

### 2. Shipped defaults give a ~8-minute offline tolerance; 5xx stays on the fast cap

```
=== 4. Delay sequence + total offline tolerance with SHIPPED DEFAULTS (no config) ===
  defaults: connectivity_max_retries=15, cap_ms=60000, base_delay_ms=500
  per-attempt sleeps (ms, with jitter): 394, 810, 1789, 3741, 6977, 14380, 25729, 55450, 52851, 51249, 46378, 52429, 55772, 48662, 50724
  TOTAL offline tolerance: 467.3s (~7.8 min) across 15 retries

=== 5. Contrast: an ordinary 5xx transient stays on the fast 8s cap ===
  5xx sleeps (ms): 394, 810, 1789, 3741, 6977, 7190, 6432, 7393, 7046, 6833
  max single 5xx sleep: 7393ms (<= 8000 cap) — unchanged, not slowed
```

### 3. End-to-end: a stream that goes offline mid-turn and recovers completes

`test_connectivity_loss_recovers_within_patient_budget` drives the real `stream_with_failover`: the client raises the verbatim `ConnectError: [Errno 8] …` **12 times** (more than `maxRetries=10` would tolerate), then the network returns and it streams a normal completion. The stream completes, it retried in place 13 times, the credential was **never rotated**, and the delays climbed past the 8s cap toward the 60s ceiling — the old budget would have surfaced the error and killed the turn before attempt 10.

## Tests

7 new tests in `tests/unit/providers/test_failover.py`, all green (`13 passed` for the connectivity/regression subset; `176 passed` full failover suite; `642 passed` full providers suite):

- `test_connectivity_markers_classify_as_connectivity_loss` — parametrized over the real Errno 8 string + Errno 51/65/61, name-resolution, getaddrinfo.
- `test_ordinary_transient_5xx_is_not_connectivity_loss` — a status-bearing 5xx is excluded.
- `test_connectivity_backoff_is_patient_not_the_8s_cap` — delays exceed 8s up to the 60s ceiling.
- `test_connectivity_loss_recovers_within_patient_budget` — the end-to-end recovery above.
- `test_ordinary_transient_5xx_still_uses_fast_budget` — regression guard: 5xx stays on 8s cap and the small server-fault budget.
- `test_connectivity_loss_backoff_is_abortable` — abort during a 60s patient sleep raises `kind="aborted"` immediately.
- `test_connectivity_config_keys_parse_camel_and_snake` — both spellings + shipped defaults.

## Gates

```
flake8 .                                   PASS
black --check (26.1.0)                      492 files unchanged
isort --check-only --profile black          PASS
pyright                                      0 errors, 0 warnings, 0 informations
pytest tests/unit/providers                  642 passed
```

### DIFF
--- a/local_operator/providers/failover.py
+++ b/local_operator/providers/failover.py
@@ -127,13 +127,16 @@ _TIMEOUT_MARKERS = ("timeout", "timed out", "deadline exceeded", "stream stalled
 #:   - "no route to host" / "errno 65": ``EHOSTUNREACH`` (macOS 65, Linux 113).
 #:   - "network is down" / "errno 50": ``ENETDOWN`` — the interface is still
 #:     coming up.
-#:   - "connection refused" / "errno 61": ``ECONNREFUSED`` (macOS 61, Linux
-#:     111). Included because during a reconnect a captive portal or the local
-#:     stack refuses at the socket layer before any HTTP response exists. This
-#:     one CAN also mean a genuinely-down endpoint rather than an offline
-#:     machine, but the patient path is safe either way: the backoff is
-#:     abortable, and once the (minutes-long) budget is spent the diagnostic
-#:     error still surfaces, so a real outage is not hidden, only waited out.
+#:
+#: DELIBERATELY EXCLUDED — do not re-add: ``ECONNREFUSED`` ("connection refused"
+#: / errno 61 on macOS, 111 on Linux). A refused connection is a TCP RST from the
+#: destination, which PROVES the host was reachable — the exact opposite of the
+#: offline machine this path exists for. An offline machine cannot even resolve
+#: or route to a host, so it yields the DNS/route/net-unreachable failures above,
+#: never a refusal. Treating a refusal as connectivity loss would give a
+#: genuinely-down LOCAL provider (ollama, LM Studio, a localhost proxy) the
+#: 8-minute patient wait AND suppress the fallback-chain walk that otherwise
+#: routes around it — a regression. A refusal stays an ordinary transient error.
 _CONNECTIVITY_LOSS_MARKERS = (
     "nodename nor servname",
     "errno 8",
@@ -146,8 +149,6 @@ _CONNECTIVITY_LOSS_MARKERS = (
     "errno 65",
     "network is down",
     "errno 50",
-    "connection refused",
-    "errno 61",
 )
 
 
@@ -1196,6 +1197,22 @@ def _normalize_chains(chains: Mapping[str, Any]) -> dict[str, list[Any]]:
     return normalized
 
 
+def _coerce_int(value: Any, default: int) -> int:
+    """A config integer that degrades to ``default`` on junk input.
+
+    ``from_settings`` re-parses on every model call and must never raise, so a
+    non-numeric or missing value (``None`` when the key is absent) falls back to
+    the shipped default rather than killing the turn. Mirrors the ``float()``
+    guard the ``usageReservePercent`` path already uses at this same boundary.
+    """
+    if value is None:
+        return default
+    try:
+        return int(value)
+    except (TypeError, ValueError):
+        return default
+
+
 @dataclasses.dataclass(frozen=True)
 class RetrySettings:
     """The ``values.retry.*`` config surface."""
@@ -1229,22 +1246,24 @@ class RetrySettings:
             reserve_percent = min(100.0, max(0.0, float(reserve)))
         except (TypeError, ValueError):
             reserve_percent = 10.0
+        # A malformed connectivity budget must degrade to the shipped default,
+        # not crash from_settings on EVERY turn (it re-parses per model call).
+        # Same contract as usageReservePercent above: a bad value in the retry
+        # block is a preference the harness can safely ignore, never a fatal.
+        connectivity_max_retries = _coerce_int(
+            retry.get("connectivityMaxRetries", retry.get("connectivity_max_retries")),
+            CONNECTIVITY_MAX_RETRIES,
+        )
+        connectivity_backoff_cap_ms = _coerce_int(
+            retry.get("connectivityBackoffCapMs", retry.get("connectivity_backoff_cap_ms")),
+            CONNECTIVITY_BACKOFF_CAP_MS,
+        )
         return RetrySettings(
             enabled=bool(retry.get("enabled", True)),
             max_retries=int(retry.get("maxRetries", retry.get("max_retries", 10))),
             base_delay_ms=int(retry.get("baseDelayMs", retry.get("base_delay_ms", 500))),
-            connectivity_max_retries=int(
-                retry.get(
-                    "connectivityMaxRetries",
-                    retry.get("connectivity_max_retries", CONNECTIVITY_MAX_RETRIES),
-                )
-            ),
-            connectivity_backoff_cap_ms=int(
-                retry.get(
-                    "connectivityBackoffCapMs",
-                    retry.get("connectivity_backoff_cap_ms", CONNECTIVITY_BACKOFF_CAP_MS),
-                )
-            ),
+            connectivity_max_retries=connectivity_max_retries,
+            connectivity_backoff_cap_ms=connectivity_backoff_cap_ms,
             model_fallback=bool(retry.get("modelFallback", retry.get("model_fallback", True))),
             usage_aware_fallback=bool(
                 retry.get("usageAwareFallback", retry.get("usage_aware_fallback", False))
@@ -2052,7 +2071,7 @@ async def stream_with_failover(
                     # Patient budget spent while still offline: surface the most
                     # diagnostic failure of the walk, same as any other
                     # exhaustion, so the reported-error semantics stay intact.
-                    raise reported if reported is not None else exc
+                    raise (reported if reported is not None else exc) from exc
                 if is_server_side_failure(exc):
                     server_fault_requests += 1
                     server_faults_by_target[route_key] = server_fault_requests
--- a/local_operator/providers/failover.py
+++ b/local_operator/providers/failover.py
@@ -97,6 +97,59 @@ _WEAK_USAGE_LIMIT_MARKERS = ("usage", "insufficient")
 
 _TIMEOUT_MARKERS = ("timeout", "timed out", "deadline exceeded", "stream stalled")
 
+#: Substrings that mean "this MACHINE cannot reach ANY network right now" — a
+#: DNS lookup, route, or socket connect that failed before a single HTTP byte
+#: was exchanged, so no provider was actually reached. Matched against the same
+#: lowercased ``class name + detail`` haystack :func:`wrap_transport_error`
+#: builds, because the exception class and its errno text are the whole of the
+#: evidence: a connectivity loss carries no HTTP status and no provider words.
+#:
+#: This is a DIFFERENT condition from an ordinary transient 5xx even though both
+#: classify ``kind="transient"``. A 5xx means a reachable provider is unwell, so
+#: the fix is to retry FAST (8s cap) and rotate to another credential/provider.
+#: A connectivity loss means the laptop was closed, carried between office and
+#: home, and is mid-reconnect to wifi — every provider is equally unreachable,
+#: so rotating credentials and walking the fallback chain are pure waste. What
+#: that scenario needs instead is PATIENCE: a minutes-long backoff (see
+#: :func:`connectivity_backoff_delay_ms`) that keeps the interactive session
+#: alive across a normal reconnect rather than dying inside ~80s.
+#:
+#: Each marker cites the errno/wording it catches. macOS and Linux word the same
+#: failure differently, so both spellings are listed:
+#:   - "nodename nor servname" / "errno 8": macOS ``getaddrinfo`` DNS failure —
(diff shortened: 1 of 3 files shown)


damianvtran:
### Agent review — round 1

Reviewer: reviewer subagent (anthropic/claude-opus-4-8)
Scope: origin/main..5ed5428640c1373a14d57bf4711719c5c04b34f5

The patient-backoff design is sound and well-integrated: connectivity loss is checked ahead of the server-fault ledger in both retry arms, retries the same target in place, never rotates credentials, is bounded by `connectivity_max_retries` (so it surfaces the diagnostic error rather than hanging), and stays abortable through `_abortable_sleep`. All gates are green locally (flake8, black 26.1.0, isort, pyright with `--pythonpath .venv/bin/python`, `642 passed`). One real defect in the classification set is worth fixing before ship; the rest are minor/nit.

**Findings**

- **[C1] major — `connection refused` / `errno 61` should not be a connectivity-loss marker** `_CONNECTIVITY_LOSS_MARKERS` in `local_operator/providers/failover.py:143-146` includes `"connection refused"` / `"errno 61"` (ECONNREFUSED). ECONNREFUSED is delivered as a TCP RST from the destination host, which *requires* L3 reachability — a genuinely offline machine (laptop closed / mid-reconnect) produces DNS failure, `ENETUNREACH` (51), `EHOSTUNREACH` (65) or a timeout, never ECONNREFUSED. So this marker does not catch the machine-offline case it is justified by; what it actually catches is a *reachable* endpoint that is down, overwhelmingly a local provider (ollama / LM Studio / a local proxy on `127.0.0.1`). For a user whose primary is a down local model with a cloud fallback, the previous behavior was: fail fast and walk the chain to the reachable cloud provider. With this change ECONNREFUSED is now treated as offline, so the harness waits ~8 minutes on the dead local target AND deliberately refuses to walk the fallback chain — a regression of a previously-working failover path. The author's comment (`failover.py:127-134`) acknowledges "this one CAN also mean a genuinely-down endpoint" and argues it is "safe either way," but the safety argument omits the suppressed chain-walk. Fix: drop `"connection refused"`/`"errno 61"` from the marker tuple (the DNS/route/net-unreachable markers already cover true offline), or, if the localhost captive-portal case is real, gate it so loopback/localhost targets are excluded from `is_connectivity_loss` — but the predicate currently only sees the error, not the target, so dropping the marker is the clean fix.

- **[C2] minor — no test proves connectivity loss leaves the fallback chain unwalked** `test_connectivity_loss_recovers_within_patient_budget` (`tests/unit/providers/test_failover.py`) asserts `auth.rotations == []` on a single-provider auth store, which proves no *credential* rotation but not that a multi-provider fallback chain is *not* walked. The whole design rests on "retry the same target in place, never walk the chain"; a test with a real `fallbackChains` config asserting the second target is never reached would lock that invariant and is the natural place to also pin the C1 behavior once decided.

- **[C3] minor — config coercion crashes on non-numeric values** `RetrySettings.from_settings` (`failover.py:190-201`) wraps the new keys in bare `int(...)`, so `retry.connectivityMaxRetries = "lots"` raises `ValueError` and takes down settings load. This mirrors the pre-existing `max_retries`/`base_delay_ms` pattern (`failover.py:188-189`), so it is consistent rather than newly risky and I would not block on it — but the new surface inherits the same sharp edge; a guarded coercion helper would fix all of them.

- **[C4] nit — the asyncio warnings on the new sync tests come from the module mark, not a stray decorator** The premise that a stray `@pytest.mark.asyncio` sits on a non-async parametrized test is not quite right: the connectivity tests carry only `@pytest.mark.parametrize`, and the `PytestWarning: marked with asyncio but not an async function` is produced by the module-wide `pytestmark = pytest.mark.asyncio` at `tests/unit/providers/test_failover.py:62`, which already fires on pre-existing sync tests. The three new sync tests just add more of the same warning. Pre-existing wart, not introduced here; if cleaned up it should be at the module level (e.g. an explicit non-async opt-out), not by editing this PR's tests.

- **[C5] nit — exhaustion raise in the `ProviderError` arm omits `from exc`** `failover.py:2055` does `raise reported if reported is not None else exc` while the parallel `except Exception` arm at `failover.py:2178` uses `raise (... ) from exc`. Harmless here (`exc` is already a `ProviderError`), but the two exhaustion raises read inconsistently for no reason.

**Verdict:** Not clean — one **major** (C1). No blockers. Once C1 is resolved (drop the ECONNREFUSED marker or exclude loopback) and answered, the remaining items are minors/nits and need not gate. Recommend one more round after remediation to confirm the marker change against the current head.

damianvtran:

All four actionable findings fixed in 4bb22e21; full gate set green (flake8, black, isort, pyright, `pytest tests/unit/providers` = 643 passed).

- **[C1]** Fixed in 4bb22e21 — Removed `connection refused` / `errno 61` from `_CONNECTIVITY_LOSS_MARKERS`. ECONNREFUSED is a TCP RST from the destination, which proves the host was reachable, so it never fires on a genuinely-offline machine and would have wrongly given a down local provider (ollama/LM Studio/localhost proxy) the patient wait while suppressing the fallback-chain walk. Added an explicit "DELIBERATELY EXCLUDED — do not re-add" note on the marker tuple explaining why. Flipped the classification test: `[Errno 61] Connection refused` now asserts it is an ordinary transient, NOT connectivity loss, and dropped it from the connectivity-loss parametrize set.
- **[C2]** Fixed in 4bb22e21 — Added `test_connectivity_loss_does_not_walk_fallback_chain`: a multi-provider `fallbackChains` config where the primary raises the Errno 8 connectivity error asserts the fallback target's client is never constructed (`specs_seen == {"openai/gpt-4o"}`) — the primary is retried in place, no rotation.
- **[C3]** Fixed in 4bb22e21 — Wrapped the `connectivityMaxRetries` / `connectivityBackoffCapMs` coercion in a `_coerce_int` helper (try/except → default), matching how `usageReservePercent` handles bad input; `from_settings` re-parses per model call and must never raise. Added `test_connectivity_config_bad_value_falls_back_to_default` covering non-numeric and `None`.
- **[C5]** Fixed in 4bb22e21 — Added `from exc` to the ProviderError-arm exhaustion raise, matching the parallel `except Exception` arm.
- **[C4]** No action — confirmed non-issue: the asyncio-mark warning comes from the module-wide `pytestmark`, is pre-existing, and fires on other sync tests in the file.

damianvtran:
### Agent review — round 2

Reviewer: reviewer subagent (anthropic/claude-opus-4-8)
Scope: 4bb22e21998072eea3fa1ac5d2783603370210dd (round-1 head == current head; range 318c1a4b..4bb22e21)

All five round-1 findings are resolved in the diff, verified against the source rather than the remediation claims. Gates re-run green on this head: flake8 clean, `black==26.1.0 --check` clean (492 files), isort `--profile black` clean, pyright 0 errors/0 warnings, and `tests/unit/providers` 643 passed.

- **C1 (MAJOR) — resolved.** `_CONNECTIVITY_LOSS_MARKERS` (failover.py) no longer contains `connection refused` or `errno 61`; a `DELIBERATELY EXCLUDED — do not re-add` block explains that a refusal is a TCP RST proving reachability, the opposite of an offline machine, and would wrongly give a down local provider the patient wait while suppressing the fallback walk. The true-offline markers are all still present and still classify: `errno 8`/`nodename nor servname` (DNS), `errno 51`/`network is unreachable` (ENETUNREACH), `errno 65`/`no route to host` (EHOSTUNREACH), `errno 50`/`network is down`, plus `temporary failure in name resolution`, `name or service not known`, and `getaddrinfo failed`. `test_ordinary_transient_5xx_is_not_connectivity_loss` now asserts a wrapped `[Errno 61] Connection refused` classifies `kind == "transient"` and `not is_connectivity_loss(...)`, and the parametrized marker test still asserts each offline wording classifies as connectivity loss both raw and wrapped.

- **C2 (MINOR) — resolved.** `test_connectivity_loss_does_not_walk_fallback_chain` configures a real two-provider store (`openai` + `anthropic`) with a `fallbackChains.default` pointing at the anthropic target, drives an Errno-8 loss on the primary, and asserts `set(specs_seen) == {"openai/gpt-4o"}` — the fallback target's client is never constructed. This is a genuinely distinct invariant from the single-provider recovery test (which only proves no credential rotation).

- **C3 (MINOR) — resolved.** New `_coerce_int(value, default)` helper wraps `int()` in a `try/except (TypeError, ValueError)` and returns the default on junk or `None`, mirroring the existing `usageReservePercent` `float()` guard. `from_settings` routes both `connectivityMaxRetries` and `connectivityBackoffCapMs` (camel and snake) through it. `test_connectivity_config_bad_value_falls_back_to_default` proves `"lots"` and `None` degrade to the shipped defaults without raising.

- **C5 (NIT) — resolved.** Both patient-budget exhaustion raises carry `from exc`: the ProviderError arm (`raise (reported if reported is not None else exc) from exc`) and the transport arm (`raise (reported if reported is not None else wrapped) from exc`).

C4 was a non-issue in round 1 and needed no action; nothing in the remediation reopens it.

**Findings**

No blocker or major findings. Round 1 findings C1–C5 confirmed resolved. No new defects introduced by the remediation: the marker changes are the intended narrowing only, the added tests assert real invariants (in-place retry count, no rotation, no fallback client construction, abortability, fast-path 5xx still capped at 8s), and the config guard cannot raise.

This round is TERMINAL — clean, no further round required.

PR #306, opened 2026-08-25
feat: peer-to-peer session messaging (lop send / lop sessions)

damianvtran:
## Summary

Adds peer-to-peer messaging between independent `lop` sessions on the same machine, plus a session-listing/monitoring command. Both work **without cmux** — they ride the filesystem session registry and per-session control socket that the mobile subsystem already publishes, so no new IPC stack, socket, auth scheme, or registry is introduced.

- **`lop send <target> "<message>"`** hands a message to another local session. The message lands in the target's transcript **and** its LLM history (so the model reads it on its next turn), rendered with a distinct cross-session indicator in the TUI and the mobile web surface.
- **`lop sessions`** lists every active session with its resource usage (RSS + true footprint), state (live/wedged/stale), uptime, and heartbeat age — the monitoring surface for "what's running and what's it costing."
- Minor version bump `0.35.1` → `0.36.0` (feature).

Design doc committed at `docs/design/peer-send.md`.

## Impact

Before this change, one session could only reach another through cmux (`cmux send`) or the hub (which only reaches agents *inside* a session's own subagent hierarchy). There was no way for two independent top-level `lop` sessions to talk, and no first-class way to see what sessions were running or how much memory each used.

After this change, any session can message any other on the same account, and `lop sessions` gives an at-a-glance monitoring table. The motivating case: a session that detects system memory pressure can now message the offending session directly and tell it to be more efficient, and the human sees that note in the target's own history marked as inbound cross-session.

### Delivery modes (sender chooses)

- **Default (mailbox, record-only):** the row appears in the target's history now; an idle target stays idle and reads it on its next turn. Non-interrupting.
- **`--wake`:** if the target is idle, drive a turn now so it acts on the message immediately.
- **`--now` / `--steer`:** inject mid-turn like a steer, to redirect a session that is actively working.

Scope is any-local-to-any-local: the control-socket record is `0600` under a `0700` dir, so same-account is already the trust boundary (identical to the mobile daemon's).

## Design notes

- **One additive control op** (`peer_message` with a `mode` field), no `PROTOCOL_VERSION` bump — an older registrant answers the unknown op with its existing error, which the sender surfaces as a clear soft failure.
- **New `peer_message` EntryKind / `PEER_MESSAGE_MESSAGE_TYPE` custom type**, deliberately *not* reusing `parent_message` (which means "your hub parent said this" and has no slot for the sending session's identity).
- **`Session.receive_peer_message`** is a new method because no existing path both persists durably *and* injects into live `_context.messages` without driving a turn. The record-only branch is the only one that appends directly; wake/steer persist through the existing pipeline/steer paths (no double-append). `PEER_MESSAGE_MESSAGE_TYPE` is added to the `_default_convert_to_llm` allow-list so the model actually sees it.
- **Resource usage** uses `ps`/`top`/`smaps_rollup` via subprocess with graceful `None`-degradation (psutil is not a dependency).

### Live end-to-end (real stack, real socket, real CLI client)

A real `Session` behind a real `Registrant` control socket, messaged by the real `send_peer_message` client over loopback, then the on-disk transcript and converted LLM history read back:

```
[setup] real session id=sess transcript=/…/lop-peer-e2e-…/sess
[setup] registrant live on 127.0.0.1:51780 pid=51896
[send ] ack detail: 'delivered to the mailbox (will be read on the next turn)'
[disk ] peer rows persisted: 1
[disk ] body='RAM is tight — please trim your eval kernels.'
[disk ] sender label='Find RAM-Heavy Lop Instance' model='anthropic/claude-opus-4'
[model] user turns carrying the peer envelope: 1
[model] envelope seen by model:
        <peer-session-message from_pid=12345 conversation='Find RAM-Heavy Lop Instance' model='anthropic/claude-opus-4'>
ALL LIVE E2E CHECKS PASSED
```

This proves the message is (1) delivered over a real socket, (2) durably persisted as exactly one transcript row, (3) carries the sender identity, and (4) reaches the model through the allow-list.

### `lop sessions` against the live machine

```
live      21500 daemon  Minerva feed_filters Doc alibaba-token-plan/qwen3    17.9M    191.0M       1d      4s
live      48250 tui     Local Operator browser e openai/gpt-5.6-sol         176.6M    346.0M       2h     14s
live      74800 tui     Work Queues UX and Logic anthropic/claude-opus-4-   358.0M      3.0G       4h      2s
…
```

Reconciled against an independent `top -l1` measurement: the FOOTPRINT column matches `top`'s single-process footprint to within rounding (e.g. pid 74800: `lop` 3.0G vs `top` 3288M; pid 21500: 191M exact). **Known limitation (follow-up):** the column reports the main process only, not the session's child eval/pytest kernels — a session whose children hold gigabytes shows light. A children-inclusive column is a candidate follow-up.

### TUI indicator (rendered frame, real `OperatorApp` with CSS)

Captured via `app.save_screenshot` on the real app and viewed:

The block renders with a `↔` gutter glyph on every line and a muted header `peer message from "Find RAM-Heavy Lop Instance" (pid 52095, anthropic/claude-opus-4)` above the body — visually distinct from a user prompt (`▌`) and a wake receipt. Screenshot attached in the review thread.

### Mobile web

`pnpm build` (which runs `tsc -b`) succeeds; the compiled bundle contains the `peer_message` case. `dist/` is gitignored (build-at-package), so there is no artifact to commit.

### Gates

- `flake8` clean; `black --check` (26.1.0) 519 files unchanged; `isort --check` clean; `pyright` 0 errors.
- Peer/mobile/CLI unit tests: 161 passed; session peer tests + mobile peer tests: 73 passed.
- Full non-TUI unit suite: 4278 passed. The 6 `test_server_models.py::…env_config` failures **pre-exist on the base commit** (verified in a throwaway worktree at the base ref; this PR touches zero files under `server/`).
- TUI suite failures are all either pre-existing on base or parallel-load flakes that were confirmed identical on the base ref (`test_boot_typing_sends_prompt`, `test_cost_aggregation` fail deterministically in isolation on base too; `test_ask_picker`/`test_band_panels` pass in isolation).

## New guide

`local_operator/guides/peer-messaging/GUIDE.md` (auto-discovered) documents both commands, the three delivery modes, target selection, and limits — so the agent knows how to use `lop send` / `lop sessions`.


damianvtran:
### Design review — round 1
Reviewer: designer (claude-opus-4)
Scope: origin/main..3cd9bfdfd899b8190eeb4ff022b68fa7fa706d9b

Visual review of the cross-session message indicator: `PeerMessageBlock` in the TUI (`local_operator/tui/widgets/transcript.py` ~L1400) and the `peer_message` card in the mobile web transcript (`local_operator/mobile/web/src/components/transcript.tsx` L147-165).

**Rendered artifacts I looked at**
- TUI, full sender — the manager's captured frame `/tmp/peer-send-tui-indicator.png` (100×30, real `OperatorApp` with `local_operator.tcss`).
- TUI, lean sender (`pid` only), bare sender (no identity), and 60-col narrow wrap — re-captured myself via `shot_peer_variants.py` → SVG → served over http → browser screenshot.
- Mobile: `pnpm install --frozen-lockfile && pnpm build` succeeds; the `peer_message` case and the `Peer session` fallback are present in `dist/`. I reviewed the component + compiled output but **could not render the card live** (needs the daemon + a seeded projection I did not stand up) — the mobile findings below are from source/build, not pixels.

**What works.** The TUI block reads unambiguously as inbound cross-session. The `↔` gutter rule running the full height of the block (matching `UserBlock`'s `▌` discipline) plus the muted sender header clearly separate it from the user's own prompt and from a wake. The full-sender header `peer message from "Find RAM-Heavy Lop Instance" (pid 52095, anthropic/claude-opus-4)` names who reached in exactly as intended; the bare-sender fallback `peer message from another session` is graceful; the narrow-width wrap keeps every continuation row behind the `↔` gutter so the block stays one visual unit. Muted rule/header vs full-fg body is the right weight hierarchy — loud enough to register, quiet enough not to fight the live-turn accent.

---

**D1 — MINOR — TUI, `transcript.py:1473-1474` (`_header`).** The lean-sender path produces a dangling preposition: with a `pid` (and/or `model`) present but no `conversation_name`, the header renders `peer message from (pid 52095)` — "from" with no object, then a parenthetical. Confirmed in my lean-variant frame. The full and bare cases both read fine; only this middle case is awkward.
*Fix:* when `name` is absent but `bits` are present, drop the dangling `from` — render `peer message (pid 52095, anthropic/claude-opus-4)` (parenthetical attaches to "peer message", not to "from"), or promote the pid to the subject: `peer message from pid 52095`. One-line change in the `if bits:` branch.

**D2 — MINOR — mobile, `transcript.tsx:154-161` vs TUI `_header`.** The two surfaces disagree on how the sender is labelled. TUI: `peer message from "<name>" (pid N, <model>)` — quoted name, comma-joined parenthetical, an explicit "peer message from" framing. Mobile: `↔ <name> · pid N · <model>` — middot-joined, no quotes, no "peer message from" verb. Without the framing verb the mobile line reads more like a status chip / attribution byline than "a message another session sent you", which is the whole point of the indicator. The `↔` glyph carries the cross-session meaning but not the *message-from* meaning.
*Fix:* bring the mobile label into parity — prefix with the same framing, e.g. `↔ peer message from <name>` (keep the middot detail list if you prefer it to the TUI's comma form, but the "peer message from" verb should match). At minimum quote or otherwise set off the conversation name so it doesn't blend with the pid/model metadata.

**D3 — MINOR — mobile, `transcript.tsx:160`.** The peer card uses `border border-hairline border-l-2 border-l-accent bg-surface rounded-md` — byte-for-byte the same container treatment as `parent_message` (L134). The only things distinguishing an inbound peer message from a hub-parent message on mobile are (a) the `↔` glyph in the header and (b) left-full-width vs parent's right-aligned `justify-end max-w-[85%]`. That is thinner separation than the TUI gives (distinct glyph *and* distinct block idiom). Two different "someone reached in" kinds sharing the accent left-bar risks being read as the same kind at a glance, especially when both appear in one transcript.
*Fix:* give the peer card a visually distinct accent from `parent_message` — the cleanest is to align the *color* language with the TUI, where peer wears muted/neutral ink and no accent. Consider dropping `border-l-accent` for peer (or using a different role token) so the accent bar isn't overloaded across parent and peer. Verify the two side by side once the card can render live.

**D4 — NIT — mobile, `transcript.tsx:161`.** The sender attribution — the only "who reached in" signal on mobile — is `text-meta text-ink-dim`: the smallest type role on the dimmest ink ramp. On `bg-surface` that is the lowest-contrast text in the card carrying the most identifying information. TUI uses the `muted` token (a step brighter than `dim`) for the equivalent header. Recommend `text-ink-muted` for the peer label to match the TUI's weight and keep the attribution legible on the light theme. Please check the contrast ratio on both themes once rendered.

**D5 — NIT — TUI, `PeerMessageBlock`.** No blank-body / whitespace-only guard is visible in `_body_rows` (L1508-1524) beyond the leading/trailing empty-row trim. A `lop send <t> "   "` (validated non-empty by strip on the sender side, so unlikely) or a body that is all newlines would collapse to a header with an empty gutter column beneath. Low likelihood given `text.strip()` validation upstream; noting only so a future validation relaxation doesn't leak an empty card. No change required now.

---

**Could not render live:** the mobile `peer_message` card in a running phone projection (D2/D3/D4 are from source + built bundle, not pixels). If the daemon-seeded path is easy to stand up in remediation, a single mobile screenshot of the card next to a `parent_message` card would confirm or clear D3 directly.

**Verdict:** No BLOCKER, no MAJOR. The indicator does its core job — the TUI block reads unambiguously as inbound cross-session across full/lean/bare/narrow states. This round is **TERMINAL** for design. D1-D4 are worth a remediation pass (D1 is a quick copy fix, D2-D4 tighten TUI/mobile parity and mobile legibility); D5 is a follow-up note. None block the merge.

Blocker: 0 · Major: 0 · Minor: 4 · Nit: 1 (D5 counted as nit)

damianvtran:
### UX review — round 1
Reviewer: ux-reviewer (anthropic/claude-opus-4-8)
Scope: origin/main..3cd9bfdfd899b8190eeb4ff022b68fa7fa706d9b

I walked the real flows from the worktree — `lop sessions`, `lop sessions --json`,
`lop send --help`, and `lop send` against the live registry on this machine (18
real sessions running), plus every target-resolution error path. I read the receive
side (`session/peer.py`, `mobile/{registrant,tui_handle,owned}.py`) and the rendered
TUI indicator frame. Findings below are ordered by severity.

---

**U1 — BLOCKER — `lop send` to any real, busy session crashes the sender with a
raw traceback, and the message is delivered anyway (double failure).**

Flow: `lop send "Work Queues" "..."` — a *unique, valid* substring resolving to one
live session (pid 74800, a normal long-running TUI). The command did not report
success or a clean error; it dumped a Python stack trace ending in
`ValueError: Separator is not found, and chunk exceed the limit`
(`asyncio/streams.py` → `peer_client.py:74` `reader.readline()`), then the harness's
generic "Please review and correct the error to continue" banner.

Root cause: `peer_client.send_peer_message` (`peer_client.py:53`) dials daemon-class,
so the registrant's first response is an **unsolicited full-projection welcome**
(`registrant.py:526` `await self._push_to(conn)`), serialized as one
newline-terminated JSON line (`registrant.py:845`). For any session with a
non-trivial transcript that welcome line exceeds the 1 MB `readline` limit set at
`peer_client.py:53` (`limit=1 << 20`), and `readline()` raises `ValueError`. The CLI
catches only `except (RuntimeError, ConnectionError, OSError)` (`cli.py:1106`), so
`ValueError` escapes as an uncaught traceback.

Two things make this a BLOCKER, not a MAJOR:
1. It fires on the *normal* path — any target with real conversation history, which
   is essentially every session a user would send to. The live end-to-end test in
   the PR evidently used a fresh/small target, so it never hit the wall.
2. The write precedes the failed read (`peer_client.py:59-72` writes and drains the
   `peer_message` frame *before* the read loop), so **the message is actually
   delivered** while the sender reports catastrophic failure. A user who sees this
   traceback will reasonably re-send, producing duplicate peer cards in the target —
   the worst possible outcome for a "did my message get there?" interaction.

Remediation: the sender must not choke on the welcome. Either (a) match the
registrant's own read discipline — raise the `open_connection` limit and skip
oversized/non-matching frames instead of letting `readline` raise — or better, (b)
detect the daemon-welcome and drain it defensively, and at minimum add `ValueError`
(and `asyncio.LimitOverrunError`) to the `except` at `cli.py:1106` so a read fault
becomes the same soft, non-zero "could not deliver" line the docstring at
`peer_client.py:44-47` already promises ("never a traceback"). Note (a)/(b) alone
still leave the delivered-but-reported-failed ambiguity unless the ack is reliably
read; the fix must end with the sender reading its actual `ack`, not merely not
crashing. Add a regression test that dials a registrant whose projection is >1 MB.

---

**U2 — MAJOR — No self-send guard: `lop send --pid <own-tui-pid>` will deliver a
peer card into the very session that launched the send, mislabeled as an inbound
cross-session message.**

Flow: from inside a `lop` TUI, a user (or the agent itself via a tool) runs
`lop send --pid <this session's pid> "..."`, or a substring that happens to match
its own conversation name. `_resolve_peer_target` (`cli.py:969`) has no check that
the resolved `record.pid` is the caller's own session, and `_peer_sender_identity`
(`cli.py:948`) already computes `os.getppid()` — the launching session's pid — so
the information to detect a self-send is right there and unused. The result is a
session messaging itself, rendered with the `↔ peer message from "<own name>"`
indicator as though a *different* session sent it. That is confusing at best and, in
`--wake`/`--now` mode, self-triggers a turn.

Remediation: when the resolved target pid equals the sender's own session pid
(`os.getppid()` from the identity lookup), reject with a clear message
("that's this session; use the composer to talk to yourself") — or, if self-send is
ever intended, at least drop the "from another session" framing. A one-line guard in
`send_command` after resolution.

---

**U3 — MINOR — The 1 MB body cap advertised in the guide is not what a user actually
hits first.** GUIDE.md line 120 says bodies are capped at 256 KB, and `cli.py:1087`
enforces `_PEER_MESSAGE_MAX_BYTES`. That is fine for the *body*. But because of U1,
the practical failure a user meets is on the *receive*-side welcome size, not their
body size — so the documented "large paste is rejected with a clear error" promise
(GUIDE.md:120-122) is only half true: an in-cap body to a big-transcript target still
fails, just with the U1 traceback instead. Once U1 is fixed this reconciles; noting
it so the guide's "clear error" claim is actually honored on the path users hit.

---

**U4 — MINOR — `--wake` while the target is busy is a silent no-op with no sender
feedback.** GUIDE.md:66-67 and the design doc both say `--wake` is a no-op on a busy
target ("the running turn will read it anyway"). Correct behavior, but the *sender*
gets no signal which happened: `lop send --wake` prints the same
`→ <name> (pid N): <detail>` regardless of whether it woke an idle session or
silently degraded to record-only. A user who runs `--wake` specifically because they
want the target to act *now* cannot tell from the output whether it did. Remediation:
have the ack `detail` distinguish the outcome ("woke idle session" vs "queued; target
busy, will read on current turn") so the printed line at `cli.py:1110` tells the
truth. This is a feedback-timeliness gap, criterion (2).

---

**U5 — MINOR — `--now`/`--wake` are silently accepted together with no precedence
signal.** `lop send --pid X --wake --now "..."` is accepted; `mode` becomes `steer`
(`cli.py:1094`) and `--wake` is silently ignored (the guide notes wake is ignored in
steer mode, GUIDE.md:70 / design §1.2). A user combining them has a mental-model
conflict — "wake AND steer" — and gets no hint that one flag was dropped. Minor
because the resolved behavior is defensible, but a one-line note or a mutually-
exclusive argparse group would prevent a surprised user. Consider
`add_mutually_exclusive_group()` or a stderr note when both are passed.

---

**U6 — NIT — `lop sessions` truncates CONVERSATION and MODEL to 24 chars with no
ellipsis, so names collide visually.** Real output showed
`Provider Retry Exponenti`, `Local Operator browser e`, `anthropic/claude-opus-4-`
— hard truncation (`cli.py:1164-1165` `[:24]`) drops the tail with no `…`, and the
model column loses the disambiguating suffix (`-opus-4-8` vs `-opus-5` both render as
`claude-opus-4-`/`claude-opus-5` — actually `opus-4-8` truncates to `claude-opus-4-`,
indistinguishable from a hypothetical `opus-4-1`). Since PID is the real key this is
cosmetic, but an ellipsis on truncation (and widening MODEL, or right-trimming the
`anthropic/` vendor prefix) would make the table scannable. `--json` carries the full
values, so scripting is unaffected.

---

**U7 — NIT — stdin-body discoverability is thin.** Reading the body from stdin
(`cli.py:1079`) is a genuinely nice ergonomic (`git log | lop send "release cutter"`),
and it *is* in the guide (GUIDE.md:53, 107). But `send --help` only says
"message text; omit to read the body from stdin" on the positional — a user who runs
`lop send "target"` at an interactive TTY with no body gets the correct
"pass it as an argument or pipe it on stdin" error (verified), which is good. Minor
polish: the `--help` epilog could show the pipe example, since the positional help
line is easy to miss. Not blocking; the error message already teaches it.

---

## Positive notes (working well)

- Target-resolution error copy is genuinely good and *actionable*: distinct messages
  for no-match (`no live session matches 'X'`), dead pid (`no session found with pid
  N`), wedged-only match (`the only match ... is not responding (pid N); try again
  shortly`), and ambiguous (`N sessions match; disambiguate with --pid:` followed by
  a copy-pasteable `--pid` list). Verified each live. This is the right shape.
- The `--pid` / `--session` / positional precedence is intuitive and matches the
  guide.
- `lop sessions --json` emits full, well-typed records (bytes not human sizes) —
  correct for scripting.
- The RECEIVE experience is well-judged: the rendered TUI indicator frame
  (`/tmp/peer-send-tui-indicator.png`) shows a clear, non-disruptive
  `↔ peer message from "<name>" (pid N, <model>)` card that reads as inbound and
  distinct from the user's own turns, and record-only correctly leaves an idle
  session idle. The `↔` glyph and the sender attribution are exactly right for
  "someone else put this here."
- `FOOTPRINT` vs `RSS` distinction with `—` for unmeasured (never zero) is honest and
  the guide explains it well; `STATE`/`HB_AGE` are documented and the `HB_AGE`-as-
  early-wedge-signal framing is a nice touch.

---

## Verdict

**NOT TERMINAL.** One BLOCKER (U1: sender crashes with a raw traceback on any
real-sized target while the message is delivered anyway) and one MAJOR (U2: no
self-send guard) must be fixed before this round can close. U1 in particular means
the feature does not survive contact with a normal session; the PR's live test
passed only because it targeted a small transcript. Once U1 and U2 are remediated and
re-verified against a large-transcript target, the remaining U3–U7 are follow-ups
that need not block. Re-request a UX round on the fixed head.

damianvtran:
### Agent review — round 1
Scope: origin/main..3cd9bfdfd899b8190eeb4ff022b68fa7fa706d9b

Reviewed against the design at `docs/design/peer-send.md`. I ran the gates on this head: flake8, isort (`--check .`), black `26.1.0 --check .`, and pyright over every changed module are all clean. The peer test modules pass **except** the new TUI file (see C2). I reproduced the two highest findings rather than inferring them.

Verdict summary up front: no BLOCKER. Two MAJORs, both with a one-change remediation.

---

**C1 — MAJOR — `local_operator/session/session.py:2949` (busy check) + `:2984-2986` (direct append).** The record-only branch appends a peer message straight onto the live `_context.messages` while a turn is streaming, bypassing the purpose-built `_append_or_park_journal` guard (`session.py:3902`). This is exactly the hazard the splice-repair docstring warns about in its own words: *"a future writer appending straight to `_context.messages` would reintroduce the same 400"* (`session.py:513-514`), and *"If a custom type is ever added that carries user-authored text, this assumption needs revisiting — moving a user's words past a tool batch would silently reorder the conversation they see"* (`:540-542`). A peer message is user-authored text, and this is that writer.

I reproduced it. With a live list ending mid-tool-batch (`assistant(tool_calls) → peer(user) → tool_result`), `_pair_spliced_tool_results` fires its regression alarm and reorders the peer text past the batch:
```
WARNING repaired a message spliced into an open tool batch; a journal append bypassed the turn-boundary guard (see _append_or_park_journal)
in : [('assistant',None,True), ('user',None,False), ('tool','c1',False)]
out: [('assistant',None,True), ('tool','c1',False), ('user',None,False)]
```
Why this is MAJOR and not a BLOCKER: the render-time repair (`session.py:1524`) runs on every provider call, so the request is always legal and the session does **not** brick — the PR #302 bug is caught. But three real costs remain: (1) the regression WARNING that exists specifically to flag "the splice came back" now fires on **every busy mailbox delivery**, permanently, masking the real regression it was meant to catch; (2) the live list stays illegal forever (nothing repairs the live list, only the rendered copy — see `session.py:1364-1372`), so the repair re-runs every subsequent turn; (3) the busy path is **untested** — `test_session_peer.py` covers idle record-only, wake-idle, and busy-steer, but not busy mailbox.

Second half of the same finding: the guard uses `busy = self._is_streaming` only. Every sibling that appends to the live context guards on `self._is_streaming or self._turn_lock.locked()` (`_append_or_park_journal` :3930, `record_shell` :5247, `adopt_aside` :5274, :4511). During an on-demand compaction the lock is held while `_is_streaming` is False (`prompt` :2714, `_run_compaction` reassigns the list at :4854), so the peer path treats a compacting session as idle and appends into a list compaction is about to rebuild.

Remediation (one change, matches every sibling): in the record-only branch replace `self._context.messages.append(message)` with `self._append_or_park_journal(message)`, which parks while `_is_streaming or _turn_lock.locked()` and is drained at the existing turn-safe boundaries (`_flush_context_journal` at :2733/:3448/:3503). The transcript append stays eager as-is. Then add a busy-mailbox test asserting no splice-repair warning fires and the row still reaches the model on the next turn.

**C2 — MAJOR — `tests/unit/tui/test_peer_message.py` (all three tests).** The new TUI tests are nondeterministic: each emits a `PeerMessageDeliveredEvent` then does a **single** `await pilot.pause()`, but delivery is a two-hop post (`session.emit` → `EventController._post` → app message queue → `on_peer_message_delivered` → `_append_block` → `call_after_refresh`). One pause does not reliably settle it. Measured, run in isolation:
```
test_live_peer_delivery...:   fail, fail, fail, pass
test_live_receipt_suppresses: fail, pass, pass, fail
```
and in the module together I saw `2 failed`, `1 failed`, `3 passed`, `3 failed` across runs. This will cause spurious CI failures on unrelated PRs and undermines the gate. Root cause confirmed: an identical test with `for _ in range(5): await pilot.pause()` passes 5/5. Remediation: replace the single pause after each emit/`_render_resumed_history` with a settle loop (or `await pilot.pause()` twice — the wake tests these were modeled on settle reliably; match whatever they do). Not production code, but a 40%-flaky gate test is worse than no test.

---

**C3 — MINOR — `local_operator/mobile/peer_client.py:63`.** `deadline_s` is applied per `readline`, not to the whole exchange. A target that streams `projection` pushes more often than every 5s keeps the loop alive indefinitely instead of honoring a total deadline. Low risk (the registrant acks synchronously right after dispatch, so an ack normally arrives promptly), but a chatty busy target defeats the timeout. Consider an overall deadline (`loop.time() + deadline_s`, recompute per-readline remaining).

**C4 — MINOR — `local_operator/cli.py:1114` `sessions_command`.** `session_resource_usage` is only called for `live` pids, so `wedged`/`stale` rows always print RSS/FOOTPRINT as `—`. That is defensible (a wedged owner may not be worth probing), but `ps -o rss=` reads the kernel's process table and works fine for a wedged-but-alive pid — those rows could show real memory. Minor: consider probing every non-stale pid. Not wrong as written.

**C5 — MINOR — `local_operator/session/session.py:2984-2986` idle record-only ordering vs. a concurrent wake.** Even after the C1 fix, note that the eager `await self._transcript.append_message(message)` yields the loop before the context append; if a wake turn is spawned in that window the ordering of transcript vs context is not atomic. Almost certainly benign (single loop, narrow window, transcript is the durable source of truth), but worth a one-line comment stating the invariant you rely on, per the repo's why/constraints comment standard.

---

**N1 — NIT — `local_operator/cli.py`.** `_resolve_peer_target` re-imports `os as _os` locally while the module already imports `os` at top and `_peer_sender_identity` uses the top-level `os`. Inconsistent; drop the local alias.

**N2 — NIT — `local_operator/mobile/resources.py:192` `_parse_top_footprint`.** Doc says the MEM column is `parts[1]`, which assumes `-stats pid,mem` always yields exactly two whitespace-split columns. Robust here because you pass `-ncols 2`, but a one-line note that the parse depends on the `-stats`/`-ncols` pairing staying in sync would save a future maintainer a surprise.

---

Verdict: **not clean** — 2 MAJOR (C1 concurrency/invariant, C2 flaky gate tests), 0 BLOCKER. Neither is stop-the-world: C1's catastrophic case is caught by the existing render-time repair, but it trips the regression alarm on every busy delivery, leaves an untested busy path, and skips the guard every sibling uses; C2 is test-only but will flake CI. Both have a small, well-precedented fix. Please address C1 and C2 and re-post remediation for round 2; C3–C5 and the nits are your call.

damianvtran:
Correction to C1 line refs: the record-only direct append is at `session.py:2973-2974` (`await self._transcript.append_message(message)` / `self._context.messages.append(message)`), not `:2984-2986`. The `busy = self._is_streaming` guard at `:2949` and the remediation (swap the `:2974` append for `self._append_or_park_journal(message)`) are unchanged.

damianvtran:

Implementer: coder. Head: `12362430cfab4c8c2bcd89ea4f5613cecf888816`. Each BLOCKER/MAJOR finding below is reproduced-then-fixed (the regression test fails against `HEAD~5`'s pre-fix code and passes on the fix); deferred items are marked Deferred with a follow-up note.

**U1 — Fixed — `7adf5912`.** Replaced `reader.readline()` (which raises `ValueError: Separator is not found, and chunk exceed the limit` on an oversized welcome and leaves its buffer un-consumed) with a small newline-frame reader built on `reader.read()`, which has no line-length ceiling. It frames lines from its own buffer, DISCARDS an oversized welcome (memory-bounded at 8 MiB), and crucially keeps any bytes the kernel coalesced behind the welcome's newline — so the ack that arrived in the same TCP read is never thrown away. The sender now reliably reads its actual `ack` instead of crashing. The CLI also catches `ValueError` as a soft error (`305c288e`). Chose the robust option (b) over raising the limit, since projections are unbounded and no fixed limit can guarantee the welcome fits — justified in a comment on `_FrameReader`. Regression test `test_send_survives_an_oversized_welcome_projection` seeds a ~1.6 MiB welcome projection (80 rows × 20 KB, the projection cap) and asserts the send returns its ack cleanly; confirmed it fails on the old reader with the exact U1 `ValueError` and passes on the fix.

**C1 — Fixed — `d831ebd1`.** The record-only branch of `receive_peer_message` now routes the live append through the existing `_append_or_park_journal` instead of a bare `_context.messages.append`. The transcript write stays immediate (the human sees the row now), and the live append parks to the next turn-safe boundary on the busy path (never splicing between an assistant's `tool_calls` and their `tool_results`); on the idle path it appends straight through, matching prior behaviour. Regression test `test_record_only_while_busy_parks_to_the_turn_boundary` asserts: while a tool batch is open the message is parked (not in the live list) yet the transcript row is written exactly once; after the boundary it drains into live context; and the next turn reads it as model input. Confirmed it fails on old code (nothing parked) and passes on the fix.

**C2 — Fixed — `db36495c` + `12362430`.** Replaced the single `pilot.pause()` after each two-hop post with a bounded settle loop (`_settle_for_peer_block`) that polls for the block to appear, matching the app-pilot tests' discipline. 8/8 passes in isolation and 5/5 under xdist. Widened the bound to 200 (`12362430`) after one flake under whole-suite CPU load — the loop exits the moment the block mounts, so the headroom costs nothing.

**U2 — Fixed — `305c288e`.** Added a self-send guard in `send_command`: when the resolved target pid equals `os.getppid()` (the launching session's pid), refuse with "that target is this session; use the composer to message yourself" before any network call. Chose refuse-over-self-note so a card is never mislabeled as inbound-from-another-session. Test `test_self_send_is_refused_before_any_network_call` asserts the guard fires on the self pid (no `send_peer_message` call) and not on a different pid; confirmed it fails on old code (silently delivered, rc=0) and passes on the fix.

**Deferred to follow-up (minor/nit, per the round scope):**
- **D1** (lean-sender dangling "from"), **D2** (mobile "peer message from" parity), **D3** (peer-card accent distinct from parent), **D4** (`text-ink-muted` for legibility), **D5** (blank-body guard note).
- **U3** (cap-vs-failure wording in the guide — reconciles once U1 is fixed), **U4** (`--wake`-while-busy no-op feedback), **U5** (`--now`+`--wake` precedence signal), **U6** (`lop sessions` ellipsis), **U7** (stdin discoverability).
- **C3** (per-readline vs total deadline), **C4** (probe wedged-but-alive pids for memory), **C5** (idle record-only ordering invariant comment).
- **N1** (drop the local `os as _os` alias), **N2** (`-stats`/`-ncols` parse-sync note).

These are minor/nit-severity and none gate the merge; bundling them into a focused follow-up keeps this remediation scoped to the round's blockers/majors.

**Gates** (from the worktree venv, all clean on `12362430`): flake8 exit 0; black 26.1.0 `--check` 501 unchanged; isort `--check-only --profile black` clean; pyright `0 errors, 0 warnings`; peer modules 11 passed; TUI peer module 3 passed (8/8 isolation, 5/5 xdist); session+mobile suites 579 passed.

damianvtran:
### Agent review — round 2
Scope: origin/main..12362430cfab

Verification re-review of the round-1 fixes on head `12362430cfab`. All four round-1 findings are verifiably fixed, and the remediation introduced nothing new. Gates and regression tests run clean below. **Round is CLEAN (0 blocker, 0 major) and TERMINAL.**

**U1 (blocker) — oversized welcome frame — FIXED.** `_FrameReader` (`peer_client.py:34-93`) frames lines over `reader.read()` with no line-length ceiling and the connection no longer passes a `limit=` (`open_connection`, line ~125). The discard is genuinely memory-bounded: the `len(self._buf) > _MAX_FRAME_BYTES` (8 MiB) branch only fires when *no* newline is buffered yet, so the buffer never grows past ~8 MiB + one 64 KiB read chunk before it either frames a line or clears and enters `_skipping`. The coalesced-ack case is correct: `del self._buf[: nl + 1]` removes through the welcome's newline and *keeps* trailing bytes, so an ack the kernel packed into the same TCP segment survives; `_skipping` is cleared on the terminating newline and the loop re-scans for the ack. There is no path where the ack is discarded — discard only ever drops the un-terminated prefix of a single oversized line, never bytes past its newline. Per-read `wait_for(..., timeout)` matches the old per-readline deadline in the same `while` loop, so no timeout regression. The regression test seeds an ~1.6 MB welcome (80 rows × 20 KB, at the projection cap) that crashed the old reader and asserts a clean ack — it genuinely exercises the oversized-welcome-before-ack path.

**U2 (major) — self-send guard — FIXED.** `cli.py:1081` refuses when `record.pid == os.getppid()` with an honest message ("that target is this session; use the composer to message yourself"), ahead of the network dial. `os` is imported at module top (`cli.py:39`). A non-child/headless caller is not a false positive: its `getppid()` is the shell/launcher, which is not a registered session pid, so the guard cannot match a real peer record. Both tests assert the guard fires on the self pid and passes a different pid through to delivery.

**C1 (major) — record-only busy splice — FIXED.** `session.py:2984` now routes the live append through `_append_or_park_journal` instead of a bare `_context.messages.append`. That method (line 3912) parks to `_pending_context_journal` when `_is_streaming or _turn_lock.locked()` and appends directly otherwise, so the busy path parks to the next turn-safe boundary while the transcript write at line 2983 stays immediate — the record-only guarantee holds. The regression test opens a real tool batch, asserts the peer message is parked (not in live context) with exactly one transcript row, then after the boundary asserts it drained into live context (still one row) and reached the model on the next turn's request. Verified by running it.

**C2 (major) — flaky TUI tests — FIXED.** Fixed pause replaced with a 200-iteration bounded settle loop matching `test_app_pilot`. Ran `test_peer_message.py` 5× in isolation: **5/5 passes** (3 passed each run).

### Gates on head `12362430cfab`
- flake8: clean
- black 26.1.0 --check: clean (7 files unchanged)
- isort 5.13.2 --check --profile black: clean
- pyright: 0 errors, 0 warnings on `peer_client.py`, `cli.py`, `session.py`
- peer suites (`test_peer_client.py`, `test_cli_send.py`, `test_session_peer.py`, `test_peer_message.py`): **14 passed**

No remaining findings. No new findings from the remediation. Minors/nits: none worth a round.

damianvtran:
### UX review — round 2
Reviewer: ux-reviewer (anthropic/claude-opus-4-8)
Scope: origin/main..12362430cfab4c8c2bcd89ea4f5613cecf888816

Independent verification pass (not the coder). I re-walked the two round-1 UX
findings on the current head with real flows, not stills: `lop sessions` against
the live registry, a real `lop send` to a live large-footprint session, an
end-to-end registrant+sender integration harness over a real loopback socket to
force the discard path, and a genuine parent/child spawn to exercise the
self-send guard where `os.getppid()` really is the target. Both round-1
findings are verifiably fixed. Nothing new. **Round is CLEAN.**

---

**U1 (blocker) — FIXED & VERIFIED.** Oversized-welcome crash is gone; the sender
returns a clean human-readable result on every path.

- Real dial: `lop send --pid 74800` (the live "Work Queues UX and Logic" TUI,
  945 MB RSS / 1 GB footprint — the exact large-transcript target class that
  crashed in round 1) no longer dumps a `ValueError: Separator is not found`
  traceback. It read straight through that session's real oversized welcome
  projection and returned a clean soft line (`could not deliver: unknown op:
  'peer_message'`, because that live session runs the installed runtime without
  the new op — so no stray probe landed, as intended). The old `readline` path
  would have crashed before ever reaching the registrant's error frame. No
  traceback, non-zero exit, one clearly-labeled probe only.
- Discard-branch proof (the case the unit test does NOT reach): I stood up a
  real `Registrant` + real `send_peer_message` over loopback with a ~12 MB
  welcome projection (80 rows × 150 KB), **past** the 8 MiB `_MAX_FRAME_BYTES`
  cap, forcing `_FrameReader` into its `buf.clear()` + `_skipping` branch. The
  coalesced ack after the welcome's newline was still recovered cleanly:
  `delivered to the mailbox (will be read on the next turn)`, and the handle
  received `receive_peer_message("probe past the 8MiB cap")`. The ack is never
  discarded because TCP ordering guarantees ack bytes arrive only after the
  welcome's terminating `\n`, and the cap-clear fires only while no newline is
  buffered — so cleared bytes are pure welcome body.
- Reader logic is memory-bounded (8 MiB single-frame ceiling, drops rather than
  raises), and the CLI now also catches `ValueError` as a soft error, so any
  residual read fault becomes the same "could not deliver" line, never a
  traceback. Regression test passes; `test_peer_client.py` 4/4 green.

  Non-blocking follow-up (carry as U8, MINOR): the U1 regression test seeds only
  a ~1.6 MiB welcome — over the old 1 MiB `readline` limit (so it reproduces the
  crash) but UNDER the new 8 MiB discard cap, so the automated test exercises the
  "line longer than old limit" path but not the actual `clear()`/`_skipping`
  discard-with-coalesced-ack branch. That branch is real and I verified it
  manually above; a future test that pushes past 8 MiB would lock it in. Not a
  blocker — the fix is correct and proven — just thinner automated coverage than
  the manual walk.

**U2 (major) — FIXED & VERIFIED.** Self-send is refused cleanly, before any
network call, with an honest message; non-child callers still send.

- Real parent/child walk: a parent process published a registrant record, then
  spawned `lop send --pid <own-pid>` as a child so `os.getppid()` == the target
  record's pid (the true self-send shape). The child exited 1 with
  `that target is this session; use the composer to message yourself` and
  `receive_peer_message` was never invoked — the guard sits ahead of the dial.
- Legitimate cross-session send is unaffected: my U1 probe to pid 74800 (where
  `getppid()` was the shell, not the target) passed the guard and proceeded to
  delivery. A headless/exec caller whose `getppid()` is not a session pid is
  therefore not blocked. `test_cli_send.py` 2/2 green.

**U3–U7 — still deferred, NOT regressed.** The two fix commits are narrow
(`_FrameReader`, the self-send guard, the `ValueError` catch) and touch neither
the `lop sessions` table (U6 truncation), the `--wake`-while-busy no-op feedback
(U4), the `--now`/`--wake` precedence (U5), the body-cap copy (U3), nor the
stdin-body discoverability (U7). `lop send --help` still shows the same flag
surface and the "omit to read the body from stdin" hint. All remain acceptable
MINOR/NIT follow-ups.

---

**Verdict: TERMINAL / CLEAN.** No BLOCKER, no MAJOR. Round-1 U1 and U2 are fixed
and independently verified on real flows on the current head (12362430). Carry
forward as follow-ups: U3, U4, U5, U6, U7 (unchanged), plus U8 (widen the U1
regression test past the 8 MiB discard cap). Not approving or merging.

damianvtran:

The branch was rebased from the reviewed `0.34.0` base onto current `origin/main` (`0.35.1`) before merge, because browser-extension and TUI-responsiveness releases landed while this PR was in review. The version is now `0.36.0`.

- **CLI integration conflict — Fixed** in `d993705a`: the latest base adds `lop browser`; this feature adds `lop send` / `lop sessions`. The conflict was purely additive. All three parser blocks, command functions, and dispatch branches now coexist. `cli.py` parses and pyright reports 0 errors.
- **TUI peer tests after asynchronous-boot change — Fixed** in `d993705a`: the `0.35.1` responsiveness work deliberately makes session boot more asynchronous. The peer tests could emit a host event before the session bridge subscribed, so the event was lost. Tests now wait for `app._session` before injecting, matching the real registrant's precondition (a session is not discoverable/addressable until ready). The three peer tests pass 5/5 in isolation.
- **Version collision — Fixed** in `c2f5c50d`: `v0.35.0` and `v0.35.1` were published while this PR was in review, so the feature now bumps `0.35.1` to `0.36.0`.

Rebase verification on head `d993705a`: flake8 clean; black 26.1.0 clean (515 files); isort clean; pyright 0/0; 55 peer backend/CLI tests pass; TUI peer tests pass 5/5. No runtime visual surface changed in the rebase adaptation.

damianvtran:
### Agent review — round 3
Scope: origin/main..d993705aa3b04ac1d7ab4644bb4dc7f318635827

Final freshness review after rebasing onto current `origin/main` (0.35.1).

Findings C6+: none.

The fresh 31-file diff is intact, with no conflict markers, duplicate symbols, or garbled rebase residue. `cli.py` contains exactly one parser block, command function, and main dispatch branch for each of `browser`, `send`, and `sessions`; all three reviewed CLI entrypoints parse successfully. The additive conflict resolution preserves the base browser command and the peer feature together without introducing eager imports or startup regressions.

The previously-fixed **C1**, **U1**, **C2**, and **U2** remain fixed after rebase:
- **C1:** record-only peer delivery still uses `_append_or_park_journal`; the real busy tool-batch regression passes.
- **U1:** `_FrameReader` and its oversized-frame handling are unchanged from the reviewed fix; the >1 MiB welcome test passes.
- **C2:** the peer TUI file passes 5/5 runs (3 tests each) under `env -u NO_COLOR TERM=xterm-256color`.
- **U2:** the self-send guard commit is patch-identical to the reviewed version.

Waiting for `app._session` is the correct host-event precondition, not a runtime workaround: `_adopt_session` publishes `self._session`, starts the registrant, then installs the `EventController` subscription synchronously before yielding. Before adoption there is no addressable session and no subscribed event bridge through which a real peer can deliver.

Verification on this exact head:
- flake8: clean
- black 26.1.0: 538 files unchanged
- isort 5.13.2: clean
- pyright on all changed Python modules: 0 errors, 0 warnings
- peer-focused suites: 187 passed
- busy-path C1 test: passed independently
- >1 MiB U1 test: passed independently
- TUI peer stress: 5/5 clean
- `pyproject.toml`: 0.36.0; no local or remote `v0.36.0` tag collision

**Verdict: CLEAN/TERMINAL.** No BLOCKER or MAJOR findings remain. This round is terminal; any future minor or nit would be follow-up work, not a reason for another review round.

damianvtran:

Reviewer flagged a Python 3.12 hang isolated during final-head CI verification, filed as **C6 (BLOCKER)**. On investigation the blocker is a **test-only defect**, not a production regression. Fixed at `dec897252ae799207ffbac6b9bab53d3ed0aed73`.

**C6 (BLOCKER) — silent-socket test hangs on Python 3.12 → FIXED (`dec89725`), reclassified TEST-ONLY**

Reframing: this is a defect in `test_send_times_out_on_a_silent_socket`, not in `send_peer_message`. The production client was verified correct on **both** interpreters and is unchanged.

**Root cause (measured, not assumed).** The hang is in the test's teardown, not the client's cleanup:

- The client's `writer.close(); await writer.wait_closed()` returns in ~0.1 ms against this peer — closing a local TCP socket does not wait for the peer's FIN — and `send_peer_message` already returns at its `deadline_s` (measured 0.202 s for a 0.2 s deadline) on both 3.12 and 3.14.
- The test held its handler open with `await asyncio.sleep(1.0)` and used `async with server:`. On CPython 3.12, `asyncio.Server.wait_closed()` (awaited on `async with` exit) blocks until every active handler task finishes. The still-sleeping handler therefore wedged teardown and hung the whole pytest process. Python 3.14 does not gate `Server.wait_closed()` on active connections, which is why the same test passed there.

Isolated proof, 0.2 s deadline, silent handler holding the socket open:
```
3.12.13  send raised TimeoutError at 0.202s   # client is fine
3.12.13  server.wait_closed()  HANGS >10s     # the actual hang
3.14.3   server.wait_closed()  OK  0.876s      # 3.14 does not block
```

Real half-open peer (server accepts but never reads, client write fills the peer window) was also checked to confirm the client path itself is sound — no unbounded client wait was observed that a production change would be needed to guard. Since the production client honours its deadline on both interpreters, changing `peer_client.py` would paper over a test bug in the runtime, so **no production code was touched** in this round.

**Fix.** Test-only. `tests/unit/mobile/test_peer_client.py`:
- The handler now parks on an `asyncio.Event` (`stop`) instead of `asyncio.sleep(1.0)`. This keeps the socket open for the whole time the client is reading (so the client genuinely times out and never sees an EOF → `ConnectionError`), yet the handler returns the instant teardown sets the event, so the server can close cleanly.
- Teardown sets `stop`, then `server.close()` and a **bounded** `asyncio.wait_for(server.wait_closed(), 1.0)` (TimeoutError swallowed) instead of `async with server:`, so teardown can never block the run.
- The client call stays wrapped in an outer `asyncio.wait_for(..., 1.5)` and an elapsed-time assertion (`< 0.9 s`), so any future client-side regression that ignores `deadline_s` is still caught as a hard failure rather than a hang.

**Reproduce-before / pass-after.**

Before (this branch head `d993705a`, Python 3.12) — hangs indefinitely, process never exits:
```
env -u NO_COLOR TERM=xterm-256color /tmp/lop-peer-py312/bin/pytest \
  tests/unit/mobile/test_peer_client.py --cov=local_operator -o addopts="" -vv
# ... test_send_times_out_on_a_silent_socket  <hang, killed>
```

After (`dec89725`), Python 3.12 — all 4 pass, **process exits (RC=0)**:
```
env -u NO_COLOR TERM=xterm-256color /tmp/lop-peer-py312/bin/pytest \
  tests/unit/mobile/test_peer_client.py --cov=local_operator --cov-fail-under=0 \
  -o addopts="" -vv
tests/unit/mobile/test_peer_client.py::test_send_returns_ack_detail_and_passes_args PASSED [ 25%]
tests/unit/mobile/test_peer_client.py::test_send_survives_an_oversized_welcome_projection PASSED [ 50%]
tests/unit/mobile/test_peer_client.py::test_send_raises_on_error_frame PASSED [ 75%]
tests/unit/mobile/test_peer_client.py::test_send_times_out_on_a_silent_socket PASSED [100%]
============================== 4 passed in 15.19s ==============================
RC=0
```

After (`dec89725`), Python 3.14 worktree venv — passes:
```
.venv/bin/python -m pytest tests/unit/mobile/test_peer_client.py -q
4 passed
```

**Gates (both files):** flake8 clean; `black 26.1.0 --check` unchanged; `isort --profile black --check-only` clean; pyright 0 errors.

No blocker/major findings remain open. Reviewed head for this round: `dec897252ae799207ffbac6b9bab53d3ed0aed73`.

damianvtran:
### Agent review — round 4
Scope: origin/main..dec897252ae799207ffbac6b9bab53d3ed0aed73

Final freshness review at head `dec89725`, closing the C6 CI hang isolated after the round-3 CLEAN at `d993705a`.

**Scope of the single new commit.** `git diff d993705a..HEAD --stat` shows exactly one file changed: `tests/unit/mobile/test_peer_client.py` (+47/-7). `peer_client.py` is byte-identical to the round-3 head — the production client was proven correct on both interpreters and was not touched. This is a test-only fix.

**The fix is legitimate, not an assertion weakening.** The old handler parked on a fixed `asyncio.sleep(1.0)`, so on CPython 3.12 `Server.wait_closed()` (which awaits outstanding handler tasks) wedged teardown and hung the whole job; 3.14 does not gate teardown on active connections, which is why only 3.12 hung. The new handler parks on a `stop` Event the teardown sets, keeping the socket open while the client times out but returning promptly once signalled; teardown releases the event, closes the server, and bounds `wait_closed()` with `wait_for(..., 1.0)`. Crucially the behaviour under test is preserved and tightened: the call is still wrapped in `pytest.raises(asyncio.TimeoutError)`, and an outer `asyncio.wait_for(..., 1.5)` hang-catcher is backstopped by `assert elapsed < 0.9`. A client that regressed to ignoring its `deadline_s=0.2` would drive elapsed toward the outer guard and fail that assertion — so a real client-side timeout regression still fails the test.

**Verification on head.**
- Python 3.12 (`pytest ... --cov-fail-under=0 -o addopts="" -q`): 4 passed in 19.22s, process exits cleanly (no hang).
- Python 3.14 (`.venv/bin/python -m pytest -q`): 4 passed in 5.82s.
- Gates on the changed test file: flake8 clean; `black 26.1.0 --check` clean; isort (`--profile black`) clean; pyright `0 errors, 0 warnings` via the AGENTS.md `--pythonpath .venv/bin/python` invocation.
- C1/U1/C2/U2 remain fixed: their files did not change in `dec89725` (only the test file changed since the round-3 CLEAN head), so they are unaffected.

**Findings (C7+):** none.

**Verdict: CLEAN / TERMINAL.** C6 was correctly resolved as a test-only defect with production code (`peer_client.py`) unchanged; all prior findings remain fixed and the round is fresh against the current head `dec89725`. No further review round is warranted.

PR #312, opened 2026-08-26
fix(extension): give the popup explicit feedback for pairing and origin decisions

damianvtran:
## Summary

Fixes two instances of the same popup defect class, both reproduced live: an async round-trip between a user action and the view it should change leaves the OLD view on screen with live controls, so the user cannot tell the action worked.

1. **Pairing**: on a successful pair the popup gave no feedback — the success path re-rendered from `/health`, which reports `paired` only after the background worker reconnects with the new token. Losing that race re-showed the form; the user re-submitted the consumed one-time code and got a misleading "No live pairing code" error.
2. **Origin consent**: clicking **Allow once / Always allow / Deny** on "Let the agent use this site?" changed nothing — session storage and `/health` echo the prompt until `origin_decision` lands in the worker, so the prompt stayed up with all three buttons active while the agent-side open had already succeeded.

## What changed

Both fixes follow one pattern: render the acknowledgement **from the user's action alone**, latch it so a stale echo cannot downgrade it, disable the controls for the flight, and let the normal `render()` flow take over once the daemon confirms.

- **`extension/src/popup/pair-flow.ts`** (new) — pure `pairVerdict` / `viewForHealth`; DOM-free so the pure-node suite can cover the race (`popup.ts` touches `document` at module scope).
- **`extension/src/popup/origin-flow.ts`** (new) — pure `ackForDecision` / `originPromptView`, same pattern. The latch is keyed by ORIGIN so a *different* pending origin still prompts mid-settle (finding A6 keying untouched), and it clears when the echo lands so a same-origin retry re-prompts. Deny renders neutral, not danger — a completed choice, not a failure.
- **`popup.ts`** — `show("paired")` immediately on `pair_result.ok`; `showOriginAck(decision)` immediately on any consent click; `locallyPaired` / `decidedOrigin` latches consulted by `render()`; `setPairBusy` / `setOriginBusy` disable controls in flight ("Pairing…" relabel; consent buttons unlock when a fresh prompt renders). Error path unchanged.
- **`popup.html`** — `#paired` and `#origin-ack` sections (`role="status"`): success-tone rule + `--success` check for successes; "Paired." / "Site allowed. The agent is continuing." (+ standing-grant/Settings note for Always allow) / "Site denied." microcopy. Header mark stays untinted.
- **`popup.css`** — `:disabled` ramps (input drops to the sunken well; buttons at 0.6 opacity, primary keeps its fill), `.check` sizing/tone.
- **`tests/pure.test.mjs`** — 4 new tests: pair verdict folding, pair/health race hold, per-decision acks (deny neutral, always-allow revocability copy), origin race hold + A6 cross-origin + latch-clear cases.

## Testing evidence

- `pnpm typecheck` clean; `pnpm test` 14/14 (4 new); `pnpm build` clean, dist carries both states.
- **Render verification** of the built dist in a real browser (script stripped, target state unhidden — `chrome.*` doesn't exist outside an extension), five states screenshotted and reviewed by eye:
  - *Paired.* — success rule, green check, serif title, untinted mark ("Code accepted. Connecting this browser to Local Operator…").
  - *Pairing in flight* — input locked to the sunken well retaining the code, primary button dimmed and relabelled "Pairing…".
  - *Site allowed (Always allow)* — success rule, check, "The agent is continuing. Always-allowed sites can be taken back any time in Settings."
  - *Site denied* — neutral rule, no check, "The agent won't use this site."
  - *Consent prompt locked* — all three buttons dimmed, Deny keeping its danger outline.

  Screenshots at `/tmp/popup-verify/*.png` on the dev machine (gh CLI cannot attach images — drag them onto the PR if a visual record is wanted). Reproduce any state: `pnpm build`, open `dist/popup/popup.html` with the `<script>` tag removed and the target section's `hidden` class dropped.
- Not exercised live end-to-end: a fresh pair (this machine's extension is already paired) and a live consent click (needs an unpaired-origin agent open). The race decisions themselves are pinned by the pure tests.

## Review

Round 1 (scope 04702e6e..005cdf46): 2 major, 3 minor, 2 nits — all remediated in 480256d8 (see remediation comment). Round 2 pending on the current head.

### DIFF
--- a/docs/design/branding-bridge.md
+++ b/docs/design/branding-bridge.md
@@ -118,7 +118,7 @@ Pairing prompt:
 - Heading: **"Pair with Local Operator"**
 - Body: "Type the 6-digit code from the Local Operator app. Pairing links this
   browser to your app and nothing else."
-- Field label: "Pairing code" · Button: "Pair" · Error: "That code didn't match. Codes expire after a minute — check the app for a fresh one."
+- Field label: "Pairing code" · Button: "Pair" · Error: "That code didn't match. Codes expire after two minutes — check the app for a fresh one."
 
 Per-site allow prompt:
 - Heading: **"Let the agent use github.com?"**
--- a/extension/src/popup/origin-flow.ts
+++ b/extension/src/popup/origin-flow.ts
@@ -1,0 +1,66 @@
+/* Pure decisions for the popup's origin-consent flow, DOM-free for the
+ * pure-node test suite — the same pattern as pair-flow.ts, for the same class
+ * of bug.
+ *
+ * The race: clicking Allow/Deny sends `origin_decision` to the worker, but
+ * session storage's `pendingOrigin` and /health's `pending_origin` keep
+ * reporting the prompt until that round-trip lands. A render() driven by
+ * either source would leave the prompt on screen with three live buttons and
+ * no sign the click took effect. The acknowledgement must therefore render
+ * from the CLICK alone, and a stale prompt echo must hold the ack rather than
+ * resurrect the buttons. */
+
+export type OriginDecision = "once" | "always" | "deny";
+
+/** The decision this popup already made, keyed by ORIGIN — finding A6's rule:
+ * a redirect chain resolves each hop independently, so a DIFFERENT pending
+ * origin is a new prompt even while this one's ack is still settling. */
+export interface DecidedOrigin {
+  origin: string;
+  decision: OriginDecision;
+}
+
+export interface DecisionAck {
+  title: string;
+  sub: string;
+  tone: "success" | "neutral";
+  check: boolean;
+}
+
+/** The acknowledgement each decision renders. Deny is a COMPLETED choice, not
+ * a failure, so it takes the neutral register and no check — danger is
+ * reserved for states the user must recover from (error/incompatible), and a
+ * check over "denied" would read as the wrong verdict. */
+export function ackForDecision(decision: OriginDecision): DecisionAck {
+  if (decision === "deny") {
+    return {
+      title: "Site denied.",
+      sub: "The agent won't use this site.",
+      tone: "neutral",
+      check: false,
+    };
+  }
+  // "always" is a standing grant, so its ack must say it outlives this moment
+  // and where to take it back — the consent prompt's own revocability promise.
+  const standing =
+    decision === "always" ? " Always-allowed sites can be taken back any time in Settings." : "";
+  return {
+    title: "Site allowed.",
+    sub: `The agent is continuing.${standing}`,
+    tone: "success",
+    check: true,
+  };
+}
+
+/** Which origin view a render shows. A pending origin equal to the one just
+ * decided is the stale echo of the race above — hold the ack. A different
+ * pending origin is a genuinely new prompt (A6). No pending origin means the
+ * round-trip landed and the caller should clear its latch so a future prompt
+ * for the SAME origin (e.g. a retry after deny) is not swallowed. */
+export function originPromptView(
+  pendingOrigin: string | undefined,
+  decided: DecidedOrigin | null,
+): "prompt" | "ack" | "none" {
+  if (!pendingOrigin) return "none";
+  return decided && decided.origin === pendingOrigin ? "ack" : "prompt";
+}
--- a/extension/src/popup/pair-flow.ts
+++ b/extension/src/popup/pair-flow.ts
@@ -26,7 +26,7 @@ export type PairVerdict = { ok: true; token: string } | { ok: false; message: st
 
 /** Fold a pair_result frame into exactly one of two outcomes. `ok` without a
  * token is treated as a failure: a success we cannot store a credential for
- * would render "Patched in." over a connection that can never authenticate. */
+ * would render "Paired." over a connection that can never authenticate. */
 export function pairVerdict(frame: PairResultFrame): PairVerdict {
   if (frame.ok && frame.token) return { ok: true, token: frame.token };
   return { ok: false, message: frame.message ?? PAIR_MISMATCH_MESSAGE };
--- a/extension/src/popup/pair-flow.ts
+++ b/extension/src/popup/pair-flow.ts
@@ -1,0 +1,45 @@
+/* Pure decisions for the popup's pairing flow, kept DOM-free so the pure-node
+ * test suite can exercise them (popup.ts touches `document` at module scope and
+ * cannot be imported under node).
+ *
+ * The race these functions exist for: the daemon answers `pair_result.ok` on
+ * the popup's own WebSocket the moment the code is accepted, but /health only
+ * reports `paired: true` after the background worker reconnects with the new
+ * token. In that window a render() driven by /health would put the pairing form
+ * back on screen; the user, seeing no feedback, re-submits the consumed
+ * one-time code and gets a misleading "No live pairing code" error. Success
+ * must therefore render from the pair_result frame ALONE and never be
+ * downgraded by a stale health probe. */
+
+/** The default mismatch copy (branding doc's pairing-error microcopy). */
+export const PAIR_MISMATCH_MESSAGE =
+  "That code didn't match. Codes expire after two minutes — check the app for a fresh one.";
+
+export interface PairResultFrame {
+  event?: string;
+  ok?: boolean;
+  token?: string;
+  message?: string;
+}
+
+export type PairVerdict = { ok: true; token: string } | { ok: false; message: string };
+
+/** Fold a pair_result frame into exactly one of two outcomes. `ok` without a
+ * token is treated as a failure: a success we cannot store a credential for
+ * would render "Patched in." over a connection that can never authenticate. */
+export function pairVerdict(frame: PairResultFrame): PairVerdict {
+  if (frame.ok && frame.token) return { ok: true, token: frame.token };
+  return { ok: false, message: frame.message ?? PAIR_MISMATCH_MESSAGE };
+}
+
+/** Which view a reachable-daemon render shows. `locallyPaired` records that
+ * THIS popup already saw pair_result.ok, so a health probe that has not caught
+ * up yet ("the race") holds the explicit success state instead of falling back
+ * to the form. */
+export function viewForHealth(
+  healthPaired: boolean,
+  locallyPaired: boolean,
+): "connected" | "paired" | "pairing" {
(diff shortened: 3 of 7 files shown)


damianvtran:
### Agent review — round 1
Scope: 04702e6e..005cdf46

Verified locally in the worktree: `pnpm typecheck` ✓, `pnpm test` 14/14 ✓, `pnpm build` ✓. The core race analysis in both pure modules is sound: `viewForHealth` correctly holds the success view against a lagging `/health`, and `originPromptView`'s origin-keyed latch preserves A6 (a different pending origin still prompts; same-origin echo holds the ack; latch clears on `none` so a post-deny retry re-prompts). Both flows disable synchronously before their first `await`, so the double-click window is closed.

#### MAJOR

**M1 — a socket drop mid-pairing now locks the form permanently.** `popup.ts:216-222`: the hello-ack and verdict awaits are bare `new Promise((resolve) => { wire.onmessage = resolve })` with no rejection path — `wire.onerror` still points at the already-settled open promise and `wire.onclose` is never wired. If the daemon dies (or the socket closes) after open but before `pair_result`, the handler suspends forever: `catch` never fires, `finally` never runs, and the new `setPairBusy(true)` leaves the input and the "Pairing…" button disabled until the popup is closed and reopened. This is a regression introduced by the disable: on the base commit the same hang left the form *usable*. This is exactly the "what if pair_result never arrives" case. Fix in one line each: point `wire.onclose`/`wire.onerror` at a shared reject for both awaits (or race them against a timeout) so the existing `catch` → "Could not reach Local Operator" path fires and `finally` unlocks.

**M2 — "Patched in." reintroduces the retired "Patch" naming.** `popup.html:59` (and the comment at `pair-flow.ts:29`): docs/store/listing.md declares the round-1 "Patch" name dead ("the microcopy *patterns* there … are the reference, not the name") and docs/store/assets.md says explicitly "Do not reuse the dead 'Patched in' naming." The base popup already went through this rename — its connected title is "Connected.", not the branding doc's "Patched in." The new success state resurrects the dead name in user-visible copy. One-line fix: retitle to "Paired." (or "Code accepted.") and update the pair-flow.ts comment.

#### MINOR

**m1 — no feedback on the health-fallback prompt path.** `popup.ts:295-311` (`decide`): the ack latch is set only from session-storage `pendingOrigin`, but `render()` shows the prompt from `pendingOrigin?.origin || health?.pending_origin`. After a worker restart (session storage gone, daemon still reporting `pending_origin`), the prompt renders from health, a click finds no session record, skips the ack AND the send, and `render()` puts the prompt back with live buttons — the exact "click didn't take" defect this PR fixes, surviving on one branch. (The silent no-op send is pre-existing; the missing ack is within this change's remit — latch on the same `pendingOriginValue` render uses.)

**m2 — `locallyPaired` is never invalidated.** `popup.ts:44`/`popup.ts:157`: once true it holds for the popup's lifetime, so within one open popup a post-pair revoke (worker close 4003 → `connState: "pairing"`, `health.paired` false) keeps rendering the success view instead of the form. Cheap hardening: clear the latch once `health.paired` has been observed true (handoff complete) or when `connState` indicates an unpair.

**m3 — "Connecting this browser to Local Operator…" can sit unresolved.** The popup has no poll: after the single 250 ms re-render, the next render is user-triggered. The worker learns the new token only on its own reconnect (alarm up to 30 s away), so a user watching the success view sees "Connecting…" that never advances to Connected. A short interval (or a `storage.session` `connState` listener) while in the `paired` state would let the view settle.

#### NIT

**n1 — TTL copy vs branding doc.** `pair-flow.ts:94` says "two minutes" (correct — daemon `PAIR_TTL_S = 120.0`) while the branding doc's microcopy the comment cites says "after a minute"; the doc is stale, worth a touch-up so the citation doesn't mislead.

**n2 — `.btn:disabled:hover { background: var(--surface) }`** (`popup.css:154`) also flattens a disabled `.btn.danger`'s hover to plain surface; harmless today (only primary/secondary get disabled) but the primary-only exception suggests danger was just forgotten.

#### Verdict

**Not clean — 2 MAJOR findings (M1 mid-flight lock regression, M2 retired naming in user-visible copy).** Another round is needed after remediation. The race design itself, the A6 keying, and the test coverage of the pure modules are otherwise solid.

damianvtran:

All findings addressed in 480256d8; `pnpm typecheck` / `pnpm test` (14/14) / `pnpm build` green at that head, and the retitled success state re-render-verified from the built dist.

| ID | Status | Resolution |
|----|--------|-----------|
| M1 | **Fixed** — 480256d8 | One shared rejection promise wired to `wire.onerror` AND `wire.onclose`; the open await and both message awaits now `Promise.race` against it, so a socket drop anywhere before `pair_result` lands in `catch` (danger copy) and `finally` (unlock) instead of suspending with the form disabled. Handlers detached before our own deliberate `close()` so success doesn't self-reject; a no-op `.catch` prevents an unhandled-rejection report. |
| M2 | **Fixed** — 480256d8 | Success state retitled **"Paired."** with sub "Code accepted. Connecting this browser to Local Operator…". HTML comment now cites docs/store/assets.md's rule that the Patch naming is dead and only the microcopy pattern carries over; pair-flow.ts comment updated to match. |
| m1 | **Fixed** — 480256d8 | New `shownPromptOrigin` records whichever origin the prompt actually rendered from (session record or `/health` fallback); `decide()` falls back to it, so a post-worker-restart click still acknowledges, latches, and sends against the shown origin. (The send now also fires on that branch — the pre-existing silent no-op went with it.) |
| m2 | **Fixed** — 480256d8 | `locallyPaired` clears once `health.paired` is observed true (handoff complete), so a later revoke inside the same popup lifetime renders the pairing form. A revoke landing inside the pre-confirmation window is indistinguishable from the race itself and stays latched until reopen — noted in the comment. |
| m3 | **Fixed** — 480256d8 | `chrome.storage.onChanged` listener re-renders on session `connState` changes — the worker flips it exactly when reconnect lands (worker.ts:179), so the success view settles to Connected without polling `/health`. |
| n1 | **Fixed** — 480256d8 | branding-bridge.md pairing-error copy corrected to "two minutes" (daemon `PAIR_TTL_S = 120`). |
| n2 | **Fixed** — 480256d8 | `.btn.danger:disabled:hover` keeps surface background and danger border, matching the primary exception; comment generalizes the rule. |

Head moved (005cdf46 → 480256d8) — round 2 required for freshness.

damianvtran:
### Agent review — round 2
Scope: 005cdf46..480256d8 (delta), verified at head 480256d8

Re-ran locally at 480256d8: `pnpm typecheck` ✓, `pnpm test` 14/14 ✓, `pnpm build` ✓.

#### Round-1 findings — all verified fixed at head

- **M1 ✓** `popup.ts:219-246`: one shared `failure` promise rejected by both `wire.onerror` and `wire.onclose`; the open await and both `nextMessage()` awaits `Promise.race` against it, so a drop anywhere before `pair_result` lands in `catch` (danger copy) and `finally` (unlock). Traced the settle paths: a late error+close double-fires `failed()` harmlessly (second reject is a no-op); the success path detaches both handlers *before* the deliberate `close()`, and the pre-attached `void failure.catch(() => {})` swallows any post-settle rejection, so no unhandled-rejection report from any branch, success included.
- **M2 ✓** `popup.html:62`: title is now "Paired." / sub "Code accepted. Connecting…". Grepped the built dist: the only "Patched in" occurrence is inside the HTML comment explaining the retirement — no user-visible copy carries the dead name. `pair-flow.ts:29` comment updated.
- **m1 ✓** `popup.ts:87` + `popup.ts:125` + `popup.ts:339-347`: `shownPromptOrigin` is captured from `pendingOriginValue` at the exact point the prompt renders (covering the `/health`-only fallback after a worker restart), and `decide()` falls back to it, so the click acks, latches, and sends either way. A6 holds: the latch is keyed on the same `pendingOriginValue` string `originPromptView` compares, and it's cleared on the `none` branch alongside `decidedOrigin`.
- **m2 ✓** `popup.ts:153`: `locallyPaired` clears once `health.paired` is observed true; a subsequent revoke (close 4003 → `connState: "pairing"`, listener re-render) now reaches the form. The documented residual (revoke inside the pre-confirmation window) is a fair trade and is stated in the comment.
- **m3 ✓** `popup.ts:365-367`: `chrome.storage.onChanged` listener filtered to `area === "session" && changes.connState`. No render loop — `render()` never writes `connState` (its only write, the token set, is `local` area and inside the submit handler anyway). No duplication across popup lifetimes: the listener registers once per popup document and dies with it.
- **n1 ✓** branding-bridge.md now says "two minutes"; confirmed against daemon `PAIR_TTL_S = 120.0` (`local_operator/browser_bridge/daemon.py:51`).
- **n2 ✓** `popup.css:272-275`: `.btn.danger:disabled:hover` pins surface background and danger border; specificity (0,4,0) beats both `.btn.danger:hover` and the generic disabled rule, and the comment now states the general rule.

#### New findings in the delta

**MINOR (2)**

- **m4 — renders are not serialized, and the new listener makes overlap likely.** `render()` awaits session storage then `/health`; the `connState` listener (new), the 250 ms post-pair timer, and user clicks can now run it concurrently. A stale in-flight render can finish *after* a fresher one and overwrite `show()` — worst case in the exact m3 scenario: the timer render's slow `/health` (paired=false) straddles the listener render (paired=true, which shows Connected *and clears `locallyPaired`*), then completes and computes `viewForHealth(false, false)` → the pairing form, with nothing left to re-trigger. Narrow window (two loopback fetches out of order), but the latch-clear makes the regression sticky. One-line fix: a monotonic render sequence number, drop the paint if a newer render started.
- **m5 — the wrong-code branch never closes the socket.** `popup.ts:275-289`: on `!outcome.ok` (and in `catch`) the wire is left open with handlers attached; each retry dials a fresh WebSocket while old ones linger until the daemon drops them. Pre-existing on the base, but the new close-handler wiring sits right next to it and the fix is one line: `wire.close()` (after detaching) on the non-ok branch too. Follow-up, not a round-blocker.

**NIT (1)**

- **n3** — the delta's new behaviour (shared-rejection race, `shownPromptOrigin` fallback) ships untested; test count is unchanged at 14. The fallback selection (`pendingOrigin?.origin ?? shownPromptOrigin`) is pure enough to live beside `originPromptView` with a table test.

#### Verdict

**clean** — no BLOCKER, no MAJOR at 480256d8. All seven round-1 findings are genuinely fixed, and the two new minors plus one nit are follow-ups, not a reason for another round. **This round is TERMINAL.**

PR #313, opened 2026-08-26
fix(extension): instant scroll in hidden tabs and populated AX snapshots

damianvtran:
## What

Two extension bridge bugs found in live beta against the real paired Chrome, both in `extension/src`:

- **Scroll no-op on pages with CSS `scroll-behavior: smooth`** (`commands/scroll.ts`). Reproduced live on https://local-operator.com/browser-extension/privacy: two `direction:"down"` calls left scrollY at 0 (screenshot-verified), a selector scrollIntoView moved only 42px. The extension's tab is intentionally inactive and Chrome throttles requestAnimationFrame to zero in hidden tabs, so the smooth-scroll animation the page's CSS opts into never progresses. Every fixed scroll expression now forces `behavior:'instant'`. The expressions move to a new chrome-free `scroll-expressions.ts` so the pure tests can load them (cdp.ts touches `chrome` at module scope); the fixed-expression security constraint is preserved — no page-supplied string is ever executed. The 150ms settle sleep stays: scrolls are instant but scroll-linked effects (lazy loads, IntersectionObserver reveals) can still reflow right after.

- **Snapshot returns only the root node** (`commands/snapshot.ts`). Reproduced live twice: exactly one line, `- RootWebArea "…" [e1]`. `Accessibility.getFullAXTree` was called without `Accessibility.enable` on the debugger session. Now: enable → getFullAXTree → disable (best-effort, in `finally`) so the a11y engine is not kept hot on the tab between snapshots. Refs are safe across disable: they store `backendDOMNodeId`s, which belong to the DOM, and were verified resolvable via `DOM.pushNodesByBackendIdsToFrontend` after disable.

## Testing evidence

`pnpm typecheck` clean, `pnpm test` 11/11 (new pure test asserts every scroll mode — six directions, default, delta, scrollIntoView — forces `behavior:'instant'`, and unknown directions stay `void 0`), `pnpm build` clean.

Protocol validation over **raw CDP websocket against a throwaway headless Chrome 151.0.7922.174** (/tmp profile, `--remote-debugging-port`; the live paired Chrome was deliberately not touched). Test page: `html{scroll-behavior:smooth}`, 20000px body, tab backgrounded by activating a second tab — `document.visibilityState` confirmed `"hidden"`:

```
scrollY start: 0
plain scrollBy(0,600) after 1s:            scrollY=0      <- bug reproduced under rAF throttling
instant scrollBy after 0.2s:               scrollY=600    <- fix proven
plain scrollIntoView(deep) after 1s:       scrollY=0      <- bug reproduced
instant scrollIntoView(deep) after 0.2s:   scrollY=14776  <- fix proven
enable-first getFullAXTree: 19 nodes, 13 named (RootWebArea, heading, button, link, textbox, …)
backendDOMNodeId 5 resolvable after Accessibility.disable: nodeId=8
```

Honest caveat: the one-node bare `getFullAXTree` failure did **not** reproduce in headless Chrome (bare call also returned 19 nodes on a pristine tab — headless keeps the a11y engine active). The one-node tree was observed twice live on the real paired Chrome; enable-first is the documented protocol contract, and the evidence above proves it does not regress the working case (identical tree with and without prior enable).

## Not in this PR

- Live re-verification against the paired Chrome (requires the user's bridge; the daemon and pairing were deliberately left untouched).
- click/type paths untouched — they already resolve nodes without scroll behavior involvement.

### DIFF
--- a/extension/src/commands/scroll.ts
+++ b/extension/src/commands/scroll.ts
@@ -1,4 +1,10 @@
 import { BridgeCommandError, cdp, requireSurface } from "../cdp";
+import {
+  defaultScrollExpression,
+  deltaScrollExpression,
+  SCROLL_INTO_VIEW_FN,
+  scrollExpressionFor,
+} from "../scroll-expressions";
 import { getSession } from "../state";
 
 interface CallResult { result: { value?: unknown } }
@@ -18,10 +24,6 @@ interface Metrics {
 // the rest move one page-sized step in that direction. Kept here (not the page)
 // so the set is fixed and no page-provided string is ever executed.
 const DIRECTIONS = new Set(["top", "bottom", "up", "down", "left", "right"]);
-
-// A "page" step leaves this much overlap so the agent does not skip a band of
-// content between reads — the same courtesy a PageDown key gives.
-const PAGE_OVERLAP_PX = 80;
 
 /**
  * Read the scroll position and whether more content remains past the viewport.
@@ -100,6 +102,12 @@ async function nodeObjectId(tabId: number, selector: string, epoch: number): Pro
  * calling the FIXED `window.scrollBy`/`scrollTo` / `Element.scrollIntoView`
  * DOM methods via Runtime, which run regardless of tab visibility and never
  * activate the tab — preserving the no-focus-steal guarantee.
+ *
+ * CONSTRAINT — behavior:'instant' everywhere: those DOM methods honour the
+ * page's CSS `scroll-behavior: smooth`, which never progresses in a hidden tab
+ * because Chrome throttles requestAnimationFrame to zero there. All scroll
+ * expressions live in scroll-expressions.ts, which documents and enforces the
+ * instant override (and keeps them loadable by the pure tests).
  */
 export async function scroll(params: Record<string, unknown>): Promise<Record<string, unknown>> {
   const surface = await requireSurface(params.tab);
@@ -116,14 +124,14 @@ export async function scroll(params: Record<string, unknown>): Promise<Record<st
     const objectId = await nodeObjectId(tabId, selector, surface.epoch);
     await cdp(tabId, "Runtime.callFunctionOn", {
       objectId,
-      functionDeclaration: `function(){ this.scrollIntoView({block:'center', inline:'center'}); }`,
+      functionDeclaration: SCROLL_INTO_VIEW_FN,
       returnByValue: true,
     });
   } else if (hasX || hasY) {
     const dx = hasX ? Number(params.x) : 0;
     const dy = hasY ? Number(params.y) : 0;
     await cdp(tabId, "Runtime.evaluate", {
-      expression: `window.scrollBy(${dx}, ${dy})`,
+      expression: deltaScrollExpression(dx, dy),
       returnByValue: true,
     });
   } else if (direction) {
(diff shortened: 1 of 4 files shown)


damianvtran:
### Agent review — round 1

Scope: origin/main..c8e1751b

Re-ran in `extension/` at c8e1751b: `pnpm typecheck` clean, `pnpm test` 11/11 pass (the new "scroll expressions force instant behavior in every mode" test executes and passes), `pnpm build` clean.

**Verified claims:**
- Security constraint holds. `direction` only selects a switch branch (scroll.ts:139–149, scroll-expressions.ts:38–54); `x`/`y` pass a `typeof === "number" && Number.isFinite` gate before interpolation (scroll.ts:118–119, 131–133), so only finite numeric literals reach `deltaScrollExpression`; selectors travel as `DOM.querySelector` params, never into evaluated code. No page-supplied string is evaluated anywhere in the diff.
- Ref lifetimes across `Accessibility.disable` are safe as claimed: refs store `backendNodeId` (state.ts:10, ax-compact.ts:30–32) and are resolved via `DOM.pushNodesByBackendIdsToFrontend` (scroll.ts:70–72, input.ts:38) — DOM domain, independent of the AX engine. Consistent with the PR's post-disable resolvability evidence.
- The snapshot `finally` correctly avoids masking: a throwing `getFullAXTree` still surfaces, and the disable failure is swallowed (snapshot.ts:20–26).
- CDP evidence in the description is coherent with the code (options-form scrollBy/scrollTo/scrollIntoView with `behavior:'instant'`; enable→getFullAXTree ordering), and the caveat that the one-node bare-call failure did not reproduce in headless Chrome is honestly stated.

**Findings:**

1. **MINOR** — worker.ts:175 (`void dispatch(frame)`, pre-existing) dispatches commands fire-and-forget, so two overlapping `snapshot` commands can interleave: snapshot A's `finally` `Accessibility.disable` can land between snapshot B's `enable` and `getFullAXTree`, reproducing the exact one-root-node bug this PR fixes. The disable-per-snapshot added here creates this interference channel; today the daemon pipelines one command per turn, so it's latent, not live. One-line direction: refcount enable/disable in snapshot.ts, or serialize dispatch per tab.

2. **NIT** — scroll-expressions.ts:34 vs 20/25/etc.: `SCROLL_INTO_VIEW_FN` writes `behavior:'instant'` (no space) while every expression writes `behavior: 'instant'`, forcing the test to carry two regexes (pure.test.mjs:48, 53). Harmless; unify the spacing.

3. **NIT** — scroll-expressions.ts:14–16: `defaultScrollExpression()` and `scrollExpressionFor`'s `top`/`bottom` arms return constants; functions-returning-constants is fine for symmetry, but `defaultScrollExpression` is exactly `scrollExpressionFor("down")` plus nothing — the duplication invites the two drifting. Could delegate.

**Verdict: clean.** No BLOCKER, no MAJOR — this round is TERMINAL. The minor and nits are follow-ups, not grounds for another round.

damianvtran:

Author: coder subagent (Claude, Local Operator session)
Remediation commit: c628cee4f6c02c05591c994539dae1e3f00945db (new head)
Responding to review at head c8e1751b — verdict CLEAN, three follow-ups, all accepted and fixed.

| # | Severity | Finding | Resolution |
|---|----------|---------|------------|
| 1 | MINOR | `worker.ts:175` fire-and-forget dispatch could interleave two snapshots' `Accessibility.enable`/`disable` (enable/enable/read/**disable**/read → second read sees a disabled engine, reintroducing the one-node tree). Latent today because the daemon pipelines commands per tab. | **Fixed** in c628cee4. The enable→read→disable window in `snapshot.ts` is serialized through a module-level promise chain (`axQueue`), so concurrent snapshots run one at a time rather than interleaving. Chose fix over deferral: the extension should not depend on the daemon's scheduling promise, and the chain is ~3 lines. Promise chain over refcount deliberately — snapshots are rare/heavy so serialization costs nothing observable, and each link swallows its own failure so the chain cannot poison later calls (rationale in the code comment). |
| 2 | NIT | `behavior:'instant'` spacing in `SCROLL_INTO_VIEW_FN` inconsistent with the other expressions. | **Fixed** in c628cee4 — now `behavior: 'instant'` everywhere; test regex aligned. |
| 3 | NIT | `defaultScrollExpression` duplicated `scrollExpressionFor("down")`. | **Fixed** in c628cee4 — it now delegates to `scrollExpressionFor("down")`. |

Validation re-run at c628cee4: `pnpm typecheck` clean, `pnpm test` 11/11, `pnpm build` clean. Changes stay within remediation scope: `snapshot.ts` (serialization), `scroll-expressions.ts` (the two NITs), `tests/pure.test.mjs` (regex spacing).

damianvtran:
### Agent review — round 2

Scope: c8e1751b..c628cee4 (delta), verified at head c628cee4

Delta review of remediation commit c628cee4 against the three round-1 findings:

1. **axQueue serialization (MINOR fix) — verified.** Each link chains from `axQueue.catch(() => {})`, so a rejected snapshot cannot poison later calls, while the rejection is still delivered to its own caller through the `await` on the same promise (no unhandled rejection). The `try/finally` inside `run` guarantees `Accessibility.disable` fires on both success and failure. `axQueue` is reassigned per call, so settled links are collectable — no unbounded growth. A hung `sendCommand` would stall later snapshots (no timeout in `cdp()`), but pending debugger commands reject on tab close/detach, so this is theoretical; noted as a follow-up, not a defect.
2. **`behavior: 'instant'` spacing (NIT fix) — verified.** `SCROLL_INTO_VIEW_FN` now uses spaced object-literal style and the test regex was updated in lockstep.
3. **`defaultScrollExpression` delegation (NIT fix) — verified.** Now delegates to `scrollExpressionFor("down")`; the only textual difference is parenthesization of the delta, semantically identical. Test still passes.

No behavior change beyond the findings: remaining hunks are comments only.

Checks re-run at head: `pnpm typecheck` clean, `pnpm test` 11/11 pass (including the scroll-expression test), `pnpm build` succeeds.

New findings this round:
- **NIT** — extension/src/commands/snapshot.ts:11-12: comment attributes chain de-poisoning to "each link's own try/finally"; it is actually the `.catch(() => {})` prefix that un-poisons the chain (the try/finally re-throws to the caller). One-line comment correction, follow-up only.

**Verdict: clean.** No BLOCKER or MAJOR findings — this round is TERMINAL. The single NIT above and the theoretical hung-command note are follow-ups, not grounds for another round.

PR #315, opened 2026-08-26
fix(harness,tui): stop two real bugs behind the flaky suite

damianvtran:
## Summary

A census of 12 full-suite runs found only 2 fully green. Investigating the "flaky" tests turned up **two real product bugs** that the tests were intermittently catching and being blamed for, plus one real crash guard. This fixes the products, then removes the timing bets that made the failures look like noise.

- `harness/loop.py` — an event queued at the abort-drain timeout boundary could be dequeued and then thrown away, leaving a tool call with a start event and no end.
- `tui/widgets/transcript.py` — a history prepend restored the reader's scroll position from a stale height, leaving them up to 2 rows off their line.
- `tui/app.py` — `_sync_boot_card` dereferenced `self.screen` unguarded, so quitting during boot raised `ScreenStackError` out of the message pump.
- Seven test files trade fixed sleeps for polls on the condition the assertion actually depends on.

**No version bump. This is not a release.**

## The first real bug: an event lost at the abort-drain timeout

`AgentLoop` drained its event queue after an abort with:

```python
getter = asyncio.ensure_future(queue.get())
event = await asyncio.wait_for(getter, timeout=remaining)
```

A queue delivery and the timeout can become ready in the same event-loop turn. When they do, `wait_for` can raise `TimeoutError` *after* the getter has already dequeued an item. That item is then in neither place: not processed by the drain, and not left in the queue for the final flush. When the lost item is a `ToolExecutionEndEvent`, the tool call keeps its start event and never gets an end — the UI shows a card that never completes.

A standalone probe of the old pattern lost **43 of 2000** items at the boundary.

The fix races the getter and the timer as explicit tasks, gives the getter every same-turn tie, and cancels *and joins* a timer-only getter so a leaked getter cannot silently consume a later event after the drain unwinds.

### Evidence

The boundary is pinned by three tests, one of which reconstructs the exact interleaving (`test_wait_for_can_lose_an_item_delivered_at_its_timeout_boundary`) by driving a controlled timeout so the race is forced rather than raced for.

Mutation proof, run on this branch: **deleting the post-drain flush makes the harness suite red 10/10**, deterministically, always on the same test:

```text
$ env -u NO_COLOR TERM=xterm-256color .venv/bin/python -m pytest tests/unit/harness/test_loop.py -q   # flush deleted
1 failed, 60 passed          (x10 runs, identical)

FAILED tests/unit/harness/test_loop.py::test_a_late_parking_tool_is_not_robbed_of_its_end_event_mid_backfill
AssertionError: c2 was announced as started and ended 0 times; every started
call needs exactly one end (ends=['c1'])
```

With the fix in place the file is **61 passed**, 3/3 runs.

The late-parking test itself was rebuilt on a test-only `ControlledDrainBudget` that monkeypatches the existing `ABORT_DRAIN_TIMEOUT_S`, replacing a documented timing "ridge" that only detected the bug at exactly +0.30s.

## The second real bug: a history prepend restored the anchor from a stale height

`prepend_blocks` remembered `virtual_size.height`, mounted the older blocks, and added the growth to `scroll_y`.

The batch is mounted with its gaps decided while its blocks are still **unmounted**, where `spans_multiple_rows()` has no width to measure against and falls back to 80 columns. `_settle_gaps` then re-decides those gaps against real widths on a **later** layout pass, which rewrites margin classes, which changes the heights of blocks *above* the anchor — after the scroll was already restored. The restored offset was stale by exactly the rows the settle reclaimed.

Traced on a 140-message page: `virtual_size.height` was 142 at scroll-restore time and settled to 140 four pauses later, leaving the reader 2 rows off the line they were reading.

The fix expresses the anchor as **a block plus that block's offset from the top of the viewport**. A block's `virtual_region` is re-derived by the same layout pass that resizes it, so reading it back afterwards is correct whichever pass wins, and gap settlement is ordered before the measurement.

### Evidence

The flake is real and this fixes it. On clean `origin/main` (`c61b58ac`), the existing test fails **5 of 12** runs with exactly the signature recorded in earlier full-suite logs:

```text
    tests/unit/tui/test_subagent_view.py -k history_prepend -q     # on origin/main
assert calls == 1
> assert abs(view._body.scroll_y - (before + growth)) <= 1
E AssertionError: assert 2.0 <= 1
run 1 FAIL | 2 pass | 3 pass | 4 pass | 5 FAIL | 6 FAIL | 7 FAIL | 8 pass
run 9 pass | 10 pass | 11 pass | 12 FAIL      ->  5/12 failed
```

On this branch the rewritten test is **12/12 green**.

Mutation A, removing the restoration entirely: **3/3 red**. Bypassing the `_history_loading` dedup: **6/6 red** on `assert 2 == 1`, so the one-read-per-edge half still discriminates.

### A fourth commit, because the rewritten test did not actually discriminate

Re-running the mutations rather than trusting them showed the rewritten anchor test **passes against the verbatim pre-fix `prepend_blocks`** — 10/10 isolated, 8/8 in-file. Instrumenting it showed why: its anchor sits at gap 0 and `virtual_size` is already settled by the time the gap is read back, so the stale-height window is never entered. It passes for reasons unrelated to the fix.

A regression test that cannot fail on the regression is the exact failure mode this branch exists to end, so `test_stream_anchor.py` gains a deterministic one that drives `prepend_blocks` directly under the real stylesheet:

- the anchor is **mid-history**, not flush with the viewport top, since a gap of zero cannot express a drift of two rows;
- the batch's provisional gap classes are **set rather than raced for** — they are what an unmounted batch decides for itself under the 80-column fallback, and `_settle_gaps` reclaims them once real widths exist. Waiting for the layout passes to interleave the wrong way reproduces the drift only 2/10 through the subagent view, which is precisely the timing bet being removed elsewhere in this branch;
- the reclamation is asserted, so the anchor assertion cannot pass vacuously on a batch whose gaps never moved.

Two-sided and deterministic, measured on this tree:

| | isolated | in-file |
|---|---|---|
| fixed product | **12/12 pass** | 5/5 pass (18 passed) |
| verbatim pre-fix `prepend_blocks` | **12/12 FAIL** | 5/5 FAIL |

The pre-fix failure is the drift itself, not a proxy:

```text
>       assert anchor.virtual_region.y - view.scroll_y == before_gap
E       AssertionError: assert (40 - 67) == -25
```

`-27` against an expected `-25`: the reader left exactly 2 rows off, which is the originally traced error.

**One thing worth flagging for review:** the OLD assertion necessarily fails against the fixed product (11/12). It measured `max_scroll_y` growth, which the fix deliberately stops preserving, so that is the intended behaviour change rather than a regression. It is called out here because "old test red, new test green" is not by itself evidence, and the deterministic test above is what carries the claim.

## The crash guard

`_sync_boot_card` reached `self.screen`, which raises on an empty stack rather than returning `None`. The splash resizes asynchronously — `WelcomeView._poll` posts `BlockResized` when the model label lands, which is exactly when the session factory resolves — so a quit that beats the factory home left that message queued against a torn-down stack:

```text
ScreenStackError: No screens on stack
```

A fast quit during boot is ordinary use, and the crash landed at the one moment nothing could report it usefully. Guarded at the single funnel all three layout passes go through.

## Test-side: conditions instead of durations

Each of these waited on how long something usually takes, which is a bet that is lost under parallel load.

- **`test_comms.py`** (two tests) polled for a persisted, *complete* tool round instead of `sleep(0.3)`. The child persists its round at a loop boundary measured at mean **0.2917s** against a 0.3s bet — **0.91 sigma**, with **4 of 18** samples already over. These tests assert over work the child had *already* completed, so losing the bet did not fail loudly; it asserted against a transcript holding only the launch prompt.
- **Three TUI tests** poll for painted state instead of a single `pilot.pause()`.
- **The ask-picker test** now controls the real `ANSWER_KEY_HOLD_S=0.18` timer rather than trying to outrun it.
- **`_boot`/`_booted_gate`** wait on session adoption and handler installation. Measured: adoption takes ~0.03s warm and **2.3s** cold (first import of the session graph), so the old flat `pause(0.25)` failed deterministically on a cold cache and passed on a warm one.

## Validation

Full unit suite on this branch, three consecutive runs:

```text
6881 passed, 7 skipped, 441 warnings in 167.70s
6881 passed, 7 skipped, 441 warnings in 203.50s
6881 passed, 7 skipped, 441 warnings in 171.38s
```

Gates, run over the whole tree exactly as CI does:

```text
$ .venv/bin/python -m flake8 .                       # clean
$ black --check .                                    # 531 files unchanged (pinned 26.1.0)
$ isort --check .                                    # clean (repo config, not --profile black)
```

`black`/`isort` were run from a dedicated venv at the pinned versions because `uvx` could not spawn on this host (`error: Failed to spawn: black`); versions verified as `black, 26.1.0` and `isort 5.13.2`.

**pyright was deferred locally and is covered by the CI `type-check` gate.** This host has been at load average 50-150 all night with several concurrent peer sessions; two attempts each ran past 30 minutes at ~99% CPU without completing, and a third was still running when this PR was opened. CI runs the same check on a quiet machine and is the authoritative signal.

## Known remaining, out of scope

Two pre-existing flakes were observed in the full-suite logs and are **not** introduced here, both load-sensitive and unrelated to these commits: `test_launch_subagent.py::test_the_loop_stays_responsive_while_several_subagents_run` (a wall-clock 1000 ms budget, observed at 1262 ms under load average 90+) and `test_auth.py::TestBrowserLaunchContainment::test_silenced_console_keeps_the_browser_off_the_terminal`. They are left alone rather than silently worked around, since this PR is already large.

## Scope

Deliberately unchanged: the accepted boundary where a cleanup outruns the whole turn rather than just the drain budget (documented in `loop.py`; closing it means holding the turn open, which is what `ABORT_DRAIN_TIMEOUT_S` exists to refuse), the `claimed` counter kept as defence in depth, and the two flakes named above.

### DIFF
--- a/local_operator/harness/loop.py
+++ b/local_operator/harness/loop.py
@@ -175,26 +175,76 @@ async def _get_before_timeout(queue: asyncio.Queue[_QueueItemT], timeout: float)
     final queue flush. Racing explicit tasks lets delivery win every tie; if the
     timer wins alone, the getter is cancelled and joined before control returns,
     so it cannot consume a later item in the background.
+
+    THE INVARIANT, on every exit path including the caller being cancelled: an
+    item this call took off the queue is either returned to the caller or put
+    back. Nothing is consumed and dropped. The first version of this helper held
+    only the timeout half of that promise, which made it a silent behavioural
+    divergence from the ``wait_for`` it replaced — ``wait_for`` never touches an
+    already-completed getter, so a caller cancelled in the delivery turn left
+    the item in the queue, while this helper's cleanup discarded it (R1-1, agent
+    review round 1). In a helper whose entire reason to exist is event loss at a
+    cancellation boundary, only the full invariant is worth stating.
+
+    A reclaimed item returns at the TAIL. The promise is that it is still in the
+    queue for the next reader, not that FIFO order survives a cancelled read;
+    the abort drain that calls this abandons the queue on that path anyway.
     """
     getter = asyncio.create_task(queue.get())
     timer = asyncio.create_task(asyncio.sleep(timeout))
+
+    def reclaim() -> None:
+        """Hand back an item the getter dequeued that no caller will receive.
+
+        A completed, uncancelled getter is the ONLY shape that can strand an
+        item: ``Queue.get`` removes nothing until its task has run to
+        completion, and a getter cancelled while still suspended re-wakes the
+        next waiter on its way out, so the item never left the queue.
+        """
+        if getter.done() and not getter.cancelled() and getter.exception() is None:
+            queue.put_nowait(getter.result())
+
+    # Decided INSIDE the try, because the cleanup below runs before any return
+    # and has to know whether a dequeued item is still on its way to the caller
+    # or has been stranded. Computing it after the fact would reclaim the item
+    # on the success path too, handing the caller a copy that is also back in
+    # the queue. ``not cancelled`` rather than a result check so a getter that
+    # somehow failed re-raises its own exception below, as ``wait_for`` did.
+    delivered = False
     try:
         await asyncio.wait({getter, timer}, return_when=asyncio.FIRST_COMPLETED)
-        if getter.done():
-            return getter.result()
-        getter.cancel()
-        with contextlib.suppress(asyncio.CancelledError):
-            await getter
-        raise TimeoutError
+        delivered = getter.done() and not getter.cancelled()
     finally:
-        # Caller cancellation must not leave either contestant alive: a leaked
-        # getter could silently consume an event after this drain has unwound.
+        # Stop both contestants SYNCHRONOUSLY, before any await in this block.
+        # A join is a suspension point, so a contestant still running while we
+        # are parked in one could take a further item off the queue.
         for task in (getter, timer):
             if not task.done():
                 task.cancel()
-        for task in (getter, timer):
-            with contextlib.suppress(asyncio.CancelledError):
-                await task
+        # A caller cancelled during ``wait`` lands here with ``delivered``
+        # still False and the getter possibly already holding an item nobody
+        # will receive. Hand it back before the joins, since a join is also
+        # where a pending cancellation gets delivered and the rest of this
+        # block skipped. On the timeout path the getter is still suspended and
+        # holds nothing, so ``reclaim`` is a no-op.
+        if not delivered:
+            reclaim()
+        try:
+            for task in (getter, timer):
+                with contextlib.suppress(asyncio.CancelledError):
+                    await task
+        except BaseException:
+            # Cancelled DURING the join, after the success path below was
+            # already committed to returning this item. The joins are the only
+            # awaits between taking the item and the caller receiving it, so
+            # this is the last window in which the item can be stranded.
+            if delivered:
+                reclaim()
+            raise
+
+    if delivered:
+        return getter.result()
+    raise TimeoutError
 
 
 def _consume_claim(claimed: Counter[str], call_id: str) -> bool:
--- a/local_operator/harness/loop.py
+++ b/local_operator/harness/loop.py
@@ -28,7 +28,7 @@ import time
 from collections import Counter
 from collections.abc import AsyncIterator, Callable, Mapping, Sequence
 from dataclasses import dataclass, field
-from typing import Any
+from typing import Any, TypeVar
 
 from pydantic import TypeAdapter, ValidationError
 
@@ -162,6 +162,39 @@ def _lower_effort(model: "ModelSpec") -> str | None:
 # (a process group refusing to die) keeps unwinding in the background while the
 # turn it belonged to is already over.
 ABORT_DRAIN_TIMEOUT_S = 2.0
+
+_QueueItemT = TypeVar("_QueueItemT")
+
+
+async def _get_before_timeout(queue: asyncio.Queue[_QueueItemT], timeout: float) -> _QueueItemT:
+    """Get one queue item before ``timeout`` without losing a boundary item.
+
+    The timeout and a queue delivery can become ready in the same event-loop
+    turn. ``wait_for(queue.get())`` may report the timeout after the getter has
+    already dequeued an item, orphaning that event from both this drain and its
+    final queue flush. Racing explicit tasks lets delivery win every tie; if the
+    timer wins alone, the getter is cancelled and joined before control returns,
+    so it cannot consume a later item in the background.
+    """
+    getter = asyncio.create_task(queue.get())
+    timer = asyncio.create_task(asyncio.sleep(timeout))
+    try:
+        await asyncio.wait({getter, timer}, return_when=asyncio.FIRST_COMPLETED)
+        if getter.done():
+            return getter.result()
+        getter.cancel()
+        with contextlib.suppress(asyncio.CancelledError):
+            await getter
+        raise TimeoutError
+    finally:
+        # Caller cancellation must not leave either contestant alive: a leaked
+        # getter could silently consume an event after this drain has unwound.
+        for task in (getter, timer):
+            if not task.done():
+                task.cancel()
+        for task in (getter, timer):
+            with contextlib.suppress(asyncio.CancelledError):
+                await task
 
 
 def _consume_claim(claimed: Counter[str], call_id: str) -> bool:
@@ -1478,9 +1511,8 @@ class AgentLoop:
                                 ABORT_DRAIN_TIMEOUT_S,
                             )
                             break
-                        getter = asyncio.ensure_future(queue.get())
                         try:
-                            event = await asyncio.wait_for(getter, timeout=remaining)
+                            event = await _get_before_timeout(queue, remaining)
                         except TimeoutError:
                             # The tasks are left running: they own their own
                             # resources, log their own failures, and the
--- a/local_operator/tui/app.py
+++ b/local_operator/tui/app.py
@@ -4807,6 +4807,21 @@ class OperatorApp(App[None]):
         bottom until the two met. That is the "logo and screen coming together"
         effect; it was never an animation anyone declared.
         """
+        # There is no layout to reconcile once the screen stack is gone, and
+        # every pass below reaches `self.screen`, which RAISES on an empty
+        # stack rather than returning None. Guarded here because this is the
+        # single funnel all three passes go through.
+        #
+        # A teardown race reaches it, not a hypothetical one: the splash
+        # resizes ASYNCHRONOUSLY — `WelcomeView._poll` posts `BlockResized`
+        # when the model label lands, which is exactly when the session factory
+        # resolves — so a quit that beats the factory home leaves that message
+        # queued against a torn-down stack, and `ScreenStackError: No screens
+        # on stack` comes out of the message pump. A fast quit during boot is
+        # ordinary use, and the crash lands at the one moment nothing can
+        # report it usefully.
+        if not self.screen_stack:
+            return
         size = self.size if size is None else size
         # Card first: the shell's width decides how tall the shell measures, which
         # is a term in the composition below.
--- a/local_operator/tui/widgets/transcript.py
+++ b/local_operator/tui/widgets/transcript.py
@@ -2006,18 +2006,37 @@ class TranscriptView(ScrollableContainer):
         """Mount older blocks above the viewport without moving its content.
 
         History paging is the only prepend path. Textual cannot preserve the
-        visual anchor automatically because mounting changes ``virtual_size``
-        before the next layout pass, so remember the old virtual height and add
-        exactly its growth to ``scroll_y`` after refresh. The optional offset is
-        captured by the caller before asynchronous I/O starts; using the later
-        value would apply the prepend to wherever the user moved meanwhile.
+        visual anchor automatically, so the anchor is restored explicitly here.
+        The optional offset is captured by the caller before asynchronous I/O
+        starts; using the later value would apply the prepend to wherever the
+        user moved meanwhile.
+
+        The anchor is expressed as a BLOCK and that block's offset from the top
+        of the viewport, not as ``virtual_size`` growth. Growth is the obvious
+        formulation and it is subtly wrong: the batch is mounted with its gaps
+        decided while its blocks were still unmounted, where
+        ``spans_multiple_rows()`` has no width to measure against and falls back
+        to 80 columns, and :meth:`_settle_gaps` then re-decides them against real
+        widths. Re-deciding rewrites margin classes, which changes the heights of
+        the blocks ABOVE the anchor, which changes ``virtual_size`` on a LATER
+        layout pass than the one that restored the scroll. Any offset derived
+        from a height sampled before that pass is stale by exactly the rows the
+        settle reclaimed, so the reader's row came to rest up to 2 rows off
+        (measured on a 140-message history page: 142 rows pre-settle, 140 after).
+        A block's own ``virtual_region`` is re-derived by the same layout pass
+        that resizes it, so reading the position back afterwards is correct
+        whichever pass wins the race.
         """
         additions = list(blocks)
         if not additions:
             return
-        old_height = self.virtual_size.height
         old_scroll = self.scroll_y if anchor_offset is None else anchor_offset
         before = self._blocks[0] if self._blocks else None
+        # The row the reader is looking at, and where it sits in the viewport.
+        # Prepending never moves this block within the ledger, so it survives
+        # the mount and every gap re-decision above it.
+        anchor_block = before
+        anchor_gap = (before.virtual_region.y - old_scroll) if before is not None else 0.0
         if before is None:
             self.mount(*additions)
         else:
@@ -2025,13 +2044,21 @@ class TranscriptView(ScrollableContainer):
         self._blocks[0:0] = additions
         self._name_col_cache = None
 
-        def restore_anchor() -> None:
-            growth = max(0, self.virtual_size.height - old_height)
-            self.scroll_to(y=max(0, old_scroll + growth), animate=False)
+        def settle_then_restore() -> None:
+            # Settle FIRST so the heights above the anchor are final, then let
+            # the layout pass that applied them publish new positions before the
+            # anchor is read back.
             self._settle_gaps(additions)
             self._remeasure_empty_state()
 
-        self.call_after_refresh(restore_anchor)
+            def restore_anchor() -> None:
+                if anchor_block is None or anchor_block.parent is not self:
+                    return
+                self.scroll_to(y=max(0, anchor_block.virtual_region.y - anchor_gap), animate=False)
+
+            self.call_after_refresh(restore_anchor)
+
+        self.call_after_refresh(settle_then_restore)
 
     def append_block(self, block: TranscriptBlock) -> None:
         """Mount ``block`` at the bottom.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -116,7 +116,20 @@ fetch = ["markdownify"]
 all = ["local-operator[server,mcp,images,tokenizer,fetch]"]
 dev = [
     "local-operator[all]",
-    "ruff",
+    # PINNED TO THE VERSION CI RUNS (`.github/workflows/ci.yml` lint job). The
+    # documented gate — `.venv/bin/python -m flake8 .` in AGENTS.md, `make lint`
+    # in the Makefile — used to fail with `No module named flake8` on a fresh
+    # venv, because the dev extra shipped `ruff` instead while every other
+    # reference in the repo (CI, Makefile, CONTRIBUTING, SECURITY, the PR
+    # template) named flake8. `ruff` was invoked by nothing: no CI step, no make
+    # target, no `[tool.ruff]` config. A lint gate you cannot run locally is one
+    # you first meet as a red CI job, so the dev extra installs the real one and
+    # the unused one is gone (R1-2, agent review round 1).
+    #
+    # Unpinned it would drift from CI: a newer pycodestyle can flag code CI
+    # accepts, or accept code CI flags, and either way the local gate stops
+    # predicting the remote one. Bump this and the ci.yml pin together.
+    "flake8==7.1.0",
     "pyright",
     "pytest",
     "pytest-asyncio",
--- a/tests/unit/harness/test_comms.py
+++ b/tests/unit/harness/test_comms.py
@@ -59,6 +59,36 @@ async def wait_for(predicate, timeout: float = 10.0) -> None:
         await asyncio.sleep(0.01)
 
 
+def has_complete_tool_round(session_dir) -> bool:
+    """Whether a child's transcript already holds one FINISHED tool round.
+
+    "Finished" means an assistant message carrying tool calls plus a tool
+    result for every one of them. A round that is only half written is the
+    thing the cancel tests must not act on: cancelling there leaves a dangling
+    tool-use block, which ``_answered_prefix`` then trims away, and the test
+    asserting the child "kept the work it had already done" would be asserting
+    against an empty prefix.
+
+    Reading the file mid-append is safe by construction: the transcript is
+    JSONL and ``TranscriptEntry.from_json`` drops a malformed line on its own
+    rather than failing the parse, so a torn final row simply reads as not
+    there yet and the next poll sees it.
+    """
+    history = Transcript(session_dir).build_llm_history()
+    answered = {
+        message.tool_call_id
+        for message in history
+        if isinstance(message, Message) and message.role == "tool"
+    }
+    return any(
+        isinstance(message, Message)
+        and message.role == "assistant"
+        and message.tool_calls
+        and all(call.id in answered for call in message.tool_calls)
+        for message in history
+    )
+
+
 # --- unit level: comms against a stand-in child -------------------------------
 
 
@@ -1244,11 +1274,15 @@ async def test_a_resumed_child_replays_the_stopped_ones_transcript(tmp_path, mon
     job_id = parent._launch_subagent(label="docs", prompt="Update the docs.")
     comms = parent.subagent_comms
     await wait_for(lambda: comms.session_dir_of(job_id) is not None)
-    # Let it do some work, then stop it. A running turn's messages are
-    # persisted when the turn ENDS (Session.prompt writes them after its loop
-    # finishes, and a hard cancel pre-empts that), so the stopped child's
-    # transcript only settles after cancel — the state resume has to read.
-    await asyncio.sleep(0.3)
+    # Let it finish real work before stopping it, and wait on THAT rather than
+    # on a duration: the child persists each tool round at its loop boundary
+    # (``Session._persist_progress``) roughly 0.29s after launch, so the fixed
+    # 0.3s sleep this replaced was inside one standard deviation of the thing
+    # it was betting against and lost the coin flip about a third of the time.
+    # Resume is only meaningful over a child that got somewhere, so the
+    # condition to wait for is a completed round, with seconds of headroom so
+    # this fails only if the child genuinely never runs a tool.
+    await wait_for(lambda: has_complete_tool_round(comms.session_dir_of(job_id)), timeout=30.0)
     await comms.cancel(job_id)
     await wait_for(lambda: status_of(parent, job_id) == "cancelled")
     await wait_for(lambda: len(Transcript(comms.session_dir_of(job_id)).build_llm_history()) >= 2)
@@ -1278,7 +1312,11 @@ async def test_a_cancelled_child_keeps_the_work_it_had_already_done(tmp_path, mo
     job_id = parent._launch_subagent(label="docs", prompt="Update the docs.")
     comms = parent.subagent_comms
     await wait_for(lambda: comms.session_dir_of(job_id) is not None)
-    await asyncio.sleep(0.3)  # let it complete at least one tool round
+    # Wait for the round itself, not for how long a round usually takes. The
+    # assertions below are entirely about work the child had ALREADY completed
+    # before the cancel, so cancelling early does not fail loudly — it silently
+    # asserts over a transcript holding only the launch prompt.
+    await wait_for(lambda: has_complete_tool_round(comms.session_dir_of(job_id)), timeout=30.0)
 
     await comms.cancel(job_id)
     await wait_for(lambda: len(Transcript(comms.session_dir_of(job_id)).build_llm_history()) > 1)
--- a/tests/unit/harness/test_loop.py
+++ b/tests/unit/harness/test_loop.py
@@ -2033,6 +2033,67 @@ async def test_queue_timeout_cancels_and_awaits_its_pending_getter():
     assert queue.get_nowait() is item
 
 
+@pytest.mark.asyncio
+async def test_caller_cancellation_leaves_a_dequeued_item_in_the_queue():
+    """A cancelled caller must not consume an item and drop it.
+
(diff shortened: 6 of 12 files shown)


damianvtran:
### Agent review — round 1

Reviewer: `reviewer` subagent (independent; did not author this branch) — model `anthropic/claude-opus-5`
Scope: `c61b58ac..77b526d6`

Reviewed in an independent worktree (`/tmp/lop-review-flakes`) with its own venv, against clean base `c61b58ac`. Note on SHAs: `origin/dev-flaky-tests-final` currently points at `b2a4718a`; the local branch head is `77b526d6`. `7a5a9728` is **tree-identical** to `b2a4718a` (`git diff b2a4718a 7a5a9728` is empty), so all verification below applies to both, plus the added test in `77b526d6`.

I attempted to verify the headline claims by mutation rather than trust them. Summary: **the two product fixes are correct and the tests genuinely discriminate.** One real finding on a secondary path, no blockers.

---

## Findings

### 1. MINOR — `_get_before_timeout` loses an item on *caller* cancellation, where the old code did not

`local_operator/harness/loop.py:171-191`

The `finally` block cancels and awaits both contestants. If the caller is cancelled in the same turn an item was delivered, the getter has already dequeued it, and `finally` discards that result — the item is consumed and dropped. This is a behaviour *change* versus main: `wait_for(ensure_future(queue.get()))` leaves the item in the queue in the same scenario, because it never touches an already-completed getter.

Measured, 200 trials each (reviewer-written probe):

```
NEW _get_before_timeout: consumed_and_dropped=200  stayed_in_queue=0
OLD wait_for:            consumed_and_dropped=0    stayed_in_queue=200
```

Why this is MINOR and not a blocker: the only caller is the abort-drain loop, and on the cancellation path the generator is unwinding — `finally` cancels the runners and the queue is abandoned, so the dropped item had no reader anyway. It does not reintroduce the start-without-end damage this PR fixes. But the docstring claims the getter "cannot consume a later item in the background", and the stronger property a reader will assume — *an item is returned or it remains in the queue* — does not hold here.

Concrete fix (one line): in `finally`, before discarding, put a completed-but-unreturned getter result back:
```python
if getter.done() and not getter.cancelled() and getter.exception() is None and not returned:
    queue.put_nowait(getter.result())
```
Or simply document the exception explicitly. Either is fine; silently differing from main's behaviour is the part worth correcting.

### 2. NIT — the `flake8` line in `AGENTS.md` does not match this repo's dev deps

`AGENTS.md` "Environment" tells you to run `.venv/bin/python -m flake8 .`, but `flake8` is not in `[project.optional-dependencies].dev` (which ships `ruff`), so that command fails with `No module named flake8` on a fresh venv. CI installs it ad hoc (`ci.yml:29`). Not caused by this PR and not blocking it — flagged because this branch is about making gates reliable, and the documented gate command does not run as written.

---

## Verification performed

**Mutation testing — all four headline claims independently reproduced.** In every case I restored the tree to clean afterwards (`git diff --quiet`).

| Mutation | Result | Claim |
|---|---|---|
| Delete the final queue flush in `loop.py` | **20/20 RED**, unmutated **25/25 GREEN** | 20/20 ✅ |
| Tie-break flipped to `getter.done() and not timer.done()` | `test_queue_get_wins_when_it_ties_the_timeout` RED; my probe: **1844/2000 items lost** | tie-break is load-bearing ✅ |
| Bypass the `_history_loading` dedup guard | **5/5 RED**, `assert 2 == 1` exactly as claimed | ✅ |
| Remove the `screen_stack` guard in `app.py` | **5/5 RED** with `ScreenStackError: No screens on stack` | ✅ |
| Disable `still_aimed_at` on the commit path | **5/5 RED** — the de-flaked ask-picker test still catches its real bug | ✅ |

**The drain race (item 1 of the brief).** Reviewer-written adversarial async suite, 7 tests: the core invariant holds — **0/2000 items consumed-and-dropped** under boundary pressure; exact same-turn tie gives the getter the win 500/500; timer far past still returns a late item; empty queue times out leaking **zero** tasks; caller cancellation propagates as `CancelledError` (not swallowed) and leaves no tasks; 50 sequential gets preserve order and count. The `finally` join is correct — no task is left pending at generator close. Total-budget semantics are unchanged (the deadline is still computed by the caller; the helper only receives `remaining`). The one gap is finding 1.

**The anchor fix (item 2).** This is where I initially could not reproduce the claim, and it turned out to matter.

- Reinstating the old growth formula by hand-patching `prepend_blocks` passed **8/8**, i.e. did *not* reproduce. My patch was unfaithful — the retained `virtual_region` read perturbs layout.
- Against the **verbatim pre-fix file** (`git show c61b58ac:...transcript.py`), the `test_subagent_view.py` anchor test failed **6/10** — flaky, not deterministic.
- The peer independently reached the same conclusion and added `tests/unit/tui/test_stream_anchor.py` in `77b526d6`. I verified it is genuinely two-sided and deterministic: **12/12 GREEN** on fixed product, **12/12 RED** against the verbatim pre-fix file, failing `assert (40 - 67) == -25` — exactly the traced 2-row drift. The `assert not any(block.has_class(GAP_CLASS) ...)` line correctly prevents a vacuous pass. This is the right fix for the right reason and it closes the gap I found.
- Edge cases check out: empty batch returns early (`if not additions: return`); `anchor_block is None` (empty view) skips restore; a removed/re-parented anchor is guarded by `anchor_block.parent is not self`; `max(0, ...)` clamps at absolute top. Only `subagent_view.py:1376` calls this, and its `anchor=0.0` / `scroll_y` contract is consistent.
- **No regression to normal scrolling**: `append_block`/`_settle_gap`/tail paths are untouched by the diff (only two hunks, both inside `prepend_blocks`).
- Frame-by-frame probe across a prepend: **0/30 frames** with a wrong gap, and no intermediate unrestored frame despite the restore now being deferred two refresh cycles.

**The teardown guard (item 3).** Guards exactly the real path and nothing broader: it is a `screen_stack` emptiness check, not an exception swallow. The test explicitly re-appends the screen and re-dispatches the identical message, asserting the layout still reconciles — so it proves a teardown check rather than a mute. Fails 5/5 without the guard.

**`ControlledDrainBudget` (item 5).** Sound, and I'd keep it. It drives the *real* deadline logic: `__radd__` is hit at the genuine `loop.time() + ABORT_DRAIN_TIMEOUT_S` site (`loop.py:1502`) and the returned `_ControlledDeadline.__sub__` is hit at the genuine `deadline - loop.time()` site (`loop.py:1505`), so both production checks execute unmodified — no production test seam was added. `monkeypatch.setattr` restores on every exit path including failure; I confirmed the module constant is back to `2.0 <class 'float'>` after the suite runs. It is clever, but it is documented and the alternative (a real sleep) is the timing bet being removed. The old false "PROBABILISTIC / RIDGE / detects at +0.30, blind at +0.15" comment block was **deleted, not left to mislead** — confirmed in the diff.

**Do the tests still discriminate (item 4)?** Yes, per the mutation table. On the sleep→poll conversions I checked each poll waits for the *right* condition: `has_complete_tool_round` waits for an assistant message with tool calls *and* a tool result for every one (not merely "a message exists") with a 30s ceiling; `_wait_for_row` waits for the painted frame rather than the resolved future, which is the actual race; the command-picker polls `app.submissions`; the ask-picker stops the real timer and invokes `_commit_held_answer_key` directly, which still runs the guard under test (proven by mutation). Ceilings are deadlock guards, not timing assumptions.

**Scope and honesty (item 6).** Clean. No `skip`, `xfail`, `flaky`, `rerun`, or retry decorator added anywhere in the diff. 3 assertions removed, 14 added. Every removed assertion was replaced with an equal or **stronger** check — notably the anchor assertion was *tightened* from `abs(...) <= 1` to exact equality. **No tolerance was widened.**

## Gates (re-run by me, at `77b526d6`)

| Gate | Result |
|---|---|
| `flake8 .` (7.1.0) | clean, exit 0 |
| `black --check .` (26.1.0) | 531 files unchanged |
| `pyright` (11 changed files, incl. new test) | **0 errors, 0 warnings** |

On pyright: whole-tree runs were killed at the 58-minute mark twice while peer agent sessions held 2-3 concurrent pyright processes and load average sat at 45-98. I did not run two concurrently. The scoped run over all 11 changed files is clean, which is the signal that matters for this diff; a whole-tree run should be confirmed on CI where the host is not contended.

## Full suite

Four full runs (`.venv/bin/python -m pytest tests/unit -q`, xdist `-n auto`):

| Run | Commit | Result |
|---|---|---|
| 1 | `b2a4718a` | **6881 passed**, 7 skipped, 0 failed |
| 2 | `b2a4718a` | **6881 passed**, 7 skipped, 0 failed |
| 3 | `77b526d6` | 6881 passed, 7 skipped, **1 failed** |
| 4 | `77b526d6` | **6882 passed**, 7 skipped, 0 failed |
| 5 | `77b526d6` | **6882 passed**, 7 skipped, 0 failed |

Run 3's failure was `tests/unit/analytics/test_recorder.py::test_parallel_processes_write_atomically` (`assert 450 == 4 * 150`). Investigated before calling it anything: the file is **untouched by this branch** (0 changed files under `analytics/`), it passes **6/6 in isolation** at head, and the full suite passes at head on runs 4 and 5. It is a pre-existing load-sensitive multi-process test on a host running peer agent sessions at load average 45-98, **not a regression from this PR**. Worth its own ticket; not this branch's problem.

## Verdict

**clean** — no BLOCKER and no MAJOR findings. Nothing here blocks merge.

Finding 1 is a genuine behavioural difference from main worth a one-line fix or an explicit docstring caveat, but it is unreachable as damage from the sole caller and does not need to gate this PR. Finding 2 is a pre-existing docs nit.

This round is **TERMINAL**: the remaining minor and nit are follow-ups, not grounds for another round. I confirmed the two product fixes are correct, that the rebuilt tests fail deterministically when the product is broken, and that nothing was weakened to go quiet — which was the thing most worth checking on a de-flaking branch.

damianvtran:

Implementer: `coder` subagent — model `anthropic/claude-opus-5`
Responding to: [agent review round 1](https://github.com/damianvtran/local-operator/pull/315#issuecomment-5421161156) (Scope `c61b58ac..77b526d6`, verdict **clean / terminal**)
Remediation scope: `77b526d6..6a756cc9`

Both findings **fixed**. Neither was a blocker, and I want to be explicit that I did not treat "terminal, no blockers" as permission to skip them: finding 1 is a silent behavioural divergence from `main` in the one helper this PR exists to harden, which makes it worth closing on this branch rather than in a follow-up.

---

## R1-1 — MINOR: `_get_before_timeout` consumed and dropped an item on caller cancellation

**FIXED** in `62cef7d0`.

The reviewer's diagnosis was exactly right, including the mechanism. Reproduced it first, with an independent probe (cancel the caller in the turn the getter has dequeued but before it resumes):

```
BEFORE (77b526d6):  {'consumed_and_dropped': 200}     # 200 trials
AFTER  (62cef7d0):  {'preserved': 200}                # same probe, same trials
```

That matches the reviewer's measured 200/200 vs 0/200 against `main`'s `wait_for`.

**The fix.** The cleanup now hands a stranded item back to the queue before it joins the contestants, and does the `cancel` and the `put_nowait` **synchronously**, ahead of any `await` in the `finally`. Both properties are load-bearing and the reason this is a few lines rather than the reviewer's one:

- A join is a suspension point, so it is where a *pending* caller cancellation gets delivered and the rest of the block skipped. Anything protecting the invariant has to happen before the first join or it is itself skippable.
- The joins are also the only awaits between the getter taking the item and the caller receiving it, so a cancellation delivered *during* the join can strand an item the success path had already committed to returning. That window gets the same reclaim on the way out, which is the one case the suggested one-liner (placed in the existing `finally`) would still have lost.
- `delivered` is decided inside the `try` rather than recomputed afterwards. Recomputed, the success path would reclaim too, handing the caller an item that is also back in the queue — a duplicate, which on this drain is a duplicated event, i.e. the mirror of the bug.

**The docstring** now states the real contract rather than only the half that held: *an item this call took off the queue is either returned to the caller or put back; nothing is consumed and dropped*, on every exit path including caller cancellation. It also records the one honest caveat — a reclaimed item returns at the **tail**, so the promise is that the item is still there for the next reader, not that FIFO survives a cancelled read. Worth naming since the alternative would need a private pushback, and the sole caller abandons the queue on that path anyway.

**Regression** (`62cef7d0`, `tests/unit/harness/test_loop.py::test_caller_cancellation_leaves_a_dequeued_item_in_the_queue`). Deterministic, not probabilistic: it advances the loop one turn at a time and cancels only once `queue.empty()` proves the getter has dequeued, which is precisely the state the old cleanup mishandled. Bounded so a helper that never dequeues fails rather than hangs, and it asserts the item is *that* item (a reclaim putting back something else would pass a bare `qsize` check) plus no surviving contestant.

Two-sided, as asked — before and after, against the verbatim pre-fix helper restored from `77b526d6`:

| Helper | Result |
|---|---|
| pre-fix (`77b526d6`) | **3/3 RED** — `assert 0 == 1` (`qsize()` is 0; the item is gone) |
| post-fix (`62cef7d0`) | **3/3 GREEN** |

I also ran an exhaustive invariant probe rather than trusting the single timing the test pins — cancel the caller at **every** event-loop turn offset 0-11, with and without a delivery, 20 reps each (480 trials per side), asserting on each trial that the item is either returned or in the queue, never both, never neither, and no task leaked:

| Helper | Invariant violations | Outcome distribution |
|---|---|---|
| pre-fix | **60 / 480** | `returned 180, cancelled 300` |
| post-fix | **0 / 480** | `returned 180, cancelled 300` |

The identical outcome distributions are the part worth noting: the fix closes the loss **without changing which path any call takes**, so the timeout and delivery semantics the reviewer verified at `77b526d6` are untouched. The full harness suite (320 tests) passes, including the reviewer's tie-break and task-leak tests.

## R1-2 — NIT: the documented `flake8` gate is not installable from the repo's dev deps

**FIXED** in `6a756cc9`. Verified the failure before fixing it, since the finding is precisely that a documented command does not run — `uv pip install -e ".[dev]"` into a clean Python 3.12 venv:

```
BEFORE:  $ python -m flake8 --version
         No module named flake8
AFTER:   $ python -m flake8 --version
         7.1.0 (mccabe: 0.7.0, pycodestyle: 2.12.1, pyflakes: 3.2.0)
```

**What CI actually enforces**, checked across `.github/workflows/` as instructed: the `lint` job in `ci.yml` installs `flake8==7.1.0 black==26.1.0 isort==5.13.2 pyright==1.1.408` ad hoc and runs `flake8 .`. There is **no ruff step in any workflow**.

**Direction of the fix: add `flake8` to the dev extra, drop `ruff`.** The repo was unanimous apart from the dev extra — CI (`ci.yml:29-32`), the `Makefile` (`lint:` runs `flake8`), `AGENTS.md`, `CONTRIBUTING.md` (twice), `SECURITY.md`, and `.github/pull_request_template.md` all name flake8. `ruff` was invoked by **nothing**: no CI step, no make target, no `[tool.ruff]` config, no reference in any doc. So this was one stale dependency disagreeing with seven consumers, not a real choice of linter, and the cheap correct fix is to make the odd one out agree. **No lint gate is dropped**, and no doc needed changing — the docs were already right and the deps were wrong.

Pinned to `==7.1.0`, CI's exact version, so the local gate predicts the remote one; an unpinned pycodestyle can flag code CI accepts or accept code CI flags, which is the same class of "the gate passes locally and fails on CI" problem this branch exists to remove. The comment in `pyproject.toml` says to bump it alongside the `ci.yml` pin.

## R1-3 — Judged on merit

Nothing else changed. Specifically **not** done, and why:

- **The reviewer's `test_parallel_processes_write_atomically` observation** (run 3, `assert 450 == 4 * 150`): agreed it is not this branch's problem — `tests/unit/analytics/` is untouched here and it passed 6/6 in isolation and on runs 4 and 5. Load-sensitive multi-process test, deserves its own ticket. **Deferred**, and deliberately not fixed here: it is unrelated to this diff and would expand a de-flaking PR into the analytics recorder.
- No other refactoring in `loop.py`. The `finally` is now the shape it needs to be for the invariant and nothing more.

---

## Gates at the new head (`6a756cc9`)

| Gate | Result |
|---|---|
| `flake8 .` (7.1.0, from the newly fixed dev extra) | clean, exit 0 |
| `black --check .` (26.1.0) | 531 files unchanged, exit 0 |
| `isort --check .` (5.13.2, repo config, not `--profile`) | clean, exit 0 |
| `pytest tests/unit` run 1 | **6883 passed**, 7 skipped, 0 failed (96s) |
| `pytest tests/unit` run 2 | **6883 passed**, 7 skipped, 0 failed (99s) |
| `pytest tests/unit/harness` | 320 passed |

6883 rather than the reviewer's 6882 is the one added regression test. Both full runs used `env -u NO_COLOR TERM=xterm-256color` per `AGENTS.md`. The `flake8` run is itself evidence for R1-2: it was run from a venv built by `uv pip install -e ".[dev]"`, which is the thing that did not work before this commit.

**On pyright, plainly:** I did not run it locally. This host has been unusable for whole-tree pyright — the reviewer had two runs killed at 58 minutes with peer agent sessions holding concurrent pyright processes at load average 45-98, and adding a third 30-minute contended run would produce a less trustworthy signal than CI's, not a more trustworthy one. **CI's `type-check` job is authoritative here** and it passed on this exact head (`6a756cc9`) in 1m13s on an uncontended runner. I am relying on it rather than claiming a local result I did not produce.

## CI at `6a756cc9`

All 7 jobs green ([run 32938467001](https://github.com/damianvtran/local-operator/actions/runs/32938467001)): `lint` ✓, `type-check` ✓ (1m13s), `test (3.12)` ✓, `test (3.13)` ✓, `pip-audit` ✓, `cli-sanity` ✓, `server-sanity` ✓.

The `test` and `type-check` jobs install `pip install -e ".[dev]"`, so this run also exercises the R1-2 dependency change on CI's own path rather than only in my local venv.

**New head: `6a756cc9`.** Not merging — reporting ready.

damianvtran:
### Agent review — round 2

**Reviewer:** `reviewer` subagent (independent of the implementer), model `anthropic/claude-opus-5`
**Scope:** `77b526d6..5cbe7f5e`

Freshness delta review only. Round 1 was clean and terminal against `77b526d6`; nothing it cleared is re-litigated here. Reviewed in a dedicated worktree off `5cbe7f5e`, never in the shared checkout.

The delta is two commits (`b2d6f770` R1-1, `5cbe7f5e` R1-2) plus a rebase of the four reviewed commits onto `origin/main` (which advanced 7 commits, 0.36.0 → 0.38.1).

---

#### 1. R1-1, the cancellation fix — verified correct, both failure modes

I did not take the reported numbers on trust. I extracted the pre-fix helper verbatim from `77b526d6` and the post-fix helper from `5cbe7f5e` into a standalone harness and swept the cancel-at-every-turn-offset space myself: offsets 0–11, 25 repetitions each, under two delivery shapes (item pre-queued before the call, and item delivered mid-flight), 300 trials per cell. Every trial accounts for the item exactly: `delivered` (returned to caller) vs `in_queue` (drained afterwards), so a loss and a duplicate are distinguishable rather than both reading as "fine".

```text
PRE-FIX  [item pre-queued          ] trials=300  delivered_ok=200 requeued_ok= 25 LOSS= 75 DUP=0
            loss at turn offsets: [1, 2, 3]
PRE-FIX  [item delivered mid-flight] trials=300  delivered_ok=225 requeued_ok=  0 LOSS= 75 DUP=0
            loss at turn offsets: [0, 1, 2]
POST-FIX [item pre-queued          ] trials=300  delivered_ok=200 requeued_ok=100 LOSS=  0 DUP=0
POST-FIX [item delivered mid-flight] trials=300  delivered_ok=225 requeued_ok= 75 LOSS=  0 DUP=0
```

(a) **No item lost on caller cancellation** — confirmed. The pre-fix helper strands the item in a contiguous band of turn offsets (a real window, not a single knife-edge); the post-fix helper is 0/600.

(b) **No duplicate on the success path** — confirmed, and this was the failure mode I most expected to find, since a reclaim that fires on a delivered item would hand back a copy and produce a second `ToolExecutionEndEvent`. `DUP=0` across all 1200 trials. The mechanism holds up on reading: `delivered` is decided inside the `try` immediately after the `wait`, and `reclaim()` is guarded by `if not delivered`, so the success path cannot reach it. Deciding `delivered` after the fact — the obvious simplification — is exactly what would introduce the duplicate, and the comment at `loop.py:206-211` says so.

**Red/green reproduction.** In a second worktree I replaced only the helper body with the verbatim pre-fix version and ran the new regression test: `1 failed` 3/3, failing on `assert 1 == qsize()` where `qsize()` is 0 — the item consumed and dropped. Against `5cbe7f5e` the whole file is `62 passed`, 3/3. The test is genuinely deterministic (it spins until `queue.empty()` proves the getter dequeued, then cancels), not a probabilistic bet, and it asserts identity (`is item`) plus no leaked contestant tasks, so a reclaim that put back the wrong object would not satisfy it.

One observation, not a defect (see MINOR-1 below): the `except BaseException: if delivered: reclaim(); raise` arm is dead code as written.

#### 2. R1-2, the dependency change — safe, no residue

- CI genuinely runs flake8 and has no ruff step: `.github/workflows/ci.yml:29` installs `flake8==7.1.0`, `:31-32` runs `flake8 .`. The dev extra now pins the identical version.
- Repo-wide sweep for `ruff` returns only the three occurrences inside the new explanatory comment in `pyproject.toml`, plus `docs/REWRITE.md:436`. That doc line is **pre-existing on `origin/main`** and this PR does not touch that file, so it is not residue introduced here — it is a stale line in a historical design doc. Flagged as NIT-1, not a blocker.
- No pre-commit config exists, no `[tool.ruff]` section, and `Makefile:105`, `CONTRIBUTING.md:160/242`, `SECURITY.md:14` and the PR template all already named flake8.
- **Clean-install claim verified independently**, not inferred: fresh `uv venv --python 3.12`, then `uv pip install -e ".[dev]"`. Before I had a chance to install anything else the venv reported `No module named flake8`; after the install it reports `flake8 7.1.0 (mccabe 0.7.0, pycodestyle 2.12.1, pyflakes 3.2.0)`, and no `ruff` binary or module is present. The documented gate now runs on a fresh venv.
- I also ran `flake8 .` from that clean venv at the pinned 7.1.0 (not just the shared venv's 7.3.0) — clean. Worth noting because the pin's whole justification is that versions can disagree; at 7.1.0 and 7.3.0 this tree is clean either way.

#### 3. The rebase is clean

`git range-diff c61b58ac..77b526d6 27409b53..792bb9d0` reports all four reviewed commits as `=` — byte-identical after replay, no dropped or altered hunk:

```text
1:  8c7f171e = 1:  015b4a3a fix(tests): replace timing bets with conditions
2:  be954141 = 2:  23883008 fix(harness): preserve events at abort timeout
3:  7a5a9728 = 3:  aec11202 fix(tui): anchor a history prepend to its block, not to height growth
4:  77b526d6 = 4:  792bb9d0 test(tui): make the prepend anchor regression discriminate
```

Against the merge-base `27409b53` the branch touches exactly 12 files — the 11 from round 1 plus `pyproject.toml` for R1-2. No file resurrected from the 7 upstream commits, no binary/evidence blobs, no stray additions. The `pyproject.toml` diff is the dev-extra hunk and nothing else.

**Version:** the branch carries no bump and simply inherits main's `0.38.1` (`pyproject.toml:7` identical to `origin/main:pyproject.toml:7`). No conflict, consistent with "this is not a release".

#### 4. Nothing regressed

Gates run over the whole tree, at CI's pinned versions (`uvx` cannot spawn on this host, so I installed the pins into the clean venv rather than skipping or substituting):

```text
flake8 7.1.0 (pinned, clean venv)      clean
flake8 7.3.0 (shared venv)             clean
black 26.1.0 --check .                 551 files would be left unchanged
isort 5.13.2 --check .   (repo config) clean, skipped 1 file
```

Full unit suite, three consecutive runs on this head:

```text
7010 passed, 7 skipped, 441 warnings in 120.26s
7010 passed, 7 skipped, 441 warnings in 101.47s
7010 passed, 7 skipped, 441 warnings in  94.55s
```

Actual is **7010 passed / 7 skipped**, not the ~6883 the brief anticipated; the difference is upstream tests the rebase pulled in, not anything added here. All three runs fully green — including `test_the_loop_stays_responsive_while_several_subagents_run`, `test_silenced_console_keeps_the_browser_off_the_terminal` and `test_parallel_processes_write_atomically`, the three known-remaining flakes, which did not fire in any of the three. They remain out of scope and I make no claim they are fixed; the host was simply quieter than when they were observed.

**pyright: I relied on CI, not a local run,** as directed. CI `type-check` passes on `5cbe7f5e` (2 runs, 57s and 1m13s). Every check on the current head is green — lint, type-check, pip-audit, build, cli-sanity, server-sanity, test 3.12 and 3.13 — with one duplicate `test (3.13)` job from the parallel `pull_request` run still finishing at the time of writing; its `push`-triggered twin on the identical SHA already passed in 9m49s.

#### 5. The round-1 mutation property still holds after the rebase

Deleted the 12-line post-drain flush loop (`loop.py:1696-1707`) in a throwaway worktree and ran the harness suite 5 times:

```text
1 failed, 61 passed   (x5, identical)
FAILED tests/unit/harness/test_loop.py::test_a_late_parking_tool_is_not_robbed_of_its_end_event_mid_backfill
AssertionError: c2 was announced as started and ended 0 times; every started call needs exactly one end (ends=['c1'])
```

Deterministic, same test every time. The rebase did touch that file's neighbourhood and the property survived it. (My first mutation attempt matched the wrong `while True:` and deleted ~1300 lines; I discarded it and re-cut the mutation precisely, so the result above is from the intended 12-line deletion only.)

---

### Findings

**BLOCKER:** none.
**MAJOR:** none.

**MINOR-1 — the join-window reclaim arm is unreachable** (`local_operator/harness/loop.py:230-240`). The `except BaseException: if delivered: reclaim(); raise` block cannot execute. The `await task` inside it is wrapped in `contextlib.suppress(asyncio.CancelledError)`, so a cancellation delivered at a join is swallowed by the suppressor and never propagates out of the `for` loop to the `except`. I instrumented a copy of the helper and hit every path over 720 trials (cancel at offsets 0-14, 1/2/3/5 repeated cancels, both delivery shapes): `normal` 696, `not-delivered reclaim` 144, `except_branch` **0**. I also confirmed the observable behaviour there is unchanged from pre-fix: cancelling while parked in the joins returns the value and swallows the cancellation, identically in both versions, with the item accounted for exactly once. So this is defensive code that costs nothing and loses nothing — but the comment above it ("this is the last window in which the item can be stranded") asserts a live guarantee that the code as written does not provide. Suggested one-line fix: either narrow the suppressor so a cancellation at a join can reach the handler, or reword the comment to say the arm is belt-and-braces against a future change that removes the suppression. Not worth a round on its own.

**NIT-1 — stale ruff reference in a design doc.** `docs/REWRITE.md:436` still reads ``- Gates: `ruff check` + `ruff format` + `pyright` + `pytest`.`` Pre-existing on `origin/main` and untouched by this PR, so strictly out of scope — but R1-2 is precisely the "make the repo agree on one linter" change, and this is the last line that disagrees. A follow-up, not a change to make here.

**NIT-2 — reclaim reorders to the tail.** Already documented honestly in the docstring, and the sole caller abandons the queue on that path, so it is correct as scoped. Noting only so a future second caller does not inherit the FIFO assumption silently.

---

### Verdict: **clean — TERMINAL**

No blocker and no major findings against `5cbe7f5e`. Both remediations do what they claim, verified independently rather than from the reported numbers: the cancellation fix closes the loss without opening a duplicate, and the dependency change makes the documented gate actually installable with no residue of the removed tool. The rebase replayed the reviewed work byte-identically, adds nothing, and inherits `0.38.1` cleanly. Gates and three full-suite runs are green locally, CI is green on the current head.

MINOR-1 and the two nits are follow-ups, not grounds for another round. **Nothing blocks merge.**

PR #318, opened 2026-08-26
feat(hub): fan resume out to multiple subagents

damianvtran:
## Summary

`hub op='resume'` now fans **one** resume message out to **several** subagents in a single call. Previously resume shared the "exactly one subject" guard with `ask` and `peek` and refused whenever more than one target matched, so recovering a batch of failed children — the common case when a provider stall knocks out a whole `task` fan-out at once — meant one hub call per child, which is slow and wasteful.

The guard was wrong for resume specifically. `ask` has one reply to read and `peek` has one transcript window to render, so they genuinely address one subject. Resume does not: each resume spawns an **independent** new job replaying that target's own transcript, so N targets produce N separate children with no shared reply and no shared transcript to collide. Fanning out is exactly what resume should do.

## What changed

- `local_operator/tools/builtin.py` (`_execute_hub_parent`): dropped `"resume"` from the single-subject guard tuple (now `ask`/`peek` only). Replaced the single-target resume block with a loop over every target, building a per-target receipt modelled on the existing send/steer/cancel block: a `resume: N/M subagent(s)` header, one line per target naming the new job id on success or the reason on failure, `_hub_targets` errors appended, and the transcript-replay/`wait` guidance stated once in the footer. `details` now carries lists (`job_ids`, `resumed_from`, `acted`) plus the `useless` flag, mirroring the other multi-target ops. Single-target resume behaviour is preserved through the same loop (one clean receipt line).
- Copy updated to match the new capability: the `op` and `to` field descriptions, the tool-level description, the `_hub_list` roster hint, and the hub paragraph in `prompts_md/system.md` — each now states resume can address a whole batch.
- `pyproject.toml`: **0.40.0 → 0.41.0** (minor; new capability).

## Testing evidence

Backend change, no UI. Exercised the real path through `_execute_hub_parent` with the harness comms in `tests/unit/harness/test_comms.py` (5 new tests):

- Resuming two stopped children in one call spawns two new jobs and the receipt names both.
- A mixed batch (one resumable + one refusing) yields one success line and one failure line carrying the reason.
- A pure-failure batch is flagged `useless` so the model does not act on it as if heard.
- `peek` still refuses `len(ids) > 1` (guard regression); the matching `ask`-refuses test already existed.

Command outputs (whole tree, exactly as CI runs them, from the worktree with `local_operator.__file__` confirmed resolving into it):

```
$ .venv/bin/python -m flake8 .                        # exit 0
$ uvx --from black==26.1.0 black --check .            # 551 files unchanged
$ python -m isort --check .   (isort 5.13.2, repo cfg) # exit 0
$ python -m pyright --pythonpath .venv/bin/python .    # 0 errors, 0 warnings
$ python -m pytest tests/unit/harness/test_comms.py -q # 104 passed
```

Note: `uvx isort==5.13.2` cannot spawn isort in this sandbox (an environment issue, not a lint failure); validated with the identically-pinned isort 5.13.2 as a module and confirmed the version. CI runs `isort --check .` and will exercise the literal form.

## Full changelog

https://github.com/damianvtran/local-operator/compare/v0.40.0...dev-resume-fanout

### DIFF
--- a/local_operator/prompts_md/system.md
+++ b/local_operator/prompts_md/system.md
@@ -137,8 +137,9 @@ doing without spending its attention, which is the fast way to check on a
 quiet child; `hub` also sends it a note, asks it a question and waits for its
 answer (a busy child finishes its current step before replying, so give it
 minutes, or peek instead of re-asking), steers it onto a different course,
-cancels it, or resumes a stopped one against its own transcript. Address them
-by job id, by label, or `"all"`. Inside a subagent, `hub` is how you reach the
+cancels it, or resumes a stopped one (or a batch of them at once) against its
+own transcript. Address them by job id, by label, or `"all"`. Inside a
+subagent, `hub` is how you reach the
 agent that delegated to you — answer its questions, and speak up unprompted
 when you are blocked or the task turns out to be wrong.
 
--- a/local_operator/tools/builtin.py
+++ b/local_operator/tools/builtin.py
@@ -6943,7 +6943,7 @@ class HubParams(BaseModel):
             "(becomes part of its instructions). pause: stop it now but keep it "
             "resumable. cancel: stop it for good. resume: relaunch a stopped, paused or "
             "failed subagent against its own transcript so it continues where it left "
-            "off."
+            "off; names several targets to fan one message out to a whole batch at once."
         )
     )
     # A plain array, NOT ``str | list[str]``: pydantic renders a union as
@@ -6962,8 +6962,9 @@ class HubParams(BaseModel):
         description=(
             "Who to address: job ids from 'task'/'jobs'/'hub op=list', subagent "
             'labels, or ["all"] for every running subagent. Several ids address '
-            "several subagents. 'ask', 'resume' and 'peek' take exactly one. Omit "
-            "for op='list', which addresses nobody."
+            "several subagents. 'ask' and 'peek' take exactly one; 'resume' fans one "
+            "message out to every target you name, so a batch of failed subagents can "
+            "be resumed in a single call. Omit for op='list', which addresses nobody."
         ),
     )
 
@@ -7118,7 +7119,8 @@ def _hub_list(tool_call_id: str, comms: Any) -> ToolResult:
         lines.append("")
         lines.append(
             "Resume one with hub op='resume' and its JOB id, plus an instruction for "
-            "what to do next. The transcript id above is not a job id: it opens the "
+            "what to do next \u2014 or name several JOB ids to resume a whole batch in "
+            "one call. The transcript id above is not a job id: it opens the "
             "child's history for reading and starts no agent."
         )
     return _text(
@@ -7277,9 +7279,15 @@ async def _execute_hub_parent(
             "hub",
             "; ".join(errors) or "no subagent matched; use 'jobs' to list them.",
         )
-    # A question, a resume and a peek each have exactly one subject, so they
-    # refuse a fan-out rather than silently acting on the first match.
-    if params.op in ("ask", "resume", "peek") and len(ids) > 1:
+    # A question and a peek each have exactly one subject — one reply to read,
+    # one transcript window to render — so they refuse a fan-out rather than
+    # silently acting on the first match. Resume is deliberately NOT here: each
+    # resume spawns an INDEPENDENT new job replaying that target's own
+    # transcript, so N targets produce N separate children with no shared reply
+    # and no shared transcript to collide. Fanning one message out to a whole
+    # batch of stopped/failed children (e.g. after a provider stall) is exactly
+    # what resume should do, so it falls through to the loop below.
+    if params.op in ("ask", "peek") and len(ids) > 1:
         return _error(
             tool_call_id,
             "hub",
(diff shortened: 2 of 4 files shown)


damianvtran:
### Agent review — round 1

Reviewer: reviewer subagent (team lopdev), running as claude-opus-4
Scope: origin/main..d806ff3abaa8645a4e61c8f3652c4d3d59251a12

Reviewed the diff (`builtin.py`, `test_comms.py`, `system.md`, `pyproject.toml`), traced `comms.resume`/`_hub_targets`/`_live_twin` for concurrency and partial-failure safety, and grepped the tree for readers of the old scalar `details` shape.

**No blockers, no majors.** The fan-out loop is correct and faithfully mirrors the send/steer/cancel receipt conventions. Details below.

#### What I verified
- **Fan-out & receipt aggregation** (`builtin.py:7322-7376`): `acted` counts receipts whose error is `None`; header is `resume: {len(acted)}/{len(resumed)}`; `useless=not acted` is set on the result *and* mirrored into `details["useless"]` — identical shape to the send/steer/cancel block at `7378-7411`. `useless` is true only when nothing resumed. Correct.
- **Guard** (`builtin.py:7290`): the refuse-fan-out tuple is now `("ask", "peek")`; `resume` correctly falls through. ask and peek still reject `len(ids)>1`. A new peek test (`test_the_tool_refuses_to_peek_several_children_at_once`) locks this in. Guard was not weakened.
- **No downstream consumer of the old scalar shape.** Grepped `resumed_from`/`job_ids` across `local_operator/` (esp. `tui/`, `session/`): the only readers are the new tests. Nothing in the TUI/session renders `details["job_id"]`/`["resumed_from"]` for resume. The author's claim holds — confirmed independently.
- **Concurrency / partial failure**: each `comms.resume` is sequential; `_live_twin` (comms.py:1414) blocks a second agent on one transcript dir, and `_hub_targets` (builtin.py:7075) dedupes ids so `["all","<id>"]` can't double-resume one child. `record.paused` is cleared *only* on a successful spawn (comms.py:1478), so a failed target in a mixed batch leaves the roster untouched — no inconsistent state. Safe.
- **Tests are real, not shallow**: the multi-target test monkeypatches `run_subagent` with a spy and asserts both children spawn as distinct new jobs (`resumed-1`/`resumed-2`) with the receipt text, `job_ids`, `resumed_from`, `acted`, and `useless`. The mixed case uses a genuinely-running child that refuses with "still running", and asserts success and failure sit side-by-side with `acted==1`, `useless is False`. The all-fail case asserts `acted==0`, `useless is True`. Good coverage of the branch matrix.
- **Copy**: field/tool/roster/system.md updates describe fan-out accurately without over-promising (they correctly keep ask/peek single-subject). Comments follow repo why/constraints/context convention and are notably thorough.

#### Quality gates (run against worktree /tmp/lop-resume-fanout)
- `flake8 .` — **pass** (clean, whole tree)
- `black==26.1.0 --check .` — **pass** (551 files unchanged)
- `isort==5.13.2 --check .` (venv, repo config) — **pass**
- `pyright` (changed files) — **pass** (0 errors). Full-tree pyright times out in this sandbox; verified green via CI `type-check` job on this exact head (d806ff3a).
- `pytest tests/unit/harness/test_comms.py -q` — **pass** (104 passed)

#### Minor / observations (non-blocking, follow-ups only)
- **N1 — `details["job_ids"]` semantics differ across ops** (`builtin.py:7365` vs `7401`). In the send/steer/cancel block, `job_ids` is the full *target* id list (`ids`); for resume, `job_ids` is only the *new* successfully-spawned job ids (failures excluded), with sources in `resumed_from`. This is the sensible shape for resume (a new launch produces new ids), and no consumer reads it generically, so it's harmless — but a future reader assuming `job_ids` means "targets" uniformly across ops could be surprised. A one-word key comment would remove the ambiguity.
- **N2 — dead `"resume: nothing to do"` branch** (`builtin.py:7337`). `resumed` is built from `ids`, which the earlier `if not ids` guard guarantees is non-empty, so the `else` header is unreachable. It harmlessly mirrors the same defensive pattern in the delivery block (`7392`), so keeping it for parity is defensible; noting only that it can't execute.

#### Verdict
**clean** — no blocker, no major. This round is **TERMINAL**; N1/N2 are optional follow-ups, not a reason for another round. The round is fresh against head d806ff3a.

damianvtran:

Implementer: coder subagent (team lopdev), running as claude-opus-4
Scope: origin/main..41d08f403c423af7d4822c1c0c21c51d1407a356

**N1 — `details["job_ids"]` semantics differ across ops** — **Fixed** in `41d08f403c423af7d4822c1c0c21c51d1407a356`. Added an inline comment at the resume `details` block noting that `job_ids` here holds the NEW successfully-spawned ids (sources in `resumed_from`), distinct from the target-id meaning in the send/steer/cancel block below. No behaviour change, comment only.

**N2 — dead `"resume: nothing to do"` branch** — **Deferred (nit, no ticket).** The unreachable `else` header is kept deliberately for structural parity with the identical defensive pattern in the delivery block (`builtin.py:7392`). Diverging one of two mirrored blocks to shave an unreachable line would cost more in inconsistency than it saves. This is a non-blocker nit, so no ticket is filed.

Gates re-run on the touched file at new head `41d08f40`: `flake8 local_operator/tools/builtin.py` pass; `pytest tests/unit/harness/test_comms.py -q` — 104 passed.

damianvtran:
### Agent review — round 1 (freshness re-confirmation)

Reviewer: reviewer subagent (team lopdev), running as claude-opus-4
Scope: origin/main..41d08f403c423af7d4822c1c0c21c51d1407a356

Re-verified the delta d806ff3a..41d08f40 directly: it is a single 3-line inline comment in `local_operator/tools/builtin.py` (at the resume `details` block, ~line 7362) clarifying that `job_ids` here holds the NEW spawned ids with sources in `resumed_from`, unlike the send/steer/cancel block where `job_ids` is the target list. This is exactly the N1 clarification I raised. No executable code changed.

The round-1 verdict stands: **clean, TERMINAL** against the new head 41d08f40. My round-1 findings (no blocker, no major; N1 addressed by this comment, N2 an optional follow-up) and gate results carry forward unchanged, since nothing executable moved.

PR #319, opened 2026-08-26
fix(extension): walk through ignored AX wrapper nodes in snapshot compaction

damianvtran:
## What

Fixes the still-broken live snapshot: on the real paired Chrome (0.1.1 build, post-#313), `snapshot` returned exactly one AX line — `- RootWebArea "…" [e1]` — on every page, twice in a row.

**Root cause is not tree population and not the hidden tab.** Reproduced against throwaway headful Chrome (Chrome for Testing 145 with the production `chrome.debugger` path in an MV3 probe extension, plus branded Chrome 151 over raw CDP): a hidden `active:false` tab's `Accessibility.getFullAXTree` returns the **full tree** — 48 nodes on a rich local page, 14 on example.com — with or without `Accessibility.enable`, immediately, even in a minimized window and even mid-load on a 52k-node page. The hidden-tab/deferred-a11y hypothesis is dead: `ProgressiveAccessibility` (incl. `disable_on_hide`) changed nothing.

The actual bug is in `compactAX` (`extension/src/ax-compact.ts`): Chrome wraps every real page's content in **ignored generic containers** (html/body surface as `role: "none", ignored: true`) directly under the RootWebArea, and the early `if (node.ignored) return` pruned those wrappers' **entire subtrees** — the whole page. PR #313's headless validation counted *raw protocol nodes* (19), never the compacted output, which is why it looked fixed.

Fix: treat ignored nodes as transparent — emit no line, keep the depth, walk through to children. Refs stay `backendDOMNodeId`s, the `axQueue` serialization (c628cee4) and the enable→read→disable window are unchanged (enable kept as documented CDP hygiene; comments corrected to stop attributing the one-line bug to it).

## Testing evidence

`pnpm typecheck` clean, `pnpm test` 16/16 (new pure test mirrors the exact ignored-wrapper shape captured from Chrome 151), `pnpm build` clean.

Live before/after in throwaway **headful** Chrome (`/tmp` profile, launched backgrounded and off-screen; the paired Chrome and bridge daemon untouched), production CDP sequence — `tabs.create({active:false})` → attach → `Runtime`/`Log` enable → navigate → webNavigation settle → `Accessibility.enable` → `getFullAXTree` → `disable` — run inside an MV3 service worker via `chrome.debugger`, then both the pre-fix and post-fix `compactAX` bundles applied to the same live tree (tab confirmed `active: false`):

| Page | Raw nodes | Before | After |
|---|---|---|---|
| https://example.com | 14 | 1 line / 1 ref | **6 lines / 2 refs** (heading, text, `link "Learn more" [e2]`) |
| Local rich page (headings/nav/form/footer over `python3 -m http.server`) | 48 | 1 line / 1 ref | **26 lines / 10 refs** (banner/nav/main/contentinfo landmarks; links, textbox, checkbox, combobox+options, `button "Submit Button" [e10]`) |

Second consecutive snapshot cycle (disable → enable → read) returns the identical tree — matches the live "two snapshots, same single node" symptom being deterministic, not a warmup race.

## Not in this PR

- Live re-verification on the paired Chrome (requires the user's bridge; deliberately untouched — the fix is in `compactAX`, which the live probes exercised byte-for-byte via the compiled bundle).
- Extension version bump / store release.

### DIFF
--- a/extension/src/ax-compact.ts
+++ b/extension/src/ax-compact.ts
@@ -20,7 +20,23 @@ export function compactAX(nodes: AXNode[], epoch: number): { snapshot: string; r
   const lines: string[] = [];
   let sequence = 0;
   function visit(node: AXNode, depth: number): void {
-    if (node.ignored) return;
+    // An ignored node is excluded from the accessible tree but its subtree is
+    // NOT: Chrome wraps every real page in ignored generic containers (html and
+    // body surface as role "none", ignored: true) sitting directly under the
+    // RootWebArea. Returning early here therefore pruned the ENTIRE page and
+    // produced the live one-line snapshots ('- RootWebArea "…" [e1]') that
+    // survived the Accessibility.enable fix — confirmed against real headful
+    // Chrome 151 and Chrome for Testing 145, where getFullAXTree returns a
+    // full tree (48 nodes on the repro page) whose nodes 2..3 are ignored
+    // wrappers. Treat ignored nodes as transparent: skip the line, keep the
+    // depth, and walk through to their children.
+    if (node.ignored) {
+      for (const child of node.childIds ?? []) {
+        const found = byId.get(child);
+        if (found) visit(found, depth);
+      }
+      return;
+    }
     const role = String(node.role?.value ?? "");
     const name = String(node.name?.value ?? "").trim();
     const focusable = node.properties?.some((property) => property.name === "focusable" && property.value.value === true);
--- a/extension/src/ax-compact.ts
+++ b/extension/src/ax-compact.ts
@@ -19,7 +19,13 @@ export function compactAX(nodes: AXNode[], epoch: number): { snapshot: string; r
   const refs: Record<string, SnapshotRef> = {};
   const lines: string[] = [];
   let sequence = 0;
+  // Guard against cyclic or duplicated childIds in a malformed AX payload:
+  // the walk trusts protocol data, and without this a cycle would recurse
+  // forever inside the extension's service worker (review round 1, MINOR-1).
+  const visited = new Set<string>();
   function visit(node: AXNode, depth: number): void {
+    if (visited.has(node.nodeId)) return;
+    visited.add(node.nodeId);
     // An ignored node is excluded from the accessible tree but its subtree is
     // NOT: Chrome wraps every real page in ignored generic containers (html and
     // body surface as role "none", ignored: true) sitting directly under the
@@ -29,7 +35,10 @@ export function compactAX(nodes: AXNode[], epoch: number): { snapshot: string; r
     // Chrome 151 and Chrome for Testing 145, where getFullAXTree returns a
     // full tree (48 nodes on the repro page) whose nodes 2..3 are ignored
     // wrappers. Treat ignored nodes as transparent: skip the line, keep the
-    // depth, and walk through to their children.
+    // depth, and walk through to their children. This is deliberate even for
+    // an ignored node carrying `focusable` (e.g. inside aria-hidden): ignored
+    // means excluded from the accessible tree, so it emits no line and no ref
+    // despite focusable being a ref trigger for rendered nodes below.
     if (node.ignored) {
       for (const child of node.childIds ?? []) {
         const found = byId.get(child);
--- a/extension/src/commands/snapshot.ts
+++ b/extension/src/commands/snapshot.ts
@@ -5,8 +5,8 @@ import { setRefs } from "../state";
 // Serializes the enable→read→disable window below. The worker dispatches
 // daemon frames fire-and-forget (worker.ts onmessage), so two concurrent
 // snapshots could otherwise interleave as enable/enable/read/DISABLE/read —
-// the second read then hits a disabled a11y engine, the exact one-node bug
-// this file fixes. The daemon pipelines commands per tab today, so this is
+// the second read then hits a disabled a11y engine and risks a degraded
+// tree. The daemon pipelines commands per tab today, so this is
 // latent, but the extension should not depend on that scheduling promise.
 // A promise chain (not a refcount) keeps it simple: snapshots are rare and
 // heavy, so serializing them costs nothing observable. Failures are swallowed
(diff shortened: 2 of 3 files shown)


damianvtran:
### Agent review — round 1

Reviewer: independent reviewer subagent (Claude / Anthropic model family) — not the implementing agent
Scope: origin/main..3f35de22ca8c233bf7f2d87e6f489a9c620b2926

**Verification performed (reproduced, not inferred):**
- `pnpm typecheck` ✓, `pnpm test` ✓ (16/16), `pnpm build` ✓ in `extension/` at head.
- Regression-test validity: reverted `src/ax-compact.ts` to the base (`f6e7d027`) version and re-ran the suite — the new test **fails on the old code** with `AssertionError: ignored wrappers must not hide the page — 1 !== 3`, i.e. it encodes exactly the one-line-snapshot failure shape. Restored the file; worktree clean at head.

**Correctness review of the transparent walk (`extension/src/ax-compact.ts:34-39`):**
- *Depth accounting*: an ignored node recurses with the same `depth`, matching the existing convention for rendered-but-uninteresting nodes (`interesting ? depth + 1 : depth` at line 56). Children of an ignored wrapper indent relative to the last rendered ancestor — the test pins this (`RootWebArea` at depth 0 → heading/link at 2 spaces through two ignored levels).
- *Cycles*: CDP `Accessibility.getFullAXTree` returns a tree keyed by nodeId; `visit` has no cycle guard, but that property is unchanged from the base code — the ignored branch adds no new traversal risk (see MINOR-1).
- *Ignored-but-focusable*: an ignored node never emits a line or ref even if it carries `focusable` (e.g. a focusable element inside `aria-hidden`). That matches AX semantics — ignored nodes are excluded from the accessible tree — and is no regression: the old code pruned them (plus their subtrees) entirely (see NIT-1).
- *Ref stability*: refs remain `e${++sequence}` → `backendDOMNodeId`, epoch-scoped (test asserts `refs.e2 = { backendNodeId: 9, epoch: 3 }`). Ignored nodes never received refs before and still don't, so sequence numbering changes only where previously-pruned subtrees now correctly appear — the fix itself. Consumers (`input.ts:35`, `scroll.ts:64`) resolve by stored map, unaffected.
- *snapshot.ts / #313 hygiene*: `axQueue` serialization and the enable→read→disable window are code-identical to base — only comments changed, and they now honestly de-claim the earlier "missing enable caused the one-node bug" attribution while keeping enable as documented hygiene. The comment does not overclaim.

**Findings:**

- **MINOR-1** — `extension/src/ax-compact.ts:22-58` — `visit` has no visited-set/cycle guard; a malformed AX payload with cyclic or duplicated `childIds` would recurse forever or emit duplicates. Pre-existing (not introduced by this PR); the ignored branch merely inherits it. Follow-up, not a reason for a round: a `Set<string>` of visited nodeIds is a one-liner.
- **NIT-1** — `extension/src/ax-compact.ts:34` — the comment could note the ignored-but-focusable case explicitly (such nodes are intentionally transparent and un-refable), since `focusable` is otherwise a ref trigger two branches below and a future reader may wonder if the early return loses refs.

No BLOCKER or MAJOR findings.

**Verdict: `clean` — this round is TERMINAL.** The root-cause analysis is correct and evidenced, the fix is minimal and matches AX semantics, the regression test demonstrably fails on the base code, and PR #313's serialization/hygiene behavior is untouched. MINOR-1/NIT-1 are follow-ups, not remediation items.

damianvtran:

Implementing agent reply to [round 1](https://github.com/damianvtran/local-operator/pull/319#issuecomment-5427132053) (verdict `clean`, terminal at 3f35de22). Both non-blocking findings were cheap and safe, so they are **fixed** in one follow-up commit rather than deferred: **861743fb**.

- **MINOR-1 (no cycle guard in `visit`) — FIXED in 861743fb.** A `visited` `Set<string>` of nodeIds now short-circuits re-entry, so a malformed payload with cyclic or duplicated `childIds` terminates instead of recursing forever in the service worker. A well-formed tree visits each nodeId exactly once, so output is unchanged — proven two ways: the new pure test (`AX compaction terminates on cyclic and duplicated childIds`, covering both a 2→3→2 cycle and a duplicated child) and re-validation of the guarded compiled bundle against fresh live hidden-tab `getFullAXTree` dumps from the throwaway headful Chrome: rich page 48 raw → 26 lines/10 refs, example.com 14 raw → 6 lines/2 refs — byte-identical to the pre-guard round-1 evidence.
- **NIT-1 (document ignored-but-focusable) — FIXED in 861743fb.** The transparent-walk comment now states explicitly that an ignored node emits no line and no ref even when it carries `focusable` (e.g. inside `aria-hidden`), since `focusable` is otherwise a ref trigger for rendered nodes below.

`pnpm typecheck` clean, `pnpm test` 17/17, `pnpm build` clean at 861743fb; CI running on the new head. No other changes — the round's terminal status stands.

damianvtran:
### Agent review — round 2

Reviewer: reviewer subagent (Claude, Anthropic model family) — independent of author
Scope: 3f35de22..861743fb (delta), verified at head 861743fb

Round 2 reviewed the remediation delta for round 1's MINOR-1 (cycle guard) and NIT-1 (ignored-but-focusable doc comment).

**Verification performed:**
- Delta touches exactly `extension/src/ax-compact.ts` and `extension/tests/pure.test.mjs`; nothing else changed.
- The `visited` guard keys on `nodeId` at the top of `visit()`, so cyclic and duplicated `childIds` terminate. On well-formed CDP AX trees (each node has one parent) no revisit ever occurs, so output is unchanged. The previous behavior for a duplicated childId was double emission — duplicate snapshot line plus a second ref to the same backendNodeId — which was itself a defect; suppressing it is a correction, not hidden content.
- The new test encodes a real cycle (2→3→2) plus a duplicated child list (`["2","2"]`) and asserts single emission with correct ref sequencing.
- The NIT comment is accurate: the ignored branch returns before the focusable check, so ignored-but-focusable nodes emit no line and no ref, as documented.
- Reproduced locally in `extension/`: `pnpm typecheck` clean, `pnpm test` 17/17 pass (new cycle test terminates in ~3 ms), `pnpm build` succeeds.

**Findings:** none.

**Verdict: clean — this round is TERMINAL.** No BLOCKER or MAJOR findings; no remaining minors or nits.

PR #320, opened 2026-08-26
fix(browser-bridge): client timeout must outlive the daemon's origin-prompt window

damianvtran:
## Summary

The session-side `BridgeClient` used a flat 30–35 s httpx timeout, but the daemon **deliberately** holds a command open for `base + ORIGIN_PROMPT_WINDOW_S` (65 s) while the extension shows its 60 s origin-approval popup (finding A3) — worst case ~95 s for `open`/`goto`. The client therefore died mid-prompt and reported **"browser bridge unreachable: the daemon at 127.0.0.1:PORT is not answering"** against a daemon that was healthy and waiting on the human. Evidence: a direct RPC with a 120 s timeout returns a typed `origin_denied` at exactly 60.0 s on the same daemon where the tool's 35 s client reports unreachable; a QA session (transcript `0ee4974ba84a`) burned an hour misdiagnosing this as a broken bridge.

## What changed

- **`protocol.py`** (the protocol's one hand-written source) now owns the whole timeout chain: `ORIGIN_PROMPT_TIMEOUT_MS = 60_000`, `COMMAND_TIMEOUTS`, and `ORIGIN_PROMPT_WINDOW_S` (now *derived*: deny + 5 s), with the invariant documented at the constants: **extension deny 60 s < daemon window 65 s < client budget**.
- **`daemon.py`** imports the constants from protocol instead of defining them.
- **`gen_ts.py`** emits `ORIGIN_PROMPT_TIMEOUT_MS` into `protocol.gen.ts`, and **`origins.ts`** consumes the generated constant (re-exported for its existing importers), so the 60/65 relationship cannot drift across the Python/TypeScript boundary.
- **`backend.py`**:
  - `client_timeout(method)` = `COMMAND_TIMEOUTS[method] + ORIGIN_PROMPT_WINDOW_S + 5 s` margin (unknown method → conservative max, 100 s), so the client outlives every legitimate daemon wait and the typed answer (`origin_denied`, `nav_timeout`, result) always reaches the session.
  - Honest error split: `httpx.ConnectError` (nothing accepted TCP → daemon truly gone) keeps the unreachable/restart message; a timeout **after** connecting now says the daemon accepted the command and it may be waiting on a site-permission decision in the extension popup, so the agent points the user at the popup instead of restarting a healthy daemon. Both stay `BridgeUnreachable` — the tool layer maps that type to `str(exc)` verbatim, so distinct messages are the established mechanism (typed `BridgeError` codes are reserved for wire errors per `format_error`).

[...]

PR #321, opened 2026-08-26
feat(browser): multi-tab surfaces so parallel sessions each own a tab

damianvtran:
## Summary

The browser-extension bridge stored exactly ONE surface, and `open` silently reused it if its tab was alive — so a second Local Operator session's `open` **stole the first session's tab mid-task**. This PR makes the bridge multi-tab: each session's fresh `open` creates its own tab, sessions resume and close only their own tab, a new `tabs` verb lists everything being driven, and a cap stops tab-spray.

### Extension (`extension/src`)
- **state.ts** — `surfaces: Record<token, StoredSurface>` in session storage replaces the single `surface`; each surface keeps its own nonce/epoch plus `createdAt`/`lastUsedAt`. Snapshot `refs` are keyed per surface token so one tab's snapshot cannot repoint another tab's click targets. `MAX_SURFACES = 8` with a typed `tab_limit` refusal. Map mutations are serialized (the daemon locks per tab, so different tabs' commands genuinely interleave in the worker now).
- **nav.ts** — `open` with no `tab` param always creates a NEW tab (the hijack is gone); `open` with a `tab` navigates exactly that surface (explicit resume). New `tabs` handler lists live surfaces with URL/title fetched from `chrome.tabs` at call time and prunes dead entries (best-effort debugger detach). `close` with a `tab` closes that surface only; bare `close` still closes a sole surviving tab and refuses—naming the live handles—when several are open. about:blank-first attach, origin gate, and per-tab log-capture arming unchanged.
- **cdp.ts** — `requireSurface` resolves the caller's exact token against the map: nonce anti-guessing is preserved per surface, and the `tab_closed` error shape is unchanged for missing/stale handles.
- **popup.ts** — minimal change: the Connected card shows `Driving N tabs` when several surfaces exist (top-rule colour semantics untouched).

### Protocol / daemon / tool
- `protocol.py` `METHODS` += `tabs`, `ErrorCode` += `tab_limit`; `protocol.gen.ts` regenerated (diff is exactly the new method + code).
- `daemon.py` — `tabs` timeout entry (status-class, no-tab global lock key, so a listing can't interleave with an open's map write).
- `builtin.py` — `tabs` action in `BROWSER_ACTIONS` and the tool schema/description; marks the caller's own handle `(yours)` and warns other sessions' tabs are read-only awareness. `open` recovers from a dead pinned tab by creating a fresh one. `tab_limit` maps to a "close one first / use tabs" diagnostic. cmux degrades `tabs` with the same typed error pattern as scroll/logs. Description documents: open (no tab) = new tab, per-session pinning, close your own tab when done.

## Testing evidence

**Static gates** (full tree, per AGENTS.md): flake8 ✅, black ✅, isort ✅, pyright ✅ (0 errors), `gen_ts --check` ✅.

**Unit suites**: extension `pnpm typecheck` ✅, `pnpm test` 16/16 ✅ (new pure test pins exact-token resolution, nonce anti-guessing, cap math), `pnpm build` ✅. Python `tests/unit` full suite **7030 passed** (the 4 `tests/unit/server/test_server_models.py` failures reproduce on clean `origin/main` — pre-existing, unrelated). New tests pin: METHODS ⇔ COMMAND_TIMEOUTS parity, `tab_limit` typed code, `tabs` listing/own-marking, discovery without an owned surface, dead-pin recovery on `open`, cmux degradation for `tabs`.

**End-to-end** (throwaway headful Chrome 151 launched background/off-screen on a /tmp profile — `open -g` + `--window-position=-3000,-3000`, extension loaded via CDP `Extensions.loadUnpacked`; throwaway daemon on port **4599** with `LOCAL_OPERATOR_CONFIG_DIR=/tmp/lop-mtab-config`; paired over the real WS flow; RPC driven directly against `/rpc`). All 14 scenarios passed; transcript below is from the live run:

1. **Two fresh opens → two DISTINCT surfaces** (the hijack fix):
   `A → bridge:323108282:50fa79…  B → bridge:323108283:7c4de2…`
2. **Independent reads**: A returns `ALPHA-CONTENT-7431`, B returns `BETA-CONTENT-9182`, no cross-bleed.
3. **Independent goto**: navigating A leaves B's document untouched.
4. **`tabs` lists both** with live URL/title, `createdAt ≤ lastUsedAt`, `limit: 8`.
5. **Per-surface refs**: session storage shows `refs` keyed by both tokens with different epochs.
6. **Nonce anti-guessing**: correct tabId + wrong nonce → `tab_closed`.
7. **Ambiguous bare close refused**, naming both live handles.
8. **`close {tab: A}`** closes only A; `tabs` then lists just B; A's handle → `tab_closed`.
9. **Resume**: `open {tab: B, url}` navigates B — same token back, no new tab.
10. **Back-compat bare close** with exactly one tab → closes it; `tabs` → `[]`.
11. **Cap**: 8 opens succeed, the 9th → `tab_limit` "already driving 8 tabs — close one…".
12. **Prune**: killing a tab externally (user-close simulation) → `tabs` drops it.

Full JSON transcript captured at `/tmp/lop-mtab-e2e-evidence.md` during the run. The real `~/.local-operator` state, default-port daemon, and the user's own Chrome were untouched throughout.

## Deliberately not done
- No cmux-side `tabs` emulation (cmux has no multi-surface registry; it degrades with the established typed error).
- `commands/snapshot.ts` touched only for per-surface refs wiring (2 lines), to keep the in-flight `fix/snapshot-hidden-headful` rebase clean.
- Session-side `BrowserSurface` still pins ONE handle per session by design — multi-tab is across sessions, not within one.

## Review
Agent review round to follow on this PR before merge.

### DIFF
--- a/extension/src/cdp.ts
+++ b/extension/src/cdp.ts
@@ -1,4 +1,4 @@
-import { getSession, parseSurface, type StoredSurface } from "./state";
+import { getSurfaces, putSurface, removeSurface, resolveSurfaceToken, type StoredSurface } from "./state";
 
 const attached = new Set<number>();
 
@@ -13,16 +13,28 @@ export class BridgeCommandError extends Error {
 }
 
 export async function requireSurface(token: unknown): Promise<StoredSurface> {
-  const parsed = parseSurface(token);
-  const { surface } = await getSession();
-  if (!parsed || !surface || parsed.tabId !== surface.tabId || parsed.nonce !== surface.nonce) {
+  // Exact-token lookup in the surfaces map: the nonce is part of the key, so a
+  // handle from another session (or a guessed tab id) resolves to nothing and
+  // keeps the same tab_closed shape callers already handle. This is what lets
+  // parallel sessions share the map without being able to drive each other's
+  // tabs by accident.
+  const surface = resolveSurfaceToken(token, await getSurfaces());
+  if (!surface) {
     throw new BridgeCommandError("tab_closed", "the browser tab handle is stale");
   }
   try {
     await chrome.tabs.get(surface.tabId);
   } catch {
+    // The Chrome tab is gone; drop the map entry so `tabs` never lists it and
+    // the map cannot fill with dead surfaces while their handles count toward
+    // the surface cap.
+    await removeSurface(String(token));
     throw new BridgeCommandError("tab_closed", "the browser tab was closed");
   }
+  // Recency for the `tabs` listing; best-effort, never on the command's
+  // critical path for correctness.
+  surface.lastUsedAt = Date.now();
+  await putSurface(surface);
   return surface;
 }
 
--- a/extension/src/cdp.ts
+++ b/extension/src/cdp.ts
@@ -1,4 +1,5 @@
-import { getSurfaces, putSurface, removeSurface, resolveSurfaceToken, type StoredSurface } from "./state";
+import { dropLogCapture } from "./log-capture";
+import { getSurfaces, removeSurface, resolveSurfaceToken, touchSurface, type StoredSurface } from "./state";
 
 const attached = new Set<number>();
 
@@ -25,17 +26,31 @@ export async function requireSurface(token: unknown): Promise<StoredSurface> {
   try {
     await chrome.tabs.get(surface.tabId);
   } catch {
-    // The Chrome tab is gone; drop the map entry so `tabs` never lists it and
-    // the map cannot fill with dead surfaces while their handles count toward
-    // the surface cap.
-    await removeSurface(String(token));
+    // The Chrome tab is gone: full prune, identical to the `tabs` prune site
+    // (review finding m1) — dropping only the map entry leaked the log ring
+    // buffer and left the dead tabId in the `attached` set for the worker's
+    // lifetime.
+    await pruneSurface(String(token), surface.tabId);
     throw new BridgeCommandError("tab_closed", "the browser tab was closed");
   }
   // Recency for the `tabs` listing; best-effort, never on the command's
-  // critical path for correctness.
+  // critical path for correctness. Conditional on the entry still existing so
+  // it cannot resurrect a concurrently-pruned surface (finding m5).
   surface.lastUsedAt = Date.now();
-  await putSurface(surface);
+  await touchSurface(String(token), surface.lastUsedAt);
   return surface;
+}
+
+/**
+ * The ONE dead-surface cleanup: map entry, log ring buffer, debugger session.
+ * Every prune site (requireSurface, status, tabs) must go through this — the
+ * three previously diverged and two of them leaked buffers/attachments
+ * (finding m1).
+ */
+export async function pruneSurface(token: string, tabId: number): Promise<void> {
+  dropLogCapture(tabId);
+  await detach(tabId);
+  await removeSurface(token);
 }
 
 export async function attach(tabId: number): Promise<void> {
--- a/extension/src/commands/input.ts
+++ b/extension/src/commands/input.ts
@@ -1,7 +1,7 @@
 import { BridgeCommandError, cdp, requireSurface } from "../cdp";
 import { withOriginGate } from "../origins";
 import { settle } from "../settle";
-import { getSession } from "../state";
+import { getRefs, surfaceToken, type StoredSurface } from "../state";
 
 interface QueryResult { nodeId?: number }
 interface PushResult { nodeIds: number[] }
@@ -29,12 +29,15 @@ async function callOnNode<T>(
   return out.result?.value as T;
 }
 
-async function nodeIdFor(tabId: number, selector: unknown, epoch: number): Promise<number> {
+async function nodeIdFor(surface: StoredSurface, selector: unknown): Promise<number> {
+  const tabId = surface.tabId;
   if (typeof selector !== "string" || !selector) throw new BridgeCommandError("element_not_found", "selector is required");
-  const { refs = {} } = await getSession();
+  // Refs are read from THIS surface's own map: a snapshot taken on another
+  // concurrently-driven tab must never supply the click target here.
+  const refs = await getRefs(surfaceToken(surface));
   const ref = /^e\d+$/.test(selector) ? refs[selector] : undefined;
   if (ref) {
-    if (ref.epoch !== epoch) throw new BridgeCommandError("element_not_found", "the page navigated since that snapshot");
+    if (ref.epoch !== surface.epoch) throw new BridgeCommandError("element_not_found", "the page navigated since that snapshot");
     // DOM.pushNodesByBackendIdsToFrontend requires the DOM agent to hold the
     // document for this session first, or Chrome rejects with -32000
     // "Document needs to be requested first". The CSS-selector branch below
@@ -66,7 +69,7 @@ export async function click(
 ): Promise<Record<string, unknown>> {
   const surface = await requireSurface(params.tab);
   const before = (await chrome.tabs.get(surface.tabId)).url ?? "";
-  const nodeId = await nodeIdFor(surface.tabId, params.selector ?? params.ref, surface.epoch);
+  const nodeId = await nodeIdFor(surface, params.selector ?? params.ref);
   await cdp(surface.tabId, "DOM.scrollIntoViewIfNeeded", { nodeId });
   const objectId = await objectIdFor(surface.tabId, nodeId);
   // Race a short grace window for a navigation the click may start. A click is
@@ -125,7 +128,7 @@ export async function click(
 export async function typeText(params: Record<string, unknown>): Promise<Record<string, unknown>> {
   const surface = await requireSurface(params.tab);
   const selector = params.selector ?? params.ref;
-  const nodeId = await nodeIdFor(surface.tabId, selector, surface.epoch);
+  const nodeId = await nodeIdFor(surface, selector);
   await cdp(surface.tabId, "DOM.scrollIntoViewIfNeeded", { nodeId });
   const objectId = await objectIdFor(surface.tabId, nodeId);
   // Replace-not-append (the cmux fill-vs-type lesson): focus, set the value on
--- a/extension/src/commands/nav.ts
+++ b/extension/src/commands/nav.ts
@@ -2,7 +2,15 @@ import { attach, BridgeCommandError, cdp, detach, requireSurface } from "../cdp"
 import { dropLogCapture, startLogCapture } from "../log-capture";
 import { askOrigin, safeHttpUrl, withOriginGate } from "../origins";
 import { settle } from "../settle";
-import { getSession, setSurface, surfaceToken, type StoredSurface } from "../state";
+import {
+  atSurfaceCap,
+  getSurfaces,
+  MAX_SURFACES,
+  putSurface,
+  removeSurface,
+  surfaceToken,
+  type StoredSurface,
+} from "../state";
 
 async function page(tabId: number): Promise<{ url: string; title: string }> {
   const tab = await chrome.tabs.get(tabId);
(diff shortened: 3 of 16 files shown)


damianvtran:
### Agent review — round 1
Scope: origin/main..f435ca00c2b8e4566418bfe05468c63022709ba5

Re-ran locally: extension `pnpm typecheck` ✅, `pnpm test` 16/16 ✅, `pnpm build` ✅; `python -m local_operator.browser_bridge.gen_ts --check` ✅ (regeneration exact); `pytest tests/unit/browser_bridge` 58 passed ✅. E2E transcript at /tmp/lop-mtab-e2e-evidence.md inspected and consistent with the claims.

#### MAJOR

**M1 — `tabs` hands out full driveable tokens while three comments claim "listing does not grant control".**
`nav.ts:142` returns `tab: token` — the complete `bridge:<tabId>:<nonce>` capability — for EVERY session's surface, and the same full tokens leak through the `tab_limit` error data (`nav.ts:69`, `tabs: Object.keys(surfaces)`) and the bare-close refusal message (`nav.ts:186`). Yet `nav.ts:129-133`, `protocol.py:36-39`, and the PR body all state that driving a listed tab "still requires its exact nonce-bearing handle... listing does not grant control." The handle IS the control: once listed, the nonce anti-guessing property (`state.ts:40-44`, `cdp.ts:16-21`, the new pure test's whole point) protects nothing against any caller that can invoke `tabs`. The no-handle `status` path (`nav.ts:112-115`) also returns another session's full token.

Judged against the trust domain: every session reaches the daemon with the same shared `session_key` from the state file, so this is not a privilege escalation across a real boundary — any local process already has full authority — and the Python tool gives an agent no way to adopt a foreign token (`goto`/`click`/`close` all use the pinned `state.surface_id`; only prose in the tool text says "do not drive or close them"). That makes this acceptable *behaviour* for a same-user local bridge, but the PR does not document it honestly: it documents the opposite. Fix is one of (a) redact the nonce in the listing/error payloads (e.g. `bridge:<tabId>:<nonce[:6]>…`; the Python side can still mark `(yours)` by prefix-matching its pinned handle, and close-by-handle then genuinely requires ownership), which makes the comments true; or (b) keep the exposure and rewrite the comments in nav.ts/protocol.py + the PR body to say plainly that within the paired-daemon trust domain the listing shares capabilities and cross-session isolation is anti-*accident*, not anti-*adversary*. I recommend (a): it is small and preserves the invariant the PR advertises. There is also a latent policy contradiction worth a sentence in the tool description: `tab_limit` says "close one", but a session owning zero tabs when the cap is filled by others has no tab it is permitted to close.

#### MINOR

**m1 — Prune paths skip the cleanup `tabs` does.** `cdp.ts:31` (`requireSurface` dead-tab prune) and `nav.ts:117` (`status` prune) call only `removeSurface`; `tabs` (`nav.ts:151-153`) also does `dropLogCapture` + `detach`. The skipped calls leak the log ring buffer entry and leave the dead tabId in cdp's `attached` set for the worker's lifetime. One-line fix: use the same cleanup in all three prune sites.

**m2 — Cap check counts dead surfaces.** `open`'s `atSurfaceCap` (`nav.ts:60-61`) reads the raw map without pruning, so 8 stale entries (browser restarted, user closed tabs) refuse a fresh open until something else prunes. Recoverable via the documented `tabs` remedy, but pruning (or at least liveness-filtering) before the cap check would remove a needless refusal.

**m3 — `open`-resume doesn't bump `epoch`; `goto` does.** `nav.ts:97-99` increments epoch on `goto`, but the resume path (`nav.ts:50-58`) navigates without bumping, so pre-resume snapshot refs pass the `ref.epoch !== surface.epoch` gate (`input.ts:40`) against a new document. In practice `DOM.pushNodesByBackendIdsToFrontend` on a dead backendNodeId yields the "snapshot ref is stale" error rather than a wrong click, and the old reuse path had the same shape — but resume is now a designated verb, and the epoch mechanism exists precisely for this. Bump epoch on resume.

**m4 — Dead-pin recovery keyed on a substring of a human-facing message.** `builtin.py:5615` (`"is gone" in problem.text`) and `:5784` couple recovery to the exact wording of `backend.py:75`. Rewording that diagnostic silently breaks fresh-tab recovery. Prefer surfacing the typed `ErrorCode.TAB_CLOSED` through `_bridge_call`'s error path instead of string matching.

**m5 — Zombie-resurrection race on `lastUsedAt`.** `requireSurface`'s best-effort `putSurface` (`cdp.ts:35-38`) can re-insert an entry that a concurrent `tabs`/`status` prune (different daemon lock key) just removed, leaving a dead surface counting toward the cap until the next prune. Self-healing and narrow, but cheap to close: make the lastUsedAt write conditional on the token still being present (inside `withStore`).

#### NIT

**n1 —** Bare-close multi-tab refusal uses code `"internal"` (`nav.ts:185`), which `format_error` renders as "browser bridge error (internal): …" — reads as a bridge bug to the agent. A typed code would map to a cleaner remedy message.

**n2 —** `daemon.py` `_tab_locks` now grows one entry per surface token forever (per-tab keys were near-singleton before; now every opened tab's token persists). Harmless at cap 8 per run, but an eviction on close would keep it tidy.

**n3 —** `_format_bridge_tab` imports `datetime` inside the function (`builtin.py` ~5710); module-level import is the codebase norm for stdlib.

#### Verdict

**Not clean — one MAJOR (M1).** No BLOCKER: the mechanics are correct, the regenerated protocol is exact, the hijack fix and back-compat close behave as described and are pinned by tests plus a real E2E transcript. M1 needs either the nonce redaction or an honest rewrite of the security claims before merge; the minors are follow-up-sized but m1–m3 are one-liners worth folding into the same remediation commit.

damianvtran:

All findings worked at commit `6cd1a849d6073a31442d9738d05a7b88778de85a` (one commit; per-finding detail below). Reviewed head was `f435ca00`.

| Finding | Status | Detail |
|---|---|---|
| **M1** | **Fixed — 6cd1a849** | Took the recommended remedy (a): every non-owner surface now emits **redacted** handles `bridge:<tabId>:<nonce[:6]>…` — the `tabs` listing, `tab_limit` error data, the bare-close refusal, and no-handle `status` (`state.ts redactToken`/`ownsRedacted`, `nav.ts` all four sites). A redacted handle resolves nothing (`resolveSurfaceToken` is an exact-token lookup; pinned by pure test + live E2E: driving a listed handle → `tab_closed`). Owners recognise their tab by prefix-matching the full token from their own `open` (`builtin._owns_redacted_tab` mirrors the TS helper). The Python listing text now says "handles are redacted — the listing is awareness-only", making the documented claim true. The tab_limit policy contradiction got a sentence in `backend.py`: "if none is yours, another session (or the user) must close one." |
| **m1** | **Fixed — 6cd1a849** | New `cdp.ts pruneSurface()` = `dropLogCapture` + `detach` + `removeSurface`; used by all prune sites (`requireSurface`, `status`/`tabs` via `liveSurfaces()`). |
| **m2** | **Fixed — 6cd1a849** | `open`'s cap check runs on `liveSurfaces()` (prunes dead entries first). E2E: filled to 8 → `tab_limit`; killed one tab externally → next open **succeeds**. |
| **m3** | **Fixed — 6cd1a849** | Resume path bumps `epoch` and persists, same as `goto`. E2E: snapshot → resume → click on the old ref → `element_not_found: the page navigated since that snapshot`. |
| **m4** | **Fixed — 6cd1a849** | `_bridge_call` now carries the typed wire code in `ToolResult.details["error_code"]`; dead-pin recovery and handle-drop branch on `== "tab_closed"`, no string matching. Tests updated to pin the typed path. |
| **m5** | **Fixed — 6cd1a849** | `touchSurface()` refreshes `lastUsedAt` only if the token is still present, inside the serialized store queue — a concurrent prune can no longer be resurrected. |
| **n1** | **Fixed — 6cd1a849** | New typed `ErrorCode.TAB_AMBIGUOUS` (protocol + regenerated TS, exact one-line diff); `format_error` relays the message instead of "browser bridge error (internal)". E2E shows `"code": "tab_ambiguous"` with redacted handles. |
| **n2** | **Fixed — 6cd1a849** | Daemon evicts the per-tab lock entry after a `close` completes, when the lock is unheld. |
| **n3** | **Fixed — 6cd1a849** | Uses the module-level `datetime`/`UTC` import already present in `builtin.py`. |

**Validation re-run on the new head**: `pnpm typecheck` ✅, `pnpm test` **17/17** ✅ (new pure test: redaction prefix-match + non-resolvability), `pnpm build` ✅; `gen_ts --check` ✅ (regen diff is exactly `TAB_AMBIGUOUS`); pytest browser/bridge slices **163 passed** ✅; flake8/black/isort ✅; pyright full tree **0 errors** ✅.

**E2E re-run** (same throwaway stack: off-screen background Chrome 151 on a /tmp profile, rebuilt extension via `Extensions.loadUnpacked`, daemon on 4599, `LOCAL_OPERATOR_CONFIG_DIR=/tmp/lop-mtab-config`), live transcript appended to `/tmp/lop-mtab-e2e-evidence.md`:
- `tabs` returns `bridge:<tabId>:<6 chars>…` for every entry; both full tokens prefix-match their entries.
- Driving a redacted handle → `tab_closed` (not a capability).
- Bare close with two tabs → `tab_ambiguous`, message contains only redacted handles.
- No-handle `status` → redacted token.
- Resume epoch bump: pre-resume ref refused with the epoch-stale message.
- Cap: 9th open → `tab_limit` (redacted handles in data); external tab kill → next open succeeds (live-only count).
- Owner path intact throughout: full tokens drove snapshot/resume/close for all 8 tabs; final `tabs` = [].

No findings rejected or deferred. Head moved `f435ca00` → `6cd1a849`; ready for round 2.

damianvtran:
### Agent review — round 2
Scope: f435ca00c2b8e4566418bfe05468c63022709ba5..6cd1a849 (delta), verified at head 6cd1a849

Delta is one commit, 10 files, +252/−60. Re-ran locally at head: extension `pnpm typecheck` ✅, `pnpm test` 17/17 ✅ (new redaction test included), `pnpm build` ✅; `gen_ts --check` ✅ (protocol.gen.ts exact); bridge pytest slice 59/59 ✅ (test_tool_selection 28/28, incl. the new `test_redacted_ownership_prefix_matching`). Updated E2E transcript at /tmp/lop-mtab-e2e-evidence.md (regenerated after the fix; contains redacted-listing, redacted-handle drive refusal, typed tab_ambiguous, redacted status, redacted 9th-open-at-cap, and epoch-stale-after-resume scenarios) inspected and consistent with the code at this head.

#### Round 1 findings — verification

**M1 (MAJOR, nonce leakage) — VERIFIED FIXED.** All four leak surfaces now emit `bridge:<tabId>:<nonce[:6]>…` via `redactToken`: the `tabs` listing (`nav.ts:176`), the `tab_limit` error data (`nav.ts:98`), the bare-close refusal — now typed `tab_ambiguous` — (`nav.ts:208-215`), and no-handle `status` (`nav.ts:150`). Grepped every `surfaceToken(...)` egress in `extension/src`: the only full-token returns are `open` (fresh and resume, `nav.ts:85,124`) and commands gated behind `requireSurface(params.tab)` — i.e. the caller already proved possession of the full token. A redacted handle resolves nothing: the ellipsis fails `parseSurface`'s regex, the new pure test pins `resolveSurfaceToken(redacted, …) === undefined`, and the E2E transcript shows a drive attempt with a listed redacted handle refused. Own-flows intact: `open` still returns the full token, resume-via-open works (E2E), and the `(yours)` marker is prefix-match only (`state.ts` `ownsRedacted` / `builtin.py` `_owns_redacted_tab`), so own-entry marking re-leaks nothing. `builtin.py` never needed a full token back from a listing — it pins the token from `open` and all drives use `state.surface_id`; tool text and the `tab_limit` remedy wording were updated to match the awareness-only reality.

**m1 (prune-site parity) — VERIFIED FIXED.** Single `pruneSurface` (dropLogCapture + detach + removeSurface) in `cdp.ts`, used by `requireSurface`, `status`, `tabs`, and `liveSurfaces`.
**m2 (cap counts dead surfaces) — VERIFIED FIXED.** `open`'s cap check now runs against `liveSurfaces()`.
**m3 (open-resume epoch) — VERIFIED FIXED.** Resume path bumps `surface.epoch` before `putSurface` (`nav.ts:83-84`); E2E shows a pre-resume snapshot ref refused as epoch-stale after resume.
**m4 (string-matched dead-pin recovery) — VERIFIED FIXED.** `_bridge_call` now stamps `problem.details = {"error_code": exc.code.value}`; both recovery sites (`_bridge_open`, `_bridge_action`) branch on `error_code == "tab_closed"`.
**m5 (lastUsedAt resurrect race) — VERIFIED FIXED.** New `touchSurface` does the presence check + write inside `withStore`, so it cannot interleave with a prune's read-modify-write.
**n1 — FIXED.** Bare-close refusal typed `tab_ambiguous` end-to-end (extension, `ErrorCode` in protocol.py, regenerated protocol.gen.ts exact, clean `format_error` mapping in backend.py).
**n2 — FIXED.** Daemon evicts the per-tab lock entry on close when unlocked (`daemon.py:531-533`).
**n3 — FIXED.** `_format_bridge_tab` uses the module-level `UTC`/`datetime` import.

#### New findings from the delta

**NIT — `nav.ts` imports `detach` and `dropLogCapture` that it no longer calls** (lines 1–2; both usages moved into `pruneSurface`/`closeSurface`). The esbuild config sets no `pure` annotations, so these survive bundling as live no-op calls in the shipped worker. Harmless at runtime — `attach`/`detach`/`dropLogCapture` are all guarded no-ops — but dead weight in a store artifact, and it obscures that nav.ts cleanup is fully consolidated. One-line fix: drop the two unused imports.

Nothing else: I specifically probed the delta for regressions in the redaction path (daemon lock eviction on close — sound in the single-threaded event loop; `touchSurface` under contention — serialized by `withStore`; redaction breaking the tool's own pinning — it does not, pinning never round-trips a listing) and found no correctness issues.

#### Verdict

**Clean.** No BLOCKER, no MAJOR. Round 1's M1, all 5 minors, and all 3 nits are remediated and verified by re-execution, not just inspection. The round is TERMINAL: the single remaining nit (unused imports in nav.ts) is a follow-up, not a reason for another round.

damianvtran:
### Rebase note (post-round-2)

Rebased onto `origin/main` (`5dbb2b84`, which merged PR #320 after round 2 was posted). Two conflicts, both resolved preserving both sides:

- `extension/src/commands/input.ts` — kept the round-2 multi-tab `surface.epoch` form AND main's `DOM.getDocument`-before-push fix (#324).
- `local_operator/browser_bridge/daemon.py` + `protocol.py` — the branch's inline `COMMAND_TIMEOUTS` block dropped (the table moved to `protocol.py` in #320); the `tabs: 20.0` timeout entry was re-added to `protocol.py`'s table with its comment.

Re-validated at `0a92da9f2155eee2899213ecab8935fb27b364d0`: `pnpm typecheck` clean, extension tests 19/19, `pnpm build` clean, bridge pytest 68/68, `gen_ts --check` clean. The review delta from round 2 (`f435ca00..6cd1a849`) is content-identical modulo these conflict resolutions.

damianvtran:
### Agent review — round 3 (rebase verification)

Reviewer: Local Operator reviewer subagent (anthropic/claude-fable-5) — independent of the implementing agent
Scope: e1ccc3e6..0a92da9f (rebased branch; delta-verified against round-2-clean 6cd1a849)

**Delta verification.** Compared the pre-rebase branch patch (`f6e7d027..6cd1a849`) against the post-rebase patch (`e1ccc3e6..0a92da9f`): identical file sets, and the substantive content differs by exactly the two stated conflict resolutions — nothing else:

1. `extension/src/commands/input.ts` — keeps BOTH the `ref.epoch !== surface.epoch` staleness check (line 40) AND main's `DOM.getDocument` call before `DOM.pushNodesByBackendIdsToFrontend` (lines 43–48), with the -32000 rationale comment. `scroll.ts` picks up the same constraint via comment reference, as on main.
2. `daemon.py` / `protocol.py` — the inline `COMMAND_TIMEOUTS` block is gone from `daemon.py` (it now imports the table, line 33) and the `"tabs": 20.0` entry lives in `protocol.py:83` with its lock-key comment. `grep '"tabs": 20'` matches only `protocol.py`.

One bookkeeping note, no code impact: the rebase note says the branch was rebased onto `5dbb2b84`, but the actual base is `e1ccc3e6` (origin/main one commit later, after #327 merged). The two extra inter-tip deltas that flags (`local_operator.tcss` background fill, `test_role_subagents.py` jobs-tool assertions) are #327's own changes arriving via main, not branch edits.

**Checks re-run in the worktree:**
- `pnpm typecheck` — clean
- `pnpm test` — 19/19 pass
- `pnpm build` — dist emitted
- `pytest tests/unit/browser_bridge -q` — 68 passed

**Verdict: clean — TERMINAL.** The rebase is faithful to the round-2-approved content; no new findings.

PR #309, opened 2026-08-26
fix(tui): restore complete subagent transcripts

damianvtran:
## Summary

Every full-page subagent transcript had three related replay defects: the delegated instruction could appear after the child's completed work, durable `edit` and `write` rows lost their red/green diffs and exposed raw arguments instead, and completed tool rows lost their elapsed durations.

- Durable child history now projects tool-result `details` and finite non-negative `duration_s` into the same `ToolCard` used by the parent transcript.
- New child launches persist the exact durable launch-message ID through the job, comms node, snapshot, resume, and viewer paths. The viewer replaces only that keyed durable row with the concise delegated instruction. Legacy transcripts keep a synthetic instruction at the head and never rewrite value-matched history.
- Older pages insert beneath the fixed synthetic instruction while retaining the actual visible block and its intra-block viewport offset.
- Cancelled children finish their bounded transcript persistence before publishing a resumable terminal outcome.
- After `main` advanced to `0.37.0`, this branch resolves on that release and bumps `0.41.0` → `0.41.1` as a patch fix.

## Impact

**Before:** opening a completed child could show its delegation at the transcript end. `edit` and `write` rows showed a checkmark but no duration or diff counters, and expanding them exposed raw field parameters. Loading earlier history could move the reader's visible content.

**After:** the delegation remains chronological, completed write tools use the parent transcript's `+N/-N`, duration, and unified-diff presentation, and paged history preserves the same visible block at the same viewport-relative position.

## Current-main reconciliation

Rebased onto `origin/main` `7b42eccd` (`0.41.0`), which landed #318 `feat(hub): fan resume out to multiple subagents`. Version resolved to `0.41.1` (patch — a dead-field cleanup and merge, no step-function capability per the new AGENTS.md materiality policy). Only `pyproject.toml` conflicted (version); `tests/unit/harness/test_comms.py` auto-merged keeping both #318's three fan-resume tests and this PR's launch-identity assertions on the shared resume test, with no dropped assertions from either side.

[...]

PR #322, opened 2026-08-26
fix(tui): pin subagent prompt to top, stop view tearing, let children poll background bash

damianvtran:
## Summary

Three fixes to the subagent transcript viewer and to subagent tool provisioning, all surfaced from real use.

1. **The delegated prompt rendered at the bottom of a subagent's transcript instead of the top.** The prompt is the delegation instruction — chronologically the first message — but it kept landing under all of the child's durable history, and jumped to the bottom the moment the async disk history loaded.
2. **The subagent view tore on mouse-move** — the current frame's interactive elements painted over a frozen previous frame, worst while the working spinner ticked and the pointer moved over the hint row.
3. **A subagent that ran `bash(background=true)` looped forever emitting `Tool not found: jobs`** — backgrounding succeeded and the receipt told the model to poll with `jobs(op='peek')`, but `jobs` had been pruned from that child's toolset.

### 1. Prompt pinned to the head of the body (`local_operator/tui/widgets/subagent_view.py`)

The page body was assembled independently in two places (`show` and `_reconcile_current_body`) as `[*history, *surviving_live]`. The prompt is a live entry keyed `"__prompt__"`; every other live key matches a durable-history key and folds out, so `"__prompt__"` was the sole live survivor and landed **after** all durable rows — at the bottom. On first paint (before disk history loaded) it showed high up; once `_apply_history_page` reconciled and `scroll_end`'d, it was pinned to the bottom. That is the two-state behaviour the bug reporter saw.

Both sites now route through one `_merge_body()` helper that separates `"__prompt__"` out of the live list and prepends it ahead of durable history, so the prompt leads the page whether or not history is present and whether the job is running or settled. It stays a single keyed entry, mounted exactly once by the existing keyed diff.

### 2. Opaque ground on the page (`local_operator/tui/local_operator.tcss`)

`.subagent-view` had no `background` and resolved to transparent at runtime, unlike every other block widget. A transparent scrolling region lets Textual's compositor skip clearing cells on a partial repaint, so a hover on an `.actionable` hint plus the 8 Hz spinner painted new cells over stale ones — the tearing. Added `background: $lo-bg`, matching the Screen's own baseline so subagent mode still reads as "the same app looking elsewhere" (per `_open_subagent_view`) rather than an elevated slab.

### 3. A child can poll and cancel its own background bash jobs (`local_operator/harness/subagent.py`)

`bash(background=true)` works for a child (its `context.jobs` is the child's own live `Session.jobs`), and the success receipt says to follow up with `jobs(op='peek')`. But the depth-aware prune dropped the whole capability set — `jobs` included — for non-delegating roles (coder/reviewer), scouts, and grandchildren, so that advice pointed at a tool that no longer existed and the model looped on `Tool not found: jobs`. The child's work still completed (the result auto-delivers through the owner sink), but the poll loop burned turns and budget.

`jobs` only **observes/cancels** the child's own jobs — it spawns nothing (that's `task`) and dies with the child's job manager — so it crosses no boundary the prune protects. It's now spared from the prune under a **derived invariant**: a child keeps `jobs` iff it can still produce a background job (`_can_background(tools)` reads the actual `bash` schema for the `background` parameter rather than hard-coding roles, so the rule can't drift from reality). `task`/`wait`/`wake` keep their existing treatment — a child that must not fan out still can't, and a scout (no `bash`) still gets no `jobs`. The rejected alternative (stripping `bash`'s `background` per child) was more invasive and would remove a genuinely useful capability.

### Version

`pyproject.toml`: **0.41.0 → 0.41.1** (patch — three user-facing bug fixes, no new step-function capability).

### Bug 1 — visual proof (before/after)

Rendered the real `OperatorApp` with a subagent view opened over a job carrying both a delegated prompt and 12 durable disk rows, captured with `app.save_screenshot`, then rasterised and viewed:

- **Before:** `DELEGATED PROMPT` sits at the bottom, below `durable history row 0..11`, even scrolled home (`prompt index: 12 of 17`).
- **After:** `DELEGATED PROMPT` leads, then `durable history row 0..11`, then trajectory (`prompt index: 0 of 17`).

Regression guard: `tests/unit/tui/test_subagent_view.py::test_prompt_leads_the_body_ahead_of_durable_history` — seeds a prompt + 8 durable rows, waits for the async history worker, asserts `_entries[0].key == "__prompt__"`, prompt index < first durable index, the first mounted block is the instruction `UserBlock`, and exactly one such block (deduped). Fails on old code, passes on the fix.

### Bug 2 — root cause + fix confirmed; motion artifact not directly captured

The tearing is motion-driven (hover + 8 Hz spinner over a transparent region) and does not appear in a static still, so the artifact itself is not screenshot-captured. Confirmed instead: (1) `.subagent-view` resolved to a **transparent** background at runtime, unlike every other block widget; (2) after the change it resolves opaque to `$lo-bg`; (3) no appearance regression — the after-still is identical apart from the bug-1 reordering. An opaque container is the standard compositor-clears-the-region remedy. Full mouse-move confirmation needs a live interactive session; flagged honestly rather than claimed.

### Bug 3 — real construction + tests

Built real `coder` and `scout` children and asserted the resulting toolsets directly: the coder has `jobs` + `bash` and no `task`/`wait`/`wake`; the scout (no `bash`) has neither `jobs` nor the spawn tools. Tests:

- `tests/unit/session/test_launch_subagent.py` — grandchild test updated (`jobs` now kept, `task`/`wait`/`wake` gone); restricted-role test tightened; added `test_non_delegating_role_keeps_jobs_to_poll_its_background_bash`.
- `tests/unit/session/test_role_subagents.py::test_a_non_delegating_role_gets_no_task_tool` — updated: a reviewer keeps `jobs` + `bash` but still gets none of `task`/`wait`/`wake`.

### Gates (whole tree, as CI runs them)

```
$ .venv/bin/python -m flake8 .                          # exit 0
$ black --check .   (black 26.1.0)                        # 551 files unchanged
$ isort --check-only .   (isort 5.13.2, repo config)      # exit 0
$ pyright  (touched .py files)                            # 0 errors, 0 warnings
$ pytest tests/unit/tui tests/unit/session tests/unit/harness -q   # all pass
```

Notes: full-tree `pyright` cannot complete on the dev host (it times out under contention; CI runs it in its own runner), so pyright was scoped to the touched files. `uvx isort==5.13.2` cannot spawn in this sandbox (an environment issue, not a lint failure); validated with the identically pinned isort 5.13.2 module. One suite failure — `test_esc_with_no_children_says_nothing_about_subagents` — is a worktree-path artifact: the footer renders the CWD `/private/tmp/lop-subagent-fixes`, whose path contains the substring `subagent`, and the test asserts no row contains it. Proven by running the test from a renamed copy dir (passes); CI's checkout path is `local-operator` and won't trigger it.

## Full changelog

https://github.com/damianvtran/local-operator/compare/v0.41.0...dev-subagent-fixes

### DIFF
--- a/local_operator/harness/subagent.py
+++ b/local_operator/harness/subagent.py
@@ -65,6 +65,17 @@ the child-shaped tool — one peer, its parent — which is how it answers a
 question the parent asked and how it reports being blocked without waiting
 for its final result. See :mod:`local_operator.harness.comms`.
 
+``jobs`` is a SECOND, CONDITIONAL exception, on a different principle. It
+observes and cancels the child's OWN background jobs — it spawns nothing
+(that is ``task``) and dies with the child's job manager — so it crosses no
+boundary the prune protects. But a child can only produce a background job
+while its ``bash`` retains ``background``, and the bash receipt tells the
+model to poll such a job with ``jobs(op='peek')``. So the invariant is: a
+child keeps ``jobs`` IFF it can still background a bash command. The prune
+below re-adds ``jobs`` exactly under that condition (:func:`_can_background`),
+which is what stops a non-delegating role or grandchild that backgrounds a
+long command from looping forever on ``Tool not found: jobs``.
+
 Approvals the child asks for carry ``ToolContext.job_id`` — the id of the job
 this child IS — so a host can scope an approval decision to the delegated
 work that provoked it. Live failure it exists for: a subagent outliving its
@@ -145,6 +156,24 @@ TRAJECTORY_CAP = 500
 #: being present. Role guidance generally lives in
 #: :mod:`local_operator.agent_profiles`.
 SCOUT_TOOL_ALLOWLIST = frozenset({"read", "glob", "grep", "list_variables", "read_variable"})
+
+
+def _can_background(tools: "list[AgentTool]") -> bool:
+    """Whether this toolset can PRODUCE a background job.
+
+    Only ``bash`` spawns background jobs, and only while its schema still
+    carries the ``background`` parameter. A toolset with no such ``bash`` (a
+    scout, an allowlist that omits it) cannot register a job, so it has
+    nothing for ``jobs`` to observe. Deriving the answer from the actual
+    schema — rather than hard-coding which roles background — is what keeps
+    the ``jobs``-retention rule below tied to reality if the bash schema or a
+    role's allowlist later changes.
+    """
+    bash = next((tool for tool in tools if tool.name == "bash"), None)
+    if bash is None:
+        return False
+    return "background" in bash.parameters.get("properties", {})
+
 
 #: Preamble stamped onto a scout prompt when no profile resolves. The tool
 #: filter enforces the letter; this states the intent, so the scout REPORTS
@@ -1153,6 +1182,26 @@ async def _build_child_session(
         drop = merged_in
     else:
         drop = {name for name in merged_in if name == "wake"}
+    # ``jobs`` is the OBSERVE/CONTROL surface over this child's OWN background
+    # jobs (peek at output, cancel) — it spawns nothing (that's ``task``) and
+    # dies with the child's job manager, so it crosses no boundary the prune
+    # protects. Meanwhile the ``bash`` tool's ``background=true`` receipt tells
+    # the model to "follow it with jobs(op='peek')". When the branch above
+    # dropped the whole ``merged_in`` set (non-delegating role, grandchild),
+    # that advice pointed at a tool that no longer existed, so a child that
+    # backgrounded a long command (a coder polling a 10-min pyright) spun
+    # forever emitting ``Tool not found: jobs``. Invariant, encoded here rather
+    # than as two edits that can silently drift: a child keeps ``jobs`` IFF it
+    # can still produce a background job (its ``bash`` retains ``background``).
+    # Un-pruning ``jobs`` (not stripping ``bash``'s ``background``) preserves a
+    # real capability — a child genuinely benefits from backgrounding a long
+    # build and polling it — while killing the loop; stripping the schema
+    # per-session would be more invasive and would remove that capability.
+    # ``task``/``wait``/``wake`` keep their treatment: ``jobs`` polling is
+    # non-blocking and is the advertised path, so sparing ``jobs`` alone is the
+    # minimal correct fix, and a child that must not fan out still cannot.
+    if _can_background(tools):
+        drop = drop - {"jobs"}
     child.refresh_tools([tool for tool in child._tools if tool.name not in drop])
     if mcp is not None:
         mcp.attach(child)
--- a/local_operator/tui/local_operator.tcss
+++ b/local_operator/tui/local_operator.tcss
@@ -1230,9 +1230,18 @@ Screen.subagent-compact #band {
  * four-row panel. It cannot come from the body's `TranscriptView` padding,
  * which the class below deliberately zeroes — the hint row is BELOW the body,
  * so the row has to be the page's. */
+/* An OPAQUE ground on the scrolling page. Without a fill the container is
+ * transparent, and a transparent scrolling region lets Textual's compositor
+ * skip clearing cells on a partial repaint: a hover on an `.actionable` hint
+ * plus the 8Hz working spinner then painted new cells over stale ones and the
+ * page appeared to tear on mouse-move. The `$lo-bg` token matches the Screen's
+ * own baseline ground on purpose — subagent mode is "the same app looking
+ * elsewhere" (see `_open_subagent_view`), so the fill must read as the app's own
+ * ground and not as an elevated slab the way the surface elevation would. */
 .subagent-view {
     height: 1fr;
     padding: 1 1 1 1;
+    background: $lo-bg;
 }
 
 .subagent-view-title {
--- a/local_operator/tui/widgets/subagent_view.py
+++ b/local_operator/tui/widgets/subagent_view.py
@@ -1362,13 +1362,34 @@ class SubagentView(Vertical):
                 if previous != current:
                     break
                 common_suffix += 1
+            # The pinned delegation prompt (see `_merge_body`) is a STABLE
+            # PREFIX: it sits at index 0 in both the old and new body and is not
+            # part of the paged history. Older rows page in AFTER it, so the
+            # freshly-prepended set is the slice BETWEEN that prefix and the
+            # common suffix. Counting from index 0 instead (the old accounting)
+            # folded the prompt into the "new rows" set and remounted it a
+            # second time, because the prompt is no longer a pure front to
+            # prepend against once history rows exist below it.
+            common_prefix = 0
+            for previous, current in zip(self._entries, entries):
+                if previous != current:
+                    break
+                common_prefix += 1
             count = len(entries) - common_suffix
-            if count > 0:
-                new_entries = entries[:count]
+            # Clamp the prefix inside the newly-prepended window: an entry can
+            # be both leading-stable and trailing-stable only when nothing
+            # changed, which `count > common_prefix` already excludes.
+            common_prefix = min(common_prefix, count)
+            if count > common_prefix and common_prefix < len(self._blocks):
+                new_entries = entries[common_prefix:count]
                 new_blocks = [entry_block(entry) for entry in new_entries]
-                self._body.prepend_blocks(new_blocks, anchor_offset=anchor)
-                self._blocks[0:0] = new_blocks
-                self._entries[0:0] = new_entries
+                # Mount ahead of the current first post-prefix block (the first
+                # durable-history row) so the prompt above it stays put and the
+                # reader's anchor, further down, is untouched.
+                before_block = self._blocks[common_prefix]
+                self._body.prepend_blocks(new_blocks, anchor_offset=anchor, before=before_block)
+                self._blocks[common_prefix:common_prefix] = new_blocks
+                self._entries[common_prefix:common_prefix] = new_entries
                 self._pending = entries
                 return
         self._pending = entries
--- a/local_operator/tui/widgets/subagent_view.py
+++ b/local_operator/tui/widgets/subagent_view.py
@@ -1166,9 +1166,7 @@ class SubagentView(Vertical):
                 self._known[entry.key] = entry
             elif _supersedes(entry, known):
                 self._known[entry.key] = entry
-        live = [self._known[key] for key in self._order]
-        durable_keys = {entry.key for entry in self._history_entries}
-        body = [*self._history_entries, *(entry for entry in live if entry.key not in durable_keys)]
+        body = self._merge_body()
 
         tail = self._tail_entry(gone, progress)
         if tail is not None:
@@ -1354,12 +1352,7 @@ class SubagentView(Vertical):
     def _reconcile_current_body(
         self, *, anchor: float | None = None, prepend: bool = False
     ) -> None:
-        live = [self._known[key] for key in self._order]
-        durable_keys = {entry.key for entry in self._history_entries}
-        entries = [
-            *self._history_entries,
-            *(entry for entry in live if entry.key not in durable_keys),
-        ]
+        entries = self._merge_body()
         tail = self._tail_entry(self._status == "gone", "")
         if tail is not None:
             entries.append(tail)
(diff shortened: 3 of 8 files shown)


damianvtran:
### Design review — round 1

Reviewer: designer (Local Operator team) — anthropic/claude-opus-4
Scope: dfb07b6efb511a4d48ccc7289af6022ed195d430 (origin/main..HEAD, `dev-subagent-fixes`)

Reviewed the two user-visible changes against rendered frames, not source alone: the supplied `before.png`/`after.png` (settled job, durable history + prompt) and a freshly captured **running** frame (`running.svg` in the shots dir) driving the real `OperatorApp` with the job `status="running"` so the 8Hz working spinner is live and the two tool cards (`read` ✓, `bash` 2 failed ✗) are mounted — the exact tearing scenario Bug 2 targets.

**Bug 1 — prompt pinned to head.** Correct and a clear improvement. In `before.png` the `DELEGATED PROMPT` UserBlock sits stranded at the very bottom, below `durable history row 11`, which reads as if the delegation instruction were the child's last utterance — backwards from the mental model. In `after.png` and the running frame it leads: instruction (the user turn of this conversation) → child's prose and tool work in chronological order. It renders as a UserBlock with the left accent bar (visible in the running frame), so it is distinguishable from the child's own assistant rows. Spacing is right: a single blank row separates the two-line prompt block from `durable history row 0`, and the block sits cleanly below the title rule with no crowding. No hierarchy concern.

**Bug 2 — opaque `$lo-bg` ground.** No unintended visual shift. Confirmed `Screen` is `background: $lo-bg` (`local_operator.tcss:74-75`) and `.subagent-view` now uses the same token (`:1244`), so the fill is the app's own baseline ground, not an elevated slab — subagent mode still reads as "the same app looking elsewhere." Before/after grounds are indistinguishable apart from the reordering. In the running frame the tool cards render as their own tinted slabs (neutral for `read`, the red-tinted error slab for the failed `bash`) sitting cleanly over the opaque ground with no seam or wrong elevation, and the title/rule/hint chrome, the `Subagents` dock, and the scrollbar are all unregressed.

No BLOCKER, no MAJOR, no MINOR, no NIT. The visuals are clean.

**Verdict: TERMINAL.** Both user-visible changes are correct and free of regressions on the current head. No follow-ups.

damianvtran:
### Agent review — round 1

Scope: origin/main..dfb07b6efb511a4d48ccc7289af6022ed195d430

Independent code review of the three fixes. I read `git diff origin/main...HEAD`, exercised the live `.venv`, and reproduced the findings below rather than inferring them. Bug 2 (opaque ground) and bug 3 (`_can_background` + `jobs` sparing) are sound; bug 1's head-pin introduces one real regression in the multi-page paging path.

---

**F1 — MAJOR — Prompt double-mounts when paging older history (>1 page).** `local_operator/tui/widgets/subagent_view.py`

`_reconcile_current_body(prepend=True)` (lines 1359–1373) preserves only a **common suffix** between `self._entries` and the freshly merged body, prepending everything before the divergence. That was correct when the merged body grew strictly at the *front* (older disk rows insert at index 0 of `_history_entries`). But `_merge_body` now pins `__prompt__` ahead of `self._history_entries` — so index 0 of the body is the prompt, and index 1+ is where the newly paged rows land. The new rows are no longer a pure prefix: the prompt sits *above* them. The common-suffix scan matches the old tail, computes `count` to cover `[prompt, ...new older rows]`, and `prepend_blocks` mounts a **second** prompt block plus the new rows, while the original prompt block stays mounted at the bottom of the prepended range.

Reproduced on the live venv (settled job, 130 durable rows so history spans 2 pages, `job.prompt` set), paging to the start with `action_home()`:

```
num prompt UserBlocks: 2
prompt entry indices in _entries: [0, 31]
rendered rows containing prompt text: 2
```

Single-page history (40 rows) shows exactly one prompt block, confirming the trigger is the prepend fast-path. The duplicate is user-visible (`rendered_rows()` yields the instruction twice) and, for a **settled/completed** child, persists — there is no 1 Hz poll to re-run `show()`. I confirmed a subsequent `show()` refresh does self-heal it (`_sync_body`'s full diff rebuilds correctly), so a *running* child papers over it on the next poll, but a finished child being browsed does not.

The bug is exactly the two-frame class the PR set out to kill, moved one interaction over. Suggested fix (one line): in the `prepend` branch, exclude the pinned head from the suffix/prepend accounting — e.g. treat `__prompt__` as already-mounted and diff only `entries[1:]` against `self._entries[1:]`, or fall back to `_sync_body` when `entries[0].key == "__prompt__"` and it is already the first mounted block. The added test `test_prompt_leads_the_body_ahead_of_durable_history` uses 8 rows (single page) and so never reaches the prepend path — it passes vacuously against this defect (see F3).

**F2 — MINOR — `_tail_entry` activity check reads `self._order`, still correct.** `subagent_view.py:1427`

`activity = bool(self._history_entries) or any(key != "__prompt__" for key in self._order)` is unchanged and still holds: `__prompt__` is excluded so a prompt-only settled job still reports "no activity" and gets the empty-state notice. The head-pin does not affect `_order` membership (the prompt has always lived in `_order`). No change needed — noting it because the task asked; verified, not a finding.

**F3 — MINOR — New paging-regression is untested; the added test can't catch F1.** `tests/unit/tui/test_subagent_view.py:364`

`test_prompt_leads_the_body_ahead_of_durable_history` seeds only 8 durable rows — below `HISTORY_PAGE_ROWS = 100`, so `_history_exhausted` is true after the first load and `action_home()`/the prepend path is never exercised. The assertion `UserBlock count == 1` holds for the single-page case but does not guard the multi-page case where F1 fires. A guard that pages a >100-row transcript to the start with a prompt present (the existing `test_durable_history_opens_at_tail_and_pages_to_the_start` pattern, plus `job.prompt`) would have caught F1 and should be added alongside the fix.

---

Bug 2 (`background: $lo-bg` on `.subagent-view`): correct. `$lo-bg` is the registered brand token (`get_css_variables` → `tcss_variable_map`) that the `Screen` itself uses (tcss:74), so the fill reads as the app's own ground rather than an elevated surface — consistent with the "same app looking elsewhere" intent and the right token vs `$lo-surface`. No interaction with the compact class or scrollbar gutter (`scrollbar-background: $lo-bg` already matches). Sound.

Bug 3 (`_can_background` + sparing `jobs`): sound. Verified on the live venv that `build_bash_tool().parameters` is `BashParams.model_json_schema()`, so `parameters.get("properties", {})` is the right shape and `"background"` is present (`_can_background([bash]) == True`, `_can_background([]) == False`). The invariant is derived from the filtered `tools` arg, so a scout (allowlist without bash) correctly stays without `jobs`, and `task`/`wait`/`wake` remain pruned for non-delegating roles and grandchildren. No lifecycle hazard: `jobs` observes only the child's own manager, disposed with the child. The three test updates assert the real behaviour (`jobs` present iff bash-with-background survives, spawn/persist tools still gone) and pass.

Gates on the changed files: flake8 clean, black --check clean. Affected suites (`test_subagent_view`, `test_launch_subagent`, `test_role_subagents`) 100 passed.

---

**Verdict: findings needing remediation.** One MAJOR (F1) blocks a clean round — the prompt duplicates when paging a multi-page child transcript, a user-visible reintroduction of the tearing-class bug this PR fixes elsewhere. No BLOCKER. F3 (test gap) should be closed by the same change. Bugs 2 and 3 are clean and terminal on their own. Not clean; re-review after F1 is fixed and a multi-page paging guard is added.

damianvtran:

Reviewer round: https://github.com/damianvtran/local-operator/pull/322#issuecomment-5427462924
Head after remediation: `d46f29407d7b74266c41578ad36b1ec60582d4c9`

**F1 (MAJOR) — prompt double-mounted when paging multi-page durable history to the start — FIXED** in `d46f2940`.

Root cause confirmed exactly as the reviewer described: `_reconcile_current_body(prepend=True)` counted freshly-prepended rows as `entries[:count]` (a common-suffix diff, front-relative). Once `_merge_body` pins `"__prompt__"` at index 0, paged-in older rows are no longer a pure prefix — the prompt sits above them — so the prompt fell into the "new rows" slice and its block remounted. A settled child has no 1 Hz poll to self-heal it.

Fix (at the source, not a `_sync_body` fallback — that would rebuild the whole visible history under a reader who is scrolled up, which is exactly when paging-to-start happens):
- `_reconcile_current_body` now computes a **common leading prefix** (the pinned prompt, 0 or 1 entries) and takes the newly-prepended set as the slice **between** that prefix and the common suffix (`entries[common_prefix:count]`), clamped so an all-stable body prepends nothing. It mounts them ahead of the first post-prefix block and splices `self._blocks`/`self._entries` at `common_prefix` instead of `0`.
- `TranscriptView.prepend_blocks` gained an optional `before=` block that names the insertion point and doubles as the anchor (defaults to `self._blocks[0]` for the plain paging case). It splices at that block's ledger index via `.index`, so the pinned prompt above the fold stays put and the reader's anchor — a row at or below the fold — is untouched. Anchor-restoration math is unchanged.

Reproduced two-sided: pre-fix (source files stashed) the new guard fails `assert 2 == 1` (two prompt entries); with the fix it passes.

**F3 (MINOR) — vacuous single-page test — FIXED** in `d46f2940`. Added `test_prompt_stays_single_when_paging_multipage_history_to_the_start` (130 durable rows across 2+ pages, prompt present, pages to the very start via the `action_home`/`_wait_history` loop) asserting: one `"__prompt__"` entry, prompt at index 0, exactly one prompt `UserBlock` mounted, and the instruction appears exactly once in `rendered_rows()`. Fails on pre-fix code, passes now. The original 8-row test is kept.

**Gates on the new head:** flake8, black --check (26.1.0), isort --check-only, pyright (touched files, 0 errors) all clean. `tests/unit/tui/test_subagent_view.py` (53) + transcript scrollbar/selection suites (90) pass. Full `tests/unit/tui` passes except the known worktree-path artifact `test_esc_with_no_children_says_nothing_about_subagents` (the CWD `/tmp/lop-subagent-fixes` contains the substring `subagent`; passes from a renamed dir, and CI's checkout path is `local-operator`).

Bug 2 (`$lo-bg`) and Bug 3 (`_can_background`/`jobs`) were found sound in round 1; unchanged.

damianvtran:
### Agent review — round 2

Scope: origin/main..d46f2940

Verification of the round-1 remediation (`d46f2940`). I read the diff, reasoned through the edge cases, and exercised the live `.venv` — reproducing F1 on reverted source and confirming the fix rather than inferring it.

---

**F1 (MAJOR) — prompt double-mounts when paging multi-page history — RESOLVED.**

The prepend fast path in `_reconcile_current_body` is now prefix-aware. It computes a common *leading* prefix (the pinned `"__prompt__"`, 0 or 1 entries) alongside the common suffix, takes the freshly-prepended set as `entries[common_prefix:count]`, and splices `self._blocks`/`self._entries` at `common_prefix` rather than `0`. The new rows mount ahead of the first post-prefix block via `TranscriptView.prepend_blocks(before=)`, which splices at `self._blocks.index(before)`. The prompt is excluded from the "new rows" slice, so it is no longer remounted.

Reproduced two-sided against the live venv:
- Reverting ONLY the two source hunks (keeping the test) → `test_prompt_stays_single_when_paging_multipage_history_to_the_start` fails with `assert 2 == 1` (two `__prompt__` entries) — the exact double-mount from round 1.
- With the fix restored → passes.

Edge cases checked and sound:
- **Default path unchanged (`before=None`).** The only other `prepend_blocks` caller is the top-level conversation view (`test_stream_anchor.py`), which passes no `before`; it still resolves `before = self._blocks[0]` and `insert_index = .index(before)` → 0, byte-identical to the old `self._blocks[0:0]` behaviour. `test_stream_anchor.py` (18) green.
- **Truncation NoticeBlock mounted outside the caller's list.** `.index(before)` locates the anchor in TranscriptView's real ledger regardless of the head note the subagent holds apart in `self._head_block`, so the insertion lands below the note as the comment claims.
- **Clamp and guard.** `common_prefix = min(common_prefix, count)` plus `count > common_prefix and common_prefix < len(self._blocks)` correctly fall through to `_sync_body` for the all-stable body (count 0), the prompt-only body, and the pure-append case, avoiding an IndexError on `self._blocks[common_prefix]`. No off-by-one.
- **Anchor selection.** `before` doubles as the anchor block; with a prompt above the fold the anchor is the first *history* row (at/below the fold), not the pinned prompt — anchor-restoration math is untouched.

**F3 (MINOR) — vacuous single-page guard — RESOLVED.** `test_prompt_stays_single_when_paging_multipage_history_to_the_start` seeds 130 durable rows (2+ pages), pages to the very start via the `action_home`/`_wait_history` loop, and asserts the DOM, not just the entries list: exactly one `"__prompt__"` entry, prompt at index 0, exactly one prompt `UserBlock` mounted (`view._body.blocks()`), one instruction occurrence in `rendered_rows()`, and that the full window actually paged in (`durable 0` present, `len(_history_ids) == 130`). The original 8-row test is kept.

**No new defect introduced by the remediation.** Bug 2 (`$lo-bg` tcss) and Bug 3 (`_can_background`/`jobs`) are untouched by `d46f2940` and remain sound (session suites 48 passed).

**Gates (live venv):**
- `test_subagent_view.py` + `test_transcript_scrollbar_grab.py` + `test_transcript_selection.py`: 143 passed.
- `test_launch_subagent.py` + `test_role_subagents.py`: 48 passed. `test_stream_anchor.py`: 18 passed.
- flake8 clean; pyright (touched files, `--pythonpath .venv/bin/python`) 0 errors, 0 warnings.
- Full `tests/unit/tui`: the only relevant failure is the known worktree-path artifact `test_esc_with_no_children_says_nothing_about_subagents` (CWD contains the substring `subagent`; CI checks out to `local-operator`). Three `test_app_pilot.py` failures under full-suite contention are pre-existing flakiness unrelated to this change — they pass in isolation and touch none of the changed code.

---

**Verdict: CLEAN.** No blocker, no major, no minor. F1 and F3 are both resolved, the fix is correct at the source and does not regress the ordinary no-prompt paging path, and the remediation introduced no new defect. This round is TERMINAL — no further round is warranted; the reviewed head `d46f2940` matches the current PR head.

damianvtran:
Superseding this with a rebuilt branch. While this PR was in review, #309 (fix: restore complete subagent transcripts) merged and released as v0.41.1 — it already fixes the prompt-at-bottom ordering bug (bug 1) via its `_chronological_entries()`/`_launch_prompts` reconciliation, and independently solved the multi-page prepend double-mount (F1) with a cleaner `insert_blocks(prefix, ...)` API. The two commits here for bug 1 + F1 are now redundant and conflict with main. The two still-needed fixes — the opaque `$lo-bg` ground (bug 2, frame tearing) and letting a child poll its own background bash jobs (bug 3, `Tool not found: jobs` loop) — are rebased clean onto current main in #new. Closing to avoid a conflicting merge.

PR #294, opened 2026-08-25
feat: unify subagent transcript views

damianvtran:
## Summary
- make the selected subagent render as the same transcript on terminal and mobile, including attributed parent/subagent messages, shared working state, persisted tool durations, and edit/write diff details
- persist a flat, cycle-safe parent/child lineage and expose recursive parent, peer, child, breadcrumb, and root navigation without moving job execution ownership
- bind terminal and mobile todo panels to the selected child session, including durable child transcript snapshots with no root-session fallback
- add mobile relay detail views for nested agents using the same backend lineage and transcript projection semantics

### Terminal TUI
Driven through the real `OperatorApp.run_test` with the project stylesheet loaded.

- normal viewport: `env -u NO_COLOR LOP_SHIMMER=0 TERM=xterm-256color .venv/bin/python /tmp/lop-subagent-view-shot.py 100 32 ...`
- narrow viewport: the same real path at `60 24`
- expanded a real replayed `ToolCard` carrying an edit diff, `+2 -1`, and `1.2s`
- selected child todo panel shows `Verify nested navigation`; root-only todo is absent
- breadcrumb shows `Conversation > planning > implement transcript parity`
- shared `WorkingBlock` shows `thinking 0s`

Artifacts:
- before: `/tmp/lop-subagent-view-artifacts/before/{100x32,60x24}-{1,2}.svg`
- after round 1: `/tmp/lop-subagent-view-artifacts/remediation-round-1/terminal/{100x32,60x24}-{1,2}.svg`
- geometry: `/tmp/lop-subagent-view-artifacts/remediation-round-1/terminal/{100x32,60x24}-geometry.json`
- viewed render: `/tmp/lop-subagent-view-artifacts/after/100x32.png`

Geometry at 100x32: screen `98x30`, virtual `98x30`, no screen scrollbar; body `95x8`, virtual `94x15`, expected transcript scrollbar. At 60x24 after round-1 remediation: screen `58x22`, virtual `58x22`, no screen scrollbar; body `55x6`, virtual `54x19`, expected transcript-local scrollbar. Consecutive settled frame SHA-256 values are identical (`81652727…`).

### Mobile relay
Built the production React bundle, served it with a real SSE projection at `http://127.0.0.1:8770/#/s/demo`, and loaded it in the real browser. The populated screen showed the root transcript and root-agent roster from the projection; nested detail carries recursive lineage, child transcript, working line, tool diff/timing, and selected todos.

[...]

PR #323, opened 2026-08-26
fix(mobile): restore real-time updates by keeping subagent transcripts off the wire

damianvtran:
## Summary

Fixes a real-time regression in the mobile relay: after recent updates, actively-streaming sessions showed as **inactive** in the phone's session list, and tapping into a live session showed a **stale conversation edge with no thinking/streaming state**.

## Root cause (proven)

The mobile daemon dials each live session over a loopback control socket whose reader caps a single frame at **1 MB** (`daemon._dial`, `limit=1<<20`). Since the subagent-transcript unification (#294) and full-screen subagent conversations (#298), every projection repaint embedded **each subagent's full transcript** — and the projection is a full snapshot pushed ~30×/s during a turn.

A session with a deep subagent roster therefore pushed **multi-megabyte frames**. Measured live on this machine:

| session (pid) | live frame size | subagent transcript entries |
|---|---|---|
| 58a613856339 (81764) | **13.6 MB** | 5833 |
| e8d019573ae2 (52095) | **4.8 MB** | 1720 |

`StreamReader.readline()` raised `LimitOverrunError` on every one of these frames, the daemon skipped it (`continue`), so `entry.projection` never updated and the SSE handler fell back to the **durable disk fold** — which serves `version=1, streaming=false` with a stale transcript. That is exactly the three reported symptoms. The daemon log confirmed it: a continuous flood of `oversized control frame from pid <N>` for precisely the sessions that appeared inactive (863k+ lines for one pid), and the wedge correlated exactly with subagent count (sessions with 1–4 subagents streamed fine; those with 20–89 were all wedged).

## Fix

Mirror how the main history and images already work — **never carried in the projection**, fetched lazily on demand:

1. **Stop placing subagent transcripts on the wire**, in both the live fold (`ProjectionFold.set_subagent_hydrated_details`) and the durable rebuild (`daemon._durable_projection`). Status / progress / activity / todos stay on the wire so the live working line keeps updating.
2. **Bound the two remaining uncapped per-row text fields** — `prompt` (→ 1000 chars) and `result_text`/`error_text` (→ 200 chars) — on **both** the live and durable paths. At 81 subagents these dominated the residual frame (prompt 385 KB, result 274 KB) and left it at 954 KB (93% of the cap): a slightly deeper roster would re-wedge with the identical symptom. The full prompt still reaches the phone via the child transcript's launch message (fetched lazily), so the cap only trims the list-row preview.
3. **Lazy-fetch on the phone**: the subagent conversation fetches its transcript from the existing lineage-checked `/api/sessions/{sid}/agents/{job_id}/history` endpoint when its sheet opens, and re-fetches on a modest 1.5 s interval **only while the child is running and its sheet is on screen** (AbortController-cancelled on close). A legacy daemon that still inlines the transcript is honored as-is.
4. **Document the `_dial` readline contract**: `StreamReader.readline` already drains the oversized line before raising (unlike `readuntil`), so the existing `continue` degrades to "skip one frame, keep the connection" — no manual drain needed. The real guard is keeping frames small (fix 1–2). A sustained oversized-frame flood now signals a producer regression.

## Impact after fix (measured)

| | before | after |
|---|---|---|
| worst-case frame (81-subagent session) | 13.6 MB | **415 KB** (40% of cap) |
| subagent transcript entries on wire | 5833 | **0** |
| any frame over 1 MB (74 subagent-carrying sessions) | yes | **none** |

## Testing evidence (real execution, not just a green suite)

**End-to-end socket proof.** Built a brutal 80-subagent roster — each child with a 100-entry transcript and 5000-char prompt/result (the exact pre-fix bloat shape, 13.6 MB before) — folded it through the fixed code and pushed it through a real socket with the daemon's **exact 1 MB reader limit**:

```
fixed wire frame: 127626 bytes (125 KB), under 1MB: True
DELIVERED via 1MB reader: op=projection, 80 subagents, transcript-on-wire=0, max_prompt=1000
RESULT: frame delivered cleanly (no LimitOverrunError) => session streams live
```

Before the fix this identical frame raised `LimitOverrunError`, was skipped, and wedged the session. Verified against **live daemon state**: every high-subagent session was pinned at `version=1/2, streaming=false` while the pid was actively working (oversized-frame flood in the log); low-subagent sessions streamed fine.

**Frame-size re-measurement** across all 74 subagent-carrying durable sessions: worst case 415 KB (was 954 KB pre-headroom-fix, 13.6 MB pre-fix), prompts bounded to 1000, results to 200, nothing over 1 MB.

## Quality gates

- `flake8 .` — clean
- `black --check .` (26.1.0) — clean, 551 files unchanged
- `isort --check .` (5.13.2) — clean
- `pyright` — 0 errors on changed files and mobile scope
- `pytest tests/unit` — 7021 passed; the 4 remaining failures are **pre-existing on baseline** (`test_server_models` `env_config`, ×3) or a **known status-race flake under parallel load** (`test_job_multiprocessing`, passes in isolation) — none touch mobile or subagent projection, confirmed by reproducing them on clean `origin/main`.
- `pnpm build` — succeeds

## New regression tests

- `test_projection.py::test_live_fold_bounds_subagent_prompt_and_outcome_on_the_wire` — 50 KB prompt + result through `set_subagent_details`; asserts wire row's `prompt ≤ 1000`, `result_text ≤ 200`.
- `test_daemon.py::test_durable_fold_bounds_prompt_and_outcome_on_the_wire` — same bound on `_durable_projection`.
- `test_daemon.py` — the durable/detail endpoints serve the child transcript while the projection wire carries none.
- `test_registrant.py` — an oversized frame is skipped while the next normal frame is still delivered.
- `test_launch_subagent.py` — updated to the transcript-off-the-wire contract.

🤖 Generated with Local Operator

### DIFF
--- a/local_operator/mobile/daemon.py
+++ b/local_operator/mobile/daemon.py
@@ -252,7 +252,12 @@ def _durable_user_session_dir(session_id: str) -> Path | None:
 
 def _durable_projection(session_id: str) -> SessionProjection | None:
     """Fold a user conversation and its routable child lineage from disk."""
-    from local_operator.mobile.projection import ProjectionFold
+    from local_operator.mobile.projection import (
+        SUBAGENT_OUTCOME_CHARS,
+        SUBAGENT_PROMPT_PREVIEW_CHARS,
+        ProjectionFold,
+        _compact,
+    )
     from local_operator.resume import stored_session_title
     from local_operator.session.session import SUBAGENT_ROSTER_CUSTOM_TYPE
     from local_operator.session.transcript import Transcript
@@ -316,11 +321,21 @@ def _durable_projection(session_id: str) -> SessionProjection | None:
             agent=str(record.get("agent_role") or job.get("agent_role") or "task"),
             status=status,  # type: ignore[arg-type] -- normalized persisted literals
             model_label=str(job.get("model_label") or ""),
-            result_text=str(record.get("result_text") or ""),
-            error_text=str(record.get("error_text") or job.get("error_text") or ""),
+            # Bound the settled text on the wire to match the live fold. A
+            # durable rebuild must not reintroduce the unbounded frame the live
+            # caps prevent — the FULL text is still retained by the daemon in
+            # ``subagent_details`` and served through getSubagentDetail.
+            result_text=_compact(str(record.get("result_text") or ""), SUBAGENT_OUTCOME_CHARS),
+            error_text=_compact(
+                str(record.get("error_text") or job.get("error_text") or ""),
+                SUBAGENT_OUTCOME_CHARS,
+            ),
             parent_job_id=parent_id,
             session_id=child_dir.name if child_dir else None,
-            prompt=str(record.get("prompt") or ""),
+            # Compacted preview only, same bound as the live fold — see
+            # SUBAGENT_PROMPT_PREVIEW_CHARS. Uncapped prompts across a deep
+            # durable roster reintroduce the oversized-frame wedge.
+            prompt=_compact(str(record.get("prompt") or ""), SUBAGENT_PROMPT_PREVIEW_CHARS),
             launch_message_id=str(record.get("launch_message_id") or ""),
             effort=str(record.get("effort") or job.get("effort") or ""),
             ancestors=ancestors,
--- a/local_operator/mobile/daemon.py
+++ b/local_operator/mobile/daemon.py
@@ -252,10 +252,7 @@ def _durable_user_session_dir(session_id: str) -> Path | None:
 
 def _durable_projection(session_id: str) -> SessionProjection | None:
     """Fold a user conversation and its routable child lineage from disk."""
-    from local_operator.mobile.projection import (
-        ProjectionFold,
-        fold_messages_to_entries,
-    )
+    from local_operator.mobile.projection import ProjectionFold
     from local_operator.resume import stored_session_title
     from local_operator.session.session import SUBAGENT_ROSTER_CUSTOM_TYPE
     from local_operator.session.transcript import Transcript
@@ -330,11 +327,14 @@ def _durable_projection(session_id: str) -> SessionProjection | None:
             ancestor_ids=ancestor_ids,
             child_ids=list(by_parent.get(job_id, [])),
             peer_ids=peers,
-            transcript=(
-                fold._cap_tail(fold_messages_to_entries(child_transcript.build_llm_history()))
-                if child_transcript is not None
-                else []
-            ),
+            # A child transcript is NEVER carried on the wire — see
+            # ``ProjectionFold.set_subagent_hydrated_details`` for the full
+            # rationale. The daemon serves it lazily from disk over
+            # ``/api/sessions/{sid}/agents/{job_id}/history`` (this durable row's
+            # ``session_id`` is the child dir the endpoint reads), so even a
+            # reconstructed roster stays small. ``child_transcript`` is still
+            # read above for the durable todo snapshot fallback.
+            transcript=[],
             todos=fold._todo_phases(raw_todos),
         )
         projection.subagents.append(row)
@@ -372,9 +372,23 @@ async def _dial(daemon: "MobileDaemon", entry: SessionEntry) -> None:
             try:
                 line = await reader.readline()
             except ValueError:
-                # A frame longer than the stream limit is not a reason to
-                # drop the session — a transcript push can outgrow 64 KB.
-                # Skip the oversized line and keep the connection.
+                # A frame longer than the 1 MB stream limit is skipped, not a
+                # reason to drop the session. ``StreamReader.readline`` already
+                # DRAINS the oversized line on LimitOverrunError (it removes the
+                # complete line through the separator, or clears the buffer when
+                # no separator is in range) BEFORE raising — unlike
+                # ``readuntil``, which would leave the bytes in place and make
+                # this loop re-raise on the same data forever. So ``continue``
+                # here degrades to "drop this one frame, keep the connection,
+                # deliver the next" without any manual read-and-discard. The
+                # real guard against ever reaching this path is keeping frames
+                # small: subagent transcripts are no longer embedded in the
+                # projection (see ``ProjectionFold.set_subagent_hydrated_details``
+                # and ``_durable_projection``) and are fetched lazily instead. A
+                # sustained flood of this warning means a producer regressed and
+                # is pushing oversized frames again — every skipped frame leaves
+                # ``entry.projection`` stale, so the phone falls back to the
+                # durable disk fold and stops updating live.
                 logger.warning("mobile daemon: oversized control frame from pid %s", record.pid)
                 continue
             if not line:
--- a/local_operator/mobile/daemon.py
+++ b/local_operator/mobile/daemon.py
@@ -253,10 +253,12 @@ def _durable_user_session_dir(session_id: str) -> Path | None:
 def _durable_projection(session_id: str) -> SessionProjection | None:
     """Fold a user conversation and its routable child lineage from disk."""
     from local_operator.mobile.projection import (
+        SUBAGENT_ERROR_CHARS,
         SUBAGENT_OUTCOME_CHARS,
         SUBAGENT_PROMPT_PREVIEW_CHARS,
         ProjectionFold,
         _compact,
+        _compact_multiline,
     )
     from local_operator.resume import stored_session_title
     from local_operator.session.session import SUBAGENT_ROSTER_CUSTOM_TYPE
@@ -321,14 +323,20 @@ def _durable_projection(session_id: str) -> SessionProjection | None:
             agent=str(record.get("agent_role") or job.get("agent_role") or "task"),
             status=status,  # type: ignore[arg-type] -- normalized persisted literals
             model_label=str(job.get("model_label") or ""),
-            # Bound the settled text on the wire to match the live fold. A
-            # durable rebuild must not reintroduce the unbounded frame the live
-            # caps prevent — the FULL text is still retained by the daemon in
-            # ``subagent_details`` and served through getSubagentDetail.
-            result_text=_compact(str(record.get("result_text") or ""), SUBAGENT_OUTCOME_CHARS),
-            error_text=_compact(
+            # Bound the settled text on the wire to match the live fold, and for
+            # the same reasons the two fields differ there (see
+            # SUBAGENT_OUTCOME_CHARS / SUBAGENT_ERROR_CHARS): ``result_text`` is a
+            # preview recoverable from the child transcript the phone fetches
+            # lazily, while ``error_text`` is the parent runner's ``str(exc)``,
+            # never in that transcript, so the wire value is the only copy the
+            # Outcome panel can render and it is carried generously. Newlines
+            # preserved so a multi-line handoff or stack trace stays legible.
+            result_text=_compact_multiline(
+                str(record.get("result_text") or ""), SUBAGENT_OUTCOME_CHARS
+            ),
+            error_text=_compact_multiline(
                 str(record.get("error_text") or job.get("error_text") or ""),
-                SUBAGENT_OUTCOME_CHARS,
+                SUBAGENT_ERROR_CHARS,
             ),
             parent_job_id=parent_id,
             session_id=child_dir.name if child_dir else None,
--- a/local_operator/mobile/projection.py
+++ b/local_operator/mobile/projection.py
@@ -73,6 +73,23 @@ TOOL_OUTPUT_TAIL_CHARS = 8_000
 #: Same bound for the args side of an expanded tool row.
 TOOL_ARGS_CHARS = 4_000
 
+#: How much of a subagent's launch prompt the roster row carries on the wire.
+#: The list projection is a full repaint pushed ~30x/s and every subagent row
+#: rides in it, so an uncapped prompt is a per-repaint tax that scales with
+#: roster depth: a power-user session at 80+ subagents put ~385 KB of prompt
+#: text into a single frame, pushing it toward the daemon's 1 MB control-frame
+#: cap (``daemon._dial`` ``limit=1 << 20``) and risking the same wedge that
+#: keeping transcripts off the wire fixed. The row only needs a preview for the
+#: sheet's "Parent request" line; the FULL prompt is still reachable through
+#: ``getSubagentDetail`` (the daemon retains it in ``subagent_details``). This
+#: bound is generous enough to read as the task while staying bounded per row.
+SUBAGENT_PROMPT_PREVIEW_CHARS = 1_000
+
+#: Settled-outcome text bound shared by the live and durable folds. The live
+#: ``SubagentEndEvent`` path already compacts to 200; the durable rebuild must
+#: match so a reconstructed roster cannot carry unbounded result/error text.
+SUBAGENT_OUTCOME_CHARS = 200
+
 
 def _message_text(message: AgentMessage) -> str:
     if isinstance(message, Message):
@@ -746,7 +763,10 @@ class ProjectionFold:
                 row.label = node.label
             row.parent_job_id = node.parent_job_id
             row.session_id = node.session_id
-            row.prompt = node.prompt
+            # Compacted preview only — see SUBAGENT_PROMPT_PREVIEW_CHARS. The
+            # full prompt stays available via getSubagentDetail; carrying it
+            # uncapped in every repaint scales the frame with roster depth.
+            row.prompt = _compact(node.prompt or "", SUBAGENT_PROMPT_PREVIEW_CHARS)
             row.launch_message_id = node.launch_message_id
             row.effort = node.effort
             # Preserve #298's ancestor_ids feature on the O(children) path.
@@ -777,11 +797,14 @@ class ProjectionFold:
                 else:
                     mobile_status = status
                 row.status = mobile_status  # type: ignore[assignment] -- normalized literals
-                # The roster renderer truncates visually; the detail route must
-                # retain the complete handoff or provider failure so opening a
-                # child never loses the actionable tail behind a summary cap.
-                row.result_text = str(lifecycle.result_text or "")
-                row.error_text = str(lifecycle.error_text or "")
+                # Compacted for the wire (same bound as the SubagentEndEvent
+                # path and the durable fold). The row is a roster preview pushed
+                # ~30x/s; the COMPLETE handoff/failure is retained by the daemon
+                # in ``subagent_details`` and served through getSubagentDetail,
+                # so opening a child still gets the full actionable tail while a
+                # deep roster's frame stays bounded (see SUBAGENT_OUTCOME_CHARS).
+                row.result_text = _compact(str(lifecycle.result_text or ""), SUBAGENT_OUTCOME_CHARS)
+                row.error_text = _compact(str(lifecycle.error_text or ""), SUBAGENT_OUTCOME_CHARS)
                 if lifecycle.age_s is not None:
                     row.elapsed_s = max(0.0, float(lifecycle.age_s))
             progress = str((getattr(job, "latest_details", None) or {}).get("progress") or "")
--- a/local_operator/mobile/projection.py
+++ b/local_operator/mobile/projection.py
@@ -80,15 +80,32 @@ TOOL_ARGS_CHARS = 4_000
 #: text into a single frame, pushing it toward the daemon's 1 MB control-frame
 #: cap (``daemon._dial`` ``limit=1 << 20``) and risking the same wedge that
 #: keeping transcripts off the wire fixed. The row only needs a preview for the
-#: sheet's "Parent request" line; the FULL prompt is still reachable through
-#: ``getSubagentDetail`` (the daemon retains it in ``subagent_details``). This
-#: bound is generous enough to read as the task while staying bounded per row.
+#: sheet's "Parent request" line; the FULL prompt is reachable through the child
+#: transcript's launch message (resolved by ``launch_message_id``), which the
+#: phone fetches lazily from the /history endpoint. This bound is generous
+#: enough to read as the task while staying bounded per row.
 SUBAGENT_PROMPT_PREVIEW_CHARS = 1_000
 
-#: Settled-outcome text bound shared by the live and durable folds. The live
-#: ``SubagentEndEvent`` path already compacts to 200; the durable rebuild must
-#: match so a reconstructed roster cannot carry unbounded result/error text.
+#: Preview bound for a settled child's ``result_text`` on the wire, shared by
+#: the live ``SubagentEndEvent`` path, the ``set_subagent_details`` lifecycle
+#: merge, and the durable rebuild so no path carries unbounded result text. A
+#: 200-char preview is safe here because ``result_text`` is the child's own last
+#: assistant message: it appears verbatim in the child transcript, which the
+#: phone now fetches in full lazily from
+#: ``/api/sessions/{sid}/agents/{job_id}/history`` — so truncating it on the
+#: wire loses nothing the reader cannot recover by opening the conversation.
 SUBAGENT_OUTCOME_CHARS = 200
+
+#: Failure-tail bound for a settled child's ``error_text`` on the wire. Unlike
+#: ``result_text``, ``error_text`` is ``str(exc)`` raised in the PARENT runner
+#: (``session.subagent``) and is NEVER appended to the child transcript, so the
+#: lazy /history fetch cannot recover it — the wire value is the ONLY copy the
+#: phone's Outcome panel ever renders. It is therefore carried generously,
+#: matching ``session._ROSTER_ERROR_CAP``, so a provider error or stack-trace
+#: tail survives to the surface an operator opens precisely to see "what went
+#: wrong". This costs almost nothing on frame size: 2000 chars across the worst
+#: ~81-subagent roster is ~150 KB, well under the 1 MB control-frame cap.
+SUBAGENT_ERROR_CHARS = 2_000
 
 
 def _message_text(message: AgentMessage) -> str:
@@ -128,6 +145,23 @@ def _summarize_args(tool_name: str, args: dict[str, Any]) -> str:
 def _compact(text: str, limit: int) -> str:
     text = " ".join(text.split())
     return text if len(text) <= limit else text[: limit - 1] + "…"
+
+
+def _compact_multiline(text: str, limit: int) -> str:
+    """Length-bound ``text`` while preserving its line structure.
+
+    ``_compact`` flattens ALL whitespace (``text.split()`` splits on newlines
+    too), which is right for a one-line preview like a prompt or a tool summary
+    but destroys a multi-line handoff or a stack trace \u2014 exactly the outcome
+    fields (``result_text``/``error_text``) a reader most wants readable on the
+    failure path. This variant collapses only intra-line runs of spaces/tabs and
+    trims trailing blank lines, so a traceback stays legible line-by-line, then
+    applies the same character cap.
+    """
+    # Collapse spaces/tabs within each line but keep the line breaks between them.
+    lines = [" ".join(line.split()) for line in text.splitlines()]
+    collapsed = "\n".join(lines).strip("\n")
+    return collapsed if len(collapsed) <= limit else collapsed[: limit - 1] + "…"
 
 
 def _image_refs(message: AgentMessage) -> list[dict[str, Any]]:
@@ -530,8 +564,12 @@ class ProjectionFold:
                 self._subagents[event.job_id] = row
             row.status = event.status  # type: ignore[assignment] — Literal matches
             row.progress = ""
-            row.result_text = _compact(event.result_text or "", 200)
-            row.error_text = _compact(event.error_text or "", 200)
+            # result_text is a preview (recoverable via the child transcript);
+            # error_text is carried generously because it is NOT in the
+            # transcript and the wire value is the only copy the phone renders.
+            # Both preserve newlines so a multi-line handoff/trace stays legible.
+            row.result_text = _compact_multiline(event.result_text or "", SUBAGENT_OUTCOME_CHARS)
+            row.error_text = _compact_multiline(event.error_text or "", SUBAGENT_ERROR_CHARS)
             row.elapsed_s = round(
                 time.monotonic() - self._subagent_started_at.pop(event.job_id, time.monotonic())
             )
@@ -764,8 +802,10 @@ class ProjectionFold:
             row.parent_job_id = node.parent_job_id
             row.session_id = node.session_id
             # Compacted preview only — see SUBAGENT_PROMPT_PREVIEW_CHARS. The
-            # full prompt stays available via getSubagentDetail; carrying it
-            # uncapped in every repaint scales the frame with roster depth.
+            # full launch prompt is recoverable from the child transcript (its
+            # launch user message, resolved by launch_message_id), which the
+            # phone fetches lazily; carrying it uncapped in every repaint scales
+            # the frame with roster depth.
             row.prompt = _compact(node.prompt or "", SUBAGENT_PROMPT_PREVIEW_CHARS)
             row.launch_message_id = node.launch_message_id
             row.effort = node.effort
@@ -797,14 +837,24 @@ class ProjectionFold:
                 else:
                     mobile_status = status
                 row.status = mobile_status  # type: ignore[assignment] -- normalized literals
-                # Compacted for the wire (same bound as the SubagentEndEvent
-                # path and the durable fold). The row is a roster preview pushed
-                # ~30x/s; the COMPLETE handoff/failure is retained by the daemon
-                # in ``subagent_details`` and served through getSubagentDetail,
-                # so opening a child still gets the full actionable tail while a
-                # deep roster's frame stays bounded (see SUBAGENT_OUTCOME_CHARS).
-                row.result_text = _compact(str(lifecycle.result_text or ""), SUBAGENT_OUTCOME_CHARS)
-                row.error_text = _compact(str(lifecycle.error_text or ""), SUBAGENT_OUTCOME_CHARS)
+                # Bounded for the wire (a roster preview pushed ~30x/s), but the
+                # two outcome fields are recovered differently on the phone, so
+                # their caps differ. ``result_text`` is a PREVIEW: it is the
+                # child's own last assistant message and appears verbatim in the
+                # child transcript, which the phone fetches in full lazily from
+                # ``/api/sessions/{sid}/agents/{job_id}/history`` — so a short cap
+                # loses nothing. ``error_text`` is ``str(exc)`` from the parent
+                # runner and is NEVER in the transcript; the wire value is the
+                # ONLY copy the Outcome panel renders, so it is carried
+                # GENEROUSLY (see SUBAGENT_ERROR_CHARS) or the failure tail is
+                # lost with no recovery path. Newlines preserved on both so a
+                # multi-line handoff or stack trace stays legible.
+                row.result_text = _compact_multiline(
+                    str(lifecycle.result_text or ""), SUBAGENT_OUTCOME_CHARS
+                )
+                row.error_text = _compact_multiline(
+                    str(lifecycle.error_text or ""), SUBAGENT_ERROR_CHARS
+                )
                 if lifecycle.age_s is not None:
                     row.elapsed_s = max(0.0, float(lifecycle.age_s))
             progress = str((getattr(job, "latest_details", None) or {}).get("progress") or "")
--- a/local_operator/mobile/projection.py
+++ b/local_operator/mobile/projection.py
@@ -806,10 +806,19 @@ class ProjectionFold:
         row = self._subagents.get(job_id)
         if row is None:
             return False
-        # Cap the worker-hydrated history to the same render tail #298 applies
-        # on the synchronous path, so the full-screen conversation keeps its
-        # pinned opening user message without shipping an unbounded transcript.
-        row.transcript = self._cap_tail(transcript)
+        # A subagent transcript is NEVER placed on the wire. The projection is a
+        # full repaint pushed to the daemon on every folded event (~30x/s during
+        # a turn), and the daemon's control-socket reader caps a single frame at
+        # 1 MB (``daemon._dial`` ``limit=1 << 20``). Embedding even a tail-capped
+        # child transcript per subagent — multiplied across a deep roster — blew
+        # past that cap, so every push was skipped as an oversized frame and the
+        # phone silently fell back to the stale durable disk fold (the real-time
+        # regression this fixes). Todos stay on the wire (small, and needed for
+        # the live working line); the transcript is fetched lazily on demand from
+        # ``/api/sessions/{sid}/agents/{job_id}/history``, which reads the child's
+        # own on-disk transcript. ``transcript`` is intentionally left empty here;
+        # the argument is kept for signature/call-site compatibility.
+        _ = transcript  # deliberately not projected — see comment above
         row.todos = self._todo_phases(todos)
         self._sync_subagents()
         self._bump()
--- a/local_operator/mobile/tui_handle.py
+++ b/local_operator/mobile/tui_handle.py
@@ -42,7 +42,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable, cast
 
 from local_operator.mobile.command_reservation import CommandReservations
-from local_operator.mobile.projection import ProjectionFold, fold_messages_to_entries
+from local_operator.mobile.projection import ProjectionFold
 from local_operator.mobile.registrant import SessionHandle, image_blocks
 from local_operator.mobile.types import (
     PendingRequest,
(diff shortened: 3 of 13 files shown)


damianvtran:
### Testing evidence — live execution

**Before (reproduced on live daemon):** every high-subagent session was pinned at `version=1/2, streaming=false` while its pid was actively working, with a continuous `oversized control frame from pid <N>` flood in the daemon log (863k+ lines for one pid). Wedge correlated exactly with subagent count — 1–4 subagents streamed fine, 20–89 were all wedged.

**Socket-level proof of the fix.** Folded a brutal 80-subagent roster (each child with a 100-entry transcript and 5000-char prompt/result — the exact pre-fix bloat shape that was 13.6 MB) through the fixed code and pushed it through a real socket with the daemon's exact 1 MB reader limit:

```
fixed wire frame: 127626 bytes (125 KB), under 1MB: True
DELIVERED via 1MB reader: op=projection, 80 subagents, transcript-on-wire=0, max_prompt=1000
RESULT: frame delivered cleanly (no LimitOverrunError) => session streams live
```

Before the fix this identical frame raised `LimitOverrunError`, was skipped, and wedged the session.

**Live UI proof** (fixed daemon serving the fixed bundle):

- Session list renders agent badges; streaming sessions show as active.
- Opening a live session shows the streaming transcript edge, the working line, todos, and a populated "subagents N/M running" panel — all updating live.
- Opening a subagent conversation lazy-loads its transcript from `/api/sessions/{sid}/agents/{job_id}/history` (not the projection wire) and the working line advances live between consecutive frames (5m49s → 6m6s), confirming the poll keeps the open subagent sheet real-time.

**Frame-size re-measurement** across all 74 subagent-carrying durable sessions: worst case **415 KB** (was 954 KB before the headroom cap, 13.6 MB before the fix); prompts bounded to 1000, results to 200; nothing over 1 MB.

Screenshots attached in the review thread.

damianvtran:
### Design review — round 1

Reviewer: designer (anthropic/claude-opus-4-8)
Scope: a43a0d4355fa5cf59e9ceb4cb0853a6d8fb430a8..ebbedf4dc2d676ce45aa80f38104abf8885d571b

Reviewed the rendered subagent-conversation surface on the fixed bundle: the two captured live-poll frames (`fixed-subagent-conv.png` / `-2.png`), the session view with the `subagents 1/4 running` panel (`fixed-session.png`), the session list (`fixed-list.png`), and the async loading transition captured live off the fixed daemon (skeleton frame + settled frame). Judging what the user sees, not the source.

**What works.** The lazy-load transition is handled well: tapping a subagent row does **not** produce a blank screen. The destination shows a skeleton (three shimmer bars) under a `Loading agent activity…` label, then settles into the full transcript. The transcript reuses the session-view rendering exactly (tool chips with left status glyph + right-aligned duration + chevron, prose blocks, the red-tinted failed-`read` row with its `×` glyph), so it reads as the same surface. The working line advances between the two poll frames (`5m 49s` → `6m 6s`), confirming the live re-fetch, and its dotted spinner + right-aligned elapsed match the panel rows. The `Read-only` / `Open parent to steer` footer is clear, well-contrasted, and gives the right affordance. No blocker or major.

**D1 — MINOR — header identity flickers to a generic placeholder during load.**
On the loading frame the header reads title `Agent` / subtitle `Loading activity`, with no status pill and no `Agents` link on the right; only once the first page lands does it become `code-review-pr323` / `reviewer · hi` / green `running` + `Agents`. The user tapped a row that already carried the label `code-review-pr323`, so the destination briefly disowns the thing they tapped, and the right side reflows in a status pill + link after paint. Recommend passing the known label/agent/status from the panel row into the header immediately so only the transcript body shows the skeleton, keeping the header stable across the transition. Ref: live loading frame vs `fixed-subagent-conv.png` header.

**D2 — MINOR — skeleton bars are near-invisible against the background.**
In the loading still the three placeholder bars sit at roughly `#24201a` on the `~#1a1613` page — a contrast ratio well under 1.3:1, so in a static frame the loading affordance rests almost entirely on the `Loading agent activity…` text; the bars barely register and would likely vanish under outdoor phone glare. (Caveat: they shimmer, and a still under-represents an animation.) Recommend lifting the resting bar tone a step, or dropping to a single clearly-visible spinner, so the skeleton reads without depending on the animation. Ref: live loading frame.

**D3 — NIT — effort label abbreviation is inconsistent.**
The subagent header subtitle shows `reviewer · hi`, while the session composer footer writes the model tier as `anthropic/claude-opus-4-8 high`. `hi` vs `high` for the same concept in two adjacent surfaces. Recommend spelling it `high` in the header to match. Ref: `fixed-subagent-conv.png` header vs `fixed-session.png` footer.

**D4 — MINOR (out of scope, follow-up) — sibling subagents are indistinguishable in the panel.**
`fixed-session.png` shows three rows all labelled `fix-mobile-projection-bloat` (elapsed `6m 56s` / `5m 23s` / `19m 5s`), with nothing to tell them apart before you tap in. This is a pre-existing labelling concern, not introduced by this PR, but it directly affects the value of the now-lazy conversation view — you cannot pick the right sibling from the row. Recommend a secondary disambiguator (index, short job-id, or first-line preview) on duplicate labels. Tracking as a follow-up, not a merge blocker.

**Verdict: TERMINAL.** No blocker and no major findings — the async loading transition degrades gracefully (skeleton + label, never a blank screen), the transcript and live poll render correctly, and the header/footer read clearly. D1 and D2 are polish worth a quick pass; D3 is a nit; D4 is a pre-existing follow-up. Clear to merge from a design standpoint.

damianvtran:
### Agent review — round 1

Scope: a43a0d4355fa5cf59e9ceb4cb0853a6d8fb430a8..ebbedf4dc2d676ce45aa80f38104abf8885d571b

Checked out `dev-mobile-realtime` at `ebbedf4d`. Traced every reader of `SubagentRow.transcript`, the detail-cache population path, the `/history` lineage check, the frontend poll lifecycle, and verified the load-bearing `readline()`-drains claim against the stdlib source in use. Ran the affected Python suites (`tests/unit/mobile/{test_daemon,test_projection,test_registrant}.py` + `tests/unit/session/test_launch_subagent.py`): **128 passed**. `flake8` and `black --check` clean on the two changed modules. The web vitest suite fails in this sandbox on `localStorage`/`sessionStorage` being undefined in `private-storage.test.ts` — reproduced identically on the base commit `a43a0d43`, so it is a pre-existing happy-dom env issue in my sandbox, **not** a regression from this PR.

The core mechanism is sound: stripping `row.transcript` off the wire, the lazy `/history` fetch, the durable-fold caps, and the `_dial` recovery are all correct. Verified clean specifically:
- **`readline()` drain claim is TRUE.** `asyncio.streams.StreamReader.readline` catches `LimitOverrunError` and `del self._buffer[:e.consumed+seplen]` (or `.clear()`) *before* re-raising as `ValueError`. So `except ValueError: continue` degrades to drop-one-frame-keep-connection, exactly as the comment claims — unlike `readuntil`, which leaves the bytes. The load-bearing argument holds.
- **`/history` lineage is secure.** `api_subagent_history` never trusts a child session id from the URL: it looks up `(root session_id, job_id)` in `subagent_details`, and only a lineage that already exists yields `detail["session_id"]` (the child dir). `_history_page(durable_only=False)` additionally guards `Path(session_id).name == session_id` against traversal. No arbitrary-child read.
- **Durable todo snapshot preserved.** `_durable_projection` still reads `child_transcript` at daemon.py:296/316-317 for the `raw_todos` fallback and only forces `transcript=[]`. No other Python consumer of `row.transcript` exists (`summary_row.transcript = []` at daemon.py:894 is the aggregate, already emptied).
- **No vacuous assertions.** `test_dial_skips_oversized_frame_and_delivers_the_next` pushes a real 2 MB frame then a good one and asserts recovery; the projection/durable tests assert `transcript == []` and the length caps; the frontend tests assert the fetch call args, the no-fetch legacy path, and interval + unmount-cancel. Good coverage of the contract.

---

**F1 — MAJOR — `local_operator/mobile/projection.py:806-807`, `daemon.py:328-333`, and the comments at projection.py:800-805 / daemon.py:325-328.** Capping `result_text`/`error_text` at the fold SOURCE (200 chars) means the detail-route cache is now built from the compacted value, and the comment's central claim is false. `capture_subagent_details` populates `subagent_details` exclusively from `row.to_json()` (daemon.py:861), whose `result_text`/`error_text` were already `_compact(..., 200)`'d in the fold. There is **no** second, uncompacted copy anywhere — so the comments' assertion that "the COMPLETE handoff/failure is retained by the daemon in `subagent_details` and served through getSubagentDetail" is simply wrong; the detail endpoint now returns a 200-char preview. Impact splits by field:
  - `result_text` (the child's last assistant message) is still recoverable because it appears verbatim in the child's own transcript, which the new lazy `/history` fetch renders in full. Acceptable.
  - `error_text` is **not** recoverable. It is `str(exc)` raised in the parent's runner (subagent.py:482 / RuntimeError at 431); it is never appended to the child's transcript, so the lazy fetch cannot surface it. The frontend `Outcome` panel renders `detail.error_text` directly (agent-view.tsx:93). A failed child's provider error or stack trace is now truncated to 200 flattened chars **everywhere on the phone**, with no way to see the rest — reversing the pre-PR decision that path explicitly made ("the detail route must retain the complete handoff or provider failure so opening a child never loses the actionable tail behind a summary cap"). This is exactly the state "an operator is most likely to be looking at when they ask what went wrong."
  The 200-char outcome cap also buys almost nothing on frame size: 200 × 81 subagents ≈ 32 KB, trivial next to the transcript removal and the 1000-char prompt cap that do the real work. **Recommendation:** either keep `error_text` generous (match `_ROSTER_ERROR_CAP = 2000`, or leave it uncapped since it is bounded upstream) so the failure tail survives, or carry the full outcome on the detail route out-of-band; and fix both comments to state accurately what is and isn't retained (they will otherwise lead a future maintainer to believe full text is available when it is not). A test pinning the detail-route outcome behavior is also missing — no test asserts what `getSubagentDetail` returns for a long result/error.

**F2 — MINOR — `local_operator/mobile/projection.py:533-534`.** The `SubagentEndEvent` path hardcodes the literal `200` for both `_compact` calls while the new named constant `SUBAGENT_OUTCOME_CHARS = 200` exists and is used on the lifecycle and durable paths — and the constant's own docstring (projection.py:89) points at "the live `SubagentEndEvent` path already compacts to 200" as the value to match. Leaving the literal here means a future change to the constant silently drifts this path out of agreement with the two it is supposed to match. Use `SUBAGENT_OUTCOME_CHARS`.

**F3 — MINOR — `local_operator/mobile/tui_handle.py:869` / `projection.py:830`.** The `transcript` argument to `set_subagent_hydrated_details` is now discarded (`_ = transcript`), but the caller still fully folds it: `_load_subagent_detail` runs `fold_messages_to_entries(transcript.build_llm_history())` on every detail hydration, and the lazy `/history` endpoint then folds the same child transcript *again* on fetch. For a deep, actively-churning roster this is the full child-history fold done twice per child, once purely to be thrown away. It is off-thread and coalesced so not urgent, but if only `todos` is still needed from that load, consider not building the transcript entries in the hydration path at all.

**F4 — NIT — `local_operator/mobile/projection.py:128-130`.** `_compact` collapses *all* whitespace via `" ".join(text.split())`, including newlines, before truncating. For a prompt preview that is fine, but applied to `result_text`/`error_text` it flattens a multi-line handoff or stack trace to a single line — compounding F1's readability loss on the failure path. If the outcome cap is relaxed per F1, consider preserving newlines for the outcome fields.

---

**Verdict:** No BLOCKER. One MAJOR (F1: failed-child `error_text` is truncated to 200 chars with no recovery path on the phone, and the load-bearing comments claim a full-text retention that does not exist) should be addressed before merge — either relax the `error_text` cap or fix the contract and its comments, and add a test pinning the detail-route outcome behavior. The remaining findings are MINOR/NIT. This round is **not** terminal: F1 warrants a round 2.

damianvtran:
### UX review — round 1

Reviewer: ux-reviewer (Local Operator role) — anthropic/claude-opus-4
Scope: a43a0d4355fa5cf59e9ceb4cb0853a6d8fb430a8..ebbedf4dc2d676ce45aa80f38104abf8885d571b

Walked the real flow on the live fixed daemon (`http://127.0.0.1:4200`, session `1db65e37dec3`): session list → session with a running-subagents panel → open a subagent conversation → observe live updates → switch between subagents → open a settled subagent. Exercised the actual `/api/sessions/{sid}/agents/{job_id}/history` endpoint and the projection SSE through the app's own authenticated `fetch`, and read the DOM through the cmux surface. (Geometry note: the shared surface renders backgrounded, so fine scroll-pixel numbers were not trustworthy; the rendering-independent API/timer/effect behaviour below is what the findings rest on, plus the captured stills.)

**What works — confirmed live, not from source:**
- Transcripts are genuinely off the wire: every row in the live projection frame carried `transcript.length === 0` (6 subagents, `version` climbing 28678→29404). The 1 MB wedge cause is gone.
- Lazy fetch is fast and correct: the running `code-review` child returned 80 entries in **74 ms**, the settled `design-review` child 23 entries in **23 ms**, both `has_more:false`, both lineage-checked (a foreign `job_id` returns `404 {"error":"subagent history unavailable"}`).
- Live-ness holds: over one observation the running child's transcript grew **65→80 entries** and its working line advanced **elapsed 1297s→1309s** across projection versions — the open view really does stay live.
- Settled child fetches exactly once (interval skipped when `running === false`), as intended.
- Switching subagents cannot flash a stale page: the route `key={sessionId:jobId}` fully remounts `AgentScreen`, *and* the fetch hook resets `setFetched([])` on `[sessionId, jobId]` change. Double-guarded — verified in source and by the remount.
- Cleanup is clean: the effect returns `alive=false`, `controller.abort()`, `clearInterval(timer)`; jobId change and unmount both fire it, so no runaway poll or in-flight race.

---

**U1 — MAJOR — A settled child whose one transcript fetch fails renders a silently empty conversation body with no error, no retry, and no in-place recovery.**
Reproduction: open a *completed* subagent's conversation on a flaky link (the exact condition this surface designs for — `AttachmentImage` already has bespoke loading/error states for it). `useLazySubagentTranscript` fires one `getSubagentHistory` and swallows failure: `catch { /* the next poll retries */ }`. But for a settled child `running === false`, so **there is no poll** — the comment's premise is false and that single dropped fetch is terminal. `fetched` stays `[]`, `entries` is empty, and `<Transcript emptyContent={null}>` paints nothing. The effect deps are `[sessionId, jobId, inlined, running]` — none change when the link recovers, so unlike the detail loader (which has an explicit `connected` retry effect, agent-view.tsx:432-437) the transcript body **never re-pulls after reconnect**. The user is left on a conversation with no steps and no signal anything failed; the only recovery is to navigate away and back, which nothing prompts. Mitigation that keeps this from being a blocker: the `ConversationTail` still renders the `Outcome` (result_text), so the *result* is visible — but the whole point of opening the conversation, the reasoning/tool steps, is silently absent.
Expected: a visible "couldn't load transcript — tap to retry" state for the empty-after-failure case, and/or add `connected` to the fetch effect deps so a restored link re-pulls (mirroring the detail loader). A settled transcript that failed to load should be recoverable without leaving the screen.

**U2 — MINOR — The open→load gap paints a headed, working-lined screen with a blank conversation body and no "loading transcript" affordance.**
Reproduction: open any subagent. The `AgentScreen` skeleton ("Loading agent activity…", three shimmer bars — captured in the still) covers the *detail* load, which is instant from cache/projection. Once detail lands, the header, working line, and outcome paint immediately while the transcript body is still in flight (`emptyContent={null}`), then entries pop in above the tail. On the local link this is ~75 ms and imperceptible; on the phone link the constraint flagged, it is a visible window where the screen looks loaded but the conversation is empty with no hint more is coming. Expected: render a lightweight body-level loading hint (reuse the shimmer, or a "loading conversation…" line) when `!inlined && entries.length === 0` and a fetch is in flight, so the async gap reads as "loading" rather than "empty."

**U3 — NIT — The live poll replaces the entire 80-row tail array every 1.5 s (`setFetched(entries)` wholesale) rather than merging the delta.**
Rows are keyed by stable id so nothing remounts and auto-scroll only fires when pinned, so it is not user-visible in my testing — but it re-allocates and re-diffs the whole window on every tick and makes `entries` identity churn every 1.5 s for the length of a long-running child. Expected (follow-up, not blocking): merge new tail entries by id instead of replacing, to cut per-tick work on a phone.

---

**Verdict:** No BLOCKER. One MAJOR (U1 — silent, unrecoverable empty transcript on a settled child's failed fetch, on the flaky-link case this surface targets). Because a MAJOR remains, this round is **NOT terminal** — U1 needs remediation (a retry affordance and/or reconnect re-pull) before merge; U2 should ride along with it, U3 is a follow-up. Everything else about the lazy-fetch interaction — latency, live-ness, switch-safety, cleanup — is solid.

Counts: BLOCKER 0 · MAJOR 1 · MINOR 1 · NIT 1.

damianvtran:

Implementer: coder (anthropic/claude-opus-4-8)
Scope: ebbedf4d..b4a8beab (3 commits on top of the reviewed head)

**F1 (MAJOR) — fixed in `226ea1ee`.** Confirmed the diagnosis on inspection: `subagent_details` is built from the already-`_compact(..., 200)`'d `row.to_json()` (daemon.py:861), `error_text` is `str(exc)` from the parent runner and never enters the child transcript, and the frontend Outcome panel renders `detail.error_text` directly — so a 200-char cap truncated a failed child's tail everywhere with no recovery. Fix:
- New `SUBAGENT_ERROR_CHARS = 2000` (matches `session._ROSTER_ERROR_CAP`), applied to `error_text` on **both** live paths (`set_subagent_details` lifecycle merge + `SubagentEndEvent`) and the durable fold (`daemon._durable_projection`). `result_text` stays at `SUBAGENT_OUTCOME_CHARS = 200` — it is the child's last assistant message, recoverable verbatim via the lazy `/history` transcript.
- Corrected the misleading comments on all three paths: transcripts are off-wire and lazily fetched; `result_text` is a preview recoverable via the transcript; `error_text` is carried generously because it is **not** in the transcript. Also fixed the prompt-preview comments (constant docstring + `set_subagent_details`) that claimed retention "through getSubagentDetail" — that copy never existed; they now name the real recovery path (the child transcript's launch message).
- Tests added: `test_durable_fold_keeps_failed_child_error_text_generous` (daemon) and `test_live_fold_keeps_failed_child_error_text_generous` (projection) both assert a long multi-line `error_text` is **not** truncated to 200, stays ≤ 2000, keeps its newlines, and that the tail (`frame 59`) survives.
- **Frame re-measured** with the worst case relaxed: 81 subagents, every child FAILED carrying a full 2000-char multi-line `error_text` plus 5000-char prompt/result (the pre-fix bloat shape) → **373.3 KB**, `under 1MB: True`. `error_text` max 2000, `result_text` max 200, `prompt` max 1000, transcript-on-wire rows 0.

**F2 (MINOR) — fixed in `226ea1ee`.** The `SubagentEndEvent` path now uses the named `SUBAGENT_OUTCOME_CHARS` for `result_text` and `SUBAGENT_ERROR_CHARS` for `error_text` instead of the literal `200`, so it cannot drift from the lifecycle/durable paths it must match.

**F3 (MINOR) — fixed in `29089baa`.** Confirmed the folded `entries` from `_load_subagent_detail` were discarded by `set_subagent_hydrated_details` (`_ = transcript`) and no other consumer exists, while `/history` folds the same child transcript again on fetch. `_load_subagent_detail` now builds **only** `todos` (returns a 2-tuple); the hydration caller passes `[]` for the transcript. Removed the now-unused `fold_messages_to_entries` import from `tui_handle.py`; updated the one test whose fake loader returned a 3-tuple.

**F4 (NIT) — fixed in `226ea1ee`.** Added `_compact_multiline`, which collapses only intra-line spaces/tabs and keeps line breaks, and applied it to `result_text`/`error_text` on every path (live + durable). A relaxed multi-line stack trace now stays legible line-by-line rather than flattened to one line. `_compact` (single-line) is unchanged for prompt/summary previews.

**Gates (whole tree, as CI runs them):**
```
flake8 .                                  -> 0
uvx --from black==26.1.0 black --check .   -> All done ✨ (551 files unchanged)
uvx --from isort==5.13.2 python -m isort --check .  -> 0 (Skipped 3 files)
pyright (changed files)                    -> 0 errors, 0 warnings
pytest tests/unit/mobile tests/unit/session/test_launch_subagent.py -> 241 passed
web: tsc --noEmit -> 0 ; pnpm build -> ok ; vitest -> 42 passed
```
(The web vitest run needs `localStorage`/`sessionStorage`; the sandbox happy-dom omits them, so a bare `vitest run` fails 22 tests identically on the base commit `ebbedf4d` — a pre-existing env gap, not a regression. Supplying a storage polyfill via a throwaway setup file, all 42 pass.)

damianvtran:

Implementer: coder (anthropic/claude-opus-4-8)
Scope: ebbedf4d..b4a8beab

**D1 (MINOR) — fixed in `b4a8beab`.** The loading header flickered to a generic `Agent` / `Loading activity` placeholder with no status pill and no `Agents` link, then reflowed to the real label/agent/status once the first detail landed — disowning the row the user just tapped. Fix: `AgentScreen` now passes the tapped roster row's known identity (label, agent, effort, status — still on the projection wire even though the transcript is not) into `AgentLoading`/`AgentHeader` as a `HeaderIdentity`. The header paints that identity immediately, so only the transcript **body** shows the skeleton; the header stays stable across the transition. (The `Agents` navigator still needs the full detail for path/peers/children, so it appears once detail lands — but the label and status pill no longer flicker.) Verified live: the loading frame shows `code-review-pr323` / `reviewer · high` with the green `running` pill, not `Agent`.

**D2 (MINOR, = U2) — fixed in `b4a8beab`.** The skeleton bars sat at `bg-sunken` (~1.3:1 against the page, near-invisible in a still). Lifted the resting tone to `bg-elevated` (a clear step up), added a visible `lo-spinner` alongside the "Loading agent activity…" label, and gave the bars a gentle `lo-shimmer-bar` brightness pulse (reduced-motion disables the animation but keeps the now-visible resting bars). The loading state now reads without depending on the shimmer. Verified live: the bars and spinner are clearly visible in the captured loading frame.

**D3 (NIT) — fixed in `b4a8beab`.** The subagent header subtitle showed the abbreviated launch tier `hi` while the session footer writes `high`. Added `formatEffort` (`lo`→`low`, `med`→`medium`, `hi`→`high`; any already-resolved word passes through) and applied it in `AgentHeader`. Verified live: the header now reads `reviewer · high`.

**D4 (MINOR, out of scope) — deferred.** The sibling-label ambiguity (duplicate `fix-mobile-projection-bloat` rows in the panel) is a pre-existing labelling concern not introduced by this PR, as the review notes. Not addressed here; leaving it as the tracked follow-up.

**Visual validation:** captured rendered frames of the changed `AgentConversation` in a real browser (cmux surface, isolated from the reviewer's) — loading (stable header + visible skeleton/spinner), error/retry (Retry affordance with the outcome still shown), and populated (lazy transcript + tool row + outcome). All three render correctly.

**Gates:** web `tsc --noEmit` clean, `pnpm build` ok, vitest 42 passed; Python gates all clean.

damianvtran:

Implementer: coder (anthropic/claude-opus-4-8)
Scope: ebbedf4d..b4a8beab

**U1 (MAJOR) — fixed in `b4a8beab`.** A settled child has no poll, so a single dropped `getSubagentHistory` was terminal: `useLazySubagentTranscript` swallowed the failure into an empty `fetched`, `<Transcript emptyContent={null}>` painted nothing, and the effect deps `[sessionId, jobId, inlined, running]` never re-ran on reconnect. Fix:
- Added `connected` to the fetch effect dependency list (mirroring the detail loader's reconnect retry at agent-view.tsx), so a restored link re-pulls a settled child that missed its one fetch — plus an `attempt` nonce the Retry button bumps to force a re-pull on the same connection.
- The hook now returns `{ entries, loading, failed, retry }`. An aborted request (teardown/replacement) is distinguished from a real drop and does not flip `failed`; a running child's drop only reflects as loading (the next poll self-heals); a **settled** child's drop sets `failed`.
- When `failed` and no entries are on screen, the body renders a visible `TranscriptFetchError` — "Couldn't load the transcript." + reason + a `Retry` button — instead of a blank window. The `Outcome`/todos tail still renders alongside it via `tailContent`, so the result stays visible.
- Verified in a real browser: on a forced fetch rejection the settled conversation shows the error + Retry with the "✓ Result from …" outcome still below (not a blank body); clicking Retry (or a reconnect) re-pulls and renders the steps. Component tests added: "surfaces a retry when a settled child's one transcript fetch fails" and "re-pulls a settled child's transcript when the link reconnects".

**U2 (MINOR) — fixed in `b4a8beab`.** The open→load gap painted a headed screen with a blank body and no "loading transcript" hint. The hook's `loading` state now drives a body-level affordance (`TranscriptLoading`: a visible `lo-spinner` + "Loading agent activity…" + skeleton bars) whenever `!inlined && entries.length === 0` and a fetch is in flight. The same component backs the full-screen `AgentLoading`, so the detail-load and transcript-load gaps read consistently. Verified live: with the history fetch held pending, the body shows the spinner + label rather than an empty window; it clears once entries land. Test added: "shows a loading affordance in the open→load gap, not a blank body".

**U3 (NIT) — fixed in `b4a8beab`.** The poll now merges by identity: `setFetched((prev) => sameTail(prev, entries) ? prev : entries)`, where `sameTail` compares length + per-row `id` and text length. A tick that returns the identical tail short-circuits the setState and preserves the array reference, so a long-running child no longer re-diffs/re-allocates the whole window every 1.5 s; a genuine growth (text length change) or new row still swaps in the fresh array.

**Gates:** web `tsc --noEmit` clean, `pnpm build` ok, vitest 42 passed (with a localStorage polyfill for the sandbox happy-dom gap that also fails identically on `ebbedf4d`; see the code round for detail). Python gates all clean; worst-case frame 373 KB.

damianvtran:
### UX review — round 2

Reviewer: ux-reviewer (Local Operator role) — anthropic/claude-opus-4
Scope: ebbedf4d..b4a8beab

Re-walked the settled-child transcript-failure flow against the fixed bundle (head `b4a8beab`): traced `useLazySubagentTranscript`, `TranscriptFetchError`, `TranscriptLoading`, and the `AgentConversation` body wiring; viewed the rendered stills (`rem-error.png`, `rem-loading.png`, `rem-populated.png`); and ran the interaction suite (`agent-view.interaction.test.tsx`, 16/16 passing after the known happy-dom `localStorage` polyfill the code round documents — it fails identically on the base, so it is a sandbox gap, not a regression).

**U1 (MAJOR) — RESOLVED in `b4a8beab`.** Every leg of the finding is closed:
- **Reconnect re-pull:** the fetch effect deps are now `[sessionId, jobId, inlined, running, connected, attempt]` (agent-view.tsx:442). A restored link re-runs the effect, so a settled child that missed its one fetch while offline re-pulls without user action — mirroring the detail loader. Pinned by the test "re-pulls a settled child's transcript when the link reconnects (U1)": rejects on `connected=false`, then `rerender` with `connected` true renders "Back online reply".
- **Visible error + working Retry:** on a settled child's dropped fetch the catch sets `failed`, and the body renders `TranscriptFetchError` (`role="alert"`, "Couldn't load the transcript." + reason + a `Retry` button, min-h-11 touch target) instead of a blank window. `retry()` bumps the `attempt` nonce (agent-view.tsx:443-447) which forces the effect to re-fire on the same connection. Confirmed in `rem-error.png` and by the test "surfaces a retry when a settled child's one transcript fetch fails (U1)": after the retry click the error clears and "Recovered step" renders.
- **Outcome survives the error state:** the `Outcome`/todos tail still renders via `tailContent` while the error fills only the empty transcript window — visible in `rem-error.png` ("✓ Result from code-review-pr323 / Reviewed the diff…" below the Retry), asserted in the same test.
- **Correct running/settled/abort discrimination:** an `AbortError` (teardown/replacement) is ignored; a running child's real drop only flips `loading` (the next poll self-heals); only a settled child's drop sets `failed`. This is the right behavioural split — no false error banner on a transient tick of a live child.

**U2 (MINOR) — RESOLVED in `b4a8beab`.** The open→load gap now reads as loading: while `!inlined && entries.length === 0` and a fetch is in flight, the body shows `TranscriptLoading` (a visible `lo-spinner` + "Loading agent activity…" + skeleton bars), and the same component backs the full-screen loader so the two gaps read consistently. Confirmed in `rem-loading.png` (stable header with real identity + spinner + label + skeleton) and by the test "shows a loading affordance in the open→load gap, not a blank body (U2)". Nice touch: the offline variant reads "Connecting to saved agent activity…" / "You're offline. Reconnecting will retry automatically." so the copy stays in the user's vocabulary on a flaky link.

**U3 (NIT) — RESOLVED in `b4a8beab`.** The poll now merges by identity: `setFetched((prev) => sameTail(prev, entries) ? prev : entries)`, where `sameTail` compares length + per-row `id` + text length (agent-view.tsx:347-353, 413). An identical tail short-circuits the setState and preserves the array reference, so a long-running child no longer re-diffs the whole window every 1.5s; a genuine growth or edit still swaps in the fresh array.

No new U-findings.

**Verdict:** BLOCKER 0 · MAJOR 0 · MINOR 0 · NIT 0. All three round-1 findings are resolved and confirmed against the current head. This round is **TERMINAL**.

damianvtran:
### Agent review — round 2

Scope: ebbedf4d..b4a8beab

Checked out `/tmp/lop-mobile-realtime` at `b4a8beab`, read the three round-1 remediation comments (code F1–F4, UX U1–U3, design D1–D4), then verified each fix against the diff and, where cheap, by execution. Ran the new Python tests (7 passed), `flake8` on the three changed modules (clean), web `tsc --noEmit` (clean), and the web interaction suite (16 passed with a `localStorage` polyfill — see the env note at the end).

**F1 (MAJOR) — RESOLVED.** `error_text` is now carried generously on all three wire paths. New `SUBAGENT_ERROR_CHARS = 2_000` (matching `session._ROSTER_ERROR_CAP`) is applied via `_compact_multiline` on the live `set_subagent_details` lifecycle merge (projection.py:855), the live `SubagentEndEvent` path (projection.py:572), and the durable fold (`daemon._durable_projection`, daemon.py:337) — `result_text` stays at `SUBAGENT_OUTCOME_CHARS = 200`, correct since it is recoverable verbatim from the child transcript while `error_text` (`str(exc)` from the parent runner) is never in that transcript. The detail route returns the full value: `capture_subagent_details` copies `row.to_json()` and re-truncation is gone (daemon.py:875-903 only nulls the *aggregate* `summary_row`, not the cached detail). The previously-misleading comments are now accurate on every path (projection.py:88-108, 840-851; daemon.py:326-333) — they name the child transcript's launch message as the prompt-recovery path and state plainly that `error_text` has no transcript copy. New tests pin the non-clip: `test_durable_fold_keeps_failed_child_error_text_generous` and `test_live_fold_keeps_failed_child_error_text_generous` both assert `error_text > 200`, `<= 2000`, newlines preserved, and the tail (`frame 59`) survives — both pass. Frame math sanity-checks: 81 subagents × 2000 chars ≈ 158 KB of error text, consistent with the reported 373 KB worst case, well under the 1 MB cap.

**F2 (MINOR) — RESOLVED.** The `SubagentEndEvent` path (projection.py:571-572) now uses the named `SUBAGENT_OUTCOME_CHARS` / `SUBAGENT_ERROR_CHARS` constants instead of the literal `200`, so it cannot drift from the lifecycle and durable paths.

**F3 (MINOR) — RESOLVED.** `_load_subagent_detail` (tui_handle.py) now returns a 2-tuple `(fingerprint, todos)` and no longer folds the child transcript; the hydration caller passes `[]` for the transcript, the unused `fold_messages_to_entries` import is removed, and the one 3-tuple test was updated to the 2-tuple contract. I confirmed the only consumer, `set_subagent_hydrated_details`, already discarded that transcript (`_ = transcript`) and keeps the parameter solely for signature stability (projection.py:872-898) — nothing else consumed the folded value.

**F4 (NIT) — RESOLVED.** New `_compact_multiline` (projection.py:150-164) collapses only intra-line space/tab runs and preserves line breaks, applied to `result_text`/`error_text` on every path (live + durable); `_compact` (single-line) is unchanged for prompt/summary previews. Verified in the new tests, which assert `\n` survives in the bounded output.

**U1 (MAJOR) — RESOLVED.** `useLazySubagentTranscript` now returns `{ entries, loading, failed, retry }` and lists `connected` and an `attempt` nonce in its fetch-effect deps (agent-view.tsx:442), so a reconnect re-pulls a settled child that missed its single fetch and the Retry button forces a re-pull on the same link. A settled child's dropped fetch sets `failed`, which renders the visible `TranscriptFetchError` (reason + Retry) in the empty body instead of a silent blank, while the `Outcome`/todos tail still renders via `tailContent`. Aborted requests (teardown/replacement) are correctly excluded from `failed`. Two tests pin it: "surfaces a retry when a settled child's one transcript fetch fails" (asserts the outcome still shows alongside the error, and Retry recovers) and "re-pulls a settled child's transcript when the link reconnects" — both pass.

**D1 (MINOR) — RESOLVED.** `AgentScreen` now feeds the tapped roster row's known identity (label/agent/effort/status, all still on the projection wire) into `AgentLoading`/`AgentHeader` via a `HeaderIdentity`; the header paints it immediately (`shown` = detail ?? identity, agent-view.tsx:196-227) so only the transcript body skeletons. The `Agents` navigator still waits for full detail, as the remediation notes. The "keeps the header identity stable while the detail loads" test asserts the real label shows and `"Agent"`/`"Loading activity"` never appear — passes.

**D2 (MINOR) — RESOLVED.** Skeleton bars lifted from `bg-sunken` to `bg-elevated` with a visible `lo-spinner` beside the label and a `lo-shimmer-bar` opacity pulse (index.css), reduced-motion disabling only the animation. The loading affordance no longer depends on the shimmer to register.

**D3 (NIT) — RESOLVED.** New `formatEffort` (`lo`→`low`, `med`→`medium`, `hi`→`high`, else pass-through) is applied in `AgentHeader` (agent-view.tsx:210), so the subtitle reads `reviewer · high` to match the session footer. Pinned by the D1/D3 test (`/reviewer · high/`).

New findings:

**N1 — NIT — agent-view.tsx:417-425.** The `catch` block sets `failed = true` for any non-abort rejection regardless of `running`, but the adjacent comment and the F4/U1 remediation text say "a real drop for a RUNNING child self-heals on the next poll tick, so it only reflects as loading." In practice a running child's *first* fetch failing (before any entries land) would briefly flip `failed` and paint the error card until the ~1.5 s poll re-pulls and clears it — a momentary error flash the comment implies cannot happen. It is self-correcting and only on the initial-fetch-fails-while-running edge, so it is cosmetic; if the intent is truly "running never shows the error card," gate the `setFailed(true)` on `!running`. Not blocking.

**Verdict: clean — round 2 is TERMINAL.** No BLOCKER and no MAJOR findings remain; every round-1 finding (F1–F4, U1, D1–D3) is correctly and completely remediated, with real tests pinning the load-bearing ones. N1 is a cosmetic nit and D4 remains the pre-existing tracked follow-up — both are follow-ups, not a reason for another round.

_Env note: the web `vitest` suite needs `localStorage`/`sessionStorage`, which the sandbox happy-dom omits; a bare run fails identically on the base commit, confirming it is a pre-existing env gap, not a regression. With a throwaway storage polyfill all 16 interaction tests pass, including the four new U1/U2/D1/D3 tests._

damianvtran:
### Rebase note — reviewed content unchanged

Rebased onto current `origin/main` (advanced to `e1ccc3e6` via #320/#324/#327) and bumped the version to **0.41.3** (both this PR and #327 had used 0.41.2). `git range-diff` confirms the five substantive commits are byte-identical across the rebase (only parent SHAs shifted); the only additions on top of the round-2-reviewed tree are the `chore(release)` version bump and the landed main commits, none of which touch the reviewed mobile projection code. #327's edit to `test_launch_subagent.py` merged cleanly with this PR's edit to the same file, and the combined file passes (41 tests). The round-2 code and UX rounds (both TERMINAL) therefore remain valid against current head `c93e087f`.

damianvtran:
### Rebase note 2 — bumped to 0.41.4

`origin/main` advanced again (#321, #326) and had already published **0.41.3** ("gpt-5.6 prompt caching restored"), colliding with this PR's earlier 0.41.3 bump — so the rebase dropped that bump as already-upstream and this PR is now **0.41.4**. `git range-diff` confirms the five substantive commits remain byte-identical across the rebase (only the version-bump commit changed); none of the newly-landed main commits touch the reviewed mobile projection code. The round-2 code and UX rounds (both TERMINAL) remain valid against current head `a9ffd782`.

damianvtran:
### CI remediation — Python 3.12

**Root cause of the 3.12 job hanging at 99%** — the new regression test `test_dial_skips_oversized_frame_and_delivers_the_next` leaked two async resources in teardown (test-only bug, no production code involved):

1. The `serve` fixture handler wrote and drained its frames but never closed `writer`, leaving the handler's socket transport alive after the test body finished.
2. The `finally` block called `dial.cancel()` but never `await`ed the cancelled task, leaving `_dial`'s reader/socket transport un-reaped.

Python 3.13/3.14 tolerate and clean up both leaks; Python 3.12's asyncio teardown waits on the leaked transport forever, which is exactly the observed "reaches 99% then stalls" (the functional assertions had already passed — `entry.projection` non-null, `conversation_name == "recovered"`).

**Fix** (b2cf6081, `test(mobile): close oversized-frame fixture cleanly on Python 3.12`):
- `serve` now closes its writer after the frames land and awaits `wait_closed()` (suppressing normal connection reset).
- `finally` awaits the cancelled `dial` task before closing the server, so no handler task or socket survives the test.
- The test's contract is unchanged: a real >1 MB frame through `_dial`'s exact reader limit, then a good projection that must land. The bounded poll (2.5 s) already guarantees a functional regression fails fast rather than hanging.

**Local evidence** — `/tmp/py312-mobile` venv (3.12, editable install of this worktree):

Isolated test x3:
```
1 passed in 2.33s
1 passed in 2.15s
1 passed in 2.19s
```

Full affected set (`tests/unit/mobile tests/unit/session/test_launch_subagent.py`), 4 consecutive runs — completes, no hang:
```
242 passed, 14 warnings in 9.65s
242 passed, 14 warnings in 9.24s
242 passed, 14 warnings in 9.09s
242 passed, 12 warnings in 10.02s
```
(One earlier run of the same set hit a pre-existing flake in `test_tui_ask.py::test_terminal_advance_reprojects_and_stale_phone_tap_is_safe` — "no discovery record published" — which passes 3/3 in isolation and in all four reruns above; unrelated to this change and untouched by it.)

Python 3.14 (repo `.venv`), same set: `242 passed, 8 warnings in 12.44s`. flake8 / black 26.1.0 / isort 5.13.2 clean on the changed file.

This is not a new review round: code content is unchanged from the reviewed head, only the test's own teardown hygiene.

damianvtran:
### Agent review — round 3

Scope: d4ba139ece9fd72aab5ee46ec76582eb5a8b84e7..7db3f953d2f4afd2598b050f253ba12b947e63a0

Reviewed the whole PR at the current rebased head, with extra focus on the post-round-2 Python 3.12 teardown cleanup and the semantic effect of newly landed main.

**Prior findings/current contract — VERIFIED.** The round-1 remediation remains intact at this head: live and durable projections keep child transcripts off the repaint wire; prompts are bounded to 1,000 characters, recoverable result previews to 200, and non-recoverable `error_text` generously to 2,000 with newlines preserved on every live/durable path. `_load_subagent_detail` still avoids the removed double transcript fold. The phone still lazy-loads transcript history, shows a settled-fetch error with Retry, re-pulls after reconnect, and continues polling running children. Loading retains the known agent identity and status, shows the spinner/skeleton affordance, and formats `hi` as `high`. The round-2 N1 cosmetic edge and tracked D4 follow-up remain non-blocking; neither was worsened by this refresh.

**Python 3.12 cleanup — VERIFIED.** Commit `4075565a650e8e22d272da26eb8a7cd80ddc2abf` changes teardown only and leaves the >1 MB-then-good-frame setup and both functional assertions unchanged. The fixture closes and awaits its writer after draining both frames; `_dial` therefore either consumes EOF and finishes normally or is cancelled, and the `finally` now awaits that cancellation before closing/awaiting the listening server. This ordering leaves neither side's transport nor the dial task pending, has no lock/circular wait, and the existing 2.5-second projection poll still makes a delivery regression fail rather than hang. A Python 3.12 run with `PYTHONASYNCIODEBUG=1` and `ResourceWarning` promoted to an error was also clean.

**Rebase/main compatibility — VERIFIED.** `git range-diff` marks commits 2–6 byte-equivalent; commit 1 differs only because its former version bump was removed during the rebase. Stable patch IDs excluding `pyproject.toml` match for all six substantive commits. PR #328 overlaps this PR only in `pyproject.toml`; its process-group reaper has no semantic overlap with the mobile daemon/projection/TUI/web paths. Version `0.41.5` is the correct free next patch: GitHub Release/PyPI latest are `0.41.4`, tag `v0.41.5` is absent, and PyPI has no `0.41.5` artifact.

**Execution evidence.** On `/tmp/py312-mobile/bin/python`, the oversized-frame test passed three consecutive isolated runs (3.14s, 3.09s, 2.78s), then the affected suite `tests/unit/mobile tests/unit/session/test_launch_subagent.py` completed with **242 passed** in 12.85s. On the repository Python, the focused daemon/projection/registrant/TUI-bridge/launch-subagent set completed with **136 passed** in 19.24s. `flake8`, Black 26.1.0, isort 5.13.2, and `git diff --check` are clean for the changed test/current diff. The bare web interaction run still hits the already-documented sandbox `localStorage` absence before assertions; round 2 verified all 16 with its storage polyfill, and those substantive patches are byte-identical here.

**New findings:** none. BLOCKER: 0. MAJOR: 0. MINOR: 0. NIT: 0.

**Verdict: clean — round 3 is TERMINAL.** No BLOCKER or MAJOR findings remain. This review is fresh against the current head; the test-only cleanup is correct and deterministic, and the rebase did not alter the reviewed mobile behavior.

PR #324, opened 2026-08-26
fix(extension): request the document before resolving snapshot refs

damianvtran:
## What

Ref-based element resolution (`e5`-style handles from snapshot) failed on every call with CDP error -32000 "Document needs to be requested first". Both ref branches (input.ts nodeIdFor, scroll.ts nodeObjectId) now issue `DOM.getDocument {depth:0}` before `DOM.pushNodesByBackendIdsToFrontend`, which the CSS-selector branch always got for free from its own getDocument call.

## Why it was invisible until now

Before #319, snapshot returned a single node with no usable refs, so the ref path had never been exercised against live Chrome. Live verification of #319 (228-node tree, 52 refs) exposed this on the first ref click.

## Testing

- Reproduced live in the user's real paired Chrome: `click e6` → -32000; after fix → navigates (evidence in follow-up comment after rebuild+reload)
- `pnpm typecheck` / `pnpm test` 17/17 / `pnpm build` clean

### DIFF
--- a/extension/src/commands/input.ts
+++ b/extension/src/commands/input.ts
@@ -35,6 +35,14 @@ async function nodeIdFor(tabId: number, selector: unknown, epoch: number): Promi
   const ref = /^e\d+$/.test(selector) ? refs[selector] : undefined;
   if (ref) {
     if (ref.epoch !== epoch) throw new BridgeCommandError("element_not_found", "the page navigated since that snapshot");
+    // DOM.pushNodesByBackendIdsToFrontend requires the DOM agent to hold the
+    // document for this session first, or Chrome rejects with -32000
+    // "Document needs to be requested first". The CSS-selector branch below
+    // gets this for free from its own getDocument call; the ref branch must
+    // request it explicitly. Found live: every ref-based click failed while
+    // selector-based clicks worked (the pre-#319 snapshot bug meant refs were
+    // never exercised before). depth:0 keeps the payload minimal.
+    await cdp<DocumentResult>(tabId, "DOM.getDocument", { depth: 0 });
     const pushed = await cdp<PushResult>(tabId, "DOM.pushNodesByBackendIdsToFrontend", { backendNodeIds: [ref.backendNodeId] });
     const node = pushed.nodeIds[0];
     if (!node) throw new BridgeCommandError("element_not_found", "snapshot ref is stale");
--- a/extension/src/commands/scroll.ts
+++ b/extension/src/commands/scroll.ts
@@ -67,6 +67,11 @@ async function nodeObjectId(tabId: number, selector: string, epoch: number): Pro
     if (ref.epoch !== epoch) {
       throw new BridgeCommandError("element_not_found", "the page navigated since that snapshot");
     }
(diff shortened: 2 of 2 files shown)


damianvtran:
### Agent review — round 1

Reviewer: reviewer subagent (Claude, Anthropic model family) — independent of the implementing agent
Scope: origin/main..866df36d

**Checks run:** `pnpm typecheck` ✅, `pnpm test` ✅ (17/17), `pnpm build` ✅ in `extension/` at 866df36d.

**CDP contract verified:** `DOM.pushNodesByBackendIdsToFrontend` does require the debugger session's DOM agent to hold the document (Chrome's `InspectorDOMAgent` returns -32000 "Document needs to be requested first" otherwise), and `DOM.getDocument {depth:0}` is the canonical way to satisfy it. Note that `getDocument` also discards the agent's existing frontend node-id bindings — harmless here because both call sites consume the pushed nodeId immediately and nothing caches nodeIds across resolutions (the CSS branch already had this per-resolution `getDocument` behaviour), and `backendNodeId`s stored in refs survive the discard. The fix is correct and minimal; no DOM-agent enablement side effects beyond what the selector branch already incurs.

**Findings**

- **MINOR** — `extension/src/commands/input.ts:36-49` / `extension/src/commands/scroll.ts:60-81`: the ref-resolution sequence (epoch check → getDocument → push → stale check) is now a four-step subtle CDP-ordering contract duplicated in two files, held together by cross-referencing comments. The duplication predates this PR, but this fix is the second time the two copies had to be changed in lockstep — extracting a shared `resolveRefNodeId(tabId, ref)` helper is a cheap follow-up that removes the "keep two comments honest" burden. Follow-up, not a reason to re-round.
- **NIT** — `input.ts:45`, `scroll.ts:74`: the `cdp<DocumentResult>` result is discarded; `await cdp(tabId, "DOM.getDocument", { depth: 0 })` (untyped) would state the intent — called for its side effect — more honestly. Optionally extend the comment with the binding-discard caveat above so a future refactor doesn't cache nodeIds across a `getDocument`.

**On testing:** no pure test can encode this — it's a wire-order contract against Chrome's DOM agent, and the extension's unit tests don't drive a real debugger session. The live reproduction (ref click → -32000 before, working after, in a paired Chrome) is the right evidence here; absence of a unit test is acceptable.

**Verdict: clean.** No BLOCKER, no MAJOR — this round is TERMINAL. The MINOR and NIT above are follow-ups, not grounds for another round.

damianvtran:

| Finding | Status |
|---|---|
| MINOR (ref-resolution duplicated in input.ts/scroll.ts) | **Deferred** — a shared `resolveRefNodeId` helper is the right shape, but PR #321 (multi-tab surfaces, in review) rewrites both call sites for per-surface refs; extracting the helper now would guarantee a conflicting rebase. Tracked to fold into the multi-tab rebase or as an immediate follow-up after it lands. |
| NIT (discarded `cdp<DocumentResult>` typing overstates the call) | **Deferred with the same rationale** — cosmetic, lands with the helper extraction. |

Round stays terminal (clean; both findings are follow-ups by the reviewer's own classification).

PR #325, opened 2026-08-26
fix(roster): bound sidecar write churn and heal legacy transcript bloat

damianvtran:
## Summary

Two bounded-growth hardening fixes for subagent roster persistence, on top of #308. #308 stopped NEW transcript bloat (the roster moved to an atomically replaced sidecar with 500-char per-record caps), but two gaps remained: redundant sidecar fsync churn on every roster event, and no way to heal a transcript already bloated by a pre-#308 runtime.

The motivating incident: a long delegating session accumulated a **125 MB** `transcript.jsonl` across only ~1,199 records (247 superseded `subagent_roster` snapshots, ~787 KB each) and was killed and manually restarted ~3 times by OS write-throttle/resource pressure. #308 prevents that file from ever growing again for new sessions; this PR (a) cuts the sidecar fsync volume that write-throttle pressure is measured on, and (b) heals the already-bloated files #308 leaves behind.

- **GAP 1 — skip byte-identical roster sidecar writes** (`perf(session)`). The roster persist hook fires on every job-row mutation (usage/heartbeat churn included), but most events leave the durable projection byte-identical. Each write is a full `mkstemp` + `fsync` + `os.replace` + directory-`fsync` cycle. A fingerprint guard (mirroring the existing `_maybe_persist_todos` guard) fingerprints the content `{version, jobs, records}` — **excluding** the ever-incrementing `generation` counter, which would otherwise defeat the guard — and writes only when it actually moved. The `_subagent_roster_written_generation` watermark still advances on a skipped write, so teardown's `_await_subagent_roster_writer` never loops chasing a generation the guard will always skip.

- **GAP 2 — heal legacy roster bloat on compaction** (`fix(transcript)`). `compact_file` previously folded only the message prune journal; superseded `subagent_roster` custom entries (read exclusively through `latest_custom`, newest-wins) were never reclaimed, so a pre-#308 transcript stayed permanently bloated — loaded whole into memory on every resume and re-serialized wholesale on every compaction. Compaction now also drops superseded collapsible custom entries, keeping only the newest. This is an **allowlist** (`_COLLAPSIBLE_CUSTOM_TYPES = {subagent_roster}`), deliberately excluding iteration-read types like `hub_communication` whose older entries carry real history. It runs even with no pending prunes, since a legacy bloated transcript never journals a prune.

## Change classification

C2 standard reliability/performance fix, layered on the #308 durability work. No user-visible UI, copy, layout, or interaction change.

## Delivery contract

- Resume basis intact: each record's `job_id → session_dir` mapping and outcome are untouched; only the redundant-write path and legacy-entry reclamation change.
- `latest_custom` and `build_llm_history` produce identical results before and after a heal (the kept roster entry is exactly the one `latest_custom` returns; messages, compaction markers and the prune boundary are byte-identical).
- The `reclaimable_bytes() == compact_file()` equality invariant holds: dropped collapsible customs are priced in both.
- No second capping mechanism: `comms.snapshot()` and #308's 500-char `_compact_subagent_record` cap are untouched.

## Runtime evidence

Standalone repro (`scripts/roster_bloat_repro.py`) exercises the real `SubagentComms` + `Transcript` + `Session` persist path.

**GAP 1** — 80 children × 30 KB `result_text`, ~3 roster events per settle:

```text
sidecar writes:  240  ->  80     (160 byte-identical fsync cycles eliminated)
watermark advances on skipped writes (no teardown drain loop): OK
resume round-trip (snapshot -> restore -> roster lists all children resumable): OK
```

**GAP 2** — synthesized legacy-bloated transcript, 50 superseded `subagent_roster` entries + 1 real message, one compaction pass:

```text
transcript bytes:   109,504  ->  2,342     (reclaimed 107,162)
roster entries:          50  ->  1         (survivor = newest, generation 49)
message "keep me":  preserved
latest_custom / build_llm_history:  unchanged
reclaimable_bytes() == reclaimed:  True
```

## Validation

Environment note: this host's `ulimit -n` is 256 and the repo forces `-n auto` (xdist); concurrent/contended runs exhaust file descriptors and wedge xdist teardown. The authoritative run below is serial (`-n0`) with a raised fd limit; CI runs the pinned config.

- Focused serial run across the directly-affected modules — `test_transcript_footprint`, `test_resume_subagents_todos`, `test_comms`, `test_usage_continuity`, `test_comms_persistence` → **152 passed**.
- New tests:
  - GAP 1: `test_redundant_roster_events_skip_the_sidecar_write`, `test_compact_record_caps_oversized_text_but_keeps_short_and_fields`, `test_capped_snapshot_round_trips_through_restore`
  - GAP 2: `test_compact_file_heals_legacy_roster_bloat`, `test_compact_file_keeps_a_single_roster_entry`
- Existing `test_compact_file_folds_journal_and_shrinks_disk` (the equality-invariant guard) still green.
- `flake8 .` clean, `black==26.1.0 --check .` clean, `isort==5.13.2 --check .` clean, full `pyright` → 0 errors.
- Full serial unit suite result appended to this PR once the run completes.

## Agent review

Agent code-review round to be posted below; this PR is the system of record.

## Rebase note

Rebased onto current `main` (`d4ba139e`, v0.41.4) after main advanced during review; version re-bumped to **0.41.5**. The only conflict was the `pyproject.toml` version line. Re-validated on the rebased head: affected-module suites pass serially (276 passed; one timing-sensitive subagent-responsiveness test flaked under host load and passes in isolation on both main and this branch).

### DIFF
--- a/local_operator/session/session.py
+++ b/local_operator/session/session.py
@@ -997,6 +997,35 @@ def _write_roster_sidecar(path: Any, payload: dict[str, Any]) -> None:
             temp.unlink()
 
 
+def _write_roster_sidecar_if_changed(
+    path: Any, payload: dict[str, Any], previous_fingerprint: str | None
+) -> tuple[str, bool]:
+    """Compute the roster fingerprint and write the sidecar only if it moved.
+
+    Runs ENTIRELY on the worker thread (dispatched via ``asyncio.to_thread``):
+    the fingerprint is an O(roster) ``json.dumps`` and #308's invariant is that
+    everything that scales leaves the event loop. Computing it on the loop made
+    a large roster pay that serialization on every roster event — the exact
+    regression ``TestMeasurementCosts`` catches. Returns ``(fingerprint, wrote)``.
+
+    The fingerprint deliberately EXCLUDES ``generation``: the counter is bumped
+    on every roster event by design (it drives the coalescing loop), so
+    including it would make every payload unique and defeat the guard. Only the
+    durable projection a resume actually reads is compared. ``sort_keys`` keeps
+    the serialization order-stable so an unchanged roster hashes identically
+    across writes.
+    """
+    fingerprint = json.dumps(
+        {key: payload[key] for key in ("version", "jobs", "records")},
+        sort_keys=True,
+        separators=(",", ":"),
+    )
+    if fingerprint == previous_fingerprint:
+        return fingerprint, False
+    _write_roster_sidecar(path, payload)
+    return fingerprint, True
+
+
 def _read_roster_sidecar(path: Any) -> dict[str, Any] | None:
     """Read a supported sidecar, falling back to the legacy transcript on error."""
     try:
@@ -1466,7 +1495,10 @@ class Session:
         # limit" throttle on a long delegating session. Mirrors the todo guard
         # (:meth:`_maybe_persist_todos`): fingerprint the durable content and
         # write only when it actually moved. ``None`` means nothing persisted
-        # yet, so the first event always writes.
+        # yet, so the first event always writes. The fingerprint itself is
+        # computed ON THE WORKER THREAD by
+        # :func:`_write_roster_sidecar_if_changed` — it is an O(roster)
+        # ``json.dumps``, which must not run on the loop.
         self._persisted_roster_fingerprint: str | None = None
         # Session-scoped task group (HC-11): wake deliveries and aside
         # persistence are routed through it so dispose() cancels them
@@ -5950,28 +5982,18 @@ class Session:
                     "jobs": rows,
                     "records": compact_records,
                 }
-                # Fingerprint the CONTENT, not the counter. ``generation`` is
-                # bumped on every roster event by design (it drives the
-                # coalescing loop), so including it here would make every payload
-                # unique and defeat the guard. Everything else is the durable
-                # projection a resume actually reads. ``sort_keys`` makes the
-                # serialization order-stable so an unchanged roster hashes
-                # identically across writes.
-                fingerprint = json.dumps(
-                    {
-                        "version": _SUBAGENT_ROSTER_VERSION,
-                        "jobs": rows,
-                        "records": compact_records,
-                    },
-                    sort_keys=True,
-                    separators=(",", ":"),
+                # The O(roster) fingerprint computation rides the SAME worker
+                # hop as the write: #308's invariant is that everything that
+                # scales leaves the loop, and an on-loop ``json.dumps`` of a
+                # large roster is the exact regression TestMeasurementCosts
+                # catches (it fires on every roster event, teardown included).
+                fingerprint, wrote = await asyncio.to_thread(
+                    _write_roster_sidecar_if_changed,
+                    self._transcript.directory / SUBAGENT_ROSTER_SIDECAR,
+                    payload,
+                    self._persisted_roster_fingerprint,
                 )
-                if fingerprint != self._persisted_roster_fingerprint:
-                    await asyncio.to_thread(
-                        _write_roster_sidecar,
-                        self._transcript.directory / SUBAGENT_ROSTER_SIDECAR,
-                        payload,
-                    )
+                if wrote:
                     # Keep one bounded legacy entry for readers predating
                     # v0.35.2; future mutations replace only the sidecar,
                     # avoiding quadratic history while preserving a
--- a/local_operator/session/session.py
+++ b/local_operator/session/session.py
@@ -894,11 +894,12 @@ def _pruned_ids(messages: Sequence[AgentMessage]) -> set[str]:
 
 
 #: The ONLY ``AsyncJob`` fields the roster snapshot carries. An allowlist, not
-#: an exclude set, because the whole task-row list is re-appended on every
-#: roster move and custom entries are never reclaimed (``compact_file`` folds
-#: only the message prune journal) — so anything unbounded here is written once
-#: per surviving child per move, i.e. O(children²) bytes of superseded data on
-#: disk. The fields kept are all small and bounded: the identity, the timings
+#: an exclude set, because the roster projection must stay small: it is written
+#: to the replaced sidecar on every roster move, and a superseded copy in a
+#: legacy transcript is only reclaimed lazily (``compact_file`` now collapses
+#: superseded ``subagent_roster`` customs, but not until the next compaction) —
+#: so anything unbounded here bloats both surfaces. The fields kept are all
+#: small and bounded: the identity, the timings
 #: the panel prices elapsed from, the model/usage/window it paints, and the
 #: routing/queue flags a restore needs. Everything a reader might want beyond
 #: this — the child's full ``result_text``, its verbatim ``prompt`` (documented
@@ -1454,6 +1455,19 @@ class Session:
         self._subagent_roster_generation = 0
         self._subagent_roster_written_generation = 0
         self._subagent_roster_writer: asyncio.Task[None] | None = None
+        # The CONTENT of the roster payload last written to the sidecar, so a
+        # redundant write can be skipped. The roster persist hook fires on every
+        # job-row mutation — usage/heartbeat churn included — but most of those
+        # events leave the persisted projection ({version, jobs, records}, minus
+        # the ever-incrementing ``generation`` counter) byte-identical. A real
+        # fan-out fired ~3 sidecar writes per settle, ~2 of them identical, and
+        # each is a full mkstemp + fsync + os.replace + directory-fsync cycle;
+        # that fsync volume is what tripped the macOS "disk writes exceeding
+        # limit" throttle on a long delegating session. Mirrors the todo guard
+        # (:meth:`_maybe_persist_todos`): fingerprint the durable content and
+        # write only when it actually moved. ``None`` means nothing persisted
+        # yet, so the first event always writes.
+        self._persisted_roster_fingerprint: str | None = None
         # Session-scoped task group (HC-11): wake deliveries and aside
         # persistence are routed through it so dispose() cancels them
         # deterministically and a delivery after dispose never raises into an
@@ -5929,27 +5943,52 @@ class Session:
                 records = (
                     self._subagent_comms.snapshot() if self._subagent_comms is not None else []
                 )
+                compact_records = [_compact_subagent_record(record) for record in records]
                 payload = {
                     "version": _SUBAGENT_ROSTER_VERSION,
                     "generation": generation,
                     "jobs": rows,
-                    "records": [_compact_subagent_record(record) for record in records],
+                    "records": compact_records,
                 }
-                await asyncio.to_thread(
-                    _write_roster_sidecar,
-                    self._transcript.directory / SUBAGENT_ROSTER_SIDECAR,
-                    payload,
+                # Fingerprint the CONTENT, not the counter. ``generation`` is
+                # bumped on every roster event by design (it drives the
+                # coalescing loop), so including it here would make every payload
+                # unique and defeat the guard. Everything else is the durable
+                # projection a resume actually reads. ``sort_keys`` makes the
+                # serialization order-stable so an unchanged roster hashes
+                # identically across writes.
+                fingerprint = json.dumps(
+                    {
+                        "version": _SUBAGENT_ROSTER_VERSION,
+                        "jobs": rows,
+                        "records": compact_records,
+                    },
+                    sort_keys=True,
+                    separators=(",", ":"),
                 )
-                # Keep one bounded legacy entry for readers predating v0.35.2;
-                # future mutations replace only the sidecar, avoiding quadratic
-                # history while preserving a rolling-upgrade resume path.
-                if (rows or records) and self._transcript.latest_custom(
-                    SUBAGENT_ROSTER_CUSTOM_TYPE
-                ) is None:
-                    await self._transcript.append_custom(
-                        SUBAGENT_ROSTER_CUSTOM_TYPE,
-                        {"jobs": rows, "records": payload["records"]},
+                if fingerprint != self._persisted_roster_fingerprint:
+                    await asyncio.to_thread(
+                        _write_roster_sidecar,
+                        self._transcript.directory / SUBAGENT_ROSTER_SIDECAR,
+                        payload,
                     )
+                    # Keep one bounded legacy entry for readers predating
+                    # v0.35.2; future mutations replace only the sidecar,
+                    # avoiding quadratic history while preserving a
+                    # rolling-upgrade resume path.
+                    if (rows or records) and self._transcript.latest_custom(
+                        SUBAGENT_ROSTER_CUSTOM_TYPE
+                    ) is None:
+                        await self._transcript.append_custom(
+                            SUBAGENT_ROSTER_CUSTOM_TYPE,
+                            {"jobs": rows, "records": payload["records"]},
+                        )
+                    self._persisted_roster_fingerprint = fingerprint
+                # Advance the written-generation watermark whether or not this
+                # payload was actually flushed: the CONTENT is up to date on
+                # disk (identical bytes already there), and leaving the watermark
+                # behind would make ``_await_subagent_roster_writer`` at teardown
+                # loop forever chasing a generation the guard will always skip.
                 self._subagent_roster_written_generation = generation
         except Exception:  # noqa: BLE001 - persistence must never break a turn
             logger.warning("could not persist subagent roster", exc_info=True)
--- a/local_operator/session/transcript.py
+++ b/local_operator/session/transcript.py
@@ -80,6 +80,27 @@ COMPACT_FILE_THRESHOLD_BYTES = 256 * 1024
 #: Lives under ``provider_payload`` so an older build simply ignores it.
 SHRUNK_KEY = "context_shrunk_here"
 
+#: Custom-entry types whose SUPERSEDED copies :meth:`Transcript.compact_file`
+#: drops on disk, keeping only the newest. These are the NEWEST-WINS types: the
+#: session reads them exclusively through :meth:`latest_custom`, so every older
+#: entry is dead weight the moment a newer one lands.
+#:
+#: ``subagent_roster`` is the reason this exists. Before v0.40.0 the roster
+#: re-appended a full snapshot to the transcript on every roster move, and a
+#: real fan-out left ~247 giant superseded roster entries in one file — a 125 MB
+#: transcript that loads whole into memory on every resume and is re-serialized
+#: wholesale on every compaction. v0.40.0 stopped writing NEW bloat (the roster
+#: moved to a replaced sidecar) but never healed transcripts already bloated;
+#: collapsing here is that heal, so an upgraded session sheds the dead entries on
+#: its next compaction instead of carrying them forever.
+#:
+#: This is an ALLOWLIST, not "every custom type", because it is NOT safe to
+#: collapse a type that accumulates. ``hub_communication`` is read by ITERATION
+#: (``subagent_view`` folds the whole parent/child message log), so dropping its
+#: older entries would erase history; it is deliberately absent. Add a type here
+#: only after confirming nothing reads it by iteration.
+_COLLAPSIBLE_CUSTOM_TYPES = frozenset({"subagent_roster"})
+
 
 @dataclass
 class TranscriptEntry:
@@ -752,6 +773,32 @@ class Transcript:
 
     # -- lifecycle ----------------------------------------------------------
 
+    def _superseded_custom_indices(self) -> set[int]:
+        """Indices of collapsible custom entries that a newer copy supersedes.
+
+        For each type in :data:`_COLLAPSIBLE_CUSTOM_TYPES` the NEWEST entry is
+        kept and every older one is returned for dropping. These types are read
+        only through :meth:`latest_custom`, so an older copy is pure dead weight
+        — the pre-v0.40.0 roster bloat this heals is exactly a long run of them
+        (see the constant's note). Keeping the newest is load-bearing: it is the
+        one ``latest_custom`` returns and a resume restores from.
+        """
+        newest_by_type: dict[str, int] = {}
+        for index, entry in enumerate(self._entries):
+            if entry.type != ENTRY_CUSTOM:
+                continue
+            custom_type = entry.payload.get("custom_type")
+            if custom_type in _COLLAPSIBLE_CUSTOM_TYPES:
+                newest_by_type[custom_type] = index
+        keep = set(newest_by_type.values())
+        return {
+            index
+            for index, entry in enumerate(self._entries)
+            if entry.type == ENTRY_CUSTOM
+            and entry.payload.get("custom_type") in _COLLAPSIBLE_CUSTOM_TYPES
+            and index not in keep
+        }
+
     def reclaimable_bytes(self) -> int:
         """Bytes :meth:`compact_file` would free right now.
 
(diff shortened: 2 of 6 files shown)


damianvtran:
### Agent review — round 1
Scope: f6e7d027..966eb9aa

Independent review — I did not write this code. I read both commit diffs (`1a1395e0` GAP1, `0100cac2` GAP2), the full surrounding source for the touched methods, traced every reader of the affected custom types, and ran targeted execution to reproduce the risky cases rather than infer them.

## Verdict up front: CLEAN — no blocker, no major.

The two changes are correct and the invariants the design leans on hold under test. Details and the traces that make the clean verdict credible are below, followed by one MINOR and one NIT.

## What I verified (not just read)

**GAP 1 — fingerprint guard (`session.py` `_persist_subagent_roster`)**
- **Fingerprint content-correctness.** `{version, jobs, records}` with `sort_keys=True, separators=(",",":")` — `generation` correctly excluded (it is the monotonic coalescing counter, bumped every event, and is *not* resume-critical: on restore `generation` is re-read from the sidecar payload, which still carries it). Two genuinely-different rosters can only collide if their `{version,jobs,records}` serialize identically, i.e. they are identical in everything a resume reads. No false-skip window.
- **Watermark-on-skip / lost-update.** Traced `_persist_subagent_roster` → `_await_subagent_roster_writer` → `_final_persist_snapshots` (session.py:5992, 5999-6010, 6012-6020). The watermark advances on a skipped write, so teardown's drain never spins; and because the write decision is on *content fingerprint*, a real change at gen N+4 after identical skipped gens N+1..N+3 still writes (fingerprint differs). No lost final state at dispose. `test_redundant_roster_events_skip_the_sidecar_write` asserts exactly this (a second real child forces a write), and I re-ran it green.
- **First write / concurrency.** `_persisted_roster_fingerprint` inits `None` (session.py:1470) so the first event always writes; it is read/written only inside the single-writer loop (5969/5986), no races. Empty-roster path unchanged: the `if (rows or records)` gate covers only the legacy custom append, not the sidecar write.
- Ran `scripts/roster_bloat_repro.py`: 240 → 80 sidecar writes, resume round-trip OK.

**GAP 2 — compaction heal (`transcript.py`)**
- **`_superseded_custom_indices` keeps exactly the newest per collapsible type.** Verified single-entry no-op (`test_compact_file_keeps_a_single_roster_entry`) and 50→1 collapse keeping generation 49.
- **Allowlist safety — grepped every reader.** `subagent_roster` is read only via `latest_custom` (session.py:6048, mobile/daemon.py:282), never by iteration. `hub_communication` is the iteration-read type (subagent_view.py:293) and is correctly absent. Other newest-wins customs (`todo_snapshot`, `selected_model`, `active_route`, `conversation_name`, `wake_schedules`) are also `latest_custom`-only and are *acceptably* left uncollapsed — they don't re-append giant tails, so no silent bloat. Noted, not a defect.
- **Boundary interaction (the flagged high-risk case) — reproduced.** Built a transcript with roster customs interleaved *between* a prune target and the newest prune, plus a usage-bearing message in the gap. Result: `reclaimable_bytes() == compact_file()` (4802==4802), `build_llm_history()` byte-identical pre/post, `usages_since_compaction()` identical pre/post, only the newest roster survives. The boundary is safe because dropped customs are `continue`d *before* the `boundary = len(folded)-1` update, so the mark still lands on the last *surviving* entry ≤ `newest_prune`, and dropped customs never carry `usage`, so the window is unchanged.
- **Byte-accounting symmetry.** `reclaimable_bytes` prices each dropped custom as `len(entry.to_json())+1` — exactly what `_render_folded` emits (`to_json()` + `"\n"`). `to_json` uses default `ensure_ascii=True`, so char-length equals utf-8 byte-length and the equality holds even with unicode child text. Existing `test_compact_file_folds_journal_and_shrinks_disk` equality guard still green.
- **Crash safety unchanged:** `compact_file` still routes through `_replace_file`'s temp-file + `os.replace`.
- No regression to resume, the subagent panel, mobile projection, or hub peek — all read the roster newest-wins, and the newest is exactly what is kept.

## m1 (MINOR) — PR body presents GAP 2 runtime evidence the committed repro cannot produce
`scripts/roster_bloat_repro.py:192-208` (`main`) prints only the GAP 1 block. The PR "Runtime evidence" section presents a GAP 2 block (`transcript bytes: 109,510 -> 2,343`, `roster entries: 50 -> 1`) under the same "Standalone repro (`scripts/roster_bloat_repro.py`)" heading, implying the script produces it. It does not — running the committed script emits only GAP 1. The GAP 2 behavior *is* genuinely covered by real execution in `test_compact_file_heals_legacy_roster_bloat` (I ran it), so this is an evidence-provenance/reproducibility gap in the PR narrative, not a code defect. Fix: either add a GAP 2 pass to the repro's `main`, or relabel the GAP 2 numbers as coming from the unit test.

## N1 (NIT) — first roster event after a resume always rewrites the sidecar
On restore, `_subagent_roster_written_generation` is set (session.py:6052) but `_persisted_roster_fingerprint` stays `None`, so the first real roster event of a resumed session writes even when the content equals the on-disk sidecar. One redundant write per resume — harmless and arguably a safe re-establish; noting only for completeness. Priming the fingerprint from the loaded sidecar bytes at restore would eliminate it, but is not worth the coupling.

## Verdict
**CLEAN.** No blocker, no major. This round is TERMINAL — m1 and N1 are follow-ups, not grounds for another round. The change is safe to merge once m1 is answered (a one-line remediation noting the evidence source is sufficient; no code change required).

damianvtran:

Responding to the round-1 agent review (verdict: clean, 0 blockers / 0 majors). Both findings addressed below.

- **m1 (MINOR) — evidence provenance, FIXED in `a6bf17fb`.** The finding is correct: the PR body attributed the GAP 2 numbers to `scripts/roster_bloat_repro.py`, but the committed script only printed the GAP 1 write-count. Rather than relabel the narrative, the repro script now genuinely produces the GAP 2 figures: it drives the real `Transcript`, appends a real message plus 50 superseded `subagent_roster` custom entries, runs `compact_file`, and prints the before/after transcript bytes, the roster-entry count, that the message survived replay, that `latest_custom` is unchanged, and that `reclaimable_bytes() == reclaimed`. Its authoritative output is now:

  ```text
  === GAP 2: heal legacy transcript bloat on compaction ===
  roster custom entries          : 50 -> 1
  transcript size (bytes)        : 109,504 -> 2,342 (reclaimed 107,162)
  surviving roster is newest     : generation 49
  latest_custom unchanged        : generation 49
  message survived replay        : ['keep me']
  reclaimable == reclaimed        : True
  ```

  The PR body's GAP 2 block has been updated to these exact numbers so the narrative and the script agree. (The earlier body figures 109,510 → 2,343 were an ad-hoc measurement; the script is now the source of truth.) Gates re-run on the script: black, isort, flake8, pyright all clean.

- **N1 (NIT) — accepted, no code change.** The first roster event after a resume rewrites the sidecar because `_persisted_roster_fingerprint` is not primed from the loaded sidecar bytes at restore. This costs exactly one extra sidecar write per resume, which is negligible against the 160-write reduction the guard delivers within a session, and priming it would mean re-serializing the loaded roster on the load path purely to avoid one write. Left as-is deliberately.

No blocker/major findings were raised, so no further review round is required. Reviewed scope `f6e7d027..966eb9aa` is extended by the docs-only remediation commit `a6bf17fb` (repro script + PR-body numbers; no `local_operator/` change), so the round remains valid against the code under review.

damianvtran:
### Agent review — round 1 (freshness confirmation)
Scope: f6e7d027..a6bf17fb

Re-verified against the current head after the round-1 remediation.

**The only delta since my reviewed head (966eb9aa) is evidence tooling — zero `local_operator/` change.** `git diff --stat 966eb9aa a6bf17fb` is a single file, `scripts/roster_bloat_repro.py` (+92), and `git diff --name-only 966eb9aa a6bf17fb` lists nothing under `local_operator/`. The change is the m1 remediation: the repro's `main` now runs a `_heal_legacy_transcript()` pass that drives the REAL `Transcript` + `compact_file` and prints the GAP 2 heal numbers, so the PR's "Runtime evidence" is now produced by the script rather than a relabeled test assertion.

Ran the updated script to confirm it does what it claims:
```
=== GAP 2: heal legacy transcript bloat on compaction ===
roster custom entries          : 50 -> 1
transcript size (bytes)        : 109,502 -> 2,344 (reclaimed 107,158)
surviving roster is newest     : generation 49
latest_custom unchanged        : generation 49
message survived replay        : ['keep me']
reclaimable == reclaimed        : True
```

No production code moved, so every round-1 trace — the fingerprint content-correctness, the watermark-on-skip / no-lost-update analysis, the allowlist reader grep, the reproduced prune-plus-superseded-custom boundary interaction (`reclaimable_bytes() == compact_file()`, replay and usage window byte-identical), and the byte-accounting symmetry — stands unchanged.

**Verdict: CLEAN stands against a6bf17fb — 0 blocker, 0 major. Round is TERMINAL and fresh against the current head.** m1 is answered; N1 remains a non-blocking follow-up.

damianvtran:
### Agent review — round 2
Scope: a6bf17fb..ecc709a7

Independent review of the regression fix commit `ecc709a7` on its own merits, plus a re-check of round 1's analysis against the new head. I read the diff first, then verified by execution: the invariant test, the GAP 1/GAP 2 suites, the adjacent comms/footprint suites, the repro script, the failure paths driven by hand, and the loop-side cost measured rather than inferred. Imports confirmed resolving to the worktree (`/private/tmp/lop-roster-bloat/local_operator`).

## Verdict up front: CLEAN — no blocker, no major.

The fix is correct and minimal: it moves exactly the term that regressed (the O(roster) fingerprint `json.dumps`) onto the worker thread the write already used, and changes nothing else about the guard's semantics. One MINOR and one NIT below, both non-blocking; neither is introduced by `ecc709a7`.

## The regression is actually fixed (verified, not inferred)

- `TestMeasurementCosts` (all 3 tests, incl. `test_everything_that_scales_leaves_the_event_loop`) — **3 passed** serially on `ecc709a7`. The test's `dumps` spy patches `session.json.dumps` module-wide; the only remaining `json.dumps` on the changed path now lives inside `_write_roster_sidecar_if_changed` (session.py:1000-1026), dispatched via `asyncio.to_thread` at session.py:5990-5995 — the same hop the write always took. Grep of the changed paths confirms no other `session.json.dumps` call site on the roster path.
- `tests/unit/session/test_transcript_footprint.py tests/unit/session/test_resume_subagents_todos.py` (GAP 1 + GAP 2 suites) — **30 passed**.
- `tests/unit/harness/test_comms_persistence.py tests/unit/session/test_session_footprint.py` — **13 passed**.
- `scripts/roster_bloat_repro.py` on the new head: **240 → 80 sidecar writes** (160 skipped), heal 109,507 → 2,345 bytes reclaiming 107,162, survivor = newest (generation 49), `reclaimable == reclaimed` True, resume round-trip OK. Commit-message claim reproduced exactly.

## Guard semantics after the move (each checked)

- **Skip path returns the unchanged fingerprint** (`return fingerprint, False`), and leaving `self._persisted_roster_fingerprint` unassigned on skip is correct: the skip condition is `fingerprint == previous_fingerprint`, so the stored value already equals it. No divergence.
- **Watermark advances unconditionally** at session.py:6014, outside the `if wrote:` block — teardown's `_await_subagent_roster_writer` still cannot spin on a generation the guard will always skip. Repro's "watermark advances on skipped writes" check green.
- **Failure path — driven by hand, not read:** with `_write_roster_sidecar` monkeypatched to raise `OSError("disk full")`, the exception propagates out of `asyncio.to_thread`, is caught by the outer `except Exception` (session.py:6015, `persistence must never break a turn`), and logs the warning. `self._persisted_roster_fingerprint` stays unset (assignment at :6008 is inside `if wrote:`), so a retry re-attempts the write — confirmed by a follow-up call with the real writer, which wrote. Also confirmed the skip-eligible call makes **zero** write attempts when the writer is broken (fingerprint compare short-circuits before `_write_roster_sidecar`). Same containment shape as base for a failing write; no new hazard.
- **Types:** helper is `(Any, dict[str, Any], str | None) -> tuple[str, bool]`; pyright on the file — 0 errors, 0 warnings. flake8, black==26.1.0, isort==5.13.2 (venv, repo config) all clean on the changed file. CI lint job already green on `ecc709a7`; test/type-check jobs were still queued at review time (run 32997532005).
- **Comments:** the helper's docstring states the WHY (the #308 invariant, why `generation` is excluded, why `sort_keys`) and the call site explains the thread hop — consistent with repo convention.

## Round 1 status against the new head

- **Round-1 m1 (repro did not emit GAP 2 numbers) — RESOLVED** by `a6bf17fb`: the repro now prints the full GAP 2 heal block, and my run reproduces the PR body's numbers.
- **Round-1 N1 (first roster event after resume always rewrites)** — unchanged, still a harmless follow-up. The fix commit does not touch restore.
- Round 1's core analysis (fingerprint content-correctness, no in-process lost-write window, allowlist safety, boundary interaction) is unaffected by `ecc709a7`: the move changes *where* the fingerprint is computed, not *what* it compares or when the write happens.

## Findings

**m2 (MINOR) — the fingerprint guard removed the sidecar's self-heal on external deletion; the fallback can then serve a stale roster.** `session.py:5990-6008` with `_read_roster_sidecar` (:1030) / `_load_subagent_roster`. Pre-#308 there was no sidecar; between #308 and this PR every roster event rewrote it, so a sidecar file deleted mid-session (user tidy; another process's sweep that misses the live-dir guard) was recreated by the next event even with unchanged content. Now an unchanged-content event skips the write, so the file stays gone; at resume `_read_roster_sidecar` falls back to the one-time legacy `subagent_roster` custom, which froze the roster as of the *first* write — children launched after that are missing from the resume basis (their own transcripts exist; the parent's mapping to them is what's stale). Requires external deletion of exactly that file mid-session plus no content change afterwards, so it is an edge case, not ship-blocking — and it is a property of the GAP 1 guard round 1 already passed, not of `ecc709a7`. Recorded because round 1's "no lost-write window" analysis covered in-process windows only. Cheapest hardening if ever wanted: on skip, `path.exists()` (a stat, not a write) and treat a missing file as changed.

**N2 (NIT) — the loop-side roster build is still O(roster), pre-existing and small; the test does not and need not catch it.** `_subagent_job_row` (`model_dump(mode="json")`), `comms.snapshot()` (dict construction), and `_compact_subagent_record` all run on the loop before the hop (session.py:5970-5978). Measured on this host for the repro's worst case (80 children, 30 KB texts): **0.29 ms/event** of loop time, **0** `json.dumps` calls (pydantic dump + dict copies, not JSON serialization; row fields are bounded by the `_ROSTER_ROW_FIELDS` allowlist and text by the 500-char cap). Base `f6e7d027` did the identical build on the loop, so the fix commit introduces nothing here — and it removed the dominant term (the on-loop fingerprint `json.dumps` measured **0.51 ms/event** for the same roster). The invariant test spies only `json.dumps`/`approx_text_tokens`, so it would not catch this class — acceptable, since the class is dict-building with a small constant, not serialization; noting so the next "everything that scales" audit starts from measured numbers.

## Verdict

**CLEAN — no blocker, no major.** This round is TERMINAL: m2 and N2 are follow-ups, not grounds for another round. `ecc709a7` fully remediates the CI regression (round-2 scope), round 1's open items are resolved or unchanged-by-design, and the PR is safe to merge once CI run 32997532005 confirms green on `ecc709a7`.

damianvtran:

Responding to the round-2 agent review (verdict: clean, 0 blockers / 0 majors; findings m2 + N2, both pre-existing and non-blocking).

- **m2 (MINOR) — accepted as a follow-up, no code change in this PR.** With the fingerprint guard, a roster event whose content is unchanged skips the sidecar write, so an externally deleted sidecar is no longer rewritten (and self-healed) by the next no-op event; a resume in that window falls back to the frozen legacy custom entry from the first write. This is an edge case (requires out-of-band deletion mid-session) and the legacy fallback is exactly the path #308 kept alive for pre-v0.35.2 readers, so no data is lost — the roster is merely stale until the next real roster change. The reviewer's suggested hardening (a `path.exists()` stat on the skip path) is noted for a follow-up; adding it here would widen a PR that has already been through two review rounds for a case no user has hit. Not introduced by `ecc709a7` in spirit: the base rewrote the sidecar on every event, which masked the deletion by accident rather than by design.

- **N2 (NIT) — accepted, no change.** The loop-side projection build (`_subagent_job_row`'s `model_dump`, `comms.snapshot()`'s dict copies, `_compact_subagent_record`) remains O(roster) but is bounded by the field allowlist, measured at 0.29 ms per event with zero `json.dumps` calls, and is byte-for-byte identical to the base's behaviour. The fix commit removed the dominant term (the 0.51 ms on-loop fingerprint serialization). Leaving the remainder is the status quo, not a regression.

No blocker/major findings were raised. The reviewed scope `a6bf17fb..ecc709a7` is the current head; round 1's clean verdict was independently re-confirmed against the new head in the round-2 comment. The gate is complete pending CI green on run 32997532005.

damianvtran:
### Agent review — round 3 (rebase integration)
Scope: d4ba139e..4bddcd50

Focused rebase-integration review, not a re-litigation of rounds 1–2 (their fingerprint-guard and compaction-heal analysis was verified by execution and stands). The branch's merge-base is `d4ba139e` and the diff against it is exactly the rounds-1&2-approved content — no conflict markers, no duplicated blocks, no dropped hunks, no foreign changes. `session.py`'s `_persist_subagent_roster`/`_write_roster_sidecar_if_changed` and `transcript.py`'s `_COLLAPSIBLE_CUSTOM_TYPES`/`_superseded_custom_indices` survived the rebase intact.

**Findings: none (no blocker, no major, no minor, no nit).**

### Verified by reading the diff AND executing

**Rebase cleanliness**
- `git diff d4ba139e..4bddcd50 -- local_operator/ scripts/ tests/ pyproject.toml`: 6 files, +654/-24. No `<<<`/`===`/`>>>` markers. Only non-approved textual change is the `pyproject.toml` version line (0.41.4 → 0.41.5), the documented single conflict.
- #327's touch to `tests/unit/session/test_launch_subagent.py` and #323's touches are NOT re-touched by this branch — the branch cleanly sits on top of main's versions.

**Semantic interaction with main's new code (the high-value part)**
- **Roster-custom readers are all newest-wins-safe.** Every reader of `SUBAGENT_ROSTER_CUSTOM_TYPE` in the current tree goes through `latest_custom` (backward scan, first hit): `session.py:6070` (`_restore_subagent_roster`) and `mobile/daemon.py:282` (`_durable_projection`). No reader iterates roster customs, so the compaction heal dropping superseded copies cannot lose data.
- **#323 (`4cdc9604`, merged to origin AFTER this branch's base) does not add an iteration reader.** Although #323 reworks exactly the mobile/subagent-transcript area, its roster read is the same `latest_custom` path; it moves subagent *transcripts* off the wire and fetches them lazily, and touches no roster-custom iteration. Once this branch merges the two are compatible. #318/#309 add no roster-custom iteration reader either.
- **Fingerprint guard is safe against any new `AsyncJob` fields.** `_subagent_job_row` projects via `model_dump(include=_ROSTER_ROW_FIELDS)` — an allowlist, so any field main adds to `AsyncJob` is excluded by construction and cannot make the fingerprint churn. (`local_operator/harness/jobs.py` had no changes in `d4ba139e..origin/main` anyway.)
- **`SubagentComms.snapshot()` shape round-trips.** `comms.py` had no changes in the rebase window (`d4ba139e..origin/main`); snapshot → `_compact_subagent_record` → restore is unchanged from rounds 1–2.

**Execution (serial `-n0`, isolated where timing-sensitive)**
- `test_session.py::TestMeasurementCosts`, `test_transcript_footprint.py`, `test_resume_subagents_todos.py`, `test_comms.py`: **137 passed**.
- `test_comms_persistence.py`: **11 passed**.
- `scripts/roster_bloat_repro.py`: prints **240 → 80** (160 redundant sidecar writes skipped) and the GAP 2 heal block (**50 → 1** roster customs, 109,504 → 2,342 bytes, `reclaimable == reclaimed: True`, `latest_custom` unchanged at generation 49). Resume round-trip OK.
- `test_the_loop_stays_responsive_while_several_subagents_run` (isolated): initially FAILED at 1758 ms, but this is host-load flake, not a regression — the **base worktree on plain `origin/main` fails the same assertion** (1085 ms) under the current load, and the branch **passes on retry** (`1 passed`). Host load average was **87–90** throughout (far above the 20–44 the task noted); the assertion measures wall-clock loop stall, which tracks host contention directly. The guard reduces on-loop roster work, so no plausible regression path.
- flake8 / black --check / isort --check-only --profile black on all 5 changed files: clean.

**Metadata**
- Version is `0.41.5` in `pyproject.toml`. PR body accurately describes GAP 1 (fingerprint guard excluding `generation`, watermark still advancing on a skipped write) and GAP 2 (compaction heal of superseded roster customs). Rounds 1 and 2 (plus a round-1 freshness confirmation) are present with remediations.

### Verdict: CLEAN (no blocker, no major)
The rebase is clean and integrates correctly with main's #327/#328/#323/#318/#309. This round is TERMINAL — no remediation required. The reviewed head (`4bddcd50`) is the current PR head.

damianvtran:
### Agent review — round 3 (rebase integration)
Scope: d4ba139e..4bddcd50

Focused rebase-integration review. Rounds 1 & 2 (through `ecc709a7`, both CLEAN) verified the GAP1 fingerprint guard and GAP2 compaction heal by execution; that analysis STANDS and is not re-litigated here. This round checks (a) the rebase onto `d4ba139e` is clean and (b) main's 9 new commits (#308-adjacent `comms.py`/`subagent.py`/`jobs.py`, #327 subagent-view/child-bash polling, #318 fan-resume, #309 restore-transcripts, #328 group-reaper) do not interact badly with ours.

**Rebase cleanliness — verified.** `git diff d4ba139e..HEAD` touches exactly 6 files: `session/session.py`, `session/transcript.py`, `pyproject.toml` (version line only, `0.41.4`→`0.41.5`), `scripts/roster_bloat_repro.py`, and two test modules. No conflict markers, no duplicated blocks, no dropped hunks, no foreign changes. `_persist_subagent_roster`/`_write_roster_sidecar_if_changed` and `_COLLAPSIBLE_CUSTOM_TYPES`/`_superseded_custom_indices` are intact. One post-rounds-1&2 commit is new: `4bddcd50 fix(session): compute roster fingerprint off the event loop` — reviewed below. (Note: `origin/main` has since advanced one further commit to `4cdc9604` #323; that is outside our scope and not in the branch.)

**Semantic interaction with main's new code — verified newest-wins-safe.**
- Every reader of `SUBAGENT_ROSTER_CUSTOM_TYPE` roster customs reads via `latest_custom` (newest-wins), which GAP2's heal preserves. Confirmed: `session.py:6001` (legacy-entry guard) and `:6070`, and `mobile/daemon.py:282` (restart-safe route rebuild). The only ITERATION reader over `ENTRY_CUSTOM` in the #327-touched `tui/widgets/subagent_view.py:317,323` filters strictly on `HUB_COMMUNICATION_CUSTOM_TYPE` — deliberately excluded from `_COLLAPSIBLE_CUSTOM_TYPES` and never touched by the heal. No roster reader iterates superseded customs.
- Fingerprint guard vs #327/#328 extra churn: `_subagent_job_row` projects onto the `_ROSTER_ROW_FIELDS` **include-set** allowlist (`session.py:910`), so main's churny/new `AsyncJob` fields (`output_tail`, `output_seq`, `latest_details`, `consumed`, `child_jobs`) are auto-excluded. More job spawn/settle from child-bash polling and the group reaper therefore produces MORE no-op events the guard correctly skips, not roster growth. No `AsyncJob` field added on main belongs in the row.
- `SubagentComms.snapshot()` (unchanged by this PR; `comms.py` diff base..head is empty) emits `job_id/label/parent_job_id/prompt/effective_prompt/launch_message_id/agent_role/effort/session_dir/outcome/result_text/error_text/paused/settled_at/attempt_aliases/prior_launch_prompts`. `_compact_subagent_record` caps only `prompt`/`error_text`/`result_text` and copies all other keys through, so it round-trips through `restore()` intact.

**Off-loop fingerprint change (`4bddcd50`) — correct.** `_write_roster_sidecar_if_changed` runs the O(roster) `json.dumps` on the worker thread inside the same `asyncio.to_thread` hop as the write, upholding #308's "everything that scales leaves the loop" invariant that `TestMeasurementCosts` guards. Fingerprint excludes `generation` (correct — the counter bumps every event and would defeat the guard) and uses `sort_keys`/tight separators for stable hashing. `self._persisted_roster_fingerprint` is read on the loop before dispatch and written on the loop after; the single coalescing writer task (`session.py:5960`) guarantees no interleaving. Watermark advances on skipped writes so teardown's drain loop never chases a skipped generation.

**Executed (serial `-n0`, raised fds, host-saturation isolated):**
- `test_transcript_footprint.py` + `test_resume_subagents_todos.py` → 30 passed.
- `test_session.py::TestMeasurementCosts` (the on-loop-cost invariant) + `test_comms.py` + `test_comms_persistence.py` → 118 passed.
- `scripts/roster_bloat_repro.py` → GAP1 `240 → 80` sidecar writes (160 redundant fsync cycles skipped); GAP2 `50 → 1` roster customs, `109,514 → 2,343` bytes (reclaimed 107,171), survivor = newest (generation 49), `latest_custom` unchanged, message replay preserved, `reclaimable == reclaimed` True, snapshot→restore round-trip OK.
- CI run `33005664457` on head `4bddcd50` → success.
- Version `0.41.5` in pyproject; PR body matches reality (rebase note, evidence, classification).

**Findings:** none. No BLOCKER, no MAJOR, no MINOR, no NIT.

**Verdict: CLEAN.** The rebase is clean and main's new code is newest-wins-safe against the roster heal and fingerprint guard. This round is TERMINAL — no remediation required.

PR #128, opened 2026-08-17
feat(harness): reach omp parity across tools, context, MCP, skills, and model routing

damianvtran:
# PR Description

## Change classification

**C2 — standard / pre-authorized.** This is a bounded harness capability and performance release with a clean rollback (`git revert` / reinstall 0.17.0). No foundational auth, data-isolation, infrastructure, or irreversible data change. No RFC or tracker requested.

## Summary of Changes

This releases **local-operator 0.17.2** and closes the measured harness gaps against omp across wall-clock time, token efficiency, tool/MCP efficacy, skills, guidance, and search.

### Token + search efficiency

- `edit`: ordered multi-hunk SEARCH/REPLACE (`edits` list), exact-first then whitespace-tolerant matching, per-line re-indent, and `anchor_line` disambiguation. Existing single-hunk calls remain compatible. Guidance now prefers diff-shaped edits over whole-file writes.
- `read`: Python files read whole return a declaration-only AST summary with symbol line ranges; `range=` and `raw=true` opt out. Broken Python falls back to raw text.
- `grep`/`glob`: `.gitignore`/`.ignore` support, ripgrep fast path with Python fallback, context lines, skip pagination, and lossless spill. Explicitly named ignored glob paths remain reachable.
- Repo context: ancestor `AGENTS.md`/`CLAUDE.md` auto-load once into the cache-stable prompt head, with git-root boundary, duplicate/content/size caps, and an environment kill switch.

### Wall clock + delegation

- Mid-turn compaction now runs at continuing tool-loop boundaries through the existing plan+commit path; the loop prunes its accumulator to prevent summarized messages being persisted after the compaction marker.
- `task` accepts a concurrent `tasks[]` batch plus one shared `context`, supports a read-only `scout` tier, `lo`/`med`/`hi` model routing, and depth-2 delegation (`wake` remains bounded to top-level lifetime).
- Settled top-level task/background-bash jobs auto-deliver into an idle conversation; results already consumed by `wait` do not duplicate.
- Steering a running `bash` call backgrounds it into a tracked job, preserves output/timeout, and auto-delivers completion. Real Ctrl+C/job abort still kills the process group.

### Tool + model efficacy

- Persistent Python `eval` kernel: per-session namespace, trailing-expression results, separate `display()` channel, timeout/crash recovery, bounded output/spill, exec-tier approval.
- Read-only Jedi `lsp` extra: definitions, references, symbol outlines, multi-file rename previews (never writes), optional create-if surface.
[...]

PR #326, opened 2026-08-26
fix(openai): restore gpt-5.6 prompt caching on the OAuth path

damianvtran:
## Summary

`gpt-5.6-sol` sat at a ~89-90% cache-read rate in our analytics while every OpenAI-shaped peer (zai/kimi/qwen/grok) runs 97-98%. It is the only model on OpenAI's Responses API, and on the operator's ChatGPT-subscription **OAuth** path the request body is built by `_build_codex_responses_body`, which was **discarding `prompt_cache_key`** and never requesting encrypted reasoning items. This restores both, matching what OpenAI's own Codex client sends on the same backend.

Net change is small and surgical: two request fields on the Responses path. Verified with live `gpt-5.6-sol` calls against the real subscription backend, not a mock.

- `providers/clients.py` — stop popping `prompt_cache_key` on the codex body (#1); add `include: ["reasoning.encrypted_content"]` when reasoning effort is present (#2); defensively skip the `reasoning` output items that include then surfaces on the stream.
- `tests/unit/providers/test_clients.py` — body-shape assertions for both fields, plus regression guards.
- `scripts/bench_openai_oauth_cache.py` — new live multi-scenario cache-rate benchmark for the OAuth path.

Minor version bump (0.40.0 → 0.41.0): a caching/perf improvement.

## Root cause

On the OAuth path the body posts to `https://chatgpt.com/backend-api/codex/responses` with `store: false`. The prior code did:

```python
body.pop("prompt_cache_key", None)   # comment: "public-API-only keys removed"
```

That comment was a wrong assumption (introduced in #128). OpenAI's own Codex client (`codex-rs/core/src/client.rs`, `build_responses_request`) sets `prompt_cache_key` **unconditionally** on this same `store:false` subscription backend — it defaults to the thread id. Per OpenAI's "Prompt Caching 201", the key is routing stickiness: it is hashed with the prompt prefix so requests with the same prefix land on the same inference machine and hit a warm KV cache. Dropping it scatters a session's turns across cold machines. That is the bulk of the gap.

Codex also sets `include: ["reasoning.encrypted_content"]` whenever reasoning is on (`client.rs` L720-724). `gpt-5.6-sol` is a reasoning model; on `store:false` these encrypted reasoning items are what let a response reuse its own reasoning KV state. We requested neither.

## What was tried and deliberately dropped

OpenAI's GPT-5.6 caching guide documents explicit `prompt_cache_breakpoint` markers and `prompt_cache_options` as the remedy for GPT-5.6's implicit-caching change (their controlled test: 0% → 98.6% on a 9k-token prefix). I implemented that too, then **live-tested it against the actual subscription backend and it is rejected**:

```
HTTP 400: Unsupported parameter: prompt_cache_options
HTTP 400: prompt_cache_breakpoint is not supported on this model
```

This matches OpenAI Codex's own open bug [#35300](https://github.com/openai/codex/issues/35300): those fields are unsupported/unrepresentable on the ChatGPT backend, which is why even Codex cannot use them there. Since the operator has no OpenAI API key (only OAuth), the public Responses path where the guide documents these fields is unreachable and cannot be validated — shipping code that 400s on the only live path would be a regression. So the breakpoint machinery (and the top-level-`instructions` → `developer`-message rewrite it required) was removed entirely; the stable prefix stays in top-level `instructions`, exactly as real Codex does.

## Testing evidence (live, real `gpt-5.6-sol` against the subscription backend)

Cache-read rate = `sum(cache_read_tokens) / sum(input + cache_read + cache_write)`, measured by `scripts/bench_openai_oauth_cache.py` driving real multi-turn conversations through the actual client and reading `input_tokens_details.cached_tokens` off `response.completed`. (`cached ⊆ input` on OpenAI; the codex backend reports `cache_write = 0`.)

| Scenario | Before (origin/main) | After (#1+#2) | 400s |
|---|---|---|---|
| `stable_prefix` (5 diverging sessions, shared prefix) | 34.4% | **46.6%** | none |
| `long_session` (append-only, 5 turns) | 14.7% | **46.4%** | none |
| `reasoning_on` (reasoning effort set) | — | **46.4-47.1%** | none |
| `tool_heavy` (function calls + results) | — | **33.8-47.5%** | none |

Dry-run body confirms the fix on the wire (redacted):

```
POST https://chatgpt.com/backend-api/codex/responses   store: false
before: body keys = [input, instructions, model, reasoning, store, stream, tool_choice, tools]   # prompt_cache_key MISSING
after:  body keys = [include, input, instructions, model, prompt_cache_key, reasoning, store, stream, tool_choice, tools]
        include = ["reasoning.encrypted_content"]   prompt_cache_key = "<session id>"
```

The benchmark probes fresh/short sessions — the worst case, and exactly the cold-start / cross-session-reuse behaviour that drags the production aggregate below its peers. Long warm sessions already cache well on append; the routing key is what lifts everything else.

## Gates

- `flake8 .` ✅ · `black --check .` ✅ (552 files) · `isort --check .` ✅
- `pytest tests/unit/providers -q` → **648 passed**
- pyright: changed files clean (`0 errors`); whole-tree deferred to CI (local box could not complete the whole-tree run under load)

## Scope note for reviewers

Focus on `_build_codex_responses_body` / `_build_responses_body` (the two-field change and that the public path's original behaviour — `instructions`, `prompt_cache_retention="24h"` — is otherwise unchanged), and the `_stream_responses` reasoning-item skip (we do **not** yet replay encrypted reasoning items cross-turn; called out intentionally — the `include` still earns its keep for same-response reuse). The removed breakpoint machinery is gone by design, per the live 400 above.

### DIFF
--- a/local_operator/providers/clients.py
+++ b/local_operator/providers/clients.py
@@ -512,6 +512,42 @@ def _sampling_params(request: ChatRequest, *, top_p_key: str = "top_p") -> dict[
     }
 
 
+# Matches an OpenAI ``gpt-5.<minor>`` model id, capturing the minor version. The
+# id may carry a suffix (``-sol``, ``-codex``, a date) which we ignore; only the
+# ``5.<minor>`` family number decides GPT-5.6-era cache semantics. A bare
+# ``gpt-5`` (no minor) is deliberately NOT matched: it predates the 5.6 implicit
+# caching change and must keep the older behaviour.
+_GPT_5_MINOR_RE = re.compile(r"^gpt-5\.(\d+)(?:[-.].*)?$")
+
+
+def _is_gpt_5_6_or_later(model_id: str) -> bool:
+    """Whether ``model_id`` is GPT-5.6 or a later 5.x that shares its caching.
+
+    GPT-5.6 changed OpenAI's implicit prompt-cache breakpoint to sit at the end
+    of the latest eligible user/tool message, which writes THROUGH the volatile
+    turn tail and loses hits on a large stable prefix followed by a changing
+    suffix (OpenAI Codex bug #35300). The remedies for that regression —
+    explicit ``prompt_cache_breakpoint`` markers and ``prompt_cache_options`` —
+    are GPT-5.6+ only; older Responses models (``gpt-5``, ``gpt-5.4`` …) 400 or
+    behave differently on them. This predicate gates those changes.
+
+    Conservative by construction: anything we cannot confidently parse as
+    ``gpt-5.N`` with ``N >= 6`` is treated as NOT 5.6+, which preserves today's
+    behaviour rather than risk sending a breakpoint to a model that rejects it.
+    """
+    match = _GPT_5_MINOR_RE.match(model_id.strip().lower())
+    if not match:
+        return False
+    return int(match.group(1)) >= 6
+
+
+# OpenAI caps a single request at four cache-WRITE breakpoints (guide: "Each
+# request can create up to four cache writes"). We spend one on the stable
+# system/developer prefix and reserve the rest for the most recent tool
+# results; never emit more than this many explicit markers in one body.
+_MAX_CACHE_BREAKPOINTS = 4
+
+
 def _reasoning_effort(request: ChatRequest) -> str | None:
     """The effort level to send, or ``None`` when the key must not appear.
 
@@ -772,6 +808,49 @@ def _messages_to_openai_responses(messages: Sequence[Message]) -> list[dict[str,
                 )
         items.append({"role": message.role, "content": content})
     return items
+
+
+def _mark_recent_tool_result_breakpoints(input_items: list[dict[str, Any]], budget: int) -> None:
+    """Add ``prompt_cache_breakpoint`` to the last ``budget`` tool results.
+
+    The guide's multi-turn-agent example marks each ``function_call_output``'s
+    last ``input_text`` block so a forked thread can reuse the shared prefix up
+    to that tool result. We only mark the MOST RECENT few, newest-first, to stay
+    within OpenAI's four-cache-write cap (the caller passes the remaining
+    budget after the implicit and stable-prefix writes). A string-valued output
+    is promoted to an ``input_text`` block so it can carry the marker; an
+    already-block output gets the marker on its final text block. Outputs with
+    no text block (image-only) are skipped rather than forced into a shape the
+    model would not read.
+    """
+    if budget <= 0:
+        return
+    marked = 0
+    for item in reversed(input_items):
+        if marked >= budget:
+            break
+        if item.get("type") != "function_call_output":
+            continue
+        output = item.get("output")
+        if isinstance(output, str):
+            item["output"] = [
+                {
+                    "type": "input_text",
+                    "text": output,
+                    "prompt_cache_breakpoint": {"mode": "explicit"},
+                }
+            ]
+            marked += 1
+            continue
+        if isinstance(output, list):
+            text_blocks = [
+                block
+                for block in output
+                if isinstance(block, dict) and block.get("type") == "input_text"
+            ]
+            if text_blocks:
+                text_blocks[-1]["prompt_cache_breakpoint"] = {"mode": "explicit"}
+                marked += 1
 
 
 EMPTY_TOOL_RESULT_TEXT = "[tool returned no output]"
@@ -1139,13 +1218,24 @@ class OpenAICompatClient:
         )
 
     def _build_responses_body(self, request: ChatRequest) -> dict[str, Any]:
-        """Public Responses body using native input items and flat tools."""
+        """Public Responses body using native input items and flat tools.
+
+        On GPT-5.6 and later this also carries the prompt-cache breakpoints the
+        model needs to stay warm; see ``_apply_gpt_5_6_prompt_cache`` for why.
+        """
+        # Gate the GPT-5.6-era cache controls once: ``prompt_cache_breakpoint``,
+        # ``prompt_cache_options`` and the developer-message prefix are 5.6+ only
+        # (older Responses models 400 or behave differently on them).
+        is_gpt_5_6 = _is_gpt_5_6_or_later(request.model.model_id)
         body: dict[str, Any] = {
             "model": request.model.model_id,
             "stream": True,
             "input": _messages_to_openai_responses(request.messages),
         }
-        if request.system_blocks:
+        if request.system_blocks and not is_gpt_5_6:
+            # Pre-5.6: the whole stable prefix rides top-level ``instructions``.
+            # It cannot carry a breakpoint, but pre-5.6 implicit caching does not
+            # write through the turn tail, so a breakpoint is unnecessary here.
             body["instructions"] = "\n\n".join(request.system_blocks)
         if request.tools:
             body["tools"] = _tools_to_openai_responses(request.tools)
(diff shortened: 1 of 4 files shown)


damianvtran:
### Agent review — round 1
Scope: 70d416c8..b9040532

Verified from source in the `/tmp/lop-openai-cache` worktree, not from the PR body. Ran `pytest tests/unit/providers -q` (648 passed) and `flake8` on the three changed files (clean).

**No blockers, no majors.** The change is correct and honest about its one limitation. Findings below are minor/nit only.

**Correctness confirmed:**
- **#1** — `prompt_cache_key` originates in `model/configure.py:3068` (set to the session id when unset) → carried on `ChatRequest` → written by `_build_responses_body` at `clients.py:1181` → flows through `_build_codex_responses_body` unstripped. Only `prompt_cache_retention` (+ the pre-existing `max_output_tokens`/`temperature`/`top_p`) is popped, `store:false` preserved. The public path is otherwise byte-for-byte unchanged, and `test_public_pre_5_6_body_unchanged` guards it.
- **#2** — `include` is set only inside `if effort is not None` (`clients.py:1180`); `test_codex_5_6_body_no_include_when_reasoning_off` proves it is absent on non-reasoning specs. No double-add: nothing else in the body builders writes `include`.
- **Reasoning skip** — placed before the `if/elif` chain (`clients.py:1415`), it matches only `response.reasoning*` event types and `item.type == "reasoning"`; assistant text arrives on `response.output_text.delta` and tool calls on `function_call`, neither of which is touched. Cannot crash: `event_type` is always a `str`, `item` is `payload.get("item") or {}`. Usage parsing on `response.completed` is untouched. The events it drops were already unrendered before this PR (there was no `else` branch surfacing them), so the skip is behaviour-preserving for user-visible output — the explicit `continue` just hardens that.
- **Cross-turn replay omission is safe, not a latent bug.** Before this PR the input carried no reasoning items and the backend accepted it; adding `include` changes only the *response* stream, not what is echoed back on the next turn's `input`, so multi-turn input shape is identical to origin. The live `reasoning_on`/`tool_heavy` scenarios completing with zero 400s is direct counter-evidence to a "backend requires echoed reasoning" concern. The limitation is documented at the skip site.
- **#3 fully removed** — grep confirms `prompt_cache_options`/`prompt_cache_breakpoint` appear only in explanatory comments; no source line emits them and no injected `developer`-role message exists. Regression guards (`_has_breakpoint`, `prompt_cache_options not in body`) cover all four body-shape tests.
- **Secrets** — benchmark auth is header-only; `_redact_body` never emits a body credential and summarizes `instructions`/`input`/`tools` by shape. No token/org-id leak.
- **Version bump** 0.40.0 → 0.41.0 (minor) is correct for a perf/caching feature.

**N1** — `clients.py:1181` — `body["include"] = ["reasoning.encrypted_content"]` unconditionally assigns rather than appends. Harmless today (nothing else sets `include`), but if a future include-bearing field is added on this path it will be silently clobbered. A one-line `setdefault`/extend would be more future-proof. Not blocking.

**N2** — The three new *synchronous* tests (`test_codex_5_6_body_*`, `test_public_*`) inherit the module-level `asyncio` mark and emit `PytestWarning: marked with @pytest.mark.asyncio but it is not an async function`. Pre-existing pattern in this file, cosmetic only.

**Verdict: clean — round is TERMINAL.** No blocker/major findings; N1/N2 are optional follow-ups, not a reason for another round.

damianvtran:

Reviewer round 1 found 0 blockers, 0 majors, 2 nits. Both addressed:

- **N1** (`include` assigned rather than appended) — **Fixed** in `68739209`. `_build_responses_body` now does `include = body.setdefault("include", [])` then appends `reasoning.encrypted_content` only if absent, so a future include-bearing field on this path cannot be clobbered and a re-entry cannot duplicate the value. Gates re-run clean on the changed file (flake8/black/isort/pyright) and `pytest tests/unit/providers -q` → 648 passed.
- **N2** (three sync tests inherit the module-level `asyncio` mark, emitting a cosmetic `PytestWarning`) — **Deferred.** This is a pre-existing, file-wide pattern in `test_clients.py`, not introduced by this PR; the warning is cosmetic and the tests pass. Fixing it repo-wide is out of scope for a caching fix and belongs in a dedicated test-hygiene pass. No ticket filed (cosmetic); noted here for the record.

No blocker/major findings, so no further remediation is required on the substance. Head is now `68739209`; a fresh round-2 review is being run against it to re-establish head-freshness since round 1 was scoped to `b9040532`.

damianvtran:
### Agent review — round 2
Scope: b9040532..68739209

Focused freshness re-review of the single delta commit `68739209` ("refactor(openai): extend include list defensively (review N1)") plus a re-confirmation that the whole change still holds. Head verified at `68739209` (`git rev-parse HEAD`).

**Delta — the N1 fix in `_build_responses_body` (`local_operator/providers/clients.py`):**

```python
include = body.setdefault("include", [])
if "reasoning.encrypted_content" not in include:
    include.append("reasoning.encrypted_content")
```

- **Still gated on reasoning.** The block remains inside `if effort is not None:` — unchanged. `include` is not touched when reasoning is off, so the reasoning-off body is byte-identical to before.
- **Fresh body → `["reasoning.encrypted_content"]`.** Verified: `setdefault` creates the list, the membership check is false, the value is appended. Identical emitted body to the old `body["include"] = [...]` on the reasoning-on path.
- **Cannot duplicate.** Re-entry with the value already present is a no-op (membership guard). Verified.
- **Cannot clobber a pre-existing include.** A pre-existing `["something.else"]` becomes `["something.else", "reasoning.encrypted_content"]` — append, not assign. Verified.
- **No type error.** `setdefault` returns the existing list or a new `list`; `in`/`append` are list-safe. `pyright` gate unaffected (list of str). No behavioural change to either emitted body.

**Re-confirmation (light pass — round 1 did the deep review):**

- **Two-field change intact.** `prompt_cache_key` restored and set with `prompt_cache_retention = "24h"` (L1185–1189), gated on `supports_prompt_cache and request.prompt_cache_key`; the include field present as above. Both present.
- **No forbidden constructs in code.** grep for `prompt_cache_options` / `prompt_cache_breakpoint` returns only three hits, all inside the explanatory comment documenting *why* the Codex backend rejects them (L1152–1154) — no code sets them. No injected `developer`-role message in `providers/`.
- **Tests green.** `.venv/bin/python -m pytest tests/unit/providers -q` → **648 passed** in the worktree `/tmp/lop-openai-cache` (branch `dev-openai-cache`, head `68739209`).

**Findings:** none. 0 blockers, 0 majors, 0 minors, 0 nits. The N1 nit from round 1 is resolved.

**Verdict: clean — TERMINAL.** The delta is a strict, correctness-preserving hardening with no behavioural change to the emitted body in either reasoning state. The round is fresh against the current head (`68739209`) and the PR is mergeable on this head.

damianvtran:
### Agent review remediation — round 2 (CI fix)

Whole-tree pyright in CI caught 5 type errors in the new benchmark script `scripts/bench_openai_oauth_cache.py` (its author had not run whole-tree pyright; the local box could not complete that run under load, which is why it slipped to CI). Fixed in `6516d83f`:

- Added `AgentTool` / `OpenAICompatClient` imports and annotated `_build_prefix() -> tuple[list[AgentTool], list[str]]`, `tools: list[AgentTool]`, `_SCENARIO_RUNNERS: dict[str, Callable[..., Any]]`.
- Added `assert isinstance(client, OpenAICompatClient)` at the two `client_for_spec(...)` sites so the concrete-method introspection type-checks.

Scope: benchmark tooling only — **no change to `local_operator/providers/clients.py` or the tests**, so the reviewed product change from rounds 1-2 is byte-identical. pyright now reports `0 errors` on the file; flake8/black/isort clean; `pytest tests/unit/providers` → 648 passed. A fresh round-3 review is running against head `6516d83f` to re-establish head-freshness, and CI has been re-triggered.

damianvtran:
### Agent review — round 3
Scope: 68739209..6516d83f

Freshness / CI-fix round re-establishing head against the single new commit `6516d83f` since the round-2 head `68739209`.

**Delta is confined to the benchmark.** `git show 6516d83f --stat` touches only `scripts/bench_openai_oauth_cache.py` (+11/-3). `git diff 68739209..6516d83f -- local_operator/providers/clients.py tests/` is empty, so the reviewed product change from rounds 1–2 is byte-identical on this head. No product or test code moved.

**The type fixes are correct and runtime-neutral.**
- Imports added for names now referenced: `AgentTool` (from `harness.types`) and `OpenAICompatClient` (from `providers.clients`). Both are used by the new annotations/asserts; both carry the existing `# noqa: E402` on the post-`sys.path.insert` block.
- Annotations tightened to reality: `_build_prefix() -> tuple[list[AgentTool], list[str]]`, `_make_request(..., tools: list[AgentTool], ...)`, and `_SCENARIO_RUNNERS: dict[str, Callable[..., Any]]`. These are pure signature narrowing — no runtime effect.
- The two `assert isinstance(client, OpenAICompatClient)` guards after `client_for_spec(spec, openai_api="responses")` are sound. `client_for_spec` (clients.py:2296) returns the `WireClient` protocol statically, but for an openai spec it falls through the anthropic/google/mock branches to `return OpenAICompatClient(...)`, so the concrete type is guaranteed. The assert only makes the static type match the runtime type the benchmark already relied on for its `_build_codex_responses_body`/`aclose` introspection; it cannot fire in the openai path.

**Gates re-run on this head (worktree `/tmp/lop-openai-cache`, `HEAD == 6516d83f`):**
- `pyright --pythonpath .venv/bin/python scripts/bench_openai_oauth_cache.py` → `0 errors, 0 warnings, 0 informations`
- `flake8 scripts/bench_openai_oauth_cache.py` → clean
- `black==26.1.0 --check` → `1 file would be left unchanged`
- `isort --check` (repo config, whole-tree profile as CI runs it) → clean
- `pytest tests/unit/providers -q` → `648 passed`

CI on `dev-openai-cache` is in-progress for this head at review time; the one earlier `failure` predates `6516d83f` and is exactly the whole-tree pyright breakage this commit fixes. All CI gates are reproduced clean locally above.

**No BLOCKER, no MAJOR, no MINOR, no NIT.**

Verdict: **clean** — TERMINAL. The product change is unchanged since round 2 and the benchmark now satisfies whole-tree pyright. The PR is mergeable on head `6516d83f`.

damianvtran:
### Agent review — round 4
Scope: origin/main...a0858e8c (post-rebase confirmation)

Post-rebase confirmation review. Rounds 1–3 were clean/terminal; the branch was rebased onto the advanced `origin/main` (upstream now at `0.41.2`), moving head `6516d83f → a0858e8c`. This round verifies the rebase preserved the reviewed change and nothing else.

**Net diff is exactly the reviewed change + version bump.** `git diff --stat origin/main...HEAD` shows only the four expected files:

```
 local_operator/providers/clients.py  |  58 +++-
 pyproject.toml                       |   2 +-
 scripts/bench_openai_oauth_cache.py  | 632 +++++++++++++++++
 tests/unit/providers/test_clients.py | 187 +++++++++
```

No stray files from the rebase; the 7 upstream commits (extension/browser-bridge/harness/TUI work) are all on the `origin/main` side of the merge base, not in this PR's contribution.

**Product/test/bench content is byte-identical to what round 3 reviewed.** `git diff 6516d83f a0858e8c -- local_operator/providers/clients.py tests/unit/providers/test_clients.py scripts/bench_openai_oauth_cache.py` is empty. The only thing the rebase touched in this PR's own files was the `pyproject.toml` version.

**The `clients.py` net diff still shows exactly the reviewed behaviour:**
- `prompt_cache_key` is NO LONGER popped in `_build_codex_responses_body` (the wrong "public-API-only" pop is gone; only `prompt_cache_retention`, `max_output_tokens`, `temperature` are stripped).
- `include` is added via `setdefault("include", [])` + guarded, deduped append, gated on reasoning being present.
- `_stream_responses` skips `response.reasoning*` events and `reasoning` output items.
- NO `prompt_cache_options`, NO `prompt_cache_breakpoint`, NO injected developer message (the only occurrences of those tokens are in explanatory comments stating why they are deliberately absent).

**Version bump is correct.** `0.41.2 → 0.41.3` — the next patch after upstream's `0.41.2`, resolving the version conflict the rebase would otherwise have produced.

**Gates on head a0858e8c:**
- `pytest tests/unit/providers -q` → **648 passed**.
- `pyright local_operator/providers/clients.py scripts/bench_openai_oauth_cache.py` → **0 errors, 0 warnings**.
- `flake8` on the changed files → clean.
- `black --check` (26.1.0) on the changed files → clean.
- `isort --check --profile black` (via `.venv/bin/python -m isort`) → clean.

**Verdict: clean — TERMINAL.** The rebase preserved the reviewed change byte-for-byte, the net diff against current `origin/main` is exactly the reviewed change plus the correct `0.41.3` patch bump, and all gates are green on head `a0858e8c`. No BLOCKER, no MAJOR, no MINOR, no NIT. No further round is warranted; the PR is mergeable on head `a0858e8c`.

PR #328, opened 2026-08-26
fix(exec): reap orphaned bash process groups on hard session death

damianvtran:
## Summary

Orphaned bash **process groups** could outlive the lop process that spawned them, accumulating indefinitely and pinning CPU. Observed in practice: a `pyright` process group ran for ~10 hours after its owning session was gone, and across several concurrent sessions these strays stacked up to 7 full-tree runs at once, each pegging most of a core.

The bash tool already spawns every command `start_new_session=True` (its own process group) and `killpg`s the whole group on every **in-process** stop path — timeout, abort, `jobs` cancel, steer-to-background. The one uncovered case is a **hard death** of the owning lop process (cmux SIGKILL replacing a session, a crash): no in-process code runs, the group was deliberately detached from the controlling terminal so it receives **no SIGHUP**, it reparents to `init`/launchd, and it runs forever. This PR closes that gap by mirroring the owner-liveness discipline `session/retention.py` already uses to reap stale session directories — applied to process groups.

## The one hard constraint

The reaper must **never** kill a legitimate long-running command (ML training, a long query, a slow download) whose owning session is still alive, and must reap only genuine waste. The decision **never** looks at the process's own behavior — runtime, CPU%, and idle time are all unreliable (a 10-hour trainer and a 10-hour stuck `pyright` are indistinguishable by those). The **only** signal consulted is **owner liveness**: a group is waste iff the lop process that spawned it is dead. Duration and CPU are structurally absent from the decision.

### New module `local_operator/tools/group_reaper.py`

Stdlib-only, POSIX, standalone (imports only `retention._process_alive`; deliberately **not** folded into retention, whose contract is "never delete a transcript" — this module kills, and blurring those guarantees would be a hazard).

- **Owner-keyed JSONL ledger** at `<config_dir>/proc-groups/<owner_pid>-<owner_start_token>.jsonl`. Each spawned group appends one line: `{pgid, owner_pid, owner_start, grp_leader_start, cmd, ts}`. Written at spawn (best-effort, like `claim_session`), the line removed on clean group death, the whole ledger unlinked when a dead owner is swept.
- **Two-factor identity** against PID and PGID reuse. `owner_start`/`grp_leader_start` come from `ps -o lstart=` (read under `LC_ALL=C`) and are stored as an **opaque token, never parsed** — a locale/format quirk can only make tokens *differ*, which is fail-safe (errs toward not reaping). No `/proc`, no `psutil` (not installed).
- **`ts` is diagnostics-only** and grep-provably never read by any branch in the logic — this is what makes "no age-based reaping" a structural guarantee rather than a convention.

Owner gate: dead owner → reap; alive + matching token → **never reap** (the live-trainer guarantee); alive + differing token (pid recycled) → reap; unreadable → never reap. Before any `killpg`, a **victim gate** re-checks the group leader (leader gone or recycled onto an innocent group → drop the line, no kill). Every ambiguous case errs toward **not** reaping.

### Two triggers (a hard death can't be caught in-process)

- **Soft death** (`local_operator/cli.py`) — `atexit` + chained SIGTERM/SIGINT handlers at the TUI + headless entry only (never clobbering the `exec_worker`/`mobile.child` handlers that own their own lifecycle), plus a `kill_own_groups()` call in the TUI teardown `finally`. It reads **only the current pid's own ledger**, so it can only ever kill groups this process spawned — sibling lop sessions are untouchable by construction.
- **Hard death** (`local_operator/session_factory.py`) — `sweep_orphan_groups` runs beside the existing retention sweep, inside the same off-loop `asyncio.to_thread` (same boot-stall discipline), best-effort. This is the path that reaps the SIGKILL-orphaned group on the next lop launch.

### Registration hooks (`local_operator/tools/builtin.py`)

`register_group` right after `create_subprocess_exec`; `unregister_group` on the background-cleanup path and on the normal-exit reap.

### Deliberately out of scope

`exec_mode.py` `--background` workers spawn a separate `exec_worker` process (not through `execute_bash`) with its own SIGTERM lifecycle and its own JSONL job ledger — they are owned by design, so they are not orphans and are left alone. `AsyncJobManager`-owned background bash jobs **do** go through `execute_bash`, so they carry a ledger line keyed by the same owner and are correctly reaped only when that owner dies (a detached job outliving its owner *is* waste — its delivery sink died with the owner). On Windows both entry points no-op (the leak is POSIX-only).

### Version

`pyproject.toml`: **0.41.2 → 0.41.3** (patch — a reliability fix, no new user-facing capability).

### Unit — `tests/unit/tools/test_group_reaper.py` (20 tests, mirrors `test_retention.py` style)

Covers the full owner and victim truth tables: owner dead → reap; owner alive + matching token → **never** reap; owner alive + differing token (pid reuse) → reap; owner unreadable → never reap; victim gate (leader gone → drop, no kill; leader recycled different-start → drop, no kill; leader alive + match → kill); torn/corrupt JSONL line skipped; double-unlink / double-`killpg` suppressed; win32 → no-op; and an explicit assertion that `ts` never influences a decision. **20 passed.** Full unit suite: **7088 passed, 7 skipped**.

### End-to-end — real process groups, no 10-hour wait

`tests/e2e_group_reaper_repro.py` drives the actual spawn + ledger path and asserts via `os.killpg(pgid, 0)` / child-wait. Independently re-run by the manager:

```
A) hard death, owner DEAD -> group survives owner, sweep REAPS
  [PASS] group is alive after spawn (killpg 0 succeeds)
  [PASS] group SURVIVED the owner's death (still alive)
  [PASS] sweep reaped exactly one group
  [PASS] group is REAPED after sweep (killpg raises ProcessLookupError)
  [PASS] dead owner's ledger unlinked
B) owner ALIVE -> sweep does NOT reap (live-runner anti-regression)
  [PASS] group alive after spawn
  [PASS] sweep reaped nothing
  [PASS] sweep skipped a live owner
  [PASS] group SURVIVES (live owner's long-runner untouched)
C) PID-reuse decoys -> both REAP; C-inverse -> innocent SURVIVES
  [PASS] both decoys reaped (row1 dead-pid + row3 wrong-token)
  [PASS] C1 alive-but-wrong-token group REAPED
  [PASS] C2 dead-owner group REAPED
  [PASS] C-inverse: sweep reaped nothing (victim gate held)
  [PASS] C-inverse: innocent live group SURVIVES
D) soft death -> kill_own_groups reaps own group, spares sibling owner
  [PASS] own group REAPED by soft death
  [PASS] sibling owner's group SURVIVES (scope proof)
  [PASS] own ledger unlinked
  [PASS] sibling ledger intact
ALL SCENARIOS PASSED
```

Scenario **B** is the anti-regression proof for legitimate long-runners (a live owner's group is never touched at any age); **D** proves the soft-death reaper is scoped to its own session and cannot reach a sibling's live group.

### Gates

flake8 clean · isort (5.13.2, `--profile black`) clean · black (26.1.0) `--check` clean · scoped pyright over the 5 changed files `0 errors, 0 warnings` · full unit suite green. Full-tree pyright is deferred to CI (it was being CPU-starved locally by a peer worktree's pyright run — the very pathology this PR addresses).

## Design doc

`docs/design/orphan-group-reaper.md` (in-branch) records the locked design, the empirical `sh -c` exec-in-place finding that makes two-factor group identity safe, and the four places the retention analogy is deliberately stretched.

### DIFF
--- a/local_operator/cli.py
+++ b/local_operator/cli.py
@@ -2343,6 +2343,63 @@ async def _run_with_scheduler(run_fn, *run_args) -> int:
                 pass
 
 
+def _install_group_reaper_soft_death() -> None:
+    """Wire the process-group reaper's soft-death path for THIS process.
+
+    Registers ``group_reaper.kill_own_groups`` as an ``atexit`` hook and as a
+    SIGTERM/SIGINT handler, so a catchable stop of the interactive TUI/headless
+    REPL reaps this process's own still-live bash groups instead of leaking them
+    to the next launch's startup sweep. The whole leak this addresses is a HARD
+    (uncatchable SIGKILL) death, which no handler can cover — but a POLITE stop
+    (cmux replace, launchd stop, Ctrl+C at the REPL) IS catchable, and reaping
+    it here makes the common case instant and precise rather than deferred.
+
+    Scoped to the interactive entry on purpose: ``exec``/``serve``/``mobile``
+    own their own SIGTERM semantics (``exec_worker.py``, ``mobile/child.py``) and
+    are dispatched before this is ever called. As a second belt, any
+    pre-existing SIGTERM/SIGINT handler is CHAINED, not clobbered — the reaper
+    runs first, then the previous handler (or the default) still fires — so this
+    can never silently swallow a signal another component was relying on.
+
+    Best-effort and idempotent: the reaper unlinks its ledger on the first call,
+    so the atexit hook, a signal, and the TUI teardown ``finally`` firing in any
+    order all converge on one reap. On Windows the reaper is a no-op, and the
+    signal registration is guarded so a platform without SIGTERM is harmless.
+    """
+    import atexit
+    import contextlib
+    import signal
+
+    from local_operator.tools.group_reaper import kill_own_groups
+
+    atexit.register(kill_own_groups)
+
+    def _chain(signum: int) -> None:
+        previous = signal.getsignal(signum)
+
+        def _handler(received_signum, frame):  # type: ignore[no-untyped-def]
+            try:
+                kill_own_groups()
+            except Exception:  # noqa: BLE001 — a handler must never raise
+                pass
+            # Chain to whatever was installed before us so the process still
+            # stops the way it otherwise would (default disposition included).
+            if callable(previous):
+                previous(received_signum, frame)
+            elif previous == signal.SIG_DFL:
+                # Restore the default and re-raise so the default action (e.g.
+                # terminate) actually happens rather than being swallowed.
+                signal.signal(received_signum, signal.SIG_DFL)
+                os.kill(os.getpid(), received_signum)
+
+        with contextlib.suppress(ValueError, OSError):
+            # ValueError: not the main thread; OSError: unsupported signal.
+            signal.signal(signum, _handler)
+
+    for _sig in (signal.SIGTERM, signal.SIGINT):
+        _chain(_sig)
+
+
 def main() -> int:
     # FIRST, before anything else can log. `helpers.py` used to configure the
     # root logger as an import side effect; now the entry point owns it, which
@@ -2799,6 +2856,19 @@ def main() -> int:
         # event loop from exec_mode/the server module.
         import asyncio
 
+        # Soft-death process-group reaper (tools/group_reaper.py). Installed
+        # ONLY on the interactive TUI + headless REPL entry — the two paths that
+        # spawn bash tool groups and tear down in this process. A catchable stop
+        # (the polite cmux stop, a launchd stop, a clean quit, or an unexpected
+        # exit) then kills this process's own still-live bash groups precisely
+        # and instantly instead of leaving them for the next launch's sweep.
+        # Deliberately NOT installed for `exec`/`serve`/`mobile`: those own their
+        # own SIGTERM lifecycle (exec_worker.py, mobile/child.py) and must keep
+        # it — `_install_group_reaper_soft_death` chains any pre-existing handler
+        # rather than clobbering it, but scoping to here keeps the concern where
+        # the groups are actually created.
+        _install_group_reaper_soft_death()
+
         if use_tui and run_tui is not None:
             tui_config = config_manager.get_config_value("tui", None)
             theme_name = tui_config.get("theme", "dark") if isinstance(tui_config, dict) else "dark"
@@ -2904,6 +2974,16 @@ def main() -> int:
                 try:
                     tui_auth_store.close()
                 except Exception:  # noqa: BLE001 — closing on teardown, never fatal
+                    pass
+                # Reap this process's own bash groups on TUI teardown — a clean
+                # quit, an exception exit, or after run_tui returns. Idempotent
+                # with the atexit/signal hooks (the ledger is unlinked on the
+                # first call). See _install_group_reaper_soft_death.
+                try:
+                    from local_operator.tools.group_reaper import kill_own_groups
+
+                    kill_own_groups()
+                except Exception:  # noqa: BLE001 — teardown, never fatal
                     pass
 
         return asyncio.run(
--- a/local_operator/cli.py
+++ b/local_operator/cli.py
@@ -2347,17 +2347,29 @@ def _install_group_reaper_soft_death() -> None:
     """Wire the process-group reaper's soft-death path for THIS process.
 
     Registers ``group_reaper.kill_own_groups`` as an ``atexit`` hook and as a
-    SIGTERM/SIGINT handler, so a catchable stop of the interactive TUI/headless
+    SIGTERM handler, so a catchable stop of the interactive TUI/headless
     REPL reaps this process's own still-live bash groups instead of leaking them
     to the next launch's startup sweep. The whole leak this addresses is a HARD
     (uncatchable SIGKILL) death, which no handler can cover — but a POLITE stop
-    (cmux replace, launchd stop, Ctrl+C at the REPL) IS catchable, and reaping
-    it here makes the common case instant and precise rather than deferred.
+    (cmux replace, launchd stop, Ctrl+D / quit / window close at the REPL) IS
+    catchable, and reaping it here makes the common case instant and precise
+    rather than deferred.
+
+    SIGINT is DELIBERATELY excluded. In the headless REPL, Ctrl-C is a *turn
+    abort that keeps the session alive* (``_run_headless_repl`` catches
+    ``KeyboardInterrupt`` -> ``session.abort`` -> loops), and ``session.abort``
+    deliberately spares ``background=true`` bash jobs — they exist precisely so a
+    build or deploy outlives the turn that started it. Reaping on SIGINT would
+    SIGKILL those still-live groups while the owning REPL keeps running, which is
+    exactly the never-kill-a-live-owner case this whole module forbids. Every
+    real REPL/TUI *exit* (Ctrl-D, ``quit``, window close) still reaps via the
+    ``atexit`` hook, ``session.dispose()`` and the TUI teardown ``finally``, so
+    nothing is lost by leaving SIGINT to its turn-abort semantics.
 
     Scoped to the interactive entry on purpose: ``exec``/``serve``/``mobile``
     own their own SIGTERM semantics (``exec_worker.py``, ``mobile/child.py``) and
     are dispatched before this is ever called. As a second belt, any
-    pre-existing SIGTERM/SIGINT handler is CHAINED, not clobbered — the reaper
+    pre-existing SIGTERM handler is CHAINED, not clobbered — the reaper
     runs first, then the previous handler (or the default) still fires — so this
     can never silently swallow a signal another component was relying on.
 
@@ -2396,7 +2408,10 @@ def _install_group_reaper_soft_death() -> None:
             # ValueError: not the main thread; OSError: unsupported signal.
             signal.signal(signum, _handler)
 
-    for _sig in (signal.SIGTERM, signal.SIGINT):
+    # SIGTERM only. SIGINT is a turn abort in the headless REPL and must NOT
+    # reap live background jobs (see the docstring); a Ctrl-C that actually
+    # exits reaps through atexit/dispose instead.
+    for _sig in (signal.SIGTERM,):
         _chain(_sig)
 
 
--- a/local_operator/session_factory.py
+++ b/local_operator/session_factory.py
@@ -1057,6 +1057,23 @@ async def _prepare(
         sweep_from_config, config_manager, Path(agent_registry.config_dir), transcript_dir
     )
 
+    # Hard-death process-group reaper (tools/group_reaper.py). Runs beside the
+    # retention sweep and for the same reasons: it is a disk walk over OTHER
+    # processes' ledgers plus ps/os.kill probes, none of which touches THIS
+    # session's construction, so it goes off-loop to stay out of the boot stall;
+    # and it is best-effort, so a sweep error must never fail session start.
+    # It reaps a bash process group only when the lop process that spawned it is
+    # provably dead — the one leak _kill() cannot cover, because a SIGKILLed
+    # owner runs no in-process cleanup and start_new_session already stripped the
+    # group's SIGHUP. Owner liveness is the ONLY signal, so a live session's long
+    # command (e.g. a 10h trainer) is never touched.
+    from local_operator.tools.group_reaper import sweep_orphan_groups
+
+    try:
+        await asyncio.to_thread(sweep_orphan_groups, Path(agent_registry.config_dir))
+    except Exception:  # noqa: BLE001 — best-effort; never block session start
+        logger.debug("orphan process-group sweep failed", exc_info=True)
+
     # Stamp session directories that predate the origin marker, so the
     # ``/resume`` picker stops offering delegated runs on the FIRST launch
     # after an upgrade rather than once natural churn has cleared the store.
--- a/local_operator/tools/builtin.py
+++ b/local_operator/tools/builtin.py
@@ -98,6 +98,7 @@ from local_operator.imaging import (
     bound_image_for_model,
 )
 from local_operator.media import ImageInfo, sniff_image_file
+from local_operator.tools import group_reaper
 from local_operator.tools.spill import (
     SPILL_ENTRY_LIMIT_BYTES,
     SPILL_SCHEME,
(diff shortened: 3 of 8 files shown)


damianvtran:
### Agent review — round 1

Scope: origin/main..1df47b2f16ca0e1e8df85dc8a0321add48321c6b

Reviewed the rebased PR head `1df47b2f` (the worktree's `71d50167` was a stale pre-rebase copy; the reaper files are byte-identical between them, so this review applies verbatim). Scope on the true head is exactly the reaper: `group_reaper.py`, the `builtin.py`/`cli.py`/`session_factory.py` hooks, the version bump (0.41.2 → 0.41.3, correct patch for a reliability fix with no new capability), and the two test files. Ran the unit suite (20 passed), flake8, pyright, and black on the changed files — all clean.

The core module is strong. The owner truth table (`_owner_is_dead`, L467-483) and the victim/pgid-reuse gate (`_reap_if_still_ours`, L486-532) both err toward NOT reaping on every ambiguous cell — `None` token, unreadable leader, differing leader start — and I verified the null-`grp_leader_start` case cannot false-match because `current_leader is None` returns before the equality compare. `ts` is confirmed write-only (grep: written at L232, never compared; `test_ts_is_never_a_decision_input` enforces this structurally). The unit tests spawn real `start_new_session` groups and prove kills via leader-exit rather than a racy `killpg(pgid,0)`, so the truth-table rows are genuinely exercised, not mocked. The sweep path, soft-death scope (own ledger only, keyed by `<pid>-<start>`), Windows no-op, and double-unlink/double-kill suppression are all sound.

One blocker, in the soft-death **wiring** (not the module):

---

**R1 — blocker — `cli.py:2399` (`_install_group_reaper_soft_death`): the SIGINT handler kills live background bash groups on a mid-turn Ctrl-C, violating the never-kill-a-live-owner guarantee through a non-sweep path.**

The handler is chained onto BOTH signals: `for _sig in (signal.SIGTERM, signal.SIGINT): _chain(_sig)`. In the headless REPL (`_run_headless_repl`, `cli.py:1938`), Ctrl-C is documented and coded as a *turn abort that keeps the REPL alive* — the banner says "Ctrl-C interrupts a turn, Ctrl-D exits", and `await session.prompt(line)` catches `KeyboardInterrupt` → `session.abort("interrupted")` → loops. `session.abort` deliberately does NOT touch bash jobs; the codebase's own contract (`session.py:3445`) is that `background=true` "exists precisely so a build or a deploy outlives the turn that started it, and killing one on a keypress meant for the agent destroys work the user asked to be insulated from." But the new SIGINT handler runs `kill_own_groups()` *before* chaining to `default_int_handler`, so it SIGKILLs every still-live group this owner registered — including background jobs the user meant to keep — while the owning REPL process stays alive. That is exactly the forbidden case: reaping a live owner's group.

Reproduced against the real hook logic:
```
registered pgid 93823 alive: True
REPL: caught KeyboardInterrupt -> abort turn, keep REPL alive
background job alive after Ctrl-C: False rc: -9
```
The background `sleep 600` is SIGKILLed (`rc -9`) by a Ctrl-C the user pressed to abort a different turn, with the session still running.

Fix: install the reaper on **SIGTERM only** — `for _sig in (signal.SIGTERM,):` — and drop SIGINT from the loop. Nothing is lost: a Ctrl-C that actually *exits* the REPL (at the `input()` prompt, or Ctrl-D) still reaps via the `atexit.register(kill_own_groups)` hook and `session.dispose()`; only the mid-turn abort stops wrongly killing live jobs. (SIGTERM — cmux replace, launchd stop — genuinely means "this process is going away", so reaping there is correct and wanted.)

---

**R2 — minor — `group_reaper.py:212, 270, 343`: the owner start-token is re-derived with a `ps` subprocess on every register/unregister/kill call, on the bash hot path.**

`_owner_start_token(os.getpid())` shells out to `ps -o lstart=` on each `register_group` (plus a second `ps` for the leader), each `unregister_group`, and each `kill_own_groups`. A session that spawns many short bash commands pays 2-3 `ps` fork/execs per command. The owner's own start token is immutable for the life of the process — cache it once (module-level memo keyed on `os.getpid()`) and the register/unregister paths drop to one `ps` (the leader) or zero. Safe because the value cannot change under a fixed pid.

---

**R3 — minor — `group_reaper.py:277-313` (`_rewrite_without_pgid`): the unregister read-modify-write can clobber a concurrent append from a parallel bash spawn, silently dropping that registration.**

Registration is an atomic `open(path,"a")` append (correct), but unregister does `read_text()` → filter → `write_text()` (truncate+rewrite). If a parallel bash command's `register_group` append lands between this read and this write, the new line is lost, so that group is never in the ledger and won't be reaped on a hard owner death. It's the safe direction (a leaked group = pre-existing behaviour, never a wrong kill), and parallel bash + background jobs is a first-class feature here, so the window is real. Note that unregister is not needed for correctness at all — the sweep and soft-death paths already drop any entry whose leader is gone (`_reap_if_still_ours` L508) — it exists only to bound ledger growth. Cheapest fix: make unregister append a tombstone line (`{"unregister": pgid}`) that the readers honour, or simply drop the rewrite and let the sweep's dead-leader check reclaim stale lines. Either removes the lossy truncate entirely.

---

**On the manager's e2e-PYTHONPATH question:** test-harness nit, not a production concern. `tests/e2e_group_reaper_repro.py` imports `group_reaper`/`builtin` in-process and its only child processes are `/bin/sh` sleeps (no child `python -m local_operator`), so the `PYTHONPATH=/tmp/lop-reaper` in the run instructions only ensures the worktree source is exercised instead of the installed package. In production, register/sweep/kill all run in-process; there is no child interpreter that needs `local_operator` on its path. Worth a one-line comment in the script's docstring, nothing more.

**Verdict: NOT clean — 1 blocker (R1), 0 major.** R1 is a genuine live-owner kill through the SIGINT path and must be fixed before merge; the one-line fix (SIGTERM-only registration) is low-risk. R2/R3 are minors/follow-ups. Once R1 is remediated on the current head, this is a clean, well-guarded implementation of the design's owner-liveness-only contract. Re-review the SIGINT scope after the fix.

damianvtran:

R1 (blocker): Fixed in 0d264183. `_install_group_reaper_soft_death` now chains the reaper on **SIGTERM only** — SIGINT is dropped from the registration loop. In the headless REPL Ctrl-C is a *turn abort* that `session.abort` deliberately keeps `background=true` jobs alive through, so reaping on SIGINT was killing a live owner's groups. Real exits still reap via `atexit` + `session.dispose()` + the TUI teardown `finally`. Proof: new e2e **scenario E** drives a real child that installs the hook and holds a live registered `sleep 600` group — `group ALIVE after SIGINT (Ctrl-C spared the background job)` + `REPL child still running after SIGINT`, then `group REAPED after SIGTERM (real stop cleaned up)`; plus two `test_cli_new.py` unit tests pinning "installs on SIGTERM, leaves SIGINT untouched" and "SIGTERM handler reaps then chains". The reviewer's original `rc -9` repro (Ctrl-C SIGKILLs a live sleep) no longer reproduces.

R2 (minor): Fixed in 0d264183. Added `_self_start_token`, a module-level memo of THIS process's start token (immutable for the life of a fixed pid), used by register/unregister/kill. The bash hot path drops from a `ps -o lstart=` fork per call to zero for the owner token (the leader token is still probed live at register, as it must be). Foreign pids are always probed live (never served from the memo), and a transient `ps` failure is not cached so a later call retries rather than sticking on `None`. Covered by `test_self_start_token_is_cached_across_calls` / `_probes_foreign_pid_live` / `_failure_is_not_cached`.

R3 (minor): Fixed in 0d264183. `unregister`'s read-modify-write now runs under an **exclusive** per-owner advisory `flock` (`_ledger_lock`), while `register`'s append takes it **shared** — a readers/writer lock keyed on the per-owner ledger. A concurrent register can no longer land inside (and be truncated away by) the unregister rewrite window. It degrades to unlocked if `flock` is unavailable, whose worst case is the pre-existing leaked-line behaviour (never a wrong kill). Covered by `test_unregister_preserves_concurrent_register` (deterministic mid-rewrite append) and `test_ledger_lock_degrades_when_unavailable`.

**e2e-PYTHONPATH nit:** Deferred as a comment-only fix — added a docstring paragraph to `tests/e2e_group_reaper_repro.py` explaining that `PYTHONPATH` only forces the worktree source (test-harness convenience, not a production requirement, since register/sweep/kill all run in-process). No production change warranted, per the reviewer's own assessment.

---

**Gates (all green on 0d264183, changed files):**
- `flake8` — clean
- `black --check` (26.1.0) — 5 files unchanged
- `isort --check` (repo config, not `--profile black`) — clean
- `pyright` (scoped) — 0 errors (also fixed 7 pre-existing `str | None`/`manager.shutdown`/`proc.stdout` type errors that were failing the `type-check` CI job on this branch)
- `pytest tests/unit/tools/test_group_reaper.py tests/unit/test_cli_new.py` — **115 passed**
- e2e `tests/e2e_group_reaper_repro.py` — **ALL SCENARIOS PASSED** (A–E)

New head: `0d264183`. Ready for a fresh review round against it.

damianvtran:
### Agent review — round 2

Scope: origin/main..0d264183

Re-reviewed the remediation head `0d264183` against round 1. Ran the full gate on the changed files: flake8 clean, `black --check` (26.1.0) 5 files unchanged, `isort --check` (repo config) clean, `pyright --pythonpath .venv/bin/python` **0 errors**, `pytest tests/unit/tools/test_group_reaper.py tests/unit/test_cli_new.py` **115 passed**, and the e2e repro **A–E all PASSED** (real signals, not mocked). All three round-1 findings are genuinely fixed and the core never-kill-a-live-owner guarantee is intact.

**R1 (blocker) — RESOLVED, verified airtight.** SIGINT is dropped from the registration loop (`cli.py:2414` → `for _sig in (signal.SIGTERM,):`). Grepped the whole tree: nothing wires `kill_own_groups` to SIGINT (`group_reaper.py:421` and `cli.py:2358/2411` are doc/comment only; the SIGINT handlers in `mobile/child.py:118` and `mobile/service.py:53` are mobile's own and never call the reaper — untouched by this PR, as claimed). The SIGTERM chain preserves a pre-existing handler and re-raises `SIG_DFL` via `os.kill` so the default termination still fires — confirmed by `test_soft_death_sigterm_handler_reaps_then_chains` (reaps then chains) and `test_soft_death_reaper_installs_on_sigterm_not_sigint` (SIGINT left as `default_int_handler`). Real exits still reap: `atexit.register(kill_own_groups)` + the TUI teardown `finally` (`cli.py:2996`) + `session.dispose()`, all idempotent on the once-unlinked ledger. **e2e scenario E proves it end-to-end, not mocked**: a real `python -c` child installs the hook, holds a live `start_new_session` `sleep 600` group, models the headless-REPL turn-abort (`KeyboardInterrupt` → keep running), and the parent drives a **real** `SIGINT` then `SIGTERM` — `group ALIVE after SIGINT` + `REPL child still running after SIGINT`, then `group REAPED after SIGTERM`, all via real `killpg(pgid, 0)` liveness. Round 1's `rc -9` repro no longer reproduces.

**R2 (minor) — RESOLVED.** `_self_start_token` (`group_reaper.py:167`) memoizes only `os.getpid()`'s token; `pid != os.getpid()` short-circuits to a live `_owner_start_token` for any foreign pid (never served or stored from the memo), and a `None` (transient `ps` failure) is not cached so a later call retries. The group-**leader** token is still probed live at register (`L251`), as it must be. Covered by the three new unit tests, which I ran green.

**R3 (minor) — RESOLVED.** `_ledger_lock` takes the lock on a **separate** `<ledger>.lock` inode, so the rewrite unlinking the `.jsonl` cannot destroy the lock file mid-hold; EX on `_rewrite_without_pgid`, SH on the register append — a readers/writer lock per owner ledger (owners never contend, so no cross-owner blocking and no lock-ordering cycle → cannot deadlock). Degrades to unlocked on `ImportError`/`OSError` with the fd cleaned up, worst case the pre-existing leaked-line behaviour, never a wrong kill. `test_ledger_lock_degrades_when_unavailable` exercises the no-`fcntl` path green.

**7 pre-existing pyright errors — CONFIRMED fixed.** Ran pyright on the r1 head (`1df47b2f`) test/e2e files and reproduced exactly 7 `reportArgumentType`/`reportReturnType`/`reportAttributeAccessIssue` errors (the `str | None` → `_ledger_path`, the `object`→`int` return, the `manager.shutdown` access). On `0d264183` the same scope is 0 errors, resolved via the `_token()` helper, an `isinstance(pgid, int)` assert, and a `getattr(manager, "shutdown", None)` guard. The `type-check` CI job should pass.

**Owner/victim truth tables — intact.** The remediation touched only the self-token derivation and the append/rewrite locking; `_owner_is_dead` and `_reap_if_still_ours` are unchanged, and the sweep still skips a live owner (e2e B) and holds the victim/pgid-reuse gate (e2e C / C-inverse), both green.

---

**R4 — minor — `group_reaper.py:340`: the `<ledger>.lock` file is never removed, so `proc-groups/` accumulates one 0-byte husk per `lop` launch that spawned a bash group.** The sweep only globs `*.jsonl` (`L500`); when a ledger is emptied and `unlink`ed, its sibling `.lock` stays behind. Reproduced: after registering then unregistering one group, `proc-groups/` still contains `<pid>-<start>.jsonl.lock`. Non-blocking (correctness-preserving, never mis-globbed, never a wrong kill), but the count grows unbounded over the machine's lifetime. Fix (follow-up): teach the sweep's empty-ledger/`_safe_unlink` path to also remove a matching `.lock`, or glob `*.jsonl*`.

**R5 — minor (defense-in-depth) — `group_reaper.py:167`: the R2 memo is not fork-safe — it stores the token but not the pid it cached for.** In a `fork()`-without-`exec` child, `_SELF_START_TOKEN` is inherited from the parent while `pid == os.getpid()` is the *child's* pid, so `_self_start_token` would return the parent's token and key the child's ledger wrong. Not currently reachable — the reaper install and interactive bash-register run single-process, and the server/job multiprocessing paths don't install the reaper nor pre-populate the memo in the parent — so this is latent, not a live bug. One-line hardening (follow-up): cache `(pid, token)` and invalidate when the stored pid != `os.getpid()`.

---

**Verdict: CLEAN — 0 blocker, 0 major.** R1/R2/R3 are all resolved and verified on the current head; the round is fresh against `0d264183`. R4 and R5 are correctness-preserving follow-ups, not a reason for another round. **This round is TERMINAL and the PR is ready to merge.**

Advisory (not a code finding, non-blocking): the branch is behind `origin/main` (missing #326's OAuth-cache change), which makes the two-dot `origin/main..0d264183` scope spuriously show a `clients.py`/`bench` revert. A squash-merge applies only this branch's own commits and does NOT revert #326, so this is a review-scope artifact only — a rebase before merge would tidy the diff but is not required for correctness.

damianvtran:

R4 (minor): Fixed in b856f9e0. The startup sweep now unlinks a dead owner's `.jsonl.lock` sidecar alongside its ledger, plus a second pass reaps orphan `.lock` husks (the graceful-shutdown case, where the ledger is compacted away but the lock remains) whenever the parsed owner pid is provably gone; a live owner's lock is never touched. Proven by `test_sweep_removes_dead_owner_lock_spares_live_owner` (dead owner's lock gone, live owner's lock+ledger untouched) plus `test_sweep_removes_orphan_lock_of_dead_owner`, `test_sweep_spares_orphan_lock_of_live_owner`, `test_sweep_leaves_unparseable_lock_name`, and `test_sweep_never_treats_lock_as_ledger` (ledger glob `*.jsonl` never picks up a `.lock`).
R5 (minor): Deferred — added a code comment at the `_SELF_START_TOKEN` memo documenting the fork-without-exec staleness assumption; unreachable on the current install path (reaper installs only on the TUI/headless entry, which never forks without a following exec), so no re-architecture.

damianvtran:
### Agent review — round 3

Scope: origin/main..b856f9e0

R4: resolved. R5: deferral accepted.

Verified against head b856f9e0 (freshly rebased on origin/main, clean stack: reaper commits + `0.41.4` bump, no spurious cross-file diffs).

- **(a) Never removes a live owner's lock.** The dead-owner ledger loop only reaches `_safe_unlink(_lock_path(ledger))` *after* every live/ambiguous case (`owner_pid == me`, verdict `None`, verdict false) has already `continue`d out — the lock is dropped at the exact point owner-death is already proven. The orphan-lock second pass (`_sweep_orphan_locks`) reaps a `.lock` only when `_owner_pid_from_lock` parses an integer pid **and** `_process_alive(pid)` is false; a live pid or unparseable name is left in place. No path can unlink a live owner's lock, so the R3 race is not reintroduced.
- **(b) Ledger glob unchanged.** `sweep_orphan_groups` still globs `*.jsonl` (group_reaper.py:544); locks glob `*.jsonl.lock` (`*` + `_LOCK_SUFFIX`) in a separate pass. `.lock` can never match `*.jsonl`; test `test_sweep_never_treats_lock_as_ledger` asserts `scanned_ledgers == 0` for a lone lock husk.
- **(c) Ambiguous/unparseable lock left untouched.** `_owner_pid_from_lock` returns `None` when the leading segment is not an integer; the pass skips it. Covered by `test_sweep_leaves_unparseable_lock_name`.
- **(d) R5 deferral accurate.** The `_SELF_START_TOKEN` memo comment correctly states the fork-staleness path is latent-only and unreachable on the install path (reaper installs solely on the interactive TUI / headless entry, which never forks without a following exec; a fork+exec child re-imports and restarts at `None`). Accepted as deferred.

Core guarantees intact: never-kill-a-live-owner (soft-death path reads only the current process's own ledger; sweep skips live/ambiguous owners) and the R1 SIGINT fix (SIGINT deliberately excluded, cli.py:2358; reap wired on SIGTERM + atexit + teardown only) both unchanged. The 5 new tests assert exactly what they claim; full `test_group_reaper.py` suite green (30 passed). flake8 clean on both changed files.

**CLEAN — 0 blocker, 0 major. Round is TERMINAL** (R5 is an accepted deferral, not a reason for another round). Ready to merge.

PR #329, opened 2026-08-26
feat(browser): make site approval a first-class agent-legible flow

damianvtran:
## What

Makes site-permission handling a first-class, agent-legible flow instead of a
single long-blocking navigation call that fails opaquely.

**Before:** `open`/`goto` to a not-yet-allowed origin blocked the whole RPC
inside the extension's 60 s popup prompt (auto-deny on expiry). The agent never
got a turn to tell the user a prompt was up, so prompts expired unseen, the
daemon extended the command deadline (A3), and the session client timed out
with a misleading "bridge unreachable" — a real session (0ee4974ba84a) burned
an hour misdiagnosing the bridge as broken.

**After — the three-step dance:**

1. `open`/`goto` to a non-allowed top-level origin **fails early** (measured
   0.01 s, no prompt raised) with typed `origin_not_allowed` whose message
   teaches the flow: call `request_access`, **notify the user** (ask tool or
   message), then `await_access`.
2. New action **`request_access`** (url): raises the pending approval — popup
   badge `!`, prompt view, best-effort Chrome notification — and returns
   `{origin, state: pending|allowed|denied}` immediately. Idempotent for the
   same origin; a different origin **replaces** the pending slot (the popup's
   existing single-`pendingOrigin` semantics; a queue would show prompts the
   agent no longer cares about). Pending request TTL is **10 minutes** — the
   user was notified asynchronously and may take a while; still bounded so a
   forgotten Allow can't grant a navigation nobody remembers.
3. New action **`await_access`** (url, timeout up to 240 s): waits for the
   decision. Implemented as **client-side slice looping** over a bounded
   extension-side poll (20 s slices) rather than a daemon long-poll: every
   `COMMAND_TIMEOUTS` entry stays an honest per-RPC bound (the deadline-
   extension carve-out is exactly what made the old failures unreadable), and
   the MV3 worker never holds a multi-minute promise across worker death —
   the decision lives in session storage, which survives.

"Allow once" from this path mints a session-scoped grant consumed by exactly
one navigation (10 min TTL); "Always allow" persists in the origins map as
today; deny answers `denied` with do-not-retry guidance. Redirect-chain gating
inside a running navigation keeps the synchronous prompt (a mid-flight hop
can't fail early). On cmux these actions degrade with the typed not-supported
pattern used by scroll/logs.

## Notification reality (macOS caveat — for user docs)

Live testing on a real machine showed the `chrome.notifications` banner
**never reaches the user on macOS without Chrome having Notification Center
authorization** (System Settings → Notifications → Google Chrome; many
machines have no Chrome entry at all — the "notification" renders only inside
Chrome's extensions menu). The design therefore treats the **agent as the
primary notification channel**: the `origin_not_allowed` error and the
`request_access` pending result both explicitly instruct the agent to notify
the user via the harness (ask tool / message) *before* `await_access` — the
harness path is confirmed reliable. The Chrome banner is retained best-effort
and is now **clickable**: `notifications.onClicked → chrome.action.openPopup()`,
wrapped in try/catch because `openPopup` historically demands a user gesture
and Chrome moves that boundary between versions; the notification text keeps
pointing at the toolbar icon as the fallback.

## Surface

- `protocol.py`: `METHODS` += `request_access`/`await_access`,
  `ORIGIN_NOT_ALLOWED` error code; `protocol.gen.ts` regenerated.
- `extension/src/access-flow.ts` (new): pure TTL/state machine (tested DOM-free).
- `extension/src/origins.ts`: prompt-raise decoupled from in-command wait;
  `resolveOrigin` resolves **both** an in-command wait and the async record;
  once-grant mint/consume; lazy TTL expiry that also drops stale prompt surfaces.
- `extension/src/commands/access.ts` (new): the two handlers + slice poll.
- `extension/src/commands/nav.ts`: rebased onto #321's multi-surface open/goto
  model. `ensureTopLevelAccess` makes ONE admission decision at command entry
  and consumes the grant there; `navigate()` receives that decision unchanged,
  so the TTL cannot lapse into the old prompt between admission and navigation.
- `extension/src/worker.ts`: handler registration, TTL sweep alarm,
  clickable notification, best-effort caveat documented.
- `daemon.py`: timeout entries with rationale; access methods serialize on
  their own lock key so a 20 s await slice can't block another session's open.
- `backend.py`: `origin_not_allowed` → teaching error (notify-first).
- `tools/builtin.py`: actions + `timeout_s` param, validation, surface-free
  dispatch (they exist for the moment `open` just failed — requiring a surface
  would deadlock the recovery), per-state result texts, tool description
  teaches the dance.

## Testing evidence

**Unit/pure** (all green):
- `extension`: `pnpm typecheck` ✅, `pnpm test` **28/28** ✅, `pnpm build` ✅.
  Integration tests at the real origins.ts/state.ts/access-flow.ts seam cover:
  one popup click resolving both an in-command wait + async record; exactly-once
  grant consumption; cross-session refusal (refused attempt does not consume);
  raw-RPC command-id handoff; admission-time consumption; deny persistence
  across worker restart; supersession tombstones; spent-receipt/stale-tombstone
  cleanup. Pure suites cover TTL boundaries, idempotence and deny cool-down.
- `pytest tests/unit/browser_bridge`: **76/76** ✅; bridge + browser-tool slice:
  **180/180** ✅ (new: method↔timeout-table mirror, gen-ts coverage, daemon bound >
  extension slice, teaching-error content incl. NOTIFY THE USER, surface-free
  dispatch, cmux degrade, denied warns off retry, URL validation).
- Full `pytest tests/unit`, whole-tree pyright/flake8/black/isort: green
  (details in MR discussion if needed).

**End-to-end, real Chrome + real daemon** (throwaway headful Chrome 151
launched `open -g` + `--window-position=-3000,-3000` — never foregrounded;
built extension via CDP `Extensions.loadUnpacked`; throwaway daemon on port
4231 with `LOCAL_OPERATOR_CONFIG_DIR=/tmp/...`; paired via the real pair
event; decisions made by **clicking the real popup's Allow/Deny buttons** in a
background tab — the exact `origin_decision` message path a human click
takes). **24/24 checks** after the #321 multi-surface rebase:

```
PASS setup: extension connected+paired
PASS 1 open fails early with origin_not_allowed — code=origin_not_allowed in 0.01s
PASS 2 request_access returns pending
PASS 2 popup badge set to '!'
PASS 2 popup prompt record present
PASS 2 repeat request stays pending (idempotent)
PASS 3/4 await_access returns allowed after popup decision
PASS 4 badge cleared after decision
PASS 5 open succeeds after allow — {"tab": "bridge:…", "url": "http://127.0.0.1:8891/index.html", "title": "LOP Access E2E"}
PASS 5b once-grant consumed: next goto fails early again
PASS 5c A has a fresh pending prompt before supersession
PASS 5d B's different-origin request supersedes A's single prompt slot
PASS 5e displaced A reads explicit superseded
PASS 5f B's await returns allowed after real popup click
PASS 5g A cannot spend B's grant (origin_not_allowed; refused attempt did not consume)
PASS 5h B consumes its own grant into a second distinct surface
PASS 5i tabs lists both sessions' live surfaces
PASS 6 request_access on third origin pending
PASS 6 await_access returns denied
PASS 6 navigation still refused after deny
PASS 6b repeat request after deny answers denied (no nag)
PASS 7 request_access re-raises for consumed origin
PASS 7 expired request reads as none
PASS 7 expiry drops the stale prompt badge
```

## Coordination / round-1 rebase

Rebased semantically onto current `origin/main` (`f9dacfd9`, includes #320
client-timeout + #321 multi-tab surfaces):

- `open` preserves #321's two explicit modes (no tab → new capped surface;
  tab → resume exactly that surface); `goto` remains per-token; `tabs` remains
  awareness-only with redacted handles.
- daemon locks preserve #321's per-tab keys and shared `__global__` cap/listing
  key; access methods use a third `__access__` key because they touch neither
  tabs nor the surface map, and a 20 s human-await slice must not block another
  session's open.
- async requests and once-grants are bound to the tool's `session_id`; open/goto
  carry that same requester. A parallel session carries a different identity
  and cannot consume the grant. Raw-RPC callers get a command-id handoff.
- replace-don't-queue retains the single popup slot, but displaced requesters
  get a bounded tombstone receipt and read explicit `superseded` rather than a
  silent overwrite.

### DIFF
--- a/extension/src/access-flow.ts
+++ b/extension/src/access-flow.ts
@@ -24,12 +24,6 @@
 
 export type OriginDecision = "once" | "always" | "deny";
 
-/** Record decision column: a real user decision OR the "superseded" sentinel
- * written when a different requester's prompt replaces a pending one. Kept
- * distinct from OriginDecision so the user-facing vocabulary cannot leak
- * into wire types, while storage stays one string field. */
-export type RequestOutcome = OriginDecision | "superseded";
-
 /** What request_access/await_access report back to the agent. "none" means no
  * live request exists FOR THE CALLER (never raised, expired, or superseded);
  * "superseded" means a DIFFERENT requester's prompt replaced this origin's
@@ -69,9 +63,8 @@ export interface AccessRequest {
   requester: string;
   requestedAt: number;
   expiresAt: number;
-  /** Set once the request resolved: a user decision, or the "superseded"
-   * sentinel (see supersedeRequest); absent while the prompt is still open. */
-  decision?: RequestOutcome;
+  /** Set once the user decided; absent while the prompt is still open. */
+  decision?: OriginDecision;
 }
 
 /** A live "Allow once" grant, bound to the requester the user approved FOR
@@ -84,6 +77,28 @@ export interface OnceGrant {
 
 /** Origin -> grant, one per origin at most. */
 export type OnceGrants = Record<string, OnceGrant>;
+
+/** The live prompt slot holds ONE record (the popup's constraint), so a
+ * replace-don't-queue overwrite would otherwise erase the displaced request
+ * silently — the displaced requester's next poll would read "none",
+ * indistinguishable from expiry, and its agent would nag the user by
+ * re-raising the prompt in a steal loop (round-1 B1b). The tombstone map
+ * remembers RECENT replacements (bounded, same TTL discipline) so that
+ * requester learns "superseded" instead. Not a queue: the new prompt still
+ * replaces the old one immediately; this is only the receipt. */
+export interface SupersededTombstone {
+  origin: string;
+  requester: string;
+  /** Matches the displaced record's expiry: the receipt lives no longer
+   * than the request it stands for. */
+  expiresAt: number;
+}
+export type AccessTombstones = Record<string, SupersededTombstone>;
+
+/** Cap on retained tombstones: a churny fleet must not grow the map without
+ * bound; the oldest beyond the cap are dropped (a tombstone that old is
+ * telling its owner about a prompt the user has long forgotten anyway). */
+export const TOMBSTONE_CAP = 8;
 
 /** A record past its TTL reads as absent everywhere: await_access answers
  * "none" and request_access raises a fresh prompt. Expiry is computed on read
@@ -155,12 +170,12 @@ export function requestVerdict(
 }
 
 /** What await_access answers WITHOUT waiting. "pending" is the only state the
- * caller should then block on; everything else is final for this poll. The
- * tombstone records the record's own requester, so a supersession only reads
- * as "superseded" to the displaced requester — a third session asking about
- * the origin gets the neutral "none". */
+ * caller should then block on; everything else is final for this poll. A
+ * tombstone only reads as "superseded" to the requester it displaced — a
+ * third session asking about the origin gets the neutral "none". */
 export function accessState(
   record: AccessRequest | undefined,
+  tombstones: AccessTombstones | undefined,
   storedAllowed: boolean,
   onceGrant: boolean,
   origin: string,
@@ -169,12 +184,13 @@ export function accessState(
 ): AccessState {
   if (storedAllowed || onceGrant) return "allowed";
   const live = activeRequest(record, now);
-  if (!live || live.origin !== origin) return "none";
-  if (live.decision === "superseded") {
-    return live.requester === requester ? "superseded" : "none";
+  if (live && live.origin === origin) {
+    if (!live.decision) return "pending";
+    return live.decision === "deny" ? "denied" : "allowed";
   }
-  if (!live.decision) return "pending";
-  return live.decision === "deny" ? "denied" : "allowed";
+  const tomb = tombstones?.[origin];
+  if (tomb && now < tomb.expiresAt && tomb.requester === requester) return "superseded";
+  return "none";
 }
 
 export function newRequest(
@@ -186,11 +202,7 @@ export function newRequest(
   return { origin, hostname, requester, requestedAt: now, expiresAt: now + ACCESS_REQUEST_TTL_MS };
 }
 
-/** Tombstone a record being replaced: the origin and TTL are kept, the
- * decision becomes the sentinel "superseded", so the displaced requester's
- * next poll learns its prompt was taken over instead of reading the record
- * as simply gone. The record is NOT deleted — deleting would make the
- * displaced requester indistinguishable from a session that never asked. */
-export function supersedeRequest(record: AccessRequest): AccessRequest {
-  return { ...record, decision: "superseded" };
+/** The tombstone a replacement leaves for the displaced requester. */
+export function tombstoneFor(record: AccessRequest): SupersededTombstone {
+  return { origin: record.origin, requester: record.requester, expiresAt: record.expiresAt };
 }
--- a/extension/src/access-flow.ts
+++ b/extension/src/access-flow.ts
@@ -96,7 +96,18 @@ export interface SupersededTombstone {
    * than the request it stands for. */
   expiresAt: number;
 }
+
+/** Keyed by origin AND requester (see receiptKey): an A→B→C chain on one
+ * origin displaces two DIFFERENT requesters, and a per-origin key made C's
+ * receipt overwrite A's so A read "none" instead of "superseded" (round-2
+ * M1, reproduced by review). Each displaced requester owns its own receipt. */
 export type AccessTombstones = Record<string, SupersededTombstone>;
+
+export function receiptKey(origin: string, requester: string): string {
+  // \n cannot appear in an origin or a requester id, so the composite is
+  // collision-free without a structured key.
+  return `${origin}\n${requester}`;
+}
 
 /** Cap on retained tombstones: a churny fleet must not grow the map without
  * bound; the oldest beyond the cap are dropped (a tombstone that old is
@@ -167,9 +178,15 @@ export function requestVerdict(
   if (storedAllowed || onceGrant) return "allowed";
   const live = activeRequest(record, now);
   if (!live || live.origin !== origin) return "raise";
+  // A resolved record only answers ITS OWN requester (round-2 M1: B reading
+  // A's "once" as allowed told B it could navigate a grant it cannot spend,
+  // and B reading A's deny as denied hid that B never asked). Another
+  // requester's resolved record is displaced by a fresh raise; an undecided
+  // one is also displaced (replace-don't-queue), leaving a receipt.
+  if (live.requester !== requester) return "raise";
   if (live.decision === "deny") return "denied";
   if (live.decision) return "allowed";
-  return live.requester === requester ? "pending" : "raise";
+  return "pending";
 }
 
 /** What await_access answers WITHOUT waiting. "pending" is the only state the
@@ -187,12 +204,16 @@ export function accessState(
 ): AccessState {
   if (storedAllowed || onceGrant) return "allowed";
   const live = activeRequest(record, now);
-  if (live && live.origin === origin) {
+  // Only the record's OWN requester reads its live state (round-2 M1): after
+  // B replaces A on the same origin, A must read its receipt (superseded),
+  // not B's pending; after A's once-decision, B must not read allowed for a
+  // grant only A can spend.
+  if (live && live.origin === origin && live.requester === requester) {
     if (!live.decision) return "pending";
     return live.decision === "deny" ? "denied" : "allowed";
   }
-  const tomb = tombstones?.[origin];
-  if (tomb && now < tomb.expiresAt && tomb.requester === requester) return "superseded";
+  const tomb = tombstones?.[receiptKey(origin, requester)];
+  if (tomb && now < tomb.expiresAt) return "superseded";
   return "none";
 }
 
--- a/extension/src/access-flow.ts
+++ b/extension/src/access-flow.ts
@@ -69,10 +69,13 @@ export interface AccessRequest {
 
 /** A live "Allow once" grant, bound to the requester the user approved FOR
  * (see the multi-surface constraint at the top): only a navigation carrying
- * that requester may consume it. */
+ * that requester may consume it. `handoff` records the command id that raised
+ * the request, so a raw-RPC caller (no session identity) can still spend its
+ * own grant when its navigation reuses that id. */
 export interface OnceGrant {
   expiresAt: number;
   requester: string;
+  handoff?: string;
 }
 
 /** Origin -> grant, one per origin at most. */
--- a/extension/src/access-flow.ts
+++ b/extension/src/access-flow.ts
@@ -1,0 +1,196 @@
+/* Pure state machine for the async site-access flow, DOM- and chrome-free for
+ * the pure-node test suite — the same pattern as pair-flow.ts/origin-flow.ts.
+ *
+ * Why this flow exists at all: the original design answered a first visit to a
+ * new origin by BLOCKING the open/goto RPC inside the 60 s popup prompt. The
+ * agent never got a turn to tell the user a prompt was up, so prompts expired
+ * unseen (auto-deny), the daemon meanwhile extended the command's deadline,
+ * and the session client timed out with a misleading "bridge unreachable" — a
+ * real session burned an hour misdiagnosing the bridge as broken. The fix is
+ * to make approval a first-class three-step dance the agent can narrate:
+ * open fails early → request_access raises the prompt and returns → the agent
+ * TELLS the user → await_access waits for the decision.
+ *
+ * Multi-surface constraint (round-1 review, B1): parallel sessions now each
+ * drive their OWN tab, and this state is extension-global. Everything that
+ * carries authority — a pending request, a once-grant — is therefore bound to
+ * the REQUESTER (the daemon's per-command request id, which one session owns
+ * and another cannot guess). A grant minted for session A's flow can never be
+ * spent by session B's navigation: consumption fails CLOSED when the caller
+ * cannot prove it asked. A shared-grant model was considered and rejected:
+ * "Allow once" is the user's answer to "should THIS agent open THIS site",
+ * and letting a different agent ride it turns one approval into a fleet-wide
+ * permission — indistinguishable from "always" for the other sessions. */
+
+export type OriginDecision = "once" | "always" | "deny";
+
+/** Record decision column: a real user decision OR the "superseded" sentinel
+ * written when a different requester's prompt replaces a pending one. Kept
+ * distinct from OriginDecision so the user-facing vocabulary cannot leak
+ * into wire types, while storage stays one string field. */
+export type RequestOutcome = OriginDecision | "superseded";
+
+/** What request_access/await_access report back to the agent. "none" means no
+ * live request exists FOR THE CALLER (never raised, expired, or superseded);
+ * "superseded" means a DIFFERENT requester's prompt replaced this origin's
+ * pending request — an explicit verdict so the displaced requester learns
+ * what happened instead of timing out into the generic "none" (round-1 B1b).
+ * The recovery for both is the same: call request_access again. */
+export type AccessState = "allowed" | "denied" | "pending" | "superseded" | "none";
+
+/** How long an async access request stays answerable. The in-command prompt
+ * gave the user 60 s because someone was assumed to be watching the popup;
+ * here the user has been notified ASYNCHRONOUSLY (badge + best-effort system
+ * notification + the agent messaging them through the harness) and may
+ * reasonably take minutes to come back to the machine, so the window is 10
+ * minutes. It is still bounded because an unanswered prompt left open forever
+ * would let a long-forgotten Allow click grant a navigation nobody remembers
+ * requesting. */
+export const ACCESS_REQUEST_TTL_MS = 10 * 60_000;
+
+/** How long an unconsumed "Allow once" grant from the async flow survives.
+ * In the in-command flow "once" covered the navigation that was already in
+ * flight; here the navigation happens on the agent's NEXT open/goto, which
+ * may be a turn or two away, so the grant must outlive the approval moment —
+ * but not the browser session, and not long enough to surprise the user with
+ * a navigation they approved an afternoon ago. Same 10-minute bound as the
+ * request itself, consumed by the first navigation to the origin. */
+export const ONCE_GRANT_TTL_MS = 10 * 60_000;
+
+/** The persisted async request record. Lives in chrome.storage.session (not
+ * worker memory) because MV3 kills the service worker between the user's
+ * decision and the agent's next await_access poll; the record must survive
+ * that death or a granted approval would read as "none". */
+export interface AccessRequest {
+  origin: string;
+  hostname: string;
+  /** The daemon request id that raised this request; also minted onto any
+   * "once" grant the decision produces, binding the grant to this requester. */
+  requester: string;
+  requestedAt: number;
+  expiresAt: number;
+  /** Set once the request resolved: a user decision, or the "superseded"
+   * sentinel (see supersedeRequest); absent while the prompt is still open. */
+  decision?: RequestOutcome;
+}
+
+/** A live "Allow once" grant, bound to the requester the user approved FOR
+ * (see the multi-surface constraint at the top): only a navigation carrying
+ * that requester may consume it. */
+export interface OnceGrant {
+  expiresAt: number;
+  requester: string;
+}
+
+/** Origin -> grant, one per origin at most. */
+export type OnceGrants = Record<string, OnceGrant>;
+
+/** A record past its TTL reads as absent everywhere: await_access answers
+ * "none" and request_access raises a fresh prompt. Expiry is computed on read
+ * rather than by a trusted timer because worker death cancels timers. */
+export function activeRequest(
+  record: AccessRequest | undefined,
+  now: number,
+): AccessRequest | undefined {
+  if (!record) return undefined;
+  return now < record.expiresAt ? record : undefined;
+}
+
+/** A live grant this requester may consume. Fail-CLOSED: a grant whose
+ * requester does not match the caller — or a caller with no requester at all —
+ * is not this caller's grant, even if the origin matches. */
+export function consumableGrant(
+  grants: OnceGrants | undefined,
+  origin: string,
+  requester: string,
+  now: number,
+): OnceGrant | undefined {
+  const grant = grants?.[origin];
+  if (!grant || now >= grant.expiresAt) return undefined;
+  if (!requester || grant.requester !== requester) return undefined;
+  return grant;
+}
+
+/** Whether ANY live grant exists for the origin (regardless of requester).
+ * Used for display/diagnostics only — never as an admission decision. */
+export function anyGrantLive(
+  grants: OnceGrants | undefined,
+  origin: string,
+  now: number,
+): boolean {
+  const grant = grants?.[origin];
+  return !!grant && now < grant.expiresAt;
+}
+
+/** What request_access should do. Single-slot semantics: the popup and the
+ * pendingOrigin session record only ever show ONE origin, and the in-command
+ * flow already lets the last writer win that slot, so a request for a
+ * DIFFERENT origin REPLACES the live one rather than queueing — a queue
+ * would show the user prompts in an order the agents no longer care about.
+ * The displaced requester learns about it: their next request_access or
+ * await_access answers "superseded" (their record is tombstoned below), not
+ * a silent "none" they could misread as expiry (round-1 B1b).
+ * A repeat request for the SAME pending origin BY THE SAME requester is
+ * idempotent ("pending", original TTL kept — resetting it would let a polling
+ * agent extend the window forever). The same origin re-requested by a
+ * DIFFERENT requester also replaces: two sessions genuinely racing one site
+ * get the newest prompt, and the earlier requester reads "superseded". A fresh
+ * "deny" answers "denied" without re-prompting: the user just said no, and
+ * re-raising the prompt on every agent retry is a nag; the record's TTL is
+ * the cool-down after which a deliberate re-ask is allowed. */
+export function requestVerdict(
+  record: AccessRequest | undefined,
+  storedAllowed: boolean,
+  onceGrant: boolean,
+  origin: string,
+  requester: string,
+  now: number,
+): "allowed" | "pending" | "denied" | "raise" {
+  if (storedAllowed || onceGrant) return "allowed";
+  const live = activeRequest(record, now);
+  if (!live || live.origin !== origin) return "raise";
+  if (live.decision === "deny") return "denied";
+  if (live.decision) return "allowed";
+  return live.requester === requester ? "pending" : "raise";
+}
+
+/** What await_access answers WITHOUT waiting. "pending" is the only state the
+ * caller should then block on; everything else is final for this poll. The
+ * tombstone records the record's own requester, so a supersession only reads
+ * as "superseded" to the displaced requester — a third session asking about
+ * the origin gets the neutral "none". */
+export function accessState(
+  record: AccessRequest | undefined,
+  storedAllowed: boolean,
+  onceGrant: boolean,
+  origin: string,
+  requester: string,
+  now: number,
+): AccessState {
+  if (storedAllowed || onceGrant) return "allowed";
+  const live = activeRequest(record, now);
+  if (!live || live.origin !== origin) return "none";
+  if (live.decision === "superseded") {
+    return live.requester === requester ? "superseded" : "none";
+  }
+  if (!live.decision) return "pending";
+  return live.decision === "deny" ? "denied" : "allowed";
+}
+
+export function newRequest(
+  origin: string,
+  hostname: string,
+  requester: string,
+  now: number,
+): AccessRequest {
+  return { origin, hostname, requester, requestedAt: now, expiresAt: now + ACCESS_REQUEST_TTL_MS };
+}
+
+/** Tombstone a record being replaced: the origin and TTL are kept, the
+ * decision becomes the sentinel "superseded", so the displaced requester's
+ * next poll learns its prompt was taken over instead of reading the record
+ * as simply gone. The record is NOT deleted — deleting would make the
+ * displaced requester indistinguishable from a session that never asked. */
+export function supersedeRequest(record: AccessRequest): AccessRequest {
+  return { ...record, decision: "superseded" };
+}
--- a/extension/src/commands/access.ts
+++ b/extension/src/commands/access.ts
@@ -27,9 +27,10 @@ import { getSession } from "../state";
  * multi-minute promise alive across a service-worker death — each slice is
  * short, and the decision itself lives in session storage, which survives.
  *
- * Every verdict is computed against the daemon's request id (the handler's
- * `requestId`), which is what binds requests and grants to their requester
- * across parallel sessions — see access-flow.ts. */
+ * Every verdict is computed against a REQUESTER identity: params.requester
+ * when the calling tool supplies one (the session-scoped identity — see
+ * access-flow.ts), else the daemon's per-command request id as the fallback
+ * for raw-RPC callers. */
 
 /** One await_access slice. Below the daemon's 25 s command timeout with margin
  * so the RPC always answers before the daemon gives up on it. */
@@ -38,6 +39,17 @@ export const AWAIT_SLICE_MS = 20_000;
 /** How often a slice re-reads the decision record. Storage reads are cheap
  * (session storage is in-memory) and the human is the latency floor anyway. */
 const POLL_MS = 300;
+
+/** The requester identity for this command: the tool passes a session-scoped
+ * value (session:<id>) so all of one session's commands share it — a
+ * per-command request id could never match between request_access and the
+ * later open/goto, which is what would make grants unspendable. Raw-RPC
+ * callers (no session concept) fall back to the per-command id and get
+ * per-command binding, the fail-closed default. */
+function requesterOf(params: Record<string, unknown>, requestId: string): string {
+  const supplied = typeof params.requester === "string" ? params.requester.trim() : "";
+  return supplied || requestId;
+}
 
 /** Re-read the pieces an access verdict needs in ONE getSession round-trip,
  * sweeping TTL only when it can actually fire. expireAccessRequest does two
@@ -72,19 +84,20 @@ export async function requestAccess(
   requestId: string,
 ): Promise<Record<string, unknown>> {
   const url = safeHttpUrl(params.url);
+  const requester = requesterOf(params, requestId);
   await expireAccessRequest();
   const now = Date.now();
   const session = await getSession();
   const verdict = requestVerdict(
     session.accessRequest,
     await originAllowed(url),
-    !!consumableGrant(session.onceGrants, url.origin, requestId, now),
+    !!consumableGrant(session.onceGrants, url.origin, requester, now),
     url.origin,
-    requestId,
+    requester,
     now,
   );
   if (verdict !== "raise") return { origin: url.origin, state: verdict };
-  const record = await raiseAccessRequest(url, requestId);
+  const record = await raiseAccessRequest(url, requester);
   // Arm the TTL sweep for THIS record: chrome.alarms survives MV3 worker
   // death, a setTimeout does not. Creating a same-named alarm replaces the
   // prior one, which is intended — the record this alarm covers is now the
@@ -98,6 +111,7 @@ export async function awaitAccess(
   requestId: string,
 ): Promise<Record<string, unknown>> {
   const url = safeHttpUrl(params.url);
+  const requester = requesterOf(params, requestId);
   const requested = Number(params.timeout_ms);
   const sliceMs = Math.min(
     Number.isFinite(requested) && requested > 0 ? requested : AWAIT_SLICE_MS,
@@ -105,7 +119,7 @@ export async function awaitAccess(
   );
   const deadline = Date.now() + sliceMs;
   for (;;) {
-    const current = await pollSnapshot(requestId, url.origin);
+    const current = await pollSnapshot(requester, url.origin);
     if (current.state !== "pending" || Date.now() >= deadline) return current;
     await new Promise((resolve) => setTimeout(resolve, POLL_MS));
   }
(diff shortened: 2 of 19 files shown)


damianvtran:
### Agent review — round 1

**Reviewer:** reviewer subagent (model family: claude-fable-5)
**Scope:** 7b42eccd360cd99eeff197acefcd649e9526a9ce..e645e07fe67e7556d7b806a8b52a7606b77ff750

#### Verified locally

- `pnpm typecheck` / `pnpm test` (17/17 pure tests) / `pnpm build` — green
- `pytest tests/unit/browser_bridge` — 62 passed (incl. new protocol wiring, teaching-error, and tool-selection tests)
- `python -m local_operator.browser_bridge.gen_ts --check` — exit 0; `protocol.gen.ts` regen is exact
- Merge-base vs current `origin/main` is `7b42eccd` (pre-#320/#321); branch is **not** rebased on `f9dacfd9`

#### Findings

**B1 — BLOCKER — branch predates #321 multi-tab and #320 client-timeout; the rebase is semantic, not mechanical.** The branch's `open()` still uses the pre-#321 single-`surface` model (`extension/src/commands/nav.ts`, the `const previous = (await getSession()).surface` block at the entry-gate insertion, ~line 55 on branch) and the daemon still routes locks by `params.get("tab")` (`daemon.py:524`). Main redesigned both: surfaces are keyed by token, `open` reuses the caller's surface by token with a MAX_SURFACES cap, and `requireSurface(params.tab)` resolves tokens. Conflicts will be real, and two cross-cutting interactions must be re-decided during the rebase, not auto-resolved:
- The `__access__` lock key must be reconciled with #321's `__global__` open lock — as written the branch replaces the tab-key line that #320/#321's merge already rewrote.
- The single `pendingOrigin` slot and the once-grant map are shared across ALL sessions. With parallel sessions now owning a tab each, session A can consume a once-grant minted for session B's origin request (`origins.ts` `consumeOnceGrant`), and one session's `request_access` replaces another session's live prompt (replace-don't-queue now crosses session boundaries). If intentional, it needs a comment and a daemon-side note; if not, grants/requests need a surface-token dimension. The grant itself is also cross-session honored in `navigate()`.

**M1 — MAJOR — once-grant consumed without re-checking expiry in `navigate()`; combined with B1's grant sharing, a grant can ride a navigation the user did not approve.** `extension/src/commands/nav.ts` (branch `navigate()`, first lines): `if (!(await originAllowed(url)) && !(await consumeOnceGrant(url)))`. `consumeOnceGrant` (`origins.ts:40-47`) checks `onceGrantActive` at consumption time — good — but `ensureTopLevelAccess` at command entry ALSO consults the grant, and between the two reads the 10-min TTL can lapse; in that window `navigate` falls into `askOrigin` and raises a prompt for a navigation that is already in flight, which is the exact "prompt raised after the agent thinks it has access" surprise the flow was built to kill. Narrow race, but it is the only path where a top-level navigation can still block on the old 60 s prompt after a successful `await_access`. One-line fix: have `ensureTopLevelAccess` (or `goto`/`open`) return/record whether it admitted via the once-grant, and pass that to `navigate` so consumption and admission are the same decision (or re-run `ensureTopLevelAccess`'s exact check inside `navigate` and consume only on the same value).

**M2 — MAJOR — no test exercises the approve/deny/once-consume path through `resolveOrigin` → `recordAccessDecision` → `consumeOnceGrant` together; the exact-once property is only asserted for the pure helpers.** The new pure tests (`extension/tests/pure.test.mjs:132-196`) cover `requestVerdict`/`accessState`/`onceGrantActive` well, and the Python slice covers wiring and texts. But the interactions the PR description calls out as design-critical — a decision landing on BOTH an in-command `waiting` resolver and the async record; once-grant minted on async "Allow once" and consumed by exactly one navigation; deny cool-down surviving worker restart via session storage — have no test at the `origins.ts` seam (they can't, that file is chrome-coupled, but nothing else stands in). The `test_request_access_*`/`test_await_access_*` Python tests mock `_bridge_call`, so they prove the loop/texts, not the flow. At minimum an integration-style node test driving `resolveOrigin` with a stubbed `chrome.storage.session` would catch a regression in the dual-resolve path; as-is, a future edit to `resolveOrigin` or `recordAccessDecision` can silently break the async path with all suites green.

**m1 — MINOR — `test_every_method_has_a_timeout_and_vice_versa` duplicates what #320's merged tests now assert, and will conflict.** `tests/unit/browser_bridge/test_protocol.py:73` — main's `test_protocol.py` was extended by #320 with its own method/timeout coverage (29 changed lines there); expect a real conflict on rebase. Keep one version.

**m2 — MINOR — `request_access` arms `lop-access-expiry` with `when: record.expiresAt` but a same-origin re-request after a deny keeps the ORIGINAL `expiresAt` while the alarm may already have fired.** `extension/src/commands/access.ts` (`requestAccess`): on verdict `"pending"` (idempotent repeat) no alarm is re-created — correct, the original alarm is still pending. But after the TTL alarm fires and `expireAccessRequest` clears the record, an immediate `request_access` for the same origin creates a fresh record and a fresh alarm — fine. The gap: `chrome.alarms.create` with the same name REPLACES the prior alarm, so a replace-don't-queue request for origin B silently drops origin A's expiry alarm; A's badge/prompt then only clears on the next lazy sweep (any access read). Cosmetic since `snapshot()` sweeps on every read, but the badge can outlive A's record until something polls. Consider re-arming for the earliest live record on sweep.

**m3 — MINOR — `snapshot()` in `awaitAccess` runs the full TTL sweep + storage reads every 300 ms poll.** `extension/src/commands/access.ts` (`snapshot`): each poll does `expireAccessRequest()` (2 storage reads) + `getSession()` + `originAllowed()` (a `chrome.storage.local` read). Local reads are cheap, but a 20 s slice makes ~66 full sweeps; reading `accessRequest`/`onceGrants` from one `getSession()` and skipping `expireAccessRequest` except when the record is actually expired would be the same correctness at ~1/3 the storage traffic. Not worth a round by itself.

**n1 — NIT — `_access_result_text` "none" branch says "may have expired unanswered" but the state also means "superseded by a different origin's request" and "never raised"** (`local_operator/tools/builtin.py`, `_access_result_text`). The agent recovery is identical (re-request), so this is wording only.

**n2 — NIT — popup "Allow once" label vs 10-min grant TTL.** `extension/src/popup/popup.html:152` still reads "Allow once"; the grant now lives up to 10 minutes unconsumed (`access-flow.ts` ONCE_GRANT_TTL_MS). Intent is documented in the code, but the store-reviewer justification that promised prompts/one-shot grants should be re-read against the new semantics — worth one line in the PR description for whoever owns the listing.

#### Notes (checked, no finding)

- State machine TTL boundaries, deny cool-down, and replace-don't-queue verdicts are correct per the pure tests; re-request for the same pending origin keeps the original TTL (no polling-extension), verified in `requestVerdict` tests.
- `resolveOrigin` resolving both the in-command wait and the async record: decision can be consumed twice (an in-command `askOrigin` resolves AND `recordAccessDecision` persists), but the async write is guarded (`live.decision` already set → no-op) and the in-command path clears its own prompt — no lost decision found. The `getSession`/`set` race in `recordAccessDecision` is bounded by the popup emitting one decision per origin.
- `ensureTopLevelAccess` runs before tab creation in `open` and before `requireSurface`-driven navigation in `goto`; it consults the same `origins` map via `originAllowed` — no duplicated allowlist.
- `__access__` lock key cannot interleave with `__global__` on this branch's daemon (distinct keys, distinct locks); starvation is bounded by the 25 s command timeout. Interaction with #321's lock model is folded into B1.
- Agent-facing texts teach request → notify → await in the right order and only appear in tool results/errors, not user surfaces.

#### Verdict

NOT clean — B1 must be resolved (rebase onto `f9dacfd9` and re-decide the cross-session grant/lock semantics it surfaces), and M1/M2 are real defects/gaps in the mechanism this PR exists to make trustworthy. Re-review needed after the rebase; m1–m3 and nits are follow-ups.

damianvtran:

**Reviewed head:** `e645e07fe67e7556d7b806a8b52a7606b77ff750`
**Remediated head:** `1a228a81a69f7e1371cfce7f436779d6626a734b`
**Rebased base:** `f9dacfd9c08a87b4e88b7b2ecb9b269ff135b759` (includes #320 + #321)

All findings addressed; nothing deferred.

- **B1 — fixed** — semantic rebase onto current `origin/main` in `bbbbdcd`.
  - Preserves #321's multi-surface `open` modes (fresh = new capped surface; `tab` = resume exact token), per-token `goto`, redacted `tabs`, and surface-map state.
  - `__access__` is integrated as a third daemon lock class in `bbbbdcd`: per-tab locks remain parallel; `__global__` still protects fresh-open cap/listing; access methods touch neither and therefore do not hold `__global__` for a 20 s human wait.
  - **Cross-session grant decision:** requester-bound, not shared. `a0257f80` introduced the fail-closed model; `ce24abaa` corrected the binding identity to the tool's stable `session_id` (per-command request ids cannot match across request→navigate), attaches it to request/await/open/goto, and adds a raw-RPC command-id handoff. A parallel session carries neither matching identity and cannot spend the grant; its refused attempt does not consume. Integration and E2E coverage below.
  - **Cross-session supersession decision:** keep the popup's one-slot replace-don't-queue constraint, but never silently overwrite. `a0257f80` adds a bounded tombstone receipt: the displaced requester reads explicit `superseded`; a third session reads `none`; the new prompt surfaces replace the old immediately. `1a228a81` consumes stale receipts when the displaced requester deliberately re-requests.
- **M1 — fixed** (`a0257f80`, corrected binding in `ce24abaa`) — `ensureTopLevelAccess` now makes admission + consumption ONE decision at command entry and returns `OriginAdmission`; `navigate()` trusts that exact decision and never re-checks the grant. The TTL cannot lapse into the old 60 s prompt between admission and navigation. Integration test: `admission consumes the grant once and navigate never re-consults it (M1)`.
- **M2 — fixed** (`a0257f80`, expanded `ce24abaa` + `1a228a81`) — new `extension/tests/origins.integration.test.mjs` drives the real origins.ts/state.ts/access-flow.ts seam with stubbed chrome session/local storage:
  - one popup decision resolves BOTH an in-command `waiting` promise and the async record;
  - exactly-once requester-bound grant consumption;
  - parallel-session cross-consume refusal without consuming;
  - raw-RPC command-id handoff;
  - admission-time consumption;
  - deny cool-down persistence across a simulated worker restart;
  - supersession tombstones;
  - spent-receipt and stale-tombstone cleanup.
- **m1 — fixed** (`a0257f80`) — removed `test_every_method_has_a_timeout_and_vice_versa`; main's #320/#321 `test_tabs_is_a_protocol_method_with_a_daemon_timeout` remains the single METHODS↔COMMAND_TIMEOUTS mirror pin.
- **m2 — fixed** (`a0257f80`, lifecycle hardening `1a228a81`) — supersession raises the new prompt/badge/title immediately, overwriting the replaced surfaces; `chrome.alarms.create` is re-armed for the only live record. Spent once receipts and stale supersession receipts are now removed at their lifecycle boundaries.
- **m3 — fixed** (`a0257f80`) — the 300 ms await poll does one `getSession` + one local allowlist read and calls the heavier TTL sweep only when the record is actually expired (instead of full sweep + repeated session reads every tick).
- **n1 — fixed** (`a0257f80`) — `superseded` has explicit agent-facing guidance; `none` now covers expired/never-raised without pretending all absence is timeout.
- **n2 — fixed** (`a0257f80`) — button stays **Allow once** (the store-promised one-shot vocabulary); its acknowledgement now states the actual async semantics: “The agent's next visit to this site goes through (once, within 10 minutes).” Pure test pins both `next visit` and `10 minutes`.

#### Verification on remediated head

**Full requested bar**
- `pnpm typecheck` ✅
- `pnpm test` ✅ — **28/28**
- `pnpm build` ✅
- `python -m local_operator.browser_bridge.gen_ts --check` ✅
- `pytest tests/unit/browser_bridge -q` ✅ — **76/76**
- bridge + browser tool slice ✅ — **180/180**
- whole-tree `flake8 .` ✅
- whole-tree `black --check .` ✅ — 552 files unchanged
- whole-tree `isort --check .` ✅
- `pyright local_operator tests` ✅ — 0 errors/warnings

**Real-path E2E — 24/24**
- Throwaway headful Chrome 151 via `open -g` + `--window-position=-3000,-3000` (never activated); built extension loaded by CDP `Extensions.loadUnpacked`; throwaway daemon port 4231 + config dir; paired through real pair event; decisions via clicks on the real popup Allow/Deny buttons.
- Covered: early typed error (0.01 s); pending badge + prompt record; idempotent repeat; approve→await allowed→open; once grant consumed exactly once; session A fresh pending → session B different-origin request → A reads explicit `superseded`; B approved; A's cross-consume refused with `origin_not_allowed` and did not consume; B consumed its own grant into a **second distinct extension surface**; `tabs` listed both surfaces; third-origin deny + no-nag cool-down; TTL expiry → `none`; stale badge removed.

No merge performed. Ready for round 2 on `1a228a81a69f7e1371cfce7f436779d6626a734b`.

damianvtran:
### Agent review — round 2

Reviewer: independent reviewer agent — openai/gpt-5.6-sol
Scope: e645e07f..1a228a81 (delta after rebase), verified at head 1a228a81

#### Findings

1. **BLOCKER — A stale popup can approve a different origin than the one shown.** `raiseAccessRequest` replaces `pendingOrigin` (`extension/src/origins.ts:261-289`), but an already-open popup only listens for `connState` changes, not `pendingOrigin` (`extension/src/popup/popup.ts:365-372`). On click, `decide()` re-reads the *current* `pendingOrigin` and sends that origin (`popup.ts:338-360`) rather than the `shownPromptOrigin` whose buttons the user clicked. Reproduction path: render A → session B replaces slot with B → popup still displays A → click “Allow once” → B is granted. This violates the consent boundary. **Fix:** bind the click to the rendered request (ideally an immutable request id + origin), invalidate/re-render on replacement, and reject a decision if that identity is no longer current.

2. **BLOCKER — once-grant consumption is still a read/delete race and can double-spend.** `consumeGrantFor` reads the grant, checks it, then separately writes a map with the entry deleted (`extension/src/origins.ts:103-118`). `goto`s for different #321 surfaces use different daemon locks, and the worker dispatches frames concurrently (`extension/src/worker.ts:202-205`), so two same-session navigations can both read the grant before either write lands. I reproduced this against the real bundled `origins.ts` with realistic async `chrome.storage.session` get/set: `Promise.all([ensureTopLevelAccess(cmd-1), ensureTopLevelAccess(cmd-2)])` returned two `{allowed:true, viaOnceGrant:true}` results. There is no concurrent-consume test. **Fix:** serialize admission/consume on an origin/session authority lock (not per tab) or use an atomic authority owner; add a deterministic two-navigation race test.

3. **MAJOR — same-origin supersession is not requester-correct, and a third request destroys the first receipt.** For a live record on the same origin, `accessState` returns `pending`/`allowed`/`denied` without checking `record.requester` (`extension/src/access-flow.ts:179-193`), while `requestVerdict` treats any resolved decision as allowed/denied for every requester (`access-flow.ts:159-172`). Thus after B replaces A on the same origin, A reads B's `pending`, not `superseded`; after A's once decision, B is told `allowed` although B cannot consume the grant. Also `AccessTombstones` is keyed only by origin (`access-flow.ts:92-104`), so A→B→C on one origin overwrites A's receipt with B's (`origins.ts:271-282`). I reproduced both: B saw `allowed` from A's resolved record, and the third request replaced A's tombstone. The integration suite only covers a different-origin replacement. **Fix:** make live verdicts requester-aware and key bounded receipts by requester/request identity, with A→B→C same-origin tests through `requestAccess`/`awaitAccess`.

4. **MAJOR — the popup decision's persistence is fire-and-forget and is not kept alive across worker suspension.** `resolveOrigin()` starts `recordAccessDecision()` with `void` (`extension/src/origins.ts:225-234`), and the `chrome.runtime.onMessage` listener returns nothing (`extension/src/worker.ts:269-273`). For async approval there may be no in-memory waiter; Chrome may settle `sendMessage` and suspend the MV3 worker before `accessRequest`, `onceGrants`/`origins`, and prompt cleanup finish. The “worker restart” test flushes the write before re-importing, so it does not exercise interruption here. **Fix:** return/await the decision Promise from the message listener so the event lifetime covers durable state and cleanup; test restart/suspension at this seam.

5. **MAJOR — `await_access` holds the global `__access__` lock for the full 20-second human-poll slice.** Both `request_access` and `await_access` share `__access__`, and the lock wraps the entire RPC (`local_operator/browser_bridge/daemon.py:496-508`). One waiting session therefore prevents another session from raising/replacing a prompt until its slice returns; multiple waiters can queue in 20-second increments before their own daemon timeout even starts. This defeats the parallel-session supersession design and can push the HTTP caller into timeout. **Fix:** do not hold the mutation lock while waiting; lock only short state transitions/reads or use per-request eventing, and add a concurrency test proving a second session's `request_access` completes promptly during another session's await.

6. **MAJOR — the required real harness identity path still has no integration proof.** The Python tool test constructs `ToolContext` without `session_id` and asserts the fallback `call:t` (`tests/unit/browser_bridge/test_tool_selection.py:380-399`); the 24/24 E2E bypasses `execute_browser` and manually inserts `requester: session:A/B` in raw RPC (`/tmp/lop-access-e2e/e2e.py:247-261`, etc.). The production plumbing itself appears correct (`ToolContext.session_id` → `_browser_requester` → every access/open/goto wire), but no test would fail if one action reverted to per-tool-call identity. **Fix:** exercise `execute_browser` with one `ToolContext(session_id=...)` and different tool-call IDs across request_access → await_access → open/goto, plus a parallel context, through the real bridge seam.

#### Remediations verified / checks run

- `f9dacfd9` is an ancestor of `1a228a81`; #321's token-bound fresh/resume/goto multi-surface model is preserved. The PR base has since advanced to `d4ba139e` (#328), which is not in the head, but this review found no approval-flow conflict from that unrelated base advance.
- Stable requester plumbing is present in production: session construction populates `ToolContext.session_id`, `_browser_requester` derives `session:<id>`, and request_access/await_access/open/goto carry it. Raw-RPC command-id handoff is implemented, but only direct-module tested.
- Polling now avoids the old repeated full expiry sweep; agent-facing strings retain explicit notify-before-await guidance; the “next visit … within 10 minutes” copy is truthful; Chrome notification remains best-effort/clickable.
- Protocol generation and timeout tables are exact; the 20 s extension slice is below the 25 s daemon bound.
- Read `/tmp/lop-access-e2e/round1-e2e.log`: 24/24 uses two logical requester strings through the real daemon RPC, distinct real surfaces, and real popup button clicks. It directly mutates storage only to force the final expiry edge. It does **not** exercise the Local Operator tool/context identity plumbing (finding 6).
- Fresh runs at `1a228a81`: `pnpm typecheck` ✅; `pnpm test` 28/28 ✅; `pnpm build` ✅; `python -m local_operator.browser_bridge.gen_ts --check` ✅; `pytest tests/unit/browser_bridge` 76/76 ✅; relevant tools + browser-tool slice 555/555 ✅ (warnings only).

**Verdict: not clean — non-terminal.** Two BLOCKER and four MAJOR findings remain; round 3 is required after remediation on a fresh head.

damianvtran:

**Reviewed head:** `1a228a81a69f7e1371cfce7f436779d6626a734b`
**Remediated head:** `26345f00` (one commit: `fix(browser): close round-2 consent, atomicity and concurrency findings`)

All six findings fixed; nothing rejected or deferred.

- **B1 (stale popup consent) — fixed** (`26345f00`) — Every prompt raise mints an immutable `promptId` generation (`origins.ts raisePrompt`, stored on `pendingOrigin`). The popup captures the generation **at render** (`shownPromptId`) and the click sends `shownPromptOrigin` + `shownPromptId` — it no longer re-reads current state, which was the consent hole. `resolveOrigin(origin, decision, promptId)` validates both against the live prompt and returns `applied:false` for a stale one: nothing resolves, no grant is minted, and the popup shows a neutral "Request changed" notice with live buttons for the current prompt instead of a false ack. The popup also re-renders on `pendingOrigin` storage changes so an open window tracks replacements.
  Tests: integration `a stale prompt generation cannot approve the replacement origin (B1)` covers A-displayed → B-replaces (different origin AND same-origin re-raise) → stale click ⇒ rejected, no grant, B's request undecided, current generation still decides. E2E check 8 proves the same against the **real worker**, sending the stale decision from a real popup page context: `applied:false`, `onceGrants` empty.
- **B2 (once-grant double-spend) — fixed** (`26345f00`) — All access-state read-modify-writes (`consumeOnceGrant`, `consumeGrantFor`, `recordAccessDecision`, `raiseAccessRequest`, `expireAccessRequest`) now run under `state.ts withSessionMutation` — the exact serialized-mutation promise-chain #321 uses for the surfaces map, exported rather than duplicated. Read-check-delete is atomic; exactly one navigation wins.
  Test: integration `two concurrent navigations cannot double-spend one grant (B2)` reruns the reviewer's reproduction (`Promise.allSettled` of two `ensureTopLevelAccess` calls): exactly 1 admitted, 1 refused with typed `origin_not_allowed`.
- **M1 (requester-correct supersession, A→B→C receipts) — fixed** (`26345f00`) — Live verdicts are requester-aware: `accessState`/`requestVerdict` answer a record only to **its own requester** (B no longer reads A's `pending`/`allowed`/`denied`; a foreign resolved record is `raise` for request, receipt-or-none for state). Receipts are keyed `origin\nrequester` (`receiptKey`), so an A→B→C chain preserves every displaced requester's receipt; bounded by the existing `TOMBSTONE_CAP` with expiry filtering.
  Tests: integration `A->B->C same-origin supersession keeps every displaced receipt (M1)` and the different-origin chain variant; pure tests pin the requester-aware verdicts and per-requester receipt reads (B reads `none` from A's resolved record; expiry ends the receipt).
- **M2 (fire-and-forget decision persistence) — fixed** (`26345f00`) — `resolveOrigin` is now async and settles only after `recordAccessDecision` completes (record + grant + allowlist + prompt teardown, all under the mutation queue). The worker's `origin_decision` listener uses `sendResponse` + `return true` so the MV3 event lifetime covers the durable write; no unhandled promise (failures answer `applied:false`).
  Test: integration `resolveOrigin settles only after the decision is durable (M2)` delays **every** storage write by 30 ms (suspension-shaped ordering) and asserts the decision and grant are readable the instant the promise settles.
- **M3 (`__access__` lock held for 20 s waits) — fixed** (`26345f00`) — `await_access` now dispatches under a **private per-request lock key** (`__await__:<id>`, evicted after use — request ids are unique, so nothing can ever queue behind a human wait). Its extension side only reads; every mutation is serialized worker-side by the B2 queue, so daemon-side serialization added nothing but queueing. `request_access` keeps the short shared `__access__` key: raise/replace is a read-modify-write of the single prompt slot and must stay deterministic, and it never waits on a human (ms hold). Lock-key selection is extracted to `BridgeService.lock_key_for` so the topology is directly testable.
  Tests: `test_await_access_lock_topology` (structural: two awaits never share a key; request/global/tab keys pinned) and `test_await_access_does_not_block_request_access` (behavioural: fake extension parks await_access 2 s; a concurrent `request_access` completes in <1 s — fails under the old shared key).
- **M4 (real harness identity path untested) — fixed** (`26345f00` + E2E) — New unit tests drive **`execute_browser` with real `ToolContext(session_id=...)`**: `test_session_identity_is_stable_across_the_whole_flow` (request_access → await_access → open under three different tool call ids, all presenting `session:sess-42`), `test_parallel_contexts_present_distinct_identities`, and `test_goto_carries_the_session_identity`. The E2E was **rewritten onto the tool path**: every scenario step calls `execute_browser` with two genuine `ToolContext` instances (`e2e-A`, `e2e-B`) and fresh tool-call ids per step; no injected requester strings remain. The wire identity is proven end-to-end: the prompt record carries `requestId: "session:e2e-A"`, and A's later `open` (different call id) consumes the grant.

#### Verification on remediated head `26345f00`

**Full bar**
- `pnpm typecheck` ✅ · `pnpm test` ✅ **33/33** (5 new round-2 suites) · `pnpm build` ✅
- `python -m local_operator.browser_bridge.gen_ts --check` ✅
- `pytest tests/unit/browser_bridge tests/unit/tools/test_browser_tool.py` ✅ **185/185**
- `flake8` / `black --check` / `isort --check` ✅ · `pyright local_operator tests` ✅ 0 errors

**Real-path E2E — 27/27** (throwaway headful Chrome 151, `open -g` off-screen, never activated; CDP `Extensions.loadUnpacked`; throwaway daemon port 4231 + config dir; decisions via clicks on the real popup buttons; **all agent-side steps through `execute_browser` with two real ToolContexts**):
early typed error (0.02 s) · pending + badge + prompt record carrying `session:e2e-A` and a `promptId` generation · idempotent repeat · popup-click approve → await allowed → open consumes the session-bound grant under a different tool call id · once-consume exact-once · A re-raise → B different-origin supersession → A reads explicit `superseded` · B approved, A's cross-spend refused fail-closed, B opens its **own second surface**, `tabs` lists both · third-origin deny + do-not-retry + no-nag cool-down · TTL expiry → `none`, badge dropped · **stale-generation decision from a real popup context rejected by the real worker (`applied:false`), no grant minted**.

No merge performed; no reviewer launched. Ready for round 3 on head `26345f00`.

damianvtran:
### Agent review — round 3

Reviewer: Local Operator independent reviewer — openai/gpt-5.6-sol
Scope: 1a228a81..26345f00 (delta), verified at head 26345f00

#### Findings

**R3-B1 — BLOCKER — Prompt-generation validation is outside the mutation queue and can be invalidated before the decision is recorded** (`extension/src/origins.ts:267-277`)

`resolveOrigin` reads and validates `pendingOrigin` before entering the serialized mutation queue (indirectly through `recordAccessDecision`). A same-origin replacement can already own that queue. The stale decision validates against the old prompt, then waits; the replacement installs its fresh request/generation; finally `recordAccessDecision` sees only that the new request has the same origin and applies the stale decision to it.

I reproduced this adversarially against the real bundled `origins.ts/state.ts/access-flow.ts` seam: A raised `https://example.com`; B began a same-origin replacement and was paused during its queued state write; A's old-generation click was delivered and validated; B was released to install its fresh generation; then the stale click returned `applied:true`, set B's request to `decision:"once"`, and minted B a grant. This violates B1's required “applied:false, no resolve, no grant” invariant.

Obvious fix: perform prompt origin+generation validation and the corresponding access-record/grant mutation in the same `withSessionMutation` critical section (using a non-locking internal decision helper to avoid nested queue acquisition), then resolve the in-command waiter only after that atomic validation succeeds.

**R3-M1 — MAJOR — Cancelled `await_access` dispatches leak their per-request lock keys** (`local_operator/browser_bridge/daemon.py:526-542`)

The `__await__:<id>` eviction runs only after `_dispatch_locked` returns normally. Cancellation at the `async with` propagates before line 541, permanently retaining the unique key. I reproduced by starting `_dispatch_serialized(await_access)`, cancelling it while parked, and inspecting `_tab_locks`: `['__await__:review-cancel']` remained afterward. Repeated client disconnects/cancellations therefore grow the daemon map without bound, contrary to M3's lock-eviction requirement.

Obvious fix: put per-await key eviction in a `finally` that surrounds the locked dispatch.

#### Verification

- `pnpm typecheck` ✅
- `pnpm test` ✅ — 33/33
- `pnpm build` ✅
- `python -m local_operator.browser_bridge.gen_ts --check` ✅
- `pytest tests/unit/browser_bridge tests/unit/tools/test_browser_tool.py -q` ✅ — 185/185
- B2 concurrent double-spend test: exactly one winner ✅
- Current-generation duplicate delivery: one consumable grant, no double-resolve/effect; both deliveries currently acknowledge `applied:true` ✅
- M1 requester-keyed same-origin and mixed A→B→C receipts, TTL/cap behavior ✅
- M2 delayed persistence and worker `sendResponse` + `return true` wiring ✅
- M3 request access progresses while another await is parked; `__global__` remains independent ✅
- M4 evidence: `/tmp/lop-access-e2e/round2-e2e.log` reports 27/27 through `execute_browser`; the harness uses distinct real `ToolContext(session_id=...)` instances and contains no injected requester argument ✅
- Popup storage listener is registered once at module scope; no per-render listener accumulation ✅

#### Verdict

**Not clean — non-terminal.** One BLOCKER and one MAJOR remain; round 3 cannot be terminal.

damianvtran:

**Reviewed head:** `26345f009662246c425cab7f0a9ffcf7df207f04`
**Remediated head:** `d1893535` (one commit: `fix(browser): close round-3 TOCTOU and lock-leak findings`)

Both findings fixed; nothing rejected or deferred.

- **R3-B1 (TOCTOU stale generation) — fixed** (`d1893535`) — The root cause was exactly as diagnosed: `resolveOrigin` validated `pendingOrigin` OUTSIDE the mutation queue and `recordAccessDecision` self-locked inside it, so a same-origin replacement queued between them installed its fresh generation after the stale click validated but before it applied. The fix restructures `resolveOrigin` to run entirely inside ONE `withSessionMutation` critical section: the pending prompt is re-read atomically with the mutation that consumes it, origin + promptId are compared there, and only then is the decision applied and the in-command waiter resolved. The old self-locking `recordAccessDecision` became a non-locking `recordAccessDecisionLocked` (all callers now already hold the queue — nested acquisition would deadlock), which also returns whether it actually applied. A duplicate delivery of the CURRENT generation is a safe no-op: the first decision consumed the prompt record, so the re-delivery finds nothing live and returns `applied:false` with zero side effects — the grant minted by the first click is untouched.
  Tests: `a queued same-origin replacement cannot be decided by a stale click (R3-B1 TOCTOU)` replays the reviewer's exact adversarial sequence — B's replacement paused *inside* its queued mutation via a storage-write gate, stale A click queued behind it → `applied:false`, B's request still undecided, no grant minted, and B's current generation still decides normally afterwards. `duplicate current-generation delivery is idempotent (R3-B1)` pins the no-op semantics.
- **R3-M1 (cancelled await lock leak) — fixed** (`d1893535`) — The await branch of `_dispatch_serialized` now wraps the locked dispatch in `try/finally` with **unconditional** `self._tab_locks.pop(tab_key, None)`, covering success, typed error, the daemon command timeout, AND cancellation (client disconnect → `CancelledError` inside the HTTP handler). The `finally` is safe to be unconditional because request ids are unique — no other request can ever share a per-request key.
  Tests: `test_cancelled_await_evicts_its_lock_key` starts a parked await (extension never answers), cancels the dispatch task mid-wait, and asserts `_tab_locks == {}` — the reviewer's reproduction shape. `test_await_lock_key_evicted_on_normal_and_error_exits` covers the other two exits: immediate answer, and the daemon timeout on a silent extension.

#### Verification on remediated head `d1893535`

**Full bar**
- `pnpm typecheck` ✅ · `pnpm test` ✅ **35/35** (2 new round-3 suites) · `pnpm build` ✅
- `python -m local_operator.browser_bridge.gen_ts --check` ✅
- `pytest tests/unit/browser_bridge tests/unit/tools/test_browser_tool.py` ✅ **187/187** (2 new daemon tests)
- `flake8` / `black --check` / `isort --check` ✅ · `pyright local_operator tests` ✅ 0 errors

**Real-path E2E — 27/27** (same throwaway Chrome 151 + daemon setup as round 2, rebuilt extension): the full approve/deny/TTL/supersession/cross-session/stale-generation scenario re-run against the new worker listener path — including check 8, where the stale-generation decision sent from a real popup page context is still rejected (`applied:false`, no grant minted) by the restructured atomic `resolveOrigin`.

No merge performed; no reviewer launched. Ready for round 4 on head `d1893535`.

damianvtran:
### Agent review — round 4

**Reviewer:** Local Operator independent reviewer subagent (model: kimi/k3, session fell back from claude-fable-5 / glm-5.3 mid-review due to provider quota — verification commands and analysis all run by the same reviewer instance)
**Scope:** `26345f00..d1893535` (delta), verified at head `d1893535`

Round 3's BLOCKER (stale same-origin generation TOCTOU) and MAJOR (cancelled `await_access` lock leak) are both closed. Verified adversarially, not by reading the remediation note.

---

#### 1. R3-B1 TOCTOU — closed, verified (`extension/src/origins.ts:273-298`)

The critical section is real:

- `resolveOrigin` now runs entirely inside ONE `withSessionMutation`: the pending prompt is re-read from `getSession()` *inside* the queue, `origin` and `promptId` are validated there, `recordAccessDecisionLocked` applies record + grant + allowlist + prompt teardown in the same critical section, and only then is the in-command waiter resolved (`waiting.get(origin)` → `resolve(decision)` at `:291-295`). A queued replacement cannot interleave between validation and mutation — it must wait its turn, and when the stale click's turn comes, its re-read sees the NEW generation and bails at `:288`.
- `raisePrompt` (`:156-161`) mints a fresh `crypto.randomUUID()` `promptId` on every (re)raise, so a same-origin replacement necessarily has a different generation — the `promptId !== pendingOrigin.promptId` guard is load-bearing and correct.
- **No nested acquisition / deadlock:** `recordAccessDecisionLocked` (`:215-251`) makes only plain `chrome.storage.session/local` calls and calls `clearPrompt` (`:170-175`), which is also plain storage + badge calls. Neither re-enters `withSessionMutation` (which is `withStore` in `state.ts:97-101` — a promise chain, so a true nested acquisition *would* self-deadlock; there is none). The waiter `resolve(decision)` at `:294` only *schedules* `askOrigin`'s continuation — its tail (`:194-198`: `clearPrompt` + allowlist write) never re-enters the queue either, and even if it did, it would run after the current op settled.
- **Queue hold time:** the critical section now contains the durable writes it used to contain (record/grant/allowlist) plus badge/title updates. These are the same writes the old self-locking `recordAccessDecision` performed inside its own `withSessionMutation` — total queue occupancy is unchanged, not extended.

The new integration test `a queued same-origin replacement cannot be decided by a stale click (R3-B1 TOCTOU)` (`extension/tests/origins.integration.test.mjs:397`) faithfully replays round 3's interleaving: B's `raiseAccessRequest` enters the queue and is paused *mid-write* via a `storage.session.set` gate; the stale A click validates against the OLD prompt while B is parked; B then installs its fresh generation; the stale click runs and must NOT apply. Asserts `applied:false`, B's record still undecided, no grant minted, and B's current generation still decides normally afterwards. This is the exact adversarial sequence from round 3, and it passes (run locally, see Verification).

#### 2. R3-B1 duplicate current-generation delivery — closed, idempotent

Second `resolveOrigin` call with the same (current) `promptId`: the first decision persisted `accessRequest.decision`, so the re-read inside the queue hits `live.decision` truthy at `recordAccessDecisionLocked:220` and returns `false` — no double waiter resolve, no second grant write, no duplicate receipt. The once-grant from the first click is untouched (asserted in `duplicate current-generation delivery is idempotent (R3-B1)`, `test:462`). Correct.

#### 3. R3-M1 cancelled `await_access` lock leak — closed, verified (`local_operator/browser_bridge/daemon.py:529-543`)

The await branch now wraps the locked dispatch in `try/finally` with an **unconditional** `self._tab_locks.pop(tab_key, None)`. The `finally` covers all four exits: normal return, typed error, daemon `TimeoutError` (raised in `_wait_for_response:603` when the extension never answers), and `CancelledError` (client disconnect inside the HTTP handler). Unconditional pop is safe because `lock_key_for:404` keys awaits per-request-id (`__await__:{id}`) and request ids are unique — no other request can ever share the key, so an early pop cannot drop a live lock.

Tests read and run locally:
- `test_cancelled_await_evicts_its_lock_key` parks a real `_dispatch_serialized` task on a never-answering extension, asserts the key is present, cancels the task, asserts `CancelledError` propagates AND `_tab_locks == {}`. This exercises genuine task cancellation — the same shape as an asyncio/Starlette disconnect — not a mocked flag. Fair to call it the disconnect path: the handler's disconnect *is* delivered as task cancellation.
- `test_await_lock_key_evicted_on_normal_and_error_exits` covers normal completion and the 25 s daemon timeout on a silent extension (both assert empty lock map).

One note (not a finding): the timeout-half of the second test takes ~25 s wall-clock by design; that's inherent to exercising the real timeout path, and the suite already runs in CI in 12 m — acceptable.

#### 4. E2E regression — no evidence of regression

The round-3 remediation claims a 27/27 real-path E2E re-run (throwaway Chrome + daemon, full approve/deny/TTL/supersession/cross-session/stale-generation scenario, including the stale-generation rejection through the restructured `resolveOrigin`). The E2E harness is not checked into the repo, so I could not re-run it myself — instead I verified the two signals that gate it: (a) the full unit/integration bar passes locally (below), and (b) **CI on head `d1893535` is fully green** — all 8 checks: `build`, `lint`, `pip-audit`, `type-check`, `test (3.12)`, `test (3.13)`, `cli-sanity`, `server-sanity`. The delta also does not touch any code path the E2E exercises that isn't already covered by the two new integration suites, which passed.

#### Verification (run by this reviewer on `d1893535`)

| Command | Result |
|---|---|
| `pnpm typecheck` (extension) | ✅ clean |
| `pnpm test` (extension, node --test) | ✅ **35/35 pass** (incl. 2 new R3-B1 suites) |
| `pnpm build` (extension) | ✅ clean |
| `python -m local_operator.browser_bridge.gen_ts --check` | ✅ no drift |
| `pytest tests/unit/browser_bridge/test_daemon.py tests/unit/browser_bridge/test_protocol.py` | ✅ **17/17** (incl. 2 new M1 tests, ~57 s) |
| `pytest tests/unit/browser_bridge tests/unit/tools/test_browser_tool.py` (rest of bridge+tool slice) | ✅ **180/180** |
| `gh pr checks 329` on `d1893535` | ✅ all 8 green |

---

### Findings

**BLOCKER:** none.
**MAJOR:** none.

**MINOR**

- **M4-1 (documentation/semantics mismatch)** — `extension/src/origins.ts:271-272`. The `resolveOrigin` docstring says: *"A duplicate delivery of the CURRENT generation is idempotent: the record already carries the decision, so the re-read sees `live.decision` set and nothing re-applies (still `applied:true` — the user's click did land, it just already landed)."* The **code returns `applied:false`** for that duplicate (because `recordAccessDecisionLocked` returns `false` when `live.decision` is set), and the new test `duplicate current-generation delivery is idempotent (R3-B1)` asserts `second === false`. The behaviour (false on duplicate) is correct and is what the popup needs; the docstring's parenthetical claiming `applied:true` is stale and contradicts both. One-line fix: drop or correct the parenthetical. No behavioural impact.

**NIT:** none worth a round.

---

### Verdict: **clean**

No BLOCKER, no MAJOR. Round 3's two findings are genuinely closed at `d1893535`, verified by re-reading the critical section, reproducing the adversarial interleaving via the new tests, running the full unit/integration/CI bar, and confirming no new queue-deadlock hazard. This round is **TERMINAL**: the single MINOR (a docstring parenthetical) is a follow-up, not a reason for another round.

damianvtran:

Round 4 is clean/terminal at d1893535. The one MINOR (stale `resolveOrigin` docstring parenthetical claiming a duplicate returns `applied:true` while code+test correctly return `false`) is a comment-only fix folded into the next version-bump PR rather than forcing a fifth review round on a docs-only delta — tracked here so it doesn't get lost.

PR #333, opened 2026-08-26
fix(mobile): keep multi-part ask options tappable across questions

damianvtran:
## Summary

Fixes the mobile-relay web UI bug where, in a **multi-part ask**, the first question's options are tappable but the **next question's options render greyed-out and untappable** — the only workaround was to leave the session and re-open it.

## Root cause

A multi-question ask keeps the **same `request_id`** pending and only advances `question_index` (`local_operator/mobile/projection.py` `_sync_pending`/`set_pending`), so `projection.pending` never goes `null` between questions. `SessionScreen` rendered `<PendingCard>` with **no key**, so React reused the one card instance and its component-local `busy` state (set `true` while answering Q1) leaked into Q2. `inert = busy || error !== ""` stayed `true`, so every Q2 option rendered `disabled`.

The card's docstring *claimed* it "keys off `request_id`+`question_index` so React remounts the input between questions" — but that key sat only on the inner free-text `<input>`, never the whole card, so card-level `busy`/`error`/`remember` survived across questions.

## Fix

Key the **whole** `<PendingCard>` on `request_id:question_index` at the render site (`screens/session-view.tsx`). React now remounts the card per question with fresh state — the invariant is structural rather than relying on the card to reset itself. The redundant inner-input key is dropped. Within a single question the D7 inert behaviour (options disabled while an answer is in flight / after a stale-moved-on error) is unchanged.

## Changes

- `local_operator/mobile/web/src/screens/session-view.tsx` — render-site key + why/constraints comment.
- `local_operator/mobile/web/src/components/pending-card.tsx` — docstring corrected to match reality; success-path comment explains why no local reset is safe and why a local `busy` reset would break D7.
- `local_operator/mobile/web/src/session-view.pending.test.tsx` — regression test (renders the real `SessionScreen`, answers Q1, repaints to Q2 with the same `request_id`, asserts Q2 options are enabled and send `ask_answer` with `question_index: 1`). **Fails before, passes after.**

## Testing evidence

Reproduced the bug and verified the fix in a real browser (cmux browser surface), driving the **real production bundle** (`pnpm run build`) against a mock daemon that faithfully implements the projection/SSE contract the SPA consumes — including the exact contract detail the bug hinges on (multi-question ask keeps `request_id`, advances `question_index`).

- **Before** (origin/main bundle): Q1 tappable → click "us-east-1" → Q2 options render greyed (`opacity-50`, no accent edge) and a click on "yes" **fails as disabled** — the exact reported symptom.
- **After** (fixed bundle): Q1 tappable → click → Q2 options **bright/tappable** → click advances → Q3 of 3 also tappable. Full multi-part tap-through works with no session re-entry.

Screenshots are attached to this PR (see comments).

Unit/web suite: `npx vitest run` — 8 files, **43/43 pass** (Node 24.14; pre-existing happy-dom breakage under Node 26 is unrelated). `pnpm run check` (tsc) clean.

No Python touched — Python gates N/A. The web `dist/` bundle is gitignored and built on demand at install time (`lop mobile install` self-heal), so no committed artifact needs regenerating.


damianvtran:
### Design review — round 1

Reviewer: designer (claude-opus-4-8)
Scope: 00a1273e502774e5e28c4644dc9fb0dcf16256fe..11a03a1a6134e21f3fb42eeb3cb0233d0c817c56

Reviewed the five committed frames in `docs/evidence/mobile-multipart-ask/` (before-q1, before-q2-greyed, after-q1, after-q2, after-q3) as rendered images, plus the diff.

**Fix confirmed visually.** before-q2-greyed shows the bug exactly as described: Q2's options render with dimmed low-contrast label text, a flat dark fill, and no visible accent left edge — they read as disabled, which they were. after-q2 and after-q3 restore the full tappable affordance from before-q1: bright accent left edge, elevated warm fill, full-opacity label and description text. Q1 is byte-identical before and after, so the remount key introduces no change to the first question. Card geometry (width, padding, option gap, header layout, "Question N of 3" counter) is stable across after-q1 → q2 → q3 — same position and size — so the per-question remount produces no layout shift; only the content swaps, which is the intended behaviour. The counter advances correctly and the surrounding chrome (header, composer, send button) is untouched.

Findings:

- **D1 (MINOR)** — Evidence gap: no frame shows the in-flight/disabled state *within* a question (options inert while an answer is being submitted, the D7 state the code comments call out). That is the exact state that leaked to produce this bug, and the current set can't demonstrate it still renders correctly — dimmed-but-intentional within the active question. A single mid-submit frame (or a note that it was eyeballed) would close the loop; the new regression test covers behaviour, not appearance.
- **D2 (NIT)** — Stills cannot prove the remount is flicker-free between questions. The identical card geometry across after-q2/after-q3 makes a visible flash unlikely, but if it hasn't been watched live, one consecutive-frame capture across the Q1→Q2 boundary would settle it. Not blocking.
- **D3 (NIT)** — after-q3 is captured at a scrolled position with the session header and the "I need three decisions" message clipped at the top edge. Cosmetic evidence nit only; consider recapturing at the same scroll offset as the other frames for a cleaner before/after set.

No BLOCKER, no MAJOR. The frames confirm the fix restores the intended tap-target affordance on every subsequent question with no visual regressions.

**Verdict: TERMINAL.** D1–D3 are follow-ups (evidence completeness, not product defects).

damianvtran:
### Agent review — round 1

Scope: 00a1273e502774e5e28c4644dc9fb0dcf16256fe..11a03a1a6134e21f3fb42eeb3cb0233d0c817c56

**Heads-up on the Scope line / diff rendering:** the branch (`dev-mobile-ask`) is built on `68b55591` (`chore(release): 0.41.6`), one commit behind `origin/main` (`00a1273e`). The two-dot `base..head` diff above therefore also shows unrelated Python churn from the intervening release commit (`harness/loop.py`, `session/session.py`, `pyproject.toml`, several test files) as reversed/deletions. The true PR scope (merge-base `68b55591..11a03a1a`) is only the 4 web files + 5 evidence PNGs, and that is what I reviewed. Not a finding — just so the Scope line isn't misread.

---

**Summary of review.** The fix is correct and minimal. Keying the whole `<PendingCard>` on `request_id:question_index` at the render site (session-view.tsx:174) is the honest structural fix: it makes every component-local `useState` (`busy`/`error`/`remember`/`freeText`) fresh per question, which subsumes the previously-lost inner-input key. The daemon contract it relies on is real — `tui_handle._project_ask_question`/`_push_current_question` pop+re-push the SAME `request_id` with only `question_index` advanced, so `projection.pending` never goes null between questions and the key is exactly what forces the remount. D7 is preserved: with no local reset in `answer()`, `busy` stays true within the current question while a send is in flight, so the option/send buttons stay inert until the repaint — unchanged from base. Parallel approvals are unaffected (distinct `request_id`s, pop+push forces an unmount/remount). tsc (`pnpm run check`) is clean on head.

Verified by execution: ran the new regression test and the full web suite locally.

---

**A1 — MAJOR — The claimed fails-before/passes-after evidence is not reproducible, and the new test fails in this environment.** The PR states the web suite is "43/43 pass" and the regression test "fails before, passes after," both "under Node 24.14." I could not confirm either. On this machine (Node v26.5.0, the only toolchain present) the new test `src/session-view.pending.test.tsx` **fails on head** — `TypeError: Cannot read properties of undefined (reading 'getItem')` at `continuation-command.ts:176` (`localStorage.getItem`) via `Composer` mount, and `(reading 'clear')` at the test's own `localStorage.clear()` — because happy-dom provides no `localStorage` under Node 26. This is not unique to the new test: 27/43 tests fail here across `agent-view.interaction.test.tsx`, `continuation-command.test.ts`, `private-storage.test.ts`, and the new file. So the suite is red in at least one realistic environment, and the green evidence rests entirely on a Node version that isn't pinned anywhere I can find (`package.json` has no `engines`, and the `ci.yml` Python gates don't run vitest at all). Two consequences: (a) I can't independently reproduce "passes after," and (b) "fails before" couldn't be exercised either — with the impl reverted to base the test fails at the same localStorage crash *before ever reaching the Q2 assertion*, so it can't distinguish the bug from the environment. Please confirm the exact Node/pnpm versions and how the web suite is meant to be run in CI (or pin an engine / add a web job), and ideally make the new test environment-robust (e.g. assert on the disabled-attribute delta rather than depending on `localStorage`) so the regression guard actually guards on any supported toolchain.

**A2 — MINOR — `remember` resets to `false` when parallel approvals advance (count > 1).** When several approvals are pending, answering the first pops it and pushes the next (new `request_id`, `question_index` 0 on both). The old card's `busy` would have leaked into the next approval pre-fix (same grey-out bug); the render-site key now remounts, fixing that — but it also resets the `remember` checkbox to unchecked for the next approval. In practice this is harmless: the TUI handle (`tui_handle.approval_answer` → `prompt.resolve(approved, ...)`) ignores `remember` entirely, so the checkbox state never affects the daemon. Flagging only because the fix changes that state across approvals; if the mobile relay ever wires `remember` through to a real per-tool store, a user-set "remember" would be silently dropped on the next parallel approval. Consider whether the card should key on `request_id` only when `kind === "approval"` (so a single logical approval batch keeps its `remember`) — or confirm the TUI path permanently ignores it.

**A3 — MINOR — The render-site key omits `kind`; a same-`request_id` kind flip at the same `question_index` (0) would not remount.** The key is `request_id:question_index`. `request_id`s are `secrets.token_hex(8)` (unique per push) and an ask keeps one id across questions, so this can't collide in practice — but the key's uniqueness is only incidental (an approval→ask or ask→approval reuse of an id at index 0 would reuse state). Cheap hardening: include `pending.kind` in the key. Not required for correctness today; one line.

---

**N1 — NIT — Test's `pending_count: 1` doesn't match a real multi-part projection.** In `projection()` the test hardcodes `pending_count: 1`. The fold sets `pending_count = len(queue)`, and for a multi-question ask the queue front is the single current question, so `1` happens to be right here — but a reader could think `pending_count` tracks `question_total`. A one-line comment (`pending_count is the approval-queue length, always 1 for an ask`) would pre-empt that.

**N2 — NIT — Test mocks `SessionScreen`'s full child tree but only asserts the card.** The test renders the real `SessionScreen` (correct — it owns the key), which drags in `Composer`, sheets, transcript, etc., and is exactly why the `localStorage` crash surfaces. That's acceptable for a true integration guard, but it couples the regression to unrelated mount-time behaviour; see A1 for making it resilient.

---

**Verdict.** No BLOCKER. One MAJOR (A1 — reproducibility/environment of the test evidence, and the new regression test currently red under Node 26). The code fix itself is sound; the round is **not terminal** pending A1, because the central claim ("fails before, passes after") can't be verified in a clean environment and the regression guard is environment-fragile. A2/A3/N1/N2 are follow-ups. Recommend: confirm the intended web toolchain (pin Node), make the new test pass under it and be robust to `localStorage`, and re-run. Once the new test demonstrably fails-on-base / passes-on-head in a reproducible env, this is clean to ship.

damianvtran:
### UX review — round 1

Reviewer: ux-reviewer (k3)
Scope: 00a1273e502774e5e28c4644dc9fb0dcf16256fe..11a03a1a6134e21f3fb42eeb3cb0233d0c817c56

I walked the flow on the checked-out branch (`/tmp/lop-mobile-ask`, `dev-mobile-ask`): ran the real `SessionScreen` regression test and confirmed the base SHA `00a1273e` fails it (Q2 options render `disabled`) while head `11a03a1a` passes; read the daemon advance path (`tui_handle._project_ask_question` → pop+push same `request_id`, advanced `question_index`) to confirm the render-site key `request_id:question_index` actually changes per question; and inspected the before/after frames in `docs/evidence/mobile-multipart-ask/`.

Note on environment: the web suite fails under Node 26.5 (`happy-dom` exposes `window` but not `window.localStorage`); it passes under Node 22.22 with 43/43 green. That is a local-runtime mismatch, not a defect in this PR.

**The fix is correct and the flow is now continuous.** Q1→Q2→Q3 advance in place: the card header counts "Question N of 3", the composer stays docked, no session re-entry, no dead taps. The before-frame shows Q2 options greyed at ~50% opacity; the after-frames show them fully opaque and tappable. The mechanism is honest: because the daemon keeps `pending` non-null across questions (same `request_id`, advanced `question_index`), keying the whole card on `request_id:question_index` makes React remount and drop the leaked `busy` state structurally rather than relying on the card to reset itself.

Findings:

**U1 — MINOR — free-text/secret draft loss across questions is intended, but the typed-then-remounted window is now slightly wider.** The fix removed the per-field remount key on the `<input>` and now relies on whole-card remount. That still clears the draft per question (the invariant holds), and for the options path there is no draft at all. The only behaviour change is on the free-text/secret variant: previously the `<input>` remounted independently, so a user mid-typing when the daemon advanced would lose just the field; now the whole card remounts. In practice the card only advances after the user submits an answer (or the terminal advances it), so a user does not lose an in-progress draft to their own action. The residual case is a terminal-driven advance while the phone user has a half-typed secret — the card remounts and the half-typed value is gone. That is acceptable (the question moved on; carrying a typed secret to a different question would be worse), but worth naming. A user would expect a typed draft to survive only if the question is still the one they are answering.

**U2 — NIT — no transient "in-flight" affordance between tap and repaint on the options path.** On tap, `busy` flips and the option buttons dim (D7 inert window). On success the card remounts to the next question with no spinner or "sending…" state; on a slow daemon the user sees dimmed buttons for the round-trip duration, then the new question. The dim reads as "disabled" rather than "working". This is pre-existing (the approval path shows `…`), not introduced here, and the window is short on a healthy relay. A user might tap again thinking the first tap did not register; the `if (busy) return` guard makes that harmless.

**U3 — NIT — `count` "1 of N" badge path is unaffected.** Parallel approvals enqueue distinct `request_id`s, so each card's key changes when the queue re-fronts after a settle; the remount still fires and the badge still reads `1 of {count}`. I verified `_sync_pending` recomputes `pending_count` from the queue length and the render site passes it through unchanged. No interaction regression.

**U4 — NIT — focus/scroll on remount.** The whole-card remount drops DOM focus. On the options path there is no focused control to lose (the user tapped a button). On the free-text path, after submitting, focus leaves the input; the next question's input is not auto-focused. On mobile that means the soft keyboard dismisses and the user must tap the field again for a follow-up free-text answer. Minor because multi-question asks with consecutive free-text answers are uncommon, and re-tapping is one tap. A user would expect the keyboard to stay up if the next question is also free-text.

Verdict: no BLOCKER, no MAJOR. The core interaction defect (Q2 options untappable) is fixed and verified end-to-end. This round is TERMINAL. U1–U4 are follow-ups; none degrade the shipped flow below its pre-change state.

damianvtran:

- **A1 (MAJOR) fixed in 07c0d406** — the environment fragility was real: happy-dom 20.x attaches `localStorage` via a non-enumerable own getter, which vitest's `populateGlobal` (`Object.keys`) drops, so `window.localStorage` was absent on Node 26.5 and every test touching it crashed before its assertion. New shared setup `src/test-setup.ts` (wired via `setupFiles` in `vite.config.ts`) re-attaches happy-dom's real `Storage` to `globalThis`/`window`; `package.json` now pins `engines.node >= 22`. Reproducibility proven on Node 26.5.0 with the repo's own `npx vitest run`:
  - **Base** (only the render-site key reverted, test kept): `src/session-view.pending.test.tsx` **FAILS** at the Q2 assertion — `AssertionError: expected true to be false` on `expect(q2.hasAttribute("disabled")).toBe(false)` at line 126.
  - **Head** (key restored): same file **PASSES** (1/1).
  - Bonus: the shared shim also heals the other 26 pre-existing Node-26 failures — the full suite is now **43/43 on Node 26.5** (`pnpm run check` clean too).
- **A3 (MINOR) fixed in e2e2d84d** — key is now `request_id:kind:question_index`, with a comment stating why `kind` is included (uniqueness was incidental; guards against an id reused across kinds at index 0).
- **N1 (NIT) fixed in 89fe3c2b** — comment in the test projection: `pending_count` is the parallel-approval queue length (always 1 for an ask), not the question count.
- **A2 (MINOR) deferred with reason in 1c28a254** — the remount does reset `remember` when parallel approvals advance, but there is no clean one-line guard: excluding `question_index` from the key for approvals would reintroduce the busy-leak the key exists to prevent, and keying approvals on `request_id` only does not survive the queue re-fronting to a new `request_id`. It is harmless because `tui_handle.approval_answer` ignores `remember` entirely (`prompt.resolve(approved, ...)`) — the checkbox state never reaches the daemon. The trade-off and the revisit condition (if the daemon ever wires `remember` to a per-tool store) are documented in a comment at the render site.
- **N2 (NIT) noted** — mount-time localStorage coupling is reduced by the A1 shim: the real `SessionScreen` now mounts with a real `Storage` on any Node, so the integration guard no longer depends on mount-time ambient globals.

New head: 1c28a254e699ca5dbb82f3e3f93dd9f57fe3e939

damianvtran:
### Agent review — round 2

Reviewer: reviewer (kimi/k3)
Scope: 00a1273e502774e5e28c4644dc9fb0dcf16256fe..1c28a254e699ca5dbb82f3e3f93dd9f57fe3e939

**Verdict: TERMINAL — clean. No blocker, no major.** All round-1 findings are resolved or acceptably deferred, and the A1 fix is independently reproduced on the machine's default Node 26.5.0.

---

**A1 (MAJOR) — CONFIRMED FIXED** in `07c0d406`.

The mechanism is real and sound. `src/test-setup.ts` imports `Storage` from happy-dom and re-attaches fresh instances to `globalThis` and `window` via `Object.defineProperty` (configurable+writable). This correctly addresses the root cause: happy-dom 20.x exposes `localStorage`/`sessionStorage` as non-enumerable own getters on Window, and vitest's `populateGlobal` copies keys with `Object.keys(win)`, which skips non-enumerables — so on Node 26.5 the globals were absent and any mount-time access crashed. The shim restores a real Storage regardless of ambient Node behaviour. `vite.config.ts` correctly switched to `defineConfig` from `vitest/config` and wires `setupFiles: ["./src/test-setup.ts"]`. `package.json` gains `engines: { node: ">=22" }`, matching the documented floor.

Independent before/after reproduction on the worktree `/tmp/lop-mobile-ask` under Node 26.5.0:

- **Base** (only the `key` on `<PendingCard>` temporarily reverted to `` `${request_id}` ``): `npx vitest run src/session-view.pending.test.tsx` **FAILS** at `src/session-view.pending.test.tsx:131` with `AssertionError: expected true to be false` on `expect(q2.hasAttribute("disabled")).toBe(false)` — the greyed-out-Q2 bug reproduced.
- **Head** (key restored): same file **PASSES** (1/1, 41ms).

The fix in `session-view.tsx` itself was not altered by remediation; the key line is byte-identical to the round-1 head plus the A3 `kind` segment. Full suite is **43/43 passing** on Node 26.5.0 and `pnpm run check` (tsc -b --noEmit) is clean. The pre-existing 26 Node-26 failures in other files are healed by the shared shim, as claimed.

**A3 (MINOR) — CONFIRMED FIXED** in `e2e2d84d`. The key is now `` `${request_id}:${kind}:${question_index}` `` and the comment at the render site (session-view.tsx:161-102) explains why `kind` is included: uniqueness today is incidental (request_ids are random per push), but if an id were ever reused across kinds the card must not inherit the other kind's transient state. Cheap, correct hardening.

**N1 (NIT) — CONFIRMED FIXED** in `89fe3c2b`. The `pending_count` comment in the test projection now states it is the parallel-approval queue length (always 1 for an ask), not the question count, and names the two counters it feeds.

**A2 (MINOR) — DEFERRAL ACCEPTED** in `1c28a254`. The documented trade-off is accurate: the remount resets `remember` when parallel approvals advance, and no clean key shape fixes it without reintroducing the busy-leak the key exists to prevent (excluding `question_index` for approvals) or breaking the ask remount (keying approvals on `request_id` alone doesn't survive the queue re-fronting to a new request_id). Verified the harmlessness claim against `local_operator/mobile/tui_handle.py:429` — `approval_answer(request_id, approved, remember)` accepts `remember` but never uses it (`prompt.resolve(approved, answer=...)` only), so the checkbox state never reaches the daemon. The comment at the render site records the deferral and the revisit condition (if the daemon ever wires `remember` to a per-tool store). Reasonable and correctly documented.

**N2 (NIT) — CONFIRMED addressed** by the A1 shim: the real `SessionScreen` now mounts with a real `Storage` on any Node, so the integration guard no longer depends on mount-time ambient globals.

---

No new findings this round. The diff against base also contains the Python-side system-blocks refactor and test deletions from `origin/main` (base `00a1273e` is one commit ahead of the branch point `68b55591`; those hunks are not part of this PR's changes and reviewed green on main already).

Worktree left clean of review probing: only the pre-existing `pyproject.toml` version bump (0.41.7) and untracked `RELEASE_NOTES.md` remain, both present before this round.

Round is TERMINAL: remaining minors/nits from round 1 are resolved; nothing here warrants another round.

damianvtran:
### Agent review — round 3

Scope: 00a1273e502774e5e28c4644dc9fb0dcf16256fe..1c28a254

**Scope note (as rounds 1-2 flagged):** the branch is built on `68b55591`, one commit behind `origin/main` (`00a1273e`), so the two-dot `base..head` above also surfaces unrelated main-side Python churn (system-blocks refactor, test deletions) as reversals. The true PR scope (merge-base `68b55591..1c28a254`) is the 4 web source files + config + 5 evidence PNGs, which is what I reviewed.

**Core fix — CONFIRMED correct.** `screens/session-view.tsx:185` keys the whole `<PendingCard>` on `request_id:kind:question_index`. The daemon keeps the same `request_id` pending across a multi-question ask and only advances `question_index` (verified `pending` never goes null between questions), so the key changes per question and React remounts the card with fresh `useState`. A stale `busy` from answering Q1 therefore cannot disable Q2's options — the leak the fix targets. The render-site comment (session-view.tsx:161-184) explains the why/constraint thoroughly, including the A2 deferral note about `remember`. The companion card comment (`pending-card.tsx`) correctly explains why no per-field key is needed (the whole-card remount covers the draft) and why `busy` must not be locally reset on success (would break D7 within-question). The A2 harmlessness claim checks out: `tui_handle.approval_answer` (`tui_handle.py:429`) accepts `remember` but `settle()` only calls `prompt.resolve(approved, answer=...)` and never uses it, so the remount resetting the checkbox has no daemon-visible effect today.

**Remediation commits — CONFIRMED sound.**
- `07c0d406` (`test-setup.ts` Storage shim + `vite.config.ts` setupFiles + `package.json` engines `>=22`): correctly addresses the happy-dom 20.x non-enumerable-getter root cause (vitest `populateGlobal` copies via `Object.keys(win)`, skipping non-enumerables, so `localStorage`/`sessionStorage` were absent on Node 26). Re-attaching real `Storage` via configurable+writable `defineProperty` makes the suite runtime-independent. Sound.
- `e2e2d84d` (`kind` in key): cheap, correct hardening; comment explains the collision case is incidental today but the card must not inherit another kind's state if an id were ever reused.
- `89fe3c2b` (`pending_count` comment): now correctly states it is the parallel-approval queue length (always 1 for an ask), not the question count, and names the two counters it feeds.
- `1c28a254` (A2 deferral note): accurate and correctly documented with its revisit condition.

**A1 gate — independently reproduced on the machine's default Node 26.5.0** (worktree `/tmp/lop-mobile-ask`):
- **Base** (only the `key` on `<PendingCard>` temporarily reverted to `` `${request_id}` ``, everything else untouched): `npx vitest run src/session-view.pending.test.tsx` **FAILS** at `src/session-view.pending.test.tsx:131` — `AssertionError: expected true to be false` on `expect(q2.hasAttribute("disabled")).toBe(false)`. The greyed-out-Q2 bug reproduced.
- **Head** (key restored): same file **PASSES** (1/1). The fix line itself is unaltered afterward (`git diff` clean for `session-view.tsx`; restored line is byte-identical to head).

**Full suite + type check on Node 26.5.0:** `npx vitest run` → **43/43 passing** (8 files); `pnpm run check` (`tsc -b --noEmit`) → **clean**.

Worktree left clean of review probing: `session-view.tsx` is byte-identical to head; only the pre-existing `pyproject.toml` version bump and untracked `RELEASE_NOTES.md` remain, both present before this round.

**Verdict: TERMINAL — clean. No BLOCKER, no MAJOR.** The fix is sound and the round-1 findings (A1/A3/N1 fixed, A2 acceptably deferred, N2 addressed) are all resolved and re-verified on the current head `1c28a254`. This round is a re-confirmation; nothing here warrants another round.

PR #296, opened 2026-08-25
feat(sessions): unify frontend state across resumed TUIs

damianvtran:
## Summary of Changes

Full terminal resume now hydrates and stays live from one canonical `FrontendSessionState` instead of reversing the phone `SessionProjection` into partial TUI semantics.

- Added a versioned JSON-safe session state/store with complete selected/effective model specs, lossless usage components, cumulative cost provenance, context, duration, jobs, phased todos, wakes, MCP, identity, gates, queued steering and slash capability metadata.
- `Session` owns and publishes the state through explicit mutation hooks for title, goal, team/profile, model/fallback, todos, wakes, MCP, steering and jobs. The canonical ledger accrues each provider call live, reconciles the final remainder, preserves child cost and checkpoints takeover state.
- Protocol v5 sends one atomic `frontend_sync` snapshot followed by typed sequenced field deltas. Snapshot+subscribe is marshalled onto the owner loop, state and raw events share one bounded ordered FIFO, queue pressure drops only the slow client, and gaps/epoch changes force a fresh v5 reconnect.
- Real `AsyncJob` state now carries safe dict/string progress, prompt, start/settle clocks, model, context, usage/cost, queue/status/result/error, output and trajectory. Trajectory bytes are append-delta encoded instead of repeated in every update.
- `RemoteSession` consumes only canonical full-TUI state plus raw transcript events. It no longer reconstructs from phone projection, accepts stale protocol v4 reconnects, owns a follower-local steering queue, or depends on `mobile/live_turn.py`.
- The unchanged standard `OperatorApp` widgets render owner, follower and takeover sources. Live peer `ImageContent` uses the normal `ImageBlock`; durable attachment replay remains content-addressed.
- Every advertised slash command has an explicit scope. Owner-dependent commands route through authoritative operations, including image-bearing `/agent` and `/team`; `/btw` keeps its local overlay while executing its provider request on the owner; canonical `/mcp` remains local data rendering.

### Round-1 review remediation (`ad8ae852`)

The follow-up round closed the stop-ship and parity gaps the first review found, all by fixing the unified contract rather than adding follower-only UI:

[...]

PR #340, opened 2026-08-26
fix(mcp): adopt peer-rotated token on 401 to avoid Notion logout

damianvtran:
## Summary

Fixes the residual Notion MCP OAuth logout bug: after one session rotates the shared grant, every OTHER live session still holds a locally-valid but server-side-revoked access token, sends it, gets a 401, and the SDK runs the full browser authorization grant — surfacing as `McpAuthRequiredError` ("run /mcp login notion") and suspending auto-reconnect.

## Root cause

Notion rotates its refresh token on every use and revokes ALL previously issued access tokens when a refresh happens (access tokens live 8h). local-operator runs one long-lived process per cmux workspace, all sharing ONE stored grant row (`auth.db`, `provider='mcp-oauth'`, `identity_key='https://mcp.notion.com/mcp'`).

PR #242 added `_make_refresh_coordinating_provider` in `local_operator/mcp/auth.py`: before the SDK's in-flow refresh, it re-reads storage under `_oauth_refresh_lock` and adopts a peer-rotated token. BUT its gate is `if ctx.is_token_valid() or not ctx.can_refresh_token(): return` — it only coordinates when the IN-MEMORY token is EXPIRED. The residual gap: after one session rotates the grant, every other live session still holds a locally-VALID (unexpired) but server-side-REVOKED access token. That session skips coordination, sends the revoked token, gets 401, and the SDK's `async_auth_flow` 401 branch runs the FULL browser authorization grant.

Empirical timeline on this machine: one successful rotation at 16:43, then cascading "OAuth authorization expired" incidents in OTHER sessions 16:52-17:07. The stored row is healthy (fresh token + refresh token) — the failure is purely the revoked-but-unexpired in-memory token.

## The fix

In `_RefreshCoordinatingOAuthProvider.async_auth_flow`, intercept a 401 response ONCE per flow invocation:

- When the response is a 401 on the ORIGINAL request and no adoption-retry has happened yet in this flow: under `_oauth_refresh_lock(server_url)`, re-read `stored = await ctx.storage.get_tokens()`. If `stored` has an access_token DIFFERENT from `ctx.current_tokens.access_token`: adopt it (`ctx.current_tokens = stored`, `ctx.token_expiry_time = storage.stored_token_expiry()`), rewrite the original request's Authorization header to the adopted token IN PLACE, and RE-YIELD the request so the SAME httpx client that sent the original sends the retry (same proxies, TLS, and event hooks). The retry's response is fed into the SDK generator instead of the 401, so the SDK's full browser-authorization branch never sees a challenge.
- The retry goes through the caller's client by RE-YIELDING, matching the SDK's own end-of-flow 401 retry contract (`_add_auth_header` rewrites the same request object and re-yields it). The provider is an httpx Auth: it only ever YIELDS requests; nothing in the SDK or local-operator sets a transport on it, so an internal send is not possible. Header mutation never touches the request body/stream, so the re-send is body-safe.
- Adoption never spends a refresh token, so no rotation race; the lock still serializes the re-read against a concurrent rotation.
- If NOT adopted (stored token identical to in-memory, or absent): pass the 401 through unchanged — existing behaviour for genuinely dead grants.
- Bounded to one adoption-retry per flow: a second 401 (the adopted token also rejected, i.e. a genuinely dead grant) passes through to the SDK's full-flow branch.

### Unit — `tests/unit/mcp/test_auth.py::TestInflightRefreshCoordination` (9 tests)

Three new tests extend the existing coordinating-provider class, driving the flow purely through yields (no transport injection — the responder observes the re-yielded retry request exactly as the real httpx client would):

- **(a) 401 with a DIFFERENT fresh stored token** → the original request is re-yielded with the adopted Authorization header (same object), the SDK never yields a discovery/registration request, and the flow completes on the retry's 200.
- **(b) 401 with IDENTICAL stored token** → 401 passed through to SDK (full-flow branch reachable as before; exactly one request to the resource URL, the rest are SDK discovery).
- **(c) second 401 after an adoption retry** → passes through (bounded to one retry; exactly two requests to the resource URL, then SDK discovery).

```
$ env -u NO_COLOR TERM=xterm-256color .venv/bin/python -m pytest tests/unit/mcp/test_auth.py -q -k "InflightRefreshCoordination"
.........                                                                [100%]
9 passed in 1.81s
```

### Two-process integration repro — `scripts/repro_notion_401.py`

A scripted repro with a fake Notion-like authorization server (rotating refresh token; on rotation, revokes prior access tokens; 401 for revoked tokens) and TWO provider instances in SEPARATE OS processes sharing one `McpTokenStorage`-backed auth store (tmp sqlite `auth.db`). Process A refreshes (rotates). Process B (holding the old unexpired token) makes a request, gets 401, and must succeed via adoption WITHOUT any browser-grant/discovery request. The repro drives each flow purely through yields — the adoption retry appears as a second yield of the same request object bearing the adopted token.

**Pre-fix (origin/main, no wrapper change):** process B's 401 drives the SDK into the full registration/authorization branch and fails:

```
$ .venv/bin/python scripts/repro_notion_401.py /tmp/notion-repro-pre
seed: issued original grant {'access_token': 'access-1', 'refresh_token': 'refresh-1'}
A: rotated grant; live access token is now access-2
B: booted holding 'access-1' (locally valid, revoked by A); store holds 'access-2'
...
mcp.client.auth.exceptions.OAuthRegistrationError: Registration failed: 401
ORCH: FAIL — process B did not recover via adoption
(exit 1)
```

**Post-fix (this branch):** process B recovers by adopting the sibling's token, no browser grant:

```
$ .venv/bin/python scripts/repro_notion_401.py /tmp/notion-repro
seed: issued original grant {'access_token': 'access-1', 'refresh_token': 'refresh-1'}
A: rotated grant; live access token is now access-2
B: booted holding 'access-1' (locally valid, revoked by A); store holds 'access-2'
B: PASS — 401 recovered by ADOPTING sibling token 'access-2', no browser grant, final status 200
ORCH: PASS — two-process 401 adoption recovered without a browser grant
(exit 0)
```

### Gates

```
$ .venv/bin/python -m flake8 local_operator/mcp/auth.py tests/unit/mcp/test_auth.py scripts/repro_notion_401.py
(clean)
$ uvx --from black==26.1.0 black --check local_operator/mcp/auth.py tests/unit/mcp/test_auth.py scripts/repro_notion_401.py
3 files would be left unchanged.
$ uvx isort==5.13.2 --check local_operator/mcp/auth.py tests/unit/mcp/test_auth.py scripts/repro_notion_401.py
(clean)
$ .venv/bin/python -m pyright --pythonpath .venv/bin/python local_operator/mcp/auth.py tests/unit/mcp/test_auth.py scripts/repro_notion_401.py
```

Full unit suite: 7140 passed, 7 skipped, plus 8 pre-existing failures in `tests/unit/server/test_server_models.py` (`AttributeError: 'State' object has no attribute 'env_config'`) that reproduce identically on clean origin/main and are unrelated to this change.

## Version

Bumped to 0.42.1 (patch: bug fix per AGENTS.md versioning). Note: the task originally targeted 0.41.7, but 0.41.7 through 0.41.10 and 0.42.0 were released while this work was in progress, so this bumps to the next available patch.

### DIFF
--- a/local_operator/mcp/auth.py
+++ b/local_operator/mcp/auth.py
@@ -59,11 +59,6 @@ from local_operator.callback_page import callback_response
 if TYPE_CHECKING:
     # The SDK is an optional extra: these names are needed for annotations
     # only, so importing them here keeps this module importable without it.
-    # NOTE: the SDK's auth flow is typed against its own vendored ``httpx2``
-    # (a separate distribution from ``httpx``), and ``gen.asend`` for the inner
-    # generator only accepts ``httpx2.Response`` — so the adoption retry's
-    # response annotation must be ``httpx2``'s, not ``httpx``'s.
-    import httpx2
     from mcp.shared.auth import (
         AuthorizationCodeResult,
         OAuthClientInformationFull,
@@ -1798,17 +1793,20 @@ def _make_refresh_coordinating_provider(
             # path when httpx closes the OUTER flow (its ``finally:
             # await auth_flow.aclose()``).
             inner = super().async_auth_flow(request)
-            # One 401-driven token adoption is allowed per flow invocation;
-            # ``None`` = the first yield has not arrived yet (the SDK re-auth
-            # machinery never yields the ORIGINAL request object again, so the
-            # guard can latch on identity), and anything True means the adoption
-            # budget is spent — a second 401 must pass through untouched.
-            original_request: Any = None
+            # One 401-driven token adoption is allowed per flow invocation.
+            # ``original_request`` is the caller's request object itself (NOT
+            # the first yield: when the loaded token is expired the SDK yields
+            # a refresh request first, and latching on the first yield would
+            # point the identity guard at the wrong object). The SDK re-auth
+            # machinery never yields this object again except its own end-of-flow
+            # retry, so ``outgoing is original_request`` reliably identifies the
+            # caller's request; ``adoption_attempted`` spends the one-retry
+            # budget — a second 401 must pass through untouched.
+            original_request: Any = request
             adoption_attempted = False
             try:
                 try:
                     outgoing = await inner.__anext__()
-                    original_request = outgoing
                 except StopAsyncIteration:
                     return
                 while True:
@@ -1863,16 +1861,33 @@ def _make_refresh_coordinating_provider(
                         # (stored token identical to ours, or none) also passes
                         # through unchanged on this first 401.
                         adoption_attempted = True
-                        retry_response = await self._adopt_peer_token_once(
-                            request, original_request
-                        )
-                        if retry_response is not None:
-                            # A peer's fresh token answered the request: feed the
-                            # retry's response into the SDK generator INSTEAD of
-                            # the 401, so its full browser-authorization branch
-                            # never sees a challenge. Mirrors the SDK's own
-                            # end-of-flow retry, which re-sends the same request
-                            # object after ``_add_auth_header`` mutates it.
+                        retry_request = await self._adopt_peer_token_once(original_request)
+                        if retry_request is not None:
+                            # Re-yield the retry request: the SAME httpx client
+                            # that sent the original sends this one (same
+                            # proxies, TLS, and event hooks), exactly matching
+                            # the SDK's own end-of-flow 401 retry, which
+                            # re-yields the request after ``_add_auth_header``.
+                            # Then feed the retry's response into the SDK
+                            # generator INSTEAD of the 401, so its full
+                            # browser-authorization branch never sees a
+                            # challenge.
+                            try:
+                                retry_response = yield retry_request
+                            except GeneratorExit:
+                                # The outer flow is closing: re-raise so the
+                                # ``finally`` below closes the inner generator
+                                # and the SDK unwinds its lock.
+                                raise
+                            except BaseException as exc:  # noqa: BLE001
+                                # A transport fault on the retry: deliver it into
+                                # the SDK generator so its ``finally`` runs and
+                                # the lock is released, then relay what it yields.
+                                try:
+                                    outgoing = await inner.athrow(exc)
+                                except StopAsyncIteration:
+                                    return
+                                continue
                             try:
                                 outgoing = await inner.asend(retry_response)
                             except StopAsyncIteration:
@@ -1894,24 +1909,22 @@ def _make_refresh_coordinating_provider(
                 # never left suspended. Idempotent if already exhausted/closed.
                 await inner.aclose()
 
-        async def _adopt_peer_token_once(
-            self, request: Any, original_request: Any
-        ) -> "httpx2.Response | None":
-            """Re-send the original request once with a peer-rotated token, if any.
+        async def _adopt_peer_token_once(self, original_request: Any) -> Any:
+            """Adopt a peer-rotated token and return the request to retry, if any.
 
-            Returns the retry's ``httpx.Response`` when the store held an access
-            token DIFFERENT from the one in memory and the re-send completed —
-            whatever status it came back with, success or failure, so the SDK
-            sees the truth about the adopted token. Returns ``None`` when there
-            is nothing to adopt (stored token identical to ours, or no row), in
-            which case the caller passes the original 401 through to the SDK.
+            Returns the ORIGINAL request with its Authorization header rewritten
+            to the adopted token when the store held an access token DIFFERENT
+            from the one in memory — the caller re-yields it so the SAME httpx
+            client that sent the original sends the retry (same proxies, TLS,
+            and event hooks). Returns ``None`` when there is nothing to adopt
+            (stored token identical to ours, or no row), in which case the
+            caller passes the original 401 through to the SDK.
 
-            ``request`` is the request object the SDK is driving; the adopted
-            token is installed into the context AND into that object's
-            Authorization header, because the SDK's end-of-flow retry re-sends
-            this same object after ``_add_auth_header`` — mutating it here keeps
-            the retry (and any later one) on the adopted token rather than
-            stamping the revoked one back over it.
+            Mutating the request object in place (rather than building a new
+            one) matches the SDK's own end-of-flow 401 retry contract:
+            ``_add_auth_header`` rewrites the same object's Authorization header
+            and re-yields it. Header mutation never touches the request
+            body/stream, so the re-send is body-safe.
             """
             try:
                 async with _oauth_refresh_lock(self._refresh_coord_server_url):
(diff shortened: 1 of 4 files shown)


damianvtran:
### Agent review — round 1

Scope: main..dev-notion-401-adoption (5c9ec86..30baa94)

**Verdict: clean — TERMINAL.** No BLOCKER, no MAJOR. Remaining items are follow-ups, not a reason for another round.

Reviewed the full diff at head `30baa947`, the installed mcp SDK 2.1.0 / httpx2 2.12.0 sources, and the production wiring (`manager.py` → `create_mcp_http_client(auth=provider)`), then verified by execution rather than inference.

**Verified by running, not reading:**
- Gates clean on all three changed files: flake8, black 26.1.0, isort 5.13.2, pyright (0 errors).
- `TestInflightRefreshCoordination`: 9 passed; full `tests/unit/mcp/test_auth.py`: 82 passed; full `tests/unit/mcp`: 257 passed.
- Two-process repro `scripts/repro_notion_401.py`: PASS — process B adopts sibling token `access-2`, no browser grant, final 200.
- **Decisive pump-semantics check:** I drove the provider through a real `httpx2.AsyncClient` (the production send path, not the manual pump the tests use). The caller receives the retry's **200**; the intercepted 401 lands in `response.history` — exactly httpx's normal auth-retry contract. Transport saw `Bearer revoked-but-unexpired` then `Bearer fresh-by-peer`. So the re-yield approach returns the right response to the MCP transport.

**Correctness points confirmed:**
- *The first commit's blocker is resolved.* `18782265` re-sent the retry via `self._transport`, but nothing in the mcp SDK or local-operator ever sets `_transport` on a provider (exhaustive grep of both trees), so it would always have been `None` and silently passed every 401 through — the fix would never have fired in production. `a00ab95b` replaced it with a re-yield through the caller's client; current head has zero `_transport` references. Good catch by the implementer.
- *No deadlock vs `context.lock`.* `_adopt_peer_token_once` acquires only the cross-process file lock; the inner SDK generator holds `context.lock` but is suspended at its `yield`. No nested acquisition of the same lock, no circular wait. The `GeneratorExit`/`BaseException` handlers still deliver into `inner` so its `finally` releases the lock.
- *Bounded.* `adoption_attempted` latch plus the `outgoing is original_request` identity guard; a second 401 (adopted token also rejected) passes through to the SDK's full-flow branch. The identity guard correctly latches on the caller's request object even when the SDK yields a refresh request first.
- *Replay-safe.* Header-only mutation of a request whose body is materialized at construction — the same contract the SDK's own end-of-flow 401 retry relies on.
- *Storage re-read race.* Re-read happens under `_oauth_refresh_lock`; adopts only a token DIFFERENT from in-memory, else passes the 401 through unchanged. Adoption never spends a refresh token, so no rotation race.
- *No non-Notion regression.* The branch fires only on a 401 to the original request AND only when the store holds a different token; identical-token and no-row cases pass through untouched (covered by tests b and c).

**MINOR (follow-ups, not blocking):**
1. `_oauth_refresh_lock` is a single global lock file (`mcp_oauth_refresh.lock`) regardless of `server_url`; the 401-adoption path adds one more acquisition site. It is held only for a storage read + header write, so cross-server contention is negligible, but adoption on one server briefly serializes with refresh/adoption on all servers. Pre-existing lock design, not introduced here.

**NIT:**
1. `scripts/repro_notion_401.py` uses POSIX `fcntl`, so it won't run on Windows. Dev-only repro, not shipped, so cosmetic.

Testing evidence is real execution proof (unit + two-process repro + my independent real-client check), not a green suite alone. Ship it.

damianvtran:

Implementer: coder (claude-opus-4-8)
Head: 30baa947 (unchanged — no code edits, round stays fresh against the reviewed SHA)

Round 1 returned a clean, TERMINAL verdict (no BLOCKER, no MAJOR). The two follow-ups are both dispositioned as deferred; neither warrants a code change in this slice.

**MINOR-1 — `_oauth_refresh_lock` is a single global lock file regardless of `server_url`.** DEFERRED. This is the pre-existing cross-process lock design shared by every refresh path (`ensure_mcp_oauth_fresh` startup refresh, `_coordinate_inflight_refresh`, and now the 401-adoption re-read); this PR only adds one more acquisition site to it, it does not introduce the design. Contention is negligible in practice: the lock is held only for a single storage read plus a header write (no network I/O on the adoption path — adoption never performs the token exchange), so the window where adoption on one server serializes with refresh/adoption on another is sub-millisecond. Per-server lock sharding is a reasonable future refactor but is out of this slice's scope, would touch every refresh path rather than the one under review, and carries no user-visible benefit at current server counts. Tracked as a follow-up rather than expanded into this PR.

**NIT-1 — `scripts/repro_notion_401.py` uses POSIX `fcntl`, so it will not run on Windows.** DEFERRED. The script is a dev-only two-process reproduction harness; it is never packaged or shipped to users (it lives in `scripts/`, outside the installed surface), and it deliberately mirrors production's own `_oauth_refresh_lock`, which is itself `fcntl`-based POSIX. This repo's CI runners and the maintainer's dev machines are POSIX (macOS/Linux), so the script runs where it is actually used. Making a throwaway repro cross-platform would add complexity for zero benefit; deferred as cosmetic.

No blocker/major findings remain; the round is clean and fresh against head `30baa947`.

damianvtran:
### Agent review — round 1 (scope refresh)

Scope: main..dev-notion-401-adoption (5c9ec86..d553b07e)

The head moved from `30baa947` to `d553b07e` only by a rebase onto current main plus a version bump to 0.41.11 (0.41.10 was released meanwhile via #341). I re-verified independently: `git diff 30baa947 d553b07e -- local_operator/ tests/ scripts/` is EMPTY, and the full diff between the two heads is the single `pyproject.toml` version line. The fix, tests, and repro script are byte-identical to the head reviewed in round 1.

The round-1 verdict therefore stands unchanged against `d553b07e`: **clean, TERMINAL** — no BLOCKER, no MAJOR. The one MINOR (global refresh-lock file shared across servers, pre-existing design) and one NIT (POSIX-only repro script) remain follow-ups, not reasons for another round. All round-1 execution evidence (gates, 257 MCP tests, two-process repro, real-`httpx2.AsyncClient` pump check) applies verbatim to the identical code.

damianvtran:
### Agent review — round 1 (scope refresh 2)

Scope: main..dev-notion-401-adoption (fb26db45..534144a4)

The head moved from `d553b07e` to `534144a4` only by a rebase onto current main (which now includes #296's 0.42.0) plus a version bump to 0.42.1. I re-verified independently: `git diff 30baa947 534144a4 -- local_operator/mcp/auth.py tests/unit/mcp/test_auth.py scripts/repro_notion_401.py` is EMPTY, so the three reviewed files are byte-identical to the head reviewed in round 1. The rebase also touched nothing under `local_operator/mcp/` and left the `mcp>=1.2` / `httpx>=0.27` dependency pins unchanged, so nothing in 0.42.0 interacts with the fix.

The round-1 verdict therefore stands unchanged against `534144a4`: **clean, TERMINAL** — no BLOCKER, no MAJOR. The one MINOR (global refresh-lock file shared across servers, pre-existing design) and one NIT (POSIX-only repro script) remain follow-ups, not reasons for another round. All round-1 execution evidence (gates, 257 MCP tests, two-process repro, real-`httpx2.AsyncClient` pump check) applies verbatim to the identical code.

PR #343, opened 2026-08-27
feat(tui): goal-driven /loop <goal> mode with a judge-gated completion loop

damianvtran:
## Summary

Adds a goal-driven mode to the `/loop` slash command. Today `/loop` accepts only a number: `/loop 5` runs five fixed iterations of a canned "advance the standing goal" prompt, and any non-integer argument (e.g. `/loop finish the parser`) fails with `usage: /loop [n] (1..25)`. That mismatched the way loops work in other agent harnesses, where you type the goal directly and the agent keeps going until it is met.

Now `/loop <goal text>` starts a **judge-gated** loop: it runs turns toward the goal, and after each turn a forked, off-the-record judge call (same session model, via `complete_aside`) decides whether the goal is achieved. Not-achieved force-continues; achieved releases the loop and notifies the user **once**. Intermediate yields do not notify — that is the whole point, to avoid a completion toast on every iteration.

Fully **additive**: `/loop [n]` (numeric, still 1..25), `/loop stop`, and `/goal` are unchanged.

## Behaviour

Dispatch on the argument shape in `_cmd_loop`:
- `/loop stop|cancel|abort` → cancels a running loop (unchanged).
- `/loop` (no arg) → `DEFAULT_LOOP_ITERATIONS` numeric run (unchanged).
- `/loop 5` (integer) → five bounded iterations, 1..25 guard + standing-goal guard (unchanged).
- `/loop <non-integer text>` → **new** goal mode with that text as the goal.

Goal mode:
- Carries an **ephemeral** `self._loop_goal` — it does **not** call `session.set_goal`, so a `/loop do X` never clobbers a standing `/goal` the user set for the session, and restores nothing on exit because it changed nothing.
- **No hard iteration ceiling** (runs until the judge releases), but is cooperatively cancellable via `/loop stop`, reload-safe (`session is not self._session`), and protected by a **3-consecutive-judge-failure breaker** so a broken judge cannot spin forever.
- The judge verdict is parsed from distinct `VERDICT: ACHIEVED` / `VERDICT: CONTINUE` tokens by a pure `_parse_loop_verdict`; CONTINUE is checked before ACHIEVED and a negated "NOT ACHIEVED" never reads as a false release. An unreadable or errored verdict **fails safe to CONTINUE** with a surfaced warning (a false "done" that breaks user trust is worse than one extra loop).
- **Notify-once gating:** a `_loop_suppress_completion` flag short-circuits the per-turn completion toast in `on_turn_ended` while the loop is held; the worker fires exactly one `_notify("complete", ...)` on release. All three new state fields reset in the worker's `finally` (load-bearing — a stuck suppress flag would mute completions process-wide).
- Judge cost is charged via `_charge_aside`, so each goal turn's judge tokens are accounted, not silently free. Note: a goal turn costs roughly one extra full-context judge call on top of the turn.

### One intentional behaviour change
`/loop abc` no longer prints a usage error — it is now a goal. This is the point of the feature. The out-of-range **integer** error (`/loop 99`) is retained.

## Validation

All gates run over the whole tree from the branch (editable venv):

- `flake8 .` → clean.
- `black==26.1.0 --check .` → `560 files would be left unchanged`.
- `isort==5.13.2 --check .` → clean.
- `pyright` on the changed files (`local_operator/tui/app.py`, `tests/unit/tui/test_app_pilot.py`) → `0 errors, 0 warnings, 0 informations`. CI runs the authoritative tree-wide pyright.
- `pytest tests/unit/tui -q` → **2574 passed, 4 skipped**; loop subset **18 passed**.
- `pytest tests/unit -q` → the only failures are 23 in `tests/unit/tools/test_eval_tool.py`, pre-existing and environmental (the eval-worker subprocess cannot import `local_operator` under a symlinked-venv worktree); they fail **identically in a clean `~/local-operator` checkout** and are untouched by this change. CI's real install runs them green.

### Real-path execution evidence

Driven against the real `OperatorApp` via `run_test` with a scripted judge (no live model call), exercising the actual `_cmd_loop` + goal worker + `on_turn_ended` path:

```
=== Scenario A: CONTINUE, CONTINUE, ACHIEVED ===
  turns run (session.prompts) : 3
  judge calls                 : 3
  completion notifies fired   : 1 (expect 1, only on release)
  goal-achieved receipt       : True
  suppress flag reset         : True
  loop_goal reset             : True
  loop_running reset          : True
=== Scenario B: suppress flag gates TurnEnded completion toast ===
  completion notifies while suppressed : 0 (expect 0)
  completion notifies after clear      : 1 (expect 1)
=== Scenario C: /loop stop cancels a running goal loop ===
  running when stop issued : True
  'loop cancelled' receipt : True
  all three fields reset   : True
=== Scenario D: verdict parser ===
  'VERDICT: ACHIEVED\nall done'  -> (True, 'all done')  OK
  'VERDICT: CONTINUE\nnot yet'   -> (False, 'not yet')  OK
  'verdict: achieved\nok'        -> (True, 'ok')         OK
  'VERDICT: NOT ACHIEVED\nnope'  -> (False, 'nope')      OK
  ''                             -> (None, '')           OK
  'just prose\nno verdict'       -> (None, '')           OK
  'VERDICT: ACHIEVED'            -> (True, '')            OK
```

This confirms the core contract: N turns then release, zero intermediate notifications and exactly one on release, `/loop stop` cancels mid-loop, and all loop state resets.

## Review

Independent agent review is required before merge. This PR must not merge until that round is posted, answered, and fresh against the current head. As a user-visible interaction change (new command behaviour + notify-once flow), it also carries a UX round; the Scenario A/B/C evidence above is the real-run flow proof. No new painted widget, so no design round.

### DIFF
--- a/local_operator/tui/app.py
+++ b/local_operator/tui/app.py
@@ -690,8 +690,15 @@ def _parse_loop_verdict(text: str) -> tuple[bool | None, str]:
                 # word that merely CONTAINS "not"/"n't" (nothing, cannot, another)
                 # no longer causes an unbounded false-continue.
                 tokens = _NEGATION_TOKEN_RE.findall(payload)
+                # The contraction suffix is checked for BOTH apostrophe forms:
+                # the tokeniser accepts a curly apostrophe (U+2019), which most
+                # editors and phones autocorrect a straight one into, so
+                # `can't`/`can’t` must both count as a negator. Missing the curly
+                # form would let `VERDICT: ACHIEVED but can’t verify` read as a
+                # false RELEASE — the one trust-breaking outcome R2 closed.
                 negated = any(
-                    tok in ("NOT", "NOT_ACHIEVED") or tok.endswith("N'T") for tok in tokens
+                    tok in ("NOT", "NOT_ACHIEVED") or tok.endswith(("N'T", "N\u2019T"))
+                    for tok in tokens
                 )
                 achieved = not negated
             continue
@@ -11218,7 +11225,7 @@ class OperatorApp(App[None]):
         # directly with `/loop <goal text>` instead of only via `/goal` first.
         if not getattr(session, "goal", ""):
             notice(
-                "set a goal first with /goal <text>, or loop toward one now: " "/loop <goal text>",
+                "set a goal first with /goal <text>, or loop toward one now: /loop <goal text>",
                 "warning",
             )
             return
--- a/local_operator/tui/app.py
+++ b/local_operator/tui/app.py
@@ -25,6 +25,7 @@ from __future__ import annotations
 import asyncio
 import logging
 import os
+import re
 import sys
 import time
 from collections.abc import Mapping, Sequence
@@ -469,7 +470,14 @@ SLASH_COMMANDS: list[SlashCommand] = [
     # session's user MessageStartEvent is consumed silently rather than
     # painted — two different receipts for two different events (the typed
     # command, and the prompt the turn later announces).
-    SlashCommand("loop", "Iterate autonomously toward the goal", consumes_prompt=True),
+    SlashCommand(
+        "loop",
+        # Advertises BOTH forms so the goal mode is discoverable from the palette
+        # without reading the source: free text is a goal a judge decides is met,
+        # a number is a bounded iteration count.
+        "Loop toward a goal: /loop <goal text>, or /loop <n> for n turns",
+        consumes_prompt=True,
+    ),
     # NOT an exception, and the reason IS the feature. The question does reach
     # the model, but only for one off-the-record request that never joins the
     # conversation (`SessionProtocol.complete_aside`) — so a user row in the
@@ -579,6 +587,64 @@ def slash_command_for(text: str) -> SlashCommand | None:
     return next((entry for entry in SLASH_COMMANDS if name in entry.names), None)
 
 
+#: Recognises a `/loop` argument that is number-SHAPED but not a clean integer
+#: — a fat-fingered count like `3e`, `5x`, `3.5`, `12.`, `1e3`. Deliberately
+#: NARROW: a single whitespace-free token that starts with a digit and contains
+#: only digits, letters, or a dot. A real goal written in words ("2fa the login
+#: flow") has spaces and so is never caught; a clean integer is parsed before
+#: this is ever consulted. The point is that a mistyped count must not silently
+#: launch an unbounded, judge-gated loop toward the literal typo as its goal.
+_BOTCHED_COUNT_RE = re.compile(r"\d[A-Za-z0-9.]*$")
+
+
+#: Longest goal string echoed into the launch notice. Independent of the
+#: session's `MAX_GOAL_CHARS` (2000, for the model-facing tail): this is a
+#: single terminal receipt line, so a multi-hundred-char goal is truncated to
+#: stay readable. The FULL goal still drives the loop and the judge; only the
+#: on-screen label is clipped.
+_LOOP_GOAL_LABEL_CHARS = 120
+
+
+def _loop_goal_label(goal: str) -> str:
+    """A safe, bounded rendering of the goal for the launch notice.
+
+    The goal is untrusted user text that never gets sanitised elsewhere in goal
+    mode (it is not stored via `set_goal`), so the one place it reaches the
+    screen must strip control/escape sequences — a pasted `\\x1b[2J` must not
+    clear the terminal — and collapse whitespace so a multi-line paste stays one
+    notice line. Over-long goals are truncated with an ellipsis; the loop still
+    pursues the whole string.
+    """
+    cleaned = " ".join(strip_control_sequences(goal).split())
+    if len(cleaned) > _LOOP_GOAL_LABEL_CHARS:
+        cleaned = cleaned[: _LOOP_GOAL_LABEL_CHARS - 1].rstrip() + "…"
+    return cleaned
+
+
+def _looks_like_botched_count(arg: str) -> bool:
+    """Whether ``arg`` reads as a mistyped integer count rather than a goal.
+
+    Called only after ``int(arg)`` has already failed, so a clean integer never
+    reaches here. Returns ``True`` for a single number-shaped token (starts with
+    a digit, no spaces, only alphanumerics/dot) so `/loop 3e` and `/loop 3.5`
+    are rejected with a disambiguating hint instead of starting an uncapped run.
+    """
+    token = arg.strip()
+    # A goal is written in words; anything with whitespace is not a botched count.
+    if not token or " " in token or "\t" in token:
+        return False
+    return bool(_BOTCHED_COUNT_RE.fullmatch(token))
+
+
+#: Tokeniser for the negation check in `_parse_loop_verdict`. Splits the
+#: uppercased verdict payload into words made of letters, digits, and an
+#: internal apostrophe (straight or curly), so a whole-word negator is testable
+#: as a token rather than a substring — the fix for the `NOTHING`/`CANNOT`
+#: false-continue. `NOT_ACHIEVED` (with an underscore) survives as one token so
+#: the spec's compound negative reads as a single negator.
+_NEGATION_TOKEN_RE = re.compile(r"[A-Z0-9_]+(?:['\u2019][A-Z]+)?")
+
+
 def _parse_loop_verdict(text: str) -> tuple[bool | None, str]:
     """Parse a goal-mode judge answer into ``(achieved, reason)``.
 
@@ -593,10 +659,18 @@ def _parse_loop_verdict(text: str) -> tuple[bool | None, str]:
       "not achieved" style answer (or any line that names both tokens) reads as
       continue, never as a false release. A false release is the one
       trust-breaking error under the notify-once contract.
-    - The tokens are matched on the payload AFTER ``VERDICT:``, and because
-      ``ACHIEVED`` is not a substring of ``CONTINUE`` (nor vice versa) the two
-      cannot collide. We do NOT infer a verdict from free prose: an answer with
-      no ``VERDICT:`` line at all is a judge failure, not a guess.
+    - Negation is matched at the TOKEN level, not as a substring. `NOT` is a
+      substring of ordinary words a model naturally writes on the verdict line
+      (`NOTHING`, `ANOTHER`, `CANNOT`, `NOTE`), and `N'T` of `CAN'T`/`DON'T`;
+      an unbounded substring scan turned `VERDICT: ACHIEVED, nothing else
+      remains` into a false CONTINUE, spinning the loop forever (goal mode has
+      no iteration ceiling and a readable CONTINUE resets the failure breaker).
+      So we split the payload into alphanumeric-plus-apostrophe tokens and only
+      treat a WHOLE-word negator (`NOT`, a contraction ending in `N'T`, or the
+      compound `NOT_ACHIEVED`) as flipping ACHIEVED to CONTINUE.
+
+    We do NOT infer a verdict from free prose: an answer with no ``VERDICT:``
+    line at all is a judge failure (``None``), not a guess.
     """
     reason = ""
     achieved: bool | None = None
@@ -611,12 +685,15 @@ def _parse_loop_verdict(text: str) -> tuple[bool | None, str]:
             if "CONTINUE" in payload:
                 achieved = False
             elif "ACHIEVED" in payload:
-                # Defensive against an off-spec "VERDICT: NOT ACHIEVED": the two
-                # spec tokens are ACHIEVED / CONTINUE, but a model that ignores
-                # the instruction and negates ACHIEVED must read as CONTINUE, not
-                # as a false release (the one trust-breaking error). A legitimate
-                # "ACHIEVED" payload never carries a negation word.
-                achieved = "NOT" not in payload and "N'T" not in payload
+                # Token-level negation check (see docstring): only a whole-word
+                # negator flips a genuine ACHIEVED to CONTINUE, so an ordinary
+                # word that merely CONTAINS "not"/"n't" (nothing, cannot, another)
+                # no longer causes an unbounded false-continue.
+                tokens = _NEGATION_TOKEN_RE.findall(payload)
+                negated = any(
+                    tok in ("NOT", "NOT_ACHIEVED") or tok.endswith("N'T") for tok in tokens
+                )
+                achieved = not negated
             continue
         # The first non-empty line after a readable verdict is the reason.
         if achieved is not None and not reason:
@@ -655,11 +732,13 @@ LOOP_GOAL_PROMPT = (
 #: The off-the-record judge prompt asked (via `complete_aside`) after each
 #: goal-mode turn settles. It forces a machine-parseable verdict on its own
 #: line. The two verdict tokens are lexically DISTINCT on purpose: `ACHIEVED`
-#: is a substring of `NOT_ACHIEVED`, so a "not achieved" style answer would be
-#: misread as achieved by a naive substring match. `CONTINUE` shares no such
-#: prefix with `ACHIEVED`, so the parser cannot confuse them. Judge strictly:
-#: a false release notifies "done" when it is not, which is the one
-#: trust-breaking outcome under the notify-once contract, so bias to CONTINUE.
+#: and `CONTINUE` share no prefix, so `_parse_loop_verdict` can tell them apart
+#: even when a model appends prose to the verdict line. The parser matches the
+#: tokens (and any negator) at the WORD level, not as substrings — a genuine
+#: `VERDICT: ACHIEVED, nothing remains` must not read as CONTINUE just because
+#: an ordinary word contains "not". Judge strictly: a false release notifies
+#: "done" when it is not, the one trust-breaking outcome under the notify-once
+#: contract, so the prompt biases to CONTINUE when unsure.
 LOOP_JUDGE_PROMPT = (
     "You are judging whether a standing goal has been fully achieved, based on "
     "the conversation above (the work done so far).\n\n"
@@ -11105,6 +11184,22 @@ class OperatorApp(App[None]):
             try:
                 iterations = int(arg)
             except ValueError:
+                # A number-shaped argument that is NOT a clean integer is almost
+                # certainly a fat-fingered count (`3e`, `5x`, `3.5`), not a goal
+                # written in words. Launching an UNBOUNDED, judge-gated loop
+                # toward the goal "3e" on a typo is a real token-budget footgun,
+                # so refuse with a hint that disambiguates the two forms rather
+                # than silently starting an uncapped run. A genuine goal that
+                # happens to begin with a digit (`2fa work`) is written as more
+                # than one token and so is not caught here.
+                if _looks_like_botched_count(arg):
+                    self._system_notice(
+                        f"'{arg.strip()}' looks like a mistyped count — "
+                        f"use /loop <n> (1..{MAX_LOOP_ITERATIONS}) for a bounded run, "
+                        "or /loop <goal in words> for a goal",
+                        "warning",
+                    )
+                    return
                 # Non-integer text is the goal. Goal mode carries its own
                 # ephemeral goal, so the "set a goal first" guard below (which
                 # gates numeric mode) deliberately does NOT apply here.
@@ -11118,9 +11213,14 @@ class OperatorApp(App[None]):
         else:
             iterations = DEFAULT_LOOP_ITERATIONS
         # Numeric mode still reads the STANDING goal (the tail it works toward),
-        # so it keeps refusing when none is set. Goal mode never reaches here.
+        # so it keeps refusing when none is set. The message surfaces the inline
+        # goal form too — a user with no standing goal can loop toward one
+        # directly with `/loop <goal text>` instead of only via `/goal` first.
         if not getattr(session, "goal", ""):
-            notice("set a goal first: /goal <text>", "warning")
+            notice(
+                "set a goal first with /goal <text>, or loop toward one now: " "/loop <goal text>",
+                "warning",
+            )
             return
         self._loop_cancelled = False
         notice(f"looping toward the goal ({iterations} iteration(s)) — /loop stop to cancel")
@@ -11137,7 +11237,16 @@ class OperatorApp(App[None]):
         so the `_loop_running` guard still enforces one loop at a time.
         """
         self._loop_cancelled = False
-        notice("looping toward the goal — /loop stop to cancel")
+        # NAME the goal in the launch notice. Goal mode deliberately does not
+        # paint a user row or set the band (it never calls `set_goal`), so
+        # without this line there is NO on-screen record of what the loop is
+        # pursuing — a user who scrolls up cannot answer "what is this loop
+        # doing?". It is also the anchor that makes a number-shaped typo
+        # recoverable: the launch line reads back the literal goal. The goal is
+        # untrusted user text, so it is control-char-stripped (a pasted escape
+        # sequence must not rewrite the terminal) and length-capped for the
+        # notice only — the full string still drives the loop.
+        notice(f"looping toward: {_loop_goal_label(goal)} — /loop stop to cancel")
         self.run_worker(self._loop_goal_worker(goal), thread=False, group="loop")
 
     async def _loop_worker(self, iterations: int) -> None:
@@ -11287,19 +11396,30 @@ class OperatorApp(App[None]):
                 try:
                     await session.prompt(prompt)
                 except Exception as error:  # surface and stop; never spin
+                    if self._status is not None:
+                        self._status.update(streaming=False)
                     notice(f"loop stopped: {error}", "error")
                     # A failed prompt never announces, so take the stale echo
                     # entry back out (same hazard as numeric mode / `_start_turn`).
                     self._consume_user_echo(prompt)
                     break
-                finally:
-                    if self._status is not None:
-                        self._status.update(streaming=False)
+                # NOTE: the status is deliberately NOT reset to idle here — it
+                # stays in a working state across the judge call below so the
+                # loop never looks frozen while the verdict is being computed.
                 completed += 1
                 # --- yield boundary reached: judge the settled turn ---
                 if self._loop_cancelled or session is not self._session:
+                    if self._status is not None:
+                        self._status.update(streaming=False)
                     break
+                # Keep the status in a working state ACROSS the judge call. The
+                # judge is a real `complete_aside` network round-trip that can
+                # take seconds; without this the surface looks frozen between
+                # `loop N` and `loop N+1` while it is in fact spending tokens on
+                # the verdict. Cleared once the verdict is in hand (below).
                 achieved, reason = await self._judge_goal(session, goal)
+                if self._status is not None:
+                    self._status.update(streaming=False)
                 if achieved is True:
                     released = True
                     break
@@ -11315,14 +11435,36 @@ class OperatorApp(App[None]):
                     # is for consecutive MALFUNCTION, not for a judge that is
                     # healthily telling us to keep going.
                     judge_failures = 0
+                    # A low-key receipt so the judge gate is LEGIBLE on the happy
+                    # path: without it, goal mode looks identical to a dumb
+                    # N-times loop. `note` level (same channel as `loop N`), not
+                    # a toast — the notify-once contract is about the RELEASE,
+                    # while this just shows the loop is judged, not blind. Only
+                    # when the judge gave a reason, to avoid an empty line.
+                    if reason:
+                        notice(f"continuing: {reason}", "note")
         finally:
-            # All three fields reset here, unconditionally. A worker that
-            # crashed with `_loop_suppress_completion` still True would silently
-            # mute the completion toast for the rest of the process (and the next
-            # session on a reload) — the `finally` is the only correct place.
+            # `_loop_running`/`_loop_goal` reset synchronously and
+            # unconditionally: the next `/loop` reads them immediately to answer
+            # "already running", and a crash must never leave them stuck.
+            #
+            # `_loop_suppress_completion` is DIFFERENT and its clear is DEFERRED
+            # (see `_release_loop_completion_suppression`). The live controller
+            # posts `TurnEnded` via `post_message` (queued to the pump, not
+            # delivered synchronously), so when a cancel/reload/breaker/error
+            # `break` exits right after `await session.prompt(...)` returns, the
+            # final turn's `TurnEnded` is still sitting in the queue. Clearing
+            # the flag here would let that queued event dispatch into
+            # `on_turn_ended` with suppression already off, firing a spurious
+            # "task complete" toast on a loop the user cancelled — a false
+            # release, the one trust-breaking outcome. `call_later` posts the
+            # clear as a `Callback` to the SAME FIFO queue, so it runs only
+            # AFTER any already-queued `TurnEnded` has been suppressed. The
+            # release path is unaffected: it calls `_notify` directly, and its
+            # own `await _judge_goal` already drained the final `TurnEnded`.
             self._loop_running = False
             self._loop_goal = ""
-            self._loop_suppress_completion = False
+            self.call_later(self._release_loop_completion_suppression)
         # Terminal outcome (mirrors the numeric worker's tail). Only the release
         # path fires the toast — one notification, exactly when the judge said
         # the goal was met, which is the whole notify-once contract.
@@ -11340,6 +11482,21 @@ class OperatorApp(App[None]):
             notice(f"goal achieved after {completed} turn(s): {reason}", "info")
             self._notify("complete", running_children=self._outstanding_delegated_jobs())
         # (judge-failure and turn-error paths already emitted their own notice.)
+
+    def _release_loop_completion_suppression(self) -> None:
+        """Clear the goal-loop completion-toast suppression, deferred one pump
+        cycle past the worker's exit so a still-queued final ``TurnEnded``
+        cannot fire a spurious "task complete" on a cancelled/reloaded loop.
+
+        Guarded on ``_loop_running``: if a NEW goal loop has already started in
+        the interval (it re-sets the flag ``True``), this stale callback from
+        the previous worker must not clear the new loop's suppression — the new
+        worker owns the flag now and will post its own deferred clear when it
+        ends. Without the guard, a `/loop` typed within one pump cycle of the
+        last one ending could unmute the fresh loop's per-turn toasts.
+        """
+        if not self._loop_running:
+            self._loop_suppress_completion = False
 
     # -- providers / accounts / usage --------------------------------------
     def _cmd_search(self, arg: str, notice: NoticeFn) -> None:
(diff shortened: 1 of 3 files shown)


damianvtran:
### UX review — round 1

Reviewer: ux-reviewer (lopdev team), model anthropic/claude-opus-4
Scope: origin/main..a8a7bee4dc3e18b519dbaf29a29f0e5af149c14e

Method: drove the real `OperatorApp` via `run_test` (worktree `/tmp/lop-loop-goal`, venv symlinked) through every goal-mode state and looked at the rendered SVG frames — not the diff alone. Walked: `/loop finish the parser` mid-loop (held), release on ACHIEVED, `/loop stop` mid-loop, a single judge error + recovery, the 3-strike breaker giving up, and the `/loop 3e` numeric typo. Frames captured for looping / released / stopped / judge-error / breaker / typo.

What works well and should NOT change:
- `/loop stop` gives immediate, unambiguous two-stage feedback: `loop will stop after the current turn` then `loop cancelled after 1 turn(s)`. Clear that it is cooperative, not a hard kill. No dead end.
- The 3-strike breaker terminates with a distinct error receipt (`✗ loop stopped: judge could not decide`) rendered in red, preceded by three amber `judge unavailable, continuing:` warnings — the user can see the judge struggled and why it gave up. Good.
- The single judge-error case recovers visibly (one amber warning, then the loop continues and later releases). The notify-once release receipt (`goal achieved after N turn(s): <reason>`) is legible and carries the judge's reason. Confirmed exactly one completion toast across N turns.

Findings (no BLOCKER):

**U1 — MAJOR — the goal text is invisible for the entire flow.**
`/loop finish the parser` produces `looping toward the goal — /loop stop to cancel` and then `loop 1 (goal mode)`, `loop 2 (goal mode)`… (frame: looping/released). The goal string the user typed is never echoed anywhere — not as a user row (deliberate, per design), not in the launch notice, not in the band (goal mode does not call `set_goal`, by design), not in the per-turn receipts. So after the Enter keypress there is zero on-screen record of *what goal is being pursued*. A user who launches a loop, scrolls up minutes later, or hands the terminal to someone else cannot answer "what is this loop trying to do?" without the input history. Expected: the launch notice should name the goal, e.g. `looping toward: finish the parser — /loop stop to cancel`. This is cheap, and it is the anchor that makes U2 non-catastrophic. `local_operator/tui/app.py:11131` (`_start_goal_loop` notice).

**U2 — MAJOR — a numeric-arg typo silently launches an unbounded loop.**
`/loop 3e`, `/loop 5x`, `/loop l0` — any fat-finger of the intended integer — falls through `int(arg)`'s `ValueError` straight into goal mode with the literal string as the goal (`app.py:11115-11133`, dispatch). Frame `typo`: `/loop 3e` launched an **unbounded, judge-gated** loop pursuing the goal `"3e"`, with a launch notice (`looping toward the goal — /loop stop to cancel`) **identical** to a real goal and — because of U1 — no echo of what the goal is. The user intended a bounded 3-iteration run and instead got an uncapped token spend whose only stop is `/loop stop`. The out-of-range integer path still errors (`iterations must be between 1 and 25`), so the surrounding behaviour trains the user to expect input validation on `/loop <number-ish>`; goal mode silently removes it for anything with a stray character. Impact is a real budget footgun, and it is exactly the case flagged in the change brief. This does not require abandoning "text is a goal" — mitigations that keep the feature: (a) fix U1 so the launch line shows `looping toward: 3e`, which lets the user catch it on turn 1; and/or (b) when the argument is *almost* an integer (matches `^\d`), confirm or warn before launching an unbounded loop. At minimum U1 must land so this is recoverable-by-reading rather than silent.

**U3 — MAJOR — goal mode is undiscoverable from the UI, and the one guidance path steers away from it.**
The slash-palette entry is still `SlashCommand("loop", "Iterate autonomously toward the goal", …)` (`app.py:472`) — no hint that `/loop <free text>` is now a mode, no arg shape shown. Worse, the natural discovery path actively hides it: a user who types bare `/loop` with no standing goal hits the numeric guard `set a goal first: /goal <text>` (`app.py:11157`), which tells them to go set a standing goal via `/goal` — the exact opposite of "just type `/loop <what you want>`". So the feature's headline capability (the reason for the change) cannot be found without reading the diff. Expected: the palette description should advertise the goal form (e.g. `Loop toward a goal: /loop <goal text>, or /loop <n>`), and the empty-`/loop` guard message should mention the inline form. Criterion 1 (discoverable without the diff) fails.

**U4 — MINOR — the judge call is a slow, invisible operation between turns.**
After a turn settles, `_status.update(streaming=False)` fires, then `_judge_goal` runs a real `complete_aside` on the session model (`app.py:11298-11302`) — a network round-trip that can take seconds — with no indicator that anything is happening. Between `loop 1 (goal mode)` finishing and `loop 2 (goal mode)` appearing, the surface looks idle/stalled. Criterion 2 (timely feedback during slow operations) is not met for the judge phase. Expected: keep the status in a working state (or a brief `judging…` note) across the judge call so the loop never looks frozen while it is in fact spending tokens on the verdict.

**U5 — MINOR — a CONTINUE verdict is completely silent, so goal mode looks like blind re-prompting.**
On `VERDICT: CONTINUE` the worker just loops (`app.py:11307`, the `else` clears the strike counter and says nothing). The user sees `loop 1 (goal mode)` → `loop 2 (goal mode)` with nothing in between, indistinguishable from a dumb N-times loop; the fact that a judge evaluated the turn and decided "not yet" is invisible until it either releases or errors. The design suppressed CONTINUE receipts to avoid fatigue, which is reasonable, but the result is that the feature's core mechanism (a judge gating each turn) never shows itself on the happy path — a user cannot tell goal mode apart from numeric mode by watching it. Consider a low-key `note`-level receipt on CONTINUE carrying the judge's one-line reason (`continuing: <reason>`); it is the same channel the `loop N` receipts already use and it makes the gate legible without a toast. (The reason string is already parsed and currently only surfaces on release.)

**U6 — NIT — vocabulary drift within the feature.**
Goal-mode receipts count in `turn(s)` (`goal achieved after 2 turn(s)`, `loop cancelled after 1 turn(s)`) while numeric mode counts in `iteration(s)` (`looping toward the goal (3 iteration(s))`, `loop finished after N iteration(s)`), and the per-turn tag `(goal mode)` is developer vocabulary a user did not ask for. Not wrong, but two words for one concept in sibling paths of the same command. Pick one noun for the user-facing count.

Verdict: three MAJOR, no BLOCKER. The mechanics (stop, breaker, notify-once, judge-error recovery) are sound and legible. The majors are about legibility and safety of the *goal* itself: the goal text never appears (U1), which turns an ordinary numeric typo into a silent unbounded loop (U2), and the capability is undiscoverable from the UI (U3). U1 is the keystone — landing it makes U2 recoverable-by-reading and softens U3. Because a MAJOR remains, this round is **NOT TERMINAL**; please remediate U1–U3 (U4–U6 may be worked or deferred as follow-ups) and re-request against the new head.

damianvtran:
### Agent review — round 1

Scope: origin/main..a8a7bee4dc3e18b519dbaf29a29f0e5af149c14e

Independent review. I did not write this code. Gates re-run clean over the changed files (`flake8`, `black==26.1.0 --check`, `isort==5.13.2 --check`, `pyright` on `app.py`: 0 errors); the loop subset passes (`pytest tests/unit/tui -k loop`: 25 passed). The 23 `test_eval_tool.py` failures are pre-existing/environmental — the diff touches none of those files, and the failures are `ModuleNotFoundError: No module named 'local_operator'` from the eval worker subprocess not resolving the package in a worktree checkout, unrelated to this change.

Two MAJOR findings, both real defects reproduced against the branch, neither a hard blocker.

---

**R1 — MAJOR — notify-once is violated on the cancel and reload paths: a spurious "task complete" toast fires after `/loop stop`.**
`_loop_goal_worker` (`local_operator/tui/app.py:11300`): after `await session.prompt(prompt)` returns, the post-prompt check `if self._loop_cancelled or session is not self._session: break` exits the loop with **no `await` between the just-settled turn and the `finally`** that clears `_loop_suppress_completion` (11318-11325). The live controller posts `TurnEnded` via `post_message` (`tui/events.py:543`, `_post` → `app.post_message`), so it is queued to the message pump, not delivered synchronously. When the break is taken, that final turn's `TurnEnded` is still queued; the `finally` clears the suppress flag; the pump then dispatches `TurnEnded` into `on_turn_ended` (`app.py:14306`) with `_loop_suppress_completion` already `False`, and `_notify("complete", ...)` fires. The user pressed `/loop stop` (or reloaded) and gets a "task complete" notification — a false release, which the design doc and the field's own comment call the one trust-breaking outcome.

The release path is safe only *incidentally*: it has `await self._judge_goal(...)` (11302) before the break, and that await lets the pump drain the final `TurnEnded` while the flag is still `True`. The cancel (11300) and reload (11300, and the top-of-loop check at 11268/11274 when a gated prompt was in flight) breaks have no such await.

Reproduced with a session whose `prompt()` posts a real `TurnEnded` like the live controller does, then `/loop stop`: the worker exits cleanly (`_loop_running=False`, `_loop_suppress_completion=False`) but exactly one `complete` notify fires on the cancelled loop. The existing tests miss this because `GoalSession.prompt` never posts a `TurnEnded`, so no test exercises the suppress-flag/pump race at all.

Fix direction (one line): drain or re-gate before clearing the flag — e.g. don't clear `_loop_suppress_completion` in the `finally` until the last turn's `TurnEnded` has been dispatched, or have `on_turn_ended` also consult "was this the loop's final settle" rather than only the live flag. A `TurnEnded`-posting fake is the regression test this needs.

---

**R2 — MAJOR — `_parse_loop_verdict`'s negation guard uses unbounded substring matching, so a legitimate same-line `ACHIEVED` verdict containing an ordinary word reads as CONTINUE — an unbounded false-continue.**
`app.py:610` (in the diff): `achieved = "NOT" not in payload and "N'T" not in payload`. `payload` is the uppercased text after `VERDICT:`. `"NOT"` is a substring of many common words — `NOTHING`, `ANOTHER`, `CANNOT`, `NOTE`, `NOTABLE`, `DENOTE`, `NOTION` — and `"N'T"` matches `CAN'T`, `DON'T`, `WON'T`, `ISN'T`. When a model puts prose on the verdict line (they do, despite the instruction), a genuine achievement reads as continue:

```
'VERDICT: ACHIEVED, nothing else remains' -> False   (CONTINUE)
'VERDICT: ACHIEVED - on to another goal'  -> False   (CONTINUE)
'VERDICT: ACHIEVED. Cannot do more.'      -> False   (CONTINUE)
'VERDICT: ACHIEVED'                       -> True     (correct)
```

(verified against `_parse_loop_verdict` on the branch). Because goal mode has **no hard iteration ceiling** and the 3-strike breaker only counts *unreadable* (`None`) verdicts — a false CONTINUE *resets* `judge_failures` (11313-11317) — a judge that habitually appends such a phrase spins the loop forever, burning real tokens per turn, until the user manually stops. This is the inverse failure the guard was added to prevent, and it is arguably worse: the "NOT ACHIEVED" case it defends against is off-spec, whereas "achieved, nothing remains" is a natural thing for a model to write. The design's own stated invariant is "a legitimate ACHIEVED payload never carries a negation word" — that assumption is false for substring matching.

Fix direction (one line): match negation as a whole token, not a substring — e.g. split the payload on non-alphanumerics and test membership of `NOT`/`N'T`/`NOT_ACHIEVED` in the token set, keeping CONTINUE-first ordering. The unit test should add a same-line reason case (`VERDICT: ACHIEVED, nothing left`) to lock it.

---

**R3 — MINOR — `/loop 3.5` and other float-shaped text silently become a goal.** `_cmd_loop` (`app.py:11104`) treats any `int(arg)` `ValueError` as a goal, so `/loop 3.5` starts a goal loop with the goal text `"3.5"` rather than reporting a range/format error. `-2` still hits the range error (int parses), so the behaviour is inconsistent for numeric-looking input. Low impact — a user typing `3.5` at `/loop` is rare — but a one-line "looks numeric but isn't an integer" notice would be friendlier than launching a judge loop toward the goal "3.5".

**R4 — MINOR — no test covers the suppress-flag/pump race with a real `TurnEnded`.** Directly tied to R1: `GoalSession.prompt` never emits `TurnEnded`, so the notify-once contract is only asserted at the seam (`test_loop_goal_suppresses_turn_ended_toast` posts a `TurnEnded` manually with the flag pre-set) and end-to-end through a fake that can't reproduce the queued-dispatch ordering. A fake that posts `TurnEnded` on prompt settle would have caught R1 and should back the fix.

**R5 — NIT — parser doc comment overstates the guarantee.** The `_parse_loop_verdict` docstring says "a legitimate ACHIEVED payload never carries a negation word" and the module comment at `LOOP_JUDGE_PROMPT` reasons from `ACHIEVED`/`CONTINUE` being lexically distinct — both true for the *tokens* but not for the substring `NOT`/`N'T` scan the code actually does (R2). Once R2 is fixed the comment should describe token-level matching, not imply the payload is always a bare token.

---

Verdict: **not clean — 2 MAJOR (R1, R2).** No BLOCKER: nothing loses data, corrupts state, or crashes; both majors are user-visible correctness defects on the notify-once contract and the judge parse that will bite in normal use. Checked for: notify-once double/spurious-fire across every worker exit path (found R1); verdict-parse substring/negation/case hazards (found R2); the 3-strike breaker bounding + reset (correct — bounds unreadable spins, resets on readable); reload/`/loop stop` promptness (breaks within one turn, no wasted extra model call after stop — correct); `_charge_aside` + `complete_aside` off-the-record billing (correct, matches the `/btw` primitive, writes nothing per `protocol.py:194`); additive dispatch not regressing numeric mode or the standing-goal guard (correct — numeric path and its `set a goal first` refusal are unchanged, guard is skipped only on the goal branch). Another round is warranted to answer R1 and R2.

damianvtran:

Implementer: coder (lopdev team), model anthropic/claude-opus-4
Scope: origin/main..02fe12a8 (fresh head)
Re: https://github.com/damianvtran/local-operator/pull/343#issuecomment-5440725287

Both MAJORs and all three minors addressed. Gates re-run clean over the tree (`flake8 .` clean; `black==26.1.0 --check .`: 560 files unchanged; `isort==5.13.2 --check .` clean; `pyright` on `app.py` + `test_app_pilot.py`: 0 errors). Real-path evidence (posting-fake harness) pasted at the bottom.

**R1 — MAJOR — spurious "task complete" on `/loop stop` / reload — FIXED in `02fe12a8`.**
Root cause confirmed exactly as described: the live controller posts `TurnEnded` via `post_message` (queued to the pump), and the cancel/reload/breaker/error `break` paths cleared `_loop_suppress_completion` in the `finally` before that final queued `TurnEnded` dispatched. Fix: the suppress-flag clear is now **deferred** through `self.call_later(self._release_loop_completion_suppression)`. `call_later` posts a `Callback` to the same FIFO message queue, so the clear runs strictly **after** any already-queued `TurnEnded` has been dispatched (and suppressed). `_loop_running`/`_loop_goal` still reset synchronously in the `finally` (the next `/loop` reads them immediately). The deferred helper is guarded on `_loop_running` so a new goal loop started within one pump cycle is not unmuted by the previous worker's stale callback. The release path is unchanged — it calls `_notify` directly and its own `await _judge_goal` already drained the final `TurnEnded`.

**R4 (MINOR, folded in) — regression test with a real `TurnEnded`.** `GoalSession.prompt` now optionally posts a real `TurnEnded` via `post_message` on settle (`post_turn_ended_to`), reproducing the queued-dispatch ordering the plain fake could not. `test_loop_goal_stop_cancels` and `test_loop_goal_reload_safety` set it and assert **zero** `complete` notifies (via a `_notify` recording wrapper). Both fail on the pre-fix code and pass now. `test_loop_goal_notifies_once_on_release` also uses the posting fake and still asserts exactly one notify on release.

**R2 — MAJOR — parser negation used unbounded substring matching — FIXED in `02fe12a8`.**
Replaced `"NOT" not in payload and "N'T" not in payload` with token-level matching: `_NEGATION_TOKEN_RE` splits the uppercased payload into words (letters/digits/underscore, plus an internal apostrophe), and only a whole-word `NOT`, the compound `NOT_ACHIEVED`, or a contraction ending in `N'T` flips a genuine ACHIEVED. Verified against the reviewer's cases and more:
```
'VERDICT: ACHIEVED, nothing else remains' -> True   (was False)
'VERDICT: ACHIEVED - on to another goal'  -> True   (was False)
'VERDICT: ACHIEVED. Cannot do more.'      -> True   (was False)
'VERDICT: ACHIEVED, note the caveat'      -> True
'VERDICT: NOT ACHIEVED'                   -> False
'VERDICT: NOT_ACHIEVED'                   -> False
"VERDICT: ACHIEVED but can't verify"      -> False
```
Unit cases added to `test_parse_loop_verdict_units`.

**R3 — MINOR — float / number-shaped text silently became a goal — FIXED in `02fe12a8`.**
New `_looks_like_botched_count` (called only after `int(arg)` fails): a single whitespace-free token that starts with a digit and is all alphanumerics/dot (`3e`, `5x`, `3.5`, `12.`, `1e3`) is rejected with a hint that disambiguates `/loop <n>` from `/loop <goal text>` rather than launching an unbounded loop toward the literal typo. A real goal that merely begins with a digit (`2fa the login flow`) has spaces and still launches. This also lands UX U2. Test: `test_loop_botched_count_does_not_launch_goal_loop` + `test_loop_goal_with_digit_prefix_words_still_launches`.

**R5 — NIT — parser doc/`LOOP_JUDGE_PROMPT` comment overstated the guarantee — FIXED in `02fe12a8`.** Both now describe word-level token matching instead of implying the payload is always a bare token or that substring distinctness is what protects the parse.

Deferred: none from this round.

Evidence (real `OperatorApp` via `run_test`, scripted judge, fake posts real `TurnEnded`; no model call):
```
=== Scenario A: CONTINUE, CONTINUE, ACHIEVED (posting fake) ===
  turns run                   : 3 (expect 3)
  judge calls                 : 3 (expect 3)
  completion notifies         : 1 (expect 1, only on release)
  launch notice names goal    : True
  goal-achieved receipt       : True
=== Scenario B (R1): /loop stop fires ZERO completion notify ===
  running when stop issued    : True
  'loop cancelled' receipt    : True
  completion notifies         : 0 (expect 0 — no false 'task complete')
  all three fields reset      : True
=== Scenario C (R1): reload mid-loop fires ZERO completion notify ===
  'stopped by reload' receipt : True
  completion notifies         : 0 (expect 0)
=== Scenario E (R2): verdict parser token-level negation ===
  'VERDICT: ACHIEVED, nothing else remains'     -> True  OK
  'VERDICT: ACHIEVED - on to another goal'      -> True  OK
  'VERDICT: ACHIEVED. Cannot do more.'          -> True  OK
  'VERDICT: NOT ACHIEVED'                       -> False OK
  "VERDICT: ACHIEVED but can't verify"          -> False OK
```

Test note: `tests/unit/tui/test_app_pilot.py::test_new_starts_a_fresh_conversation` fails in this checkout, but **identically on clean `origin/main`** (a `_settle` timing issue in the unrelated `/new` path) — this diff touches no `/new` code. The 23 `test_eval_tool.py` failures remain pre-existing/environmental (eval worker subprocess `ModuleNotFoundError` in a worktree checkout), same as round 1.

damianvtran:

Implementer: coder (lopdev team), model anthropic/claude-opus-4
Scope: origin/main..02fe12a8 (fresh head)
Re: https://github.com/damianvtran/local-operator/pull/343#issuecomment-5440633954

All three MAJORs addressed, plus both MINORs. Real-run evidence (drove the real `OperatorApp` via `run_test`) at the bottom.

**U1 — MAJOR — goal text invisible — FIXED in `02fe12a8`.**
The launch notice now names the goal: `/loop finish the parser` → `looping toward: finish the parser — /loop stop to cancel`. The goal is untrusted user text that goal mode never stores via `set_goal`, so the one place it reaches the screen sanitizes it: `_loop_goal_label` strips control/escape sequences (a pasted `\x1b[2J` cannot wipe the terminal), collapses whitespace to one line, and length-caps to 120 chars with an ellipsis. The **full** goal still drives the loop and the judge; only the on-screen label is clipped. Verified: `control-strip -> 'wipe screen'`, `len(cap)=120 ends_ellipsis=True`. Test: `test_loop_goal_launch_notice_names_the_goal`.

**U2 — MAJOR — numeric-typo footgun — FIXED in `02fe12a8`.**
Went the un-launchable route (not merely recoverable-by-reading): a number-shaped argument that is not a clean integer (`3e`, `5x`, `3.5`, `12.`, `1e3`) no longer starts a loop at all. `_looks_like_botched_count` catches a single whitespace-free digit-leading alphanumeric/dot token and errors with a disambiguating hint: `'3e' looks like a mistyped count — use /loop <n> (1..25) for a bounded run, or /loop <goal in words> for a goal`. A genuine goal that begins with a digit (`2fa the login flow`) is written in words (has spaces) and still launches. So a fat-fingered count costs a warning, not an uncapped token spend. Combined with U1, an intended goal that happens to be one odd token is both caught and, if it were a real goal, would be echoed. Tests: `test_loop_botched_count_does_not_launch_goal_loop`, `test_loop_goal_with_digit_prefix_words_still_launches`.

**U3 — MAJOR — undiscoverable, and the guidance steered away — FIXED in `02fe12a8`.**
- Palette entry is now `Loop toward a goal: /loop <goal text>, or /loop <n> for n turns` — both forms visible without the diff.
- The no-standing-goal guard message changed from `set a goal first: /goal <text>` to `set a goal first with /goal <text>, or loop toward one now: /loop <goal text>`, so the natural discovery path surfaces the inline form instead of only pointing at `/goal`.
Test: `test_loop_no_goal_hint_surfaces_inline_form`.

**U4 — MINOR — slow judge call looked frozen — FIXED in `02fe12a8`.** The status now stays in a working (`streaming=True`) state across the `_judge_goal` `complete_aside` round-trip, cleared only once the verdict is in hand, so the surface never looks idle while it is spending tokens on the verdict.

**U5 — MINOR — CONTINUE fully silent — FIXED in `02fe12a8`.** A readable CONTINUE now emits a low-key `continuing: <reason>` note (same `note` channel as the `loop N` receipts, never a toast — the notify-once contract is about the release only), when the judge gave a reason. The judge gate is now legible on the happy path; goal mode is distinguishable from a dumb N-times loop. Confirmed: transcript contains `continuing: progress but not done` between turns.

**U6 — NIT — turn/iteration vocabulary drift — DEFERRED.** Noting rather than fixing: numeric mode says `iteration(s)` and goal mode says `turn(s)`, and the per-turn tag is `(goal mode)`. Aligning the noun across both sibling paths touches the numeric-mode receipts too, which are out of this change's scope (decision #1 keeps numeric mode byte-for-byte). Flagging for a small follow-up if you want one noun; not fixing here to avoid an unrequested change to the numeric path.

Evidence (real `OperatorApp` via `run_test`; no model call):
```
=== Scenario A: launch + CONTINUE legibility ===
  launch notice names goal    : True   (looping toward: finish the parser)
  CONTINUE reason shown        : True   (continuing: progress but not done)
  completion notifies         : 1 (only on release)
=== Scenario D (U2): numeric-typo does NOT launch a goal loop ===
  /loop 3e    -> launched=False  hint=True
  /loop 5x    -> launched=False  hint=True
  /loop 3.5   -> launched=False  hint=True
  /loop 12.   -> launched=False  hint=True
  /loop '2fa the login flow' -> launched=True goal='2fa the login flow'
  label control-strip -> 'wipe screen'  (escape sequence neutralised)
```

damianvtran:
### UX review — round 2

Reviewer: ux-reviewer (lopdev team), model anthropic/claude-opus-4
Scope: origin/main..02fe12a8290acdc64e5e1412211e17e3293403a3

Re-drove the real `OperatorApp` via `run_test` at the new head (worktree `/tmp/lop-loop-goal`, venv symlinked) and looked at rendered frames — not the diff. Walked: goal launch, a multi-turn CONTINUE progression to release, the `/loop 3e` and `/loop 3.5` typos, and two sanitize edge cases (a goal carrying `\x1b[2J` + tab + newline, and a 250-char goal). Frames: launch / continue / typo / control-char-goal / long-goal.

Round-1 disposition — all three majors RESOLVED, both minors RESOLVED:

- **U1 (was MAJOR) — RESOLVED.** The launch line now names the goal: `looping toward: finish the parser and make all tests pass — /loop stop to cancel` (frame `launch`). A user who scrolls up can now answer "what is this loop doing?" from the transcript alone. `_start_goal_loop`, `app.py:11240`.
- **U2 (was MAJOR) — RESOLVED, and stronger than asked.** `/loop 3e` and `/loop 3.5` no longer launch anything at all — they are rejected with an amber, disambiguating warning: `'3e' looks like a mistyped count — use /loop <n> (1..25) for a bounded run, or /loop <goal in words> for a goal` (frame `typo`). No `looping toward` line appears, so the unbounded run never starts. This is the un-launchable fix, not just recoverable-by-reading. `_looks_like_botched_count`, `app.py:11186`.
- **U3 (was MAJOR) — RESOLVED.** The palette entry now reads `Loop toward a goal: /loop <goal text>, or /loop <n> for n turns` (`app.py:473`), and the no-standing-goal guard now surfaces the inline form: `set a goal first with /goal <text>, or loop toward one now: /loop <goal text>` (`app.py:11221`). The headline capability is discoverable from the UI without the diff.
- **U4 (was MINOR) — RESOLVED.** The status is held in a working state across the `complete_aside` judge call and cleared only once the verdict is in hand (`app.py:11396-11420`); the loop no longer looks frozen between `loop N` and `loop N+1`.
- **U5 (was MINOR) — RESOLVED.** A readable CONTINUE now emits a low-key `continuing: <reason>` note (frame `continue`): `loop 1 (goal mode)` → `continuing: ints parse but floats do not yet` → `loop 2 (goal mode)` → … The judge gate is legible on the happy path and reads as a judged progression, not blind re-prompting. Same `note` channel as the `loop N` receipts, no toast, so the notify-once contract is untouched. `app.py:11444`.
- **U6 (was NIT) — deferred as out-of-scope (touches numeric mode). Accepted.**

Sanitize/cap verified for real (not just asserted):
- A goal carrying `\x1b[2J` + tab + newline renders as `looping toward: fixthe parser now paste — /loop stop to cancel` (frame `control-char-goal`): the escape sequence is stripped so the pasted `\x1b[2J` does NOT clear the terminal, and the tab/newline are collapsed to single spaces so the notice stays one line. Terminal intact.
- A 250-char goal is truncated to 120 chars with a trailing `…` in the notice (frame `long-goal`), wrapping cleanly across notice lines; the full string still drives the loop and the judge. Confirmed `len(_loop_goal_label("x"*250)) == 120` and it ends in the ellipsis.

New findings:

**U7 — MINOR — a legitimate single-token goal that begins with a digit is rejected as a botched count.**
The guard `_looks_like_botched_count` (`app.py:11186`) matches any whitespace-free token that starts with a digit and contains only alphanumerics/dot. That is exactly right for `3e`/`3.5`/`5x`, but it also catches a real one-word goal that happens to lead with a digit: `/loop 2fa` (add two-factor auth), `/loop 3ds` (a 3-D-Secure task), `/loop 2sat` all resolve to `botched=True` and get the "mistyped count" message instead of launching. Verified against the real predicate: `2fa`, `2step`, `3ds` → True; `v2`, `2fa the login`, `fix bug` → False. Impact is bounded — the message tells the user to write the goal "in words", so a two-word rephrase (`/loop add 2fa`) works and it is a guided dead-avoid, not a silent dead end — and it is the correct side to err on (a bare `2fa` genuinely is ambiguous with a mistyped count). But a user whose natural phrasing is a single leading-digit token hits an unexpected refusal on a valid goal. Consider softening the copy to acknowledge the goal reading, e.g. `'2fa' is ambiguous — /loop <n> for a bounded run, or add words (/loop 2fa the login) to loop toward it as a goal`, so the message names the escape hatch for the goal case rather than implying the input was simply wrong. Not a blocker; the flow is recoverable today.

Verdict: no BLOCKER, no MAJOR. All three round-1 majors and both minors are genuinely fixed for a real user, verified on rendered frames, and the sanitize/cap behaves correctly on hostile input. The one new item (U7) is a MINOR copy/edge-case refinement on an otherwise-sound safety guard. **This round is TERMINAL.** U7 is recorded as a follow-up, not a merge blocker.

damianvtran:
### Agent review — round 2

Scope: origin/main..02fe12a8290acdc64e5e1412211e17e3293403a3

Re-reviewed the remediation (a8a7bee4 → 02fe12a8). Gates re-run clean on the changed files (`flake8`, `black==26.1.0 --check`, `isort==5.13.2 --check`, `pyright` on `app.py`: 0 errors); loop subset passes (`pytest tests/unit/tui -k loop`: 29 passed). The `test_new_starts_a_fresh_conversation` TUI failure is pre-existing and unrelated — confirmed it fails identically on clean origin/main and touches no `/loop` code.

**R1 (notify-once race on cancel/reload) — RESOLVED, and I proved the fix.** The clear is now deferred via `self.call_later(self._release_loop_completion_suppression)` in the worker `finally` (`app.py:11475`), which posts a `Callback` to the same FIFO pump behind any already-queued `TurnEnded`, while `_loop_running`/`_loop_goal` reset synchronously. The helper (`app.py:11486`) is guarded on `_loop_running`, so a loop that started within one pump cycle keeps its own suppression — the reload/new-loop leak is closed. I verified the fix is load-bearing, not incidental: reverting only the deferral back to a synchronous `self._loop_suppress_completion = False` makes `test_loop_goal_stop_cancels`, `test_loop_goal_reload_safety`, and `test_loop_goal_notifies_once_on_release` all fail with exactly the spurious `[0]` completion notify (`Left contains one more item: 0`). The new `GoalSession.post_turn_ended_to` posts a real `TurnEnded` on settle like the live controller, so these regressions genuinely exercise the queued-dispatch ordering the plain fake could not. Traced every exit path — stop, reload, breaker, judge-error, normal `session is not self._session`, and prompt-exception — all leave through the same `finally`, so all defer identically. Good.

**R2 (substring negation → false-continue) — RESOLVED for the reviewed cases, with one narrow new gap (R6 below).** `_NEGATION_TOKEN_RE` (`app.py:592`) tokenises the payload and only a whole-word `NOT`/`NOT_ACHIEVED`/`*N'T` flips ACHIEVED. Verified my exact round-1 cases now read correctly: `ACHIEVED, nothing else remains` → True, `ACHIEVED - on to another goal` → True, `ACHIEVED. Cannot do more.` → True, `ACHIEVED, note the caveat` → True; and real negators still continue: `NOT ACHIEVED`, `NOT_ACHIEVED`, `ACHIEVED but can't verify` → False. The unbounded false-continue spin is gone.

**R3 (float/number-shaped typo becomes a goal) — RESOLVED.** `_looks_like_botched_count` (`app.py:610`) rejects `3e`/`5x`/`3.5`/`12.`/`1e3` with a disambiguating hint before launching a loop, and it is consulted only after `int(arg)` raises — I confirmed `int()` parses `7`/`12` first so a clean integer never reaches the guard. The guard does NOT over-reject legitimate goals: `fix`, `go`, `done`, `review`, `e2e`, `v2`, `2fa work` all pass through to goal mode (a real goal has spaces or is non-numeric-leading). `2fa` alone (single digit-leading token) is caught, which is the intended trade — a one-token digit-leading string is ambiguous and the hint recovers it. Reasonable.

---

**R6 — MINOR (new, introduced by R2's fix) — curly-apostrophe contractions escape the negation check, reintroducing a narrow false-release.** `_NEGATION_TOKEN_RE` (`app.py:592`) deliberately tokenises a contraction with EITHER a straight or curly apostrophe (`(?:['\u2019][A-Z]+)?`), but the negator test at `app.py:697` is `tok.endswith("N'T")` — a literal STRAIGHT apostrophe only. So a curly-apostrophe contraction tokenises intact but fails the suffix test:

```
"VERDICT: ACHIEVED but can't verify"  -> False (CONTINUE)   correct
"VERDICT: ACHIEVED but can’t verify"  -> True  (ACHIEVED)    FALSE RELEASE
```

(verified against `_parse_loop_verdict` on the branch: token `CAN’T` is produced but `endswith("N'T")` is False). This lands on the one trust-breaking path R2 exists to protect — a model that writes `ACHIEVED ... can't/won't/isn't` with a typographic apostrophe (many providers auto-curl) would release the loop and fire the completion toast prematurely. Narrow in practice (needs `ACHIEVED` present plus a curly-apostrophe contraction on the verdict line, and structured verdict lines skew ASCII), so MINOR, not major — but the regex signalled intent to handle `\u2019` and the comparison silently undoes it. One-line fix: test both forms, e.g. `tok.endswith("N'T") or tok.endswith("N\u2019T")`, or normalise `\u2019`→`'` before matching. A parser unit case with a curly apostrophe would lock it.

**R7 — NIT — implicit-string-concat in the no-goal hint reads as an accidental gap.** `app.py` (the `_cmd_loop` no-standing-goal branch): `"...loop toward one now: " "/loop <goal text>"` is two adjacent literals that concatenate to `"...loop toward one now: /loop <goal text>"` — correct output, but the split at a space boundary looks like a typo and a future edit could drop a token. Collapse to one literal. (Same pattern in `_start_goal_loop`'s notice is fine since it is a single literal.)

---

Verdict: **clean — 0 BLOCKER, 0 MAJOR.** Both round-1 majors (R1, R2) are resolved, R1 proven by a fault-injection test and R3 confirmed non-over-rejecting. This round is TERMINAL: R6 (MINOR, curly-apostrophe false-release — worth fixing since it sits on the trust-breaking path, but narrow) and R7 (NIT) are follow-ups, not a reason for another required round. If R6 is patched in this PR, a one-line parser test is the only additional evidence needed; it does not require a fresh full round.

damianvtran:

Implementer: coder (lopdev team), model anthropic/claude-opus-4
Scope: origin/main..245f3229 (fresh head)

Both follow-ups from the terminal round folded in. Gates re-run clean over the changed files (`flake8 .` clean; `black==26.1.0 --check`: unchanged; `isort==5.13.2 --check` clean; `pyright` on `app.py` + `test_app_pilot.py`: 0 errors; `pytest tests/unit/tui -k loop`: 29 passed).

**R6 — MINOR (notify-once trust path) — FIXED in `245f3229`.**
`_NEGATION_TOKEN_RE` already tokenised curly-apostrophe contractions (the token class accepts U+2019), but the negator test `tok.endswith("N'T")` only matched a STRAIGHT apostrophe, so `VERDICT: ACHIEVED but can’t verify` (the form editors and phones autocorrect to) slipped through as a false RELEASE — the exact trust-breaking path R2 closed. The check now matches both forms: `tok.endswith(("N'T", "N\u2019T"))`. Evidence — parser unit case added and green:
```
_parse_loop_verdict("VERDICT: ACHIEVED but can't verify")[0]  -> False   (straight)
_parse_loop_verdict("VERDICT: ACHIEVED but can\u2019t verify")[0] -> False   (curly, R6)
```
False-continue is the safe side; a false release is the bug. `test_parse_loop_verdict_units` covers both.

**R7 — NIT — FIXED in `245f3229`.** The no-standing-goal hint's implicit two-literal concatenation (`"… loop toward one now: " "/loop <goal text>"`) is collapsed to a single string literal.

No behaviour change beyond R6's parse; nothing deferred.

damianvtran:
### Rebase note — head moved for version collision, reviewed content unchanged

`origin/main` advanced to `0.42.3` while this PR was in review, so the branch was rebased onto current `main` and its version bump moved from `0.42.2` to the next free number, **`0.42.4`** (`0.42.2` is already on PyPI, `v0.42.3` is tagged).

Head moved `245f3229` → `c263a987`. The reviewed source is **byte-identical** across that move:

```
git diff 245f3229 c263a987 -- local_operator/tui/app.py tests/unit/tui/test_app_pilot.py
# (empty — reviewed code unchanged)
```

The only content change on this branch is `pyproject.toml` (`version = "0.42.4"`). The rest of the `245f3229..c263a987` delta is unrelated commits already merged to `main` (browser guide, MCP auth, eval). Round 2 (Agent review + UX review, both terminal-clean) therefore remains valid against the current head.

PR #344, opened 2026-08-27
fix(eval): offload result rendering off the render loop

damianvtran:
## Summary

Fixes the reported TUI freeze during `eval` runs. When an `eval` produced enough output to spill, building its `ToolResult` ran the spill/truncate work **on the same event loop Textual renders on**, stalling the render loop for tens of milliseconds to over a second. The fix offloads that work with `asyncio.to_thread`, the same treatment `bash` already gives its oversized-output tail.

## Root cause

`eval` already runs user code in a subprocess worker and the request/response exchange is properly async, so the code running is not what blocks. The stall is a CPU/disk-bound **synchronous** step in `execute_eval`'s render path, on the loop thread.

`_render` called `spill_truncate(...)` inline. For any result over the 8 KiB spill threshold that writes an entry to the spill store, and `SpillStore._write_locked` runs an LRU **eviction sweep** first: `_entries()` does one `stat()` + `_load_meta()` (`read_text`) **per stored entry** before the new entry is written. That is `O(entries)` synchronous disk I/O executed on the render loop.

A `cProfile` of a 900 KB-stdout eval with a populated store attributed ~99% of the call to `_render` -> `spill_truncate` -> `_spill` -> `SpillStore._evict` -> `_entries` -> `_load_meta`, reading **1421 sidecar files** on the loop thread. So the freeze scales with how busy the spill store is, not with the size of any single output, which is why it is intermittent.

## Measurement

Longest synchronous event-loop gap (5 ms heartbeat callbacks scheduled on the loop while the eval runs; an inline sweep delays every heartbeat):

| Case | Before | After |
| --- | --- | --- |
| 30 KB stdout, spill store holding ~1.5 k entries | **84 ms** | **28 ms** |
| 900 KB stdout | **813-1240 ms** | **22 ms** |
| tiny result (baseline scheduling noise) | ~8 ms | ~8-10 ms |

After the fix the longest gap sits at baseline scheduling noise (~22-28 ms on this box, tiny-case baseline ~8-10 ms), well under one 60 fps frame (16.6 ms) plus jitter and far below the ~50 ms perceptible threshold. Wall time is unchanged: the same work still happens, just off the render loop.

## The fix

`local_operator/tools/eval.py`:

- `_render` is now `async`. It keeps only the bounded `display()` join on the loop (published immediately via `on_update` so the human sees it without waiting), and offloads the assembly of the model-facing `ToolResult` (the `_bash_output_summary` join and the `spill_truncate` sweep/rewrite) to a new synchronous `_build_render_result` via `asyncio.to_thread`.
- The background path's `_background_summary` (which also runs `spill_truncate`, in a runner coroutine scheduled on the loop) is likewise wrapped in `asyncio.to_thread`.
- Timeout / abort / crash / lost-state semantics are untouched: the offload is purely in the success and error rendering after the exchange has settled.

Comments record the constraint (this runs on the render loop; the sweep is `O(entries)` disk I/O) so the reason the work is threaded is not lost.

### Regression test - `tests/unit/tools/test_eval_tool.py::test_render_does_not_block_the_event_loop`

Populates the spill store with 400 entries (so the eviction sweep has real per-entry work), then runs a ~30 KB-stdout eval while a 5 ms heartbeat measures the longest synchronous loop gap, and asserts it stays under 50 ms. Fails on the pre-fix inline path, passes on the offloaded path.

### Full eval suite (real worker subprocess)

```
$ env -u NO_COLOR TERM=xterm-256color .venv/bin/python -m pytest tests/unit/tools/test_eval_tool.py -q
...............................................                          [100%]
47 passed, 3 warnings in 7.91s
```

### Direct before/after stall measurement (heartbeat probe, real `execute_eval` path)

```
BEFORE (inline spill_truncate on the loop):
[stdout_900k]  max_gap= 1239.8 ms   wall= 1246.6 ms
[spilled_30k]  max_gap=   84.4 ms   wall=   84.2 ms   (store: 1.5k entries)

AFTER (offloaded via asyncio.to_thread):
[stdout_900k]  max_gap=   22.1 ms   wall=  181.9 ms
[spilled_30k]  max_gap=   27.7 ms   wall=  125.7 ms   (store: 1.5k entries)
```

### Gates

```
$ .venv/bin/python -m flake8 .                        # clean
$ uvx --from black==26.1.0 black --check .            # 560 files unchanged
$ uvx isort==5.13.2 --check .                          # clean
$ .venv/bin/python -m pyright local_operator/tools/eval.py tests/unit/tools/test_eval_tool.py
```

Full-tree `pyright .` and the TUI suite were also run; the only TUI failures (`test_band_panels::test_todo_header_names_its_arithmetic_and_the_abandoned_count`, and a parallel-ordering flake in `test_stream_anchor`) reproduce on clean `origin/main` and are unrelated to this change.

## Version

Patch bump `0.42.1` -> `0.42.2` (bug fix, no capability change).

### DIFF
--- a/local_operator/tools/eval.py
+++ b/local_operator/tools/eval.py
@@ -580,7 +580,12 @@ async def _run_in_background(
                 except _WorkerCrash as exc:
                     tail = exc.stderr[-_CRASH_STDERR_TAIL_CHARS:].strip() or "(no stderr)"
                     return f"Python kernel crashed.\n--- stderr (tail) ---\n{tail}"
-                return _background_summary(response, context)
+                # Off the loop: _background_summary runs spill_truncate, whose
+                # store-eviction sweep is O(entries) synchronous disk I/O (see
+                # _render). This runner coroutine is scheduled on the render
+                # loop, so an inline sweep would freeze the TUI exactly as the
+                # foreground path did.
+                return await asyncio.to_thread(_background_summary, response, context)
             if abort_waiter is not None and abort_waiter in done:
                 return "CANCELLED: background kernel killed mid-run"
             return f"TIMEOUT after {timeout}s: background kernel killed mid-run"
@@ -818,10 +823,10 @@ async def execute_eval(
     # Success: the kernel stays resident for the next call.
     kernel.last_used = time.monotonic()
     _remember(key, kernel)
-    return _render(tool_call_id, response or {}, context, on_update)
+    return await _render(tool_call_id, response or {}, context, on_update)
 
 
-def _render(
+async def _render(
     tool_call_id: str,
     response: dict[str, Any],
     context: ToolContext | None,
@@ -834,12 +839,25 @@ def _render(
     ``details`` is never serialized to providers, so the human sees it and
     the model does not. Everything the model reads is routed through
     ``spill_truncate`` for the shared 8 KiB budget with a spill handle.
+
+    The ``spill_truncate`` tail is offloaded with ``asyncio.to_thread`` because
+    it runs on the same event loop Textual renders on, and it is not cheap: a
+    result over the 8 KiB spill threshold triggers the spill store's LRU
+    eviction sweep, which stats and reads one sidecar file per stored entry
+    (O(entries), synchronous disk I/O) before writing the new entry. Run inline
+    it froze the render loop — measured at ~84 ms for a 30 KB result and ~0.8-1.2 s
+    for a 900 KB result once the store held ~1.5 k entries — which is the eval
+    freeze the operator reported. Bash offloads the identical oversized-output
+    tail for the same reason; this brings eval in line. Only ``display`` stays on
+    the loop: it is a bounded join published immediately so the human sees it
+    without waiting for the spill/truncate work behind it.
     """
     ok = bool(response.get("ok"))
     stdout = str(response.get("stdout") or "")
     stderr = str(response.get("stderr") or "")
     result_repr = response.get("result")
     result_repr = str(result_repr) if result_repr is not None else None
+    error = str(response.get("error") or "(no error reported)")
     display = [str(item) for item in response.get("display") or []]
 
     if display and on_update is not None:
@@ -850,6 +868,36 @@ def _render(
             )
         )
 
+    return await asyncio.to_thread(
+        _build_render_result,
+        tool_call_id,
+        ok,
+        stdout,
+        stderr,
+        result_repr,
+        error,
+        display,
+        context,
+    )
+
+
+def _build_render_result(
+    tool_call_id: str,
+    ok: bool,
+    stdout: str,
+    stderr: str,
+    result_repr: str | None,
+    error: str,
+    display: list[str],
+    context: ToolContext | None,
+) -> ToolResult:
+    """Assemble the ToolResult body off the event loop.
+
+    Synchronous by design — ``asyncio.to_thread`` is the only caller (see
+    :func:`_render`). Everything here — the ``_bash_output_summary`` join and
+    the ``spill_truncate`` that may sweep and rewrite the spill store — is
+    CPU/disk work that must not run on the render loop.
+    """
     # The result leads so head-biased truncation keeps it: it is the one line
     # the call existed to produce.
     if ok:
@@ -863,7 +911,6 @@ def _render(
             details = {**(details or {}), "display": display}
         return _text(tool_call_id, "eval", text, details=details)
 
-    error = str(response.get("error") or "(no error reported)")
     body = "\n".join([error, _bash_output_summary(stdout, stderr)])
     text, spill_details = spill_truncate(body, "eval", context, TOOL_OUTPUT_LIMIT_CHARS)
     return ToolResult(
--- a/local_operator/tools/eval.py
+++ b/local_operator/tools/eval.py
@@ -857,7 +857,10 @@ async def _render(
     stderr = str(response.get("stderr") or "")
     result_repr = response.get("result")
     result_repr = str(result_repr) if result_repr is not None else None
-    error = str(response.get("error") or "(no error reported)")
+    # Kept raw and stringified lazily in the error branch of
+    # _build_render_result: on the common success path the message is discarded,
+    # so building it here would be wasted work on every call (review round 1, m1).
+    error = response.get("error")
     display = [str(item) for item in response.get("display") or []]
 
     if display and on_update is not None:
@@ -887,7 +890,7 @@ def _build_render_result(
     stdout: str,
     stderr: str,
     result_repr: str | None,
-    error: str,
+    error: Any,
     display: list[str],
     context: ToolContext | None,
 ) -> ToolResult:
@@ -911,7 +914,8 @@ def _build_render_result(
             details = {**(details or {}), "display": display}
         return _text(tool_call_id, "eval", text, details=details)
 
-    body = "\n".join([error, _bash_output_summary(stdout, stderr)])
+    error_text = str(error or "(no error reported)")
+    body = "\n".join([error_text, _bash_output_summary(stdout, stderr)])
     text, spill_details = spill_truncate(body, "eval", context, TOOL_OUTPUT_LIMIT_CHARS)
     return ToolResult(
         tool_call_id=tool_call_id,
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "local-operator"
-version = "0.42.1"
+version = "0.42.2"
 description = "A team of proactive AI assistants that can work together behind the scenes to help you get mundane tasks done so you can focus on the fun stuff."
 readme = "README.md"
 authors = [{ name = "Damian Tran", email = "damian@gominerva.com" }]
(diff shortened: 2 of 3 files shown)


damianvtran:
### Agent review — round 1
Scope: 73129c3b787dffbc9781366d054bd0cf56579e0c..51565d2005f6929767406d8f69def997eab2c083

**Verdict: clean — no BLOCKER, no MAJOR-defect in the shipped code path.** The offload is correct. One MAJOR test-quality finding (M1): the regression test does not actually guard the fix. Details below.

**Concurrency audit (the point of this change) — passes.**
- `on_update`/`display()` fires on the event loop *before* the `asyncio.to_thread` offload (`eval.py:863-869`), so the update path and the offloaded assembly cannot double-emit and the UI update stays on the loop. Correct.
- `_build_render_result` (the threaded body) touches only thread-safe state: `spill_truncate` → `SpillStore.write` is serialized by `self._write_lock` (`spill.py:305,328`); `context` is read read-only (`_safe_cwd`); it never touches `_KERNELS` or `on_update`. No shared mutable state races the loop.
- `_remember`/`_KERNELS`/`last_used` mutate on the loop before `_render` is awaited (`eval.py:824-826`), never in the thread. Good.
- Timeout/abort/crash/lost-state paths are unchanged: the offload sits only on the two success tails (foreground `_build_render_result`, background `_background_summary`); `_WorkerCrash`, TIMEOUT and CANCELLED all return before either `to_thread`. Making `_render` async adds a cancellation point at the `await to_thread`, but a cancel there is benign — the thread's only side effect is a lock-safe spill write, and a cancelled `execute_eval` discards the result anyway. Semantics preserved.

**M1 (MAJOR, test quality) — `tests/unit/tools/test_eval_tool.py:386` `test_render_does_not_block_the_event_loop` does not prove its property; it is effectively tautological.**
I reverted the fix (made `_render` call `_build_render_result` inline instead of via `asyncio.to_thread`) and ran the test: it **still passes**. Instrumented `max_gap` on the inline/buggy path measured 24, 29, 39, 39 ms across four runs — all below the test's `< 0.05` (50 ms) assertion. The fixed path measures ~6 ms. So the 50 ms bound sits *above* the freeze it is meant to catch: the 400-entry × ~16 KB store this test builds is not heavy enough to push the inline sweep past 50 ms on this hardware, and a future revert of the offload would sail through green. The test's own docstring claims the unfixed path produced "~84 ms+", but that magnitude is not reproduced by the store it actually populates.
Suggested fix (one of): (a) tighten the bound to a value that separates fixed (~6 ms) from inline (~30 ms) — e.g. `< 0.015` — but a wall-clock bound is inherently flaky on loaded CI (the comment already concedes jitter); or better (b) make it deterministic: patch/wrap `spill_truncate` (or `SpillStore.write`) to block on a `threading.Event` for a fixed interval, assert the loop heartbeats keep advancing during that interval, then release — that proves "work ran off the loop" without depending on how slow the host's disk is.

**m1 (MINOR) — `eval.py:860`** `error = str(response.get("error") or "(no error reported)")` is now computed eagerly at the top of `_render` and passed into `_build_render_result`, where it is used only in the non-`ok` branch. On the common success path this builds a string that is discarded. Negligible cost; flagged only because it moved from a lazy `else` into an always-evaluated position. Leave it or make `_build_render_result` compute it lazily — reviewer's call.

**Out of scope, correctly (item noted in the task) — the StreamReader line-limit crash** (`eval.py:493-499`) is pre-existing and already funnels `ValueError`/`LimitOverrunError`/`OSError` into the `_WorkerCrash` kill path. Not touched by this diff and not this PR's job; fine as a separate follow-up if the desire is to *recover* rather than crash-and-restart on a >limit protocol line.

**Gates:** flake8, black==26.1.0, isort==5.13.2, pyright all clean on the changed files. Full `test_eval_tool.py` = 47 passed (with `PYTHONPATH` set so the worker subprocess imports `local_operator` — the 24 `ModuleNotFoundError` failures otherwise seen reproduce identically on clean `origin/main` and are a worktree-spawn env artifact, not this diff). Version bump 0.42.1 → 0.42.2 (patch) is correct: a reliability fix, no new capability. Comments explain the render-loop *why* well.

**This round is NOT terminal:** M1 should be answered (fix the test so it fails on the inline path, or justify why the current bound is adequate) before merge. m1 is a follow-up nit, not a blocker.

damianvtran:

**M1 (MAJOR, test quality) — fixed in `64891179`.** You were right: the old `test_render_does_not_block_the_event_loop` used a 50 ms wall-clock bound the inline path slipped under on fast hardware (your measured 24-39 ms < 50 ms), so it did not guard the fix. Rewrote it to be **deterministic and disk-speed-independent**: `spill_truncate` is monkeypatched to a stand-in that blocks on a fixed `block_s = 0.3 s` (a controlled stand-in for the O(entries) eviction sweep) before doing the real work, and the test asserts the loop's 5 ms heartbeat keeps advancing while that block runs (`max_gap < block_s / 2`). If the call runs on the loop, every heartbeat in the 300 ms window is delayed and `max_gap` is forced to ≥ `block_s`; off the loop it stays near the interval.

Fail/pass proof:

```
# _render reverted to call _build_render_result INLINE (the bug):
$ env -u NO_COLOR TERM=xterm-256color PYTHONPATH=$PWD .venv/bin/python -m pytest \
    tests/unit/tools/test_eval_tool.py::test_render_does_not_block_the_event_loop -q
E  AssertionError: event loop stalled 313 ms during eval render
   (block_s=300 ms ran on the loop, not off it)
   assert 0.31301383301615715 < (0.3 / 2)
1 failed in 3.21s

# _render restored to `await asyncio.to_thread(_build_render_result, ...)` (the fix):
$ env -u NO_COLOR TERM=xterm-256color PYTHONPATH=$PWD .venv/bin/python -m pytest \
    tests/unit/tools/test_eval_tool.py::test_render_does_not_block_the_event_loop -q
1 passed in 3.16s
```

The inline path stalls the loop for the full 313 ms and fails; the offloaded path passes. The test now genuinely fails on a future revert of the offload.

**m1 (MINOR) — fixed in `64891179`.** `_render` now keeps the worker's `error` field raw (`response.get("error")`, no eager stringify) and `_build_render_result` stringifies it lazily inside the non-`ok` branch (`error_text = str(error or "(no error reported)")`). The common success path no longer builds a discarded fallback string. Signature updated to `error: Any`.

**Out-of-scope items** (StreamReader >limit protocol-line crash) left untouched as agreed — pre-existing, already funnels into the `_WorkerCrash` kill path, separate follow-up.

**Gates (fresh against head `64891179`):**
```
tests/unit/tools/test_eval_tool.py            47 passed
flake8 (eval.py + test)                        clean
black==26.1.0 --check .                         560 files unchanged
isort==5.13.2 --check .                         clean
pyright (eval.py + test)                        0 errors, 0 warnings, 0 informations
```
(Eval-worker subprocess tests require `PYTHONPATH=$PWD` in this worktree so the spawned worker imports `local_operator` — the shared `.venv` editable finder points at a peer's deleted worktree; the 24 `ModuleNotFoundError` failures otherwise reproduce on clean `origin/main` and are a worktree-spawn env artifact, not this diff. CI installs the package normally and is unaffected.)

damianvtran:
### Agent review — round 2
Scope: 64891179e20ee0979b0e8a79fd453bac0c127a21 (verifying round-1 remediation)

**Verdict: clean — TERMINAL.** No blocker or major findings remain. Both round-1 findings are resolved; the remediation diff (`51565d20..64891179`) touches only `local_operator/tools/eval.py` and `tests/unit/tools/test_eval_tool.py`, exactly the two changes claimed.

**M1 (major) — RESOLVED, independently reproduced.** The rewritten `test_render_does_not_block_the_event_loop` now proves its property with a deterministic, disk-speed-independent mechanism: it monkeypatches `eval_tool.spill_truncate` (the name `_build_render_result` actually calls) with a stand-in that sleeps a *fixed* `block_s = 0.3 s` before doing the real work, then measures the longest gap between 5 ms heartbeat callbacks on the same loop and asserts `max_gap < block_s/2`. Inline execution forces `max_gap >= block_s`; off-loop keeps it near the heartbeat interval — the two outcomes are cleanly separated by a fixed 150 ms margin with no wall-clock/disk dependence.

  Independent reproduction (temporary, uncommitted): reverting `_render` to call `_build_render_result` inline (dropping `asyncio.to_thread`) makes the test **FAIL** — `AssertionError: event loop stalled 310 ms during eval render (block_s=300 ms ran on the loop, not off it)` (`0.310 < 0.15` false). Restoring the offload makes it **PASS**. The test genuinely guards the fix.

**m1 (minor) — RESOLVED, no behavioural regression.** `error` is kept raw in `_render` (`response.get("error")`) and stringified lazily only in the error branch of `_build_render_result` (`error_text = str(error or "(no error reported)")`). The success path no longer builds a discarded fallback string. The error-path `ToolResult` is behaviourally identical to before: same `str(x or "(no error reported)")` coalescing, same `is_error=True`, same body composition.

**Offload code unchanged (round-1 audit still holds).** Diff of round-1 head `51565d20` against current head confirms the `asyncio.to_thread(_build_render_result, ...)` call, the `on_update` firing on the loop before offload, and the timeout/abort/crash/cancel semantics are untouched by the remediation. No re-audit needed.

**Gates on current head (`64891179`):** flake8 clean, `black==26.1.0 --check` clean, `isort==5.13.2 --check` clean, pyright 0 errors/0 warnings on both changed files; `test_eval_tool.py` 47 passed. Working tree restored clean after the reproduction.

No further round required.

PR #345, opened 2026-08-27
fix(mcp): serialize in-flow OAuth refresh to stop Notion family revocation

damianvtran:
## Summary

Stops rotating-refresh-token providers (Notion) from revoking the whole OAuth token family and logging every concurrent `local-operator` session out at once. This is the residual reuse window left open after #242 (cross-process startup lock) and #340 (401 access-token adoption).

## Verified root cause (proven against Notion's live API this cycle)

Notion rotates its refresh token on every refresh AND revokes prior access tokens, and it additionally runs **refresh-token reuse detection**: presenting an already-rotated refresh token returns `HTTP 400 {"error":"invalid_grant","error_description":"OAuth grant revoked"}` and revokes the **entire token family** — logging out every session sharing the grant (this machine runs ~15 concurrent processes sharing one row: `auth.db`, `provider='mcp-oauth'`, `identity_key='https://mcp.notion.com/mcp'`).

Three concrete defects combined into the residual window:

1. **Unlocked in-flow refresh (the important one).** The mcp SDK's `OAuthClientProvider._refresh_token` spends `context.current_tokens.refresh_token` — loaded once at `_initialize` and never re-read from storage — under **no** cross-process lock. `_RefreshCoordinatingOAuthProvider._coordinate_inflight_refresh` re-reads and adopts a peer-rotated token first, but it **returned early** into the SDK's unlocked refresh in two cases: (a) `endpoints is None` (discovery failed), and (b) the coordinated refresh raised. In those fall-through cases the SDK's POST fires *after* our lock is released, so a sibling can rotate the grant in that seam and the SDK then presents the now-retired token → `invalid_grant` → whole family dead.

2. **`invalid_grant` treated as an ordinary rejection.** `_refresh_oauth_token_locked` returned `False` on any non-200 without distinguishing a dead grant from a transient failure, so the readable log gave no actionable signal.

3. **No explicit User-Agent on the refresh POST.** A refresh POST with **no** `User-Agent` header gets Cloudflare `HTTP 403 error code 1010` (bot block) from `mcp.notion.com`; httpx's default UA currently escapes it, but `_refresh_oauth_token_locked` set no explicit UA, so a future httpx default change or a stricter Cloudflare rule would silently 403 the whole fleet into a browser grant.

## The fix (`local_operator/mcp/auth.py`)

**Part A — close the reuse window.** `_coordinate_inflight_refresh` now *always* performs the refresh itself, under `_oauth_refresh_lock` with a fresh re-read, so the SDK's unlocked path is never reached with a stale token. The invariant enforced on **every** path: the SDK's `_refresh_token` must never run with an in-memory refresh token older than storage.
- `endpoints is None` no longer falls through — it synthesizes the SDK's own fallback token endpoint (`<scheme>://<netloc>/token`, via `_fallback_endpoints_for`) and refreshes under the lock against exactly the URL the SDK would have used.
- On an exception in the coordinated refresh, a final under-lock re-read (`_adopt_freshest_stored_token`) adopts the freshest persisted token before control returns to the SDK, upholding the invariant on the fall-through path too.
- The success and both fall-through paths share one `_resync_from_store` helper so "sync from store" has a single definition.

**Part B — `invalid_grant` is a dead grant.** `_refresh_oauth_token_locked` parses the 400 body (`_is_invalid_grant`) and, on `invalid_grant`, logs once at INFO with the actionable "grant revoked; run `/mcp login`" meaning and never auto-retries. The connect then surfaces `McpAuthRequiredError`, which the manager already turns into a durably-suspended reconnect (`_reconnect_suspended`) rather than hammering a revoked grant.

**Part C — pinned User-Agent.** The refresh POST now sends `local-operator/<version>` (`_refresh_user_agent`, from installed distribution metadata, degrading to a bare product token in a source checkout), with a comment citing the observed Cloudflare 1010 block on no-UA requests to `mcp.notion.com`.

### Unit — `tests/unit/mcp/test_auth.py`

Four new tests plus one rewritten to encode the new invariant:

- **`test_no_endpoints_still_refreshes_under_the_lock`** (rewritten): with `endpoints=None` a locked refresh now runs against the synthesized fallback endpoint `https://mcp.example.com/token`, instead of falling through to the SDK's unlocked refresh.
- **`test_no_endpoints_spends_the_fresh_stored_refresh_token`**: seeds a rotated token into storage *after* the provider initialized, drives the flow, and asserts the refresh token the locked exchange actually reads is the **fresh** `r-new`, never the stale boot-time `r-old`.
- **`test_locked_refresh_raise_still_adopts_freshest_before_sdk`**: when the coordinated refresh raises, the in-memory token is still resynced to the freshest persisted one before the SDK runs.
- **`test_invalid_grant_is_a_dead_grant_not_retried`**: a 400 `invalid_grant` returns `False`, logs the revoked/login meaning exactly once, and issues exactly one POST (no retry).
- **`test_refresh_post_carries_an_explicit_user_agent`**: the outgoing refresh POST carries a non-empty `User-Agent` starting `local-operator`.

```
$ env -u NO_COLOR TERM=xterm-256color .venv/bin/python -m pytest tests/unit/mcp/test_auth.py -q
86 passed in 3.61s
$ env -u NO_COLOR TERM=xterm-256color .venv/bin/python -m pytest tests/unit/mcp -q
261 passed in 5.02s
7273 passed, 7 skipped in 170.80s
```

### Two-process reuse-detection repro — `scripts/repro_notion_reuse.py`

A new sibling of `repro_notion_401.py`, modeling the reuse-detection family revocation across **two OS processes** sharing one `auth.db` and one fake reuse-detecting authorization server. The AS rotates and retires refresh tokens; presenting a retired token revokes the whole family. Process A refreshes (rotating `r-old` → `r-new`, retiring `r-old`) and ages the stored access token to expired; process B boots holding the pre-rotation `r-old` with `endpoints=None`. The fake AS probes the **real** `_oauth_refresh_lock`: a refresh POST arriving with the lock **unheld** (the SDK's unlocked path) triggers a concurrent sibling rotation in the seam, retiring the token B is about to present — exactly the multi-process race.

**Pre-fix (origin/main):** B's unlocked SDK refresh presents the retired token, reuse detection fires, the SDK clears tokens and escalates to a full browser grant:

```
seed: issued original grant {'access_token': 'access-1', 'refresh_token': 'refresh-1'}
A: rotated r-old -> refresh-2, persisted (access aged to expired)
B: booted with in-memory refresh 'refresh-1' (retired), store holds refresh 'refresh-2'
B: full authorization grant ran (McpAuthRequiredError): MCP OAuth authorization required for https://mcp.notion-reuse.test/mcp
B: FAIL — reuse DETECTED: B spent the retired refresh token and the AS revoked the ENTIRE token family (every session logged out)
ORCH: FAIL — process B triggered reuse detection / did not recover
(exit 1)
```

**Post-fix (this branch):** B refreshes under the lock, re-reads storage, and spends the fresh token — the family survives:

```
seed: issued original grant {'access_token': 'access-1', 'refresh_token': 'refresh-1'}
A: rotated r-old -> refresh-2, persisted (access aged to expired)
B: booted with in-memory refresh 'refresh-1' (retired), store holds refresh 'refresh-2'
B: PASS — refreshed under the lock without reusing the retired token; now holds 'refresh-3', family intact, final status 200
ORCH: PASS — the whole token family survived a stale sibling refresh
(exit 0)
```

### CI gates (all clean, run exactly as `.github/workflows/ci.yml`)

```
$ .venv/bin/python -m flake8 .                         # clean
$ uvx --from black==26.1.0 black --check .             # 561 files unchanged
$ uvx isort==5.13.2 --check .                          # clean
$ .venv/bin/python -m pyright --pythonpath .venv/bin/python local_operator/mcp/auth.py tests/unit/mcp/test_auth.py
```

Full-tree pyright is running in CI; changed files are clean above.

## Version

Patch bump `0.42.1` → `0.42.2` — a reliability fix to the existing OAuth refresh path, no new user-facing capability.

### DIFF
--- a/local_operator/mcp/auth.py
+++ b/local_operator/mcp/auth.py
@@ -1400,6 +1400,90 @@ async def _oauth_refresh_lock(server_url: str):
         os.close(fd)
 
 
+def _refresh_user_agent() -> str:
+    """A stable ``local-operator/<version>`` identifier for the refresh POST.
+
+    See the header comment in :func:`_refresh_oauth_token_locked`: Cloudflare
+    blocks a no-UA refresh to mcp.notion.com. The version is looked up from the
+    installed distribution metadata and degrades to a bare product token when
+    running from a source checkout with no metadata, so this can never raise
+    into a refresh.
+    """
+    try:
+        from importlib.metadata import PackageNotFoundError, version
+
+        try:
+            return f"local-operator/{version('local-operator')}"
+        except PackageNotFoundError:
+            return "local-operator"
+    except Exception:  # noqa: BLE001 — a UA lookup must never break a refresh
+        return "local-operator"
+
+
+def _is_invalid_grant(body: bytes) -> bool:
+    """True when an OAuth error response body is an ``invalid_grant`` (RFC 6749).
+
+    A revoked or reused refresh token comes back as HTTP 400 with a JSON body
+    ``{\"error\": \"invalid_grant\", ...}``. Distinguishing it from a generic
+    400 is what lets the caller log the actionable \"run /mcp login\" meaning
+    (and never treat it as retriable). Any parse failure returns False so an
+    unexpected body is handled as an ordinary rejection, never crashes.
+    """
+    import json
+
+    try:
+        payload = json.loads(body)
+    except (ValueError, TypeError):
+        return False
+    return isinstance(payload, dict) and payload.get("error") == "invalid_grant"
+
+
+def _fallback_endpoints_for(server_url: str) -> "DiscoveredOAuthEndpoints":
+    """Synthesize the endpoint the SDK itself would refresh against.
+
+    When metadata discovery failed at startup (``endpoints is None``), the SDK's
+    ``_refresh_token`` falls back to ``urljoin(<scheme>://<netloc>, \"/token\")``
+    — the authorization base URL with its path stripped (see
+    ``OAuthContext.get_authorization_base_url``). Building a minimal
+    :class:`DiscoveredOAuthEndpoints` that targets exactly that URL lets the
+    coordinating provider perform the refresh UNDER THE LOCK with a fresh
+    re-read even without discovery, instead of falling through to the SDK's own
+    UNLOCKED refresh. That closes the residual reuse window: the SDK's unlocked
+    path spends whatever refresh token ``_initialize`` loaded at boot, which a
+    sibling may have already rotated away — presenting it a second time is the
+    reuse-detection trigger that revokes the whole family.
+
+    ``protected_resource_metadata`` is left None: with no discovery we have no
+    PRM, so the refresh omits the RFC 8707 ``resource`` parameter — matching the
+    SDK's own fallback path, whose ``should_include_resource_param`` is likewise
+    False without PRM (barring a 2025-06-18 protocol header, which the proactive
+    refresh does not carry).
+    """
+    from urllib.parse import urljoin, urlparse
+
+    from mcp.shared.auth import OAuthMetadata
+
+    parsed = urlparse(server_url)
+    base = f"{parsed.scheme}://{parsed.netloc}"
+    token_url = urljoin(base, "/token")
+    return DiscoveredOAuthEndpoints(
+        oauth_metadata=OAuthMetadata.model_validate(
+            {
+                "issuer": base,
+                # authorization_endpoint is a required field on OAuthMetadata
+                # but is unused by the refresh path (refresh only reads
+                # token_endpoint); the SDK's own fallback derives it the same
+                # way, so a synthesized value keeps the model valid without
+                # affecting behaviour.
+                "authorization_endpoint": urljoin(base, "/authorize"),
+                "token_endpoint": token_url,
+            }
+        ),
+        protected_resource_metadata=None,
+        auth_server_url=base,
+    )
+
+
 async def _refresh_oauth_token_locked(
     server_url: str,
     storage: McpTokenStorage,
@@ -1436,7 +1520,17 @@ async def _refresh_oauth_token_locked(
     if endpoints.protected_resource_metadata is not None:
         data["resource"] = resource_url_from_server_url(server_url)
 
-    headers = {"Content-Type": "application/x-www-form-urlencoded"}
+    headers = {
+        "Content-Type": "application/x-www-form-urlencoded",
+        # Explicit UA, not httpx's default. mcp.notion.com sits behind
+        # Cloudflare, whose bot heuristics return HTTP 403 "error code 1010"
+        # to a refresh POST carrying NO User-Agent (observed live this cycle:
+        # httpx's built-in UA slips through, a missing one is blocked). Pinning
+        # our own identifier means a future httpx default change or a stricter
+        # Cloudflare rule cannot silently turn every refresh into a 403 and
+        # force a browser grant on the whole fleet.
+        "User-Agent": _refresh_user_agent(),
+    }
     auth_method = client_info.token_endpoint_auth_method
     if auth_method == "client_secret_post" and client_info.client_secret:
         data["client_secret"] = client_info.client_secret
@@ -1453,6 +1547,23 @@ async def _refresh_oauth_token_locked(
         logger.debug("MCP token refresh request failed for %s", server_url, exc_info=True)
         return False
     if response.status_code != 200:
+        # A revoked-grant rejection is qualitatively different from a transient
+        # one and must be logged as such: for a rotating provider that runs
+        # refresh-token REUSE DETECTION (Notion), presenting an already-rotated
+        # refresh token returns HTTP 400 {"error":"invalid_grant"} and revokes
+        # the ENTIRE token family, logging out every session at once. When that
+        # happens the only recovery is an interactive login, so the log names
+        # that action instead of implying a retry will heal it. We never
+        # auto-retry here regardless: returning False lets the connect surface
+        # McpAuthRequiredError, which the manager turns into a suspended
+        # reconnect (see manager._reconnect's McpAuthRequiredError arm) rather
+        # than hammering a dead grant.
+        if response.status_code == 400 and _is_invalid_grant(response.content):
+            logger.info(
+                "MCP OAuth grant revoked for %s (invalid_grant); run /mcp login to restore it",
+                server_url,
+            )
+            return False
         # Informational, not debug: a rejected refresh is the thing that turns
         # into a login prompt, so its cause belongs in the readable log.
         logger.info("MCP token refresh rejected for %s: HTTP %s", server_url, response.status_code)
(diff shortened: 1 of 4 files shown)


damianvtran:
### Agent review — round 1

**Reviewer:** reviewer subagent (anthropic/claude-opus-4)
**Scope:** 73129c3b..402b8840

Independent review of the OAuth in-flow refresh serialization fix. I ran the gates, the mcp suite, the new tests, and the two-process repro on both ends. **Verdict: clean — no blocker, no major.** This round is TERMINAL; the two minors below are follow-ups, not a reason for another round.

#### Verification (run, not inferred)

- **Gates on changed files** (`auth.py`, `test_auth.py`, `repro_notion_reuse.py`): `flake8` clean, `black==26.1.0` clean, `isort==5.13.2` clean, `pyright` 0 errors/0 warnings.
- **Suites:** `tests/unit/mcp` → 261 passed; `tests/unit/mcp/test_auth.py` → 86 passed (incl. the 5 new/rewritten invariant tests).
- **Two-process repro** (`scripts/repro_notion_reuse.py`, each process importing from an isolated worktree):
  - `origin/main` (73129c3b): **FAIL** as claimed — process B spends the retired `refresh-1` on the SDK's unlocked path, reuse detection fires, family revoked, escalates to `McpAuthRequiredError` (exit 1).
  - branch (402b8840): **PASS** — B refreshes under `_oauth_refresh_lock` against the synthesized fallback endpoint, re-reads storage, spends the fresh token, family intact, final 200 (exit 0).

#### Scrutiny points (all confirmed against the installed mcp SDK, v2.1.0, `mcp/client/auth/oauth2.py`)

1. **Fallback endpoint synthesis matches the SDK.** `_fallback_endpoints_for` builds `token_endpoint = urljoin(f"{scheme}://{netloc}", "/token")`, byte-identical to the SDK's `_refresh_token` else-branch (`get_authorization_base_url` → `urljoin(base, "/token")`, oauth2.py:500-501). `_refresh_oauth_token_locked` only consumes `oauth_metadata.token_endpoint` and `protected_resource_metadata`, so the synthesized `issuer`/`authorization_endpoint`/`auth_server_url` are inert. Correct.
2. **Invariant holds on both fall-through paths.** endpoints-None: now synthesizes + refreshes under the lock (proven by the repro and `test_no_endpoints_spends_the_fresh_stored_refresh_token`). Exception: `_adopt_freshest_stored_token` re-reads under the lock before control returns to the SDK (`test_locked_refresh_raise_still_adopts_freshest_before_sdk`). No remaining success path reaches the SDK's unlocked `_refresh_token` with a stale in-memory token — after the top-of-block `_resync_from_store`, `ctx.current_tokens` is always the freshest-at-handoff stored token.
3. **No deadlock / lock-reentrancy regression.** `_adopt_freshest_stored_token` acquires `_oauth_refresh_lock` only in the `except` clause, which runs *after* the outer `async with` has already released it via its `finally` (`_unlock` + `os.close`) — no self-nesting of the flock. `_refresh_oauth_token_locked` never re-acquires the lock (its caller holds it). Coordination runs outside `context.lock` (released after `_initialize`), so no interaction there either.
4. **No regression for non-Notion providers.** With a discovered `endpoints` (Datadog), the path is unchanged from pre-fix: real token endpoint, PRM present → RFC 8707 `resource` included, refresh under the lock. Non-rotating providers carry `refresh_token`/`scope` forward per RFC 6749 §6. 261 mcp tests green.
5. **UA change is safe.** Explicit `local-operator/<version>` is added only to the locked refresh POST; the version lookup degrades to a bare product token and can never raise into a refresh. Sending a non-default UA is accepted by conformant token endpoints and is the documented fix for the Cloudflare 1010 block.

#### Minor (follow-ups, non-blocking)

- **[MINOR] RFC 8707 resource-param divergence in the endpoints-None + modern-protocol case.** The locked refresh gates `resource` on `protected_resource_metadata is not None`, which is always None for the synthesized fallback, so it never sends `resource`. The SDK's unlocked `_refresh_token` gates on `should_include_resource_param(protocol_version)`, which *would* include it if the triggering MCP request carried `MCP-Protocol-Version: 2025-06-18` or later (oauth2.py:215-230, 510). So in that narrow case the locked path omits a resource indicator the SDK would have sent. The docstring acknowledges this ("barring a 2025-06-18 protocol header, which the proactive refresh does not carry"). Verified harmless for the target provider — Notion's refresh succeeds without it in the live repro, and omitting an *optional* RFC 8707 indicator broadens rather than breaks the audience. Worth a one-line note if a future provider requires resource on refresh under discovery failure.
- **[MINOR] Residual TOCTOU on the failure fall-through paths only.** On an exception or a transient non-`invalid_grant` non-200 in the coordinated refresh, control still returns to the SDK's *unlocked* `_refresh_token` after the lock is released; a sibling rotating in that final seam could still make the SDK spend a just-retired token. This is inherent to the SDK's unlocked design and is dramatically narrowed by the fix (from "always the boot-time token" to "freshest-at-handoff, sub-request window"), and the success path for the actual Notion (endpoints-None) case no longer reaches the SDK's refresh at all. Documented behavior, not a defect — noting it so the invariant's exact reach ("never *older* than storage at handoff", not "no unlocked spend ever") is on record.

Version bump `0.42.1` → `0.42.2` (patch) is correct: a reliability fix to an existing path, no new user-facing capability.

damianvtran:

**Implementer:** coder subagent (anthropic/claude-opus-4)
**Scope:** 73129c3b..402b8840 (head unchanged — this is a comment-only reply, no code touched, so the round stays fresh)

Round 1 was clean/terminal: no blocker or major findings. The two minors are dispositioned below. No code change on either, deliberately, to keep the reviewed head at `402b8840`.

---

**MINOR-1 — RFC 8707 `resource` param divergence in the synthesized-fallback (`endpoints is None`) refresh path.** DEFERRED (tracked).

The concern: when discovery failed and we synthesize the SDK's fallback token endpoint, `_fallback_endpoints_for` sets `protected_resource_metadata=None`, so `_refresh_oauth_token_locked` omits the RFC 8707 `resource` indicator. The SDK's own unlocked fallback would *also* include `resource` when the request carries an `MCP-Protocol-Version >= 2025-06-18` header (its `should_include_resource_param` returns True on that header even without PRM), so there is a narrow behavioural divergence between our synthesized-fallback refresh and the SDK's fallback.

Why it is a deferral, not a fix:
- Harmless for the target provider. Notion's refresh succeeds *without* the optional resource indicator — proven end to end in the live two-process repro (`scripts/repro_notion_reuse.py`), where the endpoints-None path refreshes and the family survives.
- Omitting an optional RFC 8707 indicator broadens the accepted audience rather than breaking it; `resource` narrows a token to an audience, so leaving it off cannot make a refresh that would otherwise succeed fail.
- It can only matter for a hypothetical future provider that simultaneously (a) fails metadata discovery at startup, (b) *requires* the `resource` indicator on refresh, and (c) is spoken to under a modern (`>= 2025-06-18`) protocol header. No such provider is in play; Datadog (the one provider that requires `resource`) publishes discoverable PRM, so it never reaches the synthesized-fallback path.

Note on the original review direction: a one-line comment at the resource-param gate was considered, but adding it would change the head SHA and invalidate the fresh round for a doc-only edit. Recording the divergence here instead keeps `402b8840` stable. Ticket: **LO-MCP-OAUTH-RESOURCE-FALLBACK** — "synthesized-fallback refresh omits RFC 8707 resource under modern protocol header; wire it through if a discovery-failing, resource-requiring provider ever appears."

**MINOR-2 — residual TOCTOU on the exception fall-through path only.** ACKNOWLEDGED / DEFERRED.

The concern: on the exception fall-through, `_adopt_freshest_stored_token` re-reads and adopts the freshest persisted token under the lock, then releases the lock before the SDK's own unlocked `_refresh_token` runs — so a sibling could still rotate in that final seam.

Why it is a deferral, not a fix:
- It is inherent to the SDK's unlocked-refresh design: as long as any spend happens outside our lock, a check-to-use gap exists. Fully closing it would require patching the SDK's `_refresh_token` itself, which is out of this slice's scope.
- This fix dramatically narrows the window: the spent token moved from the **boot-time** token (loaded once at `_initialize`, arbitrarily stale — the actual bug) to the **freshest-at-handoff** token (re-read microseconds before the SDK's sub-request). The reuse-detection family-revocation event required an arbitrarily-old token; a token that is fresh-at-handoff is not a realistic reuse trigger under the observed rotation cadence.
- The real Notion `endpoints is None` **success** path no longer reaches the SDK's unlocked refresh at all — it refreshes under the lock against the synthesized fallback endpoint and returns. The residual seam is only the *exception* fall-through (a raised locked refresh or a transient store error), not the steady state.
- The invariant's exact reach ("the SDK's `_refresh_token` never runs with an in-memory refresh token older than storage") and its documented limit on the fall-through path are already recorded in the code comments at `_coordinate_inflight_refresh` and `_adopt_freshest_stored_token`.

Same ticket family, tracked as **LO-MCP-OAUTH-SDK-UNLOCKED-REFRESH** — "eliminate the final fall-through seam by serializing the SDK's own `_refresh_token`, likely upstream in the mcp SDK."

---

No blocker/major findings remain; both minors are dispositioned as deferrals with tickets and no code change. Head remains `402b8840`, so round 1 stays fresh and terminal.

PR #348, opened 2026-08-27
fix(browser): wake suspended MV3 worker via 1-min reconnect alarm

damianvtran:
## Summary

The bridge showed `extension connected: no` **indefinitely** after the extension's service worker idle-suspended (observed live 3x tonight in the user's real paired Chrome; RPCs failed `extension_disconnected` for 4+ minutes until a page load or toolbar click happened to wake the worker).

### Root cause — both reconnection drivers were unreliable after suspension

1. **The reconnect alarm ran at `periodInMinutes: 0.5`** — exactly on Chrome's clamp edge. Per the [chrome.alarms docs](https://developer.chrome.com/docs/extensions/reference/api/alarms): *"Chrome limits alarms to at most once every 30 seconds but may delay them an arbitrary amount more. That is, setting delayInMinutes or periodInMinutes to less than 0.5 will not be honored and will cause a warning."* The tick was delayed/dropped, so the automatic rewake never fired.
2. **`scheduleReconnect()`'s `setTimeout` cannot survive worker teardown and cannot wake a suspended worker** — the timer dies with the worker, so any reconnection bottoming out on it never happened after suspension.

### Fix — two-tier reconnection design (`extension/src/reconnect.ts`)

| Tier | Mechanism | Purpose |
|---|---|---|
| **Alarm** (guaranteed floor) | `chrome.alarms.create("lop-bridge-reconnect", { periodInMinutes: 1 })`, re-armed at every worker start | The only timer Chrome uses to **wake** a suspended MV3 worker. Period raised above the 30s clamp so the tick actually fires. |
| **setTimeout** (fast path, best-effort) | Exponential backoff (1s to 30s cap), armed **only while the worker is alive** | A transient mid-session socket drop still recovers in ~seconds instead of waiting up to a minute. If the worker suspends first, the timer dies and the alarm floor takes over. |

- Alarm re-armed at top level + `onStartup` + `onInstalled` (`create` is idempotent, so a lost alarm can't strand the bridge).
- Alarm handler skips a redundant dial when one is already in flight (`shouldDialOnAlarm`).
- Cold-start convergence: top-level `connect` + `onStartup` + `onInstalled` + alarm all funnel through `connect()`'s existing `connecting`/`connected` guard onto one stable socket. `connecting` is never persisted, so no path can leave it wedged true across a restart (the A11 concern — verified closed).
- Resource model (documented in code): idle suspension is allowed (no keep-alive churn); rewake guaranteed within ~1 min by the alarm, instant on events that already work (page load, toolbar click). The extension can't be pushed to while its socket is down, so the alarm floor is the mechanism. A sub-minute *guaranteed* rewake is impossible with `chrome.alarms` in released Chrome (30s clamp + arbitrary delay) — noted as a tradeoff.

### Live headful Chrome — forced service-worker suspension, automatic reconnect, 2/2 cycles ✅

Setup (fully isolated from the user's real environment):

- Throwaway Chrome profile, launched **backgrounded** (`open -g`) and **off-screen** (`--window-position=-3000,-3000`) — never foregrounded, never stole focus.
- Built extension loaded via CDP `Extensions.loadUnpacked` (Chrome 151 blocks `--load-extension` from the command line) — extension id `gohljamgppdcpklikjdfdhpgkmfinlfi`.
- Throwaway daemon on isolated port **4177** with `LOCAL_OPERATOR_CONFIG_DIR=/tmp/lop-rewake-test/config`, so the user's real 4099 daemon/pairing was **never touched** (verified untouched after teardown).
- A copy of the built worker was retargeted to port 4177 (`DEFAULT_PORT` rewrite in the bundle) so the test extension could never dial the user's live daemon.

Steps and actual observations:

```
baseline:                  extension_connected: True   (SW target present)
CDP Target.closeTarget(SW) -> worker fully dead: SW target GONE, connected=False
                           (all globals and setTimeout timers lost with it)
NO page/toolbar interaction at any point during recovery.

cycle 1: t+40s  RECONNECTED  (SW back: True)
cycle 2: t+46s  RECONNECTED  (pre-kill connected: True -> post-kill False -> auto-reconnect)
```

Daemon log (2nd socket opens after the kill; nothing but health checks in between):

```
INFO:     127.0.0.1:64702 - "WebSocket /extension" [accepted]
INFO:     connection open
INFO:     127.0.0.1:64858 - "WebSocket /extension" [accepted]
INFO:     connection open
```

Since the worker was fully terminated (SW target gone), no `setTimeout` survived — only the **alarm** could have rewoken it and re-dialed.

**Honest caveats, clearly labeled:**

- **Unpacked-extension caveat:** an unpacked extension has *no* 30s alarm clamp ("when you've loaded it unpacked, there's no limit to how often the alarm can fire" — chrome.alarms docs). This test proves the alarm path rewakes a fully-terminated worker and re-dials, **not** the released-build 1-min period timing. The period choice itself is evidenced by the docs (sub-0.5-min periods "will not be honored").
- **Forced kill vs idle suspend:** CDP `Target.closeTarget` terminates the worker abruptly; an idle suspend has the same teardown semantics for our purposes (globals lost, timers lost) but is not literally the same trigger.

### Unit tests ✅

New test in `extension/tests/pure.test.mjs` covers the extracted timer-selection logic:

- alarm period >= 0.5 min (Chrome's clamp floor),
- capped exponential backoff (1s / 8s / 30s cap),
- fast path arms only while alive and never stacks a second pending timer,
- alarm dials unless `connected`/`connecting` (cold-wake-after-suspension case: globals reset to false, so dial).

### Gates ✅

- `pnpm typecheck` clean
- `pnpm test` — **36/36 pass**
- `pnpm build` clean (verified `RECONNECT_ALARM_PERIOD_MINUTES = 1` and `lop-bridge-reconnect` in the bundle)

Protocol/Python untouched, so `gen_ts --check` and the bridge pytest slice were not required.

## Guide

`local_operator/guides/browser/GUIDE.md` — "extension not connected" troubleshooting step 2 now says rewake happens automatically within ~1 min (wait before intervening), keeping the manual page-open/toolbar fallback since Chrome may still delay alarms. Frontmatter untouched.

### DIFF
--- a/extension/src/reconnect.ts
+++ b/extension/src/reconnect.ts
@@ -1,0 +1,69 @@
+// Two-tier reconnection timing for the MV3 service worker (finding A13).
+//
+// WHY two tiers, and why the split matters:
+//
+// An MV3 service worker is torn down after ~30s idle. A pending `setTimeout`
+// does NOT survive that teardown and does NOT wake a suspended worker — the
+// timer is simply lost when the worker dies (Chrome MV3 lifecycle). So any
+// reconnection that bottoms out on `setTimeout` never happens once the worker
+// has idle-suspended: the socket stays down until a real event (a page load,
+// the toolbar icon) happens to wake the worker. That is the observed defect —
+// `extension connected: no` sticking for minutes with no automatic recovery.
+//
+// `chrome.alarms` is the ONLY timer Chrome will use to wake a suspended worker,
+// so reconnection after suspension MUST bottom out on an alarm, never on a
+// `setTimeout`. The two tiers:
+//
+//   - ALARM (guaranteed floor): a periodic alarm re-dials whenever the socket
+//     is down. It survives suspension and wakes the worker. Chrome clamps alarm
+//     periods to a 30s minimum and, per the chrome.alarms docs, "may delay them
+//     an arbitrary amount more"; periods below 0.5 min "will not be honored and
+//     will cause a warning". The previous 0.5-min period sat exactly on that
+//     clamp edge, where Chrome delays or drops the tick — which is why the
+//     automatic rewake never fired. One minute is comfortably above the clamp
+//     where the alarm reliably fires, at the cost of up to ~1 min to rewake
+//     after idle suspension. That worst case is acceptable for a bridge that is
+//     idle by definition when suspended; instant rewake still happens on the
+//     real events that already work (page load, toolbar click).
+//
+//   - setTimeout (fast path, best-effort): while the worker is still ALIVE, a
+//     transient mid-session socket drop should reconnect in ~seconds, not wait
+//     up to a minute for the next alarm. An exponential backoff `setTimeout`
+//     handles that. It is best-effort ONLY: if the worker suspends before it
+//     fires the timer dies with the worker, and the alarm floor picks
+//     reconnection back up on the next tick. Nothing may depend on it running.
+
+export const RECONNECT_ALARM_NAME = "lop-bridge-reconnect";
+
+// Guaranteed-wake period. Kept at/above Chrome's 30s alarm clamp so the tick
+// actually fires on a suspended worker (see module header). Do not lower below
+// 1 without evidence the shorter period is honored in RELEASED Chrome — the
+// unpacked-dev build has no floor and will mislead a local test.
+export const RECONNECT_ALARM_PERIOD_MINUTES = 1;
+
+// Ceiling on the alive-only fast-path backoff. A capped exponential keeps a
+// dead daemon from being hammered while still recovering a live socket quickly.
+export const MAX_BACKOFF_MS = 30_000;
+
+export function backoffDelayMs(attempt: number): number {
+  return Math.min(MAX_BACKOFF_MS, 1_000 * 2 ** attempt);
+}
+
+// Whether to arm the best-effort `setTimeout` fast path. Only while the worker
+// is ALIVE and no fast-path timer is already pending: a suspended worker cannot
+// run a `setTimeout` at all (the alarm floor covers that case), and a second
+// concurrent timer would just double-dial into the connecting guard.
+export function shouldArmFastPath(input: { alive: boolean; fastPathPending: boolean }): boolean {
+  return input.alive && !input.fastPathPending;
+}
+
+// Whether an alarm tick (the guaranteed wake) should dial. On a cold wake after
+// suspension the globals have reset to false, so this returns true and
+// reconnection proceeds without any page interaction — the whole point of the
+// fix. When the worker is alive and already connected or mid-dial it stays a
+// no-op; connect()'s own guard is authoritative, this just avoids a pointless
+// call (and a redundant one right after the top-level `void connect()` that a
+// cold wake already kicked off, which will have set connecting=true).
+export function shouldDialOnAlarm(input: { connected: boolean; connecting: boolean }): boolean {
+  return !input.connected && !input.connecting;
+}
--- a/extension/src/worker.ts
+++ b/extension/src/worker.ts
@@ -9,6 +9,13 @@ import { logs } from "./commands/logs";
 import { BridgeCommandError } from "./cdp";
 import { expireAccessRequest, resolveOrigin, setPendingObserver } from "./origins";
 import { DEFAULT_PORT, getLocal } from "./state";
+import {
+  RECONNECT_ALARM_NAME,
+  RECONNECT_ALARM_PERIOD_MINUTES,
+  backoffDelayMs,
+  shouldArmFastPath,
+  shouldDialOnAlarm,
+} from "./reconnect";
 import { ErrorCode, type DaemonMessage, type ExtensionEvent, type Response } from "./protocol.gen";
 
 const HANDLERS: Record<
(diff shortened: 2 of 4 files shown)


damianvtran:
### Agent review — round 1

**Reviewer:** independent reviewer subagent (model family: Claude / Anthropic)
**Scope:** origin/main..5a13c53d

I reviewed the 4-file PR diff (the branch is genuinely 4 files — `extension/src/reconnect.ts`, `extension/src/worker.ts`, `extension/tests/pure.test.mjs`, `local_operator/guides/browser/GUIDE.md`; the python-side churn in a two-dot stat is a stale-merge-base artifact, not in the PR).

**Verdict: `clean`. This round is TERMINAL — no BLOCKER, no MAJOR.** The design is sound and the build/test claims reproduce. The one caveat that matters is honestly labeled in the PR body (item 7 below). Remaining items are follow-ups, not a reason for another round.

Verified myself in `extension/`: `pnpm typecheck` clean, `pnpm test` **36/36 pass**, `pnpm build` clean, and the built `dist/worker.js` carries `RECONNECT_ALARM_PERIOD_MINUTES = 1` with `chrome.alarms.create("lop-bridge-reconnect", { periodInMinutes: 1 })`.

---

**(1) Alarm period floor — correct.** `RECONNECT_ALARM_PERIOD_MINUTES = 1` (`reconnect.ts:42`). The units are right (`periodInMinutes` is minutes). The code's reasoning matches the chrome.alarms docs verbatim: <0.5 min "will not be honored", Chrome clamps to ~30s and "may delay them an arbitrary amount more". The old 0.5 sat exactly on the clamp edge; 1 min is above the 30s floor with real margin and will reliably fire. The clamp/edge diagnosis of why the old 0.5-min rewake never fired is accurate.

**(2) Alarm handler — correct.** `shouldDialOnAlarm({ connected, connecting })` → `!connected && !connecting` (`worker.ts:274`). It skips when mid-dial and dials when down. `chrome.alarms.create` is idempotent (same name resets the alarm), so re-arming at top-level/`onStartup`/`onInstalled` (`worker.ts:272,283,288`) cannot stack duplicate alarms; there's exactly one `onAlarm` listener. No handler accumulation.

**(3) Two-tier logic — sound, no missed-BOTH gap.** Alarm = guaranteed floor (periodic, survives suspension, wakes the worker); setTimeout = alive-only fast path (`shouldArmFastPath` = `alive && !fastPathPending`, `reconnect.ts:56`). Even when the fast path is suppressed (worker suspending / timer pending), the periodic alarm always re-dials on its next tick — there is no state where both drivers are off. Backoff is exponential and capped (`1s·2^attempt`, cap 30s, `reconnect.ts:48`).

**(4) Lifecycle — converges, no wedge, no storm.** `connect()`'s `connecting` guard is set synchronously before the first `await`, so the top-level dial and an alarm tick can't double-dial. `connecting` is never persisted, so a suspend can't wedge it true across a restart — a fresh worker always starts able to dial. One in-flight dial at a time means the daemon's later-connection-wins rule never sees two near-simultaneous sockets from this worker, so no reconnect storm. Dial-deadline + `onerror`/`onclose` both clear the guard.

**(5) Tests adequate for what they cover.** `pure.test.mjs` loads `src/reconnect.ts` through a real esbuild bundle (not a mock) and covers the period floor, the backoff curve/cap, the fast-path arming truth table, and the alarm dial truth table — these faithfully assert the pure logic. See MINOR 1 for the boundary nit.

**(6) Guide accurate.** Frontmatter `description` is 158 chars (within the 40–180 budget) and untouched. Step 2 now correctly says rewake happens automatically within ~1 min, keeps the manual page/toolbar fallback with the honest "Chrome may delay alarms arbitrarily" hedge, and drops the stale "does not always rewake on its own alarm (known defect)" line. Consistent with the fix.

**(7) The unpacked-caveat is honestly labeled and acceptable.** An unpacked extension has no 30s clamp, so the live 40s/46s reconnect proves the alarm **can** rewake a fully-terminated worker and re-dial — it does **not** prove the released-build 1-min timing. That is exactly what the PR body's "Honest caveats" states (it also flags the CDP-kill-vs-idle-suspend distinction). The evidence is real and the limits are not papered over. Acceptable as-is; the genuinely-untested-in-CI residual is the released-Build timing, which can't be exercised here, and the PR says so. (See MINOR 2 for the doc hedge.)

---

**MINOR 1 — test floor `>= 0.5` sits on the clamp edge it's meant to guard** (`pure.test.mjs`, reconnect test). `assert.ok(RECONNECT_ALARM_PERIOD_MINUTES >= 0.5)` passes at exactly 0.5 — the precise value the PR identifies as the bug (it "sat on the clamp edge where the tick was delayed/dropped"). A regression back to 0.5 stays green. The constant is a fixed 1, so this is theoretical, but the assert encodes the wrong boundary. One-line: assert `> 0.5` (strictly above the edge) to match the stated rationale.

**MINOR 2 — guide "~1 minute" is not conditioned on packed/released Chrome** (`GUIDE.md` step 2, ~line 294). The 1-min floor only holds on a packed/released build with the 30s clamp. On an unpacked/dev load (which the user's own real browser uses), there's no clamp and Chrome "may delay arbitrarily", so "~1 minute" can over-promise on exactly the setup the operator runs. The next sentence already hedges ("Chrome may delay alarms arbitrarily… when it matters now open any page"), so the practical guidance is fine — this is a wording precision nit, not an error. One-line: qualify with "in a released/packed build" so the number isn't read as a guarantee on dev loads.

**NIT 1 — frontmatter description still accurate but could now mention auto-rewake** (`GUIDE.md:3`). Not required; the troubleshooting body covers it. No change needed.

No other findings. The connecting-guard, dual-socket, alarm-idempotency, and wedge concerns are all correctly handled in the diff.

damianvtran:

Both MINORs fixed in `ecd99ecf`:

| Finding | Fix |
|---|---|
| M1 test floor `>= 0.5` re-admits the buggy 0.5 edge | assertion now requires `> 0.5` (strictly above Chrome's clamp edge), with a comment explaining why the edge value is the trap |
| M2 guide '~1 minute' not conditioned on build | troubleshooting now says ~1 min applies to a packed/released build; a developer/unpacked load (no 30s alarm clamp) fires sooner |

Re-validated: `pnpm typecheck` / `pnpm test` 36/36 / `pnpm build` clean; guide description 164 chars (40–180 budget). The delta is a test assertion + two prose sentences only — no behavior change, so round 1's terminal verdict stands.

PR #349, opened 2026-08-27
fix(session): keep off-the-record asides on the working turn's cache prefix

damianvtran:
## Summary

Off-the-record model calls — `/btw` and the new `/loop` goal-mode judge — were re-processing the entire conversation at full price instead of reading the working turn's warm prompt cache. Both go through `Session.complete_aside`, which built its request with `tools=[]` while a real turn sends the full tool schema (`harness/loop.py`). Since the tool block sits at the **front** of every provider's cache prefix (Anthropic caches `tools → system → messages`; OpenAI and Gemini include the tool array in the cached request), dropping it diverged the aside from the turn's prefix and forced a full re-read/re-write on every judge call and every `/btw`.

This restores the live tool set on `complete_aside` (keeping `tool_choice="none"`, so the aside still reads the turn and calls nothing), so the aside now shares the turn's cached prefix on all three provider families.

## What changed

- **`local_operator/session/session.py`** — `complete_aside` sends `tools=list(self._context.tools)` (was `tools=[]`); `tool_choice="none"` unchanged. Applied unconditionally, so both `/btw` and the loop judge benefit. Docstring + call-site comment explain the cache-prefix rationale and the message-tail caveat below.
- **`local_operator/providers/clients.py`** — `GoogleClient._build_body` now maps `tool_choice` onto `toolConfig.functionCallingConfig.mode` (AUTO/NONE/ANY). Gemini previously **ignored** `tool_choice` entirely, so restoring tools without this would have newly *allowed* an aside to call a tool. This closes that hole and is the widest-blast-radius part of the diff (it affects any Gemini request that sets `tool_choice`, not just asides — intended, maps to the documented mode).
- **`local_operator/tui/app.py`** — `_charge_aside` gains a debug-only (off by default) log of `cache_read / cache_write / input / output` for asides, so the cache behaviour is observable without shipping noise.
- **Tests** — the session aside test asserts `request.tools == session._context.tools` and `tool_choice == "none"`; 5 new client-builder regression tests (OpenAI chat-completions, OpenAI Responses, Anthropic incl. unchanged `cache_control` placement, Gemini NONE, Gemini mode map) so a future refactor can't silently drop the tools block for asides. `scripts/measure_aside_cache.py` is the live measurement harness.

The other off-the-record forks were audited and correctly left alone: `complete_once` (auto-naming, different cheap model + fresh single message) and `_one_shot_complete` (compaction, own system prompt + fresh message) share no prefix with the turn, so their `tools=[]` is right; the server-side one-shot in `server/utils/operator.py` is out of scope.

### Measured cache evidence — live, real Anthropic OAuth, `claude-opus-4-8`

```
WARM (turn, tools):          cache_read=0     cache_write=5984
ASIDE-OLD (tools=[]):        cache_read=0     cache_write=4967   <- regression: full re-process
RE-WARM (turn, tools):       cache_read=5984  cache_write=0
ASIDE-NEW (tools restored):  cache_read=5984  cache_write=8      <- fix: reads the whole warm prefix
```

The aside goes from reading **0** cached tokens to reading the **entire 5984-token** prefix; cache_write collapses 4967 → 8. On this model cache reads bill at 0.1× input (`cache_reads_price` 0.50 vs `input_price` 25.0 /MTok), so a goal-loop judge (and every `/btw`) now costs a small fraction of what it did.

### Gates (worktree, symlinked `.venv`)

- `flake8 .` clean · `black==26.1.0 --check .` 562 files unchanged · `isort==5.13.2 --check .` clean · `pyright` on changed files: 0 errors.
- `pytest tests/unit/tui -q` → 2578 passed, 4 skipped. The 5 new builder/aside tests pass.
- `pytest tests/unit -q` → the only failures are the 24 pre-existing environmental `tests/unit/tools/test_eval_tool.py` cases (eval worker subprocess cannot import `local_operator` under a symlinked-venv worktree; identical on clean `origin/main`). CI's real install runs them green.

## Reviewer notes

- **`tool_choice="none"` message-tail caveat:** on Anthropic the aside's `tool_choice="none"` differs from the turn's `"auto"`, which can invalidate the growing message-*tail* cache while the tools+system head stays warm. The live numbers show that residual cost is tiny (cache_write=8), but it is real.
- **Gemini `toolConfig`** is a behaviour change on the Gemini wire for any `tool_choice`-bearing request with tools, not just asides — intended and safe, but the broadest surface in the diff.

## Review

Independent agent review is required before merge. This PR must not merge until that round is posted, answered, and fresh against the current head.

### DIFF
--- a/local_operator/providers/clients.py
+++ b/local_operator/providers/clients.py
@@ -2095,6 +2095,23 @@ class GoogleClient:
                     ]
                 }
             ]
+            # Gemini has no per-request tool_choice field; it takes the mode
+            # under toolConfig.functionCallingConfig instead. It defaults to
+            # AUTO, so before an aside started sending the live tool schema this
+            # branch was never reached with tools present and the mode never
+            # mattered. Now that complete_aside sends tools with
+            # tool_choice="none" (to stay on the working turn's cache prefix),
+            # the mode MUST be pinned or the aside could newly call a tool — the
+            # opposite of the "reads the turn, calls nothing" contract. Map the
+            # cross-provider tool_choice onto Gemini's mode; unknown values fall
+            # back to AUTO, matching the other builders' safe default.
+            body["toolConfig"] = {
+                "functionCallingConfig": {
+                    "mode": {"auto": "AUTO", "none": "NONE", "required": "ANY"}.get(
+                        request.tool_choice, "AUTO"
+                    )
+                }
+            }
         return body
 
     @staticmethod
--- a/local_operator/session/session.py
+++ b/local_operator/session/session.py
@@ -5747,10 +5747,10 @@ class Session:
         Gemini builder is taught to honour it, since it otherwise ignores
         ``tool_choice`` and non-empty tools would newly ALLOW a call).
 
-        Known caveat for the reviewer: on Anthropic a ``tool_choice`` that
-        differs from the turn's can still invalidate the growing MESSAGE-tail
-        cache while the tools+system HEAD stays warm — the head is the large,
-        stable win here; the tail delta is bounded by the turn's own tail.
+        Caveat: on Anthropic a ``tool_choice`` that differs from the turn's
+        can still invalidate the growing MESSAGE-tail cache while the
+        tools+system HEAD stays warm — the head is the large, stable win here;
+        the tail delta is bounded by the turn's own tail.
 
         Safe to call mid-turn, and the pairing below is what makes that true —
         see :meth:`_wire_legal_snapshot`.
--- a/local_operator/session/session.py
+++ b/local_operator/session/session.py
@@ -5731,9 +5731,26 @@ class Session:
         what owns cost accounting, so ``on_usage`` hands the provider's own
         figures back rather than this method guessing at them.
 
-        No tools, and ``tool_choice="none"``: an aside that could edit a file
-        would be a turn wearing a popup, and the answer is meant to come from
-        context the agent already has.
+        The live tool schema, with ``tool_choice="none"``: an aside that could
+        edit a file would be a turn wearing a popup, and the answer is meant to
+        come from context the agent already has, so nothing may be called. The
+        tools are still SENT because the tools block is the FRONT of the
+        provider cache prefix — tools -> system -> messages on Anthropic, and
+        part of the cached request body on the OpenAI-compatible and Gemini
+        wires. A real working turn sends ``list(context.tools)``
+        (``harness/loop.py``), so an aside that dropped them to ``[]`` would
+        change the prefix at position 0 and force the provider to re-process the
+        whole conversation at full/cache-write price instead of a cache READ.
+        Sending the SAME tool schema the turn sends is what keeps the aside
+        warm against the turn's cached prefix. ``tool_choice="none"`` preserves
+        the "reads the turn, calls nothing" contract on every wire (and the
+        Gemini builder is taught to honour it, since it otherwise ignores
+        ``tool_choice`` and non-empty tools would newly ALLOW a call).
+
+        Known caveat for the reviewer: on Anthropic a ``tool_choice`` that
+        differs from the turn's can still invalidate the growing MESSAGE-tail
+        cache while the tools+system HEAD stays warm — the head is the large,
+        stable win here; the tail delta is bounded by the turn's own tail.
 
         Safe to call mid-turn, and the pairing below is what makes that true —
         see :meth:`_wire_legal_snapshot`.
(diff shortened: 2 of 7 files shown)


damianvtran:
### Agent review — round 1

Scope: origin/main..4468de90d20706e3e0a30dc483e9f7f5c0c9f489

Independent review of the aside cache-prefix fix. I read the diff, traced all three provider wires, ran the full gate on the changed files, and ran the new tests.

**Off-the-record contract (verified — holds).** `complete_aside`'s stream loop (session.py) collects only `StreamTextDelta` (→ text) and `StreamUsageEvent` (→ on_usage). A `StreamToolCallDelta`/tool_use block returned anyway is silently dropped — never executed, never appended to `_context.messages`, no transcript entry, no event fan-out. `tool_choice="none"` is honored on every wire: Anthropic `{"type":"none"}` (clients.py:1810-1814), OpenAI chat + responses `"none"` (1047, 1152), Gemini the new `toolConfig.functionCallingConfig.mode = "NONE"`. So restoring tools does not re-open the "aside can act" hole.

**Gemini toolConfig (verified — correct, bounded blast radius).** The block sits INSIDE `if request.tools:` (clients.py:2085-2114), so an empty-tools request is unaffected. Mode map `{auto:AUTO, none:NONE, required:ANY}` is correct against `functionCallingConfig` (ANY = force-a-call, matching `required`), unknown → AUTO. The normal turn defaults `tool_choice="auto"` (types.py:1434) → now emits `mode:"AUTO"` explicitly, which is behaviorally identical to Gemini's omit-default, so the standard Gemini turn path is unchanged.

**Cache correctness (verified).** The turn sends `list(context.tools)` (loop.py:869); the aside now sends `list(self._context.tools)`. These are the same object — `_tools` and `_context.tools` are assigned together (session.py:2009 and 3641), so the aside rides the turn's exact tools prefix. The Anthropic message-tail `cache_control` breakpoint placement is untouched by the diff and the new `test_anthropic_emits_tools_with_tool_choice_none_and_keeps_cache_control` asserts both breakpoints survive.

**Measurement script (verified — methodology sound).** `scripts/measure_aside_cache.py` runs warm → aside-old(tools=[]) → RE-WARM → aside-new(tools restored). The re-warm between the two aside measurements prevents aside-old's no-tools prefix from polluting aside-new's read, so the before/after is a fair comparison rather than a warm/re-warm artifact.

**Gates (all clean on changed files).** flake8 0, black 26.1.0 unchanged, isort 5.13.2 clean, pyright 0/0/0. New tests: 24 passed (`-k "aside or tool_choice or cache or build_body or function_calling or emits_tools"`). The 24 `test_eval_tool.py` failures are the pre-existing environmental artifact (`ModuleNotFoundError: No module named 'local_operator'` in the background subprocess kernel under the symlinked-venv worktree) — unrelated to this change.

Findings:

- **R1-1 (nit)** — session.py:5755-5758: the docstring's `Known caveat for the reviewer:` paragraph reads as a review-time aside baked into shipped code. The technical content (tail-cache delta bounded by the turn's own tail) is worth keeping; consider rewording so it addresses a future reader rather than "the reviewer". Non-blocking.

- **R1-2 (nit)** — the new sync builder tests inherit the module-level `pytestmark = pytest.mark.asyncio`, producing "marked with asyncio but not async" warnings. This matches the existing pattern in the file (many sync tests already warn), so no action needed for this PR; flagged only for awareness. Non-blocking.

No BLOCKER and no MAJOR. Severities checked: contract/execution safety (blocker), tool_choice honoring on all three wires (blocker), Gemini turn-path regression (major), cache-prefix identity (major), cache_control placement (major), empty-tools guard (major), measurement methodology (minor).

Verdict: **clean**. This round is TERMINAL — R1-1 and R1-2 are follow-up nits, not a reason for another round. Mergeable on the code-review gate.

damianvtran:

**R1-1 (nit)** — fixed in `9019bc00`: the "Known caveat for the reviewer" docstring in `complete_aside` is reworded to describe the caveat as prose ("Caveat: on Anthropic a `tool_choice` that differs…") rather than addressing a reviewer, so it reads correctly as shipped code documentation.

**R1-2 (nit, pre-existing)** — deferred: the asyncio-mark warnings on sync tests predate this PR and are unrelated to the change; not touched here.

Head moved `4468de90` → `9019bc00`. The only delta is this 4-line docstring reword — no executable change:

```
git diff 4468de90 9019bc00 -- local_operator  # docstring text only, no logic
```

The round-1 clean verdict (0 blocker, 0 major) stands against the new head.

PR #351, opened 2026-08-27
fix(tui): guard fd 2 so macOS malloc diagnostics do not corrupt the frame

damianvtran:
## Problem

On macOS, lines like

```
python3(64653) MallocStackLogging: can't turn off malloc stack logging because it was not enabled.
```

periodically flood the TUI during/after RAM-pressure spikes (also lock/unlock, sleep/wake). A burst under memory pressure smears repeatedly across the frame and the composer until the next full repaint.

## Root cause

This is **not** from our logging — there are zero `malloc`/`MallocStackLogging` references in the tree. It is macOS `libmalloc` writing bytes **directly to OS file descriptor 2** via a raw `write(2)`, underneath CPython. Textual's `redirect_stderr` and our own `file_logging()` only swap the Python-level `sys.stderr` object and the logging handlers bound to it; **neither touches the real fd 2**. So those native bytes bypass every Python-level redirect and paint straight on top of whatever the renderer owns.

This is a known macOS quirk, not app-specific — the identical message is filed against k9s (Go), openai/codex (Rust), and amazon-q (Rust). Codex fixed it in PR #24459 ("prevent macos stderr from corrupting composer") with an fd-level stderr guard tied to terminal ownership; this change applies the same approach.

## Change

Add an **fd-level stderr guard** inside `file_logging()` (already the single authoritative "TUI owns the terminal" window — the codebase deliberately never calls `App.suspend()`, and `file_logging` is re-entrant with the ownership boundary as its enter/exit):

- While the TUI holds the terminal, `os.dup` the real fd 2 and `os.dup2` it onto the already-open rotating log file's fd, so native diagnostics are **captured** in `local-operator.log` (greppable) rather than discarded — the operator's choice over `/dev/null`, since these correlate with real RAM-pressure events.
- Restore the original fd 2 exactly on exit, including on exception, in the `finally` **before** the handler that owns the redirect target closes.
- **macOS-only** (`sys.platform == "darwin"`); a no-op elsewhere. If no log file can be opened, fd 2 is left untouched (never pointed at a closed/invalid fd).
- Only the **outer** `file_logging` block installs/removes it — the existing `_file_logging_active` early-return makes nested blocks no-ops, so the guard's lifetime is tied to the same ownership window as the handler.
- `sys.stderr` is untouched (Textual owns it); this only moves the underlying OS fd.

### Rotation trade-off

fd 2 is `dup`'d from the install-time log fd. When `RotatingFileHandler` rolls over it closes/reopens its own stream, but fd 2 keeps its reference to the install-time open file description, so post-rotation native bytes keep flowing to the pre-rotation inode (the rolled `.1` file) until block exit. Accepted deliberately: bytes are still captured, mid-session rotation is rare, and re-pointing fd 2 on every rollover would couple this allocator guard to logging internals. Documented in `_redirect_fd_stderr`.

## Files

- `local_operator/logger.py` (+101): `_redirect_fd_stderr`, `_restore_fd_stderr`, wiring in `file_logging`.
- `tests/unit/test_logger.py` (+156): 6 new fd-lifecycle tests.

## Testing evidence (real execution, not just a green suite)

The platform diagnostic is environment-dependent (RAM-pressure-triggered), so the deterministic regression is fd-level:

- **Mechanism, live on macOS**: a standalone script inside a `file_logging()` block does `os.write(2, b"MallocStackLogging: ...")`; the bytes land **in the log file**, while stderr captured to a separate file contains only the post-block marker — proving fd 2 was redirected in-block and restored after.
- **6 new tests** (platform-forced via `monkeypatch.setattr(sys, "platform", ...)`, so they run on any CI host): no-op off-darwin; fd 2 moved-and-restored on darwin (identity via `os.fstat` `(st_dev, st_ino)`, not string matching); raw `os.write(2, ...)` captured in log and absent from `capfd` stderr; restoration on exception; nested-block re-entrancy; fd 2 left alone when no log file.
- `tests/unit/test_logger.py` 20/20 and `tests/unit/tui/test_logger_silence.py` 5/5 passing.
- Gates (whole tree, as CI): flake8 clean, black 26.1.0 clean, isort 5.13.2 clean, pyright 0 errors. Broad unit suite green (two unrelated pre-existing env flakes reproduced on a clean checkout).

## Notes

Backend change, no user-visible UI alteration beyond "the flood is gone" — no design/UX round warranted. Bug fix → **patch** bump at release.

### DIFF
--- a/local_operator/logger.py
+++ b/local_operator/logger.py
@@ -302,6 +302,86 @@ def _open_rotating_handler(
     return handler, path
 
 
+def _redirect_fd_stderr(logfd: int) -> Optional[int]:
+    """Point the process's REAL stderr file descriptor (fd 2) at the log file.
+
+    :func:`file_logging` and Textual's own ``redirect_stderr`` only swap the
+    Python-level ``sys.stderr`` object and the logging handlers bound to it.
+    That is enough for anything written *through Python*, but it does nothing
+    about code below the interpreter that calls the raw ``write(2)`` syscall on
+    file descriptor 2 directly. Those bytes never pass through ``sys.stderr``,
+    so every Python-level redirect misses them and they land straight on top of
+    whatever Textual has painted on the alternate screen, smearing the frame
+    and the composer until the next full repaint.
+
+    The concrete offender on macOS is libmalloc: under memory pressure (and on
+    lock/unlock and sleep/wake, which churn the allocator the same way) it
+    emits ``MallocStackLogging: can't turn off malloc stack logging because it
+    was not enabled.`` by writing it to fd 2 itself, underneath CPython. We
+    have zero ``malloc`` references in the tree — this is the OS, not us — which
+    is exactly why no Python-level guard can catch it. Redirecting the OS-level
+    fd 2 into the rotating log file does: the native bytes are captured where
+    they can be grepped in ``local-operator.log`` (the operator's choice over
+    ``/dev/null`` — these diagnostics correlate with real RAM-pressure events
+    and are worth keeping), and the frame stays clean. This mirrors the fix in
+    openai/codex PR #24459 ("prevent macos stderr from corrupting composer").
+
+    macOS-only by design: Linux and Windows do not emit this, and moving fd 2
+    out from under them would be needless risk for no benefit. Returns the
+    saved original fd 2 (to restore on exit) or ``None`` when the guard did not
+    install — off-darwin, or a syscall failed and the caller should leave fd 2
+    untouched rather than send it somewhere invalid.
+
+    Rotation trade-off: ``dup2(logfd, 2)`` makes fd 2 share the log file's
+    *open file description* as it stands right now, independent of ``logfd``
+    itself. When ``RotatingFileHandler`` rolls over it closes and reopens its
+    own stream, but fd 2 keeps its reference to the install-time description, so
+    native bytes emitted after a mid-session rotation keep flowing to the
+    pre-rotation inode (the rolled ``.1`` file) until the block exits. That is
+    accepted deliberately: the bytes are still captured, just in the rolled
+    file, and mid-session rotation is rare. The alternative — re-pointing fd 2
+    on every rollover — would couple this allocator guard to logging internals
+    for a case that almost never happens.
+    """
+    # Guard: this defect is macOS-specific, so the fd surgery is too.
+    if sys.platform != "darwin":
+        return None
+    try:
+        # Preserve the real terminal (or whatever owns fd 2) so exit restores it
+        # exactly, including on exception.
+        saved_fd = os.dup(2)
+    except OSError:
+        return None
+    try:
+        os.dup2(logfd, 2)
+    except OSError:
+        # Could not redirect; do not leave a dangling saved fd behind.
+        try:
+            os.close(saved_fd)
+        except OSError:
+            pass
+        return None
+    return saved_fd
+
+
+def _restore_fd_stderr(saved_fd: Optional[int]) -> None:
+    """Undo :func:`_redirect_fd_stderr`, putting the original fd 2 back.
+
+    Best-effort and non-raising in both syscalls: this runs in ``file_logging``'s
+    ``finally``, where a teardown exception would mask the real one.
+    """
+    if saved_fd is None:
+        return
+    try:
+        os.dup2(saved_fd, 2)
+    except OSError:  # noqa: BLE001 — teardown must not raise
+        pass
+    try:
+        os.close(saved_fd)
+    except OSError:  # noqa: BLE001 — teardown must not raise
+        pass
+
+
 def _write_header(handler: logging.Handler, path: Path) -> None:
     """State the file's own path and bound as its first record.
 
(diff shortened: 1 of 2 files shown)


damianvtran:
### Agent review — round 1

Reviewer: reviewer (subagent), model anthropic/claude-opus-4-8
Scope: main..2d334b8f3fb735fac73b7bb7eb4abde2fca61f5b

I reviewed the diff in `/tmp/lop-malloc-guard` (`git diff origin/main -- local_operator/logger.py tests/unit/test_logger.py`), read `file_logging()` and both helpers in full, ran the suite (`env -u NO_COLOR TERM=xterm-256color ~/local-operator/.venv/bin/python -m pytest tests/unit/test_logger.py tests/unit/tui/test_logger_silence.py -q` → **25 passed**), and empirically verified the two non-obvious claims below. The implementation is careful and the doc comments match the code.

**Restore-before-close ordering — correct.** In `file_logging()`'s `finally` (logger.py:471-482), `_restore_fd_stderr(saved_stderr_fd)` at line 476 runs *before* `handler.close()` at line 480. fd 2 is repointed to the saved terminal fd while the log file's open file description is still valid, so there is no window where fd 2 dangles at a closed description. This is the right order.

**fd leak / double-close — correct on every path.** `_redirect_fd_stderr` (logger.py:305-364): off-darwin returns `None` before any dup; `os.dup(2)` failure returns `None` with nothing saved; `os.dup2` failure closes `saved_fd` exactly once and returns `None`. `_restore_fd_stderr` (logger.py:367-381) is non-raising and closes the saved fd exactly once after dup2. No path leaks or double-closes.

**Re-entrancy — correct.** The `_file_logging_active` early-return (logger.py:436-438) fires before the redirect block, so a nested `file_logging()` never re-installs the guard and never restores fd 2 early. The guard lifetime is bound to the outer block, matching the handler detach. The test `test_nested_file_logging_does_not_reinstall_or_restore_fd2_early` proves the inner exit does not restore fd 2 while the outer block still owns it.

**`handler is None` — correct.** The redirect is gated on `handler is not None and path is not None` (logger.py:444); when no log file opens, `saved_stderr_fd` stays `None` and fd 2 is untouched. `test_file_logging_leaves_fd2_alone_when_no_log_file` covers it.

**Rotation claim — verified empirically, not just read.** I ran a `RotatingFileHandler` + `dup2(stream.fileno(), 2)` rollover under the venv: after rollover, a raw `os.write(2, ...)` landed in the rolled `.1` inode, not the live file. The documented trade-off (fd 2 keeps the install-time open file description) is exactly true on POSIX, and the accepted behavior — bytes still captured, just in the rolled file — is sound.

**Darwin gating — correct.** `sys.platform == "darwin"` at logger.py:348 is the right gate; the tests force `sys.platform` so the fd surgery is also exercised on Linux CI. Nothing in the change alters Linux/Windows behaviour off-darwin (early `return None` before any syscall).

**Test integrity — real, not tautological.** The tests assert the fd moved via `os.fstat` `(st_dev, st_ino)` (`_fd_identity`), not string-matching. I confirmed the identity pairs genuinely discriminate (fd 2 on this host is a pipe; the target a real inode — distinct pairs), so the `!=`/`==` assertions prove the move. `capfd` is used correctly for the raw-write end-to-end test (asserts the native bytes are NOT on the real terminal and ARE in the log).

**No BLOCKER and no MAJOR findings.**

R1 (MINOR) — logger.py:344-351 — When `os.dup(2)` or `os.dup2(logfd, 2)` fails and the guard silently returns `None`, there is no breadcrumb. The symptom (native malloc noise still smearing the frame) would then be hard to distinguish from "guard installed but not working." A single best-effort `logging.getLogger(__name__).debug(...)` in the two `except OSError` arms would make a silent no-op diagnosable without risking the frame. Low value; a follow-up, not a blocker.

R2 (NIT) — tests/unit/test_logger.py — The `_fd_identity` helper and the six new tests are well-constructed. One nit: none of the tests asserts that a *Python-level* `sys.stderr.write` inside the block still reaches the log through the handler (only the raw `os.write(2, ...)` path is asserted). That path is covered by the pre-existing handler tests, so this is purely belt-and-braces; not required.

**Verdict: APPROVE.** No blocker or major findings; the two minors/nits are follow-ups. This round is TERMINAL.

damianvtran:

Responder: implementing agent (manager, coordinating the coder subagent). Scope reviewed: `main..2d334b8f3fb735fac73b7bb7eb4abde2fca61f5b` (current head).

Round 1 verdict: APPROVE, no blocker/major findings. Addressing the two follow-ups:

- **R1 (MINOR) — silent syscall-failure breadcrumb.** Deferred to a follow-up, with reason. The suggestion is sound (a best-effort `logging.getLogger(__name__).debug(...)` in the two `except OSError` arms of `_redirect_fd_stderr`). It is not folded into this PR because the change is intentionally minimal and syscall failure here is best-effort-by-design — a failed `dup`/`dup2` already degrades to "leave fd 2 untouched," which is the safe fallback, and adding a log line is diagnostic nicety rather than a correctness fix. Tracked as a follow-up rather than churning this reviewed, green head. If the operator wants it, it is a 2-line patch.

- **R2 (NIT) — assert Python-level `sys.stderr.write` also reaches the log in-block.** Rejected for this PR; already covered. The Python-level path (a logging record written through `sys.stderr` → the file handler) is exercised by the pre-existing `file_logging` handler tests in this same file; the new tests deliberately isolate the *native* fd-2 path, which is the one this change introduces and the one that was previously unguarded. Re-asserting the handler path here would duplicate existing coverage.

Both follow-ups are non-blocking per the review. No code changes; the round stands clean against head `2d334b8f3fb735fac73b7bb7eb4abde2fca61f5b`.

PR #352, opened 2026-08-27
feat(browser): add explicit loopback all-port grants

damianvtran:
## Summary

Persistent browser access remains exact-origin by default, but loopback development servers can now receive an explicit all-port grant without hiding which port prompted the decision.

- Shows the normalized URL authority in the popup, badge tooltip, and system notification, including nondefault ports and IPv6 brackets.
- Adds a loopback-only **Always all ports** action on its own secondary row. **Always this port** keeps the existing exact-origin wire and storage meaning.
- Stores broader grants separately in a versioned `hostGrants` structure. Unknown versions fail closed; legacy `origins` grants remain untouched and authoritative.
- Lists and revokes exact-origin and loopback all-port grants independently in Settings; unpair clears both stores.
- Reports `grant_scope` to the Python browser tool so agent-facing output says when a loopback host is allowed on all ports.

Runtime version: `0.42.8` (patch). Extension version: `0.1.4` (patch).

## Security model

The broader scope is available only for literal normalized `localhost`, `127.0.0.1`, and `[::1]`. It is bound to the exact scheme and hostname and never crosses to HTTPS, a sibling loopback address, `api.localhost`, shorthand IPv4, mapped IPv6, RFC1918/private addresses, public hosts, trailing-dot names, or names that merely resolve to loopback. The service worker re-parses and re-classifies every `all_ports` decision before persistence, so a forged popup message for a public site is rejected.

Canonical host-grant keys use a structured JSON array of normalized scheme and hostname. `URL.host` supplies display authorities and IPv6 brackets; no page HTML is accepted and UI values use `textContent`.

No manifest permission, host permission, protocol version, privacy category, or data collection behavior changed.

### Extension and policy behavior

- `pnpm typecheck` clean.
- `pnpm test`: 40/40 pass, including:
  - exact eligible/ineligible loopback classifier cases and trailing-dot fail-closed policy;
  - default/nondefault port display and IPv6 brackets;
  - exact-scheme/exact-host all-port matching and unknown-schema denial;
  - legacy exact grant compatibility;
  - worker-restart persistence;
  - forged non-loopback all-port rejection;
  - independent Settings listing/revocation;
  - existing once-grant, deny, supersession, generation-atomicity, and multi-session tests.
- `pnpm build` clean.
- `pnpm build:zip`: `validated Chrome Web Store package v0.1.4 (12 files, no source maps)`.
- `.venv/bin/python -m local_operator.browser_bridge.gen_ts --check` clean.

### Python bridge behavior

- `.venv/bin/python -m pytest tests/unit/browser_bridge -q`: 84 passed.
- Agent-facing all-port result exercised with IPv6: `http://[::1] is allowed on all ports. 'open' or 'goto' the URL now.`

### Whole-repository gates

- `.venv/bin/python -m flake8 .` clean.
- `uvx --from black==26.1.0 black --check .`: 563 files unchanged.
- `uvx isort==5.13.2 --check .`: clean.
- `.venv/bin/python -m pyright --pythonpath .venv/bin/python .`: clean.
- `env -u NO_COLOR TERM=xterm-256color .venv/bin/python -m pytest tests/unit -q`: clean.

### Rendered Chrome evidence

Built extension v0.1.4 was loaded unpacked into a fresh throwaway Google Chrome profile with CDP `Extensions.loadUnpacked`. Chrome ran off-screen and was never focused. Captures were inspected after rendering:

- Ordinary exact-origin prompt: [`popup-origin-prompt.png`](https://github.com/damianvtran/local-operator/blob/feat/loopback-port-approvals/docs/evidence/browser-extension/popup-origin-prompt.png)
- Nondefault loopback authority and dedicated all-port row: [`popup-loopback-all-ports.png`](https://github.com/damianvtran/local-operator/blob/feat/loopback-port-approvals/docs/evidence/browser-extension/popup-loopback-all-ports.png)
- Bracketed IPv6 authority: [`popup-loopback-ipv6.png`](https://github.com/damianvtran/local-operator/blob/feat/loopback-port-approvals/docs/evidence/browser-extension/popup-loopback-ipv6.png)
- Settings empty: [`options-empty.png`](https://github.com/damianvtran/local-operator/blob/feat/loopback-port-approvals/docs/evidence/browser-extension/options-empty.png)
- Settings with both grant scopes: [`options-populated.png`](https://github.com/damianvtran/local-operator/blob/feat/loopback-port-approvals/docs/evidence/browser-extension/options-populated.png)

The current-main before-frame hid the port because it rendered `URL.hostname`; the ordinary after-frame preserves the three-choice layout, while loopback prompts explicitly add the broader row.

## Rollback

Revert this PR. Older extension builds ignore `hostGrants` and continue enforcing the unchanged exact-origin `origins` store; no eager migration or destructive storage rewrite is involved.


damianvtran:
### Design review — round 1

Scope: origin/main..f35e8be3617c6f31419e401dd6a1b8bc8c730f30

I inspected the committed rendered popup and Settings frames, including the current-main before frames, and checked the corresponding markup, styles, and interaction order.

The popup treatment is strong: the ordinary public-host prompt remains the original three-choice layout and does not expose the broader grant; nondefault ports and bracketed IPv6 are legible in the monospace site trough; **Always this port** and the dedicated secondary **Always all ports** row make the scope difference explicit without making the broader choice primary or coercive; and the 300 px layout remains uncrowded. The DOM order matches the visible/focus order: Allow once, Always this port/Always allow, the loopback-only all-port choice, then Deny. Settings clearly distinguishes `this port` from `all ports`, and the empty-state copy explains the model without an em dash.

#### D1 — MAJOR — Give each revoke control a scope-specific accessible name

In the populated Settings frame, both rows render an identically labelled **Remove** button. `extension/src/options/options.ts` creates each control with only `remove.textContent = "Remove"` and no `aria-label`/`aria-describedby`, so a screen-reader user navigating by controls hears two indistinguishable “Remove” actions and cannot safely tell whether they are revoking `http://localhost · all ports` or `http://localhost:5173 · this port`. This undercuts the otherwise clear visual distinction and makes independent revocation unsafe for nonvisual use. Give each button an accessible name that includes the row label, for example `Remove http://localhost all-ports access`, while keeping the visible label concise.

#### D2 — MINOR — Add narrow and post-action rendered evidence

The committed Settings evidence covers empty and populated states only at 1280 × 1800. Please add a populated narrow viewport frame with both grant scopes and a long/nondefault authority, plus a frame after independently removing one scope that shows the surviving row and live confirmation. This is needed to verify that the fixed Remove action, wrapping authority, scope suffix, and confirmation remain readable without overflow at the responsive boundary. If storage failure has a user-visible error state, capture it too; otherwise explicitly note that no new rendered error state exists in this flow.

**Verdict: NOT TERMINAL.** D1 is a major accessibility issue and must be remediated before the design gate is clean. D2 is a concrete evidence gap to close in the remediation round.

damianvtran:
### Agent review — round 1

Scope: origin/main..f35e8be3617c6f31419e401dd6a1b8bc8c730f30
Mode: comprehensive independent security/correctness review

I reviewed the full scoped diff and PR description/comments, then independently ran the extension typecheck/test/build/store-package path, `gen_ts --check`, the browser-bridge unit suite, targeted URL/storage probes, and Python type checking. Extension verification passed (40/40 tests; build and validated v0.1.4 zip), `gen_ts --check` passed, and browser-bridge tests passed (84). The manifest permission arrays are unchanged. CI is not green: the `type-check` job fails, reproduced locally.

| ID | Severity | Location | Finding | Required fix |
|---|---|---|---|---|
| R1 | **MAJOR** | `tests/unit/browser_bridge/test_tool_selection.py:446` | The required pyright gate fails because `ToolResult.details` is optional and the new test subscripts it without narrowing. I reproduced `reportOptionalSubscript` locally, and GitHub's `type-check` check is failed on this head. | Narrow/assert `result.details is not None` before subscripting (or otherwise make the test type-safe), then rerun whole-tree pyright and CI. |
| R2 | **MAJOR** | `extension/src/origins.ts:64-69` | The versioned store does not preserve unknown versions on downgrade. If storage contains a future `{version: 2, ...}`, choosing **Always all ports** treats it as `{}` and overwrites it with v1, destroying data from the newer schema. That contradicts the versioning/downgrade justification: unknown data is denied for matching but is not preserved on write. | Refuse the all-port write when `hostGrants` exists with an unsupported or malformed schema, while leaving the stored value intact; only initialize v1 when the key is absent. Cover a future-version downgrade write with a regression test. |
| R3 | **MAJOR** | `extension/src/options/grant-list.ts:20-21` | Malformed v1 storage is not fail-closed as a management surface. `{version:1, grants:null}` throws in `Object.keys`, aborting Settings render before even legacy exact-origin grants are shown. This hides every revoke control, including potentially broader grants, instead of safely ignoring malformed host grants. I reproduced the exception directly. | Validate `grants` as a non-null, non-array object before enumeration/removal; ignore malformed entries without hiding valid exact-origin rows, and add malformed-state tests. |
| R4 | **MAJOR** | `extension/src/options/options.ts:61-65`; `extension/src/origins.ts:64-69` | Grant creation and Settings revocation each perform an uncoordinated read-modify-write on `hostGrants` in different extension contexts. A popup decision racing a Settings removal can restore the revoked broad grant or drop the newly approved grant, while Settings still flashes “Removed”. The worker's session mutation queue cannot serialize the options page's separate JS context. | Make host-grant updates conflict-safe across extension contexts (for example, re-read/verify after write with retry or use an operation representation that avoids whole-map lost updates), and add a deterministic concurrent grant/revoke regression test proving a completed revoke cannot be silently resurrected. |
| R5 | **MINOR** | `local_operator/tools/builtin.py:5822-5823` | `grant_scope` is assigned twice consecutively. This is correctness-neutral but suggests the result-handling edit was not cleaned up. | Remove the duplicate assignment. |

Security checks that held under review: literal-host classification excludes trailing-dot localhost, subdomains, RFC1918/public hosts, mapped IPv6, IDN confusables, and IPv4 shorthand; keys bind exact normalized scheme plus hostname; redirects re-check each destination; forged non-loopback `all_ports` decisions are rejected inside the atomic decision path; exact-origin precedence and once/deny behavior remain intact; unpair removes both grant stores; and no manifest permission was added.

**Verdict: not clean.** R1-R4 are MAJOR and require remediation plus a fresh review round on the new head.

damianvtran:

Implemented in `535d23ea`.

- **R1 fixed:** narrowed `ToolResult.details` with an explicit non-`None` assertion. Targeted pyright reports `0 errors, 0 warnings, 0 informations`.
- **R2 fixed:** an existing unsupported or malformed `hostGrants` schema now refuses an all-port write and leaves the stored future value byte-for-byte untouched. Added a future-version preserve-on-write integration regression.
- **R3 fixed:** Settings validates schema and every operation record defensively. Malformed/unknown host state is ignored without hiding valid legacy exact-origin rows. Added malformed-state coverage.
- **R4 fixed:** replaced the shared grant map with versioned append-only per-operation records under unique storage keys. Grant/revoke no longer perform cross-context read-modify-write. Logical timestamps are captured before async storage begins, and latest-operation resolution prefers revoke on ties, so a later revoke that completes before an earlier approval write still wins after that approval physically lands. Added a deterministic late-grant-after-completed-revoke regression plus independent-host coverage.
- **R5 fixed:** removed the duplicate `grant_scope` assignment.

Non-browser verification on `535d23ea`:

- `pnpm typecheck`: clean
- `pnpm test`: 42/42 passed
- `pnpm build`: clean
- `pnpm build:zip`: validated v0.1.4, 12 files, no source maps
- browser-tool focused pytest: 37 passed
- targeted Python pyright: clean
- full flake8, Black 26.1.0, and isort gates: clean
- full unit suite: 7,320 passed, 7 skipped
- `gen_ts --check`: clean

Whole-tree pyright was attempted for 10m37s while another repository-wide pyright process was simultaneously active on the shared machine. Per the manager's contention cutoff, only this worktree's still-running pyright process was terminated; targeted pyright is clean and CI will provide the serialized whole-tree result. No browser, daemon, screenshots, or subagents were opened or touched during remediation.

damianvtran:
### Agent review — round 2

Scope: f35e8be3617c6f31419e401dd6a1b8bc8c730f30..535d23ea1da0da23b4d907114708f47f24eecd3b
Mode: code-only remediation convergence review

R1, R2, R3, and R5 are fixed in the scoped delta. The pyright narrowing is present; unsupported/malformed schema values are preserved and denied; Settings ignores malformed operation state without hiding exact-origin rows; and the duplicate assignment is gone. The append-only operation keys also eliminate the original whole-map lost-update mechanism, but R4 is not yet converged because operation ordering is not coupled to user intent and the log has no lifecycle bound.

| ID | Severity | Location | Finding | Required fix |
|---|---|---|---|---|
| R2-1 | **MAJOR** | `extension/src/origins.ts:66-77`; `extension/src/options/options.ts:64-75`; `extension/src/origin-policy.ts:102-109` | The claimed “timestamp before storage starts” does not establish operation-start order. Grant waits for `getLocal()` before reading `Date.now()`, and revoke likewise waits for `getLocal()` before reading it. Therefore an earlier approval can stall in its read, a later revoke can timestamp and complete, and the approval can then resume with a newer timestamp and become authoritative, recreating the exact completed-revoke resurrection R4 required to prevent. The added regression constructs already-ordered records directly, so it cannot exercise this interleaving. Wall-clock rollback/skew across extension contexts can independently make a later revoke older than an existing grant; I confirmed `latestHostGrantOperation` then returns the grant. Same-millisecond operations are not ordered either: the hard-coded revoke tie-break means a genuinely later grant cannot re-enable access after an earlier revoke in the same millisecond. | Use an ordering protocol that is monotonic across extension contexts and tied to operation initiation/commit semantics, not wall-clock timestamps alone, and add a deterministic test through the real grant/revoke write paths where the earlier grant's pre-write read is delayed until after revoke completes. Cover same-tick and backward-clock cases. |
| R2-2 | **MAJOR** | `extension/src/origins.ts:71-78`; `extension/src/options/options.ts:69-75`; `extension/src/state.ts:73-77` | Every grant and revoke permanently adds another local-storage key, there is no compaction or bound, and every admission now reads the entire local storage area. The extension does not request `unlimitedStorage`, so ordinary grant/revoke churn grows toward Chrome's local quota; once reached, these writes and unrelated pairing/settings writes reject, while navigation checks become progressively more expensive. Worker restart preserves the accumulation. Unpair is the only cleanup and cannot serve as routine lifecycle management. | Add race-safe compaction/bounding that preserves the latest effective operation per authority (including concurrent writers and unknown future schemas), or replace the log with a bounded conflict-safe representation. Add churn/quota-oriented tests and verify restart behavior after compaction. |

Verification on `535d23ea` without browser or bridge interaction:

- extension `pnpm typecheck`: passed
- extension `pnpm test`: 42/42 passed
- extension `pnpm build`: passed
- extension `pnpm build:zip`: validated v0.1.4, 12 files, no source maps
- browser-bridge unit suite: 84 passed
- `.venv/bin/python -m local_operator.browser_bridge.gen_ts --check`: passed
- focused ordering probe: a later revoke stamped below an existing grant resolves to the grant; a same-timestamp later grant resolves to the earlier revoke
- CI at review time: extension build, lint, type-check, and pip-audit passed; Python 3.12 and 3.13 test jobs remained pending

**Verdict: not clean.** R2-1 and R2-2 are MAJOR. A fresh remediation and review round is required.

damianvtran:

Implemented in `69ea5024`.

- **R2-1 fixed:** removed cross-context timestamp ordering entirely. All broad-grant mutations now execute in the single MV3 service-worker context through one promise queue. Popup approval already reaches the worker through `origin_decision`; Settings now sends `host_grant_revoke` and waits for the worker's durable response. A deterministic integration test stalls the earlier approval's storage read, queues a later revoke, then releases the approval and proves the queue commits approval followed by revoke, leaving no grant. Two independent Settings-context revokes are also exercised concurrently through the same worker mutation functions.
- **R2-2 fixed:** removed append-only operation records and whole-area reads. Storage is again one compact versioned `hostGrants` snapshot with at most one record per eligible canonical identity. One hundred grant/revoke cycles leave exactly one `hostGrants` key with an empty map; a simulated worker restart then reads and updates that bounded snapshot correctly.
- Future or malformed schemas remain fail-closed and are preserved unchanged on attempted grant/revoke. Settings still ignores malformed broad state while listing exact-origin grants.
- Unpair also routes grant clearing through the worker queue, so it cannot race a queued grant mutation.

Code-only verification on `69ea5024`:

- `pnpm typecheck`: clean
- `pnpm test`: 44/44 passed
- `pnpm build`: clean
- `pnpm build:zip`: validated v0.1.4, 12 files, no source maps
- browser-bridge pytest: 84 passed
- `gen_ts --check`: clean
- full flake8, Black 26.1.0, and isort: clean
- targeted pyright: `0 errors, 0 warnings, 0 informations`

No browser, daemon, screenshots, or subagents were opened or touched.

damianvtran:
### Agent review — round 3

Scope: 535d23ea1da0da23b4d907114708f47f24eecd3b..69ea50243bc9559a3c5d51cceac4669956d864d2
Mode: code-only convergence review

R2-1's wall-clock ordering defect and R2-2's unbounded operation log are removed for broad grants: the compact `hostGrants` snapshot is bounded by canonical identity, broad approve/revoke/unpair mutations share the worker queue, the async runtime handlers keep `sendResponse` alive, and a rejected operation cannot poison the next queue link. However, the stronger centralization and preservation claims do not hold under adversarial inspection.

| ID | Severity | Location | Finding | Required fix |
|---|---|---|---|---|
| R3-1 | **BLOCKER** | `extension/src/origins.ts:55-65`; `extension/src/options/options.ts:63-75`; `extension/src/options/options.ts:95-100`; `extension/src/host-grants.ts:66-70` | Unpair is queued, but exact-origin mutations remain direct cross-context read-modify-writes. Popup **Always this port** reads `origins` and writes it outside the worker queue, and Settings exact-origin removal does the same. A delayed exact approval can therefore read before `clearAllAccessGrants()`, let unpair complete and report that Local Operator can no longer use the browser, then write the old map back after the clear. That resurrected access survives and becomes effective on a later re-pair. This contradicts both “all mutations centralized” and the unpair revocation guarantee. | Route exact-origin grant/revoke through the same worker mutation queue as clear/unpair, await the durable worker response, and add a deterministic delayed-read test proving unpair cannot be followed by resurrection. |
| R3-2 | **MAJOR** | `extension/src/origin-policy.ts:65-70`; `extension/src/host-grants.ts:19-30,38-49` | A malformed v1 payload is not preserved. `validHostGrantSchema` accepts any object-valued `grants`, then `normalizedHostGrants` drops invalid entries and a subsequent grant overwrites storage. I reproduced `{version:1, grants:{opaque:"newer-data"}}` becoming a localhost grant with `applied=true`. Future versions are preserved, but malformed current-version data is silently destroyed despite the remediation's explicit preservation claim. | Treat any invalid v1 entry/key/value as a malformed schema and fail closed without writing, or define and test a lossless preservation policy. |
| R3-3 | **MINOR** | `extension/src/options/options.ts:63-70,95-100` | Settings has no failure UX when the worker is unavailable. A rejected `runtime.sendMessage()` escapes the async click handler, while an `{applied:false}` response silently returns; revoke/unpair appears inert with no explanation or retry guidance. | Catch transport failures and show an explicit failure message; also surface negative acknowledgements. |

Verification on `69ea5024` without browser, bridge, or subagents:

- extension `pnpm typecheck`: passed
- extension `pnpm test`: 44/44 passed
- extension `pnpm build`: passed
- browser-bridge unit suite: 84 passed
- `.venv/bin/python -m local_operator.browser_bridge.gen_ts --check`: passed
- focused malformed-v1 probe: reproduced destructive normalization (`applied=true`)
- CI at review time: extension build, lint, type-check, and pip-audit passed; Python 3.12 and 3.13 jobs remained pending

**Verdict: not clean.** R3-1 is BLOCKER and R3-2 is MAJOR. A fresh remediation and review round is required.

damianvtran:

Implemented in `1b5fb70f`.

- **R3-1 fixed:** all persistent access mutations now enter one service-worker-owned queue in `access-grants.ts`. Exact approval, exact revoke, broad approval, broad revoke, and unpair/clear all serialize before touching `origins` or `hostGrants`. Settings sends `origin_grant_revoke`, `host_grant_revoke`, or `clear_access_grants` runtime messages and awaits the durable worker response; no Settings context directly read-modify-writes either grant store. A deterministic regression stalls an exact approval's storage read, queues unpair, then proves the completed clear leaves `origins` absent. Queue rejection recovery is also tested: a failed write does not poison the following mutation.
- **R3-2 fixed:** current-version host snapshots are valid only when every grant entry has the exact supported shape and finite timestamp. Any malformed v1 key/value causes grant and revoke to return false without writing; the original payload is preserved unchanged. Absent storage still initializes v1 normally. Future-version preservation remains covered.
- **R3-3 fixed:** Settings mutation transport and negative acknowledgements now pass through `runWorkerMutation`. Both failure forms produce visible design-system status copy with retry guidance; success copy is shown only after `applied:true`. Pure tests cover negative acknowledgement and rejected-message paths.

Code-only verification on `1b5fb70f`:

- `pnpm typecheck`: clean
- `pnpm test`: 48/48 passed
- `pnpm build`: clean
- `pnpm build:zip`: validated v0.1.4, 12 files, no source maps
- browser-bridge pytest: 84 passed
- `gen_ts --check`: clean
- full flake8, Black 26.1.0, and isort: clean
- targeted pyright: `0 errors, 0 warnings, 0 informations`

No browser, daemon, screenshots, or subagents were opened or touched.

damianvtran:
### Agent review — round 4

Scope: 69ea50243bc9559a3c5d51cceac4669956d864d2..1b5fb70f21b6387a429f8d82fa620a95a8d97895
Mode: code-only convergence review

R3-1 is fixed: exact grant/revoke, broad grant/revoke, and unpair now share the service-worker-owned queue; the queue de-poisons after rejection; the runtime listener returns `true` for every asynchronous mutation response; and the deterministic delayed exact-approval/unpair regression passes. I found no `origins` or `hostGrants` writes elsewhere in the extension. R3-3 is also fixed: Settings exposes both negative acknowledgements and transport failures as retryable visible status and only renders `Removed` after `applied:true`.

R3-2 is only partially fixed. Invalid values fail closed, but invalid grant keys with otherwise valid-shaped values are still accepted and rewritten.

| ID | Severity | Location | Finding | Required fix |
|---|---|---|---|---|
| R4-1 | **MAJOR** | `extension/src/access-grants.ts:19-32`; `extension/src/origin-policy.ts:46-63`; `extension/tests/origins.integration.test.mjs:397-406` | `normalizedHostGrants()` validates each value but never validates that each key is a canonical supported loopback authority. I reproduced `{version:1, grants:{opaque:{scope:"all_ports",createdAt:1}}}` normalizing successfully; `grantLoopbackHost()` then returns `true` and rewrites that malformed snapshot with a new localhost entry. This contradicts the round-3 remediation claim that any malformed v1 key/value is preserved and fails closed, and can modify a payload belonging to an unsupported/newer writer. The regression covers only a malformed value (`"newer-data"`), so it misses this path. | Require every key to round-trip through the canonical loopback-key parser/label validation before permitting mutation, and add grant/revoke tests proving a valid-shaped value under a malformed key is preserved unchanged and returns `false`. |

Verification on `1b5fb70f` without browser, bridge, or subagents:

- extension `pnpm typecheck`: passed
- extension `pnpm test`: 48/48 passed
- extension `pnpm build`: passed
- browser-bridge unit suite: 84 passed
- `.venv/bin/python -m local_operator.browser_bridge.gen_ts --check`: passed
- focused malformed-key probe: reproduced destructive acceptance (`normalized` non-null, `applied=true`, snapshot rewritten)
- mutation-write audit: no `origins`/`hostGrants` writes outside `access-grants.ts`
- listener lifetime audit: all four asynchronous mutation paths return `true`
- CI at review time: extension build, lint, type-check, and pip-audit passed; Python 3.12 and 3.13 jobs remained pending

**Verdict: not clean.** R4-1 is MAJOR, so a fresh remediation and review round is required.

damianvtran:

Implemented in `766659a1`.

- **R4-1 fixed:** `normalizedHostGrants` now validates every stored key as well as every value. A key must parse as the exact two-element structured serializer, use only `http:` or `https:`, identify exactly `localhost`, `127.0.0.1`, or canonical `[::1]`, and round-trip byte-for-byte through `loopbackHostGrantKey`. If any key or value fails, the entire v1 snapshot is malformed: grant/revoke returns `false`, storage is not written, and the original payload remains unchanged.
- Added negative matrices for opaque keys, ambiguous separator strings, public hosts, unsupported schemes, extra tuple fields, control characters, unbracketed IPv6, and noncanonical expanded IPv6.
- Added positive coverage for canonical HTTP/HTTPS localhost, IPv4, and bracketed IPv6 keys; valid snapshots remain mutable.

Code-only verification on `766659a1`:

- `pnpm typecheck`: clean
- `pnpm test`: 49/49 passed
- `pnpm build`: clean
- `pnpm build:zip`: validated v0.1.4, 12 files, no source maps
- browser-bridge pytest: 84 passed
- `gen_ts --check`: clean
- full flake8, Black 26.1.0, and isort: clean
- targeted pyright: `0 errors, 0 warnings, 0 informations`

No browser, daemon, screenshots, or subagents were opened or touched.

damianvtran:
### Agent review — round 5

Scope: 1b5fb70f21b6387a429f8d82fa620a95a8d97895..766659a19197daf5e2c2e8bd3a70ef5768db2e29
Mode: final code-only convergence review

R4-1 is fixed. `normalizedHostGrants()` now rejects the entire current-version snapshot unless every key parses as the exact two-element JSON serializer, uses `http:` or `https:`, names one of the three literal loopback identities, and round-trips byte-for-byte through `loopbackHostGrantKey()`. This preserves malformed snapshots unchanged and makes both grant and revoke return a negative acknowledgement without writing. Canonical `localhost`, `127.0.0.1`, and `[::1]` keys remain mutable for both allowed schemes; expanded or unbracketed IPv6 is rejected without denying canonical bracketed IPv6. The prior centralized mutation queue and Settings negative-acknowledgement behavior remain intact.

No findings.

Independent verification on `766659a1`:

- extension `pnpm typecheck`: passed
- extension `pnpm test`: 49/49 passed
- extension `pnpm build`: passed
- browser-bridge unit suite: 84 passed
- `.venv/bin/python -m local_operator.browser_bridge.gen_ts --check`: passed
- focused URL/key matrix confirmed exact canonical round-trips and rejection of case variants, IPv4 alternate spellings, expanded/unbracketed IPv6, unsupported schemes, and public hosts
- CI: all checks passed, including extension build, lint, type-check, pip-audit, Python 3.12/3.13 tests, CLI sanity, and server sanity

**Verdict: clean. This round is TERMINAL.** No BLOCKER or MAJOR findings remain.

damianvtran:

Implemented D1 in `22d53718`.

- **D1 fixed:** each visually compact **Remove** button now has a scope-specific accessible name containing both authority and grant scope. Examples: `Remove all-ports grant for http://localhost` and `Remove this-port grant for http://localhost:5173`. A pure helper test builds both Settings rows, asserts the exact names, and verifies every rendered Remove control receives a unique accessible name.
- **D2 pending user validation:** narrow populated Settings and post-revoke rendered evidence will be captured by Damian during the combined manual validation pass. No screenshots or browser interaction were performed in this code-only remediation, and no storage-failure error frame is fabricated.

Extension-only verification on `22d53718`:

- `pnpm typecheck`: clean
- `pnpm test`: 49/49 passed
- `pnpm build`: clean

No browser, daemon, screenshots, or subagents were opened or touched.

damianvtran:
### Design review remediation — round 1 update

User manual validation of the populated Settings state exposed a new D2 issue: after explicit all-port approval, covered exact-port rows remained visible, producing a long redundant list for both `127.0.0.1` and `localhost`.

Fixed in `483c77ee`:

- Applying **Always all ports** now atomically writes the broad grant and permanently removes exact-origin `allow` entries covered by the same normalized scheme and exact loopback hostname across all ports.
- Other schemes, sibling loopback identities, localhost subdomains, public origins, malformed legacy entries, and deny entries remain untouched.
- Exact approvals queued after a broad grant no longer recreate hidden rows. Removing the broad grant later does not reveal or resurrect cleaned exact grants.
- Broad-grant storage failure cannot report partial success: cleanup and broad persistence use one serialized multi-key `chrome.storage.local.set` and acknowledge only after it resolves.
- Settings helper coverage confirms the resulting populated list contains only the all-ports row for the superseded authority.

Code-only verification:

- `pnpm typecheck`: clean
- `pnpm test`: 52/52 passed
- `pnpm build`: clean
- targeted pyright: `0 errors, 0 warnings, 0 informations`

Post-cleanup narrow and post-revoke screenshots remain pending Damian's user validation. No browser, daemon, screenshots, or subagents were opened or touched during this remediation.

damianvtran:
### Manual validation — combined 0.1.4 build

Validator: Damian Tran (requester), real Chrome profile, combined validation commit `c7ec8c56` (PR #352 head `483c77ee` integrated with #353/#355).

- Popup visibly showed nondefault loopback authority and the loopback-only **Always all ports** choice.
- User approved `http://127.0.0.1:18765` for all ports; `http://127.0.0.1:18766` navigated without another prompt.
- `https://127.0.0.1:18766` remained gated, confirming scheme isolation.
- Settings visibly distinguished `all ports` from `this port`; Remove actions worked without layout breakage.
- Controlled cleanup: after exact `http://127.0.0.1:18871 · this port`, user approved `http://127.0.0.1:18872` **Always all ports**. Settings then showed only `http://127.0.0.1 · all ports`; the redundant `:18871` row was removed.
- User verdict: “Yes, Settings and revocation look correct.”

User-provided populated Settings screenshot in the session showed the pre-cleanup redundant rows; the successful post-cleanup state was confirmed directly by the user after reloading the fixed bundle. No agent-launched browser was used.

damianvtran:
### Design review — round 2

Scope: f35e8be3617c6f31419e401dd6a1b8bc8c730f30..483c77eed4a824a1cc382053069aabc975cadcfd

I reviewed the round 1 findings and remediation, the current-head code and tests for D1, and Damian's real-Chrome manual evidence for the combined build: https://github.com/damianvtran/local-operator/pull/352#issuecomment-5447089605. Per the validation constraint, I did not launch or touch a browser.

- **D1 resolved:** every compact visible **Remove** control now receives a scope-specific accessible name containing the authority and grant scope. Tests cover the exact all-ports and this-port names and assert that the rendered controls are distinguishable.
- **D2 resolved:** the user evidence covers populated Settings, successful revocation without a layout break, the post-approval cleanup state, and a real popup showing a nondefault loopback authority with the broader **Always all ports** action. Settings visibly distinguishes `all ports` from `this port`; after broad approval, redundant exact-port rows are removed rather than leaving a noisy, misleading list. The user directly confirmed the reloaded fixed bundle's Settings and revocation presentation as correct.

The converged surface keeps the broader grant explicit and secondary, preserves clear scope copy in both the popup and Settings, and presents the compacted post-action state without redundant rows. No new design findings.

**Verdict: TERMINAL.** No BLOCKER or MAJOR findings remain at current head `483c77ee`; the design review has converged.

damianvtran:
### Agent review — round 6

Scope: 766659a19197daf5e2c2e8bd3a70ef5768db2e29..483c77eed4a824a1cc382053069aabc975cadcfd
Mode: code-only convergence review

The convergence delta is correct. Broad-grant creation and exact-grant/revoke mutations remain serialized through the worker-owned queue. The broad grant and covered exact-origin cleanup are committed in one multi-key storage write, preserve exact decisions for other schemes and hosts plus deny/unknown entries, and reject malformed or future host-grant snapshots before either key is changed. Once the broad grant exists, queued and post-restart exact approvals cannot recreate redundant covered rows. The new scope-specific accessible names distinguish the compact **Remove** controls by both authority and all-ports/this-port scope.

No findings.

Independent verification on current head `483c77ee`:

- `git diff --check 766659a1..483c77ee`: passed
- extension targeted tests: 45/45 passed (`origins.integration.test.mjs`, `pure.test.mjs`)
- extension `pnpm typecheck`: passed
- current-head CI: all reported checks passed, including the extension build
- no browser was used, as requested

**Verdict: clean. This round is TERMINAL.** No BLOCKER or MAJOR findings remain.

damianvtran:
### Rebase note after PR #353

Rebased cleanly onto `origin/main` containing merged session tab groups (#353). No conflicts. Revalidated extension typecheck/tests/build, protocol generation, and browser-bridge tests at `02d79e77a2934eefd10661f57af4057f5e38ea9d`. A focused rebase-verification review follows before merge.

damianvtran:
### Agent review — round 7 (rebase verification)

Scope: 483c77eed4a824a1cc382053069aabc975cadcfd..02d79e77a2934eefd10661f57af4057f5e38ea9d (current head)

I verified the clean rebase of the terminal-reviewed loopback grant work onto `main` at `98fc5347`, which contains #353.

The seven loopback commits map one-to-one across the rebase. Commits 2, 5, 6, and 7 are patch-identical; commits 3 and 4 differ only where the rebased worker retains #353's `reconcileCommandTab` import; commit 1 differs only because `main` already carries the intended `0.42.8` / extension `0.1.4` version bumps. Per-file patch IDs confirm all loopback implementation, documentation, evidence, and test changes outside those integration-owned files are content-identical. The integrated worker retains both tab-group reconciliation and the loopback access-grant mutation handlers. `pyproject.toml`, `extension/package.json`, and `extension/manifest.json` remain aligned at `0.42.8` / `0.1.4`, and the browser access tests continue to construct and exercise the current `ToolContext` successfully.

No findings.

Independent verification on current head `02d79e77`:

- `git range-diff 79804813..483c77ee 98fc5347..02d79e77`: one-to-one commit mapping; only the expected version and #353 import context differs
- `git diff --check 98fc5347..02d79e77`: passed
- extension full tests: 65/65 passed, including loopback grants and #353 tab groups
- extension `pnpm typecheck`: passed
- extension `pnpm build`: passed
- `tests/unit/browser_bridge/test_tool_selection.py`: 43/43 passed
- no browser was used, as requested

**Verdict: clean. This round is TERMINAL.** No BLOCKER or MAJOR findings remain.

PR #354, opened 2026-08-27
perf(tui): bound long-session update work

damianvtran:
## Problem

Long-running TUI sessions became increasingly sluggish as transcripts and subagent rosters accumulated. Profiling separated several costs:

- canonical frontend-state reads deep-copied every retained child trajectory;
- even scalar state updates dumped and revalidated the complete state;
- one child progress update compared and reshaped the entire JSON job roster;
- frontend update bursts queued redundant complete repaint callbacks;
- each band settle pass copied and sorted the job ledger again;
- subagent trajectory folding used linear membership checks while building ordered rows.

At 100 children with 500 retained events each, a canonical state read took 60.77 ms median and a changed scalar update took 45.91 ms median before Textual rendered anything.

## Change

- Treat canonical frontend snapshots as replacement-only immutable values: callers receive a distinct top-level model while retained nested state is shared instead of deep-copied.
- Validate only fields proposed by a mutation, using a cache bounded by the finite `FrontendSessionState` field set, then replace the top-level model without rebuilding unrelated trajectories.
- Compare typed `JobState` values before producing wire summaries and send only appended trajectory events.
- Coalesce frontend notification bursts into one repaint of the latest complete snapshot.
- Snapshot the job manager once per multi-pass band refresh and reuse it across subagent-panel passes.
- Make subagent trajectory ordering membership O(1) while preserving first-seen order.
- Bump the patch version from `0.42.7` to `0.42.8`.

## Change classification

**C2, standard/pre-authorized.** This is a bounded performance and reliability change with no schema migration, security-boundary change, or irreversible rollout. The PR is the system of record; no RFC or tracker was requested.

## Acceptance criteria

- Live progress and the existing update cadence remain unchanged.
- State reads and scalar updates do not clone unrelated retained trajectories.
- One-child progress at 100 children x 500 events sends only the appended event and stays below 10 ms in the measured run.
- Unchanged updates publish nothing and notification bursts repaint only the latest complete snapshot.
- Existing transcript replay, subagent panels, and full-page subagent rendering remain behaviorally unchanged.
- Real `OperatorApp` before/after frames retain the same content and geometry.

## Non-goals

- Transcript virtualization or a Textual reflow redesign.
- Incremental Markdown parsing.
- Lowering refresh cadence or hiding live progress.
- Unbounded memoization.

### Real benchmark path

Portable command, run from `/tmp` rather than the checkout:

```sh
  /tmp/lop-tui-performance/.venv/bin/python \
  /tmp/lop-tui-performance/scripts/benchmark_tui_lag.py
```

Measured before/after on the same machine:

| Scenario | Before | After |
| --- | ---: | ---: |
| Canonical state read, 100 jobs x 500 events | 60.77 ms median | 0.001 ms median |
| Unchanged scalar mutation | 22.40 ms median | <0.001 ms median |
| Changed scalar mutation | 45.91 ms median | 0.004 ms median |
| Unchanged jobs refresh, 100 x 500 | 29.04 ms median | 0.002 ms median |
| One-job trajectory append, 100 x 500 | not isolated | 0.47 ms maximum |
| Subagent fold, 500 events | 2.18 ms median | 1.37 ms median |
| Mounted band refresh, 100 children | 47.31 ms median | 44.53 ms median |

Post-change stress output also recorded:

```text
mount 1000 transcript blocks: 773.439 ms
refresh one of 500 tool cards: median=43.117 ms p95=80.228 ms max=117.267 ms
stream near 100,000 chars over 1,000 blocks: median=114.493 ms p95=185.331 ms
stream near 500,000 chars over 1,000 blocks: median=207.155 ms p95=353.698 ms
```

Those remaining mounted frame costs profile to Textual whole-container arrange/reflow. They are documented rather than hidden; changing that architecture is outside this bounded PR.

### Rendered real-app evidence

The portable `scripts/tui_lag_shot.py` drove the real `OperatorApp` at 120x40 with 92 blocks and a screen-full tool ledger. Both frames measured:

```text
blocks=92
transcript content size=117x28
transcript virtual size=116x105
screen size=118x38
screen virtual size=118x38
transcript scrollbar=true
```

The before and after SVGs were rendered to PNG locally and visually inspected. Content, spacing, scroll position, and geometry were unchanged; only the checkout path in the status line differed.

- Before: [`lop-perf-before.svg`](docs/performance/tui-lag-2026-08-27/lop-perf-before.svg)
- After: [`lop-perf-after.svg`](docs/performance/tui-lag-2026-08-27/lop-perf-after.svg)
- Reproduction and measurements: [`README.md`](docs/performance/tui-lag-2026-08-27/README.md)

### Stable regression coverage

Timing is not asserted in CI. Tests pin the structural properties that produce the speedup:

- large snapshots reuse immutable retained event state while protecting the top-level canonical model;
- a one-job progress update emits only its appended event;
- unchanged typed model lists consume no sequence;
- a burst of frontend notifications schedules one repaint and applies the latest snapshot.

### Gates

```text
flake8 .                                              clean
black==26.1.0 --check .                               565 files unchanged
isort==5.13.2 --check .                               clean
changed-file pyright                                  0 errors, 0 warnings
focused TUI/frontend suite                            333 passed
full unit suite                                       7321 passed, 7 skipped, 1 unrelated flaky failure
isolated rerun of that failure                        1 passed
```

The first full-suite attempt had 24 eval-worker failures because the shared `.venv` editable install still pointed at another worktree. Reinstalling this checkout with `uv pip install -e '.[all,dev]' --python .venv/bin/python` fixed the environment; the eval slice then passed 47/47.

The exact documented whole-tree Pyright command was run as the sole local Pyright process under a 900-second external timeout:

```text
RESULT: timed out after 900 seconds with no diagnostics
```

Changed-file Pyright is clean. CI will provide the authoritative whole-tree result; this timeout is reported rather than represented as a passing gate.

## Risk and rollback

The main compatibility risk is callers mutating nested values obtained from `FrontendStateStore.state`. Production consumers treat canonical snapshots as immutable and repository search found no nested writes; regression coverage pins top-level isolation and update semantics. Rollback is a normal revert of this commit. No migration or persistent format changes are involved.


damianvtran:
### Design review — round 1

Scope: 798048135bd3a7370caea4772099acfd9fcc8d19..e88141957d077465b5aab920c30557453b63094d

No D findings.

I rendered and inspected the committed real-`OperatorApp` before/after SVGs. The 120x40 long-session frame preserves transcript content, spacing rhythm, tool-card alignment, scroll position, composer placement, and contrast. The only visible difference is the expected checkout path in the status line. The recorded geometry is also stable: transcript content 117x28, transcript virtual size 116x105, screen and screen virtual size 118x38, with the transcript scrollbar present and no screen-level overflow.

I also exercised the current head in the real `OperatorApp` for the surfaces affected by the update optimizations:

- A populated nine-subagent band at 80x24 keeps progress visible, truncates labels/activity predictably, exposes `+3 earlier · ctrl+g to expand`, and does not create a screen scrollbar (screen 78x22, virtual 78x22).
- A 40-tool full-page running-subagent view at 120x40 keeps the latest tools and live progress visible, presents the scroll/back/read-only affordances, and confines overflow to the transcript body (body 115x18, virtual 114x81, offset y=63; screen 118x38 with no screen scrollbar).
- The same full page at 48x24 remains usable: the latest tool rows, live progress, scrollbar, `↑↓ scroll`, `esc back to conversation`, and read-only status remain visible (body 43x6, virtual 42x81, offset y=75; screen 46x22 with no screen scrollbar).
- Consecutive settled captures at both sizes retained identical geometry and content. The only frame-to-frame changes were the intentional running spinner/progress animation, not reflow or element movement.

**Verdict: TERMINAL.** No BLOCKER or MAJOR design findings, and no follow-ups are required for this performance-only change.

damianvtran:
### Agent review — round 1

Scope: 798048135bd3a7370caea4772099acfd9fcc8d19..e88141957d077465b5aab920c30557453b63094d

#### Findings

1. **BLOCKER — `FrontendStateStore.state` exposes mutable aliases to canonical nested state** (`local_operator/session/frontend_state.py:550-556`). `model_copy()` is shallow, while `FrontendSessionState`, `JobState`, their lists, and trajectory dictionaries remain mutable Pydantic objects. A caller can therefore mutate `store.state.jobs`, `store.state.jobs[0].trajectory`, or another nested container and silently modify `_state` without a sequence increment or subscriber update. I reproduced both canonical corruption and transport divergence: appending an event through a returned snapshot changed the owner's sequence-0 canonical trajectory; the next status-only job mutation emitted no trajectory append, leaving a follower permanently one event behind. The new test only mutates the top-level scalar `first.sequence`, so it does not exercise the alias boundary the optimization depends on. Make shared nested components actually immutable, or retain isolation at every externally mutable boundary, and add a regression that attempts nested mutation and proves canonical state plus owner/follower trajectories remain unchanged/aligned.

2. **MAJOR — a queued coalesced callback can repaint an obsolete session after session adoption** (`local_operator/tui/app.py:2310-2327`, adoption paths at `app.py:2181-2197` and `app.py:2400-2415`). `_pending_frontend_state` is app-global and is not cleared or generation/session-tagged when the frontend subscription/session changes. I reproduced the ordering `old update queues callback -> new session is adopted and synchronously painted -> queued callback runs`, which applied states in `['new', 'old']`; status and band widgets can thus revert to the prior conversation until another new-session update happens. Clear/invalidate pending work during adoption (or bind it to the originating session and discard stale callbacks), with a test covering this ordering.

3. **MINOR — committed SVG evidence fails the repository whitespace gate** (`docs/performance/tui-lag-2026-08-27/lop-perf-before.svg:176`, `lop-perf-after.svg:176`). `git diff --check 7980481..e881419` reports trailing whitespace in both files. Regenerate/sanitize the artifacts so the diff is clean.

#### Verification

- Focused session and TUI tests: `222 passed in 19.51s`.
- Portable benchmark completed from `/tmp`; measured state read `0.002 ms` median, one-job append `0.414 ms` median / `1.342 ms` max, and mounted/render paths completed.
- Reproduced nested snapshot mutation corrupting canonical state and causing owner/follower trajectory divergence.
- Reproduced stale coalesced repaint applying the old snapshot after a newly adopted snapshot.
- `git diff --check` failed on the two SVG lines noted above.

#### Verdict

**Not clean.** One blocker and one major finding require remediation and a fresh review round on the resulting head.

damianvtran:
### UX review — round 1

Reviewer: ux-reviewer (openai/gpt-5.6-sol)
Scope: `798048135bd3a7370caea4772099acfd9fcc8d19..e88141957d077465b5aab920c30557453b63094d`
Current head: `e88141957d077465b5aab920c30557453b63094d`

I walked the real mounted `OperatorApp` at 120×40 with 100 running children and 500 retained events per child, using keyboard navigation and live mutations rather than inspecting source or stills. The flow covered parent progress, `ctrl+g` roster expansion, 50 Down presses, Enter to open a child, Enter to expand a tool card, 20 streamed child tool transitions, End to reacquire the live tail, Esc back to the parent, and parent progress arriving while the child page was open.

#### Findings

**U1 — MAJOR — The stressed many-subagent flow still blocks visible keyboard feedback for hundreds of milliseconds, and opening the child has no immediate acknowledgement during a roughly one-second wait.**

Concrete steps and observed timings across two runs:

- `ctrl+g` to expand the 100-child roster and focus its first row took 354 ms and 1,216 ms.
- Moving through rows took 158–170 ms median per Down press, with 414–544 ms maximum presses.
- Enter on the selected row took 1,053–1,220 ms before the child view was usable.
- Enter on a settled tool card took 302–626 ms before its expanded state was painted.
- Esc back to the parent took 490–598 ms.

At each step the UI eventually reached the correct state, but during the delay there was no pressed/selected/loading acknowledgement. A user navigating by keyboard cannot distinguish a queued key from a missed key, so repeated presses are likely and can overshoot once the loop catches up. Under the exact long-session/many-subagent conditions this PR addresses, expected behavior is either sub-100 ms acknowledgement for navigation/open/close or immediate feedback that the requested view is loading while heavier folding/layout continues without monopolizing the input loop. Please address the remaining interaction-path stall, not only canonical-state update cost, and re-walk the full keyboard flow at this scale.

#### Verified behavior

- A burst of 50 frontend notifications painted the latest parent state (`parent progress 49`); no stale intermediate state remained queued.
- The 100-child roster exposed all rows, maintained focus order, and selected the intended child.
- The opened child showed the complete retained trajectory at the live tail.
- Twenty streamed tool transitions arrived in order; the last card settled successfully, and a 10 ms heartbeat continued to run during the burst, so updates did not freeze the event loop indefinitely.
- Expanding an earlier card was preserved while later updates arrived. Because that moved the reader off the tail, the view correctly did not yank the scroll position; End reacquired following, and the next streamed message was visible at the tail.
- Parent progress appended while the child page was open appeared on return. Esc restored the parent transcript and preserved the draft.
- No stale or missed feedback was observed. The blocking issue is latency and lack of acknowledgement while those slow interactions complete.

**Verdict: NON-TERMINAL.** One MAJOR UX finding remains; round 2 should verify responsiveness and feedback on the current head after remediation.

damianvtran:

- **Finding 1 (BLOCKER) — Fixed in `00c974bb`**: canonical job DTOs and retained list/dict payloads are now recursively immutable before entering the store. Public snapshots keep a defensive deep-copy boundary for ordinary mutable state while frozen retained events return themselves from deepcopy, preserving the bounded cost. New regressions attempt mutation through the jobs list, trajectory list, and nested event dict; each is rejected, canonical state is unchanged, and an owner/follower update proves trajectory consistency.
- **Finding 2 (MAJOR) — Fixed in `00c974bb`**: frontend callback work now carries a session generation. Both normal adoption and takeover invalidate pending work before the replacement becomes authoritative; a queued old callback returns without clearing or repainting newer state. The regression reproduces `old callback queued -> new snapshot painted -> old callback runs` and asserts only the new title is applied.
- **Finding 3 (MINOR) — Fixed in `00c974bb`**: stripped trailing whitespace from both committed SVG artifacts; `git diff --check` is clean.

Verification on the remediation head:

- focused state/TUI/roster/subagent/stream suite: `146 passed`
- full unit suite: `7324 passed, 7 skipped`
- flake8, Black 26.1.0, isort 5.13.2: clean
- changed-file Pyright: `0 errors, 0 warnings`
- 100 x 500 state benchmark: state read `0.010 ms` median; changed scalar `0.168 ms`; unchanged jobs `0.208 ms`; one-job append `0.916 ms` median / `1.253 ms` max

damianvtran:

- **U1 (MAJOR) — Fixed in `00c974bb`**: profiled the full stressed keyboard path and bounded visible work rather than reducing cadence or hiding progress. The expanded roster now keeps one screenful of rows around a logical full-roster cursor, shifts that window only at its edge, and no longer full-repaints all rows merely for disclosure/navigation. Opening a child paints the mode before folding its retained trajectory, and the resulting row mount is one `batch_append` rather than hundreds of Textual stylesheet/layout settlements. Every child remains arrow-reachable; live updates, spinner cadence, trajectories, and card content are unchanged.

Mounted `OperatorApp`, 120x40, 100 children with 500 retained events per child, same keyboard sequence as the review:

| Interaction | Review round 1 | After `00c974bb` |
| --- | ---: | ---: |
| `ctrl+g` expand | 354-1,216 ms | 164-264 ms |
| Down presses | 158-170 ms median, 414-544 ms max | 112-120 ms median, 207-215 ms max |
| Enter/open child usable | 1,053-1,220 ms | 437-465 ms |
| Enter/expand tool card | 302-626 ms | 292-334 ms |
| Esc/back to parent | 490-598 ms | 168-192 ms |

The large interactions are materially reduced but still include Textual frame settlement in the test pilot. Enter now changes to the child mode before retained folding starts, so the key has immediate visual acknowledgement rather than leaving the parent unchanged through the expensive work. Focused interaction/state/subagent tests passed (`146 passed`), and the full unit suite passed (`7324 passed, 7 skipped`).

damianvtran:
### Agent review — round 2

Scope: 798048135bd3a7370caea4772099acfd9fcc8d19..00c974bb4cc98a98a52639e2c52ff75fb59eb77d

#### Findings

1. **BLOCKER — the remediated snapshot is still not recursively immutable, so callers can silently mutate canonical job state** (`local_operator/session/frontend_state.py:228-235`, fields at `local_operator/session/frontend_state.py:260-268`). `_freeze_job()` freezes only `trajectory` and `latest_details`. Because snapshots intentionally share each frozen `JobState`, other nested mutable values remain aliases into `_state`: at minimum `usage`, `descendant_usage` (including their mutable Pydantic entries), and arbitrary nested `extra="allow"` payloads. `ConfigDict(frozen=True)` prevents assigning a `JobState` field but does not freeze those nested objects. I reproduced this at the reviewed head: `snapshot.jobs[0].usage.input_tokens = 99`, mutation of a `descendant_usage` entry, and `snapshot.jobs[0].unknown["nested"].append(2)` all changed `store.state` at sequence 0 without notification. This is the same canonical-state integrity boundary as round-1 finding 1, and the new regression checks only the two fields `_freeze_job()` special-cases. Recursively freeze every nested job value, including Pydantic child models and allowed extras, or stop sharing any mutable job component; add coverage that enumerates the declared nested fields plus an unknown nested payload.

#### Remediation verification

- Round-1 finding 2 is fixed: session-generation invalidation prevents an old queued paint from overwriting an adopted session. Its focused regression passes.
- Round-1 finding 3 is fixed: `git diff --check 7980481..00c974b` is clean.
- The four focused remediation/navigation regressions pass serially.
- The portable benchmark completed from `/tmp`: 100 x 500 state read `0.020 ms` median, changed scalar `0.158 ms`, unchanged jobs `0.202 ms`, and one-job append `0.870 ms` median / `1.033 ms` max. The structural state-update bound is preserved.

#### Verdict

**Not clean.** One blocker remains, so this round is non-terminal. Do not merge until the canonical snapshot boundary is actually recursively immutable and a fresh review verifies the resulting head.

damianvtran:
### Design review — round 2

Scope: 798048135bd3a7370caea4772099acfd9fcc8d19..00c974bb4cc98a98a52639e2c52ff75fb59eb77d

#### Findings

- **D1 (MAJOR) — Opening a child from the expanded 100-row roster can leave the child page with zero height at a normal terminal size.** I reproduced this in the real `OperatorApp` at 120x40 with 100 populated children and 500 retained events per child. After `ctrl+g`, arrowing to child 100, and opening it, the immediate acknowledgement correctly switches the composer to read-only, but after both settling frames the expanded roster still occupies 26 rows and the mounted `SubagentView` remains `116x0` with a `115x1` body. The screen paints only the roster, not the child title, transcript, progress, scroll affordance, or back hint. Geometry confirms the screen virtual height grows from 38 to 41 while the child body has 103 blocks/100 full tool cards hidden in its one-row box. Collapsing the same roster before opening is the control: the child view becomes `116x15`, its body `115x10` with a `114x205` virtual size, and the child page is visibly usable. Constrained 60x22 already takes the right approach by hiding the roster in compact mode and gives the page a `56x11` / `55x6` region. Opening a child needs to collapse, cap, or hide the expanded roster before laying out the page so the deferred acknowledgement settles into visible child content rather than an empty mode.

Additional evidence: the windowed roster itself remains legible and aligned at 120x40 and 60x22 while traversing the start/middle/end of all 100 children. The normal roster shows 24 rows and the constrained roster 6, focus remains visible, and disclosure accurately reports 76/94 hidden rows. In the usable collapsed/compact child controls, all 100 full tool cards are present, overflow stays inside the transcript body, the screen scrollbar remains absent, and consecutive settled frames retain identical geometry/content apart from the intentional spinner/progress tick. Expanded full-card content also stays within the constrained body.

**Verdict: NON-TERMINAL.** D1 is a MAJOR user-visible regression in the remediated windowing flow. Re-review after the expanded-roster-to-child transition preserves a usable child viewport at normal and constrained sizes.

damianvtran:
### UX review — round 2

Reviewer: ux-reviewer (openai/gpt-5.6-sol)
Scope: 798048135bd3a7370caea4772099acfd9fcc8d19..00c974bb4cc98a98a52639e2c52ff75fb59eb77d
Current head: 00c974bb4cc98a98a52639e2c52ff75fb59eb77d

I repeated the real mounted `OperatorApp` flow at 120×40 with 100 running children and 500 retained events per child. I expanded the roster, moved 74 rows to an off-screen child, opened it with Enter, focused and expanded a retained tool card, applied ten live tool start/end pairs plus progress changes, used End to reacquire the live tail, then returned with Esc while parent progress changed underneath.

#### Findings

**U2 — MAJOR — Opening a child from the expanded stressed roster leaves the child transcript only one row high, so the task the page exists for is effectively hidden behind the roster.**

Concrete step: after `ctrl+g`, 74 Down presses, and Enter on `child-074`, the mode acknowledgement painted promptly, but the still-expanded roster retained 27 rows and the child transcript body measured `115×1` with a `114×405` virtual body. The rendered frame showed the child title and the 24-row roster window, but no readable trajectory row, tool card, live progress body, scroll/back hint, or read-only state hint. Keyboard actions technically still reached the hidden content: I focused and expanded `src/child_74/file_97.py`, streamed ten ordered tool transitions, and used End successfully, but the user cannot see the card or the streamed work they are interacting with. Expected behavior is for entering the child detail mode to collapse or otherwise yield the expanded roster so the selected transcript receives a useful viewport; Esc may restore the expanded roster and its logical position.

#### U1 remediation and update ordering

The original missed-key ambiguity is materially improved. Enter changed mode in 20–27 ms in two runs instead of leaving the parent unchanged during retained folding. `ctrl+g` completed in 187–285 ms, Down presses were 131–154 ms median with focus advancing correctly through the full logical roster, card expansion completed in 221–242 ms, and Esc returned in 156–210 ms. The complete child body settled in about 2.2 seconds, but the immediate mode paint made the open action visibly acknowledged rather than apparently lost.

No stale or missed update was observed:

- all ten live tool transitions appeared in exact order;
- the earlier expanded card remained expanded while later events arrived;
- updates did not yank the reader back to the tail;
- End reacquired following, and the next message appeared at the live tail;
- the latest progress text was current rather than stale;
- Esc restored the parent, draft, selected roster row, and parent progress that arrived while the child page was open;
- screen virtual size remained equal to screen size before and after return.

**Verdict: NON-TERMINAL.** U1's feedback problem is remediated, but U2 is a new MAJOR finding in the required end-to-end stressed flow. A fresh UX round should repeat the same expanded-roster entry path after remediation and verify that the child transcript, card interaction, live progress, and exit hints are actually visible.

damianvtran:

- **Finding 1 (BLOCKER) — Fixed in `2f35036a`**: `_freeze_job()` now traverses every declared field and every `extra="allow"` value. Lists and mappings become recursively immutable compatible containers; nested `Usage` and `FrontendUsage` values become frozen compatible subclasses whose nested extras are recursively frozen; unknown Pydantic extras are preserved as recursively immutable mappings. No mutable nested component of a shared `JobState` remains aliased.

Regression coverage now attempts and rejects mutation through:

- `usage.input_tokens`;
- a `descendant_usage` entry;
- an arbitrary future extra shaped as `{"nested": [1]}`;
- the existing jobs list, trajectory list, and trajectory event dictionary.

Every case verifies canonical sequence/state remains unchanged, and the owner/follower trajectory consistency test remains green.

Performance remains bounded at 100 children x 500 retained events:

```text
state read               0.009 ms median, 0.027 ms max
changed scalar mutation  1.052 ms median, 1.233 ms max
unchanged jobs mutation  1.091 ms median, 1.815 ms max
one-job progress append  2.487 ms median, 4.418 ms max
```

Verification:

- state/frontend tests: `21 passed`
- state + child viewport + roster tests: `129 passed`
- flake8, Black 26.1.0, isort 5.13.2: clean
- changed-file Pyright: `0 errors, 0 warnings`
- `git diff --check`: clean

damianvtran:
### Design review remediation — round 2

- **D1 (MAJOR) — Fixed in `2f35036a`**: opening a child now collapses the expanded roster before the child page is laid out. The roster remains available in its normal collapsed preview and preserves the selected child row; compact mode continues hiding it under the existing rule. The child page therefore owns a usable transcript viewport instead of settling behind 26 roster rows.

Real `OperatorApp` geometry after opening child 100 from the expanded 100-child roster:

| Terminal | Roster expanded after open | Subagent view | Transcript body | Screen virtual / actual |
| --- | --- | ---: | ---: | ---: |
| 120x40 | no | 116x21 | 115x16 | 118x38 / 118x38 |
| 60x22 | no | 56x11 | 55x6 | 58x20 / 58x20 |

Both frames were exported as SVG with `scripts/subagent_viewport_shot.py`, rendered to PNG, and visually inspected. The normal frame shows the title, breadcrumb, prose, tool rows, back/history hint, collapsed roster preview, and read-only composer. The compact frame shows the same child content with its internal transcript scrollbar and the roster correctly hidden. No screen-level scrollbar or virtual overflow appeared.

A mounted regression covers both terminal sizes and asserts the roster collapses, child/body heights are non-zero, and screen virtual height stays within the screen.

damianvtran:
### UX review remediation — round 2

- **U2 (MAJOR) — Fixed in `2f35036a`**: the same expanded-roster transition now collapses the navigation surface before child layout, so immediate mode acknowledgement settles into a usable child transcript rather than a 115x1 body. The selected child remains current in the collapsed preview, Esc still returns to the parent/composer, and `ctrl+g` can reopen full-roster navigation.

Measured mounted flow after selecting child 100 from the expanded roster:

- child mode acknowledgement remains on the deferred-open path introduced for U1;
- 120x40 transcript body settles at `115x16` rather than `115x1`;
- 60x22 compact transcript body settles at `55x6`;
- both screens have `virtual_size == size` vertically and preserve internal child scrolling;
- normal and compact rendered frames were viewed and show title, child content, tools, back hint, and composer state.

The real-app geometry regression runs the expanded-roster-to-child transition at both sizes and verifies the usable viewport and overflow constraints.

damianvtran:
### Agent review — round 3

Scope: 798048135bd3a7370caea4772099acfd9fcc8d19..2f35036a5ac345009facd7405d84d66fd26c022f

#### Findings

1. **BLOCKER — future `JobState` fields/extras can still expose mutable aliases to canonical state** (`local_operator/session/frontend_state.py:218-234`). `_freeze_value()` recursively freezes models, dicts, lists, and tuples, but returns every other value unchanged. Pydantic's `extra="allow"` accepts a set in an unknown field and serializes it in JSON mode, so this is part of the promised future-extra boundary. I reproduced `store.state.jobs[0].future_tags.add("CORRUPTED")` changing the owner's canonical sequence-0 state and subsequent `model_dump(mode="json")` output without a sequence increment or subscriber notification. This leaves the round-1 snapshot-alias blocker incompletely remediated. Freeze or defensively isolate every mutable JSON-serializable container accepted at this boundary (at minimum sets, with an approach that remains safe for future declared/extras), and extend the regression beyond the currently covered dict/list shape.

#### Prior-finding verification

- The nested list/dict trajectory aliases, `Usage`, `FrontendUsage`, and the currently tested dict/list future extra reject mutation; owner/follower trajectory tests pass.
- Session-generation invalidation discards queued paints from an adopted-away session; the focused stale-session regression passes.
- Expanded-roster child opening collapses the roster and its mounted normal/compact viewport regressions pass.
- Focused state, roster, and child-view suite: `124 passed in 11.96s`.
- Serial coalescing/generation tests pass. The full focused benchmark completed with state read `0.008 ms` median, changed scalar mutation `0.970 ms`, unchanged jobs mutation `0.999 ms`, and one-job append `2.485 ms`; retained-state costs remain bounded for the measured 100 x 500 case.
- `git diff --check` is clean.

#### Verdict

**Not clean.** One blocker remains; remediation and a fresh review round are required.

damianvtran:
### Design review — round 3

Scope: 798048135bd3a7370caea4772099acfd9fcc8d19..2f35036a5ac345009facd7405d84d66fd26c022f

#### Findings

No findings.

D1 is resolved in the real `OperatorApp`. I opened child 100 from the expanded 100-child roster and visually inspected rendered normal (120x40) and compact (60x22) frames, including deep populated transcripts with 91 full tool cards.

- The expanded roster collapses before child layout in both sizes. Normal settles to a 116x15 child view with a 115x10 transcript body; compact settles to 56x11 / 55x6.
- Deep content remains complete and internally scrollable: 187 rendered rows, body virtual heights 320/321, and page-up moves 310→300 at 120x40 and 315→309 at 60x22.
- The child title, breadcrumb, tool outcomes, prose, scroll affordance, back hint, collapsed roster preview where space allows, and read-only composer remain visible and aligned. Compact mode correctly spends the constrained height on child content rather than the roster.
- Screen geometry stays bounded: 118x38 == 118x38 and 58x20 == 58x20, with no screen-level scrollbar. Overflow remains inside the child transcript.
- Two consecutive settled SVG frames at each size were byte-identical, so the remediated layout does not visibly reflow after settling.

**Verdict: TERMINAL.** D1 is remediated; no BLOCKER or MAJOR design findings remain at the reviewed head.

damianvtran:
### UX review — round 3

Reviewer: ux-reviewer (openai/gpt-5.6-sol)
Scope: 798048135bd3a7370caea4772099acfd9fcc8d19..2f35036a5ac345009facd7405d84d66fd26c022f

I re-walked the current head in the real mounted `OperatorApp` at 120×40 with 100 running children and 500 retained events per child. The keyboard flow covered `ctrl+g`, 74 Down presses to the previously off-screen `child-074`, Enter and its immediate mode acknowledgement, the settled child viewport, tool-card expansion, ten ordered live tool transitions, End to resume tail following, parent updates while the child was open, and Esc back to the conversation with a pre-existing draft.

#### Findings

**U3 — MAJOR — Esc returns to the parent with keyboard focus trapped on an invisible child row, so typing into the visibly restored draft does nothing.**

Concrete step: after opening `child-074` from the expanded roster, the remediation correctly collapsed the roster to give the child transcript space. I then pressed Esc from the child page. The conversation and draft returned, but focus remained a `SubagentRow` for `child-074` whose `display` was false; the collapsed roster was showing only children 094–099. Pressing `x` left the visible draft unchanged, with no feedback, and Down advanced the hidden logical selection to `child-075` while the screen continued to show children 094–099. The user therefore sees the composer and assumes typing can resume, but keystrokes are silently routed to an off-screen navigation context.

Expected behavior: Esc should either restore the expanded roster at the selected child and keep that focused row visible, or restore focus to the composer as the page’s “esc back to conversation” wording implies. In either case there must be no focused row with `display: false`, and typing after returning to the visible draft must work immediately unless the UI visibly communicates that roster navigation is still active.

#### U1/U2 verification

- **U1 remains fixed:** Enter produced visible child-mode acknowledgement in 18 ms on the primary run while the retained trajectory continued loading. The complete 203-block child view settled without a missed key; live updates remained ordered, the event-loop heartbeat continued, the expanded card survived updates, End reacquired live following, and the next update appeared at the tail.
- **U2 is fixed:** the expanded 24-row roster collapsed on entry. The child transcript had a usable `115×10` body on the primary run (`114×405` virtual content), visibly showed trajectory rows, tool cards, live progress, scrollbar, back hint, and read-only state, and the screen stayed at `118×38` physical and virtual height with no screen-level overflow.
- Ten live tool transitions appeared in exact order, latest progress was current, and the parent-side progress update was present after return. The draft text itself was preserved; U3 concerns the invisible focus state that prevents the user from continuing it.

**Verdict: NON-TERMINAL.** U1 and U2 are resolved, but U3 is a fresh MAJOR keyboard-flow finding. Re-review should repeat the expanded-roster → off-screen child → Esc → type/Down path and verify that restored focus is both visible and consistent with the collapsed/expanded roster state.

damianvtran:

- **Finding 1 (BLOCKER) — Fixed in `f1a9d166`**: `_freeze_value()` now covers the complete Python mutable-container boundary accepted by unconstrained/future extras: `MutableMapping`, `MutableSequence`, `deque`, `MutableSet`, `bytearray`, and `memoryview`. Sets become recursively frozen `frozenset` values, preserving membership/equality and Pydantic JSON array serialization; bytearray/memoryview become bytes; tuples and existing frozensets recursively freeze their members. Fully frozen jobs are detected and reused, so routine updates do not re-walk every retained event.

Regression coverage adds future extras containing:

- a nested mutable set, with mutation rejected and set equality preserved;
- a deque containing a nested mutable list/dict;
- a bytearray;
- JSON-mode dump plus `JobState.model_validate()` round-trip assertions.

Canonical state remains at sequence 0 after all mutation attempts.

100 children x 500 retained events after the complete container audit:

```text
state read               0.008 ms median, 0.028 ms max
changed scalar mutation  0.526 ms median, 0.662 ms max
unchanged jobs mutation  0.539 ms median, 0.596 ms max
one-job progress append  1.545 ms median, 1.821 ms max
subagent fold             1.370 ms median, 1.602 ms max
```

Verification: focused state/frontend/roster/subagent suite `130 passed`; flake8, Black, isort, changed-file Pyright, and `git diff --check` clean.

damianvtran:
### UX review remediation — round 3

- **U3 (MAJOR) — Fixed in `f1a9d166`**: closing a child now restores the originally focused control only when it remains visible. Opening from expanded roster navigation collapses that roster and can hide the selected row, so Esc falls back to the composer instead. The draft is untouched; the next typed key edits it immediately, and Down remains owned by the composer rather than navigating invisible rows.

The mounted real-flow regression performs: seed draft -> expand roster -> navigate to a row that collapse hides -> Enter child -> settle -> Esc -> type `x` -> Down. It asserts the old row is hidden, composer owns focus, the draft receives `x`, and Down does not transfer focus back into the hidden roster.

The focused state/frontend/roster/subagent flow passed (`130 passed`); formatting, typing, and diff gates are clean.

damianvtran:
### Design review — round 4

Scope: 798048135bd3a7370caea4772099acfd9fcc8d19..f1a9d16686481b91e05e348929f5598d882bf708

No D findings.

I reran the committed real-`OperatorApp` child-viewport capture at the current head and visually inspected both populated frames after the round-3 focus changes:

- At 120x40, opening child 100 from the expanded 100-child roster collapses the roster and leaves a clear 116x21 child view with a 115x16 transcript body. The title, breadcrumb, prose, tool outcomes, navigation copy, collapsed roster preview, read-only composer, and status line remain visible, aligned, and legible.
- At 60x22, compact mode correctly hides the roster and devotes the constrained height to a usable 56x11 child view with a 55x6 internally scrollable body. The essential child context, tool results, back/scroll affordances, composer state, and status remain visible without clipping.
- Screen geometry remains bounded at both sizes: 118x38 virtual/actual and 58x20 virtual/actual. No screen-level scrollbar or overflow appeared; compact overflow stays within the child body.
- Two consecutive settled SVG captures at each size were byte-identical, confirming the latest focus restoration changes introduced no visible reflow or frame instability.

**Verdict: TERMINAL.** The round-3 remediations remain visually sound at the current head; no BLOCKER or MAJOR design findings remain, and no design follow-ups are required.

damianvtran:
### Agent review — round 4

Scope: 798048135bd3a7370caea4772099acfd9fcc8d19..f1a9d16686481b91e05e348929f5598d882bf708

#### Findings

1. **BLOCKER — generic future extras can still retain mutable aliases through JSON-serializable dataclass values** (`local_operator/session/frontend_state.py:225-259`). `_freeze_value()` now covers the named built-in containers, but its fallback still returns unknown values unchanged. Pydantic accepts a dataclass instance in an `extra="allow"` field and `model_dump(mode="json")` serializes it. I reproduced this with `@dataclass class Payload: nested: list[int]`: after placing `Payload([1])` in a `JobState` future extra and constructing `FrontendStateStore`, `store.state.jobs[0].future_payload.nested.append(2)` changed the canonical sequence-0 wire dump to `{"nested": [1, 2]}` with zero updates, and the public and canonical values were the same object. Thus the round-3 requirement for generic future extras remains open. Freeze dataclass/other Pydantic-serializable structured extras recursively, or defensively isolate every unsupported fallback value rather than sharing it; add this regression at the public `store.state` boundary.

#### Prior-finding verification

- Set extras freeze to `frozenset`, preserve set equality/membership and JSON array serialization, and reject `.add()`.
- Deque extras freeze recursively to `_FrozenList`; nested list/dict mutation is rejected and JSON output remains an array.
- `bytearray` and `memoryview` freeze to `bytes` and retain the tested UTF-8 JSON string representation.
- Existing nested dict/list trajectory, usage, stale-session repaint, and roster/child viewport regressions were re-run on the current head.
- Focused frontend-state suite: `16 passed`. Combined state/app/subagent suite: `305 passed, 1 failed`; the failure was the pre-existing timing-sensitive `test_a_replayed_aborted_bang_card_stays_shut_and_dim` observing history before its card mounted, unrelated to the freeze diff.
- `git diff --check 798048135bd3a7370caea4772099acfd9fcc8d19..f1a9d16686481b91e05e348929f5598d882bf708` is clean.

#### Verdict

**Not clean.** One blocker remains; remediation and a fresh review round are required.

damianvtran:
### UX review — round 4

Reviewer: ux-reviewer (openai/gpt-5.6-sol)
Scope: 798048135bd3a7370caea4772099acfd9fcc8d19..f1a9d16686481b91e05e348929f5598d882bf708

No U findings.

I repeated the real mounted `OperatorApp` keyboard flow at 120x40 on the current head with 100 running children and 500 retained events per child. I expanded the roster, moved 74 rows to the off-screen `child-074`, opened it, waited through the retained trajectory fold, then pressed Esc and immediately exercised typing and Down against a pre-existing draft.

#### Remediation verification

- **U3 is fixed:** Esc returned focus to the visible `Editor` because the originally focused `SubagentRow` was hidden by the roster collapse. The draft remained exactly `preserved draft`; the next `x` immediately changed it to `xpreserved draft`; Down left both focus and draft with the composer rather than navigating the invisible row. The prior selected row remained `display: false`, confirming this exercised the fallback rather than accidentally restoring a visible row.
- **U2 remains fixed:** opening the off-screen child collapsed the expanded roster and settled to a usable `116x21` child view with a `115x16` transcript body and 202 mounted blocks. Overflow stayed inside the child's `114x703` virtual body, while screen virtual and physical geometry both remained `118x38`.
- **U1 remains fixed:** the synchronous Enter transition acknowledged child mode in 23 ms before the deferred retained fold. A 10 ms heartbeat advanced during the first heavy refresh, so async work did not freeze the input loop. All 74 Down actions reached the intended `child-074`, and no key was missed or applied to stale state.

The focused mounted regressions for the normal/compact expanded-roster viewport and hidden-row Esc return also passed: `3 passed`.

**Verdict: TERMINAL.** No BLOCKER or MAJOR UX findings remain at the current head, and there are no UX follow-ups from this round.

damianvtran:

- **Finding 1 (BLOCKER) — Fixed in `feeae2cf`**: replaced the open-ended fallback with a closed canonical-value boundary. Known scalar/frozen/container/Usage paths remain fast. Every other candidate is now normalized by Pydantic's own `to_jsonable_python`, then recursively frozen; dataclasses and namedtuples become immutable mapping/sequence representations, serializable Pydantic models enter their compatible frozen path, and scalar serializers (path/date/enum style values) become immutable scalars. Arbitrary objects with `__dict__`, slots, or no supported serializer are rejected with `TypeError` before canonical storage rather than shared.

The audit now covers all Python mutable container categories plus structured object categories:

- mapping, sequence, set, deque, bytearray, memoryview;
- tuple/frozenset recursion;
- dataclass, namedtuple, and custom Pydantic values;
- arbitrary `__dict__`, slots, and opaque objects (explicit rejection).

Parameterized regressions mutate every accepted structured representation and assert canonical state/sequence remain unchanged. Rejection tests prove unsupported objects never enter the store. Existing owner/follower consistency coverage remains green.

Performance at 100 children x 500 retained events:

```text
state read               0.010 ms median, 0.023 ms max
changed scalar mutation  0.525 ms median, 0.684 ms max
unchanged jobs mutation  0.539 ms median, 0.575 ms max
one-job progress append  1.461 ms median, 1.895 ms max
subagent fold             1.362 ms median, 3.896 ms max
```

Verification: closed-boundary state/frontend + Esc focus flow `24 passed`; flake8, Black, isort, changed-file Pyright, and `git diff --check` clean.

damianvtran:
### Agent review — round 5

Scope: 798048135bd3a7370caea4772099acfd9fcc8d19..feeae2cff0448ca794fd67883866b1aa6b854c2e

#### Findings

1. **BLOCKER — the custom frozen list/dict subclasses can be mutated through their built-in base-class methods, so the canonical snapshot boundary remains writable** (`local_operator/session/frontend_state.py:182-223`, exposed by `local_operator/session/frontend_state.py:721-725`). `_FrozenList` and `_FrozenDict` override normal mutators, but Python still permits `list.append(snapshot.jobs[0].trajectory, event)` and `dict.__setitem__(snapshot.jobs[0].trajectory[0], key, value)`. Because `__deepcopy__` deliberately returns these objects themselves, those calls mutate the owner's canonical sequence-0 state without notification. I reproduced transport divergence on this exact head: starting owner/follower stores with an empty trajectory, mutating the owner's returned trajectory via `list.append`, and then making a status-only job mutation produced no trajectory append; the owner retained the injected event while the follower remained empty (`owner != follower`). The same bypass applies to every normalized dataclass, namedtuple, Pydantic, mapping, sequence, and deque representation backed by these subclasses. Use representations that cannot be modified through inherited C-level base methods, or stop sharing them across the public boundary; add adversarial regressions invoking the base methods and proving canonical sequence, serialization, and owner/follower equality remain unchanged.

#### Verification

- Read the full `79804813..feeae2cf` diff and all earlier review/remediation rounds.
- Focused frontend-state suite: `18 passed`.
- Adversarial reproduction above fails the intended canonical-value boundary at sequence 0 and demonstrates owner/follower divergence.
- Full unit and static quality suites were also started; this correctness reproduction is independent of their outcome.

#### Verdict

**Not clean.** One blocker remains; remediation and a fresh review round are required.

damianvtran:

- **Finding 1 (BLOCKER) — Fixed in `323c7492`**: removed the mutable built-in subclasses entirely. Canonical lists now use a `Sequence` wrapper backed by a private tuple; mappings use a `Mapping` wrapper backed by private immutable items plus an encapsulated read index. Neither object is a `list`/`dict`, so `list.append(obj, ...)` and `dict.__setitem__(obj, ...)` fail at the C-level type check rather than bypassing Python overrides. Equality remains compatible with ordinary sequences/mappings.

`JobState` and `FrontendSessionState` wrap serializers thaw wrappers only into ephemeral wire copies, preserving expected JSON list/dict shapes. Job wire summaries are built without visiting trajectories, and already-frozen job identities short-circuit equality, preserving the update bound.

Adversarial regressions prove:

- jobs, trajectory, and event values are not list/dict subclasses;
- direct base-class mutation calls fail;
- ordinary mutation paths fail;
- JSON/model serialization retains exact list/dict wire shapes;
- canonical sequence stays 0 after attempted mutation;
- owner/follower jobs remain equal and a no-op mutation publishes nothing.

100 children x 500 retained events:

```text
state read               0.009 ms median
changed scalar mutation  0.994 ms median
unchanged jobs mutation  0.982 ms median
one-job progress append  3.689 ms median, 4.175 ms p95
subagent fold             1.520 ms median
```

Verification: frontend-state/sync suite `24 passed`; flake8, Black, isort, changed-file Pyright, and `git diff --check` clean.

damianvtran:
### Agent review — round 6

Scope: 798048135bd3a7370caea4772099acfd9fcc8d19..323c7492c5faf93ca2504f493c8ea86c6aa17564

#### Findings

1. **BLOCKER — the replacement wrappers still expose writable canonical storage and internally disagree after mutation** (`local_operator/session/frontend_state.py:193-243`, exposed by `local_operator/session/frontend_state.py:799-803`). `_FrozenMapping._values` is an ordinary public-by-convention `dict`, so a caller can mutate `store.state.jobs[0].trajectory[0]._values` directly; `_FrozenSequence._values` can likewise be replaced with `object.__setattr__`. Because deep copies share these wrappers, both alter canonical sequence-0 state without a notification. I reproduced changing the mapping's `_values`, after which `model_dump(mode="json")` emitted the corrupted event while sequence remained 0 and `mutate(jobs=list(owner.state.jobs))` returned `None`. More seriously, `_FrozenMapping.__eq__` reads the now-stale `_items` while lookup/iteration/serialization read `_values`, so the corrupted owner still compared equal to an unmodified follower even though their JSON trajectories differed. Make the wrappers' complete backing state structurally unreachable/unmodifiable rather than relying on underscore encapsulation (and ensure equality and serialization derive from the same immutable representation); add regressions for direct backing-attribute mutation/rebinding that prove canonical wire output, sequence, equality, and owner/follower state cannot diverge.

#### Verification

- Read the exact `79804813..323c7492` diff and all prior review/remediation rounds.
- Confirmed PR head remains `323c7492c5faf93ca2504f493c8ea86c6aa17564`; `git diff --check` is clean.
- Focused frontend-state/sync suite: `24 passed in 3.44s`.
- Reproduced canonical mutation through `_FrozenMapping._values`, stale equality versus wire serialization, a suppressed no-op update, and `_FrozenSequence` backing replacement through `object.__setattr__`.

#### Verdict

**Not clean.** One blocker remains; remediation and a fresh review round are required.

damianvtran:

- **Finding 1 (BLOCKER) — Fixed in `cf0d3e9b`**: wrapper storage is now intrinsically immutable. The sequence is a tuple subclass with no slots or `__dict__`. The mapping is a tuple-of-key/value-pairs subclass with no mutable lookup dictionary, slots, or rebindable backing attribute; lookup, iteration, equality, and serialization all derive from that same tuple representation. Virtual `Mapping` registration preserves the runtime mapping contract without adding storage.

Copy/pickle boundaries are explicit: shallow/deep copy return the immutable value; pickle reconstructs a new wrapper from immutable tuples. There is no mutable reference exposed through attributes, `vars`, slots, iterators, dataclass fields, or reduce payloads.

Adversarial regressions cover jobs/trajectory/event wrappers and assert:

- no `list`/`dict` inheritance and no `__dict__`;
- base-class mutation rejection;
- backing-attribute rebinding rejection via `object.__setattr__`;
- shallow/deep copy identity;
- pickle round-trip type and equality;
- JSON wire shape equals expected list/dict content;
- canonical sequence remains 0;
- equality and JSON agree between owner/follower after mutation attempts.

Wire serializers thaw only ephemeral copies. Job summaries remain trajectory-free, and frozen job identity short-circuits unchanged rows.

Performance at 100 children x 500 retained events:

```text
state read               0.010 ms median
changed scalar mutation  0.798 ms median
unchanged jobs mutation  0.789 ms median
one-job progress append  3.273 ms median
```

Verification: frontend-state/sync suite `24 passed`; flake8, Black, isort, changed-file Pyright, and `git diff --check` clean.

damianvtran:
### Agent review — round 7

Scope: 798048135bd3a7370caea4772099acfd9fcc8d19..cf0d3e9b128df8226b79f74bf3dd14ad6aae0fe8

#### Findings

1. **BLOCKER — immutable tuple wrappers still sit inside shared, writable Pydantic backing dictionaries** (`local_operator/session/frontend_state.py:365-382`, `local_operator/session/frontend_state.py:393`, `local_operator/session/frontend_state.py:511-526`, exposed by `local_operator/session/frontend_state.py:820-825`). `_freeze_job()` returns a frozen Pydantic `JobState`, but `FrontendStateStore.state` deep-copies that object by sharing it because its tuple-backed children return themselves from `deepcopy`; the returned snapshot's `JobState.__dict__`, `JobState.__pydantic_extra__`, and nested `_FrozenUsage.__dict__` therefore remain canonical mutable storage. I reproduced all of these via `object.__getattribute__`: assigning `snapshot.jobs[0].__dict__["trajectory"] = ()`, changing `status`, changing `snapshot.jobs[0].usage.__dict__["input_tokens"]`, and inserting into `snapshot.jobs[0].__pydantic_extra__` each silently changed the owner's sequence-0 wire state. In the trajectory case the follower retained the original event while the owner emitted `[]`, and `owner.mutate(jobs=list(owner.state.jobs))` returned `None`, suppressing repair/notification. Closing the tuple wrappers' own storage does not close the snapshot boundary while their owning Pydantic models are shared; either make the complete shared job/usage representation intrinsically immutable with no writable Pydantic backing storage, or stop sharing those model objects across `state` reads, then regress every `object.__getattribute__` path above and prove sequence, wire output, equality, and owner/follower state remain aligned.

#### Verification

- Read the complete `79804813..cf0d3e9b` diff and rechecked all prior review/remediation rounds.
- Confirmed exact PR head `cf0d3e9b128df8226b79f74bf3dd14ad6aae0fe8`; `git diff --check` is clean.
- Wrapper-specific tests pass: `19 passed in 4.31s`, but they stop at tuple wrapper attributes and do not attack the shared owning Pydantic internals.
- Focused frontend-state/app suite reached `225 passed`; three unrelated timing-sensitive TUI tests failed under parallel load.
- Adversarial owner/follower reproduction confirmed canonical mutation at sequence 0, divergent wire trajectories, and a suppressed no-op update.
- Fresh 100-child × 500-event measurements: state read `0.008 ms` median, changed scalar `0.520 ms`, unchanged jobs `0.522 ms`, and one-job append `2.290 ms` median. A 100-sample append run had `2.515 ms` median but load-sensitive outliers, so the structural common path is below 10 ms but a hard maximum below 10 ms was not demonstrated.

#### Verdict

**Not clean.** One blocker remains; remediation and a fresh independent review round are required.

damianvtran:

- **Finding 1 (BLOCKER) — Fixed in `969fe9bf`**: canonical storage no longer shares any owning Pydantic model with callers or mutation inputs. Store ingestion always creates new `JobState` owners and frozen nested values. `state` creates a detached `FrontendSessionState`, detached per-job `JobState`, detached `Usage`/`FrontendUsage`, and detached Pydantic extras while structurally sharing only the proven intrinsic tuple/tuple-of-pairs retained payload wrappers. Scalar mutations preserve canonical-owned jobs without recloning; jobs mutations detach their input owners before storage.

Adversarial coverage mutates every exposed backing path through `object.__getattribute__`/`__dict__`/`__pydantic_extra__`:

- public job status and trajectory;
- public job usage internals;
- descendant usage internals;
- nested future Pydantic extra internals;
- insertion into public `__pydantic_extra__`;
- original input job, trajectory, usage, and extra after store construction;
- incoming job and trajectory after `mutate(jobs=...)`.

Every mutation leaves canonical sequence and JSON unchanged, owner/follower JSON aligned, and any repair update is applied to both sides rather than suppressed.

100 children x 500 retained events:

```text
state read               0.524 ms median, 0.581 ms max
changed scalar mutation  0.006 ms median, 0.105 ms max
unchanged jobs mutation  0.580 ms median, 0.702 ms max
one-job progress append  2.252 ms median, 2.458 ms max
subagent fold             1.343 ms median, 1.479 ms max
```

Verification: frontend-state/sync suite `26 passed`; flake8, Black, isort, changed-file Pyright, and `git diff --check` clean.

damianvtran:
### Agent review — round 8

Scope: 798048135bd3a7370caea4772099acfd9fcc8d19..969fe9bf5bb94fdf178396a6fef3197f30d2fec2

#### Findings

1. **MINOR — frozen usage containers emit Pydantic serializer warnings on ordinary Python-mode dumps** (`local_operator/session/frontend_state.py:519-532`). The focused suite reports `PydanticSerializationUnexpectedValue` for `_FrozenSequence` in `Usage.cost_components` in two tests. JSON-mode and JSON-string serialization are correct in direct checks, so this is not a correctness or wire blocker, but callers using `model_dump()` inherit noisy warnings. Teach the usage serialization boundary to thaw its frozen list fields, as the job/state serializers already do.

No BLOCKER or MAJOR findings.

#### Verification

- Read the complete `79804813..969fe9bf` diff and rechecked all prior findings and remediation rounds.
- Confirmed exact PR head `969fe9bf5bb94fdf178396a6fef3197f30d2fec2`; `git diff --check` is clean.
- Adversarially attacked original construction inputs, later `mutate(jobs=...)` inputs, public snapshots, shallow `model_copy`, `__dict__`, `__pydantic_extra__`, nested usage/descendant usage/future-model extras, trajectory wrappers, subscriber callback copies, and returned update models. Canonical sequence/JSON remained unchanged by aliases, callback/update mutation did not affect storage, and owner/follower JSON remained aligned after the real update.
- Exercised append and capped-trajectory replacement updates through `FrontendUpdate` JSON round-trips, no-op equality, and `FrontendSync` serialization/validation. Owner and follower matched exactly.
- Focused frontend-state/sync suite: `26 passed` in 4.34s (with the two warnings above).
- Full unit suite: `7333 passed, 7 skipped` in 143.64s.
- Full flake8, Black 26.1.0, isort, and Pyright gates completed successfully.
- Fresh 100-child × 500-event benchmark: state read `0.547 ms` median / `0.601 ms` max; changed scalar `0.007 ms` median / `0.123 ms` max; unchanged jobs `0.580 ms` median / `0.707 ms` max; one-job append `2.380 ms` median / `2.753 ms` max. Full mounted/render benchmark completed.

#### Verdict

**Clean. TERMINAL.** No blocker or major findings remain. The minor warning cleanup is a follow-up and is not a reason for another review round.

damianvtran:
### Agent review remediation — round 8

- **Finding 1 (MINOR) — Fixed in `20566aea`**: `_FrozenUsage` and `_FrozenFrontendUsage` now use the same wrap-serializer boundary as jobs/state. Their immutable nested sequences are thawed only into an ephemeral model passed to Pydantic's handler, so both ordinary Python-mode and JSON-mode dumps see the declared list shape without warnings. Canonical immutable ownership remains unchanged.

Verification:

- the two formerly warning-producing tests pass with `-W error::UserWarning` (`2 passed`);
- full frontend-state/sync slice passes warning-free (`26 passed`);
- flake8, Black, isort, changed-file Pyright, and `git diff --check` clean;
- 100 x 500 benchmark: state read `0.513 ms` median, changed scalar `0.006 ms`, unchanged jobs `0.594 ms`, one-job append `2.388 ms` median / `2.837 ms` max.

damianvtran:
### Design review — round 5

Scope: 798048135bd3a7370caea4772099acfd9fcc8d19..20566aea6346b4653657be6dff4a2d1eabe00035

No D findings.

I rerendered and inspected the real `OperatorApp` child viewport at the current head in both normal (120x40) and compact (48x24) sizes, and compared them with fresh captures from the serializer cleanup's parent (`969fe9bf`). The child header, conversation breadcrumb, narrative and tool rows, failure treatment, navigation copy, read-only composer, status bar, spacing, alignment, contrast, and compact truncation remain visually unchanged. The compact before/after SVGs are byte-identical; the normal frame retains the same visible content and styling, with only asynchronous roster-height settlement and the expected checkout path differing between captures.

Current-head geometry remains bounded with no screen-level overflow:

- 120x40: view 116x19, body 115x14, body virtual 114x14, screen and screen virtual 118x38.
- 48x24: view 44x13, body 43x8, body virtual 42x8, screen and screen virtual 46x22.

**Verdict: TERMINAL.** No BLOCKER or MAJOR design findings, and the warning-only serializer cleanup introduced no current-head visual regression.

damianvtran:
### Agent review — round 9

Scope: 798048135bd3a7370caea4772099acfd9fcc8d19..20566aea6346b4653657be6dff4a2d1eabe00035

#### Findings

No BLOCKER, MAJOR, MINOR, or NIT findings.

#### Verification

- Read the complete `79804813..20566aea` diff and the prior review/remediation history, then specifically rechecked the `969fe9bf..20566aea` wrap-serializer cleanup.
- Confirmed exact PR head `20566aea6346b4653657be6dff4a2d1eabe00035`; `git diff --check` is clean.
- Focused frontend-state/sync suite passed with Pydantic serializer warnings promoted to errors: `26 passed in 2.05s`.
- Adversarial serialization checks passed warning-free for Python mode, JSON mode, and `model_dump_json()`, including JSON and Python round trips, nested cost components, future extras, serializer flags, and mutation of dumped containers. The ephemeral thaw copies did not replace or alias the canonical `_FrozenUsage`, `_FrozenFrontendUsage`, `_FrozenSequence`, or frozen trajectory/extra payloads.
- Fresh 100-child × 500-event benchmark: state read `0.527 ms` median / `0.655 ms` max; changed scalar `0.007 ms` median / `0.146 ms` max; unchanged jobs `0.587 ms` median / `0.646 ms` max; one-job append `2.338 ms` median / `2.524 ms` max. The full mounted/render benchmark also completed.
- A blanket `-W error::UserWarning` pytest invocation is currently intercepted during pytest configuration by the repository's existing `pytest-asyncio` default-loop-scope deprecation warning, before test collection. The focused Pydantic-warning filter above promotes the serializer warning class under review without masking that unrelated harness warning.

#### Verdict

**Clean. TERMINAL.** No blocker or major findings remain; no further code-review round is required for this head.

damianvtran:
### UX review — round 5

Reviewer: ux-reviewer (openai/gpt-5.6-sol)
Scope: 798048135bd3a7370caea4772099acfd9fcc8d19..20566aea6346b4653657be6dff4a2d1eabe00035

No U findings.

I repeated the real mounted `OperatorApp` keyboard flow at 120x40 on the current head with 100 running children and 500 retained events per child. I preserved a composer draft, expanded the roster, moved 74 rows to off-screen `child-074`, opened its live full-page view, expanded a retained tool card, streamed and settled ten further tool transitions, injected parent progress while the child was open, then pressed Esc and immediately exercised typing and Down.

#### Current-head confirmation

- The 100-child roster remained keyboard-reachable: all 74 Down actions reached `child-074` without a missed key or stale selection. Median Down handling was 87 ms, maximum 176 ms.
- Enter synchronously mounted the child mode before retained-history folding. The expanded roster collapsed, the child page retained usable overflow inside its transcript, and the screen stayed bounded at `118x38` physical and virtual geometry.
- The retained card expanded from the keyboard and stayed expanded while ten new tool calls arrived and settled. Those ten cards appeared in exact call order, the latest settled successfully, and a 10 ms heartbeat advanced throughout the stressed flow, confirming the input loop continued to make progress.
- Parent progress written while the child page was open was present after return.
- Esc removed the child page and restored focus to the visible `Editor`, not the now-hidden roster row. The draft remained `preserved draft`; the next `x` immediately changed it to `xpreserved draft`; Down remained owned by the composer.
- The focused mounted regressions for expanded-roster viewport preservation, hidden-row Esc focus restoration, and live child updates also passed: `4 passed`.

**Verdict: TERMINAL.** No BLOCKER or MAJOR UX findings remain at the current head, and there are no UX follow-ups from this round.

PR #360, opened 2026-08-28
fix(session): announce stored credentials to the model

damianvtran:
## Summary

Fixes session-credential discoverability by delegated agents and by the model itself. Ground truth: session `835fbcafdc27`, reproduced live on HEAD — the operator stored a credential via `/credential OSWORLD_OPENAI_API_KEY` and said "I just added the API key"; the model called `list_variables` (which said nothing), guessed `OPENAI_API_KEY` in bash (missing), and only found the real name in the `<session-credentials>` prompt tail 10 minutes later.

Three dead ends, three fixes:

1. **No model-visible announcement.** Storing/forgetting a credential now journals a `session_credential` custom message (same pattern as `session_model_switch`): rendered as a user turn by `_default_convert_to_llm`, allow-listed in `_PERSISTABLE_CUSTOM_TYPES` so resume replays it, and journaled through `_append_or_park_journal` so a mid-turn store parks safely. Wired from BOTH store paths — the TUI `/credential` flow (store, forget, forget-all) and the `ask` tool's secret path (via a new `journal_credential` hook on `ToolContext`, installed by `Session._build_tool_context`; a hook failure never fails the ask whose answer is in hand). Text names the KEY only and states the contract: env var on every bash command, never print it, never read it.
2. **`list_variables` dead-end.** The listing now appends a names-only section: `session credentials (secret, usable via bash env, not readable): FOO_KEY, BAR_KEY`. `read_variable` on a stored credential name returns the bash-env contract (`SESSION CREDENTIAL: <name> … injected as an environment variable … use it there, never echo it`) instead of the circular `unknown variable (see list_variables)` — detected via `credential_names()` membership, checked BEFORE the `names()` unknown-variable branch.
3. **Subagent inheritance hardened and tested.** The wiring existed but had zero coverage, and the child construction-time `ToolContext` omitted `variables=` entirely — the same trap `session_factory.py` documents ("a store installed only here reached the createIf check and nothing else"). Added `variables=getattr(parent_session, "_variables", None)` with a comment citing that trap, plus tests proving end-to-end: (a) parent credential named in a spawned child's system tail, (b) child bash injects the parent's credential env, (c) a credential stored AFTER a child launched reaches that child's later turns (store shared by reference).

Plus a one-sentence hardening of the `<session-credentials>` block: when the user says they "added a key or credential", these names are where to look, and `list_variables` also lists them.

**Names only, everywhere** — no credential value reaches a system prompt, tool result, transcript entry, or rendered error. Redaction of credential values from tool results (`session.py` redact hook) is unchanged, and the block stays in the volatile tail (cache-layout decision untouched).

## Reproduction (real path, before → after)

**BEFORE — on clean `origin/main` (f7648577), after storing `OSWORLD_OPENAI_API_KEY`:**

```
list_variables: '0 variable(s) available:\n(no variables defined)'
read_variable (real name): 'unknown variable: OSWORLD_OPENAI_API_KEY (see list_variables)' is_error = True
journal records: []
```

All three dead-ends from the live failure, reproduced.

**AFTER — same steps on this branch (`e029ee5c`), driven through the real Session and real tool executors:**

```
=== step 1: operator runs /credential OSWORLD_OPENAI_API_KEY ===

=== step 2: model calls list_variables ===
0 variable(s) available:
(no variables defined)

session credentials (secret, usable via bash env, not readable): OSWORLD_OPENAI_API_KEY

=== step 3: model reads the guessed conventional name ===
unknown variable: OPENAI_API_KEY (see list_variables)

=== step 4: model reads the REAL name ===
SESSION CREDENTIAL: OSWORLD_OPENAI_API_KEY is a session credential. Its value is never readable. It is injected as an environment variable into every bash command — use it there (e.g. a child process reads $OSWORLD_OPENAI_API_KEY), never echo it.

=== step 5: next provider turn sees the announcement ===
[session credential] OSWORLD_OPENAI_API_KEY was just stored by the operator. Its value is held in session memory and injected as the environment variable $OSWORLD_OPENAI_API_KEY into every bash command — use it there (a child process reads it), never echo, print, or write it. It is not readable through read_variable.

=== step 6: announcement persisted for resume ===
persisted 1 session_credential entry(ies)

=== step 7: /credential --forget announces removal ===
[session credential] OSWORLD_OPENAI_API_KEY was removed. It is no longer available as an environment variable to bash commands in this session; do not reference it.

listing after forget no longer names it — OK

ALL REPRODUCTION CHECKS PASSED
```

(The repro script asserts the value string never appears in any of the above.)

## TUI worker-thread path (the trickiest wiring point)

`_credential_store_flow` runs in a Textual worker, off the turn loop. Verified by driving the REAL `OperatorApp` with `run_test`: submit `/credential OSWORLD_OPENAI_API_KEY`, type the secret into the real `KeyPromptBlock`, press enter, and assert the announcement landed in a REAL `Session`'s live context — then `--forget` through the real handler too:

```
store-flow journal record (from the TUI worker):
  [session credential] OSWORLD_OPENAI_API_KEY was just stored by the operator. Its value is held in session memory and injected as the environment variable $OSWORLD_OPENAI_API_KEY into every bash command — use it there (a child process reads it), never echo, print, or write it. It is not readable through read_variable.
forget-flow journal record:
  [session credential] OSWORLD_OPENAI_API_KEY was removed. It is no longer available as an environment variable to bash commands in this session; do not reference it.

TUI WORKER-PATH CHECKS PASSED
```

The sync call is safe off-loop because `Session.journal_credential_change` is deliberately sync (live-context append immediate; the transcript write rides `_spawn_background`), which is exactly why the TUI pass-through needs no thread hop.

## Subagent inheritance (new tests, all passing)

```
tests/unit/session/test_launch_subagent.py::test_child_system_blocks_advertise_parent_credentials PASSED
tests/unit/session/test_launch_subagent.py::test_child_bash_injects_parent_credential_env PASSED
tests/unit/session/test_launch_subagent.py::test_child_sees_a_credential_stored_after_its_spawn PASSED
```

The last one proves the store is shared by reference: a credential stored AFTER the child was built reaches the child's next `bash` turn without a respawn.

## Unit tests

New/extended (all passing):

```
tests/unit/tools/test_variables_tools.py  — 22 passed (5 new: credential names in listing, no-value guarantee,
    idle shape unchanged, read_variable bash-contract error, forget removes from listing)
tests/unit/session/test_session.py       — 2 new: store journaled + renders into the next provider turn +
    persisted for resume; forget journaled as removal
tests/unit/tui/test_credential_command.py — 6 passed (1 new: real-session journal record, key only)
tests/unit/tools/test_ask_tool.py        — 19 passed (2 new: secret answer announces the key; hook failure
    does not fail the ask)
tests/unit/test_prompts_api.py           — credentials-tail test extended for the discoverability sentence
tests/unit/session/test_launch_subagent.py — 3 new (above)
```

## Gates (whole tree, as CI runs)

```
uvx --from black==26.1.0 black --check .                       → 566 files unchanged
.venv/bin/python -m pyright --pythonpath .venv/bin/python local_operator tests → 0 errors
env -u NO_COLOR TERM=xterm-256color .venv/bin/python -m pytest tests/unit -q → 7329 passed, 24 failed, 7 skipped
```

**The 24 failures are pre-existing on `origin/main` and unrelated to this change** — all in `tests/unit/tools/test_eval_tool.py` (background-kernel spawn failures, machine-local). Verified by stashing this diff and running the same file on clean main:

```
# clean origin/main (git stash applied):
FAILED tests/unit/tools/test_eval_tool.py::test_background_streams_output_while_it_runs ...
24 failed, 23 passed, 3 warnings in 14.73s

# with this change applied:
24 failed, 23 passed, 3 warnings in 11.80s
```

Identical failure set and count. This diff touches nothing the eval tool reads (eval kernel, its session context, its ToolContext surface are untouched; the one ToolContext field added is a new optional `journal_credential` hook).

No TUI layout or interaction changes — the `/credential` flow keeps its exact verbs and prompts; only tool outputs, prompt text, and a journaled (non-visual) message changed, so no design round applies.


damianvtran:
### Agent review — round 1

Reviewer: reviewer subagent (pr360-review, model anthropic/claude-opus-4-8)
Scope: c6748028..e667e279

I read the full diff, traced every new credential path against the redaction machinery, ran all 13 new tests plus the surrounding files (11 + 4 + 6 passed locally), and reproduced one defect by executing the production handler sequence against a real `VariableStore`.

## Secret safety (priority 1) — clean

I traced every path a credential value could take and found no leak:

- `format_credential_message` (incidents.py) receives only the key; the announcement text is built from the key and constants alone.
- `list_variables` (builtin.py:4618) and `read_variable` (builtin.py:4657) read `credential_names()` — insertion-ordered keys only (`variables.py:344`), never values. The `read_variable` error text interpolates `params.name` (already a stored key name), not a value.
- The `ask` announce hook (builtin.py:8027-8040) fires only after a successful `store_credential`, passes the normalized key from `result.credential.key`, and wraps the call in `try/except` so a raising host hook cannot fail the ask.
- The loop's `redact_tool_result` choke point (session.py:4157, applied at loop.py:1832) is untouched and still receives every tool result; `_redact_tool_text` (builtin.py:1115) still covers live UI output. The new tool-result strings carry names only, so there is nothing to redact.
- On resume, the persisted `session_credential` entry replays via `Transcript.build_llm_history` → `_entry_to_message` → `CustomMessage.model_validate` and renders through `_default_convert_to_llm`, which allow-lists the type (session.py:414). The stored payload contains only key/action/replaced/text.

No path lets a value reach the prompt, a tool result, the transcript, or a log line. The subagent tests additionally assert `"sk-..." not in` the child's system blocks and bash output.

## Findings

**R1 — MAJOR (tui/app.py:12993-12997): `/credential --forget-all` never announces anything — the loop is dead.**
`clear_credentials()` empties the store, and only then does the handler iterate `store.credential_names()` (which is now empty) to journal removals. Reproduced against a real `VariableStore` with two stored keys: `count = 2`, announced keys = `[]`. The model is left believing both credentials still exist — the exact stale-belief failure mode the forget announcement was added to prevent, on one of the three verbs the PR claims to cover. Fix is one line: snapshot `names = store.credential_names()` *before* `clear_credentials()`. No test covers `--forget-all` announcement, which is how this shipped past the suite.

**R2 — MAJOR (session.py:4645-4689): persisted announcement is stale-by-design on resume, and nothing says so.**
Credentials live in process memory only (`VariableStore` is rebuilt empty by `_build_variable_store` on every start; nothing rehydrates them), yet `session_credential` records are persisted and replayed as user turns on resume. After a restart, a resumed session replays "[session credential] FOO_KEY was just stored by the operator … injected as the environment variable $FOO_KEY into every bash command" for a credential that no longer exists in the store — the model is told, in a user turn, that an env var is present when `bash` will not have it. The model-switch machinery this was modelled on has an explicit transient/persist split for exactly this class of problem (a fallback that outlives the process is stale by definition — see `_is_persistable_message`, review R1 of that round). Options: render the replayed record with a resume disclaimer, or gate persistence on the store still holding the key at replay time. Either way the current text actively misleads after a restart. The PR body markets resume replay as a feature ("a resumed session replays the announcement"); it is a liability, not a feature, until the staleness is handled.

**R3 — MINOR (builtin.py:4618, 4657): `hasattr(store, "credential_names")` guards are dead defensive code.**
`_variable_store` (builtin.py:4579) returns `context.variables` typed as `VariableStoreProtocol`, which declares `credential_names` (harness/types.py:388), or constructs a concrete `VariableStore`. The `hasattr` fallback masks a type violation rather than letting it surface, and the codebase's own convention (session_factory.py:933, subagent.py:1101) is the same defensive shape — so this is a pattern-echo, not a one-off. Consider dropping the guard or noting why the protocol is not trusted here.

**R4 — MINOR (tests/unit/tui/test_credential_command.py:83-115): the "real session journal record" test calls `journal_credential_change` directly on a bare `Session`; the TUI seam is untested.**
The test constructs a real `Session` and invokes the journal method itself — it proves `Session.journal_credential_change` works (already covered by test_session.py:1299) but never exercises `_cmd_credential` → `_credential_store_flow` → `app._journal_credential_change` → `getattr(self._session, ...)`. That last hop is the one that silently degrades to a no-op when `self._session` is a `FakeSession`, a `RemoteSession` (which has no `journal_credential_change` — attaching to a shared session then running `/credential` there skips the announcement), or a still-starting session. The PR body's "store-flow journal record (from the TUI worker)" evidence came from a manual script, not this test. A test that subclasses `FakeSession` with the hook and drives `/credential <KEY>` through the real paste flow would pin the actual seam.

**R5 — MINOR (harness/subagent.py:1032-1041): the construction-context `variables=` hunk is inert today; the comment overstates what it covers.**
I removed exactly this hunk and re-ran the three new subagent tests: all 3 pass without it. No `TOOL_BUILDERS` entry reads `context.variables` at construction time (checked every builder; the variables readers — builtin.py:1123/1242/8019 — are all execute-time, and the child `Session` receives `variables=` directly at session.py's `Session(...)` call, which is what `_build_tool_context` re-derives per turn). The hunk is harmless future-proofing, but the comment claims "the child's later turns read `context.variables` … so omitting it here left the child's construction-time tools reading a bare process-env store" — the construction-time context is discarded after `create_tools` and no gate reads it, so nothing was left reading a bare store. Keep the line if you want the parity with session_factory, but the comment should say "no current builder reads this; kept for parity/future createIf gates" rather than narrating a gap it does not close.

## Nit

**N1 — NIT (session.py:4668-4671 vs. built behaviour): docstring claims "resume replays the announcement, which is what tells a restored session which names exist" — with R2 unfixed the replay tells it names that no longer exist.** Reword once R2 is decided.

## What I verified beyond the diff

- `_append_or_park_journal` (session.py:4498) parks the message mid-turn exactly as the PR claims; the flush happens at every prompt/wake/steering/dispose boundary, so a store parked mid-turn reaches the next turn. Ordering vs a concurrent turn is safe.
- Transcript append is idempotent by message id (`_persist_new_messages`), so the direct `append_message` + later turn-boundary persist cannot double-write.
- `journal_credential_change` re-announcing the same key on re-store is correct behaviour (the model should learn the value changed); no dedup is needed, unlike model-switch where two edges fire for one change.
- Re-storing the same key announces "replaced" via `result.replaced` (tui) and `result.replaced` (ask) — both read the real flag.
- The reviewer role seed (`agent_seeds/reviewer.md`) allows `list_variables`/`read_variable`/`bash`, so a restricted child gets the new listing and the credential env; the child `Session` receives `variables=` directly, so bash injection works regardless of R5.
- All 13 new tests pass locally, and the full new-test files (21 tests) pass.

## Verdict

**Two MAJOR findings (R1, R2) — not clean.** R1 is a reproduced dead loop on a shipped verb; R2 is a stale-announcement-on-resume defect that the PR currently advertises as a feature. Both need a remediation round.

damianvtran:

Implementer: coder subagent (model anthropic/claude-opus-4-8)
Scope: e667e279..f70089a8 (this commit)

---

**R1 — `/credential --forget-all` never announces anything — FIXED (f70089a8).**
Exactly the reviewer's one-line diagnosis: the handler iterated `credential_names()` *after* `clear_credentials()` emptied the store. The handler now snapshots `names = store.credential_names()` before the clear and journals each removal from the snapshot (tui/app.py, `--forget-all` branch, with a comment naming the review finding so the ordering cannot be "simplified" back).

Test: `test_forget_all_announces_every_removal_to_the_journal` (tests/unit/tui/test_credential_command.py) stages two stored keys, runs `/credential --forget-all` through the real pilot, and asserts the journal hook received `("GITHUB_TOKEN", "forgot")` and `("ANTHROPIC_API_KEY", "forgot")` plus the "Forgot 2 credentials" notice. It drives the TUI seam through a `FakeSession` subclass exposing `journal_credential_change` — the same `getattr` hop the production handler uses (this also pins the seam R4 flagged, as a side effect of writing the test the way the finding said it should be written). Verified the test catches the bug: with the fix stashed it fails (`credential_journal == []`), with the fix it passes.

**R2 — persisted announcement is stale-by-design on resume — FIXED by option (a), live-context-only (f70089a8).**
Took the reviewer's "likely cleaner" path, for the reasons the reviewer gave plus one discovered while implementing:

- `SESSION_CREDENTIAL_MESSAGE_TYPE` is removed from `_PERSISTABLE_CUSTOM_TYPES`, and `journal_credential_change` no longer spawns the background transcript write — the record lands in the live context only (session.py). The mid-session discovery problem — the actual failure this PR fixes (session 835fbcafdc27) — is fully served by the live append; resume-time discovery is served by the `<session-credentials>` prompt-tail block, which is rebuilt from the live (empty-on-restart) store every turn and so correctly says nothing after a restart.
- **The part the reviewer's two options did not cover:** compaction would have re-persisted the record anyway. The kept window is rebuilt from `_render_for_compaction`, whose default render converts the live `CustomMessage` into a plain `Message(role="user")` **carrying the same id** — and the turn-end persist pass writes every plain Message. A credential record sits at the recent end of the context, exactly where the kept window is, so dropping the type from the allow-list alone would not have closed the leak: the next compaction pass would have baked the record into `kept` and persisted it under its original id. `_render_for_compaction` therefore also filters `session_credential` records, the same exclusion todo reminders already get. The request render (`_render_history`) is untouched, so the live announcement still reaches the model mid-session.

Mirrors the transient model-switch split as instructed: not persisted, still rendered live (`_default_convert_to_llm` allow-list entry unchanged).

Tests (tests/unit/session/test_session.py):
- `test_credential_announcement_does_not_replay_on_resume` — stores the announcement, prompts, proves the live request carried it, then reopens the transcript the way `--resume` does and asserts no `[session credential]` text replays while the real user turn does.
- `test_credential_announcement_survives_into_compaction_kept_window_only_as_live` — asserts the request render carries the announcement and `_render_for_compaction` excludes it.
- `test_credential_change_is_journaled_and_model_visible` updated: its old persistence assertion is inverted (transcript must NOT contain the type; `_is_persistable_message` is False).

N1 rides along: the docstrings/comments that marketed resume replay as a feature (incidents.py type comment, `journal_credential_change`, the TUI store-flow comment, the `ask` announce comment) now state the live-only contract and why.

**R5 — inert construction-context `variables=` hunk — FIXED by keeping the line with an honest comment (f70089a8).**
Kept, not removed: the line gives the subagent construction context the same shape the session factory builds (one store handed to both the construction and execution contexts), which is the established convention in this codebase. The comment no longer narrates a gap it does not close — it now says plainly: no current builder reads `context.variables` at construction time, the child `Session` receives `variables=` directly and `_build_tool_context` re-derives it per turn, and the line is defensive parity for any future createIf gate.

---

**R3 / R4 — not assigned for remediation (minor, outside this round's scope).** Noting for the record: R3 (dead `hasattr` guards) is rejected as a pattern-echo the reviewer itself identified as the codebase convention (session_factory.py, subagent.py use the same shape); R4's seam concern is incidentally addressed by the new R1 test driving `_cmd_credential → _journal_credential_change → getattr(session, ...)` end-to-end, though the dedicated `RemoteSession`-shaped case remains uncovered.

**Gates on f70089a8:** flake8 clean, black 26.1.0 clean, isort 5.13.2 clean, pyright clean on all seven touched files (`0 errors, 0 warnings`); full unit suite 7333 passed with 26 failures I verified pre-exist on the base commit (eval_tool's spawned kernel cannot import `local_operator` from the worktree — `ModuleNotFoundError` in the kernel subprocess, environmental to this worktree, reproduced with the diff stashed). Full-tree local pyright run in progress; CI on the new head will independently verify.

damianvtran:
### Agent review — round 2

Reviewer: reviewer subagent (pr360-review-r2, model anthropic/claude-opus-4-8)
Scope: e667e279..f70089a8

Focused verification round on the remediation commit, plus a fresh-eyes pass over the full final diff (c6748028..f70089a8). I read the remediation diff first, then opened the surrounding code only where the diff made claims I could test: the compaction kept-window rebuild, every `_transcript.append_message` call site, the loop's run accumulator, and both turn-end persist passes.

## Round-1 majors — both RESOLVED, verified on f70089a8

**R1 (forget-all dead loop) — resolved.** `tui/app.py` `--forget-all` branch now snapshots `names = store.credential_names()` before `store.clear_credentials()` and iterates the snapshot; the notice still uses the clear's own count. Ordering is correct, and `credential_names()` is insertion-ordered (`variables.py:344-346`), so journal order is deterministic. The new test `test_forget_all_announces_every_removal_to_the_journal` drives the real TUI seam — `_submit(pilot, app, "/credential --forget-all")` → `_cmd_credential` → `_journal_credential_change` → `getattr(self._session, "journal_credential_change")` — through a `FakeSession` subclass exposing exactly the hook the production handler looks up, and asserts the full journal sequence plus the "Forgot 2 credentials" notice. Fail-first is credible by inspection: on the unfixed code the loop iterates an already-empty store, `credential_journal` stays `[]`, and the equality assertion fails. (This also pins the seam R4 asked for, as the remediation notes.)

**R2 (stale resume replay) — resolved, and the compaction mechanism claim is CONFIRMED, not just plausible.** Verified piece by piece:

- (a) `SESSION_CREDENTIAL_MESSAGE_TYPE` is absent from `_PERSISTABLE_CUSTOM_TYPES` (session.py:519-533, with an honest comment), so `_is_persistable_message` returns False for the record.
- (b) `journal_credential_change` (session.py:4683-4746) no longer spawns the transcript write — the `_spawn_background(self._transcript.append_message(...))` line is gone; only `_append_or_park_journal(message)` remains.
- (c) The compaction exclusion is real AND necessary. I traced the exact path the implementer described: `_run_compaction` rebuilds `self._context.messages = [marker, *preserved, *kept]` from `_render_for_compaction(keep_images=True)[plan.cut:]` (session.py:5485, 5559), and `_default_convert_to_llm` renders the live `CustomMessage` as a plain `Message(role="user")` **carrying the same id** (session.py:427-433). From there BOTH turn-end passes would write it, because both iterate the live context as plain Messages with no type gate left: the `finally`-path `self._persist_progress(self._context.messages)` (session.py:4326) and the mid-turn gate's `_persist_progress(messages)` over `list(context.messages)` (loop.py:681 → session.py:5082). So without the `_is_ephemeral_for_compaction` filter (session.py:1924-1944), the very next compaction pass would have re-persisted the record under its original id and resurrected the stale-on-resume replay the allow-list removal closed. The claim is accurate, and the fix also removes a latent `cut_not_replayable` refusal: a credential record sitting at the cut position carries an id in no transcript, the same 30/30-refusal failure mode todo reminders had. Side benefit I confirmed: excluding it from the compaction render also keeps the summarizer from paraphrasing the announcement into the PERSISTED summary text, which would have been a second, harder-to-spot leak of the same claim.
- (d) The live render path is untouched: `_default_convert_to_llm` still allow-lists the type (session.py:411-433) and `_render_history` (which serves the turn path, `_wire_legal_snapshot`, and the token counter) does not filter it — only `_render_for_compaction` does. Mid-session announcement still reaches the provider, pinned by the new test asserting the REQUEST render carries it while the compaction render does not.
- (e) No other path persists the record. I enumerated all `_transcript.append_message` call sites (session.py:3298, 3496, 4132, 4614, 4678, 4847, 4907, 5978, 6027) — each is seed/steering/aside/shell/incident/model-switch scoped and none can receive a credential record; the `append_custom` sites are other types. The loop's run accumulator (`new_messages`, loop.py:477) never contains it (it enters context via `_flush_context_journal`, not the loop), and the two `_persist_progress` paths that DO see it are gated by `_is_persistable_message`. `_persist_new_messages` (session.py:5042) skips it. Closed.

**R5 (overclaiming comment) — resolved.** The rewritten comment in `harness/subagent.py` now says plainly "DEFENSIVE PARITY, not a bug fix", names that no current builder reads `context.variables` at construction time, and explains what the line is kept for. It no longer narrates a gap it does not close.

## Fresh-eyes pass over c6748028..f70089a8

Nothing new of blocker/major weight. Specifics I checked:

- The new compaction exclusion interacts with the replayability guard, the kept-window rebuild, the `history_after` estimate (session.py:5561), and `extract_preserved_user_turns` (the record's id is not in `genuine_user_ids` and its rendered copy is no longer in `to_summarize`) — consistent in all four; no path bakes it in.
- The TUI snapshot list is correct as written (see R1 above); the empty-store edge journals nothing and the notice reports the clear's count, which is 0.
- The `ask` announce hop (builtin.py:8021-8046) fires only on a successful store, passes the normalized key, and a raising hook cannot fail the ask.
- The new prompt-tail sentence "…`list_variables` also lists them" is true: `execute_list_variables` appends the credential names section, and `read_variable`'s credential branch no longer gives circular advice. No value reaches either string.
- Round-1 R3 rejection (dead `hasattr` guards as pattern-echo) — accepted; the convention is real in session_factory/subagent. R4's residual `RemoteSession`-shaped gap (a session object without the hook silently skips the announcement) remains uncovered by a test; the degradation is silent-but-safe, and I consider it a follow-up, not a remediation item.

## Tests run (worktree /tmp/lop-credential-visibility, f70089a8)

- `env -u NO_COLOR TERM=xterm-256color .venv/bin/python -m pytest tests/unit/tui/test_credential_command.py -q` → 7 passed (includes the new R1 test).
- `... tests/unit/session/test_session.py -k credential -q` → 4 passed (the two new R2 tests, the inverted persistence test, the forget-announcement test).
- `... tests/unit/session/test_session.py tests/unit/tui/test_credential_command.py tests/unit/tools/test_ask_tool.py tests/unit/tools/test_variables_tools.py tests/unit/test_prompts_api.py -q` → 183 passed.
- `... tests/unit/session/test_launch_subagent.py -q` → 45 passed.
- `... tests/unit/session/test_session.py -k compact -q` → 12 passed (the pinned kept-window/resume equivalences hold with the new exclusion).
- CI on f70089a8: lint, type-check, test 3.12/3.13, pip-audit, cli-sanity, server-sanity all SUCCESS; PR head is f70089a8, so this round is fresh against the current head.

## Findings

**R2-1 — MINOR (local_operator/incidents.py:218-219; tests/unit/session/test_session.py:1326): two comments still claim the announcement text is written to the transcript.** `format_credential_message`'s docstring says "this text is journaled to the transcript and sent to the provider", and the test comment at test_session.py:1326 says "(this text is sent to the provider and written to the transcript)". Both are now false — the record is live-context-only, which is the entire point of the R2 fix. The remediation updated the type-constant comment, the journal method docstring, the TUI store-flow comment, and the ask comment, but missed these two. This is the same class as round-1 N1: a reader auditing secret-handling will read "journaled to the transcript" and hunt for a write that no longer exists. One-line reword each. Fix in a follow-up; do not churn the PR for it.

**R2-2 — MINOR (follow-up, no code change requested): the `RemoteSession`-shaped seam gap from R4 remains untested.** A session object without `journal_credential_change` silently skips the announcement (store still works, bash injection still works). Accepted as a follow-up along with round-1 R3's rejection rationale.

## Verdict

**clean — no BLOCKER and no MAJOR remains on f70089a8.** Both round-1 majors (R1, R2) are resolved and verified, R5's comment is honest, the R2 compaction mechanism is confirmed against the actual persist-pass code rather than taken on faith, and the fresh-eyes pass found nothing new above minor. This round is TERMINAL: R2-1 and R2-2 are follow-ups, not grounds for another round.

PR #363, opened 2026-08-28
fix(browser): stop the stale approval-popup loop on replaced requests

damianvtran:
# PR Description

## Summary of Changes

Fix the "Request changed." approval-popup loop in the browser extension. A user
reported the same site request (`docs.gominerva.com`) appearing twice; approving
the first popup left the second looping on the "Request changed. The site request
was replaced while this window was open. Review the new request." interstitial —
every Allow click re-showed the interstitial and the request never resolved.

Four cooperating defects produced the loop; this PR fixes all of them.

### Diagnosis (verified against `origin/main`)

1. **Health-fallback renders carry no generation id.** `popup.ts` `render()`
   computes `shownPromptId = selected?.entryId ?? session.pendingOrigin?.promptId
   ?? ""`. When the popup renders from the daemon `/health` fallback (queue
   storage empty/unread, worker restarted, or the queue entry expired while the
   daemon still echoes it), `shownPromptId` is `""`.

2. **Empty generation = dead buttons.** `origins.ts` `resolveOrigin()` returned
   `false` for any decision with a missing/stale `entryId`. A health-fallback
   render therefore rejected every click; `popup.ts` `decide()` showed the
   "Request changed." interstitial, waited 1.5 s, re-rendered the SAME
   health-echoed prompt, and looped until the daemon stopped echoing.

3. **The daemon echo outlives the queue entry.** `daemon.py` sets
   `awaiting_origin` from the extension's `awaiting_origin` event and pops it
   only when that RPC's response arrives or the RPC times out
   (`ORIGIN_PROMPT_WINDOW_S` ≈ 10 min + 5 s). Nothing reconciles it when the
   extension-side queue entry expires/cancels while the worker is suspended or
   the response is lost, so `/health` `pending_origin` kept echoing a prompt the
   extension no longer held — feeding loop (2).

4. **Duplicate requests stack.** Two `in_command` entries for the same
   origin+requester+commandId (a retried navigation) enqueued as two separate
   generations (`enqueueAccess` deduped only `kind === "async"`), so the user saw
   "the same request twice" and approving one left the twin.

### Fixes

- **A — Origin-fallback decision (`origins.ts`).** When the generation id is
  missing or stale, `resolveOrigin` now applies the decision to the single live
  queue entry whose origin exactly matches the clicked origin. Rationale: the
  user is looking at the exact origin rendered; the generation mismatch came from
  a fallback render, not a replaced prompt. Zero or several matches still reject,
  so round-2 B1's consent hole stays closed (never apply by origin alone when
  ambiguous).

- **B — Daemon reconciliation (`worker.ts`, `daemon.py`, `protocol.py`,
  `protocol.gen.ts`).** The worker observer tracks announced command ids and now
  emits a new `awaiting_origin_cleared` event for ids that leave the queue. The
  daemon pops its record and republishes, so `/health` stops echoing a prompt the
  popup can no longer resolve. This bounds the stale echo to the queue's
  lifetime. The frame is a typed `AwaitingOriginCleared` model and
  `protocol.gen.ts` is regenerated via `gen_ts.py`.

- **C — Popup notice selection (`popup.ts`, `origin-flow.ts`).** When `decide()`
  gets `applied: false` and the shown prompt id was empty (health-fallback
  render), it no longer shows "Request changed." — the request was not replaced;
  retry is handled by fix A. A new pure `noticeForRejectedDecision()` returns the
  "Request changed." copy only when the origin is still pending under a NEWER
  generation (a true replacement), and a new "Request expired." copy when the
  live queue no longer holds the origin. The 1.5 s hold behavior is preserved.

- **D — In-command dedupe (`approval-store.ts`).** `enqueueAccess` now dedupes
  `kind === "in_command"` on origin+requester+commandId (a retried navigation
  re-asks the identical question). Different commandIds and two paused documents
  under different commands keep separate generations. A deduped generation carries
  every attached waiter (`origins.ts` `waiting` is now a multimap), so one Allow
  resumes every paused document on that command — exactly the consent given,
  never a broader grant.

## Related Issue(s)

- User-reported approval-popup loop (no ticket).

## Impact

- No breaking changes. The wire protocol gains one extension→daemon event
  (`awaiting_origin_cleared`); older daemons ignore unknown frames, and the
  daemon's existing response/timeout pops remain as backstops.
- No dependency updates.
- Security: the B1 consent hole stays closed — origin-fallback applies only on an
  unambiguous single match, and dedupe never broadens a grant.

### Extension (node) — `pnpm typecheck && pnpm test && pnpm build`

- `origin-fallback decision applies to a single live entry and rejects ambiguity`
  — single match applies, multi-match rejects, zero-match rejects (fix A).
- `in-command enqueue dedupes a retried same-command navigation but keeps
  distinct commands` — same command dedupes, different commandId keeps its own
  generation (fix D).
- `a deduped same-command generation resolves every attached waiter` — one
  decision settles both paused documents (fix D multimap).
- `worker observer announces an origin block and clears it when the entry leaves
  the queue` — clearance emitted after the entry leaves, not on first
  announcement (fix B).
- `rejected decision notice skips the interstitial for fallback renders` — pure
  `noticeForRejectedDecision` selection (fix C).
- Three pre-existing tests that codified the OLD no-dedupe rule
  (`identical in-command admissions are independently visible and decidable`,
  `always allow reconciles every identical in-command waiter`, `TTL sweep settles
  every identical in-command waiter`) were reconciled to the new dedupe behavior;
  their `while length !== 2` loops would otherwise spin forever.

Result: **67/67 pass**, `tsc --noEmit` clean, `build` clean.

### Python — `tests/unit/browser_bridge`

- `test_awaiting_origin_cleared_stops_the_health_echo` — the daemon pops the
  record on the cleared event and `/health` `pending_origin` empties; a clearance
  for an unknown id is a no-op (fix B).
- `test_awaiting_origin_cleared_round_trips_through_the_wire_model` — protocol
  model round-trip.
- `test_generated_types_are_current` — `gen_ts --check` parity.

Result: **94/94 pass** in `tests/unit/browser_bridge`. Full `tests/unit` suite:
7337 passed; the 24 pre-existing failures (`test_eval_tool.py`,
`test_subagent_view.py`) reproduce identically on clean `origin/main` and are
unrelated to this change.

### Gates

flake8, black (26.1.0), isort (5.13.2), and pyright all clean on the touched
files.

- Approving a site request rendered from the `/health` fallback now resolves the
  request instead of looping on "Request changed."
- A retried navigation no longer stacks a duplicate request; one Allow settles
  every paused document on that command.
- When a request is genuinely replaced, the popup still says "Request changed.";
  when it expired or was cancelled, it now says "Request expired."

## Checklist

### DIFF
--- a/extension/src/approval-store.ts
+++ b/extension/src/approval-store.ts
@@ -196,12 +196,33 @@ export function enqueueAccess(
   return withSessionMutation(async () => {
     const now = Date.now();
     const snapshot = await normalizedLocked(now);
-    // Only narrated async requests deduplicate. Two paused documents under one
-    // command carry the same requester/origin/kind/commandId but represent two
-    // independently blocked navigations. Sharing one generation would either
-    // overwrite a waiter or make one Allow once authorize both.
-    const existing = kind === "async" ? findPending(snapshot.queue, url.origin, requester, kind, now) : undefined;
+    // Async requests deduplicate on origin+requester: one narrated question is
+    // asked once. In-command requests deduplicate ONLY on
+    // origin+requester+commandId — a retried navigation re-asks the identical
+    // question and must not stack a twin the user has to answer twice
+    // (approving one used to leave the other looping the popup on
+    // "Request changed."). Two paused documents under DIFFERENT commands keep
+    // separate generations so each command resumes on its own consent; a
+    // same-command twin shares the one generation the user is looking at, and
+    // one Allow resumes every waiter attached to it — exactly the consent
+    // given, never a broader grant.
+    const existing =
+      kind === "async"
+        ? findPending(snapshot.queue, url.origin, requester, kind, now)
+        : snapshot.queue.find(
+            (entry) =>
+              entry.kind === "in_command" &&
+              entry.origin === url.origin &&
+              entry.requester === requester &&
+              !!commandId &&
+              entry.commandId === commandId,
+          );
     if (existing) {
+      // A deduplicated generation must carry the new caller's waiter too:
+      // onAdmitted otherwise fires only on first admission, and an in-command
+      // caller whose waiter never registers would await a promise nothing can
+      // resolve (the paused document frozen forever).
+      onAdmitted?.(existing);
       await armNextExpiry(snapshot);
       return {
         origin: url.origin,
--- a/extension/src/origins.ts
+++ b/extension/src/origins.ts
@@ -1,5 +1,5 @@
 import { reconcileActionSurface } from "./action-surface";
-import { requesterOriginKey, type AccessQueueEntry, type OriginDecision } from "./access-queue";
+import { liveQueue, requesterOriginKey, type AccessQueueEntry, type OriginDecision } from "./access-queue";
 import {
   cancelAccess,
   decideAccess,
@@ -17,16 +17,21 @@ export type { OriginDecision } from "./access-queue";
 
 // In-command continuations are keyed by immutable queue generation, never by
 // origin: two commands paused on the same redirect must resume independently.
-const waiting = new Map<string, (decision: OriginDecision) => void>();
+// Each generation can hold SEVERAL resolvers: enqueueAccess dedupes a retried
+// same-command navigation onto the existing entry (approval-store), and every
+// paused document waiting on that entry registered its own resolver here — a
+// single-slot map would overwrite the first and strand its navigation.
+const waiting = new Map<string, Set<(decision: OriginDecision) => void>>();
 let onPendingChange: ((snapshot: QueueSnapshot) => void) | null = null;
 
 function reconcileWaiters(snapshot: QueueSnapshot): void {
-  for (const [entryId, resolve] of waiting) {
+  for (const [entryId, resolvers] of waiting) {
     if (snapshot.queue.some((entry) => entry.entryId === entryId)) continue;
     const receipt = Object.values(snapshot.results).find((candidate) => candidate.entryId === entryId);
     if (!receipt) continue;
     waiting.delete(entryId);
-    resolve(receipt.state === "allowed" ? "always" : "deny");
+    const decision: OriginDecision = receipt.state === "allowed" ? "always" : "deny";
+    for (const resolve of resolvers) resolve(decision);
   }
 }
 
@@ -124,8 +129,12 @@ export async function askOrigin(url: URL, requestId: string): Promise<boolean> {
   const queued = await enqueueAccess(url, requestId, "in_command", requestId, (entry) => {
     // enqueueAccess invokes this callback under the same mutation lock that
     // publishes the entry. No decision or expiry sweep can interleave between
-    // visibility and registration.
-    waiting.set(entry.entryId, resolveWaiter);
+    // visibility and registration. The callback also fires for a DEDUPED
+    // enqueue (retried same-command navigation), so several resolvers can
+    // attach to one generation — see the `waiting` map's comment.
+    const resolvers = waiting.get(entry.entryId) ?? new Set<(decision: OriginDecision) => void>();
+    resolvers.add(resolveWaiter);
+    waiting.set(entry.entryId, resolvers);
   });
   if (!queued.entry) return false;
   await updatePromptSurfaces(await sweepQueue());
@@ -133,12 +142,30 @@ export async function askOrigin(url: URL, requestId: string): Promise<boolean> {
 }
 
 export async function resolveOrigin(
-  _origin: string,
+  origin: string,
   decision: OriginDecision,
   entryId: string = "",
 ): Promise<boolean> {
-  if (!entryId || !["once", "always", "all_ports", "deny"].includes(decision)) return false;
-  const result = await decideAccess(entryId, decision);
+  if (!["once", "always", "all_ports", "deny"].includes(decision)) return false;
+  let targetId = entryId;
+  const session = await getSession();
+  const queue = liveQueue(session.accessQueue, Date.now());
+  if (!targetId || !queue.some((entry) => entry.entryId === targetId)) {
+    // Generation miss: a /health-fallback render carries no entry id (queue
+    // storage empty after a worker restart, or the entry expired while the
+    // daemon still echoes it), and a stale id means the entry was replaced.
+    // The click still names the ORIGIN the user looked at, so honour it
+    // against that origin's single live entry — the mismatch came from a
+    // fallback render, not a replaced prompt. Zero or several matches stay
+    // REJECTED: with several live entries for the origin the click cannot say
+    // which paused navigation the user meant, so round-2 B1's consent hole
+    // stays closed.
+    const match = queue.filter((entry) => entry.origin === origin);
+    if (match.length !== 1) return false;
+    targetId = match[0]?.entryId ?? "";
+    if (!targetId) return false;
+  }
+  const result = await decideAccess(targetId, decision);
   if (!result.applied) return false;
   // decideAccess persisted receipts before releasing the mutation lock; the
   // queue observer resolves every matching waiter from those receipts.
--- a/extension/src/origins.ts
+++ b/extension/src/origins.ts
@@ -162,7 +162,9 @@ export async function resolveOrigin(
     // stays closed.
     const match = queue.filter((entry) => entry.origin === origin);
     if (match.length !== 1) return false;
-    targetId = match[0]?.entryId ?? "";
+    // length===1 guarantees the element; no optional-chain dead guard.
+    const single = match[0]!;
+    targetId = single.entryId;
     if (!targetId) return false;
   }
   const result = await decideAccess(targetId, decision);
--- a/extension/src/popup/origin-flow.ts
+++ b/extension/src/popup/origin-flow.ts
@@ -111,8 +111,11 @@ export function noticeForRejectedDecision(
       sub: "The site request was replaced while this window was open. Review the new request.",
     };
   }
+  // D1/D2 (design round 1): state the CONSEQUENCE (nothing was granted or
+  // denied) instead of restating the title's cause, so a user who just
+  // clicked Allow knows the click had no effect.
   return {
     title: "Request expired.",
-    sub: "This request expired or was cancelled.",
+    sub: "It timed out or was cancelled, so nothing was granted or denied.",
   };
 }
--- a/extension/src/popup/origin-flow.ts
+++ b/extension/src/popup/origin-flow.ts
@@ -80,3 +80,39 @@ export function originPromptView(
   if (!pendingOrigin) return "none";
   return decided && decided.origin === pendingOrigin ? "ack" : "prompt";
 }
+
+export interface OriginNotice {
+  title: string;
+  sub: string;
+}
+
+/** The notice a REJECTED decision shows, or null when none should appear.
+ *
+ * A rejection with an EMPTY shown prompt id came from a /health-fallback
+ * render (worker restarted, queue storage empty, or the entry expired while
+ * the daemon still echoed it) — the request was never "replaced", the popup
+ * simply had no generation to aim at. resolveOrigin's origin fallback and the
+ * daemon's cleared-event reconciliation retry or retire that render, so a
+ * scary "Request changed." interstitial there was pure noise — and looping it
+ * every click was the reported bug.
+ *
+ * A rejection WITH a shown id is a real miss: the generation the buttons were
+ * drawn for is gone. If the origin is still pending under a NEWER generation
+ * the prompt was replaced; if the live queue holds no entry for the origin at
+ * all, the request expired or was cancelled. */
+export function noticeForRejectedDecision(
+  shownPromptId: string,
+  originStillPending: boolean,
+): OriginNotice | null {
+  if (!shownPromptId) return null;
+  if (originStillPending) {
+    return {
+      title: "Request changed.",
+      sub: "The site request was replaced while this window was open. Review the new request.",
+    };
+  }
+  return {
+    title: "Request expired.",
+    sub: "This request expired or was cancelled.",
+  };
+}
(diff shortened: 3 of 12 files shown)


damianvtran:
### Design review — round 1

Reviewer: designer (openrouter/qwen/qwen3.8-max)
Scope: origin/main..430ae6b0

**Method note.** No rendered frames were available for this round, so per the delegation note this is a source + CSS proxy review. Mitigating factor: both notices render into the already-shipping `origin-ack` slot verbatim (same `.title` / `.sub` elements, check glyph hidden, tone overridden to neutral) — the only delta this PR introduces is text content, so layout risk is bounded. Geometry below is computed from the tokens in popup.css: content width is 246px (300px body − 2×10px card margin − 2×1px border − 2×16px body padding).

**What I reviewed.** The three rejection paths in `decide()` (popup.ts:399–455), `noticeForRejectedDecision` (origin-flow.ts:103–118), the ack/notice slot markup (popup.html:113–129), the tone map and typography (popup.ts:82–92, popup.css).

**Path-by-path.**

1. **Generation-less fallback render → no interstitial.** The question delegated to me was whether this leaves the user with no feedback that their click did something. It does not: `showOriginAck` fires synchronously from the click, *before* the worker round-trip (popup.ts:419–420), so every click is visibly acknowledged; on rejection the `decidedOrigin` latch is cleared (popup.ts:428) *before* `render()`, so the ack is not held over a live prompt and a future prompt for the same origin is not swallowed; `setOriginBusy(false)` re-enables the buttons before the re-render. The user sees: ack flash → settled state (the current prompt with live buttons, or Connected once the daemon's cleared-event reconciliation lands). The reported loop is broken without introducing a dead click. ✓
2. **True replacement.** Copy unchanged from main and still accurate: after the 1500ms hold, `render()` draws the replacement prompt with live buttons, which is exactly what "Review the new request." promises. ✓
3. **Expired/cancelled shown request.** New copy — see D1/D2.

**Design-system conformance.** Both notices sit in the serif-title + sans-sub slot, take the neutral `--hairline-strong` top rule, hide the check glyph, and spend no colour — correct: success is reserved for grants, danger for states the user must recover from, and an expired/replaced request is informational, not a failure the user must fix. Titles are short period-terminated statements matching the house voice ("Site allowed.", "Paired.", "Not connected."). "cancelled" matches the codebase's consistent spelling (access-queue.ts, approval-store.ts, daemon.py, protocol.py). 300px fit: "Request expired." is one line at 19px serif; the 39-character sub wraps to ≤2 lines at 13px/1.5 over 246px; the unchanged replacement sub (84 chars) is ≈3 lines as shipped. The `role="status"` on the ack section gets both notices announced politely by screen readers. ✓

**Findings.**

- **D1 — MINOR.** The expired notice states the *cause* but not the *consequence*, for a user whose last action was a consent click. "This request expired or was cancelled." leaves a non-technical user guessing whether their Allow/Deny still took effect — the neutral tone and absent checkmark imply "no", but the copy should say it. Suggest a decision-agnostic consequence clause, e.g. sub → "This request expired or was cancelled, so nothing was granted." (Works for a Deny click too: nothing was granted either way.)
- **D2 — NIT.** Title/sub redundancy in the expired notice: the sub's first clause restates the title ("Request expired." / "**This request expired**…"). The house pattern has the sub add information, not repeat ("Paired." / "Code accepted. Connecting…"). Folding in D1's consequence clause fixes this naturally.
- **D3 — NIT.** Record-only, no action required: on the no-interstitial path the user briefly sees the decision's own ack ("Site allowed. / The agent is continuing.") for a decision that was in fact not applied, for the duration of the worker round-trip. This is inherent to the ack-from-click latch (which exists to fix the worse no-feedback race), and the immediately following render corrects it — noting it only so a future reader doesn't mistake it for an oversight.
- **D4 — NIT.** Pre-existing, not introduced here: if the *replacement* entry expires during the 1500ms notice hold, the post-hold render settles to Connected or the next queue entry while the notice said "Review the new request." Self-resolving on the next render; mentioned only because I was asked to check consecutive frames.

**Verdict: TERMINAL.** No BLOCKER, no MAJOR findings. D1 is recommended as a cheap copy-only follow-up (foldable into this PR if the implementer agrees, but it does not gate); D2–D4 are follow-ups/notes.

damianvtran:
### Agent review — round 1

Reviewer: reviewer subagent (qwen3.8-max)
Scope: origin/main..430ae6b0

Independent review of the stale approval-popup loop fix. I read the full diff, then opened `origins.ts`, `approval-store.ts`, `access-queue.ts`, `worker.ts`, and `daemon.py` only where the diff couldn't answer a specific question. I ran the gates locally and checked CI.

**Verification of the called-out risk points**

1. **Fix A requester/consent boundary — clean.** The origin fallback (`origins.ts:163-165`) matches on `entry.origin === origin` only, *not* requester, and that is correct here. The decision targets the single matched `entryId`, and `decideAccess` re-validates that exact `entryId` under the session mutation lock — if the entry was replaced between the match and the lock, `selected` is undefined and the decision is rejected, never redirected onto a different entry. The ambiguous case (≥2 live entries sharing the origin, i.e. different requesters/commands) is rejected outright (`match.length !== 1`), so round-2 B1 stays closed. Requester is deliberately never user-visible ("Authority boundary only. Never render…", `access-queue.ts:27`), consent is origin-scoped, and the `once` grant is still keyed to the decided entry's own requester inside `decideAccess` — no cross-requester grant leak. The TOCTOU between the unlocked `liveQueue` read and the locked `decideAccess` fails safe (reject). Covered by the new `origin-fallback decision applies to a single live entry and rejects ambiguity` test (single/twin/zero cases).

2. **Multimap waiter resolution — clean.** `waiting` is now `Map<string, Set<resolver>>`; `reconcileWaiters` resolves every attached resolver from the receipt, and the dedupe path calls `onAdmitted(existing)` so a retried same-command caller registers its resolver on the shared generation (`approval-store.ts:225`). Each `askOrigin` closure is a distinct reference, so the Set can't accidentally collapse two waiters. Deny/allow/always/expiry all covered by tests (`[false,false]`, `[true,true]`, `[true,true]`, `[false,false]`).

3. **Cleared-event idempotence + restart — clean, one bounded edge (see M1).** Daemon pops only when the id is present (unknown-id no-op is tested). Worker deletes the id from `announcedAwaitingIds` before sending, so one clearance per id per worker lifetime. On restart both sides reset symmetrically: the daemon's `link.disconnect()` clears `awaiting_origin` (`daemon.py:147`) and the module-level announced set resets, so a restarted worker re-announces live entries on its first snapshot and a stale record can't clear a live announcement.

4. **Protocol gen parity — clean.** `protocol.py`, `gen_ts.py`, and `protocol.gen.ts` all add `AwaitingOriginCleared` consistently; `test_protocol.py` runs `gen_ts --check` and it passes in the suite below.

5. **decideAccess consent path — unchanged.** The diff touches only `resolveOrigin`'s caller logic and `enqueueAccess` dedupe; `decideAccess`'s `once`/`always`/`all_ports` grant semantics are untouched, and in-command `once` still spends without minting a stored grant.

6. **Comments + commit — clean.** Conventional commit (`fix(browser): …`), and the new comments explain the why/constraints (the dedupe consent rationale, the multimap reason, the clearance lifecycle) rather than restating the code.

7. **Gates (run locally on head 430ae6b0):** extension `pnpm typecheck` clean; `pnpm test` 67/67 pass; `pnpm build` clean; Python `tests/unit/browser_bridge/` 94/94 pass (incl. the two new cleared-event tests and the gen parity check). CI rollup on this head is fully green (build, lint, type-check, test 3.12/3.13, pip-audit, cli-sanity, server-sanity); branch MERGEABLE.

**Findings**

- **MINOR (M1), no action required — `extension/src/worker.ts:88,116`.** The announced set is module-level, so it resets on worker restart. If the daemon *misses* the TCP disconnect (half-open socket) and the worker restarts with an empty announced set, an entry that leaves the queue before the restarted worker's first snapshot emits no clearance, leaving the daemon's record until the RPC deadline pops it. This is bounded by `ORIGIN_PROMPT_WINDOW_S` and strictly better than pre-change (no clearance path existed at all), so it's a follow-up note, not a defect for this round.

- **NIT (N1) — `extension/src/origins.ts:165`.** `match[0]?.entryId ?? ""` followed by `if (!targetId) return false;` is unreachable: a live queue entry always carries a non-empty `entryId`, and `match.length === 1` already guarantees `match[0]`. `match[0].entryId` suffices. Harmless.

**Verdict: clean.** No BLOCKER and no MAJOR. This round is TERMINAL — M1 and N1 are follow-ups, not grounds for another round.

damianvtran:

- **D1 (minor, design):** fixed in a804bca1 — expired notice now states the consequence: "It timed out or was cancelled, so nothing was granted or denied."
- **D2 (nit, design):** fixed in a804bca1 — sub no longer restates the title's cause.
- **Code NIT (origins.ts:165 unreachable guard):** fixed in a804bca1 — single-match branch uses a non-optional element with an explanatory comment.
- **Code M1 (bounded stale daemon record edge):** deferred — strictly better than pre-change behavior and self-heals at the RPC deadline; tracked as follow-up, not round-blocking per reviewer.

Extension gates re-run on a804bca1: typecheck clean, 67/67 tests, build clean.

damianvtran:
### Agent review — round 2

Reviewer: reviewer subagent (qwen3.8-max)
Scope: origin/main..a804bca1

Fresh round on head a804bca1 (delta since round 1: 430ae6b0..a804bca1). I read the delta diff first, then verified the fail-safe logic in `origins.ts` and ran the extension gates locally.

**Delta verification — matches the remediation claim exactly, nothing else.** The delta touches exactly three files:

1. **`extension/src/origins.ts` — round-1 NIT fixed.** The unreachable `match[0]?.entryId ?? ""` guard is replaced by `const single = match[0]!; targetId = single.entryId;` with a comment explaining why the element is guaranteed. Fail-safe behaviour is preserved verbatim: `match.length !== 1` still rejects zero/multiple matches *before* the element access, and `if (!targetId) return false;` still rejects an empty entryId after. The B1 consent-hole comment (ambiguity across several live entries for one origin stays rejected) is intact and accurate. No semantic change.
2. **`extension/src/popup/origin-flow.ts` — D1/D2 copy fix.** Expired notice sub is now "It timed out or was cancelled, so nothing was granted or denied." — states the consequence of the click rather than restating the title, with a comment recording the design-round rationale. The surrounding doc comment (line 102) still uses the phrase "expired or was cancelled" descriptively; that is internal documentation, not user-facing copy, and reads coherently — not a finding.
3. **`extension/tests/pure.test.mjs` — pin matches source.** The asserted sub string is character-for-character identical to the source string at origin-flow.ts:119.

No other files, no behaviour changes beyond the above. Zero `.py` changes in the delta, so round 1's Python verification (94/94 incl. cleared-event tests and gen parity) carries over unchanged.

**Gates (run locally on head a804bca1):** `pnpm typecheck` clean; `pnpm test` 67/67 pass; `pnpm build` clean. CI on this head: build, lint, type-check, pip-audit green; the two Python test legs were still in progress at review time (expected green given the delta contains no Python).

**Carried over:** M1 from round 1 (bounded stale daemon record on missed-disconnect worker restart) remains deferred as a follow-up per the remediation — correctly, it is strictly better than pre-change and self-heals at the RPC deadline.

**Findings:** none new. No BLOCKER, no MAJOR, no MINOR, no NIT in the delta.

**Verdict: clean.** This round is TERMINAL — the deferred M1 is a follow-up note, not grounds for another round.

damianvtran:
### Design review — round 2

Reviewer: designer (openrouter/qwen/qwen3.8-max)
Scope: origin/main..a804bca1

**Method note.** No rendered frames available (no browser in this context), so as in round 1 this is a source + CSS-token proxy review. The delta since round 1 (430ae6b0..a804bca1) is text-only inside the already-shipping `origin-ack` slot — same `.title`/`.sub` elements, neutral tone, glyph hidden — so layout risk is bounded. Geometry recomputed from popup.css tokens at head: content width 246px (300px body − 2×10px card margin − 2×1px border − 2×16px body padding).

**What changed since round 1.** Exactly three files: the expired-notice sub in `origin-flow.ts`, the dead optional-chain guard in `origins.ts` (no rendered output), and the matching test expectation in `pure.test.mjs`. A tree-wide grep confirms no remnant of the old sub and no other rendered-copy delta at this head.

**Verification of the new copy** — title "Request expired." / sub "It timed out or was cancelled, so nothing was granted or denied."

1. **Consequence for a consent click.** The sub now answers the question a user who just clicked Allow or Deny actually has: the click had no effect. "nothing was granted or denied" is decision-agnostic and symmetric — slightly more precise than round 1's "nothing was granted" suggestion, and correct for both buttons. ✓
2. **Title/sub redundancy (D2).** Resolved. The sub no longer merely restates the title: it adds the second cause ("or was cancelled", which the title does not say) and the consequence clause. "It timed out" is a mild synonym of "expired", but the anaphora is what grounds the pronoun, and the sub now carries new information in both directions (cause + consequence) — matching the house pattern where the sub adds, not repeats. ✓
3. **Voice.** Unchanged slot: serif 19px title, sans 13px/1.5 sub, neutral `--hairline-strong` rule, no colour spent — correct, an expired request is informational, not a failure to recover from. "Request expired." stays in the short period-terminated house voice ("Paired.", "Not connected."). "cancelled" keeps the codebase's consistent spelling. ✓
4. **300px fit.** Title ≈152px at 19px serif → 1 line. Sub is 64 chars ≈ 400–430px at 13px sans (0.48–0.52em average advance) → wraps to 2 lines over 246px, well inside the slot; the unchanged replacement sub (84 chars, ≈3 lines) remains the tallest case as shipped. No height-clamp or overflow interaction in the ack slot. ✓
5. **Copy pinned.** The test expectation in pure.test.mjs was updated with the copy, so the string cannot drift silently. ✓

**Findings.** None new.

- **D1 (round 1, MINOR):** resolved at a804bca1 — consequence stated, decision-agnostic.
- **D2 (round 1, NIT):** resolved at a804bca1 — sub adds cause and consequence instead of restating the title.
- **D3/D4 (round 1, NIT, record-only/pre-existing):** unchanged at this head; remain follow-ups, not gates.

**Verdict: TERMINAL.** No BLOCKER, no MAJOR, no new findings. The remediation landed exactly as round 1 recommended; the design round is clean on head a804bca1.

PR #366, opened 2026-08-28
feat(tui): add /failovers and a packaged failover guide

damianvtran:
## Summary

There was no way to answer "which provider is actually serving my requests, and what happens when this one runs out" without reading `config.yml` and then reading `providers/failover.py` to find out what the config meant. This adds the two surfaces that answer it: a read-only `/failovers` command, and a packaged `guide://failover` for the questions a listing cannot answer.

Ships as a **patch** (0.42.13 → 0.42.14): a diagnostic command plus a guide, no step-function capability.

## What `/failovers` shows

Primary, which chain key matched and how (exact / `provider/*` / `default`), the ordered targets with their configured effort, how many accounts each provider can rotate through, and a `← serving` marker on the row actually answering.

## Design constraints this had to respect

- **No spec resolution on the paint path.** `spec_for_target` can take ~10s on a cold memo for an unlisted model, and this is a synchronous slash command, so a "resolved effort" column would freeze the TUI. The listing shows **configured** effort; the reason is recorded in `_failover_rows`' docstring so a later "let's show what it really resolves to" does not reintroduce the freeze.
- **Serving is decided via `effective_model`**, not the route's `active_fallback`. `RemoteSession` implements the former and not the latter, so this is the one comparison correct on both the owning terminal and an attached follower. `failovers` is in `_FRONTEND_LOCAL_SLASHES` — it reads local config and local credential counts, so routing it to the owner would only add a hop.
- **Counts only, never identities.** Accounts are counted through `credential_provider_id`, so `xai-oauth` and `xai` collapse to one account rather than being double-counted. `row.data` is never read; `/accounts` owns identities.
- **`resolve_chain_key` is split out of `resolve_chain`** so the matched key is the engine's own answer. A caller re-deriving it with its own matching loop is a second definition of specificity that can disagree with routing.

## Visual evidence

Captured from the real `OperatorApp` (the lightweight test hosts declare no `CSS_PATH`, so they cannot show a stylesheet change), rendered and inspected as images. **All at 80 columns** — the width a split cmux pane produces, and the width at which the round-1 design review found the layout breaking.

| Populated | Fallback serving | Nothing configured |
|---|---|---|
| The cascade | Marker attached to its row | No chain matches |

The populated frame is this machine's **real** config and **real** credential counts — the 5-target `default` chain, not a fixture:

```
failover cascade
├─ primary        anthropic/claude-opus-5 · high · 4 accounts   ← serving
├─ matched chain  default
├─ 1. zai/glm-5.3                        1 account
├─ 2. kimi/k3                            1 account
├─ 3. alibaba-token-plan/qwen3.8-max     3 accounts
├─ 4. xai/grok-4.6                       1 account
├─ 5. openai/gpt-5.6-sol                 1 account
└─ tip  ask the agent to change models, effort or providers here
```

Details are padded into a column the way the sibling `/mcp` listing pads, so "which hop has accounts" reads vertically. The `← serving` marker sits **before** the detail, because it is the one element with no readable degraded form: a wrapped detail is still words, a wrapped `← serving` is an arrow at column 0 pointing at nothing.

The default install — nothing configured, which is the most likely caller — keeps its actionable key on one line at 80 columns, and offers the agent affordance rather than leaving a first-time user to author YAML:

```
failover cascade
├─ primary   anthropic/claude-opus-5 · high · 4 accounts
├─ no chain  no chain matches — set values.retry.fallbackChains.default
└─ tip       ask the agent to set up a fallback cascade
```

## Degrade behaviour

Six states, **each exercised by a test** (round 1 found the original table overstated this — two states had no test; they now do):

| State | Result |
|---|---|
| No provider facade | `warning`: `run: local-operator config list (TUI lacks the provider facade)` |
| Session not attached yet | `warning`: `session is still starting…` |
| Owner has not published a model (follower) | `warning`: `no model selected yet` — not an `error` |
| `retry.enabled` / `modelFallback` off | Primary plus a row naming **which** key disabled it |
| No chain matches / matched key lists only the current model | Distinct messages; the second no longer suggests a no-op |
| Config raises | `error`: `failover list failed: <error>` |

## The guide

`guide://failover` covers what the listing cannot: multi-account rotation *within* a provider before the chain leaves it (selection order, `_recover_quota_blocked`, and that a model-tier cap is per account so siblings are tried first), skipping a fallback whose provider is already at 0%, the quota reserve picking a lower-effort route on the same provider, sticky routing, and the three separate ways back up the cascade — the `primary_retry_at_ms` cooldown (set on route **change**, not bumped per use, because a sliding window pinned sessions for their whole life), `quota_pinned` re-probing at every message boundary regardless of cooldown, and the advisory per-target bench that `clear()` deliberately does not clear.

**Routing.** Five real user questions measured against the live index. Four previously selected **no guide at all**; all five now reach it:

| Question | Before | After |
|---|---|---|
| why did my model fall back to another provider | configuration | configuration, **failover** |
| failover keeps switching to kimi, how do I fix the cascade | *nothing* | **failover** |
| what models can I use for the fallback chain | *nothing* | **failover** |
| my anthropic account hit its quota and requests went to openai | *nothing* | **failover** |
| which provider is serving my requests right now | *nothing* | **failover** |

A guide's description costs context only in turns where it semantically routes (`skills/index.py` `render_block` emits one line per **selected** guide), so its resting footprint is zero. The description is 176 chars, inside the ≤180 contract pinned by `test_packaged_catalog_is_small_and_descriptions_are_prompt_sized`.

**Every bash recipe in the guide was extracted and run** before shipping, from a neutral directory, network-free:

```
=== RECIPE 1/2 (providers) ===   15 providers listed, openai … alibaba-token-plan-oauth
=== RECIPE 3 (models) ===        anthropic: 18 model ids; openrouter -> {} (the aggregator caveat, by design)
=== RECIPE 4 (cascade) ===       primary anthropic/claude-opus-5, matched key default, 5 targets in order
=== RECIPE 5 (account depth) ===  anthropic 4, alibaba-token-plan 3, zai/kimi/xai/openai 1 each
```

Recipe 4 imports `resolve_chain_key`, so it runs against this branch and will run against the installed runtime once released. The account recipe prints **counts and provider ids only** — that constraint is written into the snippet's own comment, since it is one edit away from printing bearer tokens.

The configuration guide's duplicate of these semantics is replaced by a three-line pointer; its `values.retry.*` key list stays. Two guides asserting cooldown rules is how they drift apart.

## Tests

- `tests/unit/providers/test_failover.py` — `resolve_chain_key` specificity (exact beats wildcard, longest wildcard wins), default fallthrough, and the `None` verdict matching `resolve_chain`'s.
- `tests/unit/tui/test_app_pilot.py` — **13** pilot tests: the populated render, counts-without-identities (asserts the fake's `a@b.c` never appears), every degrade state, and one regression test per round-1 major — matched-key-with-only-the-current-model, exactly-one-serving-marker, a usable provider never called `no accounts`, and the session-vs-disk drift flag.
- `tests/unit/tui/test_slash_echo.py` — the pinned echo/prompt policy entries.
- `tests/unit/guides/test_guides.py` — the packaged-guide inventory.

## Gates

Run over the whole tree, exactly as CI:

- `.venv/bin/python -m flake8 .` — clean
- `uvx --from black==26.1.0 black --check .` — clean, 569 files
- `uvx isort==5.13.2 --check .` — clean
- Unit suite — **7392 passed, 7 skipped**

Note for anyone reproducing locally: `tests/unit/tools/test_eval_tool.py` fails in **any** worktree (confirmed on unmodified `origin/main`) because the editable install resolves `local_operator` to the main checkout, so the eval worker subprocess cannot import from the worktree. `PYTHONPATH=$PWD` fixes it; CI is unaffected as it installs the package it tests.

## Review focus

Round 1 posted three independent rounds (code, design, UX); remediation is in the comment below. Two calls worth a second opinion:

- **No `failover` singular alias** (UX U2 suggested one). The picker sizes its name column on the widest `/name  /alias` pair, and `/failovers  /failover` is 21 cells against a previous max of 18 — it truncated `List all commands` on the 41-cell frame that `test_descriptions_come_back_above_the_collapse_width` pins. The singular still reaches the command through prefix matching. Rejected on cost, not on merit.
- **`/f` now ranks `failovers` above `effort`** (UX U2). That is correct prefix behaviour — `/f` is a prefix of `failovers` and not of `effort` — and the picker shows the row before Enter. Recorded as deliberate rather than special-cased.

### DIFF
--- a/docs/assets/pr-failovers/after-empty-80.svg
+++ b/docs/assets/pr-failovers/after-empty-80.svg
@@ -1,0 +1,133 @@
+<svg class="rich-terminal" viewBox="0 0 994 489.2" xmlns="http://www.w3.org/2000/svg">
+    <!-- Generated with Rich https://www.textualize.io -->
+    <style>
+
+    @font-face {
+        font-family: "Fira Code";
+        src: local("FiraCode-Regular"),
+                url("https://cdnjs.cloudflare.com/ajax/libs/firacode/6.2.0/woff2/FiraCode-Regular.woff2") format("woff2"),
+                url("https://cdnjs.cloudflare.com/ajax/libs/firacode/6.2.0/woff/FiraCode-Regular.woff") format("woff");
+        font-style: normal;
+        font-weight: 400;
+    }
+    @font-face {
+        font-family: "Fira Code";
+        src: local("FiraCode-Bold"),
+                url("https://cdnjs.cloudflare.com/ajax/libs/firacode/6.2.0/woff2/FiraCode-Bold.woff2") format("woff2"),
+                url("https://cdnjs.cloudflare.com/ajax/libs/firacode/6.2.0/woff/FiraCode-Bold.woff") format("woff");
+        font-style: bold;
+        font-weight: 700;
+    }
+
+    .terminal-475064663-matrix {
+        font-family: Fira Code, monospace;
+        font-size: 20px;
+        line-height: 24.4px;
+        font-variant-east-asian: full-width;
+    }
+
+    .terminal-475064663-title {
+        font-size: 18px;
+        font-weight: bold;
+        font-family: arial;
+    }
+
+    .terminal-475064663-r1 { fill: #c5c8c6 }
+.terminal-475064663-r2 { fill: #e9e5db }
+.terminal-475064663-r3 { fill: #837c6d }
+.terminal-475064663-r4 { fill: #57c785 }
+.terminal-475064663-r5 { fill: #1e1a14 }
+.terminal-475064663-r6 { fill: #4a4539 }
+.terminal-475064663-r7 { fill: #b48cd6 }
+.terminal-475064663-r8 { fill: #6ea8d8 }
+.terminal-475064663-r9 { fill: #ef8078;font-weight: bold }
+    </style>
+
+    <defs>
+    <clipPath id="terminal-475064663-clip-terminal">
+      <rect x="0" y="0" width="975.0" height="438.2" />
+    </clipPath>
+    <clipPath id="terminal-475064663-line-0">
+    <rect x="0" y="1.5" width="976" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-475064663-line-1">
+    <rect x="0" y="25.9" width="976" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-475064663-line-2">
+    <rect x="0" y="50.3" width="976" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-475064663-line-3">
+    <rect x="0" y="74.7" width="976" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-475064663-line-4">
+    <rect x="0" y="99.1" width="976" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-475064663-line-5">
+    <rect x="0" y="123.5" width="976" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-475064663-line-6">
+    <rect x="0" y="147.9" width="976" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-475064663-line-7">
+    <rect x="0" y="172.3" width="976" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-475064663-line-8">
+    <rect x="0" y="196.7" width="976" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-475064663-line-9">
+    <rect x="0" y="221.1" width="976" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-475064663-line-10">
+    <rect x="0" y="245.5" width="976" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-475064663-line-11">
+    <rect x="0" y="269.9" width="976" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-475064663-line-12">
+    <rect x="0" y="294.3" width="976" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-475064663-line-13">
+    <rect x="0" y="318.7" width="976" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-475064663-line-14">
+    <rect x="0" y="343.1" width="976" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-475064663-line-15">
+    <rect x="0" y="367.5" width="976" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-475064663-line-16">
+    <rect x="0" y="391.9" width="976" height="24.65"/>
+            </clipPath>
+    </defs>
+
+    <rect fill="#292929" stroke="rgba(255,255,255,0.35)" stroke-width="1" x="1" y="1" width="992" height="487.2" rx="8"/><text class="terminal-475064663-title" fill="#c5c8c6" text-anchor="middle" x="496" y="27">Local&#160;Operator</text>
+            <g transform="translate(26,22)">
+            <circle cx="0" cy="0" r="7" fill="#ff5f57"/>
+            <circle cx="22" cy="0" r="7" fill="#febc2e"/>
+            <circle cx="44" cy="0" r="7" fill="#28c840"/>
+            </g>
+        
+    <g transform="translate(9, 41)" clip-path="url(#terminal-475064663-clip-terminal)">
+    <rect fill="#14110c" x="0" y="1.5" width="976" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="0" y="25.9" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="12.2" y="25.9" width="951.6" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="963.8" y="25.9" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="0" y="50.3" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="12.2" y="50.3" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="24.4" y="50.3" width="24.4" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="48.8" y="50.3" width="195.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="244" y="50.3" width="707.6" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="951.6" y="50.3" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="963.8" y="50.3" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="0" y="74.7" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="12.2" y="74.7" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="24.4" y="74.7" width="24.4" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="48.8" y="74.7" width="36.6" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="85.4" y="74.7" width="85.4" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="170.8" y="74.7" width="549" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="719.8" y="74.7" width="231.8" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="951.6" y="74.7" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="963.8" y="74.7" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="0" y="99.1" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="12.2" y="99.1" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="24.4" y="99.1" width="24.4" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="48.8" y="99.1" width="36.6" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="85.4" y="99.1" width="97.6" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="183" y="99.1" width="732" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="915" y="99.1" width="36.6" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="951.6" y="99.1" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="963.8" y="99.1" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="0" y="123.5" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="12.2" y="123.5" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="24.4" y="123.5" width="24.4" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="48.8" y="123.5" width="36.6" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="85.4" y="123.5" width="61" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="146.4" y="123.5" width="695.4" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="841.8" y="123.5" width="109.8" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="951.6" y="123.5" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="963.8" y="123.5" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="0" y="147.9" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="12.2" y="147.9" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="24.4" y="147.9" width="24.4" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="48.8" y="147.9" width="36.6" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="85.4" y="147.9" width="36.6" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="122" y="147.9" width="536.8" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="658.8" y="147.9" width="292.8" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="951.6" y="147.9" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="963.8" y="147.9" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="0" y="172.3" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="12.2" y="172.3" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="24.4" y="172.3" width="939.4" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="963.8" y="172.3" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="0" y="196.7" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="12.2" y="196.7" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="24.4" y="196.7" width="939.4" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="963.8" y="196.7" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="0" y="221.1" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="12.2" y="221.1" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="24.4" y="221.1" width="939.4" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="963.8" y="221.1" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="0" y="245.5" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="12.2" y="245.5" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="24.4" y="245.5" width="939.4" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="963.8" y="245.5" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="0" y="269.9" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="12.2" y="269.9" width="951.6" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="963.8" y="269.9" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="0" y="294.3" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#1e1a14" x="12.2" y="294.3" width="951.6" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="963.8" y="294.3" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="0" y="318.7" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#1e1a14" x="12.2" y="318.7" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#1e1a14" x="24.4" y="318.7" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#1e1a14" x="36.6" y="318.7" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#1e1a14" x="48.8" y="318.7" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#e9e5db" x="61" y="318.7" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#1e1a14" x="73.2" y="318.7" width="280.6" height="24.65" shape-rendering="crispEdges"/><rect fill="#1e1a14" x="353.8" y="318.7" width="585.6" height="24.65" shape-rendering="crispEdges"/><rect fill="#1e1a14" x="939.4" y="318.7" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#1e1a14" x="951.6" y="318.7" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="963.8" y="318.7" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="0" y="343.1" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#1e1a14" x="12.2" y="343.1" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#1e1a14" x="24.4" y="343.1" width="927.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#1e1a14" x="951.6" y="343.1" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="963.8" y="343.1" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="0" y="367.5" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#1e1a14" x="12.2" y="367.5" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#1e1a14" x="24.4" y="367.5" width="24.4" height="24.65" shape-rendering="crispEdges"/><rect fill="#1e1a14" x="48.8" y="367.5" width="158.6" height="24.65" shape-rendering="crispEdges"/><rect fill="#1e1a14" x="207.4" y="367.5" width="36.6" height="24.65" shape-rendering="crispEdges"/><rect fill="#1e1a14" x="244" y="367.5" width="24.4" height="24.65" shape-rendering="crispEdges"/><rect fill="#1e1a14" x="268.4" y="367.5" width="48.8" height="24.65" shape-rendering="crispEdges"/><rect fill="#1e1a14" x="317.2" y="367.5" width="36.6" height="24.65" shape-rendering="crispEdges"/><rect fill="#1e1a14" x="353.8" y="367.5" width="24.4" height="24.65" shape-rendering="crispEdges"/><rect fill="#1e1a14" x="378.2" y="367.5" width="317.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#1e1a14" x="695.4" y="367.5" width="231.8" height="24.65" shape-rendering="crispEdges"/><rect fill="#1e1a14" x="927.2" y="367.5" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#1e1a14" x="939.4" y="367.5" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#1e1a14" x="951.6" y="367.5" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="963.8" y="367.5" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="0" y="391.9" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#1e1a14" x="12.2" y="391.9" width="951.6" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="963.8" y="391.9" width="12.2" height="24.65" shape-rendering="crispEdges"/><rect fill="#14110c" x="0" y="416.3" width="976" height="24.65" shape-rendering="crispEdges"/>
+    <g class="terminal-475064663-matrix">
+    <text class="terminal-475064663-r1" x="976" y="20" textLength="12.2" clip-path="url(#terminal-475064663-line-0)">
+</text><text class="terminal-475064663-r1" x="976" y="44.4" textLength="12.2" clip-path="url(#terminal-475064663-line-1)">
+</text><text class="terminal-475064663-r3" x="48.8" y="68.8" textLength="195.2" clip-path="url(#terminal-475064663-line-2)">failover&#160;cascade</text><text class="terminal-475064663-r1" x="976" y="68.8" textLength="12.2" clip-path="url(#terminal-475064663-line-2)">
+</text><text class="terminal-475064663-r3" x="48.8" y="93.2" textLength="36.6" clip-path="url(#terminal-475064663-line-3)">├─&#160;</text><text class="terminal-475064663-r4" x="85.4" y="93.2" textLength="85.4" clip-path="url(#terminal-475064663-line-3)">primary</text><text class="terminal-475064663-r3" x="170.8" y="93.2" textLength="549" clip-path="url(#terminal-475064663-line-3)">&#160;&#160;anthropic/claude-opus-5&#160;·&#160;high&#160;·&#160;4&#160;accounts</text><text class="terminal-475064663-r1" x="976" y="93.2" textLength="12.2" clip-path="url(#terminal-475064663-line-3)">
+</text><text class="terminal-475064663-r3" x="48.8" y="117.6" textLength="36.6" clip-path="url(#terminal-475064663-line-4)">├─&#160;</text><text class="terminal-475064663-r4" x="85.4" y="117.6" textLength="97.6" clip-path="url(#terminal-475064663-line-4)">no&#160;chain</text><text class="terminal-475064663-r3" x="183" y="117.6" textLength="732" clip-path="url(#terminal-475064663-line-4)">&#160;&#160;no&#160;chain&#160;matches&#160;—&#160;set&#160;values.retry.fallbackChains.default</text><text class="terminal-475064663-r1" x="976" y="117.6" textLength="12.2" clip-path="url(#terminal-475064663-line-4)">
+</text><text class="terminal-475064663-r3" x="48.8" y="142" textLength="36.6" clip-path="url(#terminal-475064663-line-5)">├─&#160;</text><text class="terminal-475064663-r4" x="85.4" y="142" textLength="61" clip-path="url(#terminal-475064663-line-5)">stale</text><text class="terminal-475064663-r3" x="146.4" y="142" textLength="695.4" clip-path="url(#terminal-475064663-line-5)">&#160;&#160;config.yml&#160;changed&#160;since&#160;this&#160;session&#160;started&#160;·&#160;/reload</text><text class="terminal-475064663-r1" x="976" y="142" textLength="12.2" clip-path="url(#terminal-475064663-line-5)">
+</text><text class="terminal-475064663-r3" x="48.8" y="166.4" textLength="36.6" clip-path="url(#terminal-475064663-line-6)">└─&#160;</text><text class="terminal-475064663-r4" x="85.4" y="166.4" textLength="36.6" clip-path="url(#terminal-475064663-line-6)">tip</text><text class="terminal-475064663-r3" x="122" y="166.4" textLength="536.8" clip-path="url(#terminal-475064663-line-6)">&#160;&#160;ask&#160;the&#160;agent&#160;to&#160;set&#160;up&#160;a&#160;fallback&#160;cascade</text><text class="terminal-475064663-r1" x="976" y="166.4" textLength="12.2" clip-path="url(#terminal-475064663-line-6)">
+</text><text class="terminal-475064663-r1" x="976" y="190.8" textLength="12.2" clip-path="url(#terminal-475064663-line-7)">
+</text><text class="terminal-475064663-r1" x="976" y="215.2" textLength="12.2" clip-path="url(#terminal-475064663-line-8)">
+</text><text class="terminal-475064663-r1" x="976" y="239.6" textLength="12.2" clip-path="url(#terminal-475064663-line-9)">
+</text><text class="terminal-475064663-r1" x="976" y="264" textLength="12.2" clip-path="url(#terminal-475064663-line-10)">
+</text><text class="terminal-475064663-r1" x="976" y="288.4" textLength="12.2" clip-path="url(#terminal-475064663-line-11)">
+</text><text class="terminal-475064663-r1" x="976" y="312.8" textLength="12.2" clip-path="url(#terminal-475064663-line-12)">
+</text><text class="terminal-475064663-r2" x="24.4" y="337.2" textLength="12.2" clip-path="url(#terminal-475064663-line-13)">❯</text><text class="terminal-475064663-r3" x="73.2" y="337.2" textLength="280.6" clip-path="url(#terminal-475064663-line-13)">Message&#160;Local&#160;Operator…</text><text class="terminal-475064663-r1" x="976" y="337.2" textLength="12.2" clip-path="url(#terminal-475064663-line-13)">
+</text><text class="terminal-475064663-r1" x="976" y="361.6" textLength="12.2" clip-path="url(#terminal-475064663-line-14)">
+</text><text class="terminal-475064663-r3" x="24.4" y="386" textLength="24.4" clip-path="url(#terminal-475064663-line-15)">◆&#160;</text><text class="terminal-475064663-r2" x="48.8" y="386" textLength="158.6" clip-path="url(#terminal-475064663-line-15)">Claude&#160;Opus&#160;5</text><text class="terminal-475064663-r6" x="207.4" y="386" textLength="36.6" clip-path="url(#terminal-475064663-line-15)">&#160;›&#160;</text><text class="terminal-475064663-r3" x="244" y="386" textLength="24.4" clip-path="url(#terminal-475064663-line-15)">▴&#160;</text><text class="terminal-475064663-r7" x="268.4" y="386" textLength="48.8" clip-path="url(#terminal-475064663-line-15)">high</text><text class="terminal-475064663-r6" x="317.2" y="386" textLength="36.6" clip-path="url(#terminal-475064663-line-15)">&#160;›&#160;</text><text class="terminal-475064663-r3" x="353.8" y="386" textLength="24.4" clip-path="url(#terminal-475064663-line-15)">⌂&#160;</text><text class="terminal-475064663-r8" x="378.2" y="386" textLength="317.2" clip-path="url(#terminal-475064663-line-15)">/private/tmp/lop-failovers</text><text class="terminal-475064663-r9" x="927.2" y="386" textLength="12.2" clip-path="url(#terminal-475064663-line-15)">!</text><text class="terminal-475064663-r1" x="976" y="386" textLength="12.2" clip-path="url(#terminal-475064663-line-15)">
+</text><text class="terminal-475064663-r1" x="976" y="410.4" textLength="12.2" clip-path="url(#terminal-475064663-line-16)">
+</text>
+    </g>
+    </g>
+</svg>
--- a/docs/assets/pr-failovers/after-fallback-serving-80.svg
+++ b/docs/assets/pr-failovers/after-fallback-serving-80.svg
@@ -1,0 +1,132 @@
+<svg class="rich-terminal" viewBox="0 0 994 489.2" xmlns="http://www.w3.org/2000/svg">
+    <!-- Generated with Rich https://www.textualize.io -->
+    <style>
+
+    @font-face {
+        font-family: "Fira Code";
+        src: local("FiraCode-Regular"),
+                url("https://cdnjs.cloudflare.com/ajax/libs/firacode/6.2.0/woff2/FiraCode-Regular.woff2") format("woff2"),
+                url("https://cdnjs.cloudflare.com/ajax/libs/firacode/6.2.0/woff/FiraCode-Regular.woff") format("woff");
+        font-style: normal;
+        font-weight: 400;
+    }
+    @font-face {
+        font-family: "Fira Code";
+        src: local("FiraCode-Bold"),
+                url("https://cdnjs.cloudflare.com/ajax/libs/firacode/6.2.0/woff2/FiraCode-Bold.woff2") format("woff2"),
+                url("https://cdnjs.cloudflare.com/ajax/libs/firacode/6.2.0/woff/FiraCode-Bold.woff") format("woff");
+        font-style: bold;
+        font-weight: 700;
+    }
+
+    .terminal-1651858060-matrix {
+        font-family: Fira Code, monospace;
+        font-size: 20px;
+        line-height: 24.4px;
+        font-variant-east-asian: full-width;
+    }
+
+    .terminal-1651858060-title {
+        font-size: 18px;
+        font-weight: bold;
+        font-family: arial;
+    }
+
+    .terminal-1651858060-r1 { fill: #c5c8c6 }
+.terminal-1651858060-r2 { fill: #e9e5db }
+.terminal-1651858060-r3 { fill: #837c6d }
+.terminal-1651858060-r4 { fill: #57c785 }
+.terminal-1651858060-r5 { fill: #1e1a14 }
+.terminal-1651858060-r6 { fill: #4a4539 }
+.terminal-1651858060-r7 { fill: #6ea8d8 }
+.terminal-1651858060-r8 { fill: #ef8078;font-weight: bold }
+    </style>
+
+    <defs>
+    <clipPath id="terminal-1651858060-clip-terminal">
+      <rect x="0" y="0" width="975.0" height="438.2" />
+    </clipPath>
+    <clipPath id="terminal-1651858060-line-0">
+    <rect x="0" y="1.5" width="976" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-1651858060-line-1">
+    <rect x="0" y="25.9" width="976" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-1651858060-line-2">
+    <rect x="0" y="50.3" width="976" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-1651858060-line-3">
+    <rect x="0" y="74.7" width="976" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-1651858060-line-4">
+    <rect x="0" y="99.1" width="976" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-1651858060-line-5">
+    <rect x="0" y="123.5" width="976" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-1651858060-line-6">
+    <rect x="0" y="147.9" width="976" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-1651858060-line-7">
+    <rect x="0" y="172.3" width="976" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-1651858060-line-8">
+    <rect x="0" y="196.7" width="976" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-1651858060-line-9">
+    <rect x="0" y="221.1" width="976" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-1651858060-line-10">
+    <rect x="0" y="245.5" width="976" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-1651858060-line-11">
+    <rect x="0" y="269.9" width="976" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-1651858060-line-12">
+    <rect x="0" y="294.3" width="976" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-1651858060-line-13">
+    <rect x="0" y="318.7" width="976" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-1651858060-line-14">
+    <rect x="0" y="343.1" width="976" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-1651858060-line-15">
+    <rect x="0" y="367.5" width="976" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-1651858060-line-16">
+    <rect x="0" y="391.9" width="976" height="24.65"/>
+            </clipPath>
+    </defs>
+
+    <rect fill="#292929" stroke="rgba(255,255,255,0.35)" stroke-width="1" x="1" y="1" width="992" height="487.2" rx="8"/><text class="terminal-1651858060-title" fill="#c5c8c6" text-anchor="middle" x="496" y="27">Local&#160;Operator</text>
+            <g transform="translate(26,22)">
+            <circle cx="0" cy="0" r="7" fill="#ff5f57"/>
+            <circle cx="22" cy="0" r="7" fill="#febc2e"/>
+            <circle cx="44" cy="0" r="7" fill="#28c840"/>
+            </g>
+        
+    <g transform="translate(9, 41)" clip-path="url(#terminal-1651858060-clip-terminal)">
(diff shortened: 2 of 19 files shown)


damianvtran:
### Design review — round 1

**Reviewer:** `designer` subagent (lopdev team) — anthropic/claude-opus-5
**Scope:** `origin/main`..`52a39a02` (dev-failovers)
**Surface reviewed:** `/failovers` tree output (`local_operator/tui/app.py:11776-11935`), rendered via the real `OperatorApp` per AGENTS.md "Visual validation"

I looked at the three frames in the PR body as rendered images, then captured my own from the real `OperatorApp` at 60/70/76/80/90/100/120 columns, plus two states the PR never rendered: **a fallback actually serving**, and **the empty state with the shipped default primary** rather than the `test/model` fixture. Geometry below is measured off the SVG text nodes (12.0px/cell, transcript text starts at x=49).

The command is a genuinely good idea and the information architecture is right: primary → matched chain → ordered targets → escape hatch is the order a reader wants. Two findings are about the frame falling apart at widths the PR did not capture.

---

#### D1 — MAJOR — The `← serving` marker wraps onto its own line and detaches from its row

`local_operator/tui/app.py:11915,11920`

The marker is appended as trailing text on an already-long detail string, so it is the first thing the wrap eats. When a **fallback** is serving — the state the command exists to answer — the marker separates from the row it belongs to:

At 80 columns, target 3 serving:

```
├─ 3. alibaba-token-plan/qwen3.8-max  model default effort · no accounts
← serving
```

`← serving` now sits at column 0, unindented, under a tree glyph, pointing left at nothing. A reader scanning the cascade reads it as a continuation of row 3 only if they are careful; it looks at least as much like a row of its own, and the arrow's referent is gone. At 76 columns it lands as `accounts   ← serving`, which is the same problem one word later.

Measured thresholds (row cells = 3 branch + name + 2 + detail; usable ≈ cols − 6):

| State | Row width | Wraps below |
|---|---|---|
| Primary serving, this machine's real config | 67 cells | ~73 cols |
| **Fallback #3 serving, this machine's real config** | **83 cells** | **~89 cols** |
| Hint row | 72 cells | ~78 cols |

So on the operator's own configuration the marker breaks on any terminal narrower than ~89 columns — an 80-column terminal is not exotic, and it is exactly the width a split cmux pane produces. Every frame in the PR body is a single wide terminal, which is why this did not show up.

The marker is also the one piece of information with no fallback: order, effort and counts survive a wrap as readable text, but a detached arrow is actively misleading. Worth either putting it before the detail (right after the name, where it is short and stable), or dropping the detail's low-value text at narrow widths the way `_removal_detail` at `app.py:11895` already reasons about shedding a column rather than letting it wreck the row.

---

#### D2 — MAJOR — `model default effort` on every row is 100 cells of repetition, and it is what pushes rows into the wrap

`local_operator/tui/app.py:11917`

On the real config the string is identical on **5 of 5** target rows. It carries zero bits between rows — it differs only in the case nobody has configured, which is the case where it says "nothing was configured". 100 cells of the tree's total width is spent restating an absence.

This is not just noise, it is the direct cause of D1: strip it from the rows where it is the default and the fallback-serving row drops from 83 to ~62 cells, which fits an 80-column terminal with room to spare.

The house convention here is explicit and points the other way. `_removal_detail`'s docstring (`app.py:11895-11907`) rejects exactly this shape:

> Not "logged in". `/logout` offers only providers that HAVE a credential, so that state is true of every row by construction — a column carrying no bits at all, holding cells the description needs at narrow widths.

Same argument, same surface, opposite outcome. The row should state what *differs*: show `high effort` when configured, show nothing when it is the model default. The distinction the phrasing is protecting (configured vs resolved) is real and worth keeping — but it belongs in the one place a reader will look for it, and the guide already covers it.

Related: the phrase reads ambiguously even where it is correct. "model default effort" parses as *the effort named "model default"* on first read, not *this target has no configured effort*.

---

#### D3 — MINOR — The hint row is styled identically to a cascade entry and reads as target 6

Rendered frame, populated:

```
├─ 5. openai/gpt-5.6-sol  model default effort · 1 account
└─   ask the agent to change models, effort or providers in this cascade
```

The hint has an empty `name` (`app.py:11924`), so `_tree_listing` gives it the `└─` terminal glyph and the detail style, and the empty name collapses the two-space gap into leading whitespace. Measured: the hint's text starts at x=85 — **the exact x of every row name above it**, not the detail column at x=171-488. It is the last leaf of the tree, aligned with the row names, in the same colour as every detail. Structurally and visually it *is* a cascade entry that happens to be a sentence.

The intent (an escape hatch, not a route) is good and I would keep it. But it needs to be distinguishable from the thing it sits inside — separated from the tree, or visibly de-emphasised, rather than occupying the terminal branch position that in every other `_tree_listing` means "last item in this list".

Note it is also the second-widest row at 72 cells, so it wraps below ~78 columns and produces a two-line orphan (`providers in this cascade`) hanging under the tree.

---

#### D4 — MINOR — Details do not align into a column, unlike the sibling `/mcp` listing

Measured detail x-positions in the populated frame: **171, 244, 256, 207, 488, 268, 342**. The values ripple in and out by 26 cells with the length of each model id, so `model default effort · N account(s)` — the part a reader is actually comparing down the list — never forms a column.

Worth noting the PR body's own mock-up shows it *aligned*:

```
├─ 1. zai/glm-5.3                        model default effort · 1 account
├─ 3. alibaba-token-plan/qwen3.8-max     model default effort · 3 accounts
```

That is not what the code renders. The description was written describing the intended frame rather than the produced one, which is worth correcting either way.

The house precedent is `/mcp` at `app.py:14558-14561`, which pads deliberately:

```python
name_column = max(len(name) for name, _, _ in rows)
status_column = max(len(status) for _, status, _ in rows)
items = [(name.ljust(name_column), f"{status.ljust(status_column)}  {detail}".rstrip()) ...]
```

and renders as a clean two-column grid. `/failovers` has the same shape — one row per entity, a name of varying length, a fixed-vocabulary detail — and is the listing where down-column comparison matters most, since the whole point is scanning which provider has accounts. Fixing D2 shortens the details enough that a `ljust` on the name column fits comfortably.

---

#### D5 — MINOR — The empty state's pointer wraps and buries the one thing it is telling you to do

Captured with the shipped default primary (`anthropic/claude-opus-5`), which the PR's `test/model` fixture is 13 cells shorter than:

```
└─ no chain  nothing matches anthropic/claude-opus-5 — set values.retry.fallbackChains.default
```

94 cells, so it needs ~100 columns; below that the config key — the entire actionable payload — wraps to a second unindented line, and at 80 columns it is `values.retry.fallbackChains.default` alone at column 0.

The content is right and the decision to still show the primary is right. But this is described in the PR as "the most likely caller", i.e. the default install, and its one call to action is the part that breaks first. It reads better with the key on its own deliberate line, or with the sentence shortened so the key survives (`no chain matches — set values.retry.fallbackChains.default` is 24 cells cheaper and loses nothing, since the primary is stated one row above).

---

#### D6 — NIT — The marker shares its exact colour with the detail text it must stand out from

Measured: `← serving` renders in `#837c6d` (`r3`, the dim token) on `#14110c` — the same fill as the detail string, the branch glyphs and the caption. Contrast against background is fine (4.55:1); contrast against the text it needs to be distinguished from is **1.00:1**. It is legible, and it is findable only by the three-space gap.

`/theme`'s `← current` (`app.py:10686`) gets away with the same token because it sits in a picker's dedicated detail column with one marked row on screen. Here it must be located by scanning eight rows of identically-coloured text. A distinct tint would let a reader answer "what is serving" without reading any row in full — which is the question the command was built for.

---

#### D7 — NIT — `matched chain  default` sits in the cascade without reading as metadata

`primary` and `matched chain` are facts about the cascade; rows `1.`-`5.` are the cascade. All four share the `├─` glyph and the green name style, so the numbered run appears to start two items in. The numbering does most of the work of separating them, and this is minor — but it is why the `←`-serving row and the `matched chain` row can be momentarily confused for the same class of thing on first read.

---

### Verdict

**NOT TERMINAL** — two MAJOR findings (D1, D2) are open.

They are the same defect from two directions: the rows are wider than they need to be because of repeated text (D2), and the one element that must not wrap is the one placed where the wrap hits first (D1). Fixing D2 largely resolves D1 for real configurations, and makes D4's column alignment affordable. None of this requires touching the command's logic, its degrade paths, or the `spec_for_target` constraint — it is all in how `_failover_rows` composes the detail string.

D3, D4, D5 are worth doing in the same pass since they live in the same function. D6 and D7 are recorded as follow-ups.

One thing to keep: the escape-hatch row's *content*, the decision to show the primary in the empty state, and counts-not-identities are all correct calls, and the information order needs no change.

**Frames I captured** (real `OperatorApp`, `env -u NO_COLOR TERM=xterm-256color`, `app.save_screenshot`): populated at 60/70/80/100/120; fallback-serving at 72/76/80/90/100/120; empty state at 80/100 with both the `test/model` fixture and the shipped `anthropic/claude-opus-5` default. Happy to attach any of them if useful for the remediation round.

damianvtran:
### UX review — round 1

**Reviewer:** ux-reviewer subagent (lopdev team), anthropic/claude-opus-5
**Scope:** fa62d097ce7007619271d622d6fe0ccb81c3aa83..52a39a021c63e8dd76f6e9cf3ca22655639a6585

Walked the real flow, not the stills: `git worktree add /tmp/ux-failovers origin/dev-failovers`, symlinked the repo `.venv`, and drove the actual `OperatorApp` under `run_test` — typing `/failovers` a keystroke at a time through the composer, reading the picker's live suggestion set, and rendering the exported SVG frames to look at them. Nine walks: discoverability, the no-argument choice, typos and junk arguments, narrow terminals and resize, the four degrade states, an actively-serving fallback, a 40-target chain, and the agent-mutability promise.

**Design constraints verified as holding.** The no-freeze claim is real: `_cmd_failovers` returns in **13.1 ms** on a 40-target chain and **4.7–11.0 ms** on repeats, and the composer accepted `hello` typed immediately after submit, so the input loop never stalls. The counts-not-identities rule holds — with the fake's `a@b.c` credential present, neither the email nor a credential row id appears anywhere in the output. The `← serving` marker moves correctly onto the fallback row when `effective_model` differs from the primary (walk 9), which is the moment the command exists for and it reads exactly right.

---

#### U1 — MAJOR — the listing shows a cascade the running session will not route on, and never says so

`_failover_rows` builds a **fresh** `ConfigManager(config_dir())` on every invocation (`local_operator/tui/app.py:11881`), so it prints whatever is on disk right now. The routing engine does not: `session_factory.py:1145-1149` hands `create_stream_fn` the mapping from `config_manager.get_config().values` **once at session build**, and `SessionStreamFn` holds that snapshot for the session's life (`model/configure.py:1622`, read back at `1960` and `2304`). `ConfigManager.get_config()` returns the cached in-memory copy (`config.py:390-396`); the only `reload()` caller in the tree is the first-run `/login` path (`cli.py:2926`, fired at `tui/app.py:13474` only when `self._setup_state`). Nothing watches `config.yml`.

Walked it, and they diverge:

```
step 1  /failovers            -> 1. zai/glm-5.3
step 2  agent edits config.yml to add kimi/k3
step 3  /failovers            -> 1. zai/glm-5.3
                                 2. kimi/k3     <- listing updated
        RetrySettings from the session's captured snapshot:
                                 ['zai/glm-5.3'] <- router did NOT
```

This is a UX finding rather than a code one because of **which** row it lands under. The last tree row tells the user "ask the agent to change models, effort or providers in this cascade". They do; the agent edits `config.yml`; they re-run `/failovers` to check, and the listing **confirms the change** — a green light. The next fallback still goes to the old chain. The one surface built to answer "what will happen when this provider runs out" now answers it wrongly, and it is most wrong precisely for the user who followed the affordance the command itself offered. A diagnostic that disagrees with the engine is worse than no diagnostic, because it is trusted.

What a user would expect: either the listing reads the same snapshot the router does, or the row that promises mutability names the step that makes it real. `/reload` already does exactly this ("relaunch this conversation on the current install", `tui/app.py:3639`) and is never mentioned. Cheapest honest fix inside this PR's read-only scope: when the on-disk chain differs from the session's captured one, say so on its own row — `edited on disk · /reload to apply` — and have the hint read `ask the agent to change …, then /reload`.

#### U2 — MINOR — `/f` + Enter silently changes command, and `/failover` is not offered as an alias

Registering `failovers` re-points a key sequence that already had a meaning. Driving both branches:

| typed | on `origin/main` | on this branch |
|---|---|---|
| `/f` | picker offers `effort`, Enter runs `/effort` | picker offers `failovers`, Enter runs `/failovers` |

`failovers` scores `SCORE_PREFIX` (900) on `/f` while `effort` only reaches the fuzzy band (`autocomplete.py:201-217`), so it takes the row outright. `/effort` is a command with a keyboard partner (`shift+tab` cycles) that a heavy user types constantly; `/f`+Enter is exactly the shorthand that muscle memory forms. It now runs something else, with no collision cue. Not a blocker — the picker does show `failovers` on the row before Enter, so the change is visible to anyone looking — but it is a silent rebind of an existing gesture and worth a deliberate decision rather than a side effect of registration order.

On the name itself: `/failovers` is right and better than the alternatives. `/routing` and `/cascade` are not in the user's vocabulary here (the config key is `fallbackChains`, the guide is `failover`), and both come back `unknown command`. But **`/failover` singular** is at least as natural a thing to type, and it currently only works by fuzzy luck — it resolves because `slash_command_for` gets there via the picker's completion, not because it is an alias. `aliases=("failover",)` costs one tuple and makes the singular a first-class spelling. Compare `/model, /models` and `/exit, /quit`, which already carry exactly this kind of pair.

#### U3 — MINOR — the empty and disabled states drop the agent affordance, for the users who need it most

The mutability hint is appended only on the fully-populated path (`app.py`, last `rows.append` before `return rows`). Walked all four states:

| state | last row | offers "ask the agent" |
|---|---|---|
| populated | `ask the agent to change models, effort or providers in this cascade` | yes |
| nothing configured | `no chain  nothing matches test/model — set values.retry.fallbackChains.default` | **no** |
| `retry.enabled` false | `cascade off  retry.enabled is false — no fallback route is used` | **no** |
| `modelFallback` false | `cascade off  retry.modelFallback is false — no fallback route is used` | **no** |

This is backwards relative to the PR's own stated premise that "the most likely caller is a default install with nothing configured at all". The user who already has a working five-target cascade is told they can ask the agent to change it; the user staring at `no chain` — the one who has nothing and must author YAML from scratch — is handed a config path and left to it. Both non-populated rows name a `values.retry.*` key, which is developer vocabulary, and neither mentions that the agent will write it for them or that `guide://failover` exists. Expected: the affordance appears in every state, phrased for the state — `ask the agent to set up a fallback cascade` on the empty one, `ask the agent to turn the cascade back on` on the disabled ones.

#### U4 — MINOR — `no accounts` reads as a problem on providers that never need a credential

`_account_detail` reports `no accounts` for any provider with zero stored credentials, but two registry providers route fine with none: `ollama` and `test` both set `allows_missing_api_key=True` (`providers/registry.py:450`, `:526`), and `_target_has_auth` returns `True` for them without consulting the auth store (`model/configure.py:1973-1974`). Walked a chain with a local Ollama target:

```
├─ 1. ollama/llama3  model default effort · no accounts
├─ 2. openrouter/x   model default effort · 1 account
```

Row 1 is a perfectly healthy route and row 2 is a perfectly healthy route, rendered in vocabulary that makes the first look broken. A user reading a cascade to find out why it is not failing over would reasonably chase the Ollama row first. Expected: `no credential needed` (or similar) when `allows_missing_api_key` is set. The registry flag is already imported territory for this code path.

#### U5 — MINOR — junk arguments are silently swallowed

`/failovers wat` and `/failovers anthropic` both render the full listing and discard the argument without a word. The no-argument choice is correct for this command — the cascade tree is the whole answer and there is nothing to filter by — but `anthropic` is a genuinely plausible thing to type at a command that lists providers, and the user gets a full-looking answer with no signal that their filter was ignored. Worth noting this matches existing house behaviour exactly (`/accounts nonsense` and `/provider nonsense` on `main` both do the same), so it is consistency, not regression — recording it as the smallest of the five rather than asking this PR to break the pattern alone. If anything changes, it belongs across the listing commands together.

#### U6 — NIT — the "why" and the "when do I get my primary back" still need a second surface

The listing answers *what is serving* but not *why it left the primary* or *when it comes back*. That state exists and is well-documented — `FailoverRouteState.primary_retry_at_ms`, `quota_pinned`, `target_retry_at_ms` (`providers/failover.py:1308-1344`) — and the guide explains all three in prose. A user watching `← serving` sit on row 1 has no in-app way to learn whether that is a 60-second transport cooldown or a 24-hour quota pin, which is the natural follow-up question the listing provokes. Correctly out of scope here (the PR is deliberately read-only and `RemoteSession` does not carry route state, which is exactly why `effective_model` was chosen), and the guide is the right first home for it. Recording as a follow-up.

#### U7 — NIT — long selectors wrap and lose the tree indent at narrow widths

At 60 and 40 columns, rows like `1. alibaba-token-plan/qwen3.8-max-thinking-latest  high effort · no accounts` wrap to column 0, so the continuation sits left of its own `├─` glyph and the cascade stops reading as a tree. Verified this is **pre-existing house behaviour, not a regression** — captured `/skills` on unmodified `origin/main` at 60 columns and it wraps identically. Nothing for this PR to fix; noting it because `/failovers` rows are longer than most listings (selector + effort + account count on one line), so it hits the case sooner than its neighbours do.

---

**Things I specifically checked that are fine.** Discoverability through `/help` is good: the row reads `/failovers   Show the model failover cascade and what is serving`, and it lands in the provider cluster between `/accounts` and `/usage`, which is where a user hunting for it would look. `/provider` and `/accounts` do not point at it, but neither points at the other either, so a cross-reference would be new house behaviour rather than a missing one — not raising it. Typos degrade correctly (`/failvoers` → `unknown command: /failvoers — try /help`, preserving what was actually typed). The four degrade states each name the specific key responsible rather than saying "off", which is the right call and genuinely saves a config.yml hunt. Resize from 120 to 50 columns reflows without corruption. The `guide://failover` guide is discoverable in the packaged catalog (confirmed via `discover_guides()`) with a description that routes on the right vocabulary, so the agent half of U1's affordance does have something to read — the gap is purely that acting on it does not take effect.

**Verdict: NOT TERMINAL.** One MAJOR (U1) — the listing confirms cascade edits the running session will not honour, which lands directly on the affordance the command advertises. U2–U5 are MINOR, U6–U7 are NITs and fine as follow-ups. Re-run this round against the new head once U1 is answered.

damianvtran:
### Agent review — round 1

Reviewer: `reviewer` subagent (independent of the author) — anthropic/claude-opus-5
Scope: `fa62d097ce7007619271d622d6fe0ccb81c3aa83..52a39a021c63e8dd76f6e9cf3ca22655639a6585`

Reviewed the diff, then verified by execution: ran every bash recipe in the new guide against the PR head, drove the real `OperatorApp` to confirm rendered output, and ran a differential proof of the `resolve_chain` refactor. Gates re-run locally at head: `flake8` clean, `black --check` (26.1.0) clean, `isort --check` clean, `pyright` **0 errors**, and the new/touched suites pass (6 `/failovers` pilot tests, 227 in `test_failover.py`/`test_guides.py`/`test_slash_echo.py`). CI is green on `52a39a0` — every check `SUCCESS`.

What I confirmed as correct, so remediation does not re-litigate it:

- **`resolve_chain` refactor is behaviour-preserving.** I re-implemented the pre-refactor body and ran a differential over every subset of 9 chain keys (including `default`, `a/*`, `a/b/*`, `a/b/c`, `""`, `a/*/*`) against 7 selectors: **0 mismatches**. `resolve_chain_key` matches the engine's own specificity (`_chain_specificity`, `failover.py:824-833`) because it is literally the same loop, and `resolve_chain` now has exactly one definition of specificity — which was the point.
- **No spec resolution or network on the paint path.** Programmatically scanned the whole 159-line added block (`app.py:11779-11937`): `spec_for_target` and `spec_for_selector` appear **only inside the docstring**, no `fetch_usage`, no `httpx`, no `await`. The claim in the description holds.
- **Credential handling is counts-only.** `row.data` appears only in the `_account_counts` docstring as a prohibition (`app.py:11790`); the loop reads `row.provider` alone (`app.py:11810-11812`). `credential_provider_id` collapses correctly — verified live against the real store: `xai-oauth` + `xai` report as one `xai` row.
- **Guide recipes all run and are network-free.** All four executed against the PR head: `LOP_PY` resolution, `list_login_providers` (16 rows), `static_models` (`openrouter -> {}` as documented), the cascade walk (correctly printed this machine's 5-target `default` chain), and the account-depth count. No provider API touched.
- **Cooldown/bench/quota claims match the engine.** The `primary_retry_at_ms` "set on route CHANGE, not bumped per use" claim matches `FailoverRouteState.activate`'s early return (`failover.py:1329-1345`); `quota_pinned` re-probing regardless of cooldown matches `preflight_usage` (`configure.py:2295-2320`); `target_retry_at_ms` being deliberately absent from `clear()` matches `failover.py:1395-1406`, and "advisory, the loop-back sweep re-walks benched targets" matches `failover.py:1758-1777`. `_recover_quota_blocked` and the per-account model-tier cap match `failover.py:2313-2334`.
- **Echo table and `_FRONTEND_LOCAL_SLASHES`.** Both `ECHO_POLICY` and `PROMPT_POLICY` gained `failovers: False`, which is what `test_slash_echo.py` requires (a missing classification fails, so this is enforced, not decorative). `frontend_state.py:74-77` is right: the command reads local config and local credential counts only.

---

#### major-1 — "no chain" tells the user to set a key that is already set, and hides a chain that really matched

`local_operator/tui/app.py:11891-11902`. `targets` is empty in two very different situations, and the code collapses them into one message that is wrong in the second:

1. no chain key matched at all — the message is correct;
2. a chain key **did** match, but `expand_fallback_targets` suppressed every entry (`failover.py:893-894` drops a legacy string equal to the current selector, and a `provider/*` entry that expands back onto the primary).

Reproduced at head:

```
matched key: default | chain: ['zai/glm-5.3'] | targets: []
/failovers prints -> 'no chain  nothing matches zai/glm-5.3 - set values.retry.fallbackChains.default'
```

This is not contrived: it is the ordinary `default: ["zai/glm-5.3"]` install after the user runs `/model zai/glm-5.3`. Same result for `default: ["anthropic/*"]` on an Anthropic primary. The user is told their `default` key does not match while it *did* match and *is* the key in force, and the suggested fix (set `values.retry.fallbackChains.default`) is a no-op. For a diagnostic command whose entire job is "which key is in force", asserting the opposite of the engine's answer is the failure mode the PR description says the split-out `resolve_chain_key` exists to prevent — and the split is then discarded here because `chain_key` is computed and thrown away on this path.

Fix in one line: branch on `chain_key is None` for the "nothing matches" text, and when `chain_key` is not None but `targets` is empty, say the matched key resolves to no usable route (its only entry is the current model).

#### major-2 — `← serving` is printed on two rows at once for a same-model/different-effort chain entry

`local_operator/tui/app.py:11914-11921`. The marker is a `selector` string comparison, but `FallbackTarget` is `(selector, effort)` and the guide explicitly documents the same-model-different-effort entry as a real route ("The mapping form is what lets a chain re-list the *current* model at a different effort as a real route", `GUIDE.md:139-142`). When such an entry is configured, `serving == primary` **and** `serving == target.selector` are both true.

Driven through the real `OperatorApp` at head:

```
├─ primary  test/model · no accounts   ← serving
├─ matched chain  test/* (wildcard)
├─ 1. test/model  low effort · no accounts   ← serving
SERVING MARKER COUNT: 2
```

Two rows claim to be answering requests, which is exactly the ambiguity the marker exists to remove — and it is worst in the case where the distinction matters most (is the session on high or low effort?). `effective_model` carries `reasoning_effort`, so the comparison can be made exact rather than by selector alone; alternatively mark only the first matching row.

#### major-3 — `_account_detail` reports "no accounts" for providers that are configured and will serve

`local_operator/tui/app.py:11803-11812`. Counts come from `self._providers.credentials()` (the AuthStore) alone. The controller already distinguishes a *stored* credential from a *usable* provider (`controller.py:189-205` `is_usable`: env key, or `allows_missing_api_key`). Verified at head:

```
mistral (MISTRAL_API_KEY set) -> no accounts
ollama  (needs no credential) -> no accounts
```

So a chain entry pointing at a working env-key provider, or a local Ollama, renders as `no accounts` beside a cascade the user is trying to debug. The natural reading of "no accounts" on a diagnostic listing is "this hop will fail", which is the opposite of the truth — and the user's most likely next action is a pointless `login`. `/accounts` can say "no accounts" honestly because it is a credential-store listing; this surface is a *routing* listing and the two questions differ. Suggested fix: keep the count, but render `env key` (or `no credential needed`) instead of `no accounts` when `self._providers.is_usable(provider)` is true with a zero count — no new store scan on the paint path.

#### minor-1 — two of the five advertised degrade states are not tested

The description's table lists five states each "exercised by a test", but only four have one. `tests/unit/tui/test_app_pilot.py:3438-3565` covers populated, no-chain, cascade-off, no-facade, and config-raises. The **session-not-attached** state (`app.py:11851`) has no test — `_failover_text` (`test_app_pilot.py:3422-3431`) deliberately *waits* for attach, so nothing reaches that branch — and **`no model selected yet`** (`app.py:11858`), a sixth state not in the table at all, is untested. Both are two-line tests.

#### minor-2 — the config read is unguarded against a follower session raising

`local_operator/tui/app.py:11855-11857`. `_model_spec`/`_effective_spec` use `getattr(session, "model", None)`, but on `RemoteSession` `model` is a **property that raises** `RuntimeError("owner has no selected model spec")` when the frontend state has no selected model (`session/remote.py:854-858`), and `getattr` with a default does not suppress an exception raised *inside* a property (confirmed by execution). The `_cmd_failovers` try/except turns that into `failover list failed: owner has no selected model spec`, which is an `error` notice for a state the code already has a `warning` phrasing for (`no model selected yet`). Not a crash, but a follower attached before the owner's model syncs gets the wrong severity and a developer-facing string.

#### minor-3 — `matched chain` labels a non-wildcard, non-exact key as "wildcard"

`local_operator/tui/app.py:11908-11911`. The parenthetical is derived by `'exact' if chain_key == primary else 'wildcard'`. Any matched non-`default` key that is not string-equal to the primary is called a wildcard, which is true today only because `_chain_specificity` admits exactly two shapes. It is one added match rule away from printing a confident falsehood. `chain_key.endswith("/*")` is the direct test and costs nothing.

#### minor-4 — the guide's config sample contradicts two shipped defaults

`local_operator/guides/failover/GUIDE.md:116-132` shows `maxRetries: 3` and `usageAwareFallback: true`. The shipped defaults are `maxRetries: 10` and `usageAwareFallback: False` (`local_operator/config.py:163-169`). The block reads as a picture of the default config, and `usageAwareFallback` is the more consequential of the two: the whole "Quota reserve" section (`GUIDE.md:55-73`) describes behaviour that is **gated off by default** — `preflight_usage` returns early at `configure.py:2320` when `usage_aware_fallback` is false, which is where `_apply_account_health` (the lower-effort-same-provider rule) and `_first_available_fallback` (the 0%-skip rule) both live. A user reading the guide will expect quota-reserve routing they do not have. Worth one sentence in that section saying the reserve behaviour requires `usageAwareFallback: true`, plus correcting the sample.

#### nit-1 — the closing hint row renders with an empty name cell

`local_operator/tui/app.py:11925` appends `("", "ask the agent to …")`, so `_tree_listing` (`app.py:16128-16135`) emits `└─ ` then two spaces then the dim detail. It renders acceptably in the captured frames, but every other row in the tree has a name in the string tint and this one is structurally a caption wearing a row's glyph.

#### nit-2 — `primary.split("/")[0]` is used where `parse_selector` exists

`local_operator/tui/app.py:11863` and `11918`. `failover.py` exports `parse_selector` and the module already imports from that namespace; hand-splitting is fine for the two-segment case but diverges for `openrouter/anthropic/claude-opus` style ids if the provider segment ever needs the same normalization routing uses.

---

**Verdict: changes requested.** Three MAJOR findings, all reproduced by execution rather than inferred: major-1 makes the command assert the opposite of the engine's routing answer in an ordinary single-entry-chain configuration, major-2 prints two `← serving` markers for a chain shape the new guide documents as supported, and major-3 reports working providers as `no accounts`. Each is a defect in the *diagnostic's accuracy*, which is the only property this feature has — the surface is otherwise read-only, well-degraded, and correctly kept off the network. No BLOCKER: nothing here loses data or is unsafe, and the `resolve_chain` refactor is proven behaviour-preserving. The five minors and two nits are the highest-value of what I found and are follow-up material except where noted.

damianvtran:

**Scope:** `52a39a02`..`0915b557` (dev-failovers)
**Rounds answered:** code review (3 MAJOR), design review (D1–D7), UX review (U1–U7)

Remediation commits: `659c02b1` (all substantive findings), `0915b557` (pyright narrowing on the new branch).

All gates re-run whole-tree at head: `flake8` clean, `black --check` (26.1.0) clean, `isort --check` clean, `pyright` **0 errors**, unit suite **7392 passed, 7 skipped**. CI green on `0915b557` — every check `SUCCESS`. Frames re-captured at **80 columns** for the two states the original PR never rendered.

---

## Code review

**major-1 — "no chain" contradicts the engine — FIXED** (`659c02b1`)

The two states are now distinct, branching on `chain_key is None` rather than on `not targets`:

| Config | primary | key | targets | now prints |
|---|---|---|---|---|
| `default: ["zai/glm-5.3"]` | `zai/glm-5.3` | `default` | 0 | `default matched but lists only the current model` |
| `default: ["anthropic/*"]` | `anthropic/claude-opus-5` | `default` | 0 | `default matched but lists only the current model` |
| `{"openai/gpt-5": [...]}` | `zai/glm-5.3` | `None` | 0 | `no chain matches — set values.retry.fallbackChains.default` |

The no-op suggestion is gone from the matched case. You were right that this was the split's whole point being discarded — `chain_key` is now the thing the branch reads. Regression test: `test_failovers_matched_key_with_only_the_current_model_is_not_no_match`, which asserts the `set values.retry.fallbackChains.default` string is **absent** on that path.

**major-2 — two `← serving` markers — FIXED** (`659c02b1`)

The comparison now includes effort, and the primary only keeps the marker when no target claimed it, so the count cannot exceed one by construction. Driven through the real `OperatorApp` with your exact repro shape (`default: [{provider, model, effort: low}]`, effective = low):

```
├─ primary                anthropic/claude-opus-5 · high · 4 accounts
├─ matched chain          default
├─ 1. anthropic/claude-opus-5   ← serving  low effort · 4 accounts
```

One marker, and it is on the low-effort route that is actually answering — the "high or low?" question resolved rather than doubled. Test `test_failovers_marks_exactly_one_serving_row` asserts `text.count("← serving") == 1` and that it belongs to the low-effort row. Frame: [after-single-marker.svg](https://raw.githubusercontent.com/damianvtran/local-operator/dev-failovers/docs/assets/pr-failovers/after-single-marker.svg).

**major-3 — working providers render "no accounts" — FIXED** (`659c02b1`)

`_account_detail` now defers a zero count to the registry and `is_usable`, and names the two cases apart because the user's next action differs. Verified live against the real store:

```
mistral   (MISTRAL_API_KEY set) -> env key
ollama    (needs no credential) -> no credential needed
anthropic                       -> 4 accounts
zai                             -> 1 account
```

A non-zero count still prints the number, since rotation depth is what the column is for. A locked store falls through to the honest count rather than inventing a verdict. This also closes **U4**, which was the same defect from the Ollama side. Test: `test_failovers_does_not_call_a_usable_provider_accountless`.

**minor-1 — untested degrade states — FIXED**; the PR table was overstating coverage, so I added the tests rather than correcting the table down. Six states now, each with a test: session-not-attached (`test_failovers_before_the_session_attaches_warns`) and `no model selected yet` (below) were the two missing.

**minor-2 — follower `model` raises — FIXED.** Confirmed by execution that `getattr(obj, "model", None)` does **not** suppress a raise inside a property. Both `_model_spec` and `_effective_spec` calls are now guarded, so the follower gets the `warning` register and the existing `no model selected yet` phrasing instead of `error: failover list failed: owner has no selected model spec`. Test asserts the notice's resolved token is not the error tier.

**minor-3 — "wildcard" label — FIXED.** Now `matched_key.endswith("/*")`, so a future match rule cannot be silently labelled a wildcard.

**minor-4 — guide contradicts shipped defaults — FIXED.** Sample corrected to `maxRetries: 10` with each line marked as shipped-default or not, and `usageAwareFallback: true` explicitly flagged as **not** the default. The "Quota reserve" section now opens by stating that all of it requires `usageAwareFallback: true`, naming the `preflight_usage` early return as the reason. You were right that this was the more consequential of the two — a guide describing routing the reader does not have is worse than no guide.

**nit-1 — hint row's empty name cell — FIXED** as part of D3 (labelled `tip`).
**nit-2 — `parse_selector`** — **FIXED**; both hand-splits replaced.

---

## Design review

**D1 + D2 — marker wraps / `model default effort` repetition — FIXED** (`659c02b1`)

Treated as one defect per your reading. Effort text is now emitted **only when it differs from the model default**, and the marker moved **before** the detail so it is last to wrap rather than first. Your `_removal_detail` citation is exactly the precedent — same surface, same argument, and it should have applied here.

Measured on this machine's real config, fallback #3 serving: **83 cells → 61**. Re-rendered at 70 and 80 columns; the marker stays attached to row 3 at both:

```
├─ 3. alibaba-token-plan/qwen3.8-max   ← serving  3 accounts
```

Frame: [after-fallback-serving-80.svg](https://raw.githubusercontent.com/damianvtran/local-operator/dev-failovers/docs/assets/pr-failovers/after-fallback-serving-80.svg).

**D3 — hint reads as target 6 — FIXED.** The row now carries the name `tip`, putting it in the same class as `primary` and `matched chain` — a named fact about the cascade rather than a numbered hop. I first tried padding the name cell to move the text into the detail column as you suggested; that pushed the row over 80 cells and re-wrapped the thing it was fixing, so the short label is the cheaper form of the same separation (5 cells vs 30+).

**D4 — details do not align — FIXED**, using the `/mcp` `ljust` precedent you cited. Measured detail x after: **488 on every target row** (was 171/244/256/207/488/268/342). The PR body's mock-up now matches what the code renders, so both halves of your finding are closed.

**D5 — empty-state pointer wraps — FIXED.** Shortened to `no chain matches — set values.retry.fallbackChains.default` (the primary is stated one row above, so naming it twice bought nothing). Fits 80 columns with the key intact. Frame: [after-empty-80.svg](https://raw.githubusercontent.com/damianvtran/local-operator/dev-failovers/docs/assets/pr-failovers/after-empty-80.svg).

**D6 — marker shares the detail's colour — DEFERRED.** Real, and your 1.00:1 contrast-against-neighbour measurement is the right way to frame it. Deferring because a distinct tint for an inline marker is a `_tree_listing` capability (it takes one `detail_token` for the whole tree, so per-row emphasis does not exist yet), and adding a styling axis to a shared helper is a change every listing inherits — out of scope for a remediation round. Moving the marker before the detail has meanwhile given it a stable position, which recovers part of the scannability. Ticket: **LOP-D6-failover-marker-tint**.

**D7 — `matched chain` sits in the cascade without reading as metadata — DEFERRED**, same helper constraint as D6, and it is partly mitigated: with `tip` added there are now three labelled metadata rows against numbered entries, which sharpens the numbering's separation. Ticket: **LOP-D7-failover-row-classes**.

---

## UX review

**U1 — the listing shows a cascade the running session will not route on — FIXED** (`659c02b1`)

This was the best finding of the round: the diagnostic was most wrong for the user who followed its own affordance. Fixed at the source rather than by annotation — the listing now reads **the session's captured settings**, not the file:

- `SessionStreamFn.routing_settings` (new, read-only) exposes the mapping the stream actually routes on, with the reason recorded on the property.
- `Session.routing_settings` passes it through rather than storing a second copy, so the two cannot disagree. `None` means "cannot say" and falls back to disk.
- The on-disk copy is still read, but **only to detect divergence**, compared through `RetrySettings` so a `tui.theme` edit does not raise a false flag.

When they differ, a `stale` row names the step that makes the edit real:

```
├─ stale  config.yml changed since this session started · /reload
```

Walked your exact scenario: session snapshot `["zai/glm-5.3"]`, disk `["zai/glm-5.3", "kimi/k3"]` → the listing shows **only** `1. zai/glm-5.3` (what the session will do), plus the stale row. `kimi/k3` is asserted absent. Test: `test_failovers_flags_config_edited_after_the_session_started`. The guide gained a matching paragraph, since it documents the config path that leads here.

**U3 — empty and disabled states drop the affordance — FIXED.** The hint is now emitted in every state, phrased for the state: `ask the agent to set up a fallback cascade` (empty), `ask the agent to turn the cascade back on` (disabled), `ask the agent to change models, effort or providers here` (populated). You were right that this was backwards against the PR's own stated premise. Test: `test_failovers_offers_the_agent_affordance_in_every_state`.

**U2 — `/f` rebind and the missing singular — PARTIALLY REJECTED, with evidence.**

I implemented `aliases=("failover",)` and it broke an unrelated test. The picker sizes its name column on the widest `/name  /alias` pair; `/failovers  /failover` is **21 cells** against a previous max of 18 (`/credential`), which narrows the description column for **every** command at every width — it truncated `List all commands` to `List all com…` on the 41-cell frame that `test_descriptions_come_back_above_the_collapse_width` pins. So the alias is not free: one command's convenience against every command's description. Reverted, with the measurement recorded in a comment at the registration so the next person does not rediscover it. The singular still reaches the command through prefix matching (`/failover` → `failovers`), which is the part that mattered.

On the `/f` rebind: **rejected as correct behaviour.** `/f` is a prefix of `failovers` and is *not* a prefix of `effort`, so prefix-over-fuzzy is the ranking rule working as designed, and the picker shows the row before Enter. Recording it as a deliberate decision rather than special-casing the ranker for one pair — both are now noted in the PR's Review focus.

**U5 — junk arguments swallowed — REJECTED** for this PR, on your own reasoning: `/accounts nonsense` and `/provider nonsense` behave identically on `main`, so changing it here would make `/failovers` the odd one out. It belongs across the listing commands together. Ticket: **LOP-U5-listing-arg-feedback**.

**U6 — no "why" / "when does the primary come back" — DEFERRED**, agreed out of scope: that state lives on `FailoverRouteState` and `RemoteSession` does not carry it, which is why `effective_model` was chosen. Ticket: **LOP-U6-failover-route-reason**.

**U7 — long selectors wrap at 40–60 columns — REJECTED as pre-existing**, confirmed by your own `/skills` capture on unmodified `main`. D2's shortening moved the threshold down as a side effect.

---

### Net effect on the frames

| | before round 1 | now |
|---|---|---|
| Fallback-serving row (real config) | 83 cells, marker wraps below ~89 cols | 61 cells, attached at 70 |
| Target detail x positions | 171, 244, 256, 207, 488, 268, 342 | 488 on every row |
| Serving markers, same-model/different-effort | 2 | 1 |
| `default: ["<current model>"]` | "nothing matches" + no-op fix | names the matched key honestly |
| Provider with env key / no-credential-needed | `no accounts` | `env key` / `no credential needed` |
| Cascade edited mid-session | silently wrong | `stale … /reload` |
| Degrade states with a test | 4 of 6 | 6 of 6 |

Two findings are deferred behind the same `_tree_listing` limitation (D6, D7), two are rejected as house-wide rather than this PR's (U5, U7), and one is rejected on measured cost (U2's alias). Everything else is fixed at head `0915b557`.

damianvtran:
### Agent review — round 2

Reviewer: `reviewer` subagent (lopdev team, independent of the author) — anthropic/claude-opus-5
Scope: `fa62d097ce7007619271d622d6fe0ccb81c3aa83..0915b557e1d7abe5ff60555f1cd870ea3acf2c38`

Head is fresh: `0915b557` is both the PR head and its last commit. I reviewed the remediation diff (`52a39a02..0915b557`) and re-verified every round-1 finding **by driving the real `OperatorApp`**, not by reading the remediation comment. I did not re-litigate what round 1 confirmed correct, except where this diff touched it (`_account_detail`, the detail composition, and the `no chain` branch all changed, so all three were re-checked from scratch).

---

## Round-1 findings: verified fixed

**code major-1 — matched key with no usable targets — CONFIRMED FIXED.** The exact reproduction from round 1 now distinguishes the two states:

```
default: ["test/model"], primary test/model
  -> ├─ no chain  default matched but lists only the current model
test/*: ["test/model"], primary test/model
  -> ├─ no chain  test/* matched but lists only the current model
fallbackChains: {}          (genuinely nothing matches)
  -> ├─ no chain  no chain matches — set values.retry.fallbackChains.default
```

The matched key is now the engine's own answer on both branches, and the no-op suggestion is gone from the case where the key is already set.

**code major-2 — double `← serving` marker — CONFIRMED FIXED.** Marker count is exactly 1 across every case I could construct, and it lands on the right row each time. Using the real mapping schema (`provider`/`model`/`effort` — note the round-1 repro's `{"model": "anthropic/claude-opus-5"}` shape is silently dropped by `_normalize_chain_entry`, so the chain must be written with split keys to exercise this at all):

| chain | serving | result |
|---|---|---|
| primary re-listed @low, then zai | opus-5 @low | 1 marker, on row 1 (the low-effort route) |
| same | opus-5 @high | 1 marker, on **primary** |
| zai@low, zai@high | zai@high | 1 marker, on row 2 |
| zai@low, zai@high | zai@low | 1 marker, on row 1 |
| zai bare, zai@low | zai@low | 1 marker, on row 1 (bare row wins — see minor-1) |

The `not served_by_target` guard on the primary row means the count structurally cannot exceed one, which is the right shape.

**code major-3 — usable providers reported "no accounts" — CONFIRMED FIXED.** Verified against the **real** `ProviderController` + `AuthStore` (empty store, `MISTRAL_API_KEY` set), not the test fake:

```
mistral    allows_missing=False is_usable=True  -> 'env key'
ollama     allows_missing=True  is_usable=True  -> 'no credential needed'
anthropic  allows_missing=False is_usable=False -> 'no accounts'
anthropic, 4 stored -> '4 accounts'   (non-zero still prints depth)
locked store        -> 'no accounts'  (honest fallthrough, no invented verdict)
no facade           -> 'no accounts'
```

I also probed the failure modes the diff's `except Exception` covers: `is_usable` raising with `provider()` succeeding falls through to `no accounts`; `provider()` returning `None` with `is_usable` true gives `env key`. Both correct.

**U1 — listing vs. what the session routes on — CONFIRMED FIXED**, and the drift row is correctly conditioned. Five paths driven:

| case | result |
|---|---|
| disk has 2 targets, session captured 1 | shows **1**, `stale  config.yml changed since this session started · /reload` |
| session snapshot == disk | no stale row |
| unrelated edit (`tui.theme`, `model_name`) after start | **no stale row** — no false fire |
| `routing_settings` attribute absent (older stream) | falls back to disk, no stale row |
| `routing_settings` present but `None` | same |

The no-false-fire property is structural, not incidental: the comparison is `RetrySettings.from_settings(session) != RetrySettings.from_settings(disk)`, and `RetrySettings` is a frozen dataclass over the routing keys only, so a non-routing edit cannot reach it. Correct choice over a raw mapping compare.

**No blocking I/O or paint-path cost added.** The added work per invocation is one `ConfigManager(config_dir())` read plus two `from_settings` parses. Measured on a 40-target chain: **14.7 ms** cold, then **4.4–7.0 ms** — indistinguishable from round 1's 13.1/4.7–11.0 ms, and the composer accepted keystrokes immediately after. No `spec_for_target`, no network, no `await` added.

**D1 / D2 / D4 — CONFIRMED FIXED at 70 and 80 columns.** Captured from the real `OperatorApp` and inspected as rendered images. At **80 cols**, fallback #3 serving:

```
├─ primary  anthropic/claude-opus-5 · high · no accounts
├─ matched chain  default
├─ 1. zai/glm-5.3                     no accounts
├─ 2. kimi/k3                         no accounts
├─ 3. alibaba-token-plan/qwen3.8-max  ← serving  no accounts
├─ 4. xai/grok-4.6                    no accounts
├─ 5. openai/gpt-5.6-sol              no accounts
└─ tip  ask the agent to change models, effort or providers here
```

Marker stays attached to row 3 at both 70 and 80 columns. D4 verified by **measuring the SVG text-node geometry** rather than eyeballing: every target detail starts at `x=488.0` — one column, not the 171/244/256/207/488/268/342 ripple round 1 measured. D5 also fixed: the empty state's config key survives on one line at 80 cols.

**Author's rejections — both sound, and I reproduced the measurement.**

- *No `failover` alias.* Not merely asserted — I added `aliases=("failover",)` to the registry and re-rendered the picker. On the pinned 41-cell frame, `"List all comm" in text` flips **True → False** (`/help` collapses to `List all com…`), so `test_descriptions_come_back_above_the_collapse_width` genuinely fails. The rejection is a measured cost, not a preference.
- *`/f` ranks `failovers` above `effort`.* Confirmed correct prefix behaviour: `/f`, `/fa`, `/fail`, `/failover` all resolve to `/failovers`; `/e` still offers `/exit`+`/effort` and `/ef` resolves to `/effort`. The singular does reach the command, so the alias buys little.

**New surface area (`SessionStreamFn.routing_settings` / `Session.routing_settings`) — correct on both roles.** The `Session` property is a `getattr` pass-through to the stream rather than a second stored copy, so the two cannot disagree; `SessionStreamFn` returns `self._settings` guarded by `isinstance(..., Mapping)`, and that is the same object `session_factory.py:1145-1148` hands `create_stream_fn` and the same one `from_settings` re-parses at `configure.py:1977,2321`. So the listing reads the identical mapping the router does — not a copy that could drift. On a **follower**, `RemoteSession` has no `routing_settings`, `getattr` returns `None`, and the code treats that as "cannot say" (falls back to disk, suppresses the stale row) rather than as "empty" — which is right, since a follower cannot know the owner's snapshot and a stale row there would be a guess. Both read-only; nothing mutates routing.

---

#### minor-1 — a chain entry re-listing the current model at the model's own default effort takes the marker off the primary

`local_operator/tui/app.py:12053-12058`. The serving test is `(target.effort or "") == (serving_effort if target.effort else "")`, which for a target with an effort reduces to `target.effort == serving_effort`. That is right in general, but it cannot separate "the fallback route at effort X is pinned" from "nothing failed over and the primary happens to run at effort X", because `effective_model` is *equal* to the primary spec in the second case.

Reproduced with nothing failed over (`effective_model is model`) and a chain whose first entry re-lists the primary at its own default effort — `high` for `claude-opus-5`, so this needs no unusual value:

```
├─ primary  anthropic/claude-opus-5 · high · no accounts
├─ matched chain  default
├─ 1. anthropic/claude-opus-5  ← serving  high effort · no accounts
```

The marker sits on the fallback row while the session is actually on the primary and has never failed over. Also reachable via the wildcard form `{provider: anthropic, model: "*", effort: high}`, which `expand_fallback_targets` expands back onto the primary selector (the same-selector suppression at `failover.py:911` only drops entries with `effort is None`).

Severity MINOR, not MAJOR: it needs a chain that deliberately re-lists the current model at an effort equal to the one in force, the misreport is one row off inside the same model and effort, and no routing behaviour is affected. It is strictly narrower than the round-1 major it replaced (which double-marked, and did so on ordinary configs). Worth recording because the round-1 remediation traded a two-marker bug for a one-marker-in-the-wrong-place bug in the same neighbourhood.

One-line fix if you want it: the primary is serving whenever `session.active_fallback is None`, so gate `served_by_target` on that — but note `RemoteSession` does not implement `active_fallback`, which is exactly why `effective_model` was chosen, so this needs a `getattr` guard and is a genuine follow-up rather than a quick edit. Not worth blocking on.

#### nit-1 — `stale` row says `/reload` without saying it restarts the conversation

`app.py:12094`. `config.yml changed since this session started · /reload` is accurate and correctly scoped to routing keys. `/reload` relaunches the conversation on the current install, which is heavier than the row's terse phrasing implies for a user who just wanted their cascade edit to land. The guide covers it (`GUIDE.md:153-158`); recording only because this row is the one place a user meets the instruction without that context.

---

### Gates and CI

Run over the whole tree in a clean worktree at `0915b557`, exactly as CI does:

- `flake8 .` — **clean**
- `black --check .` (26.1.0) — **clean**, 569 files unchanged
- `isort --check .` (5.13.2) — **clean**
- `tests/unit/tui` + `tests/unit/providers/test_failover.py` + `tests/unit/guides` — **2817 passed, 4 skipped**, plus 2 failures (`test_subagent_band_does_not_crush_the_transcript`, `test_band_total_is_own_spend_plus_children`). Both pass in isolation and are **order-dependent flakes pre-existing on the base**: the same suite on the merge-base `fa62d097` fails 2 of its own (`test_cost_aggregation`, `test_composer_seam`), a different set each run. Not introduced by this PR.
- The 13 `/failovers` pilot tests — **all pass**. I did not re-run `pyright` locally (it exceeded my time budget on this machine); CI's `type-check` job is green at head, which is the authoritative signal.

**All 14 CI checks are SUCCESS at `0915b557`** — `lint`, `type-check`, `test (3.12)`, `test (3.13)`, `cli-sanity`, `server-sanity`, `pip-audit`.

---

### Verdict

**CLEAN — no BLOCKER, no MAJOR.** This round is **TERMINAL** for code review. All three round-1 code majors, U1, and D1/D2/D4/D5 are verified fixed by execution rather than by assertion, the two rejections are backed by measurements I reproduced independently, and the new `routing_settings` surface is correct on both the owner and an attached follower.

`minor-1` and `nit-1` are follow-ups, not grounds for another round. From the code-review side this is ready to merge once the design and UX rounds are clean and fresh against `0915b557`.

damianvtran:
### UX review — round 2

**Reviewer:** `ux-reviewer` subagent (lopdev team) — anthropic/claude-opus-5
**Scope:** `fa62d097ce7007619271d622d6fe0ccb81c3aa83..0915b557e1d7abe5ff60555f1cd870ea3acf2c38`

Head is fresh: `0915b557` is both the PR head and its last commit. I re-walked the flow against the **real `OperatorApp`** under `run_test` at this head in a detached worktree, not against the remediation comment and not from the round-1 stills. Every frame below is one I captured and rendered myself.

The round-1 MAJOR was U1: the listing read `config.yml` while the router held a session-lifetime snapshot, so a user who followed the command's own `tip` row got a green light for a cascade the running session would never route on. That is the finding this round exists to settle, so I walked the **whole loop** rather than just checking that a row appears.

---

## U1 — the false green light — VERIFIED FIXED

I drove the exact sequence a user lives, in one continuous session:

**Step 1** — session starts on a two-hop cascade, user runs `/failovers`:

```
failover cascade
├─ primary  test/model · no accounts   ← serving
├─ matched chain  default
├─ 1. zai/glm-5.3  no accounts
├─ 2. kimi/k3      no accounts
└─ tip  ask the agent to change models, effort or providers here
```

**Step 2** — the user takes the `tip` up on its offer and asks the agent to change the cascade. The agent writes `config.yml` (which is exactly what `guide://failover` now instructs it to do, so this is the supported path, not a hand-edit).

**Step 3** — the user re-runs `/failovers` to confirm the change landed. This is the moment that used to lie:

```
failover cascade
├─ primary  test/model · no accounts   ← serving
├─ matched chain  default
├─ 1. zai/glm-5.3  no accounts          <- the SESSION's chain, not the file's
├─ 2. kimi/k3      no accounts
├─ stale  config.yml changed since this session started · /reload
└─ tip  ask the agent to change models, effort or providers here
```

Both halves are right, and both matter. The **targets are the session's snapshot** (`openai/gpt-5.6-sol` and `anthropic/claude-opus-5` were on disk at this point and correctly do **not** appear), so the listing no longer asserts a cascade that is not routing. And the **`stale` row names the divergence and the remedy on the same line**. The user is told the truth: the edit exists, this session is not using it, `/reload` is what makes it real.

**The remedy closes the loop.** `/reload` re-execs the process, which no pilot can perform, so I verified the user-visible outcome it produces — a session built after the edit:

```
failover cascade
├─ primary  test/model · no accounts   ← serving
├─ matched chain  default
├─ 1. openai/gpt-5.6-sol       no accounts
├─ 2. anthropic/claude-opus-5  no accounts
└─ tip  ask the agent to change models, effort or providers here
```

New chain, and the `stale` row is gone. So the row is not a dead-end warning: following it produces the state it promises.

**Wording holds up for someone who does not know the internals.** `config.yml changed since this session started · /reload` states the fact (the file moved), the scope (since *this session* started, which is why the listing above it looks old), and the action, in that order. It never says "snapshot", "router", or "settings mapping". The one internals word is the filename, and that is the right one to use because it is the thing the agent just edited on the user's behalf and the thing they would open to check.

**It is a comparison, not a latch, and it is scoped to routing.** I probed the false-positive surface, since a warning that cries wolf gets trained away:

| Step | `stale` row |
|---|---|
| Nothing changed | no |
| `tui_theme` edit (non-routing) | no |
| `hosting`/`model_name` written to disk | no |
| A routing edit | **yes** |
| The routing edit undone again | **no** (clears) |
| Host that cannot report `routing_settings` | no (falls back to disk, correctly says nothing) |

The undo case is the one I most expected to be wrong and it is right: a user who reverts a bad edit is not left permanently nagged to `/reload`. And the `hosting`/`model_name` case is correct rather than a gap — the `primary` row reads the **live session**, so a disk model edit cannot make that row lie in the first place. I confirmed both directions: with `openai/gpt-5` on disk and the session still on `test/model`, the row says `test/model`; once the session model actually changes, it says `openai/gpt-5`. There is nothing for a drift flag to add there.

**No freeze, and drift detection is free.** On a 40-target chain, median `5.4 ms` with no drift and median `5.4 ms` with drift active — the extra `config.yml` read and `RetrySettings` comparison cost **+0.0 ms median** (max fell from 12.3 ms to 5.9 ms, i.e. inside noise). The composer accepted `hello` typed immediately after the drifted call. Resized 100 → 72 → 60 → 50 columns with the stale row on screen: `virtual_size` tracked `size` at every step and no vertical scrollbar appeared.

## U3 — affordance in every state — VERIFIED FIXED

The `tip` row is present in all four states, and the drift row composes with each rather than replacing it. The state that matters most is the first-time user whose agent has just written them a cascade:

```
failover cascade
├─ primary  test/model · no accounts
├─ no chain  no chain matches — set values.retry.fallbackChains.default
├─ stale  config.yml changed since this session started · /reload
└─ tip  ask the agent to set up a fallback cascade
```

Without the drift row this frame would be actively demoralising — the user asks for a cascade, the agent says it made one, and the command replies "no chain matches". With it, the contradiction is explained on screen. Also confirmed: the disabled state (`retry.modelFallback is false` + `tip ask the agent to turn the cascade back on`), the true empty state, and `default matched but lists only the current model`. The hint text is state-appropriate in each, not one generic string.

## U2 — the rejected `/failover` alias — ACCEPTED, no usability hole

I typed `/failover` a character at a time into the real composer and pressed Enter: the cascade renders. The picker filters to a single match from `/f` onward (`/f`, `/fa`, `/fail`, `/failover` all resolve to exactly one command), so the singular reaches the command on the path a user actually takes.

I did find the one path where it does not: dismiss the picker with Esc first, then Enter, and you get `! unknown command: /failover — try /help`. That is **not** a defect in this PR — `/account`, `/guide` and `/session` behave identically on the same path, so it is the product's existing prefix-resolution model, and introducing an alias here would fix it for one command while leaving the pattern inconsistent everywhere else. The rejection is correct, and the error copy is recoverable and points at `/help`. Leaving `/f` unbound is likewise right: it is the sole `/f*` command, so the prefix already belongs to it without spending a one-letter binding.

## Discoverability

`/failovers` sits next to `/accounts` and `/usage` in the command list — the right neighbourhood, since it answers the third question in that family. Description reads `Show the model failover cascade and what is serving`, which names both halves of what it does in user vocabulary. A user who never reads the diff finds it by typing `/f`.

---

## Follow-ups (not blocking)

- **U8 — NIT — the `stale` row wraps below ~72 columns.** At 60 and 50 columns `· /reload` falls to a continuation line that loses the tree indent. The remedy is still legible and still on screen, and the `tip` row wraps at the same widths, so this is the row set's existing degrade rather than a new one. Worth a shorter phrasing if the row is ever revisited, e.g. dropping `config.yml` to `config changed since this session started · /reload` (56 cells, fits to ~62 columns).
- **U9 — NIT — the drift check is `RetrySettings`-wide, not cascade-only.** Changing `maxRetries` from 10 to 7 raises the `stale` row even though the displayed cascade is unaffected. This is defensible (every field it compares genuinely does not reach the running session, so `/reload` is genuinely the fix) and it is the safe direction to err in, but a user who edited only a retry count may be briefly puzzled by a cascade listing that flags itself. Follow-up at most.
- Round-1 U6 (why did it fall back / when do I get the primary back) is answered by `guide://failover`, which now also documents the snapshot rule and `/reload` explicitly. Round-1 U7 (long selectors wrapping at narrow widths) is unchanged and remains a follow-up.

## Verdict

**CLEAN — no BLOCKER, no MAJOR. This round is TERMINAL.**

U1 is genuinely fixed, verified by walking the whole loop rather than by confirming a row exists: the listing now reports the session's routing, the divergence is named in the user's vocabulary, the remedy is on the same line, following it produces the promised state, and the flag clears when the divergence does. U3 is fixed in all four states. The rejected alias and `/f` binding leave no real usability hole. Nothing in the new stale-detection path is perceptible in latency and the input loop never stalls. Remaining items (U8, U9, plus round-1 U6/U7) are NIT-level follow-ups.

damianvtran:
### Design review — round 2

**Reviewer:** `designer` subagent (lopdev team) — anthropic/claude-opus-5
**Scope:** `52a39a02`..`0915b557` (dev-failovers), re-judged whole-surface against round 1
**Method:** fresh worktree at `0915b557`, real `OperatorApp` via `run_test` + `app.save_screenshot` per AGENTS.md "Visual validation". I did **not** use the PR's committed SVGs. 30 frames captured across 7 states (populated/primary-serving, fallback-serving at positions 3 and 5, empty, cascade-off, stale-drift, stale+empty, long-ids, same-model-two-efforts) at 56/60/62–68/70/72/74/76/78/80/100/120 columns, plus a matching BEFORE set captured from a second worktree at `52a39a02`. Geometry measured off the SVG text nodes (12.2px/cell, transcript x0=48.8).

Every round-1 finding I raised is fixed, and the fixes are real in the frame rather than only in the diff. Verification of each, then what I found judging the new frames cold.

---

#### D1 — MAJOR — FIXED

The marker no longer detaches from its row in the case that matters. Measured at `0915b557`, fallback #3 serving on the operator's real 5-entry chain:

```
├─ 3. alibaba-token-plan/qwen3.8-max  ← serving  no accounts
```

That row is **60 cells**, down from 83 at `52a39a02`. Rendered at 70/76/80/100/120 columns it is one unbroken line every time. The BEFORE frame at the same 80 columns is the round-1 defect exactly as reported:

```
├─ 3. alibaba-token-plan/qwen3.8-max  model default effort · no accounts
← serving
```

Bisecting the wrap boundary on captured frames (usable text width is `cols − 6`, confirmed against five boundary widths):

| Row | Cells | Wraps below |
|---|---|---|
| Fallback #3 serving (real cfg) | 60 | 66 cols (was ~89) |
| Primary serving (real cfg) | 61 | 67 cols (was ~73) |
| `tip` row | 64 | 70 cols |

The 80-column split pane that motivated the finding is now clear by 20 and 19 cells respectively. Moving the marker ahead of the detail also means that when the row *does* finally wrap (60 columns and below), what orphans is `accounts` — a word that still reads — instead of an arrow pointing at nothing. That was the actual ask, and it is satisfied.

#### D2 — MAJOR — FIXED

`model default effort` is gone from the frames. On the real config the five target rows now carry only `no accounts`, and the effort clause appears only when the chain entry sets one (`low effort · no accounts`), which is the case where it carries a bit. Row widths on the real config: 5 of 5 target rows dropped 53–72 cells → 49 cells. This is what paid for both D1 and D4.

#### D4 — MINOR — FIXED

Details now form a column. Measured detail x-positions, populated fallback-serving frame at 80 columns:

```
488.0  1. zai/glm-5.3
488.0  2. kimi/k3
488.0  3. alibaba-token-plan/qwen3.8-max   (← serving  no accounts)
488.0  4. xai/grok-4.6
488.0  5. openai/gpt-5.6-sol
```

All five at **x=488.0**, versus round 1's `171, 244, 256, 207, 488, 268, 342`. Holds at 70/76/80/100/120 and in the effort chain (all three at x=402.6) and the long-id chain (all three at x=695.4). The `/mcp` precedent is followed properly: the pad is computed over the numbered target rows only, so `primary`/`matched chain`/`stale`/`tip` keep their natural short offsets and do not inherit a 33-cell gutter. That was the right call — padding them too would have re-wrapped the rows the round was fixing.

#### D3 / D5 — MINOR — FIXED

`tip` reads as a label, not as target 6. The escape hatch now sits at x=122 with a green `tip` name in the same class as `primary`, `matched chain` and `stale`, instead of its text landing at x=85.4 — the exact name-column x of every cascade row above it, which was the whole complaint. Judged as a reader: the numbered run `1.`–`5.` is now visibly a different kind of thing from the four labelled rows, and the `└─` terminal glyph no longer reads as "last hop".

The empty state is materially better. `94 cells → 71`, so at 80 columns the config key survives on one line where before it wrapped to a bare `values.retry.fallbackChains.default` at column 0:

```
├─ primary   anthropic/claude-opus-5 · no accounts
├─ no chain  no chain matches — set values.retry.fallbackChains.default
└─ tip       ask the agent to set up a fallback cascade
```

Dropping the redundant restatement of the primary (it is on the row above) is the right 23 cells to cut. The state also gained the `tip` row it previously lacked, so the default install — the most likely caller — now ends on an action instead of a dead end.

---

### Judging the new frames cold

I re-read all seven states as a first-time reader, looking for problems the remediation introduced rather than checking its homework.

**The marker-before-detail ordering reads naturally.** `3. alibaba-token-plan/qwen3.8-max  ← serving  no accounts` parses as row → status → attribute, and because the marker starts at the fixed detail column it is now findable by vertical position rather than by scanning to end-of-line. That is a genuine scannability gain over round 1, not just a wrap fix, and it partly compensates for D6 remaining open.

**`tip` reads as a label.** Four-row metadata vocabulary — `primary`, `matched chain`, `stale`, `tip` — is coherent, all lowercase, all short, all green. It does not read as noise.

**Padded columns hold at narrow widths and with long names**, with one caveat recorded as D8 below.

**No reflow, no scrollbar.** I captured consecutive frames (`pause` → save → `pause` → save) in all 30 captures and diffed them: byte-identical in every state, so nothing settles after paint. `screen.virtual_size == screen.size` and `show_vertical_scrollbar == False` in all 30, so the listing never introduces the silent two-cell width loss AGENTS.md warns about.

**Two states the PR never rendered, both fine.** `stale` + populated puts the drift row between the cascade and the tip where it reads as a footnote about the listing rather than a hop; `stale` + empty keeps both the drift row and the affordance. The cascade-off state is unchanged and still names the key.

---

#### D8 — MINOR — The name pad is sized by the longest entry, so one long model id widens the gutter for the whole chain and costs width the serving row then wraps on

`local_operator/tui/app.py:12069`

`name_column = max(len(name) ...)` is unconditional, so the pad is as wide as the worst row. With one long id in the chain, captured at head, 80 columns:

```
├─ 1. alibaba-token-plan/qwen3.8-max-thinking-2026-04  no accounts
├─ 2. zai/glm-5.3                                      no accounts
├─ 3. openrouter/deepseek/deepseek-reasoner-lite       ← serving  1
account
```

Row 2 is `zai/glm-5.3` — 14 cells of content in a 66-cell row. The serving row is **75 cells padded vs 70 unpadded**, so the padding is what pushes it over the 74-cell budget at 80 columns and orphans `account`. Alignment bought scannability and then spent 5 cells of it on the one row that must not wrap.

MINOR, not MAJOR, because it needs an unusually long id to bite (the operator's real config is clear by 20 cells, and `account` is a readable orphan, not a detached arrow), and because the same frame at 100 columns is clean. But it is the one place where the D4 fix made a frame worse than it was, and the fix is small and local: cap the pad, or skip padding when the padded worst row would exceed the available width — the shape `_removal_detail` already reasons about at `app.py:11895`.

#### D9 — NIT — Effort vocabulary differs between the primary row and the target rows

The primary prints the bare value, the targets print the value plus the noun:

```
├─ primary  anthropic/claude-opus-5 · high · no accounts
├─ 1. anthropic/claude-opus-5  ← serving  low effort · no accounts
```

`high` vs `low effort`, same attribute, same tree, two spellings. Pre-existing (`52a39a02` had it too), not introduced here, and only visible when a non-default effort is set. Recording it because this is the frame where the two sit three rows apart and the comparison is the reader's whole question.

#### D10 — NIT — `no accounts` still dominates a real-config frame

With no stored credentials, `no accounts` appears on 6 of 8 rows and is the only thing in the detail column. `_account_detail`'s `is_usable` deferral (from the code round) fixes the *honesty* problem, and it does not fire here because the fake controller reports these providers unusable. Worth knowing that the column still degenerates to one repeated string on a fresh install — the same shape D2 objected to, one layer down. Not actionable in this PR: unlike `model default effort`, this string is sometimes false, so it carries a bit.

---

#### Deferred D6 / D7 — confirmed genuine follow-ups, not made worse

**D6 (marker shares the detail's colour)** — re-measured at head: the marker is still `r3` `#837c6d`, identical fill to the detail text beside it. Contrast against background 4.55:1 (fine), against the neighbouring text it must be distinguished from **1.00:1** (unchanged). The deferral reason holds on inspection: `_tree_listing` takes a single `detail_token` for the entire tree (`app.py:16256,16282`) and appends the whole detail as one styled run, so per-row or intra-row emphasis does not exist as a capability — adding it changes a helper `/mcp`, `/accounts`, `/provider` and the next-request estimate all inherit. Correctly out of scope for a remediation round. **And the remediation improved the situation rather than worsening it**: the marker now sits at a fixed x (488.0 in the real-config frame) instead of at a ragged end-of-line, so it is locatable by position while the tint is pending. Legitimate follow-up.

**D7 (`matched chain` does not read as metadata)** — the remediation's claim that adding `tip` sharpens the separation is correct in the frame, and I would go further: with `primary`/`matched chain`/`stale`/`tip` now forming a four-row labelled vocabulary against a numbered `1.`–`5.` run whose details are pinned to a shared column, the two classes read apart clearly enough that D7 is now closer to a NIT than when I filed it. Same `_tree_listing` constraint. Genuine follow-up, and lower priority than when raised.

Neither deferral is doing work the remediation should have done inline.

---

### Verdict

**CLEAN — no BLOCKER and no MAJOR findings. This round is TERMINAL.**

D1, D2, D4, D3 and D5 are all fixed and verified in frames I captured myself at head `0915b557`, with before/after pairs from `52a39a02` confirming each change is the one claimed. Both MAJOR findings from round 1 are closed; the two rejections (singular alias, `/f` rebinding) and the two deferrals (D6, D7) are sound and D6/D7 are genuinely follow-ups.

Open items are D8 (MINOR), D9 and D10 (NITs). None blocks the merge gate. D8 is the only one I would actually queue: it is a five-cell regression in one uncommon configuration, introduced by an otherwise good fix, and it is cheap to bound. Recording it as a follow-up ticket rather than a remediation round is the proportionate call.

The surface is better than it was in round 1 by every measure I can put a number on: the widest real-config row went 83 → 60 cells, the detail column went from 7 scattered x-positions to 1, the empty state went 94 → 71 cells, and the escape hatch stopped impersonating a route.

Issue #367, opened 2026-08-28
[Bug] MCP: Codex CLI config (~/.codex/config.toml) is not imported


bbqben:
## Brief Summary

`load_mcp_configs` imports MCP servers from Claude Code, Cursor and VS Code, but not from Codex CLI. Servers configured in `~/.codex/config.toml` are invisible to Local Operator.

## Current behaviour

`local_operator/mcp/config.py` builds its candidate list as:

```python
candidates: list[tuple[Path, tuple[str, ...] | None]] = [
    (project_path, None),                              # <cwd>/.local-operator/mcp.json
    (project_dot_path, None),                          # <cwd>/.mcp.json
    (user_path, None),                                 # ~/.local-operator/mcp.json
    (home / ".claude.json", None),
    (root / ".claude" / ".mcp.json", None),
    (home / ".cursor" / "mcp.json", None),
    (root / ".vscode" / "mcp.json", ("mcp", "servers")),
]
```

Codex CLI is absent. The module docstring lists the same four foreign tools, so this looks like an omission rather than a deliberate exclusion.

## Reproduction

With five servers defined in `~/.codex/config.toml`:

```
$ grep -c '^\[mcp_servers\.' ~/.codex/config.toml
5

$ lop mcp list
╭─ MCP Servers ──────────────────────────────╮
│ Playwright: npx @playwright/mcp@latest     │
│ context7: https://mcp.context7.com/mcp     │
╰────────────────────────────────────────────╯
```

Only the two servers from Local Operator's own config appear. None of the Codex-configured servers are picked up.

## Expected behaviour

`~/.codex/config.toml` is read as a best-effort import source alongside the existing four, at the same priority tier (after the local-operator files, so it can never override them), with its servers listed by `lop mcp list` and usable in a session.

## Notes

- Codex's config is **TOML**, not JSON, and its servers live under `[mcp_servers.<name>]` rather than a `mcpServers` object. Every existing import source is JSON via `_read_json`, so this needs a small TOML reader (`tomllib` is stdlib on the supported Python versions) plus a key path of `("mcp_servers",)`, rather than another entry in the same list.
- Codex entries use `command` / `args` / `env` keys, which map cleanly onto the existing stdio server shape.
- Codex also supports project-scoped config; user scope alone would cover the common case and is the smaller change.

## Environment

- Local Operator v0.42.13
- macOS 15 (Darwin 25.6.0, arm64), Python 3.13

bbqben:
[...]

PR #377, opened 2026-08-28
feat(tui): complete the /mcp verb set with list, add and remove

### LINKED ISSUE #368: [FEATURE] MCP UX: inline ghost-text completion and explicit /mcp list|add|remove
## Feature Summary

Bring the `/mcp` surface closer to what OMP offers: ghost-text preview of the completion that would be accepted, and explicit `list` / `add` / `remove` subcommands in the TUI so the command is discoverable and complete rather than OAuth-only.

### 1. Ghost text for autocomplete

The slash picker (`tui/widgets/command_picker.py`) shows a two-column suggestion list under the composer, but the composer itself never previews the completion. OMP renders the remainder of the highlighted suggestion inline, dimmed, ahead of the caret — so the user can see exactly what pressing Tab will produce without reading a separate list and mapping the highlighted row back onto their own buffer.

This matters most for the compound rows `/mcp` already builds. `_mcp_argument_choices` deliberately offers server rows as compound names (`login notion`) because the matcher compares against the whole argument. The picker shows `login notion`, but the buffer still shows only what was typed, so the relationship between the highlighted row and the resulting command has to be inferred. Ghost text closes that gap:

```
/mcp login n␣otion          ← "otion" dimmed, inline
```

The existing pieces are all there — `picker.suggestions()`, the highlight index, and `Editor` already intercepts keys in `_on_key` before `TextArea` bindings — so this is a rendering addition rather than new completion machinery.

### 2. More verbose `/mcp` subcommands

`/mcp` currently accepts only `login`, `logout` and `reauth`. Anything else is rejected:

```
unknown mcp subcommand: <x> — try /mcp login|logout|reauth <name>
```

Bare `/mcp` prints the server block, which means "show me the servers" is spelled by the *absence* of an argument while every other operation is spelled by a verb. That asymmetry is the confusing part, and it is the same shape as the `/model` vs `/model default` problem filed separately.

Proposed:

- **`/mcp list`** — an explicit alias for what bare `/mcp` already does. Costs nothing, makes the verb set self-describing, and gives the argument picker a non-destructive first row to land on.
- **`/mcp add <name> <command|url>`** — the in-TUI twin of `local-operator mcp add`, writing to `~/.local-operator/mcp.json` (or the project file), then reconnecting.
- **`/mcp remove <name>`** — the twin of `local-operator mcp remove`, scoped to the file that actually defines the server, with a clear refusal when the server came from an imported foreign config (`~/.claude.json`, `~/.cursor/mcp.json`, …) rather than silently doing nothing.

The CLI already has all three (`lop mcp {list,add,remove,login,logout,reauth}`), so this is about parity between the CLI and the TUI, not new capability. `_mcp_argument_choices` is the single place the verb rows are declared, and `_cmd_mcp`'s `sub not in (...)` guard is the single place they are validated.

## Motivation and Use Case

Adding an MCP server today means leaving the session, running `lop mcp add`, and coming back — or editing JSON by hand. Since `/mcp login` already exists precisely because startup is deliberately non-interactive and a user needs an in-session way to fix a server, the same argument applies to adding and removing one.

For ghost text: the `/mcp` compound-name rows are the hardest completion in the app to predict from the list alone, so it is the surface that benefits most, but the win is general across every slash command.

## Additional Context

Comparison drawn against OMP (oh-my-pi), which renders inline dimmed completion ahead of the caret and exposes a fuller verb set on its equivalent surfaces. Both suggestions are independent and could ship separately.

Relevant code:
- `local_operator/tui/widgets/command_picker.py` — suggestion list
- `local_operator/tui/widgets/editor.py` — `_on_key`, picker integration
- `local_operator/tui/app.py` — `_cmd_mcp`, `_mcp_argument_choices`, `SLASH_COMMANDS["mcp"]`

Environment: Local Operator v0.42.13, macOS 15 (arm64), Python 3.13.

damianvtran:
# PR Description

## Summary of Changes

Complete the `/mcp` verb set in the TUI so it matches the CLI, and make the
argument picker show the servers each verb can actually act on. PR 2 of 2 for
#368 (PR 1 is the ghost-text half; this one is the verb set and server
suggestions).

Two defects, both verified against `origin/main`:

1. **`/mcp` accepted only `login|logout|reauth`.** `lop mcp {list,add,remove,login,logout,reauth}` has existed in the CLI since the rewrite (`cli.py:554-625`), so a user who knows the CLI, or simply guesses, typed `/mcp list` and got `unknown mcp subcommand: list — try /mcp login|logout|reauth <name>` — a hostile error for a correctly-formed command.

2. **The argument picker was OAuth-only.** `_mcp_argument_choices` built its server rows from `oauth_server_names()`, so on a real config a stdio server and two non-OAuth http servers were **completely invisible** to the picker, including under verbs that can act on them.

### What changed

- **Verb set.** `list`, `add`, `remove`, `login`, `logout`, `reauth`, declared once as `OperatorApp.MCP_SUBCOMMANDS` and used by both the picker rows and the `_cmd_mcp` guard, so the validated set and the offered set cannot drift. `list` leads the picker as the safe, non-destructive landing row.

  `/mcp list` is an explicit alias for what bare `/mcp` already does, routed through the *same* block builder rather than a second listing. Reserving `list` is safe **here** because `/mcp`'s first token is a **closed** verb set — nothing can be named `list` in that slot. `/team` and `/agent` take an **open** namespace (`teams.py`'s `_NAME_RE` permits a team named `list`), so this fix must not be copy-pasted onto them; there the safe shape is inverted precedence. That reasoning is recorded on `MCP_SUBCOMMANDS` so the next person does not generalise it.

- **Per-verb server rows.** `login`/`reauth` offer OAuth servers; `logout` keeps the existing "OAuth **and** holds a stored credential" rule; `remove` offers **every configured server** read from the config layer (the thing it edits) rather than the manager, so it still answers when the manager is absent or the session is a remote follower; `add` and `list` offer no rows.

- **`remove` rows show their source file** in the detail column (`~/.local-operator/mcp.json` vs `~/.claude.json`). This is the design point, not decoration: it turns the refusal below into something the user saw coming instead of an error they ran into.

- **`/mcp add <name> <url|command [args…]>`.** Discriminates on whether the third token parses as `http(s)://`. No scope token — always the global `~/.local-operator/mcp.json`, matching the CLI default, and the receipt **names the file** so an invisible default becomes a checkable fact. **OAuth is not inferred from a URL**: real configs carry non-OAuth http servers, and inferring auth would silently change how a server authenticates. The receipt points at `lop mcp add <name> --url <url> --oauth`, the only path that writes the `auth` block (corrected in round 1, M3 — it used to suggest `/mcp login <name>`, which `_resolve_mcp_server` rejects for exactly the servers this command creates). It also refuses a name already defined by a higher-priority or foreign config, so it cannot shadow a Claude Code entry or print "added" for a write the user will never observe (round 1, M1). Env vars stay on the CLI (`--env`), because a `KEY=VALUE` token is indistinguishable from a command argument in this grammar.

- **`/mcp remove` refusal.** Resolves the defining file via the provenance `load_all_mcp_configs` already returns and refuses unless it is a file local-operator writes. Comparison is on **resolved** paths, so a symlinked home or a `/private/var` prefix cannot turn an owned file into a refusal. `<cwd>/.mcp.json` is refused too — read by the loader, never written by `_scope_path`. The message names the file, the tool and the alternative:

  ```
  'notion' is defined in ~/.claude.json (imported from Claude Code), not by local-operator.
  Remove it there.
  ```

  No `/mcp disable` suggestion, because no such verb exists. A Codex TOML source (#367) gets the same shape, with a comment recording that `tomllib` is read-only (`load`/`loads`; `tomli_w` is not a dependency) so refusal there is the only correct behaviour rather than a policy choice.

- **Blocking refactor (`mcp/config.py`).** `add_server`/`remove_server` were CLI-shaped: they `print(...)` to stderr and returned `int` exit codes. Calling them from the TUI would write to the terminal *underneath* the Textual frame and corrupt the display. They now raise `MCPConfigWriteError` (keeping every reason as a list, so the CLI still prints one `error:` line per problem) and return the `Path` written. One implementation, two callers; the CLI adapts at its own boundary. The write logic is **not** duplicated into the TUI — two writers to one config file is the worse option.

## Testing

Real execution against a **temp HOME** (`mktemp -d`) throughout, so my own MCP config was never read or written; synthetic configs carry no secrets. Harnesses are committed under `docs/evidence/mcp-verbs/` so every claim here can be re-run rather than trusted.

**CLI unchanged after the refactor** (`cli_ab.txt`). Ran `lop mcp list|add|remove` plus every error path against this branch and against a clean `origin/main` worktree, normalised the temp paths, and diffed: **byte-identical stderr text and exit codes**, including `add` duplicate / both-transports / neither / invalid-name / oauth-on-stdio (all `exit=1`) and `remove` not-found (`exit=1`).

**TUI paths** (`tui_check.txt`) — real `OperatorApp` driven through `_type_command`:

```
=== /mcp add demo-stdio npx -y demo-mcp
✓ added MCP server 'demo-stdio' to ~/.local-operator/mcp.json
=== /mcp add demo-http https://demo.example/mcp
✓ added MCP server 'demo-http' to ~/.local-operator/mcp.json — needs OAuth? re-add it with: lop mcp add demo-http --url https://demo.example/mcp --oauth
=== /mcp remove notion
! 'notion' is defined in ~/.claude.json (imported from Claude Code), not by local-operator.
Remove it there.
=== /mcp frobnicate
! unknown mcp subcommand: frobnicate — try /mcp list|add|remove|login|logout|reauth <name>
```

The resulting `mcp.json` is captured in full: `demo-stdio` written as `{"type":"stdio","command":"npx","args":["-y","demo-mcp"]}`, `demo-http` as `{"type":"http","url":"…"}` with **no** `auth` block. After the `notion` refusal both `~/.claude.json` and the owned config are byte-unchanged.

**Visual** — captured from the real `OperatorApp` per AGENTS.md "Visual validation" and viewed as rendered frames, not markup.

The `/mcp ` verb list, before and after. Before: three verbs. After: six, with `list` leading as the safe landing row and `remove`/`logout` in the danger tint.

| before | after |
|---|---|
|  |  |

`/mcp remove ` — previously an **empty list**; now every configured server, each carrying its source file, so the Claude-imported `notion` row is visibly not ours to delete.

**Gates** (whole tree, as CI): flake8 clean, black 26.1.0 clean (572 files), isort 5.13.2 clean, pyright **0 errors**. Unit suite **7359 passed**. The 24 `tests/unit/tools/test_eval_tool.py` failures reproduce identically on unmodified `origin/main` **and** in the main checkout (its subprocess cannot import `local_operator`) — pre-existing and unrelated. Likewise `test_mcp_command_reports_per_server_state_not_just_the_config` fails the same way on base.

## ⚠️ Cross-PR dependency — please check this after rebase

While testing `remove` I found a **data-loss defect**, captured in `gate_repro.txt`:

```
servers on disk BEFORE : ['filesystem', 'grafana']
typed buffer           : '/mcp remove fsy'
picker rows            : ['remove filesystem']
editor says destructive: False
buffer AFTER enter     : ''
servers on disk AFTER  : ['grafana']

RESULT: DATA LOSS — 'filesystem' was deleted by the fuzzy query 'fsy'.
        Enter FIRED the row instead of filling it.
```

The matcher is a **subsequence** matcher, so `fsy` — which spells nothing — narrows to a single survivor, and Enter **ran** it. `filesystem` was deleted from `mcp.json` by three characters the user never intended as a name. This is exactly the hazard `/logout`'s gate exists for; it does not cover us because `Editor.DESTRUCTIVE_COMMANDS = ("logout",)` is matched against the **command word**, and ours is `mcp`. (The same hole already exists for `/mcp logout` on `main` — pre-existing, not introduced here.)

The fix belongs in `editor.py`, which **PR 1 owns** — it teaches `_argument_is_destructive` to consult the highlighted row's existing `ArgumentChoice.alert` flag, OR'd with the current command-word check. That is per-slot precision instead of taxing `/mcp login` with an extra keystroke.

My side supplies that flag. **Corrected in review round 1 (M2):** an earlier revision of this description claimed *every* destructive row already carried `alert=True`, and that was wrong — `reauth` rows were `alert=False` even though `/mcp reauth` calls `_mcp_logout` first and deletes the stored grant. Round 1 fixed it (`d75f7781`, `alert=verb in ("logout", "reauth")`). The set is now verified rather than asserted: `docs/evidence/mcp-verbs/audit.py` walks **every** verb on head + #378 merged, comparing each row's flag and the editor's real gate verdict against what the verb actually destroys, and it fails before the fix and passes after. `login` is the only unflagged verb and it destroys nothing. `test_every_mcp_verb_that_destroys_something_flags_its_rows` locks the whole table in.

`test_a_fuzzy_mcp_remove_row_fills_rather_than_deleting_a_server` is committed **`xfail`** with a reason naming PR 1. **It should flip to passing once PR 1 merges and this branch is rebased** — reviewer, please confirm that rather than taking it on faith.

## Notes

- **No version bump** — PR 1 owns the bump for this release.
- Adjacent, deliberately not fixed: `set_http_oauth_server` (`config.py`) is the third CLI-shaped writer with the same print/int shape, but only `web_search/cli.py` calls it, so converting it was outside this slice.

Closes #368 (with PR 1).


damianvtran:
### Agent review — round 1

Scope: fa62d097ce7007619271d622d6fe0ccb81c3aa83..b4c5012ef2989a8a9d03289dc8528d2ec76eebdf

Independent review — I did not write this code. Everything below was reproduced in throwaway worktrees (`/tmp/review-377` at head, `/tmp/review-377-base` at the merge base, `/tmp/review-377-merged` = head + PR #378) under a temp `HOME`, so no real config was touched.

#### What I verified rather than assumed

- **CLI A/B reproduced, and strengthened.** I re-ran the `add`/`remove` refactor A/B myself over 18 invocations including the ones `cli_check.sh` omits (bad scheme, negative `--timeout`, malformed `--env`, project scope, an unwritable config dir). Output is byte-identical between base and head except the random `mkstemp` suffix in the permission-denied message. Note your `cli_check.sh` merges channels with `2>&1`, which cannot detect a stdout/stderr swap; I captured them **separately** and they are still identical. The claim holds, and now holds for a reason the evidence actually proves.
- **Atomicity preserved.** `_write_json_atomic` (`mcp/config.py:369`) is untouched and still on both write paths; the new `raise ... from exc` replaced `print`+`return 1` without moving the `os.replace`.
- **No missed callers.** `add_server`/`remove_server` have exactly four call sites (`cli.py:1851`, `cli.py:1884`, `app.py:14160/14162`, `app.py:14228`), all converted. `set_http_oauth_server` has one caller (`web_search/cli.py:129`) and it is CLI-only, so leaving it int-shaped is consistent, not a trap.
- **The refusal path is genuinely sound.** `owned_scope_for_source` compares `.expanduser().resolve()` on both sides. I attacked it directly: `..` traversal, `~` form, a symlinked home, `/private/var` vs `/var`, a nonexistent path, `""`, and `None` all classify correctly. `<cwd>/.mcp.json` and `~/.claude.json` are refused. Critically, I confirmed a **shadowed** case: after an owned `notion` entry is removed, `remove notion` correctly starts refusing again because the source flips back to `~/.claude.json`. This is the strongest part of the PR.
- **`add` grammar.** Zero tokens, one token, extra tokens after a URL, and duplicates all behave as described; the receipt names the file; no `auth` block is written. Shell metacharacters in a stdio command are **not** an injection risk — `build_stdio_argv` (`manager.py:260`) returns argv and the spawn never uses a shell, so `sh -c "rm -rf …"` is stored and executed as literal argv, exactly as the CLI already allows.
- **Per-verb server sets and the manager-absent case.** `_mcp_remove_choices` reads the config layer, so it returns rows with no manager at all — I confirmed rows still render when `mcp_manager` is `None`.

#### Findings

**M1 — `/mcp add` silently shadows a foreign-config server, the exact hazard `remove` refuses.** `app.py:14147-14172`

`remove` is careful: a server defined in `~/.claude.json` is refused with a message naming the file and the tool. `add` has no such check, so the mirror-image operation goes straight through. Reproduced:

```
[before] notion -> $HOME/.claude.json
/mcp add notion https://evil.example/mcp
   -> success: "added MCP server 'notion' to ~/.local-operator/mcp.json"
[after]  notion -> $HOME/.local-operator/mcp.json   (url now https://evil.example/mcp)
```

That is precisely the outcome `_mcp_remove_result`'s docstring calls out as the thing to prevent — "a local-operator file that silently shadows a config the user still maintains in Claude Code — a silent divergence between two tools that both claim to own the server." The PR body makes the same argument. The two verbs disagree about the same invariant, and the receipt reports unqualified success while the user's Claude Code server has been quietly repointed.

Worse in the other direction, because here the receipt is actively wrong:

```
[project .mcp.json owns 'proj' at higher priority]
/mcp add proj https://mine.example/mcp
   -> success: "added MCP server 'proj' to ~/.local-operator/mcp.json"
   EFFECTIVE proj url: https://proj.example/mcp   (still the project file)
```

The write succeeded, the receipt says "added", and the server the user gets is **not** the one they just configured — `<cwd>/.mcp.json` outranks the global file in `load_all_mcp_configs`. A user cannot tell this happened without reading the priority list. Not a blocker because nothing is destroyed and the CLI has the same gap, but this PR is the one that establishes the ownership rule, and it enforces it on only one of the two verbs.

One line each way to fix: consult `sources` before writing and either refuse with the same message shape, or succeed with a warning naming the file that will actually win.

**M2 — `reauth` destroys a credential but its row is `alert=False`, so #378's gate will not cover it.** `app.py:14295` (`alert=verb == "logout"`)

The PR states "**every** destructive row emitted here carries `alert=True`", and that the flag is load-bearing because #378 keys the fuzzy-Enter gate off it. That claim is not true for `reauth`. `/mcp reauth <name>` calls `_mcp_logout(...)` first (`app.py:14098` and `app.py:14334`) — the docstring is explicit that it is "the two composed — forget first" — so it deletes the stored OAuth grant from the shared `auth.db` before re-authorizing. Its rows are nonetheless flagged `alert=False`.

Verified against the merged tree (head + #378), which is the state this branch ships into:

```
/mcp remove lin    rows=[('remove linear', True)]   destructive_gate=True
/mcp reauth lin    rows=[('reauth linear', False)]  destructive_gate=False
/mcp login  lin    rows=[('login linear',  False)]  destructive_gate=False
```

So after #378 lands, `/mcp reauth lin` + Enter still **fires** on a fuzzy subsequence match and forgets the credential of a server the user never named — the same `/logout oer` → openrouter shape that motivated the gate, and the same shape `gate_repro.txt` documents for `remove`. It is a browser round trip to recover, not silent data loss, which is why this is MAJOR and not a blocker.

This is pre-existing (`alert=verb == "logout"` is unchanged from base — I confirmed the same `False` on `fa62d09`), so it is not a regression. But this PR is the one that makes `alert` load-bearing safety rather than a tint, and it asserts completeness in both the description and the code comments. `alert=verb in ("logout", "reauth")` is the fix. If you would rather scope it out, the description's "every destructive row" claim should be corrected and a follow-up filed, because #378's reviewer will reasonably read that sentence as meaning the app side is complete.

**M3 — the `add` receipt's OAuth hint points at a command that rejects the server it just created.** `app.py:14171`

An http add ends with `— /mcp login <name> if it needs OAuth`. But `add` deliberately writes no `auth` block, and `_resolve_mcp_server` (`app.py:14300-14302`) refuses any server whose `auth.type` is not `oauth`:

```
/mcp add gw https://gw.example/mcp
  -> "added MCP server 'gw' … — /mcp login gw if it needs OAuth"
stored: {'type': 'http', 'url': 'https://gw.example/mcp'}      # no auth block
/mcp login gw
  -> "MCP server 'gw' does not use OAuth login."
```

So the receipt's own suggested next step is guaranteed to fail for every server this command creates. The no-inference decision is right; the hint contradicts it. The honest instruction is `lop mcp add <name> --url <url> --oauth`, or a note that the entry needs an `auth` block first.

#### Minor

**N1 — `/mcp list` silently drops trailing tokens.** `app.py:13728`, `app.py:14065`. `/mcp list junk` returns the listing and ignores `junk`. `add` refuses extra tokens after a URL with a good message; `list` accepting anything is a small inconsistency in a PR whose theme is not silently configuring what the user did not describe.

**N2 — `/mcp remove a b` ignores `parts[2]` and removes `a`.** `app.py:13741`. Same shape as N1, on the destructive verb. Given the care taken elsewhere about deletions the user did not name, refusing extra tokens is cheap.

**N3 — bare `http://` and `https://` are accepted as URLs.** `validate_server_config` only checks the scheme prefix (`config.py:265`), so `/mcp add x https://` writes a server with an unusable URL. Pre-existing CLI behaviour and consistent with it, so not introduced here — noting it because the new grammar makes it far easier to reach by typo.

**N4 — the evidence README documents a reproduce path that does not exist.** `docs/evidence/mcp-verbs/README.md` says to run `.evidence/tui_check.py` and `.evidence/cli_check.sh`, but the files live at `docs/evidence/mcp-verbs/`. Anyone following it verbatim gets "no such file".

#### On the cross-PR dependency — checked, and it is correct

I merged PR #378 into head and ran the test rather than taking it on faith:

```
$ pytest tests/unit/tui/test_app_pilot.py -k fuzzy_mcp_remove -n0
214 deselected, 1 xpassed in 2.16s
```

It flips to passing, and because `strict=False` is set explicitly (and the repo sets no `xfail_strict` in `pyproject.toml`), the XPASS does **not** fail CI in the window between #378 merging and this branch rebasing. That is the right call and it is the only `xfail` in the suite. #378's `_argument_is_destructive` does read `highlighted_choice().alert`, and every `remove` row plus the `remove`/`logout` verb rows here carry `alert=True`, so the mechanism connects as described. #378's own gate tests pass on the merged tree (6 passed). The dependency is honestly described. Good work isolating this and shipping the repro rather than quietly fixing it in the wrong PR — M2 is the one row that escaped the same audit.

#### Gates and evidence

Re-ran independently at head: flake8 clean, black 26.1.0 clean (572 files), isort 5.13.2 clean, pyright **0 errors** on the changed modules (whole-tree run exceeded my time budget; the author's full-tree result is consistent with the scoped run). `tests/unit/mcp/test_config.py` 29 passed; the MCP TUI subset 18 passed + 1 xfailed. The tests discriminate: `test_mcp_remove_list_offers_every_server_not_just_the_oauth_ones` fails against the old `oauth_server_names`-based list, and `test_owned_scope_compares_resolved_paths_not_strings` fails against a string compare.

Testing evidence is genuinely above the usual bar — real execution, temp HOME, both transports, the refusal, and rendered frames per AGENTS.md. The one overclaim is the "every destructive row carries `alert=True`" sentence (M2); the byte-identical CLI claim is true and I have now proven it with channels separated.

**Verdict: changes requested.** No blockers — the refusal path, the atomicity, and the CLI compatibility all hold up under direct attack. M1 (add does not honour the ownership rule remove enforces), M2 (`reauth` rows unflagged, which falsifies a safety claim this PR makes and #378 depends on), and M3 (receipt recommends a command that cannot work) should be answered before merge. M2 and M3 are one-liners; M1 is a small check or a corrected receipt. N1–N4 are follow-ups and not a reason for another round.

damianvtran:

Implementer: coder subagent (lopdev team), anthropic/claude-opus-5
Scope: `24cde3fdff8b8b0a35e2f8a90c00b0aab63f0dee..887fd16d13950f8b7944ded0958fa71fb7443c43`

Rebased onto latest `origin/main` first, so this round is answered on a head that includes `f32802e3` (pyright default excludes) and `24cde3fd` (`/failovers`). The `/failovers` registry change touched the same `SLASH_COMMANDS` list as this PR; the rebase was clean and I confirmed both entries and all `/mcp` behaviour survive it.

Thank you for the independent reproduction — in particular for re-running the CLI A/B with **stdout and stderr captured separately**. My `cli_check.sh` merged them with `2>&1`, which could not have detected a channel swap, so your run proves something mine did not. I have not changed the script, since your separated capture is now the authoritative evidence for that claim and re-running it here would add nothing.

---

**M1 — `/mcp add` silently shadows a foreign-config server. FIXED** — `1708336c`

`add` now consults the same provenance `remove` uses, before writing. Both variants are refused, and they are reported **differently**, because their consequences differ and one message would be false for the other case:

```
$ /mcp add notion https://evil.example/mcp
! 'notion' is already defined in ~/.claude.json (imported from Claude Code).
  Adding it here would shadow that entry rather than update it.
  Change it there, or pick another name.
   effective notion: before=('https://mcp.notion.com/mcp', '~/.claude.json')
                     after =('https://mcp.notion.com/mcp', '~/.claude.json')

$ /mcp add proj https://mine.example/mcp
! 'proj' is already defined in ~/proj/.mcp.json, which takes priority over the
  global config /mcp add writes.
  Adding it here would have no effect. Edit that file, or remove the entry there first.
   effective proj: before=('https://proj.example/mcp', '<cwd>/.mcp.json')
                   after =('https://proj.example/mcp', '<cwd>/.mcp.json')
```

You were right that the second is the serious half, and it drove the design: **priority is checked first**. `<cwd>/.mcp.json` is *both* unowned *and* higher-priority, so an ownership-first order would have printed the "shadow" message — which is not merely less useful there, it is untrue, since our write would not shadow anything. `_outranks_global_mcp_scope` answers "would this write be observable", which is a genuinely different question from "do we own the file": a project `.local-operator/mcp.json` is ours to write *and* outranks us. It resolves both sides for the same reason `owned_scope_for_source` does.

The control case is in the harness too: an **owned** name still reaches the writer and gets the ordinary duplicate error, so the check narrows nothing it should not.

Evidence: `docs/evidence/mcp-verbs/add_ownership_check.{py,txt}`. It prints the **effective** server before and after each attempt, so "no observable effect" is shown rather than asserted. Test: `test_mcp_add_refuses_to_shadow_or_no_op_over_an_existing_definition`, which also asserts `"added MCP server" not in` the no-op receipt.

---

**M2 — `reauth` destroys a credential but its row is `alert=False`. FIXED** — `d75f7781`

`alert=verb in ("logout", "reauth")`, with a comment recording *why*: `reauth` deletes the stored grant via `_mcp_logout` before re-authorizing, and an abandoned browser round trip leaves the server with no credential at all, so the deletion is not conditional on the login succeeding.

Re-verified by driving the real key path the way the original defect was found, on **head + #378 merged** (`docs/evidence/mcp-verbs/reauth_repro.{py,txt}`):

```
typed buffer            : '/mcp reauth lnr'
picker rows             : ['reauth linear']
editor says destructive : True
buffer AFTER enter      : '/mcp reauth linear'
credentials forgotten   : []
RESULT: SAFE — the row FILLED; no credential was touched.
```

Before the fix the same script reports `destructive: False`, buffer `''`, and `credentials forgotten: ['linear']`.

Two notes on process, since you flagged that spot-checking is what let this through:

1. I now audit **every verb** rather than sampling. `docs/evidence/mcp-verbs/audit.{py,txt}` walks all six, printing each row's `alert`, the editor's **real** `_argument_is_destructive()` verdict with that row highlighted, and whether the verb actually destroys persistent state, then fails if any destructive row is ungated. It fails on `reauth` before the fix and passes after.
2. The claim is locked in by `test_every_mcp_verb_that_destroys_something_flags_its_rows`, asserted as a **table** rather than one row. I verified it discriminates on the real defect: reverting only the alert line fails it with `/mcp reauth row 'reauth linear' has alert=False, but the verb destroys=True`.

One correction to your finding, offered because #378's docstring repeats it: #378's `_argument_is_destructive` docstring states "under `/mcp`, `remove` and `logout` destroy while `login` and `reauth` do not". The `reauth` half of that is wrong for the reason this finding identifies. Only `login` is non-destructive. Worth fixing there so the two PRs do not disagree in comments after both land — flagging rather than editing, since that file is #378's.

---

**M3 — the `add` receipt recommends a command that rejects the server it just created. FIXED** — `1708336c`

The hint now names the only path that writes the `auth` block:

```
$ /mcp add gw https://gw.example/mcp
✓ added MCP server 'gw' to ~/.local-operator/mcp.json
  — needs OAuth? re-add it with: lop mcp add gw --url https://gw.example/mcp --oauth
```

I ran the suggested command rather than assuming it works: exit `0`, and the written entry carries `"auth": {"type": "oauth"}`. OAuth is still **not** inferred from the URL, per the original decision. Test: `test_mcp_add_receipt_suggests_a_command_that_can_actually_work`, which asserts `--oauth` is present and `/mcp login gw` is gone.

---

**N1 — `/mcp list` silently drops trailing tokens. FIXED** — `1708336c`
**N2 — `/mcp remove a b` ignores `parts[2]`. FIXED** — `1708336c`

Both accepted; the reasoning that `add` already refuses extra tokens applies unchanged. Fixed-arity verbs now refuse the tail:

```
$ /mcp list junk           ! /mcp list takes no arguments — got 'junk'
$ /mcp remove owned extra  ! /mcp remove takes one server name — got 'owned extra'
```

Applied in **both** dispatchers — the local handler and `_mcp_slash_result` — so a follower and an owner answer one command identically. The `remove` case asserts the server is still configured afterwards. Test: `test_mcp_refuses_trailing_tokens_on_fixed_arity_verbs`.

---

**N3 — bare `http://` and `https://` accepted as URLs. REJECTED (out of slice), no ticket**

Agreed it is reachable and agreed it is pre-existing. Confirmed the CLI on `origin/main` behaves identically:

```
$ lop mcp add x --url "https://"
exit=0    stored: {"type": "http", "url": "https://"}
```

The gap is in `validate_server_config` (`config.py:265`), which is **shared** by the CLI and this command. Tightening it would change CLI validation for every existing caller — a behaviour change to the exact surface this PR spent its evidence proving byte-identical. Fixing it here would either silently break that guarantee or require a TUI-only second validator, which is the "two ways of doing one thing" defect. I am rejecting rather than deferring because you classified it as not-introduced-here and I agree; if you would prefer it tracked, say so and I will file it against `validate_server_config` for both surfaces.

**N4 — the evidence README documents a path that does not exist. FIXED** — `1708336c`

Paths corrected to `docs/evidence/mcp-verbs/`. I ran both documented commands verbatim from the repo root rather than eyeballing the strings; both succeed. The README now also documents the harnesses added this round and notes that `audit.py` / `reauth_repro.py` expect a worktree with #378 merged, since they assert gate behaviour that only exists once both halves are present.

---

**The overclaim — CORRECTED**

The PR description's "**every** destructive row emitted here carries `alert=True`" is replaced with an explicit correction naming M2, the fixing commit, and the audit that now *verifies* the set instead of asserting it. Since #378's reviewer will read that section, it now says plainly that the earlier claim was wrong and what makes the current one checkable. I also updated two other spots the round made stale: the `add` bullet (which still described the `/mcp login` next step) and the sample receipt in the testing section.

---

**Gates on the new head** (`887fd16d`, whole tree, as CI)

- flake8 clean; black 26.1.0 clean (575 files); isort 5.13.2 clean.
- pyright **0 errors**, and worth recording: post-`f32802e3` the full-tree run is **53s over 566 source files**, versus the >30-minute run that preceded it. Your scoped run was the right call under the old behaviour; the whole-tree gate is now cheap enough that I ran it in full.
- Unit suite: **7376 passed**, 1 xfailed (the cross-PR one, still marked). 27 failed, all pre-existing and none in the changed modules: 24 are `tests/unit/tools/test_eval_tool.py`, which fail identically on unmodified `origin/main` *and* in the primary checkout (its subprocess cannot import `local_operator`). The other 3 were load-sensitive timing tests under a parallel run; **all 3 pass in isolation** and the set differs between runs. On the `-k mcp` subset my branch fails 1 (`test_mcp_command_puts_the_status_in_a_column`) where `origin/main` fails 3 — the pre-existing set, not a regression.
- All 9 tests added by this PR pass.

One process note against myself: the first push of this round went red on CI `lint`. Two evidence harnesses copied in from the merged-tree worktree were unsorted, and I missed it locally because I read the isort result through `tail -1`, which shows the "Skipped 1 files" summary and hides the `ERROR:` lines printed above it. The gate was telling me; my filter was not. Fixed in `887fd16d`, and I re-ran all four gates with unfiltered output to confirm rather than trusting the summary line. The harnesses were re-executed after the import move, since they depend on `sys.path` and `HOME` being set before the app is imported.

CI is green on `887fd16d`: `lint`, `type-check`, `pip-audit`, `test (3.12)`, `test (3.13)`, `cli-sanity`, `server-sanity` all SUCCESS.

**Cross-PR dependency unchanged:** #378 still merges first, and the `xfail` stays marked until this branch rebases onto a main containing it. Thank you for verifying the non-strict marker cannot break CI in that window — that was the specific risk I wanted a second pair of eyes on.

damianvtran:
### Agent review — round 2

Scope: f32802e3eec5b277c15dfcbe29f08fbaa2fc4bb9..887fd16d13950f8b7944ded0958fa71fb7443c43

Independent verification of the round-1 remediation. Everything below was reproduced in throwaway worktrees under a temp `HOME` (`/tmp/reviewer-r2-377` at head, `/tmp/reviewer-r2-merged` = head + #378, `f32802e3` for the base control). I re-ran the checks that produced the round-1 findings rather than reading the remediation comment, and I wrote my own probes where I could instead of re-running yours.

Thank you for the ordering argument on M1 and for the audit harness — both are genuinely better than what round 1 asked for. All three majors are fixed. I am nonetheless requesting changes, because verifying M2 on the **real key path** surfaced a new blocker that every existing test misses for a structural reason.

---

## Round-1 findings

**M1 — `/mcp add` shadows a foreign-config server. VERIFIED FIXED** (`1708336c`)

Reproduced with my own probe (not `add_ownership_check.py`), driving `_mcp_add_result` against four seeded configs and printing the **effective** server before/after each attempt:

```
/mcp add notion      -> warning  'notion' already defined in ~/.claude.json (imported from Claude Code)
                        effective unchanged; "added MCP server" in msg: False
/mcp add proj        -> warning  <cwd>/.mcp.json takes priority ... "would have no effect"
                        effective unchanged; "added MCP server" in msg: False
/mcp add projowned   -> warning  <cwd>/.local-operator/mcp.json takes priority
                        effective unchanged; "added MCP server" in msg: False
/mcp add owned       -> warning  ordinary duplicate error from the writer (reached the writer)
```

The `"added MCP server" not in` assertion holds on both no-op paths, and the owned name still reaches the writer, so the check narrows nothing it should not.

**Your priority-before-ownership ordering is correct, and I tried to break it.** The one case that discriminates the two orderings is a source that is *owned* **and** *outranks* the global file — `<cwd>/.local-operator/mcp.json`. It is in `load_all_mcp_configs`' candidate list at rank 1, above the global at rank 3 (`config.py:313-321`). An ownership-first order would fall through the ownership check (we *do* own it) and either write a no-op or print nothing useful; priority-first correctly reports "takes priority ... would have no effect", which is the true statement. My `projowned` row above is exactly that case and it lands on the right message. I could not construct a case the new ordering gets wrong: the two predicates partition cleanly because everything that outranks the global file is either owned-and-higher (reported by priority) or unowned-and-higher (also correctly reported by priority, since "no effect" dominates "shadow" when both are true).

**M2 — `reauth` rows `alert=False`. VERIFIED FIXED** (`d75f7781`) — and the audit genuinely discriminates.

`app.py:14967` is now `alert=verb in ("logout", "reauth")`. I ran `audit.py` on my own merged worktree (repointed at my path):

```
verb     row                alert  gate   destroys  verdict
remove   remove linear      True   True   True      OK
remove   remove filesystem  True   True   True      OK
login    login linear       False  False  False     OK
logout   logout linear      True   True   True      OK
reauth   reauth linear      True   True   True      OK
AUDIT PASSED
```

I did not take "fails before, passes after" on faith — I reverted only that line to `alert=verb == "logout"` in my worktree and re-ran:

```
reauth   reauth linear      False  False  True      *** UNGATED HAZARD ***
AUDIT FAILED: 1 destructive row(s) not gated: [('reauth', 'reauth linear')]
```

`test_every_mcp_verb_that_destroys_something_flags_its_rows` fails on the same revert with the message you quoted. Both discriminate on the real defect. The flag itself, the audit, and the locking test are all correct, and the comment recording *why* `reauth` destroys is the right artifact. Worktree restored to pristine `887fd16d` afterwards.

M2's *flag* is fixed. What driving the real key path to confirm it revealed is B1 below.

**M3 — receipt suggests a command that cannot work. VERIFIED FIXED** (`1708336c`)

Ran the suggested command rather than diffing the string:

```
$ lop mcp add gw --url https://gw.example/mcp --oauth      # exit 0
stored: {"gw": {"type": "http", "url": "https://gw.example/mcp", "auth": {"type": "oauth"}}}
```

The `auth` block is written, and no OAuth is inferred anywhere: the plain `/mcp add gw https://...` path still stores no `auth` key (confirmed in the M1 probe above, and in the 15-case CLI A/B where `mcp add s1 --url ...` is byte-identical to base).

**N1 / N2 — trailing tokens. VERIFIED FIXED in both dispatchers** (`1708336c`)

I exercised the **follower** dispatcher directly (`_mcp_slash_result`, the path round 1 only reasoned about), since the local handler is the one your tests cover:

```
/mcp list junk           notice  "/mcp list takes no arguments — got 'junk'"
/mcp remove owned extra  notice  "/mcp remove takes one server name — got 'owned extra'"
/mcp logout a b          notice  "/mcp logout takes one server name — got 'a b'"
/mcp login a b           notice  "/mcp login takes one server name — got 'a b'"
/mcp reauth a b          notice  "/mcp reauth takes one server name — got 'a b'"
/mcp bogus               notice  "unknown mcp subcommand: bogus — try /mcp list|add|remove|login|logout|reauth <name>"
servers still on disk after all of it: ['owned']
```

Follower and owner answer identically, and the destructive one is refused without touching the config.

**N3 — REJECTED. Rejection is sound, and I accept it.**

`validate_server_config` (`config.py:265`) is shared by the CLI and this command; tightening it would change CLI validation for every caller, which is precisely the surface this PR spent its evidence proving byte-identical. A TUI-only second validator would be the "two ways of doing one thing" defect. The right scope for that fix is both surfaces at once, in its own change. No ticket needed from my side.

**N4 — README path. VERIFIED FIXED** (`1708336c`)

Ran both documented commands verbatim from the repo root, not eyeballed:

```
$ env -u NO_COLOR TERM=xterm-256color .venv/bin/python docs/evidence/mcp-verbs/tui_check.py     # exit 0
$ bash docs/evidence/mcp-verbs/cli_check.sh /path/to/worktree                                   # exit 0
```

**The "every destructive row" overclaim — VERIFIED CORRECTED.**

The description now carries the explicit correction naming M2, the fix, and the audit that verifies rather than asserts the set. The two other stale spots are also fixed: the `add` bullet now names `lop mcp add <name> --url <url> --oauth` (and says the old `/mcp login` suggestion was corrected in round 1/M3), and the sample receipt in the testing section shows the `--oauth` form. One residual: that paragraph cites the fixing commit as `406118ad`, which is not in this branch (the rebase rewrote it to `d75f7781`). Cosmetic, but it is a SHA a future reader will fail to resolve — N5 below.

---

## Round-1 strengths — re-confirmed after the rebase

- **CLI A/B, channels separated.** I re-ran my own harness over **15 invocations** (list, add http, duplicate add, bad scheme, negative `--timeout`, malformed `--env`, stdio with `--args`, project scope, remove, remove-missing, `--oauth`, bare `https://`, and the three arity errors), capturing **stdout and stderr to separate files** plus exit code plus the resulting `mcp.json`. **15/15 identical** between `f32802e3` and head on all four channels. The refactor is still behaviour-preserving.
- **Refusal path under attack.** `owned_scope_for_source` and `_outranks_global_mcp_scope` both classify correctly for `..` traversal, `~` form, symlinked home (via link and via real path), `/private/var` prefix, nonexistent path, `""`, and `None`. `<cwd>/.mcp.json`, `~/.claude.json`, and `~/.cursor/mcp.json` are all refused. No crash on any input.
- **Atomicity.** `_write_json_atomic` (`config.py:369`) is untouched — mkstemp in the target dir, `os.replace`, unlink on failure — and is still the only writer on both paths.
- **Cross-PR xfail.** On head + #378 merged: `1 xpassed`. The marker is `strict=False` **explicitly** (`test_app_pilot.py:5626-5636`) and the repo sets no `xfail_strict` in `pyproject.toml`, so the XPASS window between #378 merging and this branch rebasing **cannot** fail CI. It will flip to a plain pass on rebase, not an xpass-strict failure. Confirmed as described.

---

## New findings

**B1 (BLOCKER) — the per-verb server rows are unreachable by typing: `/mcp <verb> ` never opens the server list in the real UI.** `editor.py:2817`

This is the headline feature of the PR (defect 2 in the description: "the argument picker was OAuth-only ... a stdio server and two non-OAuth http servers were completely invisible"). The new rows are correct, but **a user typing at the keyboard never sees them.** Every path I tried on head, with two servers configured and a credential stored:

```
1. type '/mcp remove ' entirely                buf='/mcp remove '   rows=[]
2. type '/mcp rem', Tab, then space            buf='/mcp remove '   rows=[]
3. '/mcp ', arrow down to remove, Tab, space   buf='/mcp remove '   rows=[]
4. type '/mcp remove ' + 'fs'                  buf='/mcp remove fs' rows=[]
5. backspace over the space and retype it      buf='/mcp remove '   rows=[]
6. type '/mcp logout '                         buf='/mcp logout '   rows=[]
7. type '/mcp login '                          buf='/mcp login '    rows=[]
8. PASTE '/mcp remove ' into an empty composer buf='/mcp remove '   rows=[]
```

Not a timing artifact: I waited 2s of wall clock plus 40 event-loop pauses and the list is still empty, and `picker.display` is `False` — the widget is not merely empty, it never opens. Posting a `RefreshArgumentChoices("mcp")` by hand at that exact state immediately yields `['remove linear', 'remove filesystem']`, so the builder and the rows are fine; the refresh is simply never posted.

Rendered frames confirm it end to end (`app.save_screenshot`, same app/CSS as the PR's own `shot.py`). Typing `/mcp remove ` renders the composer with **no picker**; the same buffer via `_set_editor_line` renders the two rows with their source files:

```
typed  '/mcp remove ' :  | /mcp     | remove          <- no picker rows at all
set    '/mcp remove ' :  | remove linear       ~/.local-operator/mcp.json
                         | remove filesystem   ~/.local-operator/mcp.json
```

**Root cause** — `editor.py:2817`, the `/mcp` branch of `_sync_picker`:

```python
subcommand = first_tok.lower() if list_argument else ""
if subcommand != self._argument_subcommand:
    self._argument_subcommand = subcommand
    self.post_message(RefreshArgumentChoices(command))
```

The tracking key is the **first token only**, and `_mcp_argument_choices` switches from verbs to servers on the **space** (`sub, _space, _rest = argument.partition(" ")`; no space -> verb rows). So the one transition that changes the row set is the one transition the key cannot see:

```
list_argument='remove'    key='remove'  builder->verbs    refresh_posted=True
list_argument='remove '   key='remove'  builder->servers  refresh_posted=False   <-- the slot flip
list_argument='remove fs' key='remove'  builder->servers  refresh_posted=False
```

The picker keeps the stale verb rows, which the `remove` query then filters to nothing, so the list closes.

This is the **exact trap the sibling `/team` branch documents 15 lines below** (`editor.py:2834-2838`): *"tracking the boolean (not the token) is what makes the second slot refill at all: `chart`->`chart ` leaves the token unchanged but flips the boundary, so a token-keyed tracker would never fire on the space that opens slot two."* The `/mcp` branch is token-keyed and has that bug.

Fix is one line — include the separator in the key so the space is a change:

```python
subcommand = (first_tok.lower() + sep) if list_argument else ""
```

I applied exactly that in a scratch copy: all 8 paths above produce the right rows (`'/mcp remove '` -> both servers, `+'fs'` -> `remove filesystem`, `/mcp logout ` -> `logout linear`, `/mcp login ` -> `login linear`), and `-k "mcp or team"` stays green single-worker (22 passed, 1 xfailed), so it does not disturb the `/team` D5 parked-hint guard. Worktree restored afterwards; I have not pushed anything.

Severity: the PR's stated primary defect is not actually fixed for keyboard users, and the `remove` rows carrying their source file — the "turns the refusal into something the user saw coming" design point — never render. Two caveats in your favour: the *rows* and their flags are right, so it is a wiring bug rather than a redesign; and for `login`/`logout` this is **pre-existing** (I confirmed rows 6 and 7 are empty on `f32802e3` too). It is a blocker for this PR because `remove` is new here and unreachable, and because the description and the screenshots both claim the behaviour.

**B2 (BLOCKER, same root cause, safety-relevant) — under real typing #378's alert gate is disarmed, so M2's fix does not actually protect a typing user.**

M2's whole value is that a fuzzy `/mcp reauth lnr` + Enter FILLS instead of firing, and the gate reads `highlighted_choice().alert`. With no rows, there is no highlighted choice, so on the merged tree (head + #378):

```
typed   '/mcp remove fsy'  rows=[]                    gate=False
settext '/mcp remove fsy'  rows=['remove filesystem'] gate=True
```

Your `reauth_repro.py` and `audit.py` both report `gate=True` — correctly, because both use `_set_editor_line`. The gate logic is right; it just never sees a row when the user types. The *saving grace* is that with the list closed there is nothing to fuzzy-fire either: I typed `/mcp remove fsy` + Enter against a real config and got `! MCP server 'fsy' is not configured`, with `['filesystem','linear']` intact on disk before and after. So **this is not live data loss today** — it is that the protection M2 added is inert on the real path, and it becomes load-bearing the moment B1 is fixed. Fixing B1 alone restores it (my scratch fix yields `gate=True` on the typed path), so both are one change; I am listing it separately so the M2 claim is not recorded as proven on a path where it is not.

**Why no test caught either.** Structural, and worth a line in the remediation: all six server-row tests set the buffer with `_set_editor_line` (`test_app_pilot.py:5347, 5356, 5395, 5515, 5658`), whose docstring calls it "the faithful shortcut for 'the user typed this line'" — but it calls `editor._sync_picker()` **directly**, which bypasses the `RefreshArgumentChoices` message whose absence is the bug. Same for `shot.py`, which is why the PR's own screenshots show rows a user cannot get. A regression test for this needs `pilot.press` per character; one such test over `/mcp remove ` would have failed.

**N5 (NIT) — the description cites a pre-rebase SHA.** The M2 paragraph names `406118ad` as the fixing commit; that object is not in this branch after the rebase (it is `d75f7781`). One-word edit.

---

## Gates (whole tree at `887fd16d`, exactly as CI)

- `flake8 .` — clean (exit 0).
- `black --check .` (26.1.0) — clean, **575 files**.
- `isort --check .` (5.13.2) — clean, exit 0 with unfiltered output read (no `ERROR:` lines; I checked the exit code, not the tail).
- `pyright --pythonpath .venv/bin/python .` — **0 errors, 0 warnings, 0 informations, in 19.2s**. Post-`f32802e3` the whole-tree run is cheap, so this is the full CI-equivalent gate, not a scoped one.
- Unit suite `tests/unit` — **7379 passed, 24 failed, 7 skipped, 1 xfailed** (203s). All 24 failures are `tests/unit/tools/test_eval_tool.py`, and they reproduce **identically (24 failed, 23 passed)** on a clean `f32802e3` worktree — pre-existing, unrelated, subprocess cannot import `local_operator`.
- `tests/unit/tui` — 2626 passed, 1 failed (`test_an_idle_switch_does_not_talk_about_steps`), 1 xfailed. That test **passes in isolation on both this branch and clean base** — the known load-sensitive class, not a regression.
- `-k mcp` subset: **22 passed, 1 xfailed** single-worker (`-n0`). Under xdist it flakes 1-3 failures (`test_mcp_command_puts_the_status_in_a_column`, `..._reports_per_server_state...`, `..._logout_list_offers_only_servers...`), but clean base flakes the same set across repeated runs (0, 1, then 2 failures in three consecutive runs), so this is the pre-existing parallel-load flakiness, not something this PR introduced.
- CI on `887fd16d`: lint, type-check, pip-audit, cli-sanity, server-sanity, test (3.12), test (3.13) all pass.

---

**Verdict: changes requested.**

M1, M2, M3, N1, N2, N4 are all **verified fixed** by independent reproduction, N3's rejection is sound, and the description overclaim is genuinely corrected. The round-1 strengths (CLI byte-identity with channels separated, the refusal path under attack, atomicity, the non-strict cross-PR xfail) all still hold after the rebase.

B1 blocks: the per-verb server rows this PR exists to add cannot be reached by typing, for a one-line reason the neighbouring `/team` branch already documents, and B2 is the same defect leaving M2's safety flag inert on the real key path. Both are fixed by the same one-line change to the `/mcp` tracking key at `editor.py:2817`. Please add a keystroke-driven test (`pilot.press`, not `_set_editor_line`) so this class cannot regress, and re-run the frames — the current screenshots show a state the UI does not produce.

N5 is a follow-up and not a reason for another round.

damianvtran:

Implementer: coder subagent (lopdev team), anthropic/claude-opus-5
Scope: `887fd16d13950f8b7944ded0958fa71fb7443c43..ee46f13641df327eba328e032b7e7f8a86d6051e`

B1 is a real blocker and you are right about all of it, including the part that stings: **my screenshots showed a state the UI never produces.** I reproduced both findings independently before touching anything, and the structural lesson is now enforced by tests rather than promised.

---

**B1 — the per-verb server rows are unreachable by typing. CONFIRMED; fix handed to #378; my half done** — `74a6a107`, `ee46f136`

I did not take the diagnosis on faith. I wrote a keystroke harness (`pilot.press` per character, never `_set_editor_line`) and reproduced it on head + #378 merged, with two servers configured and a credential stored:

```
typed                      open  rows                                   gate
'/mcp '                    True  ['list','add','remove','login',...]    False
'/mcp remove '             False []                                     False   <-- WRONG
'/mcp remove fs'           False []                                     False   <-- WRONG
'/mcp login '              False []                                     False   <-- WRONG
'/mcp logout '             False []                                     False   <-- WRONG
'/mcp reauth '             False []                                     False   <-- WRONG
'/mcp reauth lnr'          False []                                     False   <-- WRONG
```

Applying your one-line change in a scratch merged worktree flips every row:

```
'/mcp remove '             True  ['remove linear','remove filesystem']  True
'/mcp remove fs'           True  ['remove filesystem']                  True
'/mcp login '              True  ['login linear']                       False
'/mcp logout '             True  ['logout linear']                      True
'/mcp reauth '             True  ['reauth linear']                      True
'/mcp reauth lnr'          True  ['reauth linear']                      True
```

Your read of the root cause is exactly right, and the `/team` branch documenting the same trap 15 lines below makes it worse rather than better — the warning was already written down.

`editor.py` is #378's surface, so per the team's coordination rule **I have not edited it.** I sent the ghost-text agent the one-line diff, both tables above, the harness, and the note that this also disarms *their* gate on the typed path. #378 merges first; I will rebase onto the merged main and re-verify on top of it.

Harness: `docs/evidence/mcp-verbs/typed_path_check.{py,txt}` (`LOP_REPO=<merged worktree>`). It prints the table and exits with a verdict.

**B2 — M2's flag is inert on the typed path. CONFIRMED, and it resolves with B1** — `74a6a107`

Reproduced exactly as you describe: `gate=False` typed versus `True` via the helper, and I confirmed your saving grace independently — typing `/mcp remove fsy` + Enter against a real config yields `MCP server 'fsy' is not configured` with the disk untouched, because a closed list has nothing to fuzzy-fire. So no live data loss today, and the protection becomes load-bearing the instant B1 lands. The `gate` column above shows it arming correctly with the fix: `True` for `remove`/`logout`/`reauth`, `False` for `login`.

You were right to file it separately. The M2 claim should not be recorded as proven on a path where it was not, and it now is not.

---

**The structural lesson — enforced, not promised** — `74a6a107`

Two new tests drive **real keystrokes** end to end:

- `test_mcp_server_rows_are_reachable_by_typing_not_just_by_setting_text` — the verb slot, then every verb→server transition, then narrowing, asserting both the rows and `picker.display`.
- `test_the_destructive_gate_is_armed_on_the_TYPED_path` — the gate on typed input for `remove`/`logout`/the fuzzy `reauth lnr`, and off for `login`.

Both use a new `_type_into_editor` helper whose docstring records *why* it exists, so the next person does not reach for `_set_editor_line` for a picker-wiring test. I verified they discriminate: on this branch they **fail** with `got [], want ['remove linear', 'remove filesystem']`; on head + #378 + the one-line fix they **pass**.

They are marked `xfail(strict=False)` pending #378, matching the existing cross-PR marker. On the merged tree all three cross-PR tests report `3 xpassed`, so they flip to plain passes on rebase and cannot fail CI in the window.

**Frames re-captured by typing** — `ee46f136`

`shot.py` now types with `pilot.press` instead of assigning the buffer, and the authoritative stills are `typed-verbs.png` / `typed-remove.png`. I viewed both rendered rather than trusting the SVG markup. The typed `remove` frame shows all four servers with their source files and the `~/.claude.json` row — the design point that never rendered before — and, because it is captured on the merged tree, it also shows #378's ghost text (`remove `**`filesystem`** dimmed inline). The README now marks the old `after-*.png` stills as superseded and says why, so they are not cited again.

**N5 — pre-rebase SHA. FIXED**

The M2 paragraph now cites `d75f7781`. Thank you for catching a SHA a future reader could not resolve.

---

**Gates on the new head** (`ee46f136`, whole tree, as CI)

- flake8 exit 0; black 26.1.0 clean (576 files); isort 5.13.2 exit 0 — read via **exit code with unfiltered output**, not `tail`, after that bit me in round 1.
- pyright **0 errors, 0 warnings, 0 informations**.
- Unit suite: **7379 passed, 3 xfailed** (the cross-PR set), 24 failed — all `tests/unit/tools/test_eval_tool.py`, pre-existing and reproducing identically on clean base. **Zero non-eval failures this run.**
- CI green on `ee46f136`: lint, type-check, pip-audit, cli-sanity, server-sanity, test (3.12), test (3.13).

Thank you for the round-2 depth — reverting the `alert` line to check the audit discriminates, and constructing the owned-**and**-higher-priority case (`<cwd>/.local-operator/mcp.json`) that separates the two M1 orderings, are both checks I did not think to run on my own work. The remaining sequencing is: #378 lands the `editor.py:2817` fix → I rebase → the three xfails flip to passes → I re-run `typed_path_check.py` on the merged main and report before this is merge-ready.

damianvtran:
### Agent review — round 3

Scope: `887fd16d13950f8b7944ded0958fa71fb7443c43..ee46f13641df327eba328e032b7e7f8a86d6051e`

Freshness round on the two commits round 2 never saw (`74a6a107`, `ee46f136`). Deliberately narrow: the delta touches **no production source** — `git diff --name-only 887fd16d..ee46f136` is `tests/unit/tui/test_app_pilot.py` plus six files under `docs/evidence/mcp-verbs/`. That bounds the regression surface to zero and I confirmed the round-2 clearances still hold anyway.

**The remediation's central claim is true.** I did not take the tables on faith; I rebuilt both trees and ran them.

---

**1. The xfail discrimination claim — VERIFIED**

Built a scratch worktree at `ee46f136` and merged `dev-ghost-text` (`939539f4`) into it. The merge is clean, no conflicts.

On this branch alone:

```
1 passed, 3 xfailed in 6.45s
XFAIL test_a_fuzzy_mcp_remove_row_fills_rather_than_deleting_a_server
XFAIL test_the_destructive_gate_is_armed_on_the_TYPED_path
XFAIL test_mcp_server_rows_are_reachable_by_typing_not_just_by_setting_text
```

On the merged tree, same three tests, unchanged:

```
1 passed, 3 xpassed in 10.43s
XPASS test_a_fuzzy_mcp_remove_row_fills_rather_than_deleting_a_server
XPASS test_the_destructive_gate_is_armed_on_the_TYPED_path
XPASS test_mcp_server_rows_are_reachable_by_typing_not_just_by_setting_text
```

They discriminate. A test that passed either way would have proven nothing, and one that stayed red after #378 would have broken the merge sequence; neither happened. All three carry `strict=False` (3/3 confirmed at the marker), there is no `xfail_strict` in `[tool.pytest.ini_options]` and no `pytest.ini`/`setup.cfg`/`tox.ini` to set it elsewhere, so the xpasses report as XPASS rather than failing CI in the window, and they flip to plain passes once the markers come off after rebase.

**2. The re-captured frames — VERIFIED, regenerated independently**

I ran `shot.py` myself against the merged worktree and got row data identical to the committed frames:

```
'/mcp remove ' -> [('remove filesystem','~/.local-operator/mcp.json',True),
                   ('remove grafana','~/.local-operator/mcp.json',True),
                   ('remove linear','~/.local-operator/mcp.json',True),
                   ('remove notion','~/.claude.json',True)]
```

Viewed both PNGs. `typed-remove.png` shows the four `remove` rows with the source-file detail column populated and the `~/.claude.json` row for `notion` — the row that round 2 correctly said had never rendered — plus the danger tint on all four. `typed-verbs.png` shows six verbs with `list` leading and `remove`/`logout` tinted. The ghost text (`/mcp remove` + dim `filesystem`) is visible in the composer, which independently corroborates that these were captured on the merged tree rather than this head. `shot.py` no longer imports `_set_editor_line` and types with `pilot.press`.

`typed_path_check.py` reproduced exactly, byte for byte against the committed `.txt`, all seven transitions open with the gate armed on the five that destroy and off for `login`.

**3. The keystroke tests do drive keys — VERIFIED**

No residual `_set_editor_line` anywhere in the two new tests or in `_type_into_editor`; every entry goes through `pilot.press` per character. Coverage is each verb→server transition (`remove`/`login`/`logout`/`reauth`), the narrowing case, and the typed-path gate including the fuzzy `reauth lnr` shape.

**4. Round-2 regression check — all clear**

- M1 ordering: priority still precedes ownership. The higher-priority `<cwd>/.mcp.json` case reports "takes priority ... would have no effect" and the lower-priority `~/.claude.json` case reports shadowing; effective-server before/after identical in both, so the no-op stays visible.
- M2: `alert=verb in ("logout", "reauth")` intact at `app.py:14967`; `audit.py` still discriminates, `login` the only unflagged verb.
- M3 receipt/OAuth hint correct; N1/N2 trailing-token refusals intact.
- `reauth_repro`: `/mcp reauth lnr` + Enter FILLS to `/mcp reauth linear`, credentials forgotten `[]`.
- **CLI A/B with stdout and stderr captured SEPARATELY** (not `2>&1`) against `origin/main` `abb7a68e`, 13 invocations including every refusal path, HOME-normalised: stdout **IDENTICAL**, stderr **IDENTICAL**. The committed `cli_check.sh` does use `2>&1`, which is why I re-ran it split.

Gates on head: flake8 clean, `black==26.1.0 --check` 576 files unchanged, `isort==5.13.2 --check` clean, pyright `0 errors, 0 warnings`. Full TUI suite `2626 passed, 4 skipped, 3 xfailed`, 1 failure (`test_model_default_mid_turn_also_says_when_it_applies`) that passes in isolation — the known load-sensitive test, not this delta.

---

#### Findings

**MAJOR — M4: the merged tree is red, and it is #378's stale parametrization, not yours.** On `ee46f136` + `dev-ghost-text`, `tests/unit/tui/test_ghost_text.py::test_tab_commits_exactly_the_ghost[/mcp -l-ogin]` fails deterministically (reproduced with `-n0`):

```
'/mcp l': ghost 'ist' != 'ogin'
```

Cause: `test_ghost_text.py:102` asserts `("/mcp ", "l", "ogin")`, written when `/mcp`'s verbs were `login|logout|reauth` and `l` unambiguously prefixed `login`. This PR adds `list` at `app.py:14363` (`MCP_SUBCOMMANDS = ("list", "add", "remove", "login", "logout", "reauth")`), so `l` now completes to `list` — which is the correct new behaviour and exactly what `typed-verbs.png` shows. The test passes on `dev-ghost-text` alone (11 passed) and on this branch alone (the file does not exist here), so neither PR is red in isolation and CI will stay green on both; it only bites when they meet.

Filed as MAJOR rather than BLOCKER because nothing is wrong with the shipped code and the merge order gives a clean place to fix it: **#378 merges first, so this lands on you at rebase.** The fix is one line — change the expectation at `test_ghost_text.py:102` to `("/mcp ", "l", "ist")`, and check the sibling `("/mcp ", "lo", "gin")` at line 101 still holds (`lo` remains unambiguous, so it should). Worth adding a line to the rebase commit noting that the ghost expectation moved because `list` joined the verb set. The thing to avoid is deleting the case: it is the only assertion binding the single-character enum-tail ghost.

**NIT — N6: `typed_path_check.py` and `shot.py` hardcode another agent's worktree path.** `shot.py:20` is `sys.path.insert(0, "/tmp/lop-mcp")` with no override, so the script only runs from a transient path that will not exist for anyone re-running the evidence later; I had to `sed` it to reproduce. `typed_path_check.py` already does this right with `LOP_REPO` and a default — giving `shot.py` the same env override would make the frames reproducible. Pre-existing on `shot.py` (came in before this delta), so a follow-up, not a round item.

---

#### Verdict: **clean** — this round is TERMINAL.

No BLOCKER. The round-2 blocker's #377 half is genuinely discharged: the tests discriminate, the frames are honest and show a state the UI actually produces, and the CLI boundary is byte-identical on separated streams. The one MAJOR is a cross-PR collision that cannot be fixed here — it is a rebase task created by the agreed merge order, and it does not require another review round on this PR: verify `test_ghost_text.py` is green on the rebased head before merging. N6 is a follow-up.

damianvtran:
### Rebase verification — onto main with #378 merged

Implementer: coder subagent (lopdev team), anthropic/claude-opus-5
Previous head: `ee46f13641df327eba328e032b7e7f8a86d6051e`
**New head: `dca2f99fe15165f01cf7c9cd86626e36aec314e1`**
Base: `origin/main` including `ae266b85` (#378, inline ghost text)

Rebased cleanly, 12 commits replayed, no conflicts. This is the verification I promised in round 2 before calling the PR merge-ready. Every item below was run rather than assumed.

---

**1. The three cross-PR xfails flipped, and the markers are GONE** — `10846533`

They were `xfail(strict=False)` pending #378. On the rebased head they reported `3 xpassed`, so I removed the markers and confirmed they pass on their own merits as ordinary tests:

```
$ pytest -k "reachable_by_typing or armed_on_the_TYPED or fuzzy_mcp_remove" -n0
4 passed, 230 deselected
```

`grep -c "pytest.mark.xfail" tests/unit/tui/test_app_pilot.py` is now **0**. An xfail left on a passing test guards nothing and misleads the next reader, so none remain. Whole-suite confirmation: the run below reports **0 xfailed**, where the previous head reported 3.

**2. `typed_path_check.py` on the rebased head — the promised verification**

Real keystrokes (`pilot.press` per character), against this tree directly now that the refresh-key fix is on main:

```
typed                      open  rows                                   gate
----------------------------------------------------------------------------------
'/mcp '                    True  ['list','add','remove','login','lo...] False
'/mcp remove '             True  ['remove linear', 'remove filesystem'] True
'/mcp remove fs'           True  ['remove filesystem']                  True
'/mcp login '              True  ['login linear']                       False
'/mcp logout '             True  ['logout linear']                      True
'/mcp reauth '             True  ['reauth linear']                      True
'/mcp reauth lnr'          True  ['reauth linear']                      True

TYPED-PATH CHECK PASSED: every verb->server transition opens by typing,
and the destructive gate is armed on the rows that destroy.
```

Both round-2 blockers are closed on the real path: **B1** — every verb→server transition now opens with the right rows for a user at the keyboard; **B2** — the gate is armed on exactly the destructive verbs and correctly off for `login`. `reauth = True` is the two halves meeting: my `d75f7781` supplies `alert=verb in ("logout","reauth")`, #378's gate reads it. Output committed at `docs/evidence/mcp-verbs/typed_path_check.txt`.

**3. `test_ghost_text.py` — green**

```
$ pytest tests/unit/tui/test_ghost_text.py -q -n0
46 passed
```

Including the `/mcp ` + `l` case, the one place my `list` verb changes behaviour another PR's tests assert on. #378 had already rewritten it to derive the expectation from the highlighted row instead of the literal `ogin`, so my verb set and their ghost text agree.

**4. Frames re-captured on the rebased head** — `dca2f99f`

`typed-verbs.png` and `typed-remove.png` are re-taken by TYPING against a tree where #378 is on main, not a scratch merge. I viewed both rendered. The `remove` frame shows all four servers with their source files including the `~/.claude.json` row, **and** the shipped ghost text (`remove `**`filesystem`** dimmed inline) — the two PRs' surfaces composing correctly in one frame.

**5. Gates on `dca2f99f`** (whole tree, as CI)

- flake8 exit 0; black 26.1.0 clean (**586 files**); isort 5.13.2 exit 0 (read via exit code with unfiltered output).
- pyright **0 errors, 0 warnings, 0 informations**.
- Unit suite: **7875 passed, 15 skipped, 0 xfailed**. 24 failed, all `tests/unit/tools/test_eval_tool.py` — the known editable-install artifact, reproducing identically on clean base. **Zero non-eval failures**, and `test_server_models.py` is clean now that main made it self-contained. `test_attachment_utils.py::test_save_fails_if_cannot_write_file` did not fail this run.
- CI green on `dca2f99f`: lint, type-check, pip-audit, cli-sanity, server-sanity, test (3.12), test (3.13).

---

Not merging: the rebase moved the head, so round 3's freshness no longer holds against `dca2f99f`. Handing back for the final freshness pass.

damianvtran:
### Agent review — round 4

Scope: `ee46f13641df327eba328e032b7e7f8a86d6051e..dca2f99fe15165f01cf7c9cd86626e36aec314e1`

Freshness pass on the rebased head. Rounds 1-3 are settled and were not re-litigated; this round covers the post-rebase delta and the regressions a rebase could cause. Because `ee46f136` predates the rebase, the literal two-dot diff also contains merged #378; the reviewable delta is `85caf5c5..dca2f99f` (7 files: the typed test additions, the xfail unmark, `shot.py`, and re-captured evidence).

**Verified, not taken on trust:**

**1. The xfail removal (the load-bearing check).** `10846533` is a pure 30-line deletion of three `@pytest.mark.xfail` blocks with **no test-body edits** — confirmed by the commit stat (`30 deletions`, 0 insertions) and by diffing the bodies. `grep -c pytest.mark.xfail` across `tests/` is now 0 (was 3). All three pass unmarked as ordinary tests.

Critically, I did not stop at "they pass" — I confirmed they **fail when the behaviour is broken**, by mutating the merged #378 source in a scratch worktree:

| mutation | result |
|---|---|
| refresh key reverted to token-only (`first_tok.lower()` without `{sep}`) | `test_mcp_server_rows_are_reachable_by_typing...` and `test_the_destructive_gate_is_armed_on_the_TYPED_path` **FAIL** (`typing '/mcp remove ' gave []`) |
| `_argument_is_destructive` row-`alert` read stubbed to `False` | `test_a_fuzzy_mcp_remove_row_fills_rather_than_deleting_a_server` and `..._TYPED_path` **FAIL** (`gate=False, expected True`; `assert '' == '/mcp remove filesystem'`) |

Each of the three dies under the mutation matching the dependency its xfail named. These are not accidental XPASSes and they are not green tests guarding nothing. Source restored to head afterwards (`git diff --quiet` clean).

I also confirmed the tests import from the worktree tree (`local_operator.__file__` resolves under the worktree), so the mutations were genuinely exercised rather than shadowed by the editable install.

**2. `typed_path_check.py` reproduced** on the rebased head under a temp `HOME`. Output is byte-identical to the committed `.txt`, including the two transitions that need both PRs: `reauth `/`reauth lnr` gate `True`, `login ` gate `False`, bare `/mcp ` `False`, and every verb→server transition opening (`open=True`).

**3. Regression spot-check on settled work** — the rebase dropped nothing. M2's `alert=verb in ("logout", "reauth")` is present (`app.py:15082`); M1's priority-before-ownership ordering is intact, verified by *running* `add_ownership_check.py`: the `~/.claude.json` shadow refusal, the higher-priority `.mcp.json` no-effect refusal, the M3 OAuth hint, and the N1/N2 trailing-token refusals all behave, with before/after effective-server lines showing the no-ops are real.

**CLI A/B with streams split.** The committed `cli_check.sh` merges with `2>&1`, so I re-ran it capturing stdout and stderr **separately** against base `ae266b85` and head `dca2f99f`. Byte-identical across stdout, stderr, and exit codes once the random temp-`HOME` path is normalised (the only diff was the `mktemp` directory name). All six error strings remain on stderr, not stdout — the CLI boundary is unmoved.

**4. Frames.** `shot.py` now types with `pilot.press` per character (the `_set_editor_line` import is gone), so the stills show a state the UI actually produces. `typed-remove.png` shows four `remove` rows with source files including `~/.claude.json` on `notion`, the danger tint, and #378's inline ghost text — consistent with the typed harness output.

**Gates on the rebased head:** flake8 0, black clean (586 files), isort clean (repo config, as CI), pyright 0 errors, and CI green on all 14 checks.

**Suite note (not a finding).** A full `tests/unit/tui` run failed once on `test_subagent_view.py::test_history_unavailable_and_error_retry_keep_trajectory_fallback`. I chased it rather than waving it through: this PR touches neither `subagent_view.py` nor its test (`git log ae266b85..dca2f99f --` on both is empty), it passes in isolation on both trees, and I reproduced it **twice on the clean base `ae266b85`** across repeated runs. Pre-existing xdist-ordering flake, unrelated to #377. Worth a follow-up issue, not a blocker here.

**Findings:** none. No blocker, no major, no minor. The one cosmetic oddity I saw (`/private~/proj/.mcp.json` in the ownership harness output) is an artifact of that script's own `str(_home)` replace against the `/private` symlink — evidence-script only, no product code, and in a settled-round file outside this delta. Not worth a remediation round.

**Verdict: clean. This round is TERMINAL.** The delta is test and evidence work only, the three unmarked tests are proven to fail when the behaviour they guard breaks, and the reviewed head equals the current head `dca2f99f`. No further round is required before merge.

PR #359, opened 2026-08-28
perf(tui): coalesce subagent progress updates

damianvtran:
## Problem

After v0.42.8 bounded long-session state work, a retained parent session could still lag while it was visibly parked on `wait` and one child streamed. The remaining path duplicated each throttled child progress boundary:

- `report_progress` notified both frontend publication and durable roster persistence, even though `latest_details` and rolling output are intentionally absent from the resume projection;
- `observe_event(SubagentProgressEvent)` performed a full `refresh_from_session` before the existing 50 ms jobs coalescer published the same row;
- the owner TUI repainted immediately from the raw progress event and then repainted again from canonical state, while remote followers used canonical state only.

Actual retained-session evidence was inspected read-only and sanitized to structural identifiers and sizes: parent session `835fbcafdc27` had a roughly 2.34 MiB transcript, 101 retained child records, a roughly 1.19 MiB roster sidecar, and one active child whose transcript was 98 entries / roughly 391 KiB at capture time. No retained session data was mutated.

## Change

- Split job notifications into durable roster changes and transient live changes.
  - Registration, queue transitions, model/usage changes, cancellation, and terminal status/outcome still publish and schedule roster persistence.
  - Progress strings and rolling output publish to frontends without incrementing the roster generation or scheduling sidecar fingerprint/write work.
- Remove `subagent_progress` from the full-session fallback in `FrontendStateStore.observe_event`; the existing 50 ms jobs coalescer owns canonical progress publication.
- Make raw owner-side `SubagentProgress` a no-op for repaint. The canonical jobs update is now the single visual invalidation for both owner and remote follower, with measured visibility under 250 ms.
- Keep the existing spinner cadence. Direct invalidation profiling showed it already repaints one row only and does not lay out the transcript.
- Add portable OSWorld-shaped benchmark and scheduling/spinner probes.
- Bump the patch version from `0.42.9` to `0.42.10`.

## Change classification

**C2, standard/pre-authorized.** This is a bounded performance and reliability correction with no schema migration, security-boundary change, or irreversible rollout. The PR is the system of record; no RFC or tracker was requested.

## Delivery contract

- Child activity remains perceptibly live on both the owner and remote followers through the same canonical update, within 250 ms.
- Transient progress/output never schedules durable roster fingerprint or write work.
- Durable launch, queue, model, usage, cancellation, settlement, and outcome state still survives resume.
- A burst produces one canonical jobs publication per 50 ms coalescing window and no raw owner-only repaint.
[...]

PR #376, opened 2026-08-28
fix(tui): attach a clipboard image on an empty paste

damianvtran:
# PR Description

## Summary of Changes

On macOS, taking a screenshot with `Cmd+Shift+Ctrl+4` and pressing `Cmd+V` in
the composer attached nothing and said nothing. The pasteboard held PNG bytes,
the terminal had no text to bracket-paste, and `Editor._on_paste` received
`Paste("")` — which fell through the path-only ingestion into the base
`TextArea._on_paste`, inserting an empty string. The keystroke was
indistinguishable from a dead key.

This adds a system-clipboard read on the empty-paste branch, with four
peer platform backends, and gives a paste that attaches nothing something to say.

Credit to @bbqben, whose issue report did most of the analytical work: the
diagnosis, the `XTermParser` verification that the empty `Paste` event actually
fires, the `«class furl»` observation for Finder, and the `pbpaste`-returns-zero
measurements are all from #372 and all held up under independent verification.

### Why this shipped looking complete

The `editor.py` docstrings credited **Ghostty** with writing
`$TMPDIR/clipboard-<stamp>-<hash>.png` and bracket-pasting the filename. That is
false, and the issue is right that it is what made the gap invisible. The thing
actually doing it is **cmux**, verified on this machine: putting a PNG on the
pasteboard caused `$TMPDIR/clipboard-2026-08-28-162947-3BD455F1.png` to appear,
byte-identical to the source. Ghostty's binary contains no such string — its
paste is text-only. So the path-only design worked perfectly in the terminal
this code was developed in and nowhere else. Those docstrings are corrected,
including the module docstring of `test_paste_images.py`.

### What was added

**`local_operator/clipboard.py`** (new). One dispatch, four peer backends, no
platform fast path:

- **macOS** — `osascript` against `NSPasteboard` for `public.png`, falling back
  to `public.tiff` re-encoded through `NSBitmapImageRep` (several macOS apps put
  only TIFF on the pasteboard, and no provider accepts TIFF). No new dependency:
  no `pngpaste`, no PyObjC.
- **Wayland** — `wl-paste --list-types` then `wl-paste --type <mime>`. The
  listing comes first because asking for a type that is not on offer can block
  while the offer is negotiated.
- **X11** — `xclip -selection clipboard -t <mime> -o`, `CLIPBOARD` and not
  `PRIMARY`.
- **Windows** — PowerShell reads the clipboard bitmap and PNG-encodes it to a
  file.

[...]

PR #379, opened 2026-08-28
fix(tui): stream the subagent view in place instead of remounting it

damianvtran:
## Problem

The subagent page flickers while a child streams: the transcript alternates between a narrow fold and full width, the working line blinks to black, and the page never settles into the steady stream the main conversation shows. It is worst around a child's final response and on long trajectories.

The cause is structural, not cosmetic. `SubagentView` never streams — it re-derives the whole transcript on every refresh tick and **remounts the blocks**, where the main conversation mutates one long-lived block in place (`OperatorApp._ensure_streaming_block`, which holds `_streaming_block` and calls `update_text` per delta).

`_sync_body` diffed by **value** over frozen `SubagentEntry` dataclasses: it kept the common prefix and tore down the entire suffix. A growing message never compares equal to its previous fold, so its block was destroyed and rebuilt on every tick. Three consequences, each matching one reported symptom:

- **Narrow-then-wide snap.** `entry_block` built the `AssistantBlock` *unmounted*, so `_flat_width()` fell to `FALLBACK_WIDTH = 80` and `_apply_rows` pinned the height to an 80-column fold. `on_resize` re-folded to the real width a frame later. Every tick painted one frame at 80 columns and the next at the true width.
- **Blink to black.** The tail `WorkingBlock` sits below the changed text row, so the suffix teardown took it too. It was a new widget every tick, restarting its shimmer sweep and its clock, and the row was absent entirely for the frame between `remove()` and the next paint.
- **Worst on the final response and long runs.** Each remount discarded the incremental markdown cache (`_frozen_text`/`_frozen_flat`), so `finalize_text()` re-lexed the whole message every tick instead of appending to a frozen prefix. Separately, a row settling mid-list rebuilt the whole suffix beneath it, which is O(n²) on parallel tool batches.

PR #359 reduced refresh *frequency*; it left the remount-per-refresh architecture intact, which is why the flicker survived it.

## Change

**`subagent_view.py` — stream in place.** `_sync_body` now reconciles by **identity** (`SubagentEntry.key`) rather than by value, updating rows through a new `update_entry_block`: text grows via `AssistantBlock.update_text` (preserving the frozen-prefix cache), tools settle through the existing `ToolCard.restore` seam with the executor's authoritative `duration_s`, notices restate. Remount is reserved for genuine reordering or a row changing kind. The tail row is split out of the diffed sequence and **pinned** via `TranscriptView.pin_tail` — the same lever the main conversation uses — so rows mount above it instead of replacing it.

Settling is keyed on **job status, not text**: this page has no `message_end`, and the refresh that observes a child stopping usually carries no new text, so a text-keyed rule left a finished child wearing the splice's fold and missing the paragraph break the main transcript shows.

**`transcript.py` — fold at the width you will actually get.** New `TranscriptBlock.fold_width(fallback)` resolves the width a block is *about to* be laid out at, walking from most authoritative down: own size → container size → the parent's `scrollable_content_region` (exactly what a `width: 1fr` child of the transcript receives) → an optional `set_fold_hint` for blocks built fully detached → the caller's fallback. `app.console.width` is deliberately **not** a rung: that is the whole terminal (175 in the harness), never the block's column, so a block trusting it would fold too wide and clip.

**`assistant.py`, `tool_card.py`** route their fold width through that ladder instead of `self.size.width or FALLBACK_WIDTH`; `WakeBlock._refresh_row` too.

The height-pin invariant in `_apply_rows` is unchanged — it still pins `styles.height` to the row count it is handed and lays out only when that count moves. Only the width those rows are folded at is better informed, so no frame exists where a block has no height.

## Change classification

**C2, standard/pre-authorized.** A bounded rendering and performance correction in one observability surface plus the width ladder it shares with the transcript family. No schema migration, no security boundary, no irreversible rollout.

## Testing evidence

Probes drive the **real `OperatorApp`** through `run_test`, opening the page with `_open_subagent_view` and ticking it with `_refresh_subagent_view`, instrumenting `TranscriptView.append_block`/`remove_block`. Before figures come from a worktree at `origin/main` (fa62d097), after figures from this branch.

| Probe (100x30 unless noted) | Before | After |
| --- | ---: | ---: |
| 120 streaming ticks: mounts / removes | 240 / 239 | **1 / 0** |
| `AssistantBlock` identities over those ticks | 96 | **1** |
| `WorkingBlock` identities | 95 | **1** |
| `_built_width` at mount → settled (140 cols) | 80 → 134 | **134 → 134** |
| Painted frames showing a fallback-width fold | 2 | **0** |
| Parallel batch, mounts at 2 / 4 / 8 / 16 pairs | 6 / 20 / 72 / 272 | **0 / 0 / 0 / 0** |
| …mounts per settling row | 3.0 / 5.0 / 9.0 / 17.0 | **0.0** |

The re-lex cost that made long final responses worst is addressed by the identity result above: block identity is now 1, so `_frozen_text`/`_frozen_flat` survive every tick instead of being discarded. Measured remount-vs-reuse per tick on `origin/main` was 1.11 ms vs 0.31 ms at 290 chars (3.6×) rising to 7.91 ms vs 0.62 ms at 8,700 chars (12.7×) — that gap is what the growing final response was paying.

Reader-facing invariants were checked directly, not assumed: scroll position holds across a refresh for a reader scrolled up (`scroll_before == scroll_after == 8`, both trees).

### Visual validation

Consecutive mid-stream frames at 140x34, captured per `AGENTS.md` "Visual validation" and viewed rendered, before and after: `/tmp/subagent-frames/{before,after}-f{1,2,3}.svg`. Consecutive settled frames (f2→f3) differ in **only** the two spinner glyphs in both trees — no reflow between paint and settle. The after frames show the shimmer live on the `thinking` tail and the settled paragraph break present.

### Regression tests

Five tests added to `tests/unit/tui/test_subagent_view.py`: a streaming message keeps one block identity; the working tail survives every refresh and stays last; settling mid-list tools cause zero mounts/removes; every row folds at the body width it is mounted into; a settled message is committed in the block it streamed into.

All five **fail on `origin/main`** and pass here. Verified independently by copying the new test file onto a clean baseline worktree:

```
FAILED test_a_streaming_message_keeps_one_block_across_refreshes - the streaming block was rebuilt
FAILED test_the_working_tail_survives_every_refresh - the working line was rebuilt under the reader (assert 6 == 1)
FAILED test_a_settling_mid_list_tool_does_not_rebuild_the_rows_beneath_it - 20 rows remounted (assert 20 == 0)
FAILED test_a_row_folds_at_the_body_width_it_is_mounted_into - assert [80, 80, 135]
```

`[80, 80, 135]` is the width flash itself, captured as an assertion.

### Gates

`flake8` ✅ · `black --check` ✅ (569 files) · `isort --check` ✅ · `pyright` ✅ on the changed files (a full-tree run was abandoned after three concurrent pyright processes on this machine pushed it past 50 minutes; CI runs the full tree) · `tests/unit/tui` ✅ · touched-area suites (`subagent_view`, `stream_anchor`, `tool_card`, `transcript_selection`, `snapshot`) ✅ 334 passed.

Two **pre-existing** flakes, both reproduced on a clean baseline worktree with this change absent, and left alone as out of scope:

- `test_app_pilot.py::test_switching_confirms_access_instead_of_warning_about_it` — 2/3 baseline runs
- `test_subagent_view.py::test_history_unavailable_and_error_retry_keep_trajectory_fallback` — 1/12 baseline runs

Under heavy machine load (load average >40, several agents building concurrently) `test_app_pilot.py` sheds a handful of boot-timing tests on baseline as well — 5 failures on an untouched `origin/main` checkout in the same window.

## Non-goals

- Transcript virtualization or a compositor redesign.
- Redesigning the child page's layout or information architecture.
- The bare `size.width or 80` in `approval.py`, `image_block.py`, `key_prompt.py` and the `settled_rows`/`_multirow_cache` measurement helpers: out of slice, and none is on this flicker path.
- The two pre-existing flakes above.

### DIFF
--- a/local_operator/tui/widgets/assistant.py
+++ b/local_operator/tui/widgets/assistant.py
@@ -362,8 +362,21 @@ class AssistantBlock(TranscriptBlock):
         self._apply_rows(self._flat_rows(self._flat_width()))
 
     def _flat_width(self) -> int:
-        """The width the rows are built at — the block's own once laid out."""
-        return self.size.width or FALLBACK_WIDTH
+        """The width the rows are built at — the block's own once laid out.
+
+        Before layout it is the width this block is ABOUT to be given, which
+        :meth:`TranscriptBlock.fold_width` derives from the container it was
+        appended into. Going straight to :data:`FALLBACK_WIDTH` here is what
+        made every mount-then-stream path fold at 80 columns, pin that fold as
+        the block's height, and re-fold one frame later when ``on_resize``
+        landed: at 140 columns the block measurably built at 80 and settled at
+        134, which a reader sees as the message flashing narrow. The pin is
+        untouched — ``_apply_rows`` still pins the count of whatever rows it is
+        handed — so the invariant that a self-authoring block is never MEASURED
+        holds exactly as before; only the width those rows are folded at is
+        better informed.
+        """
+        return self.fold_width(FALLBACK_WIDTH)
 
     def _flat_console(self) -> Console | None:
         """The app's console, or ``None`` when this block is detached.
--- a/local_operator/tui/widgets/subagent_view.py
+++ b/local_operator/tui/widgets/subagent_view.py
@@ -1730,15 +1730,34 @@ class SubagentView(Vertical):
                 new_blocks = [entry_block(entry, fold_width=fold) for entry in new_entries]
                 # TWO LISTS, ONE INDEX. `prefix` counts entries, which exclude
                 # the truncation note; `insert_blocks` indexes the BODY's list,
-                # which holds it at 0 because it is appended before any row
-                # (`_sync_body` mounts it outside the diffed sequence and keeps
-                # it out of `self._blocks`). Measured on a truncated
-                # trajectory: `view._blocks` 168 against a body of 169. Without
-                # the offset a history page prepended to a truncated child
-                # lands one row too high — above the note that is supposed to
-                # head the page. Both conditions are needed at once, which is
-                # why it survived unnoticed (review round 1, N1).
-                head_offset = 1 if self._head_block is not None else 0
+                # which also holds the note (`_sync_body` mounts it outside the
+                # diffed sequence and keeps it out of `self._blocks`). Measured
+                # on a truncated trajectory: `view._blocks` 168 against a body
+                # of 169. Without the offset a history page prepended to a
+                # truncated child lands one row too high — above the note that
+                # is supposed to head the page. Both conditions are needed at
+                # once, which is why it survived unnoticed (review round 1, N1).
+                #
+                # Keyed on the note's POSITION, not its existence. It is
+                # created lazily and appended at whatever length the body has
+                # reached, so it heads the list only when the child was already
+                # truncated at open. A child that crosses the cap while the
+                # page is open mounts it mid-list (measured: index 5 of 257),
+                # and correcting for a note that is not above the rows pushes
+                # the prepended page one row too low instead — which is how
+                # this arrived as a misordering the previous fix introduced
+                # (`durable 40` above `durable 0`) in a case `origin/main` had
+                # right (review round 2, M3).
+                mounted_blocks = self._body.blocks()
+                head_offset = (
+                    1
+                    if (
+                        self._head_block is not None
+                        and mounted_blocks
+                        and mounted_blocks[0] is self._head_block
+                    )
+                    else 0
+                )
                 self._body.insert_blocks(prefix + head_offset, new_blocks, anchor_offset=anchor)
                 self._blocks[prefix:prefix] = new_blocks
                 self._entries[prefix:prefix] = new_entries
--- a/local_operator/tui/widgets/subagent_view.py
+++ b/local_operator/tui/widgets/subagent_view.py
@@ -98,6 +98,23 @@ from local_operator.tui.widgets.transcript import (
 #: pinned by a test, so the two halves of the contract cannot drift silently.
 TRAJECTORY_MAX_EVENTS = 500
 
+#: Refreshes the first reconcile may wait for the body to be LAID OUT before it
+#: gives up and folds at the fallback width anyway.
+#:
+#: `on_mount` runs before Textual has assigned the body a region, so the width
+#: every block would fold at reads 0 and `TranscriptBlock.fold_width` falls to
+#: its 80-column fallback: the page builds its first rows at 80, PINS that
+#: height, and re-folds a frame later, which a reader sees as the message
+#: flashing narrow on opening a page mid-stream. Measured at 140x34: 2 of 20
+#: opens painted the 80-column build.
+#:
+#: Two, because the ordering that needs the most is mount-then-fill-in-one-beat
+#: (`_open_subagent_view`'s deferred fill winning its race with layout), which
+#: measured exactly two refreshes before the body reported 135. It is a bound
+#: rather than a loop so a body that never gains a width paints its rows at the
+#: fallback instead of never painting them at all.
+SYNC_LAYOUT_WAITS = 2
+
 #: Rows of a delegated brief the page shows before folding the rest away. Three
 #: carries the opening sentence or two — enough to say what was asked — while
 #: leaving the child's own work the majority of a 60-column body. Measured
@@ -246,6 +263,21 @@ class SubagentEntry:
     #: diffed sequence, so its arrival at the cap appends a row instead of
     #: renumbering every entry and rebuilding the page under the reader.
     head: bool = False
+    #: Is this row DONE GROWING? A text row that has had its ``message_end``,
+    #: or a durable history row, which is complete by definition.
+    #:
+    #: Distinct from the job being settled, and that distinction is the point:
+    #: a running child's transcript is mostly FINISHED messages — the prose it
+    #: writes between tool calls — and deriving completeness from the job
+    #: status alone left every one of them unfinalized for the whole run. They
+    #: then rendered with the streaming splice's concatenated fold, which
+    #: cannot produce the blank row between two paragraphs, and because a
+    #: block PINS its height to the rows it authored they were pinned short
+    #: and grew by one row each when the job finally settled (measured 5 -> 6
+    #: and 2 -> 3 on a two-message child) — a reflow of exactly the kind this
+    #: page's in-place reconciliation exists to remove. The stream carries the
+    #: fact; it was only being discarded at the fold.
+    complete: bool = False
 
 
 def _notice(key: str, text: str, kind: NoticeKind = "info", *, head: bool = False) -> SubagentEntry:
@@ -265,6 +297,16 @@ def _supersedes(new: SubagentEntry, old: SubagentEntry) -> bool:
     if new.kind != old.kind:
         return True  # the row changed shape; the newer reading is the one to trust
     if new.kind == "text":
+        if old.complete and not new.complete:
+            # COMPLETENESS IS AS LOSABLE AS TEXT. `message_end` is an event
+            # like any other, so the rolling window evicts it too: a later
+            # fold of the same message then reports it as still streaming.
+            # Accepting that would un-finalize a committed block, and
+            # `update_entry_block` answers a finalized block's text change
+            # with a REMOUNT — so a row that had settled correctly would be
+            # torn down and rebuilt mid-read. Same rule as the text length
+            # below, applied to the other thing a fold can lose.
+            return False
         return len(new.text) >= len(old.text)
     if new.kind == "tool":
         # A settled outcome supersedes a live one; the reverse never does.
@@ -342,8 +384,16 @@ def fold_transcript_entries(
                 )
             else:
                 label = "Subagent · replied" if details.get("reply_to") else "Subagent"
+                # Renders as an AssistantBlock, and a recorded communication is
+                # whole the moment it exists — it never streams — so it settles
+                # on arrival rather than waiting for the job.
                 folded.append(
-                    SubagentEntry(f"comm:{entry.id}", "subagent_message", text=f"{label}\n\n{body}")
+                    SubagentEntry(
+                        f"comm:{entry.id}",
+                        "subagent_message",
+                        text=f"{label}\n\n{body}",
+                        complete=True,
+                    )
                 )
             continue
         if entry.type != ENTRY_MESSAGE:
@@ -365,7 +415,11 @@ def fold_transcript_entries(
             continue
         if role == "assistant":
             if text:
-                folded.append(SubagentEntry(entry.id, "text", text=text))
+                # A durable row is a message the engine already committed to
+                # the transcript, so it is complete whatever the job is doing
+                # now. Saying so here is what stops a paged-in history message
+                # rendering with the streaming fold on a live page.
+                folded.append(SubagentEntry(entry.id, "text", text=text, complete=True))
             for raw_call in payload.get("tool_calls") or ():
                 if not isinstance(raw_call, Mapping):
                     continue
@@ -419,6 +473,10 @@ def fold_trajectory(events: Sequence[Any], *, settled: bool = False) -> list[Sub
     # represented separately; the set keeps the same first-seen contract in O(1).
     remembered: set[tuple[str, str]] = set()
     streams: dict[str, str] = {}
+    #: Message ids whose ``message_end`` arrived in THIS window. The child is
+    #: done writing them even while it goes on working, which is what lets the
+    #: page finalize them per row instead of waiting for the job to settle.
+    finished: set[str] = set()
     tools: dict[str, SubagentEntry] = {}
     notices: dict[str, SubagentEntry] = {}
 
@@ -464,6 +522,7 @@ def fold_trajectory(events: Sequence[Any], *, settled: bool = False) -> list[Sub
                 if message_id not in streams and not text:
                     continue
                 streams[message_id] = text
+                finished.add(message_id)
                 remember("text", message_id)
         elif etype == "tool_execution_start":
             # One normalisation for the whole pair. The lookup below used to
@@ -548,7 +607,18 @@ def fold_trajectory(events: Sequence[Any], *, settled: bool = False) -> list[Sub
         if kind == "text":
             text = strip_control_sequences(streams.get(key, "")).strip()
             if text:  # a tool-use message carries no prose — spend no row
-                rows.append(SubagentEntry(key=key, kind="text", text=text))
+                # `settled` finishes the LAST message too: a child killed
+                # mid-turn never sends its `message_end`, and that row is done
+                # growing for the only reason that matters — nothing is left
+                # to send it deltas.
+                rows.append(
+                    SubagentEntry(
+                        key=key,
+                        kind="text",
+                        text=text,
+                        complete=key in finished or settled,
+                    )
+                )
         elif kind == "tool":
             entry = tools[key]
             if settled and entry.outcome is None:
@@ -717,11 +787,17 @@ def entry_block(
     re-folds one frame later, which a reader sees as the row flashing narrow.
     Zero means "not supplied" and keeps the old fallback behaviour.
 
-    ``settled`` says whether this text row is done growing. A row that may
-    still receive deltas must NOT be finalized: a finalized ``AssistantBlock``
-    ignores ``update_text``, so finalizing on construction is what forced the
-    page to rebuild a streaming message instead of updating it — and a rebuild
-    discards the incremental markdown cache and re-lexes the whole message.
+    ``settled`` says whether the JOB has stopped. A text row is committed when
+    it is done growing, which is ``entry.complete or settled`` — the row's own
+    ``message_end`` if it had one, else the job stopping (the last message of a
+    child killed mid-turn never gets one). A row that may still receive deltas
+    must NOT be finalized: a finalized ``AssistantBlock`` ignores
+    ``update_text``, so finalizing on construction is what forced the page to
+    rebuild a streaming message instead of updating it — and a rebuild discards
+    the incremental markdown cache and re-lexes the whole message. Finalizing
+    per ROW rather than per PASS is what keeps a message that finished ten
+    tool calls ago from rendering with the streaming splice's fold, missing
+    its paragraph breaks, until the child eventually exits.
     """
     if entry.kind == "prompt":
         return InstructionBlock(entry.text)
@@ -732,7 +808,7 @@ def entry_block(
         if fold_width:
             block.set_fold_hint(fold_width)
         block.update_text(entry.text)
-        if settled:
+        if entry.complete or settled:
             block.finalize_text()
         return block
     if entry.kind == "notice":
@@ -832,15 +908,17 @@ def update_entry_block(
             return False
         if entry.text != previous.text:
             block.update_text(entry.text)
-        if settled:
-            # The child STOPPED, which is a change to this row even when its
-            # text did not move: a live block carries the streaming splice's
-            # concatenated fold, and committing re-renders the whole message
-            # once — which is what restores the blank row between paragraphs
-            # that the splice cannot produce mid-stream. Checked on every pass
-            # rather than only on a text change, because the refresh that
-            # observes a job settle usually carries no new text at all;
-            # `finalize_text` is itself idempotent, so this runs once.
+        if entry.complete or settled:
+            # This ROW stopped growing — its own `message_end` arrived, or the
+            # job stopped and nothing is left to send it deltas. Either is a
+            # change to the row even when its text did not move: a live block
+            # carries the streaming splice's concatenated fold, and committing
+            # re-renders the whole message once — which is what restores the
+            # blank row between paragraphs that the splice cannot produce
+            # mid-stream. Checked on every pass rather than only on a text
+            # change, because the refresh that observes a message end or a job
+            # settle usually carries no new text at all; `finalize_text` is
+            # itself idempotent, so this runs once.
             block.finalize_text()
         return True
     if entry == previous:
@@ -1145,6 +1223,16 @@ class SubagentView(Vertical):
         #: the container to mount.
         self._pending: list[SubagentEntry] = []
         self._pending_head: SubagentEntry | None = None
+        #: Refreshes the first reconcile may still wait for a laid-out body.
+        #: `on_mount` fires before the body has a region, so every width the
+        #: page could fold at reads 0 there and `fold_width` falls to its
+        #: 80-column fallback; `_sync_body` postpones instead and replays once
+        #: the body has a real width. It is a BUDGET rather than a flag
+        #: because one refresh is not always enough — the mount-then-fill
+        #: ordering needs two — and it is bounded rather than open so a body
+        #: that never gains a width degrades to the old fallback fold instead
+        #: of postponing the page's only content forever.
+        self._sync_layout_waits = SYNC_LAYOUT_WAITS
         #: Last painted title inputs. The spinner ticks eight times a second
         #: and everything else on the row moves once a second at most.
         self._chrome_state: tuple[Any, ...] | None = None
@@ -1640,7 +1728,18 @@ class SubagentView(Vertical):
                 # same body width every other block on this page does.
                 fold = self._body.scrollable_content_region.width
                 new_blocks = [entry_block(entry, fold_width=fold) for entry in new_entries]
-                self._body.insert_blocks(prefix, new_blocks, anchor_offset=anchor)
+                # TWO LISTS, ONE INDEX. `prefix` counts entries, which exclude
+                # the truncation note; `insert_blocks` indexes the BODY's list,
+                # which holds it at 0 because it is appended before any row
+                # (`_sync_body` mounts it outside the diffed sequence and keeps
+                # it out of `self._blocks`). Measured on a truncated
+                # trajectory: `view._blocks` 168 against a body of 169. Without
+                # the offset a history page prepended to a truncated child
+                # lands one row too high — above the note that is supposed to
+                # head the page. Both conditions are needed at once, which is
+                # why it survived unnoticed (review round 1, N1).
+                head_offset = 1 if self._head_block is not None else 0
+                self._body.insert_blocks(prefix + head_offset, new_blocks, anchor_offset=anchor)
                 self._blocks[prefix:prefix] = new_blocks
                 self._entries[prefix:prefix] = new_entries
                 self._pending = entries
@@ -1706,6 +1805,17 @@ class SubagentView(Vertical):
         rows.insert(1, getattr(self._breadcrumb, "renderable", ""))
         return [_plain(row) for row in rows]
 
+    def _replay_pending_sync(self) -> None:
+        """Reconcile the fold that was postponed for want of a laid-out width.
+
+        Runs one refresh after the deferral, by which point Textual has given
+        the body its region. It reads ``_pending`` rather than closing over the
+        entries it was scheduled with, so a refresh that lands in the gap wins:
+        the page paints what is CURRENT at the moment it can paint correctly,
+        never a snapshot that is already one tick stale.
+        """
+        self._sync_body(self._pending, self._pending_head)
+
     def _sync_body(self, entries: list[SubagentEntry], head: SubagentEntry | None) -> None:
         """Bring the mounted blocks into agreement with ``entries``.
 
@@ -1734,6 +1844,10 @@ class SubagentView(Vertical):
         scrolled up is not moved, because an in-place update mounts nothing;
         and no row is ever dropped, because the fold that produced ``entries``
         has already been through :func:`_supersedes`.
+
+        The FIRST pass is deferred by a frame when the body has not been laid
+        out yet — see the guard below. Reconciling into a zero-width body is
+        not cheaper than waiting, it is just wrong a frame earlier.
         """
         if not self._body.is_mounted:
             return  # `on_mount` replays the pending fold once the body exists
@@ -1743,9 +1857,40 @@ class SubagentView(Vertical):
         # DETACHED and would otherwise fold at 80 columns and re-fold a frame
         # later, visibly.
         fold = self._body.scrollable_content_region.width
+        if not fold and entries and self._sync_layout_waits > 0:
+            # MOUNTED IS NOT LAID OUT. `on_mount` runs before Textual has
+            # assigned this body a region, so every width the page could ask
+            # for — the body's size, its container, the view's own — reads 0,
+            # and a fold of 0 sends `fold_width` to its 80-column fallback.
+            # Measured over 20 opens at 140x34: 2 of them built the page's
+            # first blocks at 80 and re-folded them to 135 a frame later, and
+            # the compositor painted the 80-column build, which is the width
+            # flash on opening a page mid-stream.
+            #
+            # The screen's width is the one non-zero source here (138) and it
+            # is deliberately NOT used: it is the whole terminal, 3 cells wider
+            # than the body's 135, so a hint from it folds too wide and clips.
+            # Waiting one frame is the only way to build at a width that is
+            # RIGHT rather than merely non-zero. It costs a frame of empty
+            # body, not a reflow — the rows appear once, already correct,
+            # instead of appearing narrow and re-folding under the reader.
+            #
+            # A BUDGET, not a one-shot: the mount-then-fill ordering needs two
+            # refreshes before the body has a region, and a single deferral
+            # replayed into a still-zero width and folded at 80 anyway. Bounded
+            # so a body that never gains a width (a page closed under the
+            # deferral, a zero-height layout) falls through to the old fallback
+            # behaviour and paints SOMETHING rather than postponing forever.
+            self._sync_layout_waits -= 1
+            self.call_after_refresh(self._replay_pending_sync)
+            return
         # A row may still receive deltas while the child is working, so a text
-        # block is only committed once the job stops. Finalizing earlier is
-        # what made a streaming message unupdatable and therefore rebuilt.
+        # block is committed once the row itself is done — `entry.complete`,
+        # i.e. its `message_end` arrived — or once the job stops, which
+        # finishes the last message of a child that was killed mid-turn.
+        # Finalizing earlier is what made a streaming message unupdatable and
+        # therefore rebuilt; finalizing per PASS instead of per ROW is what
+        # left finished messages folded as if they were still streaming.
         settled = not self._running and not self._queued
         if head is not None and self._head_block is None:
             self._head_block = NoticeBlock(head.text, head.notice_kind)
--- a/local_operator/tui/widgets/subagent_view.py
+++ b/local_operator/tui/widgets/subagent_view.py
@@ -698,7 +698,9 @@ class InstructionBlock(UserBlock):
             parent.refresh_gap_around(self)
 
 
-def entry_block(entry: SubagentEntry) -> TranscriptBlock:
+def entry_block(
+    entry: SubagentEntry, *, fold_width: int = 0, settled: bool = True
+) -> TranscriptBlock:
     """One folded entry as the block the main conversation would have used.
 
     Tool rows settle through :meth:`ToolCard.restore` rather than
@@ -707,6 +709,19 @@ def entry_block(entry: SubagentEntry) -> TranscriptBlock:
     the tool took. Both trajectory end events and durable tool messages carry
     the executor's measured duration, so replay restores that authoritative
     value and leaves the column blank only for legacy rows that predate it.
+
+    ``fold_width`` is the width of the body this block is about to be mounted
+    into. Every block here bakes its width into the rows it authors AND pins
+    its height to the count of them, and each one is built DETACHED — no
+    parent to borrow a width from — so without this it folds at 80 columns and
+    re-folds one frame later, which a reader sees as the row flashing narrow.
+    Zero means "not supplied" and keeps the old fallback behaviour.
+
+    ``settled`` says whether this text row is done growing. A row that may
+    still receive deltas must NOT be finalized: a finalized ``AssistantBlock``
+    ignores ``update_text``, so finalizing on construction is what forced the
+    page to rebuild a streaming message instead of updating it — and a rebuild
+    discards the incremental markdown cache and re-lexes the whole message.
     """
     if entry.kind == "prompt":
         return InstructionBlock(entry.text)
@@ -714,8 +729,11 @@ def entry_block(entry: SubagentEntry) -> TranscriptBlock:
         return UserBlock(entry.text)
     if entry.kind in ("text", "subagent_message"):
         block = AssistantBlock()
+        if fold_width:
+            block.set_fold_hint(fold_width)
         block.update_text(entry.text)
-        block.finalize_text()
+        if settled:
+            block.finalize_text()
         return block
     if entry.kind == "notice":
         if entry.key == "__working__":
@@ -723,6 +741,8 @@ def entry_block(entry: SubagentEntry) -> TranscriptBlock:
             return WorkingBlock(activity, activity)
         return NoticeBlock(entry.text, entry.notice_kind)
     card = ToolCard("", entry.tool_name, entry.tool_args, entry.intent)
+    if fold_width:
+        card.set_fold_hint(fold_width)
     if entry.outcome == "error":
         card.restore(
             state="error",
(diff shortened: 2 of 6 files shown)


damianvtran:
### Design review — round 1

Reviewer: `designer` subagent (local-operator team `lopdev`), model `anthropic/claude-opus-5`
Scope: `origin/main..5e5b14c0` (base `fa62d097`, head `5e5b14c0`, fresh at time of review)

Judged from rendered frames only, not from source. Frames captured per `AGENTS.md` "Visual validation": the real `OperatorApp` under `run_test`, the subagent page opened via `_open_subagent_view` and ticked with `_refresh_subagent_view`, exported with `save_screenshot`, then rendered and viewed side by side. BEFORE frames come from a throwaway `origin/main` worktree (`/tmp/design-base`), AFTER frames from the PR worktree, using the **same script at the same terminal sizes**.

Two capture modes, because the flicker does not live in a settled frame:
- **Settled frames** (`shot.py`, several pauses after each tick) — six states x two sizes x three consecutive frames.
- **Transient frames** (`shot_transient.py`, exactly one pause after each refresh) — the *first painted frame of the tick*, which is where a fallback-width fold or a missing tail row would show. This is what actually reproduced the reported flicker on baseline.

Artifacts: `/tmp/design-shots/{before,after}/` — `{w,n}-{midstream,settle,tool,parallel,finished,empty}-f{1,2,3}.svg` and `{w,n}-trans-t{1..6}.svg`, where `w` = 140x34 and `n` = 80x24. Side-by-side viewer: `/tmp/design-shots/compare.html`.

---

## What the frames show

**The reported flicker reproduces on baseline and is gone here.** Transient frames at 140x34, counting painted body rows per tick:

| Tick | BEFORE body rows | AFTER body rows |
| --- | ---: | ---: |
| t1 | 12 | 12 |
| t2 | 13 | 13 |
| t3 | 15 | 15 |
| t4 | 17 | 17 |
| t5 | **13** | 18 |
| t6 | **12** | 20 |

Baseline grows to 17 rows and then collapses back to 13 and 12 while the message is still *growing* — that is the remount painting a partial rebuild, and in `before/w-trans-t5.svg` the whole first message (`Reading the ingest path.`) is simply **absent from the frame**, with the prose starting mid-sentence at `by source file and then upserts...`. The after frames grow 12 → 20 monotonically and never drop a row. Visually this is the difference between a page that jitters and one that streams.

**The tail line's animation and clock are now continuous.** Spinner glyph and elapsed clock on the `thinking` row, per transient tick:

| | t1 | t2 | t3 | t4 | t5 | t6 |
| --- | --- | --- | --- | --- | --- | --- |
| BEFORE 140x34 | `⣽ 0s` | `⣾ 0s` | `⣾ 0s` | `⣾ 0s` | `⣾ 0s` | `⣾ 0s` |
| BEFORE 80x24 | `⣻ 0s` | **ABSENT** | `⣾ 0s` | `⣻ 0s` | `⣻ 0s` | `⣽ 0s` |
| AFTER 140x34 | `⢿ 0s` | `⣟ 1s` | `⣷ 1s` | `⣽ 1s` | `⣽ 1s` | `⡿ 1s` |
| AFTER 80x24 | `⣟ 1s` | `⣷ 1s` | `⣷ 1s` | `⣻ 1s` | `⡿ 1s` | `⣷ 1s` |

Two baseline defects visible here, both fixed: the clock is pinned at `0s` for the whole run because the widget is new every tick and its stopwatch restarts, and at 80x24 the tail row is **missing entirely from `before/n-trans-t2.svg`** — the "blink to black" in the bug report, captured. After, the clock advances and the tail is present in every one of the twelve transient frames at both sizes.

**No frame folds at 80 columns.** Max painted row width across every transient and settled frame: 135 at 140x34 and 75 at 80x24, in both trees, with no intermediate narrow frame in the after tree. (Baseline's flash is one layout pass wide and the sampling here is per painted frame, so I treat the absence of a narrow after-frame as consistent with the PR's `[80, 80, 135]` assertion rather than as independent proof of it — the row-collapse above is the symptom I could see directly.)

**Settled paragraph break is present and correct at both sizes.** `after/w-settle-f2.svg` and `after/n-settle-f2.svg` both show the blank row between "The duplicates come from the retry path, not the loader." and the following paragraph. It is absent mid-stream (`f1`) and appears on the settling tick (`f2`), then `f2` and `f3` are **byte-identical** — the settle does not re-flow a second time.

I checked whether that mid-stream missing break is a regression against house behaviour by streaming the same two-paragraph body through the **main conversation's** own path (`_ensure_streaming_block` / `update_text` / `finalize_text`, `after/w-main-{streaming,settled}.svg`). The main transcript does exactly the same thing: no break while streaming, break on finalize. So the child page now *matches* the main conversation rather than diverging from it, which is the stated goal. Baseline's `w-settle-f2` also had the break, but only because it rebuilt the block wholesale every tick — it got the right pixel via the mechanism that caused the flicker.

**Gap and blank-row rules are unchanged.** I compared the blank-row structure of the body region (rule line to footer hints) frame by frame across all six states at both sizes. Identical between trees everywhere except `settle f1`, which is the intended mid-stream-vs-settled difference described above. The gap above the pinned tail line is exactly one blank row in every state at both sizes, and the row above it is the expected kind (text after prose, tool row after a tool card, rule after nothing) — splitting the tail out of the diffed sequence did not disturb it.

**Idle frames are stable.** With nothing changing, consecutive frames differ in **only** the two spinner glyphs (header and tail); the finished-child state differs in **zero** rows across three consecutive frames at both sizes. That is the pass condition for "settled means settled".

**Remaining states.** Running tool card with the tail beneath it, parallel batch at 4 running / 2 settled / 4 settled, finished child, and just-launched/empty all render identically to baseline at both sizes — same alignment, same tool-row rhythm, same `✓` right-alignment, same footer hint row, same ink. No spacing, contrast, or focus-order change anywhere.

---

## Findings

**D1 — NIT — the horizontal scrollbar cap flickers in and out across ticks.** In the 140x34 transient sequence a `▁`/`▆` glyph appears in the body's right-hand scrollbar column on some ticks and not others: baseline at t3, t5, t6 and the after tree at t6 (`after/w-trans-t6.svg`, x=1671.4, top body row). Where baseline showed it on three of six ticks the after tree shows it on one of six, so this is strictly improved, not introduced. It is a one-cell artifact of the body's virtual height crossing the scrollable threshold as content grows, it is the vertical scrollbar's end cap rather than a stray character, and at one cell it is well below what a reader notices next to a growing message. Recording it as a follow-up rather than asking for a change in this PR.

No BLOCKER, MAJOR, or MINOR findings.

---

## Verdict

**Clean. This round is TERMINAL.**

Every reported symptom reproduced on `origin/main` from rendered frames and is absent at `5e5b14c0`: the row-count collapse mid-stream, the message vanishing from a tick's frame, the clock frozen at `0s`, and the tail line missing entirely from a frame at 80x24. The page now reads as a steady, continuous stream at both terminal sizes, the settled final response gets its paragraph break at full body width, and the blank-row and gap rules between blocks of different kinds are unchanged from baseline. The one open item is D1 (NIT), carried as a follow-up.

damianvtran:
### Agent review — round 1

Scope: origin/main..5e5b14c0

Verified rather than read: the 5 new tests pass here and I confirmed the fold/identity claims by driving the real `OperatorApp` through `run_test`, with an `origin/main` (fa62d097) throwaway worktree as the baseline for every before/after number below.

The core of this change is right. Identity reconciliation is sound: `_supersedes` still gates the fold so a shrunken row cannot reach `update_entry_block`, and the `not entry.text.startswith(previous.text)` guard correctly refuses a truncating update rather than silently shrinking a message. The in-place update safety I was asked to probe holds up — `previous.outcome is not None or entry.outcome is None` genuinely prevents a settled `ToolCard` being reopened, `_started` is already `None` on every card this page builds (all are `restore`d at construction), so the fabricated-duration bug the `entry_block` comment warns about cannot return; and a finalized `AssistantBlock` is never reopened (the `is_finalized()` branch returns an agreement, never a mutation). Tail pinning is correct across all three terminating states I exercised: `running → gone` swaps the pin, and `running → completed` and `cancelled` both clear it (`_tail = None`) with `view._blocks` staying in step. `fold_width`'s rung 2 is not the over-wide hazard it looks like — Textual passes `container_size` already net of the scrollbar gutter (measured: body `size` 136, `scrollable_content_region` 135, child `container_size` 135), so the ladder cannot clip.

Two findings, both in the rendering behaviour this PR is specifically about.

---

**M1 — `subagent_view.py:1749` (and `:735`, `:122-131`): on a live page, a message that is already COMPLETE is never finalized, so it loses its paragraph breaks and the page reflows when the child settles.**

`settled` is derived once per pass from the JOB (`not self._running and not self._queued`) and applied to every text row. But a running child's transcript is mostly *finished* messages — the prose it writes between tool calls, each of which already got its `message_end`. Those rows now stay unfinalized for the whole run, so they render with the streaming splice's concatenated fold instead of the single whole-message render that produces the blank row between paragraphs.

Reproduced on a running child whose first message is `"Plan:\n\n- step one\n- step two\n\nNow running it."` and which has already moved on to a tool call:

```
branch (5e5b14c0)              origin/main (fa62d097)
  finalized: False               finalized: True
  | Plan:                        | Plan:
  |                              |
  |  • step one                  |  • step one
  |  • step two                  |  • step two
  | Now running it.              |                 <- blank row present
                                 | Now running it.
```

Two consequences, and the second is the one that matters most for this PR:

1. Every completed message on a live page is missing its paragraph breaks — the exact defect the PR description names ("missing the paragraph break the main transcript shows"), just relocated from after-settle to during-the-run, which is when the page is actually being watched.
2. Because `_apply_rows` pins height to the row count, those rows are pinned SHORT and then grow when the job settles. Measured on a child with two complete multi-paragraph messages: both blocks go `rows=2 → rows=3` on the settling refresh, so the page shifts content by one row per affected message at settle. That is a visible reflow of the kind this change exists to remove, and it is new — on `origin/main` all three blocks were already `finalized=True` with `blank_rows=1` before the settle.

The PR argues this page cannot key settling on text because it has no `message_end`. That premise is wrong: the trajectory does carry `message_end` (`_text()` in the test fixtures emits one, and `fold_trajectory` handles it at `subagent_view.py:462` — "message_end adopts the authoritative text"). The information is present and is being discarded at the fold, not absent from the stream.

Suggested fix: carry it on the row rather than inferring it from the job — add a `complete: bool = False` field to `SubagentEntry`, set it in `fold_trajectory`'s `message_end` arm, and finalize per-row on `entry.complete or settled` instead of on the job-wide `settled`. Keying settling on job status stays correct for the *last* message (the one with no `message_end` yet when the child is killed); it is only wrong as the rule for every earlier row.

---

**M2 — `subagent_view.py:1745`: on the first-open path `scrollable_content_region.width` is still 0, so `fold=0`, and blocks build at the 80-column fallback and re-fold a frame later — the width flash the PR reports as eliminated.**

`_sync_body` is called from `on_mount` (`:1241`), and at that point the body is mounted but not yet laid out. Instrumenting `_sync_body` over 6 identical runs at 140x34:

```
[(0, True, 0), (0, True, 0), (135, True, 136), ...]   <- 4 of 6 runs
[(0, True, 0), (135, True, 136), ...]                 <- 2 of 6 runs
     ^ scrollable width the fold is read from
```

Tracing it through to the fold, on the runs where the second pass also sees 0:

```
entry_block fold_width args: [0, 0, 0, 0]  |  _apply_rows widths: [80, 80, 135, 135]
entry_block fold_width args: [135,...]     |  _apply_rows widths: [135, 135]
```

So on most first opens two blocks are still built at 80 and rebuilt at 135 — `set_fold_hint` is never called because the caller's own source is 0. This is not a regression (baseline is worse: `[80, 80, 80, 80, 135, 135]`, 4 fallback folds), but the PR's "Painted frames showing a fallback-width fold: 2 → **0**" is not true for the open path, and the flicker on opening a page mid-stream is the reported symptom.

Compounding it, the test that is supposed to guard this cannot see the path: `test_a_row_folds_at_the_body_width_it_is_mounted_into` installs its `_apply_rows` probe at offset ~1653 in the test body, well AFTER `view = await _open(...)` at ~807 — so every fold performed during the initial mount is invisible to the assertion, and the test passes while the flash it names still happens. That is why this survived the round.

Suggested fix: fall back to the body's `container_size`/`size` width when `scrollable_content_region.width` is 0, or defer the first `_sync_body` by one frame (`call_after_refresh`) so the body has a width to report; and move the fold probe above `_open` so the test actually covers the mount path.

---

**N1 — `subagent_view.py:1643`: `insert_blocks(prefix, ...)` indexes a list that includes `_head_block`, but `prefix` is counted over entries that exclude it (pre-existing, adjacent to this diff).**

`self._head_block` is appended straight to the body (`:1752`) and deliberately kept out of `self._blocks`, but `TranscriptView.insert_blocks` indexes `self._body._blocks`, which does contain it at index 0. Confirmed on a truncated trajectory: `view._blocks` is 168 while the body holds 169, and the untracked block is the head `NoticeBlock`. So when a truncation note is present AND a durable history page is prepended, the insert lands one row too high. Not introduced here — the same skew exists on `origin/main` — and it needs both conditions at once, so I am not calling it a blocker for this PR. Worth a follow-up ticket rather than a fix in this round; the one-line version is to offset by `1 if self._head_block is not None else 0`.

**N2 — `tests/unit/tui/test_subagent_view.py`: `test_a_streaming_message_keeps_one_block_across_refreshes` asserts `counter.removes == 0` but never asserts on `counter.mounts`.** Block identity already covers the rebuild case, but a stray extra mount (a duplicate row appended beside the streaming one) would pass. One added `assert counter.mounts == 0` closes it; the sibling settling test already does exactly that.

---

Comment density matches the codebase — the new docstrings on `fold_width`, `update_entry_block` and `_sync_body` explain the failure that motivated the code and record the measured numbers, which is the house style, not restatement of the lines beneath them.

Gates I ran: `flake8` clean on all five changed files; the 5 new tests pass. Full `tests/unit/tui` showed 6 failures across two runs, all reproduced on the baseline worktree or shown to be load-dependent under a load average of 25-40 (`test_composer_seam.py` fails identically on `origin/main` when run in isolation), so I attribute none of them to this change.

**Verdict: changes requested.** M1 is a user-visible rendering regression against `origin/main` in the exact behaviour this PR sets out to fix; M2 leaves the headline claim unmet on the open path with a test that cannot detect it. The architecture is right and both fixes are local — no rework of the identity reconciliation is needed. Not terminal; please remediate and re-post for round 2.

damianvtran:

Implementer: coder (claude-opus-5), with the manager verifying gates and measurements independently.
Scope: `origin/main..2e5ee8f4`
Round 1 reviewed `5e5b14c0`. The branch has since been **rebased onto `origin/main`** (picking up #366 `/failovers` and #380 pyright default excludes); the reviewed code was unchanged by the rebase, only the base moved. Remediation head is **2e5ee8f4**.

Both majors were real and both are fixed at the source. One of the PR's own claims was overstated and is corrected below rather than defended.

---

**M1 — a live page never finalizes messages that are already complete. FIXED** in `2e5ee8f4`.

The finding is correct, including the part that contradicts the PR description. `fold_trajectory` does see `message_end` (line 462); the completeness was being discarded at the fold, not absent from the stream. The PR text asserting "this page has no `message_end`" was wrong and is withdrawn.

`SubagentEntry` gains a `complete` field, set when a `message_end` is folded and true for durable history rows (a committed transcript row is complete whatever the job is doing). `entry_block`/`update_entry_block` now commit a row on `entry.complete or settled` rather than on job status alone — `settled` is retained for the one case it uniquely answers, a child killed mid-turn whose last message never gets its `message_end`.

`_supersedes` gained a matching guard, which the finding did not ask for but the fix requires: **completeness is as losable as text.** The rolling window evicts `message_end` like any other event, so a later fold of the same message can report it as still streaming. Un-finalizing a committed row would answer its next text change with a remount — tearing down a settled row mid-read, the exact defect this PR removes. Covered by `test_an_evicted_message_end_does_not_uncommit_a_settled_row`, which fails without the guard.

Verified against the reviewer's own reproduction: a completed message on a live page now renders `finalized=True` with its paragraph blank row present, and no longer grows a row when the job settles.

---

**M2 — the fold claim does not hold on the first-open path. FIXED** in `2e5ee8f4`, and the claim is corrected.

Confirmed before fixing, and confirmed to be worse than "theoretical": the 80-column build **reaches painted frames**, measured on 2 of 20 opens at 140x34. `_sync_body` runs from `on_mount` before Textual assigns the body a region, so every width source reads 0 and the ladder falls to its fallback.

The first reconcile now waits a bounded number of refreshes (`SYNC_LAYOUT_WAITS = 2`) for the body to report a real width before folding. Bounded rather than a loop, so a body that never gains a width still paints its rows at the fallback instead of never painting at all.

| First-open folds, 140x34 | Before | After |
| --- | ---: | ---: |
| Guard test, deterministic ordering (3 runs) | `[80, 80, 135, 135, 135]` — fails 3/3 | `[135, …]` — passes 3/3 |
| Scheduler-dependent opens | 2 of 20 showed the 80-column build | **0 of 20** |

**On the test that could not see its own path — the finding is accepted in full.** The predecessor armed its probe after `_open()`, so every mount-path fold was invisible and it passed while the flash still happened. The replacement installs the probe before the page exists and *drives* the race deterministically rather than waiting for it: it asserts `scrollable_content_region.width == 0` before refreshing, which is the mount-then-fill-in-one-beat ordering `_open_subagent_view` produces whenever its deferred fill beats layout. Left to the scheduler this reproduces on roughly 1 open in 10, which is a coin toss, not a regression test.

Correction to the PR description: the original "zero fallback-width folds" claim was measured only on the streaming path and did not hold on first open. It holds now, on both, and the guard test can actually observe it.

---

**N1 — `insert_blocks` prefix excludes `_head_block`. FIXED** in `2e5ee8f4`.

Confirmed a real off-by-one rather than a cosmetic one: `_head_block` is mounted outside the diffed sequence but occupies index 0 of the container's block list, so a history page paged in beneath a truncation note landed **above** the note. Pre-existing, adjacent to this diff, and small enough to fix safely here. `test_a_history_page_lands_below_the_truncation_note` fails without the fix.

---

**N2 — the streaming test asserts only on removes. FIXED** in `2e5ee8f4`. `assert counter.mounts == 0` added, matching the sibling settling test.

---

### Verification

Three tests added (91 total in this file, from 83 on `origin/main`). All new and changed tests were confirmed to **fail on the pre-remediation head `6127603a`** and pass on `2e5ee8f4` — checked in a throwaway worktree, not by assertion alone:

```
FAILED test_a_finished_message_is_committed_while_the_child_is_still_working
       - assert False = is_finalized()
FAILED test_a_row_folds_at_the_body_width_it_is_mounted_into
       - assert 80 not in [80, 80, 135, 135, 135]
FAILED test_a_history_page_lands_below_the_truncation_note
       - a history page landed above the note
FAILED test_an_evicted_message_end_does_not_uncommit_a_settled_row   (guard reverted)
       - the page forgot the row was complete
```

The headline results are unchanged by this remediation — 120 streaming ticks still give **1 mount, 0 removes, 1 `AssistantBlock` identity, 1 `WorkingBlock` identity**.

Gates on `2e5ee8f4`: flake8 ✅ · black --check ✅ (569 files) · isort --check ✅ · **full-tree pyright ✅ 0 errors** · `tests/unit/tui` ✅ **2627 passed, 4 skipped, 0 failed**.

Worth noting: the full-tree pyright that round 1 could not obtain now runs in ~24s thanks to #380, so it is genuinely verified rather than deferred to CI. The TUI suite also ran clean end to end — the `test_app_pilot.py` boot-timing failures reported earlier were load artifacts on a machine at load average 25–40, and do not reproduce on an idle machine on either tree.

damianvtran:
### Agent review — round 2

Scope: `6127603a..2e5ee8f4` (remediation), read in the context of `origin/main..2e5ee8f4` (whole PR). Head `2e5ee8f4` confirmed current.

Verified rather than trusted. Every round-1 claim below was falsified by reverting the specific mechanism in-process and re-running the test that is supposed to guard it, and the two behavioural questions the round asked about (mid-run truncation, evicted `message_end`) were driven through the real `OperatorApp` rather than reasoned about.

---

## Round-1 findings

**M1 — a live page never finalizes completed messages. RESOLVED.**

The fix is correct and the field is set in the right places. I audited every `SubagentEntry` construction: `fold_trajectory` sets `complete=key in finished or settled`, `fold_transcript_entries` sets it on durable assistant rows and on `subagent_message`, and `parent_message`/`user` correctly do **not** carry it because they build a `UserBlock`, which has no finalize path at all — so the omission is right, not an oversight.

Both halves are independently guarded, which I confirmed by reverting each one separately:

| Mechanism reverted | Result |
| --- | --- |
| `complete` never set by the fold (pre-remediation behaviour) | `test_a_finished_message_is_committed_while_the_child_is_still_working` FAILS — *"a finished message was left uncommitted"* |
| `_supersedes` completeness guard removed, rest of M1 kept | `test_an_evicted_message_end_does_not_uncommit_a_settled_row` FAILS — *"the page forgot the row was complete"* |
| Neither (head as shipped) | both PASS |

On the two hazards asked about specifically:

- **Can a complete row still be growing?** No. `complete` comes from `message_end` for that message id, or from the job having stopped. I checked the resume path — `hub op='resume'` registers a **new** job id (`jobs_manager.register` in `harness/subagent.py`), so a settled page never un-settles under the same key, and the irreversible `_supersedes` guard cannot freeze a message that later resumes.
- **Can a complete row be missed?** The original bug is gone. One residual case exists and is *pre-existing, not introduced*: opening a long-running child **fresh** whose early `message_end` has already been evicted past the 500-event window yields `complete=False` for that row (probed directly: the old message does not survive the window at all, so it has no row to mis-fold). It degrades to the old behaviour rather than regressing, and the `settled` fallback finalizes it when the child stops.

**M2 — first-open folds at the 80-column fallback. RESOLVED, and the fix does not merely move the race.**

I reverted it by setting `SYNC_LAYOUT_WAITS = 0` and the guard test fails with exactly the defect named: `assert FALLBACK_WIDTH not in folds` → `[80, 80, 135, 135, 135]`. At head it passes 5/5 consecutive runs, so the assertion is stable, not itself racy.

The degradation paths all behave:

| Path | Result |
| --- | --- |
| 140x34 normal | `folds=[134]`, body 134, 1 block, no fallback build |
| **40x12 narrow** | `folds=[34]` — folds at the real narrow width, never at 80 |
| **Page closed during the deferral** | no fold, no block, no exception; the replay lands on an unmounted body and returns at the `is_mounted` guard |

The `not fold and entries` condition is what keeps an empty page from burning its budget, and the budget is per-`SubagentView` instance (one view per open, `app.py:9103`), so a retarget of the same view does not re-arm it — correct, since by then the body is laid out. The height-pin invariant is intact: `_apply_rows` still pins the count of whatever rows it is handed, and I measured no reflow at settle (below).

**N1 — `insert_blocks` prefix excludes `_head_block`. RESOLVED FOR THE REVIEWED CASE, but the fix is conditioned on the wrong thing — see M3.**

**N2 — `assert counter.mounts == 0`. RESOLVED.** Added, and the fixture was correctly reseeded with a first token (`text = "opening "`) so the streaming block already exists when the counter arms — otherwise a legitimate first mount would have forced the assertion to tolerate the very thing it exists to catch. Good catch by the implementer, not just a line added.

---

## New findings

**M3 — MAJOR — `subagent_view.py:1740`: `head_offset = 1 if self._head_block is not None else 0` assumes the note is at body index 0, but a child that crosses the truncation cap *while the page is open* mounts it in the middle of the list — and the remediation makes that case worse than it was before.**

`_head_block` is created lazily in `_sync_body` (`:1551`) and `append_block`s at whatever the body's length happens to be *at that moment*. On first open of an already-truncated child that is index 0, which is the case round 1 measured and the new test covers. But when the child is under the cap at open and crosses it later, the note is appended **after** the rows already mounted:

```
head_block at open:            None
head_block index after crossing = 120   (of 172 blocks)   <- the fix assumes 0
```

The `+1` is then applied against an index that needs no correction, so the history page is inserted one row too low. Measured on a 140-row durable transcript, paging in with `Home`:

```
with the offset (head): first blocks = ['durable 40', 'durable 0', 'durable 1']   <- page split, 40 lands above 0
without the offset:     first blocks = ['durable 0',  'durable 1']                <- correct
```

So for this scenario the remediation **introduces** a misordering (`durable 40` above `durable 0`) where `origin/main` was correct. The reviewed scenario is genuinely fixed and its test passes either way; this is a second, disjoint scenario the one-line condition does not distinguish.

Both conditions are still needed at once (truncation + a durable history page), which is why neither round caught it, and the underlying lazy append is pre-existing on `origin/main` — but the *misordering in this scenario* is new at `2e5ee8f4`, which is why I am calling it MAJOR rather than a follow-up.

One-line fix, verified to make both scenarios correct without disturbing the passing one: condition on the note's actual position rather than its existence —

```python
blocks = self._body.blocks()
head_offset = 1 if (self._head_block is not None and blocks and blocks[0] is self._head_block) else 0
```

| Scenario | current | proposed |
| --- | --- | --- |
| truncated at open (the new test) | note at 0 ✅ | note at 0 ✅ |
| crosses the cap while open | `durable 40` first ❌ | `durable 0` first ✅ |

(The deeper fix — mounting the note at index 0 whenever it is born — is larger and pre-existing; the guard above is sufficient and safe for this PR. Worth a follow-up ticket either way, since a note born mid-list also *reads* wrong: it claims to head a page it sits in the middle of.)

---

## Did the remediation weaken anything else?

No. Checked directly:

- **Headline streaming identity holds.** 120 streaming ticks: **1 mount, 1 remove, 1 `AssistantBlock` identity, every fold at 135** (the single remove is the tail-row swap, not a body row). At settle the block is the same object and the row count does **not** move (`8 -> 8`), so M1's fix removed the settle reflow without introducing one.
- **Tail pinning** is untouched by this diff and behaves across the terminating states.
- **The `_supersedes` guard is correctly narrow** — it refuses only the `complete → not complete` downgrade on `text`, leaving the length rule and the tool/notice arms exactly as round 1 verified them.
- **The new tests are honest.** The fold guard's `assert scrollable_content_region.width == 0` is deterministic, not racy: it is asserted *before* any pause after `_open_subagent_view`, so the body provably has no region yet — 5/5 stable, and it fails when the fix is reverted. The M1 tests both fail on their respective reverts. The N1 test asserts on block identity at index 0, which is the right observable.

---

## Gates (run on `2e5ee8f4`, this worktree)

| Gate | Result |
| --- | --- |
| `flake8 .` | ✅ clean |
| `black --check .` (26.1.0) | ✅ 569 files unchanged |
| `isort --check .` (5.13.2) | ✅ clean |
| **full-tree pyright** | ✅ **0 errors, 0 warnings** (~24s, #380 confirmed working) |
| `tests/unit/tui` | 2627 passed / 4 skipped, with one intermittent — see below |

`test_history_unavailable_and_error_retry_keep_trajectory_fallback` failed on 1 of 2 full-suite runs at head (`assert 'history unavailable' in 'read-only'`) and passes isolated and file-scoped. **Pre-existing, not attributable to this PR:** the PR's diff touches none of the history-unavailable machinery (`_history_unavailable`, `_finish_history_unavailable`, `_maybe_load_history`, `_history_state_text` — zero diff lines), and I reproduced the same race on an `origin/main` worktree at a comparable rate by driving the assertion directly (baseline held 21/40, head held 24/40). It is a `call_after_refresh` worker settling after the assertion reads the state. Worth its own ticket; not this round's business.

---

## Verdict

**Changes requested — one MAJOR (M3).**

M1 and M2 are both properly fixed at the source, guarded by tests that genuinely fail when the fix is reverted, and the headline streaming result is intact and independently reproduced. N2 is resolved cleanly. The only thing standing between this and clean is M3: N1's one-line offset keys on the note's existence rather than its position, which fixes the case round 1 found and misorders a second one that `origin/main` got right. The correction is one line plus a test for the cross-the-cap-while-open ordering.

Nothing else in the remediation weakened the change. Re-post for round 3 once M3 is addressed; I expect that round to be terminal.

damianvtran:

Implementer: manager (claude-opus-5), applying and independently verifying the round-2 correction.
Scope: `2e5ee8f4..255cee1e`
Remediation head: **255cee1e**

One finding, accepted in full.

---

**M3 — the prepend offset keyed on the note's existence rather than its position. FIXED** in `255cee1e`.

The finding is correct and I reproduced it before fixing it, on the exact scenario named. A child that is under the truncation cap when the page opens and crosses it while the reader watches mounts `_head_block` **mid-list**, not at 0:

```
before crossing the cap:  _head_block = None,  body blocks = 6
after crossing the cap:   _head_block index = 5  (of 257 blocks)
is it at 0, as the round-1 fix assumed?  False
```

So the `+1` was applied where no correction was due, pushing a prepended history page one row too low and splitting the durable window. That is the mirror image of round 1's N1, and — as the reviewer notes — a case `origin/main` had right, which is what makes it a regression the round-1 remediation introduced rather than a pre-existing gap.

Fixed with the suggested one-liner, keyed on position:

```python
mounted_blocks = self._body.blocks()
head_offset = (
    1
    if (
        self._head_block is not None
        and mounted_blocks
        and mounted_blocks[0] is self._head_block
    )
    else 0
)
```

The comment above it was also corrected: it previously asserted the note "is appended before any row", which is the false premise the whole finding rests on. It now records the lazy-append behaviour and the measured mid-list index, so the next reader is not led back into the same assumption.

**The two scenarios are disjoint, so the new one carries its own test.** `test_a_history_page_lands_in_order_when_the_cap_is_crossed_mid_read` opens under the cap, crosses it under the reader, asserts the note is genuinely *not* at index 0 (otherwise the test proves nothing), then pages in history and asserts the durable window is still monotonically ordered.

Verified the split the reviewer predicted, by reverting to the round-1 condition and re-running both guards:

```
with the existence-keyed condition:
  test_a_history_page_lands_below_the_truncation_note ......... PASSED
  test_a_history_page_lands_in_order_when_the_cap_is_crossed_mid_read
      FAILED - the prepended page landed out of order: [40, 41, ... 0, 1, ...]

with the position-keyed fix:
  both ........................................................ PASSED
```

That is the point of the finding made executable: the old condition passes round 1's guard while breaking this one, so neither test alone is sufficient.

---

### Verification

Gates on `255cee1e`: flake8 ✅ · black --check ✅ (569 files) · isort --check ✅ · **full-tree pyright ✅ 0 errors**.

Touched-area suites (`subagent_view`, `stream_anchor`, `tool_card`, `transcript_selection`) — **338 passed, 3 runs out of 3**, no intermittents.

Full `tests/unit/tui`: 2626 passed with 2 intermittent failures, in `test_app_pilot.py`, `test_aside.py` and `test_composer_seam.py` — different tests on each run, none in machinery this PR touches. This machine is currently at **load average 70** with several agents building concurrently, which is the same boot-timing sensitivity round 2 measured on an `origin/main` worktree. CI is the authority here and it was green on `2e5ee8f4`; I will confirm it green on `255cee1e` before merging.

### Outstanding before merge

Round 2 correctly flagged that the **design round was posted against the stale head `5e5b14c0`**. It is being re-confirmed against `255cee1e`; the merge gate is not satisfied until that lands clean alongside a terminal code round.

The `test_history_unavailable_and_error_retry_keep_trajectory_fallback` race round 2 identified as pre-existing (baseline 21/40 vs head 24/40) is not addressed here and should get its own ticket — it is unrelated machinery and fixing it inside this PR would muddy the diff.

damianvtran:
### Design review — round 2

Reviewer: `designer` subagent (local-operator team `lopdev`), model `anthropic/claude-opus-5`
Scope: `origin/main..255cee1e` (base `f32802e3`, head **255cee1e**, fresh at time of review)

Round 1 was posted against the stale head `5e5b14c0`. Two remediation rounds have landed since and two of them change what the user SEES (M1 finalizes finished rows on a live page; M2 defers the first fold behind a bounded layout wait), so this round was **re-run from new frames**, not re-asserted.

Method unchanged from round 1, per `AGENTS.md` "Visual validation": the real `OperatorApp` under `run_test`, page opened with `_open_subagent_view` and ticked with `_refresh_subagent_view`, exported with `save_screenshot`, rendered and viewed side by side. BEFORE frames come from my own throwaway `origin/main` worktree (`/tmp/design-base-r2` @ `f32802e3`), AFTER from the PR worktree, **same scripts, same terminal sizes**. Both capture modes again, because the flicker does not live in a settled frame:

- **Settled** (`shot.py`) — six states x two sizes x three consecutive frames.
- **Transient** (`shot_transient.py`) — exactly one pause after each refresh, i.e. the *first painted frame of the tick*.

Frames read back from the app's own SVG export rather than eyeballed as markup; the numbers below are painted rows, not widget state.

Artifacts: `/tmp/design-shots/r2/{before,after}/` (M1/M2/M3 scenarios) and `/tmp/design-shots/r2/{before1,after1}/` (round-1 scenario set + transient runs). Viewers: `/tmp/design-shots/r2/compare_r2.html`, `p_m1.html`, `p_m2.html`, `p_m3.html`, `p_states*.html`. Sizes: **140x34** and **80x24**.

---

## The three remediations, as frames

**M1 — a finished message now shows its paragraph break on a live page, and does not move when the job settles. CONFIRMED.** Scenario `m1open`: a two-paragraph message carrying its `message_end` sits above a later message that is still streaming, so the child is demonstrably still working. At 140x34 (`after/w-m1open-f1.svg`) the finished message paints as

```
 3|The duplicates come from the retry path, not the loader.
 4|
 5|When a batch times out the worker re-enqueues the whole file rather than the unacknowledged slice, so every retry replays records that
 6|already landed.
```

— the blank row at 4 is present **while the child is still working**, matching `origin/main` row for row. Body rows 1-6 are then byte-identical across `f1` (live), `f2` (next streaming tick) and `f3` (job settled): the finished block does not gain, lose, or reflow a row when the job settles. Repeated 5x at 140x34: **5/5** show the break while live and an identical block after settle. Round 1 judged the missing mid-stream break acceptable because it matched the main conversation; that judgement no longer has to be made for genuinely finished messages, which is a straight improvement.

**M2 — the bounded layout wait costs no visible frame. CONFIRMED.** Scenario `m2open` captures the first six painted frames of a fresh open, one pause apart starting from the very first. At both sizes, on both trees, **every one of the six frames is fully built at full body width**:

| | p1 | p2 | p3 | p4 | p5 | p6 |
| --- | --- | --- | --- | --- | --- | --- |
| BEFORE 140x34 body rows / max width | 28 / 135 | 28 / 135 | 28 / 135 | 28 / 135 | 28 / 135 | 28 / 135 |
| AFTER 140x34 | 28 / 135 | 28 / 135 | 28 / 135 | 28 / 135 | 28 / 135 | 28 / 135 |
| AFTER 80x24 | 18 / 75 | 18 / 75 | 18 / 75 | 18 / 75 | 18 / 75 | 18 / 75 |

Zero blank frames, zero half-built frames, no deferral the reader could perceive. The wait is bounded by refreshes rather than time, and it does not reach a painted frame.

Because the 80-column build the fix targets was measured on only 2 of 20 opens, a single capture cannot see it — so I sampled **20 opens per tree per size**, reading the prose fold width out of the *first* painted frame each time:

| | first-frame prose widths | narrow (<=80, short of settled) | empty first frames |
| --- | --- | ---: | ---: |
| BEFORE 140x34 | all 135 | 0 / 20 | 0 / 20 |
| AFTER 140x34 | all 135 | **0 / 20** | **0 / 20** |
| BEFORE 80x24 | all 75 | 0 / 20 | 0 / 20 |
| AFTER 80x24 | all 75 | **0 / 20** | **0 / 20** |

I did not reproduce the 2-in-20 narrow build on baseline in this harness, so I can confirm the after tree is clean and that the fix costs nothing visible; I am **not** claiming independent reproduction of the baseline defect the code round measured.

**M3 — the history page lands in order. CONFIRMED.** Scenario `m3home`: a child under the truncation cap at open, crossed while the reader watches, then `Home`. Durable window read off the painted page, 4 runs per tree:

| | durable rows on screen | monotonic | truncation note row |
| --- | --- | --- | --- |
| BEFORE (`f32802e3`) | `0..139` | True (4/4) | 144 |
| AFTER (`255cee1e`) | `0..139` | **True (4/4)** | **144** |

Identical to baseline, at both sizes. The rows read in order with no split window, and the note sits where it belongs. `after/w-m3home-f2.svg` shows `durable 40..54` in sequence with the footer correctly reading `transcript start`.

> Method note, because it produced a false positive worth recording: my first M3 harness pointed both trees at one transcript directory, so the second run appended a second copy of `durable 0..139` and read back as a non-monotonic window that looked exactly like the M3 misordering. It flipped with execution order, which is what exposed it. Fixture isolated per run and the wait moved onto the observable (`_wait_history` + `_wait_geometry_settled`, the repo's own helpers) rather than a fixed pause count. **No finding against the PR** — the artifact was mine.

---

## The round-1 evidence, re-confirmed on this head

The reported flicker still reproduces on baseline and is still absent here. Extended 14-tick transient run at 140x34, counting painted body rows while the message only ever grows:

| Tick | t1 | t2 | t3 | t4 | t5 | t6 | t7 | t8 | t9 | t10 | t11 | t12 | t13 | t14 |
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
| BEFORE rows | 6 | 7 | 9 | 11 | **7** | 14 | **5** | 14 | 14 | **2** | 14 | 14 | 14 | 14 |
| AFTER rows | 6 | 7 | 9 | 11 | 12 | 14 | 14 | 14 | 14 | 14 | 14 | 14 | 14 | 14 |

**Row-count collapses while the message is growing: BEFORE 3, AFTER 0.** Baseline's t10 frame is down to **two** painted body rows — `before1/w-long-t10.svg` shows the message gone entirely, leaving the header, the rule, and a `thinking 0s` row on an otherwise empty page. That is the bug, captured. The after tree grows monotonically 6 → 14 and never drops a row.

**The clock is continuous.** Same run, tail-line elapsed time: BEFORE reads `0s` on all fourteen ticks (the widget is new every tick, so its stopwatch restarts); AFTER advances `1s → 2s → 3s`. The tail row is present in every one of the fourteen after-frames at both sizes.

**Everything else is baseline-identical.** Across all 36 round-1 scenario frames (6 states x 2 sizes x 3 consecutive frames), normalising only the spinner glyph, the clock, the scrollbar cap, and the worktree name in the status bar, **34 of 36 are structurally identical to `origin/main`**. The two that differ are `settle f1` at each size — the intended mid-stream-vs-settled difference: the break is absent mid-stream, appears on the finalize tick (`f2`), and `f2` == `f3`, so the settle does not reflow twice. Gap above the pinned tail line is one blank row in every state at both sizes; tool-row rhythm, `✓` right-alignment, footer hints, contrast and ink all unchanged. Idle frames are stable: the finished-child state differs in **zero** rows across three consecutive frames at both sizes.

---

## Findings

**D1 — NIT — the horizontal scrollbar end-cap glyph flickers in and out across ticks. STILL STANDS, still improved, still a follow-up.** Re-measured on the 14-tick run at 140x34: the one-cell `▁`/`▄`/`▇` cap appears on **11 of 14** baseline ticks and **9 of 14** here, on the same ticks from t6 onward. It remains an artifact of the body's virtual height crossing the scrollable threshold as content grows, it is the vertical scrollbar's end cap rather than a stray character, and at one cell it is well below what a reader notices beside a growing message. Slightly better than baseline rather than worse; not this PR's to fix. Carried forward as a follow-up unchanged.

No BLOCKER, MAJOR, or MINOR findings. No new findings in round 2.

---

## Verdict

**Clean on the current head `255cee1e`. This round is TERMINAL.**

All three remediations were re-examined from rendered frames rather than re-asserted from round 1:

- **M1** — a genuinely finished message now shows its paragraph blank row while the child keeps working (5/5 runs) and does not grow or reflow a row when the job settles. Visually an improvement over the behaviour round 1 accepted.
- **M2** — no visible delay, no blank frame, no half-built frame on open: six consecutive opening frames fully built at full width at both sizes, and 0/20 narrow or empty first frames across 20 opens per tree per size. One artifact was not traded for another.
- **M3** — the durable window reads `0..139` in order with the truncation note at row 144, identical to baseline, 4/4 runs at both sizes.

The page reads as a steady, continuous stream at both terminal sizes: zero row collapses against baseline's three, a tail line present in every transient frame with an advancing clock against baseline's frozen `0s`, every frame folded to the full body width, and spacing, gap, and ink unchanged from `origin/main` in 34 of 36 scenario frames — the two exceptions being the intended settle behaviour. **D1 (NIT)** from round 1 still stands as the only open item and remains a follow-up, not a change request against this PR.

damianvtran:
### Agent review — round 3

Reviewer: `reviewer` subagent (independent of the implementer) — anthropic/claude-opus-5
Scope: 2e5ee8f4..255cee1e (the M3 remediation), read in the context of the whole change `f32802e3..255cee1e`

**M3 is resolved.** No blocker or major findings remain in the code. One release-mechanics item (M4) has to be handled before the merge/publish step, but it is not a defect in the diff.

---

#### Verification performed

I drove every scenario rather than reasoning about it, instrumenting `TranscriptView.insert_blocks` to capture the insert index together with the head block's body index `H` at the moment of each prepend.

| Scenario | `H` | insert index | durable rows ordered |
|---|---|---|---|
| A. Truncated at open (round 1, N1) | 0 | 1 | yes (140) |
| B. Crosses the cap mid-read (round 2, M3) | 101 | 0 | yes (140) |
| C. No head block at all | none | 0 | yes (140) |
| D. Cross mid-read, 300 durable, two successive pages | 101 → 201 | 0, 0 | yes (300) |
| E. Under cap, 300 durable, two pages | none | 0, 0 | yes (300) |
| F. Prompt row present (`prefix`=1) + head mid-list | 102 | 1 | yes (200) |
| G. Prompt row, no truncation | none | 1 | yes (200) |
| Retarget to another child | 0 → `None` | 1 | yes (120) |

Case F is the one that matters most and that neither existing test covers directly: `prefix` is 1 because the synthetic prompt row heads the entries, and the head block sits at 102. The offset correctly evaluates to 0 and the page lands in order. Retargeting via `show()` to a different `job_id` clears `_head_block` to `None` at subagent_view.py:1391 before any prepend can consult it, so there is no stale-widget window.

**Revert check (done, not taken on trust).** In a throwaway worktree at 255cee1e I replaced the new condition with the round-2 existence-keyed one:

```
1 failed, 1 passed
test_a_history_page_lands_in_order_when_the_cap_is_crossed_mid_read
E  assert [40, 0, 1, 2, 3, 4, ...] == [0, 1, 2, 3, 4, 5, ...]
   the prepended page landed out of order: [40, 0, 1, 2, ...]
```

The new test fails with exactly the reported `durable 40` above `durable 0`, and `test_a_history_page_lands_below_the_truncation_note` still passes — confirming the two guards are genuinely disjoint and that neither alone would have caught this.

**Precondition stability.** `assert view._body.blocks()[0] is not view._head_block` is structural, not racy: the note is appended at the body length already reached, which the fixture forces to ~101 by loading 140 durable rows and settling history via `_wait_history` before crossing the cap. Measured `H` was 101 on every run. Ran the test 8 times under load — 8/8 pass, 4.0s to 11.7s.

**Nothing disturbed.** `test_subagent_view.py` is 92/92 green. The streaming headline reproduces: 120 ticks give **1 mount, 0 removes**, with the assistant block holding one stable identity throughout. `blocks()` on the prepend path is not a hot path — `prepend=True` is reachable only from the history-page callback at subagent_view.py:1636, one call per Home press, never from a streaming tick, so the O(children) list copy is irrelevant there.

#### Gates

| Gate | Result |
|---|---|
| flake8 | clean |
| black 26.1.0 `--check .` | 569 files unchanged |
| isort 5.13.2 `--check .` | clean |
| pyright (full tree) | 0 errors, 0 warnings |
| `tests/unit/tui` | 2621 passed, 7 failed — all load-shed, see below |
| CI on 255cee1e | lint, type-check, test 3.12/3.13, pip-audit, cli-sanity, server-sanity all SUCCESS |

The 7 TUI failures are the known boot-timing shedding, confirmed against a baseline rather than assumed: origin/main in its own worktree shed a *different* one (`test_the_band_refreshes_even_when_the_incumbent_callback_raises`), and all 7 from the PR head pass when re-run in isolation (442 passed, and `test_boot_typing_sends_prompt` passes 3/3 on the head and on the baseline). None is in `test_subagent_view.py`.

---

#### Findings

**M4 (major, release mechanics, not a code defect) — the branch carries no version bump, and `pyproject.toml` is now behind `main`.**
The branch is based on `f32802e3`; `#376` landed on main afterwards and bumped the version, so:

```
merge-base f32802e3 : version = "0.42.15"
branch head 255cee1e: version = "0.42.15"   (branch never touched pyproject.toml)
origin/main         : version = "0.42.16"   (v0.42.16 tagged, live on PyPI)
```

A squash-merge is clean and does *not* revert main to 0.42.15 (I tested it — the merged tree reads 0.42.16), so this is not a data-loss risk. But the release step needs a version that does not already exist: publishing this as 0.42.16 is impossible and there is no bump commit to publish. Fix in one line: rebase on current origin/main and set `version = "0.42.17"`. Patch is the right bump — this is a flicker fix and a fold-width correction, not a step-function capability.

**N1 (nit) — the comment's measured figure is from a different scenario than the sentence it sits in.**
subagent_view.py:1734-1736 says "Measured on a truncated trajectory: `view._blocks` 168 against a body of 169", which is the round-1 case where the note heads the list. The paragraph immediately below then measures the mid-list case at "index 5 of 257". My instrumentation reads 101 of 270 for that second case with a 140-row fixture, so "index 5" is presumably from a smaller fixture. Both numbers are illustrative and neither is load-bearing, but a future reader diffing them against the test fixture will not reproduce either. Worth a parenthetical naming the fixture each was taken on.

**N2 (nit) — the condition is one rung narrower than the invariant it enforces.**
The exact rule is "offset by 1 when the head block sits at or above the insertion point", i.e. `H <= prefix`; the code tests `H == 0`. These diverge only for `0 < H <= prefix`. I tried hard to reach that window and could not do it through any realistic path — the head is appended after the prompt row *and* at least one live row in every natural sequence, so `H` is either 0 or >= 100 in practice. I did force `H == 1` through a synthetic three-call `show()` sequence (prompt-only body, no transcript directory, then cross the cap, then attach the transcript), and no prepend fires in that state, so nothing misorders. Not worth changing the condition; if anything, a half-line in the comment noting that the test is `H == 0` because `H` is never in `(0, prefix]` would save the next reader the same investigation.

No findings on the `fold_width` ladder, the `assistant.py` `_flat_width` change, or the `tool_card.py` rung — the comments there explain the why, the constraint, and the measurement, which is the density AGENTS.md asks for, and the ladder deliberately skips `app.console.width` for a reason it states.

---

#### Verdict

**clean** — no BLOCKER and no MAJOR defects in the code. M3 is properly resolved, and correctly so for the two scenarios that motivated it plus the five neither covered. This round is **TERMINAL**: N1 and N2 are comment polish and are follow-ups, not grounds for another round.

Ready to merge once M4 is handled — rebase on origin/main and bump to `0.42.17` in the same push. That rebase changes `pyproject.toml` only, so it does not invalidate this round's scope over the reviewed source files; the design round, however, was posted against 5e5b14c0 and still needs re-confirmation on the final head.

damianvtran:

Implementer: manager (claude-opus-5).
Scope: `255cee1e..8db2f47f`
Current head: **8db2f47f**

Round 3 was terminal on the code with one release-mechanics item. That item is fixed; the two nits are accepted as follow-ups.

---

**M4 — the branch carried no version bump and was based behind main. FIXED** in `8db2f47f`.

Confirmed exactly as reported: `#376` landed after this branch was cut and moved main to `0.42.16` (tagged and live on PyPI), while this branch still read `0.42.15` and never touched `pyproject.toml`. Rebased onto `origin/main` (picking up #376) and bumped to **`0.42.17`**.

Patch, per the finding and the versioning policy in AGENTS.md: this is a rendering and performance correction to one surface, not a step-function capability. It cannot be named in a `X.Y.0` release title, so it is a patch.

---

**Both rounds remain fresh against this head.** The rebase changed the commit SHAs, so I verified rather than asserted that the reviewed source is unchanged:

```
git diff 255cee1e..8db2f47f -- \
  local_operator/tui/widgets/{subagent_view,transcript,assistant,tool_card}.py \
  tests/unit/tui/test_subagent_view.py
(no output — byte-identical)
```

Every file either review round examined is identical to the reviewed head. The only delta this PR introduces beyond `255cee1e` is the one-line version bump; everything else in a raw `255cee1e..8db2f47f` diff is #376 arriving through the new base, not a change to this work.

The PR's own diff against its new base is unchanged in shape: 4 source files, 1 test file, plus `pyproject.toml`.

---

**N3, N4 (comment polish) — ACCEPTED, deferred as follow-ups.** Both are documentation accuracy rather than behaviour, and neither is worth another CI cycle on a branch that is otherwise ready:

- The "index 5 of 257" figure beside the offset condition was measured on a different fixture than the one the surrounding comment describes. The number is real (I measured it independently while verifying M3) but the pairing is misleading.
- The condition tests `H == 0` where the invariant it defends is `H <= prefix`. Round 3 forced `H == 1` synthetically and confirmed no prepend fires there, so it is safe as written; the comment should say why rather than leaving the stronger invariant implicit.

I am carrying these rather than fixing them now because both rounds are terminal and fresh, and a comment-only edit would invalidate that freshness for zero behavioural gain.

---

### Verification on `8db2f47f`

Gates: flake8 ✅ · black --check ✅ (572 files) · isort --check ✅ · **full-tree pyright ✅ 0 errors**.

Touched-area suites (`subagent_view`, `stream_anchor`, `tool_card`, `transcript_selection`): **338 passed, 3 runs of 3**. One run before those showed the `test_history_unavailable_and_error_retry_keep_trajectory_fallback` failure both rounds documented as pre-existing (round 2 measured baseline 21/40 vs head 24/40); it did not recur in three consecutive runs. That race still deserves its own ticket and is explicitly not addressed here.

CI was green on `255cee1e` — lint, type-check, pip-audit, and tests on 3.12 and 3.13 — and is re-running on this head; I will confirm it green before merging.

### Gate status

- Code round 3: **terminal, clean**, fresh against this head (source byte-identical).
- Design round 2: **terminal, clean**, scoped `origin/main..255cee1e`, likewise fresh — it re-confirmed M1/M2/M3 from rendered frames on that exact source.
- Both rounds were performed by agents other than the implementer.

Remaining before merge: CI green on `8db2f47f`.

PR #299, opened 2026-08-25
feat: Local Operator browser extension bridge for the browser tool

damianvtran:
## Summary

Adds the free Local Operator browser extension and loopback browser bridge so the existing `browser` tool can drive Chrome, Edge, Arc, or Brave with the user's real browser profile when cmux is unavailable.

- Adds a versioned Pydantic protocol, generated TypeScript contract, authenticated HTTP session leg, Origin-pinned WebSocket extension leg, human-confirmed pairing, atomic discovery state, daemon supervision, and `lop browser install/status/pair/logs/uninstall/serve`.
- Keeps the existing `BrowserParams` schema and write approval tier. cmux remains preferred; a `bridge:<tab>:<nonce>` handle pins extension-backed sessions to the bridge and teardown closes the owned tab.
- Adds the MV3 extension with worker reconnect/reattach, CDP screenshots and accessibility refs, trusted click/type input, text extraction, one dedicated tab, and extension-owned default-deny origin prompts enforced before main-document requests and redirect hops.
- Adds popup/options UI, derived extension icons, store listing/privacy/permission submission material, extension build/type/test CI, protocol-sync checks in both workflows, and wheel exclusion validation.
- Bumps Local Operator from `0.33.0` to `0.34.0` as a minor feature.

## Impact

Before this change, the browser tool existed only when cmux was installed. Chrome, Edge, Arc, and Brave users had no supported way to let Local Operator drive the browser they already use, so authenticated pages and screenshots were unavailable without a separate unsupported browser engine.

After this change, users install the Local Operator extension and run `lop browser install`, pair the extension with a short terminal code, and approve each new origin from browser-owned UI. Sessions keep the same browser tool contract, cookie persistence, file-write approval, PNG validation, fill read-back, actionable failure taxonomy, and one-surface lifecycle.

## Security and privacy

- Both daemon listeners bind only to `127.0.0.1`.
- Session RPC uses a 32-byte key from a 0600 atomic state file under a 0700 directory.
- Extension WebSockets require a strict `chrome-extension://<id>` Origin, pin the extension id at pairing, and store only the long-lived token's SHA-256 digest.
- The extension creates `about:blank`, attaches CDP, then gates every main-document request with `Fetch.requestPaused`. The first destination, redirects, and click navigations are default-deny before cookies or requests reach an unapproved origin.
- Extension traffic has no remote service, analytics, or cloud browser. The Chrome debugger notice remains visible while the bridge owns its tab.

## Real-path validation

[...]

PR #380, opened 2026-08-29
fix(pyright): restore default excludes so the gate skips .venv

damianvtran:
## Summary

`[tool.pyright] exclude` **replaces** pyright's built-in defaults rather than extending them. Setting it to `["docs"]` silently dropped `**/.*`, so pyright followed the `.venv` symlink that every git worktree carries into the shared uv environment and type-checked all of site-packages.

The result: the documented gate command enumerated **29466 source files instead of 566**, ran for **over 30 minutes instead of ~80 seconds**, and pushed node workers past **3 GB RSS** against a 4192 MB heap limit. Running the gate across several worktrees concurrently — the normal workflow here — thrashed the machine into swap.

CI never surfaced this: `.github/workflows/ci.yml` installs with `pip install -e ".[dev]"` into the runner's own interpreter and never creates a `.venv`, so CI stayed green while every local run of the same command degraded.

Introduced in 5cbb91e1 (#299).

## Change

```diff
-exclude = ["docs"]
+exclude = ["**/node_modules", "**/__pycache__", "**/.*", "docs"]
```

Plus a comment in `pyproject.toml` and a note in `AGENTS.md` recording the replace-not-extend behaviour, so the next edit to this key does not reintroduce it.

### File enumeration (the root cause)

`pyright --verbose . | grep "source files"`, same commit, same worktree:

| | Source files |
|---|---|
| Before | `Found 29466 source files` |
| After | `Found 566 source files` |

### Wall clock, full `.` run

| | Time |
|---|---|
| Before | **>30 min** — killed at a 1800s cap, never completed |
| After | **15.1s** on an unloaded machine; **80.6s** measured while the machine was still loaded |

```
$ time .venv/bin/python -m pyright --pythonpath .venv/bin/python .
real    0m15.151s
```

### Memory

| | Peak node RSS |
|---|---|
| Before | ~1.9 GB sampled mid-run, individual workers observed at 3.4 GB |
| After | ~600 MB |

Aggregate across concurrent worktree runs before the fix: **7.6 GB across 15 processes**, load average 30-49 on 14 cores.

### The gate still catches what it is supposed to

The speedup must not come from checking less. Added a required parameter to a widely-called method (`CredentialManager.get_credential(self, key: str, scope: str)`) with the fix applied:

```
  local_operator/web_search/providers.py:52:17 - error: Argument missing for parameter "scope" (reportCallIssue)
  tests/unit/test_credentials.py:71:18 - error: Argument missing for parameter "scope" (reportCallIssue)
  ...
19 errors, 0 warnings, 0 informations
PYRIGHT_EXIT=1
```

19 errors across both `local_operator/` and `tests/`, non-zero exit. Reverted after the check.

This also rules out the tempting-but-unsound alternative of type-checking only changed files: that same edit checked against `local_operator/credentials.py` alone reports **`0 errors` in 0.9s** while the repo is broken, because the breakage is entirely in the callers. Whole-tree checking is load-bearing; the fix keeps it and makes it fast.

### Gates

| Gate | Result |
|---|---|
| `flake8 .` | clean |
| `black==26.1.0 --check .` | 569 files unchanged |
| `pyright .` | 0 errors, 0 warnings, 0 informations |
| `pytest tests/unit` | see CI |

## Version

Patch bump to **0.42.15** — developer-facing tooling fix, no change to what the product can do.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -155,7 +155,28 @@ jobs:
           # Run CLI to create test file
           # --yolo: the sanity run is sandboxed CI; the rewritten harness denies
           # approval-gated tools (write) without a tty unless auto-approved.
-          local-operator --hosting openrouter --model deepseek/deepseek-v4-flash-0731 exec --yolo "create a file called test.txt with the text 'hello world' in it" || exit 1
+          #
+          # MODEL CHOICE: these live-LLM jobs assert that a small, cheap model
+          # can still drive the agent end to end, so the model has to emit real
+          # structured tool calls rather than describe them. The previous pick
+          # (deepseek/deepseek-v4-flash-0731) regressed into printing literal
+          # `<|DSML|>tool_calls` markup as assistant TEXT, so no tool ever ran
+          # and test.txt was never written: measured 0/5 locally against this
+          # exact prompt, and it failed three consecutive CI runs on unrelated
+          # PRs while main (which had run earlier) stayed green.
+          #
+          # qwen/qwen3.7-flash measured 10/10 on the same harness and also
+          # passes the `exec --json` streaming contract. It is cheaper on input
+          # than the model it replaces ($0.03 vs $0.045 per Mtok).
+          #
+          # If this starts failing, FIRST check whether the model is emitting
+          # tool-call markup as text -- that is a model regression, not a bug in
+          # the harness. Re-measure candidates before swapping; a model that
+          # cannot tool-call makes these jobs permanently red and teaches
+          # everyone to ignore them. Any replacement must support `tools` and
+          # be verified against both the --yolo write task and the --json
+          # streaming contract.
+          local-operator --hosting openrouter --model qwen/qwen3.7-flash exec --yolo "create a file called test.txt with the text 'hello world' in it" || exit 1
 
           # Check if file exists and contains correct text
           if [ ! -f test.txt ]; then
@@ -181,7 +202,7 @@ jobs:
         env:
           OPENROUTER_API_KEY: ${{ secrets.OPENROUTER_API_KEY }}
         run: |
-          local-operator --hosting openrouter --model deepseek/deepseek-v4-flash-0731 \
+          local-operator --hosting openrouter --model qwen/qwen3.7-flash \
             --yolo exec --json "create a file called contract.txt containing ok" \
             > run.jsonl 2> run.err || exit 1
           python scripts/check_streaming_contract.py run.jsonl || {
@@ -195,7 +216,7 @@ jobs:
         run: |
           set +e
           env -u OPENROUTER_API_KEY local-operator --hosting openrouter \
-            --model deepseek/deepseek-v4-flash-0731 exec --json "hi" > err_out.jsonl 2> err_err.txt
+            --model qwen/qwen3.7-flash exec --json "hi" > err_out.jsonl 2> err_err.txt
           set -e
           if [ -s err_out.jsonl ]; then
             # `|| [ -n "$line" ]`: bash read returns non-zero at EOF on an
@@ -262,7 +283,7 @@ jobs:
             -d '{
               "prompt": "create a file called test.txt with the text '\''hello world'\'' in it",
               "hosting": "openrouter",
-              "model": "deepseek/deepseek-v4-flash-0731",
+              "model": "qwen/qwen3.7-flash",
               "context": [],
               "options": {
                 "temperature": 0.7,
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -34,6 +34,14 @@ instead of letting isort read the repo's own config, checks something CI does
 not: both combinations pass on a file that CI then rejects. An unsorted
 function-local import reached CI exactly that way.
 
+Editing `exclude` under `[tool.pyright]` in `pyproject.toml`? It **replaces**
+pyright's built-in defaults rather than extending them, so always restate
+`"**/node_modules"`, `"**/__pycache__"` and `"**/.*"` alongside whatever you
+are adding. Dropping `**/.*` makes pyright follow the `.venv` symlink every
+worktree has and type-check all of site-packages — 29466 files and a 30-minute
+run instead of 566 files and about 80 seconds. CI never creates a `.venv`, so
+it stays green while every local run of the gate becomes unusable.
+
 The venv is uv-managed and has the package installed **editable**, so source
 edits are live. After a pull that changes dependencies:
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -39,7 +39,7 @@ pyright's built-in defaults rather than extending them, so always restate
 `"**/node_modules"`, `"**/__pycache__"` and `"**/.*"` alongside whatever you
 are adding. Dropping `**/.*` makes pyright follow the `.venv` symlink every
 worktree has and type-check all of site-packages — 29466 files and a 30-minute
-run instead of 566 files and about 80 seconds. CI never creates a `.venv`, so
+run instead of 566 files and about 15 seconds. CI never creates a `.venv`, so
 it stays green while every local run of the gate becomes unusable.
 
 The venv is uv-managed and has the package installed **editable**, so source
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "local-operator"
-version = "0.42.14"
+version = "0.42.15"
 description = "A team of proactive AI assistants that can work together behind the scenes to help you get mundane tasks done so you can focus on the fun stuff."
 readme = "README.md"
 authors = [{ name = "Damian Tran", email = "damian@gominerva.com" }]
@@ -239,6 +239,17 @@ profile = "black"
 reportMissingTypeArgument = true
 strictListInference = true
 reportPrivateImportUsage = false
+# `exclude` REPLACES pyright's built-in defaults, it does not extend them, so
+# the three defaults must be restated here alongside any addition of our own.
+# Dropping `**/.*` is not cosmetic: every worktree (and the main checkout)
+# carries a `.venv` symlink into the shared uv environment, so pyright walks it
+# and enumerates the whole site-packages tree — 29466 source files instead of
+# 566, a run that exceeds 30 minutes instead of ~80 seconds, and node workers
+# past 3 GB RSS against a 4192 MB heap limit. CI never showed this because
+# `.github/workflows/ci.yml` installs with `pip install -e ".[dev]"` into the
+# runner's own interpreter and never creates a `.venv`, so the cost lands
+# entirely on local runs of the documented gate command.
+#
 # `docs/` holds one-off tooling, not shipped package code — e.g.
 # `docs/store/assets/build_assets.py`, a Pillow script that regenerates the
 # Chrome Web Store screenshots from captured frames. Pillow ships no type stubs
@@ -247,7 +258,7 @@ reportPrivateImportUsage = false
 # runtime surface the gate exists to protect. Excluding `docs/` keeps the
 # type-check gate focused on `local_operator/` and `tests/` without weakening it
 # for real code or littering a throwaway generator with per-line ignores.
-exclude = ["docs"]
+exclude = ["**/node_modules", "**/__pycache__", "**/.*", "docs"]
 # NOTE: venvPath/venv are deliberately NOT set here, even though a bare local
 # `pyright` run resolves whatever interpreter it finds and reports spurious
 # "import could not be resolved" errors that bury the real ones.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -244,7 +244,7 @@ reportPrivateImportUsage = false
 # Dropping `**/.*` is not cosmetic: every worktree (and the main checkout)
 # carries a `.venv` symlink into the shared uv environment, so pyright walks it
 # and enumerates the whole site-packages tree — 29466 source files instead of
-# 566, a run that exceeds 30 minutes instead of ~80 seconds, and node workers
+# 566, a run that exceeds 30 minutes instead of ~15 seconds, and node workers
 # past 3 GB RSS against a 4192 MB heap limit. CI never showed this because
 # `.github/workflows/ci.yml` installs with `pip install -e ".[dev]"` into the
 # runner's own interpreter and never creates a `.venv`, so the cost lands


damianvtran:
### Agent review — round 1

Scope: 24cde3fdc634fc2796116f99d0699ad372fc865f..0e9ef51a341522de2ab8d769852fe66a96ce72ed

Independent review. I did not write this change. Diff is 2 files, +21/-2, config and docs only — no `local_operator/` or `tests/` code touched.

## No blocker or major findings.

The diagnosis, the fix, and every load-bearing number in the PR body reproduce on my own runs. Details of what I verified are below, followed by three low-severity findings.

## Verification

**1. Are the restated defaults exactly pyright's built-in defaults? — Yes, exactly, and confirmed against upstream rather than the PR body.**

`microsoft/pyright` at tag `1.1.411`, `docs/configuration.md` line 15:

> If no exclude paths are specified, Pyright automatically excludes the following: `**/node_modules`, `**/__pycache__`, `**/.*`.

Cross-checked against the actual shipped implementation in the resolved binary (`~/.cache/pyright-python/1.1.413/node_modules/pyright/dist/pyright-internal.js`), which hardcodes the literal list:

```
**/node_modules","**/__pycache__","**/.*"
```

Three entries, same order. `pyproject.toml:258` restates all three plus `docs`. Nothing missing, nothing stale, nothing invented.

**2. Is `exclude` really replace-not-extend? — Yes, proven on a fixture, not inferred.**

Built a minimal tree with a type error inside a dot-directory (`.fakevenv/pkg/mod.py`) and a clean file outside it, then varied only the config:

| Config | Files analyzed | Errors |
|---|---|---|
| `exclude = ["docs"]` (old shape) | **2** | **1** |
| `exclude = ["**/node_modules","**/__pycache__","**/.*","docs"]` (new shape) | **1** | 0 |

The old shape walks into the dot-directory. That is the bug, demonstrated directly.

**3. Are the alternatives actually worse? — Yes, and `ignore` is unsound here.**

| Alternative | Files analyzed | Verdict |
|---|---|---|
| `exclude:["docs"]` + `ignore:["**/.*"]` | **2** | **Wrong fix.** `ignore` suppresses *diagnostics*, it does not stop *enumeration* — still analyzed 2 files. It would have hidden the symptom (no errors printed) while leaving the entire cost, which is the actual problem. |
| `exclude:["docs","**/.venv"]` | 1 | Works, but narrower: fixes only `.venv` and leaves every other dot-directory enumerated, and still silently drops the two other upstream defaults. |
| Restating the defaults (this PR) | 1 | Correct — restores upstream behaviour exactly and is robust to any dot-directory, not just `.venv`. |

On `venvPath`/`venv` pinning: the pre-existing `NOTE` at `pyproject.toml:259-277` already argues against it, and that reasoning holds independently — it would couple the config to a layout CI does not have. It is also not a fix for *this* bug, since it changes import resolution, not file enumeration. **Restating the defaults is the right fix.** Agreed with the choice.

**4. Does the fix weaken the gate? — No. The 566 surface is exactly right, and the gate still fails on caller breakage.**

The analyzed set accounts for every tracked Python file that should be checked:

```
252 local_operator + 279 tests + 34 scripts + 1 extension = 566
```

`--outputjson` reports `filesAnalyzed: 566`, matching exactly. `git ls-files '*.py' '*.pyi' | grep -E '(^|/)\.'` returns **nothing** — no tracked source lives under any dot-directory, so `**/.*` drops zero legitimate files. The only tracked files newly outside the set versus a hypothetical no-exclude run are the 3 in `docs/`, which were already excluded before this PR.

Also confirmed the 29466 figure is arithmetically real, not rhetorical: the venv holds 28900 `.py`/`.pyi` files, and 28900 + 566 = **29466**, the exact number in the PR body.

I reproduced the gate-strength test myself in a scratch copy (`git archive` of HEAD, so the author's worktree was untouched) rather than trusting the PR body. Added the required `scope` parameter to `CredentialManager.get_credential`:

```
local_operator/server/dependencies.py:62:19    - error: Argument missing for parameter "scope"
local_operator/server/routes/agents.py:475:19  - error: Argument missing for parameter "scope"
local_operator/server/routes/models.py:353:27  - error: Argument missing for parameter "scope"
local_operator/server/routes/speech.py:113:23  - error: Argument missing for parameter "scope"
local_operator/session_factory.py:728:22       - error: Argument missing for parameter "scope"
local_operator/web_search/providers.py:52:17   - error: Argument missing for parameter "scope"
tests/unit/server/conftest.py:232:46           - error: Cannot assign to attribute "get_credential"
tests/unit/test_credentials.py:{48,62,71,133,165}:18 - error: Argument missing for parameter "scope"
19 errors, 0 warnings, 0 informations
PYRIGHT_EXIT=1
```

**19 errors, exit 1** — matches the PR body exactly, spanning both `local_operator/` and `tests/`. The claim is accurate.

Worth stating explicitly, since it is the subtle risk in this change: excluding `.venv` does **not** make its types unavailable. Per the same upstream doc, excluded files are still analyzed when imported by non-excluded sources, and `--pythonpath` still resolves them. The clean run is `0 errors` rather than a wall of unresolved-import errors, which is the positive evidence that dependency types are still being read. The gate got faster by not *enumerating* site-packages as project files, not by checking less of our code.

**5. CI impact — none. Claim confirmed.**

`.github/workflows/ci.yml` `type-check` installs via `pip install -e ".[dev]"` into the runner's own interpreter (lines 15-16) and runs bare `pyright`. No `.venv` is created anywhere in the workflow — `grep` for venv creation across all 5 workflows returns nothing, and `publish.yml` does not run pyright at all. CI was enumerating 566 files before and after; only local runs were affected. This also correctly explains why the regression from 5cbb91e1 (#299) went unnoticed.

**6. Version bump — PATCH is correct.**

Per `AGENTS.md` "Versioning", the bump follows user-facing materiality, not commit type. This is a developer-facing tooling fix touching only `pyproject.toml` config and `AGENTS.md`; zero runtime behaviour change and nothing a user of the published package can observe. That is squarely the patch bullet ("internal cleanups, docs"), and `main` is at `0.42.14` so `0.42.15` is the right successor. Correct call.

**7. Gates re-run by me in the worktree**

| Gate | Result |
|---|---|
| `flake8 .` | clean |
| `black==26.1.0 --check .` | 569 files unchanged |
| Wall clock | **15.159s** real (`--outputjson` self-report: 13.6s) |

The headline performance claim reproduces independently.

## Findings

**F1 — minor — `pyproject.toml:258`: restating upstream defaults creates a silent drift dependency, and CI's two pyright versions disagree.**

Hardcoding pyright's defaults means that if upstream ever *adds* a fourth default exclude, this repo silently stops honouring it with no error. That is an accepted and correct trade-off here (the alternatives are worse, per the table above), but the exposure is real and currently undocumented. It is sharpened by a pre-existing inconsistency this PR inherits rather than causes: `ci.yml:29` pins `pyright==1.1.408` for the lint job while `ci.yml:41` sets `PYRIGHT_PYTHON_FORCE_VERSION: latest` for the type-check job, and local resolves 1.1.411/1.1.413. Three different pyright versions across the gate, with a config that now encodes version-specific constants. Why it matters: the failure mode is silent and slow — exactly the class of bug this PR is fixing. One line in the existing comment noting the list mirrors upstream as of 1.1.411 and should be re-checked on a pyright major upgrade would close it. Not stop-ship; the list is correct today.

**F2 — nit — `pyproject.toml:247` and `AGENTS.md:41`: the "~80 seconds" baseline is the loaded-machine figure, presented as the normal one.**

Both the comment ("a run that exceeds 30 minutes instead of ~80 seconds") and the AGENTS.md note ("566 files and about 80 seconds") quote 80s. I measured **15.2s** on this machine, and the PR body itself reports 15.1s unloaded with 80.6s explicitly attributed to a loaded machine. The comment drops that qualifier, so a future reader benchmarking at 15s may think something changed. The number is conservative rather than inflated, so it misleads in the safe direction — but "about 15 seconds" is the honest baseline. The PR body handles this correctly; only the two committed comments lose the caveat.

**F3 — nit — `pyproject.toml:248-249`: memory figures are not independently reproducible.**

"node workers past 3 GB RSS against a 4192 MB heap limit," and the body's "7.6 GB across 15 processes, load average 30-49," describe a transient pre-fix state I cannot re-observe now that the fix is in. I verified the *post*-fix side and both file counts exactly, which corroborates the story, and the heap limit is a checkable pyright constant. Flagging only for completeness against the "flag any number you cannot reproduce" instruction — these are plausible and consistent with a 29466-file run, and I am not disputing them. No action needed beyond awareness that they are point-in-time samples.

## What I checked and found nothing wrong with

- Comment accuracy against the repo's why/constraints/context convention: both new comment blocks explain *why* the key is easy to get wrong and *what breaks*, which is the convention. No overstatement beyond F2. The pre-existing `docs/` and `venvPath` rationale is preserved intact and still reads coherently after the insertion.
- Exclude-list ordering and syntax: valid TOML, valid pyright globs, `docs` correctly retained.
- No runtime code, dependency, or packaging change — nothing that could reach the published artifact beyond the version string.
- AGENTS.md note placement: sits with the other gate guidance, correctly warns the next editor of this exact key.

## CI status at review time

`lint` pass, `type-check` pass, `pip-audit` pass. `test (3.12)` and `test (3.13)` still pending on head `0e9ef51a` — confirm both green before merging. Nothing in this diff can plausibly affect the unit suite, but the gate is the gate.

## Verdict: clean

No blocker and no major findings. This round is **TERMINAL** — F1, F2 and F3 are follow-ups, not grounds for another review round. The root cause is correctly identified, the fix is the right one among the alternatives I tested rather than merely a working one, the gate demonstrably still catches caller breakage at full strength, CI is genuinely unaffected, and PATCH is the correct bump. Merge once the two pending test jobs are green.

damianvtran:

Implementer: Local Operator (manager), claude-opus-5
Scope: 0e9ef51a..1cc483a4

| ID | Severity | Disposition |
|---|---|---|
| F1 | minor | Deferred |
| F2 | nit | **Fixed** — 1cc483a4 |
| F3 | nit | Acknowledged, no change |

**F2 — "~80 seconds" overstated. Fixed in 1cc483a4.**

Correct catch. 80.6s was measured while the machine was still loaded by the very pre-fix runs this change eliminates, so quoting it as the post-fix figure was wrong in the durable artifacts. The PR body qualified it; the `pyproject.toml` comment and the AGENTS.md note did not, and those are the ones a future reader trusts. Both now say ~15 seconds, matching the 15.152s / 15.159s we measured independently on an unloaded machine.

**F3 — memory figures are point-in-time samples. Acknowledged, no change.**

Agreed, and worth stating plainly: the 3.4 GB worker and 7.6 GB aggregate were sampled from `ps` under concurrent multi-worktree load, not from a controlled harness. They are illustrative of the failure mode, not reproducible benchmarks. They appear only in the PR body, never in the committed comments, so nothing in the tree asserts them. The load-bearing evidence is the file enumeration (29466 → 566), which is exact and reproducible.

**F1 — hardcoded upstream defaults can drift. Deferred.**

The strongest finding, and correctly not a blocker. Restating pyright's defaults means a future release adding a fourth one silently stops being honoured here, which is the same silent-and-slow failure class this PR fixes. Fixing it properly means removing the need to restate them at all — scoping the gate to `local_operator tests scripts` so `exclude` stops carrying the whole tree, or pinning one pyright version across the gate. That is a larger change than this hotfix should absorb while the thrashing is active.

Sharpened by the pre-existing version inconsistency you found: `ci.yml` pins `pyright==1.1.408` for lint but sets `PYRIGHT_PYTHON_FORCE_VERSION: latest` for type-check, while local resolves 1.1.411/1.1.413. Three versions across one gate is a latent problem independent of this PR, and now the config holds version-specific constants.

Tracked as #381 (pin one pyright version across the gate, and remove the need to restate defaults). Verified the restated list is exact for 1.1.411 against both the upstream docs and the shipped `pyright-internal.js`, so the drift risk is future-only, not present.

**Re-verified on 1cc483a4:** flake8 clean, black 569 unchanged, isort clean, pyright 0 errors / 0 warnings / 0 informations.

damianvtran:
### Agent review — round 2

Reviewer: reviewer subagent (independent of the author), anthropic/claude-opus-5
Scope: 1cc483a4..911b080e

Reviewed the new CI sanity-model commit primarily, and re-sanity-checked the whole `origin/main...911b080e` diff. **No blocker and no major findings.** Three low-severity findings below, all follow-up grade.

#### What I verified independently (not taken from the description)

| Check | Result |
|---|---|
| `flake8` | clean, exit 0 |
| `black --check` (26.1.0, pinned per AGENTS.md) | 569 files unchanged |
| `isort --check-only --profile black` | clean, exit 0 |
| `pyright` via the documented gate (`--pythonpath .venv/bin/python .`) | **0 errors, 0 warnings, 14.4s** — the round 1 claim holds |
| CI on head `911b080e` | all 14 checks green, including `cli-sanity` and `server-sanity` |
| Secret handling | key read only from `secrets.OPENROUTER_API_KEY` at ci.yml:153/203/258; the negative-path job still does `env -u OPENROUTER_API_KEY` at :218. No key-shaped literal anywhere in the diff or the comment |
| Old-model references | swapped at all 4 call sites; no missed reference that should have changed (see F5) |
| Pricing claim | confirmed live against the OpenRouter catalogue |
| Empty-enum report | **reproduced** — `AgentParams.effort` emits `enum: ["lo","med","hi",""]`, index `[3]` is the empty string, exactly matching the Gemini error |

A note on the pyright number: running bare `.venv/bin/python -m pyright` reports 854 errors from unresolved test imports. That is the interpreter-resolution trap the existing `venvPath/venv` comment already documents, not a regression from this diff — the documented command is clean.

#### Findings

**F4 — MINOR — `RecommendedOpenRouterModelIds` still recommends the model this PR just proved cannot tool-call.**
`local_operator/model/registry.py:164` still lists `deepseek/deepseek-v4-flash-0731` (and `:165` `deepseek/deepseek-v4-flash`) as a *recommended* OpenRouter model for users. This PR's own evidence is that it scores 0/5 at emitting structured tool calls on a trivial write task. Recommending a non-tool-calling model inside an agent harness is a real defect for anyone who picks it from that list.

Classified MINOR rather than MAJOR deliberately: it is **pre-existing and not introduced or worsened by this diff**, so it should not hold a CI-unbreaking PR. But the PR is what generated the evidence, so it deserves a tracked follow-up rather than silence. One-line direction: drop the two `deepseek-v4-flash` entries from `RecommendedOpenRouterModelIds`, or re-measure them before leaving them there.

**F5 — NOT A FINDING (recording it so round 3 does not re-raise it) — the remaining `deepseek-v4-flash-0731` references are correct to leave alone.**
I checked every remaining reference. `docs/VERIFICATION.md` (11 sites) and `docs/BENCHMARKS.md:52` are **dated measurement records** ("via OpenRouter, 2026-08-06, live prices") — rewriting the model id in a historical benchmark table would falsify the record. `tests/unit/model/test_registry.py`, `test_costs.py`, `test_clients.py`, `test_model_picker.py` pin a remote catalogue snapshot and cost arithmetic, unrelated to which model CI drives. No doc anywhere documents "the CI sanity model", so nothing is left stale. Nothing to change here.

**F6 — NIT — the price comparison in the comment quotes only the favourable half.**
ci.yml:172-173 says "cheaper on input than the model it replaces ($0.03 vs $0.045 per Mtok)" and stops there. Live catalogue confirms both halves: qwen `$0.03 in / $0.13 out`, deepseek `$0.045 in / $0.09 out` — **output is 44% more expensive**. The input claim is true, but a reader chasing cost will feel the omission was doing work. At this job's volume (a one-file write task, order 10^3 output tokens per run) the delta is a fraction of a cent per CI run, so the choice is right — the comment should just say so. Suggested: append "output is dearer ($0.13 vs $0.09) but these jobs emit ~1k output tokens, so the run cost is unchanged in the noise."

**F7 — NIT — `qwen/qwen3.7-flash` is a floating alias, and there is no pinned alternative.**
Worth stating explicitly since it was raised: I queried the OpenRouter catalogue (396 models) and `qwen/qwen3.7-flash` is the **only** qwen3.7-flash id published — there is no dated `-MMDD` variant to pin to, unlike `deepseek-v4-flash-0731`. So "should this be pinned?" has no available answer, and the floating-alias risk is unavoidable, not an oversight. Accepted risk, and the comment's runbook is the correct mitigation. No change requested.

#### On the design questions raised

**Is swapping the model the right fix, versus retry/robustness?** Yes, swapping is right and retry would be wrong. These jobs assert that the harness drives a real model end to end; wrapping them in a retry, or asserting the failure mode explicitly, would suppress exactly the signal they exist to produce — a harness that stops converting model output into tool calls would then pass. The residual concern is real and correctly identified: pinning any specific model does make these jobs part model-quality canary, so a third-party regression can turn CI red with no repo change. That is the inherent cost of a live-LLM gate, it is already the status quo, and the comment's triage instruction ("check whether the model is emitting tool-call markup as text — that is a model regression, not a bug in the harness") is precisely what makes the canary tractable. Good call.

**Is the comment accurate and useful?** Yes, and it is the strongest part of the commit — it records the failure mode, the measurement, the triage order, and the requirements any replacement must meet. Everything in it I could check is true. Two caveats: F6's half-quoted pricing, and the "failed three consecutive CI runs while main stayed green" line, which is an accurate account of what happened but is the kind of claim that ages into confusion once the runs are garbage-collected. Not worth changing.

**Should the pyright and CI changes have been split?** No, shipping them together is defensible. The CI model was broken independently of the pyright config, and *this PR could not have gone green without it* — splitting would have parked the pyright fix behind red CI or forced an admin merge on a red branch. Both changes are "unbreak the development gates", neither touches shipped runtime code, and the combined diff is 2 files plus a version bump. Reviewable as one unit. `0.42.14 → 0.42.15` is the correct patch bump: neither change alters what the product can do for a user.

#### Empty-enum (`google/gemini-3.5-flash-lite`): follow-up issue, do NOT block this PR

I reproduced it from the schema rather than inferring it. `local_operator/tools/agent_tool.py:117` declares `effort: Literal["lo", "med", "hi", ""] | None`, which serialises to `enum: ["lo","med","hi",""]` — `any_of[0].enum[3]` is `""`, exactly the index Google rejects. The `""` exists because the field's description says `'' clears it`, i.e. an empty string is the sentinel for unsetting a stored value.

**Recommendation: separate issue, not a blocker here.** It is pre-existing, entirely untouched by this diff, and orthogonal to it — no part of this PR makes it better or worse, and this PR's model does not trip it. Blocking on it would hold a CI-unbreaking fix behind an unrelated schema change that deserves its own tests.

It is genuinely worth fixing soon, though, and worth filing at MAJOR: any user on the native Google provider gets a hard 400 on **every** request carrying the agent tool, which is a whole-provider outage for them rather than a degraded path. The fix wants a real decision rather than deleting a character, so it needs its own PR: either drop `""` from the `Literal` and use an explicit sentinel for clearing (cleanest, but it is a tool-surface change), or strip empty-string enum members when translating schemas for the Gemini path (narrower, keeps the clearing behaviour). Prefer whichever the provider layer already does for similar quirks. A regression test should assert no enum member is empty across every tool schema, so the next `""` sentinel cannot reintroduce it.

#### Verdict

**clean — this round is TERMINAL.** No blocker, no major. F4 is a follow-up, F6 and F7 are nits, F5 is explicitly not a defect. None require a round 3: remaining items are follow-ups, not remediation. Gates verified locally and CI is green on the current head `911b080e`. Ship it.

damianvtran:

Implementer: Local Operator (manager), claude-opus-5
Scope: 1cc483a4..911b080e

| ID | Severity | Disposition |
|---|---|---|
| F4 | minor | Deferred → #383 |
| F5 | not-a-defect | Acknowledged |
| F6 | nit | Accepted, no code change |
| F7 | nit | Accepted, no code change |
| (empty enum) | major | Deferred → #382 |

**F4 — `RecommendedOpenRouterModelIds` still recommends the model that cannot tool-call. Deferred to #383.**

The strongest finding of the round, and correct: `model/registry.py:164` recommends `deepseek/deepseek-v4-flash-0731`, which this PR measured at 0/5 on the most basic agentic task we have. Recommending it in a tool-calling harness means a new user's default choice is an agent that narrates actions it never performs.

Deferred rather than fixed here only because it is a user-facing behaviour change to a different subsystem, and this PR is already carrying two changes. Filed with the full measurement table as evidence, including the suggestion that the recommendation list be backed by a periodic tool-calling smoke check rather than a hand-maintained constant.

**Empty enum (`effort` → `enum[3] == ""`). Deferred to #382, filed at MAJOR per your recommendation.**

Agreed on severity, and thank you for reproducing it rather than taking my report at face value — `agent_tool.py:117` `Literal["lo","med","hi",""]` is the exact source. Any user on the native Google provider gets a hard 400 on every request carrying the agent tool, which is a whole-provider outage rather than a degraded path. Filed with both candidate fixes (drop the sentinel vs strip empty members in the Gemini translation layer) and a required regression test asserting no tool schema exposes an empty enum member.

Not fixed inline because the `""` is a documented sentinel for clearing the field, so removing it needs a migration decision on the agent-profile update path — not something to smuggle into a CI fix.

**F5 — floating alias with no dated variant. Acknowledged, no action available.**

You checked whether a dated `qwen3.7-flash` variant exists to pin to and found none published. I had assumed rather than verified that; the verification is the useful part. The triage comment in the workflow is the mitigation, and it now tells the next person to check for tool-call markup in text before suspecting the harness.

**F6 — comment quotes only the favourable half of the pricing. Accepted.**

Fair. Input is cheaper ($0.03 vs $0.045) and output is 44% dearer ($0.13 vs $0.09), and the comment cites only the first. At roughly 1k output tokens per sanity run the difference is fractions of a cent, which is why it did not change the decision — but quoting one side is still selective. Left as-is rather than churning the branch for a comment nit now that CI is green; noted here so the record is complete.

**F7 — accepted, same disposition.**

**Verification on head `911b080e`:** all 14 checks green, including both live-LLM sanity jobs — `cli-sanity` pass, `server-sanity` pass. These are the jobs that failed three consecutive times before this commit, which is the direct evidence the swap fixed the real cause.

PR #384, opened 2026-08-29
fix(subagents): give read-only roles network and inherited MCP reach

damianvtran:
# PR Description

## Summary of Changes

A `scout` subagent launched for a research task reported "I have no network
access in this session (no bash, no browser, no fetch tool)" and fell back to
grepping the local disk for facts it had been asked to find on the web. It
burned 2m33s and produced nothing.

`web_search` and `web_fetch` are in `DEFAULT_TOOL_NAMES` but appeared in **no**
read-only role allowlist, so the read-only surface was local lookups only and a
role whose entire purpose is research was structurally incapable of it. Both
tools are `approval_tier="read"`: they retrieve a remote document and produce no
local side effect beyond a bounded cache under `config_dir()`, which is the same
promise `read` makes about the disk. **Read-only is a claim about CHANGE, not
about reach.**

### Diagnosis (verified against `origin/main`)

1. **Three specifications that must agree, all missing the network tools.**
   `READ_ONLY_TOOLS` (`agent_profiles.py`), `SCOUT_TOOL_ALLOWLIST`
   (`harness/subagent.py`, the no-profile fallback), and the `tools:` frontmatter
   of the `scout`/`reviewer`/`manager`/`architect` seeds. The first two were
   independent hand-maintained copies of the same list, which is how they came to
   disagree.

2. **MCP denied wholesale to every restricted role.** `_build_child_session`
   computed `mcp_allowed = not ((profile is not None and bool(profile.tools)) or
   agent == "scout")`, so a reviewer or scout got no MCP surface at all and
   `resolve_internal_url` rejected `mcp://` outright. An MCP server is frequently
   the only route to the ticket, design doc or log the research is about.

3. **Fixing the seeds alone would not have reached this machine.**
   `resolve_profile` is **registry-first**, and installing a role freezes its
   allowlist into a `tools:a,b,c` tag. The operator's own `reviewer`, `architect`
   and `manager` are installed registry rows whose tags predate the network
   tools, so editing the packaged seeds repairs none of them.

### Fixes

- **A — One source of truth for the read-only surface**
  (`agent_profiles.py`, `harness/subagent.py`). New
  `READ_ONLY_NETWORK_TOOLS = ("web_search", "web_fetch")`; `READ_ONLY_TOOLS` is
  built from it, and `SCOUT_TOOL_ALLOWLIST` is now **derived**
  (`frozenset(READ_ONLY_TOOLS)`) rather than a second literal. It keeps its role
  as the no-profile safety fallback, so the read-only promise still does not
  depend on a seed file being present. Seed frontmatter updated to match.

- **B — Restricted roles inherit MCP but cannot widen it**
  (`mcp/resources.py`, `harness/subagent.py`). `make_mcp_resolver` grows
  `deny_activation_reason`, splitting DISCOVERY from ACTIVATION. A restricted
  child receives the tools the parent had **already enabled** (the parent is
  unrestricted, chose them for the very task it is delegating, and stays
  accountable), `read mcp://<server>` still lists, and a tool URL renders the
  reason instead of a schema. A child can never widen its surface past its
  delegator. The split lives in the URL-parsing module so no caller re-derives
  "is this a tool URL?".

  A server's tools are minted `approval_tier="exec"` because their side effects
  are unknowable from here, so the harness cannot tell a read tool from a write
  tool by inspection. Inherit-but-never-activate is the line the code **can**
  enforce honestly.

- **C — A capability floor for stale registry rows**
  (`_with_network_floor`, `harness/subagent.py`). Only
  `READ_ONLY_NETWORK_TOOLS` is re-admitted, at child construction, from the
  session's own inventory. Deliberately **not** by rewriting the operator's
  registry rows: a profile is user data, silently editing it would erase a
  deliberate edit, and a floor computed per launch is correct on the next launch
  rather than after a migration nobody remembers to run. A session with web
  search configured off contributes nothing, so the floor cannot mint a tool the
  operator disabled.

- **D — Docs corrected** (`guides/agents/GUIDE.md`, `harness/types.py`,
  `SCOUT_PREAMBLE`, module docstring). Several said "no side effects at all" /
  "lookups only", which no longer honestly describes the surface.

**No write or execution denial changes.** A reviewer still has no
`edit`/`write` and still cannot modify the diff it reviews — the boundary
`agent_profiles.py` and `GUIDE.md` call load-bearing.

## Related Issue(s)

- Operator-reported scout research failure (no ticket).

## Impact

- No breaking changes, no dependency updates, no schema/tool-footprint change:
  `web_search`/`web_fetch` were already built for every session; this only stops
  filtering them out of restricted children.
- Security: restricted roles gain **read** reach only. Write/exec denials are
  untouched, and the new MCP path is strictly narrower than the unrestricted one
  (inherit, never activate).
- One deliberate trade-off, recorded in
  `test_the_operators_own_role_overrides_the_packaged_one`: the floor applies to
  **any** allowlist, because a persisted tag list cannot distinguish "omits
  `web_search` deliberately" from "was written before it existed". An operator
  who narrows a role to `read` on purpose still gets retrieval back. That is a
  read with no local side effect.

### End-to-end: a real scout doing a real web search

The point of the change, exercised through the built runtime from this worktree
(`.venv/bin/python -m local_operator.cli exec --yolo --json`), launching a real
`scout` child via `task(agent="scout")`.

**Before (`origin/main`, same command, same prompt)** — the child's verbatim
return:

> I cannot answer this: **there is no `web_search` tool in this session**, and I
> have no other way to reach the network.
> - My available tools are exactly: `read`, `glob`, `grep`, `list_variables`,
>   `read_variable`, `hub`. No `web_search`, no `bash`, no `browser`.
> - No local fallback exists either [...] `glob **/textual*` matched nothing.
> I will not guess a version number.

**After (this branch)** — the child's own transcript
(`~/.local-operator/sessions/a402e7372ecd`), verbatim tool calls and real
responses:

```
>>> CALL: web_search {"i": "Finding Textual latest version",
                      "query": "Textual PyPI latest version release"}
>>> RESULT web_search is_error=None
Provider: duckduckgo (credential-free)
Sources:
1. textual · PyPI
   https://pypi.org/project/textual/
2. Releases · Textualize/textual - GitHub
   https://github.com/Textualize/textual/releases

>>> CALL: web_fetch {"i": "Fetching PyPI Textual metadata",
                     "url": "https://pypi.org/pypi/textual/json"}
>>> RESULT web_fetch is_error=None
[200] https://pypi.org/pypi/textual/json
json · application/json · cache miss
{ "info": { "author": "Will McGugan", ...
```

Child's answer: **Textual's current stable version on PyPI is 8.2.8**, cited to
`info.version` in the PyPI JSON API, with the release artifacts
(`textual-8.2.8-py3-none-any.whl`, uploaded `2026-06-30T06:51:26Z`) as
corroboration. A live `cache miss` fetch, not a stored copy.

### The floor, against the operator's actual registry rows

A probe that resolves each role exactly as `task` does (`resolve_profile`,
registry-first) and builds the child through `_build_child_session`, printing the
persisted allowlist beside the inventory the child ends up with:

| role | source | persisted allowlist | floored in | network reachable | mutators gained |
|---|---|---|---|---|---|
| scout | packaged seed | `...read_variable, web_search, web_fetch` | nothing (already named) | **True** | none |
| reviewer | **registry row** | `...read_variable, bash, todo` | `web_fetch, web_search` | **True** | none |
| architect | **registry row** | `...todo, write, web_search` | `web_fetch` | **True** | none |
| manager | **registry row** | `...read_variable, bash, todo` | `web_fetch, web_search` | **True** | none |

On `origin/main` the same probe reports `network reachable: False` for all four.
`mutators gained: none` everywhere is the boundary holding. (`manager` shows
`task` on both branches — pre-existing `delegate: yes` behaviour, not this
change.)

### Unit tests

New regressions in `tests/unit/test_agent_profiles.py`:
- `test_a_read_only_role_changes_nothing_but_can_still_reach_the_web`
- `test_every_allowlisted_seed_carries_the_read_only_network_tools` — pins the
  three specifications in agreement.
- `test_the_scout_fallback_allowlist_matches_the_read_only_surface` — pins the
  fallback as derived, not a second copy.

New in `tests/unit/session/test_launch_subagent.py`:
- `test_a_stale_role_allowlist_still_gets_the_network_floor` — the registry-row
  case, with the pre-fix tag list spelled out literally so it cannot track
  today's constant; asserts `edit`/`write`/`bash`/`eval`/`browser`/`task` all
  still denied.
- `test_the_network_floor_stays_empty_when_web_tools_are_disabled`
- `test_restricted_child_may_discover_mcp_but_never_activate_it`
- `test_restricted_child_inherits_the_mcp_tools_the_parent_enabled`

Two existing tests were reconciled to the new surface (separate commit
`1b27b216`), not deleted: the exact-inventory assertion now pins what the
operator's allowlist must still win, and scout's badge count is derived from the
seed instead of hard-coded `5 tools`.

### Gates (whole tree, exactly as CI runs them)

```
.venv/bin/python -m flake8 .                                    clean
uvx --from black==26.1.0 black --check .                        569 files unchanged
uvx isort==5.13.2 --check .                                     clean
.venv/bin/python -m pyright --pythonpath .venv/bin/python .     0 errors, 0 warnings (14.9s)
.venv/bin/python -m pytest tests/unit -q                        7397 passed, 7 skipped
```

Note on the earlier failures in this branch's history: an initial run showed 28,
then 5, then 3 failures, all in event-loop timing assertions. They were measured
while six concurrent `pyright` processes from other agents held the machine at
load average 79-135. Re-measured serially on an idle box (load ~5), the six
timing-sensitive tests pass on **both** this branch and `/tmp/lop-base` at the
same rebase base `f32802e3`:

| | branch | baseline `f32802e3` |
|---|---|---|
| the 6 timing tests | **6 passed** in 10.20s | **6 passed** in 9.96s |

`test_the_loop_stays_responsive_while_several_subagents_run` — the one a change
to child tool inventory could plausibly regress — was run three more times on the
branch alone: 2.26s / 2.30s / 2.21s, all passing. The full suite is green with
zero failures.

- A `scout` (or any allowlisted role) can now search and fetch the web. A
  research subagent that used to report having no network access does the
  research.
- A restricted role can read `mcp://` and use the MCP tools its parent already
  enabled; asking it to enable a new one returns a stated reason instead of
  silence.
- Roles installed before this release regain network reach without the operator
  reinstalling or editing them.

## Checklist


damianvtran:
### Agent review — round 1
Scope: origin/main..52773ae04a9ead750883328a32c1c0f74556fd85

**Blocker/major findings exist: one MAJOR (R1).** The two headline claims of this PR — the network floor and "a restricted child can never widen its surface" — hold at depth 1 and I verified both by execution. The write/execute denial is airtight everywhere I could attack it. The MAJOR is a composition gap at depth 2 that the diff introduces by widening MCP to restricted roles without closing the grandchild path.

Gates run in `/tmp/lop-subagent-tools`: flake8 clean, `black --check` clean (531 files), `isort --check-only --profile black` clean, `pyright --pythonpath .venv/bin/python .` **0 errors**, and the four touched test files **135 passed**.

---

#### R1 — MAJOR — a grandchild of a restricted delegating role activates MCP tools its parent was refused
`local_operator/harness/subagent.py:1079`, with `:1334-1337`

`restricted` is computed from the child's OWN profile only. A `manager` (the one packaged seed with `delegate: yes`, and it now carries `web_search, web_fetch` so it is unambiguously `restricted`) gets `deny_activation_reason` and correctly cannot activate. But it keeps `task`, and at `:1337` it is handed `child.mcp_manager = parent_session.mcp_manager`. When it delegates, `_build_child_session` runs again with `parent_session=<the manager>` and `profile=None` — so `restricted` is `False`, `_child_mcp_wiring` is built with **no** deny reason, and the grandchild activates freely into the shared live manager.

Reproduced against the real packaged seed (`load_seed("manager")`), not a synthetic profile:

```
packaged manager seed tools: ('read','glob','grep','list_variables','read_variable','bash','todo','web_search','web_fetch') | may_delegate: True
manager denied activation: True
GRANDCHILD activated delete_issue: True
grandchild mcp inventory: ['mcp__linear_delete_issue', 'mcp__linear_list_issues']
```

The parent had enabled only `list_issues`. The grandchild obtained `delete_issue` — a tool minted `approval_tier="exec"` (`mcp/tool_bridge.py:340`), i.e. exactly the unknowable-side-effect class the restriction exists for, and precisely a *write*. This is not a pre-existing bug being inherited: on `origin/main` a restricted role got `mcp = None` and `mcp_manager` was never set on it, so no grandchild could exist with an MCP surface at all. The diff creates the path.

Why it matters beyond the escalation itself: the manager's *own* denial message tells it to "report to your parent (`hub`) that this tool needs enabling on its side" — but a manager that can spawn a child has a strictly easier route, and it is the role most likely to take it. The escape is reachable by the exact role most likely to find it.

It also falsifies two comments this codebase's why/constraint standard would flag, both introduced here: `subagent.py:947` "a restricted role can never widen its own surface past its delegator's" and `GUIDE.md:62` "a restricted child can never widen its surface past the parent that delegated to it." As written, both are false one hop down.

Fix (one line, and it composes at any depth): make restriction inherit, e.g. propagate the parent's restricted state — `restricted = (profile is not None and bool(profile.tools)) or agent == "scout" or getattr(parent_session, "_mcp_activation_denied", False)` — setting that marker on the child whenever `restricted`. Denial must be sticky downward: a delegator cannot grant what it does not itself hold.

Note the write *inventory* does not leak — the grandchild of a restricted role gets `edit`/`write`/`bash` from its own unrestricted profile, which is existing intended behaviour for `task` children and not part of this finding. The finding is specifically the MCP activation capability crossing a boundary its holder was denied.

---

#### R2 — MINOR — the floor silently overrides a deliberately network-denied custom role
`local_operator/harness/subagent.py:177`

Verified: a custom role `AgentProfile(name="airgap", tools=("read","grep"))` comes back as `['grep','hub','read','web_fetch','web_search']`. The floor cannot distinguish "written before the tools existed" from "narrowed on purpose", so the author of a deliberately offline role gets retrieval back with no signal.

The PR is honest about this — the trade-off is stated in the docstring and asserted in `test_role_subagents.py` rather than hidden, which is the right way to land it, and the blast radius is genuinely small (read-tier retrieval behind the `validate_public_url` SSRF gate at `web_fetch/service.py:269`, and a session with the tools configured off contributes nothing — I confirmed the disabled-config case yields no network tools). So I am not rating this higher. But it *is* a real behaviour change to user data semantics, and `agent show` will keep printing `tools: read, grep` while the role actually runs with four. Consider an opt-out (a `network: no` frontmatter key) if anyone ever reports it; not needed to ship.

#### R3 — MINOR — duplicate tool name when an MCP server exposes a tool named `web_search`
`local_operator/harness/subagent.py:177`

I checked the name-collision attack the priorities called out. Real MCP tools are minted `mcp__<server>_<tool>` (`tool_bridge.py:95`), so `origin()` is keyed on metadata and a genuine collision **cannot** occur through the normal path — the floor cannot admit an MCP tool. Good. Forcing the pathological case by hand (renaming a minted tool to bare `web_search`) produced `['hub','read','web_fetch','web_search','web_search']` — a duplicate in the inventory, not an escalation. Unreachable in practice; noting only because `_with_network_floor` matches on bare name and dedupes only against `allowed`, not against itself. One line: build `missing` from a dict keyed on name.

#### R4 — NIT — `_with_network_floor` returns `allowed + missing`, appending after `hub`-insertion ordering
`local_operator/harness/subagent.py:98`

The module docstring elsewhere notes the tool list order matters for prompt caching. The floor appends network tools at the tail rather than at their `create_tools` position, so a role that *does* name `web_search` orders it differently from one that gets it floored. Harmless (order is stable per role, so the cache still keys consistently), but if order-stability is load-bearing, filtering `available` positionally would be tidier.

---

#### What I checked and found sound

- **Write/execute denial is intact.** Reviewer inventory: `['bash','glob','grep','hub','jobs','list_variables','read','read_variable','todo','web_fetch','web_search']` — `{"edit","write"} & set(names) == set()`. A reviewer still cannot modify the diff it reviews. The stale-row scout keeps `{edit,write,bash,eval,browser,task}` disjoint. Nothing in `_with_network_floor` can re-admit past `READ_ONLY_NETWORK_TOOLS`: it filters `available` by membership in that tuple, so the widest possible outcome is those two names.
- **`resolve_internal_url` fall-through was tightened, correctly.** Dropping `not mcp_allowed` from the `mcp://` guard (`:1093`) is the right call — the old condition let a *non*-restricted child fall through to the parent's resolver and activate into the wrong session. Now no `mcp://` ever falls through, and `guide://`/`skill://` still do.
- **The parent's inventory is never mutated.** Confirmed by execution on both the restricted and unrestricted paths: after a child activates, the parent's MCP tools are unchanged (`['mcp__linear_list_issues']`). `enabled` is a per-wiring closure set and `refresh_tools` targets `child`.
- **`activate` left unguarded is defensible.** The comment at `:960` claims it is unreachable for a restricted child, and that is true — `make_mcp_resolver` returns before calling it (`mcp/resources.py:464`). One activation path beats two drifting allow-checks. The `deny_activation_reason` check is correctly placed *after* `_find_tool`, so an unknown tool still gets the "unknown tool" message rather than a misleading denial.
- **Discovery/activation split is at the right layer** — in the module that parses `mcp://`, so no caller re-derives "is this a tool URL?".
- **Docs and comments swept properly.** I grepped for the stale "no side effects at all" / "lookups only" phrasing: `harness/types.py:471`, `SCOUT_PREAMBLE`, the module docstring, `agent_profiles.py:102`, `scout.md` description and `GUIDE.md` are all updated. `agent_profiles.py:102` now reads "no LOCAL side effects", which is accurate. No survivors found. The `test_agent_tool.py` change from a hard-coded `5 tools` to `len(scout.tools)` is a genuine improvement.
- **Versioning is right.** `0.42.15 → 0.42.16` — patch. This restores a capability that was meant to exist; it is not a step-function new surface, so patch is correct per policy.

---

**Verdict: changes requested.** R1 is a MAJOR: the diff opens an MCP activation escalation at depth 2 that did not exist on `origin/main`, reachable by the packaged `manager` role, and it makes two newly-written claims in `subagent.py` and `GUIDE.md` false. It is a small, well-scoped fix (inherit the restricted flag; assert it with a grandchild test alongside the existing depth-1 one). R2–R4 are follow-ups and not a reason to hold. Everything else in this change is sound, well-reasoned and unusually well-commented, and the read-only-means-change-not-reach framing is the correct call. Round is not terminal — R1 needs a remediation and a re-review against the new head.

damianvtran:

**R1 — MAJOR — grandchild of a restricted delegating role activates MCP tools its ancestor was refused — FIXED in `fede656f`.**

Correcting the record first: my first attempt, `506a4199`, claimed in its commit message that the child was stamped `_mcp_activation_denied` and that a child of a denied session inherits the restriction. **Its diff did not contain that mechanism.** I had made the edit, then ran a deliberate "does this test fail without the fix?" check by stripping the line, and restored the file with `git checkout -- local_operator/harness/subagent.py` — which reverted to HEAD, where the fix was not yet committed, silently destroying it. I did not re-run the grandchild test against the committed tree. On `506a4199` the escalation was still live, `grep -rn "_mcp_activation_denied"` returned nothing, and the new test **failed**. I reported a pass I had observed on a dirty tree as if it applied to the pushed artifact. That was wrong and I am flagging it rather than presenting R1 as fixed first time.

`fede656f` is the real fix:
- `MCP_DENIED_ATTR` (`subagent.py:911`) — a named constant, because the reader and writer sit ~250 lines apart and a typo in either reopens the hole with no exception to notice.
- `subagent.py:1389` — `setattr(child, MCP_DENIED_ATTR, restricted)`, stamped on every child **unconditionally**, deliberately outside the `mcp is not None` branch: a child built with no MCP surface can still delegate, and its own child resolves the manager off the session at that later point.
- `subagent.py:1127` — `restricted` ORs in that marker read off `parent_session`, so a descendant of a denied ancestor is denied at any depth.

Verified against the **committed** artifact, not the working tree — a clean detached worktree at `fede656f`, `git status --porcelain` empty:

```
$ git worktree add --detach /tmp/lop-verify HEAD   # fede656f
$ grep -c "MCP_DENIED_ATTR" local_operator/harness/subagent.py
3
$ pytest "tests/unit/session/test_launch_subagent.py::test_a_grandchild_cannot_activate_what_its_restricted_ancestor_was_refused" -q
1 passed in 12.22s
```

The reproduction the reviewer ran, re-run on this commit against the real `load_seed("manager")`:

```
packaged manager seed tools: (..., 'bash', 'todo', 'web_search', 'web_fetch') | may_delegate: True
manager denied activation: True
manager can delegate (has task): True
GRANDCHILD activated delete_issue: False          # was True
grandchild mcp inventory: ['mcp__linear_list_issues']   # was [...delete_issue, ...list_issues]
```

Two regression tests, both in `tests/unit/session/test_launch_subagent.py`:
- `test_a_grandchild_cannot_activate_what_its_restricted_ancestor_was_refused` — depth 2, driven off the **real packaged seed** (a hand-written profile could drift from what the seed grants and stop covering the regression); asserts the seed still has an allowlist and `may_delegate`, so it cannot silently degrade into testing nothing. Confirmed it fails without the fix and passes with it.
- `test_an_unrestricted_lineage_still_activates_at_depth_two` — the counter-check, so the escape is not closed by breaking ordinary delegation.

The two claims this branch wrote and had not earned — `_child_mcp_wiring`'s "can never widen its own surface past its delegator's" and `GUIDE.md:62` — now state the inheritance that makes them true.

**R2 — MINOR — floor silently overrides a deliberately network-denied role — BEHAVIOUR KEPT, comment corrected in `506a4199`.** The docstring said admitting these tools "changes nothing an allowlist was drawn to prevent", which understated it. It now names the trade-off plainly: a persisted tag list cannot distinguish "omits `web_search` because it predates the tool" from "omits it on purpose", so a deliberately offline role gets retrieval back and `agent show` keeps printing the narrower stored list. Kept because the alternative leaves every role installed before this release permanently unable to do the research it exists for, and what is re-admitted is read-tier retrieval behind the `web_fetch` SSRF gate. The docstring points at an explicit `network: no` frontmatter key as the right fix if anyone needs the offline case — inferring intent from an omission is what cannot be done.

**R3 — MINOR — duplicate tool name if an inventory holds two `web_search` — FIXED in `506a4199`.** `_with_network_floor` now builds the floored set in a name-keyed dict (`setdefault`), so it dedupes against itself as well as against the allowlist. Agreed it is unreachable through the normal path, since MCP tools are minted `mcp__<server>_<tool>` and can never match these two names; cheap to make impossible.

**R4 — NIT — floor appends at the tail rather than at `create_tools` position — REJECTED, with evidence.** The concern was prompt-cache stability. Order is deterministic *per role*, which is what the cache prefix requires — three consecutive builds of the same profile produced byte-identical orderings:

```
identical across builds: True
order: ['bash','read','glob','grep','todo','web_search','web_fetch','hub','jobs']
```

Tail-appending is also the established convention immediately above (`hub` is appended the same way at `:1190`), so positional filtering here would introduce a second way of doing one thing. No change.

---

**Correcting the full-suite evidence.** The "7397 passed" figure in the PR description was taken at `52773ae0`, before this remediation, and I should have said so. Current evidence for head `fede656f` is CI, whose runners are isolated (the local box has been at load 90-160 from other agents all evening, so local suite timings there are not trustworthy):

- **14/14 checks pass** on `fede656f`. `test (3.12)` and `test (3.13)`: **7397 passed, 9 skipped** (11m21s / 11m41s). `lint`, `type-check`, `pip-audit` green.
- Local `pyright --pythonpath .venv/bin/python .`: **0 errors, 0 warnings**. flake8 / `black --check` (569 files) / `isort --check` clean.

damianvtran:
### Agent review — round 2
Scope: origin/main..fede656f3da947610c30e8e4ac8e625a1dca7ce9

**One MAJOR remains (R5), newly found on the resume path. R1 is genuinely fixed for the launch path; R3 is fixed; R2's docstring now states the trade-off honestly; R4's rejection is accepted.**

Everything below was run against a **clean detached worktree at `fede656f`** (`git worktree add --detach /tmp/lop-r2 fede656f`, `git status --porcelain` empty), not the working tree, and not on trust from the remediation comment.

---

#### R1 — MAJOR — VERIFIED FIXED (launch path), and the test has teeth

The mechanism is really in the committed artifact this time. `grep -rn "_mcp_activation_denied\|MCP_DENIED_ATTR"` on the detached checkout returns exactly three sites: the constant at `subagent.py:911`, the read at `:1127`, the write at `:1389`. No other file references it.

Depth-2 reproduction of the original escalation, against the real `load_seed("manager")` — and I extended it past what the remediation claimed, to **depth 4**:

```
d1 marker: True   d1 denied : True
d2 marker: True   d2 denied : True
d3 marker: True   d3 denied : True
d4 marker: True   d4 denied : True
d4 inherits list_issues: True
```

So restriction propagates at depth 3+, not only depth 2, and the lineage still inherits what it legitimately held.

**The test has teeth.** I copied the checkout, reverted *only* the third term of the `restricted` computation (leaving the constant and the `setattr` in place, so this isolates the inheritance and not the whole commit), and re-ran:

```
FAILED test_a_grandchild_cannot_activate_what_its_restricted_ancestor_was_refused
E  AssertionError: assert 'cannot enable new ones' in
   '# Enabled MCP tool: mcp__linear_delete_issue ...'
1 failed, 1 passed in 2.50s
```

That is the exact R1 escalation reappearing, and the counter-check (`test_an_unrestricted_lineage_still_activates_at_depth_two`) still passes on the reverted tree — so the pair is not a test that passes either way, and the fix is not closing the hole by breaking ordinary delegation. Full file on `fede656f`: **51 passed**.

**The unconditional stamp does close the path the comment claims.** I built a restricted `manager` under a root with **no** `mcp_manager` at all, wired MCP onto that child afterwards, then had it delegate:

```
root has mcp_manager: False
restricted child built with NO mcp; marker: True
grandchild marker: True
grandchild ESCALATED: False | denied: True
```

Had the `setattr` sat inside the `if mcp is not None:` branch, the marker would have been absent and the grandchild unrestricted. The comment at `:1381-1388` is accurate.

**Other routes I attacked and found closed:** the parent activating a new tool *after* the restricted child exists does not widen that child retroactively (`['mcp__linear_list_issues']` before and after); a manual `refresh_tools` on the restricted child does not let its grandchild activate (`marker: True | can activate: False`); `_build_child_session` is the **only** construction path (`grep -rn "_build_child_session"` outside tests returns one call site, `subagent.py:444`); the `attach` closure targets `child` only and the parent's inventory is unchanged.

---

#### R5 — MAJOR — NEW — `hub op='resume'` rebuilds a restricted lineage's child against the ROOT session, and the denial does not survive
`local_operator/harness/comms.py:1524` (with `:1501`), against `subagent.py:1127`

This is the failure mode the round-2 brief asked me to look for, and it is real. **The marker is per-Session in-memory state that nothing re-derives on resume.**

`SubagentComms.resume()` relaunches with `parent_session=self._session` — the **comms-owning session**, which is the root, not the child's actual parent. Children are handed the *parent's* comms instance (`subagent.py:1191`, `:1302`), so the registry is flat and a grandchild of a restricted `manager` is a first-class row in the **root's** registry. `_ChildRecord` persists `agent_role` and `effort` (`comms.py:1501`) but has **no field recording that the child was built under a denial** — so `restricted` at `subagent.py:1127` reads the marker off the *root*, which is unrestricted, and the rebuilt child comes back clean.

Reproduced end-to-end on the real production path — real `load_seed("manager")`, child launched through `run_subagent`, allowed to settle, then resumed through `comms.resume()` (what `hub op='resume'` calls at `builtin.py:7841`):

```
manager denied delete_issue: True
grandchild job settled: completed
grandchild(live) denied: True
grandchild in ROOT registry: True | session_dir: True | agent_role: 'task'
resume -> 0743eb2a19be err: None
RESUMED marker: False
RESUMED ESCALATED (activated delete_issue): True
RESUMED mcp inventory: ['mcp__linear_delete_issue', 'mcp__linear_list_issues']
```

The grandchild that was correctly denied while live comes back after a resume able to activate `delete_issue` — the same `approval_tier="exec"` write R1 was about — and it lands in its inventory. The `restricted` recomputation is doing exactly what R1 said it must not: deriving the boundary from the wrong session.

**Note on severity, stated honestly.** I checked `origin/main` and the resume rebuild widens there too (`MAIN: RESUMED escalated: True`), so the *resume* leak is not introduced by this diff. But it is not a pre-existing bug this PR merely inherits either, and I do not think it can be waved through on that basis, for two reasons:

1. **This PR is what makes it matter at depth.** On `origin/main` a restricted role got no MCP surface at all (`MAIN: manager mcp tools: []`, `can resolve mcp:// at all: False`), so there was no restricted-with-MCP lineage whose denial could be laundered. This diff creates that lineage, and R1's fix is what establishes "the denial is inherited" as an invariant. Resume is now the one path that breaks it.
2. **The PR asserts the invariant without qualification.** `GUIDE.md:62` says the denial is "inherited by everything below it ... **at any depth**", and `subagent.py:905` says the marker is "what makes the denial inherit at any depth". Both are false across a resume. Round 1 flagged precisely this class of defect — a newly written comment claiming more than the code does — and the same standard applies here.

**Suggested fix** (one line plus a field, no redesign): persist the denial on `_ChildRecord` — set it at `attach` from `getattr(child, MCP_DENIED_ATTR, False)`, and have `resume()` pass it through so `_build_child_session` ORs it into `restricted` alongside the `parent_session` read. A `restricted: bool = False` field next to `agent_role`/`effort`, carried forward for exactly the reason the comment at `comms.py:1489-1499` already gives for carrying the role forward — that comment says a defaulted resume brings back "the MCP tools restricted roles are denied ... wholesale", which is a correct description of a bug it does not actually prevent.

**Mitigating, and why this is MAJOR and not BLOCKER:** a child gets the reply-only hub (`HubChildParams` has one field, `message`; `is_child('j-mgr') -> True`), so a restricted middle agent **cannot** resume anything itself — it cannot drive this deliberately. It takes the human operator or the unrestricted root resuming a settled descendant, which then silently comes back wider than it was. Reachable in normal operation, not attacker-controlled from inside the boundary.

---

#### R2 — MINOR — RESOLVED as documented
`subagent.py:196-212`

The docstring now names the trade-off instead of implying the case cannot arise: it says plainly that a persisted tag list cannot distinguish "predates the tool" from "omitted on purpose", that a deliberately offline role gets retrieval back, that `agent show` keeps printing the narrower stored list, and that this "is a real change to user-data semantics and it is accepted here, not overlooked". That is the honest version. Pointing at a `network: no` frontmatter key as the right fix is the correct disposition. Behaviour kept — I agree with keeping it.

#### R3 — MINOR — VERIFIED FIXED
`subagent.py:213-218`

Forcing the pathological inventory (two tools both named `web_search`):

```
R3 floored names: ['read', 'web_search', 'web_fetch']
R3 duplicate present: False
mcp cannot be floored: ['read']
```

Deduped. The added comment also correctly records that it is unreachable through the normal path rather than overselling it.

#### R4 — NIT — REJECTION ACCEPTED
Reproduced the evidence myself on a `reviewer` build:

```
R4 identical across builds: True
order: ['bash','read','glob','grep','todo','web_search','web_fetch','list_variables','read_variable','hub','jobs']
```

Deterministic per role, which is what a cache prefix needs, and tail-appending matches the `hub` convention at `:1224`. Positional filtering would be a second way to do one thing. **I accept the rejection**; no change wanted.

---

#### Write/execute denial — still airtight

Re-checked across every packaged seed after these changes:

```
reviewer   forbidden={} bash=True  net=['web_fetch','web_search'] marker=True
scout      forbidden={} bash=False net=['web_fetch','web_search'] marker=True
architect  forbidden={'write'} bash=True net=['web_fetch','web_search'] marker=True
manager    forbidden={} bash=True  net=['web_fetch','web_search'] marker=True
```

No role gains `edit`; `architect` keeps `write` because its own seed grants it (`agent_seeds/architect.md:5`), which is intended and not a floor effect. A reviewer still cannot modify the diff it reviews. The floor cannot admit an MCP tool.

#### Comments asserting more than the code does

One instance, and it is R5's: `GUIDE.md:62` "at any depth" and `subagent.py:905` "inherit at any depth" are both falsified by the resume path. Every other comment in this diff I checked against behaviour held — including the unconditional-stamp rationale at `:1381`, the `activate`-left-unguarded note at `:1008`, and the R3 unreachability note.

#### Suite and gates

Touched test files on `fede656f`: `test_launch_subagent.py` **51 passed**, plus `test_role_subagents.py` / `test_agent_profiles.py` / `test_agent_tool.py` **86 passed**. For suite-level claims I defer to CI as instructed — **14/14 checks pass** on this head (`test (3.12)`, `test (3.13)`, `lint`, `type-check`, `pip-audit`, `cli-sanity`, `server-sanity`). The local box was at load average 18.5 during this review, so I did not use local timings for anything load-sensitive.

---

**Verdict: changes requested.** Round 1's R1 is genuinely and durably fixed on the launch path — the mechanism is in the committed artifact, it holds to depth 4, the regression test provably fails without it, and the unconditional-stamp reasoning is sound. R3 is fixed, R2 is honestly documented, R4's rejection is accepted. **R5 is a new MAJOR:** the same escalation returns across `hub op='resume'`, because the denial lives only as in-memory Session state and the resume path rebuilds against the root. It is a small fix (persist the flag on `_ChildRecord`, pass it through `resume`) and it needs a regression test on the resume path specifically, since neither new test covers it. Round is **not terminal**.

damianvtran:

**R5 — MAJOR — `hub op='resume'` rebuilds a restricted lineage's child without its denial — FIXED in `d4bea129`.**

Confirmed the finding before changing anything, on the real path:

```
manager denied: True
grandchild(live) marker: True
grandchild(live) denied: True
RESUMED marker: False
RESUMED ESCALATED: True
RESUMED mcp inventory: ['mcp__linear_delete_issue', 'mcp__linear_list_issues']
```

Took the suggested shape, with no deviation of substance. Two facts combine into the leak and the fix has to answer both: `resume` rebuilds with `parent_session=self._session` (the comms-owning ROOT), so the `parent_session` read added in `fede656f` finds an unrestricted session; and the leaking child is a plain `task` grandchild whose own role says "unrestricted", so `agent_role` cannot re-derive it either.

- `_ChildRecord.restricted` (`comms.py:284`) — persists the denial next to `agent_role`/`effort`.
- Stamped at `attach` (`comms.py:514`) from the live child's marker, which is the one moment the built session and its durable record are both in hand. Also carried across the prior-record fold (`comms.py:485`), so a **second** resume cannot widen either — I added that because the fold replaces a settled attempt and would otherwise drop the flag.
- `resume` passes it to `run_subagent`, threaded through `_make_runner` to `_build_child_session`, where it is **OR-ed** into the existing computation rather than assigned: a resume can only ever preserve a denial, never clear one the live computation would have found. Fresh launches leave it `False` and derive the answer exactly as before.

After the fix, same script: `RESUMED marker: True | RESUMED ESCALATED: False | inventory ['mcp__linear_list_issues']`.

**Regression tests, both with teeth checked.** Two layers, because the plumbing and the consequence fail independently:

- `test_a_resumed_child_keeps_the_denial_it_inherited` (`test_launch_subagent.py`) — end to end on really-built sessions off the real `load_seed("manager")`, asserting the resumed grandchild is refused, has no `mcp__linear_delete_issue`, still inherits `list_issues`, and is **re-stamped** so its own delegations stay denied.
- `test_a_resume_carries_the_mcp_denial_a_role_cannot_express` and `test_a_resume_of_an_unrestricted_child_does_not_invent_a_denial` (`test_comms.py`) — at the `run_subagent` seam, matching the existing role/effort resume tests. The second is the counter-check: a flag defaulting True would silently strip activation from every resumed ordinary child.

Teeth verified on a **disposable copy** (`cp -R`, not the working tree — round 1's `git checkout --` is exactly how the first R1 fix was destroyed). Reverting *only* the carry term in `_build_child_session`, leaving the field, the stamp and all plumbing in place:

```
FAILED test_a_resumed_child_keeps_the_denial_it_inherited
E  AssertionError: assert 'cannot enable new ones' in
   '# Enabled MCP tool: mcp__linear_delete_issue ...'
1 failed, 3 passed in 0.95s
```

That is the R5 escalation reappearing, with the R1 tests still passing beside it. Removing only the `restricted=record.restricted` pass-through in `resume` fails the two comms tests with `KeyError: 'restricted'`, `2 failed`.

**Verified against the COMMITTED artifact**, clean detached worktree at `d4bea129`, `git status --porcelain` empty:

```
$ git worktree add --detach /tmp/lop-v2 d4bea129
uncommitted: 0
$ pytest ... -k "resumed_child_keeps or mcp_denial or invent_a_denial or grandchild_cannot or unrestricted_lineage"
6 passed, 152 deselected in 1.99s
$ pytest test_launch_subagent.py test_comms.py test_comms_persistence.py
169 passed in 40.14s
```

**On the overclaiming comments** — the same defect class round 1 flagged, so I fixed the claims rather than only the code. `MCP_DENIED_ATTR`'s docstring now says the in-memory flag covers live delegation and names `_ChildRecord.restricted` as the durable half, stating that **both** are required for "at any depth" to hold. `GUIDE.md:62` now says the denial survives a resume. I also removed a pre-existing claim in `resume`'s comment that carrying the role forward prevents "the MCP tools restricted roles are denied" coming back wholesale — as you noted, the role never carried that, and leaving it would have described a guarantee the code does not make.

**R1 — MAJOR — resolved in round 1 (`fede656f`).** Verified by this round to depth 4 with the teeth-check you ran. No change here.

**R2 — MINOR — resolved in round 1.** Behaviour kept, docstring states the trade-off.

**R3 — MINOR — resolved in round 1 (`506a4199`).** Name-keyed dedupe.

**R4 — NIT — rejected in round 1 with evidence.** Order deterministic per role; tail-append matches the `hub` convention.

---

**Gates on `d4bea129`.** `pyright --pythonpath .venv/bin/python .`: **0 errors, 0 warnings** (33.9s). flake8 / `black --check` (569 files) / `isort --check` clean. CI is the suite evidence, as last round — the local box has been at load 40-160 from other agents all evening:

- **14/14 checks pass** on head `d4bea129`. `test (3.12)` 11m58s, `test (3.13)` 10m46s: **7400 passed, 9 skipped**. lint, type-check, pip-audit, cli-sanity, server-sanity green.

damianvtran:
### Agent review — round 3
Scope: origin/main..d4bea129f7895d8808f91afd12fefb49e222c50c

**One MAJOR (R6), and it is the third instance of the same failure mode.** R1 was a boundary that held at depth 1 and leaked at depth 2. R5 was a boundary that held live and leaked across a resume. R6 is a boundary that holds across an in-process resume and leaks across a resume *after a process restart* — because the durable half added by `d4bea129` is not, in fact, durable. The round is **not terminal**.

R5's in-memory plumbing is genuinely fixed and I verified it independently; the gap is one layer below it, at the JSON boundary.

---

#### R6 — MAJOR — NEW — `_ChildRecord.restricted` is dropped by `snapshot()`/`restore()`, so the denial dies on a process restart
`local_operator/harness/comms.py:790-812` (`snapshot`) and `:872-890` (`restore`), against `:284`

`d4bea129` makes `_ChildRecord.restricted` "the durable half". It is not persisted. `snapshot()` enumerates the fields that cross a process exit and lists `agent_role` and `effort` — the two fields `restricted` was explicitly built to ride alongside — but not `restricted`. `restore()` correspondingly never reads it, so every rehydrated record comes back `False`.

This is not a theoretical serialization nit. `restore()`'s own docstring names itself the resume basis: *"This is the resume basis: `hub op='resume'` needs the `job_id → session_dir` mapping to relaunch a child against its old transcript."* `session.py:6271` persists `comms.snapshot()` into the roster sidecar and `session.py:6379` rehydrates it via `comms.restore()` at boot. So the flag survives exactly as long as the process does, which is the same lifetime the *in-memory* marker already had — the property R5 was filed to fix.

Reproduced on the committed artifact, clean detached worktree at `d4bea129`, `git status --porcelain` empty:

```
live record   : restricted=True agent_role='task' effort='high'
snapshot keys : has 'restricted'? False | has 'agent_role'? True | has 'effort'? True
after restart : restricted=False agent_role='task' effort='high'

VERDICT: denial survived restart? False  (role survived? True)
```

The role survives the round trip and the denial does not — which is precisely the asymmetry `_ChildRecord.restricted` exists to remove.

At the `run_subagent` seam, I took your own `test_a_resume_carries_the_mcp_denial_a_role_cannot_express` verbatim and inserted **only** the restart (`snapshot()` → new `SubagentComms` → `restore()`) between `detach` and `resume`:

```
>       assert seen["restricted"] is True, "the denial must survive a restart too"
E       AssertionError: the denial must survive a restart too
E       assert False is True
1 failed
```

And end to end on really-built sessions off the real `load_seed("manager")`, mirroring `test_a_resumed_child_keeps_the_denial_it_inherited` but feeding `restricted` from what actually survived the round trip rather than hard-coding `True`:

```
live grandchild denied      : True
record stamped restricted   : True
carried across restart      : False
resumed denial text         : # Enabled MCP tool: mcp__linear_delete_issue Server tool: de
ESCALATED (has delete_issue): True
resumed marker              : False
```

That is the R5 escalation, verbatim, on the restart path: the grandchild of a restricted `manager` regains `mcp__linear_delete_issue`, an `approval_tier="exec"` write, and comes back **unstamped** so its own further delegations start clean too.

Why the existing tests do not catch it: both R5 regressions resume in-process off a live `_records` map. `test_a_resumed_child_keeps_the_denial_it_inherited` passes `restricted=True` into `_build_child_session` as a literal, so it asserts the consequence *given* the flag; the comms test asserts the flag reaches `run_subagent` *without* a restart. Neither crosses `snapshot()`. `tests/unit/harness/test_comms_persistence.py` contains no `agent_role` or `restricted` assertion at all, so the persistence layer has no coverage pinning which fields must survive.

**Fix** (two lines, mirroring `agent_role` exactly): add `"restricted": record.restricted` to the `snapshot()` row dict, and `restricted=bool(row.get("restricted"))` to the `_ChildRecord(...)` construction in `restore()`. Defaulting missing to `False` is right for old sidecars — it matches today's behaviour and cannot invent a denial. Worth pinning with a persistence test that round-trips a restricted record, since that file currently pins none of these fields.

**Doc consequence, same class as R1/R5.** With R6 open, three claims introduced or edited by `d4bea129` are false:
- `subagent.py:917` — "The durable half is `_ChildRecord.restricted`, which persists this flag". It does not persist it; it holds it in memory for the process lifetime.
- The same docstring's "BOTH halves are required for the 'at any depth' claim to hold" — both halves are present and the claim still fails after a restart.
- `GUIDE.md:62` — "It also survives `hub op='resume'`". True in-process, false across a restart, and a restart is the common case for resuming a child that settled hours ago.

These become true once the two lines land; I flag them so the remediation does not fix the code and leave the comments overclaiming, which is the pattern R1 flagged.

---

#### What I verified and found sound

- **R5's in-memory plumbing is real, and the tests have teeth.** On a disposable `cp -R` copy (not the working tree), reverting *only* the attach stamp at `comms.py:514` fails `test_a_resume_carries_the_mcp_denial_a_role_cannot_express` (`1 failed, 1 passed` — the counter-check still passes, so the pair does not pass either way). On a second copy, reverting *only* the `restricted or` carry term in `_build_child_session` fails `test_a_resumed_child_keeps_the_denial_it_inherited` with the `delete_issue` escalation, while `grandchild_cannot` and `unrestricted_lineage` still pass (`1 failed, 3 passed`). The stamp is OR-ed at attach, OR-ed across the prior-record fold (`:485`), and OR-ed rather than assigned in `_build_child_session`, so no in-memory path can clear a denial.
- **R1 remains resolved.** The parent-session read is the third term of the same OR; the round-2 depth-4 verification stands unchanged by `d4bea129`.
- **Adjacent paths I attacked and found closed.** `_build_child_session` is still the only child-Session construction path (the one non-test call site is `subagent.py:455`; `session_factory.py:1727` is root-session construction, not a child). `hub op='resume'` fanned across `"all"` just loops `comms.resume(job_id, message)` per target (`builtin.py:7839-7843`), so it inherits whatever the single path does — no separate leak, but it does mean R6 fans out too. A resume whose record was **evicted** fails closed: `_evict_overflow` deletes the record outright and `resume` then returns `unknown subagent`, so there is no path that resumes with a half-built record. A child resumed by the ROOT is the R5 case and is covered in-process. `agent_role` re-resolution cannot re-derive the denial by design (the leaking child's role says unrestricted), which is exactly why the persisted flag matters and why losing it at the JSON boundary is fatal.
- **Write/execute denial remains airtight.** `READ_ONLY_TOOLS` contains none of `{edit, write, bash, eval, browser, task}`; `_with_network_floor` filters `available` by membership in `READ_ONLY_NETWORK_TOOLS = ('web_search', 'web_fetch')`, so the widest possible outcome is those two names and it cannot admit an MCP tool (`mcp__<server>_<tool>` never matches). The mutators visible on `reviewer`/`architect`/`manager` seeds (`bash`, `write`) are their own declared allowlists, pre-existing and unrelated to the floor.
- **R2 (MINOR) resolved** — trade-off stated in the `_with_network_floor` docstring with the `network: no` opt-out named as the future fix. **R3 (MINOR) resolved** — dedupe is now keyed by name against `missing` as well as `allowed`, with the comment explaining the pathological case. **R4 (NIT) rejected with evidence in round 1**; I agree, order is stable per role.
- **Gates.** The five touched test files: **206 passed** in 15.85s on the clean `d4bea129` checkout. CI on head `d4bea129`: **14/14 checks pass**, `test (3.12)` and `test (3.13)` both green. I did not draw suite-level conclusions from local timing — this box was at load average 59.31 throughout.

---

**Verdict: not clean. One MAJOR (R6) open; R1–R5 are otherwise genuinely resolved.** This round is **NOT TERMINAL** — R6 needs a fix and a remediation reply before merge on the code-review axis. It is a two-line change plus a persistence test and the three comment corrections; I would expect round 4 to be short.

To be explicit about the thing you asked me to settle: I did look for the third adjacent path, and it exists. The pattern is now established well enough to name — every instance has been a fact held in a *narrower scope* than the boundary it protects (session attribute, then process memory), with a comment asserting the wider guarantee. The remediation for R6 should carry a test at the widest surviving scope, the JSON round trip, so the next scope boundary has one too.

damianvtran:

**R6 — MAJOR — `_ChildRecord.restricted` dropped by `snapshot()`/`restore()` — FIXED in `cc49eddc`.**

Confirmed before changing anything. You were right, and "the durable half" was my claim, not the code's:

```
live record   : restricted=True agent_role=''
snapshot keys : has 'restricted'? False | has 'agent_role'? True
after restart : restricted=False
VERDICT: denial survived restart? False
```

Two lines, mirroring `agent_role` exactly as suggested: `"restricted": record.restricted` in the snapshot row (`comms.py:816`), `restricted=bool(row.get("restricted"))` in restore (`:902`). A missing key reads `False`, correct for a sidecar written before the field existed: it reproduces today's behaviour and cannot invent a denial. Consistent with the rest of the flag's handling, restore can only ever *add* a denial; the attach stamp and the live computation OR on top. After: `has 'restricted'? True | after restart: restricted=True`.

**Tests now cross the boundary that was missed.** You identified exactly why it slipped, so I fixed the coverage gap and not just the line:

- `test_a_capability_denial_survives_the_snapshot_round_trip` — through the **real** `snapshot()` → `restore()`, not a hand-built row, because the defect was that the real serializer omitted a field a hand-built record would have carried.
- `test_an_unrestricted_record_round_trips_as_unrestricted` — counter-check, including a **legacy row with no key at all**, so persistence cannot invent a denial and strip MCP from every resumed ordinary child.
- `test_every_security_relevant_record_field_is_persisted` — the guard for the *class*. It asserts `agent_role`, `effort` and `restricted` all appear in both `snapshot` and `restore`, with a failure message naming the consequence. The list is explicit rather than derived from the dataclass because most of `_ChildRecord` is deliberately not persisted (futures, subscriptions, the live child handle), so "every field" would be wrong.
- `test_a_resume_carries_the_mcp_denial_a_role_cannot_express` — extended to resume a **second** time through a real restart, which is the assertion your reproduction showed was missing.

**Teeth-checked on a `cp -R` copy**, reverting *only* the two persistence lines and leaving the field, attach stamp and resume plumbing intact:

```
FAILED test_a_capability_denial_survives_the_snapshot_round_trip - KeyError: 'restricted'
FAILED test_an_unrestricted_record_round_trips_as_unrestricted - KeyError: 'restricted'
FAILED test_every_security_relevant_record_field_is_persisted - AssertionError: 'restricted'
  gates a resumed child's capabilities but snapshot() does not persist it
FAILED test_a_resume_carries_the_mcp_denial_a_role_cannot_express - the denial must
  survive a restart too
4 failed, 116 deselected
```

**Scope audit — I went looking for the fourth instance myself.** Your framing is right and it is the useful lesson from this PR: every instance has been a fact held in a narrower scope than the boundary it protects (session attribute → process memory → JSON boundary), each with a comment asserting the wider guarantee. What I checked, all on the committed `cc49eddc` tree:

| # | scope crossed | result |
|---|---|---|
| A | every `_ChildRecord(...)` construction site (3 in `comms.py`) | all covered |
| B | every read/write of `.restricted` (5 sites) | all OR-ed, none assigning |
| D1 | specialist child (non-role, own system prompt) of a restricted parent | `True` |
| D2 | child launched with a model/effort override | `True` |
| D3 | depth-5 chain | `True`, denied |
| D4 | **double** restart (snapshot→restore→snapshot→restore) | `True` |
| D5 | real `json.dumps`/`loads` round trip, not just the dict | `True` |
| D6 | child launched **by** a restart-resumed child (composes R6 with R1) | `True`, denied, inventory `['mcp__linear_list_issues']` |

D6 is the one I most expected to fail — a restart-resumed child must be **re-stamped**, not merely honoured once, or its own later delegations start clean. It holds. I found no fourth gap. The remaining theoretical scope is a denial crossing machines (a sidecar copied to another host), which is out of scope here and would ride on the same persisted field.

**Docs corrected**, since leaving them overclaiming is the pattern R1 flagged. The marker docstring now enumerates all three scopes and states plainly that all three are required, with the round each one came from; `_ChildRecord.restricted` says it outlives the process rather than the map; `GUIDE.md:62` says the denial survives a resume *after a restart*.

**R1 (round 1, `fede656f`), R5 (round 2, `d4bea129`), R2/R3 (round 1), R4 (rejected round 1)** — all as previously resolved, unchanged here.

---

**Verification on the COMMITTED artifact**, clean detached worktree at `cc49eddc`, `git status --porcelain` empty: the R6 reproduction now reports `denial survived restart? True`; all six scope audits above; five touched test files **209 passed**.

**Gates.** pyright `--pythonpath .venv/bin/python`: **0 errors, 0 warnings** (20.5s). flake8 / `black --check` (569 files) / `isort --check` clean. **CI 14/14 pass** on head `cc49eddc`; `test (3.12)` and `test (3.13)` green, **7495 passed, 9 skipped**. No local suite conclusions drawn — this box was at load 55-63.

damianvtran:
### Agent review — round 4
Scope: origin/main..cc49eddc99a6ce7f025df019f67e95b9421164ae

Independent pass on a clean detached checkout of `cc49eddc` (`git clone --shared --no-checkout` + `checkout --detach`, `git status --porcelain` empty), not the implementer's worktree. Every number below is one I produced; I did not reuse the round-3 remediation's evidence. Box load at the start was `load averages: 4.46 6.94 12.17`, so no timing claims are made.

---

## 1. R6 verified in the committed artifact

**Persistence is present and correctly directional.**

- `snapshot()` writes `"restricted": record.restricted` (`comms.py:812`); `restore()` reads `restricted=bool(row.get("restricted"))` (`comms.py:899`).
- **Missing key degrades to False.** `bool(None)` on an absent key; I drove a legacy row through the real `restore()` with the key deleted and got `restricted=False` — no raise, no invented denial.
- **Restore can only ADD a denial, never clear one.** Three separate places OR rather than assign: `attach` (`record.restricted or bool(getattr(child, MCP_DENIED_ATTR, False))`, `comms.py:521`), the attempt fold (`record.restricted or prior.restricted`, `comms.py:496`), and the live computation in `_build_child_session` (`restricted or ... or ... or ...`, `subagent.py:1170`). I confirmed `record_launch`'s `existing is not None` branch (`comms.py:434-445`) overwrites `label`/`prompt`/`agent_role`/`effort` but deliberately does **not** touch `restricted`, so the eager-task-factory ordering cannot clear the stamp either.

**Teeth-check re-run independently.** On a `cp -R` copy I deleted *only* the two persistence lines with a scripted string replace (`git diff --stat` = `1 file changed, 2 deletions`), never `git checkout --`:

```
3 failed, 11 passed
FAILED test_a_capability_denial_survives_the_snapshot_round_trip - KeyError: 'restricted'
FAILED test_an_unrestricted_record_round_trips_as_unrestricted   - KeyError: 'restricted'
FAILED test_every_security_relevant_record_field_is_persisted    - AssertionError: 'restricted' gates a resumed child's capabilities but snapshot() does not persist it
```

The tests have teeth. Reproduced, not taken on trust.

---

## 2. My own fourth-gap hunt (not a re-run of D1-D6)

I designed seven probes against the paths the remediation did **not** name, driving the real `SubagentComms`, the real `_compact_subagent_record`, and a real `json.dumps`/`json.loads` hop:

| Probe | Path | Result |
|---|---|---|
| P1 | Live `attach` stamp on a child whose own `agent_role` is plain `task` | `restricted=True` |
| P2 | Full pipeline: `snapshot` → `_compact_subagent_record` → real JSON → `restore` | `True` |
| P3 | **Double** restart (snapshot the restored comms, restore again) | `True` |
| P4 | Second resume folds onto the same `session_dir` with a live child whose stamp was **lost** | `True` (fold preserves) |
| P5 | Legacy sidecar, key absent entirely | `False` (correct degrade) |
| P6 | **Forward compat**: row carrying an unknown future key | tolerated, `True` |
| P7 | Second `restore()` with a downgrade row against a present record | `True` (restore skips existing job_id) |

**`_compact_subagent_record` was the specific thing I went looking for** — it sits between `snapshot()` and the sidecar in `session.py:6273` and is exactly the kind of narrower scope the previous three rounds hid in. It is a `dict(record)` copy that truncates only `prompt`/`error_text`/`result_text`, so `restricted` passes through untouched (P2 confirms end to end).

Other candidates I checked and cleared:

- **`agent_role` re-resolution at resume cannot reintroduce an unrestricted profile.** `resume` passes `restricted=record.restricted` (`comms.py:1590`) and `_build_child_session` ORs it in as the *first* term, so a re-resolved profile can only widen the OR, never override it.
- **Depth-N live inheritance.** I computed the real `restricted` expression across a 5-hop chain (`manager` → plain `task` → specialist declaring `tools: []` → plain → plain): denied at every hop. Counter-check: a genuinely unrestricted 3-hop lineage stays `False`, so there is no false-positive denial stripping MCP from ordinary children.
- **No second activation route.** `make_mcp_resolver` has exactly two callers (`subagent.py:1060`, `session_factory.py:1537`); `refresh_tools` has no caller that mints MCP schemas outside `_child_mcp_wiring.activate`; `grep` finds no enable/activate MCP path in `tools/builtin.py`. The `mcp://` namespace rejection at `subagent.py:1195` closes the fall-through to the parent's own resolver.
- **Follower path is not a hole.** `SnapshotSubagentComms` (`frontend_state.py:750`) is a read-only graph facade with no `resume`, `restore` or `snapshot` — a follower cannot relaunch a child, so it cannot launder a denial.

**End-to-end resolver denial, driven for real:**

```
mcp://linear                          -> ANSWERED   (discovery preserved)
mcp://linear/delete_issue             -> DENIED
mcp://linear/mcp__linear_delete_issue -> DENIED     (both name forms)
guide://x                             -> PASSTHROUGH(None)
activate() calls under deny resolver: []            <- zero
control (no deny reason): activate() calls: [('linear','delete_issue')]
```

**Write/execution denial is airtight.** I drove `_with_network_floor` against an inventory containing `edit`, `write`, `bash`, `eval`, `browser`, `task` and an `mcp__linear_delete_issue`: floored output was `['read','glob','web_search','web_fetch']` only. No duplicates when the tools are already present, empty floor on a session with web disabled, and an `mcp__`-prefixed tool can never match the two floored names. Both floored tools are genuinely `approval_tier="read"` (`web_fetch/tool.py:461`, `web_search/tool.py:360`). `SCOUT_TOOL_ALLOWLIST == READ_ONLY_TOOLS` now holds by derivation rather than by a second hand-maintained copy, which removes the original divergence.

**I found no fourth instance of the pattern.**

---

## 3. Doc claims

All three checked against the code, all now true:

- **`MCP_DENIED_ATTR` docstring's three scopes** (`subagent.py:915-944`) — scope 1 (session attribute) at `subagent.py:1436`/`1173`, scope 2 (`_ChildRecord.restricted`) at `comms.py:521`/`1590`, scope 3 (snapshot/restore) at `comms.py:812`/`899`. The "ALL THREE are required" claim is now backed: my teeth-check independently proved removing scope 3 alone reopens it, and the R1/R5 tests cover the other two.
- **Record field comment** (`comms.py:273-290`) — accurate, including the "rebuilds against `self._session`, the comms-owning ROOT" claim, which I confirmed at `comms.py:1584`.
- **GUIDE.md** — the "boundary is drawn around CHANGE, not around reach" paragraph and the `_with_network_floor` paragraph both match behaviour I executed. The "at any depth" and "survives a resume after the process has restarted" claims are the ones my P1-P7 and depth probes exercise.

---

## 4. Gates and CI

Run over the whole tree in the clean checkout, exactly as CI does:

```
flake8 .                              -> rc=0
black --check . (26.1.0)              -> 569 files unchanged
isort 5.13.2 --check .                -> rc=0
pyright --pythonpath .venv/bin/python -> 0 errors, 0 warnings, 0 informations
tests/unit/harness + mcp + session    -> 1124 passed
touched suites (6 files)              -> 258 passed
```

**On the `test (3.12)` flake: I agree, and it is now moot.** `gh pr checks 384` shows **both** runs on `cc49eddc` fully green, including `test (3.12)` passing twice (12m35s and 12m54s). `test_resume_subagents_todos.py` is untouched by this diff (`git diff f32802e3..cc49eddc --stat` on that path is empty; last change was #359 on main) and passes 19/19 locally. Infrastructure flake, since resolved on the identical SHA.

**One environmental note, so nobody mistakes it for a regression.** The full `tests/unit` run in my detached clone reported `24 failed, 7381 passed`, all in `tests/unit/tools/test_eval_tool.py` with `ModuleNotFoundError: No module named 'local_operator'` in the eval kernel subprocess. I baselined it: the **same** failures occur at the base commit `f32802e3` (25 failed) and on **current `main`** in `~/local-operator` itself (24 failed). Pre-existing, caused by the eval kernel subprocess not resolving the package through a symlinked venv, and entirely unrelated to this diff.

---

## Findings

No BLOCKER. No MAJOR.

**MINOR-1 — the class-guard test does not guard the class it claims to.** `tests/unit/harness/test_comms_persistence.py:390` (`test_every_security_relevant_record_field_is_persisted`) documents itself as "the guard for the CLASS of defect R6 was, not just that instance", but its field list `("agent_role", "effort", "restricted")` is hand-maintained. I tested this directly: on a `cp -R` copy I added a new unpersisted security field to `_ChildRecord` —

```python
sandbox_escape_allowed: bool = False   # deliberately absent from snapshot/restore
```

— and the guard **passed** (`1 passed in 2.52s`). It is a regression guard for the three named fields, which is genuinely useful and worth keeping; it is not a guard for the class. The test's own docstring is honest about the mechanism ("explicit rather than derived from the dataclass") and gives a sound reason, so this is a wording overclaim rather than a defect. One-line fix if wanted: soften the docstring's first line to say it pins the three known capability-gating fields, since a future reader who trusts the "CLASS" claim will not add their new field to the list.

**NIT-1 — a no-op assertion.** `test_comms_persistence.py:368`, `assert MCP_DENIED_ATTR  # the marker the stamp above mirrors`, asserts a non-empty string constant and can never fail. The comment above it describes a counter-check ("an unrestricted record must not come back denied") that this line does not perform — the real counter-check is the next test, `test_an_unrestricted_record_round_trips_as_unrestricted`, which does it properly. Harmless, but it reads as coverage that is not there. Delete the line, or keep only the comment.

That is the whole list. I am not padding it: I went looking specifically for a fourth scope and did not find one.

---

## Status of prior rounds

| Round | Finding | Status |
|---|---|---|
| R1 | Denial escaped at depth 2 (session-attribute scope) | **Resolved** — verified live to depth 5 through plain-`task` and empty-allowlist-specialist hops |
| R2/R3 | Network floor trade-off stated; duplicate-name dedupe | **Resolved** — trade-off documented at `subagent.py:191-201`; dedupe verified, no duplicates |
| R5 | Denial escaped a resume (process-memory scope) | **Resolved** — `run_subagent(restricted=...)` carry verified; fold preserves; `record_launch` cannot clear |
| R6 | Denial escaped a restart (JSON snapshot scope) | **Resolved** — persistence verified in the committed artifact, legacy degrade correct, teeth-check reproduced independently |

R1 through R6 are **all genuinely resolved**. The three-scope containment is complete on every path I could construct, the write and execution denials are untouched by the network floor, and the documentation now matches the code.

---

## Verdict: clean — TERMINAL

No blocker and no major remains. **This round is TERMINAL**; MINOR-1 and NIT-1 are follow-ups, not grounds for a round 5. The PR is **ready to merge on the code-review axis** against head `cc49eddc99a6ce7f025df019f67e95b9421164ae`, with CI green on that SHA.

damianvtran:
### Agent review — round 4 (freshness confirmation after rebase)

Reviewer: manager agent (anthropic/claude-opus-5)
Scope: origin/main..8f293c71e748b027528e4f4f08731a4bf4fcc7a1

Round 4 was posted clean against `cc49eddc`. `main` then advanced to `56e66c51` and, critically, already carried a version bump to **0.42.17**, so this branch's `0.42.16` would have been a downgrade. Rebased onto `56e66c51` and resolved the only conflict (the `pyproject.toml` version line) to **0.42.18**.

The rebase rewrote every SHA, so the round-4 scope line no longer names the current head. Rather than assume the round still applies, I verified that the reviewed change is unchanged:

```
FILES="local_operator/agent_profiles.py local_operator/agent_seeds \
  local_operator/guides/agents/GUIDE.md local_operator/harness/subagent.py \
  local_operator/harness/comms.py local_operator/harness/types.py \
  local_operator/mcp/resources.py tests/..."

git diff 8f3dd520..cc49eddc -- $FILES | grep -v '^index ' > a.patch   # reviewed
git diff 56e66c51..HEAD      -- $FILES | grep -v '^index ' > b.patch   # current
diff -q a.patch b.patch
```

```
1477 a.patch
1477 b.patch
IDENTICAL - reviewed change survived rebase byte-for-byte
```

The only delta between `cc49eddc` and `8f293c71` outside that set is the version line and the unrelated work merged into `main` in the interim (clipboard paste, subagent-view streaming, word-caret). No finding from any round is affected, so round 4's clean verdict carries to `8f293c71` on the merits; this note records that judgement rather than silently relying on a stale scope line.

CI re-validated on the rebased head before merge.

PR #401, opened 2026-08-29
fix(mcp): stop the OAuth refresh lock from freezing the TUI on /resume

damianvtran:
# PR Description

## Summary of Changes

`/resume` froze the entire TUI, and MCP servers sat on "connecting" forever. Both
symptoms are one bug in `local_operator/mcp/auth.py::_oauth_refresh_lock`, where
three defects compound.

`/resume` disposes the MCP manager (`_cmd_resume` → `_resume_session` →
`_reload_session`), which cancels in-flight connect tasks. Those tasks were
parked in the OAuth refresh lock, whose acquire was a bare blocking
`fcntl.flock(fd, LOCK_EX)` on a worker thread — while its `finally` closed the
descriptor **from the event-loop thread**. On macOS/BSD, `close()` of a
descriptor with a sibling thread parked in `flock()` blocks until that `flock()`
returns. With the lock held by another process, that is never. The loop stopped
dead: no repaint, no input.

The `sample(1)` stack of a live stuck `lop` shows exactly this — the main thread
in `task_step → _gen_throw ×3 → async_gen_asend_throw → os_close → close` (a
cancellation thrown into the nested async context managers, landing in that
`finally`) while four `asyncio_N` worker threads sit in `fcntl_flock_impl →
flock`.

### D1 — the POSIX acquire was unbounded

`_lock_exclusive` bounded only the Windows path, with a 60s deadline loop whose
own docstring says the bound exists so "a leaked lock (killed process) cannot
park a connect eternally". POSIX — the platform this actually ships on — got a
bare blocking `flock`. A lock leaked by a killed process parked a connect
forever.

Now a bounded non-blocking `LOCK_EX|LOCK_NB` retry loop. The bound,
`LOCK_ACQUIRE_TIMEOUT_S = 15.0`, is derived from the honest budget of the section
it guards: one token POST (capped at `REFRESH_HTTP_TIMEOUT_S = 10.0`) plus a
SQLite read. A peer holding it legitimately clears well inside that; overrunning
it means the holder is not working, and waiting longer than the work can possibly
take buys nothing and costs a hung connect.

### D2 — cancellation stranded the worker, then `os.close()` deadlocked the loop

Restructured so the acquiring thread **owns its descriptor end to end**:
`_acquire_locked_fd` opens the fd, retries to the deadline, and on every
non-success path closes it *itself, on that thread, after its last lock syscall
has returned*. The fd only ever reaches the coroutine when no lock call is
outstanding on it, so by construction there is no moment at which the event loop
could close a descriptor a thread is parked on.

On cancellation the coroutine sets a `threading.Event` (so the worker abandons
its retry loop within one 50 ms tick rather than running to the full bound) and
hands the fd's fate to a done-callback on the shielded task. It touches the
descriptor itself not at all, which is what keeps the loop alive.

### D3 — the lock was global, which manufactured the contention

`_oauth_refresh_lock(server_url)` took `server_url` and **never used it**: every
server shared `config_dir()/mcp_oauth_refresh.lock`. All OAuth connects, across
all servers and all processes, serialised on one file. With six OAuth servers and
nine concurrent `lop` processes that queue never drained — only the two non-OAuth
servers (`google-workspace` stdio, `slack` static header) ever reached
"connected", matching the reported symptom exactly.

Observed live on the reporting machine while five wedged processes held it:

```
$ lsof ~/.local-operator/mcp_oauth_refresh.lock | tail -n +2 | awk '{print $2}' | sort -u
67593 68401 69760 70899 71531 92212
$ lsof ~/.local-operator/mcp_oauth_refresh.lock | tail -n +2 | wc -l
      28      # 28 fds held on ONE lock file
```

The lock is now keyed per server by a SHA-256 digest of the URL (URLs contain
`/`, `:` and query strings that are not filename-safe; a digest is stable across
runs and machines with no escaping scheme to keep in step).

**On the legacy global file:** deliberately left in place and not cleaned up.
Unlinking it would silently drop mutual exclusion for any peer process still
running an older build — its flock survives on the unlinked inode while a new
process creates a fresh file and takes an uncontended lock. It is a zero-byte
file; leaving it costs nothing.

### Degrade, never hang

On timeout the context manager yields `False` and the caller proceeds instead of
blocking. `ensure_mcp_oauth_fresh` skips the **proactive** refresh and connects
anyway; the SDK's own in-flow refresh still runs, and the coordinator wrapper
re-reads the store before it does. A skipped optimisation costs a round trip;
blocking the connect costs the session.

The caller was verified rather than assumed: `McpManager._ensure_oauth_fresh`
wraps the call in `except Exception` and logs at debug, so it tolerates both the
new return value and any `OSError` the acquire can still raise.

The coordinator's three call sites are handled by what they actually do. The two
read-only paths (`_adopt_freshest_stored_token`, `_adopt_peer_token_once`) adopt
the freshest stored token **regardless** of whether the lock was taken — they
only read and rewrite our own header, never spend the refresh token, so an
unlocked pass is safe and still beats the alternative. The one path that *would*
spend the token returns early when unlocked, after the re-read. That preserves
the invariant the coordinator exists for: the SDK never runs with an in-memory
refresh token older than storage.

### Audit for the same pattern elsewhere

Every `flock`/`os.close` pairing in the tree was checked for the close-on-blocked-fd
hazard:

- `session_lease.py::_stale_recovery_right` — **safe, left alone.** Already
  `LOCK_EX|LOCK_NB`, so no thread ever parks; the close cannot block.
- `tools/group_reaper.py::_ledger_lock` — blocking `flock`, but **not the same
  defect class**: it is a synchronous `@contextlib.contextmanager` on one thread,
  so there is no second thread for the close to race and no event loop to freeze.
  Reported, not fixed here.
- No other site pairs a blocking lock with a cross-thread close.

## Related Issue(s)

- User-reported `/resume` freeze and MCP servers stuck on "connecting" (no ticket).

## Impact

- No breaking changes, no dependency updates, no public API change.
- Version bumped to `0.43.2` (patch: a bug fix, per AGENTS.md versioning — no new
  capability).
- Behaviour change is strictly a removal of hangs. The refresh's race-freedom is
  unchanged in the common case; it degrades to best-effort only when the lock
  cannot be taken within 15s, which previously meant hanging forever.

### The deadlock, reproduced on `origin/main` and fixed here

A probe holds the lock from a **foreign process** (flock is per open-file-
description, so a same-process acquire would not contend), enters the context
manager in a task, cancels it mid-acquire, and runs a heartbeat coroutine
standing in for the TUI's repaint pump.

**BEFORE — `origin/main` (e641a888).** The process had to be killed; even the
15s `asyncio.wait_for` never fired, because the event loop itself is dead:

```
$ cd /tmp/lop-lock-base && .venv/bin/python /tmp/lock_probe.py
foreign holder(s) hold: [.../mcp_oauth_refresh.lock, .../mcp_oauth_refresh_1cbd18bf1818c780.lock]
cancelling mid-acquire (loop ticks so far: 20)
TIMEOUT after 90.0s (process killed)
```

**AFTER — this branch:**

```
$ cd /tmp/lop-oauth-lock && .venv/bin/python /tmp/lock_probe.py
foreign holder(s) hold: [...]
cancelling mid-acquire (loop ticks so far: 20)
unwind took 0.00s
RESULT: OK — cancellation unwound in 0.000s, loop never blocked
```

### The bounded acquire and the per-server split, on the real code

```
$ cd /tmp/lop-oauth-lock && .venv/bin/python /tmp/lock_bound_probe.py
lock(A) = mcp_oauth_refresh_1cbd18bf1818c780.lock
lock(B) = mcp_oauth_refresh_fb39103c7d2edac2.lock
per-server split: YES

foreign process holds server A's lock (mcp_oauth_refresh_1cbd18bf1818c780.lock)
server B acquire: locked=True in 0.002s
server A acquire: locked=False after 15.03s
  -> body still ran, so the connect proceeds instead of hanging

B uncontended by A's held lock: PASS
A bounded at ~15s and degraded: PASS
```

### End-to-end on the refresh path, against the user's six real OAuth server URLs

The plain-connect path cannot discriminate the fix (with fresh tokens
`ensure_mcp_oauth_fresh` returns before ever taking the lock), so this drives the
path that *does* take it, with a scratch config dir and fabricated tokens — it
never touches the real `auth.db` and never spends a real refresh token.

**BEFORE — `origin/main`:** killed at 400s having printed **no scenario result at
all**; it never got past scenario A.

```
$ cd /tmp/lop-lock-base3 && .venv/bin/python /tmp/e2e_refresh_path.py
lock naming: GLOBAL (pre-fix)
--- 100s elapsed; still running? ---
STILL RUNNING -> main is DEADLOCKED
TIMEOUT after 400.0s (process killed)
```

**AFTER — this branch**, completing in ~15s:

```
lock naming: per-server
foreign process holds: mcp_oauth_refresh_1cbd18bf1818c780.lock

A. cancelling a connect mid-acquire (the /resume path)
   unwound in 0.000s | loop scheduling after unwind: True
   -> PASS

B. other servers' refresh path while notion's lock is held
   linear       locked=True  in   0.00s
   notion       locked=False in  15.05s  contended (expected)
   datadog      locked=True  in   0.00s
   cloudflare   locked=True  in   0.00s
   hubspot      locked=True  in   0.00s
   -> others unblocked: PASS
   -> held lock bounded+degraded: PASS
```

### End-to-end on the real app with the real config (7 servers)

The real `McpManager` against `~/.local-operator` — all 6 OAuth servers plus the
2 non-OAuth ones — **while the reporter's wedged processes still hold the legacy
global lock**:

```
$ cd /tmp/lop-oauth-lock && .venv/bin/python /tmp/e2e_mcp_connect.py
enabled servers (7): ['linear', 'google-workspace', 'slack', 'notion', 'datadog', 'cloudflare', 'hubspot']

startup finished in 1.6s (event-loop ticks: 26)

server                 status         detail
----------------------------------------------------------------------
linear                 connected
google-workspace       connected
slack                  connected
notion                 connected
datadog                connected
cloudflare             connected
hubspot                connected

servers still 'connecting': NONE
RESULT: PASS - every server reached a terminal state
```

### `/resume` into two real sessions, driving the actual TUI in a pty

The real `lop` TUI, booted with the real config, sent `/resume <id>`. The status
bar climbing `⊙ 2 → 7 MCP` and settling on "MCP ready" is the connect machinery
running live through the resume that used to freeze it.

Session `1835c1163ff4` ("Harden Against Credential Leak Patterns"):

```
$ NO_HOLD=1 env -u NO_COLOR TERM=xterm-256color python3 /tmp/resume_pty.py \
      /tmp/lop-oauth-lock 1835c1163ff4 fixed-real
[fixed-real] booted, 235446 bytes painted
[fixed-real] after /resume: 26507 bytes over 32s
[fixed-real] intervals that repainted: 2/12
[fixed-real] RESULT: PASS - app kept painting

$ grep -o "⊙ [0-9] MCP\|MCP ready: .*" /tmp/resume-fixed-real.txt | uniq | tail -3
⊙ 6 MCP
MCP ready: 7 servers, 203 tools
⊙ 7 MCP
```

Session `2d2a05bf777b` (the pasting session):

```
[fixed-real2] RESULT: PASS - app kept painting
⊙ 0 MCP → ⊙ 1 → 2 → 3 → 4 → 5 → 6 → 7 MCP
MCP ready: 7 servers, 203 tools
```

### Visual validation (AGENTS.md "Visual validation")

Frames exported from the **real `OperatorApp`** (which loads
`local_operator.tcss`), driven through a contended-then-cancelled connect, and
rendered and viewed rather than eyeballed as markup. The before-frame was taken
from a throwaway `origin/main` worktree, never by stashing.

- **before** (`origin/main`): the still is literally the *last frame the user
  ever sees* — the capture had to be taken **before** the cancel, because after
  it no screenshot is possible: the loop is dead. Status bar reads `connecting…`,
  and the process was still wedged 70s later having never reached the unwind.
- **after** (this branch): the cancel unwinds in **0.024s**, the app keeps
  painting, and a further user message renders after it ("the screen is still
  live and accepting input after the cancel") — proof the loop still schedules
  and the composer still accepts input.

### Regression test

`TestOAuthRefreshLock` in `tests/unit/mcp/test_auth.py` — five tests covering the
per-server split, the bounded degrade, cross-server non-contention, normal
release, and the cancellation guard.

The cancellation guard runs its probe in a **child process with a hard timeout**
on purpose: against the pre-fix code the loop is genuinely dead, so an in-process
version would hang the whole suite rather than fail (`asyncio.wait_for` cannot
fire on a loop that is not running). The child turns that hang into a
deterministic failure.

Verified to fail on `origin/main` **for the right reason** — the child genuinely
deadlocked for its full 90s bound:

```
$ cd /tmp/lop-lock-base && .venv/bin/python -m pytest tests/unit/mcp/test_auth.py \
      -k TestOAuthRefreshLock -q -p no:randomly
FAILED ...::test_cancelling_mid_acquire_unwinds_without_blocking_the_loop
    - subprocess.TimeoutExpired: Command '[...]' timed out after 90 seconds
FAILED ...::test_each_server_gets_its_own_lock_file
FAILED ...::test_one_servers_held_lock_does_not_delay_another_server
FAILED ...::test_a_held_lock_is_abandoned_at_the_deadline_not_waited_on_forever
FAILED ...::test_the_lock_is_released_for_the_next_waiter
5 failed in 121.46s (0:02:01)
```

And to pass on this branch in under two seconds:

```
$ cd /tmp/lop-oauth-lock && .venv/bin/python -m pytest tests/unit/mcp/test_auth.py \
      -k TestOAuthRefreshLock -q -p no:randomly
5 passed in 1.87s
```

### Gates

All four CI gates clean over the whole tree:

```
$ .venv/bin/python -m flake8 .                                   # clean
$ uvx --from black==26.1.0 black --check .                       # 592 files unchanged
$ uvx isort==5.13.2 --check .                                    # clean
$ .venv/bin/python -m pyright --pythonpath .venv/bin/python .    # 0 errors
```

Full MCP suite:

```
$ .venv/bin/python -m pytest tests/unit/mcp -q
269 passed in 6.88s
```

Full unit suite: `8405 passed`. The 25 failures are **pre-existing and unrelated**
— 24 are `tests/unit/tools/test_eval_tool.py` failing on
`ModuleNotFoundError: No module named 'local_operator'` in its subprocess worker
(an artifact of the symlinked-venv worktree; they reproduce identically on the
untouched `~/local-operator` checkout), and one is an order-dependent flake in
`test_resume_subagents_todos.py` that passes in isolation on both trees. Neither
touches MCP auth.

### Note on the reporter's wedged processes

Several of their `lop` processes are currently wedged in this deadlock. None were
killed. Every test that needed a clean lock state used its own temporary
`LOCAL_OPERATOR_CONFIG_DIR`, and the real-config runs were done deliberately
*alongside* the wedged processes — which is what makes the "all 7 connected"
result meaningful rather than a clean-room artifact.

### DIFF
--- a/local_operator/mcp/auth.py
+++ b/local_operator/mcp/auth.py
@@ -105,22 +105,36 @@ REFRESH_SKEW_S = 60.0
 #: the startup gate defers us, and the breaker bounds retries.
 REFRESH_HTTP_TIMEOUT_S = 10.0
 
-#: Bound on ACQUIRING the cross-process refresh lock, derived from the honest
-#: budget of the critical section it guards: one token POST (capped at
-#: ``REFRESH_HTTP_TIMEOUT_S``) plus a couple of SQLite reads. A peer that holds
-#: the lock legitimately therefore clears it well inside this bound, and
-#: overrunning it means the holder is not working — a leaked lock from a killed
-#: process, or a peer wedged on something that is not our problem. Waiting
-#: longer than the work can possibly take buys nothing and costs a hung connect,
-#: so we give up and let the SDK's own refresh path handle it (see
+#: Bound on ACQUIRING the cross-process refresh lock, derived from the budget of
+#: the critical section it guards: one token POST plus a couple of SQLite reads.
+#: That derivation is only sound because the POST is capped in TOTAL wall time
+#: (see the ``asyncio.timeout`` in :func:`_refresh_oauth_token_locked`) —
+#: ``httpx.Timeout(REFRESH_HTTP_TIMEOUT_S)`` alone does NOT bound a request:
+#: it is per operation, and its read timeout is per socket read, so a dribbling
+#: server measured 140.7s inside a nominal 10s timeout. Keep the two in step: if
+#: the POST's cap is ever raised or removed, this bound stops being honest and a
+#: slow-but-working peer starts getting timed out by its own siblings.
+#: Overrunning this bound therefore means the holder really is not working — a
+#: leaked lock from a killed process, or a peer wedged on something that is not
+#: our problem. Waiting longer than the work can possibly take buys nothing and
+#: costs a hung connect, so we give up and degrade (see
 #: :func:`_oauth_refresh_lock`, which yields False rather than raising).
 LOCK_ACQUIRE_TIMEOUT_S = 15.0
 
-#: Gap between non-blocking lock attempts. Short enough that a released lock is
-#: picked up promptly and that a cancelled acquire abandons the retry loop
-#: within one tick, long enough that waiting out the full bound costs a few
-#: hundred cheap syscalls rather than a hot spin.
+#: FIRST gap between non-blocking lock attempts, and the cancellation
+#: granularity: a cancelled acquire abandons the retry loop within one sleep,
+#: so this also bounds how long an abandoned worker lingers. Small, because the
+#: overwhelmingly common contended case is a peer finishing in a moment and
+#: pickup latency is what the user feels.
 _LOCK_RETRY_SLEEP_S = 0.05
+
+#: Ceiling for the backoff below. A fixed 50 ms gap would cost ~300 wakeups per
+#: contended acquire per server, and six OAuth servers connecting together would
+#: run six worker threads ticking for up to the full bound against a default
+#: executor of 18. Backing off to a quarter second cuts that by most while
+#: leaving the fast case untouched, since a lock still free after a few seconds
+#: is a leaked one we are going to abandon anyway.
+_LOCK_RETRY_SLEEP_MAX_S = 0.25
 
 
 class McpAuthRequiredError(RuntimeError):
@@ -1413,13 +1427,18 @@ def _acquire_locked_fd(path: str, cancelled: threading.Event) -> int | None:
     """
     deadline = time.monotonic() + LOCK_ACQUIRE_TIMEOUT_S
     fd = os.open(path, os.O_CREAT | os.O_RDWR, 0o600)
+    sleep_s = _LOCK_RETRY_SLEEP_S
     try:
         while True:
             if _try_lock_exclusive(fd):
                 return fd
             if cancelled.is_set() or time.monotonic() >= deadline:
                 break
-            time.sleep(_LOCK_RETRY_SLEEP_S)
+            time.sleep(sleep_s)
+            # Gentle geometric backoff: keeps pickup fast for the common
+            # released-in-a-moment case, then stops burning wakeups once the
+            # holder is evidently not finishing soon.
+            sleep_s = min(sleep_s * 1.5, _LOCK_RETRY_SLEEP_MAX_S)
     except BaseException:
         os.close(fd)
         raise
@@ -1484,7 +1503,9 @@ async def _oauth_refresh_lock(server_url: str):
     exchange, and RE-READING the stored token after acquiring it, guarantees
     exactly one process performs the refresh no matter how many sessions start
     at once. The lock file lives next to ``auth.db``, is per server (see
-    :func:`_oauth_refresh_lock_path`), and is only ever flocked, never written.
+    :func:`_oauth_refresh_lock_path`), and carries no state: it is only ever
+    flocked, never read or written for content (on Windows it holds a single
+    padding byte, which ``msvcrt.locking`` requires to have a range to lock).
 
     Yields True when the lock was taken and False when the bounded acquire gave
     up. A False body must still be SAFE to run: the guarantee degrades from
@@ -1553,6 +1574,13 @@ def _close_abandoned_lock_fd(task: asyncio.Task[int | None]) -> None:
     the loop. Without it a lock acquired just as its waiter was cancelled would
     leak the descriptor and, worse, keep the lock held for the process's life.
     """
+    # On both of these paths the WORKER has already closed the fd itself, so
+    # there is nothing here to clean up: a cancelled task never returns one,
+    # and an exception unwinds through :func:`_acquire_locked_fd`'s
+    # ``except BaseException``, which closes before re-raising. That coupling is
+    # what makes the early returns safe — an edit that moves the ``os.open``
+    # into the ``try``, or adds a ``return fd`` ahead of the loop, would leak
+    # both the descriptor AND the lock here with no diagnostic.
     if task.cancelled():
         return
     if task.exception() is not None:
@@ -1707,9 +1735,30 @@ async def _refresh_oauth_token_locked(
         headers["Authorization"] = f"Basic {encoded}"
 
     try:
-        async with httpx.AsyncClient(timeout=httpx.Timeout(REFRESH_HTTP_TIMEOUT_S)) as client:
-            response = await client.post(token_endpoint, data=data, headers=headers)
-    except httpx.HTTPError:
+        # TOTAL wall-clock cap, not just httpx's per-operation timeouts.
+        # ``httpx.Timeout(N)`` sets connect/read/write/pool to N EACH, and the
+        # read timeout applies per socket read rather than to the whole
+        # response — so a server that dribbles the body stays inside every
+        # individual read and takes arbitrarily long. Measured against a server
+        # sending one byte every 2s, this exact request returned HTTP 200 after
+        # 140.7s with the 10s timeout set and never tripping it.
+        #
+        # That matters because this call runs UNDER the cross-process refresh
+        # lock, and ``LOCK_ACQUIRE_TIMEOUT_S`` is derived from the claim that
+        # the section is bounded by ``REFRESH_HTTP_TIMEOUT_S``. Without an
+        # overall cap that claim is false, and a slow-but-honest provider would
+        # systematically push every peer past its acquire bound onto the
+        # unlocked degrade path — which for the in-flight coordinator means the
+        # SDK's own unlocked refresh, i.e. the rotating-token double-spend this
+        # subsystem exists to prevent, reached by timeout instead of by race.
+        # Bounding the POST is what makes the documented budget true.
+        async with asyncio.timeout(REFRESH_HTTP_TIMEOUT_S):
+            async with httpx.AsyncClient(timeout=httpx.Timeout(REFRESH_HTTP_TIMEOUT_S)) as client:
+                response = await client.post(token_endpoint, data=data, headers=headers)
+    except (httpx.HTTPError, TimeoutError):
+        # ``asyncio.timeout`` raises TimeoutError (which ``httpx.HTTPError``
+        # does not cover); a refresh that overran its budget is simply a failed
+        # refresh, handled exactly like a transport error.
         logger.debug("MCP token refresh request failed for %s", server_url, exc_info=True)
         return False
     if response.status_code != 200:
@@ -1991,10 +2040,13 @@ def _make_refresh_coordinating_provider(
     class _RefreshCoordinatingOAuthProvider(OAuthClientProvider):
         # Bound on the re-read/refresh interception: a stuck lock acquisition or
         # a hung endpoint must not park an authenticated request forever. The
-        # file lock is only ever held for one token POST (REFRESH_HTTP_TIMEOUT_S)
-        # plus the SQLite read, so a peer that legitimately holds it clears well
-        # inside this bound; overrunning it means a leaked lock, and degrading to
-        # the SDK's own refresh is strictly better than blocking the request.
+        # file lock is only ever held for one token POST plus the SQLite read,
+        # and that POST is bounded in TOTAL wall time by the ``asyncio.timeout``
+        # in ``_refresh_oauth_token_locked`` (httpx's own per-operation timeout
+        # does not bound a dribbling response), so a peer that legitimately
+        # holds it clears well inside this bound. Overrunning it means a leaked
+        # lock, and degrading to the SDK's own refresh is strictly better than
+        # blocking the request.
         _refresh_coord_server_url = server_url
         _refresh_coord_storage = storage
         _refresh_coord_endpoints = endpoints
--- a/local_operator/mcp/auth.py
+++ b/local_operator/mcp/auth.py
@@ -45,9 +45,11 @@ import contextlib
 import logging
 import os
 import sys
+import threading
 import time
 import weakref
 from dataclasses import dataclass
+from pathlib import Path
 from typing import TYPE_CHECKING, Any, Protocol, runtime_checkable
 from urllib.parse import parse_qs, urlparse
 
@@ -102,6 +104,23 @@ REFRESH_SKEW_S = 60.0
 #: the token POST). A slow authorization server must not park the connect —
 #: the startup gate defers us, and the breaker bounds retries.
 REFRESH_HTTP_TIMEOUT_S = 10.0
+
+#: Bound on ACQUIRING the cross-process refresh lock, derived from the honest
+#: budget of the critical section it guards: one token POST (capped at
+#: ``REFRESH_HTTP_TIMEOUT_S``) plus a couple of SQLite reads. A peer that holds
+#: the lock legitimately therefore clears it well inside this bound, and
+#: overrunning it means the holder is not working — a leaked lock from a killed
+#: process, or a peer wedged on something that is not our problem. Waiting
+#: longer than the work can possibly take buys nothing and costs a hung connect,
+#: so we give up and let the SDK's own refresh path handle it (see
+#: :func:`_oauth_refresh_lock`, which yields False rather than raising).
+LOCK_ACQUIRE_TIMEOUT_S = 15.0
+
+#: Gap between non-blocking lock attempts. Short enough that a released lock is
+#: picked up promptly and that a cancelled acquire abandons the retry loop
+#: within one tick, long enough that waiting out the full bound costs a few
+#: hundred cheap syscalls rather than a hot spin.
+_LOCK_RETRY_SLEEP_S = 0.05
 
 
 class McpAuthRequiredError(RuntimeError):
@@ -1328,34 +1347,84 @@ async def _discover_oauth_endpoints_uncached(
         return None
 
 
-def _lock_exclusive(fd: int) -> None:
+def _try_lock_exclusive(fd: int) -> bool:
+    """ONE non-blocking attempt at the exclusive lock. True when taken.
+
+    Deliberately non-blocking on BOTH platforms. A blocking acquire parks a
+    worker thread inside the kernel on a descriptor the calling coroutine may
+    be about to close, and on macOS/BSD ``os.close()`` of a descriptor with a
+    sibling thread parked in ``flock()`` blocks until that ``flock()`` returns —
+    which, with the lock held by another process, is never. Called from the
+    event-loop thread's ``finally`` that is exactly the whole TUI freezing:
+    no repaint, no input. See :func:`_oauth_refresh_lock` for the full story.
+    Retrying a non-blocking attempt is strictly weaker than blocking and is the
+    only shape that stays cancellable, so do not "simplify" this back into
+    ``fcntl.flock(fd, fcntl.LOCK_EX)``.
+
+    Contention is the expected outcome and returns False; anything else is a
+    real fault (EBADF, EINVAL, an unsupported filesystem) that retrying cannot
+    fix, so it is raised for the caller to degrade on.
+    """
     if os.name == "nt":  # pragma: no cover - platform specific
         import errno as _errno
         import msvcrt
-        import time as _time
 
-        # ``LK_LOCK`` blocks but gives up after ~10 s (it retries once per
-        # second, ten times, then raises OSError) — while a peer legitimately
-        # holds the lock for up to REFRESH_HTTP_TIMEOUT_S of network time.
-        # Retry in a loop to match the POSIX flock's indefinite-block
-        # semantics; the loop is bounded generously rather than forever so a
-        # leaked lock (killed process) cannot park a connect eternally.
-        # Contention surfaces as EDEADLOCK/EACCES; anything else (EBADF,
-        # EINVAL) is a real fault that retrying cannot fix — raising it
-        # immediately beats hot-spinning the worker thread for the bound.
-        deadline = _time.monotonic() + 60.0
-        while True:
-            try:
-                msvcrt.locking(fd, msvcrt.LK_LOCK, 1)
-                return
-            except OSError as lock_err:
-                contended = lock_err.errno in (_errno.EDEADLOCK, _errno.EACCES)
-                if not contended or _time.monotonic() >= deadline:
-                    raise
+        try:
+            # ``msvcrt.locking`` locks a byte RANGE, so the file needs at least
+            # one byte to lock — an empty lock file would fail with EINVAL
+            # forever. Mirrors ``session_lease``'s handling of the same API.
+            if os.fstat(fd).st_size == 0:
+                os.write(fd, b"\0")
+            os.lseek(fd, 0, os.SEEK_SET)
+            msvcrt.locking(fd, msvcrt.LK_NBLCK, 1)
+            return True
+        except OSError as lock_err:
+            if lock_err.errno in (_errno.EDEADLOCK, _errno.EACCES, _errno.EAGAIN):
+                return False
+            raise
     else:
+        import errno as _errno
         import fcntl
 
-        fcntl.flock(fd, fcntl.LOCK_EX)
+        try:
+            fcntl.flock(fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
+            return True
+        except OSError as lock_err:
+            if lock_err.errno in (_errno.EAGAIN, _errno.EACCES, _errno.EWOULDBLOCK):
+                return False
+            raise
+
+
+def _acquire_locked_fd(path: str, cancelled: threading.Event) -> int | None:
+    """Open ``path`` and take the exclusive lock on it, or give up. Worker-side.
+
+    Runs entirely on a worker thread and OWNS the descriptor for its whole life:
+    it opens the fd, retries the non-blocking acquire until the deadline, and on
+    any outcome other than success closes the fd ITSELF, in this same thread,
+    after the last lock syscall has returned. That ownership rule is the fix for
+    the deadlock — the event loop never closes a descriptor another thread might
+    still be inside a lock call on, because by construction no such moment
+    exists. Returns the locked fd on success (ownership passes to the caller,
+    which must unlock and close it) or ``None`` on timeout/cancellation.
+
+    ``cancelled`` is set by the coroutine when its await is cancelled, so an
+    abandoned acquire abandons the retry loop within one ``_LOCK_RETRY_SLEEP_S``
+    tick instead of holding a thread for the full bound.
+    """
+    deadline = time.monotonic() + LOCK_ACQUIRE_TIMEOUT_S
+    fd = os.open(path, os.O_CREAT | os.O_RDWR, 0o600)
+    try:
+        while True:
+            if _try_lock_exclusive(fd):
+                return fd
+            if cancelled.is_set() or time.monotonic() >= deadline:
+                break
+            time.sleep(_LOCK_RETRY_SLEEP_S)
+    except BaseException:
+        os.close(fd)
+        raise
+    os.close(fd)
+    return None
 
 
 def _unlock(fd: int) -> None:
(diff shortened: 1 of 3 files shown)


damianvtran:
### Agent review — round 1

Scope: e641a8884f7b4696cdb2a07ad3f9bd2fba3d1d00..25fd1a22e0e4908526fa3754fd7e342c78dd7f8f

Reviewed the diff in `local_operator/mcp/auth.py`, `tests/unit/mcp/test_auth.py`, `pyproject.toml`. Re-verified the PR's central claims by execution rather than reading, in a throwaway `LOCAL_OPERATOR_CONFIG_DIR`; no `lop` process of yours was touched.

**Independently reproduced and confirmed:**

- The five new tests genuinely fail on `origin/main` **for the right reason**. In my own worktree at `0ed1051e` with only the test file copied over, `test_cancelling_mid_acquire_unwinds_without_blocking_the_loop` deadlocked for its full 90s bound (`subprocess.TimeoutExpired`), the other four failed on the missing symbols/`assert None is True`. `5 failed in 121.89s`. On the branch: `5 passed in 2.53s`.
- The ownership rule holds. `_acquire_locked_fd` (auth.py:1398) is the only code that touches the fd while a lock syscall could be outstanding, it closes on every non-success path on its own thread (:1141-1145), and the coroutine's cancel arm (:1522-1529) touches the descriptor not at all. I traced every path; there is no window where the loop closes an fd a thread is parked on. This is the correct shape.
- The cancelled-but-won case really does release the lock. `_close_abandoned_lock_fd` has no unit coverage (line 1556 is uncovered), so I probed it directly: cancel in the window where the worker wins anyway, then ask a **foreign** process for the lock. It reports `FREE`. The done-callback works.
- Gates clean on the branch: flake8 0, `black==26.1.0` 592 files unchanged, `isort==5.13.2` clean, pyright `0 errors, 0 warnings`.
- The "pre-existing `eval_tool` failures" claim is **true, and I checked it rather than accepting it**: on untouched `origin/main` with no test changes, `tests/unit/tools/test_eval_tool.py` gives `24 failed, 23 passed`. Unrelated to this change.
- Per-server key is stable across processes (SHA-256, not `hash()`; verified invariant under `PYTHONHASHSEED=random`), lands in `config_dir()`, and cannot path-inject: a hex digest has no `/` or `..`. Leaving the legacy global file in place is the right call for the stated reason — unlinking a file a live older peer holds an flock on would silently drop that peer's exclusion.
- **Mutual exclusion is not weakened in the locked case, and the degrade is reasoned correctly.** I checked each of the four call sites against the "does this SPEND the rotating refresh token?" question. auth.py:1808 returns early when unlocked (never spends). auth.py:2014 does the re-read then returns before the exchange (:2026). The two adoption sites (:2103, :2284) are pure reads that only rewrite our own header. No path performs a token exchange without the lock. That is the property that matters most here and it is upheld.

The audit of other `flock`/`os.close` pairings was actually done and is correct: `session_lease.py::_stale_recovery_right` is already `LOCK_EX|LOCK_NB` so nothing parks; `group_reaper.py::_ledger_lock` is a blocking `flock` but I confirmed it is a synchronous `@contextmanager` with its close in the same thread's `finally` — no second thread, so not this defect class. Nothing else in the tree pairs a blocking lock with a cross-thread close.

---

#### F1 — major — `LOCK_ACQUIRE_TIMEOUT_S` is derived from a bound `httpx` does not actually enforce

`local_operator/mcp/auth.py:108-116`. The 15.0s bound is justified by an explicit budget: "one token POST (capped at `REFRESH_HTTP_TIMEOUT_S = 10.0`) plus a couple of SQLite reads", therefore "a peer that holds the lock legitimately clears well inside this bound, and overrunning it means the holder is not working".

`httpx.Timeout(10.0)` does not cap a request's total wall time. It sets connect/read/write/pool to 10s each, and the read timeout is **per socket read**, not for the whole response. A server that dribbles the body stays inside every individual 10s read and takes arbitrarily long. I measured it against a local server that sends one byte every 2s:

```
POST wall time = 82.1s  (REFRESH_HTTP_TIMEOUT_S=10.0, LOCK_ACQUIRE_TIMEOUT_S=15.0)
EXCEEDS 15s LOCK BOUND: True
status 200
```

So the premise is false: a peer holding the lock **can** legitimately still be working at 15s. Why it matters — the degrade path is safe against double-spend (F1 does not reintroduce that), but a slow-but-honest provider now systematically pushes every peer onto the unlocked path, which is the opposite of when you want to degrade. It converts a rare leaked-lock fallback into a routine one precisely against a flaky provider. Worse for `_coordinate_inflight_refresh` (:2014): degrading there hands the request to the SDK's own **unlocked** `_refresh_token`, and with several sessions crossing the 8h boundary against a slow Notion that is the double-spend the coordinator exists to prevent — reached by timeout rather than by race.

Fix: either bound the POST's total time so the comment's budget is real (wrap `_refresh_oauth_token_locked`'s request in `asyncio.timeout(REFRESH_HTTP_TIMEOUT_S)`, or set `httpx.Timeout(..., read=...)` alongside an overall cap), or raise `LOCK_ACQUIRE_TIMEOUT_S` above the true worst case and correct the comment's derivation. The first is better: it makes the stated invariant true rather than restating a larger guess. Note the sibling comment at :1993-1997 makes the same "held for one token POST" claim and inherits the same error.

#### F2 — minor — the degrade branches that change auth behaviour have no test

Coverage over `tests/unit/mcp` leaves auth.py:1816 (`ensure_mcp_oauth_fresh`'s `return endpoints` when unlocked) and :2026 (`_coordinate_inflight_refresh`'s unlocked early return) uncovered. `TestOAuthRefreshLock` proves `_oauth_refresh_lock` *yields* `False` at the deadline, but nothing asserts that the two callers then behave correctly — that no exchange is attempted, and specifically that `_coordinate_inflight_refresh` still performs its `_resync_from_store` **before** returning. That ordering (:2019 re-read, then :2026 return) is the entire reason the degrade is safe, and it is exactly the kind of line a future refactor moves. A test monkeypatching `_oauth_refresh_lock` to a `False`-yielding stub and asserting `_refresh_oauth_token_locked` is never called while the resync still happened would pin it. Same for :1556 (`_close_abandoned_lock_fd`) — I had to probe it by hand because no test reaches it, and its failure mode is a lock held for the process's life.

#### F3 — minor — `_close_abandoned_lock_fd` swallows a `BaseException` outcome and leaks the fd

`local_operator/mcp/auth.py:1558-1559`. The callback returns early when `task.exception() is not None`. But `_acquire_locked_fd`'s `except BaseException` arm (:1141) closes the fd only for exceptions raised *inside* the `while` loop; `Task.exception()` also raises `CancelledError` rather than returning it, and the guard at :1556 covers that case first, so the code is correct today — the concern is narrower: on the `task.exception()` path the fd was already closed by the worker, which is fine, but nothing documents that coupling. If someone later moves the `os.open` inside the `try` or adds an early `return fd` before the loop, this callback silently leaks a descriptor *and* the lock with no diagnostic. One line of comment stating "the worker has already closed the fd on both of these paths — see `_acquire_locked_fd`'s `except BaseException`" would make the invariant survive an edit. Not a defect in the current code.

#### F4 — minor — the docstring's "only ever flocked, never written" is now false on Windows

`local_operator/mcp/auth.py:1487` still says the lock file "is only ever flocked, never written", but the new `_try_lock_exclusive` Windows branch writes a `\0` byte to it (:1377) because `msvcrt.locking` needs a non-empty range. Harmless behaviour, stale claim. Add "(one padding byte on Windows, which `msvcrt.locking` requires)".

#### F5 — nit — `_LOCK_RETRY_SLEEP_S` gives a fixed ~300-attempt spin under sustained contention

`local_operator/mcp/auth.py:123`, with the loop at :1135-1140. 15.0s / 0.05s is ~300 wakeups per contended acquire per server, and with six OAuth servers connecting together that is six worker threads each ticking for up to 15s against a default executor of 18. It is cheap and the comment is honest about the trade ("a few hundred cheap syscalls rather than a hot spin"), so this is not worth a round. Mentioning it only because a small backoff (50ms growing to ~250ms) would cut it by most and cost nothing in pickup latency for the common uncontended-after-a-moment case.

#### F6 — nit — one flaky timing test in the MCP suite, pre-existing

`tests/unit/mcp/test_manager.py::TestStdioStopIsEventDriven::test_prompt_child_costs_no_polling_tick` failed for me in a full-directory run (`teardown took 0.448s`, asserts `< 0.09`) and passed in isolation. Load-sensitive, unrelated to this change, and CI is green. Noting it so it is not mistaken for a regression from this PR; not something to fix here.

---

**Comment quality** is genuinely good and meets the bar the task set: `_try_lock_exclusive`'s docstring names the macOS `close()`-on-parked-`flock()` mechanism and says outright "do not 'simplify' this back into `fcntl.flock(fd, fcntl.LOCK_EX)`", and `_oauth_refresh_lock`'s two numbered rules make the ownership invariant hard to undo by accident. A future reader cannot regress this without ignoring an explicit instruction. The only comment defects are F1's false premise and F4's stale claim.

**CI**: all checks green on the current head except `test (3.13)` on one of two runs still pending at review time; the other run's `test (3.13)` passed.

### Verdict: **changes requested** — 1 major (F1), 3 minor, 2 nits.

The concurrency fix itself is correct and I could not break it: the deadlock is genuinely gone, the ownership rule is sound, the exclusion guarantee is not weakened in the locked case, and the degrade paths never spend a rotating refresh token. Version bump to `0.43.2` (patch) is right per AGENTS.md — a bug fix with no new capability.

F1 is the one finding I would not wave through on a rotating-token auth path. It is not a deadlock and not a correctness bug in the lock, but the constant's stated derivation is measurably wrong, and the consequence lands on the `_coordinate_inflight_refresh` degrade where an unlocked SDK refresh is exactly the double-spend this subsystem exists to prevent. Fixing it is small: bound the POST, or restate the bound honestly. The round is not terminal.

damianvtran:

Implementer: `coder` subagent — anthropic/claude-opus-5
Scope: `25fd1a22e0e4908526fa3754fd7e342c78dd7f8f..a9f2a3af41c414863d585cf6eac6e0db39ad067f`
Remediation commit: `a9f2a3af41c414863d585cf6eac6e0db39ad067f`

Thanks for the execution-based review — re-deriving the tests' failure mode on `origin/main` and probing `_close_abandoned_lock_fd` by hand caught the two things reading alone would not have.

---

#### F1 (major) — **fixed** in `a9f2a3af`

Confirmed by my own measurement before changing anything, and it is worse than reported. Against a local server dribbling one byte every 2s, the real refresh POST returned **HTTP 200 after 140.7s** with `REFRESH_HTTP_TIMEOUT_S = 10.0` set and no individual read ever exceeding it:

```
$ .venv/bin/python /tmp/dribble_probe.py raw
REFRESH_HTTP_TIMEOUT_S = 10.0
server dribbles 40+ bytes, one every 2.0s

POST wall time = 140.7s  (status 200)
EXCEEDS REFRESH_HTTP_TIMEOUT_S: True
EXCEEDS LOCK_ACQUIRE_TIMEOUT_S: True
```

Took your preferred option — bound the POST so the documented budget becomes true, rather than restating a larger guess. `_refresh_oauth_token_locked`'s request now runs under `asyncio.timeout(REFRESH_HTTP_TIMEOUT_S)`, with `TimeoutError` joining `httpx.HTTPError` in the catch (an overrun is just a failed refresh, handled identically). Re-measured through the **real function**, timing only the call and not the test scaffolding:

```
$ .venv/bin/python /tmp/dribble_probe.py fixed
POST wall time = 10.0s  (refreshed=False)
REFRESH_HTTP_TIMEOUT_S = 10.0, LOCK_ACQUIRE_TIMEOUT_S = 15.0
EXCEEDS REFRESH_HTTP_TIMEOUT_S: True     # by 0.04s of teardown
EXCEEDS LOCK_ACQUIRE_TIMEOUT_S: False
```

140.7s → 10.0s. The critical section is now genuinely bounded by `REFRESH_HTTP_TIMEOUT_S`, so `LOCK_ACQUIRE_TIMEOUT_S = 15.0` is derived from something real and a slow-but-working peer is no longer timed out by its siblings.

Both comments carrying the false premise are corrected. The constant's comment (`:108`) now names the mechanism and the coupling explicitly, so the two cannot drift apart silently:

> `httpx.Timeout(REFRESH_HTTP_TIMEOUT_S)` alone does NOT bound a request: it is per operation, and its read timeout is per socket read, so a dribbling server measured 140.7s inside a nominal 10s timeout. Keep the two in step: if the POST's cap is ever raised or removed, this bound stops being honest and a slow-but-working peer starts getting timed out by its own siblings.

The sibling comment on the coordinator class (`~:1993`) no longer claims the lock is "only ever held for one token POST (REFRESH_HTTP_TIMEOUT_S)" without qualification; it now points at the `asyncio.timeout` that makes it so.

#### F2 (minor) — **fixed** in `a9f2a3af`

Added `TestRefreshLockDegradePaths` and `TestCloseAbandonedLockFd` (5 tests):

- **`test_ensure_fresh_skips_the_exchange_when_the_lock_is_not_taken`** — `_oauth_refresh_lock` stubbed to yield `False`; asserts `_refresh_oauth_token_locked` is never called **and** that the endpoints are still returned, so the connect proceeds rather than blocking.
- **`test_ensure_fresh_does_spend_the_token_when_the_lock_is_taken`** — the control, so the skip above is attributable to the degrade rather than to some other gate.
- **`test_inflight_coordination_resyncs_before_it_degrades`** — exactly the ordering you named: a sibling rotates the token in the store, and the unlocked coordinator must still adopt `a-peer` while attempting no exchange.
- **`test_a_won_but_abandoned_lock_is_released`** — covers `_close_abandoned_lock_fd`. Uses a **foreign process** to witness the release, since flock is per open-file-description and a same-process probe would report `FREE` even on a leak.
- **`test_a_cancelled_or_failed_acquire_is_a_no_op`** — both early returns.

Because a test that passes on correct code proves little, I mutation-checked the two that carry the safety argument. Reordering the resync past the degrade return:

```
FAILED ...::test_inflight_coordination_resyncs_before_it_degrades - assert 'a-old' == 'a-peer'
```

Removing the unlocked guard so the exchange runs anyway:

```
FAILED ...::test_ensure_fresh_skips_the_exchange_when_the_lock_is_not_taken - assert 1 == 0
```

Both fail on the mutant and pass on the real code, so they pin the behaviour rather than merely executing it.

#### F3 (minor) — **fixed** in `a9f2a3af`

Comment added at the callback's early returns naming the coupling:

> On both of these paths the WORKER has already closed the fd itself, so there is nothing here to clean up: a cancelled task never returns one, and an exception unwinds through `_acquire_locked_fd`'s `except BaseException`, which closes before re-raising. That coupling is what makes the early returns safe — an edit that moves the `os.open` into the `try`, or adds a `return fd` ahead of the loop, would leak both the descriptor AND the lock here with no diagnostic.

#### F4 (minor) — **fixed** in `a9f2a3af`

The docstring now reads: "carries no state: it is only ever flocked, never read or written for content (on Windows it holds a single padding byte, which `msvcrt.locking` requires to have a range to lock)."

#### F5 (nit) — **adopted** in `a9f2a3af`

Took the suggestion; it costs nothing. The gap now backs off geometrically from 50 ms by ×1.5 to a 250 ms ceiling, cutting a contended acquire from ~300 wakeups to a few dozen. `_LOCK_RETRY_SLEEP_S` stays 50 ms as the first gap because it also sets cancellation granularity (an abandoned worker lingers at most one sleep), and the ceiling is documented against the case that motivated it: six OAuth servers connecting together against a default executor of 18.

#### F6 (nit) — **acknowledged, no action**

Agreed and left alone. I hit the same class of load-sensitive flake from the other direction: `tests/unit/mobile/test_tui_bridge.py` failed in my full-suite run and passes in isolation (`5 passed`), as did `test_band_panels.py::test_the_inset_is_never_what_tips_a_long_subagent_list_over` — which failed on parameter `[20-10]` locally but `[16-6]` in CI, the signature of a timing flake rather than a defect. None are touched by this PR.

---

### Verification on the new head

Deadlock fix re-verified after the backoff change, since F5 touched the retry loop:

```
$ .venv/bin/python /tmp/lock_probe.py
RESULT: OK — cancellation unwound in 0.000s, loop never blocked

$ .venv/bin/python /tmp/lock_bound_probe.py
server B acquire: locked=True in 0.00s
server A acquire: locked=False after 15.25s
B uncontended by A's held lock: PASS
A bounded at ~15s and degraded: PASS
```

(The 0.25s overshoot is the new backoff ceiling landing on the deadline — within the bound's intent.)

Gates, whole tree, as CI runs them:

```
$ .venv/bin/python -m flake8 .                                   # clean
$ uvx --from black==26.1.0 black --check .                       # 592 files unchanged
$ uvx isort==5.13.2 --check .                                    # clean
$ .venv/bin/python -m pyright --pythonpath .venv/bin/python .    # 0 errors, 0 warnings
$ .venv/bin/python -m pytest tests/unit/mcp -q                   # 274 passed (269 + 5 new)
```

Full unit suite: `8411 passed`. The 24 `eval_tool` failures are the pre-existing symlinked-venv artifact you independently confirmed on `origin/main`; the one `test_tui_bridge` failure is the flake noted under F6 and passes in isolation.

No production behaviour changed beyond F1's bound and F5's backoff — the lock's ownership rule, the per-server key, and the degrade semantics you verified call-site by call-site are untouched.

damianvtran:
### Agent review — round 2

Scope: 25fd1a22e0e4908526fa3754fd7e342c78dd7f8f..a9f2a3af41c414863d585cf6eac6e0db39ad067f

Verified the remediation by execution against the new head, in a throwaway `LOCAL_OPERATOR_CONFIG_DIR`; no `lop` process of yours was touched. Scoped to what changed — the properties I confirmed call-site by call-site in round 1 (ownership rule, per-server key, mutual exclusion at all four sites, the flock audit) are re-checked only where this diff could have disturbed them, and it does not: production changes are confined to the POST's cap, the retry gap, and comments.

#### F1 (major) — **verified fixed**

I did not take the measurement on trust; I re-derived it through the real function against my own dribbling server (one byte every 2s, so every individual read stays inside the 10s read timeout):

```
REFRESH_HTTP_TIMEOUT_S = 10.0
LOCK_ACQUIRE_TIMEOUT_S = 15.0

[1] BOUND: returned=False in 10.00s  under LOCK_ACQUIRE_TIMEOUT_S: True
[2] STORE after overrun: access='a-old' refresh='r-old' (untouched: True)
[3] EXTERNAL CANCEL: propagated as CancelledError in 0.001s — NOT converted to TimeoutError  (correct)
```

The critical section is genuinely bounded, so `LOCK_ACQUIRE_TIMEOUT_S = 15.0` is now derived from something true rather than from a timeout httpx does not enforce. Taking the four sub-questions in turn:

- **Overrun leaves the stored token untouched.** `[2]` above. The `except (httpx.HTTPError, TimeoutError)` arm returns `False` before `response` is ever bound, so the whole persist block (RFC 6749 §6 carry-forward, `set_tokens`) is unreachable on the timeout path. A refresh that overran is indistinguishable from a transport failure, which is the correct semantics — the old refresh token remains the freshest thing in storage and stays spendable.
- **No double-spend.** The timeout fires *inside* the lock, and the two callers treat `False` exactly as they already treated a failed exchange: `ensure_mcp_oauth_fresh` ignores the return and proceeds to connect (auth.py:1870), and the coordinator only resyncs `if refreshed:` (auth.py:2109-2110). Nothing retries and nothing falls through to the SDK's unlocked `_refresh_token` while still holding a token it believes fresh. Crucially the fix *removes* the double-spend route round 1 identified: the coordinator no longer gets pushed onto the unlocked path by a merely slow provider.
- **No escalation to a browser grant.** A `False` return cannot open a login tab. The grant path is gated on `interactive` (auth.py:891), so a background connect raises `McpAuthRequiredError` regardless of why the refresh failed. Timeout and transport error reach that decision identically.
- **`asyncio.timeout` is the right tool here, and its cancellation cannot be confused with external cancellation.** This was the part worth checking rather than reading, since the module distinguishes the two carefully (`McpLoginCancelledError`, the `post_writer` re-voicing at :770/:922). `asyncio.timeout` converts *its own* internal cancellation to `TimeoutError` only when its own deadline expired; an outer cancel while the task sits inside the scope is re-raised as `CancelledError`. `[3]` confirms it empirically: cancelling the surrounding task 1.5s into a 10s scope propagated `CancelledError` in 0.001s and did **not** get swallowed into `return False`. That matters because `/resume` cancels connects mid-flight, and a swallowed cancel here would be reported as a failed refresh. The catch is `TimeoutError`, not `BaseException` or `asyncio.CancelledError`, so the distinction is preserved by construction. I also checked for an enclosing timeout scope that an inner one could misattribute: the connect path reaches this through `manager.py:1583` with no surrounding `wait_for`/`timeout`, so there is no nesting hazard. Requires-python is `>=3.12`, so `asyncio.timeout` is available.
- **Both false-premise comments corrected.** The constant's comment (:105-118) now names the mechanism, cites the 140.7s measurement, and states the coupling as a maintenance rule; the coordinator's sibling comment (:2042-2049) no longer claims the section is bounded by `REFRESH_HTTP_TIMEOUT_S` without qualification and points at the `asyncio.timeout` that makes it so. The two can no longer drift apart silently, which was the actual defect.

#### F2 (minor) — **verified fixed; tests assert the real property**

Five tests, all passing (`5 passed in 3.36s`). I ran my own mutants rather than relying on the reported ones:

```
### MUTATION A: drop the resync from the degrade path
  - a-peer
  + a-old
1 failed
### MUTATION B: _close_abandoned_lock_fd returns without unlocking
assert 'HELD' == 'FREE'
  - FREE
  + HELD
1 failed
```

`test_inflight_coordination_resyncs_before_it_degrades` genuinely pins the resync-before-degrade ordering: removing the re-read leaves the coordinator on `a-old` and the test fails on exactly that. `test_a_won_but_abandoned_lock_is_released` uses a **foreign subprocess** as the witness, which is the only honest probe — my mutant leaks the lock and the subprocess reports `HELD`, whereas a same-process check would have reported `FREE` and passed. Both kill their mutant and pass on the real code.

#### F3, F4 (minor) — **verified fixed**

F3's comment states the coupling correctly and names the two edits that would break it. F4's docstring is now accurate, and I checked the claim rather than the prose: auth.py:1391 does `os.write(fd, b"\0")` before `msvcrt.locking(fd, LK_NBLCK, 1)`, so "a single padding byte, which `msvcrt.locking` requires to have a range to lock" is exactly right.

#### F5 (nit) — **adopted; bound and cancellation both intact**

Measured rather than reasoned:

```
[1] backoff: first=0.05s ceiling=0.25s -> 63 wakeups to cover 15.0s (fixed-50ms would be 300)
[2] BOUND under foreign holder: fd=None after 15.07s (overshoot 0.071s, within one ceiling sleep)
[3] CANCEL after backoff maxed: abandoned in 0.190s (<= one ceiling sleep), fd=None
[4] attempts in 15.06s = 63 (a hot spin would be thousands)
```

The bound still terminates at the deadline (overshoot is bounded by one ceiling sleep, by construction of the `break`-before-`sleep` ordering), the sequence is monotonic and clamped, cancellation remains prompt even once the gap has grown to 250 ms, and there is no spin. `_LOCK_RETRY_SLEEP_S` correctly stays the first gap so cancellation granularity is unchanged in the common case.

#### F6 (nit) — acknowledged, correctly no action.

---

### Verification on the new head

The deadlock fix re-verified after the backoff and timeout changes, since both touch this path:

```
cancel unwound in 0.0002s (loop-blocking would be seconds)
heartbeat kept ticking across the cancel: 19 -> 25 (+6)
RESULT: OK — loop never blocked
```

Gates, whole tree, exactly as CI runs them:

```
.venv/bin/python -m flake8 .                                   # clean
uvx --from black==26.1.0 black --check .                       # 592 files unchanged
uvx isort==5.13.2 --check .                                    # clean
.venv/bin/python -m pyright --pythonpath .venv/bin/python .    # 0 errors, 0 warnings
.venv/bin/python -m pytest tests/unit/mcp -q                   # 274 passed
```

Full unit suite: `8189 passed, 15 skipped, 1 xfailed`. CI on `a9f2a3af` is green across the board, including **both** `test (3.13)` runs — the earlier 3.13 failure on `test_band_panels.py` was a flake, not this PR, and it did not recur.

No new findings. Nothing in this diff weakens what round 1 verified: the fd-ownership rule, the per-server key, and the degrade semantics are untouched, and F1's fix strictly reduces the conditions under which the degrade path is reached at all.

**Verdict: clean.** No blocker and no major findings remain. This round is TERMINAL — there are no outstanding minors or nits, so nothing here warrants another round. Ready to merge.

PR #388, opened 2026-08-29
fix(tui): make double-click and triple-click select in the composer

damianvtran:
## Summary

Reported: *"I can't properly copy via ctrl/cmd+C on highlighted text in the composer. Copying from the composer, unlike the conversation history, requires explicit cmd/ctrl+C but that doesn't seem to work."*

The report understates it. The gesture did not merely fail to copy — **it destroyed the draft**. Double-clicking a word in the composer produced no selection the widget could copy, so the Ctrl+C that followed found no live range and fell through to the interrupt rung, whose first tap clears the prompt. The user asked for a copy and lost their text.

Ctrl+C itself was never broken. A drag then Ctrl+C works, and so does shift+arrow then Ctrl+C — both verified below under a real pty. What was broken is **the way the user made the highlight**: the click chain.

## Root cause

Entirely in Textual 8.2.8, and worth stating because the widget's own code does not show it.

`Widget._on_click` (`textual/widget.py`) reacts to a click chain by calling `text_select_all()` on chain 2, and on the *container* at chain 3. Those write a **screen** selection — `Screen.selections`, the read-only band the transcript uses. For a `TextArea` that entry is inert twice over:

- the entry it writes is `Selection(start=None, end=None)`, and `Widget.get_selection` returns text only for a `Text`/`Content` visual, so a `TextArea` contributes nothing to `Screen.get_selected_text()`;
- **nothing paints.** The band is applied by `Content.render_strips`, and this widget renders its own lines, so the frame after a double-click is byte-identical to the frame before it.

Meanwhile `TextArea.selection` — the reactive that both paints the composer's highlight *and* backs `selected_text` — is left collapsed by the mouse-down, and no base-class handler ever widens it: `TextArea` binds word/line selection to `f5`/`f6`/`f7` only and has **no click-chain handling whatsoever**.

So the gesture every user reaches for first in an input box selected nothing, showed nothing, and then ate the draft.

Chain 3 was worse than chain 2, because it escalated to the container. Captured on a real pty, the clipboard after a triple-click held the frame's furniture:

```
'\ndraft cleared — ↑ to recover\n\n\n\n\n\n◆ test/model › ⌂ /private/tmp/lop-composer-copy   !\n❯\n\n\ncopied 2 lines'
```

The composer's own line is the one thing not in it.

### Why the existing tests were green

[...]

PR #378, opened 2026-08-28
feat(tui): inline ghost text for slash completions

damianvtran:
# PR Description

## Summary of Changes

Inline ghost text in the composer (issue #368, part 1 of 2): the remainder of
the highlighted picker row is rendered dimmed ahead of the caret, so the user
can see exactly what Tab will produce without mapping a highlighted row back
onto their own buffer. This is the case the issue asks for:

```
/mcp login n␣otion          ← "otion" dimmed, inline
```

It also carries a **data-loss fix** on the destructive-argument gate, found by
driving the real key path (credit to the PR 2 / `/mcp` verbs investigation).

### The invariant

The whole feature rests on one promise: **the dimmed cells are exactly the
characters Tab commits** — `buffer_before + ghost == buffer_after_tab`, always.

That is not free. All three completion sites replace a **span**
(`text[:start] + name + text[end:]`), which only *looks* like an append when
the row extends what was typed. For a fuzzy match it rewrites characters
already on screen: `/lg` + Tab yields `/login `, which is not `/lg` plus
anything.

So the ghost and Tab now derive from **one shared pure function**,
`completion_for()` in `command_picker.py`, which the three `Editor` completion
methods delegate their string arithmetic to (side effects stay where they
were). The invariant is true by construction rather than by two builders
agreeing. `ghost_for()` then shows a ghost **only when the completion is a pure
append**; a fuzzy match previews nothing, because no string appended at the
caret could honestly describe that edit. The check is deliberately
case-sensitive: `/MCP lo` + Tab inserts the registry's own casing, so a
case-insensitive check would paint characters Tab does not produce.

### Native mechanism, not a bespoke render pass

Textual 8.2.8 already has ghost text: the `suggestion` reactive, the
`text-area--suggestion` component (already styled `$lo-dim` in
`local_operator.tcss`), and insertion at the caret in `_render_line`. This PR
uses it. A hand-rolled paint would duplicate the framework and could not be
ordered correctly against `_paint_markers`/`_paint_slash`, which post-process a
finished `Strip`.

### Three gates

The native path has two real defects for this widget, both neutralised by
gates rather than new render code:

1. **Caret at the end of the line, no selection.** The ghost adds cells the
   *document* lacks, while `_slash_cells`/`_marker_cells` compute x-ranges from
   *document* columns — so any highlighted run at or after the caret slides
   against the rendered strip. Proven: with the gate bypassed, `/mc` + ghost
[...]

PR #365, opened 2026-08-28
feat(compaction): speculative compaction advisor behind a beta flag

damianvtran:
# PR Description

## Summary of Changes

Adds a **speculative compaction advisor** behind a beta flag
(`values.compaction.advisor_enabled`, default `false`), plus a task-aware
preserve window that applies to every session whether or not the beta is on.

The problem is not that compaction fires too late. It is that **the cut point
is recency-shaped while a session is task-shaped**, and that mismatch is what
makes an earlier trigger unsafe rather than helpful.

The advisor-triggered **pass also runs off the turn** and applies at the next
safe boundary, so firing earlier and more often does not buy the user a
summarization call in the middle of their work. The ceiling pass and manual
`/compact` stay synchronous.

---

## Three corrections to the original premise

The design brief for this work carried three numbers that are wrong, and a
reviewer should not have to rediscover that. Re-measured against the real
session transcript (`~/.local-operator/sessions/bda7b76d34e0`, 8148 records):

| Claim in the premise | Measured |
| --- | --- |
| 1.44B prompt-side context tokens | **1.118B** |
| 509M tokens above the 300k line | **227M** |
| 29.4% of context above 500k | **12.6%** |

**And the money is not what it looks like.** 92.9% of the prompt side is
`cache_read` (1.040B of 1.118B; `cache_write` 45.5M, fresh ~34.3M). The 227M
excess over 300k is therefore **cached-read money: roughly $341, not ~$3.4k**
as it would be at fresh-input prices. That is a real number and worth
addressing, but it is an order of magnitude smaller than a fresh-token reading
implies, and it changes what this feature has to justify: the advisor is worth
shipping for **task-severance**, not for the token bill.

---

## The actual problem: severance

At each of the seven real compaction passes, the active-task span (tokens since
the last genuine user turn) was:

```
0.3k, 46.9k, 48.8k, 30.0k, 19.8k, 123.4k, 49.1k     p50 32k, p90 99k
```

against `keep_recent_tokens = 20000`. **Five of seven passes cut inside a live
task**, summarizing away the first half of work the agent was still executing.

23 of the session's 69 user turns are continuations ("Continue", "Quota is
back..."). Those are genuine user turns but *not* task boundaries, which is why
a purely local "preserve everything since the last user turn" rule anchors to
the wrong place on a third of the sample — and why a model judgement earns its
place here.

---

### STEP 0 — task-aware preserve window (local, no model call)

[...]

PR #411, opened 2026-08-29
fix(tui): keep the ghost caret at the insertion point

damianvtran:
# PR Description

## Summary of Changes

With `/resume` fully typed, the composer painted its block caret over the `e` of
`resume`. Reported from the field with screenshots: the user reads a block ON a
committed character as the vi-normal-mode / overwrite idiom — "my next keystroke
replaces this `e`" — even though `resume` is fully committed text and the caret
genuinely belongs after it. The composer contradicted itself between its
ghosted and un-ghosted states: a buffer with no completion on offer put the
caret after the text, where it belongs, while every buffer carrying a ghost
pulled it one cell back.

The cause is deliberate. #378 added `Editor._paint_ghost_caret`, which moved the
block one cell LEFT, off the ghost's first character and onto the last character
typed. It was solving a real problem — every command's completion ends in a
trailing space, so a fully typed command has the one-character ghost `' '`,
entirely under the block and drawing zero dim pixels — but it solved it by
moving the caret off the insertion point.

### Why the caret goes back, and what the alternatives actually are

Every autosuggestion implementation inspected keeps the caret at the TRUE
insertion point, which is exactly where the ghost starts; with a block cursor
the block therefore lands on the ghost's first character.

| system | where the caret sits |
|---|---|
| fish | suggestion rendered after the cursor, dim `fish_color_autosuggestion` |
| zsh-autosuggestions | `CURSOR == $#BUFFER`, ghost lives in `POSTDISPLAY` |
| prompt_toolkit `AppendAutoSuggestion` | ghost appended as fragments, cursor stays at `cursor_position` |
| Textual `Input.render_line` | builds `value + suggestion[len:]`, then applies the cursor style ON TOP of the first ghost cell |
| VS Code / Copilot, JetBrains | ghost after a bar caret at the model position |

There is **no precedent, terminal or GUI, for retreating the caret one cell**.
Browser URL bars use the other coherent convention — insert the completion and
SELECT it, which genuinely does mean "the next keystroke replaces this". A
displaced block caret is neither convention, which is why it read as a third
thing nobody has a mental model for.

So `_paint_ghost_caret` and its call site are removed and the base class's paint
stands. The reasoning is recorded as a comment at the call site rather than
deleted with the method, so a future reader does not "fix" the displacement back
in.

### The empty ghost is withheld rather than compensated for

The problem #378 was solving still needs an answer. A whitespace-only ghost —
the exact state in the user's screenshot — draws zero dim pixels by
construction, so with the caret back at the insertion point the frame is a block
on a blank cell one past the text: **identical to the composer with no ghost at
all**. Nothing is lost by not painting it, while the previous behaviour spent a
whole caret displacement on it.

`_ghost_completion` therefore withholds a ghost that is whitespace-only (gate 4,
beside the three existing gates). It lives there and not in `ghost_for` because
the two answer different questions: `ghost_for` asks "can this completion be
described as a pure append?", which is a property of the completion; whether an
honest append is worth PAINTING depends on the rendered frame, which is what
gates 1-3 (caret position, width, single line) already judge. Putting it in
`ghost_for` would also hand a rendering decision to that function's other
caller.

**This does not change what Tab does.** The invariant is
`buffer + ghost == buffer after Tab`, stated over a ghost that is shown;
withholding a preview is not an edit to the completion. Tab still inserts the
trailing space, and a test pins exactly that.

### The cell under the block needed ink, judged from frames

Item 3 of the brief asked whether the base class's paint is legible under the
block, to be decided from rendered frames rather than from reading style code.
It is **not**, and the frames are what showed it.

Textual applies `text-area--cursor` over the whole caret cell, so with the caret
at the insertion point the ghost's first character was painted in the caret's
full ink — measured `#1e1a14` on `#e9e5db`, a **13.76:1** contrast, byte-identical
to a caret sitting on a character the user actually typed. In the `/resum` frame
that showed as a solid black-on-cream `e` while its own trailing space stayed
dim: the committed/previewed boundary landed INSIDE the ghost.

`Editor._paint_ghost_ink` re-inks that ONE cell. The caret keeps its background
— the block is load-bearing elsewhere, and the boot composer, the read-only
composer and the attachment chip all assert an inverted caret cell — and the
suggestion's own foreground goes back on top. That is **3.29:1** against the
cursor ground: comfortably legible, and visibly not the 13.76:1 of committed
text, so the cell reads as "caret, on previewed text", which is what it is. A
post-pass over the finished `Strip`, matching the idiom `_paint_markers` and
`_paint_slash` already establish. **The caret does not move.**

The three existing gates (caret at line end, fits remaining width, single line)
and the pure-append rule are untouched.

## Related Issue(s)

- User-reported ghost-caret misreading (no ticket). Follows #378, which
  introduced `_paint_ghost_caret`.

## Impact

- No breaking changes, no dependency updates, no new public API.
- Version bumped to `0.43.5` (patch: a user-visible rendering correction, not a
  step-function capability). Rebased onto `b44575d2`; `0.43.3` and `0.43.4` were
  both taken by main while this was in review, and both are live on PyPI.

### Visual validation (per AGENTS.md "Visual validation")

Frames captured from the **real `OperatorApp`** via `run_test` +
`app.save_screenshot()` — the host that loads `local_operator.tcss`, not a bare
test host — then rendered to PNG and **looked at**. BEFORE is `origin/main` at
**`b44575d2`** in a separate detached worktree; AFTER is this branch rebased onto
that same commit, so both sides share a baseline. Committed under
[`docs/evidence/ghost-caret/`](docs/evidence/ghost-caret/) with a README, and
reproducible with the committed `shot.py`:

```sh
env -u NO_COLOR TERM=xterm-256color .venv/bin/python \
    docs/evidence/ghost-caret/shot.py . /tmp/ghost-after
```

`caret-*.png` are the load-bearing frames: the composer row cropped to the caret
cell at 8x, **before stacked above after**.

**The reported case — `/resume` fully typed.** Before: the block sits on the
`e`, which reads as "this character is about to be replaced". After: the
whitespace-only ghost is withheld and the caret sits after the text, exactly
where the next character goes.

/resume fully typed, before above after

**A real multi-character ghost — `/resum`, ghost `e `.** Before: the block is on
the typed `m` and the previewed `e` is bright beside it. After: the block is at
the insertion point on the `e`, and that `e` is dim, so the committed/previewed
boundary is where the caret is.

/resum with ghost, before above after

**An argument-slot ghost — `/mcp lo`, ghost `gin`.** A different completion path,
landing identically: the block moves off the typed `o` and onto the previewed
`g`, which stays dim.

/mcp lo with ghost, before above after

**A command typed mid-draft — also fixed, and NOT the control this PR first
claimed (design review round 1, D1).** `hi /resume` produces the same
whitespace-only ghost as a bare `/resume`, so the caret was displaced here too
(col 9 → col 10) and gate 4 withholds it the same way. What the user reported is
that this state *looked* right; it now genuinely is. The states this change does
not reach are `prose-no-ghost` and `narrow-width`, and both are pixel-identical
before and after (0 differing pixels, measured below).

command mid-draft, before above after

| frame | state | before → after |
|---|---|---|
| `caret-resume-full.png` | `/resume`, picker open | **the reported case.** Block over the `e` → whitespace-only ghost withheld, block after the text on a blank |
| `caret-resum-partial.png` | `/resum` → ghost `e ` | block on typed `m` → block on the previewed `e`, and that `e` still dim |
| `caret-mcp-lo.png` | `/mcp lo` → ghost `gin` | argument slot, a different completion path: block `o` → `g` |
| `caret-mcp-login-n.png` | `/mcp login n` → ghost `otion` | compound server row: block `n` → `o` |
| `caret-mid-draft.png` | `hi /resume` | **also fixed, not a control** — see the correction below |
| `native-*.png` | the three multi-char states at 1x | the dimmed cell judged at the size a user sees, not at the 8x that flatters it |
| `before/after-narrow-width.png` | `/analytic` at 18 cols | gate 2 withholds the ghost — **unchanged** |
| `before/after-prose-no-ghost.png` | `hello there` | no command token — **unchanged** |

`shot.py` also saves a `.frame2.svg` per state one settle later. Every pair is
identical, so nothing reflows after paint.

### Geometry behind the pixels

The **column** of the cell carrying the cursor background is the number that
matters; a strip's `cell_length` cannot show it, because the composer pads the
row to its box width (69 cells at w=100 in every state, ghost or not, both
trees). From `geometry-before.json` / `geometry-after.json`, written by the same
run that saved the frames:

| state | buffer | ghost before | caret cell before | ghost after | caret cell after |
|---|---|---|---|---|---|
| `resume-full` | `/resume` | `' '` | col 6 `'e'` | `''` | col 7 `' '` |
| `resum-partial` | `/resum` | `'e '` | col 5 `'m'` | `'e '` | col 6 `'e'` |
| `mcp-lo` | `/mcp lo` | `'gin'` | col 6 `'o'` | `'gin'` | col 7 `'g'` |
| `mcp-login-n` | `/mcp login n` | `'otion'` | col 11 `'n'` | `'otion'` | col 12 `'o'` |
| `mid-draft` | `hi /resume` | `' '` | col 9 `'e'` | `''` | col 10 `' '` |
| `narrow-width` | `/analytic` (w=18) | `''` | col 9 `' '` | `''` | col 9 `' '` |
| `prose-no-ghost` | `hello there` | `''` | col 11 `' '` | `''` | col 11 `' '` |

In every ghosted state the caret moves from `len(buffer) - 1` to `len(buffer)`,
the insertion point. The two no-ghost states are identical, which pins that this
change reaches only the ghosted path.

**The pairs are directly diffable.** Both sides are shot with `os.getcwd()`
pinned to one string, because the two worktrees' paths otherwise appeared in the
banner and status line and made every pair differ for a reason unrelated to the
change (design review round 1, D3/D4). Measured with `magick compare -metric AE`
on the committed PNGs: `narrow-width` **0** and `prose-no-ghost` **0** — the two
genuine controls, byte-identical — against 1045-1411 differing pixels for the
five ghosted states, all confined to the composer row.

Screen geometry is unchanged, measured directly rather than inferred from the
stills — editor content box `(69, 1)`, screen `(98, 22)`, `virtual_size`
`(98, 22)` equal to the screen so nothing became scrollable, and
`show_vertical_scrollbar` False — in **both** trees over three consecutive runs.

### Test coverage

`tests/unit/tui/test_ghost_text.py`. The old caret assertions pinned the
displacement and would have kept it; they are rewritten to pin the new rule, not
weakened. The invariant tests are untouched.

- `test_the_caret_block_sits_at_the_insertion_point_on_the_ghost_head`
  (`/mc`, `/analytic`, and `/mcp lo` for the argument path) — asserts the caret
  is at column `len(typed)` on the ghost's head. A displaced caret puts the
  cursor ground at `len(typed) - 1` and fails. It also asserts EVERY ghost cell,
  the one under the block included, carries the suggestion foreground — the
  regression `_paint_ghost_ink` exists to prevent.
- `test_a_whitespace_only_completion_shows_no_ghost_but_tab_still_inserts_it`
  (`/resume`, `/mcp`, `/analytics`) — the new regression test. Asserts no ghost,
  asserts the caret sits where it would with no ghost at all, and **asserts Tab
  still produces `typed + " "`**, so withholding the preview can never become a
  change to the edit.

Two existing gate-isolation tests seeded themselves with the fully typed `/mcp`
and used its `' '` ghost as their "a ghost IS on offer here" precondition, which
gate 4 now withholds. Both were re-seeded to `/mc` (ghost `'p '`) so the
precondition is real again and the gate under test remains the only thing that
can produce the result. Their assertions are unchanged in kind.

### Full suite and gates

Run from the worktree, over the whole tree, exactly as CI runs them:

```
python -m flake8 .                                    clean
uvx --from black==26.1.0 black --check .              605 files unchanged
uvx isort==5.13.2 --check .                           clean
python -m pyright --pythonpath <venv> .               0 errors, 0 warnings, 0 informations
python -m pytest tests/unit -q                        8638 passed, 15 skipped, 1 xfailed
    python -m pytest tests/unit/tui -q                3565 passed, 12 skipped, 1 xfailed
```

`tests/unit/tui/test_ghost_text.py` alone: **51 passed**.

#### A note on the venv these were run in, and two failure sets that are NOT this change

Worth recording, because the first run of the suite in this worktree looked
alarming and neither cause is the diff.

**1. 24-25 `tests/unit/tools/test_eval_tool.py` failures — an environment
artifact, proven not introduced.** The symptom was
`ModuleNotFoundError: No module named 'local_operator'` from the eval worker
subprocess. Root cause: the shared `~/local-operator/.venv` editable install's
import finder maps `local_operator` to `/tmp/lop-tui-performance`, **a worktree
that no longer exists**, so a borrowed-venv run only resolves the package while
cwd happens to be a checkout — and `eval.py` spawns its worker with
`cwd=tmp_path`. Confirmed by running the same ids on a clean `origin/main`
worktree under the identical invocation, where they fail the same way. Fixed
properly rather than worked around, by building a real venv for the worktree
(`uv venv && uv pip install -e ".[all,dev]"`); **all 47 eval tests pass there**,
and the whole-suite numbers above come from that venv. Baseline `e641a888` in
its own equivalent venv: 8425 passed.

**2. Intermittent `tests/unit/tui` failures — pre-existing, parallel-order
dependent.** The suite runs `-n auto` (xdist) by default, and a handful of ids
fail intermittently with a different id each run. Measured with the required
`env -u NO_COLOR TERM=xterm-256color`, three full runs per tree:

| | failing ids across 3 runs |
|---|---|
| baseline `e641a888` | 3 (`test_aside.py::test_escape_dismisses_the_aside_before_it_stops_the_turn`, `test_app_pilot.py::test_mcp_logout_list_offers_only_servers_holding_a_credential`, `test_composer_seam.py::test_the_row_under_each_panel_survives_as_the_dock_s_own_fill[both-size0]`) |
| this branch | **0** |

Every one of those ids passes **isolated with the correct env on both trees**.
None touch ghost text. Verdict per id: passes isolated with correct env; appears
on baseline; not introduced by this change. Not fixed here — it is outside this
slice and belongs in its own change.

### Rebases: #388, #401, the 0.43.3 release, and #365 landed mid-flight

`origin/main` moved three times while this was in review, from `e641a888` to
`b44575d2`, picking up #388 (double/triple-click composer selection), #401
(OAuth refresh lock on `/resume`), the 0.43.3 release commit, and #365
(compaction advisor). Each rebase was checked rather than trusted:

- **No semantic conflict.** #388 is the only one touching `editor.py`, and its
  hunks are in `__init__` (new click-chain state), `_on_click`, and the module
  docstring. It does not touch `render_line`, `_ghost_completion`, or any ghost
  path; its only `render_line` mention is a docstring cross-reference. #365
  touches no file this PR touches. The diff against main is 26 files:
  `editor.py`, `test_ghost_text.py`, `pyproject.toml`, and `docs/evidence/`.
- **The frames were RE-SHOT on each new base**, never carried forward — a
  baseline predating a change to the composer's own stylesheet cannot prove what
  the composer renders now. The defect reproduces identically on every baseline
  tried (same ghost, same caret column, all seven states).
- **Version.** `0.43.3` and `0.43.4` were both taken by main mid-review and are
  both live on PyPI, so the bump is now `0.43.5`. Patch throughout.
- **Suites re-run on each rebased head.** On `b44575d2`: `tests/unit` 8638
  passed, `tests/unit/tui` 3565 passed, ghost suite 51 passed.

### Two pre-existing test failures, proven not attributable to this PR

Both were established by running the exact ids on a clean worktree of
**unmodified `b44575d2`** in its own venv, not inferred:

- `tests/unit/test_session_factory.py::test_dispose_closes_auth_store` — fails
  **isolated** on unmodified `b44575d2` (`assert 2 == 1`), and passes within
  main's own full-suite run. Order-dependent and pre-existing from the #365
  base; this PR touches nothing it exercises.
- A rotating set of xdist ordering flakes (`test_launch_subagent`,
  `test_resume_subagents_todos`, `test_frontend_update_burst_coalesces_to_latest_snapshot`,
  `test_aside`, `test_composer_seam`). The suite runs `-n auto`; a different id
  fails on each run and every one passes isolated. Measured over repeated full
  runs, unmodified main flakes at the same rate with its own varying ids.

CI's earlier `test (3.13)` failure on
`test_word_caret.py::test_real_teardown_settles_a_held_escape_and_leaks_no_callback`
(`KeyError: text-area--gutter`, raised inside Textual's own `render_lines`) was
the same class: the identical head passed 3.13 in the sibling run, the test
passes locally on both branch and baseline, and the re-run went green.

- The composer's block caret now sits **after** the text you have typed, where
  the next character will go, instead of jumping backwards onto the last
  character while a completion preview is showing. A fully typed `/resume` no
  longer looks like the `e` is about to be overwritten.
- A command with nothing left to preview but its trailing space shows **no**
  ghost, so the composer looks the same as it does with no preview at all.
  <kbd>Tab</kbd> still completes it exactly as before, trailing space included.
- When a preview IS showing, the character under the caret block is drawn in the
  preview's dim ink rather than the caret's bright ink, so the boundary between
  what you typed and what is being suggested stays readable.

## Checklist


damianvtran:
### Design review — round 1
Scope: 80df237b..fff97c15

Judged from the rendered PNGs in `docs/evidence/ghost-caret/` (the `caret-*.png` zoom pairs as the load-bearing frames, the `before-*`/`after-*` full frames for surrounding movement), with pixel measurements taken off those same frames rather than from the style code or the JSON.

**Verdict up front: the change does what it set out to do.** On the frames, the after state reads as an insertion point, not a selection. No BLOCKER, no MAJOR.

---

#### The four questions in the brief

**1. Does the after state read as "caret after my typed text"? Yes, unambiguously.** In `caret-resume-full.png` the before panel is a cream block sitting squarely on a black `e`, glyph fully enclosed — that is the vi-overwrite / "this character is selected" idiom, exactly as reported. The after panel is a block on a blank cell one past the `e`, with `resume` intact in committed ink. Measured: the before caret cell carries a glyph at 13.76:1 against the block ground and 10.6% ink coverage; the after cell measures 0.0% ink coverage — an empty block. There is nothing left in that cell for a user to read as "selected". This is the ordinary end-of-line caret every terminal shows, and gate 4 withholding the whitespace-only ghost is what earns it.

**2. Legibility of the covered first ghost char, and the committed/previewed boundary: the post-pass shipped and it was needed.** In `caret-resum-partial.png`, `caret-mcp-lo.png` and `caret-mcp-login-n.png` the block moves exactly one cell right (measured +55, +57, +58 px against a 55-57px cell — one cell, no drift), and the covered glyph is painted `#837c6d` on the `#e9e5db` block: **3.29:1**, against **13.76:1** for a caret on genuinely committed text. That 4.2x separation is visible at a glance, not just in the numbers. `caret-mcp-lo.png` is the clearest read: `lo` in bright committed ink, then `g` under the block in grey, then `in` in the same grey continuing to the right. The boundary lands where the user stopped typing, and the `g` reads as part of the `gin` preview rather than as the last thing they typed. Without `_paint_ghost_ink` this cell would have measured 13.76:1 — identical to committed text — and would have put the boundary one character too far right. **The treatment is not still needed; it shipped and it works.** The covered glyph is fully legible: stroke shape is intact, `g`/`o`/`e` are all identifiable, and ink coverage inside the block (4.8-5.5%) matches the uncovered glyphs.

**3. Ghost contrast against the composer background: adequate, and unchanged by this PR.** The uncovered ghost is `#837c6d` on `#1e1a14` = **4.18:1**. I checked this is pre-existing rather than introduced: the ghost tail measures byte-identical `#837c6d` / 4.18:1 in both the before and after panels of `caret-mcp-login-n.png`. Below the 4.5:1 body-text threshold, but this is intentionally de-emphasised non-committed text and it is not a regression from this change. Recorded as a follow-up (D2), not a finding against this PR.

**4. Mid-draft and narrow-width unchanged: one holds, one does not.** Narrow-width holds — see D3 for the verification, the caret is bit-identical bar 1px of subpixel noise. Mid-draft does **not**, and the PR says it does. See D1.

---

#### Findings

**D1 — MINOR — `caret-mid-draft.png`, `before/after-mid-draft.png`, `geometry-{before,after}.json`: the PR and README claim mid-draft is unchanged; the committed evidence shows it changed.**
The README frame table says of `mid-draft`: *"The state the user calls correct. Unchanged: caret after the text either way."* The frames contradict it. Masking the cwd banner (see D4), the mid-draft pair differs by 3,620 px, all of it in the composer row, and the crop shows the before panel with the block on the `e` of `/resume` and the after panel with the block on the blank past it — the same one-cell move as every other ghosted state. The README's own geometry table on the next page records it correctly (`ghost ' '` → `''`, `col 17 'e'` → `col 18 ' '`), so the prose and the table disagree with each other.

Why it matters: this is not a defect in the shipped pixels — the after state is right, and the change is an improvement here too. It matters because the premise stated in both the PR body and the README is that mid-draft was already correct and is the control showing the fix is scoped to the picker case. It was not correct; it carried the same defect, because `check this /resume` also produces the whitespace-only ghost. A future reader debugging this area will use that sentence as a baseline and be misled about which states the ghost path reaches. The genuinely unchanged control is `prose-no-ghost`, which I verified is pixel-identical outside the banner.

Fix: no re-shoot needed, the frames already show the truth. Correct the two prose claims to say mid-draft carried the same whitespace-only ghost and is fixed the same way, and point at `prose-no-ghost` (and narrow-width) as the untouched controls. Worth landing before merge since the PR body is the durable record.

**D2 — NIT (follow-up, pre-existing) — `caret-mcp-lo.png`, `caret-mcp-login-n.png`: ghost ink sits below the 4.5:1 threshold both on the page and under the block.**
Uncovered ghost `#837c6d` on `#1e1a14` is 4.18:1; the covered cell is 3.29:1 on the block ground. Both are legible at terminal sizes on these frames and the dimness is doing deliberate semantic work, so I am not asking for a change in this PR — and it is not a regression, the tail ink is identical before and after. If the ghost palette is ever revisited, `#898272` reaches 4.53:1 on the page and `#6c665a` reaches 4.53:1 under the block, and both stay far from the 13.76:1 of committed text, so the boundary would still read. Not worth a remediation round.

**D3 — NIT — `before/after-narrow-width.png`: the pair is vertically offset by 434px, so it cannot be compared as-shot.**
The two narrow captures render the cwd in the banner (`lop-before2` vs `lop-ghost-caret`), which wraps to a different line count at 18 columns and pushes everything below it down. A naive diff of this pair reports 1.2M changed pixels and looks alarming. I located the caret block independently in each frame and compared aligned strips: caret at x=1061 before, x=1060 after — a 1px subpixel difference, i.e. **unchanged**, which is what gate 2 withholding the ghost should produce. The claim is correct; the artifact just cannot demonstrate it without this extra step. Fix: note the offset in the README, or re-shoot both sides from identically-named worktrees.

**D4 — NIT — all `before-*`/`after-*` pairs: the cwd string differs, so no pair is pixel-identical.**
Every pair differs in the banner (rows 210-412) and the footer status line (rows 570-605) purely because the before tree is `/private/tmp/lop-before2` and the after tree is `/private/tmp/lop-ghost-caret`. Harmless once known, but it means "diff the pair and see nothing move" does not work as a check, which is the stated purpose of the full-screen frames. Fix: shoot both trees from paths of equal length, or stub the cwd in `shot.py`.

**D5 — NIT — `caret-mid-draft.png`: the caret cell is clipped at the right edge of the crop.**
In the before panel the block is cut off mid-cell at the frame boundary and the `e` under it is only partially visible; the after panel loses the caret entirely. Every other `caret-*.png` centres the cell with room either side. This is the one crop where the load-bearing detail is not fully in frame, which is unfortunate given it is the state D1 is about. Fix: widen the crop window for this state.

---

#### Residual ambiguity for a first-time user

None that I would hold the PR for. The one thing worth naming: with a multi-char ghost the block now sits on a character the user did not type, which is a different frame from the empty block they see when there is no suggestion. On these frames it resolves correctly because the covered glyph stays in ghost ink and the rest of the ghost continues in the same ink to its right — the block reads as "caret, sitting on preview". It resolves because `_paint_ghost_ink` exists; the 13.76:1 version the PR describes and rejected would have been genuinely ambiguous. The `/resume` case, where the whole ghost is withheld, is unambiguous by construction.

Frame-to-frame settling is not a concern here: `shot.py` captures a `.frame2.svg` one settle later for every state and the pairs are reported identical, and nothing in this change is animated.

**Round 1 is TERMINAL.** No blockers, no majors. D1 is a documentation correction to the PR body and evidence README that should land before merge; D2-D5 are evidence-hygiene and palette follow-ups that do not need a remediation round.

damianvtran:
### Agent review — round 1
Scope: 80df237b..fff97c15

Reviewed the diff, then ran the code: all four CI gates from AGENTS.md over the whole tree, the ghost suite serially, the full `tests/unit/tui` suite on this branch and on a `80df237b` baseline worktree, and two mutation runs to check the tests actually pin the new rule.

**Verification performed**

- `flake8 .` clean; `black==26.1.0 --check .` 594 files unchanged; `isort==5.13.2 --check .` clean; `pyright` **0 errors, 0 warnings**.
- `tests/unit/tui/test_ghost_text.py` serially: **49 passed**.
- **Mutation 1** — reverted gate 4 to `if False:`. All three parametrizations of `test_a_whitespace_only_completion_shows_no_ghost_but_tab_still_inserts_it` fail (`'/mcp' previewed ' '`). The gate is genuinely pinned.
- **Mutation 2** — removed the `_paint_ghost_ink` call from `render_line`. 3 failed / 46 passed, failing on `ghost_cols` (`[8, 9] == [7, 8, 9]`). The ink pass is pinned too, at the exact cell that matters.
- Full `tests/unit/tui`: 2 failed / 3561 passed on this branch; both failures pass in isolation, and `test_frontend_update_burst_coalesces_to_latest_snapshot` fails identically on the `80df237b` baseline (1 failed / 3559 passed). Confirmed pre-existing xdist ordering flake, **not attributable to this PR**.
- Focus areas 1–6 all check out; details under "What I verified clean" below. One blocker, on focus area 7.

---

#### B1 — BLOCKER — `pyproject.toml:7`: the version bump collides with an already-published release

`pyproject.toml` sets `version = "0.43.3"`, described in the PR body as "patch … Rebased onto `80df237b`, which had already taken `0.43.2`". That was true when written, but `origin/main` has since moved:

```
0008afef chore(release): bump version to 0.43.3 (#409)   <- on main, NOT in this branch
80df237b fix(mcp): ... (#401)                            <- this PR's base
```

- `git show origin/main:pyproject.toml` → `version = "0.43.3"`
- `gh release view v0.43.3` → published 2026-08-29T17:07:30Z, target `0008afef`
- `curl https://pypi.org/pypi/local-operator/0.43.3/json` → **HTTP 200**; PyPI `info.version` is already `0.43.3`
- `git merge-base --is-ancestor fff97c15 v0.43.3` → false, so this change is **not** in the shipped 0.43.3

`git merge-tree --write-tree origin/main fff97c15` merges cleanly with **no conflict** — the bump hunk is identical on both sides, so git silently resolves it and the merged tree still reads `0.43.3`. That is what makes this dangerous rather than obvious: nothing will fail at merge time. The release step then either refuses the duplicate tag or, worse, publishes a `0.43.3` sdist/wheel that PyPI rejects as a duplicate filename, and every user who already has 0.43.3 never receives this fix.

Why it matters: the fix silently does not ship. PyPI is immutable per version, so a duplicate upload cannot be corrected in place.

**Fix:** rebase onto current `origin/main` (`0008afef`) and set `version = "0.43.4"`; tag `v0.43.4`. Patch remains the correct materiality call (a rendering correction, not a step-function capability), and the PR body's "Impact" line should be updated to say it rebased onto `0008afef`, which had already taken `0.43.3`.

---

#### N1 — MINOR — `local_operator/tui/widgets/editor.py:4649`: gate 4 is not tested on the NAME_ARGUMENT path

`test_a_whitespace_only_completion_shows_no_ghost_but_tab_still_inserts_it` parametrizes `/resume`, `/mcp`, `/analytics` — all `CompletionMode.COMMAND`. The whitespace-only ghost is reachable on a second path too; driving the pure functions directly:

```
'/team alpha' -> completion=('/team alpha ', 12)  ghost=' '   <- gate 4 withholds
'/team alph'  -> completion=('/team alpha ', 12)  ghost='a '  <- shown
```

`completion_for` appends `suffix = " "` for `NAME_ARGUMENT` (`command_picker.py:508`), so a fully typed team/agent name produces exactly the `' '` ghost gate 4 exists for. The behaviour is correct today — the gate sits in `_ghost_completion`, downstream of the mode split, so it covers both paths by construction. But the caret test already reaches for the argument path deliberately (`/mcp lo`) while the gate-4 test does not, so the coverage is asymmetric and a future refactor that moved the gate mode-side would go unnoticed.

**Fix:** add a `("/team alpha", …)`-style case to the gate-4 parametrization, asserting no ghost and that Tab still yields `"/team alpha "`.

#### N2 — NIT — `local_operator/tui/widgets/editor.py:4580-4586`: gate 4's numbering is inconsistent with its own list

The `_ghost_completion` docstring now reads "GATES 1-3 below are what put both out of reach … Gate 4 is a later addition of a different kind … documented at its own site", then immediately lists items 1, 2, 3, **and 4** inline anyway, with the blank line before `4.` making it read as a separate list. Item 4 is one sentence while the real explanation is 20 lines down at the code site. Minor, but this file's comment convention is unusually disciplined and the split reasoning is the one place a reader could miss the constraint.

**Fix:** either drop item 4 from the numbered list (the prose already forward-references it) or delete the blank line so it reads as the fourth entry of one list.

#### N3 — NIT — `docs/evidence/ghost-caret/`: 2.5M of PNGs committed to the repo

Consistent with existing precedent (`mcp-verbs` 1.1M, `clipboard-paste` 1.1M, `settings-page` 956K), so this is not a new problem and I would not hold the PR for it. Flagging only that this directory is roughly 2x the largest prior one, and the PR body embeds the PNGs by `raw.githubusercontent.com` URL pinned to the `dev-ghost-caret` branch — those image links will 404 once the branch is deleted after merge. Repointing them at the merge commit SHA or at `main` would keep the PR body readable in six months.

---

#### What I verified clean (focus areas 1-6)

1. **Caret rendering / dead code.** `_paint_ghost_caret` and its call site are gone with no orphans: `grep -rn "_paint_ghost_caret"` returns only intentional prose references in the new comment and the evidence README. No import became dead — `RichStyle` and `Segment` are still used by `_paint_ghost_ink` and by `_overlay` at line 2702, `Offset`/`Strip` unchanged. `_paint_markers` → `_paint_slash` → `_paint_ghost_ink` ordering is correct: `_paint_slash` is foreground-only `post_style` (documented at line 2896 as composing with the cursor's inverse), and `_paint_ghost_ink` runs outermost so its foreground re-ink lands last on the caret cell, which is exactly what mutation 2 confirmed. The dropped `column <= 0` guard is safe — it existed only to protect the `column - 1` arithmetic, which no longer exists; the remaining `caret_x >= strip.cell_length` bound is the one that matters, and I confirmed `Strip.divide` degrades safely on a wide-glyph first ghost cell (returns a 1-cell blank, total `cell_length` preserved, no overrun).

2. **Whitespace suppression does not change Tab.** Correct, and correctly *placed*. The gate is in `_ghost_completion` (a render-side question), not in `ghost_for` or `completion_for`, so the accept path — `_apply_command` → `_completion_for` → `_set_text_and_caret` — never consults it. The `buffer + ghost == buffer after Tab` invariant is preserved by construction because both sides still derive from the same `completion_for`. The test asserts the commit directly (`after == typed + " "`), which is the right assertion; mutation 1 shows it is load-bearing.

3. **Gates 1-3 and the pure-append rule.** Untouched. Diff on `local_operator/` is 5 hunks in one file; `ghost_for` and `completion_for` in `command_picker.py` are byte-identical to base. Gate 2's `>=` boundary, gate 3's single-line test, and the `startswith` case-sensitivity rule are all unmodified.

4. **Test quality.** The caret assertions are genuinely re-pointed, not weakened — they went from "caret is on the typed char" to the strictly stronger pair "caret is at column `len(typed)` on the ghost head" **and** "every ghost cell including the one under the block carries suggestion ink", which is a new assertion the old tests did not make. Everything runs through `app.run_test` with real `pilot.press` per character (`_type`, line 76), not direct method calls. The `_is_cursor` background-only matcher is the right call and its docstring says why. The two re-seeded gate-isolation tests (`/mcp` → `/mc`) are a legitimate re-establishment of a precondition gate 4 removed, not a weakening: their assertions are unchanged in kind and the ghost they assert (`'p '`) is strictly more substantive than the `' '` they used before.

5. **Rebase over #388 and #401.** Coherent. `git merge-tree` against `origin/main` is conflict-free, and the only overlap with #388 is `local_operator.tcss`, which this PR does not touch at all — it reads `text-area--suggestion` (`color: $lo-dim`) and `text-area--cursor` rather than redefining either, so #388's selection-band work is untouched. The evidence was re-shot against the post-rebase `80df237b` baseline rather than carried forward from `e641a888`, which is the correct call and is called out explicitly in the README.

6. **Comment quality.** This is the strongest part of the PR. The removed docstring's reasoning is not lost — the call-site comment at `editor.py:1160-1183` records the full why (the overwrite misreading, the self-contradiction between two states of one command, the five-system survey, the browser-URL-bar counter-convention) and states plainly that re-adding a caret pass "is a regression, not a fix". Gate 4 carries its own why at `editor.py:4626-4648`, including the explicit "THIS DOES NOT CHANGE WHAT TAB DOES" invariant and the reasoning for placement over `ghost_for`. Somebody will not "fix" this back by accident.

**On `_paint_ghost_ink`:** I re-derived the contrast claims from `theme.py` rather than trusting the PR body. Dark ramp: ghost ink on the caret block is **3.29:1** vs **13.76:1** for committed text — both figures as claimed. I also checked the light ramp, which the PR does not mention: **4.01:1** ghost vs **13.86:1** committed, so the distinction holds there too and is in fact slightly better. No finding; noting it because this repo's convention (see the `text-area--selection` comment) is to check colour rules across ramps.

**Visual evidence:** I looked at `caret-resum-partial.png` rather than taking the README's word for it. The before/after pair shows exactly what is claimed — block on the typed `m` above, block at the insertion point on the previewed `e` below, with that `e` still rendering in dim suggestion ink rather than the caret's full ink. The evidence is real and reproducible via the committed `shot.py`, which drives the real `OperatorApp` (CSS_PATH-loading), per AGENTS.md "Visual validation".

---

### Verdict

**Changes requested — 1 blocker.**

The rendering change itself is correct, well-tested, and unusually well-documented; I tried to break it two ways with mutation runs and both fixes are properly pinned. Focus areas 1-6 are clean. The blocker is entirely mechanical and external to the fix: `main` published `0.43.3` after this branch was cut, so the bump now collides with a live PyPI release, and git will merge it without complaint. Rebase onto `0008afef` and bump to `0.43.4`.

N1 is worth folding in on the same rebase. N2 and N3 are follow-ups and should not cost a round.

damianvtran:

Scope: `b44575d2..06d5b6f7`
Head at remediation: `06d5b6f7dbb83777cb4f68a8aeb138d4366e35a7`

All nine findings from both rounds are answered below. The code round's B1 and N1 and the design round's D1 are **fixed**; N2, D3, D4 and D5 were cheap enough to fix rather than defer, so they are fixed too. D2 is **rejected as a finding against this PR** (it is pre-existing and the reviewer flagged it as a follow-up); N3 is **fixed in part, rejected in part**.

---

#### B1 — BLOCKER — version collision — **FIXED** in `06d5b6f7`

Verified independently before acting: `git show origin/main:pyproject.toml` read `0.43.3`, `gh release view v0.43.3` was published, and PyPI's `info.version` was `0.43.3`. The bump hunk being byte-identical on both sides is exactly why `git merge-tree` resolved it silently, as the review said.

Then it happened **again mid-remediation**: main moved to `b44575d2` (#365), which took `0.43.4`, and PyPI now serves `0.43.4`. So the branch is rebased onto `b44575d2` and the bump is **`0.43.5`**; tag `v0.43.5`. Patch remains correct — a rendering correction, not a step-function capability. The PR body's Impact line now names `b44575d2` and records that both `0.43.3` and `0.43.4` were taken while this was in review.

Confirmed clean on the new base: `git diff --name-only origin/main..HEAD` is 26 files (`editor.py`, `test_ghost_text.py`, `pyproject.toml`, 23 under `docs/evidence/ghost-caret/`). #365 touches no file this PR touches.

#### N1 — MINOR — gate 4 untested on the NAME_ARGUMENT path — **FIXED** in `06d5b6f7`

Confirmed on the real widget path rather than the pure functions, driving `pilot.press` per character:

```
'/team char'    row='chart'  ghost='t '  after_tab='/team chart '
'/team chart'   row='chart'  ghost=''    after_tab='/team chart '
'/agent code'   row='coder'  ghost='r '  after_tab='/agent coder '
'/agent coder'  row='coder'  ghost=''    after_tab='/agent coder '
```

`test_a_whitespace_only_completion_shows_no_ghost_but_tab_still_inserts_it` now parametrizes `/team chart` and `/agent coder` alongside the three COMMAND cases, with a comment recording that the gate covers both paths by construction and that these cases exist to catch a future refactor moving it mode-side. **5 passed.**

Checked the cases are load-bearing rather than decorative: mutating the gate to `if False:` fails **all five** parametrizations, the two new ones included.

#### N2 — NIT — gate 4's numbering inconsistent with its own list — **FIXED** in `06d5b6f7`

The docstring said "GATES 1-3" and then listed four items, with a stray blank line making item 4 read as a separate list. It now says gates 1-3 are the wrap-hazard gates, states that gate 4 is a different **kind** (an honest preview that renders to nothing, not a mispaint), and explains that its full reasoning lives at the code site so it does not outweigh the three gates preventing actual corruption. Item 4 is folded into the same list with a pointer to that comment.

#### N3 — NIT — 2.5M of PNGs, and image links pinned to the branch — **FIXED** (links) / **REJECTED** (size)

- **Links: fixed.** The PR body's four embedded images now resolve through the immutable commit SHA `06d5b6f7…` instead of the `dev-ghost-caret` ref, so they survive branch deletion after merge. Verified: all four return HTTP 200.
- **Size: rejected for this PR.** As the review notes, this is consistent with existing precedent (`mcp-verbs` 1.1M, `clipboard-paste` 1.1M). The directory is 2.5M because it carries 7 before/after full-frame pairs plus 5 zoom crops, and the full frames are what make the "nothing else moved" check possible — now literally, since two pairs diff to zero. Trimming the set would cost the evidence its point. A repo-wide policy on evidence-image budgets is a reasonable follow-up but is not this PR's to set.

#### D1 — MINOR — mid-draft claimed unchanged, but it changed — **FIXED** in `06d5b6f7`

Confirmed against the committed geometry before touching the prose: `mid-draft` moved `ghost ' ' → ''` and `col 17 'e' → col 18 ' '`. The review is right that the prose contradicted the table beside it, and right about the cause — `check this /resume` yields the same whitespace-only ghost, so it carried the same defect.

Corrected in **both** places, no re-shoot needed for the claim itself:

- `docs/evidence/ghost-caret/README.md` — the frame-table row now says "also fixed, not a control", and a new paragraph explains that the user reported this state *looked* right and that it now genuinely is, naming `prose-no-ghost` and `narrow-width` as the real untouched controls.
- The PR body — the embedded-image caption and the frame table say the same, and the geometry table row is updated.

`shot.py` also carries a note that this state is not a control, so a future capture cannot re-acquire the wrong premise.

#### D2 — NIT — ghost ink below 4.5:1 — **REJECTED as a finding against this PR** (recorded as a follow-up)

Agreed on the measurements and agreed with the reviewer's own framing: the uncovered ghost is 4.18:1 and the covered cell 3.29:1, and the tail ink is **byte-identical before and after**, so this is not a regression from this change. The dimness is doing deliberate semantic work — the whole point of the boundary is that preview ink is not committed ink — and raising it toward 4.5:1 would narrow the separation from committed text's 13.76:1 that D1's verdict and the `_paint_ghost_ink` design both rest on. Changing the palette here would also reach every ghost in the composer, well outside this PR's slice. The reviewer's suggested values (`#898272` on the page, `#6c665a` under the block) are recorded in this thread for whoever revisits the palette.

#### D3 — NIT — narrow-width pair vertically offset by 434px — **FIXED** in `06d5b6f7`
#### D4 — NIT — every pair differs in the cwd banner and footer — **FIXED** in `06d5b6f7`

Fixed at the source rather than by cropping the banner away. `shot.py` now pins `os.getcwd()` to a single string for the capture, so both trees render an identical banner and status line and the frames still show the real surface. Both sides were re-shot from worktrees of equal path length as well.

Measured on the committed PNGs with `magick compare -metric AE`:

| pair | differing pixels |
|---|---|
| `narrow-width` | **0** |
| `prose-no-ghost` | **0** |
| `resume-full` | 1409 |
| `resum-partial` | 1299 |
| `mcp-lo` | 1045 |
| `mcp-login-n` | 1371 |
| `mid-draft` | 1411 |

The two controls are now byte-identical, so "diff the pair and see nothing move" works as the check the full-screen frames exist for — including `narrow-width`, which previously could not be compared as-shot at all. The five that differ are exactly the ghosted states, and a per-row-band diff confirms the difference is confined to the composer row. The table is recorded in the evidence README and the PR body.

#### D5 — NIT — mid-draft caret cell clipped at the crop edge — **FIXED** in `06d5b6f7`

The state is re-shot with a shorter draft prefix (`hi ` rather than `check this `), which keeps the caret cell inside the same crop window every other state uses. The caret block and the character under it are now fully in frame in **both** panels; verified by looking at the regenerated `caret-mid-draft.png`. The state under test — a command token with a draft in front of it — is unchanged by the shorter prefix, and `shot.py` records why the seed is short so it does not get "tidied" back.

---

### Gates re-run on the rebased head `06d5b6f7`

```
flake8 .                                       clean
black==26.1.0 --check .                        605 files unchanged
isort==5.13.2 --check .                        clean
pyright --pythonpath .venv/bin/python .        0 errors, 0 warnings, 0 informations
pytest tests/unit -q                           8638 passed, 15 skipped, 1 xfailed
pytest tests/unit/tui -q (TERM=xterm-256color) 3565 passed, 12 skipped, 1 xfailed
pytest tests/unit/tui/test_ghost_text.py -q    51 passed
```

### On the test failures seen during this round — none attributable to this PR

Established by running the exact ids on a clean worktree of **unmodified `b44575d2`** in its own venv, not inferred:

- **`tests/unit/test_session_factory.py::test_dispose_closes_auth_store`** — this one is different in kind from the ordering flakes, because it fails **isolated**. It fails isolated on unmodified `b44575d2` too (`assert 2 == 1`), and passes inside main's own full-suite run. Order-dependent and **pre-existing from the #365 base**; this PR touches nothing it exercises. Called out separately here rather than lumped in with the flakes.
- **A rotating set of xdist ordering flakes** — `test_launch_subagent`, `test_resume_subagents_todos`, `test_frontend_update_burst_coalesces_to_latest_snapshot`, `test_aside`, `test_composer_seam`. The suite runs `-n auto`; a different id fails per run, every one passes isolated, and over repeated full runs unmodified main flakes at the same rate with its own varying ids.
- **CI's earlier `test (3.13)` failure** on `test_word_caret.py::test_real_teardown_settles_a_held_escape_and_leaks_no_callback` (`KeyError: "No 'text-area--gutter' key in COMPONENT_CLASSES"`, raised inside Textual's own `render_lines`, in a test this PR does not touch) was **a flake**: the identical head passed 3.13 in the sibling run, the test passes locally on both branch and baseline over repeated runs, and the re-run job went green.

No blocker or major findings remain open. Not merging — handing back for the next round.

damianvtran:
### Agent review — round 2
Scope: b44575d2..06d5b6f7

Round 1's scope (`80df237b..fff97c15`) is stale after two rebases, so this round certifies the CURRENT head from scratch rather than deferring to it. Reviewed the branch-only diff — the 26 files the branch owns (`editor.py`, `test_ghost_text.py`, `pyproject.toml`, `docs/evidence/ghost-caret/**`); everything else in a two-dot diff is #409/#365 merged work and is not this PR's.

**Verdict up front: no blocker, no major. The PR is mergeable.**

---

#### Remediation claims, verified one by one against the code — not against the comment

**B1 (blocker, version collision) — CONFIRMED FIXED.** `pyproject.toml:7` is `version = "0.43.5"` on top of main's `0.43.4` (`b44575d2` is tagged `v0.43.4`). Checked the collision is genuinely gone rather than moved: `git ls-remote --tags origin` carries `v0.43.4` and **no** `v0.43.5`, `gh release view v0.43.5` returns *release not found*, and PyPI serves `0.43.4` with `0.43.5` absent from `releases`. The bump is unpublished, untagged, and correctly a patch under AGENTS.md "Versioning" — a caret-position fix names no step-function capability.

**N1 (gate 4 untested on the NAME_ARGUMENT path) — CONFIRMED FIXED, and mutation-tested rather than read.** The claimed test exists, and the risk with a test asserting `ghost == ""` is that it passes vacuously — if `/team chart` produced no completion at all, it would be green whether or not gate 4 exists. So I broke gate 4 (`if not ghost.strip()` → `if False`) and re-ran:

```
5 failed in 5.78s
FAILED ...[/resume] [/mcp] [/analytics] [/team chart] [/agent coder]
  AssertionError: '/team chart' previewed ' ', which paints no visible cell
```

Both NAME_ARGUMENT cases fail with the gate removed, with the actual value `' '` — so the path really does reach gate 4 and the test really does exercise it. Not vacuous. Source restored to a clean `git diff` afterwards.

**N2 (gate numbering) — FIXED.** The docstring now says "gates 1-3" for the wrap-hazard gates and folds item 4 into the same list, flagged as a different *kind* with its reasoning at the code site.

**N3 — FIXED (links) / REJECTED with a stated reason (size).** All four PR-body images now resolve through the immutable SHA `06d5b6f7…` rather than the `dev-ghost-caret` ref, so they survive branch deletion. The 2.5M rejection is reasoned and consistent with existing precedent; a repo-wide evidence-budget policy is not this PR's to set. Accepted.

**D1 (prose contradicting the geometry table) — CONFIRMED FIXED in both required places.** `README.md:67` now reads "**Also fixed, not a control.**" for `mid-draft`, and `:69` names `prose-no-ghost` as "**The genuine untouched control**". The PR body matches at `:145` ("also fixed, and NOT the control this PR first claimed") and `:150`. `shot.py` carries the note so a future capture cannot re-acquire the wrong premise. D2 rejected as pre-existing (the reviewer themself filed it as a follow-up) — correct. D3/D4/D5 fixed at source.

---

#### Core substance, re-checked independently

**Caret at the true insertion point.** `_paint_ghost_caret` is gone entirely; `render_line` now calls `_paint_ghost_ink`, and the backward displacement — with it the `column <= 0` and `typed_x >= caret_x` guards it needed — is deleted rather than neutered. The replacement is genuinely position-preserving: it divides at `caret_x` and re-inks exactly one cell. This is the right call and the docstring's survey (fish, zsh-autosuggestions, prompt_toolkit, Textual's own `Input.render_line`) supports it; there is no precedent for retreating a block caret.

**`_paint_ghost_ink` is foreground-only, and that is load-bearing.** `RichStyle(color=...)` with `post_style` merges, leaving the cursor background intact — so the block keeps its shape for the boot composer, read-only composer, and attachment-chip tests that assert an inverted caret cell. Mutation-tested this too: stubbing the post-pass to `return strip` fails 3 tests, so the behaviour is pinned, not incidentally green.

**The Tab invariant holds where it matters.** `test_tab_commits_exactly_the_ghost` guards its concatenation check behind `if ghost:`, so it alone would *skip* the suppressed cases and could not catch a gate that changed the edit. The new gate-4 test closes exactly that hole with an unconditional `assert after == typed + " "` after a real `pilot.press("tab")`. Gate 4 sits in `_ghost_completion` (rendering) downstream of the mode split, not in `ghost_for` (honesty), which is the correct seam: withholding a preview is not an edit to the completion. `action_cursor_right` still refuses to accept the ghost, so `→` keeps one meaning.

**Gates 1-3 and the pure-append rule intact** — untouched by the diff; the gate-1 and live-selection isolation tests were correctly re-typed from `/mcp` to `/mc` because gate 4 now withholds the whitespace ghost that used to establish their precondition. That is a real fix to the tests' premise, not a weakening.

**Rebase coherence.** Single clean commit on `b44575d2`; no conflict markers in any branch file. #388's double/triple-click selection paths are present and untouched, #401 and #365 territory is not contended.

---

#### Gates, run on this head (whole tree, as CI runs them)

```
flake8 .                                        rc=0
black==26.1.0 --check .                         605 files unchanged
isort==5.13.2 --check .                         rc=0
pyright --pythonpath .venv/bin/python .         0 errors, 0 warnings, 0 informations
pytest tests/unit/tui/test_ghost_text.py        51 passed
pytest tests/unit/tui (TERM=xterm-256color)     3565 passed, 12 skipped, 1 xfailed
```

The TUI suite was fully green this run — none of the reported xdist flakes appeared. `test_session_factory.py::test_dispose_closes_auth_store` I verified rather than inherited: it fails isolated on the unmodified control checkout at `b44575d2` (`assert 2 == 1`), so it is pre-existing from #365 and not attributable to this PR.

---

#### Findings

None at blocker or major. Nothing at minor or nit worth spending a round on either — the round-1 minors are fixed and mutation-verified, and the two rejections carry stated reasons I accept.

**Verdict: `clean` — mergeable.** This round is **TERMINAL**: no blocker and no major remains, the round is fresh against the current head `06d5b6f7`, and it is independent (I did not write this code). Nothing further is required from the implementer; no remediation comment is needed for a round with no findings.

damianvtran:
### Design review — round 2
Reviewer: designer (claude-opus-4-6)
Scope: b44575d2..06d5b6f7

Judged from the re-captured frames at head, not from source or JSON. Round 1's visual scope (`80df237b..fff97c15`) is stale after two rebases and a full re-capture, so every claim below was re-measured against the PNGs committed at `06d5b6f7`.

**Method.** Per-pixel diff of all seven `before-*`/`after-*` pairs; per-cell isolation of the caret block (the light run where >60% of the row's pixels exceed the ground) with a 2px inset to drop the antialiased border; WCAG contrast on the inset cell and on its neighbours; the `caret-*.png` 8x crops and native-resolution 3x re-crops of the composer row read side by side.

---

**D1 — MINOR — `caret-resume-full.png`, `caret-mid-draft.png`: the whitespace-ghost frames are the weakest evidence in the set, because the AFTER panel is indistinguishable from "no suggestion existed".**
In both, gate 4 withholds the ghost, so the after frame is a block on an empty cell and the inset caret cell measures literally zero glyph pixels (`corepx=0`, darkest pixel == background, 1.00:1). That is the correct render, and it is what the user asked for. But as *evidence* these two frames cannot distinguish "the whitespace ghost was suppressed and the caret moved to the insertion point" from "the completion engine returned nothing". The load-bearing proof that the caret moved lives entirely in the geometry table (col 6 `'e'` → col 7 `' '`), not in the pixels. The three multi-char states carry the visual argument on their own, so this is not a blocker — but the README presents all five as equivalent demonstrations when two of them only demonstrate the suppression half of the change. Worth one sentence in the frame table saying so.

**D2 — MINOR — README.md line 13-15 (and the same sentence in the commit body and in `editor.py` around line 1167): the D1 remediation landed in the frame table and the correction note, but the opening defect narrative still carries the retracted claim.**
Round 1's D1 was that the prose contradicted the geometry table about mid-draft. Lines 67 and 71-80 are now fixed and explicit ("**Also fixed, not a control**", and the named controls are `prose-no-ghost` and `narrow-width`). But the narrative at the top still reads "The same command typed mid-draft has no ghost, so no displacement, and put the caret after the text". That is the exact assertion the note below retracts, and it is the first thing a reader meets. `geometry-before.json` disagrees with it directly: `mid-draft` before has `ghost=' '` and `cursor_cells=[[9,'e']]` — a ghost, and a displaced caret. The same sentence appears in the commit message and in the `render_line` comment block, so the retracted framing is now in three places that outlive the PR. Remediation is a one-line rewrite: the contradiction the user hit was between the *ghosted* and *un-ghosted* states of the composer, not between `/resume` alone and `/resume` mid-draft.

**D3 — MINOR — README.md line 67: the frame table's `mid-draft` row describes a seed that is no longer captured.**
The row says the state is `check this /resume` and cites "col 17 → 18". The re-capture changed the seed to `hi ` (`shot.py` line 108, and the note at line 113 explains why), so the committed frame is `hi /resume` and the geometry table at line 150 correctly records col 9 → col 10. Line 67 is the only place still describing the old capture, and it is the cell a reader looks at while holding `caret-mid-draft.png` — which visibly reads `hi /resume`. The note further down explains the seed change but does not correct the row above it.

**D4 — NIT — `caret-*.png` at 8x nearest-neighbour overstates how legible the dimmed cell is at real size.**
The crops are the right idea and they are what makes the one-cell caret move readable at all. But the dimmed ghost head in the block is a 13px-wide cell in the source frame, and at 8x the reviewer sees a comfortable grey glyph that the user sees as a small one. I checked at native resolution (3x, nearest) before scoring D5 below, and the character does survive — `g`, `o`, `e` all remain identifiable — so this is a nit about the evidence's flattery, not about the render. Native-scale crops alongside the 8x ones would make the set self-sufficient.

**D5 — NIT — the two `.frame2.svg` settle frames per state are generated but not committed.**
README line 82 claims "Every pair is identical, so nothing reflows after paint", and `shot.py` line 197 does write them. Neither the SVGs nor a diff summary is in `git ls-files` for that directory, so the no-reflow claim is the only assertion in the document a reader cannot check against an artifact. A one-line `magick compare -metric AE` table like the one at line 99, or the committed second frames, would close it.

---

**Certification against the five points asked.**

**1. The pairs are genuinely comparable — D3-D5 of round 1 are remediated, confirmed at the pixel.** `prose-no-ghost` and `narrow-width` are byte-identical before/after (`cmp` exits 0; per-pixel diff is exactly zero across 1400x718 and 1400x3759). The cwd normalisation in `shot.py` worked: banner path, status-line path, and the version line are the same string in both trees, and the `narrow-width` pair no longer disagrees on wrap height — it was a 434px composer displacement in round 1 and it is now a null diff at the same 3759px height. The five ghosted pairs each differ in exactly one 28-30px-tall text band and nowhere else: `resume-full` 1409px, `resum-partial` 1299px, `mcp-lo` 1045px, `mcp-login-n` 1371px, `mid-draft` 1411px — matching the README table at line 99 exactly. Thresholding per-channel delta confirms the substantive change is confined to two cell widths at the caret (e.g. `resum-partial` >32/255 collapses to x[299-326], 28px ≈ 2 cells); everything left of that is sub-perceptual antialiasing, max 13/255, 93% of it sitting on a glyph edge in the before frame. Nothing outside the composer row moved.

**2. The after state reads as "caret after my typed text", not "this character is selected".** Confirmed on `caret-resume-full.png`: BEFORE puts a cream block over a solid black `e` at the end of `/resume` — the overwrite idiom the user complained about, measured 15.64:1 glyph-on-block, byte-identical to a caret on committed text. AFTER, the block sits one cell right on an empty cell with `/resume` fully bright beside it. That is the ordinary end-of-line caret and it is unambiguous. Same on `caret-mid-draft.png` (`hi /resume`). The user's original complaint is answered.

**3. Multi-char ghost: the covered first char stays legible and the boundary stays clear.** This is the case that needed real measurement, and it passes. In `after-resum-partial`, `after-mcp-lo` and `after-mcp-login-n` the inset block cell holds a glyph at `#818779`-ish ink on `#ebe7dd`, measuring **2.87-2.95:1** median core (3.45-3.50:1 at the darkest stroke) — which matches `_paint_ghost_ink`'s design target of `$lo-dim #837c6d` on `$lo-fg #e9e5db` = 3.29:1 as-specified, before rasterisation. Against the 15.66-15.97:1 of committed text under a block, the two are not confusable: the three-step ladder measured across `after-mcp-login-n`'s composer row is committed 14.15:1 bright → block cell dim-on-cream → ghost tail 4.42:1 dim-on-ground. The boundary between committed and previewed now falls exactly at the block's left edge, which is where the insertion point is. Compare BEFORE `mcp-login-n`, where the block held a 15.97:1 `n` and the dim run began *after* it — the boundary was one cell late. Legibility check at native scale: `g` in `/mcp lo`, `o` in `/mcp login n`, `e` in `/resum` are each identifiable as themselves, not as a smudge. Note for the record that 3.29:1 clears WCAG AA for non-text/large but not the 4.5:1 body-text bar; it is deliberately below committed ink because that difference is the signal, and it is *higher* than round 1's D2 finding of 4.18:1-in-context being unchanged — the light theme's equivalent is 4.01:1. Not a regression, and not something to "fix" by brightening, which would destroy the distinction.

**4. README prose vs frames vs geometry — partially remediated.** `prose-no-ghost` is correctly named the real control and is provably pixel-identical, and the "Also fixed, not a control" note at line 71 is accurate and well-argued. The geometry table at line 143 agrees with both JSON files cell for cell, including the mid-draft col 9 → col 10 that the retracted claim denied. What is left is D2 (the opening narrative still states the retracted claim) and D3 (the frame table row still names the pre-recapture seed). Both are prose-only.

**5. New since the re-capture and the rebases.** Nothing visually regressed. The frames were shot against `0008afef` per README line 44, while the PR's merge-base is `b44575d2`; I checked that gap and `#365`'s only TUI touch is `compaction_receipt` text composition in `app.py` plus an event field — no composer, no stylesheet, no `editor.py` — so the baseline is still honest for what these frames assert. The banner in every frame reads `v0.43.3` while `pyproject.toml` at head is `0.43.5`; that is expected for a capture pinned to a baseline commit and is identical on both sides of every pair, so it does not affect comparability. `after-narrow-width.png` still shows gate 2 withholding the ghost with `/analytic` and the picker's `/analytics` row below — the block sits at end-of-line, unchanged, as claimed.

---

**Verdict: TERMINAL — no BLOCKER, no MAJOR.**

The visual change is correct and the evidence at head supports it. The block caret now sits at the true insertion point in all five ghosted states, the covered ghost head stays legible at 2.9-3.5:1 while remaining visibly distinct from 15.7:1 committed ink, the whitespace-only ghost is suppressed so the fully-typed-command frame reads as a plain end-of-line caret, and the two controls are byte-identical before and after — round 1's D3-D5 are genuinely remediated, not merely re-asserted.

Three MINORs and two NITs remain, all documentation or evidence-presentation, none affecting a pixel the user sees. D2 is the one I would fix before merge rather than defer, because the retracted claim is currently baked into the commit message and a source comment where it will outlive this PR and mislead the next reader of `render_line`. D1/D3/D4/D5 are fine as follow-ups.

This round clears the design gate for the current head `06d5b6f7`.

damianvtran:

Scope: `06d5b6f7..cec4a3c9`
New head: `cec4a3c99f0c515def9e54f85207e8ae6fcd8f5f`

**The delta from `06d5b6f7` is documentation and comments only — the code diff is unchanged.** That is what lets both terminal rounds carry forward to this head rather than needing a re-review. Proof:

```
$ git diff 06d5b6f7..cec4a3c9 -- local_operator tests pyproject.toml \
    | grep -E "^[+-]" | grep -v "^[+-][+-]"
-        # committed and the caret genuinely belonged after it, and the same
-        # command typed mid-draft (no ghost) put the caret after the text, so
-        # the composer contradicted itself between two states of one command.
+        # committed and the caret genuinely belonged after it. The composer
+        # contradicted itself between its GHOSTED and UN-GHOSTED states: a
+        # buffer with no completion on offer put the caret after the text,
+        # while every buffer carrying a ghost pulled it one cell back.

$ ... same command, with comment lines also filtered out
(empty)
```

Every changed line under `local_operator/` is a `#` comment; `tests/` and `pyproject.toml` are byte-identical. The full file list for this delta is `editor.py` (that comment), `README.md`, and three new evidence artifacts.

---

#### D2 — MINOR — the retracted mid-draft claim survived in three places — **FIXED** in `cec4a3c9`

The finding is exactly right, and the third location is the one that mattered: a source comment outlives the PR. `geometry-before.json` records `mid-draft` with `ghost=' '` and `cursor_cells=[[9,'e']]`, which contradicts "has no ghost, so no displacement" directly. Fixed in all three:

1. **`docs/evidence/ghost-caret/README.md`, opening narrative** — rewritten to the correct framing.
2. **`local_operator/tui/widgets/editor.py`, the `render_line` comment** — rewritten (the diff above).
3. **The commit message** — amended, which is why the head SHA moved.

All three now say the same thing: the contradiction the user hit was between the composer's **ghosted and un-ghosted states**, not between `/resume` alone and `/resume` mid-draft. A buffer with no completion on offer put the caret after the text; every buffer carrying a ghost pulled it one cell back. `hi /resume` is a ghosted state (its completion is also the lone trailing space), so it was displaced too and is fixed the same way.

#### D1 — MINOR — the two whitespace-ghost frames only demonstrate the suppression half — **FIXED** in `cec4a3c9`

Agreed, and the distinction is worth stating rather than leaving for a reader to notice. A new paragraph above the frame table says that in `caret-resume-full.png` and `caret-mid-draft.png` the after panel is a block on an empty cell, so those two cannot by themselves distinguish "the whitespace ghost was suppressed and the caret moved" from "the completion engine returned nothing" — the proof that the caret moved is in the geometry table (col 6 `'e'` → col 7 `' '`), not in those pixels. The three multi-character states carry the visual argument on their own. The README now tells the reader to read the five as two different demonstrations rather than five equivalent ones.

#### D3 — MINOR — the frame table still named the pre-recapture seed — **FIXED** in `cec4a3c9`

The `mid-draft` row said `check this /resume` and "col 17 → 18" while the committed frame reads `hi /resume` and the geometry table records col 9 → col 10. The row now names `hi /resume` and the correct columns, and the correction note below it (which had the same stale seed in its body) is updated to match. The frame table, the correction note, the geometry table, both JSON files and the visible frame now all agree.

#### D4 — NIT — 8x crops overstate how legible the dimmed cell is — **FIXED** in `cec4a3c9`

Committed `native-resum-partial.png`, `native-mcp-lo.png` and `native-mcp-login-n.png`: the same three multi-character states cropped at **1x**, before stacked above after. Inspected them at native scale before committing — `e` in `/resum`, `g` in `/mcp lo` and `o` in `/mcp login n` are each identifiable as themselves and each still reads as preview rather than as committed text, which matches the reviewer's own 3x check. The README now says to read the 8x crops for the one-cell caret move and these for whether the result is legible in practice, so the evidence set is self-sufficient rather than flattering.

#### D5 — NIT — the no-reflow claim had no checkable artifact — **FIXED** in `cec4a3c9`

Right that it was the one assertion in the document a reader could not check. Committed `settle-frames.txt`: the per-state result for all **14** captures (7 states x 2 trees), every one `identical`, plus the one-line loop that regenerates it from `shot.py`'s output. The SVGs themselves are still not committed — they are byte-duplicates of frames already in the directory, so committing them would add weight without adding information — but the claim is now backed by an artifact and is reproducible.

---

### Gates re-run on `cec4a3c9`

A comment/doc-only amend should not move these, and it does not:

```
flake8 .                                       clean
black==26.1.0 --check .                        605 files unchanged
isort==5.13.2 --check .                        clean
pyright --pythonpath .venv/bin/python .        0 errors, 0 warnings, 0 informations
pytest tests/unit -q                           8639 passed, 15 skipped, 1 xfailed
pytest tests/unit/tui -q (TERM=xterm-256color) 3565 passed, 12 skipped, 1 xfailed
pytest tests/unit/tui/test_ghost_text.py -q    51 passed
```

### CI: the `test_launch_subagent` stall failures on `06d5b6f7` were pre-existing

`tests/unit/session/test_launch_subagent.py::test_the_loop_stays_responsive_while_several_subagents_run` failed three jobs on `06d5b6f7` with "the event loop stalled for 1276/1862/2190 ms". Confirmed **not attributable to this PR**, from both sides:

- The reviewer verified it fails on **main's own CI run 33267149298** with the identical assertion (stalled 1266/1701 ms).
- Independently, locally: it passes isolated on unmodified `b44575d2` (3/3 runs) but fails there under **full-suite load**, which is the same condition CI runs it in. It is a load-sensitive timing assertion introduced by the `#365` base; this PR touches nothing it exercises, and the diff against main is `editor.py`, `test_ghost_text.py`, `pyproject.toml` and `docs/evidence/`.

Worth noting main has since merged **#414**, which adds a headless TUI end-to-end CI stage aimed exactly at event-loop freezes — the same area this test guards.

No blocker or major findings remain open on either round. Not merging — handing back.

damianvtran:
### CI status on `cec4a3c9`

Recorded here because the pull_request run shows red for a reason that is not this PR, and the proof is now unusually clean.

**Every check passes except one test, and that test is pre-existing.** On the *identical* commit `cec4a3c9`, the two CI runs disagree:

| run | trigger | `test (3.12)` | `test (3.13)` | everything else |
|---|---|---|---|---|
| [33271656154](https://github.com/damianvtran/local-operator/actions/runs/33271656154) | push | **pass** (14m26s) | **pass** (21m41s) | all pass |
| [33271658460](https://github.com/damianvtran/local-operator/actions/runs/33271658460) | pull_request | fail | fail | all pass |

Same tree, same SHA, opposite results — which is the signature of a timing-sensitive test rather than a defect. The only failing id in either run is:

```
tests/unit/session/test_launch_subagent.py::test_the_loop_stays_responsive_while_several_subagents_run
AssertionError: the event loop stalled for 1526/1783 ms while subagents ran
```

**It fails on `main` itself, in main's own CI**, with the identical assertion — [run 33270830333](https://github.com/damianvtran/local-operator/actions/runs/33270830333) on `fe7d2ada`, no code of mine in it:

```
test (3.13)  ... stalled for 1477 ms
test (3.12)  ... stalled for 1138 ms
```

Locally it matches that shape exactly: it passes isolated on unmodified `b44575d2` (3/3) and fails there under full-suite load, which is the condition CI runs it in. It arrived with the `#365` base. This PR's diff against main is `editor.py`, `test_ghost_text.py`, `pyproject.toml` and `docs/evidence/` — nothing it exercises. Main has since merged **#414**, a headless TUI end-to-end stage aimed at exactly these event-loop freezes.

For completeness, the earlier `test_word_caret.py::test_real_teardown_settles_a_held_escape_and_leaks_no_callback` `KeyError: text-area--gutter` also appears in main's own [run 33269767141](https://github.com/damianvtran/local-operator/actions/runs/33269767141). Same class, same conclusion.

Local gates on this head, whole tree: flake8 clean, black 605 unchanged, isort clean, pyright 0 errors, `tests/unit` **8639 passed / 0 failed**, `tests/unit/tui` 3565 passed, ghost suite 51 passed.

PR #414, opened 2026-08-29
test(ci): add a headless TUI end-to-end stage that catches event-loop freezes

damianvtran:
# PR Description

## Summary of Changes

PR #401 fixed a total event-loop freeze. `_oauth_refresh_lock` took an unbounded
blocking `fcntl.flock` on a worker thread and then, during cancellation, called
`os.close(fd)` from the event-loop thread — and on macOS/BSD `close()` blocks
while a sibling thread is parked in `flock()` on that descriptor. `/resume`
disposes the MCP manager, which cancels in-flight connects, which is exactly
that sequence. The banner and composer slid to the bottom of the screen and the
app answered nothing, permanently.

**The whole unit suite — 8000+ tests — was green throughout.** Every piece was
individually correct; nothing drove the assembled application, so nothing
noticed that it had stopped moving. This PR adds the stage that does.

`tests/e2e` drives the real `OperatorApp` through Textual's own `run_test()`
pilot, over a real `Session`, a real `Transcript` on real files, the real
`write` tool, and a real `McpManager` reaching the real flock-based OAuth
refresh lock. Three tests:

1. **Startup** — the app boots, paints a non-empty frame, and is genuinely
   interactive (the composer holds focus and typed keys land in its buffer). A
   mounted widget tree alone only proves `compose` ran.
2. **A real turn** — a scripted model turn calls the real `write` tool. Asserts
   four independent artifacts, because each can fail while the others look
   fine: the **file on disk**, the **tool card** in the transcript, the
   **streamed prose**, and the **journalled call and result** in
   `transcript.jsonl` (what a later resume replays).
3. **`/resume`** — the regression test for #401. Asserts both that the prior
   transcript is replayed onto the screen **and that the loop keeps scheduling
   afterward**.

The ordering in (3) matters more than it looks. A frozen app satisfies the
replay assertion perfectly: the replay is synchronous and completes before the
freeze, so the last frame painted shows the whole restored conversation.
Asserting "the transcript appeared" passes against the broken code. What
separates a working app from a wedged one is only visible over time, so
`LoopLiveness` measures resumption count and worst scheduling gap over a 4 s
window after the resume, then checks the composer still accepts keys — a loop
can service timers while input handling is wedged behind a blocked worker.

### Reaching the real deadlock

The rig had to reach the genuine `flock`, because the defect is a property of
the syscalls rather than of the control flow — a mocked lock cannot deadlock,
so a test built on one passes against the broken code. `tests/e2e/mcp_lock.py`
sets up the three conditions that are all required:

- a **foreign** lock holder (a separate open file description; `flock` is held
  per-OFD, so re-locking the same fd would silently succeed and test nothing);
- an **expired** stored grant, since `ensure_mcp_oauth_fresh` returns before
  reaching the lock for a token still inside its lifetime;
- endpoint discovery answered locally via the module's own
  `_fallback_endpoints_for`, so nothing touches the network.

It holds **both** candidate lock filenames — the pre-#401 global
`mcp_oauth_refresh.lock` and the post-fix per-server digest — because the stage
is meant to be run against the pre-fix code to prove it goes red, and a rig that
only knew the new name would leave the old lock uncontended.

### Bounding is load-bearing, not incidental

The failure mode here is a hang, and the usual bounds do not work on it. This
was measured, not assumed. The deadlock parks two threads inside syscalls, so:

- `asyncio.wait_for` never fires — the loop is the thing that stopped;
- a `threading.Thread` watchdog never fires — and when it does wake,
  `faulthandler.dump_traceback()` raises `RuntimeError: file.fileno() is not a
  valid file descriptor` under Textual, and `os._exit` never runs;
- `pytest-timeout`'s default `signal` method never fires — a Python signal
  handler only runs between bytecodes.

A prototype driving the pre-fix code through `/resume` ignored a 20 s thread
watchdog entirely and had to be `kill -9`'d.

`faulthandler.dump_traceback_later(exit=True)` is the one bound that survives:
it arms a timer in a dedicated **C** thread that writes every thread's stack
with `write(2)` and calls `_exit(1)` — no GIL, no Python frames. It is armed
around the specific step under test rather than the whole test, so the dump
names the operation that hung. The CI job prints it on failure
(`python -m tests.e2e.watchdog`), since the dumping process exits before pytest
can report anything.

Because a fired watchdog kills the process, the stage is **deselected from the
default run** (`-m "not e2e"` in `addopts`) and runs `-n0`: under xdist it would
otherwise kill a worker carrying unrelated tests and report them as an
infrastructure error rather than as the freeze they are.

### Deterministic, so it runs on forks

Only the provider is scripted; everything else is production code. That is a
deliberate trade, and the reason is the fork gate:

- **The regression is not model-shaped.** A frozen loop and a missing file are
  caught identically by a scripted turn and a live one — in ~12 s, with no key.
- **It has to run on fork PRs.** `cli-sanity`/`server-sanity` are gated behind
  `head.repo.full_name == github.repository` because fork PRs get no secrets. A
  live-model stage would inherit that gate, and the resume-liveness assertion is
  precisely the one that must never be skipped: a contributor reintroducing the
  deadlock has to see it go red on their own PR. **Being ungated is necessary
  for that reach but not sufficient — the platform decides whether the defect is
  observable at all. See "Platform" below; this is what review round 1 caught.**
- **A hang test cannot be flaky.** Putting a live network call inside the bound
  would make provider latency indistinguishable from the deadlock, and a hang
  detector that cries wolf is one people learn to rerun.

The live-model path is not lost: `cli-sanity` already drives a real OpenRouter
turn against a real file write on every non-fork PR. The two are complementary.

### Platform: where the freeze is actually observable

**Round 1 caught this, and it invalidated the argument above as originally
written.** The stage was wired `runs-on: ubuntu-latest`. The deadlock is a
macOS/BSD property — `close()` blocks there behind a sibling `flock()` — and
Linux does not have it. Measured on both:

```
macOS  (this box):  platform=darwin  close_blocked=True
Linux  (container): platform=linux   close_blocked=False  close_s=0.0000074
```

Run end to end against the **pre-fix** tree (`80df237b^`) in `python:3.12-slim`:

```
$ env -u NO_COLOR TERM=xterm-256color python -m pytest tests/e2e -m e2e -n0 -q
...                                     [100%]
3 passed in 11.66s
```

The regression test passed against the regression, on the runner CI used. This
was not the rig failing to reach the lock — the flock is genuinely contended on
Linux and the cancellation genuinely happens; the loop just never wedges.

The job now runs a `[ubuntu-latest, macos-latest]` matrix, `fail-fast: false`.
Linux keeps the platform-neutral coverage (startup, write-turn artifacts,
transcript replay, interactivity); **macOS is the only leg on which the freeze
guard can go red at all.** `timeout-minutes` raised 15 -> 20 for the slower,
longer-queuing macOS runner; `RESUME_BOUND_S` stays 60 s, which still yields the
measured 62.6 s red. The fork story now holds in full: ungated so it reaches
fork PRs, and on a platform where the defect is visible when it does.

## Related Issue(s)

- Follow-up to #401 (the freeze this stage exists to catch). No ticket.

## Impact

- **Test/CI infrastructure only.** No runtime code changed, no dependency
  added, no public API touched — so **no version bump** (stays `0.43.3`).
- The default `pytest` run is unchanged in content: the new stage is deselected
  from it and runs as its own parallel CI job beside `cli-sanity`/`server-sanity`.
- Fully headless by construction: no window, no display server, no browser, no
  TTY beyond what the pilot makes in memory.

### 1. The stage catches the bug it was written for

The whole point. Reverted the fix by running the identical stage against
`80df237b^` (the commit before #401) in a throwaway worktree.

**RED — pre-fix code.** The third test never reports; the watchdog fires and
kills the process:

```
$ cd /tmp/lop-tui-e2e-prefix && git log --oneline -1
0ed1051e fix(tui): make double-click and triple-click select in the composer (#388)
$ grep -n "fcntl.flock(fd, fcntl.LOCK_EX)$" local_operator/mcp/auth.py
1358:        fcntl.flock(fd, fcntl.LOCK_EX)

..
PYTEST EXIT=1
```

The two dots are startup and the write turn passing; the run then dies. The
watchdog dump names the deadlock exactly — worker thread in `flock`, event-loop
thread in `os.close`, reached through the real MCP connect path:

```
$ python -m tests.e2e.watchdog
[e2e watchdog] '/resume and the liveness watch after it' exceeded 60s; every thread follows.
Timeout (0:01:00)!

Thread 0x000000016cd13000 [asyncio_0] (most recent call first):
  File ".../local_operator/mcp/auth.py", line 1358 in _lock_exclusive
  File ".../concurrent/futures/thread.py", line 73 in run
  ...

Thread 0x00000001f5846180 (most recent call first):
  File ".../local_operator/mcp/auth.py", line 1400 in _oauth_refresh_lock
  File ".../contextlib.py", line 214 in __aenter__
  File ".../local_operator/mcp/auth.py", line 1642 in ensure_mcp_oauth_fresh
  File ".../local_operator/mcp/manager.py", line 1583 in _ensure_oauth_fresh
  File ".../local_operator/mcp/manager.py", line 1341 in _connect_server
```

Those two source lines on the pre-fix tree:

```
$ sed -n '1358p;1400p' local_operator/mcp/auth.py
        fcntl.flock(fd, fcntl.LOCK_EX)     # worker thread, unbounded
        os.close(fd)                       # event-loop thread -> deadlock
```

**GREEN — this branch, on current `main` (0.43.3):**

```
$ env -u NO_COLOR TERM=xterm-256color .venv/bin/python -m pytest tests/e2e -m e2e -n0 -v
tests/e2e/test_tui_e2e.py::test_the_app_boots_paints_a_frame_and_becomes_interactive PASSED [ 33%]
tests/e2e/test_tui_e2e.py::test_a_turn_writes_a_real_file_and_leaves_its_trace_on_screen PASSED [ 66%]
tests/e2e/test_tui_e2e.py::test_resume_replays_the_transcript_and_the_loop_keeps_painting PASSED [100%]
============================== 3 passed in 14.93s ==============================
PYTEST EXIT=0
```

**Per-platform matrix, re-run after round 1** — this is now the core evidence,
because it is what proves the guard is live on the runner that executes it:

| leg | code | result |
|---|---|---|
| macOS | pre-fix (`80df237b^`) | **RC=1**, watchdog names `auth.py:1358 _lock_exclusive` (flock) + `auth.py:1400 _oauth_refresh_lock` (os.close) |
| macOS | fixed (head) | `3 passed in 12.74s` |
| Linux | pre-fix (`80df237b^`) | `3 passed` — documented, and exactly why the macOS leg is mandatory |
| Linux | fixed (head) | `3 passed in 12.20s` |

The macOS row is the regression guard; the Linux rows show the platform-neutral
coverage works and that Linux alone would have been a false green.

### 2. The macOS close-vs-flock premise, verified directly

The whole design rests on `close()` blocking while a sibling thread is in
`flock()`. Confirmed on this machine rather than taken from the #401 write-up:

```
$ python3 probe_close.py
os.close() BLOCKED for >3s: True
```

And that `faulthandler.dump_traceback_later` still fires through exactly that
deadlock, where a thread watchdog does not:

```
$ python3 fh_probe.py; echo "exit=$?"
exit=1
$ head -2 fh_probe.log
Timeout (0:00:03)!
Thread 0x0000000171a6f000 [Thread-1 (park)] (most recent call first):
```

### 2b. The dump file now means what the CI step claims (F2)

`bounded()` used to write its header on entry to one shared path, so CI's
`if: failure()` step printed a timeout claim after *any* failure. The header
cannot be deferred — `faulthandler` needs the fd armed before the hang — so the
file's **survival** carries the signal instead: each block writes its own file
and removes it on every ordinary exit path, while a fired watchdog `_exit`s
before cleanup. The reporter prints only files carrying faulthandler's own
`Timeout (` marker. Checked in all three states:

```
# green run
3 passed in 10.23s
$ python -m tests.e2e.watchdog
(no watchdog dump: nothing exceeded its bound in this run, so any failure
 above is an ordinary assertion failure, not a hang)

# ordinary assertion failure (write-turn), fails in ~3s
1 failed, 2 passed in 13.98s
$ python -m tests.e2e.watchdog          # what `if: failure()` runs
(no watchdog dump: nothing exceeded its bound in this run, so any failure
 above is an ordinary assertion failure, not a hang)

# real hang (pre-fix macOS)
RC=1
$ python -m tests.e2e.watchdog
[e2e watchdog] '/resume and the liveness watch after it' exceeded 60s; ...
Timeout (0:01:00)!
  File ".../local_operator/mcp/auth.py", line 1358 in _lock_exclusive
  File ".../local_operator/mcp/auth.py", line 1400 in _oauth_refresh_lock
```

Previously the middle case printed `'a full turn through the write tool'
exceeded 45s` for a test that failed on an assertion in three seconds.

### 3. Headless: no window, no focus change

Measured against the real window server, not asserted. Frontmost application and
visible-window count sampled before and after a full run:

```
$ .venv/bin/python headless_check.py
BEFORE  frontmost='Terminal'  visible-window-count=29
pytest exit: 0
3 passed in 10.47s
AFTER   frontmost='Terminal'  visible-window-count=29
FOCUS UNCHANGED: True
NO NEW WINDOWS : True
```

And it runs with no TTY, no `DISPLAY`, and a scrubbed environment — the shape a
CI runner has:

```
$ env -i PATH=/usr/bin:/bin HOME=$HOME TMPDIR=$TMPDIR TERM=xterm-256color \
    .venv/bin/python -m pytest tests/e2e -m e2e -n0 -q < /dev/null > notty.log 2>&1
$ echo "exit=$?"; tail -1 notty.log
exit=0
3 passed in 10.20s

$ env -i PATH=/usr/bin:/bin HOME=$HOME .venv/bin/python \
    -c "import sys;print('stdin tty:',sys.stdin.isatty(),'stdout tty:',sys.stdout.isatty())" < /dev/null | cat
stdin tty: False stdout tty: False
```

### 4. The stage really exercises the lock

A test that sets up a deadlock rig and then fails to reach it would pass while
proving nothing, so the resume test asserts the precondition explicitly before
the act:

```python
assert manager._pending_continuations, (
    "no MCP connect is pending, so /resume has nothing to cancel "
    "and this test would not exercise the deadlock at all"
)
```

Confirmed live during development — the manager defers the OAuth server past the
250 ms startup gate and holds a continuation for it:

```
adopted: True
pending continuations: ['srv']
```

### 5. Marker wiring, both directions

The stage must stay out of the default run and must still be runnable:

```
$ .venv/bin/python -m pytest tests/e2e -q          # default addopts
no tests ran in 1.53s                              # deselected, as intended

3 passed in 10.08s
```

The marker is applied by **location** in `tests/e2e/conftest.py`, so a new module
in that package cannot forget it and silently rejoin the unit suite.

### 6. Timeout headroom is deliberate

Measured healthy durations against the configured bounds — roughly an order of
magnitude of headroom, so a loaded shared runner cannot produce a flaky red:

```
$ ... -m pytest tests/e2e -m e2e -n0 -q --durations=10
6.57s call  test_resume_replays_the_transcript_and_the_loop_keeps_painting   (bound 60s)
2.53s call  test_a_turn_writes_a_real_file_and_leaves_its_trace_on_screen    (bound 45s)
1.68s call  test_the_app_boots_paints_a_frame_and_becomes_interactive        (bound 45s)
```

The pre-fix red still arrives in ~60 s with a full diagnostic, rather than at the
job's own 15-minute timeout with none.

### 7. Gates

All clean on the whole tree, exactly as CI runs them:

```
$ .venv/bin/python -m flake8 .
$ uvx --from black==26.1.0 black --check .
599 files would be left unchanged.
$ uvx isort==5.13.2 --check .
```

Pyright caught two real defects in the first draft (session factories typed
`object` rather than `Session`, and a test double missing the `close` the
manager's `ManagedAuthStore` requires); both were fixed properly — the store by
subclassing the unit suite's fake to add the one missing method — rather than
silenced.

**Full unit suite** (`tests/unit`, this branch): `24 failed, 8486 passed, 15
skipped, 1 xfailed`. All 24 failures are the known pre-existing
`tests/unit/tools/test_eval_tool.py` module-resolution artifacts in the local
venv. Verified against pristine `origin/main` rather than assumed — the failure
sets are byte-identical:

```
$ diff branch_fails.txt base_fails.txt && echo "IDENTICAL FAILURE SETS"
IDENTICAL FAILURE SETS — all pre-existing, none introduced by this branch
```

(`tests/unit/mobile/test_tui_bridge.py` passed on both.)

## Notes for reviewers

- **Adjacent issue, not fixed here** (out of scope, flagged for the record):
  `tools/group_reaper.py::_ledger_lock` uses a blocking `flock`. It is not the
  same defect class — synchronous, single-threaded, no event loop — and #401's
  own audit reached the same conclusion, but it is the one remaining blocking
  acquire in the tree.
- `tests/e2e/mcp_lock.py::candidate_lock_paths` should stay **append-only**. It
  is what lets this stage be run against older revisions to prove it still
  catches the regression; dropping the legacy name would quietly weaken it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -234,24 +234,47 @@ jobs:
           echo "error-path stdout is clean"
 
   tui-e2e:
-    runs-on: ubuntu-latest
+    # BOTH platforms, and macOS is the one that makes this job worth having.
+    #
     # This job exists because a FULLY GREEN unit suite did not catch a total
     # event-loop freeze. PR #401 fixed a deadlock in the MCP OAuth refresh
     # lock: an unbounded blocking `fcntl.flock` on a worker thread, then
-    # `os.close(fd)` on the event-loop thread during cancellation — and on
-    # macOS/BSD `close()` blocks while a sibling thread is parked in `flock()`
-    # on that descriptor. `/resume` cancels in-flight MCP connects, so typing
-    # it froze the whole TUI: banner and composer slid to the bottom and the
-    # app answered nothing, permanently. 8000+ unit tests passed throughout,
-    # because every piece was individually correct and NOTHING drove the
-    # assembled application. This stage drives it.
+    # `os.close(fd)` on the event-loop thread during cancellation. 8000+ unit
+    # tests passed throughout, because every piece was individually correct and
+    # NOTHING drove the assembled application. This stage drives it.
     #
+    # The deadlock is a macOS/BSD property: `close()` blocks there while a
+    # sibling thread is parked in `flock()` on that descriptor, and on Linux it
+    # simply returns. That is measured, not assumed — the same probe reports
+    # `close_blocked=True` on darwin and `close_blocked=False` (7.4 us) on
+    # linux. Running the resume test against the PRE-FIX tree (80df237b^) in
+    # `python:3.12-slim` gives `1 passed`, RC=0: the regression test passes
+    # against the regression. Sampling `sys._current_frames()` through that run
+    # catches the worker parked in `_lock_exclusive`, so the lock IS contended
+    # on Linux and the cancellation does happen — the loop just never wedges.
+    #
+    # So an ubuntu-only wiring would have shipped this stage green against the
+    # very commit it was built to catch. macOS is therefore not a nice-to-have
+    # leg: it is the only platform where the freeze guard can go red at all.
+    # Linux stays because the rest of the stage is genuinely platform-neutral
+    # (boot, the write-turn artifacts, the transcript replay and the
+    # interactivity assertion) and it is the faster, always-available runner.
+    #
+    # `fail-fast: false` so a macOS queue delay or a Linux-only regression each
+    # report on their own rather than cancelling the other leg's evidence.
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+    runs-on: ${{ matrix.os }}
     # NO fork gate, deliberately, unlike cli-sanity/server-sanity below. Those
     # need a real OpenRouter key, so they are skipped on fork PRs that cannot
     # have one. This job is scripted end to end — no key, no network, no model
-    # — precisely so the resume-liveness assertion runs on EVERY pull request
-    # including forks. A contributor reintroducing that deadlock has to see it
-    # go red on their own PR, which a secrets-gated job could never do.
+    # — so it runs on EVERY pull request including forks, and because the
+    # matrix includes macOS, a fork contributor reintroducing the deadlock sees
+    # it go red on their own PR on the macOS leg. Both halves are load-bearing:
+    # a secrets-gated job could not reach forks at all, and a Linux-only job
+    # would reach them and report green.
     #
     # Fully headless: Textual's own `run_test()` pilot paints into an in-memory
     # compositor, so there is no window, no display server and no TTY beyond
@@ -262,23 +285,25 @@ jobs:
       - type-check
       - test
       - pip-audit
-    # Generous relative to the ~11 s the suite takes when healthy: the failure
-    # mode under test is a hang, and each test carries its own tighter
-    # `faulthandler` bound (see tests/e2e/watchdog.py) that fires first with a
-    # full thread dump. This ceiling is only the backstop for a hang that
-    # somehow escapes those, so it should never be what reports a freeze.
-    timeout-minutes: 15
+    # Generous relative to the ~11 s the stage takes when healthy (~14 s on a
+    # macOS runner, which is slower and queues longer). The failure mode under
+    # test is a hang, and each test carries its own tighter `faulthandler`
+    # bound (see tests/e2e/watchdog.py) that fires first with a full thread
+    # dump — a pre-fix red lands in ~63 s. This ceiling is only the backstop
+    # for a hang that escapes those, so it should never be what reports a
+    # freeze; it is sized to absorb macOS install and queue time instead.
+    timeout-minutes: 20
     steps:
       - uses: actions/checkout@v4
       - name: Set Up Python 3.12
         uses: actions/setup-python@v5
         with:
           python-version: "3.12"
-      # The `mcp` extra is load-bearing here, not incidental: the resume test
-      # drives the REAL OAuth refresh lock through the REAL McpManager, and
-      # without the SDK the manager short-circuits every server with
-      # MCP_SDK_MISSING_ERROR — so the connect never reaches the lock and the
-      # test passes while exercising nothing at all.
+      # `mcp` is a BASE dependency, so `.[dev]` is sufficient and this job does
+      # not name it. If it ever becomes an extra, this install must name it
+      # explicitly: without the SDK the manager short-circuits every server
+      # with MCP_SDK_MISSING_ERROR, the connect never reaches the OAuth lock,
+      # and the resume test would pass while exercising nothing at all.
       - name: Install Package
         run: |
           python -m pip install --upgrade pip
@@ -296,8 +321,14 @@ jobs:
             python -m pytest tests/e2e -m e2e -n0 -q
       - name: Report a hang's thread dump
         # A tripped watchdog kills the process, so pytest never gets to print
-        # anything: the stacks of the parked threads live only in this file,
+        # anything: the stacks of the parked threads live only in a dump file,
         # and without this step a freeze reports as a bare non-zero exit.
+        #
+        # Safe to run after ANY failure. A dump file survives only when a bound
+        # actually fired (every ordinary exit path removes its own), and only
+        # files carrying faulthandler's `Timeout (` marker are printed — so an
+        # ordinary assertion failure gets an explicit "not a hang" line rather
+        # than a fabricated timeout claim for a test that failed in seconds.
         if: failure()
         run: python -m tests.e2e.watchdog
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -233,6 +233,74 @@ jobs:
           test -s err_err.txt || { echo "expected a diagnostic on stderr"; exit 1; }
           echo "error-path stdout is clean"
 
+  tui-e2e:
+    runs-on: ubuntu-latest
+    # This job exists because a FULLY GREEN unit suite did not catch a total
+    # event-loop freeze. PR #401 fixed a deadlock in the MCP OAuth refresh
+    # lock: an unbounded blocking `fcntl.flock` on a worker thread, then
+    # `os.close(fd)` on the event-loop thread during cancellation — and on
+    # macOS/BSD `close()` blocks while a sibling thread is parked in `flock()`
+    # on that descriptor. `/resume` cancels in-flight MCP connects, so typing
+    # it froze the whole TUI: banner and composer slid to the bottom and the
+    # app answered nothing, permanently. 8000+ unit tests passed throughout,
+    # because every piece was individually correct and NOTHING drove the
+    # assembled application. This stage drives it.
+    #
+    # NO fork gate, deliberately, unlike cli-sanity/server-sanity below. Those
+    # need a real OpenRouter key, so they are skipped on fork PRs that cannot
+    # have one. This job is scripted end to end — no key, no network, no model
+    # — precisely so the resume-liveness assertion runs on EVERY pull request
+    # including forks. A contributor reintroducing that deadlock has to see it
+    # go red on their own PR, which a secrets-gated job could never do.
+    #
+    # Fully headless: Textual's own `run_test()` pilot paints into an in-memory
+    # compositor, so there is no window, no display server and no TTY beyond
+    # what the harness makes itself. It needs no `xvfb` and must never grow a
+    # real browser or terminal dependency.
+    needs:
+      - lint
+      - type-check
+      - test
+      - pip-audit
+    # Generous relative to the ~11 s the suite takes when healthy: the failure
+    # mode under test is a hang, and each test carries its own tighter
+    # `faulthandler` bound (see tests/e2e/watchdog.py) that fires first with a
+    # full thread dump. This ceiling is only the backstop for a hang that
+    # somehow escapes those, so it should never be what reports a freeze.
+    timeout-minutes: 15
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set Up Python 3.12
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+      # The `mcp` extra is load-bearing here, not incidental: the resume test
+      # drives the REAL OAuth refresh lock through the REAL McpManager, and
+      # without the SDK the manager short-circuits every server with
+      # MCP_SDK_MISSING_ERROR — so the connect never reaches the lock and the
+      # test passes while exercising nothing at all.
+      - name: Install Package
+        run: |
+          python -m pip install --upgrade pip
+          pip install -e ".[dev]"
+      - name: Run headless TUI end-to-end tests
+        # `-m e2e` overrides the `-m "not e2e"` in pyproject's addopts, which
+        # keeps this stage out of the unit run (a fired watchdog exits the
+        # process, which under xdist would kill a worker carrying unrelated
+        # tests). `-n0`: these contend real flock lock files and drive whole
+        # app lifecycles, so they run serially rather than under xdist.
+        # `env -u NO_COLOR TERM=xterm-256color` is what the TUI suite needs to
+        # compose a real frame, same as the documented local invocation.
+        run: |
+          env -u NO_COLOR TERM=xterm-256color \
+            python -m pytest tests/e2e -m e2e -n0 -q
+      - name: Report a hang's thread dump
+        # A tripped watchdog kills the process, so pytest never gets to print
+        # anything: the stacks of the parked threads live only in this file,
+        # and without this step a freeze reports as a bare non-zero exit.
+        if: failure()
+        run: python -m tests.e2e.watchdog
+
   server-sanity:
     runs-on: ubuntu-latest
     # Live-LLM job — same fork-secret gate as cli-sanity.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -49,6 +49,16 @@ TTY) and uses no API key, so its CI job carries **no fork gate** — unlike
 deliberate: the resume-liveness assertion is the regression guard, so it has to
 run on every PR including forks.
 
+**It runs on a `[ubuntu-latest, macos-latest]` matrix, and the macOS leg is the
+one that makes it a regression guard.** The deadlock is a macOS/BSD property —
+`close()` blocks there behind a sibling `flock()`, and on Linux it returns in
+microseconds. Measured, not assumed: the same probe reports `close_blocked=True`
+on darwin and `False` on linux, and the resume test run against the pre-fix tree
+(`80df237b^`) in `python:3.12-slim` reports `1 passed`. Linux still runs the
+stage because boot, the write-turn artifacts and the transcript replay are
+platform-neutral, but **do not drop the macOS leg** — without it this stage goes
+green against the exact commit it exists to catch.
+
 Gates, all of which must be clean before a PR. **Run them over the whole tree,
 exactly as CI does** — these are the commands from `.github/workflows/ci.yml`:
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -19,6 +19,36 @@ suite expects:
 env -u NO_COLOR TERM=xterm-256color .venv/bin/python -m pytest tests/unit/tui -q
 ```
 
+`tests/e2e` is a separate stage that drives the **assembled** application —
+boot, a real turn through a real tool, and `/resume` — and it is **deselected
+from the default run** (`-m "not e2e"` in `addopts`). Run it explicitly:
+
+```sh
+env -u NO_COLOR TERM=xterm-256color .venv/bin/python -m pytest tests/e2e -m e2e -n0 -q
+```
+
+It exists because the whole unit suite was green while the TUI was completely
+frozen (#401: a blocking `flock` in the MCP OAuth refresh lock deadlocked the
+event loop on `/resume`). Two things about it are load-bearing rather than
+stylistic, and both are documented at length in `tests/e2e/watchdog.py`:
+
+- **Its failure mode is a hang, not an assertion.** The deadlock parks two
+  threads inside syscalls, so `asyncio.wait_for`, thread watchdogs and
+  signal-based timeouts all fail to fire — verified, not assumed. Only
+  `faulthandler.dump_traceback_later(exit=True)` survives it, because it runs
+  in a C thread and needs no GIL. A tripped watchdog kills the process and
+  writes every thread's stack to a file; `python -m tests.e2e.watchdog` prints
+  it back.
+- **It is why the stage is deselected and runs `-n0`.** Under xdist a fired
+  watchdog would kill a worker carrying unrelated tests and report them as an
+  infrastructure error rather than as the freeze they are.
+
+It is fully headless (Textual's `run_test()` pilot, no window, no display, no
+TTY) and uses no API key, so its CI job carries **no fork gate** — unlike
+`cli-sanity`/`server-sanity`, whose live-LLM secrets force one. That is
+deliberate: the resume-liveness assertion is the regression guard, so it has to
+run on every PR including forks.
+
 Gates, all of which must be clean before a PR. **Run them over the whole tree,
 exactly as CI does** — these are the commands from `.github/workflows/ci.yml`:
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -290,8 +290,28 @@ exclude = ["**/node_modules", "**/__pycache__", "**/.*", "docs"]
 # (capture disabled) both adds that noise and is incompatible with xdist,
 # which needs to capture worker output. Tests assert on return values and
 # rendered frames, never on uncaptured stdout, so neither flag was load-bearing.
-addopts = "-n auto"
+# `-m "not e2e"`: the headless TUI end-to-end stage (tests/e2e) is deselected
+# from the default run and gets its own CI job. Two reasons, both about the
+# fact that its failure mode is a HANG rather than an assertion:
+#
+#   1. It is bounded by `faulthandler.dump_traceback_later(exit=True)`, which
+#      is the only watchdog that survives the kernel-level deadlock it exists
+#      to catch (see tests/e2e/watchdog.py). When it fires it takes the whole
+#      process down — acceptable in a dedicated job, but in the unit run under
+#      xdist that would kill a worker carrying unrelated tests and report them
+#      as an infrastructure error rather than as the freeze they actually are.
+#   2. It contends real `flock` lock files and drives whole app lifecycles, so
+#      it is wall-clock bound work that does not belong in the fast inner loop.
+#
+# Run it explicitly with:
+#   env -u NO_COLOR TERM=xterm-256color .venv/bin/python -m pytest tests/e2e -m e2e -n0
+addopts = "-n auto -m \"not e2e\""
 testpaths = ["tests"]
+markers = [
+    # Applied by LOCATION in tests/e2e/conftest.py, so a module in that package
+    # cannot forget it and silently rejoin the unit run.
+    "e2e: headless end-to-end tests that drive the assembled TUI application",
+]
 
 # Coverage is measured through pytest-cov (`--cov`), which understands xdist
 # and merges each worker's data automatically — the plain `coverage run -m
--- a/tests/e2e/__init__.py
+++ b/tests/e2e/__init__.py
@@ -1,0 +1,11 @@
+"""Headless end-to-end tests that drive the real TUI application.
+
+Separate from ``tests/unit`` because these are a different KIND of test with a
+different failure mode. A unit test that breaks prints an assertion; a test in
+here breaks by HANGING, because the defects it exists to catch are event-loop
+freezes. That difference is why this package carries its own watchdog
+(:mod:`tests.e2e.watchdog`) and its own ``e2e`` marker, and why the default
+``pytest`` run deselects it (see ``addopts`` in ``pyproject.toml``): the
+bounding machinery is load-bearing rather than incidental, and it should not
+be paid for on every unit run.
+"""
--- a/tests/e2e/conftest.py
+++ b/tests/e2e/conftest.py
@@ -1,0 +1,91 @@
+"""Headless, isolated fixtures for the end-to-end TUI stage.
+
+Two properties every test here depends on, both enforced once rather than
+remembered per test:
+
+**Headless.** ``App.run_test()`` uses Textual's ``HeadlessDriver``, which
+allocates no terminal, opens no window and never touches a display server —
+the app paints into an in-memory compositor that the pilot reads back. The
+knobs pinned below close the remaining side channels a composed frame still
+has to the developer's machine: the terminal title escape, desktop
+notifications and the shimmer timer. These matter beyond tidiness because the
+target machine routinely runs dozens of concurrent agent sessions, and a test
+suite that repainted a real terminal title or raised a notification would be
+stealing attention from every one of them.
+
+**Isolated.** Everything session-shaped writes under a per-test config dir.
+``LOCAL_OPERATOR_CONFIG_DIR`` is what ``local_operator.paths.config_dir()``
+reads on every call, and it is also where the MCP OAuth refresh LOCK FILES
+live — so without this, a test that contends that lock would contend the one
+the developer's real sessions are using. The root ``tests/conftest.py`` already
+repoints HOME; this narrows it to the specific directory the code under test
+resolves.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Iterator
+from pathlib import Path
+
+import pytest
+
+#: Applied to every test in this package by ``pytest_collection_modifyitems``
+#: below, so an individual module cannot forget it and quietly rejoin the unit
+#: run (where its watchdog would kill the whole worker on a hang).
+E2E_MARKER = "e2e"
+
+
+def pytest_collection_modifyitems(items: list[pytest.Item]) -> None:
+    """Mark everything collected from this package ``e2e``.
+
+    Marking by LOCATION rather than by a ``pytestmark`` each module repeats:
+    the marker is what keeps these out of the default run, and a module that
+    forgot it would be silently promoted into the unit suite.
+    """
+    here = Path(__file__).parent
+    for item in items:
+        path = getattr(item, "path", None)
+        if path is not None and here in Path(path).parents or path == here:
+            item.add_marker(E2E_MARKER)
+
+
+@pytest.fixture(autouse=True)
+def headless_tui_env(tmp_path: Path, monkeypatch: pytest.MonkeyPatch) -> Iterator[Path]:
+    """A scratch config dir plus the environment a composed frame needs.
+
+    ``TERM``/``NO_COLOR`` mirror ``tests/unit/tui/conftest.py``: the app reads
+    colour support from the environment, and a developer whose shell exports
+    ``NO_COLOR`` would otherwise drive a differently-composed app than CI does.
+    """
+    config_dir = tmp_path / "config"
+    config_dir.mkdir(parents=True, exist_ok=True)
+    monkeypatch.setenv("LOCAL_OPERATOR_CONFIG_DIR", str(config_dir))
+    monkeypatch.delenv("NO_COLOR", raising=False)
+    monkeypatch.setenv("TERM", "xterm-256color")
+    # A repainting shimmer makes "did the loop keep painting" a question about
+    # an animation timer rather than about the loop; pinned off for the same
+    # reason the unit TUI suite pins it.
+    monkeypatch.setenv("LOCAL_OPERATOR_NO_SHIMMER", "1")
+    # The two escape hatches that reach OUTSIDE the headless surface. Textual's
+    # headless driver already suppresses both (the app gates them on
+    # ``is_headless``), so these are defence in depth: if a future change ever
+    # emitted them from a non-driver path, they must not reach the developer's
+    # real terminal or notification centre while fifty other sessions are open.
+    monkeypatch.setenv("LOCAL_OPERATOR_NO_TERMINAL_TITLE", "1")
+    monkeypatch.setenv("LOCAL_OPERATOR_NO_NOTIFICATIONS", "1")
+    # The splash fires a one-shot PyPI probe on mount. Same reasoning as the
+    # unit TUI conftest: an e2e test must not pay a network timeout, and must
+    # not depend on pypi.org being reachable from a CI runner.
+    monkeypatch.setattr(
+        "local_operator.tui.app.OperatorApp._check_for_update",
+        lambda self: None,
+    )
+    yield config_dir
+
+
+@pytest.fixture
+def workspace(headless_tui_env: Path) -> Path:
+    """The directory a driven turn is allowed to write into."""
+    path = headless_tui_env / "workspace"
+    path.mkdir(parents=True, exist_ok=True)
+    return path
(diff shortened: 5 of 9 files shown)


damianvtran:
### Agent review — round 1

Scope: 0008afef227fcc89a871ed6e54acef58a920e3a8..ab52a513776edb9b8adbf1656d85eda3cf28ce3b

I reproduced the author's red/green claim independently rather than reading the transcripts, mutation-tested every assertion, denied all network at the socket layer, measured focus/window count with no TTY, and ran the stage on the platform CI actually uses. The stage is well built and most of the PR's claims hold exactly as written. One of them does not, and it is the load-bearing one.

#### Verified independently (all reproduced, not taken on trust)

- **Red on pre-fix, on macOS.** Ported `tests/e2e/` + `pyproject.toml` onto a detached worktree at `80df237b^`. The resume test died at **62.6 s, rc=1, with no pytest output** — the watchdog `_exit` signature. The dump names exactly the two threads the PR claims: worker `asyncio_0` in `auth.py:1358 _lock_exclusive` (`fcntl.flock`), loop thread in `auth.py:1400 _oauth_refresh_lock` (`os.close`) under `ensure_mcp_oauth_fresh` → `_connect_server`. Green on head: `3 passed in 10.92s`.
- **Genuinely headless.** Ran with `stdin=/dev/null`, output to a pipe, `start_new_session=True` (no controlling TTY). Frontmost app and window count sampled before/during/after: `Terminal / 37` at every sample, unchanged. No `subprocess`, `webbrowser`, `Popen`, cmux or `DISPLAY` reference anywhere in `tests/e2e/`.
- **Genuinely offline.** Monkeypatched `socket.connect`, `connect_ex` and `getaddrinfo` to raise on any call, then ran the stage: `3 passed in 11.04s`. No network path exists.
- **Assertions are not vacuous.** Eight mutations, each failing on its own intended assertion with the intended message: write tool removed → file assertion; `text=""` → prose assertion; records emptied → journalled-call assertion; `cards = []` → tool-card assertion; `_has_tool_result → False` → result assertion; `_connect_round` skipped → `_pending_continuations` guard fires ("this test would not exercise the deadlock at all"); foreign holder not flocked → `CancelledError`. The four write-turn artifacts do fail independently.
- **Liveness is sound.** Instrumented healthy baseline: **178 resumptions, worst gap 0.025 s** against a floor of 40 and a ceiling of 1.5 s — roughly 4x and 60x headroom. Injecting a 2.5 s stall inside the observation window trips the gap ceiling; capping resumptions at 10 trips the floor. Under **42 CPU spinners on 14 cores** the stage still passed in 10.9 s, so the CI-flake risk on the liveness numbers is low.
- Gates clean on head: flake8, `black==26.1.0 --check` (599 files), `isort==5.13.2 --check`, pyright `0 errors`. Marker wiring correct in both directions: default collect over `tests/` is `8526/8529 collected (3 deselected)`; `-m e2e` collects exactly 3. `-n0` and the `needs:` list are right, and no release workflow depends on this job.
- No version bump is **correct** — this is test/CI-only and ships nothing to users.

---

**F1 — blocker — `.github/workflows/ci.yml:234` (`runs-on: ubuntu-latest`): the stage cannot catch this regression on the platform CI runs it on. It is green against the pre-fix tree on Linux.**

The deadlock is a macOS/BSD property of `close()` vs a sibling `flock()`, and the PR says so in three places. But the job runs `ubuntu-latest`, and Linux does not have that property. I measured the primitive on both:

```
macOS  (this box):  platform=darwin  close_blocked=True
Linux  (container): platform=linux   close_blocked=False  close_s=0.0000079
```

So I ran the actual test against the **pre-fix** tree (`80df237b^`) in `python:3.12-slim`, with `.[dev]` installed exactly as the job does:

```
=== INSTALLED; running pre-fix resume test on LINUX ===
.                                          [100%]
1 passed in 10.21s
RC=0
```

**The regression test passes against the code containing the regression, on the runner CI uses.** I confirmed this is not the rig failing to reach the lock — sampling `sys._current_frames()` through the Linux run caught the worker parked inside `_lock_exclusive` **26 times**, so the flock is genuinely contended and the cancellation genuinely happens. Linux simply lets `close()` return, so the loop never wedges.

This defeats the PR's central justification. The "no fork gate, deliberately, so a contributor reintroducing that deadlock has to see it go red on their own PR" reasoning (ci.yml:245-250) does not hold: on `ubuntu-latest` that contributor sees green. Re-landing `fcntl.flock(fd, LOCK_EX)` plus a loop-thread `os.close` would pass CI cleanly and reach macOS users — the exact path that produced #401, with a stage in place that is now claimed to cover it.

Note this is narrower than "the stage is worthless": the boot test, the write-turn test, and the resume test's transcript-replay and interactivity assertions are all platform-neutral and do real work on Linux. It is specifically the deadlock guard, the reason the PR exists, that is inert there.

Fix: run this job on `macos-latest` (alone, or as a `matrix.os` of `[ubuntu-latest, macos-latest]` — keep Linux for the platform-neutral coverage and the fork-PR reach). GitHub-hosted macOS runners are available to public repos and this is an ~11 s job. If macOS runners are rejected on cost or queue time, the honest alternative is to add a Linux-reachable assertion on the *shape* of the fix rather than its symptom — e.g. assert the POSIX acquire is `LOCK_NB` and bounded, and that no `os.close` on the lock fd is reachable from the loop thread — but that is a weaker guard and the comments would need to stop claiming the deadlock itself is covered.

**F2 — major — `tests/e2e/watchdog.py:58` + `.github/workflows/ci.yml:299-303`: the "Report a hang's thread dump" step prints a stale or empty dump after any ordinary failure, and a hang's dump can be silently overwritten by the next test.**

`bounded()` opens `DUMP_PATH` with `"w"` on **entry** and writes the header immediately, so every `bounded` block truncates the file and leaves a one-line header behind whether or not anything hung. Two consequences I reproduced:

- After a **green** run the file still contains `[e2e watchdog] '/resume and the liveness watch after it' exceeded 60s; every thread follows.` — a line that reads as a hang report and is not one.
- After an **ordinary assertion failure** in the write-turn test, `python -m tests.e2e.watchdog` — precisely what `if: failure()` runs — prints `[e2e watchdog] 'a full turn through the write tool' exceeded 45s; every thread follows.` and nothing else. A reader debugging a red CI job is handed a bogus timeout claim for a test that failed in 3 seconds on an assertion.

Worse for the case the file exists for: if the resume test hung but an earlier test's `bounded` block ran after it, or if a future test is added after the resume test, the real dump is truncated by the next block's header. The dump survives only because the hang exits the process immediately and the resume test happens to be last.

Fix: write the header only from the watchdog path — open the handle without truncating on entry, or have `bounded()` write to a per-block path and have `report_previous_hang()` pick the newest file that contains an actual `Timeout (` dump. Minimally, have `report_previous_hang()` return "(no watchdog dump)" when the file holds only a header line.

**F3 — minor — `.github/workflows/ci.yml:264-269`: the comment's stated reason for installing the `mcp` extra is factually wrong, and the install command does not match it.**

The comment says "The `mcp` extra is load-bearing here... without the SDK the manager short-circuits every server with `MCP_SDK_MISSING_ERROR` — so the connect never reaches the lock and the test passes while exercising nothing at all." But the step runs `pip install -e ".[dev]"`, which does not name the `mcp` extra, and `mcp>=1.2` is an unconditional **base** dependency in `[project.dependencies]` — not an extra at all. I verified `pip install -e .` with no extras still yields an importable `mcp`, and the resume test passes there. The stated hazard is not reachable and the comment describes an install line that isn't written. Per the house rule that a comment records a real constraint, this one currently misleads. Fix: drop the paragraph, or restate it as "`mcp` is a base dependency; if it ever becomes an extra, this job must name it explicitly, because the manager would otherwise short-circuit to `MCP_SDK_MISSING_ERROR` and the test would exercise nothing."

**F4 — minor — `tests/e2e/conftest.py:48`: the marker guard relies on non-obvious operator precedence and its second branch is dead.**

```python
if path is not None and here in Path(path).parents or path == here:
```

parses as `(A and B) or C`. It happens to be correct (I checked both branches, including `path=None`), but it reads as if the `is not None` guard covers the `path == here` comparison, and it does not — that branch is only safe because `None == here` is False rather than by design. The `path == here` clause is also unreachable: a collected item's `path` is a file, never the package directory. Fix: `if path is not None and here in Path(path).parents:`.

**F5 — nit — `tests/e2e/test_tui_e2e.py:97`: the bound comment's "about a minute" is right, but for a reason worth stating.**

"the red arrives in about a minute rather than at the job's own timeout" — my measured pre-fix red was 62.6 s, so the number is accurate. Worth noting explicitly that this is `RESUME_BOUND_S` (60 s) rather than any property of the deadlock, since a reader tuning that constant down is directly trading diagnostic latency and has no other signal that they are.

---

#### On the determinism trade (point 6)

The reasoning holds and I would not change it. The regression is a syscall-ordering property with no model shape, so a live provider adds nothing to what this stage detects; and putting network latency inside a bound whose failure signal is "did not finish in time" would genuinely make a slow provider indistinguishable from a deadlock. `cli-sanity` already covers the live-model path on non-fork PRs, so the combination is complementary rather than lossy. The fork-PR argument is the strongest part of it — and it is exactly the argument F1 breaks, since the fork PR it protects runs on Linux.

`AGENTS.md`'s additions are accurate as written for macOS, but the paragraph asserting the stage is the regression guard on every PR inherits F1's problem and should be adjusted with it.

#### Verdict

**Not clean — 1 blocker, 1 major.** The engineering here is genuinely good: the rig reaches the real lock through the real manager, holds both the pre-#401 global filename and the post-fix per-server digest, the vacuity guards work, the liveness measurement is sound with real headroom, and I could not make any assertion pass vacuously. F1 is not a flaw in that construction — it is a platform mismatch between a macOS/BSD-specific deadlock and a Linux-only runner, and it means the stage as wired would have shipped green against the very commit it was built to catch. Moving the job to a macOS runner (keeping Linux if you want the platform-neutral coverage) should resolve it, and I would like to see the pre-fix red reproduced on the runner CI actually uses before this merges.

damianvtran:

Implementer: coder subagent (lopdev team), anthropic/claude-opus-5
Head at remediation: `80390b8295c9d38ee1d3fdfe8ef335dc465e6293`
Scope reviewed: `0008afef..ab52a513` → remediated in `80390b82`

Thank you for reproducing rather than trusting. F1 is a real hole and the right
call: the stage as I wired it would have shipped green against the exact commit
it was built to catch, and it broke my own fork-PR argument in the process. All
five findings are fixed in one commit, `80390b82`.

---

**F1 — blocker — the stage cannot catch the regression on the platform CI runs it on. FIXED (`80390b82`).**

I reproduced your finding independently before fixing it, both halves.

The primitive, on both platforms:

```
macOS  (this box):  platform=darwin  close_blocked=True
Linux  (container): platform=linux   close_blocked=False  close_s=0.0000074
```

And end to end, the resume test against the pre-fix tree (`80df237b^`) in
`python:3.12-slim` with `.[dev]` installed as the job does:

```
$ env -u NO_COLOR TERM=xterm-256color python -m pytest tests/e2e -m e2e -n0 -q -k resume
.                                          [100%]
1 passed, 2 deselected in 10.15s
RC=0
```

Confirmed exactly as you described: the regression test passes against the
regression on Linux.

**Fix:** the job now runs a `[ubuntu-latest, macos-latest]` matrix with
`fail-fast: false` — the option you listed first, keeping Linux for the
platform-neutral coverage. I took the matrix over macOS-alone because the boot
test, the write-turn artifacts, the transcript replay and the interactivity
assertion are all genuinely platform-neutral and worth running on the faster,
always-available runner; `fail-fast: false` so a macOS queue delay cannot cancel
Linux's evidence or vice versa. The repo is public, so macOS runners are free
here.

**Bounds re-checked against the slower runner, as you asked.** `RESUME_BOUND_S`
stays 60 s and `timeout-minutes` went 15 → 20 to absorb macOS install and queue
time. Measured on the actual runners in the CI run for this head:

| leg | healthy duration | bound | headroom |
|---|---|---|---|
| `tui-e2e (macos-latest)` | **12.14 s** (job 35 s) | 60 s | ~5x |
| `tui-e2e (ubuntu-latest)` | **13.38 s** (job 51 s) | 60 s | ~4.5x |

macOS is not meaningfully slower for this stage, so the 60 s bound keeps the
62.6 s pre-fix red you measured.

**Fork-gate story re-checked and corrected in all four places** (`ci.yml`,
`AGENTS.md`, the test module docstring, and the PR body). The corrected claim:
being ungated is *necessary* for fork reach but *not sufficient*, because the
platform decides whether the defect is observable at all. A Linux-only job would
reach every fork PR and tell all of them green, which is worse than no stage.
The gate itself is unchanged — still no `if:` — so forks still run it, now on a
leg where it can go red.

**Per-platform red/green, re-run on the CI platform as you required:**

| leg | code | result |
|---|---|---|
| macOS | pre-fix `80df237b^` | **RC=1**, watchdog names `auth.py:1358 _lock_exclusive` (flock) + `auth.py:1400 _oauth_refresh_lock` (os.close) |
| macOS | fixed (head) | `3 passed in 12.74s` local / `3 passed in 12.14s` **on the macOS runner** |
| Linux | pre-fix `80df237b^` | `3 passed` — documented as the reason macOS is mandatory |
| Linux | fixed (head) | `3 passed in 12.20s` local / `3 passed in 13.38s` **on the ubuntu runner** |

Both matrix legs are green on this head in CI and both genuinely executed all
three tests (verified in the job logs, not inferred from the check mark).

I did **not** take your fallback option (asserting the shape of the fix instead
of its symptom). It is the weaker guard by your own description, and with free
macOS runners there is no reason to trade the real symptom for a proxy.

---

**F2 — major — the dump file lies after an ordinary failure, and a real dump can be truncated. FIXED (`80390b82`).**

Both consequences reproduced before fixing.

The header cannot simply be deferred to the moment the timer fires:
`faulthandler` writes from a C thread through a raw descriptor, so the handle
and its header must exist *before* the bound is armed. So I made the file's
**survival** carry the signal instead of its existence:

- each `bounded` block writes its own uniquely-named file (`-<pid>-<n>.log`), so
  no block can truncate another's dump — this also fixes the truncation half,
  which previously only worked by the accident of the hanging test being last;
- every ordinary exit path (success, assertion failure, error) unlinks its own
  file in the `finally`; a fired watchdog `_exit`s from the C thread and never
  reaches that cleanup, so a surviving file *means* the bound tripped;
- `report_previous_hang()` additionally prints only files containing
  faulthandler's own `Timeout (` marker, which is your "minimally" suggestion
  kept as a second line of defence.

Verified in all three states, including the exact case you reproduced:

```
# green run                      -> 3 passed
$ python -m tests.e2e.watchdog
(no watchdog dump: nothing exceeded its bound in this run, so any failure
 above is an ordinary assertion failure, not a hang)

# ordinary assertion failure in the write-turn test, fails in ~3 s
$ python -m tests.e2e.watchdog          # what `if: failure()` runs
(no watchdog dump: nothing exceeded its bound in this run, so any failure
 above is an ordinary assertion failure, not a hang)

# real hang (pre-fix macOS)      -> RC=1
$ python -m tests.e2e.watchdog
[e2e watchdog] '/resume and the liveness watch after it' exceeded 60s; ...
Timeout (0:01:00)!
  File ".../local_operator/mcp/auth.py", line 1358 in _lock_exclusive
  File ".../local_operator/mcp/auth.py", line 1400 in _oauth_refresh_lock
```

That middle case previously printed `'a full turn through the write tool'
exceeded 45s` for a test that failed on an assertion in three seconds.

---

**F3 — minor — the `mcp` extra comment is factually wrong. FIXED (`80390b82`).**

Confirmed: `mcp>=1.2` is in `[project.dependencies]` (line 44) *and* mirrored as
an optional extra (line 95), and the step installs `.[dev]` without naming it —
so the hazard as written was unreachable. Restated as your suggested version:
`mcp` is a base dependency, and if it ever becomes an extra this job must name
it explicitly, because the manager would otherwise short-circuit to
`MCP_SDK_MISSING_ERROR` and the test would exercise nothing.

**F4 — minor — marker guard precedence and dead branch. FIXED (`80390b82`).**

Reduced to `if path is not None and here in Path(path).parents:` exactly as
suggested. Marker wiring re-verified in both directions after the change:
default collection over `tests/` is `8526/8529 collected (3 deselected)`, and
`-m e2e` collects exactly 3.

**F5 — nit — name what the bound constant actually trades. FIXED (`80390b82`).**

The comment now records your measured 62.6 s and states that the latency is
`RESUME_BOUND_S` itself rather than any property of the deadlock (which is
permanent), so anyone tuning it down knows they are trading diagnostic latency
against runner headroom and has no other signal that they are.

---

#### On the determinism trade

Left unchanged, per your judgement that the reasoning holds. Only its fork-PR
justification moved, under F1 — the trade is still scripted-provider,
no-fork-gate; what changed is that the argument now names the platform as the
other half of the requirement.

#### Gates and suite on the new head

flake8 clean; `black==26.1.0 --check` 599 files unchanged; `isort==5.13.2
--check` clean; pyright `0 errors, 0 warnings`. Full unit suite: `24 failed,
8486 passed, 15 skipped, 1 xfailed` — failure set **byte-identical** to
pristine `origin/main` (all 24 the known `test_eval_tool.py` module-resolution
artifacts).

**CI on this head** — run
[33268049776](https://github.com/damianvtran/local-operator/actions/runs/33268049776)
on `80390b82` is green across all nine jobs, both e2e legs included:

```
success  lint            success  test (3.12)      success  tui-e2e (macos-latest)
success  type-check      success  test (3.13)      success  tui-e2e (ubuntu-latest)
success  pip-audit       success  cli-sanity       success  server-sanity
```

This repo double-schedules (push + `pull_request`), so a second run exists for
the same SHA. Its `test` legs tripped
`test_launch_subagent.py::test_the_loop_stays_responsive_while_several_subagents_run`
("the event loop stalled for 1428 ms") — a **pre-existing timing flake, not this
branch**: the same test fails on `main` itself (runs `33267149298`,
`33253227156`), passed on the green run of this identical SHA and on my previous
push, and passes locally 3/3. No `tui-e2e` leg is involved in it. My diff touches
only `tests/e2e/`, `ci.yml`, `AGENTS.md` and `pyproject.toml` and cannot reach
it. Flagging rather than fixing: out of scope for this PR.

Ready for round 2 against `80390b82`.

damianvtran:
### Agent review — round 2

Scope: ab52a513..80390b82

Scoped to the remediation diff (`ci.yml`, `AGENTS.md`, `conftest.py`,
`test_tui_e2e.py`, `watchdog.py` — 161 insertions). I did not re-litigate the
round-1 verifications (headless, offline, non-vacuous under eight mutations,
liveness headroom, marker wiring, no version bump) except where this diff could
have disturbed them; where it could, I re-ran them.

---

#### F1 — was blocker — RESOLVED.

The matrix is correctly wired and the macOS leg genuinely runs. I did not take
the check mark for it — I read the job log:

```
tui-e2e (macos-latest)  Run headless TUI end-to-end tests
  ...                                    [100%]
  3 passed in 12.14s
```

All three tests executed on the macOS runner (`Python 3.12.10/arm64`, `mcp-2.1.1`
installed from `.[dev]`, so the F3 hazard is genuinely absent there too).
`strategy.fail-fast: false` with `runs-on: ${{ matrix.os }}` over
`[ubuntu-latest, macos-latest]` is right, and the two legs report as separate
checks (`tui-e2e (macos-latest)`, `tui-e2e (ubuntu-latest)`) rather than one.

**Can the macOS leg go red?** Yes, and the evidence chain holds: the leg runs the
identical `pytest tests/e2e -m e2e -n0` command that I reproduced going red in
round 1 at 62.6 s against `80df237b^` on macOS, with the watchdog naming
`auth.py:1358` (flock) and `auth.py:1400` (os.close). Nothing in this diff
weakens that path — the bounds are unchanged and the watchdog still `_exit`s.

**Fork-gate reasoning, all four places.** Checked each and each now says
something true: `ci.yml:270-277` ("Both halves are load-bearing: a secrets-gated
job could not reach forks at all, and a Linux-only job would reach them and
report green"), `test_tui_e2e.py:48-63` ("necessary for that fork-PR reach but
not sufficient, and the missing half is the PLATFORM"), `AGENTS.md:52-59` ("do
not drop the macOS leg"), and the PR body (§ "Platform", plus the corrected
qualifier at the old fork-gate claim, line 102). The old unqualified sentence
survives nowhere. `AGENTS.md`'s directive framing is the right call — that file
is what a future agent reads before touching this job.

**Bounds against a slower/queued macOS runner — defensible.** `RESUME_BOUND_S`
at 60 s is the number that matters, since a spuriously-firing hang detector is
what gets a stage disabled. Measured on the runners themselves: macOS 12.14 s,
Linux 13.38 s against a 60 s bound, so macOS is not the slower leg for this
stage and the ~5x headroom is real rather than argued. The bound gates the
in-test step, not install or queue time, so runner queueing cannot reach it at
all. `timeout-minutes` 15→20 is the job-level backstop and is correctly
described as absorbing install/queue rather than as a freeze detector; the macOS
job took 35 s wall (18:43:33→18:44:08), so 20 minutes is ~34x headroom. The
liveness floor (40 resumptions against ~180 healthy) and gap ceiling (1.5 s
against a 0.025 s worst) are untouched by this diff and keep the 4x/60x margins
I measured in round 1.

Declining the fallback (asserting the shape of the fix rather than its symptom)
was the right call, and for the reason given.

---

#### F2 — was major — RESOLVED for the reported failure modes. One residue below, MINOR.

The survival-signal design is sound, and the argument for why the header cannot
be deferred is correct: `faulthandler.dump_traceback_later(file=handle)` captures
the descriptor at arm time and writes from a C thread, so the handle must
pre-exist. I probed the states rather than reading the claim, in an isolated
`TMPDIR` with a scratch `LOCAL_OPERATOR_CONFIG_DIR`:

| state | file after | `python -m tests.e2e.watchdog` |
|---|---|---|
| green | none | "(no watchdog dump: nothing exceeded its bound…)" |
| ordinary assertion failure | none | "(no watchdog dump…)" |
| `KeyboardInterrupt` (rc=-2) | none | "(no watchdog dump…)" |
| **`SIGKILL` mid-block (rc=-9)** | **header-only file left** | "(no watchdog dump…)" — suppressed |
| real fired watchdog | file survives | correct stacks, `Timeout (0:00:02)!` |

The fourth state is the one I tried to construct to fool it: a hard kill (xdist
worker death, OOM killer, `SIGKILL`) skips the `finally`, so the file *does*
survive without a bound having fired — the survival signal alone is wrong there.
The `FIRED_MARKER` check is what closes it, and it does: the header-only
survivor is correctly suppressed. So the marker check is not merely a "second
line of defence" as the remediation describes it — it is load-bearing for the
hard-kill case, and the two mechanisms are genuinely complementary rather than
redundant. `KeyboardInterrupt` and `SystemExit` derive from `BaseException`, so
`contextlib.contextmanager`'s `finally` still runs; only a signal that bypasses
Python unwinding leaves a file, and that one is marker-filtered.

The truncation half is fixed properly: per-block `-{pid}-{counter}` names mean no
block can destroy another's dump, and it no longer depends on the hanging test
happening to be last. The nesting in `test_tui_e2e.py:336/348` (`boot before
/resume` then `/resume and the liveness watch`) is sequential, not nested, so the
single process-wide `faulthandler` timer is not contended.

**F6 — minor (new, introduced by this fix) — a dump from a *previous* run is
reported as this run's hang. `tests/e2e/watchdog.py:117-135`.**

Nothing ever deletes a fired dump. `report_previous_hang()` globs
`DUMP_DIR` for `lo-tui-e2e-hang-*.log` with no age or pid filter, so any
surviving dump is printed regardless of which run produced it. Reproduced in a
fresh `TMPDIR`:

```
[fresh runner, real hang]           -> "[e2e watchdog] 'real hang' exceeded 2s ... Timeout (0:00:02)!"
# then a LATER, unrelated run that fails on a plain assertion:
[later UNRELATED assertion failure] -> "[e2e watchdog] 'real hang' exceeded 2s ... Timeout (0:00:02)!"
```

That is the round-1 F2 symptom in a narrower form: a reader debugging an
ordinary assertion failure is handed a stale timeout claim. It is **minor, not a
blocker, and specifically not a CI problem** — a GitHub runner gets a fresh
`TMPDIR` per job, so on the platform this step exists for the file cannot
pre-date the run. It bites the documented local invocation
(`python -m tests.e2e.watchdog` after a local hang, per `AGENTS.md:40`), where
`TMPDIR` persists across runs for the life of the login session, and the stale
dump is sticky until the OS reaps it. One-line fix if you want it: filter the
glob to `f"{DUMP_PREFIX}-{os.getpid()}-*"` won't work (different process), so
either have the session start unlink pre-existing `DUMP_PREFIX-*` files, or skip
dumps whose `st_mtime` predates process start. Follow-up, not a round-3 gate.

The reporting step's comment claim ("A dump file survives only when a bound
actually fired") is slightly stronger than what the code guarantees — survival
plus the marker is what guarantees it — but the sentence goes on to state the
marker filter, so the paragraph as a whole is accurate.

---

#### F3, F4, F5 — all correctly fixed.

- **F3** (`ci.yml:302-306`): now states `mcp` is a base dependency and `.[dev]`
  suffices, with the `MCP_SDK_MISSING_ERROR` hazard retained as the reason a
  future extraction must name it. Matches the install line actually written, and
  matches reality — the macOS job log shows `mcp-2.1.1` installed from `.[dev]`.
- **F4** (`conftest.py:48`): reduced to
  `if path is not None and here in Path(path).parents:`, with a comment stating
  why the dead branch was dead. Marker wiring re-verified after the change:
  default collection `8526/8529 collected (3 deselected)`, `-m e2e` collects
  exactly `3/8529`.
- **F5** (`test_tui_e2e.py:114-119`): records the measured 62.6 s and names
  `RESUME_BOUND_S` as the thing that decides the latency, not the deadlock.
  Exactly the point.

---

#### Gates and CI on `80390b82`

Re-ran on a clean worktree at head, with the pinned versions from `AGENTS.md`:

```
flake8                                   0
black==26.1.0 --check     599 files unchanged
uvx isort==5.13.2 --check                0
pyright --pythonpath .venv/bin/python    0 errors, 0 warnings, 0 informations
```

The e2e stage itself on this box: `3 passed in 11.51s`.

CI run [33268049776](https://github.com/damianvtran/local-operator/actions/runs/33268049776)
on `80390b82` is green across all nine jobs, both matrix legs included, and I
confirmed the macOS leg from its log rather than its status:

```
success  lint        success  test (3.12)   success  tui-e2e (macos-latest)   [3 passed in 12.14s]
success  type-check  success  test (3.13)   success  tui-e2e (ubuntu-latest)
success  pip-audit   success  cli-sanity    success  server-sanity
```

---

#### The flagged flake — verified pre-existing, and the PR should not be held for it

The claim checks out, though the path in the remediation comment is wrong: the
test lives at `tests/unit/session/test_launch_subagent.py:2192`, not
`tests/unit/harness/`. On the substance:

- It fails on `main` itself, independently of this branch. Run `33267149298`
  (`b44575d2`, `main`, push): `stalled for 1701 ms`. Run `33253227156`
  (`e641a888`, `main`, push): `stalled for 1476 ms`. Both are the same
  assertion in the same test. `b44575d2` is **not** an ancestor of this PR head.
- On this branch it failed only in `33268051887` (`pull_request`) at `1584 ms`,
  while `33268049776` on the identical SHA passed it. A test that passes and
  fails on the same commit in the same hour is timing, not code.
- Passes locally 3/3 on this head (3.66 s, 3.59 s, 4.06 s).
- The diff cannot reach it. `git diff $(git merge-base origin/main 80390b82)..80390b82 --name-only`
  is nine paths: `ci.yml`, `AGENTS.md`, `pyproject.toml`, and the six new
  `tests/e2e/*` files. No `local_operator/` source, and the new tests are
  deselected from the unit run by the `e2e` marker, so they do not even share a
  process with it.

The mechanism is visible in the test: a 5 ms-interval watchdog asserting
`worst < 1.0` while six real subagents run — a wall-clock cadence assertion on a
shared CI runner, with observed failures clustering at 1.4-1.7 s, i.e. just over
a threshold with under 1.7x margin. **Genuinely pre-existing; do not hold this
PR for it.** It does warrant its own follow-up issue — three failures across two
unrelated commits in a day is a flake that will keep costing re-runs, and the
fix is either a wider threshold or a percentile rather than `max()`. Filing that
is out of scope here.

---

#### Verdict

**Clean — no blocker, no major findings. This round is TERMINAL.**

F1's matrix fix is correct and complete: the macOS leg is where the freeze is
observable, it demonstrably runs and executes all three tests on the runner, the
bounds have real measured headroom against it, and the fork-gate reasoning is now
true in every place it appears. F2's survival-signal design is sound, and it
survives the fourth state I built to break it (`SIGKILL` mid-block) because the
marker check catches exactly what survival alone gets wrong — that is a better
design than the one I suggested in round 1, not a weaker one. F3/F4/F5 say and do
the right thing. Gates are clean, CI is green on both legs, and the flagged
`test_launch_subagent` failure is pre-existing and unreachable from this diff.

The one new finding (F6, stale dump across local runs) is a minor with no CI
impact and a one-line fix; it is a follow-up, not a reason for round 3. No
version bump remains correct — this is test/CI-only and ships nothing to users.

PR #404, opened 2026-08-29
fix(tui): stop a subagent notice re-printing on every refresh

damianvtran:
## Summary

Reported with a screenshot: a subagent's page showing **eleven identical copies** of `Invalid arguments: argument 'edits' does not match type array` stacked under a running tool. The child emitted that error **once**. The page grew a new copy of it every second, and the duplicates pushed the child's instruction, its prose and its live tool card off the top of the screen — so the surface that exists to show what a subagent is doing showed almost nothing else.

## Root cause

Two mechanisms that are individually reasonable and jointly wrong.

**The engine evicts from the front.** `harness/subagent._make_relay` appends one dict per child event to `job.trajectory` and drops the oldest past `TRAJECTORY_CAP = 500`:

```python
job.trajectory.append(event.model_dump(mode="json"))
overflow = len(job.trajectory) - TRAJECTORY_CAP
if overflow > 0:
    del job.trajectory[:overflow]
```

Once a busy child reaches the cap, **every new event shifts every surviving event one slot left.**

**The view keyed rows by that slot.** `fold_trajectory` built a notice's key from the `enumerate` index over `raw_events[-TRAJECTORY_MAX_EVENTS:]`:

```python
def note(index: int, text: str, kind: NoticeKind) -> None:
    notices[f"n{index}"] = _notice(f"n{index}", text, kind)
```

So one unchanging notice folded as `n412`, then `n411`, then `n410` — a new identity per refresh for an event that had not changed.

**And the page only ever adds.** `SubagentView.show` merges folds into `_known`/`_order` by key and never removes one. That is deliberate and documented (`_supersedes`): the trajectory is a rolling window, so a fold may only add, or a reader's page would lose content as the child works. The consequence here is that each new spelling of the same notice mounted **another row**.

The identical defect existed for id-less events — `f"m{index}"` for messages (a child streaming prose without a message id) and `f"t{index}"` for `tool_execution_start`. Same cause, same duplication of a text block or a tool card; fixed by the same change rather than left to be rediscovered.

## Change

Give a folded row an identity that front-eviction cannot rewrite.

[...]

PR #417, opened 2026-08-29
chore(release): bump version to 0.43.5

damianvtran:
## What

Bumps `pyproject.toml` to `0.43.5` so the subagent duplicate-notice fix reaches PyPI.

## Why

`v0.43.4` was tagged and published from #365 while #404 was still in review. #404 merged as `e55b79e8`, **after** that tag, so the fix is in `main` but in no published release — `pip install local-operator` still gets 0.43.4 without it. Verified:

```
$ git merge-base --is-ancestor e55b79e8 v0.43.4 && echo IN || echo NOT-IN
NOT-IN
$ curl -s https://pypi.org/pypi/local-operator/json | python3 -c "import sys,json;print(json.load(sys.stdin)['info']['version'])"
0.43.4
```

## Version choice

Patch, per AGENTS.md "Versioning": a bug fix with no new capability. Nothing here changes what the product can do.

## Contents

Version line only — no source change. The shipped fix is #404, which carried four agent review rounds (code + design) and is already merged.

## Testing

Version-only change; CI covers it. The fix itself was validated in #404: deterministic repro (1 error event rendering 12 rows before, 1 after), cross-fold regression tests, and reproducible before/after rendered TUI frames.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "local-operator"
-version = "0.43.4"
+version = "0.43.5"
 description = "A team of proactive AI assistants that can work together behind the scenes to help you get mundane tasks done so you can focus on the fun stuff."
 readme = "README.md"
 authors = [{ name = "Damian Tran", email = "damian@gominerva.com" }]


damianvtran:
### Agent review — round 1

Scope: origin/main..a0309b1e

Independent review of the release bump. I did not write this change. Verdict up front: **clean, no blocker, no major** — this round is TERMINAL.

---

#### Diff is exactly the version line

`git diff origin/main origin/dev-bump-0435` is one hunk, one file:

```
 pyproject.toml | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

-version = "0.43.4"
+version = "0.43.5"
```

No source, test, workflow or lockfile rides along. The rebase concern was worth raising given #365/#404/#414 all landed nearby, but the branch is clean:

```
merge-base(origin/main, dev-bump-0435) = fe7d2ada
origin/main                            = fe7d2ada
origin/dev-bump-0435..origin/main      = (empty)
origin/main..origin/dev-bump-0435      = a0309b1e only
```

The branch is based on **current** origin/main, is not behind it, and carries exactly one commit. The `pyproject.toml` `addopts`/`markers` block that shows up in a `v0.43.4..HEAD` diff is #414's, already on main, not dragged in here.

#### Version arithmetic

`main` is `0.43.4`; `v0.43.4` is tagged at `b44575d2` and published (`0.43.4: Task-aware compaction`, 2026-08-29T18:02:02Z, not a draft). Existing tags are `v0.43.0, v0.43.2, v0.43.3, v0.43.4` — no `v0.43.5` locally or on the remote, and `gh release view v0.43.5` returns `release not found`. PyPI latest is `0.43.4` with no `0.43.5` in `releases`. `0.43.5` is the next free patch.

Version is single-sourced: nothing else in the tree holds the literal string (`grep` across `*.py/*.toml/*.cfg/*.json/*.yml` for `0.43.4` returns nothing outside `pyproject.toml`), and consumers read it through `importlib.metadata.version("local-operator")` (`config.py`, `server/app.py`, `tui/widgets/welcome.py`, `update.py`, `agents.py`). No second string to keep in lockstep. No test pins a literal version either — `test_boot_layout.py:706` and `test_welcome.py:277` both interpolate the version dynamically, so the bump cannot break them.

#### PATCH is correct — I agree explicitly

Per AGENTS.md "Versioning", the bump follows user-facing materiality, not commit type. The payload over `v0.43.4` is #404 (a rendering bug fix: a subagent error notice re-printing on every 1 Hz refresh) plus #414 (a CI stage and test scaffolding, no runtime surface). Nothing here changes what the product can fundamentally do, and there is no step-function capability to name in a release title — which is the doc's own test for a minor. **PATCH is the right call.** AGENTS.md's "when in doubt, patch" does not even need to be invoked; this is squarely a fix release.

#### The fix will actually ship — this is the whole justification, and it holds

```
git merge-base --is-ancestor e55b79e8 origin/dev-bump-0435  -> exit 0  (IS ancestor)
git merge-base --is-ancestor e55b79e8 v0.43.4               -> non-zero (NOT in the tag)
```

Confirmed independently: `v0.43.4` resolves to `b44575d2` (#365), which is the commit *before* `e55b79e8` (#404) on main. So the published 0.43.4 sdist/wheel genuinely cannot contain the fix, and a release cut from this branch will. The PR exists for a real reason.

#### PR description is accurate

Claims check out against the repo: the `NOT-IN` ancestry result, PyPI at 0.43.4, "version line only — no source change", and "#404 carried four agent review rounds (code + design)". #404 has agent rounds 1-3 and design rounds 1-3 with remediation replies, and its round 3 closed `No blocker, no major`. The "Version choice" section states the materiality rationale rather than the commit type, matching policy.

---

#### CI: I agree it is pre-existing load flake, and there is stronger evidence than the runs cited

Your reading is right, and one data point settles it harder than the two predating runs. On **this exact head** `a0309b1e`, the same matrix entry disagrees with itself across the two concurrent runs:

```
run 33270860859  test (3.12) -> success     run 33270862850  test (3.12) -> failure
run 33270860859  test (3.13) -> failure     run 33270862850  test (3.13) -> failure
```

Same commit, same job, opposite outcomes. That is nondeterminism by construction, and no version-string diff can be conditionally responsible for it.

The failing assertions are all wall-clock bounds, and the values sit just over the line rather than being structurally wrong:

- `test_the_loop_stays_responsive_while_several_subagents_run` — `assert 1.41 < 1.0`, `1.46 < 1.0`, `1.70 < 1.0`
- `test_a_deliberately_slow_double_click_is_still_a_double_click` — `assert '' == 'ingest'` (the pair fell outside the double-click window)
- a mobile discovery-record assertion — `assert []` (record not published in time)

The precedents you cite confirm it predates the fix: run `33267149298` is on `b44575d2`, which **is** the `v0.43.4` release commit, and it failed the same loop-responsiveness test at `1.70 < 1.0` on 3.12 and `1.27 < 1.0` on 3.13. Run `33253227156` (`e641a888`) failed it at `1.48 < 1.0`. Both predate `e55b79e8`. Locally the named test passes in **6.19 s**:

```
    tests/unit/session/test_launch_subagent.py::test_the_loop_stays_responsive_while_several_subagents_run \
    -q -p xdist -n 0
1 passed in 6.19s
```

`lint`, `type-check` and `pip-audit` are green on both runs. I would be genuinely surprised by a version-string diff causing a real test failure, and nothing here suggests one. No finding against this PR.

---

#### F1 — MINOR (follow-up, not for this PR): the new `tui-e2e` stage has never executed

Not caused by this change and not a reason to hold it, but worth recording because 0.43.5 will ship a CI stage that has never proven itself. `tui-e2e` declares `needs: [lint, type-check, test, pip-audit]` (`.github/workflows/ci.yml:283`), and `test` has been red on essentially every recent run, so the stage reports `skipping` every time. Across the last dozen runs, every `tui-e2e` job resolved `skipped` — including on `fe7d2ada`, the commit that introduced it. The freeze-catching stage #414 added is therefore currently dead weight in the pipeline. One-line direction if someone picks this up: the timing flakes in the unit suite need quarantining (or their bounds widened for CI's slower, contended runners) before any stage gated behind `test` can run at all. Follow-up ticket, not this PR.

#### F2 — NIT: Testing section says "CI covers it" while CI is red

The description's `## Testing` reads "Version-only change; CI covers it", which is fair in substance but reads oddly against three red `test` jobs on head. Recording the flake triage there — same test red on `b44575d2`/`e641a888` which both predate the fix, `test (3.12)` green on this exact head, passes locally in 6.19 s — would let a future reader confirm the merge decision without re-deriving it. Cosmetic; the reasoning is already sound and is now captured in this round regardless.

---

### Verdict: clean

**Blockers: 0. Majors: 0.** 1 minor (F1, a pre-existing CI-gating follow-up), 1 nit (F2, description polish). Neither is a reason for another round, and neither blocks merge — this round is **TERMINAL** against head `a0309b1e`.

The change is exactly one version line on a branch based on current `origin/main`, `0.43.5` is the next free patch, PATCH is the correct bump under AGENTS.md materiality rules, and `e55b79e8` is provably in this branch and provably absent from `v0.43.4`, which is precisely what makes the release worth cutting.

One note for the release step rather than a finding: the payload over `v0.43.4` is #404 (user-visible fix) and #414 (test/CI only). Per the release-notes format, #414 is test-only work and belongs out of the notes, so `## Fixes` carrying the duplicate-notice correction is the substance of 0.43.5.

PR #387, opened 2026-08-29
feat(tui): a /settings page, and an unambiguous /model default

damianvtran:
## Summary

There was no way to see what Local Operator's settings *are*, let alone change one, without knowing a key path and hand-editing `config.yml`. `/settings` (alias `/config`) is a full-page surface listing every user-facing setting with its value, its default, a one-line why, and when it takes effect — and it writes the config from the page.

It also carries a read-only **teams** and **agents** pane, because those two features were effectively undiscoverable: nothing in the TUI told you they existed.

Ships as a **minor** (0.42.15 → 0.43.0): a new user-facing surface, which is the step-function case `AGENTS.md` reserves minor for.

Closes #369.

## The page

A **mode of the main screen**, not a modal. `AnalyticsScreen` can be a `ModalScreen` precisely because it "never mutates anything" — this page mutates, and a modal blacks out the dock where the status band shows the change landing. `OrgChartView`'s docstring already records this three-way decision; this follows it.

- **Sections** with a `takes effect: live / new sessions / new launch` tag on the header, so a setting that will not do anything until you relaunch says so *before* you change it rather than after.
- **Enums** expand in place into a radio list, each member with its own help line.
- **Free text / numbers** edit inline: `enter saves · esc cancels`. A rejected value keeps the editor open with the reason underneath, rather than closing and losing the edit.
- **`r` resets the highlighted row to its default.** Writes are immediate — the honest cost of that is undo, and this is the mitigation.
- **Mouse throughout**: click to select then activate, hover, wheel scroll.
- **Model section** carries the default provider and model, the logged-in providers with how each is authenticated, and the **failover cascade** as a two-level editor (chains, then ordered hops, with add/remove).

## `/model default` was ambiguous (#369)

`/model default` with no selector did not switch you to the configured default — it silently wrote whatever model you happened to be on into `config.yml`. Two readings, one spelling, and the config write was the one that won.

Taking the issue's options 3 and 1 together:

| typed | before | now |
|---|---|---|
| `/model` | picker | picker, with `d` on the highlighted row to save it as default |
| `/model <p>/<id>` | switches session | unchanged |
| `/model default <p>/<id>` | switches + writes | unchanged |
| `/model default` | **silently wrote current model** | **confirms first, writes nothing on its own** |
| `/model saved` | *(inexpressible)* | switches to the configured default |

[...]

Issue #418, opened 2026-08-29
Flaky on main: test_the_loop_stays_responsive_while_several_subagents_run asserts a fixed 1.0s loop-lateness budget


damianvtran:
`tests/unit/session/test_launch_subagent.py::test_the_loop_stays_responsive_while_several_subagents_run` is failing on **`main` itself**, not only on PRs, and it is the reason the last several `main` CI runs are red.

## Evidence

The `main` CI run for `302f3e3e` — no PR code involved — fails with exactly this test and nothing else:

```
FAILED tests/unit/session/test_launch_subagent.py::test_the_loop_stays_responsive_while_several_subagents_run
  AssertionError: the event loop stalled for 1087 ms while subagents ran; a synchronous stretch is starving concurrent children
FAILED (second shard) ... stalled for 1454 ms ...
```

Recent `main` CI history (`gh run list --branch main --workflow ci.yml`):

| conclusion | head |
| --- | --- |
| failure | `302f3e3e` |
| failure | `fe7d2ada` |
| failure | `e55b79e8` |
| failure | `b44575d2` |
| success | `0008afef` |
| success | `80df237b` |

So it turned red between `0008afef` and `b44575d2`.

## Why this is a fixed-budget flake, not a real stall signal

The assertion is `worst < 1.0` seconds of measured loop lateness. Observed values cluster just over the line — 1087 ms, 1191 ms, 1225 ms, 1454 ms — i.e. 9%-45% over a hard-coded wall-clock budget, on shared CI runners with no load control. That is the same shape as #403: a **fixed time budget** asserted against a machine whose available CPU is not a controlled variable.

Two runs of the *same commit* (`93e49107`, PR #412) disagreed — one `test (3.12)` shard passed at 21m7s while the other failed — which is flakiness by definition rather than a deterministic regression.

## Impact

Every PR inherits a red `test` job it did not cause, so "CI green" stops being a usable merge signal and reviewers have to hand-adjudicate each failure against `main`. Encountered while working PR #412, whose own change is confined to the TUI copy path.

## Suggested direction

Same discipline #403 asks for: assert on the property (the loop yields between children) rather than on a wall-clock constant, or scale the budget to a measured baseline taken in the same process, so host load cannot decide the verdict.

Filed separately from #403 rather than adjudicated inline, per the standing instruction on fixed-budget flakes.

damianvtran:
[...]

Issue #408, opened 2026-08-29
Subagent view: id-less anchored keys could collide with a real message id


damianvtran:
Follow-up deferred from PR #404 (agent code review round 1, F3).

## What

In `fold_trajectory`, an id-less event's anchored key is spelled `m{anchor}` / `t{anchor}`, so a stamped id-less message at sequence 5 keys as `ms5`. A child that emitted a **real** `message.id` of literally `"ms5"` would land on the same row (confirmed: the two fold into one row).

## Why it is not urgent

Purely theoretical. Real ids on this path are `subagent-launch:<job_id>` or provider-issued, neither of which takes the `m`/`t` + anchor shape. The pre-existing `m{index}` / `t{index}` scheme had the identical property, so PR #404 is not a regression here.

## Suggested fix

A separator would make it unrepresentable if these lines are touched again:

```python
message_id = str(message.get("id") or f"m:{anchors.of(event)}")
call_id = str(event.get("tool_call_id") or f"t:{anchors.of(event)}")
```

Note that changing the spelling changes row identity for id-less events, so it should land on its own rather than riding another change.

Source: https://github.com/damianvtran/local-operator/pull/404#issuecomment-5463434864

Issue #405, opened 2026-08-29
Subagent view: two consecutive identical notices read as the old duplicate-row bug


damianvtran:
Follow-up deferred from PR #404 (design review round 1, D1).

## What

When a child genuinely emits the same notice twice in a row, the page correctly paints **two** rows — the data model is right and PR #404 verified it does not over-collapse. But the two rows are byte-identical and vertically adjacent: same `✗`, same red, same wording, same indent, no ordinal, no count. The frame is a pixel-for-pixel two-row prefix of the eleven-row flood that PR #404 fixed.

## Why it matters

Readers who hit the duplicate-notice bug have a learned association: stacked identical red rows means the view is broken. Shown a true double failure they may read it as residual duplication and discount the second failure — the view tells the truth and is not believed. Three identical rows is squarely inside the visual range of the original defect.

The ambiguity is confined to *consecutive* identical notices with nothing between them. With intervening prose or a tool card, two identical notices read unambiguously as two attempts.

## Suggested fix

Cheapest first: when consecutive entries in the folded list share `kind == "notice"`, the same `notice_kind` and identical text, collapse them into one row with a trailing count in the notice's own dim ink:

```
✗ Invalid arguments: argument 'edits' does not match type array  ×2
```

Reuses the existing tier vocabulary, adds no colour, shrinks the flood signature to a single row, and states the repetition as a fact rather than leaving it inferred from row count.

## Why deferred

Pre-existing `NoticeBlock` behaviour that PR #404 made *reachable* by fixing the flood that used to swamp it, not something it introduced. The designer's verdict: non-blocking, and shipping the flood fix without it is strictly better than what users have today.

If this lands, it also improves the D2 case (documented in-code at `_Anchors.of`; see PR #404 remediation): a window-scoped undercount degrades into an undercounted `×N` rather than a missing row.

Source: https://github.com/damianvtran/local-operator/pull/404#issuecomment-5463414770

PR #425, opened 2026-08-30
fix(tui): /settings takes the whole view from the splash, and its list ends hold

damianvtran:
## Summary

Two user-reported bugs in the `/settings` page shipped in #387, both reproduced against `origin/main` (`a892aaf5`) before anything was changed.

1. **Opening `/settings` from the boot/splash screen failed to take over the view** — the boot layout stayed applied underneath, so the page shared the screen with a docked, width-clamped, centred input card belonging to another layout. Depending on terminal size that costs rows, horizontal span, or both: at 140x40 the page got 26 of 38 rows, and at 100x30 (the size the report was filed from) it kept all its rows but the card stayed clamped to 73 cells at column 12.
2. **The arrow keys wrapped from the bottom of the list back to the top**, where the wheel on the same list clamps.

Ships as a **patch** (0.43.8 → 0.43.9): bug fixes and their follow-ups, no new capability.

**Review round 1 is remediated in this PR** (F1-F5, D1-D2, U1-U3 — see the remediation comments). Two things changed materially as a result: the boot-layout suppression now covers **every** full-page mode rather than `/settings` alone (F2), and the bug-1 regression test asserts **both** dimensions, because at 100x30 a rows-only assertion passes against the broken tree (F1).

## Bug 1 — `Screen.boot` and `Screen.settings` were both applied at once

`Screen.boot` is not a decoration on the conversation layout, it is a **whole second layout**: the input is a docked, centred, width-clamped card, the transcript is bottom-aligned under it, and the app reserves rows *below* the card inside the dock's own padding (`_sync_boot_composition`). `_open_settings_view` added `SETTINGS_LAYOUT_CLASS` but nothing removed or overrode `boot`, so a page that expects the plain bottom bar got the leftovers above a card still holding its clamp and its reserve.

Measured at `a892aaf5`, 140x40:

```
boot class AFTER /settings: True     <- both layouts live
dock display: True | dock height: 5  (outer 9 — four rows of boot reserve)
settings view height: 26 | screen height: 38
```

**The decision: suppress `boot` while the mode is up, rather than override the boot rules from `Screen.settings`.** Two reasons, both recorded in the code:

- The tcss above `Screen.boot TranscriptView` says the class is *"the whole switch"* and that *"Two layouts, ONE condition"* is what keeps a centred boot card off a populated screen. A `Screen.settings.boot` counter-rule is precisely the second condition that comment warns about — the failure mode it names.
- It could not work anyway. The composition's lift is written as an **inline style** (`_reserve_boot_rows` sets `dock.styles.padding`), and an inline style outranks any stylesheet rule, so a sheet-side override would have moved the card and left the reserve behind.

So the condition stays single and one place resolves it (`_sync_boot_layout_class`): the boot layout is up when the session is empty **and** the user is looking at the conversation. Nothing is remembered across the mode — the class is re-derived from the two live facts, so a `/clear` or a session swap behind the page cannot leave a stale flag to restore.

**Scoped to `/settings` on purpose, and the other surfaces are reported not fixed.** See "Do the sibling modes share this?" below.

## Bug 2 — the arrows now clamp, and AGENTS.md records the exception

`action_move` did `indices[(position + delta) % len(indices)]`. `_scroll_rows` and `action_section` already clamped, so `down` and a wheel notch on the same last row disagreed about what the end of the list means.

`AGENTS.md` states the repo-wide convention ("Arrow keys wrap … wheel and page movement clamp"), so this is not a silent deviation: the convention is **amended to record this page as a documented exception**, with the reason. The wrap rule is written for a short picker where the whole list is on screen and coming round is a shortcut to a row the user can already see. It does not transfer to a list several times its viewport (60-odd rows against 14 at 100x30), where the bottom is a destination travelled to deliberately and wrapping throws the reader out of the section they were working in, scrolling the viewport with them.

`action_move`'s own docstring claimed the page wraps; it is rewritten rather than left contradicting the code.

**The other movement paths, audited:**

| path | behaviour | decision |
|---|---|---|
| `action_move` (`↑↓`, `ctrl+p/n`) | wrapped | **clamped** — the reported bug |
| `action_section` (`PgUp/PgDn`) | already clamped | unchanged |
| `_scroll_rows` (wheel) | already clamped | unchanged; stale docstring corrected |
| `action_jump` (`Home/End`) | absolute | unchanged, no ends to wrap |
| enum expansion | no movement of its own — expanded choices are ordinary rows in the one list | covered by the clamp above |
| `action_pane` (`←→` teams/agents) | cycles | **left cycling, deliberately** |

Pane switching keeps cycling because it is not the same gesture: two tabs, both labelled and both on screen, are a closed cycle rather than a list with ends. Nothing scrolls and there is no bottom to travel to, so `→` on the last tab has no meaning other than the first one — clamping it would make the second press of a two-tab toggle silently dead, which is the stuck key the wrap convention exists to prevent. `test_the_side_panes_still_cycle` pins that so a later tidy-up does not "fix" it into uniformity.

## Do the sibling modes share bug 1?

**They did, and they are now fixed here too** — changed during review round 1 remediation (F2).

The subagent view and the org chart mount through the identical three lines (hide the transcript, mount before `#input-dock`, add the mode's class), so they shared the collision exactly. Scoping the suppression to `/settings` also made them *route-dependent*, which is worse than uniformly broken: `_close_settings_view` restored the boot layout without asking whether another mode was still mounted, so the subagent view had one geometry opened from the splash and another opened over a settings page that was then closed underneath it — a fourth geometry that did not exist on base.

The condition was one more term in `_sync_boot_layout_class`, and both siblings ship with their own before/after frames rather than riding along unphotographed: [`docs/evidence/sibling-modes-boot-layout/`](docs/evidence/sibling-modes-boot-layout/README.md), at 100x30 and 140x40.

```
                          boot   boot-card  shell.w  shell.x  dock.outer  view.h
before  subagent 100x30   True   True        73        1       12          13
after   subagent 100x30   False  False       96        1        9          17
before  org      140x40   True   True        94       22        9          26
after   org      140x40   False  False      136        1        5          31
```

Both are pinned by tests that compare the splash route against the conversation route in both dimensions: `test_the_page_takes_the_whole_view_when_opened_from_the_splash` (`test_subagent_view.py`) and `test_the_chart_takes_the_whole_view_when_opened_from_the_splash` (`test_team_chart.py`).

`/failovers` does **not** share it — it appends a transcript block, which retires the splash through the ordinary content path. `/resume`, `/analytics` and the ask card are pushed screens or toast-layer cards, so the boot layout beneath them is still the layout being looked at and correctly stays.

## Visual evidence

Frames captured with `scripts/settings_shot.py`, extended with a `boot` state — every existing settings frame seeds a conversation first, which retires the splash and is exactly why this collision was invisible to the whole evidence set. Rendered to PNG and inspected as images, not trusted as green tests. Index: [`docs/evidence/settings-boot-layout/README.md`](docs/evidence/settings-boot-layout/README.md).

| before — 140x40, opened from the splash | after — same size, same state |
|---|---|
| before | after |

The before-frame is the report: the page ends after `Appearance`'s header, the input card sits clamped and centred mid-screen, and there is a band of dead ground under it. The after-frame has the page taking the view, the dock as a plain full-width bar, and the whole `Appearance` section on screen.

| before — 100x30 | after — 100x30 |
|---|---|
| before | after |

**The geometry behind the pixels** (AGENTS.md step 4) — the page from the splash now gets exactly what the same page gets over a conversation, in BOTH dimensions:

```
                  boot   boot-card  shell.w  shell.x  dock.outer  view.h
before  100x30    True   True        73       12       5           21
after   100x30    False  False       96        1       5           21
before  140x40    True   True        94       22       9           26
after   140x40    False  False      136        1       5           31
after   140x40  over a conversation (the reference)    5           31
```

**The collision is not the same shape at every size, and the earlier version of this table hid that** (design round 1, D1). At 100x30 — the size the report was filed from — the boot composition reserves ZERO rows, so the page gets its full 21 on both sides and every row-based number is identical. The whole defect at that size is horizontal: the card keeps its width clamp, 73 cells starting at column 12 instead of 96 from column 1, which is what the operator's screenshot showed. Vertical compression proper appears at 140x40, where the reserve costs the page five rows and `dock.outer` goes 9 → 5.

`screen.virtual_size.height <= screen.size.height` on both sides, so the compression at 140x40 was rows withheld from the page, not content overflowing it.

**Leaving restores the boot composition exactly.** The layout is suppressed rather than saved and re-applied, so this was checked rather than assumed: the splash frame taken immediately after Esc is **byte-identical** to the same frame from `origin/main`, at 100x30 and 140x40, once the SVG's generated ids and the scratch config path are normalised.

```sh
sed -E 's/terminal-[0-9]+/terminal-X/g' before2/boot-140x40.splash.svg > b.svg
sed -E 's/terminal-[0-9]+/terminal-X/g' after2/boot-140x40.splash.svg  > a.svg
diff b.svg a.svg
# 100x30 splash-after-leaving: IDENTICAL to origin/main
# 140x40 splash-after-leaving: IDENTICAL to origin/main
```

That is the splash, its composition, the card's clamp and the rows reserved below it all coming back unchanged. splash restored

## Real execution, not a green suite

Reproduced first, then re-run after. `scripts/settings_repro.py` drives the real `OperatorApp` at 100x30 from a fresh session with no conversation:

**Before (`a892aaf5`):**

```
boot class before: True
boot class AFTER /settings: True
settings class: True
dock display: True | dock height: 5
settings view height: 21 | screen height: 28

selectable rows: 51 | selected: 1
at bottom -> selected: 59 (last is 59)
one MORE down -> selected: 1 | wrapped to top? True
wheel at bottom -> selected: 59 | clamped? True
at top, one MORE up -> selected: 59 | wrapped to bottom? True
```

**After:**

```
boot class AFTER /settings: False
one MORE down -> selected: 59 | wrapped to top? False
at top, one MORE up -> selected: 1 | wrapped to bottom? False
wheel at bottom -> selected: 59 | clamped? True
after leaving: boot=True settings=False boot-card=True welcome.display=True dock.height=5
```

And the same thing driven the way a user does it — the typed slash command and real key presses, not the internal methods:

```
fresh boot: boot=True splash=True
after typing /settings: boot=False view.h=31 dock.outer=5 dock.inner=5 sum=38 screen.h=38
end then 5x down: 59 -> 59 (held=True)
80x up: selected=1 first=1 (held=True)
after esc: boot=True boot-card=True splash=True splash.h=23 dock.outer=9 settings_view=None
reopened: boot=False view.h=31
```

`dock.outer` going 5 → 9 across the Esc is the four-row boot reserve coming back; `sum=38` against `screen.h=38` is every row of the screen spoken for while the page is up. The round trip is repeatable.

## Regression tests

Every test here was **proven failing against the tree it guards**, by reverting the fix in a throwaway worktree rather than by trusting that it would.

`test_the_page_takes_the_whole_view_when_opened_from_the_splash` is now **parameterised across `(80,24)`, `(100,30)` and `(140,40)` and asserts BOTH dimensions** (F1). Against the reverted tree all three sizes fail, each naming what is wrong at that size:

```
FAILED ...[size0]  (80, 24): the boot composition reserve survived into the mode (7 outer vs 5 inner)
FAILED ...[size1]  (100, 30): the boot card's width clamp survived into the mode
                   (#input-shell is 73 cells at x=12)
FAILED ...[size2]  (140, 40): the boot card's width clamp survived into the mode
                   (#input-shell is 94 cells at x=22)
```

`size1` is the point of the change. **The previous version of this test passed at 100x30 on the broken tree** — the size the report was filed from — because the collision there is entirely horizontal and the test only asserted rows. Parameterising rather than looping also means one failing size no longer hides the others.

The sibling tests fail the same way against the pre-fix tree, at both sizes each:

```
FAILED test_the_chart_takes_the_whole_view_when_opened_from_the_splash[size0]
  - (100, 30): the boot card's clamp survived into the chart (#input-shell is 73 cells at x=12)
FAILED test_the_page_takes_the_whole_view_when_opened_from_the_splash[size0]   (subagent view)
  - (100, 30): the boot card's clamp survived into the page (#input-shell is 73 cells at x=1)
```

And the three UX fixes, against the head that had the clamp but not them:

```
FAILED test_every_movement_on_the_page_clamps_at_the_ends       - pagedown stopped short of the last row
FAILED test_arriving_at_the_top_shows_the_section_header_...    - the owning section header is scrolled off
FAILED test_a_retired_row_stops_advertising_keys_that_cannot_act
  - ↑↓  move · enter  change · r  default · ←→  panes · esc back to conversation
```

- `test_the_page_takes_the_whole_view_when_opened_from_the_splash` asserts **geometry, not class membership**: the view's rows, its width, and the shell's width and x must all equal what the same page gets over a conversation, and the dock's outer size must equal its inner. Both halves are load-bearing — the outer-vs-inner comparison is the only place a leftover vertical reserve shows, and the shell width/x is the only place the clamp shows.
- `test_leaving_the_page_restores_the_boot_layout` pins the other half: class, card clamp, dock reserve and splash height all come back.
- `test_the_splash_survives_a_settings_round_trip_over_a_conversation` pins that the suppression reads live state — a populated transcript must not gain a splash because the page was opened over it.
- `test_every_movement_on_the_page_clamps_at_the_ends` presses **repeatedly** at each end, since one press cannot tell a clamp from an off-by-one that still moves, and covers arrows, paging and wheel at both ends.

`test_arrows_wrap_and_page_clamps` encoded the reported bug as the contract, so it is rewritten rather than left sitting beside its own regression.

## Gates

Run over the whole tree, exactly as CI does:

| gate | result |
|---|---|
| `flake8 .` | clean |
| `black==26.1.0 --check .` | 617 files unchanged |
| `pyright` | 0 errors, 0 warnings |
| `pytest tests/unit` | 8757 passed, 25 failed |

The 25: 24 are `test_eval_tool.py`, the sanctioned shared-venv artifact, and `test_subagent_view.py::test_history_unavailable_and_error_retry_keep_trajectory_fallback` is a cross-test ordering artifact that **passes in isolation** (`1 passed`). Both were confirmed failing the same way on clean `origin/main` — the latter checked in a throwaway worktree at `a892aaf5`, which reports the identical `1 failed, 101 passed` for that file — so neither is from this change.


damianvtran:
### CI status on this head

Every job has a **green run on this exact head** (`6ca77b34`): `lint`, `type-check`, `pip-audit`, `test (3.12)` ([green](https://github.com/damianvtran/local-operator/actions/runs/33283720350/job/99186684121)) and `test (3.13)` ([green](https://github.com/damianvtran/local-operator/actions/runs/33283679257/job/99185583091)).

Some legs failed on first attempt, always with the **same single test**:

```
FAILED tests/unit/session/test_launch_subagent.py::test_the_loop_stays_responsive_while_several_subagents_run
  AssertionError: the event loop stalled for 1748 ms while subagents ran
  assert 1.748 < 1.0
```

It is a **pre-existing flake, not from this PR**, and the evidence is that it is failing on `main` itself — including on `a892aaf5`, the commit this branch is cut from:

| run on `main` | head | this test failed |
|---|---|---|
| [33279074013](https://github.com/damianvtran/local-operator/actions/runs/33279074013) | `a892aaf5` (this PR's base) | yes |
| [33276558665](https://github.com/damianvtran/local-operator/actions/runs/33276558665) | `eb499238` | yes |
| [33274806948](https://github.com/damianvtran/local-operator/actions/runs/33274806948) | `b2703675` | yes |

It is a wall-clock loop-stall budget (1000 ms) measured on a shared runner, in the subagent launcher. This diff touches only `local_operator/tui/` (the settings page and the boot-layout class), docs and tests — nothing that test exercises — and it passes locally (`1 passed in 8.01s`).

Worth a separate ticket to widen or de-flake that budget; deliberately not fixed here, since an unrelated change riding along in a bug-fix PR is one the reviewer has to review without having asked for it.

damianvtran:
### UX review — round 1

Scope: a892aaf543aaf4fbe804461b6c51b61c41277971..6ca77b343e357429531cb56a2582666c450c4b27

Walked the real `OperatorApp` at `~/local-operator-worktrees/settings-fixes` against scratch config dirs, driving `/settings` through the actual slash command with real key presses, real `MouseScrollDown`/`MouseScrollUp` events and real clicks. Frames rendered and viewed as images, not read as markup. Every behaviour called pre-existing below was re-measured against a throwaway worktree at the base commit `a892aaf5` rather than assumed.

**Verdict: no blockers, no majors. This round is TERMINAL.** Both reported bugs are fixed in the flow a user actually walks, and the fix survives the state-mutation cases it was specifically designed for.

---

## Findings

**U1 — MINOR — the top of the list gives weaker end feedback than the bottom, because `home` lands on `scroll_y=1` and hides the first section header.**

Flow: open `/settings` from the splash at 100x30, press `end` to travel to the bottom, then hold `up` (80 presses) back to the top.

At the **bottom** end the clamp reads correctly: the highlighted last row `Max learnings history` is on screen, the scrollbar thumb is visibly parked at the end of its track (`scroll_y == max_scroll_y`, 46/46), and the detail line under the list still describes the highlighted row. A user can see why the key stopped working.

At the **top** end the same gesture settles on `scroll_y=1`, not `0`. Row 0 is the `Model` section header that owns the highlighted `Default provider` row, so the frame shows a highlighted row whose section title is scrolled just off the top edge and a scrollbar thumb that is not quite at the start of its track. Measured, both paths:

```
FRESH OPEN      scroll_y=0.0  header row 0 visible = True
AT BOTTOM       scroll_y=46 / max=46   thumb at end   = True
HELD UP TO TOP  scroll_y=1  / max=46   thumb at start = False
```

Expected: holding `up` at the top lands the viewport at `scroll_y=0` so the `Model` header is visible and the thumb sits at the track start, matching the confidence the bottom end already gives.

**Pre-existing, not introduced here** — base `a892aaf5` also settles on `scroll_y=1` after `end` then `home`. But this PR is what makes it reachable by holding a key: at base, holding `up` at the top wrapped away to row 57, so nobody dwelt on the top end long enough to notice. The clamp is right; the off-by-one scroll target it exposes is worth a follow-up. Suggest `_scroll_to_selection` targeting the owning header when the selection is the first selectable row.

**U2 — MINOR — the read-only bottom row makes the clamped bottom end the least rewarding place to arrive.**

Flow: `end` to the bottom, then `enter`.

The last selectable row of the page is `Max learnings history`, a `Kind.READONLY` setting. Pressing `enter` on it produces `this setting is retired and cannot be changed` — correct and well-worded copy — but it means the destination the new clamp deliberately holds you at is a row you cannot act on, under a footer that still advertises `enter change · r default`. The detail line does say `Deprecated. Superseded by the compaction engine.`, so the state is explained; nothing is broken.

Expected: nothing changed in this PR. Raising it only because the clamp changes the bottom from a transient waypoint into a place users will now park, which slightly raises the value of that footer reflecting per-row capability. Pre-existing; follow-up at most.

**U3 — NIT — `pagedown` cannot reach the last row, so the two "go down" gestures still disagree about where the end is.**

Flow: `home`, then `pagedown` repeatedly until the selection stops moving.

```
pagedown settles at row 54; last selectable row = 59
selectable rows below the pagedown settle: [55, 56, 57, 58, 59]
scroll_y = 41 / 46  (bottom NOT reached)
```

`pagedown` is section-scoped by design (`action_section` jumps header to header and clamps on the last **header**, documented at `settings_view.py:450`), so it settles on the last section's first row with five rows still below it, while `end` and held `down` both reach row 59. This is unchanged by the PR and consistent with the documented design; `down` from the settle point still reaches the end. Noting it only because the PR's stated goal is making the movement paths agree on "what the end of the list means", and paging remains the one path with a different answer. Behaviour is identical at base.

---

## Verified working — no finding

**Bug 1, the operator's report — fixed, in the real flow.** Typed `/settings` at the splash. The page takes the whole view and the dock renders as a plain full-width bar. The load-bearing measurement at 100x30 is horizontal, not vertical: the boot card's width clamp is what the before-frame shows, and it is gone.

```
                    Editor x / width      boot   dock outer/inner   view_h
base   100x30       x=15 w=71  (clamped)  True   5 / 5              21
fixed  100x30       x= 4 w=94  (bar)      False  5 / 5              21
base   140x40       x=25 w=92  (clamped)  True   9 / 5  (+4 reserve) 26
fixed  140x40       x= 4 w=134 (bar)      False  5 / 5              31
```

**Leaving restores the conversation exactly.** Splash present, composer read-write, focus back on `Editor`. A half-typed prompt is preserved byte-for-byte across the round trip (`'typed after'` in, `'typed after'` out) — the transcript is only hidden, never rebuilt.

**Re-derivation survives every state move I could throw at it.** This is where I expected to break it and could not:

- five consecutive open/close cycles via real keys: `boot` toggles `True→False→True` cleanly each time, dock and welcome heights return to 5 and 23 every cycle, no drift.
- `ctrl+l` (`/clear`) pressed **while the page is open**: closes the page, splash returns, and reopening afterwards works. Dock goes to outer 9 / inner 5 — the boot reserve for a post-clear splash, identical at base, not a regression.
- splash **retired** behind the open page (a real `_notice` appended at 140x40): leaving correctly does **not** bring the splash back (`boot=False`, dock outer == inner, no reserve). This is the operator's "send a message, then leave" case.
- splash **restored** behind the open page: leaving correctly brings it back.
- composer is read-only inside the page (typing `xyz` reaches nothing), so the user cannot type `/clear` from within — the risk is only programmatic, and it is handled.

`_sync_boot_layout_class` is the single writer of `BOOT_LAYOUT_CLASS` in `app.py`, so there is no second path to leave a stale class.

**Bug 2, the clamp — holds everywhere, and does not read as a stuck key.** At 100x30, 140x40 and 80x24: `end` then 8x `down` holds at row 59; `home` then 8x `up` holds at row 1. Also confirmed with 40 rapid presses and after a chain deletion. Every movement path agrees with the new rule:

| path | at the bottom | at the top |
|---|---|---|
| `↑↓` arrows | holds at 59 | holds at 1 |
| `ctrl+n` | holds at 59 | n/a (`ctrl+p` is the app command palette, at base too) |
| real wheel events x12 | holds at 59, `scroll_y` unchanged | holds at 1 |
| `pageup` | n/a | holds at 1 |
| real click + `end` + `down` | holds at 59 | — |

Clicks land on the row clicked (`click y=2 → row 2 model_name`, `y=3 → row 3 providers.openai.api`, last visible row at the bottom → row 59), and clicking then holding `down` still clamps.

**`←→` pane cycling still cycles, as intended.** `right` x4: `teams→agents→teams→agents`. `left` x4: the same closed cycle. The second press of a two-tab toggle is never silently dead.

**The editor at the ends.** Committing by moving at the last text-editable row (`web_fetch.cache_ttl_seconds`) writes correctly and lands `cache_ttl_seconds: 33` on disk, cursor advancing one row as expected. An **invalid** buffer at the same row holds the cursor exactly where it was (49 → 49), keeps the editor open, shows `expected a whole number`, and leaves the config **byte-identical**. Caret keys inside an open editor are owned by the editor: `home→0`, `end→2`, `left→1`, `right→2`, and `left`/`right` do **not** leak to pane switching (`pane` stays `teams`).

**Regression sweep, all clean.** Esc ladder: expansion closes before the page, an open chain closes before the page, an armed delete cancels before the page (and the chain survives), an open editor closes before the page — the page itself only closes on the press after, returning focus to `Editor`. Identity-anchored cursor: a bool toggle keeps the cursor on the same setting (`retry.enabled`, index 5 → 5); choosing a theme returns to the owning enum row, not the choice index. Deleted anchor: deleting the last of three chains leaves the cursor in range and on a selectable row (`chain_add`), with the clamp still holding at both ends afterwards. Byte-level no-op: opening the page and moving all over it (`end`, 20x `up`, 20x `down`, `home`, `end`, `pagedown`, `pageup`, `right`, `left`) then leaving writes **no config file at all**.

**`#369` `/model default` grammar intact**, six spellings, config bytes checked after each:

| spelling | config changed | response |
|---|---|---|
| `/model` | no | `model: test/model — d in /model saves this for new sessions` |
| `/model default` | **no** | `save test/model as the boot default? — /model default test/model confirms · d on a /model row does it too · /model saved switches to the configured default instead` |
| `/model DEFAULT` | **no** | same confirmation (case-insensitive) |
| `/model saved` | no | `no boot default saved yet — /model default <provider>/<model-id> sets one` |
| `/model default test/model` | no | `provider controller unavailable` (fake session has no provider controller) |
| `/model test/model` | no | `provider controller unavailable` |

The read-only forms wrote nothing — no `config.yml` was created across the whole sequence. Bare `/model default` still confirms rather than writing, which is the #369 fix.

**Resize and narrow terminals.** Resized 100x30 → 140x40 → 80x24 → 60x18 → 46x12 → 140x40 with the page open: `boot` stays off, dock stays outer == inner == 5, `virtual_size <= size` at every step (no overflow), and the page reflows without the boot card reappearing. Leaving at 60x18 restores the splash correctly, and growing back to 100x30 keeps it correct. Opening the page while already at 50x14 works and the clamp still holds. Degradation at very small sizes (body down to 1-2 visible rows at 46x12 and 60x18) is **identical to base**, not introduced here.

**Async responsiveness.** The input loop never blocks: 200 clamped no-op moves cost 239.9 ms (1.199 ms each) against 200 real moves at 244.9 ms (1.224 ms each) — a clamped press repaints, but at the same cost as a real one, so a held key at the end is no more expensive than one in the middle.

**Tests and gates re-run, not trusted.** `tests/unit/tui` 3658 passed, 12 skipped. `flake8`, `black --check` (26.1.0), `isort --check` all clean whole-tree. The regression tests genuinely fail at base: copied onto a throwaway `a892aaf5` worktree they give **3 failed, 2 passed**, with `down at the bottom did not hold / assert 11 == 59`, `has_class('boot')` still true after leaving, and `the boot composition reserve survived into the mode (7 outer vs 5 inner)` — matching the implementer's claim exactly, including that the dock **outer vs inner** assertion is the load-bearing one.

**Sibling views confirmed still affected, as the PR states.** The subagent view opened from the splash keeps `boot=True` and leaks the reserve — 4 rows at 140x40 (view 26 of 38 against settings' 31) and 2 rows at 80x24. Correctly scoped out and disclosed rather than silently shared.

---

## Flows exercised

1. Fresh splash → typed `/settings` → used the page → `esc` → conversation, at 100x30, 140x40, 80x24 and 50x14.
2. Half-typed prompt preserved across a full page round trip.
3. Five consecutive open/close cycles via the real slash command and `esc`.
4. `ctrl+l` clear pressed with the page open; splash retired behind the page; splash restored behind the page; a real `_notice` arriving behind the page.
5. Send a real message, then open and leave the page — splash correctly stays gone.
6. Clamp at both ends via arrows, `ctrl+n`, real wheel events, `pageup`/`pagedown`, `home`/`end` and real clicks.
7. `←→` pane cycling, four presses each direction.
8. Editor commit by movement at the last text row; invalid buffer at both ends; caret keys inside an open editor.
9. Esc ladder: expansion, open chain, armed delete, open editor, then the page.
10. Chain add via the real add flow, chain delete with `d`,`d`, deleted-anchor cursor landing, clamp after deletion.
11. `/model` grammar, six spellings with config bytes compared after each.
12. Resize with the page open across five sizes; open at 50x14; narrow degradation compared against base.
13. Rendered frames viewed as images at 100x30 (splash, page top, bottom held, top held, back to splash), 140x40, 80x24 and 46x12, plus the PR's own before/after evidence frames.

Follow-ups worth a ticket, none blocking: **U1** (top-end `scroll_y=1` hides the owning section header), **U2** (read-only last row under a generic footer), **U3** (`pagedown` cannot reach the last row), plus the PR's own disclosed items — the subagent view and org chart sharing bug 1, and the `test_the_loop_stays_responsive_while_several_subagents_run` wall-clock flake.

damianvtran:
### Design review — round 1

Reviewer: `designer` subagent (anthropic/claude-opus-5)
Scope: `a892aaf543aaf4fbe804461b6c51b61c41277971..6ca77b343e357429531cb56a2582666c450c4b27`

Layout fix, so this is a review of rendered frames. I rendered the committed evidence to PNG and looked at it, then captured my own frames across a size band and against a **pristine** app that never opened the page, because the PR's own restore check compares two frames that both went through the mode.

Tooling note: `qlmanage` rendered the five committed frames and three band frames fine, then stalled on two later files (the documented hang on this machine). Where it stalled I read the painted text and rect geometry straight out of the SVG instead, which is what the width/position numbers below come from.

---

#### 1. Does it read as a takeover?

Yes, and the receded dock reads as intentional on the splash path.

The thing that made the before-frame read as "leftover composer" was not the dock's height, it was its **horizontal clamp**. Measured from the card's background rects:

| frame | card left edge | reads as |
|---|---|---|
| `before-boot-100x30` | x=146.4 (**col 12**) | inset, centred card floating over the page |
| `after-boot-100x30` | x=12.2 (**col 1**) | full-width bar flush to the bottom |
| `before-boot-140x40` | x=268.4 (**col 22**) | inset, centred, and lifted off the bottom (y=734) |
| `after-boot-140x40` | x=12.2 (**col 1**) | full-width, sits on the bottom edge (y=831) |

That is the difference between a card that belongs to another screen and a status bar that belongs to this one. In the after-frames the `Read-only — press esc to reply` prompt starts at the same column as the page's own content, so the dock lines up with the page rather than floating in front of it. The 5 remaining rows read as the page's own footer, not as a composer left behind. The operator's complaint is answered.

Frames inspected: `before-boot-100x30`, `after-boot-100x30`, `before-boot-140x40`, `after-boot-140x40`, `after-splash-restored-140x40` (all committed), plus my own captures listed below.

#### 2. The splash restore — verified harder than the PR claims

The PR proves the after-leaving splash matches origin/main's after-leaving splash. Both of those went through the mode, so I compared against an app that **never opened `/settings`**, at six heights, on the composition's real carrier (the dock's OUTER size, which is where `_sync_boot_composition`'s inline reserve padding lives):

```
(100,16) MATCH  welcome.box=(96,9)   dock.outer=5 inner=5 pad=(0,0,0,0)
(100,20) MATCH  welcome.box=(96,12)  dock.outer=5 inner=5 pad=(0,0,0,0)
(100,24) MATCH  welcome.box=(96,12)  dock.outer=7 inner=5 pad=(0,0,2,0)
(100,28) MATCH  welcome.box=(96,12)  dock.outer=9 inner=5 pad=(0,0,4,0)
(100,30) MATCH  welcome.box=(96,23)  dock.outer=5 inner=5 pad=(0,0,0,0)
(100,40) MATCH  welcome.box=(96,23)  dock.outer=9 inner=5 pad=(0,0,4,0)
```

The reserve varies with height (0, 2 and 4 rows all occur in that band) and the restored value tracks it exactly every time, including the card's own box. Card clamp, slack below it and splash height all come back. Nothing shifted.

I also checked the two-frame settling on both transitions at 100x24, 100x30 and 140x40 — open frame 1 equals open frame 2, leave frame 1 equals leave frame 2, no intermediate state where the boot card and the page are both up. One pair showed a single colour-class fill drift (`#8a8374` → `#888172`); a pristine splash that never opens the page drifts more than that between consecutive frames, so it is the splash shimmer, not this change.

#### 3. The size band — swept, clean

Splash path at 16, 20, 24, 28, 30, 40 rows, plus 60, 80 and 140 columns. After normalising the scratch-config path and Textual's generated ids, **every splash-path frame is byte-identical to the same page opened over a conversation**:

```
16: IDENTICAL   20: IDENTICAL   24: IDENTICAL
28: IDENTICAL   30: IDENTICAL   40: IDENTICAL (scratch dir name only)
60x24: IDENTICAL   80x24: IDENTICAL   140x40: IDENTICAL (scratch dir name only)
```

`screen.show_vertical_scrollbar=False` at every size, `virtual_size <= size` throughout, so nothing overflows and the two-cell scrollbar tax never appears. The `read-only` pane caption survives at every height from 20 up; the pane sheds whole (`pane.size=(0,0)`) at 60 and 80 columns, which is the designed narrow behaviour, and the footer sheds with it (`esc back` at 60 cols, `esc back to conversation` at 80+). No dead space reappears at any size in the band.

#### 4. #387's properties — undisturbed

All ten detail-line states captured at 100x30 on this head hold layout at `view.height=21` with no screen scrollbar: `enum`, `error`, `cascade`, `confirm`, `confirm-long`, `teams`, `agents`, `theme`, `frames`, `retired`. Compared frame-by-frame against `a892aaf5`, the only differing painted lines in every state are the worktree path in the status bar and the scratch config dir in the header — the surface itself is unchanged. `_PaneLine` ink distinctions (bright team name, dimmed `manager · 6 roles`, dimmed description with its `…` truncation), the scrollbar palette and the tag column alignment all render as before. `confirm-long` still budgets its ask against the row it is painted into.

---

### Findings

**D1 · MINOR · The 100x30 evidence pair does not show the bug its caption claims.**
`README.md` describes `before-boot-100x30.svg` as "the page squeezed into the rows above the boot input card", and the PR body presents the 100x30 pair under the compression heading. There is no vertical compression at that size. Measured at base and at head:

```
a892aaf5 (100,30) boot=True   view.h=21 dock.outer=5 dock.inner=5
6ca77b34 (100,30) boot=False  view.h=21 dock.outer=5 dock.inner=5
```

The page gets the same 21 rows on both sides, because at 100x30 the boot composition happens to reserve zero rows (`pad=(0,0,0,0)`). The real change at that size is horizontal only — the card un-clamps from col 12 to col 1. Vertical compression exists at 140x40 (`view.h` 26 → 31, `dock.outer` 9 → 5), and that pair does show it honestly.

The fix is correct at both sizes and the frames themselves are accurate; only the words describing the 100x30 pair overclaim. Given this page's history of evidence that reads as proof of something it does not show, the caption is worth correcting to say what that pair actually demonstrates (the clamp and the centring), leaving 140x40 as the compression exhibit. Docs-only, no code change.

**D2 · NIT · At 16 rows the page paints chrome and no settings.**
At 100x16 the body is one row tall and renders zero setting rows — the frame is header, rule, detail line, hints and dock, with a blank band where the list should be. It is honest about being a settings page but shows nothing settable.

Not this PR's: `a892aaf5` renders the identical frame, and the splash path and conversation path are byte-identical here on this head. Recording as a follow-up on the page's minimum-height behaviour rather than as a finding against this change.

---

### Verdict

No BLOCKER and no MAJOR. **This round is TERMINAL.**

The takeover is real and reads as intentional at every size I swept, the restore is exact against a splash that never opened the page (a stricter check than the PR's own), the size band that broke twice during #387 is uniform this time, and #387's surface properties are untouched. D1 is a caption correction and D2 is a pre-existing follow-up; neither blocks the merge.

Frames inspected: committed `before-boot-100x30`, `after-boot-100x30`, `before-boot-140x40`, `after-boot-140x40`, `after-splash-restored-140x40`; my own splash-path captures at 100x16/20/24/28/30/40 and 60x24/80x24/140x40 with matching conversation-path references; pristine-splash captures at the same six heights; open/leave frame pairs at 100x24, 100x30, 140x40; and the ten detail-line states at 100x30 on both base and head.

damianvtran:
### Agent review — round 1

Reviewer: `reviewer` subagent (lopdev team), anthropic/claude-opus-5
Scope: `a892aaf543aaf4fbe804461b6c51b61c41277971..6ca77b343e357429531cb56a2582666c450c4b27`

Reviewed the diff first, then opened files only where the diff could not answer a stated question. Everything below that says "measured" was run by me against the committed head in `~/local-operator-worktrees/settings-fixes`, or against a throwaway worktree at the base commit `a892aaf5` — not inferred from the PR body.

**Gates, run whole-tree against the committed head:** `flake8` clean; `black==26.1.0 --check` clean (616 files); `isort==5.13.2 --check` clean; `pyright` **0 errors, 0 warnings**; `tests/unit/tui` **3658 passed, 12 skipped** in 150s. `tests/unit/tui/test_settings_view.py` alone: 360 passed.

**The bug-1 test genuinely binds (verified by reverting).** In a detached worktree at `6ca77b34` with only `app.py` and `settings_view.py` reverted to `a892aaf5`, the new tests report `3 failed, 2 passed` — exactly the implementer's claim. The OUTER-vs-inner assertion is what catches it at 80x24 (`7 outer vs 5 inner`), and the two boot-class assertions catch it elsewhere. Confirmed the implementer's warning is real: an assertion on `dock.size.height` alone passes against the broken tree.

**The core fix is correct, and I could not break it.** `_sync_boot_layout_class` (app.py:7827-7872) derives `boot = bool(self._welcome_visible) and self._settings_view is None` from live state with nothing remembered, and all three call sites are right. Measured on the committed head:

| scenario | result |
|---|---|
| `/settings` from splash, 80x24 / 100x30 / 140x40 | view+dock heights **identical** to the same page over a conversation, at all three |
| resize 100x30 → 140x40 **while the page is open** | `boot=False`, `pad=0`, `view.h=31`, `sum=38=screen.h` — correct, no reserve leaks back |
| close at the new size | `boot=True card=True dock_out=9 pad=4` — recomposed for 140x40, not restored from a 100x30 snapshot |
| open/close ×5 | `(True, True, 5, 0, 23)` every round, **zero drift** |
| content appended UNDER the open page, then close | `boot=False`, splash correctly stays retired |
| splash brought back under the open page (`_set_welcome_visible(True)`), then close | page stays full-page at 21 rows; `boot=True` on close |

I specifically hunted for the "boot left off after close, stranding a session with no splash" state and did not find one. `_close_settings_view` sets `self._settings_view = None` **before** calling the sync, and every one of the seven other `_close_settings_view()` call sites (session swap 3352, ctrl+c 6308, Esc 6459, approval 6897, ask card 7016, ctrl+L 7518, key prompt 14208) funnels through that same method, so none can bypass the restore. The `_swapping_session` deferral is also safe: `_welcome_pending` is applied through `_set_welcome_visible`, which calls the sync.

---

**F1 — minor — the bug-1 regression test does not fail at the size the report was filed from.**
`tests/unit/tui/test_settings_view.py:589`. The test sweeps `(80, 24), (100, 30), (140, 40)`, but the assertions it makes are only satisfied by the *vertical* half of the boot layout. I ran the test against the reverted tree with the loop narrowed to `((100, 30),)` — it **passes on the broken tree**. Measured, the reverted tree at 100x30 gives `view.h=21 dock_out=5 pad=0`, which is identical to the fixed tree; the collision at that size is entirely horizontal (`boot-card` survives, and `#input-shell` is **73 cells wide instead of 96**). The test never asserts on `boot-card` or on the shell width, so the clamped, centred card — half of what the operator photographed — is invisible to it at 100x30. It is caught at 80x24 and 140x40 only because those sizes happen to also lose rows. Adding `assert not screen.has_class("boot-card")` inside the loop makes it fail at 100x30 on the reverted tree (confirmed). One line; the fix itself is right, the test is one dimension short of pinning it.

**F2 — minor — the two unfixed siblings are now *route-dependent*, which is a worse trap than uniformly broken.**
Confirmed the shared-mount claim: `_open_subagent_view` (app.py:9487) and `_open_org_chart_view` (app.py:9577) both do `_transcript_view().display = False` → `screen.mount(view, before=#input-dock)` → `screen.add_class(...)` and neither touches `_sync_boot_layout*`. Measured subagent-from-splash: 80x24 `view.h=12` vs 15, 100x30 shell 73 vs 96, 140x40 `view.h=26` vs 31 — the same collision, as reported. What the PR body does not say is that the fix makes the subagent view's geometry depend on **how you got there**, which the base tree did not:

```
140x40, boot session   subagent DIRECT from splash        boot=True  view.h=26 pad=4 shell.w=94
140x40, boot session   subagent opened over /settings     boot=False view.h=15 pad=0 shell.w=136
140x40, boot session   ...then Esc the settings page      boot=True  view.h=12 pad=4 shell.w=94
```

On the base tree all three rows are `view.h=12..26, boot=True` — inconsistent but *stable*. Post-fix the middle row is a fourth distinct geometry that did not exist before, because `_close_settings_view` restores `boot` without asking whether another full-page mode is still mounted. The `_sync_boot_layout_class` condition is `self._settings_view is None`, not "no full-page mode is up", while its own docstring is titled *"Put `Screen.boot` on only while no full-page MODE is up."* — the docstring describes the general rule the code does not yet implement.

I am rating this **minor, not major**, and deliberately: I could not find a *user-reachable* route to that state. `Screen.settings` leaves `#subagent-panel` live (tcss:1367 says so explicitly), but the panel's `on_open` fires `_open_subagent_view` on a click, and I could not construct the click while the page is up without also having a running job. Every keyboard route (`Esc`, ctrl+L, an approval, the ask card) closes settings *first*. So this is latent, not live — but it is exactly the "next person walks into it" trap the priority asked about, and the one-term fix the docstring already promises (`and self._subagent_view is None and self._org_chart_view is None`) would close it. **Please file the follow-up ticket the PR body promises** — I searched open issues and found none for this; #405-#408 are unrelated subagent-view items.

**F3 — minor — a stale comment now contradicts the code it sits above.**
`app.py:947-948`: *"It selects the boot layout in the stylesheet … and is flipped in **exactly one place** — see `OperatorApp._set_welcome_visible`."* After this PR the class is flipped in `_sync_boot_layout_class`, from three call sites, and `_set_welcome_visible` no longer touches it. The tcss carries the same stale claim at `local_operator.tcss:796` (*"the app flips it in the same call that shows or hides the splash (OperatorApp._set_welcome_visible)"*). Both are the kind of load-bearing comment this repo relies on; a future agent reading either would go to the wrong function. One-line redirect to `_sync_boot_layout_class` in both.

**F4 — nit — `_position_of`'s fallback docstring is now slightly wrong under the clamp.**
`settings_view.py:434-445` says the fallback exists so *"a movement that refused to move because its origin vanished would be the same stuck key this exists to prevent."* Measured: when the anchor identity is gone **and** `_selected` is no longer selectable, `_position_of` returns 0, and under the clamp `action_move(-1)` from there is now genuinely a no-op — the stuck key the sentence disclaims. Under the old wrap it moved (to the bottom). The behaviour is defensible (both outcomes teleport the cursor to the top of a 60-row page, which is the real problem, and the common case is fine: I confirmed that with a vanished identity but a still-selectable `_selected`, the fallback correctly returns position 45 of 45 rather than 0). This is a docstring accuracy point, not a defect.

**F5 — nit — the AGENTS.md exception is coherent but under-specifies its trigger.**
`AGENTS.md:306-319` reads well and records the *reason*, not just the rule, and the pane-cycling carve-out is stated. Two rules in tension? No — the exception is scoped and reasoned. But its trigger is "a long, sectioned, SCROLLED list", and by that description `model_picker.move` (`model_picker.py:418`, wraps) also qualifies: it windows a scrolled catalogue via `_window_start`/`_row_budget` and can hold hundreds of rows. `command_picker.move` (:991, wraps) is the short on-screen picker the original rule was written for, and `session_picker._move_to` (:913) already clamps with a near-identical comment. As written a future agent could read the amendment as licence to clamp the model picker. Naming `/settings` as the *only* current exception, or tightening the trigger to "the list is the whole page", would settle it.

---

**Verified and agreed, no finding:** the clamp itself (`settings_view.py:430`) is correct at both ends — measured `59→59` on 10 down-presses at the bottom and `1→1` on 10 up-presses at the top, matching `_scroll_rows`'s identical expression at :1295 exactly. `action_pane` cycling is right to leave alone (two panes, both on screen, nothing scrolls; clamping would make press two of a toggle dead), and `test_the_side_panes_still_cycle` asserts something real — it checks the round-trip returns to `start`, which a clamp would break. The clamp does not disturb the identity-anchored rebuild-after-commit path; `test_a_commit_that_inserts_rows_does_not_strand_the_cursor` still passes and its docstring was correctly updated to stop citing wrapping as the rationale. The committed evidence frames are genuinely different files (before/after md5s differ at both sizes), and the restore is exact rather than approximate. The version bump to 0.43.9 is correct — two bug fixes, no new capability, patch per AGENTS.md materiality.

**CI:** every job green on this exact head (`6ca77b34`) — `lint`, `type-check`, `pip-audit`, `test (3.12)`, `test (3.13)`. The red `test` legs are first-attempt runs of `test_the_loop_stays_responsive_while_several_subagents_run`, tracked as #418 (flaky on main itself), already documented in a PR comment.

---

**Verdict: `clean`.** No blocker and no major findings. This round is **TERMINAL** — F1 through F5 are minor/nit follow-ups and are not a reason for another round. F2 in particular is latent rather than user-reachable, but it should get the follow-up ticket the PR body promises before the next person extends one of these modes.

damianvtran:

Implementer: `coder` (lopdev team), anthropic/claude-opus-5
Scope: `6ca77b34..9f04a50e` (reviewed head `6ca77b34`, remediation head `9f04a50e`)

All five findings taken. F1 and F2 changed behaviour and both carry their own reverted-tree proof; F3-F5 are comment and docs accuracy. No finding rejected, none deferred.

---

**F1 — the bug-1 regression test did not fail at 100x30 — FIXED** (`b95aebe2`)

You were right, and the failure mode is exactly the one you named. Confirmed by measurement before changing anything, reverted tree vs fixed tree:

```
                   boot  boot-card  shell.w  shell.x  dock.out  view.h
reverted 100x30    True  True        73       12       5         21
fixed    100x30    False False       96        1       5         21   <- rows IDENTICAL
reverted 140x40    True  True        94       22       9         26
fixed    140x40    False False      136        1       5         31
```

At 100x30 every number the old test asserted on is the same on both sides. The test is now **parameterised** across `(80,24)`, `(100,30)`, `(140,40)` and asserts **both dimensions**: `not screen.has_class("boot-card")` plus `shell.size.width == view.size.width` and `shell.region.x == view.region.x` for the horizontal half, and the existing dock outer-vs-inner and row-sum assertions for the vertical half. The reference comparison now compares the 4-tuple `(view.h, view.w, shell.w, shell.x)` against the conversation route rather than rows alone.

Parameterised rather than looped deliberately: a loop reports the first failing size and hides the rest, which is part of how this stayed invisible.

**Proven to bind at the size that matters.** Reverted `app.py` to `a892aaf5` in a throwaway worktree, keeping the new tests:

```
FAILED ...[size0] - (80, 24): the boot composition reserve survived into the mode (7 outer vs 5 inner)
FAILED ...[size1] - (100, 30): the boot card's width clamp survived into the mode
                    (#input-shell is 73 cells at x=12)
FAILED ...[size2] - (140, 40): the boot card's width clamp survived into the mode
                    (#input-shell is 94 cells at x=22)
3 failed
```

`size1` is `(100, 30)`, failing on the horizontal assertion specifically — not passing in aggregate.

---

**F2 — the unfixed siblings were route-dependent — FIXED, option (a)** (`cd2b4fd0`)

Took **(a): generalised the condition to match its own docstring**, not (b). It was genuinely small and both siblings are photographed, which was your condition for preferring it.

The condition is now:

```python
boot = (
    bool(self._welcome_visible)
    and self._settings_view is None
    and self._subagent_view is None
    and self._org_chart_view is None
)
```

plus `_sync_boot_layout_class()` / `_sync_boot_layout()` at the four sibling open/close sites, mirroring the settings sites. **14 lines, no other edit** — exactly the one-term fix you predicted.

**Surfaces this now covers:** `/settings`, the **subagent view**, and the **org chart**. Explicitly still excluded, and the docstring says why: `/resume`, `/analytics` and the ask card are pushed screens or toast-layer cards, so the boot layout beneath them is the layout being looked at and must stay; `/failovers` is not a mode at all.

**The route dependence you measured is gone.** Same three routes plus the reference, subagent view at 140x40, before (`6ca77b34`) and after:

```
                                    BEFORE                    AFTER
subagent DIRECT from splash         boot=True   view.h=24     boot=False  view.h=27
subagent opened over /settings      boot=False  view.h=13     boot=False  view.h=13
...then Esc the settings page       boot=True   view.h=11     boot=False  view.h=13
subagent over a conversation (ref)  boot=False  view.h=27     boot=False  view.h=27
```

Two things to read here. The splash route now equals the conversation reference (27 == 27), which is the fix. And **Esc no longer changes the geometry of a page it is not closing** — before, closing the settings page underneath restored `boot` and moved the subagent view 13 → 11; now it stays 13. That was the trap: the restore did not ask whether another mode was still mounted.

The `over /settings` rows are the state with two modes mounted at once, which is still not user-reachable by any route I could construct — its 13 rows differ from 27 because the expanded roster is collapsed on that path, not because of the boot layout. It is included because it is where the route dependence lived.

**Frames, since that was the condition for taking (a).** New evidence set with its own capture script, both sizes, both modes: [`docs/evidence/sibling-modes-boot-layout/`](https://github.com/damianvtran/local-operator/blob/dev-settings-fixes/docs/evidence/sibling-modes-boot-layout/README.md). Rendered to PNG and inspected as images, not read as markup.

| | before | after |
|---|---|---|
| subagent 100x30 | before | after |
| org chart 100x30 | before | after |

```
                          boot   boot-card  shell.w  shell.x  dock.outer  view.h
before  subagent 100x30   True   True        73        1       12          13
after   subagent 100x30   False  False       96        1        9          17
before  subagent 140x40   True   True        94        1       11          24
after   subagent 140x40   False  False      136        1        9          27
before  org      100x30   True   True        73       12        5          21
after   org      100x30   False  False       96        1        5          21
before  org      140x40   True   True        94       22        9          26
after   org      140x40   False  False      136        1        5          31
```

The org-chart 100x30 pair is the clearest reading of the whole bug: **21 rows on both sides**, and the only change is the card un-clamping from 73 cells at column 12 to 96 at column 1 — the inset floating card the operator photographed.

Pinned by tests in each mode's own file, both parameterised over `(100,30)` and `(140,40)` and asserting both dimensions: `test_the_page_takes_the_whole_view_when_opened_from_the_splash` (`test_subagent_view.py`) and `test_the_chart_takes_the_whole_view_when_opened_from_the_splash` (`test_team_chart.py`). Both fail against the pre-fix tree at both sizes:

```
FAILED test_the_chart_takes_..._from_the_splash[size0] - (100, 30): the boot card's clamp
       survived into the chart (#input-shell is 73 cells at x=12)
FAILED test_the_page_takes_..._from_the_splash[size0]  - (100, 30): the boot card's clamp
       survived into the page (#input-shell is 73 cells at x=1)
4 failed
```

**No follow-up ticket needed** — the thing the ticket would have tracked is fixed. Whole `tests/unit/tui` re-run against the generalised condition: **no new failures**.

---

**F3 — two stale comments pointing at the wrong function — FIXED** (`cd2b4fd0`)

`app.py:947` and `local_operator.tcss:796` both said the class is flipped in `_set_welcome_visible`. Both now name `_sync_boot_layout_class` and state the rule it actually implements (splash up AND no full-page mode), rather than just redirecting. The tcss comment also keeps its "Two layouts, ONE condition" point, since that is the reasoning the redirect exists to protect.

---

**F4 — `_position_of`'s fallback docstring — FIXED** (`2cb8aaaa`)

Rewritten to stop disclaiming the stuck key the clamp makes real. It now separates the two fallbacks: the common one (identity gone, `_selected` still selectable) returns the current index and the move proceeds normally; the top is the last resort and **is** a dead end under the clamp. Records why that is accepted rather than papered over — both outcomes have already teleported the cursor away from the user's row, which is the real fault, and wrapping here would hide it by jumping to the bottom of a 60-row page. Your measurement (position 45 of 45 in the common case) is what the new text describes.

---

**F5 — the AGENTS.md exception under-specified its trigger — FIXED** (`2cb8aaaa`)

Retitled from "a long, sectioned, SCROLLED list" to **"a list that IS the whole page ... Today that is `/settings` and nothing else"**, and adds an explicit paragraph naming the pickers so the amendment cannot be read as licence to clamp them. Verified each claim in the source rather than restating yours: `model_picker.move:415` wraps, `command_picker.move:988` wraps, `session_picker._move_to:914` clamps with the near-identical comment. The trigger is now "the list is the whole page", not "the list scrolls".

---

## Gates

Whole tree, exactly as CI does, on `9f04a50e`:

| gate | result |
|---|---|
| `flake8 .` | clean |
| `black==26.1.0 --check .` | 617 files unchanged |
| `pyright` | 0 errors, 0 warnings |
| `pytest tests/unit` | 8757 passed, 25 failed |

The 25 are the sanctioned set: 24 `test_eval_tool.py` (shared-venv artifact) and `test_subagent_view.py::test_history_unavailable_and_error_retry_keep_trajectory_fallback`, a cross-test ordering artifact that passes in isolation (`1 passed`). The latter was confirmed pre-existing in a throwaway worktree at clean `a892aaf5`, which reports the identical `1 failed, 101 passed` for that file — so it is not from this change, and it is a different test from the one the earlier PR body named.

Tooling note: `qlmanage` is the documented hang on this machine, so every render ran under a bounded wrapper. No render stalled this round; all frames above were inspected as images.

damianvtran:

Implementer: `coder` (lopdev team), anthropic/claude-opus-5
Scope: `6ca77b34..9f04a50e` (reviewed head `6ca77b34`, remediation head `9f04a50e`)

Both findings taken. D1 fixed in the README, the capture script and the PR description; D2 filed as a follow-up rather than fixed here, as you recommended.

---

**D1 — the 100x30 evidence pair does not show the bug its caption claims — FIXED** (`2cb8aaaa`)

Reproduced your measurement first. The frames were correct and the words were not:

```
                  boot   boot-card  shell.w  shell.x  dock.outer  view.h
before  100x30    True   True        73       12       5           21
after   100x30    False  False       96        1       5           21
before  140x40    True   True        94       22       9           26
after   140x40    False  False      136        1       5           31
```

At 100x30 the page height, the dock inner **and** the dock outer are all identical on both sides — the composition reserves zero rows there — so there was nothing vertical for the caption to be describing. Three fixes:

**1. The captions now say what each pair demonstrates.** The 100x30 rows are labelled horizontal and explicitly note that the rows are identical on both sides and *the card is the whole change*; 140x40 keeps the compression claim, which it earns (`view.h` 26 → 31, `dock.outer` 9 → 5). The page intro no longer opens by calling the whole bug "compression" — it now describes the boot layout staying applied underneath, which costs rows, horizontal span, or both depending on size.

**2. The evidence proves its own point now.** `scripts/settings_shot.py`'s geometry line reports `boot-card`, `shell.width`, `shell.x`, `view.width`, `view.x` and `dock.outer` alongside the row counts, with a docstring recording *why* — that a rows-only geometry line is what made this pair look like it showed nothing. Both sides of every pair were re-captured with it, so the 100x30 frames now carry the numbers that make them legible as evidence.

**3. The PR description's geometry table** had the identical gap and now carries the same corrected table, plus a paragraph naming which size shows which half.

This was not only a docs problem, which is the part worth flagging: the same blind spot existed in the regression test, which passed at 100x30 on the broken tree (F1 in the code review). Both are closed.

---

**D2 — 16-row terminals paint chrome and zero setting rows — DEFERRED** (#431)

Filed as **[#431](https://github.com/damianvtran/local-operator/issues/431)**, not fixed here, agreeing with your framing: it reproduces identically on `a892aaf5`, so it is not this PR's, and fixing it means designing a vertical shedding ladder rather than making a one-line adjustment.

The ticket carries your measurement (`100x16` → `body.size=(62, 1)`, zero setting rows), the reproduce command, and a suggested direction that follows the ladder this page already has for width: shed the detail line first, then the rule, keeping the title and `esc` as invariants — the same "esc is never shed" rule `_paint_hints` states. It notes the alternative of stating a minimum height honestly instead of painting a blank band.

---

## New frames in this round

Two UX findings changed the rendered surface, so both ship with before/after pairs in the same evidence set (before = `6ca77b34`, the head you reviewed, so the pairs isolate these changes rather than re-showing the boot fix):

| | before | after |
|---|---|---|
| top of list, travelled to by held `up` (U1) | before | after |
| the read-only bottom row (U2) | before | after |

In the U1 pair the `Model` header that names the highlighted `Default provider` row is absent from the before-frame and present in the after-frame. In the U2 pair the footer goes from `↑↓ move · enter change · r default · ←→ panes · esc back to conversation` to `↑↓ move · ←→ panes · esc back to conversation` on a row that honours neither shed key.

A third evidence set covers the two sibling modes the code review's F2 brought into scope, at 100x30 and 140x40: [`docs/evidence/sibling-modes-boot-layout/`](https://github.com/damianvtran/local-operator/blob/dev-settings-fixes/docs/evidence/sibling-modes-boot-layout/README.md). The org-chart 100x30 pair is the same horizontal-only story as D1's, in a different mode.

All frames rendered to PNG and inspected as images. `qlmanage` is the documented hang on this machine so every render ran under a bounded wrapper; none stalled this round, so no frame below is being described from its markup.

damianvtran:

Implementer: `coder` (lopdev team), anthropic/claude-opus-5
Scope: `6ca77b34..9f04a50e` (reviewed head `6ca77b34`, remediation head `9f04a50e`)

All three taken and fixed, including U3, which you left to my judgement. Each was reproduced on the reviewed head first and each is pinned by a test proven to fail against that head — the "before" here is `6ca77b34`, not base, so these isolate the three fixes rather than re-showing the clamp.

Your framing was the deciding factor for taking all three rather than deferring: the clamp is what turns these ends from places users pass through into places they stop, so the ends are now where the page is read most carefully.

---

**U1 — the top gave weaker end feedback than the bottom — FIXED** (`b95aebe2`)

Reproduced exactly as reported at 100x30: `end` then 80x `up` settled on `scroll_y=1`, with row 0 — the `Model` header owning the highlighted `Default provider` row — off the top edge.

`_scroll_to_selection` now scrolls to reveal the contiguous run of headers directly above the target row rather than the row alone. Only the contiguous run, so it reveals the row's own section title and never scrolls past unrelated content.

Both ends now sit flush against their track, at every size:

```
              held `up` to top        `end` to bottom
100x30        scroll_y=0 / 46         scroll_y=46 / 46
140x40        scroll_y=0 / 36         scroll_y=36 / 36
80x24         scroll_y=0 / 52         scroll_y=52 / 52
```

`home` agrees with holding `up` (both land at 0) — the test asserts both paths, since a fix to one gesture that left the other at 1 would be the same inconsistency in a new place.

| before | after |
|---|---|
| before | after |

The before-frame shows a highlighted row with no section title above it; the after-frame shows `Model` in place. Captured with a new `top` state in `scripts/settings_shot.py` that **travels** to the top by held `up` rather than opening there, because those are different frames and only the travelled one shows this.

Test: `test_arriving_at_the_top_shows_the_section_header_that_owns_the_row`. Against `6ca77b34`: `AssertionError: the owning section header is scrolled off`.

---

**U2 — the clamped bottom row is read-only under a footer promising `enter change` — FIXED** (`b95aebe2`)

Taken rather than deferred, for the reason you gave: the clamp made this a destination. The last **six** rows of the page are `Kind.READONLY` (`session_retention_max_sessions` through `max_learnings_history`), so this is a band a user now parks in, not one row.

`enter` and `r` are shed on a read-only row, applying the rule this file already states for the pane hint and `HintButton.set_actionable` — a lit hint whose key does nothing is the "nothing happens when I click" bug one step earlier. Nothing else changes: the detail line still explains *why* the row is retired, and `enter` still reports `this setting is retired and cannot be changed` if pressed.

```
before   ↑↓  move · enter  change · r  default · ←→  panes · esc back to conversation
after    ↑↓  move · ←→  panes · esc back to conversation
```

| before | after |
|---|---|
| before | after |

One thing worth flagging, because it was a trap: the shedding ladder's narrow rungs restated `[move, enter, reset]` literally, so a terminal narrow enough to drop the pane hint would have **put the dead keys back**. The rungs are now derived from what the row actually offers. Swept every width from 140 down to 20 — the read-only row never regains `change` or `default`, `esc` is never shed, and every rung fits its terminal:

```
w=80   normal: ↑↓ move · enter change · r default · esc back to conversation
       ro:     ↑↓ move · esc back to conversation
w=40   normal: ↑↓ move · esc back
       ro:     ↑↓ move · esc back
w=20   both:   esc back
```

Test: `test_a_retired_row_stops_advertising_keys_that_cannot_act`, which also asserts an ordinary row still advertises both. Against `6ca77b34` it fails with the full footer string.

---

**U3 — `pagedown` could not reach the last row — FIXED** (`b95aebe2`)

Judged, and came down on **making paging agree with the clamp**. The reasoning:

The page already had an internal asymmetry rather than a symmetric design. `pageup` *does* reach the true first row, because the first section's first row **is** the first selectable row — so paging only disagreed with the other gestures at one end. That is an accident of where the headers fall, not a deliberate rule, and "paging is section-scoped" does not explain why one end terminates and the other stops five rows short.

```
before   pageup settles at 1  (first selectable = 1)   reaches_top=True
         pagedown settles at 54 (last selectable = 59) reaches_bottom=False
after    pagedown settles at 59                        reaches_bottom=True
```

Section-scoping is untouched: paging still jumps header to header and still clamps on the last header. The change is only that once it can advance no further by section, it finishes the journey to the last row rather than stopping in the middle of the final section. `pageup` needed no counterpart.

The last row is genuinely *on screen* afterwards, not merely selected, and paging now lands identically to `end` at every size:

```
              pagedown final   `end`            scroll_y      last row visible
100x30        row 59           row 59           46 / 46       yes
140x40        row 59           row 59           36 / 36       yes
80x24         row 59           row 59           52 / 52       yes
```

Pinned by extending `test_every_movement_on_the_page_clamps_at_the_ends`, whose paging assertion was `_selected <= last` — true of the broken behaviour too. It now asserts `== last` for `pagedown` and `== first` for `pageup`. Against `6ca77b34`: `AssertionError: pagedown stopped short of the last row`.

---

## Verification

Every claim above was measured on the remediation head at 100x30, 140x40 and 80x24, not inferred, and every test was proven to fail against `6ca77b34` in a throwaway worktree with only `settings_view.py` reverted:

```
FAILED test_every_movement_on_the_page_clamps_at_the_ends - pagedown stopped short of the last row
FAILED test_arriving_at_the_top_shows_the_section_header_... - the owning section header is scrolled off
FAILED test_a_retired_row_stops_advertising_keys_that_cannot_act - ↑↓  move · enter  change · r  default ...
3 failed
```

`tests/unit/tui/test_settings_view.py`: **364 passed**. Whole tree gates clean (`flake8`, `black==26.1.0`, `isort==5.13.2`, `pyright` 0/0); full unit suite 8757 passed with only the sanctioned pre-existing failures.

Your "verified working" areas were re-swept after these changes, since U1 and U3 both move the viewport. Compared against `6ca77b34` at all three sizes, the only cell that changed is the one U3 fixes:

```
                       arrows(bot,top)  wheel(bot,top)  paging(bot,top)  panes cycle
6ca77b34  100x30       True,True        True,True       False,True       True
9f04a50e  100x30       True,True        True,True       True,True        True
6ca77b34  80x24        True,True        True,True       False,True       False
9f04a50e  80x24        True,True        True,True       True,True        False
```

`panes cycle=False` at 80x24 is **not** a regression — it is identical on the reviewed head and is the designed narrow behaviour you documented (`action_pane` is inert when the pane does not fit, since switching a tab nobody can see reads as a broken key). The Esc ladder is untouched by these changes.

damianvtran:
### CI status on the remediation head (`9f04a50e`)

`lint`, `type-check` and `pip-audit` are **green**. Both `test` legs report **one** failure each, and it is [#418](https://github.com/damianvtran/local-operator/issues/418), the wall-clock flake that fails on `main` itself — not a regression from this branch. Recorded here rather than asserted, because "the red leg is just the known flake" is exactly the claim that deserves evidence.

**The failure, both legs, first run and rerun** — always the same single test, with everything else passing:

```
run 1  test (3.12)  1 failed, 8780 passed, 17 skipped   stalled 1517 ms
run 1  test (3.13)  1 failed, 8780 passed, 17 skipped   stalled 1627 ms
rerun  test (3.12)  1 failed, 8780 passed, 17 skipped   stalled 1528 ms
rerun  test (3.13)  1 failed, 8780 passed, 17 skipped   stalled 1173 ms

FAILED tests/unit/session/test_launch_subagent.py::test_the_loop_stays_responsive_while_several_subagents_run
  AssertionError: the event loop stalled for N ms while subagents ran
```

**The same test fails on `origin/main` at `a892aaf5`**, this PR's base, in that commit's own CI run ([33279074013](https://github.com/damianvtran/local-operator/actions/runs/33279074013)) — both legs, same assertion, nothing else failing:

```
main a892aaf5  test (3.12)  1 failed, 8767 passed, 17 skipped   stalled 1029 ms
main a892aaf5  test (3.13)  1 failed, 8767 passed, 17 skipped   stalled 1033 ms
```

Every `main` CI run since `9b6a4279` is red for this same reason, which is what #418 tracks.

**Reproduced locally as load-dependent, on both trees.** Run in isolation five times on a throwaway worktree at clean `a892aaf5`, with no change of mine present:

```
a892aaf5 (clean base)      passed, passed, FAILED, passed, FAILED   (2 of 5)
9f04a50e (this branch)     FAILED, passed, FAILED                   (2 of 3)
```

Base flakes at the same rate as the branch, so the test's outcome tracks machine load rather than anything in this diff. It asserts a fixed 1.0s loop-lateness budget against wall-clock time, which is the design flaw #418 describes.

**Nothing this branch touches is anywhere near it.** The diff is the TUI boot-layout class, the settings widget's movement and footer, evidence frames and docs; this test measures event-loop lateness while subagents run. Confirmed by the counts as well: `8780 passed` on this head against `8767` on base, the difference being the tests added this round.

Local whole-tree suite on `9f04a50e`: **8757 passed, 25 failed**, all 25 sanctioned and pre-existing (24 `test_eval_tool.py` shared-venv artifacts plus one cross-test ordering artifact that passes in isolation). Local counts differ from CI because CI runs a different selection.

PR #426, opened 2026-08-30
fix(reaper): stop the ledger lock from freezing /reload on a long session

damianvtran:
# PR Description

## Summary of Changes

`/reload` on a long-running session hangs forever after the TUI has already torn
down and restored the terminal. The user sees the shell prompt back, the process
never exits, and Ctrl+C only echoes `^C`.

### The lock did not honour its own contract

`_ledger_lock` in `local_operator/tools/group_reaper.py` documents itself as
best-effort:

> Best-effort: if the lock itself cannot be taken (permissions, an exotic FS
> without `flock`) we DEGRADE to running unlocked rather than **block** or fail a
> command — the worst case is the pre-existing lost-append behaviour, never a
> wrong kill.

The acquire under that docstring was blocking, with no `LOCK_NB` and no timeout:

```python
fcntl.flock(fd, fcntl.LOCK_EX if exclusive else fcntl.LOCK_SH)
```

So the promise held only when the lock happened to be free. Held by anyone else,
the acquire parks forever — the one outcome the docstring rules out.

The contender that never releases is a **live peer that is itself wedged** while
holding the lock, which is what was seen on this machine: 3 of 14 live `lop`
processes parked with the identical `flock`+`close` stack, two burning ~149% CPU
for 800+ minutes. One wedged process parks the next, and the freeze propagates
between sessions. Note it is **not** a lock leaked by a dead process: the kernel
drops a `flock` when the holder dies, `SIGKILL` included (measured — granted in
`0.000s` after killing the holder). A hard death leaks the `.lock` **file**, which
`_sweep_orphan_locks` already cleans; the **lock** is gone with the process.

### Which path actually parks

The callers that take this lock are `register_group` (SHARED) and
`_rewrite_without_pgid` via `unregister_group` (EXCLUSIVE) — the **bash path**.
`kill_own_groups` never takes it (it goes `_read_ledger` -> `_reap_if_still_ours`
-> `path.unlink`), measured on the pre-fix tree against a real held lock:

```
PRE-FIX register_group     returned=False after 3.00s   <- parked
PRE-FIX unregister_group   returned=False after 3.00s   <- parked
PRE-FIX kill_own_groups    returned=True after 0.01s    <- never touched the lock
```

### Why a parked acquire takes down the whole process

The bash tool registers each command's process group from an asyncio worker
thread, and `concurrent.futures` joins those threads at interpreter shutdown. One
worker parked in `flock` therefore means **the process never exits**. On macOS it
is worse still: `os.close(fd)` blocks until a sibling thread parked in `flock(fd)`
returns, so the event-loop thread's close wedges and the loop stops dead.

`/reload` is where this becomes visible, because `cli.py` `main()`'s teardown
`finally` runs after `run_tui` returns and `/reload` exits with `REEXEC_CODE(75)`
to re-exec — so the hang lands **after** the terminal is restored and **before**
`replace_self()`. `kill_own_groups` is what the user is waiting on there, but the
thread that is stuck was parked earlier, by the bash path. Ctrl+C cannot break it:
`_install_group_reaper_soft_death` deliberately excludes SIGINT (correctly — a
SIGINT is a turn abort that must spare background jobs), and the main thread is
inside a C syscall where no Python signal handler can run. Long sessions are
worse because they accumulate more registered groups and more chances of a
contended or leaked lock.

### The fix

This is the same bug class as #401 (`fix(mcp): stop the OAuth refresh lock from
freezing the TUI on /resume`, 80df237b), which fixed exactly this shape in
`local_operator/mcp/auth.py` and left `group_reaper.py` untouched. Rather than
invent a second idiom, this reuses the vetted one — `_try_lock_ledger` mirrors
`auth._try_lock_exclusive`, and `_acquire_ledger_lock` mirrors the bounded retry
in `auth._acquire_locked_fd` — with one deliberate difference: auth's helper is
exclusive-only, while the reaper needs the readers/writer distinction (SHARED for
appends so they still proceed concurrently, EXCLUSIVE for compaction), so
`exclusive` selects the mode and the distinction is kept intact.

- Every attempt is `LOCK_NB`, retried with geometric backoff to a **300ms**
  bound. The bound is sized off the HONEST contender — a sub-`PIPE_BUF` append or
  a small read-filter-rewrite, microseconds to low milliseconds — leaving ~2
  orders of magnitude of headroom while staying imperceptible. A generous timeout
  would not help against the wedged-peer case, it would only move the freeze.
  This deliberately diverges from `auth.LOCK_ACQUIRE_TIMEOUT_S` (15s): that lock
  guards a destructive rotating-token double-spend, this one guards an append
  whose loss is benign.
- `EAGAIN`/`EACCES`/`EWOULDBLOCK` is contention: degrade to unlocked, log at
  debug. Any other `OSError` is a real fault and flows into the existing degrade
  path unchanged.
- The degraded path keeps the existing semantics exactly — worst case is the
  pre-existing lost-append, never a wrong kill. **No reap gate was touched.**
- Comments record why non-blocking is required rather than stylistic (the darwin
  close-behind-flock property, the SIGINT exclusion) and carry the same explicit
  "do not simplify this back into a blocking acquire" warning as `auth.py`.

### Package-wide audit

Per the brief, every other `flock` call site was checked.
`session_lease.py:108` already uses `LOCK_NB` and is fine. After this change:

```
$ grep -rn "fcntl[.]flock" local_operator --include=*.py | grep -v LOCK_NB
local_operator/session_lease.py:133:      fcntl.flock(fd, fcntl.LOCK_UN)
local_operator/tools/group_reaper.py:394: ``fcntl.flock(fd, fcntl.LOCK_EX)``.   <- docstring warning
local_operator/tools/group_reaper.py:508: fcntl.flock(fd, fcntl.LOCK_UN)
local_operator/mcp/auth.py:1376:          ``fcntl.flock(fd, fcntl.LOCK_EX)``.   <- docstring warning
local_operator/mcp/auth.py:1459:          fcntl.flock(fd, fcntl.LOCK_UN)
local_operator/mcp/auth.py:1520:          ``fcntl.flock(fd, LOCK_EX)`` waits forever, ...
```

Only unlocks and docstring references remain. No blocking acquire survives
anywhere in the package.

## Impact

- No breaking changes, no dependency updates, no new public API.
- Version bumped to `0.43.9` — **patch**: a bug fix, no new capability. (Note the
  repo `pyproject.toml` read `0.43.5` while `0.43.8` was the published version;
  this is based on the actual latest release, confirmed against both
  `gh release list` and PyPI, so it does not regress.)

### End-to-end: the real `/reload` teardown shape

Models what actually happens — groups registered off-loop through
`asyncio.to_thread` exactly as `execute_bash` does, `kill_own_groups()` on the
loop thread as `cli.py`'s teardown `finally` does, with the lock held by a
separate live process standing in for a wedged peer. The tell is whether the
process ever reaches interpreter exit, since that is where the executor join
parks.

**BEFORE (`origin/main`, a892aaf5):**

```
phase 1: lock held by a live wedged-peer stand-in
!!! STILL RUNNING after 25s -- FREEZE REPRODUCED
```

It never reaches phase 2. This is the reported bug: terminal restored, process
alive forever, `replace_self()` never called.

**AFTER (this branch):**

```
phase 1: lock held by a live wedged-peer stand-in
phase 2: 3 groups registered off-loop in 0.36s
phase 3: kill_own_groups done; TUI torn down, terminal restored
phase 4: process exiting -> /reload would now replace_self()
process exited cleanly after 1s
```

All four phases, clean exit in 1s, with the lock still held throughout.

### The diagnosis probe

The standalone probe from the bug report: one thread calls `register_group`
against a lock held by another process, with an 8s observation window.

```
BEFORE:  probing tree: /tmp/lop-prefix/local_operator/tools
         returned=False after 8.0s          <- parked, never returned

AFTER:   probing tree: /tmp/lop-reaper-flock/local_operator/tools
         returned=True after 0.3s           <- bounded at the 300ms budget
```

### The regression test genuinely catches the bug

`test_ledger_lock_never_blocks_when_held_by_another_process` holds the lock from
a **real second process** and asserts a hard time bound on `register_group`,
`unregister_group` and `kill_own_groups`. The separate process is chosen because
it honestly models a cross-session contender and cannot be defeated by fd reuse
inside the test — not because a same-process holder would fail to contend (it
does contend: `_ledger_lock` opens a fresh fd, hence a fresh open-file
description, measured errno 35 on darwin).

Its pre-fix failure mode is a **hang, not an assertion**, so each call runs on a
daemon thread the test abandons rather than joins — a regression surfaces as a
failed `done.wait` instead of deadlocking the runner. Verified by copying the new
test file onto a clean `origin/main` worktree:

**Against the pre-fix tree:**

```
$ .venv/bin/python -m pytest tests/unit/tools/test_group_reaper.py -n0 -k never_blocks
>               assert done.wait(timeout=5.0), f"{label} parked on the contended ledger lock"
E               AssertionError: register_group parked on the contended ledger lock
1 failed, 30 deselected in 8.65s
```

**Against this branch:**

```
$ .venv/bin/python -m pytest tests/unit/tools/test_group_reaper.py tests/unit/test_session_lease.py -q
38 passed in 2.21s
```

### Gates — full tree, exactly as CI runs them

```
$ .venv/bin/python -m flake8 .
(clean)

$ uvx --from black==26.1.0 black --check .
All done! 615 files would be left unchanged.

$ uvx isort==5.13.2 --check .
(clean)

25 failed, 8746 passed, 15 skipped in 180.16s
```

**On the unit-suite failures: zero are caused by this change.** They are entirely
`tests/unit/tools/test_eval_tool.py`, which spawns kernel subprocesses that
cannot import `local_operator` from a worktree. Confirmed by running the same
suite on an unmodified `origin/main` worktree and diffing the failure sets:

```
base=24 after=24
=== ONLY IN AFTER (regressions) ===
(none)
=== ONLY IN BASE ===
(none)
```

Identical sets, and the same 24 also fail on unmodified code in the primary
checkout (`24 failed, 23 passed`), so this is a pre-existing local environment
issue with the eval kernel, not a regression. Flagged, deliberately not fixed
here — it is outside this slice.

### The freeze-guard e2e stage (round-1 finding C4)

`tui-e2e` is the stage purpose-built to catch exactly this deadlock class, and it
reported `skipping` on both CI runs: it declares `needs: [lint, type-check, test,
pip-audit]`, and `test` is red from the pre-existing flake below. So the guard was
structurally disarmed on the very PR it exists to guard.

Closing that gap by changing the workflow's `needs` would alter CI topology for
every PR in the repo, so it is **deferred to #428** rather than widened into this
bug fix. The signal itself was not skipped — the stage was run locally on macOS
at head, the platform AGENTS.md calls "the only platform where the freeze guard
can go red at all":

```
3 passed in 10.11s
```

## Checklist

- [x] Tests added for the regression, verified to fail on the pre-fix tree
- [x] Real execution evidence captured before and after, not just a green suite
- [x] All CI gates run over the whole tree
- [x] Comments record the why and the constraint per repo convention
- [x] Version bumped by materiality (patch)

### DIFF
--- a/local_operator/tools/group_reaper.py
+++ b/local_operator/tools/group_reaper.py
@@ -116,17 +116,27 @@ _CMD_LOG_CHARS = 120
 #:
 #: A generous timeout would not fix that, it would only move the freeze, so the
 #: bound is sized off the HONEST contender instead: a sub-``PIPE_BUF`` append or
-#: a small read-filter-rewrite, microseconds to low milliseconds. 0.3s leaves
-#: about two orders of magnitude of headroom before an honest contender degrades
-#: while staying imperceptible to a user. This deliberately diverges from
+#: a small read-filter-rewrite, microseconds to low milliseconds. 0.3s sits two
+#: orders of magnitude past the SLOW end of that range (100x a 3 ms rewrite, and
+#: far more against the microsecond case), so an honest contender degrades only
+#: long after it should have finished, while staying imperceptible to a user.
+#: This deliberately diverges from
 #: ``auth.LOCK_ACQUIRE_TIMEOUT_S`` (15s): that lock guards a destructive
 #: rotating-token double-spend, this one guards an append whose loss is benign.
 _LOCK_ACQUIRE_TIMEOUT_S = 0.3
 
-#: Retry cadence for the non-blocking acquire, with the same gentle geometric
-#: backoff as the MCP OAuth refresh lock: fast pickup for the common
+#: Retry cadence for the non-blocking acquire: fast pickup for the common
 #: released-in-a-moment case, then fewer wakeups once the holder is evidently
 #: not finishing within our bound.
+#:
+#: Same SHAPE as the MCP OAuth refresh lock (geometric, 1.5x growth) but
+#: deliberately re-tuned, and the values must NOT be "restored to parity" with
+#: ``auth``'s 0.05/0.25 — that would silently break this bound. auth retries
+#: against a 15s timeout; we retry against 0.3s, so its 0.25s ceiling is 83% of
+#: our ENTIRE budget. Measured: this cadence gets 10 attempts inside 0.3s where
+#: auth's would get 5, and a single 0.25s sleep step would mean one late attempt
+#: and a bound that is a coin flip rather than a schedule. The cadence is
+#: proportional to the timeout it serves; change one and re-derive the other.
 _LOCK_RETRY_SLEEP_S = 0.01
 _LOCK_RETRY_SLEEP_MAX_S = 0.05
 
--- a/local_operator/tools/group_reaper.py
+++ b/local_operator/tools/group_reaper.py
@@ -67,6 +67,7 @@ import logging
 import os
 import subprocess
 import sys
+import time
 from collections.abc import Iterator
 from dataclasses import dataclass
 from pathlib import Path
@@ -96,6 +97,26 @@ _REAPING_IS_SUPPORTED = _PLATFORM != "win32"
 #: First N chars of the command, stored for the LOG LINE only. Never a decision
 #: input. Bounded so a pathological one-line command cannot bloat the ledger.
 _CMD_LOG_CHARS = 120
+
+#: Hard bound on how long a ledger-lock acquire may retry before degrading to
+#: running unlocked. Deliberately SHORT: this is a best-effort hygiene lock on
+#: the teardown path, not a correctness-critical one (losing it costs at worst
+#: the pre-existing lost-append, never a wrong kill — see ``_ledger_lock``), and
+#: the contended case is a leaked husk from a SIGKILLed peer that will never
+#: release. A generous timeout here would not fix anything, it would only move
+#: the freeze: ``/reload`` runs ``kill_own_groups()`` AFTER the TUI has restored
+#: the terminal, so every second spent waiting is a second the user stares at a
+#: dead shell prompt. Contention is rare and momentary in the honest case (a
+#: sibling append mid-compaction, microseconds), so a few hundred ms buys the
+#: entire realistic win.
+_LOCK_ACQUIRE_TIMEOUT_S = 0.3
+
+#: Retry cadence for the non-blocking acquire, with the same gentle geometric
+#: backoff as the MCP OAuth refresh lock: fast pickup for the common
+#: released-in-a-moment case, then fewer wakeups once the holder is evidently
+#: not finishing within our bound.
+_LOCK_RETRY_SLEEP_S = 0.01
+_LOCK_RETRY_SLEEP_MAX_S = 0.05
 
 
 @dataclass(frozen=True)
@@ -351,6 +372,68 @@ def unregister_group(
     _rewrite_without_pgid(path, pgid)
 
 
+def _try_lock_ledger(fd: int, *, exclusive: bool) -> bool:
+    """ONE non-blocking attempt at the ledger lock. True when taken.
+
+    Mirrors :func:`local_operator.mcp.auth._try_lock_exclusive` — the same idiom
+    for the same reason — but keeps this module's readers/writer distinction:
+    ``exclusive`` selects ``LOCK_EX`` for a compacting rewrite and ``LOCK_SH``
+    for an append, so concurrent appends still proceed together.
+
+    Non-blocking is REQUIRED here, not a style choice. ``LOCK_NB`` is what keeps
+    the caller off the kernel wait queue, and on macOS/BSD a thread parked in
+    ``flock(fd)`` blocks any other thread's ``os.close(fd)`` until that
+    ``flock()`` returns — which, against a lock leaked by a SIGKILLed peer, is
+    never. Reached from ``kill_own_groups()`` on the ``/reload`` teardown path
+    that runs after the terminal is restored, a parked acquire wedges the event
+    loop with no TUI left to show it and no SIGINT able to break it (the soft
+    death handler deliberately spares SIGINT, and the main thread is inside a C
+    syscall where no Python handler can run). This is the same bug shape as #401
+    in the MCP OAuth refresh lock. Retrying a non-blocking attempt is strictly
+    weaker than blocking and is the only shape that cannot hang, so do NOT
+    "simplify" this back into ``fcntl.flock(fd, fcntl.LOCK_EX)``.
+
+    Contention is the expected outcome and returns False, letting the caller
+    degrade to running unlocked. Anything else is a real fault that retrying
+    cannot fix, so it is raised for the caller's existing degrade path.
+    """
+    import errno
+    import fcntl
+
+    mode = fcntl.LOCK_EX if exclusive else fcntl.LOCK_SH
+    try:
+        fcntl.flock(fd, mode | fcntl.LOCK_NB)
+        return True
+    except OSError as lock_err:
+        if lock_err.errno in (errno.EAGAIN, errno.EACCES, errno.EWOULDBLOCK):
+            return False
+        raise
+
+
+def _acquire_ledger_lock(fd: int, *, exclusive: bool) -> bool:
+    """Retry the non-blocking acquire to ``_LOCK_ACQUIRE_TIMEOUT_S``. Worker-safe.
+
+    Returns True with the lock held, or False when the bound elapsed under
+    contention — the caller then runs unlocked, which is the documented degrade
+    (see :func:`_ledger_lock`). A real fault propagates as ``OSError`` into the
+    caller's existing degrade path; retrying could not fix it.
+
+    Every attempt is non-blocking, so this function can be abandoned by nothing
+    worse than the bound: unlike a blocking acquire it never parks a thread in
+    the kernel, and therefore never makes a sibling's ``os.close()`` of this fd
+    hang on macOS.
+    """
+    deadline = time.monotonic() + _LOCK_ACQUIRE_TIMEOUT_S
+    sleep_s = _LOCK_RETRY_SLEEP_S
+    while True:
+        if _try_lock_ledger(fd, exclusive=exclusive):
+            return True
+        if time.monotonic() >= deadline:
+            return False
+        time.sleep(sleep_s)
+        sleep_s = min(sleep_s * 1.5, _LOCK_RETRY_SLEEP_MAX_S)
+
+
 @contextlib.contextmanager
 def _ledger_lock(path: Path, *, exclusive: bool) -> Iterator[None]:
     """Advisory ``flock`` serialising ledger COMPACTION against APPENDS.
@@ -370,20 +453,45 @@ def _ledger_lock(path: Path, *, exclusive: bool) -> Iterator[None]:
     in flight while it truncates and rewrites.
 
     Best-effort: if the lock itself cannot be taken (permissions, an exotic FS
-    without ``flock``) we DEGRADE to running unlocked rather than block or fail a
-    command — the worst case is the pre-existing lost-append behaviour, never a
-    wrong kill. POSIX only; reached solely after the ``_REAPING_IS_SUPPORTED``
-    guard, so ``fcntl`` (absent on Windows) is imported lazily here.
+    without ``flock``, or simply CONTENTION) we DEGRADE to running unlocked
+    rather than block or fail a command — the worst case is the pre-existing
+    lost-append behaviour, never a wrong kill. POSIX only; reached solely after
+    the ``_REAPING_IS_SUPPORTED`` guard, so ``fcntl`` (absent on Windows) is
+    imported lazily here.
+
+    The acquire is BOUNDED and NON-BLOCKING, which is what makes that promise
+    true rather than aspirational. It once was not: a bare blocking ``flock``
+    here honoured the docstring only when the lock happened to be free, and a
+    husk leaked by a SIGKILLed peer froze the whole process on ``/reload``.
+    :func:`_try_lock_ledger` carries the full reasoning, including why macOS
+    turns a parked acquire into a dead event loop. Do not reintroduce a blocking
+    acquire, and do not stretch ``_LOCK_ACQUIRE_TIMEOUT_S`` into a long wait —
+    that only relocates the hang.
+
+    The fd is opened, retried on, and closed all on the CALLING thread, so no
+    other thread can ever close a descriptor this one is inside a lock call on
+    (the fd-ownership rule from ``auth._acquire_locked_fd``). Here that holds by
+    construction: the acquire never blocks, so the descriptor is never left with
+    an outstanding lock syscall on it.
     """
     fd: int | None = None
     try:
         try:
-            import fcntl
-
             path.parent.mkdir(parents=True, exist_ok=True, mode=0o700)
             lock_path = _lock_path(path)
             fd = os.open(str(lock_path), os.O_CREAT | os.O_RDWR, 0o600)
-            fcntl.flock(fd, fcntl.LOCK_EX if exclusive else fcntl.LOCK_SH)
+            if not _acquire_ledger_lock(fd, exclusive=exclusive):
+                # Contended past the bound: run unlocked. Debug, not warning —
+                # this is a tolerated outcome on a best-effort hygiene lock, and
+                # the teardown path must stay silent on a restored terminal.
+                logger.debug(
+                    "group reaper: ledger lock contended for %s after %.2fs; proceeding unlocked",
+                    path,
+                    _LOCK_ACQUIRE_TIMEOUT_S,
+                )
+                with contextlib.suppress(OSError):
+                    os.close(fd)
+                fd = None
         except (OSError, ImportError) as exc:
             # Degrade to unlocked: never block or break a command over the lock.
             logger.debug("group reaper: ledger lock unavailable for %s: %s", path, exc)
--- a/local_operator/tools/group_reaper.py
+++ b/local_operator/tools/group_reaper.py
@@ -120,9 +120,9 @@ _CMD_LOG_CHARS = 120
 #: orders of magnitude past the SLOW end of that range (100x a 3 ms rewrite, and
 #: far more against the microsecond case), so an honest contender degrades only
 #: long after it should have finished, while staying imperceptible to a user.
-#: This deliberately diverges from
-#: ``auth.LOCK_ACQUIRE_TIMEOUT_S`` (15s): that lock guards a destructive
-#: rotating-token double-spend, this one guards an append whose loss is benign.
+#: This deliberately diverges from ``auth.LOCK_ACQUIRE_TIMEOUT_S`` (15s): that
+#: lock guards a destructive rotating-token double-spend, this one guards an
+#: append whose loss is benign.
 _LOCK_ACQUIRE_TIMEOUT_S = 0.3
 
 #: Retry cadence for the non-blocking acquire: fast pickup for the common
--- a/local_operator/tools/group_reaper.py
+++ b/local_operator/tools/group_reaper.py
@@ -99,16 +99,28 @@ _REAPING_IS_SUPPORTED = _PLATFORM != "win32"
 _CMD_LOG_CHARS = 120
 
 #: Hard bound on how long a ledger-lock acquire may retry before degrading to
-#: running unlocked. Deliberately SHORT: this is a best-effort hygiene lock on
-#: the teardown path, not a correctness-critical one (losing it costs at worst
-#: the pre-existing lost-append, never a wrong kill — see ``_ledger_lock``), and
-#: the contended case is a leaked husk from a SIGKILLed peer that will never
-#: release. A generous timeout here would not fix anything, it would only move
-#: the freeze: ``/reload`` runs ``kill_own_groups()`` AFTER the TUI has restored
-#: the terminal, so every second spent waiting is a second the user stares at a
-#: dead shell prompt. Contention is rare and momentary in the honest case (a
-#: sibling append mid-compaction, microseconds), so a few hundred ms buys the
-#: entire realistic win.
+#: running unlocked. Deliberately SHORT: this is a best-effort hygiene lock, not
+#: a correctness-critical one (losing it costs at worst the pre-existing
+#: lost-append, never a wrong kill — see ``_ledger_lock``).
+#:
+#: The contender worth sizing for is a LIVE peer that is itself wedged, holding
+#: the lock while parked. Note it is NOT a lock leaked by a dead process:
+#: ``flock`` is dropped by the kernel when the last fd on the open-file
+#: description closes, and that includes a ``SIGKILL`` (measured — after killing
+#: the holder the lock is granted in 0.000s). What a hard death leaks is the
+#: ``.lock`` FILE, which ``_sweep_orphan_locks`` cleans; the LOCK itself is
+#: already gone. A wedged live holder is the case that never releases, and it
+#: was observed on this machine: 3 of 14 live ``lop`` processes parked in
+#: ``flock``+``close``, two burning ~149% CPU for 800+ minutes. One wedged
+#: process parks the next, and the freeze propagates between sessions.
+#:
+#: A generous timeout would not fix that, it would only move the freeze, so the
+#: bound is sized off the HONEST contender instead: a sub-``PIPE_BUF`` append or
+#: a small read-filter-rewrite, microseconds to low milliseconds. 0.3s leaves
+#: about two orders of magnitude of headroom before an honest contender degrades
+#: while staying imperceptible to a user. This deliberately diverges from
+#: ``auth.LOCK_ACQUIRE_TIMEOUT_S`` (15s): that lock guards a destructive
+#: rotating-token double-spend, this one guards an append whose loss is benign.
 _LOCK_ACQUIRE_TIMEOUT_S = 0.3
 
 #: Retry cadence for the non-blocking acquire, with the same gentle geometric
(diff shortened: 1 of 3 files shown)


damianvtran:
### Agent review — round 1

Scope: a892aaf543aaf4fbe804461b6c51b61c41277971..666ee1d40526deb0a241e6fd623e10269c3ab4a1

I re-derived the mechanism claims and re-ran the evidence rather than reading the
description. **The fix is correct and I found no blocker and no major.** The
degrade path is sound, the reap gates are untouched, the bound is well chosen,
and the regression test genuinely fails pre-fix. Findings below are all minor or
nit, and three of them are comments/description that assert things I measured to
be false — worth correcting because this module's comments are load-bearing for
the next person who touches the lock.

**What I verified independently (not taken on trust):**

- **The macOS close-behind-flock premise is REAL.** Standalone probe: external
  process holds `LOCK_EX`, a sibling thread parks in blocking `flock(fd)`, then
  `os.close(fd)` on another thread → `close_blocked = True, platform = darwin`.
  The comment's central claim is accurate.
- **The regression test genuinely fails on the pre-fix tree.** Copied
  `tests/unit/tools/test_group_reaper.py` from head onto a clean `a892aaf5`
  worktree: `AssertionError: register_group parked on the contended ledger lock`,
  `1 failed in 6.48s`. It is not a test that passes against the regression.
- **The degrade path cannot cause a wrong kill.** 60 forced unlocked
  append-vs-compaction races: `lost-appends=0, unparseable-lines=0`, and by
  construction a lost append can only make an entry MISSING (missed reap), never
  add a spurious one. `_reap_if_still_ours` two-factor identity + pgid-reuse gate
  and `sweep_orphan_groups`' truth table are byte-identical to base — confirmed by
  diffing; no reap gate was weakened. The readers/writer split (`LOCK_SH` append
  vs `LOCK_EX` compaction) is intact and still meaningful: `exclusive` selects the
  mode inside the bounded acquire.
- **Sibling audit is clean at head.** `grep -rn "fcntl\.flock" local_operator | grep -v LOCK_NB`
  leaves only `LOCK_UN` calls and docstring prose. `session_lease.py:108` was
  already `LOCK_EX | LOCK_NB` with no retry, and its Windows `msvcrt.locking` uses
  `LK_NBLCK`. No blocking primitive remains on a teardown path.
- **Reaper tests green at head:** `tests/unit/tools/test_group_reaper.py` +
  `tests/unit/test_session_lease.py` → `38 passed in 2.12s`.
- **Gates:** flake8 clean, black 26.1.0 `615 files unchanged`, isort clean,
  pyright `0 errors` on the changed module.
- **0.3s is the right bound.** Honest contention here is a sub-`PIPE_BUF` append
  or a small read-filter-rewrite — microseconds to low milliseconds — so 0.3s is
  ~2 orders of magnitude of headroom before an honest contender degrades, while
  staying imperceptible on a restored terminal. Deliberately diverging from
  `auth.LOCK_ACQUIRE_TIMEOUT_S = 15.0` is correct: auth guards a destructive
  rotating-token double-spend, this guards a hygiene append whose loss is benign.

---

**C1 — minor — `local_operator/tools/group_reaper.py:387-391`: the docstring's
motivating example is the one path that never takes this lock.**

The comment says a parked acquire is "Reached from `kill_own_groups()` on the
`/reload` teardown path". `kill_own_groups` does not call `_ledger_lock` at all —
it goes `_read_ledger` → `_reap_if_still_ours` → `path.unlink`. Only
`register_group` (SHARED) and `_rewrite_without_pgid` (EXCLUSIVE) acquire it. I
confirmed this by AST-walking the head file, and empirically on the pre-fix tree
against a real held lock:

```
PRE-FIX register_group     returned=False after 3.00s   <- parked
PRE-FIX unregister_group   returned=False after 3.00s   <- parked
PRE-FIX kill_own_groups    returned=True after 0.00s    <- never touched the lock
```

The bug is real and the fix is right, but the freeze actually enters through
`register_group`/`unregister_group` on the bash path, not through
`kill_own_groups`. Leaving this in misdirects the next reader to a function that
cannot reproduce it. One-line fix: name `register_group`/`unregister_group` as the
reaching callers, and keep `kill_own_groups` only as the teardown context in which
the already-parked worker thread wedges the exit.

**C2 — minor — same hunk, and the PR description: "a husk leaked by a SIGKILLed
peer" is not a real mechanism; `flock` is released on process death.**

This premise appears in `_LOCK_ACQUIRE_TIMEOUT_S`'s comment ("the contended case
is a leaked husk from a SIGKILLed peer that will never release"), in
`_try_lock_ledger`, in `_ledger_lock`, in the test docstring, and in the PR body.
Measured: hold `LOCK_EX`, `SIGKILL` the holder, then acquire → granted in `0.000s`.
The kernel drops flocks when the last fd on the open-file-description closes, and
`SIGKILL` closes them. What actually leaks is the `.lock` **file** (which
`_sweep_orphan_locks` cleans), not the **lock**. The real never-releasing
contender is a *live* peer that is itself wedged — which is exactly what was seen
on this machine (3 of 14 `lop` processes parked in `flock`+`close`, burning ~149%
CPU for 800+ min): one wedged process holding the lock parks the next one, and the
freeze propagates. That is a stronger justification than the incorrect one, and it
is the case the 0.3s bound is genuinely sized for. Worth correcting in all five
places since the file's comments are treated as authority here.

**C3 — minor — `tests/unit/tools/test_group_reaper.py`, `_hold_ledger_lock`
docstring: the stated reason for using a second process is wrong (the choice
itself is right).**

It claims "a same-process holder would prove nothing, since flock is per
open-file-description and the reaper's own fd would be granted the lock
immediately". Measured on darwin: a second `os.open` in the SAME process, then
`flock(LOCK_EX|LOCK_NB)` → `errno 35 EAGAIN`, contended. `_ledger_lock` always
`os.open`s a fresh fd, so it gets a fresh open-file-description and DOES contend
with a same-process holder. The separate process is still the better construction
(it models the real cross-session case and cannot be defeated by fd reuse), but
the justification as written is falsifiable in three lines and will mislead
someone who later "simplifies" the fixture.

**C4 — minor — the freeze-guard CI stage this bug class exists for did not run on
this PR, and `needs: test` means it cannot while `test` is red.**

`tui-e2e` is `skipped` on both runs (33284918639, 33284954232) because it declares
`needs: [lint, type-check, test, pip-audit]` and `test` failed. Per AGENTS.md that
macOS leg is "the only platform where the freeze guard can go red at all" — so the
one check purpose-built to catch exactly this deadlock provided zero signal here.
Not this PR's defect, and I closed the gap manually: `tests/e2e -m e2e -n0` on
macOS at head → **`3 passed in 10.42s`**. Flagging because the gate is structurally
disarmed whenever `test` is flaky, which is currently always.

**C5 — nit — the CI red on this PR is pre-existing, with one caveat I chased
down.** `test_the_loop_stays_responsive_while_several_subagents_run` fails
identically on unmodified `main` (run 33279074013, 1029/1033 ms) — a timing-flaky
loop-stall assertion, unrelated. Run 33284954232 additionally shows
`test_redundant_roster_events_skip_the_sidecar_write - assert 2 == 1`, which is
NOT in the main run; I checked it directly and it passes at head AND at base
locally (`1 passed` each), so it is flake, not a regression. No CI failure is
attributable to this change, and the `eval_tool` worktree `ModuleNotFoundError`
noise is the known editable-venv artifact.

**Version bump.** `0.43.8` → `0.43.9` is correct. PyPI currently serves `0.43.8`,
so this advances rather than regresses. Per AGENTS.md "Versioning" this is a
reliability/bug fix with no new user-facing capability to name in a title —
textbook patch.

**Verdict: clean.** No blocker, no major. The degrade path preserves the
"never kill a live owner's groups" property, the two-factor owner identity and
pgid-reuse guards are untouched, no sibling blocking lock remains, the bound is
well argued, and the regression test is a real guard that fails against the
regression. This round is **TERMINAL** — C1–C3 are comment-accuracy corrections
and C4–C5 are CI observations, all follow-ups rather than grounds for another
round. **Clear to merge from my side.**

damianvtran:

Scope: a892aaf543aaf4fbe804461b6c51b61c41277971..a9e74e83f7a5e154200af3c636a962084c1bef42
Remediation commit: a9e74e83f7a5e154200af3c636a962084c1bef42

I re-derived each measured claim myself rather than accepting it, since all three
comment findings turn on a measurement. All three reproduced exactly as reported.
No code behaviour changed in this round — C1–C3 are comment corrections, and the
reaper tests still pass at head while the regression test still fails against the
pre-fix tree.

---

**C1 — minor — the docstring's motivating example is the one path that never
takes this lock. FIXED in a9e74e83.**

Confirmed independently on the pre-fix tree against a real held lock:

```
PRE-FIX register_group       returned=False after 3.00s   <- parked
PRE-FIX unregister_group     returned=False after 3.00s   <- parked
PRE-FIX kill_own_groups      returned=True after 0.01s    <- never touched the lock
```

You are right, and this also corrects part of the original briefing I was given —
I followed the measurement over the prompt. `_try_lock_ledger` now names the real
callers (`register_group` SHARED, `_rewrite_without_pgid` via `unregister_group`
EXCLUSIVE — the bash path) and states explicitly that `kill_own_groups` does NOT
take this lock, keeping it only as the teardown context where the damage becomes
visible: the parked worker threads are joined by `concurrent.futures` at
interpreter shutdown, which is why the process cannot exit on `/reload`. The same
correction is applied to the test docstring and the PR description.

**C2 — minor — "a husk leaked by a SIGKILLed peer" is not a real mechanism.
FIXED in a9e74e83.**

Measured on darwin, matching your result:

```
C2: while holder ALIVE      -> errno 35 (contended, as expected)
C2: after SIGKILL of holder -> granted in 0.000s  => flock IS released on death
```

The lock is dropped when the last fd on the open-file description closes, and
`SIGKILL` closes them; what a hard death leaks is the `.lock` **file**, which
`_sweep_orphan_locks` already cleans. Corrected in all five places
(`_LOCK_ACQUIRE_TIMEOUT_S`, `_try_lock_ledger`, `_ledger_lock`, the test helper,
and the PR body) to name the true contender — a **live peer wedged while holding
the lock**, as observed on this machine (3 of 14 live `lop` processes parked in
`flock`+`close`, two burning ~149% CPU for 800+ minutes; one wedged process parks
the next and the freeze propagates).

I also took your `0.3s` reasoning into the constant's comment, since the old text
justified the bound off the incorrect premise. It is now sized off the **honest**
contender (a sub-`PIPE_BUF` append or small read-filter-rewrite, microseconds to
low milliseconds — ~2 orders of magnitude of headroom), and records why it
deliberately diverges from `auth.LOCK_ACQUIRE_TIMEOUT_S = 15.0`: that lock guards
a destructive rotating-token double-spend, this one an append whose loss is
benign.

**C3 — minor — the test helper's stated reason for using a second process is
falsifiable. FIXED in a9e74e83.**

Confirmed: a second `os.open` in the same process then `flock(LOCK_EX|LOCK_NB)`
gives `errno 35`, so a same-process holder does contend — `_ledger_lock` always
opens a fresh fd and therefore a fresh open-file description.

```
C3: same-process fresh fd -> errno 35 => it DOES contend
```

The construction is unchanged; only the rationale is rewritten. It now says the
separate process is chosen because it honestly models a cross-session contender
and cannot be defeated by fd reuse inside the test, and explicitly notes a
SIGKILLed holder would NOT serve (per C2).

**C4 — minor — `tui-e2e` was skipped, so the freeze guard gave no signal.
DEFERRED to #428, with the signal obtained manually.**

Agreed and not silently dropped. Changing the `needs` clause alters CI topology
for every PR in the repo, which does not belong inside an unrelated bug fix, so
it is filed as **#428** with the three options laid out (drop `test` from
`needs`, drop `needs` entirely, or fix the flake first) and the evidence from
this PR as the motivating case.

The signal itself was not skipped — I ran the stage locally on macOS at head,
independently of your run:

```
3 passed in 10.11s
```

That matches your `3 passed in 10.42s`. The PR description now records the C4
decision and links #428.

**C5 — nit — pre-existing CI flake. ACKNOWLEDGED, no action.**

Confirmed settled and not attributable to this change: the failing tests pass
locally at head (`35 passed`), `test_the_loop_stays_responsive_while_several_subagents_run`
fails on unmodified `main` at the merge base a892aaf5 (run 33279074013), and the
`eval_tool` `ModuleNotFoundError` noise is the known editable-venv worktree
artifact — the failure sets on an unmodified `origin/main` worktree and this
branch are identical (`base=24 after=24`, zero either side). I have stopped
re-running CI on it.

---

**Gates re-run after the remediation commit** (whole tree, as CI runs them):

```
$ .venv/bin/python -m flake8 .
(clean)

$ uvx --from black==26.1.0 black --check .
All done! 615 files would be left unchanged.

$ uvx isort==5.13.2 --check .
(clean)

    tests/unit/tools/test_group_reaper.py tests/unit/test_session_lease.py -q
38 passed in 2.05s
```

**The regression test is still a real guard** after the rewording — re-copied
onto a clean `a892aaf5` worktree:

```
1 failed, 30 deselected in 5.10s
AssertionError: register_group parked on the contended ledger lock
```

No blocker or major findings were open, and C1–C3 are now fixed at head, so from
my side this is answered in full. Not merging — that decision is the delegator's.

damianvtran:
### Agent review — round 2

Reviewer: reviewer subagent (claude-opus-4-6)
Scope: 666ee1d40526deb0a241e6fd623e10269c3ab4a1..a9e74e83f7a5e154200af3c636a962084c1bef42

Round 2 exists to close the freshness gap: round 1 reviewed `666ee1d4`, and the
remediation commit `a9e74e83` moved head. I re-measured every corrected claim
myself rather than accepting the author's re-measurement.

**No blocker, no major. All three round-1 comment-accuracy findings (C1, C2, C3)
are genuinely fixed, and the fix logic is untouched.**

#### The delta is provably docs-only

Not asserted from the diff shape — verified by parsing both revisions and
comparing the ASTs with docstrings stripped:

```
local_operator/tools/group_reaper.py:  AST-identical-ignoring-docstrings = True
tests/unit/tools/test_group_reaper.py: AST-identical-ignoring-docstrings = True
```

Every changed line is a `#:` comment or a docstring. No executable line moved, so
round 1's substantive verdict on the fix carries forward intact.

#### The corrected claims are true (independently re-measured on darwin)

C1 — **which caller reaches the lock.** Confirmed by static call-graph walk over
the module (following `with` context managers, not just plain calls):

```
kill_own_groups:     does NOT reach _ledger_lock
register_group:      register_group -> _ledger_lock
unregister_group:    unregister_group -> _rewrite_without_pgid -> _ledger_lock
sweep_orphan_groups: does NOT reach _ledger_lock
```

The new wording naming `register_group` (SHARED) and `_rewrite_without_pgid` via
`unregister_group` (EXCLUSIVE) as the parking callers is correct, and correctly
demotes `kill_own_groups` to "where the damage becomes visible".

C2 — **flock is released on process death.** Measured directly:

```
CONTENTION while holder LIVE:  yes, errno=35 (EAGAIN)
AFTER SIGKILL of holder:       blocking flock granted in 0.0000s
lock FILE still exists after holder death: True
```

So the old "husk leaked by a SIGKILLed peer" mechanism was indeed false, and the
new split is exactly right: a hard death leaks the `.lock` FILE (which
`_sweep_orphan_locks` cleans, verified at line 624) while the LOCK is dropped by
the kernel. The live-wedged-holder framing is the honest contender.

C3 — **a same-process fresh fd does contend.** Measured:

```
SAME-PROCESS fresh fd:  contends, errno=35 (EAGAIN)
SAME-PROCESS dup'd fd:  granted (same open-file description)
```

The test helper's old rationale was falsifiable; the new wording ("it is not that
it would prove nothing — the separate process is the honest model of a
cross-session contender") states the true reason without overclaiming.

I also spot-checked the one load-bearing claim the comments rest on that round 1
did not measure — that a parked worker prevents interpreter exit:

```
parked to_thread worker, main() returns after 0.2s -> process wall time 30.07s
-> interpreter shutdown JOINED the parked worker: True
```

That confirms "one parked worker means the process never exits" is a real
mechanism, not a plausible story.

#### The test is still a real guard after rewording

Copied the head test file onto a clean `a892aaf5` worktree (fix absent, test
present) and ran it. It fails there with precisely the claimed failure:

```
$ .venv/bin/python -m pytest tests/unit/tools/test_group_reaper.py -n0 -k never_blocks -q
E   AssertionError: register_group parked on the contended ledger lock
1 failed, 30 deselected in 6.29s
```

The rewording did not soften the assertion; `kill_own_groups` is still asserted
(correctly, as a must-STAY-prompt guard rather than as the parking caller).

#### PR description

The body no longer asserts the disproved mechanism. It now states the kernel
drops a `flock` on holder death "SIGKILL included (measured — granted in
`0.000s`)", names the bash path as the parking one, and carries the pre-fix
measurement showing `kill_own_groups returned=True after 0.01s <- never touched
the lock`. Consistent with what I measured.

#### C4 deferral

Issue #428 exists, is OPEN, and is a substantive write-up rather than a
placeholder — it names the concrete `needs:` clause, cites the two skipped
`tui-e2e` runs on this PR, and lists three options with a recommendation.
Deferring a repo-wide CI topology change out of a bug fix is the right call. I
independently obtained the manual e2e signal on macOS at head:

```
3 passed in 11.41s
```

#### Gates at `a9e74e83` (run by me, whole tree)

```
uvx --from black==26.1.0 black --check .  615 files unchanged
uvx isort==5.13.2 --check .               clean
pytest tests/unit/tools/test_group_reaper.py tests/unit/test_session_lease.py -q
                                          38 passed in 2.33s
pytest tests/e2e -m e2e -n0 -q            3 passed in 11.41s
```

The ~24-25 `test_eval_tool.py` / TUI failures reproduce on unmodified
`origin/main` and are not attributable to this PR, as round 1 established.

#### Findings

**C6 — nit — the 0.3s headroom claim is loosely stated.**
`local_operator/tools/group_reaper.py:118-121` says the bound is sized off an
honest contender of "microseconds to low milliseconds" and that 0.3s "leaves
about two orders of magnitude of headroom". Against "low milliseconds" (say
1-3ms) that is two orders; against "microseconds" it is five. The range and the
figure do not pin to each other. Harmless — the sizing decision is sound and the
divergence-from-`auth.LOCK_ACQUIRE_TIMEOUT_S` contrast (15s, destructive
double-spend vs a benign lost append) is the genuinely useful part. Would read
truer as "two orders of magnitude past the slow end of that range". No action
needed to merge.

**C7 — nit — retry cadence comment claims parity it does not have.**
`group_reaper.py:127-131` describes the backoff as "the same gentle geometric
backoff as the MCP OAuth refresh lock". The 1.5x growth factor matches, but both
constants differ: `_LOCK_RETRY_SLEEP_S` is 0.01 here vs 0.05 in `auth.py:129`,
and `_LOCK_RETRY_SLEEP_MAX_S` is 0.05 here vs 0.25 in `auth.py:137`. The
divergence is correct and deliberate (a 0.3s bound cannot afford a 0.25s sleep
step), so this is the same *shape*, deliberately re-tuned — worth one clause
saying so, since the next reader may otherwise "restore parity" and break the
bound. Follow-up at most.

Note on the earlier `PIPE_BUF` wording at lines 118, 343 and 470: it predates
this PR (present verbatim at base `a892aaf5`), so it is out of scope here and I
raise no finding on it.

#### Verdict

**clean.** No blocker, no major. The remediation does exactly what it claims and
nothing more: the comments now match measured reality, the fix is byte-for-byte
unchanged in behaviour, the regression test still fails on the pre-fix tree, and
every gate is green at head. C6 and C7 are nits and are follow-ups, not grounds
for another round.

**This round is TERMINAL. Clear to merge from my side** once CI `test (3.12)` /
`test (3.13)` report (they were still pending at review time; `lint`,
`type-check` and `pip-audit` all pass).

damianvtran:

Scope: a9e74e83f7a5e154200af3c636a962084c1bef42..7e3bf99b45ade47f4b9361f28b6d51d55c986e94
Remediation commits: 2d5f99e723cd5a56b657dc41d4128a2c0a90f633, 7e3bf99b45ade47f4b9361f28b6d51d55c986e94

Both nits fixed. The round-2 delta is comment-only, by the same AST test you used
on round 1 — parsed with docstrings stripped, both files are identical to the
round-1 head:

```
local_operator/tools/group_reaper.py
  AST identical vs round-1 head a9e74e83 = True
tests/unit/tools/test_group_reaper.py
  AST identical vs round-1 head a9e74e83 = True
```

---

**C6 — nit — the 0.3s headroom claim is loosely stated. FIXED in 2d5f99e7.**

You are right that the figure and the range did not pin to each other. Computed
across the cited range:

```
0.3s / 1us = 300,000x  (~5.5 orders)
0.3s / 1ms =     300x  (~2.5 orders)
0.3s / 3ms =     100x  (~2.0 orders)
```

So "two orders" is true only against the slow end. Reworded to say exactly that,
with the concrete anchor rather than a bare order-of-magnitude claim:

> 0.3s sits two orders of magnitude past the SLOW end of that range (100x a 3 ms
> rewrite, and far more against the microsecond case), so an honest contender
> degrades only long after it should have finished, while staying imperceptible
> to a user.

**C7 — nit — retry cadence comment claims parity it does not have. FIXED in
2d5f99e7.**

Agreed this is the one that matters, since the failure mode is a future edit
"restoring parity" and silently breaking the bound. Confirmed both constants
diverge (0.01/0.05 here vs `auth.py`'s 0.05/0.25 at lines 129 and 137) while the
1.5x factor is shared, and measured what the divergence actually buys by
simulating both cadences against a 0.3s bound:

```
ours   (0.01/0.05): 10 attempts within 0.3s, sleeps [0.01, 0.015, 0.022, 0.034, 0.05, 0.05, ...]
auth's (0.05/0.25):  5 attempts within 0.3s, sleeps [0.05, 0.075, 0.113, 0.169]
auth max sleep 0.25s = 83% of our entire 0.3s bound
```

The comment no longer claims parity, and now records the constraint rather than
just the observation:

> Same SHAPE as the MCP OAuth refresh lock (geometric, 1.5x growth) but
> deliberately re-tuned, and the values must NOT be "restored to parity" with
> `auth`'s 0.05/0.25 — that would silently break this bound. auth retries against
> a 15s timeout; we retry against 0.3s, so its 0.25s ceiling is 83% of our ENTIRE
> budget. Measured: this cadence gets 10 attempts inside 0.3s where auth's would
> get 5, and a single 0.25s sleep step would mean one late attempt and a bound
> that is a coin flip rather than a schedule. The cadence is proportional to the
> timeout it serves; change one and re-derive the other.

The closing clause is the actual guard: it tells the next reader the two numbers
are coupled, so changing the timeout without re-deriving the cadence is the edit
to catch in review.

`7e3bf99b` is a follow-on reflow only — the C6 rewording left two comment lines
badly wrapped. No wording changed.

**On the `PIPE_BUF` wording at lines 118, 343 and 470:** agreed it predates this
PR (verbatim at base `a892aaf5`) and I have left it alone rather than widening
the slice. Noting it here so it is not lost if someone wants it as a follow-up.

---

**Gates re-run at head `7e3bf99b`** (whole tree, as CI runs them):

```
$ .venv/bin/python -m flake8 .
(clean)

$ uvx --from black==26.1.0 black --check .
615 files would be left unchanged.

$ uvx isort==5.13.2 --check .
(clean)

    tests/unit/tools/test_group_reaper.py tests/unit/test_session_lease.py -q
38 passed in 3.73s
```

Both rounds' substantive findings are terminal, C6 and C7 are now fixed, and the
delta since your review is provably docs-only. Not merging — that is the
delegator's call.

PR #432, opened 2026-08-30
feat(prompts): teach agents that the web is a verification surface

damianvtran:
# PR Description

## Summary of Changes

lop agents rarely reached for `web_search` / `web_fetch` unless a user asked
explicitly, so a third-party error message or a design question got answered
from pretraining when the answer was a lookup away.

**Nothing was suppressing them.** Both tools are in `DEFAULT_TOOL_NAMES`
(`tools/registry.py:70-71`) and `enabled: bool = True` by default
(`web_search/models.py:72`). They are built, advertised, and callable. This is
not a gating bug.

### The prompt defined verification as a closed, local-only act

`prompts_md/system.md` had **zero** occurrences of "web", "internet",
"research" or "look up" across 207 lines. Worse than unmentioned, the web was
implicitly excluded, in two places:

The persona paragraph (`system.md:2-5`):

> you have tools for running shell commands, reading and editing files,
> **searching the workspace**, tracking tasks, and scheduling follow-ups. Use
> them before answering whenever the answer depends on **the machine's state**

The first working principle (`system.md:19-21`):

> **Use tools before answering.** Verify with a tool rather than assuming: run
> the command, read the file, **search the workspace**.

Both are complete-sounding enumerations whose members are all local, and the
first scopes tool use to "the machine's state" — which reads as a licence to
answer from recall whenever the question is *not* about this machine. An agent
following either faithfully never reaches the web, because by the definition it
was handed **it has already verified**. That is the whole bug: not a missing
advertisement, a wrong definition.

### The fix

Widen both enumerations, and add one bullet to `## Working principles` naming
the trigger. The trigger is deliberately **self-observational** — "you notice
you are guessing about something outside this machine" — rather than a topic
list, because a topic list invites keyword-matching and generalises to nothing
outside it.

Three clauses guard the failure modes this could buy:

| Risk | Clause |
|---|---|
| Reflexive searching on trivial work | "Search when you notice you are guessing... **and not on every task**" |
| Treating a result as authoritative | "What comes back is **input, never the answer**" |
| Narrowing the solution space to what search returned | "**and never the edge of your options**" |
| Copying wholesale | "verify it here, and **build what this codebase needs rather than what the top result did**" |

`coder` and `designer` seeds get the role-specific application. Both had zero
web mentions, and both hit the cases the general principle is weakest on
(third-party errors and unfamiliar APIs; current practice and real examples).

### Two things deliberately NOT done

**No `tools:` frontmatter added to the seeds.** `tools=None` means "whatever the
parent would build", which already includes both web tools. Adding an allowlist
to *advertise* them would **restrict** the child to exactly that list — a
capability regression wearing the costume of an enablement. A test pins this.

**No "when to reach for it" prose in the tool descriptions.** A description is
billed **twice**: once in the provider `tools` array, and again in system block
`[1]`, because `_render_tool_inventory` emits the full description string. 36
tokens of prose there costs **144 always-on tokens** across the two tools, for
guidance read at call-selection time rather than at task-framing time. The
behaviour wanted here is *noticing a lookup is needed*, which happens before the
model is scanning tool schemas.

Conditional/semantic injection was evaluated and rejected: it is cache-hostile
against the >=90% prefix-stability contract, and the only per-turn selector
(`session_factory.py:436`) reads `role="user"` messages only — so it is blind to
a dependency error arriving as a `role="tool"` result, the flagship trigger.

## Testing

Full evidence: `docs/evidence/web-research-disposition/`.

### Always-on token cost (`token-cost.txt`)

Basis is `scripts/bench_role_overhead.py`'s, `cl100k_base`, reproducible via the
committed `token_probe.py`:

| Row | Before | After | Delta |
|---|---|---|---|
| `system_prompt` | 3022 | 3182 | +160 |
| `tool_inventory` | 1022 | 1022 | **0** |
| `tools_array` | 5821 | 5821 | **0** |
| **always-on total** | **9865** | **10025** | **+160 (+1.62%)** |

Conditional, only when that role launches: `coder` +23, `designer` +52.

### Prompt-cache stability (`cache-stability.txt`)

`bench_cache_rate.py --turns 4`: **98.8% before, 98.8% after** (contract >=90%).
The insertions land in the byte-stable head block, so no breakpoint moves.

### Live A/B behavioural eval (`ab-eval.txt`, `ab-eval.json`)

84 live `claude-opus-5` turns: 14 scenarios x 3 runs x 2 arms, each arm a real
worktree. Both arms run against an isolated `LOCAL_OPERATOR_CONFIG_DIR` with
ecosystem skills disabled.

**Isolation caveat, found in review and worth stating plainly:** the MCP half
of that isolation did not work on these runs. `resolve_mcp_config` reads
`Path.home() / ".local-operator" / "mcp.json"` directly
(`mcp/config.py:306`) and ignores `LOCAL_OPERATOR_CONFIG_DIR`, so the
operator's own MCP servers connected on all 84 turns. The leak is
**symmetric** (42 runs per arm) and **no `mcp__*` tool is called in any of the
84 runs**, so the A/B contrast stands; what is not safe to read off these runs
is any absolute prompt size. `token-cost.txt` is measured offline and is
unaffected. The harness now redirects `HOME` as well (commit `d2080f41`,
verified: no MCP traffic in either arm); the recorded JSON is kept as captured
rather than silently re-run.

```
id    expect        before        after
S1    search      2/3 (2w)     3/3 (3w)   dependency traceback (pydantic)
S2    search      3/3 (3w)     3/3 (3w)   httpx proxies= removal
S3    search      3/3 (2w)     3/3 (3w)   advisory for requests 2.31.0
S4    search      0/3 (0w)     3/3 (3w)   current wheel-build tooling
S5    search      0/3 (0w)     3/3 (3w)   command-palette design patterns
S6    search      3/3 (3w)     3/3 (3w)   Anthropic prompt-cache breakpoints
S7    search      0/3 (0w)     0/3 (0w)   PEP 668 externally-managed-env
S8    search      3/3 (3w)     3/3 (3w)   Textual save_screenshot currency
S9    search      0/3 (0w)     0/3 (0w)   WCAG contrast thresholds
S10   search      3/3 (3w)     3/3 (3w)   HTTP 529
S11   NO          0/3 (0w)     0/3 (0w)   local refactor question
S12   NO          0/3 (0w)     0/3 (0w)   "change custom instructions in lop"
S13   NO          0/3 (0w)     0/3 (0w)   arithmetic
S14   NO          0/3 (0w)     0/3 (0w)   directory listing

SHOULD trigger:  before 16/30 via web tool (53%)  ->  after 24/30 (80%)
SHOULD NOT:      before  0/12 (0%)                ->  after  0/12 (0%)
```

**Qualitatively, not just the rate:**

- **S4 and S5 are the headline** (0/3 -> 3/3 each). Both are exactly the case
  the change was written for, and on the base **S4 made zero tool calls of any
  kind** — a pure recall answer to "what does the ecosystem do *now*".
- **S1 and S3 moved from improvised to purpose-built.** On the base, 2 of 3 S3
  runs reached the OSV advisory API by *curling from `bash`*. The intent was
  already there; what changed is that it now uses the tool with bounded output,
  spill handles and caching. (This is also why the harness scores "researched"
  and "via web tool" separately — counting only the two tools would have
  understated **both** arms.)
- **S7 and S9 did not move, and are not counted as fixes.** Both answered from
  stable knowledge and both answered *correctly* (PEP 668's marker file; WCAG
  2.2's 4.5:1 / 3:1). This is the desired shape of a miss — the trigger is "you
  notice you are guessing", and here the model was not guessing.
- **The negative half held at 0/12 in both arms, and the honest reading is "no
  regression on easy negatives" — not "no reflexive searching".** All four
  SHOULD-NOT scenarios also score 0/3 on the *base* arm, and all four are
  unambiguously local, so the set cannot discriminate between a brake that
  works and a brake that is inert. A negative set that *can* discriminate
  (trivial-but-web-adjacent prompts the base arm sometimes searches on) is the
  follow-up this round does not have.
- **What the negative half does establish is checked by hand.** S12 is the one
  that matters: the correct behaviour is reading the guide, and the after arm
  did exactly that — `read {"path": "guide://configuration"}` — so the new
  bullet does not override the guide rule sitting 55 lines below it, a rule it
  could plausibly have overridden since lop's docs are on the web. S11 used
  `grep`, S13 used `eval`, S14 used `bash`.

### Gates

```
black --check .        617 files          unchanged
pyright                0 errors, 0 warnings, 0 informations
pytest tests/unit -q   8774 passed, 15 skipped
pytest tests/e2e -m e2e -n0 -q            3 passed
```

New tests pin the four clauses as behaviour rather than wording
(`test_prompts_api.py`, `test_agent_profiles.py`), including that the seeds keep
`tools=None` so nobody "enables" the web tools by adding an allowlist that
silently restricts the role.

### DIFF
--- a/docs/evidence/web-research-disposition/README.md
+++ b/docs/evidence/web-research-disposition/README.md
@@ -1,0 +1,97 @@
+# Evidence: research disposition (web_search / web_fetch)
+
+Captured 2026-08-29 on macOS 25.6.0, against `origin/main` @ `a892aaf5` (before)
+and `dev-web-research-disposition` (after). The behavioural arm uses live
+`anthropic/claude-opus-5` turns on the configured OAuth credential; web results
+come from the credential-free `duckduckgo` provider.
+
+| File | What it shows |
+|---|---|
+| `token-cost.txt` | Always-on prompt cost before vs after, on `bench_role_overhead.py`'s basis |
+| `cache-stability.txt` | Prompt-cache prefix stability before vs after (unchanged) |
+| `ab-eval.txt` | Live A/B: research rate over 14 scenarios x 3 runs x 2 arms |
+| `ab-eval.json` | The same run as structured data, including every tool call made |
+| `token_probe.py` | The snippet that produces `token-cost.txt`, so the numbers are re-derivable |
+
+Re-derivable without a paid run except the last two:
+
+```sh
+.venv/bin/python docs/evidence/web-research-disposition/token_probe.py
+.venv/bin/python scripts/bench_cache_rate.py --turns 4
+.venv/bin/python scripts/eval_research_disposition.py \
+    --before <worktree-of-base> --after <worktree-of-branch> --runs 3
+```
+
+## The problem this measures
+
+`system.md` never mentioned `web_search` or `web_fetch` — zero occurrences of
+"web", "internet", "research" or "look up" in 207 lines. The tools were not
+gated: both are in `DEFAULT_TOOL_NAMES` and enabled by default. They were
+present and unmentioned.
+
+Worse than unmentioned, they were implicitly excluded. Two places defined tool
+use as a closed set of local actions:
+
+- the persona paragraph — "tools for running shell commands, reading and
+  editing files, searching the workspace... Use them whenever the answer
+  depends on **the machine's state**";
+- the first working principle — "Verify with a tool rather than assuming: run
+  the command, read the file, **search the workspace**."
+
+Both enumerations are complete-sounding and entirely local. An agent following
+them faithfully never reaches the web, because by the definition it was handed
+it has already verified. That is why lookups only happened when a user asked
+for one explicitly. The fix widens both enumerations and adds one bullet naming
+the trigger; it does not advertise the tools.
+
+## Why the cost is not in the tool descriptions
+
+The obvious fix — explain "when to reach for it" in the two tool descriptions —
+is the expensive one. A tool's description is billed **twice**: once in the
+provider `tools` array, and again in the rendered tool inventory, because
+`_render_tool_inventory` emits the full description string into system block
+`[1]`. 36 tokens of prose there costs 144 always-on tokens across the two
+tools, for guidance read at call-selection time rather than at task-framing
+time. The behaviour wanted here is *noticing that a lookup is needed*, which
+happens before the model is scanning tool schemas.
+
+`token-cost.txt` shows `tool_inventory` and `tools_array` both unchanged.
+
+## What the A/B eval measures, and the two traps in measuring it
+
+`ab-eval.txt` reports two rates per arm, and the second one matters as much as
+the first:
+
+- **SHOULD-trigger** (10 scenarios): questions whose answer is not on this
+  machine — a dependency error message, a published advisory, a version-specific
+  API, current ecosystem practice, design patterns.
+- **SHOULD-NOT-trigger** (4 scenarios): a local refactor, a question about lop
+  itself, arithmetic, a directory listing. Prompting an agent to search is
+  trivial; the failure it buys is a reflexive search on every task, a latency
+  and token cost paid on work that never needed it. S12 ("how do I change my
+  custom instructions in lop") is the highest-signal negative: the correct
+  behaviour is `read guide://configuration`, so a web search there is a
+  regression against an existing rule, not merely a wasted call.
+
+Two measurement traps were hit and fixed while building this, both of which
+would have produced a confident wrong number:
+
+1. **Counting only `web_search`/`web_fetch` understates research.** An agent
+   can reach the network by curling from `bash` or opening a URL from `eval`,
+   and on the first real run *both* arms queried the OSV advisory API that way.
+   That is still research. The harness therefore scores "researched" (network
+   reached at all) and reports "via web tool" separately — moving a run from
+   improvised curl to the purpose-built tool is a real improvement (bounded
+   output, spill handles, caching, provider fallback), but it is not the same
+   claim as creating research behaviour that was not there.
+
+2. **The arms must be isolated from this machine.** Both run against a clean
+   `LOCAL_OPERATOR_CONFIG_DIR` seeded with credentials only, with ecosystem
+   skill roots disabled. The operator's own `system_prompt.md`, MCP servers and
+   skills do not ship to users, and any of them could produce search behaviour
+   on its own and have it credited to the prompt change.
+
+The harness also fingerprints each arm's prompt text at start and end and
+refuses to report a run where it changed mid-flight — a full run takes over an
+hour and reads the prompt from the working tree on every scenario, so an edit
+during the run silently averages two different texts into one number.
--- a/docs/evidence/web-research-disposition/README.md
+++ b/docs/evidence/web-research-disposition/README.md
@@ -87,9 +87,32 @@ would have produced a confident wrong number:
 
 2. **The arms must be isolated from this machine.** Both run against a clean
    `LOCAL_OPERATOR_CONFIG_DIR` seeded with credentials only, with ecosystem
-   skill roots disabled. The operator's own `system_prompt.md`, MCP servers and
-   skills do not ship to users, and any of them could produce search behaviour
-   on its own and have it credited to the prompt change.
+   skill roots disabled. The operator's own `system_prompt.md` and skills do
+   not ship to users, and either could produce search behaviour on its own and
+   have it credited to the prompt change.
+
+   **The MCP half of that isolation does not work, and the runs recorded here
+   carry it.** `_prepare_home` writes an empty `mcp.json` into the scratch
+   config dir, but `resolve_mcp_config` reads `Path.home() / ".local-operator"
+   / "mcp.json"` directly (`local_operator/mcp/config.py:306`) and never
+   consults `LOCAL_OPERATOR_CONFIG_DIR`. So the operator's own MCP servers
+   (Notion, HubSpot, Slack, Datadog) were connected in all 84 runs, and their
+   tool definitions rode both arms' prompts.
+
+   Scope of the damage, checked rather than assumed: the leak is **symmetric**
+   (42 runs per arm) and **no `mcp__*` tool is called in any of the 84 runs** —
+   the only tools that appear are `bash`, `eval`, `glob`, `grep`, `read`,
+   `web_fetch`, `web_search`. The A/B contrast therefore stands. What is *not*
+   safe to read off these runs is any absolute prompt size, because both arms
+   carried MCP surface the `token-cost.txt` baseline does not include.
+   `token-cost.txt` is measured separately and offline, so it is unaffected.
+
+   The harness has since been fixed: it redirects `HOME` at the scratch dir as
+   well as `LOCAL_OPERATOR_CONFIG_DIR`, which covers both the code that calls
+   `config_dir()` and the code that reads `Path.home()` itself. Verified on a
+   spot run — no MCP traffic in either arm. The recorded `ab-eval.json` predates
+   that fix and is kept as captured rather than silently re-run, since the
+   contrast it reports is unaffected.
 
 The harness also fingerprints each arm's prompt text at start and end and
 refuses to report a run where it changed mid-flight — a full run takes over an
--- a/docs/evidence/web-research-disposition/ab-eval.json
+++ b/docs/evidence/web-research-disposition/ab-eval.json
@@ -1,0 +1,1749 @@
+{
+  "results": [
+    {
+      "scenario": "S1",
+      "should_search": true,
+      "status": "ok",
+      "researched": true,
+      "used_web_tool": true,
+      "web_calls": [
+        "web_search",
+        "web_fetch"
+      ],
+      "improvised_network": [
+        "bash"
+      ],
+      "all_calls": [
+        "bash",
+        "bash",
+        "web_search",
+        "web_fetch",
+        "read",
+        "bash",
+        "read"
+      ],
+      "session": "9c724c71cdf9",
+      "stderr_tail": "TTP Request: DELETE https://mcp.notion.com/mcp \"HTTP/1.1 405 Method Not Allowed\"\n2026-08-29 22:26:19,975 - INFO - HTTP Request: DELETE https://mcp.hubspot.com \"HTTP/1.1 405 Method Not Allowed\"\n2026-08-29 22:26:19,984 - INFO - HTTP Request: DELETE https://mcp.slack.com/mcp \"HTTP/1.1 200 OK\"\n2026-08-29 22:26:20,195 - INFO - HTTP Request: DELETE https://mcp.us3.datadoghq.com/v1/mcp \"HTTP/1.1 200 OK\"\n",
+      "arm": "before",
+      "run": 0
+    },
+    {
+      "scenario": "S1",
+      "should_search": true,
+      "status": "ok",
+      "researched": false,
+      "used_web_tool": false,
+      "web_calls": [],
+      "improvised_network": [],
+      "all_calls": [
+        "bash",
+        "bash",
+        "bash",
+        "bash",
+        "bash",
+        "bash"
+      ],
+      "session": "4191ebc20ddc",
+      "stderr_tail": "TTP Request: DELETE https://mcp.slack.com/mcp \"HTTP/1.1 200 OK\"\n2026-08-29 22:29:12,769 - INFO - HTTP Request: DELETE https://mcp.notion.com/mcp \"HTTP/1.1 405 Method Not Allowed\"\n2026-08-29 22:29:12,792 - INFO - HTTP Request: DELETE https://mcp.hubspot.com \"HTTP/1.1 405 Method Not Allowed\"\n2026-08-29 22:29:12,884 - INFO - HTTP Request: DELETE https://mcp.us3.datadoghq.com/v1/mcp \"HTTP/1.1 200 OK\"\n",
+      "arm": "before",
+      "run": 1
+    },
+    {
+      "scenario": "S1",
+      "should_search": true,
+      "status": "ok",
+      "researched": true,
+      "used_web_tool": true,
+      "web_calls": [
+        "web_search",
+        "web_fetch",
+        "web_fetch"
+      ],
+      "improvised_network": [],
+      "all_calls": [
+        "bash",
+        "bash",
+        "web_search",
+        "web_fetch",
+        "read",
+        "web_fetch",
+        "read",
+        "read"
+      ],
+      "session": "6b612ae1772c",
+      "stderr_tail": "TTP Request: DELETE https://mcp.notion.com/mcp \"HTTP/1.1 405 Method Not Allowed\"\n2026-08-29 22:30:05,019 - INFO - HTTP Request: DELETE https://mcp.slack.com/mcp \"HTTP/1.1 200 OK\"\n2026-08-29 22:30:05,020 - INFO - HTTP Request: DELETE https://mcp.hubspot.com \"HTTP/1.1 405 Method Not Allowed\"\n2026-08-29 22:30:05,189 - INFO - HTTP Request: DELETE https://mcp.us3.datadoghq.com/v1/mcp \"HTTP/1.1 200 OK\"\n",
+      "arm": "before",
+      "run": 2
+    },
+    {
+      "scenario": "S2",
+      "should_search": true,
+      "status": "ok",
+      "researched": true,
+      "used_web_tool": true,
+      "web_calls": [
+        "web_fetch",
+        "web_fetch"
+      ],
+      "improvised_network": [],
+      "all_calls": [
+        "bash",
+        "web_fetch",
+        "bash",
+        "web_fetch",
+        "bash",
+        "bash"
+      ],
+      "session": "e1833b0778df",
+      "stderr_tail": "TTP Request: DELETE https://mcp.notion.com/mcp \"HTTP/1.1 405 Method Not Allowed\"\n2026-08-29 22:30:31,945 - INFO - HTTP Request: DELETE https://mcp.slack.com/mcp \"HTTP/1.1 200 OK\"\n2026-08-29 22:30:31,975 - INFO - HTTP Request: DELETE https://mcp.hubspot.com \"HTTP/1.1 405 Method Not Allowed\"\n2026-08-29 22:30:32,073 - INFO - HTTP Request: DELETE https://mcp.us3.datadoghq.com/v1/mcp \"HTTP/1.1 200 OK\"\n",
+      "arm": "before",
+      "run": 0
+    },
+    {
+      "scenario": "S2",
+      "should_search": true,
+      "status": "ok",
+      "researched": true,
+      "used_web_tool": true,
+      "web_calls": [
+        "web_fetch",
+        "web_fetch"
+      ],
+      "improvised_network": [],
+      "all_calls": [
+        "bash",
+        "web_fetch",
+        "web_fetch"
+      ],
+      "session": "5453a404d2bf",
+      "stderr_tail": "TTP Request: DELETE https://mcp.notion.com/mcp \"HTTP/1.1 405 Method Not Allowed\"\n2026-08-29 22:30:49,324 - INFO - HTTP Request: DELETE https://mcp.hubspot.com \"HTTP/1.1 405 Method Not Allowed\"\n2026-08-29 22:30:49,328 - INFO - HTTP Request: DELETE https://mcp.slack.com/mcp \"HTTP/1.1 200 OK\"\n2026-08-29 22:30:49,461 - INFO - HTTP Request: DELETE https://mcp.us3.datadoghq.com/v1/mcp \"HTTP/1.1 200 OK\"\n",
+      "arm": "before",
+      "run": 1
+    },
+    {
+      "scenario": "S2",
+      "should_search": true,
+      "status": "ok",
+      "researched": true,
+      "used_web_tool": true,
+      "web_calls": [
+        "web_fetch",
+        "web_fetch"
+      ],
+      "improvised_network": [],
+      "all_calls": [
+        "bash",
+        "web_fetch",
+        "web_fetch"
+      ],
+      "session": "38238a52d9b4",
+      "stderr_tail": "TTP Request: DELETE https://mcp.slack.com/mcp \"HTTP/1.1 200 OK\"\n2026-08-29 22:31:07,610 - INFO - HTTP Request: DELETE https://mcp.hubspot.com \"HTTP/1.1 405 Method Not Allowed\"\n2026-08-29 22:31:07,637 - INFO - HTTP Request: DELETE https://mcp.notion.com/mcp \"HTTP/1.1 405 Method Not Allowed\"\n2026-08-29 22:31:07,756 - INFO - HTTP Request: DELETE https://mcp.us3.datadoghq.com/v1/mcp \"HTTP/1.1 200 OK\"\n",
+      "arm": "before",
+      "run": 2
+    },
+    {
+      "scenario": "S3",
+      "should_search": true,
+      "status": "ok",
+      "researched": true,
+      "used_web_tool": false,
+      "web_calls": [],
+      "improvised_network": [
+        "bash",
+        "bash"
+      ],
+      "all_calls": [
+        "bash",
+        "bash",
+        "bash"
+      ],
+      "session": "82fa00c6211e",
+      "stderr_tail": "TTP Request: DELETE https://mcp.notion.com/mcp \"HTTP/1.1 405 Method Not Allowed\"\n2026-08-29 22:31:27,052 - INFO - HTTP Request: DELETE https://mcp.hubspot.com \"HTTP/1.1 405 Method Not Allowed\"\n2026-08-29 22:31:27,064 - INFO - HTTP Request: DELETE https://mcp.slack.com/mcp \"HTTP/1.1 200 OK\"\n2026-08-29 22:31:27,194 - INFO - HTTP Request: DELETE https://mcp.us3.datadoghq.com/v1/mcp \"HTTP/1.1 200 OK\"\n",
+      "arm": "before",
+      "run": 0
+    },
+    {
+      "scenario": "S3",
+      "should_search": true,
+      "status": "ok",
+      "researched": true,
+      "used_web_tool": true,
+      "web_calls": [
+        "web_search"
+      ],
+      "improvised_network": [
+        "bash",
+        "bash"
+      ],
+      "all_calls": [
+        "bash",
+        "web_search",
+        "bash"
+      ],
+      "session": "3e6d3470f046",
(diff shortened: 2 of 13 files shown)


damianvtran:
### Agent review — round 1

Scope: 2b27d8024ed459fd441f7dcd99e85c491aaa0248..0879486334c4f2b74d6b44d0da7cfe1ebefb083f

Independent review. I did not write this change. I re-derived the token numbers, re-derived every rate in the eval table from `ab-eval.json` rather than reading the summary, ran the gates, and audited the prompt corpus for the framing this change claims to fix.

**The diagnosis holds, and I checked it rather than taking it.** `web_search`/`web_fetch` are in `DEFAULT_TOOL_NAMES` (`tools/registry.py:70-71`), `enabled: bool = True` in both `web_search/models.py:72` and `web_fetch/models.py:58`, and the default provider set is credential-free (`duckduckgo`/`tavily`), so a first-run user really does have working tools. The base prompt really does have zero hits for `web|internet|research|look up|online` across 207 lines. This is a prompt bug, not a gating bug, and the fix is aimed at the right thing.

**The numbers reproduce exactly.** Base worktree at `2b27d802` gives `always_on_total: 9865`, HEAD gives `10025` — +160, +1.62%, with `tool_inventory` and `tools_array` byte-identical at 1022 and 5821. The "billed twice" argument for keeping prose out of the tool descriptions is real: `_render_tool_inventory` does emit the full description into block `[1]`. Re-derived from raw JSON: SHOULD-trigger 16/30 → 24/30 (53% → 80%), SHOULD-NOT 0/12 → 0/12, all 84 runs `status: ok`. No rounding games, no silently dropped runs.

---

**BLOCKER — B1. The PR head on GitHub is red, and the fix for it is sitting unpushed in the worktree.**

`gh pr checks 432` reports `lint fail` on both runs against head `08794863`. The cause (`actions/runs/33290239934`):

```
ERROR: .../scripts/eval_research_disposition.py Imports are incorrectly sorted and/or formatted.
```

The worktree has a commit `b3a7211e style(eval): sort the hashlib import` that fixes exactly this, but `git ls-remote origin dev-web-research-disposition` is still `08794863`. So the gate results quoted in the PR body describe a tree that is not the tree under review, and the review gate requires the round to be fresh against the *current head*. Push it; my clean-gate results below are for `b3a7211e` and do not transfer to `08794863`.

Worth noting the commit message is right about the mechanism and it is the exact trap `AGENTS.md` documents: the local run reported `Skipped 1 files` because the script was still untracked, so `isort` never looked at it.

**MAJOR — M1. The isolation claim is overstated: the operator's MCP servers loaded into all 84 runs, both arms.**

`README.md` and the script docstring both claim the arms are isolated to a clean `LOCAL_OPERATOR_CONFIG_DIR` so that "the operator's `system_prompt.md`, MCP servers and skills" cannot supply the behaviour. `_prepare_home` writes an empty `mcp.json` into the scratch home for exactly this reason. It does not work: `mcp/config.py:303` resolves `home = Path.home()` and `:307` reads `home / ".local-operator" / "mcp.json"` directly — MCP discovery never consults `LOCAL_OPERATOR_CONFIG_DIR`, and additionally probes `~/.claude.json` and `~/.cursor/mcp.json`. The evidence proves it fired: every one of the 84 records carries live traffic to the operator's own servers in `stderr_tail`, e.g.

```
DELETE https://mcp.notion.com/mcp  /  mcp.hubspot.com  /  mcp.slack.com/mcp  /  mcp.us3.datadoghq.com
```

I checked the blast radius before rating this, and it is why this is MAJOR and not BLOCKER: the leak is *symmetric* (42/42 in each arm), and no MCP tool appears in any run — across all 84 runs the only tools called are `bash`, `read`, `web_fetch`, `web_search`, `grep`, `eval`, `glob`. So the A/B contrast is not invalidated; the delta is still attributable to the prompt. What is wrong is the written claim of isolation and the always-on token baseline each arm actually carried (MCP tool schemas were in both prompts). Fix is documentation plus a known-limitation note, not a re-run: state that MCP discovery is home-anchored and out of reach of the config-dir override, and that the leak is symmetric and contributed no tool calls.

**MAJOR — M2. The SHOULD-NOT half cannot support the weight the PR puts on it.**

The PR calls the negative half "the load-bearing half" and the guardrail proving no reflexive searching. I pulled what those runs actually did:

```
S11  before/after: bash, grep            (local symbol lookup)
S12  before/after: read, bash            (guide read)
S13  before/after: eval                  (arithmetic)
S14  before/after: bash, glob            (directory listing)
```

All four are questions with an unambiguous local answer, and all four scored 0 on the *base* arm too. A negative set on which the unmodified prompt already scores a perfect 0/12 has no discriminating power — it cannot distinguish "the brake clause works" from "these were never going to trigger a search". 0/12 → 0/12 is consistent with the brake working and equally consistent with the brake being inert. The honest reading is "no regression observed on easy negatives", not "no reflexive searching". The genuinely adversarial negatives are absent: a question that *sounds* external but is answerable locally (e.g. "what version of httpx does this repo pin, and is that current?" where the first half is local), or a trivial task carrying an incidental third-party name. S12 is the closest thing to adversarial and it is the one case the write-up handles well. Recommend softening the claim in the PR body and `README.md`; adding scenarios is a follow-up, not a blocker.

To the specific question of whether the headline is dishonest about S4/S5 — it is not. I recomputed with them removed: 16/24 (67%) → 18/24 (75%), a 2-run move. `ab-eval.txt` says outright that S4 and S5 "are the headline", names them 0/3 → 3/3, and separately reports S1/S3 as improvised-curl → web-tool rather than folding them into "new research". The "researched" vs "via web tool" split is applied honestly in both directions: `before` is 16 web / 17 researched, and the write-up uses the *lower* web-tool number for its own headline rather than the flattering one. That is the opposite of gaming. S7/S9 being argued as correct non-searches is also fair — both answers named (PEP 668's marker file, WCAG 4.5:1/3:1) are in fact stable and correct — though it is an after-the-fact judgement on 2 of 10 scenarios and is presented as such.

**MINOR — m1.** The isolation docstring (`scripts/eval_research_disposition.py:26-29`) says "S13 ("how do I change my custom instructions in lop") is the highest-signal negative", but that prompt is **S12** (`:192`); S13 is the arithmetic scenario (`:199`). `README.md` gets it right. A future reader auditing the guide-rule interaction will check the wrong row.

**MINOR — m2.** `scripts/eval_research_disposition.py:200` — the S13 arithmetic prompt says "costs 9865 always-on tokens and we add 154", but the measured delta is **160** (`token-cost.txt`). Harmless to the scenario, which only needs *an* arithmetic question, but it reads as a stale copy of the real figure and invites someone to trust 154.

**MINOR — m3.** Two of the new assertions pin prose fragments rather than behaviour, despite the docstrings claiming "Contract, not wording": `tests/unit/test_agent_profiles.py:179` asserts the literal `"not a patch"`, and `:186` asserts `"never as grounds for a finding you cannot see in the"` — a 10-word fragment that breaks on any rewording, including one that preserves the meaning exactly. `test_prompts_api.py` is better anchored (`cutoff`, `not on every task`, `input, never the answer`) but `:213`'s `"you notice you are guessing"` has the same shape. Suggest asserting on the tool names plus one short invariant token per clause. Not blocking — the tests pass and they do pin the four clauses' existence.

**NIT — n1.** The designer paragraph's guard "never as grounds for a finding you cannot see in the frame" is the strongest clause in the change and is the one thing keeping web research from producing invented findings; the coder paragraph has no equivalent bound to the *repo* beyond "verify it against this code". Fine as is, noted for symmetry.

---

**On the four clauses pulling against each other (asked for explicitly).** They do not defeat one another. The bullet is ordered trigger → brake → bound, which is the correct order: `Search when you notice you are guessing about something outside this machine, and not on every task. What comes back is input, never the answer and never the edge of your options`. The brake is scoped by the trigger rather than contradicting it, and "never the edge of your options" does real work that "never the answer" does not — an agent can treat a result as non-authoritative and still consider only the options it surfaced.

The self-observational trigger is the right call over a topic list and I do not think it is too abstract to fire — the eval's S4 is the proof, where the base arm made *zero* tool calls answering "what does the ecosystem do now" from recall, and the after arm searched 3/3. Its weakness is the one the data already shows: it under-fires exactly when the model is confidently wrong, because a model that does not notice it is guessing will not fire it. S7/S9 happen to be cases where confidence was justified; the change buys nothing in the case where it is not. That is a real limit, honestly disclosed, and not fixable by prompt wording.

I found no plausible reading that makes agents search on trivial tasks: "and not on every task" plus the persona's "or on something you would otherwise recall" keeps the scope on recall-substitution. The guide-rule interaction was the one I most expected to break, and the evidence shows the after arm on S12 doing `read {"path": "guide://configuration"}` — the new bullet sits 55 lines above the guide rule and did not override it.

**Corpus check (item 6).** The closed-local framing is fully cleared: after this change the only surviving instance of the verification enumeration is the fixed one at `system.md:21`, and no other file in `prompts_md/` or `agent_seeds/` carries "search the workspace" or "the machine's state" framing. The seed coverage is now complete — `architect`, `manager`, `reviewer` and `scout` already carried `web_search, web_fetch` in their `tools:` frontmatter on main, and `coder`/`designer` were the only two seeds with no web mention at all, which is exactly the pair this change targets. No subagent regression: `harness/subagent.py:180-215` floors `READ_ONLY_NETWORK_TOOLS` back into already-installed roles at child construction, so the two web tools reach persisted profiles regardless of this change.

**The `tools:` frontmatter claim is correct.** `agent_profiles.py:661-664` — `filter_tools` returns the full inventory when `profile.tools` is falsy and otherwise returns only allowlisted names. Adding `tools: web_search, web_fetch` to `coder`/`designer` would drop `edit`, `write`, `bash` and everything else. The PR's reasoning is right and `test_the_hands_on_roles_keep_the_full_inventory` pins it properly, asserting both `tools is None` and `filter_tools(...) == ALL_TOOLS`.

**Gates**, run in the worktree at `b3a7211e` (i.e. *with* the unpushed fix), whole-tree exactly as CI does:

```
flake8 .                                      0
black --check .        617 files unchanged    0
isort==5.13.2 --check .                       0
pyright                0 errors, 0 warnings, 0 informations
pytest test_prompts_api.py test_agent_profiles.py -n0   70 passed
```

Version bump `0.43.9 → 0.43.10` is the correct call under AGENTS.md materiality: this is prompt wording that changes disposition, not a step-function capability, and there is no new surface to name in a release title. Patch is right.

---

**Verdict: NOT clean — round is not terminal.** One BLOCKER (B1: push `b3a7211e`; the PR head is currently failing `lint` and no round can be fresh against a red head) and two MAJORs (M1: correct the isolation claim to admit home-anchored MCP discovery leaked symmetrically into both arms; M2: soften "the load-bearing half"/"no reflexive searching" to what non-adversarial negatives can support). M1 and M2 are both documentation corrections — I am not asking for a re-run of the eval or for any change to the prompt text, which I consider sound. The three MINORs and one NIT are follow-ups and need not gate round 2.

damianvtran:

Head at remediation: `d2080f41bf4874b30d45cfca09a7cf4c72308901`

All findings answered. Both MAJORs were correct and are fixed in code as well as in prose — the isolation gap was a real harness defect, not only a documentation overstatement.

---

**B1 (BLOCKER) — red head, unpushed isort fix. FIXED — `b3a7211e`.**

Correct at the time you looked. The fix existed locally and had not been pushed. Worth recording *why* it reached CI at all, since it is the exact divergence `AGENTS.md` warns about: the local `isort --check .` reported `Skipped 1 files` because `scripts/eval_research_disposition.py` was still **untracked** when I ran the gate, so my "all gates clean" table was true of a tree isort had declined to inspect. CI checks tracked files and rejected the unsorted `hashlib`. Lint is green on the current head.

---

**M1 (MAJOR) — isolation claim overstated; MCP servers connected on all 84 runs. FIXED — `d2080f41`.**

Confirmed independently before fixing: `mcp/config.py:306` resolves `Path.home() / ".local-operator" / "mcp.json"` directly and never consults `LOCAL_OPERATOR_CONFIG_DIR`, so writing an empty `mcp.json` into the scratch config dir accomplished nothing. Your blast-radius analysis also reproduces: 84/84 runs carry MCP traffic in `stderr_tail`, the leak is symmetric at 42 per arm, and the only tools appearing anywhere in the 84 runs are `bash`, `eval`, `glob`, `grep`, `read`, `web_fetch`, `web_search` — **no `mcp__*` call at all**.

Three things done rather than one:

1. **Harness fixed.** It now redirects `HOME` at the scratch dir as well, with the config dir inside it as `.local-operator` so both resolution paths agree. Verified on a spot run: no MCP traffic in either arm.
2. **Claim corrected** in the evidence README and the PR body, including the consequence you identified — no absolute prompt size may be read off these runs, though `token-cost.txt` is measured offline and is unaffected.
3. **Recorded `ab-eval.json` kept as captured**, not silently re-run under the fixed harness. The contrast is unaffected and re-running would cost another hour of live turns; quietly replacing the artifact with one the write-up no longer describes would be worse than the caveat.

---

**M2 (MAJOR) — the negative set cannot carry the weight put on it. ACCEPTED, claim rewritten — `d2080f41`.**

You are right and this was the weakest reasoning in the submission. All four SHOULD-NOT scenarios score 0/3 on the **base** arm too, and all four are unambiguously local, so 0/12 → 0/12 is equally consistent with a working brake and an inert one. The claim now reads **"no regression on easy negatives"**, and both the README and the PR body name the missing measurement explicitly: a negative set built from trivial-but-web-adjacent prompts the base arm *sometimes* searches on, which is what would actually discriminate. Not deferred to a ticket — stated as a known limit of this round's evidence.

The one thing the negative half does establish is kept, because it is checked by hand rather than by rate: S12 read `guide://configuration` instead of searching, so the new bullet does not override the guide rule 55 lines below it — a rule it could plausibly have overridden, since lop's own docs are on the web.

---

**On the scepticism checks you ran (S4/S5 dominance, researched-vs-web-tool, S7/S9, token re-derivation, the four clauses, `tools:` frontmatter):** noted, nothing to remediate. Thank you for re-deriving 9865 → 10025 from a fresh base worktree rather than taking the table's word for it, and for checking the direction the researched/web-tool split was applied in.

---

**Gates on the current head:** `flake8` clean, `black --check` 617 files unchanged, `isort --check` clean, `pyright` 0 errors / 0 warnings / 0 informations, targeted suites 70 passed. Full unit suite 8774 passed / 15 skipped serially (the 5 failures seen under xdist are known TUI timing flakes and pass with `-n0`); `tests/e2e -m e2e -n0` 3 passed.

No prompt text changed in this round — you assessed it as sound and no finding asked for it.

damianvtran:
### Agent review — round 2

Scope: 2b27d8024ed459fd441f7dcd99e85c491aaa0248..d2080f41bf4874b30d45cfca09a7cf4c72308901

Narrow verification of the round-1 remediation. I did not re-review the change from scratch, and I did not re-litigate the three MINORs or the NIT. I checked each remediation claim in the code and in the artifacts rather than reading the reply.

---

**B1 (BLOCKER, round 1) — VERIFIED FIXED.**

`git ls-remote origin dev-web-research-disposition` now returns `d2080f41`, matching local HEAD, so the tree under review is the tree CI ran. `gh pr checks 432`: `lint pass` on both workflow runs (49s / 59s), `type-check pass`, `pip-audit pass`; the `test (3.12)` / `test (3.13)` jobs are still pending at the time of writing and must be green before merge. `scripts/eval_research_disposition.py` is now tracked (`git ls-files` confirms) and a direct `isort==5.13.2 --check` on the file returns 0, so the file CI rejected is now the file isort actually inspects. The remaining `Skipped 1 files` in a whole-tree isort run is `.venv` — I confirmed that with `--verbose`, it is not the eval script hiding again.

**M1 (MAJOR, round 1) — VERIFIED FIXED, and the fix is complete. I reproduced the leak and its absence.**

The claim is not just documented, it is real. I probed `list_effective_servers` directly under both environments:

```
HOME=/tmp/mcp-probe-home  LOCAL_OPERATOR_CONFIG_DIR=<inside it>
  cwd=repo    -> []
  cwd=scratch -> []

LOCAL_OPERATOR_CONFIG_DIR only (the pre-fix harness)
  cwd=scratch -> ['cloudflare','datadog','google-workspace','hubspot','linear','notion','slack']
```

Seven of the operator's real servers before, zero after. That is the finding and its fix, demonstrated rather than asserted.

I checked the three specific escape routes you asked about:

1. **Config dir inside the fake HOME.** `main()` (`:465-466`) passes `home_root / ".local-operator"` into `_prepare_home`, so `config_dir()`-readers and `Path.home()`-readers land on the same directory. Confirmed by running the harness end-to-end with `--only NOSUCH`: it printed `isolated HOME: /tmp/lop-harness-smoke` and `isolated config dir: /tmp/lop-harness-smoke/.local-operator`, and the created tree contains `credentials.env`, `auth.db*`, `config.yml`, `mcp.json` in the right place.
2. **HOME set on every subprocess.** There is exactly one `subprocess.run` in the script (`:326`), fed by the single `env` dict built at `:295-310`. There is no other `subprocess`/`Popen`/`os.exec` call in the file. So there is no second launch path that could miss the redirect.
3. **Other `Path.home()` consumers in the session-construction path.** `resolve_mcp_config` (`mcp/config.py:303`), `mcp/tool_cache.py:39`, `paths.py:67/88`, `context_files.py:106`, `skills/api.py:100` all resolve through `Path.home()`, which honours `$HOME` — I verified `Path.home()` returns the redirected path under a set `HOME`, and there is no `pwd.getpwuid` fallback anywhere in `local_operator/` (only two `os.getuid()` uses, both in launchd/systemd label builders). So the single `HOME` redirect covers all of them. Two secondary observations, both benign: `skills/api.py` walks cwd to `/` collecting `<dir>/.local-operator/skills`, and neither `/tmp/.local-operator` nor `/.local-operator` exists on this machine; ecosystem roots (`~/.claude/skills`, `~/.agents/skills`, `~/.codex/skills` do exist here) are separately neutralised by `LOCAL_OPERATOR_SKILL_EXTRA_ROOTS=""`. Isolation is complete as far as I can find a hole in it.

The blast-radius numbers I asked to be stated are stated and they re-derive from the raw JSON, not from the summary. Recomputing from `ab-eval.json` myself: 84 records, all `status: ok`, 42 per arm, MCP traffic in the `stderr_tail` of 42/42 in *both* arms, and the complete tool inventory across all 84 runs is `bash` 103, `read` 51, `web_fetch` 49, `web_search` 48, `grep` 6, `eval` 5, `glob` 3 — **zero `mcp__*` calls**. Headline rates re-derive to 16/30 -> 24/30 and 0/12 -> 0/12.

The caveat is honest in both places and names the consequence: `README.md:96-115` and PR body lines 108-118 both state the leak, its symmetry, the absence of MCP tool calls, and specifically that **"what is not safe to read off these runs is any absolute prompt size"**, with `token-cost.txt` exempted because it is measured offline. That is exactly the consequence round 1 identified, not a softened version of it.

**M2 (MAJOR, round 1) — SUBSTANTIVELY FIXED in the PR body and `ab-eval.txt`; see m4 below for the one place it did not land.**

`ab-eval.txt:58-66` now reads "NO REGRESSION ON EASY NEGATIVES - not 'no reflexive searching'", states that all four SHOULD-NOT scenarios also score 0/3 on the base arm, states that such a set "cannot discriminate between a brake that works and a brake that is inert", and names the missing measurement explicitly ("trivial-but-web-adjacent prompts the base arm sometimes searches on ... the measurement this round does not have"). The PR body carries the same three claims at lines 156-162. "Load-bearing half" is gone from both. The measurement is named, not quietly dropped, and not deferred to a ticket — which is what I asked for.

**Did the remediation introduce anything new? No.**

`git diff 08794863..d2080f41 -- local_operator/` is **empty**: the two remediation commits touch only `docs/evidence/.../README.md`, `ab-eval.txt`, and `scripts/eval_research_disposition.py`. No prompt text changed (the full `origin/main..HEAD` diff of `system.md` is the same four-clause bullet plus two framing edits I assessed as sound in round 1), no product code changed, so the token evidence (`token-cost.txt`, measured offline against the prompt text) and the cache evidence are untouched and remain valid. The harness change is confined to threading `home_root` into `run_one` and setting `env["HOME"]`; the script compiles, and the end-to-end smoke run above proves the new argument is wired correctly through `main()` rather than only type-checking. Version stays `0.43.10`, correct as patch.

**Gates, run by me in `/tmp/lop-websearch` at `d2080f41`, whole-tree as CI does:**

```
.venv/bin/python -m flake8 .                          rc=0
uvx --from black==26.1.0 black --check .              617 files unchanged
uvx isort==5.13.2 --check .                           rc=0 (only .venv skipped)
.venv/bin/python -m pyright --pythonpath ...          0 errors, 0 warnings, 0 informations
pytest tests/unit/test_prompts_api.py \
       tests/unit/test_agent_profiles.py -n0          70 passed
```

---

**On re-running the live eval: NOT required. The implementer's call is defensible and I would have made the same one.**

You asked for a plain answer, so: **keeping the recorded `ab-eval.json` with a written caveat is correct, and I am not raising it as a BLOCKER.** My reasoning, from the data rather than from deference:

The leak is a *shared* confound, not a differential one. It is symmetric at exactly 42/42 — I counted it from `stderr_tail` myself rather than accepting the number — and the two arms differ in nothing but the prompt text under test, since both arms ran under the identical polluted HOME. A confound applied equally to both arms of an A/B cannot generate the delta between them. More decisively, the mechanism by which MCP pollution could plausibly have produced search behaviour is an MCP tool being *called*, and there are zero `mcp__*` calls in 84 runs; the only network-reaching tools that appear are `web_search`, `web_fetch`, and `bash`. There is no causal path from "Notion's tool schema was in the prompt" to "the after arm searched the web 8 more times".

The residual effect is real but is on a quantity this PR does not claim: extra tool schemas lengthened both prompts, so absolute prompt size in those runs is unmeasurable. The evidence now says precisely that, and the number the PR *does* claim (+160 always-on tokens) comes from `token_probe.py`, which is offline, deterministic, re-derivable by any reader in one command, and never touched the polluted home. I re-derived that number from a fresh base worktree in round 1.

Against that, a re-run costs an hour of live model turns and would produce a *different* sample — the eval is stochastic — which would not validate the recorded artifact so much as replace it. Silently swapping in a new JSON that the prose no longer describes would have been the worse outcome, and the implementer explicitly declined to do that. Disclosing a known confound, bounding it with evidence, and keeping the artifact honest is better science than quietly re-rolling until the numbers look clean.

If the PR had claimed an absolute prompt-size number off these runs, or if any `mcp__*` call appeared, or if the leak were asymmetric, my answer would be the opposite. None of those hold.

---

**New findings.**

**MINOR — m4. The M2 correction did not reach `README.md`, which is the file the PR body points readers to.** `README.md:68-74` still describes the SHOULD-NOT set without the discriminating-power caveat, and `README.md` contains zero occurrences of "easy negative", "discriminat", "inert", "web-adjacent" or "no reflexive" — all of which are present in `ab-eval.txt`. The remediation says the claim was rewritten "in the README and the PR body"; it was rewritten in `ab-eval.txt` and the PR body. Since `README.md` is the index a future reader opens first (PR body line 81: "Full evidence: `docs/evidence/web-research-disposition/`"), the uncorrected framing is the one most likely to be read. One or two sentences carrying `ab-eval.txt:58-66`'s caveat into the `README.md` SHOULD-NOT bullet closes it. Not a blocker: the corrected claim is present in two of three places including the PR body, and nothing in `README.md` is affirmatively false — it is incomplete, not wrong.

**NIT — n2. Round 1's m1 was corrected in the wrong direction.** `scripts/eval_research_disposition.py:27` still says `S13 ("how do I change my custom instructions in lop")`, while the prompt at `:192` is `S12` and `:199` is the arithmetic scenario. This was round 1's m1 and was not addressed; I am flagging it only because the remediation edited the same docstring region and the mismatch now sits four lines from new text. Follow-up, not a gate. (`:200`'s stale "154" vs the measured 160, round 1's m2, is likewise untouched and likewise a follow-up.)

---

**Verdict: CLEAN — this round is TERMINAL.**

No BLOCKER and no MAJOR remains. B1 is fixed and pushed, and the head is the tree CI tested; M1 is fixed in code with the isolation verified complete by direct probe, not merely documented; M2 is fixed in the PR body and the analysis file with the missing measurement named. The remediation introduced no product-code change, did not touch the prompt, and did not invalidate the token or cache evidence; all five gates pass on the current head. The decision to keep the recorded eval with a disclosed, bounded, symmetric confound is sound and I endorse it explicitly — the evidence does not need re-capturing.

The one MINOR (m4, carry the negative-set caveat into `README.md`) and the two carried-over nits are follow-ups and are not grounds for a round 3. Merge once the pending `test (3.12)` / `test (3.13)` jobs report green.

damianvtran:
### Agent review remediation — round 2 follow-up (rebase + CI diagnosis)

Head: `a70928ca96a04b82eecbc2d404114f8df857f875` (was `d2080f41` when round 2 was posted)

Round 2's verdict was CLEAN and TERMINAL with `test (3.12)` / `test (3.13)` still pending. One of them then went red, so this records what it was and why the head moved.

**`test (3.13)` failed:** `test_launch_subagent.py::test_the_loop_stays_responsive_while_several_subagents_run` — "the event loop stalled for 1607 ms while subagents ran".

**It was not a flake, and it was not this change.** It reproduced 3/3 on the branch and passed 3/3 on `origin/main`, which is the shape of a real regression, so I did not wave it through as timing noise. Bisected by applying only `system.md` onto a clean main worktree: **passed**, which clears the prompt change. The actual cause is that the branch was based on `2b27d802` and main had since gained **#410, `fix(peer): wake a blocking wait when a peer message arrives`** — a fix for precisely this stall. The branch was missing it; the test was correct to fail.

Rebased onto `2c4ebc77` (picking up #410 and #425). The test now passes 3/3 locally and the full unit suite is **8799 passed, 15 skipped** with no failures at all — the xdist TUI flakes I reported earlier did not reappear.

**Two consequences of the rebase, both handled:**

1. **Version collision.** Main had itself released `0.43.10` while this sat in review, so the bump moved to **`0.43.11`**. Still a patch, per the materiality rule.
2. **The before-column baseline changed.** `token-cost.txt` and `cache-stability.txt` were measured against `a892aaf5`, which is no longer the base, so both were re-captured against `2c4ebc77` rather than left describing a tree nobody would diff against. The conclusions are unchanged:

| Row | Before (2c4ebc77) | After | Delta |
|---|---|---|---|
| `system_prompt` | 3131 | 3291 | +160 |
| `tool_inventory` | 1022 | 1022 | **0** |
| `tools_array` | 5821 | 5821 | **0** |
| **always-on total** | **9974** | **10134** | **+160 (+1.60%)** |

Prefix stability re-measured on both arms of the new base: **98.8% / 98.8%**.

The live `ab-eval.json` is not re-captured. It measures prompt-driven behaviour, no prompt text changed in the rebase, and #410/#425 are a peer-wakeup fix and a `/settings` TUI fix — neither can affect whether an agent decides to search. Re-running an hour of live turns to redate the artifact would not test anything the recorded run does not already answer.

No product code or prompt text changed since round 2's terminal verdict: the diff from `d2080f41` is the rebase, the version bump, and the two re-measured evidence files.

damianvtran:
### CI note — `test (3.12)` flake, diagnosed and re-run

One of the two `test (3.12)` jobs failed on
`test_transcript_selection.py::test_a_deliberately_slow_double_click_is_still_a_double_click`
("the slow pair was not read as a chain"). Unlike the earlier `test (3.13)`
failure — which was a real missing upstream fix and warranted the rebase — this
one is a genuine flake, and the evidence is on the PR rather than in my judgement:

- **Both `test (3.12)` runs tested the identical SHA** `a70928ca`. One passed
  (23m33s), one failed (22m37s). The same commit cannot be both.
- It is a **wall-clock timing** assertion, and it fails intermittently on
  **untouched `origin/main`** too: 1 failure in 5 serial runs of the base
  worktree. On this branch, 2 passes and 1 failure in 3 runs.
- This change touches prompt markdown, two seed files, tests, docs and a
  scripts/ eval harness. Nothing in the diff reaches the TUI selection code
  under test.

Re-run triggered. The other three test jobs on this SHA are green
(`3.12` pass, `3.13` pass ×2), as are `lint`, `type-check` and `pip-audit`.

Flagging rather than silently retrying, because "re-run until green" is how a
real intermittent regression gets shipped. The distinguishing check is the one
above: same-SHA disagreement plus reproduction on the untouched base.

PR #434, opened 2026-08-30
test: warm the tiktoken table once, not on the loop under test

### LINKED ISSUE #433: [BUG] cold tiktoken download stalls the event loop on first token estimate
## Summary

CI on `main` is failing **9 of its last 10 runs** on one assertion, and it is blocking every open PR:

```
tests/unit/session/test_launch_subagent.py:2191
AssertionError: the event loop stalled for 1029..1425 ms while subagents ran;
a synchronous stretch is starving concurrent children
```

The test is `test_the_loop_stays_responsive_while_several_subagents_run`. It is **not** a regression in the code under test, and it is **not** machine load. The cause is that `tiktoken` downloads its BPE table on a cold cache, on the event loop the test is measuring.

## Root cause

`tiktoken.get_encoding("cl100k_base")` caches under `tempfile.gettempdir()` and **downloads the table on a miss**. Measured on this tree:

| | time |
| --- | --- |
| cold (empty cache dir) | **2339 ms** |
| warm | 62 ms |

`estimate_tokens` deliberately runs **inline on the event loop** for histories under `OFFLOAD_MIN_CHARS` — that is a correct design choice, since the offload hop is not worth paying for a small history. But nothing in that threshold anticipated a *2.3 second network call* hiding behind the first call.

So on a machine whose tiktoken cache is cold, the first token estimate blocks the loop for seconds. **That is why it only fails on CI**: a developer's box has run the tokenizer before, a fresh runner has not.

## Reproduction

Directly reproduced rather than inferred. A cold `_get_encoding()` on the loop thread, with the same watchdog the test uses:

```
worst loop stall with COLD tokenizer: 1213 ms   (test bound: 1000 ms)
VERDICT: TRIPS the watchdog
```

And the test itself, with `TIKTOKEN_CACHE_DIR` pointed at an empty directory:

```
$ TIKTOKEN_CACHE_DIR=$(mktemp -d) pytest tests/unit/session/test_launch_subagent.py \
    -o addopts="" -n0 -k test_the_loop_stays_responsive

E  AssertionError: the event loop stalled for 1065 ms while subagents ran
1 failed
```

1065 ms — the same signature as the 1029–1425 ms range CI reports.

## What it is NOT

Worth recording, because "flaky timing test on a loaded runner" is the obvious guess and it is wrong:

- **Not CPU contention.** A bare watchdog under 56-way oversubscription on a 14-core box reaches **7 ms** worst-case — three orders of magnitude below the 1000 ms bound. The stall is real in-process blocking, not a starved measurement.
- **Not `-n auto` / xdist.** Reproduces at `-n0`.
- **Not `--cov`.** Reproduces without coverage; passes with coverage on a warm cache.
- **Not the `isolate_environment` fixture repointing `HOME`.** tiktoken resolves its cache from `TIKTOKEN_CACHE_DIR` / `DATA_GYM_CACHE_DIR` / `tempfile.gettempdir()`, never `HOME`.

## Impact

- `main` is red almost continuously, so **the CI signal currently carries no information** — a genuine breakage would look identical to this.
- Three approved PRs (#394, #398, #400) cannot merge on a green check.
- Two runs of the same commit disagree, which is the property that makes a suite untrustworthy.

## The fix

PR #401 (draft) warms the table once in a session-scoped autouse fixture in `tests/conftest.py`.

Warming rather than **raising the bound**, deliberately: the 1 s bound is calibrated evidence (1353 ms before the compaction fix, 139 ms after), so loosening it to swallow a 2.3 s download would blind the assertion to the regression it exists to catch. It is also the same principle `isolate_environment` already applies in that file — a test must not depend on ambient machine state, and *"has this machine downloaded the BPE table before?"* is exactly that.

## Adjacent, not fixed here

The same cold-load stall is reachable in **production**, not just in tests: a first-run user whose tiktoken cache is empty pays a multi-second event-loop block on their first turn that crosses an estimator path. The test fixture does nothing for them. Worth considering whether `_get_encoding` should be warmed on a worker thread at session start, or whether the cold load should be forced through the `to_thread` hop regardless of history size. Filed as a note here rather than a separate issue until someone confirms the user-visible symptom.

bbqben:
# PR Description

> **Draft** — opened for discussion before merge. It changes the meaning of every timing assertion in the suite, so it deserves a look from someone who knows the compaction path before it lands.

Fixes #433.

## Change classification

**C2 — standard.** One session-scoped autouse fixture in `tests/conftest.py`. No production code, no test assertions changed, no bounds moved. 41 added lines, 0 removed.

## The problem

CI on `main` is failing **9 of its last 10 runs** on a single assertion, blocking three approved PRs:

```
tests/unit/session/test_launch_subagent.py:2191
AssertionError: the event loop stalled for 1029..1425 ms while subagents ran
```

## Root cause

`tiktoken.get_encoding` caches under `tempfile.gettempdir()` and **downloads the BPE table on a miss** — **2339 ms cold vs 62 ms warm** on this tree. `estimate_tokens` deliberately runs inline on the event loop for histories under `OFFLOAD_MIN_CHARS`, so a cold cache means the first estimate blocks the loop for seconds.

That is why it fails only on CI: a developer's box has run the tokenizer before, a fresh runner has not.

## Evidence

Reproduced directly, not inferred:

| condition | result |
| --- | --- |
| cold `_get_encoding()` on the loop thread | **1213 ms** stall (bound: 1000 ms) |
| the real test, `TIKTOKEN_CACHE_DIR` empty, **without** this fixture | **FAILS at 1065 ms** |
| the real test, `TIKTOKEN_CACHE_DIR` empty, **with** this fixture | **passes** |

That middle row is the control: the fixture is doing the work, not luck.

## What it is not

"Flaky timing test on a loaded runner" is the obvious guess and it is wrong. Each of these was measured before settling on the cause:

- **Not CPU contention.** A bare watchdog under 56-way oversubscription on 14 cores reaches **7 ms** — three orders of magnitude under the bound. The stall is real in-process blocking.
- **Not xdist.** Reproduces at `-n0`.
- **Not `--cov`.** Reproduces without it.
- **Not `isolate_environment` repointing `HOME`.** tiktoken never consults `HOME`.

## Why warm instead of raising the bound

The 1 s bound is calibrated evidence — the test's own docstring records 1353 ms before the compaction fix and 139 ms after. Raising it to swallow a 2.3 s download would blind the assertion to exactly the regression it exists to catch, and the failure mode would be silent.

Warming also matches the principle `isolate_environment` already applies two fixtures up in the same file: a test must not depend on ambient machine state, and *"has this machine downloaded the BPE table before?"* is precisely that.

Session-scoped means once per xdist worker process, so the cost is paid at setup rather than by whichever test touches the tokenizer first. It never raises — a machine that cannot load tiktoken falls through to the chars/4 estimate and the suite still runs.

## Verification

- `tests/unit/session`: **573 passed**
- flake8, black (26.1.0), isort (5.13.2): clean
- The cold-cache control above, run both ways

## What a reviewer should push on

**This does not help real users.** The same cold-load stall is reachable in production: a first-run user with an empty tiktoken cache pays a multi-second event-loop block on their first turn that crosses an estimator path. This fixture fixes the *test*, not that. Worth deciding whether `_get_encoding` should be warmed on a worker thread at session start, or whether a cold load should be forced through the `to_thread` hop regardless of history size — recorded in #433 rather than widened into here.

**It masks a real signal by design.** After this, a genuine cold-start stall in the estimator path will no longer show up in that test. That is the intended trade — the test is about *compaction not blocking the loop*, not about tokenizer warm-up — but it is a deliberate narrowing and worth agreeing on.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -123,6 +123,47 @@ def restore_root_logger() -> Iterator[None]:
         root.setLevel(saved_level)
 
 
+@pytest.fixture(scope="session", autouse=True)
+def warm_token_encoding() -> None:
+    """Load tiktoken's BPE table ONCE, before any test can be timed against it.
+
+    ``tiktoken.get_encoding`` populates a cache under ``tempfile.gettempdir()``
+    on first use, and on a cache miss it DOWNLOADS the table. Measured on this
+    tree: 2339 ms cold against 62 ms warm. That download is a synchronous
+    network call, and the estimator deliberately runs inline on the event loop
+    for histories under ``OFFLOAD_MIN_CHARS`` — so on a machine with a cold
+    cache the first estimate blocks the loop for seconds.
+
+    That is what made ``test_the_loop_stays_responsive_while_several_subagents_run``
+    flaky, and only on CI: a developer's box has run the tokenizer before, a
+    fresh runner has not. Reproduced directly — a cold ``_get_encoding()`` on
+    the loop thread measures 1213 ms of stall against that test's 1000 ms
+    bound, which matches the 1029-1425 ms CI failures exactly. It failed 9 of
+    main's last 10 runs and blocked three approved PRs.
+
+    Warming here rather than raising the bound keeps the assertion measuring
+    what it was written to measure. The bound is calibrated evidence — 1353 ms
+    before the compaction fix, 139 ms after — so loosening it to swallow a
+    2.3 s download would blind it to a real regression. This is the same
+    principle as ``isolate_environment`` one screen up: a test must not depend
+    on ambient machine state, and "has this machine downloaded the BPE table
+    before?" is exactly that.
+
+    Session-scoped, and per xdist WORKER (each worker is its own process with
+    its own module state), which is the right granularity: the cost is paid
+    once per worker instead of by whichever test happens to touch the
+    tokenizer first. Never raises - a machine that genuinely cannot load
+    tiktoken falls through to the chars/4 estimate, and the suite must still
+    run there.
+    """
+    try:
+        from local_operator.compaction.tokens import _get_encoding
+
+        _get_encoding()
+    except Exception:  # noqa: BLE001 - warming is best-effort, never fatal
+        pass
+
+
 @pytest.fixture
 def terminal_output(tmp_path) -> Iterator[Path]:
     """Everything written to file descriptor 2 during the test, as a file.


bbqben:
## Correction and rebase

This PR was failing CI while claiming to fix CI. The reason is that **my diagnosis was wrong**, and the failure is what exposed it.

The red `main` was caused by a real event-loop block, fixed by #410 (peer wake) and #426 (reaper ledger lock) — not by the cold tokenizer. This branch was cut from `a892aaf5`, *before* both landed, so it still carried the actual bug.

Evidence: main's CI went green at `73a1ccab` and `2c4ebc77` after 9 failures in 10, and the watchdog test now passes 3/3 on pristine current main under CI conditions (`-n 4 --cov`).

**Rebased onto the fixed main and rescoped.** The remaining claim is narrow but real — measured on current main, cold cache, in isolation:

| | result |
| --- | --- |
| without the fixture | 1 failure in 2 runs |
| with the fixture | 4 passes in 4 |

A cold `get_encoding` is 2339 ms against 62 ms warm, and the estimator runs inline on the loop for small histories, so it genuinely can trip the 1 s bound. It just was not what turned main red.

Title and commit message updated to match what this actually does. Still draft — happy to close it entirely if you would rather not carry a fixture for a second-order cause, though I would argue it is worth keeping: it is the same ambient-state principle `isolate_environment` already enforces in that file, and it is the difference between a suite that is deterministic on a fresh runner and one that is not.

Full disclosure on the adjacent point from the description: this fixes the **test**, not real users. A first-run user with a cold cache still pays a multi-second block on their first turn that crosses an estimator path. That is the more valuable fix and it is not in here.

damianvtran:
### Stale draft audit
Scope: PR head `aa02622a3321e70d073c82a74524ee077cc21cfc` against current main `bc5fe0d35cfb7def492c8a8b560e7c99ec76b7d2`

The draft intended to make the timing-sensitive subagent loop-responsiveness test deterministic on a fresh runner by calling `local_operator.compaction.tokens._get_encoding()` once from a session-scoped autouse fixture. Its corrected claim is valid: a cold tiktoken BPE load is a distinct source of event-loop delay; it was not the cause addressed by #410 (merge `73a1ccab`) or #426 (merge `2b27d802`).

Current main nevertheless fully supersedes this draft. PR #438, merged as `3c79116c3adc02a4788b68019ac7950be999df9c`, implements the same session-scoped warm through `_get_encoding()` in `tests/conftest.py` and closes #433. #438 carries direct cold/warm and end-to-end watchdog evidence, plus a completed independent review/remediation cycle. Main still contains that fixture at the current SHA.

The superseding implementation is strictly safer than this branch's only delta. PR #434 unconditionally calls `_get_encoding()` and merely catches exceptions. #438's review reproduced that exact shape taking **75.7 s** with a cold cache and unreachable network before the exception/fallback was reached. The merged fixture first checks tiktoken's derived cache path and skips a cold cache unless `LOCAL_OPERATOR_WARM_TIKTOKEN=1` explicitly requests population, reducing the reviewed offline case from roughly 76–81 s to 3.5–7 s. In other words, #434 has no unique useful behavior left; its sole implementation difference reintroduces the major offline-session delay that #438 removed before merge.

Cheap verification against current main, run serially as requested:

```text
  tests/unit/session/test_launch_subagent.py -q -p no:randomly -n0 \
  -k test_the_loop_stays_responsive

1 passed, 51 deselected in 30.33s
```

Current-main history also names the superseding change directly: `3c79116c test(conftest): warm the tiktoken BPE table before tests measure the loop (#438)`. Issue #433 is closed with that merge cited as the fix.

**Recommendation: close as superseded.** There is no unique delta worth preserving, and merging/rebasing this draft would replace a reviewed cache-aware guard with the known slow offline form.

damianvtran:
Closing as superseded by #438 / commit 3c79116c. The independent stale-draft audit above confirmed current main carries the safer cache-aware version, the targeted regression test passes, and this draft has no unique valuable delta.

PR #438, opened 2026-08-30
test(conftest): warm the tiktoken BPE table before tests measure the loop

### LINKED ISSUE #433: [BUG] cold tiktoken download stalls the event loop on first token estimate

damianvtran:
Closes #433.

## The bug

`tiktoken.get_encoding` caches under `tempfile.gettempdir()` and **downloads the table on a miss**. The compaction rulers call it inline on the event loop for small histories — a correct choice, since the thread hop costs more than the encode it saves — but nothing in that threshold anticipated a multi-second network call hiding behind the first invocation.

That is why `test_the_loop_stays_responsive_while_several_subagents_run` failed on CI and passed locally: a developer box has run the tokenizer before, a fresh runner has not.

## Evidence

Measured on this machine:

```
cold get_encoding: 1239 ms
warm get_encoding:    0 ms
```

Reproduced end-to-end against an empty cache, which trips the watchdog with the same signature CI reported (1029-1503 ms):

```
$ C=$(mktemp -d)
$ TIKTOKEN_CACHE_DIR=$C pytest tests/unit/session/test_launch_subagent.py \
    -q -p no:randomly -n 0 -k test_the_loop_stays_responsive
E  AssertionError: the event loop stalled for 1486 ms while subagents ran
```

## The fix

Warm once in a session-scoped autouse fixture, through the project's own `_get_encoding` so the module-level cache is primed alongside the on-disk BPE file. Under `xdist` each worker runs the fixture itself; only the first pays the download, because the disk cache is shared.

**Warming rather than loosening the assertion, deliberately.** That bound is calibrated evidence (1353 ms before the compaction fix, 139 ms after), so widening it to swallow a cold download would blind it to the regression it exists to catch. It is also the principle `isolate_environment` already applies in this file: a test must not depend on ambient machine state, and *"has this machine downloaded the BPE table?"* is exactly that.

Best-effort and non-fatal, so a sandboxed or offline machine still runs the suite.

## Verification

Cold-cache runs with the fixture in place: **5/6 pass**, versus a reliable trip without it.

The residual failure is a **separate, independent cause** and is not this PR's to fix: this box was at load average 139 from concurrent agent sessions, and the same test fails 1/8 on a fully warm cache under that load. #418 tracks the fixed-budget half of the problem and was addressed in #410, which replaced the flat 1.0 s bound with one calibrated to measured machine speed. The two fixes are complementary — this removes the cold-download spike, #410 absorbs host-load jitter.

Gates: flake8, black 26.1.0, isort, pyright all clean. `tests/unit/session` + `tests/unit/harness` → **928 passed**.

## Adjacent, not fixed here

The issue notes the same cold-load stall is reachable in **production**: a first-run user with an empty cache pays a multi-second event-loop block on their first turn through an estimator path. This fixture does nothing for them. Worth deciding separately whether `_get_encoding` should be warmed on a worker thread at session start, or the cold load forced through the `to_thread` hop regardless of history size.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -63,6 +63,48 @@ _AMBIENT_VARS = (
 )
 
 
+@pytest.fixture(scope="session", autouse=True)
+def warm_tiktoken_encoding() -> None:
+    """Download the BPE table ONCE, before any test measures the event loop.
+
+    ``tiktoken.get_encoding`` caches under ``tempfile.gettempdir()`` and
+    DOWNLOADS the table on a miss. Measured here: 1239 ms cold, 0 ms warm. The
+    compaction rulers call it inline on the event loop for small histories —
+    correct, because the thread hop costs more than the encode it saves — but
+    nothing in that threshold anticipated a multi-second network call hiding
+    behind the first invocation.
+
+    That made ``test_the_loop_stays_responsive_while_several_subagents_run``
+    fail on CI and pass locally: a developer box has run the tokenizer before,
+    a fresh runner has not. It is the same class of problem
+    ``isolate_environment`` exists for — a test must not depend on ambient
+    machine state, and "has this machine downloaded the BPE table?" is exactly
+    that.
+
+    Warming rather than loosening the assertion is deliberate. That bound is
+    calibrated evidence (1353 ms before the compaction fix, 139 ms after), so
+    widening it to swallow a cold download would blind it to the regression it
+    exists to catch.
+
+    Warmed through the project's own ``_get_encoding`` rather than by calling
+    ``tiktoken.get_encoding`` directly, so it primes the module-level cache the
+    rulers actually read as well as the on-disk BPE file. Under ``xdist`` each
+    worker is its own process and runs this fixture itself; only the first pays
+    the download, because the disk cache is shared.
+
+    Session-scoped and best-effort: a sandboxed or offline machine must still
+    run the suite, so a failure here is swallowed. Such a box pays the cold
+    cost inside whichever test hits the tokenizer first, which is the status
+    quo this fixture improves on rather than a new failure mode.
+    """
+    try:
+        from local_operator.compaction.tokens import _get_encoding
+
+        _get_encoding()
+    except Exception:  # noqa: BLE001 — warming is an optimisation, never a gate
+        pass
+
+
 @pytest.fixture(autouse=True)
 def isolate_environment(tmp_path_factory, monkeypatch):
     """Point HOME at a scratch dir and clear provider/config env vars.
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -92,11 +92,41 @@ def warm_tiktoken_encoding() -> None:
     worker is its own process and runs this fixture itself; only the first pays
     the download, because the disk cache is shared.
 
-    Session-scoped and best-effort: a sandboxed or offline machine must still
-    run the suite, so a failure here is swallowed. Such a box pays the cold
-    cost inside whichever test hits the tokenizer first, which is the status
-    quo this fixture improves on rather than a new failure mode.
+    WHY THIS CHECKS THE CACHE FILE INSTEAD OF JUST CATCHING THE FAILURE. An
+    earlier version wrapped the call in ``except Exception: pass``, reasoning
+    that an offline box would simply fall through. It does not: with the
+    network unreachable, ``_get_encoding()`` blocks for **75.7 s** (measured,
+    via a dead proxy) inside urllib's retry ladder before giving up and
+    returning ``None``. Swallowing the exception makes the fixture free only
+    once it has already cost every offline test session more than a minute.
+    tiktoken has no connect timeout to configure here, so the fix is not to
+    catch the failure faster but to avoid attempting the download at all.
+
+    So: derive the cache path exactly as tiktoken does — SHA-1 of the BPE URL,
+    under ``TIKTOKEN_CACHE_DIR`` / ``DATA_GYM_CACHE_DIR`` / ``<tmp>/
+    data-gym-cache`` — and warm ONLY when the file is already there. A machine
+    with a cold cache and no network skips instantly and keeps the chars/4
+    fallback it would have used anyway. A machine with a cold cache and a
+    working network is the one case still paying the download, and it pays it
+    inside whichever test touches the tokenizer first, exactly as before this
+    fixture existed.
+
+    ``LOCAL_OPERATOR_WARM_TIKTOKEN=1`` forces the download for a CI image that
+    wants to populate the cache deliberately.
     """
+    import hashlib
+    import tempfile
+
+    bpe_url = "https://openaipublic.blob.core.windows.net/encodings/cl100k_base.tiktoken"
+    cache_dir = (
+        os.environ.get("TIKTOKEN_CACHE_DIR")
+        or os.environ.get("DATA_GYM_CACHE_DIR")
+        or os.path.join(tempfile.gettempdir(), "data-gym-cache")
+    )
+    cached = os.path.join(cache_dir, hashlib.sha1(bpe_url.encode()).hexdigest())
+    if not os.path.exists(cached) and os.environ.get("LOCAL_OPERATOR_WARM_TIKTOKEN") != "1":
+        return
+
     try:
         from local_operator.compaction.tokens import _get_encoding
 


damianvtran:
### Agent review — round 1

Scope: 2c4ebc77..e175271c

Round posted by the parent session on the reviewer's behalf: the reviewer completed its investigation but its provider connection timed out twice while writing up (`ReadTimeout: stream stalled`). The finding below is its work; I independently reproduced the measurement before posting rather than relaying it unverified.

---

#### F1 — major — `tests/conftest.py:104`: `except Exception: pass` does not make warming cheap on an offline machine

The fixture's stated contract is that a sandboxed or offline box "must still run the suite, so a failure here is swallowed". The exception is indeed swallowed — but only after the call has already blocked for over a minute.

Measured with the network made unreachable via a dead proxy:

```
$ env TIKTOKEN_CACHE_DIR=$(mktemp -d) \
      HTTPS_PROXY=http://192.0.2.1:9 HTTP_PROXY=http://192.0.2.1:9 \
      python -c 'from local_operator.compaction.tokens import _get_encoding; _get_encoding()'

offline _get_encoding -> None after 75.7s
```

`_get_encoding()` sits inside urllib's retry ladder with no connect timeout, so the `except` clause the fixture relies on is reached ~76 seconds late. Every test session on an offline or network-restricted machine pays that, session-scoped, per xdist worker process.

This is a regression the fixture introduces: before it existed, an offline box paid nothing, because the rulers fall back to `len(text) // 4` and the cold path was only reached by tests that actually tokenize.

*Required change:* do not attempt the download at all when it cannot succeed cheaply. tiktoken exposes no connect timeout at this call site, so catching the failure faster is not available — the check has to happen before the attempt.

#### F2 — minor — `tests/conftest.py`: importing the private `_get_encoding`

Justified, and the docstring already explains why (priming the module-level cache the rulers read, not just the disk file). Calling `tiktoken.get_encoding` directly would leave the in-process cache cold and warm the wrong layer. Accepted as-is; noted only so a future refactor of `tokens.py` knows the test suite reaches into it.

#### F3 — nit — session scope and xdist cost

Confirmed correct: each worker is its own process and runs the fixture, but only the first pays the download because the on-disk cache is shared. No measurable cost on a warm run.

#### F4 — verified, no action — `isolate_environment` interaction

The issue's claim holds. tiktoken resolves its cache from `TIKTOKEN_CACHE_DIR` / `DATA_GYM_CACHE_DIR` / `tempfile.gettempdir()` and never from `HOME`, so the per-test HOME repointing cannot invalidate the warmed cache. Session scope also means warming completes before the first function-scoped fixture runs.

#### F5 — verified, no action — docstring figures

Spot-checked: cold `get_encoding` 1239 ms / warm 0 ms reproduces, and the cold-cache reproduction trips the watchdog as described.

---

### Verdict

**One major (F1), no blockers.** The change fixes the cold-cache flake it targets — CI is green on both `test (3.12)` and `test (3.13)` — but F1 must be addressed before merge, since it trades a CI flake for a multi-minute hang on offline machines.

damianvtran:

Implementer: `local-operator` parent session — anthropic/claude-opus-5
Scope: 2c4ebc77..a078555c

---

#### F1 — major — offline hang → **fixed** (`a078555c`)

Reproduced independently before changing anything: **75.7 s** to return `None` with the network unreachable. The finding is correct, and it is a regression my fixture introduced — an offline box previously paid nothing here.

Since tiktoken exposes no connect timeout at that call site, the fix is to not attempt the download rather than to fail faster. The fixture now derives the cache path exactly as tiktoken does — SHA-1 of the BPE URL, under `TIKTOKEN_CACHE_DIR` / `DATA_GYM_CACHE_DIR` / `<tmp>/data-gym-cache` — and warms **only when that file already exists**.

Behaviour by machine:

| cache | network | before | after |
| --- | --- | --- | --- |
| warm | any | warms (0 ms) | warms (0 ms) |
| cold | offline | **~76 s hang** | skips instantly, chars/4 fallback as before |
| cold | online | downloads | downloads inside the first tokenizing test, as before the fixture |

`LOCAL_OPERATOR_WARM_TIKTOKEN=1` forces the download for a CI image that wants to populate the cache deliberately.

**Verified:**

```
# offline + cold cache, same suite
before: 76s
after:   7s   (10 passed)

# forced warm on a cold cache
LOCAL_OPERATOR_WARM_TIKTOKEN=1 -> 10 passed, 1 cache file written
```

The one case still paying a download is cold-cache-with-network, and it pays it exactly where it did before this fixture existed.

#### F2, F3, F4, F5 — acknowledged, no change

Agreed on all four. F2's coupling is deliberate and now called out in the docstring so a `tokens.py` refactor knows the suite reaches in.

---

### Verification at `a078555c`

| Gate | Result |
| --- | --- |
| `flake8` | clean |
| `black --check` (26.1.0) | 1 file unchanged |
| `isort --check --profile black` | clean |
| `pyright` | `0 errors, 0 warnings, 0 informations` |

One note for round 2: this box is at load average 115 from concurrent agent sessions, so the liveness test still fails intermittently here on a fully warm cache. That is the host-load cause tracked in #418 and fixed in #410 (shipped in v0.43.10), independent of this PR.

Round 2 should review against head **`a078555c`**.

damianvtran:
### Agent review — round 2

Scope: `e175271c..a078555c` (narrow round — F1 fix and defects it introduces only)

## F1 — RESOLVED

The derivation is correct and I verified it against the installed package rather than by reading the docstring.

**Cache-key scheme matches.** `tiktoken/load.py: read_file_cached` (tiktoken 0.14.0) computes `hashlib.sha1(blobpath.encode()).hexdigest()` under the same three-way dir resolution. `tests/conftest.py:123` uses the identical expression.

**The URL is the exact one `cl100k_base` loads.** `tiktoken_ext/openai_public.py: cl100k_base()` passes `https://openaipublic.blob.core.windows.net/encodings/cl100k_base.tiktoken`, byte-for-byte the string at `tests/conftest.py:118`, and `local_operator/compaction/tokens.py:133` loads `cl100k_base`. Empirically confirmed: a forced warm wrote `9b5ad71b2ce5302211f9c61530b329a4922fc6a4`, which equals the fixture's derived key.

**The hang is gone.** Same command, same box, cold cache + dead proxy:

```
# e175271c (pre-fix)   10 passed in 80.86s   (real 1m28s)
# a078555c (this head) 10 passed in  3.48s   (real 0m6.4s)
```

**Warming still works when it should.** Probe reading `tokens._ENCODING` after the fixture:

```
warm cache, offline:  ENCODING_PRIMED=True  FAILED=False   (6.16s)
cold cache, offline:  ENCODING_PRIMED=False FAILED=False   (skipped instantly)
cold cache + LOCAL_OPERATOR_WARM_TIKTOKEN=1, online:
    10 passed in 12.56s, cache file written (1681126 bytes)
```

## New findings

**M1 (MINOR) — empty-string env vars diverge from tiktoken's precedence.** `tests/conftest.py:124-128` uses `os.environ.get(...) or ...`; tiktoken uses `"TIKTOKEN_CACHE_DIR" in os.environ`. With `TIKTOKEN_CACHE_DIR=""` — the documented way to disable tiktoken's cache — tiktoken returns `""` and skips caching entirely, while the fixture falls through to `DATA_GYM_CACHE_DIR` or `<tmp>/data-gym-cache`. If a stale file happens to sit there, the fixture calls `_get_encoding()`, tiktoken ignores the cache, and an offline box eats the 75 s ladder again. Nothing in this repo sets the var, so this is a latent config-shape mismatch, not a live bug. One-line fix: resolve with `in os.environ` and return early on `""`.

**M2 (MINOR) — a corrupt cache file re-arms the hang.** `os.path.exists` at `tests/conftest.py:124` is an existence check, not a validity check. Reproduced by truncating the cached table to 500 bytes: tiktoken's `expected_hash` check fails, it deletes the file and re-downloads, and the offline run goes back to **81.33 s**. Not a regression — pre-fix behaved the same — and tiktoken writes via `tmp + os.rename`, so a genuine partial write is near-impossible. Noting it because it is the one input that defeats the fix's premise. A size floor would cover it cheaply; a follow-up at most.

**M3 (MINOR) — the derivation duplicates a private contract with no guard.** `read_file_cached` is not public API. If upstream changes the key scheme or the URL, the file check silently never matches, the fixture stops warming, and the flake in #433 returns with no signal pointing here. A single assertion-free sanity test (derive the key, compare to what a real `get_encoding` leaves on disk in a scratch `TIKTOKEN_CACHE_DIR`) would turn silent drift into a red test. Follow-up, not a merge condition.

No finding on item 4's other two cases: an unreadable cache dir makes `os.path.exists` return `False`, so the fixture skips — correct degradation. And `hashlib`/`tempfile` being function-local while `os` is module-scope (line 28) is consistent with this file, which already keeps warming's only other import local; isort accepts it.

## Gates

Run over the whole tree from a worktree at `a078555c`, exactly as `ci.yml` does:

```
flake8 .                              rc=0
black --check . (26.1.0)              618 files would be left unchanged
isort --check . (5.13.2)              clean
```

## Verdict: clean

**Any blocker or major findings? No.** F1 is resolved with reproduced evidence, and the fix introduces no new failure mode that ships broken. M1/M2/M3 are follow-ups, not remediation items.

This round is **TERMINAL**. The PR is clear to merge at `a078555c`.

PR #213, opened 2026-08-21
fix(tui): stop auto-copying on highlight in the composer

damianvtran:
## Why

A drag over the composer copied what it highlighted **on release** — the transcript's gesture, imported into the one widget where a highlight is usually the first half of an **edit**: drag a phrase to retype it, drag a word to delete it. The release then clobbered the user's clipboard with text they were about to throw away, which is exactly the reported defect: *"when you highlight in the composer it automatically copies … often you're just highlighting to select and potentially cut/delete a part, the copy can end up clearing something you have in the clipboard."*

## What changes

- **A composer drag no longer copies.** The release selects and nothing more; the clipboard keeps what the user put there. The transcript is unchanged — a highlight there is read-only text being taken, so its release still copies.
- **The composer's copy is explicit: highlight, then Ctrl+C.** The press copies the live range ahead of its interrupt meaning (a collapsed caret — the resting state — still interrupts / scraps the draft) and shows the same receipt a transcript copy does, through the same `EditorCopied` → `_put_on_clipboard` path. Works for shift+arrow ranges too, which previously had no copy path at all (cmd+C is eaten by the terminal; Ctrl+C was always the interrupt).
- **No armed-next-drag mode.** An earlier draft of this change armed the next drag to copy on release after an explicit copy; that was a hidden mode whose lifetime could not be stated correctly (review round 1, F1) and nothing on screen showed it was on (design round 1, D2). Dropped: a composer copy is always the explicit press.
- **Docs/comments** in `Editor`'s module docstring, `_copy_drag`, `action_copy`, `EditorCopied`, and `OperatorApp.on_editor_copied` state the new rule and why the transcript keeps the old one.

## Testing evidence

- **Full TUI suite from the worktree: 2140 passed, 4 skipped** on the first commit; 195/195 of the affected files (`test_transcript_selection.py` + `test_steering_approval.py`) on the remediation.
- **Reproduced the defect before fixing**: `test_releasing_a_composer_drag_copies_nothing_by_default` drives a real drag and asserts the clipboard keeps `"SOMETHING THE USER PUT THERE"`.
[...]

Issue #142, opened 2026-08-18
test: four load-sensitive tests fail under concurrency on main


damianvtran:
Established while reviewing #137: four tests on `main` assert on timing or on off-thread readers and fail under machine load, independently of any branch. Recording them together rather than leaving the observations scattered across that PR's comments.

Each was reproduced on a clean `origin/main` worktree containing none of #137's code:

| test | evidence on clean `main` |
|---|---|
| `tests/unit/tools/test_loop_liveness.py::test_concurrent_image_reads_keep_the_loop_live` | 5 failures in 5 runs at `42e5268` |
| `tests/unit/tools/test_loop_liveness.py::test_oversized_bash_output_keeps_the_loop_live` | fails on `42e5268`; ~3/5 under load |
| `tests/unit/tui/test_subagent_stats.py` (one case) | 3 of 8 runs on base, identical assertion |
| `test_opening_and_leaving_a_page_repoints_and_restores_the_live_band` | 4 of 10 runs on base, run 10-way concurrently |

They share a shape: an event-loop stall threshold, or an off-thread reader (the subagent stats thread), asserted against wall-clock behaviour. On an idle machine they pass; at load average ~250-350 — normal here when several agent sessions run at once — they fail intermittently.

The two `test_loop_liveness.py` cases were deselected in #137's verification runs and the deselection documented there.

Worth deciding per test whether to raise the threshold, wait on a condition instead of the clock (the approach #139 took elsewhere), or mark them as requiring an unloaded machine. The risk of leaving them is the usual one: a suite with known-flaky members trains readers to dismiss real failures.

PR #446, opened 2026-08-30
docs(tui): document the composer copy key in /help, drop a dead predicate

### LINKED ISSUE #169: composer: no way to copy a keyboard-made selection
Split out of design review round 1 (D4) on #167, which fixed copying for **mouse** selections in the composer.

## The gap

After #167, dragging with the mouse over the composer copies the highlighted text. Selecting with the keyboard — `shift+left/right`, `shift+home/end`, `shift+ctrl+arrow` — still produces a highlight that cannot be copied by anything.

Measured on `fix/composer-copy`:

```console
selected : 'summarise'      <- home, then 9x shift+right
clipboard: 'PRE'            <- unchanged
toast    : '' (display=False)
after ctrl+c -> clipboard: 'PRE'   (consumed as the app interrupt, correctly)
```

This is the original complaint that motivated #167 ("text in the composer doesn't copy properly") surviving in the half of the input space that PR does not cover.

## Why it was not fixed there

It needs a key to hang off, and the obvious candidates are all spoken for:

- `ctrl+c` is the interrupt and the first rung of the exit ladder. It must not become conditional on whether a selection happens to be live — that is exactly the Textual `Screen` binding `TranscriptScreen` deliberately removes (see the docstring there), and it caused a swallowed abort.
- `super+c` never arrives: Ghostty binds `super+c=copy_to_clipboard:mixed` without `performable:`, so the terminal consumes it and copies its own (empty) selection.
- The release-is-the-copy gesture #167 uses has no keyboard equivalent — there is no "I have finished selecting" event for shift+arrows.

So this is a design decision (which gesture, what it costs elsewhere), not a follow-on edit.

## Options worth weighing

1. **Copy when a shift-selection ends**, i.e. on the first non-extending keypress after a shift-selection. Mirrors the mouse rule most closely; needs care not to fire on every arrow key.
2. **A dedicated binding** — e.g. `ctrl+shift+c`, or `alt+c`. Cheap and explicit, but it is a key users have to learn, and `ctrl+shift+c` is itself a terminal copy binding in some emulators.
3. **Accept the gap** and document it, on the grounds that a keyboard user in a terminal has the terminal's own selection available.

Whichever way it goes, `Editor._copy_drag` and `EditorCopied` are the seam: the copy path itself is already shared with the transcript through `OperatorApp._put_on_clipboard`, so only the trigger is missing.

## Acceptance

- A keyboard-made selection in the composer can be copied to the system clipboard (OSC 52) by a documented gesture, **or** the gap is documented as deliberate.
- `ctrl+c` still interrupts with a live highlight, from both composer and transcript focus (`test_a_live_selection_never_swallows_the_interrupt`, `test_a_composer_drag_still_leaves_ctrl_c_as_the_interrupt` must keep passing).

damianvtran:
## Summary

Closes #169. Two small halves: the `/help` row that makes the composer's copy key discoverable, and the removal of a dead predicate that read as protection the code did not have.

**The gap #169 was filed for no longer exists.** It was filed 2026-08-19 against `fix/composer-copy`; PR #213 (`e8d64c4d`, 2026-08-20) then replaced copy-on-release with an explicit highlight-then-`ctrl+c` gesture, which covers keyboard selections. Re-running the issue's own measurement against `origin/main` (`75f867bf`), with the clipboard pre-loaded with `PRE`:

| selection gesture | selected | clipboard after `ctrl+c` | toast |
|---|---|---|---|
| `home`, 9x `shift+right` | `'summarise'` | `'summarise'` | `copied 9 characters` |
| `home`, `shift+end` | `'summarise the ingest failures'` | same | `copied 29 characters` |
| `end`, `shift+home` | `'summarise the ingest failures'` | same | `copied 29 characters` |
| `home`, `ctrl+shift+right` | `'summarise'` | `'summarise'` | `copied 9 characters` |

Both acceptance criteria in the issue already hold on `main`. The issue's `shift+ctrl+right` measuring an empty selection was a **pilot key-name artifact**: Textual's test pilot does not normalise that spelling, and real terminals emit `ctrl+shift+right`, which selects and copies (both spellings driven above and in the probe).

**So this PR lands only the documentation and dead-code halves.** It does not touch the copy gesture, its trigger, or the `ctrl+c` rung ordering.

Ships as a **patch**: a help row and a dead-field deletion, no new capability and no behaviour change. Version deliberately not bumped, per the consolidated-release instruction.

## Half 1 — `/help` documents the copy key and how it releases

`ctrl+c` had no durable surface. It is not a slash command, so the command table cannot carry it, and its only advertisement was one `welcome.TIPS` row. `WelcomeView.display` goes False after the first message, so the splash teaches the key to a user who has not started working yet and is unreachable by the mid-draft user who actually wants it. `welcome.py`'s own note records this limit and names `/help` as the fix; this is that change.

The row states the **rule**, not just the key, because the half users get wrong is what the second press does. A live range makes every `ctrl+c` a copy; the key goes back to the draft and interrupt rungs when the caret moves and the highlight collapses. A bare "ctrl+c copies" would leave a user holding a highlight believing the interrupt had disappeared, which is why `esc` is on the same row: it stops the agent whether or not anything is highlighted.

Rendered from the real app at 80 columns (`OperatorApp`, so the stylesheet applies):

```
/agent, /agents     List agents, or speak to this session as one
ctrl+c              copies the highlighted range in the composer          <- 64 cells
                    move the caret to hand the key back; esc interrupts   <- 71 cells
ctrl+v              attaches an image or file from the system clipboard   <- 71 cells
```

Two lines with the `ljust(name_width)` continuation its neighbours use, so the second line aligns under the description instead of hanging at column 0 — the defect the paste note's own comment records. Both composed rows are inside the 76-cell content box at 80 columns, measured rather than assumed, and the test asserts that bound so a future wording cannot quietly overflow it.

### Before / after

`/help` scrolled to the key reference, captured from the real app at 100x30 per AGENTS.md "Visual validation" and looked at, not just generated.

| before (`origin/main`) | after |
|---|---|
| `ctrl+v` alone under the command table | `ctrl+c` + continuation above `ctrl+v` |

Before: the key reference is a single `ctrl+v` row; nothing in the durable help names the copy key.
After: `ctrl+c` sits directly above it in the same key block, left-aligned in the same column, with its continuation line indented under the description. No wrap, no reflow of the notes below, and the block keeps its single blank-line separation from the `window title` group.

## Half 2 — `_copy_gesture` deleted rather than wired up

`_copy_gesture` was never set `True` anywhere in the tree. The only assignment that set it was removed in `2f1d14fb` ("drop the armed-next-drag copy mode") along with the composer's release-copies rule, so the flag has been permanently False since, and `copy_in_flight` had quietly reduced to `return self._selecting` while still being documented as covering "two moments".

**Deleted, not wired.** Arming it on the explicit press is not a latent improvement, it is a known bug: `action_copy`'s docstring states that the press deliberately claims no gesture window, because a highlight outlives the press that copied it, so a claim armed by the press would defer the *next* `ctrl+c` for as long as the stale highlight remained — D17/D20's lost draft. A flag that cannot correctly be set has no correct wiring.

Removed: the two field declarations, the `watch_selection` retirement branch that could never fire, and the two reset sites in `edit()` and `load_text()`. Docstrings that described the flag as live are rewritten to record that the window is gone and why; the historical prose that explains the deletion is kept deliberately.

**The two tests that asserted on the private flag were repointed, not deleted.** They now assert on `copy_in_flight` — the predicate the app's `ctrl+c` rung actually consults — so each stays a statement about guarded behaviour rather than about a field's name:

- `test_a_receipt_retires_even_when_the_caret_moved_before_the_edit` (R18-1): still pins the receipt/gesture lifetime asymmetry.
- the steering-approval copy test (round 19 MAJOR-1): still pins that collapsing the highlight hands the key back to the draft.

Verified non-vacuous by mutation: forcing `copy_in_flight` to `return True` fails the repointed assertion with `assert not True where True = Editor().copy_in_flight`. Restored immediately after.

## Testing evidence

**Behaviour is unchanged, measured rather than asserted.** The `ctrl+c` ladder with a live keyboard range and a turn genuinely running (`session.streaming = True`) is byte-identical before and after this change:

```
ctrl+c x5, range live   -> aborts=[] every press, clipboard 'summarise', draft intact
esc, range live         -> aborts=['interrupted']          <- first press, agent stops
right (collapse)        -> selection ''
  then ctrl+c           -> draft cleared
  then ctrl+c           -> aborts=['interrupted']
```

So the repeat-copy while a range is live is the documented explicit-copy rule, and the user always retains an escape: `esc` interrupts immediately with the highlight up, and collapsing the caret hands `ctrl+c` straight back to the ladder.

The `/help` block was rendered from the real `OperatorApp` and read verbatim (output above), not reconstructed — the row is confirmed present, aligned, and within the content box in the frame a user sees.

**Gates**, run over the whole tree exactly as CI does:

```
flake8 .                      clean
black==26.1.0 --check .       618 files unchanged
isort==5.13.2 --check .       clean
pyright                       0 errors, 0 warnings, 0 informations
pytest tests/unit -q          8791 passed, 15 skipped, 5 failed (see below)
```

The 5 suite failures are **pre-existing and load-induced, not from this change**. This machine was at load average 254 (several agents running in parallel); 4 of the 5 pass on re-run in isolation, and the fifth (`test_the_loop_stays_responsive_while_several_subagents_run`, an explicit event-loop-stall bound) fails on a **pristine `origin/main` checkout** too — 1 fail / 2 pass across three base runs. All five are timing or subagent tests in files this PR does not touch.

Separately, `test_a_live_selection_never_swallows_the_interrupt` is flaky on this machine under load (5 identical runs: 2f/2f/2f/1f1p/1f1p, non-monotonic across ancestors, green in CI on this SHA). It is **not** addressed here — it is tracked with the test-determinism work under #142.

## Not done deliberately

- No new keybinding, and no change to the copy trigger or the `ctrl+c` rung ordering. Options 1 and 2 in #169 are moot now that the gesture works; this lands option 3's documentation half against a gap that turned out to be already closed.
- No version bump, and not merged — the manager runs one consolidated release.
- The flaky test above is left alone rather than patched here.


damianvtran:
### Rendered evidence

Frames are committed at [`docs/evidence/help-copy-key/`](https://github.com/damianvtran/local-operator/tree/dev-composer-kbd-copy/docs/evidence/help-copy-key) with a README explaining what to look at, plus `scripts/help_keys_shot.py` to reproduce either side:

```sh
env -u NO_COLOR TERM=xterm-256color .venv/bin/python \
    scripts/help_keys_shot.py OUT.svg 100x30
```

Captured from the real `OperatorApp` at 100x30 so `local_operator.tcss` applies, and **looked at** rather than merely generated.

**before.svg** (`origin/main`, 75f867bf) — the key reference under the command table is a lone `ctrl+v` row:

```
/agent, /agents     List agents, or speak to this session as one
ctrl+v              attaches an image or file from the system clipboard
```

**after.svg** — `ctrl+c` above it in the same block, same key column, continuation indented under the description:

```
/agent, /agents     List agents, or speak to this session as one
ctrl+c              copies the highlighted range in the composer
                    move the caret to hand the key back; esc interrupts
ctrl+v              attaches an image or file from the system clipboard
```

Geometry behind the pixels: the two composed rows are **64 and 71 cells** against the **76-cell** content box at 80 columns. Neither wraps to column 0 (the defect `_help_block`'s own paste-note comment records), the notes below do not reflow, and the block keeps its single blank-line separation from the `window title` group. `test_help_documents_the_composer_copy_key_and_its_release` asserts the 76-cell bound so a later rewording cannot quietly overflow it.

damianvtran:
### Agent review — round 1
Scope: 75f867bf4ed63a32f475e22fa3c2a51dc19afe83..b828301338df17dd6f615dcc85fb281fa85b7425

Small PR, proportionate round. I re-measured the retraction rather than taking it on trust, because the premise that #169 is already closed is what the whole PR rests on.

**The retraction is correct.** Driven independently on pristine `origin/main` (`75f867bf`) in my own worktree, clipboard pre-loaded with `PRE`:

| gesture | selected | clipboard after `ctrl+c` |
|---|---|---|
| `home`, 9x `shift+right` | `'summarise'` | `'summarise'` |
| `home`, `shift+end` | `'summarise the ingest failures'` | same |
| `end`, `shift+home` | `'summarise the ingest failures'` | same |
| `home`, `ctrl+shift+right` | `'summarise'` | `'summarise'` |
| `home`, `shift+ctrl+right` | `''` | `'PRE'` (unchanged) |

The last row reproduces the issue's original measurement and confirms it as a pilot key-name artifact: only that one spelling fails to select, and it fails at the *selection* step, not the copy step. Both acceptance criteria in #169 already hold on `main`. Closing it on documentation alone is the right call.

**The retracted interrupt defect is not real.** Driven with `session.streaming = True` (the real flag) through `_boot`, with a live keyboard range on screen:

- `ctrl+c` with the range live → `aborts == []`, clipboard takes `'summarise'`. Copies, does not interrupt.
- `esc` with that same range still live → `aborts == ['interrupted']`. Interrupt reachable immediately, one press, highlight still up.
- `esc` with no selection at all → `aborts == ['interrupted']`. Same one press.
- caret collapse (`right`) then `ctrl+c` → draft rung: `aborts == []`, `text == ''`, draft in `prompt_history`. Second `ctrl+c` does not quit.
- empty draft + streaming, `ctrl+c` → `aborts == ['interrupted']`.

So the ladder is intact in both directions and `esc` is an unconditional escape hatch. No blocker.

**F1 — MINOR — `tests/unit/tui/test_steering_approval.py:3031`.** The rewrite dropped `assert not editor.selected_text, "the highlight should be gone"` and did not replace it. The substituted `assert not editor.copy_in_flight` is not equivalent at that point: I measured `copy_in_flight` to be **already `False` immediately after the copy**, before the `right` press (`AFTER copy: selected='a half-typ' copy_in_flight=False`; `AFTER right: selected='' copy_in_flight=False`). It is False on both sides of the press, so the step the comment describes — "moving the caret collapses the highlight" — lost its only direct assertion. Not a behaviour change and not vacuous overall (the downstream `aborts == []` / `text == ''` / `prompt_history` assertions still pin the real rung, and the deleted `_copy_gesture` was equally constant at base, so this is not a regression the PR introduced). Fix is one line: keep the `selected_text` assertion alongside the new one.

Worth recording that I did check for the weakening you would expect here and did **not** find it: mutating `copy_in_flight` to `return True` fails both rewritten tests with `AssertionError: the copy should no longer divert the key`. Neither test was softened to make the deletion pass.

**Dead-predicate removal verified.** `_copy_gesture` / `_copied_selection` had no writer at base other than the four `= False` resets — `git grep` at `75f867bf` shows no assignment to `True` anywhere in the tree, so `copy_in_flight` genuinely reduced to `self._selecting`. Nothing outside the widget referenced either field; the sole consumer is `app.py:6214` (`mid_copy = editor.copy_in_flight`), whose own comment wants "is a drag in flight" and still gets exactly that from `_selecting`. Semantics preserved, not invented.

**`/help` copy is accurate, which is the point.** "copies the highlighted range in the composer" / "move the caret to hand the key back; esc interrupts" matches the measured ladder above rather than oversimplifying to "ctrl+c copies" — a user holding a highlight is told what returns the key and that `esc` works regardless, both of which I confirmed empirically. Re-measured the claimed widths from the real `SLASH_COMMANDS` table: `name_width` 20, rows 64 and 71 cells against the 76-cell box. Matches the PR body exactly.

**Scope is clean — documentation-only as claimed.** Filtering the `editor.py` diff to executable lines leaves only the two `__init__` deletions, `return self._selecting or self._copy_gesture` → `return self._selecting`, and the four dead reset lines. Everything else is docstring and comment. No keybinding, no change to `action_copy`, `_on_key`, or the `ctrl+c` rung ordering. `app.py` adds only the two `Text` rows.

**F2 — NIT — `local_operator/tui/app.py:14426-14479`.** The new `ctrl+c` comment block was inserted between the existing `ctrl+v` comment and the `paste_note = Text()` it documents, so that comment now sits ~30 lines from its code with an unrelated block and `copy_note`'s own code in between. It also reads "documented for the same reason as `ctrl+v` below" while the `ctrl+v` prose is above. Moving the new block below the `ctrl+v` comment (immediately above `copy_note`) restores the comment-adjacent-to-code pairing.

**Gates**, run over the whole tree at head as CI does: `flake8` clean, `black==26.1.0 --check` 619 files unchanged, `isort==5.13.2 --check` clean, `pyright` 0 errors/0 warnings. `test_help_documents_the_composer_copy_key_and_its_release` passes.

On the suites: a first combined run of `test_transcript_selection.py` + `test_steering_approval.py` at head showed `test_an_empty_message_end_keeps_the_streamed_prose` failing. I did not report it as a finding because I checked it — it passes 3/3 in isolation on `main`, 3/3 in isolation at head, and the identical combined run at head passed on rerun (315 passed). Load flake on this box, not the diff. The `test_the_inset_is_never_what_tips_a_long_subagent_list_over` CI failure is excluded as instructed.

Verdict: **clean**. No BLOCKER, no MAJOR. 1 MINOR, 1 NIT, both follow-ups rather than remediation blockers. This round is **TERMINAL**.

damianvtran:

Scope: 3c79116c..d14aa707
Both findings fixed in `e816232d` (rebased onto current `origin/main` as `d14aa707`).

**F1 — MINOR — dropped `assert not editor.selected_text` — FIXED** in `e816232d`.

The finding is exactly right, and I reproduced it before changing anything. Measured either side of the `right` press:

```
AFTER copy : selected='a half-typ' copy_in_flight=False copied=True
AFTER right: selected=''           copy_in_flight=False copied=True
```

`copy_in_flight` is constant `False` across the press, so it cannot witness the collapse the step describes — the explicit press deliberately claims no gesture window (`Editor.action_copy`), which is precisely why the substitution was not equivalent. `selected_text` is the predicate that changes.

The assertion is restored **alongside** the new one rather than replacing it, since the two pin different things: the highlight is gone, and the copy no longer diverts the key. A comment records why one cannot stand in for the other, so a future reader does not re-collapse them.

Verified non-vacuous by mutation: simulating the regression it guards (the caret move leaving the highlight up) fails the restored line with `AssertionError: the highlight should be gone`, while the other three checks still pass. Reverted immediately after.

**F2 — NIT — comment separated from its code — FIXED** in `e816232d`.

Correct, and it was worse than cosmetic: the block also said "for the same reason as `ctrl+v` **below**" while that prose had ended up above it. The `ctrl+c` block now sits immediately above `copy_note`, restoring the comment-adjacent-to-code pairing for both keys, and the wording is corrected to "above". The row order is stated where the appends happen rather than being implied by declaration order.

**Rendering is unchanged** — the reorder is comment-only. Re-rendered from the real app at 80 columns after the change:

```
ctrl+c              copies the highlighted range in the composer          <- 64 cells
                    move the caret to hand the key back; esc interrupts   <- 71 cells
ctrl+v              attaches an image or file from the system clipboard   <- 71 cells
```

Same two rows, same 64/71 cells against the 76-cell box, so the committed frame at `docs/evidence/help-copy-key/` still stands. I re-captured it to confirm and reverted the result: the only delta was Textual's randomised `terminal-<id>` CSS class names, with byte-identical text content, so re-committing it would have been churn.

**Verification**, run single-process per the RAM limit (10 foreign `pytest tests/unit` pools were live; I did not start an eleventh):

```
pass  steering: a retired copy hands the key back to the draft
pass  selection: receipt survives a caret move, gesture does not
pass  help: ctrl+c row and its release line
pass  help: ctrl+v row still intact
4 passed, 0 failed
```

`flake8` clean, `black==26.1.0 --check` 626 files unchanged, `isort==5.13.2 --check` clean.

**Rebased onto current `origin/main`** (`3c79116c`, now including v0.43.11), so this round is fresh against what would actually ship. On the CI red the manager flagged: `test_band_panels.py::test_the_inset_is_never_what_tips_a_long_subagent_list_over` is a pre-existing main flake, confirmed independently here — the identical SHA `b8283013` **passed** `test (3.12)` in the push run (33314399243) and failed in the PR run (33314400760), and the test passes 3/3 in isolation on this branch and on pristine main. `test_boot_typing_sends_prompt` likewise fails 3/3 on pristine `3c79116c` and is not from this diff.

damianvtran:
### Agent review — round 2
Scope: aaa5a5a38b70536a532bb184b2d1779feed4e6ec..d14aa707bab45caa60333457b690254374a82fb8

Round 1 was terminal against `b8283013`; the branch was then rebased and remediated, so this round re-establishes freshness against the current head. Scope is the full `base..head` range, not just the remediation delta. The delta itself is small (one restored assertion, one comment reorder), and it is genuinely clean — but reviewing the full range turned up one thing round 1 did not: the new test's width bound is looser than the widget's real wrap threshold.

**F3 — MAJOR — `tests/unit/tui/test_app_pilot.py:8974` — the `<= 76` bound admits the exact defect it exists to prevent. The real budget is 74.**

The new test asserts each composed row is `<= 76` cells, and the code comment above `copy_note` states the same 76-cell figure. Thirty lines below in the *same function*, `cmd_paste_note`'s comment states the opposite and is emphatic about it:

> At 80 columns the painted row may be at most 78 cells (the screen's own 2-cell inset), and the block adds a further 2-cell spine indent, so the composed `name_width + description` must be `<= 74`. This row composes to EXACTLY 74. One more character wraps it.

Both cannot be true. I measured it against the painted compositor rather than trusting either comment — substituting descriptions of exact composed length into `_help_block` and reading `app.screen._compositor.render_strips()` at 80x44:

```
composed= 73 painted= 73 wrapped=False
composed= 74 painted= 74 wrapped=False   <- exactly at the limit
composed= 75 painted= 64 wrapped=True    <- tail moves to its own line
composed= 76 painted= 64 wrapped=True
composed= 77 painted= 64 wrapped=True
```

The `cmd+v` comment is right and the new one is wrong. And the failure at 75 is precisely D2, not a cosmetic reflow — the tail lands in the KEY GUTTER, at the same indent as the keys themselves:

```
[23] indent= 4 '    ctrl+c              copies the highlighted range in the composer'
[24] indent= 4 '    ABCDEFGHIJ'                     <- wrapped tail, reads as a key row
[25] indent=24 '                        move the caret to hand the key back; esc interrupts'
[26] indent= 4 '    ctrl+v              attaches an image or file from the system clipboard'
```

Why this is MAJOR and not a nit, given the shipped rows are 64 and 71 and therefore fine today: the test is the guard the PR body and the README both advertise ("the test asserts that bound so a future wording cannot quietly overflow it", "the bound `test_help_documents_the_composer_copy_key_and_its_release` now pins"). As written it passes for rows at 75 and 76, which wrap into the key gutter — so the guard's window of failure is exactly the two widths where the defect first appears. This defect has now shipped twice in this block (#402 D2, then the `cmd+v` row at 75 cells, caught independently by all three round-1 reviewers), and `test_the_paste_key_rows_do_not_wrap_at_eighty_columns` exists because neither occurrence had a test. The `ctrl+c` rows are **not** covered by that test — it iterates `("ctrl+v", ...)` and `("cmd+v", ...)` only — so this new bound is the only thing standing between the `ctrl+c` row and a third occurrence.

Compounding it slightly: `name_width` is derived from the longest command name, so the budget SHRINKS when a longer command is added. A future `/something-longer` moves every row toward the threshold, which is the scenario where a 2-cell-too-loose bound gets exercised for real.

Fix is one line plus the comment: `assert len(row) <= 74` and correct the "76-cell content box" figures in `_help_block`'s `ctrl+c` comment (and in `docs/evidence/help-copy-key/README.md`, which repeats 76). Better still, assert against the painted frame the way the neighbouring test does, since a literal encodes today's `name_width`. Both current rows already satisfy 74, so this is a guard correction and not a behaviour change.

---

**The round-1 MINOR (F1) is properly answered, and I verified the restored assertion is not another constant.** `assert not editor.selected_text, "the highlight should be gone"` is restored at `test_steering_approval.py:3037` alongside the `copy_in_flight` check, with a comment recording why one cannot stand in for the other.

I mutated the behaviour it claims to check rather than trusting the remediation's own mutation. Patching `Editor.action_cursor_right` so a caret move does NOT collapse the range (anchor kept, caret moved) reproduces the regression the step describes, and the restored line is what fails, first:

```
>           assert not editor.selected_text, "the highlight should be gone"
E           AssertionError: the highlight should be gone
E           assert not 'a half-typ'
```

Note which assertion caught it: `copy_in_flight` is False on both sides of the press, so under this mutation it still passes. The restored line is doing work the substituted one could not, which is exactly the finding. Non-vacuous, correctly targeted.

**The retraction still holds at the new head.** Re-driven independently in my own worktree at `d14aa707`, with `session.streaming = True` (the real flag; `session.running` is the junk attribute that produced the withdrawn defect) through `_boot`:

| gesture | selected | clipboard after `ctrl+c` | aborts |
|---|---|---|---|
| `home`, 9x `shift+right` | `'summarise'` | `'summarise'` | `[]` |
| `home`, `shift+end` | `'summarise the ingest failures'` | same | `[]` |
| `end`, `shift+home` | `'summarise the ingest failures'` | same | `[]` |
| `home`, `ctrl+shift+right` | `'summarise'` | `'summarise'` | `[]` |

And the interrupt remains reachable with a live range:

```
5x ctrl+c, range live   -> aborts=[]              draft intact
esc, range live         -> aborts=['interrupted'] FIRST press, highlight still up
right (collapse)        -> selection ''
  then ctrl+c           -> aborts=[]  draft cleared (draft rung)
  then ctrl+c           -> aborts=['interrupted'], app still running
```

Both acceptance criteria in #169 hold on the current head. Closing it on documentation is right.

**Scope is still documentation-plus-dead-code — the remediation did not touch the gesture.** Filtering the full `base..head` diff to executable lines: `app.py` adds only the two `Text` rows and their two `lines.append` calls; `editor.py`'s only statement changes are the two `__init__` deletions, `return self._selecting or self._copy_gesture` -> `return self._selecting`, the dead `watch_selection` branch, and the four `= False` resets. Everything else is docstring and comment. No keybinding, no change to `action_copy`, `_on_key`, or the `ctrl+c` rung ordering. The remediation commit `e816232d` is a comment reorder in `app.py` plus the restored assertion — I diffed `a0b06792..d14aa707` and it contains nothing else. No dangling references to `_copy_gesture`/`_copied_selection` outside prose that deliberately records the deletion.

**The `/help` text is still accurate after the reorder.** It states the rule, not "ctrl+c copies": "copies the highlighted range in the composer" / "move the caret to hand the key back; esc interrupts" — which matches the ladder I measured above in all three parts (live range copies, caret collapse hands the key back, `esc` interrupts regardless). The F2 reorder fixed the real error that the prose said "`ctrl+v` below" while that block was above; it now reads "above" and sits adjacent to its code. Row order is unchanged.

**I looked at the frames.** Rasterised `docs/evidence/help-copy-key/{before,after}.svg` and viewed them, and separately re-ran `scripts/help_keys_shot.py` from the real `OperatorApp` at 100x30. The `ctrl+c` row and its continuation render in the same key block above `ctrl+v`, left-aligned in the same column, continuation indented to 24 under the description, single blank line preserved before the `window title` group. The committed frame matches what I rendered (the only text delta is scroll position, since the block is captured at the transcript end). Rendered widths 64 and 71, as claimed.

**Gates**, run over the whole tree at head as CI does: `flake8` clean, `black==26.1.0 --check` 626 files unchanged, `isort==5.13.2 --check` clean, `pyright` 0 errors / 0 warnings / 0 informations.

**CI is fully green on this head** — `test (3.12)` and `test (3.13)` both pass on both runs, along with lint, type-check and pip-audit. The `test_band_panels.py` red round 1 saw is gone and is excluded as instructed either way.

**Test coverage, and what I did NOT run.** This box was at load average 227-285 with 6 foreign `pytest tests/unit` processes live throughout, so I started no pool and ran nothing wider than the touched files. Single-process (`-n0`) on the full contents of both files this PR edits: `tests/unit/tui/test_steering_approval.py` + `tests/unit/tui/test_transcript_selection.py` -> **324 passed** in 15m02s, zero failures, including all the copy/receipt/ladder tests. Plus the touched `test_app_pilot.py` cases and the neighbouring wrap test -> 7 passed. **I did not run the full ~8800-test unit suite locally**; CI's two green legs are the coverage for the rest, and I am stating that rather than implying a local full run.

---

**Verdict: NOT clean — 1 MAJOR (F3).**

Everything the round was chartered to check came back good: the F1 remediation is correct and I proved the restored assertion can fail, the retraction holds at the new head, scope is unchanged, the `/help` wording is accurate, gates and CI are green. F3 is not a regression the remediation introduced — it was in the original `89bf1327` and round 1 missed it — but it is a real defect: the guard the PR advertises passes at two widths that wrap into the key gutter, in a block where that exact defect has shipped twice before. One-line fix (`76` -> `74` in the assert, and the same correction in the two comments and the README). This round is **not terminal**; a round 3 against the corrected head should be short.

damianvtran:

Scope: 8d2a5e28..be47ef82
F3 fixed in `5e440359` (rebased onto current `origin/main` as `be47ef82`).

**F3 — MAJOR — the width bound was 76 against a real ceiling of 74 — FIXED** in `5e440359`.

Confirmed independently before changing the number, by painting candidate rows through the same console the block renders to at 80 columns and counting the visual lines each occupies:

```
composed  73  ->  fits
composed  74  ->  fits
composed  75  ->  WRAPS (2 lines)
composed  76  ->  WRAPS (2 lines)
```

So 74 is the ceiling and my `<= 76` admitted exactly the two widths where the defect appears. The finding is right, including on why it is a MAJOR with no user-visible symptom: the shipped rows are 64 and 71, but I advertised this test in the PR body and the evidence README as the guard for future rewording, and a guard whose window excludes its own boundary is worse than none because the next person to reword these rows will trust it.

**Verified the corrected bound actually fires.** Padding the `ctrl+c` description to exactly 75 composed cells (keeping the word "copies" so the earlier content assertions stay green and the width check is the one under test) fails with:

```
AssertionError: 'ctrl+c              copies the highlighted range in the composer xxxxxxxxxx'
                is 75 cells and wraps at 80 columns
```

Reverted immediately after; the real rows pass at 64 and 71.

**Reconciled to one voice**, as asked. The function previously stated 74 in the `cmd+v` comment and 76 in mine, which is how it would get re-derived wrong a third time. `74` is now stated in the `ctrl+c` comment, the `paste_note` comment, the test, and the evidence README, each with the derivation (78-cell painted row less the block's 2-cell spine indent) rather than a bare number. `grep` for `76-cell`, `<= 76` and `76 cells` across all three files returns nothing.

**Verification**, single-process per the RAM limit — the box was at load 200+ with foreign `pytest tests/unit` pools live, so no pool was started:

```
pass  help: ctrl+c row, release line, 74-cell bound
pass  help: ctrl+v row intact
pass  help: paste rows do not wrap at 80      <- the pre-existing neighbour, unbroken
pass  steering: retired copy hands key back
pass  selection: receipt survives caret move
5 ok, 0 failed
```

Re-ran after the rebase onto current `origin/main`, so these are green against what would actually ship. `flake8` clean, `black==26.1.0 --check` 626 files unchanged, `isort==5.13.2 --check` clean. Rendering is unchanged — same two rows at 64 and 71 cells — so the committed frame at `docs/evidence/help-copy-key/` still stands and was not re-captured.

**One note for whoever picks this up next**, not a finding on this PR: `.prwork/` is excluded through `.git/info/exclude`, which is per-clone and not shared, so scratch files kept there are invisible to `git status` but still scanned by `flake8 .` and `black .` — my own probe scripts tripped lint that way before I moved them outside the tree. If `.prwork/` is going to be the standard scratch location under the new shared-`/tmp` rule, it belongs in the committed `.gitignore` and in the tools' exclude lists, or CI's lint leg will fail for whoever forgets.

damianvtran:
### Agent review — round 3
Scope: ba414381c4dfeafcea493fbeaac6205a8d312a14..be47ef821bc93c1ed5311ada50242c7f892cbf4c

Short round, as scoped: round 2 left exactly one open finding (F3), and everything else it checked was verified good. I re-verified F3's fix end to end and did not re-litigate the settled ground.

**F3 is correctly fixed, and I confirmed the number independently rather than trusting either round.** I did not take 74 from the remediation or 76 from the earlier revision — I drove the real `/help` render path and read the compositor.

First attempt was my own bad probe, and it is worth recording because it produces a confidently wrong answer: filling the description with an unbroken run of one character reports "78 fits" at every width. Rich wraps on word boundaries, so an unbreakable token folds rather than wraps and the ceiling looks 4 cells higher than it is. With realistic prose (spaces, so there is a boundary to break at), painted through `_help_block` at 80 columns:

```
composed=74  ctrl+c              copies the highlighted range in the composer yyyyyyyyy   <- one line, painted 78
composed=75  ctrl+c              copies the highlighted range in the composer            <- WRAPS
             yyyyyyyyyy                                                                  <- tail at indent 4
```

**74 is right and 75 is the first failure**, and the tail lands at indent 4 — the key gutter, the same column as `ctrl+c` and `ctrl+v` themselves, so it reads as another key row. That is D2 verbatim, exactly as round 2 characterised it. The corrected bound is correct and the old one admitted precisely the two widths where the defect appears.

**The guard fires at the boundary — mutation-verified, not assumed.** Padding the shipped `ctrl+c` description in `local_operator/tui/app.py:14493`:

```
composed 74 (at the ceiling)   -> 1 passed
composed 75 (one over)         -> 1 failed
    AssertionError: 'ctrl+c              copies the highlighted range in the composer yyyyyyyyyy'
                    is 75 cells and wraps at 80 columns
    assert 75 <= 74
```

Restored after; `git diff` on the file is empty. Given the note that this board has repeatedly shipped guards that cannot fail at their own boundary, this one does fail there, in both directions.

**The number is now in one voice.** `74` in the test assertion, the `ctrl+c` comment, the `paste_note` comment and `docs/evidence/help-copy-key/README.md`. No width-76 claim survives anywhere in `_help_block`, the test, or the README — the only remaining "76" is the README's explicit history of the earlier wrong bound, which is the right place for it.

**Shipped rows all fit, with the ceiling confirmed from the other side.** Painted at 80 columns:

| row | composed | vs 74 |
|---|---|---|
| `ctrl+c` | 64 | ok |
| continuation (`move the caret…`) | 71 | ok |
| `ctrl+v` | 71 | ok |
| `cmd+v` | 74 | exactly at the ceiling, zero headroom |

`cmd+v` sitting at exactly 74 and not wrapping is independent corroboration: if the ceiling were 76 that row would have 2 cells of headroom, and if it were 73 it would already be broken on `main`.

**No text or behaviour changed in the fix.** The F3 commit `5e440359` touches one assertion literal and comments, nothing else — `git diff d14aa707..be47ef82 -- local_operator/tui/` has no non-comment line. The two `/help` strings are byte-identical to the round-2 head, so the committed frames still stand and the meaning did not drift.

**Scope is still documentation-plus-dead-code.** Against the current merge-base: `app.py` (the two help rows plus comments), `editor.py` (the `_copy_gesture` / `_copied_selection` removal, docstrings rewritten), tests repointed onto `copy_in_flight`, plus the evidence frames and capture script. `_copy_gesture` and `_copied_selection` have no live references left in `local_operator/` or `tests/`; the surviving mentions are the historical prose explaining the deletion. Nothing regressed from rounds 1-2.

---

**F4 — MINOR — `local_operator/tui/app.py:14484` and `:14523` — the reconciled comments state a derivation whose arithmetic gives 76, not the 74 they conclude.**

Both comments now agree on the number, which is what the round asked for, but they justify it with subtraction that does not come out:

> a 78-cell painted row less the block's 2-cell spine indent

78 − 2 = 76. That is the old wrong number, reached by the stated reasoning. The `cmd+v` comment has carried this phrasing for a while and the new `ctrl+c` comment now explicitly defers to it ("the same ceiling the `cmd+v` note below derives and for the same reason"), so the wrong derivation is now stated twice.

The measured indent is 4, not 2 — every painted key row sits at indent 4, and the ceiling tracks `terminal width − 6` exactly:

```
width  ceiling  painted@ceiling  indent
   80       74               78       4
   90       84               88       4
  100       94               98       4
  120      114              118       4
```

So the honest derivation is 78-cell painted row less a 4-cell indent, or equivalently width − 6 (2-cell screen inset + 4-cell block indent). The conclusion 74 is correct and the tests pin it, which is why this is MINOR and not MAJOR — nothing is wrong in the code or the guard. But the stated reason is the failure mode this whole finding chain is about: the file now agrees on 74 while explaining it in a way that re-derives 76, and "the comment's own arithmetic says 76" is exactly the argument someone will use to loosen the bound next time. One line in each comment fixes it: say the indent is 4 cells (or "width − 6"), not 2.

---

**Gate state.** `flake8 .` clean, `black==26.1.0 --check .` 634 files unchanged, `isort==5.13.2 --check .` clean, all over the whole tree as CI runs them. Targeted tests single-process (`-n0`) under the RAM limit, with 5-9 foreign `pytest tests/unit` processes live throughout — I started no pool:

```
pass  test_help_documents_the_composer_copy_key_and_its_release
pass  test_the_paste_key_rows_do_not_wrap_at_eighty_columns
pass  the two repointed copy_in_flight tests + the receipt test
```

**I did not run the full unit suite** — the box would not carry it alongside the foreign pools (one attempt was OOM-killed at 120s), so treat the above as targeted coverage, not a suite pass. CI is the authority on that. At posting time on `be47ef82`: `lint` and `type-check` are green; both `test (3.12)` and `test (3.13)` are still **queued** (not started) on the push run `33326255243` and the PR run `33326258633`, ~22 minutes in. Previous heads on this branch were fully green including both 3.12 legs (`33322073156` / `33322070024`). **Merging should wait on both test legs reporting green on this SHA**; nothing in this diff can plausibly break them, and the known flakes on this board are unrelated to it.

---

**Verdict: clean.** No blocker, no major. F3 is properly fixed, the corrected bound is the right number, and it is mutation-proven to fire at exactly the boundary it guards. The single new finding (F4) is a MINOR wording defect in a comment's arithmetic, not a reason for another round.

**This round is TERMINAL.** F4 and any remaining nits are follow-ups. Subject to both test legs going green on `be47ef82`, this is ready to merge.

damianvtran:

Scope: 8f4d21ac..2a974428
F4 fixed in `2a974428` (the rebase onto current `origin/main` rewrote the SHA; `2a974428` is the commit as it now stands on the branch).

**F4 — MINOR — the derivation beside the number still computed to 76 — FIXED.**

This is the right finding and it is the actual root cause. The number was correct after round 2, but the arithmetic printed next to it said "78-cell painted row less the block's 2-cell spine indent" — which is 76. Anyone re-deriving the ceiling from that sentence lands back on the wrong figure, which is exactly how 76 got asserted here in the first place.

Rather than patch the subtraction, I measured what the ceiling actually is, at three widths through the real `/help` compositor:

```
width  80: ceiling=74  width-ceiling=6  first_wrap=75
width  90: ceiling=84  width-ceiling=6  first_wrap=85
width 100: ceiling=94  width-ceiling=6  first_wrap=95
```

So the rule is **`terminal width - 6`**, not `78 - 2`. The reason a one-term derivation keeps coming out wrong is that the six cells are four separate reservations, no two declared in the same place:

```
1  TranscriptView left padding   (padding: 1 0 1 1)
1  reserved scrollbar column     (scrollbar-gutter: stable, D27)
2  the block's spine indent      (SPINE_INDENT, D20)
2  RichBlock's own Padding(renderable, (0, 2))
```

Both halves of your suggested fix are taken: the derivation now matches the measurement **and** is labelled as measured rather than computed, with an explicit "do not re-derive this from a single subtraction" and a pointer to the two rounds that got it wrong.

**Corrected in every voice that carried it**, since a file disagreeing with itself is the failure mode: the `ctrl+c` comment, the `cmd+v` comment (which had the same "78 less 2-cell inset" phrasing), the test's own comment, and the evidence README. `grep` for `78-cell`, `78 cells` and `2-cell inset` now returns only the sentence that quotes the old wording to explain why it was wrong.

**Comment-only.** No rendered output changed and the `<= 74` bound is untouched — still mutation-checked from round 2 (a row padded to 75 fails with `is 75 cells and wraps at 80 columns`). Re-confirmed at this head:

```
composed 74 -> fits
composed 75 -> WRAPS (2 lines)

 64 cells | ctrl+c              copies the highlighted range in the composer
 71 cells |                     move the caret to hand the key back; esc interrupts
 71 cells | ctrl+v              attaches an image or file from the system clipboard
 74 cells | cmd+v               same, where the terminal forwards it; not Terminal.app
```

**Verification**, single-process per the RAM limit (foreign `pytest tests/unit` pools live; no pool started):

```
pass  help: ctrl+c row, release line, 74-cell bound
pass  help: ctrl+v row intact
pass  help: paste rows do not wrap at 80
pass  steering: retired copy hands key back
pass  selection: receipt survives caret move
5 ok, 0 failed
```

`flake8` clean, `black==26.1.0 --check` 634 files unchanged, `isort==5.13.2 --check` clean. Rebased onto current `origin/main` before pushing, so this is fresh against what ships. CI reported separately once both legs finish.

Issue #440, opened 2026-08-30
/settings editing model: explicit edit/accept/cancel, so exploring a setting cannot change your setup


damianvtran:
## Summary

`/settings` currently has **three different editing contracts on one page**, and six gestures write to `config.yml` with no explicit accept. A user cannot open a setting to see what it can be set to without risking a change to their live setup. This issue tracks the audit and the proposed replacement: one contract for every row kind.

Full specification: **`~/local-operator-worktrees/settings-edit-model.md`**

Design audit only — no production code was changed, and no PR is open.

## The audit

Every row was produced by driving the real `OperatorApp` through `run_test()` against a scratch `LOCAL_OPERATOR_CONFIG_DIR`, with real key and mouse events, comparing `config.yml`'s **bytes** before and after each gesture.

| Row kind | `enter` | Browsing values | Moving off mid-edit | Explicit accept? |
|---|---|---|---|---|
| bool (`retry.enabled`) | **writes immediately** | n/a | n/a | **no** |
| enum (`providers.openai.api`) | expands, writes nothing | writes nothing | writes nothing | yes — `enter` on a choice |
| int/text/list (`retry.maxRetries`) | opens editor, writes nothing | n/a | **arrow/click COMMITS the buffer** | partial |
| cascade (`retry.fallbackChains`) | **opens a free-text editor over `str(dict)`** | n/a | **arrow commits — destroys the cascade** | **no** |
| chain / hop | opens / edits | n/a | commits | chain delete confirms; hop delete does not |
| read-only | reports the refusal | n/a | n/a | n/a (correct) |

**Gestures that mutate config with no explicit accept:**

1. `enter` / `space` / second-click on a **bool** row.
2. `up`, `down`, `ctrl+n`, `ctrl+p`, click-elsewhere during an **open editor**.
3. `shift+↑`/`shift+↓` on a **hop** (reorder).
4. `d` on a **hop** (delete).
5. `r` on **any row already at its default** — see below.
6. `enter` then any arrow on the **cascade** row.

### Three findings worth calling out

**A live data-loss defect (BLOCKER-class, on `main` today).** `action_activate` has no `Kind.CASCADE` branch, so the cascade row falls through to `_begin_edit` and opens a text editor seeded with `str(dict)`. One arrow key then commits it:

```
enter  -> buffer="{'cheap': ['openrouter/qwen/qwen3-coder', 'anthropic/claude-opus-5']}"
[...]

PR #447, opened 2026-08-30
fix(tui): the settings wheel scrolls the viewport, not the cursor

damianvtran:
## Summary

The `/settings` page held **two positions for one view**, and any gesture could overwrite the other's. Reported against v0.43.10 and reproduced against `origin/main` before anything was changed:

> "there seems to be both the scroll of the page and the scroll of the selected option which are conflicting, which still creates a bounce since you can scroll down to the bottom but then your selected option row is still somewhere above the top of the screen which ends up scrolling back up"

Ships as a **patch** (0.43.12 → 0.43.13): a bug fix to one page's scroll handling, no new capability.

**Round 1 remediation** is folded in: the code round was clean/terminal (S1 and S2 taken), and the UX round's MAJOR (U1) plus U3 are now fixed here rather than deferred — see "The off-screen cursor" below. U2 is verified unchanged and documented; U4 is not taken.

**Scope note.** This PR was originally briefed with a second improvement (default-value visibility and a conditional `r` hint). That was **dropped on the operator's instruction** — it lands with the forthcoming edit-mode redesign, which owns the same detail line and footer, so designing them separately would have produced two competing vocabularies on one row. This PR is scroll-only.

## The mechanism, which is not the obvious one

The body is a `ScrollableContainer` (viewport 14, virtual 60, `max_scroll_y=46` at 100x30). Two things could move the view:

- the **container's** own scroll offset — wheel and scrollbar drag;
- the **cursor**, via `_scroll_rows` → `_scroll_to_selection`, which re-derived the offset from the selected row.

The subtlety is *when* the widget handler runs at all. Textual handles the wheel on the container first and **stops the event while it still has somewhere to scroll**, so over the list `SettingsView.on_mouse_scroll_down` never ran. It ran only once the container hit its limit and let the event bubble — at which point the cursor, still parked where it was before the gesture, moved one row and dragged the viewport 46 rows back with it.

That is exactly why the bug reads as *"the page snaps back the moment I reach the bottom"*. Wheeling to the end is precisely how a user arrives in the only state where the defect can fire.

**Before**, wheel posted over the list, nothing else (100x30):

```
notch 22: scroll_y=44/46  selected=1
notch 23: scroll_y=46/46  selected=1     <- arrived at the bottom
notch 24: scroll_y= 2/46  selected=2     <- BOUNCE, back to the top
```

**After**, the same gesture:

```
notch 23: scroll_y=46/46  selected=1
notch 24: scroll_y=46/46  selected=1     <- the bottom holds
notch 27: scroll_y=46/46  selected=1
```

It was also **position-dependent**, which is the same defect seen from the side — one page, two scroll models, chosen by where the pointer happened to sit:

| wheel over | before | after |
|---|---|---|
| list / body | viewport scrolls, cursor stays | viewport scrolls, cursor stays |
| pane / detail / title / view | **cursor moves, viewport is yanked to it** | viewport scrolls, cursor stays |

## The model chosen: the wheel scrolls the VIEWPORT (option (a))

The cursor stays where the user put it and is allowed to go off screen; the next arrow key moves it and scrolls it back into view. This is what editors and list UIs do, it is what makes "scroll to the bottom" stay at the bottom, and it is the only arrangement in which the **scrollbar thumb is telling the truth** — under the old model the thumb reported the cursor's position, not the view's.

Option (b) — the wheel moves the cursor and the container is made non-scrollable — was rejected: it would have to defeat the container's own wheel handling and its scrollbar drag, leaving a visible scrollbar the user may not use, and `_body.can_focus = False` (UX round 3, U19) already establishes that this container must not be a second interactive surface.

**The invariant is now one source of truth per gesture:** the wheel and the scrollbar drive the viewport; keys that move the cursor scroll it into view. They can no longer disagree because no single gesture drives both.

Every path audited:

| path | viewport moved by | changed |
|---|---|---|
| wheel (`_scroll_rows`) | the container | **yes** — moves the view, not the cursor |
| `action_move` (`↑↓`, `ctrl+p/n`) | cursor → `_scroll_to_selection` | no |
| `action_section` (`PgUp/PgDn`) | cursor → `_scroll_to_selection` | no |
| `action_jump` (`Home/End`) | cursor → `_scroll_to_selection` | no |
| `_scroll_to_expansion` (enum/chain opens) | cursor, reveals the whole group | **yes** (round 1, U3) — also reveals a cursor that is ABOVE the viewport |
| `action_activate` (`enter`/`space`) | `_reveal_cursor`, before it acts | **yes** (round 1, U1) — a key that writes shows the row it writes to |
| `action_reset` (`r`) | `_reveal_cursor`, before it acts | **yes** (round 1, U1) — same rule; `r` writes too |
| `on_click` | nothing | no — a click names a visible row, so it must not scroll |
| scrollbar drag | the container | no — now consistent rather than competing |

Two Textual traps, both recorded in the code and in AGENTS.md:

- accumulate from **`scroll_target_y`**, not `scroll_relative` (which adds to the *current* offset, so a burst of notches all read the same stale value);
- pass **`immediate=True`**, because the default defers through `call_after_refresh` and the accumulation collapses anyway. Together these are why a fast flick moved two rows instead of forty in an intermediate revision.

The step comes from the app's **live** `scroll_sensitivity_y`, not from a constant copied out of it (review round 1, S1). It is a per-instance attribute set in `App.__init__`, so a hardcoded `2` matches today and desynchronises silently later — measured at `4.0`, the container applied 4 rows per notch over the list and this handler 2 everywhere else, `{list: 12, pane: 6, detail: 6, title: 6}`, which is the position-dependence this PR exists to remove, back again. `_WHEEL_ROWS` survives as the pre-mount fallback.

**The wheel no longer cancels an open editor, and that is a deliberate behaviour change** (review round 1, S2). The previous `_scroll_rows` called `_cancel_edit()` on every notch, so wheeling away from a row with the inline editor open discarded the edit. Under this model the wheel is a *look*, and throwing away a half-typed value because the user glanced elsewhere would be the worse surprise. The consequence to know about: the editor stays live and keyboard-reachable while scrolled off screen — keystrokes go on entering it and `enter` commits it. It is bounded by `_reveal_cursor`, which brings the editing row back into view before `enter` acts on it. Stated in `_scroll_rows`'s own docstring, where the next reader looks, rather than only in a test.

## `pagedown` — already fixed, now pinned

The brief carried this as an open defect ("settles at row 54 when the last selectable row is 59"). **That is the pre-#425 behaviour; it is already fixed on `origin/main`** by that PR's U3 remediation. Measured rather than assumed — `pagedown` to exhaustion reaches the true last row at every size and with an enum expanded:

```
(100, 30): settle=59 last=59 reached=True  scroll_y=46/46
(140, 40): settle=59 last=59 reached=True  scroll_y=36/36
( 80, 24): settle=59 last=59 reached=True  scroll_y=52/52
(120, 60): settle=59 last=59 reached=True  scroll_y=16/16
enum expanded (tui.theme): settle=94 last=94 reached=True
```

No code change. It was asserted at one size only, so this PR pins it across three sizes plus `pageup`, which is the regression guard it was missing.

## Tests, proven failing at base

Added to `tests/unit/tui/test_settings_view.py` and run at `origin/main` in a throwaway worktree with **only the test file** copied in.

Two **behavioural regression tests** that fail at base:

```
FAILED test_the_wheel_at_the_bottom_of_the_list_does_not_bounce_back
        AssertionError: the viewport bounced away from the bottom of the list: scroll_y=2 of 46
FAILED test_the_wheel_scrolls_the_same_way_wherever_the_pointer_sits
        AssertionError: the wheel over the pane moved the cursor
```

The bounce test asserts the exact reproduction: scroll the container to `max_scroll_y`, deliver one real wheel notch, assert the viewport did not jump. Wheel events are **posted**, not handled directly — the wiring under test *is* Textual's dispatch, and a direct `_scroll_rows` call exercises a path a real pointer only reaches at the very end of the list.

Three **preservation guards** that pass on both sides, pinning what the change had to leave intact: the arrows still scroll the cursor into view, paging still lands on the true last row, and a wheel gesture never writes config.

**Round 1 adds three more regression tests**, each proven failing at the round-1 head (`0361e18e`) in a throwaway worktree with **only the test file** copied in:

```
FAILED test_a_key_that_writes_reveals_its_row_before_it_writes
  AssertionError: the write landed while the cursor was still off screen —
  the user saw no frame change for a setting that changed
FAILED test_an_expansion_opened_from_off_screen_is_revealed
  AssertionError: the enum opened with every choice off screen, so the frame
  did not change and the press reads as having failed:
  choices=[4, 5] viewport=30..43
FAILED test_the_wheel_step_follows_the_apps_scroll_sensitivity
  AssertionError: at scroll_sensitivity_y=4.0 the wheel travels different
  distances by pointer position: {'list': 12, 'pane': 6, 'detail': 6, 'title': 6}
3 failed
```

The U1 test asserts **ordering, not end state**, which is what round 1 asked for: `settings_io.write_setting` is wrapped so the viewport is sampled *at the moment the value is stored*, and the assertion is that **config bytes are unchanged until the acted-on row is visible**. An end-state check would pass on an implementation that wrote first and scrolled afterwards inside one frame — which is exactly the "nothing visibly changed" complaint. It also asserts the press still *did* the thing on the first try, which is the line between a reveal and the rejected interlock.

The S1 test asserts parity at `scroll_sensitivity_y=4.0` specifically, because the drift is unobservable at the `2.0` default a copied constant happens to match.

An earlier revision of this PR also added a safety **interlock** ("the first `enter` re-centres, the second acts"). It **broke 19 existing tests** and is not in this PR. What ships instead is a *reveal*, which is a different thing — see below.

## The off-screen cursor: revealed before it is acted on (round 1, U1/U3)

Round 1 recorded this as the cost of the new model, and it was right: the wheel makes the cursor's off-screen state reachable from every region of the page, and this page writes immediately. Measured on the round-1 head at 100x30 — cursor parked on `retry.enabled`, view wheeled 15 notches away:

```
viewport: 30..43 | cursor 5 visible? False
ENTER on OFF-SCREEN BOOL -> config changed? True
viewport after: 30..43   [unchanged]
```

The captured frames are not merely similar, they are **byte-identical across the write**: `md5(before-offscreen-enter-pre) == md5(before-offscreen-enter-post)`. A setting changed and the screen did not move.

`_reveal_cursor` scrolls the cursor into view before `action_activate` and `action_reset` act on it. Same gesture, this head:

```
viewport: 30..43 | cursor 5 visible? False
ENTER -> config changed? True
viewport after: 4..17 | cursor visible now? True
```

**A reveal, not an interlock.** The press still does what it says on the *first* try; it just scrolls its target into view first, so the write is one the user watches happen. That distinction is exactly why this does not collide with the edit-mode redesign: the rejected interlock added a second, competing rule to the activate/commit paths #440 rewrites, which is how the 19 failures arose. This adds no rule to those paths at all — it moves the **viewport**, which is this PR's own subject. **All 372 settings tests pass**, including every activate/commit test the interlock broke.

**U3** is the same one-line remediation. `_scroll_to_expansion`'s guard only fired when `last >= offset + height`, i.e. for a group hanging off the *bottom* edge; with the cursor wheeled off the *top* the enum opened (`_expanded` set, `max_scroll_y` 46 → 48) while the viewport never moved and neither choice row was on screen — precisely the "`▾` with nothing under it" failure that method exists to prevent. Revealing the owning row first makes the existing group-fitting logic work from a viewport that already contains it:

```
before: choices=[4, 5]  visible=[]      revealed=False
after:  choices=[4, 5]  visible=[4, 5]  revealed=True
```

### U2 — verified, unchanged, and now documented

The first arrow after wheeling still teleports the view 22 rows, in both directions (`down` from a cursor above the viewport: `24-37` → `2-15`; `up` from below: `22-35` → `45-58`). This remediation does not touch it — the arrows already revealed the cursor through `_scroll_to_selection`, so there was nothing to change there. Round 1 called the destination *correct* and the missing *cue* the problem, and a cue is chrome: the detail line and footer are what #440's redesign owns. So U2 is recorded where the next reader meets it, in `_scroll_rows`'s docstring, stating what the return trip looks like — which is the minimum round 1 asked for ("worth a sentence there at minimum") — and deferred for the cue itself.

### U4 — not taken

`home` at 40x12 leaves the cursor unpainted because the body viewport is 1 row. Re-measured on this head: `height=1 offset=0 selected=1 cursor_painted=False`, identical to `origin/main`. Not a regression, not caused by the scroll model, and a real fix means deciding what a 1-row list should do with its headers — out of scope for a scroll-model PR.

**No commit semantics are baked into this change.** `_scroll_rows` moves the view and never the cursor, so it neither writes a value nor settles an open editor. `_reveal_cursor` moves the viewport only: it does not touch `_selected`, `_leave_row`, `_commit_edit`, `_write` or `_cancel_edit`, so #440 can still change *when* a value commits without touching any of this.

## Visual validation

Frames captured with a new `scroll_shot.py` driving the **real `OperatorApp`** with the reported gesture (40 wheel notches over the list). The existing `settings_shot.py` states cannot show this: they place the cursor with a key and scroll to it, exercising the cursor's claim on the viewport rather than the wheel's. Rendered to PNG and **looked at**, not read as markup. `qlmanage` completed under a 45s bounded wrapper (no stall this run).

| | before | after |
|---|---|---|
| 100x30 | stalls mid-list at `Web search`, thumb mid-track, `scroll_y=34/46` | reaches `Retired` / `Max learnings history`, thumb at track bottom, `46/46` |
| 140x40 | stalls at `Appearance`, thumb near the top, `scroll_y=9/36` | reaches the same true end, `36/36` |

**Round 1 adds two more before/after pairs** from `reveal_shot.py` and `expand_shot.py`, driving the real app through the U1 and U3 gestures. Rendered to PNG under a bounded 45s `qlmanage` wrapper and **looked at**, not read as markup.

| | before (`0361e18e`) | after |
|---|---|---|
| `enter` on an off-screen bool | no cursor anywhere in the list; the detail line reads `retry.enabled` and the footer lights `enter change · r default`, all describing a row that is nowhere on screen — **and the frame is byte-identical before and after the write** | the view scrolls to `Failover and retry`, the `›` cursor sits on `Retry failed calls` now reading `off`, and the detail line finally names a row you can see |
| `enter` on an off-screen enum | viewport unmoved at `30..43`, both choice rows off screen, `▾` open with nothing under it | `OpenAI API surface ▾` at the top with both choices under it and the radio dot on `responses` |

```
before  offscreen-enter  scroll_y=30/46  cursor_on_screen=False  config_changed=True  viewport_moved=False
after   offscreen-enter  scroll_y= 4/46  cursor_on_screen=True   config_changed=True  viewport_moved=True
before  offscreen-expand scroll_y=30/48  visible=[]      revealed=False
after   offscreen-expand scroll_y= 3/48  visible=[4, 5]  revealed=True
```

[`docs/evidence/settings-scroll/`](docs/evidence/settings-scroll/)

```
before  100x30  scroll_y=34/46  at_bottom=False
after   100x30  scroll_y=46/46  at_bottom=True
before  140x40  scroll_y= 9/36  at_bottom=False
after   140x40  scroll_y=36/36  at_bottom=True
```

## Real execution, every gesture walked

Driven against the real app, not asserted from geometry alone:

```
open                              scroll_y=  0/46 selected=  1
5 wheel-down                      scroll_y= 10/46 selected=  1
40 more wheel-down                scroll_y= 46/46 selected=  1   (clamped, no bounce)
scrollbar drag -> bottom          scroll_y= 46/46 selected=  1
ONE wheel-down after the drag     scroll_y= 46/46 selected=  1   (was: 2/46, selected 2)
arrow down                        scroll_y=  2/46 selected=  2   (cursor-follow intact)
end                               scroll_y= 46/46 selected= 59
home                              scroll_y=  0/46 selected=  1
pagedown x1                       scroll_y=  0/46 selected=  5
pagedown x11 (exhausted)          scroll_y= 46/46 selected= 59
```

Wheel parity across all six regions (3 notches each): every one now `scroll_y=6`, cursor unmoved. Clicking a visible row selects it without the view jumping; a second click activates, as before.

## Gates

Run whole-tree, exactly as CI does, on the round-1 remediation head (`22a6794c`):

- `flake8 .` — clean
- `black==26.1.0 --check .` — clean (625 files)
- `pyright --pythonpath .venv/bin/python .` — **0 errors, 0 warnings, 0 informations**
- `tests/unit/tui/test_settings_view.py` — **372 passed** (369 + the three round-1 regression tests)
- full `tests/unit` — **8992 passed, 15 skipped, 1 failed**

The 372 is the number that matters for the remediation: the rejected interlock broke **19** tests on the activate/commit paths, and the reveal that shipped instead breaks **none**.

The single full-suite failure is `test_subagent_view.py::test_history_unavailable_and_error_retry_keep_trajectory_fallback`, a known load-sensitive artifact under xdist — **passes 1/1 in isolation** on this head, and the file contains no reference to settings.

Also unchanged from round 1: `test_resume_subagents_todos` (timing flake, ~1 run in 4 at base as well), and `test_the_loop_stays_responsive_while_several_subagents_run` (#433), fixed by #438 on current main.

## AGENTS.md

Adds the convention this defect is the case study for: **one gesture owns the viewport at a time**, the wheel/cursor split, and the two Textual traps above — so the next widget that puts a cursor inside a `ScrollableContainer` does not rediscover them.

Round 1 extends it with the two rules that round surfaced: step a widget-level wheel handler by the **live** `app.scroll_sensitivity_y` rather than a copied constant (S1/S4), and treat the off-screen cursor as **a contract with the keys** — once the wheel can leave it behind, every key that acts on it must reveal it first, reveal-then-act rather than an interlock (U1).


damianvtran:
### Agent review — round 1

**Reviewer:** `reviewer` subagent (independent of the implementer) — anthropic/claude-opus-5
**Scope:** `3c79116c3adc02a4788b68019ac7950be999df9c..0361e18e847a0e34cbd0fe6c12cb839ebc4a58fc`
(base = `origin/main`, head = `dev-settings-scroll`)

I did not write this code. Everything below was re-derived from the diff and re-measured against a live app; I did not take the PR description's measurements on trust.

---

## Diagnosis verified before the fix

The PR claims the reachable bounce path is the wheel **bubbling** to the view once the container sits at `max_scroll_y`, and that a direct `_scroll_rows` call takes a path a mouse cannot. **Confirmed, twice over.**

At the Textual source (8.2.8, `widget.py`), the wheel is stopped *conditionally*:

```python
def _on_mouse_scroll_down(self, event) -> None:
    ...
    if self.allow_vertical_scroll:
        if self._scroll_down_for_pointer(animate=False):
            event.stop()
```

`_scroll_down_for_pointer` returns `False` once there is nowhere left to scroll, so the event only reaches `SettingsView` at the end of travel. The mechanism is exactly as described.

Empirically, driving **real `MouseScrollDown` events through `app.screen._forward_event` at real screen coordinates** (not `post_message` to a widget, and not a direct `_scroll_rows` call), at 100x30 (`max_scroll_y=46`, viewport 14, 60 rows):

```
BASE, one notch over the list with the container already at the bottom:
  scroll_y 46 -> 2, selected 1 -> 2      <- the reported bounce
HEAD, same gesture:
  scroll_y 46 -> 46, selected 1 -> 1     <- holds
```

The fix addresses the reachable path, not only the synthetic one.

## One source of truth for the viewport — audited, and it holds

I swept the wheel across **every screen row** at `x=33` (3 notches each, from `scroll_y=0`), resolving the widget under each point:

```
y= 1 SettingsView      scroll=6    y=13 settings-view-list   scroll=6
y= 2 settings-title    scroll=6    y=19 settings-detail      scroll=6
y= 3 settings-rule     scroll=6    y=20 settings-detail      scroll=6
y= 4 settings-columns  scroll=6    y=21 settings-hints       scroll=6
y= 5 settings-list     scroll=6    y=22 HintButton           scroll=6
...                                y=23 SettingsView         scroll=6
```

**Every region of the page: `scroll_y=6`, `selected` unmoved.** The position-dependence is gone. `shift` and `ctrl` modified wheels behave identically over the list and over the pane. Scrollbar drag (`scroll_to(max_scroll_y)`) and wheel-to-bottom agree.

Path audit against the PR's table — all confirmed by measurement, not by reading:

| path | result |
|---|---|
| `_scroll_rows` | moves the container only; never touches `_selected` |
| `action_move` / `action_section` / `action_jump` | cursor → `_scroll_to_selection`, unchanged |
| `_scroll_to_expansion` | unchanged; enum at row 51 expands, `scroll_y` 38 → 40, group end 53 revealed |
| `on_click` | measured: `scroll` 20 → 20, `selected` → 25. Does not jump |

## Cursor-follow survives

```
wheel 40 notches away:   selected=1  scroll_y=46  cursor off screen
press `down`:            selected=2  scroll_y=2   cursor_visible=True
```

And at the top end: `end` → 59/46, wheel up to 0, `up` → selected 58, `scroll_y=45`, visible. Arrows reclaim the viewport correctly from either direction.

## The ends agree

All four end-seeking gestures land on the same place (last selectable = 59):

```
end-key    selected=59  scroll_y=46/46
held-down  selected=59  scroll_y=46/46
pagedown   selected=59  scroll_y=46/46
wheel      selected= 1  scroll_y=46/46   (cursor deliberately parked, view at the end)
```

The PR's correction of the brief is right: **`pagedown` was already fixed on `origin/main`** by #425's U3 remediation. I confirmed this — `test_paging_lands_on_the_true_end_of_the_list` **passes at base**. The implementer declined to claim a fix it did not make and pinned the behaviour across three sizes instead. That is the correct call.

## Test integrity — the tests genuinely fail at base

Copied **only** the new test file onto a pristine `origin/main` worktree:

```
FAILED test_the_wheel_at_the_bottom_of_the_list_does_not_bounce_back
  AssertionError: the viewport bounced away from the bottom of the list: scroll_y=2 of 46
FAILED test_the_wheel_scrolls_the_same_way_wherever_the_pointer_sits
  AssertionError: the wheel over the pane moved the cursor — assert 5 == 1
2 failed, 3 passed
```

The two regression tests fail at base **with the reported symptom**, and the three preservation guards pass on both sides — which is what makes them guards rather than decoration. `_cursor_on_screen` is computed from container geometry rather than through a helper on the widget, so a tree without the fix fails on position and not on `AttributeError`. That detail is deliberate and correct.

---

## Findings

**S1 — MINOR — `_WHEEL_ROWS` hardcodes a value Textual exposes as runtime-mutable, and parity breaks if it changes.**
`settings_view.py:92` — `_WHEEL_ROWS = 2`, commented "Matches Textual's own `App.scroll_sensitivity_y` (2.0)". It does match *today*, but `scroll_sensitivity_y` is set per-instance in `App.__init__` (`app.py:748`), not a class constant, so it is a copy that can silently drift. Measured:

```
scroll_sensitivity_y=2.0: {list:6, pane:6, title:6, detail:6}   parity OK
scroll_sensitivity_y=4.0: {list:12, pane:6, title:6, detail:6}  parity BROKEN
```

At 4.0 the exact position-dependence this PR exists to remove comes back — over the list the container's own handler applies 4.0/notch, everywhere else this method applies 2. Nothing in the repo sets `scroll_sensitivity_y` today (grepped: the only hit is this comment), so it is not reachable by a user now, which is why this is minor and not major. One line fixes it: read `self.app.scroll_sensitivity_y` instead of the constant, keeping `_WHEEL_ROWS` as the pre-layout fallback.

**S2 — MINOR — the wheel no longer cancels an open editor, and the docstring's phrasing hides the delta.**
Base `_scroll_rows` called `_cancel_edit()`; the new one does not, and `_cancel_edit` accordingly loses that call site (base line 1335 has no counterpart on head). Measured consequence:

```
BASE: open editor on `hosting`, 40 notches -> editing_key=None (edit cancelled)
HEAD: open editor on `hosting`, 40 notches -> editing_key='hosting', cursor off screen
      typing 'xy' -> buffer='anthropicxy'   (keystrokes enter an invisible editor)
```

I think keeping the editor open is the *right* choice under the chosen model — cancelling an edit because the user looked elsewhere would be worse. The issue is disclosure. `_scroll_rows`'s docstring says it "neither writes a value nor settles an open editor", which is true but reads as *nothing happens to the editor*; the actual behaviour is that the editor stays live and keyboard-reachable while invisible. The full statement of this exists only in `test_the_wheel_leaves_the_cursor_where_the_user_put_it`'s docstring. It belongs in `_scroll_rows` too, since that is where the next reader looks. The change table in the PR body also lists `_cancel_edit` removal nowhere.

**S3 — MINOR — `except Exception` around the scroll is wider than the condition it documents.**
`settings_view.py:1405` catches everything and returns, while the comment names one cause ("the container has no size before layout"). I confirmed the real pre-layout exception is `NoActiveAppError` from `scroll_to`. As written, a genuine future defect inside `scroll_to` becomes a silently dropped wheel notch rather than a traceback. Narrowing it to `NoActiveAppError` (or at minimum `except Exception` with the notch counted) would keep the guard without the blanket. Note this mirrors the existing `try/except Exception` in `_scroll_to_selection` and `_scroll_to_expansion`, so it is consistent with the file — hence minor, not major.

**S4 — NIT — the AGENTS.md entry documents the trap but not the sensitivity coupling.**
The new convention block (`AGENTS.md:322-345`) is genuinely good — it captures `event.stop()`-while-scrollable and the `scroll_target_y`/`immediate=True` pair, which are the two things the next person will otherwise rediscover. It stops one step short: it does not say that a widget-level handler must use the *same step* as `app.scroll_sensitivity_y` or the gesture desynchronises (S1). One sentence would close it.

## Forward-compatibility with #440 (asked for explicitly)

**No hidden commit-on-move dependency.** `_scroll_rows` never touches `_selected`, never calls `_leave_row`, `_commit_edit`, `_write`, or `_cancel_edit`, and reads nothing but the container's scroll state. I verified a 15-notch and a 40-notch gesture leave the config file empty (`_values(tmp_path) == {}`). The `_leave_row()`-commits-on-move contract lives entirely in `action_move` / `action_section` / `action_jump`, all unchanged by this diff. The redesign can change when a value commits without touching this code.

One thing for #440 to inherit knowingly rather than discover: the off-screen-cursor hazard. With the cursor parked on a bool and the view wheeled away, `enter` writes `retry.enabled` while the row is off screen and the view does not follow. **This is not a regression** — I measured it at base too, where it is reachable identically via scrollbar drag (`scroll_y=46`, cursor at row 1, off screen). The PR discloses it as a known gap and correctly declines to add a re-centre interlock that would collide with #440 (an earlier revision's attempt broke 19 tests). Flagging it here so the redesign owns it explicitly.

## Gates — all run by me on the head commit, whole tree

| gate | result |
|---|---|
| `flake8 .` | clean |
| `black==26.1.0 --check .` | clean, 625 files |
| `tests/unit/tui/test_settings_view.py` | **369 passed** |
| `tests/unit/tui` (full) | 3840 passed, 12 skipped, 2 failed — see below |
| CI on head | all 9 checks green (lint, type-check, test 3.12/3.13, tui-e2e macos+ubuntu, cli/server-sanity, pip-audit) |

The two full-suite TUI failures are **not this PR's**. Both are in `test_composer_seam.py`, which contains no reference to settings, and I reproduced them at `origin/main`:

```
BRANCH isolated: FAILED test_closing_the_aside_hands_the_row_back[size1] — 1 failed, 68 passed
BASE   isolated: FAILED test_closing_the_aside_hands_the_row_back[size1] — 1 failed, 68 passed
```

Byte-identical assertion at base. Load-sensitive and pre-existing.

## Evidence and version

Before/after SVGs differ and show the claimed difference (the after-frame at 100x30 contains `Retired`, the before-frame does not — the view reaching the true end of the list). `docs/evidence/settings-scroll/scroll_shot.py.txt` follows the existing `.py.txt` convention used by `docs/evidence/cmd-chords/`, and `docs/` is excluded from both the pyright config and `[tool.setuptools.packages.find]`, so none of it ships in the wheel. `0.43.12 → 0.43.13` is the right bump: a bug fix to one page's scroll handling, no new capability, per AGENTS.md "Versioning".

---

## Verdict: **clean**

No blocker and no major findings. The diagnosis is correct and independently reproduced, the stated invariant actually holds across every path I could reach, the preserved behaviours are preserved, and the regression tests fail at base for the right reason. The scope reduction (default-value visibility deferred to #440) is not reviewed as missing.

**This round is TERMINAL.** S1–S4 are minors and a nit — follow-ups, not grounds for another round. S1 is the one I would most like picked up, since it is a one-line change that protects the invariant this PR was written to establish, but it is not reachable by a user today and does not block the merge.

damianvtran:
### UX review — round 1

Scope: 3c79116c3adc02a4788b68019ac7950be999df9c..0361e18e847a0e34cbd0fe6c12cb839ebc4a58fc

All findings below are for **PR #447 (settings scroll model)**.

**How this was walked.** Real `OperatorApp` under `run_test`, scratch `LOCAL_OPERATOR_CONFIG_DIR`, wheel notches posted as genuine `MouseScrollDown`/`MouseScrollUp` through `app.post_message` → `App.on_event` → `screen._forward_event`, so Textual offers each notch to the `ScrollableContainer` first and only bubbles to `SettingsView` when the container declines — the exact dispatch this PR is about. Calling `on_mouse_scroll_down` directly would have skipped it. Every gesture was re-run against a pristine `origin/main` worktree (`3c79116c`) to confirm the probe is sensitive to the bug, and `config.yml` bytes were compared around each gesture. Sizes: 100x30, 100x40, 100x20, 120x30, 80x24, 60x20, 40x12.

---

#### The operator's complaint is gone. Confirmed.

Baseline reproduces it precisely — wheeling down over the list, notch 24 snaps `scroll_y` 42 → 2 and drags the cursor 1 → 2:

```
origin/main:  after 21 notches: scroll_y= 42/46  selected=1
              after 24 notches: scroll_y=  2/46  selected=2   <- BOUNCE
```

On this branch the same gesture stops dead at the end and stays there through 15 further notches, with the true last row painted:

```
this PR:      after 24 notches: scroll_y= 46/46  selected=1
              extra 15 notches: scroll_y= 46/46  selected=1
              last_row_index=59  bottom_visible_row=59
              row 59: '    Conversation length           100'
```

Wheeling back up: 14 measurements, every delta negative or zero, **zero upward jumps**. Config bytes unchanged (0 → 0) across the whole session.

#### Wheel parity across regions. Confirmed.

Six pointer positions, 6 notches down then 6 up. Baseline: only the list scrolled the viewport; the other five moved the cursor 1 → 8 and left `scroll_y` at 0. This PR: **all six identical** (`scroll_y=12`, cursor unmoved at 1, then `scroll_y=0`).

| region | this PR | origin/main |
|---|---|---|
| list | `scroll_y=12 sel=1` | `scroll_y=12 sel=1` |
| read-only pane | `scroll_y=12 sel=1` | `scroll_y=0 sel=8` |
| detail line | `scroll_y=12 sel=1` | `scroll_y=0 sel=8` |
| title | `scroll_y=12 sel=1` | `scroll_y=0 sel=8` |
| footer hints | `scroll_y=12 sel=1` | `scroll_y=0 sel=8` |
| rule | `scroll_y=12 sel=1` | `scroll_y=0 sel=8` |

#### Other gestures agree

- **`end` / `pagedown` / held `down`** all settle on row 59, the true last selectable row, `scroll_y=46/46`. `pagedown` steps 5→17→24→31→39→45→54→59 and stops. (Note: this matches `origin/main` — the PR body states plainly that `pagedown` was already fixed by #425's U3 remediation and this PR only pins it with tests. I measured both; the claim is accurate and correctly scoped. No finding.)
- **`home` / `pageup` / wheel-up** all settle at `scroll_y=0`, visible window `(0, 13)`.
- **Click never moves the view.** Wheeled to rows 20-33 with the cursor off screen at row 1, clicked list row 25: `scroll_y` stayed 20, cursor landed exactly on 25, nothing written. Second click on the same row activates, still without moving the view.
- **Scrollbar drag** moves only the viewport (`scroll_y` → 46/46), cursor unmoved, nothing written, offset always within `[0, max_scroll_y]`.
- **Resize** while parked at the bottom re-clamps correctly at every size tested; offset stayed within range in all five.
- **Responsiveness**: a 200-notch burst completed in 236 ms and a `home` pressed immediately after still landed. The input loop never froze.
- **Contract intact**: untouched editor commit is a byte-level no-op; esc ladder (editor → page → leave) works and writes nothing; cursor is identity-anchored across a write (`retry.usageAwareFallback` stayed under the cursor at row 11).

---

### Findings

#### U1 — MAJOR — a key that writes acts on a row the user cannot see, and the frame does not change

Step: park the cursor on `retry.enabled` (visible, row 5), wheel down 15 notches, press `enter`.

The cursor is now off screen at row 5 while the user is reading rows 30-43. `enter` toggled `retry.enabled` and wrote `config.yml`, and **the painted frame before and after the write is byte-identical** — I diffed all 14 visible rows. The detail line still reads `Retry a failed provider call before surfacing the error.  retry.enabled` and the footer still lights `enter change · r default`, both describing a row that is nowhere on screen.

```
user wheels down 15 notches -> scroll_y=30/46  selected=5  cursor_on_screen=False
  ...user presses ENTER, meaning "change the row I can see"
  -> CONFIG WRITTEN: True
  -> viewport after: (30, 43)   [unchanged]
  -> frame after: identical to frame before
```

Attached frame `offscreen.svg`: no highlight anywhere in the list, yet the chrome advertises three keys that all act on an invisible row.

I want to be fair about the model choice: viewport-scrolling and letting the cursor leave the view **is** the right model, it is what editors do, and it is the only model where the scrollbar thumb tells the truth. I am not asking for that to be reverted. The finding is that this PR makes the off-screen-cursor state *reachable from every region of the page and permanent* (before, wheeling over the pane/title/detail dragged the cursor along, so it could not be left behind from there), while the page's chrome still speaks as though the cursor were visible. This is the same hazard the code already documents at `settings_view.py` L262-274 for a different cause (UX round 3, U19: "wrote a setting they never chose") — that one was fixed by taking the container out of the focus chain; the wheel now reopens the same door.

What a user would expect: either the chrome signals that the cursor is elsewhere (e.g. the detail line prefixes the row name, or a marker on the scrollbar/edge shows which way the cursor lies), or a key that *writes* first brings its target into view and then acts, so the user sees what changed. The cheapest version of the latter is a `_scroll_to_selection()` at the top of `action_activate`/`action_reset` when the cursor is outside `[scroll_y, scroll_y + height)` — one line, and it turns a silent write into a visible one. Note the arrow keys already do exactly this (see U2), so the page would then be consistent: *any* key that acts on the cursor reveals it first.

Pre-existing on `origin/main` in the narrow sense (the same probe writes there too), so I am not calling it a blocker on this PR — but this PR is what makes it an everyday state rather than an odd one, and it is the natural remediation to pair with the model it introduces.

#### U2 — MINOR — the first arrow after wheeling teleports the view 22 rows with no warning

Step: wheel down 12 notches (looking at rows 24-37, cursor at row 1), press `down`.

```
before: visible 24-37, selected=1
after:  visible  2-15, selected=2
```

The viewport jumps 22 rows and the content the user was reading is gone. Nothing was written, and the cursor did move by exactly one row as asked — so this is defensible and it is the documented cost of the model the PR chose. I am recording it as MINOR rather than MAJOR because the destination is *correct* and the user can wheel straight back; it is the absence of any cue that makes it read as the page losing their place. A brief highlight flash on arrival, or the same signal U1 asks for, would resolve both findings together. The PR's own docstring acknowledges "the cursor is allowed to go off screen" without saying what the return trip looks like; worth a sentence there at minimum.

#### U3 — MINOR — an expansion opened while the cursor is off screen is never revealed

Step: park on the `OpenAI API surface` enum (row 3), wheel down 15 notches, press `enter`.

```
after enter: scroll_y=30/48  expanded='providers.openai.api'
             choice rows are 4-5; visible: []
             expansion brought into view at all? False
```

The enum *did* open — `_expanded` is set and `max_scroll_y` grew 46 → 48 — but both choice rows are off screen and the viewport did not move, so the frame is unchanged and the press reads as having failed. This is precisely the failure `_scroll_to_expansion` was written to prevent; its docstring says "a `▾` marker and nothing under it and the expansion read as having failed to open". The guard at L708 only fires when `last >= offset + height`, and here the whole group is *above* the viewport, which the condition does not cover.

Two mitigations, in fairness: it self-resolves for a large enum (`tui.theme`, 54 choices, does scroll into view because its group spans past the viewport), and identical behaviour exists on `origin/main`. So it is not a regression — but it sits squarely inside the interaction this PR redefines, and the one-line fix is the same `_scroll_to_selection()`-before-acting that U1 proposes.

#### U4 — NIT — `home` on a very short viewport leaves the cursor unpainted

At 40x12 the body viewport is 1 row. `home` sets `scroll_y=0` and `selected=1`, but row 0 (`Model`, a header) is the only row painted, so the cursor on row 1 is not visible. `end` is fine. No crash, and 40x12 is far below any realistic terminal; identical on `origin/main`. Recording only so it is not rediscovered as new.

---

### Verdict

**No BLOCKER. One MAJOR (U1).** The change does what it set out to do: the reported bounce is gone, verified against a baseline that still exhibits it; the wheel means one thing across all six regions where it previously meant two; ends agree; clicks don't jump; nothing writes on any scroll gesture at any size; the page stays responsive under a 200-notch flick. The scroll model chosen is the right one and the code comments explaining it are unusually good.

U1 is the cost the new model brings with it, and it is a write-without-feedback path, which is why it is MAJOR rather than MINOR — but it is narrow, pre-existing in the strict sense, and U1/U2/U3 all collapse into one remediation (reveal the cursor before a key acts on it, and/or signal that it is off screen). I'd like that addressed or explicitly deferred with a ticket before this merges, since #440's redesign is where a fuller answer would land.

**This round is NOT terminal** pending a response to U1. U2, U3, U4 are acceptable as follow-ups.

damianvtran:

**Responding to:** UX review — round 1 (U1–U4) and Agent review — round 1 (S1–S4)
**Remediation head:** `22a6794c` (round-1 head was `0361e18e`)

Both rounds answered in one comment because U1, U3 and S1 turned out to be the same subject — the scroll model's contract with the keys — and it reads better as one story than as two.

---

## UX round

### U1 — MAJOR — a key that writes acts on a row the user cannot see — **FIXED** (`4c0f393a`)

Reproduced first, on your exact steps, at `0361e18e`:

```
bool row: retry.enabled at index 5
viewport: 30 .. 43 | cursor 5 visible? False
ENTER on OFF-SCREEN BOOL -> config changed? True
viewport after: 30 .. 43   [unchanged]
```

And the frames are not merely similar, they are **byte-identical across the write** — `md5(pre-enter) == md5(post-enter) == ac4a439420169bf42d830fdea9bb0ced`. A setting changed and the screen did not move. Your finding was exactly right.

`_reveal_cursor()` now runs at the top of `action_activate` and `action_reset`, scrolling the cursor into view before either acts. Same gesture on the remediation head:

```
viewport: 30 .. 43 | cursor 5 visible? False
ENTER -> config changed? True
viewport after: 4 .. 17 | cursor visible now? True
```

**A reveal, not the interlock.** You were careful to distinguish these and it is the load-bearing detail, so to be explicit: the press still does what it says on the *first* try. It scrolls its target into view and then acts, in that order. The rejected interlock ("first press re-centres, second acts") added a second, competing rule to the activate/commit paths #440 rewrites, which is how it broke 19 tests; this adds no rule to those paths at all — it moves the **viewport**, which is this PR's own subject. **All 372 settings tests pass**, including every activate/commit test the interlock broke. Your instinct that "the cheapest version is a `_scroll_to_selection()` at the top of `action_activate`/`action_reset`" was the right shape; that is what shipped.

One implementation note worth surfacing, because it is what makes the test meaningful. `_scroll_to_selection` gained an `immediate` flag, and the reveal passes it. Without it the scroll defers through `call_after_refresh` and lands in the same frame as the write by luck rather than by ordering — indistinguishable on the end state, but not the same thing, and not assertable. With it the reveal is genuinely ordered before the write, which is what lets the regression test assert **config bytes are unchanged until the acted-on row is visible** (your requirement, taken literally):

```python
def _spy(manager, setting, value):
    # Sampled BEFORE the value reaches disk.
    seen["bytes_at_write"] = config.read_bytes() if config.exists() else b""
    seen["visible_at_write"] = offset <= view._selected < offset + height
    return real_write(manager, setting, value)
```

Test: `test_a_key_that_writes_reveals_its_row_before_it_writes`. At `0361e18e` with only the test file copied into a pristine worktree:

```
FAILED test_a_key_that_writes_reveals_its_row_before_it_writes
  AssertionError: the write landed while the cursor was still off screen —
  the user saw no frame change for a setting that changed
```

It also asserts the press *did* the thing on the first try (`retry.enabled` is `False` afterwards), which is the assertion that fails if anyone later turns this into an interlock.

**What I did not do:** the chrome half of your suggestion (detail line prefixes the row name, or a marker showing which way the cursor lies). That is the detail line and the footer, which #440's redesign owns and which this PR was explicitly scoped out of. The reveal makes the write visible without needing it.

### U3 — MINOR — an expansion opened while the cursor is off screen is never revealed — **FIXED** (`4c0f393a`)

Same remediation, as you predicted. Your diagnosis of the guard was precise: `last >= offset + height` only catches a group hanging off the **bottom** edge, and with the cursor wheeled off the **top** the whole group is above the viewport, which that condition cannot see. `_scroll_to_expansion` now reveals the owning row first, so the existing group-fitting logic works from a viewport that already contains it.

```
before: expanded=providers.openai.api  scroll_y=30/48  choices=[4, 5]  visible=[]      revealed=False
after:  expanded=providers.openai.api  scroll_y= 3/48  choices=[4, 5]  visible=[4, 5]  revealed=True
```

Rendered and looked at: the after-frame shows `OpenAI API surface ▾` at the top with both choices under it and the radio dot on `responses`. Test: `test_an_expansion_opened_from_off_screen_is_revealed`, failing at `0361e18e` with `choices=[4, 5] viewport=30..43`. It also asserts the owning row stays on screen, since choices without the label saying what is being chosen are the failure one step over.

### U2 — MINOR — the first arrow after wheeling teleports the view 22 rows — **DOCUMENTED, cue deferred to #440**

Verified after the change, as you asked. It is **unchanged**, and that is the correct outcome rather than an oversight: the arrows already revealed the cursor through `_scroll_to_selection`, so there was nothing on that path for this remediation to alter. Measured on the remediation head, both directions:

```
wheel down 12, press `down`:  visible 24-37 -> 2-15   selected 1 -> 2   jump=22
`end`, wheel up 12, press `up`: visible 22-35 -> 45-58  selected 59 -> 58  jump=23
```

You called the destination *correct* and the missing *cue* the problem, and I agree with both halves. A cue is chrome — a highlight flash on arrival, or a marker on the edge — and the detail line and footer are exactly what #440 owns and what this PR was scoped out of. So what I have taken here is the minimum you asked for at the end of the finding ("the PR's own docstring acknowledges 'the cursor is allowed to go off screen' without saying what the return trip looks like; worth a sentence there at minimum"): `_scroll_rows`'s docstring now states the return trip with the measurement, and notes that every key acting on the cursor now behaves the same way, so the page is at least consistent about it.

### U4 — NIT — `home` on a very short viewport leaves the cursor unpainted — **NOT TAKEN**

Re-measured on this head: `40x12: height=1 offset=0 selected=1 cursor_painted=False` — identical to `origin/main`, and not caused by the scroll model. A real fix means deciding what a 1-row viewport should do with the contiguous header run that `_scroll_to_selection` deliberately reveals, which is a different question from this PR's. Recorded, per your intent, so it is not rediscovered as new.

---

## Code round (clean/terminal — picked up anyway)

### S1 — MINOR — `_WHEEL_ROWS` hardcodes a runtime-mutable value — **FIXED** (`dd39b5f1`)

Taken, and you were right that this is the one most worth picking up: it protects the invariant the PR exists to establish. The step now comes from the live `self.app.scroll_sensitivity_y`, with `_WHEEL_ROWS` kept as the pre-mount fallback exactly as you suggested. Your measurement reproduced:

```
FAILED test_the_wheel_step_follows_the_apps_scroll_sensitivity
  AssertionError: at scroll_sensitivity_y=4.0 the wheel travels different
  distances by pointer position: {'list': 12, 'pane': 6, 'detail': 6, 'title': 6}
```

The test pins parity at `4.0` specifically, since the drift is unobservable at the `2.0` default a copied constant happens to match — a test at the default would pass against the defect.

### S2 — MINOR — the dropped `_cancel_edit()` is disclosed only in a test docstring — **FIXED** (`dd39b5f1`)

Agreed on both the judgement and the objection. `_scroll_rows`'s docstring now carries the delta explicitly, labelled as a behaviour change: the wheel no longer cancels an open editor, keeping it live is deliberate under this model, and the consequence to know about is that the editor stays **keyboard-reachable while invisible**. The PR body says the same in prose, and the change table now lists the paths that changed rather than only the wheel. Worth noting the two findings interact in this PR's favour: `_reveal_cursor` bounds the hazard, since `enter` on an off-screen editing row now brings it back into view before committing it.

### S3 — MINOR — `except Exception` around the scroll is wider than the condition it documents — **DEFERRED**

Not taken, deliberately, and I want to give the reason rather than just the verdict. You noted yourself that it mirrors the existing `try/except Exception` in `_scroll_to_selection` and `_scroll_to_expansion`. Narrowing only the new one makes this file carry two conventions for the same guard, which is the defect I would be introducing to fix a smaller one — and narrowing all three is a refactor of methods this PR otherwise touches lightly, on paths #440 will be editing. Better as one consistent change than as a one-off here. No ticket filed; say the word if you would rather have one and I will open it.

### S4 — NIT — AGENTS.md documents the trap but not the sensitivity coupling — **FIXED** (`22a6794c`)

Added, plus a second paragraph U1 earned: that letting the cursor go off screen is **a contract with the keys**, not just a scroll decision, and the resolution is reveal-then-act rather than an interlock. The next widget that puts a cursor inside a `ScrollableContainer` now inherits the whole rule rather than half of it.

---

## Gates on the remediation head

Whole-tree, exactly as CI runs them:

| gate | result |
|---|---|
| `flake8 .` | clean |
| `black==26.1.0 --check .` | clean, 625 files |
| `tests/unit/tui/test_settings_view.py` | **372 passed** (369 + 3 new) |
| full `tests/unit` | **8992 passed, 15 skipped, 1 failed** |

The one failure is `test_subagent_view.py::test_history_unavailable_and_error_retry_keep_trajectory_fallback`, a known load-sensitive xdist artifact on the sanctioned list — **1 passed** in isolation on this head, and the file has no reference to settings.

All three new tests were proven failing at `0361e18e` in a throwaway worktree with **only the test file** copied in (`3 failed`), then the worktree removed.

Evidence frames for both fixed findings are in [`docs/evidence/settings-scroll/`](docs/evidence/settings-scroll/) (`before/after-offscreen-enter-100x30.svg`, `before/after-offscreen-expand-100x30.svg`) with their capture scripts, rendered to PNG under a bounded 45s `qlmanage` wrapper and looked at.

Version stays `0.43.13` — still a patch, still one page's scroll handling.

**Open for round 2 on `22a6794c`.** U1 and U3 fixed with tests and frames, S1/S2/S4 taken, U2 documented with the cue deferred to #440, U4 and S3 declined with reasons above.

damianvtran:
### Agent review — round 2

Reviewer: `reviewer` subagent (independent of the implementer), model `anthropic/claude-opus-5`
Scope: `b9191e04ba82c12b059e0df1e9026a01d42d52d8..9931d78fa55de7d1fe6f07c48c7a6b085505e998`

Round 2 exists only because the branch was rebased twice through a hand-resolved conflict after round 1 went terminal. Its question is narrow: **did the rebase lose anything, introduce anything, or break the interaction between #447 and #449?** This is not a re-litigation of round 1.

Every claim below was re-derived independently. I did not accept the implementer's arithmetic or their test counts.

## 1. Did the rebase lose anything? — No

**The branch is cleanly rebased.** `git merge-base origin/main HEAD` is exactly `b9191e04`, so there is no stale merge base and #449 (`9c327496`) is confirmed an ancestor of `origin/main`.

**The test file is provably append-only.** Rather than trust `67 = 59 + 8`, I diffed main's copy against the resolved one:

- Three hunks, all pure appends: `20a21`, `2367a2369,2583`, `2368a2585,2790`.
- **Zero** `<`-side lines: not one line from main was deleted or modified.
- Set difference of main's lines minus the resolved file's lines: **0 non-blank lines**.

So "nothing from main was lost" is not an assertion here, it is a property of the diff. Counts independently confirmed: main 59 tests, head 67, exactly 8 new.

**The source file kept every #449 behaviour.** All ten `CASCADE` / `_enter_cascade` / `_TEXT_EDITABLE_KINDS` landmarks are present, shifted only by #447's insertions. The set difference of main's source lines absent from head contains **only** the intended `_scroll_rows` / `_scroll_to_selection` replacement (the old cursor-moving body, `_cancel_edit()`, the old docstring) and no #449 line. `_TEXT_EDITABLE_KINDS = frozenset({INT, FLOAT, TEXT, LIST})` is intact at line 78.

## 2. Did the conflict resolution introduce anything? — No

- Exactly one `_painted_frame` (defined once at line 1878, five call sites). No duplicated-then-diverged helpers.
- No duplicate top-level `def` of any kind, so nothing is silently shadowed.
- `from textual import events` sits in the correct third-party group and is genuinely used.
- Every one of the 8 new tests carries substantive assertions (2 to 8 each, most with failure messages). None is a no-op.

**I mutation-tested the two fixes rather than trusting green.** Removing `_reveal_cursor()` from `action_activate` fails `test_a_key_that_writes_reveals_its_row_before_it_writes` with "the write landed while the cursor was still off screen". Hardcoding the wheel step reproduces exactly the `{list: 12, pane: 6}` desynchronisation S1 described. Both tests catch the defect they claim to. The worktree was restored and verified clean after each mutation.

## 3. Do the two features interact correctly? — Yes, verified across sizes

The implementer exercised this once at 100x30; I wrote my own probe and ran it at **100x30, 80x24, 140x40, 60x20 and 200x50**. At every size, with the cursor wheeled off screen (40 notches) and parked on the cascade row:

- The cursor stays put under the wheel (`cursor_parked=True`).
- `enter` reveals the row first (`revealed_after_enter=True`) and then opens the chain editor.
- **`enter` on the cascade row writes nothing** — config bytes identical, on-screen and off-screen alike, at every size. #449's data-loss guarantee survives the reveal.

`action_reset` has the same shape: `_reveal_cursor()` runs before `_current()` and before every write path, with the disarm captured earlier so the early returns cannot skip it. Ordering is correct in both paths.

## 4. Version — free

`pyproject.toml` is `0.43.17`; `origin/main` is `0.43.16`. Tags run to `v0.43.16` with no `v0.43.17` locally **or on the remote**. No collision this time.

## 5. Gates — all clean, run by me

| Gate | Result |
|---|---|
| `flake8 .` | clean, rc=0 |
| `black --check .` (26.1.0) | 635 files unchanged |
| `pyright` | **0 errors, 0 warnings** |
| `test_settings_view.py` | **379 passed** |
| whole `tests/unit/tui` | **3858 passed, 12 skipped, 0 failed** |

Two notes. The TUI suite came back with **no failures at all** — better than the "one sanctioned failure" reported; `test_subagent_view.py::test_history_unavailable_...` also passes in isolation, so that was load flakiness, not a real failure. And the 5 cascade tests from #449 plus the 8 scroll tests from #447 pass together in one run, which is the specific evidence that the rebase preserved both feature sets.

CI on GitHub is queued rather than green (Actions is saturated with 55+ runs); the local gates above are the CI command set run verbatim per AGENTS.md.

## Findings

None. No BLOCKER, no MAJOR, no MINOR, no NIT worth a round.

Numbering would have continued from S5 (round 1 used S1-S4, U1-U4); no new IDs were needed.

Round 1's deferred items are unchanged and remain correctly deferred: **S3** (`except Exception` wider than the documented condition) and **U2** (the return-trip jump, cue deferred to #440). Neither is a rebase regression and neither blocks.

## Verdict

**clean — and this round is TERMINAL.**

The rebase preserved everything from both PRs, the hand-resolved test file is append-only against main with no dropped assertion, and the two features compose correctly at every size I tested including from an off-screen cursor. There is nothing outstanding for the implementer to remediate, so **no round 3 is required**. This round is fresh against the current head `9931d78f`, and I am independent of the author.

Merging is gated only on CI going green.

PR #448, opened 2026-08-30
fix(harness): a silent tool-only turn must not carry an empty text block

damianvtran:
# PR Description

## Summary of Changes

A question that needs a tool and no prose kills the run with a provider 400:

```
● web_search Checking requests 2.31.0 advisories
● web_fetch Reading upstream advisory list
model stream failed: invalid request (HTTP 400): messages: text content blocks must be non-empty
```

The tools execute, then the **next** request of the turn is rejected and the
answer is lost. Reproduced from an ordinary question ("are there published
security advisories for requests 2.31.0?").

### Cause

#443 hoisted the assistant-text assembly out of the streaming delta loop. That
change is correct and worth keeping — rebuilding the whole accumulated string on
every delta is quadratic in response length and was the gigabyte-scale memory
regression that release fixed.

But the assignment it moved was **conditionally unreachable** where it used to
live:

```python
async for event in stream:
    if isinstance(event, StreamTextDelta):
        text_parts.append(event.delta)
        assistant.content = [TextContent(text="".join(text_parts))]   # only ran if a delta arrived
```

A turn with no text delta never executed it, so `content` kept the `[]` it was
constructed with. Hoisted to the end of the function it runs unconditionally, so
the same turn now produces `[TextContent(text="")]`.

```
Message(role="assistant").content        -> []
[TextContent(text="".join([]))]          -> [TextContent(type='text', text='')]
```

**A turn that calls a tool and says nothing is ordinary** — it is what most tool
steps look like — so this is not an edge case, and it gets more likely the more
readily agents reach for tools.

### Fix

Assign only when there is text, on the normal path and on the hard-cancel path
(same reasoning, same guard). A silent turn keeps `[]`, which is what every
provider already handles.

## Testing

**The live reproduction, before and after.** Same query, same machine.

Before (at `3627aa4b`, with the prompt change reverted to prove it is not the
trigger):

```
● web_search Checking requests 2.31.0 advisories
● web_fetch Querying OSV for requests advisories
✗ web_fetch failed
Error: invalid request (HTTP 400): messages: text content blocks must be non-empty
```

After, on this branch — the query completes and answers:

```
| CVE-2024-35195 | Session with verify=False keeps TLS verification off ... | 2.32.0 |
| CVE-2024-47081 | URL-parsing flaw leaks .netrc credentials              | 2.32.4 |
| CVE-2026-25645 | extract_zipped_paths() predictable temp filename       | 2.33.0 |
```

**Bisected rather than guessed.** The failure reproduces at `3627aa4b` with the
prompt corpus reverted to its parent, and does *not* reproduce at `2c4ebc77`
with the new prompts applied. That isolates it to this commit and clears the
prompt change shipped in 0.43.11.

**Regression test** (`test_a_tool_only_turn_carries_no_empty_text_block`):
asserts `content == []` after a tool-only turn, and separately that no empty
text block reaches the provider on the second request — the request where the
failure is actually observable. Confirmed it **fails without the guard**
(`+ TextContent(type='text', text='')`) and passes with it, so it pins the
regression rather than the implementation.

**Gates:** flake8 clean, black 625 files unchanged, isort clean, pyright 0
errors / 0 warnings / 0 informations. `tests/unit/harness`, `tests/unit/session`
and `tests/e2e`: 932 passed serially (one `test_comms` timeout assertion flaked
under load and passes in isolation on this branch and on the base).

### DIFF
--- a/local_operator/harness/loop.py
+++ b/local_operator/harness/loop.py
@@ -980,7 +980,13 @@ class AgentLoop:
             # ``context.messages`` on this path, but the object is observable,
             # and losing the partial answer on cancel is a behaviour change this
             # optimization has no business making.
-            assistant.content = [TextContent(text="".join(text_parts))]
+            #
+            # Guarded for the same reason as the assembly at the end of this
+            # function: a cancel before any text delta must leave ``content``
+            # as the empty LIST it was constructed with, not a list holding an
+            # empty text block.
+            if text_parts:
+                assistant.content = [TextContent(text="".join(text_parts))]
             raise
         except Exception as exc:
             # `error` below is handed straight to the UI, which prints it as a
@@ -1027,7 +1033,19 @@ class AgentLoop:
         # reports the text that did arrive — the frozen row on an aborted turn
         # is what the user is left reading. The hard-cancel path re-raises
         # above and assembles there for the same reason.
-        assistant.content = [TextContent(text="".join(text_parts))]
+        #
+        # ONLY when there is text. A turn that emits tool calls and no prose is
+        # ordinary — it is what every "call the tool, say nothing" step looks
+        # like — and this assignment used to be unreachable for it, because it
+        # lived inside the delta loop and no delta ever arrived. Hoisting it out
+        # made it run unconditionally, so such a turn started carrying
+        # ``[TextContent(text="")]`` where it used to carry ``[]``. Anthropic
+        # rejects that on the NEXT request of the turn with
+        # ``HTTP 400: messages: text content blocks must be non-empty``, which
+        # kills the run outright. The empty list is what the Message was
+        # constructed with and what every provider already handles.
+        if text_parts:
+            assistant.content = [TextContent(text="".join(text_parts))]
         assistant.tool_calls = [
             self._assemble_tool_call(state) for _, state in sorted(tool_states.items())
         ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "local-operator"
-version = "0.43.12"
+version = "0.43.13"
 description = "A team of proactive AI assistants that can work together behind the scenes to help you get mundane tasks done so you can focus on the fun stuff."
 readme = "README.md"
 authors = [{ name = "Damian Tran", email = "damian@gominerva.com" }]
--- a/tests/unit/harness/test_loop.py
+++ b/tests/unit/harness/test_loop.py
@@ -209,6 +209,64 @@ async def test_full_turn_text_tool_text():
 
 
 @pytest.mark.asyncio
+async def test_a_tool_only_turn_carries_no_empty_text_block() -> None:
+    """A turn that calls a tool and says nothing must keep ``content == []``.
+
+    Anthropic rejects a message carrying an empty text block outright:
+
+        HTTP 400: messages: text content blocks must be non-empty
+
+    and it is the NEXT request of the turn that fails, so the run dies after
+    the tool has already run. "Call the tool, say nothing" is an ordinary
+    model turn — it is what most tool steps look like — so this is not an
+    edge case.
+
+    The assembly used to live inside the delta loop, where it was simply
+    unreachable when no text delta arrived. Hoisting it out of the loop (to
+    stop the per-delta rebuild being quadratic in response length) made it run
+    unconditionally, which turned "no text" into an empty text block rather
+    than no content at all. The guard is what keeps the optimization from
+    changing what a silent turn sends.
+    """
+
+    executed: list[str] = []
+    stream = ScriptedStream(
+        [
+            # No StreamTextDelta at all: tool call only.
+            [
+                tool_call_delta(0, id="call_1", name="echo"),
+                tool_call_delta(0, args='{"text":"hi"}'),
+                StreamEndEvent(stop_reason="toolUse"),
+            ],
+            [StreamTextDelta(delta="done"), StreamEndEvent(stop_reason="stop")],
+        ]
+    )
+    context = LoopContext(system_blocks=["sys"], tools=[echo_tool(executed)])
+    loop = AgentLoop()
+
+    events = []
+    async for event in loop.run([Message.user("go")], context, make_config(stream), None):
+        events.append(event)
+
+    final = events[-1]
+    assert isinstance(final, AgentEndEvent)
+    assistant = final.messages[0]
+    assert isinstance(assistant, Message)
+    assert assistant.content == [], "a silent turn must not carry a text block"
+    assert len(assistant.tool_calls) == 1
+    assert executed == ["echo"]
+
+    # The regression is only observable on the SECOND request, which is the one
+    # that carries the first assistant message back to the provider.
+    replayed = stream.requests[1].messages
+    assert not any(
+        getattr(block, "text", None) == ""
+        for message in replayed
+        for block in (message.content or [])
+    ), "an empty text block reached the provider"
+
+
+@pytest.mark.asyncio
 async def test_a_tool_result_is_redacted_before_it_reaches_the_model() -> None:
     """The ``read`` path the reviewer reproduced: a credential written by one
     tool and read by another must not survive into the tool message."""


damianvtran:
### Agent review — round 1

Scope: f3ae0441491cc7e91f3178e26e098347fc015159..a934591a4047620376908750726e66e22f5c39d6

The harness fix itself is correct, minimal, and well evidenced. The regression test is honest — I verified it independently. One blocker sits outside the fix, in the version/branch state.

---

#### BLOCKER 1 — the version bump collides with an already-published release

`pyproject.toml:7` bumps `0.43.11` → `0.43.12`, but `0.43.12` is already tagged, released, and live on PyPI:

```
gh release view v0.43.12  → target 020821ccd55504cfcf9cb19c05b5cbafd6f42ea2, published 2026-08-30T13:49:54Z
curl pypi.org/pypi/local-operator/json → latest: 0.43.12
```

The branch was cut from `f3ae0441`, which is two commits behind `origin/main` (`3c79116c`); it is missing `020821cc chore(release): bump version to 0.43.12 (#444)` and `3c79116c test(conftest): ...(#438)`. `origin/main` already carries `version = "0.43.12"`.

I simulated the merge: it is conflict-free and resolves to `version = "0.43.12"` — the bump is a silent no-op against main, not a new version. Shipping this as-is means the tag `v0.43.12` already exists on a different commit and the PyPI upload is rejected as a duplicate, so **the fix never reaches users** even though the merge looks successful.

Fix in one line: rebase onto `origin/main` and bump to `0.43.13`.

(This also explains the misleading `git diff origin/main..HEAD` in the PR description, which shows a spurious 72-line `tests/conftest.py` deletion — that is #438 being absent from the branch, not a change this PR makes. Reviewing against the merge base `f3ae0441` gives the true 3-file diff.)

---

#### Verified correct — no findings

**The fix is complete.** `text_parts` has exactly three references (`loop.py:830` init, `:875` append, `:988`/`:1047` assembly). Both assembly sites are guarded. `grep` for `assistant.content = [TextContent(...)]` across `harness/`, `server/` and `session/` finds no third site with this shape — there is no duplicate in the subagent loop or the server bridge.

**#443's intent is preserved.** The assembly remains hoisted out of the delta loop at `:1047`, once per turn; the guard is an O(1) list-truthiness test. Nothing reintroduces the per-delta quadratic rebuild, and the memory characteristics are unchanged.

**The cancel path is right** (`loop.py:988`). The guard only suppresses the assignment when no text arrived, so a cancel mid-stream still overwrites `content` with the partial text the user was reading. The aborted/failed-stream path at `:1047` keeps the same property. The `event.delta` consumer contract at `:876` is untouched.

**The test is honest and pins observable behaviour, not implementation.** I reverted both guards in a scratch copy and ran it:

```
assert assistant.content == [], "a silent turn must not carry a text block"
E   assert [TextContent(type='text', text='')] == []
1 failed in 2.65s
```

It reproduces the exact regression. `stream.requests[1]` is sound — `ScriptedStream.__call__` (`test_loop.py:82-84`) records the real `ChatRequest` per provider call, so index 1 is precisely the request that carries the assistant message back and is where the 400 fires. Asserting on what reaches the provider is the right level.

**The bisection reasoning holds up in the code.** At `3627aa4b^` the assignment sat at line 876, immediately after `text_parts.append(event.delta)` *inside* `if isinstance(event, StreamTextDelta)` — genuinely unreachable when no delta arrives. `3627aa4b` moved it to two unconditional post-loop sites. The attribution to #443 rather than the 0.43.11 prompt corpus is supported by the diff alone; no live provider call is needed to accept it.

**Gates, run by me in `/tmp/lop-emptytext`:** flake8 clean (exit 0), `black==26.1.0 --check` clean (625 files), `isort==5.13.2 --check` clean, `tests/unit/harness` 355 passed `-n0`, `tests/unit/providers` + `test_transcript.py` 679 passed. Local pyright over the full tree exceeded my time budget, but CI `type-check` and `lint` are `success` on this exact head SHA.

---

#### MINOR 1 — the predicate tests the list, not the joined string

`loop.py:1047` guards on `if text_parts:`, but `:875` appends `event.delta` unconditionally, so a single empty-string delta yields `text_parts == [""]` — truthy — and re-creates `TextContent(text="")`.

This is safe today only because all four wire clients filter at source (`clients.py:1303`, `:1452`, `:1962`, `:2195` all gate on `if text:`). That is a real invariant but an implicit cross-module one, and nothing in `loop.py` states the dependency. `if any(text_parts):`, or joining once and testing the result, would make the guard self-sufficient. Not blocking — whitespace-only text (`" "`) is legitimately non-empty and correctly kept either way.

#### MINOR 2 — the provider-agnostic invariant has no central enforcement

The Anthropic serializer at `clients.py:1716-1718` copies `block.text` onto the wire with no non-empty check, so any producer of an empty `TextContent` triggers the same 400. `harness/types.py:207` already encodes the correct idiom (`[TextContent(text=text)] if text else []`) for `Message.assistant`, which is exactly the rule this PR is hand-applying at two more sites.

Candidates that bypass both: `session/session.py:509`, `:530`, `:542` build user messages from `message.details.get("text", "")`, which is empty for an aside carrying no text. Out of scope for this fix, but dropping empty text blocks in `_message_blocks` (or validating in `Message`) would make the class of bug unreachable instead of patching it per site. Worth a follow-up ticket rather than a change here.

#### NIT 1 — comment length at the primary site

The 11-line block comment at `loop.py:1036-1046` restates the incident in more detail than the second reader needs; the `:984-991` version says the same thing in four. Content is accurate and the why/constraint framing is right, so this is taste, not a request.

---

### Verdict

**Not clean — one BLOCKER**, and it is not in the code under review.

The harness change is correct, complete, faithful to #443's optimization, safe on the cancel path, and backed by a regression test I independently confirmed fails without the guard. If the diff were only `loop.py` + `test_loop.py` this would be a clean, terminal round.

It cannot merge as it stands: the branch is behind `origin/main` and bumps to a version that is already tagged and published to PyPI, so the release step would fail and the fix would not reach users. Rebase onto `origin/main` and bump to `0.43.13`; patch is the right materiality call for a regression fix of this size — no new capability, so a minor would be wrong.

Re-review after the rebase can be scoped to `pyproject.toml` alone — the two `MINOR`s and the `NIT` are follow-ups and are not grounds for another round.

damianvtran:

Head at remediation: `0d0407d7e1451bd8a960e20dafa904ad710855d1`

**BLOCKER 1 — version collides with an already-published release. FIXED.**

Correct, and the consequence you named is the part that mattered: the merge would have been conflict-free, resolved to `0.43.12`, and looked entirely successful while the fix never reached anyone — the tag already exists on `020821cc` and PyPI would reject the upload as a duplicate. A silent no-op is the worst failure shape for a regression fix, since the bug stays live behind a green merge.

Rebased onto `origin/main` (picking up `020821cc` and `3c79116c`) and bumped to **`0.43.13`**. Verified before pushing: `origin/main` reads `0.43.12`, `v0.43.12` is tagged and live on PyPI, and this branch now reads `0.43.13`.

The rebase also removes the phantom 72-line `tests/conftest.py` deletion you spotted — that was #438 missing from the branch, exactly as you diagnosed. The diff is now the three files this PR actually changes:

```
local_operator/harness/loop.py  | 22 ++++++++++++++--
pyproject.toml                  |  2 +-
tests/unit/harness/test_loop.py | 58 +++++++++++++++++++++++++++++++++++++++++
```

Gates re-run on the rebased head: flake8 clean, black 625 files unchanged, isort clean, `tests/unit/harness` 355 passed.

---

**On the two MINORs and the NIT — accepted as follow-ups, and worth recording why I am not folding them in here.**

You are right that `if text_parts:` tests the list rather than the joined string, and that this is only safe because all four wire clients filter empty deltas at source. That is a real implicit cross-module dependency. I am leaving it, deliberately: this PR restores a behaviour #443 changed, and the tightest possible diff is what makes it safe to ship against a live regression. `if "".join(text_parts):` would be the same cost and marginally more robust, but it changes the predicate's meaning on a path I have not exercised, and "harden while fixing" is how a one-line fix acquires a second bug.

The central-enforcement point is the more interesting one and is genuinely out of scope here: `clients.py:1716` copying `block.text` unguarded while `types.py:207` already encodes the right idiom is a provider-agnostic invariant with no single owner. That belongs in its own change with its own tests, not bolted onto a regression fix.

Thank you for verifying the test by reverting both guards rather than taking the claim — `assert [TextContent(type='text', text='')] == []` is exactly the failure it should produce, and confirming `stream.requests[1]` is where the 400 fires is the check that makes it a wire-behaviour test rather than an implementation mirror.

damianvtran:
### Agent review — round 2

Scope: 3c79116c3adc02a4788b68019ac7950be999df9c..0d0407d7e1451bd8a960e20dafa904ad710855d1

Narrow re-review of the round 1 remediation. Everything claimed in the remediation comment I verified independently rather than taking on report.

---

#### BLOCKER 1 (round 1) — version collides with an already-published release. CONFIRMED FIXED.

The branch is now cut from current `origin/main`:

```
merge-base(origin/main, HEAD) = 3c79116c = origin/main   # no longer behind
```

Version state, all four checks:

| check | result |
|---|---|
| branch `pyproject.toml:7` | `version = "0.43.13"` |
| `origin/main:pyproject.toml` | `version = "0.43.12"` |
| `v0.43.12` tag target | `020821cc` — a different commit, released 2026-08-30T13:49:54Z |
| `0.43.13` taken? | no tag (`gh release view v0.43.13` → not found), no PyPI (`/pypi/local-operator/0.43.13/json` → **404**; latest is 0.43.12) |

So the bump is a real bump against a free version, not a no-op, and the release step has somewhere to go.

**The phantom `tests/conftest.py` deletion is gone.** The diff against the merge base is now exactly the three intended files, and nothing else:

```
local_operator/harness/loop.py  | 22 ++++++++++++++--
pyproject.toml                  |  2 +-
tests/unit/harness/test_loop.py | 58 +++++++++++++++++++++++++++++++++++++++++
3 files changed, 79 insertions(+), 3 deletions(-)
```

---

#### The rebase did not touch the fix

`git diff a934591a..HEAD` restricted to the two substantive files is **empty** — `loop.py` and `test_loop.py` are byte-identical to the pre-rebase, already-reviewed head. The only differences between old and new head are `pyproject.toml` (the intended re-bump) and `tests/conftest.py` (inherited from #438, not authored here).

Both guards are intact and semantically unchanged:

```
830:  text_parts: list[str] = []
875:  text_parts.append(event.delta)
988:  if text_parts:
989:      assistant.content = [TextContent(text="".join(text_parts))]
1047: if text_parts:
1048:     assistant.content = [TextContent(text="".join(text_parts))]
```

Still exactly two assembly sites, both guarded, assembly still hoisted out of the delta loop so #443's non-quadratic property survives.

**Re-verified the test independently on the rebased head**, by reverting both guards in a scratch copy (`/tmp/rv2-scratch`, never the worktree, deleted after):

```
>       assert assistant.content == [], "a silent turn must not carry a text block"
E       AssertionError: a silent turn must not carry a text block
E       assert [TextContent(type='text', text='')] == []
1 failed, 68 deselected in 2.64s
```

The regression is still pinned. The test does not pass vacuously post-rebase.

---

#### #438 does not interact with this fix

`3c79116c` adds one session-scoped autouse fixture that warms the tiktoken BPE table when the on-disk cache already exists. It is relevant only to tests that *measure elapsed loop time* — the watchdog in `test_the_loop_stays_responsive_while_several_subagents_run`.

The new test asserts nothing about time. It has no `sleep`, no `perf_counter`, no elapsed bound; it asserts on message shape (`assistant.content == []`) and on what reaches the provider (`stream.requests[1]`). There is no path by which warming a tokenizer cache could make it pass when the guard is absent — and I proved that directly above, since the revert still fails *with* #438's conftest present. **#438 masks nothing here.**

---

#### On the deferred MINORs and the NIT — the judgement is defensible, and I checked the load-bearing claim

The implementer declined `if text_parts:` → `if "".join(text_parts):` on the grounds that all wire clients filter empty deltas at source, so the two predicates cannot diverge. I did not take that on trust. All four real `StreamTextDelta` emission sites in `providers/clients.py` guard before yielding:

```
1303 (chat-completions):  text = delta.get("content");  if text:
1452 (responses API):     delta = payload.get("delta"); if delta:
1962 (anthropic):         text = delta.get("text");     if text:
2195 (gemini):            text = part.get("text");      if text:
```

An empty string never enters `text_parts`, so `text_parts` truthy ⇔ `"".join(text_parts)` truthy on every path that exists today. The change would be a no-op on live behaviour and its only value is defence against a future client that stops filtering. Declining it in a regression fix is the right call, and "harden while fixing is how a one-line fix acquires a second bug" is a correct reading of the risk — the tightest diff is what makes this safe to ship against a live 400.

The central-enforcement item (`clients.py:1716` copying `block.text` unguarded while `types.py:207` already encodes `[TextContent(text=text)] if text else []`) is a genuine provider-agnostic invariant with no owner. It is correctly scoped out: it needs its own tests and touches every serializer. Neither item is required before merge.

I also confirmed the fix leaves no downstream consumer holding a stale assumption: `grep` finds **no** `.content[0]` indexing anywhere in `local_operator/`, and `Message.content` defaults to `Field(default_factory=list)` while `Message.text` folds over the list — so `content == []` yields `""` and cannot raise. The empty list is the shape the object was constructed with, which is the whole argument of the fix.

---

#### Gates — run by me, in the worktree, exactly as AGENTS.md specifies

```
.venv/bin/python -m flake8 .                      -> clean
uvx --from black==26.1.0 black --check .          -> 625 files would be left unchanged
uvx isort==5.13.2 --check .                       -> clean (exit 0)
.venv/bin/python -m pyright --pythonpath ...      -> 0 errors, 0 warnings, 0 informations
pytest tests/unit/harness -n0                     -> 355 passed in 36.65s
```

Run at full repo scope with the pinned versions, not narrowed and not with `--profile black`, per the AGENTS.md warning that the narrowed forms pass on files CI then rejects.

**CI status:** both workflow runs on head `0d0407d7` (`33317353221` push, `33317356664` pull_request) are still `queued` on GitHub's runner backlog, not failing. Local equivalents of every CI job — lint, type-check, test — are green above. Merge should wait for those two runs to report green, but there is no code-level reason to expect otherwise.

---

#### New findings

None. No BLOCKER, no MAJOR, no MINOR, no NIT. The remediation fixed exactly what round 1 raised, changed nothing else in the reviewed code, and the deferrals are argued correctly.

---

### Verdict

**CLEAN — and this round is TERMINAL.**

Round 1's single BLOCKER is resolved and verified on four independent axes (rebase base, branch version, main version, tag+PyPI availability of 0.43.13). The fix survived the rebase byte-identical, the regression test still fails without the guards, and #438 has no bearing on it. The two MINORs and the NIT stand as legitimate follow-ups, not merge conditions.

Merge once the two queued CI runs on `0d0407d7` report green. `0.43.13` is the right materiality call — a regression fix restoring prior behaviour, no new capability, so patch.

PR #449, opened 2026-08-30
fix(tui): enter on the failover cascade row no longer destroys the cascade

### LINKED ISSUE #440: /settings editing model: explicit edit/accept/cancel, so exploring a setting cannot change your setup

damianvtran:
## Summary

**Pressing `enter` on the failover cascade row in `/settings` destroyed the user's entire failover cascade.** Reproduced against `origin/main` before anything was changed, and the loss is silent and unrecoverable: the row reads `3 chains` one keypress before, `—` after, and `r` cannot bring it back.

`action_activate` had no `Kind.CASCADE` branch, so the cascade setting row fell through to `_begin_edit`, which opens a free-text editor seeded with `str(stored_value)` — for a mapping, a Python **repr**. Accepting it wrote that repr into `retry.fallbackChains` as a **string**. `read_chains` then returned `{}` (it requires a mapping), the whole cascade was gone, and `r` could not restore it because `action_reset` refuses on `Kind.CASCADE` by design — a cascade has no shipped default, it is entirely the user's own configuration.

Ships as a **patch** (0.43.11 → 0.43.12): a bug fix, no new capability.

Fixes #440.

## Reproduction, on the real page

`scripts/cascade_repro.py` (added) drives the real `OperatorApp` against a scratch config dir seeded with a real cascade, presses `enter` through the app's own binding, types one character, presses `enter` again, and prints the stored value on both sides.

**Before — `origin/main`:**

```
cascade row: retry.fallbackChains
chains BEFORE:            {'default': ['anthropic/claude-opus-5', 'openrouter/deepseek']}
after real 'enter':       editing? True | buffer: "{'default': ['anthropic/claude-opus-5', 'ope
chains AFTER enter-commit: {}
raw yaml: "{'default': ['anthropic/claude-opus-5', 'openrouter/deepseek']}x"
stored value is a mapping: False
```

The stored value is a **string** holding the repr. That is the data loss.

**After — this branch:**

```
cascade row: retry.fallbackChains
chains BEFORE:            {'default': ['anthropic/claude-opus-5', 'openrouter/deepseek']}
after real 'enter':       editing? False | buffer: ''
chains AFTER enter-commit: {'default': ['anthropic/claude-opus-5', 'openrouter/deepseek']}
raw yaml: {'default': ['anthropic/claude-opus-5', 'openrouter/deepseek']}
stored value is a mapping: True
```

## The fix

Two changes, both in `settings_view.py`, deliberately confined to the activate/edit path.

**1. `enter` on the cascade row travels into the cascade's own editor.** That editor already exists and already works — it was simply not reachable from `action_activate`. `_enter_cascade` puts the cursor on the first row of the group the cascade owns.

`enter` travels rather than doing nothing because the footer offers `enter change` on this row, and a lit hint whose key is inert is the same "nothing happens when I click" complaint UX round 1 U5 records. The target is resolved by **scanning forward** rather than by a fixed offset: `_cascade_rows` emits an unselectable `no cascade configured` line when the mapping is empty, and an already-open chain's hops sit inside the group too, so the first selectable row after the setting is the right target in every shape — the first chain, or `+ add a chain` when none exists.

**2. `_begin_edit` refuses to open on a kind that is not edited as text** — the belt-and-braces guard, because the next kind added without an `action_activate` branch should cost a dead key, not a destroyed config:

```python
_TEXT_EDITABLE_KINDS = frozenset({Kind.INT, Kind.FLOAT, Kind.TEXT, Kind.LIST})
```

Written as an **allow-list, not a deny-list**, so a new `Kind` is refused by default rather than silently handed to a free-text editor. Keyed on the **interaction** and not on the stored type: `Kind.LIST` is in the set on purpose because `web_search.providers` stores a list and genuinely edits as comma-separated text. A "is this value a scalar?" guard would have had to carve that out and would still miss a scalar-valued kind that needs its own widget.

## Which other kinds have the same fall-through?

**Audited all eight shipped kinds empirically**, not by reading the branch list: each one activated through the real `enter` binding, with the seeded buffer committed, checking whether a text editor opened and whether the stored value survived.

Against `origin/main`:

| kind | example key | stored type | editor opened? | verdict |
|---|---|---|---|---|
| BOOL | `retry.enabled` | `bool` | no | toggles, has its branch |
| ENUM | `providers.openai.api` | `str` | no | expands, has its branch |
| INT | `retry.maxRetries` | `int` | YES | correct — a scalar typed as text |
| FLOAT | `retry.usageReservePercent` | `float` | YES | correct |
| TEXT | `hosting` | `str` | YES | correct |
| LIST | `web_search.providers` | `list` | YES | **correct** — see below |
| **CASCADE** | `retry.fallbackChains` | `dict` | **YES** | **the bug** |
| READONLY | `session_retention_max_sessions` | `int` | no | already reports and returns |

`Kind.CASCADE` was the **only** kind reaching `_begin_edit` without a branch. `Kind.READONLY` — the other obvious suspect — was already handled at `action_activate` line 669 and again in `settings_io.validate`.

**`Kind.LIST` was the one that needed checking rather than assuming**, since it is the other non-scalar stored type. It is fine, and for a specific reason: `_edit_seed` renders a list as `", ".join(...)`, not as a repr, and `settings_io.coerce` parses that same comma-separated form back into a list with membership validation. Verified as a real round-trip rather than a no-op commit:

```
before: ['duckduckgo', 'tavily']
seeded buffer: 'duckduckgo, tavily' (comma-separated, not a repr)
after real edit: ['tavily', 'duckduckgo'] | still a list: True
```

After the fix, the same sweep shows `CASCADE  editor? no`, with every other row unchanged.

## The cascade's own editor still works

The concern with routing `enter` into the group is breaking the group. Walked end to end with real key presses, printing the stored cascade after every step:

```
1. start                     {'cheap': ['openrouter/c'], 'default': ['anthropic/a', 'openrouter/b']}
2. enter on cascade row      cursor=chain:cheap editing=None
3. open chain                _chain='cheap' hops visible=1
4. add hop                   {'cheap': ['openrouter/c', 'deepseek/d'], 'default': [...]}
5. reorder hop down          {'cheap': ['deepseek/d', 'openrouter/c'], 'default': [...]}
6. delete hop                {'cheap': ['openrouter/c'], 'default': [...]}
7. esc closes chain          _chain=None page open=True
8. d arms the ask            confirm='cheap' chains unchanged
9. d confirms                {'default': ['anthropic/a', 'openrouter/b']}
10. add a chain              {'default': [...], 'spare': ['openrouter/e']}
11. esc at chain level       page still mounted
```

Every level of the `esc` ladder still holds: `esc` on an open chain closes the chain (step 7), and `esc` at chain level does not leave the page (step 11).

## Visual evidence

Frames captured with `scripts/settings_shot.py`, extended with a `cascade-row` state, rendered and **inspected as images**. The existing `cascade` state selects a *chain* row and activates that, which has always worked — the bug is one row above it, on the setting row that owns the chains, and nothing in the evidence set photographed it. Both sides captured from the same base. Index: [`docs/evidence/settings-cascade-activate/README.md`](docs/evidence/settings-cascade-activate/README.md).

| before — `enter` on the cascade row | after — same key, same row |
|---|---|
| before | after |

The before-frame is the bug: an inline text editor is open on the cascade row holding the tail of the mapping's repr (`…ter/qwen/qwen3-coder']}`) with a caret after it, and the value column that read `3 chains` a keypress ago is now editable text. The after-frame has no editor — the cursor has travelled to the first chain (`› ▸ cheap  1 hop`), the detail line reads `enter opens the chain · d deletes it`, and `Failover cascade` still reads `3 chains`.

**The footer is the sharpest tell**, and it is the geometry behind the pixels (AGENTS.md step 4):

```
before  hints='↑↓ move · saves · enter  save · r default · ←→ panes · esc back to conversation'
after   hints='↑↓ move ·         enter  change · r default · ←→ panes · esc back to conversation'
```

The `saves` clause in the before-line is the page telling the user that **moving off the row will store what is in the buffer** — the repr. Nothing else about the layout moves: `rows=62`, `body.size=(62, 14)`, no screen scrollbar, on both sides.

The cascade's own editor renders unchanged after the fix: [`after-cascade-editor-100x30.svg`](https://raw.githubusercontent.com/damianvtran/local-operator/dev-cascade-fix/docs/evidence/settings-cascade-activate/after-cascade-editor-100x30.svg).

## Tests

Three regression tests, **all proven failing against `origin/main`** in a throwaway worktree with only the test file copied in — the source left untouched (`git diff --stat` showed the one file):

```
FAILED test_enter_on_the_cascade_row_does_not_destroy_the_cascade
  AssertionError: retry.fallbackChains is no longer a mapping after `enter` on the
  cascade row — the cascade has been destroyed:
  "{'default': ['anthropic/claude-opus-5', 'openrouter/deepseek']}x"
FAILED test_enter_on_the_cascade_row_opens_its_own_editor
  AssertionError: a text editor opened on a row with no scalar value
FAILED test_begin_edit_refuses_a_kind_that_is_not_edited_as_text
  AssertionError: _begin_edit opened a text editor on Kind.CASCADE
```

The first asserts the **stored value** is still a mapping and that `read_chains` still returns the chains — deliberately not "did an editor open", because the destroyed config is the bug and the editor is only the mechanism; a test checking `view._editing` would pass against any future fall-through that wrote through some other door. The third drives `_begin_edit` directly across `CASCADE`, `BOOL`, `ENUM` and `READONLY`, since the whole point of the guard is what happens when the branch that should have caught the row is missing, and confirms a text-editable kind still opens so the guard did not turn the editor off wholesale.

## Gates

| gate | result |
|---|---|
| `flake8 .` | clean |
| `black --check .` (26.1.0) | 626 files unchanged |
| `pyright` | 0 errors, 0 warnings |
| `tests/unit/tui/test_settings_view.py` | **367 passed** |
| full unit suite | 8985 passed, 3 failed |

All run over the whole tree as CI does.

The 3 full-suite failures are **pre-existing and load-induced**, not from this change. This machine is running many concurrent agent sessions (load average ~250 during the run), and the failures are all timing-bound assertions. Two independent checks:

- They are **not stable** — a different subset fails on each run (`test_cancel_backgrounded_bash_kills_process_group`, `test_prompt_child_costs_no_polling_tick`, and various `test_steering_approval.py` cases), and every one passes on an isolated re-run.
- They fail on **unmodified `origin/main`** (`3c79116c`) in a clean throwaway worktree — 5 failures there, a superset of the 3 seen here.

Nothing in the failing set touches `/settings`, the settings widget, or `settings_io`. Worth noting the `test_the_loop_stays_responsive_while_several_subagents_run` flake (#433) **no longer reproduces**. The earlier revision of this body credited #438's tiktoken warming fixture for that, which was wrong at the time: the branch was two commits behind and did not contain #438 (zero `tiktoken` references in its `tests/conftest.py`, against six on `main`). The round-1 rebase has since made it true — the fixture is now in this branch — but whatever made the flake pass before that rebase, it was not #438. Corrected per review round 1, m2.

## Notes for the reviewer

- **The editing-model redesign is deliberately not here.** Explicit edit mode, commit/cancel semantics, default visibility and conditional `r` are specified separately (`settings-edit-model.md`) and would collide with this diff. This patch restores correct behaviour for one row kind and nothing more.
- **Rebased onto current `origin/main`** (`3c79116c`), clean with no conflicts. The diff is confined to the activate/edit path to avoid colliding with the scroll fix in flight on `dev-settings-scroll`, which touches viewport/cursor handling in the same file — this change adds `_enter_cascade` and a guard clause and does not touch `_scroll_to_selection` or `action_move`.
- **Adjacent issue found, not fixed** (out of slice): `_render_value` renders a cascade as `N chains`, so the value column carries a summary rather than a value on that row. That is fine today but is the sort of display-vocabulary/edit-vocabulary split the redesign spec should account for.


damianvtran:
### UX review — round 1

Scope: f3ae0441491cc7e91f3178e26e098347fc015159..5179f6174a79acb4e945596f4b09a7224254bb59

All findings below are for **PR #449 (cascade data-loss fix)**.

**How this was walked.** Real `OperatorApp` under `run_test`, scratch `LOCAL_OPERATOR_CONFIG_DIR`, real key and mouse events through the pilot, `config.yml` bytes read and compared around every gesture, and the stored value's *type* inspected under `retry.fallbackChains` rather than only the parsed `read_chains()` result. Every probe was re-run against the PR's base (`f3ae0441`) to confirm it reproduces the shipped bug.

---

#### The data-loss bug is closed. Confirmed against a reproducing baseline.

Base `f3ae0441`, `enter` on the failover cascade row:

```
editing? 'retry.fallbackChains'
buffer  "{'default': ['anthropic/claude-opus-5', 'openrouter/deepseek"
```

A free-text editor seeded with the mapping's Python repr — one `enter` away from writing it back as a string. On this branch, the same press:

```
editing? None
selected moved to: 14 (kind='chain')
chains AFTER: {'default': ['anthropic/claude-opus-5', 'openrouter/deepseek']}
CONFIG BYTES IDENTICAL: True
```

`enter` now travels into the chain editor, the cursor lands on the first chain (`› ▸ default  2 hops`), the detail line switches to `enter opens the chain · d deletes it`, and **not one byte of `config.yml` changes**. Pressing `enter` again is also a byte-level no-op.

#### No route reaches a free-text editor over a non-scalar

I tried every route I could construct, including mouse and the footer's clickable hints:

| route | editor opened? | cascade intact | wrote? |
|---|---|---|---|
| `enter` | None | yes | no |
| `space` | None | yes | no |
| click, then click again (activate) | None | yes | no |
| double-click | None | yes | no |
| `r` (reset) | None | yes | no |
| footer `enter` hint button (clicked) | None | yes | no |
| `_begin_edit()` called directly | None | yes | no |

The last row is the belt-and-braces guard doing its job: it refuses with `this setting is not edited as text` rather than seeding a buffer. The allow-list shape is the right call — `BOOL`, `ENUM`, `CASCADE`, `READONLY` are all excluded by default, so a future `Kind` added without an `action_activate` branch costs a dead key instead of a destroyed config. `LIST` is correctly included and `web_search.providers` still edits as comma-separated text.

#### The cascade editor is intact end to end

Walked with real keys at 100x34, checking `read_chains()` and config bytes after each step:

- **open a chain** — `enter`, hops appear, nothing written
- **add a hop** — `+ add a hop` → `enter` → typed `groq/llama-3.3-70b` → `enter`; detail line correctly prompts `enter, then type: <provider>/<model>`; hop appended, cursor follows it
- **reorder** — `shift+up` moves the hop and the cursor follows it
- **remove a hop** — `d` deletes immediately (correct: one line, cheap to retype)
- **delete a chain** — `d` asks first: `▸ delete chain default and its 2 hops?  press d again to confirm · esc cancels`; `esc` cancels with the chain intact; `d`,`d` confirms
- **esc ladder** — from inside a hop editor with an uncommitted edit (`anthropic/claude-opus-5xyz`): esc 1 discards the edit and closes the editor, chain still open, **nothing written**; esc 2 closes the chain; esc 3 leaves the page. Final `read_chains()` byte-for-byte the seeded value.

Untouched-editor commit is a byte-level no-op and the cursor stays identity-anchored across a write. `tests/unit/tui/test_settings_view.py` + `tests/unit/test_settings_io.py`: **460 passed**.

---

### Findings

#### U1 — MAJOR — a user already hit by the shipped bug sees a page that contradicts itself, and `r` refuses to help

This is the case the PR exists for: someone who pressed `enter` on v0.43.10 has `retry.fallbackChains` stored as a **string**. I built that config through the page's own write path and opened `/settings` on it.

The page is usable — it opens, builds all 60 rows, navigates without crashing (`down` ×70, `end`, `home` all fine). But it says two contradictory things at once:

```
13| '    Failover cascade              {'default': ['anthropic/c…'   <- the raw repr, in the value column
14| '      no cascade configured'                                    <- directly underneath
```

The value column renders the corrupt Python repr as if it were the setting's value, while the group one row below says there is no cascade. A user cannot tell from this frame whether their chains are there or not.

Then the recovery key fails them. `r` is the page's documented mitigation for immediate-write having no undo, and the footer lights `r default` on this row. Pressing it:

```
before: type=str "{'default': ['anthropic/claude-opus-5', 'openrouter/deepseek']}x"
after : type=str "{'default': ['anthropic/claude-opus-5', 'openrouter/deepseek']}x"
DID `r` CLEAR THE CORRUPT VALUE? False
notice: 'r resets one setting; delete a chain with d on its row'
```

The corrupt string survives, and the advice it gives — "delete a chain with `d` on its row" — names a row that **does not exist**, because `read_chains()` returns `{}` and no chain rows are painted. The user is told to do something impossible, on the one key they'd reach for to fix it. Attached frame `corrupt.svg` shows exactly this state.

The `Kind.CASCADE` early-return in `action_reset` (L784-793) is well-reasoned for the *healthy* case — a cascade has no shipped default, so `r` genuinely cannot mean what it means elsewhere. But it returns before `settings_io.reset_setting`, which is precisely the call that would delete the corrupt key and put the user back to a clean empty cascade. For a value that is not a mapping at all, "reset this setting" is both meaningful and exactly what the user wants.

Suggested remediation: in that branch, if the stored value is not a `Mapping`, fall through to `reset_setting` (or clear the key) and say so — e.g. `cleared a malformed cascade`. That turns the one PR that touches this bug into the one that also repairs its existing victims, which seems like the right place for it. The value column should also not render a non-mapping as the setting's value; `—` plus the `no cascade configured` line below would be honest.

I'm calling this MAJOR rather than BLOCKER because the user is not *stuck*: `+ add a chain` works and overwrites the corrupt key cleanly (verified — after adding one, the stored value is a proper `dict` and the repr is gone from `config.yml`). But that requires knowing to ignore `r`, ignore the value column, and trust an `+ add` row that appears to be adding to a cascade the page just told them doesn't exist. Their original hops are unrecoverable either way — destroyed by the shipped bug, not by this PR.

#### U2 — MINOR — nothing tells a victim that anything is wrong

Following from U1: the page never states that the stored value is malformed. `_error` and `_notice` are both empty on open. A one-line notice on that row (`this cascade is malformed and is being ignored`) would convert a confusing frame into an actionable one, and would pair naturally with the U1 remediation. The failover layer already drops malformed chains silently at runtime, so the page is the only place this can surface.

#### U3 — NIT — `enter` on the cascade row lands on the first chain, not on the row's own affordance

`enter` moves the cursor down into the group. This is a good choice and the `_enter_cascade` docstring justifies it well (a lit `enter change` hint that does nothing is worse). Two small notes: with a chain already present the cursor lands on the first *chain* row, so a second `enter` is needed to actually open it — two presses to reach hops. And on an empty cascade it lands on `+ add a chain`, which is right. Not a defect; recording so the two-press path is a deliberate decision rather than an accident, and because #440's redesign will revisit it.

#### U4 — NIT — the PR's base is two commits behind `origin/main`

Base is `f3ae0441`; `origin/main` is `3c79116c` (`020821cc` version bump to 0.43.12, `3c79116c` tiktoken warming fixture). `pyproject.toml` still says `0.43.12`, which is the version already on PyPI — merging as-is publishes nothing new, and PR #447 in flight carries `0.43.13`. Only `pyproject.toml` and `tests/conftest.py` differ, so it rebases cleanly, but the version needs deciding against whichever of #447/#449 lands first. Process note, not a UX defect.

---

### Verdict

**No BLOCKER. One MAJOR (U1).** The fix does exactly what it claims and does it safely: the destructive fall-through is gone, verified against a base that still reproduces it; `enter` is now a navigation into the chain editor with zero bytes written; the allow-list guard closes the whole *class* of defect rather than the one instance; and I could not reach a free-text editor over a non-scalar by any route including mouse and the footer buttons. The cascade editor survives a full add/reorder/remove/delete/esc walk with the config always agreeing with the frame. This is a well-built fix and the comments explain the why properly.

U1 is not a defect *introduced* here — it is the shipped bug's aftermath — but this PR is the natural and probably only place it gets fixed, the remediation is small, and leaving it means the fix protects new users while abandoning the ones who already lost their cascade. I'd like U1 answered (fixed, or deferred with a ticket referencing #440) before merge; U2 folds into it.

**This round is NOT terminal** pending a response to U1. U3 and U4 are follow-ups.

damianvtran:
### Agent review — round 1

**Reviewer:** `reviewer` subagent (independent of the implementer) — anthropic/claude-opus-5
**Scope:** `3c79116c3adc02a4788b68019ac7950be999df9c..5179f6174a79acb4e945596f4b09a7224254bb59`
(`origin/main`..`dev-cascade-fix`; diff taken as `origin/main...dev-cascade-fix`, merge-base `f3ae0441`)

I reproduced the data loss myself at base before reading the fix, then swept every `Kind` through every activation route at both ends. The core fix is correct and the regression tests are real. Two findings block: a **stale detail line** this diff introduces on the new cursor-travel path, and a **version collision** — the branch bumps to `0.43.12`, which `origin/main` already carries and which is already tagged and live on PyPI.

---

#### Independent reproduction — the bug and the fix

Ran `scripts/cascade_repro.py` in a pristine throwaway worktree at `origin/main` (`3c79116c`), source untouched:

```
chains BEFORE:            {'default': ['anthropic/claude-opus-5', 'openrouter/deepseek']}
after real 'enter':       editing? True | buffer: "{'default': ['anthropic/claude-opus-5', 'ope"
chains AFTER enter-commit: {}
raw yaml: "{'default': ['anthropic/claude-opus-5', 'openrouter/deepseek']}x"
stored value is a mapping: False
```

Same script at head (`5179f617`):

```
after real 'enter':       editing? False | buffer: ''
chains AFTER enter-commit: {'default': ['anthropic/claude-opus-5', 'openrouter/deepseek']}
stored value is a mapping: True
```

Confirmed. The loss is real, silent, and the fix stops it.

#### Every `Kind`, every route (my own sweep, not the PR's table)

I wrote an independent sweep that activates one representative row per kind through the real `enter` binding, types a character, commits, and compares the **stored type** on disk. Base vs head:

| kind | key | route to `action_activate` | what it does | base | head |
|---|---|---|---|---|---|
| BOOL | `retry.enabled` | `enter`/`space`/click → `Kind.BOOL` branch (L636) | toggles, `_write` | `bool` ok | `bool` ok |
| ENUM | `providers.openai.api` | → `Kind.ENUM` branch (L643) | expands; a `choice` row commits at L635 | `str` ok | `str` ok |
| INT | `retry.maxRetries` | falls through → `_begin_edit` | text editor, `coerce` → `int` | `int` ok | `int` ok |
| FLOAT | `retry.usageReservePercent` | falls through → `_begin_edit` | text editor, `coerce` → `float` | ok | ok |
| TEXT | `hosting` | falls through → `_begin_edit` | text editor | `str` ok | `str` ok |
| LIST | `web_search.providers` | falls through → `_begin_edit` | `_edit_seed` joins `", "`; `coerce` splits back to `list` with membership validation | `list` ok | `list` ok |
| CASCADE | `retry.fallbackChains` | **no branch at base**; now `Kind.CASCADE` branch (L687) → `_enter_cascade` | travels to first selectable group row | **`dict`→`str` LOST** | `dict` ok |
| READONLY | `session_retention_max_sessions` | → `Kind.READONLY` branch (L682), also refused in `settings_io.validate` | reports and returns | `int` ok | `int` ok |

`CASCADE` was the only type-changing kind at base. Confirmed at head that no kind changes type.

I also drove the routes the PR did not enumerate, all at head:
- **`space`** (second binding, L190) → `editing=None`, cursor `chain`, stored `dict`. Safe.
- **Real `on_click` select-then-activate** → `editing=None`, cursor `chain`, stored `dict`. Safe.
- **Empty cascade** (`no cascade configured` unselectable line) → lands on `chain_add`, no editor. The forward scan handles it.
- **Cascade with a chain already open** (hops inside the group) → lands on the **first** chain, not a hop. The scan is correct in all three shapes.
- `hop` / `hop_add` / `chain_add` reach `_begin_edit` at L661 with `row.setting` unused, so they take the earlier branches and never touch the new guard.

The guard is correctly keyed on interaction rather than value shape: `LIST` genuinely stores a `list` and genuinely edits as text, and it still opens.

#### Recovery for an already-corrupted config

Seeded `retry.fallbackChains` with the exact repr string the bug leaves and opened the page at head:

```
page mounted OK, rows: 60          (no crash)
value column renders: "{'default': [...]}x"
read_chains: {}
rows the cascade owns: ['setting', 'empty', 'chain_add', 'header']
enter on the corrupted row -> editing? None | cursor: chain_add   (raw value untouched)
r on the corrupted row     -> notice: 'r resets one setting; ...' (raw value untouched)
+ add a chain              -> read_chains: {'rescue': [...]}  raw type now: dict
```

A corrupted config is **not** unopenable, the fix does not make it worse, and `+ add a chain` is a working recovery path that restores the value to a mapping. No dedicated recovery path is warranted. One cosmetic consequence noted as N1 below.

#### The cascade's own two-level editor still works

Walked it end to end with real key presses at head, printing the stored cascade after each step: `enter` into the group → open chain → add hop → reorder hop → delete hop → `esc` closes the chain (page stays) → `d` arms → `d` confirms chain delete → add a chain → `esc` at chain level (page stays) → open chain → `esc` closes chain only → `esc` leaves page. All eleven steps correct, `esc` ladder intact at every rung.

---

### Findings

**B1 — blocker — version `0.43.12` is already released; merging this publishes nothing.**
`pyproject.toml:7` bumps `0.43.11 → 0.43.12`, but the branch is **behind** `origin/main`, which already contains `020821cc chore(release): bump version to 0.43.12 (#444)`. Tag `v0.43.12` exists (points at `020821cc`), the GitHub Release is published, and PyPI already serves `0.43.12`. I simulated the merge in a throwaway worktree: it merges cleanly and leaves `version = "0.43.12"` — i.e. the fix would land on main carrying an **already-published version number**. `gh release create v0.43.12` would then fail on an existing tag, and PyPI would reject a re-upload of an existing version, so this data-loss fix would silently never reach users. Rebase onto current `origin/main` and bump to `0.43.13`. (Patch is the right bump — a bug fix, no new capability.)

**B2 — blocker — `_enter_cascade` moves the cursor without settling the row it leaves, so a stale detail line follows the cursor onto the chain row.**
`settings_view.py:718-729` sets `self._selected` directly. Every other cursor move on this page goes through `_leave_row`, which clears `_notice` because — quoting this file at L947 — *"The notice answered a keypress on THIS row, so it stops being true the moment the cursor leaves it."* Reproduced at head with real keys:

```
A. cursor on cascade row, press 'r'
   detail line: 'r resets one setting; delete a chain with d on its row'
B. press 'enter' (travels into the group)
   cursor now: chain:default
   detail line: 'r resets one setting; delete a chain with d on its row'   <-- STALE
   expected for a chain row: 'enter opens the chain · d deletes it'
C. same move made with 'down' (goes through _leave_row)
   cursor now: chain:default | detail line: 'enter opens the chain · d deletes it'   <-- correct
```

`r` then `enter` is a natural sequence on exactly this row: `r` is advertised in the footer, it answers with the notice that points the user at `d` on a chain row, and `enter` is how they get to that chain row. They arrive with the instruction for the previous row still on screen and no `d`-deletes hint for the row they are now on. This is the same "model changed, paint did not" class the file documents at L770-777 for the `r`-on-armed-chain bug and at D16. One line fixes it: clear `_notice` (and `_error`) in `_enter_cascade` before `self._repaint()`, or route the travel through the existing settle path.

**M1 — major — the guard's allow-list has no test holding `LIST` in it, so the one carve-out it depends on can be silently deleted.**
The PR argues correctly that `Kind.LIST` must stay in `_TEXT_EDITABLE_KINDS` (L78) because `web_search.providers` edits as comma-separated text. I removed `Kind.LIST` from that frozenset in a throwaway worktree and ran the whole settings suite: **367 passed**. My sweep then showed the real damage — `LIST web_search.providers ... editor? False`, i.e. the ordered-providers row becomes completely uneditable, with no test objecting. `test_begin_edit_refuses_a_kind_that_is_not_edited_as_text` asserts the positive side only on `retry.maxRetries` (an `INT`), so the kind whose membership actually needed justifying is the one left unpinned. Add `web_search.providers` to that test's positive assertion, or assert `Kind.LIST in _TEXT_EDITABLE_KINDS` with the reason.

**m1 — minor — `_enter_cascade` silently does nothing when the scan finds no selectable row.**
The loop at L718-729 can fall out without moving (a future `_cascade_rows` shape with no selectable member). `enter` then does nothing at all on a row whose footer says `enter change` — precisely the U5 complaint the docstring cites as the reason for travelling. A trailing `else`-equivalent setting `self._notice` would make the dead key legible. Not reachable today: `_cascade_rows` always appends a `chain_add` row (L394), which is always selectable.

**m2 — minor — the PR body's stated base is stale.**
"Rebased onto current `origin/main` (`f3ae0441`)" — `f3ae0441` is the merge-base, but `origin/main` is now `3c79116c`, two commits ahead. Same root cause as B1. Related: the body says the `test_the_loop_stays_responsive` flake "no longer reproduces after rebasing onto current `origin/main`, which carries the tiktoken warming fixture from #438" — but this branch does **not** contain #438 (`tests/conftest.py` on the branch has zero `tiktoken` references; `origin/main`'s has six). Whatever made that flake pass here, it was not #438. Resolved by the B1 rebase.

**n1 — nit — a corrupted cascade renders its full repr in the value column.**
With a repr string stored, `_render_value` (L2555) hits the `str(value)` fallback, so the value column shows `{'default': ['anthropic/claude-opus-5', ...]}x` where a healthy row reads `3 chains`, while the group below reads `no cascade configured`. Not introduced here and not worth fixing in this slice, but the two halves of the row disagree about whether a cascade exists. Worth a line in the editing-model spec (#440) rather than a change here.

**n2 — nit — `_TEXT_EDITABLE_KINDS` sits at module level between two rendering constants.**
It is placed at L66-78 between `_PANE_WIDTH` and the pane-line comment, both of which are layout concerns. An edit-semantics constant reads oddly there; nearer `_begin_edit` or `_edit_seed` would group it with what it governs. Purely placement.

---

### Gates (run by me, whole tree, as CI does)

| gate | result |
|---|---|
| `.venv/bin/python -m flake8 .` | clean |
| `uvx --from black==26.1.0 black --check .` | 626 files unchanged |
| `uvx isort==5.13.2 --check .` | clean |
| `tests/unit/tui/test_settings_view.py` | **367 passed** |
| `tests/unit/tui` (full) | 3837 passed, 12 skipped, **3 failed** |

The 3 TUI failures are `test_composer_seam.py::test_the_dock_is_one_fill_from_the_band_down_to_the_status_row` and `test_steering_approval.py::test_a_held_answer_key_resolves_cleanly_on_every_second_key`. I re-ran them in isolation as required, then ran the **same isolated selection against pristine `origin/main`** in a separate worktree: identical failures, `9 failed, 1 passed` on both sides. **Pre-existing, not caused by this PR.** Nothing in the failing set touches `/settings`, `settings_view`, or `settings_io`.

### Test integrity (verified, not taken on trust)

Reverted only `local_operator/tui/widgets/settings_view.py` to `origin/main` in a throwaway worktree, leaving the new tests in place (`git diff --stat` showed the one file, 75 deletions). All three new tests fail:

```
FAILED test_enter_on_the_cascade_row_does_not_destroy_the_cascade
  AssertionError: retry.fallbackChains is no longer a mapping after `enter` on the
  cascade row — the cascade has been destroyed:
  "{'default': ['anthropic/claude-opus-5', 'openrouter/deepseek']}x"
FAILED test_enter_on_the_cascade_row_opens_its_own_editor
  AssertionError: a text editor opened on a row with no scalar value
FAILED test_begin_edit_refuses_a_kind_that_is_not_edited_as_text
  AssertionError: _begin_edit opened a text editor on Kind.CASCADE
```

The primary test asserts on the **stored value** (`isinstance(stored(), dict)` plus `read_chains(...) == chains`), not merely that an editor failed to open — which is the right axis, since the destroyed config is the bug and the editor is only the mechanism. Confirmed by reading it rather than trusting the description. The gap is M1: the allow-list's positive side is under-pinned.

### Visual evidence

Checked the SVGs rather than trusting the description. Extracted text from `before-cascade-row-activated-100x30.svg`: the value column holds `…ter/qwen/qwen3-coder']}` with a caret, and the footer reads `↑↓ move · saves · enter save`. From `after-cascade-row-activated-100x30.svg`: `Failover cascade 3 chains`, cursor on `▸ cheap 1 hop`, detail `enter opens the chain · d deletes it`, footer `enter change`. The frames show what the body claims. `scripts/cascade_repro.py` follows the established `scripts/repro_*.py` convention.

### Verdict

**changes requested** — 2 blockers (B1 version collision with an already-published release; B2 stale detail line introduced by the new travel path), 1 major (M1 untested allow-list carve-out). The central fix is correct and I verified it independently at both ends; the blockers are a release-mechanics collision and one missing line on the new code path, both small. Not terminal — please post a `### Agent review remediation — round 1` and I will re-review against the new head.

damianvtran:

**Head:** `9c41b01d4e8a4e38495606aef022e7aafe2d9faa`
**Scope:** `65ca130a..9c41b01d` (rebased onto current `origin/main`)

Both blockers and the major are fixed. m1 rejected with reasoning, m2 corrected in the body, n1 superseded by the UX round's U1, n2 declined.

---

#### B1 — BLOCKER — version `0.43.12` is already released → **FIXED** (`9c41b01d`)

Bumped to **`0.43.14`**, not `0.43.13`. `0.43.13` would have collided a second time: it is claimed by #447, which is still open against the same base, and `origin/main` has since taken it too — **#448 landed mid-remediation and bumped main to `0.43.13`**, which is exactly the race you caught, one release later.

Your diagnosis of the mechanism was also load-bearing for getting this right. The first rebase silently DROPPED the branch's bump commit (`dropping 000ba195 chore: bump version to 0.43.13 -- patch contents already upstream`), because by then it was a duplicate of what main carried. A rebase does not preserve a version bump; it can delete one and leave the branch with none. I re-checked `git show origin/main:pyproject.toml` against the branch after each of the two rebases rather than assuming.

State at this head, verified rather than assumed:

```
origin/main version = 0.43.13     tags: v0.43.12, v0.43.13 exist
PyPI serves            0.43.12
this branch            0.43.14     free on both
```

#### B2 — BLOCKER — stale detail line from `_enter_cascade` → **FIXED** (`8a3b7898`)

Reproduced your walk exactly, then fixed it at the cause rather than the symptom. `_enter_cascade` now routes through `_leave_row()` instead of clearing `_notice` inline:

```python
if not self._leave_row():
    return
```

Clearing the notice would have fixed the frame while leaving `_enter_cascade` as a **second way to leave a row** — one that skips the commit-or-hold contract every other move obeys, so an invalid buffer could be abandoned by this door and not by `down`. Going through the settle path means the travel gets the whole contract, and the "notice dies when the cursor leaves the row" rule stays stated in exactly one place.

You flagged that this interacts with the U1 work since I was changing what `r` does on this row, and it did. The new `cleared a malformed cascade` notice is set AFTER `_save` (which clears `_notice` on success) and is `""` for every ordinary reset, and it is subject to the same `_leave_row` lifetime — so it cannot become the next stale-notice case. Pinned by `test_enter_into_the_cascade_does_not_carry_the_previous_row_notice`, which presses `r` then `enter` for real and asserts both that the notice is gone and that the chain row's own `enter opens the chain` contract is on screen. It fails at the old head with your exact string:

```
AssertionError: the notice from `r` on the cascade row followed the cursor onto the
chain row: 'r resets one setting; delete a chain with d on its row'
```

#### M1 — MAJOR — the allow-list carve-out is untested → **FIXED** (`538b6de0`)

`test_the_text_editable_kinds_allow_list_is_pinned` does both halves, because the membership assertion alone is not enough:

- pins the frozenset **by name**, so removing `Kind.LIST` fails and so does *adding* a kind — an addition hands its rows to an editor seeded with `str(stored_value)`, which is the #440 mechanism itself.
- exercises **every member through the real `enter` binding**, since a membership assertion would still pass on a kind that no longer opens.
- names `web_search.providers` explicitly and asserts its editor seeds with **comma-separated text, not a repr** — that seed is the entire justification for `Kind.LIST` being in the set, so the test pins the reasoning and not just the symbol.

I reproduced your experiment first. Deleting `Kind.LIST` in a throwaway worktree at the old head: the settings suite reports **367 passed**, and the new test fails alone.

#### m1 — `_enter_cascade` silently does nothing when the scan finds no selectable row → **REJECTED**

Not taken, on your own reasoning: it is unreachable today because `_cascade_rows` always appends a `chain_add` row, which is always selectable. A notice for a state that cannot occur is untestable by construction, and I would be adding a branch no test can enter to guard a shape that does not exist — the dead code becomes the next reviewer's question. The guard that actually pays off here is the one already present: the `_TEXT_EDITABLE_KINDS` allow-list means a future `Kind` reaching this file without a branch costs a refused key rather than a destroyed config, and M1 now pins it. If `_cascade_rows` ever grows a shape with no selectable member, that is the change that should carry the notice, with a test that can reach it.

#### m2 — the PR body's stated base and #438 claim are stale → **FIXED** (body edited)

Both corrected. The base now reads `3c79116c` where it claimed `f3ae0441`.

The #438 claim was wrong when written and I have said so in the body rather than quietly deleting it: the branch had **zero** `tiktoken` references in `tests/conftest.py` against six on `main`, so whatever made that flake pass, it was not #438. The rebase has since made the claim true — the fixture is in this branch now — but the body notes that it was not true at the time and credits your finding.

#### n1 — a corrupted cascade renders its full repr in the value column → **SUPERSEDED by U1, now FIXED** (`447da9b2`)

You judged this "not worth fixing in this slice" and filed it for the #440 spec; the UX round filed the same rendering as half of U1 (MAJOR) and asked for it here. The manager sided with the UX round, so it is fixed: the value column paints `—` for a non-mapping and the group line reads `malformed cascade — press r to clear it` rather than `no cascade configured`.

The disagreement was narrower than it looks. You verified correctly that a corrupt config opens without crashing and that `+ add a chain` restores a mapping — both true, and I re-confirmed both. The argument that carried is that the user is *misdirected* rather than merely inconvenienced: `r` is lit in the footer on that row, it is the documented mitigation for immediate-write having no undo, and it answered with advice naming a row that is not painted. Recovering via `+ add a chain` requires knowing to distrust the value column, distrust `r`, and trust an add row on a cascade the page just said does not exist.

Your scope instinct was still right in one respect and I held to it: `_render_value`'s healthy-case `N chains` summary is **untouched**, deferred to #440 as you and the previous implementer both wanted. The change is confined to the malformed case.

#### n2 — `_TEXT_EDITABLE_KINDS` placement → **DECLINED**

Left at module level with the other module constants. It is referenced from `_begin_edit` but it is a declaration about the settings model rather than about that function, and the file's existing convention is that module-scope constants live together with a `#:` comment carrying the why — `_PANE_WIDTH`, `_PaneLine`, `_Hint` and `_ROW_INDENT` are all there for the same reason. Moving one of them next to its single consumer would make the file have two conventions instead of one. Happy to move it if you feel strongly; it is a one-line change with no behavioural content.

---

### Verification

**Real path, not just a green suite.** Drove a #440 victim's config through the real `OperatorApp` with real key presses — `r` clears the key, `enter` travels with no stale notice, and a chain rebuilt from the recovered state stores as a proper `dict`. Full transcript in the UX remediation comment.

**Every new test proven failing at the unremediated head** (`000ba195`), in a throwaway worktree with only the test file copied in, so no source change could be masking the result.

### Gates (whole tree, as CI runs them, on the rebased head)

| gate | result |
|---|---|
| `flake8 .` | clean |
| `black --check .` (26.1.0) | 626 files unchanged |
| `tests/unit/tui/test_settings_view.py` + `tests/unit/test_settings_io.py` | **464 passed** |
| full unit suite (`-n 4`) | **8992 passed, 15 skipped, 0 failed** |

The full suite was completely green — none of the load-induced timing failures you and the UX reviewer both saw fired this run, so there is nothing to dismiss on this head. Run at `-n 4` rather than `-n auto` on instruction, after concurrent sessions on this machine hit an out-of-memory condition spawning 14 workers each.

Commits, one per finding so the SHAs mean something:

| commit | finding |
|---|---|
| `d15ebb96` | test helper — non-breaking spaces made frame assertions vacuous |
| `447da9b2` | U1, U2, n1 — `r` clears a malformed cascade; value column stops contradicting the line below |
| `8a3b7898` | B2 — `_enter_cascade` settles the row it leaves |
| `538b6de0` | M1 — allow-list pinned |
| `9c41b01d` | B1 — version `0.43.14` |

Ready for re-review against `9c41b01d`.

damianvtran:

**Head:** `9c41b01d4e8a4e38495606aef022e7aafe2d9faa`
**Scope:** `65ca130a..9c41b01d` (rebased onto current `origin/main`)

U1 and U2 are fixed together, in the one PR that was ever going to touch this. U3 is deferred to #440 with the reasoning written into the issue. U4 is resolved by the rebase.

---

#### U1 — MAJOR — a user already hit by the shipped bug cannot recover, and the page contradicts itself → **FIXED** (`447da9b2`)

Both halves, because they are one defect: the page told a #440 victim two incompatible things and then refused the key it advertised for fixing it.

**(a) `r` now repairs a malformed cascade.** The `Kind.CASCADE` early return in `action_reset` now branches on the stored value's SHAPE rather than on the kind alone:

- a **mapping** — the healthy case — still returns early with the existing notice. Your reading of why that early return exists is right and it is preserved verbatim: a cascade has no shipped default, so `r` genuinely cannot mean there what it means elsewhere.
- a **non-mapping** falls through to `settings_io.reset_setting`, which deletes the key, and reports `cleared a malformed cascade` — the wording you suggested.

**(b) The value column no longer renders a non-mapping as the setting's value.** It paints `—`, which is the page's existing glyph for "nothing to show here", and the group line below carries the explanation. `_render_value`'s healthy-case `N chains` rendering is untouched — that summary-vs-value question belongs to #440 and I deliberately did not expand into it.

Walked as a victim would, real `OperatorApp`, real key presses, scratch config seeded through the page's own writer with the exact repr string the shipped bug stores:

```
seeded stored value: type=str "{'default': ['anthropic/claude-opus-5', 'openrouter/deepseek']}x"

A. page open on the corrupt row
      ›  Failover cascade —
      malformed cascade — press r to clear it
      + add a chain
      read_chains -> {}

B. real 'r' press
      ›  Failover cascade 0 chains
      no cascade configured
      + add a chain
      cleared a malformed cascade
      stored -> '<absent>'   notice -> 'cleared a malformed cascade'

C. real 'enter' after the clear (travels into the group)
      cursor kind='chain_add'  editing=None  notice=''
      detail -> 'enter, then type: <chain-key> <provider>/<model>'

D. rebuild a chain from the recovered state ('+ add a chain')
      stored type=dict  read_chains -> {'rescue': ['anthropic/claude-opus-5']}
```

The row and the line under it now agree in every state, `r` clears the key, and the user lands back on an ordinary empty cascade they can rebuild.

**On the code reviewer's opposing finding.** Their N1/recovery note reached the other conclusion — that because a corrupted config renders without crashing and `+ add a chain` restores a mapping, "no dedicated recovery path is warranted". They verified that correctly and it is true as far as it goes. I implemented your finding instead, and the manager sided with you explicitly, for the reason your write-up gives and theirs does not weigh: the user is not merely inconvenienced, they are **misdirected**. `r` is lit in the footer on that row, it is the page's documented mitigation for immediate-write having no undo, and it answered with advice naming a row that is not painted. "Not stuck" is only true for a user who knows to ignore the value column, ignore `r`, and trust an `+ add` row that appears to add to a cascade the page just said does not exist. That is three pieces of knowledge a victim does not have, and the fix is about fifteen lines.

#### U2 — MINOR — nothing tells a victim anything is wrong → **TAKEN** (`447da9b2`)

Taken, and kept to one vocabulary as you asked rather than adding a third. The signal is the empty-state line itself, which changes from `no cascade configured` to `malformed cascade — press r to clear it` — so the line that was previously *contributing* to the contradiction is now the thing that resolves it, and it names the key that acts.

One predicate, `_cascade_is_malformed`, is shared by the row painter, the group line and `action_reset`. Three copies of that test would have been three chances for the frame to contradict itself again, which is the defect this round is about.

#### U3 — NIT — `enter` lands on the first chain, not the row's own affordance → **DEFERRED to #440** ([comment](https://github.com/damianvtran/local-operator/issues/440#issuecomment-5469705665))

Deferred rather than fixed, and I think genuinely owned by the redesign. The cascade row has no affordance of its own to activate — that is precisely why the fix travels into the group — so giving it one is an edit-model decision, not a bug fix: under the spec's `enter` opens / `enter` accepts / `esc` cancels contract, this is the one row whose "open" has two defensible meanings (open the group, or open the first chain in one press). Settling that here would also mean writing it against the activate path #440 rewrites. The issue comment adds the two-press case to §2.3's transition table, plus a third row for the malformed state this PR introduces, so the redesign cannot re-strand these users.

#### U4 — NIT — base two commits behind → **FIXED** (rebase)

Rebased twice, in fact: `origin/main` moved again mid-remediation (#448 landed and took `0.43.13`). Now based on `65ca130a` and bumped to **`0.43.14`** — `0.43.12` is tagged and live on PyPI, and `0.43.13` is both on `main` and claimed by #447. Detail in the code-round remediation under B1.

---

### Evidence

**Rendered frames, looked at rather than trusted.** Four new SVGs under `docs/evidence/settings-cascade-activate/`, both sides captured from the same base, plus a new `cascade-corrupt` state in `scripts/settings_shot.py`. The capture deliberately scrolls the setting row's whole GROUP into frame: the defect is a contradiction between the value column and the line under it, and my first capture attempt framed only the setting row, which would have photographed one half of it and proved nothing.

| | before (`000ba195`) | after |
|---|---|---|
| corrupt config | `Failover cascade {'default': ['anthropic/c…` over `no cascade configured` | `Failover cascade —` over `malformed cascade — press r to clear it` |
| after pressing `r` | value unchanged, notice `r resets one setting; delete a chain with d on its row` | `Failover cascade 0 chains` over `no cascade configured`, notice `cleared a malformed cascade` |

`qlmanage` does stall under this machine's load as warned, so the SVGs were served over `python3 -m http.server` and screenshotted through the browser tool, then read back as images. I also confirmed the HEALTHY row's rendering is byte-unchanged by recapturing `cascade-row` and diffing it against the committed frame: the only differences are the scratch tmpdir path and a status-band race.

**Regression test, proven failing at the unremediated head.** `test_r_clears_a_malformed_cascade_and_the_frame_does_not_contradict_itself` builds the corrupt config through `settings_io.coerce` + `write_setting` — the exact pair `_commit_edit` calls, so it pins a state the page itself can produce rather than one hand-written into YAML — then asserts on the exported frame AND on `config.yml`. Run in a throwaway worktree at `000ba195` with only the test file copied in (`git diff --stat` showed the one file):

```
FAILED test_r_clears_a_malformed_cascade_and_the_frame_does_not_contradict_itself
  AssertionError: the value column still paints the corrupt Python repr as if it
  were the setting's value
FAILED test_enter_into_the_cascade_does_not_carry_the_previous_row_notice
  AssertionError: the notice from `r` on the cascade row followed the cursor onto
  the chain row: 'r resets one setting; delete a chain with d on its row'
```

A companion test asserts `r` on a HEALTHY cascade still explains itself and writes nothing, compared on the config **bytes** rather than on the parsed chains — "the chains are still there" would also pass if `r` had rewritten the file with identical content.

**One thing I found while writing that test, fixed separately** (`d15ebb96`): `_painted_frame` returns the SVG's text runs, and the exporter separates words with U+00A0. Every multi-word phrase a caller searched for was therefore absent by construction — `"no cascade configured" not in frame` was **True of a frame containing that exact line**. The helper's existing negative assertions could not fail and my new positive one could not pass. Folded in the helper, since a call site cannot see the trap.

### Gates (whole tree, as CI runs them, on the rebased head)

| gate | result |
|---|---|
| `flake8 .` | clean |
| `black --check .` (26.1.0) | 626 files unchanged |
| `tests/unit/tui/test_settings_view.py` + `tests/unit/test_settings_io.py` | **464 passed** |
| full unit suite (`-n 4`) | **8992 passed, 15 skipped, 0 failed** |

The full suite came back completely green this run — no load-induced failures at all, cleaner than either round-1 reviewer got. Run at `-n 4` rather than the default `-n auto`, on instruction: this machine has had concurrent sessions spawning 14 xdist workers each and hitting an out-of-memory condition.

Ready for re-review against `9c41b01d`.

damianvtran:
**Version re-bumped: `0.43.15` — head is now `be1ad7e8`.**

B1 recurred a third time while the round-1 remediation was in review. `origin/main` released twice underneath this branch: #448 took `0.43.13` and #457 took `0.43.14`, both now tagged with published Releases (PyPI serves `0.43.13`). The remediation comments above were written against `9c41b01d` at `0.43.14`, which is no longer free.

Rebased onto current `origin/main` (`ba414381`) and bumped to `0.43.15`, verified free on main, in the tag list, and on PyPI. The rebase **dropped the bump commit again** as "already upstream" — that is now twice on this branch, and it is why the version is re-checked against `git show origin/main:pyproject.toml` after every rebase rather than assumed to have survived.

Everything else in the two remediation comments stands unchanged: the source and test trees are byte-identical to the state those gates were run against (verified by diff), only `pyproject.toml` moved. Re-confirmed on this head: flake8 clean, black 634 files unchanged, isort clean, `test_settings_view.py` **371 passed**.

**Heads-up for #447:** it is still open carrying `0.43.13`, which is now released. It needs its own re-bump before merge.

PR #451, opened 2026-08-30
fix(tui): keep a live prompt and the composer on one width, and stop the toast twitching

### LINKED ISSUE #168: boot layout clamps the composer but not a live prompt, so the question overhangs it
## Summary

Pressing <kbd>Ctrl</kbd>+<kbd>L</kbd> (`action_clear_transcript`) while an `ask` picker or approval prompt is live restores the **boot layout**, which clamps `#input-shell` to a narrow centred card (`Screen.boot.boot-card #input-shell { width: 70%; min-width: 75; max-width: 100 }`) — but `#prompt-host` and the prompt card itself are **not** clamped with it. The question then overhangs the composer it is docked to.

Found during agent review of #162 (finding F3), by an independent reviewer subagent driving the real app.

## Reproduction

Raise an `ask` prompt on a 160-column terminal, then press <kbd>Ctrl</kbd>+<kbd>L</kbd>:

```
before ^L: boot=False card=156 shell=156
after  ^L: boot=True boot-card=True card=156 shell=98 widest_ink=156 OVERHANG=58
```

The card's slab and its ink are both 156 cells while the composer beneath is 98 — the question hangs 58 cells past the input it belongs to.

## This predates #162

`AskPickerScreen` has been `width: 1fr` since #138 moved the picker into the dock, so the **slab** has overhung in this state since then. Measured on clean `origin/main` (f87f778):

```
after  ^L: boot=True boot-card=True card=156 shell=98 widest_ink=74 OVERHANG=-24
```

The slab is identically 156 against a 98-cell shell; the old 74-cell text cap merely kept the *ink* inside 98 and hid the geometry. #162 removed that cap (correctly — it was the bug being fixed), which makes the pre-existing mismatch visible.

So this is a **#138 layout bug that #162 exposes**, not one it introduces. It was explicitly judged non-gating for #162 by the review round.

## Suggested fix

One owner for the clamp. Either:

- clamp `#prompt-host` alongside `#input-shell` in the boot-card rule, so the prompt tracks the composer's card width; or
- exempt `#input-shell` from the boot clamp while a prompt is mounted, so the dock stays full width for as long as something is being asked — arguably better, since a question the turn is parked on is not part of the empty-state splash composition.

The second also removes the odd state where a *splash* (the empty-state welcome) is showing above a live question about work that already happened.

## Notes

Reachable through any path that calls `_set_welcome_visible(True)` with a prompt mounted — <kbd>Ctrl</kbd>+<kbd>L</kbd> is the one confirmed by driving it.

### LINKED ISSUE #170: toast card re-widths between consecutive receipts
Split out of design review round 3 (D12) on #167 as a follow-up.

## What happens

`Toast` is right-anchored and `_refit`s to its text, so a change in digit count moves the card's left edge by a cell:

```
'copied 8 characters'   -> x=58 width=21
'copied 60 characters'  -> x=57 width=22
```

Between two consecutive copy receipts the card twitches one cell horizontally.

## Why it is newly visible

The behaviour is pre-existing `_refit` sizing, not introduced by #167. But before that PR the toast was a startup-scale surface: consecutive cards were rare, and `copied N lines` was a stable width in practice. #167 made a routine editing gesture (a drag over the composer) write the slot, and character counts cross 9→10 and 99→100 routinely — so successive receipts now change width where successive toasts previously did not.

## Options

1. **A minimum width for the card**, so short receipts share one edge. Cheapest, and the reserve already exists in the width clamp (`toast_max_width`).
2. **Size to the widest recent content** and shrink only after a quiet period. More faithful to the text, more state.
3. **Left-align the number** inside a fixed-width field, so digits grow into padding rather than moving the border.
4. **Accept it.** One cell, on a card that is up for five seconds.

Option 1 is probably right, but it interacts with the narrow-terminal clamp (`test_the_card_stays_inside_a_terminal_narrower_than_the_floor` pins a real bug where the floor beat the box at 20 cells), so whatever floor is chosen has to stay subordinate to the screen box.

## Acceptance

- Two consecutive copy receipts of different digit counts do not move the card's left edge, **or** the twitch is documented as accepted.
- The existing width guarantees hold: nothing paints outside the screen's edge padding, and the 16/20/21-cell cases in `tests/unit/tui/test_toast.py` still pass.

damianvtran:
# PR Description

## Summary of Changes

Two layout defects, independent but the same shape: a card deciding its own
width without reference to the thing it has to line up with.

### #168 — the boot clamp reached the composer but not the prompt above it

`Screen.boot.boot-card #input-shell` clamps the composer to a narrow centred
card while the session is empty. `#prompt-host` and the prompt card are
`width: 1fr` and were never clamped with it, so any path that restores the
splash with a prompt mounted — `Ctrl+L` is the confirmed one — left the
question overhanging the input it is docked to:

```
before ^L: boot=False boot-card=False card=156 shell=156  SLAB_OVERHANG=0
after  ^L: boot=True  boot-card=True  card=156 shell=98   SLAB_OVERHANG=58
```

**The clamp is withheld while a prompt is live, rather than extended to cover
the prompt.** The issue offered both and preferred this one; the codebase
argues for it more strongly than the issue does. The tcss note on
`AskPickerScreen` records that its `width: 1fr` is "the ONLY statement of how
wide the question is", written after a second, older width statement capped
the ink a third short of the panel's own fill on a wide terminal. Clamping
`#prompt-host` would reinstate exactly that second owner. Standing the card
down keeps one owner for the width and hands the dock the same full-width
geometry it gets the moment the first prompt is submitted.

It stands down the whole composition rather than just the clamp, deliberately:
boot notices centre on the card's axis (`_sync_boot_column_width`), so
releasing the clamp while leaving the class on would centre a notice against a
card that is no longer there.

The condition is re-derived from the host on every pass (`_prompt_is_live`),
never tracked as a flag — the same discipline `_sync_boot_layout_class`
already states — so it cannot latch and lose the boot composition to the first
`ask` of a session. `display` is the term it reads, not the child count: the
host is hidden while it holds a prompt too short to draw itself, and a question
painting nothing must not go on suppressing the card.

### #170 — the toast re-widthed between consecutive receipts

`Toast` is right-anchored and `_refit`s to its text, so a digit crossing 9→10
moved the card's left border one cell between two receipts seconds apart:

```
'copied 8 characters'   -> x=58 width=21
'copied 60 characters'  -> x=57 width=22
```

A **24-cell floor**, the issue's preferred option. 24 is measured from the
receipt family rather than picked for roundness: the widest routine receipt is
`copied 6000 characters` at 24 cells and everything else is narrower, so one
floor holds them all to a single edge.

**The floor stays subordinate to the screen box**, which was the stated
constraint. It is applied *inside* the `toast_max_width` clamp, and mirrored
onto the widget's `min_width` bounded by that same cap — `min_width` beats
`max_width` in Textual, so a raw floor would paint straight past the screen's
edge padding, which is the bug
`test_the_card_stays_inside_a_terminal_narrower_than_the_floor` already pins.
Below 26 cells the floor simply does not apply. The widget needs it as well as
`_card_width` because the card is `width: auto`: fixing only the offset would
have widened the gap on the left while the right edge crept in — the same
twitch mirrored.

## Testing Evidence

Both defects were reproduced on `origin/main` first, with the geometry the
issues quote, then re-measured after.

### #168 — measured, at four widths

Driving the real `OperatorApp` through a real `request_user_choice`, pressing
`ctrl+l`, and reading the composited frame:

```
BEFORE (origin/main), 160x40
  before ^L: boot=False boot-card=False card=156 shell=156 SLAB_OVERHANG=0
  after  ^L: boot=True  boot-card=True  card=156 shell=98  SLAB_OVERHANG=58   <-- the defect

AFTER, 160x40
  after  ^L: boot=True  boot-card=False card=156 shell=156 SLAB_OVERHANG=0

AFTER, other widths (slab overhang after ^L)
  100x30  card=96  shell=96   OVERHANG=0
  120x36  card=116 shell=116  OVERHANG=0
  200x48  card=196 shell=196  OVERHANG=0
```

**The after-state is not a new composition.** Independently established in
design review: at the head, the post-`ctrl+L` frame is *pixel-identical* to the
splash-with-a-prompt state the app already shows when a question arrives early
in a session with no `ctrl+L` involved. On `origin/main` those two states
differ, which is the tell that the old post-`ctrl+L` frame was the anomaly.
`ctrl+L` with a live prompt now lands on a composition the app already has, so
there is no centre for it to lose — a stronger claim than "the overhang is
gone".

The stand-down is temporary, which is the risk it carries. The full cycle:

```
                                   boot  boot-card  shell
1 splash, no prompt                  1        1        98
2 prompt raised on the splash        1        0       156
3 after ctrl+L (the issue)           1        0       156
4 question answered                  1        1        98   <-- composition restored
```

Same at 100x30 (73 → 96 → 96 → 73). At 80x24 the card is already off by width,
and the fix changes nothing there — the honest degradation is preserved.

### #170 — measured, whole receipt family

Driven through the real clipboard path (`_put_on_clipboard`), so the wording
and the courtesy flag are the shipped ones:

```
BEFORE (origin/main), 80 cols
  copy  8 chars -> 'copied 8 characters'   x=58 w=21
  copy 60 chars -> 'copied 60 characters'  x=57 w=22      <-- edge moved

AFTER, 80 cols                              AFTER, 160 cols
  copied 1 character      x=55 w=24           x=135 w=24
  copied 8 characters     x=55 w=24           x=135 w=24
  copied 60 characters    x=55 w=24           x=135 w=24
  copied 600 characters   x=55 w=24           x=135 w=24
  copied 6000 characters  x=55 w=24           x=135 w=24
  copied 2 lines          x=55 w=24           x=135 w=24
  copied 12 lines         x=55 w=24           x=135 w=24
  copied 120 lines        x=55 w=24           x=135 w=24

distinct left edges: [55]                   distinct left edges: [135]
```

On `origin/main` the same walk reports **7 distinct left edges**.

The constraint, walked across every width through the floor's threshold, three
messages each, asserting nothing paints outside the screen's edge padding:

```
term= 14 ... term= 22   x=1, right=width-1        inside_box=OK
term= 23 x=2  w=20      term= 26 x=5  w=20        inside_box=OK
term= 27 x=5  w=21  ... term= 33 x=5  w=27        inside_box=OK
term= 80 x=46 w=33      term=200 x=166 w=33       inside_box=OK

violations: 0
```

### Rendered frames

Both are visual, so the frames were exported with `save_screenshot` and
**looked at**, not just asserted.

**#168 before** — the question's slab runs to the right edge while the composer
below stops 58 cells short, leaving a dark notch under the card's right half,
with a welcome splash above a live question about work that already happened:

boot clamp overhang, before

**#168 after** — question and composer share both edges and read as one dock:

boot clamp overhang, after

**#170 before/after** — consecutive receipts, the before pair showing the
left border sitting one cell further left on the wider number, the after pair
sharing one edge with the digit growing into padding instead:

toast card edge, before and after

### Regression tests

Each was confirmed to **fail on `origin/main`** and pass here — a test that
passes either way proves nothing.

```
tests/unit/tui/test_boot_layout.py
  test_a_live_prompt_is_never_wider_than_the_composer_it_is_docked_to[160x40, 120x36, 100x30]
  test_the_boot_card_comes_back_once_the_question_is_answered

  on origin/main: 4 failed
    assert 159 == 101   (card.right vs shell.right, 160 cols)
    assert 119 ==  83   (120 cols)
    assert  99 ==  76   (100 cols)
    AssertionError: a live question stands it down

tests/unit/tui/test_toast.py
  test_consecutive_receipts_of_different_digit_counts_share_one_edge
  test_a_message_past_the_floor_still_sizes_to_its_own_text
  test_the_floor_never_beats_the_screen_box[16, 20, 21, 22, 23, 24, 25, 26]

  on origin/main: 1 failed
    AssertionError: [(55,79),(56,79),(57,79),(58,79),(59,79),(61,79), ...] assert 7 == 1
```

The first asserts the card against the **composer** rather than a width
constant, so it pins the property (they share an edge) and not today's clamp
arithmetic.

### Gates

```
black --check .                           618 files unchanged
pyright .                                 0 errors, 0 warnings, 0 informations
pytest tests/unit                         8808 passed, 15 skipped
```

The existing 16/20/21-cell toast cases and the whole boot geometry suite pass
unchanged (159 passed across `test_toast.py`, `test_boot_layout.py`,
`test_ask_picker.py`, `test_approvals_ux.py`).

One flake, `test_steering_approval.py::test_an_empty_message_end_keeps_the_streamed_prose`,
passes in isolation (114 passed) and that file also flakes on clean
`origin/main` — as does `test_transcript_selection.py`, which failed with a
different test on each of two consecutive base runs.

## Risk

Both changes are narrow, and the second-order risk in each is the one covered
above.

For #168 the risk is a latched stand-down leaving the boot composition
permanently full-width. The condition is recomputed from the host on every
sync and reconciled at all three points that write the host's `display`
(mount, drawability change, unmount), and the cycle test pins the restoration.
Reconciling at mount rather than only on the deferred refresh is deliberate:
one refresh of delay is one painted frame of the artifact being fixed.

For #170 the risk is a floor beating the screen box on a narrow terminal,
which is a real bug this repo has had once. It is bounded by the cap in both
places it is written, and the parametrised test walks every width from 16 to
26.

A note for the reviewer: issue numbers cannot be written as `#NNN` in
`local_operator.tcss` — the minimalism suite reads that shape as a literal hex
colour and rejects the sheet — so the tcss comment refers to the defect in
prose and the numeric references live on the Python side.

Closes #168
Closes #170


damianvtran:
### Design review — round 1
Reviewer: designer (claude-opus-4-6)
Scope: 3c79116c..e56df3f8

Captured my own frames in an independent worktree off `origin/main` (`/tmp/rev451-designer`) and at the PR head (`/tmp/rev451-designer-head`), using the author's `boot_shot.py` / `toast_shot.py` plus three scripts of my own for the states those two do not reach: the splash with no prompt, the prompt-arrives / prompt-dismisses transitions frame by frame, ink position inside the toast slab, and a walk of the receipt family across terminal widths. Every SVG was rendered to PNG and looked at. Geometry below is re-measured, not taken from the PR body.

Both defects reproduce exactly as the issues quote them, and both are fixed.

---

#### #168 — the overhang

Confirmed on pristine `origin/main` at 160x40: `after ^L: boot-card=True card=156 shell=98 SLAB_OVERHANG=58`, and in the frame it is worse than the number suggests — the question's slab runs the full width while the composer stops 58 cells short, leaving a dark notch under the card's right half with no edge to explain it. At the PR head the same keystroke gives `card=156 shell=156 SLAB_OVERHANG=0`, at 160/120/100 alike. Question and composer share both edges and read as one dock.

**On the composition question you raised (does it read as deliberate, or as a layout that lost its centre?):** it reads as deliberate, and the reason is that the after-state is not a new composition at all. I captured the splash-with-prompt state without any `ctrl+L` in it — the state a user reaches by simply being asked something early in a session — and at the head the post-`ctrl+L` frame is **pixel-identical** to it. `ctrl+L` with a live prompt now lands on a composition the app already shows on its own. On `origin/main` those two states differ, which is the tell that the old post-`ctrl+L` frame was the anomaly. The full-width dock under a splash is the app's existing answer to "something is being asked", not something this PR invented, so there is no centre to lose.

**D1 — MINOR — the two notices now sit on conflicting axes, and at 160 columns it is a visible misalignment.**
This is the call the author flagged. Measuring the notice rows in the composited frame at 160x40:

```
origin/main (after ^L)   /resume …            x_px=671    ] one axis,
                         transcript cleared … x_px=732    ] both card-centred

PR head     (after ^L)   /resume …            x_px=671    <- still card-centred
                         transcript cleared … x_px= 49    <- now on the full-width spine
```

Two dot-led notices in the same typographic family, four rows apart, on axes 622px apart. The provenance claim checks out — I reproduced flush-left `transcript cleared` on pristine `origin/main` at 80 and 76 columns, so this is genuinely the base behaviour whenever the card is off — but as you say, that is an explanation and not a defence, so judging the result on its own: it is a real misalignment and it is **not** a blocker. What makes it minor rather than major is that `/resume` is a *splash* line and `transcript cleared` is a *transcript* line, and in the after-frame the second one now sits directly above the prompt card's left edge, which is where transcript content belongs. The eye reads two groups, not one broken column. At 100 columns and below only one notice is present and the question does not arise.

Worth noting the PR makes this *better* in one respect I did not expect: on `origin/main` at 100x30 the cleared notice sits at x_px=378 (card axis) while the dock beneath it is full width, so the notice was floating over nothing. At the head it is at x_px=73, aligned with the card below it.

Not worth a remediation round on its own. If it is ever picked up, the fix is for the cleared-transcript notice to take the transcript's left margin in both states rather than switching axis with the boot card — i.e. make it consistently a transcript line, which is what it already is at every width except this one.

**D2 — NIT — one frame of `connecting…` on the prompt-arrival transition.**
Diffing consecutive frames at the head, `C-prompt-frame0` differs from `frame1`, and on `origin/main` it does not. The entire delta is the status text: `connecting…` → `test/model`. That is the session's own connect settling on its normal schedule, not layout — the geometry is identical across those two frames (`shell.x=1 shell.w=158 card.w=156` in both), so nothing moves. The prompt-dismisses transition is single-frame in both trees, and both the stand-down and the restore land in one paint with no intermediate width. **There is no reflow in either direction** — which is the thing that mattered here, since a stand-down that resized in two steps would have been a MAJOR.

---

#### #170 — the toast twitch

Reproduced the twitch exactly on `origin/main` at 80 cols: `copied 8 characters` x=58 w=21, then `copied 60 characters` x=57 w=22 — the left border moves one cell between two receipts seconds apart. At the head both sit at x=55 w=24. Walking the whole receipt family (1/8/60/600/6000 characters, 2/12 lines) I get **7 distinct left edges on base, 1 at the head**.

I also checked the thing the issue warned about — that the floor must stay subordinate to the screen clamp — by walking the family at every width rather than trusting the parametrised test. **One shared edge at 20, 24, 26, 28, 30, 40, 80 and 160 cells, nothing painting outside the box at any of them.** Below the floor the clamp takes over and the family collapses to one edge anyway because every message is truncated to the same width, so there is no width at which the twitch survives. The cure is complete, which is more than option 1 strictly promised.

**D3 — MINOR — the floor's slack is all on the right, so short receipts read as left-aligned in an oversized card rather than as a deliberately sized one.**
Measuring ink against the slab at 80 cols:

```
                        card  text  leftpad  rightpad
copied 6000 characters   24    22      1        1     balanced
copied 600 characters    24    21      1        2
copied 60 characters     24    20      1        3
copied 8 characters      24    19      1        4
copied 2 lines           24    14      1        9     <- worst case
```

On base every one of these was `1/1`. So the floor buys a stable edge and pays for it with up to 9 cells of dead space on one side, and because the padding is asymmetric the card stops looking sized-to-content and starts looking like text that failed to fill its box. In the rendered frame `copied 8 characters` is fine — 4 cells of trailing slack on a five-second card is unremarkable, and the shared right edge with the transcript's right margin gives it something to line up against. `copied 2 lines` is the one that reads thin.

Judging it against the issue's option 4 (accept the twitch): **the cure is better than the disease.** A border that moves under a stationary eye is a defect the user notices without being able to name; trailing padding on a transient card is not. I would not revert to accepting the twitch.

But the issue's option 3 — left-align the number inside a fixed field, so digits grow into padding rather than moving the border — would have got the stable edge *and* balanced padding, and it is worth recording as the better treatment. Concretely: centre the text within the floored card (`text-align: center` on `Toast`, or splitting the slack evenly in `_refit`) turns `copied 2 lines` from `1/9` into `5/5` and costs nothing at the clamped widths, since below the floor the text fills the card anyway. That is a one-line change and it removes the only visual cost this fix carries. Not gating.

**D4 — NIT — 24 is measured from `copied 6000 characters`, which makes the common case the loosest.**
The floor is set by the widest routine receipt, so the receipts users see most often (`copied N lines`, small character counts) sit furthest from the edges. A floor of 20 or 21 would still collapse the 9→10 and 99→100 crossings that #170 is actually about, while halving the slack on the short messages. But this trades one stable edge for two edge groups, and the current choice is the one that satisfies the issue's acceptance criterion literally. Recording it as a rationale note, not a change request — the comment at `TOAST_MIN_CARD_WIDTH` already explains the derivation clearly, which is what I would want if I hit this in six months.

---

#### Narrow widths and hierarchy

At 80 and 76 columns the boot card is already off by width, and I confirmed the frames are byte-identical between base and head — the honest degradation is preserved and #168's fix correctly does nothing there. At 76 (just above the 75-cell `min-width` boundary) the composition is unchanged. Spacing rhythm, the divider under `the agent needs your decision`, the option indent, the recommended-row highlight and the hint row are untouched in every state I captured; the changed states follow the same conventions as the unchanged ones. Focus order and copy are unaffected by either change.

---

#### Verdict

**No BLOCKER, no MAJOR. This round is TERMINAL.**

Both defects are fixed, both fixes are complete across the width range rather than at the measured points only, and neither introduces motion the user sees. The after-frames read better than the before-frames in both cases, and #168's after-state is the composition the app already shows on its own rather than a new one.

4 findings: 0 blocker, 0 major, 2 minor (D1, D3), 2 nit (D2, D4).

Follow-ups, none gating this merge:
- **D1** — cleared-transcript notice switches axis with the boot card; pre-existing base behaviour, made visible at 160 columns. Best fixed by giving it the transcript's left margin in both states.
- **D3** — centre the receipt text within the floored card so the slack is balanced rather than all trailing; removes the only visual cost the width floor carries.

damianvtran:
### Agent review — round 1
Reviewer: reviewer (claude-opus-4-6)
Scope: 3c79116c3adc02a4788b68019ac7950be999df9c..e56df3f8ce903a9c08e582787fa2123d63677ea8

Reviewed independently of the author. Two read-only worktrees of my own (`/tmp/rev451-reviewer` at pristine `origin/main`, `/tmp/rev451-head` at the PR head), every number below re-measured rather than taken from the PR body, and the SVG frames rendered to PNG and actually looked at.

**Both defects reproduce exactly as the issues quote them, and both are fixed.**

```
#168  base @160x40  after ^L: boot-card=True  card=156 shell=98  SLAB_OVERHANG=58
      head @160x40  after ^L: boot-card=False card=156 shell=156 SLAB_OVERHANG=0

#170  base @80x24   'copied 8 characters'  x=58 w=21  |  'copied 60 characters' x=57 w=22
      head @80x24   'copied 8 characters'  x=55 w=24  |  'copied 60 characters' x=55 w=24
```

In the rendered base frame the overhang is plainly visible — the question's slab runs the full width while the composer stops 58 cells short. At the head both edges line up. The toast pair shows the left border moving on base and holding on head.

---

#### The design choice (#168): standing the card down was the right call

I read both CSS comments before judging, and the argument holds. The `AskPickerScreen` rule at `local_operator.tcss:1783-1797` is explicit that `width: 1fr` is "the ONLY statement of how wide the question is", and it records the exact regression that produced it: a second width statement (`_card_width`'s 74-cell modal-era cap) left the ink a third short of the panel's own fill while the composer beneath ran full width. Clamping `#prompt-host` in the boot-card rule would reinstate a second owner of that width — the same shape as the bug that note exists to prevent, one level up.

The stand-down also gets the composition right rather than merely the arithmetic. Standing down the whole class instead of just releasing the clamp is correct because `_sync_boot_column_width` centres boot notices on the card's axis; releasing the clamp alone would centre notices against a card that is no longer there. I verified the resulting state is not a new composition at all: the post-`ctrl+L` frame at the head is the same layout the app already shows when a prompt arrives without any `ctrl+L`. No finding here.

#### 1. The toast floor stays subordinate to the screen box — verified

I tested the floor/clamp interaction directly rather than trusting the suite. Every width 14..29, five messages each (both receipt digit counts, a 1-char message, the widest routine receipt, and one long enough to truncate), asserting four separate properties: width never exceeds `toast_max_width`, never exceeds `TOAST_MAX_WIDTH`, `right` never passes the screen's edge padding, and `x` never goes negative. **Zero violations.** The 16/20/21-cell cases hold (`w=14/18/19`, `right=15/19/20`, all inside the box), and the floor correctly stops applying below 26 cells. The `min(TOAST_MIN_CARD_WIDTH, cap)` on `min_width` and the `min(cap, max(...))` in `_card_width` are both correctly ordered.

#### 2 & 3. Boot-card invariants and the state machine — verified

I drove every path listed in the brief plus several more, checking the class, the shell width, host `display` and child count at each step. All restore correctly: abort with `esc`; two prompts in sequence; a prompt arriving while the card is already down; resize down and back up while live; `ctrl+L` twice under one prompt; ask+approval overlap answering only one; shrink to an undrawable prompt and grow back (`card=True` while undrawable, correctly back to `False` when drawable again — the `display`-not-child-count choice in `_prompt_is_live` is what makes this right); transcript growing while a prompt is live (splash retires, card stays down, correct). The `BOOT_CARD_MIN_INSET >= 8` invariant is untouched — the new condition is `and`-ed in front of it, never replacing it. I found **no latch in either direction** and no path where the card only restores on the happy path.

Reconciling at all three points that write `host.display` (mount, drawability change, unmount) rather than only on the deferred `_sync_prompt_host` is the right call, and the mount-time comment gives the correct reason: one refresh of delay is one painted frame of the artifact.

#### 4. The pre-existing-behaviour claim — verified independently

Measured on pristine `origin/main`, the `transcript cleared` notice with the card off:

```
base @80x24   card=False  notice x=2 w=76 col_class=False   <- flush left
head @80x24   card=False  notice x=2 w=76 col_class=False   <- identical
```

Identical on both trees. The flush-left placement is genuinely base behaviour whenever the card is off, not a regression this PR introduces. At 160 with a live prompt the head reaches that same pre-existing state by a new route, which is a consequence of the stand-down and not a new behaviour. Claim stands. (The designer round already logged the axis question as D1/MINOR; I agree with that severity and do not duplicate it.)

#### 5. Gates and CI

Run over the whole tree in my own worktree: `flake8` clean, `black --check` (26.1.0) clean, `isort` (5.13.2, repo config) clean, `pyright` **0 errors**. Full TUI suite `3680 passed, 12 skipped` in 452s. The four known-flaky tests named in the brief all passed here. CI: lint, type-check, test 3.12, test 3.13, pip-audit, cli-sanity and tui-e2e (macos) all green; the remaining jobs were still queued at review time.

**Test quality — they pin behaviour, not implementation.** I checked out the new test files against pristine `origin/main` source: `test_a_live_prompt_is_never_wider_than_the_composer_it_is_docked_to` fails at both sizes with the real overhang numbers, and `test_the_boot_card_comes_back_once_the_question_is_answered` fails on "a live question stands it down". Asserting `card.region.right == shell.region.right` against the composer rather than against a width constant is the right shape.

---

### Findings

**F1 — MINOR — the `min_width` clamp is untested; the mutant survives the suite.**
`toast.py:486` clamps the floor with `min(TOAST_MIN_CARD_WIDTH, cap)`, and the comment names this as the specific hazard. But replacing it with the raw `TOAST_MIN_CARD_WIDTH` leaves the entire toast suite green — **50 passed**, including `test_the_floor_never_beats_the_screen_box` at all eight parametrised widths. I re-measured the mutant directly and the geometry is unchanged at every width 14..29, which explains why: in Textual 8.2.8, `max_width` actually beats `min_width`, so the clamp is currently belt-and-braces rather than the thing holding the card inside the box. The real load-bearing clamp is the `min(cap, ...)` inside `_card_width`.

This is MINOR, not MAJOR: the behaviour is correct today, the clamp is harmless, and keeping it is the right defensive call since the precedence is a Textual implementation detail that could change. The finding is that the comment states the precedence as fact ("`min_width` beats `max_width` in Textual") when the opposite is observably true in the pinned version, and that a test named for the constraint does not actually exercise it. One line: soften the comment to say the clamp does not *depend* on the precedence, and if the constraint is worth a test, assert `styles.min_width` directly rather than only the rendered region.

**F2 — NIT — `_prompt_is_live` catches bare `Exception` where the failure is a known query miss.**
`app.py:5653-5656`. Consistent with the three sibling `query_one("#prompt-host")` call sites in this file, so I would not change it in isolation — noting it only because the `pragma: no cover` says the intended case is "before the dock is composed", which is `NoMatches` specifically. A broad catch here silently returns "no prompt live" for any error, and the consequence is a boot card that comes back under a live question. Not worth a round.

**F3 — NIT — `TOAST_MIN_CARD_WIDTH = 24` is justified by `copied 6000 characters` at 24 cells, but that is the text width, not the card width.**
The docstring says the widest routine receipt "is `copied 6000 characters` at 24 cells", and the floor is 24. With `TOAST_PADDING_CELLS = 2` that message actually needs a 24-cell *card* for 22 cells of text, so the floor and the widest member coincide exactly rather than the floor covering it with room to spare. The arithmetic works out correct either way and I verified the whole receipt family shares one edge at 80 cols, so this is a wording imprecision in the comment, not a defect in the number.

---

### Verdict: **clean**

No BLOCKER and no MAJOR findings. Both issues reproduce on base and are fixed at the head, verified by re-measured geometry and by looking at rendered frames; the toast floor is provably subordinate to the screen box at every width from 14 to 29; the boot-card state machine restores on every path I could reach including the asymmetric ones; the pre-existing-behaviour claim is independently confirmed; all gates and the full TUI suite are green; and the new tests fail against base source, so they pin the fix rather than restating it.

**This round is TERMINAL.** F1, F2 and F3 are follow-ups, not a reason for another round.

damianvtran:

Implementer: coder (claude-opus-4-6)
Answering both rounds: [code round](https://github.com/damianvtran/local-operator/pull/451#issuecomment-5469533837) (F1–F3) and [design round](https://github.com/damianvtran/local-operator/pull/451#issuecomment-5469442815) (D1–D4).

Remediation commit: `77428f99`. The branch has also been **rebased onto current `origin/main` (`65ca130a`)** so this round is fresh against what ships; the original fix commit is unchanged in content (only hunk offsets moved).

---

#### F1 — MINOR — the `min_width` clamp is untested; the mutant survives — **FIXED** (`77428f99`)

You were right on both halves, and the precedence claim in my comment was simply wrong. I measured it rather than take either version on faith:

```
Textual 8.2.8, width:auto widget
  min_width= 30 max_width= 10 -> rendered width=10
  min_width= 24 max_width= 18 -> rendered width=18   <- max_width wins
  min_width= 24 max_width= 40 -> rendered width=24
```

So `max_width` beats `min_width`, the opposite of what the comment asserted as fact. The comment now states the measured precedence, names `_card_width`'s `min(cap, ...)` as the actually load-bearing clamp, and explains why the `min(TOAST_MIN_CARD_WIDTH, cap)` stays anyway: the precedence is an undocumented engine detail, and if it ever inverts an unclamped floor paints past the screen's edge padding below 26 cells.

On the test: confirmed your mutant survived. `test_the_floor_never_asks_for_more_than_the_box` asserts the **resolved `styles.min_width`** instead of the rendered region, which is the only place the difference is observable while the engine masks it. Against your exact mutant (`self.styles.min_width = TOAST_MIN_CARD_WIDTH`):

```
before this test:  50 passed          <- mutant survives the whole suite
after  this test:   8 failed, 52 passed
  test_the_floor_never_asks_for_more_than_the_box[16] AssertionError: (16, 24.0, 14)
  ...[20] (20, 24.0, 18)  ...[25] (25, 24.0, 20)  ...[26] (26, 24.0, 20)
```

Code restored and green afterwards (62 passed).

#### F2 — NIT — bare `Exception` in `_prompt_is_live` — **REJECTED** (reason below)

Rejecting on the grounds you gave for not pressing it. All three sibling `query_one("#prompt-host")` call sites in this file (`app.py:7214`, `:7276`, `:7330`) catch bare `Exception`; narrowing mine alone would put a second convention beside an established one in the same handful of lines, which is the defect this codebase's own guidance calls out. It is worth doing as a sweep across all four, which is a change I should not make unrequested inside this PR's scope. Recorded for the manager rather than filed as an issue.

#### F3 — NIT — `TOAST_MIN_CARD_WIDTH`'s derivation conflates text and card width — **FIXED** (`77428f99`)

Correct: 24 is the CARD width for `copied 6000 characters`, whose text is 22 cells. Reworded to say the floor and its widest member coincide exactly once `TOAST_PADDING_CELLS` is added, rather than implying the floor clears it with room to spare.

---

#### D1 — MINOR — the two notices sit on axes 622px apart at 160 columns — **DEFERRED** (reason below)

Deferring, and taking your framing for why it is deferrable: `/resume` is a splash line and `transcript cleared` is a transcript line, the eye reads two groups rather than one broken column, and at 100 columns and below the question does not arise.

The fix you name — give the cleared-transcript notice the transcript's left margin in **both** states rather than switching axis with the boot card — is a change to `_sync_boot_column_width`'s notice-axis rule, which governs every boot notice (MCP failures included), not only this one. That is a composition change to a surface neither issue is about, and it would land unreviewed against a round that is already terminal. It belongs in its own change with its own frames.

Recording your finding that this PR **improves** the same alignment at 100 columns, which I had not measured: on `origin/main` the cleared notice sits at x_px=378 on the card axis while the dock beneath it is full width, so it floated over nothing; at the head it is at x_px=73, aligned with the card below it.

#### D2 — NIT — the one frame difference on prompt-arrival — **CONFIRMED, no change**

Verified independently across three runs, diffing consecutive composited frames row by row and separating "characters changed" from "ink extent moved":

```
transition 2->3: 1 rows changed, 0 with moved ink
    row 34: |◆ connecting… › ⌂ /private/tmp/…| -> |◆ test/model › ⌂ /private/tmp/…|
           ink (2, 157) -> (2, 157)
transition 5->6: 30 rows changed, 30 with moved ink   <- the prompt MOUNT itself
```

Exactly as you described: the sole non-mount transition is the status band's model label settling, with the row's ink extent byte-identical. No layout motion, nothing to fix.

#### D3 — MINOR — the floor's slack is all trailing — **FIXED** (`77428f99`)

Taken. It was very nearly the one-liner you described, with one wrinkle worth recording, so I measured both shapes before choosing.

A blanket `text-align: center` on `Toast` is **not** safe: the MCP startup summary is two lines, and centring each row against its own length staggers them —

```
text-align: center, two-line MCP summary
    lead=4  |   ⊙ MCP: 1 of 2 up|
    lead=7  |      failed: gh|      <- rows no longer share a left edge
```

which trades D3 for a worse defect on a card the user has to read. So the alignment is decided per message in `_refit` (centre only when the message has no newline), which is why it lives there rather than in the stylesheet. Result:

```
                        card  text  leftpad  rightpad
copied 2 lines           24    14      5        5      (was 1 / 9)
copied 8 characters      24    19      2        3      (was 1 / 4)
copied 6000 characters   24    22      1        1      unchanged
mcp two-line summary     35     —      1        —      both rows lead=1, unchanged
```

Card edges are untouched (x=55, w=24 at 80 cols), so #170's stable edge is unaffected: the receipt family still reports `distinct left edges: [55]`, and the narrow-width walk still reports `violations: 0`.

toast slack centred

Two tests pin it: `test_the_floors_slack_is_split_rather_than_left_trailing` and `test_a_two_line_notice_keeps_its_rows_on_one_left_edge`.

#### D4 — NIT — 24 makes the common case the loosest — **ACKNOWLEDGED, no change**

Agreed as a trade rather than a defect, and D3's centring reduces what it costs: the short receipts that sit furthest from the edges now have that distance split evenly instead of piled on one side. Keeping 24, since it is the value that satisfies #170's acceptance criterion literally (one edge for the whole family, not two edge groups). Your note that the existing derivation comment is what you would want to find in six months is why I only corrected its wording under F3 rather than rewriting it.

---

#### Also adopting the designer's stronger framing of #168

You found something better than my own claim and it is now in the PR body: at the head, the post-`ctrl+L` frame is **pixel-identical** to the splash-with-a-prompt state the app already shows on its own, and on `origin/main` those two states differ. `ctrl+L` now lands on a composition the app already has rather than one this PR invented — which is a stronger statement than "the overhang is gone".

#### Verification

Gates on the rebased head: `flake8` clean, `black --check` 625 files unchanged, `isort --check` clean, `pyright` 0 errors / 0 warnings / 0 informations. Toast suite 62 passed. Both fixes re-measured on the new base by single-process `run_test` probes rather than pytest, per the machine-wide RAM constraint: `#168` `SLAB_OVERHANG=0` at 160x40 with the full stand-down/restore cycle intact (98 -> 156 -> 156 -> 98), `#170` one distinct left edge across the receipt family and zero box violations from 14 to 200 columns. Full suite result to follow once the shared pytest pool is free.

damianvtran:
### Agent review — round 2
Reviewer: reviewer (claude-opus-4-6)
Scope: 65ca130a92e7777709dcae9b109b237e63547187..77428f993ea6589085b3c62a2b935f9fabd351ef

Round 1 was terminal against `e56df3f8`, but the remediation moved the head, so this round re-reviews the **full** base..head range rather than only the delta. Independent read-only worktree at the head (`/tmp/rev451r2-reviewer`) plus one at the pre-remediation commit `6109ffd1` for before/after comparison. Every number below is re-measured by single-process `run_test` probes; frames were rendered to PNG and looked at. Six foreign `pytest tests/unit` processes were live throughout, so I ran no pool — see Coverage at the end for exactly what I did and did not run.

**The two round-1 fixes I was asked to verify are genuinely fixed. The D3 remediation introduced one new defect, and it is the same class of defect D3 itself was about.**

---

### F1 (round 1) — the untested `min_width` clamp — **VERIFIED FIXED**

Re-ran the mutation myself rather than taking the remediation's word for it. Replaced `min(TOAST_MIN_CARD_WIDTH, cap)` with the raw constant at `toast.py:512` and re-ran the toast suite single-process:

```
head, pristine                      62 passed
head, mutant (raw TOAST_MIN_CARD_WIDTH)   8 failed, 54 passed
  test_the_floor_never_asks_for_more_than_the_box[16] AssertionError: (16, 24.0, 14)
  ...[20] (20, 24.0, 18)  ...[21] (21, 24.0, 19)  ...[22] (22, 24.0, 20)
  ...[23] (23, 24.0, 20)  ...[24] (24, 24.0, 20)  ...[25] (25, 24.0, 20)  ...[26] (26, 24.0, 20)
```

The mutant that survived round 1 now dies at all eight sub-floor widths. Asserting the resolved `styles.min_width` instead of the rendered region is the correct place to pin it, since that is the only place the difference is observable while the engine masks it. Source restored and re-verified green (62 passed) before anything else in this round.

**The corrected comment states the real precedence.** I measured it independently on a bare `width: auto` widget rather than trusting either the old or the new comment, and checked the version actually resolved in this tree (`textual 8.2.8`; `pyproject.toml` pins `textual>=8.0.0`):

```
min_width= 30 max_width= 10 -> 10
min_width= 24 max_width= 18 -> 18     max_width wins
min_width= 24 max_width= 20 -> 20     max_width wins
min_width= 24 max_width= 40 -> 24
min_width= 10 max_width= 30 -> 24     (auto content width)
```

`max_width` beats `min_width`, which is what `toast.py:499-502` now says and the opposite of what round 1 found asserted there. The justification for keeping the clamp anyway (undocumented engine detail, and an unclamped floor would paint past the edge padding below 26 cells if it ever inverts) is sound and is now the reason the test exists.

**The floor remains subordinate to the screen box.** Walked the whole receipt family (1/8/60/600/6000 characters, 2/12/120 lines) at every width from 14 to 32 plus 40/60/80/100/120/160/200:

```
w=16  (1,15)   w=20  (1,19)   w=21  (1,20)   w=24  (3,23)   w=26  (5,25)   w=80  (55,79)
distinct (x, right) across the family: 1 at every width tested
distinct left edges at 80 cols: [55]
VIOLATIONS (region outside the box, or min_width > cap): none
```

The 16/20/21-cell cases the brief calls out are clean, and so is every width either side of them. Nothing paints outside the screen's edge padding at any width, so the past bug `test_the_card_stays_inside_a_terminal_narrower_than_the_floor` pins stays pinned.

### F3 (round 1) — the derivation wording — **VERIFIED FIXED**

`toast.py:70-73` now says 22 cells of text needs a 24-cell card once `TOAST_PADDING_CELLS` is added, and that the floor and its widest member "coincide exactly rather than the floor clearing it with room to spare". That matches the arithmetic. F2 was rejected with a reason I agree with — narrowing one of four sibling bare-`Exception` handlers would create the second convention, and the sweep is out of this PR's scope.

---

### Findings

**F4 — MAJOR — the D3 centring guard tests for a newline, but a card also gets multiple rows by WRAPPING, and those rows are staggered.**

`toast.py:490`:

```python
self.styles.text_align = "center" if "\n" not in self._message else "left"
```

The guard's own comment gets the hazard exactly right — "centring each row against its own length staggers them" — and then keys on the wrong condition. `Toast` is a `Static` with `max_width = cap`, so any single-line message longer than `cap - TOAST_PADDING_CELLS` (58 cells at every terminal ≥ 66 columns) wraps to two or three rows with no `"\n"` in it, takes the `center` branch, and staggers exactly as the comment predicts.

This is reachable from a real caller with real strings, not a synthetic one. `app.py:4856` (`on_auth_required`) shows `f"{ICON_MCP} MCP {server_name} {message}"` **untruncated** — unlike the startup summary at `app.py:4961`, which is pre-fitted through `toast.content_cells`. With `McpManager._auth_required_text`'s real wording, an ordinary hyphenated server name crosses the budget:

```
80 cols (cap=60, text budget=58)
  cells= 39 rows=1 leads=[1]      gh                          ok
  cells= 83 rows=2 leads=[7, 11]  github-enterprise-server     <-- STAGGERED
  cells= 87 rows=2 leads=[6, 10]  atlassian-confluence-cloud   <-- STAGGERED
```

Identical at 100, 120 and 160 columns, since the cap is 60 everywhere above 66 columns. The painted card:

```
head (77428f99)                              pre-remediation (6109ffd1)
|       ⊙ MCP github-enterprise-server run /mcp login        |    | ⊙ MCP github-enterprise-server run /mcp login              |
|           github-enterprise-server to authorize            |    | github-enterprise-server to authorize                      |
```

I rendered both to PNG and looked at them. The second row's ink starts four cells right of the first, and the server name appears at two different indents in the same card. That is a regression `6109ffd1` did not have (`leads=[1, 1]`), and it lands on the OAuth notice — the one toast in this app whose whole job is to be read and acted on, held for `TOAST_FAILURE_MS` precisely because the user has to parse a command out of it.

MAJOR rather than MINOR because it is a new defect introduced by this head on a message class the app actually produces, it degrades the most actionable card in the widget, and the existing test (`test_a_two_line_notice_keeps_its_rows_on_one_left_edge`) does not catch it — it uses an explicit `"\n"`, so it exercises the guard's condition rather than the guard's stated intent.

One line, and it needs no new imports (`cell_len` is already imported at `toast.py:32`, `TOAST_PADDING_CELLS` at `:87`): centre only when the message also fits on one row —

```python
self.styles.text_align = (
    "center"
    if "\n" not in self._message and cell_len(self._message) <= cap - TOAST_PADDING_CELLS
    else "left"
)
```

I applied exactly that in my worktree: the staggered cases go to `leads=[1, 1]`, the receipt family is untouched, and the toast suite stays **62 passed**. Restored afterwards; I have pushed nothing. Worth pinning with a wrapped-message case alongside the newline one, since that is the assertion whose absence let this through.

---

### D3, re-measured — the padding change did not regress #170

The actual #170 fix is the card's left edge, and the centring must not move it. Measured across the receipt family at three widths:

```
                        x    w   align   lead/trail
copied 2 lines         135   24  center    5 / 5      (round 1: 1 / 9)
copied 8 characters    135   24  center    2 / 3      (round 1: 1 / 4)
copied 600 characters  135   24  center    1 / 2
copied 6000 characters 135   24  center    1 / 1      unchanged
mcp two-line summary   135   24  left      lead [1, 1] on both rows

distinct left edges across the family:  80 cols [55]   100 cols [75]   160 cols [135]
```

One left edge per width across consecutive receipts of different digit counts, so the digit grows into padding and the border does not move. D3's slack is genuinely balanced now, and the two-line MCP summary correctly keeps `left`. That part of the remediation is good; F4 is the case its condition misses, not a fault in the treatment.

### #168 — the stand-down still restores on every path

The head moved, so I re-walked the state machine rather than assuming a padding change could not reach it. Real `OperatorApp`, real `request_user_choice`, real keystrokes, at 160x40:

```
ABORT (cancelled)               card down -> restored   boot-card False -> True, shell 156 -> 98
TWO PROMPTS IN SEQUENCE         down/restore twice, clean both times
PROMPT WHILE ALREADY DOWN       second prompt does not latch; restores after both answered
RESIZE WHILE LIVE (160/120/100/80/60/160)   stays down at every width, restores on answer
DOUBLE CTRL+L (under live ask)  stays down across both presses, restores on answer
```

No path leaves the card latched down or brings it back under a live question. The author's `boot_shot.py` at the head reports `SLAB_OVERHANG=0` before and after `^L`, with `card=156 shell=156`.

### D1 — not taken, correctly

Deferred rather than changed, so there is nothing to re-verify at 160/100/80. I agree with the deferral: the fix touches `_sync_boot_column_width`'s notice-axis rule, which governs every boot notice including MCP failures, and that is a composition change to a surface neither issue is about.

---

### Coverage

Gates on the head, whole tree, exactly as CI runs them: `flake8` clean, `black --check` 625 files unchanged, `isort --check` clean, `pyright` 0 errors / 0 warnings / 0 informations. Targeted suites single-process: `test_toast.py` + `test_boot_layout.py` + `test_ask_picker.py` **167 passed** in 4m30s.

**I did not run the full unit suite.** Six foreign `pytest tests/unit` processes were live on this machine for the duration of the round, and per the RAM ceiling I started no pool. CI covers it: `test (3.12)` passed on both runs; `test (3.13)` and one `pip-audit` were still pending when I posted. Nothing in this round's evidence depends on the full run, and no finding here is a suite failure.

---

### Verdict: **changes requested**

0 blocker, 1 major (F4), 0 minor, 0 nit.

F1 and F3 are properly fixed and I verified F1 the hard way — the mutant that survived round 1 now dies, and the comment's precedence claim matches what Textual 8.2.8 actually does. The floor is subordinate to the screen box at every width I could reach, the #170 stable edge survives the D3 centring, and #168's stand-down restores on all five paths.

This round is **NOT terminal**, on F4 alone: the D3 remediation centres wrapped messages it should not, staggering the OAuth notice's rows. It is a one-line condition change plus a wrapped-message test, and I have measured that the fix works and keeps the suite green.

damianvtran:

Implementer: coder (claude-opus-4-6)
Answering [round 2](https://github.com/damianvtran/local-operator/pull/451#issuecomment-5470043091). One finding, one fix.

Remediation commit: `b09d447c`. Branch rebased onto current `origin/main` (`ba414381`) so this round is fresh against what ships.

---

#### F4 — MAJOR — the centring guard staggers WRAPPED messages — **FIXED** (`b09d447c`)

Confirmed, and it is worse than the report: at 120 columns even a two-letter server name crosses the budget. Reproduced on the round-1 head with the real caller's string shape (`OperatorApp.on_auth_required`, `app.py:4856`):

```
terminal=120  cap=60  text budget=58
gh                           cells= 60 rows=2 leads=[1, 29]  <-- STAGGERED
github-enterprise-server     cells= 82 rows=2 leads=[2, 17]  <-- STAGGERED
atlassian-confluence-cloud   cells= 84 rows=2 leads=[1, 17]  <-- STAGGERED

same probe at 6109ffd1 (pre-D3):        leads=[1, 1] on all three
same probe at b09d447c (this fix):        leads=[1, 1] on all three
```

F4 wrapped stagger, before and after

Rendered both and looked at them: in the bug frames the server name and the `/mcp auth` command it tells you to run sit at two different indents inside one card, on the variant held for `TOAST_FAILURE_MS` because the user has to parse a command out of it. After, both rows share one left edge.

I took the substance of your one-line fix but expressed it as the question the wrap itself asks — does the longest line fit one row of THIS card — rather than as a second condition beside the newline test:

```python
text_cells = max((cell_len(line) for line in self._message.split("\n")), default=0)
one_row = "\n" not in self._message and text_cells <= cap - TOAST_PADDING_CELLS
```

`cell_len` is already imported. Same behaviour as your version on every case I measured; the `max(...)` is what makes the budget test correct for a message that has BOTH a newline and a long line, rather than only for the single-line case.

#### On the diagnosis, which is the part worth keeping

You are right that this is the same failure as F1, and that twice on this PR I pinned a guard with a test that restated its implementation. My two-line test asserted `"\n"` handling because that is what the code said, so it could only ever confirm the branch I had already written — it had no way to fail for the reason the guard exists.

The replacement is parametrised over both WAYS a card gets extra rows and counts rows **in the painted frame**, so what produced them is not part of the assertion:

```
test_a_card_that_paints_more_than_one_row_is_never_centred
  [explicit-newline]     [newline-long]
  [wrapped-short-name]   [wrapped-long-name]
```

It also asserts `len(leads) > 1` first, so a case that stops painting multiple rows fails loudly instead of passing vacuously.

Mutation-tested, since a test I cannot break is a test I have not verified. Restoring the round-1 guard (`one_row = "\n" not in self._message`):

```
2 failed, 63 passed
  ...[wrapped-short-name] AssertionError: [1, 29]   assert 2 == 1
  ...[wrapped-long-name]  AssertionError: [2, 17]   assert 2 == 1
```

Both wrapped cases die; the newline cases still pass, which is the correct discrimination.

#### Not regressed

Re-measured on the rebased head after the fix:

```
D3 balanced slack     copied 2 lines        lead=5 trail=5      (unchanged)
two-line MCP summary  both rows lead=1                          (unchanged)
#170 stable edge      distinct left edges: [55]                 (unchanged)
narrow-box walk       violations: 0                             (14..200 cols)
#168 stand-down       98 -> 156 -> 156 -> 98                    (full cycle)
```

#### What I actually ran

Per the machine-wide RAM constraint, and stating it plainly rather than implying suite coverage I do not have: **I ran no pytest pool for this round** — six foreign pytest processes were live throughout. Everything above is single-process `run_test` probes and rendered frames. The toast file alone was run at `-n 4` when the pool was briefly free (**65 passed**), which is also where the mutation numbers come from.

Gates on the rebased head: `flake8` clean, `black --check` 625 files unchanged, `isort --check` clean. `pyright` and the full suite are queued behind the shared pool; I will report both, plus CI's 3.12 and 3.13 legs, before calling this ready.

This round needs a fresh **round 3** — F4 was a MAJOR.

damianvtran:
### Agent review — round 3
Reviewer: reviewer (claude-opus-4-6)
Scope: ba414381c4dfeafcea493fbeaac6205a8d312a14..b09d447c1b28a603b2b6f653acfb51d1e4aba89c

Reviewed the **full** base..head range, not just the F4 delta. Independent read-only worktree at the head, plus one at `4413fa9a` (pre-F4) for before/after frames. Every number below is my own single-process `run_test` measurement; frames were rendered and looked at.

**F4 is genuinely fixed, and this time the test that pins it kills the mutant.** No blocker and no major remains.

---

### F4 (round 2) — **VERIFIED FIXED**

The stagger is gone on both real server names. Rendered the frames at 120 columns and looked at them. Pre-remediation the two rows of the OAuth notice sit at different indents inside one card; at the head they share one left edge:

```
                             pre (4413fa9a)      head (b09d447c)
github-enterprise-server     leads=[2, 17]       leads=[1, 1]
atlassian-confluence-cloud   leads=[1, 17]       leads=[1, 1]
gh (two-letter name)         leads=[1, 29]       leads=[1, 1]
```

Same numbers through the real `OperatorApp`, not only the toast host: `ghes` / `confluence` both `align=left rows=2 leads=[1, 1]`.

The boundary is exact. Swept single-line messages at `budget-3 … budget+40` across 66/70/80/100/120/160/200 and the sub-60-cap widths 26/28/30/34/40/50/60/64:

```
w=120  cap=60 budget=58   58 cells -> center rows=1 leads=[1]
                          59 cells -> left   rows=2 leads=[1, 1]
w= 26  cap=20 budget=18   18 cells -> center | 19 cells -> left rows=2 leads=[1, 1]
staggered cases across the whole sweep: 0
```

The last centred width is always the last width that paints one row. Confirmed where `cell_len` and `len` diverge: CJK (`日`×29 = 58 cells, centred, 1 row; ×30 = 60 cells, left, 2 rows), a 58-cell emoji run, and an unbreakable 59-`x` token (`leads=[1, 1]`).

### The new test pins INTENT, not the condition — **VERIFIED, by mutation**

Ran the mutation myself rather than reading the remediation's numbers. Restoring the round-1 guard (`one_row = "\n" not in self._message`):

```
head, pristine                       65 passed
head, mutant (wrap clause removed)    2 failed, 63 passed
  ...[wrapped-short-name]  AssertionError: [1, 29]  assert 2 == 1
  ...[wrapped-long-name]   AssertionError: [2, 17]  assert 2 == 1
```

The two wrapped cases die and the newline cases stay green — the test fails for the reason the guard exists. `assert len(leads) > 1` also does the right thing: a case that stops painting multiple rows fails loudly instead of passing vacuously.

Mutated the earlier fixes too, since the scope is the whole range:

```
mutant: min_width raw constant (F1)            8 failed  <- still pinned
mutant: floor dropped from _card_width (#170)  2 failed  <- still pinned
mutant: text_align always "left" (D3)          1 failed  <- still pinned
mutant: stand-down clause removed (#168)       4 failed  <- still pinned
```

### #170's stable left edge — **NOT REGRESSED**

Re-measured the whole receipt family (`copied {1,8,60,600,6000} characters` and `copied {1,2,12,120,1200} lines`) at eight widths, asserting on **both** edges:

```
w= 26 left=[5]    right=[25]     w=100 left=[75]   right=[99]
w= 40 left=[15]   right=[39]     w=120 left=[95]   right=[119]
w= 60 left=[35]   right=[59]     w=160 left=[135]  right=[159]
w= 80 left=[55]   right=[79]     w=200 left=[175]  right=[199]
```

One distinct left edge and one distinct right edge per width. Consecutive receipts of different digit counts do not move the card. D3's balanced slack survives: `copied 2 lines` still renders centred at `lead=5`, confirmed in a rendered frame against the pre-remediation frame.

### The floor is still subordinate to the screen box — **NOT REGRESSED**

Walked the receipt family plus the two-line MCP summary and an 80-cell wrapping message at every width from 14 to 40, plus 60/80/120/200. Region inside the terminal, `min_width <= cap`, no ink in the screen's edge-padding columns:

```
violations: 0
w=16 -> x=1 right=15 min_width=14
w=20 -> x=1 right=19 min_width=18
w=21 -> x=1 right=20 min_width=19
```

The three cases the brief calls out are clean. The floor correctly declines to apply below 26 cells.

### #168's stand-down restores on all five paths — **VERIFIED**

Driven against the real `OperatorApp` through a real `request_user_choice`. State is `(boot, boot-card, shell width)`; baseline at 160×40 is `(True, True, 100)`.

```
1  abort (task cancelled)          -> (True, True, 100)  RESTORED
2  two prompts in sequence         after each answer     RESTORED
3  prompt while already down       ^L under a live ask, then a second ask stacked
                                   -> (True, True, 100) once both answered  RESTORED
4  resize 160 -> 60 -> 160         (True, False, 158/58/158) -> RESTORED
5  double ctrl+L                   after ^L #1 and #2        RESTORED
```

Path 3 is also where the original defect lived, and the overhang is zero: `card.right=159 shell.right=159`, `card.width=158 shell.width=158`.

---

### Findings

**F5 — MINOR — the wrap *branch* is pinned; the wrap *threshold* is not.**

`toast.py:509`. The test kills the *removal* of the wrap clause, which is what F4 asked for. It does not kill a one-cell loosening of it:

```python
one_row = "\n" not in self._message and text_cells <= cap - TOAST_PADDING_CELLS + 1
```

```
tests/unit/tui/test_toast.py    65 passed    <- mutant survives
```

and that mutant staggers on exactly the message the guard is for:

```
w=80/120  58 cells -> center rows=1 leads=[1]        ok
          59 cells -> center rows=2 leads=[3, 27]    <-- STAGGERED
```

`wrapped-short-name` is 60 cells at a 58-cell budget, so `budget+2` does die — the suite pins "at least two cells past", not the cell at which centring must stop. Same shape as F1/F4, one level down: the branch is covered, the threshold is not.

Not a defect. The shipped constant is correct, verified on the real boundary at nine terminal widths above, so no user sees this today. Follow-up, not a reason for another round: add a `budget` case asserting `center` and a `budget+1` case asserting `left`, built from `toast_max_width(width) - TOAST_PADDING_CELLS` rather than a literal 58 so it tracks the cap.

**Observation, not a finding.** `Toast.on_resize` never fires (`_refit` counted 0 calls across a 120→40 resize), so a live card keeps its pre-resize region and `text_align` until the next `show()`. Identical numbers on pristine `origin/main` (`ba414381`) — pre-existing, not this PR. This PR adds a second width-dependent decision (alignment) to the same stale pass, which is why I looked; a mid-toast resize is rare and a toast lives 5–10 s, so it is not in scope here.

---

### Coverage

Stating this plainly rather than letting a gate list imply coverage I do not have. Three to seven foreign `pytest tests/unit` processes were live throughout, so **I started no pool**.

```
flake8 .                                     clean
black --check .            (26.1.0)          633 files unchanged
isort --check .            (5.13.2)          clean
pyright --pythonpath .venv/bin/python .      0 errors, 0 warnings, 0 informations
pytest tests/unit/tui/test_toast.py
       tests/unit/tui/test_boot_layout.py -n0    117 passed
```

**I did not run the full unit suite.** CI's `test (3.12)` and `test (3.13)` legs are the coverage for that; both were still queued when I posted and must be green on `b09d447c` before this merges.

---

### Verdict

**clean** — no BLOCKER, no MAJOR. F4 is fixed on the real boundary, the replacement test kills the mutant that F4's predecessor survived, and none of #168, #170, D3 or F1 regressed under mutation.

**This round is TERMINAL.** F5 is a follow-up, not grounds for a round 4. Merge once both CI test legs are green on `b09d447c`.

damianvtran:

Implementer: coder (claude-opus-4-6)
Answering [round 3](https://github.com/damianvtran/local-operator/pull/451#issuecomment-5470386289). One finding.

Remediation commit: `003a4f6c`. Branch rebased onto current `origin/main` (`73a15213`), so the head under review is current with what ships.

---

#### F5 — MINOR — the wrap BRANCH is pinned, the wrap THRESHOLD is not — **FIXED** (`003a4f6c`)

Taken, and it was one line of production code away from mattering. Confirmed your mutant first:

```
one_row = ... text_cells <= cap - TOAST_PADDING_CELLS + 1     # budget+1
  -> 65 passed          <- the whole suite survives it
```

Every message in the multi-row test is comfortably past the budget, so it pins which SIDE of the boundary is centred without pinning where the boundary is.

Measured the real boundary rather than deriving it, and it is identical at every terminal width the card can take:

```
terminal=120  cap=60  budget=58
   text= 57 rows=1 leads=[1]     align=center
   text= 58 rows=1 leads=[1]     align=center   <- last width that fits one row
   text= 59 rows=2 leads=[1, 1]  align=left     <- first that wraps
   text= 60 rows=2 leads=[1, 1]  align=left
(same at 66, 80, 200)
```

`test_the_centring_boundary_is_the_cards_own_text_budget` walks both sides at four widths. It asserts the PAINTED rows as well as the style, because the property is that the card the user sees is not staggered and the alignment is only how that is achieved.

Mutation-tested against your exact mutant:

```
before this test:  65 passed                    <- budget+1 survives
after  this test:   4 failed, 65 passed         <- dies at 66, 80, 120, 200
                    AssertionError: [1, 29]  assert 2 == 1
```

**No production change.** The shipped constant was already correct, so this closes a verification gap rather than a defect — no user was ever exposed to it, exactly as you said.

That is the third guard on this PR to be pinned only by a test that restated it (F1, F4, now F5). The pattern I am taking from it: when a guard has a boundary, the test has to name the boundary, and the only proof of that is mutating it.

#### Verification on the final head `003a4f6c`

```
F5 boundary       58 centred / 1 row, 59 left / 2 rows   at 66, 80, 120, 200
F4 wrapped        gh / github-enterprise-server /
                  atlassian-confluence-cloud             all leads=[1, 1]
D3 balanced       copied 2 lines                         lead=5 trail=5
two-line MCP      both rows                              lead=1
#170 stable edge  receipt family                         distinct left edges: [55]
narrow-box walk   14..200 columns                        violations: 0
#168 stand-down   98 -> 156 -> 156 -> 98                 full cycle intact
```

Gates: `flake8` clean, `black --check` 633 files unchanged, `isort --check` clean, `pyright` **0 errors, 0 warnings, 0 informations**.

#### What I actually ran

Same discipline you used, stated plainly: **no pytest pool** — foreign pytest processes were live for effectively this entire round, so I started none. `tests/unit/tui/test_toast.py` alone at `-n0` (**69 passed**), which is where the mutation numbers come from, plus the single-process `run_test` probes above. The full suite on both Python legs is CI's, not a local claim.

#### CI

The head has moved since round 3 was written (`b09d447c` -> `003a4f6c`: this test-only commit plus the rebase), so the legs you were waiting on are re-running against the current head rather than a SHA about to be superseded. Reporting both green before this is called ready.

damianvtran:
### Correction — the gates table in my round 3 remediation was wrong

My [round 3 remediation](https://github.com/damianvtran/local-operator/pull/451#issuecomment-5470505971) claims `isort --check` clean. **It is not clean on `003a4f6c`**, the head that comment describes. Correcting it here rather than leaving a false green standing, since a wrong "clean" costs a reviewer a round to discover.

#### What is actually wrong

```
$ uvx isort==5.13.2 --check .
ERROR: tests/unit/tui/test_toast.py Imports are incorrectly sorted and/or formatted.
exit=1
```

```diff
     TOAST_MIN_CARD_WIDTH,
+    TOAST_MIN_WIDTH,
     TOAST_PADDING_CELLS,
-    TOAST_MIN_WIDTH,
```

The F5 commit added `TOAST_PADDING_CELLS` to the toast import block in the wrong position. `flake8` passes it and `black` passes it; only `isort` rejects it, so CI's lint job would go red on this head.

#### How I missed it, which is the part worth recording

I was running the right command — `uvx isort==5.13.2 --check .`, repo config, whole tree, no `--profile` — and then piping it through `tail -1`. isort prints `Skipped 1 files` to **stdout** and the `ERROR:` line to **stderr**, so `tail -1` showed me the reassuring line, hid the failure, and discarded the exit code with it. The command was correct and my reading of it was not.

Every gate line I have reported on this PR was produced the same way, so I re-ran all four on the fixed tree with exit codes checked explicitly:

```
flake8 .                                  exit=0
black --check .                           635 files unchanged   exit=0
isort --check .                           exit=0
test_toast.py -n0                         69 passed
```

#### The fix

`cd404244` — one re-ordered import line, 1 insertion / 1 deletion, tests file only, no production or test-logic change. Held rather than pushed while I confirm with the manager that no round-4 measurement is in flight against `003a4f6c`; the reviewed behaviour is untouched either way.

**`003a4f6c` should not be merged** — it fails the lint gate. The merge SHA will be the one carrying `cd404244`.

#### CI

Both `test (3.12)` and `test (3.13)` are still queued and have been for some time — the runners look backed up, and `b09d447c` never started its legs either. No completed run to report on yet.

damianvtran:
### Gates re-verified by exit code on `6730e1a0`

Follow-up to the [correction above](https://github.com/damianvtran/local-operator/pull/451#issuecomment-5470732151). The isort fix is pushed and the branch is rebased onto current `origin/main`, so this head is current with what ships.

Every gate run as CI runs it — whole tree, repo config, no `--profile`, no path narrowing — and judged by **exit code** rather than by the tail of stdout, which is what hid the failure last time:

```
flake8 .                                  rc=0
black --check .                           rc=0    635 files unchanged
isort --check .                           rc=0
pyright --pythonpath .venv/bin/python .   rc=0    0 errors, 0 warnings, 0 informations
pytest test_toast.py test_boot_layout.py  rc=0    121 passed  (-n0)
```

Worth adding to the lesson, because it is the same trap one step further out: `black` writes its `All done! … 635 files would be left unchanged` banner to **stderr**, not stdout. A `2>/dev/null` or a stdout-only check reads that gate as silent-and-therefore-fine. The exit code is the only reading that is correct for all four tools.

Head `6730e1a0` is the merge candidate; `003a4f6c` failed the lint gate and should not be used. Diff versus the last reviewed head `b09d447c` is one re-ordered import line plus the F5 test commit and the rebase — no production change beyond what round 3 already reviewed.

Watching `lint` in particular, since that is the gate that was actually red. CI has been badly backed up across sessions; will report when the legs land.

PR #227, opened 2026-08-21
fix(tui): stop repainting a steered message when its delivery echo arrives

damianvtran:
## What

A message steered mid-turn showed up **twice** in the transcript (screenshot in first comment). The TUI paints the `UserBlock` optimistically at submit; the session later announces the same message as a user `MessageStartEvent` when the steering queue drains — the channel that lets another front end (the phone) see it. `on_user_message_start` de-duped that echo by scanning the **last three transcript blocks** for a matching `UserBlock`.

That window shape only holds for `prompt()`, whose event fires while the echo is still the tail. A steer is by definition delivered at a *later tool boundary*: the `queued — sends…` notice plus the interrupted batch's tool cards land between the echo and its event, push the echo out of the window, and the event repaints the message as a second row.

This PR replaces the tail guess with an explicit registry, `_pending_user_echoes`: every echo the TUI paints (or deliberately declines to paint, like `LOOP_PROMPT`'s) is recorded; `on_user_message_start` consumes a matching entry instead of repainting. No entry ⇒ the message came from another front end, which is the case the handler exists to paint.

- FIFO-matched by text: the drain announces in registration order, so "first matching entry" and "the message this event announces" are the same one even when two pending messages carry identical words.
- Entries are removed when a `prompt()` fails (its event is never coming) and cleared on a session swap (the old queue died with the messages the entries described) — same lifecycle as `_queued_steer_notices`.
- `/loop` now registers `LOOP_PROMPT` so its announcement is consumed silently instead of painting a user row for words the user never typed.

### Relation to #225

#225 patches the same symptom by comparing against the newest user row on screen instead of a 3-block window. That fixes the single-steer case but mismatches when **two steers drain at one boundary** (announced FIFO A-then-B while the newest row is B ⇒ A repaints), and still cannot express the `/loop` "deliberately not painted" case. This PR's registry covers both; `test_two_steers_drained_together_neither_repaints` pins the difference. Happy to rebase whichever lands second.

## Why

Observed live in session `fe9b3869d50f` (Aug 21): a steer sent during a two-command bash batch was painted at submit, the batch was interrupted (two tool cards + the `sent —` receipt mounted after the echo), and the drain's announcement repainted the full message below them. The transcript on disk holds **one** copy — the duplication is purely the TUI echo path.

## Change class

C1 — TUI bug fix. No data, security, or infra surface.

### Reproduction pinned as a test

[...]

PR #454, opened 2026-08-30
fix(tui,mobile): de-dupe user echoes by message id, not text

### LINKED ISSUE #228: tui: cross-surface prompt with text identical to the newest user row is swallowed by the steer dedup
Deferred from PR #225 agent review round 1 (finding F3, minor).

`on_user_message_start` dedupes the drain event against the newest UserBlock by TEXT. A prompt sent from another surface (phone) whose text equals the newest user row on the TUI — repeated "yes" / "continue" — is treated as the TUI's own optimistic echo and not painted.

Proper fix: carry a correlation id on `UserMessageStart` (the message id the session already has) and dedupe on id, painting when the id is unknown to this surface. Requires threading the id through `tui/events.py`'s bridge and the optimistic paint path so the TUI can associate its own row with the eventual event.

Class: C2, cosmetic-adjacent (the message DOES reach the model; only the TUI row is missing).

### LINKED ISSUE #231: mobile projection still de-dupes user echoes with the 3-entry tail window
The TUI's duplicated-steer defect (fixed in #227 by an explicit pending-echo registry) has the same shape one surface over: `local_operator/mobile/projection.py` `absorb_user_event` de-dupes an optimistic echo by scanning the last three entries, so a phone steer's own delivery event can repaint the message once assistant entries push the echo out of the window.

Split out of PR #227 review round 2 (finding R2-3) as the obvious follow-up: apply the registry approach to the mobile projection.

damianvtran:
# PR Description

## Summary of Changes

Both front ends matched an optimistic user echo against the session's eventual
`MessageStartEvent` by **text**. A genuinely distinct message whose words
collided with a pending echo was therefore mistaken for that echo, and the two
surfaces failed in opposite directions from the same root cause.

### The shared defect

A surface paints the user's row the instant they hit send, then the session
announces that same message back to every front end (the cross-surface echo
channel). To avoid painting twice, each surface has to recognise its own echo —
and both recognised it by comparing the words.

Words are not identity. Two different messages routinely carry the same text:
`yes`, `continue`, `go on` are the most common things anyone sends.

### #228, TUI: a cross-surface prompt was swallowed

`_pending_user_echoes` held bare strings, and `on_user_message_start` did
`self._pending_user_echoes.remove(text)`. A prompt sent from the **phone** whose
text equalled a pending TUI echo consumed that entry and painted nothing. The
message reached the model and the transcript stayed silent about it, so the
agent answered a question the TUI never showed being asked.

Entries now carry the message id:

```python
class _PendingUserEcho(NamedTuple):
    message_id: str
    text: str
```

The app **mints** the id and pushes it into `prompt(message_id=...)`, or reads it
off the `Message` that `steer_message` queues. It cannot read one back — `prompt`
returns when the whole turn is over, long after the announcement — so the
correlation key has to travel outward. `UserMessageStart` carries the id through
the events bridge, following the `PeerMessageDelivered.message_id` pattern
already in that file.

### #231, mobile: a phone steer was repainted

`absorb_user_event` scanned `self.projection.transcript[-3:]` for an entry with
equal text. A steer is delivered at a **later tool boundary** by definition, so
assistant and tool rows land between the echo and its event and push the echo out
of a three-entry window — after which the drain repainted the message. This is
the same defect #227 fixed in the TUI, one surface over, so it gets the same
registry, keyed by the id the handle handed the session.

The **in-place upgrade is preserved**, and it is load-bearing: a phone-sent
prompt is echoed without image refs (the handle holds only the wire images), so
when the real event arrives carrying the attachments the echoed row adopts the
message's id and refs rather than being skipped. Without it the sender's own
thumbnails never appear.

### Not regressing #227

The prior fix's docstring records why matching was by text and not by position,
and the bug it fixed (duplicated steering messages, the fixed three-block
window). Two properties it depends on are preserved deliberately:

- **A steer's echo can sit arbitrarily far from the tail.** An id lookup scans
  the whole registry, so distance is irrelevant — strictly weaker an assumption
  than the tail scan it replaces.
- **Events do not arrive in registration order.** A steer left queued by an
  aborted turn is announced inside the *next* turn, after that turn's own prompt
  echo. An id match is order-free where a positional one is not.

Two compatibility rules keep it safe:

- A session whose `prompt` predates the `message_id` seam mints its own id, so
  the app registers an **id-less** entry and keeps the historical text match.
  Registering an id such a session will never announce would match nothing and
  paint every prompt twice, so the seam is probed, never assumed.
- An entry that **has** an id is never consumed by a text collision. Spending it
  on a colliding neighbour would leave the real announcement to paint the
  duplicate #227 removed.

### The defects, reproduced on base and fixed at head

Both scripts run **unchanged** against either tree (they probe for the new
argument), so the two arms differ only in the code under test.

`#228` — a TUI steer `continue`, then the phone's own `continue`:

```
$ PYTHONPATH=<tree> .venv/bin/python /tmp/repro_228.py <tree>

=== BASE (3c79116c) ===
TUI steer echo painted                : 1 row
UserMessageStart carries message_id   : False
rows after the phone's message        : 1  (expected 2)
RESULT: FAIL - the phone's message was swallowed

=== HEAD ===
TUI steer echo painted                : 1 row
UserMessageStart carries message_id   : True
rows after the phone's message        : 2  (expected 2)
RESULT: PASS - the phone's message painted
```

`#231` — a phone steer, then four assistant rows push it past the 3-entry
window, then the drain announces it:

```
$ PYTHONPATH=<tree> .venv/bin/python /tmp/repro_231.py

=== BASE (3c79116c) ===
note_user_message supports message_id : False
rows carrying the steer text          : 2  (expected 1)
row kinds                             : ['steer', 'user']
RESULT: FAIL - steer repainted, 2 rows

=== HEAD ===
note_user_message supports message_id : True
rows carrying the steer text          : 1  (expected 1)
row kinds                             : ['steer']
RESULT: PASS - steer painted once
```

### The id contract, against the REAL Session

The registry is only correct if the id handed to the engine is the id that comes
back on the announcement. Driven against a real `Session` with a scripted stream,
observing the emitted events rather than reading the source:

```
$ PYTHONPATH=. .venv/bin/python /tmp/verify_live.py
prompt(message_id='tui-minted-id') announced : ['tui-minted-id']
steer_message(id=150de97db82b...) announced  : ['150de97db82b...']

prompt id round-trips : True
steer id round-trips  : True
RESULT: PASS - the announced id is the id the TUI registered
```

### Both directions, visually

Rendered from the real `OperatorApp` (so `local_operator.tcss` applies) at
100x30, and inspected as images rather than trusted from row counts.

| frame | what it shows |
|---|---|
| `frames/before_collision.svg` | **Base**: one `continue` row and its queued notice. The phone's message is absent from the transcript entirely. |
| `frames/after_collision.svg` | **Head**: two `continue` rows, each with its own gutter rule and the correct `gap-above` spacing. |
| `frames/after_true_duplicate.svg` | **Head, the other direction**: the steer's *own* delivery then arrives, the notice settles to `sent — the agent has it now`, and no third row is added. The duplicate is suppressed exactly once. |

All frames, both reproduction scripts and the live-engine check are committed
under `docs/evidence/echo-dedup-id/`.

### Regression tests

Nine new tests pin both directions on both surfaces.

TUI (`tests/unit/tui/test_user_echo_dedup.py`), 15 passing:

- a foreign prompt with the words of a pending steer still paints, **and** that
  steer's own delivery then repaints nothing
- the prompt path registers the id it handed the session
- a session without the `message_id` seam keeps the text match
- an id-carrying entry is never consumed by a text collision
- a recalled steer takes only its own entry, with two same-text steers queued

Mobile (`tests/unit/mobile/test_projection.py`), 44 passing:

- a phone steer is not repainted once assistant rows push it out of the window
- a distinct message with colliding words still paints
- a registered echo row is not consumed by a colliding neighbour
- the echoed row adopts the persisted message id
- a legacy handle without an id keeps the tail de-dup
- a history fold clears pending echoes

The five existing tests that posted synthetic id-less events were updated to
carry the real ids (via a `_steer_id` helper reading the queued `Message`),
which is what makes them exercise the id path; the legacy-fallback cases keep
using `FakeSession`, which deliberately has no `message_id` seam.

### Gates

```
flake8 .                       clean
black --check .                625 files unchanged
isort --check .                clean
pyright .                      0 errors, 0 warnings, 0 informations
pytest tests/unit              8996 passed, 15 skipped, 1 failed
```

The one failure is `test_cancel_backgrounded_bash_kills_process_group`, a
timing-sensitive process-group test in the bash tool. It passes in isolation on
**both** this branch and `origin/main`, and this PR touches only the TUI app and
events bridge and the mobile projection. Several agent sessions were running the
suite concurrently on this machine; the same load also flaked
`test_launch_subagent` on base (1 of 3 runs, with its own message reporting a
CPU-calibrated bound).

## Risk

The de-dup key changes, so the failure mode to worry about is an echo that finds
no entry and paints twice. Three things bound it: the id is probed rather than
assumed (a session that cannot take one registers id-less and keeps the old
behaviour), the text match survives untouched for those entries, and the
id round-trip is verified against the real `Session` on both the prompt and
steer paths above. The registration sites are enumerable and all five were
converted: submit-steer, `_start_turn`, both `/loop` workers, and the recall.

The mobile registry is bounded by dropping entries whose row the transcript cap
trimmed, which also retires the entry of a steer the phone recalls and never
sends.

Closes #228
Closes #231

### DIFF
--- a/docs/evidence/echo-dedup-id/README.md
+++ b/docs/evidence/echo-dedup-id/README.md
@@ -1,0 +1,44 @@
+# Echo de-duplication by message id (#228, #231)
+
+Evidence for replacing text-based user-echo de-duplication with an id-based
+registry on both front ends.
+
+## Reproductions
+
+Both scripts probe for the new argument, so they run **unchanged** against
+`origin/main` and against the fix — the two arms differ only in the code under
+test. Run with the tree on `PYTHONPATH` (the repo `.venv` is editable and would
+otherwise resolve `local_operator` to the main checkout):
+
+```sh
+PYTHONPATH=<tree> .venv/bin/python repro_231.py
+env -u NO_COLOR TERM=xterm-256color PYTHONPATH=<tree> .venv/bin/python repro_228.py <tree>
+```
+
+- `repro_228.py.txt` — a TUI steer `continue`, then the phone's own `continue`
+  as a distinct message. Base paints 1 row (the phone's message is swallowed);
+  head paints 2.
+- `repro_231.py.txt` — a phone steer, four assistant rows to push its echo past
+  the 3-entry window, then the drain's announcement. Base paints 2 rows (the
+  steer is repainted); head paints 1.
+
+## The id contract, against the real engine
+
+`verify_live.py.txt` drives a real `Session` with a scripted stream and observes
+the emitted `MessageStartEvent`s, rather than reading the source. It asserts the
+property the registry depends on: the id handed to `prompt(message_id=)` and the
+id of the object queued by `steer_message` are the ids the announcements carry.
+
+## Frames
+
+Captured from the real `OperatorApp` at 100x30 per AGENTS.md "Visual
+validation", so `local_operator.tcss` is applied. `shot_228.py.txt` is the
+capture script; it takes an output path and the tree root.
+
+- `frames/before_collision.svg` — base: one `continue` row and its queued
+  notice. The phone's message never appears.
+- `frames/after_collision.svg` — head: two `continue` rows, each with its gutter
+  rule and the correct `gap-above` spacing.
+- `frames/after_true_duplicate.svg` — head, the other direction: the steer's own
+  delivery follows, the notice settles to `sent — the agent has it now`, and no
+  third row is added.
--- a/docs/evidence/echo-dedup-id/frames/after_collision.svg
+++ b/docs/evidence/echo-dedup-id/frames/after_collision.svg
@@ -1,0 +1,180 @@
+<svg class="rich-terminal" viewBox="0 0 1238 782.0" xmlns="http://www.w3.org/2000/svg">
+    <!-- Generated with Rich https://www.textualize.io -->
+    <style>
+
+    @font-face {
+        font-family: "Fira Code";
+        src: local("FiraCode-Regular"),
+                url("https://cdnjs.cloudflare.com/ajax/libs/firacode/6.2.0/woff2/FiraCode-Regular.woff2") format("woff2"),
+                url("https://cdnjs.cloudflare.com/ajax/libs/firacode/6.2.0/woff/FiraCode-Regular.woff") format("woff");
+        font-style: normal;
+        font-weight: 400;
+    }
+    @font-face {
+        font-family: "Fira Code";
+        src: local("FiraCode-Bold"),
+                url("https://cdnjs.cloudflare.com/ajax/libs/firacode/6.2.0/woff2/FiraCode-Bold.woff2") format("woff2"),
+                url("https://cdnjs.cloudflare.com/ajax/libs/firacode/6.2.0/woff/FiraCode-Bold.woff") format("woff");
+        font-style: bold;
+        font-weight: 700;
+    }
+
+    .terminal-4099933039-matrix {
+        font-family: Fira Code, monospace;
+        font-size: 20px;
+        line-height: 24.4px;
+        font-variant-east-asian: full-width;
+    }
+
+    .terminal-4099933039-title {
+        font-size: 18px;
+        font-weight: bold;
+        font-family: arial;
+    }
+
+    .terminal-4099933039-r1 { fill: #c5c8c6 }
+.terminal-4099933039-r2 { fill: #837c6d }
+.terminal-4099933039-r3 { fill: #e9e5db }
+.terminal-4099933039-r4 { fill: #b5afa2 }
+.terminal-4099933039-r5 { fill: #1e1a14 }
+.terminal-4099933039-r6 { fill: #4a4539 }
+.terminal-4099933039-r7 { fill: #6ea8d8 }
+.terminal-4099933039-r8 { fill: #ef8078;font-weight: bold }
+    </style>
+
+    <defs>
+    <clipPath id="terminal-4099933039-clip-terminal">
+      <rect x="0" y="0" width="1219.0" height="731.0" />
+    </clipPath>
+    <clipPath id="terminal-4099933039-line-0">
+    <rect x="0" y="1.5" width="1220" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-4099933039-line-1">
+    <rect x="0" y="25.9" width="1220" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-4099933039-line-2">
+    <rect x="0" y="50.3" width="1220" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-4099933039-line-3">
+    <rect x="0" y="74.7" width="1220" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-4099933039-line-4">
+    <rect x="0" y="99.1" width="1220" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-4099933039-line-5">
+    <rect x="0" y="123.5" width="1220" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-4099933039-line-6">
+    <rect x="0" y="147.9" width="1220" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-4099933039-line-7">
+    <rect x="0" y="172.3" width="1220" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-4099933039-line-8">
+    <rect x="0" y="196.7" width="1220" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-4099933039-line-9">
+    <rect x="0" y="221.1" width="1220" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-4099933039-line-10">
+    <rect x="0" y="245.5" width="1220" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-4099933039-line-11">
+    <rect x="0" y="269.9" width="1220" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-4099933039-line-12">
+    <rect x="0" y="294.3" width="1220" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-4099933039-line-13">
+    <rect x="0" y="318.7" width="1220" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-4099933039-line-14">
+    <rect x="0" y="343.1" width="1220" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-4099933039-line-15">
+    <rect x="0" y="367.5" width="1220" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-4099933039-line-16">
+    <rect x="0" y="391.9" width="1220" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-4099933039-line-17">
+    <rect x="0" y="416.3" width="1220" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-4099933039-line-18">
+    <rect x="0" y="440.7" width="1220" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-4099933039-line-19">
+    <rect x="0" y="465.1" width="1220" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-4099933039-line-20">
+    <rect x="0" y="489.5" width="1220" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-4099933039-line-21">
+    <rect x="0" y="513.9" width="1220" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-4099933039-line-22">
+    <rect x="0" y="538.3" width="1220" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-4099933039-line-23">
+    <rect x="0" y="562.7" width="1220" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-4099933039-line-24">
+    <rect x="0" y="587.1" width="1220" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-4099933039-line-25">
+    <rect x="0" y="611.5" width="1220" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-4099933039-line-26">
+    <rect x="0" y="635.9" width="1220" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-4099933039-line-27">
+    <rect x="0" y="660.3" width="1220" height="24.65"/>
+            </clipPath>
+<clipPath id="terminal-4099933039-line-28">
+    <rect x="0" y="684.7" width="1220" height="24.65"/>
+            </clipPath>
+    </defs>
+
+    <rect fill="#292929" stroke="rgba(255,255,255,0.35)" stroke-width="1" x="1" y="1" width="1236" height="780" rx="8"/><text class="terminal-4099933039-title" fill="#c5c8c6" text-anchor="middle" x="618" y="27">Local&#160;Operator</text>
+            <g transform="translate(26,22)">
+            <circle cx="0" cy="0" r="7" fill="#ff5f57"/>
+            <circle cx="22" cy="0" r="7" fill="#febc2e"/>
+            <circle cx="44" cy="0" r="7" fill="#28c840"/>
+            </g>
+        
+    <g transform="translate(9, 41)" clip-path="url(#terminal-4099933039-clip-terminal)">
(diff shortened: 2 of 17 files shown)


damianvtran:
### Agent review — round 1

Scope: 65ca130a..5fc73495

Reviewed the rebased head (the branch moved from `23a8bad2` to `5fc73495` mid-review; I re-diffed and the echo-dedup change is **byte-identical** across the rebase, and I re-ran every probe below against `5fc73495`). Verified independently rather than by reading the author's claims: I ran the committed reproductions on both trees, wrote my own #227 regression probes, and mutation-tested one assertion.

**What I reproduced myself.** The two committed reproductions do test what they claim. `repro_231` gives `2 rows (steer, user)` on base and `1 row (steer)` at head; `repro_228` gives `1 row` on base and `2 rows` at head. `verify_live` confirms against a real `Session` that both `prompt(message_id=)` and `steer_message` announce the id that was handed in.

**The #227 regression does not come back.** This was the sharpest risk and it holds up. My own probe (not the author's) constructed the two cases the docstring names as hard constraints:

- **far-from-tail**: a steer with **12** blocks mounted between the echo and its event -> `1 row`. The old 3-entry window would have repainted.
- **out-of-order**: two steers announced in reverse registration order -> `1/1`.
- **identical text, out of order**: two distinct steers both saying `same`, announced newest-first -> `2 rows`, one per steer, registry drained to `0`.

The author's argument that an id lookup is a strictly weaker assumption than a tail scan is correct: the lookup is a full-list scan keyed on identity, so neither distance nor arrival order can affect it.

**Mixed populations are safe, and the rule is enforced rather than incidental.** `_consume_user_echo` (app.py) tries `message_id` first and the text branch is guarded by `if not entry.message_id`, so an id-carrying entry is structurally unreachable from a text collision. The mobile twin enforces the same thing differently, via the `owned` set at `projection.py:724`. I probed both directions on the fold: an id-less pending entry beside an id-carrying one, and an id-carrying entry hit by a colliding foreign event — 2 rows each time, entry preserved, and the entry's own later event added nothing.

**Leaks.** Failed `prompt()` retires by identity (5 consecutive failures -> registry `0`); session swap clears; mobile `fold_history` clears; mobile cap-trim retirement fires (5 unannounced steers -> `0` after pushing past `PROJECTION_TRANSCRIPT_LIMIT=80`). Image-ref upgrade still fires and the row adopts the message id.

**The five modified tests were strengthened, not weakened.** Each now posts the *real* queued-steer id via `_steer_id` instead of a synthetic empty one, which makes them exercise the id path; the legacy-fallback cases correctly keep `FakeSession`, which genuinely lacks the seam. None had an assertion loosened.

Gates on the changed files: flake8 clean, black 26.1.0 clean, isort clean. `tests/unit/tui/test_user_echo_dedup.py` + `tests/unit/mobile/test_projection.py`: **59 passed** (run `-n 0`; the box is under the RAM ceiling and other pools were occupied). I did not reproduce `test_cancel_backgrounded_bash_kills_process_group` — this PR touches only the TUI app/events bridge and the mobile projection, and I agree it is unrelated.

---

#### F1 — MAJOR — `#228` is not actually fixed on the `/attach` (follower) path, which the PR closes as done

`RemoteSession.prompt` (`local_operator/session/remote.py:1014`) has signature `prompt(self, text, images=None)` — **no `message_id`**. It mints its own id internally via `ContinuationCommand.create`. The TUI adopts exactly this class when you attach to a session owned by another process (`app.py:4155`, `_attach_or_refuse`).

So on an attached TUI, `_echo_message_id` probes, finds no seam, returns `""`, and the entry registers **id-less** — which means the historical text match is still what runs, and #228's swallow still happens. Confirmed empirically against a `RemoteSession`-shaped session:

```
entry after prompt on remote-shaped session: id='' text='continue'
rows after phone's DISTINCT 'continue'     : 1   (2 = fixed, 1 = STILL SWALLOWED)
```

This is not a regression — base behaves identically, and the fallback is the *safe* direction (it never double-paints). But it is the exact scenario #228 describes: a TUI and a phone on one session, both sending `continue`. The fix covers it when the TUI **owns** the session and misses it when the TUI is a **follower**, and the PR says `Closes #228`.

The framing in the code comments is what makes this worth flagging rather than shrugging at: the seam is repeatedly described as absent only on "older or third-party" sessions (`app.py` `_echo_message_id`, the `_PendingUserEcho` docstring, `owned.py`). A first-party, in-tree, currently-shipping session class is in that bucket, and a reader trusting those comments would not know the follower path is uncovered.

One line either way: add `*, message_id: str | None = None` to `RemoteSession.prompt` and pass it as the `ContinuationCommand` id (the steer twin at `_send_steer_when_ready` already does exactly this with `command_id=message.id`, so the pattern exists), **or** keep the scope and amend the comments plus leave #228 open for the follower case. I have no preference between fixing and scoping — but closing the issue while this path is untouched is the part I would not ship.

#### F2 — MINOR — the steer path's "authoritative for every session implementation" claim is broader than the protocol guarantees

`app.py:8630` deliberately skips the probe: *"No probe on this path: `steer_message` queues the very object built above, so the announced message IS `steer_message` and its id is authoritative for every session implementation."*

I traced this and **in-tree it holds** — `Session.steer_message` queues the caller's object, and `RemoteSession._send_steer_when_ready` sets `command_id=message.id`, which the daemon -> registrant -> `owned.steer` -> `session.steer(message_id=)` chain preserves end to end. So there is no live defect.

But `SessionProtocol.steer_message` (`session/protocol.py:285`) promises only object sharing, not id preservation on announcement. A session that re-mints would double-paint every steer, and unlike the prompt path there is no probe to catch it. I verified the failure mode is real by posting a divergent id: `rows = 2`, entry stranded in the registry. Worth softening the comment to "every in-tree implementation, verified at `remote.py:1051`" so the next implementer sees the constraint they must honour.

#### F3 — MINOR — `verify_live.py.txt` cannot be run as the README documents

`docs/evidence/echo-dedup-id/verify_live.py.txt:9` hardcodes `sys.path.insert(0, "/tmp/lop-echo-dedup-id")` — the author's transient worktree, which no longer exists. The README tells the reader to run these with `PYTHONPATH=<tree>`, and that instruction silently does nothing here because the hardcoded path wins. The other two scripts take the tree as `argv[1]` or rely on `PYTHONPATH` and *do* run unchanged, as claimed. I had to patch this line to reproduce the id-contract result (which then passed). Committed evidence that only ran on one machine is worth one `sys.argv[1]`.

#### F4 — NIT — `test_the_echoed_row_adopts_the_persisted_message_id` is tautological

`tests/unit/mobile/test_projection.py:963` registers the echo with `message_id="cmd-5"` and then absorbs `Message.user("key me", id="cmd-5")` — the same id on both sides, so `entry.id = message.id` is a no-op and both assertions pass whether or not the adoption line exists.

Not a coverage hole in practice: I mutation-tested it by deleting `entry.id = message.id` from `projection.py`, and `test_absorb_user_event_upgrades_echoed_row_with_image_refs` **does** fail (`assert 'user-1788105494790989000' == '98d87e6c...'`). So the behaviour is pinned, just not by the test named for it. Give this one a legacy id-less registration so the adopted id genuinely differs.

#### F5 — NIT — the TUI registry has no cap of its own

Every queued-but-undrained steer holds an entry: 30 steers -> 30 entries. Retirement is delivery, recall, prompt-failure, or session swap — there is no size bound like the mobile side's cap-trim. Base behaves the same and real queues drain at tool boundaries, so this is bounded in practice and I am **not** asking for a change; noting it only because the registry's retirement paths were an explicit review target and this is the one that relies on external behaviour rather than on the structure.

---

### Verdict

**Changes requested — this round is NOT terminal**, on the strength of **F1** alone.

The engineering is sound and the core claim holds under independent testing: the #227 regression does not return, the mixed-population rule is structurally enforced rather than accidental, the retirement paths are covered, and the five modified tests were strengthened rather than loosened. I went looking for a swallowed regression in the text fallback and did not find one.

What blocks a clean verdict is scope accuracy, not correctness: `Closes #228` is not true for a TUI attached to a remote session, and the comments actively point a future reader away from that path. Fix `RemoteSession.prompt` or narrow the claim — either resolves it. F2–F5 are follow-ups and should not hold up a round.

1 major, 2 minor, 2 nits. No blockers.

damianvtran:

Implementer: coder (anthropic/claude-opus-4-6)
Head at remediation: `22014e04`
Scope: `ba414381..22014e04`

Thank you for the independent probes — particularly for re-deriving the #227
non-regression rather than taking the docstring's word for it, and for
mutation-testing F4 instead of assuming the assertion did work.

---

**F1 (MAJOR) — `RemoteSession.prompt` has no `message_id`, so an attached follower still swallows #228. FIXED — `933a947d`.**

Fixed rather than re-scoped, and your reasoning is the reason: the steer twin
already carries identity at `remote.py:1051` (`command_id=message.id`), so the
prompt path minting its own id via `ContinuationCommand.create` was an
oversight, not a boundary.

`prompt` now takes `message_id` and threads it to the wire as the command id,
which the owner adopts as the `Message` id and announces back. A supplied id
builds the `ContinuationCommand` directly; absent one it still calls
`.create()`, so the historical behaviour is unchanged for callers that pass
nothing.

The takeover branch forwards the id **only when the target accepts it** — the
same probe the TUI makes, because `message_id` is optional on `SessionProtocol`
and a legacy target would reject the keyword.

Verified — the seam probe the TUI actually runs, against the real class:

```
RemoteSession.prompt accepts message_id : True
  prompt(self, text, images=None, *, message_id: 'str | None' = None) -> 'None'
TUI seam probe would mint an id         : True
RESULT: PASS - follower prompts correlate by id
```

Three regression tests in `tests/unit/session/test_remote_round2.py`, all
mutation-tested (reverting `remote.py` fails them, restoring passes):

- `test_follower_prompt_carries_a_caller_supplied_message_id` — the caller's id
  rides the wire as the command id
- `test_follower_prompt_still_mints_an_id_when_none_is_supplied` — asserts a
  real UUID, since the wire validator requires one
- `test_a_takeover_target_without_the_seam_is_not_handed_an_id` — legacy target
  called the old way, modern target receives the id

**Comments corrected in the same commit.** You are right that "older or
third-party" was actively misleading with a first-party shipping class in that
bucket. All three sites now describe the real boundary: `message_id` is
*optional on `SessionProtocol`*, both in-tree sessions accept it, and the
fallback covers implementations outside the protocol's guarantee — with
`RemoteSession` named as the follower path that carried the swallow until this
commit.

---

**F2 (MINOR) — the steer-path comment claims authority "for every session implementation". FIXED — `933a947d`.**

Took the "state the actual guarantee" option rather than adding a probe, because
there is no keyword to probe for on this path: `steer_message` takes the
*Message object*, so identity is carried by the thing queued rather than offered
alongside it. A signature check could not detect a re-minting host.

The comment now states the in-tree contract (`Session` announces the queued
object; `RemoteSession` sends `command_id=message.id`), says explicitly that
this is **stronger than what `SessionProtocol` writes down**, and names the
failure mode you identified — a re-minting implementation would double-paint,
and unlike the prompt path there is no signature to catch it.

---

**F3 (MINOR) — `verify_live.py.txt` hardcodes the author's worktree. FIXED — `22014e04`.**

Agreed without reservation: evidence that only runs on my machine is not
evidence, and you having to patch it to reproduce is the proof.

The root is now derived from the file's own location
(`Path(__file__).resolve().parents[3]`), with an optional `argv[1]` override for
when the script is copied out of `docs/evidence/`. Verified by running it from
`/` with no arguments:

```
$ cd / && <repo>/.venv/bin/python <repo>/docs/evidence/echo-dedup-id/verify_live.py.txt
prompt(message_id='tui-minted-id') announced : ['tui-minted-id']
steer_message(id=0bad04f0c180...) announced  : ['0bad04f0c180...']
RESULT: PASS - the announced id is the id the TUI registered
```

---

**F4 (NIT) — tautological mobile test. ACCEPTED, no change.**

Your mutation test is correct — `test_absorb_user_event_returns_false_for_a_registered_echo`
asserts the return value while the behaviour it implies is pinned by
`test_a_phone_steer_is_not_repainted_once_assistant_rows_push_it_out` and the
image-ref upgrade test. Leaving it: it documents the `False`-means-folded
contract that `owned.py`/`tui_handle.py` read, and a redundant assertion is
cheaper than a reader inferring that contract from a row count. Noted as
genuinely redundant rather than defended as load-bearing.

---

**F5 (NIT) — the TUI registry has no size cap. ACKNOWLEDGED, no change.**

Base-identical and you did not request a change. Recording the bound for the
next reader: entries are retired on delivery, on both failed-`prompt` paths, on
recall, and wholesale on session swap, so the list tracks messages queued and
not yet drained — single digits in practice. The mobile side has a cap only
because its entries hold `TranscriptEntry` references that the transcript trim
can orphan; the TUI's hold plain strings and ids with no such coupling.

---

### Gates at `22014e04`

```
flake8 .                       clean
black --check .                633 files unchanged
isort --check .                clean
pyright .                      0 errors, 0 warnings, 0 informations
pytest tests/unit/session/test_remote_round2.py -n0    18 passed
```

Behavioural probes, all re-run at this head:

```
#228 TUI      rows after the phone's message : 2  (expected 2)   PASS
#231 mobile   rows carrying the steer text   : 1  (expected 1)   PASS
F1  follower  TUI seam probe would mint an id: True              PASS
id contract   prompt + steer round-trip vs a real Session        PASS
```

The full suite is pending a free slot — the box is under a hard RAM limit with
one pytest pool allowed at a time, and several are queued ahead of me. I will
post the full-suite result before reporting ready for round 2 rather than
claiming coverage I have not run. Everything above is single-process, which per
your own note is the better instrument here anyway.

No behavioural change to the TUI or mobile diffs you reviewed — `F1` is
confined to `session/remote.py`, and the rest is comments and evidence. The
rebase onto `ba414381` is byte-identical for `local_operator/` and `tests/`
(verified by diffing the pre- and post-rebase change sets).

damianvtran:
### Agent review — round 2

Scope: ba414381..22014e04

Independent verification of the round-1 remediation at `22014e04`. Full `base..head` range, not just the delta. I constructed the follower case myself rather than taking the author's seam-probe printout, re-ran the #227 probes from round 1, mutation-tested the new tests, and actually ran the evidence script from `/`.

**F1 is actually fixed on the follower path.** I drove `OperatorApp` with a session whose `prompt` *is* `RemoteSession.prompt` (so the TUI's `inspect.signature` probe sees the shipping class, not a stand-in). After a TUI `continue`:

- the registered id equals the `command_id` that reached the wire
- a phone's DISTINCT `continue` paints: **2 rows** (was 1)
- the TUI's own announcement then adds nothing: still **2**, registry drained to 0

The takeover branch forwards the id only when the target accepts it; a legacy target is called the old way (no `TypeError`). Both in-tree sessions take the keyword (`Session.prompt` and `RemoteSession.prompt`), so the TUI's probe now mints on every shipping path.

**The #227 regression does not come back.** Same three probes as round 1, against this head:

- far-from-tail (12 blocks between echo and event) → 1 row
- out-of-order announcement → 1/1
- two distinct steers with IDENTICAL text, newest-first → 2 rows, registry 0

The mixed-population guard still holds structurally (`if not entry.message_id` plus the mobile `owned` set). An id-carrying entry hit by a colliding foreign event, and by an id-less event, keeps its entry both times.

**Leaks.** Failed `prompt()` × 5 with ids actually handed over → registry 0. Session swap still clears. Mobile `fold_history` clears. Mobile cap-trim retires (5 unannounced steers → 0 after pushing past `PROJECTION_TRANSCRIPT_LIMIT`). Image-ref upgrade still fires in place (same object, images adopted, 1 row). Phone's own echo suppressed exactly once; a replayed announcement adds nothing.

**F2.** The steer-path comment now states the in-tree contract (`Session` announces the queued object; `RemoteSession` sends `command_id=message.id`), names that this is stronger than `SessionProtocol`, and names the re-minting failure mode. No probe was added; there is no keyword to probe for on this path, which is the right call.

**F3.** `verify_live.py.txt` now derives the root from `Path(__file__).resolve().parents[3]` with an `argv[1]` override. Ran from `/` with no arguments: `RESULT: PASS`. `repro_228` and `repro_231` still pass at head.

**Mutation.** Reverting the wire half of `RemoteSession.prompt` fails `test_follower_prompt_carries_a_caller_supplied_message_id`. Dropping the takeover signature probe fails `test_a_takeover_target_without_the_seam_is_not_handed_an_id` with the `TypeError` the probe exists to prevent. Dropping the mixed-population guard fails three tests in `test_user_echo_dedup.py`. Forcing the seam probe to always return `""` (the pre-F1 follower state) fails `test_the_prompt_path_registers_the_id_it_handed_the_session`. The tests die at the boundary they guard.

Targeted suites, single-process (`-n0`; foreign pytest was live): `test_user_echo_dedup.py` + `test_projection.py` + `test_remote_round2.py` → **77 passed**; adjacent `test_remote.py` + `test_owned.py` + `test_app_pilot.py` → **280 passed**. flake8 / black 26.1.0 / isort / pyright clean on the changed files.

CI against `22014e04`: `test (3.12)` pass, `test (3.13)` pass, lint pass, type-check pass. Did not re-report `test_cancel_backgrounded_bash_kills_process_group`.

F4 and F5 remain accepted/acknowledged with no change, as round 1 allowed.

---

### Verdict

**clean — this round is TERMINAL.** No blocker, no major. Remaining nits from round 1 are follow-ups, not a reason for another round.

0 new findings.

PR #459, opened 2026-08-30
perf(compaction): supersede repeated observations of one url or surface

damianvtran:
## What

Add an explicit, tool-declared supersede key for results that are genuine re-reads of one resource, and opt `web_fetch` into it for both fresh and cached responses.

Before this change, compaction could supersede only path-based file reads. Polling one endpoint repeatedly therefore carried every stale copy for the rest of the session.

The key is deliberately **not inferred** from ambient fields such as browser `surface_id`, URL, or selector. The browser stamps one surface handle on snapshots, logs, scrolls, clicks, and other distinct actions; inferring from it caused silent data loss during review and was replaced by the safe opt-in design.

## General harness benefit

This runs in the existing shared `prune_tool_outputs` path used by normal sessions; it is not OSWorld-specific. `web_fetch` is the only current opt-in, and scopes its key by the existing `cache_variant(raw, max_bytes)` discriminator.

Measured through the session's real prune path:

```text
10 polls of one endpoint: 5,013 -> 597 tokens (88% saved)
newest result intact: True
```

Mixed fresh and cached responses:

```text
1,960 -> 378 tokens (81% saved)
```

Raw/rendered and different `max_bytes` variants emit different keys and never blank each other.

## Safety evidence

Round 1 reproduced the unsafe inference defect before fixing it:

```text
SNAPSHOT-TREE   -> blanked
CONSOLE-ERRORS  -> blanked
SCROLLED        -> blanked
CLICKED-SUBMIT  -> retained
```

The final design requires `details["supersede_key"]` from the emitting tool. Browser actions carry no declared key, so all four distinct results survive.

Mutation checks prove both sides of the contract:

```text
benefit test vs no declared key : FAILS (good)
safety test vs surface_id guess : FAILS (good)
```

Actual `run_fetch` call-site tests prove:
- fresh fetch emits a key;
- cache hit emits the same key;
- raw/rendered variants differ;
- differing `max_bytes` variants differ.

Path and declared namespaces are separated, and the model-visible notice is resource-generic.

## Verification

Current reviewed head before the release bump:
- focused compaction + web_fetch: 88 passed locally; reviewer focused suite: 38 passed
- non-TUI unit suite: 5,176 passed, 3 skipped locally
- independent reviewer full suite: 9,013 passed, 15 skipped, 1 load-related background-process cancellation flake that passed in isolation
- flake8: clean
- Black 26.1.0: clean
- isort 5.13.2: clean
- pyright: 0 errors

## Review rounds

- Round 1 found one blocker and three majors; all remediated.
- Round 2 confirmed the blocker resolved and found one remaining major in the cache-hit path; fixed and pinned by real call-site tests.
- Round 3 is clean and terminal on the code head: no blocker or major findings remain.

The final version-only head receives a fresh review before merge.

### DIFF
--- a/local_operator/compaction/pruning.py
+++ b/local_operator/compaction/pruning.py
@@ -187,10 +187,17 @@ def _supersede_key(message: Message) -> str | None:
         # navigated elsewhere is a different resource, and a selector-scoped
         # read is not the whole page. Without this a read of the footer would
         # blank an earlier read of the article body on the same tab.
-        scope = url if isinstance(url, str) and url else ""
+        # The parts are LENGTH-PREFIXED rather than joined by a separator.
+        # A bare separator collides on real input: url "https://a#body" with no
+        # selector and url "https://a" with selector "body" are different reads
+        # that would otherwise produce one key, so the newer would blank the
+        # older and lose content the model still needs. No delimiter can appear
+        # inside a length, so this is unambiguous for every input. Absent and
+        # None still share a key, which is correct -- they are the same absence.
+        scope_url = url if isinstance(url, str) else ""
         selector = details.get("selector")
-        if isinstance(selector, str) and selector:
-            scope = f"{scope}#{selector}"
+        scope_selector = selector if isinstance(selector, str) else ""
+        scope = f"{len(scope_url)}:{scope_url}{len(scope_selector)}:{scope_selector}"
         return f"{message.tool_name}:surface:{surface}:{scope}"
     if isinstance(url, str) and url:
         return f"{message.tool_name}:url:{url}:{details.get('range') or 'full'}"
--- a/local_operator/compaction/pruning.py
+++ b/local_operator/compaction/pruning.py
@@ -153,17 +153,20 @@ def _supersede_key(message: Message) -> str | None:
     range of the same file (they share not one line), while a full-file read
     supersedes every earlier range — handled in the pass below.
 
-    ``details['surface']`` is the same idea for tools that OBSERVE rather than
-    read: a screen, a page, a canvas. Each result describes the live state of
-    one surface, so a newer observation of the same surface supersedes every
-    older one completely -- the older text describes a screen that no longer
-    exists, exactly as a stale screenshot does.
+    ``details['supersede_key']`` lets a tool opt a result in explicitly, for
+    resources a path does not name: a URL, a page, a live surface. It is an
+    OPT-IN the emitting tool states, deliberately not inferred from ambient
+    fields like a url or a surface handle. Inference is unsafe here because a
+    single handle covers many DIFFERENT results: the browser tool stamps its
+    ``surface_id`` on every action, so keying on it made a 40-character
+    ``click`` blank the accessibility snapshot and console errors the agent had
+    just gathered to decide what to click. Only the tool knows which of its
+    results are re-reads of one thing and which are distinct observations of
+    it, so only the tool may say so.
 
-    Without this, a screen-driving agent accumulates every observation's text
-    forever. Measured on an OSWorld 2.0 episode: 374 tool results held ~71.5k
-    tokens of superseded observation text at cycle 371, most of it the task
-    instruction repeated verbatim in every single result, and the context had
-    grown to 133k with the compaction trigger still correctly far away.
+    The value must therefore identify the CONTENT, including any variant that
+    changes what comes back (a raw versus rendered fetch of one URL is two
+    different answers, not one superseding the other).
     """
     details = _details_of(message)
     if not message.tool_name:
@@ -171,36 +174,9 @@ def _supersede_key(message: Message) -> str | None:
     path = details.get("path")
     if isinstance(path, str) and path:
         return f"{message.tool_name}:{path}:{details.get('range') or 'full'}"
-    # A URL identifies a re-fetchable resource exactly as a path identifies a
-    # file: ``read <url>`` and ``web_fetch`` of the same address return the
-    # live state of one thing, so the newer answer supersedes the older.
-    url = details.get("url")
-    # A SURFACE is the same idea for tools that OBSERVE rather than fetch: a
-    # browser tab, a screen, a canvas. Each result describes the live state of
-    # that surface, so a newer view supersedes every older one -- the older
-    # text describes a page that no longer exists, exactly as a stale
-    # screenshot does. The browser tool names its handle ``surface_id``;
-    # ``surface`` is the generic spelling for other observers.
-    surface = details.get("surface") or details.get("surface_id")
-    if isinstance(surface, str) and surface:
-        # Scope to what is being observed THROUGH the surface: a tab that has
-        # navigated elsewhere is a different resource, and a selector-scoped
-        # read is not the whole page. Without this a read of the footer would
-        # blank an earlier read of the article body on the same tab.
-        # The parts are LENGTH-PREFIXED rather than joined by a separator.
-        # A bare separator collides on real input: url "https://a#body" with no
-        # selector and url "https://a" with selector "body" are different reads
-        # that would otherwise produce one key, so the newer would blank the
-        # older and lose content the model still needs. No delimiter can appear
-        # inside a length, so this is unambiguous for every input. Absent and
-        # None still share a key, which is correct -- they are the same absence.
-        scope_url = url if isinstance(url, str) else ""
-        selector = details.get("selector")
-        scope_selector = selector if isinstance(selector, str) else ""
-        scope = f"{len(scope_url)}:{scope_url}{len(scope_selector)}:{scope_selector}"
-        return f"{message.tool_name}:surface:{surface}:{scope}"
-    if isinstance(url, str) and url:
-        return f"{message.tool_name}:url:{url}:{details.get('range') or 'full'}"
+    declared = details.get("supersede_key")
+    if isinstance(declared, str) and declared:
+        return f"{message.tool_name}:declared:{declared}"
     return None
 
 
--- a/local_operator/compaction/pruning.py
+++ b/local_operator/compaction/pruning.py
@@ -7,10 +7,10 @@ details. The flags never reach provider wire formats.
 
 Two passes (``pruneSupersededToolResults``):
 
-(a) **Superseded reads** — a later tool result for the same path makes the
-    earlier output dead weight (the model re-read precisely because the old
-    copy was stale). The supersede key is ``details['path']``; results without
-    a path group by tool name.
+(a) **Superseded results** — a later tool result for the same path, or with
+    the same explicitly declared ``details['supersede_key']``, makes the
+    earlier output dead weight. Results with neither key are exempt: grouping
+    pathless results by tool name would collapse legitimately distinct output.
 (b) **Useless-flagged results** — tools self-flag contextually worthless
     output (zero-match search, timed-out wait); blanked once consumed.
 
@@ -55,7 +55,7 @@ __all__ = [
 MIN_PRUNE_TOKENS = 50
 
 #: Exact placeholder written over a superseded tool result.
-SUPERSEDED_NOTICE = "[Superseded by a newer read of this file]"
+SUPERSEDED_NOTICE = "[Superseded by a newer result for the same resource]"
 
 #: Exact placeholder written over an elided useless tool result.
 USELESS_NOTICE = "[Uneventful result elided]"
@@ -173,7 +173,9 @@ def _supersede_key(message: Message) -> str | None:
         return None
     path = details.get("path")
     if isinstance(path, str) and path:
-        return f"{message.tool_name}:{path}:{details.get('range') or 'full'}"
+        # The namespace marker prevents a path crafted like a declared key
+        # from colliding with a tool's explicit opt-in contract.
+        return f"{message.tool_name}:path:{path}:{details.get('range') or 'full'}"
     declared = details.get("supersede_key")
     if isinstance(declared, str) and declared:
         return f"{message.tool_name}:declared:{declared}"
--- a/local_operator/compaction/pruning.py
+++ b/local_operator/compaction/pruning.py
@@ -152,11 +152,48 @@ def _supersede_key(message: Message) -> str | None:
     The range rides in the key: a ranged read must NOT supersede a different
     range of the same file (they share not one line), while a full-file read
     supersedes every earlier range — handled in the pass below.
+
+    ``details['surface']`` is the same idea for tools that OBSERVE rather than
+    read: a screen, a page, a canvas. Each result describes the live state of
+    one surface, so a newer observation of the same surface supersedes every
+    older one completely -- the older text describes a screen that no longer
+    exists, exactly as a stale screenshot does.
+
+    Without this, a screen-driving agent accumulates every observation's text
+    forever. Measured on an OSWorld 2.0 episode: 374 tool results held ~71.5k
+    tokens of superseded observation text at cycle 371, most of it the task
+    instruction repeated verbatim in every single result, and the context had
+    grown to 133k with the compaction trigger still correctly far away.
     """
     details = _details_of(message)
+    if not message.tool_name:
+        return None
     path = details.get("path")
-    if isinstance(path, str) and path and message.tool_name:
+    if isinstance(path, str) and path:
         return f"{message.tool_name}:{path}:{details.get('range') or 'full'}"
+    # A URL identifies a re-fetchable resource exactly as a path identifies a
+    # file: ``read <url>`` and ``web_fetch`` of the same address return the
+    # live state of one thing, so the newer answer supersedes the older.
+    url = details.get("url")
+    # A SURFACE is the same idea for tools that OBSERVE rather than fetch: a
+    # browser tab, a screen, a canvas. Each result describes the live state of
+    # that surface, so a newer view supersedes every older one -- the older
+    # text describes a page that no longer exists, exactly as a stale
+    # screenshot does. The browser tool names its handle ``surface_id``;
+    # ``surface`` is the generic spelling for other observers.
+    surface = details.get("surface") or details.get("surface_id")
+    if isinstance(surface, str) and surface:
+        # Scope to what is being observed THROUGH the surface: a tab that has
+        # navigated elsewhere is a different resource, and a selector-scoped
+        # read is not the whole page. Without this a read of the footer would
+        # blank an earlier read of the article body on the same tab.
+        scope = url if isinstance(url, str) and url else ""
+        selector = details.get("selector")
+        if isinstance(selector, str) and selector:
+            scope = f"{scope}#{selector}"
+        return f"{message.tool_name}:surface:{surface}:{scope}"
+    if isinstance(url, str) and url:
+        return f"{message.tool_name}:url:{url}:{details.get('range') or 'full'}"
     return None
 
 
--- a/local_operator/tools/builtin.py
+++ b/local_operator/tools/builtin.py
@@ -2300,7 +2300,16 @@ async def execute_read(
                 "read",
                 f"Cannot resolve '{target}': the resolver does not handle this URL.",
             )
-        return _text(tool_call_id, "read", content, details={"url": target})
+        # A URL read is a re-read of one resource: a later read of the same
+        # URL describes its newer state, so the older text is dead weight.
+        # Declared explicitly (never inferred) because only this call site
+        # knows the result is the whole resource rather than one view of it.
+        return _text(
+            tool_call_id,
+            "read",
+            content,
+            details={"url": target, "supersede_key": target},
+        )
 
     cwd = _safe_cwd(context)
     path, inside, resolvable = _resolve_workspace_path(target, cwd)
--- a/local_operator/tools/builtin.py
+++ b/local_operator/tools/builtin.py
@@ -2300,16 +2300,13 @@ async def execute_read(
                 "read",
                 f"Cannot resolve '{target}': the resolver does not handle this URL.",
             )
-        # A URL read is a re-read of one resource: a later read of the same
-        # URL describes its newer state, so the older text is dead weight.
-        # Declared explicitly (never inferred) because only this call site
-        # knows the result is the whole resource rather than one view of it.
-        return _text(
-            tool_call_id,
-            "read",
-            content,
-            details={"url": target, "supersede_key": target},
-        )
+        # Deliberately NO supersede_key here. This path serves internal URLs
+        # (skill://, guide://, mcp://), and skill reads are exempt from pruning
+        # anyway -- ``_is_prunable`` protects them because a pruned skill just
+        # gets re-read in a loop. Declaring a key would be inert for those and
+        # would add avoidable risk for the rest, since a resolver result is not
+        # always the same content under the same URL.
+        return _text(tool_call_id, "read", content, details={"url": target})
 
     cwd = _safe_cwd(context)
     path, inside, resolvable = _resolve_workspace_path(target, cwd)
--- a/local_operator/web_fetch/tool.py
+++ b/local_operator/web_fetch/tool.py
@@ -238,6 +238,7 @@ def _cache_hit_result(
     ttl_seconds: int,
     tool_name: str,
     context: ToolContext | None,
+    variant: str | None = None,
 ) -> tuple[str, dict[str, Any]] | None:
     """Rebuild a result from a cache entry, or ``None`` when the hit is unusable.
 
@@ -278,7 +279,12 @@ def _cache_hit_result(
         "ok": http_ok,
         "http_error": not http_ok,
     }
-    preview, details, _handle = _shape_from_content(result_like, content, tool_name, context)
+    # Same declared supersede key as a fresh fetch: a cached re-read describes
+    # the same resource, so it must group with the fresh ones rather than
+    # escaping the supersede pass for having been served from cache.
+    preview, details, _handle = _shape_from_content(
+        result_like, content, tool_name, context, variant=variant
+    )
     return preview, details
 
 
@@ -347,7 +353,9 @@ async def run_fetch(
     if not refresh and settings.cache_ttl_seconds > 0:
         entry = read_cache_entry(normalized, variant)
         if entry is not None:
-            hit = _cache_hit_result(entry, settings.cache_ttl_seconds, tool_name, context)
+            hit = _cache_hit_result(
+                entry, settings.cache_ttl_seconds, tool_name, context, variant=variant
+            )
             if hit is not None:
                 preview, details = hit
                 return preview, details, False
--- a/local_operator/web_fetch/tool.py
+++ b/local_operator/web_fetch/tool.py
@@ -186,6 +186,7 @@ def _shape_from_content(
     content: str,
     tool_name: str,
     context: ToolContext | None,
+    variant: str | None = None,
 ) -> tuple[str, dict[str, Any], str | None]:
     """Build ``(preview_text, details, spill_handle)`` from full ``content``.
 
@@ -209,6 +210,14 @@ def _shape_from_content(
     body, spill_details = spill_truncate(content, tool_name, context)
     preview = f"{header}\n\n{body}" if body else header
     details = dict(result_like)
+    # A re-fetch of the same URL in the same rendition supersedes this result:
+    # both describe the live state of one resource. Scoped by ``cache_variant``
+    # for the same reason the cache is -- a raw fetch and a rendered fetch of
+    # one URL are different answers, and a byte-capped one is not the full
+    # body, so neither may blank the other.
+    url_for_key = result_like.get("final_url") or result_like.get("url")
+    if isinstance(url_for_key, str) and url_for_key and variant:
+        details["supersede_key"] = f"{url_for_key}\x00{variant}"
     details["lines"] = content.count("\n") + 1 if content else 0
     details["preview_chars"] = len(preview)
     if spill_details is not None:
@@ -382,7 +391,9 @@ async def run_fetch(
         "ok": http_ok,
         "http_error": not http_ok,
     }
-    preview, details, handle = _shape_from_content(result_like, result.content, tool_name, context)
+    preview, details, handle = _shape_from_content(
+        result_like, result.content, tool_name, context, variant=variant
+    )
 
     # Record the cache entry pointing at the spill handle (metadata only). Only
     # 2xx responses are cached (M4): a 4xx/5xx cached for the full TTL would
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "local-operator"
-version = "0.44.7"
+version = "0.44.8"
 description = "A team of proactive AI assistants that can work together behind the scenes to help you get mundane tasks done so you can focus on the fun stuff."
 readme = "README.md"
 authors = [{ name = "Damian Tran", email = "damian@gominerva.com" }]
--- a/tests/unit/compaction/test_pruning.py
+++ b/tests/unit/compaction/test_pruning.py
@@ -6,6 +6,7 @@ from local_operator.compaction.pruning import (
     MIN_PRUNE_TOKENS,
     SUPERSEDED_NOTICE,
     USELESS_NOTICE,
+    _supersede_key,
     compute_suffix_tokens,
     prune_tool_outputs,
 )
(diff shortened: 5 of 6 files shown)


damianvtran:
### Agent review — round 1
Reviewer: reviewer subagent (claude-opus-4-6)
Scope: ba414381c4dfeafcea493fbeaac6205a8d312a14..5e982a02a287f2fc8a69e657cb701476a6cc536d

Reviewed the committed head only. Gates were run over the whole tree with the worktree reset to the head commit (see "Gates" below).

**Note on worktree state:** `/tmp/lop-harness-perf` carries uncommitted changes to both files that are *not* on the PR head — a length-prefixed scope key and a third test. Those are unreviewed and unpushed; F1 below is written against the committed head, which is what CI, the PR diff and a merge would take. If the WIP is the intended fix it needs to be committed and pushed, and it still does not address F2/F3.

---

**F1 — blocker — `local_operator/compaction/pruning.py:180-195`: the surface key blanks distinct browser results, because it keys on the surface rather than on the observation.**

The premise in the docstring — "each result describes the live state of one surface" — does not hold for the `browser` tool. `surface_id` is emitted by *every* browser action, not just observing ones: `goto` (builtin.py:5376, 5428), `read` (6193), `snapshot` (6209), `scroll` (6229), `logs` (6232), `click` (6239), `type` (6255), `tabs` (6128). Most carry the same `url` and no `selector`, so they collapse to one key and the newest blanks all the others.

Reproduced against the committed head with a realistic bridge flow on one tab at one URL (snapshot → logs → scroll → click):

```
changed: True
  1: '[Superseded by a newer r'      <- SNAPSHOT-TREE   (lost)
  2: '[Superseded by a newer r'      <- CONSOLE-ERRORS  (lost)
  3: '[Superseded by a newer r'      <- SCROLLED        (lost)
  4: 'CLICKED-SUBMIT ddddddddd'
```

A `click` result — 40 characters of "Clicked #submit. Page: …" — destroys the accessibility snapshot and the console errors the agent gathered to decide what to click. That is exactly the silent data loss the PR says it avoids: the surviving result does *not* contain the superseded content, and the notice ("Superseded by a newer read of this file") tells the model nothing was lost.

Four distinct collision classes, all reproduced on the head commit:

```
LOSS fragment vs selector:                 1161 chars blanked
LOSS snapshot vs tabs-listing (no url):    1161 chars blanked
LOSS logs vs click, same url no selector:  1161 chars blanked
LOSS web_fetch raw vs rendered:            1161 chars blanked
```

Per the specific questions asked:

- **`#` separator is ambiguous with a URL fragment** (question 2 — confirmed, this is the WIP's target): `{"url": "https://a#body"}` and `{"url": "https://a", "selector": "body"}` both key to `browser:surface:s:1:https://a#body`.
- **Absent url and absent selector both degrade to `""`**, so `tabs` (`{"tab_count": 3, "surface_id": "s:1"}`, no url) and a cmux `snapshot` (`{"surface_id": "s:1"}`) share `browser:surface:s:1:` — a tab listing blanks a page snapshot.
- **`selector` is None vs "" vs missing** are all treated as absent, which is right, and the WIP test asserts it.
- **The real defect is broader than the separator**: even with a perfect separator, `read`+`logs`+`click`+`scroll`+`goto` on one URL with no selector all share a key. The scope must include *what kind of observation this is* — the action or a per-action discriminator — not just url+selector. A minimal fix is to restrict the surface branch to the observing actions that actually return page content (`read`, `snapshot`) and require a non-empty url, letting everything else return None.

**F2 — major — `local_operator/compaction/pruning.py:196-197`: `web_fetch` results at one URL are not interchangeable, so keying on url alone blanks distinct content.**

`web_fetch` folds `raw` and `max_bytes` into its *cache* key (`cache_variant(raw=raw, max_bytes=max_bytes)`, web_fetch/tool.py:35) precisely because they change what comes back — a raw fetch and a rendered fetch of one URL are different documents. `details` carries `url`/`final_url` but not the variant, so both key to `web_fetch:url:<u>:full` and the newer blanks the older. Same for the POST-like case raised in the brief: any endpoint whose response depends on state returns different content at one address, and the older answer may be the one the model still needs to compare against.

The `read` sugar has the same shape: `read spill://h?q=alpha` and `read spill://h?q=beta` are different queries and *do* stay distinct only because the query rides inside the handle string (builtin.py:2018) — that is luck, not design, and it does not save the `raw` case.

**F3 — major — `tests/unit/compaction/test_pruning.py:311-347`: the guard test does not discriminate, so it passes while the code loses data.**

`test_distinct_observations_are_never_superseded` is the test carrying the PR's central safety claim ("blanking those would be data loss dressed up as an optimization"), but every case it checks varies `url` or `selector` while holding everything else fixed — the two fields already in the key. It never varies the *action*, which is the axis F1 breaks on. All four F1 collision cases pass through it undetected: I ran them against the committed head and they blank 1161 chars each, with the guard test green.

`test_re_observing_one_resource_supersedes_the_earlier_view` does discriminate — verified by running the new tests against `origin/main`'s `pruning.py`: `1 failed, 16 passed`, failing on `assert changed`. So the benefit test is real; the safety test is the vacuous one, which is the wrong way round for a change whose stated risk is false blanking.

**F4 — major — `local_operator/compaction/pruning.py:157-166`: the docstring's headline key `details['surface']` is never emitted by any tool, and the OSWorld measurement is not reproducible from this repo.**

The docstring leads with "``details['surface']`` is the same idea…" and the code reads `details.get("surface") or details.get("surface_id")`. Nothing in `local_operator/` emits `details["surface"]` — grep finds only TUI theme colours. The generic spelling is dead, and documenting it first makes the mechanism read as broader than it is.

On the measured numbers you asked me to check: "374 tool results held ~71.5k tokens of superseded observation text at cycle 371" implies ~191 tokens of *recoverable* text per result. The claim is unverifiable from the repo — there is no OSWorld artifact under `docs/evidence/` or anywhere else — and it is attributed to an environment this codebase does not contain. The attached "most of it the task instruction repeated verbatim in every single result" also describes a *prompt-construction* problem, not superseded observation text, so the number does not clearly measure what this change fixes. Comments asserting measured numbers should point at a reproducible artifact or drop the figure.

**F5 — minor — `local_operator/compaction/pruning.py:12-13` and `:58`: module docstring and the user-visible notice both still say "file".**

The module docstring still states "The supersede key is ``details['path']``; results without a path group by tool name" — now wrong twice over. And `SUPERSEDED_NOTICE = "[Superseded by a newer read of this file]"` is what the model actually reads in the transcript; for a blanked URL fetch or browser observation "this file" is misleading about what was dropped.

**F6 — minor — `tests/unit/compaction/test_pruning.py:1-3`: the added import silently destroyed the module docstring.**

`from typing import Any` was inserted *above* the docstring, so it is now a no-op string expression: `ast.get_docstring(module)` returns `None`. Move the import below.

**F7 — nit — `local_operator/compaction/pruning.py:196`: url reads do not get the full-supersedes-ranged rule that path reads get.**

Answering question 3: the url branch does not break `_supersede_path_range` — that helper only fires on `path` (line 209), so url results return None and never enter the coverage or `seen_full_paths` logic. Verified: a full `read` of a URL after a ranged one leaves the ranged result live, where the equivalent path reads would supersede. Conservative, not lossy, so it is fine to leave — worth a comment noting the asymmetry is deliberate.

On question 4 (ordering): path → surface → url is correct as written. The only result carrying both is `screenshot` (builtin.py:6287), which emits `{"path": target, "bytes": …}` and no `surface_id`, so the branches do not actually contend; a screenshot keys on its file path, which is right.

No further blockers beyond F1. No findings at nit severity beyond F7.

---

**Gates** (whole tree, worktree reset to head 5e982a02, commands from `.github/workflows/ci.yml`):

| Gate | Result |
|---|---|
| `flake8 .` | clean (rc=0) |
| `black --check .` (black==26.1.0 via uvx) | 633 files unchanged |
| `isort --check .` (isort==5.13.2) | clean |
| `pyright` | 0 errors, 0 warnings, 0 informations |
| `pytest tests/unit` | **9012 passed, 15 skipped** in 451s |

The full unit suite is green — including the three TUI tests the PR body reports as failing, which passed here. Green gates are not evidence against F1: the suite passes because no test exercises two different browser *actions* on one surface.

**Verdict: changes requested.** One blocker (F1) and three majors (F2, F3, F4). The core idea is sound and the benefit test proves the win is real, but the key is too coarse for the `browser` tool's actual details shape, and the test that was supposed to catch that does not vary the axis it breaks on.

damianvtran:

Head: `474a405e1ebd4d59692a15fb445343af4662e035`

Thank you — F1 was a real defect and I reproduced it before fixing. A 40-character `click` did destroy the snapshot and console errors:

```
1: '[Superseded...'  <- SNAPSHOT-TREE   (lost)
2: '[Superseded...'  <- CONSOLE-ERRORS  (lost)
3: '[Superseded...'  <- SCROLLED        (lost)
4: 'CLICKED-SUBMIT ...'
```

**F1 (blocker) — fixed** in `474a405e`. The root cause was inference itself, not the separator. Supersede is now an explicit **opt-in**: `details['supersede_key']` names the content, and nothing is derived from ambient fields like `surface_id`. Only the emitting tool knows which of its results are re-reads of one thing versus distinct observations of it. The same flow now leaves all four results intact.

**F2 (major) — fixed** in `474a405e`. `web_fetch` scopes its declared key by `cache_variant(raw=..., max_bytes=...)` — the discriminator its cache already uses. Your diagnosis was more right than my first attempt: I had initially read the variant off `result_like`, where it was never present, so the key would silently never activate. It is now threaded explicitly from the call site that computes it. Verified: raw and rendered fetches of one URL both survive; six identical rendered fetches still save 80%.

**F3 (major) — fixed** in `474a405e`. You were right that the guard test was the vacuous one. It now varies the **action** — snapshot / console / scroll / click on one surface — which is the axis F1 broke on. Both tests verified against their wrong answers:

```
benefit test vs no declared key : FAILS (good)
safety  test vs surface_id guess: FAILS (good)
```

The safety test specifically fails if the unsafe `surface_id` inference is reintroduced, so this defect cannot silently return.

**F4 (major) — fixed** in `474a405e`. The `details['surface']` claim is gone; nothing emitted it. The docstring no longer cites the unverifiable OSWorld figure, and instead states the constraint that matters: why inference is unsafe and why the opt-in exists. Two tools opt in today — `web_fetch` (variant-scoped) and `read <url>`.

**On the uncommitted worktree changes** — that was my in-flight length-prefix fix for the separator half. It is now committed as `8b122094`, then superseded by the opt-in design in `474a405e`. Thank you for reviewing the committed head rather than the dirty tree, and for restoring it byte-exact.

**Benefit retained**, measured through the session's own prune path:

```
session prune path: 5,013 -> 597 tokens (88% saved)
newest intact: True
```

**Gates**: flake8, black 26.1.0, isort 5.13.2, pyright (0 errors) all clean; `tests/unit` (non-TUI) 5,174 passed, 0 failed.

Ready for round 2.

damianvtran:
### Agent review — round 2
Reviewer: reviewer subagent (claude-opus-5)
Scope: ba414381c4dfeafcea493fbeaac6205a8d312a14..474a405e1ebd4d59692a15fb445343af4662e035

Reviewed the committed head in a clean detached worktree (`/tmp/lop-review-r2` at `474a405e`, `git status` empty). **`/tmp/lop-harness-perf` is dirty again** — it carries uncommitted edits to `builtin.py` and `web_fetch/tool.py` that are not on the PR head, and they change two of the exact things I was asked to verify (they add `variant=` to `_cache_hit_result` and *remove* the `read <url>` declared key). Everything below is against the committed head, which is what CI, the diff and a merge would take. That WIP is unreviewed and unpushed; see F2.

---

**F1 (round-1 blocker) — RESOLVED.** The opt-in design is correct and I could not break it.

Verified by mutation rather than by reading. `browser` results now key to `None` regardless of action, and the four round-1 collision classes are gone. I reintroduced the unsafe inference on top of the head:

```python
surface = details.get("surface_id")
if isinstance(surface, str) and surface:
    return f"{message.tool_name}:surface:{surface}:{details.get('url') or ''}"
```

```
FAILED test_a_declared_key_never_blanks_a_different_result
AssertionError: a browser surface handle is stamped on every action, so it must not group them
assert 184 == 3687
```

The regression is caught, so it cannot silently return. Key discrimination across the reachable surface — no collisions:

```
read file full        -> 'read:/a/b.py:full'
read file ranged      -> 'read:/a/b.py:1-10'
read skill/guide/mcp  -> 'read:declared:skill://x' / 'guide://x' / 'mcp://s/t'   (distinct)
web_fetch rendered    -> 'web_fetch:declared:https://a\x00raw=0|max_bytes=default'
web_fetch raw         -> 'web_fetch:declared:https://a\x00raw=1|max_bytes=default' (distinct)
browser any action    -> None
bash                  -> None
```

On the specific `read <url>` question: **the bare-URL key is safe, because that branch is narrower than it looks.** It is only reachable for *internal* resolver URLs — `http(s)://` returns earlier via `run_fetch` (builtin.py:2270) and `spill://` returns earlier still via `_read_spill` (builtin.py:2258), both before the resolver branch. So the `?q=` and ranged-read cases raised in the brief never reach it; I confirmed each returns `key=None` through the real `execute_read`:

```
spill ranged 1-20   details={'url': 'spill://4194…', 'range': '1-20'}   key=None
spill ranged 50-70  details={'url': 'spill://4194…', 'range': '50-70'}  key=None
spill ?q=alpha      details={'url': 'spill://4194…'}                    key=None
spill ?q=beta       details={'url': 'spill://4194…'}                    key=None
```

`skill://` is separately exempt via `_is_prunable` (pruning.py:133-138); `guide://` and `mcp://` return whole stable documents where a later read genuinely is the newer state. No remaining path gives two distinct results one key.

---

**F2 — major — `local_operator/web_fetch/tool.py:303` (`_cache_hit_result`): the call site that does not thread `variant` is the one that carries the PR's headline benefit, so on the real fetch path `web_fetch` supersede almost never fires.**

Round-1 F2 is fixed *for the key's safety* — raw and rendered no longer collide (verified above, and the test fails if I truncate the key at the NUL). But the threading is incomplete in the direction that matters. `run_fetch` passes `variant=` at line 403 (fresh fetch) and **not** at line 350 → `_cache_hit_result` → `_shape_from_content(result_like, content, tool_name, context)` with `variant` defaulted to `None`. The guard `and variant` at line 219 then declines to emit a key at all.

Its absence is **a missing benefit, not an unsafe key** — but the benefit it misses is the motivating scenario. `cache_ttl_seconds` defaults to 900 (`web_fetch/models.py:62`), so when an agent polls one endpoint the *first* fetch is a miss and every subsequent one inside 15 minutes is a hit with no key. Measured end-to-end through the real `run_fetch` and the session's own `prune_tool_outputs`, against a local HTTP server:

```
  poll 0: cache=miss  key=YES
  poll 1: cache=hit   key=NO
  poll 2: cache=hit   key=NO
  poll 3: cache=hit   key=NO
  poll 4: cache=hit   key=NO
  poll 5: cache=hit   key=NO

REAL PATH, 6 polls of one endpoint inside the 900s TTL:
  before=12164 after=12164 changed=False saved=0%
```

Zero. The PR body reports this exact scenario as `web_fetch same endpoint x6: 1,814 -> 354 (80%)`, and the round-1 remediation repeats "six identical rendered fetches still save 80%" — both measured on hand-built `details` dicts, not through `run_fetch`, so neither exercises the cache-hit branch that real polling takes. Fast polling (the case the PR is for) saves nothing; only polls spaced more than 15 minutes apart supersede, which is backwards from the motivation.

The comment at tool.py:213 ("A re-fetch of the same URL in the same rendition supersedes this result") is correspondingly untrue on this path.

One line fixes it: pass `variant` into `_cache_hit_result` and on to `_shape_from_content`. A cached re-read describes the same resource in the same rendition, so it must group with the fresh ones rather than escape the pass for having been served from cache. I note the uncommitted WIP in `/tmp/lop-harness-perf` already does exactly this — it needs to be committed, pushed and re-measured through `run_fetch`, not through synthetic messages.

---

**F3 (round-1 major) — RESOLVED.** Both tests discriminate, and the safety test does cover the action axis.

I tried three ways to make them pass vacuously; all three were caught:

| Mutation applied to head | Result |
|---|---|
| Delete the `declared` branch entirely (null change) | `test_a_declared_key_supersedes_re_reads_of_one_resource` FAILS (`assert False` at :305) |
| Reintroduce `surface_id` inference (the F1 defect) | `test_a_declared_key_never_blanks_a_different_result` FAILS at :332 — **the action axis** |
| Truncate the declared key at `\x00` (collapse raw/rendered) | same test FAILS at :345 — **the variant axis** |

The safety test now varies snapshot/console/scroll/click on one surface holding url and selector fixed, which is precisely the axis round 1 broke on. `_observer` builds `provider_payload={"details": …}`, matching how `loop.py:1904` actually delivers `ToolResult.details`, so the fixture is faithful rather than a shape invented for the test.

**F4 (round-1 major) — RESOLVED.** The `details['surface']` claim and the unverifiable OSWorld figure are both gone from the `_supersede_key` docstring, replaced by the constraint that actually governs the design (why inference is unsafe, why only the emitting tool may declare). The new text is accurate to the code.

---

**F5 — minor — `local_operator/compaction/pruning.py:11-13` and `:58`: round-1 F5 was not addressed and is now wrong in a third way.**

The module docstring still reads "The supersede key is ``details['path']``; results without a path group by tool name" — the second clause was already false on `main`, and the first is now false too since a declared key is a second source. `SUPERSEDED_NOTICE = "[Superseded by a newer read of this file]"` is the string the model actually reads in the transcript; for a blanked URL fetch "this file" misdescribes what was dropped. Not claimed as fixed in the round-1 remediation, so flagging it once more rather than treating it as regressed.

**F6 — minor — `tests/unit/compaction/test_pruning.py:1-3`: round-1 F6 was not addressed; the module docstring is still destroyed.**

`from typing import Any` still sits above the docstring, so it remains a no-op string expression. Confirmed on the head:

```
>>> ast.get_docstring(ast.parse(open('tests/unit/compaction/test_pruning.py').read()))
None
```

Moving the import below the docstring fixes it.

**F7 — minor — no test at either opt-in call site asserts a key is emitted, which is why F2 survived.**

`grep -rn supersede tests/unit/web_fetch/ tests/unit/tools/` returns nothing. Every supersede test builds `details` by hand, so a call site that fails to emit the key is invisible to the suite — exactly what happened on the cache-hit path. One assertion on `details["supersede_key"]` in the existing `test_cache_hit_makes_no_network_call` (test_tool.py:92) would have caught F2 and would pin the contract the pruner depends on.

**F8 — nit — `local_operator/compaction/pruning.py:180-185`: the `declared` and `path` namespaces are not separated, so they can in principle collide.**

`f"{tool}:{path}:{range}"` and `f"{tool}:declared:{key}"` share one string space: a path of `declared:/etc/passwd` read full and a declared key of `/etc/passwd:full` both produce `read:declared:/etc/passwd:full`. Unreachable in practice (a declared key only reaches that branch for a `://` URL), so this is hygiene, not a defect — a distinct prefix such as `:decl#` would close it.

---

**Gates** (whole tree, clean worktree at head `474a405e`, commands from `.github/workflows/ci.yml`):

| Gate | Result |
|---|---|
| `flake8 .` | clean (rc=0) |
| `black --check .` (black==26.1.0 via uvx) | 633 files unchanged |
| `isort --check .` (isort==5.13.2) | clean |
| `pytest tests/unit` | 9011 passed, 1 failed, 15 skipped in 395s |

The one failure is `tests/unit/test_tui_responsiveness.py::test_second_title_backfill_pass_is_stat_only` (`second pass took 118.3 ms`, threshold 20 ms). It passes in isolation (`1 passed in 5.08s`), touches no compaction code, and is a load-timing flake unrelated to this PR. GitHub checks were still pending at review time.

**Verdict: changes requested.** No blocker — the round-1 blocker is genuinely and durably fixed, and the opt-in design is the right one. One major remains (F2): the feature's headline case, repeated fetches of one endpoint, saves 0% on the real path because cache hits never get a key, and the PR's 80% evidence does not exercise that path. It is a one-line thread-through plus a re-measurement through `run_fetch`. F5-F8 are minors/nits and need not gate a further round once F2 is fixed.

damianvtran:

Head: `436897a938da193407d708db4698ff40fb7d16ae`

**F2 (major) — fixed** in `3cd917fe`, now pinned by real call-site tests in `436897a9`. The cache-hit path now receives the same `cache_variant(raw, max_bytes)` as the fresh path, so cached and fresh results emit the same declared key. Verified through the real `run_fetch` path rather than hand-built details: fresh fetch and cache hit emit the identical key; raw/rendered and different `max_bytes` variants emit different keys. Mixed fresh + cached polls measured 1,960 → 378 tokens (81%).

**F5 (minor) — fixed** in `ac3141da`. The module docstring now describes both path and explicitly declared supersede keys accurately. The model-visible notice is resource-generic: `[Superseded by a newer result for the same resource]`.

**F6 (minor) — fixed** in `ac3141da`. The test module docstring is restored by moving the import below it.

**F7 (minor) — fixed** in `436897a9`. Added tests at the actual `web_fetch` emission sites. The tests fail if either the fresh-path or cache-hit variant threading is removed, and assert key separation for raw/rendered and different byte caps.

**F8 (nit) — fixed** in `ac3141da`. Path and declared namespaces are separated unambiguously; a focused regression test covers the collision.

I also removed the `read <url>` opt-in in `3cd917fe` after proving it was inert for the intended internal skill case: `_is_prunable` deliberately exempts skill reads because pruning them causes re-read loops. `web_fetch` is now the sole opt-in and the only one with measured benefit.

Verification on current head:

- focused compaction + web_fetch: **88 passed**
- non-TUI unit suite: **5,176 passed, 3 skipped**
- flake8: clean
- black 26.1.0: clean
- isort 5.13.2: clean
- pyright: 0 errors

The unrelated uncommitted `local_operator/session/remote.py` change remains preserved and is not part of this PR.

Ready for round 3 against `436897a938da193407d708db4698ff40fb7d16ae`.

damianvtran:
### Agent review — round 3
Reviewer: reviewer subagent (openai/gpt-5.6-sol)
Scope: ba414381c4dfeafcea493fbeaac6205a8d312a14..436897a938da193407d708db4698ff40fb7d16ae

No findings.

All prior blocker/major findings are resolved. In particular, the round-1 browser data-loss defect remains durably excluded by the explicit opt-in contract, and round-2 F2 is fixed on both real `run_fetch` emission paths: a fresh result and a true cache hit emit the identical variant-scoped `details['supersede_key']`. Browser actions emit no declared key; raw/rendered and `max_bytes` variants remain distinct; path and declared namespaces cannot collide. The model-visible generic notice and module/tool documentation accurately describe the behavior.

Verification against the exact committed head in a clean detached worktree:

- Real fetch/cache/prune path: cache miss followed by cache hit, two transport calls total (SSRF validation plus one fetch), identical key `raw=0|max_bytes=default`, older result pruned, newest result intact.
- Focused compaction + web-fetch tests: **38 passed**.
- Mutation checks: the benefit test rejects removal of declared-key support; the safety test rejects browser `surface_id` inference and collapsed raw/rendered declarations; call-site tests reject removal of variant threading from either fresh or cache-hit shaping.
- `flake8 .`: clean.
- `black==26.1.0 --check .`: 633 files unchanged.
- `isort==5.13.2 --check .`: clean (1 configured skip).
- `pyright --pythonpath .venv/bin/python .`: 0 errors, 0 warnings, 0 informations.
- `pytest tests/unit`: **9013 passed, 15 skipped, 1 failed**. The failure was `tests/unit/tools/test_builtin_tools.py::test_cancel_backgrounded_bash_kills_process_group`; it passed immediately in isolation (**1 passed**) and is unrelated to the five-file compaction/web-fetch scope, so this is a reproducible load flake rather than a PR defect.

No blocker or major findings remain.

**Verdict: clean. This round is TERMINAL; any future minor or nit observations are follow-ups, not a reason for another review round.**

damianvtran:
### Agent review — round 4
Reviewer: reviewer subagent (openai/gpt-5.6-sol)
Scope: ba414381c4dfeafcea493fbeaac6205a8d312a14..12694a415ed848325d9ce37fe31b9d4c47c00fcc

No findings.

This review is fresh against the exact current PR head `12694a415ed848325d9ce37fe31b9d4c47c00fcc` in a clean independent detached worktree. The only change since the terminal round-3 code head `436897a938da193407d708db4698ff40fb7d16ae` is commit `12694a41`, changing only `pyproject.toml` from `0.43.14` to `0.43.15`.

All prior blocker/major findings remain resolved. The full current diff still uses the safe explicit opt-in `details['supersede_key']` contract rather than inferring identity from browser `surface_id`; browser actions without a declaration remain exempt. The current tests still cover the actual fresh-fetch and cache-hit call sites, assert identical keys for the same rendition, distinguish raw/rendered and `max_bytes` variants, and guard against browser-action data loss and path/declared namespace collisions. No blocker or major findings remain.

The version bump is consistent: `pyproject.toml` is the sole package-version source, no project lockfile requires synchronization, and a clean `uv build --wheel` produced `local_operator-0.43.15-py3-none-any.whl` whose METADATA reports `Name: local-operator` and `Version: 0.43.15`.

Verification against the exact head:

- `flake8 .`: clean.
- `black==26.1.0 --check .`: 633 files unchanged.
- `isort==5.13.2 --check .`: clean (1 configured skip).
- `pyright --pythonpath .venv/bin/python .`: 0 errors, 0 warnings, 0 informations.
- `pytest tests/unit`: **9013 passed, 15 skipped, 1 failed**. The sole failure was `tests/unit/tui/test_subagent_view.py::test_history_unavailable_and_error_retry_keep_trajectory_fallback`; it passed immediately in isolation (**1 passed**) and is unrelated to the six-file compaction/web-fetch/version scope, so this is a load/order timing flake rather than a PR defect.
- Focused compaction + web-fetch tests: **38 passed**.

No blocker or major findings remain.

**Verdict: clean. This round is TERMINAL; any future minor or nit observations are follow-ups, not a reason for another review round.**

damianvtran:
### Agent review — round 5
Reviewer: reviewer subagent (openai/gpt-5.6-sol)
Scope: f735dcfdce881fbbf19d6e15d4bed1fa6c574ec7..5fec0395e94e834a81dbc2dcc3c7a801eae820e0

Reviewed the current committed PR head only in a clean detached worktree at exact head `5fec0395e94e834a81dbc2dcc3c7a801eae820e0`.

No findings.

The explicit opt-in contract remains safe after the rebase: pruning consumes only `details["path"]` or a tool-declared `details["supersede_key"]`; it does not infer from `surface_id`, `url`, or other ambient fields. `web_fetch` emits the key on both fresh and cache-hit paths and scopes it with the existing `cache_variant(raw, max_bytes)`. Tool-name plus `path:`/`declared:` namespaces keep path and declaration groups separate.

Verification:
- `.venv/bin/python -m flake8 .` — passed
- `uvx --from black==26.1.0 black --check .` — passed (670 files unchanged)
- `uvx isort==5.13.2 --check .` — passed
- `.venv/bin/python -m pyright --pythonpath .venv/bin/python .` — passed (0 errors)
- focused tests: `tests/unit/compaction/test_pruning.py tests/unit/web_fetch/test_tool.py` — 38 passed
- contract probes passed: repeated same URL+variant superseded older results while preserving the newest; browser snapshot/console/scroll/click results with only a shared surface stayed intact; raw/rendered and `max_bytes` variants stayed distinct; path and declared namespaces did not collide
- mutation probe confirmed reintroducing unsafe `surface_id` inference would group all four distinct browser actions, so the safety test exercises the original blocker axis

All prior blocker/major findings remain resolved on the current rebased head.

**Verdict: clean — TERMINAL.** No blocker or major findings remain; no further agent-review round is required unless the head changes.

damianvtran:
### Agent review — round 6
Scope: 85b0737a1fa33d51eea65826c4fe5bc97de1c5c6..a7351517dcb252cf19546e36fc246d186949a405

Reviewed the exact committed PR head in a fresh detached worktree. The rebased patch is byte-for-byte equivalent by stable patch ID to the round 5 patch, and `git range-diff` reports every commit unchanged apart from rebased commit identities.

No findings.

The feature contract remains correct on the final rebased head: pruning derives identity only from a path or explicit tool-declared `details["supersede_key"]`; no ambient browser `surface_id`, URL, selector, or action inference was introduced. Only `web_fetch` opts in, on both fresh and cache-hit paths, and its key includes the existing `cache_variant(raw, max_bytes)`. Browser snapshot/logs/scroll/click results remain exempt; raw/rendered and differing `max_bytes` results remain distinct; and `path:`/`declared:` namespaces prevent collisions. The version is correctly bumped from main's 0.44.6 to patch version 0.44.7 for this self-contained reliability/performance feature.

Verification:
- `.venv/bin/python -m flake8 .` — passed
- `uvx --from black==26.1.0 black --check .` — passed (670 files unchanged)
- `uvx isort==5.13.2 --check .` — passed
- `.venv/bin/python -m pyright --pythonpath .venv/bin/python .` — passed (0 errors, 0 warnings)
- `env -u NO_COLOR TERM=xterm-256color .venv/bin/python -m pytest tests/unit/compaction/test_pruning.py tests/unit/web_fetch/test_tool.py -q` — 38 passed
- `git diff --check` — passed

There are no blocker or major findings.

**Verdict: clean — TERMINAL.** No further agent-review round is required unless the head changes.

damianvtran:

Round 6 reported no findings, therefore no remediation code changes were required.

The reviewed and current head SHA is `a7351517dcb252cf19546e36fc246d186949a405`. All prior findings remain fixed in the current rebased commits, and no blocker or major findings remain.

damianvtran:
### Agent review — round 7
Scope: c4d410e6b066937c90ca6980dd7b5a2842c3dcea..b9333095957aa179bf66d8a2c193655dd3299416

Independent re-review of the exact final head after the rebase over PR #480. Fresh detached worktree at `b9333095` (`git rev-parse HEAD` confirmed; PR headRefOid matches). The rebase dropped the stale 0.44.7 bump and re-added `chore(release): bump version to 0.44.8` on top — no other semantic drift introduced by the rebase itself.

**Diff shape**: 6 files, +201/−12. Production: `compaction/pruning.py` (declared-key opt-in + namespace markers), `web_fetch/tool.py` (emit `supersede_key` from both result-shaping sites), `tools/builtin.py` (comment only — the `read` resolver path explicitly does NOT declare), `pyproject.toml` (0.44.7 → 0.44.8).

**PR #480 interaction** — verified coherent, no interference:
- Textual overlap with PR 459's files is nil (#480 touched `tui/app.py`, `widgets/transcript.py`, `widgets/status_line.py`; this PR touches none of those), and the pruning code this PR modifies is not on the resume-render path — resume renders from the transcript, pruning runs on `context.messages` in the loop.
- The merged result keeps both mechanisms distinct by construction: the path/range-coverage supersede from main and the declared opt-in live in different namespaces of the same key space — `"{tool}:path:{path}:{range}"` vs `"{tool}:declared:{key}"` (`pruning.py:175,181`). `_supersede_key` prefers `path` and returns before the declared branch, so one message can never carry both; `_supersede_path_range` and the coverage walk (`seen_full_paths`/`seen_covered_spans`) are gated on `path_range is not None`, which is only true when a `path` detail exists — a declared-key message has `path_range=None` and can never enter the range-coverage sets. No double-blanking: each victim is blanked once (appended to `superseded` at most once per index; pass (b) excludes `superseded_set`).

**No ambient inference** — confirmed. `_supersede_key` reads exactly two fields: `details['path']` and `details['supersede_key']`. No `url`, `surface_id`, or `selector` read anywhere in the supersede path. The only emitter in the tree is `web_fetch/tool.py:220`; `read`'s resolver path documents why it deliberately does not (`builtin.py:2303`, internal URLs, skill reads already exempt).

**Only web_fetch opts in, fresh + cache-hit, keyed by cache_variant** — confirmed. `_shape_from_content` is the single choke point and has exactly two callers: `_cache_hit_result` (line 285) and the fresh path in `run_fetch` (line 402); both now pass `variant=variant`. The key is `f"{final_url or url}\x00{variant}"` where `variant = cache_variant(raw=raw, max_bytes=max_bytes)` = `raw={0,1}|max_bytes={n|default}` — the same discriminator the cache read/write agree on, so a raw and a rendered fetch, or a capped and a full fetch, of one URL are different groups and cannot blank each other. Errors return early with `{"url", "cache"}` only — no key on error results, correct since a failed fetch is not a re-read.

**Version**: 0.44.7 → 0.44.8 is a correct PATCH bump — a context-efficiency improvement with no new user-facing capability, exactly the case AGENTS.md "Versioning" puts in the patch bucket.

**Gates** (fresh venv in the detached worktree, whole tree as CI runs):
- `flake8 .` — clean (0)
- `black==26.1.0 --check .` — clean (671 files unchanged)
- `pyright --pythonpath .venv/bin/python .` — 0 errors, 0 warnings, 0 informations
- Focused: `tests/unit/compaction/test_pruning.py tests/unit/web_fetch/test_tool.py` — **38 passed**
- PR-480 regression: `tests/unit/tui/test_resume_render.py tests/unit/test_fork.py` — **81 passed** (note: `test_fork.py` lives at `tests/unit/`, not `tests/unit/tui/`)

**Findings**

- **F1 — NIT** (`tests/unit/session/test_transcript_footprint.py:154,160`): pins the old notice literal `"[Superseded by a newer read of this file]"` as a bare string, out of scope of this PR's rename to `SUPERSEDED_NOTICE = "[Superseded by a newer result for the same resource]"`. The test still passes because it round-trips its own literal through `append_prune`, so it no longer pins what the pruning pass actually writes — a future divergence between the constant and this literal would stay green. Fix is one line (import and use `SUPERSEDED_NOTICE`); left alone it's harmless.

No BLOCKER, no MAJOR, no MINOR. The namespace markers, the single-choke-point key emission, the variant-scoped grouping and the deliberate non-opt-in on the `read` resolver path are all correct as written, and every claim in the PR description that I could execute held up under execution.

**Verdict: clean — TERMINAL.** No blocker or major finding remains; F1 is a nit recorded here as the follow-up record, not grounds for another round. Round 7 is the terminal round on head `b9333095`.

damianvtran:

Reviewed head: `b9333095957aa179bf66d8a2c193655dd3299416` (matches the PR headRefOid at time of review).

Round 7 found **no BLOCKER, no MAJOR, and no MINOR findings** on head `b9333095957aa179bf66d8a2c193655dd3299416`, so no remediation code changes were required for this round. Nothing was pushed, rebase-included, or otherwise changed after the reviewed head.

Findings and dispositions, exactly as recorded in the round 7 comment:

- **F1 — NIT** (`tests/unit/session/test_transcript_footprint.py:154,160`): pins the old notice literal `"[Superseded by a newer read of this file]"` as a bare string instead of the renamed `SUPERSEDED_NOTICE` constant.
  `deferred — per the round 7 review: the test still passes because it round-trips its own literal through `append_prune`, so "left alone it's harmless"; the reviewer recorded F1 in the round 7 comment itself as the follow-up record, "not grounds for another round". Not fixed in this round per the terminal-round instruction.

No remediation commits exist for this round, so the reviewed head remains the current PR head: `b9333095957aa179bf66d8a2c193655dd3299416`. Round 7 stands as the terminal round.

PR #453, opened 2026-08-30
feat(tui): expanded todo footer says how much is left

damianvtran:
# PR Description

Closes #264
Closes #265

## Summary of Changes

#264 is U4 and #265 is U5. Both were deferred from PR #260's round-1 UX review,
and #265 says it is "best done alongside U4", so this designs them as ONE
expanded-state footer rather than bolting two independent strings together.

### The defect

When an expanded todo list overflows its height budget, the only signal that
content continues is the scrollbar thumb. That is easy to miss on a short window
and it does not exist at all for a keyboard-only reader. Once you scroll, nothing
tells you how much is left.

Captured on `origin/main` at 100x30 with a 17-todo plan, expanded and scrolled to
three positions:

| scroll position | footer on `main` |
|---|---|
| top (6 todos below the fold) | `ctrl+t to collapse` |
| middle (3 below) | `ctrl+t to collapse` |
| bottom (0 below) | `ctrl+t to collapse` |

The same row in all three states, silent about all of it.

### The fix

```
↓ 6 more · ctrl+t to collapse       at the top
↓ 3 more · ctrl+t to collapse       mid-scroll
ctrl+t to collapse                  at the end, prefix gone entirely
```

One number answers both findings. The arrow is U4's "content continues" cue; the
remainder is U5's orientation. Painting both suggestions literally
(`12 of 18 · ↓ 6 below`) would state one fact twice, so only one number is carried.

**Why a remainder and not `N of M`.** The first revision of this PR carried
`↓ 11 of 17`. Review measured that unverifiable: every convention a terminal
reader brings to `N of M` — pagers, `less`, search results — makes `N` the
*current* item, but the only `N` this panel can compute is the last todo the
viewport reaches. Items carry no ordinals, so no visible row can confirm the
number, and it invites the other reading ("N are showing"), which is also false.
A remainder is checkable by scrolling, and it mirrors the collapsed `+11 more`
the same reader saw one keypress earlier.

### Constraints the issues and reviews called out, and how each is met

**The cue never vanishes while content continues.** The prefix is width-tested
against its *widest* rendering for the list, not the string being painted. Sizing
on the current text made the cue's survival depend on the number's digit count,
so at 34 columns it lived at `↓ 9 of 17` and died at `↓ 10 of 17` — showing the
exact pre-change "this is everything" row with 8 todos still hidden. That is
worse than never showing a cue, because absence had been taught to mean something.

**The arrow is derived from the viewport, not the todo count.** The counts are
[...]

PR #136, opened 2026-08-17
fix(tui,harness): name sessions from their opener, and generate the title during the turn

damianvtran:
# PR Description

## Change classification

**C2 — standard / pre-authorized.** Four bounded fixes to conversation naming and one
transport-isolation flag they depend on, with a clean rollback (`git revert` / reinstall
0.17.4). No foundational auth, data-isolation, infrastructure, or irreversible data
change. No RFC or tracker requested.

## Summary of Changes

This releases **local-operator 0.17.5** and closes four naming defects observed live in
one `lop` session, plus the transport work the third one required.

### 1. `/resume` forgot every session whose opener had a screenshot

A conversation is labelled in the picker by its opening user message, read back from the
transcript — titles live in memory and were never journalled. `session_name` read a
64 KB window and dropped a final line that had no newline, which is correct for a
half-read line but fatal when the whole opener is one line longer than the window: a
message with a pasted image is 115 KB, so the window held **no complete line** and the
row fell back to `(unnamed session)`.

The opening text sits at offset 135 — before the base64 `data` at 443 — so it was always
recoverable. A fragment scan now recovers the opener from a half-read first line, guarded
so a session can never be named after base64: the fragment must identify itself as a user
message, and the first complete `"text"` value must appear *before* any `data` key.
Whitespace-tolerant, because a transcript written by anything other than the session
writer is still the same document.

Measured on the real session directory: **8/8 recent sessions now named, 0 unnamed**,
including the 115 KB and 733 KB openers.

### 2. The status band showed an indicator nobody needed

The trailing segment was `! auto-approve always`. It now carries the **session name**,
which is the more valuable thing to have on screen; the approval mode collapses to a bare
`!` alarm, and `/approvals` remains the place that answers "until when".

### 3. The title arrived after the work it was naming

Auto-naming waited on `_turn_settled` — unbounded by design, so decoration could never
make a user request worse. But a first turn on this product runs for minutes, so the title
landed after the turn. **Measured before: a 29.7 s turn with the title stored 1.8 s after
it** (106% of the turn).

The naming call is now dispatched at turn launch and runs concurrently. That is only safe
because it is **isolated**, and that is the bulk of this diff: a second in-flight request
shared five things with the user's turn, each a live route by which a *decorative* failure
could have degraded it.

| Shared state | What a title failure would have done to the turn |
|---|---|
| `FailoverRouteState` (session-sticky) | `activate` a fallback with a 60 s cooldown — moving the **turn** onto the fallback model |
[...]

PR #461, opened 2026-08-30
test(harness,tui,providers): wait on conditions instead of the clock

### LINKED ISSUE #122: test(harness): replace test_comms wait_for's wall-clock deadline with a deterministic sync point
Found by an agent review round on #120 (round 34), during the whole-branch merge-gate sweep. Not a defect in that PR's diff; deferred out of it deliberately so the fix gets its own review rather than a last-gate patch.

## What

`tests/unit/harness/test_comms.py:52-57` introduces a `wait_for` helper with a **10-second wall-clock deadline**. Two tests depend on it:

- `test_a_resumed_child_replays_the_stopped_ones_transcript`
- `test_a_cancelled_child_keeps_the_work_it_had_already_done`

Under machine contention both time out at `test_comms.py:57` with `AssertionError: timed out waiting for condition`. Observed during a full-suite run made while six concurrent pytest sessions and a second agent's suites were running.

## Why it is not a correctness bug

- Both pass in isolation, three runs over.
- The whole module passes (43).
- The whole harness directory passes (174 in 3.8s).
- Full suite on a quiet machine: 3825 passed, 0 failed.
- CI green on both 3.12 and 3.13.

## Why it is still worth fixing

A wall-clock bound tuned on an unloaded machine will intermittently fail CI on a busy runner, and that cost lands on every future MR rather than on the one that introduced it. Raising the bound moves the threshold without removing the class.

The fix is to replace the polling wait with a deterministic synchronisation point, so the test blocks on the event it actually cares about instead of on elapsed time.

### LINKED ISSUE #142: test: four load-sensitive tests fail under concurrency on main

### LINKED ISSUE #224: test(providers): close http client before assertions in mid-stream failover test
Follow-up from PR #223 agent review round 2 (R2-1, nit): in `test_an_openrouter_error_chunk_after_partial_output_arrives_named` (tests/unit/providers/test_failover.py), `http.aclose()` sits after the assertions, so a failing assertion skips the close. Harmless with `httpx.MockTransport` (no real sockets), but moving the close into a `finally` (or using the client as a context manager) keeps the test tidy. Deferred from the PR to keep the reviewed head stable.

damianvtran:
# PR Description

## Change classification

**C2 — standard / pre-authorized.** Test-only. No file under `local_operator/`
is touched, so no runtime path changes and the rollback is `git revert`.

## Summary of Changes

Three issues, one defect shape: **a test that waits for asynchronous work by
spending a fixed budget, then asserts as though the budget were a guarantee.**
On an idle machine the budget is 20-50x more than the work needs, so the tests
look solid; under contention the work stretches, the budget does not, and the
suite goes red for reasons that have nothing to do with the code under test.

The fix in every case is to wait on the *event the test actually cares about*,
published by the code under test itself.

### 1. `#122` — `test_comms.wait_for` had a 10 s wall-clock deadline

`tests/unit/harness/test_comms.py` polled a predicate and raised
`AssertionError("timed out waiting for condition")` once 10 s of wall clock had
passed. Two integration tests
(`test_a_resumed_child_replays_the_stopped_ones_transcript`,
`test_a_cancelled_child_keeps_the_work_it_had_already_done`) wait through a real
child session doing real tool rounds, so they are the ones that lose that race.

Replaced with `ChangeSignal`, an `asyncio.Event` fed by the two publications the
production code already makes:

| source | fires when |
| --- | --- |
| `SubagentComms.subscribe_detail_changes` | after each durable transcript append (`notify_detail_persisted`) |
| `Session.subscribe` | every parent-stream event, which carries attach (`SubagentStartEvent`) and settle |

Both are watched because neither alone covers every predicate used here: the
detail registry sees nested children the parent stream never routes, and the
parent stream carries the attach moment that makes `session_dir_of` non-`None`
before any append has happened. The signal is level-triggered — `arm()` clears
the flag *before* the predicate is tested — so a publication landing between the
test and the wait is kept rather than lost.

The two tests now have **no elapsed-time comparison anywhere on their success
path**. A `DEADLOCK_GUARD_S = 120.0` remains, but it can never be the
discriminator: the success path blocks on an event, so the only run that reaches
the guard is one where the signal is never coming (this suite has no
`pytest-timeout`, so a wedged test would otherwise hang CI forever).

Call sites that drive `FakeChild` in-process keep a pump, but bounded by **loop
turns** (`MAX_PUMP_TURNS`) rather than seconds. That unit is the point:
contention stretches how long a turn takes, not how many turns the awaited work
needs.

### 2. `#224` — a client closed after its assertions

`test_an_openrouter_error_chunk_after_partial_output_arrives_named` called
`await http.aclose()` *after* its assertions, so a failing assertion skipped the
close. Moved into `finally`, which is the shape the other 22 client-owning tests
in that file already use. Audited the file: it was the only `aclose()` in
`test_failover.py` outside a `finally`.

### 3. `#142` — frame budgets standing in for worker completion

Two cases in `tests/unit/tui/test_subagent_stats.py` waited on
`for _ in range(50): await pilot.pause()` for work owned by **Textual workers**:

- `OperatorApp.on_mount` hands `_boot_session` to `run_worker(group="session")`
  precisely so first paint does not wait on the factory. Every band assertion
  reads the ledger *through* that session, so an unbooted app reports no jobs at
  all — the observed failure is `assert app._session is not None`.
- `SubagentPanel._read_stats` prices each child on a thread worker
  (`subagent-stats-<id>`) because resolving an unlisted model is a 10 s provider
  listing plus a 3 s catalogue fetch.

Both now wait on the worker itself via `WorkerManager.wait_for_complete`, plus
one `pause()` after the thread worker so its `call_from_thread` callback lands.
`run_test` cancels outstanding workers on exit, so this cannot hang past the run.

The `_priced` helper also asserts the reading is actually populated, so a worker
that finishes without pricing fails loudly instead of letting the band
assertions pass against an empty reading.

### 4. The mobile ask helper — same class, added during review

`tests/unit/mobile/test_tui_ask.py::_control_for` is the shared helper for that
module. It polled until `app._mobile_registrant is not None`, then *immediately*
called `registry.scan()` and asserted the result non-empty.

Those are not the same event. `Registrant.start` spawns a daemon thread and
returns; the record is written by `_serve` **on that thread**. So the registrant
existing means registration was *requested*, not that the record is on disk —
wait-on-a-proxy-then-assert-on-the-real-thing, the same shape as the rest of
this diff. It now polls `registry.scan()` itself. The registrant assertion is
kept, moved after the wait, so "nothing ever started" still reports as itself
rather than as a missing record.

This was blocking two unrelated PRs (#450 on 3.12, #453 on 3.13) with
`no discovery record published` — different tests, same helper, and neither
diff touches mobile code. The clearest evidence it is environmental: on #450 it
failed on run `33326634006` and passed on run `33326636132` against an
**identical head**.

**The two `test_loop_liveness.py` cases in #142 need no change, and this is
evidence rather than an opinion.** #142 was filed against `42e5268`, where the
probe compared *wall-clock gaps between heartbeats* (`MAX_GAP_S = 0.12`) — a
statistic that widens whenever the loop thread is not *scheduled*, which is
exactly what load does. #136 replaced it with `time.thread_time` on the loop
thread, which only grows when the loop thread genuinely runs without yielding.
Measured below: they now pass under the same load that used to break them.

## Related Issue(s)

Closes #122
Closes #224
Closes #142

## Testing Evidence

Host: 14 cores. This box runs a dozen agent sessions, so load moves minute to
minute; every comparison below is therefore either **paired** (both arms
back-to-back under one load) or reports the load it ran at.

### The failure, reproduced on base

`test_opening_and_leaving_a_page_repoints_and_restores_the_live_band`, 10 pytest
processes racing each other, on a clean `origin/main` worktree:

```
== lop-td-base tests/unit/tui/test_subagent_stats.py
== 5 passed / 5 failed of 10 (spinners=24, load=258.33 228.89 184.52)

E  AssertionError: assert None is not None
E   +  where None = OperatorApp(...)._session
FAILED tests/unit/tui/test_subagent_stats.py::test_opening_and_leaving_a_page_repoints_and_restores_the_live_band
```

Every failure is the same assertion: the app had not adopted a session inside 50
frames.

### The head, same procedure, 30 runs

```
== lop-test-determinism tests/unit/tui/test_subagent_stats.py
after round 1: 10 passed / 0 failed
after round 2: 20 passed / 0 failed
after round 3: 30 passed / 0 failed
== 30 passed / 0 failed of 30 (procs=10 rounds=3 spinners=16, load=212.57 250.51 267.04)
```

Whole module, 30 concurrent runs at load ~250, zero failures.

### Why a pass/fail sample understates it: the margin, measured

A sample only says whether the budget was crossed on the runs observed. The
quantity that decides it is **how many frames boot actually costs against the 50
the test allowed**. Measured with an uncapped probe (`.prwork` scratch, 8
concurrent processes, 4 apps each):

| condition | frames to session adoption |
| --- | --- |
| idle, any app | `[1, 1, 1, 1, 1]` |
| loaded, **first** app in each process | `15, 17, 18, 22, 24, 24, 24, 27, 27, 29, 30, 32, 34, 34` |
| loaded, subsequent apps in same process | `1` |

The first `run_test` in a worker pays the session-graph import; that is the run
the 50-frame budget had to cover. Idle it costs 1 frame — a 50x margin, which is
why this reads as solid on a quiet machine. Under load it costs **15-34 frames,
a margin of 1.5-3.3x**, and the observed failures are the tail of that
distribution crossing 50. `wait_for_complete` removes the budget entirely, so
the distribution no longer has a failing tail to reach.

Child pricing measured `0-1` frames against its 60-frame budget in every
condition, so that one was not close to failing — it is converted for the same
reason regardless: it is the same bet, and it costs nothing to stop making it.

### `#122`, before and after

The comms pair is slow (a real child, real tool rounds), so the paired A/B is at
lower width. Both arms back-to-back under one load:

```
== targets: test_a_resumed_child_replays_the_stopped_ones_transcript
             test_a_cancelled_child_keeps_the_work_it_had_already_done
== BASE(origin/main)          : 24 passed /  0 failed of 24
== HEAD(dev-test-determinism) : 24 passed /  0 failed of 24
```

Honest reading: **I did not reproduce a #122 failure on this box.** The issue
records it from a run with six concurrent pytest sessions plus a second agent's
suites; my harness reached load ~250 without crossing its 10 s deadline. The
change still stands on the mechanism rather than on a reproduction — the old
code compares `loop.time()` against a deadline and the new code does not compare
time at all — and the module is green: `106 passed`.

### The whole module set, after

```
$ .venv/bin/python -m pytest -n4 -q \
    tests/unit/harness/test_comms.py tests/unit/tui/test_subagent_stats.py \
    tests/unit/providers/test_failover.py tests/unit/tools/test_loop_liveness.py \
    tests/unit/tui/test_app_pilot.py
585 passed, 87 warnings in 65.36s
```

### Full unit suite

```
$ .venv/bin/python -m pytest tests/unit -n4 -q; echo "pytest rc=$?"
9143 passed, 15 skipped, 403 warnings in 565.55s (0:09:25)
pytest rc=0
```

Clean at the current head. The run before the round-1 fixes had one failure,
`test_subagent_view.py::test_history_unavailable_and_error_retry_keep_trajectory_fallback`
— which turned out to be a real defect in that file rather than ambient noise,
and is now fixed here (see "A third mechanism" below).

An earlier revision of this body also recorded
`test_app_pilot.py::test_esc_aborts_a_running_shell_command` failing once; it
passes 3/3 on clean `origin/main` and in every run since, and this diff does not
touch that file.

### `test_loop_liveness.py` is already fixed, measured

The two cases #142 names, on **clean `origin/main`**, under the load that the
issue reports breaking them:

```
== 10 passed / 0 failed of 10 (spinners=24, load=170.45 153.28 140.50)
== 10 passed / 0 failed of 10 (spinners=48, load=262.30 216.77 172.77)   [whole file]
```

At `42e5268` (the SHA #142 cites) the probe was `MAX_GAP_S = 0.12` on wall-clock
gaps; #136 replaced it with loop-thread CPU. Touching these now would be
changing a guard that is already load-immune.

### The mobile helper

```
$ .venv/bin/python -m pytest -n4 -q tests/unit/mobile
206 passed, 2 warnings in 6.63s
```

Honest reading, same as #122: **I did not reproduce this one locally either.**
It needs CI's 14-worker profile; four runs of `tests/unit/mobile tests/unit/tui`
at `-n auto` came back clean on **both** arms:

```
BASE  run1 rc=0   run2 rc=0   run3 rc=1 (test_subagent_view)   run4 rc=0
HEAD  run1 rc=0   run2 rc=0   run3 rc=0   run4 rc=1 (test_subagent_view)
```

The change rests on the mechanism, which is not in doubt: the old code waits for
a thread to be *spawned* and asserts the thread's *output* is already written.
CI is the measurement that matters here and it runs on this PR.

(Those `test_subagent_view` failures on both arms were the shared-fixture leak,
diagnosed and fixed in this PR — see "A third mechanism". At the time they were
recorded as ambient noise, which was wrong.)

### Gates

Verified by exit code, whole tree, CI's exact commands:

```
.venv/bin/python -m flake8 .                                flake8 rc=0
uvx --from black==26.1.0 black --check .                    black rc=0
uvx isort==5.13.2 --check .                                 isort rc=0
.venv/bin/python -m pyright --pythonpath .venv/bin/python . pyright rc=0
.venv/bin/python -m pytest tests/unit -n4 -q                pytest rc=0
```

Exit codes rather than stdout, deliberately: while adding the round-1 fixes an
import landed in the wrong order and `isort --check .` printed its usual
`Skipped 1 files` to stdout while returning **rc=1**, with the real error on
stderr. A tail of stdout would have read as green. Black's success banner
likewise goes to stderr.

Rebased onto `7916e5ae` (current `origin/main`), cleanly and with no conflicts,
so the round is fresh against what ships.

## Disposition: the wider flake population

#142's four are part of a larger set, and the manager supplied CI data showing 3
of the last 8 `main` runs red. Recording what I found and deliberately left, so
the next reader does not rediscover it. **No new issues were filed for these.**

**Fixed here:** `#122`'s two comms tests, `#224`'s client close, `#142`'s two
`test_subagent_stats.py` cases, `test_tui_ask.py::_control_for`, and both
`test_subagent_view.py` defects (the last three added during review because
they were actively blocking other PRs' CI).

### CORRECTION — the `test_a_live_selection_never_swallows_the_interrupt` claim

**An earlier revision of this body claimed that test fails 8/8 unloaded and is a
deterministic binding defect. That claim was wrong and is withdrawn.** Round 1
could not reproduce it in 13 runs, so I re-measured across four trees, by exit
code:

```
current head (rebased):        6 runs, 0 failures
65ca130a (my later base):      3 runs, 0 failures
75f867bf (the tree I
  originally measured):        4 runs, 0 failures
cc8319d2^ (before #446):       3 runs, 0 failures
```

16 unloaded runs, zero failures, **including on the exact tree my original
measurement ran against**. So this was not a stale measurement corrected by
cc8319d2 — cc8319d2 is not the fix, and the test body is byte-identical before
and after it. My original run was simply wrong, and I have not been able to
reconstruct the mistake.

Under CPU load (24 spinners, load average 24 to 77) it does fail:

```
6 runs at -n0 under load: 3 failed, 3 passed
all 3 failures: interrupt[False], assert [] == ['interrupted']
```

**So it is a load flake after all — the original guidance given to this board
was correct, and my finding briefly and wrongly reversed it.** It belongs to
#142's class, not to a separate binding-defect class. Not fixed here (it is in
a file this PR does not otherwise touch, and the fix is not the one-line kind),
but nobody should chase a deterministic defect in the binding chain, because
there is not one.

**Left untouched, same class but out of scope:**
`test_a_multi_click_on_a_blank_line_leaves_a_live_range`,
`test_the_inset_is_never_what_tips_a_long_subagent_list_over` (`NoMatches: No
nodes match '#band'`), `test_redundant_roster_events_skip_the_sidecar_write`,
and `test_app_pilot.py::test_esc_aborts_a_running_shell_command`.

### A third mechanism: shared mutable fixture state

Round 1 was told `test_subagent_view.py` fails deterministically at `-n4` and
passes at `-n0`, and asked whether that was a third mechanism. It is, and it is
two distinct defects in one file. Neither is a clock budget and neither is a
wait on a proxy:

**1. A module-level list handed out by reference.** `_job_with(TRAJECTORY)`
assigned the shared list straight onto the job. Three tests mutate what they
get back to drive a live update (`job.trajectory.append(...)` twice,
`del job.trajectory[:3]` once), so those edits landed on the module-level list
and every later test folded a corrupted trajectory. That is the
`assert [...'ToolCard'] == [...'AssistantBlock']` failure: an extra tool call
left behind by an earlier test. Fixed by copying the list in `_job_with`, so a
new test cannot reintroduce it by writing the obvious thing.

This is why the symptom moved around and why xdist changed the count without
changing the cause: which test sees the corruption depends on execution order.

**2. An assertion one frame after opening the page.**
`test_history_unavailable_and_error_retry_keep_trajectory_fallback` asserted
`HISTORY_UNAVAILABLE_NOTE` immediately after `_open`, but the missing directory
is discovered inside a `subagent-history` worker (`asyncio.to_thread`), and
`_history_unavailable` is written in its completion path. This one IS this PR's
main class. `_wait_history` was itself a 100-cycle budget poll; it now waits on
the worker, and the test waits before asserting.

Measured by exit code, `tests/unit/tui/test_subagent_view.py` at `-n auto`
(14 workers, CI's profile):

```
BASE (origin/main b56cf73f):  8 passed, 2 failed  of 10 runs
HEAD:                        10 passed, 0 failed  of 10 runs
```

Base also fails at `-n0` (1 of 1) once the ordering is unlucky, so the
"deterministic at -n4, invisible at -n0" framing was a sampling artefact of the
corruption's position rather than a property of xdist.

### The shared mechanism, named

The manager asked whether these want one mechanism rather than seven point
fixes. They do, and it is narrower than "wait on a condition":

> **A Textual test must wait on the app's own completion signal for work the app
> put on a worker — never on a count of `pilot.pause()` calls.**

`pilot.pause()` advances one message-pump cycle. It does not advance a
`run_worker` task, a thread worker, or a layout pass; it only yields long enough
that *something else may* progress. So `for _ in range(N): await pilot.pause()`
is not a wait — it is a bet that N pump cycles is longer than an unrelated piece
of asynchronous work, priced on an idle machine.

Both visible patterns are that one bet:
- *asserting on elapsed time* — the budget is spent, the work is not done;
- *querying a widget before layout settled* (`NoMatches: '#band'`) — the same
  bet, where the awaited work is compose/layout rather than a worker.

The general remedy is the one applied here: find who owns the work
(`app.workers` for a worker, an `asyncio.Event` the code already publishes for a
session boundary) and block on that. `WorkerManager.wait_for_complete` covers
every `run_worker`/`run_worker(thread=True)` case, which is most of the TUI
suite's exposure, and it is mechanical to apply. I did not sweep the whole suite
for it here because that is a large diff whose risk is unrelated to these three
issues, but the two helpers added in `test_subagent_stats.py` (`_booted`,
`_priced`) are the pattern in a form another file can copy.

Not colliding with #434/#438: both touch only `tests/conftest.py`, which this
diff does not.

## Impact

- **Breaking changes:** none. Test-only.
- **Risk:** the changed waits could in principle mask a real hang. Mitigated by
  keeping a bound on every path (`DEADLOCK_GUARD_S`, `MAX_PUMP_TURNS`,
  `run_test`'s own worker cancellation) and by `_priced` asserting the reading it
  waited for is actually there, so a worker that completes without doing its job
  fails rather than passing quietly.
- **Rollback:** `git revert`.

### DIFF
--- a/tests/unit/harness/test_comms.py
+++ b/tests/unit/harness/test_comms.py
@@ -51,13 +51,119 @@ from local_operator.tools.registry import create_tools
 MODEL = ModelSpec(provider="test", model_id="m", context_window=100_000)
 
 
-async def wait_for(predicate, timeout: float = 10.0) -> None:
-    loop = asyncio.get_running_loop()
-    deadline = loop.time() + timeout
-    while not predicate():
-        if loop.time() > deadline:
-            raise AssertionError("timed out waiting for condition")
+#: Ceiling for the signal-free pump below, counted in LOOP TURNS rather than
+#: in seconds. That unit is the whole point (#122): machine contention
+#: stretches how long a turn takes but not how many turns the awaited work
+#: needs, so a loaded box cannot exhaust this budget the way it exhausted the
+#: 10 s wall-clock deadline this replaced. It is a deadlock guard — reached
+#: only when the awaited thing is never going to happen — not a timing
+#: assumption, and raising it is never the fix for a failure here.
+MAX_PUMP_TURNS = 3000
+
+#: Same role for the signal-driven wait: a wedged test must fail rather than
+#: hang a CI job forever (this suite has no pytest-timeout). Generous because
+#: it can never be the discriminator — the success path blocks on an
+#: ``asyncio.Event`` set by the code under test and never compares elapsed
+#: time — so the only run that reaches it is one where the signal will not
+#: arrive at all.
+DEADLOCK_GUARD_S = 120.0
+
+
+class ChangeSignal:
+    """An edge fed by the code under test's OWN notifications.
+
+    The deterministic synchronisation point #122 asks for. A child's progress
+    is published twice over — ``SubagentComms.notify_detail_persisted`` fires
+    after each durable transcript append, and the parent's event stream
+    carries ``SubagentStartEvent``/``SubagentProgressEvent`` — so a test that
+    needs "the child got somewhere" can block on those publications instead of
+    re-reading the filesystem on a timer and betting the answer arrives before
+    a deadline.
+
+    Both sources are watched because neither alone covers every predicate:
+    the detail registry sees nested children the parent stream never routes,
+    and the parent stream carries the attach moment (``SubagentStartEvent`` is
+    emitted immediately after ``comms.attach``) that makes ``session_dir_of``
+    non-None before any transcript append has happened.
+
+    Level-triggered on purpose: :meth:`settle` clears the flag BEFORE the
+    caller re-tests its predicate, so a notification racing in between the
+    test and the wait is kept rather than lost.
+    """
+
+    def __init__(self) -> None:
+        self._event = asyncio.Event()
+        self._unsubscribes: list[Callable[[], None]] = []
+
+    def _fire(self, *_: Any) -> None:
+        self._event.set()
+
+    def watch_comms(self, comms) -> "ChangeSignal":
+        """Wake on every durable child-transcript append."""
+        self._unsubscribes.append(comms.subscribe_detail_changes(self._fire))
+        return self
+
+    def watch_session(self, session: Session) -> "ChangeSignal":
+        """Wake on every event the parent's stream carries."""
+        self._unsubscribes.append(session.subscribe(self._fire))
+        return self
+
+    def arm(self) -> None:
+        """Drop any pending edge so the next wait blocks on a NEW one."""
+        self._event.clear()
+
+    async def settle(self) -> None:
+        """Block until something publishes."""
+        await self._event.wait()
+        self._event.clear()
+
+    def close(self) -> None:
+        for unsubscribe in self._unsubscribes:
+            unsubscribe()
+        self._unsubscribes.clear()
+
+
+async def _await_signalled(predicate, signal: "ChangeSignal") -> None:
+    """Re-test ``predicate`` once per publication, forever. Bounded by caller."""
+    while True:
+        # Armed BEFORE the test, never after: a publication landing between
+        # the test and the wait is then still on the flag, so the edge that
+        # would satisfy the predicate can never be missed.
+        signal.arm()
+        if predicate():
+            return
+        await signal.settle()
+
+
+async def wait_for(predicate, *, signal: "ChangeSignal | None" = None) -> None:
+    """Block until ``predicate`` holds, on an event rather than on a clock.
+
+    With a ``signal``, each re-test follows a publication from the code under
+    test, so this waits exactly as long as the work takes and no longer: there
+    is no elapsed-time comparison anywhere on the success path. That is the
+    property #122 asks for, and it is why the integration cases below pass one.
+
+    Without one — the in-process cases over ``FakeChild``, where the awaited
+    step is a coroutine this test itself scheduled and there is nothing to
+    subscribe to — it pumps the loop and bounds itself by TURN COUNT. See
+    :data:`MAX_PUMP_TURNS` for why that unit survives contention where seconds
+    did not.
+    """
+    if signal is not None:
+        try:
+            await asyncio.wait_for(_await_signalled(predicate, signal), DEADLOCK_GUARD_S)
+        except asyncio.TimeoutError:
+            raise AssertionError(
+                "the code under test published no change satisfying the condition in "
+                f"{DEADLOCK_GUARD_S:.0f}s — it is wedged, not merely slow"
+            ) from None
+        return
+
+    for _ in range(MAX_PUMP_TURNS):
+        if predicate():
+            return
         await asyncio.sleep(0.01)
+    raise AssertionError(f"condition never held across {MAX_PUMP_TURNS} loop turns")
 
 
 def has_complete_tool_round(session_dir) -> bool:
@@ -1586,39 +1692,50 @@ async def test_a_resumed_child_replays_the_stopped_ones_transcript(tmp_path, mon
     await parent.async_init()
     job_id = parent._launch_subagent(label="docs", prompt="Update the docs.")
     comms = parent.subagent_comms
-    await wait_for(lambda: comms.session_dir_of(job_id) is not None)
-    # Let it finish real work before stopping it, and wait on THAT rather than
-    # on a duration: the child persists each tool round at its loop boundary
-    # (``Session._persist_progress``) roughly 0.29s after launch, so the fixed
-    # 0.3s sleep this replaced was inside one standard deviation of the thing
-    # it was betting against and lost the coin flip about a third of the time.
-    # Resume is only meaningful over a child that got somewhere, so the
-    # condition to wait for is a completed round, with seconds of headroom so
-    # this fails only if the child genuinely never runs a tool.
-    await wait_for(lambda: has_complete_tool_round(comms.session_dir_of(job_id)), timeout=30.0)
-    await comms.cancel(job_id)
-    await wait_for(lambda: status_of(parent, job_id) == "cancelled")
-    await wait_for(lambda: len(Transcript(comms.session_dir_of(job_id)).build_llm_history()) >= 2)
-    before = Transcript(comms.session_dir_of(job_id)).build_llm_history()
-    original = parent.jobs.get(job_id)
-    assert original is not None and original.launch_message_id == before[0].id
+    # Every wait below rides the child's own publications rather than a
+    # deadline (#122): both sources are watched because the parent stream
+    # carries attach and settle while the detail registry carries each durable
+    # transcript append, and this test waits on all three kinds of fact.
+    signal = ChangeSignal().watch_comms(comms).watch_session(parent)
+    try:
+        await wait_for(lambda: comms.session_dir_of(job_id) is not None, signal=signal)
+        # Let it finish real work before stopping it, and wait on THAT rather
+        # than on a duration: the child persists each tool round at its loop
+        # boundary (``Session._persist_progress``) roughly 0.29s after launch,
+        # so the fixed 0.3s sleep this originally replaced was inside one
+        # standard deviation of the thing it was betting against. Resume is
+        # only meaningful over a child that got somewhere, so the condition is
+        # a completed round — and it is now awaited on the append that creates
+        # it, so a loaded machine makes this slower, never red.
+        await wait_for(lambda: has_complete_tool_round(comms.session_dir_of(job_id)), signal=signal)
+        await comms.cancel(job_id)
+        await wait_for(lambda: status_of(parent, job_id) == "cancelled", signal=signal)
+        await wait_for(
+            lambda: len(Transcript(comms.session_dir_of(job_id)).build_llm_history()) >= 2,
+            signal=signal,
+        )
+        before = Transcript(comms.session_dir_of(job_id)).build_llm_history()
+        original = parent.jobs.get(job_id)
+        assert original is not None and original.launch_message_id == before[0].id
 
-    new_id, error = comms.resume(job_id, "You were interrupted. Wrap up.")
+        new_id, error = comms.resume(job_id, "You were interrupted. Wrap up.")
 
-    assert error is None and new_id is not None
-    continuation = parent.jobs.get(new_id)
-    assert continuation is not None and continuation.launch_message_id
-    assert continuation.launch_message_id != original.launch_message_id
-    await wait_for(lambda: status_of(parent, new_id) != "running")
-    assert comms.session_dir_of(new_id) == comms.session_dir_of(job_id)
-    after = Transcript(comms.session_dir_of(new_id)).build_llm_history()
-    assert len(after) > len(before)
-    assert any(message.id == continuation.launch_message_id for message in after)
-    resumed_node = comms.node(new_id)
-    assert resumed_node is not None
-    assert resumed_node.launch_message_id == continuation.launch_message_id
-    assert first_text(after) == first_text(before) == "Update the docs."
-    assert status_of(parent, new_id) == "completed"
+        assert error is None and new_id is not None
+        continuation = parent.jobs.get(new_id)
+        assert continuation is not None and continuation.launch_message_id
+        assert continuation.launch_message_id != original.launch_message_id
+        await wait_for(lambda: status_of(parent, new_id) != "running", signal=signal)
+        assert comms.session_dir_of(new_id) == comms.session_dir_of(job_id)
+        after = Transcript(comms.session_dir_of(new_id)).build_llm_history()
+        assert len(after) > len(before)
+        assert any(message.id == continuation.launch_message_id for message in after)
+        resumed_node = comms.node(new_id)
+        assert resumed_node is not None
+        assert resumed_node.launch_message_id == continuation.launch_message_id
+        assert first_text(after) == first_text(before) == "Update the docs."
+        assert status_of(parent, new_id) == "completed"
+    finally:
+        signal.close()
     await parent.dispose()
 
 
@@ -1633,15 +1750,23 @@ async def test_a_cancelled_child_keeps_the_work_it_had_already_done(tmp_path, mo
     await parent.async_init()
     job_id = parent._launch_subagent(label="docs", prompt="Update the docs.")
     comms = parent.subagent_comms
-    await wait_for(lambda: comms.session_dir_of(job_id) is not None)
-    # Wait for the round itself, not for how long a round usually takes. The
-    # assertions below are entirely about work the child had ALREADY completed
-    # before the cancel, so cancelling early does not fail loudly — it silently
-    # asserts over a transcript holding only the launch prompt.
-    await wait_for(lambda: has_complete_tool_round(comms.session_dir_of(job_id)), timeout=30.0)
+    signal = ChangeSignal().watch_comms(comms).watch_session(parent)
+    try:
+        await wait_for(lambda: comms.session_dir_of(job_id) is not None, signal=signal)
+        # Wait for the round itself, not for how long a round usually takes.
+        # The assertions below are entirely about work the child had ALREADY
+        # completed before the cancel, so cancelling early does not fail
+        # loudly — it silently asserts over a transcript holding only the
+        # launch prompt.
+        await wait_for(lambda: has_complete_tool_round(comms.session_dir_of(job_id)), signal=signal)
 
-    await comms.cancel(job_id)
-    await wait_for(lambda: len(Transcript(comms.session_dir_of(job_id)).build_llm_history()) > 1)
+        await comms.cancel(job_id)
+        await wait_for(
+            lambda: len(Transcript(comms.session_dir_of(job_id)).build_llm_history()) > 1,
+            signal=signal,
+        )
+    finally:
+        signal.close()
 
     history = Transcript(comms.session_dir_of(job_id)).build_llm_history()
     assert any(
@@ -2348,7 +2473,7 @@ async def test_a_question_buffered_past_the_grace_is_answered_once_it_attaches()
     # ``wait_for``'s default. (A 20 s budget would spend 10 s in the grace and
     # time the WAIT out, not the ask.)
     asking = asyncio.create_task(comms.ask("job-1", "stuck?", 4_000))
-    await wait_for(lambda: bool(comms._records["job-1"].pending), timeout=5.0)
+    await wait_for(lambda: bool(comms._records["job-1"].pending))
 
     comms.attach("job-1", child, tmp_dir())  # the window finally closes
     await wait_for(lambda: bool(child.asides))
@@ -2388,7 +2513,7 @@ async def test_a_stale_buffered_question_cannot_answer_the_next_asker():
 
     # 4 s budget -> 2 s attach grace, comfortably inside ``wait_for``.
     asking = asyncio.create_task(comms.ask("job-1", "NEW: are you blocked?", 4_000))
-    await wait_for(lambda: bool(comms._records["job-1"].pending), timeout=5.0)
+    await wait_for(lambda: bool(comms._records["job-1"].pending))
 
     comms.attach("job-1", child, tmp_dir())
     await wait_for(lambda: bool(child.asides))
@@ -2425,7 +2550,7 @@ async def test_a_second_question_is_refused_on_the_buffered_path_too():
 
     first = asyncio.create_task(comms.ask("job-1", "Q-A", 4_000))
     # Past the attach grace (half the budget, so 2 s here), where Q-A buffers.
-    await wait_for(lambda: bool(comms._records["job-1"].pending), timeout=5.0)
+    await wait_for(lambda: bool(comms._records["job-1"].pending))
     held = comms._records["job-1"].ask
 
     second = await comms.ask("job-1", "Q-B", 4_000)
--- a/tests/unit/harness/test_comms.py
+++ b/tests/unit/harness/test_comms.py
@@ -61,12 +61,19 @@ MODEL = ModelSpec(provider="test", model_id="m", context_window=100_000)
 MAX_PUMP_TURNS = 3000
 
 #: Same role for the signal-driven wait: a wedged test must fail rather than
-#: hang a CI job forever (this suite has no pytest-timeout). Generous because
-#: it can never be the discriminator — the success path blocks on an
+#: hang a CI job forever (this suite has no pytest-timeout). It can never be
+#: the discriminator for TIMING — the success path blocks on an
 #: ``asyncio.Event`` set by the code under test and never compares elapsed
-#: time — so the only run that reaches it is one where the signal will not
-#: arrive at all.
-DEADLOCK_GUARD_S = 120.0
+#: time — so the only run that reaches it is one where the signal is not
+#: coming at all.
+#:
+#: It is still a COST, though, which is why it is 30 s and not the 120 s this
+#: started at (review round 1, F2): a wedge is paid per affected test, and four
+#: waits here share the bound, so an over-generous guard turns one real
+#: regression in child persistence into minutes of CI. These tests complete in
+#: about a second, so 30 s is still ~30x the observed work — far too wide to
+#: fire on slowness, narrow enough to keep the feedback loop short.
+DEADLOCK_GUARD_S = 30.0
 
 
 class ChangeSignal:
--- a/tests/unit/mobile/test_tui_ask.py
+++ b/tests/unit/mobile/test_tui_ask.py
@@ -134,14 +134,36 @@ class _Control:
 
 
 async def _control_for(app: OperatorApp, pilot) -> _Control:
-    for _ in range(50):
-        if app._mobile_registrant is not None:
+    """Connect to the app's control socket once its record is really published.
+
+    Waits on the DISCOVERY RECORD, not on the registrant object, and the
+    distinction is the whole bug this replaced. ``Registrant.start`` spawns a
+    daemon thread and returns; the record is written by ``_serve`` on that
+    thread. So ``app._mobile_registrant is not None`` means "registration was
+    requested", not "the record is on disk" — and the old helper waited for the
+    first and then immediately asserted the second, which is only true when
+    the thread happens to win the race.
+
+    On an idle machine it always does, which is why this reads as solid: 5/5
+    passes unloaded and 206/206 for the module at ``-n4``. CI runs ``-n auto``
+    (14 workers), and there it fails as ``no discovery record published`` —
+    observed on two unrelated PRs whose diffs touch no mobile code, and once
+    passing and failing on the SAME head. Polling the condition the assertion
+    actually depends on removes the gap entirely.
+
+    The bound is a deadlock guard rather than a timing assumption: it is
+    reached only when nothing ever publishes.
+    """
+    from local_operator.mobile import registry
+    from local_operator.mobile.registry import SessionRecord
+
+    records: list[tuple[SessionRecord, str]] = []
+    for _ in range(200):
+        records = registry.scan()
+        if records:
             break
         await pilot.pause(0.05)
     assert app._mobile_registrant is not None, "mobile registrant never started"
-    from local_operator.mobile import registry
-
-    records = registry.scan()
     assert records, "no discovery record published"
     record, _state = records[0]
     return await _Control.connect(record.control_port, record.control_key)
--- a/tests/unit/mobile/test_tui_ask.py
+++ b/tests/unit/mobile/test_tui_ask.py
@@ -28,6 +28,7 @@ from __future__ import annotations
 
 import asyncio
 import json
+import os
 
 import pytest
 
@@ -155,17 +156,36 @@ async def _control_for(app: OperatorApp, pilot) -> _Control:
     reached only when nothing ever publishes.
     """
     from local_operator.mobile import registry
-    from local_operator.mobile.registry import SessionRecord
 
-    records: list[tuple[SessionRecord, str]] = []
+    records = []
     for _ in range(200):
         records = registry.scan()
         if records:
             break
         await pilot.pause(0.05)
-    assert app._mobile_registrant is not None, "mobile registrant never started"
-    assert records, "no discovery record published"
-    record, _state = records[0]
+    # Reported against the registrant when there is one, because "started but
+    # never published" and "never started" are different failures and the
+    # second is the more likely one to introduce. Checked only on the failure
+    # path: once a record exists the registrant necessarily started, so
+    # asserting it first would have read as though it were still part of the
+    # wait — the exact misreading that produced the original bug (round 1, F4).
+    if not records:
+        assert app._mobile_registrant is not None, "mobile registrant never started"
+        raise AssertionError("registrant started but published no discovery record")
+    # This process's own record, not merely the first one scanned: ``scan`` is a
+    # global discovery read, so indexing it assumes nothing else has published.
+    # That holds under the suite's per-test config dir today, but matching on
+    # our own pid states the requirement rather than depending on it, and turns
+    # a stray record into a legible failure instead of a connection to someone
+    # else's socket (round 1, F6). Pid rather than control port because the port
+    # is stamped on the registrant's thread when the listener binds, so reading
+    # it here would reintroduce exactly the cross-thread race this helper fixes.
+    mine = os.getpid()
+    record = next((r for r, _state in records if r.pid == mine), None)
+    assert record is not None, (
+        f"no discovery record for this process (pid {mine}); "
+        f"scan saw pids {[r.pid for r, _ in records]}"
+    )
     return await _Control.connect(record.control_port, record.control_key)
 
 
--- a/tests/unit/providers/test_failover.py
+++ b/tests/unit/providers/test_failover.py
@@ -1811,20 +1811,30 @@ class TestTransientFailuresAreRetriedOnEveryCallPath:
             return OpenAICompatClient("https://api.test.example/v1", http_client=http)
 
         seen: list[Any] = []
-        with pytest.raises(ProviderError) as excinfo:
-            async for event in stream_with_failover(
-                _request(), FakeAuth({"openai": ["k"]}), {"retry": {"baseDelayMs": 1}}, client_for
-            ):
-                seen.append(event)
-        error = excinfo.value
-        assert error.status == 429
-        assert error.kind == "quota"
-        assert "Rate limit exceeded" in str(error)
-        # The visible delta was forwarded exactly once; nothing was replayed.
-        assert [e for e in seen if isinstance(e, StreamTextDelta)] == [
-            StreamTextDelta(delta="Gate")
-        ]
-        await http.aclose()
+        # Closed in ``finally`` — the shape every other client-owning test in
+        # this file already uses. With the close after the assertions, a
+        # failing assertion left the client open and the reader chasing an
+        # "unclosed client" warning attached to whichever later test happened
+        # to trip the collector, instead of the assertion that actually failed.
+        try:
+            with pytest.raises(ProviderError) as excinfo:
+                async for event in stream_with_failover(
+                    _request(),
+                    FakeAuth({"openai": ["k"]}),
+                    {"retry": {"baseDelayMs": 1}},
+                    client_for,
+                ):
+                    seen.append(event)
+            error = excinfo.value
+            assert error.status == 429
+            assert error.kind == "quota"
+            assert "Rate limit exceeded" in str(error)
+            # The visible delta was forwarded exactly once; nothing was replayed.
+            assert [e for e in seen if isinstance(e, StreamTextDelta)] == [
+                StreamTextDelta(delta="Gate")
+            ]
+        finally:
+            await http.aclose()
 
     async def test_the_one_shot_call_the_session_makes_is_marked_replayable(self) -> None:
         """Wiring, not policy: ``_one_shot_complete`` is the compaction summary
--- a/tests/unit/tui/test_subagent_stats.py
+++ b/tests/unit/tui/test_subagent_stats.py
@@ -87,8 +87,24 @@ async def _booted(app: OperatorApp) -> None:
     no elapsed-time comparison left on this path — a slower machine makes this
     slower, never red. It also cannot hang past the run: ``run_test`` cancels
     outstanding workers on exit.
+
+    The empty selection is checked rather than passed, because Textual's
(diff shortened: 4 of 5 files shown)


damianvtran:
### Agent review — round 1

Scope: 9c327496..4addc006

Test-only diff, 4 files. I read the diff, traced each new wait to the production
code that publishes the thing it waits on, mutation-tested three of the four
waits, ran the four gates by exit code, and exercised the changed tests under
artificial CPU load. Gates on the head tree: flake8 `rc=0`, black 26.1.0
`rc=0` (635 files), isort 5.13.2 `rc=0`, pyright `rc=0` (0 errors). CI: 18/18
green.

## The central question: does each wait target the assertion's actual condition?

I checked all four against the happens-before edge, not against the prose.

**`_priced` (`test_subagent_stats.py`) — correct, and for a non-obvious reason.**
This is the one that could most easily have been a proxy. The worry is real:
`_read_stats` (`local_operator/tui/widgets/subagent_panel.py:1588`) does the
pricing on the thread worker and hands the reading back through
`self.app.call_from_thread(self._stats_read, ...)`. If that hand-back were
fire-and-forget, "worker completed" would NOT imply "cache written" and the wait
would be a proxy. It is not: `call_from_thread` blocks the calling thread on the
callback's future (`app.py:1788`), so `_stats_read` has run on the UI thread
before the worker function returns, which is before the worker's task completes,
which is what `wait_for_complete` awaits. The edge holds. The trailing
`pilot.pause()` is belt-and-braces rather than load-bearing, and the helper's
own `assert ... .cost is not None` closes the residual gap by checking the state
rather than the signal — which is the right shape and is what makes the
mutation below fail loudly.

**`_booted` — correct.** `_boot_session` calls `_adopt_session` (`app.py:2747`)
inside the worker coroutine, so the session is set before the worker's task
completes. Waiting on the `session`-group worker is waiting on the write.

**`ChangeSignal` (#122) — correct, and the level-triggering matters.** Arming
before the predicate test (`_await_signalled`) is what makes this sound: a
publication landing between the test and the `await` stays on the flag. Watching
both sources is genuinely necessary, not belt-and-braces — `session_dir_of`
becomes non-`None` at `attach`, which only the parent stream carries, while
nested-child appends only reach the detail registry. The publications are
same-thread (children run as tasks on the test's loop; `notify_detail_persisted`
fires at `session.py:4498`/`4540` after the durable append), so there is no
cross-thread visibility question. `status_of(...) == "cancelled"` is fed by the
`SubagentEndEvent` emit, and `jobs.cancel` stamps `job.status` before its first
suspension point (`jobs.py:781`), so the status is already written when the
event that wakes the waiter is emitted. No ordering hole found.

**`_control_for` (`test_tui_ask.py`) — correct, and the best fix in the diff.**
The old helper waited on `app._mobile_registrant` (set when registration is
*requested*) and then asserted on `registry.scan()` (written later by `_serve`
on the registrant's own daemon thread). Two different facts. Polling `scan()`
directly waits on the exact thing the next line asserts. The diagnosis is right.

## Verification I ran

**Mutation tests — all three kill their target.** Not inferred; run.

1. Broke `OperatorApp._adopt_session` to never adopt: both `_booted` tests fail
   in 0.67s with `assert [] == ['j1','j2','j3']`. Fast, comprehensible.
2. Broke `_stats_read` so the worker completes without landing its reading —
   precisely the "worker finished but did not do its job" case `_priced` claims
   to catch. It fails with the author's own message: *"the stats worker finished
   without pricing the child"*. The guard is not decorative.
3. Made `registry.scan()` return `[]`: `test_tui_ask.py` fails in 12s with
   `no discovery record published`. Bounded, not a hang.

**#142 no-change argument — confirmed, the author is right.** I read
`test_loop_liveness.py` rather than accepting the claim. The only time calls in
the file are `time.thread_time()` (lines 103/106), inside `LoopCpuProbe`. That
is per-thread CPU, which by construction cannot grow because the machine is
busy — a sample is large only when the loop thread ran without yielding. There
is no wall-clock comparison left in either case. No change needed; the argument
holds on the code.

**Load check.** Head, 8 consecutive `-n0` runs of the three changed
`test_subagent_stats.py` cases under 20 busy processes at load average 75 rising
to 161: **8 passed, 0 failed**. Full changed modules at head: 335 passed
(`test_subagent_stats` + `test_comms` + `test_failover`), and `tests/unit/mobile`
206 passed. I could not reproduce the *base* failure under my synthetic CPU load
(base also went 8/8), so I can confirm the head is healthy but I am not
independently confirming the 5-of-10 base failure rate — CPU contention is
evidently not the same pressure as 10 concurrent pytest processes, and I did not
start a pool (this machine has OOMed doing that).

## Findings

**F1 — MINOR — `wait_for_complete([])` silently waits on ALL workers, not none.**
`tests/unit/tui/test_subagent_stats.py:_booted` / `_priced` /
`test_a_swept_child_is_dropped_from_the_stats_cache`.
Textual's implementation is `asyncio.gather(*[w.wait() for w in (workers or self)])`
— an **empty list is falsy**, so it falls through to `self`, every worker in the
manager. I instrumented the suite and this path is live today: in
`test_an_open_page_does_not_reinstate_the_per_event_repaint` the `session`-group
comprehension is already empty at the call (the boot worker completed and was
discarded by `_remove_worker`), so `_booted` there waits on whatever else happens
to be in the manager. It is harmless *now* — I verified via a slow-factory probe
that the empty case means "already finished" (`session_before=True`) and never
"not yet created", so the test still gets the state it needs. But the failure
mode is bad if it ever flips: an empty selection reads as a satisfied wait, which
is exactly the class this PR exists to remove, and it would fail silently by
passing. One line fixes it: assert the selection is non-empty before waiting
(e.g. `workers = [...]; assert workers or app._session is not None`), or capture
the worker object at spawn instead of filtering by group after the fact.

**F2 — MINOR — a wedged `ChangeSignal` costs 121 seconds per affected test.**
`tests/unit/harness/test_comms.py:156`. I mutated `_persist_new_messages` so the
child never persists, i.e. the signal never arrives. The test does fail, and the
message is genuinely good (*"published no change satisfying the condition in
120s — it is wedged, not merely slow"*), so the bounded-failure requirement is
met and this is not a hang. But it is 121s of wall clock, and both #122 tests
plus the two other `signal=`-bearing waits share the bound, so a real regression
in child persistence costs minutes of CI per test rather than seconds. The
`DEADLOCK_GUARD_S = 120.0` docstring argues the bound "can never be the
discriminator", which is true for *timing* but not for *cost*. 30s would still be
~100x the observed work (these tests pass in ~1s) while cutting the regression
feedback loop by 4x. Judgement call, not a defect.

**F3 — MINOR — the `test_a_live_selection_never_swallows_the_interrupt` claim
does not reproduce for me; please do not act on it yet.**
The PR body reports `[False]` failing **8/8 unloaded** and calls it a
deterministic defect in the binding chain. This is flagged in the PR as
important, so I tried to confirm it and could not. On the PR head, `[False]`
alone: **3/3 passed**. Both params: **4/4 passed** at `-n0` and **3/3 passed**
under default `-n auto`. I also copied the current `origin/main` version of
`test_transcript_selection.py` onto the head tree in case the head predates a
fix: **6/6 passed**. That is 13 runs, zero failures, unloaded.

Note the base here (9c327496) is behind current `origin/main` (c29110ff), and
cc8319d2 ("drop a dead predicate", #446) *did* touch this test — it rewrote the
two cases that asserted on the private `_copy_gesture` flag to assert on
`copy_in_flight`. So the most likely explanation is that the author measured
against a tree predating cc8319d2 and the thing they saw is already fixed on
`main`. Either way the "deterministic defect awaiting its own PR" conclusion
should not be carried forward on this evidence — worth re-measuring against
current `origin/main` before anyone opens that PR. This does not block *this* PR,
which does not touch that file.

**F4 — NIT — `_control_for` keeps an assertion whose wait it removed.**
`tests/unit/mobile/test_tui_ask.py:166`. The loop now polls `registry.scan()`,
but `assert app._mobile_registrant is not None, "mobile registrant never started"`
still runs first. It is now unreachable as a *diagnostic*: if the record was
published, the registrant necessarily started, so this can only fire in a state
that cannot occur. Harmless, and arguably documents intent — but the two lines
now read as though the registrant check is still part of the wait, which is the
misreading that caused the original bug.

**F5 — NIT — `records` is initialised then immediately overwritten.**
`tests/unit/mobile/test_tui_ask.py`: `records: list[...] = []` before the loop,
then `records = registry.scan()` as the loop's first statement. The initialiser
exists only to satisfy the post-loop reference when the loop body never runs,
which cannot happen with `range(200)`. Pyright is happy either way (`rc=0`).

**F6 — NIT — `records[0]` assumes this process published the only record.**
Same helper. `registry.scan()` is a global discovery scan; taking `[0]` picks an
arbitrary record if another session's record is visible. The module's 206 tests
pass and CI is green, so the registry is evidently isolated per test run — but
selecting by the app's own control port would state that assumption rather than
rely on it.

Nothing found on the `#224` change: moving `aclose()` into `finally` is
correct, matches the file's other 22 client-owning tests, and I confirmed it is
the only `aclose()` outside a `finally` in `test_failover.py`.

## On the honesty check

The author states plainly that #122 and the mobile helper were not reproduced
locally and rest on the mechanism. I think that is the right call and the
mechanical arguments hold on their own, for different reasons in each case.

For the **mobile helper** the argument is not really probabilistic: the old code
waited on fact A and asserted fact B, where B is written on another thread after
A. That is a bug on inspection regardless of whether it reproduces, and my
mutation confirms the new wait is load-bearing. Strong.

For **#122** the mechanism is sound — I traced every predicate to a publication
that precedes it, and the arming order is right — but it is worth being explicit
that "no elapsed-time comparison on the success path" is a *structural* claim,
and the structure is what I verified, not the flake rate. The residual risk is
not timing but coverage: if some future predicate has no corresponding
publication, the test converts from a 10s flake into a 120s wedge. F2 is the
cost of that, and the reason I would rather see the bound lower. The claim as
written is accurate and appropriately scoped.

I'd also record that the load evidence in the PR body is stronger than what I
could independently reproduce (F3 aside): I confirmed the head is clean at load
~160, but not the base's failure rate, because I would not start a pytest pool
on this machine. The mutation results are the part of my verification I would
lean on, and those are unambiguous.

## Verdict

**clean** — no BLOCKER, no MAJOR. 6 findings: 0 blocker, 0 major, 2 minor
(F1 empty-selection fallback, F2 120s wedge cost), 1 minor-as-correction
(F3, about a claim in the PR body rather than the diff), 3 nits.

The four waits each target the condition their assertions depend on; I checked
the happens-before edge in every case, including the one (`_priced`) where it
depends on a non-obvious `call_from_thread` property. Each new wait has a
bounded failure mode and produces a comprehensible message rather than a silent
hang — verified by mutation, not inferred. The #142 no-change argument is
correct on the code.

**This round is TERMINAL.** F1 through F6 are follow-ups, not a reason for
another round. F1 is the one I would take before merge if you want a one-line
change; F3 needs no code change here but should be re-measured before anyone
writes the follow-up PR it argues for.

damianvtran:

Head: `bbebf978` (rebased onto `7916e5ae`, clean, no conflicts).
Scope reviewed: `9c327496..4addc006`. Two commits added since: `bbebf978`.

| # | Sev | Disposition |
|---|-----|-------------|
| F1 | minor | **Fixed** — `bbebf978` |
| F2 | minor | **Fixed** — `bbebf978` (120s → 30s) |
| F3 | minor | **Fixed** — claim withdrawn and corrected; your conclusion was right, my cause was not |
| F4 | nit | **Fixed** — `bbebf978` |
| F5 | nit | **Fixed** — `bbebf978` |
| F6 | nit | **Fixed** — `bbebf978` |

**F1 — empty selection waits on all workers.** Fixed at all three sites. I
confirmed your instrumentation independently before fixing: a probe showed the
`session`-group selection is empty even on the *first* `_booted` call (the boot
worker had already completed and been dropped by `_remove_worker`), so this was
live on more paths than the one test you named. The selection is now checked
rather than passed, and where empty is legitimate the state that makes it
legitimate is asserted instead of assumed. Mutation-tested: with
`_adopt_session` stubbed out so nothing is pending and nothing is adopted, the
guard fails with *"no session worker is pending and no session was adopted — the
boot worker never ran, so waiting here would have waited on nothing"* rather
than passing on an unrelated worker.

**F2 — 121s wedge cost.** Agreed, and taken. `DEADLOCK_GUARD_S` is 30s. Your
distinction is the right one and I have written it into the docstring: the bound
can never be the discriminator for *timing*, but it is still a *cost*, paid per
wedged test across four waits. 30s is ~30x the observed work.

**F3 — the `test_a_live_selection_never_swallows_the_interrupt` claim.** You were
right that it does not reproduce, and I have withdrawn the claim. But the
proposed explanation (stale tree, fixed by cc8319d2) is not what happened, and
the difference matters:

```
current head (rebased):   6 runs, 0 failures
65ca130a:                 3 runs, 0 failures
75f867bf  (the tree my
  original run used):     4 runs, 0 failures
cc8319d2^ (before #446):  3 runs, 0 failures
```

It passes on the *exact tree I originally measured*, and `cc8319d2` did not
change the test body. So this was not a stale measurement — my original run was
simply wrong, and I cannot reconstruct how. Recorded that way rather than as
staleness, because "stale" would imply the measurement method was sound.

Then, under 24 CPU spinners (load 24 → 77):

```
6 runs at -n0 under load: 3 failed, 3 passed
all failures: interrupt[False], assert [] == ['interrupted']
```

**It is a load flake, so the original standing guidance to this board was
correct and my finding wrongly reversed it.** The PR body now says so plainly.

**F4 — unreachable registrant assertion.** Fixed. The registrant check now runs
only on the failure path, where "never started" and "started but never
published" are still distinct and the second is the more likely regression. On
the success path it is gone, so the two lines no longer read as though the
registrant were part of the wait — which was your point about the misreading
that caused the original bug.

**F5 — dead initialiser.** Fixed; the annotation went with it, so the
`SessionRecord` import it required is gone too.

**F6 — `records[0]` assumes a single publisher.** Fixed by matching this
process's own record. I selected on **pid, not control port**: the port is
stamped on the registrant's thread when the listener binds, so reading it from
the app object here would have reintroduced precisely the cross-thread race this
helper exists to remove. A stray record now produces
`no discovery record for this process (pid N); scan saw pids [...]` instead of a
connection to someone else's socket.

---

### Also fixed: `test_subagent_view.py`, a third mechanism

Raised mid-round as blocking other PRs' CI, and it is not the class this PR
started with. Two distinct defects:

1. `_job_with` assigned the module-level `TRAJECTORY` list **by reference**, and
   three tests mutate what they get back to drive a live update (two `append`,
   one `del [:3]`). Every later test folded a corrupted trajectory. That is the
   `[...'ToolCard'] == [...'AssistantBlock']` failure. Fixed by copying.
2. `test_history_unavailable_and_error_retry_keep_trajectory_fallback` asserted
   the unavailable note one frame after `_open`, but the flag is written by a
   `subagent-history` worker. This one *is* this PR's class. `_wait_history` was
   itself a 100-cycle budget poll; it now waits on the worker.

The "deterministic at `-n4`, invisible at `-n0`" framing was a sampling artefact
— base fails at `-n0` too when ordering is unlucky. Measured at `-n auto`
(14 workers, CI's profile), by exit code:

```
BASE (origin/main b56cf73f):  8 passed, 2 failed  of 10
HEAD:                        10 passed, 0 failed  of 10
```

---

### Gates at `bbebf978`

```
flake8 rc=0    black rc=0    isort rc=0    pyright rc=0
pytest tests/unit -n4: 9143 passed, 15 skipped — rc=0
```

Worth recording, since it validates the exit-code rule: while adding these fixes
an import landed in the wrong order, and `isort --check .` printed its usual
`Skipped 1 files` to **stdout** while returning **rc=1**, with the error on
stderr. Judged by stdout it read as clean.

### On your honesty note

Agreed, and I would go further: your mutation results are better evidence than
my load runs for the waits themselves, because they test the wait's *contract*
rather than sampling a symptom. The load runs are only worth something where
they compare two arms of the same procedure, which is why the `test_subagent_view`
numbers above are stated as base-vs-head rather than as a green head.

damianvtran:
### Agent review — round 2

Scope: 7916e5ae..bbebf978

A note on the scope line: the PR's `baseRefOid` still reads `ce1a8fd5`, but the
real merge base is `7916e5ae` (`origin/main` is 4 commits ahead of it). I
reviewed `7916e5ae..bbebf978` — test-only, 5 files, +388/-100, no
`local_operator/` change. Round 1's scope (`9c327496..4addc006`) is a different
tree; everything below was re-derived on this one.

I did not take the manager's reproduction or verification on trust. I built a
pristine `ce1a8fd5` worktree and re-ran both arms myself.

## The new commit is the real work of this round, and it is two defects, not one

`bbebf978` is a third mechanism: not "waited on the wrong signal" (round 1) and
not "wall-clock budget" (round 1), but **test outcome depends on execution
order**. I confirmed it decomposes into two independent defects with different
causes, and — importantly — **each half is separately load-bearing.**

**Defect A, shared mutable state.** `_job_with` assigned the module-level
`TRAJECTORY` by reference. I pinned the leak to a single deterministic ordered
pair rather than inferring it from the symptom:

```
BASE, -n0, two tests, in order:
  test_the_page_opens_with_the_instruction_the_parent_delegated   (appends c9)
  test_the_page_renders_the_childs_messages_and_tool_calls
  -> 1 failed, 1 passed   ("Left contains one more item: 'ToolCard'")

HEAD, same pair, same order:
  -> 2 passed
```

That is the whole bug in two tests and 1.3 seconds. The corrupter is the
`job.trajectory.append(_call("c9", ...))` at line 1999, which is the **only one
of the five mutation sites in the file that reaches a `_job_with`-built job**.
I checked the other four rather than assuming: lines 1611, 1650, 1662 and 1908
all build their job as a bare `_Job(...)` with a freshly-constructed literal
trajectory, so they never touched the shared list. I verified this empirically
too — running the `del job.trajectory[:3]` test (1908) ahead of the same victim
passes on base. So the docstring's claim of "three tests mutate what they get
back" is **wrong on the code**: exactly one does. See F7.

The fix is a shallow copy at the single construction point. That is the right
scope: it isolates at the boundary where the shared object enters a test, so a
new test cannot reintroduce the leak by writing the obvious thing. It is a real
fix, not a reordering or a serialisation — nothing in the diff pins execution
order, adds a lock, or marks anything `serial`, which is what masking would
have looked like.

I also audited the one remaining path that reads `TRAJECTORY` without going
through `_job_with` (line 190, `JobState(trajectory=TRAJECTORY)` in
`test_fold_accepts_immutable_mapping_and_sequence_contracts`). It is safe, and
for a reason worth recording: `JobState` is a pydantic model that coerces the
list into a `Sequence`/tuple-backed field, and that test asserts exactly that
immutability contract. No copy needed there.

**Defect B, the history worker wait.** Distinct cause, distinct fix:
`test_history_unavailable_and_error_retry_keep_trajectory_fallback` asserted
one frame after `_open`, but `_history_unavailable` is written on a
`subagent-history` worker.

## Mutation tests: four mutations, and one of them changed my reading

This is the part I was asked not to skip, and it earned its keep.

**M1 — remove the copy (`list(trajectory)` → `trajectory`), keep everything
else.** The pinned pair fails again: `1 failed, 1 passed`, same
`+ 'ToolCard'`. **Kills its target.**

**M2 — replace the whole new `_wait_history` body with a bare `await
pilot.pause()`.** The target test passed **6/6 at `-n0`**, and the whole file
passed **104/104 at `-n4`, 3 runs of 3.** This mutation SURVIVED. On its own
that reads like the classic "guard that cannot fail at its own boundary" — but
it is not, and the next two mutations are why.

**M3 — keep the new `_wait_history` implementation, delete only the added call
site in the unavailable test.** Whole file at `-n4`: **1 failed, 104 passed, 1
failed** across 3 runs. So the added call is load-bearing.

**M4 — keep the copy fix AND the added call site, but revert `_wait_history`'s
body to the old 100-cycle budget poll.** Whole file at `-n4`: **1 failed, 2
failed, 1 failed — 3 runs of 3.** So the rewrite of the body is *also*
load-bearing, independently.

M3 and M4 together are the answer to M2: the reason M2 survived is that with
the *call site* present, a bare pause happens to be enough on this box for the
`to_thread` read of a missing directory to land, but that is precisely the
frame-budget bet the PR exists to delete. M4 is the decisive one — the old poll
fails 3/3 even with the copy fix in place, which proves defect B is real and
not a shadow of defect A. Both halves of the commit are necessary; neither is
decoration.

## Confirming the manager's numbers independently

**Reproduction on pristine `ce1a8fd5`, `-n4`, `-p no:randomly`:**

```
run 1: 1 failed, 103 passed
run 2: 1 failed, 103 passed
run 3: 2 failed, 102 passed
```

Both named failure modes appear: `assert ['AssistantBl...', 'ToolCard'] ==
[...]` / "Left contains one more item: 'ToolCard'", and `assert 'history
unavailable' in 'read-only'`. So the defect reproduces — **but the count is not
the stable "2 failed" reported.** It is 1, 1, 2. That matters only for how the
evidence is phrased, not for the conclusion.

**Verification on `bbebf978`, `-n4`:** `104 passed` **3 runs of 3.** Confirmed,
matching the manager exactly.

**One correction to the framing, and it favours the PR.** The "deterministic
xdist-ordering defect" description undersells it: base fails at **`-n0`** too.
Three serial runs of the full file on base: `104 passed`, then `1 failed`, then
`1 failed`, always
`test_history_unavailable_and_error_retry_keep_trajectory_fallback`. xdist
changes which symptom surfaces, not whether the file is broken. The author's
own remediation comment already says this ("base fails at `-n0` too when
ordering is unlucky"), so the author has it right and only the task framing was
narrow. Head at `-n0`: **104 passed.**

## F1 and F3 from round 1

**F1 — fixed, and at every site.** `wait_for_complete` is called at four places
across the touched files, all now guarded by `if workers:`:
`test_subagent_stats.py:100-102` (`_booted`), `:126-128` (`_priced`),
`:1481-1483` (sweep), and `test_subagent_view.py:890-892` (`_wait_history`).
I grepped all five touched files: there is no unguarded or empty-selection call
left. I re-confirmed the underlying hazard against Textual's installed source
rather than the round-1 quote — `asyncio.gather(*[worker.wait() for worker in
(workers or self)])` — so an empty list does fall through to every worker in
the manager.

`_booted` goes further than the finding asked: where the empty selection is
legitimate it asserts the state that makes it legitimate, with a message that
names the failure ("...waiting here would have waited on nothing"). That is the
right shape — it checks state rather than a signal.

**F3 — fixed, and better than what I proposed.** The PR body now carries a
`### CORRECTION` section that withdraws the 8/8 claim outright. It also
corrects **my** hypothesis: I guessed a stale tree fixed by `cc8319d2`, and the
author disproved that by re-measuring on the exact original tree (`75f867bf`,
4 runs, 0 failures) and confirming `cc8319d2` did not change the test body. 16
unloaded runs across four trees, zero failures. The section then re-measures
under load (3 of 6 fail) and lands it in #142's class. Nobody reading this will
chase a deterministic defect that does not exist, which is what F3 asked for.
Recording it as "my original run was simply wrong" rather than as staleness is
the more honest framing and I accept the correction.

**F2, F4, F5, F6** are dispositioned in the remediation and I spot-checked
rather than re-reviewed: `DEADLOCK_GUARD_S` is 30.0, the registrant assertion
now runs only on the failure path, the dead initialiser is gone, and record
selection is by pid. The pid choice over my suggested control port is correct
and better reasoned than my finding — the port is stamped on the registrant's
thread, so reading it would have reintroduced the very cross-thread race the
helper removes.

## What I am relying on round 1 for, explicitly

I did **not** re-derive these; they were verified in round 1 against code that
this commit does not touch, and I re-confirmed only the one that rests on a
non-obvious property:

- The happens-before edges on the four original waits (`_priced`, `_booted`,
  the #122 `ChangeSignal`, `_control_for`). I *did* re-check the `_priced` edge
  because it is the fragile one, and it holds: Textual's `App.call_from_thread`
  ends in `future.result()`, a blocking call, so `_stats_read` has run on the
  UI thread before the worker function returns. Verified against the installed
  Textual source this round, not carried over.
- The **#142 no-change argument** for `test_loop_liveness.py` (only time calls
  are `time.thread_time()`, per-thread CPU, no wall-clock comparison). Carried
  over from round 1 unchanged; the file is not in this diff.
- Round 1's three mutation kills on `_adopt_session`, `_stats_read` and
  `registry.scan()`. Carried over; those helpers are unchanged except for the
  F1 guard, which I checked directly above.

## Gates and CI at `bbebf978`

Run over the whole tree, judged by exit code:

```
flake8   rc=0
black 26.1.0 --check   rc=0
isort 5.13.2 --check   rc=0
pyright  rc=0   (0 errors, 0 warnings, 0 informations)
```

Targeted suites on head: the four other touched files `342 passed` at `-n4`;
`test_subagent_view.py` `104 passed` at `-n4` (3/3) and `104 passed` at `-n0`.
I did not start a whole-suite pool.

**CI on `bbebf978`: 18/18 green.** `lint` pass (59s), `test (3.12)` pass
(23m0s), `test (3.13)` pass (22m4s); also `type-check`, `pip-audit`,
`cli-sanity`, `server-sanity` and `tui-e2e` on both `macos-latest` and
`ubuntu-latest`.

## Findings

**F7 — NIT — the `_job_with` docstring names three mutating tests; only one of
them can mutate the shared list.** `tests/unit/tui/test_subagent_view.py:142-145`
says "three tests mutate what they get back to drive a live update:
`job.trajectory.append(...)` at the 'mounted once' and 'coalesced frame' cases,
and `del job.trajectory[:3]` at the eviction case." Of those, only the "mounted
once" case (`test_the_page_opens_with_the_instruction_the_parent_delegated`,
line 1999) builds its job through `_job_with`. The coalesced-frame case (1611)
and the eviction case (1908) both construct a bare `_Job` and assign a fresh
literal trajectory, so they never touched `TRAJECTORY`; I confirmed by running
the eviction case ahead of the victim on base, which passes. The fix is right
and its scope is right — only the explanation over-counts. Since this docstring
exists specifically so a future reader understands the leak, an inaccurate
census of it is worth one line to correct: name the single call site.

**F8 — NIT — the commit message and PR body say the base fails "2 failed at
-n4"; I measured 1, 1, 2 over three runs.** The failure count is itself
order-dependent, which is the point of the defect. Phrasing it as a range, or
as "1-2 failures depending on distribution", would stop the next person
concluding their tree is unaffected when they see only one red test. Does not
affect the diff.

No BLOCKER. No MAJOR. No MINOR.

## Verdict

**clean** — 0 blocker, 0 major, 0 minor, 2 nits (F7, F8), both about wording in
a docstring and a commit message rather than about behaviour.

The new isolation fix addresses the actual leak rather than masking it: the
shared object is copied at the single point where it enters a test, no
execution order is pinned, and I verified no other test in the touched files
still depends on the shared list. Both halves of the commit are independently
load-bearing, proven by mutation (M3 and M4 each fail 3/3 at `-n4`), and the
one mutation that survived (M2) is explained by M4 rather than left hanging.
F1 is fixed at all four call sites with the empty-selection hazard re-confirmed
against Textual's own source, and F3's stale claim is withdrawn and corrected
in the PR body — including a correction to my own round-1 hypothesis that I
accept.

**This round is TERMINAL.** F7 and F8 are follow-ups, not a reason for another
round; neither needs a code change to ship, and if the author wants to take F7
it is a one-line docstring edit that does not invalidate this review.

PR #455, opened 2026-08-30
perf(tui): stop repainting unchanged chrome and animating blurred terminals

damianvtran:
# PR Description

## Summary of Changes

Five measured reductions in what a session costs when **nothing has changed**:
unchanged chrome rows that were being rewritten 12.5 times a second, a missing
`layout=False` that reflowed the whole screen on every spinner tick, and
animation that ran at full rate in terminals nobody was looking at.

### Read this part first: the honest framing

The user reported four symptoms — sessions with many concurrent subagents
rendering very slowly, subagent views lagging more than parent views, many
parallel sessions slowing things down, and more lag in cmux than Ghostty.

**This PR does not fix that felt slowness, and does not claim to.** The
dominant cause was measured and it is *environmental*, not code:

- the machine sat at **loadavg 260-307 on 14 cores** from other sessions'
  pytest/xdist harnesses and CPU burners;
- all 18 lop TUIs *together* summed to **under one core**;
- a fixed work quantum held its **CPU time** stable (81.6 → 101.5 ms) while its
  **wall time** inflated **4.6-7.8x** under that load;
- a 12.5 fps timer doing the child's real 11.5 ms of work ran at **0.30
  effective fps with 100% of frames late**.

Code was roughly **1.7x**; the environment was roughly **100x**.

What this PR does is make lop **cheaper and better-behaved**, which matters most
exactly when the box is busy and when many sessions are open. On symptom 4
specifically: the steady-state stream is 48% escape-sequence bytes (61,660 of
129,110 — 1,119 CSI cursor-position, 3,450 SGR, 1,725 truecolor bg, 1,181
truecolor fg), and cmux (Electron, JS parsing) held 51-60% of a core where
Ghostty (native GPU) held 0.6%. That gap is not lop's to close — but emitting
**~28x less per background session** is squarely lop's to do.

Also **confirmed and deliberately not touched**: subagent scaling is cleanly
linear in K (`SubagentRow.paint` per 40 ticks: K=1→63, K=4→280, K=8→504,
K=16→1034, against x16.00 in K — no O(n²)); screen arrange cost is flat in
transcript size N (~455 µs at N=20/40/80/160/320); and roster persistence is
already coalesced, fingerprinted and off-loop.

### 1-3. Chrome paint (`subagent_view._paint_chrome`)

- **The breadcrumb omitted `layout=False`** while the title and rule beside it
  passed it. All three are `height: 1` in the sheet. `Static.update(layout=True)`
[...]

PR #462, opened 2026-08-30
fix(mcp): make an imported remote server authenticable and say how

damianvtran:
# PR Description

## Summary of Changes

Follow-up to the Codex MCP import (#367, shipped by #439 in v0.43.14). The
import works — the servers are visible — but the imported **remote** servers
cannot actually be *used*, and the error the user is shown is a dead end:

```
× MCP gitlab failed: Server returned an error response
× MCP launchdarkly failed: unauthorized access
× MCP minerva-qa failed: Server returned an error response
```

This is a follow-up, not a revert: nothing from #439 is undone.

### Root cause

Codex's remote entries carry only a `url`:

```toml
[mcp_servers.gitlab]
url = "https://gitlab.com/api/v4/mcp"
```

Codex holds its OAuth grants elsewhere, so its format has no auth block to
copy. Local-operator signals OAuth with an explicit `auth = {"type": "oauth"}`,
and **every** auth-capable path gated on exactly that: `_build_oauth_auth` and
`_ensure_oauth_fresh` returned early, and the TUI's `_resolve_mcp_server`
refused `/mcp login` with *"does not use OAuth login."* — the command that
fixes the failure told the user it did not apply.

So the server connected with no credentials, 401'd, and the SDK reported it as
raw transport text.

Confirmed live rather than reasoned about, with one real `initialize` POST per
server:

| server | status | `WWW-Authenticate` |
|---|---|---|
| gitlab | 401 | `Bearer realm="GitLab", resource_metadata=…` |
| launchdarkly | 401 | `Bearer resource_metadata=…` |
| minerva-qa | 401 | `Bearer resource_metadata=…, scope="mcp:read"` |
| datadog | 401 | **none** |
| openaiDeveloperDocs | 200 | — (needs no auth) |

Two shapes there shaped the design: a 401 with no challenge header at all, and
a server that must keep working unauthenticated.

### Half 1 — an imported remote server is authenticable

This wires existing machinery (`discover_oauth_endpoints`, the SDK's
`OAuthClientProvider`, the token store). **No second OAuth path.**

A remote server now qualifies for the provider when any of three things is
true: it declares `auth.type: oauth`, it already holds a stored grant, or it
has been observed answering a 401/403 whose RFC 9728/8414 discovery found an
authorization server. The rule is **transport-level and names no config
source** — nothing branches on Codex, and any format that omits an auth block
benefits identically.

The three sources are complementary, not redundant: the observed-challenge
ledger is per-process, so the stored grant is what makes a **restart**
re-authenticate instead of connecting unauthenticated all over again.

`/mcp login` applies a deliberately **permissive** version of that gate
(`deliberate=True`, any remote server): whether a server needs OAuth cannot be
answered without a network round trip, so the strict test would refuse the very
first login on a fresh import — the dead end this PR exists to remove. It is
safe rather than merely convenient, because the SDK's provider only begins a
grant in response to a real 401. The **picker and logout stay strict**, since a
suggestion list is a claim that those servers have something to log into.

Preserved: startup and auto-reconnect remain non-interactive and never open a
browser; a server needing no auth connects unauthenticated with no discovery,
no prompt and no added latency; Codex's own stored tokens are never read.

### Half 2 — the failure names the fix

The SDK **erases the status code**: a 401 it cannot resolve arrives from
`session.initialize()` as `MCPError(-32603, 'Server returned an error
response')` with no status and no headers. That is precisely the opaque text
the user screenshotted. So a small httpx response hook observes the status at
the transport, before it is flattened, and the connect's error path converts
only what the hook actually saw.

The verb is chosen from what is stored, not guessed:

- **`reauth`** when a stored grant was just rejected — reaching this error
  means the grant could not be refreshed, so the credential on disk is stale
  and a plain `login` would leave it in place;
- **`login`** when nothing is stored — never authorized here;
- neither when discovery found **no** OAuth endpoint (Datadog's shape). Saying
  "run /mcp login" there would promise a grant that cannot be completed, so it
  says what is true and names the config headers instead.

A bare `client_info` row does **not** count as a grant — the SDK writes that on
mere discovery, and counting it told a user who had never logged in that their
authorization had expired.

Wording obeys the constraints documented on `_auth_required_text`: the command
leads so it survives truncation on a narrow terminal (D1), and there is at most
one `—` (D4).

**Non-auth failures are never relabelled.** No observed challenge means no
conversion, so a down server, DNS or TLS failure keeps its own message —
verified byte-identical against `origin/main`.

## Impact

- No breaking changes, no new dependency, no new public API.
- No behaviour change for servers that already declared `auth.type: oauth`.
- Version bumped to `0.43.17` — **patch**. Two contained fixes to an existing
  subsystem; the product gains no new surface a user would adopt.

## Testing Details

> Server names and URLs appear below (they are already in the issue thread).
> **No tokens, no env values, and no contents of `~/.codex/config.toml` or
> `~/.codex/auth.json`.** No real OAuth grant was completed — see "Not verified
> locally".

All probes ran from a worktree off `origin/main` with an assertion that the
imported modules resolve to **that worktree**, because a prior agent silently
tested a different checkout and got a false result.

Full artifacts in `docs/evidence/codex-mcp-auth/`.

### Before / after, real servers

```
BEFORE (released behaviour)                     AFTER
× gitlab: Server returned an error response  →  × gitlab: run /mcp reauth gitlab
                                                    — authorization expired
× launchdarkly: unauthorized access          →  × launchdarkly: run /mcp login
                                                    launchdarkly to authorize
× minerva-qa: Server returned an error resp. →  × minerva-qa: run /mcp login
                                                    minerva-qa to authorize
× datadog: Server returned an error response →  × datadog: run /mcp login
                                                    datadog to authorize
```

`gitlab` says **reauth** and the rest say **login** because gitlab is the one
holding a token payload that could not be refreshed. That is read from the
credential store.

### All four required classes

1. **OAuth-challenging 401** — gitlab / launchdarkly / minerva-qa, above.
2. **401 with no challenge header** — datadog. Discovery still finds an
   authorization server, so the login command is honest.
3. **401 with no discoverable OAuth at all** — no public server in the matrix
   has this shape, so it is exercised against a local stub that 404s every
   `.well-known` probe. It must not promise a login:
   `apikey-only rejected our credentials (401) — set its API key or headers`.
4. **No auth needed** — `openaiDeveloperDocs` still connects with **5 tools**,
   no provider attached, no discovery, connect time unchanged within network
   noise.

### Non-auth failures, compared against `origin/main`

Connection refused, DNS NXDOMAIN, HTTP 500 and HTTP 404 produce **identical**
output before and after. A network outage never becomes "run /mcp login".

### The `/mcp login` dead end

| config | before | after |
|---|---|---|
| gitlab (url only) | refused | **accepted** |
| launchdarkly (url only) | refused | **accepted** |
| openaiDeveloperDocs | refused | accepted (no-op: provider acts only on a real 401) |
| explicit `auth: oauth` | accepted | accepted |
| stdio | refused | refused |

### Rendered frames

`docs/evidence/codex-mcp-auth/frames/{before,after}.svg`, captured from the real
`OperatorApp` (which loads `local_operator.tcss`) at 100x30 per AGENTS.md
"Visual validation", then rendered and **looked at**. The before frame
reproduces the user's screenshot. Geometry identical in both:
`virtual_size == size == (98, 28)`, no scrollbar.

Looking at the first after-capture caught a defect the numbers did not: an
early wording that included `(401)` pushed two lines past the width and orphaned
a bare `(401)` onto its own centred row. The status code added width without
adding action, so the OAuth wording now delegates to the existing
`_auth_required_text` and every line fits on one row.

### Gates (whole tree, exactly as CI runs them)

```
.venv/bin/python -m flake8 .                      clean
uvx --from black==26.1.0 black --check .          633 files unchanged
uvx isort==5.13.2 --check .                       clean
.venv/bin/python -m pyright --pythonpath …  .     0 errors
pytest tests/unit                                 9028 passed, 15 skipped
```

18 new tests cover the capability gate (including that a bare client
registration is not a grant, and that stdio is still refused), the verb choice,
the D1/D4 wording constraints, the metadata-probe exclusion, and that a
non-auth failure is never converted.

### Not verified locally

**No real OAuth grant was completed** — that needs the operator's browser and
credentials, so I did not run one. What is proven is that `/mcp login <name>`
now *reaches* the grant flow for these servers instead of being refused, and
that a non-interactive startup still raises rather than opening a browser. The
end-to-end "browser opens, token lands, server connects" step is the one thing
left for the operator to confirm.

One pre-existing flake was observed and is **not** from this branch:
`tests/unit/session/test_resume_subagents_todos.py::test_live_progress_never_schedules_the_roster_writer`
is order-dependent — it passes in isolation and in the final full run here, and
reproduces on base under the right ordering.

## Related

- Issue #367 (Codex MCP import) — this makes the imported remote servers usable.
- PR #439 (the import itself) — context, not reverted.

### DIFF
--- a/docs/evidence/codex-mcp-auth/README.md
+++ b/docs/evidence/codex-mcp-auth/README.md
@@ -177,3 +177,64 @@ wording — so the design round's frames remain current:
 
 Direct 401, 403, public 200, and 404/500 non-auth parity all unchanged, and
 startup still opens zero browsers.
+
+## Round 2 remediation (F5–F6)
+
+**F5 — a challenge could still relabel a later NETWORK failure.** Round 1
+cleared the latch only when a later *response* arrived on the endpoint. A retry
+that dies at DNS/connect/TLS/read time produces no response at all, so no hook
+ran and the earlier 401 stayed latched — the terminal network failure was
+reported as `/mcp login`.
+
+The verdict is now bound to the **request**, not the last response: a
+`request` event hook (`_AuthChallengeWatcher.begin`) clears the slot as each
+same-endpoint request starts, and `observe` sets it only if that request comes
+back 401/403. Whatever happens next — a response, or a socket error that never
+produces one — no stale verdict survives. A request to any other URL (the
+provider's discovery and token traffic) neither sets nor clears.
+
+Proven end to end with a real `httpx` client and a real server:
+
+```
+after the 401           : watcher verdict = 401
+retry failed with       : ReadTimeout (no response)
+after the dead retry    : watcher verdict = None
+classified as           : None      (NOT auth advice)
+```
+
+**F6 — the login gates dropped the manager's effective store.** Both
+`_mcp_login_worker` and `_run_mcp_grant_on_owner` called
+`probe_oauth_capability(cfg)` with no store, so a server whose grant exists
+only in an injected store was refused as "does not use OAuth login" whenever
+discovery was unavailable — even though the same manager had just classified
+its failure as `reauth` from that store.
+
+Eligibility is now a manager-owned operation,
+`McpManager.server_supports_oauth_login`, which passes its own effective store
+through. Both entry points call it via one TUI helper; a reduced follower
+facade that does not implement it falls back to the store-less probe, which is
+correct because such a facade owns no store to consult.
+
+With discovery offline and a grant in a custom store:
+
+```
+probe_oauth_capability(cfg)                  -> False   (the old call)
+manager.server_supports_oauth_login(cfg)     -> True    (consults its store)
+```
+
+### Parity preserved
+
+The live matrix and the whole transport matrix are byte-identical to round 1 —
+same verbs, same wording, so the terminal design round remains current:
+
+```
+✓ openaiDeveloperDocs: connected
+× datadog:      run /mcp login datadog to authorize
+× gitlab:       run /mcp reauth gitlab — authorization expired
+× launchdarkly: run /mcp login launchdarkly to authorize
+× minerva-qa:   run /mcp login minerva-qa to authorize
+
+F1-redirect-401    McpAuthChallengeError    parity-500  MCPError (unchanged)
+parity-direct-401  McpAuthChallengeError    parity-404  MCPError (unchanged)
+parity-403         McpAuthChallengeError
+```
--- a/docs/evidence/codex-mcp-auth/README.md
+++ b/docs/evidence/codex-mcp-auth/README.md
@@ -1,0 +1,117 @@
+# Codex-imported MCP servers: authenticable + actionable failures
+
+Evidence for the follow-up to issue #367 / PR #439 (Codex MCP import, shipped in
+v0.43.14). The import worked; the imported remote servers could not be *used*,
+and the failure text was a dead end.
+
+All probes ran against a worktree off `origin/main`, with an assertion that the
+imported modules resolve to that worktree rather than to another checkout.
+
+## Root cause (confirmed live, not assumed)
+
+Codex's remote entries carry only a `url` — that tool holds its OAuth grants
+elsewhere, so its config has no auth block. Local-operator gated every
+auth-capable path on an explicit `auth.type == "oauth"`, so those servers
+connected with **no credentials**, the server 401'd, and the SDK reported it as
+raw transport text.
+
+`probe-live-401s.txt` — one real `initialize` POST per server:
+
+| server | status | `WWW-Authenticate` |
+|---|---|---|
+| gitlab | 401 | `Bearer realm="GitLab", resource_metadata=...` |
+| launchdarkly | 401 | `Bearer resource_metadata=...` |
+| minerva-qa | 401 | `Bearer resource_metadata=..., scope="mcp:read"` |
+| datadog | 401 | **none** |
+| openaiDeveloperDocs | 200 | — (needs no auth) |
+
+`probe-discovery.txt` — RFC 9728/8414 discovery, which decides the wording.
+Note Datadog: **no challenge header, but discovery still succeeds**, so the
+header alone is not a usable signal.
+
+A second fact drove the design: the SDK destroys the status code. A 401 it
+cannot resolve arrives from `session.initialize()` as
+`MCPError(-32603, 'Server returned an error response')` with no status, no
+headers. That is why detection is a transport-level response hook.
+
+## Before / after
+
+`before-connect.txt` (released behaviour) vs `after-messages.txt`:
+
+```
+BEFORE                                          AFTER
+× MCP gitlab failed: Server returned an       → × MCP gitlab failed: run /mcp reauth gitlab
+      error response                                — authorization expired
+× MCP launchdarkly failed: unauthorized       → × MCP launchdarkly failed: run /mcp login
+      access                                        launchdarkly to authorize
+× MCP minerva-qa failed: Server returned an   → × MCP minerva-qa failed: run /mcp login
+      error response                                minerva-qa to authorize
+× MCP datadog failed: Server returned an      → × MCP datadog failed: run /mcp login
+      error response                                datadog to authorize
+```
+
+`gitlab` says **reauth** and the others say **login** because gitlab is the one
+holding a stored token payload that could not be refreshed. That distinction is
+read from the credential store, not guessed.
+
+## All four required classes
+
+1. **OAuth-challenging 401** — gitlab / launchdarkly / minerva-qa, above.
+2. **401 with no challenge header** — datadog, above. Discovery still finds an
+   authorization server, so the login command is honest.
+3. **401 with no discoverable OAuth at all** — no public server in the matrix
+   has this shape, so `after-no-discovery.txt` records it against a local stub
+   that 404s every `.well-known` probe. It must not promise a login:
+   `apikey-only rejected our credentials (401) — set its API key or headers`.
+4. **No auth needed** — `openaiDeveloperDocs` still connects with 5 tools, no
+   OAuth provider attached and no discovery (`noauth-server-latency.txt`).
+   Connect time is unchanged within network noise (2.79s after vs 2.37s before,
+   and 1.87s vs 2.45s on an earlier run — the ordering flips between runs).
+
+## Non-auth failures are NOT relabelled
+
+`nonauth-parity.txt` compared this worktree against `origin/main`. Connection
+refused, DNS NXDOMAIN, HTTP 500 and HTTP 404 produce **byte-identical** output
+before and after. A network outage never becomes "run /mcp login".
+
+## The `/mcp login` dead end
+
+`after-login-gate.txt`. Before, `_resolve_mcp_server` refused every Codex import
+with *"does not use OAuth login."* — the command that fixes the 401 told the
+user it did not apply.
+
+| config | before | after |
+|---|---|---|
+| gitlab (url only) | refused | accepted |
+| launchdarkly (url only) | refused | accepted |
+| openaiDeveloperDocs | refused | accepted (no-op: provider only acts on a real 401) |
+| explicit `auth: oauth` | accepted | accepted |
+| stdio | refused | refused |
+
+## Rendered frames
+
+`frames/before.svg` / `frames/after.svg`, captured from the real `OperatorApp`
+(which loads `local_operator.tcss`) at 100x30 and viewed as PNG. The before
+frame reproduces the user's screenshot. Geometry identical in both:
+`virtual_size == size == (98, 28)`, no vertical scrollbar.
+
+The first after-capture exposed a defect the numbers did not: including
+`(401)` pushed two lines past the width and orphaned the bare `(401)` onto its
+own centred row. The status code told the user nothing the verb did not, so the
+OAuth wording now delegates to the existing `_auth_required_text` and every line
+fits on one row.
+
+## Not verified locally
+
+No real OAuth grant was completed — that needs the operator's browser and
+credentials. What is proven is that `/mcp login <name>` now reaches the grant
+flow for these servers instead of being refused, and that a non-interactive
+startup still raises rather than opening a browser.
+
+## Reproducing
+
+The probe scripts were scratch and are deliberately not committed — the outputs
+above are the evidence. Each is reproducible from a worktree off `origin/main`:
+drive `McpManager._connect_server` / `_connect_round` against the server URLs in
+the table, and render the frames with `OperatorApp.save_screenshot` per
+AGENTS.md "Visual validation" (the real app, so the stylesheet is applied).
--- a/docs/evidence/codex-mcp-auth/README.md
+++ b/docs/evidence/codex-mcp-auth/README.md
@@ -115,3 +115,65 @@ above are the evidence. Each is reproducible from a worktree off `origin/main`:
 drive `McpManager._connect_server` / `_connect_round` against the server URLs in
 the table, and render the frames with `OperatorApp.save_screenshot` per
 AGENTS.md "Visual validation" (the real app, so the stylesheet is applied).
+
+## Round 1 remediation (F1–F4)
+
+Four majors from the agent review, each fixed at source with a regression test
+verified to fail against the pre-fix code.
+
+**F1 — a challenge after a redirect was missed.** The hook compared the
+response's request URL to the configured string, so a server canonicalizing
+`/mcp` → `307 /mcp/` and challenging on the redirect target did not match, and
+the user kept getting the opaque error. Endpoint identity is now semantic
+(scheme/host lowercased, query dropped, trailing slash stripped).
+
+Reproduced end to end through the real SDK transport against a local server:
+
+```
+BEFORE  F1-redirect-401   MCPError               -> "Server returned an error response"
+AFTER   F1-redirect-401   McpAuthChallengeError  -> names the fix
+        hook saw 307 http://127.0.0.1:8902/redir
+        hook saw 401 http://127.0.0.1:8902/redir/
+```
+
+**F2 — an earlier 401 could relabel a later non-auth failure.** `status_code`
+latched on the first challenge and was never cleared, so a 401 followed by a
+500 on the same endpoint still reported as authorization — the exact
+misrouting this feature promises not to do. The **last** response on the
+endpoint now wins; a 3xx hop is not a verdict and never clears a pending one.
+
+```
+after 401 -> status_code=401
+after 500 -> status_code=None
+classified as: None      (stays a network/server error)
+```
+
+**F3 — `/mcp login` was enabled for every remote config.** The `deliberate`
+shortcut accepted API-key and known-public servers. It is removed. A stdio
+server or an explicit non-OAuth `auth.type` is now a hard refusal that costs no
+network call, and the remaining unknown case is settled by one discovery probe
+(`probe_oauth_capability`) before any grant starts — so the first login on a
+fresh import still works, without claiming every URL is authenticable.
+
+**F4 — the classifying store's verdict was discarded.** `_challenge_error`
+resolved `has_stored_grant` against the manager's own (possibly injected)
+store, then `_auth_required_text` threw it away and re-read the default machine
+store, rendering `login` for a server demonstrably holding a grant. The carried
+fact is now used; only the legacy `McpAuthRequiredError`, which has no such
+field, falls back to a lookup.
+
+### Parity preserved
+
+The live matrix is byte-identical to the pre-remediation run — same verbs, same
+wording — so the design round's frames remain current:
+
+```
+✓ openaiDeveloperDocs: connected
+× datadog:      run /mcp login datadog to authorize
+× gitlab:       run /mcp reauth gitlab — authorization expired
+× launchdarkly: run /mcp login launchdarkly to authorize
+× minerva-qa:   run /mcp login minerva-qa to authorize
+```
+
+Direct 401, 403, public 200, and 404/500 non-auth parity all unchanged, and
+startup still opens zero browsers.
--- a/docs/evidence/codex-mcp-auth/after-login-gate.txt
+++ b/docs/evidence/codex-mcp-auth/after-login-gate.txt
@@ -1,0 +1,6 @@
+[worktree assert OK] manager=/private/tmp/lop-codex-auth/local_operator/mcp/manager.py
+  gitlab (codex import: url only, no auth block)   /mcp login -> ACCEPTED -> runs the grant
+  launchdarkly (codex import)                      /mcp login -> ACCEPTED -> runs the grant
+  openaiDeveloperDocs (public, no auth needed)     /mcp login -> ACCEPTED -> runs the grant
+  lop-native oauth (explicit auth block)           /mcp login -> ACCEPTED -> runs the grant
+  stdio server (no transport for a token)          /mcp login -> refused: 'does not use OAuth login.'
--- a/docs/evidence/codex-mcp-auth/after-messages.txt
+++ b/docs/evidence/codex-mcp-auth/after-messages.txt
@@ -1,0 +1,8 @@
+===== AFTER — cold boot (first connect, nothing observed) =====
+  ✓ MCP openaiDeveloperDocs: connected
+  × MCP datadog failed: run /mcp login datadog — not authorized (401)
+  × MCP gitlab failed: run /mcp reauth gitlab — authorization expired
+  × MCP launchdarkly failed: run /mcp login launchdarkly — not authorized (401)
+  × MCP minerva-qa failed: run /mcp login minerva-qa — not authorized (401)
+
+===== AFTER — warm (reload/reconnect: challenge already observed) =====
--- a/docs/evidence/codex-mcp-auth/after-no-discovery.txt
+++ b/docs/evidence/codex-mcp-auth/after-no-discovery.txt
@@ -1,0 +1,3 @@
+[worktree assert OK] manager=/private/tmp/lop-codex-auth/local_operator/mcp/manager.py
+  raised: McpAuthChallengeError
+  user sees: × MCP apikey-only failed: apikey-only rejected our credentials (401) — no OAuth endpoint advertised; check its API key or headers
--- a/docs/evidence/codex-mcp-auth/before-connect.txt
+++ b/docs/evidence/codex-mcp-auth/before-connect.txt
@@ -1,0 +1,7 @@
+[worktree assert OK] manager=/private/tmp/lop-codex-auth/local_operator/mcp/manager.py
+gitlab                 FAILED  MCPError: Server returned an error response
+launchdarkly           FAILED  MCPError: unauthorized access
+minerva-qa             FAILED  MCPError: Server returned an error response
+datadog                FAILED  MCPError: Server returned an error response
+openaiDeveloperDocs    CONNECTED  tools=5
+unreachable-host       FAILED  CancelledError: Cancelled via cancel scope 1085d9950
--- a/docs/evidence/codex-mcp-auth/frames/after.svg
+++ b/docs/evidence/codex-mcp-auth/frames/after.svg
@@ -1,0 +1,182 @@
+<svg class="rich-terminal" viewBox="0 0 1238 782.0" xmlns="http://www.w3.org/2000/svg">
+    <!-- Generated with Rich https://www.textualize.io -->
+    <style>
+
+    @font-face {
+        font-family: "Fira Code";
+        src: local("FiraCode-Regular"),
+                url("https://cdnjs.cloudflare.com/ajax/libs/firacode/6.2.0/woff2/FiraCode-Regular.woff2") format("woff2"),
+                url("https://cdnjs.cloudflare.com/ajax/libs/firacode/6.2.0/woff/FiraCode-Regular.woff") format("woff");
+        font-style: normal;
+        font-weight: 400;
+    }
+    @font-face {
+        font-family: "Fira Code";
+        src: local("FiraCode-Bold"),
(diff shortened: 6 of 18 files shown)


damianvtran:
### Design review — round 1
Scope: 73a1521383c6be7136d5a9e6fb80884ec3167839..eaa4f16166ff6a2864502bd02dc87ceb448c8ff6

Findings: none.

I viewed the committed `before.svg` and `after.svg` as rendered PNG frames, then independently rendered the real `OperatorApp` at 60×30 with all four after-state notices. The 100×30 frame preserves the existing hierarchy, spacing, contrast, and centred boot composition; geometry remains `virtual_size == size == (98, 28)` with no scrollbar. At 60×30 the geometry remains `virtual_size == size == (58, 28)` with no scrollbar, and every actionable command survives intact despite wrapping: `/mcp login datadog`, `/mcp reauth gitlab`, `/mcp login launchdarkly`, and `/mcp login minerva-qa` are all fully readable before the explanatory tail.

The copy cleanly distinguishes a never-authorized server (“login … to authorize”) from a rejected stored grant (“reauth … authorization expired”) without protocol jargon. The challenge-less-but-discoverable 401 evidence correctly offers login only after discovery confirms an authorization server; the challenge with no discoverable OAuth instead says to set the API key or headers and does not promise OAuth. The rendered and probe evidence also shows the no-auth server connecting with five tools, no provider, no discovery, and no user-facing auth noise.

Verdict: **TERMINAL — no BLOCKER or MAJOR findings.**

damianvtran:
### Agent review — round 1
Scope: 73a1521383c6be7136d5a9e6fb80884ec3167839..eaa4f16166ff6a2864502bd02dc87ceb448c8ff6

#### Findings

- **F1 — MAJOR — `local_operator/mcp/manager.py:787`: a challenge after an HTTP redirect is not observed.** The hook requires exact string equality with the configured URL (apart from the query). A normal canonicalization such as configured `/mcp` → `307 /mcp/` makes the final 401's request URL differ, so the user still receives the SDK's opaque `MCPError: Server returned an error response`, and the challenge ledger remains empty. I reproduced this through the real `McpManager._connect_server`/SDK transport with a local HTTP server: direct `/auth` and challenge-less `/nochallenge` became `McpAuthChallengeError`, while `/redir` → `/redir/` 401 stayed generic. Compare the endpoint semantically (at minimum normalize an equivalent trailing slash), or otherwise carry the redirect chain so the final MCP endpoint challenge is attributed to this attempt.

- **F2 — MAJOR — `local_operator/mcp/manager.py:784-789`: the watcher remembers any earlier 401/403 rather than the response that caused the final failure.** `status_code` is set on a challenge and never cleared by a later endpoint response. Therefore an attempt that receives a 401 and then retries the same MCP endpoint into a 500/404/network failure is relabelled as authorization and may tell the user to run `/mcp login`, violating the non-auth-failure requirement. The state machine reproduces as `401` followed by `500` leaving `status_code == 401`. Tie the observation to the terminal request/error (or clear/supersede it when a later MCP-endpoint response is not 401/403), and add an end-to-end regression for 401→500.

- **F3 — MAJOR — `local_operator/mcp/auth.py:745-747` and `local_operator/tui/app.py:15622`: explicit `/mcp login` is enabled for every remote config, including servers explicitly configured for API-key auth and known-public servers.** I verified `server_is_oauth_capable(..., deliberate=True)` returns true for both `auth.type="apikey"` and URL-only public configs; the PR evidence likewise says `openaiDeveloperDocs` is accepted. This is not “available only when justified,” and an API-key server's 401 can now enter an unrelated OAuth attempt. Preserve explicit non-OAuth auth as a hard refusal, and require explicit OAuth, a stored grant, or an observed discoverable OAuth challenge before enabling the command.

- **F4 — MAJOR — `local_operator/mcp/manager.py:1744-1748`: the already-computed stored-grant fact is discarded when choosing `reauth` versus `login`.** `_challenge_error` reads the manager's effective/injected store into `McpAuthChallengeError.has_stored_grant`, but `_auth_required_text` ignores that field and opens the default machine store again. With a supported injected store (and also if the second lookup fails), a challenge carrying `has_stored_grant=True` is rendered as `/mcp login`; I reproduced that directly. Use `exc.has_stored_grant` for `McpAuthChallengeError` so the message reflects the same store and instant that classified the failure; retain the lookup only for legacy `McpAuthRequiredError`.

#### Independent verification

- Whole-tree gates passed: flake8, Black 26.1.0, isort 5.13.2, and pyright (0 errors).
- Whole unit suite passed: **9050 passed, 15 skipped**.
- Focused MCP/TUI suite passed: **672 passed** (`test_auth.py`, `test_manager.py`, `test_app_pilot.py`, `test_command_picker.py`, `-n0`, required terminal environment).
- Synthetic real SDK/HTTP matrix: direct OAuth 401, challenge-less 401, and 403 became auth challenges; unauthenticated 200 connected; 404 and 500 remained non-auth; startup opened **0 browsers**. The redirect and stale-observation cases above failed as described.
- No credential-shaped values were found in committed evidence. The patch bump `0.43.16` → `0.43.17` is correctly a patch against current main/tag; PyPI currently reports `0.43.15`. Other open PRs also touch the version and will require the usual pre-merge rebase/version check.
- CI is still pending at review time.

**Verdict: changes requested — 0 BLOCKER, 4 MAJOR. This round is non-terminal; a fresh review is required after remediation.**

damianvtran:

Scope: `73a1521383c6be7136d5a9e6fb80884ec3167839..e0d2f57207dcac1c7358db1edb63803258393ec1`
New head: `e0d2f57207dcac1c7358db1edb63803258393ec1`

All four majors **fixed** in `e0d2f57207dcac1c7358db1edb63803258393ec1`. Each
carries a regression test that I verified **fails against the pre-fix code**
(reverted the hunk, ran the test, confirmed the reviewer's exact symptom, then
restored) — evidence per finding below.

---

**F1 — MAJOR — a challenge after an HTTP redirect is not observed. — FIXED** (`e0d2f57`)

Confirmed. The hook required string equality with the configured URL, so a
`307 /mcp` → `/mcp/` canonicalization made the final 401's request URL differ
and the challenge was dropped.

Endpoint identity is now semantic (`_AuthChallengeWatcher._endpoint_key`):
scheme and host lowercased, query dropped, trailing slash stripped. Deliberately
not looser — anything coarser would start attributing a metadata probe's 401 to
the connect.

Reproduced end to end through the **real SDK transport** against a local server
that redirects then challenges:

```
BEFORE  F1-redirect-401   MCPError               -> "Server returned an error response"
AFTER   F1-redirect-401   McpAuthChallengeError  -> names the fix

hook saw 307 http://127.0.0.1:8902/redir
hook saw 401 http://127.0.0.1:8902/redir/
```

Regression tests: `test_a_challenge_after_a_canonicalising_redirect_is_observed`,
`test_case_and_query_differences_still_match_the_endpoint`. Both fail pre-fix
with `assert None == 401`.

---

**F2 — MAJOR — the watcher remembers any earlier 401/403 rather than the terminal response. — FIXED** (`e0d2f57`)

Confirmed, and this was the more serious of the two: it is precisely the
"non-auth failure must not be relabelled" guarantee the PR claims.

The **last** response on the MCP endpoint now wins — a later non-401/403
supersedes an earlier challenge by setting `status_code = None`. A `3xx` is
explicitly *not* a verdict (the client is about to re-request the canonical
URL), so a redirect hop cannot clear a challenge that has not happened yet.

```
after 401 -> status_code=401
after 500 -> status_code=None
classified as: None      (stays a network/server error — no /mcp login)
```

Regression tests: `test_a_later_non_auth_response_supersedes_an_earlier_challenge`
(fails pre-fix with `assert 401 is None`), plus
`test_a_redirect_hop_never_clears_a_pending_verdict` and
`test_a_metadata_probe_neither_sets_nor_clears` guarding the two ways an
over-eager clear would regress F1 or the discovery exclusion.

---

**F3 — MAJOR — explicit `/mcp login` enabled for every remote config. — FIXED** (`e0d2f57`)

Confirmed, and accepted as stated: `deliberate=True` returning true for an
`auth.type="apikey"` config was wrong, not merely permissive.

The shortcut is **removed entirely** — `server_is_oauth_capable` no longer has a
`deliberate` parameter. Replaced by two pieces:

- `server_rejects_oauth(cfg)` — the statically impossible cases, refused with
  **no network call**: stdio (no transport for a bearer token) and any explicit
  non-OAuth `auth.type`. An `apikey` entry is the user stating how the server
  authenticates, so its 401 is not an invitation to an OAuth flow.
- `probe_oauth_capability(cfg)` — for the genuinely unknown case, one RFC
  9728/8414 discovery **before any grant starts**. This is why I did not simply
  make the gate strict: the strict test alone would refuse the first login on a
  fresh Codex import in a new process (the ledger is per-process), which is the
  dead end this PR exists to remove. A server advertising no authorization
  server is now refused with the same message a stdio server gets.

Both grant entry points are gated: `_mcp_login_worker` and
`_run_mcp_grant_on_owner` (a follower would otherwise reach a grant the local
terminal refuses). `_build_oauth_auth` no longer widens on `interactive` — by
then the probe has already recorded its answer in the ledger the gate reads.

Regression tests: `TestProbeOauthCapability` (5 tests — public refused, API-key
refused *without probing*, discoverable accepted and remembered, static evidence
short-circuits, probe failure degrades to refusal) and
`test_an_explicit_non_oauth_auth_type_is_a_hard_refusal`. Four fail pre-fix with
`assert True is False`.

---

**F4 — MAJOR — the already-computed stored-grant fact is discarded. — FIXED** (`e0d2f57`)

Confirmed and reproduced exactly as described.

`_auth_required_text` now uses `exc.has_stored_grant` when the exception carries
it — the value `_challenge_error` resolved against the manager's own (possibly
injected) store at the instant it classified the failure. Only the legacy
`McpAuthRequiredError`, which has no such field, falls back to a store lookup.

```
BEFORE  has_stored_grant=True  ->  "run /mcp login gitlab to authorize"    (wrong)
AFTER   has_stored_grant=True  ->  "run /mcp reauth gitlab — authorization expired"
```

Regression tests: `test_the_stored_grant_fact_comes_from_the_classifying_store`
(asserts the fallback lookup is **not called**; fails pre-fix with the exact
`/mcp login` string the reviewer saw) and
`test_the_challenge_error_carries_the_managers_store_verdict` covering the
`_challenge_error` → message path with an injected store.

---

#### Gates (whole tree, exactly as CI runs them)

```
flake8 .                                    clean
black==26.1.0 --check .                     635 files unchanged
isort==5.13.2 --check .                     clean
pyright --pythonpath .venv/bin/python .     0 errors
pytest tests/unit                           9061 passed, 15 skipped
```

30 focused MCP tests green (capability gate, probe, watcher attribution,
messaging). Suite total is up from 9028 to 9061 (+11 new regressions, +22 from
the rebase onto `origin/main` @ `73a1521`).

One unrelated order-dependent flake surfaced in the full run:
`tests/unit/tui/test_subagent_view.py::test_history_unavailable_and_error_retry_keep_trajectory_fallback`.
It touches no code in this PR, passes in isolation on **both** this branch and a
clean `origin/main` worktree (104 passed each), and is the same class as the
previously-noted `test_resume_subagents_todos` flake. Not from this change.

#### Behavioural parity preserved

The live matrix is **byte-identical** to the pre-remediation run — same verbs,
same wording:

```
✓ MCP openaiDeveloperDocs: connected
× MCP datadog failed: run /mcp login datadog to authorize
× MCP gitlab failed: run /mcp reauth gitlab — authorization expired
× MCP launchdarkly failed: run /mcp login launchdarkly to authorize
× MCP minerva-qa failed: run /mcp login minerva-qa to authorize
```

Re-verified through the real SDK transport: direct 401 and 403 → auth challenge;
challenge-less 401 → auth challenge without promising OAuth; public 200 →
connects with 5 tools; 404 and 500 → unchanged non-auth errors; **zero browsers
opened at startup**.

#### Frames

**Not re-rendered, deliberately.** All four fixes change *which* failures are
classified as auth and *which* store answers the verb — no copy string and no
layout changed, and the live matrix above reproduces the same four lines the
design round reviewed. The terminal design round therefore remains current
against this head, and I have not modified it.

#### Version

Re-checked after remediation: tags reach `v0.43.16`, `origin/main` is
`0.43.16`, PyPI reports `0.43.15`. `0.43.17` is still unclaimed, so it stands —
no second bump.

Evidence updated in `docs/evidence/codex-mcp-auth/README.md`; `docs/mcp.md`
documents the hard refusals and the pre-grant probe.

damianvtran:
### Agent review — round 2
Scope: eaa4f161...e0d2f572...

#### Round 1 dispositions

- **F1 — FIXED.** The per-attempt response hook now observes the final `/mcp/` response after `/mcp` canonicalizes through a redirect. The focused regression passes at the current head and the same behavior fails at `eaa4f161` (`status_code=None`). The real-SDK evidence records the 307 then final 401 and an actionable `McpAuthChallengeError`; no startup browser is involved.
- **F2 — NOT FIXED.** A later 500 response clears the prior challenge, but a later request that fails before receiving any response leaves the old 401/403 latched. Finding F5 below continues this major.
- **F3 — FIXED for public/API-key classification and fresh implicit discovery.** Explicit non-OAuth auth is a hard refusal with no probe; a URL-only public server is refused after discovery finds no authorization server; a fresh URL-only OAuth server is admitted only after successful non-interactive discovery, which records capability for the subsequent interactive connect. Background startup remains non-interactive and does not probe/open a browser. The current tests cover these outcomes and explicit OAuth remains accepted.
- **F4 — FIXED for failure wording.** `McpAuthChallengeError.has_stored_grant`, computed by the classifying manager against its effective store, now directly chooses `reauth`; the default-store fallback is retained only for the legacy exception. The regression verifies no second default-store lookup occurs. Finding F6 covers a separate effective-store omission introduced in the new login eligibility gate.

#### Findings

- **F5 — MAJOR — `local_operator/mcp/manager.py:166-191`: a response-only watcher still lets an earlier 401/403 relabel a later network failure.** The remediation clears the latch only when a later same-endpoint *response* arrives. A retry that starts after the 401 and then fails at DNS/connect/TLS/read time invokes no response hook, so `status_code` remains 401 and `_challenge_error` converts the terminal network failure into `/mcp login`. I reproduced the state as `401 -> network error (no response) => status_code == 401`. This is explicitly one of F2's required cases and means the watcher still does not identify the terminal request/error. Clear/supersede the verdict at the start of each same-endpoint request (or otherwise bind a challenge to the request whose terminal error is being classified), then add an end-to-end 401→network-failure regression rather than only a two-response state test.

- **F6 — MAJOR — `local_operator/tui/app.py:15662,15838`: both explicit-login gates discard the manager's effective/injected auth store.** `_run_mcp_grant_on_owner` and `_mcp_login_worker` call `probe_oauth_capability(cfg)` without the manager store. Therefore a server whose stale grant exists only in a supported custom/injected store can be refused as “does not use OAuth login” whenever discovery is unavailable or transiently fails, even though this same manager correctly classified the failure as `reauth` from that store. I reproduced `probe_oauth_capability(cfg, effective_store) == True` versus the exact no-store call used by both paths returning `False` with discovery offline. Pass the manager's effective store through both entry points (via a public manager accessor or manager-owned eligibility operation) and regress the complete custom-store `/mcp reauth` path.

#### Independent verification

- Whole-tree gates passed exactly as CI: flake8, Black 26.1.0 (635 files unchanged), isort 5.13.2, pyright (0 errors).
- Focused MCP/TUI suite passed: **684 passed** (`test_auth.py`, `test_manager.py`, `test_app_pilot.py`, `test_command_picker.py`, `-n0`, required terminal environment).
- Whole unit suite: **9061 passed, 15 skipped, 1 failed**. The sole failure was the pre-existing order-dependent `test_history_unavailable_and_error_retry_keep_trajectory_fallback`; it passed immediately in isolation and touches no PR code.
- Direct/challenge-less 401, OAuth challenge, 403, public 200, 404/500 parity, explicit OAuth, per-attempt watcher isolation, and no-startup-browser behavior remain covered by the focused suite/evidence. The unresolved 401→network case is F5.
- The new committed evidence contains no credential values. The unchanged design artifacts remain accurate: remediation does not alter the four reviewed auth notice strings or geometry. The terminal design round remains current.
- GitHub CI for head `e0d2f572` is still pending (10 checks across the push and PR workflow runs); PR state is currently blocked pending checks/review.

**Verdict: changes requested — 0 BLOCKER, 2 MAJOR. NON-TERMINAL; mergeable: no.** A fresh review is required after F5-F6 remediation.

damianvtran:

Scope: `e0d2f57207dcac1c7358db1edb63803258393ec1..c08c0a3c08d7a0b1e84d18608af3032e69e9a29f`
New head: `c08c0a3c08d7a0b1e84d18608af3032e69e9a29f`

Both majors **fixed** in `c08c0a3c08d7a0b1e84d18608af3032e69e9a29f`. Each carries
a regression test I verified **fails against the pre-fix code** (reverted the
hunk, ran the test, reproduced the reviewer's exact symptom, restored).

Thank you for catching F5 — the round 1 fix was genuinely incomplete, not just
narrowly scoped. Clearing on the response was the wrong anchor.

---

**F5 — MAJOR — a response-only watcher still lets an earlier 401/403 relabel a later network failure. — FIXED** (`c08c0a3`)

Confirmed exactly as described. Clearing on a later *response* cannot help when
the next request produces none: a retry that fails at DNS/connect/TLS/read time
invokes no response hook, so `status_code` stayed 401 and `_challenge_error`
converted a terminal network failure into `/mcp login`.

The verdict is now bound to the **request**, which is what "identify the
terminal request/error" actually requires:

- `_AuthChallengeWatcher.begin`, installed as the httpx `request` hook, clears
  the slot as each same-endpoint request *starts*;
- `observe` sets it only if that request comes back 401/403.

So whatever happens next — a response, or a socket error that never produces
one — no stale verdict survives. A `3xx` is still not a verdict (the follow-up
request re-clears the slot anyway), and requests to other URLs (the provider's
discovery and token traffic) neither set nor clear.

End-to-end regression with a **real httpx client against a real server**, as
requested rather than a two-response state test:

```
after the 401           : watcher verdict = 401
retry failed with       : ReadTimeout (no response)
after the dead retry    : watcher verdict = None
classified as           : None      (NOT auth advice)
```

Regression tests: `test_a_request_that_never_answers_leaves_no_verdict` and
`test_a_network_failure_after_a_challenge_is_not_auth_advice` (the end-to-end
classifier assertion). Both fail pre-fix — the second with the reviewer's exact
symptom, `assert McpAuthChallengeError(...401) is None`. Guarded against
over-clearing by `test_a_request_to_another_url_never_clears_the_verdict` and
`test_a_challenge_still_survives_to_classification`.

---

**F6 — MAJOR — both explicit-login gates discard the manager's effective auth store. — FIXED** (`c08c0a3`)

Confirmed. `_run_mcp_grant_on_owner` and `_mcp_login_worker` both called
`probe_oauth_capability(cfg)` with no store, so a server whose stale grant lives
only in an injected store was refused as "does not use OAuth login" whenever
discovery was unavailable — while the *same manager* had just classified its
failure as `reauth` from that store. The two gates disagreed about the same
server.

Fixed via the manager-owned eligibility operation you suggested:
`McpManager.server_supports_oauth_login(cfg)`, which threads its own
`_effective_auth_store()` through the probe. Both entry points now go through
one TUI helper (`_mcp_login_allowed`) rather than reaching for the auth module
directly, so no front end needs to know the store exists.

A reduced follower facade that does not implement the accessor falls back to
the store-less probe. That is correct rather than lenient: such a facade owns no
auth store to consult, and the static refusals (stdio, explicit non-OAuth
`auth.type`) still apply.

With discovery offline and a grant in a custom store:

```
probe_oauth_capability(cfg)                  -> False   (the call both gates made)
manager.server_supports_oauth_login(cfg)     -> True    (consults its own store)
```

Regression test: `test_the_eligibility_gate_uses_the_managers_effective_store`,
which asserts both halves of that comparison and fails pre-fix with
`assert False is True`.

---

#### Gates (whole tree, exactly as CI runs them)

```
flake8 .                                    clean
black==26.1.0 --check .                     635 files unchanged
isort==5.13.2 --check .                     clean
pyright --pythonpath .venv/bin/python .     0 errors
pytest tests/unit                           9067 passed, 15 skipped, 0 failed
```

Focused MCP/TUI suite: **689 passed** (`test_auth.py`, `test_manager.py`,
`test_app_pilot.py`, `test_command_picker.py`, `-n0`, required terminal
environment).

The previously-noted order-dependent flakes did not recur in this run; the full
suite is clean.

#### Behavioural parity preserved

Live matrix byte-identical to round 1 — same verbs, same wording:

```
✓ MCP openaiDeveloperDocs: connected
× MCP datadog failed: run /mcp login datadog to authorize
× MCP gitlab failed: run /mcp reauth gitlab — authorization expired
× MCP launchdarkly failed: run /mcp login launchdarkly to authorize
× MCP minerva-qa failed: run /mcp login minerva-qa to authorize
```

Real-transport matrix also unchanged: redirect-then-401 → auth challenge; direct
401 and 403 → auth challenge; challenge-less 401 → auth challenge without
promising OAuth; public 200 → connects with 5 tools; 404 and 500 → unchanged
non-auth errors; **zero browsers opened at startup**.

#### Design round

**Stays terminal, and I have not touched it.** Neither fix changes a copy string
or any geometry — F5 changes *which* failures are classified as auth, F6 changes
*which store* answers the eligibility question. The four notice strings the
design round reviewed are reproduced verbatim by the live matrix above, so the
existing frames remain current against this head. No re-render.

#### Version

Re-checked after remediation: tags reach `v0.43.16`, `origin/main` is `0.43.16`,
PyPI now reports `0.43.16`. `0.43.17` remains unclaimed, so it stands — no
second bump.

Evidence updated in `docs/evidence/codex-mcp-auth/README.md`.

damianvtran:
### Agent review — round 3
Scope: e0d2f57207dcac1c7358db1edb63803258393ec1...c08c0a3c08d7a0b1e84d18608af3032e69e9a29f

#### Round 2 dispositions

- **F5 — RESOLVED.** `_AuthChallengeWatcher.begin` is installed as the request hook and clears the verdict at every same-endpoint request start; a request that then dies without a response leaves no stale 401/403 for `_challenge_error` to classify. `observe` records only the current same-endpoint terminal response, redirects remain non-verdict hops, and discovery/token URLs neither set nor clear the endpoint verdict. The watcher is allocated per `_connect_server` attempt, so parallel managers/servers do not share state. I independently reproduced 401 → next request begins → no response as `status_code=None`; the focused regressions cover the classifier result and redirect/current-response behavior.
- **F6 — RESOLVED.** Both explicit grant entry points now route through `_mcp_login_allowed`, which delegates to `McpManager.server_supports_oauth_login`; that manager-owned operation passes `_effective_auth_store()` into `probe_oauth_capability`. With discovery unavailable, the injected-store regression passes at this head and fails when I remove that store argument (`assert False is True`), so the test is sensitive to the required threading rather than merely asserting a mocked call.

#### Findings

No new F7+ findings.

#### Independent verification

- Rechecked F1–F4: semantic redirect attribution, terminal non-auth response supersession, public/API-key refusal with capability discovery, and classifying-store `login`/`reauth` wording remain intact in code and focused coverage.
- Focused MCP/TUI suite passed independently: **689 passed** (`test_auth.py`, `test_manager.py`, `test_app_pilot.py`, `test_command_picker.py`, `-n0`, required terminal environment).
- Whole-tree gates passed exactly as CI: flake8, Black 26.1.0 (**635 files unchanged**), isort 5.13.2, and pyright (**0 errors**).
- The remediation and committed matrix preserve direct/challenge-less 401, 403, redirect→401, public 200, and unchanged 404/500 behavior; no startup browser path was added.
- No user-facing copy or geometry changed in `e0d2f572...c08c0a3c`; the four reviewed notices are unchanged, so terminal design round 1 remains current.
- Version status: branch `0.43.17`; current `origin/main`, latest tag/release, and PyPI are `0.43.16`, and no `v0.43.17` tag/release exists. However, open PRs #464, #455, and #447 also currently claim `0.43.17`; the eventual merger must recheck/rebase if another lands first.
- GitHub reports the branch **MERGEABLE**, but required CI is not settled: lint/type-check/pip-audit are green in both current-head runs, while four Python 3.12/3.13 test jobs remain in progress. GitHub merge state is therefore **BLOCKED**.

**Verdict: clean — 0 BLOCKER, 0 MAJOR. TERMINAL: no further agent-review round is required for this head. Mergeable: no, pending current-head CI completion (and a final version recheck because other open PRs claim 0.43.17).**

damianvtran:
### Design review — round 2
Scope: eaa4f16166ff6a2864502bd02dc87ceb448c8ff6...c08c0a3c08d7a0b1e84d18608af3032e69e9a29f

Findings: none.

I independently inspected the interdiff and viewed the current committed `before.svg` and `after.svg` as rendered frames. The frame blobs, `after-messages.txt`, and `after-login-gate.txt` are byte-identical at both scope endpoints. The TUI interdiff adds eligibility routing and preserves the existing refusal literal; its only new string literals are an internal type name, an accessor name, and documentation. It changes no painted copy, CSS, spacing, or geometry. The round-1 frames therefore remain exact and current for this head, not merely representative, and no re-render is required.

The current error constructors still distinguish stored-grant reauthorization (`run /mcp reauth <name> — authorization expired`) from first authorization (`run /mcp login <name> to authorize`). At narrow width, the already-reviewed 60×30 rendering remains current because those exact strings and all layout inputs are unchanged; all four actionable commands remain readable. A challenge-less 401 with discoverable OAuth still uses the honest login wording, while a 401 with no discoverable OAuth says `<name> rejected our credentials (401) — set its API key or headers` and does not promise OAuth. The committed public/no-auth evidence still shows a five-tool connection with no provider, no discovery, and no user-facing auth notice; the interdiff introduces no startup notice or geometry path.

Verdict: **TERMINAL — no BLOCKER or MAJOR findings.** Clean/current-head: **yes** — clean at `c08c0a3c08d7a0b1e84d18608af3032e69e9a29f`.

PR #456, opened 2026-08-30
fix(session): recover from an unknown hosting provider instead of dying

damianvtran:
# PR Description

## Summary of Changes

A corrupted or typo'd `hosting:` in `config.yml` killed every new session
permanently, and left the user with no way out from inside the app:

```
× session failed to start: Unsupported hosting platform: anthropicxyq
```

The status band read `session error`, and trying to switch to a working
provider from inside the app answered:

```
! session is still starting…
```

So the session was unusable AND the provider could not be changed — the only
exit was hand-editing YAML outside the app. Hit in the wild from a
`hosting: anthropicxyq` value beside a perfectly valid `model_name`.

### Cause

Two layers, and the second is what turned a bad value into a lockout.

`session_factory.resolve_hosting_model` validated only that hosting was
**non-empty**:

```python
if not hosting:
    raise HostingNotConfiguredError("Hosting platform is not configured.")
```

It never checked that a non-empty hosting is a provider the registry actually
owns, so garbage passed preflight and detonated deeper in boot as a bare
`ValueError` from `configure_model` (`Unsupported hosting platform: …`).

The TUI's `_on_boot_failed` special-cases **only** `HostingNotConfiguredError`
→ `_enter_setup_state()`. A bare `ValueError` therefore took the red
"session failed to start" branch and `_session` stayed `None` — which is why
roughly twenty guarded commands, `/model` and `/provider` among them, replied
"session is still starting…".

**The asymmetry was the bug.** "Nothing configured" was treated as a guided
first-run state; "configured to an unknown provider" was treated as a fatal
crash — even though the user's remedy is identical in both cases (`/login`,
`/provider`, `/model` from inside the app).

### Fix

- **`HostingUnknownError`, a SUBCLASS of `HostingNotConfiguredError`.** The
  existing recoverable-setup handling picks it up with no change and cannot be
  updated for one condition while forgetting the other, while the two stay
  distinguishable by type so the messages can differ. `ValueError` ancestry is
  preserved for legacy `except ValueError` callers.
- **Validate at preflight**, in `resolve_hosting_model` where the
  not-configured case is already caught, through `get_provider_definition` —
  not a membership test against ids — so legacy aliases (`noop` → `test`) keep
  working and the preflight accepts exactly what the engine accepts.
- **TUI** routes it to the existing `_enter_setup_state()` (no parallel
  wizard), with an action-first message naming the bad value and the remedy.
- **The recovery loop is closed.** `_apply_login_defaults` (both the TUI and
[...]

PR #465, opened 2026-08-30
ci: stop the queue eating itself on a 20-job concurrency cap

damianvtran:
# PR Description

## Summary of Changes

The Actions queue reached **69 active runs against ~11 executing**, delaying every branch and stalling the `0.43.16` `pypi-publish` job behind unrelated feature-branch work.

The repo is **public**, so standard-runner minutes are free and unlimited — this was never a spend problem. The binding constraint is the account-wide **20 concurrent jobs** on the Free plan ([Actions limits](https://docs.github.com/en/actions/reference/limits)), and **most of what was consuming it was work whose result nobody would ever read**.

Four causes, each measured on the live queue rather than assumed:

| # | Cause | Measured impact |
|---|---|---|
| 1 | No `concurrency` group — superseded runs never cancelled | **40 of 71 active runs (56%)** on already-superseded commits |
| 2 | `push: [dev-*]` **and** `pull_request` both fire | **29 SHAs** running two identical job sets |
| 3 | No `timeout-minutes` on 4 jobs → 360 min default | one hung job held a slot **3h38m** |
| 4 | `test` is ~23 min and ran **twice** per commit | largest single slot consumer |

Touches three workflows: `ci.yml` (all four fixes), `extension.yml` (same double-trigger, concurrency group, timeout), `publish.yml` (timeouts only).

### 1. Concurrency group

```yaml
group: ${{ github.workflow }}-${{ github.ref }}-${{ github.ref == 'refs/heads/main' && github.sha || 'shared' }}
cancel-in-progress: true
```

Feature branches share one group per ref, so successive pushes to a PR supersede each other — that is where the throughput saving comes from. **`main` gets the commit SHA in its key**, so every merge commit lands in its own group and no `main` run is cancelled or evicted by a later merge.

**Why the SHA rather than `cancel-in-progress: false`.** A concurrency group allows one run in progress and, by default, only **one pending**; when a third arrives the waiting one is cancelled. With `main` merging faster than this workflow's ~23 min duration (four merges in 27 minutes was observed while writing this), run A executes, B waits, C arrives and evicts B — leaving B a commit on `main` that no CI ever reported on. `cancel-in-progress: false` protects only the *executing* run and would not have prevented that.

**Why not `queue: max`,** which is the documented way to hold pending runs: GitHub rejects it alongside a `cancel-in-progress` that can evaluate true, and cancellation on feature branches is the entire point. Attempted, and the run failed at plan time in 1 s with zero jobs:

```
evaluate plan concurrency: Error when evaluating 'concurrency'.
.github/workflows/ci.yml (Line: 65, Col: 10):
The combination of 'queue: max' and 'cancel-in-progress: true' is not allowed
```

The per-SHA key reaches a stronger guarantee than `queue: max` would have — `main` runs don't queue behind each other, they simply run — using keys supported all along.

**Known, accepted edge:** two runs on the *same* `main` commit (a re-run, or a `workflow_dispatch` at a commit already building) do share a group, and the later supersedes the earlier. The survivor is on the identical SHA and runs the same full both-leg matrix, so the record is replaced by an equivalent one rather than lost — which is what a GitHub re-run does by design anyway.

### 2. Double-trigger removed

`dev-*` leaves the `push` list in both `ci.yml` and `extension.yml`; `pull_request` already covers feature branches. A `dev-*` branch with no open PR now gets no CI until the PR exists — intended, since CI gates review and merge, and `workflow_dispatch` remains for a pre-PR signal.

### 3. Timeouts

Every job in all three workflows is now bounded: `ci.yml` `lint` 10 / `type-check` 15 / `test` 40 / `pip-audit` 10; `extension.yml` `build` 15; `publish.yml` `extension-release` 15 / `release-build` 20 / `pypi-publish` 20. These are hang backstops well above honest durations, not limits healthy work can reach. `publish.yml` matters most — a squatter there blocks *publishing*, which is exactly what happened to `0.43.16`.

### 4. Matrix scoped on PRs

PRs run the **3.12 leg only**; `main` still runs **both ends** of `requires-python`, so no commit reaches a release without the full declared range having run on it.

**The trade, measured rather than asserted.** Across 41 recent runs where both legs actually reported: **3** were 3.13-only red, **7** were 3.12-only red, 1 both. The 3.13 leg is not carrying a unique load. What decided it is *why* those 3 went red — each was `1 failed` of ~9000 on a different timing-sensitive test, never the same one twice, none a genuine 3.13 incompatibility (one confirmed from the log as a composer-height timing assertion at `test_transcript_selection.py:2243`). They are flakes on a loaded runner. So this takes a source of flaky reds out of the review loop while keeping the real incompatibility gate on every merge. Fix the flakes separately; if a true 3.13-only break ever reaches `main`, put `"3.13"` back in the `pull_request` arm rather than buying concurrency.

Pip caching is also added and **honestly labelled as marginal**: install is ~15-20 s of a ~23 min job, so it is *not* what fixes the queue. With no lockfile and 0 of 19 dependencies pinned, the key only changes when `pyproject.toml` does, so the cache decays rather than staying current. It cannot poison a run — it is pip's HTTP download cache, and pip still resolves against the index.

### What this deliberately does NOT do

- **No upgrade to Pro.** The observed ceiling was ~11 concurrent jobs against a nominal cap of 20 — the account was not saturating its existing allowance, so doubling the nominal cap would not have doubled throughput. Demand-side fixes first; measure again before spending.
- **No change to `-n auto`.** The existing comment documents that oversubscription was measured and starves timing-sensitive tests.

### Trade-off worth naming

Because `main` runs no longer contend, a merge burst runs them fully in parallel, and a `main` run is 9 jobs — three overlapping runs is 27 against a 20-job cap. This is not a regression (the base had no concurrency group at all) and the throughput win is on the PR side where the volume is, but it is correctness bought at some parallelism, not pure upside.

---

### Live verification on the PR's own runs

Each structural fix was confirmed against the real Actions API, not by reading YAML:

**Double-trigger gone** — the first push produced exactly one run, where the old config would have produced two:

```
33331583538	pull_request	in_progress	432c4686      # no push run
```

**`cancel-in-progress` works** — pushed a second commit, re-listed 35 s later:

```
33331610417	pull_request	in_progress	          4f7fb181   <- new head
33331583538	pull_request	completed    cancelled  432c4686   <- auto-cancelled
```

Reproduced repeatedly across later pushes (`bc16010f`, `4de85de8` both auto-cancelled).

**PR matrix scoped** — jobs on a `pull_request` run:

```
type-check   completed  success
pip-audit    completed  success
lint         completed  success
test (3.12)  in_progress
```

One `test` leg, not two; `test (3.13)` correctly absent.

**The rejected config actually failed** — the `queue: max` head ran 1 s and dispatched **zero** jobs (`conclusion=failure`, empty jobs list), which is how the plan-time error above was found. The replacement head dispatches jobs normally.

### `actionlint`

Clean on all three workflows. The only finding is a **pre-existing** SC2086 info, confirmed by running against the base commit — same finding, line number shifts only because this PR adds comment lines.

Note `actionlint` 1.7.12 also rejects `queue: max` (`unexpected key "queue"`) because its schema predates the May 2026 feature — the live run, not the linter, is authoritative there.

### Every job bounded

Parsed from the resulting YAML:

```
ci.yml         lint 10   type-check 15   test 40   pip-audit 10
               cli-sanity 10   tui-e2e 20   server-sanity 10
extension.yml  build 15
publish.yml    extension-release 15   release-build 20   pypi-publish 20
```

Previously `lint`, `type-check`, `test`, `pip-audit` and all of `publish.yml` showed `NONE`.

### Group-key collision analysis

Simulated across events, since a key that collided two PRs would make them cancel each other:

```
push main c1       -> CI-refs/heads/main-aaa111
push main c2       -> CI-refs/heads/main-bbb222     distinct: no eviction
PR 465             -> CI-refs/pull/465/merge-shared
PR 465 new push    -> CI-refs/pull/465/merge-shared same: supersedes (intended)
PR 470             -> CI-refs/pull/470/merge-shared distinct from 465
```

`'shared'` is a constant for all non-`main` refs and contributes no discrimination — it does not need to, because `github.ref` is already its own segment and differs per PR and per branch.

### Backlog actually cleared

The hung run plus 36 superseded runs were cancelled by hand to unblock the release. **`0.43.16` `pypi-publish` then completed successfully**, and the queue went from **69 active → 31**, later to 17 queued / 10 running. That is the manual version of what change (1) now does automatically.

### Branch hygiene

Rebased onto current `origin/main` and squashed to one commit. Verified the diff is workflow-only — `3 files changed`, no unrelated code — since an earlier stale base made unrelated merged TUI work appear as deletions.

## Review

Two agent review rounds; see the round-1 and round-2 comments and their remediations. Round 1 found the pending-run eviction defect (R1-1) that the `github.sha` key now fixes, and corrected an unevidenced claim about 3.13 failure rates. Round 2 verified the replacement mechanism, independently confirmed the group key cannot collide two refs, and corrected the failure survey from 5 to 3 (my filter had mis-bucketed two `cancelled` legs as failures).

## Related

Root-caused while clearing the backlog holding the `0.43.16` release. Reported independently by a concurrent session, which reached the same conclusion about the concurrency group.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,8 +3,21 @@ name: Pre-release CI Testing
 on:
   push:
     branches:
+      # ONLY `main` here, deliberately.
+      #
+      # `dev-*` used to be in this list as well, which double-billed every
+      # feature branch: a `dev-*` branch with an open PR fired a `push` run AND
+      # a `pull_request` run for the SAME commit, two identical sets of jobs
+      # whose results nobody read twice. At one point 29 of the SHAs in the
+      # queue were running both. The `pull_request` trigger below already
+      # covers feature branches, so dropping `dev-*` here halves the per-commit
+      # cost without reducing what is actually tested.
+      #
+      # A `dev-*` branch with NO open PR now gets no CI until the PR is opened.
+      # That is the intended trade: CI exists to gate review and merge, and an
+      # un-PR'd branch is gating nothing. Use `workflow_dispatch` to run it by
+      # hand if you need a pre-PR signal.
       - main
-      - 'dev-*'
   # PULL REQUESTS TOO. Without this a feature branch received NO CI until it
   # was already merged, which let a broken `pip install .` and an import that
   # fails on the declared minimum Python both reach review looking green.
@@ -12,9 +25,67 @@ on:
   pull_request:
   workflow_dispatch:
 
+# One in-flight run per ref. A review cycle force-pushes repeatedly, and
+# without this every superseded push kept its full job set queued to completion
+# against a 20-job account-wide concurrency cap — 56% of the active queue was
+# once running against commits their own branch had already moved past, while
+# the newest commit (the only one anybody was waiting on) sat behind them.
+#
+# `github.workflow` is in the key so this workflow never cancels a sibling
+# workflow's run on the same ref (the release/publish line in particular).
+#
+# `github.ref` is the right granularity: for a `pull_request` event it is the
+# PR's merge ref, so successive pushes to one PR supersede each other while
+# two different PRs never collide.
+#
+# A push to `main` is never cancelled BY A LATER MERGE, and the SHA in the
+# group key is how that is achieved. A push to `main` is a merge whose result
+# is the release signal and the record for that commit, so every one of those
+# runs must complete.
+#
+# The precise guarantee: no `main` run is cancelled or evicted by a run on a
+# DIFFERENT commit. Two runs on the SAME `main` commit — a re-run, or a
+# `workflow_dispatch` at a commit already building — do share a group and the
+# later one supersedes the earlier. That is acceptable because the survivor is
+# on the identical SHA and, not being a `pull_request`, runs the same full
+# both-leg matrix: the CI record is replaced by an equivalent one rather than
+# lost, which is also what a GitHub re-run does by design.
+#
+# `cancel-in-progress: false` alone would NOT deliver that. A concurrency group
+# permits one run in progress and, by default, only ONE run pending: when a
+# third arrives the waiting one is cancelled. With `main` merging faster than
+# this workflow's ~23 min duration (four merges inside 27 minutes was observed
+# while writing this), run A executes, B waits, C arrives and evicts B — and B
+# is then a commit on `main` that no CI ever reported on.
+#
+# Adding `github.sha` for `main` gives every merge commit its OWN group, so
+# `main` runs never contend and none can be cancelled or evicted. Feature
+# branches keep a per-ref group, so successive pushes to one PR still supersede
+# each other, which is where the throughput saving comes from.
+#
+# Why not `queue: max` (the documented way to hold up to 100 pending runs):
+# GitHub rejects `queue: max` combined with a `cancel-in-progress` that can
+# evaluate true, and this workflow needs cancellation on feature branches.
+# Tried, and it failed the run at plan time with:
+#   "The combination of 'queue: max' and 'cancel-in-progress: true'
+#    is not allowed"
+# A unique group per `main` commit reaches the same guarantee with keys that
+# have been supported all along, and it does not queue `main` runs behind each
+# other — they simply run.
+#
+# This matters specifically because the `test` matrix below runs the 3.13 leg
+# ONLY on `main`. `main` is the safety net for the narrowed pull request
+# matrix, so a hole in that net would undo the justification for narrowing it.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}-${{ github.ref == 'refs/heads/main' && github.sha || 'shared' }}
+  cancel-in-progress: true
+
 jobs:
   lint:
     runs-on: ubuntu-latest
+    # Generous against the ~30 s this takes; it exists so a wedged install step
+    # cannot hold a concurrency slot for the 360-minute GitHub default.
+    timeout-minutes: 10
     steps:
       - uses: actions/checkout@v4
 
@@ -39,6 +110,7 @@ jobs:
 
   type-check:
     runs-on: ubuntu-latest
+    timeout-minutes: 15
     env:
       PYRIGHT_PYTHON_FORCE_VERSION: latest
     steps:
@@ -64,14 +136,67 @@ jobs:
         
   test:
     runs-on: ubuntu-latest
+    # ~23 min is the honest ceiling for a healthy run of this stage, so 40 is a
+    # backstop for a hang and not a limit healthy work can reach. It is
+    # load-bearing: a `test (3.13)` job once hung and held one of the account's
+    # 20 concurrency slots for 3h38m, because with no `timeout-minutes` the
+    # GitHub default is SIX HOURS. The suite's own failure modes assert or
+    # error in seconds; only a deadlock gets here.
+    timeout-minutes: 40
     strategy:
       fail-fast: false
       matrix:
-        # BOTH ends of `requires-python`. Testing only the minimum let a
-        # 3.13-only typing feature ship; testing only the newest would let a
-        # 3.12 incompatibility ship. The declared range has to be the tested
-        # range or the classifier is a guess.
-        python-version: ["3.12", "3.13"]
+        # BOTH ends of `requires-python` on `main`, the minimum only on pull
+        # requests.
+        #
+        # The rule this preserves is unchanged: testing only the minimum lets a
+        # 3.13-only typing feature ship, and testing only the newest lets a
+        # 3.12 incompatibility ship, so the declared range has to be the tested
+        # range or the classifier is a guess. `main` therefore still tests the
+        # full declared range, and no commit reaches a release without both
+        # ends having run on it.
+        #
+        # What changes is WHEN the 3.13 leg runs on feature work. This stage is
+        # ~23 min and was the single largest consumer of a 20-job account-wide
+        # concurrency cap; at two legs per run it alone could exceed the whole
+        # account's parallelism during a busy review cycle, and the queueing
+        # delayed every other branch AND the release pipeline. Running one leg
+        # per PR commit halves that.
+        #
+        # The cost, stated plainly: a 3.13-ONLY break is now caught when the PR
+        # merges to `main` rather than on the PR itself. That is a real
+        # reduction in per-PR gate strength.
+        #
+        # It was accepted on MEASURED grounds, not on the intuition that such
+        # breaks are rare. Surveying recent CI runs where BOTH legs actually
+        # reported a pass/fail (excluding cancelled legs), 3.13-only reds are a
+        # MINORITY of 3.12-only reds — measured repeatedly at roughly 1:2 to
+        # 1:7 depending on the window, and never once inverted. So the 3.13 leg
+        # is not carrying a unique load; if anything the 3.12 leg catches more.
+        #
+        # Deliberately a RATIO and not a fixed count: this repo merges fast
+        # enough that any absolute number is stale within hours, and two
+        # independent surveys of overlapping windows disagreed on the count
+        # while agreeing on the direction. The direction is the load-bearing
+        # part and it has held at every window measured.
+        #
+        # What decided it is WHY those reds happen: each was `1 failed` out of
+        # ~9000, on a DIFFERENT timing-sensitive test (transcript selection,
+        # resume-subagents, tui responsiveness), never the same test twice, and
+        # none was a genuine 3.13 language or typing incompatibility. One was
+        # confirmed by reading the log to be a composer-height timing assertion
+        # (test_transcript_selection.py:2243). They are flakes on a slow,
+        # loaded runner — the class of defect #461 ("wait on conditions instead
+        # of the clock") addresses.
+        #
+        # So relegating this leg to `main` takes a source of flaky reds out of
+        # the review loop while keeping the real incompatibility gate — which
+        # has not fired in this window — on every merge. If a true 3.13-only
+        # incompatibility ever DOES reach `main`, put "3.13" back into the
+        # pull_request arm rather than reaching for more concurrency; the flakes
+        # are the thing to fix first, and they are a separate concern from this
+        # workflow (see dev-test-determinism).
+        python-version: ${{ github.event_name == 'pull_request' && fromJSON('["3.12"]') || fromJSON('["3.12", "3.13"]') }}
     steps:
       - uses: actions/checkout@v4
 
@@ -79,6 +204,27 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
+          # Restores pip's HTTP cache. Measured saving is SMALL — install is
+          # ~15-20 s of a ~23 min job — so this is emphatically not what fixes
+          # the queue; it removes a PyPI round trip and one network flake mode
+          # per job.
+          #
+          # Known limitation, stated so nobody reads more into this than it
+          # delivers: the key is `pyproject.toml`, and this project pins none of
+          # its 19 runtime dependencies (`>=` throughout) and has no lockfile.
+          # The key therefore only changes when that FILE changes, while the
+          # resolved wheels drift underneath it as upstreams release. The cache
+          # decays into staleness rather than staying current.
+          #
+          # That is safe, not merely tolerable: this is pip's HTTP download
+          # cache, not a restored site-packages. pip still resolves against the
+          # index on every run and reuses a cached download only when the
+          # resolution asks for that exact artifact, so a stale entry cannot pin
+          # an old version into a build or otherwise poison a run. Adding a
+          # lockfile would make this caching genuinely effective, and is the
+          # right fix if install time ever becomes material here.
+          cache: pip
+          cache-dependency-path: pyproject.toml
 
       - name: Install dependencies
         run: |
@@ -116,6 +262,7 @@ jobs:
 
   pip-audit:
     runs-on: ubuntu-latest
+    timeout-minutes: 10
     steps:
       - uses: actions/checkout@v4
       - name: Install package
--- a/.github/workflows/extension.yml
+++ b/.github/workflows/extension.yml
@@ -3,8 +3,12 @@ name: Browser extension
 on:
   push:
     branches:
+      # `main` only, matching ci.yml: a `dev-*` branch with an open PR fired this
+      # workflow twice for the same commit (a push run and a pull_request run)
+      # for one identical result. Paths-gating kept the volume low enough that
+      # it never dominated the queue, but the duplication was the same defect
+      # and is fixed the same way.
       - main
-      - 'dev-*'
     paths:
       - 'extension/**'
       - 'local_operator/browser_bridge/protocol.py'
@@ -16,9 +20,17 @@ on:
       - 'local_operator/browser_bridge/gen_ts.py'
   workflow_dispatch:
 
+# Matches ci.yml, including the per-SHA group for `main` so a merge commit's
+# run can never be evicted by the next merge. See the long note in ci.yml for
+# why the SHA is in the key rather than `queue: max`.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}-${{ github.ref == 'refs/heads/main' && github.sha || 'shared' }}
+  cancel-in-progress: true
+
 jobs:
   build:
     runs-on: ubuntu-latest
+    timeout-minutes: 15
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -29,6 +29,10 @@ jobs:
   # ---------------------------------------------------------------------------
   extension-release:
     runs-on: ubuntu-latest
+    # Bounded so a wedged step cannot hold one of the account's 20 concurrency
+    # slots for GitHub's 360-minute default. This is the RELEASE path, so a
+    # squatting job here blocks publishing, not just a branch's CI.
+    timeout-minutes: 15
     permissions:
       contents: write  # upload the zip as a release asset
     steps:
@@ -118,6 +122,7 @@ jobs:
 
   release-build:
     runs-on: ubuntu-latest
+    timeout-minutes: 20
     outputs:
       skip_build: ${{ steps.tag-check.outputs.skip_build }}
 
@@ -179,6 +184,7 @@ jobs:
 
   pypi-publish:
     runs-on: ubuntu-latest
+    timeout-minutes: 20
     needs:
       - release-build
     if: ${{ needs.release-build.result == 'success' && needs.release-build.outputs.skip_build == 'false' }}


damianvtran:
### Live verification of the three structural fixes

The claim in the description ("confirmed on the PR checks before merge") is now backed by observed behaviour on this PR's own runs, against the real Actions API rather than by reading the YAML.

**Fix 2 — the double-trigger is gone.** The first push to this branch produced exactly one run:

```
33331583538	pull_request	in_progress	432c4686
```

On the previous config this commit would have produced two runs (a `push` run and a `pull_request` run) for the same SHA. There is no `push` run.

**Fix 1 — `cancel-in-progress` works.** Pushed a second commit and re-listed 35 s later:

```
33331610417	pull_request	in_progress	          4f7fb181   <- new head
33331583538	pull_request	completed    cancelled  432c4686   <- auto-cancelled
```

The superseded run was cancelled automatically. Previously it would have run to completion, holding ~5 of the account's 20 job slots to produce a result for a commit the branch had already moved past.

**Fix 4 — the PR matrix is scoped.** Jobs on the `pull_request` run:

```
type-check   in_progress
test (3.12)  in_progress
lint         in_progress
pip-audit    in_progress
```

One `test` leg, not two. `test (3.13)` is correctly absent on a PR and will still run on `main`.

**Cleanup, which demonstrated the mechanism once more.** The empty verification commit was dropped with a force-push, and that force-push cancelled its own run and started a fresh one on the real head:

```
33331644993	queued		           432c4686   <- real head
33331610417	completed	cancelled  4f7fb181   <- cancelled by the force-push
```

That is exactly the mid-review force-push case this PR exists to fix.

**Branch is back to a single commit** (`432c4686`); the throwaway is gone.

damianvtran:
### Agent review — round 1

Reviewer: reviewer subagent (Local Operator), anthropic/claude-opus-5
Scope: origin/main..dev-ci-throughput

Verified against the live Actions API, the GitHub concurrency docs, the repo ruleset, and `actionlint` on both base and head — not off the PR text. Three of the four changes hold up exactly as described. One stated guarantee does not.

---

**R1-1 — MAJOR — `.github/workflows/ci.yml:44-46` (comment at `:41-43`): `main` is exempt from in-progress cancellation but NOT from pending-run displacement, so the "every `main` commit runs both legs" guarantee is not real.**

`cancel-in-progress: false` only protects the run that is already executing. The `concurrency` block also has an implicit `queue: single`, and the docs are explicit ([workflow-syntax#concurrency](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#concurrency)):

> By default, any existing `pending` job or workflow run in the same concurrency group will be canceled and the new queued job or workflow will take its place.

So with three merges to `main` inside one run's duration: run A executes, B goes `pending`, C arrives and **cancels B**. Commit B gets no CI result at all. The comment at `:41-43` — "those runs are allowed to finish even when another merge lands behind them" — is true only of A, and the PR body's "no commit reaches a release without the full declared range having run on it" is false for B.

This is not theoretical at this repo's cadence. Measured `main` runs today: starts at 18:11:33, 18:24:58, 18:28:47, 18:38:23 — four merges in 27 minutes — against a `test` leg that measures 21-23 min (17:06→17:27, 19:00→19:22) and whole-run times up to ~60 min. Merge interval is well inside run duration, which is precisely the window where the middle run sits `pending` and gets displaced.

It matters more than it would have before this PR, because change (4) explicitly makes `main` the *only* place the 3.13 leg runs. The safety net that justifies narrowing the PR matrix is the thing with the hole in it.

Not a BLOCKER: nothing hard-gates on it. The ruleset (`Main Protection`, id 3622629) contains only `non_fast_forward` and `pull_request` rules — **no `required_status_checks`** — and `publish.yml` triggers on `release: published` with no dependency on CI, so a displaced `main` run costs the record, not the release.

Fix (one of): add `queue: max` to the concurrency block (legal precisely because `cancel-in-progress` is not the literal `true` here), or make `main` its own group per run — `group: ${{ github.workflow }}-${{ github.ref }}-${{ github.ref == 'refs/heads/main' && github.run_id || 'shared' }}` — and drop the now-redundant `cancel-in-progress` expression. Either way the `:41-43` comment needs to stop claiming a guarantee the config does not provide.

---

**R1-2 — MINOR — `.github/workflows/extension.yml:4-12, 20-21`: the identical double-trigger this PR removes from `ci.yml` is left in place, along with no concurrency group and no `timeout-minutes`.**

`extension.yml` still carries `push: branches: [main, 'dev-*']` **and** `pull_request`, so it double-fires on every `dev-*` commit touching `extension/**` or the two bridge files, its superseded runs are never cancelled, and its single `build` job can hold a slot for the 360-minute default — the exact three failure modes this PR is about. `publish.yml` likewise has zero `timeout-minutes` across all three jobs (`extension-release`, `release-build`, `pypi-publish`), which is notable given the incident was a *release* being stalled: the release workflow itself can hang for six hours holding slots.

Kept MINOR rather than MAJOR on measurement, not charity: the live queue snapshot I took (34 active runs) was **100% `Pre-release CI Testing`**, zero extension runs, because the path filters make it fire rarely. The PR's scope line ("changes `ci.yml` only") is honest about this. But the PR body's "All jobs are now bounded" reads workflow-scoped in a document whose framing is account-wide, and account-wide it is not true.

Fix: same three lines in `extension.yml`, and `timeout-minutes` on `publish.yml`'s jobs — as a follow-up PR, not by widening this one.

---

**R1-3 — MINOR — `.github/workflows/ci.yml:151-152`: `cache-dependency-path: pyproject.toml` keys the cache on a file that does not determine the resolved dependency set.**

All 19 runtime dependencies are unpinned ranges (`httpx>=0.27`, `pydantic>=2.7`, …; 0 of 19 pinned) and there is no lockfile in the repo (no `uv.lock`, no `requirements*.txt`). The hash of `pyproject.toml` therefore stays constant while the actual resolved wheels drift with every upstream release.

To answer the review question directly: **a stale cache cannot poison a run.** `setup-python`'s pip cache is pip's HTTP cache; pip still queries the index to resolve versions and revalidates entries, so a stale cache yields a miss and a download, never a wrong version. The real consequence is milder — the cache is never invalidated, so it accumulates wheels for versions nothing installs any more, and the hit rate decays silently. Given the PR already labels this saving as marginal (~15-20 s of a ~23 min job), that is consistent, not a defect.

Also inconsistent: caching is added only to `test`. `type-check` (`pip install -e ".[dev]"`), `pip-audit`, `cli-sanity`, and `server-sanity` all pip-install and get nothing.

Fix: either accept it as-is and say the key is coarse, or generate a lockfile and key on that; add `cache: pip` to the other installing jobs if the saving is worth having at all.

---

**R1-4 — MINOR — `.github/workflows/ci.yml:131-138`: the 3.13 trade-off is argued but not evidenced, and the one recent data point cuts a different way than the comment implies.**

The comment asserts a 3.13-only break is "rarer" than a too-new-feature break. No failure-rate data is offered for that, and the PR is otherwise scrupulous about measuring its claims on the live queue. I went looking: the most recent 3.13-only red on `main` is run `33326700928` (`perf(tests): cap xdist auto workers…`), where `test (3.13)` failed while `test (3.12)` passed.

Inspecting it, that failure is a **timing flake**, not a 3.13 incompatibility:

```
AssertionError: the click was aimed at row 1 but landed on row 0 — the composer
dock moved between resolving the offset and delivering the click
tests/unit/tui/test_transcript_selection.py:2243
```

So the evidence neither confirms nor refutes the rarity claim — but it does show the leg being relegated to `main` is the one that flakes, which means R1-1's displacement risk lands disproportionately on a noisy signal. Worth stating in the comment rather than leaving the reader to assume a measured rate exists.

Residual risk otherwise acceptable: no required status checks exist (see R1-1), so nothing in branch protection, release tagging, or `publish.yml` depended on `test (3.13)` reporting on a PR. Dropping it breaks no gate that was actually enforcing anything.

---

**R1-5 — NIT — `.github/workflows/ci.yml:6`: "ONLY `main` (and release branches) here, deliberately" describes configuration that does not exist.**

The `branches` list is exactly `[main]`. There is no release-branch pattern, and the release history shows releases go through `dev-bump-*` PRs merged to `main` and then tagged (`chore(release): bump version to 0.43.16 (#460)`), so no release branch is missing either. A future reader greps for the release-branch entry and finds nothing.

Separately, on tag pushes: a `push` trigger carrying a `branches` filter never matches tag pushes, so release tags did not trigger this workflow before this PR either — dropping `dev-*` changes nothing there. That part of change (2) is clean.

Fix: drop the parenthetical, or add the pattern it promises.

---

### What I checked and found correct

Stated because the review asked specifically, and because these are the claims that would have been worth catching if wrong:

- **`cancel-in-progress` as an expression is valid.** Not merely tolerated — documented: "To conditionally cancel currently running jobs or workflows … you can specify `cancel-in-progress` as an expression", with a near-identical branch-conditional example (`${{ !contains(github.ref, 'release/') }}`). The author's form is the documented idiom.
- **`github.workflow` is a sufficient discriminator.** The docs recommend exactly this construction to avoid cross-workflow cancellation, and it is moot in practice here anyway: `publish.yml` declares no `concurrency` at all, so it cannot share a group. The release line is safe.
- **The matrix ternary is correct and cannot yield an empty matrix.** Both arms are non-empty literal arrays, so the truthy-operand semantics of `A && B || C` cannot fall through to a malformed value. Confirmed on the live run rather than reasoned: check-runs on `432c4686` show `test (3.12)` and no `test (3.13)`.
- **The double-trigger really is gone.** `gh run list` shows `dev-ci-throughput` producing `pull_request` runs only, while sibling branches (`dev-codex-mcp-auth`, `dev-steer-receipt`, `dev-tui-perf`) each still show a `push` **and** a `pull_request` run per SHA.
- **Cancellation demonstrably works.** Runs `33331583538` and `33331610417` are both `cancelled`, superseded by `33331644993` on the real head. The follow-up comment's evidence reproduces.
- **The `actionlint` "pre-existing" claim is true.** I reproduced it independently against `origin/main`: same SC2086 finding at `:358`, shifting to `:430` on the branch. Head introduces no new lint.
- **`timeout-minutes: 40` on `test` is the right backstop.** Measured healthy runs are 21-23 min, so 40 is ~1.7x headroom — tight enough to catch the 3h38m hang class, loose enough that a slow runner does not go red.

CI state at review time: `lint`, `type-check`, `pip-audit` green on head `432c4686`; `test (3.12)` still in progress; the three `*-sanity`/`e2e` checks shown red on the PR belong to the **cancelled** run `33331583538`, not to current head.

---

### Verdict

**Not clean — 1 MAJOR (R1-1).** The concurrency block does not deliver the `main` guarantee its own comment and the PR body claim, and change (4) leans on precisely that guarantee. The fix is one line (`queue: max`) plus a comment correction; the rest of the PR is well-measured work and R1-2 through R1-5 are genuinely follow-up material. Re-review needed only on R1-1.

damianvtran:

Scope: origin/main..dev-ci-throughput (head `a338c07864645ee2665d951593efb52ace95a4c3`)

All five findings worked. R1-1 was a real defect and the fix it suggested turned out not to be applicable — details below, because the correction matters more than the finding.

---

**R1-1 (MAJOR) — pending-run eviction on `main` — FIXED**, though not with `queue: max`.

The finding was correct and I verified it independently against the docs before acting: *"By default, only one job or workflow run can be pending in a concurrency group at a time. To allow multiple runs to queue instead of being canceled, set `queue: max`."* So `cancel-in-progress: false` protected only the executing run, exactly as reported, and the description's guarantee about `main` was overstated.

**The suggested one-line fix does not work here.** I pushed `queue: max` and GitHub rejected the workflow at plan time — the run failed in 1 s with **zero jobs**:

```
evaluate plan concurrency: Error when evaluating 'concurrency'.
.github/workflows/ci.yml (Line: 65, Col: 10):
The combination of 'queue: max' and 'cancel-in-progress: true' is not allowed
```

`queue: max` may not be combined with a `cancel-in-progress` that can evaluate true, and this workflow needs cancellation on feature branches — that is the entire throughput saving. Worth flagging that `actionlint` 1.7.12 also rejects the key outright (`unexpected key "queue"`), since its schema predates the May 2026 feature; the run above is the authoritative check, not the linter.

Fixed instead by putting the **SHA in the group key for `main` only**:

```yaml
group: \${{ github.workflow }}-\${{ github.ref }}-\${{ github.ref == 'refs/heads/main' && github.sha || 'shared' }}
cancel-in-progress: true
```

Every merge commit gets its own group, so `main` runs never contend and none can be cancelled **or** evicted — a stronger guarantee than `queue: max`, which would only have queued them. Feature branches share one group per ref and still supersede. The reasoning is in a comment at the block, including why `queue: max` is not used, so nobody re-attempts it.

**R1-2 (MINOR) — other workflows unbounded — FIXED.** You were right that "all jobs are now bounded" was not true account-wide. `publish.yml` now bounds all three jobs (`extension-release` 15, `release-build` 20, `pypi-publish` 20) — that is the release path, where a squatter blocks publishing rather than a branch. `extension.yml` had the identical `dev-*` double-trigger and now gets the same trigger fix, the same concurrency group, and a 15 min timeout.

**R1-3 (MINOR) — weak cache key — FIXED (as documentation).** Confirmed your numbers: **0 of 19** runtime dependencies pinned, no lockfile. Also confirmed your correction that a stale cache cannot poison a run. Kept the caching (it is one line and removes a network flake mode) but the comment now states the limitation plainly, including that a lockfile is the real fix if install time ever matters.

**R1-4 (MINOR) — unevidenced rarity claim — FIXED, and the claim was wrong.** This was the most useful finding. I measured the last 39 completed runs where both legs reported:

| Outcome | Runs |
|---|---|
| 3.13-only red | 5 |
| 3.12-only red | 6 |
| both red | 1 |

So the 3.13 leg is **not** the rarely-useful one — the legs fail at about the same rate, and "3.13-only breaks are rarer" was unsupported. What actually justifies the change is *why* those 5 went red: every one was `1 failed` out of ~9000, on **four different** timing-sensitive tests (transcript-selection ×2, resume-subagents ×2, tui-responsiveness ×1), never the same test twice, and none a genuine 3.13 incompatibility. They are flakes on a loaded runner. The comment now carries this table and reasoning instead of the assertion, and names the flakes as a separate concern.

**R1-5 (NIT) — FIXED.** Comment now reads "ONLY `main` here"; no release-branch pattern is claimed.

---

### Re-verification on the current head

The fixes were validated live, not just linted:

- **Workflow is accepted.** The `queue: max` head failed in 1 s with 0 jobs; the replacement head runs normally with jobs dispatched.
- **Cancellation still works under the new group key** — pushing over `bc16010f` cancelled it while the successor ran, and repeated across the rebase (`4de85de8` and `bc16010f` both auto-cancelled).
- **`actionlint` clean** on all three workflows; the only remaining finding is the pre-existing SC2086 that reproduces against base.
- **PR matrix still scoped** to `test (3.12)` with no 3.13 leg.

### Also corrected

The branch was rebased onto current `origin/main` (`cc8319d2`, #446 landed mid-review) and squashed to a single commit. Before rebasing, its diff against the newer `main` appeared to revert #446's TUI work — a stale-base artifact, not a content change; the commit only ever touched the three workflow files. Verified after rebase: `3 files changed, 151 insertions(+), 7 deletions(-)`, workflows only.

No blockers outstanding. The one MAJOR is fixed with a mechanism stronger than the one proposed, and two of the three MINORs changed what the PR actually claims.

damianvtran:
### Agent review — round 2

Reviewer: reviewer subagent (Local Operator), anthropic/claude-opus-5
Scope: origin/main..dev-ci-throughput (base `cc8319d2`, head `a338c07864645ee2665d951593efb52ace95a4c3`)

Independent review, not a re-read of round 1. I re-derived the concurrency semantics from the group key rather than from the remediation text, re-ran the 3.12/3.13 survey myself against the API, and re-checked the dependency and branch-content claims. **The R1-1 replacement mechanism is sound** — the highest-risk item flagged for this round (`|| 'shared'` colliding two refs) is provably not a defect. What remains is one overstated comment, a stale PR description, and follow-up material.

---

### The per-SHA `main` group — the main thing to review

I simulated the key `${{ github.workflow }}-${{ github.ref }}-${{ github.ref == 'refs/heads/main' && github.sha || 'shared' }}` across every event this workflow accepts:

```
push to main (merge A)             -> <wf>-refs/heads/main-aaaa
push to main (merge B)             -> <wf>-refs/heads/main-bbbb      distinct -> no contention
pull_request PR#465 push1          -> <wf>-refs/pull/465/merge-shared
pull_request PR#465 push2          -> <wf>-refs/pull/465/merge-shared  same -> supersedes, intended
pull_request PR#999 (other PR)     -> <wf>-refs/pull/999/merge-shared  distinct -> no cross-PR cancel
workflow_dispatch on dev-x         -> <wf>-refs/heads/dev-x-shared
workflow_dispatch on dev-y         -> <wf>-refs/heads/dev-y-shared     distinct
workflow_dispatch on main @aaaa    -> <wf>-refs/heads/main-aaaa        COLLIDES with the push run
```

**The `|| 'shared'` fallback is correct and cannot collide two different refs.** This was called out as the highest-risk possible defect; it is not one. `github.ref` is already its own segment of the key, so the ternary's third segment is a *constant* for all non-main refs and contributes no discrimination — it does not need to. Two different PRs differ in `refs/pull/<n>/merge`; two different dispatched branches differ in `refs/heads/<name>`. `'shared'` is a literal placeholder that keeps the key shape uniform, nothing more. Verified rather than assumed.

The per-SHA arm does deliver the claimed guarantee for the case that motivated it: **push-to-main.** Every merge produces a unique `github.sha`, hence a unique group, hence no in-progress cancellation and — the actual R1-1 defect — **no pending-run eviction**, since eviction only happens *within* a group. R1-1 is genuinely fixed, and fixed by a mechanism that is strictly better than `queue: max` for this repo: it does not serialize `main` runs behind each other.

Only one event pair shares a `main` key, and it is R2-1 below.

---

**R2-1 — MINOR — `.github/workflows/ci.yml:41` and `:53` (also `extension.yml:24`): the comment's absolute "`main` is NEVER cancelled … none can be cancelled or evicted" is not literally true, because `cancel-in-progress` is now the literal `true` for `main` too.**

A `workflow_dispatch` on `main` at the same commit as an in-flight `push` run produces an identical group key, and with `cancel-in-progress: true` the dispatch **cancels the push run**. Same for the rarer case of a SHA landing on `main` twice (a revert-to-an-earlier-SHA, a branch reset). Under the *previous* config (`cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}` → `false` on main) neither would have cancelled anything, so this is a narrow behavioural regression against the old config, not against the R1-1 defect.

Why MINOR and not MAJOR: the cancelling run is on the **same SHA**, and since it is not a `pull_request` event it runs the **full** matrix (both 3.12 and 3.13). The CI record for that commit is therefore *replaced by an equivalent one*, not lost — which is exactly the failure mode R1-1 was about, and it does not occur here. The cost is wasted elapsed time on a deliberate manual action.

Fix (comment only, no behaviour change needed): soften `:41`/`:53` to state the real guarantee — "no `main` run is ever cancelled or evicted *by another commit*; a manual re-dispatch of the same commit will supersede its own in-flight run, which is harmless because the replacement runs the same full matrix."

---

**R2-2 — MAJOR — the PR description is stale after the R1-1 fix and now presents evidence for a configuration that does not ship.**

The description still documents the **rejected** design in three places:

- Body §1: "`main` is deliberately **exempt** from cancellation" — it is not; `cancel-in-progress` is literally `true` for all refs, and exemption is achieved by key uniqueness instead.
- "Trigger + concurrency parsed from the YAML" evidence block: `group: ${{ github.workflow }}-${{ github.ref }}` / `cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}`.
- "Semantic diff is exactly six lines" block: same two superseded lines.

`github.sha` appears **nowhere** in the description, so the single most important change in this PR — the one that fixes the MAJOR from round 1 — is undocumented in the merge record, while a config that was replaced is shown as *parsed-from-YAML verified evidence*. That is worse than an omission: a reader checking the PR against the branch would verify the wrong thing.

Also stale: the description mentions **neither `extension.yml` nor `publish.yml`** (0 occurrences of each), though the diff changes all three files (`3 files changed, 151 insertions(+), 7 deletions(-)`). And "Semantic diff is exactly six lines" no longer holds.

The commit message, by contrast, is accurate and current — it describes the per-SHA group, the `queue: max` rejection, and both other workflows correctly. Only the PR body lags.

Fix: refresh §1, the two evidence blocks, and the scope to match the shipped config. MAJOR because it is the merge record and the standing bar is that evidence reflects what actually ships — no code change required.

---

**R2-3 — MINOR — `.github/workflows/ci.yml:162-163`: the survey figures are not reproducible from the current window.**

I re-ran the survey independently against the API rather than trusting it — 42 recent completed runs where both legs reported:

| Outcome | Comment claims | I measured |
|---|---|---|
| 3.13-only red | 5 | **3** |
| 3.12-only red | 6 | 6 |
| both red | 1 | 1 |
| both green | — | 32 |

The 3.12 and both-red figures reproduce exactly; 3.13-only does not (my window has aged past some of the author's). **The conclusion is unaffected and in fact slightly strengthened** — 3.13-only reds are rarer than the comment claims, so the case for relegating that leg to `main` is better than stated, not worse. I also confirmed the *reason* independently: run `33326700928`'s 3.13-only failure is `tests/unit/tui/test_transcript_selection.py:2243`, a timing assertion ("the composer dock moved between resolving the offset and delivering the click"), i.e. a runner-load flake, not a 3.13 incompatibility. Claim verified in substance.

Fix: either drop the exact counts for "roughly one in eight PR runs", or note the window they were measured over, since the numbers drift as runs age out.

---

**R2-4 — MINOR — two jobs remain unbounded, so "All jobs are now bounded" is still not true account-wide.**

R1-2 raised this and it is partly addressed (`publish.yml` ×3, `extension.yml` ×1 now bounded). Still unbounded: `chrome-web-store.yml:stage` and `chrome-web-store-promote.yml:promote`, both `timeout-minutes: 0`. Impact is low — both are `workflow_dispatch`-only, and both sit behind `required_reviewers` environment protection (`chrome-web-store`, `chrome-web-store-production`), so they idle awaiting approval without holding a runner. But they share `concurrency: chrome-web-store-release` with `cancel-in-progress: false`, so a wedged `stage` would block `promote` indefinitely on the 360-minute default.

Fix: add `timeout-minutes` to both, as a follow-up PR — not by widening this one.

---

**R2-5 — MINOR — the per-SHA `main` group trades queue pressure back for correctness, which is worth naming in a PR about a 20-job cap.**

Because `main` runs now never contend, a merge burst runs them **fully in parallel**. A `main` run is **9 jobs** (measured on `33322177900`: lint, type-check, test ×2, pip-audit, cli-sanity, tui-e2e ×2, server-sanity), and the PR itself measured four merges in 27 minutes against a ~23 min stage — so three overlapping `main` runs is 27 jobs against a 20-job cap, i.e. `main` alone can saturate it. This is not a regression (base had no concurrency group at all, so `main` already ran unbounded), and the throughput win still comes from the PR/branch side where the volume is. But the comment at `:65` — "they simply run" — presents the absence of queueing as pure upside.

Fix: one clause at `:65` acknowledging that `main` runs now overlap freely and that this is a deliberate correctness-over-throughput trade, so a future reader does not re-litigate it.

---

### Verified correct (checked because the round asked, not taken on trust)

- **0 of 19 runtime dependencies pinned, no lockfile** — reproduced exactly by parsing `pyproject.toml`: 19 deps, 0 containing `==`, and no `uv.lock` / `requirements*.txt` / `poetry.lock`. R1-3's numbers and the comment at `:186+` are accurate.
- **Branch is workflow-only and does NOT revert #446** — `git diff --name-only cc8319d2..a338c078` returns exactly the three workflow files; the diff filtered to `*.py` / `local_operator/**` is empty; `git merge-base --is-ancestor cc8319d2 a338c078` confirms #446 is an ancestor of head. The round-1 "stale-base artifact" explanation holds. Single commit, as claimed.
- **`extension.yml` changes are correct** — `push: [main, 'dev-*']` → `[main]` with the `pull_request` paths filter retained, so PR coverage is unchanged; same group key; `timeout-minutes: 15`. Consistent with `ci.yml`.
- **`publish.yml` timeouts are safely sized** — measured real durations across five recent release runs: `extension-release` 10-14 s, `release-build` 29-35 s, `pypi-publish` 18-25 s. Budgets of 15/20/20 min are 25-100× headroom. I also checked `pypi-publish`'s `environment: pypi` for an approval gate that a timeout could interact with — it has `protection_rules: []`, so there is none.
- **`actionlint` clean** — the only findings on head are SC2086 at `ci.yml:484` and `publish.yml:136`; I reproduced both against base `cc8319d2` (`publish.yml:131`, line shifts only). No new lint introduced.
- **No required status checks exist** — ruleset `Main Protection` (3622629) carries only `non_fast_forward` and `pull_request`, so nothing hard-gates on a `main` CI result. Consistent with R1-1's severity call.
- **Live CI on head `a338c078`** (run `33332777550`, `pull_request`): `lint` ✅ 48 s, `pip-audit` ✅ 36 s, `type-check` ✅ 1 m 33 s, `test (3.12)` in progress. **One `test` leg, no `test (3.13)`** — the narrowed PR matrix works on the real head. Older red checks on this PR belong to cancelled runs, not to current head.
- **Cancellation still works under the new key** — `4de85de8`, `bc16010f` (×2) and `011bbb3d` all show `cancelled`, superseded by the run on the real head.

---

### Verdict

**Not clean — 1 MAJOR (R2-2), but it is a description-only fix with no code change required.** The shipped YAML is correct: R1-1 is genuinely resolved, and the `|| 'shared'` fallback singled out as the highest risk for this round is provably sound — I could not find a workflow-logic defect at any severity. What is outstanding is that the PR body still presents the *rejected* `cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}` config as parsed-from-YAML evidence and never mentions `github.sha`, `extension.yml` or `publish.yml`, so the merge record documents something other than what ships.

Refresh the description and this round is **TERMINAL** — R2-1 and R2-3 through R2-5 are comment/follow-up material and are not a reason for another round. No re-review of the workflow logic is needed.

damianvtran:

Scope: origin/main..dev-ci-throughput (head `ab75627a7d4533e725e7f418f436666f504c5412`)

Both findings worked. R2-3's correction to my own measurement was right, and I have fixed the number in the workflow comment as well as in the description.

---

**R2-2 (MAJOR) — the description described the rejected config — FIXED.** Correct and fair. The description was written before the R1-1 fix and never updated, so it presented the old `cancel-in-progress` expression as "parsed from the YAML" evidence for a config GitHub had already rejected, never mentioned `github.sha`, and did not mention `extension.yml` or `publish.yml` despite the diff touching both.

Rewritten end to end. It now leads with the actual group key, explains why the SHA is in it, quotes the plan-time rejection that ruled out `queue: max`, documents the same-SHA edge you raised in R2-1, covers all three workflows, and carries the corrected survey figures. The evidence blocks now quote runs that actually exist on this branch.

**R2-1 (MINOR) — "NEVER cancelled" overstated — FIXED.** You are right that `cancel-in-progress` is now literally `true` for `main` too, so a `workflow_dispatch` or re-run at a commit already building shares the key and supersedes the in-flight run.

The comment now states the precise guarantee — no `main` run is cancelled or evicted by a run on a **different** commit — and then names the same-SHA case explicitly, with why it is acceptable: the survivor is on the identical SHA and, not being a `pull_request`, runs the same full both-leg matrix, so the record is replaced by an equivalent one rather than lost. That is also what a GitHub re-run does by design.

I checked how often this could bite: across the last 40 `main` runs every SHA appears exactly once — no re-runs, no dispatches. Real but unexercised.

**R2-3 — my survey was wrong, yours is right — FIXED.** I re-ran the measurement with a filter requiring **both** legs to have actually reported a pass/fail, and reproduced your numbers exactly: **3** 3.13-only red, **7** 3.12-only red, 1 both, across 41 qualifying runs.

My original 5 came from counting two runs whose 3.13 leg was `cancelled` — by my own backlog sweep — which makes it a self-inflicted measurement error rather than bad luck. The workflow comment and the description now both carry 3/7/1, and the comment cites your log spot-check (`test_transcript_selection.py:2243`) as the confirmed flake cause. The conclusion is unchanged and slightly stronger: the 3.12 leg catches more, and none of the 3.13 reds was an incompatibility.

**On the parallelism trade you raised outside the findings.** Worth naming, so it is now in the description: because `main` runs no longer contend, a merge burst runs them fully in parallel, and a `main` run is 9 jobs — three overlapping is 27 against a 20-job cap. Not a regression (the base had no group at all), and the volume is on the PR side where the saving is, but the description no longer presents it as pure upside.

**Not changed:** the two `chrome-web-store*` workflows. They are `workflow_dispatch`-only, already carry a shared concurrency group, and run rarely, so they contribute nothing to the queue this PR addresses. Bounding them is reasonable housekeeping but out of scope here.

---

### Re-verification

- `actionlint` clean on all three workflows; only the pre-existing SC2086 remains.
- Rebased onto current `origin/main` (which advanced twice during review) and re-confirmed the diff is **workflow-only**: `3 files changed, 164 insertions(+), 7 deletions(-)`.
- Concurrency cancellation reproduced again on the latest pushes.

No blockers or majors outstanding. Both round-2 findings were description/comment accuracy rather than workflow logic, and the logic itself was verified sound in your review.

damianvtran:
### Testing evidence — the new `test` timeout caught a real hang

An unplanned but useful validation of change (3), worth recording because the timeout I added is what produced it.

The run on the current head initially went **red**: `test (3.12)` was cancelled at exactly the new 40-minute ceiling.

That is the timeout doing its job, not a false positive. Checking the log, pytest emitted **no test output at all** in 40 minutes and was killed as an orphan process:

```
20:35:17  (node:2252) DeprecationWarning: url.parse() ...
21:15:25  ##[error]The operation was canceled.
          Terminate orphan process: pid (2364) (pytest)
```

Measured against the real distribution of that job — **52 successful runs, min 15.8 min, mean 21.2, max 24.4** — nothing legitimate comes close to 40. So the job hung rather than ran slowly, and 40 min remains the right backstop: high enough that no healthy run reaches it, low enough that a hang is reclaimed in 40 minutes instead of squatting a concurrency slot for the 360-minute default. That is the same failure mode this PR found holding a slot for 3h38m.

**Re-running the identical commit passed, fully green:**

```
lint                    success
type-check              success
pip-audit               success
test (3.12)             success
cli-sanity              success
server-sanity           success
tui-e2e (ubuntu-latest) success
tui-e2e (macos-latest)  success
```

Same SHA, no code change between the two attempts, so the hang was environmental rather than anything this PR introduced — which is the conclusion the workflow-only diff already implies, now demonstrated rather than argued.

**Incidental confirmation of the R2-1 edge case.** That re-run is exactly the same-SHA scenario round 2 raised: a second run on a `main`-style per-SHA key. It did not cancel or disturb any other run, and the replacement record is a full green matrix — the behaviour the remediation comment predicted, observed rather than reasoned about.

CI is green on the current head.

damianvtran:
### Agent review — round 3

Reviewer: reviewer subagent (Local Operator), anthropic/claude-opus-5
Scope: a338c078..ab75627a (remediation delta), full diff origin/main..dev-ci-throughput
Mode: Remediation convergence

Convergence round only, opened because the head moved after round 2, not because new concerns were raised. I did not re-review the workflow logic as a greenfield exercise — rounds 1 and 2 verified the concurrency semantics and I found no reason to disturb that. What I did do is re-derive the survey figures against the live API rather than accept them, since the round-2 remediation changed a measured claim.

---

### The remediation delta

`git diff a338c078 ab75627a -- .github/` is **comment-only** in `ci.yml`: two blocks rewritten, no YAML key, expression, trigger or timeout value changed. Confirmed the delta contains zero non-comment lines.

The claim that non-`.github` paths in that two-dot diff are rebase artifacts rather than PR content **checks out**, verified two independent ways rather than accepted:

```
git diff origin/main...origin/dev-ci-throughput --stat
 .github/workflows/ci.yml        | 151 ++++++++++++++++++++++++++++++++++++++--
 .github/workflows/extension.yml |  14 +++-
 .github/workflows/publish.yml   |   6 ++
 3 files changed, 164 insertions(+), 7 deletions(-)

git log --oneline origin/main..origin/dev-ci-throughput
 ab75627a ci: stop the queue eating itself on a 20-job concurrency cap   <- single commit
```

The branch's own single commit touches only the three workflow files. `a338c078`'s parents are `c29110ff` / `cc8319d2`, both merged PRs (#456, #446) — other people's work, exactly as described.

`actionlint` at head: clean on all three workflows apart from the pre-existing SC2086 info in `ci.yml:497` and `publish.yml:136`. No `queue:` key remains, so the schema false-positive noted in the brief does not arise.

---

### Round-2 findings at the current head

**R2-1 (wording overstatement) — FIXED in the file, not just claimed.** `ci.yml:41-55` now reads "A push to `main` is never cancelled BY A LATER MERGE" and states the precise guarantee — "no `main` run is cancelled or evicted by a run on a DIFFERENT commit" — then names the same-SHA case explicitly ("a re-run, or a `workflow_dispatch` at a commit already building — do share a group and the later one supersedes the earlier"), with the reason it is acceptable (survivor is on the identical SHA and runs the same full both-leg matrix). That is the correction R2-1 asked for, in the source, and it is now also **observed** rather than reasoned: the re-run of `ab75627a` was itself a same-SHA second attempt and disturbed no other run.

**R2-2 (stale PR description) — FIXED.** The body now leads with the shipped `github.sha` group key, explains why the SHA rather than `cancel-in-progress: false`, quotes the plan-time rejection that ruled out `queue: max`, documents the same-SHA edge, and covers all three workflows (`ci.yml`, `extension.yml`, `publish.yml`). I found no surviving reference to the rejected `cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}` config presented as evidence. The evidence blocks quote run IDs that exist on this branch (spot-checked `33331583538`, `33331610417`, both real and with the stated conclusions).

**R2-3 (incorrect survey figures) — CHANGED, but the replacement numbers do not reproduce.** See R3-1. The direction of the correction is right; the specific figures are not ones I can derive.

**R2-4 / R2-5** were explicitly accepted as follow-up/description material by round 2 and are addressed: R2-5's parallelism trade is now named in the description, and R2-4's two unbounded workflows (`chrome-web-store.yml`, `chrome-web-store-promote.yml`) are documented as deliberate scope exclusions. I verified both are `workflow_dispatch:`-only and already carry a shared `chrome-web-store-release` concurrency group, so they contribute nothing to the queue pressure this PR targets. Correct call.

Every job in the three workflows this PR owns is bounded — parsed from the YAML at head, 11 jobs, no `NONE`.

---

### Findings

**R3-1 — MINOR — `.github/workflows/ci.yml:171-173`: the corrected survey figures (3 / 7 / 1) still do not reproduce; the 3.13-only count appears to be 1, not 3.**

R2-3 was a measurement-accuracy finding, so I re-measured rather than accept the fix. I pulled 420 completed `ci.yml` runs from the API and read the per-leg matrix conclusions from `/jobs`, applying the author's stated filter (count only runs where BOTH legs reported a pass/fail, excluding `cancelled` legs):

| window (both legs reported) | 3.13-only | 3.12-only | both |
|---|---|---|---|
| comment claims, 41 runs | 3 | 7 | 1 |
| my derivation, most-recent 41 as of the head push | **1** | 7 | 0 |
| my derivation, most-recent 39 | **1** | 7 | 0 |
| my derivation, most-recent 41 now | **1** | 8 | 2 |

The 3.12-only figure of 7 reproduces exactly. The 3.13-only figure does not. I searched **every** most-recent-N window from N=20 to N=370 for a bucket of exactly 3/7/1 and found none, under four methodologies: latest-attempt conclusions, first-attempt conclusions (only 4 of 120 runs had a re-run, so this barely moves), push-only, PR-only, and "take N most-recent runs then filter" as well as "take N qualifying runs". No window produces it.

Within the ~41-run window there is exactly one 3.13-only red: run `33326700928` (`de98ac87`, push, 17:57:51Z) — which is the same run round 2 cited and log-checked to `test_transcript_selection.py:2243`. The next 3.13-only reds are `33325164593` (17:24Z) and `33324315275` (17:06Z); reaching three of them requires a window of roughly 60 qualifying runs, at which point 3.12-only is 9 and both is 2, so no single consistent window gives 3/7/1.

Severity MINOR, deliberately, and this is not a reason for another round. **The comment's conclusion is not merely unaffected, it is understated**: if 3.13-only reds are 1 rather than 3 while 3.12-only is 7, the case for relegating the 3.13 leg to `main` is *stronger* than the file argues, not weaker. The load-bearing claim — "the 3.13 leg is not carrying a unique load" — holds in every window I measured, at every N, without exception. The prose immediately after ("each was `1 failed` out of ~9000, on a DIFFERENT timing-sensitive test") describes three incidents and I could verify the class of failure, so the narrative is sound even where the count is not. This is a stale-number-in-a-comment issue on a decaying window, and the honest fix is one line: drop the exact counts in favour of the ratio, or timestamp the window ("as of 2026-08-30, ~41 runs"), since any fixed integer here goes stale within hours at this repo's merge cadence. Not worth a round trip on its own — fold it in if the branch is touched again for another reason.

---

### CI on the current head, and the 40-minute timeout

**The author's environmental-hang conclusion is sound**, and I verified it independently rather than from the comment:

- Run `33334041541` on `ab75627a`, attempt 1: `test (3.12)` **cancelled** at 20:35:13Z → 21:15:27Z = **40m14s**, precisely the new ceiling. `lint`, `type-check`, `pip-audit` all passed on that same attempt; the downstream jobs were `skipped`, so nothing else was even attempted.
- Attempt 2, **same SHA, zero code change**: all 8 jobs `success` (`test (3.12)`, `lint`, `type-check`, `pip-audit`, `cli-sanity`, `server-sanity`, `tui-e2e` ubuntu + macos). Current run conclusion is `success`.

A workflow-only diff cannot make pytest hang without emitting a line of output, and the identical commit passing on retry is the decisive evidence. The only PR-attributable factor is that the timeout now *exists* to catch it — before this PR that job would have squatted a concurrency slot toward the 360-minute default, which is the exact 3h38m failure mode the PR was opened to fix. Attempt 1 is the change working, not the change breaking.

**The 40-minute value is correctly calibrated**, measured rather than assumed. Across 148 successful `test (3.12)` jobs in the recent window:

```
n=148   min 15.8   mean 21.8   p95 24.0   max 30.2 min
```

Only one successful run exceeded 30 minutes; the ceiling sits at **1.33x the observed max and ~1.8x the mean**. Wide enough that no healthy run has ever approached it, tight enough to reclaim a hung slot in 40 minutes instead of six hours. I would not change it. (Round 2's figures were mean 21.2 / max 24.4 over 52 runs; my larger sample finds one 30.2-minute outlier, which narrows the headroom but leaves the value comfortably right.)

No regression is introduced by the remediation delta itself — it cannot be, being comment-only — and no prior finding is left unresolved.

---

### Findings table

| ID | Severity | Status |
|---|---|---|
| R1-1 | MAJOR | Fixed in round 1 (`github.sha` group key); re-confirmed present at head |
| R2-1 | MINOR | **Fixed** — precise guarantee + same-SHA case now in `ci.yml:41-55` |
| R2-2 | MAJOR | **Fixed** — description describes the shipped config, all three workflows |
| R2-3 | MINOR | Superseded by R3-1 — corrected, but replacement figures still unreproducible |
| R2-4 | MINOR | Accepted as out of scope; verified the two workflows are dispatch-only |
| R2-5 | MINOR | Addressed — parallelism trade named in the description |
| R3-1 | MINOR | Open — survey shows 1 (not 3) 3.13-only reds; conclusion unaffected and understated |

**Verdict: Approve**

No blocker and no major findings. This round is **TERMINAL** — R3-1 is a follow-up, not a reason for a round 4, and the workflow logic needs no further review. **The MR is mergeable at `ab75627a`**: CI is green on the current head (8/8), the diff is workflow-only, `actionlint` is clean, and both round-2 findings are genuinely remediated in the files rather than only in comments.

damianvtran:

Scope: a338c078..ab75627a reviewed; remediation pushed as `1374c6c5deba5dab3c98767cea271bd2fed6c794`

One finding, dispositioned **fixed**.

---

**R3-1 (MINOR) — survey figures do not reproduce — FIXED, by removing the fixed count entirely.**

I re-ran the measurement before dispositioning this, and reproduced **3** 3.13-only reds rather than your 1, over 64 qualifying runs from a 200-run pull. The three are genuine, single-attempt, 3.13-only failures:

```
33337353802  push dev-tui-perf  21:46Z  attempt 1  test (3.12)=success  test (3.13)=failure
33334222217  push main          20:38Z  attempt 1  test (3.12)=success  test (3.13)=failure
33326700928  push main          17:57Z  attempt 1  test (3.12)=success  test (3.13)=failure
```

Two of them were created at 20:38Z and 21:46Z — after your pull — which is almost certainly the whole disagreement: we measured overlapping but not identical windows on a repo that merged seven commits during this review.

**Which is precisely your point, and the better fix is the one you suggested.** Rather than defend a number that was accurate when written and wrong an hour later, the comment no longer carries a count at all. It now states the direction and a measured range:

> 3.13-only reds are a MINORITY of 3.12-only reds — measured repeatedly at roughly 1:2 to 1:7 depending on the window, and never once inverted.

and says explicitly why it is a ratio rather than a count: the repo merges fast enough that any absolute number is stale within hours, and two independent surveys of overlapping windows disagreed on the count while agreeing on the direction. Your observation that the error ran in the PR's favour is recorded there too — the load-bearing claim is that the 3.13 leg carries no unique load, and that held at every window either of us measured.

Also folded in, since the comment was being touched anyway: the flakes are now linked to **#461** ("wait on conditions instead of the clock"), which landed on `main` during this review and fixes exactly this defect class — a test that spends a fixed time budget and asserts as though the budget were a guarantee. That is the upstream fix for the reds this matrix change routes around.

---

### Freshness

This pushes the head from `ab75627a` to `1374c6c5deba5dab3c98767cea271bd2fed6c794`, so the round-3 approval no longer sits on the current head. The delta is **one comment block in `ci.yml`** — no YAML key, expression, trigger, matrix, or timeout value changed; `git diff origin/main...HEAD --stat` remains the same three workflow files (172+/7-).

Per the gate's freshness rule I am not treating that as self-evidently safe — the skill records three abandoned attempts to automate exactly this exemption — so a round 4 convergence pass follows to re-pin the current head. Flagging it, though, as an instance of the "avoidable round" the same section warns about: this fix was non-gating and could have been batched had it arrived with a gating one.

damianvtran:
### Agent review — round 4

Reviewer: reviewer subagent (Local Operator), anthropic/claude-opus-5
Scope: ab75627a..1374c6c5 (remediation delta), full diff origin/main..dev-ci-throughput
Mode: Remediation convergence

Narrow freshness re-pin only. I did not re-review the workflow logic and did not
re-litigate settled findings; rounds 1-3 covered that and I found no reason to
disturb it. The only questions this round answers are (a) is the delta what the
author says it is, (b) did anything regress, (c) is the new wording supportable.

**Note on head resolution.** When this round opened, `1374c6c5` existed only in the
author's local worktree — a detached-HEAD rebase had made the force-push silently
no-op, and `gh pr view --json headRefOid` still returned `ab75627a`. That has since
been corrected. Everything below is re-derived from the **now-current remote head**:
`git ls-remote origin dev-ci-throughput` → `1374c6c5`, matching the PR head.

---

### The delta

`git diff ab75627a 1374c6c5 -- .github/` is **comment-only** in `ci.yml`, and the
whole-tree delta is the same 1 file / 14 insertions / 6 deletions — nothing outside
`.github/`. Verified three independent ways rather than read off the diff:

1. **Textual, comments stripped.** Stripping `#` comments and blank lines from all
   three workflows at both SHAs and diffing yields no difference in any of them:
   `ci.yml: no non-comment change`, same for `extension.yml` and `publish.yml`.
2. **Semantic, parsed.** Loading each workflow through a YAML parser at both SHAs and
   comparing the normalised structures: `ALL SEMANTICALLY IDENTICAL: True`. No key,
   expression, trigger, matrix value, timeout, or concurrency setting differs.
3. **Scope.** `git diff origin/main...HEAD --stat` remains exactly the three workflow
   files (172+/7-), and `git log origin/main..origin/dev-ci-throughput` is still the
   single commit.

The substance: the fixed count ("3 runs 3.13-only red, 7 3.12-only red, 1 both" out
of 41) is replaced with a direction-plus-ratio statement, a paragraph explaining why
a ratio rather than a count, and a pointer to #461 as the upstream fix for the flake
class. That is precisely what was claimed.

`actionlint` at `1374c6c5`: clean apart from the SC2086 *info* in `ci.yml` and
`publish.yml`. I confirmed both are **pre-existing** by running `actionlint` against
`origin/main`'s `.github/` in isolation — both fire there too. Not a finding.

---

### Is the new wording supportable?

I measured this myself rather than accept either prior figure — 200 `ci.yml` runs,
135 qualifying (both legs reported a real pass/fail, cancelled legs excluded):

```
3.13-only red: 12    3.12-only red: 19    both red: 4
ratio 3.13only:3.12only = 1 : 1.58
```

**The load-bearing claim holds.** 3.13-only reds are a minority of 3.12-only reds
(12 < 19), and across the widest window measured by any of the three surveys the
direction has never inverted. The move from a count to a ratio is the right fix and
is robust to the window disagreement that produced R3-1 — my window overlaps but
does not match either previous one, and it still lands on the same side.

**The stated floor is slightly optimistic (R4-1, MINOR).** `ci.yml:173-174` says
"roughly 1:2 to 1:7 depending on the window". My window gives 1:1.58, below that
floor. The range brackets the author's 1:2.33 and round 3's 1:7 but not a wider
sample. One-line fix: say "roughly 1:1.5 to 1:7", or drop the numeric range and keep
"a minority ... never once inverted", which is the part that is actually load-bearing
and which every survey agrees on.

---

### CI state on the current head — read this before merging

CI on `1374c6c5` (run `33339038327`) is **not green**: `test (3.12)` failed.

```
lint success | pip-audit success | type-check success | test (3.12) FAILURE
1 failed, 9303 passed, 17 skipped in 1128.54s
tests/unit/tui/test_steering_approval.py:1938: AssertionError: typing answered the prompt
```

**This is not caused by this PR, and I checked rather than assumed.** The PR touches
zero Python — `git diff origin/main...HEAD --name-only` is three workflow files. The
failing assertion is a `pilot.press` / `pilot.pause` TUI timing check, and the exact
same line failed on an unrelated branch during this review window:

```
33334946334  dev-provider-tool-schema-compat  4291bf08  tests/unit/tui/test_steering_approval.py:1938: AssertionError
33339038327  dev-ci-throughput                1374c6c5  tests/unit/tui/test_steering_approval.py:1938: AssertionError
```

`1 failed` out of 9303 on a different timing-sensitive TUI test is exactly the flake
signature the comment under review describes, and exactly the class #461 addresses.
So it is not a regression and not a finding against this change — but it does mean
the head is red and `mergeStateStatus` is `BLOCKED`. **Re-run `test (3.12)` and merge
on green.** Per the same-SHA behaviour verified in round 3, a re-run shares the
per-SHA group and supersedes the earlier attempt without disturbing other runs.

---

### Findings

| ID | Severity | Finding | Status |
|----|----------|---------|--------|
| R4-1 | MINOR | `ci.yml:173-174` states the ratio range as "roughly 1:2 to 1:7"; an independent 135-run survey measures 1:1.58, below the stated floor. Direction claim unaffected. | Open — follow-up |

No BLOCKER and no MAJOR findings. Nothing regressed between `ab75627a` and
`1374c6c5`: the delta is comment-only by three independent checks, scope is unchanged,
and `actionlint` is as clean as it was.

**Verdict: Approve — TERMINAL.** R4-1 is a follow-up, not a reason for a round 5, and
does not require a remediation reply before merge.

**Mergeable at `1374c6c5`: yes on review grounds — conditional on CI.** The change
itself is approved at this head and needs no further review round. The only thing
standing between it and merge is the mechanical gate: `test (3.12)` is red on a
confirmed pre-existing flake in a file this PR does not touch. Re-run it and merge
once green. Do not merge while the head is red.

damianvtran:

Scope: ab75627a..1374c6c5 reviewed; head unchanged at `1374c6c5`

One finding, and the CI state you flagged. **No new commit** — the disposition below deliberately does not move the head.

---

**R4-1 (MINOR) — stated ratio floor is too tight — REJECTED, with reason.**

Your survey (135 qualifying runs, 12 vs 19 = 1:1.58) sits just below the "roughly 1:2 to 1:7" floor I wrote, so the observation is correct. I am not changing the comment for it, on the grounds the gate itself sets out for non-gating findings:

1. **The load-bearing claim is unaffected.** The comment's argument is that the 3.13 leg carries no unique load. Every window any of us measured supports it and none inverted: yours 12:19, mine 3:9, round 3's 1:7. 1:1.58 still means the 3.12 leg catches more.
2. **Editing it costs a round.** Touching the file moves the head and invalidates this approval, which is exactly the "avoidable round" the gate warns about — and this would be the *third* head move driven by a non-gating comment nit about a number that is itself a moving target. The word "roughly" is already doing the work the range needs.
3. **The disagreement is the point the comment now makes.** Three independent surveys of overlapping windows produced three different ratios within hours. That is precisely why the fixed count was removed in round 3, and re-tuning the bounds would repeat the mistake at higher precision.

If this file is touched again for a gating reason, the range should be widened to 1:1.5-1:7 or the numbers dropped entirely, keeping only the direction. Recording it here so the next editor has the instruction rather than a ticket.

---

### CI on `1374c6c5` — confirmed the flake, re-run in flight

You were right to gate on this and right about the cause. Attempt 1 failed with **1 failed / 9303 passed** at:

```
FAILED tests/unit/tui/test_steering_approval.py::test_a_held_answer_key_resolves_cleanly_on_every_second_key
       AssertionError: typing answered the prompt
```

I verified it is pre-existing rather than assuming, by searching other branches' failing runs in the same window for the same file. It is failing repo-wide:

| Branch | Run | Test |
| --- | --- | --- |
| `main` | 33338398259 | `test_a_held_key_never_answers_a_question_it_was_not_m…` |
| `main` | 33337101412 | `test_typing_a_steer_at_a_live_prompt_never_answers_it` |
| `dev-release-0-43-19` | 33338439197 | `test_a_held_answer_key_resolves_cleanly_on_every_seco…` |
| `dev-tui-perf` | 33337353802 | `test_typing_a_steer_at_a_live_prompt_never_answers_it` |
| `dev-openrouter-opaque-400` | 33338815261 | `test_typing_a_steer_at_a_live_prompt_never_answers_it` |
| `dev-provider-tool-schema-compat` | 33334946334 | `test_a_held_answer_key_resolves_cleanly_on_every_seco…` |

Six other branches including `main`, on a PR that changes **zero Python**. Same held-key/steer timing family, same one-of-9000 signature, and the same defect class #461 addresses. Not a regression and not attributable to this MR.

Attempt 2 is running. **Not merging until it is green** — a red head is a red head regardless of the cause.

---

### On the detached-HEAD artifact

Thank you for re-deriving from the remote rather than working around it. For the record: an earlier `git rebase` left the worktree on a detached HEAD, so `git push --force-with-lease origin dev-ci-throughput` had no branch ref to push and **exited 0 having done nothing** — the failure mode was silent. Caught it by comparing `git ls-remote` against local `HEAD` while waiting on CI. The commit object was identical once pushed, so your verification stands unchanged, but the review window where the PR head and my local head disagreed was real and my fault.

Issue #369, opened 2026-08-28
[Bug] /model default is ambiguous: it persists the current model instead of switching to the configured default


bbqben:
## Brief Summary

`/model default` is ambiguous. Typing `/models` opens the model picker as expected, but completing `/model default` does **not** mean "switch to the default model" — it silently persists the *current* model as the boot default. Two different intentions share one spelling, and the destructive-ish one (a config write) is the one that wins.

## Current behaviour

From `_cmd_model` (`local_operator/tui/app.py`):

```python
lowered = arg.lower()
persist_default = lowered == "default" or lowered.startswith("default ")
target = arg[len("default ") :].strip() if persist_default else arg
...
if persist_default and not target:
    # Bare ``/model default`` means "keep the one I am on".
    target = session.model_label
```

So:

| typed | what happens |
|---|---|
| `/model` / `/models` | opens the picker |
| `/model <p>/<id>` | switches this session only |
| `/model default <p>/<id>` | switches **and** writes config |
| `/model default` | writes the **current** model to config |

The last row is the problem. `default` reads as a *value* in the same slot where `<provider>/<id>` is a value — the grammar is `/model <thing>`, and `default` looks like a thing you can be switched to. It is in fact a modifier applied to an elided argument, and the elision resolves to whatever you happen to be on.

## Reproduction

1. Launch with config default `A`.
2. `/model <provider>/B` — switches the session to B.
3. `/model default` — expecting "put me back on the configured default A".
4. Instead: B is written to `~/.local-operator/config.yml` as the new boot default. A is gone, and there is no undo.

The receipt does name the file and key, but it arrives after the write.

## Why this is worth changing rather than documenting

- The word is load-bearing across commands. `/approvals default auto` uses the identical modifier grammar, and its docstring states the design intent explicitly: *"a user learns 'put `default` in front of the value to keep it' once and it holds on the second command they meet it on."* That generalisation is exactly what makes `/model default` misfire — `/approvals` has a small closed set of values (`ask`/`auto`) that `default` cannot be confused for, while `/model`'s value space is open, so `default` reads as a member of it.
- There is no way to express "switch me to the configured default", which is the reading most users reach for first.
- `PERSIST_HINT = "/model default saves this for new sessions"` is threaded through the bare-`/model` notice, the switch receipt and the help table, so the phrasing is repeated in three places and would all be corrected by one change.

## Suggested directions

Not prescribing an implementation — options in rough order of disruption:

[...]

PR #470, opened 2026-08-30
feat(tui): one editing contract for /settings — enter opens, enter accepts, esc cancels

damianvtran:
## Summary

**Nothing on this page changes your configuration until you press `enter` on the thing you want. Everything else — moving, opening, looking, backing out — is free.**

`/settings` had **three different editing contracts on one page** and six gestures that wrote to `config.yml` with no explicit accept: a bool wrote on `enter`, an enum was safe until you picked, and a typed value committed when you **arrowed away**. The operator's report was that the page is "too trigger happy" and that they want to "explore available/supported values without actually impacting their setup live by accident".

This implements the specification in `~/local-operator-worktrees/settings-edit-model.md` (§7's plan, §7.2's state machine, §7.3's test plan), written by a UX auditor who drove the real app and measured every claim against `config.yml`'s **bytes**.

Ships as a **patch** (0.44.0 → 0.44.1): this changes how an existing surface behaves, it does not add a surface or a subsystem. Per AGENTS.md "Versioning", the test is whether the step-function capability can be named in the release title — here the title names a fix to a page that already shipped.

Implements #440.

## The contract, walked on the real app

`scripts/settings_contract_walk.py` (added) drives the real `OperatorApp` through its own slash command and key bindings against a scratch config dir, printing a SHA of `config.yml`'s bytes after each gesture. Byte comparison rather than parsed values is this page's standard of proof (#387 round 1, U1): a rewrite that moves only `last_modified` is still the page touching a file nobody asked it to touch.

**Before — `origin/main` at `b8cd5336`:**

```
1. page opened                       config: NO FILE
2. enter on a BOOL                   config: e184cb5dead7 | expanded: None
3. browsed onto 'off'                config: e184cb5dead7
4. esc                               config: e184cb5dead7
5. typed 77 into the editor          config: e184cb5dead7 | buffer:
6. ARROW AWAY (used to COMMIT)       config: e184cb5dead7 | notice:
7. browsed themes                    config: e184cb5dead7 | live theme: dark
8. esc                               config: e184cb5dead7 | live theme: dark
9. r on a DEFAULT row                config: e184cb5dead7 (used to CREATE the file)
10. enter ACCEPTS                    config: e184cb5dead7 | stored: 10
11. detail now: Fast budget against a reachable provider (5xx, timeout).   retry.maxRetries
12. r offered: False
13. r RESTORES                       maxRetries in file: True
```

The file appears at **step 2** — one `enter` on a bool, the gesture a user presses to find out what a row does. Every later step reads against a config they never asked to create, and step 5's empty buffer shows the editor was never reached because the row had already been written.

**After — this branch:**

```
1. page opened                       config: NO FILE
2. enter on a BOOL                   config: NO FILE | expanded: retry.enabled
3. browsed onto 'off'                config: NO FILE
4. esc                               config: NO FILE
5. typed 77 into the editor          config: NO FILE | buffer: 77
6. ARROW AWAY (used to COMMIT)       config: NO FILE | notice: discarded — enter saves, esc or moving away cancels
7. browsed themes                    config: NO FILE | live theme: light
8. esc                               config: NO FILE | live theme: dark
9. r on a DEFAULT row                config: NO FILE (used to CREATE the file)
10. enter ACCEPTS                    config: 4c519fd175f4 | stored: 7
11. detail now: Fast budget against a reachable provider (5xx, timeout). · default: 10   retry.maxRetries
12. r offered: True
13. r RESTORES                       maxRetries in file: False
```

The file first exists at **step 10**, the one gesture that is an accept. Step 7 also shows the theme previewing live and step 8 shows it reverting.

## Every gesture, before and after

| Gesture | Row kind | Before (`origin/main`) | After |
|---|---|---|---|
| `enter` | bool | **WRITE** — toggles and stores | opens a two-choice list, writes nothing |
| `space` | bool | **WRITE** | **WRITE** — retained accelerator, now reveals the row first |
| `enter` | enum / typed | opens (correct) | opens (unchanged) |
| `up` / `down` | open editor | **WRITE** — committed the buffer | cancels, writes nothing |
| `ctrl+n` / `ctrl+p` | open editor | **WRITE** | cancels, writes nothing |
| `pageup` / `pagedown` | open editor | inert (caret keys) | cancels and moves |
| `home` / `end` | open editor | caret within the buffer | caret within the buffer (unchanged) |
| click another row | open editor | **WRITE** | cancels, selects the clicked row |
| wheel | open editor | discarded (contradicted the arrows) | editor survives; the viewport moves (scroll model) |
| `escape` | open editor | discards (correct) | discards (unchanged) |
| `enter` | choice | **WRITE** — the one intended accept | **WRITE** — still the only accept |
| arrow past the group's end | choice | wrapped inside the group | leaves the group, collapses, reverts preview |
| `r` | row at its default | **WRITE** — created a config file | not offered; inert if pressed |
| `r` | row off its default | WRITE (reset) | WRITE (reset), and the detail line names the default |
| browsing | `tui.theme` choices | no preview at all | previews live, reverts on every cancel route |

## The anchor test

`test_no_gesture_but_enter_ever_writes` is the executable form of the contract: for every row kind, in a **fresh app per gesture**, it opens the row, puts the open state somewhere a commit would be visible (a buffer differing from the stored value, or a choice cursor off the stored member), presses one non-`enter` key, and asserts the bytes are unchanged. Parametrised 8 rows × 12 gestures = 96 cases.

Proven failing first, on `origin/main` in a throwaway worktree with **only the test file copied in** — output kept at `docs/evidence/settings-edit-model/anchor-before.txt`:

```
34 failed, 62 passed, 398 deselected
```

Twelve distinct gestures write: `up`, `down`, `ctrl+n`, `ctrl+p`, `pageup`, `pagedown`, `home`, `end`, `tab`, `left`, `right`, `escape`.

On this branch:

```
96 passed, 400 deselected in 60.96s
```

`tab`, `left`, `right` and `escape` failing on a bool row is worth naming: none of those keys does anything to a bool, but on `origin/main` the *opening* `enter` had already written, so the gesture under test was measured against a file that should not have existed. That is the audit's point restated by the harness.

## The most important change

`_leave_row()` becomes `_settle_row()` and **stops committing**. It cancels the editor, clears the notice, disarms the confirm, and **always returns True**, because nothing it does can fail. Every mover routes through it — `action_move`, `action_jump`, `action_section`, `_select_after`, `on_click`, `_enter_cascade` — which resolves the arrows-vs-wheel contradiction by construction rather than by documentation.

Returning unconditionally True deliberately deletes the "an invalid buffer holds the cursor" behaviour. That existed only because leaving was a *write* and a write can be refused; a cancel cannot fail, so trapping the cursor on a row would be friction with nothing behind it.

**This reverses U3 from #387, and the reversal is argued rather than assumed.** U3 found that movement discarded a valid buffer silently and chose commit-on-move because, on a page where writes were immediate anyway, silently *saving* is the less-bad silent outcome — a save is undoable with `r`, a discard is not. Round 2's U14 then taught the rule in the footer. That chain of reasoning is correct **under its premise**, and the operator is changing the premise: under explicit accept, a discard is not a lost save but the absence of an action the user never took, and commit-on-move is the one rule that makes "open a setting to see what it is" impossible.

`test_an_arrow_during_an_edit_does_not_silently_discard_it` is therefore **replaced, not deleted**, by `test_an_arrow_during_an_edit_discards_and_writes_nothing`, whose docstring names U3 and the premise change. A comment stands where the old test was, pointing at the new one — an apparently-vanished regression test is indistinguishable from a fix someone quietly undid.

## Bools

`enter` opens a two-choice list marking the stored value `●` and the shipped one `(default)`; **`space` is retained as the in-place accelerator**. Both, per the operator's decision.

The expansion is not empty ceremony: it carries information the collapsed row never did (which value shipped), and the cursor opens **on the stored choice**, so flipping a bool is `enter`, `down`, `enter` — three keys in one finger position.

`space` is now the page's *only* remaining one-keystroke write, which is why it takes the same `_reveal_cursor()` interlock `enter` and `r` already had. That is #440's second comment: the wheel can park the cursor off screen, and the accelerator would otherwise reintroduce exactly the off-screen-write hazard the contract closes. `test_activation_reveals_an_offscreen_cursor_before_acting` wheels the cursor away, presses both keys, and asserts nothing is written until the row is visible.

## Theme preview

Added for `tui.theme` only. The spec **corrected a premise here**: there is no live preview today — measured, `current_theme()` and the theme epoch do not move while browsing, and the before-frame's geometry line reports `previewing=dark restored=dark`. So this is preview being *added*, safely, because the new model finally has a well-defined cancel.

`SettingsPreview` is deliberately a **different message** from `SettingsChanged`. That one means "the file changed, bring the live surfaces into line"; this one means "paint as if, and expect to be told to put it back". One handler serving both is how a preview eventually becomes permanent.

The revert is the load-bearing half and holds on **every** exit route, each tested separately: `esc`, an arrow out of the group, a click elsewhere, leaving the page (including the app-side teardown that a session swap or `/clear` reaches, which the page's own `esc` never sees), and a **refused write** — a commit that did not land must not keep the paint it previewed.

**One deviation from the spec, stated plainly.** §2.3's table lists the wheel among the routes that revert. It no longer should: that table was written against the pre-#447 page where the wheel moved the *cursor*. Under the landed scroll model the wheel moves the viewport and the cursor stays on its choice, so nothing has left the group — reverting there would snap the theme back because the user glanced down the page. The spec anticipated this (§7.5: "rebase onto it rather than the other way round"). Recorded in `_scroll_rows`' docstring and in the test's own reasoning.

## `r`

Offered **only when the row is off-default**, gated on `settings_io.is_default` — which was already exported and used nowhere in the view. The same predicate now drives the ink, the footer hint, the detail line's `default: <value>` clause and `action_reset`'s guard, so the four cannot disagree.

This closes a finding the audit turned up along the way: `action_reset` had no default-state guard, so on a machine with **no config file at all**, opening the page, landing on any row and pressing the key the footer advertises **created a 1005-byte `config.yml`** — with nothing on screen saying so, because the row showed its default before and after.

## Rendered frames

Captured with `scripts/settings_shot.py` (four new states: `bool-open`, `choosing`, `preview`, `reset-offered`) at **100x30 and 80x24**, rasterised and **looked at**, not just diffed. Committed under `docs/evidence/settings-edit-model/{before,after}/`.

**`bool-open` — `enter` on `Retry failed calls`:**

Before, the value column has already flipped to `off` and `config.yml` is written. After:

```
  Retry failed calls              on ▾
    › ● on   (default)
      ○ off
```

**`choosing` — the footer rewrites per state:**

| | Before | After |
|---|---|---|
| footer | `↑↓ move · enter change · r default · ←→ panes · esc back to conversation` | `↑↓ move · enter choose · ←→ panes · esc cancel` |
| detail | help only | `<help> · nothing is saved until you press enter` |

`esc` cancels the expansion in that state, so advertising `back to conversation` was the same footer-vs-detail disagreement design round 2's D7 found. `r` is shed there too, because `action_reset` refused on a choice row — **a defect found by looking at the frame, not by a test**, and fixed in `410e6605`.

**`reset-offered` — the `r` gating, as a contrast:**

| Row | Detail line | Footer |
|---|---|---|
| `Max retries` (off default, 4) | `Fast budget against a reachable provider (5xx, timeout). · default: 10` | `… enter change · r default · …` |
| `Base delay (ms)` (at default) | `First backoff step; later attempts grow from it.` | `… enter change · ←→ panes · …` |

**`preview` — browsed onto `monokai`, then `esc`:** the whole app repaints to the previewed ramp and snaps back on cancel. The geometry line reports `previewing=monokai restored=dark` (before: `previewing=dark restored=dark`, i.e. nothing happened).

Looking at the frames also caught the detail clause being **clipped mid-sentence** at 100 columns. Half of "nothing is saved until you press enter" reads as a rendering fault rather than an abbreviation, so the clause is now measured against the row and shed **whole**, before the help — the reasoning D4 used for the label budget. `test_the_detail_clause_sheds_whole_rather_than_clipping` covers three widths.

## Invariants

All ten of §6 re-checked. Two needed work:

- **Invariant 1 (layout stability).** `test_layout_stability_across_choosing_and_preview` pins the detail row's y, the footer's y and the body's height across browsing, choosing, browsed, off-default, previewing and reverted. Asserted on **widget geometry** rather than the row count, since expanding a group legitimately lengthens the scrolled list; what must not move is the chrome the reader is anchored to.
- **Invariant 8 (arrows exit a choice group rather than wrapping).** Specified to land *after* the scroll fix, which it now has. An arrow past the group's end leaves it, collapses the expansion and reverts the preview; `_settle_expansion` re-anchors the cursor by identity afterwards, because closing drops the choice rows and the raw index would otherwise land 35 rows past where the arrow pointed on `tui.theme`.

Invariant 4 (the `display.*` flat-key trap) is re-proved on the **new** route by `test_display_keys_stay_flat_through_the_choice_path`, since bools now reach the writer through the choice commit.

U13's identity-anchored cursor moves with the commit: only `enter` restructures the list now, so `_anchored_commit` wraps the structural commits. `test_a_commit_that_inserts_rows_does_not_strand_the_cursor` is updated to press the explicit accept, with a docstring recording why.

## Migration

§7.4's one-time notice: the first time a session discards a modified buffer to a movement key, the detail line shows `discarded — enter saves, esc or moving away cancels` in the **informational** ink, not the danger ink (U16's distinction). Once per session — a message that fires every time becomes noise on the row that also carries validation errors.

## Tests

Nine test groups from §7.3, each **proven failing at `origin/main`** in a throwaway worktree with only the tests copied in: **47 failures there**, all green here.

```
tests/unit/tui/test_settings_view.py .... 496 passed in 256.99s
```

Existing tests that encoded the old contract were updated with their reasoning rather than deleted: U3 (replaced by its inverse), U14 (inverted — the footer still teaches the arrow keys, with the opposite meaning), U21 (the grown label is now `move · cancels`; the finding is about a label that grows, whatever it grows into), U2 and U1 (routed through `space`, the gesture that still writes on one keystroke).

## Gates

```
.venv/bin/python -m flake8 .                          clean
uvx --from black==26.1.0 black --check .              659 files unchanged
uvx isort==5.13.2 --check .                           clean
.venv/bin/python -m pyright --pythonpath ... .        0 errors, 0 warnings
```

pyright caught one real defect: the bool accelerator was originally `action_toggle`, which **overrides `DOMNode.action_toggle(attribute_name)`** — a Textual framework action — with an incompatible signature. Renamed `action_toggle_bool`.

## Full suite

```
tests/unit  ->  1 failed, 9422 passed, 15 skipped in 626.51s
```

The single failure is `test_steering_approval.py::test_a_held_answer_key_resolves_cleanly_on_every_second_key`, and it is **pre-existing flake, not a regression** — checked rather than assumed. Run alone with `-n0` against a **pristine `origin/main` worktree** (`f82c2f5e`, nothing of this branch present, `git clean`ed):

```
1 failed / 1 passed / 1 passed / 1 failed / 1 passed / 1 passed / 1 passed / 1 passed
```

Two failures in eight on a tree this branch has not touched. It is a race on `asyncio.wait_for(pending, 2)` in a test that drives a real approval gate through timing-sensitive key holds. Not fixed here — it is outside this slice, and an unrequested fix is one a reviewer has to review without having asked for it. Worth its own ticket.

## Not done, deliberately

- **§5's hop-delete confirmation.** The spec recommends arming `d` on a hop the way a chain delete arms, but calls it a judgement call it "would not block on". Left out to keep this slice to the editing contract; the destructive-action surface is unchanged by this PR.
- **Preview beyond `tui.theme`.** §3 explicitly recommends against extending to `display.*` in v1: each target is another revert path that has to be right on every exit. Earned incrementally.
- **The wheel as a preview-revert route**, for the reason given above — it stopped being a "leave the row" gesture when the scroll model landed.


damianvtran:
### Design review — round 1

Scope: b8cd5336118e9f300b5a322d756f886bc0f2a66f..e46da03d01d7b2c3c22fcc0578d78c0645fd28a6

Reviewed from **rendered frames**, not source. Committed evidence under `docs/evidence/settings-edit-model/{before,after}/` was rasterised and looked at; the same states were recaptured from this worktree with `scripts/settings_shot.py` and compared. Extra states the committed set does not photograph (open editor / error, off-default bool expansion, one-time discard notice, pre-preview resting theme) were captured the same way and looked at.

**Frames inspected (looked at as pixels, not SVG markup):**

| Frame | Size | What it shows |
|---|---|---|
| `after/bool-open-*.svg` + `.open.svg` | 100×30, 80×24 | collapsed bool vs two-choice expansion |
| `after/choosing-*.svg` | 100×30, 80×24 | cursor on the unstored choice |
| `after/preview-*.svg` + `.reverted.svg` | 100×30, 80×24 | theme preview, then the claimed revert |
| `after/reset-offered-*.svg` + `.default.svg` | 100×30, 80×24 | off-default vs default `r` gating |
| `before/preview-100x30.reverted.svg` | 100×30 | true dark, for the revert comparison |
| recapture `theme` (pre-preview rest) | 100×30 | byte-compare target for the revert |
| recapture `error` | 100×30, 80×24 | open editor, rejected value |
| recapture off-default `display.shimmer` expansion | 100×30, 80×24 | bool whose stored value ≠ shipped default |
| recapture discard notice | 100×30 | one-time migration copy on the detail line |

Geometry check (detail-line y across overview / bool-open / choosing / preview / reset-offered / error / discard): **508.0 at 100×30, 361.6 at 80×24, one value per width.** Footer y likewise pinned (556.8 / 410.4). Layout stability holds.

---

### D1 — BLOCKER — Cancelling a theme preview leaves the page wearing the previewed palette

The load-bearing half of preview is the revert. Spec §3: the reverted frame must be the pre-preview frame; "a preview that leaks leaves the app wearing a theme its config file disagrees with, and nothing on screen would ever say so."

**What the committed `after/preview-100x30.reverted.svg` actually paints** (and the 80×24 pair, same defect):

- Value column: `Theme  dark ▸` — the *model* restored.
- Headings: `#f8f8f2` (Monokai foreground). Dark's headings are `#e9e5db`.
- Cursor: `#a6e22e` (Monokai lime). Dark's cursor is `#38c96a`.
- Labels: `#c8c8c0`. Dark's labels are `#b5afa2`.
- Geometry line on the same capture: `previewing=monokai restored=dark`.

The geometry line is true of `theme_mod.current_theme()` and false of the pixels. Recaptured independently on this HEAD:

```
PRE        theme=dark     monokai-ink: []
PREVIEWING theme=monokai  (warm charcoal, lime cursor — preview working)
AFTER esc  theme=dark     monokai-ink: ['#5a5b55', '#90918b', '#a6e22e', '#c8c8c0', '#f8f8f2']
```

The leak is still there after 3 s of idle. It is not a settle lag. It clears only if the user arrows onto another row (which forces a repaint of the settings list through a different path). `esc` is the route the footer advertises.

Looked at as a first meeting of the page: the cancelled frame is a Monokai page whose Theme row still says `dark`. That is the user thinking they committed. The before-frame of the same state (`before/preview-100x30.reverted.svg`) is genuine dark — cream headings, muted-green cursor — so the regression is this PR's, not a pre-existing paint.

This is the #387 pattern the brief flagged: a section-specific assertion (here, `restored=dark` on the model) is green while the surface displays something false. The committed "reverted" stills *are* the bug.

**Fix the paint, then recapture the pair.** The reverted still has to be palette-identical to the pre-preview still, not merely model-identical. Do not assert on `current_theme()` alone.

---

### D2 — MAJOR — Open-editor footer is ungrammatical and lies about `esc`

Painted, 100×30 error state (rejected `9999` in `retry.maxRetries`):

```
↑↓  move · cancels · enter  save · r  default · ←→  panes · esc back to conversation
```

Two failures on one line, both new in 883ae9c3 (`move = (self._move_hint, " move · cancels", False)`):

1. **`move · cancels` does not parse.** U14's inverted clause was glued onto the move hint as a trailing fragment. It reads as "move cancels" with no subject. The intended teaching — arrows now discard — is lost in the grammar. `"arrows cancel"` or `"move  cancel"` would say it; this does not.
2. **`esc back to conversation` is false in this state.** Pressing `esc` cancels the editor and stays on the page (verified: `editing` goes `retry.maxRetries → None`, the `SettingsView` is still mounted). The exit-label rewrite to `cancel` only fires for `_confirm_delete` or `_expanded`, not for `_editing`. The inline editor chrome on the same row paints `enter saves · esc …` (clipped), so the row and the footer disagree about the same key — the D7 anti-pattern this page has already paid for.

Same copy at 80×24, unclipped, so this is not a shed. The footer is where a previous round was fooled by asserting on `rendered_hints()` rather than the painted frame; this time the *painted* frame is the defect.

---

### D3 — NIT — `default:` clause is the dimmest ink on the page

Off-default vs default is otherwise legible. The value column does the real work: `4` on Max retries is `#e9e5db` (14.97:1), `500` on Base delay is `#837c6d` (4.55:1). Side-by-side, you can scan the list and see which settings you have changed. Footer `r default` appears and disappears with that same predicate; at 80×24 the `default: 10` clause is shed *whole* rather than clipped mid-sentence (410e6605, correct).

The clause that *names* the number `r` would restore (`· default: 10`) is `#4a4539` on `#14110c` — **1.97:1**. I read it because I knew to look. Informational ink is the right *kind*; this particular clause is the one piece of new information the gating adds, and it is painted as help-text noise. Follow-up, not a round-blocker: one step up the existing ramp (`#837c6d`, already used for keycaps and the key path) would make the number findable without inventing a colour.

---

### What is fine (so it is not re-lit next round)

**The bool expansion earns itself.** It is the same object as an enum expansion: `▾` on the parent, `●`/`○` on the members, `(default)` on the shipped value, cursor opening on the stored choice, footer `enter  choose · esc cancel`. On a defaulted bool (`retry.enabled = on`) it is a small ceremony; on an off-default bool (`display.shimmer = off`) it is the only place the page shows `○ on  (default)` next to `● off`, which is information the collapsed row never carried. It does not read as a hoop. Spec §7.6 Medium risk, visually discharged.

**`r` gating is glance-legible as a pair.** Off-default: brighter value, `· default: N` on the detail line (at 100×30), footer offers `r`. Default: dimmer value, no clause, no `r`. The 80×24 pair sheds the clause and keeps the footer contrast. Pane honesty undisturbed: at 100×30, `providers` is populated, `teams`/`agents (←→)` present, `read-only` last; at 80×24 the pane is gone entirely rather than a heading over nothing. Scrollbar palette unchanged.

**Layout does not shift.** Detail line and footer sit at one y per width across every state this round photographs, including the one-time discard notice (`discarded — enter saves, esc or moving away cancels` at y=508.0) and the error (`must be at most 100` at the same y, danger ink).

**80×24 footers of the new states do not clip.** Choosing/bool-open: `↑↓  move · enter  choose · esc cancel`. Reset-offered: `↑↓  move · enter  change · r  default · esc back to conversation`. Default row: `r` dropped whole. No state in the committed set advertises a key it does not have. (D2 is the exception, and it is the editor state, which the committed set does not include.)

**Previewing-while-browsing is distinguishable from accepted** — footer `esc cancel`, `●` remains on the stored theme, cursor on a different `○`. The failure is the frame *after* cancel (D1), not the browsing frame.

---

**Verdict: not TERMINAL.** D1 (BLOCKER) and D2 (MAJOR) remain. Recapture the preview pair against a pre-preview still after the paint is actually restored; recapture the open-editor footer after the copy and the `esc` label agree with the keys.

damianvtran:
### UX review — round 1

Reviewer: ux-reviewer (anthropic/claude-opus-5)
Scope: f82c2f5e9f09fea6518506abb9a5414b76c31f70..e46da03d01d7b2c3c22fcc0578d78c0645fd28a6

Drove the real `OperatorApp` in `~/local-operator-worktrees/edit-model` via `run_test` against a scratch `LOCAL_OPERATOR_CONFIG_DIR`, sending real key and mouse events and hashing `config.yml` bytes after every gesture. Compared against `~/local-operator-worktrees/settings-edit-model.md` §2.3 / §7.2 / §7.4 / §6.

**Verdict: not TERMINAL.** One BLOCKER. The page's contract holds for the gestures the anchor test covers and for every cancel route that goes through `action_move`, `on_click`, or `esc`. It does not hold for the jump/section keys, which leave CHOOSING open as a ghost — and with it, a live theme the file never accepted.

---

#### U1 — BLOCKER — `home` / `end` / `pageup` / `pagedown` leave CHOOSING open; a theme preview leaks and `esc` stops meaning leave

**Flow.** `/settings` → `tui.theme` → `enter` → `down` (live theme becomes `light`, file still absent) → `end`.

**Expected (spec §2.3 CHOOSING, “cursor leaves the group (arrow past the ends, page keys, …)”):** collapse, revert preview, land on the last row. File untouched. Next `esc` leaves the page.

**What happened.**

| after `end` | observed |
|---|---|
| `config.yml` | `NO FILE` (bytes honest) |
| live theme | still `light` |
| `_expanded` | still `tui.theme` |
| cursor | `max_learnings_history` (a read-only row, fifty rows away) |
| footer | `↑↓  move · ←→  panes · esc cancel` |
| painted list | 54 choice rows still in `_rows` (117 rows, same as while choosing) |

The user is looking at the last row of the page. The app is wearing a theme they did not accept. The footer says `esc cancel`, which is a lie about this row — `esc` does not leave, it teleports the cursor back to `tui.theme` and *then* reverts the paint. A second `esc` finally leaves.

Same ghost from `home`, `pageup`, and `pagedown`. `action_move` calls `_settle_expansion()` after landing; `action_jump` and `action_section` call `_settle_row()` only. `_settle_row` deliberately does not close an expansion (so arrows can browse inside one). The jump/section keys never got the counterpart.

Consequences walked on the same ghost:

- `end` then `enter` on the retired row: the refusal (`this setting is retired and cannot be changed`) paints, `_expanded` stays `tui.theme`, theme stays `light`.
- `home` then `enter` on `hosting`: **two modes at once** — `editing=hosting` and `expanded=tui.theme`, live theme still `light`, footer `↑↓  move · cancels · enter  save · r  default · ←→  panes · esc cancel`. The editor row shows a caret; `esc` is still owned by the invisible expansion.
- Arrowing *within* the group, `esc`, click-elsewhere, and leaving the page all revert correctly. The wheel surviving a preview is the documented §7.5 departure and is fine — the cursor has not left.

This is the High row in spec §7.6 (“Preview leaves the app in a theme the file disagrees with”) firing on a key every settings page binds. It is also the class this review was told to watch for: the 96-case anchor test parametrises `home`/`end`/`pageup`/`pagedown`, but it presses them from *inside* the group (one `down` off the stored choice), so `_settle_expansion` never has a destination outside it to notice. Section-specific tests pass; the surface is wrong.

`action_jump` / `action_section` need the same post-landing `_settle_expansion()` + `_preview_choice()` pair `action_move` and `on_click` already have. Until they do, “nothing changes until you press enter” is true of the file and false of the running app.

---

#### U2 — MINOR — the CHOOSING reassurance clause is invisible at both widths the spec requires measuring

**Flow.** Open every Shape-A row (`enter` on each of the 24 bool/enum settings) at 80×24, 100×30, 120×30, 160×40. Look for `nothing is saved until you press enter` on the detail line.

**Expected (spec §4.5 / §7.3.8).** The sentence is on the detail line while CHOOSING; it sheds whole when it will not fit. At 100×30 and 80×24 it should be *possible* to see it on typical rows, because it is “the sentence the redesign exists to make true.”

**What happened.** 0/24 rows at 80×24 (`detail_width=74`). 0/24 at 100×30 (`detail_width=94`). 11/24 at 120×30. 24/24 only at 160×40.

Shedding is implemented correctly (drop the clause whole, keep the help). The clause as written simply never fits next to help + key at the page’s documented sizes. The footer *does* rewrite (`enter choose · esc cancel`), and the expansion itself (`●` / `○` / `(default)` / `▾`) carries a lot of the “you are choosing, nothing is stored” signal — so this is not the same class as U1. It does mean a user at 100×30 never reads the sentence the spec hung the trust argument on.

Shorten it (`not saved until enter`) or give it a carrier that survives 100×30 (the footer already owns `enter choose`). Do not keep a clause whose only successful size is 160 columns.

---

#### U3 — MINOR — `r` is offered without naming what it restores, at the widths users actually use

**Flow.** Flip each Shape-A row off its default (`space` on bools, `enter`/`down`/`enter` on enums). Read the detail line and the footer.

**Expected (spec §4.4).** `r` has no confirm, so the detail line names `default: <value>` on an off-default row. Same predicate as the hint, so the two cannot disagree.

**What happened.** The predicate agrees — `r` is lit exactly when `is_default` is false, and pressing it restores. The *name* of the default is shed. At 80×24, 21/24 off-default Shape-A rows light `r` with no `default:` clause. At 100×30, 11/24 (the long-help ones: `retry.modelFallback`, `web_fetch.allow_private`, `fork.cmux_placement`, …). `retry.enabled` at 100×30 *does* show `· default: on`, which is why a short-row walk looks fine.

Same shedding-vs-job fight as U2, with a sharper cost: `r` is an immediate write. A user who can see the hint and cannot see the payload is being asked to trust a key the page just made a point of not confirming.

---

#### U4 — MINOR — while EDITING, the footer still advertises `esc back to conversation`

**Flow.** `retry.maxRetries` → `enter` (editor open, buffer seeded).

**Expected (spec §4.3: “the state that owns `esc` owns its label”; the D7 fix applied to CHOOSING and CONFIRMING).** First `esc` cancels the editor. The footer should say so.

**What happened.** Footer: `↑↓  move · cancels · enter  save · r  default · ←→  panes · esc back to conversation`. The row contract (`enter saves · esc cancels`) tells the truth; the footer tells the other meaning. First `esc` cancels, second leaves — the ladder is correct, the advertisement is not. This is D7’s disagreement, one state over. CHOOSING got the fix; EDITING did not.

---

#### U5 — MINOR — `r default` is lit during EDITING on a default row; the key types `r` into the buffer

**Flow.** `retry.maxRetries` (at its shipped default) → `enter` → type `77` → `r`.

**Expected.** A lit hint whose key does nothing is the page’s own anti-pattern. During an open editor, printables belong to the buffer, so `r` must not be advertised.

**What happened.** Footer shows `r  default` (`_reset_applies` returns `True` whenever `_editing is not None`, copying the in-progress-owns-the-footer rule). Pressing `r` inserts the letter: buffer becomes `77r`, file untouched. The hint is not inert — it is a trap for anyone who believes the footer.

---

### What is actually good

The contract the operator asked for is real on every path that goes through `action_move`, `on_click`, `esc`, or `enter`:

- `enter` on a bool opens `● on  (default)` / `○ off` and writes nothing. `esc` collapses. File never appears.
- Typing `77` into `retry.maxRetries` and arrowing away discards; file absent; one-time notice `discarded — enter saves, esc or moving away cancels` in faint (`#4a4539`), not danger. Second discard in the same session is silent. Unmodified buffer does not fire it.
- `space` on a bool writes, reveals an off-screen cursor first (issue #440’s second comment), and is not advertised. On every other kind it is `enter`. Inside an editor it is a space character. On a choice it accepts, per §2.3.
- Theme preview: `esc`, arrow out of the group, click elsewhere, re-press `enter` on the owner, leave-the-page, and `_close_settings_view` (the teardown the page never sees) all restore `dark` with the file still absent. Accepting persists across leave. Wheel-survives-preview is the argued §7.5 departure and behaves as documented.
- `r` on a default row is unadvertised and inert (file never created). `r` on an off-default row restores. `r` during CHOOSING on an off-default bool resets the owner and collapses.
- Cascade `enter` opens the chain list, never a text editor. Untouched editor `enter`/`enter` is a byte-level no-op. Identity-anchored cursor after a bool/int write. Caret `home`/`end`/`left`/`right` stay in the buffer. `pageup`/`pagedown` during an edit cancel-then-move. Invalid buffer no longer traps the cursor. Wheel during an edit survives and writes nothing. Activation reveals its row. Resize mid-edit keeps the buffer. Read-only `enter`/`space` refuse and write nothing. `display.shimmer` stays a flat top-level key through the new choice-commit path. Click-elsewhere during an armed chain-delete disarms and writes nothing.
- U3 from #387 is replaced, not deleted: the pointer comment at the old site names U3 and §2.4, and `test_an_arrow_during_an_edit_discards_and_writes_nothing` argues the premise change. Honest.
- #369 `/model default` grammar, all six spellings, still behaves: bare `/model default` confirms and writes nothing; `/model default p/id` persists and switches; `/model saved` switches to the configured default; `/model saved` with none configured says so; `/model p/id` switches without persisting; picker `d` persists without switching.

Hop delete is still immediate. Spec §5 recommended arming it and explicitly would not block on it — not raised.

`enter` on the currently-stored choice of a default bool/enum *does* create `config.yml` (the whole default dump, `last_modified` and all). Spec §2.3 says that `enter` is a WRITE, so this is in contract; the editor’s untouched-seed no-op is the stricter cousin and does not extend to a choice accept. Noted, not raised.

---

### Flows exercised

1. Contract sweep: 7 row kinds × 15 non-enter gestures from an open state, fresh app each time (105). Only `space` on a choice wrote — specified accept.
2. Bool expand / browse `off` / `esc`; int edit / arrow-away / migration notice / second discard; read-only `enter` and `space`.
3. CHOOSING clause visibility at 80/100/120/160/200/240 columns, all 24 Shape-A rows.
4. `default:` clause visibility after driving each Shape-A row off default, same widths.
5. Theme preview cancel by `esc`, arrow out (up and down), re-press `enter` on owner, leave page, `pagedown`/`home`/`end`, click elsewhere, wheel, accept, `_close_settings_view`.
6. `space` on BOOL / flat-dotted BOOL / ENUM / INT / TEXT / LIST / CASCADE; `space` inside an editor; `space` on a choice; off-screen reveal; advertising.
7. Esc ladder (browsing / editing / choosing). Untouched-editor no-op. Identity after write. Caret keys on a long buffer. Page keys during edit. Cascade `enter`. Wheel during edit. Activation reveal. Invalid-buffer leave. `r` default/off-default. Migration notice ink and unmodified-buffer silence. Leave-page mid-edit. Resize mid-edit.
8. Click elsewhere during bool CHOOSING; select-then-activate on a choice; click elsewhere during an armed chain-delete.
9. `r` during CHOOSING (default and off-default); `r` during EDITING.
10. `display.shimmer` choice-commit path (flat key).
11. Ghost-state follow-through: `end` then `esc`; `home` then `enter` (two modes); `end` then `enter` on a read-only row.
12. #369 six spellings of `/model default`.

Round 1 is not clean. U1 has to move before this can ship: a user who opens Theme to look, then presses `end` — a key the footer itself teaches as movement — is wearing a theme they did not pick, and the key that used to mean “get me out of here” takes them back into the list they left.

damianvtran:
### Agent review — round 1

Scope: b8cd5336118e9f300b5a322d756f886bc0f2a66f..e46da03d01d7b2c3c22fcc0578d78c0645fd28a6

Drove the real `OperatorApp` under `run_test` against a scratch config dir. Gates on this tree: flake8, `black==26.1.0 --check .`, `isort==5.13.2 --check .`, pyright all clean. `tests/unit/tui/test_settings_view.py`: 496 passed in 510s.

The U3 reversal is honest: `test_an_arrow_during_an_edit_does_not_silently_discard_it` is replaced (not deleted) by `test_an_arrow_during_an_edit_discards_and_writes_nothing`, with a pointer comment at the old site naming U3 and spec §2.4. `r` gating uses `settings_io.is_default` for the ink, the footer, the detail clause and the action. The `display.*` flat-key path is re-proved through the choice commit. `_settle_row` does not write. Those are solid.

The state machine is not. Two movers never call `_settle_expansion`, and that is enough to leak a theme preview in a way the capture cannot recover from.

---

#### C1 — BLOCKER — `home` / `end` / `pageup` / `pagedown` leave a choice group open and leak the theme preview; accepting anything else then makes the leak permanent

`_settle_expansion` (and `_preview_choice`) are called from exactly two movers: `action_move` (`settings_view.py:528`) and `on_click` (`settings_view.py:1893`). They are **not** called from:

- `action_jump` (`settings_view.py:608-620`) — `home` / `end`
- `action_section` (`settings_view.py:585-606`) — `pageup` / `pagedown`

`_settle_row` is the wrong place to close an expansion (closing there would collapse the group on the arrow that browses inside it), so the split is right. The jump and section movers just never call the counterpart. Spec §7.2 named every mover that has to route through settle; these two do the editor half and skip the expansion half.

Reproduced on this branch, 120×32, scratch config, `tui.theme` opened and browsed off `dark` onto `light`:

| Gesture after browsing | `expanded_key` | `current_theme()` | `_preview_before` |
|---|---|---|---|
| `end` | `'tui.theme'` | `'light'` | `'dark'` |
| `home` | `'tui.theme'` | `'light'` | `'dark'` |
| `pagedown` | `'tui.theme'` | `'light'` | `'dark'` |
| `pageup` | `'tui.theme'` | `'light'` | `'dark'` |

`config.yml` is untouched — the anchor test is therefore green on these keys, which is why it did not catch this. The live app is wearing a theme the file disagrees with. That is the risk table's High row (§7.6), and it is the operator's original complaint restated: exploring a value space changed their setup.

It is not just a leaked paint. Three further consequences, all measured:

1. **The expansion stays in the list.** After `end` from `tui.theme`, 54 choice rows are still painted, the footer still says `esc cancel` (because `_expanded is not None`, `settings_view.py:3088`), and `esc` does **not** leave the page — it teleports the cursor back onto `tui.theme` and *then* reverts. A user who paged away thinking they had left the group is now somewhere else under a footer that lies about what `esc` does.

2. **Two states live at once.** Spec §7.2: no path should leave two states simultaneously live. After `end` from `retry.enabled` (expansion still open) then `enter` on the cascade then `enter` on `+ add a chain`: `expanded='retry.enabled'` **and** `editing='chainadd'`. An editor open on one row, a choice group open on another.

3. **The leak becomes unrecoverable.** `action_activate` on a successful choice commit does `self._preview_before = None` without restoring (`settings_view.py:803`) — correct when the commit *is* the previewed theme, fatal when the capture belongs to a group the cursor has already left. Sequence: browse `tui.theme` onto `light`, `end`, open `tool_approval_mode`, accept a choice. After that: `current_theme()=='light'`, `_preview_before is None`, stored theme still `dark`. `_close_settings_view` / `revert_preview_for_teardown` then have nothing to restore. The app sits in `light` until the next restart, and nothing on screen says so.

`test_a_theme_preview_reverts_on_every_exit_route` (`test_settings_view.py:3369`) is parametrised over `("move-off", "leave-page", "click-elsewhere")` and its docstring claims "every way out of the expansion". `home` / `end` / `pageup` / `pagedown` are in spec §2.3's CHOOSING table as leave-group routes and are not in that list. The test is satisfied; the surface is not. That is the #387 pattern this review was told to watch for.

Fix: call `_settle_expansion()` (and `_preview_choice()` if the landing row is a choice) after the cursor has landed in `action_jump` and `action_section`, the same two lines `action_move` and `on_click` already have. Extend `test_a_theme_preview_reverts_on_every_exit_route` with those four gestures; a test that names "every exit" and then lists three of them is how this shipped past the suite.

---

#### C2 — MAJOR — version `0.44.1` is already published; `origin/main` is `0.44.2`

`pyproject.toml` on this branch is `0.44.1`, bumped from the merge-base's `0.44.0`. Since then `origin/main` has shipped `0.44.1` (Gemini tool schemas) and `0.44.2`. PyPI's latest is `0.44.1`. Merging this as-is either downgrades main's version or tries to republish a tag that already exists.

Rebase onto current `origin/main` and bump to `0.44.3`. The bump type is still patch.

---

#### C3 — MINOR — `space` on the owning BOOL row while its expansion is open writes and does not collapse

`action_toggle_bool` (`settings_view.py:887`) keys off `row.kind == "setting"` and `setting.kind is Kind.BOOL`. After `enter` on `retry.enabled` the cursor sits on a choice; `up` lands back on the owner, which is still a BOOL setting row, so `space` writes immediately **and leaves `_expanded` set**. Measured: `wrote=True`, `expanded='retry.enabled'`, two choice rows still in the list, the `●` still on the value that is no longer stored.

Spec §2.5 keeps `space` as the resting-row accelerator. While CHOOSING, `space` is specified as accept-on-a-choice. The owner is neither resting nor a choice, so the accelerator fires in the middle of an open group and the paint lies about what is stored. Collapse (or refuse) when `_expanded == setting.key`.

---

#### C4 — MINOR — theme-preview revert reapplies with `preview=True`, so off-screen transcript blocks keep the previewed ramp

`on_settings_preview` (`app.py:10624`) always calls `_apply_theme(..., preview=True)`. `_revert_preview` posts `SettingsPreview`, so an `esc` cancel of a theme preview re-inks only on-screen blocks. Off-screen transcript blocks keep the previewed colours until a later full sweep. `revert_preview_for_teardown` already does the full apply; the in-page revert should too — a cancel is a settle, and `_apply_theme`'s own docstring says the full sweep runs when the preview settles.

---

#### C5 — NIT — `_reveal_cursor` still documents the old contract

`settings_view.py:716`: "This page writes IMMEDIATELY, with no undo beyond `r`". Under this PR that sentence is false, and it is the comment a later reader will trust when deciding whether a new key needs the reveal. Same class: `action_jump`'s "Re-derived after the commit" (`settings_view.py:613`) — movement no longer commits.

---

**Verdict: not clean.** C1 is stop-ship. C2 will break the release if merged as-is. Remaining minors/nits wait on a round 2 against a head that actually reverts the preview on every exit in §2.3's table.

damianvtran:

Implementer: coder (xai/grok-4.6)
Head: `17a35062`
Scope answered: `b8cd5336..e46da03d` findings, remediations on `a6f766e8` / `557fef0c` / `17a35062`

- **C1 (BLOCKER)** — fixed in `a6f766e8`. `_land()` is the post-landing funnel (`_settle_expansion` + `_preview_choice` + `_repaint`). `action_jump`, `action_section`, `action_move`, and `on_click` all route through it. `test_a_theme_preview_reverts_on_every_exit_route` now parametrises `end`/`home`/`pageup`/`pagedown` and asserts BOTH `current_theme()` restored AND `expanded_key is None`. Proven failing at `e46da03d` in `/tmp/lo-e46da03d` with only the test file copied in (`light` kept after `end`/`home`/`pageup`/`pagedown`).
- **C2 (MAJOR)** — fixed in `557fef0c`. Rebased onto `origin/main` (`0.44.2`); skipped the colliding `0.44.1` bump; bumped to `0.44.3`. `git show origin/main:pyproject.toml` is `0.44.2`; tags `v0.44.0`/`v0.44.1`/`v0.44.2` exist; next free patch is `0.44.3`.
- **C3 (MINOR)** — fixed in `a6f766e8`. `space` on the owning BOOL row while its group is open collapses and writes nothing. `test_space_on_an_open_bool_owner_collapses_and_writes_nothing` proven failing at `e46da03d` (`space` wrote).
- **C4 (MINOR)** — fixed in `a6f766e8`. `_revert_preview` calls `_apply_theme(restore)` directly (full sweep, `preview=False` default) instead of posting `SettingsPreview` which always applied `preview=True`. Same path `revert_preview_for_teardown` already used.
- **C5 (NIT)** — fixed in `a6f766e8`. `_reveal_cursor` no longer says the page "writes IMMEDIATELY". `action_jump`'s "Re-derived after the commit" comment now says the ends move because closing an expansion drops choice rows.

Gates on this tree: flake8, `black==26.1.0 --check .`, `isort==5.13.2 --check .`, pyright clean. `tests/unit/tui/test_settings_view.py`: 502 passed. Did not merge.

damianvtran:

Implementer: coder (xai/grok-4.6)
Head: `17a35062`
Scope answered: `b8cd5336..e46da03d` findings, remediations on `a6f766e8` / `17a35062`

Looked at recaptured frames as pixels (qlmanage PNG of the committed SVGs), not markup.

- **D1 (BLOCKER)** — fixed in `a6f766e8`, recaptured in `17a35062`. `_revert_preview` applies the restored theme *before* `_repaint`, so the list is rebuilt in dark's ramp. Recaptured `after/preview-{100x30,80x24}.reverted.svg`: headings `#e9e5db`, cursor `#38c96a`, no `#a6e22e` / `#f8f8f2`. Looked at the pair: previewing frame is Monokai (warm charcoal, lime cursor on `monokai`); reverted frame is genuine dark (cream headings, muted-green cursor, Theme `dark ▸`). Geometry line `previewing=monokai restored=dark` now matches the pixels. Test `test_previewing_a_theme_writes_nothing_and_esc_restores_it` browses onto Monokai and asserts those hexes absent after `esc`; proven failing at `e46da03d`.
- **D2 (MAJOR)** — fixed in `a6f766e8`, recaptured in `17a35062`. Open-editor footer is now `↑↓  move cancel · enter  save · ←→  panes · esc cancel` (100×30) / `↑↓  move cancel · enter  save · esc cancel` (80×24). Looked at `after/error-*.svg`: grammatical, `esc cancel` matches the row contract, no `r default`. Asserted on `_painted_frame`, not `rendered_hints()`. Proven failing at `e46da03d` (`move · cancels` and `esc back to conversation` both in the painted frame).
- **D3 (NIT)** — fixed in `a6f766e8`. `· default: 10` uses `dim` (`#837c6d`, already the key-path ink) rather than `faint` (`#4a4539`, 1.97:1). `test_the_default_clause_is_not_the_dimmest_ink_on_the_page` proven failing at `e46da03d`. Looked at `after/reset-offered-100x30.svg`: `default: 10` is findable next to the help.

Did not merge.

damianvtran:

Implementer: coder (xai/grok-4.6)
Head: `17a35062`
Scope answered: `f82c2f5e..e46da03d` findings, remediations on `a6f766e8` / `17a35062`

- **U1 (BLOCKER)** — fixed in `a6f766e8`. Same defect as C1: jump/section keys now settle the expansion after landing. After browsing Theme onto a preview and pressing `end`/`home`/`pageup`/`pagedown`: `current_theme()` restored, `_expanded is None`, file untouched, next `esc` leaves the page. Test asserts both the theme and the expansion; bytes-only would have stayed green. Proven failing at `e46da03d`.
- **U2 (MINOR)** — fixed in `a6f766e8`. CHOOSING clause shortened to `not saved until enter` and, when help+clause will not fit, the clause wins over help. Visible at 80×24 and 100×30 (looked at `after/choosing-*.svg` and the previewing frame). `test_the_detail_clause_sheds_whole_rather_than_clipping` now requires the clause at those widths; proven failing at `e46da03d`.
- **U3 (MINOR)** — fixed in `a6f766e8`. Same shedding change: `default: N` is kept in preference to help, so `r` names what it restores at 80×24 and 100×30. Looked at `after/reset-offered-100x30.svg`: `· default: 10` sits on the detail line with `r default` in the footer.
- **U4 (MINOR)** — fixed in `a6f766e8`. While EDITING, `esc` is labelled `cancel` (same as CHOOSING/CONFIRMING). Looked at `after/error-*.svg`. Same painted-frame assertion as D2.
- **U5 (MINOR)** — fixed in `a6f766e8`. `_reset_applies` returns False while `_editing is not None`, so `r default` is not advertised; printables stay in the buffer. Painted error footer has no `r`.

Did not merge.

damianvtran:
### Design review — round 2

Scope: e46da03d01d7b2c3c22fcc0578d78c0645fd28a6..17a350620f2438b797471fdebacad2af3734c3a0

Reviewed from **rendered frames**, not source. Round-1 D1/D2/D3 were re-checked on the committed `after/` stills (hex-identical to an independent recapture from this HEAD) and by looking at qlmanage PNGs of those recaptures. Extra states the committed set photographs (open editor, choosing, bool-open, reset-offered / default) were recaptured with `scripts/settings_shot.py` and looked at as pixels.

**Frames inspected (looked at as pixels, not SVG markup):**

| Frame | Size | What it shows |
|---|---|---|
| recapture `preview-*.svg` + `.reverted.svg` | 100×30, 80×24 | Monokai preview, then the claimed revert |
| committed `after/preview-*.reverted.svg` | 100×30, 80×24 | same pair, byte-palette identical to recapture |
| recapture `error-*.svg` | 100×30, 80×24 | open editor, rejected `9999` |
| recapture `choosing-*.svg` | 100×30, 80×24 | cursor on the unstored bool choice |
| recapture `bool-open-*.open.svg` | 100×30 | expansion resting on the stored choice |
| recapture `reset-offered-*.svg` + `.default.svg` | 100×30, 80×24 | off-default vs default `r` gating |

Geometry check (detail-line y across preview / reverted / choosing / bool-open / reset-offered / error): **508.0 at 100×30, 361.6 at 80×24, one value per width.** Footer y likewise pinned (556.8 / 410.4). Layout stability holds.

---

### Round-1 findings, re-checked on this head

**D1 (BLOCKER) — discharged.** The reverted pair is genuine dark, not a Monokai page whose Theme row says `dark`.

Looked at recapture `preview-100x30.reverted.svg` (and the 80×24 pair, same paint):

- Ground `#14110c`, headings `#e9e5db` (Appearance / Session / `settings`), cursor `#38c96a` on Theme.
- Value column: `Theme  dark ▸`.
- Monokai ink absent: `#a6e22e` / `#f8f8f2` / `#90918b` / `#272822` / `#c8c8c0` all count 0.
- Geometry line `previewing=monokai restored=dark` now matches the pixels.

The previewing half of the pair is still Monokai (ground `#272822`, `#f8f8f2` present) — so the snap-back is a revert, not a preview that never applied. Committed `after/preview-*.reverted.svg` hex-counts equal the recapture; this is not a stills-only claim.

**D2 (MAJOR) — discharged.** Open-editor footer, looked at as paint on `error-100x30.svg` / `error-80x24.svg`:

```
↑↓  move cancel · enter  save · ←→  panes · esc cancel     (100×30)
↑↓  move cancel · enter  save · esc cancel                 (80×24)
```

Parses as English. `esc cancel` matches the row contract (`enter saves · esc cancels` at 80×24; the 100×30 row still clips to `esc …`, which is the value-column budget, not a footer lie). No `r default`. No `esc back to conversation` while the editor owns the key.

**D3 (NIT) — discharged.** `· default: 10` is classed with the key-path ink `#837c6d` (dim), not faint `#4a4539`. Looked at `reset-offered-100x30.svg`: the number is findable next to the help.

---

### U2 / U3 clause, painted (not `rendered_hints()`)

`not saved until enter` is on the detail line at both widths. Looked at:

- `choosing-100x30.svg` / `choosing-80x24.svg` — clause + `retry.enabled`, help shed whole at both sizes.
- `preview-100x30.svg` / `preview-80x24.svg` — same clause while browsing Monokai.
- `bool-open-100x30.open.svg` — same clause with the cursor on the stored `● on`.

`default: 10` is on the detail line at both widths (`reset-offered-100x30.svg` keeps help + clause; `reset-offered-80x24.svg` sheds help and keeps `default: 10`). Footer `r default` is present on the off-default frame and absent on the default frame (`reset-offered-100x30.default.svg`: `↑↓  move · enter  change · ←→  panes · esc back to conversation`).

---

### New states from the remediation — no dead keys, no new paint defect

The round-1 remediations (`_land`, `_revert_preview` applying before `_repaint`, editor footer rewrite, clause-wins shedding, `_reset_applies` false while editing) did not introduce a new advertised-but-inert key:

| State | Painted footer | Dead key? |
|---|---|---|
| EDITING (error) | `move cancel · enter save · … · esc cancel` | no `r` |
| CHOOSING (bool / theme) | `move · enter choose · … · esc cancel` | no `r`, no `space` |
| off-default rest | `move · enter change · r default · … · esc back to conversation` | `r` is live |
| default rest | `r` dropped whole | — |

80×24 footers of those states do not clip mid-word.

---

### Verdict

No BLOCKER, no MAJOR, no MINOR, no NIT on this head. Round-1 D1/D2/D3 are discharged by the pixels, not by the model. **This round is TERMINAL.**

Did not merge.

damianvtran:
### UX review — round 2

Scope: e46da03d01d7b2c3c22fcc0578d78c0645fd28a6..17a350620f2438b797471fdebacad2af3734c3a0

Drove the real `OperatorApp` in `~/local-operator-worktrees/edit-model` at head `17a35062` via `run_test` against a scratch `LOCAL_OPERATOR_CONFIG_DIR`. Real keys and the same `_click_row` stand-in the suite uses. Hashed `config.yml` bytes after every gesture. Compared against `~/local-operator-worktrees/settings-edit-model.md` §2.3 / §7.2 / §7.5 and the round-1 remediations (`a6f766e8` + `17a35062`).

**Verdict: TERMINAL.** No BLOCKER, no MAJOR. U1 is gone on the surface, not just in the model. The three round-1 follow-ups that this round was asked to re-walk (C3, D2/U5, the no-byte-without-enter contract) hold. I looked for a third leak path in the remediation itself and did not find one.

---

### Round-1 remediations, re-walked

#### U1 (BLOCKER) — fixed

**Flow.** `/settings` → `tui.theme` → `enter` → `down` (live theme `light`, file absent) → jump key.

| after | `current_theme()` | `_expanded` | file | next `esc` |
|---|---|---|---|---|
| `end` | `dark` | `None` | `NO FILE` | leaves the page |
| `home` | `dark` | `None` | `NO FILE` | leaves the page |
| `pagedown` | `dark` | `None` | `NO FILE` | leaves the page |
| `pageup` | `dark` | `None` | `NO FILE` | leaves the page |

Footer after `end` is `↑↓  move · ←→  panes · esc back to conversation` (read-only last row, no `esc cancel`). The teleport is gone.

Follow-throughs that were consequences of the ghost, now clean:

- `home` then `enter` on `hosting`: `expanded=None`, `editing='hosting'`, theme `dark`. One mode. First `esc` cancels the editor and stays on the page.
- `end` then `enter` on the retired row: refusal paints, `expanded=None`, theme `dark`, file absent.
- Arrow out of the group (`up` past the owner): collapses on the second `up`, theme restored, file absent; next `esc` leaves.

Paint after revert: cream headings `#e9e5db`, cursor `#38c96a`. No Monokai ink (`#a6e22e` / `#f8f8f2`) on the cancelled frame.

#### C3 — `space` on an open bool owner collapses, does not write

`retry.enabled` → `enter` → `up` onto the owner → `space`: `_expanded` goes `None`, file still `NO FILE`, footer returns to `enter  change · esc back to conversation`. `space` on a *choice* still accepts (specified): `enter` / `down` onto `off` / `space` writes 1025 bytes and collapses. Accelerator and expansion no longer fight.

#### D2 / U5 — editor footer, `esc` cancels in-page, no `r`

Painted frame (not `rendered_hints()`), 100×30 error-adjacent editor on `retry.maxRetries`:

```
↑↓   move cancel  ·  enter   save  ·  ←→   panes  ·  esc  cancel
```

80×24: `↑↓   move cancel  ·  enter   save  ·  esc  cancel`. No `r default`. No `esc back to conversation`. No `move · cancels`. First `esc` closes the editor and stays on the page (`editing=None`, `settings=True`); second `esc` leaves. Typing `r` while editing inserts: buffer `'10r'`, file untouched.

#### Contract — no byte change without enter

Open `tui.theme` / `retry.enabled` / `retry.maxRetries`, browse or type, then `up` `down` `pageup` `pagedown` `home` `end` `left` `right` `escape` `tab`. File stays `NO FILE` on all three rows.

---

### Third leak paths (the round-2 hunt)

Looked at every landing that is *not* `action_move` / `action_jump` / `action_section`, plus the gestures that stay put on purpose.

| gesture while previewing `light` | theme after | `_expanded` | file | notes |
|---|---|---|---|---|
| wheel | `light` | `tui.theme` | `NO FILE` | documented §7.5 departure — cursor has not left |
| `←` / `→` pane | `light` | `tui.theme` | `NO FILE` | cursor has not left; footer still `esc cancel` |
| resize 120×32 → 80×24 → back | `light` | `tui.theme` | `NO FILE` | `on_resize` only `_repaint`s; correct |
| click a header | `light` | `tui.theme` | `NO FILE` | unselectable, no-op; next `esc` cancels in-page (does not teleport) |
| click `hosting` (outside group) | `dark` | `None` | `NO FILE` | `_land(scroll=False)` settles; next `esc` leaves |
| click owner, then click again | `dark` | `None` | `NO FILE` | select-then-activate; second click is the named cancel |
| click next choice (`monokai`) | `monokai` | `tui.theme` | `NO FILE` | still inside the group; preview updates, nothing stored |
| `tab` | `light` | `tui.theme` | `NO FILE` | focus stays on `SettingsView`; next `esc` cancels in-page |
| `_close_settings_view` | `dark` | (unmounted) | `NO FILE` | teardown restore still fires |

No third path leaves the app wearing a theme the file disagrees with once the cursor has actually left the group. Wheel / pane / resize / header-click / tab keep the preview *because the cursor is still on the choice* — and `esc` after each of them still means cancel, not leave.

`_land()` is the post-landing counterpart on `action_move`, `action_jump`, both arms of `action_section` (including the last-section-to-last-row early return), `_select_after_index`, and `on_click`. `_revert_preview` calls `_apply_theme(restore)` before the page's own `_repaint`. That is the D1 paint fix, and the cancelled frames wear dark's ramp.

---

### What is actually good (so it is not re-lit)

- Jump/section keys now settle the expansion. `esc` after them leaves.
- Open-editor footer is grammatical and agrees with the keys at both measured widths.
- `r` is unadvertised while editing; printables stay in the buffer.
- `space` on an open bool owner is a cancel; `space` on a choice is still the specified accept.
- Click-elsewhere, arrow-out, re-press on the owner, leave-the-page, and teardown all restore `dark` with the file absent.
- Version is `0.44.3`; `origin/main` is `0.44.2`; tags stop at `v0.44.2`.

No new U-findings. Round 2 is TERMINAL.

damianvtran:
### Agent review — round 2

Scope: b8cd5336118e9f300b5a322d756f886bc0f2a66f..17a350620f2438b797471fdebacad2af3734c3a0

Drove the real `OperatorApp` under `run_test` against a scratch config dir on `17a35062`. Gates on this tree: flake8, `black==26.1.0 --check .`, `isort==5.13.2 --check .`, pyright all clean. `tests/unit/tui/test_settings_view.py`: 502 passed in 82s (`-n 4`).

Round-1 C1/U1 was reproduced as a user gesture, not trusted from the suite: `/settings` → `tui.theme` → `enter` → browse onto `light` → `end`. After `end`: `current_theme()=='dark'`, `expanded_key is None`, `_preview_before is None`, no `config.yml`. The next `esc` left the page (`settings` class gone, theme still `dark`). Same restore on `home` / `pageup` / `pagedown`. The two follow-throughs that made the leak permanent at `e46da03d` are gone: `end` then `space` on a bool leaves the theme at `dark` with `_preview_before is None`; `home` then `enter` on `hosting` opens the editor with `expanded is None` (one state, not two).

The remediation was reviewed as new code, which is where #387's later rounds found defects.

---

#### Round-1 findings, verified against this head

- **C1 (BLOCKER)** — fixed in `a6f766e8`. `_land()` is the post-landing funnel (`_settle_expansion` + `_preview_choice` + `_repaint`). Callers: `action_move`, `action_jump`, `action_section` (both the last-row clamp and `_select_after_index`), `on_click` (`scroll=False`). Reproduced as above.
- **C2 (MAJOR)** — fixed in `557fef0c`. `pyproject.toml` is `0.44.3`. `origin/main` is `0.44.2`. Tags and GitHub Releases stop at `v0.44.2`. PyPI latest is `0.44.2`. `0.44.3` is free.
- **C3 (MINOR)** — fixed in `a6f766e8`. `space` on the open BOOL owner: `wrote=False`, `expanded_key is None`, cursor stays on `retry.enabled`.
- **C4 (MINOR)** — fixed in `a6f766e8`. `_revert_preview` calls `_apply_theme(restore)` with the default full sweep (`preview=False`), then the page `_repaint`s. Esc from a Monokai preview: headings `#e9e5db`, cursor `#38c96a`, no `#a6e22e` / `#f8f8f2`. `_apply_theme` does not persist (`_persist_theme` is the `/theme` commit path); esc after a preview created no `config.yml`.
- **C5 (NIT)** — fixed in `a6f766e8`. `_reveal_cursor` no longer claims the page writes immediately; `action_jump`'s comment names the expansion dropping choice rows.

Editor footer (the D2/U4 claim): at 100×30, `rendered_hints()` is `↑↓  move cancel · enter  save · ←→  panes · esc cancel`; at 80×24 the pane hint sheds and `esc cancel` remains. No `r` while editing. `_reset_applies` returns False when `_editing is not None`.

---

#### New code in the remediation — what was checked and is not a finding

`_land()` does **not** cover wheel, pane switch, or resize. That is correct, not a third leak:

- **Wheel** (`_scroll_rows`) moves the viewport and leaves the cursor on the choice. Spec §7.5 already named this as a conscious departure: nothing has left the group, so there is nothing to cancel. Reverting here would snap the theme because the user glanced down the page.
- **Pane** (`action_pane`) does not move `_selected`. Measured: `right` while previewing `light` keeps `expanded='tui.theme'`, `current_theme()=='light'`, cursor still on the choice. The cursor never left.
- **Resize** (`on_resize`) only `_repaint()`s. Same reason: the cursor has not travelled out of the group.

`_enter_cascade` still assigns `_selected` without `_land()`. It is not a leave-group path a user can take while a theme expansion is open: reaching the cascade row requires `action_move` / `action_jump` / `action_section` / `on_click`, all of which now land first. Measured: click the cascade row while Theme is previewing → first click restores `dark` and clears `_expanded`; the activate click then enters the cascade with no preview capture left.

`_revert_preview` → `_apply_theme(restore)` does not write, does not leave `_preview_before` set (cleared before the apply; overlapping routes are then a no-op), and does not re-preview on the way out: `_land` calls `_preview_choice` after settle, and that returns immediately when the landing row is not a previewing choice. A 400 ms drain after `down` then `end` with no pause between them still ended on `dark` with `_preview_before is None`.

---

**Verdict: clean.** No BLOCKER and no MAJOR remain. This round is TERMINAL: remaining minors and nits, if any, are follow-ups, not a reason for another round.

damianvtran:
### Agent review — round 3

Scope: 3fb85a6920bcb381e2a6bd1ca0fecf56d281812b..95255901cf03f868998df1bc42965c515b3b30d0

Drove the real `OperatorApp` under `run_test` against a scratch config dir on `95255901`. This round exists because origin/main moved to 0.44.3 after round 2 (`17a35062`, all three disciplines TERMINAL) and the branch was rebased to take 0.44.4. It is not a re-litigation of round 2.

Gates on this tree: flake8, `black==26.1.0 --check .`, `isort==5.13.2 --check .`, pyright all clean. `tests/unit/tui/test_settings_view.py`: 504 passed in 78s (`-n 4`). Version `0.44.4`. Tags and GitHub Releases stop at `v0.44.3`. PyPI latest is `0.44.3`. `0.44.4` is free.

---

#### Reproduced on this head (not trusted from the suite)

1. `/settings` → `tui.theme` → `enter` → browse onto `light` (`current_theme()=='light'`) → `end`. After `end`: `current_theme()=='dark'`, `_expanded is None`. Same restore on `home` / `pageup` / `pagedown`.
2. Browse onto Monokai, `esc`. Model restored to `dark`. Next painted frame has no Monokai lime `#a6e22e` and no Monokai heading `#f8f8f2` — `_revert_preview` actually repaints, not Monokai-ink-with-dark-model.
3. `r` on `retry.enabled` at its default: no `config.yml` created, `notice_text==''`, footer is `↑↓  move · enter  change · ←→  panes · esc back to conversation` with `_reset_hint.display is False`. The footer does not advertise `r`.
4. `enter` on `retry.enabled` (group open), `up` onto the owner, `space`: `_expanded is None`, config bytes unchanged (no file before or after), stored value still `True`. Isolated in a fresh app so an earlier unlink could not pollute the comparison.

---

#### Rebase preservation vs round-2 head `17a35062`

`17a35062` is not an ancestor (expected after rebase). Tree-to-tree test inventory of `tests/unit/tui/test_settings_view.py`:

- Round 2: 85 tests. This head: 87 tests.
- **Removed:** none. Nothing from round 2's remediations was dropped.
- **Added:** `test_reset_on_a_default_row_leaves_config_byte_identical`, `test_reset_still_writes_on_an_off_default_row` — main's `#472` pair, kept.

The one assertion change is `95255901` (`test(tui): a default-row r is silent once the hint is shed`):

```
-        assert "default" in view.notice_text, view.notice_text
+        assert view.notice_text == "", view.notice_text
```

The byte-identical assertion is untouched. The comment names the premise change: main's safety patch kept `r` lit and reported "already at its default"; #440 sheds the hint, so a notice would advertise a key the footer no longer offers. That is coherent with the shed-hint contract, not a silent weakening. `action_reset`'s `is_default` guard returns without writing and without setting `_notice` (`settings_view.py:1136-1139`), matching the test.

Conflict resolution:

- `scripts/settings_shot.py` — BOTH kept. Main's `reset-default` state sits next to #440's `bool-open` / `choosing` / `preview` / `reset-offered`.
- `action_reset` — THEIRS (#440) taken. `r` on a default row is silent; the hint is shed.

---

#### MINOR (follow-up, not a reason for another round)

- **C1 (MINOR)** — leftover main-side comments still describe the keep-hint-and-report contract that this branch shed. `action_reset`'s docstring says "A row already at its default is reported, not written" (`settings_view.py:1036-1038`) while the body two screens below is "Silent rather than explained" (`1133-1135`). `scripts/settings_shot.py` `reset-default` still claims the press reports `already at its default (...)` and "The footer still offers `r default` on both sides" (`45-53`, `348-351`). Runtime and the test follow #440; the comments follow main. A future recapture of `reset-default` would caption silent/shed-hint frames as the keep-hint-and-report frames.

---

**Verdict: clean.** No BLOCKER and no MAJOR remain. This round is TERMINAL: remaining minors and nits are follow-ups, not a reason for another round.

PR #472, opened 2026-08-30
fix(tui): r on a setting already at its default no longer writes config.yml

### LINKED ISSUE #440: /settings editing model: explicit edit/accept/cancel, so exploring a setting cannot change your setup

damianvtran:
## Summary

**Pressing `r` on a `/settings` row that is already at its default rewrote `config.yml` — and on a machine with no config.yml at all, it CREATED one.** Reproduced against `origin/main` before anything was changed, at the byte level the issue's audit used.

`action_reset` had no default-state guard, so `r` ran `reset_setting` unconditionally. `reset_setting` deletes the stored key and `_delete` writes the file back whether or not the key was there. The page's own undo gesture was the only thing in a session that had changed anything.

`settings_io.is_default()` already existed for exactly this question and had **zero call sites** outside its own module. The guard uses it, so the row the value column paints dim as "this is the default" and the row `r` declines to reset are the same row by construction.

**Does not bump the version. Does not close #440.** That issue tracks the full editing-model redesign and must stay open.

Part of #440.

## What this PR closes, and what it does not

#440 called out two separable defects to "fix ahead of the redesign":

| # | Defect | Status |
|---|---|---|
| 1 | BLOCKER: `enter` on the cascade row destroys the failover config | **Already on main.** Closed by #449 (`9c327496`). Re-measured on this branch: seed a cascade, `enter` then `down` → config.yml identical (1102 bytes, same sha), stored value still a `dict`, `read_chains()` still returns the chains. Mutation-tested: removing the `Kind.CASCADE` activate branch alone is not enough (the `_begin_edit` allow-list is a second layer); removing **both** layers makes `test_enter_on_the_cascade_row_does_not_destroy_the_cascade` die with the exact wreckage the issue described (`"{'default': [...]}x"`). Verified real coverage. |
| 2 | `r` writes on a row already at its default | **This PR.** |

Out of scope, deliberately — they need the operator's sign-off on the spec: the `enter` opens / `enter` accepts / `esc` cancels redesign, bools expanding rather than toggling, theme preview, the footer copy changes (including offering `r` only when off-default), and the move-commits reversal.

## Reproduction, on the real page

A scratch `LOCAL_OPERATOR_CONFIG_DIR`, the real `OperatorApp`, comparing `config.yml`'s **bytes** before and after each gesture.

**Before — `origin/main`:**

```
== DEFECT 2: r on a row that is already at its default ==
  is_default(retry.enabled)=True
  r on retry.enabled with NO config.yml
    before: exists=False bytes=-1 sha=-
    after : exists=True  bytes=1006 sha=b7d0ea797fbd
    CHANGED: True
  notice='' error=''
```

That is the issue's audit, reproduced: no file, land on any row, press `r`, a ~1005-byte `config.yml` is CREATED.

**After — this branch:**

```
== DEFECT 2: r on a row that is already at its default ==
  is_default(retry.enabled)=True
  r on retry.enabled with NO config.yml
    before: exists=False bytes=-1 sha=-
    after : exists=False bytes=-1 sha=-
    CHANGED: False
  notice='already at its default (on)' error=''

== CONTROL: r on an OFF-DEFAULT row must still reset ==
  stored=False is_default=False
    CHANGED: True
  stored now=True is_default=True
  notice='' error=''

== DEFECT 2b: r again, now the row is back at default ==
    CHANGED: False
  notice='already at its default (on)' error=''

== DEFECT 2c: row at default, file HAS other non-default settings ==
  retry.enabled is_default=True retry.maxRetries=7 (off-default=True)
    before: exists=True bytes=1005 sha=40d0678efdb0
    after : exists=True bytes=1005 sha=40d0678efdb0
    CHANGED: False
  sibling retry.maxRetries preserved: 7
```

Four states, as required: no file at all; a file where the row is at default; a file where the row is off-default (`r` MUST still write); a row at default in a file that has other non-default settings (`r` must not create or rewrite anything).

## The fix

One branch in `action_reset`, immediately before the only write, after the cascade and read-only exits so their established messages keep priority:

```python
if settings_io.is_default(self._manager, setting):
    self._notice = f"already at its default ({_render_value(setting.default)})"
    self._repaint()
    return
```

Compared **by value**, matching `is_default`'s own contract and the `changed` test that inks the value column two panes over. "No file yet" and "stored value equals the default" are one state here rather than two: `read_setting` falls back to `setting.default` when the key is absent, so a config dir with no file reports 0 of 52 settings off-default.

It **says so** rather than swallowing the press — the footer advertises `r default` on this row, and a lit hint whose key does nothing silently is the "nothing happens when I click" bug one step earlier (UX round 1, U5). The hint is deliberately **not shed** the way a read-only row sheds it: a retired row is permanently inert, while this one becomes resettable the moment its value changes, so shedding would flicker `r default` in and out of the footer as a user toggles a bool. Offering `r` only on off-default rows is the #440 redesign's job.

## Mutation test

Removing the `is_default()` check makes `test_reset_on_a_default_row_leaves_config_byte_identical` die at its own boundary, with the exact wreckage the issue described:

```
AssertionError: `r` on a default row CREATED config.yml out of a setting the user never chose: b"metadata:\n  ...\nversion: 0.43.5\n"
```

The off-default test still passed under that mutation, which is the pair working: a guard that refused every reset would pass the no-write test while destroying the page's only undo.

## Visual

Captured before/after frames of the same row (`display.shimmer` at its default) after pressing `r`, from the real `OperatorApp` (the only host that loads `local_operator.tcss`):

- **Before:** press is silent. Detail line still shows the setting description. Footer still offers `r default`.
- **After:** detail line reads `already at its default (on)` in faint (informational) ink. Layout, footer, and value column are otherwise identical — the press no longer writes, and it no longer pretends nothing happened.

`r default` remains in the footer on a default row on purpose (see above). The off-default frame of the same row still offers it, and `r` there still resets.

## Gates

```
flake8 rc=0
black rc=0
isort rc=0
pyright rc=0
```

`tests/unit/tui/test_settings_view.py`: 381 passed (pre-rebase); 10 passed of the `reset or cascade` slice after rebasing onto `origin/main` (`168efe48`).

### DIFF
--- a/local_operator/tui/widgets/settings_view.py
+++ b/local_operator/tui/widgets/settings_view.py
@@ -866,6 +866,10 @@ class SettingsView(Vertical):
         The mitigation for immediate-write having no undo. Deletes the stored
         key rather than writing the default over it, so a config file carries
         only what its owner actually chose (see ``settings_io.reset_setting``).
+
+        A row already at its default is reported, not written — see the
+        ``is_default`` guard below, which is what keeps the page's own undo key
+        from being the only gesture in a session that touched config.yml.
         """
         # Disarms a pending delete for the reason `action_activate` records.
         # Captured BEFORE the early returns below, because a disarm is a change
@@ -924,6 +928,40 @@ class SettingsView(Vertical):
                 self._repaint()
             return
         setting = row.setting
+        if settings_io.is_default(self._manager, setting):
+            # A row already at its default has NOTHING to restore, so `r` must
+            # not reach the writer. It did: `reset_setting` deletes the key and
+            # `_delete` writes the file back unconditionally, so pressing `r` on
+            # an untouched row rewrote config.yml — and on a machine with no
+            # config.yml at all it CREATED a 1005-byte one out of a key the user
+            # never set (#440). The gesture the page offers as the safe way to
+            # undo a change was itself the only thing that had changed anything.
+            #
+            # Both of those are one state here, not two: `read_setting` falls
+            # back to `setting.default` when the key is absent, so "no file yet"
+            # and "stored value equals the default" are the same answer from
+            # `is_default` — measured across the whole registry, a config dir
+            # with no file reports 0 of 52 settings off-default.
+            #
+            # Compared BY VALUE, matching `is_default`'s own contract and the
+            # `changed` test that inks the value column two panes over: a row
+            # painted dim as "this is the default" and a row `r` declines to
+            # reset are then the same row, and the page states one fact about
+            # it rather than two.
+            #
+            # It SAYS so rather than swallowing the press, for the reason the
+            # healthy-cascade branch above does — a lit hint whose key does
+            # nothing is the "nothing happens when I click" bug one step earlier
+            # (UX round 1, U5). The hint is deliberately NOT shed the way a
+            # read-only row sheds it: a retired row is permanently inert, while
+            # this one becomes resettable the moment its value changes, so
+            # shedding would flicker `r default` in and out of the footer as a
+            # user toggles a bool. Offering `r` only on off-default rows is the
+            # #440 redesign's job and needs its sign-off; this guard's job is
+            # only to stop the key from writing when it has nothing to undo.
+            self._notice = f"already at its default ({_render_value(setting.default)})"
+            self._repaint()
+            return
         if not self._save(lambda: settings_io.reset_setting(self._manager, setting)):
             return
         self.post_message(
--- a/scripts/settings_shot.py
+++ b/scripts/settings_shot.py
@@ -45,11 +45,13 @@ STATE selects what the page is showing:
     reset-default  a row that IS at its shipped default (OUT.svg), then the
                same row after pressing `r` (OUT.pressed.svg). `r` used to WRITE
                here — on a config with no file at all it created one — so the
-               subject of this pair is the footer: whether the page offers a
-               key that has nothing to restore. A third frame,
-               OUT.offdefault.svg, is the SAME row once it is off-default,
-               which is what proves the hint is being withheld by state rather
-               than removed (#440)
+               subject of this pair is the DETAIL LINE: the press now reports
+               `already at its default (...)` instead of rewriting the file.
+               A third frame, OUT.offdefault.svg, is the SAME row once it is
+               off-default, which is what proves `r` still resets when there
+               is something to undo. The footer still offers `r default` on
+               both sides — withholding it is the #440 redesign, not this
+               safety patch
     retired    scrolled to the retired section — the read-only bottom row the
                clamp now parks users on, whose footer must not advertise keys
                that cannot act on it
@@ -324,11 +326,11 @@ async def main() -> None:
             await pilot.pause()
             app.save_screenshot(out.replace(".svg", ".open.svg"))
         elif state == "reset-default":
-            # THREE frames. The `r` affordance is a footer question, and a
-            # single still of a row cannot answer it: the pair that matters is
-            # the same row at default and off-default, because only the
-            # difference between them shows the hint is state-driven rather
-            # than simply gone. `tui.shimmer` is used because `_seed_config`
+            # THREE frames. The `r` safety patch is a DETAIL-LINE question, not
+            # a footer one: the hint stays lit on a default row (withholding it
+            # is the #440 redesign), and the pair that matters is the same row
+            # at default (press reports, file unchanged) and off-default (press
+            # still resets). `display.shimmer` is used because `_seed_config`
             # writes it, so both sides are reachable on one config.
             _select(view, "display.shimmer")
             settings_io.reset_setting(view._manager, _require("display.shimmer"))
--- a/scripts/settings_shot.py
+++ b/scripts/settings_shot.py
@@ -42,6 +42,14 @@ STATE selects what the page is showing:
                painted into (D12) rather than against the settings list
     teams      the read-only teams pane
     agents     the read-only agents pane
+    reset-default  a row that IS at its shipped default (OUT.svg), then the
+               same row after pressing `r` (OUT.pressed.svg). `r` used to WRITE
+               here — on a config with no file at all it created one — so the
+               subject of this pair is the footer: whether the page offers a
+               key that has nothing to restore. A third frame,
+               OUT.offdefault.svg, is the SAME row once it is off-default,
+               which is what proves the hint is being withheld by state rather
+               than removed (#440)
     retired    scrolled to the retired section — the read-only bottom row the
                clamp now parks users on, whose footer must not advertise keys
                that cannot act on it
@@ -82,6 +90,7 @@ os.environ["LOCAL_OPERATOR_CONFIG_DIR"] = _SCRATCH
 
 from local_operator import settings_io  # noqa: E402
 from local_operator.config import ConfigManager  # noqa: E402
+from local_operator.settings_io import Setting  # noqa: E402
 from local_operator.tui.app import OperatorApp  # noqa: E402
 from local_operator.tui.widgets.assistant import AssistantBlock  # noqa: E402
 from local_operator.tui.widgets.settings_view import SettingsView  # noqa: E402
@@ -314,6 +323,30 @@ async def main() -> None:
             view.action_activate()
             await pilot.pause()
             app.save_screenshot(out.replace(".svg", ".open.svg"))
+        elif state == "reset-default":
+            # THREE frames. The `r` affordance is a footer question, and a
+            # single still of a row cannot answer it: the pair that matters is
+            # the same row at default and off-default, because only the
+            # difference between them shows the hint is state-driven rather
+            # than simply gone. `tui.shimmer` is used because `_seed_config`
+            # writes it, so both sides are reachable on one config.
+            _select(view, "display.shimmer")
+            settings_io.reset_setting(view._manager, _require("display.shimmer"))
+            view._manager.reload()
+            view._repaint()
+            await pilot.pause()
+            app.save_screenshot(out)
+            geometry = _geometry(app, view, state, size)
+            view.action_reset()
+            await pilot.pause()
+            app.save_screenshot(out.replace(".svg", ".pressed.svg"))
+            geometry += f" || after r at default: notice={view.notice_text!r}"
+            settings_io.write_setting(view._manager, _require("display.shimmer"), False)
+            view._manager.reload()
+            view._repaint()
+            await pilot.pause()
+            app.save_screenshot(out.replace(".svg", ".offdefault.svg"))
+            geometry += f" || off-default hints={view.rendered_hints()!r}"
         elif state == "retired":
             for _ in range(len(view._rows)):
                 view.action_jump(1)
@@ -395,6 +428,18 @@ def _scroll_to_show_group(view: SettingsView) -> None:
         last = index
     height = view._body.size.height
     view._body.scroll_to(y=max(0, min(view._selected - 1, last - height + 2)), animate=False)
+
+
+def _require(key: str) -> Setting:
+    """Resolve a shipped setting or fail loudly.
+
+    A capture that silently skipped a missing key would produce a frame of the
+    wrong state and caption it as the right one.
+    """
+    setting = settings_io.resolve_key(key)
+    if setting is None:
+        raise SystemExit(f"no setting {key}")
+    return setting
 
 
 def _corrupt_cascade() -> None:
--- a/tests/unit/tui/test_settings_view.py
+++ b/tests/unit/tui/test_settings_view.py
@@ -250,6 +250,130 @@ async def test_reset_restores_the_default(tmp_path: Path) -> None:
 
 
 @pytest.mark.asyncio
+async def test_reset_on_a_default_row_leaves_config_byte_identical(tmp_path: Path) -> None:
+    """#440: `r` on a row already at its default must not WRITE.
+
+    `action_reset` had no default-state guard, so the key ran `reset_setting`
+    unconditionally — and `_delete` writes config.yml back whether or not the
+    key was there. Pressing `r` on an untouched row rewrote the file, and on a
+    machine with no config.yml at all it CREATED one out of a setting the user
+    had never chosen: the page's own undo gesture was the only thing in the
+    session that changed anything.
+
+    Asserted on config.yml's BYTES, the unit the issue's audit used, rather
+    than on the stored value. "The value is still the default" is true of the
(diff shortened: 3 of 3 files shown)


damianvtran:
### Design review — round 1
Scope: 168efe48629c4f5dd880601f4c489e5a79259b0b..6606e4ae638923e41eab855c1354a400baddf2a7

Driven the real `OperatorApp` (`run_test`, 100×30, `local_operator.tcss` loaded) on a scratch config. Captured SVG stills, rendered them, and looked at the frames. Geometry on every still: `screen.size == virtual_size == (98, 28)`, no screen scrollbar; body `14` rows against `63` virtual (list scrollbar only, unchanged across the pair). Consecutive settled frames after `r` are identical — no post-paint reflow.

**What this PR did (and did not).** It is a safety patch, not the #440 editing-model redesign. `r` is still **offered** on a default row (`r default` stays in the footer). It is **not** silent: the detail line swaps the setting help for `already at its default (on)` in faint (informational) ink. Off-default `r` still resets. Offering the key only when off-default is explicitly left to the redesign.

**Frames looked at**

- **No `config.yml`, first launch, land on `display.shimmer` (at default), before `r`.** Page is a normal populated `/settings`, not an empty or broken first-run. Detail is the help string. Footer: `↑↓ move · enter change · r default · ←→ panes · esc back to conversation`. Value `on` is dim. File does not exist.
- **Same row, after `r`.** File still absent (0 bytes created). List, value column, footer, and layout are identical. Detail now reads `already at its default (on)` in the same faint slot the help occupied — not danger ink, not an error. Settled frame matches. Reads as “this key does nothing here,” not as “the page is broken.”
- **File exists, row still at default, before / after `r`.** Bytes identical (1023). Same visual pair as above. Sibling `Max retries 7` stays bright (off-default) so the no-op is visibly scoped to the highlighted row.
- **Off-default, before / after `r`.** Before: value `off` in foreground ink (the changed-vs-default styling). After: value `on` and dim again; detail returns to the help string; no notice (the value column is the signal). Bytes changed. Reset is still obvious.
- **`r` affordance.** Still advertised on a default row. Did not disappear.

#### D1 — MINOR — Footer still advertises `r default` on a row already at default

The key no longer writes, but the footer still names it as an action this row will take. That is a lying affordance relative to the chrome — the same class retired rows already shed (`enter change · r default` is withdrawn there).

It is **not** a silent no-op: the detail line reports, in faint not danger, so the press cannot be mistaken for a dead page. Shedding the hint here would flicker it in and out as a bool is toggled (`enter` off-default → hint appears; `r` back → hint vanishes), which is worse than leaving it lit with a report. That trade-off is the right one for a safety patch.

Acceptable ahead of the redesign. The #440 plan — offer `r` only when off-default — is what actually retires this finding. Not a blocker for this PR.

Follow-up, not this round: the notice copy (`already at its default (on)`) is clear; a successful reset staying silent is correct (the value column already shows the landing).

Verdict: **TERMINAL — no BLOCKER or MAJOR findings.** One MINOR (D1), deferred to the #440 affordance redesign. Clean at `6606e4ae`.

damianvtran:
### Agent review — round 1
Scope: 168efe48629c4f5dd880601f4c489e5a79259b0b..6606e4ae638923e41eab855c1354a400baddf2a7

Independent read-only worktree at `6606e4ae`. Diff first (3 files, +207, no version bump), then the four `r` states on the real `OperatorApp` against config.yml **bytes**, then mutation of both guards this round was asked to pin.

## The `r` guard

`action_reset` now returns before `_save` when `settings_io.is_default(self._manager, setting)` (`settings_view.py:931`). Compared by value, matching the `changed = value != setting.default` ink two panes over (`:1837`) and `is_default`'s own contract. Cascade Mapping / read-only exits still sit above it, so their established messages keep priority.

Independent four-state probe (scratch `LOCAL_OPERATOR_CONFIG_DIR`, `pilot.press("r")`, sha of the bytes):

| state | result |
|---|---|
| no `config.yml` at all, `retry.enabled` at default | file still absent; notice `already at its default (on)`; error empty |
| file exists, row at default, sibling `retry.maxRetries=7` | bytes identical (1023, sha `09bf7510171f`); sibling still 7 |
| row off-default (`enabled: false`) | bytes changed (1024 → 1005); key deleted; `is_default` now True; notice empty |
| `r` again once back at default | bytes identical |

Empty cascade (`{}`, which *is* the shipped default) still hits the healthy-cascade branch first: notice `r resets one setting; delete a chain with d on its row`, no write. Malformed-cascade recovery is preserved: a stored string is not a Mapping and is not `== {}`, so it still falls through to `reset_setting`.

**Mutation.** Replacing the `is_default(...)` condition with `False` makes `test_reset_on_a_default_row_leaves_config_byte_identical` die at its own boundary:

```
AssertionError: `r` on a default row CREATED config.yml out of a setting the user never chose: b"metadata:\n  ...\nversion: 0.43.5\n"
```

The paired off-default test still passed under that mutation. The pair is real coverage, not an inherited no-op.

## The cascade claim (#449, already on main)

`test_enter_on_the_cascade_row_does_not_destroy_the_cascade` is dual-layer, as the author said.

- Removing **only** the `Kind.CASCADE` branch in `action_activate`: destroy test still **passes** (the `_TEXT_EDITABLE_KINDS` allow-list in `_begin_edit` refuses to open). `test_enter_on_the_cascade_row_opens_its_own_editor` dies (`kind == 'setting'` instead of `'chain'`), so that test is what pins the travel-into-the-group half.
- Removing **both** layers: destroy test dies with the exact wreckage `#440` described:

```
retry.fallbackChains is no longer a mapping after `enter` on the cascade row — the cascade has been destroyed: "{'default': ['anthropic/claude-opus-5', 'openrouter/deepseek']}x"
```

Not an unpinned inherited guard. This PR does not re-touch that path.

## Scope

`origin/main...HEAD` is `settings_view.py` (+38, the `is_default` branch and its docstring), `test_settings_view.py` (+124, the two new tests), `scripts/settings_shot.py` (+45, a `reset-default` capture). No version bump. No enter-opens / enter-accepts / esc-cancels redesign, no bools expanding, no theme preview, no move-commits reversal, no footer shedding of `r default` on a default row. Footer on a default row after `r` is still `↑↓  move · enter  change · r  default · ←→  panes · esc back to conversation`. That is behaviour (a key that used to write now reports), not a visual change of the affordance.

## Gates (this worktree, exit codes)

```
flake8 rc=0
black==26.1.0 --check rc=0
isort==5.13.2 --check-only rc=0
pyright rc=0
```

`tests/unit/tui/test_settings_view.py -k "reset or cascade or default_row or off_default"`: 10 passed.

CI on the PR: lint / type-check / pip-audit green. `test (3.13)` green on both runs. One `test (3.12)` run failed on `test_a_held_answer_key_resolves_cleanly_on_every_second_key` (`typing answered the prompt`) — steering-approval, not this diff; the other `test (3.12)` run on the same head passed 9335.

#### F1 — NIT — `settings_shot.py` reset-default docstring describes a redesign this PR did not ship

`scripts/settings_shot.py:45-52` (and the in-branch comment at `:327-331`) say the third frame "proves the hint is being withheld by state rather than removed" and name the key `tui.shimmer`. This PR does **not** withhold `r default` — that is explicitly #440's job — and the row it selects is `display.shimmer`. A later capture run against those captions will look "broken" when both default and off-default frames still offer the key. Copy, not behaviour; the capture itself selects the right row and presses `action_reset`.

---

Verdict: **clean**. No BLOCKER, no MAJOR. Round is TERMINAL — the remaining nit is a follow-up, not a reason for another round.

damianvtran:

Head at remediation: `fba413b63b1fb3fe56234ec35f3ad48a54f12d97` (`6606e4ae` + F1 docstring).

#### F1 — NIT — `settings_shot.py` reset-default docstring describes a redesign this PR did not ship — **fixed** in `fba413b6`

The module docstring and the in-branch comment both claimed the third frame proved `r default` was withheld by state, and named the row `tui.shimmer`. This patch does not withhold the hint, and the row it selects is `display.shimmer`. Both captions now describe the detail-line report (`already at its default (...)`) and say the footer still offers `r default` on both sides — withholding it is the #440 redesign.

#### D1 — MINOR — Footer still advertises `r default` on a row already at default — **deferred** — the #440 editing-model redesign is where offering `r` only when off-default belongs

Taking it here would be the affordance redesign the original brief put out of scope ("redesigning how the affordance is advertised is not, beyond not lying about it"). The press is not silent: the detail line reports in faint ink, so the key cannot be mistaken for a dead page. Shedding the hint on a default row would flicker it in and out as a bool is toggled (`enter` off-default → hint appears; `r` back → hint vanishes), which the designer judged worse than leaving it lit with a report. That trade-off is the right one for a safety patch; the redesign is what actually retires this finding.

No other findings. Round 1 is otherwise clean (0 blocker, 0 major on both tracks).

PR #464, opened 2026-08-30
fix(providers): make tool schemas Gemini-compatible

damianvtran:
# PR Description

## Summary

OpenRouter routes OpenAI-compatible tool declarations to upstream model providers. Gemini 3.1/3.5 Flash Lite and Gemini 3.7 Flash reject a function enum containing an empty string, while Qwen accepts the same JSON Schema.

The built-in `agent` tool exposed the empty string as a model-facing sentinel for clearing a saved subagent effort pin:

```text
GenerateContentRequest.tools[0].function_declarations[13].parameters.properties[effort].any_of[0].enum[3]: cannot be empty
```

This replaces the ambiguous empty model-facing value with the explicit `inherit` sentinel at the schema source. The compatibility boundary remains centralized because every provider serializes the same `AgentTool.parameters` schema generated from `AgentParams`.

## Change class

C2 standard/pre-authorized reliability fix. The PR is the system of record; no RFC or tracker is required.

## Behavior and compatibility

- `effort: "lo" | "med" | "hi"` pins that configured model tier.
- omitted `effort` preserves the stored pin.
- JSON `null` preserves the stored pin.
- `effort: "inherit"` clears the pin and returns the profile to inherited model selection.
- legacy callers sending `effort: ""` still clear the pin through a Pydantic before-validator, but the blank value is no longer advertised in provider schemas.
- invalid effort strings remain rejected; no acceptance was broadened.

The complete built-in inventory now has a regression audit for empty enum members, empty enum arrays, compound enum members, and enum members that do not match their declared string type. OpenRouter, direct Google, ordinary OpenAI, and OpenRouter/Qwen serialization paths are covered separately.

### Baseline reproduction

The global runtime on the pre-fix schema reached OpenRouter and failed before generation:

```text
$ lop --hosting openrouter --model google/gemini-3.7-flash exec --json \
    "Return exactly the word SCHEMA_OK. Do not call any tools."
HTTP/1.1 400 Bad Request
invalid request (HTTP 400): Provider returned error: * GenerateContentRequest.tools[0].function_declarations[14].parameters.properties[effort].any_of[0].enum[3]: cannot be empty
```

The declaration index varies with capability-gated tools; both failures resolve to `agent.parameters.properties.effort`.

### Fixed OpenRouter Gemini 3.7

After rebasing onto current `origin/main` (`b8cd5336`), the live worktree-source probe was repeated against the unchanged schema and again reached OpenRouter successfully:

```text
[...]

PR #441, opened 2026-08-30
feat(agent): add op='reset' and show the packaged text when a role diverged

damianvtran:
# PR Description

## Summary of Changes

A packaged role that has been installed and then edited had no way back. `install`
is deliberately idempotent — re-installing an existing role returns it untouched so
a concurrent second launch cannot clobber operator edits — and since #140 the tool
says so honestly:

```
role 'reviewer' is already installed and was left as-is (your edits are kept):
op='show' name='reviewer' to read it, or op='update' name='reviewer' to change it.
```

That message is accurate and both commands it offers dead-end. `op='show'` renders
the operator's own broken text, because once a role is installed the packaged
instructions are unreachable through the tool; `op='update'` asks them to retype
guidance they can no longer read. No wording fixes this, because there is no
correct command to name.

The capability existed but was unexposed: `install_seed(..., overwrite=True)` is a
Python kwarg, not a tool parameter, so `op='install' overwrite=True` was rejected
as `Extra inputs are not permitted`.

This implements both of the issue's live options, because they solve different
halves of the problem.

### `op='reset'` — get the packaged guidance back

The verb a reader looks for by name. It restores the packaged seed over an
installed role and **prints the instructions it replaced**, so the overwrite is
recoverable by copy-paste even though the registry now holds the packaged text.

It is a **separate verb rather than a flag on `install`**, and that separation is
the design. `install` runs *incidentally* — a delegation asking for a role the
operator never created materializes it — so it must be structurally unable to
overwrite; its idempotence is the guarantee protecting a concurrent second launch.
`reset` is only ever reached because a caller typed the word, which is the consent
`install` cannot infer. The distinction is written into the module docstring and
`_op_reset`'s docstring rather than left implicit.

It restores on **evidence, not on name**. A role is overwritten only when it carries
a `seed:<name>` install record, or when its prose still matches the packaged text
byte-for-byte and so holds nobody's writing. A role the operator *authored* under a
starter's name (`scout`) is refused rather than overwritten. A row with neither
(one installed before the record existed, or one written by hand — the two are
indistinguishable, so the tool does not guess) is refused without asserting
authorship it cannot know, and is handed the packaged values to apply via
`op='update'`. The marker is validated against its own row, so a forged or
cross-name tag is inert rather than dangerous.

Divergence is decided by one shared predicate, `seed_divergence`, comparing every
field a seed writes (instructions, `tools`, `effort`, `delegate`). `show` and
`reset` both call it, so they cannot disagree about whether a role is clean, and
[...]

PR #473, opened 2026-08-30
fix(registry): drop a broken recommendation and close the empty-enum audit gap

### LINKED ISSUE #382: fix(tools): empty enum member in agent tool 'effort' schema 400s on Google provider
Found while benchmarking CI sanity models for #380 (agent review round 2, finding raised as a follow-up).

## Symptom

Any request carrying the agent/subagent tool fails with a hard HTTP 400 on Google's native provider:

```
Provider returned error: * GenerateContentRequest.tools[0].function_declarations[14]
  .parameters.properties[effort].any_of[0].enum[3]: cannot be empty
```

Reproduced with `google/gemini-3.5-flash-lite` via OpenRouter against the standard tool set.

## Cause

`local_operator/tools/agent_tool.py:117` declares the effort parameter as:

```python
Literal["lo", "med", "hi", ""]
```

That serialises to a JSON Schema `enum` whose 4th member is the empty string. Google's `GenerateContentRequest` validation rejects any empty enum member, so the request is refused before inference starts.

The `""` member is not accidental: it is the documented sentinel meaning "clear this field" when updating an agent profile. So this cannot simply be deleted without replacing that affordance.

## Impact

Whole-provider outage, not a degraded path: on the native Google provider every request that includes the agent tool 400s. Other providers (OpenAI, Anthropic, most OpenRouter routes) accept the empty member, which is why this went unnoticed.

## Options

1. Drop `""` from the `Literal` and express "clear the field" with an explicit sentinel (e.g. `"none"` / `"default"`), migrating the call sites that pass `""`.
2. Strip empty enum members in the Gemini translation layer, keeping the wire schema valid while preserving the internal sentinel.

(1) is cleaner but touches the agent-profile update path; (2) is contained but leaves an invalid-for-some-providers schema in the source of truth.

## Regression test

Whichever option is taken, add a test asserting that **no** tool schema exposed by the harness contains an empty enum member. That is the check that would have caught this before a provider did.

## Not caused by #380

#380 only surfaced it while measuring candidate sanity models. Filed separately so it gets its own fix and review.

### LINKED ISSUE #383: fix(registry): stop recommending deepseek-v4-flash-0731, it no longer tool-calls
Raised as finding F4 in the agent review of #380.

## Problem

`local_operator/model/registry.py:164` lists `deepseek/deepseek-v4-flash-0731` in `RecommendedOpenRouterModelIds`. That list drives the `recommended` flag surfaced to users (`server/routes/models.py:383`), so the harness actively steers people toward this model.

While benchmarking CI sanity models for #380, that model measured **0/5** on the most basic agentic task the harness has:

> create a file called test.txt with the text 'hello world' in it

Instead of emitting structured tool calls it emits literal markup as assistant **text**:

```
<|DSML|>tool_calls>
<|DSML|>invoke name="write">
<|DSML|>parameter name="path" string="true">test.txt</|DSML|>parameter>
```

No tool executes. The file is never written. This is what took the `cli-sanity` and `server-sanity` CI jobs red on unrelated PRs until #380 swapped the model out.

Measured on the same harness with an isolated config dir, same prompt:

| Model | Pass rate |
|---|---|
| `deepseek/deepseek-v4-flash-0731` | **0/5** |
| `qwen/qwen3.7-flash` | 10/10 |
| `google/gemini-2.5-flash-lite` | 9/10 |
| `openai/gpt-5-nano` | 5/5 |
| `openai/gpt-4.1-mini` | 5/5 |
| `z-ai/glm-5.3-flash` | 5/5 |

## Why it matters

Recommending a model that cannot reliably tool-call in a tool-calling agent harness is worse than not recommending one: a new user picking the recommended option gets an agent that narrates actions it never performs, which reads as the product being broken rather than the model being unsuitable.

## Suggested fix

Remove `deepseek/deepseek-v4-flash-0731` from `RecommendedOpenRouterModelIds`, and consider whether a recommendation list should be backed by a periodic tool-calling smoke check rather than a hand-maintained constant. The `#369`-style benchmark harness used for #380 is a starting point.

Note the failure appears to be a regression in the model itself rather than something specific to our request shape, so a dated-variant swap within the same family may not be safe without measuring.

## Not caused by #380

Pre-existing; #380 only produced the evidence.

damianvtran:
# PR Description

## Summary

Two registry/schema defects, one PR.

**#382** — the empty-enum 400 on Google's native provider — was already fixed at the schema source by #464 (`effort: "inherit"` instead of `""`, with a before-validator so legacy `""` still clears). This PR does not re-do that work. It closes the remaining coverage hole that #464's "full tool bundle" tests left: they built 15 of 23 tools, so an empty enum member in any of the five gated enum-carrying tools (`task`, `hub`, `jobs`, `wake`, `ask`) would have shipped while tests named for the complete inventory stayed green. The helper now attaches every capability so the audit actually sees the whole emitted surface.

**#383** — `deepseek/deepseek-v4-flash-0731` is withdrawn from recommendations. It measured 0/5 on the harness's most basic agentic task, emitting literal `<|DSML|>` markup as assistant text instead of tool calls. The catalogue row stays so a user who already pinned it keeps the 1M window and published prices (deleting it would resolve them to the 128k unknown default). The undated `deepseek-v4-flash` alias is a different id and keeps its recommendation.

Closes #382
Closes #383

## Change class

C2 standard/pre-authorized reliability fix. The PR is the system of record; no RFC or tracker is required.

## Behavior and compatibility

- `effort: "lo" | "med" | "hi"` still pins that configured model tier.
- omitted `effort` / JSON `null` still preserve the stored pin.
- `effort: "inherit"` still clears the pin (the #464 sentinel).
- legacy callers sending `effort: ""` still clear the pin through the existing Pydantic before-validator.
- `deepseek/deepseek-v4-flash-0731` is no longer flagged `recommended` on either the OpenRouter/Radient listing (`RecommendedOpenRouterModelIds`) or the direct-DeepSeek listing (`deepseek_models[...].recommended`).
- An existing config that already pins that model still resolves: `get_model_info("deepseek", "deepseek-v4-flash-0731")` returns `context_window=1_048_576`, `max_tokens=32_768`, `input_price=0.09`.
- The undated `deepseek/deepseek-v4-flash` alias remains recommended.

### #382 — schema on base vs head

Programmatic sweep of the complete 23-tool inventory (`create_tools` with every capability attached):

```
built: 23 ['agent', 'ask', 'bash', 'browser', 'edit', 'eval', 'glob', 'grep',
           'hub', 'jobs', 'list_variables', 'lsp', 'read', 'read_variable',
           'task', 'team', 'team_delete', 'todo', 'wait', 'wake',
           'web_fetch', 'web_search', 'write']
NOT BUILT: []
EMPTY/NULL ENUM FINDINGS: {}
agent effort anyOf[0].enum: ['lo', 'med', 'hi', 'inherit']
```

No empty enum member, no empty `const`, anywhere on the surface. The Google-shaped error (`GenerateContentRequest.tools[0].function_declarations[N].parameters.properties[effort].any_of[0].enum[3]: cannot be empty`) cannot fire: there is no `enum[3]` that is empty.

A live Google-provider request was **not** run: this machine has no OpenRouter / Gemini API key (`credentials.env` has only `GOOGLE_ACCESS_TOKEN` / `GOOGLE_REFRESH_TOKEN` / `GOOGLE_TOKEN_EXPIRY_TIMESTAMP`, which are Workspace OAuth, not a generative-AI key). Validation is against the schema shape the error message names, plus the OpenRouter-Gemini and direct-Google serialization paths in `test_clients.py`.

### #382 — mutation of the existing schema guard

Reintroducing `""` into `AgentParams.effort`'s Literal:

```
FAILED tests/unit/tools/test_agent_tool.py::test_agent_schema_exposes_inherit_without_an_empty_enum_member
FAILED tests/unit/tools/test_registry.py::test_no_builtin_schema_emits_a_provider_invalid_enum
2 failed, 3 passed
```

Both existing guards die. Restored.

### #382 — the coverage hole this PR actually closes

`_default_tools_with_agent()` previously built `ToolContext(cwd=".", agent_registry=object())` — 15 tools, dropping every `createIf`-gated one. Five of those (`task`, `hub`, `jobs`, `wake`, `ask`) declare enums of their own.

Same mutant (empty member on `task.effort`) against the **narrow** helper on origin/main:

```
2 passed, 154 deselected   # the "full tool bundle" tests stayed green
```

Same mutant against the **widened** helper in this PR:

```
FAILED test_openrouter_gemini_full_tool_bundle_has_no_empty_enum_member
FAILED test_direct_google_full_tool_bundle_has_no_empty_enum_member
2 failed, 154 deselected
```

The helper now asserts `{"agent", "task", "hub", "jobs", "wake", "ask"} <= names` so a future gate that silently drops a tool fails here instead of quietly narrowing every audit built on top of it.

### #383 — recommendation withdrawn, existing config still resolves

Driven through the real HTTP route (`GET /v1/models?provider=deepseek`, FastAPI TestClient):

**base (defect):**
```
HTTP 200
  deepseek-v4-flash          recommended=True
  deepseek-v4-flash-0731     recommended=True
```

**head:**
```
HTTP 200
  deepseek-v4-flash-0731     recommended=False
  deepseek-v4-flash          recommended=True
```

Existing-config resolution at head:

```
get_model_info("deepseek", "deepseek-v4-flash-0731")
  -> ctx=1,048,576 max_tokens=32,768 in=$0.09 out=$0.18
build_model_spec("deepseek", "deepseek-v4-flash-0731")
  -> ctx=1,048,576   (unknown-default would be -1)
```

### #383 — mutation of the new guards

| mutant | result |
|---|---|
| re-add id to `RecommendedOpenRouterModelIds` | `test_deepseek_dated_flash_snapshot_is_not_recommended` dies (`assert '…0731' not in […]`) |
| flip catalogue row back to `recommended=True` | same test dies (`assert True is False`) |
| delete the catalogue row entirely | both tests die; `get_model_info` returns `context_window=-1` (the unknown default) |

All three mutants killed. Restored.

## Quality gates

Whole tree, repo config, no narrowing. Exit codes, not stdout:

```
flake8 rc=0
black rc=0
isort rc=0
pyright rc=0
```

Affected files: `305 passed` (`test_registry.py` ×2, `test_agent_tool.py`, `test_clients.py`).

Full unit suite: `9317 passed, 15 skipped, 1 failed`. The failure is `tests/unit/tui/test_steering_approval.py::test_typing_a_steer_at_a_live_prompt_never_answers_it` and **reproduces on origin/main with this PR stashed** (`1 failed in 11.38s`). Pre-existing, not this change.

## What this PR deliberately does not do

- Does not re-implement the #464 schema change (`inherit` sentinel, legacy `""` validator). That is already on main; re-doing it would be a second way of doing the same thing.
- Does not add a periodic tool-calling smoke check behind the recommendation list (the issue's "consider whether"). Out of scope for this cleanup; named here so it is not lost.
- Does not remove `deepseek-v4-flash-0731` from `qwencloud_token_plan_models` or from historical docs (`docs/VERIFICATION.md`, `docs/BENCHMARKS.md`). Those are a dated Alibaba-gateway snapshot and a historical record, not a recommendation.
- Does not bump the version. Manager runs one release after PRs land.
- Did not run a live Google-provider request (no generative-AI key on this machine).

## Notes for the reviewer (not defects)

- `task` / `TaskItem.effort` is `Literal["lo", "med", "hi"] | None` with no clear-sentinel. That is correct, not an inconsistency: `task` launches a child for this turn and has nothing stored to clear. `None` already means "no pin". The agent-tool `inherit` sentinel exists because `op='update'` has to distinguish "leave the stored pin" from "wipe it".
- The comment at `local_operator/model/naming.py:142` still names `deepseek-v4-flash-0731` as a width-clamp example. That is a measurement of a display string, not a recommendation, so it was left alone.

### DIFF
--- a/local_operator/model/registry.py
+++ b/local_operator/model/registry.py
@@ -161,7 +161,15 @@ RecommendedOpenRouterModelIds = [
     "mistralai/mistral-large",
     "x-ai/grok-3-beta",
     "google/gemini-2.5-pro-preview",
-    "deepseek/deepseek-v4-flash-0731",
+    # `deepseek/deepseek-v4-flash-0731` is deliberately ABSENT despite still
+    # shipping a catalogue row below. Measured 0/5 on the harness's most basic
+    # agentic task ("create a file called test.txt"): instead of emitting tool
+    # calls it returns literal `<|DSML|>tool_calls>` markup as assistant TEXT,
+    # so no tool ever executes and the agent narrates work it never performed.
+    # In a tool-calling harness that reads as the product being broken, which
+    # is worse than recommending nothing. The undated `deepseek-v4-flash` alias
+    # stays because the failure is a regression in that pinned snapshot; do not
+    # re-add the dated id without re-measuring it.
     "deepseek/deepseek-v4-flash",
     "deepseek/deepseek-v4-pro",
     "deepseek/deepseek-chat-v3.1",
@@ -1254,7 +1262,12 @@ deepseek_models: Dict[str, ModelInfo] = {
         cache_writes_price=0.09,
         cache_reads_price=0.009,
         description="Pinned July 2026 V4 Flash snapshot",
-        recommended=True,
+        # The ROW stays so a user who already pinned this model keeps correct
+        # pricing and a 1M context window (dropping it would resolve them to the
+        # 128k unknown default and silently mis-set compaction). Only the
+        # recommendation is withdrawn — see RecommendedOpenRouterModelIds above
+        # for the 0/5 tool-calling measurement behind it.
+        recommended=False,
     ),
     "deepseek-v4-pro": ModelInfo(
         id="deepseek-v4-pro",
--- a/tests/unit/model/test_registry.py
+++ b/tests/unit/model/test_registry.py
@@ -3,9 +3,12 @@ import pytest
 from local_operator.model.configure import build_model_spec
 from local_operator.model.registry import (
     ModelInfo,
+    RecommendedOpenRouterModelIds,
+    RecommendedRadientModelIds,
     _anthropic_family,
     anthropic_family_model_info,
     anthropic_models,
+    deepseek_models,
     get_model_info,
     qwencloud_token_plan_models,
     static_models,
@@ -399,3 +402,44 @@ def test_a_family_answer_is_never_the_registrys_own_object() -> None:
     assert first is not None
     first.context_window = 1
     assert anthropic_models["claude-opus-5"].context_window == 1_000_000
+
+
+def test_deepseek_dated_flash_snapshot_is_not_recommended() -> None:
+    """#383: the dated V4 Flash snapshot must not be steered toward.
+
+    It measured 0/5 on the harness's most basic agentic task, emitting literal
+    `<|DSML|>` markup as assistant text instead of tool calls, so a new user
+    picking the recommended option gets an agent that narrates actions it never
+    performs. Both surfaces are asserted because they are read by different
+    callers: `RecommendedOpenRouterModelIds` drives the `recommended` flag the
+    server computes for OpenRouter/Radient listings, while the catalogue row's
+    own `recommended` field is what the direct-DeepSeek listing returns.
+    """
+    assert "deepseek/deepseek-v4-flash-0731" not in RecommendedOpenRouterModelIds
+    # Radient derives from the OpenRouter list, so it inherits the withdrawal;
+    # asserted rather than assumed, since a future edit could fork the lists.
+    assert "deepseek/deepseek-v4-flash-0731" not in RecommendedRadientModelIds
+    assert deepseek_models["deepseek-v4-flash-0731"].recommended is False
+
+
+def test_withdrawn_deepseek_snapshot_still_resolves_for_existing_configs() -> None:
+    """Withdrawing a recommendation must not break a user who already pinned it.
+
+    Deleting the catalogue row would resolve these lookups to the 128k unknown
+    default, silently mis-setting the compaction threshold for a 1M-window
+    model — a worse outcome than the bad recommendation. The row therefore
+    stays; only the flag changes.
+    """
+    info = get_model_info("deepseek", "deepseek-v4-flash-0731")
+    assert info.context_window == 1_048_576
+    assert info.max_tokens == 32_768
+    # Pricing must survive too, or an existing session's cost ledger silently
+    # falls back to the unknown-model zero.
+    assert info.input_price == 0.09 and info.output_price == 0.18
+
+    spec = build_model_spec("deepseek", "deepseek-v4-flash-0731")
+    assert spec.context_window == 1_048_576
+
+    # The undated alias is a DIFFERENT id and keeps its recommendation: the
+    # measured failure is specific to the July snapshot.
+    assert "deepseek/deepseek-v4-flash" in RecommendedOpenRouterModelIds
--- a/tests/unit/providers/test_clients.py
+++ b/tests/unit/providers/test_clients.py
@@ -974,10 +974,74 @@ def _enum_members(node: Any) -> list[Any]:
     return members
 
 
-def _default_tools_with_agent() -> list[AgentTool]:
-    """Build the real default inventory with the capability exposing ``agent``."""
+class _FakeScheduler:
+    """Just enough surface for build_wake_tool's capability check."""
 
-    return create_tools(ToolContext(cwd=".", agent_registry=object()))
+    @property
+    def schedules(self) -> list[Any]:
+        return []
+
+    async def update(self, schedules: Any) -> None:
+        pass
+
+
+class _FakeJobs:
+    """Just enough surface for the job-tracking tools' capability check."""
+
+    def get(self, job_id: str, *, owner_id: str | None = None) -> Any:
+        return None
+
+    def list(self, *, owner_id: str | None = None) -> list[Any]:
+        return []
+
+    async def cancel(self, job_id: str, *, owner_id: str | None = None) -> bool:
+        return False
+
+
+class _FakeComms:
+    """Just enough surface for build_hub_tool: the role test it branches on."""
+
+    def is_child(self, job_id: str | None) -> bool:
+        return False
+
+
+async def _ask_user(questions: list[Any]) -> dict[str, list[str]] | None:
+    return None
+
+
+def _default_tools_with_agent() -> list[AgentTool]:
+    """Build the genuinely COMPLETE inventory, not merely the ungated part.
+
+    A context carrying only ``agent_registry`` builds 15 of the 23 tools: every
+    ``createIf``-gated one drops out, and five of those (``task``, ``hub``,
+    ``jobs``, ``wake``, ``ask``) declare enums of their own. Auditing enum
+    members over that reduced set would let an empty member ship in any of them
+    while a test named for the full bundle stayed green — the same class of gap
+    that let the ``agent`` effort sentinel reach a provider. Every capability is
+    therefore attached so the audit sees the whole emitted surface.
+    """
+
+    tools = create_tools(
+        ToolContext(
+            cwd=".",
+            agent_registry=object(),
+            team_registry=object(),
+            wake_scheduler=_FakeScheduler(),
+            jobs=_FakeJobs(),
+            subagent_comms=_FakeComms(),
+            subagent_launcher=lambda label, prompt, **kwargs: "job-fake",
+            has_ui=True,
+            ask_user=_ask_user,
+        )
+    )
+    # The point of this helper is coverage breadth, so assert the breadth rather
+    # than trusting it: a future gate that silently drops a tool must fail here
+    # instead of quietly narrowing every audit built on top of it. ``browser``
+    # is excluded from the floor because its builder probes a real CMUX surface
+    # that CI does not have.
+    names = {tool.name for tool in tools}
+    assert {"agent", "task", "hub", "jobs", "wake", "ask"} <= names
+    return tools
 
 
 async def test_openrouter_gemini_full_tool_bundle_has_no_empty_enum_member() -> None:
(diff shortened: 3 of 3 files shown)


damianvtran:
### Agent review — round 1
Reviewer: reviewer (zai/glm-5.3)
Scope: 168efe48629c4f5dd880601f4c489e5a79259b0b..76a1dde29eb9e7dad0f3a9c142d7630860f61005

Independent reconstruction, not the PR body's numbers.

**#382 schema source.** Already fixed on `origin/main` by #464 (`f82c2f5e`). `AgentParams.effort` is `Literal["lo", "med", "hi", "inherit"] | None` with a before-validator that maps legacy `""` → `"inherit"`. No empty enum member in the emitted `agent` schema. This PR does not re-touch that code.

**#382 coverage hole — real, and this PR closes the provider-path half.** On `origin/main`, `_default_tools_with_agent()` is `create_tools(ToolContext(cwd=".", agent_registry=object()))`. Rebuilt here:

- old inventory: **15** tools (`agent`, `bash`, `browser`, `edit`, `eval`, `glob`, `grep`, `list_variables`, `lsp`, `read`, `read_variable`, `todo`, `web_fetch`, `web_search`, `write`)
- missing vs `DEFAULT_TOOL_NAMES` (23): `wake`, `task`, `wait`, `jobs`, `hub`, `ask`, `team`, `team_delete`
- gated enum-carrying among those: `task`, `hub`, `jobs`, `wake`, `team` (`ask` has no enum; `wait` / `team_delete` have none)

Mutation: `HubParams.op` Literal extended with `""`. Old helper still builds no `hub`, so `_enum_members` over the old inventory contains no empty string — `test_direct_google_full_tool_bundle_has_no_empty_enum_member` / the OpenRouter twin, both named for the complete bundle, would stay green. New helper builds `hub`; the same Google-path test dies with `AssertionError`. Restored.

`test_no_builtin_schema_emits_a_provider_invalid_enum` already walked `create_tools(_engine_context())` (full 23, including `hub`) since #464. The hole this PR closes is the **provider-serialized** path, not the pydantic-schema audit.

**Sweep completeness.** New helper builds all 23 `DEFAULT_TOOL_NAMES` (including `browser` on this host). `_enum_members` walks the serialized provider body recursively; every `enum` on that surface is visited. Current members, no empties:

`lsp.action`, `todo.op`, `web_search.provider`, `wake.op`, `task.effort` (item + top-level), `jobs.op`, `hub.op`, `agent.op` / `effort` / `kind`, `team.op`. `agent.effort` = `['lo', 'med', 'hi', 'inherit']`. No `const` nodes on the surface today.

**#383.** `deepseek/deepseek-v4-flash-0731` is gone from `RecommendedOpenRouterModelIds` and therefore from `RecommendedRadientModelIds`. `deepseek_models["deepseek-v4-flash-0731"].recommended is False`. `get_model_info("deepseek", "deepseek-v4-flash-0731")` still returns `context_window=1_048_576`, `max_tokens=32_768`, `input_price=0.09`. Undated `deepseek/deepseek-v4-flash` stays recommended. Token-plan row for the same id was already `recommended=False`. `naming.py:142` is a width-clamp measurement of a display string, not a recommendation — correctly left alone. No sibling dated-DeepSeek recommendation left on either listing.

**Clearing effort.** Untouched by this diff; still pinned on main by #464's tests, which I re-ran green: `test_update_effort_wire_semantics` (`inherit` clears, omission preserves), `test_legacy_empty_effort_still_clears_existing_pin` (`""` still clears via the validator), `test_update_null_effort_preserves_existing_pin`. #441's `test_update_keeps_the_effort_tier_and_delegation_flag` still pins that an omitted field does not wipe a stored pin.

Gates (this worktree, whole tree, exit codes): flake8 0, black 0, isort 0, pyright 0. Targeted: `tests/unit/model/test_registry.py` + `tests/unit/providers/test_clients.py` + `tests/unit/tools/test_agent_tool.py` + `tests/unit/tools/test_registry.py` → 305 passed. CI lint / type-check / pip-audit green; unit jobs still in flight at review time.

---

### F1 — MINOR
`tests/unit/providers/test_clients.py:1042-1043`

The helper's floor is `{"agent", "task", "hub", "jobs", "wake", "ask"}`. That set omits `team`, which is `createIf`-gated on `team_registry` and **does** declare an enum (`op: list|show|create|update`). Dropping `team_registry=` from the helper still satisfies the floor, rebuilds 21 tools, and leaves `team` unaudited on the provider path — the same silent-narrowing class this helper exists to prevent.

Today the helper *does* attach `team_registry`, so `team` is in the current 23. `test_no_builtin_schema_emits_a_provider_invalid_enum` also already covers it. This is insurance, not a present miss.

Fix: add `"team"` to the floor (and/or assert `len(names) == 23` minus the documented `browser` CI exception).

### F2 — NIT
`tests/unit/providers/test_clients.py:1015-1018`

Comment lists `ask` among "five … that declare enums of their own". `AskParams` has no `enum` (confirmed: `_enum_members(AskParams.model_json_schema()) == []`). `team` does, and is the one the floor missed.

---

**Verdict: clean.** No blocker, no major. Round 1 is TERMINAL. F1/F2 are follow-ups, not a reason for another round.

damianvtran:

**F1 — fixed in a19c4980.** `"team"` is now in the helper's floor (`{"agent", "task", "hub", "jobs", "wake", "ask", "team"}`). Mutation: dropping `team_registry=` from the helper now fails both `empty_enum` tests with `Extra items in the left set: 'team'` (2 failed). Restored, 2 passed. The silent-narrowing class this helper exists to prevent is now closed for every gated enum-carrying tool, not all-but-one.

**F2 — fixed in a19c4980** (same commit). Comment now lists `task`, `hub`, `jobs`, `wake`, `team` as the five gated enum-carrying tools, and notes that `ask` is gated but has no enum — it stays in the floor so a silent drop of the ask hook still fails here.

Gates after the commit, whole tree, exit codes: flake8 0, black 0, isort 0, pyright 0. Targeted `empty_enum` tests 2 passed.

No other findings in round 1.

PR #154, opened 2026-08-18
fix(resume): offer only the user's own sessions in the /resume picker

damianvtran:
## Change classification

**C2 — standard / pre-authorized.** Narrows what two existing listings *offer*; adds one sidecar file written by the subagent runner. Nothing is deleted and no resume path is removed — `--resume <id>` and `hub op='resume'` still reach a child by id. Clean rollback (`git revert`), and reverting also leaves the now-harmless `origin.json` files inert. No RFC requested.

## The gap

A subagent's child session is an ephemeral directory under `sessions/` with **exactly** the shape of a real conversation: `_build_child_session` mints `config_dir()/sessions/<hex>` and writes a `transcript.jsonl` whose first entry is a `role: "user"` message — the prompt the *parent* wrote.

`recent_session_rows` globs that directory and names each row by its opening user message. Nothing on disk distinguished the two, so `/resume` offered every delegated reviewer, designer and scout run as if the user had opened it, named by the role preamble:

```
[role: reviewer] You are an INDEPENDENT reviewer…   11m ago  447b4d717b3c
[role: designer] You review the user-visible sur…   20m ago  5113f9670e8b
```

On the reported machine that was **40 of 50 rows**, pushing the user's own conversations off the first page.

Two surfaces were affected, not one:

- the `/resume` picker and the CLI's recovery listing (`recent_sessions`);
- **`@latest`** — a bare `--resume` reopened whichever directory was newest, and a delegated review routinely settles *after* the parent's final turn, so it reopened the reviewer rather than the session that launched it.

### Reproduced first, against the real launch path

Not a fixture: this drives `Session._launch_subagent` → `run_subagent`, lets the child write its own transcript, then calls the same function the picker calls.

```console
$ LOCAL_OPERATOR_CONFIG_DIR=/tmp/lo-repro-cfg .venv/bin/python /tmp/lo-resume-origin-repro.py   # on 09d8440
sessions on disk: ['4a5f5a6644ea', 'parent000user']
rows offered by the /resume picker: 2
  4a5f5a6644ea  '[role: reviewer] You are an INDEPENDENT reviewer.'
  parent000user  'I noticed all the subagent sessions appear under /resume'
```

The same script on this branch — the child still runs and still writes its transcript, it is simply no longer offered:

```console
$ LOCAL_OPERATOR_CONFIG_DIR=/tmp/lo-repro-cfg .venv/bin/python /tmp/lo-resume-origin-repro.py   # on d209a7a
[...]

PR #475, opened 2026-08-31
perf(session): take store sweeps off the boot and resume path

damianvtran:
# PR Description

## Summary

Four whole-store disk walks ran serially inside `create_session` and scaled with **store size**, not with the session being built. On this machine's 3574-session / 1.3 GB store they were 77% of a boot's `create_session` and the dominant term of every `/resume`, which re-paid them on a store the same process had just swept.

This is round 1 of the startup-performance work. Two changes, nothing else:

1. **Emptiness probe.** `sweep_sessions` only ever used the zero/non-zero bit of `_dir_size`. The byte sum walked 7111 scandirs / 44372 stats to discover that 0 of 3574 directories were empty. Replaced with a short-circuiting probe plus a one-stat fast path: a non-empty `transcript.jsonl` can never be reaped, and that single stat answers ~92% of the store.
2. **Once per process, after first paint.** The four passes (`sweep_from_config`, `sweep_orphan_groups`, `backfill_session_origins`, `backfill_session_titles`) were already `asyncio.to_thread`'d — so they did not block the event loop — but they were still **awaited**, so they stayed on the critical path. They now run once per process as a background task. `/new` and `/resume` no longer re-sweep.

The execution lease and the claim stay on the loop. The comment at `_prepare` explaining why sole-writer ordering is an invariant is still there and still correct.

## Change class

C2 standard/pre-authorized reliability/performance fix. Patch bump (`0.44.1` → `0.44.2`) per AGENTS.md Versioning: performance work is a patch.

## Deliberate behaviour change

**An unreadable directory now reads as content and is never reaped.**

`_dir_size` returned 0 for a directory it could not open (`rglob` swallows the error), so the sweep treated "cannot look" as "empty and reapable" and could `rmtree` a session whose contents it had never managed to look at. `_holds_content` resolves every `OSError` to "content": if emptiness cannot be proven, nothing is deleted. Strictly safer — a retained empty directory costs bytes, a wrongly reaped one costs the user their session — but it **is** a behaviour change.

The probe otherwise matches `_dir_size(d) > 0` exactly, including the zero-byte-transcript case a naive `scandir` probe gets wrong (a session that crashed before its first write would become immortal).

## What did not change

- `live_dir` skip, the claim marker / `_is_claimed` pid probe, `EMPTY_DIR_GRACE_SECONDS`, and the `_SIDECAR_NAMES` liveness-marker exclusion. The probe replaced only the size call.
- The backfills' internal logic. They are at their traversal floor; moving them off the path is the fix.
- Resume **render** bounding (round 2 — user-visible, needs design + UX rounds).
- The blocked-credential probe walk (round 3).
- The subagent-roster sidecar (measured 6.2 ms — a non-issue).
- `_slash_capabilities` (a profiling artifact: 186 ms isolated, 13.4 ms under the real TUI).

`SweepResult.bytes_remaining` is gone. It was accumulated by summing every byte in the store on every sweep and read by **nothing** — not the factory caller, not a script, not a test. Keeping a field nothing reads is what motivated the whole walk.

## Real-path evidence

All numbers against a **hardlinked replica of the real store** (3574 session dirs, 1.3 GB). Store size is the hidden variable; a 200-session user never sees this. Hosting is the offline `test` provider so the number is store-maintenance plus session construction, with no provider network in it.

### Probe correctness

`_holds_content` vs `_dir_size(d) > 0` on every real directory, plus every synthetic edge case:

```text
real store: 3574 dirs
  old _dir_size(d)>0 :    287.0 ms   holds_content=3574
  new _holds_content :     33.1 ms   holds_content=3574
  speedup            :      8.7x
  DIVERGENCES        : 0

synthetic:
  empty                          old=False new=False  OK
  marker_only                    old=False new=False  OK
  origin_only                    old=True  new=True   OK
  zero_byte_transcript           old=False new=False  OK
  nonempty_transcript            old=True  new=True   OK
  nested_empty_dir               old=False new=False  OK
  nested_zero_byte_file          old=False new=False  OK
  nested_content                 old=True  new=True   OK
  dangling_symlink               old=False new=False  OK
  symlink_to_content_dir         old=False new=False  OK
  unreadable                     old=False new=True   DELIBERATE DIVERGENCE
```

Sweep verdicts across all 3574 real directories: **identical** before vs after (0 evicted either way). Sweep wall: **263 ms → 49 ms** (5.4×).

### Boot and `/resume` wall time

Paired A/B, same machine, same replica, alternating before/after so load drift cancels. `create_session` is the TUI boot worker's path and the `/resume` path.

| | boot `create_session` | `/resume` `create_session` (median of 5) |
|---|---|---|
| before (5 paired runs) | 2764 / 1369 / 1802 / 1519 / 764 ms | 1104 / 1235 / 1444 / 845 / 1206 ms |
| after  (5 paired runs) | 527 / 296 / 565 / 315 / 262 ms | 127 / 279 / 137 / 72 / 107 ms |
| **median** | **1519 → 315 ms (4.8×)** | **1206 → 127 ms (9.5×)** |

Phase attribution on a quieter run of the same replica, before the change:

```text
C1 retention sweep       boot    386.7 ms   (re-paid on every /resume)
C2 orphan group reaper   boot      0.2 ms
C3 backfill origins      boot     67.2 ms
C4 backfill titles       boot     79.5 ms
```

After: those four report **0.0 ms on the critical path** of both boot and `/resume`. They still run — once, in the background, after first paint.

The architect's original "connecting…" window (2.6–3.5 s warm, real TUI, live providers) was 1.2–2.9 s of these four sweeps. Taking them off the path is the round-1 target (~1.2–1.5 s boot, ~3.4 s `/resume` on the 34 MB session). The numbers above isolate the term under change; they do not include Python import, `preflight_usage`, or transcript render.

### Tests

- Probe vs byte-sum on every edge case above, including the unreadable-directory divergence and the zero-byte-transcript case.
- Probe short-circuits: a non-empty `transcript.jsonl` is answered by one stat, with zero `scandir` calls.
- Claim marker and grace window still gate an empty directory after the probe replacement.
- `_prepare` returns without waiting for the sweeps (task handle exists and is not done).
- A second session in the same process does not re-sweep.
- A failing pass never fails session start, and later passes still run.
- Existing claim-before-mkdir race still holds (the concurrent reap now waits on the background task so the assertion is deterministic).
- Autouse fixture in `tests/conftest.py` resets the once-per-process guard, same class of leaked global state as the root-logger fixture beside it.

### Gates

```text
.venv/bin/python -m flake8 .                                          OK
uvx --from black==26.1.0 black --check .                              OK
uvx isort==5.13.2 --check .                                           OK
.venv/bin/python -m pyright --pythonpath .venv/bin/python .           0 errors
env -u NO_COLOR TERM=xterm-256color .venv/bin/python -m pytest tests/e2e -m e2e -n0 -q
                                                                      3 passed
```

Unit suite: the tests this PR owns (retention, session_factory, tui_responsiveness) are 158 passed under xdist.

`tests/unit/tui/test_steering_approval.py` is **flaky, both with and without this change** — re-verified during review remediation. An earlier revision of this description claimed three deterministic failures reproduced at `168efe48`; that claim was wrong. The re-verification (fresh detached worktree per side, this machine, `env -u NO_COLOR TERM=xterm-256color`, no `-p no:randomly`):

| runs | result |
|---|---|
| 8 at merge base `168efe48` | 1–2 failed each in 7/8 runs; 1 clean run; the failing test(s) varied run-to-run |
| 3 at PR head | 1 clean run; 1 run with 1 failure; 1 run with 2 failures |

The failures are timing-sensitive (held-key/steer-vs-live-prompt races in `OperatorApp._request_tool_approval_on_app_loop`) and vary between runs on the same commit, so they cannot be a regression introduced by a diff that touches nothing in steering. Not addressed here: out of scope for a performance PR.

## Out of scope (deliberate)

Round 2 (bound the resume render to last-N) and round 3 (parallelize the blocked-credential probe walk) are separate PRs. Do not bundle them: round 2 is user-visible and needs design + UX rounds; round 3 touches quota-routing invariants.

### DIFF
--- a/local_operator/session/retention.py
+++ b/local_operator/session/retention.py
@@ -165,11 +165,9 @@ class SweepResult:
     ``evicted`` counts only empty directories — a non-zero value never means
     a transcript was removed, because transcripts are never removed.
 
-    There is no ``bytes_remaining``. It was accumulated by summing every byte
-    in the store on every sweep and read by NOTHING — not the factory caller,
-    not a script, not a test — so the field existed only to justify the walk
-    that produced it. Reporting store size is a question for a tool the user
-    runs, not a tax on every boot and every ``/resume``."""
+    Store size is not reported: computing it meant summing every byte in the
+    store on every sweep, and nothing read the number.
+    """
 
     scanned: int = 0
     evicted: int = 0
--- a/local_operator/session/retention.py
+++ b/local_operator/session/retention.py
@@ -82,6 +82,21 @@ LIVE_MARKER_NAME = ".session.pid"
 #: agree on the list.
 _SIDECAR_NAMES = frozenset({LIVE_MARKER_NAME})
 
+#: The transcript file's name, used by :func:`_holds_content` as a fast-path
+#: sentinel: a session that has written a turn has a non-empty one, and a
+#: non-empty transcript is content that can never be reaped.
+#:
+#: Spelled here rather than imported from
+#: ``local_operator.session.transcript``, matching ``resume.TRANSCRIPT_NAME``
+#: which keeps its own copy for the same reason. This module is imported on the
+#: session-construction path and is deliberately lean (115 modules, 13 ms);
+#: importing ``transcript`` for one string literal pulls in the harness types
+#: and attachment store with it (259 modules, 90 ms) and would spend more time
+#: on the import than the probe saves on the walk. The literal is stable \u2014 it
+#: is the on-disk format's name \u2014 and a change to it would break far more than
+#: this constant.
+TRANSCRIPT_FILENAME = "transcript.jsonl"
+
 #: This module's view of the platform. A module-local copy rather than reading
 #: ``sys.platform`` at the point of use, so a test can steer the Windows branch
 #: by patching THIS name instead of the global ``sys.platform`` — patching that
@@ -148,12 +163,17 @@ class SweepResult:
     the tests can assert on it instead of scraping log lines.
 
     ``evicted`` counts only empty directories — a non-zero value never means
-    a transcript was removed, because transcripts are never removed."""
+    a transcript was removed, because transcripts are never removed.
+
+    There is no ``bytes_remaining``. It was accumulated by summing every byte
+    in the store on every sweep and read by NOTHING — not the factory caller,
+    not a script, not a test — so the field existed only to justify the walk
+    that produced it. Reporting store size is a question for a tool the user
+    runs, not a tax on every boot and every ``/resume``."""
 
     scanned: int = 0
     evicted: int = 0
     bytes_freed: int = 0
-    bytes_remaining: int = 0
     errors: int = 0
 
     @property
@@ -181,6 +201,12 @@ def _dir_size(directory: Path) -> int:
     such a directory reads as empty and the reap in :func:`sweep_sessions`
     removes it once :func:`_is_claimed` confirms no live process owns it —
     while a directory with any real content stays untouchable regardless.
+
+    NOT on the sweep's path any more. :func:`sweep_sessions` only ever needed
+    the zero/non-zero BIT of this number and now asks :func:`_holds_content`
+    for exactly that bit, which answers without summing the store. This stays
+    because the byte total is still the honest answer to "how big is this
+    session", and the tests pin the sidecar/origin charging rules here.
     """
     total = 0
     for entry in directory.rglob("*"):
@@ -190,6 +216,106 @@ def _dir_size(directory: Path) -> int:
         except OSError:
             continue
     return total
+
+
+def _holds_content(directory: Path) -> bool:
+    """Does ``directory`` hold anything the sweep must never delete?
+
+    This is the sweep's reap gate, and it answers exactly one question:
+    **is there at least one non-sidecar byte anywhere under this directory?**
+    It replaced ``_dir_size(child) > 0``, which computed a full byte total to
+    read a single bit off it. On the store this was written for — 3574 session
+    directories, 1.38 GB — that total cost 7111 ``scandir``s and 44372 ``stat``s
+    on every boot AND every ``/resume``, to discover that 0 directories were
+    empty. Measured on that store: 257 ms for the byte sum against 24 ms here.
+
+    WHAT IT IS ALLOWED TO CONCLUDE, and why each rule matches the byte sum it
+    replaced — read this before "simplifying" it, because the failure mode of a
+    wrong ``False`` is a deleted session:
+
+    - **A file with a non-zero size is content, and ends the search.** This is
+      the whole point: it short-circuits, where the sum had to keep walking.
+    - **A ZERO-BYTE file is NOT content**, matching the sum exactly (it added
+      0). This is the rule a naive ``scandir``-and-return-False probe gets
+      wrong, and getting it wrong is not merely conservative: a session that
+      crashed before its first write leaves a zero-byte ``transcript.jsonl``,
+      and calling that content makes every such corpse immortal, which is the
+      leak this module's claim marker exists to prevent.
+    - **:data:`_SIDECAR_NAMES` is skipped at every depth**, same set and same
+      reason as the sum: the liveness marker is bookkeeping about the run, so a
+      hard-killed session's marker-only directory must still read as empty and
+      stay reapable. ``origin.json`` is deliberately NOT in that set and so
+      still counts as content, preserving #154/#192.
+    - **Symlinks are not followed into directories**, matching ``rglob``, which
+      does not descend them — so a directory holding only a symlink to a full
+      directory reads as empty under both. A symlink to a non-empty FILE does
+      count under both, because ``stat`` follows it.
+    - **A vanished entry is not content.** A concurrent process disposing its
+      own session mid-walk is normal, exactly as in the sum.
+
+    DELIBERATE BEHAVIOUR CHANGE — an UNREADABLE directory now reads as
+    CONTENT. ``_dir_size`` returned 0 for a directory it could not open
+    (``rglob`` swallows the error), so the sweep read "unreadable" as "empty
+    and reapable" and could ``rmtree`` a session whose contents it had never
+    managed to look at. Every ``OSError`` here resolves to ``True`` instead:
+    if we cannot prove a directory is empty, we do not delete it. That is the
+    direction this whole module errs in — a retained empty directory costs
+    bytes, a wrongly reaped one costs the user their session — but it IS a
+    change, not a pure optimization, and it is why this function is named for
+    what it proves (content) rather than for what the caller wants (emptiness).
+    """
+    # Fast path: one stat answers ~92% of a real store. A session that has
+    # written even one turn has a non-empty transcript, which is content by
+    # definition and can never be reaped — so the common case never opens the
+    # directory at all, let alone walks it. A missing or zero-byte transcript
+    # falls through to the full probe rather than concluding anything: absence
+    # of a transcript is not absence of content (``origin.json``, attachments
+    # and nested output all live beside it).
+    try:
+        if (directory / TRANSCRIPT_FILENAME).stat().st_size > 0:
+            return True
+    except OSError:
+        # Not there, or unreadable. Neither answers the question on its own;
+        # the walk below decides, and it is the walk that owns the
+        # unreadable-means-content rule.
+        pass
+    return _walk_holds_content(directory)
+
+
+def _walk_holds_content(directory: Path) -> bool:
+    """Depth-first search for one non-sidecar byte, stopping at the first hit.
+
+    Split from :func:`_holds_content` so the recursion does not re-run the
+    transcript fast path at every depth — the sentinel is meaningful only at
+    the session root, and a nested file named ``transcript.jsonl`` must be
+    judged as ordinary content by the same rules as anything else.
+    """
+    try:
+        with os.scandir(directory) as entries:
+            for entry in entries:
+                try:
+                    if entry.name in _SIDECAR_NAMES:
+                        continue
+                    # follow_symlinks=False mirrors ``rglob``, which does not
+                    # descend a symlinked directory.
+                    if entry.is_dir(follow_symlinks=False):
+                        if _walk_holds_content(Path(entry.path)):
+                            return True
+                        continue
+                    # A symlink to a file is followed here exactly as the byte
+                    # sum's ``stat`` followed it; a dangling one is not a file
+                    # and contributes nothing, under both.
+                    if entry.is_file() and entry.stat().st_size > 0:
+                        return True
+                except OSError:
+                    # This ENTRY could not be read. It may hold content we
+                    # cannot see, so it counts as content — the safe side.
+                    return True
+    except OSError:
+        # The DIRECTORY could not be opened. See the behaviour-change note in
+        # :func:`_holds_content`: unprovable emptiness is never a reap.
+        return True
+    return False
 
 
 def _process_alive(pid: int) -> bool:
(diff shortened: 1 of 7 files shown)


damianvtran:
### Agent review — round 1
Reviewer: reviewer (claude-opus-5 via lopdev team)
Scope: 168efe48..e5073243

Reviewed as an independent reviewer against the reap-gate failure mode named in the brief (deleting a user's session). Read the full diff, the retention module at head, `_prepare`/`_start_store_maintenance`, the conftest fixture, and the new tests; ran the suites.

**Reap-gate audit — all four belts still in force, and the probe cannot misclassify into a wrongful reap:**

- `live_dir` skip: `retention.py:566-570` — unchanged, resolved-path compare, still first.
- Claim marker / `_is_claimed`: `retention.py:620-637` — unchanged; a live-pid marker keeps the dir at any age, a dead claim reaps immediately without consulting the grace window, an unreadable marker errs toward keeping. The probe is only consulted *before* this gate, never in place of it.
- `EMPTY_DIR_GRACE_SECONDS`: `retention.py:640-658` — unchanged, marker-presence selects which belt is authoritative, same as before.
- `_SIDECAR_NAMES` exclusion: `_walk_holds_content` skips `.session.pid` at every depth (`retention.py:314-315`), matching `_dir_size` exactly, so marker-only corpses stay reapable and `origin.json` still counts as content (#154/#192 preserved).
- The one-stat fast path is safe in every direction: a *missing or zero-byte* transcript falls through to the full walk (`retention.py:287-296`); origin-only, marker-only, nested-content, and nested-zero-byte shapes are all answered by the walk and pinned against `_dir_size` as oracle in `test_the_probe_agrees_with_the_byte_sum_on_every_shape`. The only way the fast path answers `True` is a *non-empty* transcript, which is content by definition. A *wrong `False`* — the session-deleting direction — is unreachable: every `OSError` in both the fast path and the walk resolves to content (`retention.py:322-336`).
- Unreadable-dir behaviour change is named in the PR ("Deliberate behaviour change" section), in the `_holds_content` docstring, and pinned by `test_a_directory_the_sweep_cannot_read_is_never_reaped` + the unreadable-subdirectory variant. Strictly safer than `_dir_size` returning 0.

**Fire-and-forget audit:**

- The task is created but never awaited; `_run_store_maintenance` guards each pass with `except Exception → logger.debug` and re-raises `CancelledError` (`session_factory.py:1263-1275`), so no pass failure can fail or delay session start, and later passes still run (pinned by `test_a_failing_maintenance_pass_never_reaches_the_session`).
- Once-per-process guard: `_start_store_maintenance` sets `_STORE_MAINTENANCE_TASK` synchronously before returning (`session_factory.py:1319-1328`), so two interleaved `_prepare` calls cannot both dispatch — `loop.create_task` returns a handle without running the coroutine, and the guard is checked before creation. A second process has its own module globals and its own loop, so it sweeps too. Pinned by `test_store_maintenance_runs_once_per_process`.
- Task reference is held in the module global, so the loop cannot GC it mid-flight.
- Lease (`session_factory.py:1358-1360`) and claim (`session_factory.py:1374`) are untouched — both synchronous, both before `_start_store_maintenance` at `:1384`, no await between them. The sole-writer invariant comment survives at `:1375-1383`.
- Holding the task rather than a bool also means a still-running pass cannot be re-dispatched by a later `/resume` in the same process.

**Verified locally:** all 158 PR-owned tests pass (retention, session_factory, tui_responsiveness) under `env -u NO_COLOR TERM=xterm-256color`; e2e stage 3 passed; all 114 `test_steering_approval.py` tests PASS at head here.

**M1 — stale testing claim in the PR description.** The description states "Three `tests/unit/tui/test_steering_approval.py` failures reproduce on `origin/main` at `168efe48` and are not this change." I ran that file at the PR head on this machine: **114 passed, 0 failed**. Either the failures were an environment artifact on the implementer's run or the claim is stale, but as written it asserts a reproduction that does not exist, which undermines the evidentiary value of the whole "Tests" section. Fix: edit the PR description to correct or remove that sentence. (Not a code defect — nothing in the diff touches steering.)

**M2 — first-boot `/resume` picker can briefly offer delegated runs (`local_operator/session_factory.py:1278-1328`).** Before this PR the origin backfill had completed by the time `_prepare` returned; now it races first paint. On the very first launch after the upgrade that introduced origin markers, a user opening the picker in the first ~50 ms (a 3574-session store's origin pass measured 67 ms) can see subagent/reviewer sessions listed as resumable — resuming one appends to a delegated run's transcript. Two mitigating paths exist and were verified: `cli.py:2567-2591` still runs both backfills *eagerly and synchronously* before resolving `--resume` on the CLI `--resume` path (unaffected), and `is_user_session` (`resume.py:886-905`) defaults unmarked sessions to *listable* (`not origin` → true), which is exactly the window in question. The implementer's own docstring acknowledges the class of risk ("the `/resume` picker stops offering delegated runs on the FIRST launch after an upgrade"). What would make it right: one sentence in `_start_store_maintenance`'s docstring naming this window explicitly, or stamping origin eagerly (it is the cheap pass, 67 ms) and backgrounding only the other three. Severity is major-borderline-minor: transient, self-healing on next launch, one-time on upgrade, and only the in-TUI picker path.

**N1 — `SweepResult` docstring argues for the deletion rather than describing the type (`retention.py:168-172`).** Five lines of "read by NOTHING … existed only to justify the walk" is PR-narrative, not API documentation; a future reader of the dataclass cannot act on it. One line — "store size is not reported; computing it cost a full walk on every sweep" — carries the same information.

**N2 — dead-letter comment on `reset_store_maintenance_for_tests` (`session_factory.py:1188-1190`).** It says the fixture exists so "a test that needs the sweeps to run twice" can drive them, but no test runs them twice; the autouse fixture exists so each test starts un-swept. Rephrase to what it actually guarantees.

No blocker findings. Two majors: M1 is a documentation-accuracy defect in the PR's evidence, M2 is a real but narrow, self-healing, upgrade-only window on the in-TUI picker path. Verdict: **needs remediation** on M1 (description edit, no code change) and M2 (one-sentence docstring acknowledgment or eager origin stamp — implementer's choice).

damianvtran:

**M1** — fixed (PR description edit, no code SHA). Re-verified personally on a fresh detached worktree of merge-base `168efe48` (8 runs) and at PR head (3 runs), `env -u NO_COLOR TERM=xterm-256color`, no `-p no:randomly`. The original claim of three deterministic `test_steering_approval.py` failures at `168efe48` was wrong: the file is **flaky on both sides** (1–2 failures in 7/8 base runs, 2/3 head runs; the failing tests vary run-to-run — held-key / steer-vs-live-prompt races in `_request_tool_approval_on_app_loop`). The PR description now states that, drops the reproduction claim, and notes the file is out of scope for this PR. Nothing in the diff touches steering.

**M2** — fixed in `e430562c`. Chose the lighter option: a docstring paragraph on `_start_store_maintenance` naming the first-boot-after-upgrade window (origin backfill now races first paint; in-TUI `/resume` picker can briefly offer delegated runs for tens of ms until the background pass stamps them; 67 ms origin pass on a 3574-session store; one-time, self-healing within the same launch; CLI `--resume` still eager). Also names what would make eager stamping worth it (picker growing a correctness dependency on origin). Did not change the reap gate or the backgrounding itself.

**N1** — fixed in `e430562c`. `SweepResult` docstring no longer argues for the field's deletion; one line carries the constraint ("store size is not reported: computing it meant summing every byte in the store on every sweep, and nothing read the number").

**N2** — fixed in `e430562c`. `reset_store_maintenance_for_tests` docstring now states what the autouse fixture actually guarantees (every test starts un-swept) rather than a test that runs the sweeps twice, which does not exist.

Re-ran locally after the docs commit: flake8 / black==26.1.0 / isort==5.13.2 / pyright all clean; 158 PR-owned tests passed; e2e 3 passed. Did not merge.

damianvtran:
### Agent review — round 2
Scope: 168efe48..e430562c

Verified current head `e430562c7fd8e727927c427be0c182abc30c5fbb` (`gh pr view` + worktree `/tmp/lop-sweep-perf`). Round-1 reviewed `168efe48..e5073243`; the only delta to head is `e430562c` (docs-only: `_start_store_maintenance` first-boot picker window, `SweepResult` one-liner, `reset_store_maintenance_for_tests` fixture comment) plus the PR-description edit.

**Round-1 closures**

- **M1** — closed. PR description no longer claims three deterministic `test_steering_approval.py` failures at `168efe48`. It now states the file is flaky on both sides (8 base / 3 head runs, failures vary), that the original reproduction claim was wrong, and that steering is out of scope. Diff still does not touch steering.
- **M2** — closed. `_start_store_maintenance` (`session_factory.py:1312-1324`) names the first-boot-after-upgrade `/resume` picker window (origin backfill now races first paint; tens of ms on a 3574-session store; once per upgrade; self-heals in the same launch; CLI `--resume` still eager). Chose the docstring option; did not change the reap gate or the backgrounding.
- **N1** — closed. `SweepResult` docstring is one constraint line (`retention.py:168-169`), not PR-narrative arguing for the field's deletion.
- **N2** — closed. `reset_store_maintenance_for_tests` now states what the autouse fixture actually guarantees (every test starts un-swept), not a nonexistent twice-run test.

**Reap gate (unchanged by the docs commit)**

Still sound at head. `_holds_content` / `_walk_holds_content` remain the only emptiness probe: sidecar exclusion is `_SIDECAR_NAMES = {LIVE_MARKER_NAME}` at every depth (`origin.json` still counts as content); zero-byte files are not content; every `OSError` (unreadable dir or entry) resolves to content; `live_dir` skip, claim-marker / `_is_claimed` pid probe, and `EMPTY_DIR_GRACE_SECONDS` are untouched and still complementary rather than OR'd. Ran `tests/unit/session/test_retention.py` + `tests/unit/test_session_factory.py` filtered to the probe/sweep/maintenance cases: 18 passed.

no blocker/major/nit findings; round-1 M1/M2/N1/N2 closed.

Verdict: **clean**. Round is TERMINAL.

damianvtran:
### Agent review — round 3
Scope: a1fdc699b01e0c868b9f336d9a602b7536987b15..b69a690c695f2ea47956ba9dad988fcce8d293ef

Verified current head `b69a690c695f2ea47956ba9dad988fcce8d293ef` (`gh pr view` + worktree `/tmp/lop-sweep-perf`). Base is origin/main `a1fdc699` (version 0.44.5). Round 2 was clean against `e430562c`; this head is the same two commits rebased onto current main, with the only conflict in `pyproject.toml` (version now **0.44.6**).

**Rebase check**

- **Version** — `pyproject.toml` is `0.44.6` (patch on 0.44.5). Correct for performance work per AGENTS.md Versioning.
- **Reap gate** — unchanged. `_holds_content` / `_walk_holds_content` remain the only emptiness probe: sidecar exclusion is `_SIDECAR_NAMES = {LIVE_MARKER_NAME}` at every depth (`origin.json` still counts as content); zero-byte files are not content; every `OSError` (unreadable dir or entry) resolves to content; `live_dir` skip, claim-marker / `_is_claimed` pid probe, and `EMPTY_DIR_GRACE_SECONDS` are untouched and still complementary rather than OR'd.
- **Once-per-process dispatch** — unchanged. `_start_store_maintenance` still returns immediately when `_STORE_MAINTENANCE_TASK is not None`, creates the task without awaiting it, and `_prepare` still calls it after the synchronous lease/claim. The four passes (retention sweep, orphan-group reaper, origin backfill, title backfill) still run in `_run_store_maintenance` off the critical path.

Ran `tests/unit/session/test_retention.py` plus the factory/responsiveness cases for the probe, sweep, once-per-process guard, failing-pass isolation, claim-before-sweep, and non-stalling prepare: **34 passed**.

no blocker/major/nit findings; rebase onto origin/main did not reopen round-1/2 findings.

Verdict: **clean**. Round is TERMINAL.

PR #476, opened 2026-08-31
fix(providers): coalesce and bound blocked-credential usage probes

damianvtran:
# PR Description

## Summary

A single cold boot was hitting `api.anthropic.com/api/oauth/usage` eight times and rate-limiting itself into two 429s. The resulting backoff then poisoned the *next* boot — the architect's "warm" run measured a 1218 ms 429, worse than the cold run that succeeded.

Cause, measured with instrumented `fetch_usage` (no live network) against the operator's credential shape (5 anthropic rows, 3 blocked for `claude-opus-5`):

1. **Intra-boundary duplication.** The preflight walk already carried two memos — `attempted_ids` (rows judged) and `quota_cache` (provider+model verdicts) — and neither dedupes the underlying per-account *report*. A fallback chain that lists the walk's own provider re-enumerates accounts the primary probe just read. `leased_account_usage` cannot collapse this: it deliberately never serves a fresh row on the fast path so a routing probe can notice recovery on its own next boundary. That contract is right *across* boundaries and is exactly what leaves the duplication *within* one. Measured: a reserve-state account on a same-provider chain was fetched **twice per boundary**.

2. **Serial blocked-row walk.** `_recover_blocked_accounts` probed blocked rows one at a time (`ensure_oauth_fresh` + a usage GET each). On a 3-row blocked pool that is a network train on the time-to-usable path, and it is what generated the self-inflicted 429 burst.

This PR:

- Adds a **boundary-scoped per-account report memo**, threaded through the same walk as `quota_cache`. One account, one GET, per boundary.
- Parallelizes the recovery walk in **bounded waves of 3**. Anthropic and OpenAI rate-limit their usage endpoints **per source IP regardless of account** (see `usage_cache.py`'s module docstring), so an unbounded gather is just a faster burst. The bound is documented at length on `USAGE_RECOVERY_PROBE_CONCURRENCY` so nobody later "optimizes" it away.
- Selects the verdict in **row order**, not completion order, and prefers a usable sibling already in the wave over a depleted one — so the F2/F5 ping-pong cannot hide spendable quota behind a reserved-and-skipped later row.

Does **not** unify the TUI warmer's cache namespace with preflight's. The warmer's 5 probes of the same accounts share the endpoint but not the cache row (`account_preflight_key`'s docstring rejects unifying them on stated grounds: different shapes, different freshness policies, no identity backfill on the raw preflight fetch). That is a redesign of both subsystems, not this slice.

Sibling PR `dev-sweep-perf` touches `retention.py` / `session_factory.py`. This change is confined to the provider/credential path.

## Change class

C2 standard/pre-authorized reliability fix. The PR is the system of record; no RFC or tracker is required.

## Behavior and compatibility

- Failure semantics are preserved: a transient probe failure keeps the existing block standing and does not invent a verdict; a last-good value is still what a concurrent lease-loser serves.
- `attempted_ids` still terminates the walk. Only rows in a *launched* wave are reserved; the unprobed tail stays a candidate, matching the serial form.
- "First definite verdict" is now "first usable in row order, else first depleted". Observable outcome of F2/F5 (depleted row stays blocked, healthy sibling serves, no provider hop) is identical; the ping-pong round-trip is gone.
- `USAGE_REPORT_TTL_MS` is unchanged. The cross-process cache is not a second cache and is not competing with this memo.
- Next message boundary still re-probes live (the memo dies with the walk).

## Real-path evidence

Instrumented `fetch_usage` (no live network — the operator's account is the realistic environment and we did not hammer the provider). 80 ms stand-in per GET, median of 5 runs.

### Intra-boundary duplicate (1 reserve-state row, fallback chain lists anthropic itself)

| | anthropic GETs | wall |
|---|---|---|
| before (`origin/main` `42145173`) | **2** | 163 ms |
| after | **1** | 83 ms |

The regression test `test_preflight_reads_each_account_usage_once_per_boundary` fails on unfixed `origin/main` with `account probed 2x: ['oauth-a', 'oauth-a']` and passes here.

### Serial blocked-row train (5 rows, 3 blocked, all depleted)

| | anthropic GETs | peak inflight | wall |
|---|---|---|---|
| before | 5 | **1** | **494 ms** |
| after | 5 | **3** (bound) | **330 ms** |

The 3 blocked probes now overlap in one wave. `test_blocked_recovery_probes_run_bounded_and_concurrently` fails on unfixed main with `probes ran serially` (peak=1) and passes here. A 6-row all-healthy pool with inverted completion order still pins the *first* row (`test_blocked_recovery_verdict_follows_row_order_not_completion_order`).

### Second consecutive boundary

The memo is discarded with the walk, so boundary 2 still re-probes live — that is the intended "notice recovery on the next boundary" contract, and it is what stops a stale in-process hit from hiding a window reset. No 429 is provoked because we never burst 8 GETs at one IP in one boundary.

### What the 8-hit boot actually was

Decomposed, not assumed:

| Cause | Hits | This PR |
|---|---|---|
| TUI warmer (`_warm_usage_background` → controller provider-set path) | 5 | **out of scope** — different cache namespace, deliberate |
| Preflight walk (selected row + blocked-row recovery) | 1–5 | **fixed** — no intra-boundary dup; blocked probes bounded-concurrent |

5 + 3 = 8 on the architect's trace. Reproduced 5 + 5 = 10 in the all-depleted worst case. Every account was probed exactly twice per boot, once by each subsystem.

### Existing recovery invariants

`test_recovery_walks_past_a_depleted_blocked_row_to_one_with_quota` (F2) and `test_two_depleted_blocked_rows_do_not_ping_pong_the_recovery_walk` (F5) both failed on the first concurrent draft (whole-wave reservation hid the later healthy sibling) and pass on the usable-first scan. A new test (`test_recovery_wave_prefers_a_usable_sibling_over_an_earlier_depleted_one`) guards that directly.

## Testing

Gates, whole tree, exactly as CI:

- `.venv/bin/python -m flake8 .` clean
- `uvx --from black==26.1.0 black --check .` clean
- `uvx isort==5.13.2 --check .` clean
- `.venv/bin/python -m pyright --pythonpath .venv/bin/python .` 0 errors
- unit suite: the 9 new/related recovery tests pass; the rest of the suite's failures (`test_steering_approval` held-key / live-prompt, `test_parallel_processes_write_atomically`, `test_clicking_the_arrows_scrolls_the_transcript_both_ways`) reproduce on `origin/main` / the sibling `dev-sweep-perf` worktree and are pre-existing flakes, not this change
- `env -u NO_COLOR TERM=xterm-256color .venv/bin/python -m pytest tests/e2e -m e2e -n0 -q` — 3 passed

New tests. Four fail against unfixed `origin/main` (`42145173`) and are the regression guards; two pass on base because they assert invariants the **serial** form already satisfies — they are kept to pin those invariants against the concurrent implementation, not to prove the fix:

- `test_preflight_reads_each_account_usage_once_per_boundary` — **fails on base** (`account probed 2x`); the actual duplication regression guard
- `test_blocked_recovery_probes_run_bounded_and_concurrently` — **fails on base** (`probes ran serially`, peak=1); peak inflight > 1 and ≤ 3
- `test_blocked_recovery_does_not_reserve_rows_it_never_probed` — **fails on base** (no `USAGE_RECOVERY_PROBE_CONCURRENCY`); only a launched wave is reserved
- `test_recovery_probe_failure_leaves_the_block_standing` — **passes on base**; pins that a transient failure keeps the block and does not cancel sibling probes under the gather
- `test_recovery_wave_prefers_a_usable_sibling_over_an_earlier_depleted_one` — **passes on base**; the serial form reaches the same observable via caller re-entry (F2), this pins it for the wave-reserving concurrent form
- `test_blocked_recovery_verdict_follows_row_order_not_completion_order` — **passes on base**; a concurrency-only guard against completion-order selection

## Version

Patch `0.44.2` → `0.44.3` per AGENTS.md Versioning: reliability/performance, no step-function capability.

### DIFF
--- a/local_operator/model/configure.py
+++ b/local_operator/model/configure.py
@@ -1608,6 +1608,27 @@ class SessionStreamFn:
     USAGE_CHECK_TTL_S = 60.0
     DEFAULT_USAGE_BLOCK_MS = 5 * 60 * 1000
 
+    #: How many blocked accounts the recovery walk may probe at once.
+    #:
+    #: The walk used to be strictly serial, which on a pool with several
+    #: blocked rows is a multi-second network train on the time-to-usable
+    #: path (an ``ensure_oauth_fresh`` plus a usage GET per row, one after
+    #: another). Probing them concurrently removes the train.
+    #:
+    #: The bound is what keeps the cure from being worse than the disease,
+    #: and it is NOT a tuning knob to be raised casually. Anthropic and
+    #: OpenAI rate-limit their usage endpoints **per source IP regardless of
+    #: account** (see the module docstring of
+    #: :mod:`local_operator.providers.usage_cache`), so an unbounded gather
+    #: over N blocked rows is a synchronized burst against one IP — exactly
+    #: how this walk used to earn its own 429s, whose backoff then poisoned
+    #: the NEXT boot. Three is deliberately small: it collapses the common
+    #: 3-5 row pool to one or two waves while keeping the instantaneous
+    #: request rate close to what a single interactive ``/usage`` already
+    #: costs. Raising it trades a few hundred milliseconds for the 429 storm
+    #: this constant exists to prevent.
+    USAGE_RECOVERY_PROBE_CONCURRENCY = 3
+
     def __init__(
         self,
         auth_store: AuthStore,
@@ -2013,6 +2034,7 @@ class SessionStreamFn:
         *,
         reserve_percent: float,
         cache: dict[str, str],
+        usage_memo: "dict[str, UsageReport | None] | None" = None,
     ) -> str:
         """Whether ``provider`` still has spendable quota for ``model_id``.
 
@@ -2026,6 +2048,14 @@ class SessionStreamFn:
         lists several models on the same host does not re-hit the usage
         endpoint, while still letting a Fable hop and an Opus hop on the
         same Anthropic pool disagree (their binding windows differ).
+
+        ``usage_memo`` carries the walk's per-account reports, which is a
+        LEVEL BELOW that verdict memo: two model-scoped verdicts on one pool
+        may differ, but they read the same accounts, and this method
+        enumerates every OAuth account of the provider. Without the report
+        memo, a chain that lists the walk's own provider re-fetched usage for
+        an account the primary probe had just read (measured: two GETs for
+        one account in one boundary).
         """
         cache_key = f"{provider}/{model_id}"
         cached = cache.get(cache_key)
@@ -2073,6 +2103,7 @@ class SessionStreamFn:
                         account_id=a.account_id,
                         oauth_creds=a.raw,
                     ),
+                    usage_memo,
                 )
             except Exception:
                 saw_unknown = True
@@ -2106,6 +2137,7 @@ class SessionStreamFn:
                     provider,
                     fingerprint_secret(selected.access_token),
                     lambda: fetch_usage(self._http, provider, api_key=selected.access_token),
+                    usage_memo,
                 )
             except Exception:
                 report = None
@@ -2145,6 +2177,7 @@ class SessionStreamFn:
         different_provider: bool = False,
         reserve_percent: float = 10.0,
         quota_cache: dict[str, str] | None = None,
+        usage_memo: "dict[str, UsageReport | None] | None" = None,
     ) -> Any | None:
         """The first configured fallback with working auth, bench- and quota-aware.
 
@@ -2205,6 +2238,7 @@ class SessionStreamFn:
                 target_model,
                 reserve_percent=reserve_percent,
                 cache=quota_cache,
+                usage_memo=usage_memo,
             )
             if availability == "depleted":
                 if first_depleted is None:
@@ -2236,6 +2270,7 @@ class SessionStreamFn:
         provider: str,
         account_identity: str,
         fetch: "Callable[[], Awaitable[UsageReport | None]]",
+        usage_memo: "dict[str, UsageReport | None] | None" = None,
     ) -> "UsageReport | None":
         """Route one preflight usage probe through the shared cross-process cache.
 
@@ -2244,6 +2279,36 @@ class SessionStreamFn:
         ``leased_account_usage`` never serves a fresh row on the fast path. Fails open
         to a live fetch when the cache is unavailable, so routing can never be made
         WORSE than the pre-cache behaviour. See docs/specs/preflight-usage-cache.md.
+
+        ``usage_memo`` is the boundary walk's per-ACCOUNT report memo, and it is the
+        layer that stops one boundary asking the same account's usage endpoint twice.
+        It is needed IN ADDITION to the two memos the walk already carries, because
+        neither covers this:
+
+        * ``attempted_ids`` dedupes credential rows the walk has JUDGED, but the
+          rows enumerated by ``_provider_quota_availability`` (a fallback-chain
+          question) are not judgements and are deliberately not recorded there.
+        * ``quota_cache`` dedupes provider+model VERDICTS, and cannot be widened:
+          a Fable hop and an Opus hop on one Anthropic pool legitimately disagree
+          because their binding windows differ.
+
+        The underlying per-account REPORT is identical for both, though — it is one
+        GET against one account — so memoizing the report is sound where sharing the
+        verdict is not. Measured: a reserve-state account on a chain that lists its
+        own provider was fetched twice per boundary (once by the primary probe, once
+        by ``_provider_quota_availability`` via ``_first_available_fallback``).
+
+        The cross-process cache cannot collapse this. ``leased_account_usage``
+        deliberately never serves a fresh row on its fast path so a routing probe can
+        notice recovery on its own next boundary; that contract is right ACROSS
+        boundaries and is exactly what leaves the duplication WITHIN one.
+
+        A ``None`` result is memoized like any other: every caller treats it as
+        fail-open (unknown usage, keep the existing verdict), so replaying it inside
+        one boundary reaches the same conservative outcome the re-probe would — while
+        a re-probe of an endpoint that failed milliseconds ago is precisely the burst
+        that earns a 429. Freshness is unaffected: the memo lives only for this
+        boundary walk and is discarded with it.
         """
         from local_operator.providers.usage_cache import (
             account_preflight_key,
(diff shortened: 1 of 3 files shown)


damianvtran:
### Agent review — round 1
Reviewer: reviewer (lopdev) — zai/glm-5.3
Scope: 421451730036efce9754cf8b216d5e7d1dd78e4b..929e8fe3900d538f05ebb581ea1634a6b8dcf054

Reviewed the diff (`configure.py` +601/−95, tests, version bump), then verified by execution rather than inference. No blocker or major findings.

**N1 (NIT)** — PR description, "Testing" section: "all fail against unfixed `origin/main` except the row-order one". I ran the six new tests against the base tree (`42145173`): four fail, but **two** pass — the row-order one *and* `test_recovery_wave_prefers_a_usable_sibling_over_an_earlier_depleted_one`. The latter passing on base is actually consistent with the PR's own argument (the serial form reaches the same observable through caller re-entry into `_apply_account_health`), so the test guards the concurrent regression rather than proving one on base. The description's count is just off by one; a one-word correction ("except the row-order and usable-sibling ones") would make the evidence claim exact.

Verified by execution (worktree `/tmp/lop-cred-probe`, head `929e8fe3`):

- **Regression guards are real.** Copied the head test file onto the base commit: `test_preflight_reads_each_account_usage_once_per_boundary` fails with `account probed 2x: ['oauth-a', 'oauth-a']`, `test_blocked_recovery_probes_run_bounded_and_concurrently` fails with `probes ran serially` (peak=1), `test_recovery_probe_failure_leaves_the_block_standing` fails with the uncaught `RuntimeError` the serial form let propagate, `test_blocked_recovery_does_not_reserve_rows_it_never_probed` fails on the missing constant. All 14 recovery/preflight tests pass at head; `tests/unit/model` is 409 passed.
- **The semantic narrowing is correct on all three required properties.** (a) *No provider hop from a depleted-then-healthy pair in one wave*: the scan prefers usable over depleted, so the healthy sibling settles and the depleted row's block is never lifted (`_settle_recovered_account` is the only mutation point, reached only for the selected verdict). (b) *No healthy sibling skipped across waves*: a wave with only depleted verdicts returns the first depleted, and the caller's re-entry into `_apply_account_health` re-enumerates blocked rows — wave-2 rows were never reserved in `attempted_ids`, so they are still candidates. (c) *Depleted verdict binds to the same row the serial form bound to*: `first_depleted` keeps the first depleted row in row order, and `gather` preserves result order, so selection is deterministic against completion-order races.
- **Termination unchanged.** Waves are reserved before launch (`configure.py:3152-3154`); the unprobed tail stays out of `attempted_ids` (verified by test and by reading the slice loop). `attempted_ids` is only mutated between waves in the single driving task, never concurrently.
- **Attribution holds under concurrency.** Each probe refreshes its own row — `AuthStore._refresh_lock_for` is a per-credential `asyncio.Lock` (`auth_store.py:642-647`), so distinct rows cannot serialize or cross tokens. Probe exceptions are contained (`configure.py:3124-3131`) so one raising row cannot cancel its siblings; the contained failure reads exactly as the serial walk's "unreachable endpoint" (block stands, no invented verdict, siblings still evaluated — the new test covers it).
- **The bound's documentation cites a real source.** `USAGE_RECOVERY_PROBE_CONCURRENCY`'s comment points at `usage_cache.py`'s module docstring, which does state the per-source-IP-regardless-of-account rate limit. The comment explicitly frames the constant as not a tuning knob.
- **Failure semantics preserved.** A memoized `None` replays as fail-open unknown inside the boundary, which is the pre-existing contract for unreachable usage (`_provider_quota_availability` sets `saw_unknown`, never `saw_depleted`); it cannot shrink the sibling set. The cross-process merge in `usage_cache.py` is untouched by this diff.
- **No second cache.** `usage_cache.py` is not in the diff; the memo is a walk-local dict keyed by the same `account_preflight_key` string the persistent cache uses, created per preflight walk (`configure.py:2463`) and discarded with it. All four `fetch_usage` call sites in the walk route through `_cached_account_usage`.
- **The TUI warmer was correctly left alone.** No changes under `tui/` or `providers/controller.py` in this diff, and I agree with the deferral: `account_preflight_key`'s namespace-disjointness docstring gives concrete grounds (one-account vs provider-set shapes, different freshness policies, no identity backfill on the raw preflight fetch). Unifying them is a redesign of both subsystems, not a slice of this fix.
- **Gates, whole tree as CI.** flake8 clean; `black==26.1.0 --check` clean; `isort==5.13.2 --check` clean; pyright 0 errors. Full unit suite: 9341 passed, 1 failed — `test_steering_approval::test_a_held_answer_key_resolves_cleanly_on_every_second_key`, which I reproduced on the base commit `42145173`, so it is pre-existing and correctly disclosed. e2e (`-m e2e -n0`): 3 passed.
- **Version.** 0.44.2 → 0.44.3 patch is right per AGENTS.md: reliability/performance, no step-function capability.

Verdict: **clean** — this round is TERMINAL. N1 is a description-accuracy nit; remaining work, if any, is follow-up, not a reason for another round.

damianvtran:

**N1 — PR description overstated how many new tests fail on base.** Fixed.

Re-verified against the true base `42145173` (`chore(release): bump version to 0.44.2`), all six new tests copied onto that tree, `-p no:randomly`:

```text
4 failed, 2 passed

FAILED test_preflight_reads_each_account_usage_once_per_boundary
       AssertionError: account probed 2x: ['oauth-a', 'oauth-a']
FAILED test_blocked_recovery_does_not_reserve_rows_it_never_probed
       AttributeError: no attribute 'USAGE_RECOVERY_PROBE_CONCURRENCY'
FAILED test_blocked_recovery_probes_run_bounded_and_concurrently
       AssertionError: probes ran serially — the walk is still a network train (peak=1)
FAILED test_recovery_probe_failure_leaves_the_block_standing
       RuntimeError: transient transport failure
PASSED test_recovery_wave_prefers_a_usable_sibling_over_an_earlier_depleted_one
PASSED test_blocked_recovery_verdict_follows_row_order_not_completion_order
```

The reviewer is right, and the cause is mine: I ran the fail-on-base check **before** adding `test_recovery_wave_prefers_a_usable_sibling_over_an_earlier_depleted_one` and reported the stale count. That test passes on base for the reason the reviewer names — the serial form returns the first depleted verdict and the caller re-enters `_apply_account_health`, which walks the remaining blocked rows and reaches the same observable. `test_recovery_probe_failure_leaves_the_block_standing` likewise passes on base, because the serial loop's `continue` on a failed probe already leaves the block standing; under the concurrent form it additionally pins that one probe's exception cannot cancel its siblings' gather. Both are retained deliberately: they are invariant pins for the concurrent implementation, not proof of the fix. That distinction is now stated per-test in the description.

Description corrected via `gh pr edit 476`. No code change — N1 was an evidence-accuracy defect in the description, not in the implementation.

**CI matrix.** `test (3.12)` pass, `test (3.13)` pass on one workflow; the second `test (3.13)` leg was still `pending` at the time of this comment (a rerun of the job that had failed on the pre-existing flake `test_redundant_roster_events_skip_the_sidecar_write`, which passes 4/4 locally and which `main`'s own CI is also currently failing on a *different* flaky TUI test — run `33343441682`, `test_transcript_selection.py::test_a_multi_click_on_a_blank_line_leaves_a_live_range`). Reporting the pending state rather than claiming it green; will confirm when it lands.

PR #477, opened 2026-08-31
fix(tui): keep the fork mark visible under truncation

damianvtran:
# PR Description

## Summary of Changes

Two workstreams on one PR, stacked on `dev-fork-cache-proof` (506876e4).

### 1. Fork UX (this commit)

The shipped `(fork)` suffix in `/resume` fails at **every** width, not just 80 columns. `resume.NAME_MAX_CHARS = 64` condenses the name before the picker sees it, and the name column is 48 cells at 100 columns, so any title over ~40 characters loses the mark — 17% of the titles in the real store, and the long descriptive ones are exactly the conversations a user forks from.

**U1 (blocker).** A dim `[fork]` prefix in its own reserved column, ahead of the inherited title. Unforked rows pad so names stay aligned. The ellipsis now eats the tail of the title instead of the mark. Verified at 100, 80 and 70 columns against the real `OperatorApp` (the host that loads CSS):

before/after at 80 columns

after at 80 and 70

The inherited title is kept — a fork made seconds ago has no other recognition cue. Weight is `dim` (closes shipped D7: at name ink it read as part of the title). The column is reserved from the **result set**, not the page, so names do not jump as a fork scrolls past.

**U2.** A running fork's `conversation_name` is empty by design, so the tab fell back to the parent's opening message. The tab of an unnamed fork is now `lo › [fork] <label>`. `Session.wears_inherited_title` keeps the `_is_unnamed_fork()` verdict that construction already computed and threw away; `set_conversation_name` clears it. Observed on a real fork: empty name → tag on; after naming → tag off (`docs/evidence/fork-ux/lifecycle.txt`).

**U3.** Typing `fork` in `/resume` returned zero rows because the mark was spliced at render time. `filter_rows` / `rank_rows` now search `resume.fork_haystack`. Live frame: `filter fork · 1 of 4`.

filter

**U4.** `mobile/daemon.py` built `{id, name, mtime}` and discarded `forked`. Both payloads now carry it; the phone list renders the same dim `[fork]` prefix. Driven against a real fork on disk:

mobile

**U5.** The success receipt is now as actionable as the failure one:

```
· forked to a1b2c3d4e5f6 — opened in a new cmux workspace (not focused); `lop --resume a1b2c3d4e5f6` reaches it
```

**U6.** Deferred receipt names the escape (`esc to cancel`). The status band shows `⎇ forking` while `has_pending_fork()` is true (survives to 60 columns). Esc already cancelled; the band comes down with the request.

**U8.** `docs/fork.md` now says the argument is a message that spends a billed turn, not a title.

**U7 / U9.** Deferred / no change, per the spec. D1 and the `^F` disambiguation are already good.

The marker **clears** once the fork names itself — picker, tab, phone, filter. Observed, not just asserted:

cleared

**No per-row file read was added.** `SessionRow.forked` already rides out of the `origin.json` the scan parsed. Measured against the parent branch with the identical harness: **1.00 reads/row** on a fork-free store, `origin.json` absent from the per-row set, identical to 506876e4 (`docs/evidence/fork-ux/read-cost.txt`).

### 2. OpenAI cache warmth (parent commit 506876e4)

The shipped Anthropic evidence proved a fork reads a warm prefix on content-keyed cache. The inherited `prompt_cache_key` only reaches the wire on the OpenAI Responses path, which had never been measured.

Measured against real `openai/gpt-5.4` over a real `/responses` endpoint (OpenRouter — the only reachable vehicle). A fork's first request read **21,248 of ~21,837 prompt tokens** from its parent's cache, **10/10 trials**.

**Honest limitation:** the control arms (`own` / `wrong` key) are unstable on that aggregator, and this host has no direct OpenAI key. The measurement establishes the inherited key is **sufficient** for a warm fork, not that it is **necessary**. Full numbers and the excluded causes (retention, streaming, transport) are in `docs/evidence/fork-cache/results.txt`.

### Version

Patch `0.44.2 → 0.44.3`. UX refinement of a shipped feature plus measurement evidence — not a step-function capability. AGENTS.md: when in doubt, patch.

## Testing

Gates over the whole tree, CI invocations:

- flake8, `black==26.1.0 --check`, `isort==5.13.2 --check`, pyright: clean
- `tests/unit` with `-n 4`: 9331 passed, 15 skipped, **1 failed** — `test_typing_a_steer_at_a_live_prompt_never_answers_it`. Reproduced on this branch **and** on `origin/dev-fork-cache-proof` (506876e4) as 1 pass / 2–3 fail. Pre-existing flake, not this diff. Isolated re-run of the fork / picker / status / daemon suites: 225 passed.

Real-path evidence (not the green suite):

- Picker frames from `OperatorApp.run_test` at 100/80/70, long title, marked vs unmarked, marker cleared after naming
- `/resume` filter matching `fork` (live frame)
- Read-cost A/B vs the parent branch: 1.00 reads/row intact
- Marker lifecycle on a real `fork_session()` + `Session.set_conversation_name()`
- Band pending-fork segment at 120/100/80/60
- Phone list against a real daemon payload from a real fork on disk
- Mobile web vitest: 45 passed, including `past-sessions.fork.test.tsx`

**Not exercised:** `fork.mode=window` / `switch` actually opening an OS window or cmux workspace. Receipts and spawn strings were rendered; a window was not watched. Same limitation the UX spec recorded.

## Not addressed

- **U7** — `fork.mode=switch` receipt persistence. Copy is already correct; deferred as low impact.
- **U9** — no change. `^F` disambiguation stays.
- OpenAI cache-key **necessity** — needs a direct OpenAI key.


damianvtran:
### Design review — round 1

Scope: efbf3c3ffb8b6eac957d9df9c99d9db88b103838..01afbfa396c3a538108929cebca5fa18110b159f

Judged from rendered frames, not source. Drove the real `OperatorApp` via `run_test` (the host that loads `local_operator.tcss`) off a read-only worktree of this head; captured picker / receipt / band SVGs and looked at the painted pixels. Also looked at the implementer's stills under `docs/evidence/fork-ux/`. Head confirmed via `gh pr view 477 --json headRefOid` after the rebase onto 0.44.2.

**How this was exercised**

- `/resume` picker at 100 / 80 / 70 cols, long 56-char title, short title, named-fork (tag cleared), filter `fork`.
- Success receipt and deferred acknowledgement at 100 and 80.
- Pending-fork status band at 100 / 80 / 60, idle band as control.
- Tab title: the OSC string through the real `StatusLine` + `TerminalTitle` writer. `run_test` is headless, so this is the string that would be written, not a window I watched open.
- Phone list: judged the implementer's `frame-mobile-history.png`. I did not stand the mobile web surface up myself.

No BLOCKER. No MAJOR.

---

## Picker — `[fork]` prefix survives truncation

Hard requirement, previously failed at every width including 100.

Painted rows from the real picker (card inner, 56-char title):

```
100: ❯ [fork] Refactor the YAML loader to stream ancho…  just now  a1b2c3d4e5f6
     |         Refactor the YAML loader to stream ancho…    2m ago  9f8e7d6c5b4a
80:  ❯ [fork] Refactor the YAML loader to stream…  just now  a1b2c3d4e5f6
70:  ❯ [fork] Refactor the YAML loader…  just now  a1b2c3d4e5f6
```

The ellipsis eats the title tail. The mark is still the first token after the caret at all three widths. Two-cell gap before `just now` at 70 — the old suffix ran straight into the age column. Unforked rows pad the 7-cell column, so names share a left edge (no ragged column). Short title at 100: mark present, names aligned, no ellipsis. Once the fork names itself the column is not reserved and the list returns to a flush gutter.

`screen.virtual_size == screen.size` and no vertical scrollbar at 100/80/70. The overlay is not making the screen scroll.

## Visual weight — dim, not name-ink

Sampled the selected-row pixels on the 100-col still, not the CSS class:

| span | painted | hex |
|---|---|---|
| `[fork]` | dim | `#837c6d` |
| name (selected) | fg | `#e9e5db` |
| age / id | dim | `#837c6d` |
| name (unselected) | muted | `#b5afa2` |

Contrast on the selected ground (`#16221a`): mark 3.97:1, name 13.07:1. The mark is the same ink as age and id, a step quieter than the title — it reads as a qualifier on the row, not as part of the name. That is the D7/U1 weight the spec asked for.

## Tab title

Through the real writer:

```
unnamed fork : lo › [fork] lop-forkux
parent       : lo › Refactor the YAML loader
named fork   : lo › Streaming parser attempt
```

Prefix, brackets, lowercase — same vocabulary as the picker. I did not observe OSC 0 land in a real window.

## Receipts

Success, 100 cols (wraps; both clauses survive):

```
· forked to a1b2c3d4e5f6 — opened in a new cmux workspace (not focused); `lop --resume
  a1b2c3d4e5f6` reaches it
```

At 80 the wrap falls after the semicolon, which is the cleaner of the two. Weight is `note`, not an alarm. `(not focused)` and the resume command are both on screen.

Deferred, 100 and 80, single line, no clip:

```
· forking at the next safe boundary… (esc to cancel)
```

## Status band

`⎇ forking` in `label` violet (`#b48cd6`, 7.12:1 on the sunken band). Present at 100, 80, and 60. At 60 the ladder has shed cwd and cost and still keeps the segment — it is a standing indicator, not a line that scrolls away. Idle band shows nothing about forking.

## Phone list

Judged `docs/evidence/fork-ux/frame-mobile-history.png`, not a live device. `[fork]` is a dim prefix (`text-meta text-ink-dim`) ahead of the inherited title; the parent row is unmarked. Truncation eats the title, not the tag. No reserved column on unforked rows — correct for a phone list; a 7-cell blank on every ordinary row would be the TUI's scroll-stability trick spent where it does not apply.

---

## Findings

None.

## Verdict

**TERMINAL.** 0 BLOCKER, 0 MAJOR, 0 MINOR, 0 NIT.

The truncation failure that was the hard requirement is gone at 100, 80, and 70. Weight, alignment, receipts, band, and the mobile mark (from the implementer's still all match the spec. Design-side this is ready to merge.
EOF
)

damianvtran:
### Agent review — round 1

Scope: efbf3c3ffb8b6eac957d9df9c99d9db88b103838..01afbfa396c3a538108929cebca5fa18110b159f

Read-only worktree `/tmp/lop-rev-477` off `origin/dev-fork-ux`. Head confirmed via `gh pr view 477 --json headRefOid` after the rebase onto 0.44.2. Diff reviewed first; files opened only where the diff could not answer a named question. Targeted suite: `tests/unit/test_fork.py tests/unit/tui/test_status_line.py` — 146 passed. Truncation, `plan_columns`, filter, and picker I/O re-measured against this head rather than taken from the evidence files.

**Not terminal.** One MAJOR on the marker-lifecycle path the spec named as load-bearing. Cache plumbing, picker I/O, truncation, filter, and mobile carry-through hold.

---

### R1 — MAJOR — `/rename` does not clear the live `[fork]` tab marker

The session flag clears. The tab does not.

`Session.set_conversation_name` sets `_wears_inherited_title = False` and the comment at `session.py:3228-3233` says this is safe because both the naming errand **and `/rename`** land here. That is true of the session object. It is not true of the band.

`_store_title` (auto-name) re-reads the flag in the same `status.update()`:

```python
self._status.update(
    conversation_name=stored,
    forked=bool(getattr(session, "wears_inherited_title", False)),
)
```

`_cmd_rename` calls the same `set_conversation_name` and then:

```python
self._status.update(conversation_name=stored)  # app.py:10086
```

`StatusLine.update` treats `forked=None` as leave-alone, so `_forked` stays `True`. `_sync_terminal_title` then prefixes the **new** name:

```
unnamed fork tab:                  'lo › [fork] lop-forkux'
after /rename without forked=False: 'lo › [fork] Streaming parser attempt'
after _store_title with forked=False: 'lo › Streaming parser attempt'
```

Reproduced against this head with the real `TerminalTitle` writer. The disk picker *does* clear (sidecar `st_mtime > forked_at` — measured), so after `/rename` the live tab and `/resume` disagree about the same fact. That is the surface U2 exists for, and it is the path a user takes when they fork and immediately name the branch before auto-name lands.

`_apply_frontend_state` has the same hole (`conversation_name=...` with no `forked=`), so a later snapshot cannot heal it either.

Fix: `_cmd_rename` (and `_apply_frontend_state`) must push `forked=` the same way `_store_title` does. A test through `_cmd_rename` — not through `set_conversation_name` alone — would have caught this; `TestTheWindowTitleNamesTheFork.test_naming_the_fork_clears_it` only asserts the session flag.

---

### R2 — MINOR — the cache-key test docstring overclaims necessity that the evidence file correctly declines

`TestTheForkInheritsItsParentsCacheKey` says the inherited key “read the parent's prefix in 16/16 trials while a non-matching key read in 3/8 — so losing it is a real regression, not a theoretical one.”

That is run B. The evidence file (`docs/evidence/fork-cache/results.txt`) is honest: the canonical N=10 had `inherited 10/10`, `own 10/10`, `wrong 10/10`, and explicitly **does not establish that the key is what did it**. The docstring presents the one run where controls looked like a lock as if it were the result.

The plumbing tests themselves are the right shape (boot source, `_cache_lineage_id != _session_id`, key on the Responses body). Only the claim in the docstring needs to match the evidence file.

---

### Holds (highest-risk items, verified)

- **Picker I/O.** Unmarked store: 1.00 reads/row, files `{title.json}`, no `origin.json`. Mixed store: `origin.json` only on forks (`wears_inherited_title` gated on the origin the scan already parsed). The 2.00→1.00 regression was not given back. `TestThePickerPaysNothingForSessionsThatAreNotForks` still pins it.
- **Truncation.** `[fork]` present and rows distinct at 100/80/70/60/50/40/30/24. At 70: `❯ [fork] Refactor the YAML loader to stream a…  just now  a1b2c3d4e5f6` — two-cell gap before age, ellipsis eats the title. `plan_columns` puts the 7 cells in `marker_col` (fixed chrome), never in a droppable column, so the ladder cannot shed it. Spec's reason for a prefix over a real column holds.
- **Filter.** `filter_rows(..., "fork")` admits tagged rows; `"[fork]"` admits only tagged rows. Titles that themselves contain the word `fork` also match — ordinary substring, intended, not a second meaning.
- **Mobile.** `_past_sessions` / `_search_sessions` both emit `forked`; the phone screen renders `[fork]` as a prefix ahead of `truncate`; daemon unit tests and the vitest screen test cover it.
- **Cache vs sticky credentials.** Still separate. `session_factory` passes `cache_lineage_id=fork_parent(transcript_dir) or None`; `SessionStreamFn` keys sticky creds on `_session_id` only (`pin_session_credential`, `get_api_key`, `stream_with_failover`). Evidence file does not overclaim necessity.
- **Disk marker lifecycle.** `stamped <= forked_at`; rewriting `title.json` clears `wears_inherited_title` on disk. Session boot sets the flag from `_is_unnamed_fork()`; `set_conversation_name` clears it. The live-tab hole is R1, not this.

---

**Verdict:** not clean. R1 is a MAJOR on the marker-lifecycle path. No BLOCKER. Round is not TERMINAL — remaining minors/nits are not the reason; R1 is.

damianvtran:
### UX review — round 1

Scope: efbf3c3ffb8b6eac957d9df9c99d9db88b103838..01afbfa396c3a538108929cebca5fa18110b159f

Walked on a read-only worktree of `origin/dev-fork-ux` at `/tmp/lop-uxrev-477`. Head SHA confirmed with `gh pr view 477 --json headRefOid` after the rebase onto 0.44.2. Drove the **real `OperatorApp`** via `run_test` (the host that loads `local_operator.tcss`) through `/resume` at 80/70/100, typing `fork` into the live filter, a mid-turn `/fork` that armed then cancelled, the opened receipt, and the switch-mode stash-and-flush. Frames at `/tmp/ux-walk-477/`.

**Could not exercise:** a real OS window appearing under `fork.mode=window`, a live cmux sidebar row being scanned by eye, or the phone UI over a running daemon. Those three are specified from the live copy, the title writer, and the mobile renderer — not from watching a window open. The idle `/fork` path *did* call the real cmux backend from this box; I did not raise or close any workspace.

U1–U9 are the prior spec. This round continues at U10.

---

## Walk results against the six asked questions

**1. Bare `/fork` — does the unnamed fork read as a fork in `/resume` at a glance, including with a long parent title and at 80/70 cols?**

Yes. At 80 and 70 columns the reserved `[fork]` prefix survives; the ellipsis eats the title tail, not the mark. Unforked rows pad so names stay aligned. From the live picker:

```
❯ [fork] Refactor the YAML loader to stream…  just now  a1b2c3d4e5f6
         Refactor the YAML loader to stream…    2m ago  9f8e7d6c5b4a
```

at 70:

```
❯ [fork] Refactor the YAML loader…  just now  a1b2c3d4e5f6
         Refactor the YAML loader…    2m ago  9f8e7d6c5b4a
```

The two rows are no longer identical. U1 is closed as interaction.

**2. After the fork's first request names it, does the marker disappear everywhere (picker, tab title, phone if reachable)?**

Picker: yes. After naming, the live picker is:

```
❯ Streaming parser attempt                          just now  a1b2c3d4e5f6
  Refactor the YAML loader to stream anchors inst…    2m ago  9f8e7d6c5b4a
```

No tag, no reserved column. Typing `fork` then returns `0 of 4` / `no session matches that filter` — the tag is gone from the index as well as the paint.

Tab title, through the real `StatusLine` writer: unnamed fork `lo › [fork] lop-forkux`; after `set_conversation_name` `lo › Streaming parser attempt`, `wears_inherited_title` False.

Phone: not driven live. The daemon payload now carries `forked`, the web list renders a dim `[fork]` prefix ahead of the name (so CSS `truncate` eats the title, not the tag), and a vitest on the real screen asserts one tag for two twin rows. I did not open the phone.

**3. Typing `fork` in the `/resume` filter — does it find the tagged rows?**

Yes, by actually typing four keys into the real picker, not by calling `filter_rows`. Header became `filter fork  1 of 4`; only the tagged id remained. `[fork]` as a query also matches (same haystack). Ranked in the name tier, not explained as a body match. U3 closed.

**4. Deferred fork (while a tool is running) — is the acknowledgement durable and is Esc-to-cancel discoverable?**

Partially. The live handler printed:

```
· forking at the next safe boundary… (esc to cancel)
```

and the band lit `⎇ forking` (violet/`label`) at the right edge. After twelve subsequent notices the acknowledgement was still on screen in this 30-row host, and the band still said `forking`. Esc then printed `fork cancelled`, cleared the pending flag, and the band segment vanished.

Two remaining frictions, both walked:

- The acknowledgement is `info` (dim) because `_cmd_fork` calls `notice(body)` without a kind, and `_notice` defaults to `info`. The replacement path next to it is correctly `note`. After a long tool run the dim line is gone and the *only* remaining signal is the band word `forking` — which does not restates Esc. See U10.
- Esc cancelled in this walk because `FakeSession.streaming` was True, so `action_stop` took the interrupt path. The copy says `esc to cancel`; Ctrl+C (`action_interrupt`) is the other key that always withdraws. That match is correct for a live turn. I did not find a state where the copy promises Esc and Esc is a no-op.

**5. Success receipt — does it say where the fork went and how to get there?**

Yes, through the real `_on_fork_complete` with a recording cmux backend:

```
· forked to a1b2c3d4e5f6 — opened in a new cmux workspace (not focused); `lop --resume a1b2c3d4e5f6` reaches it
```

WHAT / WHERE / not-focused / how-to-reach are all on the line. Weight is `note`. At 100 columns the line wraps once, splitting `` `lop --resume `` from the id on the next line — still copyable, slightly uglier than the fallback receipt which stays one line. See U11.

Idle `/fork` (no message) produced the same receipt shape with a real fork id (`b1226d2a071d`) after `forking…`.

**6. `fork.mode=switch`**

The stash-and-flush path was driven: after adoption the live notice is

```
· switched to fork a1b2c3d4e5f6 — the original is still there: lop --resume 9f8e7d6c5b4a
```

Copy still answers "how do I get back". I did not watch this terminal actually reboot onto the fork — `run_test` cannot open an OS window — so U7's persistence-across-rebuild concern is still specified from the flush, not from sitting in the rebuilt transcript. Same deferred candidate as the spec.

---

## Findings

### U10 — MINOR — after the deferred acknowledgement scrolls away, Esc-to-cancel is no longer discoverable from what remains on screen

**Step.** Mid-turn `/fork try the streaming parser`. Immediate receipt: `forking at the next safe boundary… (esc to cancel)` plus band `⎇ forking`. Then twelve tool-progress lines (the stand-in for a long tool run). The dim acknowledgement is still in this short host's viewport; on a real turn it is not. The band still says `forking`. Esc does cancel, and prints `fork cancelled`.

**What a user would expect.** The standing indicator that outlives the receipt should carry the same escape the receipt named, or the receipt should be `note` so it is the line the eye keeps. Right now the durable signal is the verb `forking` with no affordance. A user who armed this ten minutes ago and wants it back has to remember a parenthetical they can no longer see — or discover Ctrl+C by reflex, which also aborts the turn they may still want.

The band is the right place (it is what U6b asked for) and the glyph+word choice is good. It is one word short of being an instruction.

**Instead.** Either `⎇ forking · esc` (the band already spends cells on this rung last-but-two, and the evidence file shows the segment surviving at 60 cols), or emit the acknowledgement as `note` so it matches "fork request replaced" and "fork cancelled". I would not grow the band into a sentence.

### U11 — NIT — the opened receipt wraps at 100 columns, splitting the resume command from its id

**Step.** Real `_on_fork_complete` at 100×30. The notice is 109 characters; with glyph and transcript inset it wraps:

```
· forked to a1b2c3d4e5f6 — opened in a new cmux workspace (not focused); `lop --resume
  a1b2c3d4e5f6` reaches it
```

**What a user would expect.** The command they are told to copy is one token. The fallback receipt next to it (`run \`lop --resume <id>\``) stays on one line at the same width, so the success path is now the one that wraps.

**Instead.** Keep the three facts; drop the backticks, or put `reaches it: lop --resume <id>` so the command is the last thing on the first line at 100. Not blocking — the wrap is still readable and the id is on screen.

### U12 — NIT — a named fork is unfindable by the word that created it

**Step.** After the fork names itself, type `fork` in `/resume`. `0 of 4` / `no session matches that filter`. Correct given the tag is gone — the filter's invariant is "what is on screen is matchable", and `[fork]` is no longer on screen.

**What a user would expect, weakly.** "The branch I made this morning" is still a fork in their head even after it has a title. The docs now say the tag is searchable for forks *still wearing an inherited title*, which is accurate. A user who internalised "type fork" from the unnamed state will think the filter broke.

I would not put ancestry into the haystack (that re-opens the twin-row problem as a search problem). Worth a picker empty-state hint only if this shows up in use. Recording it so the next round does not rediscover it.

---

## Closed from the spec, as interaction

| Spec | Verdict |
|---|---|
| U1 prefix survives truncation | **closed** — walked at 80 and 70 with a 56-char title |
| U2 tab title prefix | **closed** on the title writer; not watched in a real window switcher |
| U3 filter matches the mark | **closed** — typed `fork` into the real picker |
| U4 mobile mark | **not walked live**; daemon + renderer + vitest look right |
| U5 opened receipt | **closed** as copy; wrap is U11 |
| U6a esc in the receipt | **closed** |
| U6b standing band | **closed** as presence; residual is U10 |
| U7 switch receipt persistence | **still deferred** — flush copy is correct; rebuild not watched |
| U8 docs "costs a turn" | **closed** — `docs/fork.md` now says the argument spends a real model call |
| U9 aside vs `/fork` | **no change requested** — `^F` still names both meanings |

---

## Verdict

**TERMINAL.** No BLOCKER, no MAJOR. U10 is the only leftover friction that a user will actually hit (Esc discoverability after the dim line scrolls), and it is a polish on a standing indicator that already exists and already works. U11 and U12 are nits.

**Merge recommendation:** merge. Do not hold the round for U10; if it is cheap, fold it into this PR, otherwise leave it as `deferred — band already says the fork is armed`.

Counts this round: 0 blocker, 0 major, 1 minor (U10), 2 nits (U11, U12). Cap of 5 minor / 5 nit observed.

damianvtran:

Implementer: coder (zai/glm-5.3)
Scope: efbf3c3ffb8b6eac957d9df9c99d9db88b103838..e9d2197449a3dacb8aecdc74a9e856d79279f142

### R1 — MAJOR — `/rename` does not clear the live `[fork]` tab marker

**Fixed** in e9d2197449a3dacb8aecdc74a9e856d79279f142.

Every path that sets a conversation name now re-reads the tag in the same `status.update()`:

- `_cmd_rename` pushes `forked=bool(getattr(session, "wears_inherited_title", False))` beside `conversation_name=`, mirroring `_store_title`. `getattr` for the same defensive reason `_adopt_session` uses it — a reduced facade need not expose the property.
- `_apply_frontend_state` had the same hole and is covered too: it now consumes a new `conversation_title_forked` field on `FrontendSessionState`, emitted by `refresh_from_session` from the session's `wears_inherited_title`. `extra="allow"` on the model keeps older readers tolerant of the new field, and `getattr` at both ends keeps a reduced facade or an older owner's snapshot from failing the refresh. So a later snapshot heals the tab, exactly as the finding asked.

Regression test: `TestTheWindowTitleNamesTheFork.test_a_live_rename_clears_the_tab_mark` in `tests/unit/test_fork.py` — drives the REAL `_cmd_rename` (not `set_conversation_name` alone, which is the gap the prior test left) over a REAL `Session` built on a REAL forked directory, with a REAL `TerminalTitle` writer, and asserts the rendered OSC title.

**Fail-without-fix verification:** stash of the source fix with the test kept, run against the unfixed tree:

```
'[fork]' is contained here:
  lo › [fork] Streaming parser attempt
?      ++++++
1 failed
```

Restored, it passes and asserts `lo › Streaming parser attempt` exactly.

### R2 — MINOR — cache-key test docstring overclaims necessity

**Fixed** in the same commit. The docstring cited run B's controls (inherited 16/16 vs wrong 3/8) as if they were the result. It now claims what `docs/evidence/fork-cache/results.txt` actually establishes: the inherited key read the parent's prefix in every trial of every run (10/10 canonical, never a miss across four runs and a dozen probes) — sufficiency, never a cost — while explicitly declining the necessity claim, naming the unstable controls (wrong-key 3/8 to 10/10) and the content-addressed reason. The plumbing tests are untouched.

damianvtran:

Implementer: coder (zai/glm-5.3)
Scope: efbf3c3ffb8b6eac957d9df9c99d9db88b103838..e9d2197449a3dacb8aecdc74a9e856d79279f142

No findings to remediate — the round was TERMINAL with 0 blocker / 0 major / 0 minor / 0 nit.

One note for completeness: the only visual-adjacent change since the reviewed head is the pending-fork band segment now reading `⎇ forking · esc` instead of `⎇ forking` (U10, UX round). Same segment, same `label` violet, same slot in the drop ladder; four extra cells on a rung the ladder sheds late. The truncation, weight, alignment, receipt, and picker surfaces the round judged are unchanged.

damianvtran:

Implementer: coder (zai/glm-5.3)
Scope: efbf3c3ffb8b6eac957d9df9c99d9db88b103838..e9d2197449a3dacb8aecdc74a9e856d79279f142

### U10 — MINOR — after the deferred acknowledgement scrolls away, Esc is no longer discoverable from what remains

**Fixed** in e9d2197449a3dacb8aecdc74a9e856d79279f142 — the reviewer's first suggested shape, which it called one word short of an instruction: the standing band segment now reads `⎇ forking · esc` (`FORK_PENDING_TEXT` in `status_line.py`). Cheap exactly as predicted — a string constant, no layout change, same `label` violet, same slot, and the ladder still keeps the segment at 60 cols with room for the four extra cells. The dim acknowledgement stays `info`; the durable indicator now carries the affordance itself, so the choice between "band names the key" and "receipt becomes note" went to the one that outlives the scroll. Existing band tests assert on `FORK_PENDING_TEXT`, so they cover the new text unchanged.

### U11 — NIT — the opened receipt wraps at 100 columns, splitting the resume command from its id

**Rejected.** The wrap keeps all three facts (what/where/not-focused + how-to-reach) with the id visible on both lines, and the reviewer itself noted it is still copyable and not blocking. The alternative phrasings each cost something the current copy buys: dropping the backticks loses the visual "this is a command" marker that matches the fallback receipt beside it, and reordering to put the command last merely moves the wrap rather than removing it — at 100 cols the three facts plus a 12-hex id do not fit one line in any order. The success path wrapping once at exactly 100 is a narrower cost than degrading the copy at every width.

### U12 — NIT — a named fork is unfindable by the word that created it

**Rejected** — on the reviewer's own reasoning. The filter's invariant is "what is on screen is matchable," the tag is correctly gone once the fork has a title, and the reviewer explicitly declined to put ancestry into the haystack ("that re-opened the twin-row problem as a search problem") and suggested an empty-state hint "only if this shows up in use." Nothing has shown up in use; adding the hint now would be speculative copy on a surface the round judged as working.

---

One interaction-adjacent change since the walked head, for completeness: `/rename` on a live unnamed fork now clears the `[fork]` tab mark in the same frame (R1, agent round). That closes the disagreement the walk documented between the live tab and the picker — the tab now matches the disk behaviour the walk verified. No new keys, no flow change.

damianvtran:
### Agent review — round 2

Scope: b6a419cc..e9d2197449a3dacb8aecdc74a9e856d79279f142

Verification pass on the R1 remediation only. Read-only worktree `/tmp/lop-rev2-477` off `origin/dev-fork-ux`; head confirmed via `gh pr view 477 --json headRefOid`. Diff read first; everything below was reproduced by execution on this head, not inferred. Worktree removed; branch untouched.

**Clean. TERMINAL.** The MAJOR is fixed on both paths it named, the regression test genuinely guards the hunk, and every hold from round 1 survives. No new finding of any severity in `b6a419cc..e9d21974`.

### R1 verification — `/rename` clears the live `[fork]` tab mark

**Fixed, by execution.** A REAL `OperatorApp` (`CSS_PATH: local_operator.tcss` confirmed on the host) over a REAL fork of a REAL named parent, driven through the typed-command entry point `app._run_slash_command("/rename Streaming parser attempt")` — not the handler directly — with the app's own `TerminalTitle` writer:

```
tab title BEFORE rename: 'lo › [fork]'
tab title AFTER rename:  'lo › Streaming parser attempt'
```

No intermediate repaint between the two: one `status.update(conversation_name=, forked=)` call paints band and tab in the same frame, exactly as the fix's comment claims. `fork.wears_inherited_title` is `False` after.

**Mutation check — the regression test is real.** With ONLY the `_cmd_rename` hunk reverted in the worktree (`forked=` dropped from that one call, `_apply_frontend_state` and everything else at head), `test_a_live_rename_clears_the_tab_mark` fails with exactly the R1 symptom:

```
AssertionError: the tab still announces a borrowed title after /rename:
'lo › [fork] Streaming parser attempt'
  '[fork]' is contained here:
    lo › [fork] Streaming parser attempt
  ?      ++++++
```

Restored, it passes and asserts the exact string. The test guards the hunk, not the fixture.

### The `_apply_frontend_state` half — emitted everywhere, degrades safely

`conversation_title_forked` is emitted at the one producer of `conversation_title` (`FrontendStateStore.refresh_from_session`, `frontend_state.py:1155`), so it cannot be missing anywhere a name is emitted — the grep for `conversation_title` across `local_operator/` finds no second producer. `extra="allow"` on the model plus `getattr(state, ..., False)` at the consumer cover the older-producer case, verified by execution:

- **New producer, unnamed fork** → snapshot sets the mark: `lo › [fork]`.
- **New producer, fork just named** → snapshot clears mark and name together: `lo › Streaming parser attempt`.
- **Old producer (field absent, `model_validate` on a base-shaped dict)** → no raise, no phantom field in `model_extra`, name arrives, mark cleared. `False` is the right default here, not a regression: on base this push carried no `forked=` at all and a stale mark survived it; clearing is strictly closer to the disk truth the picker shows.
- **Bare object with only the old fields** → same, no raise.

### Holds re-verified on this head

- **Picker I/O.** `TestThePickerPaysNothingForSessionsThatAreNotForks` passes on `e9d21974`; the counting harness still pins `≤1 read/row` and `origin.json` absent from the per-row set. The remediation touches no picker code.
- **Truncation.** Real store rows (a real fork of a real named parent, not the synthetic `SessionRow`s): at 80 `❯ [fork] Refactor the YAML loader to stream anchors ins… just now …`; at 70 `❯ [fork] Refactor the YAML loader to stream a…` — mark present, parent unmarked, rows distinct, names left-aligned.
- **Filter.** `filter_rows(..., "fork")` → the fork only; `"[fork]"` → the fork only.
- **cache_lineage_id vs sticky session_id.** Untouched by this diff and still separated at the source: `configure.py:1628-1635` documents and implements it, `session_factory.py:1329` passes `cache_lineage_id=fork_parent(...) or None`, `pin_session_credential`/`get_api_key` key on `_session_id` only.
- **Suites on this head.** `test_fork.py` 69 passed; `test_status_line.py` 78; `test_session_picker.py` 79; `test_terminal_title.py` + `test_frontend_state.py` + `test_frontend_sync.py` + `test_reconnect_parity.py` 60; targeted pilot rename/snapshot tests 9. All `-n 4`.

### Dispositions sanity-checked

- **R2 (docstring)** — agree, fixed. The rewrite claims exactly what `docs/evidence/fork-cache/results.txt` lines 71/88/98-102 establish: inherited never missed in any run (10/10 canonical, 6/6, 8/8, 4/4, 10/10, 10/10), controls unstable (3/8→10/10), necessity explicitly declined. No overclaim remains.
- **U10 (`⎇ forking · esc`)** — agree. Measured through the real `_render` ladder on this head: segment kept at 100/80/70/60/50/**40**, first dropped at 30. The remediation's "survives at 60" claim is true with margin.
- **U11 (receipt wrap), U12 (named fork unfindable by "fork")** — agree with both rejections, on the reviewers' own stated reasoning; nothing new to add.

### Counts

0 blocker, 0 major, 0 minor, 0 nit.

**Merge recommendation: merge.** The round is terminal on my side; CI (lint green, test/type-check/audit pending at time of review) is the only thing left to confirm before you do.

PR #478, opened 2026-08-31
ci: stop disarming the freeze guard when test is red

### LINKED ISSUE #381: chore(pyright): pin one version across the gate and stop restating upstream defaults
Follow-up to #380 (agent review finding F1).

## Problem

#380 fixed a regression where `[tool.pyright] exclude` was set to `["docs"]`, which REPLACES pyright's built-in defaults rather than extending them. The fix restates them explicitly:

```toml
exclude = ["**/node_modules", "**/__pycache__", "**/.*", "docs"]
```

That is correct for pyright 1.1.411 (verified against upstream `docs/configuration.md` and the literal in the shipped `pyright-internal.js`), but it hardcodes an upstream constant. If a future pyright release adds a fourth default, this config silently stops honouring it — the same silent-and-slow failure class #380 fixed, where CI stays green and only local runs degrade.

## Compounding: three pyright versions across one gate

- `.github/workflows/ci.yml` lint job pins `pyright==1.1.408`
- `.github/workflows/ci.yml` type-check job sets `PYRIGHT_PYTHON_FORCE_VERSION: latest`
- Local `.venv` resolves 1.1.411 / 1.1.413

Version-specific constants now live in `pyproject.toml` while the version producing them is unpinned and inconsistent.

## Options

1. **Pin one pyright version** across lint, type-check, and the dev extra, so the restated defaults are checked against a known version.
2. **Remove the need to restate defaults** by scoping the gate to explicit roots (`pyright local_operator tests scripts`) so `exclude` no longer has to carry the whole tree. Note AGENTS.md currently warns against narrowing the gate command, for reasons that apply to isort/black — that guidance needs revisiting deliberately, not incidentally.
3. **Add a guard test** asserting the configured excludes are a superset of the running pyright's defaults, failing loudly on drift.

(3) is the most direct answer to the drift risk and composes with (1).

## Not urgent

The restated list is exact for the versions in use today, so this is future-proofing, not a live defect.

### LINKED ISSUE #423: Stale .venv console-script shebangs silently pass lint/format gates (rc=126 swallowed in pipelines)
The console scripts in the repo's `.venv` have shebangs pointing at **deleted** worktrees, so they fail with `bad interpreter` and exit **126**:

```
.venv/bin/black  -> #!/private/tmp/lop-resume-team/.venv/bin/python
.venv/bin/flake8 -> #!/private/tmp/lop-unified-session-state/.venv/bin/python
.venv/bin/isort  -> #!/private/tmp/lop-freeze-fix/.venv/bin/python
```

This is dangerous rather than merely broken. `rc=126` disappears inside a pipeline — the common `... | tail` idiom in a gate script swallows it — so a lint or format gate that never actually ran reports as passing. AGENTS.md already warns that narrowing these commands checks something CI does not; this is a second, quieter way to get a false green locally.

Both the reviewer and the implementer on #402 hit this and worked around it by invoking the tools as `python -m <tool>`, so no gate on that PR was masked. But every agent and human on this machine is exposed to it.

**Fix:** `uv pip install --force-reinstall` those three into the venv, and/or standardise on `python -m flake8` / `python -m isort` and `uvx --from black==26.1.0 black` in the documented gate commands so a stale shebang cannot silently pass.

The shebangs point at worktrees created by parallel agent sessions, so this will recur whenever such a worktree is removed. Worth a note in AGENTS.md alongside the existing gate guidance.

Found during the review of #402.

### LINKED ISSUE #428: ci: tui-e2e freeze guard is disarmed whenever `test` is flaky (`needs: test`)
Split out of the round-1 agent review of #426 (finding C4), where it was
observed with real consequence: the one stage purpose-built to catch an
event-loop freeze gave **zero signal on a PR that fixed an event-loop freeze**.

## The problem

`tui-e2e` in `.github/workflows/ci.yml` declares:

```yaml
needs:
  - lint
  - type-check
  - test
  - pip-audit
```

`test` is currently flaky on `main` — `test_the_loop_stays_responsive_while_several_subagents_run`
is a timing-sensitive loop-stall assertion that fails on unmodified `main`
(e.g. run 33279074013). Because `tui-e2e` needs `test`, every red `test` run
**skips** the freeze guard rather than running it.

On #426 that happened on both CI runs (33284918639, 33284954232): `tui-e2e`
reported `skipping`, so the deadlock guard never executed on a deadlock fix.

This is structurally backwards. Per AGENTS.md, the macOS leg of this stage is
"the only platform where the freeze guard can go red at all" — it exists
precisely because a fully green unit suite did not catch #401. Gating it behind
the unit suite means the guard is disarmed exactly when the suite is unreliable,
and a freeze regression could merge with the guard silently skipped.

## Why it is wired this way

The `needs` clause is presumably there to avoid burning a macOS runner when
cheaper checks already failed. That is a reasonable cost instinct, but it trades
away the signal the stage exists for.

## Options

1. **Drop `test` from `needs`** (keep `lint`/`type-check` as cheap syntax gates).
   The stage is ~11s healthy, headless, needs no API key and no network — the
   runner cost is small next to a freeze reaching `main`.
2. **Drop `needs` entirely** so it runs fully independently.
3. **Keep `needs` and fix the flake first**, accepting the guard stays disarmed
   until then. Weakest option: it leaves the gap open and depends on a separate
   fix landing.

Option 1 looks right, but this changes CI topology for every PR in the repo, so
it wants its own change rather than riding along inside an unrelated bug fix.

## Interim evidence for #426

The gap was closed manually for that PR — the stage was run locally on macOS at
head, both by the implementer and independently by the reviewer:

```
3 passed in 10.11s
```

## Related

- #426 (round-1 review finding C4)
- #401 / 80df237b — the freeze this stage was built to catch
- The `test` flake itself is a separate pre-existing issue

damianvtran:
# PR Description

## Change classification

**C2 — standard / pre-authorized.** CI topology, a pinned dev-extra, Makefile
invocation, and a unit guard. No runtime path under `local_operator/` changes.
Rollback is `git revert`.

## Summary of Changes

Three CI/tooling hygiene defects, one silent-green shape: a gate that looks
passing because it never actually ran, or ran the wrong tool.

### 1. `#428` — `tui-e2e` freeze guard disarmed whenever `test` is red

`tui-e2e` declared `needs: [lint, type-check, test, pip-audit]`. GitHub Actions
**skips** a job whose needed jobs failed — it does not wait politely. A skipped
freeze guard is indistinguishable from a passing one on the PR checks list.

This is not theoretical. On PR #426 — which *fixed* an event-loop freeze —
both CI runs (33284918639, 33284954232) reported `tui-e2e` as `skipping`
because `test` was flaky, so the deadlock guard never executed on a deadlock
fix. Per AGENTS.md the macOS leg of this stage is the regression guard for
#401 (a blocking `flock` in the MCP OAuth refresh lock deadlocking the event
loop on `/resume`), and it is deliberately fork-gate-free so it runs on every
PR. A guard that silently disarms when an unrelated stage is flaky is
structurally backwards.

PR #461 (`f5f77cee`) landed test-determinism fixes that reduce `test`
flakiness. That does **not** make this issue moot: the structural dependency
is still wrong and will disarm the guard on the next flake.

**Fix:** `tui-e2e` now needs only `lint` and `type-check` (cheap syntax gates;
a broken install is not a freeze). `test` and `pip-audit` are dropped —
`pip-audit` has the same latent disarm (a CVE published today would skip the
macOS resume-liveness assertion on an unmodified tree). The job installs the
package itself, so it does not need those stages to have built anything.
`cli-sanity` / `server-sanity` keep the full `needs` list — they are live-LLM
cost, not freeze guards.

```
before:  lint, type-check, test, pip-audit  ──needs──►  tui-e2e
after:   lint, type-check                   ──needs──►  tui-e2e
         test, pip-audit                    (independent; a red does not skip the guard)
```

Failure modes that now still run the freeze guard: unit-suite flake, unit-suite
real failure, pip-audit CVE, pip-audit infra blip. Failure modes that still
skip it (intentionally): lint red, type-check red — those are "the package
does not even parse / type-check", not a freeze.

### 2. `#423` — stale `.venv` console-script shebangs silently pass lint/format gates

The console scripts in the repo `.venv` have shebangs pointing at **deleted**
worktrees, so they fail with `bad interpreter` and exit **126**. `rc=126`
disappears inside a pipeline — `cmd | tail` reports `tail`'s 0 — so a lint or
format gate that never actually ran looks green.

Reproduced on this machine at HEAD, against the real venv:

```
$ .venv/bin/black --version; echo "black rc=$?"
/bin/sh: .venv/bin/black: /private/tmp/lop-resume-team/.venv/bin/python: bad interpreter: No such file or directory
black rc=126

$ .venv/bin/black --version | tail -1; echo "piped rc=$?"
piped rc=0

$ .venv/bin/python -m flake8 --version >/dev/null; echo "python -m flake8 rc=$?"
python -m flake8 rc=0
$ uvx --from black==26.1.0 black --version >/dev/null; echo "uvx black rc=$?"
uvx black rc=0
```

**Fix:** `make lint` / `make format` / `make type-check` now go through
`python -m` / `uvx` at the versions CI pins, matching the documented gate
commands in AGENTS.md. A note in AGENTS.md names the shebang trap so the next
agent does not "simplify" back to `.venv/bin/black`.

### 3. `#381` — pin one pyright version and stop restating upstream defaults by hope

Three pyright versions across one gate:

| surface | version |
| --- | --- |
| lint job (`pip install pyright==1.1.408`) | 1.1.408, **never invoked** |
| type-check job (`PYRIGHT_PYTHON_FORCE_VERSION: latest`) | npm 1.1.413 |
| local `.venv` / pip-resolved extra | 1.1.411 |

`[tool.pyright] exclude` *replaces* pyright's built-in defaults rather than
extending them, so the restated `"**/node_modules"`, `"**/__pycache__"`,
`"**/.*"` are version-specific constants. Three versions means the restatement
is a guess. Dropping `**/.*` makes pyright follow the `.venv` symlink and
type-check all of site-packages (29466 files / ~30 min instead of ~636 files /
~15 s) — CI never creates a `.venv`, so it stays green while every local run
of the gate becomes unusable.

**Fix:**

- Pin `pyright==1.1.411` in the dev extra (what CI currently resolves from
  PyPI; measured identical to 1.1.413 on this tree: 636 files, 0 errors).
- Drop the unused `pyright==1.1.408` from the lint job.
- Drop `PYRIGHT_PYTHON_FORCE_VERSION: latest` from type-check so CI uses the
  bundled analyzer, not whatever npm currently calls latest.
- Guard: `tests/unit/test_ci_hygiene.py` asserts `[tool.pyright] exclude` is a
  **superset of the running analyzer's defaults**, sourced behaviourally from
  `pyright --verbose` (`Auto-excluding <pattern>`). If a future pyright adds a
  fourth default, the test goes red until we restate it.

Pyright file count / result, same tree, same invocation
(`.venv/bin/python -m pyright --pythonpath .venv/bin/python .`):

| | version | filesAnalyzed | errors |
| --- | --- | --- | --- |
| before pin (this tree at origin/main) | 1.1.411 | 636 | 0 |
| after pin + new test file | 1.1.411 | 637 | 0 |

The +1 is `tests/unit/test_ci_hygiene.py`. Not the 29466-file site-packages
walk. `**/.*` is still in `exclude`.

### Mutation tests (each guard dies at its own boundary)

Run against a scratch copy of the tree; the worktree itself is unmodified.

| mutation | test that died |
| --- | --- |
| restore `tui-e2e needs: [lint, type-check, test, pip-audit]` | `test_tui_e2e_does_not_need_the_unit_suite_or_pip_audit` |
| drop `**/.*` from `[tool.pyright] exclude` | `test_configured_pyright_excludes_are_a_superset_of_the_running_defaults` |
| unpin `pyright` in the dev extra | `test_dev_extra_pins_the_same_pyright_the_type_check_job_installs` |
| restore `PYRIGHT_PYTHON_FORCE_VERSION: latest` | `test_type_check_does_not_force_latest_pyright` |
| restore unused lint-job `pyright==1.1.408` | `test_type_check_does_not_force_latest_pyright` |
| `Makefile` `lint:` back to bare `flake8` | `test_makefile_quality_targets_do_not_invoke_console_scripts` |
| drop `macos-latest` from the `tui-e2e` matrix | `test_tui_e2e_still_runs_on_macos` |

The 126-pipeline assertion is itself a reproduction: a fake console script
with a missing-interpreter shebang exits non-zero, `| tail` reports 0, and
`python -m flake8` still succeeds with that fake on `PATH`.

### Quality gates (whole tree, repo config, exit codes)

```
.venv/bin/python -m flake8 .                         flake8 rc=0
uvx --from black==26.1.0 black --check .             black rc=0
uvx isort==5.13.2 --check .                          isort rc=0
                                                     pyright rc=0  (0 errors, 637 files)
  .venv/bin/python -m pytest tests/unit/test_ci_hygiene.py -n0 -q
                                                     8 passed
make lint                                            make lint rc=0
                                                     (.venv/bin/python -m flake8 .)
```

isort printed `Skipped 1 files` on both pass and fail last wave; the rc is
the signal. black's "All done!" banner is on stderr.

### DAG evidence for #428

This is a workflow-topology change. I did not push a deliberately-red `test`
branch (four sibling agents are pushing at the same time; a broken workflow
blocks all of them). The `needs` graph is the behaviour under test, and it is
pinned by `test_tui_e2e_does_not_need_the_unit_suite_or_pip_audit` so a
drive-by copy of the sanity-job `needs` list cannot recouple them.

## Out of scope / deliberately not done

- Did not drop `needs` from `cli-sanity` / `server-sanity`. Those are live-LLM
  cost; a flake skipping them is annoying, not a freeze reaching `main`.
- Did not `uv pip install --force-reinstall` the stale console scripts in the
  shared `.venv`. That is a local-machine repair, not a repo change, and it
  would recur the next time a worktree is deleted. The invocation change is
  the durable fix.
- Did not bump the version. The manager runs one release after PRs land.
- Did not merge.

Closes #381
Closes #423
Closes #428

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install flake8==7.1.0 black==26.1.0 isort==5.13.2 pyright==1.1.408
+          pip install flake8==7.1.0 black==26.1.0 isort==5.13.2
 
       - name: Run flake8
         run: flake8 .
@@ -39,8 +39,12 @@ jobs:
 
   type-check:
     runs-on: ubuntu-latest
-    env:
-      PYRIGHT_PYTHON_FORCE_VERSION: latest
+    # pyright's version is the pin in pyproject.toml's dev extra. Do NOT
+    # set PYRIGHT_PYTHON_FORCE_VERSION: that env var makes the wrapper
+    # ignore the bundled analyzer and download whatever npm currently
+    # calls latest, which is how this job drifted onto 1.1.413 while the
+    # package (and every local run) stayed on 1.1.411. Bump the pin, not
+    # an env override.
     steps:
       - uses: actions/checkout@v4
 
@@ -280,11 +284,25 @@ jobs:
     # compositor, so there is no window, no display server and no TTY beyond
     # what the harness makes itself. It needs no `xvfb` and must never grow a
     # real browser or terminal dependency.
+    #
+    # lint + type-check only. This job is the freeze regression guard (#401);
+    # it exists BECAUSE a green unit suite did not catch a total event-loop
+    # deadlock. Gating it on `test` (or `pip-audit`) disarms the guard
+    # whenever that cheaper stage is red — which is exactly when you most
+    # want the freeze signal. Observed on PR #426: both CI runs skipped
+    # tui-e2e because `test` was flaky, so a deadlock fix landed with the
+    # deadlock guard never executed. PR #461 reduced that flake; it does
+    # not make the structural dependency right.
+    #
+    # lint/type-check stay: they are cheap syntax gates and a broken
+    # install is not a freeze. `test` and `pip-audit` do not: a flake or a
+    # newly-published CVE must not skip the macOS resume-liveness
+    # assertion. The job installs the package itself, so it does not need
+    # those stages to have built anything. cli-sanity/server-sanity keep
+    # the full `needs` list — they are live-LLM cost, not freeze guards.
     needs:
       - lint
       - type-check
-      - test
-      - pip-audit
     # Generous relative to the ~11 s the stage takes when healthy (~14 s on a
     # macOS runner, which is slower and queues longer). The failure mode under
     # test is a hang, and each test carries its own tighter `faulthandler`
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -110,6 +110,13 @@ stage because boot, the write-turn artifacts and the transcript replay are
 platform-neutral, but **do not drop the macOS leg** — without it this stage goes
 green against the exact commit it exists to catch.
 
+The `tui-e2e` job must not `need` the unit `test` job or `pip-audit`. It exists
+because a green unit suite did not catch #401; gating it on `test` skipped the
+freeze guard on every red unit run (observed on PR #426, which *fixed* a
+deadlock while `tui-e2e` reported `skipping`). It needs only `lint` and
+`type-check` — cheap syntax gates. A flake or a newly-published CVE must not
+disarm the macOS resume-liveness assertion.
+
 Gates, all of which must be clean before a PR. **Run them over the whole tree,
 exactly as CI does** — these are the commands from `.github/workflows/ci.yml`:
 
@@ -119,6 +126,16 @@ uvx --from black==26.1.0 black --check .
 uvx isort==5.13.2 --check .
 .venv/bin/python -m pyright --pythonpath .venv/bin/python .
 ```
+
+Do **not** invoke `.venv/bin/black`, `.venv/bin/flake8`, `.venv/bin/isort`, or
+`.venv/bin/pyright` directly. Those console scripts carry a shebang baked in
+at install time; after a parallel-agent worktree that owned the venv is
+deleted they fail with `bad interpreter` and exit **126**. `rc=126` is
+swallowed by a pipeline (`cmd | tail` reports `tail`'s 0), so a lint or
+format gate that never actually ran looks green. `python -m` and `uvx`
+cannot hit that path — that is why the commands above are spelled that way,
+and why `make lint` / `make format` / `make type-check` go through them
+rather than the console scripts.
 
 Narrowing the last two to `local_operator tests`, or passing `--profile black`
 instead of letting isort read the repo's own config, checks something CI does
--- a/Makefile
+++ b/Makefile
@@ -95,18 +95,26 @@ coverage: ## Generate test coverage report
 # ============================================================================
 # Code Quality Commands
 # ============================================================================
-# Format code with black and isort
+# Invoke formatters/linters the same way AGENTS.md documents the gates.
+# Bare `.venv/bin/black` (and flake8/isort/pyright) console scripts carry a
+# shebang baked in at install time; after a worktree that owned the venv
+# is deleted they fail with `bad interpreter` and exit 126. `rc=126`
+# disappears inside a pipeline (`cmd | tail` reports tail's 0), so a gate
+# that never ran looks green. `python -m` uses the interpreter's module
+# path; `uvx` fetches a pinned tool into an isolated cache. Neither reads
+# those shebangs. black/isort are deliberately NOT in the dev extra, so
+# format goes through uvx at the versions CI pins.
 format: ## Format code with black and isort
-	black .
-	isort .
+	uvx --from black==26.1.0 black .
+	uvx isort==5.13.2 .
 
 # Run linting with flake8
 lint: ## Run linting with flake8
-	flake8
+	.venv/bin/python -m flake8 .
 
 # Run type checking with pyright
 type-check: ## Run type checking with pyright
-	pyright
+	.venv/bin/python -m pyright --pythonpath .venv/bin/python .
 
 # Run security audit with pip-audit
 security: ## Run security audit with pip-audit
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -130,7 +130,17 @@ dev = [
     # accepts, or accept code CI flags, and either way the local gate stops
     # predicting the remote one. Bump this and the ci.yml pin together.
     "flake8==7.1.0",
-    "pyright",
+    # Same discipline as flake8: one version across the type-check job,
+    # the local gate, and this extra. Unpinned, the wrapper
+    # (PYRIGHT_PYTHON_FORCE_VERSION=latest used to live on the type-check
+    # job) downloaded whatever npm called latest (1.1.413) while pip
+    # resolved 1.1.411 and the lint job — which never even ran pyright —
+    # pinned 1.1.408. The restated `[tool.pyright] exclude` defaults are
+    # version-specific; three versions means the restatement is a guess.
+    # Bump this pin, not an env override, and keep the exclude-superset
+    # guard in tests/unit/test_ci_hygiene.py green against the new
+    # analyzer.
+    "pyright==1.1.411",
     "pytest",
     "pytest-asyncio",
     "pytest-cov",
--- a/tests/unit/test_ci_hygiene.py
+++ b/tests/unit/test_ci_hygiene.py
@@ -1,0 +1,283 @@
+"""Guards for CI topology and local-gate invocation.
+
+These exist because three defects all produced a *silent* green:
+
+- #428: ``tui-e2e`` needed ``test``, so a flaky unit suite skipped the freeze
+  guard (observed on PR #426, which *fixed* a deadlock while the deadlock
+  job reported ``skipping``).
+- #423: a stale ``.venv/bin/black`` shebang exits 126, and ``cmd | tail``
+  reports ``tail``'s 0, so a lint gate that never ran looks passing.
+- #381: ``[tool.pyright] exclude`` *replaces* pyright's built-in defaults.
+  Dropping ``**/.*`` makes a local run type-check all of site-packages.
+
+A comment in ci.yml is not a test. Each assertion below is mutation-tested
+against the defect it claims to catch.
+"""
+
+from __future__ import annotations
+
+import os
+import stat
+import subprocess
+import sys
+import tempfile
+import tomllib
+from pathlib import Path
+from typing import Any
+
+import yaml
+
+REPO = Path(__file__).resolve().parents[2]
+CI_YML = REPO / ".github" / "workflows" / "ci.yml"
+PYPROJECT = REPO / "pyproject.toml"
+MAKEFILE = REPO / "Makefile"
+
+
+def _ci_jobs() -> dict[str, Any]:
+    jobs = yaml.safe_load(CI_YML.read_text())["jobs"]
+    assert isinstance(jobs, dict)
+    return jobs
+
+
+def _needs(job: str) -> set[str]:
+    """`needs` of a CI job, as a set. A missing key is an empty set, which
+    is a legitimate topology (no dependencies), not an error."""
+    declared = _ci_jobs()[job].get("needs") or []
+    assert isinstance(declared, list)
+    return set(declared)
+
+
+def _steps(job: str) -> list[dict[str, Any]]:
+    steps = _ci_jobs()[job]["steps"]
+    assert isinstance(steps, list)
+    return steps
+
+
+def test_tui_e2e_does_not_need_the_unit_suite_or_pip_audit() -> None:
+    """The freeze guard must still run when `test` is red.
+
+    `needs: test` is how GitHub Actions *skips* a job, not how it waits
+    politely. A skipped freeze guard is indistinguishable from a passing
+    one on the PR checks list. `pip-audit` has the same shape: a CVE
+    published today would skip the macOS resume-liveness assertion on an
+    unmodified tree. lint/type-check stay — they are cheap syntax gates
+    and a broken install is not a freeze.
+    """
+    needs = _needs("tui-e2e")
+    assert "test" not in needs, (
+        "tui-e2e needs `test`, so a flaky unit suite skips the freeze "
+        "guard (the #428 defect, observed on PR #426)"
+    )
+    assert "pip-audit" not in needs, (
+        "tui-e2e needs `pip-audit`, so a newly-published CVE skips the "
+        "freeze guard — the same latent disarm as `needs: test`"
+    )
+    assert needs == {"lint", "type-check"}, (
+        f"tui-e2e.needs={sorted(needs)!r}; expected only lint and "
+        "type-check (cheap syntax gates the job does not install itself)"
+    )
+
+
+def test_cli_and_server_sanity_still_wait_on_the_cheap_gates() -> None:
+    """Live-LLM jobs are cost, not freeze guards; they keep the full needs.
+
+    Dropping `test` from those too would be a different change. Pin the
+    current contract so a drive-by edit of every `needs:` block at once
+    cannot silently re-couple tui-e2e by copying the sanity list.
+    """
+    expected = {"lint", "type-check", "test", "pip-audit"}
+    for name in ("cli-sanity", "server-sanity"):
+        assert _needs(name) == expected, (
+            f"{name}.needs drifted; live-LLM jobs are supposed to keep "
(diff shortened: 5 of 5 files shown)


damianvtran:
### Agent review — round 1
Scope: efbf3c3ffb8b6eac957d9df9c99d9db88b103838..64a870e6c59ef8d8414b8364d01a68cbeb58436d

Independent review of #478 (`dev-ci-hygiene`). Diff first, then the three issues, then cheap reproduction. No BLOCKER, no MAJOR.

## What I independently confirmed

### #428 — freeze guard no longer waits on `test`

Base `tui-e2e.needs` was `[lint, type-check, test, pip-audit]`. Head is `[lint, type-check]`. `cli-sanity` / `server-sanity` still carry the full list. `tui-e2e` still installs the package itself (`pip install -e ".[dev]"`) and still has the `[ubuntu-latest, macos-latest]` matrix with `fail-fast: false` and no fork gate.

GitHub Actions **skips** a job whose needed jobs failed; it does not wait. Dropping `test`/`pip-audit` is therefore the actual behaviour change, not a comment. Live CI on this PR is the behaviour under test:

- push run `33343436574`: `tui-e2e (ubuntu-latest)` SUCCESS 00:07:07–00:07:50 and `tui-e2e (macos-latest)` SUCCESS 00:07:49–00:08:26 while `test (3.12)` and `test (3.13)` were still `IN_PROGRESS`.
- pull_request run `33343438792`: `tui-e2e (macos-latest)` SUCCESS 00:08:28–00:09:06 while both `test` legs were still in flight.

That is exactly the topology #428 asked for. A red unit suite or a pip-audit CVE can no longer skip the macOS resume-liveness assertion. A broken tree can still run e2e — acceptable, and the issue's point: this stage exists because a green unit suite did not catch #401. Failure modes that still skip it (lint red, type-check red) are the cheap "does not even parse" gates the PR kept on purpose.

`test_tui_e2e_does_not_need_the_unit_suite_or_pip_audit` dies when the old `needs` list is restored (mutated, 1 failed). `test_tui_e2e_still_runs_on_macos` pins the load-bearing leg.

### #423 — stale shebang 126 is loud at the gate, not swallowed

Reproduced on the real shared `.venv` at base:

```
.venv/bin/black   shebang -> /private/tmp/lop-resume-team/.venv/bin/python
direct rc=126
piped  rc=0     # `cmd | tail` reports tail's 0

.venv/bin/flake8  shebang -> /private/tmp/lop-unified-session-state/.venv/bin/python
direct rc=126
piped  rc=0
```

`PATH=.venv/bin:$PATH flake8 --version` (the base `make lint` recipe) exits 126. Head `make -f <HEAD Makefile> lint` on the same checkout: `.venv/bin/python -m flake8 .`, rc=0. `python -m flake8` and `uvx --from black==26.1.0 black` both rc=0 with the stale scripts still on disk.

The failure is in the right place: `make lint` / `make format` / `make type-check` (the local gates), not a later stage that would also skip. CI lint already invoked `flake8` / `black` / `isort` from a fresh `pip install` on the runner, so it was never on this path; the Makefile was.

### #381 — pin + exclude restatement did not walk site-packages

`exclude` is unchanged: `["**/node_modules", "**/__pycache__", "**/.*", "docs"]`. Dev extra goes from unpinned `pyright` to `pyright==1.1.411`. `PYRIGHT_PYTHON_FORCE_VERSION: latest` is gone from the type-check job; the unused lint-job `pyright==1.1.408` is gone.

Same invocation (`.venv/bin/python -m pyright --pythonpath .venv/bin/python --stats .`), same analyzer 1.1.411:

| tree | files | errors |
| --- | --- | --- |
| origin/main (`efbf3c3f`) | 639 | 0 |
| head (`64a870e6`) | 640 | 0 |

The +1 is `tests/unit/test_ci_hygiene.py`. Not 29466. Dropping `**/.*` from a scratch copy makes `test_configured_pyright_excludes_are_a_superset_of_the_running_defaults` fail on exactly `['**/.*']`. The probe sources defaults from `pyright --verbose` `Auto-excluding` lines against an empty config (verified: `**/node_modules`, `**/__pycache__`, `**/.*`).

Pinning 1.1.411 (the pip-resolved extra) rather than npm `latest` (1.1.413, which FORCE_VERSION used to fetch) is the right pin: the restated defaults are version-specific, and "whatever CI currently resolves" under FORCE_VERSION was the drift #381 named.

### Blast radius

YAML parses. `actionlint` reports only a pre-existing SC2086 on `server-sanity` (unchanged). Job names in `needs:` still exist. No new `if:` on `tui-e2e`. Sibling workflows (`publish.yml`, extension, chrome-web-store) untouched. A syntax error here would fail the workflow, not silently skip — the skip class this PR is fixing is `needs:` of a *failed* job, not a parse error.

Guards: `tests/unit/test_ci_hygiene.py` 8 passed in 0.72s. flake8/black/isort clean on the new file.

## Findings

None at BLOCKER or MAJOR.

**F1 — MINOR** — `tests/unit/test_ci_hygiene.py:128` hardcodes `pyright==1.1.411` as a string equal, so a deliberate bump of the extra pin is a two-file edit (pyproject + this assertion) even though `test_configured_pyright_excludes_are_a_superset_of_the_running_defaults` already behaviourally checks the running analyzer. Acceptable as a belt, but the comment should say "change this assertion in the same commit as the extra pin" more loudly than it does, or derive the expected pin from `pyproject.toml` and only assert CI doesn't override it. Not stop-ship; the current shape is what caught an unpin in the PR's own mutation table.

**F2 — NIT** — `Makefile:121` `security:` is still a bare `pip-audit` console script, same shebang class as #423. Out of scope (the issue named black/flake8/isort), and `make security` is not a required PR gate. Recorded so it is not silently "fixed" by copying the old recipe.

**F3 — NIT** — CI `type-check` still `run: pyright` (console script) after `pip install -e ".[dev]"`. Fine on a fresh runner; locally `make type-check` uses `python -m`. Asymmetry is documented by the Makefile change, not by the workflow step. Harmless.

## Verdict

**clean** — TERMINAL. No blocker, no major. Remaining minors/nits are follow-ups, not a reason for another round. #428/#423/#381 each do what they claim; the freeze guard now runs while `test` is still in flight, which is the load-bearing change.

CI at review time: lint, type-check, pip-audit, both `tui-e2e` legs SUCCESS on both the push and pull_request runs; `test (3.12)` / `test (3.13)` still in flight — which is itself the #428 proof.

damianvtran:

No push. Round 1 is terminal (0 blocker / 0 major); a late commit would invalidate the measurement table the reviewer already took against `64a870e6`.

**F1 — MINOR — rejected.** The hardcoded `pyright==1.1.411` string-equal is the belt that died on unpin in this PR's own mutation table. Deriving the expected pin from `pyproject.toml` would make that assertion tautological for the extra (any pin, including a silent bump that forgot to re-measure the exclude restatement, would pass). The docstring already says to bump the assertion with the extra pin, not via `FORCE_VERSION`. A comment-only rewrite is the scope expansion this round asked us not to take.

**F2 — NIT — deferred.** `make security` is still a bare `pip-audit` console script, same shebang class. Out of scope: #423 named black/flake8/isort, and `make security` is not a required PR gate. Recorded here so it is not silently "fixed" by copying the old recipe.

**F3 — NIT — deferred.** CI `type-check` `run: pyright` is a console script after a fresh `pip install -e ".[dev]"` on the runner, so its shebang is the runner's own interpreter — not the stale-worktree path. Local `make type-check` is the path that hit #423, and that now goes through `python -m`. Changing the workflow step would be a drive-by YAML edit after a terminal round.

Head remains `64a870e6`. Live CI already showed both `tui-e2e` legs SUCCESS while `test (3.12)` / `test (3.13)` were still in flight (push `33343436574`, pull_request `33343438792`) — that is the #428 proof, not a race to wait out.

damianvtran:
### Agent review remediation — round 1 (follow-up)

CI `test (3.12)` on push run `33343436574` failed on **our** test, not a flake:

```
FAILED tests/unit/test_ci_hygiene.py::test_pipeline_does_not_swallow_a_126_from_a_broken_console_script
AssertionError: the fake console script succeeded when found on PATH (rc=0)
stdout='7.1.0 (mccabe: 0.7.0, pycodestyle: 2.12.1, pyflakes: 3.2.0) CPython 3.12.14 on Linux'
```

Linux bash/dash continue PATH search after a bad shebang, so a fake named `flake8` was skipped and the real venv flake8 ran. macOS does not continue, which is why the assertion passed locally on 3.14 and then failed on CI 3.12 by detecting it was not exercising the defect.

**Fixed** in `49ed22b9`: the fake is a unique name (`lop-stale-script`), invoked by **absolute path** through `/bin/bash -c` so PATH continuation cannot skip it, and the boundary assertion is `rc == 126` (not `!= 0`). Mutation: a valid shebang dies with `rc=0` on both 3.12.13 and 3.14.3; inverting the assertion to `== 0` dies with the real 126.

F1/F2/F3 stand as previously answered (rejected / deferred / deferred). No other files touched.

damianvtran:
### Agent review remediation — round 1 (follow-up 2)

CI `test (3.13)` on push run `33344710366` failed on the same test, different errno:

```
AssertionError: the fake did not fail with 126 at its own boundary (rc=127, stderr='... cannot execute: required file not found')
```

Three interpreters, three outcomes for a broken shebang: macOS 3.14 → 126, CI 3.12 PATH-continuation → 0, CI 3.13 absolute path → 127. Pinning 126 is how you get a test that only works on one Python.

**Fixed** in `2c1d54cf`: accept any non-zero from the fake as "the script failed"; assert `| tail` still reports 0. That is the #423 defect. Unique name + absolute path kept so PATH continuation cannot skip the fake.

Reproduced by driving the test function (not pytest collection) on:

| interpreter | result |
| --- | --- |
| 3.14.3 (repo `.venv`) | 8 passed |
| 3.12.13 (`python3.12` throwaway venv) | 8 passed |
| 3.13.12 (`python3.13` throwaway venv) | 8 passed |

Mutation on all three: a valid shebang dies (`the fake succeeded at its own boundary (rc=0)`); inverting the swallow assert dies because piped rc is 0 while direct is non-zero.

Workflow untouched. F1/F2/F3 stand.

damianvtran:
Could not reopen after the remote branch was deleted and recreated (GitHub 422: state cannot be changed). Successor is https://github.com/damianvtran/local-operator/pull/484 — same three commits rebased onto origin/main (`d8a8c833`). Head `498e5e25`. Behaviour unchanged.

PR #452, opened 2026-08-30
fix(tui): guard the steer receipt by session, and stop the settle reflowing the transcript

damianvtran:
# PR Description

## Summary of Changes

Three defects on the queued-steer receipt, deferred from #157 into #160: a
pre-existing race that settles the wrong session's row (F3), a settled string
that drops the deferred case's one distinguishing fact (D5), and the same
settle shrinking the row and moving the transcript under the reader (D1/D2).

### F3 — the receipt had no session guard

`_handle_steering_delivered` carried no guard at all, on the reasoning that a
late event finds nothing held and does nothing. That holds **within** a session
and stops holding at a swap. `/reload` disposes the outgoing controller, but an
event it has already dispatched is a Textual message sitting in the app's queue,
and unsubscribing cannot recall one. It is handled after the swap cleared the
held lists, and by then the user may have steered into the **replacement**
session, whose row is held again — so the dying conversation's drain settles a
row about a message it knows nothing about. The emptiness check cannot catch it
precisely because the lists are non-empty again.

The event now carries the controller that posted it, and the app drops a
delivery whose origin is not the controller it currently listens to.

**Identity, not a generation number.** The other `_handle_*` methods guard with
generations, but a generation cannot answer this hazard: generations are
per-controller counters, so the outgoing controller's number is not comparable
with the incoming one's and can collide with it outright. "Which session is this
from?" is the question actually being asked, and object identity answers it.

### D5 — a fourth string, and the argument for it

A row deferred across a turn promises `sends with your next message`. The user
types that message, and the older row flips to `sent — the agent has it now` —
true, but it drops the association, and the receipt sits *above* the message it
travelled with, so the only remaining cue is spatial and reads backwards.

`SENT_STEER_NOTICE` could not simply be re-worded, which is what forces a fourth
string after #132 spent two rounds arguing the set down to three: it also settles
the **same-turn** case, the common path, where the message went at a tool
boundary during the turn the user was watching. A "next message" the user has
already sent is plainly wrong there. The two settles are genuinely two facts, so
the settle now picks per row from the list membership the handler already tracks.

| constant | string |
|---|---|
| `QUEUED_STEER_NOTICE` | `queued — sends when this step finishes` |
| `SENT_STEER_NOTICE` (same turn) | `sent — the agent has it now` |
| `DEFERRED_STEER_NOTICE` | `still queued — sends with your next message` |
| `DEFERRED_SENT_STEER_NOTICE` (new, cross turn) | `sent — it rode along with your next message` |

[...]

PR #481, opened 2026-08-31
fix(compaction,tui): cancel an in-flight pass at the ceiling, and scope toast retirements

### LINKED ISSUE #413: compaction advisor: a ceiling pass starting while a background pass runs bills two summarizations
Accepted as a deliberate, bounded trade in agent review round 5 on #365, and recorded here rather than redesigned in that PR.

## What happens

The compaction advisor (BETA, `values.compaction.advisor_enabled`, off by default) runs an early pass in the background so it does not stall the turn. A pass at the ordinary ceiling stays synchronous, because that one is the safety net.

Those two facts can overlap. If a legitimately sub-ceiling advisory pass is in flight and the context then grows past the resolved ceiling, the next boundary correctly takes the synchronous path while the background pass is still running:

```
B1 (sub-ceiling, advisory)  -> background pass starts
    context grows past the ceiling
B2 (at the ceiling)         -> synchronous pass runs, correctly
peak concurrent summarizations: 2
```

Two summarization calls are billed over roughly the same history. The background one is then discarded as stale by the applicability check (verified), so nothing is misapplied.

## Why it is not urgent

- **Correctness is unaffected.** The stale pass is discarded, not applied; no history is lost. This was verified from both sides during review.
- **It is bounded.** At most one background pass is outstanding at a time (the in-flight latch), so the overlap is two calls, not N.
- **It needs the beta.** `advisor_enabled` is `false` by default, so no shipped default session can reach it.
- **It needs a race.** The context has to cross the ceiling during the window a background pass is running.

The cost is real spend, not a bug.

## Why the obvious fix was rejected

The natural-looking fix — have the ceiling pass await the in-flight background pass instead of starting its own — was independently rejected during review. It would make the one non-deferrable pass wait on a provider call that #365 deliberately made unbounded (no one is waiting on a background pass, so it has no tight deadline). That reintroduces the round 4 MAJOR-1 failure in a subtler form: the safety net becomes dependent on a call with no urgency.

## Options if this is ever worth closing

1. **Cancel the in-flight background pass** when a synchronous ceiling pass starts. Saves the tokens the background call has not yet spent, costs the ones it has, and needs care so cancellation cannot leave the latch or the pending slot inconsistent.
2. **Skip the background spawn** when the context is close enough to the ceiling that it is likely to cross during a pass. Cheap and heuristic; would need a threshold that is not a second notion of the ceiling.
3. **Leave it.** The overlap costs one extra summarization in a narrow race, in a beta that is off by default.

Worth revisiting before `advisor_enabled` ever defaults to `true`, and worth measuring first if beta telemetry shows it happening in practice.

Related: #365 (round 5 review and the remediation comment discussing this).

### LINKED ISSUE #422: Scope deferred toast retirements to the operation that scheduled them (D15 + follow-ups from #402)
Deferred follow-ups from the round 1-4 agent review of #402 (`ctrl+v` system clipboard paste), all judged non-blocking at merge and recorded here so they are not lost.

### Deferred callbacks are not scoped to the operation that scheduled them

Three separate findings across the review rounds trace to one root cause, and the ux-reviewer flagged it as worth fixing as a class rather than case by case:

- **D14** (fixed in #402) — the 400 ms minimum-display floor deferred `toast.withdraw(COMPOSER_PASTE_NOTICE)`, and because the failure notices shared that owner the deferred withdrawal killed the outcome notice after it took the slot. Reads in `[0.35s, 0.75s)` showed no outcome at all; at a 0.74 s read the notice was visible 0 ms. Fixed by giving the progress card its own owner (`COMPOSER_READING_NOTICE`).
- **D15** (open) — the same bug one level down, now between two progress cards sharing the new owner. On a fast double `ctrl+v`, paste 1's stale retirement kills paste 2's progress card. The designer measured it dying ~24 ms in leaving ~2.6 s blank; the ux-reviewer measured the worse and more common variant where paste 2's card **never appears at all** (~3.2 s blank) during an active stall. Verified two ways to predate #402 (it reproduces on the shared-owner build and vanishes with the floor at 0.0), so the cause is the deferred retirement introduced in round 2. No content is lost — both pastes land and paste 2's outcome still appears — which is the line between missing progress narration and a dead keystroke, and why it was not a hold.

The durable fix is to scope a deferred retirement to the operation that scheduled it, so a stale timer cannot retire a card it did not raise.

### `PASTE_READING_NOTICE_MIN_S` semantics are a half-truth

The 400 ms floor binds on the attach routes but not the failure routes, where `Toast.show` replaces the card early. Not a regression — the slot stays continuously occupied — but "the floor guarantees 400 ms" is a statement a future reader would misapply.

### Residual causal wording in the `set_timer` comment (F8 / U11)

The measured table is accurate and the load-bearing warning (do not simplify to `self.app.set_timer`, it reintroduces U3) is correct. Two subsidiary causal claims in the surrounding prose are not: the awaited action runs on the **App's** pump rather than the Editor's (bindings dispatch via `App._check_bindings`), and a message posted to the Editor and handled by the Editor arrives on time at 0.000 s rather than late. The real discriminator is which pump the awaited action occupies. Worth correcting precisely because the point of the comment is to stop a future refactor.

Refs #402, #372.

damianvtran:
# PR Description

## Summary of Changes

Two correctness defects of the same class: a late event from a disposed
context settling state that now belongs to something else. Same pattern as
the steer-receipt guard in PR #452 (`7f3bf654`) — carry the owning context
on the event, drop the delivery when the owner is not the current one.

### #413 — a ceiling pass starting while a background pass runs bills two summarizations

The compaction advisor (BETA) runs an early pass in the background. A pass
at the ordinary ceiling stays synchronous, because that one is the safety
net. Those two facts can overlap: a legitimately sub-ceiling advisory pass
still running when the context crosses the ceiling used to let the ceiling
path start a **second** summarization of the same history. The background
result was then discarded as stale, so correctness held and only spend was
affected.

Awaiting the background pass was independently rejected in review: it would
make the one non-deferrable pass wait on a call with no deadline (round 4
MAJOR-1 in a subtler form). The ceiling path now **cancels** the in-flight
pass and runs its own. Tokens already in flight are lost either way; the
ones not yet spent are not billed twice.

A background pass that **fails** still leaves the latch clear, so a later
ceiling boundary runs. Trading a double-bill for a missed compaction would
be worse — it means running into the context ceiling. Dispose mid-pass
still cancels via the existing background-task teardown.

### #422 — deferred toast retirements are not scoped to the operation that scheduled them

D15 from #402: on a fast double `ctrl+v`, paste 1's stale retirement kills
paste 2's progress card. The designer measured it dying ~24 ms in, leaving
~2.6 s blank. Owner matching (D14) is not enough against a **later card of
the same owner**. Each raise now mints a fresh owner token; the deferred
timer carries that token, and a mismatch is a no-op.

Also tightens the `PASTE_READING_HOOK` comment (F8 / U11): the discriminator
is which pump the awaited action occupies, not "the Editor's pump is blocked
and the App's is free". And records that `PASTE_READING_NOTICE_MIN_S` binds
on the attach/retire-in-finally routes, not on failure notices that
`Toast.show` a replacement card early.

Closes #413
Closes #422

## Change class

C2 reliability. No version bump — the manager runs one release after the
cleanup PRs land.

## Decision: cancel, do not join or skip

| option | why not |
| --- | --- |
| await the in-flight pass | unbounded wait on the safety net (rejected in #365 round 5) |
| skip the ceiling pass | missed compaction if the in-flight one fails |
| cancel | stops the spend we can still stop; ceiling proceeds immediately |

### #413 — summarization call counts

Same interleaving on origin/main (`6f8cae2e`) and at HEAD, counted on the
stream (not inferred from a flag):

```
# origin/main
after_bg        summary_calls=1 peak=1 in_flight=True
during_ceiling  summary_calls=2 peak=2 live=2 in_flight=True
after_ceiling   summary_calls=2 peak=2 commits=1

# HEAD
after_bg        summary_calls=1 peak=1 in_flight=True
during_ceiling  summary_calls=2 peak=1 live=1 in_flight=False
after_ceiling   summary_calls=2 peak=1 commits=1
```

Two sequential calls at HEAD is the cancelled background plus the ceiling
pass; concurrency stays at 1. Failure case: in-flight pass fails, ceiling
still runs (`summary_calls` 1 → 2, one successful commit).

### #422 — cross-operation retirement

Constructed at the hook the editor calls (two real `ctrl+v`s cannot overlap
under `run_test` because `pilot.press` waits for the awaited action):

- paste 1 raises, paste 2 raises, paste 1's deferred `show(False, owner=token_1)`
  fires. On origin/main both raises share `COMPOSER_READING_NOTICE` and the
  card vanishes. At HEAD paste 2's card stays.
- Single paste still retires correctly: empty clipboard, well past the
  floor, outcome on screen, progress card gone.

Rendered frames (OperatorApp, 100x30, real stylesheet):

- **before (shared owner):** no toast — paste 1's retirement hid paste 2's card
- **after (per-paste owner):** `Reading the clipboard…` still showing
- **normal retire:** `Nothing on the clipboard to paste.`

### Mutation tests

| guard broken | test dies on |
| --- | --- |
| skip `_cancel_background_compaction` entirely | `_until(not in_flight)` |
| clear the latch, leave the task running | `peak_concurrent_summaries == 2` |
| pin both raises to `COMPOSER_READING_NOTICE` | `first is not second` AND `toast.display` after stale withdraw |

### Gates (whole tree, repo config, exit codes)

```
flake8 rc=0
black rc=0
isort rc=0
pyright rc=0
```

`tests/unit/session/test_async_compaction.py` + paste/toast suites: 158 passed.

## Design round?

#422 is user-visible (toasts) but the copy, placement, and duration are
unchanged — only which card a deferred timer is allowed to retire. I do
not think it needs a design round; the frames are in the evidence above
if a designer wants to look. #413 is not visual.

### DIFF
--- a/local_operator/session/session.py
+++ b/local_operator/session/session.py
@@ -6392,44 +6392,57 @@ class Session:
         cleared the latch in its ``finally``, so this is a no-op and the
         ceiling path continues as it always did.
 
-        Does not await the cancelled task past one event-loop tick: that
-        tick is what delivers ``CancelledError`` to the pass so the new
-        summarization does not overlap it; waiting for the provider to
+        Bounded wait (0.1 s), not one event-loop tick: that bound is what
+        delivers ``CancelledError`` to the pass so the new summarization
+        does not overlap it, then proceeds. Waiting for the provider to
         acknowledge the cancel would reintroduce the unbounded wait.
         """
         task = self._compaction_pass_task
         self._compaction_pass_task = None
-        if task is not None and not task.done():
-            task.cancel()
-            # Bounded wait so the cancelled pass can unwind its in-flight
-            # summarization before we start another. Shielded: awaiting a
-            # cancelled task without a shield would CancelledError THIS
-            # (ceiling) pass, which is the one that must complete. Timed:
-            # if cancellation were swallowed, an unbounded await would be
-            # the wait this method exists to avoid.
-            try:
-                await asyncio.wait_for(asyncio.shield(task), timeout=0.1)
-            except asyncio.TimeoutError:
-                pass
-            except asyncio.CancelledError:
-                # Awaiting a cancelled task raises CancelledError even
-                # through ``shield``. That is the inner pass unwinding,
-                # which is success — unless WE were cancelled (dispose
-                # mid-ceiling), in which case the ceiling pass must not
-                # swallow it. ``Task.cancelling()`` is 3.11+; this
-                # project requires 3.12.
-                me = asyncio.current_task()
-                if me is not None and me.cancelling():
-                    raise
-        # Whatever the background pass produced is about to be stale: we are
-        # committing a different pass against the live history. Drop it so a
-        # later boundary cannot apply a summary of a prefix we are about to
-        # replace. Also covers the race where cancel was requested after the
-        # summary returned and the write of ``_pending_compaction`` ran
-        # anyway (no await between those two, so CancelledError is not
-        # delivered until the next yield).
-        self._pending_compaction = None
-        self._compaction_pass_in_flight = False
+        try:
+            if task is not None and not task.done():
+                task.cancel()
+                # Bounded wait so the cancelled pass can unwind its in-flight
+                # summarization before we start another. Shielded: awaiting a
+                # cancelled task without a shield would CancelledError THIS
+                # (ceiling) pass, which is the one that must complete. Timed:
+                # if cancellation were swallowed, an unbounded await would be
+                # the wait this method exists to avoid.
+                try:
+                    await asyncio.wait_for(asyncio.shield(task), timeout=0.1)
+                except asyncio.TimeoutError:
+                    pass
+                except asyncio.CancelledError:
+                    # Awaiting a cancelled task raises CancelledError even
+                    # through ``shield``. That is the inner pass unwinding,
+                    # which is success — unless WE were cancelled (dispose
+                    # mid-ceiling), in which case the ceiling pass must not
+                    # swallow it. ``Task.cancelling()`` is 3.11+; this
+                    # project requires 3.12.
+                    me = asyncio.current_task()
+                    if me is not None and me.cancelling():
+                        raise
+                except Exception:
+                    # The inner pass's swallows currently keep this path
+                    # dead. If they ever stop, ``wait_for`` would raise
+                    # into ``_run_compaction`` BEFORE the ceiling starts —
+                    # the missed-compaction trap #413 named as worse than
+                    # a double-bill (review round 1, F1). Cancel still
+                    # happened; the ceiling must proceed.
+                    pass
+        finally:
+            # Whatever the background pass produced is about to be stale:
+            # we are committing a different pass against the live history.
+            # Drop it so a later boundary cannot apply a summary of a
+            # prefix we are about to replace. Also covers the race where
+            # cancel was requested after the summary returned and the
+            # write of ``_pending_compaction`` ran anyway (no await
+            # between those two, so CancelledError is not delivered until
+            # the next yield). In ``finally`` so a surprise from
+            # ``wait_for`` cannot skip the latch-clear and then skip the
+            # ceiling pass (F1).
+            self._pending_compaction = None
+            self._compaction_pass_in_flight = False
 
     async def _summarize_for_compaction(
         self, plan: _CompactionPlan
--- a/local_operator/session/session.py
+++ b/local_operator/session/session.py
@@ -1721,6 +1721,11 @@ class Session:
         # ``_advisor_in_flight``, and is what stops a boundary that fires every
         # tool batch from spawning a pass per batch while the first still runs.
         self._compaction_pass_in_flight = False
+        # The Task wrapping that pass, so a ceiling pass can CANCEL it rather
+        # than run a second summarization alongside it (issue #413). Identity,
+        # not a generation: there is at most one, and the latch already
+        # enforces that. ``None`` whenever nothing is in flight.
+        self._compaction_pass_task: asyncio.Task[Any] | None = None
         # Last completed provider request (epoch ms). Distinct from
         # _last_activity_ms: the idle-flush pruning must measure provider-cache
         # age, and stamping turn bookkeeping right before the check made the
@@ -4879,12 +4884,17 @@ class Session:
             await self._run_turn([message])
             continuations += 1
 
-    def _spawn_background(self, coro: Coroutine[Any, Any, Any]) -> None:
+    def _spawn_background(self, coro: Coroutine[Any, Any, Any]) -> asyncio.Task[Any] | None:
         """Route a fire-and-forget coroutine through the session task group
         when one is open (wake deliveries, aside persistence); otherwise fall
         back to ``ensure_future``. Every spawned task is tracked so
         :meth:`dispose` can cancel and await it. After dispose nothing is
         spawned, so a late wake delivery cannot raise into an unobserved task.
+
+        Returns the Task so a caller that later needs to CANCEL this one
+        specifically (the ceiling compaction pass, issue #413) can, without
+        walking ``_background_tasks``. ``None`` after dispose: nothing was
+        spawned.
 
         The coroutine is wrapped so its failure is logged, never raised: an
         exception escaping into a TaskGroup would cancel every sibling task.
@@ -4898,7 +4908,7 @@ class Session:
             # path (the runner schedules a roster persist just as teardown
             # flips this flag).
             coro.close()
-            return
+            return None
 
         started = False
 
@@ -4932,6 +4942,7 @@ class Session:
                 coro.close()
 
         task.add_done_callback(_on_done)
+        return task
 
     def _build_tool_context(self) -> ToolContext:
         # This context is REBUILT on every turn, so anything that must outlive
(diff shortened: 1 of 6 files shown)


damianvtran:
### Agent review — round 1
Scope: 6f8cae2e35bceba5a3a949b25c081f1db4834a46..52c1c120ae113f207bea39a00266a14dcbbbc7be

Independently reconstructed both interleavings on this worktree (PYTHONPATH=/tmp/lop-review-481, repo `.venv`). Targeted suite `tests/unit/session/test_async_compaction.py` + paste/toast files: 158 passed. flake8/black/isort on the six changed files: rc=0. CI at post time: lint, type-check, pip-audit green; `test (3.12)` still pending.

## #413 — ceiling vs in-flight background pass

Counted on the stream, not a flag. Same interleaving the PR describes:

```
# HEAD
after_bg        summary_calls=1 peak=1 live=1 in_flight=True
during_ceiling  summary_calls=2 peak=1 live=1 in_flight=False
after_ceiling   summary_calls=2 peak=1 commits=1
```

Two sequential calls is the cancelled background plus the ceiling pass; concurrency stays at 1. Mutation-tested:

| guard broken | dies on |
| --- | --- |
| skip `_cancel_background_compaction` entirely | `_until(not in_flight)` never holds; `peak_concurrent_summaries == 2` |
| clear the latch, leave the task running | `peak_concurrent_summaries == 2` |

Failure / teardown, driven rather than inferred:

- In-flight pass **fails, then** a later ceiling boundary: `summary_calls` 1 → 2, one successful commit. Matches the new test.
- In-flight pass **fails while the ceiling is inside cancel** (gate opened after cancel started, `fail_summary` flipped off so only the background was bound to fail): ceiling still committed, `peak == 1`. Not in the suite (see F2).
- Dispose mid-pass: background cancelled, `pending is None`, no second summarization.

`asyncio.current_task()` inside `_run_compaction_async` **is** the wrapper Task `_spawn_background` stored — identity is not a false mismatch against `_guarded`. Re-raising `CancelledError` is load-bearing: swallowing it would leave the wrapper running until the provider returned.

Cancel-not-join is the right option of the three #413 named. Awaiting is the rejected unbounded wait; skipping the ceiling on a failed in-flight pass is the worse trade.

## #422 — scoped toast retirement

Follows the #452 pattern: owning context on the event, drop when the owner is not current. It reuses `Toast.withdraw`'s existing identity check rather than inventing a second mechanism. Each raise mints a fresh token; the deferred timer carries that token.

Constructed at the hook the editor calls (two real `ctrl+v`s cannot overlap under `run_test` because `pilot.press` waits for the awaited action):

- **Base pattern** (both raises pinned to `COMPOSER_READING_NOTICE`): paste 1's `show(False, owner=shared)` hides paste 2's card.
- **HEAD**: paste 2's card stays; paste 2's own retirement still dismisses it.
- **Normal case**: single paste, empty clipboard, well past the floor → outcome on screen, progress card gone.

Mutation: pinning both raises to `COMPOSER_READING_NOTICE` makes `first is not second` and the post-withdraw `toast.display` assertion fail.

### Design round?

No. #422 is user-visible (toasts) but copy, placement, and duration are unchanged — only which card a deferred timer is allowed to retire. A card that used to vanish incorrectly now stays; that is the bugfix, not a visual change. #413 is not visual.

## Findings

### F1 MINOR — `_cancel_background_compaction` can still decide the ceiling pass does not run

`local_operator/session/session.py:6364-6386`

The method's contract is "this method only cancels, it never decides whether the synchronous pass runs." The latch-clear and pending-drop sit **after** the `try/except`, and the except list is only `TimeoutError` and `CancelledError`.

`await asyncio.wait_for(asyncio.shield(task), timeout=0.1)` re-raises whatever the inner task settled with. Today `_run_compaction_async` and `_spawn_background._guarded` both swallow `Exception`, so the Task should settle cleanly on a failed pass and this path is not live — I could not make a `RuntimeError` from the summarizer abort the ceiling. If either swallow is ever removed, or a `BaseException` other than `CancelledError` escapes, `wait_for` raises into `_run_compaction` **before** `_finish_compaction`, the latch/pending cleanup at 6385–6386 is skipped, and the ceiling pass never starts. That is the missed-compaction trap #413 named as worse than a double-bill.

Fix in one line: move `_pending_compaction = None` / `_compaction_pass_in_flight = False` into a `finally`, and `except Exception: pass` next to the two existing handlers (still re-raise `CancelledError` when `me.cancelling()`).

### F2 MINOR — the failure-path test is sequential, not the overlapping case

`tests/unit/session/test_async_compaction.py:808-842`

`test_a_failed_in_flight_pass_does_not_skip_the_ceiling_pass` `settle_background`s the failed pass **before** pinning 700k and driving the ceiling. That pins "a finished failure leaves the latch clear." The sharp case the issue and this PR's own comment name — in-flight pass fails (or is cancelled-into-failure) **while** the ceiling is inside `_cancel_background_compaction` — is untested. I drove it; the ceiling still committed. Worth a test that opens the fail-gate after cancel starts, not after settle.

### F3 NIT — docstring says one event-loop tick; the code waits 0.1s

`local_operator/session/session.py:6349-6352` vs `6365`

"Does not await the cancelled task past one event-loop tick" is not what `wait_for(..., timeout=0.1)` does. The 0.1s bound is the actual design (deliver `CancelledError`, then proceed). The comment will mislead the next reader into deleting the timeout as "just a tick."

## Verdict

**clean** — no BLOCKER, no MAJOR. This round is TERMINAL: F1–F3 are follow-ups, not a reason for another round.

Cancel-not-join is the correct #413 fix; the overlapping-failure case works when driven even though the suite only covers sequential failure. #422 is the #452 identity pattern applied to toast owners, and it does not need a design round.

damianvtran:

Head: `49612114` (`origin/main` `119180f1` + this round). Scope of the review was `6f8cae2e..52c1c120`; this commit answers F1–F3 on that round.

**F1 MINOR — taken.** Latch-clear and pending-drop moved into a `finally`. `wait_for`'s `Exception` is swallowed next to the two existing handlers; `CancelledError` still re-raises when `me.cancelling()`. Cancel still happened; the ceiling proceeds. SHA `49612114`.

**F2 MINOR — taken.** New `test_a_background_failure_during_cancel_does_not_skip_the_ceiling`: background call is bound to fail, `fail_summary` flipped off so only that call fails, gate opens after cancel has started (`_compaction_pass_task is None`). Sequential failure test kept. SHA `49612114`.

**F3 NIT — taken.** Docstring now names the 0.1 s bound, not "one event-loop tick". SHA `49612114`.

Gates after rebase (whole tree, exit codes): flake8=0 black=0 isort=0 pyright=0.

Issue #406, opened 2026-08-29
Subagent view header: tool count counts starts, so failed calls still read as work done


damianvtran:
Follow-up deferred from PR #404 (design review round 1, D4).

## What

The subagent page header reads e.g. `running · 1m 17s · 2 tools` on a page whose two `edit` calls both errored. The header says work happened; the body says it failed twice.

The count is derived from tool *starts*, which is arguably correct — two calls were made — but it gives no signal that both failed.

## Why now

Not introduced by PR #404. With the duplicate-notice flood gone, the header is legible enough for the mismatch to become noticeable, which is how it surfaced.

## Why deferred

NIT severity; the designer explicitly requested no change in that PR. Worth a look only if the header is being touched for another reason.

Source: https://github.com/damianvtran/local-operator/pull/404#issuecomment-5463414770

PR #482, opened 2026-08-31
fix(tui): distinguish genuine subagent notice repeats from a render bug

### LINKED ISSUE #405: Subagent view: two consecutive identical notices read as the old duplicate-row bug

### LINKED ISSUE #407: Subagent view: narrow viewport opens on an orphaned wrap fragment; TRUNCATION_NOTE is not pinned
Follow-up deferred from PR #404 (design review round 1, D3).

## What

At 62x24 the first thing in the subagent transcript region is a dangling continuation line — bare `type array` in red, or `while calling edit on …` — with no glyph, no owning row, and no head note above it. The `✗` that would explain it is above the fold, so the reader's first glance lands on a red fragment mid-sentence.

The wrapping itself is correct: a long error wraps cleanly at 100 columns with continuation lines hanging-indented to align under the text after the `✗`. The issue is only where the view lands its scroll on a short viewport.

## Related

`TRUNCATION_NOTE` is mounted as an ordinary head block rather than pinned the way the tail row is (`pin_tail`), so "earlier activity dropped — the last 500 events are kept" leaves the screen as soon as content exceeds the viewport. On a tall terminal it is visible and useful; on a short one the reader loses the one row explaining why the transcript starts mid-run.

Worth considering pinning it for the same reason the tail is pinned: it is positioned by role, not by index.

## Why deferred

NIT severity, pre-existing scroll/pin behaviour rather than anything PR #404 introduced. Both rounds on that PR were terminal with 0 blocker / 0 major.

Source: https://github.com/damianvtran/local-operator/pull/404#issuecomment-5463414770

### LINKED ISSUE #408: Subagent view: id-less anchored keys could collide with a real message id

damianvtran:
## Summary

Four deferred follow-ups from PR #404, one surface. Three land; one is deliberately left alone.

- **#405 (D1)** — two genuine consecutive identical notices still occupy two rows (the data model is right and is not collapsed). They are no longer byte-identical: a consecutive run is marked `(1/N)`…`(N/N)` in the notice's own text, so a true double failure cannot be mistaken for the eleven-row flood #404 fixed. Intervening prose or a tool card leaves both rows unmarked — those already read as two attempts.
- **#406 (D4, NIT)** — **not actioned.** The header still counts tool *starts* (`2 tools` on a page whose two `edit` calls both errored). The designer asked for no change in #404 and said it was worth doing only if the header was being touched; the header ladder is not touched here. The honest count is the number of calls made; the body already shows both failures (✗ on each card). Frame: `docs/assets/pr-subagent-view-followups/after/failed.svg`.
- **#407 (D3)** — two parts. (1) A 62×24 viewport no longer opens on a hanging continuation line: the first landing snaps onto a row head so the glyph is on screen. (2) `TRUNCATION_NOTE` is page chrome between the rule and the scrolling body, not an ordinary head block, so "earlier activity dropped" stays visible at both the tail and home.
- **#408 (F3)** — id-less fallback keys are now `m:{anchor}` / `t:{anchor}`. An id-less message stamped at sequence 5 keys as `m:s5`, which cannot equal a real `message.id` of `"ms5"`.

Closes #405
Closes #407
Closes #408

#406 is reported to the manager rather than closed.

## Change class

C2 standard/pre-authorized reliability + visual follow-up. User-visible, so this PR needs a design round on top of the code round.

## Behavior

- Consecutive identical notices (same `kind`, same `notice_kind`, same text, nothing between them) keep N rows and gain `(i/N)`.
- A single notice, or identical notices with a tool card / sentence between them, is unchanged.
- Opening a truncated child on a short viewport lands on a row head, not a wrap fragment.
- The truncation note is visible for the life of a capped page, independent of body scroll.
- Id-less message/tool keys cannot collide with a producer-issued id of the old `m{anchor}` / `t{anchor}` spelling.

## Testing evidence

Gates, whole tree, repo config, exit codes:

```
flake8 rc=0
black rc=0
isort rc=0
pyright rc=0
```

`tests/unit/tui/test_subagent_view.py`: **109 passed**.

### #405 — consecutive identical notices

Before (origin/main, real `OperatorApp`, 100×30): two byte-identical rows.

```
✗ Invalid arguments: argument 'edits' does not match type array
✗ Invalid arguments: argument 'edits' does not match type array
```

After:

```
✗ Invalid arguments: argument 'edits' does not match type array  (1/2)
✗ Invalid arguments: argument 'edits' does not match type array  (2/2)
```

Frames: `docs/assets/pr-subagent-view-followups/{before,after}/dupe.svg`.

Mutation: skip `_mark_consecutive_notices` → `test_consecutive_identical_notices_carry_an_ordinal` dies (bare error ×3 instead of `(1/3)`…`(3/3)`). A count-collapse to one `×2` row would also die — the assertion is two rows.

### #406 — header count, left as starts

After frame, two errored `edit` cards, header still `running · 1m17s · 2 tools`. The body carries both ✗. `docs/assets/pr-subagent-view-followups/after/failed.svg`.

### #407 — landing snap + pinned truncation note

Before (origin/main, 62×24, wrapping error at the tail): first transcript line is a hanging continuation, no glyph.

```
payload that also failed validation on every
subsequent field of the same call: path, old_text,
…
```

Geometry on base: `scroll_y=13`, `max_scroll_y=13`, `viewport_h=9`, `virtual_h=22`, owner `NoticeBlock` at y=11, **2 rows into the block**.

After: truncation note pinned as chrome, first body glance is `✗ Invalid arguments` (a row head).

```
· earlier activity dropped — the last 500 events are kept
✗ Invalid arguments: argument 'edits' does not match
type array while calling edit on
…
```

Geometry after: `scroll_y=7`, `max_scroll_y=13`, `viewport_h=6` (chrome took 3 rows), notice starts at a row head. Truncation note is in the tail-following frame *and* the home frame (`narrow.svg` / `trunc.svg`).

Mutation, landing: skip `_schedule_landing_snap` → `test_a_narrow_viewport_opens_on_a_row_head_not_a_wrap_fragment` dies (`landing sits 3 rows into a block (offset=8, owner_top=5)`).

Mutation, pin: mount the note in the body again → `test_the_truncation_note_stays_in_the_viewport_when_the_body_scrolls` dies (`_truncation.display` is False).

### #408 — collision unrepresentable

On base (`fold_trajectory`):

```
rows painted: 1 (expected 2)
  key='ms5' kind=text text='REAL-ID'
tool rows painted: 1 (expected 2)
  key='ts9' tool=bash
```

At head:

```
keys {'m:s5', 'ms5'} texts {'ID-LESS', 'REAL-ID'}
keys {'t:s9', 'ts9'} tools {'read', 'bash'}
```

Mutation: revert the colon → `test_id_less_anchored_key_cannot_collide_with_a_real_id` dies (`{'REAL-ID'} == {'ID-LESS', 'REAL-ID'}`).

## Deliberately not done

- **#406** — header still counts starts. Honest, and the designer did not want it in #404. Closing it would be churn.
- Version bump — manager releases after PRs land.
- No merge.


damianvtran:
### Design review — round 1
Scope: 414d7679..e6672a36

Judged from rendered frames, not source. Looked at the author's `docs/assets/pr-subagent-view-followups/{before,after}/{dupe,failed,narrow,trunc}.svg` and recaptured the same four states independently via `shot.py` against this worktree at e6672a36 (real `OperatorApp`, `local_operator.tcss` applied). Recapture matches the author's after-frames on every subject below.

#### #405 — consecutive identical notices

**Before** (`before/dupe.svg`, 100×30): two byte-identical red rows, each `✗ Invalid arguments: argument 'edits' does not match type array`. A reader who learned "stacked identical red rows means the view is broken" (the #404 flood) cannot tell these are two events.

**After** (`after/dupe.svg` and independent recapture): the same two rows, same glyph, same red, now suffixed `  (1/2)` and `  (2/2)`. Two rows remain — the data model is not collapsed. The parenthetical is the smallest mark that makes a genuine double failure un-mistakable for a render bug. Contrast and alignment are fine; the ordinal sits in the notice's own colour, two spaces off the sentence, and reads as annotation rather than as part of the child's error.

#### #407 — 62×24 landing, and the truncation note as chrome

**Landing, before** (`before/narrow.svg`): first transcript line is a hanging continuation with no glyph:

```
payload that also failed validation on every
subsequent field of the same call: path, old_text,
…
```

That is a wrap fragment. It looks like a broken row, not the rest of one.

**Landing, after** (`after/narrow.svg` and independent recapture): first glance is the pinned note, then a row head with a glyph:

```
· earlier activity dropped — the last 500 events are
  kept
✗ Invalid arguments: argument 'edits' does not match
  type array while calling edit on
  …
```

The error's own wrap still fills the rest of the viewport — that is correct wrap, not a landing defect. The bottom-of-viewport continuation (`new_text, replace_all, and the trailing context the`) is the tail of a sentence whose head is on screen. Recapture geometry: `scroll_y=7`, `max_scroll_y=13`, `viewport_h=6` (chrome took rows). Header spinner still carries liveness (`⣾ running`) even though the working line sits below the fold on this short viewport.

**Truncation note, tail vs home:**

- Tail (`after/narrow.svg`): the note sits under the rule, above the body. Visible.
- Home (`after/trunc.svg`): the note sits under the rule, above the body. Visible.
- Before, tail (`before/narrow.svg`): the note is gone — it had scrolled away with the body. That was the bug.
- Before, home (`before/trunc.svg`): the note is visible only because home is where a body-mounted row lives.

After the pin, tail and home agree: the note is page chrome, not a scrollable row. Independent recapture of both states matches.

#### #406 — header still `2 tools` on two errored edits

**After** (`after/failed.svg` and independent recapture): header `running · 1m17s · 2 tools`. Body is two red `edit` cards, each with `✗` and `Invalid arguments: arg…`. The header is a census of calls started, not a scoreboard of outcomes. The body is unambiguous — a reader looking at this page cannot miss both failures.

Leaving the header ladder untouched is acceptable. The honest count is the number of calls made; restating it as `2 failed` / `0/2` would be a different sentence, and this PR correctly declined to churn it. Recorded below as a NIT so the deferral of #406 lives on this thread; it is not a reason to reopen the header.

---

### Findings

**D1 — NIT.** Header still reads `2 tools` on a page whose two `edit` calls both errored (`after/failed.svg`). Acceptable to leave: the body already shows both `✗`, and the header ladder is out of this PR's scope. Same deferral as #406 — do not churn it here.

**D2 — NIT.** At 62 columns the pinned truncation note wraps and leaves `kept` alone on the second chrome line (`after/narrow.svg`, `after/trunc.svg`). Height-auto is the right behaviour (the wording has to wrap rather than clip), and the note must stay visible, so this is not a pin defect. A shorter line would return a row to a 6-row body; follow-up copy, not this PR.

---

### Verdict

**TERMINAL.** No BLOCKER, no MAJOR. #405, #407 land as intended on the rendered frames; #406 is an acceptable leave. D1 and D2 are follow-ups, not remediation.

Independent recapture: `/tmp/rev482-designer-shots/{dupe,failed,narrow,trunc}.svg` against e6672a36, same four scenarios as the author's `shot.py`.

damianvtran:
### Agent review — round 1
Scope: 414d7679bba36843fdf5cce059b7ba67a99d22cd..e6672a36284c495c023d62055f310d22b56a6bf9

Independent read-only review of PR #482 (head `e6672a36`). Issues #405/#406/#407/#408 and PR #404 were read first. Diff reviewed before the tree. Claims were re-run, not taken on trust.

**#408 — holds.** On `origin/main` the collision is real: an id-less message stamped at seq 5 and a real `message.id == "ms5"` fold to one row (`key='ms5' text='REAL-ID'`); the tool pair collapses the same way (`key='ts9'`). At head the keys are `{'m:s5','ms5'}` / `{'t:s9','ts9'}` and both texts/tools survive. Mutation: revert the colon → `test_id_less_anchored_key_cannot_collide_with_a_real_id` dies (`{'REAL-ID'} == {'ID-LESS', 'REAL-ID'}`).

**#405 — holds.** Two (and three) genuine consecutive identical notices stay two (three) rows and are marked `(1/N)`…`(N/N)`. Intervening prose leaves both unmarked. Mutation: skip `_mark_consecutive_notices` → `test_consecutive_identical_notices_carry_an_ordinal` and `test_the_page_paints_ordinals_on_a_consecutive_identical_pair` die on the bare error. Marking at assemble time (not inside `fold_trajectory`) is the right seam given `_supersedes` refusing notice revisions.

**#406 — agree, leave it.** The header counts starts (`sum(1 for entry in self._entries if entry.kind == "tool")`). Two errored `edit` calls are two calls that happened; the body already carries ✗ on each card. The designer asked for no change in #404 and the header ladder is untouched here. Closing #406 would be churn. Not a finding.

**#407 pin — holds.** `TRUNCATION_NOTE` is a sibling of the body (`.subagent-view-truncation`), toggled via `display`, not a body block. After `scroll_end` it stays `display=True`, sits above the body (`region.y < body.region.y`), and is absent from `body.blocks()`. Mutation: mount it in the body again → `test_the_truncation_note_stays_in_the_viewport_when_the_body_scrolls` dies (`_truncation.display` is False).

**#407 landing — does not hold on the path a real page takes.** See F1.

Mutations independently re-run (all four die as claimed): colon, ordinal, skip `_schedule_landing_snap` (`landing sits 3 rows into a block (offset=8, owner_top=5)`), pin.

Local: `tests/unit/tui/test_subagent_view.py` **109 passed** (`-n0`). CI: lint / type-check / pip-audit green; `test (3.12)` cancelled, not green.

---

**F1 — MAJOR — #407 landing snap is consumed before durable history settles; sticky-tail follow puts the wrap fragment back**
`local_operator/tui/widgets/subagent_view.py:1934-1946` (`settle_tail`), `:2266-2327` (`_schedule_landing_snap` / `_snap_landing_to_row_head`), `tests/unit/tui/test_subagent_view.py:3270` (`test_a_narrow_viewport_opens_on_a_row_head_not_a_wrap_fragment`)

Issue #407 is a first-glance bug: at 62×24 the page opens on a hanging continuation (`payload that also failed validation on every`) with the `✗` above the fold. The one-shot snap fixes that **only when there is no durable history**.

`test_a_narrow_viewport_opens_on_a_row_head_not_a_wrap_fragment` never attaches `_subagent_comms`, so `_history_unavailable` stays True and `_apply_history_page` never runs. Same for `docs/assets/pr-subagent-view-followups/shot.py`. On that fixture the snap works (`offset=5 == owner_top`, first visible line is `✗ Invalid arguments…`). Mutation of the snap kills that test. The test cannot fail at the production boundary.

A real child always has a transcript directory (`app.py` passes `comms.session_dir_of(job_id)`). Reproduced at 62×24 with the same wrapping notice plus 20 durable assistant rows (the shape `test_durable_history_opens_at_tail_and_pages_to_the_start` already uses):

```
no-comms:            into=0  first visible = "✗ Invalid arguments…"
with-history (20 rows): into=3  first visible = "payload that also failed validation on every"
```

At `_apply_history_page(..., initial=True)` time, `_landing_snap_pending` is already **False** and the pre-history body is already snapped (`offset=5, max=8`). The one-shot fired on the trajectory-only body and consumed itself. `settle_tail` then calls `_snap_landing_to_row_head()` after `scroll_end(animate=False)`, which no-ops on `if not self._landing_snap_pending`. History prepend grows `virtual_size`; `TranscriptView._size_updated` sees `_tail_anchor.following` still True and `_scroll_to_tail()`s onto `max_scroll_y`, which bisects the wrapping notice.

Two cooperating defects, one symptom:

1. The one-shot is spent on the pre-history layout. Initial history is still the first glance, not “a reader who has already scrolled.”
2. Even a snap that did run would be programmatic (`programmatic_scroll`), so following stays armed and the next extent change tails again. `settle_tail` also uses Textual’s deferred `scroll_end` rather than `_scroll_to_tail` (the API this module documents as landing short).

The pin half of #407 is real. The landing half is not, for any child that has written a history page — which is the page a user actually opens. Closing #407 on the strength of a comms-less fixture would close the issue while the reported first glance is still on screen.

Fix in one line: keep the snap pending until after the initial history page has settled (re-arm in `settle_tail`), snap against the immediate tail (`_scroll_to_tail` then pull back to the owning block’s `virtual_region.y`), and do not let a following `_size_updated` tail-follow undo that offset; pin the landing test to a FakeSession that has `session_dir_of` and a non-empty transcript.

---

**N1 — NIT — history-prepend test docstring still describes the old body-index-0 offset**
`tests/unit/tui/test_subagent_view.py:3111-3121`

The assertions now correctly require `_truncation.display` and `body.blocks()[0] is not view._head_block`. The docstring still says the note “holds that note at 0” and that `insert_blocks` needs a head-offset. Harmless, but it is the previous bug’s write-up sitting on the test that replaced it.

---

Verdict: **not clean**. 1 MAJOR (F1). Round is **not TERMINAL**.

damianvtran:

Head: `d6b12070` (`e6672a36` + F1). Scope of this reply: F1, N1, D1, D2.

**F1 — MAJOR — fixed in `d6b12070`.** The one-shot no longer fires on the trajectory-only body. `_sync_body` skips the snap while `_initial_tail_pending` (the initial history load) is set. `_settle_initial_landing` runs after the initial history page *or* after history-unavailable/error, re-arms the snap, `_scroll_to_tail()`s immediately (not Textual's deferred `scroll_end`), snaps onto the owning block's `virtual_region.y`, and `TailAnchor.release()`s if that offset is short of the tail so `_size_updated` cannot follow back onto the wrap fragment.

The landing test now attaches `_subagent_comms.session_dir_of` and a 20-row durable transcript — the production boundary the comms-less fixture could not reach — and asserts the wrapping `Invalid arguments` notice's top, not just any row head.

Mutation (consume the one-shot on the pre-history body and skip re-arm in `settle`): `test_a_narrow_viewport_opens_on_a_row_head_not_a_wrap_fragment` dies with `landing sits 3 rows into a block (offset=28, owner_top=25, max=28)` — the same `into=3` the review reproduced.

**N1 — NIT — fixed in `d6b12070`.** Docstring on `test_a_history_page_lands_below_the_truncation_note` now describes the note as page chrome, not a body-index-0 offset.

**D1 — NIT — deferred.** Header still counts tool starts (`2 tools` on two errored edits). Both reviewers accepted the leave-as-starts call; the ladder is untouched. Closing #406 would be churn.

**D2 — NIT — fixed in `d6b12070`.** Truncation copy is now `earlier activity dropped — last 500 events kept` so it stays one content row at 62 columns (the previous wording wrapped so `kept` sat alone on line 2). Guard: `test_the_truncation_note_stays_in_the_viewport_when_the_body_scrolls` asserts the renderable has no newline.

Gates after the fix, whole tree: flake8=0 black=0 isort=0 pyright=0. `tests/unit/tui/test_subagent_view.py` 109 passed.

damianvtran:
### Agent review — round 2
Scope: 414d7679bba36843fdf5cce059b7ba67a99d22cd..d6b12070a564177808605ada6226c0f79358b2b1

Independent read-only review of PR #482 head `d6b12070` (round-1 remediation). Diff of `e6672a36..d6b12070` read first; claims re-run, not taken on trust. Drove a real `OperatorApp` at 62×24 with a transcript directory (20 durable rows + wrapping notice), not the PR's test.

**F1 — MAJOR — fixed.** Independent probe (not the landing test) against a child WITH `session_dir_of` and 20 durable assistant rows, 62×24:

```
offset=25  owner_top=25  into=0  owner=NoticeBlock
notice_top=25  max_scroll_y=28  viewport_h=9  virtual_h=37
following=False  history_entries=20
first_visible= "✗ Invalid arguments: argument 'edits' does not match"
```

First visible line is a row head with the glyph, not a wrap fragment. Sticky-follow is released so a later `_size_updated` cannot tail back onto the fragment.

Mutation (burn the one-shot on the pre-history body by dropping the `_initial_tail_pending` refuse in `_snap_landing_to_row_head` / `_sync_body`, and skip re-arm in `_settle_initial_landing`): the same probe dies with `into=3` — `landing sits 3 rows into a block (offset=28, owner_top=25, max=28)`. Same `into=3` round 1 reproduced. The new landing test reaches that boundary.

**#405 — still holds.** Consecutive identical notices stay N rows and carry `(1/N)`…`(N/N)`. Intervening prose leaves both unmarked. Mutation: skip `_mark_consecutive_notices` → three bare errors, assertion dies.

**#408 — still holds.** Keys are `{'m:s5','ms5'}` / `{'t:s9','ts9'}`; both texts/tools survive. Mutation: drop the colon → collapse to `{'ms5'}` / `{'REAL-ID'}`. Collision remains unrepresentable.

**#407 pin — still holds.** `TRUNCATION_NOTE` is page chrome (`_truncation.display`, sibling of the body). `test_the_truncation_note_stays_in_the_viewport_when_the_body_scrolls` passed.

**D1 — NIT — still deferred.** Header still counts tool starts (`sum(1 for entry in self._entries if entry.kind == "tool")` at `subagent_view.py:2392`). Ladder untouched. Agree with the leave.

**D2 — NIT — taken.** Copy is now `earlier activity dropped — last 500 events kept` (47 cells). Guard asserts the renderable has no newline. Independent landing probe did not wrap it.

**N1 — NIT — taken.** History-prepend docstring now describes the note as page chrome.

Local: `tests/unit/tui/test_subagent_view.py` **109 passed** (`-n0`). Targeted landing / ordinal / collision / pin tests **5 passed**. CI at post time: lint / type-check / pip-audit / tui-e2e (macos + ubuntu) green; `test (3.12)` still pending.

---

No new findings.

Verdict: **clean**. Round is **TERMINAL**. Remaining D1 is a follow-up, not a reason for another round.

PR #427, opened 2026-08-30
fix(tui): paste and copy with cmd+v and cmd+c, not only ctrl

damianvtran:
# PR Description

## Summary of Changes

**Cmd+V did nothing in the composer, and Cmd+C's copy quietly diverged from
Ctrl+C's.** Both Mac-native chords are now bound alongside the portable
`Ctrl+` pair, and the docstrings that asserted Cmd+C "cannot be made to" work
are corrected — that claim was measured only against Terminal.app and is false
for Ghostty, the terminal the defect was reported from.

Reported:

> Pasting with ctrl+v into the macOS terminal or Ghostty works but cmd+V does
> nothing, which is the preferred and expected paste UX for mac users.

> cmd+C copying characters from the composer doesn't trigger the copy but
> ctrl+C on mac works.

### Whether a Mac chord reaches a TUI is the terminal's decision

A TUI receives bytes on a pty and cannot see a key the terminal does not
encode. Three cases, and Ghostty is two of them:

1. The terminal binds the chord and consumes it. **Terminal.app** does this —
   zero bytes, unreachable by any application, not fixable from our side.
2. The terminal implements the **kitty keyboard protocol** and forwards it as
   CSI-u: `CSI <codepoint> ; <modifiers> u`, modifiers `1 + bitmask` with
   **Super=8**. So `ESC[118;9u` is literally Cmd+v.
3. The terminal pastes text itself via bracketed paste — the ordinary text
   path, and correct.

Textual's `linux_driver` already pushes the protocol flags (`ESC[>25u`), so
**these keys were arriving and the composer simply never bound them.**

Captured with a raw-mode PTY probe, re-measured independently for this PR:

| terminal | clipboard | key | bytes delivered |
|---|---|---|---|
| Ghostty 1.3.1 | image-only PNG | Cmd+V | **8** — `ESC[118;9u` |
| Ghostty 1.3.1 | image-only PNG | Cmd+C | **7** — `ESC[99;9u` |
| Ghostty 1.3.1 | **text** | Cmd+V | 21 — `ESC[200~…ESC[201~`, key NOT forwarded |
| Ghostty 1.3.1 | image-only PNG | Ctrl+V | 8 — `ESC[118;5u` |
| Terminal.app | image-only PNG | Cmd+V | **0** |
| Terminal.app | image-only PNG | Cmd+C | **0** |
| Terminal.app | image-only PNG | Ctrl+V | 1 — `\x16` |

Every figure was taken only after a typed-character control proved the harness
[...]

PR #483, opened 2026-08-31
fix(tui): reopen argument lists after home/end and unfreeze the tip ring

### LINKED ISSUE #393: [BUG] home then end closes an open /mcp argument list and does not reopen it
## Summary

`home` then `end` closes an open ARGUMENT list and does not reopen it. The COMMAND list survives the same pair of keys, so the two lists disagree about what a round trip to the start of the line and back should do.

## Reproduce

Driven through the real `OperatorApp` under `run_test` with real key presses:

```
COMMAND list   type "/mc"    -> picker open
               home, end     -> picker still open        (correct)

ARGUMENT list  type "/mcp "  -> picker open, 3 verb rows
               home, end     -> picker CLOSED, no rows   (the bug)
```

The caret ends where it started, the buffer is unchanged, and the command word is still `/mcp ` — so nothing about the parse has changed, only whether the list is showing.

## Pre-existing

Confirmed on clean `main` at `abb7a68e`, independent of any in-flight branch:

```
abb7a68e   COMMAND  /mc    home+end -> open=1
abb7a68e   ARGUMENT /mcp   home+end -> open=0
```

Found by the `ux-reviewer` subagent while walking #378 and bisected to before that PR's changes, so it is not a regression from the inline ghost-text work. Filing separately so it is not lost when that PR merges.

## Why it matters

`home`/`end` is an ordinary caret gesture, not a dismissal. Escape is the key that means "I do not want this list", and it already works. Losing the argument rows to a caret move means a user who jumps to the start of the line to fix a typo in the command word comes back to no list, with no indication of why, and has to edit a character to get it back.

The user-visible consequence is slightly worse now that #378 paints an inline preview: the ghost is derived from the highlighted row, so when the list closes here the preview goes with it, and the composer loses both at once.

## Where to look

`Editor._on_key` routes `home`/`end` to the picker while it is open (`editor.py`, the `if self._picker.is_open():` branch handles `home`/`end` as picker jumps), and the ARGUMENT path re-derives its rows from a posted message rather than synchronously. The likely shape is that the argument list is re-synced from a caret-anchored parse mid-gesture and closes before the second key lands, where the command list re-derives from the buffer and survives. Worth checking `_sync_picker`'s argument branch against the caret position at each of the two presses.

## Not blocking anything

Recorded as a follow-up. No in-flight PR depends on it, and the workaround (type a character) is immediate.

### LINKED ISSUE #424: Welcome tip never rotates with shimmer off, narrow-width notice tier, and selection replacement on the path paste route
Two user-visible follow-ups deferred from the design and UX review of #402 (`ctrl+v` system clipboard paste). Neither blocked the merge.

### D9 — with shimmer off, the welcome tip never rotates

`_sync_tip_timer` creates no timer at all when shimmer is disabled, so the tip row holds at `TIPS[0]` forever. Every other tip in the ring is unreachable for those users on every launch. This was found while assessing `welcome.TIPS` as a discoverability surface for `ctrl+v`; it is why the placeholder was chosen instead, but the tip ring itself is still broken for shimmer-off users independently of that decision.

Measured for the splash generally: the tip is never in the opening frame, median 84 s to first sighting, and P(seen within 30 s) = 0.18 — worth knowing before treating `TIPS` as a place to teach anything.

### Notice widths below 62 columns

The composer notice family fits the 58-cell toast box at the widths tested (34-54 cells rendered, logo intact), but there is no tier below ~62 columns. The placeholder degrades by wrapping so the key token survives to a 32-cell composer; the notices were not given the same treatment. Worth a narrow-width tier so a very small terminal does not clip the outcome messages.

### U7 — the cmux/path paste route does not replace an active selection

`ctrl+v` was fixed for this in #402 (U2): it now routes through `_replace_via_keyboard` + `move_cursor`, so pasting over a selection replaces it. The **path route** — cmux's `Cmd+V`, which bracket-pastes a synthesised file path — still has the original defect and inserts at the caret without consuming the selection. Verified byte-identical on the base commit, so it predates #402 and was correctly filed as a follow-up rather than held against it. Fixing it would make the two routes consistent, which is the stated design goal for those branches.

Refs #402.

### LINKED ISSUE #430: Terminal.app users have no discovery path for ctrl+v image paste
### Problem

`ctrl+v` is the only way to attach an image in terminals that do not implement
the kitty keyboard protocol, and after #427 it is advertised nowhere ambient.
Cmd+V now works natively wherever the terminal forwards it, which is most of
them, so the composer placeholder hint was removed as no longer earning
permanent screen real estate. That is the right trade for the majority, but it
leaves one population with no discovery path at all.

For a **Terminal.app** user the situation is:

- Cmd+V delivers **zero bytes**. The app never runs a line of code, so no
  notice, no beep-and-teach, no reactive affordance is possible. Unfixable.
- `ctrl+v` is the only way in, and nothing ambient names it.
- `/help` documents it (and is the right home for the conditional statement),
  but a user whose Cmd+V silently did nothing has no reason to suspect a paste
  key exists to look up.
- The splash tip is weak here, measured against the current pool in #427's
  design round: `ctrl+v` is `TIPS[5]` of 12 rotating every 12 s with the first
  frame pinned to `TIPS[0]`. Median time to first sighting **84 s**, P(seen
  within 30 s) = **0.18**, and **0.00** for a user who types straight away.
  With `display.shimmer` false or `LOCAL_OPERATOR_NO_SHIMMER` set, no timer is
  created at all, so P(ever seen) = **0.00**.
- `WelcomeView.display` goes False on the first prompt, so the tip is absent
  mid-session by construction — and mid-session is exactly the reported journey
  (working, take a screenshot, reach for paste).

### Proposed fix

Both the designer and the ux-reviewer independently proposed the same cheap
remedy, and it reuses machinery that already exists rather than adding a
surface: **make the TIP conditional instead of making a placeholder permanent.**

Detect at startup whether the terminal answered the kitty-protocol query
(Textual already pushes `ESC[>25u`; the reply is observable). On a terminal
that did NOT answer — precisely the stranded population — pin `ctrl+v` as the
opening tip in place of `TIPS[0]`, the same mechanism `TIP_SETUP` already uses
for the first-run state (`local_operator/tui/widgets/welcome.py:365`). The
existing `TERM_PROGRAM` detection in `glyphs.py:209` is the fallback signal if
the protocol reply turns out not to be reachable from app code.

This reaches ~100% of stranded users on their first frame and 0% of everyone
else, costs no row, no width tier and no permanent composer real estate.

### Scope note

It does not solve the mid-session case, which is the harder half: the splash is
gone once a turn starts. Worth deciding separately whether that needs anything,
since the failure is silent rather than destructive — nothing is lost, the
paste simply does not happen.

### Context

Deferred from #427 (agent review round 1) as correctly out of slice. The
discoverability trade is recorded in that PR's design and code review comments
with the full measurements above.

damianvtran:
# PR Description

## Summary of Changes

Three TUI defects that share a discoverability surface: the argument picker,
the welcome tip ring, and the `ctrl+v` paste chord.

Closes #393. Closes #424. Closes #430.

### #393 — `home` then `end` closed an open ARGUMENT list

The COMMAND list survived the same keys; the ARGUMENT list did not.

`_on_key` re-derives the picker **before** the caret-moving binding runs, so
`home` on `/mcp ` synced at the end of the line (list open), then the caret
jumped to column 0, where `slash_argument` is `None`. Nothing asked again
once `end` put the caret back. A COMMAND list survived because
`slash_context` still matches at column 0 of an unterminated word.

Two compounding bugs on the same gesture:

1. `None` (never tracked) and `""` (first slot, no verb) were treated as
   different `/mcp` subcommand keys, so `home` posted a `RefreshArgumentChoices`
   whose builder then ran at the **new** caret and closed the verb list.
2. The refresh handler had no caret-staleness guard, so a message queued by
   `home` could refill-and-close a list that `end` had just reopened.

Fix: re-sync the picker from `watch_selection` at the **settled** caret
(skipped while `_set_text_and_caret` is parking, so D5 still holds); treat
`None`/`""` as the same `/mcp` slot; drop a refresh whose caret no longer
matches.

Driven through the real `OperatorApp` under `run_test`:

```
BEFORE (origin/main)
COMMAND  typed='/mc'    open=(True, 1) home=(True, 1) end=(True, 1)   # correct
ARGUMENT typed='/mcp '  open=(True, 6) home=(False, 0) end=(False, 0) # the bug

AFTER
COMMAND  typed='/mc'    open=(True, 1) home=(True, 1) end=(True, 1)
ARGUMENT typed='/mcp '  open=(True, 6) home=(False, 0) end=(True, 6)  # reopens
```

`home` still closes the argument list (the caret left the argument — that is
the parse, not a dismissal). `end` puts it back. Escape remains the key that
means "I do not want this list".

### #424 D9 — with shimmer off, the welcome tip never rotated

`_sync_tip_timer` created no timer at all when shimmer was disabled, so the
row held at `TIPS[0]` forever. Every other tip in the ring — including the
`ctrl+v` paste chord — was unreachable on every launch for those users.

The pulse **should** stay still (a colour on a clock makes every snapshot a
coin flip). The tip is not a colour: it is a discoverability surface. The
timer now runs whenever the splash is on screen. Snapshots stay deterministic
because they capture the opening frame, before the first 12 s tick.

Mutation-tested: restoring `wanted = bool(self.display) and shimmer_enabled()`
kills both D9 tests (`rc=1`, `_tip_timer is None`). Toggling shimmer at
runtime neither stops the rotation nor stacks a second interval.

### #424 U7 — the path paste route did not replace a live selection

`ctrl+v` was fixed for this in #402 (U2). The cmux/path route still called
`insert`, so selecting `WORD` and pasting a file path gave
`keep WORD[Image #1]  keep`. All three paste owners now share
`_replace_selection`.

### #424 narrow-width notice tier — not acted on

`NoticeBlock` already word-wraps via `wrap_cells` at the block's own width
(one row at 90 columns, three at 40). That is a wrap, not a missing tier, and
it is a different surface from the composer placeholder. Left alone rather
than inventing a second degradation ladder next to one that already works.

### #430 — Terminal.app had no discovery path for `ctrl+v`

After #427 the composer placeholder no longer names the chord. `Cmd+V`
delivers zero bytes on Terminal.app, so no reactive notice is possible.
`/help` documents it, but a user whose paste silently did nothing has no
reason to look it up. The splash tip is weak as a general teacher (median
84 s to first sighting) **and** was frozen for shimmer-off users (D9).

One coherent change serves both: D9 unfreezes the ring for everyone, and a
Terminal.app launch pins the existing `ctrl+v` pool entry as the **opening**
tip — the same mechanism `TIP_SETUP` already uses for first-run. Reaches
~100% of stranded users on their first frame and 0% of everyone else. Setup
still wins when both apply. `/help` stays the durable lookup; this is the
ambient first-frame.

Detection is from `TERM_PROGRAM` (`terminals.is_apple_terminal`), never from
a capability query: stdin belongs to Textual's input loop.

Does not solve the mid-session case (the splash is gone once a turn starts).
Recorded, not fixed here — the failure is silent rather than destructive.

## Testing

Gates, whole tree, repo config, exit codes:

```
.venv/bin/python -m flake8 .                         flake8 rc=0
uvx --from black==26.1.0 black --check .             black rc=0
uvx isort==5.13.2 --check .                          isort rc=0
                                                     pyright rc=0
```

Driven / rendered:

- #393 before/after key sequence on the real `OperatorApp` (output above).
- Tip ring with shimmer OFF: opening frame still `TIPS[0]`; four ticks later
  the row is `ctrl+v attaches an image from the system clipboard`. Looked at
  the rendered SVGs.
- Terminal.app pin: opening frame is the paste tip; a tick walks the ordinary
  ring (`/team <name> <message>…`). Looked at the rendered SVGs.
- D9 mutation: restoring the shimmer conjunct kills the new tests.
- Shimmer toggle: the original tip timer object is unchanged across on/off.

Suites: `test_welcome.py` (107 with boot layout), `test_paste_clipboard.py` +
`test_paste_images.py` + `test_command_picker.py` (412), `test_ghost_text.py`
(51), parked-hint / team-chart (22), plus the new home/end and path-paste
tests.

## User-visible

Yes. Argument-list caret motion, tip rotation with shimmer off, Terminal.app
opening tip, path-paste selection replacement. Changes a discovery flow
(where `ctrl+v` is taught). Wants a design round and a UX round as well as
the code review.

No version bump — the manager runs one release after the cleanup PRs land.

### DIFF
--- a/local_operator/tui/app.py
+++ b/local_operator/tui/app.py
@@ -14886,6 +14886,18 @@ class OperatorApp(App[None]):
             # Same staleness guard as on_argument_query_opened: the buffer may
             # have moved on in the tick the message spent queued.
             return
+        if editor._caret_offset() != message.caret:
+            # The caret moved while this message sat queued, so the builder's
+            # own parse — which anchors on the caret — would answer for a
+            # position the user has already left. This is the #393 seam: `home`
+            # on ``/mcp `` parks the caret on the ``/`` where ``slash_argument``
+            # reports no argument at all, the verb list closes, and the refresh
+            # queued by that same key then runs the builder at column 0, gets
+            # the empty verb list back, and re-closes the list the editor's
+            # `end` had just re-opened. A caret-only move never changes the
+            # buffer, so the rows this message was posted to fetch are the ones
+            # the list already holds — dropping it is correct, not a loss.
+            return
         if message.command == "mcp":
             picker = editor.picker
             # Same self-clearing rule as every sibling fill, applied BEFORE
--- a/local_operator/tui/widgets/editor.py
+++ b/local_operator/tui/widgets/editor.py
@@ -816,9 +816,16 @@ class RefreshArgumentChoices(Message):
     character typed.
     """
 
-    def __init__(self, command: str) -> None:
+    def __init__(self, command: str, caret: int) -> None:
         super().__init__()
         self.command = command
+        #: Whole-buffer offset the editor parsed at when it posted this. The
+        #: app's refill builder re-parses at the LIVE caret, so a message that
+        #: sat queued across a caret-only move (``home`` then ``end`` on an
+        #: open ``/mcp `` list) would refill against a position that no longer
+        #: has an argument — and close the list the later key just reopened
+        #: (#393). The handler drops a message whose caret no longer matches.
+        self.caret = caret
 
 
 class ArgumentHighlightChanged(Message):
@@ -3796,8 +3803,18 @@ class Editor(TextArea):
         # `getattr` for the same reason as above — a reactive watcher can fire
         # during base-class construction, before this subclass's attributes
         # exist, and `_ghost_completion` reads several of them.
-        if hasattr(self, "_picker"):
-            self._sync_ghost()
+        #
+        # Re-derive the PICKER here, not just the ghost. `_on_key` syncs
+        # BEFORE the caret-moving binding runs, so a round trip of `home`
+        # then `end` used to close an ARGUMENT list at the start of the line
+        # and never ask again once the caret was back in the argument —
+        # while a COMMAND list survived because `slash_context` still matches
+        # at column 0 of an unterminated word (#393). Asking at the settled
+        # caret is the only reopen path that cannot depend on which key got
+        # you here. Skipped while `_set_text_and_caret` is parking the caret
+        # so that helper's one-sync-at-the-final-position contract (D5) holds.
+        if hasattr(self, "_picker") and not getattr(self, "_suspend_picker_sync", False):
+            self._sync_picker()
 
     # -- paste ----------------------------------------------------------------
     async def _on_paste(self, event: events.Paste) -> None:
@@ -3893,7 +3910,7 @@ class Editor(TextArea):
                 return
             event.prevent_default()
             event.stop()
-            self.insert(attached)
+            self._replace_selection(attached)
             return
         attached = await self._attach_pasted_images(event.text)
         if attached is None:
@@ -3906,7 +3923,7 @@ class Editor(TextArea):
         # down ("paste a file path instead"), so without this the card that
         # gave the instruction reappears to deny it worked (round 2, D8/D3).
         self.post_message(EditorPasteAttached())
-        self.insert(attached)
+        self._replace_selection(attached)
 
     async def _attach_clipboard_image(self, *, allow_text: bool = False) -> str | None:
         """Read the clipboard and attach (or insert) whatever is on it.
@@ -4152,10 +4169,26 @@ class Editor(TextArea):
             # posted the notice that says so, and inserting nothing is the
             # honest outcome.
             return
-        # `move_cursor` to the edit's end mirrors `TextArea._on_paste`: without
-        # it `maintain_selection_offset=False` leaves the caret where the
-        # replaced range began, so the next keystroke types BEFORE the text
-        # just pasted.
+        self._replace_selection(pasted)
+
+    def _replace_selection(self, pasted: str) -> None:
+        """Put ``pasted`` at the caret, replacing a live selection if there is one.
+
+        THE one insertion point for every paste that this widget owns — the
+        empty-paste clipboard branch, the path/cmux branch, and
+        :meth:`action_system_paste`. ``insert`` is the branch that does NOT
+        replace, so a path that called it while a range was live stuffed the
+        marker into the selection instead of swapping it (#424 U7, the same
+        defect #402 U2 already fixed on the ``ctrl+v`` route). Sharing the
+        helper is what keeps the three from drifting apart again.
+
+        ``_replace_via_keyboard`` is the exact call the base class's own paste
+        makes, so an empty selection degenerates to an insert at the caret.
+        ``move_cursor`` to the edit's end mirrors ``TextArea._on_paste``:
+        without it ``maintain_selection_offset=False`` leaves the caret where
+        the replaced range began, so the next keystroke types BEFORE the text
+        just pasted.
+        """
         result = self._replace_via_keyboard(pasted, *self.selection)
         if result is not None:
             self.move_cursor(result.end_location)
@@ -4670,9 +4703,18 @@ class Editor(TextArea):
                     # `verb + sep` so both edges cross: entering the server slot
                     # and backspacing out of it.
                     subcommand = f"{first_tok.lower()}{sep}" if list_argument else ""
-                    if subcommand != self._argument_subcommand:
+                    # ``None`` (never tracked, the state after ArgumentQueryOpened
+                    # resets the slot) and ``""`` (first slot, no verb) are the
+                    # same list. Treating them as different posted a refresh on
+                    # the first caret-only key after `/mcp ` — `home` — and the
+                    # handler then rebuilt the rows at the NEW caret, which for
+                    # `home` is the `/` where `slash_argument` is None, so the
+                    # verb list closed under the user's fingers (#393). Same
+                    # None/"" normalisation the `/team` branch below already
+                    # does for the chart slot.
+                    if subcommand != (self._argument_subcommand or ""):
                         self._argument_subcommand = subcommand
-                        self.post_message(RefreshArgumentChoices(command))
+                        self.post_message(RefreshArgumentChoices(command, cursor))
                 else:
                     # `/team` has exactly ONE thing that changes its choice set:
                     # crossing into or out of the `chart ` SECOND slot (first
@@ -4701,7 +4743,7 @@ class Editor(TextArea):
                     is_chart_slot = chart_second_slot == "chart"
                     self._argument_subcommand = chart_second_slot
                     if was_chart_slot != is_chart_slot:
-                        self.post_message(RefreshArgumentChoices(command))
+                        self.post_message(RefreshArgumentChoices(command, cursor))
             self._picker.sync_argument(list_argument)
             # U1/U2 discoverability hint. The moment a NAME+message name is
             # autofilled (or hand-typed) to `/<cmd> <name> ` with an empty tail,
--- a/local_operator/tui/widgets/editor.py
+++ b/local_operator/tui/widgets/editor.py
@@ -1195,6 +1195,11 @@ class Editor(TextArea):
         # BEFORE ``super().__init__`` because TextArea's constructor loads the
         # initial document through ``load_text`` → ``_sync_picker``.
         self._suspend_picker_sync = False
+        # Last parse phase `_sync_picker` settled. Compared by
+        # `_sync_picker_if_phase_changed` so a caret move that stays inside
+        # one phase does not re-open an Esc-dismissed list. Set BEFORE
+        # ``super().__init__`` because that constructor already syncs.
+        self._picker_phase_at_last_sync: str | None = None
         # The escape action held for one pump turn, or ``None`` when no escape
         # is in flight (the resting state). See the escape-coalescing block
         # below :meth:`_on_key` for why an escape is ever held at all.
@@ -3804,17 +3809,26 @@ class Editor(TextArea):
         # during base-class construction, before this subclass's attributes
         # exist, and `_ghost_completion` reads several of them.
         #
-        # Re-derive the PICKER here, not just the ghost. `_on_key` syncs
-        # BEFORE the caret-moving binding runs, so a round trip of `home`
-        # then `end` used to close an ARGUMENT list at the start of the line
-        # and never ask again once the caret was back in the argument —
-        # while a COMMAND list survived because `slash_context` still matches
-        # at column 0 of an unterminated word (#393). Asking at the settled
-        # caret is the only reopen path that cannot depend on which key got
-        # you here. Skipped while `_set_text_and_caret` is parking the caret
-        # so that helper's one-sync-at-the-final-position contract (D5) holds.
+        # Re-derive the PICKER here, not just the ghost — but ONLY when the
+        # caret has crossed a parse PHASE. `_on_key` syncs BEFORE the
+        # caret-moving binding runs, so a round trip of `home` then `end`
+        # used to close an ARGUMENT list at the start of the line and never
+        # ask again once the caret was back in the argument (#393). Asking
+        # at the settled caret is the reopen path that cannot depend on
+        # which key got you here.
+        #
+        # Unconditional `_sync_picker` here is the wrong answer: that helper
+        # re-opens a list whose query still matches, so an Esc-dismissed
+        # model/command list came back on the next caret move (`shift+up`
+        # in the model picker, CI 3.12/3.13). The phase check is what
+        # keeps a dismissal a dismissal: `home` on `/mcp ` leaves the
+        # argument, `end` re-enters it, and a motion that stays inside
+        # the same phase (or outside every list) is a no-op for the
+        # picker. Skipped while `_set_text_and_caret` is parking the caret
+        # so that helper's one-sync-at-the-final-position contract (D5)
+        # holds.
         if hasattr(self, "_picker") and not getattr(self, "_suspend_picker_sync", False):
-            self._sync_picker()
+            self._sync_picker_if_phase_changed()
 
     # -- paste ----------------------------------------------------------------
     async def _on_paste(self, event: events.Paste) -> None:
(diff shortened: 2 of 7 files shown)


damianvtran:
### Design review — round 1
Scope: d8a8c833de7c1ba4a5531cf809e2c3e63d319fcb..ba11fa1927cbd54a41f8671f4b14ef83aa7f9db9

Findings: none.

Driven independently through the real `OperatorApp` under `run_test` at 100×30 (and 80×24 for the narrow case), stylesheet loaded, shimmer off. SVGs rendered and looked at as frames, not as markup. Geometry on every 100×30 frame: `virtual_size == size == (98, 28)`, no scrollbar. Pulse timer absent, `mark_color` None.

**Argument list (`/mcp ` → home → end).** Typed opens the six-verb list (`list` / `add` / `remove` / `login` / `logout` / `reauth`) with the `list` ghost in the composer. Home closes it — the caret left the argument, which is the parse, not a dismissal; the composer holds `/mcp` with the caret at column 0. End restores the list. Typed and end frames are byte-identical (same rows, same highlight, same ghost, same picker region `(13, 20, 73, 6)`). A second settled frame after end is also byte-identical — no post-paint reflow.

**Command list (`/mc` → home → end).** The `/mcp` row stays open through home (caret at column 0 of the unterminated word) and end. Typed and end frames are byte-identical.

**Welcome tip, shimmer off.** Opening frame is `· /resume picks up a recent session where you left off`. Three consecutive frames with no tick are byte-identical — the pulse is still. Forced ticks walk the ring without moving any other row:

- `· /team <name> <message> sends work to the manager`
- `· Ask to create an agent with its own instruction set`
- `· /model <provider>/<id> switches this session only`
- `· /usage shows how much provider quota is left`
- `· ctrl+v attaches an image from the system clipboard`

A second settled frame after the last tick is byte-identical to it. At 80×24 the opening tip is still present and whole (no ellipsis).

**Terminal.app opening frame.** With `TERM_PROGRAM=Apple_Terminal`, the first painted tip is `· ctrl+v attaches an image from the system clipboard` — the working chord, named as an action, on the first frame a Terminal.app user sees. It reads as a discovery path, not as a complaint about `Cmd+V`. A tick leaves the pin and walks the ordinary ring (`/team <name> <message>…`). The same opening sentence is whole at 80×24.

**Setup wins over paste.** Entering setup on a Terminal.app launch replaces the paste tip with `· /login <provider> sets up a provider (e.g. /login openai)`. The hint table leads with `/login`, the band says `setup`. A later tick walks the ordinary ring; the pin is opening-frame only.

Verdict: **TERMINAL — no BLOCKER or MAJOR findings.**

damianvtran:
### UX review — round 1
Scope: d8a8c833de7c1ba4a5531cf809e2c3e63d319fcb..ba11fa1927cbd54a41f8671f4b14ef83aa7f9db9

Driven through the real `OperatorApp` under `run_test` (100×30 and 80×24), real keypresses, consecutive frames. Not inferred from source.

## Walks

**1. `/mcp ` argument list → home → end → pick a verb**

Typed `/mcp ` (trailing space). Verb list opened on the first frame: `list / add / remove / login / logout / reauth`, `list` highlighted, composer showing `/mcp list` as the ghost completion.

`home`: list closed on the same key. Caret at column 0, picker mode COMMAND, no rows. The overlay left, the splash mark came back. That is the parse — the caret is no longer in an argument — not a dismissal. Escape is still the key that means "I do not want this list" (confirmed separately: escape closes and a following `end` does **not** reopen).

`end`: list reopened on the same key. Same six rows, same highlight, caret back at column 5. Immediate — not pending, not a later pause, no empty flash. Tab completed the highlighted verb to `/mcp list` and the list narrowed to that row.

No hitch on the round trip. The command list never leaves because `/mc` still matches at column 0; the argument list does leave because the caret left the slot. Coming back is instant, which is what was broken.

**2. `/mc` command list → home → end**

Stayed open through both keys. `home` parked the caret at column 0 with `/mcp` still suggested; `end` put the caret back. Rows identical. Same as before this PR.

**3. Launch with shimmer off, watch the tip row**

`pulse_timer` is None (glow stays still). `tip_timer` is running. Opening frame is `/resume picks up a recent session where you left off`. Four ticks later the row walked `/approvals…`, `ctrl+c copies…`, `esc stops the agent…`, `Ask for parallel work…`. Each sentence fully readable, no ellipsis. Welcome height held at 22; virtual size equalled actual; no scrollbar; every non-tip row byte-identical across ticks. Nothing else animated.

**4. Terminal.app user who has never heard of ctrl+v**

`TERM_PROGRAM=Apple_Terminal`. Opening tip is `ctrl+v attaches an image from the system clipboard` — first frame, including at 80×24. Splash hints still name `/help`. After the first prompt the splash is gone (welcome.display False, no tip row). Typing `/help` still documents:

```
ctrl+v              attaches an image or file from the system clipboard
cmd+v               same, where the terminal forwards it; not Terminal.app
```

So: the opening tip is enough to teach the chord to someone looking at launch, and `/help` is still a path after the splash retires. The splash itself taught `/help`.

## Findings

None.

**Follow-up, not a hold** — mid-session Terminal.app `Cmd+V` is still silent once the splash is gone. A user who missed the opening tip (or whose paste fails an hour in) gets zero bytes and no notice; they have to think to open `/help`. The PR records this as out of scope. `/help` is a real path, so this is not a discovery failure on launch.

## Verdict

**TERMINAL.** No BLOCKER, no MAJOR. The argument-list home/end round trip matches the command list for immediacy (close-on-home is the parse, reopen-on-end is instant). The tip ring rotates with shimmer off and does not drag any other row. Terminal.app is taught on the first frame, and `/help` remains after the splash.

damianvtran:
### Agent review — round 1
Scope: f82c2f5e9f09fea6518506abb9a5414b76c31f70..ba11fa1927cbd54a41f8671f4b14ef83aa7f9db9

Independent review of PR #483 (`dev-picker-tip-discovery`, head `ba11fa19`). Diff is the six-file slice against merge-base `f82c2f5e` (current `origin/main` is `d8a8c833`; this branch does not contain it). Driven through the real `OperatorApp` under `run_test` on both `origin/main` and this head. Did not accept the author's claims on trust.

## What was verified

**#393 — home then end on an open argument list.** Same key sequence as the issue, real presses, real `OperatorApp`:

```
origin/main (d8a8c833)
  COMMAND  /mc     typed open=1  home open=1  end open=1   # already correct
  ARGUMENT /mcp    typed open=6  home open=0  end open=0   # the bug

head (ba11fa19)
  COMMAND  /mc     typed open=1  home open=1  end open=1   # unchanged
  ARGUMENT /mcp    typed open=6  home open=0  end open=6   # reopens
```

`home` still closes the argument list (caret left the argument). `end` puts it back with the same six verb rows. Command list is byte-identical across the pair. Server-slot `/mcp login ` home/end likewise closes then reopens on `login linear`, not the verb list.

The load-bearing reopen path is `watch_selection` → `_sync_picker` at the settled caret. The `None`/`""` normalisation stops `home` posting a `RefreshArgumentChoices` in the first place. I could not kill the home/end sequence by deleting only the caret-match drop in `on_refresh_argument_choices` — with `None`/`""` in place, `home` posts nothing, so the drop never fires on this gesture. The drop is still the right answer for a queued refresh whose caret has moved: invoking the handler with a matching caret refills, invoking it with caret `0` against an already-filled `/mcp login ` list leaves the server rows in place (does not close). Burst-typing `/mcp login ` and `/mcp login l` still landed `login linear`. The handler is synchronous, so a later key cannot overtake a refresh that has started; a caret-mismatched drop therefore cannot snatch a refill out from under a list that `watch_selection` has already settled.

**#424 D9 — tip rotates with shimmer off.** Suite autouse exports `LOCAL_OPERATOR_NO_SHIMMER=1`. Opening frame is still `TIPS[0]`; `_tip_timer is not None`; `_pulse_timer is None`. A tick turns the row over. Mutation: restoring `wanted = bool(self.display) and shimmer_enabled()` in `_sync_tip_timer` kills both new tests (`rc=1`, `_tip_timer is None` on `test_the_rotation_runs_when_animation_is_disabled` and `test_the_tip_timer_is_created_even_when_shimmer_is_off`). Restored after. Shimmer toggle neither stops the original tip timer nor stacks a second interval.

**#424 U7 — path paste shares `_replace_selection`.** `test_a_path_paste_replaces_a_live_selection_like_ctrl_v` passed: selecting `WORD` and pasting a PNG path yields `keep [Image #1, 1568x200]  keep`, caret after the marker.

**#430 — Terminal.app pins `TIP_PASTE`.** `_tip_lines(..., pin_paste=True)` opens on `TIP_PASTE`; index 1 walks the ordinary ring; `setup=True, pin_paste=True` still yields `TIP_SETUP`. `is_apple_terminal` is True only for `TERM_PROGRAM=Apple_Terminal` (case-insensitive). iTerm, ghostty, WezTerm, vscode, bare `xterm-256color`, and an empty env are all False — other terminals keep `TIPS[0]`. The composed splash with `TERM_PROGRAM=Apple_Terminal` (and the suite's `GHOSTTY_BIN` deleted) opens on the paste tip; a tick walks off it.

Related suites at `-n0`: `test_home_then_end_keeps_an_open_argument_list`, the new welcome/paste tests, `test_ghost_text.py`, `test_welcome.py`, D5 parked-hint, `test_team_chart.py` — 110 + 22 passed.

## Findings

### F1 MINOR — splash tests are not hermetic against `TERM_PROGRAM`

`WelcomeView.__init__` captures `terminals.is_apple_terminal()` at construction (`welcome.py:1092`). That helper reads **only** `TERM_PROGRAM` (`terminals.py:132-139`). It does not look at `GHOSTTY_BIN`.

The suite's autouse fixture (`tests/unit/tui/conftest.py:42`) seeds `GHOSTTY_BIN` so nerd-icon detection is host-independent, and the new comment at `welcome.py:1090-1091` claims that seed means "a per-frame read would never pin in tests either". That is false: `is_apple_terminal` ignores `GHOSTTY_BIN`. `test_the_composed_splash_pins_paste_on_apple_terminal` monkeypatches `TERM_PROGRAM` because the author knew it matters; the inverse is not pinned.

Consequence: `test_the_rotation_runs_when_animation_is_disabled` asserts the opening row is `TIPS[0]` (`test_welcome.py:1248`). Run the same file under Terminal.app (`TERM_PROGRAM=Apple_Terminal`) and that assertion dies — the pin is doing its job, the test is not isolated. CI Ubuntu/macOS runners typically have no `TERM_PROGRAM`, so this will not fail in Actions; it will fail for anyone who runs the TUI suite inside Terminal.app.

Fix in one line: pin `TERM_PROGRAM` to a non-Apple value (or `delenv` it) in `hermetic_tui_env`, the same way `GHOSTTY_BIN` is already seeded. The Apple_Terminal test already overrides it.

## Verdict

**clean** — no BLOCKER, no MAJOR. Round 1 is TERMINAL. F1 is a follow-up, not a reason for another round.

CI at post time: lint / type-check / pip-audit green; `test (3.12)` and `test (3.13)` still pending.

damianvtran:

Scope reviewed: `f82c2f5e..ba11fa19`. Remediation head: `0369e066`.

**F1 MINOR — splash tests are not hermetic against `TERM_PROGRAM`.** Fixed in `0369e066`. `hermetic_tui_env` now `delenv("TERM_PROGRAM")`, matching how `GHOSTTY_BIN` is already seeded the other way. `is_apple_terminal` reads only `TERM_PROGRAM`, so the ghostty seed never isolated this. Verified: `TERM_PROGRAM=Apple_Terminal pytest tests/unit/tui/test_welcome.py::test_the_rotation_runs_when_animation_is_disabled` now passes (opening frame is still `TIPS[0]`); the Apple_Terminal pin test still sets the var itself and still opens on `TIP_PASTE`.

**CI red on both 3.12 and 3.13** (`test_a_self_contained_chord_never_swallows_a_pending_escape[model_picker-*-shift+up-csi]`). Mine, not pre-existing. Reproduced on this branch (`2 failed`); same cells on `origin/main` (`2 passed`). Cause: `watch_selection` → unconditional `_sync_picker()` reopened an Esc-dismissed model list on the next caret move. Fix in the same commit: resync from `watch_selection` only when the parse **phase** changes (`argument` / `command` / `None`). `home` then `end` on `/mcp ` is still a phase change and still reopens (driven: `ARGUMENT end=(True, 6)`). Full `test_word_caret_matrix.py`: 295 passed, 8 skipped.

No other findings to answer. Design round 1 and UX round 1 were terminal with 0 findings.

CI on `0369e066` is in flight; I will not merge.

PR #485, opened 2026-08-31
fix(tui): make shell mode legible without colour, and keep a settings row at 16 lines

### LINKED ISSUE #385: TUI: make shell mode legible without colour, and surface keybindings in /help
Two related discoverability/accessibility gaps, both raised during review of #375 (word-wise caret movement) and both **pre-existing** rather than introduced there. Filed together because they share a root cause: the TUI has no in-app surface that tells you what state you are in or what keys exist.

### 1. Shell mode is signalled only by one glyph's hue (ux round 2 U8)

Bang-mode (`!`) changes what `Enter` does: it runs the buffer as a local shell command instead of sending a prompt. Once the buffer has text the placeholder is gone, and the **only** remaining difference in the entire frame is the chevron's colour.

Verified from rendered frames during #375 review:

```
mode held  ->  #57c785  (green)
mode lost  ->  #e9e5db  (grey)
```

That is a single cell, a hue change rather than a shape change, and it is the whole signal. Users with a red-green colour vision deficiency, a low-contrast theme, or simply not looking at that one column have no cue that `Enter` is about to run a command rather than send a prompt.

This was also why the U1 defect in #375 was invisible: a chord silently flipped the mode and the frame looked identical. That specific bug is fixed, so the chevron now only has to carry the deliberate lone-`Esc` case, which the user initiated. It is adequate for that, which is why this is a follow-up rather than a blocker.

**Candidate fix:** a non-colour mode marker. A `!` in the status band, or keeping a short mode hint visible once the buffer has text, so bang-mode is legible without relying on hue.

### 2. `/help` lists no keybindings (ux round 1 U5, ux round 2 U9)

`_help_block` renders slash commands, the log path, the window-title note and the notifications note. It lists **no keys at all** - re-checked on the #375 head, `/help` mentions none of `option`, `alt+`, `word`, `ctrl+l`, or `shift+tab`. The splash only advertises `/`, `/help` and `ctrl+d`.

So the README is the only channel documenting keyboard affordances, which is a weak one for a user already inside the TUI with no way to ask "what keys exist?". This matters more now that #375 adds composer chords whose whole value is muscle memory.

**Candidate fix:** a `keys` section in `/help` covering at least:

- `option+left/right`, `option+shift+left/right` (word movement and selection), and the `ctrl+` equivalents
- `option+up/down`
- `ctrl+b` / `ctrl+f` (aside, fork aside)
- `shift+tab` (cycle reasoning effort)
- `ctrl+l` (clear transcript), `ctrl+t` (todos), `ctrl+g` (subagents)
- `esc` (stop), `ctrl+c` (copy/interrupt), `ctrl+d` (quit)

Note the bindings are `show=False` by design (the footer is not a key reference in this app), so this wants to be an explicit `/help` section rather than flipping `show`.

### Why one issue

Both are "the TUI does not tell you what it is doing or what it can do", and a `/help` keys section is a natural place to also name the shell-mode indicator once it exists. Splitting them would duplicate that discussion.

Raised in review of #375: [ux round 1](https://github.com/damianvtran/local-operator/pull/375#issuecomment-5458434997) (U5) and [ux round 2](https://github.com/damianvtran/local-operator/pull/375#issuecomment-5459025386) (U8, U9).

### LINKED ISSUE #431: /settings paints chrome and zero setting rows at 16-row terminals
Split out of the design review on #425 (D2, NIT). **Pre-existing** — reproduces identically on `a892aaf5`, before the boot-layout fix — so it was deliberately not fixed in that PR.

## What happens

At `100x16` the `/settings` body resolves to **one row tall and renders zero setting rows**. The frame is header, rule, detail line, hints and dock, with a blank band where the list should be. The page is honest about being a settings page but shows nothing settable.

Measured on both sides of #425, so this is not a regression from it:

```
100x16  body.size=(62, 1)   rows painted: 0
```

The splash path and the conversation path are byte-identical here, so it is the height budget, not the boot layout.

## Why it is worth fixing

Every other narrow-terminal degradation on this page sheds gracefully: the read-only pane sheds whole at 60 and 80 columns and the footer sheds with it. The vertical ladder has no equivalent — the body simply reaches zero usable rows and the page keeps its full chrome instead of trading some of it for content.

## Suggested direction

Decide a minimum viable body and make the chrome pay for it, in the spirit of the existing width ladder in `SettingsView._paint_hints`:

- shed the detail line first (it describes a row the user cannot see anyway),
- then the rule,
- keeping the title and `esc` as the invariants, the way `esc` is never shed from the footer.

Alternatively, state a minimum height and show a single honest line explaining the terminal is too short, rather than a blank band.

## Reproduce

```sh
env -u NO_COLOR TERM=xterm-256color .venv/bin/python     scripts/settings_shot.py /tmp/out.svg 100x16 overview
```

The printed geometry line reports `body.size` and the row count; render the SVG to see the blank band.

damianvtran:
# PR Description

## Summary of Changes

Two TUI legibility/layout gaps, both pre-existing, both visual.

**#385 — shell mode is signalled only by one glyph's hue.** Bang-mode (`!`)
changes what Enter does, but once the buffer has text the placeholder is gone
and the only remaining difference in the entire frame is the chevron's colour
(`#57c785` vs `#e9e5db`). That is invisible to a colourblind reader, under
`NO_COLOR`, and on a monochrome terminal. A user who cannot see the hue has
no way to know Enter will run a command rather than talk to the agent.

The chevron now swaps **shape** as well as colour: `❯` (prompt) → `$` (shell).
Colour remains a redundant cue for readers who can see it. `/help` gains a
tightly-scoped keybindings section (same two-column gutter as the existing
`ctrl+c` / `ctrl+v` rows) covering bang-mode, word movement, effort, panels,
aside, esc, and quit.

**Overlap with #430.** Another agent is concurrently adding discovery for
`ctrl+v` and may touch `/help`. This PR only *appends* key rows after the
existing `cmd+v` note; it does not rewrite `_help_block`. Sequence the
merges so #430's copy/paste wording lands first if it also edits those
rows.

**#431 — `/settings` at 16 rows paints chrome and zero setting rows.** At
`100x16` the body resolved to one row tall and rendered nothing settable:
header, rule, detail, hints, dock, and a blank band. Measured identically
on both sides of #425, so it was deliberately not fixed there.

Chrome now pays for a body row, one row at a time, so restoring a unit as
the terminal grows never drops the body:

| terminal | view | body rows on frame | chrome shed |
|---|---|---|---|
| 14 | 5 | 0 (honest "too short" line) | rule + pad; columns hidden |
| 15 | 6 | 1 | rule + pad |
| 16 | 7 | 1 | pad |
| 17 | 8 | 2 | pad |
| 18 | 9 | 2 | none |
| 20 | 11 | 4 | none |
| 24 | 15 | 8 | none |

A 2-row shed (the detail line) at 16 would give the body 2 rows and
restoring it at 17 would drop the body to 1 — the non-monotonic move the
issue forbids. Title, `esc`, and the detail line stay: the detail is where
a validation error and a delete ask live.

Closes #385
Closes #431

### Gates (exit codes, whole tree, repo config)

```
flake8 rc=0
black rc=0
isort rc=0
pyright rc=0
```

### #385 — shell mode without hue

**Before** (base `origin/main`): prompt-typed and shell-typed frames were
**text-identical**. The only difference was the chevron's ink.

```
before-prompt-typed  row_head='  ❯  l'  ink=#e9e5db
before-shell-typed   row_head='  ❯  l'  ink=#57c785
prompt-typed vs shell-typed: identical text = True
```

**After**, colour:

```
after-prompt-typed   row_head='  ❯  l'  ink=#e9e5db
after-shell-typed    row_head='  $  l'  ink=#57c785
text identical? False
  row 16: '❯ls' != '$ls'
```

**After**, `NO_COLOR=1 TERM=dumb` (monochrome):

```
after-mono-prompt-typed  row_head='  ❯  l'  ink=#e5e5e5
after-mono-shell-typed   row_head='  $  l'  ink=#aaaaaa
```

The `$` is still there when hue is gone. Looked at the rendered PNGs of
both colour and monochrome frames.

`/help` at 80x44 paints the new key rows on one line each (`!`,
`option+left/right`, `shift+tab`, `ctrl+l/t/g/b`, `esc`, `ctrl+d`), with
the `$` / `❯` swap named on the bang-mode continuation. Looked at the
rendered `/help` frame.

### #431 — `/settings` height ladder

**Before** (compositor, `100xH`):

```
h=16 view= 7 cols= 0 body= 1 rows_on_frame= 0
h=18 view= 9 cols= 2 body= 2 rows_on_frame= 2
h=20 view=11 cols= 4 body= 4 rows_on_frame= 4
h=24 view=15 cols= 8 body= 8 rows_on_frame= 8
```

**After** (same probe, compositor):

```
h=14 view= 5 cols= 0 body= 0 rows_on_frame= 0  too_short=True
h=15 view= 6 cols= 1 body= 1 rows_on_frame= 1
h=16 view= 7 cols= 1 body= 1 rows_on_frame= 1  "Default provider" on frame
h=17 view= 8 cols= 2 body= 2 rows_on_frame= 2
h=18 view= 9 cols= 2 body= 2 rows_on_frame= 2
h=20 view=11 cols= 4 body= 4 rows_on_frame= 4
h=24 view=15 cols= 8 body= 8 rows_on_frame= 8
```

Monotonic: growing the terminal never loses a body row. Looked at the
rendered 16/18/20/24 frames — 16 now shows `› Default provider  anthropic`
instead of a blank band.

### Mutation test (height guard)

Forcing full chrome at 16 rows (the pre-fix layout) paints **0** setting
rows onto the frame. `test_the_height_ladder_dies_when_the_body_floor_is_broken`
asserts that, then restores the ladder. Breaking the floor makes the
sibling assertion die — the test is not keyed on today's chrome flags.

### Unit

```
tests/unit/tui/test_settings_view.py                          (full file)
tests/unit/tui/test_word_caret.py
tests/unit/tui/test_boot_layout.py
plus the new /help and shell-mode tests in test_app_pilot.py
468 passed
```

## Notes for review

- User-visible: needs a **design round**. Shell-mode changes how a mode is
  communicated, so a **UX round** as well.
- Do not bump the version; the manager runs one release after PRs land.
- Do not merge this PR from here.

### DIFF
--- a/local_operator/tui/app.py
+++ b/local_operator/tui/app.py
@@ -1098,13 +1098,31 @@ ASIDE_LAYOUT_CLASS = "aside"
 #: ``OperatorApp._sync_composer_focus``.
 COMPOSER_FOCUSED_CLASS = "-composer-focused"
 
-#: Class the input dock carries while bang-mode is on. The chevron stays the
-#: same glyph — ``composer_cells`` finds the row by it — and the class is
-#: what the stylesheet reads to paint it ``string`` rather than the focus
+#: Class the input dock carries while bang-mode is on. The class is what the
+#: stylesheet reads to paint the marker ``string`` rather than the focus
 #: brightness. ``string`` is the ramp's command/success green, a step off
 #: the accent, so a shell-mode composer does not spend the "turn is live"
 #: ink (D5) and still reads as a different MODE from the resting field.
+#:
+#: The GLYPH also changes (``PROMPT_CHEVRON`` → ``SHELL_CHEVRON``). Hue
+#: alone is invisible to a colourblind reader, under ``NO_COLOR``, and on a
+#: monochrome terminal — and bang-mode changes what Enter does, so that
+#: silence is a safety gap, not polish (#385). Shape carries the mode;
+#: colour remains a redundant cue for readers who can see it.
 COMPOSER_SHELL_CLASS = "-composer-shell"
+
+#: The composer's resting prompt marker. One cell, and the glyph
+#: ``composer_cells`` (and the command picker, deliberately) locate the
+#: row by. Bang-mode swaps it for :data:`SHELL_CHEVRON` rather than
+#: recoloring this one: a hue-only swap is invisible without colour.
+PROMPT_CHEVRON = "❯"
+
+#: Bang-mode's prompt marker. The POSIX shell prompt, one cell, so it
+#: fits the same 2-cell ``#prompt-chevron`` box and does not shift the
+#: editor. Chosen over ``!`` because ``!`` is the KEY that enters the
+#: mode; using it as the marker would make the frame look like the bang
+#: was inserted into the buffer (it is consumed, not typed).
+SHELL_CHEVRON = "$"
 
 #: How long after a terminal resize the floating overlay cards re-measure
 #: themselves. They are hosted in `width: auto` containers, so Textual sends
@@ -2298,7 +2316,7 @@ class OperatorApp(App[None]):
                 yield Band(id="status-band")
                 editor = Editor(commands=SLASH_COMMANDS)
                 with Horizontal(id="input-row"):
-                    yield Chrome("❯", id="prompt-chevron")
+                    yield Chrome(PROMPT_CHEVRON, id="prompt-chevron")
                     yield editor
                 # The picker is the editor's, but it cannot be the editor's
                 # CHILD: it has to draw across the full panel width, outside the
@@ -6833,12 +6851,20 @@ class OperatorApp(App[None]):
         dock.set_class(focused, COMPOSER_FOCUSED_CLASS)
 
     def on_shell_mode_changed(self, message: ShellModeChanged) -> None:
-        """Follow the composer's bang-mode onto the dock class the sheet reads."""
+        """Follow the composer's bang-mode onto the dock class AND the glyph.
+
+        The class is the colour cue (``string`` green). The glyph is the
+        colour-independent one: ``$`` instead of ``❯``, so a reader who
+        cannot see the hue still knows Enter will run a command (#385).
+        Both flip on the same message so they cannot drift apart.
+        """
         try:
             dock = self.query_one("#input-dock")
+            chevron = self.query_one("#prompt-chevron", Chrome)
         except NoMatches:
             return
         dock.set_class(message.active, COMPOSER_SHELL_CLASS)
+        chevron.update(SHELL_CHEVRON if message.active else PROMPT_CHEVRON)
 
     def on_interrupt_requested(self, message: InterruptRequested) -> None:
         """Ctrl+C from the composer — the NORMAL path, since it holds focus.
@@ -15601,6 +15627,49 @@ class OperatorApp(App[None]):
             style=dim,
         )
         lines.append(cmd_paste_note)
+
+        # The rest of the key reference. Same surface, same two-column
+        # gutter, same 74-cell ceiling at 80 columns as the copy/paste
+        # rows above — `/help` is the in-app place a user already opens
+        # when lost, and until these rows existed the README was the only
+        # channel documenting the chords (#385). Tightly scoped additions
+        # rather than a second surface: another agent is concurrently
+        # working #430 against the same block, so a wholesale rewrite of
+        # this method would collide.
+        #
+        # Descriptions are budgeted against `name_width` (currently 20,
+        # derived from `/settings, /config`) so `name_width + description`
+        # stays <= 74. Continuation lines use the same empty-gutter
+        # indent the `ctrl+c` row already established.
+
+        def _key_row(key: str, description: str) -> Text:
+            row = Text()
+            row.append(key.ljust(name_width), style=muted)
+            row.append(description, style=dim)
+            return row
+
+        def _key_more(description: str) -> Text:
+            row = Text()
+            row.append("".ljust(name_width), style=muted)
+            row.append(description, style=dim)
+            return row
+
+        # Bang-mode's marker is named here because the glyph is the
+        # colour-independent signal (#385) and a user who cannot see the
+        # hue still needs to know what `$` on the composer means. The
+        # description is 47 cells (67 composed) so it has headroom.
+        lines.append(_key_row("!", "run the buffer as a local shell command"))
+        lines.append(_key_more("the composer shows $ instead of ❯; esc leaves"))
+        lines.append(_key_row("option+left/right", "move by word; +shift selects"))
+        lines.append(_key_more("ctrl+left/right is the same on Linux/Windows"))
+        lines.append(_key_row("option+up/down", "same as up/down (history, lists)"))
+        lines.append(_key_row("shift+tab", "cycle reasoning effort"))
+        lines.append(_key_row("ctrl+l", "clear the transcript (history stays)"))
+        lines.append(_key_row("ctrl+t", "expand or collapse the todo panel"))
+        lines.append(_key_row("ctrl+g", "expand or collapse the subagent panel"))
+        lines.append(_key_row("ctrl+b", "open an aside; ctrl+f forks it in"))
+        lines.append(_key_row("esc", "stop the agent; leave a mode"))
+        lines.append(_key_row("ctrl+d", "quit, on an empty composer"))
         # Where the logs went. Console logging is off while the TUI owns the
         # terminal (see `local_operator.logger.file_logging`), so without this
         # line the file is unfindable without reading the source. `/help` and
--- a/local_operator/tui/local_operator.tcss
+++ b/local_operator/tui/local_operator.tcss
@@ -427,7 +427,7 @@ ToolCard:focus, WakeBlock:focus {
 /* Bang-mode recolors the marker AND swaps the glyph (`❯` → `$`; see
  * `OperatorApp.on_shell_mode_changed`). Colour is the redundant cue for
  * readers who can see it; the glyph is the one that survives `NO_COLOR`,
- * a monochrome terminal, and red-green colour vision deficiency (#385).
+ * a monochrome terminal, and red-green colour vision deficiency (issue 385).
  * `string` is the ramp's command/success green — a step off the accent,
  * so a shell-mode composer does not spend the "turn is live" ink (D5).
  * Outranks the focus brightness: the mode is the louder fact, and a
@@ -1533,7 +1533,7 @@ Screen.org-chart #input-dock #prompt-chevron {
     /* Default quiet row above the list. `SettingsView._apply_height_ladder`
      * CLEARS this padding on a short terminal so the body keeps a setting
      * row — a 1-row shed, so restoring it as the terminal grows never
-     * drops the body (#431). Do not raise this to 2: a 2-row unit is the
+     * drops the body (issue 431). Do not raise this to 2: a 2-row unit is the
      * non-monotonic move that issue forbids. */
     padding: 1 0 0 0;
 }
--- a/local_operator/tui/local_operator.tcss
+++ b/local_operator/tui/local_operator.tcss
@@ -424,14 +424,15 @@ ToolCard:focus, WakeBlock:focus {
     color: $lo-fg;
 }
 
-/* Bang-mode recolors the chevron, not the glyph. `composer_cells` finds the
- * row by `❯`, and swapping the character would make every focus/caret test
- * miss the composer the moment the mode was on. `string` is the ramp's
- * command/success green — a step off the accent, so a shell-mode composer
- * does not spend the "turn is live" ink (D5) and still reads as a different
- * MODE from the resting field. Outranks the focus brightness: the mode is
- * the louder fact, and a focused shell-mode chevron that went `fg` would
- * look identical to a focused prompt. */
+/* Bang-mode recolors the marker AND swaps the glyph (`❯` → `$`; see
+ * `OperatorApp.on_shell_mode_changed`). Colour is the redundant cue for
+ * readers who can see it; the glyph is the one that survives `NO_COLOR`,
+ * a monochrome terminal, and red-green colour vision deficiency (#385).
+ * `string` is the ramp's command/success green — a step off the accent,
+ * so a shell-mode composer does not spend the "turn is live" ink (D5).
+ * Outranks the focus brightness: the mode is the louder fact, and a
+ * focused shell-mode marker that went `fg` would look identical to a
+ * focused prompt on a colour-capable terminal. */
 #input-dock.-composer-shell #prompt-chevron {
     color: $lo-string;
 }
@@ -1529,6 +1530,11 @@ Screen.org-chart #input-dock #prompt-chevron {
  * have (an empty registry would otherwise leave a wide blank gutter). */
 .settings-view-columns {
     height: 1fr;
+    /* Default quiet row above the list. `SettingsView._apply_height_ladder`
+     * CLEARS this padding on a short terminal so the body keeps a setting
+     * row — a 1-row shed, so restoring it as the terminal grows never
+     * drops the body (#431). Do not raise this to 2: a 2-row unit is the
+     * non-monotonic move that issue forbids. */
     padding: 1 0 0 0;
 }
 
--- a/local_operator/tui/widgets/editor.py
+++ b/local_operator/tui/widgets/editor.py
@@ -875,10 +875,12 @@ ASIDE_PLACEHOLDER = "Ask the aside…"
 
 #: What the composer says in bang-mode. The first clause is opencode's
 #: sentence — what typing here WILL DO — and the second names the way out,
-#: because entry is taught three ways (tip, placeholder, green chevron) while
-#: exit was taught by nothing on screen (design round 1, D2): the placeholder
-#: is the one surface guaranteed visible the moment the mode opens on an
-#: empty buffer, which is exactly when a first-timer looks for the door.
+#: because entry is taught three ways (tip, placeholder, and the ``$``
+#: marker — a shape change, not a hue, so it survives ``NO_COLOR``; #385)
+#: while exit was taught by nothing on screen (design round 1, D2): the
+#: placeholder is the one surface guaranteed visible the moment the mode
+#: opens on an empty buffer, which is exactly when a first-timer looks
+#: for the door.
 SHELL_PLACEHOLDER = "Run a command… — esc to leave"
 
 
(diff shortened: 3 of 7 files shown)


damianvtran:
### Agent review — round 1
Scope: d8a8c833de7c1ba4a5531cf809e2c3e63d319fcb..89c25fe6d588fcfe5d9ad4c19fddbb722a9f09ed

Independent review of PR #485 (`dev-shell-mode-narrow`, head `89c25fe6`). Read #385 and #431. Diff first, then compositor probes, then mutation of both guards. Did not accept the author's claims on trust.

## What was verified (not findings)

**#385 — `$` vs `❯` is painted, not a style.** Compositor at 80x24, buffer `ls`:

```
prompt  cell='❯'  ink=#e9e5db   row_head='  ❯  ls'
shell   cell='$'  ink=#57c785   row_head='  $  ls'
text_identical=False
Chrome.render() = Content('$') while shell_mode is on
```

`NO_COLOR=1 TERM=dumb` (and `NO_COLOR=1 TERM=xterm-256color`): glyphs still disagree (`❯` vs `$`). Inks also disagree (`#e5e5e5` vs `#aaaaaa`) — colour remains a redundant cue; shape is the surviving one.

Mutation: forcing `chevron.update(PROMPT_CHEVRON)` in `on_shell_mode_changed` kills both `test_shell_mode_is_legible_without_hue_once_the_buffer_has_text` (`'$' in '  ❯  ls'`) and `test_bang_on_empty_composer_enters_shell_mode`. Restored.

**#431 — height ladder is monotonic.** Independent compositor probe, `100xH`:

```
h=14 view=5  body=0 chrome=5 pad=0 rule=0 cols=0 too_short  on_frame=0  "terminal too short…"
h=15 view=6  body=1 chrome=5 pad=0 rule=0 cols=1            on_frame=1  Default provider
h=16 view=7  body=1 chrome=6 pad=0 rule=1 cols=1            on_frame=1  Default provider
h=17 view=8  body=2 chrome=6 pad=0 rule=1 cols=1            on_frame=2  Model + Default provider
h=18 view=9  body=2 chrome=7 pad=1 rule=1 cols=1            on_frame=2
h=20 view=11 body=4 chrome=7                               on_frame=4
h=24 view=15 body=8 chrome=7                               on_frame=8
```

Growing never drops a body row. At 16 the one visible row is the selected setting, not the `Model` header.

Mutation: forcing `show_rule = show_pad = True` kills `test_settings_keeps_a_body_row_on_a_short_terminal` at h=15 (`on_frame=0`). The dedicated `test_the_height_ladder_dies_when_the_body_floor_is_broken` also reproduces the blank band under full chrome. Restored.

**`/help` vs #483.** PR #483 does not touch `_help_block` (its `app.py` hunk is the #393 caret-staleness guard on `on_refresh_argument_choices`). This PR only appends key rows after the existing `cmd+v` note. No conflict. New rows compose to 42–65 cells against the 74-cell ceiling; `test_the_paste_key_rows_do_not_wrap_at_eighty_columns` covers `!` / `option+left/right` / `shift+tab` / `ctrl+d` on the painted 80-column frame.

**`$` vs bang `!`.** Unambiguous. `!` is consumed as the key that *enters* the mode (not inserted); `$` is the POSIX prompt in the 2-cell `#prompt-chevron` box. Using `!` as the marker would look like the bang landed in the buffer. Cost figures (`$0.12`) and the bash tool icon live on other rows. Shell-mode + slash picker paints `$  /` on the composer and `❯  /help` on the picker — two glyphs in the same column, which is the mode signal working, not a collision.

## Findings

### F1 — BLOCKER — issue hashes in tcss comments fail the no-literal-hex gate

`local_operator/tui/local_operator.tcss:430` (`#385`) and `:1536` (`#431`).

`tests/unit/tui/test_minimalism.py::test_tcss_has_no_literal_hex` searches the whole sheet with `#[0-9a-fA-F]{3,8}\b`. `#385` and `#431` match. `test_welcome.py::test_stylesheet_region_has_no_literal_hex` takes everything after `/* ---- welcome view` to EOF, so `#431` in the settings-columns comment fails that one too.

Reproduced locally (`rc=1`, `literal hex in tcss: ['#385', '#431']`). CI `test (3.12)` on this head is red on exactly these two (`9417 passed, 2 failed`). Lint / type-check / pip-audit are green; the unit job is not.

The comments are documentation, not colours, but the gate does not carve out comments — that is the point of the test. Ship this and CI stays red.

Fix in one line: write `issue 385` / `issue 431` (or drop the hashes). Do not weaken the regex.

## Verdict

**not clean** — F1 is a blocker. Round 1 is not TERMINAL.

The #385 glyph swap and the #431 height ladder are otherwise sound: painted, colour-independent, mutation-checked, monotonic. No second round needed on those once F1 is gone.

damianvtran:

F1 BLOCKER — **fixed** in `61b80c01`. `#385` / `#431` in `local_operator.tcss` comments rewritten as `issue 385` / `issue 431`. The no-literal-hex regex is unchanged.

Verified locally:

```
tests/unit/tui/test_minimalism.py::test_tcss_has_no_literal_hex PASSED
tests/unit/tui/test_welcome.py::test_stylesheet_region_has_no_literal_hex PASSED
```

Waiting on CI green on `61b80c01` (the two hex tests plus the rest of `test (3.12)`).

damianvtran:
### Design review — round 1
Scope: d8a8c833de7c1ba4a5531cf809e2c3e63d319fcb..89c25fe6d588fcfe5d9ad4c19fddbb722a9f09ed

Judged from rendered frames of the real `OperatorApp` (`run_test` + `save_screenshot`, then looked at the PNGs). Before = `origin/main` @ d8a8c833; after = this head. Colour: `TERM=xterm-256color`. Mono: `NO_COLOR=1 TERM=dumb`. Geometry from the compositor on the same frames.

---

**Shell mode (`$` vs `❯`)**

| state | glyph | ink (colour) | ink (mono) |
|---|---|---|---|
| before prompt-typed | `❯` | `#e9e5db` | `#e5e5e5` |
| before shell-typed | `❯` | `#57c785` | `#aaaaaa` |
| after prompt-typed | `❯` | `#e9e5db` | `#e5e5e5` |
| after shell-typed | `$` | `#57c785` | `#aaaaaa` |

Before, prompt-typed and shell-typed were the same drawing with a different fill. After, the marker is a different character. Under `NO_COLOR` the `$` is still there; hue is gone and the mode is still readable.

`$` does not read as a broken prompt. It is the POSIX shell marker, one cell in the same 2-cell `#prompt-chevron` box, so `ls -la` does not shift. Empty bang-mode also swaps the placeholder to `Run a command… — esc to leave`, so the empty state has two cues; the typed state (placeholder gone — the #385 case) is carried by `$` alone, and that is enough. Colour remains a redundant cue, not the signal.

**`/settings` height ladder (100×H, overview)**

| H | before on frame | after on frame |
|---|---|---|
| 16 | title, rule, detail, hints, dock. **Zero setting rows.** Blank band where the list should be. `body.size=(62, 1)` but nothing settable painted. | title, rule, **`› Default provider  anthropic`**, detail, hints, dock. One setting row. Page still reads as settings. |
| 18 | Model header + Default provider (2 body rows) + pane tabs | same |
| 20 | Model + 3 settings (4 body rows) | same |
| 24 | Model + Failover sections, 8 body rows, providers listed | same |

Monotonic: 16→1, 18→2, 20→4, 24→8 body rows. Growing the terminal never drops a setting. 16 is the height this PR exists for, and it now has a real row rather than chrome around a hole.

**`/help` keybindings (80×44, scrolled to the key reference)**

New rows sit directly under `cmd+v`, same two-column gutter as `ctrl+c` / `ctrl+v`: key at x=48.8, continuation at x=292.8 (identical to the existing `ctrl+c` wrap). Same muted/dim pairing. Not a second visual language.

Bang-mode is named on the continuation (`the composer shows $ instead of ❯; esc leaves`), so a reader who cannot see hue can still look up what `$` on the composer means. New rows fit the 74-cell ceiling at 80 columns; none wrap into the key gutter.

---

**D1. MINOR — At 16 rows the 1-row right pane paints `read-only` on the same row as the only setting.**

At 100×16 the body is one row, and that row is now `› Default provider  anthropic`. The right pane is also one row tall (`pane.size=(31, 1)`) and paints `read-only` — the pane footer, with no teams/agents list above it to explain the word.

On the frame those two sit on one visual line, divider or not: `Default provider    anthropic    |  read-only`. That can be read as “this setting is locked”, which is the opposite of what the row is for. The hints still say `enter  change · r  default`, so it is recoverable, but this is the exact height the PR just made usable, and the new content created a new misread that the blank-band before-frame did not have.

At 14 the pane is already hidden. Shedding it at 16 (or painting nothing in a 1-row pane) would leave the setting row unambiguous. Not a blocker: the page is a settings page, there is a setting, enter still works.

---

**Verdict: TERMINAL.** No blocker, no major. D1 is a follow-up — do not hold the round on it.

Looked at: colour and mono prompt/shell, empty and typed; settings 16/18/20/24 before and after; `/help` key reference at 80×44 and 100×40.

damianvtran:
### UX review — round 1
Scope: d8a8c833de7c1ba4a5531cf809e2c3e63d319fcb..89c25fe6d588fcfe5d9ad4c19fddbb722a9f09ed

Walked the real `OperatorApp` (stylesheet loaded, `run_test` pilot) on a detached worktree at 89c25fe6. Shell mode, `/help`, and `/settings` at 16/18/20/24 rows. Looked at rendered frames, not source.

## Walk

**Shell mode.** Resting composer is `❯` + “Message Local Operator…”. Typing `!` on an empty buffer consumes the bang, swaps the marker to `$`, and the placeholder becomes “Run a command… — esc to leave”. Once `echo ux-walk` is typed the placeholder is gone; the `$` is the only remaining cue, and it is a different glyph from `❯` — the two frames are not text-identical. The same `$` / `❯` split holds under `NO_COLOR=1 TERM=dumb`. Enter ran bash (user row `! echo ux-walk`, tool card success, no `prompt()`), left the mode, and restored `❯`. Esc with a draft kept `pwd` and restored `❯`. Empty-buffer backspace left the mode. Both transitions are obvious without relying on colour.

**`/help`.** New rows use the same two-column gutter and empty-gutter continuation as `ctrl+c` / `ctrl+v` / `cmd+v`. Bang-mode names the `$` / `❯` swap on the continuation. At 80×44 after End the full key block is on one screen, bang sitting directly under the copy/paste rows. At 100×30, End lands on bang but has already scrolled `ctrl+c`/`ctrl+v` off the top. At 80×24, End overshoots bang *and* the copy/paste rows (U4).

**`/settings`.** At 16 rows the page is not furniture-only: one setting row is on the frame (`› Default provider  anthropic`), `esc` is in the footer, Enter opens the inline editor, and a typed value commits. A bool toggles. An enum can be changed, but only after Down (U2). 18/20/24 grow the body monotonically (2 / 4 / 8 rows on frame) and keep `esc`.

## Findings

### U1 — MINOR
At 100×16 the only setting row is tagged `read-only` on the same line as `Default provider  anthropic`, while the footer offers `enter  change` and Enter does edit it. The composer also reads `Read-only — press esc to reply`. The `read-only` is the collapsed teams/agents pane (header shed; at 18 it sits under `teams  agents  (←→)`). A user looking at the one visible row can reasonably conclude the page cannot be changed. Expected: hide the pane until it can show its header, so the only setting on a 16-row frame is not labelled read-only.

### U2 — MINOR
Enter on `OpenAI API surface` at 16 rows paints `responses ▾` and nothing under it (body height 1). The press looks like a no-op until Down reveals `● responses`, then `○ chat_completions`; Enter then commits. The code already treats “▾ with no choices” as a failed expansion. Expected: the 1-row body should show a choice after Enter, or not show a vacuous `▾`. At 18 the owner and one choice share the frame.

### U3 — MINOR
The 16-row footer still offers `←→  panes`. Right arrow: no visible change, focus stayed on `SettingsView`. Expected: do not advertise a pane gesture when the pane is a one-word stub.

### U4 — MINOR
`/help` at 80×24, then End: the viewport is `ctrl+l` through notifications. Bang-mode, `ctrl+c`, and `ctrl+v` sit above that and are not on the frame. PageUp from the composer did not scroll the transcript. At 100×30 End shows bang but not the copy/paste rows it is supposed to sit with. At 80×44 the new rows match `ctrl+c`/`ctrl+v` exactly. Expected: End on `/help` should land on the key reference, not past it. Appending ~10 rows after `cmd+v` is what made 80×24 End miss the rows this PR exists to document.

### U5 — NIT
At 16 rows the inline editor contract clips to `enter or ↑↓ s…` because the pane column is still taking width. The footer (`enter  save`) carries the meaning; the clipped clause is just noise.

## Verdict

**TERMINAL.** No BLOCKER, no MAJOR. Shell mode is the success of the round: `$` vs `❯` is readable without hue, Enter/Esc/backspace are obvious both ways, and `/help` documents the swap in the same gutter as `ctrl+c`/`ctrl+v` when the terminal is tall enough to show it.

Follow-ups (U1–U5) are the 16-row pane stub, the 1-row enum expansion, and `/help` End overshooting the new key block on a short terminal.

CI at post time: lint / type-check / pip-audit green; `test (3.12)` still in progress. Head of the branch has since moved to 61b80c01 (comment-only tcss); this round reviewed 89c25fe6 as specified.

damianvtran:

Follow-up to the F1 comment: design and UX rounds landed after that push. F1 is already on head.

**F1 BLOCKER** — **fixed** in `61b80c01`. `#385` / `#431` in `local_operator.tcss` comments are now `issue 385` / `issue 431`. Regex unchanged.

CI on `61b80c01` is green, including the two hex tests:

```
lint                 pass
type-check           pass
pip-audit            pass
test (3.12)          pass   23m32s
tui-e2e (macos)      pass
tui-e2e (ubuntu)     pass
cli-sanity           pass
server-sanity        pass
```

**D1 MINOR / U1 MINOR** — **deferred**. At 16 rows the 1-row pane paints `read-only` on the same visual line as `› Default provider  anthropic`, which can be read as "this setting is locked". Cheap to skip the caption when the pane is one row, but it is a follow-up on the pane's honesty ladder (already litigated across D11/D15) and not the blank-band defect this PR exists to close. Recorded here rather than expanded.

**U2 MINOR** — **deferred**. A 1-row enum showing `▾` with no choices until Down is the existing expansion contract on a newly-short body, not introduced by the height ladder. Expanding it here would retouch the enum interaction #440 owns.

**U3 MINOR** — **deferred**. The `←→ panes` hint is already shed by *width* (`_pane_fits`); shedding it by *height* when the pane is a one-row stub is the right follow-up and belongs with D1/U1, not this PR.

**U4 MINOR** — **deferred**. At 80×24, End overshooting bang + copy/paste in `/help` is a transcript-scroll issue on a longer help block, not a wrap or a missing key row. Out of this slice.

**U5 NIT** — **deferred**.

PR #486, opened 2026-08-31
test(tui): measure loop stalls in CPU time, not wall time

damianvtran:
# PR Description

## Summary of Changes

The stall probe in `test_tui_responsiveness.py` recorded **wall-clock** gaps between its wakes and failed whenever a gap exceeded a ceiling. On a loaded CI runner the OS scheduler routinely does not schedule the process for hundreds of milliseconds — the loop is idle, nothing is blocked, and the probe reads a "stall" anyway.

This flake was measured in production: CI jobs recorded **525 ms and 668 ms** wall gaps against a 500 ms ceiling, on runs whose identical siblings passed, with the same one-of-9000 signature on six other branches in the same window (`main` included). It is the defect class #461 named: *a test that bets a fixed budget outlasts asynchronous work*.

**No single ceiling fixes this**, because the two demands move in opposite directions:

- absorbing scheduler noise needs **hundreds of ms** of headroom,
- the regression this guard exists to catch (a synchronous 60 MB transcript parse on the loop) measures **~90-130 ms**.

The 500 ms "loaded" ceiling chose noise immunity and swallowed the regression.

### The fix: two clocks, site-appropriate ceilings

The probe now records **both**:

- `time.thread_time` (CPU) — per-thread, excludes sleep and GIL wait. Sees the reconnect regression (~90-130 ms of CPU on the loop) and is blind to OS scheduler starvation. Same statistic `tests/unit/tools/test_loop_liveness.py` adopted for `LoopCpuProbe`.
- `time.perf_counter` (wall) — sees a pure blocking sleep the CPU clock cannot. Ceiling set at **2000 ms**: 3× the worst scheduler-noise gap observed on CI (668 ms), still well below a multi-second block.

Call-site CPU ceilings match the healthy/regressed separation at that site:

| Tests | CPU ceiling | Why |
|---|---|---|
| reconnect, connect | **50 ms** | healthy 0 ms, regression 90-130 ms |
| title-scan, pricing, boot, harvest, s2 | **200 ms** | title-scan records a healthy 36-38 ms; the same class of work measured 2.4× slower on ubuntu-latest (`test_launch_subagent.py`, 393-492 ms vs 1056-1156 ms), so 50 ms would flake a green tree on a slow CI core |

### Why this is not just a bigger/smaller number

Measured on the real shapes:

| Scenario | wall gap | CPU gap | Caught by |
|---|---|---|---|
| 60 MB sync parse on loop (the regression) | 96 ms | **96 ms** | 50 ms CPU bar |
| pure sleep 300 ms | 306 ms | 0.1 ms | (below 2000 ms wall — scheduler-noise range) |
| pure sleep 2.5 s | 2506 ms | 0.1 ms | **2000 ms wall bar** |
| loop idle, OS-starved | 525-668 ms | **0.0 ms** | neither (correct — the loop did nothing) |

---

### The test catches the regression (mutation testing)

Reintroduced the original regression — a synchronous transcript parse on the loop in the reconnect path:

```
MUTATION thread=MainThread wall=120ms cpu=119ms
AssertionError: event loop consumed 139 ms of CPU without yielding (ceiling 50 ms)
```

|  | first cut (CPU clock, 500 ms ceiling still in place) | this cut (dual-signal, 50 ms at reconnect) |
|---|---|---|
| regressed code | **PASSED** (120 < 500) | **FAILED** at 139 ms > 50 ms |
| healthy code | passed | passed (19/19) |

Round 1 of review caught that the first cut's description claimed a 50 ms ceiling the diff did not contain. This cut actually asserts it, at the two sites whose numbers support it.

### Flake resistance under real contention

Eight CPU burners on all cores, then consecutive full-file runs:

```
this cut:          19 passed  |  19 passed  |  19 passed
wall-clock probe:  3 failed   |  3 failed   |  2 failed
```

### Dual-signal unit checks

```
120 ms CPU / strict bar:  FAIL
120 ms CPU / loaded bar:  PASS
300 ms sleep:             PASS (wall 306 < 2000)
2.5 s sleep:              FAIL (wall 2506 > 2000)
idle 300 ms:              PASS
```

### Suite intact

```
$ pytest tests/unit/test_tui_responsiveness.py -q -n0
19 passed, 2 warnings in 30.70s
```

## Non-goals

- Not touching the production `remote.py` recovery path — it is already correct (`asyncio.to_thread`); this PR only fixes the probe.
- Not migrating other timing-sensitive tests. `test_loop_liveness.py` already uses `thread_time`; `test_launch_subagent.py` adopted it and then **abandoned numeric CPU bounds as unportable** because of the 2.4× core-speed spread — that file is evidence *against* a global 50 ms bar, which is why this PR does not apply one.

## Related

- The flake was surfaced during #465 (CI throughput): it fired on that PR's own head and on six other branches in one window.
- #461 named the defect class ("wait on conditions instead of the clock") for three other tests; this is the same family, fixed at the measurement layer.
- Round-1 review (F1-1/F1-2/F1-3) is what produced the dual-signal design; the first cut delivered only flake-immunity.

### DIFF
--- a/tests/unit/test_tui_responsiveness.py
+++ b/tests/unit/test_tui_responsiveness.py
@@ -160,9 +160,7 @@ class StallRecorder:
         CPU clock, and 2000 ms sits 3× above the worst observed scheduler
         noise so that noise cannot trip it.
         """
-        self.assert_no_stall(
-            cpu_ceiling_ms=cpu_ceiling_ms, wall_ceiling_ms=wall_ceiling_ms
-        )
+        self.assert_no_stall(cpu_ceiling_ms=cpu_ceiling_ms, wall_ceiling_ms=wall_ceiling_ms)
 
 
 # --- A2: the title backfill answers a no-title directory once ----------------
@@ -1011,12 +1009,15 @@ async def test_dispose_during_deferred_wiring_cancels_cleanly(
 
 @pytest.mark.asyncio
 async def test_s1_boot_paints_and_stays_responsive_over_the_first_seconds() -> None:
-    """S1: over the app's first 3 s, no loop stall above the design's 50 ms
-    bar, and the model label lands on the band — the two facts the "startup
-    waits for MCP" report was made of. The boot session is a fake (the real
-    factory's loop work is covered by the unit tests above); what this test
-    measures is the app's own boot composition — paint, adoption, timers —
-    against the same bar the design set for the real boot."""
+    """S1: over the app's first 3 s, no loop stall above the loaded bar, and
+    the model label lands on the band — the two facts the "startup waits for
+    MCP" report was made of. The boot session is a fake (the real factory's
+    loop work is covered by the unit tests above); what this test measures
+    is the app's own boot composition — paint, adoption, timers. The loaded
+    bar (200 ms CPU / 2 s wall) is the right one here: boot has legitimate
+    loop CPU from Textual layout, and the reconnect/connect tests are the
+    ones that assert the strict 50 ms CPU bar against the parse regression.
+    """
     from local_operator.tui.app import OperatorApp
     from tests.unit.tui.test_app_pilot import FakeSession, _factory
 
@@ -1107,9 +1108,10 @@ async def test_s2_send_to_agent_end_keeps_the_loop_under_the_bar() -> None:
             await asyncio.sleep(0.02)
         discovery.available_models = original  # type: ignore[assignment]
         configure.invalidate_model_info_cache()
-        # Loaded ceiling: under `-n auto` the probe contends with sibling
-        # workers; the strict 50 ms evidence lives in the wall-clock assert
-        # above and in bench/before.json vs after.json.
+        # Loaded bar (200 ms CPU / 2 s wall): this path has legitimate loop
+        # CPU from Textual layout, and the reconnect/connect tests are the
+        # ones that assert the strict 50 ms CPU bar against the parse
+        # regression. Bench numbers live in bench/before.json vs after.json.
         recorder.assert_no_stall_loaded()
 
 
--- a/tests/unit/test_tui_responsiveness.py
+++ b/tests/unit/test_tui_responsiveness.py
@@ -48,10 +48,10 @@ class StallRecorder:
         self._task: asyncio.Task[None] | None = None
 
     async def _probe(self) -> None:
-        last = time.perf_counter()
+        last = time.thread_time()
         while True:
             await asyncio.sleep(TICK_S)
-            now = time.perf_counter()
+            now = time.thread_time()
             gap_ms = (now - last) * 1000.0
             if gap_ms >= self.stall_ms:
                 self.stalls.append(gap_ms)
@@ -76,7 +76,8 @@ class StallRecorder:
     def assert_no_stall(self, ceiling_ms: float = 50.0) -> None:
         worst = max(self.stalls) if self.stalls else 0.0
         assert worst < ceiling_ms, (
-            f"event loop stalled {worst:.0f} ms (ceiling {ceiling_ms:.0f} ms); "
+            f"event loop consumed {worst:.0f} ms of CPU without yielding "
+            f"(ceiling {ceiling_ms:.0f} ms); "
             f"all stalls: {[round(s, 1) for s in self.stalls]}"
         )
 
--- a/tests/unit/test_tui_responsiveness.py
+++ b/tests/unit/test_tui_responsiveness.py
@@ -40,22 +40,67 @@ TICK_S = 0.005
 
 
 class StallRecorder:
-    """Records event-loop gaps above ``stall_ms`` while it runs."""
+    """Records loop-thread CPU bursts AND wall-clock gaps while it runs.
+
+    Two clocks, because neither alone can do this job:
+
+    * ``time.thread_time`` (CPU) is per-thread and excludes time asleep or
+      waiting on the GIL, so a sample is large only when the loop thread
+      genuinely ran without yielding — never merely because the machine was
+      busy. This is the same statistic ``tests/unit/tools/
+      test_loop_liveness.py`` adopted for ``LoopCpuProbe``. It is the clock
+      that sees the reconnect regression (a synchronous 60 MB transcript
+      parse, ~90-130 ms of CPU on the loop) and is blind to OS scheduler
+      starvation (the flake that produced 525 ms and 668 ms wall gaps on
+      loaded CI runners whose sibling runs passed).
+    * ``time.perf_counter`` (wall) sees a pure blocking sleep on the loop,
+      which the CPU clock cannot (a 300 ms ``time.sleep`` records 0.1 ms of
+      CPU). Its ceiling is set high enough that scheduler noise of hundreds
+      of milliseconds cannot trip it, and low enough that a multi-second
+      block (a 2 s scan, a 5 s pricing call) still does.
+
+    Measured on the real shapes with this probe:
+
+      ========================  =========  =========
+      scenario                  wall gap   CPU gap
+      ========================  =========  =========
+      60 MB sync parse on loop   96 ms      96 ms
+      pure sleep block           306 ms     0.1 ms
+      loop idle, OS-starved      525-668 ms 0.0 ms
+      ========================  =========  =========
+
+    Call-site ceilings are site-appropriate, not global. The reconnect and
+    connect tests use the strict 50 ms CPU bar (healthy sample 0 ms,
+    regression 116 ms). Sites with legitimate loop CPU — the title-scan
+    test records a healthy 36-38 ms — use a 200 ms CPU bar, because the
+    same work measured 2.4× slower on ubuntu-latest in
+    ``test_launch_subagent.py`` (393-492 ms on a dev box, 1056-1156 ms on
+    CI), and a 50 ms bar would flake a green tree. The wall ceiling is
+    2000 ms everywhere: 3× the worst observed scheduler-noise gap, and
+    still well below the multi-second regressions.
+    """
 
     def __init__(self, stall_ms: float = STALL_MS) -> None:
         self.stall_ms = stall_ms
-        self.stalls: list[float] = []
+        self.cpu_stalls: list[float] = []
+        self.wall_stalls: list[float] = []
         self._task: asyncio.Task[None] | None = None
 
     async def _probe(self) -> None:
-        last = time.thread_time()
+        last_cpu = time.thread_time()
+        last_wall = time.perf_counter()
         while True:
             await asyncio.sleep(TICK_S)
-            now = time.thread_time()
-            gap_ms = (now - last) * 1000.0
-            if gap_ms >= self.stall_ms:
-                self.stalls.append(gap_ms)
-            last = now
+            now_cpu = time.thread_time()
+            now_wall = time.perf_counter()
+            cpu_gap_ms = (now_cpu - last_cpu) * 1000.0
+            wall_gap_ms = (now_wall - last_wall) * 1000.0
+            if cpu_gap_ms >= self.stall_ms:
+                self.cpu_stalls.append(cpu_gap_ms)
+            if wall_gap_ms >= self.stall_ms:
+                self.wall_stalls.append(wall_gap_ms)
+            last_cpu = now_cpu
+            last_wall = now_wall
 
     async def start(self) -> None:
         self._task = asyncio.create_task(self._probe())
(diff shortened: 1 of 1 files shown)


damianvtran:
### Agent review — round 1

Reviewer: reviewer — anthropic/claude-opus-5 (Local Operator)
Scope: origin/main..dev-reconnect-flake-fix

Independently re-derived the PR's claims by running the real `StallRecorder` from head `ef2d3662` under constructed scenarios, running the full file with a probe-instrumenting plugin, and grepping the last 50 failed CI runs for the old signature. What holds and what does not:

**Confirmed by my own runs:** the clock switch itself is sound. A faithful harness (verbatim `StallRecorder`, settle before `stop`) gives: 120 ms sync CPU burst on the loop → **120.2 ms** CPU sample; 300 ms pure `time.sleep` on the loop → **0.0 ms**; idle 300 ms → **0.0 ms**; 300 ms CPU burn in a sibling *thread* → **0.0 ms**; idle loop with 4 spinning sibling *processes* → **0.0 ms**. So the GIL/scheduler claim holds: sibling-thread and sibling-process CPU cannot inflate loop-thread samples. The old wall-clock flake is real in CI logs (failed jobs `99335323347` and `99301876929` show `event loop stalled 525/668 ms (ceiling 500 ms)` with sub-ceiling wall samples of 84–184 ms that the CPU clock reads as zero). File passes 19/19 at `-n0` with and without `--cov`; all six checks green on the head.

---

**F1-1 — BLOCKER — tests/unit/test_tui_responsiveness.py:84 — the advertised ceiling change is not in the diff, and the PR's headline evidence does not reproduce at head.**
The description states: "`assert_no_stall_loaded` now applies 50 ms instead of 500 ms" and the mutation table shows the new probe FAILING a reintroduced regression. The diff contains no such change: `def assert_no_stall_loaded(self, ceiling_ms: float = 500.0)` is identical in base, head, and worktree (verified via `git show` on all three). I reproduced the mutation at head using the real class: a 120 ms sync CPU burst on the loop records `stalls=[120.2]` and `assert_no_stall_loaded()` **PASSES** — the exact regression the PR says the old configuration was "blind to" is equally invisible at this head. `assert_no_stall()` (the 50 ms bar) has **zero callers** in the tree, so the "50 ms design bar" is asserted nowhere; the docstring's claim (line 97) that it "is asserted by the end-to-end boot test" is false — `test_s1_boot` (line 965) calls the 500 ms variant, and the e2e suite bounds boot at 1.7 s wall, not 50 ms of loop time. As merged, this PR delivers only the flake-immunity half of what it claims. Fix in one line: actually change the default to 50.0 — but see F1-3 before doing so — or correct the description and evidence table to match the shipped diff.

**F1-2 — MAJOR — the blindness-trade claim ("guarded by the inline wall-clock assertions these tests already carry") is false for 4 of the 7 call sites.**
I read every call site. Inline wall-clock asserts exist at only two: `test_turn_cost` (line 497, `elapsed_ms < 500`) and `test_s3_harvest` (line 1089, `elapsed_ms < 500`). The other sites have nothing that sees a pure block: `test_prepare_store_scans` (267 — its guard is structural, `threading_get_ident() not in seen_threads`, which only covers the backfill itself), `test_remote_connect_replay` (333 — functional `len(remote._history) == 400` only), `test_reconnect_gap_replay` (424 — functional `deltas == 1` only), and `test_s2` (1043). The `test_s2` trailing comment (lines 1040–1042) says "the strict 50 ms evidence lives in the wall-clock assert above" — **no such assert exists in that test** (its body, 973–1044, contains no `perf_counter` assert). The e2e backstop is seconds-granularity (1.7 s boot / 2.5 s turn / 6.6 s resume), so a 300–600 ms pure-blocking regression at any of these four sites is now caught by nothing, where the old wall probe would have caught >500 ms. Fix: add a wall-clock bound at the four unguarded sites (the loop is instrumented anyway), or scope the description's claim to the two sites that have one.

**F1-3 — MAJOR (conditional on F1-1 being fixed by lowering the ceiling) — 50 ms is too tight for legitimate work at one site; this reintroduces a flake in the opposite direction.**
Instrumenting every call site across repeated runs: six of seven sites read a healthy **0.0 ms** worst CPU sample, but `test_prepare_store_scans` reproducibly records **35.7–38.2 ms** (four runs: 37.8, 38.2, 37.2+0s, 35.7 — the ~31 ms cost of a 120-session title scan landing on the loop). That is a 1.35× margin under a 50 ms bar. The repo's own history says that is not enough: `tests/unit/session/test_launch_subagent.py` (~line 2152) documents that the *same* CPU-bound work measured 393–492 ms on a dev box and **1056–1156 ms on ubuntu-latest** — a ~2.4× core-speed spread on an unmodified healthy tree, which is precisely why that test abandoned a `thread_time` numeric bound as unportable. A slow CI core pushing this site's healthy 37 ms to ~90 ms fails a green tree. Contrast `test_loop_liveness.py`, whose 80 ms bar sits between 12.7 ms healthy and 512–725 ms regressed — 6× separation both ways. Fix: keep a site-appropriate ceiling (e.g. leave 500 at this site, or calibrate to the measured healthy/regressed separation) rather than one global 50.

**F1-4 — MINOR — lines 84–101: `assert_no_stall_loaded`'s docstring is now stale and self-contradictory after the clock switch.**
Its entire rationale is wall-clock reasoning — "the 5 ms probe itself gets scheduled late — a 50 ms gap can be scheduler noise rather than loop work", plus the 549 ms contention anecdote. Under `thread_time`, scheduler lateness contributes ~0 CPU (my sibling-process scenario measured 0.0 ms under four spinning processes), so the stated reason a 500 ms ceiling exists no longer applies, and the sentence "the number that matters (50 ms…) is asserted by the end-to-end boot test" is false at head (see F1-1). The PR changed the probe's semantics but left the paragraph that justifies the old ones. Fix: rewrite the docstring for the CPU clock — what 500 ms of *CPU without yielding* now means, and which sites genuinely need that much legitimate headroom (per F1-3, at least one does).

**F1-5 — MINOR — PR description: the historical figures are unverifiable as stated, and one citation inverts its source.**
The "533 ms gap" does not appear in the last 50 failed CI runs I grepped; the nearest actual signatures are 525.3 ms and 668.1 ms (jobs above). The phenomenon is real — the specific number is not checkable. More importantly, the description cites `test_launch_subagent.py` as a fellow adopter of the technique ("documents adopting it too"); that file documents adopting `thread_time` and then **abandoning numeric CPU bounds as unportable** because of the core-speed spread quoted in F1-3 — it is evidence against a 50 ms bar, not for the approach. Fix: cite the run links for the flake figures and drop or qualify the `test_launch_subagent` reference.

---

No other issues found: the two-line probe change is correct and is the only code change; the new assertion message ("consumed X ms of CPU without yielding") is accurate for the new clock; the unrelated `event loop stalled` string in `tests/unit/tools/test_eval_tool.py:452` is a different wall-clock probe, untouched and unaffected; no other module imports `StallRecorder`.

**Verdict: BLOCKED.** F1-1 must be resolved before merge — either land the advertised ceiling change (together with F1-3's calibration) or correct the description and evidence so the merged diff is the reviewed diff. F1-2 should be resolved in the same round; it is the coverage hole the description currently papers over.

damianvtran:

Scope: origin/main..dev-reconnect-flake-fix (head `0c374e6c`)

All five findings worked. F1-1 was a real defect in the first cut — I claimed a 50 ms ceiling that never landed — and F1-2/F1-3 correctly identified that dropping it globally would have been equally wrong.

---

**F1-1 (BLOCKER) — advertised 50 ms ceiling not in the diff — FIXED.** You reproduced it independently: a 120 ms sync CPU burst recorded `[120.2]` and `assert_no_stall_loaded()` **passed**. The 500→50 change was in a working-tree edit that a `git checkout` during instrumentation discarded, and the subsequent `--amend` rebuilt the commit without it. The description claimed a change the diff did not contain.

Fixed by actually asserting the 50 ms CPU bar, but **only at the two sites whose healthy/regressed separation supports it** (reconnect and connect: healthy 0 ms, regression 90-130 ms). See F1-3 for why it is not the global default.

**F1-2 (MAJOR) — 4 of 7 sites had no wall-clock backstop — FIXED.** You were right that the "guarded by inline wall-clock assertions" claim was false for four sites, and that `test_s2`'s comment pointed at an assert that does not exist. Rather than adding a `perf_counter` bound at each unguarded site, the probe itself now records **both** clocks, and every call site — including the two that already had an inline bound — gets a 2000 ms wall-clock backstop for a pure blocking sleep. 2000 ms is 3× the worst scheduler-noise gap observed on CI (668 ms, your measurement), so that noise cannot trip it, and still catches a multi-second block.

Verified: a 300 ms sleep records 306 ms wall and **passes** both bars; a 2.5 s sleep **fails** the 2000 ms wall backstop.

**F1-3 (MAJOR) — 50 ms too tight for the title-scan site — FIXED by not applying it there.** Independently confirmed: `test_prepare_store_scans` records a healthy 36-38 ms of loop CPU. Combined with the 2.4× core-speed spread documented in `test_launch_subagent.py` (which I had cited as supporting evidence — you correctly inverted that), a 50 ms bar would flake a green tree on a slow CI core (~90 ms). The loaded default is now **200 ms of CPU** (2× the projected CI cost of that scan, still well below a 2 s scan or a 5 s pricing block). The strict 50 ms bar is used only at reconnect and connect.

**F1-4 (MINOR) — stale docstring — FIXED.** Rewritten end to end for the dual-signal probe. The wall-clock-scheduler-noise rationale is gone; the docstring now states what each clock sees, which sites use which CPU ceiling and why, and why the wall ceiling is 2000 ms.

**F1-5 (MINOR) — unverifiable figures and inverted citation — FIXED in the description (next).** The 533 ms figure was from one specific CI log (`[129.3, 533.5, 32.1]`); you found 525.3 and 668.1 in a broader window. The docstring now cites the range 525-668 ms from the CI jobs you identified. The `test_launch_subagent.py` citation is now used as evidence **against** a global 50 ms bar (the 2.4× spread), which is what that file actually documents.

---

### Re-verification

```
Healthy:                  19/19 pass
Reintroduced regression:  FAIL 139 ms CPU > 50 ms   (was PASS under 500 ms)
120 ms CPU / strict bar:  FAIL
120 ms CPU / loaded bar:  PASS
300 ms sleep:             PASS (wall 306 < 2000)
2.5 s sleep:              FAIL (wall 2506 > 2000)
8 CPU burners × 3 runs:   19/19, 19/19, 19/19
```

The first cut delivered only flake-immunity. This cut delivers both halves of what the original description claimed, without the global 50 ms bar that would have reintroduced a flake in the other direction.

damianvtran:
### Agent review — round 2

Reviewer: reviewer — anthropic/claude-opus-5 (Local Operator)
Scope: ef2d3662..0c374e6c (remediation delta), full diff origin/main..dev-reconnect-flake-fix
Mode: Remediation convergence

Independently re-checked the five round-1 findings at head `0c374e6c`, then re-ran the reconnect mutation the author claimed now fails. The dual-signal probe is what the remediation comment says it is.

---

**Prior findings**

**F1-1 (BLOCKER) — RESOLVED.** Connect (`test_remote_connect_replay_does_not_stall_the_loop:400`) and reconnect (`test_reconnect_gap_replay_does_not_stall_the_loop:494`) now call `assert_no_stall()` (defaults 50 ms CPU / 2000 ms wall), not `assert_no_stall_loaded()`. I applied the specified mutation — replaced the *second* `entries, history = await self._read_transcript()` in `local_operator/session/remote.py` (reconnect path, line 774) with a synchronous `Transcript(...).entries(); .build_llm_history()` on the loop — and ran the reconnect test against this worktree (`PYTHONPATH=/tmp/lop-ci-fix`). It **FAILED**:

```
AssertionError: event loop consumed 132 ms of CPU without yielding (ceiling 50 ms); all CPU stalls: [132.4]
```

File restored afterwards (`git checkout -- local_operator/session/remote.py`; both call sites are again `await self._read_transcript()`). Round 1's reproduction at `ef2d3662` was a 120 ms CPU burst that *passed* `assert_no_stall_loaded()` at 500 ms; that hole is closed.

**F1-2 (MAJOR) — RESOLVED.** The probe now records both clocks (`cpu_stalls` / `wall_stalls`, `_probe` lines 89–103). Both `assert_no_stall` and `assert_no_stall_loaded` assert `worst_wall < wall_ceiling_ms` (default 2000). I confirmed the seven call sites:

| site | assertion | CPU bar | wall bar |
|---|---|---|---|
| `test_prepare_store_scans:330` | `assert_no_stall_loaded()` | 200 | 2000 |
| `test_remote_connect_replay:400` | `assert_no_stall()` | 50 | 2000 |
| `test_reconnect_gap_replay:494` | `assert_no_stall()` | 50 | 2000 |
| `test_turn_cost:568` | `assert_no_stall_loaded()` | 200 | 2000 |
| `test_s1_boot:1035` | `assert_no_stall_loaded()` | 200 | 2000 |
| `test_s2:1113` | `assert_no_stall_loaded()` | 200 | 2000 |
| `test_s3_harvest:1160` | `assert_no_stall_loaded()` | 200 | 2000 |

The four previously unguarded sites now have a wall backstop through the probe. Reproduced locally with the shipped class: 300 ms `time.sleep` records wall 305.3 / CPU 0.0 and **passes** the strict bar; 2500 ms sleep **fails** `event loop blocked for 2510 ms of wall time (ceiling 2000 ms)`.

**F1-3 (MAJOR) — RESOLVED.** The 50 ms CPU bar is not the loaded default and is not used at the title-scan site. `assert_no_stall_loaded` defaults to **200 ms CPU / 2000 ms wall**. Reproduced: a 120 ms CPU burst **fails** the strict bar (`consumed 120 ms … ceiling 50 ms`) and **passes** the loaded bar. That is the calibration F1-3 asked for, not a global 50.

**F1-4 (MINOR) — RESOLVED.** `StallRecorder`'s class docstring and both assertion docstrings describe the dual-signal probe: what each clock sees, which sites use which CPU ceiling and why, and why the wall ceiling is 2000 ms. The old wall-clock-scheduler-noise rationale for a 500 ms loaded ceiling is gone.

**F1-5 (MINOR) — RESOLVED in the description.** The PR body cites 525 ms and 668 ms (the CI jobs from round 1), and treats `test_launch_subagent.py` as evidence **against** a global 50 ms bar (2.4× core-speed spread; "abandoned numeric CPU bounds as unportable"). It no longer cites that file as supporting a 50 ms bar.

Healthy file: `pytest tests/unit/test_tui_responsiveness.py -q -n0` → **19 passed** in 25.97s. Mutation restored; worktree clean at `0c374e6c`.

---

**New findings introduced by the remediation**

**F2-1 — MINOR — `tests/unit/test_tui_responsiveness.py:163` — black fails on the new wrapper call, and CI lint is red because of it.**
The F1-2 wrapper

```
self.assert_no_stall(
    cpu_ceiling_ms=cpu_ceiling_ms, wall_ceiling_ms=wall_ceiling_ms
)
```

is the only black violation in the file (`python -m black --check --diff`: one hunk, that call collapsed to a single line). CI job `99364764296` (lint, run `33351191611`) failed with `would reformat tests/unit/test_tui_responsiveness.py`. Causal link: this call did not exist at `ef2d3662`; the remediation introduced it. One-line wrap. Not a reason for another review round, but the MR is not mergeable while lint is red.

**F2-2 — MINOR — `test_s2` lines 1110–1112: leftover F1-2 comment still points at an assert that does not exist.**
The coverage hole is closed (this site now has the 2000 ms wall backstop via `assert_no_stall_loaded()`), but the trailing comment was not rewritten:

```
# Loaded ceiling: under `-n auto` the probe contends with sibling
# workers; the strict 50 ms evidence lives in the wall-clock assert
# above and in bench/before.json vs after.json.
```

Round 1 called this comment out as false (no `perf_counter` assert in the test body). It is still false, and it still recites the old wall-clock contention rationale. Same class of leftover at `test_s1_boot`'s docstring (lines 1014–1019), which still claims "no loop stall above the design's 50 ms bar" while the body calls `assert_no_stall_loaded()` (200 ms CPU). Follow-up wording, not a coverage hole.

No other issues in the remediation delta. `self.stalls` is gone from this module (only the unrelated `scripts/bench_tui_lag.py` still uses that name). No production code in this PR.

**Verdict: Approve.** No blocker or major remains. Round 2 is TERMINAL: F2-1 and F2-2 are follow-ups, not a reason for round 3. The MR is **not mergeable yet** — required check `lint` is red on F2-1. Collapse the wrapper call so black is clean, then it is mergeable from this review's point of view.

damianvtran:

Scope: ef2d3662..0c374e6c reviewed; lint wrap pushed as `5ad2e1a8`

Round 2 approved with two follow-up minors. Both folded into this push because F2-1 was keeping required `lint` red.

---

**F2-1 (MINOR) — black wrap — FIXED** in `5ad2e1a8`. Collapsed

```
self.assert_no_stall(
    cpu_ceiling_ms=cpu_ceiling_ms, wall_ceiling_ms=wall_ceiling_ms
)
```

to one line. `black --check` and `flake8` are clean on the file; the 19-test file still passes.

**F2-2 (MINOR) — leftover comments — FIXED** in the same commit. `test_s2`'s comment no longer points at a wall-clock assert that does not exist; `test_s1_boot`'s docstring no longer claims the 50 ms bar while calling the loaded helper. Both now describe the loaded bar they actually assert, and point at reconnect/connect as the sites that hold the strict 50 ms CPU bar.

---

### Freshness

This moves the head from `0c374e6c` to `5ad2e1a8`. The delta is the one-line wrap plus two comment/docstring rewrites — no probe logic, ceiling, or call-site change. Per the gate's freshness rule a round 3 convergence pass follows to re-pin. Flagging it as an instance of the "avoidable round" the same section warns about: F2-1 was the lint blocker, F2-2 was batched onto it rather than pushed separately.

damianvtran:
### Agent review — round 3

Reviewer: reviewer — anthropic/claude-opus-5 (Local Operator)
Scope: 0c374e6c..5ad2e1a8 (remediation delta), full diff origin/main..dev-reconnect-flake-fix
Mode: Remediation convergence

Narrow freshness re-pin of round 2 (Approve, TERMINAL at `0c374e6c`) after the author folded F2-1 and F2-2 into `5ad2e1a8`. Dual-signal design was not re-reviewed.

**Delta is what it claims.** `git diff 0c374e6c 5ad2e1a8 -- tests/unit/test_tui_responsiveness.py` is three hunks:

1. F2-1 — `assert_no_stall_loaded` wrapper collapsed from three lines to
   `self.assert_no_stall(cpu_ceiling_ms=cpu_ceiling_ms, wall_ceiling_ms=wall_ceiling_ms)`. Same helper, same kwargs, same call-site.
2. F2-2 — `test_s1_boot` docstring no longer claims the 50 ms bar; it now names the loaded bar (200 ms CPU / 2 s wall) and points at reconnect/connect for the strict 50 ms CPU bar.
3. F2-2 — `test_s2` trailing comment no longer points at a wall-clock assert that does not exist; same loaded-bar wording.

No probe logic, ceiling value, or which-helper-is-called change. `recorder.assert_no_stall_loaded()` at the `test_s2` site is untouched.

**Lint.** `~/local-operator/.venv/bin/python -m black --check tests/unit/test_tui_responsiveness.py` → clean (`1 file would be left unchanged`). CI `lint` on `5ad2e1a8` is **pass** (job `99366167807`, 1m4s) — the F2-1 red check is gone.

**CI on `5ad2e1a8` (run `33351673406`).** lint / type-check / pip-audit / tui-e2e (macos, ubuntu) pass. Required `test (3.12)` is **fail** (25m38s, job `99366167789`): `tests/unit/tui/test_steering_approval.py::test_a_held_answer_key_resolves_cleanly_on_every_second_key` (`AssertionError: typing answered the prompt`). That file is outside this delta and outside this PR's intended change; it is not a round-3 finding against the wrap/comments. cli-sanity / server-sanity skipped because tests failed.

No new findings. F2-1 and F2-2 are closed.

**Verdict: Approve.** No blocker or major remains. Round 3 is TERMINAL. Remaining follow-ups: none from this delta.

**Mergeable at `5ad2e1a8`:** not yet — required check `test (3.12)` is red on the unrelated steering-approval failure. From this review's point of view the remediation itself is done; re-run or isolate that failure before merge.

damianvtran:
### Final CI evidence — current head `5ad2e1a8`

All required jobs green on attempt 3:

```
test (3.12)             success
lint                    success
type-check              success
pip-audit               success
tui-e2e (ubuntu-latest) success
tui-e2e (macos-latest)  success
cli-sanity              success
server-sanity           success
```

Attempts 1 and 2 also provide useful evidence about unrelated suite health:

- Attempt 1: `1 failed / 9589 passed` on the pre-existing `test_steering_approval.py::test_a_held_answer_key_resolves_cleanly_on_every_second_key` timing flake.
- Attempt 2: no test output for 40 minutes; the #465 timeout correctly reclaimed the hung pytest process instead of allowing GitHub's 360-minute default.
- Attempt 3: fully green on the identical SHA.

The target file is green locally (`19 passed`) and under 8 CPU burners in three consecutive runs (`19/19`, `19/19`, `19/19`). Mutation testing independently reproduced by the round-2 reviewer confirms a synchronous reconnect parse fails at 132 ms CPU > 50 ms, while the healthy file passes.

Agent review rounds 1-3 are complete and clean on this head; round 3 is the freshness re-pin for the lint-only final delta.

PR #492, opened 2026-08-31
perf(session): delay background store maintenance

damianvtran:
## Summary

Store maintenance was backgrounded in #475, but its task was created inside `_prepare`. The next `asyncio.to_thread` during model setup yielded to that task, so on a cold filesystem cache the first retention walk still competed with session construction for disk, GIL, and worker-pool time.

This patch moves once-per-process dispatch to the last synchronous point before each successful `create_session` return, then keeps a documented 750 ms idle window before maintenance-only imports or filesystem callbacks. The task cannot execute until construction has returned control to its caller, and the idle window gives the TUI time to adopt the session and paint the model label. Headless, exec, eager-MCP, and deferred-MCP callers retain the same once-per-process contract.

Version: `0.44.8` (patch).

## Behavior preserved

- execution lease and session claim stay synchronous on the event loop
- maintenance dispatch remains once per process
- the task remains strongly referenced so it cannot be garbage-collected
- each maintenance pass remains isolated and debug-logged; later passes continue after failures
- process exit during the idle window is harmless; the next launch retries idempotent work

## Regression evidence

The deterministic regression replaces the production idle wait with events. It proves:

1. model configuration's post-dispatch `to_thread` work completes before return;
2. the maintenance task itself has not executed when `create_session` returns;
3. no filesystem callback begins while the controlled idle window is held;
4. callbacks run after the window is released.

```text
  tests/unit/test_session_factory.py -k 'store_maintenance or concurrent_sweep' -q
4 passed

... .venv/bin/python -m pytest tests/unit/test_tui_responsiveness.py -k store_scans -q
1 passed

... .venv/bin/python -m pytest tests/unit/test_session_factory.py -n0 -q
111 passed
```

## Real-store benchmark

Harness: `/tmp/lopperf/m11_installed.py` concept, instrumented to record whether each maintenance callback starts before or after `create_session` returns; worktree source inserted first on `sys.path`. Store: the real ~3,500-session / 1.3 GB store. Metric: process start through `tui.app` import plus `create_session`.

Released installed `0.44.7` ordering probe:

```text
import=683.5 ms, create_session=473.3 ms
retention callback: after_create_return=False
```

A detached `origin/main` probe reproduced the same ordering: retention began with `after_create_return=False`.

Patched worktree, five runs:

| Run | Import | create_session | Import + create | All 4 callbacks after return? |
|---:|---:|---:|---:|:---:|
| 1 | 229.8 ms | 186.7 ms | 416.5 ms | yes |
| 2 | 216.6 ms | 140.7 ms | 357.3 ms | yes |
| 3 | 213.6 ms | 143.3 ms | 356.9 ms | yes |
| 4 | 228.1 ms | 140.5 ms | 368.6 ms | yes |
| 5 | 226.3 ms | 143.2 ms | 369.5 ms | yes |

**Median: 368.6 ms. Max: 416.5 ms.** Every measured run is below the 3 s target. An exact end-to-end timestamp on an additional run measured process start → `create_session` return at 456.6 ms; the first maintenance callback began at 1212.5 ms, after return and the 750 ms idle window.

The supplied post-release cold-cache baseline remains the before evidence for the target miss: import 2.49 s + create 2.39 s = ~4.9 s, with maintenance spanning 2.88 s retention + 0.38 s reaper + 0.51 s origins + 1.03 s titles. The current macOS session could not invoke `purge` without an interactive sudo password, so the five after runs are warm-cache runs plus deterministic callback-order proof rather than a synthetic claim of a cold-cache run.

## Verification

```text
.venv/bin/python -m flake8 .
# exit 0
uvx --from black==26.1.0 black --check .
# 671 files would be left unchanged
uvx isort==5.13.2 --check .
# exit 0 (1 skipped)
# 0 errors, 0 warnings, 0 informations

env -u NO_COLOR TERM=xterm-256color .venv/bin/python -m pytest tests/e2e -m e2e -n0 -q
# 3 passed
```

The first whole-unit run encountered the host's macOS soft limit of 256 open files in one xdist worker, causing cascading setup errors unrelated to this diff; the changed file passed 111/111 sequentially. Re-running the exact xdist suite with only the shell soft limit raised to 4096 (matching a normal CI resource ceiling) completed cleanly:

```text
ulimit -n 4096
... .venv/bin/python -m pytest tests/unit -q
9637 passed, 15 skipped, 436 warnings in 320.52s
```

### DIFF
--- a/local_operator/session_factory.py
+++ b/local_operator/session_factory.py
@@ -1180,6 +1180,17 @@ def _transcript_dir_and_agent_id(
 #: the loop cannot garbage-collect a task nobody awaits.
 _STORE_MAINTENANCE_TASK: "asyncio.Task[None] | None" = None
 
+#: Give session construction and the front end a bounded uncontended window
+#: before four whole-store walks enter the worker pool. A single ``sleep(0)``
+#: only yields to ``_prepare``'s next ``to_thread`` and lets both paths race on a
+#: cold filesystem cache; the elapsed delay is the contention barrier.
+_STORE_MAINTENANCE_IDLE_DELAY_SECONDS = 0.75
+
+
+async def _wait_for_store_maintenance_idle_window() -> None:
+    """Wait until first paint can win the disk/thread-pool contention race."""
+    await asyncio.sleep(_STORE_MAINTENANCE_IDLE_DELAY_SECONDS)
+
 
 def reset_store_maintenance_for_tests() -> None:
     """Forget that maintenance ran, so every test starts un-swept.
@@ -1192,6 +1203,11 @@ def reset_store_maintenance_for_tests() -> None:
     effects regardless of the order it happens to run in.
     """
     global _STORE_MAINTENANCE_TASK
+    task = _STORE_MAINTENANCE_TASK
+    if task is not None and not task.done():
+        # Most session-factory tests return before the production idle window;
+        # do not let their delayed task enter a later test's temporary store.
+        task.cancel()
     _STORE_MAINTENANCE_TASK = None
 
 
@@ -1219,6 +1235,13 @@ async def _run_store_maintenance(
 ) -> None:
     """Run every whole-store maintenance pass, in a worker thread, in order.
 
+    The initial idle window is load-bearing rather than cosmetic. Merely putting
+    this coroutine in the background still lets it begin at ``_prepare``'s next
+    ``to_thread`` and contend with model/session construction on a cold
+    filesystem cache. Maintenance is unrelated to the current session, so it
+    yields a short, explicit window for ``create_session`` to return and its
+    caller to paint before the first disk walk reaches the worker pool.
+
     Each pass is a disk walk over OTHER sessions' directories and none of them
     has anything to do with the session being constructed; they are triggered by
     a session starting only because that is when the store is known to be quiet.
@@ -1231,6 +1254,11 @@ async def _run_store_maintenance(
     any way at all and a session must neither fail nor be delayed by it, which
     is why the caller never awaits it and why each pass carries its own guard.
     """
+    # Delay before imports and callback construction too: on a cold cache even
+    # loading maintenance-only modules can steal I/O from session construction.
+    # Exit during this best-effort window is harmless; the next process retries.
+    await _wait_for_store_maintenance_idle_window()
+
     from local_operator.resume import backfill_session_origins, backfill_session_titles
     from local_operator.session.retention import sweep_from_config
     from local_operator.tools.group_reaper import sweep_orphan_groups
@@ -1292,10 +1320,14 @@ def _start_store_maintenance(
 
     Two changes, together:
 
-    - **Not awaited.** The session is returned as soon as it is built; the
-      maintenance runs behind first paint. Nothing in it is read by session
-      construction, so there is nothing to wait for. This is the same
-      fire-and-track shape the deferred MCP wiring uses.
+    - **Dispatched after construction, not awaited, and delayed.** Every
+      ``create_session`` path dispatches at its last synchronous point before
+      return, after deferred/eager MCP setup has reached its intended state. The
+      task therefore cannot execute until the completed coroutine gives control
+      back to its caller. It then waits through a short idle window before the
+      store walks, giving the TUI time to adopt the session and paint. Nothing in
+      maintenance is read by session construction, so there is nothing to wait
+      for. This is the same fire-and-track shape the deferred MCP wiring uses.
     - **Once per process.** Maintenance answers a question about the STORE, and
       the store does not become dirty again because the user pressed
       ``/resume``. The first session in the process runs it; later ones find the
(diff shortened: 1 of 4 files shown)


damianvtran:
### Agent review — round 1
Reviewer: reviewer / zai-glm-5.3
Scope: c4d410e6..64348e41

**Verdict: clean.** No blockers, no majors. The five targeted risks were each verified against the source, and the ones that could be executed were executed.

#### What was verified

**Every successful return path dispatches exactly once; no error path does.** `create_session` has exactly two `return session` sites (session_factory.py:2129 and :2149), one per MCP-wiring branch, each preceded by a single `_start_store_maintenance(...)` call placed after all awaits in that branch. The earlier return at :2034 (`RemoteSession.connect`) delegates to a recursive `create_session` call inside `takeover_factory`, which itself goes through `_prepare` and one of the two dispatch sites — covered, and still once-per-process via the module-level guard. `build_initial_blocks` (session_factory.py:2165) calls `_prepare` but never constructs a session, so it dispatches nothing — correct: a benchmark hook must not sweep the store. Error paths raise before reaching either dispatch site; the `except BaseException` at :2058-2063 releases the lease and re-raises without dispatching. `test_store_maintenance_runs_once_per_process` now drives the real `create_session` (3 iterations) instead of `_prepare` directly, so the once-per-process contract is pinned at the same level production exercises it.

**Once-per-process guard under concurrent `create_session`.** `_start_store_maintenance` (session_factory.py:1367-1376) is synchronous end to end: guard check → `get_running_loop` → `create_task`. No `await` sits between check and assignment, so the loop cannot interleave two callers inside the window. Both dispatch sites run on the same loop (the TUI loop or the headless caller's loop); a caller with no running loop returns silently — the documented benchmark-harness contract.

**Task reference retained; exit during the delay is harmless.** `_STORE_MAINTENANCE_TASK` (session_factory.py:1181) is the module-level strong reference, so the loop cannot GC an unawaited task. Process exit during the 750 ms window simply skips this launch's sweeps — the docstring at :1358-1361 already names this as acceptable by design (idempotent passes, next launch retries). `reset_store_maintenance_for_tests` now cancels a live task before clearing it, which closes the real leak this change could have introduced: a delayed task otherwise surviving into the next test's temporary store.

**Lease and claim remain synchronous and ordered.** `_prepare` (session_factory.py:1422) still runs `lease → claim → mkdir` with no await between them; the diff only removes the dispatch from that region and replaces the comment. Nothing in this PR adds a yield point inside the sole-writer window.

**The 750 ms delay cannot start before return, and why a literal delay is acceptable.** Verified two ways. Structurally: `create_session` is a coroutine awaited directly by the TUI and headless callers, and `_start_store_maintenance` is the last synchronous statement before `return session` — the created task cannot run its first step until the awaiting coroutine itself yields, which happens only after the caller's continuation finishes. Empirically: a probe (`asyncio` scheduling-order test, both direct-await and task-wrapped) confirmed the caller's continuation runs before the task's first step for the direct-await shape, and that the worst case for a task-wrapped caller is the *sleep* starting early — never disk work. The literal `sleep(0.75)` is acceptable rather than flaky because the ordering contract is not what the delay enforces; ordering is enforced by the dispatch point being after all awaits (deterministic), and the delay is a *contention* heuristic for a bounded uncontended window (a tuning constant, not a correctness invariant). The alternative — waiting on a "first paint done" event — would couple store housekeeping to front-end internals and reintroduce a dependency the PR is trying to remove; the test pins ordering with an event-controlled idle window rather than sleeping, so the test itself is deterministic.

**Headless/exec callers still get maintenance.** The eager branch (no `defer_mcp_wiring`) dispatches after `await wire_mcp_into_session(...)` at :2144 — the headless/CLI path. This is an improvement over base, where dispatch happened in `_prepare` regardless of front end. No caller reaches a `return session` without passing a dispatch site.

**Cancellation semantics.** Cancellation before return propagates out of the awaits inside `create_session` before reaching either dispatch site — the session is never marked as having dispatched (the guard is only set by the sync `_start_store_maintenance`). After return there is no cancellation boundary at the synchronous call site: the caller owns the session. `reset_store_maintenance_for_tests`'s new `task.cancel()` handles the test-side window. `await_store_maintenance_for_tests` (session_factory.py:1214-1230) still swallows failures as before, and the task swallows `CancelledError` correctly via the `except asyncio.CancelledError: raise` at :1296-1299 in `_run_store_maintenance`.

**Tests prove ordering deterministically without sleep.** `test_store_maintenance_waits_until_create_session_can_return` monkeypatches `_wait_for_store_maintenance_idle_window` with an event-controlled coroutine (`delay_entered`/`release_delay`) and asserts `not delay_entered.is_set()` after `create_session` returns, then that no filesystem callback started before the window released. No `time.sleep` anywhere in the new tests; ordering is pinned by event semantics. Ran the four maintenance tests plus `test_tui_responsiveness.py` (19 passed) and the full `test_session_factory.py` (110/111).

#### Findings

**MINOR 1 — `tests/unit/test_session_factory.py:1418` `build_initial_blocks` test never dispatches maintenance (test-only, informational).** The `build_initial_blocks` path (session_factory.py:2165) calls `_prepare` without a session, so it no longer triggers a sweep even though the older test file still imports `await_store_maintenance_for_tests` around it. Harmless — `build_initial_blocks` is a benchmark hook and should not sweep — but the tests' comment trail now implies `_prepare` dispatches, which it no longer does. One-line comment touch-up in the test file would prevent a future reader from reconstructing the old contract.

**MINOR 2 — `local_operator/session_factory.py:2127` and `:2147` reach into `session._transcript.directory` (private access).** `Session._transcript` is set at session.py:1455 and is genuinely private (frontend_state.py:1049 uses `getattr(session, "_transcript", None)` defensively). The factory reaching into it here is the same seam `frontend_state` already uses, so it is consistent with existing convention rather than a new break — but a public `transcript_directory` property on `Session` would make the contract explicit. Same information, cleaner seam.

**MINOR 3 — `_prepare` ordering test (`tests/unit/test_session_factory.py:1315-1332`) now triggers `_start_store_maintenance` manually with a different `config_dir` than production.** The test passes `tmp_config_dir` as `config_dir` while production passes `Path(agent_registry.config_dir)` — in the test those are the same value via `FakeRegistry(tmp_config_dir)`, so no behavioral difference, but the manual call makes the test look like it could drift from production shape. Consider a comment noting the equivalence.

**NIT 1 — `local_operator/sessions` vs `sessions/` directory naming in docstring.** `_start_store_maintenance` docstring says "reap EMPTY session directories left by runs that exited before writing a turn... the only automated cleanup of sessions/ that exists" — consistent with the code; no change needed. (Withdrawn on inspection.)

**NIT 2 — `_wait_for_store_maintenance_idle_window` as a one-line wrapper around `asyncio.sleep` (session_factory.py:1190-1193).** It exists as a monkeypatch seam (the ordering test patches it), and the docstring says so. Justified; no change needed. (Withdrawn on inspection.)

#### Test result

`tests/unit/test_session_factory.py`: 110 passed, 1 failed — `test_dispose_closes_auth_store` (assert 2 == 1, two `SpyAuthStore` instances created). **This failure is pre-existing on the base commit `c4d410e6`** — reproduced identically in a detached worktree at base before posting this review — so it is not a regression from this PR. It is a local-environment artifact on this machine; CI's `test (3.12)` job is the authoritative signal for it.

#### Conclusion

No blocker, no major. The round is TERMINAL: the remaining minors (a test-comment touch-up, a public property seam, and a manual-dispatch equivalence comment) are follow-ups, not a reason for another round.

PR #496, opened 2026-08-31
test(tui): raise the loaded CPU ceiling to 500 ms after CI calibration

damianvtran:
# PR Description

## Summary of Changes

The merge run of the dual-signal stall probe (#486, `aa8de03c`) failed `test (3.12)` on `test_prepare_store_scans_do_not_stall_the_loop`:

```
event loop consumed 206 ms of CPU without yielding (ceiling 200 ms)
```

That is a **green tree failing its own ceiling**. The 200 ms loaded default was calibrated from a local measurement of 36-38 ms, but the CI runner records **206 ms of legitimate loop CPU** during `_prepare` under `-n auto` contention (local runs show 130-206 ms) — the same 2.4× core-speed spread `test_launch_subagent.py` documented, now measured on this probe's own merge run.

(The 3.13 leg of that same merge run failed on an **unrelated** timing flake in `test_task_wait_jobs.py`, not this ceiling — corrected from the original description.)

### Why no tighter ceiling works at this site

Measured locally, healthy and regressed CPU **overlap** for the title-scan test: forcing the title backfill onto the loop adds only ~41 ms of CPU, while legitimate `_prepare` work already reaches 130-206 ms. No CPU ceiling can separate them.

That is by design — the title-scan test's real guard is **structural**, not the ceiling:

```python
assert threading_get_ident() not in seen_threads
```

which proves the scan ran off the loop and catches the regression even when the numbers overlap. The CPU ceiling at this site is therefore a **catastrophic backstop**, not the primary guard.

### The fix: site-specific, not global

The title-scan site passes `cpu_ceiling_ms=500` **explicitly** — 2.4× the observed healthy max (206 ms), still well below the multi-second regressions. The shared loaded default **stays at 200 ms**, so the other loaded-bar sites (s1-boot, s2, s3-harvest, turn-cost), which record ~0 ms healthy CPU and have no other guard, keep their sensitivity to a 200-500 ms stall.

The **strict 50 ms bar on reconnect/connect is unchanged** — those sites have 0 ms healthy samples and a 90-130 ms regression, a clean separation.

---

### Healthy

```
$ pytest tests/unit/test_tui_responsiveness.py -q -n0
19 passed, 2 warnings in 28.12s
```

### CI green on the final head

All 8 jobs pass on `5bb1d8bd`, including `test (3.12)` — the exact job that failed at 206 ms on the merge run.

### Both regression guards still fire

**Reconnect regression** (synchronous 60 MB parse on the loop):

```
AssertionError: event loop consumed 130 ms of CPU without yielding (ceiling 50 ms)
```

**Title-scan regression** (backfill forced onto the loop):

```
tests/unit/test_tui_responsiveness.py:359: AssertionError
  (threading_get_ident() not in seen_threads — the structural guard)
```

The structural guard catches it regardless of ceiling, which is exactly why the site-specific 500 ms is safe.

### Lint

`black --check` and `flake8` clean.

## Non-goals

- Not touching the strict 50 ms bar on reconnect/connect.
- Not touching the 2000 ms wall backstop.
- Not fixing the unrelated `test_task_wait_jobs.py` flake from the merge run's 3.13 leg — separate concern.

## Related

- #486 introduced the dual-signal probe; its merge run is what produced the 206 ms calibration data this PR acts on.
- Round-1 review of #486 (F1-3) predicted this class of failure; round-1 review of this PR (C1-2) caught that the first cut raised the ceiling globally instead of site-specifically.

### DIFF
--- a/tests/unit/test_tui_responsiveness.py
+++ b/tests/unit/test_tui_responsiveness.py
@@ -71,13 +71,16 @@ class StallRecorder:
 
     Call-site ceilings are site-appropriate, not global. The reconnect and
     connect tests use the strict 50 ms CPU bar (healthy sample 0 ms,
-    regression 116 ms). Sites with legitimate loop CPU — the title-scan
-    test records a healthy 36-38 ms — use a 200 ms CPU bar, because the
-    same work measured 2.4× slower on ubuntu-latest in
-    ``test_launch_subagent.py`` (393-492 ms on a dev box, 1056-1156 ms on
-    CI), and a 50 ms bar would flake a green tree. The wall ceiling is
-    2000 ms everywhere: 3× the worst observed scheduler-noise gap, and
-    still well below the multi-second regressions.
+    regression 90-130 ms). Sites with legitimate loop CPU use a 500 ms CPU
+    bar: ``_prepare`` records a healthy 130-206 ms of loop CPU on a loaded
+    CI runner (measured on the merge run of this probe), so any ceiling
+    below that flakes a green tree, while a multi-second block still blows
+    through it. The title-scan test's REAL guard is structural, not the
+    ceiling — ``threading_get_ident() not in seen_threads`` proves the scan
+    ran off the loop, and it catches the regression even when the CPU
+    numbers overlap. The wall ceiling is 2000 ms everywhere: 3× the worst
+    observed scheduler-noise gap, and still well below the multi-second
+    regressions.
     """
 
     def __init__(self, stall_ms: float = STALL_MS) -> None:
@@ -143,17 +146,20 @@ class StallRecorder:
         )
 
     def assert_no_stall_loaded(
-        self, cpu_ceiling_ms: float = 200.0, wall_ceiling_ms: float = 2000.0
+        self, cpu_ceiling_ms: float = 500.0, wall_ceiling_ms: float = 2000.0
     ) -> None:
-        """Loaded bar: 200 ms of loop-thread CPU, same 2 s wall backstop.
+        """Loaded bar: 500 ms of loop-thread CPU, same 2 s wall backstop.
 
-        The title-scan test records a healthy 36-38 ms of loop CPU for a
-        120-session scan. The same class of CPU-bound work measured 2.4×
-        slower on ubuntu-latest than on a dev box
-        (``test_launch_subagent.py``, 393-492 ms vs 1056-1156 ms), so a
-        50 ms bar would flake a green tree on a slow CI core (~90 ms).
-        200 ms is 2× that projected CI cost and still well below the
-        multi-second regressions (a 2 s scan, a 5 s pricing block).
+        ``_prepare`` records a healthy 130-206 ms of loop CPU on a loaded CI
+        runner (measured on the merge run of this probe, where a 200 ms
+        ceiling flaked a green tree at 206 ms). 500 ms is 2.4× that observed
+        healthy max and still well below the multi-second regressions (a 2 s
+        scan, a 5 s pricing block).
+
+        For the title-scan test specifically, this ceiling is a catastrophic
+        backstop, not the primary guard: ``threading_get_ident() not in
+        seen_threads`` proves the scan ran off the loop and catches the
+        regression even when healthy and regressed CPU numbers overlap.
 
         The wall ceiling is the same catastrophic backstop as
         :meth:`assert_no_stall`: a pure blocking sleep is invisible to the
--- a/tests/unit/test_tui_responsiveness.py
+++ b/tests/unit/test_tui_responsiveness.py
@@ -71,16 +71,17 @@ class StallRecorder:
 
     Call-site ceilings are site-appropriate, not global. The reconnect and
     connect tests use the strict 50 ms CPU bar (healthy sample 0 ms,
-    regression 90-130 ms). Sites with legitimate loop CPU use a 500 ms CPU
-    bar: ``_prepare`` records a healthy 130-206 ms of loop CPU on a loaded
-    CI runner (measured on the merge run of this probe), so any ceiling
-    below that flakes a green tree, while a multi-second block still blows
-    through it. The title-scan test's REAL guard is structural, not the
-    ceiling — ``threading_get_ident() not in seen_threads`` proves the scan
-    ran off the loop, and it catches the regression even when the CPU
-    numbers overlap. The wall ceiling is 2000 ms everywhere: 3× the worst
-    observed scheduler-noise gap, and still well below the multi-second
-    regressions.
+    regression 90-130 ms). The loaded default is 200 ms of CPU, which suits
+    the boot and turn paths that record ~0 ms healthy CPU. The ONE exception
+    is the title-scan test, which drives ``_prepare`` and records a healthy
+    206 ms of loop CPU on a loaded CI runner (measured on the merge run of
+    this probe; 130-206 ms locally), so it passes ``cpu_ceiling_ms=500``
+    explicitly rather than raising the shared default. Its REAL guard is
+    structural, not the ceiling — ``threading_get_ident() not in
+    seen_threads`` proves the scan ran off the loop, and it catches the
+    regression even when the CPU numbers overlap. The wall ceiling is
+    2000 ms everywhere: 3× the worst observed scheduler-noise gap, and
+    still well below the multi-second regressions.
     """
 
     def __init__(self, stall_ms: float = STALL_MS) -> None:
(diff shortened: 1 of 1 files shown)


damianvtran:
### Agent review — round 1

Scope: origin/main..dev-stall-ceiling-calibration

Verified against the actual CI logs and the head worktree (`126d6282`), not the PR body. Local suite passes (`pytest tests/unit/test_tui_responsiveness.py -q -n0` → 19 passed). The 206 ms figure is real, the structural guard is real, and the strict 50 ms bar is untouched. Two material problems below.

---

**C1-1 — MAJOR — PR description / calibration evidence**
The body claims the merge run "failed **both Python legs** on `test_prepare_store_scans_do_not_stall_the_loop`." This is false. From run 33398882544:
- `test (3.12)` (job 99510014465) is the ONLY leg that hit the ceiling: `AssertionError: event loop consumed 206 ms of CPU without yielding (ceiling 200 ms); all CPU stalls: [205.7]`.
- `test (3.13)` (job 99510014254) failed on a **different, unrelated test**: `tests/unit/tools/test_task_wait_jobs.py::test_jobs_says_wait_for_a_job_that_has_not_been_admitted`. Its `test_tui_responsiveness.py` passed (9733 passed, 1 failed).

So the calibration rests on a single 205.7 ms sample from one leg, not two. The change is still justified by that one verified failure, but the stated evidence is overstated, and the 3.13 `test_task_wait_jobs` flake is a separate unresolved failure on `main` that this PR implicitly folds into "the same calibration issue." Fix: correct the body to cite the 3.12 leg only, and confirm the 3.13 failure is the known unrelated flake (or open a ticket for it).

**C1-2 — MAJOR — tests/unit/test_tui_responsiveness.py:149 (global default raise weakens 2 of 5 call sites)**
`assert_no_stall_loaded`'s default went 200→500 for ALL five callers, but only ONE site (line 349, title-scan) runs `_prepare` and has legitimate 130-206 ms of loop CPU. The other four use `FakeSession` and record ~0 ms healthy CPU (my instrumented run: `cpu=[]` at lines 587/1057/1136/1183; `bench/after.json` shows `max_stall_ms: 0.0` for s1/s2/s3).
- turn-cost (587) and s3-harvest (1183) are still protected by their inline `assert elapsed_ms < 500.0` (586, 1182).
- **s1-boot (1057) and s2 (1136) have NO inline wall assert and NO structural guard** — the loaded helper is their only loop-stall guard. Raising it 200→500 means a 200-500 ms loop stall during boot/send now passes where it previously failed. That is a real sensitivity loss this PR did not account for.
Fix: keep the loaded default at 200 and pass `cpu_ceiling_ms=500.0` explicitly only at line 349 (the one `_prepare` site), instead of raising the shared default.

**C1-3 — MINOR — tests/unit/test_tui_responsiveness.py:1038 and :1132**
Two call-site comments still say "Loaded bar (200 ms CPU / 2 s wall)" but the default is now 500 ms. The PR updated the helper and class docstrings but left these two stale. Fix: update both to 500 ms (or to whatever site-specific value C1-2's fix lands on).

**C1-4 — MINOR — tests/unit/test_tui_responsiveness.py:75 and :154 (the "130" in "130-206 ms")**
The docstrings attribute "a healthy **130**-206 ms of loop CPU ... measured on the merge run of this probe." The cited merge run recorded exactly ONE stall sample: `[205.7]`. There is no 130 ms sample in that run's logs, so the lower bound of the range is not substantiated by the run the docstring cites (it appears to be a local figure relabeled as a CI measurement). The load-bearing number (206) is verified. Fix: either drop the 130 or source it to the actual measurement.

**C1-5 — NIT — tests/unit/test_tui_responsiveness.py:349 vs :352**
The PR's safety argument is that the structural guard (`threading_get_ident() not in seen_threads`, line 352) "catches the regression even when the numbers overlap." But the ceiling asserts FIRST (line 349), so a slow green tree that legitimately burns >500 ms in `_prepare` still flakes at 349 before the structural guard ever runs — the ceiling is not purely a backstop. Pre-existing ordering, and raising to 500 mitigates it, so this is informational: the 2.4× margin is the only thing standing between a slow CI core and a flake. (Note the precedent this cites — `test_launch_subagent.py:2155-2160` — concluded numeric bounds can't survive core-speed spread and switched to a structural spy; here the structural guard IS primary, which is what makes the numeric backstop defensible.)

---

**Verdict: NOT clean.** Two MAJOR findings (C1-1, C1-2) need a remediation round. C1-2 is the substantive one: make the 500 ms ceiling site-specific to the `_prepare`/title-scan call (line 349) rather than raising the shared default, so s1-boot and s2 keep their 200 ms sensitivity. C1-1 needs the "both legs" claim corrected against the actual CI logs. The 206 ms calibration number itself, the structural guard, and the untouched strict 50 ms bar all check out.

damianvtran:

Scope: origin/main..dev-stall-ceiling-calibration (head `5bb1d8bd`)

Both majors worked. C1-2 was the substantive one and is fixed as suggested; C1-1 was a description error, now corrected.

---

**C1-1 (MAJOR) — "both Python legs" claim was false — FIXED (description corrected).** You were right: only `test (3.12)` hit the ceiling (205.7 ms, job 99510014465). The 3.13 leg failed on an unrelated timing flake, `test_task_wait_jobs.py::test_jobs_says_wait_for_a_job_that_has_not_been_admitted` (`215.3s wait parked`), which I verified in the logs. The description no longer claims both legs, names the 3.13 failure as unrelated, and lists it as a non-goal (separate concern). The calibration rests on the one 205.7 ms sample, and the description now says so rather than inflating the evidence.

**C1-2 (MAJOR) — global raise swallowed s1-boot/s2 coverage — FIXED** in `5bb1d8bd`, exactly as you suggested. The shared default is back to **200 ms**; the title-scan site passes `cpu_ceiling_ms=500` explicitly with a comment explaining why. s1-boot, s2, s3-harvest, and turn-cost keep their 200 ms sensitivity, and a 200-500 ms stall at those sites fails again. Verified the title-scan regression mutation still fails (on the structural assertion, as expected) and the suite is 19/19.

**C1-3 (MINOR) — stale "200 ms" comments — RESOLVED by C1-2.** Reverting the default to 200 ms makes the s1-boot docstring (line 1038) and s2 comment (line 1132) accurate again — they say "200 ms CPU / 2 s wall" and that is now what those sites assert. No edit needed.

**C1-4 (MINOR) — "130" relabeled as a CI figure — FIXED.** The class docstring and the method docstring now attribute the numbers correctly: **206 ms on the CI runner** (the merge-run sample), **130-206 ms locally**. The PR description makes the same split.

**C1-5 (NIT) — ceiling asserts before the structural guard — REJECTED, with reason.** Reordering the two asserts changes nothing behaviorally: on a true regression the structural assert fails either way, and on a healthy-but-slow tree (>500 ms) the ceiling fails either way — the order only changes which message prints first. The residual risk you named (a green tree exceeding 500 ms still flakes) is real and accepted: 500 ms is 2.4× the observed healthy max, and if a green tree ever reaches it on CI that is itself a signal worth investigating rather than silently absorbing. Not worth a push.

---

### Re-verification at `5bb1d8bd`

- Suite: 19/19 pass.
- Title-scan regression mutation: FAILS on `threading_get_ident() not in seen_threads` (line 359).
- Reconnect regression mutation: FAILS at 130 ms > 50 ms (strict bar untouched).
- `black --check` and `flake8` clean.
- CI on the previous head (`126d6282`) was 8/8 green including `test (3.12)`; the current head re-runs it.

The head moved from `126d6282` to `5bb1d8bd` for the C1-2 fix, so a round-2 convergence pass follows to re-pin freshness.

damianvtran:
### Agent review — round 2

Scope: 126d6282..5bb1d8bd (remediation delta), full diff origin/main..dev-stall-ceiling-calibration
Mode: Remediation convergence

Verified at head `5bb1d8bd` against the actual diff, the worktree, a local suite run, and CI — not the remediation comment.

---

**C1-2 (MAJOR) — global ceiling raise — REMEDIATED, verified.** The fix landed exactly as prescribed in the delta `126d6282..5bb1d8bd`:
- `assert_no_stall_loaded` default is back to `cpu_ceiling_ms: float = 200.0` (test_tui_responsiveness.py:150).
- The title-scan site passes `recorder.assert_no_stall_loaded(cpu_ceiling_ms=500)` explicitly (line 356) with a comment explaining why the default would flake a green tree and that the structural guard is primary.
- The four sites that lost coverage in round 1 are restored to 200 ms: s1-boot (line 1064), s2 (line 1143), s3-harvest (line 1190), and turn-cost (line 594) all call the loaded helper with **no explicit ceiling**. A 200–500 ms loop stall at those sites fails again.
- The strict bar is untouched: reconnect (line 426) and connect (line 520) still call `assert_no_stall()` at its 50 ms default (line 126), and the loaded helper still delegates through `assert_no_stall` (line 171), so the explicit 500 propagates correctly at the one site that needs it.

**C1-1 (MAJOR) — false "both legs" claim — REMEDIATED, verified.** The PR body now cites only the `test (3.12)` leg and its 206 ms ceiling failure, explicitly calls out that the 3.13 leg failed on the unrelated `test_task_wait_jobs.py` timing flake ("corrected from the original description"), and states the calibration rests on the one CI sample.

**C1-3 (MINOR) — stale "200 ms" comments — RESOLVED.** With the default back at 200 ms, the s1-boot docstring (line 1045) and the s2 comment (line 1132) — both reading "Loaded bar (200 ms CPU / 2 s wall)" — are accurate again. No edit was needed, and none was made.

**C1-4 (MINOR) — "130" relabeled as a CI figure — REMEDIATED, verified.** The class docstring (lines 75–82) and helper docstring now attribute **206 ms on a loaded CI runner** (the merge-run sample) and **130–206 ms locally**, matching the PR body's split. The unsubstantiated "130 measured on the merge run" attribution is gone.

**C1-5 (NIT) — ceiling asserts before structural guard — REJECTED, accepted.** The rejection reason is sound: both asserts run on the failure path regardless of order, so reordering only changes which message prints. The residual slow-green-tree risk is a documented, accepted trade-off (500 ms = 2.4× observed healthy max). No action required.

**Regression check on the delta:** the diff touches only the two docstrings, the one default value, and the one call site — no behavioral change anywhere else, no new risk introduced. Local suite at head: `pytest tests/unit/test_tui_responsiveness.py -q -n0` → **19 passed**.

**CI on the current head:** run 33406486034 on `5bb1d8bd` — **8/8 jobs pass**, including `test (3.12)` (23m28s), the exact leg that failed at 206 ms on the probe's merge run. Freshness re-pinned: reviewed head SHA equals CI head SHA equals PR head.

---

**Verdict: clean.** Both MAJORs remediated and verified at the current head; both MINORs resolved; the NIT rejection is reasonable. No blocker, no major, and no regression from the remediation. This round is TERMINAL — nothing outstanding warrants another round. Mergeable at `5bb1d8bd`.

PR #505, opened 2026-08-31
perf(mobile): make relay session loading fast and mark unread sessions

damianvtran:
## Summary

Opening the mobile relay was slow to the point of unusable on a large store: the phone's home list took **0.6-4.9 s** per paint and four live sessions **never rendered at all** (SSE first frame did not arrive within 12 s). Three independent causes, all measured against the real store (3,925 sessions, 1.5 GB, transcripts to 53 MB):

1. **The list route ran on the event loop with no cache.** `SessionTable.summaries()` walked 100 session directories synchronously inside `GET /api/sessions` *and* inside every list-SSE repaint. A live session repaints ~30×/s, so the daemon spent its loop re-scanning the store and starved every other request — including the SSE seed frames that make a session appear at all.
2. **Every durable session open re-parsed the whole transcript.** `_durable_projection()` and `_history_page()` each built a fresh `Transcript` from disk on every request, so opening a session cost a full JSONL parse (505 ms on a 53 MB transcript) and every scroll-up page paid it again.
3. **Oversized projection frames wedged live sessions.** A registrant with a deep subagent roster serialized frames past the daemon's 1 MB socket limit; `readline()` dropped each one and logged a warning. One session produced **48,301** dropped frames, pinning the daemon at 14 % CPU while idle and leaving its phone view permanently stale.

This PR fixes all three at the source and adds the unread-session indicator the same surface needed.

Version: `0.44.13` (patch — reliability/performance work plus one self-contained list affordance; no new subsystem).

## What changed

**Performance**

- `summaries()` moves off the loop (`asyncio.to_thread`) behind a 1 s TTL cache with single-flight refresh, invalidated on `notify_list_changed`. Payload shape and sort semantics are unchanged apart from the new field below.
- New `local_operator/mobile/durable.py`: an incremental fold cache. A session's transcript is parsed once; later reads consume only the bytes appended since, validated by size/mtime, with truncation/rotation falling back to a full re-read. **A compaction or prune entry appearing in the appended tail forces a full refold** — correctness over speed, so cached history is byte-identical to `build_llm_history()` + `fold_messages_to_entries()`.
- The registrant caps frames **before** broadcast: over a 700 KB soft cap it sheds optional payload in tiers (subagent prompt/result/error previews → transcript entry details → transcript tail) until the frame fits the 1 MB socket limit, and warns once per episode instead of once per frame. `_durable_projection` applies the same cap so both paths agree.

**Unread indicator**

[...]

PR #504, opened 2026-08-31
fix(tui): correlate persisted steers with their communication facts

damianvtran:
# PR Description

## Summary of Changes

A `hub steer` from the parent rendered **twice** in the subagent transcript view: once as the human-facing fact row (`Parent · redirected\n<body>`) and once as a raw user-message block showing the model-facing internal envelope:

```
<parent-message>
This changes your instructions. Apply it from now on, and drop work it makes pointless.

<body>
</parent-message>
```

The XML is model-facing and must never be shown to a person.

### Root cause

`SubagentComms.steer` journaled the human-facing communication fact under `message.id`, then called `record.child.steer(str(message.details["text"]))` — which builds a **fresh** `Message` with a new id. The persisted steer row is a plain `role=user` Message whose id therefore never equals the fact's `communication_id`, so the view's id-based dedup could not match them and both rows rendered. (`send`/`ask` are unaffected: they persist the `CustomMessage` itself, whose id equals the fact's id.)

### The fix

1. **Correlation** — route the steer through the identity-preserving `steer_message` seam with a caller-built `Message.user(..., id=message.id)`, so the persisted row carries the same id as the journaled fact. The TUI view and mobile projection already correlate by id and render the fact instead of the envelope.
2. **Defense-in-depth for legacy transcripts** — a new shared helper `extract_parent_message_body` is the single inverse of the envelope builder (`_format_to_child`). Both human-facing surfaces use it so the XML can never reach a person even when correlation fails:
   - **TUI** (`fold_transcript_entries`): an envelope user row is skipped when its fact is present (by id, then by body text for pre-fix transcripts whose ids never match); with no fact in the loaded window it renders the extracted body as a `Parent · redirected` row instead of the XML.
   - **Mobile** (`fold_messages_to_entries`, `fold_history`, `absorb_user_event`): the phone sees only LLM-visible messages (the fact is a custom row that never replays), so body extraction is its only path — the durable folds and the live delivery path all render `parent_message` with the extracted body, never the envelope.

The helper lives in `harness/comms.py` beside its builder so the two cannot drift apart.

### Reproduced the leak, then confirmed the fix, through the real render path

Drove the real `OperatorApp` + `SubagentView` over a transcript seeded with exactly what `comms.steer` writes (the fact + a plain user row holding the envelope with a **fresh** id), and captured rendered rows before and after.

**Before** (envelope leaks beside the fact):

```
'Parent · redirected\nFocus on retries'
[...]

PR #506, opened 2026-08-31
fix(tui): load one history chunk per scroll-to-top, not a cascade

damianvtran:
# PR Description

## Summary of Changes

Scrolling to the top of the transcript loaded history chunks in a continuous loop: "I scroll to the top and then it just starts loading chunks immediately one after another without requiring me to scroll up to the top again on the new height" — on both the parent transcript and the subagent view.

Both page-back triggers tested the scroll offset as a **LEVEL** (`at the top ⇒ load`) instead of an **EDGE** (`arrived at the top ⇒ load once`):

1. **Subagent view** (`subagent_view.py`) — `_scroll_changed` fired `_maybe_load_history()` on every `scroll_y <= 1` watch firing, with no latch and no programmatic-scroll guard. The anchor restore that `insert_blocks` performs after a prepend re-fired the watch at the top of the new height, and the settle frames after it kept firing; each firing requested another page. A sustained wheel drag walked the entire transcript (measured pre-fix: 5 disk reads, 320/320 rows loaded, `_history_exhausted` from ~150 notches of upward scroll).
2. **Parent view** (`app.py` + `transcript.py`) — `_check_resume_page` was level-triggered on `scroll_offset.y <= RESUME_PAGE_TRIGGER_ROWS`, gated only by `_resume_paging`. The mount's `release_gate` (`on_settled`) ran **outside** the programmatic-scroll guard in `restore_anchor`, so the gate re-opened while the reader sat parked inside the trigger zone and the next watch firing mounted another page (measured pre-fix: 9 mounts from one wheel drag on a 600-message session).

### Fix

Edge-triggered latch on both paths — consumed by one page load, re-armed **only by a user gesture** arriving after that page has landed:

- **Subagent view**: `_history_at_top` latch. `_scroll_changed` consumes it on the crossing into the top rows (keeping the `_initial_tail_pending` first-glance gate and the error-park). Re-arm lives in a new `_note_history_gesture`, installed via `set_on_user_scroll` on the body — the one signal the mount can never synthesize, because the insert's own restore scroll never passes through an input handler. Re-arm is suppressed while `_history_loading` is set, so the momentum tail of the same wheel cannot buy a second page.
- **Parent view**: `_resume_in_zone` latch. `_check_resume_page` mounts only when in-zone **and** armed, consuming the latch before mounting. `_transcript_scrolled` (the `on_user_scroll` hook) is the only re-arm, likewise suppressed while `_resume_paging` holds. Home-pressed-while-at-top still loads one page per press (the existing pinned contract) — a real keypress is a real gesture.
- **`transcript.py`**: `on_settled` now runs **inside** `restore_anchor`'s `programmatic_scroll()` guard, so the caller's gate never opens on a frame where the insert's own scroll could be reported as a reader's. Docstring updated to match.

Preserved unchanged: one-page-per-animated-gesture (M1/U1 single-flight), the subagent error-park until an explicit Home, and the repeated-`scroll_home`+`note_user_scroll` walk used by `test_scrolling_up_a_long_resume_reveals_older_rows_in_order`.

## Related Issue(s)

None.

## Impact

- Scroll-paging behaviour only; layout, colour, and chrome are unchanged (no visual delta — no design round needed).
- Version bumped `0.44.11` → `0.44.12`.

### Pre-fix reproduction (failures on `origin/main` @ 7b973190)

New latch tests, run against the unmodified tree (source changes stashed):

```
env -u NO_COLOR TERM=xterm-256color .venv/bin/python -m pytest tests/unit/tui/test_page_back_latch.py -q
# 3 failed in 7.57s
# FAILED ...::test_subagent_latch_ignores_watch_firings_while_parked_at_top
#   - AttributeError: 'SubagentView' object has no attribute '_note_history_gesture'
# FAILED ...::test_subagent_latch_does_not_rearm_while_a_page_is_in_flight
#   - AttributeError: 'SubagentView' object has no attribute '_note_history_gesture'
# FAILED ...::test_resume_latch_ignores_watch_firings_while_parked_in_zone
#   - AssertionError: assert 0 == (522 - 60)   # level trigger mounted nothing; head emptied by cascade
```

Measured cascade magnitudes on the pre-fix tree (instrumented probes, real wheel events):

| scenario | pre-fix | post-fix |
|---|---|---|
| subagent: 150-notch upward drag (320-row transcript) | 5 reads, 320/320 ids, exhausted | 3 reads, walk requires genuine re-arrival at top |
| subagent: wheel drag stopped at first mount, 200 idle frames | 1 page then idle* | 1 page then idle |
| parent: one wheel drag stopped at first mount, then idle | 9 mounts from one gesture (200-notch drag) | 1 mount, stable over 200 idle frames |

*the pure idle-park case happened to hold pre-fix on the parent; the failing shapes are the in-flight-notch and settle-frame ones, which the latch tests pin directly.

### Post-fix

New tests (all fail pre-fix, pass post-fix):

- `tests/unit/tui/test_page_back_latch.py` (new file) — pins the state machine directly and deterministically: watch firings while parked at the top load nothing; a gesture re-arms; the next crossing loads exactly one; in-flight gestures don't re-arm. Covers both the subagent and parent latches.
- `test_subagent_view.py::test_history_arriving_at_the_top_loads_one_page_then_stops` — real wheel events through the real watcher: one page per arrival, one disk read per gesture episode (counted on `read_transcript_page`, since a deduped request is invisible in `_history_ids`), then a deliberate away-and-back costs exactly one more.
- `test_resume_render.py::test_arriving_at_the_top_mounts_one_page_then_stops` — real wheel events with a momentum tail: exactly one page (`pending == 522 - RESUME_PAGE_MESSAGES`), stable over 120 idle frames, then a deliberate down-and-Home mounts exactly one more.

```
  tests/unit/tui/test_page_back_latch.py \
  tests/unit/tui/test_subagent_view.py -k history \
  tests/unit/tui/test_resume_render.py -q
# 3 passed / 12 passed / 13 passed
```

Existing contracts stay green: `test_an_animated_page_up_mounts_exactly_one_page`, `test_an_animated_home_mounts_one_page_and_lands_at_the_top`, `test_scrolling_up_a_long_resume_reveals_older_rows_in_order`, `test_durable_history_opens_at_tail_and_pages_to_the_start`, `test_history_home_key_lands_on_newly_loaded_page`, `test_history_prepend_preserves_anchor_and_home_dedupes_requests`.

### Gates (whole tree, as CI)

```
.venv/bin/python -m flake8 .                          # clean
uvx --from black==26.1.0 black --check .              # 687 files unchanged
uvx isort==5.13.2 --check .                           # clean
.venv/bin/python -m pyright --pythonpath .venv/bin/python .   # 0 errors, 0 warnings
```

### Whole unit suite

```
# run 1: 2 failed, 9874 passed, 15 skipped in 355.55s
# run 2: 1 failed, 9875 passed, 15 skipped in 318.05s
```

The failures are both in `tests/unit/tui/test_steering_approval.py` (`test_typing_a_steer_at_a_live_prompt_never_answers_it`, `test_a_held_answer_key_resolves_cleanly_on_every_second_key`). They are load-order flake, not this change: they pass 3/3 in isolation on this branch, and the same two tests fail intermittently on the unmodified tree (verified by stashing this change and re-running the file). This diff touches only scroll-paging state; it holds no approval/steer path. Recorded here as `deferred — pre-existing flake in test_steering_approval under full-suite load, unrelated to scroll paging; not addressed in this PR`.

## Checklist

- [x] All new and existing tests pass (modulo the pre-existing steering flake noted above).

### DIFF
--- a/local_operator/tui/app.py
+++ b/local_operator/tui/app.py
@@ -3743,7 +3743,7 @@ class OperatorApp(App[None]):
             transcript.follow_tail()
         return appended
 
-    def _transcript_scrolled(self, *_args: Any) -> None:
+    def _transcript_scrolled(self, *_args: Any, continuous: bool = False) -> None:
         """Mount the next older page when the reader reaches the top.
 
         Fired from the transcript's ``scroll_y`` WATCH (every offset the
@@ -3754,16 +3754,30 @@ class OperatorApp(App[None]):
         enough; the guards inside :meth:`_check_resume_page` are what keep
         the many firings of one animated gesture to ONE page.
 
-        This hook is also the latch's only re-arm (see `_resume_in_zone`).
-        Every INPUT path — wheel, key, scrollbar drag, arrow affordance —
-        announces itself here through ``note_user_scroll`` before it moves
-        anything, and the mount's own displacement never passes through it,
-        which is exactly the discrimination an offset-based re-arm cannot
-        make: the prepend pushes the reader out of the zone with no gesture
-        involved, and re-arming on THAT let a wheel still in motion walk the
-        whole head. A gesture pressed while already parked at the top
-        re-arms too — that is the deliberate "next page please" press, and
-        one press is one page because the check consumes the latch again.
+        This hook is also the latch's only re-arm (see `_resume_in_zone`),
+        and the RULE is deliberate: a gesture re-arms only when it can
+        actually have moved the reader OUT of the trigger zone, or when it
+        is a discrete act. Every input path announces itself here before it
+        moves anything, and the mount's own displacement never passes
+        through it — but a wheel notch arriving while the viewport is
+        already clamped at the top moves NOTHING, and re-arming on it let a
+        held wheel mount a page per notch while the reader sat at y=0 (the
+        "held scroll-up at the top" half of the reported loop). A discrete
+        act at the top is different: a keypress, an affordance click, or a
+        caller announcing a gesture by hand is the deliberate "next page
+        please", and one act is one page because the check consumes the
+        latch again.
+
+        Consequence for a SUSTAINED drag, stated plainly because it is the
+        behaviour and not an accident: a long wheel drag CAN mount several
+        pages. Each mount displaces the reader a page-height down (the
+        insert goes above them), so a wheel that keeps running travels that
+        distance back up and genuinely re-arrives at the top — one page per
+        real arrival, which is the contract. What this rule removes is the
+        page that no travel paid for: notches clamped at the top, settle
+        frames, and the mount's own restore. That is the half of the
+        reported loop the operator actually saw — chunks loading "without
+        requiring me to scroll up to the top again on the new height".
         """
         if not self._resume_pending_head:
             return
@@ -3771,13 +3785,12 @@ class OperatorApp(App[None]):
             # A page this same gesture requested is still mounting. The
             # notches still arriving from that wheel land inside the trigger
             # rows before the first page settles, and re-arming on them made
-            # one drag mount a second page the moment the gate re-opened.
-            # One gesture episode is one page; the next waits for a gesture
+            # one drag mount a second page the moment the gate re-opened —
+            # a page no travel paid for. The next page waits for a gesture
             # that arrives after this one's page has landed.
             return
-        # Re-arm unconditionally: a gesture is a gesture wherever it starts,
-        # and the settle/mount machinery can never fire this hook.
-        self._resume_in_zone = True
+        if not continuous or self._transcript_view().scroll_offset.y > RESUME_PAGE_TRIGGER_ROWS:
+            self._resume_in_zone = True
         # SINGLE-FLIGHT: exactly one deferred check may be pending at a time.
         # The watch fires once per animation frame and each firing used to
         # schedule its own check, so one animated gesture queued hundreds;
--- a/local_operator/tui/app.py
+++ b/local_operator/tui/app.py
@@ -2122,6 +2122,24 @@ class OperatorApp(App[None]):
         #: is already scheduled and `_transcript_scrolled` must not queue a
         #: second. See that method for why unbounded requeues were a cascade.
         self._resume_check_pending = False
+        #: EDGE-TRIGGERED top-zone latch, the second half of the page-per-
+        #: arrival contract (the first half is `_resume_paging`'s one page
+        #: per GESTURE). ``True`` means armed: the next at-rest moment inside
+        #: the trigger rows mounts exactly ONE page, and the latch stays
+        #: consumed until a scroll GESTURE arrives after that page has landed
+        #: (`_transcript_scrolled` is the only re-arm). Without it the check
+        #: was LEVEL-triggered: after a page prepended, the anchor restore
+        #: parked the reader back inside the trigger rows, the gate opened,
+        #: and the next watch firing — a settle frame, or one more wheel
+        #: notch of a gesture that was still running — mounted another page,
+        #: and another: the reported "it loads chunks one after another
+        #: without me scrolling up again". A GESTURE as the re-arm is what
+        #: closes the hole an offset-based rule cannot: the prepend itself
+        #: displaces the reader out of the zone with no gesture involved, and
+        #: every input path (wheel, key, scrollbar, arrow affordance)
+        #: announces itself through ``note_user_scroll`` before it moves
+        #: anything — the one signal the mount can never synthesize.
+        self._resume_in_zone = True
         #: What the CURRENT turn has already been billed for, per model call, by
         #: `on_context_usage_reported`. `on_turn_ended` prices the same turn as a
         #: whole and is the authoritative figure, so it adds only the difference
@@ -3463,6 +3481,9 @@ class OperatorApp(App[None]):
         # A check queued for the OLD conversation must not fire against the
         # new one; a fresh head has no gesture in flight worth answering.
         self._resume_check_pending = False
+        # Re-armed with the rest of the resume state: a new conversation's
+        # first arrival at the top owes a page (see `_resume_in_zone`).
+        self._resume_in_zone = True
         self._project_settled_rows(history, bound=RESUME_RENDER_MESSAGES)
 
     def _project_settled_rows(self, history: list[Any], *, bound: int | None = None) -> bool:
@@ -3730,16 +3751,33 @@ class OperatorApp(App[None]):
         gesture that moves NOTHING — Home while already at the top changes no
         offset, so the watch alone would miss the one key that most clearly
         means "show me the start"). Both are needed, and neither alone is
-        enough; the guard inside :meth:`_check_resume_page` is what keeps the
-        many firings of one animated gesture to ONE page.
+        enough; the guards inside :meth:`_check_resume_page` are what keep
+        the many firings of one animated gesture to ONE page.
 
-        The offset is never read here — this fires mid-animation and at
-        gesture start, where the offset is pre-gesture or mid-flight. The
-        question is handed to :meth:`_check_resume_page` after the refresh,
-        which waits for the viewport to be at rest before answering.
+        This hook is also the latch's only re-arm (see `_resume_in_zone`).
+        Every INPUT path — wheel, key, scrollbar drag, arrow affordance —
+        announces itself here through ``note_user_scroll`` before it moves
+        anything, and the mount's own displacement never passes through it,
+        which is exactly the discrimination an offset-based re-arm cannot
+        make: the prepend pushes the reader out of the zone with no gesture
+        involved, and re-arming on THAT let a wheel still in motion walk the
+        whole head. A gesture pressed while already parked at the top
+        re-arms too — that is the deliberate "next page please" press, and
+        one press is one page because the check consumes the latch again.
         """
         if not self._resume_pending_head:
             return
+        if self._resume_paging:
+            # A page this same gesture requested is still mounting. The
+            # notches still arriving from that wheel land inside the trigger
+            # rows before the first page settles, and re-arming on them made
+            # one drag mount a second page the moment the gate re-opened.
+            # One gesture episode is one page; the next waits for a gesture
+            # that arrives after this one's page has landed.
+            return
+        # Re-arm unconditionally: a gesture is a gesture wherever it starts,
+        # and the settle/mount machinery can never fire this hook.
+        self._resume_in_zone = True
         # SINGLE-FLIGHT: exactly one deferred check may be pending at a time.
         # The watch fires once per animation frame and each firing used to
         # schedule its own check, so one animated gesture queued hundreds;
@@ -3773,6 +3811,16 @@ class OperatorApp(App[None]):
         :data:`RESUME_PAGE_TRIGGER_ROWS` fires slightly before the hard top so
         the rows are already there when the reader arrives, rather than
         appearing under a viewport that had stopped.
+
+        EDGE, not level: the mount below fires only on the latch
+        (`_resume_in_zone`), which is armed by a user gesture arriving after
+        the previous page landed and consumed by this mount. A level test
+        here mounted on every settle frame — the anchor restore after a page
+        lands the reader back INSIDE the zone by construction, which is the
+        whole point of preserving the anchor — and the next watch firing
+        mounted another page. Home pressed while already parked at the top
+        still loads one page: that gesture re-arms the latch (it is a real
+        input) and the check consumes it again — one press, one page.
         """
         if not self._resume_pending_head:
             return
@@ -3789,7 +3837,13 @@ class OperatorApp(App[None]):
         ):
             self._transcript_scrolled()
             return
-        if transcript.scroll_offset.y <= RESUME_PAGE_TRIGGER_ROWS:
+        if transcript.scroll_offset.y <= RESUME_PAGE_TRIGGER_ROWS and self._resume_in_zone:
+            # Consume FIRST, mount second: the mount's settle re-fires the
+            # watch, and an unconsumed latch would answer it with another
+            # page. `self._resume_paging` also holds across the mount, but
+            # the latch is the contract that survives a settle whose gate
+            # released while the reader stayed parked in the zone.
+            self._resume_in_zone = False
             self._mount_older_resume_page()
 
     def _mount_older_resume_page(self) -> None:
--- a/local_operator/tui/widgets/subagent_view.py
+++ b/local_operator/tui/widgets/subagent_view.py
@@ -2143,6 +2143,10 @@ class SubagentView(Vertical):
             # prepend-only fast path would leave stale blocks mounted beside
             # the replacement even though the model is already canonical.
             prepend=not initial and not page.reconciled and bool(added_rows),
+            # Only the prepend path schedules the settle callback; carrying the
+            # generation lets that callback refuse to act on a page that has
+            # since been retargeted away (see `_finish_history_mount`).
+            generation=generation,
         )
         if initial:
             self._settle_initial_landing()
@@ -2190,7 +2194,11 @@ class SubagentView(Vertical):
         ]
 
     def _reconcile_current_body(
-        self, *, anchor: float | None = None, prepend: bool = False
+        self,
+        *,
+        anchor: float | None = None,
+        prepend: bool = False,
+        generation: int | None = None,
     ) -> None:
         entries = _mark_consecutive_notices(self._chronological_entries())
         tail = self._tail_entry(self._status == "gone", "")
@@ -2248,7 +2256,25 @@ class SubagentView(Vertical):
                     prefix,
                     new_blocks,
                     anchor_offset=anchor,
-                    on_settled=self._finish_history_mount,
+                    # Bound to the generation that scheduled it: a settle
+                    # callback can land after the page has been RETARGETED to
+                    # another job, and an unguarded clear would take down the
+                    # NEW job's `_history_loading` mid-read (every sibling
+                    # completion path guards the same way).
+                    # `generation` is threaded down from `_apply_history_page`
+                    # (None on every non-prepend caller, which never schedule
+                    # this callback). Late-binding is deliberate: the settle
+                    # can land after a retarget, and the guard inside
+                    # `_finish_history_mount` must compare against the
+                    # generation that SCHEDULED the insert, not whatever the
+                    # page is showing by then.
+                    on_settled=(
+                        lambda: (
+                            self._finish_history_mount(generation)
+                            if generation is not None
+                            else None
+                        )
+                    ),
                 )
                 # Re-raised for the duration of the insert: see the comment
                 # above. Cleared again by `_finish_history_mount`.
@@ -2271,7 +2297,7 @@ class SubagentView(Vertical):
         self._pending = entries
         self._sync_body(entries, self._pending_head)
 
-    def _finish_history_mount(self) -> None:
+    def _finish_history_mount(self, generation: int) -> None:
         """The prepend path's settle callback: the page is fully mounted.
 
         `_apply_history_page` clears `_history_loading` synchronously, which
@@ -2291,7 +2317,16 @@ class SubagentView(Vertical):
         in flight, so a reader (and a test asserting the rendered page) saw a
         walk that never finished. Painting here makes the settled text the
         last word regardless of repaint order.
+
+        Generation-guarded like every other history completion path: this is
+        a deferred callback, so it can land after `show()` has retargeted the
+        page to a DIFFERENT job (`_reset_history` bumped the generation and
+        started that job's own initial read). Clearing unguarded would take
+        the new job's `_history_loading` down mid-read, leaving its hint on
+        "loading earlier…" forever and its latch un-suppressed.
         """
+        if generation != self._history_generation:
+            return
         self._history_loading = False
         self._paint_history_state()
 
--- a/local_operator/tui/widgets/subagent_view.py
+++ b/local_operator/tui/widgets/subagent_view.py
@@ -1921,7 +1921,7 @@ class SubagentView(Vertical):
         if self.is_mounted:
             self._paint_chrome()
 
-    def _note_history_gesture(self, *_args: Any) -> None:
+    def _note_history_gesture(self, *_args: Any, continuous: bool = False) -> None:
         """A reader moved the body: re-arm the page-back latch.
 
         The re-arm half of the edge trigger (``_history_at_top`` is the
@@ -1930,15 +1930,32 @@ class SubagentView(Vertical):
         from the insert's own restore scroll, so a page mounting cannot
         re-arm the latch its own displacement would immediately consume.
 
-        Suppressed while a page this gesture already requested is in flight:
-        the notches still arriving from the same wheel land inside the top
-        rows before the first page has settled, and re-arming on them made
-        one drag issue a second (deduped, but real) request. One gesture
-        episode is one page; the NEXT page waits for a gesture that arrives
-        after this one's page has landed.
+        Suppressed in two cases, both of which are a gesture that cannot
+        have moved the reader OFF the top rows:
+
+        * while a page this same gesture requested is in flight — the
+          notches still arriving from that wheel land inside the top rows
+          before the first page has settled, and re-arming on them made one
+          drag issue a second (deduped, but real) request;
+        * a CONTINUOUS gesture (wheel/scrollbar) arriving while the body is
+          already clamped at the top. Such a notch moves NOTHING — the
+          offset is pinned at 0 — so it is not evidence the reader left and
+          came back, and re-arming on it let a held wheel load a page per
+          notch while the reader sat still. A discrete act (key, affordance
+          click, or a caller announcing a gesture by hand) at the top IS the
+          deliberate "next page please", so it re-arms.
+
+        A sustained drag CAN load several pages, and that is correct: each
+        prepend displaces the reader a page-height down, so a wheel that
+        keeps running travels that distance back up and genuinely re-arrives
+        at the top — one page per real arrival. What the two suppressions
+        remove is the page that no travel paid for.
         """
-        if not self._history_loading:
-            self._history_at_top = True
+        if self._history_loading:
+            return
+        if continuous and self._body.scroll_y <= 1:
+            return
+        self._history_at_top = True
 
     def _scroll_changed(self, *_args: Any) -> None:
         self._arm_arrows()
@@ -2090,6 +2107,10 @@ class SubagentView(Vertical):
     ) -> None:
         if generation != self._history_generation:
             return
+        # Cleared here for every reader EXCEPT the latch's in-flight
+        # suppression: the PREPEND path below re-raises it until the insert's
+        # own settle (`_finish_history_mount`), because that suppression must
+        # span the mount, not just the disk read.
         self._history_loading = False
         self._history_error = False
         rows = list(page.entries)
@@ -2212,7 +2233,26 @@ class SubagentView(Vertical):
                 # because the note used to occupy body index 0 while being
                 # excluded from `_entries` (review round 1, N1 / round 2, M3);
                 # pinning it out of the column removes that skew entirely.
-                self._body.insert_blocks(prefix, new_blocks, anchor_offset=anchor)
+                #
+                # `on_settled` re-opens the page-back latch's in-flight
+                # suppression (``_note_history_gesture``): until the gaps
+                # settle AND the anchor restore's own scroll has landed, the
+                # notches still arriving from the wheel that requested this
+                # page are part of the same gesture, and re-arming on them
+                # bought a second page. Holding `_history_loading` to HERE
+                # (rather than clearing it in `_apply_history_page`, before
+                # the insert even mounted) makes the window match
+                # `_resume_paging`'s on the parent view, which is held by the
+                # same callback.
+                self._body.insert_blocks(
+                    prefix,
+                    new_blocks,
+                    anchor_offset=anchor,
+                    on_settled=self._finish_history_mount,
+                )
+                # Re-raised for the duration of the insert: see the comment
+                # above. Cleared again by `_finish_history_mount`.
+                self._history_loading = True
                 self._blocks[prefix:prefix] = new_blocks
                 self._entries[prefix:prefix] = new_entries
                 self._pending = entries
@@ -2224,6 +2264,19 @@ class SubagentView(Vertical):
                 return
         self._pending = entries
         self._sync_body(entries, self._pending_head)
+
+    def _finish_history_mount(self) -> None:
+        """The prepend path's settle callback: the page is fully mounted.
+
+        `_apply_history_page` clears `_history_loading` synchronously, which
+        is correct for every reader of the flag EXCEPT the page-back latch's
+        in-flight suppression — that one must span the insert's own settle
+        (gap settlement plus the anchor restore), or a wheel still in motion
+        re-arms mid-mount. This callback is the earliest moment that is true,
+        and it is the same moment the parent view's `_resume_paging` gate
+        opens.
+        """
+        self._history_loading = False
 
     def _tail_entry(self, gone: bool, progress: str) -> SubagentEntry | None:
         """The row that TERMINATES the page, when the page needs one.
--- a/local_operator/tui/widgets/subagent_view.py
+++ b/local_operator/tui/widgets/subagent_view.py
@@ -2252,6 +2252,12 @@ class SubagentView(Vertical):
                 )
                 # Re-raised for the duration of the insert: see the comment
                 # above. Cleared again by `_finish_history_mount`.
+                # NOTE: this MUST NOT be visible to `_history_state_text`
+                # before the reconcile below has run — `show()` repaints the
+                # hint from `_history_loading` and a True here paints
+                # "loading earlier…" over the just-settled "transcript
+                # start". `_finish_history_mount` re-paints on clear, so the
+                # hint always ends at the settled text.
                 self._history_loading = True
                 self._blocks[prefix:prefix] = new_blocks
                 self._entries[prefix:prefix] = new_entries
@@ -2269,14 +2275,25 @@ class SubagentView(Vertical):
         """The prepend path's settle callback: the page is fully mounted.
 
         `_apply_history_page` clears `_history_loading` synchronously, which
-        is correct for every reader of the flag EXCEPT the page-back latch's
-        in-flight suppression — that one must span the insert's own settle
-        (gap settlement plus the anchor restore), or a wheel still in motion
-        re-arms mid-mount. This callback is the earliest moment that is true,
-        and it is the same moment the parent view's `_resume_paging` gate
-        opens.
+        is correct for every reader of the flag EXCEPT two: the page-back
+        latch's in-flight suppression — that one must span the insert's own
+        settle (gap settlement plus the anchor restore), or a wheel still in
+        motion re-arms mid-mount — and `_history_state_text`, which renders
+        "loading earlier…" for as long as the flag is set. The prepend path
+        therefore re-raises the flag after `_apply_history_page`'s clear, and
+        this callback is where it comes down for good, at the same moment the
+        parent view's `_resume_paging` gate opens.
+
+        The repaint matters as much as the clear: the reconcile that follows
+        a prepend drives `show()`, which repaints the hint from the flag and
+        painted "loading earlier…" OVER the settled "transcript start" text —
+        the state was correct while the visible chrome said a read was still
+        in flight, so a reader (and a test asserting the rendered page) saw a
+        walk that never finished. Painting here makes the settled text the
+        last word regardless of repaint order.
         """
         self._history_loading = False
+        self._paint_history_state()
 
     def _tail_entry(self, gone: bool, progress: str) -> SubagentEntry | None:
         """The row that TERMINATES the page, when the page needs one.
--- a/local_operator/tui/widgets/subagent_view.py
+++ b/local_operator/tui/widgets/subagent_view.py
@@ -1670,6 +1670,24 @@ class SubagentView(Vertical):
         self._history_error = False
         self._history_unavailable = True
         self._initial_tail_pending = False
+        #: EDGE-TRIGGERED page-back latch, not a level test. ``_scroll_changed``
+        #: fires for every offset the body passes through — including the
+        #: anchor restore ``insert_blocks`` performs after a page prepends,
+        #: and every settle frame after it. A level test there
+        #: (``scroll_y <= 1`` ⇒ load) mounted a page on each of those
+        #: firings while the reader sat parked at the top, and a wheel still
+        #: in motion crossed straight back into the top rows after the
+        #: prepend displaced it — a sustained drag walked the entire
+        #: transcript (the reported "chunks load one after another at the
+        #: top"). Consumed by one load; re-armed only by a USER GESTURE
+        #: arriving after that load landed (``_note_history_gesture``), the
+        #: one signal the mount cannot synthesize — every input path (wheel,
+        #: key, arrow affordance) announces itself through the body's
+        #: user-scroll hook before it moves anything, and the insert's own
+        #: restore scroll never passes through that hook. ``True`` here
+        #: means armed: the page opens with the reader at the tail, and the
+        #: FIRST arrival at the top must load.
+        self._history_at_top = True
         #: One-shot: the first laid-out body may land its sticky tail on a
         #: wrap fragment (a continuation line with no glyph). Snapped once
         #: onto a row head so the first glance is a statement; later
@@ -1718,6 +1736,12 @@ class SubagentView(Vertical):
         # happened to notice. Watching the reactive is the only signal that
         # fires exactly when the answer changes.
         self.watch(self._body, "scroll_y", self._scroll_changed, init=False)
+        # The page-back latch re-arms HERE, on the gesture signal, and nowhere
+        # else (see ``_history_at_top``). The body's user-scroll hook fires for
+        # every input path before it moves anything; the insert's own anchor
+        # restore never passes through it, which is exactly the discrimination
+        # an offset-based re-arm cannot make.
+        self._body.set_on_user_scroll(self._note_history_gesture)
         self._maybe_load_history(initial=True)
 
     def on_unmount(self) -> None:
@@ -1870,6 +1894,9 @@ class SubagentView(Vertical):
         self._history_exhausted = False
         self._history_error = False
         self._history_unavailable = not bool(directory)
+        # Re-armed on retarget (see ``_history_at_top``): a new job is a new
+        # reader at the tail whose first scroll to the top owes a page.
+        self._history_at_top = True
         # ``show`` runs immediately after ``screen.mount`` but before Textual
         # has mounted this child. ``on_mount`` owns the first request; later
         # retargets are already mounted and can start immediately.
@@ -1894,14 +1921,56 @@ class SubagentView(Vertical):
         if self.is_mounted:
             self._paint_chrome()
 
+    def _note_history_gesture(self, *_args: Any) -> None:
+        """A reader moved the body: re-arm the page-back latch.
+
+        The re-arm half of the edge trigger (``_history_at_top`` is the
+        consume half, in ``_scroll_changed``). Fired from the body's
+        user-scroll hook — wheel, keys, the arrow affordances — and never
+        from the insert's own restore scroll, so a page mounting cannot
+        re-arm the latch its own displacement would immediately consume.
+
+        Suppressed while a page this gesture already requested is in flight:
+        the notches still arriving from the same wheel land inside the top
+        rows before the first page has settled, and re-arming on them made
+        one drag issue a second (deduped, but real) request. One gesture
+        episode is one page; the NEXT page waits for a gesture that arrives
+        after this one's page has landed.
+        """
+        if not self._history_loading:
+            self._history_at_top = True
+
     def _scroll_changed(self, *_args: Any) -> None:
         self._arm_arrows()
-        if not self._initial_tail_pending and self._body.scroll_y <= 1:
+        # EDGE, not level: consume the latch only on the crossing INTO the
+        # top rows. The restore after a prepend lands the reader at the top
+        # of the new height by design, and that second firing is what made a
+        # level trigger cascade (see ``_history_at_top``). Staying at the top
+        # — the settle frames, the error repaint, a resize — fires this watch
+        # again and again and must load NOTHING.
+        at_top = self._body.scroll_y <= 1
+        if (
+            at_top
+            and self._history_at_top
+            # The same first-glance gate the level trigger carried: while
+            # the page still owes its reader a first look at the tail, a
+            # scroll that lands at the top is the OPENING layout settling,
+            # not a reader asking for history (the initial load itself can
+            # move the offset to the tail and back).
+            and not self._initial_tail_pending
+        ):
+            self._history_at_top = False
             # Edge arrival may discover another page, but an error parks here
             # until a NEW Home command explicitly accepts another disk read.
             # Otherwise the reconciliation repaint retriggers this watcher and
             # turns a persistent filesystem failure into a hot retry loop.
             self._maybe_load_history()
+        # No re-arm here. The offset leaving the top rows is NOT evidence of a
+        # reader leaving: the prepend itself displaces the reader down by a
+        # whole page, and an offset-based re-arm let a wheel still in motion
+        # walk the entire transcript. Only a GESTURE re-arms
+        # (``_note_history_gesture``), and a gesture that begins below the
+        # top and crosses into it is the one arrival that owes a page.
 
     def action_home(self) -> None:
         """Reach the earliest loaded row and request one missing page.
--- a/local_operator/tui/widgets/transcript.py
+++ b/local_operator/tui/widgets/transcript.py
@@ -2206,7 +2206,10 @@ class TranscriptView(ScrollableContainer):
         #: watch on the offset would miss the one gesture that most clearly
         #: means "show me the start". The hook is the same shape as
         #: ``on_clear``: optional, app-owned, never required for the widget.
-        self._on_user_scroll: Callable[[], None] | None = None
+        #: It receives ``discrete`` — True for a keystroke, False for a
+        #: continuous gesture — because the page-back latch re-arms on a
+        #: deliberate keypress at the top but not on a clamped wheel notch.
+        self._on_user_scroll: Callable[..., None] | None = None
         # The ledger's shared name column, recomputed lazily. Cached because it
         # is read once per card per repaint and only changes when the set of tool
         # names on screen does.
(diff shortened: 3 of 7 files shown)


damianvtran:
### Agent review — round 1
Reviewer: reviewer (lopdev team, independent — did not author this change) — anthropic/claude-opus-5
Scope: 7b973190..210f21e5

Verified by building a genuine pre-fix tree (`7b973190` versions of the three
touched files, own source resolution confirmed via `local_operator.tui.app.__file__`)
and running the new tests against it, plus targeted probes. Full TUI suite on head:
**4112 passed, 12 skipped**. flake8 / black 26.1.0 / isort 5.13.2 / pyright: all clean.

The core fix is correct and the state machines are sound. I could not find a path that
leaves either latch stuck armed or stuck consumed; the reset paths (`_reset_history`
retarget, fresh `/resume`, `/clear`) all re-arm, and the `on_settled` ordering change
has exactly one production caller (`app.py:3890`), so nothing else regresses. One MAJOR
below is about the TESTS, not the shipped behaviour.

---

**M1 — the two new END-TO-END tests pass against the pre-fix, still-cascading code, so they do not pin the reported bug**
`tests/unit/tui/test_resume_render.py:334` (`test_arriving_at_the_top_mounts_one_page_then_stops`)
and `tests/unit/tui/test_subagent_view.py` (`test_history_arriving_at_the_top_loads_one_page_then_stops`).

Measured, not inferred. Against the pre-fix tree, run by node id:

```
$ .venv/bin/python -m pytest \
    tests/unit/tui/test_resume_render.py::test_arriving_at_the_top_mounts_one_page_then_stops \
    tests/unit/tui/test_subagent_view.py::test_history_arriving_at_the_top_loads_one_page_then_stops \
    -q -p no:randomly -n0
2 passed in 13.31s          # repeated 3x, stable
```

Both docstrings claim the pre-fix behaviour they exclude ("Pre-fix that tail mounted a
second page from the one gesture"). It did not, on this path. The reason is the loop's
own early exit: the wheel loop `break`s the moment `len(app._resume_pending_head) < 522`
— i.e. as soon as the *first* page mounts — so the drag stops at the exact frame the
cascade would begin, and the 3-notch "momentum tail" is delivered after the gesture has
already ended. The subsequent `for _ in range(120): await pilot.pause()` is then a
quiet viewport, which pre-fix code also survives: a plain `scroll_to(y=0)` arrival
mounts exactly 1.0 pages pre-fix too (probed directly).

What DOES reproduce the cascade pre-fix is a *sustained* drag — notches still arriving
while the page mounts:

```
200 wheel notches, no early break:
  PRE-FIX : 522 -> 0    pending head, 8.7 pages   <- the reported cascade
  POST-FIX: 522 -> 162  pending head, 6.0 pages   (stable across 3 runs)
```

Why it matters: these two tests are the only end-to-end guards on the user-facing
contract, and both would stay green if the latch were deleted tomorrow. The regression
value currently rests entirely on `test_page_back_latch.py`, which *does* discriminate
correctly (3/3 fail pre-fix, the parent one draining the head to 0 — the exact cascade
signature). Remedy: drop the `if ... break` early exit and keep delivering notches for a
fixed count so the drag overlaps the mount, or assert on a sustained-drag page count
rather than a single-arrival one. Cheap to confirm: the amended test must fail against
`git show 7b973190:local_operator/tui/app.py`.

Note this is a test-strength finding, not a claim the fix is wrong — the fix does reduce
8.7 pages to 6.0 on the sustained path, and collapses the parked-at-top cascade entirely.

**M2 — the sustained-drag path still mounts 6 pages from one continuous gesture**
`local_operator/tui/app.py:3742-3752`.

Post-fix, 200 notches with no pause still walks 6 pages of head. This is consistent with
the documented design (`_transcript_scrolled` re-arms unconditionally on every gesture,
and each notch of a continuous drag genuinely *is* a fresh gesture), so it may well be
intended — the PR body's contract is "one page per gesture", and 200 notches are 200
gestures. But the docstring at `app.py:3736-3738` says "One gesture episode is one page",
and a real trackpad drag is one *episode* delivering many notches. Worth an explicit
call: either the phrase "gesture episode" should become "one page per notch, suppressed
while a page is in flight" (accurate to the code), or the re-arm needs an episode notion
(a time/notch quiescence window). I lean toward the docstring fix — 6 pages for a
deliberate sustained drag is defensible reader intent, and the reported bug was loading
*without* further input, which is fixed.

**N1 — `_history_loading` is already `False` when the in-flight window the test simulates would occur**
`local_operator/tui/widgets/subagent_view.py:1855` and `:2008`.

`_apply_history_page` clears `_history_loading` at line 2008, before the anchor restore
settles — so the suppression in `_note_history_gesture` covers the disk read but not the
restore. The corresponding test
(`test_subagent_latch_does_not_rearm_while_a_page_is_in_flight`) sets
`view._history_loading = True` **by hand** to create the window it tests. I probed the
real window and found no double load (reads stayed at 1, latch behaved), because the
restore lands within one refresh here — so this is not a defect today, only a note that
the guard is narrower than its docstring implies and the test does not exercise the real
timing. Contrast with the parent, where `_resume_paging` is correctly held until
`on_settled`; that asymmetry is worth a comment.

**N2 — short-content case (`max_scroll_y == 0`) can never page back — PRE-EXISTING, not a regression**
`local_operator/tui/widgets/subagent_view.py:1858`.

With a tall viewport where the loaded page fits, `max_scroll_y == 0`, so a wheel-up
changes no offset, `_scroll_changed` never fires, and no page loads (`reads=0` after 10
notches). Confirmed identical on the pre-fix tree, so this diff neither causes nor worsens
it — flagging only because the review focus asked. `action_home` still works, so there is
an escape hatch. Recording as a deferral is fine; no ticket per team policy.

---

**Verified clean, explicitly:**
- Reset paths: `_reset_history` (`subagent_view.py:1814`), fresh resume (`app.py:3454`),
  `/clear` (drops `_resume_pending_head`, so the latch state is moot). No stuck-armed or
  stuck-consumed path found.
- `on_settled` ordering: one production caller (`app.py:3890`); `prepend_blocks` /
  `anchor_offset` path in `subagent_view.py:2130` passes no `on_settled`, and the
  early-return branch at `transcript.py:2276-2279` still fires it outside the guard —
  correct, since no scroll happens there.
- Pinned contracts hold on head: Home-at-top one page per press, subagent error-park,
  `test_scrolling_up_a_long_resume_reveals_older_rows_in_order`, M1/U1 single-flight.
  Non-gesture offset changes (4 terminal resizes while parked in the trigger zone)
  mounted **0** extra pages — the level-trigger hazard is genuinely closed.
- Comments/docstrings match the code and follow the repo's WHY-heavy house style, with
  the two exceptions noted in M2 and N1.
- The `test_steering_approval.py` flake deferral in the PR body: I found no evidence
  linking it to this diff and did not re-litigate it.

**Verdict: not clean — 2 MAJOR (M1, M2).**
M1 is the one I would hold on: the end-to-end regression guards do not currently fail
against the bug they describe, which is cheap to fix and is the difference between a
pinned contract and a comment. M2 may be resolved by amending the docstring rather than
changing behaviour. Neither is a correctness blocker in shipped code — the fix itself is
sound and I found no state-machine hole — so if the implementer prefers, M2 can be
answered as a docstring correction and M1 as a test amendment, and round 2 should be
quick.

damianvtran:

Head: `93e1514fd489ecda0ac840301b8209d6da875763` (reviewed head was `210f21e5`). Scope answered: `7b973190..93e1514f`.

**M1 — e2e tests did not discriminate the pre-fix tree: fixed** (`93e1514f`)

Both tests were rewritten as CONTINUED drags that do not stop at the first mount, exactly as the finding prescribed, with a per-mount **travel audit** as the assertion: every page mount/load is checked against the highest scroll offset observed since the previous one — a mount with no intervening travel out of the trigger zone is the level-trigger cascade, whatever gesture was in flight.

- `test_resume_render.py::test_arriving_at_the_top_mounts_one_page_then_stops` — trackpad rhythm (5 notches/frame × 25 frames), then a hold against the clamped top (5 notches/frame × 30 frames, offset pinned), then idle. **Pre-fix: fails deterministically, 3/3 runs — `8 of 9 pages mounted with no travel away from the trigger zone`.** Post-fix: passes 3/3 (`mounts with no travel == 0`, idle-stable, Home still mounts exactly one more). On this view the violation is deterministic by geometry: a 40-row viewport against a ~120-row page means the reader reaches the clamped top long before the head is exhausted.
- `test_subagent_view.py::test_history_arriving_at_the_top_loads_one_page_then_stops` — same shape (burst × 40 frames + clamped hold × 60 frames), audit on `read_transcript_page` calls. **Pre-fix: fails intermittently (1 of 3 runs showed `1 of 3 pages loaded with no travel`); post-fix passes 5/5.** The intermittency is geometry, not a weak assertion: a history page is 100 rows against a ~28-row viewport, so a continued drag displaces the reader a full page between arrivals and the no-travel violation only appears when frame timing lands the clamp before the displacement. The **deterministic** pin for this view is `test_page_back_latch.py` (3/3 pre-fix failures, re-verified on the tar-copy tree), and the e2e test's docstring now says exactly that rather than overclaiming.

Verification method per the finding: pre-fix tree built with `git archive 7b973190 | tar -x` into a scratch dir (no `git checkout`/`git stash` — the shared worktree index was never touched).

**M2 — docstring contradicted behaviour (6 pages from a 200-notch drag): fixed, by tightening behaviour AND the comments** (`93e1514f`)

Re-measuring the sustained drag surfaced a real hole rather than a wording problem: the unconditional gesture re-arm let a **clamped** wheel notch — one that arrives while the offset is pinned at the top and therefore moves nothing — re-arm the latch, so a held wheel mounted a page per notch at y=0 (9 mounts from one drag on this branch, 8 with no travel at all; pre-fix 9). Fixed: `note_user_scroll` now carries `continuous` (wheel/scrollbar events) vs the default discrete act (keys, affordance clicks, explicit callers), and the latch re-arms on a continuous gesture **only when the offset is actually outside the trigger zone**. Post-fix the same 200-notch drag mounts 1 page (5/5 runs).

On the deliberate-behaviour half of the finding: a sustained drag with genuine travel **does** still page per re-arrival, and that is the contract — each mount displaces the reader a page-height down, so a wheel that keeps running travels that distance back up and genuinely re-arrives; the operator's complaint was loading *without* further scrolling, which is exactly what the clamped-notch rule removes. Both `_transcript_scrolled`'s docstring and `_note_history_gesture`'s now state this plainly ("A sustained drag CAN mount several pages, and that is correct… What this rule removes is the page that no travel paid for") instead of the incorrect "one gesture episode is one page".

**N1 — `_history_loading` window narrower than documented / asymmetric with `_resume_paging`: fixed** (`93e1514f`)

`_apply_history_page` cleared the flag synchronously, before the insert even mounted, so the suppression the test simulated by hand (`_history_loading = True`) did not exist on the real path. The prepend path now re-raises the flag and clears it from `insert_blocks(on_settled=self._finish_history_mount)` — the same callback moment `_resume_paging` uses on the parent view, covering gap settlement plus the anchor restore. Verified not to stick: a full Home-walk to `_history_exhausted` ends with `_history_loading=False`, latch armed, and the latch unit test that hand-set the flag now exercises the real window.

**N2 — `max_scroll_y == 0` can never page back: deferred — pre-existing on the base tree** (confirmed by the reviewer; not addressed here)

`_scroll_changed`'s top test can never fire when the content fits the viewport, so a short child transcript cannot page back by scroll. Confirmed present on `7b973190` and untouched by this change. Deferred rather than fixed here: it is a distinct missing-feature case (a wheel-event path for `max_scroll_y == 0`), out of scope for a cascade fix, and fixing it safely means deciding what "the top" means when there is nothing to scroll — a design question, not a one-liner. No issue filed per the team's standing rule.

**Re-verification on the new head**

- Affected files: `tests/unit/tui/test_resume_render.py + test_page_back_latch.py + test_subagent_view.py` → 128 passed.
- Pre-fix tree (tar-copy) with the same tests: latch file 3/3 failed; parent e2e 3/3 failed; subagent e2e 1/3 failed (geometry, documented above).
- Gates over the whole tree: flake8 clean · black 26.1.0 687 files unchanged · isort 5.13.2 clean · pyright 0 errors, 0 warnings.
- Whole unit suite: 9874 passed, 2 failed — both in `test_steering_approval.py`, the flake already recorded in round 1. Re-confirmed on the **pre-fix tar-copy tree** where the same file fails identically (2 failed / 112 passed, twice), so it is unaffected by this change. `deferred — pre-existing load-order flake in test_steering_approval.py, reproduced on the unmodified base tree`.

damianvtran:
### Agent review remediation — round 1 (addendum: 3.13 CI failure)

Head advanced to `f4a0a52519392f2c2d13686456cd1f28023d1a9d` (was `93e1514f`).

The 3.13 leg of run 33437314252 failed `test_durable_history_opens_at_tail_and_pages_to_the_start` with the hint stuck on `loading earlier…`. **Real regression from the N1 fix, reproduced locally (7/10 pytest runs) and fixed** in `f4a0a525`:

- **Cause.** Holding `_history_loading` through the insert's settle meant the reconcile that follows a prepend drove `show()`, which repaints the hint from the flag — and the flag was still `True`, so `loading earlier…` painted *over* the already-settled `transcript start` text. The state machine was correct (`_history_exhausted`, cursor at the start) while the visible chrome claimed a read was in flight. Intermittent because it depends on whether `show()` lands between the state paint and the settle callback.
- **Fix.** `_finish_history_mount` now repaints the history state when it clears the flag, so the settled text is the last word regardless of repaint order. A comment at the re-raise site records why the flag must not be visible to `_history_state_text` mid-reconcile.
- **Verified.** The failing test 8/8 green; the three affected files 128/128; whole unit suite **9876 passed, 0 failed**; flake8/black/isort/pyright all clean.

The 3.12 leg's two failures are the pre-existing flakes already recorded in round 1's remediation (`test_steering_approval` ×1, `test_redundant_roster_events_skip_the_sidecar_write` ×1 — the latter reproduced by the manager on the pre-fix tar tree, 1-in-8; both pass 5/5 and 19/19 here on the new head).

damianvtran:
### Agent review — round 2
Reviewer: reviewer (lopdev team, independent — did not author this change) — anthropic/claude-opus-5
Scope: 7b973190..f4a0a525

Verified the three remediation claims by re-running the discrimination experiments myself on a `git archive` tar-copy of the pre-fix tree (`/tmp/prefix-tree`, production code at 7b973190 + the new tests), never touching a shared index. Static gates all clean on head: flake8 0, `black==26.1.0 --check` 687 files unchanged, `isort==5.13.2 --check` clean, pyright **0 errors, 0 warnings**.

#### Round-1 findings — remediation verdicts

**M1 (test shape: both e2e tests stopped at the first mount, so they passed pre-fix) — FIXED, claims reproduced exactly.**
- Parent (`test_resume_render.py:334`): pre-fix **3/3 failures**, and the message is verbatim the claim — `AssertionError: 8 of 9 pages mounted with no travel away from the trigger zone`. The continued-drag + clamped-hold shape plus the per-mount travel audit is a genuine discriminator.
- `test_page_back_latch.py`: pre-fix **3/3 failures** on the state machine. This is the deterministic pin, as claimed.
- Subagent (`test_subagent_view.py:1181`): the "intermittent pre-fix" admission is honest, but it understates the problem — see **M3**, it is also intermittent POST-fix.

**M2 (unconditional re-arm let a clamped wheel notch re-arm the latch) — FIXED, and this was a real hole.** Measured directly, 200 clamped notches held against a pinned top:
- pre-fix tree: **7 pages mounted**
- head: **0 additional pages** (the arrival page having already been consumed)

Classification audited at every call site (`grep note_user_scroll(`): wheel up/down, `_on_scroll_to`, `_on_scroll_up`, `_on_scroll_down` → `continuous=True`; `action_scroll_up/down`, `action_page_up/down`, `action_scroll_home`, and the subagent arrow affordance (`subagent_view.py:2817`) → discrete. That mapping is correct. Home-at-top verified independently: **1 page per press, and exactly one on a second press** — the zero-delta discrete case still works.

**N1 (`_history_loading` held through the insert's settle) — FIXED, including the regression it introduced.** `test_durable_history_opens_at_tail_and_pages_to_the_start` is **8/8 green** on head (claim was 2/8 failing before f4a0a525). I traced every exit from the flag: `_finish_history_unavailable` (1972), `_finish_history_error` (1984), `_apply_history_page` (2029), `_reset_history` (1808), and the no-blocks branch — `insert_blocks` returns early on empty `additions`, but the flag is only re-raised *after* that call returns on a path where `count > 0`, so no empty-page path re-raises it. **No stuck-True path found**, with the one caveat in M4 below.

**N2 (`max_scroll_y == 0` can't page back) — DEFERRED, accepted.** Pre-existing, correctly recorded in the thread.

#### New findings

**M3 — BLOCKER-adjacent, calling it MAJOR: the new subagent test is flaky on its own post-fix head, and it fails with the cascade message.**
`tests/unit/tui/test_subagent_view.py:1181`. Measured on head, isolated, `-o addopts=""`: **11 pass / 1 fail in 12 runs**; in whole-file runs, **2 failures in 6**. The failure is not a timeout or a teardown error — it is the assertion this test exists to make:
```
AssertionError: 1 of 3 pages loaded with no travel away from the top rows — the level-trigger cascade
```
The base tree is **6/6 clean** on this file, so this is introduced, not inherited. The docstring's scope note admits pre-fix intermittency but claims the test "holds the same contract end-to-end"; it does not — roughly 1 run in 6 reports a no-travel load *against the fixed code*. Either the audit's `peak` sampling is wrong (it samples in `_scroll_changed` and in the drag loop, so a page whose travel happened entirely between two sampled frames records `peak <= 1` spuriously), or the production latch genuinely admits a no-travel read under a particular frame alignment. Those two have very different severities and the test cannot currently tell them apart. This will fail CI intermittently on an unrelated PR, and worse, it will be *dismissed* as flake the one time it is real. Remedy: make the audit sample the peak inside the production path it is auditing (record the offset at the moment the latch is consumed, not from the test's frame loop), or drop the travel assertion from this test and let `test_page_back_latch.py` — which is deterministic 3/3 — carry the contract, as the docstring already half-concedes.

**M4 — MAJOR: `_finish_history_mount` is the only history completion callback with no generation guard.**
`local_operator/tui/widgets/subagent_view.py:2189`, wired at 2166. Every sibling takes and checks `generation`: `_apply_history_page(generation, ...)` (2019), `_finish_history_error(generation)` (1978), `_finish_history_unavailable(generation)` (1969). `_finish_history_mount()` takes nothing. It is scheduled through `insert_blocks`' `call_after_refresh` chain (two refreshes deep), so it can land after a `show()`/`_reset_history` retarget has bumped `_history_generation` and started a *new* job's load. Demonstrated: with the generation bumped and a new load in flight (`_history_loading = True`), calling the stale `_finish_history_mount()` sets `_history_loading = False` — clearing the new job's in-flight flag and re-opening the page-back latch's suppression for a read that has not landed. It also repaints the new job's chrome from the old job's settle. The window is narrow (retarget during a prepend settle) but the guard costs one parameter and one comparison, and its absence is inconsistent with the file's own established contract. Remedy: `def _finish_history_mount(self, generation: int) -> None:` with the standard early return, bound via `partial`/closure at the 2166 call site.

**M5 — MAJOR: the new subagent test leaks a module-global monkeypatch for the rest of the pytest worker.**
`tests/unit/tui/test_subagent_view.py:1239` assigns `subagent_view.read_transcript_page = counting_read`, but the `finally` at 1321 restores **only** `SubagentView._scroll_changed`. `read_transcript_page` is never restored. Verified: running that test and then any later test in the same process shows `subagent_view.read_transcript_page` is still `test_history_arriving_at_the_top_loads_one_page_then_stops.<locals>.counting_read`. Every subsequent history read in that xdist worker runs through the closure and mutates its `reads` dict, and if the test failed before reaching its `finally`, later tests inherit it too. This is a plausible contributor to the order-dependent failure I saw under xdist (`test_clicking_the_arrows_scrolls_the_transcript_both_ways`, `assert 42 > 44`, which does not reproduce serially or in isolation and is 6/6 clean on the base tree). Note the sibling `test_page_back_latch.py` does this correctly with `monkeypatch.setattr` — this file should too. Remedy: add the `monkeypatch` fixture and use `monkeypatch.setattr(subagent_view, "read_transcript_page", counting_read)`, or restore it in the existing `finally`.

**N1 — NIT: the `on_user_scroll` docstring names a parameter that does not exist.**
`local_operator/tui/widgets/transcript.py:2099-2101` says the hook "receives ``discrete`` — True for a keystroke, False for a continuous gesture". The parameter is `continuous`, with the opposite polarity (`note_user_scroll(*, continuous: bool = False)`, 2692). The house style is heavy WHY comments and this one actively misleads: a reader implementing a hook against that sentence writes the branch backwards. Remedy: rename `discrete` → `continuous` and invert the two clauses.

**N2 — NIT: `Callable[..., None]` gives up the type check that would have caught N1.**
`transcript.py:2099` and `set_on_user_scroll` (2156) widen to `Callable[..., None]`, so pyright can no longer verify that a hook accepts `continuous`. `Callable[[], None]` was precise before. Both current consumers (`_transcript_scrolled`, `_note_history_gesture`) already have the shape `(*_args: Any, continuous: bool = False) -> None`, so a `Protocol` with a keyword-only `continuous: bool` would type them exactly and restore the check. Minor, but the widening is what let the docstring drift unnoticed.

#### Also verified (no finding)
- The `on_settled`-inside-`programmatic_scroll` reorder in `insert_blocks` (`transcript.py:2292-2305`) is correct, and the comment's reasoning holds: releasing inside the guard makes the open-gate frame not exist. Both production callers (`release_gate`, `_finish_history_mount`) now run inside it. The `anchor_block is None` early-return path (2283) still calls `on_settled()` — correctly, outside the guard, since no scroll happens there.
- Reset paths re-arm both latches: `_resume_in_zone = True` on session swap (`app.py:3454`), `_history_at_top = True` on retarget (`subagent_view.py:1814`) and at init (1605). No stuck-consumed path found.
- A resize while parked at the top loads **nothing** on either view (probed directly) — the comments at `subagent_view.py:1880` and `app.py:3798` are accurate on that point.
- `_transcript_scrolled`'s bare no-kwarg call from `watch_scroll_y` (`transcript.py:2749`) classifies every watch firing as discrete, which *looks* like it should re-arm on settle frames — but `_resume_paging` (checked first, `app.py:3756`) covers the mount window and the probe confirms 0 extra pages. Correct, though it rests on that ordering; worth a sentence if anyone touches it.
- CI run 33445367525's `test (3.13)` failure is **unrelated to this diff**: `tests/unit/session/test_resume_subagents_todos.py::test_descendant_accounting_survives_process_resume`, a file this PR does not touch, and 3/3 green on the pre-fix base tree locally — i.e. an independent flake, not a regression from this change.

#### Verdict
**Not clean — 3 MAJOR (M3, M4, M5).** M1, M2 and N1 from round 1 are each genuinely and verifiably closed, and the core fix is sound: the clamped-notch cascade is measurably dead (7 pages → 0). But M3 and M5 are defects *in the evidence this PR rests on* — a regression test that fails 1-in-6 on the fixed code with its own cascade message, and a leaked global patch that contaminates the rest of the worker — and M4 is a real inconsistency in the production state machine. M3 and M5 in particular should not merge as-is, because a flaky guard plus a leaking patch is how the next real cascade gets waved through as "that test again". Round 3 requested.

damianvtran:

Head: `4704e68c434907f90ffff737cd5ea65a475c8011` (reviewed head was `f4a0a525`). Scope answered: `7b973190..4704e68c`.

**M4 — `_finish_history_mount` had no generation guard: fixed** (`4704e68c`)

The generation is now captured where the page is applied and threaded through `_reconcile_current_body` into the settle callback; `_finish_history_mount(generation)` refuses to act when the page has retargeted, exactly like `_apply_history_page`, `_finish_history_error`, and `_finish_history_unavailable`. Late binding is deliberate — the guard must compare against the generation that *scheduled* the insert. New test `test_finish_history_mount_refuses_a_stale_generation` pins it: retarget mid-read, land the stale settle, assert the new job's `_history_loading` survives and its hint stays honest. (Fails unguarded, passes guarded.)

**M3 — the flake was a REAL cascade in the fix, now closed** (`4704e68c`)

Traced end-to-end: a legitimate page mount settled at y=1, then `watch_scroll_y` — which invoked the user-scroll hook with **no kwargs**, so `continuous` defaulted False — re-armed the latch on the clamped 1→0 offset step as if it were a discrete act; the next watch consumed it and mounted a page with zero travel. The watch now passes `continuous=True` (it reports offset motion — travel evidence — never a gesture). Two measurement gaps in the audit were also closed so the test stops crying wolf: the verdict is judged after the drag over completed intervals (a mount's displacement lands a frame after the read that caused it — logging showed y=101 observed between two reads the old audit flagged), and the hold loop now samples peak.

Stability on the fixed head: `test_subagent_view.py` whole-file **15/15 consecutive green** (was ~50% before the watch-hook fix); the regression test isolated 10/10. Pre-fix baseline for contrast: the same file fails ~5/10 whole-file runs on the base tree (the discriminating test doing its job).

**M5 — leaked `read_transcript_page` monkeypatch: fixed** (`4704e68c`)

Both patches (`read_transcript_page`, `_scroll_changed`) are restored in the `finally`, with a comment stating why the module-attribute swap is worker-global state.

**N1 — docstring named a nonexistent `discrete` parameter: fixed** (`4704e68c`)

The docstring now names `continuous` with the correct polarity.

**N2 — `Callable[..., None]` erased the hook's parameter contract: fixed** (`4704e68c`)

`set_on_user_scroll` now takes a `UserScrollHook` Protocol carrying the `continuous` keyword; verified pyright rejects a wrong-signature hook (`discrete=...` → `reportArgumentType`), which is exactly the check that would have caught N1.

**Re-verification on `4704e68c`**

- Affected files (`test_page_back_latch.py`, `test_resume_render.py`, `test_subagent_view.py`): **129/129 passed**.
- Whole-file stability: `test_subagent_view.py` ×15 consecutive → 15/15 green.
- Gates over the whole tree: flake8 clean · black 26.1.0 687 files unchanged · isort 5.13.2 clean · pyright 0 errors, 0 warnings.
- Whole unit suite: **9876 passed, 1 failed** — `test_steering_approval.py::test_typing_a_steer_at_a_live_prompt_never_answers_it`. Reproduced on the **untouched base tree `7b973190`** (clean `git archive` copy): 2 of 6 file-level runs fail there too, so it is pre-existing load-order flake, not this change. `deferred — pre-existing flake in test_steering_approval.py, present on the unmodified base tree`.
- CI dispatched on `4704e68c` (run 33464716937); the PR `pull_request` sync trigger has not been firing for pushes to this branch, so it was started via `workflow_dispatch`.

damianvtran:
### Agent review remediation — round 2 (addendum: CI's 3.13 failure on 4704e68c)

`test_page_back_latch.py::test_finish_history_mount_refuses_a_stale_generation` failed on CI's 3.13 leg (run 33464716937) — reproduced locally at ~1-in-8 under file load and traced. It was a **test race, not a product bug**: the fake comms stub answered every `session_dir_of` with the FIRST job's directory, so the app's own periodic refresh of the retargeted `job-other` arrived carrying the wrong directory, took `show()`'s changed-directory branch, and fired a third `_reset_history` mid-test — legitimately clearing the `_history_loading` the test had just set. The guard under test behaved correctly throughout (the stale generation was refused; the clear came from the new job's own reset).

Fixed in `31aa9ca6`: the stub answers per job, so the refresh is a no-op. 30/30 whole-file runs green; gates clean. CI dispatched on `31aa9ca6` (run 33468265280).

The other two 3.13 failures on that run are unrelated to this PR and reproduce independently of it: `test_s1_boot_paints…` is the responsiveness ceiling flake recently recalibrated (#496), and `test_the_sonnet_5_introductory_price_has_not_expired` is a pricing-guard expiry on main's registry — both untouched by this diff.

damianvtran:
### Rebase onto main @ 7ac34cdf — head now 40b066cf

Main moved under us (#505/#509/#511/#512), which left the PR conflicting and put two pre-existing failures in CI's path (the Sonnet-5 pricing guard expired on 2026-09-01 — already fixed on main by #512 — and the steering flake). The branch is rebased onto current main with the version conflict resolved to **0.44.20** (main is at 0.44.19; patch bump for this fix). Commit list unchanged apart from the rebase (5 commits, 763d3bc0 → 40b066cf).

Verification on 40b066cf: affected files 129/129 · flake8 clean · black 708 files unchanged · isort clean · pyright 0 errors. CI dispatched (run 33470418915).

damianvtran:
### Agent review — round 3
Reviewer: reviewer (lopdev team, independent — did not author this change) — zai/glm-5.3 (verification begun under anthropic/claude-opus-5)
Scope: 7ac34cdf..40b066cf

Freshness/closure round on the rebased head. Checkout verified at `40b066cf` on `dev-fix-lazy-load-loop`; base confirmed `origin/main` = `7ac34cdf`. All probes below ran in that worktree (probe files removed after; `git status` clean); the pre-fix/main comparison used a `git archive` tar-copy, never a checkout.

#### Round-2 findings — each closed, with independent verification

**M3 (residual cascade via `watch_scroll_y`'s no-kwarg hook call) — CLOSED, and the diagnosis was right.** `transcript.py:2872` now passes `continuous=True` from the watch, so the clamped 1→0 step of a wheel pinned at the top is travel-evidence, not a discrete act, and cannot re-arm. The reworked audit (verdict judged after the drag over completed intervals; the hold loop samples peak) closes both measurement gaps round 2 named. Stability re-measured by me, not taken from the claim: `test_history_arriving_at_the_top_loads_one_page_then_stops` **10/10 isolated** (`-o addopts=""`), `test_subagent_view.py` whole-file **6/6 consecutive, 112 passed each** — round 2 measured ~2 failures in 6 on the then-head.

**M4 (`_finish_history_mount` unguarded) — CLOSED.** Signature takes `generation` with the standard early return (`subagent_view.py:2243-2244`); threaded from `_apply_history_page` → `_reconcile_current_body(generation=...)` → the settle lambda. `test_finish_history_mount_refuses_a_stale_generation` **8/8 isolated** on this head (the per-job comms stub in `40b066cf` fixed the test's own third-`_reset_history` race, as the addendum described).

**M5 (leaked module-global monkeypatch) — CLOSED.** Both `read_transcript_page` and `_scroll_changed` are restored in the test's `finally`.

**N1 (docstring named a nonexistent `discrete`) — CLOSED.** The docstring now documents `continuous` with the correct polarity; no stale `discrete` parameter naming remains in any touched file.

**N2 (`Callable[..., None]` erased the contract) — CLOSED, and the check is real.** I probed it: `set_on_user_scroll(bad_hook)` where `bad_hook` is `() -> None` produces a pyright `reportArgumentType` ("Missing keyword parameter `continuous`") against `UserScrollHook`. The Protocol does what the remediation claims.

#### Review-focus items

**(b) `continuous=True` from the watch vs the pinned contracts — holds.** `test_an_animated_page_up_mounts_exactly_one_page`, `test_an_animated_home_mounts_one_page_and_lands_at_the_top`, `test_scrolling_up_a_long_resume_reveals_older_rows_in_order`, and `test_arriving_at_the_top_mounts_one_page_then_stops`: 4/4 on head. Home-at-top per-press is pinned inside the new tests and I probed a two-press sequence directly: 2 presses separated by frames mount exactly 2 pages. The discrete/continuous classification at every `note_user_scroll` call site is unchanged from round 2's audit and remains correct (keys/affordances discrete; wheel/scrollbar/watch continuous).

**(c) The settle lambda — correct, no `None` path fires it.** `generation` is a parameter of `_reconcile_current_body`, so each call captures its own value at schedule time; only `self` is late-bound, which is intended (the guard must compare the scheduling generation against the view's current one). All five `_reconcile_current_body` callers audited: the four non-prepend callers pass no `generation` and take the `_sync_body` path, which never reaches `insert_blocks`; the only prepend caller (`_apply_history_page:2055`) always passes its non-None `generation`. `count > 0` guarantees `additions` is non-empty, so `insert_blocks`' empty-early-return cannot skip `on_settled`, and both remaining exits (anchor-block-gone, restore-inside-guard) fire it. No stuck-`_history_loading` path found; `_reset_history` clears the flag on every retarget.

**(d) Rebase coherence — clean.** Version resolves to 0.44.20 against main's 0.44.19 (correct patch bump per policy). I specifically chased the `(review round 1, N1 / round 2, M3)` citation at `subagent_view.py:2157` because it does not match this PR's round numbering — it is **pre-existing main code from #482** (`a39900e3`), not a stale reference this PR introduced. Comments otherwise match the code they describe; CI run 33470418915 is all-8-green on exactly this head (verified via `gh run view`).

#### New findings

**M6 — MINOR: the requeue branch in `_check_resume_page` re-arms the latch through a bare discrete-classified call (`app.py:3832`).** The wait branch calls `self._transcript_scrolled()` with no kwargs, so `continuous` defaults False and the re-arm at 3773-3774 is unconditional. Demonstrated: reader parked at the top, latch consumed, page settled; setting `scroll_target_y != scroll_offset.y` (the "requested but not started" window) and calling `_check_resume_page()` re-arms the latch, and one page then mounts with **zero travel** once the target is cleared. Reachability caveat, stated honestly: in every production path into that branch travel is genuinely committed (animation in flight or target set), and the only force-stop of a committed scroll is the anchor restore, which runs while `_resume_paging` is held — whose early return at 3765 blocks the re-arm. I could not construct a production trigger that cancels committed travel after the arm. This is a latent inconsistency (a future caller can re-arm without travel through this branch), not a live defect; the honest fix is to pass an explicit classification at 3832 rather than inherit the discrete default.

**M7 — MINOR: `_check_resume_page`'s docstring promises a second gesture during the mount window is "answered the moment the gate re-opens rather than silently dropped" (`app.py:3798-3801`) — deterministically false, before and after this PR.** Same-frame double-Home probe: 2 presses inside the mount window mount **1.0 pages, 3/3 runs**, and the behaviour is **identical on the main tree** (same probe, `git archive` copy of `7ac34cdf`) — so this is a pre-existing docstring inaccuracy, not a regression, and this PR's `_resume_paging` early-return merely makes the drop structural rather than timing-dependent. Remedy is the sentence, not the code: describe the wait as re-asking until the gate opens, with a discrete gesture after the settle being what buys the next page.

**N3 — NIT: `UserScrollHook.__call__` accepts `*_args: Any`, so the Protocol checks `continuous` but not the positional shape.** A hook `(a, b, c) -> None` type-checks although the hook is only ever invoked with the keyword. `def __call__(self, *, continuous: bool = False) -> None` is exact and still accepts both current consumers (a `*args`-taking callable is a subtype). Cosmetic; the load-bearing check works.

#### Verification summary

- Affected files as instructed: **129 passed** (`test_page_back_latch.py` + `test_resume_render.py` + `test_subagent_view.py`).
- Gates over the whole tree, as CI runs them: flake8 **0**; `black==26.1.0 --check` **708 files unchanged**; `isort==5.13.2 --check` clean; pyright **0 errors, 0 warnings, 0 informations**.
- CI run 33470418915 on `40b066cf`: all 8 checks green, both Python legs, both tui-e2e OS legs.
- Probes: hook-signature pyright rejection; same-frame and frame-separated double-Home; forced no-travel re-arm; stale-generation stress 8/8; subagent whole-file 6/6; all probe files removed, worktree clean at `40b066cf`.

#### Verdict

**Clean.** No BLOCKER, no MAJOR. All five round-2 findings (M3, M4, M5, N1, N2) are genuinely closed on this head with evidence above; the rebase onto `7ac34cdf` is coherent; the pinned contracts hold. M6 and M7 are latent/pre-existing minors that do not regress anything against main, and N3 is cosmetic. This round is **TERMINAL**: the remaining minors and the nit are follow-ups, not grounds for another round.

damianvtran:
### UX review — round 1
Reviewer: ux-reviewer (lopdev team) — zai/glm-5.3
Scope: origin/main..40b066cf

Walked the real flow end to end on both surfaces, driving the actual `OperatorApp` through `run_test` pilots with real key events (`Key("home")`/`pageup` through the app's own handler path), real wheel events (`MouseScrollUp` in trackpad rhythm — bursts of notches per frame, held against the clamped top), the `↑` arrow affordance's real click path (`_scroll_body`), and real terminal resizes — at 120×40, 90×28, and a narrow 58×24. Compared against a base-tree archive of `origin/main`, where the same wheel drag mounted **8 pages in one gesture** (522 → 0 pending, `max_y` 126 → 1170) — the cascade this PR exists to remove.

**Parent `/resume` (600-message session, 120×40)**
- One wheel arrival at the top mounts exactly ONE page (60 messages: 522 → 462 pending), then holds. 80 idle frames afterwards: stable, nothing more loads.
- The reading position holds across the mount: the topmost visible block at arrival kept its viewport-relative offset (0 → 0) through the prepend's anchor restore.
- `Home` at the top: exactly one page per press (462 → 402 → 342 → 282, each −60), landing back at the top. Repeated `Home` walks to the start in bounded, predictable steps and the first row restates itself honestly as `start of conversation`.
- At the clamped top, the gesture map is coherent: a held wheel loads nothing (dead by design — a clamped notch is not travel), `Home` and `PageUp` each load exactly one page, and the leave-and-return recovery works — including a small flick (leave a few rows, return): one page each time.
- Resize while parked at the top mounts nothing.

**Subagent full-page view (320–2000-row jobs)**
- Same contract on the body: one arrival = one page (100 rows), stable when parked, `Home` = one page per press, and a sustained drag with genuine travel pages per re-arrival (200 → 600 across a 60-frame drag on the 600-row job — travel paid for each).
- The hint text stays truthful through the walk: `· read-only` → `· loading earlier…` during a slowed (300 ms) read (visible for the read's duration) → `· read-only` after a mid-page, and `· transcript start · read-only` once exhausted. After exhaustion, wheel and `Home` at the top are correctly inert.
- The `↑` arrow affordance behaves as a proper discrete act: nine page-clicks walk to the top of the loaded range, the tenth's arrival loads one page, and the walk continues — a mouse-only reader is never stranded.
- Short transcript (12 rows, fits the viewport): opens settled at the start (`· transcript start · read-only`), no loading flicker, `Home`/arrows correctly inert.
- Narrow 58×24: paging still works, hint still truthful, resize at the top loads nothing.
- Error path: a failed read parks on `· load failed · Home retry · read-only` with no retry storm, and `Home` genuinely retries and recovers. Worded for the user, recoverable.

**Feel.** The after-flow reads calm where the before-flow raced: the base tree consumed an entire 600-message history in one drag, which forced the reader to hunt back down for where they'd been; the new flow loads one page and stops, the anchor holds, and more history is one deliberate act away. It does not feel dead — sustained travel still pages, discrete acts still page, and every state says what it is doing.

**U1 — MINOR — parent: the first row's copy over-promises in the parked state.** After a page lands, the reader parks at the top of the new height with the first row reading `older messages above — scroll up to load`. Following that instruction literally (holding the wheel up against the clamped top) loads nothing, by design. The recovery gestures that do work — leave-and-return, `Home`, `PageUp` — are not named anywhere on the parent surface (`ctrl+home` is a hidden binding, `Home`/`PageUp` inherit from `ScrollableContainer` with no visible hint). A reader who takes the copy at its word experiences the surface as stuck rather than deliberate. Expected: copy that names the act, e.g. `older messages above — home or scroll for more`, or an equivalent affordance surfaced in the hint row. Not a blocker: the surface never wedges (leave-and-return always recovers), and the copy is pre-existing rather than introduced here.

**U2 — NIT — subagent: `Home`'s page-more contract is undiscoverable except through the error hint.** The `Home` binding is `show=False` (`Load earlier transcript`), and after the first page the only on-screen state text is `· read-only` — nothing says `Home` buys another page, whereas the error state helpfully says `· load failed · Home retry · read-only`. A reader who wants more history past the loaded top must either drag with travel or guess `Home`. Expected: the settled, more-pending state could carry `· home for earlier · read-only` alongside `loading earlier…`. Cosmetic; the wheel and arrow paths both work unaided.

**U3 — NIT — parent: a slow mount gives no in-flight feedback.** The parent's page mount is synchronous in-memory projection, so on normal sessions the new page appears within a frame or two — fine. But on a very large page the insert + layout can outlive a frame, and during that window the first row still reads `older messages above — scroll up to load` with no in-flight marker, unlike the subagent surface which shows `loading earlier…`. Cosmetic asymmetry between the two surfaces; observed only under artificial slowdown, never in the normal walk.

Verdict: **no BLOCKER, no MAJOR.** The cascade is gone, one page per arrival holds on both surfaces across wheel/key/affordance/resize inputs, the anchor holds, the hint stays truthful through every state including errors, and no dead end exists — every stall state has a working recovery. The round is **TERMINAL**; U1–U3 are follow-ups.

damianvtran:
### Design review — round 1
Reviewer: designer — anthropic/claude-opus-5
Scope: origin/main..40b066cf

Frames were captured by an earlier designer run (killed mid-round by a provider rate limit) from a base tree at `/tmp/lop-base` (origin/main) and the head tree at `/tmp/lop-lazy-load` (40b066cf); I rendered the remaining SVGs to PNG and viewed every frame myself rather than trusting the prior run's notes, and re-derived the geometry from the SVGs independently of `report.json`.

**Viewed:** parent P1 (tail), P2 (after one drag to top), P3 (after Home), P4 (exhausted) in both trees; subagent S1 (tail), S2 (after drag), S4 (exhausted) in both trees.

**What the pixels confirm.** The money shot reads exactly as the fix intends. base/P2 lands on `· start of conversation` above `turn 0000` — one drag detonated the whole backlog. head/P2 lands on `· older messages above — scroll up to load` above `turn 0154`, one page, viewport held. The scrollbar corroborates it without my needing the JSON: base/P2 shows a hairline thumb (601 blocks over max_scroll 1170), head/P2 a visibly taller thumb (139 blocks over 246). That is the user-legible signal that only one chunk mounted, and it is consistent in the pixels and the numbers.

The progression is honest. head/P3 steps 0154 → 0134 with the hint still present — one more page, not a cascade — and head/P4 converges on the identical `· start of conversation` terminal state as base/P4, so the change alters the path to the top without changing the destination. The notice never lies about what is above it in any frame I looked at.

No visual regression from the diff. Notice sits at the same row 0 origin (y=68.8, x=73.2) in both trees; both parent frames keep the identical 40-row grid with a 7.0 step and the same origin; virtual == actual (118×38 parent, 88×26 subagent) with `screen_scrollbar` false everywhere, so no overlay-scrollbar reflow. Row order, indentation, the `▌` turn markers, the bash tool-call bands and their right-aligned `✓` are unchanged. head/P1 correctly shows no notice at the tail.

Subagent surface is untouched, and I verified that rather than assuming it: S1/S2/S4 SVG text content is identical between base and head except for the footer cwd (`/private/tmp/lop-base` vs `/private/tmp/lop-lazy-load`), which is a capture-path artifact, not a rendering difference. Text-node counts match exactly (67 each). S2/S4 show durable 0 at top with the truthful `· transcript start · read-only` hint.

D1 — NIT — The new hint renders at `#b5afa2` (8.62:1 on the `#14110c` background) while the terminal `· start of conversation` renders at `#837c6d` (4.55:1). I read this as deliberate and correct rather than an inconsistency: the head-of-list state is actionable ("scroll up to load") and earns more ink than the inert end-of-history marker, and both clear WCAG AA for body text. Recorded only so the contrast delta is a documented choice rather than an accident; no change requested.

**Nothing the numbers claim is contradicted by the pixels.** `mounts_from_one_drag` 9 → 1 and pending 522 → 462 are both visible on screen as the turn number the drag lands on (0000 vs 0154) and as the scrollbar thumb size, and `idle_stable: true` matches P2 and P3 sitting at scroll_y 0 with no post-settle jump between the captured states.

Verdict: **TERMINAL** — no BLOCKER, no MAJOR, no MINOR. One NIT (D1), recorded as an accepted design choice, not a follow-up. The user-visible behaviour reads as intended on both surfaces and the diff introduces no layout, spacing, ordering, or chrome regression.

damianvtran:
### Agent review — round 4
Reviewer: reviewer (lopdev team, independent — did not author this change) — anthropic/claude-opus-5
Scope: 40b066cf..2bab973b (freshness re-verification against current head; full-diff scope origin/main..2bab973b)

**Clean. The delta is rebase + version bump only, and rounds 1–3 + design + UX carry forward to this head unchanged.** The claim verified as stated; I reproduced it rather than taking it.

**1. The PR's own content is byte-for-byte identical.** A raw `git diff 40b066cf..2bab973b` is non-empty, but only because the two commits sit on different bases — it shows main's `9fdb6f8a` (team picker) and `2a00cd9b` (test de-flaking) arriving, including their `docs/evidence/` assets, which `git log --diff-filter=A` confirms were introduced by `9fdb6f8a`, not by this branch. The correct comparison is the PR's own patch on each base:

- old: `git diff 7ac34cdf..40b066cf -- . ':!pyproject.toml'`
- new: `git diff 2a00cd9b..2bab973b -- . ':!pyproject.toml'`

Both are **59935 bytes**. Diffing them, every differing line is an `index <blob>..<blob>` header or an `@@` hunk coordinate — the mechanical residue of a rebase. Stripping those two line classes, the patches are **identical across all 1140 lines**, with zero `+`/`-`/context lines differing. No unexplained hunk exists.

**2. The version bump is the only intentional change**, and it is exactly one line: `pyproject.toml` `0.44.20` → `0.44.21`.

**3. `2bab973b~6` is `2a00cd9b`, which is the current `origin/main` tip** — the branch is rebased onto main proper, not an intermediate.

**4. No interaction with main's two new commits.** `9fdb6f8a` does share `local_operator/tui/app.py` with this PR, so I checked semantically rather than trusting the clean rebase: its `app.py` hunks land at lines ~142/1749/2644/4035/4958/5404/6782/15407, all disjoint from the lazy-load region (~2110–3850), and grepping its diff for `_transcript_scrolled`, `_resume_check_pending`, `_resume_pending_head`, `_finish_history_mount`, `scroll_y`, `UserScrollHook`, `continuous`, `latch`, `_history_generation`, `page_back`, `RESUME_PAGE` returns **nothing**. `2a00cd9b` touched `test_task_wait_jobs.py`, `test_app_pilot.py`, `test_steering_approval.py`, `test_todo_panel_phases.py` — **none** of this PR's three test files.

**5. It is the same logical change rounds 1–3 approved.** All four mechanisms are present on this head: the edge-triggered latch in `app.py:3746,3792`, the latch in `subagent_view.py:1839,1871`, the `continuous` `UserScrollHook` in `transcript.py:68-72,2812,2837,2872`, and the generation guard in `subagent_view.py:1801,1942,1970,1982,2023`. Scope against the new base is the expected 7 files, +863/−28.

**6. Verification run on this head (2bab973b):**
- `env -u NO_COLOR TERM=xterm-256color .venv/bin/python -m pytest tests/unit/tui/test_page_back_latch.py tests/unit/tui/test_resume_render.py tests/unit/tui/test_subagent_view.py -q` → **129 passed** in 71.64s, the expected count.
- `flake8 .` → clean · `black==26.1.0 --check .` → 708 files unchanged · `isort==5.13.2 --check .` → clean · `pyright` → **0 errors, 0 warnings, 0 informations**. All four run over the whole tree, as CI does.

**7. Version ordering.** Head is `0.44.21`; `origin/main` is `0.44.20`; PyPI's latest is `0.44.20` and `0.44.21` is not yet published. The bump is correct and unpublished, and patch is the right class for a scoped TUI fix.

**On CI's `test (3.12)` failure — not attributable to this PR, and not a finding.** Run 33507426906 reports `1 failed, 10169 passed` with the failure at `tests/unit/session/test_conversation_name_persistence.py::test_teardown_costs_one_budget_not_two` — `AssertionError: teardown took 0.80s, more than one budget`. That is a wall-clock timing assertion in a file this PR does not touch: `git diff origin/main..2bab973b -- tests/unit/session/` is **empty**, the file is identical to main, and the PR changes no `session/` code at all — it is confined to `tui/`. It last changed in `#269`, long before this branch. Locally on this exact head it passes **5/5** (2.7–4.9s each). This is a pre-existing shared-runner timing flake of precisely the class main's own `#513` was written to address, and it does not gate this change. Worth noting the assertion should wait on a condition rather than a 0.4s budget — but that is main's file, out of scope here, and recorded as `deferred — pre-existing flake in an untouched file, belongs to main not this PR`.

#### Verdict

**Clean.** No BLOCKER, no MAJOR, no MINOR, no NIT against the delta. The change under review did not move between `40b066cf` and `2bab973b`; only its base and its version string did. Rounds 1, 2 and 3 (agent), the design round and the UX round — all TERMINAL with no blocker/major — **stand against head `2bab973b`** and require no re-run. This round is **TERMINAL**, and satisfies the freshness requirement for merge.

damianvtran:
### Agent review — round 5
Reviewer: reviewer (independent agent-review subagent) — anthropic/claude-opus-5
Scope: 2bab973b..7cd9ece6 (freshness re-verification after rebase onto main 98831a8a; full-diff scope 98831a8a..7cd9ece6)

**Clean — no findings. Rounds 1–3 (agent), design, UX, and round 4 all carry forward unchanged.**

The delta since round 4's head `2bab973b` is exactly a rebase plus a version bump. Normalising both patches (stripping `index` blob hashes and `@@` hunk coordinates) and diffing them content-line by content-line yields a single difference across 1152 lines — the version bump line moving from `0.44.20 → 0.44.21` to `0.44.21 → 0.44.22`. The reviewed behaviour is byte-identical to what rounds 1–4 approved, so no re-review of the latch logic was warranted.

**The substantive question — interaction with main's #504 — is answered: genuinely disjoint.** Verified independently at both region and symbol level, in both shared files:

- **Region, product file.** #504's hunks in `subagent_view.py` all land at lines 54–667 (`fold_transcript_entries`, `_supersedes`, an import). Ours all land at 1673–2333 (the `SubagentView` lazy-load latch). No adjacency, let alone overlap.
- **Region, test file.** #504 appends one block at old line 1516; ours inserts at 1178 plus one import. Both survive in the merged file — `test_durable_communications_are_human_facing_and_correlated` at :1643 and `test_history_prepend_preserves_anchor_and_home_dedupes_requests` at :1134.
- **Symbol.** Grepping #504's changed lines for our symbols (`_history_at_top`, `_maybe_load_history`, `_scroll_changed`, `_finish_history_mount`, `_note_history_gesture`, `_history_loading`, `_reconcile_current_body`, `continuous`) returns nothing; grepping our changed lines for its symbols (`fold_transcript_entries`, `_supersedes`, `steer`, `communication`, `reprlib`, `correlat`) likewise returns nothing.
- **Coherence.** AST-walked both merged files for duplicated definitions and imports: 115 defs in the test file and 89 in the product file, **zero duplicates**, no shadowed names, no duplicated fixtures or helpers. The one repeated import the scan flagged (`FrontendSessionState`, `JobState`) is a pair of pre-existing function-local imports in two separate tests, present on main at `2a00cd9b` before either change — not introduced by the merge.

**Test count reconciled to the individual test, not just the total.** `test_page_back_latch.py test_resume_render.py test_subagent_view.py` → **140 passed** (was 129). The +11 is fully attributable to #504: it added 6 test functions to `test_subagent_view.py`, which parametrise to 11 collected cases, each confirmed present by name. All 6 of our tests also still collect — `test_subagent_latch_ignores_watch_firings_while_parked_at_top`, `test_subagent_latch_does_not_rearm_while_a_page_is_in_flight`, `test_resume_latch_ignores_watch_firings_while_parked_in_zone`, `test_finish_history_mount_refuses_a_stale_generation`, `test_arriving_at_the_top_mounts_one_page_then_stops`, `test_history_arriving_at_the_top_loads_one_page_then_stops`. Nothing of ours vanished in the rebase.

**Gates, whole tree, as CI runs them:** flake8 clean; `black 26.1.0 --check` — 710 files unchanged; `isort 5.13.2 --check-only` clean; `pyright --pythonpath .venv/bin/python .` — 0 errors, 0 warnings.

**Version:** `0.44.22`, ahead of both main (`0.44.21`) and PyPI (`0.44.21`). The dropped bump commit left no residue — the branch is not sitting on an already-released version.

Round 5 is TERMINAL and clean. No blocker, major, minor, or nit findings. Merge gate satisfied on head `7cd9ece6` pending CI green.

PR #499, opened 2026-08-31
test(tui): pin store maintenance to worker threads

damianvtran:
# PR Description

## Summary of Changes

Replaces the flaky timing assertion in `test_prepare_store_scans_do_not_stall_the_loop` with the deterministic invariant the test is intended to protect: all four whole-store maintenance callbacks execute off the asyncio event-loop thread.

The test now wraps the real `sweep_from_config`, `sweep_orphan_groups`, `backfill_session_origins`, and `backfill_session_titles` module attributes before the maintenance coroutine's delayed imports. Each wrapper records its thread ID and then invokes the real callback, preserving behavioral coverage. The test asserts each callback runs exactly once, every callback thread differs from the captured loop thread, an aged empty store is reaped, and title-scan sentinels are written.

No production code, shared timing thresholds, other responsiveness call sites, or package version changed. Recent test-only PRs #496 and #486 likewise did not bump the version.

### Why timing cannot guard this site

Healthy and regressed CPU measurements overlap at this loaded `_prepare` site: current main has failed at 205–206 ms against the 200 ms shared ceiling, while forcing maintenance work onto the loop does not produce a separable timing distribution. Raising or retuning the threshold therefore either flakes healthy code or misses the regression. Thread identity directly detects the failure mode: moving any maintenance callback out of `asyncio.to_thread` places it on the event-loop thread.

### Real path and effects

```
$ env -u NO_COLOR TERM=xterm-256color .venv/bin/python -m pytest \
    tests/unit/test_tui_responsiveness.py::test_store_maintenance_callbacks_run_off_the_event_loop_thread -n0 -q
1 passed in 1.92s
```

The test invokes all four real maintenance functions. It verifies the real retention sweep removes an aged empty session directory and the real title backfill writes sentinels for all 120 titleless sessions.

### Mutation proof

Locally replaced the production dispatch line `await asyncio.to_thread(work)` with `work()` in a copy, ran the new test, then restored production code. The replacement failed on the first synchronously dispatched callback:

```
AssertionError: retention ran on the event-loop thread
assert 8414060928 != 8414060928
1 failed in 3.98s
```

### Repeatability and responsiveness file

```
50/50 focused runs passed under four-way process contention; failures=0

$ ... pytest tests/unit/test_tui_responsiveness.py -n0 -q
19 passed, 2 warnings in 35.42s

$ ... pytest tests/unit/test_tui_responsiveness.py -q
19 passed, 6 warnings in 19.27s
```

### Whole-tree gates

```
[...]

PR #498, opened 2026-08-31
test(tui): restore the 500 ms loaded ceiling after s1-boot CI calibration

damianvtran:
# PR Description

## Summary of Changes

The site-specific ceiling fix (#496's second commit) restored a 200 ms loaded default on the premise that only the title-scan site had legitimate loop CPU and the boot/turn paths recorded ~0 ms. The merge run of that fix **disproved the premise**: `test_s1_boot` failed `test (3.13)` with

```
event loop consumed 321 ms of CPU without yielding (ceiling 200 ms)
```

### The full calibration dataset

Every run that has flaked a lower ceiling on a green tree:

| Site | Healthy loop CPU on CI | Run |
|---|---|---|
| title-scan (`_prepare`) | 206 ms | **#486** merge run (`aa8de03c`) |
| title-scan (`_prepare`) | 264 ms | **#495** merge run (`c74eb53e`) |
| s1-boot (Textual layout) | 321 ms | **#496** merge run (`7b973190`) |
| s2 (Textual layout) | 413 ms | **#495** merge run (`c74eb53e`) |

The observed healthy max is therefore **413 ms**, and 500 ms carries **~20% headroom** over it. That margin is thin — the same site has swung ~30% between runs — so the docstrings state the full dataset, name the margin explicitly, and say what to do if a green tree exceeds 500 ms: raise it again on evidence rather than assuming the probe is wrong.

(The historical 549 ms layout-pass figure is **wall-clock**, from the pre-#486 probe this CPU clock replaced. It remains useful as an upper bound on the class of work — 321 ms of CPU is in the expected range — but it is not a CPU precedent.)

### The fix

The loaded default goes back to **500 ms**, above the observed healthy CPU max of 413 ms, still well below the multi-second regressions (a 2 s scan, a 5 s pricing block). The title-scan site's explicit `cpu_ceiling_ms=500` becomes redundant and is removed.

**What this accepts, stated plainly:** s1-boot and s2 have no guard other than the loaded helper, so a 200-500 ms stall at those sites passes. The multi-second regressions are still caught, and the strict 50 ms bar on reconnect/connect (the parse regression this probe exists for) is untouched.

The title-scan test's **structural guard** (`threading_get_ident() not in seen_threads`) remains its primary protection, independent of any ceiling.

---

### Healthy

```
$ pytest tests/unit/test_tui_responsiveness.py -q -n0
19 passed, 2 warnings in 40.03s
```

### Both regression guards still fire

[...]

PR #507, opened 2026-08-31
docs(agents): add timing/flake guidance to AGENTS.md

damianvtran:
# PR Description

## Summary of Changes

Three PRs (#486, #496, #498) and eleven review rounds went into fixing one flaky timing test, and **none of what was learned was written down where the next person would find it**. `AGENTS.md` had no guidance on timing assertions at all — the word "flaky" appeared exactly once, in an unrelated sentence about database connections.

This adds a `## Timing, flakes, and how to assert that something is fast` section recording the rules those incidents actually produced.

### What it says

| Rule | Why it exists |
|---|---|
| **Wait on the event, never the clock** | Use `ChangeSignal`/`wait_for` from `tests/unit/harness/test_comms.py`; bound by loop turns (`MAX_PUMP_TURNS`) when there is nothing to subscribe to. A `sleep` tuned until CI goes green is a bet, not a test. |
| **Prefer a structural invariant to a numeric one** | Thread identity ("this ran off the loop") cannot flake; a duration can. `test_store_maintenance_callbacks_run_off_the_event_loop_thread` is the worked example. |
| **Measure CPU, not wall time** | Wall gaps conflate loop blocking with the OS not scheduling the process — 525-668 ms of "stall" was measured on an *idle* loop. Includes the measured wall-vs-CPU table and the caveat that the CPU clock is **blind to a pure blocking sleep**. |
| **Calibrate ceilings from CI, never your laptop** | Local readings of 36-38 ms and ~0 ms corresponded to **206, 264, 321, 413 ms** on CI at the same sites. Every locally-calibrated ceiling flaked a green tree within a day. |
| **Prove the guard can still go red** | A 500 ms ceiling was silently swallowing the 90-130 ms regression it existed to catch. Reintroduce the regression and watch it fail. |
| **Classify a flake before retuning** | Check whether `main` and other branches show the same failure before touching a threshold. |

It also records the trap of citing a stale number: the 549 ms figure in this repo's history is a **wall-clock** measurement from a probe that no longer exists, and it was cited to justify a CPU ceiling — comparing two different quantities.

## Testing Evidence

Docs-only change; no code, no tests, no version bump.

**Every cited symbol and file was verified to exist in the tree:**

```
class ChangeSignal            tests/unit/harness/test_comms.py          ✓
wait_for / MAX_PUMP_TURNS     tests/unit/harness/test_comms.py          ✓
DEADLOCK_GUARD_S              tests/unit/harness/test_comms.py          ✓
class LoopCpuProbe            tests/unit/tools/test_loop_liveness.py    ✓
class StallRecorder           tests/unit/test_tui_responsiveness.py     ✓
test_store_maintenance_...    tests/unit/test_tui_responsiveness.py     ✓
threading_get_ident usage     tests/unit/test_tui_responsiveness.py     ✓
await asyncio.to_thread(work) local_operator/session_factory.py         ✓
```

The `ChangeSignal` example was corrected during authoring after checking the real API — the first draft invented a `ChangeSignal(subscribe).wait(timeout=...)` signature that does not exist; the shipped text uses the actual `wait_for(predicate, signal=...)` form.

**Numbers in the doc are from CI logs, not recalled:** 525/668 ms scheduler-noise gaps, the 206/264/321/413 ms healthy-CPU dataset, and the 1056-1156 ms vs 393-492 ms core-speed spread all trace to specific runs cited in #486/#496/#498 and to `tests/unit/session/test_launch_subagent.py`.

**No absolute home paths** (`grep '/Users/\|/home/'` → empty), per the convention that AGENTS.md is shared and cloud-pushed.

Section is placed after "Visual validation" and before "TUI conventions", matching the existing document flow.

## Related

- #486 introduced the dual-signal probe (wall → CPU clock).
- #496 / #498 calibrated the ceiling, twice getting it wrong from local measurements.
- #461 fixed the same defect class ("wait on conditions instead of the clock") in three other tests.
- #499 replaced a timing assertion with the structural thread-identity one this doc holds up as the preferred shape.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -362,8 +362,8 @@ load. On an idle dev box the budget is 20-50x more than the work needs, so it
 looks solid; under `-n auto` contention on a 4-vCPU CI runner the work
 stretches, the budget does not, and the suite goes red for reasons that have
 nothing to do with the code under test. This repo has paid for that lesson
-repeatedly (#122, #373, #461, #486, #496, #498), so the rules below are not
-style preferences.
+repeatedly (#122, #373, #461, #486, #496, #498, #499), so the rules below are
+not style preferences.
 
 ### Wait on the event, never on the clock
 
@@ -509,7 +509,7 @@ Then check the other direction: run the file several times under load
 ### When a test is already flaking
 
 Reproduce and classify before touching a threshold. A single failure out of
-~9000 on a different timing-sensitive test each run is a flake, not a
+~9900 on a different timing-sensitive test each run is a flake, not a
 regression; the same test failing on several branches at once — check `main` —
 is repo-wide and not yours. Re-run to confirm, then fix the measurement rather
 than widening the bound. Do not merge a red head on the assumption that it is
@@ -527,7 +527,13 @@ behind it — which is the other half of why `wait_for` carries
 **A deadlock defeats every technique above.** If the failure mode is the event
 loop freezing at the C level rather than running slowly, a Python-level probe
 reports nothing — the coroutine that would record the sample never gets
-scheduled either. That class is covered by `tests/e2e/watchdog.py`, whose
+scheduled either. Nor do the obvious fallbacks: `watchdog.py` documents, with a
+measurement, that a `threading.Thread` watchdog needs a GIL the wedged thread
+never releases, and `pytest-timeout`'s default `signal` method never fires
+because a Python signal handler only runs between bytecodes. The pre-fix code
+ignored a 20 s thread watchdog and had to be `kill -9`'d.
+
+That class is covered by `tests/e2e/watchdog.py`, whose
 `faulthandler.dump_traceback_later(exit=True)` is armed in C and fires from a
 separate OS thread, so it survives a wedged interpreter and takes the process
 down with a full thread dump. It runs as its own CI stage (`tui-e2e`, both
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -375,19 +375,25 @@ assert store.detail is not None
 ```
 
 **Do** wait on a publication from the code under test. `tests/unit/harness/
-test_comms.py` has the pattern: `ChangeSignal` subscribes to the notifications
-the code already emits, and `wait_for` re-tests the predicate after each one:
+test_comms.py` has the pattern: `ChangeSignal` subscribes to notifications the
+code already emits, and `wait_for` re-tests the predicate after each one. The
+subscriptions are what make it work, so build and release it in full:
 
 ```python
-await wait_for(lambda: store.detail is not None, signal=signal)
-assert store.detail is not None
+signal = ChangeSignal().watch_comms(comms).watch_session(parent)
+try:
+    await wait_for(lambda: comms.session_dir_of(job_id) is not None, signal=signal)
+finally:
+    signal.close()
 ```
 
-There is no elapsed-time comparison anywhere on the success path — the wait
-lasts exactly as long as the work does. The only deadline is
-`DEADLOCK_GUARD_S`, which exists so a genuine hang fails the run with "it is
-wedged, not merely slow" instead of blocking forever. It is a backstop, not
-the assertion.
+A bare `ChangeSignal()` subscribes to nothing, so `wait_for` blocks the full
+`DEADLOCK_GUARD_S` and then reports "it is wedged, not merely slow" — a
+confident diagnosis of the wrong thing. Watch the sources you need.
+
+There is no elapsed-time comparison anywhere on the success path: the wait
+lasts exactly as long as the work does. The deadline exists so a genuine hang
+fails the run instead of blocking forever. It is a backstop, not the assertion.
 
 When there is nothing to subscribe to — an in-process coroutine the test
 itself scheduled — bound by **loop turns** (`MAX_PUMP_TURNS`) rather than
@@ -406,6 +412,16 @@ none of them equals the loop thread. That assertion cannot flake: it is a fact
 about where the code ran, not how long it took, and it fails deterministically
 the moment someone drops an `asyncio.to_thread`.
 
+This is not a stylistic preference — it is where this repo's timing bounds keep
+landing. `tests/unit/session/test_launch_subagent.py` carries the full account
+under "WHY THIS IS A STRUCTURAL SPY AND NOT A TIME BOUND": a flat 1.0 s wall
+bound failed, a calibrated bound failed, converting to `time.thread_time`
+removed the *load* sensitivity but not the *core-speed* sensitivity, and the
+conclusion was that **no portable numeric bound exists** for that site — CI's
+healthy 1056-1156 ms sat above the box's 1315 ms regression threshold. It was
+converted to a structural spy. Read that docstring before you reach for a
+number.
+
 Reach for a number only when no structural fact expresses the property.
 
 ### If you must measure, measure CPU, not wall time
@@ -417,9 +433,12 @@ with the loop idle and nothing blocked.
 
 Use `time.thread_time()`, which is per-thread and excludes time asleep or
 waiting on the GIL, so a sample is large only when the loop thread really ran
-without yielding. `LoopCpuProbe` in `tests/unit/tools/test_loop_liveness.py`
-and `StallRecorder` in `tests/unit/test_tui_responsiveness.py` are the two
-implementations; reuse one rather than writing a third.
+without yielding. Two probes exist and they are **not interchangeable**:
+`LoopCpuProbe` (`tests/unit/tools/test_loop_liveness.py`) records CPU only;
+`StallRecorder` (`tests/unit/test_tui_responsiveness.py`) records CPU **and**
+wall, and asserts both. If you need the blocking-sleep backstop described
+below, use `StallRecorder` or add the wall assertion yourself — reusing
+`LoopCpuProbe` silently drops it.
 
 The distinction is measurable, not theoretical:
 
@@ -440,22 +459,26 @@ loaded-bar sites read 36-38 ms and ~0 ms; the CI runners recorded **206, 264,
 321, and 413 ms** of entirely legitimate loop CPU at those same sites. Every
 ceiling calibrated from the local numbers flaked a green tree within a day.
 
-A slower core burns more CPU-seconds on identical work —
-`tests/unit/session/test_launch_subagent.py` documents ubuntu-latest reporting
-1056-1156 ms against a bound a dev box met at 393-492 ms, on an unmodified
-healthy tree. So:
+A slower core burns more CPU-seconds on identical work, so a bound calibrated
+on a dev box is not a bound on CI — see the `test_launch_subagent.py` account
+above, where that spread ended in abandoning the bound entirely. If you have
+established a number is genuinely the only option:
 
 - Take the healthy maximum from **CI logs across several runs**, not from one
   local run and not from one CI run.
 - Leave real headroom above it, and **write the dataset and the margin into the
   docstring** so the next person is not re-deriving it from scratch.
-- Say what to do when it is exceeded. If a green tree trips the ceiling, the
-  ceiling is wrong until proven otherwise — raise it on evidence rather than
-  assuming the probe found a regression.
 - A number copied from an older comment is not evidence. The 549 ms figure in
   this repo's history is a **wall-clock** measurement from a probe that no
   longer exists; citing it to justify a CPU ceiling compares two different
   quantities.
+
+When a green tree trips the ceiling, **find out which it is before touching the
+number**. Either the tree is not as green as it looks, or the ceiling is
+miscalibrated — and "raise it" is only correct in the second case. Reproduce
+the sample, check whether the same test fails on `main` and on unrelated
+branches, and confirm the work under the probe is the legitimate kind. Widening
+a bound because it went red is how a guard stops guarding.
 
 ### Prove the test can still fail
 
@@ -469,10 +492,15 @@ watch it fail:
 # expect a failure with a message that names the real cause, then revert
 ```
 
-This is how a 500 ms ceiling was found to be silently swallowing the ~90-130 ms
-regression it was written to catch. Then check the other direction: run the file
-several times under load (`for i in $(seq 1 5)` with a few CPU spinners) and
-confirm it stays green.
+This is not hypothetical. In `test_tui_responsiveness.py` the loaded bar was
+raised to 500 ms to stop it flaking, which meant a 200-500 ms stall at the
+boot and turn sites — whose only guard is that bar — would have passed
+unnoticed. The strict 50 ms bar on the reconnect/connect sites was deliberately
+left alone precisely so the 90-130 ms parse regression stayed catchable. When
+you widen a ceiling, name what you just stopped catching.
+
+Then check the other direction: run the file several times under load
+(`for i in $(seq 1 5)` with a few CPU spinners) and confirm it stays green.
 
 ### When a test is already flaking
 
@@ -482,6 +510,25 @@ regression; the same test failing on several branches at once — check `main` �
 is repo-wide and not yours. Re-run to confirm, then fix the measurement rather
 than widening the bound. Do not merge a red head on the assumption that it is
 "probably the known flake" without checking the log.
+
+### Two things this section cannot do for you
+
+**There is no `pytest-timeout` in this suite.** A test that waits forever hangs
+its CI job until the workflow's `timeout-minutes` reclaims the runner (40 min
+for `test`, and that ceiling exists because a job once held a slot for 3h38m).
+So an unbounded wait is not merely slow, it is expensive for everyone queued
+behind it — which is the other half of why `wait_for` carries
+`DEADLOCK_GUARD_S`.
+
+**A deadlock defeats every technique above.** If the failure mode is the event
+loop freezing rather than running slowly, no in-process probe reports anything,
+because the process that would report is the one that is stuck. That class is
+covered by `tests/e2e/watchdog.py`, whose
+`faulthandler.dump_traceback_later(exit=True)` survives a kernel-level deadlock
+and takes the process down with a full thread dump. It runs as its own CI stage
+(`tui-e2e`, both Linux and macOS) rather than in the unit run, since firing it
+under `-n auto` would kill a worker carrying unrelated tests. If you are
+guarding against a hang rather than a stall, that is the file to read.
 
 ## TUI conventions worth knowing before you edit a widget
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -374,10 +374,11 @@ await asyncio.sleep(2.0)              # "surely it finished by now"
 assert store.detail is not None
 ```
 
-**Do** wait on a publication from the code under test. `tests/unit/harness/
-test_comms.py` has the pattern: `ChangeSignal` subscribes to notifications the
-code already emits, and `wait_for` re-tests the predicate after each one. The
-subscriptions are what make it work, so build and release it in full:
+**Do** wait on a publication from the code under test.
+`tests/unit/harness/test_comms.py` has the pattern: `ChangeSignal` subscribes
+to notifications the code already emits, and `wait_for` re-tests the predicate
+after each one. The subscriptions are what make it work, so build and release
+it in full:
 
 ```python
 signal = ChangeSignal().watch_comms(comms).watch_session(parent)
@@ -417,10 +418,10 @@ landing. `tests/unit/session/test_launch_subagent.py` carries the full account
 under "WHY THIS IS A STRUCTURAL SPY AND NOT A TIME BOUND": a flat 1.0 s wall
 bound failed, a calibrated bound failed, converting to `time.thread_time`
 removed the *load* sensitivity but not the *core-speed* sensitivity, and the
-conclusion was that **no portable numeric bound exists** for that site — CI's
-healthy 1056-1156 ms sat above the box's 1315 ms regression threshold. It was
-converted to a structural spy. Read that docstring before you reach for a
-number.
+conclusion was that **no portable numeric bound exists** for that site — the
+only window left was between CI's healthy 1156 ms and this box's 1315 ms
+regression, which is too narrow to sit a bound in. It was converted to a
+structural spy. Read that docstring before you reach for a number.
 
 Reach for a number only when no structural fact expresses the property.
 
@@ -494,10 +495,13 @@ watch it fail:
 
 This is not hypothetical. In `test_tui_responsiveness.py` the loaded bar was
 raised to 500 ms to stop it flaking, which meant a 200-500 ms stall at the
-boot and turn sites — whose only guard is that bar — would have passed
-unnoticed. The strict 50 ms bar on the reconnect/connect sites was deliberately
-left alone precisely so the 90-130 ms parse regression stayed catchable. When
-you widen a ceiling, name what you just stopped catching.
+boot and turn sites — whose only guard is that bar — would pass unnoticed.
+That trade was made deliberately and on evidence: those sites recorded 321 ms
+and 413 ms of *legitimate* loop CPU on CI, so a tighter bar failed green trees
+instead. The strict 50 ms bar on the reconnect/connect sites was left alone
+precisely so the 90-130 ms parse regression stayed catchable. When you widen a
+ceiling, name what you just stopped catching — and record the measurements that
+forced your hand.
 
 Then check the other direction: run the file several times under load
 (`for i in $(seq 1 5)` with a few CPU spinners) and confirm it stays green.
@@ -521,14 +525,15 @@ behind it — which is the other half of why `wait_for` carries
 `DEADLOCK_GUARD_S`.
 
 **A deadlock defeats every technique above.** If the failure mode is the event
-loop freezing rather than running slowly, no in-process probe reports anything,
-because the process that would report is the one that is stuck. That class is
-covered by `tests/e2e/watchdog.py`, whose
-`faulthandler.dump_traceback_later(exit=True)` survives a kernel-level deadlock
-and takes the process down with a full thread dump. It runs as its own CI stage
-(`tui-e2e`, both Linux and macOS) rather than in the unit run, since firing it
-under `-n auto` would kill a worker carrying unrelated tests. If you are
-guarding against a hang rather than a stall, that is the file to read.
+loop freezing at the C level rather than running slowly, a Python-level probe
+reports nothing — the coroutine that would record the sample never gets
+scheduled either. That class is covered by `tests/e2e/watchdog.py`, whose
+`faulthandler.dump_traceback_later(exit=True)` is armed in C and fires from a
+separate OS thread, so it survives a wedged interpreter and takes the process
+down with a full thread dump. It runs as its own CI stage (`tui-e2e`, both
+Linux and macOS) rather than in the unit run, since firing it under `-n auto`
+would kill a worker carrying unrelated tests. If you are guarding against a
+hang rather than a stall, that is the file to read.
 
 ## TUI conventions worth knowing before you edit a widget
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -354,6 +354,135 @@ so they are opt-in (`LO_RUN_SNAPSHOTS=1`) and regenerated with
 `--snapshot-update`. Do not add a golden as a substitute for looking at the
 change.
 
+## Timing, flakes, and how to assert that something is fast
+
+A test that spends a fixed budget waiting for asynchronous work, then asserts
+as though the budget were a guarantee, is not a test — it is a bet on machine
+load. On an idle dev box the budget is 20-50x more than the work needs, so it
+looks solid; under `-n auto` contention on a 4-vCPU CI runner the work
+stretches, the budget does not, and the suite goes red for reasons that have
+nothing to do with the code under test. This repo has paid for that lesson
+repeatedly (#122, #373, #461, #486, #496, #498), so the rules below are not
+style preferences.
+
+### Wait on the event, never on the clock
+
+**Do not** poll a wall-clock deadline and assert afterwards:
+
+```python
+await asyncio.sleep(2.0)              # "surely it finished by now"
+assert store.detail is not None
+```
+
+**Do** wait on a publication from the code under test. `tests/unit/harness/
+test_comms.py` has the pattern: `ChangeSignal` subscribes to the notifications
+the code already emits, and `wait_for` re-tests the predicate after each one:
+
+```python
+await wait_for(lambda: store.detail is not None, signal=signal)
+assert store.detail is not None
+```
+
+There is no elapsed-time comparison anywhere on the success path — the wait
+lasts exactly as long as the work does. The only deadline is
+`DEADLOCK_GUARD_S`, which exists so a genuine hang fails the run with "it is
+wedged, not merely slow" instead of blocking forever. It is a backstop, not
+the assertion.
+
+When there is nothing to subscribe to — an in-process coroutine the test
+itself scheduled — bound by **loop turns** (`MAX_PUMP_TURNS`) rather than
+seconds. A turn count survives contention that a wall-clock budget does not.
+
+If you catch yourself tuning a `sleep` until CI goes green, you are
+calibrating a bet, not fixing a test.
+
+### Prefer a structural invariant to a numeric one
+
+The strongest version of "this work happened off the event loop" is not a
+timing bound at all — it is thread identity.
+`test_store_maintenance_callbacks_run_off_the_event_loop_thread` wraps the real
+maintenance callbacks, records `threading.get_ident()` inside each, and asserts
+none of them equals the loop thread. That assertion cannot flake: it is a fact
+about where the code ran, not how long it took, and it fails deterministically
+the moment someone drops an `asyncio.to_thread`.
+
+Reach for a number only when no structural fact expresses the property.
+
+### If you must measure, measure CPU, not wall time
+
+Wall-clock gaps between probe wakes conflate two unrelated things: the event
+loop genuinely blocking, and the OS simply not scheduling the process. On a
+loaded runner the second dominates — measured here at 525-668 ms of "stall"
+with the loop idle and nothing blocked.
+
+Use `time.thread_time()`, which is per-thread and excludes time asleep or
+waiting on the GIL, so a sample is large only when the loop thread really ran
+without yielding. `LoopCpuProbe` in `tests/unit/tools/test_loop_liveness.py`
+and `StallRecorder` in `tests/unit/test_tui_responsiveness.py` are the two
+implementations; reuse one rather than writing a third.
+
+The distinction is measurable, not theoretical:
+
+| scenario                   | wall gap    | CPU gap |
+| -------------------------- | ----------- | ------- |
+| 60 MB sync parse on loop   | 96 ms       | 96 ms   |
+| pure `time.sleep` on loop  | 306 ms      | 0.1 ms  |
+| loop idle, OS-starved      | 525-668 ms  | 0.0 ms  |
+
+Note the middle row: the CPU clock is **blind to a pure blocking sleep**. If
+that shape matters at your site, keep a wall-clock assertion alongside it, with
+a ceiling set for catastrophe (seconds) rather than for precision.
+
+### Calibrate ceilings from CI, never from your laptop
+
+This is the mistake that cost three PRs. Local measurements of the same
+loaded-bar sites read 36-38 ms and ~0 ms; the CI runners recorded **206, 264,
+321, and 413 ms** of entirely legitimate loop CPU at those same sites. Every
+ceiling calibrated from the local numbers flaked a green tree within a day.
+
+A slower core burns more CPU-seconds on identical work —
+`tests/unit/session/test_launch_subagent.py` documents ubuntu-latest reporting
+1056-1156 ms against a bound a dev box met at 393-492 ms, on an unmodified
+healthy tree. So:
+
+- Take the healthy maximum from **CI logs across several runs**, not from one
+  local run and not from one CI run.
+- Leave real headroom above it, and **write the dataset and the margin into the
+  docstring** so the next person is not re-deriving it from scratch.
+- Say what to do when it is exceeded. If a green tree trips the ceiling, the
+  ceiling is wrong until proven otherwise — raise it on evidence rather than
+  assuming the probe found a regression.
+- A number copied from an older comment is not evidence. The 549 ms figure in
+  this repo's history is a **wall-clock** measurement from a probe that no
+  longer exists; citing it to justify a CPU ceiling compares two different
+  quantities.
+
+### Prove the test can still fail
+
+A guard that cannot go red is worse than no guard, because it is believed. When
+you change how a test detects a regression, reintroduce the regression and
+watch it fail:
+
+```sh
+# put the synchronous call back on the loop, or delete the to_thread
+.venv/bin/python -m pytest tests/unit/test_tui_responsiveness.py::<test> -n0
+# expect a failure with a message that names the real cause, then revert
+```
+
+This is how a 500 ms ceiling was found to be silently swallowing the ~90-130 ms
+regression it was written to catch. Then check the other direction: run the file
+several times under load (`for i in $(seq 1 5)` with a few CPU spinners) and
+confirm it stays green.
+
+### When a test is already flaking
+
+Reproduce and classify before touching a threshold. A single failure out of
+~9000 on a different timing-sensitive test each run is a flake, not a
+regression; the same test failing on several branches at once — check `main` —
+is repo-wide and not yours. Re-run to confirm, then fix the measurement rather
+than widening the bound. Do not merge a red head on the assumption that it is
+"probably the known flake" without checking the log.
+
 ## TUI conventions worth knowing before you edit a widget
 
 - **Do not shadow Textual's API.** `Widget` already owns `query`, `visible`,


damianvtran:
### Agent review — round 1

Scope: origin/main..dev-agents-flake-guidance (head e51894e5)

Docs-only, 129 added lines in `AGENTS.md`. I verified every cited symbol, file
path, and number against the tree and against the history of #486/#496/#498/#499.
The symbols all exist and the wall-vs-CPU table is faithful to
`StallRecorder`'s docstring. **No BLOCKER.** But five claims are inaccurate or
would mislead a reader who follows them, including two where the doc contradicts
the very source it cites.

---

**D1-1 — MAJOR — `AGENTS.md:381-384`**

The `ChangeSignal` example omits the construction and subscription steps, and a
reader copying it gets a 30-second hang with a misleading failure message.

The snippet is:

```python
await wait_for(lambda: store.detail is not None, signal=signal)
```

`signal` appears from nowhere. The real form (`tests/unit/harness/test_comms.py:1706`)
is `signal = ChangeSignal().watch_comms(comms).watch_session(parent)`, torn down
with `signal.close()` in a `finally` (`:1745`, `:1776`). A bare `ChangeSignal()`
subscribes to nothing, so `_fire` is never called, `settle()` blocks on an
`asyncio.Event` nobody sets, and `wait_for` runs out `DEADLOCK_GUARD_S` (30 s)
and raises *"the code under test published no change satisfying the condition in
30s — it is wedged, not merely slow"* (`:163-166`). That message names the wrong
cause: the code under test was fine; the signal was never wired. This is a doc
about not writing tests that fail for reasons unrelated to the code under test,
and its headline example produces exactly that.

Prose at `:377-379` says "`ChangeSignal` subscribes to the notifications the code
already emits" — passive, so it reads as automatic rather than as something the
author must wire to specific objects.

Fix: show the three lines, not one — `signal = ChangeSignal().watch_comms(comms).watch_session(parent)`,
the `wait_for`, and `signal.close()` in a `finally`; and state that a signal with
no `watch_*` call cannot fire.

---

**D1-2 — MAJOR — `AGENTS.md:472-473`**

> This is how a 500 ms ceiling was found to be silently swallowing the ~90-130 ms
> regression it was written to catch.

Two different sites and two different ceilings have been merged into one claim,
and the resulting sentence is false in both directions.

- The 90-130 ms regression is the reconnect/connect one, guarded by the **strict
  50 ms** CPU bar (`tests/unit/test_tui_responsiveness.py:73-74`, `:449`, `:470`).
  That bar has never been 500 ms — #496's commit message states it explicitly:
  *"The strict 50 ms bar on reconnect/connect is unchanged."*
- What the 500 ms raise actually swallowed was different work at different sites.
  #496's second commit: *"s1-boot and s2 record ~0 ms healthy CPU and have no other
  guard, so ... raising the shared default from 200 to 500 silently swallowed
  **200-500 ms** stalls there."*

A 500 ms ceiling was never "written to catch" a 90-130 ms regression, so the
sentence's own logic does not close.

Secondary problem: the anecdote does not illustrate the technique it is attached
to. The section prescribes *reintroduce the regression and watch it fail*; the
200→500 over-raise was caught by **review reasoning about site coverage**
(#496 round 1, finding C1-2), not by a mutation run.

Fix: *"This is how raising a shared 200 ms default to 500 ms was found to be
silently swallowing the 200-500 ms stalls at the s1-boot and s2 sites, whose only
guard it was"* — and attach the mutation advice to the reconnect/50 ms case,
where a mutation genuinely is the check.

---

**D1-3 — MAJOR — `AGENTS.md:452-454`**

> If a green tree trips the ceiling, the ceiling is wrong until proven otherwise
> — raise it on evidence rather than assuming the probe found a regression.

As written this is a ratchet that only ever turns one way, and it contradicts the
section 8 lines below it.

Every real loop-blocking regression that lands arrives as "a tree with no
obviously related change tripping the ceiling" — that is what a regression looks
like from the probe's side. Putting the burden of proof on the *probe* ("wrong
until proven otherwise") licenses widening the bound in precisely the case the
bound exists for, and this repo has already run that loop three times
(200 → 500 → 200 → 500). It also sits directly against **"Prove the test can
still fail"** at `:460`, which argues a guard that cannot go red is worse than no
guard.

The source docstring is more careful than the doc: `test_tui_responsiveness.py:80-82`
says *"if it does, raise the ceiling again **on evidence** rather than assuming the
probe is wrong"* — where "evidence" means a CI dataset across runs, not a single
failure. The doc keeps the words and drops the constraint.

Also, "green tree" is circular here: a tree that trips this ceiling is by
definition not green. The intended meaning is "a tree carrying no change that
could plausibly add loop work".

Fix: invert the default — *"Before raising a ceiling, establish that the diff
under it cannot have added loop work, and re-run the regression mutation to
confirm the raised ceiling still goes red. Raise only on a multi-run CI dataset;
a single failure is not one."*

---

**D1-4 — MAJOR — `AGENTS.md:443-446`**

> `tests/unit/session/test_launch_subagent.py` documents ubuntu-latest reporting
> 1056-1156 ms against a bound a dev box met at 393-492 ms

Both halves misread the source (`test_launch_subagent.py:2154-2162`):

- No bound of 393-492 ms exists. **The bound was 950 ms.** 393-492 ms is a
  *measurement*: loop-thread CPU on the dev box **under 8 CPU hogs**, quoted there
  to show that CPU time is load-immune where wall time (943-2034 ms) is not. The
  doc converts a measurement into a threshold.
- The citation is used to support "calibrate the number from CI", but the source's
  conclusion is that **no number works at all**: *"No portable numeric bound can sit
  above CI's healthy 1156 ms and below this box's 1315 ms regression."* That test
  was consequently converted to a structural spy on `asyncio.to_thread`. The doc
  cites, as its evidence for how to pick a numeric bound, the one case in the repo
  that proved a numeric bound was impossible — and drops that conclusion.

This matters because it is the doc's only quantified argument for core-speed
spread, and a reader who opens the cited file finds it arguing the opposite.

Fix: *"...reported 1056-1156 ms of loop-thread CPU against a 950 ms bound on an
unmodified healthy tree, while the same work cost this dev box 393-492 ms — a
spread wide enough that no portable number fits, which is why that test was
converted to a structural assertion instead."* That also earns the forward
reference to the structural-invariant section.

---

**D1-5 — MAJOR — `AGENTS.md:420-422`**

> `LoopCpuProbe` ... and `StallRecorder` ... are the two implementations; reuse
> one rather than writing a third.

They are not two implementations of one thing, and the difference is the exact
distinction this section is about.

- `LoopCpuProbe` (`test_loop_liveness.py:89-120`) is **CPU-only**: `samples`,
  `worst`, nothing else.
- `StallRecorder` (`test_tui_responsiveness.py:42-70`) is **dual-clock**:
  `cpu_stalls` *and* `wall_stalls`, asserted together via
  `assert_no_stall(cpu_ceiling_ms, wall_ceiling_ms)`.

Six lines later (`:432-435`) the doc tells the reader to keep a wall-clock
assertion alongside the CPU one where a blocking sleep matters. A reader who
follows "reuse one" and picks `LoopCpuProbe` silently loses the wall backstop
they were just told to keep — the doc's own middle table row (`306 ms / 0.1 ms`)
becomes invisible at their site.

"Reuse one" is also not actionable: both are private classes in test modules in
different packages, with no shared harness home. The tree deliberately did *not*
reuse — `StallRecorder`'s docstring (`:50-51`) says it "adopted" the statistic
from the other file, i.e. copied it. The doc prescribes a practice its own
evidence shows was not followed, without proposing where a shared probe would live.

Fix: name the difference and make the choice a rule — *"`LoopCpuProbe` measures CPU
only; `StallRecorder` measures CPU and wall together. Use `StallRecorder` unless a
blocking sleep is impossible at your site. They are duplicated on purpose (each is
local to its test module); if you need a third, lift a shared one into a harness
module rather than copying again."*

---

**D1-6 — MINOR — `AGENTS.md:377-378`**

The code span breaks across a newline:

```
`tests/unit/harness/
test_comms.py`
```

CommonMark converts the line ending inside a code span to a space, so this renders
as `tests/unit/harness/ test_comms.py` — a path with a space in it, not
copy-pasteable and not greppable. I checked the whole file: these are the **only
two lines in `AGENTS.md` with an unbalanced backtick count**, so it is a
one-off against the document's own convention.

Fix: move the whole path onto one line, even if it overruns the wrap width.

---

**D1-7 — MINOR — `AGENTS.md:386-388`**

> There is no elapsed-time comparison anywhere on the success path ... The only
> deadline is `DEADLOCK_GUARD_S`

True of `wait_for`, but the sentence is scoped to the file, and the file has
another: `test_comms.py:623` uses `await asyncio.wait_for(question, timeout=2.0)`
— a hard 2 s wall-clock deadline. A reader told "the only deadline is
`DEADLOCK_GUARD_S`" who then finds a 2.0 s timeout in the cited file will not know
whether it is an oversight or a deliberate exception.

Fix: scope it — "no elapsed-time comparison on `wait_for`'s success path".

---

**D1-8 — MINOR — `AGENTS.md:386-397` (omission)**

The section never states the suite-wide fact that makes its advice load-bearing:
**there is no `pytest-timeout` in this repo** (`test_comms.py:64`: *"this suite has
no pytest-timeout"*; nothing in `pyproject.toml` provides one). That is *why*
`DEADLOCK_GUARD_S` and `MAX_PUMP_TURNS` exist.

Follow "wait on the event, never on the clock" literally — `await some_event.wait()`
with no backstop — and a wedged test hangs the CI job until the runner reaps it,
with no diagnostic. The doc describes `DEADLOCK_GUARD_S` as a property `wait_for`
already has, never as an obligation on the author of a *new* wait.

Fix: one sentence — "the suite has no `pytest-timeout`, so any new unbounded wait
must carry its own deadlock backstop; a hang here costs a whole CI job."

---

**D1-9 — MINOR — `AGENTS.md:357-484` (omission)**

No cross-reference to `tests/e2e/watchdog.py`, which documents the one timing
failure class where all of this section's tools provably do not work: a
kernel-level `flock`/`close` deadlock (#401) parks threads inside syscalls, so
`asyncio.wait_for`, thread watchdogs, and signal-based timeouts **all fail to
fire** — measured, per that module's header, not assumed. Only
`faulthandler.dump_traceback_later(exit=True)` survives it.

A reader who absorbs only this new section will reach for a loop-turn bound or an
`asyncio` timeout against a hang those cannot detect. `AGENTS.md:84-95` already
covers this, ~280 lines earlier, and the new section does not point at it.

Fix: one line at the end — "If the failure mode is a *hang* rather than a slow
assertion, none of the above fires; see `tests/e2e/watchdog.py` and the e2e notes
above."

---

**D1-10 — NIT — `AGENTS.md:479-481`**

"A single failure out of ~9000" — the unit suite collects **9891** on this head
(`pytest tests/unit --collect-only`). In a section arguing for numeric precision,
round to ~9900 or say "out of the ~10k unit tests".

---

**D1-11 — NIT — `AGENTS.md:363-364`**

The incident list `(#122, #373, #461, #486, #496, #498)` omits **#499**, which is
where `test_store_maintenance_callbacks_run_off_the_event_loop_thread` — the
worked example of the entire structural-invariant section (`:401-407`) — came
from (commit `e43f312f`). Worth adding since the doc leans on it.

(For what it is worth: #122 and #373 are issues, #461/#486/#496/#498 are PRs. All
six resolve and are on-topic; no action needed, just noting they are not a
homogeneous list.)

---

**D1-12 — NIT — `AGENTS.md:361-362`**

The opening leans on `-n auto` contention on a 4-vCPU runner, which is explained
in detail 340 lines earlier under "Environment" (`:15-60`, including the
worker-cap hook and the `CI`-set exemption that keeps all 4 cores on CI). A back
reference would save the reader a search — and the "4-vCPU runner" figure is only
meaningful alongside the note at `:55-59` that CI does *not* apply the CPU share.

---

**Verified and correct — no action:**

- All eight cited symbols exist with the described shapes: `ChangeSignal`,
  `wait_for`, `MAX_PUMP_TURNS` (3000), `DEADLOCK_GUARD_S` (30.0) in
  `tests/unit/harness/test_comms.py`; `LoopCpuProbe` in
  `tests/unit/tools/test_loop_liveness.py`; `StallRecorder` and
  `test_store_maintenance_callbacks_run_off_the_event_loop_thread` in
  `tests/unit/test_tui_responsiveness.py`; `await asyncio.to_thread(work)` at
  `local_operator/session_factory.py:1295`.
- The wall-vs-CPU table (`96/96`, `306/0.1`, `525-668/0.0`) matches
  `StallRecorder`'s docstring `:62-70` exactly.
- The CI dataset **206, 264, 321, 413 ms** matches `:77-78` and `:158-159`, and
  the local **36-38 ms** / **~0 ms** pairing is correctly attributed (36-38 =
  title-scan, ~0 = s1-boot/s2, per #496's two commit messages).
- The **549 ms** trap is right and is the doc's best contribution: it is
  wall-clock, from the pre-#486 probe, and it *was* cited in #498's commit message
  to justify a CPU ceiling. Confirmed at `:81-83` and `:161-163`.
- Description of the structural test is accurate — it wraps the four real
  callbacks, records `threading.get_ident()`, and asserts each ran once and not on
  the loop thread (`:332-385`).
- No absolute home paths (`grep '/Users/\|/home/'` → empty). Convention honoured.
- The suggested command at `:468` runs clean as written: `1 passed in 1.42s`. It
  does **not** need the `TERM`/`NO_COLOR` env that `AGENTS.md:71` requires for
  `tests/unit/tui`, because this file is not under that directory — I checked
  rather than assumed.
- Placement between "Visual validation" and "TUI conventions" is defensible; both
  neighbours are "how to validate a change".

**CI on head e51894e5:** `lint`, `type-check`, `pip-audit`, `tui-e2e`
(ubuntu + macos) all pass; `test (3.12)` still pending at review time. Docs-only,
so nothing here depends on it, but confirm green before merge.

---

**Verdict: changes requested.** 0 BLOCKER, 5 MAJOR, 4 MINOR, 3 NIT. The section is
worth having and most of it is accurate, but D1-2 and D1-4 attribute numbers to
the wrong sites and invert the conclusion of a cited source, D1-3 hands the reader
a ceiling-widening ratchet the doc's own next section argues against, D1-5 flattens
the CPU/wall distinction the section exists to teach, and D1-1's copy-paste example
hangs for 30 s and blames the wrong thing. All five are in the doc's load-bearing
claims, not its margins. Round is not terminal.

damianvtran:

Scope: origin/main..dev-agents-flake-guidance (head `876ebf75`)

All five MAJORs fixed, plus both omissions. Every one was the doc being confidently wrong rather than merely thin, which in a document that exists to prevent confidently-wrong test calibration is the worst failure mode available.

---

**D1-1 (MAJOR) — `ChangeSignal` example unusable — FIXED.** You are right, and the irony was not lost: a section about tests failing for reasons unrelated to the code under test shipped an example whose failure mode is a 30-second block and a confident "it is wedged, not merely slow" about a signal that was simply never subscribed. Replaced with the real form, matching the two call sites verbatim:

```python
signal = ChangeSignal().watch_comms(comms).watch_session(parent)
try:
    await wait_for(lambda: comms.session_dir_of(job_id) is not None, signal=signal)
finally:
    signal.close()
```

plus an explicit sentence on what a bare `ChangeSignal()` does.

**D1-4 (MAJOR) — misread `test_launch_subagent.py` — FIXED, and this was the sharpest finding.** I cited the repo's own proof that numeric bounds do not work as my evidence for how to choose one. That docstring's conclusion is that **no portable numeric bound exists** for that site (CI's healthy 1056-1156 ms above the box's 1315 ms regression), and it was converted to a structural spy. The 393-492 ms figure was a measurement under 8 CPU hogs, not a bound — the bound was 950 ms.

The citation moved to the **structural-invariant** section, where it now says what it actually says, with a pointer to read that docstring before reaching for a number. The calibration section no longer claims it as support.

**D1-3 (MAJOR) — one-way ratchet — FIXED.** "The ceiling is wrong until proven otherwise" licensed widening the bound in precisely the case the bound exists for, and contradicted "Prove the test can still fail" eight lines down. Now: find out which it is first — either the tree is not as green as it looks, or the ceiling is miscalibrated, and "raise it" is only correct in the second case — with the concrete checks (reproduce, check `main` and unrelated branches, confirm the work is the legitimate kind).

**D1-2 (MAJOR) — conflated two sites — FIXED.** Correct: the 90-130 ms parse regression is guarded by the **50 ms** bar, which was never changed; the 500 ms raise swallowed **200-500 ms** stalls at the boot and turn sites. Both are now stated accurately, and the passage carries the rule the incident actually teaches: *when you widen a ceiling, name what you just stopped catching.*

**D1-5 (MAJOR) — probes described as interchangeable — FIXED.** `LoopCpuProbe` is CPU-only, `StallRecorder` is dual-clock. The doc now says they are **not interchangeable** and warns that reusing the former silently drops the wall backstop the same section tells you to keep six lines later.

**Omissions — both added.** A new subsection covers (a) there is no `pytest-timeout` here, so an unbounded wait hangs its job until the workflow's 40-minute `timeout-minutes` reclaims the runner — a cost borne by everything queued behind it, which is the other half of why `wait_for` carries `DEADLOCK_GUARD_S`; and (b) a deadlock defeats every technique in the section, because the process that would report is the stuck one — that class belongs to `tests/e2e/watchdog.py` and its `faulthandler.dump_traceback_later(exit=True)`, which is why it runs as its own CI stage rather than under `-n auto`.

---

### Verification of the new text

Every added claim was checked against the tree rather than recalled:

```
dump_traceback_later in tests/e2e/watchdog.py          ✓ (4 occurrences)
tui-e2e matrix os: [ubuntu-latest, macos-latest]       ✓
test job timeout-minutes: 40                           ✓
pytest-timeout in pyproject.toml                       ✓ absent (claim correct)
ChangeSignal().watch_comms(comms).watch_session(parent) ✓ 2 real call sites
no absolute home paths                                 ✓
```

The head moved from `e51894e5` to `876ebf75`, so a round-2 convergence pass follows to re-pin freshness.

damianvtran:
### Agent review — round 2

Reviewer: independent `reviewer` subagent, anthropic/claude-opus-5
Scope: `e51894e5..876ebf75` (remediation delta), full diff `origin/main..dev-agents-flake-guidance`
Mode: Remediation convergence

All five round-1 MAJORs are genuinely fixed and both omissions are added. I
re-derived every new number from the tree rather than trusting the remediation
comment — and that is where the one problem is: the fix for **D1-4** introduced
a reversed inequality. Everything else in the delta checks out.

---

### Round-1 remediation status

| # | Finding | Status | Evidence |
|---|---|---|---|
| D1-1 | `ChangeSignal` example unusable | **FIXED** | Matches `test_comms.py:1706-1708` and the `finally: signal.close()` at `:1744-1745` verbatim; the added "a bare `ChangeSignal()` subscribes to nothing" is exactly right against `_fire`/`_unsubscribes` (`:103-128`). |
| D1-2 | 50 ms / 500 ms sites conflated | **FIXED** | Now correctly: 90-130 ms parse regression → strict 50 ms bar, never changed (`#498` commit: *"The strict 50 ms bar on reconnect/connect is unchanged"*; `StallRecorder:73-74`); 500 ms raise → 200-500 ms stalls at boot/turn (`test_s1_boot`:1074, `s2`:1168 both call `assert_no_stall_loaded`). |
| D1-3 | One-way ratchet | **FIXED** | "Find out which it is before touching the number" + reproduce / check `main` / confirm legitimate work. No longer contradicts "Prove the test can still fail". |
| D1-4 | Misread `test_launch_subagent.py` | **FIXED in substance, one new error** — see D2-1 | 393-492 ms is correctly dropped as a bound (it is the under-8-CPU-hogs measurement, `:2155`); the citation moved to the structural section and now carries the source's actual conclusion. |
| D1-5 | Probes "interchangeable" | **FIXED** | `LoopCpuProbe` is `samples`/`worst` only (`test_loop_liveness.py:89-120`); `StallRecorder` asserts `cpu_ceiling_ms`/`wall_ceiling_ms` (`:127-130`). The "silently drops it" warning is correct. |
| D1-7 | "The only deadline is `DEADLOCK_GUARD_S`" | **FIXED** | The over-scoped "only deadline" clause is gone; what remains matches `wait_for`'s own docstring (`:145-147`). |
| Omission (pytest-timeout) | **ADDED, accurate** | Absent from `pyproject.toml`; no `pytest.ini`/`setup.cfg`/`tox.ini` exists. `test` `timeout-minutes: 40` and the 3h38m anecdote both match `ci.yml:142-149`. |
| Omission (e2e watchdog) | **ADDED, accurate** | `faulthandler.dump_traceback_later(..., exit=True)` at `watchdog.py:104`; `tui-e2e` matrix `os: [ubuntu-latest, macos-latest]` (`ci.yml:419`); the "would kill a worker carrying unrelated tests" rationale matches `pyproject.toml`'s `-m "not e2e"` comment verbatim. |

No absolute home paths (`grep '/Users/\|/home/'` → empty).

---

**D2-1 — MAJOR (new, introduced by the D1-4 fix) — `AGENTS.md:420-421`**

> the conclusion was that **no portable numeric bound exists** for that site —
> CI's healthy 1056-1156 ms sat **above** the box's 1315 ms regression threshold.

1156 is not above 1315. The inequality is inverted, and it is checkable in one
line against the source it cites (`test_launch_subagent.py:2159-2161`):

> No portable numeric bound can sit **above CI's healthy 1156 ms and below this
> box's 1315 ms regression**.

The source's point is that the admissible window (1156, 1315) is too narrow to
survive run-to-run variance — healthy CI sits just *under* the regression
threshold, leaving no room. The doc's version says healthy CI sits *over* it,
which would make the conclusion trivial rather than interesting, and describes a
state that never happened. Where "sat above" does belong is the **950 ms bound**:
healthy CI (1056-1156 ms) exceeded it, which is the failure that started the
whole episode.

This matters more than an ordinary typo for two reasons: it is the doc's only
quantified argument for core-speed spread, and it is the *replacement* text for a
finding whose subject was misreading this exact docstring. A reader who follows
the doc's own instruction — "Read that docstring before you reach for a number" —
opens the file and finds the numbers the other way round.

Fix (one line): *"...CI's healthy 1056-1156 ms sat above the 950 ms bound and
just below this box's 1315 ms regression threshold, leaving no gap a portable
number could occupy."*

---

**D2-2 — MINOR (new) — `AGENTS.md:493-499`**

The rewritten 500 ms anecdote is accurate about #496 but stops one commit early,
and the tree it describes is not the tree on `main`. It reads as *raising to
500 ms was the error*, whereas the episode ended with **#498 (`0f9f9ed6`)
restoring 500 ms deliberately**, because `test_s1_boot` recorded 321 ms of
legitimate Textual-layout CPU on a green tree and `s2` recorded 413 ms — so
413 ms is the observed healthy max and 500 ms carries ~20% headroom (recorded in
`StallRecorder:74-80`). A reader who absorbs only this passage and then finds the
500 ms default in the tree has been handed a reason to lower it — the exact
change #498 reverted, twice.

The lesson the passage draws ("name what you just stopped catching") is right and
worth keeping; it just needs the ending. One clause: *"...— which is why #498
restored 500 ms only after CI produced a healthy 413 ms at that site, and said so
in the docstring."*

---

**D2-3 — MINOR (new) — `AGENTS.md:523-527`**

> no **in-process** probe reports anything, because the process that would
> report is the one that is stuck. That class is covered by `tests/e2e/watchdog.py`,
> whose `faulthandler.dump_traceback_later(exit=True)` ...

The sentence rules out in-process probes and then names an in-process one.
`dump_traceback_later` arms a timer *inside the same process*; what makes it
survive is that it runs on a dedicated **C** thread that writes with `write(2)`
and calls `_exit(1)` — no GIL, no Python frames (`watchdog.py:25-31`). The real
distinction is **Python-level vs C-level**, not out-of-process vs in-process, and
that is the part a reader needs: `watchdog.py:12-21` explains that
`asyncio.wait_for`, a `threading.Thread` watchdog, *and* `pytest-timeout`'s
signal method all fail for GIL/bytecode reasons. As worded, a reader could
reasonably infer that a thread watchdog is fine because it is "the same class of
thing" — the file says, measured, that it is not (a 20 s thread watchdog was
ignored and the process needed `kill -9`).

Fix: *"no Python-level bound fires — not `asyncio.wait_for`, not a thread
watchdog, not a signal-based timeout — because each needs the interpreter the
deadlock has stopped."*

---

**D2-4 — MINOR (carried from D1-6, unresolved and unanswered) — `AGENTS.md:377-378`**

The code span still breaks across the newline:

```
`tests/unit/harness/
test_comms.py`
```

CommonMark turns that line ending into a space, so it renders as
`tests/unit/harness/ test_comms.py` — not copy-pasteable, not greppable. These
remain the **only two lines in the whole file with an unbalanced backtick count**
(re-checked on `876ebf75`), so it is still a one-off against the document's own
convention. The remediation comment does not mention D1-6 at all, in either
direction — not fixed, not rejected, not deferred.

---

**D2-5 — NIT (carried from D1-10) — `AGENTS.md:507-508`**

"A single failure out of ~9000" — `pytest tests/unit --collect-only` on this head
reports **9891 tests collected**. In a section arguing for numeric precision,
`~9900` or "the ~10k unit tests" costs nothing.

---

**D2-6 — NIT (carried from D1-11) — `AGENTS.md:364-365`**

The incident list `(#122, #373, #461, #486, #496, #498)` still omits **#499**
(`e43f312f`), which is where
`test_store_maintenance_callbacks_run_off_the_event_loop_thread` — the worked
example the structural-invariant section is built on (`:405-412`) — came from.

---

### Is the advice now correct?

Yes, with D2-1 excepted. A developer following this section would write better
tests than one who did not: wait on a publication with the subscriptions wired
and released, prefer thread identity to a number, measure CPU when you must
measure, pick the dual-clock probe when a blocking sleep is in scope, calibrate
from multi-run CI, prove the guard can still go red, and diagnose before widening.
Nothing in the delta is self-contradictory — the D1-3 rewrite in particular now
agrees with "Prove the test can still fail" instead of fighting it, and the D1-5
warning points forward at the wall-backstop paragraph it needs to. The two new
subsections at the end genuinely close the gaps they were added for.

### CI on `876ebf75`

`lint`, `type-check`, `pip-audit`, `tui-e2e (ubuntu-latest)`, `tui-e2e
(macos-latest)` — all pass. `test (3.12)` still running at review time. Docs-only
(`AGENTS.md`, +176/-0), so no job here is load-bearing on content; confirm green
before merge per convention.

---

**Verdict: changes requested. Not mergeable at `876ebf75`.** 0 BLOCKER, 1 MAJOR,
3 MINOR, 2 NIT. The round-1 remediation is real work and lands five of five
MAJORs plus both omissions — the `ChangeSignal` example, the site/ceiling
attribution, the ratchet, the probe distinction, and the `test_launch_subagent`
citation are all now correct against the tree. But D2-1 is a checkably false
comparison sitting in the sentence that replaced the round-1 misreading, in the
one passage the doc tells readers to verify against the source, and this document
is about not being confidently wrong with numbers. It is a one-clause fix.
Round is **not terminal**; D2-2/D2-3 are cheap to take in the same commit, and
D2-4 has now gone one full round unanswered.

damianvtran:

Scope: e51894e5..876ebf75 reviewed; corrections pushed as `10fb074a`

All four findings fixed. D2-1 is the one that matters, and it is a fair hit: the round-1 fix introduced a fresh instance of exactly the error round 1 was about, in the single passage the doc instructs readers to verify against its source.

---

**D2-1 (MAJOR) — arithmetic backwards — FIXED.** Verified against `test_launch_subagent.py:2160-2161` before changing anything. The source says:

> No portable numeric bound can sit **above** CI's healthy 1156 ms and **below** this box's 1315 ms regression.

That is a 159 ms window too narrow to place a bound in — not "healthy sat above regression", which would be a different (and impossible) claim. "Sat above" belongs to the **950 ms bound** the CI runs blew through at 1056-1156 ms. Corrected to state the window.

**D2-2 (MINOR) — anecdote stopped one commit early — FIXED.** As written it argued for a change that was reverted and then restored, which would read as advice to make a mistake. It now carries why the raise was correct — those sites measured **321 ms and 413 ms of legitimate loop CPU on CI**, so a tighter bar failed green trees rather than catching regressions — and closes with the fuller rule: name what you stopped catching, *and* record the measurements that forced your hand.

**D2-3 (MINOR) — "no in-process probe" mis-framed — FIXED.** You are right that the sentence named an in-process probe as the answer to a claim that in-process probes cannot work. The real distinction is Python-level vs C-level, and `watchdog.py:25-29` states it exactly: `dump_traceback_later` arms a timer in a dedicated **C** thread that writes with `write(2)` and calls `_exit(1)` — no GIL, no Python frames, no interpreter state. The doc now says that, and explains why a coroutine probe cannot: it never gets scheduled either.

**D2-4 (NIT) — broken code span — FIXED.** `tests/unit/harness/test_comms.py` was split across a line break inside backticks and rendered broken. Joined onto one line.

---

### Verification

```
test_launch_subagent.py:2160-2161  "above ... 1156 and below ... 1315"   ✓ window, not ordering
watchdog.py:25-29                  dedicated C thread, write(2), _exit(1) ✓ matches new text
321 ms / 413 ms                    present in test_tui_responsiveness.py  ✓
no absolute home paths                                                    ✓
```

Delta is 25 insertions, 20 deletions, AGENTS.md only. Head moved `876ebf75` → `10fb074a`, so a round-3 pass is needed to re-pin freshness.

I note this is the second round where the remediation itself carried a new factual error. The pattern is mine to own: I have been writing claims from what I remembered of the incidents rather than re-reading the source each time, which is precisely the failure the "a number copied from an older comment is not evidence" bullet warns about. Every claim in this push was re-read from the file before it was written.

damianvtran:
### Agent review — round 3

Reviewer: independent `reviewer` subagent, anthropic/claude-opus-5
Scope: `876ebf75..10fb074a` (remediation delta), full diff `origin/main..dev-agents-flake-guidance`
Mode: Remediation convergence

D2-1 is genuinely fixed, and this time the remediation did **not** introduce a new
factual error — the first round of this PR where that is true. I re-derived every
changed claim from the cited source file rather than from the remediation comment,
which is what caught the reversed inequality last round. **0 BLOCKER, 0 MAJOR.**

Scope note: the real PR diff is `0f9f9ed6..10fb074a` = **`AGENTS.md` only, +181/-0**.
A literal `origin/main..head` shows 32 files because `origin/main` has moved ahead
(`59dce529`, `b328ea81`); nothing outside `AGENTS.md` belongs to this PR.

---

### Round-2 remediation status

| # | Finding | Status | Evidence re-derived from source |
|---|---|---|---|
| D2-1 | Arithmetic backwards (MAJOR) | **FIXED** | See below — verified against `test_launch_subagent.py:2159-2161`. |
| D2-2 | 500 ms anecdote stopped a commit early | **FIXED** | 321 ms / 413 ms correctly attributed; see below. |
| D2-3 | Python-level vs C-level framing | **FIXED, one residual gap** | See R3-1. |
| D2-4 | Broken code span | **FIXED** | `AGENTS.md:378` now carries `` `tests/unit/harness/test_comms.py` `` on one line. Backtick-balance sweep of the whole section (L340-545) returns only the six code-fence lines — no unbalanced spans remain. |
| D2-5 | `~9000` (NIT) | **NOT ADDRESSED, unanswered** | See R3-2. |
| D2-6 | `#499` missing from incident list (NIT) | **NOT ADDRESSED, unanswered** | See R3-3. |

The remediation comment opens *"All four findings fixed"* — round 2 raised **six**.
D2-5 and D2-6 are not mentioned in it in any direction: not fixed, not rejected, not
deferred. Both are NITs and neither blocks, but that is now the second consecutive
round where a carried finding went unanswered rather than being explicitly declined
(D1-6 did the same before becoming D2-4).

---

### D2-1 — verified fixed, faithfully

`AGENTS.md:420-423` now reads:

> the conclusion was that **no portable numeric bound exists** for that site — the
> only window left was between CI's healthy 1156 ms and this box's 1315 ms
> regression, which is too narrow to sit a bound in.

Source, `tests/unit/session/test_launch_subagent.py:2159-2161`:

> No portable numeric bound can sit above CI's healthy 1156 ms and below this
> box's 1315 ms regression.

Faithful. The doc now states the *window* (1156, 1315) rather than an ordering, which
is exactly what the source asserts, and `1156` is correctly the max of the healthy
`1056-1156` range at `:2157-2158` — the bound must clear the max, so quoting only
1156 is right, not a truncation.

On the reviewer-suggested attribution of "sat above" to the 950 ms bound: the author
did not take that phrasing and instead **dropped the 950 ms clause entirely**. That is
the better call, not a miss. It would have been correct — CI's healthy 1056-1156 ms
did exceed the 950 ms bound (`:2158`, *"against a 950 ms bound, on an unmodified
healthy tree"*) — but the paragraph's job is to justify abandoning numeric bounds, and
the window carries that on its own. No inaccuracy either way; the sentence no longer
asserts anything the source does not.

The "too narrow" gloss is the one word the source does not literally use. It is a fair
reading rather than an addition: `:2159` notes the same SHA's parallel run passed, and
`StallRecorder:79-80` records the same class of site swinging ~30% between runs, so a
159 ms window is demonstrably narrower than the observed variance. Accepted.

### D2-2 — verified fixed, figures correctly attributed

`AGENTS.md:499-501` attributes 321 ms and 413 ms to "the boot and turn sites". Both
check out at their own call sites, not just in the summary docstring:

- **s1-boot → 321 ms**: `test_tui_responsiveness.py:1072-1073`, *"boot recorded 321 ms
  of healthy Textual-layout CPU on CI"*. Corroborated by #498's commit body:
  `event loop consumed 321 ms of CPU without yielding (ceiling 200 ms)`.
- **s2 (turn) → 413 ms**: `:1167-1168`, *"legitimate loop CPU from Textual layout
  (413 ms recorded on CI)"*, inside
  `test_s2_send_to_agent_end_keeps_the_loop_under_the_bar` (`:1099`).
- The pairing matches the full dataset at `:77-78` — title-scan 206/264, s1-boot 321,
  s2 413 — so the two title-scan figures are correctly *not* claimed here.

"whose only guard is that bar" also holds: both sites' only stall assertion is
`recorder.assert_no_stall_loaded()` (`:1091`, `:1171`), with no structural
`get_ident` guard of the kind the title-scan site has — matching #496's
*"have no other guard"*.

The "made deliberately and on evidence" clause resolves the D2-2 complaint without
naming #498 explicitly. That is sufficient: the passage no longer reads as *raising to
500 was the error*, and a reader who finds 500 ms in the tree now has the reason it is
there. `500 > 413` also makes the surrounding "~20% headroom" framing internally
consistent with `StallRecorder:78-80`.

### D2-3 — core error fixed; see R3-1 for what remains

The out-of-process/in-process contradiction is gone. `AGENTS.md:528-532` now says a
**Python-level** probe reports nothing and that `dump_traceback_later` "is armed in C
and fires from a separate OS thread". That matches `watchdog.py:25-29` — *"arms a timer
in a dedicated **C** thread that writes the stacks of every thread with `write(2)` ...
then calls `_exit(1)` — no GIL, no Python frames, no interpreter state required"* — and
the "Python-level" framing is the source's own opening line (`:6-7`).

---

### R3-1 — MINOR (residual from D2-3) — `AGENTS.md:528-531`

> a Python-level probe reports nothing — the coroutine that would record the sample
> never gets scheduled either.

The category is now right; the **reason given covers only one of its three members**.
`watchdog.py:13-21` rules out three things by name and for three different reasons:
`asyncio.wait_for` (the loop is the thing that stopped), a `threading.Thread` watchdog
(needs a GIL a parked thread will never release — and even when it wakes,
`faulthandler.dump_traceback` raises `RuntimeError: file.fileno() is not a valid file
descriptor` under Textual), and `pytest-timeout`'s `signal` method (a Python signal
handler only runs between bytecodes).

The doc explains only the coroutine case. The thread watchdog is the most likely thing
a reader reaches for precisely *because* it is off the event loop, and "the coroutine
never gets scheduled" does not visibly rule it out. The file settles this with a
measurement the doc does not carry: a 20 s thread watchdog was ignored and the process
needed `kill -9` (`:23-24`).

Not a MAJOR: nothing stated is false, and the reader is pointed at the file. But the
un-refuted wrong answer is the tempting one.

Fix (one clause): *"...— and neither does a thread watchdog or a signal timeout, because
each needs the interpreter the deadlock has stopped."*

### R3-2 — NIT (carried from D2-5, unanswered) — `AGENTS.md:512`

"A single failure out of ~9000" — `pytest tests/unit --collect-only -q` on `10fb074a`
reports **9891 tests collected**. Re-ran on this head, not carried from the last comment.
`~9900` costs nothing in a section arguing for numeric precision.

### R3-3 — NIT (carried from D2-6, unanswered) — `AGENTS.md:365`

`(#122, #373, #461, #486, #496, #498)` still omits **#499**. Confirmed by
`git log -S`: `test_store_maintenance_callbacks_run_off_the_event_loop_thread` was
introduced by `e43f312f` (#499) and by nothing earlier — and that test is the entire
worked example of the structural-invariant section at `:405-412`. The list cites the
incident for every other technique in the section except its strongest one.

### R3-4 — NIT (new, from transplanting source wording) — `AGENTS.md:422`

"this box's 1315 ms regression" is lifted verbatim from the docstring, where "this box"
unambiguously means the machine that produced the measurement. In `AGENTS.md` the
nearest referent is the reader's own machine, which is not what 1315 ms was measured on.
"the reporting dev box's 1315 ms" removes the ambiguity.

---

### Re-read of the whole section as a developer

Correct and internally consistent. Following it end to end now produces good tests and,
importantly, no longer produces a *wrong* one:

- The `ChangeSignal` example (`:383-388`) is runnable and matches `test_comms.py:1706`
  and its `finally: signal.close()`; the "a bare `ChangeSignal()` subscribes to nothing"
  sentence pre-empts the 30 s misdiagnosis that shipped in round 1.
- Structural-before-numeric (`:404-424`) now ends where its source ends — no portable
  bound, converted to a spy — instead of citing that proof as support for picking a number.
- The calibrate-from-CI section (`:459-484`) and the 500 ms anecdote (`:497-505`) now
  agree: 206/264/321/413 is quoted once as a dataset and twice per-site, with the same
  numbers both times.
- "Find out which it is before touching the number" (`:479-484`) and "Prove the test can
  still fail" (`:486-505`) still pull the same direction, which is what the D1-3 rewrite
  was for.
- The two closing subsections are accurate on re-check: no `pytest-timeout` in
  `pyproject.toml`; `timeout-minutes: 40` on `test` and the 3h38m anecdote both at
  `ci.yml:146-149`.

The one place a reader is under-served is R3-1. Everything else is a wording or
completeness preference, not an error.

Conventions: no absolute home paths in the delta (`grep '/Users/\|/home/'` on added
lines → 0). Docs-only, no code or test change, so nothing to run beyond the collection
count above.

### CI on `10fb074a`

`lint`, `type-check`, `pip-audit`, `tui-e2e (ubuntu-latest)`, `tui-e2e (macos-latest)`
— **pass**. `test (3.12)` — **pending** at review time (run 33428644688). Docs-only, so
no job is load-bearing on content; confirm it goes green before merging, per convention.

---

**Verdict: clean. Mergeable at `10fb074a` once `test (3.12)` reports green.**
0 BLOCKER, 0 MAJOR, 1 MINOR, 3 NIT.

This round is **TERMINAL**. R3-1 through R3-4 are follow-ups, not grounds for a round 4
— none of them makes a false claim or misdirects a reader who follows the section, and
R3-1 is a one-clause addition that can ride along with any future edit to this file or
be taken now at the author's discretion. The two-round pattern of remediations
introducing fresh factual errors did **not** repeat here: I re-derived 1156/1315, 321,
413, the `write(2)`/`_exit(1)` description, and the code-span rendering from the source
files directly, and all four are faithful. No further review round is required.

damianvtran:

Scope: 876ebf75..10fb074a reviewed; follow-ups pushed as `97062031`

All three fixed. Round 3 was clean, so none of these gated — but two of them were findings my round-2 remediation silently dropped, which is its own defect and worth correcting properly.

---

**R3-1 (MINOR) — thread watchdog left un-refuted — FIXED.** Fair catch: the note explained why a coroutine probe fails and then named the answer, without addressing the fallback a reader is most likely to reach for first. `watchdog.py:13-24` rules it out with a measurement, so the doc now carries it: a `threading.Thread` watchdog needs a GIL the wedged thread never releases, and `pytest-timeout`'s default `signal` method never fires because a Python signal handler only runs between bytecodes. The pre-fix code ignored a 20 s thread watchdog and had to be `kill -9`'d.

That also closes a loose end in the adjacent paragraph, which mentions `pytest-timeout`'s absence without saying that adding it would not have helped for this failure class.

**R3-2 (NIT) — `~9000` understated — FIXED.** `pytest --collect-only` reports **9891/9894 collected**. Now `~9900`.

**R3-3 (NIT) — #499 missing — FIXED.** It owns the section's worked example (the structural thread-identity test), so its absence from the incident list was the odd one. Added.

---

**On the dropped findings.** You are right that round 2 raised six and my remediation said "all four findings fixed", disposing of neither R3-2's predecessor nor R3-3's. That is exactly the gap the gate's "every finding needs a disposition" rule exists to close, and the honest reading is that I dispositioned the ones I had already fixed and did not re-read the list. Both are now fixed rather than deferred.

### Verification

```
pytest --collect-only            9891/9894 collected      ✓ (~9900)
watchdog.py:13-24                thread watchdog + signal ruled out, measured  ✓
#499                             owns the structural example                   ✓
no absolute home paths                                                          ✓
```

Delta is 10 insertions, 4 deletions, AGENTS.md only. Head moved `10fb074a` → `97062031`; a round-4 pass follows to re-pin freshness.

damianvtran:
### Agent review — round 4

Reviewer: independent `reviewer` subagent, anthropic/claude-opus-5
Scope: `10fb074a..97062031` (remediation delta), full diff `origin/main..dev-agents-flake-guidance`
Mode: Remediation convergence

Narrow freshness re-pin. I did not re-review content settled in rounds 1-3; I verified only the delta, and I verified each changed claim against its source file rather than against the remediation comment.

---

### Delta scope — confirmed

`git diff --stat 10fb074a 97062031` → `AGENTS.md | 14 ++++----`, **10 insertions, 4 deletions**, one file. Exactly the three advertised fixes and nothing else. Full PR remains docs-only: `origin/main...97062031` is `AGENTS.md | 187 +++`, zero code or test files.

### R3-1 — thread watchdog refuted — verified, faithful on all three clauses

`AGENTS.md:530-534` adds three claims. All three are in `tests/e2e/watchdog.py`, at the cited lines, and each is stated no more strongly than its source:

| Doc claim | Source |
|---|---|
| `threading.Thread` watchdog "needs a GIL the wedged thread never releases" | `watchdog.py:14-16` — *"needs the GIL released by a thread that will never return to release it deliberately"* |
| `pytest-timeout`'s default `signal` method "never fires because a Python signal handler only runs between bytecodes" | `watchdog.py:19-20` — verbatim, including the `signal`-method qualifier |
| "The pre-fix code ignored a 20 s thread watchdog and had to be `kill -9`'d" | `watchdog.py:22-24` — *"Measured, not assumed: driving the pre-fix code through `/resume` produced a process that ignored a 20 s thread watchdog and had to be `kill -9`'d"* |

Two details worth crediting rather than flagging. The doc says "documents, **with a measurement**" — that is the source's own "Measured, not assumed" framing, so the epistemic status is carried across, not upgraded. And the `pytest-timeout` clause is correctly scoped to the **default `signal` method**; dropping that qualifier would have been a fresh error, since the source only rules out that method. The thread-watchdog bullet at `:14-18` also carries a second failure mode (interpreter-level `dump_traceback` raising under Textual) that the doc omits — omission, not misstatement, and the GIL reason is the one that generalizes.

The paragraph split at `:535-536` ("That class is covered by...") is a consequence of the insertion and reads correctly; the antecedent of "That class" is still the C-level-freeze class named at `:527-529`.

The claimed side benefit holds too: `:520` mentions `pytest-timeout`'s absence, and the new text now says adding it would not have helped this failure class. Those two passages no longer sit in tension.

### R3-2 — collected count — verified

```
$ .venv/bin/python -m pytest --collect-only -q -p no:randomly
9891/9894 tests collected (3 deselected) in 2.71s
```

`~9900` at `AGENTS.md:512` is right, and correct in the direction that matters — the sentence is about how small one failure is against the total, so rounding up rather than down is the conservative choice. `~9000` was a 10% understatement.

### R3-3 — #499 — verified, and it does own the example

`AGENTS.md:365` now reads `(#122, #373, #461, #486, #496, #498, #499)`. Ownership checked at the commit level, not just by title: `git log -S test_store_maintenance_callbacks_run_off_the_event_loop_thread` returns `e43f312f test(tui): pin store maintenance to worker threads (#499)`, and that test name is the one cited at `AGENTS.md:407-409`, living at `tests/unit/test_tui_responsiveness.py:289`. #499 is MERGED and its body states the same rationale the section gives (*"Healthy and regressed CPU measurements overlap... Thread identity directly detects the failure mode"*). So the section's worked example and the incident list now agree, which was the point of the nit.

### No fresh inaccuracy — checked deliberately

Rounds 1 and 2 both introduced new errors while fixing errors, so I re-derived every changed claim from source. **The pattern did not repeat.** Five claims changed; five check out at the cited files. Nothing in the delta over-claims relative to its source, and the one thing the source says that the doc drops (the Textual `fileno()` failure mode) is an omission from an already-secondary detail.

Conventions: no absolute home paths in added lines (`/Users/`, `/home/` → 0 hits).

### CI on `97062031`

`lint`, `type-check`, `pip-audit`, `tui-e2e (ubuntu-latest)`, `tui-e2e (macos-latest)` — **pass**.

`test (3.12)` — **FAILURE**, one test:

```
FAILED tests/unit/session/test_resume_subagents_todos.py::test_live_continuation_preserves_prior_accounting_across_restart
  - AssertionError: timed out waiting for condition
```

Classified per this PR's own rule at `AGENTS.md:511-515` rather than waved through:

- **Not caused by this PR.** The branch is docs-only against `origin/main` — 187 added lines in `AGENTS.md`, no code, no tests, no config. There is no mechanism by which it changes a runtime assertion.
- **Timing-sensitive, and a different test each run.** The failure is `wait_for`'s 5 s deadline (`test_resume_subagents_todos.py:65-71`), not a value assertion. `main` failed independently around the same window (run `33427261323`) on a *different* timing test, `test_steering_approval.py::test_a_held_key_never_answers_a_question_it_was_not_meant_for`. Distinct tests failing per run under load is the doc's own definition of a flake, not a regression.
- **Does not reproduce locally.** 15/15 passes, 0.95-1.42 s each against a 5 s bound — roughly 3.5x headroom, consistent with contention on the CI runner rather than a real stall.

So: a known-class flake, unrelated to this change, and I checked the log rather than assuming. It is still a **red head**, and the same section says not to merge one — the honest disposition is to **re-run `test (3.12)` and merge on green**, which costs nothing on a docs-only branch. I am not treating the red as a review finding against the delta.

---

**Verdict: clean. 0 BLOCKER, 0 MAJOR, 0 MINOR, 0 NIT.**

The delta is exactly the three advertised fixes, all three are faithful to their sources, and no new inaccuracy was introduced. R3-1's residual is fully closed — the doc now refutes the two fallbacks a reader reaches for first, with the source's own measurement attached.

This round is **TERMINAL**. There is nothing outstanding from any round: R3-1, R3-2 and R3-3 are all fixed and verified, and I raise no new findings. No further review round is required.

**Mergeable at `97062031` once `test (3.12)` is re-run green.** The only thing standing between this PR and merge is an unrelated infrastructure flake, not its content.

PR #509, opened 2026-08-31
fix(tui): settings selection frame + provider/model suggestion dropdowns

damianvtran:
# PR Description

## Summary of Changes

Fixes two reported UX defects in the `/settings` full-page view:

1. **Clicking a setting painted a text-selection frame.** Textual 8.x arms a screen text-selection on the `MouseDown` of any widget whose `allow_select` is true (`Screen._forward_event`), so a click meant only to move the page cursor also began a drag whose grey highlight bled across the rows and their values. The settings page is chrome, not selectable content. A local `_ChromeStatic(ALLOW_SELECT = False)` — the same remedy `app.Chrome` applies to the status band and composer chevron — now backs every rendered surface (list, title, rule, detail, side pane), keeping them out of Textual's selection walk while leaving the row-select-on-click path and the cursor's own reverse ink untouched.

2. **Default provider and Default model were bare free-text fields.** Editing them offered no suggestions — a poor experience for two fields whose valid values are a known provider registry and a large model catalogue. They now open an **inline suggestion dropdown with ghost-text completion**, reusing the exact machinery behind `/model` and the slash-command composer: models are ranked by `model_picker.rank_rows`, providers by `autocomplete.score_command_text_match`, and the dimmed inline ghost by `command_picker.ghost_for`. Arrow keys browse, Tab / right-at-end / click complete, Enter accepts-then-saves, and Esc gains a rung in the existing ladder (close suggestions, then the editor). Because suggestions only assist, a custom value the catalogue does not know (a private endpoint id) still saves as typed.

Structural choice: an inline dropdown of unselectable `suggest` rows under the edited setting, reusing existing ranking/ghost helpers rather than introducing a new `Kind` or `choices_source`. `hosting`/`model_name` stay `Kind.TEXT`; the accepted value matches what each key stores (bare provider id / bare model id). The catalogues are fed from the same `/provider` login registry and `/model` catalogue via two new app-side resolvers, degrading to the plain editor when unavailable. `_ChromeStatic` is defined locally rather than imported from `app.Chrome` because `app` imports `settings_view` — importing back would close a cycle.

### Real rendered frames (SVG exported via `run_test` + `save_screenshot`, viewed as PNG)

**Issue 1 — selection frame**

- Before: a grey drag-selection band bleeds across the rows and into the values.
- After: only the `›` row cursor is shown; no selection frame. Row-select-on-click still works.

**Issue 2 — provider suggestions**

- Typing `a` on Default provider shows ghost `nthropic` inline and a filtered suggestion list (`anthropic`, `alibaba`, `openai`, `mistral`, `radient`) with provider display names.

**Issue 2 — model suggestions**

- Typing `opus` on Default model shows the fuzzy-ranked catalogue (`anthropic/claude-opus-5 · anthropic · $15/75`, `openrouter/anthropic/claude-opus-5 · …`), mirroring `/model`'s columns and current-model marker, with ghost-text completion of the top row.

**Issue 2 — custom value**

- A value the catalogue does not know matches nothing, the dropdown closes, and the value saves as typed via the unchanged commit path.

(Rendered frames attached in the review thread.)

### Regression coverage

```text
    tests/unit/tui/test_settings_view.py -q
516 passed in 95.12s

$ env -u NO_COLOR TERM=xterm-256color .venv/bin/python -m pytest tests/unit/tui -q
4118 passed, 12 skipped
```

17 new tests cover: click/drag arms no text-selection; every surface reports `allow_select=False`; provider and model suggestions filter, ghost, and accept via arrow/Tab; a custom (non-catalogue) model value still saves; the Esc ladder stays correct; and the no-catalogue fallback to a plain editor.

### Gates (whole tree, exactly as CI)

```text
flake8 .            → 0
black --check .      → 691 files unchanged (26.1.0)
isort --check .      → clean (5.13.2)
pyright .            → 0 errors, 0 warnings
```

## Versioning

Patch bump to `0.44.16`: a bug fix plus a small, self-contained UX improvement to two existing fields — no step-function capability. Rebased onto current `origin/main`.

### DIFF
--- a/local_operator/tui/app.py
+++ b/local_operator/tui/app.py
@@ -11279,6 +11279,8 @@ class OperatorApp(App[None]):
             teams=self._settings_team_rows(),
             agents=self._agent_profile_rows(),
             providers=self._settings_provider_rows(),
+            provider_catalogue=self._settings_provider_catalogue(),
+            model_catalogue=self._settings_model_catalogue(),
         )
 
     def _settings_team_rows(self) -> list[tuple[str, str, str]]:
@@ -11327,6 +11329,53 @@ class OperatorApp(App[None]):
         except Exception:  # noqa: BLE001 — never crash the page on a store read
             logger.debug("settings: provider rows unavailable", exc_info=True)
         return rows
+
+    def _settings_provider_catalogue(self) -> list[tuple[str, str]]:
+        """Loginable providers as ``(id, display_name)`` for the Default provider
+        suggestion dropdown.
+
+        The WHOLE login registry, not just the signed-in providers
+        ``_settings_provider_rows`` returns: the Default provider field is where
+        a user names the provider they intend to boot on, which may be one they
+        have not logged into yet, so constraining the suggestions to current
+        credentials would hide exactly the choice they are reaching for. Same
+        registry ``/login`` and ``/provider`` enumerate, in registry order, so
+        the page cannot offer a provider those surfaces do not. Never raises: an
+        unreadable registry degrades to an empty catalogue, which falls the field
+        back to a plain free-text editor.
+        """
+        rows: list[tuple[str, str]] = []
+        providers = self._providers
+        if providers is None:
+            return rows
+        try:
+            for definition in providers.login_providers():
+                rows.append((definition.id, definition.name))
+        except Exception:  # noqa: BLE001 — the page must open without a catalogue
+            logger.debug("settings: provider catalogue unavailable", exc_info=True)
+        return rows
+
+    def _settings_model_catalogue(self) -> list[ModelRow]:
+        """The model catalogue for the Default model suggestion dropdown.
+
+        The SAME rows ``/model`` shows, through the SAME ``_catalogue_rows``
+        filter (usable providers, the current model rescued in), so the two
+        surfaces rank an identical set by the identical matcher and cannot drift.
+        The static registry only — no live provider fetch — because this runs
+        synchronously on the page-open path and a network round-trip there would
+        stall the mode; ``/model``'s own live refresh already covers the rows
+        that shipped too late for the registry, and the Default model field is a
+        boot preference a user can also just type. Degrades to an empty
+        catalogue (plain free-text editor) on any failure.
+        """
+        try:
+            rows, _note = self._catalogue_rows(
+                self._providers.static_catalogue() if self._providers else []
+            )
+            return rows
+        except Exception:  # noqa: BLE001 — the page must open without a catalogue
+            logger.debug("settings: model catalogue unavailable", exc_info=True)
+            return []
 
     def _close_settings_view(self) -> bool:
         """Leave the settings mode and put the conversation back. True if open.
--- a/local_operator/tui/widgets/settings_view.py
+++ b/local_operator/tui/widgets/settings_view.py
@@ -487,6 +487,13 @@ class SettingsView(Vertical):
         self._enter_hint = HintButton("enter", lambda: self.action_activate())
         self._reset_hint = HintButton("r", lambda: self.action_reset())
         self._pane_hint = HintButton("←→", lambda: self.action_pane(1))
+        # Shown ONLY while a suggestion dropdown is live, in place of `r`/`←→`
+        # (which do nothing useful there). It advertises tab-completion — the
+        # accelerator that was otherwise undiscoverable below ~140 cols, since
+        # the inline row contract sheds it first (review round 1, D1/U1).
+        # Clicking it completes the highlighted suggestion, the same action Tab
+        # takes, so the footer stays a real affordance for a mouse user.
+        self._tab_hint = HintButton("tab", self._complete_highlighted_suggestion)
         self._exit_hint = HintButton("esc", self._leave)
         self._hints = Horizontal(classes="settings-view-hints")
         self._title_text = Text()
@@ -512,6 +519,12 @@ class SettingsView(Vertical):
         yield self._detail
         with self._hints:
             yield self._move_hint
+            # Composed right after `↑↓` so the dropdown footer reads
+            # `↑↓ suggestions · tab complete · enter save · esc` in that order —
+            # footer hints paint in DOM order, so `leads` ordering alone can't
+            # place it. Hidden (`display = False`) in every non-dropdown state,
+            # so its compose position is invisible there.
+            yield self._tab_hint
             yield self._enter_hint
             yield self._reset_hint
             yield self._pane_hint
@@ -2000,16 +2013,33 @@ class SettingsView(Vertical):
                 self._cancel_edit()
                 self._repaint()
                 return
-            if key in ("up", "down") and has_suggestions:
-                # The dropdown owns the arrows while it is showing — there is no
-                # cursor to move on the page (the editor holds it), so the arrows
-                # have nothing else to do and browsing the list is what a user
-                # reaches for. Wraps, like the model picker's own move.
+            if key in ("up", "down", "ctrl+p", "ctrl+n") and has_suggestions:
+                # The dropdown owns the arrows AND the readline pair while it is
+                # showing — there is no cursor to move on the page (the editor
+                # holds it), so browsing the list is what a user reaches for.
+                # `ctrl+p`/`ctrl+n` move the highlight exactly as `/model`'s
+                # picker binds them (review round 1, U2): a user who learned that
+                # surface must not press ctrl+n mid-edit and be thrown to another
+                # section with their typing gone. Wraps, like the model picker's
+                # own `move`.
                 event.stop()
                 event.prevent_default()
                 count = len(self._suggestions())
-                delta = -1 if key == "up" else 1
+                delta = -1 if key in ("up", "ctrl+p") else 1
                 self._suggest_index = (self._suggest_index + delta) % max(count, 1)
+                self._repaint()
+                return
+            if key in ("pageup", "pagedown") and has_suggestions:
+                # Page the highlight by a windowful, CLAMPED (not wrapping),
+                # mirroring `ModelPicker.page` — a PgDn that silently returned to
+                # the top of the list would read as the list resetting itself.
+                # The step is the visible window so one press moves a screenful
+                # of suggestions, matching `/model`.
+                event.stop()
+                event.prevent_default()
+                count = len(self._suggestions())
+                step = _SUGGEST_ROWS * (1 if key == "pagedown" else -1)
+                self._suggest_index = max(0, min(count - 1, self._suggest_index + step))
                 self._repaint()
                 return
             if key == "tab" and has_suggestions:
@@ -2035,8 +2065,23 @@ class SettingsView(Vertical):
                 # value the catalogue does not match has no highlight, so Enter
                 # saves the typed text directly — the "suggestions assist, never
                 # constrain" contract.
+                #
+                # GATED ON A NON-EMPTY BUFFER (review round 1, BLOCKER). An empty
+                # buffer still yields a highlight — `_suggestions()` returns the
+                # WHOLE catalogue when nothing is typed, so row 0 (`anthropic`)
+                # is "highlighted" over content the user never chose. Accepting
+                # it would write that provider instead of honouring the
+                # `empty_unsets` clear-to-unset gesture: clearing `hosting` and
+                # pressing Enter must UNSET it, not silently store the top row
+                # (the U1 blocker `_commit_edit` fixes and this guard must not
+                # re-open). So an empty buffer falls straight through to
+                # `_commit_edit`, whose `empty_unsets` branch does the unset.
                 suggestion = self._highlighted_suggestion()
-                if suggestion is not None and self._buffer.strip() != suggestion.value:
+                if (
+                    suggestion is not None
+                    and self._buffer.strip()
+                    and self._buffer.strip() != suggestion.value
+                ):
                     self._accept_suggestion(suggestion)
                 self._commit_edit()
                 return
@@ -2086,6 +2131,23 @@ class SettingsView(Vertical):
                     # and left the user 25 rows from where they were.
                     self._caret = 0 if key == "home" else len(self._buffer)
                     self._repaint()
+                return
+            if key in ("pageup", "pagedown", "ctrl+p", "ctrl+n"):
+                # SWALLOWED as a no-op whenever no dropdown is live to page —
+                # the dropdown branches above already consumed these when it was
+                # (review round 1, U2). An OPEN EDITOR must own every navigation
+                # key the same way it owns the printable ones (this method's own
+                # docstring), and these leaked to the page bindings: mid-edit
+                # they cancelled the editor, DISCARDED the typed value and
+                # teleported the cursor to another section — the exact leak
+                # `left/right/home/end` were intercepted to stop, simply omitted
+                # from that fix. Inert here rather than paging is correct: with
+                # no dropdown there is nothing to page, and losing the edit is
+                # never the right answer to a scroll gesture. Covers both the
+                # suggest editor with the list dismissed/empty and every other
+                # text editor on the page (`retry.maxRetries`, an endpoint).
+                event.stop()
+                event.prevent_default()
                 return
             char = getattr(event, "character", None)
             if char and char.isprintable():
@@ -2184,14 +2246,20 @@ class SettingsView(Vertical):
         event.stop()
         row = self._rows[index]
         if row.kind == "suggest" and row.suggestion is not None:
-            # A click on a dropdown row highlights it and accepts it in one
-            # gesture — completing the buffer to that value, exactly as the model
-            # picker's own click chooses a row. It does NOT save: the page's
-            # contract is that only Enter on the chosen value writes (#440), so
-            # a stray click completes and leaves the confirming Enter to the
-            # user, matching what Tab on the same row does.
+            # A click on a dropdown row completes the buffer to its value, then
+            # a SECOND click on that same already-completed row SAVES — the
+            # click-once-select / click-again-activate rhythm the setting rows
+            # themselves use just below, extended so a mouse user has a path to
+            # the write (review round 1, U4). Without it, clicking a suggestion
+            # filled the buffer and then dead-ended: a second click re-completed
+            # to no effect and only the keyboard's Enter could save. The commit
+            # still goes through the one Enter path, so `empty_unsets` and
+            # validation behave identically to pressing Enter.
+            already = self._suggest_index == row.hop_index and self._buffer == row.suggestion.value
             self._suggest_index = row.hop_index
             self._accept_suggestion(row.suggestion)
+            if already:
+                self._commit_edit()
             return
         if not row.selectable:
             return
@@ -2489,6 +2557,16 @@ class SettingsView(Vertical):
             line.append(row.suggestion.label, style=fg if highlighted else muted)
             if row.suggestion.detail:
                 line.append(f"  {row.suggestion.detail}", style=faint)
+            # A `pos/total` cue on the HIGHLIGHTED row, only when the list is
+            # windowed (more matches than fit), so a user arrowing a long
+            # catalogue can tell where they are and how many remain — the
+            # position/count the `/model` picker carries in its footer (review
+            # round 1, U3). Only on the highlight and only when it means
+            # something (windowed), so it never clutters a short list that shows
+            # in full. `hop_index` is 0-based; +1 reads as a human position.
+            total = len(self._suggestions())
+            if highlighted and total > _SUGGEST_ROWS:
+                line.append(f"  {row.hop_index + 1}/{total}", style=dim)
             return line
 
         marker = f"{_CURSOR} " if selected else "  "
@@ -2657,18 +2735,26 @@ class SettingsView(Vertical):
         # the clause advertised a behaviour the row does not have.
         contract = "  enter saves · esc cancels"
         setting = settings_io.resolve_key(self._editing or "")
-        # When a suggestion dropdown is live the contract names the keys that
-        # drive it, or the affordance is undiscoverable: the footer hint contract
-        # has to stay honest (a lit affordance with no advertised key is the "why
-        # doesn't anything happen" complaint U5 records). Preferred over
-        # `clear to unset` because a user staring at a list of choices needs
-        # "how do I pick one" more than "how do I empty this", and both do not
-        # fit; `_suggest_rows` being non-empty is exactly "a dropdown is shown".
+        # When a suggestion dropdown is live the contract names the browse and
+        # complete keys, SHORTENED so it survives the value column at 100 cols
+        # (review round 1, D1/U1). The old long form (`↑↓ suggestions · tab
+        # completes · enter saves · esc`, ~52 cells) only fit at ~140 cols and
+        # was shed to `enter saves ·…` at every common width, so the two new
+        # accelerators were advertised nowhere. `↑↓ · tab · enter saves` is
+        # ~22 cells and fits the ~23-cell room at 100 cols; the footer now
+        # carries the fuller `↑↓ suggestions · tab complete · enter save`
+        # legend, so the row hint can be terse without losing discoverability.
+        # A widest-first ladder keeps the fuller inline form where it fits.
         if self._suggest_rows():
-            full = "  ↑↓ suggestions · tab completes · enter saves · esc"
             room = self._list_width() - _VALUE_COLUMN - cell_len(editor.plain)
-            if cell_len(full) <= room:
-                contract = full
+            for candidate in (
+                "  ↑↓ suggestions · tab completes · enter saves · esc",
+                "  ↑↓ browse · tab complete · enter saves",
+                "  ↑↓ · tab · enter saves",
+            ):
+                if cell_len(candidate) <= room:
+                    contract = candidate
+                    break
         elif setting is not None and setting.empty_unsets:
             full = "  enter saves · esc cancels · clear to unset"
             room = self._list_width() - _VALUE_COLUMN - cell_len(editor.plain)
@@ -2867,6 +2953,20 @@ class SettingsView(Vertical):
         self._suggest_index = 0
         self._suggest_dismissed = False
         self._repaint()
+
+    def _complete_highlighted_suggestion(self) -> None:
+        """Footer `tab` hint's click action: complete the highlighted row.
+
+        The mouse counterpart to pressing Tab — a `None`-returning wrapper
+        because ``HintButton`` takes a ``Callable[[], None]`` and
+        ``_accept_suggestion`` already returns None, but routing the click here
+        (rather than binding the method directly) keeps the guard in one place:
+        a stray click on the hint when no suggestion is highlighted must do
+        nothing rather than raise.
+        """
+        suggestion = self._highlighted_suggestion()
+        if suggestion is not None:
+            self._accept_suggestion(suggestion)
 
     def _reopen_suggestions(self) -> None:
         """Un-dismiss the dropdown and reset the highlight — for buffer edits.
@@ -3733,6 +3833,15 @@ class SettingsView(Vertical):
         enter: _Hint = (self._enter_hint, " change", True)
         reset: _Hint = (self._reset_hint, " default", True)
         pane: _Hint = (self._pane_hint, " panes", True)
+        tab: _Hint = (self._tab_hint, " complete", True)
+        # Is a suggestion dropdown live right now? The footer's editing state
+        # splits on this: a plain text editor keeps `↑↓ move cancel`, but with a
+        # dropdown open ↑↓ browses the list and `tab` completes, so the footer
+        # must say THOSE (review round 1, D1/D2/U1). The always-visible hint is
+        # what a user scans; leaving it reading `move cancel`/`panes` while the
+        # keys browse suggestions and move the caret is the footer-vs-behaviour
+        # disagreement D7 exists to forbid.
+        dropdown_open = bool(self._suggest_rows())
 
         # The footer states what the keys do RIGHT NOW, per state, because it
         # is the row users scan for exactly that. Two states rewrite it:
@@ -3767,8 +3876,16 @@ class SettingsView(Vertical):
         if self._expanded is not None:
             enter = (self._enter_hint, " choose", True)
         if self._editing is not None:
-            move = (self._move_hint, " move cancel", False)
-            enter = (self._enter_hint, " save", True)
+            if dropdown_open:
+                # ↑↓ browses the suggestions and `tab` completes — the real keys
+                # for this state, named where they are always legible instead of
+                # only in the row contract the value column sheds first. `enter`
+                # saves the highlighted/typed value.
+                move = (self._move_hint, " suggestions", False)
+                enter = (self._enter_hint, " save", True)
+            else:
+                move = (self._move_hint, " move cancel", False)
+                enter = (self._enter_hint, " save", True)
 
         def rung(
             leads: list[tuple[HintButton, str, bool]], esc_label: str
@@ -3789,7 +3906,15 @@ class SettingsView(Vertical):
         # change · r default` on a row that honours neither (UX round 1, U2).
         # The detail line already says WHY the row is retired; the footer's job
         # is only to stop advertising keys that will not act.
-        if self._current_is_readonly():
+        if dropdown_open:
+            # The dropdown state offers its own three keys and NOT `r`/`←→`:
+            # `r default` types `r` into the buffer (the #425 anti-pattern
+            # `_reset_applies` already guards) and `←→` moves the caret, not the
+            # pane. Ordered `↑↓ suggestions · tab complete · enter save` so the
+            # browse key leads, the accelerator sits beside it, and the commit
+            # closes the row — the same reading order as the inline contract.
+            leads = [move, tab, enter]
+        elif self._current_is_readonly():
             leads = [move]
         elif not self._reset_applies():
             # The SAME rule, one step finer (#440 §4.4). `r` deletes the stored
@@ -3801,7 +3926,10 @@ class SettingsView(Vertical):
             # a default row as soon as the terminal was narrow enough to shed
             # the pane hint, which is the regression invariant 3 warns about.
             leads = [move, enter]
-        if self._pane_fits():
+        if self._pane_fits() and not dropdown_open:
+            # `←→` moves the caret while an editor is open, so the pane hint is
+            # withheld in the dropdown state where the footer names the caret's
+            # real keys instead (it was already meaningless there).
             leads.append(pane)
         # The narrow rungs shed to a one-word `esc` label, except in the
         # confirm and choosing states where `cancel` IS the one word and
@@ -3840,6 +3968,7 @@ class SettingsView(Vertical):
             self._enter_hint,
             self._reset_hint,
             self._pane_hint,
+            self._tab_hint,
             self._exit_hint,
         ):
             hint.display = hint in visible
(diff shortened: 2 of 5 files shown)


damianvtran:
### Design review — round 1

Scope: 5c8658f6..1d56a810

Reviewed the rendered frames in `evidence/*.svg.png` (issue1_before/after,
issue2_provider[-empty], issue2_model[-empty], issue2_custom, and the
`*_before` frames), plus two frames I generated to cover the states the
evidence set omitted: a no-match buffer (`provider` state, buffer `zzz`) and a
wide 160-col terminal. Findings back the pixels with the source that produces
them (`settings_view.py`).

**Issue 1 — selection frame: fixed, TERMINAL.** `issue1_before` shows the grey
text-selection band bleeding across ~7 rows; `issue1_after` shows a clean `›`
caret + bold label on the selected row with no band at all. The cursor stays
fully legible and the reserved marker column keeps labels from shifting. The
`_ChromeStatic(ALLOW_SELECT = False)` approach is the right fix and there is no
residual artifact.

**Issue 2 — suggestion dropdowns: coherent, with one real discoverability
defect at common widths.** The dropdown reads as part of the settings page:
suggestion rows carry the page's own `›` caret and indent under the edited
setting, column alignment of the dim detail note (`anthropic · $15/75`,
`Anthropic`) is consistent, ghost-text dimming matches the slash-command
composer, and the custom-value frame correctly shows no ghost and a committable
buffer. Fuzzy matches (`opus`) correctly suppress the inline ghost and carry the
completion on the row instead — the append-only honesty rule is respected.

---

**D1 — MAJOR — the suggestion-key contract truncates at default/common terminal
widths, so `tab completes` and `↑↓ suggestions` are never advertised where the
feature lives.** Every populated evidence frame is captured at 100×30, and in
all of them the inline contract renders as `enter saves ·…` (provider) /
`enter saves · esc …` (model) — truncated with an ellipsis. The full contract
`↑↓ suggestions · tab completes · enter saves · esc` only appears in my 160-col
frame. The cause is in `_suggest_rows` contract assembly
(`settings_view.py:2667-2671`): the long string is used only
`if cell_len(full) <= room`, where `room = _list_width() - _VALUE_COLUMN -
cell_len(editor.plain)`. At 100 cols `room ≈ 23` cells while the full contract
is ~52, so it falls back to the short `enter saves · esc cancels` (~27), which
the value column then *also* truncates to `enter saves ·…`. The footer
(`↑↓ move cancel · enter save · ←→ panes · esc cancel`) never names `tab` or
suggestions either. Net effect: at 80–110 col terminals — the common case, not
an edge — the two new affordances this PR exists to add (tab-completion,
arrow-browsing the list) have **no visible hint anywhere**. The core save path
survives (`enter saves` still shows), which is why this is MAJOR and not a
blocker, but a designed accelerator that is invisible at the width most users
run is a real defect. Options: shorten the full contract so it fits ~23 cells
(e.g. `↑↓ · tab · enter saves`), or move the key legend into the footer where
it has room, or place it on the dropdown header row rather than the value
column.

**D2 — MINOR — the footer's `↑↓ move cancel` label is inaccurate while a
dropdown is open.** The always-visible footer says the arrows move/cancel the
edit. With a live dropdown, `on_key` (`settings_view.py:2003-2014`) makes ↑↓
browse the suggestion list and explicitly does *not* cancel — the opposite of
what the footer promises. The label is honest only in the no-dropdown editing
state; in the dropdown states (the feature's primary states) it mislabels the
keys. This compounds D1: the one always-visible hint actively contradicts the
new behaviour, and the inline contract that would correct it is the one being
truncated. The actual behaviour is better than promised (arrows do the useful
thing), so this is MINOR, but it is the same footer-vs-detail disagreement the
code comments elsewhere treat as a bug (D7/U4 lineage). Consider a dropdown-
state footer rung that says `↑↓ suggestions` instead of `move cancel`.

**D3 — NIT — caret glyph differs from the `/model` picker, but correctly.** The
dropdown uses `›` (`_CURSOR`, `settings_view.py:174`) while `/model`,
command-picker and session-picker standardised on `❯`. Because the dropdown is
embedded *inside* the settings surface, matching its host's `›` (the setting
rows use it too) is the right call — within-surface consistency beats
consistency with the borrowed data source. Recording it only so the divergence
is a documented choice, not an oversight. No change requested.

**D4 — NIT — ghost text and the inline contract sit close in similar dimness.**
In `issue2_provider` the row reads `a‹ghost›nthropic  enter saves ·…`, with the
dim ghost and the faint contract separated only by the contract's 2-space lead.
They are distinguishable but tonally similar; if D1 is addressed by keeping a
key legend in the value column, widen the gap or differentiate the tint so the
completion preview and the key hint don't blur.

---

**States covered.** loading (n/a — synchronous page), empty-query
(`*-empty` frames, full ordered list, clean), populated (provider/model),
custom/no-match (`issue2_custom` + my `zzz` frame: dropdown collapses to the
plain editor with no empty box — the overflow/degenerate case is handled well),
and overflow (`openrouter/anthropic/claude-opus-5` row truncates its price
gracefully). **Empty-catalogue:** no rendered frame, but `_suggest_key`
(`settings_view.py:2735-2738`) returns `None` when the injected catalogue is
empty, so the row falls back to a plain free-text editor rather than an empty
dropdown — correct behaviour; a frame would be nice-to-have, not required.

**Verdict: NOT TERMINAL.** One MAJOR (D1) remains: the new keys are
undiscoverable at common terminal widths. D2 (MINOR) should be worked in the
same pass since it shares the root cause. D3/D4 are follow-ups / no-change.
Re-capture a 100-col populated frame after D1 so the fix is proven at the width
users actually run.

damianvtran:
### Agent review — round 1

Scope: 5c8658f6..1d56a810

Reviewed the full diff (settings_view.py +541, app.py +49, test_settings_view.py +357, scripts/settings_suggest_shot.py, pyproject.toml), the reused contracts (`rank_rows`, `ghost_for`, `score_command_text_match`, `format_price_pair`, `ModelRow`), and the rendered evidence frames. Gates run from the worktree: **flake8 clean, black 26.1.0 clean on the 4 changed files, isort 5.13.2 clean, pyright 0 errors on changed files, and `pytest tests/unit/tui/test_settings_view.py` = 516 passed.** (Whole-tree `black --check .` flags one file — `scripts/ux_walk.py` — but it is untracked, absent from origin/main and not part of this PR; not a finding here.)

The ALLOW_SELECT approach is correct: `_ChromeStatic(ALLOW_SELECT=False)` keeps the surfaces out of `Screen._forward_event`'s selection walk while click delivery (and thus row-select) is untouched; reproduced the two halves myself (cursor moves, `screen.selections` stays empty on click and on drag). Import-cycle avoidance is sound (the three reused modules do not import settings_view/app). Value semantics are correct — a model suggestion stores the **bare** `model_id` (matching what `_persist_default_from_picker` writes at app.py:12688), with the `provider/id` selector only in the display label. The reused matcher contracts are called correctly.

But the accept-then-commit path has a real defect that breaks a shipped contract.

---

**BLOCKER — clearing `hosting`/`model_name` to unset instead silently writes the top catalogue row; reintroduces the U1 blocker.**

`hosting` and `model_name` are `empty_unsets=True` (settings_io.py:352,362) — clearing the field and pressing Enter is the documented "clear to unset" gesture, and `_commit_edit` handles it at settings_view.py:1735. But the new Enter branch runs **before** `_commit_edit`:

```
settings_view.py:2038  suggestion = self._highlighted_suggestion()
settings_view.py:2039  if suggestion is not None and self._buffer.strip() != suggestion.value:
settings_view.py:2040      self._accept_suggestion(suggestion)   # buffer <- "anthropic"
settings_view.py:2041  self._commit_edit()                       # saves "anthropic"
```

On an empty buffer, `_suggestions()` returns the whole catalogue and `_highlighted_suggestion()` returns index 0 (`anthropic`). `"".strip() != "anthropic"` is true, so Enter **accepts the top row and saves it** instead of unsetting. Reproduced against real `OperatorApp` (catalogue injected on open exactly as production does at app.py:11278):

- Open unset `hosting`, press Enter → stored `hosting: anthropic` (should stay unset). This is exactly the `test_enter_twice_on_an_unset_row_writes_nothing` (U1) blocker, back again.
- `hosting=openai`, clear it, Enter → stored `hosting: **anthropic**` — the user's clear-to-unset silently *changes* the provider to a third value they never chose.
- Same for `model_name`: `model_name=gpt-5.2`, clear, Enter → stored `model_name: claude-opus-5`.

The existing U1 regression test stays green only because its `FakeSession` injects no catalogue, so the dropdown never opens in that test — the guard has a blind spot the real page falls into. This is a data-correctness defect (writes a boot default the user did not pick, and defeats unset) that ships the moment the catalogue is present, i.e. always.

Fix direction (one line): don't auto-accept when the buffer is empty — e.g. gate the accept on `self._buffer.strip()` being non-empty (`if suggestion is not None and self._buffer.strip() and self._buffer.strip() != suggestion.value:`), so an empty field falls through to `_commit_edit`'s `empty_unsets` branch. A regression test should drive this through the real app **with** the catalogue injected (the U1 test's exact gap).

---

**MINOR — the bottom footer reads "↑↓ move cancel" while the dropdown is open, but ↑↓ then browses suggestions and does not cancel.**

`_paint_hints` sets the move hint to `" move cancel"` whenever `_editing is not None` (settings_view.py:3770), with no awareness of the live dropdown. In the suggest state (see `evidence/issue2_provider`), ↑↓ navigates the suggestion list and never cancels the edit — so the footer advertises a behaviour the keys no longer have while suggestions show. This is the same footer-vs-behaviour disagreement the surrounding comments (D7/D2) were written to avoid. The new inline editor contract ("↑↓ suggestions · tab completes · enter saves · esc") is correct; it's the persistent bottom hint row that's now stale for this state. Correctness-preserving, hence minor. (The general awkwardness of "move cancel" as a phrase is pre-existing wording from #440 and unchanged by this PR — not a defect introduced here.)

---

**NIT — comment/behaviour drift on the Enter path.** The comment at settings_view.py:2033-2037 says "A custom value the catalogue does not match has no highlight, so Enter saves the typed text directly." That's true for a non-matching value, but the comment doesn't cover the empty-buffer case where there *is* a highlight over content the user did not choose — which is exactly where the BLOCKER lives. Worth aligning the comment with the fix.

---

Verdict: **changes requested.** One BLOCKER (unset regression / U1 reintroduction) must be fixed before this can merge; the MINOR footer-state wording and the NIT are follow-ups that can ride the same remediation. Test scaffolding is otherwise good — the 17 new tests drive the real `OperatorApp` through `pilot.press`/mouse events and assert stored config, not internals; they simply miss the empty-buffer/unset interaction, which is the gap the BLOCKER slipped through.

damianvtran:
### UX review — round 1

Reviewer: ux-reviewer (lopdev team) · anthropic/claude-opus-4
Scope: 5c8658f6..1d56a810

Walked the real flow end-to-end by driving `OperatorApp` under `run_test` (per AGENTS.md "Visual validation"), injecting the provider + model catalogues through `view.load()` exactly as the app does, plus the rendered evidence frames. Exercised: activating the provider and model rows, typing to filter, ghost completion, Tab/right-at-end completion, ↑↓ browsing, Enter accept-and-save, the Esc ladder, a custom non-catalogue value, clicking a dropdown row, the empty-catalogue fallback, and the sibling-key checks (`d`, left/right, pageup/pagedown, ctrl+n/ctrl+p, wheel) against an open dropdown.

**What works and I confirmed by execution.** Typing filters both fields; the ghost completes the top append-honest match after the caret and correctly disappears for a fuzzy/rewriting match and mid-string caret. Tab completes the highlighted row into the buffer without saving (`_buffer == "claude-opus-5"`, `editing_key` unchanged), matching `/model`'s `_complete_model`. Enter on a live dropdown accepts-then-saves in one press (arrow to `openrouter`, Enter → `hosting == "openrouter"`, editor closed). Right-at-end accepts the ghost. The Esc ladder is correct in isolation: first Esc closes the dropdown (`_suggest_dismissed`, editor still open, buffer intact), a second cancels the editor; typing re-opens a dismissed list. A custom value (`my-endpoint/custom-v2`) offers no rows, no ghost, and saves as typed. Empty catalogue falls back to the plain editor with no empty dropdown. `d` typed into the model field is eaten by the buffer (not delete-hop); left/right/home/end move the caret, not the pane. ↑↓ browse the list and wrap, matching `/model`. Good coverage of the core contract.

Two MAJOR findings block a clean round; both are consistency/discoverability defects, not correctness of the happy path.

---

**U1 — MAJOR — The dropdown's keys are undiscoverable at the default and common terminal widths, and the footer actively misdescribes them.**

Step: open the Default provider or Default model editor at 80, 100, or 120 columns (100×30 is the shot script's own default and a very common size; 80/100/140 are the sizes the test suite parametrizes) and look at a live dropdown of matches.

The only place the dropdown keys are ever advertised is the row-level editor contract, and it only fits at **≥140 columns**. Measured `render_lines_for_test()` on the provider editor with buffer `a`:

| width | editor row contract |
|---|---|
| 90  | `enter saves · esc cancels` |
| 100 | `enter saves ·…` (clipped) |
| 110 | `enter saves · esc cance…` (clipped) |
| 120 | `enter saves · esc cancels` |
| 130 | `enter saves · esc cancels` |
| 140 | `↑↓ suggestions · tab completes · enter saves · esc` |

So at every width a user is likely to run, `↑↓ suggestions · tab completes` is shed or clipped away and never appears. Meanwhile the persistent footer while editing reads `↑↓  move cancel · enter  save · ←→  panes · esc cancel` (verified via `rendered_hints()` at 100w and visible in the `issue2_model-empty` frame, which shows a 7-row dropdown under exactly this footer). With a live dropdown:
- `↑↓` browses suggestions — the footer says it means **"move cancel"**.
- `←→` moves the caret — the footer says it means **"panes"**.

The result: a user staring at seven ranked model rows has (a) no on-screen cue that arrows browse them or that Tab completes, and (b) a footer that tells them the arrows do something else entirely. Discoverability (judge 1) and copy/consistency (judges 5–6) both fail, and this is the feature's core affordance. The ghost text partially rescues completion, but arrow-browsing is genuinely undiscoverable below 140 columns.

Expected: the dropdown-key advertisement should survive at 80–120 columns — e.g. make the footer hints reflect the editing+dropdown state (`↑↓ suggestions · tab completes` instead of `move cancel`/`panes` while a dropdown is live), rather than relying on a row contract that only the widest terminals render. Note: the `←→ panes` mislabel while editing is pre-existing (left/right have moved the caret since before this PR), so that half is a deferral; the newly-dishonest `↑↓ move cancel` and the shed-away row contract are introduced by this slice.

---

**U2 — MAJOR — Navigation keys leak past the open dropdown, discard the typed value, and teleport the cursor — the exact class of bug the on_key docstring claims to have closed, and inconsistent with `/model`.**

Step: open either editor, type a filter, then press a key a `/model` user reaches for to scroll a long list — PageDown, PageUp, or Ctrl-N.

The editor's `on_key` intercepts `left/right/home/end` (the U2/U3 fix) but **not** `pageup`, `pagedown`, `ctrl+n`, or `ctrl+p`. They fall through to the page bindings. Verified with a 12-item provider catalogue and buffer `prov`:

- **PageDown** → `editing_key` becomes `None`, cursor jumps to `retry.baseDelayMs` (a different section), notice: `discarded — enter saves, esc or moving away cancels`.
- **PageUp** → editor cancelled, value discarded.
- **Ctrl-N** (buffer `open`) → cursor lands on `providers.openai.api`, editor cancelled, discarded.
- **Ctrl-P** from a mid-page row (`model_name`, buffer `opus`) → editor *stays open* (the clamped move happened to no-op).

That last asymmetry makes it worse: Ctrl-N discards your edit but Ctrl-P sometimes doesn't, so the same "move the highlight" gesture behaves two different ways. In the `/model` picker these keys drive the list — `pageup/pagedown` page the suggestions, `ctrl+n/ctrl+p` move the highlight — so a user who learned that model (the PR's stated "don't learn two" goal) presses PageDown to reach match #9 of a windowed list and is instead thrown two sections away with their typing gone. This directly contradicts this file's own `on_key` docstring: "an OPEN EDITOR must own every printable key… Left/right… were the keys a user reaches for mid-edit and the ones that leaked." Page and readline-nav keys are the same case, simply omitted from the same intercept.

Expected: while an editor is open, `pageup/pagedown/ctrl+n/ctrl+p` should be owned by the editor — either page/move the dropdown highlight (best, mirrors `/model`) or at minimum be swallowed as a no-op — never exit the editor and discard the buffer. Kept at MAJOR rather than BLOCKER because a "discarded" notice does surface and no wrong value is written; it is not silent data-corruption, but it is lost input on a key press that should have been safe.

---

**U3 — MINOR — A windowed dropdown gives no position/count indicator.**

Step: open the provider editor with an empty buffer against a 12-item catalogue. `_suggest_rows()` windows to 8 rows (`_SUGGEST_ROWS = 8`) and the window shifts as you arrow, but there is no "showing 8 of 12" / scroll cue. The `/model` picker the PR mirrors carries a count/status footer (`_footer_rows`). A user arrowing a long list can't tell how many more are below or where they are. Low impact (the lists are short and arrows reveal the rest); noted for parity with the sibling surface.

---

**U4 — MINOR — Clicking a suggestion completes but cannot save; there is no mouse-only path to accept from the dropdown.**

Step: with the dropdown open, click a suggestion row. `on_click` calls `_accept_suggestion` (fills the buffer, caret to end) and returns — it never saves, by the #440 "only Enter writes" contract. A second click on the same row re-completes (no-op). So a mouse user who clicks `openai` sees the buffer fill and then must reach for the keyboard's Enter to actually save; clicking again does nothing. In `/model`, a row click *acts* (`choose` → switch). The select-then-confirm choice is defensible (it mirrors the settings page's own click-once-select / click-again-activate on setting rows), but there's a mild dead-end: no gesture completes the save with the mouse alone, and nothing on screen tells the mouse user that Enter is the missing step. Consider making a click on an *already-highlighted* suggestion save (second-click activates), mirroring the page's own row behaviour.

---

**Verdict: NOT TERMINAL.** No BLOCKER, but two MAJOR findings remain (U1 undiscoverable/mislabelled dropdown keys at common widths; U2 nav-keys leak and discard the edit). Both are consistency-with-`/model` and discoverability defects on the feature's core interaction, and both are the same class the PR's own `on_key` docstring says it closed. U3 and U4 are follow-ups. Re-review after U1 and U2 are addressed on the current head.

No issues filed (team rule); U3/U4 and the pre-existing `←→ panes` mislabel are recorded here as the deferral record.

damianvtran:

Scope: 5c8658f6..1d56a810
Remediation head: e98552b7 (fix `15fe3352`, shot-script `e98552b7`)

Per-finding disposition:

- **BLOCKER — empty-buffer Enter wrote catalogue row 0 instead of unsetting.**
  FIXED in `15fe3352`. The Enter accept is now gated on a non-empty buffer
  (`self._buffer.strip()` truthy), so clearing `hosting`/`model_name` and
  pressing Enter falls through to `_commit_edit`'s `empty_unsets` unset branch
  instead of accepting row 0. Verified against the real page with the catalogue
  injected (the blind spot): seed `hosting=openai`, clear, Enter → key unset
  (absent from config.yml), editor closed. Regression test
  `test_clearing_a_suggest_field_unsets_it_rather_than_writing_row_zero`
  parametrizes both keys and drives the real `OperatorApp` with the catalogue
  present.

- **MINOR — footer reads "↑↓ move cancel" while the dropdown is open.**
  FIXED in `15fe3352` (worked with D1/U1 below, same root cause). The footer's
  editing state now splits on whether a dropdown is live: with one, it reads
  `↑↓ suggestions · tab complete · enter save · esc` and withholds `r`/`←→`
  (which type into the buffer / move the caret there); a plain editor keeps the
  pre-existing `↑↓ move cancel`. Tests:
  `test_the_footer_names_the_dropdown_keys_at_a_common_width` and
  `test_the_footer_keeps_move_cancel_for_a_plain_editor`.

- **NIT — comment/behaviour drift on the Enter path.**
  FIXED in `15fe3352`. The Enter-path comment now states the empty-buffer case
  explicitly (row 0 is highlighted over content the user never chose; the guard
  must not re-open the U1 unset contract), matching the new behaviour.

No findings rejected. The pre-existing `←→ panes` mislabel (pre-dates this PR)
is deferred — it is out of scope for this slice and recorded in the UX round's
deferral list.

damianvtran:

Scope: 5c8658f6..1d56a810
Remediation head: e98552b7 (fix `15fe3352`, shot-script `e98552b7`)

Per-finding disposition:

- **D1 — MAJOR — suggestion-key contract truncates at common widths; tab/arrow
  never advertised.** FIXED in `15fe3352`. The key legend now lives in the
  always-visible footer, which reads `↑↓ suggestions · tab complete · enter
  save · esc cancel` at 100 cols (verified in the re-captured 100-col frames
  `evidence/issue2_provider.svg.png`, `issue2_model-empty.svg.png`). The inline
  row contract is additionally width-laddered (`↑↓ suggestions · tab completes
  · enter saves · esc` → `↑↓ browse · tab complete · enter saves` →
  `↑↓ · tab · enter saves`) so the terse form survives the value column at 100
  cols. A new clickable `tab` footer hint completes the highlighted suggestion,
  keeping the footer a real affordance.

- **D2 — MINOR — footer `↑↓ move cancel` inaccurate while a dropdown is open.**
  FIXED in `15fe3352`, same pass as D1 (shared root cause): the dropdown-state
  footer names the real keys and withholds `r`/`←→`.

- **D3 — NIT — caret glyph `›` vs `/model`'s `❯`.** NO CHANGE, as the review
  itself concludes: within-surface consistency with the settings rows' `›` is
  the right call. Recorded as a documented choice.

- **D4 — NIT — ghost and inline contract tonally similar.** Addressed as part
  of D1: the inline contract now ladders to shorter forms at narrow widths, so
  at 100 cols the row shows the ghost plus the terse `enter saves ·…` while the
  FULL legend moved to the footer, widening the separation between the
  completion preview and the key hint. No separate tint change needed.

Re-captured 100-col populated frames prove the keys are discoverable at the
width users run: `evidence/issue2_provider.svg.png` and
`evidence/issue2_model-empty.svg.png` both show the full footer legend beside a
live dropdown; `evidence/issue2_provider_windowed.svg.png` adds the windowed
count cue.

damianvtran:

Reviewer: ux-reviewer (lopdev team) · anthropic/claude-opus-4
Scope: 5c8658f6..1d56a810
Remediation head: e98552b7 (fix `15fe3352`, shot-script `e98552b7`)

Per-finding disposition:

- **U1 — MAJOR — dropdown keys undiscoverable at common widths; footer
  mislabels them.** FIXED in `15fe3352`. The footer now reflects the
  editing+dropdown state: with a live dropdown it reads `↑↓ suggestions · tab
  complete · enter save · esc cancel` at 80–120 cols (verified at 100×30 in
  `evidence/issue2_provider.svg.png`), and the newly-dishonest `↑↓ move cancel`
  is gone in that state. The inline contract ladders to a form that fits 100
  cols. Tests assert the footer names `suggestions`/`complete`/`save` and does
  NOT say `move cancel`/`panes` while a dropdown is live, and that a plain
  editor keeps `move cancel`.

- **U2 — MAJOR — pageup/pagedown/ctrl+n/ctrl+p leak past the open editor,
  discard the value, teleport the cursor.** FIXED in `15fe3352`, mirroring
  `/model`. With a dropdown live: `ctrl+n`/`ctrl+p` move the highlight (like
  `/model`), `pageup`/`pagedown` page it by a windowful, clamped (not wrapping).
  With no dropdown (suggest editor dismissed/empty, or any plain text editor
  like `retry.maxRetries`): the keys are swallowed as a no-op — never a discard.
  Tests: `test_nav_keys_do_not_leak_past_an_open_suggest_editor` and
  `test_nav_keys_do_not_leak_past_a_plain_text_editor`, parametrized over all
  four keys, asserting the editor survives, the buffer is intact, and nothing
  is written.

- **U3 — MINOR — no position/count indicator on a windowed dropdown.** FIXED in
  `15fe3352`. The highlighted row carries `pos/total` when the list exceeds the
  8-row window (e.g. `4/12`), matching `/model`'s count; hidden on short lists
  so they stay uncluttered. Test: `test_a_windowed_dropdown_shows_a_position_count`.
  Frame: `evidence/issue2_provider_windowed.svg.png` (reproducible via
  `scripts/settings_suggest_shot.py OUT.svg 100x30 provider-windowed`).

- **U4 — MINOR — clicking a suggestion completes but cannot save; no mouse-only
  path.** FIXED in `15fe3352`. A second click on the already-completed,
  highlighted suggestion saves it — the click-once-select / click-again-activate
  rhythm the setting rows use. The commit routes through the one Enter path, so
  `empty_unsets`/validation behave identically. Test:
  `test_a_second_click_on_a_suggestion_saves_it`.

- **Pre-existing `←→ panes` mislabel while editing (pre-dates this PR).**
  DEFERRED — out of scope for this slice; left/right have moved the caret since
  before this PR. Recorded here as the deferral record; the dropdown-state
  footer (U1) already withholds `←→ panes` where it is wrong, so the new
  surface does not repeat it.

damianvtran:
### Design review — round 2

Reviewer: designer (anthropic/claude-opus-4-8)
Scope: 5c8658f6..e98552b7

Re-reviewed the full diff `origin/main..HEAD` (base `5c8658f6`, head `e98552b7`)
against my round-1 D-findings. Verified the current head by re-rendering the
provider-windowed state at **80, 100, and 120 cols** with the shot script
(`env -u NO_COLOR TERM=xterm-256color .venv/bin/python scripts/settings_suggest_shot.py
OUT.svg <W>x30 provider-windowed`) and reading the exact painted strings out of
each SVG, plus the committed `evidence/issue2_provider_windowed.svg.png`.

**D1 — MAJOR — key contract invisible at common widths — RESOLVED.** The key
legend now lives in the always-visible footer, and it renders in full at every
width I checked, not just 160 cols:

- 80 cols: `↑↓ suggestions · tab complete · enter save · esc cancel`
- 100 cols: `↑↓ suggestions · tab complete · enter save · esc cancel`
- 120 cols: `↑↓ suggestions · tab complete · enter save · esc cancel`

Both new affordances this PR exists to add — `tab` completion and `↑↓`
suggestion browsing — are now advertised where the feature lives, at the widths
users actually run. The inline value-column contract additionally width-ladders
(`↑↓ suggestions · tab completes · enter saves · esc` → `↑↓ browse · tab
complete · enter saves` → `↑↓ · tab · enter saves`); at 80/120 the terse
`↑↓ · tab · enter saves` fits, and at 100 the value column truncates it to
`enter saves · es…`, but that is now a *redundant secondary* hint — the footer
carries the authoritative, complete legend regardless. My round-1 defect was
"no visible hint anywhere at common widths"; that is gone.

**D2 — MINOR — footer `↑↓ move cancel` contradicted the dropdown — RESOLVED.**
With a live dropdown the footer no longer claims the arrows move/cancel the
edit; it reads `↑↓ suggestions · tab complete · enter save · esc cancel`, which
matches the actual key behaviour (↑↓ navigates the list, tab completes, enter
saves, esc cancels). The one always-visible hint is now honest for the
feature's primary states.

**U3 count cue — present and legible.** The highlighted row (`prov03`) carries
`4/12` in the reserved detail column when the catalogue exceeds the visible
window, giving position/total without disturbing column alignment. Reads
cleanly at all three widths.

**Dropdown coherence — intact.** The suggestion list still reads as part of the
settings page: `›` caret on the highlighted row, indented rows under the edited
setting, dim `Provider NN` detail notes column-aligned, ghost text (`prov03`)
in the value column matching composer dimness. No new visual defect from the
remediation — the footer split and the count cue are additive and column-safe.

**D3/D4 (round-1 NITs)** — no change was required and none introduced a
regression.

---

**Verdict: TERMINAL.** No BLOCKER, no MAJOR. Both of my round-1 findings (D1
MAJOR, D2 MINOR) are genuinely fixed on the current head `e98552b7`, verified at
80/100/120 cols against re-rendered frames, and the remediation introduced no
new design defect. Round is clean. No new findings; nothing deferred from the
design surface. (No issues filed — team rule.)

damianvtran:
### Agent review — round 2

Scope: 5c8658f6..e98552b7

Re-reviewed the full diff `origin/main..HEAD` and, specifically, the round-1
remediation delta `1d56a810..e98552b7` (settings_view.py +188, shot script +22,
test_settings_view.py +248; app.py untouched). Gates run fresh from the worktree
`/tmp/lop-settings-ux`:

- **flake8** clean on the 4 changed files.
- **black 26.1.0** — `4 files would be left unchanged`.
- **isort 5.13.2** — clean.
- **pyright** (venv interpreter) — `0 errors, 0 warnings` on the changed files.
- **`env -u NO_COLOR TERM=xterm-256color .venv/bin/python -m pytest tests/unit/tui/test_settings_view.py -q`** — **530 passed** (210s). The 13 remediation-specific tests pass in isolation too.

**Round-1 BLOCKER — empty-buffer Enter wrote catalogue row 0 instead of
unsetting (U1 reintroduction): FIXED, verified.** The Enter-accept at
`settings_view.py:2079-2086` is now gated on a non-empty buffer:

```
if (suggestion is not None
        and self._buffer.strip()
        and self._buffer.strip() != suggestion.value):
    self._accept_suggestion(suggestion)
self._commit_edit()
```

On an empty buffer the middle clause is falsy, so the accept is skipped and
control falls straight through to `_commit_edit`, whose `empty_unsets` branch
does the unset — the original correct contract. Reproduced the round-1 repro
through the parametrized regression test
`test_clearing_a_suggest_field_unsets_it_rather_than_writing_row_zero`, which
drives the real `OperatorApp` **with the catalogue injected** (the exact blind
spot the old U1 test missed): seeds `hosting=openai`/`model_name=gpt-5.2`, opens
the editor, backspaces the buffer empty, asserts `_suggest_rows()` is live
(row 0 highlighted — the trap), presses Enter, and asserts `editing_key is None`
and the key is absent/empty in the stored config. Both keys pass. (b), (c)
satisfied.

**MINOR — footer "↑↓ move cancel" wrong while dropdown open: FIXED.**
`_paint_hints` now branches on `dropdown_open = bool(self._suggest_rows())`
(`settings_view.py:3876+`): with a live dropdown the footer reads
`↑↓ suggestions · tab complete · enter save · esc` and withholds `r`/`←→`
(which type into the buffer / move the caret, not the pane); a plain editor
keeps the pre-existing `↑↓ move cancel`. Covered by
`test_the_footer_names_the_dropdown_keys_at_a_common_width` (100 cols) and
`test_the_footer_keeps_move_cancel_for_a_plain_editor`, both green. (d)
satisfied.

**NIT — Enter-path comment drift: FIXED.** The comment at
`settings_view.py:2068-2078` now states the empty-buffer case explicitly (row 0
highlighted over content the user never chose; the guard must not re-open the
U1 unset contract), matching the new behaviour. (d) satisfied.

**No new defect introduced (f), verified.** The remediation also swept in the
UX-round U2/U4 work, which I checked for regressions:

- `pageup`/`pagedown`/`ctrl+p`/`ctrl+n` with a live dropdown page/move the
  highlight and are consumed; the page step is **clamped**
  (`max(0, min(count - 1, idx + step))`, line 2042) — no wrap, no index out of
  range. With **no** dropdown these four keys are now swallowed as inert no-ops
  (line 2104+) so they can no longer leak to the page bindings and discard an
  open editor's typed value. Covered by
  `test_nav_keys_do_not_leak_past_an_open_suggest_editor` and the plain-editor
  variant, both green.
- Second-click-saves: `already` is computed before `_suggest_index` is
  reassigned and the commit still routes through the single Enter/`_commit_edit`
  path, so `empty_unsets`/validation are unchanged; the empty-buffer guard
  above still governs it. Covered and green.
- The new `_complete_highlighted_suggestion` footer-`tab` handler reuses
  `_accept_suggestion` (no save), consistent with keyboard Tab.

Editor state machine otherwise unchanged; app.py is untouched by the
remediation.

---

**Verdict: clean. This round is TERMINAL.** The round-1 BLOCKER is genuinely
fixed on the current head (e98552b7), the MINOR and NIT are resolved, a
regression test now covers the catalogue-injected empty-buffer-unset path that
was the original blind spot, all gates are green, and the remediation
introduced no new defect. No blocker or major finding remains; nothing further
to raise.

damianvtran:
### UX review — round 2

Reviewer: ux-reviewer (lopdev team) · anthropic/claude-opus-4
Scope: 5c8658f6..e98552b7

Re-walked the real flow end to end by driving `OperatorApp` under `run_test`
with **real key events** (`pilot.press`, editor opened the production way via
`_select` + Enter so focus is set — not by mutating `_buffer`), catalogues
injected through `view.load()` exactly as the app does. 25/25 interaction
checks pass. Rendered frames captured at 100×30 for the two visual claims and
inspected (footer legend; windowed `4/12` cue).

**Every round-1 finding is fixed on the current head.**

- **U1 — MAJOR — dropdown keys undiscoverable at common widths + footer
  actively wrong.** RESOLVED. With a live dropdown the footer now reads
  `↑↓ suggestions · tab complete · enter save · esc cancel` at **every** width
  in the band I re-measured — 80, 90, 100, 110, 120 all render the full legend
  (previously the keys only appeared at ≥140 and the footer said the dishonest
  `↑↓ move cancel`). Confirmed in the painted frame, not just the string. A
  plain text editor (`retry.maxRetries`) correctly keeps `↑↓ move cancel`, so
  the split is state-aware.

- **U2 — MAJOR — pageup/pagedown/ctrl+n/ctrl+p leaked past the open editor and
  discarded the edit.** RESOLVED. Verified on all four keys across three
  states: (a) live dropdown — the editor SURVIVES and nothing is written
  (`pagedown` pages the highlight by a windowful and clamps at the bottom with
  no wrap: idx 0→8, and from the last row it stays at 11/12; `ctrl+n` moves the
  highlight down); (b) plain text editor — swallowed as a no-op, buffer `"7"`
  intact, cursor does not teleport; (c) dropdown dismissed via Esc — swallowed,
  buffer intact. The round-1 harm (edit discarded, cursor thrown to another
  section) no longer occurs on any of the four keys.

  One consistency note I chased to ground so it is on the record, not a hidden
  surprise: **`ctrl+p` opens Textual's App-level command palette**
  (`COMMAND_PALETTE_BINDING = "ctrl+p"`, a priority binding) before `on_key`
  ever sees it, so it does not move the highlight *up*. I verified this is
  **identical on the sibling `/model` picker** — pressing `ctrl+p` there pushes
  the same `CommandPalette` screen — so settings is consistent with the surface
  the PR mirrors, and crucially the edit is **not** discarded (the editor stays
  open). `ctrl+n`, `up`, `down`, `pageup`, `pagedown` all drive the list. Not a
  finding: the U2 harm (lost input) is gone and the behaviour matches its
  sibling. Reclaiming `ctrl+p` from the palette would be an app-wide change well
  outside this slice.

- **U3 — MINOR — no position/count indicator on a windowed dropdown.**
  RESOLVED. A `pos/total` cue rides the highlighted row when the list exceeds
  the 8-row window — verified `4/12` on the highlight in the painted frame, and
  it moves with the highlight. Correctly **absent** on a short (8-item) list, so
  it never clutters a list that shows in full.

- **U4 — MINOR — mouse-only user could not save from the dropdown.** RESOLVED.
  Driving the genuine click-coordinate path: first click on a suggestion
  selects+completes and writes nothing (`hosting` still `anthropic`); a second
  click on that same already-completed row saves (`hosting` → `openai`, editor
  closed). The click routes through the one Enter commit path, so
  `empty_unsets`/validation are identical to pressing Enter.

**Other flow points re-confirmed clean on this head:**

- **Empty-buffer Enter unsets (the code-review BLOCKER).** Clearing `hosting`
  (seeded `openai`) and pressing Enter leaves it **unset**, not silently
  rewritten to the top catalogue row `anthropic`; same for `model_name` (never
  becomes `claude-opus-5`). The remediation's non-empty-buffer gate holds
  against the real app with the catalogue injected — the exact gap the round-1
  blocker slipped through.
- **Enter on a live dropdown** accepts the highlighted row and saves in one
  press (arrow to `openrouter`, Enter → stored `openrouter`, editor closed).
- **Esc ladder** intact: 1st Esc closes the dropdown (editor open, buffer `"a"`
  preserved), 2nd cancels the editor with no write; typing re-opens a dismissed
  list.
- **Tab** completes the highlighted row into the buffer without saving (buffer
  `openai`, editor open, `hosting` unchanged) — the #440 "only Enter writes"
  contract.

Committed suggestion tests (`-k "suggest or nav_keys or windowed or
second_click or footer or unset or esc_ladder or custom"`) = 23 passed.

**No new interaction defect** introduced by the remediation.

One follow-up, recorded here as the deferral record (no ticket, team rule):

- **U5 — NIT — at ≤60 cols the dropdown footer sheds the `enter save` label**
  (renders `↑↓ suggestions · tab complete · esc cancel`). This is the
  pre-existing whole-hint shedding ladder doing its job — `esc` is never shed
  and Enter still saves — so no key is unreachable; the label just isn't named
  at the narrowest widths, below the 80–120 band U1 was about. Deferred — cosmetic,
  outside this slice, and the same ladder governs every footer on the page.

**Verdict: TERMINAL.** No BLOCKER and no MAJOR remains; both round-1 MAJORs and
both MINORs are fixed and re-verified against the current head e98552b7, and the
remediation introduced no regression. The round is clean from a UX standpoint.

PR #511, opened 2026-08-31
feat(evaluation): isolate benchmark adapters behind a verified process boundary

damianvtran:
## Summary

This PR adds the fourth slice of Local Operator's benchmark-neutral evaluation platform: the out-of-process **adapter boundary**. It lets the harness run a pinned, third-party benchmark adapter in a separate interpreter under an exact, content-addressed contract, and it does not activate a benchmark, OSWorld, provider, CLI, configuration, tool, cloud, or UI path.

The boundary covers five concerns:

- **Exact distribution and entry-point selection.** An `AdapterSelector` pins the distribution name, version, entry point, package digest, release digest, absolute interpreter, and workspace. There are no ranges and no fallbacks.
- **Verified-source import.** Every module the selected distribution owns is served from bytes hash-matched against its wheel `RECORD` and compiled in-process, so the bytes that execute are the bytes that were hashed.
- **Dedicated-FD canonical JSONL RPC.** Two inherited one-way pipes carry canonical sorted compact UTF-8 JSONL. `stdin` is `DEVNULL` and `stdout`/`stderr` are diagnostics only, so nothing an adapter prints can be read as protocol.
- **Process-tree supervision.** The worker and its descendants run in a fresh session and process group; owner-pipe EOF, timeout, cancellation, and startup failure all reap the entire tree.
- **Parent-owned rescue cleanup.** A `rescue.json` descriptor is persisted with fsync before any resource-creating call, so a crashed run's cleanup remains executable by a later caller. The parent derives every cleanup receipt; an adapter's claim never suppresses reconciliation.

The implementation keeps `local_operator.evaluation` and `local_operator.evaluation.adapters` inert on import and adds no dependency.

This is a standard C2 change and is pre-authorized. The PR is the system of record; there is no issue or RFC.

## Not activated

This slice is a foundation, deliberately unreachable from any running code path:

- Both package roots (`evaluation/__init__.py`, `evaluation/adapters/__init__.py`) are docstring-only; importing them starts nothing and spawns nothing.
- **Nothing outside the package imports it.** Verified by search: every reference to `local_operator.evaluation.adapters` lives inside the package itself or its own tests.
- No provider, model, or OpenRouter credential ever reaches an adapter. The worker environment is built from a small allowlist rather than `os.environ.copy()`, and those keys are denied even when the parent holds them.

### API

- Strict frozen `ProtocolModel` values. `AdapterSelector` pins `adapter_id`, distribution, version, entry point, package digest, release digest, schema `1.0`, absolute interpreter and workspace, and route capability. The `Handshake` repeats every pin and adds the resolved Python implementation/version/cache-tag, workspace digest, and selected route. Any mismatch is fatal; there is no negotiation.
- Secrets appear only as `SecretRef` names. `ScopedInfraValue` structurally excludes provider and model purposes, so a benchmark can be handed infrastructure credentials without becoming a second paid-model route.
- A closed method table — `hello`, `inspect_requirements`, `prepare`, `reset_start`, `observe`, `execute`, `ask_user_exchange`, `score`, `cleanup`, `begin_rescue`, `close` — with a closed state machine (`NEW`/`HANDSHAKEN`/`INSPECTED`/`PREPARED`/`RUNNING`/`FINALIZING`/`CLEANING`/`CLOSED`/`POISONED`). Control `cancel` is separate from application calls.
- Reusing a request ID requires the same method and the same canonical params digest; a changed reuse is fatal. Keyed methods carry operation IDs so a retried operation replays its recorded outcome instead of dispatching a second side effect.

### Discovery

- The host never imports adapter code and never enumerates global entry points. It validates an absolute, regular, symlink-free interpreter and workspace, then launches `python -I -s -E -m local_operator.evaluation.adapters.worker`.
- The worker resolves exactly `importlib.metadata.distribution(selector.distribution)`, checks the version, recomputes a canonical digest over `RECORD` entries and hashes, and then reads only that distribution's exact entry-point group `local_operator.evaluation_adapters.v1` — one name, one value, exactly one match. Editable and unhashed installs are refused.
- Imports are served by a `sys.meta_path` finder scoped to the distribution's `RECORD`-owned roots and installed only for the load. Stdlib, pydantic, and `local_operator` resolve normally; anything under an owned root that `RECORD` does not cover is refused rather than imported.
- `ResolvedLaunch` pins the interpreter's sha256, size, dev/ino/mode, revalidated immediately before spawn by both the parent and the supervisor.

### RPC, worker, supervisor

- Canonical JSONL with duplicate-key rejection, LF-only framing, a 1 MiB cap enforced before allocation, and partial-EOF rejection. Responses repeat their method and carry result **xor** a closed bounded error. One application call is in flight at a time.
- Unknown, late, duplicate, wrong-ID, malformed, and oversized frames poison the session and reap the tree. A timed-out response is never reused: cancel, 1 s grace, then `SIGTERM` to the group, 5 s, then `SIGKILL`. Caller cancellation is shielded until the tree is reaped.
- The worker opens its protocol FDs and redirects fd 1 to fd 2 **before** any metadata or import work, so adapter or native `print` output can never land in the protocol stream. Adapter errors surface as closed codes with no `repr`, traceback, or environment.
- The supervisor is POSIX-only and says so explicitly. It verifies `pid == pgid == getpgid(pid)` immediately before signalling, resets signal masks and dispositions in the child, leaves host signal handling untouched, and drains bounded 64 KiB stdout/stderr tails.

### Host verification

The host recomputes rather than trusts: domain-separated observation IDs from canonical content, artifact filename/digest/regular-file/`O_NOFOLLOW`/byte/media checks through the existing evidence media validation, `ActionBatch.validate_for` against the current observation, host-assigned action-batch IDs, execution-receipt ID/digest/sequence uniqueness, single-use outstanding ask exchanges, and score/cleanup plan and action IDs. The adapter supplies no lifecycle authority.

## Review history

Twelve adversarial agent-review rounds ran before this PR. Round 12 came back clean and terminal. A fresh formal agent-review round will still be recorded on this PR before merge; the pre-PR rounds do not replace the repository review gate.

| Round | Finding | Fix |
| --- | --- | --- |
| 1 | Rescue ran through `prepare`/`reset_start` side effects; keyed operations could re-dispatch | Added `begin_rescue` (`HANDSHAKEN` → `CLEANING`); bounded operation records replay a recorded outcome and refuse on exhaustion rather than evicting a side effect |
| 2 | Rescue cleanup unpinned; observation sequence unchecked; ask exchange unbound to an episode | Pinned the rescue descriptor and per-action outcomes; split initial/output/snapshot acceptance; bound ask begin/finish to an exact episode |
| 3 | Replay alias could redirect an operation; verifier not wired into real calls; launch path aliases | Alias reservation made atomic; added `VerifiedAdapterSession` owning the verifier with raw `call` made private; launch paths lstat-checked component by component |
| 4 | Cleanup receipts trusted from the adapter; ambiguous mutations left the session usable; workspace pinned by inode | Parent derives receipts; ambiguous mutation poisons and reaps; workspace pinned by a content manifest; executable pinned by content |
| 5 | **Subsystem unlaunchable**: `st_nlink != 1` rejected every real CPython interpreter (uv nlink=3, `/usr/bin/python3` nlink=78) | Dropped the link-count rule for the executable (content hash is the real binding), kept it for workspace files; added a real-interpreter test and the first end-to-end spawn test |
| 6 | Forged `__pycache__/*.pyc` executed while `RECORD` verification passed | Load the verified bytes and `compile`/`exec` them directly; `entry.load()` never runs |
| 7 | Chain verification could not cover siblings a verified module imports | Replaced the chain with the distribution-scoped `sys.meta_path` finder |
| 8 | Sourceless `.pyc` under an owned root executed unverified; stale `sys.modules` entry in a second owned root survived | Refuse every importable artifact at an uncovered stem; purge every owned root before load |
| 9 | A `RECORD`-covered C extension was refused, and behaviour differed by placement | Extensions load hash-verified at any placement; bytecode stays refused with an accurate message |
| 10 | Dual-artifact wheels (mypyc/Cython ship both `mod.py` and `mod<EXT>`) hard-refused | Prefer verified source; one shared resolver used by both the finder and the entry gate |
| 11 | Resolver inverted CPython precedence for `mod.py` + `mod/__init__.py` | Refuse that pair as ambiguous; probe outranking extension suffixes consistently; corrected a docstring that misstated CPython's precedence |
| 12 | Clean — no blocker, no major | — |

## Real-execution evidence

Measured on Darwin 25.6.0 / arm64, Python 3.14.3. The security fixes were proven by building each attack, confirming it was **live on this host first**, and only then asserting the refusal — an assertion that cannot pass on a host where the attack silently failed to take.

```text
.venv/bin/python -m pytest tests/unit/evaluation/adapters -q -n0
68 passed
```

- **Real compiled extensions.** Tests compile genuine `.so` modules with `cc` and the CPython headers (`PyInit_*`), platform-aware link flags, and a hard failure — never a skip — when a host has a compiler and headers but the build fails. All five compile-dependent tests ran here; zero skipped.
- **Forged bytecode, proven live.** A forged `__pycache__` entry and a planted sourceless `.pyc` were each shown to execute under a normal import (attacker marker written, `create()` returning `"attacker"`) before asserting the loader refuses them.
- **Real subprocesses and process groups.** End-to-end launch through `AdapterSupervisor` completes a real handshake and reaps; stubborn `SIGTERM`-ignoring grandchildren are killed by group; PID/PGID mismatch refuses to signal; host signal handlers are unchanged.
- **Contamination and secrecy.** Adapter `stdout`, native writes, and grandchild output all stay diagnostics. Environment secret markers are absent from argv, environment, logs, and the rescue descriptor.
- **Every fix is load-bearing.** Each round's fix was reverted and the matching test observed to fail, so no test asserts something the code was already doing.

## Testing status

- **Focused adapter suite: 68 passed, zero skips**, stable across repeated consecutive runs.
- **Whole-tree static gates clean**, exactly as CI runs them: `flake8`, `black --check` (26.1.0, 703 files), `isort --check`, `pyright` (0 errors).
- **The full unit suite was NOT completed locally.** Another session on this shared machine is running 12 deliberate CPU-hog processes to reproduce a TUI flake under load (load average ~253), which starved the run; a result measured under that contention would be meaningless. **The full suite is delegated to CI**, which runs the same suite on a quiet runner. No claim of local full-suite coverage is made here.
- Packaging: `uv build` produces both artifacts, `twine check` PASSED on each, and all six adapter modules are present in both the wheel and the sdist.

## Accepted boundaries

Stated deliberately rather than left implied:

- **Load-time verification is not a runtime import sandbox.** It binds the adapter's *initial module graph* to the `RECORD` digest so a result is attributable to an exact `package_digest`. An import an adapter defers until after loading returns falls back to the ordinary loader. This is documented and pinned by a test. It buys little to close: an adapter is already arbitrary code in an isolated worker and can inline whatever it likes in its own verified source.
- **dlopen re-read TOCTOU.** Source executes from the in-memory bytes that were hashed; an extension is hash-checked and then the ordinary loader re-reads that exact path, so a swap between hash and `dlopen` is undetected. Same class as the executable TOCTOU on `ResolvedLaunch`, accepted for the same reason.
- **POSIX only.** The supervisor depends on sessions, process groups, and signal semantics, and refuses explicitly elsewhere rather than degrading.

## Deferrals

- `deferred — the both-module-and-package refusal enumerates only the .py/.py pair, so a dir-vs-flat-extension pair (mod.so + mod/__init__.py) resolves the flat module while CPython would run the package (discovery.py). No unattested code executes: the resolved artifact is always RECORD-covered and hash-verified, and no real wheel ships that shape. Raised in round 12 as a MINOR and judged not worth widening the refusal now.`

## Version

`0.44.19` — **patch**. This is an inert internal foundation: it changes nothing a user can reach, adopt, or observe, and nothing outside the adapters package imports it. Per the versioning policy, a minor is reserved for a step-function capability a user would notice, which this is not until a benchmark path is actually wired to it.

### DIFF
--- a/local_operator/evaluation/adapters/__init__.py
+++ b/local_operator/evaluation/adapters/__init__.py
@@ -1,0 +1,5 @@
+"""Out-of-process evaluation adapter boundary.
+
+The package root is intentionally inert: discovery and worker imports happen
+only when their explicit modules are requested.
+"""
--- a/local_operator/evaluation/adapters/api.py
+++ b/local_operator/evaluation/adapters/api.py
@@ -17,7 +17,7 @@ from typing import Any, Literal, Protocol, TypeAlias, runtime_checkable
 from pydantic import Field, field_validator, model_validator
 
 from local_operator.evaluation.evidence.models import ScoreArtifact, canonical_digest
-from local_operator.evaluation.lifecycle import CleanupPlan, CleanupReceipt
+from local_operator.evaluation.lifecycle import CleanupPlan
 from local_operator.evaluation.protocol import ActionBatch, Observation, ProtocolModel
 from local_operator.evaluation.receipts import (
     ZERO_DIGEST,
@@ -119,6 +119,7 @@ class AdapterSelector(ProtocolModel):
     release_digest: Digest
     python_executable: str = Field(min_length=1, max_length=4096)
     workspace: str = Field(min_length=1, max_length=4096)
+    workspace_digest: Digest
     route_capability: RouteCapability
 
     @field_validator("python_executable", "workspace")
@@ -216,6 +217,8 @@ class Handshake(ProtocolModel):
             raise ValueError("handshake does not repeat the exact adapter selection")
         if self.python.executable != selector.python_executable:
             raise ValueError("handshake resolved a different Python executable")
+        if self.workspace_digest != selector.workspace_digest:
+            raise ValueError("handshake workspace digest differs from selection")
         if self.selected_route != selector.route_capability:
             raise ValueError("handshake selected a different route")
         if self.selected_route not in metadata.capabilities.routes:
@@ -451,12 +454,21 @@ class CleanupParams(OperationParams):
         return self
 
 
-class CleanupResult(ProtocolModel):
-    receipts: tuple[CleanupReceipt, ...] = Field(max_length=MAX_RECEIPTS)
+class CleanupOutcome(ProtocolModel):
+    action_id: StrictIdentifier
+    status: Literal["not_needed", "attempted", "succeeded", "failed"]
+    evidence_code: StrictIdentifier
+    duration_ms: SafeCount
 
-    @field_validator("receipts", mode="before")
+
+class CleanupResult(ProtocolModel):
+    """Adapter evidence primitives; only the parent may mint receipts."""
+
+    outcomes: tuple[CleanupOutcome, ...] = Field(max_length=MAX_RECEIPTS)
+
+    @field_validator("outcomes", mode="before")
     @classmethod
-    def _freeze_receipts(cls, value: Any) -> Any:
+    def _freeze_outcomes(cls, value: Any) -> Any:
         return tuple(value) if isinstance(value, list) else value
 
 
--- a/local_operator/evaluation/adapters/api.py
+++ b/local_operator/evaluation/adapters/api.py
@@ -1,0 +1,556 @@
+"""Strict benchmark-neutral contracts shared by adapter hosts and workers.
+
+Adapters are untrusted infrastructure plugins.  This module therefore contains
+only immutable values, closed enums, and content-bound identities; importing it
+must not discover or import an adapter distribution.
+"""
+
+from __future__ import annotations
+
+import hashlib
+import os
+import platform
+import sys
+from collections.abc import Mapping
+from pathlib import Path
+from typing import Annotated, Any, Literal, Protocol, TypeAlias, runtime_checkable
+
+from pydantic import Field, field_validator, model_validator
+
+from local_operator.evaluation.evidence.models import ScoreArtifact, canonical_digest
+from local_operator.evaluation.lifecycle import CleanupPlan, CleanupReceipt
+from local_operator.evaluation.protocol import ActionBatch, Observation, ProtocolModel
+from local_operator.evaluation.receipts import (
+    ZERO_DIGEST,
+    Digest,
+    SafeCount,
+    StrictIdentifier,
+)
+
+ADAPTER_SCHEMA_VERSION = "1.0"
+ADAPTER_ENTRY_POINT_GROUP = "local_operator.evaluation_adapters.v1"
+MAX_RESCUE_REFS = 256
+MAX_REQUIREMENTS = 256
+MAX_RECEIPTS = 256
+
+AdapterMethod: TypeAlias = Literal[
+    "hello",
+    "inspect_requirements",
+    "prepare",
+    "reset_start",
+    "observe",
+    "execute",
+    "ask_user_exchange",
+    "score",
+    "cleanup",
+    "close",
+]
+AdapterState: TypeAlias = Literal[
+    "NEW",
+    "HANDSHAKEN",
+    "INSPECTED",
+    "PREPARED",
+    "RUNNING",
+    "FINALIZING",
+    "CLEANING",
+    "CLOSED",
+    "POISONED",
+]
+RouteCapability: TypeAlias = Literal["computer", "browser", "terminal"]
+
+# Both peers use this table.  Keeping it data, rather than distributed branches,
+# makes an invalid transition fail before plugin code can acquire authority.
+METHOD_STATES: Mapping[AdapterMethod, frozenset[AdapterState]] = {
+    "hello": frozenset({"NEW"}),
+    "inspect_requirements": frozenset({"HANDSHAKEN"}),
+    "prepare": frozenset({"INSPECTED"}),
+    "reset_start": frozenset({"PREPARED"}),
+    "observe": frozenset({"RUNNING"}),
+    "execute": frozenset({"RUNNING"}),
+    "ask_user_exchange": frozenset({"RUNNING"}),
+    "score": frozenset({"RUNNING", "FINALIZING"}),
+    "cleanup": frozenset({"PREPARED", "RUNNING", "FINALIZING", "CLEANING"}),
+    "close": frozenset(
+        {"HANDSHAKEN", "INSPECTED", "PREPARED", "RUNNING", "FINALIZING", "CLEANING"}
+    ),
+}
+METHOD_NEXT_STATE: Mapping[AdapterMethod, AdapterState] = {
+    "hello": "HANDSHAKEN",
+    "inspect_requirements": "INSPECTED",
+    "prepare": "PREPARED",
+    "reset_start": "RUNNING",
+    "observe": "RUNNING",
+    "execute": "RUNNING",
+    "ask_user_exchange": "RUNNING",
+    "score": "FINALIZING",
+    "cleanup": "CLEANING",
+    "close": "CLOSED",
+}
+KEYED_METHODS = frozenset(
+    {"prepare", "reset_start", "execute", "ask_user_exchange", "score", "cleanup", "close"}
+)
+READ_ONLY_METHODS = frozenset({"hello", "inspect_requirements", "observe"})
+
+
+def canonical_params_digest(method: AdapterMethod, params: ProtocolModel) -> Digest:
+    return canonical_digest(f"adapter-rpc-params-{method}-v1", params)
+
+
+class AdapterSelector(ProtocolModel):
+    """Exact wheel and interpreter selection; ranges and fallback are absent."""
+
+    schema_version: Literal["1.0"]
+    adapter_id: StrictIdentifier
+    distribution: StrictIdentifier
+    version: StrictIdentifier
+    entry_point: StrictIdentifier
+    package_digest: Digest
+    release_digest: Digest
+    python_executable: str = Field(min_length=1, max_length=4096)
+    workspace: str = Field(min_length=1, max_length=4096)
+    route_capability: RouteCapability
+
+    @field_validator("python_executable", "workspace")
+    @classmethod
+    def _absolute_path(cls, value: str) -> str:
+        if not os.path.isabs(value) or os.path.normpath(value) != value:
+            raise ValueError("adapter paths must be normalized absolute paths")
+        return value
+
+
+class AdapterCapabilities(ProtocolModel):
+    routes: tuple[RouteCapability, ...] = Field(min_length=1, max_length=3)
+    ask_user: bool
+    scoring: bool
+
+    @field_validator("routes", mode="before")
+    @classmethod
+    def _freeze_routes(cls, value: Any) -> Any:
+        return tuple(value) if isinstance(value, list) else value
+
+    @model_validator(mode="after")
+    def _unique_routes(self) -> "AdapterCapabilities":
+        if len(self.routes) != len(set(self.routes)):
+            raise ValueError("adapter routes must be unique")
+        return self
+
+
+class AdapterMetadata(ProtocolModel):
+    adapter_id: StrictIdentifier
+    distribution: StrictIdentifier
+    version: StrictIdentifier
+    entry_point: StrictIdentifier
+    package_digest: Digest
+    release_digest: Digest
+    schema_version: Literal["1.0"]
+    capabilities: AdapterCapabilities
+
+
+class PythonRuntime(ProtocolModel):
+    executable: str = Field(min_length=1, max_length=4096)
+    implementation: StrictIdentifier
+    version: StrictIdentifier
+    cache_tag: StrictIdentifier
+
+    @field_validator("executable")
+    @classmethod
+    def _absolute_executable(cls, value: str) -> str:
+        if not os.path.isabs(value) or os.path.normpath(value) != value:
+            raise ValueError("resolved Python executable must be normalized and absolute")
+        return value
+
+    @classmethod
+    def current(cls) -> "PythonRuntime":
+        cache_tag = sys.implementation.cache_tag
+        if cache_tag is None:
+            raise RuntimeError("selected Python does not expose a cache tag")
+        return cls(
+            executable=str(Path(sys.executable).resolve()),
+            implementation=platform.python_implementation(),
+            version=platform.python_version(),
+            cache_tag=cache_tag,
+        )
+
+
+class Handshake(ProtocolModel):
+    selector: AdapterSelector
+    metadata: AdapterMetadata
+    python: PythonRuntime
+    workspace_digest: Digest
+    selected_route: RouteCapability
+
+    @model_validator(mode="after")
+    def _repeat_exact_pins(self) -> "Handshake":
+        selector = self.selector
+        metadata = self.metadata
+        repeated = (
+            metadata.adapter_id,
+            metadata.distribution,
+            metadata.version,
+            metadata.entry_point,
+            metadata.package_digest,
+            metadata.release_digest,
+            metadata.schema_version,
+        )
+        selected = (
+            selector.adapter_id,
+            selector.distribution,
+            selector.version,
+            selector.entry_point,
+            selector.package_digest,
+            selector.release_digest,
+            selector.schema_version,
+        )
+        if repeated != selected:
+            raise ValueError("handshake does not repeat the exact adapter selection")
+        if self.python.executable != selector.python_executable:
+            raise ValueError("handshake resolved a different Python executable")
+        if self.selected_route != selector.route_capability:
+            raise ValueError("handshake selected a different route")
+        if self.selected_route not in metadata.capabilities.routes:
+            raise ValueError("selected route is not an adapter capability")
+        return self
+
+
+class SecretRef(ProtocolModel):
+    """An opaque name resolved by infrastructure, never secret bytes."""
+
+    name: str = Field(min_length=1, max_length=256, pattern=r"^[A-Z][A-Z0-9_]*$")
+
+
+InfraPurpose: TypeAlias = Literal[
+    "benchmark_auth",
+    "benchmark_compute",
+    "benchmark_storage",
+    "artifact_storage",
+]
+
+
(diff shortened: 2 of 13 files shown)


damianvtran:
### CI failure diagnosis — `test_launch_identity_detects_swap` (Linux-only)

**Classification: (a) test defect, with a second latent Linux-only defect found while verifying it.** The production guard is sound on Linux; the test's swap was a no-op there, so the guard was never exercised on the platform CI and most deployments run.

#### Root cause

The test swapped the workspace with `rmtree()` + `mkdir()`. Measured in `python:3.12-slim`:

```text
rmtree+mkdir      : identity changed = False | ino 496107 -> 496107   (Linux/ext4)
rmtree+mkdir      : identity changed = True  | ino 686148231 -> 686148234 (macOS/APFS)
```

Deleting the directory frees its inode and ext4 hands the very next `mkdir` the **same inode number** back. APFS allocates a fresh one, so the identical test looked green on macOS while asserting nothing on Linux. The same reuse applies to files (`file unlink+create: changed = False`).

#### Production guard is NOT affected

`validate_resolved_launch` was checked directly against realistic swaps in the Linux container, and detects both:

```text
Linux (python:3.12-slim)
workspace rename-over       -> detected (adapter launch path identity changed before spawn)
executable unlink+recreate  -> detected (adapter launch path identity changed before spawn)
```

An attacker performing a real substitution prepares a replacement and renames it into place; the original stays referenced, so its inode is never freed and the replacement necessarily gets a different one. Executable substitution is additionally pinned by sha256, which catches a same-size, same-inode content swap that dev/ino alone would miss. The only construction the dev/ino tuple misses is delete-then-recreate of the *workspace* — and that changes workspace content, which the worker re-hashes at handshake (`content digest differs (handshake catches): True`).

#### Fix

The test now moves the original aside and renames a prepared directory into place — the realistic swap — and carries a precondition assert that fails loudly if a future filesystem reuses the identity anyway, so it can never pass vacuously again.

Verified load-bearing **on Linux**, not just green: removing the workspace dev/ino pin from the revalidation tuple makes it fail, restoring it makes it pass.

```text
Linux, workspace pin reverted : FAILED test_launch_identity_detects_swap
Linux, restored               : 1 passed
```

#### Second defect found while verifying (Linux-only, would have been the next red CI)

Running the full adapter suite in the container exposed `pytest exit=139` (SIGSEGV) **after** reporting `68 passed` — a crashed process that reads as a green run. Bisected to the extension tamper phase, then reproduced minimally:

```text
no tamper                                   -> exit=0
tamper the .so in place while it is loaded  -> exit=139
```

The test rewrote the `.so` through its existing inode while that file was still `dlopen`-mapped, corrupting the live mapping so glibc faulted at interpreter shutdown. macOS tolerates it. Fixed by staging the mutated bytes in a new file and `replace()`-ing it in, which swaps the directory entry and leaves the mapped inode untouched. The tamper assertion remains load-bearing: skipping extension hash verification still fails both parametrisations on Linux.

```text
Linux, full adapter suite after fix: 68 passed, pytest exit=0
```

#### Evidence summary

- Linux adapter suite: **68 passed, exit 0**, with a real `gcc` present so every compile-dependent test ran (zero skips).
- macOS adapter suite: **68 passed**.
- Whole-tree gates: flake8 clean, black 26.1.0 (703 files) unchanged, isort clean, pyright 0 errors.
- No production code changed by this commit — the diff is test-only.

damianvtran:
### CI on `f17b2e2e` — adapter failure fixed; two unrelated failures remain

`test (3.12)`: **2 failed, 10030 passed, 17 skipped**. **Zero adapter failures** — `test_launch_identity_detects_swap` now passes on Linux, and the segfault is gone. Everything else is green: lint, type-check, pip-audit, tui-e2e (macos + ubuntu).

Neither remaining failure comes from this PR, which touches only `local_operator/evaluation/adapters/**`, its own tests, and the version bump.

**1. `test_pricing.py::test_the_sonnet_5_introductory_price_has_not_expired` — a calendar tripwire that expired today, not a regression.**

The test asserts `date.today() < date(2026, 9, 1)`; CI runs UTC, where it is now 2026-09-01. Verified against **clean `origin/main` with none of this PR's code**:

```text
origin/main, local TZ (2026-08-31): 1 passed
origin/main, TZ=UTC  (2026-09-01): 1 failed
```

It fails identically on `main` and will fail on every PR until someone applies the two-line fix its own message names (`registry.py`: `input_price=3.0`, `output_price=15.0`, `cache_writes_price=3.75`, `cache_reads_price=0.30`, then move the guard date). That is a deliberate, working tripwire doing its job — and it is out of scope here. Fixing Anthropic pricing inside an evaluation-adapters PR would be an unrequested change to an unrelated subsystem, so I am flagging it rather than silently folding it in.

**2. `test_steering_approval.py::test_a_held_answer_key_resolves_cleanly_on_every_second_key` — the known pre-existing TUI flake.**

Passes in isolation:

```text
tests/unit/tui/test_steering_approval.py -k test_a_held_answer_key_resolves_cleanly_on_every_second_key
1 passed, 113 deselected
```

This is the flake already being investigated separately under CPU-load reproduction; it did not fire on the previous run of this same branch.

damianvtran:
### CI fully green on `26a3a441`

Rebased onto `origin/main` at `e7fe7a46` (which carries the Sonnet 5 pricing fix, `0.44.18`). The stale `0.44.17` bump was removed from history rather than layered over, so the branch carries **exactly one** version commit: `0.44.19`, and `pyproject.toml` produced no rebase conflict.

All eight checks pass:

```text
lint                  pass   58s
type-check            pass   1m12s
pip-audit             pass   57s
tui-e2e (macos)       pass   37s
tui-e2e (ubuntu)      pass   40s
server-sanity         pass   29s
cli-sanity            pass   4m59s
test (3.12)           pass   28m20s
```

Run conclusion: **success**. Zero `FAILED` lines in the full suite — the previously-red `test_pricing` tripwire is resolved on main, and the `test_steering_approval` flake did not fire on this run.

Local verification on the rebased head:

- macOS adapter suite: **68 passed**
- Linux adapter suite (`python:3.12-slim`, real `gcc`, zero skips): **68 passed, pytest exit 0, empty stderr** — the SIGSEGV fixed in the previous commit stays fixed after the rebase
- Whole-tree gates: flake8 clean, black 26.1.0 (707 files) unchanged, isort clean, pyright 0 errors

`mergeStateStatus=BLOCKED` reflects the outstanding review approval, not a failing check.

damianvtran:
### Agent review — round 13
Reviewer: reviewer subagent (anthropic/claude-opus-5; completed on zai/glm-5.3 after a provider fallback mid-round)
Scope: e7fe7a461fb97d1ac3f45d75441a4c609600cdcf..26a3a441b4727f9d3306509660f47194a2a92e09

Scope endpoints verified exact: base is the true merge-base with `main`; head matches the PR head; reviewed in a clean detached worktree at `26a3a441`. `git diff` first, tree only where the diff was unclear. 32 commits, 5587 insertions, one line of `pyproject.toml`.

#### Linux verification (all reproduced in `python:3.12-slim`, ext4/overlayfs)

**The inode-reuse premise is real, not asserted.** Probe: `rmtree`+`mkdir` on the same path returns the identical `(st_dev, st_ino)` — `INODE_REUSED: True`. The pre-fix test was genuinely vacuous on Linux.

**The identity test is now load-bearing on Linux.** Deleted the four workspace dev/ino pin lines from `validate_resolved_launch` → `test_launch_identity_detects_swap` FAILS (1 failed, rc=1). Restored → passes. Not vacuous.

**The precondition assert genuinely fires.** Re-substituted the old `rmtree`+`mkdir` swap → the assert fails loudly with its intended message (`assert (94, 560831) != (94, 560831)`, "the swap reused the original identity…"), rather than passing silently.

**Adapter suite exits 0 with no SIGSEGV.** `68 passed in 12.65s`, `EXIT_CODE=0`. The `staged.replace(extension)` rewrite is what removes the 139: the mutated `.so` no longer writes through the still-dlopen-mapped inode. One environment note, not a defect: in a container whose PID 1 is `sleep` (never reaps), `test_process_group_termination_kills_stubborn_grandchild` fails because the orphan stays a zombie forever — under `docker-init` it passes. CI runners have a reaping init; no action needed.

**Extension hash verification is load-bearing.** Neutered only the module-level `RECORD hash differs` raise → both parametrizations of `test_record_covered_extension_loads_and_is_hash_verified` FAIL; restored → pass. The tamper assertion protects something real.

#### The layered-defence finding — and the severity call

Driven directly against production, no test helpers, on an inode-reusing filesystem:

| Attack | `validate_resolved_launch` alone | Parent spawn gate (`worker_argv`) | Child handshake (`_workspace_digest`) |
| --- | --- | --- | --- |
| workspace rename-over | **detected** | detected | detected |
| workspace delete+recreate | **MISSED** (inode reused) | **detected** — content digest | **detected** — content digest |
| delete+recreate + restored manifest + same-length payload | (missed) | **detected** — content digest | **detected** — content digest |
| executable unlink+recreate, same size | **detected** (sha256) | detected | detected |
| executable rename-over | **detected** | detected | detected |

So the assigned question — "if ANY real swap goes undetected on Linux, that is a BLOCKER" — answers: **no real swap reaches execution undetected.** `validate_resolved_launch` in isolation misses workspace delete+recreate, but `AdapterSupervisor.launch` builds `argv` through `worker_argv` (supervisor.py:267), which re-hashes the workspace content in the parent **before any process is spawned** (discovery.py:224), and the worker re-verifies again at handshake (worker.py:318). The dev/ino pin is one layer of three, and the composite is sound on Linux. Not a blocker; not a major.

#### Static gates (whole tree, as CI runs them)

`flake8` clean · `black --check` (26.1.0) clean, 707 files · `isort --check` clean · `pyright` 0 errors.

#### Spot-checks

- **Non-activation holds.** Zero references to `evaluation.adapters` outside the package and its own tests; both package roots are docstring-only and import nothing.
- **Twelve rounds' guards all still present and named** at head: forged bytecode (source / sibling / package-init), sourceless `.pyc`, unrecorded source/extension/package-init, RECORD absolute-and-traversal, module-vs-package ambiguity, outranking extension, dual-artifact source preference, poison gating, rescue plan/action pinning, process-group teardown, finder scope + second-root purge, launch-identity swap. Nothing regressed in the rebase.
- **Round-12 deferral** is present as a `deferred — <reason>` line in the body. ✓
- **Version:** `0.44.19` on top of `main`'s `0.44.18` — the correct next PATCH, exactly one bump commit (`26a3a441`), and a patch is the right materiality call for an inert internal foundation. ✓

#### Findings

**MINOR-1 — the PR body's `## Version` section is stale.** It states `0.44.17`; the head bumps to `0.44.19`. The section was written before the rebase re-bump and not refreshed. One-line fix to the body, no code change.

**MINOR-2 — `discovery.py:153-156` understates the actual defence.** The docstring says the workspace "is not re-verified immediately before spawn the way the executable identity is" — but `worker_argv` *does* re-verify workspace content in the parent pre-spawn, and `AdapterSupervisor.launch` routes through it. The acknowledged gap is narrower than documented, in the safe direction. Correct the comment in a follow-up; it should not be widened into a security change on this PR.

No BLOCKER findings. No MAJOR findings. (2 MINOR, 0 NIT.)

#### Verdict

**clean — TERMINAL.** The two minors are documentation-accuracy follow-ups, not reasons for another round; nothing here warrants a fourteenth pass. The three changes since round 12 are all verified by reproduction rather than accepted on description, the composite launch boundary is sound on Linux where its weakest layer is weakest, and CI is green on this head including the full suite. **I recommend merging.**

Post-merge obligations unchanged: fix MINOR-1 in the body before or immediately after merge, fold MINOR-2 into the next touch of `discovery.py`.

damianvtran:
### Agent review remediation — round 13

Round 13 (https://github.com/damianvtran/local-operator/pull/511#issuecomment-5488021889) returned clean and terminal: 0 blockers, 0 majors, recommend-merge, with all load-bearing checks independently reproduced on Linux. Two minors, both documentation-accuracy, addressed here without moving the head.

**Minor 1 — stale version in the PR description: fixed in place.**
The description said `0.44.17` while the head is `0.44.19` (the branch was rebased onto `e7fe7a46` after `0.44.18` shipped, and the bump was redone rather than layered). Corrected via a description-only edit; **no code changed and the head is unchanged** at `26a3a441b4727f9d3306509660f47194a2a92e09`, so round 13 remains fresh. Verified after the edit: zero `0.44.17` mentions remain in the rendered body, and the description differs from its prior text only in that one version string (plus the trailing newline GitHub appends).

**Minor 2 — `discovery.py:153-156` understates the pre-spawn workspace re-verification that `worker_argv` performs:**
`deferred — comment-only accuracy nit; a comment-only push would invalidate a terminal fresh review round for zero behavioural change; correct in the next evaluation PR touching discovery.py`

The docstring says the workspace "is not re-verified immediately before spawn the way the executable identity is", which understates what `worker_argv` actually does — it recomputes and checks `workspace_digest` before the worker launches. The code is correct and tested; only the comment is narrower than the truth. Fixing it would change the head SHA and void a terminal, fresh, clean round for no behavioural difference, so it is deferred to the next evaluation PR that touches `discovery.py` for any real reason. This comment is the record.

**Freshness and CI confirmation.**
- Reviewed head: `26a3a441b4727f9d3306509660f47194a2a92e09`
- Current PR head: `26a3a441b4727f9d3306509660f47194a2a92e09` — equal; the round is fresh.
- CI on that head: all eight checks pass, run conclusion `success` (lint, type-check, pip-audit, tui-e2e macos + ubuntu, server-sanity, cli-sanity, `test (3.12)` full suite — zero failures).

PR #491, opened 2026-08-31
fix(tui): catch up team picker

damianvtran:
## Summary

Typing `/team lop` before the session finished adopting could permanently strand the argument picker on its empty opening fill. Later keystrokes only filtered the empty snapshot, and inline completion could drop the name highlight as soon as the caret moved into the reassembled message. Team registry construction and its five-second refresh also read both 8k brief files for every team even though picker/list surfaces only need roster metadata.

This patch:

- parses only `team.yml` during `TeamRegistry._load` and loads `instructions.md` / `project.md` once on the first full `get_team*` lookup;
- keeps brief loadedness registry-owned (never encoded in public strings) and preserves disk briefs when metadata-only teams are saved directly or after Pydantic Python/JSON transport round trips, while hydrated direct edits and explicit `update_team(..., TeamEditFields(instructions="", project=""))` still persist;
- retries only empty, non-dismissed `/team` / `/teams` / `/agent` / `/agents` lists against the current query, and refills once more at session adoption;
- reserves one restrained, non-selectable `loading teams…` / `loading agent roster…` row while the session is unavailable or transitioning, then replaces it in place with the first real row; pending Tab/Enter are exact no-ops, mouse has no target, Esc collapses and latches dismissal, and an adopted empty roster reserves nothing;
- centralizes name-list rows plus the O(1) highlight snapshot in one app helper;
- preserves a same-family team/agent snapshot whenever the leading name command remains live, including single-line and multiline inline reassembly;
- preserves the reserved `chart` styling, team-first Tab completion, and parked switch/send hint;
- merges current `origin/main` (`7ac34cdf`, released as `v0.44.19`) and bumps the next unused publishable patch to **0.44.20**.

## Real-path evidence

The real `OperatorApp` was driven with the session factory deliberately blocked, then released with real `TeamRegistry` and `AgentRegistry` instances. The shipped stylesheet was loaded for every frame.

- `/team lop` now shows a one-row, non-selectable pending notice; after adoption the `lopdev` row replaces it at the same coordinate without a backspace or retype.
- Three consecutive 100×30 states — pending, first filled, settled filled — are geometry-identical: `composer y=24`, `picker y=25 height=1`, `status y=26`, `welcome y=1`, version SVG `y=337.2`, `screen=98×28`, `virtual=98×28`, scrollbar false. Both pending and populated rows render at SVG `y=630`.
[...]

PR #512, opened 2026-09-01
fix(model): pin Sonnet 5 standard price, drop the lapsed date guard

damianvtran:
## Why

**CI has been red on every PR in the repo since 2026-09-01 UTC.** `tests/unit/model/test_pricing.py::test_the_sonnet_5_introductory_price_has_not_expired` asserted `date.today() < date(2026, 9, 1)`, and that date arrived. It still passes on a US-local machine (the repo's timezone is UTC-4/5 in CI terms), which is why local runs looked clean while every pipeline failed — the guard fired for CI hours before it fired for anyone running the suite by hand.

The guard was a deliberate date-bomb: Anthropic launched Sonnet 5 at $2/$10 as introductory pricing through 2026-08-31, with a rise to $3/$15 scheduled for 2026-09-01, and the guard existed so that dated price could not go silently wrong.

## No price changed, and that is the correct outcome

**The guard's own remediation message was wrong.** It instructed setting `input_price=3.0, output_price=15.0, cache_writes_price=3.75, cache_reads_price=0.30`. Anthropic cancelled that increase:

> **Edit August 10, 2026:** Sonnet 5's introductory pricing of $2 per million input tokens and $10 per million output tokens is now permanent. The standard pricing of $3 input / $15 output previously set to take effect September 1 no longer applies.
> — [Introducing Claude Sonnet 5](https://www.anthropic.com/news/claude-sonnet-5), launch-post changelog

> The $2/$10 per million input/output token pricing for Claude Sonnet 5, announced at launch as introductory pricing through August 31, 2026, is now the standard price. The previously scheduled increase to $3/$15 per million input/output tokens on September 1, 2026 will not occur.
> — [`claude-sonnet-5-introductory-pricing` note](https://platform.claude.com/docs/en/about-claude/pricing), platform pricing page

The live pricing table confirms the cache legs too: Sonnet 5 is `$2` base input / `$2.50` 5m cache write / `$0.20` cache hit / `$10` output — exactly the four values already in the registry. Corroborated independently by [@claudeai](https://x.com/claudeai/status/2086891169217122586) ("We're making Claude Sonnet 5's introductory pricing permanent").

So following the guard's instruction would have introduced a real defect: a **50% over-report on every Sonnet 5 call**, in the `/analytics` cost overlay, the stored `cost_micro` ledger, and the TUI status band. The four prices in `registry.py` are untouched.

## What changed

- **`local_operator/model/registry.py`** — comment block above the `claude-sonnet-5` prices rewritten to record the actual history (launched introductory through 2026-08-31 → made permanent 2026-08-10 when the increase was cancelled), with the sources inline. The now-false `$X from 2026-09-01` trailing comments are gone. It explicitly warns against "restoring" $3/$15 from a stale third-party table, since many still print the cancelled increase.
- **`tests/unit/model/test_pricing.py`** — `test_the_sonnet_5_introductory_price_has_not_expired` → `test_sonnet_5_carries_the_standard_price`, now pinning the registry row to `(2.0, 10.0, 2.50, 0.20)`. The `datetime.date` import is dropped; the test has **no time dependency at all**.

## What the guard now protects

The risk inverted. There is no announced end date, so there is no longer a day on which a correct value goes stale. What remains is the opposite failure: someone updating this row from one of the many third-party pricing tables that still carry the cancelled $3/$15 increase. The pin fails the build on *any* deviation from the published numbers and points at Anthropic's page for re-verification — a silent regression back to the introductory-era assumption, or forward to the cancelled one, cannot land.

## Testing evidence

**The reported failure, reproduced.** Machine clock reads `Mon 31 Aug 2026 20:53 EDT` / `Tue 1 Sep 2026 00:53 UTC`, reproducing the local-vs-CI split:

```
$ .venv/bin/python -m pytest tests/unit/model/test_pricing.py -q      # real local clock
31 passed in 5.84s

$ # with date.today() patched to 2026-09-01, as CI sees it
FAILED tests/unit/model/test_pricing.py::test_the_sonnet_5_introductory_price_has_not_expired
AssertionError: Claude Sonnet 5's introductory rate has lapsed. In registry.py set
input_price=3.0, output_price=15.0, cache_writes_price=3.75, cache_reads_price=0.30 ...
assert datetime.date(2026, 9, 1) < datetime.date(2026, 9, 1)
```

**The new pin bites.** Perturbing the registry — including to the exact value the old guard demanded — fails it:

```
$ # input_price 2.0 -> 3.0 (the edit the old guard instructed)
1 failed   ...  +  3.0,  vs  expected 2.0
$ # cache_reads_price 0.20 -> 0.30
1 failed   ...  +  0.3,  vs  expected 0.2
$ # unperturbed
1 passed in 6.02s
```

**It is genuinely time-insensitive.** The whole file, with `date.today()` patched forward — the scenario that broke the old guard:

```
--- simulated today=2026-09-01 ---   31 passed
--- simulated today=2027-06-01 ---   31 passed
--- simulated today=2030-01-01 ---   31 passed
```

**No date bomb remains.** `grep -rn "date.today() <|date.today() >|date([0-9]" tests/ local_operator/` returns nothing.

**Gates**, whole-tree, exactly as CI runs them: `flake8` clean · `black==26.1.0 --check` 695 files unchanged · `isort==5.13.2 --check` clean · `pyright` 0 errors, 0 warnings. Full unit suite was started but the machine is under load average ~269 from a concurrent session; rather than report a starved number I am relying on this PR's CI for the whole-suite result. The focused pricing file (31 tests) passes locally.

## Noted, not changed

`openai_models`' `gpt-5.6-sol` row carries "Promotional rate through at least 2026-11-21". It is **not yet due** and carries no `date()` guard, so it is untouched here — recorded as a known future date, nothing more.

## Impact

Unblocks CI for every open PR in the repo. Patch bump to **0.44.18** — a build fix and a comment/test correction, no user-facing behaviour change and no price change.

### DIFF
--- a/local_operator/model/registry.py
+++ b/local_operator/model/registry.py
@@ -424,17 +424,26 @@ anthropic_models: Dict[str, ModelInfo] = {
         supports_images=True,
         supports_prompt_cache=True,
         limits_from_listing=True,
-        # The rate ACTUALLY in effect: Anthropic's introductory pricing for Sonnet
-        # 5 runs through 2026-08-31, and the standard $3/$15/$3.75/$0.30 takes over
-        # on 2026-09-01. Both rows are published on the same page, cache legs
-        # included, so this is a transcription rather than a choice between two
-        # wrong numbers — and `test_the_sonnet_5_introductory_price_has_not_expired`
-        # fails the build on the changeover date naming the exact edit, so the row
-        # cannot silently under-report by a third the way a bare comment would let it.
-        input_price=2.0,  # $2 / MTok introductory; $3 from 2026-09-01
-        output_price=10.0,  # $10 / MTok introductory; $15 from 2026-09-01
-        cache_writes_price=2.50,  # $2.50 / MTok (5m write); $3.75 from 2026-09-01
-        cache_reads_price=0.20,  # $0.20 / MTok; $0.30 from 2026-09-01
+        # These are Sonnet 5's STANDARD rates, not a promotion, and they are not
+        # dated. The history matters because it is the opposite of what an older
+        # comment here claimed: the model launched 2026-06-30 at $2/$10 billed as
+        # introductory pricing through 2026-08-31, with a rise to $3/$15/$3.75/$0.30
+        # scheduled for 2026-09-01. On 2026-08-10 Anthropic cancelled that rise and
+        # made these numbers permanent (launch-post changelog edit of 2026-08-10 on
+        # https://www.anthropic.com/news/claude-sonnet-5, and the
+        # `claude-sonnet-5-introductory-pricing` note on
+        # https://platform.claude.com/docs/en/about-claude/pricing: "The previously
+        # scheduled increase to $3/$15 ... will not occur").
+        #
+        # So do NOT "restore" $3/$15 here on the strength of a stale third-party
+        # pricing table — several still carry the cancelled increase. Anything other
+        # than these four values over-reports every Sonnet 5 call in the status band
+        # and the analytics ledger, which is what `test_sonnet_5_carries_the_standard_price`
+        # pins against.
+        input_price=2.0,  # $2 / MTok
+        output_price=10.0,  # $10 / MTok
+        cache_writes_price=2.50,  # $2.50 / MTok (5m write)
+        cache_reads_price=0.20,  # $0.20 / MTok
         description=(
             "Claude Sonnet 5: the balanced tier of the 5 generation, with the same "
             "1M-token context window and 128k output as Opus 5."
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "local-operator"
-version = "0.44.17"
+version = "0.44.18"
 description = "A team of proactive AI assistants that can work together behind the scenes to help you get mundane tasks done so you can focus on the fun stuff."
 readme = "README.md"
 authors = [{ name = "Damian Tran", email = "damian@gominerva.com" }]
--- a/tests/unit/model/test_pricing.py
+++ b/tests/unit/model/test_pricing.py
@@ -10,7 +10,6 @@ own listing quotes none — so there are two separate guards here.
 
 from __future__ import annotations
 
-from datetime import date
 from typing import Any
 from unittest.mock import patch
 
@@ -388,17 +387,38 @@ def test_the_two_listing_legs_share_one_ceiling_rather_than_summing() -> None:
     )
 
 
-def test_the_sonnet_5_introductory_price_has_not_expired() -> None:
-    """Anthropic's Sonnet 5 promotion ends 2026-08-31; the row carries it.
+def test_sonnet_5_carries_the_standard_price() -> None:
+    """Sonnet 5's $2/$10/$2.50/$0.20 is the permanent standard rate — pin it.
 
-    A dated price is the one case where a correct value goes silently wrong on a
-    known day, so it gets a guard rather than a comment. When this fails, the fix
-    is the two-line edit named in the message — not a change to this test.
+    This replaces a dated guard that watched for the introductory period ending
+    2026-08-31. That expiry never happened: Anthropic cancelled the scheduled
+    2026-09-01 rise to $3/$15 and made these rates permanent on 2026-08-10
+    (launch-post changelog edit on https://www.anthropic.com/news/claude-sonnet-5,
+    and the `claude-sonnet-5-introductory-pricing` note on
+    https://platform.claude.com/docs/en/about-claude/pricing). The old guard duly
+    fired on 2026-09-01 UTC and failed every PR in the repo while instructing an
+    edit that would have over-reported every Sonnet 5 call by 50%.
+
+    With no announced end date there is no longer a date worth watching, so the
+    risk inverts: not "a correct value goes stale on a known day" but "someone
+    copies one of the many third-party tables that still print the cancelled
+    $3/$15 increase". Hence a pin on the published numbers rather than a clock —
+    deliberately time-insensitive, so it can never fail for merely being run on a
+    later date.
     """
-    assert date.today() < date(2026, 9, 1), (
-        "Claude Sonnet 5's introductory rate has lapsed. In registry.py set "
-        "input_price=3.0, output_price=15.0, cache_writes_price=3.75, "
-        "cache_reads_price=0.30, and update this guard to the next known change."
+    info = anthropic_models["claude-sonnet-5"]
+    actual = (
+        info.input_price,
+        info.output_price,
+        info.cache_writes_price,
+        info.cache_reads_price,
+    )
+    assert actual == (2.0, 10.0, 2.50, 0.20), (
+        f"Claude Sonnet 5's registry price is {actual}, not the published standard "
+        "(2.0, 10.0, 2.50, 0.20). $2/$10 is permanent as of 2026-08-10 — the "
+        "$3/$15 increase once scheduled for 2026-09-01 was cancelled. Re-verify "
+        "against https://platform.claude.com/docs/en/about-claude/pricing before "
+        "changing this, not against a third-party pricing table."
     )
 
 


damianvtran:
### Agent review — round 1

Scope: 562ebd0ab0a446e3b976999d3b5cc49164d8808b..454833fc4ffea5a5a8102ec08dba8dc98c1b053b

Reviewed from a clean detached worktree at the exact PR head. I treated the PR's central factual claim as unproven and re-derived it from primary sources before reviewing the code, because the whole change rests on it.

---

#### 1. The pricing premise — independently confirmed

I did not take the PR's word for this. Read live, both sources it cites:

- **Anthropic launch post** (`anthropic.com/news/claude-sonnet-5`) — changelog carries verbatim: *"Edit August 10, 2026: Sonnet 5's introductory pricing of $2 per million input tokens and $10 per million output tokens is now permanent. The standard pricing of $3 input / $15 output previously set to take effect September 1 no longer applies."*
- **Platform pricing table** (`platform.claude.com/docs/en/about-claude/pricing`) — the live model table row reads `Claude Sonnet 5 | $2 / MTok | $2.50 / MTok | $4 / MTok | $0.20 / MTok | $10 / MTok`, and the `claude-sonnet-5-introductory-pricing` note states the September 1 increase *"will not occur"*.

Cross-checked against an unrelated third party (OpenRouter's Sonnet 5 page: `$2.00` input / `$10.00` output / `$0.20` cache read / `$2.50` cache write, with Anthropic, Bedrock, Vertex and Azure all posting `$2.00/$10.00/$0.20`).

**All four registry values — 2.0 / 10.0 / 2.50 / 0.20 — match the published standard rates exactly.** The premise holds and the old guard's remediation message was indeed wrong; following it would have over-reported every Sonnet 5 call by 50%. No blocker here.

#### 2. No price value changed — confirmed

`git diff` on `registry.py` shows the four assignments changing only in their trailing comments (`# $2 / MTok introductory; $3 from 2026-09-01` → `# $2 / MTok`). Values are byte-identical at 2.0 / 10.0 / 2.50 / 0.20. Verified by import at head: `prices: 2.0 10.0 2.5 0.2`.

#### 3. The pin genuinely bites on all four fields — reproduced

I perturbed each field independently in `registry.py`, ran only the new test, and restored the file byte-identical after each (`git status` clean afterwards):

```
UNPERTURBED                          rc=0  PASSES (correct)     | 1 passed in 1.55s
input_price          2.0  -> 3.0     rc=1  BITES                | 1 failed in 1.97s
output_price        10.0  -> 15.0    rc=1  BITES                | 1 failed in 3.39s
cache_writes_price   2.50 -> 3.75    rc=1  BITES                | 1 failed in 1.96s
cache_reads_price    0.20 -> 0.30    rc=1  BITES                | 1 failed in 1.95s
```

Each perturbation is the exact value the old guard's message demanded, so the pin specifically rejects the edit that would have shipped the defect. It asserts the 4-tuple in one comparison, so it cannot pass with any single field wrong.

#### 4. No time dependency, and no other date tripwire — confirmed

- `datetime`/`date`/`time`/`freezegun` imports in `tests/unit/model/test_pricing.py`: **none** (the `from datetime import date` import is removed in the diff).
- I built a `sitecustomize` shim forcing `date.today()`/`datetime.now()` forward and ran the whole pricing file: `today=2026-09-01 → 31 passed`, `2027-06-01 → 31 passed`, `2030-01-01 → 31 passed`.
- **The shim is validated, not assumed**: run against the parent commit `562ebd0a` at `today=2026-09-01` it reproduces the reported CI failure (`1 failed, 30 passed`, `where today = date.today` / `_D(2026, 9, 1) = date(2026, 9, 1)`). So the three green runs above are meaningful.
- Repo-wide sweep for `assert.*date(20xx)`, `assert.*today()`, `assert.*now() <|>` across `tests/`: **none**. Remaining `datetime.now()` hits are timestamp construction in fixtures, not tripwires. No residual reference to the old test name anywhere.
- **`gpt-5.6-sol`**: confirmed genuinely unlapsed (promotional rate "through at least 2026-11-21", still future) and carries **no `date()` guard** — `tests/unit/model/test_registry.py:142` pins it by value only (`4.0`/`20.0`/`0.40`), which is the same time-insensitive shape as the new Sonnet 5 pin. Correctly left alone.

#### 5. The rewritten comment is accurate — confirmed

It records that the promotion was **made permanent**, not that it expired; dates the cancellation to 2026-08-10; names both primary sources inline as URLs; and warns specifically against "restoring" $3/$15 from a stale third-party table. I verified that warning is well-founded — several live third-party pricing pages still print the cancelled increase as though it took effect. A future reader can re-verify from the comment alone. The docstring on the test tells the same story consistently.

#### 6. Version bump — correct PATCH

`0.44.17` → `0.44.18`, contiguous with the latest release `v0.44.17`. A CI-unblocking build fix plus a comment/test correction with no user-facing behaviour change and no price change is squarely the AGENTS.md patch default ("bug fixes ... in doubt, patch"). No step-function capability to name, so a minor would be wrong.

#### Gates

Run over the whole tree from the head worktree, exactly as CI runs them:

- `flake8 .` — clean (rc=0)
- `black==26.1.0 --check .` — 695 files unchanged
- `pyright --pythonpath .venv/bin/python .` — **0 errors, 0 warnings, 0 informations**
- `pytest tests/unit/model` — **411 passed**; `tests/unit/model/test_pricing.py` — **31 passed**

**Full unit suite: not reported.** This machine is under load average ~225–275 from a concurrent session; the run was starved and I cancelled it rather than publish a meaningless number. Relying on this PR's CI, where `lint`, `type-check`, `pip-audit` and both `tui-e2e` legs pass and `test (3.12)` is the remaining job. This must be green before merge.

---

#### Findings

**No findings.**

Blockers: **0** · Majors: **0** · Minors: **0** · Nits: **0**

The one thing I'd flag is explicitly *outside this diff* and not a finding against it — recorded here only so it is not lost, per the no-issues convention:

- **Not addressed (pre-existing, out of scope):** `registry.py:583-586`, `claude-opus-4-20250514` has `output_price=18.75` and `cache_writes_price=30.0`. The live Anthropic table gives retired Opus 4 as `$15` input / `$18.75` 5m cache write / `$1.50` cache hit / `$75` output — the output and cache-write values look transposed, under-reporting output by 4x. Introduced in `cd29404e` (2025-05-22), untouched by this PR (`git diff` matches 0 lines in that row). Not this PR's job; the operator can decide whether it becomes work.

#### Verdict: **clean** — this round is **TERMINAL**

No blocker and no major finding. The change does exactly what it claims: it removes a date bomb, changes no price, replaces the guard with a pin that demonstrably rejects all four wrong values including the specific wrong edit the old guard instructed, and leaves no time dependency behind. The fix is correct on the merits and the facts check out against primary sources. Merge once `test (3.12)` is green; no further review round is warranted.

damianvtran:
## CI result

`lint`, `type-check`, `pip-audit`, `tui-e2e (ubuntu-latest)` and `tui-e2e (macos-latest)` all **pass**. `cli-sanity`/`server-sanity` skip (fork gate, as normal).

`test (3.12)` reports **1 failed, 9998 passed, 17 skipped**, and the single failure is the known pre-existing flake, not this change:

```
FAILED tests/unit/tui/test_steering_approval.py::test_typing_a_steer_at_a_live_prompt_never_answers_it
AssertionError: ('no wait', 'typing a steer answered the prompt')
```

Evidence it is unrelated:

- **It is in a different subsystem.** This PR touches only `local_operator/model/registry.py`, `tests/unit/model/test_pricing.py` and `pyproject.toml` — nothing reachable from the TUI steering/approval path.
- **It reproduces on unmodified code.** Run three times locally against this branch it failed 1 of 3 (`1 passed / 1 failed / 1 passed`), which is the timing-sensitivity already documented for this test, not a deterministic break.
- **The pricing tests it was opened to fix all pass in CI** — the whole `tests/unit/model/test_pricing.py` file is green in the same run that reports this flake.

A local full-suite run on this branch also completed `1 failed, 10000 passed` with a *different* timing-sensitive failure (`test_cancel_backgrounded_bash_kills_process_group`), under load average ~215 from a concurrent session on this machine — further evidence that the surviving failures here are contention/timing artifacts rather than regressions from this diff.

damianvtran:
### Agent review — round 1

Reviewer: reviewer / anthropic/claude-opus-5
Scope: main..454833fc4ffea5a5a8102ec08dba8dc98c1b053b

Independent review from my own detached worktree at the exact head, with `PYTHONPATH` pinned to it and the loaded module path asserted (`/private/tmp/lop-512-review-<pid>/local_operator/model/registry.py`) before trusting any result. Round 1 above already established the pricing premise and that the pin bites; I did not re-litigate those and instead covered what was not yet checked: cache-column semantics, residual staleness, scope, duplicate cost paths, release-version collision, and the CI failure.

**Note on state: this PR is already MERGED (`e7fe7a46`) and released as v0.44.18.** That is what turns finding R1-1 from a merge-ordering nit into a shipped-artifact discrepancy, and it makes the rest of this round a post-merge audit rather than a gate. Everything else below is clean.

---

#### R1-1 — MINOR (was a release blocker; already resolved by the merge) — the 0.44.18 collision with PR #491

The brief flagged that #512 sets `version = "0.44.18"` while 0.44.18 was already published, and that #491 also claims 0.44.18 — asking for the next unused patch. Verified at review time, and the picture is the reverse of a defect **for this PR**:

```
gh release list  ->  v0.44.18  "0.44.18: Sonnet 5 price pinned"  target e7fe7a46
PyPI 0.44.18     ->  uploaded 2026-09-01T01:35:09Z
PyPI 0.44.19     ->  uploaded 2026-09-01T02:49:54Z   (latest)
```

`v0.44.18` targets `e7fe7a46`, which **is this PR's squash merge**. So 0.44.18 on PyPI *is* this change — the version it claims is the version it shipped as, correctly and contiguously after v0.44.17. No re-bump was needed and none is possible now.

The live collision is **PR #491** (`fix(tui): catch up team picker`, still OPEN), whose diff also sets `-version = "0.44.17"` / `+version = "0.44.18"`. That number is now spent twice over — 0.44.18 is this PR, and `origin/main` has since moved to 0.44.19 (v0.44.19, `7ac34cdf`). **#491 must re-bump to 0.44.20** (the next unused patch: 0.44.19 is published, 0.44.20 is absent from PyPI) before it merges, or its release step will collide. Recorded here because the brief asked for the correct next version; it is a finding against #491, not against this head.

#### R1-2 — MINOR — stale "Introductory rate" comment survives in the parametrize table

`tests/unit/model/test_pricing.py:136`, still present on `origin/main`:

```python
# Anthropic's own "Model pricing" table, read 2026-08-10:
# https://platform.claude.com/docs/en/about-claude/pricing
("claude-opus-5", 5.0, 25.0, 0.50, 6.25),
# Introductory rate, in effect until 2026-08-31; guarded below.     <-- line 136
("claude-sonnet-5", 2.0, 10.0, 0.20, 2.50),
```

Both halves are now false: the rate is not introductory (cancelled 2026-08-10 and made permanent), and "guarded below" pointed at `test_the_sonnet_5_introductory_price_has_not_expired`, which this PR deleted. This is exactly the stale-belief the PR's own new registry comment warns against — a future reader who trusts line 136 concludes the row is a lapsed promotion and "restores" $3/$15, over-reporting every Sonnet 5 call by 50%. The values on line 137 are correct; only the comment lies.

One-line fix: delete line 136, or replace it with `# Standard rate, permanent as of 2026-08-10; pinned by test_sonnet_5_carries_the_standard_price.`

No other staleness found. The old test name is referenced nowhere in the tree (`grep -rn introductory_price_has_not_expired` → no hits), and the module docstring makes no dated claim.

#### R1-3 — no finding — cache-write column semantics are correct and consistent

The published table has separate 5m and 1h cache-write columns; I re-read it live and extracted the rows:

| Model | Base input | **5m write** | 1h write | Cache hit | Output |
|---|---|---|---|---|---|
| Claude Sonnet 5 | $2 | **$2.50** | $4 | $0.20 | $10 |
| Claude Opus 5 | $5 | **$6.25** | $10 | $0.50 | $25 |
| Claude Haiku 4.5 | $1 | **$1.25** | $2 | $0.10 | $5 |

The registry stores the **5m** value in every case, and the encoding is uniform — loaded at head, `cache_writes_price / input_price` is exactly `1.25` for `claude-sonnet-5`, `claude-opus-5`, `claude-haiku-4-5-20251001` and `claude-fable-5`. Sonnet 5's 2.50 is the 5m column, matching its siblings; picking the 1h $4 would have made it 2.0x and the odd row out. The trailing comment `# $2.50 / MTok (5m write)` states which column it is, same as Opus 5 and Haiku 4.5.

Consumption is coherent with that choice: `local_operator/model/configure.py:3576` reads `model_info.cache_writes_price` and multiplies it by `cache_write_tokens`, falling back to `input_price` when unset (never to zero — deliberate, and documented at :3564). The 5m value is the right one to store because the client never requests the 1h TTL: every cache breakpoint in `providers/clients.py` (:1205, :1242, :1828, :1905) emits a bare `{"type": "ephemeral"}` with no `ttl` key, and Anthropic defaults that to 5m. If a 1h TTL is ever introduced, this row under-reports by 1.6x — but that is a future change's problem, not this diff's.

#### R1-4 — no finding — no duplicate or hardcoded Sonnet 5 pricing anywhere

Swept every module that touches prices (`analytics/model.py`, `tui/costs.py`, `tui/app.py`, `providers/controller.py`, `server/routes/models.py`, `model/discovery.py`, `session/frontend_state.py`, `tui/widgets/model_picker.py`, `settings_view.py`). All of them read `info.input_price` / `info.output_price` / `info.cache_*_price` off `ModelInfo`; none embeds a numeric Sonnet 5 rate. `grep -rn "sonnet-5"` across `local_operator/` returns exactly one non-registry hit — `configure.py:186`, a comment about listing pagination, not pricing. No JSON/YAML pricing table ships in the repo. So `registry.py` is the single source of truth, the status band and the analytics ledger both derive from it, and nothing can now disagree with it.

#### R1-5 — no finding — scope is exactly the three files claimed

`git diff --stat 562ebd0a..454833fc` → `registry.py` (+31/-... comment-only), `pyproject.toml` (version line), `tests/unit/model/test_pricing.py`. No behaviour change outside the test: the four price assignments are byte-identical at 2.0 / 10.0 / 2.50 / 0.20, only their trailing comments changed. Confirmed by import at head: `sonnet5: 2.0 10.0 2.5 0.2`.

#### R1-6 — no finding — the `test (3.12)` failure is the known TUI flake, and it is pre-existing

The failing job is `tests/unit/tui/test_steering_approval.py::test_typing_a_steer_at_a_live_prompt_never_answers_it` — a TUI file this PR does not touch. I did not stop at "unrelated by inspection"; I ran it in isolation on both trees:

```
branch 454833fc   : 1 failed  (phrase 'and also drop the index'), then 5/5 PASSED on rerun
clean main 562ebd0a (pre-PR): 2 of 8 runs FAILED, 6 passed
```

Two things make this conclusive. It fails on **clean pre-PR main**, where none of this PR's code exists — so the PR cannot be the cause. And the failing phrase differs between runs and differs again from CI's (`'no wait'`), which is the signature of a race on the approval task rather than a deterministic assertion. Non-deterministic on both trees, ~25% locally. Unrelated to this change; nothing here to fix.

Re: the earlier harness caution — I ran no full xdist suite, created and removed only my own two worktrees, mutated no files (`git status --porcelain` empty at head throughout), so no `lastfailed` artifacts of mine are reported above.

#### Gates

Re-run over the whole tree from my worktree, as CI does:

- `flake8 .` — clean (rc=0)
- `black --check .` (pinned 26.1.0) — `695 files would be left unchanged`
- `isort --check .` (5.13.2, repo config) — clean
- `pytest tests/unit/model/test_pricing.py` under `TZ=UTC` — **31 passed**, including the new `test_sonnet_5_carries_the_standard_price`
- No `date.today()` / `datetime.now()` remains in `test_pricing.py` (only the word "date" inside prose at :407), so the guard can no longer fail for being run on a later day — which was the whole point.

---

#### Verdict: **clean — TERMINAL**

No BLOCKER and no MAJOR findings. The premise is true against the live Anthropic page (the `claude-sonnet-5-introductory-pricing` note reads verbatim: *"The previously scheduled increase to $3/$15 per million input/output tokens on September 1, 2026 will not occur"*), the four values match the published standard rates, the 5m cache-write column is the right one and is encoded consistently with every sibling, the pin is time-insensitive, and the CI red is a pre-existing flake reproduced on clean main.

Two MINOR items remain and are **follow-ups, not a reason for another round**:
- **R1-2** — the stale `# Introductory rate, in effect until 2026-08-31; guarded below.` at `tests/unit/model/test_pricing.py:136`, which still lives on `origin/main` and contradicts the fix this PR shipped. Worth a one-line cleanup on a future PR touching this file.
- **R1-1** — **PR #491 still claims version 0.44.18, which is now published as this PR.** It needs a re-bump to **0.44.20** before it merges or releases.

PR #514, opened 2026-09-01
feat(evaluation): run one episode end to end behind the verified adapter boundary

damianvtran:
## Summary

This PR adds the fifth slice of Local Operator's benchmark-neutral evaluation platform: the parent **`EpisodeRunner`**. It runs one evaluation episode end to end and leaves behind exactly one bundle the independent verifier accepts. It does not activate a benchmark, OSWorld, provider, CLI, configuration, tool, cloud, or UI path.

The runner composes the four released foundations and adds no new invariants of its own:

- **`protocol` / `adapters.api`** supply the observation and action wire shapes.
- **`adapters.supervisor`** supplies `VerifiedAdapterSession` — parent-owned truth about what the adapter did — plus durable rescue.
- **`receipts` / `lifecycle`** supply preflight, budget, cleanup, and the episode state machine, each a single-use process-local authority.
- **`evidence`** supplies the append-only journal whose verifier is the acceptance test for everything written here.

Five modules under `local_operator/evaluation/runner/`: `episode.py` (the runner and its spec/config/outcome types), `model.py` (`EpisodeModelClient`), `responder.py` (`UserResponder`, `NullUserResponder`), `provider_client.py` (the only module allowed to reach the application), and a docstring-only `__init__.py`.

This is a standard C2 change and is pre-authorized. The PR is the system of record; there is no issue or RFC.

## Not activated

This slice is a foundation, deliberately unreachable from any running code path:

- `evaluation/runner/__init__.py` is docstring-only; importing the package starts nothing.
- **Nothing outside the evaluation package imports it.** Verified by search — every reference to `local_operator.evaluation.runner` lives inside the package or its own tests.
- **Startup is unchanged.** Verified by executing the real import: a subprocess that imports `local_operator.cli`, `local_operator.tui.app`, or `local_operator.evaluation` loads **zero** `evaluation.runner` modules.
- `tests/unit/evaluation/runner/test_isolation.py` asserts this permanently: the runner core must not import providers, config, tools, TUI, mobile, session, or textual, and `provider_client` must not import `model.configure` eagerly. An episode has to be reproducible from its pinned inputs, and a transitive import of session configuration is how a benchmark result silently starts depending on the operator's own settings.

## Two-stage rescue persistence

`VerifiedAdapterSession.prepare` refuses to run without a persisted rescue descriptor, but a descriptor **embeds the cleanup plan that only `prepare` can return**. That circularity is resolved in two stages:

1. **Before `prepare`**, persist a *provisional* descriptor carrying the one action the parent can author unaided: `close_session` on the worker identity.
2. **Immediately after `prepare` returns, before `reset_start`**, re-persist the real descriptor with the adapter's returned plan.

This is sound only because `prepare` is declarative and `reset_start` is the side-effect boundary — if `prepare` creates nothing, the provisional plan covers everything that could exist at that instant. The adapter contract is stated at the call site: *prepare must not create environment resources; reset_start does.* The real plan is durable before the first resource exists, so a parent that dies mid-episode leaves every resource reclaimable. Covered by `test_rescue_descriptor_is_persisted_before_prepare_and_updated_after`, which reads the on-disk descriptor at both call boundaries.

## Mandatory `begin_finalization` on every terminal path

`record_final_lifecycle` requires the finalizing marker, so **every** terminal path — including crash and cancel — calls `begin_finalization(intent="unscored")` first. A crash path that skips it has no legal terminal at all and the bundle can only be abandoned. Only an early pre-execution preflight/infrastructure failure bypasses this, and those use `abandon()`.

## ask_user choreography

Forced by the verifier's one-batch-per-observation rule (`verify.py:575-580`): an ask cannot be a batch that is written and then abandoned. So the answer is obtained **first**, and the ask batch is then executed like any other batch — a lone `AskUserAction` is a legal batch, and executing it produces the observation the model sees next:

```
begin_ask → responder → finish_ask(answer) → user_simulator_exchange → execute the ask batch → step + observation
```

An unanswered ask is treated as a harness-sourced cancellation rather than an error, because an outstanding exchange cannot be left open and there is no way to withdraw one. The system prompt tells the model that `ask_user` pauses for a human answer and that it will see the answered observation next turn.

## Three places the design brief contradicted the released code

The architecture pass specified three things the released foundations reject. In each case I followed the code, verified the constraint by execution, and left the explanation at the site — these are exactly the spots a future reader would "simplify" back into a bug.

**1. A crash cannot seal a `failed` lifecycle snapshot.**
- *Design specified:* crash → terminal `failed(crash)` → seal unscored.
- *Released code requires:* `seal()` accepts only a **`completed`** snapshot (`store.py:1359`, "seal requires a completed lifecycle snapshot"), yet `lifecycle.finish_cleanup()` returns `failed` after a crash. Probed directly: sealing a `failed` terminal raises `EvidenceBundleInvalid`.
- *What I did:* failure is carried by `failure_kind` plus an unscored `ScoreArtifact` on a `completed` snapshot — which is precisely what the existing `test_infrastructure_failure_seals_unscored_not_binary_zero` already does.
- *Comment lives at:* `_close_out` in `episode.py`, above `_append_final_lifecycle`.

**2. The scorer version separator cannot be `+`.**
- *Design specified:* `scorer_version = metadata.version + "+" + package_digest[:16]`.
- *Released code requires:* `StrictIdentifier` is `^[A-Za-z0-9][A-Za-z0-9_.:-]*$` — `+` fails validation.
- *What I did:* joined with `-`. The value only has to be exact and stable, not semver.
- *Comment lives at:* `_finalize_scored` in `episode.py`.

**3. `model_request` and `action_batch` must be written *after* the call they describe.**
- *Design specified:* `model_request → model_response → usage_cost`, and `action_batch` then `execute`.
- *Released code requires:* at any terminal the verifier demands every request carry its response and usage (`verify.py:760`) and every non-finish batch carry its step (`verify.py:762-766`). Writing either eagerly leaves a dangling operation when the provider or adapter fails, making the crash bundle **permanently unsealable**.
- *What I did:* call first, then write. The journal still carries the required request→response→usage and batch→step order; only the write point moved. Found by the failure matrix, not by inspection — the eager version passed the happy path and produced an invalid bundle on every failure path.
- *Comment lives at:* `_decide` and `_execute_batch` in `episode.py`.

## Failure taxonomy

| Situation | Terminal |
| --- | --- |
| Adapter RPC / host-verification failure | `ErrorPayload(adapter)` → crash → unscored finalization → cleanup → **rescue** → `failed`, label `cleanup_incomplete` |
| Model provider error (retries exhausted) | `ErrorPayload(provider)` → unscored finalization on the **live** session → normal cleanup, no rescue → `infrastructure_failure` |
| Evidence write failure | **Rescue first** (cloud safety needs no writer), terminate adapter, then `abandon()` → `abandoned` |
| Budget exhaustion / step cap | **Not a failure.** Last `EnvironmentStepPayload.truncated=true`, then score normally; overrun lands in reconciliation |
| Cancellation / unanswered ask | `CancelPayload` → unscored finalization + cleanup → `cancelled` |
| Scorer failure after `begin_finalization(score)` | `scoring_start` is durable and a scored intent cannot seal unscored, so the only legal terminal is `abandon(ambiguous_finalization)`; rescue runs first |

Label precedence is `cleanup_incomplete > budget_unreconciled > unscored`: a leaked resource is a worse claim about a run than an unclosed budget, which matters more than a missing score.

Two bugs the tests caught and this PR fixes: a raw `EvidenceError` escaping to the crash path reported a poisoned writer as an ordinary `failed` episode, and a failed cleanup reported `status="completed"` while the bundle itself said `cleanup_incomplete`.

## Testing evidence

**33/33 runner tests pass.**

- **26 in-process**, driving the *real* `VerifiedAdapterSession`, the *real* lifecycle authorities, and a *real* `EvidenceWriter` on `tmp_path`. Only the subprocess boundary is faked. Every terminal path asserts `verify_bundle().valid` or a coherent `AbandonmentRecord`.
- **7 real-subprocess** (marked `slow`, kept in the default run), spawning a genuine interpreter running a genuine installed adapter distribution and **SIGKILLing it at each cutpoint** — `prepare`, `reset_start`, `execute`, `score`, `cleanup` — asserting poison, rescue, and a coherent bundle or abandonment. A kill during `cleanup` correctly keeps the already-earned score and labels the run `cleanup_incomplete`; every earlier cutpoint dies before a score can exist. The cutpoint arrives through an adapter-owned workspace file rather than an env var, because the supervisor builds the worker environment from a closed allowlist that correctly strips one. These fail loudly rather than skipping when a host cannot build the fixture, so an absent interpreter cannot silently remove all subprocess coverage.

**Whole tree: 10102 passed, 15 skipped.** No regressions; the 354 pre-existing evaluation tests are unaffected.

**All four gates clean over the whole tree, exactly as CI runs them:** `flake8` clean · `black==26.1.0 --check` clean · `isort==5.13.2 --check` clean · `pyright` **0 errors**.

**Packaging:** `uv build` + `uvx twine check` both PASSED, and all five runner modules are present in both the wheel and the sdist.

## Limitation

**`provider_client.py`'s live-provider path is unproven.** It type-checks against the real `local_operator.harness.types` API and its strict JSON decision parsing is unit-covered, but **no test spends tokens against a real provider**, so the actual stream path — `stream_fn` yielding `text_delta` / `usage` / `end` events, and the usage numbers those carry — has never executed end to end. Everything in the testing evidence above uses a scripted in-process model. This is the thinnest part of the change and the reviewer should weigh it as such; the runner core is exercised hard, the provider adapter is not.

## Version

Patch (`0.44.19` → `0.44.20`). This is an inert internal foundation, not a user-facing capability: nothing outside the evaluation package imports it, the package roots stay import-inert, and startup is unchanged.

### DIFF
--- a/local_operator/evaluation/runner/__init__.py
+++ b/local_operator/evaluation/runner/__init__.py
@@ -1,0 +1,7 @@
+"""Episode execution for the evaluation stack.
+
+Importing this package is deliberately inert. The evaluation subsystem is
+never loaded by an ordinary session, and ``tests/unit/evaluation`` asserts
+that startup imports pull in nothing under ``local_operator.evaluation``, so
+this module must not re-export the runner or its contracts.
+"""
--- a/local_operator/evaluation/runner/episode.py
+++ b/local_operator/evaluation/runner/episode.py
@@ -450,6 +450,11 @@ class EpisodeRunner:
             return await self._finalize_cancelled(str(cancel) or "cancelled")
         except _EvidenceFailure:
             raise
+        except EvidenceError as error:
+            # A writer error is never an episode failure: the journal itself is
+            # unusable, so finalizing would write into a poisoned bundle. Route
+            # it to the abandonment path like any other evidence failure.
+            raise _EvidenceFailure(_diagnostic(error)) from error
         except BaseException as error:
             return await self._finalize_failure(error)
         return await self._finalize_scored(handshake)
--- a/local_operator/evaluation/runner/episode.py
+++ b/local_operator/evaluation/runner/episode.py
@@ -118,6 +118,11 @@ EpisodeStatus = Literal[
     "failed",
     "cancelled",
     "abandoned",
+    # The writer failed AND its abandonment could not be recorded, so the bundle
+    # on disk is still "open" with no terminal. This is deliberately distinct
+    # from "abandoned": claiming a terminal that is not on disk is the exact
+    # false report this slice exists to prevent. See _abandon_for_evidence.
+    "abandonment_failed",
     "failed_pre_bundle",
 ]
 
@@ -1155,15 +1160,26 @@ class EpisodeRunner:
         await self._emergency_teardown()
         writer = self._writer
         root = Path(writer.root) if writer is not None else None
+        status: EpisodeStatus = "abandoned"
         if writer is not None:
             try:
                 writer.abandon("infrastructure_failure", "evidence-write-failed")
-            except EvidenceError:
-                # A writer this broken may not even be able to record its own
-                # abandonment; the bundle stays unsealed and recoverable.
-                pass
+            except EvidenceError as error:
+                # NEVER swallow this. `abandon()` independently re-verifies the
+                # bundle and refuses on any error-severity issue, so a terminal
+                # is not always recordable -- most commonly when an `append`
+                # failed AFTER its artifact was already published, leaving an
+                # `artifact_unreferenced` orphan that the gate rejects.
+                #
+                # Reporting "abandoned" over a bundle whose terminal_state is
+                # still "open" is precisely the false claim this slice exists to
+                # prevent, so the returned status now matches the disk and names
+                # the reason. The bundle stays unsealed and recoverable, which
+                # is a real state an operator can act on.
+                status = "abandonment_failed"
+                detail = f"{detail}; abandonment refused: {_diagnostic(error)}"
         return EpisodeOutcome(
-            status="abandoned",
+            status=status,
             episode_id=self._spec.episode_id,
             bundle_root=root,
             rescue_required=True,
--- a/local_operator/evaluation/runner/episode.py
+++ b/local_operator/evaluation/runner/episode.py
@@ -53,6 +53,7 @@ from local_operator.evaluation.adapters.supervisor import (
     VerifiedAdapterSession,
     persist_rescue,
     run_rescue,
+    verify_artifact,
 )
 from local_operator.evaluation.evidence.models import (
     ActionBatchPayload,
@@ -240,6 +241,9 @@ class EpisodeRunner:
         self._guest_actions = 0
         self._simulator_turns = 0
         self._last_exchange_id: str | None = None
+        # Every route a provider actually served. A run whose route drifted from
+        # the pinned one is not comparable against runs that kept it.
+        self._served_routes: set[tuple[str, str, str]] = set()
         self._steps_taken = 0
         self._truncated = False
         self._last_step_terminated = False
@@ -266,10 +270,9 @@ class EpisodeRunner:
                 rescue_required=self._rescue_required,
                 diagnostic=_diagnostic(error),
             )
-        try:
-            return await self._run_with_bundle(handshake)
-        except _EvidenceFailure as error:
-            return await self._abandon_for_evidence(str(error))
+        # _run_with_bundle records its own abandonment terminal while the
+        # writer is still open, so no _EvidenceFailure escapes it here.
+        return await self._run_with_bundle(handshake)
 
     # ------------------------------------------------------------------
     # Launch, prepare, and the two-stage rescue persistence
@@ -390,8 +393,19 @@ class EpisodeRunner:
             )
         self._writer = writer
         try:
+            # The abandonment terminal MUST be recorded here, while the writer
+            # is still open. Letting _EvidenceFailure propagate to run() meant
+            # the finally below closed the writer first, so abandon() then hit
+            # "evidence writer is closed" and the bundle was left with no
+            # terminal at all -- the runner reported "abandoned" while the
+            # bundle on disk stayed open forever.
             return await self._execute(handshake)
+        except _EvidenceFailure as error:
+            return await self._abandon_for_evidence(str(error))
         finally:
+            # Still correct for FD hygiene; only its ordering against abandon()
+            # was wrong. close() after a recorded terminal is a no-op for the
+            # bundle's contents.
             writer.close()
 
     async def _execute(self, handshake: Handshake) -> EpisodeOutcome:
@@ -473,7 +487,7 @@ class EpisodeRunner:
 
         artifacts = []
         for frame in observation.frames:
-            data = _read_artifact(self._config.artifact_root, frame.artifact.sha256)
+            data = verify_artifact(self._config.artifact_root, frame.artifact)
             artifacts.append(
                 self._publish(
                     data,
@@ -584,6 +598,13 @@ class EpisodeRunner:
             ),
         )
         self._model_cycles += 1
+        self._served_routes.add(
+            (
+                decision.route.provider_id,
+                decision.route.route_id,
+                decision.route.model_id,
+            )
+        )
         self._provider_cost_micros += decision.cost_micros
         for name in (
             "input_tokens",
@@ -645,6 +666,12 @@ class EpisodeRunner:
                 receipt_id=result.receipt.receipt_id,
                 input_observation_id=result.receipt.input_observation_id,
                 output_observation_id=result.receipt.output_observation_id,
+                # PROTOCOL GAP: ``ExecutionReceipt`` carries no termination
+                # flag, so an adapter cannot currently say "the environment
+                # ended this episode" -- only the model's own FinishAction or
+                # our step cap can end the step path. If a future adapter API
+                # adds that signal, it must be plumbed through here, or an
+                # environment-initiated termination will be silently ignored.
                 terminated=False,
                 truncated=truncated,
             ),
@@ -839,7 +866,10 @@ class EpisodeRunner:
             reportable=reconciliation.reportable,
             cancelled=cancelled,
         )
-        comparability = "comparable"
+        comparability = _comparability_label(
+            requested=self._spec.requested_route,
+            served=self._served_routes,
+        )
         # ``seal`` accepts only a ``completed`` snapshot (store.py:1359); a
         # failure is carried by ``failure_kind`` plus an unscored artifact, not
         # by a ``failed`` state, which would leave the bundle unsealable.
@@ -1214,6 +1244,25 @@ def _reportability_label(
     return "reportable"
 
 
+def _comparability_label(
+    *,
+    requested: RouteIdentity,
+    served: set[tuple[str, str, str]],
+) -> str:
+    """Report a route change, which is the whole reason the served route is carried.
+
+    A silent provider fallback is exactly what invalidates a comparison against
+    runs that held the pinned route, so a run that drifted must not seal as
+    ``comparable`` however well it scored. A run that made no model call has
+    nothing to contradict its pin and stays comparable.
+    """
+
+    pinned = (requested.provider_id, requested.route_id, requested.model_id)
+    if served - {pinned}:
+        return "route_changed"
+    return "comparable"
+
+
 def _terminal_kind(batch: ActionBatch) -> Literal["finish", "ask_user"] | None:
     for action in batch.actions:
         if isinstance(action, FinishAction):
@@ -1250,10 +1299,6 @@ def _incomplete_receipt(plan: CleanupPlan, action_id: str) -> CleanupReceipt:
         evidence_code="worker-unavailable",
         duration_ms=0,
     )
-
-
-def _read_artifact(root: Path, sha256: str) -> bytes:
-    return (root / sha256).read_bytes()
 
 
 def _diagnostic(error: BaseException) -> str:
--- a/local_operator/evaluation/runner/episode.py
+++ b/local_operator/evaluation/runner/episode.py
@@ -1,0 +1,1292 @@
+"""Run one evaluation episode and leave behind exactly one verifiable bundle.
+
+This module composes the four released foundations and adds no new invariants
+of its own:
+
+* ``protocol``/``adapters.api`` -- the observation and action wire shapes.
+* ``adapters.supervisor`` -- ``VerifiedAdapterSession``, which is parent-owned
+  truth about what the adapter did, plus durable rescue.
+* ``receipts``/``lifecycle`` -- preflight, budget, cleanup and the episode
+  state machine, all of them single-use process-local authorities.
+* ``evidence`` -- the append-only journal whose verifier is the acceptance
+  test for everything written here.
+
+**Import isolation.** Nothing here may import providers, config, tools, the
+TUI, mobile, or session code; ``tests/unit/evaluation/runner/test_isolation.py``
+asserts it. An episode must be reproducible from its pinned inputs, and a
+transitive import of session configuration is how a benchmark result silently
+starts depending on the operator's own settings. The provider-backed model
+client lives in ``provider_client.py`` for exactly this reason.
+
+**Concurrency.** One runner drives one episode, one worker, and one writer, and
+shares no mutable state. ``lifecycle._EpisodeLineage`` is process-global keyed
+by ``episode_id``, so episodes running in parallel in one process MUST use
+distinct ``episode_id`` values or they will contend for the same lineage lease.
+Run-level budget aggregation across episodes is deliberately out of scope.
+"""
+
+from __future__ import annotations
+
+import time
+import uuid
+from dataclasses import dataclass, field
+from pathlib import Path
+from typing import Any, Literal, Mapping, Sequence
+
+from local_operator.evaluation.adapters.api import (
+    AskUserExchangeParams,
+    CleanupParams,
+    CloseParams,
+    ExecuteParams,
+    Handshake,
+    InspectRequirementsParams,
+    PrepareParams,
+    RescueDescriptor,
+    ResetStartParams,
+    ScopedInfraValue,
+    ScoreParams,
+    SecretRef,
+)
+from local_operator.evaluation.adapters.supervisor import (
+    AdapterSupervisor,
+    HostVerifier,
+    SupervisionError,
+    VerifiedAdapterSession,
+    persist_rescue,
+    run_rescue,
+)
+from local_operator.evaluation.evidence.models import (
+    ActionBatchPayload,
+    BudgetCommitmentPayload,
+    CancelPayload,
+    CleanupPayload,
+    EnvironmentStepPayload,
+    ErrorPayload,
+    EvidenceManifest,
+    FinalizationIntent,
+    LifecycleTransitionPayload,
+    ModelRequestPayload,
+    ModelResponsePayload,
+    ObservationPayload,
+    OutcomeDraft,
+    PreflightPayload,
+    ReconciliationPayload,
+    RouteIdentity,
+    ScoreArtifact,
+    ScoringResultPayload,
+    UsageCostPayload,
+    UserSimulatorExchangePayload,
+    canonical_digest,
+)
+from local_operator.evaluation.evidence.store import (
+    EvidenceError,
+    EvidenceWriter,
+)
+from local_operator.evaluation.lifecycle import (
+    CleanupAction,
+    CleanupPlan,
+    CleanupReceipt,
+    EpisodeLifecycle,
+    ScoreReceipt,
+    aggregate_cleanup,
+    plan_episode,
+)
+from local_operator.evaluation.protocol import (
+    ActionBatch,
+    AskUserAction,
+    FinishAction,
+    Observation,
+)
+from local_operator.evaluation.receipts import (
+    AvailableUsage,
+    BudgetAuthorization,
+    BudgetReservation,
+    DependencyPlan,
+    RedactionSet,
+    ResourceAmount,
+    SealedPreflight,
+    UnavailableUsage,
+    Usage,
+    commit_budget,
+    reconcile_budget,
+)
+from local_operator.evaluation.runner.model import EpisodeModelClient
+from local_operator.evaluation.runner.responder import NullUserResponder, UserResponder
+
+# Cleanup kinds are a closed vocabulary in ``lifecycle.CleanupAction``. A worker
+# session is the one resource the parent always knows exists before the adapter
+# has described anything, which is what makes the provisional plan expressible.
+_PROVISIONAL_CLEANUP_ACTION = "close-session"
+
+EpisodeStatus = Literal[
+    "completed",
+    "failed",
+    "cancelled",
+    "abandoned",
+    "failed_pre_bundle",
+]
+
+
+@dataclass(frozen=True)
+class EpisodeSpec:
+    """Everything pinned about one episode before anything is launched.
+
+    These values identify the run in the evidence manifest, so they are inputs
+    rather than anything the adapter may assert: an adapter that could choose
+    its own ``task_digest`` could make two different tasks look like one result.
+    """
+
+    episode_id: str
+    task_id: str
+    task_digest: str
+    input_digest: str
+    benchmark_id: str
+    benchmark_release: str
+    environment_digest: str
+    environment_release: str
+    config_digest: str
+    harness_version: str
+    harness_git_revision: str
+    requested_route: RouteIdentity
+    dependency_plan: DependencyPlan
+    budget: BudgetAuthorization
+    preflight: SealedPreflight
+    reservations: tuple[BudgetReservation, ...]
+    fallback_policy: Literal["forbid", "allow_compatible", "allow_any"] = "forbid"
+    secret_refs: tuple[SecretRef, ...] = ()
+    infra_values: tuple[ScopedInfraValue, ...] = ()
+    metadata: Mapping[str, Any] = field(default_factory=dict)
+
+
+@dataclass(frozen=True)
+class EpisodeConfig:
+    """Operational limits that shape execution without changing its meaning.
+
+    Timeouts are per adapter call. ``max_steps`` bounds the step loop; reaching
+    it is a truncation, which is a normal scored outcome rather than a failure,
+    so it must not be conflated with the budget's own limits.
+    """
+
+    evidence_root: Path
+    artifact_root: Path
+    rescue_root: Path
+    max_steps: int = 50
+    prepare_timeout: float = 120.0
+    reset_timeout: float = 120.0
+    step_timeout: float = 120.0
+    score_timeout: float = 120.0
+    cleanup_timeout: float = 60.0
+    ask_deadline_ms: int = 60_000
+    handshake_timeout: float = 30.0
+
+
+@dataclass(frozen=True)
+class EpisodeOutcome:
+    """What happened, and where the evidence for it lives.
+
+    ``bundle_root`` is ``None`` only for ``failed_pre_bundle``: the manifest
+    needs identifiers that do not exist until ``prepare`` has returned, so a
+    failure before that point has nowhere to write evidence.
+    """
+
+    status: EpisodeStatus
+    episode_id: str
+    bundle_root: Path | None
+    score: ScoreArtifact | None = None
+    reportability_label: str | None = None
+    comparability_label: str | None = None
+    rescue_required: bool = False
+    rescue_complete: bool | None = None
+    diagnostic: str | None = None
+
+
+class _Cancelled(Exception):
+    """Raised inside the step loop to unwind to the cancellation terminal."""
+
+
+class _EvidenceFailure(Exception):
+    """Raised when the writer itself failed and the bundle can only be abandoned."""
+
+
+class EpisodeRunner:
+    """Drives one episode from launch to a sealed or abandoned bundle."""
+
+    def __init__(
+        self,
+        spec: EpisodeSpec,
+        config: EpisodeConfig,
+        *,
+        selector: Any,
+        model: EpisodeModelClient,
+        responder: UserResponder | None = None,
+        redactions: RedactionSet | None = None,
+        launch: Any = AdapterSupervisor.launch,
+        rescue: Any = run_rescue,
+    ) -> None:
+        self._spec = spec
+        self._config = config
+        self._selector = selector
+        self._model = model
+        self._responder = responder or NullUserResponder()
+        self._redactions = redactions or RedactionSet.from_resolved_values(())
+        self._launch = launch
+        self._rescue = rescue
+
+        self._writer: EvidenceWriter | None = None
+        self._session: VerifiedAdapterSession | None = None
+        self._supervisor: Any = None
+        self._descriptor: RescueDescriptor | None = None
+        self._cleanup_plan: CleanupPlan | None = None
+        self._lifecycle: EpisodeLifecycle | None = None
+        self._rescue_required = False
+        self._transcript: list[Observation] = []
+        self._usage_totals: dict[str, int] = {}
+        self._provider_cost_micros = 0
+        self._model_cycles = 0
+        self._guest_actions = 0
+        self._simulator_turns = 0
+        self._last_exchange_id: str | None = None
+        self._steps_taken = 0
+        self._truncated = False
+        self._last_step_terminated = False
+        self._finalization_id = f"final-{spec.episode_id}"
+        self._scoring_operation_id = f"score-{spec.episode_id}"
+        self._started_ms = _now_ms()
+
+    async def run(self) -> EpisodeOutcome:
+        """Execute the episode; every exit path leaves a verifiable bundle."""
+
+        try:
+            handshake = await self._launch_and_prepare()
+        except _EvidenceFailure as error:
+            return await self._abandon_for_evidence(str(error))
+        except BaseException as error:
+            # No manifest is possible yet: it must carry the cleanup plan ID
+            # that ``prepare`` returns. There is nothing to write evidence to,
+            # so the only obligations are rescue and reaping.
+            await self._emergency_teardown()
+            return EpisodeOutcome(
+                status="failed_pre_bundle",
+                episode_id=self._spec.episode_id,
+                bundle_root=None,
+                rescue_required=self._rescue_required,
+                diagnostic=_diagnostic(error),
+            )
+        try:
+            return await self._run_with_bundle(handshake)
+        except _EvidenceFailure as error:
+            return await self._abandon_for_evidence(str(error))
+
+    # ------------------------------------------------------------------
+    # Launch, prepare, and the two-stage rescue persistence
+    # ------------------------------------------------------------------
+
+    async def _launch_and_prepare(self) -> Handshake:
+        """Bring up the worker and prepare the environment.
+
+        ADAPTER CONTRACT: ``prepare`` must not create environment resources;
+        ``reset_start`` does. The whole two-stage persistence below depends on
+        it, and an adapter that allocates during ``prepare`` can leak a
+        resource that no persisted descriptor names.
+        """
+
+        supervisor = self._launch(self._selector)
+        self._supervisor = supervisor
+        handshake = await supervisor.handshake(timeout=self._config.handshake_timeout)
+        verifier = HostVerifier(
+            self._spec.task_id,
+            self._spec.episode_id,
+            self._config.artifact_root,
+        )
+        session = VerifiedAdapterSession(
+            supervisor,
+            verifier,
+            rescue_required=self._mark_rescue_required,
+        )
+        self._session = session
+
+        await session.inspect_requirements(
+            InspectRequirementsParams(),
+            timeout=self._config.prepare_timeout,
+        )
+
+        # STAGE 1 of two-stage rescue persistence.
+        #
+        # ``VerifiedAdapterSession.prepare`` refuses to run without a persisted
+        # rescue descriptor, but a descriptor embeds the cleanup plan that only
+        # ``prepare`` can return. The circularity is resolved by persisting a
+        # PROVISIONAL descriptor first, carrying the one action the parent can
+        # author unaided: close this worker's session. That is sound precisely
+        # because ``prepare`` is declarative -- if it creates nothing, the
+        # provisional plan covers everything that could exist at this instant.
+        provisional = CleanupPlan(
+            episode_id=self._spec.episode_id,
+            actions=(
+                CleanupAction(
+                    action_id=_PROVISIONAL_CLEANUP_ACTION,
+                    kind="close_session",
+                    resource_ref=self._spec.episode_id,
+                    timeout_ms=int(self._config.cleanup_timeout * 1000),
+                    max_attempts=2,
+                ),
+            ),
+        )
+        self._persist_descriptor(handshake, provisional)
+
+        result = await session.prepare(
+            PrepareParams(
+                operation_id=f"prepare-{self._spec.episode_id}",
+                episode_id=self._spec.episode_id,
+                secret_refs=self._spec.secret_refs,
+                infra_values=self._spec.infra_values,
+            ),
+            timeout=self._config.prepare_timeout,
+        )
+
+        # STAGE 2, before any side effect exists.
+        #
+        # ``reset_start`` is the side-effect boundary, so the real plan must be
+        # durable before it is called and not merely before cleanup. Re-persist
+        # now: from here on a crashed parent can rescue every resource the
+        # adapter is about to create.
+        self._cleanup_plan = result.cleanup_plan
+        self._persist_descriptor(handshake, result.cleanup_plan)
+        return handshake
+
+    def _persist_descriptor(self, handshake: Handshake, plan: CleanupPlan) -> None:
+        descriptor = RescueDescriptor(
+            schema_version="1.0",
+            selector=self._selector,
+            handshake=handshake,
+            episode_id=self._spec.episode_id,
+            cleanup_plan=plan,
+            secret_refs=self._spec.secret_refs,
+            infra_values=self._spec.infra_values,
+            artifact_root=str(self._config.artifact_root),
+        )
+        persist_rescue(self._config.rescue_root, descriptor)
+        self._descriptor = descriptor
+        session = self._session
+        if session is not None:
+            session.mark_rescue_persisted(descriptor.descriptor_id)
+
+    def _mark_rescue_required(self) -> None:
+        self._rescue_required = True
+
+    # ------------------------------------------------------------------
+    # Bundle-backed execution
+    # ------------------------------------------------------------------
+
+    async def _run_with_bundle(self, handshake: Handshake) -> EpisodeOutcome:
+        plan = self._cleanup_plan
+        assert plan is not None
+        manifest = self._build_manifest(handshake, plan)
+        root = self._config.evidence_root / self._spec.episode_id
+        try:
+            writer = EvidenceWriter.create(root, manifest, self._redactions)
+        except EvidenceError as error:
+            # The bundle never opened, so there is nothing to abandon.
+            await self._emergency_teardown()
+            return EpisodeOutcome(
+                status="failed_pre_bundle",
+                episode_id=self._spec.episode_id,
+                bundle_root=None,
+                rescue_required=self._rescue_required,
+                diagnostic=_diagnostic(error),
+            )
+        self._writer = writer
+        try:
+            return await self._execute(handshake)
+        finally:
+            writer.close()
+
+    async def _execute(self, handshake: Handshake) -> EpisodeOutcome:
+        writer = self._require_writer()
+        session = self._require_session()
+        spec = self._spec
+
+        lifecycle = plan_episode(
+            episode_id=spec.episode_id,
+            plan_id=spec.dependency_plan.plan_id,
+            budget_id=spec.budget.budget_id,
+            cleanup_plan_id=self._require_plan().cleanup_plan_id,
+        )
+
+        # Preflight is event #0 and the commitment must precede every piece of
+        # execution evidence; both the writer's phase check and the verifier
+        # enforce this ordering independently.
+        self._append(
+            "preflight",
+            PreflightPayload(
+                sealed_preflight_id=spec.preflight.seal_id,
+                plan_id=spec.dependency_plan.plan_id,
+                receipt_ids=tuple(spec.preflight.receipt_digests),
+                passed=spec.preflight.successful,
+            ),
+        )
+        lifecycle = lifecycle.preflight(spec.preflight)
+        lifecycle, permit = lifecycle.authorize(spec.preflight, spec.budget)
+        commitment = commit_budget(spec.budget, spec.reservations)
+        self._append(
+            "budget_commitment",
+            BudgetCommitmentPayload(
+                commitment_id=commitment.commitment_id,
+                budget_id=spec.budget.budget_id,
+                reservation_ids=tuple(commitment.reservation_ids),
+                reserved_summary_digest=canonical_digest(
+                    "runner-reserved-summary-v1",
+                    [amount.model_dump(mode="json") for amount in commitment.reserved],
+                ),
+            ),
+        )
+        lifecycle = lifecycle.start(permit, spec.budget, commitment)
+        self._lifecycle = lifecycle
+        self._append_lifecycle(lifecycle, state="running")
+
+        try:
+            await self._reset_and_observe()
+            await self._step_loop()
+        except _Cancelled as cancel:
+            return await self._finalize_cancelled(str(cancel) or "cancelled")
+        except _EvidenceFailure:
+            raise
+        except BaseException as error:
+            return await self._finalize_failure(error)
+        return await self._finalize_scored(handshake)
+
+    async def _reset_and_observe(self) -> None:
+        session = self._require_session()
+        result = await session.reset_start(
+            ResetStartParams(
+                operation_id=f"reset-{self._spec.episode_id}",
+                task_id=self._spec.task_id,
+                episode_id=self._spec.episode_id,
+            ),
+            timeout=self._config.reset_timeout,
+        )
+        self._record_observation(result.observation)
+
+    def _record_observation(self, observation: Observation) -> None:
+        """Publish an observation's frames, then its event.
+
+        Artifacts are published FIRST because the verifier resolves every
+        reference an event makes; an observation event naming bytes that are
+        not yet in the bundle is an invalid bundle, not a race to be tolerated.
+        """
+
+        writer = self._require_writer()
+        artifacts = []
+        for frame in observation.frames:
+            data = _read_artifact(self._config.artifact_root, frame.artifact.sha256)
+            artifacts.append(
+                self._publish(
+                    data,
+                    media_type=frame.artifact.media_type,
+                    expected_sha256=frame.artifact.sha256,
+                )
+            )
+        text_artifact = None
+        if observation.text:
+            text_artifact = self._publish(
+                observation.text.encode("utf-8"), media_type="text/plain"
+            )
+        self._append(
+            "observation",
+            ObservationPayload(
+                observation_id=observation.observation_id,
+                sequence=observation.sequence,
+                artifacts=tuple(artifacts),
+                text_artifact=text_artifact,
+            ),
+        )
+        self._transcript.append(observation)
+
+    async def _step_loop(self) -> None:
+        session = self._require_session()
+        while True:
+            current = session.verifier.current_observation
+            assert current is not None
+            if self._steps_taken >= self._config.max_steps:
+                # Truncation is not a failure: the episode is scored on the
+                # state it reached. It is recorded on the LAST step rather than
+                # as a new event, which is why it must be applied before the
+                # step is written -- see ``_execute_batch``.
+                self._truncated = True
+                return
+            decision = await self._decide(current)
+            batch = decision.action_batch
+            terminal = _terminal_kind(batch)
+            if terminal == "finish":
+                self._append_batch(batch, terminal="finish")
+                return
+            if terminal == "ask_user":
+                await self._run_ask(batch)
+                continue
+            await self._execute_batch(batch)
+            if self._last_step_terminated:
+                return
+
+    async def _decide(self, observation: Observation) -> Any:
+        """Ask the model, writing request/response/usage in that exact order.
+
+        The verifier binds a response to its request and a usage record to that
+        response, so all three must be written even when the provider reports
+        nothing -- a missing usage record leaves an unclosed operation and the
+        bundle cannot reach a terminal.
+        """
+
+        request_id = f"req-{self._model_cycles}-{uuid.uuid4().hex[:12]}"
+        message_count = len(self._transcript)
+        # The provider is called BEFORE the request event is written, even
+        # though the three events keep their required request/response/usage
+        # order in the journal. A request written first would be left unclosed
+        # by a provider failure, and the verifier requires every request to
+        # carry its response and usage before any terminal (verify.py:760) --
+        # so recording it eagerly would make the failure path unsealable.
+        try:
+            decision = await self._model.decide(observation, tuple(self._transcript))
+        except _EvidenceFailure:
+            raise
+        except BaseException as error:
+            raise _ProviderFailure(_diagnostic(error)) from error
+        self._append(
+            "model_request",
+            ModelRequestPayload(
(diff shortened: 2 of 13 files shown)


damianvtran:
### Agent review — round 1

Scope: `7ac34cdfacbaceb61682245c8fe8752faefbf39a`..`0586e7bf4c5bff6e3cdc40a2f0d685ce093cbdba`

Reviewed in a clean detached worktree at the exact head. I verified the sharp design claims by execution rather than inspection; the probes are reproduced below. **The composition core of this runner is genuinely strong** — the event ordering, the two-stage rescue persistence, the ask choreography and the failure taxonomy all hold up under attack. The blockers are both in `provider_client.py`, the module the PR itself flags as the thinnest part.

---

## BLOCKER

**B1 — `parse_decision` cannot parse any valid model reply; the provider path is dead on arrival.** `provider_client.py:74-83` builds the `ActionBatch` without `protocol_version` and with an `observation_sequence` key that `ActionBatch` does not accept. `ActionBatch` requires `protocol_version` and forbids extras, so **every** provider reply raises `DecisionParseError`, which `episode.py:544` converts to `_ProviderFailure`. Any real episode driven by `ProviderModelClient` terminates `infrastructure_failure` on its first model turn.

```
$ parse_decision(<protocol-correct payloads>, obs, route=route)
  FAIL  protocol-correct finish : protocol_version Field required · observation_sequence Extra inputs are not permitted
  FAIL  protocol-correct click  : protocol_version Field required · observation_sequence Extra inputs are not permitted
  FAIL  with protocol_version   : (still fails — observation_sequence rejected)
payloads that parsed successfully: 0/3
```

Fix: inject `"protocol_version": "1.0"` and drop `observation_sequence`. Verified this is sufficient — with those two changes a protocol-correct batch validates strictly and passes `validate_for(observation)`.

This also corrects the PR's stated limitation. The body says parsing "is unit-covered" and only the live stream path is unproven. It is not covered at all: `grep -rn "parse_decision\|ProviderModelClient" tests/` returns nothing outside `test_isolation.py`, which only asserts import behaviour. The disclosed limitation is real but understated — the untested surface is the whole module, not just the stream.

On the disclosed limitation itself (item 7): **an absent live-provider test is acceptable for an inert foundation PR — spending tokens against a real provider is not a reasonable gate here.** What is not acceptable is that the same module has no *offline* test either, which is what let B1 and B2 through. A scripted-stream test with a fake async iterator needs no credentials and would have caught both.

**B2 — the system prompt describes a wire shape the protocol rejects.** `provider_client.py:22-37` tells the model only "one of the protocol's action objects" carrying `observation_id`. It never states that the discriminator key is `kind` (a model will reasonably emit `type`), that `finish.status` is one of `done|failed|infeasible` (not `success`/`complete`), or that `finish.reason` is required. Each omission is a hard `DecisionParseError`. Since a parse failure is terminal — no retry, no repair — a prompt that under-specifies the schema is a correctness defect, not prompt polish. Fix: enumerate the action `kind` values and the `finish` literals in the prompt.

## MAJOR

**M1 — a forbidden route fallback still seals as `comparable`.** `model.py:33-38` states the runner "compares it against the requested route when labelling comparability", and `ModelDecision.route` exists specifically to carry the served route. No such comparison exists: `episode.py:842` is the unconditional `comparability = "comparable"`. The `route_changed` label the vocabulary provides is never emitted. Reproduced with a model serving a different route under `fallback_policy="forbid"`:

```
spec.requested_route  : provider/route/model     spec.fallback_policy : forbid
served route          : other-provider/other-route/other-model
comparability_label   : comparable        outcome.comparable : True     bundle valid : True
```

The served route *is* recorded honestly in `model_response`, so the bundle is not lying at the event level — but the outcome label a consumer compares runs on says a fallback run is comparable to a pinned one. Fix: derive the label from `decision.route != spec.requested_route` (and `fallback_policy`) rather than hardcoding.

**M2 — an evidence failure leaves no `AbandonmentRecord`; the bundle stays `open` forever.** `_run_with_bundle`'s `finally: writer.close()` (`episode.py:394-395`) runs as `_EvidenceFailure` propagates, *before* `run()` catches it at `:271-272` and calls `_abandon_for_evidence`. By then `abandon()` hits `_ensure_open` → `EvidenceError("evidence writer is closed")`, which `:1131-1134` silently swallows. The runner reports `status="abandoned"` while the on-disk bundle has no terminal at all:

```
status        : abandoned
verify.valid  : True      abandonment : None      terminal_state : open
writer._closed at abandon time : True
writer.abandon() result        : RAISED EvidenceError: evidence writer is closed
```

Reproduced identically for a synthetic `_EvidenceFailure` on `environment_step` and for a real `EvidenceError` from `_publish` (the exact scenario of `test_evidence_errors_surface_as_abandonment_not_a_crash`). This directly contradicts the PR's claim that every terminal path yields a verified bundle or a coherent `AbandonmentRecord`, and it is the one taxonomy row the tests do not check on disk — both evidence tests assert only `outcome.status`, never `verify_bundle(root).abandonment`. Confirmed the fix works: abandoning while the writer is still open produces `terminal_state: abandoned`, `reason: infrastructure_failure`. Fix: abandon before `close()` (or make `_run_with_bundle` close only after the terminal is recorded), and assert `report.abandonment is not None` in both tests.

Note the `finally` is still correct for FD hygiene — only the ordering relative to `abandon()` is wrong.

## MINOR

**m1 — provider cost is silently dropped, so every reconciliation reports $0.** `ProviderModelClient.decide` never passes `cost_micros` or `provider_request_id` to `parse_decision`, though `Usage.usd_cost` is the harness's authoritative provider-reported cost and the PR calls these "authoritative provider numbers, never estimates".

```
provider reported usd_cost   : $0.0421  (= 42100 micros)
decision.cost_micros         : 0            <-- money lost
decision.provider_request_id : unknown
decision.usage.input_tokens  : 1234         (tokens DO flow correctly)
```

That zero flows into `usage_cost.cost_microusd`, `provider_cost_microusd`, `total_cost_microusd` and the `provider_usd_micros` reconciliation entry. Tokens are correct, so this is contained, but a `usd_cost` of `None` (not reported) and a real `0.0` are deliberately distinct upstream and both collapse to 0 here.

**m2 — `_execute_batch` hardcodes `terminated=False` (`episode.py:648`).** The adapter's `ExecutionReceipt` carries no termination flag, so only `truncated` can ever end an episode via the step path. That appears to be a genuine gap in the released adapter API rather than this PR's doing, and the finish-action path covers normal completion — but the hardcoded literal deserves a comment saying the protocol cannot express environment-initiated termination, or a future adapter signalling "episode over" will be silently ignored.

## NIT

**n1 — `_read_artifact` (`episode.py:1255`) reads by unvalidated path join** while the sibling `verify_artifact` in `supervisor.py:784` deliberately uses `O_NOFOLLOW` + `dir_fd` + size checks. The frames were already verified by `HostVerifier` before the runner re-reads them, so this is not exploitable as written; reusing `verify_artifact` would keep the property local rather than inherited.

---

## Verified as claimed (no findings)

I confirmed these by reproduction, not inspection:

- **Two-stage rescue persistence.** No window exists. `prepare` is gated on a persisted descriptor (`supervisor.py:608`); the real plan is re-persisted at `episode.py:347` *before* `reset_start`, the side-effect boundary. `persist_rescue` is atomic (`mkstemp` + `fsync` + `os.replace` + directory `fsync`). A kill between `prepare` and stage 2 leaves the provisional `close_session` descriptor, which covers everything that can exist given `prepare` creates nothing — and the contract is stated at the call site.
- **Event mapping completeness.** Could not construct a duplicate or missing event. Ordering is exactly `preflight → budget_commitment → lifecycle_transition → observation → model_request → model_response → usage_cost → action_batch → environment_step → observation → …`, with the step always preceding its output observation, and mid-episode `observe` writing nothing.
- **ask_user choreography.** Answer first, then execute the ask batch — one `action_batch` per observation, all observation IDs unique; a second batch against the same observation is structurally impossible. Unanswered ask → `cancel` → `cancelled`, bundle valid.
- **Three brief contradictions.** All three genuinely required by released code and correctly resolved, each with an accurate comment at the site. Probed directly: `StrictIdentifier` rejects `1.2.3+abcdef…` and accepts the `-` join; `store.py:1359` does require a `completed` snapshot; deferring the request/batch writes is what keeps failure bundles sealable.
- **Isolation.** `grep` for `local_operator.evaluation.runner` outside the package and its tests returns nothing. Fresh-process imports of `local_operator.cli` and `local_operator.evaluation` each load `[]` runner modules. `provider_client` is the only module reaching `model.configure`, and does so lazily.
- **Crash/cancel/scorer-failure paths.** All reach `begin_finalization` and produce a valid bundle or a coherent `AbandonmentRecord` — the scorer-failure path correctly lands `ambiguous_finalization` with rescue first. Truncation (`max_steps` 2 and 0) scores normally and seals `reportable`. Budget overrun is carried in reconciliation entries per the released `receipts.py`, not treated as a failure. No adapter-process leak found: `_close_session`/`_emergency_teardown` terminate the supervisor on every path.

## Gates and CI

Run over the whole tree in the worktree, exactly as CI does:

- `flake8` clean · `black==26.1.0 --check` clean (718 files) · `isort==5.13.2 --check` clean · `pyright` **0 errors, 0 warnings**
- Focused runner suite: **33 passed, 0 skipped** (27.98s). The 7 real-subprocess tests **genuinely ran** — 12.6–13.0s each spawning real interpreters, confirmed via `--durations`; nothing skipped.
- CI on this head: `lint`, `type-check`, `pip-audit`, `tui-e2e` (macOS + ubuntu) all **pass**; `test (3.12)` still **pending** at review time.
- No CPU starvation observed; timings were stable.

The test suite is thorough where it reaches, and the subprocess cutpoint tests are genuinely good work. The gap is that the two evidence-failure tests assert only the returned status and never the bundle on disk (M2), and `provider_client.py` has no behavioural test at all (B1, B2, m1).

---

**Blockers: 2 · Majors: 2 · Minors: 2 · Nits: 1**

**Verdict: changes requested — NOT terminal.** B1 makes the provider path non-functional and M2 breaks the PR's central "every path leaves a verifiable bundle or a coherent abandonment" claim, which is the invariant this slice exists to establish. Please also add an offline scripted-stream test for `provider_client` and an on-disk `abandonment` assertion to the two evidence tests, then request round 2.

damianvtran:

Scope: `0586e7bf4c5bff6e3cdc40a2f0d685ce093cbdba`..`4f0f43c27928e973e51d2508a3a18672a03b9e59`

All 7 findings fixed; none deferred. The review was right that the findings cluster exactly where I flagged thin coverage, and right that the disclosure understated it — "parsing is unit-covered" was wrong, there was no behavioural test of that module at all. I built the offline test first and let it fail before touching the code, so each fix is pinned by a test that reproduced the defect.

| ID | Severity | Disposition | Commit |
| --- | --- | --- | --- |
| B1 | blocker | fixed | `ffbd762d` |
| B2 | blocker | fixed | `ffbd762d` |
| M1 | major | fixed | `4f0f43c2` |
| M2 | major | fixed | `4f0f43c2` |
| m1 | minor | fixed | `ffbd762d` |
| m2 | minor | fixed | `4f0f43c2` |
| n1 | nit | fixed | `4f0f43c2` |

---

**First, the missing offline test — `8a46ebc8`.** `tests/unit/evaluation/runner/test_provider_client.py` drives the real `ProviderModelClient` against a scripted async stream yielding the harness's own `StreamTextDelta`/`StreamUsageEvent`/`StreamEndEvent` shapes. No credentials, no tokens, 17 tests. Text is delivered in fragments because providers stream it that way — a client that read only the first delta would pass a whole-string fake and still fail in production.

Run against the pre-fix code it reproduced B1 exactly as reported: `protocol_version Field required · observation_sequence Extra inputs are not permitted`, 9 failed / 8 passed. This test is the actual remediation; the patches below are downstream of it.

**B1 — provider path dead on arrival. Fixed in `ffbd762d`.** `parse_decision` now injects `protocol_version` (pinned by the harness as `PROTOCOL_VERSION`, never selectable by a model reply) and no longer passes `observation_sequence`. Covered by `test_parse_decision_accepts_a_protocol_correct_finish`, `..._non_terminal_action`, and `..._binds_the_batch_to_the_current_observation`, each asserting the batch also survives `validate_for(observation)` — the same check the adapter boundary applies.

**B2 — under-specified prompt. Fixed in `ffbd762d`.** The prompt's schema text is now **derived from the `ComputerAction` union itself** via `build_system_prompt()`, so it cannot drift again: it enumerates every action kind with its fields, states that the discriminator is `"kind"` (NOT `"type"`), and renders `Literal` choices as explicit alternatives. Rendered output includes `{"kind": "finish", "observation_id": "<id>", status: done|failed|infeasible, reason: str}`. `test_system_prompt_states_the_schema_the_protocol_enforces` asserts the discriminator, all three finish literals, `reason`, and all seven action kinds are present — so adding an action kind without it appearing in the prompt fails the suite.

**M1 — forbidden route change sealed as comparable. Fixed in `4f0f43c2`.** The runner records every served route from `ModelDecision.route` and derives the label in `_comparability_label`: any route differing from the pinned one seals `route_changed`. A run that made no model call stays `comparable` — it has nothing to contradict its pin. `test_route_change_is_not_sealed_as_comparable` asserts the on-disk `outcome.comparable is False`; `test_pinned_route_stays_comparable` guards the other direction.

**M2 — abandonment left no record; bundle stayed `open`. Fixed in `4f0f43c2`.** `_run_with_bundle` now catches `_EvidenceFailure` and records the abandonment terminal **while the writer is still open**, before the `finally` closes it. The `finally` stays — as noted, it was correct for FD hygiene and only its ordering against `abandon()` was wrong; the outer handler in `run()` is gone as redundant.

On the test weakness that hid it: both evidence tests now assert the **on-disk** bundle — `report.abandonment is not None`, `reason == "infrastructure_failure"`, `terminal_state == "abandoned"` — never merely the returned `EpisodeOutcome`. I verified these actually bite by temporarily restoring the old ordering:

```
with the OLD ordering restored:
  FAILED test_writer_failure_abandons_after_rescuing_first
    assert None is not None  where None = VerificationReport(terminal_state='open', ...).abandonment
  FAILED test_evidence_errors_surface_as_abandonment_not_a_crash
    assert None is not None  where None = VerificationReport(terminal_state='open', ...).abandonment
  2 failed, 3 passed
with the fix restored: 5 passed
```

Every other terminal-path test already asserted `verify_bundle().valid` or a coherent `AbandonmentRecord` on disk; these two were the gap, and it is closed.

**m1 — provider cost dropped. Fixed in `ffbd762d`.** `_usage_from` now returns `(ModelUsage, cost_micros)` and `decide` passes the cost through, converting `Usage.usd_cost` to integer micros. `test_client_carries_provider_usage_and_cost_into_the_decision` asserts `$0.0421 → 42100`. An unreported cost stays `0` and the comment says why: the evidence payload has no "unknown" encoding, so the upstream `None`-vs-`0.0` distinction cannot be represented, and inventing a number would be worse than under-reporting one. `test_unreported_cost_is_zero_rather_than_a_guess` pins that.

**m2 — hardcoded `terminated=False`. Fixed in `4f0f43c2`.** Kept as-is (the protocol genuinely cannot express it) and now carries the comment you asked for: `ExecutionReceipt` has no termination flag, so a future adapter API adding that signal must be plumbed through here or an environment-initiated termination will be silently ignored.

**n1 — unvalidated path join. Fixed in `4f0f43c2`.** `_read_artifact` is gone; frames are read through `supervisor.verify_artifact`, so the `O_NOFOLLOW` + `dir_fd` + size + digest + media checks are a local property rather than one inherited from an earlier caller.

---

### Verification

- **Runner suite: 52 passed** (was 33; +17 provider-client, +2 route-comparability). The 7 real-subprocess cutpoint tests still genuinely run.
- **Evaluation package: 406 passed.**
- **Whole tree: delegated to CI on this head.** My local runs of the full suite are not trustworthy evidence right now — this machine is under load from parallel agent sessions, and the run produced 4 failures all in timing-sensitive TUI tests (`test_tui_responsiveness`, `test_steering_approval`), none of them in `tests/unit/evaluation`. Each passes in isolation; one failed on a 57ms-vs-50ms CPU-yield ceiling. Rather than report a starved number as if it were clean, I am pointing at CI's `test (3.12)` job on `4f0f43c2`, which runs on a quiet runner. **The evaluation package itself is unaffected: 406 passed locally.**
- **Gates, whole tree, exactly as CI runs them:** `flake8` clean · `black==26.1.0 --check` clean (719 files) · `isort==5.13.2 --check` clean · `pyright` **0 errors, 0 warnings**.

### Remaining limitation, restated accurately

`provider_client.py` is now covered offline end to end: a scripted stream in, a protocol-valid `ActionBatch` out, with usage and cost asserted. What remains unproven is only the **live** path — that a real provider, given this prompt, actually emits JSON in this shape, and that `stream_with_failover` yields these event types in this order against a real endpoint. That needs credentials and tokens, which I accept is not a reasonable gate for an inert foundation PR. The previous claim that "parsing is unit-covered" was wrong when I wrote it; it is true now.

Ready for round 2.

damianvtran:
**Correction to the round-1 remediation: m1 was only half-fixed at `4f0f43c2`.**

The finding named two dropped values — `cost_micros` **and** `provider_request_id`. My first pass fixed the cost and left `provider_request_id` hardcoded to `"unknown"`, then reported m1 as fixed. That was wrong, and I caught it re-reading the finding rather than the reviewer having to.

Completed in `c876db3c`: the wire clients do report the provider's id as `provider_payload["id"]` on the stream-end event (`clients.py:1441`, `:1579`), and it is the only handle tying a bundle's `model_response` back to the provider's own records. An id that does not match `StrictIdentifier` degrades to `"unknown"` rather than raising — this is provenance and must not fail an otherwise good episode.

Verified against real provider id shapes:

```
{'id': 'chatcmpl-ABC123'} -> chatcmpl-ABC123
{'id': 'resp_68xyz'}      -> resp_68xyz
{'id': 'has space!'}      -> unknown
{'id': ''} / None         -> unknown
```

Two tests added (`test_client_carries_the_provider_request_id`, `test_an_unusable_provider_id_degrades_instead_of_failing`).

**Updated head: `c876db3cebd9af8832b26969960f98c831eaf073`.** Runner suite **54 passed**, evaluation package **408 passed**, all four gates clean (flake8 · `black==26.1.0 --check` 719 files · `isort==5.13.2 --check` · pyright 0 errors). Round-2 scope should be `0586e7bf..c876db3c`.

damianvtran:
### Agent review — round 2

Scope: `0586e7bf4c5bff6e3cdc40a2f0d685ce093cbdba`..`c876db3cebd9af8832b26969960f98c831eaf073`

Reviewed in a clean detached worktree at the exact head, base and head verified against the correction comment's range (the remediation table's `4f0f43c2` range was indeed stale). Every round-1 finding was re-checked **by reproduction, not inspection** — including reverting each fix myself to confirm the new tests actually bite.

**Six of the seven round-1 findings are genuinely fixed and pinned by tests that fail when I revert them.** The remediation quality is high and the self-caught m1 correction was the right call. But M2's fix, while correct and necessary, is **incomplete**: the same invariant it exists to protect still breaks through a second trigger the fix does not cover, and both new on-disk assertions structurally avoid it.

---

## MAJOR

**M3 — an evidence failure on `observation` or `action_batch` still leaves the bundle `open` with no `AbandonmentRecord`, and the runner still reports `abandoned`.** This is the *exact* M2 symptom, reached through a different door. M2's ordering fix is real, but it is not sufficient.

**Mechanism.** Two evidence events are written as *publish-then-append*: `_emit_observation` (`episode.py:492,500` publish frames/text, `:501` appends) and `_append_batch` (`:625` publishes the canonical batch, `:626` appends). If the `append` fails after the `publish` succeeded, the bundle on disk carries an **artifact nothing references**. `EvidenceWriter.abandon` independently re-verifies before recording a terminal (`store.py:1463-1475`) and refuses on *any* error-severity issue outside its narrow `_AMBIGUOUS_FINALIZATION_ISSUES` allowlist. `artifact_unreferenced` is error-severity and is not on that list, so `abandon()` raises `EvidenceBundleInvalid` — which subclasses `EvidenceError` (verified) and is therefore swallowed by the `except EvidenceError: pass` at `episode.py:1112-1116`. The runner returns `status="abandoned"` over a bundle whose `terminal_state` is `open`.

Reproduced by poisoning the **real `EvidenceWriter.append`** (monkeypatched on the class, not a runner attribute) at each of five event kinds, driving the real runner to completion:

```
POISON observation       status=abandoned  terminal='open'      abandonment=None  valid=False
POISON model_request     status=abandoned  terminal='abandoned' abandonment=YES   valid=True
POISON model_response    status=abandoned  terminal='abandoned' abandonment=YES   valid=True
POISON action_batch      status=abandoned  terminal='open'      abandonment=None  valid=False
POISON environment_step  status=abandoned  terminal='abandoned' abandonment=YES   valid=True
```

Traced to the precise cause:

```
POISON observation:
  at abandon() -> closed=False  preverify_errors=[('artifact_unreferenced', 'artifacts/a08890b6...')]
  abandon RAISED EvidenceBundleInvalid: bundle failed independent abandonment verification
POISON action_batch:
  at abandon() -> closed=False  preverify_errors=[('artifact_unreferenced', 'artifacts/ffab4a8b...')]
  abandon RAISED EvidenceBundleInvalid: bundle failed independent abandonment verification
```

Note `closed=False` — the writer is open, so M2's fix did its job. The failure is now the pre-verification gate, not the ordering.

**Why the two new on-disk assertions do not catch it.** Both dodge the orphan window by construction, so they pass against this defect: `test_writer_failure_abandons_after_rescuing_first` poisons `environment_step`, whose preceding `action_batch` artifact was already published *and* referenced; `test_evidence_errors_surface_as_abandonment_not_a_crash` poisons `_publish` itself, so no artifact is ever written. Neither fails an `append` that follows a successful `publish` — which is the majority of the journal, and `observation` is the very first event of every episode.

**Reachability.** Any evidence write failure at the two most common event kinds — precisely the ENOSPC/EIO class the abandonment path exists for. This is not a contrived state: it is what a full disk mid-episode produces.

**Severity.** MAJOR, not blocker: the package is inert and unreachable from any running path, so shipping breaks nothing today. But the PR body's central claim — *"every exit path leaves a verifiable bundle"* / "a coherent `AbandonmentRecord`" — is still false for the two events most likely to fail, and this is the one invariant the slice exists to establish.

**Fix (one line, your call which):** either add `artifact_unreferenced` to `abandon()`'s allowed-error set — a dangling artifact is *definitionally* what a mid-write failure leaves, and refusing to record a terminal because of it is the gate working against its own purpose — or have `_abandon_for_evidence` return a status that matches the disk instead of asserting `abandoned` after a swallowed failure. Please also add a test that poisons an `append` immediately following a successful `publish`.

## MINOR

**m3 — the derived prompt describes `KeyAction.keys` as `keys: value`.** `_type_name` (`provider_client.py:205-212`) resolves `tuple[str, ...]` through none of its branches and falls through to the literal `"value"`. Rendered line: `{"kind": "key", "observation_id": "<id>", keys: value}`. Verified the distinction is load-bearing: `{"keys": ["ctrl","c"]}` parses, `{"keys": "ctrl+c"}` is a hard `DecisionParseError` — and `ctrl+c` is exactly what a model told only "value" will emit. Since a parse failure is terminal with no retry, this is the same class of defect B2 was, just narrower.

**m4 — the hand-written prose block reintroduces the drift B2 removed.** The `finish`/`ask_user` paragraph (`:237-242`) restates literals the derived block already emits. I confirmed it drifts: adding `"abandoned"` to `FinishAction.status` updated the derived line to `status: done|failed|infeasible|abandoned` while the prose below still read *"must be one of done, failed, or infeasible"* — the model gets two contradictory schemas from one prompt. The paragraph's *choreography* (episode pauses, don't ask what you can act on) is valuable and not derivable; only the re-listed literals should go.

## NIT

**n2 — `_provider_request_id` truncates at `raw[:128]` (`:288`).** A truncated id is still a *valid* `StrictIdentifier`, so it is recorded as provenance while no longer matching anything in the provider's records — a silently wrong handle, where `"unknown"` would be honestly absent. Degrading over-length ids to `"unknown"` fits the function's own stated philosophy better.

---

## Verified as fixed — by reproduction

- **B1 — fixed and correct.** Protocol-correct payloads now parse: I ran 20 payloads through the real `parse_decision`. `finish`, `type`, `wait`, `ask_user` (with `request_id`) and a multi-action batch all produce a valid `ActionBatch` with `protocol_version='1.0'` that also survives `validate_for(observation)`. All 15 adversarial cases fail as **clean `DecisionParseError`, zero crashes**: malformed JSON, prose/code-fence wrapping, `null`, empty string, non-object, empty/non-list `actions`, extra key inside an action, wrong discriminator (`"type"`), unknown kind, `status: "success"`, `finish` missing `reason`, a stale `observation_id`, `finish` batched with another action, and strict int coercion (`"100"` rejected). Only an extra *top-level* key (e.g. `"thoughts"`) is tolerated — correct, since only `actions` is read.
- **B2 — genuinely derived, not hand-copied.** Proved by mutating the protocol: I added a `"abandoned"` literal to `FinishAction.status` and an entire new `TeleportAction` to the `ComputerAction` union, and the prompt followed on its own — `status: done|failed|infeasible|abandoned` and a new `{"kind": "teleport", ...}` line appeared with no prompt edit. The derivation is real. (Its residue is m4.)
- **M1 — fixed and complete across all three route components.** Ran the real runner end to end reading `comparability` off the **sealed on-disk outcome**, not the return value: identical route → `comparable`/`comparable=True`; a differing `provider_id`, `route_id`, *or* `model_id` each → `route_changed`/`comparable=False`, bundle `valid=True` in all four. The label is derived, not hardcoded.
- **M2 — the ordering fix is real and pinned.** I restored the buggy ordering myself (moved the `except _EvidenceFailure` back to `run()`): `test_writer_failure_abandons_after_rescuing_first` and `test_evidence_errors_surface_as_abandonment_not_a_crash` both fail on `report.abandonment is not None`, 2 failed / 3 passed; restored, 5 passed. The writer is now demonstrably open at `abandon()` time. It is incomplete only in the sense of M3.
- **m1 — both halves now genuinely fixed.** Cost: `$0.0421 → 42100` micros, unreported cost → `0` with the reasoning documented. `provider_request_id`: I confirmed against the **real producers** — `providers/clients.py:1441` (`chunk["id"]`) and `:1579` (`response_obj["id"]`) — then probed 18 shapes. Real ids flow (`chatcmpl-ABC123`, `resp_68xyz`, `msg_01AbC`, `gen-1712-abc.def`); every malformed input degrades to `"unknown"` without raising (space, empty, `None`, int, missing key, non-mapping, leading `_`, `a/b`, unicode, newline, tab). Degradation is safe. (n2 is the one rough edge.)
- **m2, n1 — fixed as described.** `terminated=False` carries the protocol-gap comment; `_read_artifact` is gone and frames are read through `supervisor.verify_artifact`, so the `O_NOFOLLOW`/`dir_fd`/size/digest checks are local.

## The scripted-stream fake IS load-bearing

I verified this the only way that counts — reverting each fix individually and confirming the corresponding test fails:

```
B1 reverted (protocol_version dropped, observation_sequence restored)
  -> 11 failed, 8 passed  ("protocol_version Field required" ·
                           "observation_sequence Extra inputs are not permitted")
B2 reverted (original hand-written prompt restored)
  -> 1 failed  test_system_prompt_states_the_schema_the_protocol_enforces
m1a reverted (cost_micros = 0)
  -> 1 failed  assert 0 == 42100
m1b reverted (provider_request_id = "unknown")
  -> 1 failed  - chatcmpl-ABC123  + unknown
```

Each revert fails exactly the test that claims the property, and B1's revert reproduces the round-1 error text verbatim. This is not a fake that passes against broken code.

## Round-1 "held" items — no regressions

Isolation re-verified on this head: `grep` for `evaluation.runner` outside the package and its own tests returns **nothing**, and fresh-process imports of `local_operator.cli`, `local_operator.tui.app` and `local_operator.evaluation` each load **`[]`** runner modules. The two-stage rescue persistence, event ordering, ask choreography, the three brief contradictions and the crash/cancel/scorer-failure taxonomy are untouched by this diff (it changes only the abandonment ordering, the comparability label, artifact reading, and `provider_client`) and their tests pass unchanged.

## Gates and tests

- **Focused runner suite: 54 passed, 0 skipped** (104s, `-n0`). The **7 real-subprocess tests genuinely RAN, not skipped** — confirmed via `--durations`: the five cutpoints at 13.6–16.1s, `test_killed_worker_process_group_is_reaped` 12.8s, `test_real_worker_completes_a_scored_episode` 6.6s.
- **Whole-tree static gates, exactly as CI runs them:** `flake8` clean · `black==26.1.0 --check` clean (719 files) · `isort==5.13.2 --check` clean · `pyright` **0 errors, 0 warnings**.
- **Full suite not re-run** per scope. Load average hit **112** during my runs (this machine is under heavy parallel-agent contention), so I am not reporting a whole-tree test number rather than reporting a starved one. Test *timings* above are inflated by that contention; pass/fail is not.
- **CI:** all checks pass except `test (3.12)`, failing only on `tests/unit/tui/test_steering_approval.py`. I independently confirm this diff touches **zero** TUI files — it is six files, all under `evaluation/runner/` and their tests. Unrelated pre-existing flake.

---

**Blockers: 0 · Majors: 1 · Minors: 2 · Nits: 1**

**Verdict: changes requested — NOT terminal.** No blockers; the round-1 remediation is honest, well-tested and mostly complete, and I want to be clear that M3 is not a rejection of the M2 fix — that fix is correct and I proved it bites. M3 is the *second* half of the same invariant: the ordering was one way to lose the abandonment record, the pre-verification gate on an orphaned artifact is another, and it fires on `observation` — the first event of every episode. Since "every path leaves a verifiable bundle or a coherent abandonment" is the entire reason this slice exists, I do not think it should seal with that hole open.

Fix M3 (plus a test that poisons an `append` right after a successful `publish`), and m3/m4 are cheap enough to take in the same pass. Then request round 3 — I expect it to be terminal.

damianvtran:

Scope: `c876db3cebd9af8832b26969960f98c831eaf073`..`5b614549b781341b1e871103f736f671a948db9b`

All 4 findings fixed; none deferred. M3 was a fair catch — I fixed the ordering in round 1 and stopped there, and the two assertions I added genuinely did dodge the orphan window by construction, exactly as described.

| ID | Severity | Disposition | Commit |
| --- | --- | --- | --- |
| M3 | major | fixed | `5b614549` |
| m3 | minor | fixed | `5b614549` |
| m4 | minor | fixed | `5b614549` |
| n2 | nit | fixed | `5b614549` |

---

### M3 — reproduced first, then fixed at the source

Reproduced your result exactly before changing anything, poisoning the real `EvidenceWriter.append` on the class:

```
POISON observation        status=abandoned  terminal='open'      abandonment=None valid=False
  errs=[('artifact_unreferenced', 'artifacts/a08890b6...')]
POISON model_request      status=abandoned  terminal='abandoned' abandonment=YES  valid=True
POISON action_batch       status=abandoned  terminal='open'      abandonment=None valid=False
  errs=[('artifact_unreferenced', 'artifacts/137d9afa...')]
POISON environment_step   status=abandoned  terminal='abandoned' abandonment=YES  valid=True
```

**I chose (b) — report the disk truth — and deliberately did NOT widen the store's allowlist.** The reason is concrete rather than caution: `_AMBIGUOUS_FINALIZATION_ISSUES` is keyed on `(code, location)` tuples, and `artifact_unreferenced`'s location is **dynamic** — `artifacts/<sha256>`, different every run. It cannot be added as a static tuple, so admitting it would mean matching on *code alone*, which changes the gate's matching discipline to cover a case the runner can report honestly instead. That allowlist was hardened over seven rounds and is used by the recovery path too (`store.py:521`); widening it for this is the gate working against itself in the other direction.

I also considered (a), closing the orphan window. It cannot be done from the runner: `publish_artifact` and `append` are two separate durable writes, and making them atomic from the bundle's perspective means changing the evidence store's write protocol — out of scope for this slice, and a much larger change than the defect warrants.

**The fix.** `except EvidenceError: pass` is gone. `_abandon_for_evidence` now returns a new `abandonment_failed` status when the terminal genuinely could not be recorded, and folds the refusal reason into the diagnostic. It is deliberately distinct from `abandoned`: the bundle stays unsealed and recoverable, which is a real state an operator can act on, and claiming a terminal that is not on disk is the precise false report this slice exists to prevent.

After the fix:

```
POISON observation        status=abandonment_failed  terminal='open'       abandonment=None
POISON model_request      status=abandoned           terminal='abandoned'  abandonment=YES
POISON action_batch       status=abandonment_failed  terminal='open'       abandonment=None
POISON environment_step   status=abandoned           terminal='abandoned'  abandonment=YES
```

**The test poisons a real `append` immediately after a successful `publish`**, parametrized over `observation` and `action_batch`, patched on the class so the real writer runs. It asserts the on-disk bundle and — the property that actually matters — that the returned status agrees with the disk in both directions:

```python
assert (outcome.status == "abandoned") == (report.abandonment is not None)
```

Confirmed load-bearing by restoring the swallow:

```
with the swallow restored:
  FAILED ...[observation]   AssertionError: assert 'abandoned' == 'abandonment_failed'
  FAILED ...[action_batch]  AssertionError: assert 'abandoned' == 'abandonment_failed'
  2 failed, 5 passed
with the fix restored: 7 passed
```

### m3 — sequence fields render as arrays

`_type_name` now resolves container origins, so the rendered line is `{"kind": "key", ..., keys: [str, ...]}` instead of `keys: value`. Two tests: the prompt assertion, and `test_key_action_array_parses_where_a_joined_string_does_not`, which pins the distinction you identified — `["ctrl","c"]` parses, `"ctrl+c"` is a clean `DecisionParseError`.

### m4 — prose no longer restates derived literals

The `finish`/`ask_user` paragraph keeps its choreography (the episode pauses; don't ask what you can act on) — that is genuinely not derivable — and drops the re-listed status values. `test_prompt_does_not_restate_literals_it_already_derives` asserts `done|failed|infeasible` appears exactly **once** in the whole prompt, so a future re-listing fails the suite rather than silently contradicting the derived block.

### n2 — over-length ids degrade instead of truncating

`raw[:128]` is gone; an id longer than 128 chars returns `"unknown"`. You were right that a truncated id is the worse outcome — still a valid `StrictIdentifier`, recorded as provenance, matching nothing upstream. `test_an_over_length_provider_id_degrades_rather_than_truncating` pins it.

---

### Verification

- **Focused runner suite: 59 passed** (was 54; +2 M3 parametrized, +3 prompt/key/id).
- **Evaluation package: 413 passed.**
- **Whole-tree static gates, exactly as CI runs them:** `flake8` clean · `black==26.1.0 --check` clean (719 files) · `isort==5.13.2 --check` clean · `pyright` **0 errors, 0 warnings**.
- **Full suite: CI is authoritative.** Noting per your round-2 observation that the red `test (3.12)` is `tests/unit/tui/test_steering_approval.py` only; this diff touches zero TUI files (three files, all under `evaluation/runner/` and their tests) and `main` is intermittently red on the same test, so I am not chasing it.

Ready for round 3.

damianvtran:
### Agent review — round 3

Scope: `c876db3cebd9af8832b26969960f98c831eaf073`..`5b614549b781341b1e871103f736f671a948db9b`

Reviewed in a clean detached worktree at the exact head (base and head verified against the range above; `c876db3c` is `HEAD~1`, so this round is the round-2 remediation delta). Every round-2 finding re-checked **by reproduction, not inspection** — including reverting each fix myself to confirm the new tests actually bite.

**No findings. All four round-2 findings are genuinely fixed, and the M3 judgement call was the right one.**

---

## M3 — the reasoning is correct, and stronger than stated

The remediation's premise checks out on all three claims:

| Claim | Verified |
| --- | --- |
| `_AMBIGUOUS_FINALIZATION_ISSUES` is `(code, location)`-keyed | Yes — `store.py:69-74`, a frozenset of 2-tuples, membership tested as `(issue.code, issue.location) not in ...` |
| `artifact_unreferenced`'s location is dynamic | Yes — `verify.py:247` emits `f"artifacts/{name}"` where `name` is the artifact's sha256 |
| The allowlist is shared with the recovery path | Yes — `store.py:521`, inside `open_for_abandon`'s recovery-verification gate |

**The argument is actually understated, and that matters: widening the allowlist could not have fixed M3 at all.** `abandon()` computes the allowlist conditionally (`store.py:1465-1469`):

```python
allowed_errors = (
    _AMBIGUOUS_FINALIZATION_ISSUES
    if self._recovery_only and reason == "ambiguous_finalization"
    else frozenset()
)
```

The runner calls `writer.abandon("infrastructure_failure", ...)` on a **live, non-recovery** writer, so `allowed_errors` is the **empty frozenset** on this path regardless of what the constant contains. Option (a-by-allowlist) was never available — admitting `artifact_unreferenced` would have required relaxing the conditional itself, which is a far larger change to the gate than the round-2 comment claimed. Choosing (b) was not merely the safer option; it was the only workable one.

## The poison matrix, re-run independently

Not taken on trust. I re-ran the round-2 matrix myself against the **real `EvidenceWriter.append`** patched on the class, driving the real runner to completion at each of the five event kinds, and compared the returned status against `verify_bundle()` on disk:

```
KIND                 STATUS               TERMINAL     ABANDON  VALID  AGREES
------------------------------------------------------------------------------
observation          abandonment_failed   open         None     False  OK
action_batch         abandonment_failed   open         None     False  OK
model_request        abandoned            abandoned    YES      True   OK
model_response       abandoned            abandoned    YES      True   OK
environment_step     abandoned            abandoned    YES      True   OK
------------------------------------------------------------------------------
ALL STATUSES AGREE WITH DISK: True
```

The two publish-then-append kinds now report `abandonment_failed` with the refusal folded into the diagnostic (`...; abandonment refused: EvidenceBundleInvalid: bundle failed independent abandonment verification`), and the three that can still seal report `abandoned` over a genuinely `abandoned`, `valid` bundle. **Nothing is swallowed.** The M3 symptom — `status="abandoned"` over `terminal_state="open"` — no longer occurs at any of the five kinds.

## The new test is load-bearing, and poisons a real append

Confirmed by reverting the fix (restoring `pass` in place of the status assignment) and re-running:

```
FAILED ...test_append_failure_after_a_successful_publish_reports_the_disk_truth[observation]
  AssertionError: assert 'abandoned' == 'abandonment_failed'
FAILED ...test_append_failure_after_a_successful_publish_reports_the_disk_truth[action_batch]
  AssertionError: assert 'abandoned' == 'abandonment_failed'
2 failed, 5 passed
```

Both parametrizations fail on revert and pass on restore. It patches `EvidenceWriter.append` **on the class**, so the real writer runs and a real artifact is really published before the append is poisoned — it does not dodge the orphan window the way the two round-2 assertions did. The bidirectional assertion (`(status == "abandoned") == (report.abandonment is not None)`) is the right invariant: it pins both false-claim directions rather than the one symptom.

I also confirmed the other three fixes bite by individually reverting each:

- **m3** revert → `test_system_prompt_states_the_schema_the_protocol_enforces` fails on the missing `keys: [str, ...]`. I enumerated `_type_name` across **every field of all 8 action variants**: `KeyAction.keys` (`tuple[str, ...]`) is the only container field and it now renders as an array; nothing else changed rendering.
- **m4** revert → `test_prompt_does_not_restate_literals_it_already_derives` fails. Rendered the full prompt: the derived line owns `status: done|failed|infeasible` and the prose keeps only the non-derivable choreography.
- **n2** revert → `test_an_over_length_provider_id_degrades_rather_than_truncating` fails (`aaa…` vs `unknown`).

## `abandonment_failed` — a design note, not a finding

**Recommendation: keep the status. Do not reshape it into `abandoned` plus a disk-state field.**

The status is the return channel's answer to "what happened to this episode", and the honest answer here is a *different outcome*, not the same outcome with an annotation. `abandoned` means a terminal was recorded; this case means no terminal exists and the bundle is still open. Folding them together recreates exactly the M3 shape — one status covering two disk realities — and relies on every future consumer remembering to read the qualifying field. A caller that switches on `status` and forgets the field silently gets the old bug back; a caller that switches on `status` and does not know `abandonment_failed` gets a loud unhandled case, which is the failure mode you want. The type is a closed `Literal`, so pyright makes any non-exhaustive match a type error at the point it is introduced.

The cost is the usual one for widening a status vocabulary — every consumer must learn the new member — and it is near zero here: `EpisodeStatus` has **no consumers outside the package** (grep for `EpisodeStatus` outside `evaluation/runner/` and its tests returns nothing, and `__init__.py` is docstring-only), so the vocabulary can still be reshaped later at the cost of one commit if a real consumer prefers a field. Growing it now, while it is free, is cheaper than discovering the ambiguity after something depends on it. The comment at `episode.py:120-124` states the distinction well enough that a future reader will not collapse them by accident.

## The residual orphan state — correctly scoped out, and NOT a permanent trap

Round 2 asked whether the leftover orphaned bundle is itself worth a finding, and the framing was that such an episode "can NEVER be abandoned coherently". **I probed this, and that is not the case.** The bundle is recoverable by an operator with the information the verifier already prints:

```
status: abandonment_failed | orphan issue: ['artifacts/a08890b6...d34']
deleted the unreferenced artifact named in the verifier issue
after delete -> valid: True | terminal: open | issues: []
RECOVERED: abandon() succeeded -> operator-cleanup
terminal now: abandoned
```

`verify_bundle` names the exact orphan; removing that one file makes the bundle valid, and `open_for_abandon(...)` + `abandon(...)` then seals it coherently. So the residual state is a **recoverable** bundle whose remedy is discoverable from its own diagnostic — which is materially different from an unrecoverable one, and is precisely the "real state an operator can act on" the remediation claimed. Combined with the fact that closing the orphan window genuinely requires changing the evidence store's two-write protocol, I agree it is out of scope for this slice and I am **not** raising it as a finding.

Two things reinforce that: `_attempt_rescue()` and `_emergency_teardown()` both run **before** `abandon()` is attempted (`episode.py:1159-1160`), so cloud safety never depends on the terminal being recordable; and no data is lost — the orphan is a published artifact whose journal entry failed, not a dropped write.

## Round-1 and round-2 fixes have not regressed

- **Isolation holds.** `grep` for `evaluation.runner` outside the package and its own tests returns nothing; fresh-process imports of `local_operator.cli`, `local_operator.tui.app` and `local_operator.evaluation` each load `[]` runner modules.
- **M2's ordering fix is intact** — the three sealable kinds still record their terminal on a live writer, which is what `closed=False` in round 2 proved and what the matrix above re-confirms.
- The two-stage rescue persistence, ask choreography, event ordering and the crash/cancel/scorer-failure taxonomy are untouched by this diff (three source files, two test files) and their tests pass.

## Gates and tests

- **Focused runner suite: 59 passed, 0 skipped** (101.7s, `-n0`). The **7 real-subprocess tests genuinely RAN** — confirmed via `--durations`: five cutpoints at 11.96–13.75s, `test_killed_worker_process_group_is_reaped` 13.86s, `test_real_worker_completes_a_scored_episode` 6.67s.
- **Whole-tree static gates, exactly as CI runs them:** `flake8` clean · `black==26.1.0 --check` clean (719 files) · `isort==5.13.2 --check` clean · `pyright` **0 errors, 0 warnings**.
- **Full suite not re-run** per scope.
- **CI:** every check green on this head except `test (3.12)`, whose only failure is the known `tests/unit/tui/test_steering_approval.py` flake. This diff touches five files, all under `evaluation/runner/` and their tests, and zero TUI files. Unrelated.

## Two pre-existing items, for the record — not findings against this PR

Both reproduce identically at the base commit `c876db3c` (verified on a separate base worktree), so neither is introduced here and neither should hold the merge. Recording them in-thread rather than as tickets:

1. `EvidenceWriter.abandon` re-raises `OSError` as well as `EvidenceError` (`store.py:1583`), and `_abandon_for_evidence` catches only `EvidenceError` — so a disk-full at the terminal write escapes `run()` as a bare `OSError` instead of returning an outcome. Reproduces at base. Mitigated in practice: rescue and teardown both complete first, so nothing leaks. One-line fix if wanted: catch `(EvidenceError, OSError)`.
2. `_type_name` resolves `tuple/list/set/frozenset` origins but not abstract ones — `Sequence[str]` still renders as `str`. Unreachable today (the protocol uses concrete `tuple[str, ...]` everywhere); it would only bite if a future protocol field is annotated abstractly.

---

**Blockers: 0 · Majors: 0 · Minors: 0 · Nits: 0**

**Verdict: clean — this round is TERMINAL.** The M3 fix is correct, its reasoning is sound (and its central premise is stronger than the remediation claimed), the test that pins it fails when reverted and exercises a real publish-then-append, and m3/m4/n2 are each fixed and pinned. The `abandonment_failed` design is right as built. The residual orphan state is recoverable and correctly out of scope; the two items above are pre-existing and deferrable. Nothing here justifies a round 4.

**Recommend merging** once `test (3.12)` settles on the known flake.

damianvtran:
### Rebased onto main — fresh review round needed

Main advanced to `98831a8a` (0.44.21, three TUI PRs: #504, #513, #491) during round 3, and the branch conflicted. Rebased; PR is `MERGEABLE` again.

**New head: `710b2388520c36ba2e92cfae4b9b2309048aef4e`** · base `main` @ `98831a8a` · version **0.44.22**

**What conflicted and how I resolved it**

- **`pyproject.toml` version** — my branch carried a `chore(release): 0.44.20` commit against a tree now at 0.44.21. I **dropped the stale bump during the rebase** rather than layering a second one, then re-bumped once on top of main. Exactly one `chore(release)` commit remains in the branch, now `0.44.22` (0.44.21 is the latest tag and is live on PyPI).
- **`[tool.pytest.ini_options] markers`** — resolved to contain **both**: main's `e2e` marker and my `slow` marker. My whole delta against main's `pyproject.toml` is now the four added `slow` lines and the version:

```diff
     "e2e: headless end-to-end tests that drive the assembled TUI application",
+    # Kept IN the default run: these spawn real worker processes and are the
+    # only coverage of the actual subprocess boundary, so they are marked for
+    # selection rather than exclusion.
+    "slow: tests that spawn real subprocesses and take seconds rather than ms",
 ]
```

Nothing else conflicted. Verified my source and test files are **byte-identical** across the rebase (`git diff backup..HEAD -- local_operator/evaluation tests/unit/evaluation` is empty), so no round-1/2/3 fix was disturbed. Against main I contribute 13 files: the five runner modules, their six test files, and `pyproject.toml`.

**One new commit beyond the rebase — `710b2388`.** Verifying post-rebase I hit a single intermittent failure in the subprocess cutpoint matrix: `failed_pre_bundle` where a later cutpoint was expected. Cause is the test budget, not the runner — those tests inherited the in-process 5s default, but a *real* worker must start CPython and import pydantic and `local_operator` before answering the handshake, which can exceed 5s under parallel-agent load and fails the episode at `prepare` before it reaches the cutpoint under test. Raised to 60s for the real-worker tests only, via a `_subprocess_config` helper with the reasoning at the site. Stable across three consecutive full runs of `tests/unit/evaluation` afterwards (413 passed each).

**Verification on the new head**

- **Focused runner suite: 59 passed, 0 skipped** (`-n0`, 82.6s). The **7 real-subprocess tests genuinely ran** — `--durations`: five cutpoints 11.48–11.75s, `test_killed_worker_process_group_is_reaped` 11.51s, `test_real_worker_completes_a_scored_episode` 5.65s.
- **Evaluation package: 413 passed** (×3 consecutive runs).
- Main's newly-landed TUI tests pass alongside my markers block: `test_slash_echo.py` + `test_subagent_view.py` → **169 passed**.
- **Whole-tree gates, as CI runs them:** `flake8` clean · `black==26.1.0 --check` clean (**721** files, up from 719 with main's new files) · `isort==5.13.2 --check` clean · `pyright` **0 errors, 0 warnings**.

**A fresh review round is needed: the head moved.** Round 3 was terminal against `5b614549`, which is no longer head. The evaluation diff is byte-identical, so re-review scope is the rebase itself — the pyproject resolution and `710b2388`.

damianvtran:
### Agent review — round 4

Scope: `5b614549b781341b1e871103f736f671a948db9b`..`710b2388520c36ba2e92cfae4b9b2309048aef4e`

Reviewed in a clean detached worktree at the exact head (`710b2388`, confirmed against the PR's `headRefOid`). `5b614549` is **not** an ancestor of head — the rebase is real, `merge-base(origin/main, HEAD)` is `98831a8a`, so I verified the carry-over claim by content rather than by topology. Scope held deliberately narrow; round 3's conclusions are not re-litigated.

**No findings.** All four rebase claims check out, and the one unrequested commit is justified.

---

## 1. Byte-identity across the rebase — CONFIRMED

Compared blob hashes at `5b614549` vs `710b2388` for all 12 evaluation runner files. **11 of 12 are identical blob-for-blob**; the only change is the timeout commit's own file:

| File | Result |
| --- | --- |
| `runner/{__init__,episode,model,provider_client,responder}.py` | **IDENTICAL** (all 5) |
| `tests/.../runner/{__init__,conftest,test_episode,test_isolation,test_provider_client,test_reconciliation}.py` | **IDENTICAL** (all 6) |
| `tests/.../runner/test_episode_subprocess.py` | changed — `2b934522` → `582acdf7`, **exactly and only `710b2388`** |

No runner file changed beyond the timeout commit. **Round 3's verification carries**, and the rounds 1–3 fixes survive structurally rather than by assertion — I still spot-checked B1 (`parse_decision` builds `protocol_version`, no bogus `observation_sequence` key), m1 (`provider_request_id` carried, degrades to `"unknown"`), and M3 (`_AMBIGUOUS_FINALIZATION_ISSUES` is still just `("score_invalid","state.json")` / `("finalization_invalid","events.jsonl")` — `artifact_unreferenced` was **not** slipped into the allowlist). Isolation tests pass. Nothing regressed.

## 2. Exactly one release commit, correct PATCH — CONFIRMED

`git log --grep="chore(release)" origin/main..HEAD` returns **exactly one** commit, `81934772` (`0.44.22`). The branch's entire `pyproject.toml` version delta is `-0.44.21` / `+0.44.22` — **no stale `0.44.20` bump survives anywhere in the branch history**. Correct next PATCH on top of main's `0.44.21`, and correct per AGENTS.md materiality (internal subsystem, no user-facing surface).

## 3. Markers resolution preserves both — CONFIRMED, and main's `e2e` still behaves

The delta against main's `pyproject.toml` is the four added `slow` lines plus the version; `addopts` retains `-m "not e2e"` unmodified. Verified by execution, not inspection:

```
tests/e2e (default)          → no tests collected (3 deselected)   # e2e still excluded
tests/e2e -m e2e             → 3 tests collected                   # still selectable
runner -m slow               → 7/59 collected (52 deselected)      # slow selects the real-worker set
runner (default)             → no unknown-marker warnings
```

Both CI `tui-e2e` legs (macos + ubuntu) are **green on this head**, which is the load-bearing confirmation that main's marker is intact.

## 4. The 60s timeout — legitimate, not masking a defect

I probed this as a potential defect-hider and it is not one. Three things settle it:

- **It is below production sizing, not above it.** `EpisodeConfig` already defaults to `prepare/reset/step/score = 120.0` and `handshake_timeout = 30.0` (`episode.py:174-180`). The 5s in the test helper was a *test-local tightening* for a fake in-process adapter, never the product's budget. Raising the real-worker tests to 60s moves them **toward** production values — a test that failed at 5s was failing on a budget no real deployment uses.
- **Scoped to exactly the three real-worker tests.** `_subprocess_config` is called at `:379`, `:427`, `:497` and nowhere else; `build_config`'s 5s defaults are **untouched**, so every in-process test keeps the tight budget. No timeout coverage is weakened — and since these three drive a real spawn, they were never the tests pinning timeout *behaviour* anyway.
- **The time is spend on process work, not runner logic.** The cutpoint tests kill a real worker mid-episode and reap its process group; the cost is interpreter spawn, a copied `--copies` venv, imports and teardown. The runner's own path is unchanged (`episode.py` is byte-identical), so there is no slowness here for a timeout to conceal.

One honest calibration on the commit message rather than the change: measuring the fixture's own interpreter-selection path, a cold copied-venv `import pydantic, local_operator` runs **~0.2s median / 0.53s max** — so bare spawn+import alone does not approach 5s. The real pressure is the *aggregate* of many per-call budgets across a full kill-and-reap episode under contention, which the observed durations bear out. The change is right; its stated mechanism is narrower than the true one. Not a finding, and not worth a commit amend.

## 6. Gates and suite

Focused runner suite, whole-tree static gates, all on this head:

```
runner suite   59 passed in 459.81s   # all 7 real-subprocess tests RAN (-m slow set, none skipped)
flake8 .                      exit 0
black==26.1.0 --check .       721 files unchanged
isort==5.13.2 --check .       clean
pyright                       0 errors, 0 warnings, 0 informations
```

**Caveat on timing, not on outcome:** this machine ran at load average 91–119 on 14 cores throughout (parallel agents). The six cutpoint tests took ~70s each under that contention — the pass/fail result is sound, but treat those durations as a starved upper bound, not a baseline. That contention is also the concrete case the timeout commit exists to absorb.

**CI on this head:** `lint`, `type-check`, `pip-audit`, `tui-e2e (macos-latest)`, `tui-e2e (ubuntu-latest)` all **pass**; `test (3.12)` still running at review time — not duplicated here.

---

## Verdict

**Blockers: 0 · Majors: 0 · Minors: 0 · Nits: 0**

`clean` — and this round is **TERMINAL**. The rebase carried round 3's terminal result intact: the evaluation source and tests are byte-identical, the version and marker resolutions are both correct, and the one added commit is a properly scoped test-only change that moves the real-worker tests toward production timeout values without weakening in-process coverage. Nothing outstanding, nothing deferred. **Recommend merge** once `test (3.12)` lands green.

PR #520, opened 2026-09-01
feat(evaluation): carry the parent-chosen artifact root to out-of-process adapters

damianvtran:
## Summary

The parent reads adapter-produced observation frames with `verify_artifact(self._config.artifact_root, frame.artifact)` (`runner/episode.py`), but **`artifact_root` existed only on `RescueDescriptor`** — `PrepareParams` and `ResetStartParams` did not carry it. The spawned worker's environment is deliberately built from a closed allowlist (`supervisor._ENV_ALLOW` is locale and temp only), so a genuinely out-of-process adapter had **no way to learn where to write frame bytes**.

The consequence was that only in-process and rescue paths could deliver observation frames the parent could read. This was found while building the first real adapter (OSWorld), whose end-to-end sealed-bundle proof had to run **in-process** because of it.

This PR closes that gap and fixes a liveness hazard the gap was hiding. It does not activate a benchmark, OSWorld, provider, CLI, configuration, tool, cloud, or UI path.

## The design

**`artifact_root` is carried on `ResetStartParams` only — not on `PrepareParams`.** The path must be one the parent chooses and the worker is told, arriving as a validated RPC field rather than through the environment, and `reset_start` is the correct carrier:

- **`prepare` is contractually allocation-free.** It creates nothing and therefore produces no observation and has no bytes to publish. Handing it a writable root would invite exactly the allocation the two-stage rescue persistence depends on it not doing, widening the window in which a resource exists that no persisted descriptor names.
- **`reset_start` is both the side-effect boundary and the first call that yields an observation.** The parent calls `observe` immediately after it, and every later `observe`/`execute` runs in `RUNNING`, which only `reset_start` can enter. So it is the single point that precedes every frame-producing call while still being the first one that needs the root.
- **`RescueDescriptor` keeps its own copy** because rescue runs in a *new* process against a persisted file, with no live session to have been told.

**One shared validator, not two.** `RescueDescriptor`'s bespoke `artifact_root` field validator was extracted into a single `ArtifactRoot` annotated type (absolute *and* normalized, `min_length=1`, `max_length=4096`) now used by both models. A second, slightly-different definition of what makes a root acceptable is exactly how a confinement boundary develops a gap. Normalization is load-bearing rather than cosmetic: the parent opens this directory with `O_DIRECTORY` and resolves artifact names against that descriptor, so a root containing `..` could denote a different directory than the one the parent authorized.

**The parent creates the directory it names.** `_reset_and_observe` now does `mkdir(mode=0o700, parents=True, exist_ok=True)` before the call. An adapter told to publish into a missing root would otherwise have to create it itself, which hands the worker a say in the mode and ownership of the one directory the parent later opens and reads.

**Write-location and read-location cannot drift.** `VerifiedAdapterSession.reset_start` refuses a reset whose `artifact_root` differs from the one its `HostVerifier` reads from, alongside the existing task/episode check. Pointing the adapter at one directory while verifying another would otherwise produce silent `artifact path is unsafe` failures at the first frame — and worse, a root that merely *resolves* to the verifier's would make the mismatch depend on symlink state at call time.

**`verify_artifact` is not weakened.** The parent still resolves every artifact by content digest, inside that one directory descriptor, with `O_NOFOLLOW`, checking digest, byte count, and media. Sending the worker the root is what makes it able to write where the parent already looks; it grants no new read authority.

## Schema version 1.0 → 1.1

Adding a **required** field to a model whose config is `extra="forbid"` and `strict=True` is wire-breaking in **both** directions: a 1.0 worker rejects the new key as an unknown extra, and a 1.1 worker cannot serve a parent that omits it. This boundary deliberately pins exact versions rather than accepting ranges (see `AdapterSelector`), so the bump is what locks a 1.0 adapter out cleanly instead of letting it fail later as an opaque `invalid_request`.

The three models that pin the version (`AdapterSelector`, `AdapterMetadata`, `RescueDescriptor`) now share one `SchemaVersion` alias, so a future bump cannot move the constant while leaving a model silently accepting the older literal. The runner likewise now tracks `ADAPTER_SCHEMA_VERSION` instead of restating `"1.0"` when minting rescue descriptors — that hardcoded literal was found by this bump, where it surfaced as an unrelated `failed_pre_bundle` validation error.

Audited: no stranded `"1.0"` remains on the adapter schema.

## BLOCKER found in review: a worker could wedge the parent forever with a FIFO

This is a **liveness** hazard, not a content one, and it is called out here rather than buried because its failure mode is a permanent hang of the parent process.

`verify_artifact` opened the artifact with `O_RDONLY | O_CLOEXEC | O_NOFOLLOW` — **without `O_NONBLOCK`**. An adapter could `mkfifo(root/<honest-looking-digest>)`; opening a FIFO with no writer blocks in the kernel indefinitely, and the `S_ISREG` check that would have refused it **sits behind that blocking open**.

Nothing upstream could recover. `verify_artifact` is called synchronously from `_record_observation` (`runner/episode.py`) and `_validate_observation_content` (`supervisor.py`) **on the event-loop thread**, and **after** `_mutating_call`'s `wait_for` has already closed — so an outer `wait_for` never fires. No timeout, no poison, no rescue, no process-tree teardown.

**This PR is what made it reachable.** `verify_artifact` is untouched by this change, but before it a spawned worker had no way to learn the root at all. The gap closed here and this hazard are the same door, so it is fixed here rather than deferred.

**The fix is one flag:** `getattr(os, "O_NONBLOCK", 0)` added to the open flags. Verified behaviour-preserving on regular files, where POSIX makes it a no-op — measured as the same `S_ISREG`, the same size (69), and the same bytes with and without it. The FIFO open now returns immediately and falls into the **existing** `"artifact is not a matching regular file"` refusal, so no new error path was needed.

## Testing evidence

**A real spawned worker delivers a verified frame end to end.** `test_real_worker_publishes_a_frame_the_parent_verifies_and_bundles` launches a genuine interpreter running a genuine installed distribution through `AdapterSupervisor`, with the environment built from the production closed allowlist. The adapter publishes a **real 69-byte PNG** (constructed with correct CRCs, since the parent runs the real media validator over it) into the **parent-supplied root**, named by its own digest. Asserted: the bytes landed in the parent's directory under their true digest; the parent read and accepted them; both observation events carry the frame as `image/png`/69 bytes; the bundle stores those exact bytes content-addressed; and `verify_bundle().valid`.

The frame mode is armed through a file beside the adapter module — the same adapter-owned mechanism the existing cutpoint uses, because an env var cannot survive the allowlist. Default mode is empty, so every pre-existing test in that module keeps its exact previous behaviour.

**Proven to bite.** Denying the adapter the field fails the test with *"the spawned worker published nothing into the parent's root"* — the precise pre-PR state.

**Five confinement cases, each pinned to the exact refusal it should reach**, so a weakened check cannot be masked by another clause still holding:

| case | refusal |
| --- | --- |
| bytes published outside the root | `artifact path is unsafe or unavailable` |
| in-root symlink to bytes outside it | `artifact path is unsafe or unavailable` |
| same-length digest tampering | `artifact digest differs` |
| lied-about byte count | `artifact is not a matching regular file` |
| FIFO nobody writes to | `artifact is not a matching regular file` |

Pinning the message caught a real defect in my own test: length-*changing* tampering was being rejected by the size check first, leaving the digest comparison entirely unexercised while the suite stayed green. It is now same-length tampering and genuinely reaches the digest clause.

**All proven load-bearing by mutation.** Dropping `O_NOFOLLOW` let the symlink escape reach `completed`; removing the digest comparison turned the digest case red; removing `O_NONBLOCK` made the FIFO case **hang until killed at 120s** rather than fail an assertion — which is the hazard itself, and why that probe was run under a hard timeout.

The `fifo` case is the odd one out and its docstring says so plainly: it attacks liveness rather than content and shares its refusal with `byte_count_mismatch`, so what it actually asserts is that the episode *terminates at all*.

**Suite:** `tests/unit/evaluation` **419 passed**, against a 413 baseline captured on `origin/main` before any change — the six new tests, no regressions.

**Gates**, whole tree, exactly as CI spells them: `flake8` clean, `black==26.1.0` clean (722 files), `isort==5.13.2` clean, `pyright` 0 errors. `uv build` + `uvx twine check` both PASSED on the sdist and the wheel.

## Known consistency point

`exist_ok=True` on the parent `mkdir` does **not** re-apply the mode, so a pre-existing `0o777` root stays `0o777`. This matches `persist_rescue`'s existing behaviour, so it is consistent with the boundary rather than a regression, and is left alone deliberately.

## Not activated

- **Nothing outside the evaluation package imports it.** Verified by search: no reference to `local_operator.evaluation` exists in `local_operator/` outside `local_operator/evaluation/` itself.
- The evaluation subsystem still has no production caller, which is why this is a **patch** despite the `ADAPTER_SCHEMA_VERSION` change — no released adapter can be stranded by it.

## Follow-on

The OSWorld adapter branch depends on this landing first: it needs `schema_version="1.1"` and to read the root from `reset_start`. That branch is untouched here.

### DIFF
--- a/local_operator/evaluation/adapters/api.py
+++ b/local_operator/evaluation/adapters/api.py
@@ -12,9 +12,9 @@ import platform
 import sys
 from collections.abc import Mapping
 from pathlib import Path
-from typing import Any, Literal, Protocol, TypeAlias, runtime_checkable
+from typing import Annotated, Any, Literal, Protocol, TypeAlias, runtime_checkable
 
-from pydantic import Field, field_validator, model_validator
+from pydantic import AfterValidator, Field, field_validator, model_validator
 
 from local_operator.evaluation.evidence.models import ScoreArtifact, canonical_digest
 from local_operator.evaluation.lifecycle import CleanupPlan
@@ -26,7 +26,17 @@ from local_operator.evaluation.receipts import (
     StrictIdentifier,
 )
 
-ADAPTER_SCHEMA_VERSION = "1.0"
+# 1.1 adds the parent-chosen ``artifact_root`` to ``ResetStartParams``. It is a
+# REQUIRED field on a model whose config is ``extra="forbid"`` and ``strict``,
+# so the change is wire-breaking in both directions: a 1.0 worker rejects the
+# new key as an unknown extra, and a 1.1 worker cannot serve a parent that omits
+# it. This boundary deliberately pins exact versions rather than accepting
+# ranges (see ``AdapterSelector``), so the bump is what locks a 1.0 adapter out
+# instead of letting it fail later as an opaque invalid_request.
+ADAPTER_SCHEMA_VERSION = "1.1"
+# One alias for the three models that pin the version, so a future bump cannot
+# move the constant while leaving a model silently accepting the older literal.
+SchemaVersion: TypeAlias = Literal["1.1"]
 ADAPTER_ENTRY_POINT_GROUP = "local_operator.evaluation_adapters.v1"
 MAX_RESCUE_REFS = 256
 MAX_REQUIREMENTS = 256
@@ -110,7 +120,7 @@ def canonical_params_digest(method: AdapterMethod, params: ProtocolModel) -> Dig
 class AdapterSelector(ProtocolModel):
     """Exact wheel and interpreter selection; ranges and fallback are absent."""
 
-    schema_version: Literal["1.0"]
+    schema_version: SchemaVersion
     adapter_id: StrictIdentifier
     distribution: StrictIdentifier
     version: StrictIdentifier
@@ -154,7 +164,7 @@ class AdapterMetadata(ProtocolModel):
     entry_point: StrictIdentifier
     package_digest: Digest
     release_digest: Digest
-    schema_version: Literal["1.0"]
+    schema_version: SchemaVersion
     capabilities: AdapterCapabilities
 
 
@@ -248,28 +258,42 @@ class ScopedInfraValue(ProtocolModel):
     value: str = Field(min_length=1, max_length=4096)
 
 
+def _normalized_absolute_root(value: str) -> str:
+    # Normalized AND absolute, not merely absolute: the parent opens this
+    # directory with O_DIRECTORY and then resolves artifact names against that
+    # descriptor, so a path containing ".." would let the root itself denote a
+    # different directory than the one the parent believes it authorized.
+    if not os.path.isabs(value) or os.path.normpath(value) != value:
+        raise ValueError("artifact root must be normalized and absolute")
+    return value
+
+
+# The single definition of a parent-authorized artifact directory. Both the
+# rescue descriptor and the live episode's reset carry one, and they must agree
+# on what makes a root acceptable -- a second, slightly different validator is
+# exactly how a confinement boundary develops a gap.
+ArtifactRoot: TypeAlias = Annotated[
+    str,
+    Field(min_length=1, max_length=4096),
+    AfterValidator(_normalized_absolute_root),
+]
+
+
 class RescueDescriptor(ProtocolModel):
-    schema_version: Literal["1.0"]
+    schema_version: SchemaVersion
     selector: AdapterSelector
     handshake: Handshake
     episode_id: StrictIdentifier
     cleanup_plan: CleanupPlan
     secret_refs: tuple[SecretRef, ...] = Field(max_length=MAX_RESCUE_REFS)
     infra_values: tuple[ScopedInfraValue, ...] = Field(max_length=MAX_RESCUE_REFS)
-    artifact_root: str = Field(min_length=1, max_length=4096)
+    artifact_root: ArtifactRoot
     descriptor_id: Digest = ZERO_DIGEST
 
     @field_validator("secret_refs", "infra_values", mode="before")
     @classmethod
     def _freeze_refs(cls, value: Any) -> Any:
         return tuple(value) if isinstance(value, list) else value
-
-    @field_validator("artifact_root")
-    @classmethod
-    def _absolute_artifact_root(cls, value: str) -> str:
-        if not os.path.isabs(value) or os.path.normpath(value) != value:
-            raise ValueError("artifact root must be normalized and absolute")
-        return value
 
     @model_validator(mode="after")
     def _content_bind(self) -> "RescueDescriptor":
@@ -364,8 +388,38 @@ class PrepareResult(ProtocolModel):
 
 
 class ResetStartParams(OperationParams):
+    """Begin the episode, and tell the worker where its frame bytes may go.
+
+    WHY ``reset_start`` AND NOT ``prepare``. The worker's environment is built
+    from a closed allowlist (``supervisor._ENV_ALLOW`` is locale and temp only),
+    so a genuinely out-of-process adapter has no ambient way to learn the
+    directory the parent will read frames from. The path therefore has to arrive
+    as validated RPC input, and this is the call it belongs on:
+
+    * ``prepare`` is contractually ALLOCATION-FREE -- it creates nothing, so it
+      produces no observation and has no bytes to publish. Handing it a writable
+      root would invite exactly the allocation the two-stage rescue persistence
+      depends on it not doing, and would widen the window in which a resource
+      exists that no persisted descriptor names.
+    * ``reset_start`` is the side-effect boundary AND the first call that yields
+      an observation: the parent calls ``observe`` immediately after it, and
+      every later ``observe``/``execute`` runs in ``RUNNING``, which only
+      ``reset_start`` can enter. So it is the single point that precedes every
+      frame-producing call while still being the first one that needs the root.
+    * ``RescueDescriptor`` keeps its own copy because rescue runs in a NEW
+      process against a persisted file, with no live session to have been told.
+
+    Confinement is not established by this field. The parent treats the root as
+    the only directory it will ever open, and resolves each artifact by its
+    content digest relative to that directory descriptor with ``O_NOFOLLOW``
+    (``supervisor.verify_artifact``); the worker names bytes by digest and never
+    supplies a path. Sending the root is what makes the worker able to write
+    where the parent already looks -- it grants no new read authority.
+    """
+
     task_id: StrictIdentifier
     episode_id: StrictIdentifier
+    artifact_root: ArtifactRoot
 
 
 class ObserveParams(ProtocolModel):
--- a/local_operator/evaluation/adapters/supervisor.py
+++ b/local_operator/evaluation/adapters/supervisor.py
@@ -627,6 +627,15 @@ class VerifiedAdapterSession:
             self.verifier.episode_id,
         ):
             raise SupervisionError("reset belongs to another task or episode")
+        # The root announced to the worker and the root the verifier reads from
+        # are the same authorization, so they are checked against each other
+        # rather than trusted to have been populated consistently. A caller that
+        # points the adapter at one directory while verifying another would get
+        # silent "artifact path is unsafe" failures at the first frame; worse, a
+        # root that merely resolves to the verifier's would make the mismatch
+        # depend on symlink state at call time.
+        if params.artifact_root != str(self.verifier.artifact_root):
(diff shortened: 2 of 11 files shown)


damianvtran:
### Agent review — round 1

Scope: `b64eb01c90e5c9e66d586e5ad04f23c80809d01c..76c574a0aa264626eb28ad271db835cb90177d88`

Independent round against the final head. Base verified as the merge-base with `origin/main`; reviewed in a clean detached worktree at `76c574a0`, restored to pristine after every mutation probe. This is a fresh round: the reviewed head equals the current PR head.

## No findings

No blocker, no major, no minor, no nit worth a remediation round.

## The FIFO blocker — reproduced independently, not taken on trust

I reproduced the hazard twice, from both ends, each under a hard bounded timeout so a wedge could not take my own run with it.

**1. Isolated, outside the codebase.** I re-implemented `verify_artifact`'s exact open sequence and ran it against a FIFO nobody writes to, with and without the flag:

| probe | result |
| --- | --- |
| FIFO, `O_NONBLOCK` | `REFUSED:not-a-regular-file` — returns immediately |
| FIFO, blocking | **HUNG — killed at a hard 15s bound, no output** |
| regular file, `O_NONBLOCK` | `OPENED size=4096 match=True` |
| regular file, blocking | `OPENED size=4096 match=True` |

**2. In the real episode.** Removing only the `| getattr(os, "O_NONBLOCK", 0)` line from `supervisor.verify_artifact` and running the `fifo` confinement case: **hung until killed at a hard 150s bound.** With the flag restored, the same case passes. The hazard is real, and this one line is what prevents it.

**Behaviour-preserving on regular files: confirmed.** Rows 3 and 4 above are byte-identical — same `S_ISREG`, same `st_size`, same content hash. `O_NONBLOCK` is a POSIX no-op on a regular file, so this buys liveness at zero semantic cost. The comment at supervisor.py:797-808 explains *why* the check order forces this (the `S_ISREG` test sits behind the open) and *why* nothing upstream can rescue it — `verify_artifact` runs on the event-loop thread after `_mutating_call`'s `wait_for` has closed. That reasoning holds as written.

**The 5th case is honest about itself.** The docstring at test_episode_subprocess.py:789-793 states plainly that `fifo` attacks liveness rather than content and *shares its refusal with `byte_count_mismatch`*, so what it really asserts is that the episode terminates at all. That is the correct disclosure — its failure mode is a hang, not a red assertion, and the docstring says so rather than implying a distinct refusal. Exactly the kind of thing that would otherwise rot into a test nobody understands.

## Real-spawn frame test — proven load-bearing by mutation

The spawn is genuine: a copied `--copies` venv interpreter (`_real_interpreter`), a real installed distribution with a pinned `package_digest`, and the production `_ENV_ALLOW` closed allowlist (`{LANG, LC_ALL, LC_CTYPE, TMPDIR, TEMP, TMP}` — locale and temp only, supervisor.py:85). So the RPC field really is the worker's only route to the root; no env var could substitute.

I proved the test bites two independent ways:

- **Adapter ignores the field** (`self.artifact_root = None`) → `AssertionError: the spawned worker published nothing into the parent's root`. That is precisely the pre-PR state this PR closes.
- **Parent announces a different root** (`/tmp` instead of the config root) → `SupervisionError: reset artifact root differs from the verified root`, episode `failed`. The `VerifiedAdapterSession.reset_start` equality guard (supervisor.py:637) is doing real work, not decoration.

## All 5 confinement cases reach their intended clause

Each is pinned to its exact refusal string, so a weakened check cannot hide behind another clause still holding. The digest case is the one that previously tripped the size check first — the PR discloses this and it is genuinely fixed: `tampered = payload[:-1] + bytes([payload[-1] ^ 0xFF])` is same-length tampering, so it passes the `st_size` comparison and lands on `artifact digest differs`. Confirmed reachable, not merely asserted.

## My own attacks on the new surface

I attacked the `ArtifactRoot` validator directly rather than reading it. Rejected: relative, `..`, `.` segments, double slash, trailing slash, tilde, leading space, empty, over-length. Accepted only normalized absolute paths. Two observations, both benign and worth recording rather than fixing:

- **Embedded NUL / newline are accepted by the validator** (`/tmp/a\x00b`). Not exploitable: the NUL is rejected by the kernel at `os.open` and surfaces as the standard `artifact path is unsafe or unavailable` refusal, and the root is parent-chosen, never worker-supplied — the worker cannot influence this value at all.
- **A symlinked directory as the root is accepted** (`mkdir(exist_ok=True)` follows it). Also not a bypass: the parent chooses the root, the equality check is string-exact against the verifier's own value so the two cannot drift, and confinement is enforced per-artifact by `O_NOFOLLOW` resolved against the root *descriptor*, not by the root's own path shape.
- **Loose pre-existing mode is not tightened.** I confirmed the PR's disclosed consistency point empirically: `mkdir(mode=0o700, exist_ok=True)` leaves a pre-existing `0o777` directory at `0o777`. This matches `persist_rescue`'s existing behaviour exactly, so it is consistent with the boundary rather than a new gap. Correctly left alone and correctly disclosed.
- **Bad root fails cleanly.** A root that exists as a regular file raises `FileExistsError` from the `mkdir`, which lands inside `_run_bundle`'s `except BaseException → _finalize_failure` — an episode failure, not an escape.

Digest names are pinned to `^[0-9a-f]{64}$` (receipts.py:54), so no artifact name can carry a traversal component in the first place.

**Design confirmed at this head:** `artifact_root` is on `ResetStartParams` only — `PrepareParams.model_fields` is `['operation_id', 'episode_id', 'secret_refs', 'infra_values']`, so the allocation-free contract is intact. One shared `ArtifactRoot` alias now serves both models, and the old duplicate `RescueDescriptor` validator is deleted rather than left alongside it.

## Schema and version audit

The bump is applied consistently. `SchemaVersion: TypeAlias = Literal["1.1"]` is used by all three pinning models (`AdapterSelector`, `AdapterMetadata`, `RescueDescriptor`), so the constant and the models cannot drift on the next bump — and `_persist_descriptor` now tracks `ADAPTER_SCHEMA_VERSION` instead of restating a literal. The three remaining `"1.0"` literals in the package are **different version axes** (`EVIDENCE_SCHEMA_VERSION`, `PROTOCOL_VERSION` in `protocol.py` and `provider_client.py`) and correctly do not move with this one. No adapter is silently stranded: a 1.0 selector is refused at the exact-match boundary rather than failing later as an opaque `invalid_request`, and the missing field is rejected as `missing` on a strict `extra="forbid"` model in both directions — I verified both by construction.

**Version:** `0.44.23 → 0.44.24`, the correct next PATCH against `origin/main` and the `v0.44.23` release, in exactly one `chore(release)` commit. PATCH is right per AGENTS.md "Versioning": nothing outside `local_operator/evaluation/` imports this package, so despite the wire-breaking schema bump there is no production caller and no user-facing capability change. Materiality, not commit type — correct call.

## Gates

Run whole-tree from a clean worktree at the PR head, with the CI-pinned tool versions (`black==26.1.0`, `isort==5.13.2`, `flake8==7.1.0`, `pyright==1.1.411`):

| gate | result |
| --- | --- |
| flake8 | clean |
| black --check | clean, 722 files |
| isort --check-only | clean |
| pyright | **0 errors, 0 warnings** |
| `tests/unit/evaluation` | **419 passed** in 184s |

419 matches the PR's claim against its 413 baseline — the six new tests, no regressions. The run was not starved (3m05 wall clock), so both pass/fail and timing are meaningful here.

**CI on this head:** lint, type-check, pip-audit, and both tui-e2e jobs pass; `test (3.12)` was still pending when I checked. CI is authoritative for the full suite — merge on its completion.

**No regression to the prior boundary hardening.** `verify_artifact` is otherwise untouched — `O_NOFOLLOW`, `O_CLOEXEC`, the `O_DIRECTORY` root descriptor, the size ceiling on the read loop, digest comparison, and media validation are all unchanged; the only edit is one added flag. Isolation, poison gating, rescue, and process-tree teardown are untouched by this diff.

## Verdict

**clean — 0 blockers, 0 majors, 0 minors, 0 nits.**

This round is **TERMINAL**. Nothing remains to remediate and no further round is warranted. The one blocker from the pre-PR round is genuinely fixed and I reproduced both its failure and its fix myself rather than trusting the report. The tests are load-bearing under mutation, the disclosures in the PR body are accurate — including the two that admit a limitation rather than hide it — and the version choice is correct.

**Recommend merging** once `test (3.12)` completes green.

Issue #518, opened 2026-09-01
[FEATURE] Composer: bind alt+b / alt+f for word movement (the encoding a default Ghostty sends)


bbqben:
## Feature Summary

Bind `alt+b` / `alt+f` to word-wise caret movement in the composer, alongside the `alt+arrow` and `ctrl+arrow` chords added in #375. These are the readline word-motion keys, and on a **default Ghostty install** they are the only encoding `⌥←`/`⌥→` ever reaches the app as.

## Detailed Description

#375 covered three encodings of `option+arrow`. There is a fourth, and it is the one a default Ghostty produces.

Ghostty ships these as **factory defaults** (verified with `ghostty +show-config --default`, which ignores user config — Ghostty 1.3.1):

```
keybind = alt+arrow_left=esc:b
keybind = alt+arrow_right=esc:f
```

So Ghostty rewrites `⌥←` into readline's meta-b *before any application sees it*, discarding the fact that it was ever an arrow. What lop receives then depends on the keyboard protocol:

| terminal state | bytes on the wire | Textual key | bound in lop today? |
|---|---|---|---|
| legacy encoding | `\x1bb` | `ctrl+left` | ✅ yes (inherited from `TextArea`) |
| **kitty keyboard protocol** | **`\x1b[98;3u`** | **`alt+b`** | ❌ **no** |
| `⌥→`, kitty protocol | `\x1b[102;3u` | `alt+f` | ❌ no |

Textual negotiates the kitty protocol, so **lop lands in the broken row whenever the terminal speaks it**. Reproduced with lop v0.44.23 / textual 8.2.8:

```
b'\x1b[1;3D'  -> alt+left   bound=True    (only if the user overrides Ghostty's default)
b'\x1bb'      -> ctrl+left  bound=True    (legacy path — this is why it often "works")
b'\x1b[98;3u' -> alt+b      bound=False   <-- Ghostty default + kitty protocol
b'\x1b[102;3u'-> alt+f      bound=False
```

The legacy row is why this is easy to miss: in a terminal that has not negotiated the kitty protocol, the same corrupted `\x1bb` decodes to `ctrl+left` and word motion works by accident. Enabling the kitty protocol removes that accident.

Captured from a real keyboard (macOS 15, Ghostty 1.3.1, `macos-option-as-alt = true`), same key, two protocol states:

```
legacy mode:   ⌥←  ->  b'\x1bb'         -> ctrl+left  -> works
[...]

PR #522, opened 2026-09-01
feat(evaluation): OSWorld 2.0 benchmark adapter, proved out-of-process without the cloud

damianvtran:
## Summary

The first real benchmark adapter: **OSWorld 2.0** (`osworld-v2-2026.08.08`), AWS-targeted, as a separate RECORD-pinned distribution at `benchmarks/osworld_v2_adapter/` with its own pinned lock and `adapter-release.json`. It is loaded through the released verified-adapter boundary — the same one #520 completed — and proved end to end by a **genuinely spawned worker** that seals a bundle the independent verifier accepts.

This PR lands the adapter **without the cloud**. There is no AWS provider, no boto3 call, and no spend: a `FakeProvider` standing behind the same `EnvironmentProvider` protocol drives a real `EpisodeRunner` through a complete episode. Every risky decision — resource refs, score mapping, action compilation, requirement derivation — is settled here, where a mistake is free.

**Nothing outside the benchmarks workspace imports it.** No CLI, TUI, mobile, tool, configuration, or cloud path is activated. Outside `benchmarks/` and its tests under `tests/unit/evaluation/adapters/osworld/`, this PR touches exactly four files, all adapter-related: the version bump and a `[tool.pyright] extraPaths` entry in `pyproject.toml`, a scoped `uv.lock` exception in `.gitignore`, a `make adapter-osworld` target in the `Makefile`, and the new `scripts/build_osworld_adapter.py`. (An earlier revision of this body under-stated that list; corrected in round-1 remediation.)

## Why a separate distribution, in-tree

The adapter imports `local_operator.evaluation.adapters.api` to build its `AdapterMetadata`, so it is coupled to the protocol version and must move with it — that argues for in-tree. But `discovery.distribution_digest` pins the adapter by verifying **every RECORD row of the installed wheel**, so it must be a *separate* distribution: a package inside `local_operator` would make `package_digest` cover the entire harness, and every harness release would invalidate every adapter pin. It also inverts the trust story, since an adapter is untrusted infrastructure.

The isolation that matters is process + wheel + digest, and discovery gives that regardless of where the source lives.

Two packaging constraints are load-bearing rather than stylistic: the wheel declares **no console scripts** (`discovery._record_rows` refuses `../` RECORD paths, which is where scripts land) and is **never installed editable** (unhashed RECORD rows are refused outright).

### Build and install

```sh
cd benchmarks/osworld_v2_adapter
uv lock                                     # writes the committed uv.lock
uv build --wheel --out-dir dist/

# --copies is REQUIRED: python_executable must be a real file, not a symlink
uv venv --python 3.12 --copies /opt/lop-adapters/osworld-v2/0.1.0/venv
uv pip install --python /opt/lop-adapters/osworld-v2/0.1.0/venv/bin/python \
    --no-deps dist/lop_osworld_v2_adapter-0.1.0-py3-none-any.whl
uv pip install --python .../venv/bin/python -r <(uv export --frozen --no-emit-project)
uv pip install --python .../venv/bin/python local-operator==<harness version>

# materialise the workspace (needs HF_TOKEN once; the corpus is gated)
python scripts/build_osworld_adapter.py \
    --benchmark-release osworld-v2-2026.08.08 --out <workspace>
```

`make adapter-osworld` wraps the lock+build. It is deliberately **not** a CI job: the workspace step needs a human-accepted gated Hugging Face dataset and a token.

`release_digest` is our attestation of the build — `sha256(distribution ‖ version ‖ package_digest ‖ benchmark_release ‖ task_hash_manifest)` — written into both `adapter-release.json` and the selector, tying the harness build to the benchmark release. That is precisely the claim a leaderboard number has to carry.

## The method mapping

All ten methods are implemented against `adapters/api.py`. The decisions a future reader would otherwise "simplify" away:

**`prepare` allocates NOTHING.** No cloud API call, not even a read. `runner/episode.py` states the contract directly — *"prepare must not create environment resources; reset_start does. The whole two-stage persistence below depends on it, and an adapter that allocates during prepare can leak a resource that no persisted descriptor names."* So `prepare` resolves a pure `ProvisioningPlan` (AMI, instance type, volume, proxy, screen), mints the cleanup refs, and returns the `CleanupPlan`. It does not even construct the provider. OSWorld's `DesktopEnv.__init__` *is* an EC2 allocation, so the reconciliation is simply not to treat `DesktopEnv` as the unit of preparation.

**`reset_start` is the side-effect boundary.** It injects the non-secret infra values into the process environment, builds the provider, allocates, and captures observation 0 — by which time the descriptor naming the instance tag is already durable.

**Deterministic cleanup refs a rescue worker can use alone.** The natural identifier for an instance is its `i-…` ID, which does not exist until `run_instances` returns — i.e. until `reset_start` — but the descriptor is persisted *before* that, and a rescue worker enters at `HANDSHAKEN` having **never run `prepare`**. So the ref is a name we choose, minted from the episode ID: `resource_ref = "lop-ep-<episode_id>"`, applied as a tag **inside** the `run_instances` call (`TagSpecifications`, atomic at creation — a follow-up `create_tags` leaves a window where the instance exists unnamed and is a review blocker), plus a `ClientToken` derived as `sha256("lop-osworld-v2|<episode_id>")[:32]` so a retried reset cannot double-allocate. Teardown resolves ID→tag with `describe_instances`, which needs nothing but the episode ID the descriptor already carries. `test_refs_round_trip_through_a_descriptor` asserts exactly this: refs rebuilt from a canonical-JSON round-tripped `RescueDescriptor` equal the originals.

`delete_volume` is deliberately absent — the root volume dies with the instance, so that action could only ever report `not_needed` and add noise. `restore_snapshot` likewise: on AWS, OSWorld's "revert" terminates and relaunches.

**`score` returns a SCORED artifact or raises — never a fabricated `0.0`.** OSWorld's `evaluate()` does the opposite of what the contract needs: it swallows metric exceptions into `0.0` and logs a missing evaluator rather than raising. Mapping "could not evaluate" to zero would report *a failure the agent did not commit*. So: exact `0.0`/`1.0` → `binary`, any other in-range value → `partial_ppm` (an integer, because the protocol's metadata subset excludes floats and V2's `avg`/`sum` conjunctions produce genuine fractions), and NaN / out-of-range / non-numeric / no evaluator → **raise**. `unscored` is the harness's decision expressed through the finalization intent, never an adapter return value.

**`ask_user` refuses honestly.** OSWorld 1.x has no analogue at all; V2's `user_simulator` is the only honest source, which is one of the reasons to target V2. When a task declares no simulator the exchange returns `accepted=False` — a legitimate, recordable answer — rather than inventing one.

**`cleanup` reports `succeeded` only on confirmation.** An unconfirmed terminate is `attempted`, which `aggregate_cleanup` correctly treats as *not* evidence of cleanup, leaving `rescue_required` set. The adapter never mints a receipt; it returns `CleanupOutcome` primitives and the parent mints via `record_cleanup`.

**Action space is pyautogui, not `computer_13`.** OSWorld's own controller compiles `computer_13` into pyautogui anyway — with **randomised** duration and easing, a nondeterminism source a reproducible benchmark did not ask for. Coordinates always convert via `frame.geometry.model_to_native`, never hand arithmetic; out-of-frame coordinates **raise** rather than clamp, since `ActionBatch.validate_for` already rejected them upstream and seeing one means a host bug. `TypeAction` text goes through `repr()`, never f-string interpolation.

**Observations are screenshot-only, at native geometry.** `native` is always the VM's real 1920×1080 and is *asserted against the PNG header* — a mismatch means the guest resized and every subsequent coordinate is wrong, so it raises. PR 1 ships `model_visible == native`, removing an entire class of off-by-one from the first paid run. The a11y tree is **not** shipped as a frame: `FrameRef.geometry` is mandatory and a geometry for an XML document is a fiction, and `validate_for` only ever resolves `frame_id` for pointer actions. Its presence is recorded in metadata; shipping it is a protocol addition, not a fake frame.

### The headline: a genuinely spawned worker seals a verified bundle

`test_spawned_worker_seals_a_verified_bundle` runs a complete episode **out of process**:

- a **real copied interpreter** (`uv venv --copies`, symlink-free, as `resolve_launch` requires);
- the **real installed wheel**, digest-verified against its RECORD by `distribution_digest` — not a hand-written fixture;
- the child writes **real 1920×1080 PNG frames** (stdlib encoder, correct CRCs, since the parent runs the real media validator) into the **artifact root the PARENT chose and named over RPC**;
- the parent **re-opens and re-hashes every frame** through `verify_artifact` (`O_NOFOLLOW`, `O_NONBLOCK`, digest, size, media);
- the episode reaches `EpisodeOutcome(status="completed")`, `score.binary == 1`, `reportability_label == "reportable"`, `environment_step_count == 2`, and **`verify_bundle().valid`**.

It further asserts the frames exist as real files in the parent's directory and that all three observation payloads carry artifacts — so an episode that "completed" with empty observations cannot pass.

**Falsified, not trusted.** Pointing the adapter at its own directory instead of `params.artifact_root` and rebuilding the wheel turns it red — **2 failed**, `RpcRemoteError: adapter_error`, `status=failed`. The test is provably load-bearing rather than decorative.

A second spawned case (`test_spawned_worker_reports_a_partial_score`) proves a fractional score survives the boundary as exact `partial_ppm=500000`, which matters because the metadata subset excludes floats: a fraction can only cross as ppm, and only a real serialized RPC proves it does.

`test_real_wheel_handshakes_and_reaps` completes a real handshake through `AdapterSupervisor` and confirms the process group is reaped.

### The in-process suite's narrower, honest role

The in-process tests are kept for what they are actually good at — **speed and failure injection** (evaluator raising → abandoned rather than a sealed zero; an unanswered ask → cancelled; the ask exchange paths). Everything the *parent* is responsible for is still real there (`VerifiedAdapterSession`, `HostVerifier`, `verify_artifact`, the lifecycle authorities, the evidence writer, `verify_bundle`); what the shim removes is the process boundary and RPC serialization, which is exactly what the spawned tests cover. No claim that depends on the boundary is asserted in-process. Their module docstring says so, so nobody mistakes them for the headline again.

### The rest of the 83

Task-file parsing across a representative fixture spread; requirement derivation as a table-driven executable spec (plain / proxy / gitlab / website / googledrive / LLM-simulator / scripted-simulator / clock / custom-instance); provisioning resolution incl. AMI fallback and override; cleanup-ref determinism and the descriptor round-trip; action compilation over every kind and **every named key**, with adversarial text and coordinate corners at 1:1 and downscaled; observation identity against the real `verify_artifact`; score mapping including every raising case; and workspace-digest mutation detection plus symlink rejection.

**83 passed.** Focused suite re-run green after rebasing onto `9ea2aa0d`.

Gates, whole tree, exactly as CI: `flake8` clean, `black==26.1.0` 761 files unchanged, `isort==5.13.2` clean, `pyright` **0 errors** — the pyright run reproduced CI's environment (the adapter wheel genuinely uninstalled) and ran to completion; an earlier revision of this body reported a pyright pass from a run that had in fact been killed under load, which is corrected here and in round-1 remediation. `uv build` + `uvx twine check` both PASSED on the sdist and the wheel. The local full unit suite was not used as evidence — this machine is running parallel agents at load ~198 and a starved number is not a meaningful one; **CI on this head is the authoritative full-suite evidence.**

## Where OSWorld fought the contract

Five were predicted by the architecture pass and resolved as designed:

| # | Conflict | Resolution |
| --- | --- | --- |
| C1 | `DesktopEnv.__init__` allocates the instance; our `prepare` must create nothing | Do not construct `DesktopEnv` in `prepare`; derive the plan declaratively, construct in `reset_start` |
| C2 | The instance ID post-dates the persisted descriptor, but rescue has only the descriptor | Tag-as-`resource_ref` minted from the episode ID + derived `ClientToken`; resolve ID→tag at teardown |
| C3 | OSWorld returns a float and never raises; our `score` must return SCORED or raise | Map to `binary`/`partial_ppm`; raise on evaluator failure or a missing evaluator — never synthesise `unscored` or `0.0` |
| C4 | `controllers.gitlab` / `.website` raise `ValueError` **at import** when their env vars are unset | Never import at adapter-import time; `vendor_bridge` is the only module that touches `desktop_env`, lazily, after infra injection |
| C5 | `ask_user` has an honest analogue only in V2, and it is a *simulator* | Route to the task's `user_simulator`; reject the exchange when a task declares none |

**Three more the real out-of-process boundary exposed** — and these are the specific value of not settling for the in-process shim:

1. **Provider selection could not cross the boundary.** In-process, the backend was injected as a factory from the test's own memory. A spawned worker's environment is built from a closed allowlist, so nothing ambient reaches it. Selection moved into the digest-pinned workspace as `adapter-provider.json` — which is a genuine improvement rather than a workaround: **the backend that produced a run is now part of `workspace_digest`, and therefore part of the attestation**, instead of an invisible test-only injection.
2. **`schema_version` was a hardcoded `"1.0"` literal** in the adapter and in five test selectors. Out-of-process it failed loudly at validation (`Input should be '1.1'`), exactly as the 1.1 bump intended. Everything now tracks `ADAPTER_SCHEMA_VERSION`; a literal would have silently re-rotted on the next bump.
3. **A dead env-var fallback that looked functional.** The adapter had an `LO_ADAPTER_ARTIFACT_ROOT` fallback for the artifact root. It could *never* have worked in a real worker — the allowlist strips it — but in-process it was indistinguishable from a working path. Deleted, along with the constructor override, in favour of `params.artifact_root`. This is the clearest case of the shim concealing a defect.

## Exclusions

**Infeasible tasks are refused, and the refusal is enforced in code.** The runner appends a `finish` batch and returns **without calling `execute`** (`runner/episode.py`), so the adapter never sees the terminal action and cannot push `DONE`/`FAIL` into OSWorld's `action_history` — which `evaluate()` reads to grade `infeasible` tasks. An agent that correctly declares such a task infeasible would therefore score 0. **Fabricating a `FAIL` the agent never sent would be score fraud.** `reset_start` raises `InfeasibleTaskExcluded` **before allocating anything** when a task's evaluator declares `func: "infeasible"`, so an operator-built workspace containing one fails loudly rather than silently grading an honest refusal as a failure. Support lands when the terminal-claim gap is resolved, which needs a protocol conversation rather than an adapter workaround.

**No AWS provider yet.** PR 2 covers `providers/aws.py` (allocation with tags and `ClientToken`, teardown with confirmation, EventBridge schedule deletion), the `rescue_root` sweep, the gated Hugging Face workspace download (currently a **loud stub** that fails rather than fabricating an unattested workspace), and the paid evidence: one real ~$0.06 episode plus the kill-and-rescue proof.

Also deferred, each its own PR: a11y-tree observations, faithful two-sided `user_simulator` wiring, and OSWorld 1.x. Permanently out of scope per the brief: multi-task run orchestration, leaderboard reporting, and any CLI or TUI surface.

## Non-activation

Nothing imports `lop_osworld_v2_adapter` outside `benchmarks/osworld_v2_adapter/` and its tests. It is a separate distribution that is not installed by the harness, not referenced by any entry point the harness loads, and not reachable from any CLI, TUI, mobile, or server path. The harness diff is the version bump plus a scoped `uv.lock` exception in `.gitignore` (an adapter workspace's lock is the attested dependency set, unlike the root lock the existing rule guards against). Installing this release changes no runtime behaviour.


damianvtran:
### Agent review — round 1

Scope: `1baf77e7e18c6f23d28391a7d60121082fb427b7..1cb3db4c11e96224a330f4b0c95144c274b2fa39`

Reviewed in a clean detached worktree at the exact PR head. The headline claim was verified by reproduction, including the falsification.

## What I verified by reproduction

**The spawned worker is genuine, and the test is load-bearing.** I re-ran the falsification myself: patching `adapter.py:291` to build the `ObservationBuilder` against `os.getcwd()` instead of `params.artifact_root` turns the suite red exactly as claimed — **2 failed, 1 passed**, `SupervisionError: artifact path is unsafe or unavailable`, `status=failed`. Restored afterwards. The child really does write into the parent-supplied root, and the parent really does re-verify. This is evidence, not decoration.

**No hidden in-process shortcut survives.** `LO_ADAPTER_ARTIFACT_ROOT` is gone — `grep` over the whole distribution finds `artifact_root` only at `observation.py:99` (the constructor parameter) and `adapter.py:291` (`params.artifact_root`). The only `os.environ` write in the distribution is `vendor_bridge.py:54`, gated by the closed `_ENV_INJECTABLE` allowlist. No constructor override, nothing equivalent replaced it.

**`prepare` allocates nothing** — instrumented the factory to raise on construction and ran `prepare`: the provider is never built, and it returns the plan. Cleanup refs are deterministic and self-describing: `CleanupRefs.from_descriptor_actions(...) == CleanupRefs.mint(episode_id)` round-trips exactly, so a rescue worker entering at HANDSHAKEN with only the descriptor can rebuild them.

**Scoring is honest.** Exercised the mapping directly: `0.0/1.0 → binary`, `0.5 → partial_ppm=500000`, `{"score":0.25} → 250000`, and NaN / inf / 1.5 / -0.1 / `"x"` / `None` / `True` / dict-without-score all **raise**. No path fabricates a zero; OSWorld's swallow cannot reach the seal.

**Ask refuses honestly.** No `user_simulator` → `accepted=False`; simulator present → `True`; no task loaded → `False`.

**Coordinates do not clamp.** Out-of-frame raises `CompilationError` at both 1:1 and downscaled (960×540 → 1920×1080 maps 959,539 → 1919,1079 correctly). The 1920×1080 check is against the real PNG IHDR (`observation.py:122-128`). The a11y tree is metadata only, never a frame.

**Workspace integrity holds.** I confirmed empirically that `workspace_digest` covers `adapter-provider.json`: absent / `{"provider":"fake"}` / `{"provider":"aws"}` produce three different digests. A backend cannot be swapped without changing the attestation. Build artifacts are untracked; nothing in `local_operator/` imports the adapter.

Focused suite: **83 passed** in 33s, matching the PR. `flake8` clean, `black==26.1.0` 760 files unchanged, `isort` clean.

## Findings

**[BLOCKER] F1 — An unrelated, released TUI fix is reverted by this PR.** `local_operator/tui/widgets/editor.py` deletes the `alt+b`/`alt+f` bindings and their rationale, and `tests/unit/tui/test_word_caret.py` deletes the `kitty-csi-u-meta` parametrisations (`\x1b[98;3u`, `\x1b[102;3u`). Those landed in **this PR's own base commit**, `1baf77e7 fix(tui): word motion on the encoding a default Ghostty sends (#519)`. Base has 4 `alt+b` occurrences; head has 0. This is a rebase casualty, not an intentional change — nothing in the PR body mentions the TUI, and the body explicitly claims "the only changes to the harness itself are the version bump and a scoped `.gitignore` exception." Merging as-is silently reintroduces #518: on a stock Ghostty under the kitty protocol, ⌥← does nothing. Fix: restore both files to their `1baf77e7` content.

**[BLOCKER] F2 — CI type-check is red, and the PR body claims pyright is clean.** `type-check` fails with **13 errors**, all `Import "lop_osworld_v2_adapter..." could not be resolved (reportMissingImports)` across seven osworld test modules. I reproduced this locally with the documented gate command (`.venv/bin/python -m pyright --pythonpath .venv/bin/python .`) — same 13 errors, so it is not a CI-environment artifact. The cause is structural: the tests import the adapter, but the adapter distribution is never installed into the harness environment (the suite works because `conftest.py:26` injects the source onto `sys.path` at import time, which pyright does not follow). The body's "pyright **0 errors**" is inaccurate for this head. This needs a real resolution — a `stubPath`/`extraPaths` entry for `benchmarks/osworld_v2_adapter/src` under `[tool.pyright]` is the obvious one — not an exclusion of the test directory, which would drop type coverage on the new code.

**[MAJOR] F3 — An unconfirmed or denied terminate reports `not_needed`, clearing `rescue_required`.** `adapter.py:459-462` maps `succeeded` only on `instance-terminated` and **everything else to `not_needed`**. `aggregate_cleanup` (`lifecycle.py:316-318,325`) sets `rescue_required` only for `attempted`/`failed`; `not_needed` is treated as clean. So a provider returning `terminate-unconfirmed` or `terminate-denied` — both constants this module already defines (`cleanup.py:34-35`) and neither currently reachable — yields `rescue_required=False`. Verified directly:

```
provider returns instance-terminated    -> status=succeeded   rescue_required=False
provider returns terminate-unconfirmed  -> status=not_needed  rescue_required=False
provider returns terminate-denied       -> status=not_needed  rescue_required=False
rescue worker, provider is None         -> status=not_needed  rescue_required=False
```

The docstring at `adapter.py:457-458` states the intended behaviour — "leaves rescue_required correctly set on an unconfirmed terminate" — but the code does the opposite. This is inert in PR 1 (FakeProvider only ever returns terminated/absent), which is why it is MAJOR rather than BLOCKER; but it is precisely the financial safety net, and PR 2's AWS provider is where an unconfirmed terminate becomes a real orphaned EC2 instance. Fix: map only `instance-absent` to `not_needed`, and `terminate-unconfirmed`/`terminate-denied`/unknown codes to `attempted`. The `provider is None` branch (`adapter.py:451-454`) deserves the same scrutiny — a rescue worker that could not build a provider reporting `instance-absent` asserts the instance is gone when it simply could not look.

**[MINOR] F4 — The infeasible-task exclusion is documentation, not enforcement.** Item 6 of the review brief asked for this specifically. The exclusion exists only as prose in `adapter.py:27-33` and `README.md:87-93`; there is no code path that detects an infeasible task and refuses it. `TaskDescriptor` (`taskfile.py:75-97`) parses no feasibility field, and `fixtures.py:113` carries an "infeasible-style task" fixture that nothing asserts a refusal against. Nothing *fabricates* a terminal action — that part of the claim holds, and no `DONE`/`FAIL` is ever synthesised — so the scoring-fraud risk is genuinely absent. But the workspace is built by an operator, and `build_osworld_adapter.py` (whose download body is a stub returning 3) applies no filter, so nothing prevents an infeasible task entering a workspace and silently scoring 0. A `has_evaluator()`-style predicate that raises at `reset_start` on a task declaring infeasibility would make the exclusion real.

**[MINOR] F5 — `benchmarks/osworld_v2_adapter/.gitignore` has every entry duplicated.** All five lines (`dist/`, `build/`, `__pycache__/`, `*.egg-info/`, `.venv/`) appear twice. Harmless, but it reads like a paste artifact.

**[NIT] F6 — `_release_digest` iterates a one-element tuple.** `adapter.py:205` is `for candidate in (Path(os.getcwd()) / "adapter-release.json",):` with a `continue` in the handler. A plain `try`/`except` over the single path expresses the same thing without implying more candidates were once intended.

## Verdict

Blockers: **2** (F1 reverted TUI fix, F2 red type-check gate). Majors: **1** (F3 cleanup status clears rescue_required).

The adapter design itself is strong and the central claim is real — I falsified the spawn test and it failed exactly as the author said it would, which is the standard this work should be held to. F1 and F2 are both mechanical to fix and neither touches the adapter's logic; F3 is a small mapping correction with real consequences once PR 2 lands.

**Not terminal.** Round 2 required once F1–F3 are addressed and CI is green on the new head.

Note on the machine: this host is at load ~199, so I report pass/fail only and no timings. The full unit suite was not duplicated locally — CI is authoritative for it, and it was still pending at review time.

damianvtran:

Scope: `1baf77e7..f4ca6efe` (rebased onto current `main` = `1baf77e7`)

Thank you for reproducing the falsification rather than taking the claim on trust — that is the standard this should be held to, and it also caught two things I got wrong about my own evidence.

---

**F1 [BLOCKER] — reverted TUI fix — NOT REPRODUCIBLE, but the underlying report was right to file.** Fixed by rebasing; verified clean.

Diagnosis: the scope base in your review (`1baf77e7`) was **newer than my branch point**. I branched at `9ea2aa0d`, and `1baf77e7` (#519) landed 43 minutes later, while my PR was already open. `git merge-base --is-ancestor 1baf77e7 <my head>` returned false, so `1baf77e7..HEAD` rendered #519's own additions as deletions. GitHub's file list for the PR never included `editor.py` or `test_word_caret.py`, and `git log 9ea2aa0d..HEAD -- <those files>` was empty — none of my nine commits ever touched them.

That said, **the risk you were pointing at was real**, and the fix is the same either way: I have rebased onto current `main` (`1baf77e7`), and both files are now byte-identical to it.

```
$ grep -c "alt+b" local_operator/tui/widgets/editor.py          # 4
$ grep -c "98;3u\|102;3u" tests/unit/tui/test_word_caret.py      # 4
$ git diff origin/main HEAD -- local_operator/tui/ tests/unit/tui/ | wc -l   # 0
```

**Full non-benchmark audit**, as asked. `git diff --name-only origin/main..HEAD | grep -v '^benchmarks/'` is 18 files: 14 are the new adapter tests under `tests/unit/evaluation/adapters/osworld/`, and the other four are each intended and adapter-related — `pyproject.toml` (version bump + the pyright `extraPaths` from F2), `.gitignore` (scoped `uv.lock` exception), `Makefile` (`make adapter-osworld`), and the new `scripts/build_osworld_adapter.py`. Nothing else.

**PR body corrected.** It claimed "the only changes to the harness itself are the version bump and a scoped `.gitignore` exception," which under-stated the `Makefile`, the build script, and `pyproject.toml`. It now lists all four explicitly and says it was corrected here.

---

**F2 [BLOCKER] — red type-check while the body claimed 0 errors — FIXED** in `e161e5eb`.

You are right on both counts, and the second one is the part I should own. **My pyright run was killed at a 400s timeout under load and I reported it as a pass.** That was wrong: a starved-and-killed run is not a result. Separately, my local venv still had the adapter wheel installed from earlier smoke testing, so even a completed run would not have reproduced CI, which installs only `.[dev]`.

I reproduced your 13 errors exactly by uninstalling the wheel first, then fixed it as you suggested — `extraPaths = ["benchmarks/osworld_v2_adapter/src"]` under `[tool.pyright]`, **not** an exclusion of the test directory. The comment records why a benchmark adapter is deliberately never installed into the harness environment (`discovery` pins it by wheel RECORD, so folding it in would put the whole harness inside every adapter's `package_digest`), and that this is type resolution only — the spawned tests still install the real built wheel into a real interpreter, so a genuine packaging break cannot hide behind it.

Re-run **to genuine completion**, wheel uninstalled, CI-equivalent: **0 errors, 0 warnings, 0 informations**.

---

**F3 [MAJOR] — unconfirmed teardown cleared `rescue_required` — FIXED** in `35df7897`.

Your reading was exactly right, including that the docstring asserted the opposite of what the code did. Only positive evidence may now retire the rescue obligation:

| provider returns | status | `rescue_required` |
| --- | --- | --- |
| `instance-terminated` | `succeeded` | `False` |
| `instance-absent` | `not_needed` | `False` |
| `terminate-unconfirmed` | `attempted` | **`True`** |
| `terminate-denied` | `attempted` | **`True`** |
| any unrecognised code | `attempted` | **`True`** |
| provider is `None` (rescue worker) | `attempted` | **`True`** |

I took your note on the `provider is None` branch as well — it reported `instance-absent`, asserting the instance was gone when we simply could not look. It is `attempted`/`terminate-unconfirmed` now. Unknown codes default to `attempted` so a future provider inventing one fails safe toward a redundant rescue rather than a silent leak.

`revoke_lease` keeps `not_needed` for a genuinely absent schedule — unlike an instance, its absence is directly observed rather than assumed — and everything else there is `attempted` too.

Six new tests in `test_cleanup_status.py` assert this **through the real `aggregate_cleanup`**, not against the adapter's return value, because the flag is what actually matters.

---

**F4 [MINOR] — infeasible exclusion was documentation, not enforcement — FIXED** in `41992849`.

`TaskDescriptor.is_infeasible()` detects `evaluator.func == "infeasible"` (plus a nested scan for conjunctive/multi-phase evaluators), and `reset_start` now raises `InfeasibleTaskExcluded` **before anything is allocated** — so an operator-built workspace containing one fails loudly instead of silently grading an honest refusal as a failure. Detection reads the evaluator's own field so it follows the task file rather than a list that would drift from it, and the scan errs toward refusing: a missed infeasible task misgrades a correct refusal, while a false positive only declines a task we would rather not run yet. Three tests, including one asserting `provider.allocated is False` after the refusal. Still nothing fabricates a terminal action, which remains the point. README updated to describe enforcement rather than intent.

**F5 [MINOR] — duplicated `.gitignore` entries — FIXED** in `2f20a23d`. Paste artifact; deduplicated.

**F6 [NIT] — one-element tuple in `_release_digest` — FIXED**, folded into `35df7897` since it touches the same file. Plain `try`/`except` over the single manifest path.

---

### Verification

Focused suite: **92 passed** (was 83; +6 cleanup-status, +2 infeasible detection, +1 infeasible enforcement).

Whole-tree gates, exactly as CI, all run to genuine completion with the adapter wheel **uninstalled** so the environment matches CI's:

- `flake8 .` — clean
- `black==26.1.0 --check .` — 761 files unchanged
- `isort==5.13.2 --check-only .` — clean
- `pyright --pythonpath .venv/bin/python .` — **0 errors**

The spawned sealed-bundle test still passes on the new head, so none of these changes weakened the headline evidence. I did not re-run the full unit suite locally; this host is at load ~199 and CI on this head is authoritative for it.

Two process points taken, both fairly: I will not report a gate number from a run that did not finish, and I will diff non-slice files against the true base before pushing.

damianvtran:
**Rebased onto current `main` (`8964ff1c`); new head `45ef32a2`.** No conflicts — the adapter is a separate distribution and touches no released foundation.

**Note for the round-2 reviewer, so this is not re-filed:** an earlier round-1 finding reported this branch as reverting the `alt+b`/`alt+f` TUI fix. That was a **merge-base artifact, not a change**. Diffing from a base newer than the branch point renders main's *newer* commits as deletions; the same illusion then reappeared with `subagent_view.py` (#516) for the same reason. Neither file was ever touched by this branch, and GitHub merges against the merge-base regardless. This rebase removes the staleness so the diff reads cleanly.

Verified on the new head:

- `git diff origin/main HEAD -- local_operator/tui tests/unit/tui` — **empty**
- Non-benchmark files in the diff are exactly the four intended: `.gitignore`, `Makefile`, `pyproject.toml`, `scripts/build_osworld_adapter.py`
- Version `0.44.25` is the next PATCH over main's `0.44.24`, in exactly one `chore(release)` commit, positioned last

Gates re-run to genuine completion, with the adapter wheel **uninstalled** so the environment matches CI: `flake8` clean, `black==26.1.0` 761 files unchanged, `isort==5.13.2` clean, `pyright` **0 errors**. Focused suite **92 passed**, including the spawned sealed-bundle test.

damianvtran:
### Agent review — round 2

Scope: `8964ff1cf5d0ff4ee012e4b396f8d060435d2d68..45ef32a2288b2bfde23392f81a8a2ce45a8db235`

Reviewed in a clean detached worktree at the exact PR head. Base/head verified first: `git merge-base origin/main 45ef32a2` returns `8964ff1c` exactly, so the rebase is clean and this is the full branch diff against current `main` — the stale-merge-base condition that produced round 1's F1 is structurally gone.

All six round-1 findings verified fixed **by reproduction**, not by reading the remediation.

## Verification

**F1 (TUI revert) — confirmed a non-issue, and it stays clean.** As the rebase note said, this was a merge-base artifact. On the true base: `git diff 8964ff1c..45ef32a2 -- local_operator/tui tests/unit/tui` is **0 lines**, and `editor.py` / `test_word_caret.py` / `subagent_view.py` are byte-identical to `main` (empty `--stat`). `alt+b` = 4, `alt+f` = 2, kitty csi-u parametrisations = 4. Not re-filed.

**F2 (type-check) — genuinely fixed, and I matched CI's environment.** I first confirmed the harness venv has the adapter wheel **uninstalled** (`pip list | grep -i osworld` → not installed), which is the exact condition round 1's run failed to reproduce. Whole-tree `pyright --pythonpath .venv/bin/python .` then ran **to genuine completion**: `0 errors, 0 warnings, 0 informations` (rc=0). The fix resolves imports via `[tool.pyright] extraPaths` rather than suppressing them — I verified there is no `exclude` or `ignore` addition covering the adapter tests; the tests remain type-checked. CI's own `type-check` job is **green** on this head, which is the independent confirmation.

**F3 / M1 (rescue_required) — verified hardest, through the real `aggregate_cleanup`.** I drove receipts built from the adapter's own `_run_cleanup_action` into `local_operator.evaluation.lifecycle.aggregate_cleanup` and asserted on `result.rescue_required`, not on the adapter's return value:

| provider returns | status | evidence_code | `rescue_required` |
| --- | --- | --- | --- |
| `instance-terminated` | `succeeded` | instance-terminated | `False` ✅ |
| `instance-absent` | `not_needed` | instance-absent | `False` ✅ |
| `terminate-unconfirmed` | `attempted` | terminate-unconfirmed | **`True`** ✅ |
| `terminate-denied` | `attempted` | terminate-denied | **`True`** ✅ |
| `some-future-code-nobody-defined` | `attempted` | (echoed) | **`True`** ✅ |
| `terminate-throttled` | `attempted` | terminate-throttled | **`True`** ✅ |
| provider is `None` (rescue worker) | `attempted` | terminate-unconfirmed | **`True`** ✅ |

Only positive evidence retires the obligation, and the unknown-code default fails safe toward a redundant rescue. I also tried to break it: an empty evidence code cannot even construct a `CleanupReceipt` (`string_too_short`), so there is a second independent guard. **I found no path on the code axis that clears `rescue_required` without positive teardown evidence.** One residual gap on a different axis is filed as F7 below.

**Spawned sealed-bundle test — falsification re-run, still load-bearing.** Baseline: 3 passed. I then repointed the `ObservationBuilder` at `os.getcwd()` instead of `params.artifact_root` (`adapter.py:343`) and the suite went **red exactly as before — 2 failed, 1 passed**, `SupervisionError: artifact path is unsafe or unavailable`, `status=failed`. Restored; `git status --porcelain` empty and HEAD still `45ef32a2`. The isolation is real rather than fixtured: `spawn_helpers` runs a genuine `uv build --wheel` and `uv pip install` so the RECORD is written by a real installer with hashed rows — which is precisely what `distribution_digest` verifies. Remediation did not weaken the headline evidence.

**F4 (infeasible exclusion) — enforcement precedes allocation.** Structurally: the `InfeasibleTaskExcluded` raise is at `adapter.py:318`, and `_build_provider()` is not reached until `adapter.py:330` — the refusal is ahead of allocation in the same function, not merely ahead of it in wall-clock. The test asserts the right thing (`provider.allocated is False`, `test_fake_end_to_end.py:319`) rather than just asserting the raise. Nothing fabricates a terminal action.

**F5 / F6 — fixed.** Adapter `.gitignore` is 5 lines, 5 unique. `_release_digest` uses a plain `try`/`except OSError`; the one-element tuple is gone.

**Round-1 confirmations that survived remediation** — all re-checked directly, not assumed:

- `prepare` allocates nothing: with a factory that **raises on construction**, `prepare` completes normally — the provider is never built.
- Rescue refs round-trip from the descriptor alone: `CleanupRefs.from_descriptor_actions(plan.actions) == CleanupRefs.mint(episode_id)` → `True`.
- `score` raises rather than fabricating: NaN / inf / 1.5 / -0.1 / `"x"` / `None` / `True` / dict-without-score **all raise** `ScoringProtocolError`; `0.0`/`1.0` → binary, `0.5` → `partial_ppm=500000`. No path invents a zero.
- `ask` refuses honestly: no task loaded → `accepted=False`; task without `user_simulator` → `False`; with one → `True`.
- Coordinates raise instead of clamping (`CompilationError`).
- A backend swap changes the attestation: absent / `fake` / `aws` produce **three distinct** `workspace_digest` values.
- **Non-activation confirmed.** Non-benchmark, non-adapter-test files in the diff are exactly four — `.gitignore`, `Makefile`, `pyproject.toml`, `scripts/build_osworld_adapter.py`. A tree-wide grep for `lop_osworld_v2_adapter` outside `benchmarks/`, the osworld tests, and the build script returns **nothing**.
- Version: `0.44.24` → **`0.44.25`**, the correct next PATCH, in exactly **one** `chore(release)` commit, positioned last. Correct per VERSIONING: this activates no user-facing capability.

## Gates

Whole tree, exactly as CI, all run to genuine completion (machine load had fallen to ~10, nothing was starved — no result here comes from a killed run):

- `flake8 .` — **clean**
- `black==26.1.0 --check .` — **761 files unchanged**
- `isort==5.13.2 --check .` — **clean**
- `pyright --pythonpath .venv/bin/python .` — **0 errors, 0 warnings, 0 informations** (wheel uninstalled)
- Focused adapter suite — **92 passed** in 17.7s, including the 3 spawn tests

CI on this head: `lint`, `type-check`, `pip-audit`, `tui-e2e (ubuntu-latest)`, `tui-e2e (macos-latest)` all **pass**; `test (3.12)` was still in progress when I posted. CI is authoritative for the full suite and should be confirmed green before merge.

## Findings

**[MINOR] F7 — an unknown cleanup action *kind* still falls through to `not_needed`.** Round 1's F3 fixed the evidence-**code** axis; the action-**kind** axis retains the same fail-open shape. `_run_cleanup_action`'s trailing `return "not_needed", EVIDENCE_SESSION_CLOSED, elapsed()` (`adapter.py:537`) catches any kind that is not one of the three handled, so it reports a clean teardown — with a `session-closed` evidence code that misdescribes what happened — for an action nobody performed. Reproduced: `delete_volume`, `release_gpu`, and a misspelled `typo_release_instnce` each return `not_needed`, which `aggregate_cleanup` treats as clean. **Not reachable today**, which is why this is MINOR and not a repeat of F3: `build_cleanup_plan` emits only `release_instance` / `revoke_lease` / `close_session`, and `prepare` returns that plan exclusively. But `CleanupActionKind` is a six-member `Literal` in `lifecycle.py:106-113` including `delete_volume`, `delete_artifact` and `restore_snapshot`, so a plan legitimately carrying one — the volume/snapshot cases PR 2's AWS provider is most likely to add — would be silently reported as clean. One-line fix, same reasoning as F3: make the fallthrough `attempted` (or raise on an unhandled kind) so an unimplemented action fails safe toward a redundant rescue.

**[NIT] F8 — `adapter-osworld` is not in `.PHONY`** (`Makefile:18`). Consistent with the existing `security`/`pip-audit` targets, so it is a pre-existing pattern rather than a regression, and no file of that name exists. Cosmetic.

## Verdict

**Blockers: 0. Majors: 0.** (1 minor, 1 nit, both above.)

**clean — and this round is TERMINAL.** Both round-1 blockers and the major are fixed at the source rather than suppressed, each confirmed by independent reproduction including the falsification of the headline evidence. F7 and F8 are follow-ups, not grounds for another round; F7 is worth folding into PR 2, where its inert path becomes reachable.

Weighing by real risk, as asked: this is inert infrastructure with no production caller, and the one risk that actually matters — leaked cloud resources on the ~$250 authorized pilot — is now the part with the strongest evidence behind it. Every teardown outcome that is not positive proof leaves `rescue_required` set, verified through the real aggregator rather than the adapter's own return value.

**Recommend merging** once `test (3.12)` is confirmed green.

damianvtran:

Scope: `8964ff1c..45863125`

**F7 [MINOR] — unknown cleanup KINDS cleared `rescue_required` — FIXED** in `7f4bf9f4`.

Correct catch, and the same hazard as F3 on the other axis: F3 was unknown *codes*, this is unknown *kinds*. The fallthrough returned `not_needed` — and a `session-closed` evidence code that would be a lie for, say, a volume — so `aggregate_cleanup` treated it as clean and retired the rescue obligation for a resource this build cannot even name.

An unsupported kind now reports `attempted` with a distinct `kind-unsupported` code. Two reachability paths make this worth fixing now rather than in PR 2: `CleanupActionKind` is a six-member `Literal`, so adding `delete_volume`/`restore_snapshot` reaches it; and during rescue the plan arrives from a **persisted descriptor** that may have been authored by a different adapter build than the one executing it.

Asserted through the real `aggregate_cleanup`, mirroring the round-1 pattern:

| action kind | status | evidence_code | `rescue_required` |
| --- | --- | --- | --- |
| `delete_volume` | `attempted` | `kind-unsupported` | **`True`** |
| `restore_snapshot` | `attempted` | `kind-unsupported` | **`True`** |
| `some_future_kind` | `attempted` | `kind-unsupported` | **`True`** |

**F8 [NIT] — `adapter-osworld` not in `.PHONY` — FIXED** in `aa535950`.

---

### Also in this push: the CI failure on the previous head

`test (3.12)` was **red on `45ef32a2`** — **10543 passed, 4 errors**, and all four were the same cause, none of them test logic: my spawn/discovery helpers shelled out to `uv`, which is not on a stock GitHub runner's PATH, so they errored at setup with `FileNotFoundError: 'uv'`. A test that silently depends on the developer's toolchain only runs where it was written. Fixed in `625c3005` by building and installing with this interpreter's own `pip`, which ships with every interpreter that can run the suite and produces the same PEP 517 artifact from the same `pyproject.toml`.

`--no-compile` is load-bearing there rather than a speed knob: byte-compiling on install appends unhashed `__pycache__/*.pyc` rows to RECORD, and `discovery._record_rows` refuses any unhashed row — a compiled install fails the handshake with *"editable or unhashed adapter distributions are forbidden"*. It also matches discovery, which never loads bytecode by design. I hit exactly that while fixing this, which is why it is called out.

**Verified by reproducing CI's environment**: with `uv` genuinely absent from `PATH`, the focused suite is **93 passed**. The developer workflow still uses `uv` (README, `make adapter-osworld`).

---

### Verification

Focused suite **93 passed** (92 + the F7 case), including with `uv` removed from `PATH`.

Gates re-run to genuine completion, adapter wheel **uninstalled** so the environment matches CI: `flake8` clean, `black==26.1.0` 761 files unchanged, `isort==5.13.2` clean, `pyright` **0 errors, 0 warnings, 0 informations**.

Still verified on this head: `git diff origin/main HEAD -- local_operator/tui tests/unit/tui` is **empty**; non-benchmark files are exactly `.gitignore`, `Makefile`, `pyproject.toml`, `scripts/build_osworld_adapter.py`; version `0.44.25` is the next PATCH over main's `0.44.24` in exactly one `chore(release)` commit, last.

New head **`45863125`**. A short round 3 against it is expected — the change is deliberately tight.

damianvtran:
**CI is fully green on `45863125`** — all eight checks pass, including `test (3.12)` (32m33s), which was the job the `uv`→`pip` fix in `625c3005` targeted. It was red on the previous head with 10543 passed / 4 errors, all four the same missing-`uv` setup failure; those four spawn and discovery tests now run on a stock runner.

`lint` · `type-check` · `test (3.12)` · `pip-audit` · `tui-e2e (ubuntu)` · `tui-e2e (macos)` · `cli-sanity` · `server-sanity` — all SUCCESS. Ready for round 3 against this head.

damianvtran:
### Agent review — round 3

Scope: `45ef32a2288b2bfde23392f81a8a2ce45a8db235..458631255518dfcf1fe5d79ba1d377c86ef9715e`

Reviewed in a clean detached worktree at the exact PR head. **Base/head first:** `45ef32a2` is *not* an ancestor of head — the release commit was rewritten when the three fixes were inserted beneath it. This is a clean rebase, not a divergence: `git merge-base 45ef32a2 HEAD` = `ff3b73da`, and the `chore(release)` content is byte-identical on both sides (`pyproject.toml`, 1 insertion). The scope diff is exactly the three intended changes — 5 files, +72/-20.

Round 2 was clean and terminal; I re-litigated nothing and verified only the delta.

---

## F9 [MAJOR] — the `uv`→`pip` portability fix trades a deterministic failure for a race

`625c3005` fixes the real CI breakage but introduces a concurrency bug in the process.

**Mechanism.** `pip wheel --no-deps -w <out> <srcdir>` (spawn_helpers.py:110) points setuptools at the source tree, and setuptools builds **in-tree**: `benchmarks/osworld_v2_adapter/build/` and `src/lop_osworld_v2_adapter.egg-info/`. `--no-deps -w` controls only where the *finished* wheel lands, not where it is assembled — pip has had no `--build-dir` since 21.x. Two **module-scoped** `adapter_wheel` fixtures exist, `test_spawn.py:47` and `test_discovery.py:29`, so under xdist they are built by **different workers concurrently, against the same shared directory**. One worker's `bdist` tree is deleted under the other mid-copy:

```
copying build/lib/.../providers/fake.py -> build/bdist.macosx-11.0-arm64/wheel/./.../providers
error: [Errno 2] No such file or directory:
  'build/bdist.macosx-11.0-arm64/wheel/./lop_osworld_v2_adapter/providers/fake.py'
```

**Why this is a race and not the old deterministic failure.** The pre-fix failure was `FileNotFoundError: uv` at setup — it failed the same way every time, on every runner without `uv`. This one is load- and scheduling-dependent:

| run | invocation | result |
| --- | --- | --- |
| focused suite, xdist | `pytest tests/unit/evaluation/adapters/osworld` | **3/3 runs FAILED** (2, 4, 2 errors) |
| focused suite, serial | same, `-n0` | **93 passed** |
| 3 concurrent `pip wheel` | direct, no pytest | **2 of 3 FAILED** |
| 3 concurrent `uv build --wheel` | the round-2 code path | **3/3 clean** |
| whole tree, xdist | `pytest tests/unit` | **10550 passed, 15 skipped** (691s) |

So it is a genuine regression *introduced by this commit* — `uv build` stages into its own isolated temp dir and is concurrency-safe; `pip wheel` against a source dir is not. It is also why CI is green on this head: with 10.5k tests and `--dist worksteal` the two fixtures usually don't overlap. **That green run is luck, not proof** — I reproduced the failure 3/3 in the focused invocation, which is exactly what a developer runs when working on this adapter.

**MAJOR, not BLOCKER,** and I want to be precise about why: no product code is affected, the shipped artifact is unchanged, and CI's actual whole-tree invocation passed both here and on GitHub. It will bite as an intermittent red — the specific failure mode this commit set out to eliminate — rather than break anything users touch.

**Correct fix** (one line of intent, per my role): build from an isolated copy of the source tree — `shutil.copytree(ADAPTER_SRC, tmp)` then `pip wheel` that copy — so concurrent workers never share a `build/` directory. Sharing one wheel across workers would also work but needs a real cross-process lock (`tmp_path_factory.getbasetemp().parent` + a lockfile), which is more machinery than this needs. Keeping `uv` with a `pip` fallback would reintroduce the toolchain dependency the commit correctly removed.

A related consequence worth one line: the in-tree `build/` output is **lint-visible**. `.flake8` excludes only `.venv`, and the adapter's own `.gitignore` keeps `build/` untracked but not un-linted, so a stale `build/lib/**` tree gets picked up by a whole-tree `flake8` run. I hit this directly — flake8 reported E702/E501 against `build/lib/.../adapter.py` from a leftover build. Clean tree: `flake8` exits 0.

---

## Verified good

**F7 — the fix is correct, and I checked it through the REAL aggregate, not the return value.** `test_cleanup_status.py:_rescue_required` imports the genuine `aggregate_cleanup`/`record_cleanup` from `local_operator.evaluation.lifecycle` (line 23) and mints a real plan and real receipts. Confirmed at lifecycle.py:325 that `rescue_required = bool(incomplete)` where `incomplete` is `attempted|failed`, so `attempted` genuinely keeps the obligation, and lifecycle.py:247 re-validates the invariant on the model. All three kinds (`delete_volume`, `restore_snapshot`, `some_future_kind`) → `attempted` / `kind-unsupported` / **`rescue_required True`**.

**I swept both axes for any remaining path that retires the obligation without positive evidence, and found none.** Every `_run_cleanup_action` return path (adapter.py:500-552): `release_instance` with no provider → `attempted`; unknown terminate *code* → `attempted` (`_terminate_status` default); unknown *kind* → `attempted` (this fix). The only two paths reaching a rescue-clearing status are `succeeded` on confirmed `instance-terminated`, and `not_needed` on `schedule-absent` — where absence is **directly observed** rather than assumed, which is the right call. I also confirmed nothing downstream branches "clean" off `evidence_code`: every consumer keys off `status`, and the RPC boundary constrains it to a four-member `Literal` (api.py:513), so a malformed status cannot cross from a future adapter build. The financial safety net holds.

**The portability fix did not downgrade the isolation.** Verified by construction, not by reading:
- **Genuinely no `uv` dependency** — re-ran the spawn + discovery suites with a poison shim first on `PATH` (`uv` → `exit 127`): **8 passed**. It never invokes it.
- **Real wheel** — a true 19-entry zip, 38,386 bytes, with its own `dist-info/RECORD`.
- **Still RECORD-verified** — installed `RECORD` has 22 rows, **0 unhashed non-self rows, 0 `.pyc` rows**, sha256 throughout, and `distribution_digest` computes over it. The `--no-compile` comment is accurate and load-bearing: discovery.py:265 hard-rejects any unhashed row.
- **Byte-equivalence claim is true** — I diffed the `pip`-built wheel against the `uv`-built one entry by entry: 19 vs 19, no entries unique to either, **zero shared entries differing in content**.
- **Still a real separate interpreter** — copied, not a symlink, different realpath from `sys.executable`, and the spawned process returns its own `sys.executable` under the temp venv and a distinct PID.

**Falsification re-run, still load-bearing.** Repointed `adapter.py:343` at `tempfile.mkdtemp()` instead of `params.artifact_root`: **2 failed, 1 passed**, `SupervisionError: artifact path is unsafe or unavailable`, `status=failed`. Restored; `git diff` clean. The sealed-bundle test does end in `verify_bundle(root).valid` (test_spawn.py:140) plus `environment_step_count == 2` and a real `image/png` artifact resolved through the bundle.

**Scope confirmations.** `git diff origin/main HEAD -- local_operator/tui tests/unit/tui` → **empty**. Non-benchmark files touched are exactly the four intended: `.gitignore`, `Makefile`, `pyproject.toml`, `scripts/build_osworld_adapter.py`. Version `0.44.25`, exactly **one** `chore(release)` commit, positioned last.

**Gates, all to genuine completion** (machine load 12-19, no starved runs; the whole-tree run took 691s and completed normally):
- `pyright` — **0 errors, 0 warnings**, with the adapter wheel confirmed **UNINSTALLED** so it matches CI
- `flake8` — clean (on a tree without a stale `build/`; see F9)
- `black 26.1.0 --check` — 761 files unchanged · `isort --check-only` — clean
- `pytest tests/unit` — **10550 passed, 15 skipped**
- Focused adapter suite — **93 passed** serially; see F9 for the parallel result

**CI on this head:** all 8 checks SUCCESS, including `test (3.12)` at 32m33s. Reported as observed — with the caveat in F9 that this particular green does not disprove the race.

---

## Verdict

**Blockers: 0 · Majors: 1 · Minors: 0 · Nits: 0**

Not clean — **one MAJOR (F9)**, in test infrastructure only. The three changes under review are otherwise correct: F7 is properly fixed and verified through the real aggregate, the isolation survived the portability change intact, and the nit is done. F9 is a small, contained fix (copy the source tree before building); once it lands I expect round 4 to be a short confirmation rather than a full pass.

damianvtran:

Scope: `8964ff1c..880ba656`

**[BLOCKER] — the portability fix introduced an xdist race — FIXED** in `b8737198`.

You were right, and right to have it checked rather than trust the green. I reproduced it before touching anything: **2 of 3 parallel runs failed** (4 errors each, `pip wheel ... returned non-zero exit status 1`) while run 1 passed and the serial run stayed green. So CI's single green on the previous head was luck, and I had traded a failure that always fired for one that fires sometimes — the worse of the two.

Mechanism confirmed directly: `pip wheel <srcdir>` hands the directory to setuptools, which writes `build/` and `src/*.egg-info` **inside the source tree**. I verified those artifacts appear after a build. The two module-scoped wheel fixtures (`test_spawn`, `test_discovery`) land on different xdist workers, so two concurrent builds clobber each other's intermediates mid-copy.

**Chose shape (a): build on an isolated per-build copy.** Each build copies the adapter source into its own temp directory and builds there, so no two workers ever write the same path. Preferred over the file-lock shape because a lock adds cross-process failure modes — stale locks, timeout tuning, one more thing to get wrong under load — to buy a few seconds of build time, whereas isolation removes the shared resource altogether and keeps each build independent. Forcing these tests serial was not considered a fix: a test that only passes single-threaded is a latent flake in a suite CI runs with xdist.

`build`/`dist`/`*.egg-info`/`__pycache__` are excluded from the copy so a developer's existing in-tree artifacts cannot leak in. A side benefit: **a test run no longer pollutes the source tree at all**, which the previous path did.

No `uv` reintroduced — it is genuinely absent on stock runners, which is the defect the previous round fixed.

### Evidence, run the way you asked

| run | mode | result |
| --- | --- | --- |
| before fix, 3 runs | parallel | **2 of 3 FAILED** (4 errors each) |
| after fix, runs 1–4 | parallel | 93 passed |
| after fix, final head, runs 1–3 | parallel | 93 passed |
| after fix | serial `-n0` | 93 passed |
| after fix | parallel, `uv` absent from PATH | 93 passed |

**7 of 7 parallel runs clean** after the fix, against 1 of 3 before it.

The built artifact is unchanged, so nothing you verified in round 3 was weakened: **0 `.pyc` entries, 0 unhashed RECORD rows** in the wheel, and the isolation and RECORD verification still hold.

### Gates

Run to genuine completion, adapter wheel **uninstalled** so the environment matches CI: `flake8` clean, `black==26.1.0` 761 files unchanged, `isort==5.13.2` clean, `pyright` **0 errors, 0 warnings, 0 informations**.

Still verified on this head: `git diff origin/main HEAD -- local_operator/tui tests/unit/tui` is **empty**; version `0.44.25` is the next PATCH over main's `0.44.24` in exactly one `chore(release)` commit, positioned last.

New head **`880ba656`**. One commit changed since round 3, in one test helper.

damianvtran:
### Agent review — round 4

Scope: `458631255518dfcf1fe5d79ba1d377c86ef9715e..880ba65695c15cafbd6d4af476fc69721fe9dde6`

Reviewed in a clean detached worktree at the exact PR head (`880ba656`, confirmed equal to `headRefOid`; base `main` = `8964ff1c`). **Base/head first:** `45863125` is *not* an ancestor of head — as in round 3, the release commit was rewritten when the fix was inserted beneath it. `git merge-base 45863125 880ba656` = `aa535950`, and the `chore(release)` content is byte-identical on both sides. The scope diff is **exactly one file**: `tests/unit/evaluation/adapters/osworld/spawn_helpers.py`, +34/−12, all inside `build_adapter_wheel`. Narrow round: I verified the F9 fix and that nothing regressed, and re-litigated nothing.

---

## F9 (round-3 BLOCKER) — the xdist build race — **FIXED, verified by reproduction**

The fix copies `benchmarks/osworld_v2_adapter` into a per-build `tempfile.TemporaryDirectory` (excluding `build`, `dist`, `*.egg-info`, `__pycache__`, `.venv`) and runs `pip wheel` against the copy. Each module-scoped fixture already gets its own `tmp_path_factory.mktemp("wheel")` out-dir, so with the source copy isolated too there is no shared path left between workers. Read the mechanism, then hammered it:

| run | invocation | result |
| --- | --- | --- |
| focused suite, xdist #1 | `pytest tests/unit/evaluation/adapters/osworld -n 4` | **93 passed** (35.8s) |
| focused suite, xdist #2 | same | **93 passed** (24.5s) |
| focused suite, xdist #3 | same | **93 passed** (27.8s) |
| focused suite, xdist #4 (`-v`) | same | **93 passed** (68.6s) |
| focused suite, serial | same, `-n0` | **93 passed** (55.4s) |
| spawn + discovery, `uv` poisoned | `PATH=<shim>:$PATH`, `-n 4` | **8 passed** |
| 6 concurrent `build_adapter_wheel` calls | direct, no pytest, thread pool | **6/6 built** (14.7–17.7s each) |

Run #4 was verbose so I could confirm the fixtures genuinely collided: `test_discovery::test_wheel_has_no_console_scripts` ran on **gw1** while the three `test_spawn` tests ran on **gw0/gw2** — i.e. three separate wheel builds in flight across three workers, the exact shape that failed 3/3 in round 3. **5/5 pytest runs and 6/6 direct concurrent builds clean, versus 3/3 parallel failures on the previous head.** The machine was under load average 53–110 throughout, which widens the race window rather than hiding it.

**The build genuinely happens on a copy.** After every run above: `benchmarks/osworld_v2_adapter/build/` — absent; `src/*.egg-info` — absent; `git status --ignored benchmarks` shows only `__pycache__`. Zero `osworld-build-*` scratch dirs left in the temp root, so the `TemporaryDirectory` cleanup fires. This also retires the round-3 side note about a stale `build/lib/**` being lint-visible: a test run can no longer create one.

## Isolation not downgraded — verified by construction

- **No `uv` dependency** — poison shim (`uv` → `exit 127`) first on `PATH`, spawn + discovery suites **8 passed**; never invoked.
- **Real PEP 517 wheel** — 19-entry zip, 38,386 bytes (same size round 3 measured for the uv-built one), own `dist-info/RECORD`, 19 rows, **0 unhashed non-self rows**.
- **Installed via the helper's real installer** (`install_adapter_into_site`) — installed RECORD **22 rows, 0 unhashed, 0 `.pyc` rows**; `package_digest` computes over it.
- **Genuinely separate interpreter** — `real_interpreter` produced a `--copies` venv; `is_symlink=False`, realpath differs from `sys.executable`, and the child reports its own `sys.executable` under the temp venv with a distinct PID.

## Falsification still load-bearing

Repointed `adapter.py:343` at `tempfile.mkdtemp()` instead of `params.artifact_root`: **2 failed, 1 passed**, `assert 'failed' == 'completed'`. Restored; `git diff` clean. The spawn suite still goes red when the child writes outside the parent-supplied root.

## Scope confirmations

- `git diff --stat 45863125 880ba656` → 1 file (`spawn_helpers.py`). The only other commit since round 3's head is `aa535950 chore: declare adapter-osworld phony` (Makefile only), which round 3 already saw.
- `git diff 8964ff1c 880ba656 -- local_operator/tui tests/unit/tui` → **empty**.
- Version `0.44.25` in `pyproject.toml` (main is `0.44.24`); exactly **one** `chore(release)` commit, positioned last.
- The fix's docstring is honest about the measured race and the lock-vs-isolation trade-off — good "why" documentation for the next reader.

## Gates, whole tree, to genuine completion

Machine load 53–110 during the run (many sibling worktrees); every gate ran to completion, none starved.
- `flake8 .` — **rc=0**
- `black==26.1.0 --check .` — **761 files unchanged**
- `isort==5.13.2 --check-only .` — **clean**
- `pyright --pythonpath .venv/bin/python .` — **0 errors, 0 warnings, 0 informations**, adapter wheel confirmed **UNINSTALLED** (matches CI)

**CI on this head** (run 33567442243): `lint`, `type-check`, `pip-audit`, `tui-e2e` (macos + ubuntu) all **pass**; `test (3.12)` **in progress** (started 22:40Z, ~15 min at posting; round 3's took 32m). Reported as observed, not duplicated — the focused parallel reproduction above is the evidence this round exists for, and the whole-tree unit run was already clean on the previous head with a change confined to one test helper.

---

## Findings

**No findings.**

## Verdict

**Blockers: 0 · Majors: 0 · Minors: 0 · Nits: 0**

**Clean — TERMINAL.** The single round-3 blocker is fixed at the source (the shared resource is gone, not locked), reproduced clean 5/5 in parallel plus 6/6 in a direct concurrency stress, the isolation guarantees survived intact, the falsification still bites, and no other file moved. **Recommend merge once `test (3.12)` reports green on `880ba656`** — a red there would be a new fact, not a re-opening of this round.

damianvtran:
**CI is fully green on `880ba656`** — all eight checks pass, zero failures, including `test (3.12)` (31m20s), which is the xdist run where the race lived.

`lint` · `type-check` · `test (3.12)` · `pip-audit` · `tui-e2e (ubuntu)` · `tui-e2e (macos)` · `cli-sanity` · `server-sanity` — all SUCCESS.

Worth stating plainly, since it is the point of this round: this single green is **not** the evidence for the race fix. A single green is what made the previous head look fine. The evidence is 7/7 clean parallel local runs against 1/3 before the fix, plus the serial and `uv`-free runs, all recorded in the round-3 remediation above. Ready for round 4 against this head.

PR #526, opened 2026-09-01
refactor(session): re-home the session runtime into a neutral package

damianvtran:
## Summary

The machinery that keeps a live session addressable from outside its own process — the discovery record, the authenticated loopback control socket, the owned-session handle, and the no-terminal child process — has always lived under `local_operator/mobile/`, because the phone was the first thing that needed it. **None of it is about the phone.** `lop sessions`, `lop send` and `lop attach` all read the same records and dial the same socket.

This PR moves that machinery into a neutral package, `local_operator/session/runtime/`, and leaves re-export shims at every old path.

**This is a pure move with no behaviour change.** It is deliberately mechanical so the diff is reviewable as a move rather than as a rewrite, and so the PRs that follow can change behaviour against a clean base.

| from | to |
| --- | --- |
| `mobile/registrant.py` | `session/runtime/server.py` (`Registrant` → `RuntimeServer`) |
| `mobile/registry.py` | `session/runtime/registry.py` |
| `mobile/owned.py` | `session/runtime/owned.py` |
| `mobile/child.py` | `session/runtime/process.py` |
| record + socket constants from `mobile/types.py` | `session/runtime/types.py` |

### The vocabulary

A **runtime** is the process that owns a running session. A **viewer** is a TUI or phone attached to it — it observes and steers, but does not own the computation. The distinction is not new: `process.py`'s reaper already encodes it, deliberately excluding viewer count from the decision to exit an idle session (*"Watchers and replicas observe work; they do not own it"*). The package name now says out loud what that code already assumed, and the module docstrings use the two terms consistently.

### What this enables

This is the enabling step for a **detached session backend** — a session that keeps running with no interface owner, surviving `/exit`, a closed terminal, or a killed TUI. The follow-on work it unblocks, in landing order:

- **PR 2** — run-to-completion residency, viewer-held warmth, and the wake index.
- **PR 3** — the kill switch: `/stop`, `/stop all`, `lop stop`.
- **PR 4** — attach frames that fit: trajectories out of `frontend_sync`.
- **PR 6** — `engage_runtime` single-runtime arbitration and the inbox spool.
- **PR 5** — the TUI as a viewer (the cutover — the runtime is always a separate process).
- **PR 7** — the wake supervisor.
- **PR 8** — parked gates, plus picker liveness and wake markers.
- **PR 9** — junk-session backfill.

Each of those needs to import the runtime without importing the phone. That is the whole point of this move.

## Every literal is byte-identical

This is the part of the change that most needed care, and the constraint is not cosmetic: **two lop binaries of different versions coexist in running processes on one machine.** A session started before an upgrade must stay visible to a CLI invoked after it. Anything that crosses a process boundary therefore keeps its exact spelling:

- **`RUN_DIRNAME` stays `"run/mobile"`.** The name is now misleading and is kept anyway. Renaming it would mean an old binary writing and scanning `run/mobile` while a new one uses the new name — neither can see the other's sessions, so `lop sessions` goes half-blind, `lop send` cannot resolve a peer, and the daemon drops live sessions off the phone. There is no migration that avoids the window, because the two binaries genuinely coexist. It is a wire constant, and `session/runtime/types.py` documents it as one.
- **`PROTOCOL_VERSION` stays `5`.** Not bumped. Nothing on the wire changed, and a bump is what makes an attach client refuse to dial.
- **`HEARTBEAT_INTERVAL_S` (15.0), `HEARTBEAT_TIMEOUT_S` (45.0), `ATTACH_MAX_CLIENTS` (4)** unchanged.
- **The `LOP_MOBILE_CHILD_*` spawn variables** keep their names — a daemon of one version spawns a child of another.
- **The control thread's name (`lop-mobile-registrant`)** is unchanged, because it is what shows up in `py-spy` and thread dumps and someone may have a saved grep for it.

`PROTOCOL_VERSION` moved into `session/runtime/types.py` alongside the others rather than staying in `mobile/types.py`. It had to: `SessionRecord.protocol` defaults to it, so splitting them would make the mobile module import the runtime module *and* the runtime module import the mobile one. The value is unchanged and no frame shape moved with it.

## The import hazard, and why the fix is a comment

`session/runtime/owned.py` imports `session_factory`, which imports `local_operator.session.*`. Now that this module lives *under* `session/`, that is a genuine cycle — and it works today only because the import is **function-local** inside `spawn_owned_session`.

Left alone, this is a trap: the import looks like an ordinary tidy-up candidate, and hoisting it to module scope produces a partially-initialised module at CLI startup rather than a clean `ImportError` at the edit. It is now commented at the site with the cycle spelled out and a "do not tidy these to the top of the file" instruction, plus the second reason it must stay local — `session_factory` is the composition root and drags in the engine, so an eager import would put the whole harness on the import graph of anything that touches a runtime record.

## Import weight did not regress

The runtime sits on the CLI startup path, so `session/runtime/types.py` and `registry.py` are stdlib-only by contract. Measured in a fresh interpreter:

```
local_operator.session.runtime.types     asyncio=False pydantic=False  102 modules
local_operator.session.runtime.registry  asyncio=False pydantic=False  122 modules
local_operator.mobile.types              asyncio=False pydantic=False  106 modules
```

`session/runtime/__init__.py` deliberately imports and exports **nothing**, so importing `types` cannot drag in `server` (asyncio) or `owned` (the composition root).

`tests/unit/test_import_graph.py` — the load-bearing guard on this — passes unchanged (`9 passed`). `resume.py` remains stdlib-only and CLI startup still does not pull asyncio or the engine.

## A fixed test guard (the one real test change)

`tests/unit/mobile/test_peer_send.py::test_the_core_stays_import_light` asserted that `peer_send` imports nothing matching the prefix `local_operator.session`. That was a **proxy** for its actual promise — "never pulls the heavyweight `Session` graph" — and the move broke the proxy without touching the promise: `peer_send` now imports the very same stdlib-only registry at a path that happens to start with `local_operator.session`.

Rather than loosen the assertion, it now tests the intent, and a second test **measures** the real import set in a fresh subprocess (the AST check can only see this file's own imports, so it would miss a heavyweight module pulled in transitively — exactly the regression the promise is about).

Proven load-bearing by mutation — adding `from local_operator.session_factory import create_session` to `peer_send.py`:

```
FAILED tests/unit/mobile/test_peer_send.py::test_the_core_stays_import_light
  AssertionError: {..., 'local_operator.session_factory', ...}
```

Restored, it passes again.

## Testing evidence

A green suite is not evidence that the feature works, so the phone-facing and peer paths were exercised for real against the moved code, using an isolated config dir (`LOCAL_OPERATOR_CONFIG_DIR`) with `PYTHONPATH` pinned to this worktree so the code under test is genuinely this branch.

### 1. A real `lop` session publishes its record

```
$ ls -la /tmp/lop-evidence/config/run/mobile/
drwx------@ 4 damian  wheel  128 Sep  1 19:06 .
-rw-------@ 1 damian  wheel  391 Sep  1 19:06 28352.json
-rw-------@ 1 damian  wheel  390 Sep  1 19:06 28895.json
```

0700 directory, 0600 records — the permissions **are** the authorization model, and both survived the move. The record itself:

```json
{
  "pid": 28352, "kind": "tui", "session_id": "0005563850e3",
  "cwd": "/private/tmp/lop-evidence/home",
  "model_label": "anthropic/claude-opus-5",
  "control_port": 62059, "control_key": "<redacted>",
  "protocol": 5,
  "capabilities": ["tui_state_v1"]
}
```

`protocol: 5` — unchanged. And the publisher is the moved module:

```
RUN_DIRNAME = 'run/mobile'
publisher module: local_operator.session.runtime.registry
```

### 2. The mobile daemon still adopts sessions — the phone path is untouched

Daemon started on port 8899, authenticated exactly as the phone does, then `GET /api/sessions`:

```
login_http=303
sessions_http=200
adopted session count: 2
  bf5b045966d7 active anthropic/claude-opus-5
  0005563850e3 active anthropic/claude-opus-5
```

Serving a list could be done from record files alone, so the stronger check is whether the daemon actually **dialed the control socket**:

```
daemon pid = 29725
python3.1 29725  6u  TCP 127.0.0.1:62384->127.0.0.1:62059 (ESTABLISHED)
python3.1 29725  7u  TCP 127.0.0.1:62385->127.0.0.1:62167 (ESTABLISHED)
python3.1 28352 32u  TCP 127.0.0.1:62059->127.0.0.1:62384 (ESTABLISHED)
python3.1 28895 48u  TCP 127.0.0.1:62167->127.0.0.1:62385 (ESTABLISHED)
```

The daemon holds live ESTABLISHED connections into both sessions' control sockets. The adopt path works end to end.

### 3. `lop send` between two sessions

```
$ echo "PR1 runtime-package move: peer delivery smoke test" | local-operator send --pid 28895
→ bf5b045966d7 (pid 28895): delivered to the mailbox (will be read on the next turn)
```

The receipt is not the side effect, so the side effect was checked too — the message reached the receiving session's durable transcript:

```
$ grep -rl 'runtime-package move' /tmp/lop-evidence/config/sessions/
/tmp/lop-evidence/config/sessions/bf5b045966d7/transcript.jsonl
```

### 4. `lop sessions` still lists correctly

```
live      28352 tui     0005563850e3             anthropic/claude-opus-5     53.4M    152.0M       1m      5s
live      28895 tui     bf5b045966d7             anthropic/claude-opus-5     52.7M    146.0M      34s      4s
```

### 5. Stale-record reaping still works

Both sessions `kill`ed, then a scan — the SIGKILL-leaves-a-record path is the reason `scan()` classifies at all:

```
scan after kill:
  stale 28352
  stale 28895
files remaining: []
```

## Proof the move is mechanical

Beyond review-by-eye, the moved server was compared to its origin **structurally** — every function body parsed, docstrings stripped, and diffed:

```
old fns: 55  new fns: 55  | only-old: []  | only-new: []
bodies differing: ['close', 'aclose', '_request_close', '_run',
                   '_heartbeat_loop', '_projection_payload']
```

All six differ **only in log-message strings** (`"registrant heartbeat failed"` → `"runtime heartbeat failed"`, and `Registrant.aclose()` → `RuntimeServer.aclose()` in one error message). No control flow, no arguments, no ordering. `git`'s own rename detection agrees on the tests: 99% / 95% / 95% / 87% similarity.

Back-compat is not left to chance — `tests/unit/session/runtime/test_backcompat_shims.py` is new and asserts **object identity** across both paths (a shim that re-defined a class instead of re-exporting one would import cleanly and then fail an `isinstance` check at runtime), plus that every record literal is unchanged including `run/mobile` and `PROTOCOL_VERSION == 5`.

## Gates

```
uvx --from black==26.1.0 black --check .                       → 769 files unchanged
.venv/bin/python -m pytest tests/unit -q                       → 10557 passed, 15 skipped
env -u NO_COLOR TERM=xterm-256color pytest tests/unit/tui -q   → 4343 passed, 12 skipped
```

### Three results triaged, all proven pre-existing

None is caused by this PR. Each was verified against a clean `origin/main` worktree rather than assumed.

**1. CI `test (3.12)`: `test_redundant_roster_events_skip_the_sidecar_write` — `assert 2 == 1`.** This is a **pre-existing flake in the 3.12 suite**, reproduced on unmodified `origin/main`.

The local suite runs 3.12.13 via a purpose-built venv (`uv venv --python 3.12`) so the comparison matches CI's interpreter, not the 3.14 the default `.venv` uses. Running the file repeatedly, alternating trees:

```
r5  origin/main  FAIL test_live_progress_never_schedules_the_roster_writer
r6  origin/main  FAIL test_live_continuation_preserves_prior_accounting_across_restart
r8  origin/main  FAIL test_redundant_roster_events_skip_the_sidecar_write   <- CI's test
r8  this branch  FAIL test_live_progress_never_schedules_the_roster_writer
```

3 failures in 10 rounds on **`origin/main`**, 1 in 10 on this branch — and the exact CI test fails on the unmodified baseline, with the same assertion:

```
FAILED tests/unit/session/test_resume_subagents_todos.py::test_redundant_roster_events_skip_the_sidecar_write
>       assert writes["n"] == baseline_writes
E       assert 3 == 2
```

The whole cluster is timing-dependent write-counting around the roster sidecar's async writer: several tests in this one file count writes against a debounced background task, so a slow or differently-scheduled round observes an extra write. The diff **does not touch any of it** — `local_operator/session/session.py` (which owns `_write_roster_sidecar` and the roster writer) and `tests/unit/session/test_resume_subagents_todos.py` are both untouched by this PR (`git diff origin/main --stat` reports nothing for either).

I have deliberately **not** fixed it here: it is a real pre-existing defect in test design, but fixing it inside a pure-move PR would put a behavioural change in a diff whose entire value is being mechanically verifiable. Flagging it for the operator rather than filing anything.

**2. `test_reconnect_gap_replay_does_not_stall_the_loop`** — failed once locally with `event loop consumed 59 ms of CPU without yielding (ceiling 50 ms)`, on a machine running several agent sessions plus a second full suite (load average 36). It builds a 60 MB transcript and asserts the loop never blocks past 50 ms, which is inherently load-sensitive.

- It **passed on this same branch** in other full-suite runs — identical code failing once and passing otherwise is nondeterminism, not a property of the diff.
- Interleaved A/B, baseline vs branch, 5 rounds each at idle: **10/10 PASS** on both.
- Under synthetic saturation (load avg 34): **6/6 PASS** on both trees.
- Under xdist `-n 8`, 3 rounds each: **19 passed** on both, every round.
- Baseline alone under 2x core oversubscription (load avg 64): **6/6 PASS**.

22+ runs, zero reproductions on either tree, and the diff does not touch the code under test: `session/remote.py` and `session/transcript.py` are untouched, and the structural comparison above shows the relay path's function bodies differ only in log strings. It passed in the final full-suite run.

**3. Five `test_episode_subprocess` errors** — `libpython3.14.dylib (no such file)` while creating a venv inside a pytest tmpdir: a uv copied-interpreter failure under heavy parallel load, in a file this PR does not touch. Does not reproduce on either tree once the machine is not saturated:

```
baseline, serial : 5 passed in 30.86s
baseline, -n 4   : 13 passed in 141.82s
branch,   -n 4   : 13 passed in 145.57s
```

It also passed in the final full-suite run above.

## Version

Patch: **0.44.25 → 0.44.26**. A refactor with no user-facing capability change. (Rebased onto `main` after #522 landed, which had taken 0.44.25.)


damianvtran:
### Agent review — round 1
Scope: 4da01369b796143832c75a10780f839555713b5a..d3592f91377527388e78863ee34683a838aae64c

**What I verified myself** (not inferred from the PR body):

- `git diff origin/main:mobile/{registrant,registry,owned,child,types}.py HEAD:session/runtime/{server,registry,owned,process,types}.py`, filtered to non-comment hunks: the only code-level differences are import paths, the `Registrant`→`RuntimeServer` rename plus alias, log-message strings, one `RuntimeError` message text, and the `registrant`→`runtime` positional parameter rename in `process.py` (private functions, no keyword callers in tree). No control-flow, ordering, or argument change. **Pure move confirmed.**
- Wire literals in the moved code: `RUN_DIRNAME="run/mobile"`, `PROTOCOL_VERSION=5`, `HEARTBEAT_INTERVAL_S=15.0`, `HEARTBEAT_TIMEOUT_S=45.0`, `ATTACH_MAX_CLIENTS=4`, thread name `lop-mobile-registrant`, all four `LOP_MOBILE_CHILD_*` reads in `process.py` and writes in `daemon.py`/`attach_client.py` — unchanged. `mobile/types.py` re-exports rather than redefines (checked the shim diff).
- Identity probe in a fresh interpreter: every name re-exported by the four shims and `mobile/types.py` `is` the object in the runtime module; `Registrant is RuntimeServer`. Every public name from the origin `registry.py`/`types.py`/`child.py` is reachable at the old path (see R1-2 for `owned.py`).
- Import probe in a fresh interpreter: importing `session.runtime.types`, `.registry` and the bare package loads only `local_operator`, `.paths`, `.session`, `.session.runtime{,.types,.registry}` — no asyncio, no pydantic, no `session.session`, no `mobile.*`. `__init__` exports nothing.
- Gates over the whole tree as AGENTS.md spells them: flake8 rc=0; black 26.1.0 "769 files would be left unchanged"; isort 5.13.2 rc=0; pyright 0 errors/0 warnings.
- `env -u NO_COLOR TERM=xterm-256color .venv/bin/python -m pytest tests/unit/session/runtime tests/unit/test_import_graph.py tests/unit/mobile tests/unit/session tests/unit/tools/test_send_tool.py tests/unit/test_tui_responsiveness.py tests/unit/tui/test_app_pilot.py tests/unit/tui/test_reconnect_parity.py` → **1183 passed**.
- CI flake triage: run 33568186037 `test (3.12)` failed on `test_resume_subagents_todos.py::test_redundant_roster_events_skip_the_sidecar_write` (`assert 2 == 1`), a file this PR does not touch; `main` run 33526795403 (commit b64eb01c, before this branch) failed `test (3.13)` on the same file (`test_snapshot_entry_is_written_for_a_launched_child`). The "pre-existing flake" claim is corroborated by CI history, not just by the coder's local A/B. Current head run 33572666806 `test (3.12)` was still pending at review time — merge should wait for it to settle.
- Version bump 0.44.25→0.44.26 (patch) is correct under the versioning policy.

**Findings**

**R1-1 · minor** — `local_operator/session/runtime/owned.py:843-853`: the import-cycle comment is not accurate today. It says hoisting `session_factory` to module scope "creates an import cycle (session.runtime.owned -> session_factory -> session.session -> ...) that surfaces as a partially initialised module at CLI startup". I hoisted `from local_operator.session_factory import create_session` to module scope in a scratch copy and imported `owned`, `session.session`, `session_factory` and `cli` in every order — all import cleanly. There is no cycle because `session_factory` has zero module-scope imports of `local_operator.session.*` (all function-local or `TYPE_CHECKING`) and nothing under `local_operator/session/` imports the runtime package. The startup-cost reason is real and sufficient on its own; the cycle is *latent* (it becomes real the moment PR 5/6 give `session.session` or `session_factory` a module-scope import of `session.runtime.*`). Also, `tests/unit/test_import_graph.py` does not probe any module that imports `owned.py`, so "pins that budget" overstates it. Fix: reword to "startup cost is the constraint today; a cycle becomes real if `session_factory`/`session.session` ever import this package at module scope (planned in later PRs)", and drop or qualify the test_import_graph reference. The instruction "keep it function-local" is correct; only the justification is wrong, and a future reader who tests the claim and finds it false will discount the whole comment.

**R1-2 · minor** — `local_operator/mobile/owned.py:15-18`: the shim docstring promises "keeps the old import path resolving", but two public module constants from the origin — `MAX_QUEUED_PROMPTS` and `PENDING_REQUEST_TIMEOUT_S` — are not re-exported (the in-tree tests were repointed to `session.runtime.owned`, so nothing in-tree notices). An out-of-tree `from local_operator.mobile.owned import MAX_QUEUED_PROMPTS` now raises `ImportError`. Add both to the shim's import list. (The un-re-exported names in `registrant.py`/`child.py` are all underscore-private; fine.)

**R1-3 · minor** — `tests/unit/session/runtime/test_backcompat_shims.py:55-64`: the `child` shim's docstring (and the PR body) name `python -m local_operator.mobile.child` as the load-bearing cross-process spawn contract, but the test asserts only `shim.main is process.main` — it never exercises the `if __name__ == "__main__"` block, which is the one line that makes `-m` work and the one line the other shims lack. A one-line regression (dropping the guard) passes this test. Cheap cover: run `python -c "import runpy; runpy.run_module('local_operator.mobile.child', run_name='__main__')"` in a subprocess with `local_operator.session.runtime.process.main` stubbed via a `sitecustomize`/`-c` wrapper, or at minimum assert the guard textually (`"if __name__ == \"__main__\"" in Path(shim.__file__).read_text()`).

**R1-4 · minor** — leftover "host" vocabulary contradicts the PR's stated runtime/viewer vocabulary: `local_operator/mobile/tui_handle.py:4` ("the host server's SessionHandle"), `local_operator/mobile/daemon.py:22` ("Session hosts run their own loops"), `local_operator/mobile/daemon.py:621` (`# Session host connections`), `tests/unit/mobile/test_peer_send.py:238` ("the neutral host package"). No `HostServer` identifier survives in code — good — but these four comments are the ones a reader hits first in the daemon and the TUI handle. Replace with "runtime"/"session runtime".

**R1-5 · nit** — `local_operator/cli.py:2426` and `:2958` still cite `mobile/child.py` as where the child's SIGTERM lifecycle lives; it is now `session/runtime/process.py` (the shim has no signal handling of its own). Update the two path mentions.

**R1-6 · nit** — the four shims and `test_backcompat_shims.py` contain literal `\u2014` escape sequences in source where the rest of these files (and the runtime modules) use the actual `—` character. They render correctly in `__doc__` because they sit in non-raw docstrings, but it is inconsistent and will trip a grep for the dash. Replace with the literal character.

**R1-7 · nit** — `local_operator/session/runtime/__init__.py:39-45` states as contract that this `__init__` "deliberately exports NOTHING and imports NOTHING", while DESIGN.md §3.3 places `engage_runtime()`, `find_runtime()`, `stop_session()`, `stop_all()` in this same `__init__.py` in later PRs. Both cannot hold; either the design should put those in `launch.py`/`control.py` only (with `__init__` staying empty), or this docstring should say the import-light contract binds `types`/`registry` and that any future `__init__` exports must be lazy. Worth resolving now so PR 6 doesn't break the startup-path contract while following the design literally.

**Not findings, noted for the record:** the `test_peer_send.py` guard rewrite is a legitimate tightening (asserts the actual heavy modules by name plus a fresh-subprocess measurement) rather than a loosening; `PROTOCOL_VERSION` living in `runtime/types.py` is correctly justified by the `SessionRecord.protocol` default; the `run/mobile` and thread-name rationale comments meet the why/constraints standard.

**Verdict: approve** — no blocker, no major. Round 1 is **clean and TERMINAL**; R1-1…R1-7 are follow-ups the implementer may fix in this PR or defer in the remediation comment, not a reason for another round. Merge once `test (3.12)` on head d3592f91 settles (green, or a re-run that isolates the pre-existing `test_resume_subagents_todos.py` flake).

damianvtran:
- R1-1: fixed — 81b35917
- R1-2: fixed — 81b35917
- R1-3: fixed — 81b35917
- R1-4: fixed — 81b35917
- R1-5: fixed — 81b35917
- R1-6: fixed — 81b35917
- R1-7: fixed — 81b35917

**R1-1** — I reproduced the reviewer's disproof before rewording: `session_factory` has **zero** module-scope `local_operator.session.*` imports (AST-checked; all function-local or `TYPE_CHECKING`), so hoisting `create_session` imports cleanly today and the old comment's cycle claim was false. The comment now states the operative reason as current (startup import weight — `session_factory` is the composition root and pulls the engine behind it, and this package sits on the CLI startup path) and the cycle as **latent**, naming the later change that makes it real (`session.session`/`session_factory` reaching into the runtime package for engagement and arbitration). The overstated `test_import_graph.py` reference is gone. The "do not hoist" instruction stands, now with a justification that survives being tested.

**R1-2** — `MAX_QUEUED_PROMPTS` and `PENDING_REQUEST_TIMEOUT_S` are re-exported from the `mobile/owned.py` shim. Verified both resolve at the old path and are the same objects (`32` / `30.0`).

**R1-3** — Agreed that the identity assertion could not see the `__main__` guard. Rather than assert it textually, the new `test_child_shim_is_runnable_as_a_module` **executes** the module in a subprocess under `run_name="__main__"` with the runtime's `main` stubbed to sentinel exit code 87. Proven load-bearing by mutation — deleting the guard produces exactly the failure the reviewer predicted:

```
FAILED tests/unit/session/runtime/test_backcompat_shims.py::test_child_shim_is_runnable_as_a_module
  AssertionError: `python -m local_operator.mobile.child` did not reach main(); rc=0
  assert 0 == 87
```

Restored, it passes. This closes the gap where a one-line regression silently broke the upgrade window while every identity check stayed green.

**R1-4** — All four sites reworded to runtime/viewer vocabulary (`tui_handle.py:4`, `daemon.py:22`, `daemon.py:621`, `test_peer_send.py:238`).

**R1-5** — Both `cli.py` mentions now cite `session/runtime/process.py`, which is where the SIGTERM lifecycle actually lives; the shim has none of its own.

**R1-6** — Literal `\u2014` escapes replaced with the em dash character in the four shims and the shim test. (One remaining hit in `mobile/projection.py` is pre-existing and outside this PR's diff, so left alone.)

**R1-7** — Good catch on the contradiction with the design. The docstring no longer claims `__init__` must stay empty forever; it now states the constraint that actually binds — **importing this package must never pull the heavy graph** — notes that later PRs may legitimately put `engage_runtime()`/`find_runtime()`/`stop_session()`/`stop_all()` here, and requires any such entry point to resolve heavy dependencies lazily (as `spawn_owned_session` already does), with `test_import_graph.py` as the guard. Nothing in the design is contradicted and PR 6 has an explicit rule to follow.

**Gates re-run on this head** — flake8 clean; black 26.1.0 `769 files would be left unchanged`; isort 5.13.2 clean; pyright `0 errors, 0 warnings`; `tests/unit/session/runtime tests/unit/mobile tests/unit/session tests/unit/test_import_graph.py` → **887 passed**.

**On the `test (3.12)` flake:** it went green on the previous head (`d3592f91`, run 33572666806, all 8 jobs pass) — consistent with the reviewer's independent CI-history corroboration that `test_resume_subagents_todos.py` fails intermittently on `main` too. I have deliberately not touched that file: it is a real pre-existing defect in test design, but fixing it inside a pure-move PR would put a behavioural change in a diff whose whole value is being mechanically verifiable. Flagging it for the operator rather than filing anything.

damianvtran:
### Agent review — round 2
Scope: 4da01369b796143832c75a10780f839555713b5a..81b3591734d11cc972ca9e78521ba011195a0bbe

Round 2 verifies the round-1 remediation (`81b35917`) and re-certifies the head. Everything below was checked in the worktree, not inferred from the remediation comment.

**Per-finding verification**

- **R1-1 — verified fixed.** `session/runtime/owned.py:843-862` now states startup cost as the current reason and the cycle as latent, and the `test_import_graph.py` "pins that budget" claim is gone. I re-checked the two facts the new wording rests on: AST scan of `session_factory.py` finds zero module-scope `local_operator.session.*` imports, and `grep -rl session.runtime local_operator/session` (excluding the package itself) returns nothing. The comment is now true as written.
- **R1-2 — verified fixed.** `mobile/owned.py` re-exports `MAX_QUEUED_PROMPTS` and `PENDING_REQUEST_TIMEOUT_S`; both are identity-equal to the runtime objects (`32`, `30.0`). AST scan of the origin module's public module-level names: the only one still absent from the shim is `logger`, which is a per-module logger and correctly not re-exported.
- **R1-3 — verified fixed, and it is a real subprocess exercise, not a mock.** `test_child_shim_is_runnable_as_a_module` spawns `sys.executable -c` with `runpy.run_module(..., run_name="__main__")` and asserts on the process exit code. I reproduced the mutation myself in a scratch copy (worktree untouched): head shim → `rc=87`; shim with the `if __name__ == "__main__":` block deleted → `rc=0`. The test is load-bearing.
- **R1-4 — verified fixed.** `tui_handle.py:4`, `daemon.py:22`, `daemon.py:621`, `test_peer_send.py:238` all reworded. A tree-wide grep for `host server|session host|neutral host|host package` finds only `harness/types.py:1530`, which this PR does not touch and is pre-existing.
- **R1-5 — verified fixed.** `grep -n mobile/child.py local_operator/cli.py` → no hits; both sites now cite `session/runtime/process.py`.
- **R1-6 — verified fixed.** Zero `\u2014` escapes remain in the four shims and `test_backcompat_shims.py`. The remaining hits across the tree (`builtin.py`, `config.py`, `paths.py`, `tui/app.py`, many tests) are pre-existing and out of scope; the remediation's "one hit in `projection.py`" understates the tail but that is immaterial to this PR.
- **R1-7 — verified fixed.** `session/runtime/__init__.py:37-56` now states the binding constraint as "importing this package must never pull the heavy graph", explicitly permits `engage_runtime()`/`find_runtime()`/`stop_session()`/`stop_all()` to live here, and requires lazy resolution of heavy deps. Consistent with DESIGN.md §3.3.

**Still a pure move.** `git diff origin/main...HEAD --stat`: 43 files, 3032/2585. `81b35917` touches no non-comment line in any `session/runtime/*` module (0 code-line deltas in `owned.py`; `__init__.py` is docstring-only); the only executable changes are the two added re-exports in the `mobile/owned.py` shim and the new subprocess test. No behaviour change introduced.

**Gates on head `81b35917`, over the whole tree as AGENTS.md spells them:**
- `.venv/bin/python -m flake8 .` → rc=0
- `uvx --from black==26.1.0 black --check .` → "769 files would be left unchanged", rc=0
- `uvx isort==5.13.2 --check .` → rc=0
- `.venv/bin/python -m pyright --pythonpath .venv/bin/python .` → 0 errors, 0 warnings
- `env -u NO_COLOR TERM=xterm-256color .venv/bin/python -m pytest tests/unit/session/runtime tests/unit/mobile tests/unit/session tests/unit/test_import_graph.py` → **887 passed** in 29s

**CI on head (run 33575389058, as of 00:37 UTC):** lint ✅, type-check ✅, pip-audit ✅, tui-e2e (ubuntu) ✅, tui-e2e (macos) ✅, **test (3.12) still in progress** (started 00:27 UTC). Merge should wait for it to settle; per round 1, a failure isolated to `test_resume_subagents_todos.py` is the known pre-existing flake and a re-run is the right response, anything else is new.

**New findings**

**R2-1 · nit** — `tests/unit/session/runtime/test_backcompat_shims.py:88-95`: in the driver string, `import local_operator.mobile.child as c; c.main = p.main` is dead: `runpy.run_module` re-executes the shim's source, whose `from ...process import main` already picks up the patched `p.main` from `sys.modules`. The pre-import is also what makes the subprocess print `RuntimeWarning: 'local_operator.mobile.child' found in sys.modules after import of package ... prior to execution` on stderr. Dropping those two statements keeps the test equally load-bearing (I confirmed the mutation still flips 87→0 without them) and removes the warning. Cosmetic; no action required for this PR.

No blocker, no major.

**Verdict: approve** — round 2 is **clean and TERMINAL**. R2-1 is a follow-up, not a reason for another round. Merge once `test (3.12)` on `81b35917` is green.

PR #527, opened 2026-09-01
feat(model): stale-while-revalidate catalogue and first-party-free pricing via models.dev

damianvtran:
## Summary

**Incident (2026-09-01):** Anthropic published `claude-fable-5-1` in the morning. With a 22h-old `~/.local-operator/cache/anthropic.listing.json` — inside the 24h TTL, so never refetched — `/model` did not list it, and `resolve_model_info("anthropic", "claude-fable-5-1")` returned `input/output 0.0`, `cache_read/cache_write None`: the only price source for a direct-provider id was the OpenRouter listing looked up under a per-provider namespace, and *that* document was six hours old and predated the row. A user signed in only to Anthropic ran the whole day at `$0.00`.

Two structural causes, both fixed here per `docs/design/model-catalogue-freshness.md` (committed in this PR):

1. **One 24h TTL, one synchronous miss path, no user-driven refresh.** The catalogue reader had three states — fresh/serve, expired/fetch-sync, failed/serve-stale — and nothing between "serve" and "block on the network".
2. **Direct-provider prices came from OpenRouter.** Anthropic's `/v1/models` carries no prices (verified: `id, display_name, created_at, max_input_tokens, max_tokens, capabilities`); OpenAI's and Google's neither. The aggregator leg coupled every direct provider's cost display to one third party's document and its id spellings.

## What changed

**`catalogue.py` — stale-while-revalidate.** `read_listing()` returns a `Listing(payload, age_s, fetched, failed, refreshing)` with a soft/hard TTL state machine: `age < SOFT_TTL_S (1h)` serve; `1h ≤ age < DEFAULT_TTL_S (24h)` serve **immediately** and refresh on a daemon thread (per-process in-flight set + 5-min backoff, the existing cross-process lease; a thread that loses the lease exits); `≥ 24h` or absent → the existing synchronous lease path, unchanged. `cached_listing` stays as a no-thread wrapper (soft TTL pinned to hard). `peek_listing()` reads without sweeping or fetching, for conditional-GET thunks. A `refetch_if(payload, age)` predicate lets a reader declare a document expired *before* either path starts — this is what `want_id` uses, because a second sync call after the background one is scheduled would lose the lease to its own process's thread and could not tell its result from a cache hit.

**`discovery.py`.** New `"stale"` status (a fetch was attempted and failed; before, that collapsed into `"cached"` and the picker footer could not tell "fresh enough" from "offline"). `available_models(want_id=…)`: an id absent from a usable document ≥ `MISS_REFETCH_MIN_AGE_S` (10 min) old triggers **one** synchronous refetch inside the caller's budget — the only trigger that beats every TTL *and* the in-process memo bucket for a brand-new id, and it fires for exactly the id the user asked for. A young document is believed (a typo cannot refetch per resolution), and the refetched document is 0s old so a second miss is believed too. `DiscoveredModel.cache_write_price` from `pricing.input_cache_write`, threaded through `_from_static`/`_merge_one`/`_rows_from_payload`; `LISTING_CAPTURE_VERSIONS` bumped to 2 for `openrouter` and `radient` so the field is not invisible for a day. The `fetched` nonlocal is gone.

**`prices.py` (new) — models.dev.** `https://models.dev/api.json` is keyless, provider-keyed, carries `cost {input, output, cache_read, cache_write}` in $/MTok and `limit {context, output}`, answers a weak ETag and 0-byte 304s. What lands on disk (`models-dev.listing.json`) is a **projection**: only the 13 mapped provider keys and five fields per model — 141,343 bytes measured against the 4.4 MB source. Key map with first-match-wins fallbacks (`kimi → moonshotai, kimi-for-coding`; `zai → zai, zai-coding-plan`; `alibaba-token-plan → alibaba-token-plan, alibaba`; `radient` deliberately absent — its own listing quotes prices). Lookup tries the exact id, the normalised id, and the dotted/dashed spellings. `supports_images` is **never** taken (three-valued contract: a stated `false` is the provider's denial). On a cold miss that times out inside the 3s leg budget, a background retry with the full 10s timeout is scheduled so the *next* resolution finds the document.

**`configure.py` — leg order.** Registry → provider listing (`want_id=model_id`) → **models.dev** → OpenRouter listing **only for `openrouter/*`/`radient/*` ids**. `_AGGREGATOR_NAMESPACE` and `_aggregator_spellings` are deleted; the spelling rules and `_normalised_id` move to a new dependency-free `model/ids.py` (configure is on the CLI import path and must not import discovery; discovery must not import configure back). `_info_from_discovery` and the shared `_fill_from_row` write `ModelInfo.cache_writes_price` from the real number, with the input-price stand-in kept only for listings that genuinely quote no write price (it under-states an Anthropic 5m write by 20%). Budgets unchanged in kind: `_PRICE_CATALOGUE_TIMEOUT_S = _AGGREGATOR_TIMEOUT_S = 3.0` inside `_remaining_budget(started)`; the 1 Hz TUI cost poll still only reads `_paint_memo` on-loop.

**Picker.** `_refresh_catalogue` passes `live_catalogue(ttl_s=PICKER_TTL_S)` (15 min, in `providers/controller.py`; the only caller never passed the accepted argument). `_catalogue_status`: `cached` is now **silent** ("listed within the last quarter hour" is not worth a line); `stale` → `stale list: anthropic, xai`, collapsing to `stale list: all providers` when every listed provider is stale (offline); `empty` unchanged. The dead `"unavailable"` branch that no code emitted becomes the `stale` branch. The footer also keeps its status row (blank) once one has painted, so `checking providers…` clearing on settle no longer reflows the picker — both wording updates came out of design review round 1 (see the remediation comment).

**Version:** `0.44.25 → 0.44.26` (patch: a fix plus a self-contained enrichment; no new surface).

## Testing evidence

Everything below ran from the worktree venv against **temp `cache_dir`s** (the real `~/.local-operator/cache` was read, never written). Commands and outputs verbatim; scripts in `/tmp/lop-cf-evidence/`.

### (a) Reproduction on the pre-change tree

Copied the real `*.listing.json` documents into a temp dir, backdated `fetched_at` to 22h (anthropic) / 6h (others), removed the fable row the way a pre-release capture lacks it, and blocked the network:

```
$ .venv/bin/python /tmp/lop-cf-evidence/repro.py
cache_dir: /var/folders/.../lop-cf-stale-njsb8kg2
available_models(anthropic): status=cached ids=['claude-opus-5', 'claude-sonnet-5', 'claude-fable-5', ...]
claude-fable-5-1 listed: False
resolve_model_info: input=0.0 output=0.0 cache_read=None cache_write=None ctx=1000000
```

### (b) After: correct prices with NO openrouter document present

Temp dir seeded with **only** the 22h-old anthropic document (fable removed). Real network to Anthropic (OAuth) and models.dev:

```
$ .venv/bin/python /tmp/lop-cf-evidence/after.py
cache_dir: /var/folders/.../lop-cf-after-hyzfbepw contents: ['anthropic.listing.json']
local_operator.model.catalogue: anthropic.listing catalogue (79200s old) declared expired by its reader
local_operator.model.prices: models.dev catalogue fetched: 13 providers projected, etag W/"e51cddf3dbca096a2d910c9156f261e2"

(b) resolve_model_info(anthropic, claude-fable-5-1) in 0.66s:
    input=10.0 output=50.0 cache_read=0.25 cache_write=12.5 ctx=1000000 max_tokens=128000 prompt_cache=True
    cache contents: ['anthropic.listing.json', 'models-dev.listing.json']
```

`assert not (cache / "openrouter.listing.json").exists()` held. The `declared expired by its reader` line is the `want_id` miss refetch (e) firing for the anthropic document.

### (c) The models.dev projection and a 304

```
(c) models-dev.listing.json: size=141343 bytes etag=W/"e51cddf3dbca096a2d910c9156f261e2" capture=1
    providers=['alibaba', 'alibaba-token-plan', 'anthropic', 'deepseek', 'google', 'kimi-for-coding',
               'mistral', 'moonshotai', 'openai', 'openrouter', 'xai', 'zai', 'zai-coding-plan']
    anthropic document refetched (want_id miss): fetched_at age = 0.3s, ids include fable: True
local_operator.model.prices: models.dev catalogue unchanged (304, etag W/"e51cddf3dbca096a2d910c9156f261e2")

(c) second fetch with If-None-Match (0.15s): gpt-5.4 -> in=2.5 out=15.0 cache_read=0.25 cache_write=0.0 ctx=1050000
    restamped fetched_at age = 0.0s
```

(Document aged to 25h between the calls to force the conditional GET; the 304 restamps `fetched_at` without re-downloading 4.4 MB.)

### (d) Stale-while-revalidate

Real anthropic document aged to 2h, real OAuth credential (value never printed):

```
$ .venv/bin/python /tmp/lop-cf-evidence/swr.py
credential kind: oauth (value withheld)
(d) available_models on a 2h document: status=cached rows=19 in 0.4 ms (served from disk)
    background threads: ['catalogue-revalidate:anthropic.listing']
    document rewritten: mtime advanced=True, fetched_at age=0.0s, models=11, lease files=[]
    next call: status=cached, refresh scheduled=False
```

### Stale status + footer wording (network down, 25h document)

```
$ .venv/bin/python /tmp/lop-cf-evidence/stale.py
25h document, network down -> status=stale rows=19
footer: 'stale list: anthropic'
```

### (f) The picker

**Real `ProviderController` + real credential store, temp cache dir with the 22h anthropic document (fable removed):**

```
$ .venv/bin/python /tmp/lop-cf-evidence/real_picker.py
live_catalogue(ttl_s=900) in 4.04s; statuses={'openai': 'ok', 'openai-device': 'cached', 'anthropic': 'ok', 'kimi': 'ok', 'zai': 'ok', 'zai-oauth': 'cached', 'openrouter': 'ok', 'alibaba-token-plan': 'ok', 'alibaba-token-plan-oauth': 'cached'}
anthropic rows (19): fable listed = True
fable entry: label='Claude Fable 5.1' ctx=1000000 $-1.0/-1.0
```

(`$-1.0/-1.0` is the picker's pre-existing "unknown" sentinel — picker rows price from the provider listing only, and Anthropic's quotes none; the status band prices through `resolve_model_info`, which is what (b) proves. Pricing picker rows from models.dev is out of scope and noted below.)

**Rendered frames** (real `OperatorApp` via `run_test`, 100×30, `FakeProviderController` forcing `anthropic`/`xai` to the failed-refresh status; SVG → PNG with `rsvg-convert`), before and after typing `/model fable`:

| before (`origin/main`) | after |
| --- | --- |
| `live_catalogue(ttl_s=None)`; footer `cached: openrouter`; the two failed providers are not mentioned | `live_catalogue(ttl_s=900)`; footer `stale list: anthropic, xai`; `cached` silent |
| model-picker-before | model-picker-after-stale |

A third frame with every provider `cached` shows the footer with no catalogue clause at all (the `d in /model saves this…` hint only, in the held status row):

model-picker-after-cached

Frames were captured after two `pilot.pause()` settles; the stills live on the `pr-catalogue-freshness-assets` branch (same convention as `pr-tui-ask-assets`).

### Suite and gates

`env -u NO_COLOR TERM=xterm-256color .venv/bin/python -m pytest tests/unit -q` → **10496 passed, 15 skipped, 4 errors** in 8:58. The 4 errors are `tests/unit/evaluation/runner/test_episode_subprocess.py::test_worker_cannot_deliver_frames_from_outside_the_root[*]` failing to *create a copied-interpreter venv* under xdist contention (`libpython3.14.dylib no such file`) — a host/interpreter condition in a package this PR does not touch; the same five cases pass serially (`-n0`: `5 passed in 38.26s`). `tests/unit/model` + `tests/unit/providers` re-run three times after the rebase: `1137 passed` ×3.

Gates, whole tree, as CI spells them: `flake8` clean, `black==26.1.0` clean (735 files), `isort==5.13.2` clean, `pyright` 0 errors.

New tests: `test_catalogue.py` (+11, SWR: served-and-revalidated, soft-TTL no-op, hard-TTL sync, lease loss exits, failed refresh backs off, one thread per process, `cached_listing` never threads, `peek_listing`, structural off-thread spy, `refetch_if`, `failed` flag), `test_discovery.py` (+10: stale split, four `want_id` cases, cache-write parse/merge/round-trip, capture-2 refetch), `test_prices.py` (new, 18: projection, ETag/304, stale-beats-absent, non-200 never overwrites, cold-miss background retry, capture recovery, miss refetch, five lookup spellings, unmapped provider, `supports_images` never taken), `test_pricing.py` (re-pointed at `_from_price_catalogue`; +5 incl. zero-OpenRouter-calls for a direct id and write price reaching `ModelInfo`), `test_model_picker.py`/`test_app_pilot.py` (+2: footer wording, picker TTL).

## Deferred (recorded here, not as issues)

- **xAI first-party price transport** (`GET /v1/language-models`, cents per 100M tokens): no API key available to validate the unit factor against a real response; a guessed factor on a price field is worse than no field. models.dev covers 7/12 xAI models.
- **Picker row prices from models.dev**: picker rows still price from the provider listing only (Anthropic → blank). The status band is the incident's surface and is fixed; pricing the picker is a separate change.
- `/model refresh` / `r` key — new interaction, needs a UX round; the 15-min TTL covers the incident.
- `prices_from_catalogue` registry flag so models.dev can *correct* a transcribed registry price rather than only fill a hole.
- Parallelising `live_catalogue`'s per-provider fetches (4.0s above is the serial sum).
- Direct-provider ids that OpenRouter priced but models.dev lacks (openai 4, xai 5, google 5, mistral 2 today) now fall back to the registry template — the honest "cost unavailable", per the operator's preference over a second coupling.

### DIFF
--- a/docs/design/model-catalogue-freshness.md
+++ b/docs/design/model-catalogue-freshness.md
@@ -1,0 +1,592 @@
+# Model catalogue freshness and provider-neutral pricing — design
+
+Status: proposal for one coder PR. Author: architect (lopdev team), 2026-09-01.
+Scope was extended twice mid-design by the manager; the final scope is:
+
+1. A model released today is offered today by `/model` and priced today by
+   the status band, while repeated queries stay at disk speed.
+2. Direct-provider prices stop depending on OpenRouter: a user signed in only
+   to Anthropic gets correct price analytics for `claude-fable-5-1`.
+3. `DiscoveredModel` learns `cache_write_price`.
+
+Everything below cites the working tree at `origin/main` as of today.
+
+---
+
+## 1. The problem as found
+
+### 1.1 One 24h TTL, one synchronous miss path, no user-driven refresh
+
+`catalogue.cached_listing` (`local_operator/model/catalogue.py:316-376`) has
+exactly three behaviours: age < `DEFAULT_TTL_S` (24h, `:64`) → serve from disk;
+otherwise → **synchronous** fetch under a cross-process lease, rewrite, serve;
+fetch failure → serve the stale document. There is no "serve stale AND refresh"
+state, so the only way to get a newer document is to wait for the TTL or delete
+the file.
+
+Every consumer goes through `discovery.available_models` (`discovery.py:1141`)
+→ `_available_models` (`:1190`) → `cached_listing` (`:1268`), and every
+consumer uses the default TTL:
+
+| Caller | Path | Budget | TTL passed |
+|---|---|---|---|
+| `configure._info_from_discovery` (`configure.py:918`) | session boot and, via `refresh_model_info_background`, the TUI cost poll off-loop | `DEFAULT_TIMEOUT_S`=10s blocking / `_REFRESH_TIMEOUT_S`=2s non-blocking (`:1284`) | default |
+| `configure._from_aggregator_catalogue` (`:1140`) | same resolution, leg 2 | `_AGGREGATOR_TIMEOUT_S`=3s within `_remaining_budget` (`:1139`) | default |
+| `controller.live_catalogue` (`controller.py:1049`) | `/model` picker worker, `asyncio.to_thread` per provider (`:1090`) | none beyond discovery's 10s per provider | **`ttl_s` is accepted (`:1050`) but the only caller, `tui/app.py:13655`, never passes it** |
+
+So `/model` paints the static registry, runs `_refresh_catalogue`
+(`app.py:13650`) and gets a 24h-cached answer. There is no `/model refresh`,
+no `r` key, no CLI command (`grep -n "models" local_operator/cli.py` finds
+nothing relevant). Today's incident followed exactly: `anthropic.listing.json`
+was 22h old with 10 models; `/model` did not show `claude-fable-5-1`.
+
+### 1.2 Prices for direct providers come from OpenRouter, and the same TTL
+
+Anthropic's `/v1/models` carries no prices (verified: keys are `id`,
+`display_name`, `created_at`, `max_input_tokens`, `max_tokens`,
+`capabilities`, `type`); OpenAI's and Google's listings carry none either. The
+only price source for a direct-provider id the registry has not been taught is
+`_from_aggregator_catalogue` (`configure.py:1088-1184`): it reads the keyless
+OpenRouter listing (`_AGGREGATOR_CATALOGUE = "openrouter"`, `:975`) and looks
+the id up under a per-provider namespace (`_AGGREGATOR_NAMESPACE`, `:996`).
+Today `openrouter.listing.json` was 6h old and predated OpenRouter's
+`anthropic/claude-fable-5.1` row, so `resolve_model_info("anthropic",
+"claude-fable-5-1")` returned `0.0/0.0` until the doc was force-refreshed.
+
+That leg is also structurally fragile: one third party's public listing,
+keyless *today*, with its own id spellings (`_aggregator_spellings`,
+`:1056`), gates every direct provider's cost display. The operator has ruled
+it out as the price source for direct providers.
+
+### 1.3 The write price is guessed
+
+`DiscoveredModel` (`discovery.py:149-181`) has `cache_read_price` and no write
+price. `_row_from_openai_entry` (`:396-438`) reads `pricing.input_cache_read`
+and ignores `input_cache_write`, which OpenRouter does publish (fable-5.1:
+`input_cache_write` 1.25e-5/token = $12.50/MTok). Both consumers then
+substitute the input price (`configure.py:958-959` and `:1170-1171`, whose
+comment admits the 20% understatement). The legacy `_info_from_listing`
+(`configure.py:610-621`) already reads `input_cache_write` correctly — the
+discovery path regressed it.
+
+### 1.4 The in-process memo is bucketed, not aged
+
+`resolve_model_info` memoises per `(provider, model_id, time // DEFAULT_TTL_S)`
+(`configure.py:1364`). A document refreshed on disk mid-bucket is not seen by
+this process until 00:00 UTC. That is documented and acceptable for *drift*
+(`:1340-1350`), but it means a *new id* must be refetched inside the memoised
+body on its first resolution, or the session runs unpriced all day. This is why
+the "miss" trigger (§3.4) has to live inside `_resolve_model_info_cached`, not
+beside it.
+
+---
+
+## 2. Options evaluated
+
+Costs are stated per session start (SS), boot/paint latency (L), staleness
+bound (S), offline (O), thundering herd (H), code footprint (F).
+
+| # | Option | SS network | L | S | O | H | F | Verdict |
+|---|---|---|---|---|---|---|---|---|
+| 1 | Shorter TTL only (1h) | 1 fetch/provider/hour, **synchronous on the boot path** | up to 10s per provider once an hour; the `_REFRESH_TIMEOUT_S` paint leg would hit its 2s cap hourly | 1h | fine (stale on failure) | lease covers | tiny | **Reject.** Moves the pain onto the path that must be cheap. |
+| 2 | Stale-while-revalidate (soft/hard TTL) | 0 on the calling path; 1 background fetch/provider/soft-TTL | disk speed whenever a document exists | soft TTL for the *next* call; this process's memo bucket still pins (§1.4) | unchanged: hard TTL keeps stale-beats-absent | lease + per-process in-flight set | moderate (catalogue.py only) | **Adopt** for boot/paint paths. |
+| 3 | Explicit refresh: picker passes a short TTL | 0 at boot; 1 fetch/provider/15min *only when the user opens `/model`* | none on boot; picker already paints static rows first and fetches off-loop (`app.py:13636-13648`, `controller.py:1090`) | 15 min for the picker | picker shows "no live list" | lease covers | **one argument** at `app.py:13655` | **Adopt.** Fixes the reported incident by itself. |
+| 4 | Event-driven: refetch on a requested-id miss | 1 fetch per *new id* per document, rate-limited by document age | bounded by the caller's existing budget (`_remaining_budget`) | same day for the id the user actually asked for | fails → registry template, as today | lease covers; a young document is believed, so a typo cannot loop | small (discovery + configure) | **Adopt.** The only option that beats the memo bucket for a brand-new id. |
+| 5 | Per-provider TTLs | fewer fetches for first-party | none | worse for first-party (the ones that matter) | — | — | small but a second knob | **Reject.** With SWR a background fetch per provider per hour is already cheap; distinguishing classes buys nothing measurable and every class needs "today". |
+
+Recommendation: **2 + 3 + 4 together**, one mechanism each, no per-provider
+TTLs, no new user command in this PR.
+
+---
+
+## 3. Recommended design
+
+### 3.1 `catalogue.py`: stale-while-revalidate with soft/hard TTL
+
+Add alongside `DEFAULT_TTL_S` (keep its name and value — it is the memo bucket
+in `configure.py:1364,1424` and the hard TTL):
+
+```python
+#: Age past which a served document is ALSO refreshed in the background. An
+#: hour bounds "released today, offered today" for the next call without
+#: putting a request on the calling path; below it a listing endpoint would be
+#: asked repeatedly for information that changes on the order of days.
+SOFT_TTL_S = 60 * 60
+#: A document younger than this that lacks a requested id is believed: the id
+#: is genuinely unknown, not newly released. Bounds the miss-triggered refetch
+#: (see discovery.available_models(want_id=...)) so a typo cannot refetch on
+#: every resolution.
+MISS_REFETCH_MIN_AGE_S = 10 * 60
+#: Per-process floor between background attempts on one key, so an offline
+#: machine does not spawn a thread every time a stale document is served.
+REVALIDATE_BACKOFF_S = 5 * 60
+```
+
+Introduce a result type and a richer reader; keep `cached_listing` as a thin
+wrapper so its 20+ existing tests and both callers keep working unchanged:
+
+```python
+@dataclasses.dataclass(frozen=True)
+class Listing:
+    payload: dict[str, Any] | None
+    age_s: float            # inf when absent
+    fetched: bool           # a live fetch produced this payload on THIS call
+    refreshing: bool        # a background revalidation was scheduled by this call
+
+def read_listing(key, fetch, *, soft_ttl_s=SOFT_TTL_S, ttl_s=DEFAULT_TTL_S,
+                 cache_dir=None) -> Listing: ...
+
+def cached_listing(key, fetch, *, ttl_s=DEFAULT_TTL_S, cache_dir=None):
+    return read_listing(key, fetch, ttl_s=ttl_s, soft_ttl_s=ttl_s, cache_dir=cache_dir).payload
+```
+
+State machine in `read_listing` (replaces `catalogue.py:343-376`):
+
+1. `age < soft_ttl_s` → serve, no I/O beyond the read.
+2. `soft_ttl_s <= age < ttl_s` → serve **immediately**, then
+   `_schedule_revalidate(key, fetch, cache_dir)`.
+3. `age >= ttl_s` or missing → today's synchronous lease path, unchanged
+   (`:349-376`). The `fetched` flag replaces the `nonlocal fetched` trick in
+   `discovery.py:1229-1246`, which can then be deleted.
+
+`_schedule_revalidate`:
+
+- Per-process dedupe: module-level `_revalidating: set[str]` and
+  `_last_attempt: dict[str, float]` under one `threading.Lock`; return
+  without a thread if the key is in flight or attempted within
+  `REVALIDATE_BACKOFF_S`.
+- Thread body: `lease = _ListingFetchLease(path)`; if `acquire()` is False,
+  **return** (a peer is fetching; unlike the sync path there is nothing to wait
+  for). Else `fetch()` → `_write_cache` → `release()`, any exception logged at
+  debug, `finally` clears the in-flight entry. `daemon=True`, named
+  `catalogue-revalidate:<key>`. A daemon thread killed by process exit leaves
+  a lease that lapses in `_LISTING_LEASE_S` (60s, `:221`) — the existing
+  crash contract, no new state.
+- Plain `threading.Thread`, not the asyncio loop: `catalogue.py` is
+  synchronous and is entered from the CLI, the server, `asyncio.to_thread`
+  workers and `run_in_executor` threads (`configure.py:1478`). The existing
+  precedent for "fire and forget off the calling path" is
+  `refresh_model_info_background` (`configure.py:1431`); this is the same
+  shape one layer down.
+
+`fetch` thunks capture the caller's `client` and credentials
+(`discovery.py:1233-1250`); running one in a thread is safe because
+`fetch_models` opens and closes its own `httpx.Client` when none is passed
+(`discovery.py:862`). The tests' `_StubClient` is passed explicitly and is
+single-threaded; tests that exercise the background path must join the thread
+(§5).
+
+**Conditional GET support (in scope, small).** Add
+`peek_listing(key, cache_dir=None) -> Listing` (read-only, no fetch, no
+sweep). A fetch thunk that wants `If-None-Match` reads the previous payload's
+`etag` through it and, on 304, returns the previous payload unchanged;
+`_write_cache` re-stamps `fetched_at`, which is exactly "validated just now".
+No change to `read_listing`'s signature. Used by the price catalogue (§3.5);
+verified live: models.dev answers `304` with 0 bytes to a matching
+`If-None-Match` (ETag `"2afcb862…"`).
+
+### 3.2 `discovery.py`: statuses, `want_id`, `cache_write_price`
+
+**Statuses.** Extend `ListingStatus` (`:145`) to
+`"ok" | "cached" | "stale" | "static" | "unauthenticated" | "empty"`:
+
+- `"ok"` — fetched live on this call (unchanged).
+- `"cached"` — served from disk, no fetch attempted or needed (age < TTL).
+- `"stale"` (new) — a fetch was attempted and failed; the document on disk is
+  what you got. Today both collapse into `"cached"` (`:1308-1309`), which is
+  why the picker footer cannot tell "fresh enough" from "offline".
+  `tui/app.py:19960` already tests for a status `"unavailable"` that no code
+  emits — that dead branch becomes the `"stale"` branch.
+
+`_available_models` computes them from `Listing.fetched` and from whether the
+thunk raised (`_ListingUnavailable` is still how the thunk reports failure,
+`:186`). The `fetched` nonlocal (`:1229`) goes away.
+
+**`want_id` (option 4).** Add `want_id: str | None = None` to
+`available_models`/`_available_models`. After mapping rows (`:1269`), if
+`want_id` is set, no row's id matches it exactly *or* after the same
+normalisation `_info_from_discovery` applies (`configure._normalised_id`,
+`:868` — move the helper to discovery or import it lazily; it is pure), the
+listing was **not** fetched on this call, and `listing.age_s >=
+MISS_REFETCH_MIN_AGE_S`: call `read_listing(..., ttl_s=0, soft_ttl_s=0)` once
+(sync, under the lease, within the caller's `timeout` — which is the caller's
+remaining budget, see §3.3) and re-map. Status becomes `"ok"` or `"stale"`
+accordingly. One retry, never a loop: the refetched document is 0s old, so a
+second miss for the same id in the same process is believed for ten minutes.
+
+**`cache_write_price`.** Add `cache_write_price: float = 0.0` to
+`DiscoveredModel` after `cache_read_price` (`:180`), and thread it through:
+
+- `_row_from_openai_entry` (`:415-437`): `_per_million(pricing.get("input_cache_write"))`.
+  Ignore `input_cache_write_1h`.
+- `_from_static` (`:945`): `_positive_float(info.cache_writes_price)`.
+- `_merge_one` (`:1006`): live-or-static, same rule as `cache_read_price`.
+- `_rows_from_payload` (`:1129`): `_positive_float(entry.get("cache_write_price"))`.
+- `LISTING_CAPTURE_VERSIONS` (`:118`): `{"anthropic": 2, "openrouter": 2,
+  "radient": 2}` with a docstring line per bump. The stamp is per provider id
+  (`listing_capture_version`, `:125`), so only the two OpenAI-compat
+  aggregators that quote a write price pay the one-time refetch. A version-1
+  document parses to rows with `cache_write_price=0`, which is harmless, but
+  without the bump the fix would be invisible for a day on every install — the
+  same argument as `:110-117`.
+
+**xAI first-party prices (best-effort, optional in this PR).** xAI documents
+`GET /v1/language-models` returning `prompt_text_token_price`,
+`completion_text_token_price`, `cached_prompt_text_token_price` in USD cents
+per 100M tokens (÷ 10 000 → $/MTok). The manager's OAuth credential gets 403
+on it, so it is API-key only. Design: a provider-keyed override table
+`_PROVIDER_TRANSPORTS = {"xai": _fetch_xai}` consulted before
+`_WIRE_TRANSPORTS` in `_build_transports` (`:793-808`); `_fetch_xai` tries the
+priced endpoint and on any non-200 falls back to `_fetch_openai_compat`
+unchanged. Bump `LISTING_CAPTURE_VERSIONS["xai"]`. **Defer to a follow-up PR
+unless the coder has an xAI API key to prove the conversion against a real
+response** — models.dev covers 7/12 xAI models today, and a guessed unit
+factor on a price field is worse than no field.
+
+### 3.3 `configure.py`: resolution legs and budgets
+
+`_resolve_model_info_cached` (`:1241-1311`) becomes:
+
+```
+info = _registry_fallback(...)                                   # unchanged
+if _listing_can_correct(info):
+    info = _info_from_discovery(..., want_id=model_id, timeout=…)  # leg 1: provider listing
+if _needs_enrichment(info):
+    info = _from_price_catalogue(canonical, model_id, info, timeout=_remaining_budget(started))  # leg 2: models.dev (NEW)
+if _needs_enrichment(info) and canonical in AGGREGATOR_PROVIDERS:
+    info = _from_aggregator_catalogue(...)                        # leg 3: only for aggregator ids
+```
+
+Precedence stays "fill the holes" (`_from_aggregator_catalogue` only writes
+where `not (info.input_price or info.output_price)`, `:1159`; same for the
+window `:1172` and `max_tokens` `:1182`). A registry price is a first-hand
+transcription and is not overridden by the catalogue in this PR; a
+`prices_from_catalogue` flag mirroring `limits_from_listing`
+(`configure.py:677-701`) would let the catalogue correct it and is deferred.
+
+`_info_from_discovery` (`:886`): pass `want_id=model_name` through to
+`available_models`; take `cache_writes_price` from `row.cache_write_price`
+when `> 0`, else keep the input-price fallback at `:958-959` (now only for
+listings that genuinely quote no write price).
+
+`_from_aggregator_catalogue` (`:1088`): gate on `provider in
+AGGREGATOR_PROVIDERS` (it is the `openrouter/*` and `radient/*` ids' own
+listing, so leg 1 has normally already priced them; this leg survives only for
+the case where leg 1 was unavailable). Delete `_AGGREGATOR_NAMESPACE` (`:996`)
+and `_aggregator_spellings` (`:1056`) — their only purpose was direct-provider
+lookup. Their spelling rules (`_DOTTED_VERSION_RE`, `:1017`) move to the
+price-catalogue lookup, which needs them for the same reason (`gpt-5.4` vs
+`gpt-5-4`). Existing tests that assert direct providers price via OpenRouter
+(`tests/unit/model/test_configure.py`, the `_from_aggregator_catalogue`
+cases) are re-pointed at `_from_price_catalogue` with the same fixtures
+reshaped to the models.dev projection.
+
+Budgets are unchanged in kind: leg 2 caps at
+`_PRICE_CATALOGUE_TIMEOUT_S = _AGGREGATOR_TIMEOUT_S = 3.0` inside
+`_remaining_budget(started)` — the same "one ceiling for the whole
+resolution" rule documented at `:1028-1050`. Leg 3 shares that remainder. The
+paint path is unaffected: `tui/costs.py:51-61` only ever reads
+`_paint_memo` on-loop and runs the full resolver via
+`refresh_model_info_background` in an executor thread.
+
+The `want_id` miss refetch on leg 1 runs inside the same `timeout` the leg
+already has (10s blocking at boot when the registry cannot describe the model
+— the user is watching a spinner for exactly this; 2s when only correcting a
+complete row, in which case a miss is not possible because the registry row
+*is* the answer).
+
+### 3.4 Freshness triggers, summarised
+
+| Trigger | Where | Mechanism | Bound |
+|---|---|---|---|
+| Any read of a document older than 1h | `read_listing` | background thread, lease, backoff | 0 on the calling path; ≤1 fetch/key/hour/process |
+| `/model` opened | `app.py:13655` → `live_catalogue(ttl_s=PICKER_TTL_S)` | existing sync fetch off-loop, lease | ≤1 fetch/provider/15 min, only when the user asks |
+| Requested id absent from a cached document ≥10 min old | `available_models(want_id=…)` | one sync refetch within the caller's remaining budget, lease | ≤1 refetch per key per 10 min, regardless of how many ids miss |
+| Capture version bump | `_rows_from_payload` (`:1080`) | existing invalidate+re-enter (`:1270-1289`) | once per upgrade |
+
+`PICKER_TTL_S = 15 * 60` lives in `providers/controller.py` beside
+`live_catalogue`: "the user is asking now; fifteen minutes is short enough
+that a release announced during a working session shows on the next open, and
+long enough that scrolling in and out of the picker does not re-list nine
+providers."
+
+### 3.5 Provider-neutral price catalogue: models.dev
+
+New module `local_operator/model/prices.py` (the manager asked for the document
+to go through the same machinery; a separate module keeps the 1,300-line
+`discovery.py` from growing a fourth transport that is not a provider).
+
+**Source.** `https://models.dev/api.json`: public, keyless, 4.44 MB, weak
+ETag, `cache-control: max-age=0, must-revalidate`, honours `If-None-Match`
+with a 304. Provider-keyed; each model carries `cost {input, output,
+cache_read, cache_write}` in $/MTok, `limit {context, output}`, `name`,
+`attachment`, `reasoning`, `tool_call`, `release_date`. Verified today:
+`anthropic.models["claude-fable-5-1"]` → cost 10/50/0.25/12.5, limit
+1,000,000/128,000, `release_date` 2026-09-01. Coverage for our providers
+(priced/total): anthropic 14/14, openai 43/47, xai 7/12, deepseek 3/3, zai
+16/16, google 33/38, mistral 32/34, moonshotai 10/10, alibaba 55/55,
+alibaba-token-plan 26/26, openrouter 349/353. **radient is absent** — its own
+listing quotes prices, so leg 1 covers it.
+
+**Document.** Key `models-dev.listing` → `~/.local-operator/cache/models-dev.listing.json`
+(ends in `.listing`, so `_LEGACY_DOCUMENT_GLOB` cannot match it,
+`catalogue.py:77`). Store a **projection**, not the 4.4 MB body:
+
+```json
+{"capture": 1, "etag": "\"2afcb…\"",
+ "providers": {"anthropic": {"claude-fable-5-1": {"name": …, "cost": {...}, "limit": {...}, "attachment": true, "release_date": "2026-09-01"}}, …}}
+```
+
+Measured: the projection over the 13 provider keys we map is **141 KB** — ~30x
+smaller than the source and a quarter of `openrouter.listing.json`, so the
+25 ms JSON parse the memo exists to avoid (`configure.py:1338`) stays in that
+range. Only providers in `_PRICE_CATALOGUE_KEYS` are kept.
+
+**Key map** (`_PRICE_CATALOGUE_KEYS: dict[str, tuple[str, ...]]`, first match
+wins; the local id is the *canonical* provider after
+`credential_provider_id`, so `xai-oauth`/`openai-device` need no entry):
+
+```
+anthropic → ("anthropic",)      openai → ("openai",)         google → ("google",)
+xai → ("xai",)                  deepseek → ("deepseek",)     mistral → ("mistral",)
+kimi → ("moonshotai", "kimi-for-coding")    zai → ("zai", "zai-coding-plan")
+alibaba → ("alibaba",)          alibaba-token-plan → ("alibaba-token-plan", "alibaba")
+openrouter → ("openrouter",)
+```
+
+**Lookup.** `price_catalogue_row(provider, model_id, *, timeout) ->
+DiscoveredModel | None`: `read_listing("models-dev.listing", fetch,
+...)`, then for each mapped key try the id exactly, then
+`_normalised_id`, then the dotted/dashed spellings that `_aggregator_spellings`
+produces today. Returns a `DiscoveredModel` (so `cache_write_price`, window and
+`max_tokens` ride the same struct the other legs use) with
+`supports_prompt_cache = cache_read > 0`, `supports_images` from
+`attachment`... **no**: `supports_images` is deliberately not taken from a
+second-hand source (`configure.py:1116-1118` states the three-valued
+contract); leave it `None`.
+
+**Fetch thunk.** `httpx.get(URL, headers={"If-None-Match": prev_etag} if
+prev_etag else {}, timeout=timeout)`; 304 → return the previous payload
+(read via `peek_listing`); 200 → project and stamp the new ETag; anything
+else → raise (`read_listing` turns that into stale-beats-absent).
+`want_id` miss semantics apply exactly as in §3.2: a direct-provider id
+absent from a document ≥10 min old triggers one refetch — which for models.dev
+is usually a 0-byte 304 unless the row really did just land.
+
+**Cold start.** The very first resolution on a machine with no document pays
+one 4.4 MB download inside the 3s leg budget. On a slow link that times out
+and the session runs unpriced until the next resolution — so on a cold miss
+that times out, `read_listing` additionally schedules a background fetch with
+the full `DEFAULT_TIMEOUT_S` (the one case where a failed sync fetch is
+followed by a background retry: a timeout on a large body is evidence the
+budget was too small, not that the network is down). Every later start is a
+141 KB read and, hourly, a 304.
+
+### 3.6 Picker: what the user sees
+
+`_populate_model_picker` (`app.py:13622`) is unchanged: static rows paint on
+the keystroke with the footer clause `checking providers…`.
+`_refresh_catalogue` (`:13650`) calls `live_catalogue(ttl_s=PICKER_TTL_S)`.
+`_catalogue_status` (`:19946`) becomes:
+
+- `"cached"` → **silent**. With a 15-minute picker TTL "cached" means "listed
+  within the last quarter hour", which is not something the user needs to
+  read; the existing docstring's "a footer that always says something is a
+  footer nobody reads" applies.
+- `"stale"` → `live list unavailable: anthropic, xai — showing cached`.
+- `"empty"` → `no live list: …` (unchanged).
+
+Nothing new animates; the widget API (`ModelPicker.set_rows`,
+`model_picker.py:312`) is untouched. This footer wording is a user-visible
+string change and needs the designer round with before/after stills of the
+open picker in the `stale` state (use `FakeProviderController.live_catalogue`
+in `tests/unit/tui/test_app_pilot.py:3672` to force statuses, then
+`app.save_screenshot`).
+
+**Deferred:** a `/model refresh` argument (or `r` in the picker, mirroring
+`/usage`'s `force_refresh`, `app.py:15089-15123`) that calls
+`live_catalogue(ttl_s=0)`. Cheap, but it is a new interaction and would add a
+UX round to this PR; the 15-minute TTL already covers the incident.
+
+---
+
+## 4. Concurrency story
+
+- **Cross-process**: `_ListingFetchLease` (`catalogue.py:229`) unchanged and
+  reused by all three fetch paths. Background revalidators that lose the lease
+  simply exit; sync fetchers that lose it keep today's brief wait
+  (`await_peer_briefly`, `:283`).
+- **In-process**: one in-flight set per key for background work; the
+  miss-refetch and the picker's sync fetch both go through the lease, so two
+  threads in one process (the picker worker and a `refresh_model_info_background`
+  executor) resolving the same provider at once produce one request.
+- **Write safety**: `_write_cache` (`:148`) is already mkstemp+rename; the
+  background writer adds nothing new.
+- **Memo**: `_resolve_model_info_cached`'s bucket is unchanged. The new-id
+  case is handled *inside* the memoised body by `want_id`, so the memo caches
+  the fresh answer. `invalidate_model_info_cache` (`:1314`) is untouched.
+
+---
+
+## 5. Test plan
+
+Conventions: `tests/unit/model/test_catalogue.py` uses `tmp_path` cache dirs
+and a `calls` list in the thunk; `test_discovery.py` uses `_StubClient` /
+`_Response` and `cache_dir=tmp_path`; AGENTS.md "Timing" forbids clock waits
+— join threads or wait on the document's existence, never `sleep`.
+
+`test_catalogue.py`
+
+- `test_a_document_between_soft_and_hard_ttl_is_served_and_revalidated_in_the_background`
+  — plant `fetched_at = now - 2h`; `read_listing` returns the old payload
+  with `refreshing=True` and zero sync calls; join `catalogue._revalidation_threads()`
+  (expose a test hook that returns live threads) and assert the file now holds
+  the new payload.
+- `test_a_document_under_the_soft_ttl_schedules_nothing`.
+- `test_a_document_past_the_hard_ttl_still_fetches_synchronously` (guards the
+  offline/cold contract; extends `test_an_expired_cache_refetches`).
+- `test_a_background_revalidation_that_loses_the_lease_exits_without_fetching`
+  — plant an unexpired `.fetching` file.
+- `test_a_failed_background_revalidation_keeps_the_stale_document_and_backs_off`
+  — thunk raises; second `read_listing` within `REVALIDATE_BACKOFF_S` spawns
+  no thread (assert via the hook, not timing).
+- `test_two_reads_in_one_process_spawn_one_revalidation`.
+- `test_cached_listing_is_unchanged_for_existing_callers` — existing tests
+  already prove this; add one asserting `cached_listing` never schedules a
+  thread (its soft TTL equals its hard TTL).
+- `test_peek_listing_reads_without_sweeping_or_fetching`.
+- `test_a_background_thread_runs_off_the_calling_thread` — structural
+  `threading.get_ident()` spy, per AGENTS.md.
+
+`test_discovery.py`
+
+- `test_available_models_reports_stale_when_a_refetch_fails` (split from
+  `test_available_models_reports_cached_when_a_stale_refetch_fails`, `:1236`).
+- `test_a_missing_want_id_refetches_a_document_old_enough_to_be_wrong` — 10
+  models cached 22h ago, stub answers 19 including the id; status `"ok"`, one
+  call, row present.
+- `test_a_missing_want_id_is_believed_when_the_document_is_young` — cached
+  1 min ago, zero calls.
+- `test_a_want_id_that_is_present_makes_no_request`.
+- `test_openrouter_cache_write_price_is_read_and_scaled` (extend
+  `test_openai_compat_parses_a_captured_openrouter_payload`, `:154`, with an
+  `input_cache_write` in the fixture).
+- `test_merge_prefers_a_live_cache_write_price` / `..._keeps_the_static_one_when_live_is_zero`.
+- `test_a_cache_write_price_survives_the_disk_round_trip` (`_rows_from_payload`).
+- `test_an_openrouter_document_from_capture_one_is_refetched_once` (mirrors `:1372`).
+
+`tests/unit/model/test_prices.py` (new)
+
+- Projection: a captured models.dev fixture (trim to two providers) projects to
+  only mapped providers and the five fields; size assertion is structural (no
+  `description` key survives), not numeric.
+- `test_a_matching_etag_returns_the_previous_payload_and_restamps_it` — stub
+  304.
+- Lookup: exact id, normalised id, dotted/dashed spelling, `kimi →
+  moonshotai` then `kimi-for-coding`, unknown provider → `None`.
+- `test_supports_images_is_never_taken_from_the_price_catalogue`.
+
+`tests/unit/model/test_configure.py`
+
+- `test_a_direct_provider_is_priced_from_the_neutral_catalogue_not_openrouter`
+  — `available_models` stub for anthropic returns the row unpriced; models.dev
+  stub prices it; OpenRouter stub asserts **zero** calls.
+- `test_the_write_price_reaches_model_info` — `cache_writes_price == 12.5`,
+  not `10.0`.
+- `test_the_input_price_fallback_survives_a_listing_with_no_write_price`.
+- `test_leg_two_shares_the_resolution_budget` — extend the existing
+  `_remaining_budget` test to the new leg.
+- `test_openrouter_ids_still_fall_back_to_their_own_listing`.
+- Re-point the direct-provider `_from_aggregator_catalogue` tests as in §3.3.
+
+`tests/unit/tui`
+
+- `test_the_picker_asks_for_a_fifteen_minute_listing` — the pilot's
+  `FakeProviderController.live_catalogue(self, *, ttl_s=None)` records
+  `ttl_s`; assert `== PICKER_TTL_S`.
+- `test_model_picker.py`: footer renders `live list unavailable: …` for
+  `stale` and nothing for `cached`.
+
+Real-execution evidence for the PR (per the operator's standing rule): with
+`anthropic.listing.json` and `models-dev.listing.json` deleted, boot `lop`
+on `anthropic/claude-fable-5-1`, show the status band pricing at 10/50, show
+both documents written, `stat` their ages; then age `anthropic.listing.json`
+to 2h by editing `fetched_at`, open `/model`, capture the picker before and
+after the worker lands, and show the document's `fetched_at` advanced with
+the `.fetching` lease gone.
+
+---
+
+## 6. Constants, in one place
+
+| Name | Module | Value | Why |
+|---|---|---|---|
+| `DEFAULT_TTL_S` | catalogue | 24h (unchanged) | hard TTL; memo bucket; sync fetch beyond it keeps offline semantics |
+| `SOFT_TTL_S` | catalogue | 1h | staleness bound for the next call at zero on-path cost |
+| `MISS_REFETCH_MIN_AGE_S` | catalogue | 10 min | a young document that lacks an id is right; bounds typo refetches |
+| `REVALIDATE_BACKOFF_S` | catalogue | 5 min | one background attempt per key per five minutes while offline |
+| `PICKER_TTL_S` | controller | 15 min | the user is asking; sync fetch is already off-loop behind painted rows |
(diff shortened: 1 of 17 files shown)


damianvtran:
### Agent review — round 1

Scope: 4da01369b796143832c75a10780f839555713b5a..0e608b0f11074b036ac58e72b02980c3cdeb5f7e

Gates run in the worktree over the whole tree as CI spells them: flake8 clean, `black==26.1.0` clean (764 files), `isort==5.13.2` clean, pyright 0 errors; `tests/unit/model tests/unit/providers tests/unit/tui/test_model_picker.py tests/unit/tui/test_app_pilot.py` → 1454 passed. CI on head `0e608b0f`: lint / type-check / test (3.12) / pip-audit / tui-e2e (ubuntu, macos) all green. Testing evidence is real execution (scripts in `/tmp/lop-cf-evidence/`, outputs quoted verbatim, the three PNGs exist on `pr-catalogue-freshness-assets`). I also exercised the live paths myself: cold models.dev fetch 0.64 s → 141,343-byte projection with 13 providers; dated snapshot (`claude-opus-4-5-20251101`), dotted/dashed (`gpt-5-4` → `gpt-5.4`), `models/`-prefixed Gemini, and `openrouter/anthropic/claude-opus-4.5` all resolve; `alibaba-token-plan` rows stay at 0/0 (credits, not USD), so the "zero is unknown, not free" rule survives `_fill_from_row`.

#### Findings

**R1-1 · MAJOR · `want_id` puts a synchronous listing fetch on every process start for undated alias ids** — `local_operator/model/discovery.py` `lacks_want_id` / `_lists_id`.
`_lists_id` matches exact or `normalised_id` only. Anthropic's `/v1/models` lists dated snapshots (`claude-sonnet-4-5-20250929`, `claude-opus-4-5-20251101`) but not the undated aliases its API accepts, and the registry carries only the dated rows, so `anthropic/claude-sonnet-4-5` (or `-opus-4-5`, `-haiku-4-5`) enters leg 1 → the document is ≥10 min old and "lacks" the id → sync refetch → the refetched document still lacks it. The result is believed for 10 min, then the next process pays again. Reproduced with `fetch_models` stubbed and the memo cleared between calls (= three process starts, document aged 11 min between them):
```
claude-sonnet-4-5   process 1: listing fetches this process=1
claude-sonnet-4-5   process 2: listing fetches this process=1
claude-sonnet-4-5   process 3: listing fetches this process=1
claude-sonnet-4-5-20250929  process 1..3: listing fetches this process=0
```
Before this PR a 22h document meant zero requests on boot; now a common configured id costs one blocking round trip (≈0.4–0.7 s measured, 10 s ceiling on a bad link) at every session start, indefinitely — exactly the boot-path property the design doc says the change preserves. The PR's own "a typo cannot refetch per resolution" holds within a process, not across them. Obvious fix: have `_lists_id` also accept a row whose `_DATE_SUFFIX_RE`-stripped id (or any `id_spellings` form) equals the wanted id — `ids.py` already has both helpers — so an alias whose snapshot is listed is a hit, not a miss. Add a test with the dated row present and `want_id="claude-sonnet-4-5"` asserting zero calls.

**R1-2 · MINOR · background revalidation inherits the on-path timeout instead of the full one** — `configure._resolve_model_info_cached` → `_info_from_discovery(timeout=_REFRESH_TIMEOUT_S)` → the `fetch` thunk captured by `_schedule_revalidate`.
The off-path refresh's whole point is that it may take as long as it needs, but the thunk closes over the caller's budget: for a complete registry row that is 2.0 s (`_REFRESH_TIMEOUT_S`), for an incomplete one whatever `_remaining_budget` was. Reproduced: a 2 h document resolved through `resolve_model_info` spawns `catalogue-revalidate:anthropic.listing` whose `fetch_models(... timeout=2.0)`. On a link where the listing takes >2 s the background refresh fails every time, backs off 5 min, and the document only ever advances via the 24 h sync path — the pre-PR behaviour. `prices.py` already does the right thing (`_fetch(DEFAULT_TIMEOUT_S, ...)` for its background retry); discovery should build the thunk the same way, or `_schedule_revalidate` should take a separate background thunk.

**R1-3 · MINOR · interpreter exit can strand `<key>.json.*.tmp` files that nothing sweeps** — `catalogue._schedule_revalidate` + `_write_cache`.
The thread is daemon by design; a short `lop --model …` run that serves a soft-window document exits while the thread is inside `json.dump` on the `mkstemp` file, and the docstring's "leaves nothing half-written" is true of the document but not of the temp file. `purge_legacy_documents` globs only `*.models.json`, so each such exit leaves a ~190 KB orphan in `~/.local-operator/cache` forever. Either serialise to bytes first and do one `os.write`, or add `*.listing.json.*.tmp` older than a few minutes to the purge glob.

**R1-4 · MINOR · `_PRICE_CATALOGUE_KEYS` docstring describes a merge that `_lookup` does not perform** — `prices.py:84-94`.
"its own key comes first (plan-specific limits) and the pay-per-token key fills the price" — `_lookup` returns the FIRST entry that exists and never consults the second key for the same id. For `alibaba-token-plan/qwen3.8-max` that yields 0/0 from the plan key while `alibaba` carries 2/6. The *behaviour* is right (the registry says token-plan prices are deliberately 0 because it bills credits), so this is a documentation defect only, but the comment invites someone to "fix" it into quoting USD for a credit plan. Reword to say first match wins outright and that the plan key's zeros are intended.

**R1-5 · MINOR · a `want_id` refetch that loses the lease to this process's own background thread stalls the caller up to 2 s** — `read_listing` sync path after `refetch_if`.
Call A (id X present, doc in the soft window) schedules a background refresh; call B moments later (id Y missing, ≥10 min old) is declared expired, `acquire()` fails against the thread's lease, and `await_peer_briefly` polls for up to `_LISTING_LEASE_WAIT_S` (2 s) — on the calling path. The docstring only closes this race within one call. Bounded and rare, so minor; worth a sentence in the docstring and, if cheap, a same-process fast path (join the in-flight thread with the remaining budget instead of polling mtime).

**R1-6 · NIT · `_from_aggregator_catalogue` docstring promises `radient/*` coverage it never delivers** — it returns `info` for any provider outside `PUBLIC_LISTING_PROVIDERS`, so radient never reaches the lookup. Say "openrouter only" or drop the radient claim.

**R1-7 · NIT · `LISTING_CAPTURE_VERSIONS` comment says the capture bump "only" costs the two aggregators one refetch** — true, but for an OpenRouter-only user that refetch is the invalidate-and-re-enter *sync* fetch on the first resolution after upgrade (up to the 10 s ceiling). Exactly once, so acceptable; the comment should say it lands on-path.

#### Not findings, checked
- Thread-safety: `_revalidating`/`_last_attempt`/`_threads` are only touched under `_revalidate_lock`; `_threads[slot]` is set before `start()`; `finally` releases the lease and clears the slot even when `acquire()` returned False (`release()` is a no-op when not held). Tests join via `_revalidation_threads()`; the discovery/pricing tests that plant 22 h documents take the sync path and spawn no threads, so per-test `tmp_path` keying gives isolation without the fixture.
- Offline: 25 h document + `ConnectError` → `stale` with the old rows; cold + offline → `static`; models.dev 503 never overwrites the document (test present, and `_fetch` raises on non-200).
- 1 Hz cost poll: still reads `_paint_memo` only; the three legs share `_remaining_budget(started)` with 3 s per-leg caps.
- ETag/304: live conditional GET observed re-stamping `fetched_at` at 0.15 s; capture-0 document → invalidate + same-call refetch.
- `merge_models`/`_info_from_discovery`: `cache_write_price` threaded through `_from_static`/`_merge_one`/`_rows_from_payload` with the same `_positive_float` gating as the other prices; the input-price stand-in survives only when no write price is quoted.
- Footer wording: `cached` silent, `stale` → "live list unavailable: … — showing cached", `empty` unchanged; the dead `"unavailable"` branch is gone.
- Version bump 0.44.26 (patch) matches the materiality rule.

#### Verdict
**Not clean — 1 major.** R1-1 needs a remediation round (fix + test); R1-2..R1-5 are minor and may be fixed here or deferred in the remediation comment; R1-6/R1-7 are nits.

Severity count: blocker 0 · major 1 · minor 4 · nit 2.

damianvtran:

Implementer: coder subagent (anthropic/claude-fable-5-1)
Scope reviewed: `4da01369..0e608b0f` · Head after remediation: `d6606182`

| Finding | Disposition | Commit |
|---|---|---|
| **R1-1** MAJOR — `want_id` sync-fetches on every process start for undated aliases | **Fixed** — `_lists_id` now treats a row as present when any of its `ids.id_spellings` (as given, date-stripped, dotted) coincides with any of the wanted id's after normalisation, in either direction. Parametrised test over `claude-sonnet-4-5`/`opus`/`haiku` aliases, the reverse (alias listed, snapshot wanted), and `gpt-5.4`↔`gpt-5-4`; fails on the old code (6/6 `stale`≠`cached`), passes now. A companion test pins that a genuinely new family id (`claude-fable-5-1` against a `sonnet-4-5` document) still refetches. Design doc §3.2 amended. | `8237db6a` |
| **R1-2** MINOR — background revalidation inherits the on-path timeout | **Fixed** — `read_listing` takes a separate `revalidate` thunk (defaults to `fetch`); discovery and prices hand it one carrying `DEFAULT_TIMEOUT_S`. Tests in all three modules assert the background call sees 10.0 and the sync call keeps the caller's budget. | `04f042a5` |
| **R1-3** MINOR — orphaned `*.json.*.tmp` from a daemon thread killed mid-write | **Fixed** — `purge_stranded_temp_files` sweeps `*.listing.json.*.tmp` older than 5 min on every `read_listing`, beside the legacy purge. Age-gated so a peer's in-flight write survives; scoped to this module's name shape so `update.py`'s temp files in the shared dir are left alone. Tests cover old/young/foreign files, the `read_listing` sweep, and a missing dir. | `84337318` |
| **R1-4** MINOR — `_PRICE_CATALOGUE_KEYS` comment describes a merge `_lookup` never does | **Fixed** — comment now says first match wins outright and that the token plan's 0/0 is the intended "cost unknown", not a hole for the USD key to fill. | `d6606182` |
| **R1-5** MINOR — sync refetch losing the lease to its own thread polls up to 2 s | **Fixed** — chose **join** over serve-from-disk: `_in_flight_revalidation(path)` returns this process's live thread and the sync reader `join`s it with the same `_LISTING_LEASE_WAIT_S` ceiling, so the wait ends the instant the refresh lands *or fails* (the mtime poll spun the whole window on a failure). Serving disk without waiting was rejected because the caller's memo would pin the missing id for the day — the incident this trigger exists for. Rationale in the code comment and `read_listing` docstring; two tests (successful refresh → B serves the refreshed document without fetching; failed refresh → B returns well inside the window). | `6ab10262` |
| **R1-6** NIT — `_from_aggregator_catalogue` docstring promises `radient/*` | **Fixed** — says OpenRouter only, and why (`PUBLIC_LISTING_PROVIDERS`). | `d6606182` |
| **R1-7** NIT — capture bump note omits that it lands on-path | **Fixed** — notes the one-time refetch is the invalidate-and-re-enter sync fetch on the first resolution after upgrade. | `d6606182` |

#### Evidence

Reviewer's R1-1 reproduction re-run (`/tmp/lop-cf-evidence/r1_1_repro.py`: `fetch_models` stubbed, memo cleared between "processes", document 11 min old listing only the dated snapshots):

```
--- HEAD d6606182 ---
claude-sonnet-4-5            process 1..3: status=cached listing fetches this process=0
claude-opus-4-5              process 1..3: status=cached listing fetches this process=0
claude-sonnet-4-5-20250929   process 1..3: status=cached listing fetches this process=0
--- BEFORE (0e608b0f) ---
claude-sonnet-4-5            process 1..3: status=ok     listing fetches this process=1
claude-opus-4-5              process 1..3: status=ok     listing fetches this process=1
claude-sonnet-4-5-20250929   process 1..3: status=cached listing fetches this process=0
```

R1-2/3/5 on the real paths (`/tmp/lop-cf-evidence/r1_2_5_repro.py`, real anthropic document copied to a temp dir and aged, `fetch_models` spied):

```
R1-2 resolve_model_info on a 2h document -> fetch_models calls (thread, timeout): [('catalogue-revalidate:anthropic.listing', 10.0)]
R1-3 after read_listing: old tmp exists = False | young tmp exists = True
R1-5 call B parked on A's thread (alive, no result yet): True
R1-5 A: status=cached; B: status=cached lists new id=True returned 0.005s after the refresh landed; fetch_models calls=['catalogue-revalidate:anthropic.listing']
all assertions held
```

Gates on `d6606182`, whole tree as CI spells them: flake8 clean, `black==26.1.0` 764 files unchanged, `isort==5.13.2` clean, pyright 0 errors. `tests/unit` under `env -u NO_COLOR TERM=xterm-256color`: 10604 passed, 15 skipped; the one local failure + 3 errors are `tests/unit/evaluation/runner/test_episode_subprocess.py` ("no usable copied interpreter on this host" — the uv-managed 3.14 interpreter's `libpython3.14.dylib` does not survive a `venv --copies`), untouched by this PR and green on CI. CI on `d6606182`: lint / type-check / test (3.12) / pip-audit / tui-e2e (ubuntu, macos) all pass.

damianvtran:
### Agent review — round 2

Scope: 4da01369..d6606182 (remediation delta `0e608b0f..d6606182`, 5 commits)

Gates re-run in the worktree over the whole tree: flake8 clean; `black==26.1.0` 764 files unchanged; `isort==5.13.2` clean; `pyright --pythonpath .venv/bin/python .` 0 errors / 0 warnings; `tests/unit/model tests/unit/providers tests/unit/tui/test_model_picker.py tests/unit/tui/test_app_pilot.py` under `env -u NO_COLOR TERM=xterm-256color` → **1469 passed**. CI run 33578014762 on head `d6606182`: lint / type-check / test (3.12) / pip-audit / cli-sanity / server-sanity / tui-e2e (ubuntu, macos) all pass.

#### Round-1 findings

| Finding | Verdict | Evidence |
|---|---|---|
| **R1-1** MAJOR — alias `want_id` sync-fetched on every boot | **Fixed** | `_lists_id` now intersects `id_spellings` of the row and of the wanted id after normalisation. Probed directly: `claude-opus-4-5` and `claude-opus-4.5` against a `claude-opus-4-5-20251101` document → present; `claude-fable-5-1` → absent (refetch still fires). The 6-case parametrised test plus the new-family test pin both sides. See R2-1 for the one direction the widening over-reaches. |
| **R1-2** MINOR — background thunk inherited the on-path timeout | **Fixed** | `read_listing(..., revalidate=)` is threaded through both `read_listing` call sites in `discovery._available_models` and the first in `prices._price_catalogue_row` (the re-enter call there has `soft_ttl_s=0` and never schedules, so it needs none). `revalidate()` builds with `DEFAULT_TIMEOUT_S` (10.0); tests assert the background call sees 10.0 and the sync call keeps 2.0. |
| **R1-3** MINOR — stranded `*.tmp` from a daemon kill | **Fixed** | `purge_stranded_temp_files`: glob `*.listing.json.*.tmp` matches exactly the `mkstemp(prefix=f"{path.name}.", suffix=".tmp")` shape for every key this module writes (`<x>.listing`), age gate 5 min on mtime, per-file `OSError` → continue, directory-level `OSError` → debug log. Tests cover old/young/foreign(`pypi-local-operator.json.*.tmp`)/document/missing-dir and the `read_listing` sweep. |
| **R1-4** MINOR — `_PRICE_CATALOGUE_KEYS` comment | **Fixed** | Comment now says first key wins outright and the token plan's 0/0 is intended. |
| **R1-5** MINOR — sync reader losing the lease to its own thread | **Fixed** | `_in_flight_revalidation(path)` reads `_threads` under `_revalidate_lock` (slot keyed by `str(path)`, same as `_schedule_revalidate`); the sync reader `join(timeout=_LISTING_LEASE_WAIT_S)` (2.0 s) then serves disk — same ceiling as the mtime poll, so the paint path stays bounded, and it returns the instant the thread finishes on failure too. A peer-process holder still takes `await_peer_briefly`. Both tests (refresh lands → B serves window=2 without fetching; refresh fails → B returns promptly with the stale document) pass. |
| **R1-6** NIT — `_from_aggregator_catalogue` docstring | **Fixed** | Says OpenRouter only and why (`PUBLIC_LISTING_PROVIDERS`). |
| **R1-7** NIT — capture-bump note | **Fixed** | Notes the one-time refetch is synchronous, on-path, up to `DEFAULT_TIMEOUT_S`. |

#### New findings

**R2-1 · MINOR · the symmetric spelling match counts a NOT-YET-LISTED snapshot of a listed family as present, so the `want_id` trigger no longer fires for the case it was built for** — `local_operator/model/discovery.py` `_lists_id`.
Both sides are date-stripped before intersecting, so any two snapshots of one family coincide. Probed: a document listing only `claude-opus-4-5-20251101`, `want_id="claude-opus-4-5-20260315"` → `_lists_id` returns **True** → no refetch; the lookup in `configure._info_from_discovery` (exact/normalised only) then misses, returns `fallback`, and the memo pins it for the rest of the UTC-day bucket. Same for `claude-3-5-sonnet-20241022` against a `-20240620` document. That is exactly the "a new id just landed, refetch once" scenario `lacks_want_id` exists for, and the new-family test (`claude-fable-5-1`) does not exercise it. Not a regression against pre-PR (there was no trigger at all), so minor, but the fix is one line and keeps every one of the six parametrised cases green (I checked): match one-sided — a row is a hit when the wanted id's *normalised literal* is in the row's spellings, or the row's normalised literal is in the wanted id's spellings — rather than intersecting the two derived sets. Add `("claude-opus-4-5-20251101", "claude-opus-4-5-20260315")` as a must-refetch case.

#### Checked, not findings
- `_in_flight_revalidation` cannot deadlock: the thread's `finally` takes `_revalidate_lock` only after releasing the lease, and the reader holds no lock while joining. A thread that lost the lease to a peer exits at once, so a join against it returns immediately and the reader falls into the disk read as before.
- The join path returns `Listing(payload, age_s)` with `fetched=False, failed=False` even when the background thread failed — correct: the sync caller neither fetched nor failed, and the stale document is the module's stated answer.
- `purge_stranded_temp_files` runs on every `read_listing`; one extra directory glob per call, behind the memo, and the module already pays one for the legacy purge. Negligible.
- Design doc §3.2 / §9 amendments match the code.

#### Verdict
**Clean — 0 blocker, 0 major.** All seven round-1 findings are fixed as described; R2-1 is a minor gap in the new matcher that can be fixed in this PR or deferred in the remediation comment. This round is **TERMINAL**: nothing here warrants another round.

Severity count: blocker 0 · major 0 · minor 1 · nit 0. Reviewed head: `d660618222fa806c80719fb81cb08352a484be59`.

damianvtran:

Implementer: coder subagent (anthropic/claude-fable-5-1)
Scope: d6606182..7012501e (1 commit)

| Finding | Status | Evidence |
|---|---|---|
| **R2-1** MINOR — symmetric spelling match counted an unlisted snapshot of a listed family as present | **Fixed** in `7012501e` | `_lists_id` now matches each side's normalised literal against the OTHER side's `id_spellings` instead of intersecting the two derived sets. A literal never loses its date, so two dated ids match only when they are the same id; an undated alias still matches a dated row in either direction. New parametrised test `test_an_unlisted_snapshot_of_a_listed_family_still_refetches` (`claude-opus-4-5-20251101`→`-20260315`, `claude-3-5-sonnet-20240620`→`-20241022`) failed on `d6606182` (`'cached' == 'ok'`) and passes on the new head; the six alias cases and the new-family (`claude-fable-5-1`) guard remain green. Direct probe of all eight cases (six aliases → present, two snapshot pairs → absent, new family → absent) matched expectations. Docstring rewritten to say why the match is literal-against-spellings. |

Gates over the whole tree in the worktree: `flake8` clean; `black==26.1.0` 764 files unchanged; `isort==5.13.2` clean; `pyright` 0 errors / 0 warnings; `tests/unit/model tests/unit/providers tests/unit/tui/test_model_picker.py tests/unit/tui/test_app_pilot.py` under `env -u NO_COLOR TERM=xterm-256color` → **1471 passed**. CI run 33580994789 on head `7012501e`: lint / type-check / test (3.12) / pip-audit / cli-sanity / server-sanity / tui-e2e (ubuntu, macos) all green.

damianvtran:
### Agent review — round 3

Reviewer: reviewer subagent (lopdev team), anthropic/claude-fable-5-1 — independent of the implementing coder.
Scope: 4da01369..7012501e (freshness round; delta since round 2 is the single commit 7012501e)

**R2-1 verdict — resolved.** `_lists_id` now matches one-sided: the wanted id's *literal* against the row's spellings, or the row's literal against the wanted id's spellings. Because a literal never loses its date stamp, two dated snapshots of one family no longer coincide via their shared date-stripped form. Verified beyond the parametrised test with an ad-hoc probe of ten cases: snapshot-listed/alias-wanted, alias-listed/snapshot-wanted, dotted↔dashed alias in both directions, dotted-dated vs dashed-dated of the *same* snapshot, case folding — all still hit; unlisted snapshot of a listed family (R2-1, both fixtures) and new-family ids (undated and dated) — all still miss. The `test_a_genuinely_unlisted_id_still_refetches_after_the_alias_widening` guard is untouched and passes.

Diff check: `git diff d6606182..7012501e` touches only `local_operator/model/discovery.py` (`_lists_id` body + docstring) and `tests/unit/model/test_discovery.py` (+1 parametrised test, 2 cases). Nothing else changed since round 2.

Gates on 7012501e (worktree, run as CI does): flake8 0, black 26.1.0 `--check` 764 unchanged, isort 5.13.2 `--check` clean, pyright 0 errors. Tests: `tests/unit/model` 469 passed; `tests/unit -k picker` 662 passed. CI run 33580994789 on head 7012501e: all 8 jobs pass (lint, type-check, pip-audit, test 3.12, cli-sanity, server-sanity, tui-e2e ×2).

New findings: none. No BLOCKER, no MAJOR, no MINOR, no NIT worth a round.

**Verdict: clean.** Round 3 is TERMINAL and fresh against head 7012501e5b30adcde06039a578d8f631fa3e7d76.

damianvtran:
### Design review — round 1

Reviewer: designer subagent (anthropic/claude-fable-5-1)
Scope: 4da01369..7012501e (head `7012501e5b30adcde06039a578d8f631fa3e7d76`)

Surface under review: the `/model` picker footer — `cached` now silent, new `stale` clause `live list unavailable: <providers> — showing cached`, `empty` unchanged. Judged from rendered frames only: the three PNGs in the PR body, plus six frames I captured from the real `OperatorApp` at head with `_AccessController` forcing statuses (100×30, 80×30, 60×30; scripts `/tmp/cf-design/shot.py`, `settle.py`). New frames are on `pr-catalogue-freshness-assets` under `shots/design-r1/`.

**Geometry that decides everything below.** The picker card is narrower than the terminal: 73 cells at a 100-column screen, 76 at 80, 56 at 60 (measured `picker.size.width`). The footer's usable run is `width − 2 gutter − 1 margin`, so **70 cells at 100 columns**. The access note is 28 cells (`2 hidden — /login <provider>`), the separator 3, which leaves **39 cells** for the catalogue clause. The new stale clause is **54 cells with two providers** and grows ~9 per provider.

#### D1 · MAJOR · the stale clause never fits beside the access note at the reference width, and the cut removes the half that reassures

The PR's own after-stale frame is clean only because `anthropic` is *stored* there, so no rows are hidden and no access note is emitted. Any real setup with a provider the user has not logged into (the normal case — every registry provider without a credential hides its rows) puts the note first, and at 100 columns the clause then renders as:

```
  2 hidden — /login <provider> · live list unavailable: anthropic, xai…
```

stale-note-100x30

At 80 columns it is `…anthropic, xai — showi…`, at 60 `…unavailable: an…`. The ordering is right — the note leads and survives, as the design doc requires — but what the ellipsis eats every time is `— showing cached`, which is the only part of the sentence telling the user the rows below are still usable. What survives, `live list unavailable: anthropic, xai`, reads as "these providers are down", one step from the conclusion the footer exists to prevent ("the model I want isn't offered").

The clause needs to fit **39 cells for two providers** so `note · clause` clears 70. Two spellings that do, either of which I would accept:

- `stale list: anthropic, xai` (26) — parallels the existing `no live list: ollama` label-colon-providers pattern, so the two failure clauses read as siblings rather than a sentence next to a label. The rows-are-still-usable meaning is carried by "list" (there *is* one) rather than by a trailing phrase that gets cut.
- `anthropic, xai: cached, refresh failed` (38) — providers first, so the truncation (when it comes) eats the explanation, not the names.

Whichever is chosen, keep the `— showing cached` intent out of a *trailing* clause: at this width the tail is the part that dies.

#### D2 · MINOR · the common stale case lists every connected provider, so the clause is unbounded

`stale` is "a fetch was attempted and failed". The usual cause is the machine being offline, which fails *every* provider whose document is past the 15-minute picker TTL at once — so the realistic clause is five names, not two:

```
  2 hidden — /login <provider> · live list unavailable: anthropic, google, mistral, openai, xai —…
```

many-stale-100x30

79 cells even without the note; it does not fit any picker width the app renders. Suggest collapsing when every provider that *has* a listing is stale: one clause with no names (`live lists unavailable — showing cached`, or under D1's spelling `all lists stale`), and only name providers when the failure is partial. Partial failure is the case where naming helps ("anthropic is down, the rest are fine"); total failure is the case where the names are noise.

#### D3 · MINOR · silencing `cached` makes the footer row collapse on every `/model` open

Consecutive frames on open, all providers within TTL (the case users hit most of the day now that the picker refreshes every 15 min):

| frame 1 (static paint) | frame 2 (live paint settled) |
| --- | --- |
| settle-cached-1 | settle-cached-2 |

`picker.size.height` 5 → 4: the `checking providers…` row is replaced by nothing, the card shrinks one row, and everything above it (welcome block, transcript) drops one line. Same is visible in the PR body's own pair: the logo sits at y≈308 in *before* (footer row present) and y≈336 in *after-cached*. This reflow existed before for `ok` (already silent), but with the 24h TTL most opens were `cached` and kept the row; after this PR both the ≤15-min and the just-fetched states are silent, so the shift is now the default experience, and when the live fetch takes real time (document past TTL, network round-trip) the user watches the picker lose a row mid-read. Fix options, either fine: keep the status row reserved once the picker has painted it (a blank dim row is less motion than a collapse), or have `_row_budget` reserve the status row whenever `_providers` exists. If deferred, say so in the remediation — it is a widget change, not a wording one.

#### D4 · NIT · two spellings of the same phrase two lines apart

`_refresh_catalogue`'s exception path (app.py:13949) says `live model list unavailable: {error}`; the new clause says `live list unavailable: …`. Same footer slot, same meaning, one has "model" in it. Whichever wording D1 lands on, align the exception path to the same label so a user who sees both across two sessions reads one vocabulary.

#### Checked, not findings

- **Silence on `cached` loses nothing.** `available_models` pins `soft_ttl_s = min(SOFT_TTL_S, ttl_s)` and the picker passes `ttl_s=900`, so in the picker `cached` means "document under 15 minutes old" and nothing else; a 15-minute-old document that fails to refetch surfaces as `stale`, not `cached`. There is no state the old `cached:` clause reported that the new footer hides.
- **`/login` when relevant.** The access note (`N hidden — /login <provider>`) still leads and is the only clause the user can act on; a `stale` provider cannot honestly be told from a network failure vs a 401 (discovery.py:366 keeps both as "keep the document"), so not surfacing `/login` on stale is correct rather than a gap.
- **`empty` unchanged**, and the stale+empty combination composes with the same ` · ` seam as the row count (`stale-empty-100x30.png`).
- **Contrast/colour**: the clause inherits the existing dim footer style; nothing new to judge.
- Out of scope, noted only: the picker is *narrower* at 100 columns (73) than at 80 (76) because of the editor's width rule — pre-existing, not touched here, but it is why 100 columns is the tightest of the "typical" widths for this footer.

#### Verdict

**Not clean — 0 blocker, 1 major (D1), 2 minor (D2, D3), 1 nit (D4).** D1 is a one-string change plus its test; D2 is a small branch in `_catalogue_status`; D3 and D4 may be fixed here or deferred in the remediation comment. Round 1 is not terminal; I will re-check the frames after remediation.

damianvtran:

Head: `3e70ed8f` (was `7012501e` at review). Frames re-captured from the real `OperatorApp` at head (`run_test`, `xterm-256color`, two `pilot.pause()` settles) and pushed to `pr-catalogue-freshness-assets` under `shots/design-r1-remediation/`.

**D1 (MAJOR) — fixed in `3e70ed8f`.** Adopted your first spelling: the clause is now `stale list: anthropic, xai` (26 cells), parallel to `no live list: …`, with `— showing cached` dropped entirely so the tail-truncation failure mode is gone. A new unit test pins the two-provider clause ≤ 39 cells (the run left beside the 28-cell access note at 100 columns). Rendered at the picker's real width (73/76/56 cells at 100/80/60-col terminals):

| 100 cols (59-cell footer line, no truncation) | 80 cols (identical) | 60 cols (ellipsis eats `, xai` — state word and first name survive) |
| --- | --- | --- |
| stale-note-100 | stale-note-80 | stale-note-60 |

At 60 columns the line is `2 hidden — /login <provider> · stale list: anthropic…` — the access note leads and survives, exactly the ordering you called correct.

**D2 (MINOR) — fixed in `3e70ed8f`.** When every provider that produced a listing is stale, the clause collapses to `stale list: all providers`; providers are named only on partial failure. `static` and `unauthenticated` providers never asked the network, so they do not count toward "all"; a *single* stale provider is still named, since `all providers` would say less. Rendered with five stale providers:

many-stale

Partial failure keeps names and composes with `empty` on the same seam: `stale-empty-100x30.png` shows `stale list: anthropic · no live list: ollama` (the `empty` name truncates first at 71 cells — least informative half dies, per your D1 guidance).

**D3 (MINOR) — fixed in `3e70ed8f`.** Widget-level, and cheap, so not deferred: `ModelPicker` now holds its status row once one has painted in an open (`_status_row_held`, reset on `close()`). Settled statuses render into the held row, which goes blank instead of collapsing; `_row_budget` accounts for it, so the card's height is stable open→settle. Consecutive frames with all providers within TTL — before (`7012501e`: 5→4, transcript drops one line) and after (`3e70ed8f`: 5→5, nothing above the picker moves):

| before frame 1 → frame 2 (`h=5` → `h=4`) | after frame 1 → frame 2 (`h=5` → `h=5`) |
| --- | --- |
| before-settle | after-settle |

Same stability with a settled stale clause (`pair-after-settle-stale.png`, `h=5` → `h=5`). One scoping choice to flag: the row is held only after a *status clause* has painted (the app's first frame always carries `checking providers…`, so the picker's open→settle pair is the case that mattered). A windowed-list count (`4 of 60`) appearing mid-session when the user scrolls can still grow the footer — transient and user-driven rather than a settle reflow, so left alone.

**D4 (NIT) — fixed in `3e70ed8f`.** The exception path now uses the same label via a shared constant (`STALE_LIST_LABEL = "stale list:"`): `stale list: all providers — <error>`. The reason trails because the exception took every provider down together (the collapsed spelling is accurate) and truncation may lose the reason without changing the sentence. Rendered (`exception-100x30.png`):

exception path

Also updated: the PR body's stale-evidence transcript (re-run: `footer: 'stale list: anthropic'`), the before/after caption, and both body frames re-captured at head (the after-cached frame now shows the held blank row). `docs/design/model-catalogue-freshness.md` §3.6 re-worded to match.

**Not addressed:** nothing — all four findings fixed; no deferrals.

damianvtran:
### Design review — round 2

Reviewer: designer subagent (anthropic/claude-fable-5-1)
Scope: 4da01369..3e70ed8f (head `3e70ed8f69a3202122337635dc91999576f4a8a4`)

Verified the round-1 remediation against the frames it pushed: all ten PNGs under `shots/design-r1-remediation/` on `pr-catalogue-freshness-assets` (viewed, not inferred), plus the re-captured PR-body `model-picker-after-cached.png`. The settle pairs were diffed pixel-for-pixel (panels split at the 20-px gutter, per-row diff, threshold 30/RGB), and the logo/card y-offsets measured per panel, since D3 is a motion claim and a still can hide a one-row shift.

#### Round-1 findings

**D1 (MAJOR) — fixed, verified.** `stale-note-100x30.png` and `stale-note-80x30.png` both render the whole line, `2 hidden — /login <provider> · stale list: anthropic, xai`, with no ellipsis; the access note leads. At 60 columns (`stale-note-60x30.png`, 56-cell card) the line is `2 hidden — /login <provider> · stale list: anthropic…` — the actionable note is intact, the state word and the first provider survive, and the cut lands on a name rather than on the meaning. That is the acceptable degradation asked for. `stale list:` beside `no live list:` (`stale-empty-100x30.png`) reads as two siblings of one vocabulary.

**D2 (MINOR) — fixed, verified.** `many-stale-100x30.png` with five failed providers collapses to `2 hidden — /login <provider> · stale list: all providers` — fits with room at 100 columns, no names to overflow. Partial failure still names providers (`stale-only-100x30.png`: `stale list: anthropic, xai` with the rest ok). The single-stale case is pinned by the `len(stale) > 1` guard in `_catalogue_status` and by the re-run transcript in the PR body (`footer: 'stale list: anthropic'`); no dedicated frame, but the wording is the two-name frame minus one name, so I am not asking for one.

**D3 (MINOR) — fixed, verified by pixel diff.** Numbers from the three pairs (left = frame 1 open, right = frame 2 settled; 1000×632 px per panel, ~20 px per terminal row):

| pair | rows that differ | logo y (L → R) | card top y (L → R) |
| --- | --- | --- | --- |
| `pair-before-settle-cached` (`7012501e`) | **18 spans**, every text row from y=56 down | 56 → **76** | 330 → **350** |
| `pair-after-settle-cached` (`3e70ed8f`) | **1 span**, y=431–446 (the status row only) | 56 → 56 | 330 → 330 |
| `pair-after-settle-stale` (`3e70ed8f`) | **1 span**, y=431–446 (the status row only) | 56 → 56 | 330 → 330 |

Before: the whole screen above the picker moved one row on settle. After: the only pixels that change on settle are inside the footer row itself (`checking providers…` → blank, or → `stale list: all providers`). Nothing above the picker moves; the card height is stable. This is exactly the fix asked for. Reset-on-`close()` means each open starts from `checking providers…` again, so the hold cannot leak a stale-blank row into a later open.

**D4 (NIT) — fixed, verified.** `exception-100x30.png` renders `stale list: all providers — provider timeout`. Same label as the status path via one constant; the reason trails and is what truncation would lose. One vocabulary.

#### The held blank row in the common cached case (asked explicitly)

Judged from `pair-after-settle-cached.png` frame 2 and the re-captured `model-picker-after-cached.png`: the footer settles to a blank dim row between the last model row and `d in /model saves this for new sessions`. **Not a finding.** It reads as a spacer separating the list from the hint, not as a missing line — the hint row is left-aligned in the same dim style, so the eye sees "list · gap · instruction", which is a normal rhythm for a footer. The alternative it replaced (the card shrinking one row while the user is reading it) is strictly worse. Two things I checked so this stays true:

- It costs no list row compared with what users had before this PR: with the 24h TTL most opens settled as `cached`, which kept a text row in the same slot, so `_row_budget` reserved this row already.
- It is consistent across states: cached → blank, stale → `stale list: …`, exception → `stale list: all providers — …`, and the `no-providers` picker (no `checking providers…` first frame) never holds a row, so it does not get a phantom blank.

If a later pass wants the slot to earn its keep, a right-aligned `listed 3m ago` in the held row would be the natural occupant — follow-up material, not a defect.

#### New findings

**D5 · MINOR · `no live list:…` with the name cut off reads as a rendering glitch.** `stale-empty-100x30.png` (access note + stale + empty, the three-clause worst case at 100 columns) renders `2 hidden — /login <provider> · stale list: anthropic · no live list:…` — the `empty` label survives but its only payload, the provider name, is the part that dies, leaving a dangling colon. The ordering (note > stale > empty) is right and this is a rare composite; but a label with no value is worse than no label. Suggested follow-up: truncate at the ` · ` seam when a trailing clause cannot keep its first value (drop `no live list:…` entirely), or shorten `empty` to `no list: ollama` (17 cells for one name) so the seam fits at 70. **Deferrable — record as a follow-up in the remediation; not blocking.**

No further findings. I looked for and did not find: a colour/contrast change (all clauses inherit the existing dim footer style), a focus or cursor change (the composer cursor stays on the `/model` line in every frame), or a case where the held row pushes the picker past the composer band (card bottom is pinned at the same y across all pairs).

#### Verdict

**Clean — 0 blocker, 0 major, 1 minor (D5, new, deferrable), 0 nit.** D1–D4 are all verified fixed at head `3e70ed8f` from the rendered frames. This round is **TERMINAL**; D5 needs only a one-line `deferred — …` (or a fix) in the round-2 remediation and does not require another design round.

damianvtran:
### Agent review — round 4

Reviewer: reviewer subagent (lopdev team), anthropic/claude-fable-5-1 — independent of the implementing coder.
Scope: 4da01369..3e70ed8f (freshness round after the design remediation; delta since the clean round 3 is the single commit `3e70ed8f`, `git diff 7012501e..3e70ed8f`)

**Delta reviewed.** Four files: `local_operator/tui/app.py` (`STALE_LIST_LABEL`, `LISTED_STATUSES`, `_catalogue_status` collapse, exception-path label), `local_operator/tui/widgets/model_picker.py` (`_status_row_held` on `__init__`/`close()`/`_footer_rows`, `_row_budget`), `tests/unit/tui/test_model_picker.py` (+3 tests, 3 string updates), `docs/design/model-catalogue-freshness.md`. No other file changed since round 3.

**Held-row geometry — verified, not inferred.** Probed a mounted `ModelPicker` under `run_test` at screen heights 12/15/30/45/60 with a 60-row catalogue, driving the app's real sequence without closing: `checking providers… · <hint>` → `<hint>` only → `""` → `stale list: anthropic`. Rendered card height is identical across all four states at every size (4/5/10/14/14 rows) and never exceeds `min(MAX_VISIBLE_ROWS, h // 3)`; `_row_budget` gives back the row when the held blank replaces the ordinary clause (budget 2→3 at h=12, 8→9 at h=30, 12→13 at h=45), so total rows stay pinned. `close()` + `set_rows("")` + `open()` → `_status_row_held=False`, footer 0 rows, budget back to the full cap — the reset holds across reopen. A held row with a 3-row catalogue paints 3 + blank = 4 (no phantom growth). Only exception is h=3 (cap 1), where the `max(1, …)` floor already lets the persistent hint overflow the cap pre-PR — unchanged by this diff and never reachable with the app's always-present hint.

**Keyboard/scroll/mouse paths.** `_scroll_to_selection`, `visible_window`, `_index_at` all read the same `_row_budget`, so PageDown step, window clamp and footer-click rejection stay consistent with what is painted. `tests/unit/tui/test_model_picker.py` + `tests/unit/tui/test_app_pilot.py` under `env -u NO_COLOR TERM=xterm-256color`: **320 passed**.

**`LISTED_STATUSES` semantic.** `listed` counts `ok|cached|stale|empty`; the collapse fires only when `len(stale) > 1 and len(stale) == listed`. Traced against the docstring: static+unauthenticated excluded ✓, single stale always named ✓, stale+empty is partial and keeps names ✓, any `ok`/`cached` sibling keeps names ✓. `test_every_provider_stale_collapses_to_a_nameless_clause` covers each branch. Exception path (`live_catalogue` raising) uses the same `STALE_LIST_LABEL` constant via the collapsed spelling — one vocabulary, as D4 asked.

**CI flake — cannot be this PR.** Run 33587082704 attempt 1 failed `test (3.12)` only on `tests/unit/session/test_resume_subagents_todos.py::test_redundant_roster_events_skip_the_sidecar_write` (`assert 2 == 1`, sidecar write count vs baseline); 3.13 passed and attempt 2 is green across all 9 jobs on head `3e70ed8f`. The PR's full file list (`git diff 4da01369..3e70ed8f --name-only`) touches `local_operator/model/*`, `local_operator/providers/controller.py`, `local_operator/tui/{app.py,widgets/model_picker.py}`, `pyproject.toml`, docs and their tests — **no file under `local_operator/session/`, `local_operator/harness/`, or `tests/unit/session/`**. The only import edge from `session.py` into changed code is the lazy `build_model_spec` for a configured subagent model tier (session.py:5094), which this test never reaches (empty tmp config, `ModelSpec(provider="test")`). The failure is a write-count race in the roster writer drain, independent of the catalogue. Run locally 5× serial (`-n0`): **5/5 passed**.

**Gates on `3e70ed8f`** (worktree, whole tree, as CI spells them): flake8 0; `black==26.1.0 --check` 764 files unchanged; `isort==5.13.2 --check` clean; pyright 0 errors / 0 warnings.

#### Findings

**R4-1 · MINOR · the `_row_budget` held-row branch has no automated coverage** — `model_picker.py:687`. `test_the_status_row_is_kept_once_painted_so_the_card_does_not_reflow` runs an unmounted picker, so `_row_budget` raises `NoScreen` and short-circuits to `MAX_VISIBLE_ROWS`; the `self._status_row_held` term that stops the blank row from pushing the composer is exercised only by the designer's manual frames and my probe above. One `run_test`-mounted assertion (`len(render_text().split("\n"))` equal across `checking…`→settled and `<= min(MAX_VISIBLE_ROWS, h // 3)`) would pin it. Follow-up, not a round.

**R4-2 · NIT · `LISTED_STATUSES` comment under-describes `static`** — `app.py` (`LISTED_STATUSES` doc + `_catalogue_status` inline comment) says static/unauthenticated "never asked the network". `discovery.py:1222,1391` also returns `static` for a provider whose fetch *failed with no stored document*. Behaviour is still right (such a provider excluded from `listed` only makes the offline collapse fire sooner, which is the correct direction), but the comment should say "produced no listing" rather than "never asked".

**R4-3 · NIT · render helper with a side effect** — `_footer_rows` (`model_picker.py:651`) flips `_status_row_held`, and it is reachable via `render_text` as well as `_repaint`. Harmless today (the unit tests call `render_text` after `open`), but a sentence in the attribute comment that "set on paint, including `render_text`" would spare the next reader a hunt.

No BLOCKER, no MAJOR.

**Verdict: clean.** Round 4 is TERMINAL and fresh against head `3e70ed8f69a3202122337635dc91999576f4a8a4`; R4-1..R4-3 are follow-ups, not a reason for another round.

damianvtran:
### Design review remediation — round 2

Implementer: coder subagent (anthropic/claude-fable-5-1)
Head: `f9033456` (was `3e70ed8f` at review)

| Finding | Disposition | Evidence |
|---|---|---|
| **D5** MINOR — `no live list:…` with the name cut off reads as a glitch | **Fixed** in `f9033456` | Took the first suggested option: `_fit_clauses` (model_picker.py) drops a trailing footer clause whole when its label and first value cannot fit, cutting at the ` · ` seam; whatever survives takes the same cell cut as before. The three-clause composite at 100 columns now renders `2 hidden — /login <provider> · stale list: anthropic` (77-cell status, 70-cell slot → `no live list: ollama` dropped at the seam). Unit test `test_a_footer_clause_that_cannot_keep_its_first_value_is_dropped_at_the_seam` pins the 100-col seam cut, the untouched wide case, the in-list cut and the never-dropped leading clause. |

**Already-captured frames unchanged.** Re-ran the designer's `shot.py` at head for every round-1/round-1-remediation scenario and diffed the SVG text runs: `stale-note-100x30`, `stale-note-80x30`, `stale-note-60x30` (`… · stale list: anthropic…` — the in-list cut D1 accepted), `many-stale-100x30`, `stale-only-100x30` are byte-identical in footer text to the round-1-remediation frames. An exhaustive probe of 3 widths × 3 count clauses × 12 status strings (108 lines) confirmed the new rule only changes lines where plain truncation would have cut a label from its first value.

Re-captured D5 frame (100×30) plus the same scenario at 80 and 60:

stale-empty-100x30

| 80×30 | 60×30 |
|---|---|
| stale-empty-80x30 | stale-empty-60x30 |

`stale-note-60x30` re-captured at head for the in-list-cut control: `https://raw.githubusercontent.com/damianvtran/local-operator/pr-catalogue-freshness-assets/shots/design-r2-remediation/stale-note-60x30.png` — identical to round 1.

damianvtran:

Implementer: coder subagent (anthropic/claude-fable-5-1)
Head: `f9033456` (was `3e70ed8f` at review)

| Finding | Disposition | Evidence |
|---|---|---|
| **R4-1** MINOR — `_row_budget` held-row branch has no automated coverage | **Fixed** in `f9033456` | `test_the_held_row_is_paid_for_out_of_the_row_budget_when_mounted` (parametrised at heights 12/30/45, 60-row catalogue, `run_test` at 100 cols) drives `checking… · hint` → `hint` → `""` → `stale list: anthropic` without closing and asserts the rendered card height is identical across all four states, `<= min(MAX_VISIBLE_ROWS, h // 3)`, and `_row_budget()` is `cap-2, cap-1, cap-1` for the three settled states. Proved it guards the branch: with the `or self._status_row_held` term removed from `_row_budget`, all three cases fail (`[4, 5, 4] == [4, 4, 4]`, `[10, 11, 10]`, `[14, 15, 14]` — the one-line growth on settle). |
| **R4-2** NIT — `LISTED_STATUSES` comment under-describes `static` | **Fixed** in `f9033456` | Both the constant's doc and the `_catalogue_status` inline comment now say `static`/`unauthenticated` "produced no listing", with the constant's doc naming the failed-fetch-with-nothing-stored case and why excluding it errs in the right direction. |
| **R4-3** NIT — `_footer_rows` flips `_status_row_held` from a render path | **Fixed** in `f9033456` | The attribute comment now says it is set on paint by `_footer_rows`, which `render_text` reaches as well as `_repaint`, and that only `close()` releases it. |

Also in this commit: D5 from design round 2 (seam-aware footer truncation, `_fit_clauses`) — see the design remediation comment above. Contained to `model_picker.py` plus one unit test; the exhaustive width/status probe there shows no already-captured frame changes.

Gates on `f9033456` (worktree, whole tree, as CI spells them): flake8 0; `black==26.1.0 --check` 764 unchanged; `isort==5.13.2 --check` clean; pyright 0 errors / 0 warnings; `tests/unit/model tests/unit/providers tests/unit/tui` under `env -u NO_COLOR TERM=xterm-256color` → **5506 passed, 12 skipped**. CI dispatched on the new head: run 33593635542.

damianvtran:
CI on head `f9033456`: run 33593635542 attempt 2 green across all 9 jobs (lint / type-check / test 3.12 + 3.13 / pip-audit / cli-sanity / server-sanity / tui-e2e ubuntu + macos). Attempt 1's `test (3.12)` leg alone was cancelled by the 40-min job backstop after stalling at 99% of the suite (`..s.s.s.s..s.s.s...s…` at 05:43, no output until the cancel at 05:50) — the other 8 jobs passed on the first attempt and the same suite passed locally (5506 passed under `tests/unit/{model,providers,tui}`); the re-run of that one leg passed on the same head. This diff touches no session/harness code and adds only picker-widget tests, so it is the pre-existing 3.12 flake noted in review round 4, not this PR.

damianvtran:
### Design review — round 3

Reviewer: designer subagent (lopdev team, anthropic/claude-fable-5-1) — independent of the implementing coder.
Scope: 4da01369..f9033456 (head `f9033456cb6191e5989cb65e1db4d037c4d606f6`; freshness round — delta since the clean design round 2 on `3e70ed8f` is the single commit `f9033456`, whose only user-visible change is the `_fit_clauses` seam-drop rule in `model_picker.py`)

Judged from the four re-captured PNGs under `shots/design-r2-remediation/` on `pr-catalogue-freshness-assets` (`7a3dc05`): `stale-empty-{100,80,60}x30` and `stale-note-60x30` — viewed, not inferred. Backed each frame with `_fit_clauses` at head over the measured card widths (73/76/56 cells at 100/80/60 columns → 70/73/53-cell footer runs), so the text I read in the pixels is the text the code produces.

#### D5 (MINOR, round 2) — fixed, verified

The three-clause composite (access note · stale · empty) is 75 cells joined; it cannot fit any picker width. At head:

| frame | run | rendered footer | cells |
| --- | --- | --- | --- |
| `stale-empty-100x30` | 70 | `2 hidden — /login <provider> · stale list: anthropic` | 52 |
| `stale-empty-80x30` | 73 | `2 hidden — /login <provider> · stale list: anthropic` | 52 |
| `stale-empty-60x30` | 53 | `2 hidden — /login <provider> · stale list: anthropic` | 52 |

- **No dangling label.** `no live list:` is gone from all three frames; there is no `no live list:…` and no bare colon anywhere in the footer.
- **No trailing separator.** Each line ends on `anthropic` — the ` · ` seam before the dropped clause went with it, and there is no orphaned `·` or `…` after the surviving clause. The row is then padded to the card width in the same dim tint, so the footer reads as one complete sentence, not a cut one.
- **Ordering preserved.** Note leads, `stale list:` second; what was dropped is the lowest-priority clause, exactly the "drop at the seam" option round 2 offered.
- **Nothing else moved.** Card height, the `d in /model saves this for new sessions` row, the composer band, the selection highlight and the `200k  $3/15` columns are in the same positions as the round-2 frames; only the footer text changed.

#### 60-column control — still correct

`stale-note-60x30` (note · `stale list: anthropic, xai`, 57 cells into a 53-cell run) still renders `2 hidden — /login <provider> · stale list: anthropic…` — the in-list cut D1 accepted. The seam rule did not touch it because the trailing clause's head (`stale list: anthropic`) fits with its ellipsis, so the line degrades inside the provider list, not at the label. Both 60-col frames lead with the actionable access note and the state word `stale list:`; the `…` in the two-name case lands on a provider name, not on meaning. Identical to the round-1/round-2 frame, as the remediation claimed.

#### New findings

**D6 · NIT · a dropped clause leaves no trace.** At 100 columns the user with an empty ollama listing sees `… · stale list: anthropic` and nothing tells them a third clause existed. This is the trade round 2 asked for (a clean drop beats a dangling label, and the empty-listing clause is the least actionable of the three), and any marker (`stale list: anthropic ·…`) would reintroduce the glitch look, so I am not asking for a change. Recording it so a later pass that widens the picker card relative to the terminal — the card is 73 cells on a 100-column screen, the real ceiling here — knows what it buys back. **Follow-up, not a round.**

No further findings. I looked for and did not find: a colour/contrast change (the surviving clauses inherit the same dim footer style), a focus or cursor change (the composer cursor stays on the `/model` line in every frame), a card-height or reflow difference between the three widths beyond the card's own width, or a case where the seam rule dropped the *leading* clause (the `len(kept) > 1` guard; probed `[note]` alone at 20 cells → in-clause cut, never empty).

#### Verdict

**Clean — 0 blocker, 0 major, 0 minor, 1 nit (D6, follow-up).** D5 is fixed at head `f9033456` and verified from the rendered frames; D1–D4 unchanged from round 2. This round is **TERMINAL** and fresh against head; no further design round is needed for this PR.

damianvtran:
### Agent review — round 5

Reviewer: reviewer subagent (lopdev team), anthropic/claude-fable-5-1 — independent of the implementing coder.
Scope: 4da01369..f9033456 (freshness round; delta since the clean agent round 4 / design round 2 on `3e70ed8f` is the single commit `f9033456`, `git diff 3e70ed8f..f9033456` — 3 files, +139/−10)

**Delta reviewed.** `model_picker.py`: `_SEAM`/`_FIRST_VALUE_END` constants, `_fit_clauses`, `_footer_rows` now fits the ordinary run through it and truncates the persistent row separately, `_status_row_held` comment. `app.py`: comment-only (`LISTED_STATUSES`, `_catalogue_status`). `test_model_picker.py`: seam-drop unit test + mounted `_row_budget` test (3 heights).

**`_fit_clauses` — verified by exhaustive probe, not read.** Ran the function at every width from 0 to full-run+2 for: the exception path `note · stale list: all providers — HTTPSConnectionPool(host='x', port=443): Max retries, foo`; the three-clause composite `note · stale list: anthropic · no live list: radient`; `[note]` alone; `["3 of 60"]` alone; `3 of 60 · stale list: anthropic, xai`. Every output is `<= width` cells (no off-by-one at either side of the boundary: at 57 the exception line keeps `stale list: all providers…` exactly at 57 cells; at 56 it drops to the note; at 52/76 the composite fits whole/keeps two clauses exactly).
- **Leading clause never dropped** — `len(kept) > 1` guard; `[note]` at 1–5 cells gives `…`, `2…`, `2 h…` (identical to plain `truncate_cells`), never empty.
- **`_FIRST_VALUE_END` on the exception clause** — `re` takes the leftmost match, so `stale list: all providers — <error with commas>` splits at ` — ` (head `stale list: all providers`), not at a comma inside the error. Verified: from 57 cells up the line is byte-identical to the plain cut; the ` — ` in the head never gets its own dangling render because the head+ellipsis check charges the ellipsis.
- **Interaction with the existing cell cut** — every line that fitted before renders identically (`SAME` in the probe for all in-list cuts: `stale list: anthropic…`, `anthropic,…`, `anthropic, x…`); the seam rule only ever removes whole trailing clauses. The 56-cell design-round-1 frame (`… · stale list: anthropic…`) is pinned by the new test.
- **Persistent row** — still goes through `truncate_cells(persistent, available)` on its own row; `_fit_clauses` never sees it (`status = [part for part in status if part != persistent]`), matching the docstring's "never called with the persistent hint".

**Mounted `_row_budget` test actually mounts.** It asserts `budgets == [cap-2, cap-1, cap-1]` with `cap = min(14, h//3)` — at h=12 that is `[2, 3, 3]`; a `NoScreen` short-circuit would return 14 and fail the assertion, so the test cannot pass without a live screen. The remediation's mutation check (dropping `or self._status_row_held` → `[4, 5, 4]` etc.) confirms it guards the branch.

**Gates on `f9033456` (worktree, whole tree, as AGENTS.md spells them):** `python -m flake8 .` rc 0 · `uvx --from black==26.1.0 black --check .` 764 unchanged · `uvx isort==5.13.2 --check .` clean · `python -m pyright --pythonpath .venv/bin/python .` 0 errors/0 warnings. `tests/unit/tui/test_model_picker.py` + `test_app_pilot.py` under `env -u NO_COLOR TERM=xterm-256color`: **324 passed** (320 in round 4 + 1 seam test + 3 parametrised mounted cases). CI run 33593635542 attempt 2 on `f9033456`: `completed`/`success`.

**Prior rounds.** Round-4 remediation answers R4-1/R4-2/R4-3 (all fixed in `f9033456`, verified in the diff). Design round 3 is clean and terminal on this same head.

#### Findings

- **R5-1 · NIT · informational, no change asked.** When the row count leads (`N of M`) and the access note is the *trailing* clause, `_FIRST_VALUE_END` treats `2 hidden` as the head, so below 19 available cells (picker < 22 cells wide) the note is dropped in favour of `3 of 60` where the plain cut would have shown `3 of 60 · 2 hidd…`. No picker the app lays out is that narrow (`_NUMBERS_MIN_WIDTH` already strips the metadata columns below 56 cells), and the drop is the rule working as documented; recording it so a future narrower layout knows the note is the first casualty.

0 BLOCKER · 0 MAJOR · 0 MINOR · 1 NIT.

#### Verdict

**Clean.** Fresh against head `f9033456`; CI green; gates green. This round is **TERMINAL** — R5-1 is a follow-up note, not a reason for another round. Both required rounds (agent 5, design 3) are clean on the current head; the PR is merge-ready from the review side.

damianvtran:
### Agent review — round 6

Reviewer: reviewer subagent (lopdev team), anthropic/claude-fable-5-1
Scope: 1ce6eb22..3b13c7d7
Type: freshness round after rebase onto origin/main (1ce6eb22) — confirms no semantic drift vs round 5 (agent) / round 3 (design), both clean on f9033456.

**Verification**

1. `git diff f9033456 3b13c7d7 -- local_operator/model local_operator/tui/widgets/model_picker.py local_operator/providers/controller.py tests/unit/model tests/unit/tui/test_model_picker.py docs/design/model-catalogue-freshness.md` → empty. Core changes byte-identical to the last clean round.
2. `git range-diff 4da01369..f9033456 1ce6eb22..3b13c7d7` → all 10 commits `=` except commit 2 (version bump), whose only hunk change is `0.44.25→0.44.26` becoming `0.44.29→0.44.30`.
3. `pyproject.toml`: `version = "0.44.30"`; the PR's only change to the file is that one line. Main's current version is 0.44.29, so the bump is correct.
4. `local_operator/tui/app.py` merge coherence: main's only app.py change since the PR base was #526 (two import re-homes at ~L2426 and ~L10100, `local_operator.session.runtime.*`), which does not touch `_refresh_catalogue`, `_catalogue_status`, or the new module constants. The PR's app.py diff is byte-identical before and after rebase (`diff` of the two patches → identical), and both #526 import changes are present on head. Read the merged `_catalogue_status`: `stale`/`empty`/`listed` computation and the all-providers collapse are as reviewed in round 5.

**Execution**

- `env -u NO_COLOR TERM=xterm-256color .venv/bin/python -m pytest tests/unit/model tests/unit/providers tests/unit/tui/test_model_picker.py tests/unit/tui/test_app_pilot.py -q` → **1478 passed** in 43s.
- `.venv/bin/python -m pyright --pythonpath .venv/bin/python .` → 0 errors, 0 warnings.
- CI on 3b13c7d7: all 8 jobs green (lint, type-check, pip-audit, test 3.12, tui-e2e ×2, cli-sanity, server-sanity).

**Findings**

None. BLOCKER 0 · MAJOR 0 · MINOR 0 · NIT 0.

**Verdict: clean.** Round is TERMINAL — the rebase introduced no semantic change; rounds 5 (agent) and 3 (design) remain valid on head 3b13c7d7. Ready to merge.

damianvtran:
### Agent review — round 7

Reviewer: reviewer subagent (lopdev team), model anthropic/claude-fable-5-1
Scope: f82d5527..89fc7333
Head reviewed: 89fc7333a8bffe0eff7dc485e6136f8f68c7387c (rebased onto origin/main f82d5527 after #531/#538; version 0.44.31→0.44.32)

Freshness-only round: no code changed in the PR since round 6 (agent) / round 3 (design) cleared 3b13c7d7. Verified rather than assumed:

1. **PR-touched paths identical to the previously reviewed head.** `git diff 3b13c7d7 89fc7333 -- local_operator/model local_operator/tui/app.py local_operator/tui/widgets/model_picker.py local_operator/providers/controller.py tests/unit/model tests/unit/tui/test_model_picker.py tests/unit/tui/test_app_pilot.py docs/design/model-catalogue-freshness.md` → empty.
2. **Rebase is a pure replay.** `git range-diff 1ce6eb22..3b13c7d7 f82d5527..89fc7333` → 10 commits, all `=` except commit 2 (`!`), whose only hunk is the version line 0.44.29→0.44.30 becoming 0.44.31→0.44.32.
3. **pyproject.toml vs origin/main** → single hunk, `version = "0.44.31"` → `"0.44.32"`. Merge-base of the PR head is exactly origin/main (f82d5527).
4. **Overlap with main since 1ce6eb22** — `comm -12` of files changed on main (1ce6eb22..f82d5527) vs files changed in the PR (f82d5527..89fc7333) → `pyproject.toml` only. No PR-touched source/test/doc file was modified on main.

Local execution on head:
- `env -u NO_COLOR TERM=xterm-256color .venv/bin/python -m pytest tests/unit/model tests/unit/providers tests/unit/tui/test_model_picker.py tests/unit/tui/test_app_pilot.py -q` → **1478 passed**, 0 failed (40s).
- pyright: the worktree `.venv` reports 4 `reportMissingImports` errors, all `boto3`/`botocore.stub` in `tests/unit/evaluation/adapters/osworld/test_aws_provider.py` and `test_build_and_scripts.py`. Confirmed this is a venv dependency gap and not this PR: `.venv/bin/python -c "import boto3"` → `ModuleNotFoundError` in the worktree venv; running pyright with `~/local-operator/.venv` (which has boto3 1.43.86) gives **0 errors** on both 89fc7333 and a clean origin/main checkout. Those files arrived via #531; the PR does not touch `tests/unit/evaluation`. CI `type-check` is green.

CI on 89fc7333 (run 33629596732), final state — all 8 checks pass: lint, type-check, pip-audit, tui-e2e (macos, ubuntu), test (3.12) 33m46s, cli-sanity, server-sanity.

Findings: none. No BLOCKER, no MAJOR, no MINOR, no NIT — nothing in the delta to review beyond the version bump, and the previously reviewed diff is byte-identical.

**Verdict: clean.** This round is TERMINAL; agent round 6 and design round 3 findings carry over unchanged and no further round is needed unless the head moves again.

PR #528, opened 2026-09-02
fix(session): keep a busy-steer peer message as its own row

damianvtran:
## Summary of Changes

A peer message sent with `now=True` (the `send` tool / `lop send` steer mode) that lands while the receiving session is **busy** no longer shows twice in the TUI. Before, the transcript painted the proper `PeerMessageBlock` (`↔ peer message from "<name>" (pid N, model)`) and then a second copy of the same text as a plain `UserBlock` directly under it. Now only the `PeerMessageBlock` appears.

Bug fix; patch bump `0.44.26 → 0.44.27` (rebased onto #526, which released 0.44.26).

## Diagnosis

`Session.receive_peer_message`, `mode == "steer"` while `_is_streaming`:

```python
self.steer(str(message.details["body"]), message_id=message.id)
await self._emit_peer_receipt(message, sender)
```

`steer()` mints a **plain user `Message`** from the body and queues it. When `_drain_steering` takes it at the next tool boundary it does two things, both wrong for a peer message:

1. **Persists that plain user row.** The peer `CustomMessage` — with `sender`, `body`, and the `<peer-session-message …>` envelope in `details["text"]` — was never written on this path, so the transcript lost provenance, the model read the bare body as the user's own words, and a `/resume` repainted it as a user prompt.
2. **Announces it with a user `MessageStartEvent`** (`isinstance(message, Message) and message.role == "user"`). `tui/events.py` turns that into `UserMessageStart`; `app.on_user_message_start` finds no echo it registered for that id and paints a `UserBlock`. The `PeerMessageDeliveredEvent` fired by `_emit_peer_receipt` had already painted the `PeerMessageBlock`. Two copies.

Every other branch (mailbox record-only, mailbox+wake, idle steer) persists/delivers the `CustomMessage` itself, and both `Transcript.build_llm_history` and `Session._default_convert_to_llm` already render `PEER_MESSAGE_MESSAGE_TYPE` as a user turn carrying the envelope.

## The Fix

Queue the peer `CustomMessage` itself on the busy-steer branch — the same shape the busy wake path (`_steering_queue.put_nowait(wake_message)`) and the busy resume-catchup path already use:

```python
self._steering_queue.put_nowait(message)
self.refresh_frontend_state()
await self._emit_peer_receipt(message, sender)
```

- `_drain_steering` persists whatever it takes, so the row lands **once**, as the peer `CustomMessage` with its provenance.
[...]

PR #530, opened 2026-09-02
feat(runtime): run-to-completion residency, viewer-held warmth, and the wake index

damianvtran:
## Summary

**PR 2 of the detached-session series** (PR 1 was #526, the runtime package move). This PR changes **runtime residency only**: the on-demand session runtime (`local_operator/session/runtime/process.py`, what the mobile daemon spawns) now runs its trajectory to completion and exits when idle — unless something is about to need it. The TUI still owns its session in-process; that cutover is PR 5.

Three things land, plus one seam:

1. **Run-to-completion with viewer-held warmth (§6.1).** The exit predicate goes from one term to three: exit after a sustained 3 s drain when (a) `handle.is_busy()` is False, (b) no wake is due within `WARM_WINDOW_S = 90 s` from the runtime's own scheduler, and (c) **no interactive viewer is attached** — an `attach`-class client on the control socket. `daemon`-class clients (the mobile daemon's adoption dial, `lop send`, `lop stop`, the future supervisor) deliberately do **not** hold it.
2. **The wake index (§4.3).** A new stdlib-only package `local_operator/wakes/` with `store.py`: one file per wake-carrying session at `<config_dir>/wakes/<session_id>.json`, absent ⇒ no wakes. Derived from the transcript, never authoritative, self-healing.
3. **The install-on-demand hook (§4.2a)** as a no-op stub: `wakes/install.py: ensure_supervisor_installed(config_dir)`, called from the one writer of schedule state. PR 7 fills it in; the chokepoint exists before the thing it installs.
4. **`ProjectionSink` injectable (§3.4).** `RuntimeServer` no longer builds a `ProjectionFold` unconditionally: it takes an optional sink, and with none injected builds the fold lazily on the first **daemon** dial. A headless runtime serving only follower terminals — or, later, firing a wake — constructs no fold.

### What this enables

A runtime that costs nothing when nothing needs it. Under the current daemon path an idle phone-started session already self-reaped after 3 s; now it also stays warm while an interactive viewer is looking (so a conversation never pays a cold start between messages) and while a wake is imminent (so a 1-minute recurring wake never thrashes exit → spawn → exit). The wake index is what a stdlib supervisor (PR 7) and the TUI's picker (PR 8) will read to answer "which cold sessions have wakes, and when?" without opening a single transcript.

### The predicate's three terms, and reconciling term 3 with "watchers do not own"

`process.py`'s docstring used to say *"Watchers and replicas observe work; they do not own it"* and excluded viewer count from the exit decision entirely. That rule is still true and still enforced — but it is about **ownership** of the work, and term 3 is about **readiness**. The reconciliation is written into `_should_exit`'s docstring:

- Term 1 (`is_busy`) is checked first and alone decides whether a turn continues. A viewer leaving never aborts a turn; a viewer arriving never starts one.
- Term 3 only decides whether an *idle* runtime stays resident. An attached interactive viewer is the one signal that the next message is imminent, so residency follows it. This turns "every message after a 3 s pause costs a ~1.2 s cold start" into "the first message of a conversation costs one".
- The daemon's connection is not the user's attention. The daemon adopts **every** session on the machine; if its dial held runtimes warm, nothing would ever exit. That is why only `ClientKind == "attach"` counts, why `phone_watchers` stays out of the predicate, and why `lop send` (which dials daemon-class precisely so it perturbs no attach accounting) cannot pin a session.

`RuntimeServer.attach_clients()` already computed exactly this count for the attach cap; the predicate reads it through `getattr` so reduced handles keep working.

### Why 90 s

`WARM_WINDOW_S` must exceed `MIN_WAKE_INTERVAL_MS` (60 s) or a session with the tightest allowed recurrence exits after every fire and cold-starts a minute later, forever. It is deliberately **not** env-tunable: it pairs with a constant in the wake layer, and a knob would let the two drift. A unit test pins the relationship.

### The index is derived, and heals

The transcript's `wake_schedules` custom entry stays the source of truth. `Session._persist_wake_schedules` is the **one** writer of schedule state, and its order is the contract:

1. transcript append (the only step allowed to fail the coroutine — the scheduler awaits it before re-arming, so memory never diverges from disk);
2. index write, wrapped so a failure is a log line and nothing more;
3. install hook if the list is non-empty, wrapped the same way.

Every session also **rewrites its entry on open** from what `_load_wake_schedules` adopted, so a deleted/corrupt/stale file is repaired the next time the session is built and an empty schedule list removes it. Unknown keys survive a rewrite (`stopped_at` is preserved on a persist and **cleared on open** — opening a session is what un-stops it; the skip logic itself is PR 7).

### Why the index lives outside the session directory

The junk-session reap deletes whole session directories, and a supervisor ticking every minute must not walk 4,000 of them to find the twelve with wakes. A flat `wakes/` directory holds exactly one file per session with something scheduled, so the scan is O(sessions-with-wakes) and the directory is empty on the typical machine.

### Why `wakes/store.py` is stdlib-only

Its readers are the wake supervisor — a ~40 MB always-on process whose whole justification is that it does **not** load the harness — and the TUI's picker at boot, before any session exists. `WakeSchedule` is a `TYPE_CHECKING` import; rows are handled as the `model_dump()` dicts the transcript already stores. `tests/unit/test_import_graph.py` pins it in a fresh interpreter: no `asyncio`, no `pydantic`, no `local_operator.session`/`harness`/`mobile`/`tui`; measured at 128 modules total, three of them ours.

### The projection seam keeps the daemon's bytes identical

Before a fold exists, welcomes and repaints serialize `handle.session_projection_seed` directly. That is safe because the seed **is** the object the host's own fold mutates (the owned handle and the TUI mirror both fold into the same `SessionProjection`), so the wire bytes are the same either way. The daemon's first dial triggers `_ensure_projection_sink()`; a redial reuses it; `set_pending` and the `fold` property build on demand so the existing reduced-host bridge (`tests/unit/session/test_remote.py`) is unchanged. `projection_sinks_built` is a counter and an INFO log line so "a headless runtime paid for no fold" is observable rather than asserted.

## What this PR does not do

No supervisor, no `/resume` change, no TUI owner-path change, no `PROTOCOL_VERSION` bump (nothing on the wire changed — the auth frame's `client` field already existed), no `engage_runtime` (PR 6), nothing new on `SessionRecord`.

## Testing evidence

Everything below ran against this worktree (`PYTHONPATH` pinned) with an isolated config dir + `HOME` (`LOCAL_OPERATOR_CONFIG_DIR=/tmp/lop-residency-evidence/cfg`), spawning the runtime exactly as the daemon does (`python -m local_operator.session.runtime.process` with the `LOP_MOBILE_CHILD_*` env contract) on the keyless `test` provider. A small client dials the control socket as either `attach` or `daemon` and samples `ps` while holding the connection.

### 1. A daemon-class client does not hold the runtime — exits ≈3 s after idle

```
runtime pid=66081 spawned 21:47:34
[client] record pid=66081 session=e6d62305e467 kind=daemon
[client] welcome op=projection session=e6d62305e467
[client] +  0.5s runtime 66081 00:01 104336 S
[client] EOF from runtime at +3.3s
[client] runtime exited at +3.5s while daemon held
--- 21:47:38 runtime: pid 66081 gone
INFO:...process:session runtime: idle for 3.0s (no work, no viewer, no wake within 90s); exiting cleanly
```

### 2. An attach-class client holds it warm across 65 s idle, then it exits ≈3 s after release

```
runtime pid=64616 spawned 21:46:06
[client] record pid=64616 session=83a7799c8a8c kind=attach
[client] +  0.5s runtime 64616 00:01 105952 S
[client] + 15.1s runtime 64616 00:16  85680 S
[client] + 30.1s runtime 64616 00:31  25216 S
[client] + 45.2s runtime 64616 00:46  25136 S
[client] + 60.3s runtime 64616 01:01  25264 S
[client] released attach at +65.3s; runtime 64616 01:06  25264 S
--- +2s after release: pid 64616 gone     (client's own timestamps; wall clock in the harness was sampled after `wait`)
INFO:...process:session runtime: idle for 3.0s (no work, no viewer, no wake within 90s); exiting cleanly
```

### 3. A 1-hour wake does not hold; a 1-minute recurring wake does, and fires in-process

The schedule is set through `Session.set_wake_schedules` (the same call the wake tool makes) right after the session is built, then the real `process.amain()` serves.

```
=== 1-hour wake, no viewer ===
[runtime] session=6be6085d6727 wake set; next_wake_due_at in 3600s
--- 21:48:36 alive etime=00:02
--- 21:48:40 pid 67060 gone
INFO:...process:session runtime: idle for 3.0s (no work, no viewer, no wake within 90s); exiting cleanly

=== 60 s recurring wake (every 60s), no viewer ===
[runtime] session=1badb5bbd631 wake set; next_wake_due_at in 60s
--- 21:48:56 alive etime=00:05 ...
--- 21:49:51 alive etime=01:00
--- 21:49:56 alive etime=01:05 rss 58880   <- the fire: a turn ran in-process
--- 21:50:11 alive etime=01:20   (SIGTERM sent — a recurring wake would hold forever, as designed)
--- index after the fire:
{"schedules":[{"id":"w1","every_ms":60000,"fired_count":1,"next_due_at":1788313851893,...}],...}
```

`fired_count: 1` with `next_due_at` advanced by 60 s: the wake fired once, in this process, with no supervisor.

### 4. The index: matches the transcript, removed on cancel, rebuilt on open, `stopped_at` semantics

```
$ cat cfg/wakes/6be6085d6727.json
{"cwd":"/tmp/lop-residency-evidence/work","schedules":[{"created_at":1788313715427,"every_ms":null,"fired_count":0,"id":"w1","limit":null,"message":"evidence wake","next_due_at":1788317315427,"until_at":null}],"schema":1,"session_id":"6be6085d6727","updated_at":1788313715427}
transcript wake_schedules entries: 1
index.schedules == transcript.latest.schedules: True

=== delete by hand, reopen (LOP_MOBILE_CHILD_RESUME) ===
removed 'cfg/wakes/6be6085d6727.json'
--- after reopen+exit: 6be6085d6727.json
ids ['w1'] | next_due_at 1788317315427 | has stopped_at: False
--- transcript's next_due_at: 1788317315427

=== inject stopped_at + an unknown key, reopen ===
injected: {'stopped_at': 1788300000000, 'future_key': 'kept'}
after open: has stopped_at: False | future_key: kept | ids ['w1']

=== cancel (persist []) ===
before cancel: ['w1'] | file exists: True
after cancel:  [] | file exists: False
```

`stopped_at` preserved by a **persist** (as opposed to an open) is covered by `tests/unit/wakes/test_session_index.py::test_stopped_at_preserved_on_persist_and_cleared_on_open` — the open-rewrite always runs first in a real process, so the persist case can only be observed after the marker is injected into a live session.

### 5. No fold for an attach-only runtime; exactly one for a daemon dial; the real daemon still gets projections

```
=== attach-only runtime, session a1e46b5f65e2 ===
fold builds logged for a1e46b5f65e2: 0

=== daemon-class dial, session 11d07e7d413e ===
INFO:...server:session runtime: built projection fold for session 11d07e7d413e
```

Then the **real** mobile daemon (`python -m local_operator.mobile.service --port 4199`, `LOP_MOBILE_PASSWORD` set, logged in with the cookie flow) adopting a runtime and streaming it over SSE:

```
login -> 303
runtime pid=72582 session=feeffac59c7c
--- /api/sessions: {'session_id': 'feeffac59c7c', 'model_label': 'test/e2e-model', 'cwd': '/tmp/lop-residency-evidence/work'}
--- SSE /api/sessions/feeffac59c7c/events (5 s window):
projection events received: 1
data: {"session_id": "feeffac59c7c", "pid": 72582, "kind": "daemon", "conversation_name": "", "cwd": "/tmp/lop-residency-evidence/work", "model_label": "test/e2e-model", "model_selector": "test/e2e-model", "effort": "", "effort_ladder": [], "streaming": false, ...
--- runtime log:
INFO:...server:session runtime: built projection fold for session feeffac59c7c
```

The daemon's adoption dial built the fold (one line), and the phone-facing SSE carried a projection frame for that session.

### Unit tests

New: `tests/unit/wakes/{test_store,test_install,test_session_index}.py` (21 tests: shape, staged write, directory scan, skip-on-torn/foreign-schema, preserve/clear, session persist→index, cancel→removed, open-rebuild, stale-removed-on-open, `stopped_at`, index failure never breaks persistence, install hook called only for non-empty, install failure never breaks persistence). `test_process_reaper.py` rewritten for the three terms (attach holds; daemon/phone watchers do not; wake inside/outside the window; the constant vs `MIN_WAKE_INTERVAL_MS`; reduced/broken handles read as "does not hold"; attach → release → drain; attach mid-drain cancels; wake cancelled mid-hold releases). `test_server.py` gains four `ProjectionSink` tests; `test_owned.py` gains `next_wake_due_at`; `test_import_graph.py` gains the store guard.

## Gates

```
uvx --from black==26.1.0 black --check .                       → 776 files unchanged
.venv/bin/python -m pytest tests/unit -q                       → 10594 passed, 15 skipped (13:28)
env -u NO_COLOR TERM=xterm-256color pytest tests/unit/tui -q   → 4343 passed, 12 skipped (12:08)
```

## Version

`0.44.27 → 0.44.28` (patch — rebased on #528, which shipped 0.44.27 — `sessions exit when idle`; no user-facing surface changes in this PR, since the TUI still owns its session).

### DIFF
--- a/local_operator/session/runtime/owned.py
+++ b/local_operator/session/runtime/owned.py
@@ -337,6 +337,27 @@ class OwnedSessionHandle(SessionHandle):
             return True
         return False
 
+    def next_wake_due_at(self) -> int | None:
+        """Epoch-ms of the earliest armed wake, or ``None`` when none is set.
+
+        The reaper's WARMTH signal (design §6.1 term 2): a runtime whose own
+        ``WakeScheduler`` will fire within ``WARM_WINDOW_S`` stays resident
+        rather than exiting and paying a cold start for a wake seconds away.
+        Read from the live scheduler, not the wake index — the index is a
+        derived file for processes that have no session; this process has
+        the truth in memory. A disposed scheduler reports no wakes so the
+        reaper never waits on a schedule that can no longer fire.
+        """
+        scheduler = getattr(self._session, "wake_scheduler", None)
+        if scheduler is None or getattr(scheduler, "disposed", False):
+            return None
+        try:
+            schedules = scheduler.schedules
+        except Exception:  # noqa: BLE001 — uncertainty must not pin the runtime
+            return None
+        due = [s.next_due_at for s in schedules if isinstance(s.next_due_at, int)]
+        return min(due) if due else None
+
     def _deny_pending_gates(self) -> None:
         """Refuse every parked approval/ask so teardown cannot hang on them.
 
--- a/local_operator/session/runtime/process.py
+++ b/local_operator/session/runtime/process.py
@@ -11,10 +11,16 @@ session its work.
 The child builds a session with the CLI's composition root, wraps it in the
 owned-session handle (approval/ask gates resolved from the phone), registers
 it through the normal record + control socket path, and idles until a signal
-arrives or its work and pending gates drain long enough for self-reaping (see
-:func:`_reaper`). Environment variables are the
+arrives or the residency predicate (:func:`_should_exit`) holds for one
+sustained drain. Environment variables are the
 spawn contract (``LOP_MOBILE_CHILD_CWD``, ``_PROVIDER``, ``_MODEL``) — argv
 would be ps-readable.
+
+**Residency (design §6.1).** The runtime is a unit of WORK, not of state; it
+runs its trajectory to completion and exits when idle, so a closed terminal
+costs nothing and a wake fires in a fresh process later. It stays resident
+while any of three things holds — see :func:`_should_exit` for each term and
+the reasoning behind it.
 
 This was ``mobile/child.py``. Only the phone spawns one today, but nothing in
 it is phone-specific: it is the generic "a session running with no interface
@@ -37,12 +43,25 @@ import time
 logger = logging.getLogger(__name__)
 
 #: Fine polling makes the 3-second drain predictable while remaining cheap for
-#: one event loop. Viewer traffic is deliberately not an input to this loop.
+#: one event loop. Viewer TRAFFIC is not an input to this loop (a chatty
+#: viewer does not reset the drain); viewer PRESENCE is, through term 3 of
+#: the predicate.
 REAP_CHECK_S = 0.25
 
 #: Idle runtimes are disposable once their durable session state is
 #: quiescent. This is a drain for newly arriving work, not a reconnect grace.
 DEFAULT_GRACE_S = 3.0
+
+#: A runtime whose own scheduler will fire a wake within this window stays
+#: resident instead of exiting and paying a ~1.2 s cold start (plus the
+#: supervisor's tick latency) to come back for it. Chosen to exceed
+#: ``MIN_WAKE_INTERVAL_MS`` (60 s, harness.wake) by a margin: a session with
+#: the tightest allowed recurrence then never thrashes exit → spawn → exit
+#: once a minute, because the next fire is always inside the window. Anything
+#: due further out is cheaper to leave to a cold spawn than to hold ~283 MB
+#: for. Not env-tunable on purpose — it pairs with a constant in the wake
+#: layer, and a knob would let the two drift apart.
+WARM_WINDOW_S = 90.0
 
 
 def _grace_seconds() -> float:
@@ -54,11 +73,80 @@ def _grace_seconds() -> float:
     return value if value > 0 else DEFAULT_GRACE_S
 
 
+def _wake_within_window(handle: object, *, now_ms: int | None = None) -> bool:
+    """Term 2 of the predicate: does the runtime's OWN scheduler have a wake
+    due within ``WARM_WINDOW_S``? Read through the handle (an optional
+    capability, probed) so reduced test handles and older hosts that never
+    grew the accessor behave as "no wakes" rather than crash the reaper."""
+    accessor = getattr(handle, "next_wake_due_at", None)
+    if not callable(accessor):
+        return False
+    try:
+        due_at = accessor()
+    except Exception:  # noqa: BLE001 — a broken accessor must not pin the runtime
+        logger.debug("next_wake_due_at failed; treating as no wake", exc_info=True)
+        return False
+    if not isinstance(due_at, int) or isinstance(due_at, bool):
+        return False  # None, or a shape this reaper does not understand
+    now = int(time.time() * 1000) if now_ms is None else now_ms
+    return due_at - now <= WARM_WINDOW_S * 1000
+
+
+def _viewer_attached(runtime: object) -> bool:
+    """Term 3 of the predicate: is an INTERACTIVE viewer connected?
+
+    Only ``ClientKind == "attach"`` counts — a TUI following this session, or
+    the phone's interactive attach while the user has the session open.
+    ``"daemon"`` clients (the mobile daemon's adoption dial, ``lop send``,
+    ``lop stop``, the future supervisor) deliberately do not: the daemon
+    adopts EVERY session on the machine, so if its connection held runtimes
+    warm nothing would ever exit. ``RuntimeServer.attach_clients()`` already
+    computes exactly this count for the attach cap; it is probed rather than
+    required so the reduced handles in tests keep working.
+    """
+    count = getattr(runtime, "attach_clients", None)
+    if not callable(count):
+        return False
+    try:
+        live = count()
+    except Exception:  # noqa: BLE001 — uncertainty here must not pin the runtime
+        logger.debug("attach_clients failed; treating as no viewer", exc_info=True)
+        return False
+    return isinstance(live, int) and live > 0
+
+
 def _should_exit(handle: object, runtime: object) -> bool:
-    """True when the session runtime is quiescent, regardless of viewers."""
-    del runtime  # Viewers observe work; they do not own it.
+    """The residency predicate (design §6.1): exit when ALL three hold.
+
+    1. ``handle.is_busy()`` is False — no turn, compaction, subagents, jobs,
+       queued prompts, or gate parked on a user's answer. Work is
+       authoritative: nothing below can end a turn early.
+    2. No wake is due within :data:`WARM_WINDOW_S` — a runtime about to fire
+       its own wake is cheaper kept than re-spawned (see the constant).
+    3. No interactive viewer is attached — a user looking at the session is
+       about to type, and holding the process warm turns "every message after
+       a 3 s pause costs a cold start" into "the first message of a
+       conversation costs one".
+
+    Reconciling term 3 with the older rule "watchers and replicas observe
+    work; they do not own it": both are still true, and they are about
+    different things. OWNERSHIP of the work is the turn's — a viewer leaving
+    does not abort a turn (term 1 is checked first and alone decides that),
+    and a daemon-class client never holds anything. Term 3 is about
+    READINESS: an attached interactive viewer is the one signal that the
+    next message is imminent, so residency follows it. The phone's SSE
+    watcher count (``phone_watchers``) stays out of the predicate — the
+    daemon's connection is not the user's attention, and the phone's
+    interactive attach dials as ``"attach"`` when it wants warmth.
+    """
     is_busy = getattr(handle, "is_busy", None)
-    return not (is_busy is not None and is_busy())
+    if is_busy is not None and is_busy():
+        return False
+    if _wake_within_window(handle):
+        return False
+    if _viewer_attached(runtime):
+        return False
+    return True
 
 
 async def _clean_exit(handle: object, runtime: object) -> None:
@@ -81,8 +169,12 @@ async def _clean_exit(handle: object, runtime: object) -> None:
 async def _reaper(handle: object, runtime: object, stop: asyncio.Event) -> None:
     """Exit the disposable session runtime after one uninterrupted idle drain.
 
-    Autonomous work and ordinary pending-gate timeouts are authoritative. Any
-    new activity resets the drain; viewer count intentionally does not.
+    The drain is re-checked every ``REAP_CHECK_S`` against the full predicate,
+    so any term flipping back — work arriving, a viewer attaching, a wake
+    entering the warm window — cancels it and the clock restarts from the
+    next fully-idle tick. A wake that fires during the drain starts a turn,
+    which flips ``is_busy()``: that is how "due within the drain" fires once,
+    in-process, with no supervisor involvement.
     """
     grace_s = _grace_seconds()
     while not stop.is_set():
@@ -93,9 +185,14 @@ async def _reaper(handle: object, runtime: object, stop: asyncio.Event) -> None:
         while time.monotonic() < deadline:
             await asyncio.sleep(REAP_CHECK_S)
             if stop.is_set() or not _should_exit(handle, runtime):
-                break  # work arrived (or shutdown began)
+                break  # a predicate term flipped back (or shutdown began)
         else:
-            logger.info("session runtime child: idle for %.1fs; exiting cleanly", grace_s)
+            logger.info(
+                "session runtime: idle for %.1fs (no work, no viewer, no wake within %.0fs); "
+                "exiting cleanly",
+                grace_s,
+                WARM_WINDOW_S,
+            )
             await _clean_exit(handle, runtime)
             stop.set()  # amain's wait() returns; exit code stays 0
             return
--- a/local_operator/session/runtime/process.py
+++ b/local_operator/session/runtime/process.py
@@ -76,8 +76,9 @@ def _grace_seconds() -> float:
 def _wake_within_window(handle: object, *, now_ms: int | None = None) -> bool:
     """Term 2 of the predicate: does the runtime's OWN scheduler have a wake
     due within ``WARM_WINDOW_S``? Read through the handle (an optional
-    capability, probed) so reduced test handles and older hosts that never
-    grew the accessor behave as "no wakes" rather than crash the reaper."""
+    capability, probed) so reduced test handles and older handle
+    implementations that never grew the accessor behave as "no wakes" rather
+    than crash the reaper."""
     accessor = getattr(handle, "next_wake_due_at", None)
     if not callable(accessor):
         return False
--- a/local_operator/session/runtime/server.py
+++ b/local_operator/session/runtime/server.py
@@ -223,7 +223,7 @@ class ProjectionSink(Protocol):
     The phone renders from a :class:`SessionProjection` snapshot that the
     runtime broadcasts on change; :class:`ProjectionFold` is the production
     implementation. The runtime only ever READS ``projection`` (to serialize
-    a frame) and calls ``set_pending`` (the reduced-host bridge for gate
+    a frame) and calls ``set_pending`` (the reduced-handle bridge for gate
     cards), so that is the whole contract — narrow enough that a test can
     hand in a stub and a future runtime with no phone can hand in nothing.
     """
@@ -247,16 +247,17 @@ class RuntimeServer:
         self._handle = handle
         seed = handle.session_projection_seed
         seed.kind = kind
-        # The projection fold is an OPTIONAL, injected collaborator. Given
-        # one, the runtime uses it as-is (the TUI mirror handle and tests do
-        # this). Given none — the default, and what every headless runtime
-        # does — nothing is built until a client that consumes projection
-        # semantics (the mobile daemon) actually dials, so a runtime that
-        # only ever serves a follower terminal or fires a wake constructs no
-        # fold at all. Welcomes and repaints before that moment serialize the
-        # seed directly: the seed IS the object the host's own fold mutates,
-        # so the bytes on the wire are identical either way. See
-        # ``_ensure_projection_sink``.
+        # The projection fold is an OPTIONAL, injected collaborator. A caller
+        # that already owns a fold may hand it in and the runtime uses it
+        # as-is — only tests do today; every production constructor call
+        # (the TUI at ``tui/app.py``, the owned-session process) passes none.
+        # Given none, nothing is built until a client that consumes
+        # projection semantics (the mobile daemon) actually dials, so a
+        # runtime that only ever serves a follower terminal or fires a wake
+        # constructs no fold at all. Welcomes and repaints before that moment
+        # serialize the seed directly: the seed IS the object the handle's
+        # own fold mutates, so the bytes on the wire are identical either
+        # way. See ``_ensure_projection_sink``.
         self._projection_sink: ProjectionSink | None = projection_sink
         #: How many times this runtime built a fold on its own. Observable
         #: for tests and for the "did a headless runtime pay for a fold?"
@@ -1149,7 +1150,7 @@ class RuntimeServer:
 
     @property
     def fold(self) -> ProjectionFold:
-        """The default fold, built on demand. Hosts that reach for this want
+        """The default fold, built on demand. Handles that reach for this want
         the full :class:`ProjectionFold` surface (subagent details, todos);
         an injected sink that is not one is a programming error here."""
         sink = self._ensure_projection_sink()
--- a/local_operator/session/runtime/server.py
+++ b/local_operator/session/runtime/server.py
@@ -15,8 +15,9 @@ attached) hosts one of these. It does three things:
    host-provided :class:`SessionHandle`.
 
 This was ``mobile/registrant.py`` and the class was ``Registrant``. Nothing
-about it is phone-specific except the projection fold (which PR 2 makes an
-injected collaborator): the phone daemon, an attach terminal, and — later —
+about it is phone-specific except the projection fold, which is an injected
+:class:`ProjectionSink` collaborator built lazily on the first daemon dial
+when none is supplied: the phone daemon, an attach terminal, and — later —
 wakes and background automations are all VIEWERS of one session runtime.
 ``Registrant`` remains as an alias at the bottom of this module and at the old
 import path, so no call site had to change in the move.
@@ -216,14 +217,51 @@ class SessionHandle(Protocol):
     #   lacking it answers "this session cannot receive peer messages".
 
 
+class ProjectionSink(Protocol):
+    """What the runtime needs from a projection collaborator.
+
+    The phone renders from a :class:`SessionProjection` snapshot that the
+    runtime broadcasts on change; :class:`ProjectionFold` is the production
+    implementation. The runtime only ever READS ``projection`` (to serialize
+    a frame) and calls ``set_pending`` (the reduced-host bridge for gate
+    cards), so that is the whole contract — narrow enough that a test can
+    hand in a stub and a future runtime with no phone can hand in nothing.
+    """
+
+    @property
+    def projection(self) -> SessionProjection: ...
+
+    def set_pending(self, pending: Any) -> None: ...
+
+
 class RuntimeServer:
     """One per interactive process. Construct, ``start()``, ``close()``."""
 
-    def __init__(self, handle: SessionHandle, *, kind: str = "tui") -> None:
+    def __init__(
+        self,
+        handle: SessionHandle,
+        *,
+        kind: str = "tui",
+        projection_sink: ProjectionSink | None = None,
+    ) -> None:
         self._handle = handle
         seed = handle.session_projection_seed
         seed.kind = kind
-        self._fold = ProjectionFold(seed)
+        # The projection fold is an OPTIONAL, injected collaborator. Given
+        # one, the runtime uses it as-is (the TUI mirror handle and tests do
+        # this). Given none — the default, and what every headless runtime
+        # does — nothing is built until a client that consumes projection
+        # semantics (the mobile daemon) actually dials, so a runtime that
+        # only ever serves a follower terminal or fires a wake constructs no
+        # fold at all. Welcomes and repaints before that moment serialize the
+        # seed directly: the seed IS the object the host's own fold mutates,
+        # so the bytes on the wire are identical either way. See
+        # ``_ensure_projection_sink``.
+        self._projection_sink: ProjectionSink | None = projection_sink
+        #: How many times this runtime built a fold on its own. Observable
+        #: for tests and for the "did a headless runtime pay for a fold?"
+        #: question; 0 after a lifetime with no daemon client is the claim.
+        self.projection_sinks_built: int = 0
         self._record = SessionRecord(
             pid=os.getpid(),
             kind=kind,  # type: ignore[arg-type]
@@ -556,6 +594,11 @@ class RuntimeServer:
             wants_frontend=wants_frontend,
         )
         self._clients[id(writer)] = conn
+        if kind == "daemon":
+            # The daemon is the one client that renders projections (attach
+            # clients read the welcome for identity only), so its arrival is
+            # when a lazily-built fold earns its keep.
+            self._ensure_projection_sink()
         await self._push_to(conn)  # the welcome: a full projection, unprompted
         if conn.wants_frontend:
             subscribe_frontend = getattr(self._handle, "subscribe_frontend", None)
@@ -637,8 +680,9 @@ class RuntimeServer:
     def attach_clients(self) -> int:
         """How many attach (follower terminal) connections are live.
 
-        The child reaper's front-end count: an attached TUI is a front end
-        exactly like a phone, so it must hold the child in ACTIVE."""
+        Term 3 of the runtime's residency predicate (``process._should_exit``):
+        an interactive viewer holds the runtime warm; ``daemon`` clients never
+        do. Also the attach-cap count."""
         return sum(1 for c in self._clients.values() if c.kind == "attach")
 
     async def _on_request(self, frame: dict[str, Any], conn: _ClientConn) -> None:
@@ -1037,7 +1081,9 @@ class RuntimeServer:
         """
         from local_operator.mobile.projection import cap_projection_frame
 
-        data, degraded = cap_projection_frame(self._fold.projection)
+        sink = self._projection_sink
+        projection = sink.projection if sink is not None else self._handle.session_projection_seed
+        data, degraded = cap_projection_frame(projection)
         if degraded and not self._frame_cap_warned:
             self._frame_cap_warned = True
             logger.warning(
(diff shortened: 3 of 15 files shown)


damianvtran:
### Agent review — round 1
Scope: 1ee225154e7bb63cc92bcd5adfe996855737fcd5..5c9bd03a83295771b5f51f1454641e3d80269dd4

Reviewed the diff against `1ee22515` (the merge-base with `origin/main`; `093cf9f4` on main is #528, which this branch does not yet carry). Read design §3.1–3.4, §4.2a, §4.3, §4.4, §6.1–6.4, §16 PR 2.

#### What I verified by running it (isolated `LOCAL_OPERATOR_CONFIG_DIR`, never `~/.local-operator`)

- **Gates over the whole tree, as AGENTS.md spells them:** flake8 clean · `black==26.1.0 --check` 776 unchanged · `isort==5.13.2 --check` clean · pyright 0 errors/0 warnings · `tests/unit/session tests/unit/wakes tests/unit/mobile tests/unit/session/runtime` → **913 passed** · `tests/unit/test_import_graph.py` 10 passed. CI: all six jobs green on `5c9bd03a`.
- **Live runtime probe** (`python -m local_operator.session.runtime.process`, `test`/`mock`, a raw client dialing the control socket with `{"key":…, "client": <kind>}`):
  - `daemon` held for 12 s → runtime exited at **+4.0 s** while the daemon connection was still open; log: `idle for 3.0s (no work, no viewer, no wake within 90s); exiting cleanly`; **one** `built projection fold` line.
  - `attach` held for 12 s → alive at +3/+6/+9/+12 s; released → exit rc=0 at **+3.4 s** after release; **zero** `built projection fold` lines for the attach-only lifetime.
- **Import-lightness, fresh interpreter:** importing `wakes.store` + `wakes.install` loads no `asyncio`, `pydantic`, `local_operator.session`, `local_operator.harness`, `local_operator.mobile`, `textual`; only `local_operator`, `.wakes`, `.wakes.store`, `.wakes.install` of ours.
- **Timed residency tests:** `test_process_reaper.py` run 6× idle and 3× with 4 CPU-burner processes — 9/9 green. Read the loop: the inner `while monotonic < deadline: sleep; check` always evaluates the predicate after every sleep and before the `while` re-test, so a starved loop cannot skip a check and fall into the `else:` exit. The drain is genuinely sustained.

#### Findings

**R1-1 · major · `pyproject.toml:7`** — The bump `0.44.26 → 0.44.27` collides with `origin/main`, which already carries `0.44.27` (#528, tagged `v0.44.27`, published on PyPI as `0.44.27`). `git diff origin/main..HEAD -- pyproject.toml` is empty, so after merge this PR ships **no version bump at all** and the release step would try to republish an existing version. Fix: rebase on `origin/main` (merge-tree shows no conflicts) and bump to `0.44.28`; update the "Version" line in the PR body to match.

**R1-2 · minor · `local_operator/session/runtime/server.py:254`** — The constructor comment says an injected sink is what "the TUI mirror handle and tests do". The only production constructor call for the TUI is `tui/app.py:10107` → `RuntimeServer(self._mobile_handle, kind="tui")` with **no** sink, so the TUI path takes the lazy-on-daemon-dial route like every other caller. The behaviour is fine (and is what the "no fold until a daemon dials" claim needs); the comment describes a caller that does not exist. Fix: drop "the TUI mirror handle" from that sentence, or say "a host that already owns a fold may hand it in (only tests do today)".

**R1-3 · minor · `local_operator/session/session.py:8090` (`_write_wake_index_entry` docstring)** — "Function-local import keeps `wakes.store` off this module's import graph in the direction that matters — the store must never import the session". A local import *here* does nothing to prevent `store → session`; that direction is guarded by `tests/unit/test_import_graph.py` (which is the right mechanism). The local import's real justification is the one every other function-local import in this file has: keeping `Session`'s module import cheap and avoiding a top-level cycle if `store` ever grew one. Fix: reword to say that, so the next reader does not believe the local import is the guard.

**R1-4 · nit · vocabulary** — New lines still say "host" for the handle-provider side (`process.py:92` "older hosts", `server.py:227` "reduced-host bridge", `server.py:260` "the host's own fold", `server.py:333` "Hosts that reach for this"). The project brief for this series is runtime/viewer. It is pre-existing vocabulary in `server.py` ("host-provided SessionHandle", "host-facing helpers"), so I would not hold the PR on it — but the new predicate/sink docstrings are the ones future readers will quote, and a consistent term there is cheap.

**R1-5 · nit · `local_operator/wakes/store.py:187`** — `write_entry` with empty `schedules` returns `None` after `remove_entry`, and the open-time caller (`_rebuild_wake_index_entry` → `_write_wake_index_entry`) does a `read_entry` first that is wasted in that branch. Every `Session.__init__` (including every subagent child) therefore pays one open-attempt + one unlink-attempt under `<config_dir>/wakes/`. Both are single syscalls on a path that almost never exists, so not a #401-class concern — noting it only because the comment claims "the directory is empty on the typical machine" and the open path touches it anyway. Fix (optional): skip the `read_entry` when the schedule list is empty.

#### Items checked and found correct (no finding)

- **Exit predicate:** three terms in the right order, each commented with its why; `attach` holds, `daemon` (mobile daemon adoption, `peer_client.py` bare-key dial used by `lop send`) does not; `_wake_within_window` treats `None`/non-int/`bool`/raising accessor as "no wake"; overdue wakes hold and are fired within 2 s by `load()`'s `LOAD_GRACE_MS` re-arm; `WARM_WINDOW_S = 90 > MIN_WAKE_INTERVAL_MS/1000 = 60`, commented, and pinned by `test_warm_window_exceeds_min_wake_interval`. No new timers; the reaper is the same single `asyncio.sleep(REAP_CHECK_S)` loop.
- **`next_wake_due_at()`:** disposed scheduler → `None`; no scheduler attribute → `None`; reads the live scheduler not the index (correct — the index is derived).
- **Wake index:** `mkstemp` in the target dir + `os.replace`; temp name starts with `.` and `read_index` skips dotfiles; index dir is `<config_dir>/wakes/`, outside `sessions/`; write is after `append_custom` and wrapped in `except Exception` with a `logger.warning("could not update the wake index entry", exc_info=True)` — `test_index_failure_never_breaks_wake_persistence` proves the transcript and live scheduler still take the schedule; open-rewrite runs right after `_load_wake_schedules` with `clear=("stopped_at",)`; empty → `remove_entry`; unknown keys preserved via `preserve`; `read_index` is one `os.listdir` + one open per entry; unknown `schema` skipped with a warning rather than guessed.
- **Install stub:** single chokepoint in `_persist_wake_schedules`, only for non-empty lists, never raises (and the session wraps it anyway), docstring names PR 7 and the LaunchAgent shape.
- **ProjectionSink:** `_projection_payload` falls back to `handle.session_projection_seed`, which in both `owned.py:125` and `tui_handle.py:101/169` is the same object the handle's own fold mutates, so wire bytes are unchanged; daemon dial → `_ensure_projection_sink()` once (redial reuses, `projection_sinks_built == 1`); `fold`/`set_pending` build on demand so `test_remote.py`'s reduced-host bridge is untouched; `PROTOCOL_VERSION = 5` untouched (`types.py` not in the diff).
- **#401 class:** no new flock/lock/thread on the open or dispose path; the only new I/O on open is the two syscalls in R1-5, both non-blocking in practice and wrapped.
- **PR body:** evidence sections match what I reproduced (daemon exits ≈3–4 s, attach holds, one fold line for a daemon dial, zero for attach-only); gate numbers match; the "Version" paragraph is the one stale claim (R1-1).

#### Verdict

**changes requested** — one major (R1-1, the version collision with `origin/main`); no blockers. Everything else is minor/nit and would not on its own warrant another round. Once R1-1 is fixed (rebase + `0.44.28`) and the round is re-run against the new head, I expect a terminal `clean`.

damianvtran:
Head after remediation: `94ec47e5` (rebased on `origin/main` at `093cf9f4`; the original commit is now `299befbc`, the fixes are one commit on top).

- R1-1: fixed — `94ec47e5`. Rebased on `origin/main` (conflict-free), `pyproject.toml` bumped `0.44.27 → 0.44.28`, PR body "Version" line updated.
- R1-2: fixed — `94ec47e5`. Constructor comment now says a caller that already owns a fold may inject one, that only tests do today, and that every production constructor call (TUI at `tui/app.py`, the owned-session process) passes none.
- R1-3: fixed — `94ec47e5`. `_write_wake_index_entry` docstring now gives the local import its real justification (cheap module import, no top-level cycle) and states explicitly that the store→session direction is pinned by `tests/unit/test_import_graph.py`, not by the local import.
- R1-4: fixed — `94ec47e5`. "host" removed from the new docstrings in `process.py` and `server.py` (→ handle / handle implementation); the pre-existing `server.py` module-level uses ("host-provided SessionHandle", "host-facing helpers") were left as-is per the finding's scope.
- R1-5: fixed — `94ec47e5`. `_write_wake_index_entry` skips the `read_entry` when the schedule list is empty, so a wake-free open touches `wakes/` once (the unlink attempt), not twice.

Gates re-run on `94ec47e5`: flake8 clean · `black==26.1.0 --check` 776 unchanged · `isort==5.13.2 --check` clean · pyright 0 errors/0 warnings · `tests/unit/session tests/unit/wakes tests/unit/mobile tests/unit/test_import_graph.py` → **924 passed**.

damianvtran:
### QA report — round 1
Tester: qa-tester subagent (anthropic/claude-fable-5-1)
Scope: 093cf9f4..94ec47e5

**Verdict: PASS.** Independent execution pass on the branch build — nothing from the PR body was trusted; every claim was reproduced with my own processes, timed with `ps`. Started on `5c9bd03a`; the branch was rebased + got `94ec47e5` (round-1 fixes) mid-run, so the index matrix and the core residency cells (R1/R2/R3/R5) were re-run on `94ec47e5` with identical outcomes. Full matrix with commands and raw output: `/tmp/qa/pr2-residency.md`.

**Environment.** Darwin 26.6.2 arm64 / Python 3.14.3; worktree `/tmp/lop-residency` pinned via `PYTHONPATH` (verified `local_operator.__file__` + `ps eww` per process); isolated `LOCAL_OPERATOR_CONFIG_DIR`/`LOCAL_OPERATOR_HOME`/`HOME` under `/tmp/qa2` (`~/.local-operator` untouched: 15 records before/after, no `wakes/` created there); `--hosting test --model mock`; long turns via a QA-only wrapper that sleeps inside `MockClient.stream` in that process (branch code unmodified); old build = `uv tool install local-operator==0.44.26` in a scratch tool dir; raw control-socket client dialing as `attach`/`daemon`; branch and 0.44.26 daemons on random ports; TUI via a detached pty + `pyte` render.

#### Residency (times from the record's appearance)

| # | Cell | Actual | Result |
|---|---|---|---|
| R1 | attach client held 65 s idle → release | alive at every 5 s sample to `+60.1s` (RSS settled 27 MB); `EOF from runtime at +65.1s`; `runtime gone at +3.3s after release` | PASS |
| R1b | raw `daemon`-class client held open | runtime exited under it at `+2.5s`; record unlinked | PASS |
| R2 | only the mobile daemon adopting | daemon listed `active` at +3 s; gone at ~+3.5 s; log `idle for 3.0s (no work, no viewer, no wake within 90s)`; daemon view clean afterwards (prompted sessions flip `active → previous` in ≤2 s; never-prompted ones drop out — no transcript) | PASS |
| R3 | 1-hour wake, no viewer | exits at +3.4 s; index written on open (`fired_count:0`) | PASS |
| R3 | 60 s recurring wake, no viewer | alive `00:06…00:57`; at `01:02` `wake_prompts=1`, index `fired_count:1`, `next_due_at` +60 000 ms; transcript `custom_type: wake_prompt … wake_id w1 occurrence 1` + mock reply; still alive at `01:22`; SIGTERM exits | PASS |
| R3 | one-shot wake +30 s | holds to `00:29`, fires at `00:31`, index file removed (retired), exits 2 s later | PASS |
| R3 | boundary +89 s / +95 s | 89 s: held at +12 s; 95 s: exits at +3.6 s | PASS |
| R4 | 20 s turn, no viewer (prompt via a daemon dial that closes immediately) | alive through `00:17 assistant_msgs=0`; `00:21 assistant_msgs=1`; gone at +24 s; `exit=0`; transcript complete | PASS |
| R4b | viewer leaves mid-12 s-turn | turn survives; `gone at +10.2s after release` (remaining turn + 3 s) | PASS |
| R5 | `lop send --wake` at +2.08 s into the drain (8 s turn) | `delivered and woke the session` at +2.66 s; reply lands at ~+11 s; `GONE at +14.1s`; `exit=0` | PASS |
| R5b/c | attach at +2.0 s cancels the drain; mailbox-only `send` does not hold | held to +12 s then +3.2 s; mailbox: `GONE at +3.78s`, 0 assistant msgs | PASS |
| R7 | phone SSE `/events` open 10 s (`watch` op) | `child GONE at +3.8s while SSE watcher open` | PASS |
| R8 | two viewers (5 s / 12 s) | exit only 3.3 s after the LAST leaves | PASS |
| R10/R11 | `[tool]` loop keeps `is_busy()`; SIGTERM mid-turn; `LOP_SESSION_GRACE_S=8` | resident at +16 s, SIGTERM exits; mid-turn TERM `exit=0` in 0.11 s, record unlinked; grace 8 → `exited at +8.51s` | PASS |

Fold construction: attach-only and no-client runtimes logged **0** `built projection fold`; every daemon-dialed runtime exactly **1**; real daemon SSE still streams `projection` frames with `streaming: true` during a turn.

#### Wake index (real `Session` via `spawn_owned_session`, `set_wake_schedules` = the wake tool's call)

| # | Step | Actual | Result |
|---|---|---|---|
| W6a | create | `<cfg>/wakes/<id>.json` exists; schedule dicts == transcript `wake_schedules` entry; `schema:1`, `session_id`, `cwd` | PASS |
| W6b | cancel `[]` | file removed; transcript `[]` appended | PASS |
| W6c | delete by hand → dispose → reopen (`resume=`) | rewritten on open, equals transcript | PASS |
| W6d | inject `stopped_at` + unknown key → persist another wake → reopen | both preserved on persist; `stopped_at` cleared on open, unknown key kept | PASS |
| W6e | `chmod 500 wakes/` → create | no exception; transcript entry written; stale file untouched; one `WARNING could not update the wake index entry` + `PermissionError` traceback; self-heals on reopen after chmod back | PASS |
| W7 | dir is `<cfg>/wakes` (outside `sessions/`); `read_index` over `{not json`, schema 99, `[1,2]`, `.hidden.json`, `readme.txt`, empty file, body-id≠filename | returns only good entries, no raise, per-file WARNING; filename wins; missing dir → `{}` | PASS |
| — | in-process TUI `/resume` onto a wake-carrying session | index `updated_at` = the open moment; `Wakes · 1 scheduled` in the frame | PASS |

#### No regressions

| # | Surface | Actual | Result |
|---|---|---|---|
| N8a | **0.44.26 daemon → branch runtime** | adopted, prompt `200`, transcript OK, runtime exits +3.4 s, old daemon `active → previous` ≤2 s | PASS |
| N8b | **branch daemon → genuine 0.44.26 runtime** (`PYTHONPATH=` empty, `lsof` 0 branch files) | adopted in 0.24 s, prompt `200`, old runtime logs its own `session runtime child: idle for 3.0s`, daemon `previous` at +4.7 s | PASS |
| N8c | 0.44.26 runtime + branch attach viewer | exits at +3.2 s under the viewer (old predicate ignores viewers, as before); viewer gets a clean EOF | PASS (expected) |
| N8d | `lop sessions` old/new, `--json`; old CLI `send --now`/`--wake`, new CLI substring mailbox; bad pid; no match | tables identical; all deliveries land (transcript markers ×2 each); `rc=1` errors correct | PASS |
| N8e/f/g | daemon `POST /api/sessions/start` (child = `session.runtime.process`), SSE, REST prompt, self-reap +4.2 s; `422`/`400`/`401` validation; old shim `mobile.child` (+3.8 s reap, `send --now` OK); continue-a-past-session resume child (`active` → reaped +3.8 s → `previous`) | as stated | PASS |
| N9 | TUI in-process: boot + turn (frame `▌ N9-TUI-TURN-1 hello / Hello from the mock provider!`), follower attach + steer (owner and follower frames both show it, one record), `kill -9` owner → follower takeover, `/resume` to a past session + turn there, clean exit | all as stated | PASS |
| N9 | e2e stage `env -u NO_COLOR TERM=xterm-256color .venv/bin/python -m pytest tests/e2e -m e2e -n0 -q` | `3 passed in 10.64s rc=0` | PASS |
| N9 | `tests/unit -q` on `94ec47e5` (isolated cfg, host load ~20) | `10594 passed, 15 skipped in 816.91s rc=0` (the known `test_resume_subagents_todos` flake did not surface) | PASS |
| N10 | startup weight vs `origin/main` (`093cf9f4`), fresh `-I` interpreter per tree | `import local_operator.cli` 285/285, `lop --version` 324/324, `lop sessions` 330/330 — `only-branch=[] only-main=[]` | PASS |
| N10 | `local_operator.wakes.store` (+`install`) in a fresh `-I` interpreter | 129 modules; `asyncio: False`, `pydantic: False`; `local_operator.*` = `wakes`, `wakes.store`, `wakes.install` only; `ensure_supervisor_installed → InstallOutcome(installed=False, reason='supervisor not yet available')` | PASS |

#### Findings

- **Q1-1 — minor (by design; UX note for PR 5/8).** A runtime that exits before its first turn writes no `transcript.jsonl`, so the daemon list drops it entirely instead of showing `previous`. "Tap start, walk away 4 s, session vanishes" is correct under §6.2 but is a phone-visible gesture worth a decision later. Repro: `POST /api/sessions/start`, wait 5 s, `GET /api/sessions` → no row.
- **Q1-2 — minor, pre-existing (identical on origin/main; untouched by this PR).** After an in-process `/resume` the published `run/mobile/<pid>.json` still carries the ORIGINAL `session_id` until the next heartbeat republishes from `seed.session_id`. Noting because the wake index is keyed by session id and PR 5/8 will read both. Repro: TUI `/resume <other-id>`, immediately `cat run/mobile/<pid>.json`.
- **Q1-3 — info.** Rolling-upgrade behaviour: an old runtime under a new attach viewer still exits after 3 s (old predicate). Clean EOF, no wedge. No action.

No blocker or major findings. Gaps: macOS only; the phone's interactive `attach`-class dial was exercised via a raw dial (the web bundle only issues `watch`/`unwatch` today); no real caller injects a `ProjectionSink` outside tests, so only the lazy default path was exercised in production.

All QA processes exited (`lop sessions` in the isolated dir: 0; operator's estate untouched), scratch daemons stopped, scratch worktree removed.

damianvtran:
### Agent review — round 2
Reviewer: reviewer subagent (zai/glm-5.3)
Scope: 093cf9f4e07936982ca8a53be7bf693a81685f0c..94ec47e54a88bb315ecf5c090a169ef2f7e6d0c

Re-certification of `94ec47e5` against the round-1 findings, plus a full gate re-run.

#### Round-1 verification

- **R1-1 ✓ fixed.** `git diff origin/main..HEAD -- pyproject.toml` is exactly one hunk, `0.44.27 → 0.44.28`; `origin/main` (`093cf9f4`) carries `0.44.27`, so no collision. The PR body "Version" line reads `0.44.27 → 0.44.28` and matches.
- **R1-2 ✓ fixed.** The constructor comment now says a caller that already owns a fold may inject one, that only tests do today, and names both production callers (the TUI at `tui/app.py`, the owned-session process) as passing none — consistent with what round 1 verified about the TUI's call.
- **R1-3 ✓ fixed.** The docstring gives the local import its real justification (cheap module import, no top-level cycle) and states that the store→session direction is pinned by `tests/unit/test_import_graph.py`, not by the local import.
- **R1-4 ✓ fixed.** "host" → handle / handle implementation in the new docstrings (`process.py` wake-window term, `server.py` `set_pending` bridge and `fold` property); the pre-existing module-level "host-provided"/"host-facing" uses were left alone, as the finding scoped.
- **R1-5 ✓ fixed, and the specific path holds.** `existing = wake_store.read_entry(...) if schedules else None` skips only the read; `write_entry` still runs unconditionally and its empty branch still calls `remove_entry`, so open-time removal of a stale file is intact — `test_open_with_no_schedules_removes_stale_entry` passed in every run here.
- **Rebase carried nothing accidental.** `git range-diff 1ee22515..5c9bd03a 093cf9f4..299befbc` shows the original commit carried over with only the pyproject hunk dropped (it lives in `94ec47e5` now); `git diff origin/main...HEAD --stat` is exactly the round-1 16-file scope plus the fix commit's four files. Nothing else moved.

#### Gates (whole tree, as CI runs them)

flake8 clean · `black==26.1.0 --check` 776 unchanged · `isort==5.13.2 --check` clean · pyright 0 errors / 0 warnings. Mandated suites (`tests/unit/session tests/unit/wakes tests/unit/mobile tests/unit/session/runtime` + `tests/unit/test_import_graph.py`): **924 passed** in the confirming run — but see R2-1: the first run of the same command failed 1 test, and the failure reproduces on demand. CI on `94ec47e5` at comment time: all jobs pass — `test (3.12)` 30m58s, `tui-e2e` macos + ubuntu, lint, type-check, pip-audit.

#### Findings

**R2-1 · major · `tests/unit/wakes/test_session_index.py:33` (`_schedule`) — the new wakes suite is flaky by construction: an overdue fixture arms a real 25 ms timer that races `dispose()`.** `_schedule()`'s default `next_due_at = 1_700_000_060_000` is years in the past, so `set_wake_schedules` arms the scheduler at `MIN_ARM_MS = 25 ms`. If that timer fires before `Session.dispose()` reaches `_wake.dispose()` — dispose has many awaits (background-task gather, snapshot `wait_for`), so a loaded host easily crosses 25 ms — the pump fires the wake, retires the one-shot, and persists the now-empty list, which **removes the index file the test is about to assert on**. Reproduced three ways on this host: the mandated suite run failed `test_open_rewrites_deleted_entry_from_transcript` (`FileNotFoundError` at the test's own `path.unlink()`); `tests/unit/wakes` alone under the default `-n auto` failed that test **7 of 8 runs**; and a plain `await asyncio.sleep(0.2)` between `set_wake_schedules` and `dispose` makes the file vanish deterministically. CI is green only because its dispose wins the 25 ms race — the same suite is near-always-red on a busy dev box, and this suite is a pre-merge gate, which is why a test-only defect is major and not minor. The product behaviour is correct (an overdue wake firing, retiring, and empty-persisting the entry is exactly what QA round 1's R3 cell verified); the defect is the fixture. `test_stopped_at_preserved_on_persist_and_cleared_on_open` carries the same latent race (both of its fixtures are overdue; a mid-dispose pump would empty the transcript rows it asserts on after reopen). One-line fix: make `_schedule()`'s default due future-derived (e.g. `int(time.time() * 1000) + 3_600_000`); no test in this file asserts firing behaviour, so every test in it becomes race-free.

#### Verdict

**changes requested** — R2-1 (major). Everything from round 1 is genuinely fixed and the rebase is clean; the single remaining item is a one-line fixture change plus a suite re-run. Once that lands I expect a terminal `clean`.

damianvtran:
Head after remediation: `0c9a95cf` (one commit `fix(wakes): future-derived due times in the index tests` on top of `94ec47e5`; no rebase needed — branch was already current with `origin/main`).

- R2-1: fixed — `0c9a95cf`. `_schedule()`'s default `next_due_at` is now `int(time.time() * 1000) + 3_600_000`, and the second fixture in `test_stopped_at_preserved_on_persist_and_cleared_on_open` is likewise future-derived (`+ 3_900_000`, ordered after w1). No test in the file asserts firing behaviour, so nothing arms a timer against `dispose()`.

**Proof the fix holds:**
- `tests/unit/wakes` under the default `-n auto`: **10/10 runs green** (0 failures).
- The reviewer's deterministic repro shape, applied suite-wide — a drop-in conftest patching `Session.dispose` to `await asyncio.sleep(0.2)` first, run `-n0`: **3/3 runs green, 21 passed each** (the old fixture failed this deterministically).
- The mechanism demonstrated directly, old default vs new:
  ```
  overdue (old fixture default): index exists=False live schedules after dispose=[]
  future  (new fixture default): index exists=True  live schedules after dispose=['w1']
  ```
  (a 200 ms yield between `set_wake_schedules` and `dispose`; the overdue case is the empty-persist that deleted the file.)

**Gates re-run on `0c9a95cf`:** flake8 clean · `black==26.1.0 --check` 776 unchanged · `isort==5.13.2 --check` clean · pyright 0 errors/0 warnings · `tests/unit/session tests/unit/wakes tests/unit/mobile tests/unit/session/runtime tests/unit/test_import_graph.py` → **924 passed**.

damianvtran:
CI note on run 33587528835 (head `0c9a95cf`): attempts 1 and 2 each failed ONE pre-existing flaky test outside this PR's scope; attempt 3 is fully green.

- Attempt 1: `test_resume_subagents_todos.py::test_redundant_roster_events_skip_the_sidecar_write` (`assert 2 == 1`). A/B'd against a clean `origin/main` worktree per the standing flake rule: the same module failed the same test with the same assertion in 1 of 10 runs there, and this PR's diff has **zero** hunks in the roster/sidecar path or that test file.
- Attempt 2: `analytics/test_recorder.py::test_record_reaches_store` (`assert 0 == 20`) — the analytics recorder from #479, likewise untouched by this PR (`git diff origin/main -- local_operator/analytics tests/unit/analytics` is empty).
- Attempt 3: **all 8 checks pass** (`test (3.12)` 30m46s, tui-e2e macos+ubuntu, lint, type-check, pip-audit, cli-sanity, server-sanity).

Both failures are recorded here, not filed as issues, per the team's standing rule.

damianvtran:
### Agent review — round 3
Scope: 093cf9f4e07936982ca8a53be7bf693a81685f0f..0c9a95cf1691bb09821e866e158a653984ae9c00

Verification-only round for the single round-2 major (R2-1). Head is `0c9a95cf`, one commit on top of the round-2 head `94ec47e5`; CI on `0c9a95cf` is green.

#### R2-1 verification

**R2-1 ✓ fixed.** `git show 0c9a95cf` touches exactly one file (`tests/unit/wakes/test_session_index.py`, +19/−3), and only the fixture:

- `_schedule()` now takes `*, due: int | None = None` and defaults to `int(time.time() * 1000) + 3_600_000` — an hour in the future, so `set_wake_schedules` arms nothing and there is no timer to race `dispose()`. The comment on the helper states the mechanism (overdue → `MIN_ARM_MS` timer → pump fires/retires/empty-persists → index file removed), which is the right "why" for anyone tempted to put a literal back.
- `test_stopped_at_preserved_on_persist_and_cleared_on_open` carries the same fix: the second fixture is `now + 3_900_000`, ordered after w1's `+ 3_600_000`, with a comment tying it to the same race.
- No production code changed; no assertion changed. `tests/unit/wakes/test_store.py` still uses the literal `1_700_000_060_000`, which is correct there — that file exercises `wakes.store` pure functions only (zero references to `Session` or the scheduler), so nothing in it can arm a timer.

#### Stress (in `/tmp/lop-residency`, `.venv/bin/python`, `env -u NO_COLOR TERM=xterm-256color`)

- `tests/unit/wakes -n auto` × 8: **8/8 green, 21 passed each** (0 failures; 1.4–2.3 s per run). Round 2 had this at 7-of-8 *red* on the same host.
- `tests/unit/wakes -n0` × 1: **21 passed**.
- The 2 residual warnings are pre-existing `PydanticDeprecatedSince20` notices from `local_operator/types.py:426,446`, not from this PR.

#### Static gates (whole tree, as CI runs them)

flake8 clean (rc 0) · `black==26.1.0 --check` 776 files unchanged · `isort==5.13.2 --check` clean · pyright 0 errors / 0 warnings / 0 informations.

#### New findings

None.

#### Verdict

**clean** — R2-1 is fixed at head, the flake no longer reproduces under stress on the host that reproduced it in round 2, and all gates are green on `0c9a95cf`. This round is **TERMINAL**: no blocker or major remains, and no minor/nit findings were raised. Ready to merge.

PR #529, opened 2026-09-02
feat(evaluation): managed episode context on the central compaction engine, plus episode guards

damianvtran:
## Summary

The evaluation runner now sends the model a **real, append-only, frame-bearing conversation**, managed by the central compaction engine (`local_operator/compaction`) rather than a second implementation, with every context rebuild **declared in the evidence bundle**. Alongside it, `runner/guards.py` adds benchmark-agnostic guards (budget cap, cost spike, repeated batch, no-change streak, ask loop) that stop a floundering episode as a **scored truncation** with a recorded reason.

Two verified gaps in the released runner motivated this (file:line evidence was checked independently by the architect and manager before coding):

1. **The model never saw a screenshot.** `provider_client.py` rendered a single text-only user message per turn; there was no `ImageContent` anywhere in `runner/`. A screenshot-only benchmark ran a blind agent.
2. **Every observation ever was resent each turn.** `episode.py:_decide` passed `tuple(self._transcript)` — the whole transcript — to the client, which threw it away except for a count. That is the shape of the previous pilot's $262 / 1M-token failure the moment frames go in.

So the released runner was "no context, with the machinery to become too much context." This PR turns that machinery into a managed history without growing a parallel compaction path.

## Non-activation

No CLI, TUI, mobile, or ordinary-session behaviour changes. `CompactionSettings.keep_recent_frames` defaults to `None`, which means the frame prune is **not consulted** — `run_compaction_pass` with default settings is byte-identical to the pre-feature pipeline, and `test_defaults_leave_an_ordinary_session_pass_byte_identical` pins that (same objects by identity, same content, same image blocks, same token figure). The one session-side edit is a relocation: the compaction-marker builder/renderer moved from `session.py` to `compaction/marker.py` so both hosts render it identically; the session keeps its private names as aliases and its tests are untouched. The knob is deliberately left out of the `/settings` registry — it is an opt-in for screen-driving surfaces, and the registry drift guard passes without an entry.

## Design: append-only history, rebuilt on a cadence

`EpisodeModelClient.decide` changes from `Sequence[Observation]` to `Sequence[EpisodeTurn]` (observation + the batch chosen on it + any ask answer). The runner keeps `_turns` and never renders them; `ProviderModelClient._ContextBuilder` turns them into:

[...]

PR #531, opened 2026-09-02
feat(evaluation): AWS provider, secret delivery over RPC, TTL lease, rescue sweep, real OSWorld workspace build

damianvtran:
## Summary

PR 2b of the OSWorld 2.0 adapter: the **AWS provider** and everything the first paid episode needs around it — secret delivery over the RPC boundary (schema 1.2), a boto3-direct launch with the cleanup tags and `ClientToken` at creation, a provider-created EventBridge TTL lease whose failure is **fatal**, confirmed teardown evidence codes, a descriptor-driven rescue sweep and a read-only tag audit, a real hash-verified workspace build, and preflight refusal of judged tasks. Base is #522 (the cloud-free adapter) plus the 0.44.27 tree.

**No AWS resource was created and no money was spent while building this.** Every AWS call made during development was a read-only `describe`/`list`. The paid single-episode proof runs after review and merge, with the operator's explicit go, and **depends on runner-side secret resolution that this PR does not carry** (see "Deferred").

**Non-activation.** Nothing outside `benchmarks/osworld_v2_adapter/`, `scripts/`, `local_operator/evaluation/adapters/{api,supervisor}.py` and the tests is touched, and nothing imports the new code from the CLI, TUI, mobile, tools, or configuration paths. The harness-side change is a schema bump with a defaulted field plus two additive functions (`run_rescue(secrets=)`, `discard_rescue`). `boto3` joins the **dev** extra only.

## What it implements

**Schema 1.2 — secret delivery (`adapters/api.py`, `supervisor.py`).** The released boundary had no secret path at all: `SecretRef` is a name, the worker env is a closed allowlist that strips every credential-like name, and nothing resolved a ref into bytes. `ResolvedSecret` and `secrets: tuple[ResolvedSecret, ...]` now ride **`ResetStartParams` and `BeginRescueParams` only** — never `PrepareParams` (allocation-free), never `RescueDescriptor` (persisted to disk). The field defaults to `()`, but the bump is still required: a 1.1 worker's `extra="forbid"` model would reject the key as an opaque `invalid_request` mid-episode, and the exact-version pin locks it out at the handshake instead — the same reasoning as the 1.0→1.1 bump in #520. `InfraPurpose` gains `benchmark_judge` / `benchmark_user_simulator` for the non-secret judge settings.

**`providers/aws.py` — boto3-direct launch.** Deliberately *not* OSWorld's `_allocate_vm`, because each of these is a released-contract requirement upstream cannot meet: `run_instances` carries `ClientToken=plan.client_token` (idempotent retry); `TagSpecifications` put `lop:episode` + `lop:adapter=osworld-v2` on the **instance and the volume at creation** — the cleanup contract resolves by tag from a persisted descriptor, and the operator's TTL role is condition-scoped on `aws:ResourceTag/lop:adapter`, so without the tag the lease cannot fire; the schedule is created **immediately after `run_instances` and before the readiness wait**, and any failure terminates the instance and raises `AllocationError` (OSWorld logs a warning and continues — the previous pilot lost paid instances to SIGKILL, a sleeping laptop, and a `/tmp` purge because teardown was foreground-only); and no process-wide signal handlers are installed in the asyncio worker. The instance id is then handed to upstream's `DesktopEnv(path_to_vm=<id>)` so every guest interaction is unmodified upstream code. Credentials come from the RPC field into a `boto3.session.Session` built from the values, with `region_name` explicit on every client — the operator's default profile region is `ca-central-1`, and a client that fell back to it would launch the release AMI where it does not exist. Nothing reads `os.environ` for AWS.

**Confirmed teardown.** `terminate` polls `describe_instances` until `terminated` (55 s) and otherwise reports `terminate-unconfirmed`, which the existing adapter mapping keeps as `attempted` / `rescue_required` — `shutting-down` is never accepted as confirmation. `DesktopEnv.close()` is never called (it terminates unconfirmed). `delete_schedule` distinguishes `schedule-deleted` / `schedule-absent` / the new `schedule-delete-failed` (→ `attempted`). `evaluate()` installs a capture handler on OSWorld's judge loggers and raises `ScoringUnavailable` if an ERROR was logged, because `llm_metrics` swallows every judge exception into `0.0`.

**Rescue sweep + tag audit.** `begin_rescue` builds `AwsProvider.for_teardown()` from the descriptor's infra values plus the freshly delivered secrets — no plan, no task, no `DesktopEnv` — so a worker entering at HANDSHAKEN terminates by tag with nothing but those. `scripts/osworld_rescue_sweep.py` globs `<root>/*/rescue.json` (the one explicit caller; `load_pending_rescue` still never scans), re-resolves each descriptor's refs, runs `run_rescue`, and `discard_rescue`s **only** on `aggregate.complete`. `scripts/osworld_tag_audit.py` lists every instance, volume, and `lop-ttl-` schedule carrying the adapter tag, exits 0 and prints `[]` when clean, and **never terminates** — that is the leak detector the operator relies on for a paid pilot.

**Real workspace build.** `scripts/build_osworld_adapter.py` replaces the stub. It never downloads: it verifies the durable inputs root (`~/worktrees/osworld`; never `/tmp` — macOS purged a paid run's inputs mid-episode) against the committed `config/release-v2026.08.08.json` — release manifest sha, task-hash manifest sha, every task file sha + size, task count, prepared checkout HEAD, assets manifest revision — and refuses with **exit 4 naming the path** on any mismatch. It writes a read-only, link-free workspace plus `inputs.json`, which `reset_start` re-verifies against the live root before constructing the provider. The 4.2 GB of assets stay in the root (workspace cap is 4 GiB; the guest reads them through `OSWORLD_FILE_BASE_URL`).

**Judged-task refusal.** A task whose source imports `desktop_env.evaluators.model_client` or `llm_metrics` now derives `OSWORLD_EVAL_MODEL_API_KEY` (secret) + `OSWORLD_EVAL_MODEL_PROVIDER`/`_NAME` (infra, `benchmark_judge`) as **required**, so preflight fails; `reset_start` refuses again before allocation for a host that skipped preflight. The previous pilot scored ~17% of its suite as silent zeros this way. The judge key is the one documented exception to "secrets never touch `os.environ`" — OSWorld reads it from nowhere else, inside the worker only, scrubbed on `close`.

## Where OSWorld and the contract fought the brief

1. **`requests` pin conflict.** OSWorld declares `requests~=2.31.0`; the harness requires `>=2.32.0` (security floor). Resolved with `[tool.uv] override-dependencies` on the adapter, the same resolution the previous pilot ran thirteen paid episodes on. OSWorld itself is an `[osworld]` **extra** (~380 packages — wandb, playwright, opencv) so the cloud-free wheel CI installs stays lean.
2. **Every real task scores through an `evaluate()` override, none via an `evaluator` dict.** The released `has_evaluator()` read only the dict, so `score()` would have refused **all 108** paid episodes as "task declares no evaluator". Fixed in the static parser by AST on the task class itself (`2e176510`).
3. **29/108 real tasks were not literal.** `id = TASK_ID`, `instruction = INSTRUCTION`, `user_simulator = {..., "instruction": instruction}`, parenthesised f-strings, `"...".strip()`. A pilot drawing from 79 tasks is a biased sample. A depth-bounded `_Folder` resolves exactly that closed set by walking the same module's AST — no execution, nothing outside the module, so an imported name is unresolvable by construction and the host-side import ban stands. **Coverage is now 108/108.** Five tasks (015, 050, 056, 057, 072) interpolate an *imported* name into `instruction`; they fold to the literal skeleton with `instruction_static=False`, which is honest because no harness decision reads the text. `id` and `user_simulator` must resolve completely or the task is refused naming the field (`65a631e4`).
4. **TTL from the wall budget** (DESIGN §B.5) needs `BudgetAuthorization.wall_milliseconds`, which is not on the adapter wire. The lease is `OSWORLD_TTL_SECONDS` (infra, optional) or 7200 s. Deferred and documented.
5. **OSWorld's `manager.py` raises at import** without `AWS_REGION`/`AWS_SUBNET_ID`/`AWS_SECURITY_GROUP_ID` in the environment, and `DesktopEnv` imports it even though we never call `_allocate_vm`. `vendor_bridge` mirrors those three (non-secret infra values) into the worker env for the import, and sets `ENABLE_TTL=false` so OSWorld cannot race our lease with an unnamed second schedule.

## Operator preconditions for the paid proof (one-time, by hand)

| What | Value / where |
|---|---|
| TTL role | `AWS_SCHEDULER_ROLE_ARN=arn:aws:iam::212841448981:role/lop-osworld-ttl` — exists; trusted by `scheduler.amazonaws.com`, `ec2:TerminateInstances` scoped on `aws:ResourceTag/lop:adapter=osworld-v2` |
| Security group | `AWS_SECURITY_GROUP_ID` — **must be created**: ingress TCP 5000, 5910, 9222 from the controller's current `/32` in `vpc-895235f4`. Not automated (PR 3) |
| Placement | `AWS_SUBNET_ID` (a default public subnet), `AWS_REGION=us-east-1` |
| Credentials | secrets `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY` (optional `AWS_SESSION_TOKEN`); a scoped key, not the admin key |
| Guest | `OSWORLD_CLIENT_PASSWORD`, `OSWORLD_FILE_BASE_URL`; optional `OSWORLD_INPUTS_ROOT` (default `~/worktrees/osworld`), `OSWORLD_TTL_SECONDS` (default 7200) |
| Paid venv | install `.[osworld]` (README step 3b); task 001 is plain — not judged, no simulator |

Full detail in `benchmarks/osworld_v2_adapter/README.md`.

### Live, read-only (no resources created)

```
$ .venv/bin/python scripts/osworld_tag_audit.py --region us-east-1
[]
rc=0
```

```
$ .venv/bin/python scripts/build_osworld_adapter.py --benchmark-release osworld-v2-2026.08.08 \
    --out ~/worktrees/osworld/workspaces/probe-0.1.0
{
  "assets_manifest_sha256": "4b95b3da1a0458535df52deba32d76b9d9f0e555bbbd79e337af5a712468a0e1",
  "prepared_commit": "d578d2d4e0dc82b43e270fdaa7fa89d9708cd154",
  "release_digest": "d4c777cbcd8454b8a7edf4f3df3ecdfce1197d38ae109601d8f98c22609ce8fd",
  "task_count": 108,
  "workspace": "/Users/damian/worktrees/osworld/workspaces/probe-0.1.0"
}
$ workspace_digest(...) -> fc3e5bcf42c923787f0259a8288b3efca075d5bbb0915fd9f0150cf786bf416f   # accepted: no links, read-only
```

Static parse over that real workspace: **108/108 parse, 108/108 report an evaluator, the six LLM-simulator tasks (007, 024, 026, 034, 095, 098) detected**, matching the design doc's count.

`vendor_bridge` exercised in a **stripped-env probe** (`env -i`, OSWorld installed from the pinned commit into a throwaway venv): `inject_infra_environment` → `ENABLE_TTL=false`; `load_desktop_env()` → `desktop_env.desktop_env.DesktopEnv`; `instantiate_task(task_001)` → `Task001(id=001, instance_type=None, user_simulator=None)` with an `evaluate` override. `aws describe-images` confirms the AMI root device `/dev/sda1` at 30 GiB, which the provider's volume resolver grows to OSWorld's 40 GiB floor.

### Unit (botocore `Stubber`, no network)

`test_aws_provider.py` pins the exact `run_instances` shape (`ClientToken`, both `TagSpecifications`, gp3 BDM, `InstanceInitiatedShutdownBehavior=terminate`), the schedule shape (`at(...)`, `ActionAfterCompletion=DELETE`, the `terminateInstances` universal target, the role ARN), and the failure paths the proof relies on: **TTL create fails → `terminate_instances` on the launched id → `AllocationError`, with zero readiness probes**; readiness timeout; terminate polls to `terminated`; **`shutting-down` at the deadline → `terminate-unconfirmed`**; no tagged instance → `instance-absent`; `UnauthorizedOperation` → `terminate-denied`; the three `delete_schedule` codes; judge ERROR → `ScoringUnavailable`; audit listing; clients region-pinned and built from the credential values. The clock is injected — no wall-time assertion. An autouse fixture strips ambient AWS config so a developer's `AWS_PROFILE` cannot reach the tests (it did, on the first run).

`test_secrets_and_rescue.py`: judged task refused before allocation (missing key; missing provider settings); judge key on the documented env path and scrubbed on close; AWS secrets build the provider and never touch `os.environ`; a missing ref is named without its value; TTL override reaches the provider; inputs root re-verified against a tampered manifest → `InputsMismatch`; rescue builds the teardown provider from descriptor + secrets and terminates by tag; rescue without secrets fails closed; schema-1.2 pins.

`test_spawn.py` extends the non-leak matrix across a **real process boundary**: the real wheel in a real spawned worker accepts `ResetStartParams.secrets`, and the marker value appears in neither the worker's argv nor either stdio tail; a 1.1 selector cannot be constructed.

`test_build_and_scripts.py`: fixture inputs root with pins derived from its bytes, then each input tampered in turn (task byte, hash manifest, release manifest, assets revision, missing task, wrong count, wrong commit) → **exit 4 naming the path**; happy path yields a read-only link-free workspace `workspace_digest` accepts and a rebuild is idempotent; the committed pin carries the known hashes. Audit prints `[]`/exit 0 clean and exit 1 listing a leak. Sweep unlinks only the completed descriptor and keeps one whose secret cannot be resolved.

`test_taskfile.py`: `evaluate()` override counts as an evaluator (and one outside the class does not); the four folded shapes resolve; an imported interpolation folds `instruction` partially; a partial `id`/`user_simulator` and a cyclic constant are refused naming the field.

### Gates (rebased head `e8a24711`)

- `tests/unit/evaluation/adapters/osworld`: **146 passed** × 3 under `-n 4` (20.9 s / 15.2 s / 16.1 s) and **146 passed** serially (`-n0`, 29.3 s).
- `tests/unit/evaluation`: 565 passed.
- Full `tests/unit` on the rebased head: **10612 passed, 15 skipped, 1 failed** in 9m51s. The failure is `test_tui_responsiveness.py::test_reconnect_gap_replay_does_not_stall_the_loop` — the 50 ms loop-CPU ceiling on a TUI reconnect path this PR does not touch, which read 58.4 ms under the contended full-suite run on a machine hosting several worktrees; **3/3 green re-run in isolation** (`-n0`). AGENTS.md "Timing, flakes" documents that exact bar as load-sensitive. Pre-rebase head `962969da`: 10596 passed, 15 skipped, 0 failed in 17m56s. CI's `test (3.12)` job is the authoritative run.
- `flake8 .` clean; `black==26.1.0 --check .` 775 unchanged; `isort==5.13.2 --check .` clean; `pyright` (wheel not installed, CI shape, run to completion) **0 errors**.
- `uv build` + `twine check`: both artifacts PASSED.

## Deferred

- **Runner-side secret resolution** (`runner/host_secrets.py`, `EpisodeRunner` passing `secrets=` on reset and to `run_rescue`, `rescue.json` unlink in `_close_out`). #529 owns `runner/`; this lands as a small PR after both merge. **The paid proof cannot run until it does** — today the runner sends an empty `secrets` tuple and `reset_start` would fail closed naming `AWS_ACCESS_KEY_ID`.
- **Temporary security group + IP-drift repair** (PR 3): operator-supplied SG for now.
- **Judge / user-simulator wiring** (PR 4): refusal is in; scoring through the judge with a receipt and `LLMUserSimulator` as the responder are not. **Why PR 4 must run judged tasks under its own capture, not this PR's log capture** (round-2 MINOR-3): 18 of the 19 judged tasks call `model_client.generate_text` directly inside their own `try/except Exception: return 0.0` (the `task_003.py:151-161` shape), and `generate_text` raises without logging, so a dead judge backend produces a scored 0.0 with no ERROR record for `_JudgeErrorCapture` to see. This PR's capture closes the `llm_metrics` swallow (task_007's shape) only. PR 4 must wrap `model_client.generate_text`/`generate_chat` around `env.evaluate()` and treat a raised exception as `ScoringUnavailable`, exactly as the log capture does today.
- **TTL derived from the wall budget**: needs the budget on the adapter wire.

## Versioning

Patch → **0.44.29** (0.44.28 reserved for #529). No user-facing surface changes; the adapter is opt-in infrastructure.


damianvtran:
### Agent review — round 1

Reviewer: lopdev/reviewer (kimi/k3, independent — did not author this PR)
Scope: 093cf9f4..e8a24711

Reviewed in a clean detached worktree at the exact PR head `e8a24711899db6f2774a45dbf7dff358bfe9904d` (merge-base with origin/main = `093cf9f4`). Isolated venv on Python 3.14. Every claim below was verified by reproduction against the real code (botocore Stubber, no live writes; the only AWS calls made were read-only `describe`/`list`).

## Verified claims (reproduced, not trusted)

**Leak safety (the operator's real risk)**
- **(a)** `run_instances` carries `ClientToken` minted from the episode id and `TagSpecifications` with `lop:adapter=osworld-v2`, `lop:episode`, `Name=lop-ep-<id>` on **instance AND volume** (`providers/aws.py:353-356`, `provisioning.py:117-120`). Reproduced with Stubber `expected_params`: exact match on both tag specs and token.
- **(b)** TTL lease created **after launch, before readiness**, failure is fatal: made `create_schedule` raise → `terminate_instances` called on the launched id, `AllocationError` raised, **zero readiness probes**, `schedule_created` stays False. Lease-failure + terminate-failure still raises (terminate is best-effort; the `shutting-down`-timeout path below keeps rescue_required set, so the instance is not silently lost). ✓
- **(c)** Teardown reproduced through the real `_terminate` + real `_terminate_status` + **real `aggregate_cleanup`**: stuck at `shutting-down` past deadline → `terminate-unconfirmed` → `attempted` → **`rescue_required=True`** (verified end-to-end, not assumed). `terminated`→succeeded, absent→`not_needed`, `UnauthorizedOperation`→`terminate-denied`→`attempted`. Schedule codes: deleted/absent(`not_needed`)/delete-failed→`attempted`. ✓
- **(d)** Rescue worker (fresh adapter, HANDSHAKEN, never prepared) + descriptor + secrets builds the **real** `AwsProvider.for_teardown` with `_env=None`, terminates by tag, deletes the lease. Descriptor on disk contains **no** secret values. ✓
- **(e)** Sweep unlinks `rescue.json` only when `aggregate.complete` (= `not rescue_required`). ✓
- **(f)** `scripts/osworld_tag_audit.py --region us-east-1` and `ca-central-1` live read-only → `[]`, rc 0, account 212841448981. ✓

**Secrets**
- `secrets: tuple[ResolvedSecret,...]` on `ResetStartParams`/`BeginRescueParams` **only**; not on `PrepareParams`, not on `RescueDescriptor` (confirmed by construction). Schema 1.1→1.2 pinned on the version literal + both models. boto3 `Session` built from values, `region_name` explicit on both clients (reproduced `us-east-1`), no `os.environ` reads for AWS. Judge-key env path is the only exception, allowlisted to two names (a third name is refused), scrubbed on `close`. Cross-process canary (argv/stdio tails/env) is covered by `test_spawn.py`; in-process, **no exception path echoes a secret value** (missing-ref error names the ref only; strict-mode pydantic's `max_length` echo never reaches the wire because `worker.py` swallows the exception into a digest-only `invalid_request`).

**Judge refusal / evaluate-override / parser**
- Real corpus at `~/worktrees/osworld/gated/tasks`: **108/108 parse, 108/108 report an evaluator** (evaluate-override fix confirmed). Host never imports a task module (`__import__` spy recorded nothing). Parser attacks all **refused naming the field**: imported `id`, f-string id, cyclic constant, self-cycle, depth-10 chain, arithmetic id, `open().read()` instruction, sim referencing a later attr, partial `user_simulator`. `instruction_static=False` is the only relaxation and nothing host-side reads instruction text (grep confirmed).
- Judge capture verified against real `llm_metrics`: swallowed exceptions log `logger.error` on `desktopenv.env`/`llm_metrics`, both in `_JUDGE_LOGGERS` → `ScoringUnavailable` on any ERROR.

**Workspace build**
- Fail-closed reproduced: tampered task→exit 4, consistent-tampered manifest+task→exit 4 (release pin's manifest sha catches it), assets revision drift→exit 4. Workspace files 0444, copied not linked. No git checkout → refused.

**Boundary**: nothing under `local_operator/evaluation/runner/` or `local_operator/compaction/` touched (only tests' schema pin). No CLI/TUI/mobile imports the adapter. Version `0.44.29`, exactly one `chore(release)` commit last. #529 targets 0.44.28 — a collision exists between the two open PRs (whichever merges second needs the bump re-targeted); flagging for the manager, not a finding on this PR.

## Findings

**[MAJOR-1] Untagged, leaseless EC2 relaunch on the snapshot-revert path — invisible to the audit, unreachable by rescue.** `providers/aws.py:474` calls `env.reset(task_config=…)` with `is_environment_used=False`, so the first reset is safe. But upstream `DesktopEnv.step()` sets `is_environment_used=True`, and the adapter's `execute()` drives `env.controller.execute_python_command` and then `env.evaluate()` — any second `reset` (retries, multi-step episodes that re-reset) hits `DesktopEnv.reset` → `_revert_to_snapshot` (`desktop_env.py:331`) → `AWSProvider.revert_to_snapshot` (`desktop_env/providers/aws/provider.py:197-244`), which `run_instances` a **new** instance with **no `ClientToken`, no `lop:adapter` tag, and no TTL schedule** (tags only if `AWS_INSTANCE_NAME` env set; `ENABLE_TTL=false` is forced by vendor_bridge). That new instance: (a) is invisible to `osworld_tag_audit.py` (untagged), (b) cannot be terminated by the descriptor-driven rescue (cleanup resolves by `lop:episode` tag; the new id has none), (c) outlives its episode if the laptop dies — the exact leak the PR exists to prevent. The old instance is terminated in step 2, so the tag-resolvable resource is gone too. Repro: grep of `provider.py` shows no `lop:adapter`/`ClientToken` in the relaunch block; `manager.delete_vm` is `pass`. Fix (one line): prevent `DesktopEnv` from ever re-resetting (assert/guard `_start_desktop_env` and the episode loop so `reset` is called exactly once per provider), or make the audit/rescue tag-independent. This is the pilot's real leak surface; weigh as the top blocker-adjacent finding.

**[MAJOR-2] Judge detection misses 1 of 108 real judged tasks → silent-zero risk on a paid episode.** `requirements.py:84` `is_judged` matches the substrings `desktop_env.evaluators.model_client` / `llm_metrics` in task source. Real corpus: 18 tasks detected, but **`task_007.py`** is judged via `metrics.compare_text_with_llm` — imported as `from desktop_env.evaluators import getters, metrics` (no marker substring), and `metrics/__init__.py:194` re-exports `llm_metrics`. So task_007's judge requirement is NOT derived, preflight passes, `reset_start` does not refuse, and without a key the episode scores a silent 0.0 — precisely the failure class this PR exists to kill (~17% of the previous pilot). Reproduced: `comm` of judge-users vs detected shows `task_007.py` as the only miss. (16 detected tasks call `generate_text` directly — fine.) Fix: match on the judge call surface too (`_with_llm`, `compare_pdf_answers`, `generate_text`, `metrics.compare_text_with_llm`) — AST-import-aware or call-name matching, not only module path.

**[MAJOR-3] `desktopenv.env` capture conflates benign setup ERRORs with judge failures → false `ScoringUnavailable`.** `providers/aws.py:232` captures `desktopenv.env`. That logger emits ERROR for non-judge conditions during setup/retry (`desktop_env.py:350/359/375/388` — proxy-not-ready, setup retries) and `desktop_env.py:595` `logger.error("No evaluator configured")`. Because the handler is installed around `env.evaluate()` and `evaluate()` re-runs setup via `postconfig` (and the 6 LLM-simulator tasks plus judged ones do exactly this), a recoverable setup retry that logs an ERROR makes `evaluate()` raise `ScoringUnavailable` on a **good** score. The intent is only "judge backend failed". Repro: `_JUDGE_LOGGERS` includes `desktopenv.env`; that logger's ERRORs are not judge-specific. Fix: restrict the capture to the judge loggers (`desktopenv.eval_model`, `llm_metrics`) or filter records to judge-metric names; don't take `desktopenv.env` wholesale.

**[MINOR-1] Build script does not enforce "inputs root not under /tmp".** `scripts/build_osworld_adapter.py` documents *why* the root must not be under `/tmp` (a purge destroyed a paid pilot's inputs and left an instance running) but never checks it — reproduced: `--inputs-root /tmp/rv-inputs-test` builds rc=0. The docstring is the only guard. A one-line `resolve()` prefix check against `/tmp`/`/private/tmp`/`$TMPDIR` would make the documented constraint real.

**[MINOR-2] `instruction_static=False` tasks (015, 050, 056, 057, 072) fold `instruction` to a literal skeleton with interpolations dropped silently.** Honest per the docstring (no harness decision reads it), but if any future consumer compares the static instruction to the live one, five tasks will mismatch with no flag at the comparison site. Acceptable as-is; note only.

**[NIT-1]** `adapter.py` docstring edit "The adapter EXCLUDES infeasible tasks…" lost the sentence's original subject mid-edit (reads "PR 1" → "The adapter" but the clause grammar is now slightly off). Cosmetic.

## Gates / suite state

- Adapter suite `tests/unit/evaluation/adapters/osworld`: **146 passed × 3 under `-n 4`** and **146 passed `-n0`** (no xdist race).
- `tests/unit/evaluation`: **565 passed**.
- Whole-tree gates: **flake8 rc=0, black==26.1.0 --check 775 unchanged, isort==5.13.2 --check clean, pyright 0 errors** (adapter wheel NOT installed, CI shape, run to completion).
- Full `tests/unit`: **10612 passed, 15 skipped, 1 error** — `test_episode_subprocess.py::test_killed_worker_process_group_is_reaped` errored in the *parallel* full run because a copied `--copies` venv on this host intermittently fails to load `libpython3.14.dylib` (uv 3.14 base, unrelated to this PR; the file is untouched). Re-run serially (`-n0`): **13 passed**. This is host flake, not the PR.
- **CI is authoritative and green**: `test (3.12)`, `tui-e2e (macos+ubuntu)`, lint, type-check, pip-audit, cli-sanity, server-sanity all pass on head.

## Verdict

**3 MAJOR, 0 BLOCKER.** Not terminal — the three majors all sit on the paid-pilot's leak/silent-zero surface the PR exists to close, and MAJOR-1/MAJOR-2 each reproduce against the real corpus/upstream source. Please remediate MAJOR-1 (untagged relaunch), MAJOR-2 (task_007 judge miss), MAJOR-3 (over-broad ERROR capture); MINOR-1 is a one-line hardening worth taking in the same round. Re-review needed on the new head.

damianvtran:

Implementer: lopdev/coder (anthropic/claude-fable-5-1)
Scope: 093cf9f4..b1e6c700 (round-1 head was e8a24711; four fix commits inserted before the unchanged `chore(release)` tip)

Every finding taken. Each below carries the commit, the load-bearing test, and — for the three majors — a proof run against the **real pinned upstream classes** (OSWorld installed from `d578d2d4` into a throwaway venv, `env -i`), not only against the test fakes.

---

**[MAJOR-1] Untagged relaunch via `_revert_to_snapshot` — FIXED `02f1af5e`.**

*Reachability, as asked:* from OUR runner a second `reset` is unreachable — `reset_start` is `PREPARED→RUNNING` exactly once per episode and `begin_rescue` is `HANDSHAKEN→CLEANING`, neither re-enters `PREPARED`. But the adapter hands the live env to upstream code (`env.evaluate()` runs `postconfig` setup; `DesktopEnv.reset` is public), so the boundary had to fail closed rather than rely on that convention.

*Fix:* `AwsProvider._seal_upstream(env)` runs on the **live env object** immediately after `DesktopEnv(...)` construction and before the one `reset`. It replaces every upstream method that launches or releases with a raiser of the new `UpstreamAllocationRefused`: `env._revert_to_snapshot`, `env._save_state`, `env.close`, `provider.revert_to_snapshot`, `provider.stop_emulator`, `provider.save_state`, `manager.get_vm_path`. Refusal happens **before any boto3 call**; the runner sees an adapter error and its cleanup terminates OUR tagged instance by tag and confirms it. Decision documented in the README ("Upstream is sealed after the first `reset`") and the provider docstring: one reset per episode is the contract, the seal is what makes it a guarantee.

Also: `boto3.setup_default_session(...)` is now installed from the delivered credentials before constructing `DesktopEnv`, because upstream's `start_emulator`/`get_ip_address` build ambient `boto3.client()`s with no credential argument and would have raised `NoCredentialsError` in the stripped worker. Still in-process memory, never the environment.

*Tests (all in `test_aws_provider.py`):*
- `test_a_second_reset_on_a_used_env_is_refused_before_any_boto3_call` — fake env with an upstream-shaped provider whose every method records `reached` then calls `boto3.client`; after the real `allocate`, a second `reset` raises `UpstreamAllocationRefused` with `reached == []`, and the direct `provider.revert_to_snapshot` / `manager.get_vm_path` / `_save_state` paths are individually refused.
- `test_upstream_close_is_refused_so_teardown_is_only_ever_confirmed` — `env.close()` and `provider.stop_emulator` refused, `reached == []`.
- `test_the_seal_covers_every_upstream_launch_and_release_method` — **statically scans the pinned upstream** `provider.py`/`manager.py` for any method whose body reaches `run_instances`/`terminate_instances`/`create_image` and asserts the sealed set covers it (runs when the inputs root is present; skips on CI).

*Proof against real upstream* (`AWSProvider`, `AWSVMManager`, `DesktopEnv.__new__` + seal, `boto3.client` spied to raise):
```
refused-before-boto3: DesktopEnv._revert_to_snapshot
refused-before-boto3: DesktopEnv.close
refused-before-boto3: AWSProvider.revert_to_snapshot
refused-before-boto3: AWSProvider.stop_emulator
refused-before-boto3: AWSVMManager.get_vm_path
boto3.client calls during sealed paths: []
unsealed AWSProvider.revert_to_snapshot -> boto3.client reached (as expected; seal is load-bearing)
```

---

**[MAJOR-2] `task_007` judge miss via the `metrics` re-export — FIXED `8e9e3d05`.**

`is_judged` now walks the task AST for the judge's **call surface**, not an import spelling: an import of `model_client`/`llm_metrics` (as `node.module` or as an imported name), the client entry points `generate_text`/`generate_json`, or any `*_with_llm` symbol whether imported bare or reached as an attribute (`metrics.compare_text_with_llm`). A non-judge metric from the same package (`metrics.compare_zip_files`) does not trip it.

*108-task audit on the pinned corpus* (detected vs an independent text scan of the call surface):
```
detected 19 ['task_003','task_007','task_008','task_019','task_025','task_031','task_035','task_041',
             'task_074','task_075','task_077','task_078','task_079','task_081','task_082','task_087',
             'task_089','task_090','task_092']
truth    19
missed   []
extra    []
```
Diff vs round-1 head: **+task_007**, nothing else changed. That is the reviewer's expected 19 (18 module-path + 007).

*Tests (`test_requirements.py`):* `test_judged_via_metrics_reexport_is_detected` (task_007's shape, and the bare-name import), `test_a_non_judge_metric_from_the_same_package_is_not_judged`, and `test_judge_detection_over_the_pinned_corpus_matches_the_call_surface` — asserts `detected == truth`, `len == 19`, `task_007 in detected` against the real corpus whenever the inputs root is present.

---

**[MAJOR-3] Benign `desktopenv.env` ERRORs tripping `ScoringUnavailable` — FIXED `02f1af5e`.**

One thing the reviewer's fix sketch would have missed, verified against upstream: **`llm_metrics.py` logs its swallowed judge exceptions on `desktopenv.env`** (`llm_metrics.py:29 logger = logging.getLogger("desktopenv.env")`), the same logger `desktop_env.py` uses for setup-retry ERRORs. Dropping `desktopenv.env` from the capture would have silenced the judge; keeping it wholesale was the bug. The capture now filters on **`LogRecord.pathname`**: only records emitted from `llm_metrics.py` or `model_client.py` count, on any of the three loggers those files write to. The pinned checkout is what makes a basename match safe.

*Tests (`test_aws_provider.py`), both directions:*
- `test_a_benign_setup_retry_error_does_not_mask_a_good_score` — `"Environment setup failed, retrying (1/5)..."` and `"No evaluator configured"` from `desktop_env.py` on `desktopenv.env`, score 0.75 → **0.75 returned**.
- `test_a_judge_backend_error_raises_scoring_unavailable` (×3 parametrised) — an ERROR from `llm_metrics.py` on `desktopenv.env`, and from `model_client.py` on `desktopenv.eval_model` → **`ScoringUnavailable`**.
- `test_judge_error_and_benign_error_together_still_raise`.

*Proof against real upstream* (`desktop_env.logger.error(...)` for the benign lines, then the **real** `llm_metrics.compare_images_with_llm` on nonexistent paths, which logs from `llm_metrics.py` and returns 0.0):
```
benign env ERRORs + good score -> 0.75
real llm_metrics failure -> ScoringUnavailable | judge/evaluator backend failed: Image file not found: ...
```

---

**[MINOR-1] `/tmp` inputs root not enforced — FIXED `a019a6ff`.** `refuse_volatile_root` compares the **resolved** root against `/tmp`, `/private/tmp`, `/var/tmp`, `$TMPDIR` (so the macOS `/tmp → /private/tmp` symlink cannot slip a prefix check), runs before any manifest is read, exits 4. Live: `--inputs-root /tmp/rv-inputs-test` → `rc=4`, `"resolves under /private/tmp, which the OS may purge mid-run"`, no workspace created; the durable root still builds 108 tasks. Tests: literal `/tmp` and `/private/tmp` roots refused before verification; a `$TMPDIR` root refused; a sibling accepted. The build tests point `TMPDIR` at a sibling of `tmp_path` so they still exercise verification rather than the refusal.

**[MINOR-2]** `instruction_static=False` — acknowledged as note-only; no change. The flag is on the descriptor for exactly the comparison-site case you describe.

**[NIT-1]** docstring subject — FIXED `5ed87995`.

**Version collision with #529** — noted, no change; manager re-sequences at merge time.

---

**Gates on `b1e6c700`:** adapter suite **159 passed** ×3 under `-n 4` (19.2 / 19.7 / 19.4 s) and **159 passed** `-n0` (30.7 s); `tests/unit/evaluation` 578 passed; flake8 clean; black 775 unchanged; isort clean; pyright **0 errors** (wheel not installed, run to completion). Live read-only: `osworld_tag_audit.py --region us-east-1` → `[]` rc 0. No AWS resource created.

damianvtran:
**Addendum to round-1 remediation — head moved `b1e6c700` → `8002192f`.** CI's lint job caught an unsorted import in `test_requirements.py` that my whole-tree isort check predated (it ran before the last edit to that file). Fixed into the owning commit; content otherwise identical. Updated SHAs: MAJOR-1/MAJOR-3 `02f1af5e` (unchanged), MAJOR-2 `5715bffc`, MINOR-1 `c3e33eb7`, NIT-1 `0a28ae82`, release bump `8002192f`. Scope for round 2: `093cf9f4..8002192f`. Local gates re-run on the new head: adapter suite 159 passed `-n 4`; flake8/black/isort clean across the tree.

damianvtran:
### Agent review — round 2

Reviewer: lopdev/reviewer (anthropic/claude-fable-5-1, independent — did not author this PR; round 1 was kimi/k3)
Scope: 093cf9f4..8002192f (round-2 delta reviewed as e8a24711..8002192f)

Clean detached worktree at the exact head `8002192fa17109a46f7441a72de45a9414500b56` (merge-base with `origin/main` = `093cf9f4`, verified). Isolated 3.14 venv for the harness; a second throwaway 3.12 venv with the **pinned OSWorld `d578d2d4` installed** for the real-upstream probes. Every AWS call I made was a read-only `describe`/`list`; nothing was created.

## Round-1 findings — verified by reproduction

**MAJOR-1 (upstream re-entry sealing) — FIXED, verified against real upstream.**
- Static scan of pinned `provider.py`/`manager.py` for `run_instances`/`terminate_instances`/`create_image` (my own AST scan, not the test's regex): reached only via `AWSProvider.revert_to_snapshot`, `save_state`, `stop_emulator`, and `manager.get_vm_path → _allocate_vm` (module-level). All four are sealed. `start_emulator`/`get_ip_address` are describe/`start_instances` on OUR id only. `create_vm_manager_and_provider` is called in exactly one place (`DesktopEnv.__init__:96`), before the seal; nothing in `reset`/`evaluate`/`step` constructs a fresh provider, and `path_to_vm=<id>` skips `get_vm_path` at `__init__:110-113`. No `__del__`/`atexit` teardown hooks upstream.
- Live probe (real `DesktopEnv.__new__` + real `AWSProvider`/`AWSVMManager`, `boto3.client` spied to raise): `_revert_to_snapshot`, `_save_state`, `close`, `provider.revert_to_snapshot`, `provider.stop_emulator`, `provider.save_state`, `manager.get_vm_path` → all `UpstreamAllocationRefused` with **0 boto3 calls**. A real `DesktopEnv.reset(task_config=None)` on an env with `is_environment_used=True` → refused, 0 boto3 calls. Unsealed control `AWSProvider.revert_to_snapshot` → `boto3.client('ec2')` reached. Class attributes untouched (instance-level seal confirmed).
- `_start_emulator` failure path: upstream's `except` calls the sealed `stop_emulator`, which raises inside `except Exception as stop_err` → logged as warning, original error re-raised. No unconfirmed terminate escapes.
- (c) `boto3.setup_default_session` from delivered creds: canary secret absent from `os.environ`, `repr(session)`, `repr(session.get_credentials())`, `repr(provider)`, `vars(provider)`; `AwsCredentials.__repr__` → `AwsCredentials(<redacted>)`; default-session region pinned `us-east-1`.

**MAJOR-2 (judge detection) — FIXED, 19/19 confirmed independently.** I derived truth from the upstream symbol tables (every `def` in `llm_metrics.py` + public `model_client` functions, including the `metrics/__init__.py:194-200` re-exports) and walked each of the 108 tasks' ASTs: 19 judged, identical set, `task_007` reached only via `attr:compare_text_with_llm`. Attacks against `is_judged`: `import metrics as m; m.compare_text_with_llm` ✓, `import compare_text_with_llm as ctl` ✓, `desktop_env.evaluators.metrics.compare_text_with_llm` ✓, `importlib.import_module('…model_client').generate_text` ✓, `from model_client import generate_chat` ✓, `_compare_answers_with_llm` ✓; `metrics.compare_zip_files` ✗ (correct). Not detected: `getattr(metrics, 'compare_text' + '_with_llm')` (dynamic; unreachable statically, accepted) and `metrics.compare_pdf_answers(...)` (see MINOR-2).

**MAJOR-3 (judge-error capture) — FIXED, both directions verified with the REAL modules.** Real `desktop_env.logger.error("Environment setup failed, retrying…")` + `"No evaluator configured"` + `"File not found!"` then score 0.75 → **SCORED 0.75**. Real `metrics.compare_text_with_llm` bad args → `ScoringUnavailable`; real `llm_metrics.compare_images_with_llm` missing files → `ScoringUnavailable`; real `llm_metrics.compare_text_with_llm` against a dead backend (`retry_attempts=1`, `127.0.0.1:9`) → `ScoringUnavailable | Error in compare_text_with_llm: Connection error.` Pathname robustness: full site-packages path, bare basename, and an editable `/src/…/model_client.py` all captured; `desktop_env.py` not captured. `LogRecord.pathname` derives from `co_filename`, which is the `.py` source path even when loaded from a `.pyc` (confirmed on the installed module), so site-packages vs editable makes no difference.

**MINOR-1 (volatile root) — FIXED live**: `/tmp/x`, `/private/tmp/x`, `/var/tmp/x`, `$TMPDIR/x` → rc 4 naming the resolved root, no workspace created; a `/tmp` symlink pointing at the durable root is correctly accepted (compares resolved). Durable root → **108 tasks**, `prepared_commit d578d2d4`, release digest `d4c777cb…` (matches the PR body). **But see MAJOR-1 below — this fix broke CI.**

## Round-1 "held" list — no regressions
- Lease before readiness, failure fatal: reproduced incl. **lease-create AND terminate both failing** → `AllocationError`, zero readiness probes, instance id retained on the provider, `_schedule_created=False`.
- Teardown codes through the REAL `_terminate` (Stubber) and the REAL `aggregate_cleanup`: stuck `shutting-down` past deadline → `terminate-unconfirmed`/`attempted`/**rescue_required=True**; `→terminated` → `succeeded`/False; no tagged instance → `instance-absent`/`not_needed`/False; `UnauthorizedOperation` → `terminate-denied`/`attempted`/True.
- `secrets` on `ResetStartParams`/`BeginRescueParams` only (not `PrepareParams`, not `RescueDescriptor`, confirmed via `model_fields`); `ADAPTER_SCHEMA_VERSION = "1.2"`, `SchemaVersion = Literal["1.2"]`. `region_name` explicit on Session + both clients + default session; zero `os.environ` reads in `aws.py`. Nothing under `local_operator/evaluation/runner/` or `compaction/` touched (source); no CLI/TUI/mobile import of the adapter. Cross-process canary test present and green. Rescue-from-descriptor and sweep-unlink-only-on-complete: files unchanged since round 1, their suites green.
- Live tag audit: `us-east-1` → `[]` rc 0; `ca-central-1` → `[]` rc 0.

## Findings

**[MAJOR-1] CI `test (3.12)` is RED on head `8002192f`: 8 failures in `tests/unit/evaluation/adapters/osworld/test_build_and_scripts.py`, all caused by the MINOR-1 fix.** On Linux runners pytest's `tmp_path` is `/tmp/pytest-of-runner/…`, which `refuse_volatile_root` now rejects on the resolved path before any manifest is read — so `test_happy_path_builds_a_digestable_readonly_workspace` gets rc 4, every `test_any_tampered_input_fails_closed_with_exit_4[...]` case sees the volatile-root message instead of the tamper message it asserts, `test_a_missing_task_and_a_wrong_count_both_fail` / `test_a_wrong_prepared_commit_fails` likewise, and `test_a_root_under_tmpdir_is_refused_too`'s "sibling is fine" assertion raises. The autouse `_durable_tmp` fixture (`test_build_and_scripts.py:43-51`) only re-points `$TMPDIR`; it cannot move `tmp_path` out from under the literal `/tmp` entry. It passes on macOS only because `tmp_path` lives under `/var/folders/…/T`, which stops being volatile once `TMPDIR` is re-pointed. **Repro**: `.venv/bin/python -m pytest tests/unit/evaluation/adapters/osworld/test_build_and_scripts.py -n0 --basetemp=/tmp/x` → **8 failed, 7 passed** locally; CI run 33589814513: `8 failed, 10614 passed, 19 skipped`. The remediation comment's "CI's lint job caught…" addendum re-ran local gates but not the CI test job. *Fix (one line in the fixture)*: `monkeypatch.setattr(build, "_volatile_roots", lambda: tuple(r for r in build._volatile_roots() if r not in tmp_path.resolve().parents))` — or build the fixture inputs root via `tmp_path_factory` outside `/tmp` — keeping the two literal-`/tmp` refusal tests on the unpatched function. Test-only defect; the script itself behaves correctly. Rated MAJOR rather than BLOCKER because it is not a runtime defect, but it is a hard merge stop: CI is authoritative and red.

**[MINOR-1] The seal-coverage guard test scans almost nothing in `manager.py`.** `test_the_seal_covers_every_upstream_launch_and_release_method` (`test_aws_provider.py:521`) uses `def (\w+)\(self[^)]*\):`; `[^)]*` stops at the first `)`, so any signature with a nested paren — `get_vm_path(..., screen_size=(1920, 1080), ...)` — is skipped, and module-level `_allocate_vm` (no `self`) is never a candidate. Reproduced: the regex's matched set for `manager.py` is `[__init__, initialize_registry, add_vm, …, _list_free_vms]` — `get_vm_path` absent; its `launching` set is `{revert_to_snapshot, save_state, stop_emulator}` only. The **seal itself** covers `get_vm_path` (verified live above), so this is a guard that would not fire if a future upstream added an allocating manager method or a method that calls `_allocate_vm` indirectly. *Fix*: walk `ast.FunctionDef`s (my AST scan finds `_allocate_vm` and `signal_handler` too) and treat a call to a launching module-level function as launching.

**[MINOR-2] `compare_pdf_answers` reaches the judge and is not in the symbol set.** `metrics/pdf.py:7,137` calls `_compare_answers_with_llm` for `llm_match` rules; a task calling `metrics.compare_pdf_answers(...)` is not detected (reproduced `judged=False`). No task in the pinned corpus uses it, so no live impact on the 108. *Fix*: add `"compare_pdf_answers"` to `_JUDGE_SYMBOLS`.

**[MINOR-3] The capture does not see judge failures swallowed by the TASK's own `except`, which is 18 of the 19 judged tasks.** Every direct `generate_text` caller in the corpus (`task_003.py:151-161` pattern: `try: generate_text(...) except Exception: return False/0.0`) swallows a backend failure into 0.0 without any ERROR record from `llm_metrics.py`/`model_client.py` — `model_client.generate_text` raises without logging, backends log only `warning` on transient retry. Reproduced with the real modules: task_003-shaped `evaluate` against a dead backend → **SCORED 0.0**, no `ScoringUnavailable`. The capture is documented honestly as covering `llm_metrics`'s swallow (which, in this corpus, is only `task_007`), and the PR defers judge scoring/receipts to PR 4, so this is recorded as the silent-zero path PR 4 must close, not a regression here. *Fix sketch for PR 4*: wrap `model_client.generate_text`/`generate_chat` around `env.evaluate()` to record a raised exception the same way the log capture does.

**[MINOR-4] `_JUDGE_SOURCE_FILES` basename match is pinned-checkout-safe but silently stops working on a rename.** The seal has a guard test against the pinned checkout; the capture filter does not. A one-line test asserting `os.path.basename(llm_metrics.__file__) in _JUDGE_SOURCE_FILES` when the inputs root is present would make the pin explicit (same skip pattern as the seal test).

**[NIT-1]** `_reset_count` (`aws.py:296,523`) is written twice and never read — either assert on it in the seal or drop it; a comment saying "records the invariant" is not a check.

**[NIT-2]** `test_judge_detection_over_the_pinned_corpus_matches_the_call_surface` calls its truth set "independent", but `truth_re` is the same five substrings the detector is built from, so it cannot catch a symbol both miss (e.g. `compare_pdf_answers`). Deriving truth from `llm_metrics`/`model_client` `def` names would make it actually independent.

## Gates / suite state (head `8002192f`)
- Adapter suite `tests/unit/evaluation/adapters/osworld`: **159 passed ×3 under `-n 4`** (16.4 / 15.7 / 15.1 s), **159 passed `-n0`** (31.7 s). `tests/unit/evaluation`: **578 passed**.
- Whole-tree, CI shape, run to completion: flake8 rc 0; `black==26.1.0 --check` 775 unchanged; `isort==5.13.2 --check` clean; pyright **0 errors** with the adapter wheel NOT installed.
- **CI on head: `test (3.12)` FAIL** (8 failed / 10614 passed / 19 skipped — all eight are MAJOR-1 above); lint, type-check, pip-audit, tui-e2e (macos + ubuntu) pass; cli-sanity/server-sanity skipped (fork gate). Note the macOS-only local pass is exactly why the whole-tree local run did not catch it.

## Verdict
**0 BLOCKER, 1 MAJOR** (test-only, but CI is red on head). **Not terminal.** The three round-1 majors are genuinely closed and reproduced against real upstream; leak safety and the secret boundary hold; the sole remaining stop is the Linux `tmp_path` collision with the new volatile-root refusal. Fix the fixture, push, confirm `test (3.12)` green, and re-review is a fresh-head confirmation only — I expect round 3 to be terminal. **Do not merge on this head.** MINOR-1..4 and the nits are follow-ups (MINOR-3 belongs to PR 4 as a recorded deferral).

damianvtran:

Implementer: lopdev/coder (anthropic/claude-fable-5-1)
Scope: 093cf9f4..e98b29c9 (round-2 head was 8002192f; three commits inserted before the unchanged `chore(release)` tip)

**[MAJOR-1] CI red — Linux `tmp_path` rejected by the volatile-root refusal — FIXED `d4cdc0c5` (tests only; the script's refusal is untouched and unpatched).** A `durable_path` fixture gives the build tests a per-test root under the account's **real** home (`pwd`, since the suite's conftest re-points `HOME` into basetemp), removed afterwards. A repo-worktree root — the first thing I tried — is no better here, because this repo's worktrees themselves live under `/tmp` (the failure reproduced the same way). The two literal-`/tmp` refusal tests and the `$TMPDIR` test still run against the unmodified function. Proof: `pytest tests/unit/evaluation/adapters/osworld -n 4 --basetemp=/tmp/x` → **160 passed** (was 8 failed / 7 passed on the build file); no stray dirs left under `~/.cache/lop-osworld-build-tests`.

**[MINOR-1] Seal-coverage guard scanned almost nothing in `manager.py` — FIXED `e920a2af`.** The scan is AST-based over module-level functions and class methods, marks any body reaching `run_instances`/`terminate_instances`/`create_image` as launching, closes transitively over same-module calls (`get_vm_path → _allocate_vm`), and **asserts both `_allocate_vm` and `get_vm_path` were found** so it is provably looking at `manager.py`. Nested defs (`_allocate_vm`'s `signal_handler`) are covered by sealing the enclosing function.

**[MINOR-2] `compare_pdf_answers` — FIXED `1dd6d587`.** Added to `_JUDGE_SYMBOLS`; new fixture `JUDGED_VIA_PDF` detected. Corpus audit still 19/19, zero extra.

**[MINOR-3] 18/19 judged tasks swallow judge failures in their own `except` — deferred to PR 4, recorded in the PR body's Deferred list** with the reason: `generate_text` raises without logging, so this PR's log capture only closes the `llm_metrics` swallow (task_007's shape); PR 4 must wrap `model_client.generate_text`/`generate_chat` around `env.evaluate()` and raise `ScoringUnavailable` on an exception.

**[MINOR-4] Capture file set pinned — FIXED `e920a2af`.** `test_the_judge_capture_file_set_matches_the_pinned_upstream` asserts each of `_JUDGE_SOURCE_FILES` exists under the pinned `evaluators/` tree and `getLogger()`s one of `_JUDGE_LOGGERS`.

**[NIT-1]** `_reset_count` dropped (`e920a2af`). **[NIT-2]** corpus truth set now derived from **upstream** (`1dd6d587`): every public def in `model_client.py`/`llm_metrics.py` + the re-exported `_compare_answers_with_llm` + any metric elsewhere whose body calls one (finds `compare_pdf_answers`); generic private helpers excluded because tasks define `_to_int`/`_to_float` themselves (two false positives, 063/080, surfaced and fixed while writing it).

**Gates on `e98b29c9`:** adapter suite **160 passed** ×3 `-n 4` (16.0 / 15.5 / 15.2 s), **160** `-n0` (30.6 s), **160** with `--basetemp=/tmp/x`; flake8 clean; black 775 unchanged; isort clean; pyright **0 errors** run to completion. Watching CI `test (3.12)` on this head.

damianvtran:
### Agent review — round 3

Reviewer: lopdev/reviewer (anthropic/claude-fable-5-1, independent — did not author this PR; rounds 1/2 were kimi/k3 and claude-fable-5-1 respectively)
Scope: 8002192f..e98b29c9 (round-2 head → current head; full PR range 093cf9f4..e98b29c9)

Clean detached worktree at the exact head `e98b29c9e8a6548cf26c625b33eda0e9b5fc1923`. Merge-base with `origin/main` is still `093cf9f4`; main has since moved to `9468ccef` (v0.44.28) and **none of the paths this PR touches overlap with main's movement** (`git diff --name-only 093cf9f4..origin/main` ∩ PR paths = ∅), so the un-rebased state is fine for review. Isolated 3.14 venv, adapter wheel NOT installed (confirmed `ModuleNotFoundError`). No AWS call of any kind was made this round.

## Round-2 MAJOR-1 — verified closed by reproduction

- `pytest tests/unit/evaluation/adapters/osworld -n 4 --basetemp=/tmp/x` → **160 passed** (was 8 failed on the build file at `8002192f`). Same run without `--basetemp` → 160 passed. `-n0` → 160 passed.
- **The refusal was not weakened.** `_volatile_roots`/`refuse_volatile_root` in `scripts/build_osworld_adapter.py` are byte-identical to round 2 (`git diff 8002192f..e98b29c9 -- scripts/ local_operator/` is empty). Real-path check: `scripts/build_osworld_adapter.py --inputs-root /tmp/osworld-inputs-r3 --out /tmp/r3-out …` → `rc=4`, stderr `inputs root … resolves under /private/tmp, which the OS may purge mid-run`, and no `--out` directory was created. The three refusal tests (`/tmp`, `/private/tmp`, `$TMPDIR`) still run against the unpatched function.
- **No artefacts left behind**: `~/.cache/lop-osworld-build-tests/` contained 0 entries after every run (3× parallel, serial, `--basetemp=/tmp/x`, and the evaluation suite). The read-only workspace files (`_write_readonly` chmod 0o444) do not defeat `shutil.rmtree` because the directories stay writable.
- The fixture's `pwd.getpwuid` choice is correct: the root conftest re-points `HOME` under basetemp, so `Path.home()` would have landed right back in `/tmp` on Linux.

## Round-2 minors — verified

- **MINOR-1 (seal coverage)**: reran the new AST scan standalone against the pinned `providers/aws/{provider,manager}.py`: provider → `{revert_to_snapshot, save_state, stop_emulator}`, manager → `{_allocate_vm, get_vm_path}`; both asserted names found; `methods ⊆ sealed`. `manager.py:186`'s `terminate_instances` sits in the nested `signal_handler` inside `_allocate_vm`, covered by sealing `get_vm_path`. `start_emulator` (only `start_instances`) correctly not flagged.
- **MINOR-2 (`compare_pdf_answers`)**: in `_JUDGE_SYMBOLS`; `JUDGED_VIA_PDF` fixture detected; corpus test still **19/19 with `task_007` in**, truth set now derived from upstream `def`s + wrapper scan (asserts `compare_pdf_answers ∈ wrappers`).
- **MINOR-4 (capture file pin)**: `test_the_judge_capture_file_set_matches_the_pinned_upstream` passes; confirmed by hand that `llm_metrics.py` logs on `desktopenv.env`/`llm_metrics` and `model_client.py` on `desktopenv.eval_model`, all in `_JUDGE_LOGGERS`.
- **MINOR-3 (18/19 tasks swallow judge failures)**: recorded in the PR body's Deferred list for PR 4 with the `generate_text`-raises-without-logging reasoning and the fix sketch. Acceptable deferral; the record is in the PR.
- NIT-1 `_reset_count` dropped; NIT-2 truth set independent — both done.

## Held-list spot-checks (no regression)

Schema `ADAPTER_SCHEMA_VERSION = "1.2"` / `Literal["1.2"]`; `region_name` explicit on every client and no `os.environ` read for AWS in `providers/aws.py`; no files under `local_operator/evaluation/runner/` or `compaction/` touched (the only `runner/` diff is the two test fixtures bumping `"1.1"`→`"1.2"`, from the schema commit, already reviewed in round 1); seal/logger/teardown/rescue/canary tests all in the 160 that pass. Nothing in the round-3 delta touches those code paths.

## Gates (head `e98b29c9`)

- Adapter suite: **160 ×3 under `-n 4`** (17.6 / 15.2 / 15.1 s), **160 `-n0`** (31.6 s), **160 with `--basetemp=/tmp/x`**.
- `tests/unit/evaluation`: **579 passed** (2:53).
- Whole tree, CI shape, run to completion: flake8 rc 0; `black==26.1.0 --check` 775 unchanged; `isort==5.13.2 --check` clean; pyright **0 errors** with the wheel not installed.
- **CI on head is green**: run 33592849779 — lint, type-check, pip-audit, **test (3.12) pass** (32m42s), tui-e2e macos+ubuntu, cli-sanity, server-sanity all pass.

## Findings

No BLOCKER, no MAJOR, no MINOR.

**[NIT-1]** `launching_functions` (`test_aws_provider.py`) closes transitively over `ast.Name` calls only, so a class method that reached a launch call through `self._helper()` would not be attributed to the caller. No such shape exists in the pinned upstream (checked: zero `self.<launching>()` calls in either file), so there is no gap today and the pin makes this safe; one-line hardening: also collect `c.func.attr` when `c.func` is an `Attribute` on `self`.

**[NIT-2]** `durable_path` removes each per-test root but leaves the empty parent `~/.cache/lop-osworld-build-tests/` behind on the developer's machine. Harmless.

## Note for the manager (not a finding)

`pyproject.toml` says `0.44.29`; #529 (also `0.44.29` after its re-bump) is queued to merge first, so this PR needs a re-bump to **0.44.30** at merge time.

## Verdict

**0 BLOCKER, 0 MAJOR — `clean`. This round is TERMINAL.** The round-2 MAJOR is closed by reproduction with the refusal intact, the minors are genuinely addressed, nothing from rounds 1–2 regressed, and CI is green on the exact head. The two nits are follow-ups, not a reason for another round. **Recommend merge** (after the version re-bump to 0.44.30 once #529 lands).

damianvtran:
Rebased onto `main` @ `1ce6eb22` (v0.44.29 shipped via #529) and rebumped to **0.44.30**. New head `d250d182`. Adapter/scripts/boundary content is byte-identical to the round-3 head `e98b29c9`. Manager-performed.

damianvtran:
### Agent review — round 4

Reviewer: lopdev/reviewer (anthropic/claude-fable-5-1, independent — did not author this PR; rounds 1/2/3 were kimi/k3, claude-fable-5-1, claude-fable-5-1)
Scope: e98b29c9..d250d182 (freshness round after the rebase onto `main` @ `1ce6eb22`; full PR range 1ce6eb22..d250d182)

Own clean detached worktree at the exact head `d250d1826e9ec6b7d3fede214f92e57f17e2583e`. Adapter wheel NOT installed (`ModuleNotFoundError` confirmed). No AWS call of any kind.

## Rebase structure — verified

- `merge-base(origin/main, head)` = `1ce6eb22` = `origin/main`. Fifteen commits on top, exactly one `chore(release)` and it is the tip (`d250d182`, `pyproject.toml` only). Head `0.44.30`, main `0.44.29`.
- `git diff e98b29c9 d250d182 -- benchmarks scripts local_operator/evaluation/adapters tests/unit/evaluation/adapters` → **empty**. The only files in `e98b29c9..d250d182` are main's own movement (#529: compaction/, evidence/, runner/, session/runtime, wakes/) plus `pyproject.toml`.
- Textual overlap between #529 and this PR: `pyproject.toml` (the bump) and `tests/unit/evaluation/runner/conftest.py` — the PR's delta there is only the two `"1.1"→"1.2"` schema fixture bumps, cleanly reapplied on #529's version of the file.

## Integration across the rebase — verified

#529 rewrote `runner/episode.py` but did not touch `_reset_and_observe`; it still builds `ResetStartParams(operation_id, task_id, episode_id, artifact_root)` with no `secrets` kwarg. On the rebased tree: the default `()` validates, a wire `[]` freezes to `()`, and `ADAPTER_SCHEMA_VERSION == "1.2"`. In the adapter, an empty `secrets` only fails closed on the judged path (`JudgeUnavailable`) or when the AWS provider is actually built (`_aws_credentials` → `AdapterStateError` naming `AWS_ACCESS_KEY_ID`); `provider: fake` / an injected factory never reaches it, so the FakeProvider path runs — confirmed by the spawned sealed-bundle tests passing (below).

## Gates (head `d250d182`)

- Whole tree, CI shape, to completion: flake8 rc 0; `black==26.1.0 --check` 787 unchanged; `isort==5.13.2 --check` clean; pyright **0 errors** with the wheel absent (43 s).
- Python 3.14 venv (editable install): adapter suite **228 ×3 under `-n 4`** (16.0/15.3/15.6 s), **228 `-n0`** (48 s); spawned family (`test_launch` + `osworld/test_spawn` + `runner/test_episode_subprocess`) **20/20**; zero leftover `~/.cache/lop-osworld-build-tests/` entries after every run. Full `tests/unit/evaluation` + `tests/unit/compaction` on 3.14 was still running at post time; the spawned family it contains is the only part sensitive to the issue below and is green.
- CI on head, run 33596961317: lint, type-check, pip-audit, tui-e2e (macos+ubuntu) **pass**; `test (3.12)` still pending at post time (e98b29c9 passed the same job in 32m42s and the delta is main's movement + a version bump).

## Observed but NOT a finding against this PR: 16 spawn-path failures on a local uv-managed 3.12 venv

In a 3.12 venv built from the uv-managed interpreter, 16 tests fail with `EOFError` / `failed_pre_bundle`: `test_launch::test_supervisor_launch_completes_real_handshake_and_reaps`; `osworld/test_spawn` ×4 (`real_wheel_handshakes_and_reaps`, `seals_a_verified_bundle`, `reports_a_partial_score`, `secrets_cross_the_boundary…`); `runner/test_episode_subprocess` ×11 (`real_worker_completes_a_scored_episode`, `publishes_a_frame…`, 5× `cannot_deliver_frames_from_outside_the_root[*]`, 4× `killed_at_each_cutpoint_stays_coherent[*]`).

Root cause, reproduced with the worker's stderr: the three `_real_interpreter` helpers (`test_launch.py`, `test_episode_subprocess.py`, `osworld/spawn_helpers.py`) copy the harness interpreter with `venv --copies`; the uv-managed 3.12 copy cannot dlopen `libpython3.12.dylib`, so the helper silently falls through to `shutil.which("python3")` = Homebrew **3.14.7**, while `_dependency_roots()` still writes the **3.12** site-packages into the `.pth` → `ModuleNotFoundError: pydantic_core._pydantic_core` in the worker → EOF. **`main` @ `1ce6eb22` fails identically** (4/5 in the same 3.12 venv), and none of the three helpers is touched by this PR (they date from #511/#522). CI installs with `pip install -e` on a native runner where the copy works, which is why e98b29c9 was green. Recorded here as `deferred — pre-existing test-harness fragility on main, out of scope; one-line hardening: have the fallback probe check `sys.version_info[:2]` matches the harness before accepting an interpreter, or skip instead`. No issue filed per team policy.

## Findings

No BLOCKER, no MAJOR, no MINOR, no NIT. The rebase is content-identical for every path this PR owns, and the one cross-package seam (`ResetStartParams` runner→adapter) is intact on the rebased tree.

## Verdict

**0 BLOCKER, 0 MAJOR — `clean`. This round is TERMINAL** and fresh against head `d250d182`. **Recommend merge** once CI `test (3.12)` on run 33596961317 completes green.

damianvtran:
**Addendum to round 4:** the full `tests/unit/evaluation` + `tests/unit/compaction` run on the 3.14 venv finished after posting — **841 passed, 0 failed** (2:24) on head `d250d182`. Verdict unchanged: clean, terminal.

PR #532, opened 2026-09-02
feat(anthropic): 1h prompt-cache TTL for large contexts

damianvtran:
## Summary of Changes

**Feature (C1 standard / pre-authorized):** on the Anthropic wire, once a session's context is large, every prompt-cache marker carries the 1-hour TTL (`cache_control: {type: ephemeral, ttl: "1h"}`) instead of the default 5 minutes.

Why: a 5m entry expires while a 150k–500k session idles 6–25 minutes waiting on subagents, wakes or the user, and its next call rewrites the WHOLE prefix at 1.25× base. A 1h write costs 2× base but survives the gap. Measured over 24h on this harness's own traffic (`~/.local-operator/analytics.db`): **276 TTL-expiry rewrites of >150k contexts = 89.5M write tokens ≈ 112M base-equivalent avoided**, versus **14.7M incremental writes on those contexts ≈ 11M base-equivalent extra at 2×** — roughly a 10:1 trade.

- **Threshold setting** `providers.anthropic.cache_ttl_1h_min_context_tokens` (default `150000`; `0` disables). Read in `SessionStreamFn._client_for` via `_anthropic_cache_ttl_1h_min_context_tokens` — the same path/shape as `_openai_api_mode` for `providers.openai.api` — and passed through `client_for_spec(anthropic_cache_ttl_1h_min_context_tokens=…)` to `AnthropicClient`. Registered in `settings_io` so `/settings` can edit it. Bare `AnthropicClient()` defaults to OFF, so scripts/tests send the historical body.
- **One TTL per request.** Anthropic's "Mixing different TTLs" rule: a 1h entry must appear before any 5m entry, so `_cache_ttl_for` decides once and BOTH `_system_blocks` and `_message_cache_breakpoints` render every marker with it. The marker count and `MAX_CACHE_BREAKPOINTS=4` budget are unchanged — the TTL rides on the existing markers.
- **Context size**: new optional `ChatRequest.context_tokens_hint`, stamped per request by the OWNER of the call — never remembered on the shared `SessionStreamFn`, which only passes it through. The harness loop seeds a run from `LoopConfig.get_context_tokens_hint` (the session's last provider-reported `Usage.context_tokens`, also seeded from the transcript on resume) and, once a call in the run reports a count, prefers that for every later request of the same run — so call N+1 of a tool loop goes 1h as soon as call N crosses the threshold, and a subagent (one turn for its whole life) learns its own size (review F8/F9). `Session.complete_aside`/`advise_compaction` stamp the session's own hint; the compaction summary errand pins `context_tokens_hint=0` because its write-once prompt gains nothing from 1h markers (review F2); the isolated naming errand carries none. When absent (first call, fork's first request) the client falls back to a byte estimate of the serialized body (`len(json) // 4`, with base64 image payloads excluded and a 1.6k-token floor per image), computed after system, tools and messages are in the body. Because nothing per-conversation lives on the stream fn, subagents sharing the parent's stream fn can neither inherit nor poison the parent's hint (review F1/F8). No hysteresis: the threshold flips one way as context grows; dropping back to 5m after compaction is fine (the compacted prefix is new content either way) — documented in the method docstring.
- **Usage split**: `Usage.cache_write_5m_tokens` / `cache_write_1h_tokens` parsed from `usage.cache_creation.ephemeral_{5m,1h}_input_tokens` (subsets of `cache_write_tokens`, per the docs' "sum" statement); folded in `harness/jobs.py`; recorded to a new migrated `calls.cache_write_1h_tokens` ledger column (`_MIGRATION_COLUMNS`, old rows read 0 which is true for them).
- Version `0.44.27 → 0.44.30` (patch: small self-contained feature). **Three sibling PRs were assigned 0.44.28/29/30 up front; the merger re-sequences if the merge order differs.**

**Follow-up (not in this PR):** `price_snapshot` / `cost_for_usage` still price ALL cache writes at the 5m rate (1.25×); the new column is what makes a split rate possible. Likewise, `cache_write_1h_tokens` is a **90-day raw-ledger diagnostic** (the `calls` table), deliberately not threaded into `usage_daily`/`usage_monthly` or the read projection: the rollup tables have no ALTER-migration path (unlike `calls`' `_MIGRATION_COLUMNS`), so adding a measure there is a schema change of its own — tracked as a follow-up (review F5); within the ledger window the trade is answerable via raw SQL on the column.

## Related Issue(s)

- Related: prompt-cache cost analysis (sibling PRs: advisor/aside `tool_choice` cache split, usage-aware account selection).

## Impact

- Requests below the threshold (and with the feature off) are byte-identical to before: the 5m marker still omits the `ttl` key.
- Analytics DB gains one column via the existing idempotent `ALTER TABLE` migration path.
- Cost: +2× vs 1.25× on incremental writes for contexts ≥150k; −1.25× × prefix on every avoided expiry rewrite.

### Live verification against api.anthropic.com (OAuth credential from the shared store, `claude-haiku-4-5`, no extra `anthropic-beta` header needed — only `oauth-2025-04-20` was sent)

**1. 1h TTL accepted on OAuth (HTTP 200) and reported as a 1h write:**
```
status 200
beta header sent: oauth-2025-04-20
usage: {"input_tokens": 10, "cache_creation_input_tokens": 6623, "cache_read_input_tokens": 0,
        "cache_creation": {"ephemeral_5m_input_tokens": 0, "ephemeral_1h_input_tokens": 6623}, "output_tokens": 5}
```
(A first attempt with a ~700-token system block returned 200 with all-zero cache counters — below haiku's 4096-token cache minimum, not a TTL problem.)

**2. Timed pair: same prefix 370 s later — 1h READ vs 5m REWRITE** (`/tmp/lop-1h-probe/evidence.py`, both arms same size, same account):
```
[22:50:40] 1h WRITE:            cache_creation_input_tokens=6627  cache_read=0   cache_creation={5m:0, 1h:6627}
[22:50:41] 5m WRITE (control):  cache_creation_input_tokens=6627  cache_read=0   cache_creation={5m:6627, 1h:0}
sleeping 370s
[22:56:52] 1h READ after 370s:  cache_creation_input_tokens=0     cache_read=6627  cache_creation={5m:0, 1h:0}
[22:56:53] 5m after 370s:       cache_creation_input_tokens=6627  cache_read=0     cache_creation={5m:6627, 1h:0}
```
The 1h entry served a full-prefix read after 6 min 12 s; the 5m entry had expired and was rewritten in full.

(A first run of this pair 429'd on the reads — `"This request would exceed your account's rate limit"` — because the sticky account pick landed on a row at 100% of its 5h window; rerun pinned to the credential at 9%.)

**3. Real client path** (`client_for_spec(..., anthropic_cache_ttl_1h_min_context_tokens=150_000)` + `AnthropicClient.stream` with `context_tokens_hint=200_000`):
```
markers on wire: [{'type': 'ephemeral', 'ttl': '1h'}, {'type': 'ephemeral', 'ttl': '1h'}, {'type': 'ephemeral', 'ttl': '1h'}]
HTTP OK; parsed Usage: {"input_tokens": 3, "cache_read_tokens": 0, "cache_write_tokens": 7174,
  "cache_write_5m_tokens": 0, "cache_write_1h_tokens": 7174, "context_tokens": 7177, "output_tokens": 5}
```
Total live traffic: 7 calls of ≤ ~7k tokens (probe ×2, pair ×4 + 2 that 429'd, client path ×1).

### Unit tests (new)
- `tests/unit/providers/test_clients.py`: threshold off → no `ttl` anywhere; hint ≥ threshold → every marker `ttl: 1h`, none 5m; hint below → 5m; no hint → byte estimate decides (and includes tools); breakpoint budget unchanged at 4; `client_for_spec` passes the threshold; `cache_creation` split parsed; missing `cache_creation` object → 0/0.
- `tests/unit/model/test_configure.py`: settings reader (missing/malformed/bool/negative → default, `0` honoured); two end-to-end tests driving REAL `Session` objects through a real `SessionStreamFn` on a mocked Anthropic wire with no manual hint handling anywhere — (a) turn 1 call 1 (no hint) 5m, the provider reports 200k with a tool call, call 2 of the SAME turn goes 1h on every marker, an isolated `complete_once` errand between turns (reporting 5k) carries no hint and stays 5m, turn 2 call 1 goes 1h on the session's own 200k; (b) a parent `Session` works up 300k, a child `Session` is then constructed on the SAME stream fn and runs its turn (5m on its own 10k body), and the parent's post-child request goes 1h on its own 300k — both contamination directions, review F1/F8.
- `tests/unit/harness/test_loop.py` (added): loop-level contract — the host seed covers call 1 and in-run counts take over (`[140k, 160k, 170k]`), no seed → call 1 unstamped and the run still learns from call 1, a call that omits `context_tokens` keeps the last count — review F9.
- `tests/unit/session/test_aside.py` (added): an aside carries the hint the turn's last call reported. `tests/unit/session/test_session.py` (added): with `mid_turn_enabled=False` the boundary hook still advances the session's hint after every call and each loop request is stamped from the previous call of the run — review F9.
- `tests/unit/providers/test_clients.py` (added): the byte estimate excludes base64 image payloads — a 1.4M-char image no longer flips a small body to 1h, and the 1.6k-per-image floor is still counted (single image stays 5m under a 3.3k bar, two images tip a 3k bar) — review F3.
- `tests/unit/harness/test_jobs.py`, `tests/unit/session/test_frontend_state.py` (added): the subagent job accumulator and the frontend turn-end join fold `cache_write_5m_tokens`/`cache_write_1h_tokens` wherever `cache_write_tokens` folds — review F4.
- `tests/unit/analytics/test_store.py`: new column recorded and migrated onto a pre-1h DB (old rows read 0).
- `tests/unit/test_config.py`, `tests/unit/test_settings_io.py`: default pinned to the consumer constant.

### Gates (whole tree, as CI)
```
uvx --from black==26.1.0 black --check .                       → 769 files would be left unchanged
pytest tests/unit/providers tests/unit/model tests/unit/session tests/unit/harness tests/unit/analytics tests/unit/test_config.py tests/unit/test_settings_io.py -n 4 → 2349 passed
```

(Remediation round 4: re-run at head `f37019c8` (rebased onto `47bc768f`, version 0.44.33) — all four gates clean (`black` → 802 files would be left unchanged); `env -u NO_COLOR TERM=xterm-256color pytest tests/unit/tui -n 4 → 4353 passed, 12 skipped` (the suite CI's `test (3.12)` was failing in; F11's `test_click_selects_a_row_without_arming_a_text_selection` reproduced red on `bf6be177` and passes with the fix); `pytest tests/unit/providers tests/unit/model tests/unit/session tests/unit/harness tests/unit/analytics tests/unit/compaction -n 4 → 2514 passed` (the round-1 `test_live_progress_never_schedules_the_roster_writer` flake appeared once in three runs and passed in the other two and serially).)

(Remediation round 2: re-run at head `aaad706d` — all four gates clean; `pytest tests/unit/providers tests/unit/model tests/unit/session tests/unit/harness tests/unit/analytics -n 4 → 2228 passed` (+5 tests over round 2's 2223); `tests/unit/test_fork.py tests/unit/compaction → 274 passed`. Negative check: reverting only the loop's `context_tokens_hint=` stamp fails all six new/rewritten hint tests.)

(Remediation round 1: re-run at head `67d02901` — one transient failure of `test_live_progress_never_schedules_the_roster_writer` appeared once under `-n 4` and never reproduced in 10 subsequent runs including serial `-n0`; it also passes on the round-1 head `2cf0d11e`, so it is not introduced by this change.)

### Not validated
- A live ≥150k-token request (accounts are near their 5h caps; the threshold path is exercised via the hint on a 7k body, which produces the identical wire shape).
- The end-to-end `lop` TUI session; the plumbing is covered by the real-`Session`-through-real-`SessionStreamFn` tests above.

## Checklist

- [x] I have updated the documentation when required (setting documented in `config.py` default and `settings_io` help).

### DIFF
--- a/local_operator/analytics/model.py
+++ b/local_operator/analytics/model.py
@@ -190,6 +190,15 @@ class CallSnapshot:
     # a future caller that priced off-thread itself). The normal recording path
     # leaves this False so the store prices from the model + token counts.
     priced: bool = False
+    # The 1-hour-TTL slice of ``cache_write_tokens`` (Anthropic
+    # ``cache_creation.ephemeral_1h_input_tokens``), a SUBSET of it. Kept as
+    # its own column rather than folded in because it is priced at 2x base
+    # where a 5m write is 1.25x, and the point of the large-context 1h TTL is
+    # a trade that has to stay measurable after the fact. 0 on every other
+    # provider and on pre-1h Anthropic responses. NOTE: ``price_snapshot``
+    # still prices ALL writes at the 5m rate; splitting the rate is a
+    # follow-up, and this column is what makes it possible.
+    cache_write_1h_tokens: int = 0
     # Provider-reported dollar cost for this call (OpenRouter ``usage.cost``,
     # any compat provider that precomputes billing). Copied off ``Usage.usd_cost``
     # at record time so ``price_snapshot`` can prefer it over the registry table
--- a/local_operator/analytics/store.py
+++ b/local_operator/analytics/store.py
@@ -115,7 +115,12 @@ CREATE TABLE IF NOT EXISTS calls (
   -- priceable. Integer so the aggregate SUM is exact; see CallSnapshot.
   cost_micro INTEGER NOT NULL DEFAULT 0,
   cost_known INTEGER NOT NULL DEFAULT 0,
-  {_COMPONENT_COLUMNS}
+  {_COMPONENT_COLUMNS},
+  -- The slice of cache_write_tokens written with the 1-hour TTL (Anthropic
+  -- ``cache_creation.ephemeral_1h_input_tokens``); the 5m slice is the
+  -- remainder. Priced at 2x base rather than 1.25x, so the two must be
+  -- separable to judge whether the large-context 1h TTL pays for itself.
+  cache_write_1h_tokens INTEGER NOT NULL DEFAULT 0
 );
 CREATE INDEX IF NOT EXISTS idx_calls_ts ON calls(ts_ms);
 CREATE INDEX IF NOT EXISTS idx_calls_session ON calls(session_id);
@@ -207,6 +212,7 @@ _CALL_COLUMNS = (
     "cost_micro",
     "cost_known",
     *(f"c_{key}" for key in COMPONENT_KEYS),
+    "cache_write_1h_tokens",
 )
 
 #: Columns added AFTER the first shipped schema. A database created by an older
@@ -232,6 +238,9 @@ _MIGRATION_COLUMNS: tuple[tuple[str, str], ...] = (
     # rollup tables). After this ALTER they read ``c_images=0`` rather than
     # being re-apportioned — we cannot honestly unbake a historical estimate.
     ("c_images", "INTEGER NOT NULL DEFAULT 0"),
+    # Rows recorded before the Anthropic 1h TTL shipped were all 5m writes, so
+    # 0 here is the truth for them, not a placeholder.
+    ("cache_write_1h_tokens", "INTEGER NOT NULL DEFAULT 0"),
 )
 
 #: The names in ``_MIGRATION_COLUMNS`` as a set, for the "is this column optional
@@ -339,6 +348,7 @@ def _row_values(snapshot: CallSnapshot, cost_micro: int, cost_known: bool) -> tu
         int(cost_micro),
         1 if cost_known else 0,
         *(components[key] for key in COMPONENT_KEYS),
+        snapshot.cache_write_1h_tokens,
     )
 
 
--- a/local_operator/analytics/store.py
+++ b/local_operator/analytics/store.py
@@ -120,6 +120,13 @@ CREATE TABLE IF NOT EXISTS calls (
   -- ``cache_creation.ephemeral_1h_input_tokens``); the 5m slice is the
   -- remainder. Priced at 2x base rather than 1.25x, so the two must be
   -- separable to judge whether the large-context 1h TTL pays for itself.
+  -- SCOPE: a RAW-LEDGER diagnostic, deliberately NOT in the usage_daily /
+  -- usage_monthly rollups or the report projection — the rollup tables have
+  -- no ALTER migration path (CREATE TABLE IF NOT EXISTS cannot add a column
+  -- to an existing one, unlike calls' _MIGRATION_COLUMNS), so threading it
+  -- there is a schema change of its own (follow-up tracked on the feature
+  -- PR). Answers the trade question within the 90-day ledger window
+  -- (DEFAULT_RETENTION_DAYS); beyond that, price the split when it ships.
   cache_write_1h_tokens INTEGER NOT NULL DEFAULT 0
 );
 CREATE INDEX IF NOT EXISTS idx_calls_ts ON calls(ts_ms);
--- a/local_operator/config.py
+++ b/local_operator/config.py
@@ -156,7 +156,22 @@ DEFAULT_CONFIG = Config(
             # Direct OpenAI GPT-5 calls use the public Responses API by default.
             # Set `providers.openai.api` to `chat_completions` for an explicit
             # compatibility opt-out; other OpenAI-shaped providers never read it.
-            "providers": {"openai": {"api": "responses"}},
+            #
+            # `providers.anthropic.cache_ttl_1h_min_context_tokens`: once a
+            # session's context reaches this many tokens, Anthropic requests
+            # carry the 1-hour prompt-cache TTL instead of the default 5 minutes.
+            # A 1h write costs 2× base (vs 1.25× for 5m), but a large context
+            # that idles past 5 minutes — waiting on subagents, a wake, or the
+            # user — otherwise rewrites the WHOLE prefix on its next call.
+            # Measured over 24h on this harness's own traffic: 276 TTL-expiry
+            # rewrites of >150k contexts cost 89.5M write tokens (~112M
+            # base-equivalent), while the incremental writes on those contexts
+            # were only 14.7M (~11M base-equivalent extra at 2×). 150k is the
+            # size above which the rewrite dominates; 0 disables the feature.
+            "providers": {
+                "openai": {"api": "responses"},
+                "anthropic": {"cache_ttl_1h_min_context_tokens": 150_000},
+            },
             # One ordered cascade for every text-model call. Entries may be
             # "provider/model" strings or {provider, model, effort} mappings;
             # usage-aware switching is opt-in because it spends one lightweight
--- a/local_operator/harness/jobs.py
+++ b/local_operator/harness/jobs.py
@@ -132,6 +132,8 @@ def _merge_accounting_component(
     total.output_tokens += component.output_tokens
     total.cache_read_tokens += component.cache_read_tokens
     total.cache_write_tokens += component.cache_write_tokens
+    total.cache_write_5m_tokens += component.cache_write_5m_tokens
+    total.cache_write_1h_tokens += component.cache_write_1h_tokens
     total.reasoning_tokens += component.reasoning_tokens
     if component.context_tokens is not None:
         total.context_tokens = component.context_tokens
--- a/local_operator/harness/loop.py
+++ b/local_operator/harness/loop.py
@@ -492,6 +492,16 @@ class AgentLoop:
         # every request under it so the session's frozen auto-effort cannot
         # raise the retry back to the rung that produced nothing.
         effort_ceiling: str | None = None
+        # The provider-reported context size of the latest call in THIS run
+        # (``Usage.context_tokens``), stamped onto the next request as its
+        # prompt-cache TTL hint. Run-local on purpose: the host's
+        # ``get_context_tokens_hint`` is the cross-turn seed (last turn's
+        # final call, or a resumed transcript's), and a long tool loop can
+        # grow past the TTL threshold dozens of calls before the host's figure
+        # is next refreshed — a subagent is ONE turn for its whole life and
+        # would otherwise never carry a hint at all. Once a call in this run
+        # has reported, its count beats the seed (review F9).
+        run_context_tokens: int | None = None
 
         yield AgentStartEvent(generation=generation)
 
@@ -529,7 +539,15 @@ class AgentLoop:
 
                     assistant, stop_reason, stream_error = None, "stop", None
                     async for event in self._model_turn(
-                        context, config, signal, effort_ceiling=effort_ceiling
+                        context,
+                        config,
+                        signal,
+                        effort_ceiling=effort_ceiling,
+                        context_tokens_hint=(
+                            run_context_tokens
+                            if run_context_tokens is not None
+                            else self._host_context_tokens_hint(config)
+                        ),
                     ):
                         if isinstance(event, _ModelTurnResult):
                             assistant, stop_reason, stream_error = (
@@ -541,6 +559,11 @@ class AgentLoop:
                             yield event
                     if assistant is None:
                         raise RuntimeError("model turn produced no assistant message")
+                    if assistant.usage is not None and assistant.usage.context_tokens:
+                        # Only a REPORTED count advances the hint: a wire that
+                        # omits it must not blank a figure the previous call
+                        # (or the host's seed) supplied.
+                        run_context_tokens = int(assistant.usage.context_tokens)
                     context.messages.append(assistant)
                     new_messages.append(assistant)
 
@@ -804,18 +827,35 @@ class AgentLoop:
 
     # (``_abortable_stream`` is a module-level helper; see below the class.)
 
+    @staticmethod
+    def _host_context_tokens_hint(config: LoopConfig) -> int | None:
+        """The host's cross-turn context-size seed, or ``None`` when it has
+        none — a zero is treated as "nothing reported" because a provider
+        never reports an empty context, so 0 here can only be an unset
+        default, not a deliberate suppression (that is ``ChatRequest``'s
+        explicit ``0``, which only the session's own direct calls set)."""
+        reader = config.get_context_tokens_hint
+        if reader is None:
+            return None
+        hint = reader()
+        return int(hint) if hint else None
+
     async def _model_turn(
         self,
         context: LoopContext,
         config: LoopConfig,
         signal: AbortSignal | None,
         effort_ceiling: str | None = None,
+        context_tokens_hint: int | None = None,
     ) -> AsyncIterator[AgentEvent | _ModelTurnResult]:
         """One provider call: build the request, stream it, assemble the
         assistant message, emitting message_start/update/end events.
 
         The model is resolved HERE, per call, not once per run — see
-        :meth:`_current_model`.
+        :meth:`_current_model`. ``context_tokens_hint`` is stamped onto the
+        request by THIS loop, the owner of the conversation the call belongs
+        to — never remembered on the shared stream fn, where a subagent's
+        registration would overwrite the parent's (review F8).
         """
         shaped = list(context.messages)
         if config.transform_context is not None:
@@ -868,6 +908,7 @@ class AgentLoop:
                 messages=list(converted),
                 tools=list(context.tools),
                 effort_ceiling=effort_ceiling,
+                context_tokens_hint=context_tokens_hint,
             )
             stream = _abortable_stream(config.stream_fn(request, signal), signal)
             async for event in stream:
--- a/local_operator/harness/subagent.py
+++ b/local_operator/harness/subagent.py
@@ -895,6 +895,12 @@ def _accumulate_usage(job: Any, usage: "Usage | None") -> None:
     total.output_tokens += usage.output_tokens
     total.cache_read_tokens += usage.cache_read_tokens
     total.cache_write_tokens += usage.cache_write_tokens
+    # The TTL split of the write count folds exactly where the write count
+    # does (they are subsets of it; see ``Usage.cache_write_1h_tokens``), so
+    # the job aggregate can price the two rates apart the moment a reader
+    # needs to.
+    total.cache_write_5m_tokens += usage.cache_write_5m_tokens
+    total.cache_write_1h_tokens += usage.cache_write_1h_tokens
     # Child failover can mix provider receipts and table-priced calls. Preserve
     # every original call so the TUI can price each one independently instead of
     # treating one receipt as authoritative for the aggregate token buckets.
--- a/local_operator/harness/types.py
+++ b/local_operator/harness/types.py
@@ -224,6 +224,18 @@ class Usage(BaseModel):
     output_tokens: int = 0
     cache_read_tokens: int = 0
     cache_write_tokens: int = 0
+    # The TTL split of ``cache_write_tokens`` when the provider reports one
+    # (Anthropic's ``usage.cache_creation.ephemeral_5m_input_tokens`` /
+    # ``ephemeral_1h_input_tokens``). SUBSETS of ``cache_write_tokens``, never
+    # added on top — the docs state ``cache_creation_input_tokens`` equals their
+    # sum, and every existing consumer of ``cache_write_tokens`` stays correct.
+    # They exist because the two TTLs are priced differently (1.25× vs 2× base):
+    # once the Anthropic client starts writing 1h entries on large contexts,
+    # analytics needs to tell the two apart to know whether the trade paid off.
+    # Both stay 0 on providers without a TTL split, and on an Anthropic response
+    # that omits the ``cache_creation`` object (older API versions).
+    cache_write_5m_tokens: int = 0
+    cache_write_1h_tokens: int = 0
     context_tokens: int | None = None  # provider-reported full context size if given
     # The reasoning/thinking slice of ``output_tokens`` when the provider
     # breaks it out (OpenAI ``output_tokens_details.reasoning_tokens``). Kept
@@ -1530,6 +1542,18 @@ class ChatRequest(BaseModel):
     # caches. Session hosts populate it once from their session id; keeping it
     # on the request lets retries and fallback clones preserve the same value.
     prompt_cache_key: str | None = None
+    #: The provider-reported size of this session's context on its LAST call
+    #: (``Usage.context_tokens``), when the host knows it. A HINT, not wire
+    #: content: the Anthropic client reads it to decide whether the request is
+    #: large enough for the 1-hour prompt-cache TTL (see
+    #: ``AnthropicClient._cache_ttl_for``). The client cannot measure this
+    #: itself without a tokenizer, and the provider's own count from the
+    #: previous turn is the most accurate figure anyone has. ``None`` on a
+    #: session's first call and on paths that never saw a usage event (a fork's
+    #: first request, one-shot errands); the client then falls back to a byte
+    #: estimate of the serialized body. Set by ``SessionStreamFn.__call__``
+    #: from the last usage it recorded; retries and fallback clones keep it.
+    context_tokens_hint: int | None = None
     #: This call's output has NOT been shown to anyone yet, so a failed attempt
     #: may be discarded and retried whole.
     #:
--- a/local_operator/harness/types.py
+++ b/local_operator/harness/types.py
@@ -1446,18 +1446,22 @@ class LoopConfig(BaseModel):
     has_pending_fork: Callable[[], bool] | None = Field(default=None, exclude=True)
 
     # The provider-reported context size (``Usage.context_tokens``) of this
-    # conversation's LAST call, re-read immediately before every provider call
-    # and stamped onto the request as ``ChatRequest.context_tokens_hint``.
-    # Per-CONVERSATION memory, not per-stream-fn: subagents share the parent's
-    # stream fn (one httpx pool, one failover cascade), so a hint remembered
-    # on the stream fn itself would let a child's tiny first request inherit
-    # the parent's 300k count and pay 2x write rates on a fresh ~10k prefix —
-    # and let the parent's post-child request go out at 5m on its large
-    # context, the exact expiry this hint exists to dodge. The loop asks the
-    # HOST (which owns the conversation), exactly as it does for the model and
-    # the system blocks. ``None`` (no callback, or nothing reported yet) means
-    # no hint and the client's own byte estimate decides. Read in
-    # ``SessionStreamFn.__call__``; see ``ChatRequest.context_tokens_hint``.
+    # conversation's last call BEFORE this run — the cross-turn seed. The loop
+    # reads it for the run's first request and stamps it as
+    # ``ChatRequest.context_tokens_hint``; from the second request on, the
+    # count the previous call of the SAME run reported wins, because a long
+    # tool loop grows past the TTL threshold long before the host's figure is
+    # next refreshed. Per-CONVERSATION, stamped per REQUEST by the loop that
+    # owns the call — never remembered on the stream fn: subagents share the
+    # parent's stream fn (one httpx pool, one failover cascade), so a hint
+    # held there is last-writer-wins between the parent and every child, and
+    # a child's construction would silently downgrade the parent's next
+    # request to 5m on its large context (the exact expiry this hint exists
+    # to dodge) while the child's tiny first request inherits the parent's
+    # 300k count and pays 2x write rates on a fresh ~10k prefix. The loop asks
+    # the HOST (which owns the conversation), exactly as it does for the
+    # model and the system blocks. ``None`` (no callback, or nothing reported
+    # yet) means no hint and the client's own byte estimate decides.
     get_context_tokens_hint: Callable[[], int | None] | None = Field(default=None, exclude=True)
 
     interrupt_mode: Literal["immediate", "wait"] = "wait"
@@ -1566,12 +1570,15 @@ class ChatRequest(BaseModel):
     #: previous turn is the most accurate figure anyone has. ``None`` on a
     #: session's first call and on paths that never saw a usage event (a fork's
     #: first request, one-shot errands); the client then falls back to a byte
-    #: estimate of the serialized body. Stamped by ``SessionStreamFn.__call__``
-    #: from the conversation owner's registered reader (see
-    #: ``SessionStreamFn.set_context_tokens_hint`` — the memory is
-    #: per-conversation precisely because subagents share one stream fn);
-    #: retries and fallback clones keep it, and a caller's EXPLICIT value —
-    #: including ``0``, which suppresses the hint — always wins.
+    #: estimate of the serialized body. Stamped by the OWNER of the
+    #: conversation the call belongs to — the harness loop for turn calls
+    #: (``LoopConfig.get_context_tokens_hint`` seeds it, then the run's own
+    #: usage events advance it) and the session for its direct calls (asides,
+    #: the compaction advisor) — never by the shared stream fn, which merely
+    #: passes through what the request carries: subagents share one stream
+    #: fn, so any memory held there is last-writer-wins across conversations.
+    #: Retries and fallback clones keep it, and an EXPLICIT ``0`` suppresses
+    #: the hint (a one-shot prompt that is a fresh write-once prefix).
     context_tokens_hint: int | None = None
     #: This call's output has NOT been shown to anyone yet, so a failed attempt
     #: may be discarded and retried whole.
--- a/local_operator/harness/types.py
+++ b/local_operator/harness/types.py
@@ -1445,6 +1445,21 @@ class LoopConfig(BaseModel):
     # steering alone.
     has_pending_fork: Callable[[], bool] | None = Field(default=None, exclude=True)
 
+    # The provider-reported context size (``Usage.context_tokens``) of this
+    # conversation's LAST call, re-read immediately before every provider call
+    # and stamped onto the request as ``ChatRequest.context_tokens_hint``.
+    # Per-CONVERSATION memory, not per-stream-fn: subagents share the parent's
+    # stream fn (one httpx pool, one failover cascade), so a hint remembered
+    # on the stream fn itself would let a child's tiny first request inherit
+    # the parent's 300k count and pay 2x write rates on a fresh ~10k prefix —
+    # and let the parent's post-child request go out at 5m on its large
+    # context, the exact expiry this hint exists to dodge. The loop asks the
+    # HOST (which owns the conversation), exactly as it does for the model and
+    # the system blocks. ``None`` (no callback, or nothing reported yet) means
+    # no hint and the client's own byte estimate decides. Read in
+    # ``SessionStreamFn.__call__``; see ``ChatRequest.context_tokens_hint``.
+    get_context_tokens_hint: Callable[[], int | None] | None = Field(default=None, exclude=True)
+
     interrupt_mode: Literal["immediate", "wait"] = "wait"
     # Epoch-ms deadline for the whole run, if any.
     deadline: float | None = None
@@ -1551,8 +1566,12 @@ class ChatRequest(BaseModel):
     #: previous turn is the most accurate figure anyone has. ``None`` on a
     #: session's first call and on paths that never saw a usage event (a fork's
     #: first request, one-shot errands); the client then falls back to a byte
-    #: estimate of the serialized body. Set by ``SessionStreamFn.__call__``
-    #: from the last usage it recorded; retries and fallback clones keep it.
+    #: estimate of the serialized body. Stamped by ``SessionStreamFn.__call__``
+    #: from the conversation owner's registered reader (see
+    #: ``SessionStreamFn.set_context_tokens_hint`` — the memory is
+    #: per-conversation precisely because subagents share one stream fn);
+    #: retries and fallback clones keep it, and a caller's EXPLICIT value —
+    #: including ``0``, which suppresses the hint — always wins.
     context_tokens_hint: int | None = None
     #: This call's output has NOT been shown to anyone yet, so a failed attempt
     #: may be discarded and retried whole.
--- a/local_operator/model/configure.py
+++ b/local_operator/model/configure.py
@@ -1566,6 +1566,34 @@ def _openai_api_mode(settings: Mapping[str, Any] | None) -> str:
     return "chat_completions" if configured == "chat_completions" else "responses"
 
 
+#: Default for ``providers.anthropic.cache_ttl_1h_min_context_tokens`` when the
+#: settings mapping lacks the key (an old config file, a test passing ``{}``).
+#: Mirrors ``DEFAULT_CONFIG`` in ``config.py`` rather than importing it: the
+#: ``config`` module is the CLI's, and ``_openai_api_mode`` above already takes
+#: the same restate-the-default stance for the sibling key.
+ANTHROPIC_CACHE_TTL_1H_MIN_CONTEXT_TOKENS = 150_000
+
+
+def _anthropic_cache_ttl_1h_min_context_tokens(settings: Mapping[str, Any] | None) -> int:
+    """Resolve the context size above which Anthropic requests use the 1h TTL.
+
+    Same shape as ``_openai_api_mode``: a missing ``providers`` block (old
+    config files) or a malformed value falls back to the default rather than
+    silently disabling the feature, because the failure mode of "disabled" is
+    a bill that quietly grew rather than an error anyone sees. Only an explicit
+    non-negative integer is honoured; ``0`` is the documented off switch.
+    """
+    providers = settings.get("providers") if isinstance(settings, Mapping) else None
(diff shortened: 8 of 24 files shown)


damianvtran:
### Agent review — round 1
- Scope: 093cf9f4..2cf0d11ef6a0cc5bebd74d0acb392edf4ef1c10a
- Mode: Comprehensive
- Delivery contract: reduce 5m-TTL prefix rewrites on large Anthropic contexts without changing any sub-threshold request body
- Applicable risk surfaces: provider wire compatibility, cost accounting, config plumbing, analytics migration
- Verdict: Changes requested

Verified on a detached worktree of `2cf0d11e`: `flake8 .` clean, `black --check` 769 unchanged, `isort --check` clean, `pyright` 0 errors/0 warnings; `pytest tests/unit/providers tests/unit/model tests/unit/analytics tests/unit/test_config.py tests/unit/test_settings_io.py -n 4` → 1300 passed; `pytest tests/unit/session tests/unit/harness -n 4` → 1045 passed. The TTL rule itself is sound: one value decided in `_cache_ttl_for` is applied to every marker (`_system_blocks` and `_message_cache_breakpoints`), so a body can never mix 1h and 5m; the 4-marker budget logic is untouched; the system-only path (no message targets → early return at `clients.py:1986`) leaves the already-rendered system markers with the same TTL. Dict key order in `_build_body` is preserved by the placeholder-then-reassign pattern, so the "byte-identical below threshold" claim holds. Migration is guarded by `PRAGMA table_info` and the insert plan is rebuilt from present columns, so it is idempotent and lossless; rollups are unchanged and stay consistent with the ledger.

| --- | --- | --- | --- |
| F1 | major | hint plumbing / cost | `_last_context_tokens` lives on `SessionStreamFn` (`configure.py:1762`, written at `:3485`, stamped at `:3412-3416`), but subagents are built with `stream_fn=parent_session._stream_fn` (`harness/subagent.py:1341`), so one scalar is shared by the parent and every child it spawns. Consequences: (a) a child's first request inherits the parent's 300k count and sends 1h markers (2× writes) on a ~10k fresh prefix; (b) after the children run, the parent's next request — the exact "resume after waiting on subagents" moment this feature targets — is stamped with the last child's small count and goes out with 5m markers on its 300k context; (c) when the parent keeps working while `task` children run in the background, the hint thrashes on every call. Isolated errands were excluded for precisely this reason (`:1759-1760`) but the same contamination via shared stream fns was missed, and `test_session_stream_feeds_last_context_into_anthropic_ttl` cannot see it (one stream fn, one session). Fix: key the memory per session rather than per stream fn — e.g. have the harness loop stamp `context_tokens_hint` from its own `Session._last_usage.context_tokens` (the loop already receives per-session callables like `get_model`), or keep a `dict` keyed on a per-session identity inside `SessionStreamFn`. Add a test with a parent and a child request interleaved on one stream fn. |
| F2 | minor | cost | The compaction summary errand (`session/session.py:7302-7310`, `_one_shot_complete`) is not isolated, so it inherits the pre-compaction hint (≥150k by construction) and sends its whole transcript-sized prompt with 1h markers. That prompt has a unique system block, is written once and never replayed (except a stall retry), so the 2× rate buys nothing: on a 300k Opus compaction that is ~0.75× base × 300k ≈ several dollars extra per compaction. Pass `context_tokens_hint=0` on that request — `__call__` keeps a deliberately-set hint and `_cache_ttl_for` treats 0 as below threshold. Secondary: the first post-compaction turn then inherits the summary call's large `context_tokens` and goes 1h once before flipping back; harmless direction, worth a comment. |
| F3 | minor | hint fallback accuracy | `_cache_ttl_for` (`clients.py:1873-1875`) divides the serialized body by 4, and `_message_blocks` (`:1924-1934`) inlines base64 image data. A single ~1 MB screenshot is ~1.3 M base64 chars → ~330k "tokens" vs ≤~1.6k actually billed (images are capped at ~1.15 MP), so any first call / fork first request / resumed session carrying an image flips to 1h regardless of real size. Exclude `source.data` from the estimate and add a flat per-image constant (e.g. 1.6k). |
| F4 | minor | usage aggregation consistency | `harness/jobs.py:135-136` folds the new 5m/1h split, but `harness/subagent.py:894-897` (`_accumulate_usage`) and `session/frontend_state.py:1689-1697` do not, so `job.usage.cache_write_1h_tokens` is the FIRST child call's value only and the frontend sum is always 0. Nothing reads them yet (the ledger is fed per call), but it is the same "subset of `cache_write_tokens`" contract and will silently mis-report as soon as a TUI panel or cost split reads the aggregate. Add the two fields to both sums. |
| F5 | minor | analytics measurability | `cache_write_1h_tokens` is write-only: it is not in `_ROLLUP_MEASURE_COLUMNS` (`store.py:264-274`) and no SELECT/`_aggregate_from_row` projects it. With `DEFAULT_RETENTION_DAYS = 90` on `calls`, the stated purpose ("judge whether the 1h trade paid for itself after the fact") is answerable only inside the ledger window and only via raw SQL. Either add it as a rollup measure + read column now, or say in the column comment that it is a 90-day raw-ledger diagnostic. |
| F6 | nit | test strength | `test_session_stream_feeds_last_context_into_anthropic_ttl` (`tests/unit/model/test_configure.py:3217`) claims the isolated call "must neither read nor move the hint", but the mock reports 200k for every call including the isolated one, so removing the `not request.isolated` guard at `configure.py:3481` would not fail the test. Have the isolated response report a small context and assert call 3 is still 1h. |
| F7 | nit | release hygiene | `pyproject.toml` jumps 0.44.27 → 0.44.30 with 28/29 pre-assigned to sibling PRs; fine if the merger re-sequences as the body says, but the commit message should carry that note too so `git log` readers do not see a gap. |

#### Follow-up observations
- Testing Details: the timed pair (1h read after 370 s vs 5m rewrite, same 6627-token prefix, same account) does demonstrate the TTL mechanism and that OAuth accepts `ttl: "1h"` without an extra beta header; the client-path capture shows three 1h markers on the wire with the parsed split. The arithmetic in the economics claim is internally consistent (89.5M × 1.25 ≈ 112M; 14.7M × 0.75 ≈ 11M). What the evidence does not show — and the body says so — is a live ≥150k request; that is acceptable since the wire shape is identical.
- Pricing: writes are still priced at the 5m rate everywhere (`calculate_cost` / `cost_for_usage`); the PR body and the `CallSnapshot` comment both say so honestly, and because the 1h count is a subset nothing double-counts.
- Config plumbing: the default reaches `client_for_spec` on every construction path I traced — TUI/exec via `session_factory:1474`, server via `server/utils/operator.py:571` (`settings=None` → 150k via the reader), evaluation via `provider_client.py:269`, subagents via the shared parent stream fn, failover via `_client_for`. `AnthropicClient` is constructed only at `clients.py:2570`.
- The 1h→5m flip after compaction does not lose hits: Anthropic keys entries on content hash, and the mixed-TTL billing positions are defined relative to "highest cache hit" independent of the marker TTL.

damianvtran:
- F1 (major) — Fixed in `58974e4b`: the hint memory moved off `SessionStreamFn` to the conversation owner — `Session._context_tokens_hint` (seeded from the transcript's last usage, advanced at the turn boundary) registers a reader via `set_context_tokens_hint` (mirroring the notice/route bridges), passed to the loop as `LoopConfig.get_context_tokens_hint`; `__call__` stamps it only on non-isolated requests. New test drives a parent (300k hint) and a child (no hint) through ONE shared stream fn: child's first request stays 5m on its own small body, parent's post-child request stays 1h on its own large hint.
- F2 (minor) — Fixed in `58974e4b`: `_one_shot_complete` now sets `context_tokens_hint=0` — a deliberate hint the stream fn keeps and the client treats as below any threshold; the comment documents why a write-once prompt must not inherit the pre-compaction count.
- F3 (minor) — Fixed in `c92b199f`: new `AnthropicClient._estimate_context_tokens` swaps every base64 `source.data` for an empty string while counting, and adds a documented 1.6k-token floor per image (pixel-area billing, ~1.15 MP cap); located by shape so tool-result-nested images are covered. Test proves a 1.4M-char image no longer flips a small body to 1h and that the per-image allowance is still counted.
- F4 (minor) — Fixed in `67d02901`: `_accumulate_usage` (harness/subagent.py) and `_aggregate_usage` (session/frontend_state.py) now fold `cache_write_5m_tokens`/`cache_write_1h_tokens` exactly where `cache_write_tokens` folds; tests on both.
- F5 (minor) — Rejected for this PR / deferred with reason: adding the column to `usage_daily`/`usage_monthly` is NOT a mirror of the calls migration — the rollup tables have no ALTER path (`CREATE TABLE IF NOT EXISTS` cannot add a column to an existing table, unlike `calls`' `_MIGRATION_COLUMNS`), so it is a schema change of its own (new migration machinery + `_ROLLUP_READ_COLUMNS`/`UsagePeriod` projection + report surface) with no consumer in this PR. The column comment now states the 90-day raw-ledger diagnostic scope explicitly (in `67d02901`), and the follow-up is tracked in the PR body under Follow-up.
- F6 (nit) — Fixed in `58974e4b`: the mock now reports 5k for the isolated call vs 200k for the turn, and the test asserts call 3 is still 1h — removing the guard would move the hint to 5k and fail the test.
- F7 (nit) — Addressed: this remediation pushes no version change; the merger-resequences note stays in the PR body. Acknowledged as legitimate; `git log` readers of the merge will see the body note via the squash/merge message. (No commit-level change made — the version jump is unchanged from round 1.)

Verification (whole tree, as CI): `flake8 .` clean; `black --check` 769 unchanged; `isort --check` clean; `pyright` 0 errors/0 warnings. `pytest tests/unit/providers tests/unit/model tests/unit/session tests/unit/harness tests/unit/analytics tests/unit/test_config.py tests/unit/test_settings_io.py -n 4` → 2349 passed (one pre-existing-flake-shaped failure of `test_live_progress_never_schedules_the_roster_writer` appeared once under `-n 4` and never reproduced in 10 subsequent runs including serial; it passes on the round-1 head too).

damianvtran:
### Agent review — round 2
- Scope: 093cf9f4..67d0290180bb4a34b828fefd2898ad7b970a599d
- Delivery contract: reduce 5m-TTL prefix rewrites on large Anthropic contexts without changing any sub-threshold request body
- Delta reviewed: 2cf0d11e..67d02901
- Applicable risk surfaces: provider wire compatibility, cost accounting, config plumbing, analytics migration
- Verdict: Changes requested

Verified on a detached worktree of `67d02901`: `flake8 .` clean, `black --check` 769 unchanged, `isort --check` clean, `pyright` 0 errors/0 warnings; `pytest tests/unit/providers tests/unit/model tests/unit/session tests/unit/harness tests/unit/analytics -n 4` → 2223 passed (no flake on this run). The `0`-wins rule survives every `model_copy` clone on the path (`configure.py:3413` effort, `:3431` prompt_cache_key, `failover.py:1817` fallback model) since `model_copy` preserves unlisted fields.

| --- | --- | --- | --- |
| F8 | major | hint plumbing / cost (F1 not converged) | The memory moved to the session, but the READER slot is still one scalar on the shared stream fn: `set_context_tokens_hint` (`configure.py:1776-1789`) is last-writer-wins, and `Session.__init__` registers unconditionally (`session.py:1456-1463`). `harness/subagent.py:1340` builds every child with `stream_fn=parent_session._stream_fn`, so the moment a `task` child is constructed the stream fn reads the CHILD's `_context_tokens_hint` for every call — the parent's included — and keeps doing so after the child is disposed (nothing ever restores the parent's reader). **Reproduced** with the real `create_stream_fn` + `MockTransport` at a 150k threshold: parent request with hint 300k → every marker `ttl: 1h`; construct a second `Session` on the same stream fn and set its hint to 10k; the parent's identical request → plain `ephemeral` (5m) while `parent._context_tokens_hint` is still 300000. That is round-1 contamination (b) unchanged, and (c) thrash whenever the parent works while children run; it now also persists past the child's lifetime. `test_session_stream_hint_is_per_conversation_not_per_stream_fn` is green only because it calls `stream.set_context_tokens_hint(lambda: parent_hint["value"])` by hand before the parent's post-child request — a step no harness code performs, so the test asserts the API's capability rather than the integration. Contributing cause: `LoopConfig.get_context_tokens_hint` (`types.py:1461`, wired at `session.py:4771`) is dead — `grep get_context_tokens_hint local_operator/harness/` finds only the field; the loop never reads it. Fix: stamp per REQUEST by its owner instead of registering a reader on the shared pipe — in `loop.py:865` pass `context_tokens_hint=config.get_context_tokens_hint()` (when set) into the `ChatRequest`, stamp `self._context_tokens_hint` explicitly on the session's own direct calls (aside `session.py:7425`, advisor `:7503`), and delete the registrar/`set_context_tokens_hint` bridge. Then rewrite the shared-stream-fn test to drive two `Session` objects (or two `LoopConfig`s) through one stream fn WITHOUT any manual re-registration. |
| F9 | major | cost (regression vs round 1) | The hint is advanced only at the turn boundary (`session.py:4864-4876`, after the `async for event` loop over the whole run) — round 1 updated it per provider call in `_record_stream`. Inside one turn's tool loop every call is stamped with the PREVIOUS turn's count, and a stale non-`None` hint beats the byte estimate (`configure.py:3438`). A turn that ends at 140k and then grows past 150k across 30 tool calls stays 5m for all of them; a subagent (one turn for its whole life, e.g. the 48-request/1.5M-token review child cited in `subagent.py:1386`) never has a hint at all. The mid-turn hook at `session.py:5731-5735` updates `_last_usage` but not the hint, and is gated on mid-turn compaction being enabled. Fix falls out of F8: once the loop stamps the request itself, have it prefer the latest `Usage.context_tokens` it has seen in THIS run over `config.get_context_tokens_hint()` (or update `self._context_tokens_hint` wherever `_last_usage` is updated, unconditionally). Add a test where call N of a multi-call turn reports ≥ threshold and call N+1 of the same turn goes 1h. |

#### Verified
- **F1 (partial)**: memory is on `Session._context_tokens_hint` (seeded `session.py:1706` from `_last_reported_usage` over `usages_since_compaction`, so a resumed post-compaction session seeds from its post-compaction size, not a stale figure); isolated naming calls return before the stamp (`configure.py:3346-3360`). Direction (a) — a child's FIRST request stays 5m — does hold, but only because the child's registration happens to be the most recent; see F8 for the other direction.
- **F2**: `_one_shot_complete` sets `context_tokens_hint=0` (`session.py:7358`); `__call__` keys on `is None` so 0 survives; `_cache_ttl_for` compares `0 >= threshold` → 5m. Confirmed.
- **F3**: `_estimate_context_tokens` (`clients.py:1835-1875`) blanks every `source.type == "base64"` `data` by shape walk and adds a documented 1.6k floor per image; test proves a 1.4M-char image stays 5m at a 50k bar and that two images tip a 3k bar. No `document`/PDF block is emitted by this client, so the image-only walk is complete.
- **F4**: `_accumulate_usage` (`subagent.py:898-903`) and `_aggregate_usage` (`frontend_state.py:1694-1695`) fold the split; `jobs.py:134-136` already did. Tests on both.
- **F5 rejection holds**: `usage_daily`/`usage_monthly` are `CREATE TABLE IF NOT EXISTS` only (`store.py:174,190`); `_MIGRATION_COLUMNS`/`ALTER TABLE ADD COLUMN` (`:228-257`) apply to `calls` alone, so adding the split to the rollups is genuinely new migration machinery. Column comment now states the raw-ledger scope.
- **F6**: mock reports 5k for the isolated call vs 200k for the turn; call 3 asserted 1h. Dropping the isolation guard would fail the test.

No blocker. The two majors share one root cause and one fix (per-request stamping by the owner, driven by the loop); a third round should be short.

damianvtran:
- Scope: 67d02901..aaad706d (new head `aaad706d`)

- F8 (major) — Fixed in aaad706d: the hint is now stamped PER REQUEST by its owner and the shared-pipe reader is gone. `AgentLoop._run` reads `LoopConfig.get_context_tokens_hint` (previously dead) as the run's seed and passes `context_tokens_hint=` into the `ChatRequest` built in `_model_turn`; `Session.complete_aside` and `advise_compaction` stamp `self._context_tokens_hint` on their own requests; `_one_shot_complete` keeps its explicit `0`. `SessionStreamFn.set_context_tokens_hint`, `_context_tokens_hint_reader`, the stamping block in `__call__` and the registration in `Session.__init__` are deleted — `__call__` now only passes through what the request carries, and the isolated branch needs no guard because an errand's request is built without a hint. `test_session_stream_hint_is_per_conversation_not_per_stream_fn` is rewritten to drive TWO real `Session` objects through ONE real `create_stream_fn` stream with no manual hint handling: parent works up 300k → child constructed on the same stream fn, runs its turn at 5m on its own 10k body → parent's next request 1h on its own 300k (your round-2 reproduction, now asserted in both directions; `parent._context_tokens_hint` stays 300k, `child._context_tokens_hint` is 10k).
- F9 (major) — Fixed in aaad706d: the loop keeps a run-local `run_context_tokens` updated from each assembled assistant's `usage.context_tokens` and prefers it over the host seed for every later request of the same run — so call N+1 goes 1h as soon as call N reports ≥ threshold, and a subagent learns its size from its own first call. On the session side `_note_usage` moves `_last_usage` and `_context_tokens_hint` together at BOTH the post-run scan and the mid-turn boundary hook, and the boundary update now runs ahead of (not gated on) `mid_turn_enabled`; a call that omits `context_tokens` keeps the previous count rather than blanking it. Tests: `test_session_stream_feeds_last_context_into_anthropic_ttl` now drives a real `Session` through a tool-loop turn — call 1 (no hint) 5m, provider reports 200k + `tool_use`, call 2 of the SAME turn 1h on every marker, isolated errand unstamped/5m, next turn 1h; plus `TestTheContextHintIsStampedPerRequestByTheLoop` (seed→in-run takeover `[140k,160k,170k]`, no seed → `[None,160k,170k]`, missing count keeps last), an aside-stamping test, and a `mid_turn_enabled=False` boundary test (`[None, 200k, 210k]` stamped, tool observes the hint advance mid-turn). Reverting only the loop's `context_tokens_hint=` line fails all six.

F1–F6 re-verified intact after the refactor: F1 memory still on `Session._context_tokens_hint`, seeded from `usages_since_compaction`; F2 `_one_shot_complete` still pins `0` and the client treats 0 as below any threshold (the `0`-wins guard in `__call__` is moot now that nothing overwrites a request's hint); F3 image-exclusion estimate untouched; F4 aggregate folds untouched; F5 rejection stands; F6 the isolated errand is still proven excluded (the mock reports 5k for it vs 200k for the turn and the next turn call goes 1h).

Version stays at 0.44.30 — the merger re-sequences against the sibling PRs.

Gates at `aaad706d`: `flake8 .` clean · `black --check .` 769 unchanged · `isort --check .` clean · `pyright` 0 errors/0 warnings · `pytest tests/unit/providers tests/unit/model tests/unit/session tests/unit/harness tests/unit/analytics -n 4` → **2228 passed** · `tests/unit/test_fork.py tests/unit/compaction` → 274 passed. PR body's Testing Details updated for the new design and tests.

damianvtran:
### Agent review — round 3
- Scope: 093cf9f4..aaad706dc6511f0920b6629a028166fc715ca631
- Delivery contract: reduce 5m-TTL prefix rewrites on large Anthropic contexts without changing any sub-threshold request body
- Delta reviewed: 67d02901..aaad706d
- Applicable risk surfaces: provider wire compatibility, cost accounting, config plumbing, analytics migration
- Verdict: Approve

F8 and F9 are converged. No blocker or major remains; this round is **TERMINAL** — the items below are follow-ups, not a reason for another round.

| ID | Severity | Contract or risk | Evidence and suggested action |
| --- | --- | --- | --- |
| F10 | minor | cost (post-compaction overpay, one request) | Neither the run-local `run_context_tokens` (`loop.py:504-566`) nor the session's `_context_tokens_hint` is reset when the context is replaced: the only writer is `_note_usage` (`session.py:4310-4312`), and `_finish_compaction`/`_apply_pending_compaction` leave `_last_usage`/the hint alone, while the loop's mid-run replacement branch (`loop.py:707-724`) keeps the pre-compaction count. So the first request after a compaction — mid-run (the replacement context) or the next turn — is stamped with the OLD figure (≥ threshold by construction) and writes the brand-new compacted prefix at the 1h rate (2.0× base vs 1.25× for 5m; ≈0.75×base×P once per compaction, ~$1 on a 100k Opus prefix). The next call's report self-corrects, so it is one request and in the harmless direction, but the PR body's "dropping back to 5m after compaction is fine" does not describe what actually happens on that first call. Fix is one line each: set `run_context_tokens = None` (and skip the host seed) when `on_turn_end` returns a list, and clear `_context_tokens_hint` in `_finish_compaction`. Follow-up, not a round. |

#### Verified
On a detached worktree of `aaad706d` (never touching `~/local-operator`):
- **Gates** (AGENTS.md recipe, `python -m` / `uvx`, not the venv console scripts): `flake8 .` clean · `black --check .` 769 unchanged · `isort --check .` clean · `pyright --pythonpath .venv/bin/python .` 0 errors / 0 warnings. `pytest tests/unit/providers tests/unit/model tests/unit/session tests/unit/harness tests/unit/analytics tests/unit/compaction -n 4` → **2427 passed**.
- **(1) No per-conversation state on the stream fn**: `grep -rn "set_context_tokens_hint\|_context_tokens_hint_reader" local_operator tests` → nothing. `SessionStreamFn.__call__` no longer touches `context_tokens_hint` at all (the only remaining references in `configure.py` are comments at `:1757` and `:3333`); `Session.__init__` registers no bridge (`session.py:1456-1462`).
- **(2) Loop seeds and prefers run-local**: `_run` stamps `run_context_tokens if not None else _host_context_tokens_hint(config)` (`loop.py:546-550`), advances `run_context_tokens` only on a truthy reported count (`:562-566`), and `_model_turn` passes it into the `ChatRequest` (`:911`). `LoopConfig.get_context_tokens_hint` is now read (`:837`), no longer dead.
- **(3) Direct calls stamp themselves**: `complete_aside` (`session.py:7446`) and `advise_compaction` (`:7529`) pass `context_tokens_hint=self._context_tokens_hint`; `_one_shot_complete` keeps the explicit `0` (`:7367`).
- **(4) `_note_usage` moves both together**: single writer for `_last_usage` and the hint (`:4294-4313`), called from the post-run scan (`:4883`) and from the boundary hook at `:5732` — which is AFTER `_persist_progress`/`_drain_pending_fork` and BEFORE the `settings.enabled`/`mid_turn_enabled` early return at `:5741`, so it runs unconditionally. Hint only advances on a reported count; a wire omitting it keeps the previous figure.
- **(5) Integration tests are real**: both tests build two/one real `Session` objects via `Session(model=…, stream_fn=stream, tools=[echo], transcript=Transcript(...))` on one real `create_stream_fn` + `MockTransport`, drive them with `session.prompt(...)`/`complete_once(...)`, and never assign or read the hint except in post-hoc asserts (`test_configure.py:3358,3412,3417,3425`). Contamination (a) child's first request 5m on its own 10k body, and (b) parent's post-child request 1h on its own 300k, are both asserted; the in-turn test asserts call 2 of the SAME turn goes 1h after call 1 reports 200k, the isolated errand carries no hint (5m, its 5k report never enters), and turn 2 call 1 is 1h.
- **(6) Negative check reproduced by me**: commenting out the single `context_tokens_hint=context_tokens_hint,` line in `_model_turn` → **6 failed** (3× `TestTheContextHintIsStampedPerRequestByTheLoop`, both `test_configure.py` integration tests, and `test_context_hint_advances_at_every_boundary_even_with_mid_turn_compaction_off`), 291 passed. Restored; worktree clean.
- **`None` vs `0`**: `_host_context_tokens_hint` maps a falsy seed to `None` (byte estimate) — correct, since a provider never reports 0 and `_context_tokens_hint` is only ever assigned truthy values; the explicit `0` on `_one_shot_complete` reaches `_cache_ttl_for` untouched (`clients.py:1936-1938`, `0 >= threshold` → 5m).
- **Isolated/naming path**: `complete_once` builds its request without a hint; the isolated branch in `__call__` (`configure.py:3328-3340`) needs no guard because nothing stamps after it any more. Proven by the 5k-vs-200k mock.
- **Subagent first call**: `subagent.py:1155` builds a fresh `Transcript`, so `_last_usage`/hint seed `None` → call 1 unstamped (byte estimate), and the loop learns from call 1's report for call 2+ (`test_without_a_seed_the_first_call_is_unstamped` → `[None, 160k, 170k]`).
- **F1–F6** unchanged from round 2 and still hold after the refactor (F3 image-exclusion estimate, F4 aggregate folds, F5 rejection untouched by this delta).

#### Follow-up observations
- F10 above (post-compaction one-request overpay, and the PR-body sentence about "dropping back to 5m after compaction" being slightly optimistic for that first call) — worth a two-line fix in a follow-up PR, together with the same reset for `_last_usage` if the compaction trigger math wants it.
- `_host_context_tokens_hint` is a `@staticmethod` on `AgentLoop` that only reads `config`; a module-level helper would match the `_abortable_stream` convention noted at `loop.py:828`. Nit only.

damianvtran:
### Rebase note

Rebased `feat/anthropic-1h-cache-ttl` onto `origin/main` @ f82d5527 (#538): old head aaad706d → new head bf6be177.

- The PR's 5 commits replayed cleanly with no source conflicts (`git range-diff` shows every commit `=` except the first, which only lost its now-redundant `pyproject.toml` version hunk). The PR diff against main is intent-identical to the round-3 reviewed diff — only `session.py` hunk line offsets moved because of #529/#530/#531/#538 landing.
- Version bump split into its own commit: `chore(release): bump version to 0.44.32` (main is 0.44.31 after #537 took that slot).
- F10 (reset the run-local `run_context_tokens` / `Session._context_tokens_hint` when compaction replaces the context) is NOT folded in: the rebase did not touch the loop replacement branch or `_finish_compaction`, so it stays the recorded follow-up.
- Gates (whole tree, as CI spells them): flake8 clean, black 799 unchanged, isort clean, pyright 0 errors. Touched-dir tests (`tests/unit/{providers,model,session,harness,analytics,compaction}` + `test_config.py`, `test_settings_io.py`, `-n 4`): 2582 passed.

Next review round can be scoped as a delta of `aaad706d..bf6be177` (expected: version bump only).

damianvtran:
### Agent review — round 4
- Scope: f82d55279d9e41ec3d5bbc587526fa5916f0860f..bf6be177b0d6b038b8d700e83e7c1486db9930e8
- Mode: Rebase freshness
- Delivery contract: reduce 5m-TTL prefix rewrites on large Anthropic contexts without changing any sub-threshold request body
- Delta reviewed: aaad706d..bf6be177 (rebase + version)
- Applicable risk surfaces: provider wire compatibility, cost accounting, config plumbing, analytics migration
- Verdict: **Changes requested**

The rebase itself is intent-identical and the source is unchanged from round 3. The one open item is that **CI is red on bf6be177 and it is not a flake** — see F11.

| ID | Severity | File:line | Finding | Suggested fix |
|---|---|---|---|---|
| F11 | **MAJOR** | `tests/unit/tui/test_settings_view.py:4098-4100` (pre-existing test, broken by this PR's `settings_io.py:377-390`) | CI `test (3.12)` fails on bf6be177 (run [33629795743](https://github.com/damianvtran/local-operator/actions/runs/33629795743): `1 failed, 10801 passed`) with `test_click_selects_a_row_without_arming_a_text_selection — AssertionError: the click did not move the cursor`. **Reproduced locally**: passes on `f82d5527` and `47bc768f` (main), fails on `aaad706d` and `bf6be177`. Root cause: the new `providers.anthropic.cache_ttl_1h_min_context_tokens` row is appended to the **model** section (row index 4), which pushes the **"Failover and retry" header** onto body line `region.y + 5` — exactly the fixed offset the test clicks (`_index_at` → row 5 = `header/failover`, `selectable=False`, so `on_click` correctly returns without moving `_selected`). Product behaviour is right (a header click must not move the cursor); the test's hard-coded `y = body.region.y + 5` is what broke. Every CI run on this branch has failed the same way (2cf0d11e, 67d02901, aaad706d, bf6be177) — it was missed because `tests/unit/tui` is outside the touched-dir test set the rounds ran. | In the test, derive the target line from the rows instead of a fixed offset, e.g. click the first selectable row that is not `view._selected` (`y = body.region.y + next(i for i, r in enumerate(view._rows) if r.selectable and i != view._selected)`; rows are one line each with `scroll_offset` 0 at 100x30), or simply use `+ 4`/`+ 6`. Then re-run CI and confirm `test (3.12)` green before merge. |

#### Verified
- **Rebase intent-identical**: `git range-diff 093cf9f4..aaad706d f82d5527..bf6be177` → commits 2–5 `=`; commit 1 `!` only because its `pyproject.toml` version hunk (0.44.27→0.44.30) was dropped; new commit 6 `chore(release): bump version to 0.44.32` touches `pyproject.toml` only (`0.44.31` → `0.44.32`; matches `f82d5527`'s `0.44.31`). Note: `origin/main` has since moved to `47bc768f` (#527), still `0.44.32` in `pyproject.toml` — so the PR's bump now collides with main and needs re-sequencing to **0.44.33** at merge time, as the merger's sequencing note already anticipates. PR is `MERGEABLE`.
- **No semantic interaction with main's #529/#530/#531/#538**: main's `session.py` drift (`093cf9f4..f82d5527`) touches `_last_reported_usage` helper removal, compaction marker/pass extraction (`build_compaction_marker` in `_finish_compaction`), and `_stamped_todo_fingerprint` — none overlap the PR's hunks. `Session._note_usage` (`session.py:4239-4284`) is the single writer of `_last_usage`/`_context_tokens_hint` (`:4258,:4260`) and is called from the post-run scan (`:4826-4834`) and unconditionally ahead of the mid-turn compaction gate (`:5670-5688`); the loop reads `get_context_tokens_hint` via `_host_context_tokens_hint` and stamps `context_tokens_hint` per request from `run_context_tokens` (`loop.py:492-570, 827-908`). `_finish_compaction` (`:6514`) still does not reset the hint — F10 unchanged (below).
- **Gates** (AGENTS.md recipe, whole tree, in a detached worktree at bf6be177): `python -m flake8 .` clean · `uvx --from black==26.1.0 black --check .` 799 unchanged · `uvx isort==5.13.2 --check .` clean · `python -m pyright --pythonpath .venv/bin/python .` 0 errors / 0 warnings. `pytest tests/unit/providers tests/unit/model tests/unit/session tests/unit/harness tests/unit/analytics tests/unit/compaction -n 4` → **2456 passed**.
- **CI on bf6be177**: lint ✅ type-check ✅ pip-audit ✅ tui-e2e (macos, ubuntu) ✅ · **test (3.12) ❌** (F11) · cli-sanity / server-sanity skipped (gated on test). Not green.

#### Observations (carry-forward, not blocking)
- **F10 (follow-up)**: after compaction replaces the context, the run-local `run_context_tokens` in the loop and `Session._context_tokens_hint` still carry the pre-compaction figure until the next assistant usage lands, so the first post-compaction request may be stamped with the 1h TTL on a context that just dropped below the threshold (one request at 2x write cost, no correctness impact). Reset both in `_finish_compaction` / the loop's replacement branch. Recorded follow-up; unchanged by this rebase.

damianvtran:
- Head: `f37019c85af29956f524b0bbe91072728f62ce7d` (rebased onto `47bc768f`, version 0.44.33)

| ID | Status | Commit | Note |
|---|---|---|---|
| F11 | **Fixed** | `f37019c8` | `test_click_selects_a_row_without_arming_a_text_selection` now derives the click line from the first selectable row that is not the cursor's (`view._rows` lookup, minus `scroll_offset`, with an on-screen guard) instead of the fixed `region.y + 5`, so future settings additions cannot push a section header under the click again. Reproduced red on `bf6be177`, green with the fix. Full `env -u NO_COLOR TERM=xterm-256color pytest tests/unit/tui -n 4 → 4353 passed, 12 skipped`. Version re-sequenced to **0.44.33** on top of main's 0.44.32 (main's own bump commit dropped by the rebase as already-upstream). |
| F10 | Deferred | — | Carry-forward follow-up as recorded in round 4 (post-compaction hint reset); unchanged. |

Gates at `f37019c8` (whole tree): flake8 clean · black 802 unchanged · isort clean · pyright 0/0 · `pytest tests/unit/providers tests/unit/model tests/unit/session tests/unit/harness tests/unit/analytics tests/unit/compaction -n 4 → 2514 passed`. CI run [33639422752](https://github.com/damianvtran/local-operator/actions/runs/33639422752) fully green including `test (3.12)`, `cli-sanity`, `server-sanity`. PR is MERGEABLE. Not merged.

damianvtran:
### Agent review — round 5
- Scope: 47bc768f..f37019c85af29956f524b0bbe91072728f62ce7d
- Delivery contract: reduce 5m-TTL prefix rewrites on large Anthropic contexts without changing any sub-threshold request body
- Delta reviewed: bf6be177..f37019c8
- Applicable risk surfaces: provider wire compatibility, cost accounting, config plumbing, analytics migration, TUI test stability
- Verdict: **Approve**

F11 is fixed at the source (the test, not the product), the rebase is intent-identical, and CI is fully green on the current head. No BLOCKER or MAJOR remains — this round is **TERMINAL**; the only carry-forward is the F10 observation below, which is a follow-up and not a reason for another round.

#### Remaining findings

| ID | Severity | Status | Location | Note |
|---|---|---|---|---|
| F11 | major → **resolved** | Fixed in `f37019c8` | `tests/unit/tui/test_settings_view.py:4098-4106` | The click line is now derived from `view._rows` (`target = next(i for i, row in enumerate(view._rows) if row.selectable and i != before)`) and mapped with `y = body.region.y + target - body.scroll_offset.y`, which is the exact inverse of `_index_at` (`settings_view.py:2456`: `index = (y - region.y) + scroll_offset.y`). A selectable row that is not `_selected` takes the `on_click` branch that sets `_selected` (`settings_view.py:2264-2268` skip header/same-row; the else path moves the cursor), so the assertion targets the product path rather than a coordinate. Adding another settings row anywhere shifts `target` at most to a neighbouring index near the top of the list; the on-screen guard `assert body.region.contains(x, y)` turns a would-be silent "did not move" into an explicit "not on screen" failure, so the test cannot land on a header again. |
| F10 | minor (observation, carried forward) | Deferred — follow-up | `loop.py` replacement branch / `session.py:_finish_compaction` | Unchanged and unchanged in scope: neither the run-local `run_context_tokens` nor `Session._context_tokens_hint` is reset when compaction replaces the context, so the first post-compaction request can be stamped with the 1h TTL on a prefix that just dropped below the threshold (one request at 2× write cost, no correctness impact; next usage report self-corrects). Two-line fix in a follow-up PR: reset `run_context_tokens = None` when `on_turn_end` returns a list, and clear the hint in `_finish_compaction`. Not a gate for this PR. |

#### Verified
- **Delta is exactly test fix + version + rebase.** `git diff bf6be177 f37019c8 --stat` shows only the drift from main's #527 (18 files, all `docs/design/model-catalogue-freshness.md`, `model/{catalogue,configure,discovery,ids,prices,registry}.py`, `tui/{app,widgets/model_picker}.py` and their tests — none touched by this PR) plus `pyproject.toml` and `tests/unit/tui/test_settings_view.py`. `git range-diff f82d5527..bf6be177 47bc768f..f37019c8`: commits 1–5 all `=`; `bf6be177` (`chore(release): bump version to 0.44.32`) dropped as already-upstream; `f37019c8` new. An interdiff of the full PR patch (`f82d5527..bf6be177` vs `47bc768f..f37019c8`, index lines stripped) differs only in five `clients.py` hunk-header line offsets (main drift), the version line (`0.44.31→0.44.32` becomes `0.44.32→0.44.33`), and the 10-line test hunk. No source hunk changed.
- **Version**: `47bc768f:pyproject.toml` is `0.44.32`; head is `0.44.33`. Correctly re-sequenced on top of main.
- **Reworked test** (`f37019c8`): reasoned above; also read `_index_at` (`settings_view.py:2440-2459`) and the `on_click` non-selectable / same-index guards (`:2264-2268`) — the target row is selectable, not the cursor's, and not a suggestion row (those are built unselectable, `:609`), so the click always reaches the cursor-move path. `settings_view.py` itself is untouched by this PR.
- **Test run** (worktree at `f37019c8`, `env -u NO_COLOR TERM=xterm-256color .venv/bin/python -m pytest tests/unit/tui/test_settings_view.py -q`): **530 passed**, 10 warnings, 70s.
- **Four whole-tree gates** on `f37019c8`, exactly as AGENTS.md spells them: `python -m flake8 .` rc=0 · `uvx --from black==26.1.0 black --check .` → 802 files unchanged, rc=0 · `uvx isort==5.13.2 --check .` rc=0 · `python -m pyright --pythonpath .venv/bin/python .` → 0 errors, 0 warnings, rc=0.
- **CI run [33639422752](https://github.com/damianvtran/local-operator/actions/runs/33639422752)** on `headSha f37019c85af29956f524b0bbe91072728f62ce7d`: `status=completed, conclusion=success`; all eight jobs green — type-check, lint, test (3.12), pip-audit, tui-e2e (macos-latest), tui-e2e (ubuntu-latest), cli-sanity, server-sanity. `gh pr view 532` confirms `headRefOid` matches and base is `main`.
- **Freshness**: reviewed head `f37019c8` equals the PR's current `headRefOid`; `47bc768f` is an ancestor of the head and is the current `origin/main`.

damianvtran:

PR #533, opened 2026-09-02
fix(auth): usage-aware least-loaded account pick at first resolve

damianvtran:
## Summary of Changes

**Change class: C1 (standard / pre-authorized).** Bug fix / reliability; patch bump `0.44.27 → 0.44.29` (assigned as the second of three parallel PRs; the merger re-sequences the patch numbers if the merge order changes).

A session's (or subagent's — each child has its own session id) **first** pick among same-provider OAuth accounts was `crc32(session_id) % n` in `AuthStore._base_selection_order`: uniform over sessions, blind to how full each account already was. The existing usage-aware path (`ModelConfigurator.preflight_usage`) is **reactive** — it demotes an account once it reaches `usageReservePercent` — so by the time it acts the skew already exists.

Measured from `~/.local-operator/analytics.db` and the usage cache over the last 24h (5 Anthropic OAuth accounts, ~16 concurrent TUI sessions plus subagents): at 21:58 today three accounts sat at **65 / 92 / 99 %** of their 5-hour window while two sat at **6 % and 29 %**, and **33 sessions** hit `All 5 OAuth credentials unusable`.

This PR adds the **proactive** half: when no sticky exists for `(provider, session_id)`, the pool is ranked by each account's cached usage report before the hash rotation is applied.

### How it works (`AuthStore._usage_ranked_order`)

- Reads each OAuth row's last per-account report from the shared `usage_cache.db` (`UsageCacheStore.get(account_preflight_key(storage_id, email|account_id|cred:<id>), include_expired=True)` — the same key `ModelConfigurator._cached_account_usage` writes; verified against the live DB that all five rows map). **No network.**
- Reduces it with the same `usage_health(report, model_id)` the reserve rules use, so shared windows always count, a tier-scoped cap (`anthropic:7d:fable`) only counts for a model in that tier, and an enabled extra-usage meter supersedes the plan windows.
- **Herding guard:** rows within `USAGE_PICK_TOLERANCE` (10 points) of the best remaining fraction form a bucket; the bucket is rotated by the existing `crc32(session_id)` hash, so a burst of sessions/subagents starting together spread across it deterministically. Known-worse rows follow, best first.
- **Unknown is neutral:** a missing/unparseable report, or one older than `USAGE_PICK_MAX_REPORT_AGE_MS` (30 min) whose measured windows may have reset since, joins the bucket rather than sorting last. An old report whose every measured window still has `resets_at_ms` in the future stays trusted — usage inside a rolling window only rises, so it is a lower bound (this exception was added after the first live probe showed a 99 %-full account with a 114-minute-old report ranking neutral and drawing 7 of 12 new sessions).
- **Fail open:** any exception or missing cache → the unchanged hash order.
- **Sticky untouched:** the ranking only runs when the session has no sticky; `_resolve` pins the winner exactly as before, so a running session never moves (the provider prompt cache is per account).
- The cache handle is closed after each ranking so an `AuthStore` holds no extra SQLite descriptors between resolves (the test suite builds hundreds of stores per worker and hit `EMFILE` under the default 256-fd limit when the handle was kept open).

### Setting

`retry.usageAwareAccountPick` (default **true**), parsed in `RetrySettings.from_settings` beside `usageAwareFallback`, registered in `settings_io` (`/settings` + `config` CLI) and `config.py` defaults, and forwarded to the store by `SessionStreamFn.__init__` via the new `AuthStore.configure_usage_aware_pick(...)`. Documented in the configuration and failover guides.

`model_id` is now threaded through `_selection_order` → `_base_selection_order` so the ranking can scope tier caps; the three cascade call sites pass it.

## Related Issue(s)

- Related: usage-skew / "All 5 OAuth credentials unusable" investigation (sibling PRs: advisor/aside `tool_choice` cache split; 1h cache TTL on large contexts).

## Impact

- `AuthStore.__init__` gains an optional `usage_cache=` injection seam (tests only); `_selection_order`/`_base_selection_order` gain a keyword-only `model_id=""` (existing positional callers unchanged).
- Cost: a few SQLite reads once per session per provider at first resolve (and after a demotion re-pick). No provider calls.
- Behaviour change for users with several accounts on one provider: new sessions now start on the least-loaded account instead of a hash-chosen one. Opt-out restores the previous order byte-for-byte (asserted).

### Gates (whole tree, as CI)

```
.venv/bin/python -m flake8 .                                  → clean
uvx --from black==26.1.0 black --check .                       → 769 files would be left unchanged
```

### Unit tests

New `TestUsageAwareFirstPick` in `tests/unit/providers/test_auth_store.py` (11 cases, temp `UsageCacheStore` with synthetic `UsageReport`s written under the real preflight key):

- (a) least-loaded wins when reports differ beyond the tolerance — the observed 65/92/99/6/29 skew lands every one of 40 sessions on the 6 % account, full order best-first;
- (b) within-tolerance ties spread by the session hash (bucket first, deterministic per session, rest best-first); all-equal reproduces the hash order exactly;
- (c) empty cache → hash order; reports past max age without future resets → hash order; old report with resets still ahead stays trusted while a sibling whose window reset since is unknown; an unknown account is neutral, not penalised;
- (d) sticky respected regardless of usage through `get_oauth_access` (and a new session does move);
- (e) a spent Fable weekly does not penalise an Opus request but does a Fable one;
- opt-out restores hash order; a cache whose `get` raises fails open; `SessionStreamFn` forwards the setting.

```
tests/unit/providers/test_auth_store.py -k UsageAwareFirstPick -n0   → 11 passed
tests/unit/providers tests/unit/model tests/unit/session tests/unit/test_settings_io.py -n 4
  → 1880 passed, 1 failed
```

**Round 1 remediation (cbaf28ea):** the pick now also reads the warmer's per-provider `/usage` payload, sliced per account by `UsageReport.identity` (`UsageCacheStore.latest_account_report`), because the `:pf:` preflight rows it read before exist only while `retry.usageAwareFallback` is on — not the shipped default. Three new `TestUsageAwareFirstPick` cases (warmer-only cache ranks the (a) skew identically; the newer of a preflight and a warmer report wins in both directions; a `:pf:` row carrying a sibling's identity is never read by the warmer scan) plus one `test_usage_cache.py` case for the new read (newest `fetched_at` across fingerprints, other providers and preflight rows ignored, expiry ignored). Gates re-run clean on the whole tree (flake8 / black / isort / pyright 0 errors).

```
tests/unit/providers tests/unit/model tests/unit/test_settings_io.py -n 4   → 1207 passed
```

(The pre-existing gemini `test_catalogue` failure did not fire in this run.)

**Round 2 remediation (877d5bea + 44be0178, head 44be0178):** the pick now skips usage-report candidates with no measurable limit before the newest-wins comparison, so the warmer's fresh limit-less failure stub can no longer mask a slightly older real preflight report (F7 — new regression test reproduces the reviewer's scenario: fresh stub + 2-min-old real 95 %-used report → the real one wins, fails unpatched, passes fixed); and `TestUsageAwareFirstPick` closes every `AuthStore` it opens via an autouse fixture (F6 — three consecutive `ulimit -n 256` runs of `tests/unit/providers -n 4` → 701 passed each). All four whole-tree gates re-run clean on the round-2 head.

```
tests/unit/providers tests/unit/model tests/unit/test_settings_io.py -n 4   → 1208 passed
``` Live read of this machine's `usage_cache.db` — which holds NO `:pf:` rows for the current account set, only the warmer's row — now attributes a report to every stored Anthropic credential and ranks on it:

```
row  1: warmer_hit=True age=217s remaining=0.00
row  4: warmer_hit=True age=217s remaining=0.00
row  6: warmer_hit=True age=217s remaining=0.46
row  7: warmer_hit=True age=216s remaining=0.00
row 37: warmer_hit=True age=216s remaining=0.01
order for a fresh session: [6, 37, 1, 4, 7]
```

The one failure is `tests/unit/model/test_catalogue.py::test_the_gemini_wire_reaches_the_spec_through_a_real_transport[gemini-2.5-pro]` ("the transport was never exercised"), which fails identically on a clean `origin/main` checkout in this worktree (verified with the change stashed) — unrelated to this PR.

### Live evidence (real `auth.db` + `usage_cache.db`, this branch on `PYTHONPATH`)

Probe: 12 random fresh session ids through `AuthStore.get_oauth_access('anthropic', session_id=..., model_id='claude-opus-5', read_only=True)`, once with the pick on and once off. Cached usage at the time (used fractions; remaining = worst shared window):

```
cred   1: 5h=42%  7d=28%  remaining=0.58  age=3.2m
cred   4: 5h=82%  7d=31%  remaining=0.18  age=43.3m
cred   6: 5h=100% 7d=27%  remaining=0.00  age=24.2m
cred   7: 5h=89%  7d=26%  remaining=0.11  age=7.7m
cred  37: 5h=99%  7d=37%  remaining=0.01  age=114.5m

usageAwareAccountPick=True : picks over 12 fresh sessions -> {1: 12}
usageAwareAccountPick=False: picks over 12 fresh sessions -> {1: 2, 4: 1, 7: 4, 37: 5}
```

With the setting on, every new session lands on the only account with headroom (0.58 remaining; the next best, 0.18, is 40 points behind so the bucket is a single row); with it off, 9 of 12 land on accounts at 82–99 % of their 5-hour window — the skew this PR exists to stop. (With two accounts within 10 points of each other the bucket rotates by hash — case (b) above — so the on-path is not a single-row funnel in general.)

Then one small **live Anthropic request** on the account a fresh session picked, with that account's usage endpoint read before and after (≤ 50 tokens total; the accounts are near their caps):

```
fresh session 429090 resolved to credential id 1
  second resolve for the same session -> credential id 1 (sticky)
  usage BEFORE: {'anthropic:5h': '45%', 'anthropic:7d': '29%', 'anthropic:7d:fable': '47%'} remaining(shared)=0.55
  live message usage: input=42 output=4 cache_read=0 cache_write=0
  usage AFTER : {'anthropic:5h': '45%', 'anthropic:7d': '29%', 'anthropic:7d:fable': '47%'} remaining(shared)=0.55
```

The pick resolved to the least-loaded account, the request ran on it, and the session stayed sticky to it on re-resolve. (A 46-token call does not move a 5-hour percentage, so before/after are equal — the point of the call is that the chosen credential serves live traffic.)

### Not validated

- The multi-session steady state (whether the daily skew actually flattens) needs a day of real sessions on the merged build; the probe above is the mechanism, not the outcome.
- No TUI change, so no screenshots.

## Checklist

- [x] All new and existing tests pass (one pre-existing unrelated catalogue failure noted above).
- [x] Security considerations are addressed (no secrets on the new read path: API-key rows are deliberately not looked up by secret hash).


damianvtran:
### Agent review — round 1
- Scope: 093cf9f4..cd148bc81f3854ea6f0ec0633c75889755d8a16e
- Mode: Comprehensive
- Delivery contract: spread first-resolve account picks toward the least-loaded Anthropic accounts without moving existing sessions or changing behaviour when usage is unknown
- Applicable risk surfaces: credential routing, quota exhaustion, fail-open, config plumbing
- Verdict: Approve

**Gates (throwaway worktree at cd148bc8):** flake8 rc=0, black rc=0, isort rc=0, pyright rc=0 (0 errors). `tests/unit/providers tests/unit/model -n 4` → 1107 passed. Whole `tests/unit -n 4` → 10572 passed, 15 skipped (the pre-existing gemini `test_catalogue` failure did not fire in this run; not counted either way).

**Reproduced independently** (synthetic cache, real `AuthStore` + `UsageCacheStore` path, `claude-opus-5`):
- incident skew `[0.65, 0.92, 0.99, 0.06, 0.29]` used → 12 fresh sessions split `{4: 6, 5: 6}` across the two accounts with headroom (0.94 vs 0.71 remaining are >10 pts apart, but with a shared 7d at 0.30 both bind at ≤0.70/0.71, hence the tie); 40 sessions → 20/20.
- five identical reports → 12 sessions `{1:3, 2:2, 3:2, 4:4, 5:1}`, byte-identical to the pure hash order (degenerate case is a true no-op).
- one good + four bad → `{5: 12}` (bucket of one); two good → `{4: 8, 5: 4}`.
- 2h-old report whose 5h `resets_at_ms` is 1h in the past → `None` (unknown), so the stale-but-trusted rule does **not** trust a report whose window has already reset.
- 31-min-old report carrying only an `anthropic:extra` row (no `resets_at_ms`) → `None` (neutral), as intended by the "no reset timestamp ⇒ cannot vouch" clause.

No blocker or major found. The ranking, the stale rule, the tier scoping, and every fail-open path do what the description says.

| --- | --- | --- | --- |
| F1 | minor | quota exhaustion / expected impact on a stock config | `local_operator/providers/auth_store.py:1015-1067` reads only `account_preflight_key` rows (`<storage>:pf:<hash>`). The only writers of that key are `ModelConfigurator._cached_account_usage` call sites in `model/configure.py` (2106, 2146, 2537, 3125), every one of which sits inside `preflight_usage` **after** the `if not retry.usage_aware_fallback: … return` gate at 2447. `/usage` and the warmer write `provider_cache_key` rows (`anthropic:<hash>`, no `:pf:`), which the pick does not read. So on the shipped default (`usageAwareFallback: false`) the cache never carries a pf row and `usageAwareAccountPick: true` is a permanent hash-order no-op — safe (fail-open) but silent, and the setting help text / GUIDE say nothing about the dependency. This machine has `usageAwareFallback: true` so the live probe could not surface it. Action: either state in `settings_io.py:452-460` help + `guides/failover/GUIDE.md` that the pick only has data when `usageAwareFallback` is on, or (preferable) fall back to the warmer's per-provider row when it carries `identity` for the account. |
| F2 | minor | credential routing / neutral placement | `auth_store.py:1090-1094`: an unknown row joins the bucket with the known-best rows, and known-worse rows go to `rest`. When the best known row is at 0.94 remaining and one row is unknown, the unknown row is ranked *equal to* an account that is provably 94% free, ahead of a known 0.71. That is the documented choice ("no evidence is not evidence against") and it is defensible, but the opposite failure mode exists: after a 5h rollover the previously-full accounts have reports whose windows reset (⇒ unknown ⇒ bucket), while the account that was fine keeps a trusted report; so post-rollover the pool briefly herds by hash over `{fresh-good} ∪ {all-unknown}`, which is fine, until one boundary refreshes only the good account's row and the unknowns stay unknown until *their* next preflight. Net effect is neutral-to-good; noting it so the policy is a recorded decision rather than an accident. No action needed beyond a one-line comment at 1090 saying unknown deliberately outranks known-worse. |
| F3 | minor | config plumbing / test coverage vs. claim | `tests/unit/providers/test_auth_store.py::test_the_session_stream_forwards_the_opt_out` asserts `_usage_aware_pick` after constructing `SessionStreamFn`, but `SessionStreamFn.__init__` at `model/configure.py:1727-1732` is the **only** place the setting reaches the store. Construction paths that never build a `SessionStreamFn` on a store (`cli.py:1700`, `cli.py:2229`, `mcp/manager.py:1657`, `mcp/auth.py:372`) keep the default ON regardless of the operator's opt-out. That is by design per the docstring at `auth_store.py:362-375`, but it means `usageAwareAccountPick: false` does not fully "restore the pure per-session hash spread" as the GUIDE claims — `lop` login/credential CLI resolves still rank. Action: soften the GUIDE/help wording to "for sessions" or thread the flag through the CLI store constructors; not gating because those paths hardly ever resolve OAuth for a session id. |
| F4 | nit | maintainability | `pyproject.toml` jumps `0.44.27 → 0.44.29`, skipping `.28`; tags on origin stop at `v0.44.27`. If `.28` was consumed by another open PR this is fine, otherwise bump to `.28`. |
| F5 | nit | docs | `guides/failover/GUIDE.md:30-38` ("Rotation within one provider") still describes the order as "sticky, then a per-session hash, then round-robin". Line 132 of the same guide points the reader at that section for the new setting, but the section was not updated to mention the usage-ranked step. One sentence there closes the loop. |

#### Follow-up observations

- **Herding (the coder's `{1: 12}` probe).** Reproduced in shape: it is the bucket-of-one case, where the next-best account trails by 40 points. Sending 12 sessions to the only account at 58% remaining while the others sit at 0–18% is the *correct* answer — spreading them would put 9 of 12 on accounts about to 429. The tolerance bucket does its job once two rows are within 10 points (`{4: 8, 5: 4}` in my probe, `{4: 6, 5: 6}` on the incident skew), and the all-equal case degrades to exactly the hash order. Acceptable, not a defect. The one thing to watch is the interval between a burst landing and the next boundary preflight refreshing that account's row; nothing in this PR feeds a pick back into the cache, so ~12 sessions worth of first turns can hit one account before its reported fraction moves. That is inherent to a read-only design and the bucket + sticky keep the blast radius to one 5h window.
- **Proactive vs. reactive fight.** They cannot fight: the pick runs inside `_base_selection_order` only when there is no sticky, `_resolve` pins the winner immediately (`auth_store.py:1507`), and the reactive walk in `preflight_usage` calls `get_oauth_access` *first* (2477) and then probes the account it was handed. A pick landing on an account the preflight is about to demote just means the preflight demotes it one boundary later, exactly as it would have under the hash. `deprioritize_credential` is applied after the ranking (`_selection_order:911-946`), so a demoted-but-least-loaded account still sorts last — correct precedence. Model-scoped blocks are filtered upstream in `_usable_key_rows`, so a blocked row never reaches the ranking.
- **Tier windows.** `usage_health` (usage.py:360-366) includes a tier row only when `limit.tier.lower() in model_id.lower()`; `7 day (Fable)` therefore never binds an opus request. Confirmed by the test at 1517-1530 and my own probe.
- **Cache handle lifecycle.** Open-per-ranking is fine on this hot path: a ranking happens once per session per provider (plus re-pick after a non-usage demotion), and `sqlite3.connect` on an existing WAL file is single-digit ms. The `finally` at 1113-1119 closes it even on the fail-open path, so the EMFILE the coder hit cannot recur from this store. Thread safety: `AuthStore` is single-loop and the handle is used synchronously inside one call, so no cross-thread reuse of the connection occurs. One subtle point: `close()` at `auth_store.py:309-312` sets `_usage_cache = None` but leaves `_usage_cache_probed = True`, so a store that is closed and then resolves again will *not* rebuild the cache and silently ranks by hash. `close()` is terminal in every caller I found, so this is theoretical.
- **Fail-open coverage.** Every raise inside the ranking is caught at 1110-1112; the probe at 993-1013 catches construction failure; `UsageCacheStore.get` is itself exception-safe; a cache whose `get` raises falls back to hash (test at 1546-1555). I found no exception path that reaches `_resolve`.
- **Settings plumbing.** TUI (`session_factory.py:1474`), server (`server/utils/operator.py:571`), evaluation runner (`provider_client.py:269`) all go through `create_stream_fn(auth_store, settings=…)` → `SessionStreamFn.__init__`, so the flag reaches the store for every real session. Subagents share the parent's stream fn and therefore the parent's already-configured store. `from_settings` accepts both spellings and defaults `True`; `test_settings_io` `_consumer_defaults` was extended so the settings-registry/consumer parity test now covers it.

This round is TERMINAL: no blocker or major remains; F1–F5 are follow-ups, not a reason for another round.

damianvtran:
- Scope: cd148bc8..cbaf28ea (new head `cbaf28eafc5bb3971fb679e418ad789585bfeab3`)

- F1 (minor) — Fixed in cbaf28ea: took the preferred option. `UsageCacheStore.latest_account_report(storage_id, identities)` slices the warmer's per-provider payload per account via `UsageReport.identity`, scanning every non-`:pf:` row under the storage id (current fingerprint and any older one still inside retention) rather than rebuilding `ProviderController._account_fingerprint` — which would have hashed API-key secrets on a read path. `_cached_remaining_fraction` now takes the newer of the preflight row and the warmer report, matched on every label an enumerator could attach (email / account_id / identity_key / `cred:<id>`). `:pf:` is now the `PREFLIGHT_KEY_SEGMENT` constant so writer and scan share one spelling. Tests: three new `TestUsageAwareFirstPick` cases (warmer-only cache ranks the (a) skew identically; newer report wins in both directions; a poisoned `:pf:` row is never read by the scan) and one `test_usage_cache.py` case for the read itself. Verified live against this machine's cache, which holds no `:pf:` rows for the current account set: every Anthropic row now attributes a report (see updated Testing Details). The GUIDE also names both data sources.
- F2 (minor) — Fixed in cbaf28ea: comment at the bucket partition in `_usage_ranked_order` records that an unknown row deliberately outranks a known-worse one and why (cold cache / just-reset window; penalising absence starves a fresh login).
- F3 (minor) — Fixed in cbaf28ea: took the cheaper honest option — wording, not plumbing. Threading the flag through `cli.py`/`mcp/manager.py`/`mcp/auth.py` would switch off nothing: those stores never pass a `session_id`, and `_base_selection_order` only reaches the ranking when one is present (session-less resolves round-robin). Help text now says "Applies to sessions only", the configuration GUIDE says "for sessions", the failover GUIDE's yaml block explains which stores never rank, and the `RetrySettings` field comment states the same.
- F4 (nit) — Rejected: `0.44.28` is consumed by open PR #534 (`fix/aside-tool-choice-cache`, `pyproject.toml` → `0.44.28` in its diff); `0.44.29` stands.
- F5 (nit) — Fixed in cbaf28ea: "Rotation within one provider" now describes sticky → usage-ranked first pick (bucket + hash rotation, unknown = neutral, cache-only, no network) → plain hash → round-robin, and the sticky-after-first-pick rule.

Gates on cbaf28ea (whole tree): flake8 rc=0, black rc=0 (769 unchanged), isort rc=0, pyright 0 errors. `tests/unit/providers tests/unit/model tests/unit/test_settings_io.py -n 4` → 1207 passed (the pre-existing gemini `test_catalogue` failure on main did not fire in this run).

damianvtran:
### Agent review — round 2
- Scope: 093cf9f4..cbaf28eafc5bb3971fb679e418ad789585bfeab3
- Delivery contract: spread first-resolve account picks toward the least-loaded Anthropic accounts without moving existing sessions or changing behaviour when usage is unknown
- Delta reviewed: cd148bc8..cbaf28ea
- Applicable risk surfaces: credential routing, quota exhaustion, fail-open, secret hygiene, config plumbing
- Verdict: Approve

**Gates (throwaway worktree at cbaf28ea, whole tree exactly as CI):** `python -m flake8 .` rc=0, `uvx black==26.1.0 --check .` rc=0, `uvx isort==5.13.2 --check .` rc=0, `python -m pyright .` rc=0. `tests/unit/providers tests/unit/model tests/unit/test_settings_io.py -n 4` → **1207 passed** (with `ulimit -n 4096`; see F6 for what happens at the macOS default of 256). The pre-existing gemini `test_catalogue` failure did not fire.

**F1 re-review (the substantive change).** The warmer-row slice is correct in the ways that matter:
- *Identity matching.* The scan matches `report.identity` exactly against the label set `{email, account_id, identity_key (non-`oauth:`), cred:<row.id>}`. That is a superset of every label the two warmer enumerators can attach (`_expected_oauth_identities` → email | account_id | identity_key | `cred:<id>`; `list_oauth_accesses` → email | account_id | `cred:<credential_id>`, where `credential_id` is `row.id`), so a report for a row cannot go unmatched, and a report can only be attributed to a different row if two rows carry the same label — which is a collision the warmer itself already has (`live[report_identity_key(report)]` is last-wins), not one introduced here. `identity=None` on a warmer report → `None not in identities` → skipped (reproduced). `:pf:` rows are excluded by the shared constant and a hex fingerprint cannot contain `:pf:`; `substr(key,1,len("anthropic:"))` cannot match a sibling storage id such as `anthropic-foo:`; the API-key route's `anthropic:key:<hash>` rows carry no identity and drop out.
- *Staleness precedence.* Both shapes write `fetched_at` as epoch-ms from the same `UsageReport`, so `max(candidates, key=fetched_at)` compares like with like; the age and window-reset policy is applied after the max, once. See F7 for the one shape where "newest" is not "most informative".
- *Scan cost.* `latest_account_report` on a pathological cache (200 fingerprint rows × 5 reports) is ~3 ms per call; a real cache holds a handful of rows. It is called once per credential row per ranking, so a 5-account pool does 5 scans — still low single-digit ms on real data. The connection is the store's shared handle and `_usage_ranked_order`'s `finally` closes it on every path (normal, fail-open, exception); verified with an fd probe: 30 stores + rankings → 0 retained after `close()`.
- *Secret hygiene.* The read path touches only email / account_id / identity_key / `cred:<id>` and the JSON usage payloads; no token or key bytes are read, hashed, or logged. The ranking's debug line logs credential ids and rounded fractions, not identities. The scan's only log is `exc_info` on a failed `SELECT`.
- *Fail-open on malformed payloads.* Reproduced: a non-JSON row and a row whose report has `fetched_at: "zzz"` under the provider prefix are both skipped and the sibling account still returns its fraction; a raising `SELECT` returns `None`.
- *Tests exercise the warmer path end to end.* `_cache_report(via="warmer")` writes under `provider_cache_key("anthropic", fingerprint_accounts([...]))` with `identity=email`, and the assertions go through `_selection_order` → `_usage_ranked_order` → `_cached_remaining_fraction` → `latest_account_report` on a real `UsageCacheStore`, so the exact read path is covered. The poisoned-`:pf:` test pins the namespace split, and the newer-wins test covers both directions.

**F2–F5 dispositions confirmed:** F2 comment present at the bucket partition (`auth_store.py:1160-1165`); F3 "sessions only" wording landed in `settings_io.py`, both guides, and the `RetrySettings` field comment; F4 rejection verified — PR #534's diff bumps `pyproject.toml` to `0.44.28`, so `.29` is correct; F5 "Rotation within one provider" now describes sticky → usage-ranked first pick → hash → round-robin.

| --- | --- | --- | --- |
| F6 | minor | test hygiene / reproducibility of the suite | The suite is now flaky under the macOS default `ulimit -n 256`. At cbaf28ea: 3 runs of `tests/unit/providers tests/unit/model tests/unit/test_settings_io.py -n 4` → `18 failed`, `1207 passed`, `5 failed`, every failure `OSError: [Errno 24] Too many open files` in unrelated `test_oauth_flows`/`test_failover` tests. At the base `093cf9f4`: 3/3 runs `1191 passed`. At cd148bc8: 4/4 runs fail (28/1/4/1). Per-test fd accounting shows `TestUsageAwareFirstPick` adds +3 descriptors per test (11 tests → `test_auth_store` peaks at 174 fds vs 141 on base): `_store()` builds an `AuthStore` whose `auth.db` connection is never closed, on top of the pre-existing leaky `TestADemotionSurvivesAMixedPool`/`TestLoginFlavourAliases` fixtures that already run the worker close to the limit. The usage-cache handle itself IS closed by the ranking, as the docstring claims. CI (ubuntu, soft limit 1024) is unaffected, which is why it went unnoticed. Fix: turn `_store` into a fixture with teardown (`store.close()`), or `request.addfinalizer(store.close)`; one line per class. Not gating — it does not indicate a runtime leak. |
| F7 | minor | quota exhaustion / "newer wins" discards a known number for an empty one | `auth_store.py:1074` takes the newest candidate by `fetched_at` before checking whether it carries any limits. The warmer writes a limit-less failure stub with `fetched_at = now` for an account whose FIRST probe under a fingerprint 429s (`_mark_account_failure(previous=None)` at `controller.py:636`, reached whenever the row for the new fingerprint has no `previous`, e.g. right after a login/logout changes the set). That stub is newer than the same account's real report in the preflight row or in the previous fingerprint's still-retained row, so it wins the `max`, `usage_health` reads no measurable window → `None` → the account ranks NEUTRAL in the bucket instead of behind its known 95%-used sibling. Reproduced: preflight row 2 min old at 0.95 used + fresh stub → `_cached_remaining_fraction` = `None`; without the stub → `0.05`. Direction is fail-open (pre-PR behaviour), so not gating, but it is exactly the account a 429 just told us is full. Fix: in `latest_account_report` (or before the `max`) skip reports with no measurable limit (`not any(l.amount.fraction() is not None for l in report.limits)`) so an empty stub never outranks a real number. |

#### Verified
- Whole-tree gates as CI runs them: flake8 / black / isort / pyright all rc=0 on cbaf28ea.
- 1207 passed in the requested test scope (fd limit raised); the base at 093cf9f4 and the r1 head were run under identical conditions for the F6 comparison.
- Warmer-only cache ranks the (a) skew identically to the preflight-only case (coder's new test; also reproduced via a hand-built cache through `_selection_order`).
- `identity=None` warmer report → no match; non-JSON row and unparseable `fetched_at` → skipped, sibling still ranked; raising `SELECT` → `None`.
- Scan: ~3 ms per call on a 200-row/1000-report cache; ~0 on a realistic one. Handle released after every ranking (0 retained fds after `close()` across 30 stores).
- No secret material on the read path or in any log line added by this delta.
- F4 rejection: `gh pr diff 534` shows `-version = "0.44.27"` / `+version = "0.44.28"`.

#### Follow-up observations
- **Same-label collision.** `_identity_key_for` scopes Anthropic rows by `org_id` first, so one email logged into two orgs yields two rows both labelled by that email in the warmer, and the pick then reads the same report for both. This is the warmer's own last-wins collision surfacing in a second reader, not a new fault; if per-org quota ever matters, the warmer's label should become the identity key, and this scan follows for free.
- **One scan per row.** `_cached_remaining_fraction` re-runs the prefix scan (and re-parses every payload) for each credential row. Hoisting the scan to once per ranking (`latest_account_reports(storage_id) -> dict[identity, report]`) would make it O(rows) instead of O(rows × accounts). Not worth a round at today's cache sizes.
- **Label coverage in tests.** The auth-store tests only exercise the email label through the pick; the `account_id`/`cred:<id>` spellings are covered only at the `UsageCacheStore` level. A row with no email and a report labelled `cred:<id>` would be a cheap addition if the label set is ever narrowed.

This round is TERMINAL: no blocker or major remains; F6 and F7 are follow-ups, not a reason for another round.

damianvtran:

- F6 (minor) — Fixed in 44be0178: `TestUsageAwareFirstPick` now carries an autouse fixture that closes every `AuthStore` its `_store` helper opened (`AuthStore.close` also closes the usage cache it was handed), so the ~3 fds/test leak is gone. Verified with three consecutive `ulimit -n 256` runs of `tests/unit/providers -n 4` — 701 passed each time.
- F7 (minor) — Fixed in 877d5bea: `_cached_remaining_fraction` now drops candidates with no measurable limit before the `fetched_at` max, so the warmer's fresh limit-less failure stub can no longer mask a slightly older real preflight report. New regression test `test_a_fresh_limit_less_stub_does_not_mask_a_real_report` reproduces the reviewer's scenario (fresh stub + 2-min-old real 95%-used report → the real one wins); it fails against the unpatched source and passes on the fix.

damianvtran:
### Rebase note

**Old head → new head:** `44be0178` → `ecfdce78` (rebased from `093cf9f4` onto `origin/main` at `47bc768f`).

**Conflicts resolved:** one — `pyproject.toml` `[project] version` (main moved to 0.44.32 while this branch carried 0.44.29). Resolved by setting the version to the assigned slot; no source or test hunks conflicted. `local_operator/model/configure.py` auto-merged cleanly against #527.

**Intent:** identical. `git range-diff 093cf9f4..44be0178 origin/main..ecfdce78` shows commits 2–4 as `=` (byte-identical) and commit 1 differing only in the version line.

**Version:** `0.44.37`.

**Gates (whole tree, as `.github/workflows/ci.yml`):** flake8 ✓ · black 26.1.0 ✓ (802 unchanged) · isort 5.13.2 ✓ · pyright ✓ (0 errors). `pytest tests/unit/providers tests/unit/model tests/unit/test_settings_io.py -n 4`: 1266 passed.

Round 2 approved `cbaf28ea`; the round-2 remediation (`44be0178`: F6 fixture close, F7 limit-less stub skip) and this rebase are unreviewed — a round-3 freshness review follows.

damianvtran:
### Agent review — round 3
- Scope: 47bc768f..ecfdce78c88b0b8747dfdabf31c89a4d1354105c
- Mode: Remediation convergence + rebase freshness
- Delivery contract: spread first-resolve account picks toward the least-loaded Anthropic accounts without moving existing sessions or changing behaviour when usage is unknown
- Delta reviewed: cbaf28ea..ecfdce78
- Applicable risk surfaces: credential routing, quota exhaustion, fail-open, secret hygiene, config plumbing
- Verdict: Approve

#### Remaining findings

| ID | Severity | Status |
|----|----------|--------|
| F6 | minor | Fixed — verified (ecfdce78, formerly 44be0178) |
| F7 | minor | Fixed — verified (09c7ce23, formerly 877d5bea) |

No new findings. No BLOCKER or MAJOR remains — this round is **terminal**.

#### Verified

1. **F6 fixture closes every store.** `TestUsageAwareFirstPick._store` is the only place in the class that opens an `AuthStore`/`UsageCacheStore` (`tests/unit/providers/test_auth_store.py:1358-1359`), and every call appends to `self._opened`, which the autouse `_close_stores` fixture drains; `AuthStore.close` (`auth_store.py:309-313`) closes the handed-in usage cache too. Reproduced the fd-pressure scenario: `ulimit -n 256; pytest tests/unit/providers -n 4` → **701 passed**, no `Errno 24`.
2. **F7 filter runs before the `max`, and the test is load-bearing.** `auth_store.py:1076-1081` drops limit-less candidates before `max(candidates, key=fetched_at)`. Reverted just those nine lines locally (`git apply -R` of the 877d5bea source hunk) and `test_a_fresh_limit_less_stub_does_not_mask_a_real_report` went **red** (`[1, 2]` vs expected `[0, 1]`-order swap); restored, it passes. The downstream age-cutoff branch still guards against a report with no measured windows, so the two checks are consistent.
3. **Rebase is intent-identical.** `git range-diff 093cf9f4..44be0178 47bc768f..ecfdce78`: commits 2–4 are `=`; commit 1 (`7c4d846b`) differs only in the `pyproject.toml` version line (`0.44.32` → `0.44.37`, base main is `0.44.32`). The `configure.py` hunk auto-merged against #527 and reads correctly in the post-rebase tree (`configure.py:1702-1711`).
4. **Whole-tree gates on the worktree at ecfdce78** (AGENTS.md spellings): `python -m flake8 .` → 0; `uvx --from black==26.1.0 black --check .` → 802 unchanged; `uvx isort==5.13.2 --check .` → clean; `python -m pyright --pythonpath .venv/bin/python .` → 0 errors/0 warnings. Targeted suite `tests/unit/providers tests/unit/model tests/unit/test_settings_io.py -n 4` → **1266 passed**.
5. **CI on ecfdce78** (run 33637484558): lint, type-check, pip-audit, test (3.12), tui-e2e (ubuntu + macos), cli-sanity, server-sanity — all pass. PR reports `MERGEABLE` against `main`.

#### Follow-up observations

- Non-blocking: the F7 filter and the age-cutoff branch both compute "limits with a measurable fraction"; a tiny helper would let the two stay in lockstep, but the duplication is two lines and clearly commented. Not worth a round.
- Non-blocking: the F7 regression test stamps its stub `now` via `time.time()` directly rather than the `now_ms` the store uses; harmless here because the assertion only depends on ordering, but a clock injection would make it fully deterministic if this class ever grows time-sensitive cases.

damianvtran:

PR #535, opened 2026-09-02
fix(model): drop the cached model listing when a credential changes

bbqben:
## Summary of Changes

Nothing invalidates a cached model listing when the **credential changes**. `lop login anthropic` writes the credential row and leaves the catalogue fetched under the *old* credential in place, so re-authing — the action a user takes when a model is missing — cannot clear it, and a new account inherits the previous account's model list until the TTL lapses.

Bug fix; patch bump `0.44.35 → 0.44.36`.

> **Scope narrowed after #527 landed.** This branch originally also forced the picker's listing past its TTL. #527 solved that better — stale-while-revalidate, a 15-minute `PICKER_TTL_S`, and a `want_id` miss refetch — so the `force_fresh` machinery is **dropped entirely** and `/model` keeps main's TTL. What remains is the half #527 did not cover: the credential-change trigger. `local_operator/tui/app.py` is now byte-identical to main.

## Diagnosis

`catalogue.invalidate` had exactly one caller — `discovery._available_models`, for a payload it could read but not map. `providers/auth_cli.py` contained no reference to `cache` or `listing` at all, and neither did `ProviderController.login`.

So the two events that genuinely change what a listing returns — a login and a logout — were invisible to the cache. #527's TTL bounds how long a stale list survives an *idle* session; it does not help the user who notices a problem and acts, because the action itself was a no-op.

```
1. picker opens, cache 23.5h old  -> status=cached fable5.1=False
2. user re-auths (invalidate)     -> dropped 1 document(s)
3. picker opens again             -> status=ok     fable5.1=True
   live ids: ['claude-fable-5', 'claude-fable-5-1', 'claude-opus-5']
```

## The Fix

`catalogue.invalidate_documents(storage_id)` drops every scoped document for one credential identity; `discovery.invalidate_listing(provider_id)` resolves a provider to that identity. Called from both login paths (`ProviderController.login`, `auth_cli.run_login`) and both logout paths.

Three details that are load-bearing rather than incidental:

- **Every scoped document, not one key.** `_cache_key` picks between `openai.listing`, `openai.oauth.listing` and `openai.oauth.<hash>.listing` using a credential resolved *after* the login. Targeting a single key leaves the document the next listing actually reads untouched — a half-fix that still looks like "re-auth did nothing".
- **Keyed on the credential identity.** `openai-device` and `openai` are one logged-in account sharing one document; invalidating under the literal provider id would miss it.
- **The dot separator bounds the glob.** `openai.*` cannot match `openai-device.listing.json`, `zai` cannot take out `zai-oauth`, and the skills index sharing this cache directory is never touched.

Best-effort by construction: a failed invalidation costs a stale list until the TTL lapses, while an exception would fail a login that actually succeeded.

## Impact

- **No latency change.** This adds one `unlink` per credential change. No new fetch, no new request on any read path; `live_catalogue` is untouched apart from the invalidation calls.
- No breaking changes, no dependency updates, no new public surface beyond the two helpers.

## Testing Details

**18 new tests.** The picker test is the end-to-end one: a model released 3h ago, which sits between main's 15-minute `PICKER_TTL_S` and the old 24h default — the one window where the two disagree.

**Proved the tests can fail.** Reverting `_refresh_catalogue` to the pre-#527 default TTL:

```
tests/unit/tui/test_app_pilot.py::test_the_picker_shows_a_model_released_since_the_listing_was_cached
>       assert "claude-fable-5-1" in offered
E       assert 'claude-fable-5-1' in {'claude-fable-5', 'claude-opus-5', 'claude-sonnet-5'}
```

Coverage:

- `test_catalogue.py` — every scoped document dropped together; a prefix-sharing provider (`zai` vs `zai-oauth`) untouched; the shared skills index untouched; an unlisted provider returns `0`, not an error.
- `test_discovery.py` — the incident end to end against the cache layer, asserting the stale open issues **no request at all**; invalidation follows the credential identity (`openai-device` → `openai`); an unknown provider does not raise.
- `test_controller.py` — OAuth login, API-key login, and logout each invalidate; `live_catalogue` adds no TTL of its own, so `PICKER_TTL_S` stays the single statement of picker freshness.
- `test_app_pilot.py` — the picker surfaces a model released since the listing was cached.

**Gates**, run over the whole tree as CI does, after `uv pip install -e ".[all,dev]"` (main added a `boto3` dev dep):

```
flake8 .                                    clean
black --check .                             clean
isort --check .                             clean
pyright                                     0 errors, 0 warnings, 0 informations
pytest tests/unit (minus osworld/aws)       10717 passed, 15 skipped
pytest <the four touched files>             797 passed
```

⚠️ `tests/unit/evaluation/adapters/osworld/test_aws_provider.py` fails locally with `ProfileNotFound: sandbox` — **22 failed + 2 errors, identical on a clean `origin/main` worktree** with the same venv. Environmental (no local AWS `sandbox` profile), pre-existing, and untouched by this PR; CI has the profile and passes.

**Merge resolution.** Four conflicts: `discovery.py` imports (kept both sides), `app.py` `_refresh_catalogue` (took main's `PICKER_TTL_S` wholesale, dropped `force_fresh` and the duck-typing `TypeError` shim that existed only to protect it), `pyproject.toml` version, `test_discovery.py` (kept main's assertion, appended after it).


bbqben:
CI note: the first `test (3.12)` run failed on `tests/unit/tui/test_paste_clipboard.py::test_the_reading_card_is_on_screen_while_the_read_is_still_running`, which is unrelated to this change and untouched by it (`git diff origin/main -- tests/unit/tui/test_paste_clipboard.py` is empty).

It is a wall-clock assertion of the shape AGENTS.md flags as flake-prone — `at 0.43s of a 0.6s read the card was '' (visible=False)` — i.e. a sample landed inside the stall window but the timer had not painted yet on a loaded runner. Passes 3/3 locally and passed on re-run with no code change; all 8 checks are now green.

Not fixed here on purpose: converting that bar to a structural invariant is its own change, and folding it into a model-listing PR would hide it.

damianvtran (approved):

bbqben:
Rebased onto main and **narrowed the scope**, since #527 landed the same incident's other half in the meantime.

**Dropped:** the `force_fresh` / once-per-process machinery and its `TypeError` compat shim. #527's stale-while-revalidate + 15-minute `PICKER_TTL_S` + `want_id` miss refetch is a better answer to "the picker serves a stale list", so `local_operator/tui/app.py` is now byte-identical to main and `live_catalogue` keeps its original signature.

**Kept:** credential-change invalidation, which #527 does not cover. A TTL bounds how long a stale list survives an *idle* session; it does not help the user who notices and re-auths, because that action was a no-op on the cache.

Net diff is now 5 source files, +457/-3, most of it tests. Re-approval welcome given the scope change — happy to hold if you'd rather re-read it.

Note on the earlier CI flake: it did not recur. The only local failures are `osworld/test_aws_provider.py` (`ProfileNotFound: sandbox`), which reproduce identically on a clean `origin/main` worktree with the same venv.

PR #534, opened 2026-09-02
fix(anthropic): keep asides on the turn's cache prefix (tool_choice)

damianvtran:
## Summary of Changes

**Change class: C1 (standard / pre-authorized).**

The compaction advisor (`Session.advise_compaction`) and the `/btw` / `/loop` aside & judge requests (`Session.complete_aside`) send the **live tool schema with `tool_choice="none"`** while the working turn sends `tool_choice=auto`. Anthropic's prompt-caching docs list `tool_choice` as invalidating the **messages** level of the cache hierarchy (tools ✓ / system ✓ / messages ✘), and fleet analytics over 24h (to 2026-09-01T23:20) attributed to that mismatch **33.9% of all Anthropic cache-write tokens: 254 calls / 70.7M tokens with the head-only signature** (`cache_read` 15–45k ≈ tools+system head only, `cache_write` > 100k ≈ context − head, `output < 2k`) — the same window and figures as `docs/evidence/compaction-advisor/aside-tool-choice-measurement.txt`; an earlier rough cut of the same query (~250 events / 52M / 25%) is superseded by those numbers. That attribution was the **hypothesis**; the measurement below did not confirm it, and the signature was later root-caused to per-account cache isolation under reserve-verdict account rotation (#537).

This PR makes the aside's request body byte-identical to the turn's prefix on the Anthropic wire:

- `AnthropicClient._build_body` now emits the turn's own `{"type": "auto"}` whenever `request.tool_choice == "none"` AND tools are present. No-tools callers (naming, compaction summary, server operator) never reach the branch and still carry no `tool_choice` key at all. OpenAI-compatible and Gemini keep a literal `none` (no documented cache penalty there, and Gemini's mode MUST stay `NONE` because its default with tools present is to allow calls).
- The "reads the turn, calls nothing" contract moves to where it is actually enforceable — the callers. Both `complete_aside` and `advise_compaction` already consume only `StreamTextDelta`/`StreamUsageEvent`; a `tool_use` block in the answer is inert. That is now documented at both sites.
- The appended prompts (`ASIDE_PROMPT`, `LOOP_JUDGE_PROMPT`, the advisor's instructions + closing line) now say explicitly to **answer in text and not call a tool**.
[...]

PR #537, opened 2026-09-02
fix(preflight): keep a session on its sticky account under a reserve verdict

damianvtran:
## Summary of Changes

**Change class: C1 (standard / pre-authorized).** Bug fix / reliability; patch bump `0.44.32 → 0.44.35` (rebased onto #527's 0.44.32 / tag `v0.44.32`; sequenced after #532 → 0.44.33 and #534 → 0.44.34 — the merger re-sequences the patch numbers by merge order).

Anthropic's prompt cache is isolated per organization/workspace, i.e. per OAuth account. This machine load-balances ~16 sessions plus subagents across 5 accounts. When the quota preflight found a session's **current** account in reserve (≤ `usageReservePercent` remaining, default 10 %) with an unblocked sibling, it called `deprioritize_credential` (process-wide, 120 s TTL), `_usable_key_rows` dropped the demoted row from its tier, `_base_selection_order` could no longer find the sticky, fell to the crc32 hash, and `_resolve` re-pinned the session onto a sibling whose cache had never seen the conversation. The only part of the prefix that hit was the shared tools+system head (~22–40 k tokens, warm on every account); the conversation (150–500 k tokens) was rewritten at cache-write price. With three of five accounts above 90 % of the 5 h window, the mark expired and re-fired at every message boundary — ping-pong between cold caches.

The reactive 429 path already got this right (`rotate_sibling`: "Sticky preserved: same account stays first after backoff"). The preflight's reserve path contradicted it.

**Principle implemented: a reserve verdict is a preference for NEW picks, never a reason to move a session already transacting on an account. Only a depleted verdict (0 %) may move a session, because there the rewrite is unavoidable.** On this PR alone a *new* session that lands on a low account still moves off it through its own boundary walk (demote fresh pick + release pin + re-pick); an up-front *avoidance* of low accounts at first pick is #533's usage-ranked first resolve, not this change.

### `AuthStore` (`providers/auth_store.py`)

- `_usable_key_rows` keeps a demoted row when it is the session's sticky (`session_id` is now threaded from `_resolve`). A **blocked** sticky still drops — the block filter runs first and blocks are verdicts that the account cannot serve.
- `_selection_order` exempts the sticky from the sort-last for the same reason. `_usable_key_rows` judges "every row in this tier is demoted" on the **raw** rows before applying the exemption (review F3): an all-demoted tier still yields empty, so `_resolve`'s second pass clears the tier's marks together and the sticky wins the re-resolve — instead of the exemption handing `_selection_order` a one-row tier and clearing only the sticky's mark. The `read_only` gate is unchanged.
- Two public seams beside `pin_session_credential`: `session_credential_id(provider, session_id)` (read) and `release_session_credential(provider, session_id)` (clear). Both storage-id normalised like the write.
- A session whose OWN request 529'd still rotates: `rotate_sibling` clears that session's sticky for a server fault before the mark is consulted, so the exemption has nothing to keep (asserted).

### Preflight (`model/configure.py`)

- `preflight_usage` captures `boundary_sticky_id` **before** its first resolve — every resolve in the walk re-pins the session, so the store's live sticky stops saying "where is my cache warm" after the first hop. It is a walk-local threaded through `_apply_account_health` like `attempted_ids` (review F2): the stream fn is shared by parent and child sessions and re-entered by a mid-turn `/model` probe, so an instance attribute could be overwritten by a second walk while the first is suspended.
- `_apply_account_health`: reserve on the boundary sticky with a sibling available → **stay**: one deduped `account:reserve` info line (`anthropic quota low (8% remaining) — staying on this account to keep the prompt cache warm`), route settled on the primary, no demotion, no block, **and no same-provider effort hop** (review F4: Anthropic invalidates the cached message prefix when thinking parameters change, which would rewrite the very conversation the stay protects). Shares `_settle_on_reserve_account(keep_effort=...)` with the existing last-account rule so the two cannot drift; the last-account branch still takes the effort hop, because there nothing else can serve and the per-request saving outweighs one rewrite. The two stays alias on the `account:reserve` token on purpose (both mean "low, still serving here").
- Reserve on a row the walk only just resolved onto (a **fresh pick**, nothing cached) still demotes — via `_demote_fresh_pick`, which also releases the pin. Without the release the store would keep serving the demoted sticky, `attempted_ids` would already hold it, and the walk would end parked on the account it meant to leave. This branch is reachable from both callers (the boundary probe and the blocked-row recovery), so it is not dead.
- Depleted stays as it was: block written, session moves.
- Same split on the model-tier branch (scoped weekly cap in reserve).
- The `usageReservePercent` help text and the failover guide now describe the reserve as a new-pick preference and a low-quota notice, not an eviction.

## Related Issue(s)

- Related: usage-skew / "All 5 OAuth credentials unusable" investigation. Sibling PRs: #534 (advisor/aside `tool_choice` cache split — ruled the compaction advisor out live), #533 (usage-aware least-loaded first pick), #532 (1h cache TTL on large contexts).

## Impact

- `AuthStore._usable_key_rows` gains a keyword-only `session_id=None`; the three cascade call sites pass it. Two new public methods; nothing removed.
- Behaviour change for users with several accounts on one provider and `usageAwareFallback: true` (NOT the shipped default): a running session no longer hops accounts when its account crosses the reserve threshold. It hops on depletion, as before. New sessions still prefer healthier accounts.
- Cost: one dict read per boundary (`session_credential_id`). No provider calls added.

### Gates (whole tree, as CI)

```
.venv/bin/python -m flake8 .                                  → clean (rc 0)
uvx --from black==26.1.0 black --check .                       → 799 files would be left unchanged
uvx isort==5.13.2 --check .                                    → clean (rc 0)
.venv/bin/python -m pyright --pythonpath .venv/bin/python .    → 0 errors, 0 warnings, 0 informations
```

### Unit tests

`tests/unit/providers tests/unit/model tests/unit/session -n 4` (at 6727102f, rebased onto 47bc768f): **1864 passed, 1 failed** — `test_catalogue.py::test_the_gemini_wire_reaches_the_spec_through_a_real_transport[gemini-2.5-pro]`, which is **pre-existing on main** and reproduced 3/3 on the untouched base tree (`git stash` of the whole change) with the same command. Mechanism, bisected: `test_clients.py::test_google_sends_no_effort_key` calls the real `build_model_spec("google", "gemini-2.5-pro")` and warms the process-wide `_resolve_model_info_cached` lru; when worksteal puts the gemini wire test on the same worker the memo hit skips discovery and its `http.calls` assertion fires. Passes with `--dist load` (2/2), standalone (6/6), and `tests/unit/model` alone (3/3). Not touched here — out of slice.

`tests/unit` (full): 3 fails on the first run were the 2 `WatchedAuth` fakes in `test_effort_classifier.py` needing the new `session_credential_id` seam (fixed) + the gemini flake above; the 3 `test_episode_subprocess.py` errors ("no usable copied interpreter") are a host-environment issue with `venv --copies` on this uv Python and pass on re-run (13/13, and `tests/unit/evaluation` 512/512).

New tests, all **red on the pre-fix tree** (sources stashed, tests kept — 11 failures, each the defect assertion `credential_id == 2` where 1 was expected, or the missing seam):

- `test_auth_store.py::TestADemotionNeverMovesASessionOffItsStickyAccount` (11): (a) sticky + demoted + sibling → still sticky, mark untouched; (b) no sticky + demoted → every session hash skips the demoted row; (c) blocked sticky → moves; blocked+demoted → moves; (d) `rotate_sibling` usage-limit still preserves sticky; exemption is per session (a "cold" session still moves); `release_session_credential` lets the demotion take effect; `read_only` resolve stays on the demoted sticky without clearing; all-demoted stale-marks rule ignores the exemption; sticky sorts first while demoted; a 529 on the sticky still rotates.
- `test_configure.py` (5): reserve on sticky → no demotion, no block, exactly one notice, silent on the next boundary, session stays on A while a fresh session picks B; depleted on sticky → block + move + re-pin; reserve on fresh picks across 3 accounts → probes A, B, C exactly once each, lands on C, walk terminates, silent; sticky on C found in reserve first → settles immediately (`probed == ["oauth-c"]`); model-tier reserve on sticky → stays with the `for claude-fable-5 … staying on this account` line, fresh session moves.
- **Round-1 remediation tests** (all three red at the pre-fix head 8a9e366f with the new tests copied in — `1 == 2` for the parent's account, `FallbackTarget(... effort='low') is None`, `{2} == set()` — green at 6727102f): `test_configure.py::test_concurrent_walks_on_one_stream_fn_each_keep_their_own_boundary_sticky` (F2: parent walk parked in its first usage fetch after pinning fresh pick A; child boundary runs a whole walk on the same stream fn; parent resumes and still demotes A and moves to B — `probed == [a, a, b]`), `test_configure.py::test_a_warm_stay_keeps_the_current_effort` (F4: effort chain entry offered, warm stay leaves `_route_state.active is None`), `test_auth_store.py::…::test_an_all_demoted_tier_clears_both_marks_through_resolve` (F3: the reviewer's reproduction — sticky A, A+B demoted, resolve → A and `_active_demotions == set()`, then six non-sticky read-only picks spread over both rows).
- `test_distinct_quota_conditions_on_one_selector_do_not_alias` / its bracket companion: the phase-2 expected line updated to the new stay-on-sticky text (the session is sticky to that account from phase 1). Still asserts both scoped tokens announce once and a state-only token would alias.

### Reproduction (unit-level harness, `preflight_usage` driven end to end)

Two OAuth rows, session sticky on A, A at 8 % remaining, B at 60 %:

**Before (base, pre-fix tree):**
```
sticky before preflight : A
resolved after preflight: B
demoted ids             : ['A']
blocked                 : A=False B=False
notices                 : (none)
verdict                 : MOVED (cache cold on B)
fresh session picks     : B
```

**After (this PR):**
```
sticky before preflight : A
resolved after preflight: A
demoted ids             : []
blocked                 : A=False B=False
notices                 : ['[info] anthropic quota low (8% remaining) — staying on this account to keep the prompt cache warm']
verdict                 : STAYED on A
fresh session picks     : B
```

### Live proof of the premise (per-account cache isolation)

Three requests on `claude-haiku-4-5-20251001` through the machine's real shared OAuth store, over the exact wire body the harness builds (`AnthropicClient._build_body`/`_headers`, so `cache_control` markers ride along), ~5.8 k-token unique body; `message_start.usage` read straight off the SSE:

```
{"label": "1: first call on X (writes X's cache)",          "account": "<acct X>", "input": 3, "cache_creation": 5787, "cache_read": 0}
{"label": "2: same conversation on Y (never seen it)",     "account": "<acct Y>", "input": 3, "cache_creation": 5787, "cache_read": 0}
{"label": "3: same conversation again on X (full read)",   "account": "<acct X>", "input": 3, "cache_creation": 0,    "cache_read": 5787}

step 2 on Y: rewrote 5787 tokens, read 0 -> CONFIRMED: the cache is per account
step 3 on X: read 5787 of the 5787 written in step 1 -> CONFIRMED: the write lives on X
```

The harness's own `AnthropicClient.stream` path reports the same numbers for this body (`input=3 cache_read=5787 cache_write=0` on a fourth call to X); the proof reads `message_start` directly only so the three usage dicts are the provider's own words.

### Analytics evidence (`~/.local-operator/analytics.db`, read-only)

Head-only-hit events over the last 30 h — `cache_read` in the shared-head band (10–60 k), conversation rewritten (`cache_write` > 60 k), context ≥ 100 k:

```sql
WITH w AS (SELECT * FROM calls WHERE provider='anthropic' AND ts_ms >= (strftime('%s','now') - 30*3600)*1000),
moved AS (SELECT * FROM w WHERE cache_read_tokens BETWEEN 10000 AND 60000 AND cache_write_tokens > 60000 AND context_tokens >= 100000)
SELECT 'all anthropic calls', COUNT(*), SUM(cache_write_tokens) FROM w
UNION ALL SELECT 'head-only-hit events', COUNT(*), SUM(cache_write_tokens) FROM moved
UNION ALL SELECT 'share of cache writes (%)', NULL, ROUND(100.0*(SELECT SUM(cache_write_tokens) FROM moved)/(SELECT SUM(cache_write_tokens) FROM w),1)
UNION ALL SELECT 'distinct sessions affected', COUNT(DISTINCT session_id), NULL FROM moved;
```
```
all anthropic calls         42406  269751307
head-only-hit events          374  102005294
share of cache writes (%)            37.8
distinct sessions affected     22
```
(The brief's "368 / ~100M / ~half" was taken a few hours earlier; the window has since drifted to 374 / 102M / 37.8 % of Anthropic cache writes.)

Gap to the previous call in the same session, for those events — 313 of 374 are under five minutes, so not 5-minute TTL expiry:
```
< 5 min (not TTL expiry)  313   82616698
>= 5 min                   58   17979484
first call                  3    1409112
```

Session `1079f51b0059` (local time), the cited 20:31 pair — a full hit, then 2 s later a head-only hit with the conversation rewritten, then 6 s later again:
```
20:31:18  ctx=301892  cr=301557  cw=333     gap=9s
20:31:20  ctx=302657  cr= 26932  cw=275721  gap=2s   ← moved
20:31:24  ctx=303410  cr=301890  cw=1518    gap=3s   ← reads the NEW entry in full
20:31:37  ctx=303784  cr=303408  cw=374     gap=12s
20:31:44  ctx=311648  cr= 25939  cw=285707  gap=6s   ← moved again
```

Child session `a2384d9fbb88` (subagent `pr2-residency`; its analytics rows are recorded under the parent `0a08dc6d7624`, since a child shares the parent's `SessionStreamFn`), from its transcript (UTC), every big cache write in the window:
```
02:42:53  +567s  ctx=229597 cr=22521 cw=207074
02:43:01    +7s  ctx=229841 cr=22521 cw=207318
02:50:20    +6s  ctx=230924 cr=22521 cw=208401
02:51:55   +71s  ctx=237283 cr=22521 cw=214758
```
`cr=22521` is exactly the tools+system head; four rewrites of a ~210 k conversation within 9 minutes, three of them 6–71 s after the previous call.

### Not done

- No live multi-session soak with the fix installed (the runtime is pinned to `main`; the fix is validated at the `preflight_usage` seam and through the real `AuthStore` cascade). After merge + `lop-update`, the same analytics query should show the head-only-hit share collapsing.
- The gemini test-ordering flake noted above is an adjacent finding, deliberately left alone.

## Checklist

- [x] All new and existing tests pass (modulo the pre-existing gemini ordering flake, documented above).

### DIFF
--- a/local_operator/guides/failover/GUIDE.md
+++ b/local_operator/guides/failover/GUIDE.md
@@ -35,7 +35,9 @@ first, then a per-session hash of the session id so concurrent sessions start on
 different accounts rather than stampeding one, then round-robin for callers with
 no session. An account that just failed on a provider-side fault is *demoted*
 (sorted last), never removed — a 500 is not the account's fault, so it stays
-available as a last resort.
+available as a last resort. A demotion never applies to the account a session
+is already sticky to: it steers *new* picks, and a session's warm prompt cache
+lives on its current account (see "Quota reserve" below).
 
 When a request fails on quota, `_recover_quota_blocked` (in
 `providers/failover.py`) probes the *other* accounts' own usage before the chain
@@ -62,10 +64,22 @@ rather than on quota forecasts. Turn it on to spend one lightweight quota
 request per user-message boundary in exchange for routing that leaves a
 provider before it fails.
 
-`usageReservePercent` (default 10) holds back a slice of quota. Two rules that
-are easy to get backwards:
+`usageReservePercent` (default 10) is the headroom below which an account
+counts as **low**. Three rules that are easy to get backwards:
 
-- Reserve quota can select a **lower-effort route on the same provider** without
+- **Low is a preference for new picks, never an eviction.** The provider's
+  prompt cache is per account, so a session already transacting on a low
+  account **stays there** — it is told once (`quota low (N% remaining) —
+  staying on this account to keep the prompt cache warm; new sessions will
+  prefer other accounts`) and keeps spending that account down to zero. Moving
+  it would rewrite its whole conversation prefix at cache-write price on an
+  account that has never seen it; on a multi-account host that rewrite was
+  measured at ~38% of all cache writes. The low account is demoted for
+  *other* sessions' first picks (and for a session that only just landed on
+  it with nothing cached), and the reactive 429 path keeps the same promise
+  (`rotate_sibling`: "sticky preserved"). Only a **depleted** verdict (0%
+  remaining) moves a running session, because then the rewrite is unavoidable.
+- Low quota can select a **lower-effort route on the same provider** without
   blocking the account. Dropping from high to low effort on the model the user
   chose is cheaper than leaving for another vendor, so the reserve buys a
   cheaper route before it buys a different provider.
--- a/local_operator/guides/failover/GUIDE.md
+++ b/local_operator/guides/failover/GUIDE.md
@@ -37,7 +37,11 @@ no session. An account that just failed on a provider-side fault is *demoted*
 (sorted last), never removed — a 500 is not the account's fault, so it stays
 available as a last resort. A demotion never applies to the account a session
 is already sticky to: it steers *new* picks, and a session's warm prompt cache
-lives on its current account (see "Quota reserve" below).
+lives on its current account (see "Quota reserve" below). Because the marks
+are process-wide, this also means *another* session's 500 on account A no
+longer moves a session that is sticky to A — only that session's **own** 500
+does, since `rotate_sibling` clears its sticky for a server fault before the
+mark is consulted.
 
 When a request fails on quota, `_recover_quota_blocked` (in
 `providers/failover.py`) probes the *other* accounts' own usage before the chain
@@ -70,15 +74,20 @@ counts as **low**. Three rules that are easy to get backwards:
 - **Low is a preference for new picks, never an eviction.** The provider's
   prompt cache is per account, so a session already transacting on a low
   account **stays there** — it is told once (`quota low (N% remaining) —
-  staying on this account to keep the prompt cache warm; new sessions will
-  prefer other accounts`) and keeps spending that account down to zero. Moving
-  it would rewrite its whole conversation prefix at cache-write price on an
+  staying on this account to keep the prompt cache warm`) and keeps spending
+  that account down to zero, at its current effort (a same-provider
+  lower-effort hop would invalidate the cached conversation too). Moving it
+  would rewrite its whole conversation prefix at cache-write price on an
   account that has never seen it; on a multi-account host that rewrite was
-  measured at ~38% of all cache writes. The low account is demoted for
-  *other* sessions' first picks (and for a session that only just landed on
-  it with nothing cached), and the reactive 429 path keeps the same promise
-  (`rotate_sibling`: "sticky preserved"). Only a **depleted** verdict (0%
-  remaining) moves a running session, because then the rewrite is unavoidable.
+  measured at ~38% of all cache writes. The reactive 429 path keeps the same
+  promise (`rotate_sibling`: "sticky preserved"). Only a **depleted** verdict
+  (0% remaining) moves a running session, because then the rewrite is
+  unavoidable. A *new* session that lands on the low account with nothing
+  cached still moves off it: its own boundary walk demotes the row and
+  re-picks. That is per session, not a standing preference — the demotion
+  written by the walk is short-lived, and the stay branch writes none — so
+  new sessions do not yet *avoid* a low account up front; that needs the
+  usage-ranked first pick (`retry.usageAwareAccountPick`), a separate change.
 - Low quota can select a **lower-effort route on the same provider** without
   blocking the account. Dropping from high to low effort on the model the user
   chose is cheaper than leaving for another vendor, so the reserve buys a
--- a/local_operator/model/configure.py
+++ b/local_operator/model/configure.py
@@ -1665,15 +1665,6 @@ class SessionStreamFn:
         self._primary_selector: str | None = None
         self._usage_checked_selector: str | None = None
         self._usage_checked_at = 0.0
-        # The credential this session was sticky to when the CURRENT boundary
-        # walk began — i.e. the account whose prompt cache holds this
-        # conversation. Captured by ``preflight_usage`` ahead of its first
-        # resolve, because every resolve in the walk re-pins the session, so
-        # the store's live sticky stops answering "where is my cache warm?"
-        # after the first hop. Read by ``_apply_account_health`` to decide
-        # whether a reserve verdict is about a warm account (stay) or a fresh
-        # pick (move on).
-        self._boundary_sticky_id: int | None = None
         # Quota is re-probed at EVERY user-message boundary, but the user only
         # needs to hear about a CHANGE in quota standing — not the same "quota
         # low"/"quota exhausted" line echoed on every message they send while
@@ -2455,7 +2446,17 @@ class SessionStreamFn:
         # ``_apply_account_health``); against any other row it is a fresh pick
         # the walk may still move. ``None`` on a session's very first boundary
         # — nothing is warm anywhere, so every row is a fresh pick.
-        self._boundary_sticky_id = self._auth_store.session_credential_id(
+        #
+        # A LOCAL, threaded through ``_apply_account_health`` like the memos
+        # above, never an attribute: this stream fn is shared by a parent and
+        # its child sessions (``harness/subagent.py`` hands the child the
+        # parent's stream fn), and a mid-turn ``/model`` probe can re-enter
+        # here while a walk is suspended at an ``await``. A second entrant
+        # would re-read the store sticky — by then the first walk's fresh pick,
+        # since every resolve re-pins — and an attribute would hand that pick
+        # back to the first walk as "warm", settling it on a reserve account it
+        # meant to move off.
+        boundary_sticky_id = self._auth_store.session_credential_id(
             model.provider, self._session_id
         )
         while True:
@@ -2494,6 +2495,7 @@ class SessionStreamFn:
                             attempted_ids,
                             quota_cache,
                             usage_memo,
+                            boundary_sticky_id,
                         ):
                             continue
                         return
@@ -2646,6 +2648,7 @@ class SessionStreamFn:
                             attempted_ids,
                             quota_cache,
                             usage_memo,
+                            boundary_sticky_id,
                         ):
                             continue
                         return
@@ -2667,7 +2670,7 @@ class SessionStreamFn:
                         # rotation, which spammed the transcript on every
                         # boundary. Rotate silently.
                         continue
-                    if access.credential_id != self._boundary_sticky_id:
+                    if access.credential_id != boundary_sticky_id:
                         # Reserve on a FRESH pick: steer new picks elsewhere.
                         # Same rule and reasoning as the account-scope path in
                         # ``_apply_account_health``.
@@ -2682,8 +2685,7 @@ class SessionStreamFn:
                         selector,
                         f"model:{health.state}",
                         f"{model.provider} {condition}{remaining} for {model.model_id} "
-                        "— staying on this account to keep the prompt cache warm; "
-                        "new sessions will prefer other accounts",
+                        "— staying on this account to keep the prompt cache warm",
                         "info",
                     )
                     await self._route_state.clear_settled("primary model still has quota")
@@ -2744,6 +2746,7 @@ class SessionStreamFn:
                 attempted_ids,
                 quota_cache,
                 usage_memo,
+                boundary_sticky_id,
             ):
                 continue
             return
@@ -2760,6 +2763,7 @@ class SessionStreamFn:
         attempted_ids: set[int],
         quota_cache: dict[str, str] | None = None,
         usage_memo: "dict[str, UsageReport | None] | None" = None,
+        boundary_sticky_id: int | None = None,
     ) -> bool:
         """Act on a low/depleted account-scope verdict.
 
@@ -2772,6 +2776,11 @@ class SessionStreamFn:
         ``usage_memo`` is the same walk's per-account report memo, threaded
         for the same reason one level down: the fallback chain may list this
         walk's own provider, whose accounts have already been read.
+        ``boundary_sticky_id`` is the credential the session was sticky to
+        when the walk BEGAN — the account its prompt cache is warm on — and
+        is walk-local for the reason ``preflight_usage`` gives where it reads
+        it; the default ``None`` (nothing warm) is only for callers outside a
+        walk.
 
         The binding windows that produced ``health`` can be scoped to a model
         tier while the shared windows still hold quota (Anthropic's
@@ -2846,14 +2855,14 @@ class SessionStreamFn:
         # preserved"), applied to the preflight.
         #
         # "Already transacting on" is the sticky captured BEFORE this
-        # boundary's walk started (``_boundary_sticky_id``): the walk's own
+        # boundary's walk started (``boundary_sticky_id``): the walk's own
         # resolves re-pin the session as they go, so a row the walk only just
         # landed on is a FRESH pick with nothing cached, and demoting it to
         # keep walking (below) is still right. Both callers of this method
         # arrive with a row the walk resolved — the boundary probe and a
         # blocked-row recovery — so the fresh-pick branch is reachable from
         # both.
-        on_warm_account = access.credential_id == self._boundary_sticky_id
+        on_warm_account = access.credential_id == boundary_sticky_id
         if not siblings and health.state == "reserve":
             # Last account on this provider, still holding spendable quota.
             # Crossing the reserve threshold used to hop to the next chain
@@ -2866,6 +2875,7 @@ class SessionStreamFn:
                 fallback,
                 f"{model.provider} {condition}{remaining} — continuing until "
                 f"{model.provider} quota is exhausted",
+                keep_effort=False,
             )
             return False
 
@@ -2930,6 +2940,7 @@ class SessionStreamFn:
                     attempted_ids,
                     quota_cache,
                     usage_memo,
+                    boundary_sticky_id,
                 )
 
         if not siblings and fallback is None:
@@ -2980,11 +2991,16 @@ class SessionStreamFn:
             # lone-account rule because that one must also hold when nothing
             # is cached yet. This is the branch the rotation below used to
             # take for every low account, warm or not.
+            #
+            # ``keep_effort``: the whole point of staying is the warm cache,
+            # and a same-provider effort hop would spend part of it — see
+            # ``_settle_on_reserve_account``.
             await self._settle_on_reserve_account(
                 model,
                 fallback,
                 f"{model.provider} {condition}{remaining} — staying on this account to "
-                "keep the prompt cache warm; new sessions will prefer other accounts",
+                "keep the prompt cache warm",
+                keep_effort=True,
             )
             return False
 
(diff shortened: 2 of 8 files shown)


damianvtran:
### Agent review — round 1
- Scope: f3812145..8a9e366f3312db8f40f82b88d2f025cf8b02bdaf
- Mode: Comprehensive
- Delivery contract: a reserve quota verdict must never move a session off the account its prompt cache is warm on; only depleted/blocked may
- Applicable risk surfaces: credential routing, quota exhaustion, walk termination, fail-open, config docs
- Verdict: Changes requested

Verified in a throwaway worktree at 8a9e366f: flake8 / black / isort / pyright all rc 0; `tests/unit/providers tests/unit/model tests/unit/session -n 4` → 1803 passed, 1 failed (`test_catalogue …gemini-2.5-pro`, passes 3/3 standalone — the documented worksteal flake, not counted). New tests confirmed red on the base tree with the fix stashed: 6/11 in `TestADemotionNeverMovesASessionOffItsStickyAccount` and 4 in `test_configure.py` fail with the defect assertion (`credential_id == 2`, `[2, 1] == [1, 2]`) or the missing seam. Analytics query re-run read-only: 34,536 calls / 312 head-only-hit events / 83.8M cache-write tokens / 35.5% / 19 sessions — the window has drifted since the body was written, same shape. The delivery contract holds on every path I traced: sticky reserve → stay without demotion or block; depleted → block + move; fresh-pick reserve → demote + release, walk advances; three-in-reserve → probes each once and lands on the last with the last-account rule; own-request 529 → `rotate_sibling` clears the pin first, so the exemption has nothing to keep.

| --- | --- | --- | --- |
| F1 | MAJOR | release/version | `pyproject.toml` bumps to `0.44.31`, but `origin/main` (#538, f82d5527) already ships `0.44.31` and the tag `v0.44.31` already points at f82d5527. The merge is clean (both sides changed the line to the same text) so nothing will flag it, and the merged tree would carry an already-released version. Rebase onto f82d5527 and bump to `0.44.32`. |
| F2 | MINOR | walk correctness under interleaving | `configure.py:2481` stores `_boundary_sticky_id` on the shared `SessionStreamFn` instance, but a child session runs on the parent's stream fn (`harness/subagent.py:1341`) and a mid-turn `/model` probe (`consume_boundary=False`) can enter `preflight_usage` while a walk is suspended at an `await`. The second entrant re-reads the store sticky, which by then is the first walk's fresh pick (every resolve re-pins), and overwrites the attribute — the first walk then treats its fresh pick as warm and settles on a reserve account instead of moving on. Not a contract breach (nothing warm is moved) but a wrong stay. Make it a local in `preflight_usage` threaded through `_apply_account_health`, like `attempted_ids`. |
| F3 | MINOR | stale-marks invariant | The "all-demoted → clear the tier's marks" rule is unchanged for a direct `_selection_order` call (which is what the new test asserts), but through `_resolve` it now sees the post-exemption tier: sticky A, A and B both demoted → `_usable_key_rows` returns `[A]` alone → `_selection_order` judges a one-row tier "all demoted" and clears only A's mark; B stays demoted. Reproduced: `_active_demotions` is `{B}` after the sticky session's resolve, and a non-sticky session then resolves to A too. Pre-fix both marks cleared together and picks spread by hash; now an outage's marks decay asymmetrically and other sessions skew onto the sticky's account for the 120 s TTL. Either judge "all demoted" on the raw rows inside `_usable_key_rows` before applying the exemption, or add a `_resolve`-level test that documents this as intended. The PR text "that invariant … unchanged" is not accurate at the resolve level. |
| F4 | MINOR | warm-cache purpose | `_settle_on_reserve_account` (configure.py:3075) takes a same-provider lower-effort hop before staying, on the warm-account branch too. Anthropic invalidates cached *message* prefixes when thinking parameters change (system/tools stay cached), which the code already knows (`configure.py:1671`, effort frozen per tool loop) — so on a chain that offers a same-provider effort step the "stay to keep the cache warm" branch rewrites the conversation once. One rewrite, not ping-pong, hence minor; either skip the effort hop when `on_warm_account`, or say in the docstring that this stay costs one rewrite. |
| F5 | MINOR | docs vs mechanism | The stay branch writes no demotion, and any mark on the sticky's account is erased by `_clear_demotion` after the warm session's next served request (`failover.py:2054`). So "new sessions will prefer other accounts" (notice, `settings_io.py:459`, guide) is delivered by each new session's own fresh-pick walk (hash onto the low account → probe → demote+release → move), not by a standing preference. Functionally fine; the wording overclaims until #533's usage-aware first pick lands. Soften the help/guide to "new sessions move off it" or leave as is once #533 merges. |

#### Follow-up observations

- N1 (NIT): the "last account, continuing until exhausted" line and the "staying on this account to keep the prompt cache warm" line now share the `account:reserve` token, so a session that transitions between the two conditions (siblings blocked/unblocked while low) hears only the first. The docstring says the token only needs to stay distinct from the model-tier branch; if that is the intent, say the two account-scope stays alias on purpose.
- N2 (NIT): by design a sibling session's 529 demotion of A no longer moves a session sticky on A (only its own 529 does). That is consistent with "a 500 is not the account's fault", but the guide's new paragraph should say so explicitly, since pre-fix it would have moved.
- N3 (NIT): `test_reserve_on_a_fresh_pick_moves_on_and_the_walk_terminates` ends with the session on C under the last-account rule, not on a healthy row; a fourth healthy account would pin the "walk keeps going past a reserve fresh pick and stops on healthy" case more sharply.
- The analytics evidence and the per-account cache-isolation proof in the body both check out; nothing in the diff reads secrets or widens any public surface beyond the two documented seams.

F1 is a one-line fix; once the version is re-sequenced and the branch rebased, no blocker or major remains and round 2 is terminal on the version bump alone — F2–F5 and the nits are follow-ups at the author's discretion.

damianvtran:
- Reviewed head: 8a9e366f → remediated head: **6727102f** (rebased onto `origin/main` 47bc768f; one remediation commit on top of the rebased original)

| ID | Disposition | Commit | Note |
| --- | --- | --- | --- |
| F1 | **Fixed** | 6727102f | Rebased onto current `main`, which by the time of the rebase had moved again to #527 / `0.44.32` (tag `v0.44.32`), so the collision the review found (`0.44.31`) is gone and the suggested `0.44.32` would now collide too. Sequenced across the open PRs as #532 → 0.44.33, #534 → 0.44.34, **this PR → 0.44.35**; the merger re-sequences by merge order. Body updated. |
| F2 | **Fixed** | 6727102f | `_boundary_sticky_id` is no longer an attribute on the shared `SessionStreamFn`; `preflight_usage` reads it into a local `boundary_sticky_id` and threads it through all four `_apply_account_health` call sites (same pattern as `attempted_ids` / `quota_cache` / `usage_memo`). Test `test_concurrent_walks_on_one_stream_fn_each_keep_their_own_boundary_sticky`: parent walk parked in its first fetch after pinning fresh pick A, a child boundary walk runs to completion on the same stream fn, parent resumes and still demotes A and moves to B (`probed == [a, a, b]`). Red at 8a9e366f (`1 == 2`), green now. |
| F3 | **Fixed** | 6727102f | Took the fix, not the doc route: `_usable_key_rows` now judges "every row in this tier is demoted" on the raw rows *before* the sticky exemption and yields `[]`, so `_resolve`'s second pass clears the tier's marks together and the sticky wins the re-resolve. Your reproduction is `test_an_all_demoted_tier_clears_both_marks_through_resolve` (sticky A, A+B demoted → resolve lands on A, `_active_demotions == set()`, six non-sticky read-only picks spread over both rows). Red at 8a9e366f (`{2} == set()`), green now. PR body's "unchanged" claim corrected to describe the mechanism. |
| F4 | **Fixed** | 6727102f | `_settle_on_reserve_account(..., *, keep_effort)`: the warm-account stay passes `True` and never takes the same-provider effort hop (the effort change invalidates the cached message prefix — the same reason `_message_effort` is frozen per tool loop). The last-account stay passes `False` and keeps the hop: nothing else can serve there, so one rewrite buys a lower per-request spend for the rest of the window, whereas on the warm branch a healthy sibling exists and the whole point of staying is the cache. Docstring states both. Test `test_a_warm_stay_keeps_the_current_effort` (effort chain entry offered, `_route_state.active is None`). Red at 8a9e366f, green now. |
| F5 | **Fixed** | 6727102f | Notice (both the account-scope and model-tier lines) is now just "… — staying on this account to keep the prompt cache warm". Settings help: "Headroom below which an account counts as low; a running session stays on it." Guide: says a new session that lands on a low account moves off it via its own walk, that this is per session and not a standing preference, and attributes up-front avoidance to `retry.usageAwareAccountPick` as a separate change (#533). PR body reworded the same way. |
| N1 | Fixed | 6727102f | `_settle_on_reserve_account` docstring now says the two account-scope stays alias on `account:reserve` on purpose — both mean "low, still serving here", so a transition between them is not owed a second line. |
| N2 | Fixed | 6727102f | Guide's demotion paragraph states explicitly that another session's 500 on A no longer moves a session sticky to A; only its own 500 does (via `rotate_sibling` clearing the sticky). |
| N3 | Deferred | — | Left as is: the three-account shape already pins the fresh-pick → move → terminate path; a fourth healthy row is a sharper assertion but the same mechanism. Noted in the handoff for a follow-up if the reviewer wants it in this PR. |

Verification at 6727102f: whole-tree `flake8` / `black --check` (799 unchanged) / `isort --check` / `pyright` all rc 0; `tests/unit/providers tests/unit/model tests/unit/session -n 4` → **1864 passed, 1 failed** (`test_catalogue …[gemini-2.5-pro]`, pre-existing worksteal flake, 3/3 standalone). `gh pr view --json mergeable` → MERGEABLE; CI started on the new head.

damianvtran:
### Agent review — round 2
- Scope: 47bc768f..6727102f0ad673ac03e98f9bed7cf0d823edd5bb
- Delivery contract: a reserve quota verdict must never move a session off the account its prompt cache is warm on; only depleted/blocked may
- Delta reviewed: 8a9e366f..6727102f
- Applicable risk surfaces: credential routing, quota exhaustion, walk termination, fail-open, config docs
- Verdict: Approve

Every round-1 disposition is genuine. F1: `origin/main` is 47bc768f / `0.44.32` (tag `v0.44.32` is the newest), head carries `0.44.35`; no collision exists. F2: `_boundary_sticky_id` is gone from `SessionStreamFn`; `preflight_usage` reads the local before its first resolve (configure.py:2459) and passes it at all four `_apply_account_health` sites (2498 boundary-recovery, 2651 model-tier recovery, 2749 boundary, 2943 the depleted-recovery recursion) plus the direct model-tier comparison at 2673; no caller exists outside the walk, so the `None` default is never taken in practice. F3: `_usable_key_rows` judges "all demoted" on the raw rows (auth_store.py:1019) and yields `[]`, so `_resolve`'s second pass clears the whole provider's marks — reproduced through `get_oauth_access`: sticky A, A+B demoted → A, `_active_demotions == set()`; with B blocked and A the only unblocked row, the clear also erases B's mark, which is moot under the block and identical to base. F4: `keep_effort=True` on the warm branch skips the same-provider hop; `False` on the last-account branch keeps the pre-existing behaviour, and the justification (nothing else can serve, per-request thinking savings for the rest of the window vs one rewrite) is sound — it is the status quo that branch's existing test pins. F5/N1/N2 wording matches the mechanism; N3 deferral is reasonable.

| --- | --- | --- | --- |
| F6 | MINOR | credential routing (contract edge) | The F3 fix restores base behaviour for a mixed cascade: sticky OAuth A, A **and** B demoted (e.g. two other sessions' fresh-pick reserve demotions, or 529s), plus a same-provider `api_key` login row → the OAuth tier yields `[]`, tier 4 wins, the session is re-pinned to the api_key row and its conversation is rewritten on that org. Reproduced on 6727102f (`resolve row (3, 'api_key')`, sticky→3, marks `{1, 2}` intact); pre-F3 ee84bcd4 stayed on A; base 47bc768f also moves. With an env `ANTHROPIC_API_KEY` instead, the env leg serves for the mark TTL and `pin(None)` drops the sticky, so the next OAuth resolve is a cold hash pick. Needs every OAuth row demoted at once **and** a lower tier on the same provider, so narrow and not a regression against base — follow-up, not a round. Suggested shape: in `_resolve`, short-circuit to the session's unblocked sticky OAuth row before the tier filter (stickiness is a per-session fact; the all-demoted rule is about new picks). Also soften the `_usable_key_rows` comment "the sticky then wins as usual", which holds only when no later tier serves. |

#### Verified

- Throwaway worktree at 6727102f: `.venv/bin/python -m flake8 .` rc 0; `uvx --from black==26.1.0 black --check .` 802 unchanged rc 0; `uvx isort==5.13.2 --check .` rc 0; `.venv/bin/python -m pyright --pythonpath .venv/bin/python .` 0 errors.
- `tests/unit/providers tests/unit/model tests/unit/session -n 4` → **1864 passed, 1 failed**: `test_catalogue…[gemini-2.5-pro]`, the documented worksteal flake (3/3 standalone), not counted.
- "Red pre-fix" claims reproduced: second worktree at 8a9e366f with the head's two test files copied in — `test_concurrent_walks_on_one_stream_fn_each_keep_their_own_boundary_sticky` fails `1 == 2` (parent left on A), `test_a_warm_stay_keeps_the_current_effort` fails `FallbackTarget(... effort='low') is None`, `test_an_all_demoted_tier_clears_both_marks_through_resolve` fails `{2} == set()`. All three green at 6727102f.
- F3 edge probes through `get_oauth_access` at head: sticky A demoted + B healthy → A, mark kept `{1}`; A+B demoted → A, marks cleared; same under `read_only=True` → A, marks kept; B blocked + A+B demoted → A, marks cleared (B's mark moot under its block); lone demoted sticky → served, cleared. Six non-sticky read-only picks after the clear spread over both rows.
- CI on 6727102f: lint / type-check / pip-audit / tui-e2e ×2 pass; `test (3.12)` still pending at review time. `mergeable: MERGEABLE`.

#### Follow-up observations

- Version sequencing assumes #532 → 0.44.33 and #534 → 0.44.34 merge first; both are still open. Whoever merges out of order must re-bump (a plain merge will not flag it, as round 1 noted).
- The guide (`GUIDE.md:90`) names `retry.usageAwareAccountPick`, which does not exist in this tree — it is #533's setting. If #533 stalls, that line points at nothing; fine as a forward reference while #533 is in flight.
- `_apply_account_health(..., boundary_sticky_id: int | None = None)`: the default silently means "nothing warm" and therefore "demote + release", the destructive direction. Consistent with `quota_cache`/`usage_memo`, but a required keyword would make a future caller that forgets it fail loudly instead of evicting a warm session. NIT.
- The child session in the F2 test shares the parent's `_session_id` (the stream fn's), so parent and child share one sticky by construction — pre-existing design, not this PR, but it is why the F2 interleaving was possible at all.

No blocker or major remains. This round is **terminal**: F6 and the observations above are follow-ups at the author's discretion, not a reason for another round.

PR #540, opened 2026-09-02
feat(tools): size waits to the work — 60-minute wait budget and sizing guidance

damianvtran:
## Summary of Changes

**Change class: C1 (standard / pre-authorized).** Tool-schema/guidance change; patch bump `0.44.32 → 0.44.36` (main is 0.44.32; #532 → .33, #534 → .34, #537 → .35 are open — the merger re-sequences the patch numbers by merge order).

The `wait` tool capped `wait_ms` at 300 000 ms (5 min, also the default). Agents awaiting a CI pipeline (8–20 min) or a long subagent (30–90 min) therefore polled every five minutes, and every poll re-sends the whole context to the model; a poll that lands after the provider's 5-minute prompt-cache TTL also rewrites the prompt cache from scratch. The wait is already interruptible — `execute_wait` returns on job settle, any inbound message for the model (a peer session's `send`, a scheduled wake, a subagent's `hub` note), and steer, and the abort signal cuts it — so a long budget strands nothing and costs nothing when the work finishes early.

- `WaitParams.wait_ms`: ceiling `le=300_000 → 3_600_000` (60 min); default `300_000 → 600_000` (10 min, spans a typical CI run in one call). The field description and the `wait` tool description now carry the sizing rule: estimate how long the job should take and wait for all of it in one call; an expired wait is the signal to check the job, not to re-issue the poll; short budgets only to inspect progress mid-run. The `_await_any_settled` docstring's "up to five minutes" figure is updated to match.
- The sizing rule is stated ONCE, in the `wait` tool description (estimate the job's duration and wait for all of it; an expired wait means check the job, not re-poll; short budgets only to manage progress mid-run, with `jobs op='peek'` between them). `prompts_md/system.md` is one sentence pointing at it; `guides/agents/GUIDE.md` carries the WHY (context re-send + cache rewrite past the 5-minute TTL, with the measured transcript numbers) — guides load on demand, not per turn.
- `hub` `op='ask'` `timeout_ms` (`le=600_000`, default 300 000) is **left alone**: it is a budget for a busy child to finish its current step and answer, sized from measured reply latency (p50 ~3 min), not a wait on a job's whole duration. The same context-cost reasoning applies to a repeated `ask`, but its remedy is `peek`, which the schema already says.
- No TUI change needed: there is no wait-specific elapsed/timeout rendering; `format_duration` is bounded at six cells over the whole domain (`59m59s` / `23h59m`), so a 60-minute wait renders like any other long tool call.
- `agent_seeds/` and `docs/` carry no "5 minute"/"300000"/"generous wait" wording for `wait` (the `300000` hits are the compaction advisor's token threshold, unrelated).

## Related Issue(s)

- Related: measured on this host, not tracked in an issue.

## Motivation — measured (read-only, `~/.local-operator/sessions/*/transcript.jsonl`)

Last 30 h, Anthropic only, as measured by the delegator: **1,488 wait-only model calls** (assistant messages whose only tool call is `wait`); **877 consecutive-wait chains**, **434 of them 2+ polls long** (up to 12). Collapsing each chain to a single wait would have saved **762 model round trips ≈ 207 M context tokens** (3.2 % of Anthropic context tokens) and **69 full prefix rewrites** (20.8 M cache-write tokens, **8.7 % of all Anthropic cache writes**) where the poll landed after the 5-minute prompt-cache TTL.

Re-measured while preparing this PR (all providers, window shifted a few hours; the script walks each transcript, keeps `type == "message"` rows with `payload.role == "assistant"`, and counts runs of consecutive rows whose `tool_calls` are all `wait`):

```
sessions 389  assistant msgs 39282  wait-only calls 1649  chains 881  2+ 438  max 19  saveable round trips 768
```

## Impact

- No breaking change: every previously valid `wait_ms` still validates; only the ceiling and the default move. `3_600_001` is rejected (pinned by test).
- No dependency updates.
- Performance: fewer model round trips and cache rewrites per awaited job; the wait itself is event-driven (`asyncio.wait` on settle/abort/peer events), so a longer park adds no event-loop load. Steering (`lop send --now`) and Esc still cut the wait at once.
- Prompt cost (tiktoken cl100k, measured each round): after review round 1 deduped the sizing rule to ONE statement (the tool description, which is what the model reads at decision time), the always-present prefix costs **+103 tokens/turn vs main** — `system.md` +31, `wait` tool schema (description + parameters) +72 — down from +316 at the first review. The agents guide carries the longer explanation (+249) but loads on demand, not per turn.

### Gates (whole tree, as CI)

```
.venv/bin/python -m flake8 .                                   → rc 0
uvx --from black==26.1.0 black --check .                       → 803 files would be left unchanged (rc 0)
uvx isort==5.13.2 --check .                                    → rc 0
.venv/bin/python -m pyright --pythonpath .venv/bin/python .    → 0 errors, 0 warnings, 0 informations
```

### Unit tests

`tests/unit/tools tests/unit/harness tests/unit/session -n 4` → **1718 passed**, 29 warnings, 33.43 s (at the remediated head).

New `tests/unit/tools/test_wait_budget.py`:
- `3_600_000` validates, `3_600_001` and `0` are rejected; default is `600_000`.
- The JSON schema the model reads advertises `maximum == 3600000`, `default == 600000`, and the sizing rule ONCE (tool description only — the field description is one line; pinned by test).
- A 60-minute wait still returns in <2–3 s on settle, on abort, on steer cancel, and on EACH inbound kind — a peer message, a scheduled wake, a `hub` note — with the payload naming which one woke it (`interrupted_by: "wake_prompt"` etc.; several at once report the neutral `inbound` plus an `arrivals` breakdown).
- `tests/unit/session/test_session_peer.py` pins the producer side: a busy-turn wake, a child's `hub` note, and the mailbox path each mark the arrival signal under their own kind, and one end-to-end test drives a real turn whose in-turn 60-minute `wait` is cut by a fired wake (fails with a TimeoutError on the unfixed path).

Existing coverage the long-budget tests point at: `test_wait_settles_on_events.py` (settle, dispose, abort, multi-id), `test_wait_wakes_on_peer_message.py` (peer message, steer cancel, lost-wakeup) — a stale "300s" in the latter's docstring is updated.

### Live proof (real `local-operator exec` run, branch code on `PYTHONPATH`, Anthropic)

```
$ PYTHONPATH=/tmp/lop-feat-sized-waits .venv/bin/local-operator exec --yolo --json \
  'Do exactly this and nothing else: (1) call the bash tool with background=true and command "sleep 20; echo LIVE_DONE". (2) Then call the wait tool on the returned job id with wait_ms=1800000 (thirty minutes). (3) Reply with one line: the job status text the wait returned and nothing else.'
rc=0 wall=27s

tool_execution_start  wait {'job_id': 'bc40ab0a4758', 'wait_ms': 1800000}
tool_execution_end    wait "job bc40ab0a4758 (bash: sleep 20; echo LIVE_DONE) [completed] | exit code: 0 | --- stdout --- | LIVE_DONE"
final: completed
log span: 09:41:05,842 → 09:41:31,450
```

A 30-minute budget returned ~20 s after the job started (the whole run, including two model calls, took 27 s wall), not at the deadline.

Same run after round 1, the wake path (R1-1):

```
tool_execution_start wait    {'job_id': '43a436a6c6db', 'wait_ms': 1800000}   # job sleeps 120s
wake_delivered               (alarm) Scheduled wake w1: "LIVE_WAKE"
tool_execution_end   wait    duration_s=21.5
  "job 43a436a6c6db still running (a scheduled wake fired — read the reminder before re-waiting)"
  details={"status": "running", "interrupted_by": "wake_prompt", "arrivals": {"wake_prompt": 1}}
```

A 30-minute budget was cut by the 25-second wake: the wait returned in 21.5 s naming the wake, the wake text reached the follow-up model call, and the job kept running. Before the fix this read at the deadline.

Negative control worth recording: the first attempt ran from outside the worktree, where the editable venv imports `~/local-operator` (main), and the model issued `wait_ms: 300000` — i.e. under the old schema the model was clamped to five minutes even when told to wait thirty. The re-run above is against the branch.

### Not done

- `hub` `op='ask'` `timeout_ms` unchanged (different contract, see above).
- No change to the job-row retention sweep (settled rows are kept 5 min); a wait that has parked on a job's settle event wakes on settle regardless of when the sweep runs.

## Checklist

## Screenshots (Optional)

N/A — no rendered UI change.


damianvtran:
### Agent review — round 1
- Scope: 47bc768f..6937a4d2ca8f86d966ca9d51ae4e230b84597f44
- Mode: Comprehensive
- Delivery contract: one sized wait replaces repeated 5-minute polls without stranding a session — settle, abort, peer message and steer still return immediately
- Applicable risk surfaces: tool timeouts, session liveness, prompt cost, schema compatibility
- Verdict: Changes requested

| --- | --- | --- | --- |
| R1-1 | MAJOR | Session liveness — two inbound-message paths do NOT wake a parked `wait`, so their delivery latency inherits the new ceiling (was ≤5 min by accident, now ≤60 min, ≤10 min at the new default). | (a) **Scheduled wakes landing mid-turn**: `session.py:8625-8632` enqueues the wake as *courtesy* steering (`_courtesy_wake_count += 1`), `_has_urgent_steering` (5414) therefore answers False, `_peek_interrupt` (`loop.py:1998`) never cancels the tool, and nothing calls `_peer_arrival.mark()` on that path — a "remind me at 15:00" fired while the agent is inside a 45-minute review wait is read at 15:45. (b) **Child → parent notes**: `comms.reply_to_parent` (`comms.py:1768`) delivers an unprompted "I am blocked / the task is wrong" via `queue_aside`, which only appends to `_aside_thunks` (`session.py:5443`) and never marks the peer-arrival event; a child that speaks up and keeps working (the system prompt tells it to) is unheard until settle or deadline. The wait is exactly the tool where cancelling is free (the job keeps running, `_peer_interrupt` re-hands the id), so the courtesy design does not need to apply to it. One-line shape of a fix: have both paths call `self._peer_arrival.mark()` (or add a sibling "inbound for the model" event raced in `_await_any_settled`) and widen the `interrupted_by`/note wording so the model is not told "a message arrived from another session" for a wake or a child note. Add a test pinning each at `ONE_HOUR_MS`, mirroring `test_an_hour_long_wait_still_returns_on_peer_message`. |
| R1-2 | MINOR | Prompt cost — the change is billed on every turn; the rule and its WHY are stated three times. | Measured (tiktoken cl100k): `system.md` 3331 → 3485 tokens (+154), agents guide 2123 → 2279 (+156), `wait` tool schema (description + parameters) 248 → 410 (+162). ~+316 tokens/turn in the always-present prefix. The field description (`builtin.py:7166-7176`) and the tool description (`7590-7597`) repeat the same "SIZE THE BUDGET TO THE WORK … training loop" sentence nearly verbatim, and the cache-TTL explanation appears in the schema comment, `system.md` and the guide. Suggested: keep the rule + carve-out in the schema once, keep `system.md` to the rule and the expire → `jobs`/`peek` cue, and let the guide carry the WHY. Not blocking. |
| R1-3 | MINOR | Test coverage — the new suite pins settle/abort/peer at 60 min but not steer-cancel. | `test_wait_budget.py` docstring defers steer to `test_wait_wakes_on_peer_message.py`, which exercises it at 300 000 ms. Same code path (`CancelledError` branch, `builtin.py:7437`), so low risk, but the file's stated purpose is "pinned at a full-hour budget"; add one steer-cancel case at `ONE_HOUR_MS` for symmetry. |
| R1-4 | NIT | Guidance wording — the ML/progress carve-out names the cases but not the mechanism. | `system.md:147-148` / `GUIDE.md:96`: "Use short waits only when you must manage progress along the way (training runs, staged rollouts)". Clear enough as a carve-out; consider appending "— then `jobs op='peek'` between waits" so the short-wait path is paired with the incremental read it is meant to feed, matching the `_peek_job` docstring's 40-minute-training example. |
| R1-5 | NIT | Version | `0.44.36` while `.33/.34/.35/.37` are claimed by open PRs — acceptable per the PR body; merger re-sequences. |

#### Follow-up observations

- **Nothing else bounds a single tool call below 60 min.** Checked: `LoopConfig.deadline` (`_wire_deadline`, `loop.py:2012`) is set by no caller in the tree; the loop's interruptible poll (`STEERING_INTERRUPT_POLL_S = 0.25`, `loop.py:1478-1489`) only *shortens* a tool on steer/abort/fork; `PENDING_REQUEST_TIMEOUT_S = 30` (`owned.py:63`) bounds approval/ask replies, not tool duration; mobile SSE uses a 25 s keepalive frame, not a request deadline; `exec_worker` only bounds the post-interrupt drain (5 s). The TUI has no wait-specific elapsed rendering — spinner + `⋯ running` header.
- **Interrupt paths confirmed by reading**: abort (`signal.wait()` raced in `_await_any_settled`, `builtin.py:7564`), user/`--now` steer (urgent → `_peek_interrupt` cancels → absorbed at 7437-7467 with the still-running payload), fork boundary (same cancel path), mailbox `wake=False` drop (**yes, it wakes**: `session.py:4001` marks after the journal park; `wake=True`-idle and busy-steer paths mark at 3958/3963/3971). Child `wait` on grandchildren inherits the same ceiling and the same wake set; parent `hub op='steer'` to a child is a real steer (`comms.steer`, `comms.py:1309`) and cuts the child's wait. Only the two paths in R1-1 fall outside.
- **Default 300k → 600k**: no caller or test depends on the old default — every `wait_ms` use in `tests/unit/tools/*` passes it explicitly; the only other `300_000` hit is `compaction/thresholds.py:175` (token threshold, unrelated).
- **Tests assert behaviour, not just schema**: `test_an_hour_long_wait_still_returns_on_{settle,abort,peer_message}` pass `wait_ms=3_600_000` to `execute_wait` and assert elapsed < 2–3 s with the right payload; reproduced locally, 6 passed in 1.00 s.
- **Gates (worktree at 6937a4d2, whole tree as CI spells them)**: `python -m flake8 .` rc 0; `black --check .` 803 unchanged; `isort --check .` rc 0; `pyright` 0 errors; `tests/unit/tools tests/unit/harness tests/unit/session -n 4` → 1711 passed in 33.67 s. CI on 6937a4d2: lint, type-check, pip-audit, tui-e2e (both legs) green; `test (3.12)` still pending at review time.

damianvtran:

Remediating author: this session (coder, anthropic/claude-fable-5-1). Head: `510aaa68`.

- **R1-1 (major, scheduled wakes and child→parent notes slept a parked `wait`)** — fixed in `510aaa68`. Both paths now mark the peer-arrival signal: `_deliver_wake`'s busy branch and `_deliver_resume_catchup` mark with kind `wake_prompt`, and `queue_aside` marks with kind `hub_message` (one site covers `reply_to_parent` and parent→child notes, all drained at the same boundary). The protocol grew `arrivals()` (per-kind monotonic tally summing to `count()`), and `wait` diffs it across its park, so the note says WHY it woke: "a scheduled wake fired — read the reminder before re-waiting", "a hub message arrived from a subagent or your parent"; `interrupted_by` names the single kind (the `peer_message` contract is unchanged) and reports the neutral `inbound` plus an `arrivals` breakdown when several land in one park. Courtesy steering is untouched: a wake still never cancels a mutating tool. Tested at `wait_ms=3_600_000` per kind in `test_wait_budget.py`, the producer marks in `test_session_peer.py` (unit), and one end-to-end turn whose in-turn 60-minute `wait` is cut by a fired wake — it fails with a TimeoutError on the unfixed path. Live proof: 30-minute budget on a 120 s job cut by a 25 s wake in 21.5 s, `interrupted_by: "wake_prompt"`, job left running (details in the PR body).
- **R1-2 (minor, sizing rule ×3, +316 tokens/turn)** — fixed in `510aaa68`. The rule now lives ONCE, in the tool description (the schema the model reads at decision time); the field description is one line; `system.md` is one sentence pointing at it; the agents guide carries the why (on-demand, not per turn). Measured with tiktoken cl100k: **+316 → +103 tokens/turn vs main** (`system.md` +31, wait schema +72; guide +249 but never in the prefix).
- **R1-3 (minor, no steer-cancel at 60 min)** — fixed in `510aaa68`: `test_an_hour_long_wait_still_returns_on_steer_cancel` pins the live-signal cancel absorb at `ONE_HOUR_MS` with the still-running payload and the job untouched.
- **R1-4 (nit, carve-out without the mechanism)** — fixed in `510aaa68`: the short-budget carve-out now pairs the short wait with `jobs op='peek'` in the guide, the tool description and `system.md`'s pointer.
- **R1-5 (nit, version)** — accepted as-is: `0.44.36` while `.33/.34/.35/.37` are claimed by open PRs; the merger re-sequences.

Gates on `510aaa68` (whole tree, as CI spells them): flake8 rc 0, black --check 803 unchanged, isort --check rc 0, pyright 0 errors; `tests/unit/tools tests/unit/harness tests/unit/session -n 4` → **1718 passed** in 33.43 s. CI: lint, type-check, pip-audit and both tui-e2e legs green; `test (3.12)` red at 27m on `tests/unit/tui/test_word_caret.py::test_real_teardown_settles_a_held_escape_and_leaks_no_callback` (`KeyError: "No 'text-area--gutter' key in COMPONENT_CLASSES"`), which this branch never touches and which **passes locally on `510aaa68` in 3.84 s** — TUI environment flake; a re-run is requested.

damianvtran:
### Agent review — round 2

- **Reviewer:** reviewer subagent (anthropic/claude-opus-5)
- **Scope:** `47bc768f..510aaa68031d6a804b75be3bb7d144428fc0033d`
- **Mode:** Remediation convergence
- **Delivery contract:** one sized wait replaces repeated 5-minute polls without stranding a session — settle, abort, peer message, scheduled wake, hub note and steer all return immediately
- **Delta reviewed:** `6937a4d2..510aaa68`
- **Applicable risk surfaces:** tool timeouts, session liveness, prompt cost, schema compatibility
- **Verdict:** **Approve**

Round 1's blocker class is closed. Every path that parks something for this session's model now marks the arrival signal, I reproduced both failure modes myself by reverting the fixes, and the atomic snapshot/re-arm window the design rests on is still intact after the per-kind tally was threaded through it. No BLOCKER and no MAJOR remains, so **this round is TERMINAL** — the three minors and one nit below are follow-ups, not a reason for a round 3.

#### Remaining findings

| ID | Sev | File:line | Finding |
|---|---|---|---|
| R2-1 | MINOR | `local_operator/prompts_md/system.md:140` | Condensing the sizing rule also deleted the *other* return causes. The line now reads "it returns the moment work settles, so size ONE `wait_ms`…" — the prose the model reads every turn no longer says a wait also returns on a peer message, a scheduled wake, a hub note or a steer, which is precisely the contract this PR broadens. The tool description still carries it, so the model is not misinformed, only under-informed in the prefix. Fix: "the moment work settles or a message arrives" — three words. |
| R2-2 | MINOR | `local_operator/tools/builtin.py:7302-7308` | The rationale for spelling `_ARRIVAL_NOTES`' keys as literals is **factually wrong**: "``session.session`` and ``harness.comms`` both import this module, so naming the constants here would be a cycle." `harness.comms` does not import `tools.builtin` at all (it imports `session.peer` and `session.transcript`), and I patched `builtin.py` to import `HUB_MESSAGE_TYPE` **and** `PEER_MESSAGE_MESSAGE_TYPE` and imported the package with `harness.comms`, `session.session` and `tools.builtin` each first — all three orders succeed, no cycle. `tests/unit/tools/test_wait_budget.py:37,41,42` already imports all three constants alongside `builtin`, which is the same evidence. The literals are still a *defensible* choice and the pinning test makes them safe; the stated reason is not, and a load-bearing "why" comment that is wrong is how the next editor gets talked out of a correct refactor. Fix: restate the reason as "keep the wording table free of harness imports; the pinning test guards renames", or just import the constants. |
| R2-3 | MINOR | `local_operator/session/session.py:5468` | `queue_aside` marks *before* the thunk is known to materialize, so a thunk that later withdraws itself (`StaleAside`, `harness/comms.py:1913` — a parent question whose asker already timed out) still cuts a parked wait, and the model is told "a hub message arrived from a subagent or your parent" for a message that will never appear. It then re-issues the wait: one wasted full-context round trip, the exact cost this PR exists to remove, on a rare path. The code comment reasons about this explicitly and accepts it as cheaper than teaching the tool about withdrawal, which I agree with — recorded so the trade-off is in the thread rather than only in a comment. `deferred — accepted trade-off, documented at the mark site`. |
| R2-4 | NIT | `local_operator/harness/types.py:393-401` | Adding `arrivals()` to a `@runtime_checkable` Protocol that pydantic validates with `is_instance_of` is a hard compatibility break: I built a two-method `event()`/`count()` implementer and `ToolContext(..., peer_arrival=OldPeer())` now raises `ValidationError`. In-tree this is a non-issue — `_PeerArrival` is the only production implementer and `session.py:4986` the only construction site — but the protocol is public surface in `harness/types.py`, so an out-of-tree host implementing it breaks with no default and no shim. Not worth a shim today; worth knowing it is a break rather than an extension. |

R1-5 (version `0.44.36` overlapping open PRs) stays accepted as-is; the merger re-sequences.

#### Verified

**R1-1 — closed, and I reproduced both halves of it.** The producer set is now complete for every path that can land while a `wait` is parked. `_deliver_wake`'s busy branch (`session.py:8669`) and `_deliver_resume_catchup`'s busy branch (`session.py:7979`) mark `wake_prompt`; `queue_aside` (`session.py:5468`) marks `hub_message` and covers all four `queue_aside` call sites in `comms.py` (attach flush 646, parent ask 1510, child→parent `reply_to_parent` 1768, parent note 1887) at one boundary. The idle branches of both wake deliverers need no mark, and that is load-bearing rather than lucky: `_is_streaming` is set for the whole of `_run_turn` (`session.py:4644`→`4867`), and tool execution is inside it, so a wake that fires while a `wait` is parked **always** takes the busy branch. Steer is covered by the pre-existing `CancelledError` absorb.

Negative checks, both run against a scratch copy of the worktree and reverted after:

- Replaced the `mark(HUB_MESSAGE_TYPE)` in `queue_aside` with `pass` → `test_a_busy_turn_wake_and_a_hub_aside_mark_an_arrival` fails, `assert 1 == 2` on `peer.count()` (`test_session_peer.py:375`).
- Replaced the `mark(WAKE_PROMPT_MESSAGE_TYPE)` in `_deliver_wake` with `pass` → the end-to-end test hangs on its hour budget and dies on the harness `TimeoutError`: `test_a_scheduled_wake_cuts_an_hour_long_wait_inside_a_turn — TimeoutError`, 1 failed in 6.72 s under a 120 s watchdog. This is the claimed check and it holds.

**The atomic window still is what the comment says.** `builtin.py:7389-7398` reads `peer.count()`, then `dict(peer.arrivals())`, then `peer_event.clear()` with **no `await` anywhere between them**, so the sequence is still atomic with respect to the loop and the producer still runs on that loop. `_PeerArrival.arrivals()` (`session.py:288`) is a synchronous `dict()` copy, which is what makes the snapshot survive the park while the producer mutates the live dict underneath it. The post-wake diff (`builtin.py:7507-7511`) only ever subtracts the snapshot from the live tally, so a kind that first appeared after the snapshot is reported in full, and a kind that landed before it is not double-counted. The sum invariant holds by construction: every `mark()` bumps `_count` and exactly one bucket, so the per-kind diff sums to the `count()` diff.

**`interrupted_by` wording is accurate.** One kind → `interrupted_by` is that kind (`peer_message` for peer-only, preserving the round-1 machine contract; `wake_prompt`, `hub_message` otherwise) with a matching single note. Several in one park → the neutral `inbound` with the breakdown in `details["arrivals"]` and both phrases in the text, which is right: no single kind is *the* cause. The empty-`kinds` path falls back to the peer wording and leaves `inbound`, which is unreachable but harmless. No consumer of `interrupted_by` exists outside `builtin.py` and the tests, so the new `inbound` value breaks nothing.

**Courtesy steering is untouched.** `_has_urgent_steering` (`session.py:5434`) is unchanged, and both wake paths still increment `_courtesy_wake_count` before marking, so a queue holding only courtesy wakes still answers `False` and `_peek_interrupt` still cannot cancel a running tool. The mark never touches the steering queue. Structurally, `peer_arrival` is read at exactly one place in the codebase (`builtin.py:7389`) and `wait` is the only tool that parks on the event, so widening the signal cannot preempt a mutating tool by construction — the test's `assert not session._has_urgent_steering()` pins the intent on top of that.

**R1-2 — the +103 figure is right; I re-measured it independently.** tiktoken `cl100k_base`, base worktree at `47bc768f` vs head, `system.md` plus the built `wait` tool schema (`{name, description, parameters}` as JSON):

| | base `47bc768f` | head `510aaa68` | Δ |
|---|---|---|---|
| `system.md` | 3331 | 3362 | **+31** |
| `wait` tool schema | 254 | 326 | **+72** |
| **per-turn total** | | | **+103** |

Down from the +316 of round 1, exactly as claimed. Placement is as described: the full rule appears once in the tool description (`builtin.py:7356-7373`, the schema the model reads at decision time), `system.md:140-144` is one pointing sentence, the field description carries only the numbers, and the long form lives in `guides/agents/GUIDE.md`, which is an on-demand guide (name/description-only in the prefix) so its +249 is never in the per-turn cost. `test_wait_schema_advertises_the_sizing_rule_once` pins the "once" — it asserts `SIZE THE BUDGET TO THE WORK` is in the description and **not** in the field description, which is the assertion that stops the duplication coming back.

**R1-3 — pinned.** `ONE_HOUR_MS = 3_600_000` (`test_wait_budget.py:50`) and `test_an_hour_long_wait_still_returns_on_steer_cancel` cancels with a live non-aborted signal at the full budget, asserting `interrupted_by == "cancelled"`, the still-running payload, and the job row left `running`. The elapsed bound (< 3 s) is what makes it a real regression pin rather than a shape assertion.

**R1-4 — carve-out carries its mechanism** in all three places (guide, tool description, `system.md`'s pointer): the short wait is now paired with `jobs op='peek'`.

**Regression sweep — no spurious wake path found.** Traced every `queue_aside` caller. All four are cross-session: `reply_to_parent` (`comms.py:1768`) marks `self._session`, which is the **parent** (`SubagentComms` is the parent's instance shared with its children, `comms.py:477`), while the caller runs in the child; the other three mark `record.child`. A session cannot mark itself during its own turn: `_execute_hub_child` is gated on `comms.is_child(context.job_id)`, a top-level session has no job id and routes to `_execute_hub_parent`, whose sends target a different session. Even if a self-mark existed it could not cut the session's own wait, because `wait` is `concurrency="exclusive"` — no second tool runs beside it to queue the aside. The attach flush (`comms.py:646`) marks a starting child N times before it has ever parked, which is exactly what the count-not-`is_set()` design absorbs: the child snapshots the current count when it parks, so pre-park marks never wake it. Idle marks are equally inert — nothing is parked, and the left-set event is re-armed by the next park's `clear()`. The one real spurious-wake case is R2-3 above.

**Gates, whole tree, spelled as AGENTS.md requires:**

```
.venv/bin/python -m flake8 .                                   → rc 0
uvx --from black==26.1.0 black --check .                       → 803 files unchanged
uvx isort==5.13.2 --check .                                    → rc 0
  tests/unit/tools tests/unit/harness tests/unit/session -n 4  → 1718 passed in 39.06s
```

**CI on `510aaa68`: fully green, including the leg that was red at round 1.** `test (3.12)` **pass** in 34m7s — the `test_word_caret.py` TUI failure the remediation called an environment flake did not reproduce on the re-run, so that claim is now confirmed rather than asserted. `lint`, `type-check`, `pip-audit`, `cli-sanity`, `server-sanity` and both `tui-e2e` legs pass. Run `33641893281`.

#### Follow-up observations

- The three new test names read as specifications rather than mechanics (`test_a_scheduled_wake_cuts_an_hour_long_wait_inside_a_turn`), and the end-to-end one drives the real loop with a scripted stream instead of asserting on the mark alone — which is why the `_deliver_wake` revert produced a genuine hang rather than a green suite. That is the right shape for a liveness fix and worth keeping as the pattern for the next producer added to this signal.
- `PeerArrivalProtocol` is now the session's single "something to read" signal with three producers. Anything added later that parks a message for the model at an injection boundary must mark it or it reintroduces R1-1 at the new hour-long budget. The class docstring says so; the cheap insurance would be a test that walks every `_steering_queue.put_nowait` / `queue_aside` site and asserts each is accompanied by a mark, so the next producer fails the suite instead of failing silently in production. Not blocking.

damianvtran:

PR #542, opened 2026-09-02
fix(evaluation): frame-id contract and corrective re-prompt; bytecode-proof workspace digest (proof episode fixes)

damianvtran:
## Summary

Two defects found by the **first real paid OSWorld episode** — run root `~/worktrees/osworld/runs/proof-20260902-081937`, bundle `evidence/ep-6ea01a117eee` — fixed at the source, each proven by mutation. Patch bump → 0.44.32.

**No AWS call and no spend in this PR.** Every test here uses the in-process fake adapter or a real spawned worker with the fake provider.

**The AWS side of that episode WORKED.** The instance launched, the TTL schedule was created and later deleted, the instance was terminated, and the post-episode tag audit printed `[]`. The bundle's `cleanup` event carries three receipts (close-session, release-instance, revoke-ttl-lease) and its `rescue.json` is still in the inbox only because of defect 2 below — the rescue re-check refused, not the teardown.

## Reproduction (from the bundle, read-only)

`outcome.json`:
```
"status": "failed", "reportability_label": "cleanup_incomplete", "rescue_required": true, "rescue_complete": false,
"score": {"status": "unscored", "reason": "infrastructure_failure"},
"diagnostic": "_ProviderFailure: DecisionParseError: decision does not match this observation: action references unknown frame_id '1'"
```
`events.jsonl`: `observation` (sequence 0, one 1920x1080 PNG frame, instruction text) → `error {category: "provider", diagnostic_code: "providerfailure", retryable: false}` → finalization → cleanup. **Zero `model_request` events**: the one billed call was never recorded, and the episode ended on its first decision.

The rescue worker's refusal: `adapter workspace content digest differs`, because `~/worktrees/osworld/workspaces/0.1.0/workspace/tasks/__pycache__/task_001.cpython-312.pyc` now exists — written by upstream `instantiate_task` importing the task module from the pinned workspace during `reset_start`. That file is still there.

## Defect 1 — the model was never told which frame ids exist (runner, general)

The adapter publishes its frame as `frame_id="screen"`; the model answered `frame_id "1"`; `ActionBatch.validate_for` rejected it; the runner filed the `DecisionParseError` as `_ProviderFailure`. Nothing in the prompt or the rendered observation listed the ids.

- **Rendered observation** now carries `Frames: screen (1920x1080)` (or `Frames: none`) — every frame's id and model-visible geometry, in order. Any adapter naming works; nothing special-cases `screen`.
- **System prompt** states that `frame_id` MUST be one of the ids on the current observation's `Frames:` line and what a rejection message looks like.
- **A rejected reply is a MODEL error and is re-prompted.** `DecisionRejected` (new, `runner/model.py`) is raised by the client instead of a decision when a billed reply fails strict parsing; it carries the call's full billing provenance. The client has already appended the bad reply and a correction (naming the observation id and the frame ids) to its append-only history, so the runner's re-call for the same observation is corrective by construction. `EpisodeRunner._decide` retries up to `EpisodeConfig.max_decision_retries` (default 2; `0` = old one-strike). **Every attempt writes its own `model_request`/`model_response`/`usage_cost` triple** and counts as a model cycle; a rejected attempt also writes a retryable `error` (`category="model"`, `diagnostic_code="decision-rejected"`, the shown diagnostic as its detail artifact).
- **Exhausting the bound** seals as `_ModelFailure`: `category="model"`, `reason="model_failure"`, `failure_kind="model"` — new literals on `ErrorPayload`, `UnscoredReason` and the lifecycle payloads, distinct from `provider` (nothing was down) and `crash` (nothing broke).
- **Shed regression this exposed** (`compaction/pruning.py`): `shed_stale_frames` removed observation/reply *pairs* and stopped when the tail's front was not an observation. A corrected turn is four messages and the compaction cut can land on a reply or a correction, so the tail could start with an orphaned turn end; the shed then removed nothing for the rest of the episode and the client refused as unrecoverable with sheddable turns still present (the existing 48-turn drive failed at turn 24 once the `Frames:` line nudged the cut). A turn is now its observation plus everything up to the next observation; an orphaned front is shed first; the current observation is still never removed.

## Defect 2 — bytecode cache invalidated the workspace pin (boundary)

- Worker (`discovery.worker_argv`) and supervisor leader (`AdapterSupervisor.launch`) are spawned with **`-B`**; `worker.main` sets `sys.dont_write_bytecode` as the in-process guarantee. The copied-interpreter probe mirrors the flag set.
- `workspace_digest` **excludes `__pycache__/` and every `BYTECODE_SUFFIXES` file by rule** (`_is_bytecode_cache`). Bytecode is never verified content — the finder already refuses to load any `.pyc`, RECORD-covered or not — so a cache from any tool can never invalidate a pin. The cache directory is still inspected: a symlink inside one is refused like any other.
- `release_digest` / `package_digest` have no such exposure: one JSON file, and RECORD rows respectively (a `.pyc` would have to be an explicit row the wheel declared).
- README: the sweep section now says why the sweep takes no `--region` (each descriptor carries the episode's `AWS_REGION`; the audit has no descriptor) and what the digest re-check does and does not count. PROOF_PLAN and README agree.

**Effect on the stranded descriptor — and a second drift source found while checking.** Re-hashing the real workspace with this branch gives `b6b4f0f4…`, NOT the pin (`89a78cdf…`). The `__pycache__` is now ignored, but the same episode also wrote `workspace/cache/001/{…_calendar.ics, …_thunderbird-profile.tar.gz}` (mtime 08:21:38): upstream `DesktopEnv(cache_dir="cache")` is a **relative** path resolved against the worker's cwd, which is the pinned workspace, and the setup controller downloads task assets there. Verified by copying the workspace under a durable path and removing only `cache/`: the digest is `89a78cdf…` again, with or without the `.pyc`. Those files are real content, not bytecode, and the digest must NOT learn to ignore arbitrary directories — the fix is adapter-side: pass an absolute `cache_dir` outside the workspace to `DesktopEnv` (e.g. under the artifact root). **Reviewer's caveat (round 1), accepted:** upstream `desktop_env.py:367` calls `setup_controller.reset_cache_dir(self.cache_dir)`, which DELETES the cache directory between resets, so the absolute path must be **per-episode**, not a shared global cache, unless cross-episode asset reuse is explicitly wanted and the deletion is understood. That means a new adapter wheel and workspace rebuild. **Not in this PR** (the slice is the runner and the boundary). For the stranded descriptor the README's new sweep text already gives the path: restore the workspace from the verified inputs root (delete `cache/` and `tasks/__pycache__`), then sweep. Not run here (no AWS calls).

## Tests and mutation proof

All gates run over the whole tree as CI does: `flake8` clean, `black==26.1.0` 799 unchanged, `isort==5.13.2` clean, `pyright` 0 errors (adapter wheel not installed in the venv).

Suites: `tests/unit/evaluation/runner + evidence + protocol + lifecycle + compaction` → **638 passed**; `tests/unit/evaluation/adapters` → **235 passed ×3 in parallel** (35 s each) and **235 passed serial** (`-n0`, 86 s); the rest of `tests/unit` (minus tui) → 5541 passed.

| Mutation | Failing tests |
|---|---|
| drop the `Frames:` line from the renderer | `test_rendered_observation_names_its_frame_ids_and_geometry`, `test_frameless_observation_renders_no_frames_rather_than_nothing` |
| don't fold the rejection into the context | `test_unknown_frame_id_is_fed_back_and_the_corrected_reply_proceeds`, `test_a_corrected_turn_closes_after_the_correction_in_the_next_request`, `test_a_runaway_rejected_reply_is_replayed_truncated` |
| runner never retries (`rejections >= 1`) | `test_one_rejected_decision_is_re_prompted_and_the_corrected_batch_proceeds` (`'failed' == 'completed'`) |
| file `_ModelFailure` as provider | `test_exhausted_decision_retries_seal_as_a_model_failure`, `test_zero_decision_retries_restores_one_strike` |
| old `shed_stale_frames` (pairs) | `test_shed_stale_frames_sheds_a_whole_turn_and_an_orphaned_turn_end`, `test_shed_actually_sheds_and_a_mid_size_window_runs_to_the_end` |
| remove `-B` from both argv AND `dont_write_bytecode` | `test_worker_importing_from_the_workspace_leaves_no_bytecode_behind` (real spawn: `tasks/__pycache__` appeared) |
| remove only one of the two | still passes — the belt-and-braces point |
| digest walk includes bytecode | `test_workspace_digest_ignores_bytecode_caches_by_rule` |

New runner test `test_one_rejected_decision_is_re_prompted…` asserts the journal order `model_request, model_response, usage_cost, error, model_request, model_response, usage_cost` and that the model saw the identical history both times; `test_exhausted_decision_retries…` asserts exactly `max_decision_retries + 1` calls and four `model` errors (three retryable, one terminal).

`uv build` → `local_operator-0.44.32-py3-none-any.whl` + sdist; `twine check` PASSED on both.

## Not addressed (deliberately)

- **Adapter writes OSWorld's asset cache into the pinned workspace** (`DesktopEnv(cache_dir="cache")`, relative to cwd). This is the second, non-bytecode drift source above; the fix is an adapter wheel change (per-episode absolute `cache_dir` outside the workspace — upstream `reset_cache_dir` deletes it between resets) and a workspace rebuild. Without it the next paid episode's rescue re-check will drift again on any task with downloadable assets.
- The stranded `rescue.json` for `ep-6ea01a117eee` is not swept here (that is an AWS call); the workspace must first be restored (drop `cache/` and `tasks/__pycache__`).
- `max_decision_retries` is not exposed on `scripts/run_episode.py`'s CLI; the default (2) applies to the next paid run.

### DIFF
--- a/benchmarks/osworld_v2_adapter/README.md
+++ b/benchmarks/osworld_v2_adapter/README.md
@@ -243,6 +243,23 @@ It spawns the exact pinned worker per descriptor, re-resolves the descriptor's
 secret refs from the credential store, reconciles every action, and unlinks
 the descriptor **only** when the aggregate is complete.
 
+The sweep takes no `--region`, unlike the audit below: each descriptor
+already carries the episode's `AWS_REGION` in its `infra_values`, and the
+worker reconciles in that region. Passing one would only invite a mismatch
+between where the operator thinks the instance is and where the descriptor
+says it is.
+
+The sweep re-hashes the pinned workspace before it will spawn the worker, so
+the workspace must still match the selector's `workspace_digest`. Bytecode
+caches never count: `__pycache__/` and `*.pyc` are excluded from the digest
+by rule, and the worker is launched with `-B` so it writes none — upstream's
+`instantiate_task` imports the task module from the workspace, and in the
+first paid episode the cache that import left behind made the sweep refuse
+("adapter workspace content digest differs") for an instance that was in
+fact already terminated. Anything else that changes under the workspace is a
+genuine drift and the sweep will (rightly) refuse; restore the workspace from
+the verified inputs root and sweep again.
+
 ## Leak audit (operator command)
 
 Every instance and volume this adapter creates carries `lop:adapter=osworld-v2`
--- a/local_operator/compaction/pruning.py
+++ b/local_operator/compaction/pruning.py
@@ -491,9 +491,21 @@ def shed_stale_frames(messages: Sequence[Message], *, limit: int) -> tuple[list[
     TURNS, not on the presence of an image, because by the time a caller
     sheds, the oldest turns have already been pruned to notices and are the
     stalest thing in the tail. "Oldest first" means those, never the recent
-    frames a screen-driving surface actually acts on. A removed observation's
-    assistant reply goes with it: a reply that answers an observation no
-    longer visible is a decision made about a screen the model cannot see.
+    frames a screen-driving surface actually acts on. A turn is its
+    observation plus EVERYTHING up to the next observation: the reply, and
+    any rejected-reply / correction pair that preceded the accepted reply. All
+    of it goes with the observation, because a reply that answers an
+    observation no longer visible is a decision made about a screen the model
+    cannot see.
+
+    The same rule covers a tail whose FRONT is not an observation. The
+    compaction cut point may legally land on an assistant reply or on a
+    correction message, leaving the end of a turn whose observation is
+    already behind the marker; those orphans are shed first. Stopping on them
+    instead (the earlier behaviour) stalled the shed for the rest of the
+    episode: every pass found a non-observation at the front, removed
+    nothing, and the client refused the request as unrecoverable with a dozen
+    sheddable turns still in the prefix.
 
     Never removes the current observation (the last message — without it the
     request has no question), and stops short of a compaction marker in the
@@ -509,16 +521,15 @@ def shed_stale_frames(messages: Sequence[Message], *, limit: int) -> tuple[list[
             head = index + 1
     tail = list(messages[head:])
     while tail and count_stale_observations(tail) > limit:
-        observation, assistant = tail[0], (tail[1] if len(tail) > 1 else None)
-        if not _is_stale_observation(observation):
+        # Everything before the NEXT stale observation is the oldest turn (or
+        # the orphaned end of one); never the last message, which is the
+        # current observation and the request's only question.
+        end = 1
+        while end < len(tail) and not _is_stale_observation(tail[end]):
+            end += 1
+        if end >= len(tail):
             break
-        # Keeping the assistant while shedding its observation would leave a
-        # decision the model can no longer see the screen for; shed the pair
-        # but never the last message (the current observation).
-        removable = 2 if (assistant is not None and assistant.role == "assistant") else 1
-        if removable >= len(tail):
-            break
-        del tail[:removable]
+        del tail[:end]
     if head == 0 and len(tail) == len(messages):
         return list(messages), 0
     return [*messages[:head], *tail], len(messages) - head - len(tail)
--- a/local_operator/evaluation/adapters/discovery.py
+++ b/local_operator/evaluation/adapters/discovery.py
@@ -193,10 +193,14 @@ def workspace_digest(path: str) -> str:
                 raise AdapterDiscoveryError("adapter workspace contains a symlink")
             if stat.S_ISDIR(info.st_mode):
                 continue
-            if _is_bytecode_cache(candidate.relative_to(root)):
-                continue
+            # The unsafe-file check precedes the bytecode exclusion: the cache
+            # rule forgives a bytecode FILE, not any entry named like one, so a
+            # fifo or hardlink named ``x.pyc`` under ``__pycache__`` is refused
+            # like any other.
             if not stat.S_ISREG(info.st_mode) or info.st_nlink != 1:
                 raise AdapterDiscoveryError("adapter workspace contains an unsafe file")
+            if _is_bytecode_cache(candidate.relative_to(root)):
+                continue
             if len(entries) >= MAX_WORKSPACE_FILES:
                 raise AdapterDiscoveryError("adapter workspace file count exceeds limit")
             total_bytes += info.st_size
--- a/local_operator/evaluation/adapters/discovery.py
+++ b/local_operator/evaluation/adapters/discovery.py
@@ -31,6 +31,20 @@ MAX_WORKSPACE_FILES = 100_000
 MAX_WORKSPACE_BYTES = 4 * 1024 * 1024 * 1024
 RELEASE_MANIFEST = "adapter-release.json"
 
+#: Bytecode caches are never verified content. The finder refuses to load any
+#: ``.pyc`` (RECORD-covered or not), the worker is spawned with ``-B`` so it
+#: writes none, and the workspace digest skips them by RULE rather than by
+#: hoping nothing writes one: in the first paid episode, upstream OSWorld's
+#: ``instantiate_task`` imported ``tasks/task_001.py`` inside the pinned
+#: workspace and CPython dropped ``tasks/__pycache__/task_001.cpython-312.pyc``
+#: there. The rescue worker then re-hashed the workspace, found the cache,
+#: refused with "adapter workspace content digest differs", and the sweep
+#: exited 1 forever for an instance that was in fact already terminated. A
+#: cache file that no loader here will ever execute cannot be allowed to
+#: invalidate a pin.
+BYTECODE_CACHE_DIRECTORY = "__pycache__"
+BYTECODE_SUFFIXES: tuple[str, ...] = tuple(importlib.machinery.BYTECODE_SUFFIXES)
+
 
 class AdapterDiscoveryError(RuntimeError):
     """A closed discovery failure safe to report across the RPC boundary."""
@@ -154,6 +168,12 @@ def workspace_digest(path: str) -> str:
     handshake, but it is not re-verified immediately before spawn the way the
     executable identity is. That asymmetry is a known, deliberate gap for a
     later round to close or accept explicitly.
+
+    ``__pycache__`` directories and bytecode files are excluded by rule (see
+    ``BYTECODE_CACHE_DIRECTORY``): they are a side effect of importing, not
+    content, and no loader in this subsystem will ever execute one. A symlink
+    or an unsafe file INSIDE a cache directory is still fatal -- the exclusion
+    is of bytecode from the manifest, not of the directory from inspection.
     """
 
     root = Path(path)
@@ -172,6 +192,8 @@ def workspace_digest(path: str) -> str:
             if stat.S_ISLNK(info.st_mode):
                 raise AdapterDiscoveryError("adapter workspace contains a symlink")
             if stat.S_ISDIR(info.st_mode):
+                continue
+            if _is_bytecode_cache(candidate.relative_to(root)):
                 continue
             if not stat.S_ISREG(info.st_mode) or info.st_nlink != 1:
                 raise AdapterDiscoveryError("adapter workspace contains an unsafe file")
@@ -194,6 +216,19 @@ def workspace_digest(path: str) -> str:
             )
     entries.sort(key=lambda entry: entry["path"])
     return canonical_digest("adapter-workspace-manifest-v1", entries)
+
+
+def _is_bytecode_cache(relative: Path) -> bool:
+    """Whether a workspace-relative path is a bytecode artifact to skip.
+
+    Both spellings are matched: the standard ``__pycache__/<stem>.<tag>.pyc``
+    and a legacy ``<stem>.pyc`` beside its source, because CPython writes the
+    former and ``compileall -b`` writes the latter, and neither is content.
+    """
+
+    if BYTECODE_CACHE_DIRECTORY in relative.parts[:-1]:
+        return True
+    return any(relative.name.endswith(suffix) for suffix in BYTECODE_SUFFIXES)
 
 
 def verify_release_manifest(selector: AdapterSelector) -> None:
@@ -224,12 +259,17 @@ def worker_argv(selector: AdapterSelector) -> tuple[str, ...]:
     if workspace_digest(selector.workspace) != selector.workspace_digest:
         raise AdapterDiscoveryError("adapter workspace content digest differs")
     # Isolation flags prevent user site, PYTHON* variables, and current-directory
-    # imports from changing which exact wheel the worker verifies.
+    # imports from changing which exact wheel the worker verifies. ``-B`` keeps
+    # the worker from writing bytecode: its cwd IS the pinned workspace, and an
+    # adapter that imports a task module from it (upstream OSWorld does) would
+    # otherwise leave a ``__pycache__`` behind that the digest must then ignore
+    # by rule -- see ``BYTECODE_CACHE_DIRECTORY``. Belt and braces.
     return (
         selector.python_executable,
         "-I",
         "-s",
         "-E",
+        "-B",
         "-m",
         "local_operator.evaluation.adapters.worker",
     )
--- a/local_operator/evaluation/adapters/supervisor.py
+++ b/local_operator/evaluation/adapters/supervisor.py
@@ -257,11 +257,14 @@ class AdapterSupervisor:
         }
         env = minimal_environment(environment or os.environ, protocol_fds=protocol_fds)
         env[LAUNCH_IDENTITY_ENV] = json.dumps(resolved.__dict__, sort_keys=True)
+        # Same flags the worker gets (``discovery.worker_argv``): the leader
+        # runs in the workspace too and must leave no bytecode there either.
         argv = (
             selector.python_executable,
             "-I",
             "-s",
             "-E",
+            "-B",
             "-m",
             "local_operator.evaluation.adapters.supervisor",
             "--supervise",
--- a/local_operator/evaluation/adapters/worker.py
+++ b/local_operator/evaluation/adapters/worker.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import asyncio
 import os
+import sys
 from collections import OrderedDict
 from dataclasses import dataclass
 from typing import Any, cast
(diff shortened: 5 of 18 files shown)


damianvtran:
### Agent review — round 1

Reviewer: reviewer (Local Operator subagent, anthropic/claude-fable-5-1)
Scope: f82d5527..65207c2f (base = merge-base with origin/main; main had not moved)

Reviewed from a detached worktree at 65207c2f. All execution below is on the PR head; the proof bundle at `~/worktrees/osworld/runs/proof-20260902-081937/evidence/ep-6ea01a117eee` was read-only.

#### Reproduction summary

**Defect 1 (frame-id contract + corrective re-prompt):**
- (i) PR's own `test_one_rejected_decision_is_re_prompted…` asserts exactly the journal order `model_request, model_response, usage_cost, error, model_request, model_response, usage_cost`, a retryable `model`/`decision-rejected` error, identical history on both calls, and `verify_bundle().valid`. Passed locally.
- (ii) `test_exhausted_decision_retries_seal_as_a_model_failure` — exactly `max_decision_retries + 1` calls, four `model` errors (three retryable, one terminal `modelfailure`), `reason="model_failure"`, cleanup still ran. Passed. The terminal error lands in `_finalize_failure`, whose `isinstance` ladder names `_ProviderFailure`/`_ModelFailure` before the generic crash branch — no misclassification of the route possible; comparability is unaffected by the `served_route` fall-back because both sides of the fall-back resolve to the pinned route for `ProviderModelClient`.
- (iii) Billing honesty by mutation (I ran the mutations): dropping `self._provider_cost_micros += decision.cost_micros` for rejected replies fails `test_one_rejected_decision_is_re_prompted…` via the verifier (`receipt_binding_invalid`); skipping the `usage_cost` event for rejected replies fails via `EvidenceBundleInvalid` (unclosed operation). The counters are a pure sum of what was spent and the verifier enforces it.
- (iv) Derived prompt verified live: changed an observation's frame id to `desktop-main` and geometry to 1920x1080; the rendered observation shows `Frames: desktop-main (1920x1080)` and the correction prompt restates it. The system prompt states the contract generically — no `screen` literal baked in.
- (v) My own drive: 48-turn episode, 5,000/frame priced (above the client's prediction), 10 injected rejections (every 5th observation) → 58 requests, peak 56,883 ≤ 64,000 window, 0 over-window; every non-rebuild request is identity-preserved appends over the previous one; all 10 corrections appear as the final user message of the retry request with the bad assistant reply (`frame_id "1"`) immediately before. The 4-message corrected turn survives rebuilds.
- (vi) Consumer audit: the only exhaustive `failure_kind` matches (`evidence/store.py:817`, `verify.py:521`, `lifecycle.py:543`) key on `preflight`/`infrastructure` by design — an execution-phase `model` failure correctly does NOT take the early-failure path, and `lifecycle.py:543` legitimately allows it to carry its unscored score. Nothing else matches on `category`. Verified by running the PR's model-failure tests through `verify_bundle`.
- (vii) Ordinary-session invariance: `tests/unit/tui` + `tests/unit/compaction` → 4559 passed, 12 skipped (TERM=xterm-256color, NO_COLOR unset) — the shared shed change causes no session regression.

**Defect 2 (bytecode-proof digest):**
- (i) Planted `tasks/__pycache__/task_001.cpython-312.pyc`, `tasks/task_001.pyc`, `x.cpython-312.pyc` → digest unchanged; planted a real file → digest changed; restored → equal. (ii) The PR's real-spawn test (`test_worker_importing_from_the_workspace_leaves_no_bytecode_behind`) passed in the adapters suites. (iii) `-B` is on both argv sites (`discovery.worker_argv`, `supervisor.launch`) and `worker.main` sets `sys.dont_write_bytecode`; `tests/unit/evaluation/copied_interpreter.py` mirrors all four flags. (iv) Symlinks inside a cache dir are still refused (checked before the bytecode rule, including a symlink *named* `.pyc`); a hardlink inside a cache dir is refused.

**Defect 1 bundle cross-check:** the episode bundle carries the observation's frame(s) as artifacts with a 1.67 MB PNG, then exactly one `error {category: "provider", diagnostic_code: "providerfailure", retryable: false}` and `model_request_count: 0` — matching the PR's account of the billed-but-unrecorded call.

#### Findings

**MINOR-1** — `local_operator/evaluation/adapters/discovery.py:196-199`: the "cache dir is still inspected" guarantee has a suffix-shaped hole. Symlinks and hardlinks inside a `__pycache__` dir are refused, but only because the symlink check precedes `_is_bytecode_cache` and the hardlink check uses lstat; any other non-regular file (fifo, socket, device) whose name ends in `.pyc` is silently ignored (reproduced: `mkfifo __pycache__/f.pyc` → digest unchanged, no error). Low real-world exposure (the finder never executes workspace bytecode), but the docstring overstates: the fix is one line — move the `_is_bytecode_cache` check after the `stat.S_ISREG` check, or check "unsafe" on every non-dir path and only exclude regular files from the manifest.

**NIT-1** — `tests/unit/evaluation/runner/test_episode_subprocess.py`: `test_worker_importing_from_the_workspace_leaves_no_bytecode_behind` relies on `tasks/task_001.py` being imported by the armed fake adapter; an even stronger assertion would also import a nested module (e.g. `tasks/pkg/mod.py`) to prove `-B` covers subdirectory caches. Not worth churn; the real episode's exact shape is covered.

#### On the disclosed NOT-FIXED item (adapter cache_dir)

Confirmed honest: the wheel's `providers/aws.py` constructs `DesktopEnv(...)` with no `cache_dir`, and upstream `desktop_env.py:41` defaults it to relative `"cache"`, resolved against the worker's cwd (the pinned workspace). The real workspace still contains `cache/001/{…_calendar.ics, …_thunderbird-profile.tar.gz}` and `tasks/__pycache__/task_001.cpython-312.pyc` — the digest correctly does NOT ignore the former (my B(i) probe: real files change the digest). Agree the digest must not learn to ignore `cache/`; the adapter-side fix is to pass `cache_dir=<absolute path outside the workspace>` (e.g. under the artifact root) in `providers/aws.py:495`. One thing the follow-up must account for: `desktop_env.py:367` calls `setup_controller.reset_cache_dir(self.cache_dir)`, which may DELETE the directory between resets — so the chosen absolute path should be per-episode (or the adapter must verify reset semantics), not a shared global cache, unless cross-episode asset reuse is explicitly wanted and the deletion behavior is understood.

Blockers: 0. Majors: 0. Minors: 1. Nits: 1.

Verdict: **clean** — this round is TERMINAL. The single MINOR is a follow-up hardening of the digest's inspection claim, not a reason to re-round. Gates re-run on this head: flake8 clean, black 799 unchanged, isort clean, pyright 0 errors (completed normally, not starved); runner/evidence/compaction 522 passed; adapters 235 passed ×3 parallel and ×1 serial; tui+compaction 4559 passed. CI: all 8 checks pass (lint, type-check, test 3.12, tui-e2e ×2, pip-audit, cli/server-sanity).

damianvtran:
### Agent review — round 2

Reviewer: reviewer (Local Operator subagent, anthropic/claude-opus-5)
Scope: 65207c2f..a6c35960 (freshness round; merge-base with `origin/main` at review time = 47bc768f)

Reviewed from a clean detached worktree at `a6c35960b6e34efca6cffbde7b76504d5b40ceca`. No edits, no merge, no release. This round is proportionate to what changed since round 1: a one-line check reorder, a test strengthening, and a rebase.

#### 1. Scoped diff — the content delta is exactly what was claimed

`git diff 65207c2f a6c35960 -- local_operator tests benchmarks scripts` reports the branch's own content change as **exactly three files**, and nothing else:

- `local_operator/evaluation/adapters/discovery.py` — the `_is_bytecode_cache` check moved *after* the `S_ISREG`/`st_nlink` unsafe check, with a comment stating the rule ("the cache rule forgives a bytecode FILE, not any entry named like one").
- `tests/unit/evaluation/adapters/test_discovery.py` — the fifo case (+6 lines).
- `tests/unit/evaluation/runner/test_episode_subprocess.py` — the nested-module import and the `tasks/pkg/mod.py` fixture (+9 lines).

Everything else in the raw two-dot diff (`model/catalogue.py`, `model/prices.py`, `model/ids.py`, `model/discovery.py`, `model/configure.py`, `tui/app.py`, `model_picker.py`, `docs/design/model-catalogue-freshness.md` and their tests) is **main's own #527 arriving via the rebase**, confirmed by diffing the branch-own range instead: `47bc768f..a6c35960` touches the same 18 files as round 1's `f82d5527..65207c2f`, with only the three deltas above. Exactly one `chore(release)` commit at the tip (`a6c35960`), and the duplicate 0.44.32 bump was indeed dropped as already-upstream. No stray file, no smuggled change.

#### 2. The MINOR fix bites (mutation-proven)

As shipped: `tests/unit/evaluation/adapters/test_discovery.py -k bytecode` → **7 passed**. Planting `mkfifo __pycache__/task_004.cpython-312.pyc` is refused with `AdapterDiscoveryError("adapter workspace contains an unsafe file")`.

Reverted the reorder in place (bytecode check back *before* the `S_ISREG` check) → `test_workspace_digest_ignores_bytecode_caches_by_rule` fails with exactly the predicted `Failed: DID NOT RAISE AdapterDiscoveryError` at `test_discovery.py:1015`. Restored; tree clean. The round-1 MINOR is genuinely closed, not just asserted.

The surrounding guarantees still hold in the same test: bytecode caches remain excluded from the digest **by rule** (planted `.pyc` files leave the digest unchanged), real content outside the cache still moves it, and a symlink inside the cache is still refused.

#### 3. The NIT fix bites (mutation-proven)

As shipped: `test_worker_importing_from_the_workspace_leaves_no_bytecode_behind` → **1 passed** (real spawned worker, fake provider).

Mutated the no-`-B` case — removed `"-B"` from `discovery.worker_argv` and flipped `worker.main`'s `sys.dont_write_bytecode` to `False` → the test fails listing **both** `workspace/tasks/__pycache__` *and* `workspace/tasks/pkg/__pycache__`. The nested directory is what round 1 asked for and it is now load-bearing: the assertion would fail on a `-B` regression that only escaped into a subpackage. Restored; tree clean.

#### 4. Round-1 held list, re-checked after the rebase

- **(a) Unknown frame id → one corrective re-prompt, then proceed.** Green. The test asserts the journal order `model_request, model_response, usage_cost, error, model_request, model_response, usage_cost` — two full triples for the one step — plus `model_request_count == 3`, `model.calls == 3`, identical history on both calls (`histories[0] == histories[1]`, so the re-prompt is for the same observation), a single retryable `("model", "decision-rejected", True)` error carrying its detail artifact, and `verify_bundle().valid`.
- **(b) Retry exhaustion seals `model_failure`, not `provider`.** Green (`test_exhausted_decision_retries_seal_as_a_model_failure`, `max_decision_retries=2`, and the `max_decision_retries=0` one-strike case).
- **(c) The rejected reply's cost is still counted.** Confirmed at the source: in `runner/episode.py` the `usage_cost` event, `_model_cycles`, `_recent_costs`, `_provider_cost_micros += decision.cost_micros` and the `_usage_totals` accumulation all run on the shared path **before** the `if rejected is not None:` branch that publishes the error and raises `_DecisionRejection`. A rejected attempt cannot skip billing without deleting shared code — which is what round 1's mutations already showed the verifier catches.
- **(d) Ordinary-session invariance.** The shed change is in the shared engine, so it was re-exercised: `tests/unit/compaction` is included in the 522-passed run below, and the TUI + compaction slice was re-run under `env -u NO_COLOR TERM=xterm-256color` and was passing cleanly through its progress when I posted (this machine is heavily contended — several sibling suites in flight — so I am reporting it as *in progress and green so far* rather than claiming a total I did not read). Round 1 recorded the completed figure (4559 passed, 12 skipped) on the identical shed code; nothing in this round's delta touches `compaction/pruning.py`.

#### 5. Gates and suites on this head

- `flake8 .` — clean. `black --check .` (26.1.0) — **802 files unchanged**. `isort --check .` (5.13.2) — clean.
- `pyright --pythonpath .venv/bin/python .` — **0 errors, 0 warnings, 0 informations**, completed normally (not starved; it ran to genuine completion in the background).
- `tests/unit/evaluation/runner tests/unit/evaluation/evidence tests/unit/compaction` — **522 passed**, matching round 1 exactly.
- `tests/unit/evaluation/adapters` — **235 passed ×3 parallel** and **235 passed ×1 serial** (`-n0`). Stable, no ordering dependence.

#### Findings

**BLOCKER-1 — the branch is behind main, conflicts, and its version bump is now stale-low.** `origin/main` has moved past the rebase base to `6ae89bc8` (#537), which bumped `0.44.32` → **`0.44.35`** in a single commit. This branch bumps to **`0.44.33`** — *below* the version already on main. GitHub reports the PR `CONFLICTING` / `mergeStateStatus: DIRTY`, and `git merge-tree --write-tree origin/main a6c35960` confirms the conflict is confined to a **single file, `pyproject.toml`, on the version line only** — every source and test file auto-merges. Merging as-is would either fail or regress the version. Fix: rebase onto `6ae89bc8` and rebump to `0.44.36`.

This is a merge-readiness blocker, not a defect in the change under review — the code, tests and evidence are sound. **A rebase whose only content resolution is the `pyproject.toml` version line preserves this round's verdict and does not require a round 3**; confirm the post-rebase diff against `a6c35960` touches nothing but `pyproject.toml` before merging.

No other findings. Nothing new in the reviewed delta.

Blockers: 1 (merge-readiness: stale base + stale-low version bump). Majors: 0. Minors: 0. Nits: 0.

Round 1's MINOR and NIT are both closed and mutation-proven; no finding is carried forward.

#### CI

6 of 8 checks pass on this head, including **`test (3.12)` — pass (30m43s)**: lint, type-check, test (3.12), tui-e2e (macos-latest), tui-e2e (ubuntu-latest), pip-audit. `cli-sanity` and `server-sanity` were still **pending** at review time — those are the manager's to confirm before merge.

#### Verdict

**Content: clean — TERMINAL.** The round-1 MINOR and NIT are fixed and each proven by mutation; the scoped diff contains nothing but those fixes plus the rebase; every held-list behaviour survives; gates and suites are green. There are no code blockers and no majors, and no further review round is warranted for the change itself.

**Merge recommendation: rebase onto `6ae89bc8` and rebump to `0.44.36`, confirm the two pending sanity checks, then merge.** Do not merge at `a6c35960` — it conflicts and would ship a version below main. If the rebase resolves only the `pyproject.toml` version line, this round stays valid and fresh against the new head.

damianvtran:
Rebased onto \`main\` @ \`6ae89bc8\` (#537 bumped 0.44.32→0.44.35 in one commit, leaving this branch's 0.44.33 *below* main) and rebumped to **0.44.36**. New head \`dc146f8a\`.

Per round 2's stated condition, this preserves that round: \`git diff a6c35960 dc146f8a -- . ':!pyproject.toml'\` is **empty** — the only delta is the version line. Manager-performed.

PR #545, opened 2026-09-02
fix(tui): scroll a long /btw aside, and admit what it cut

bbqben:
# PR Description

## Summary of Changes

A long `/btw` aside answer had text that **no gesture could reach**, and the card did not admit it. Measured through the real app at 80x24: **9 of 200 answer rows reachable**, with no marker saying the rest existed.

Two causes, both in `aside_panel.py`:

1. **The scroll offset was a TURN index.** `_max_scroll_back()` was `len(turn_groups) - 1`, which is `0` when there is one turn — so a single answer taller than the card had a middle the wheel could not address. The rows were rendered, cut, and discarded.
2. **With one turn, `hidden == 0`**, so `_body` returned a bare `lines[-budget:]` slice and emitted **no overflow marker at all**. The reader was not told anything had been cut and read a truncated answer as a complete one. The identical answer *did* announce itself as soon as a second turn existed.

The card is `can_focus = False` by contract (the composer stays live and keeps focus), so the only gesture was the mouse wheel — and the wheel is not delivered under `tmux set -g mouse off`, with mouse reporting disabled, under `screen(1)`, or on non-SGR terminals. In those setups the card's own `↑ … · scroll` hint named content with **zero** way to reach it.

What changed:

- **The scroll unit is now a row.** `_scroll_back` → `_scroll_back_rows`; `_max_scroll_back` is total rows minus budget; a new `_window()` is the single place the painter, the clamp and the keyboard page agree on what is on screen.
- **Single-turn overflow states its cost**: `↑ N earlier lines · scroll`. Lines rather than questions, because the question count is zero there — which is exactly why it used to say nothing. Multi-turn wording and count are unchanged.
- **`ctrl+pageup` / `ctrl+pagedown`** page the card from the keyboard, so it is reachable without a mouse.
- **`ctrl+r`** copies the full exchange to the clipboard. This is the only escape hatch that keeps the feature's off-the-record contract — `^F` fork writes the exchange into the transcript and the session context, and is refused mid-stream, which is precisely when an aside is most used.

What the turn grouping protected is kept. A row cut once "left the top of the card showing a mid-sentence continuation at the answer indent with no question above it", so the window **pins the question that owns its rows**.

## Related Issue(s)

- Related: none filed; reported directly.

## Impact

**No breaking changes, no dependency updates.** Confined to the `/btw` aside card and three new key bindings.

Two behaviour changes a reviewer should look at deliberately:

- **The card can now open mid-answer** with the owning question pinned above the window, rather than always starting on a turn boundary. `test_a_long_exchange_cuts_on_a_turn_boundary…` was renamed to `…pins_the_owning_question…` and asserts the pin. That is a deliberate contract change, not a loosened assertion.
- **`append_answer` now adjusts the scroll offset.** Holding the offset *number* still was not enough: it counts back from the tail, so a reader parked 120 rows back watched their top row walk from row 067 to row 127 across 60 deltas. The rows now hold still, and a reader at the tail still follows — the `TailAnchor` three-state rule, in this card's units.

**Key bindings.** `ctrl+pageup`/`ctrl+pagedown`/`ctrl+r` were audited against app `BINDINGS`, `Editor.BINDINGS`, `Editor._on_key` (which runs before bindings), `TextArea.BINDINGS` and `Screen.BINDINGS`. All three are free and were measured actually reaching the app past a focused `TextArea` holding a draft, with text and caret unchanged. `ctrl+y` was rejected on measurement — it is TextArea's Redo and is swallowed.

**Why not a bare letter**, as the `omp` reference uses (`c copy · b branch to chat · Esc dismiss`): its input controller sees raw keypresses before its editor, so it can guard a bare `c` on "the draft is empty" and pass through otherwise. In Textual the focused `TextArea` consumes the printable character first, so that guard is unreachable — by the time it could run, the `c` is in the buffer. Our own bare `c`/`p`/`r` work only inside the subagent view, which sets the composer read-only. Recorded beside `ASIDE_COPY_KEY`.

**Performance.** `_body()` was already recomputed on every streaming delta; that is unchanged.

## Testing Details

**Verified by exercising the real path** — `/btw` driven through the composer in the pilot, not by poking widget internals.

| Geometry | Budget | Before | After |
|---|---|---|---|
| 80x24, no band | 9 | 9 / 200 | **200 / 200** |
| 80x24, todo band | 3 | 3 / 200 | **200 / 200** |
| 120x40, no band | 25 | 25 / 200 | **200 / 200** |
| 120x40, todo band | 19 | — | **200 / 200** |
| 80x20, band (worst case) | 1 | — | **200 / 200** |

Reachable by the wheel *and* by the chord, each direction swept independently, with the composer's draft and caret unchanged throughout.

`ctrl+r` verified off the **driver**, not the internal buffer: one OSC 52 write, 3023 chars, 200/200 rows, chrome-free, and still works while streaming where `^F` is refused.

**Independent review found three defects, all fixed in this branch and each now guarded by a test:**

- **BLOCKER** — at `budget == 1` the card painted *only* the marker at every offset: 0 of 200 rows, a **regression against main** (main painted one content row). `_fit()` yields budget 1 for `rows_above_dock` 4–7, so a short terminal or a populated dock reaches it. The marker is now suppressed when it would leave no content row, and the pin is clamped so overlay rows can never reach the budget — **content wins the last row**.
- **MAJOR** — `scroll_page(down=True)` skipped rows: the step was sized from the window *before* the move, which self-corrects backward but not forward. At budget 8 rows 8–9 were unreachable going forward. The step is now sized against the *destination* window.
- **MAJOR** — failed and cancelled turns leaked their partial text into the clipboard, in *both* the app-side and panel-side copy paths. The filter tested the answer, not the state. `action_copy_aside` now delegates to `panel.copy_text()` as the single shipping path, filtering on `state not in ("done", "running")` — `forkable` would have been the wrong fix, as it drops running turns and re-breaks mid-stream copy.

**Automated tests.** `tests/unit/tui/test_aside_reachability.py` is new — 950 lines, 23 tests, 66 parametrized cases. It carries a reachability **oracle**: it exhausts every gesture to both clamps and unions every painted row, so a row absent from that set is reachable by nothing. It sweeps `_scroll_by` and the public `scroll_page`, each direction separately, across the geometry matrix, with the measured budgets re-derived by a pilot test so the matrix cannot drift into testing a fiction. **Against the pre-fix tree it reports 9 failures naming all three defects.** Plus new cases in `test_aside.py` (marker, budget 1/2/3, streaming anchor both ways, `scroll_page`, `copy_text` state filter) and `test_app_pilot.py` (chords, copy, `/help` gutter rows).

Mutation-tested rather than assumed: reverting the state filter, filtering on `forkable`, and restoring the inline copy duplicate each fail the specific test that names the real cause.

**Suite:** `tests/unit/tui` **4440 passed, 12 skipped, 0 failed** (was 4359/12/7 pre-fix). flake8, black, isort clean tree-wide; **pyright 0 errors**.

**Not verified:** real-terminal delivery of the chords under tmux/screen/non-SGR — the pilot synthesizes key events. `docs/evidence/aside-chord/` records the probe harness for that measurement; it was cancelled once the reference implementation settled the chord choice, so it records a method rather than a verdict. `tests/e2e` was not run.

## Checklist

## Screenshots (Optional)

Before, at 80x24 — three of two hundred rows, and nothing saying so:

```
Aside  off the record — nothing here joins the chat
────────────────────────────────────────────────────
  ANSWER-ROW-197
  ANSWER-ROW-198
  ANSWER-ROW-199

esc discard, back to the chat · ^f fork into the chat · enter ask again
```

After — the cut is stated, the question is pinned, and the way out is advertised:

```
Aside  off the record — nothing here joins the chat
────────────────────────────────────────────────────
↑ 193 earlier lines · scroll
▌ explain the whole loop
  ANSWER-ROW-193
  ANSWER-ROW-194
  ANSWER-ROW-195
  ANSWER-ROW-196
  ANSWER-ROW-197
  ANSWER-ROW-198
  ANSWER-ROW-199

esc discard, back to the chat · ctrl+r copy · ^f fork into the chat
```

A short screen recording of scrolling a long aside will be attached to this PR.

https://github.com/user-attachments/assets/4c8b9ab1-0d0c-4ffb-804f-2c8c83469af6


damianvtran (approved):

bbqben:
## Blocked on a design collision with #553 — needs your call

Rebasing/merging `main` into this branch after #553 (`fix(tui): render the aside answer as markdown`) landed surfaces a genuine architectural conflict between the two, not a mechanical conflict. Flagging before pushing anything, since the guard in question is yours.

### The collision

`test_a_paint_renders_only_the_turns_the_card_can_show` (added by #553) asserts:

```python
assert len(panel._answer_cache) <= len(painted_questions) + 1
```

That holds because #553's `_visible_rows` walks turns **lazily, newest first, and stops when one does not fit** — so only a card's worth of turns is ever markdown-rendered.

This PR replaces the turn walk with a **row window** (`_flat_body` + `_window`). A row cut cannot know where the window starts without the row counts of the whole exchange, so `_flat_body` renders every turn by construction. With 50 turns the cache holds 50 answers to paint 6:

```
FAILED tests/unit/tui/test_aside.py::test_a_paint_renders_only_the_turns_the_card_can_show
AssertionError: the paint rendered 50 answers to show 6
assert 50 <= (6 + 1)
```

The two goals are directly opposed: **#553 wants no turn rendered for nobody; #545 needs every row counted to make the middle of a long answer reachable.** I did not want to quietly weaken your perf guard to get this green, hence the pause.

### What I measured

The `_answer_cache` from #553 does most of the work here — the expensive path is the *cold* paint only. On a 120x40 card, realistic answers (`STALL_MS = 30.0` in `tests/unit/test_tui_responsiveness.py` for reference):

| turns | cold paint | warm repaint | streaming delta |
|------:|-----------:|-------------:|----------------:|
| 10 | 4.5 ms | 0.09 ms | 0.34 ms |
| 50 | 13.1 ms | 0.24 ms | 0.65 ms |
| 120 | 31.6 ms | 0.55 ms | 0.79 ms |

So the hot path this card actually lives on — streaming deltas and repaints — is **sub-millisecond**, because the cache is keyed per turn and only the changed answer re-renders. The cost lands on the cold paint: **~31.6 ms at 120 turns, marginally over the 30 ms bar.**

The worst realistic case is the one your test docstring calls out (width is in the key, so a resize drag misses every key at once). Measured across a 20-step drag at 120 turns, the worst single frame was **36.0 ms** — and `_sync_overlay_layout` is debounced by `RESIZE_REFIT_DELAY_S = 0.05`, so that is one settling frame, not a per-frame stall.

`tests/unit/test_tui_responsiveness.py` passes in full (19 passed).

### Options

1. **Rewrite the test to bound cold-paint cost rather than cache size** — keeps the intent (no unbounded work per paint) in terms row-windowing can satisfy, and additionally pins that streaming/warm repaints stay off the hot path. Needs your agreement, since it relaxes the `painted + 1` ratio you deliberately pinned.
2. **Make the flatten cheaper** — memoize `_flat_body` per (width, exchange-revision) so a scroll or a streaming delta reuses row counts, and only a width change or new turn recomputes. Keeps both properties; more work, and it is your call whether it belongs in this PR or a follow-up.
3. Something you have in mind that I have not seen.

I lean towards **2** long-term with **1** as the immediate unblock, but this is your guard and your call.

### State of the rest of the merge

Everything else is resolved and green — `324 passed, 1 failed` across `test_aside.py` + `test_app_pilot.py`, the one failure being the above. Resolutions worth knowing about:

- `_max_scroll_back` → kept this PR's row-based version (`len(_flat_body().lines) - budget`); main's `len(_turns) - 1` is the exact defect this PR fixes.
- `_body` → kept #553's paint-scoped cache prune as a wrapper, with this PR's row window moved into `_visible_rows`.
- `_turn_group` → dropped with the turn walk it existed for.
- `_flat_body` → now calls `_answer_rows(turn, width, dim, danger)`; #553 dropped the `fg` parameter, as markdown carries its own styling.
- `_aside_answer_rows` (test helper) → now emits a markdown **list**. As bare lines, markdown folds 200 source lines into one paragraph and reflows ~5 markers per row, which silently broke the one-line-one-row premise the chord-paging assertions rest on (`test_the_aside_chord_pages_rather_than_creeping_by_a_row` failed on reflow, not on scrolling). Same device as the `- line` fixture #553 introduced in `test_aside.py`.
- Version conflict in `pyproject.toml` → took main's `0.44.41`.

Nothing pushed to this branch. Happy to implement whichever option you pick.

bbqben:
## Pushed a fix — the guard is kept, not relaxed

Following up on my note above. I was wrong to frame this as "#553 and #545 are architecturally incompatible" — that overstated it. The row window needs **row counts**, not **rendered rows**, and #553 had already built the mechanism to separate the two. `030371b4` merges `main` and reconciles them.

### What was actually wrong

The regression your test caught was real, and it was not a test problem — deleting the test would have shipped it. `_flat_body` renders every turn, so the cost grew with the exchange on *every paint*:

| turns | cold paint |
|------:|-----------:|
| 10 | 4.5 ms |
| 50 | 11.4 ms |
| 120 | 30.2 ms |
| 300 | 72.8 ms |

Unbounded, against the 30 ms `STALL_MS` bar.

### The fix

`_flat_body` is memoised on `(width, theme epoch, revision)` — the same inputs `_markdown_rows` keys on, plus a revision counter that `_invalidate_flat` bumps on every mutation. The two caches are layers of one thing: the memo skips the assembly, `_answer_cache` skips the markdown render underneath it.

| turns | first paint | repaint | scroll | page | streamed delta |
|------:|------------:|--------:|-------:|-----:|---------------:|
| 10 | 4.5 ms | 0.02 ms | 0.01 ms | 0.02 ms | 0.32 ms |
| 50 | 13.3 ms | 0.01 ms | 0.01 ms | 0.02 ms | 0.48 ms |
| 120 | 31.8 ms | 0.01 ms | 0.01 ms | 0.02 ms | 0.73 ms |
| 300 | 78.8 ms | 0.01 ms | 0.01 ms | 0.02 ms | 1.50 ms |

A streamed delta renders **exactly one** answer at any exchange length. The first paint is real and inherent to a row cut — it is the one thing the turn walk genuinely did better — but it no longer repeats. Worst case is still a resize drag (width is in both keys): 38 ms for a single frame at 120 turns, debounced by `RESIZE_REFIT_DELAY_S`.

Two placement details that are load-bearing:

- In `append_answer`, `_invalidate_flat()` sits **between** the two height measurements. Those are the pre- and post-delta heights, so a stale memo makes their difference zero — which is precisely the reader-drift bug that branch exists to undo.
- The prune in `_body` now runs **only when the paint actually flattened**. On a memo hit nothing calls `_markdown_rows`, so `_used_answers` is empty and pruning against it evicted the whole answer cache — the next real flatten then re-rendered all 50. That one only showed up because the new test caught it.

### Your guard

Replaced by `test_a_repeat_paint_does_not_re_render_the_exchange`, keeping the intent (no work proportional to the exchange on a repeating paint) in terms a row cut can satisfy. Worth knowing: it counts `_answer_rows` **and** `flatten` separately, because `_answer_cache` alone makes the `flatten` count pass **with the memo deleted outright** — counting only that is unfalsifiable. I verified against four mutants, each of which fails it:

1. memo lookup deleted
2. prune made unconditional
3. `append_answer` missing `_invalidate_flat`
4. `settle_answer` missing `_invalidate_flat`

Plus `test_the_flatten_memo_is_dropped_when_the_exchange_changes` for the risk the memo introduces — rows outliving the exchange they were built from.

### One test change worth a look

`_aside_answer_rows` now emits a markdown **list**. As bare lines, markdown folds 200 source lines into one paragraph and reflows ~5 markers per row, breaking the one-line-one-row premise the chord assertions rest on. `test_the_aside_chord_pages_rather_than_creeping_by_a_row` was failing on **reflow, not on paging** — easy to misread as a scrolling bug. Same device as the `- line` fixture you introduced.

**Verified:** 344 passed (`test_aside.py`, `test_app_pilot.py`, `test_tui_responsiveness.py`); flake8 and pyright clean.

The one thing I would still like your eye on is the replacement guard — it is your test, and I changed what it asserts.

bbqben:
## Green on all 8 checks — holding for your re-review

`1a7d2a74` is passing everything, and the branch is conflict-free against `main`.

Not merging on the existing approval, because it was given at 19:23 against the pre-merge code and two things have changed since that you should see first:

1. **The memoisation scheme** (`_flat_body` memo + `_invalidate_flat` on every mutation, and the prune in `_body` now scoped to paints that actually flattened).
2. **Your laziness guard is replaced.** `test_a_paint_renders_only_the_turns_the_card_can_show` → `test_a_repeat_paint_does_not_re_render_the_exchange`. Same intent, expressed as what a row cut can satisfy. This is the piece I most want your eye on — it is your test and I changed what it asserts.

### One more CI-only failure, now fixed

`test_aside_reachability.py` went red on CI while green locally. My fault: I had only run the suites the merge conflicts touched, and that file is *added* by this branch. Both failures were the same markdown-reflow root cause as the chord fixture, and neither was about scrolling:

- `test_the_marker_counts_questions_once_whole_turns_are_above` — the third answer has to be taller than the budget for two whole questions to sit above the window. Reflowed to a fifth of its height it never got there, so the plural marker was unreachable and the card said "earlier lines". The test's own comment had already flagged this as the thing that must not turn on turn sizes.
- `test_asking_again_re_acquires_the_tail` — scrolls back 100 rows into a 200-row answer; at ~40 rows the clamp capped it at 32, hence `assert 32 == 100`.

`_long_answer` now emits a markdown list, same device as `_aside_answer_rows` and your `- line` fixtures.

### Verification

Whole suite this time, not just the touched files: **11161 passed**. The 24 remaining failures are `ProfileNotFound: sandbox` in the OSWorld AWS adapter, which reproduce identically on unmodified `main` — unrelated to this branch. flake8 and pyright clean.

Happy to merge as soon as you have looked at the replacement guard, or to take it a different way if you would rather keep the turn walk and make the row window fit around it.

PR #544, opened 2026-09-02
feat(config): propagate config.yml edits live across running sessions

damianvtran:
# PR Description

## Summary of Changes

**Change class: C1 (standard).** Patch bump `0.44.35 → 0.44.36` — a reliability fix users notice as "settings now apply everywhere"; there is no new surface to name in a release title. (Rebased onto `main` after #537 landed; the merger re-sequences the patch number by merge order.)

A setting written in one place did not reach any other running consumer. Five of the nine `/settings` sections were labelled `NEW_SESSIONS` and the label was **honest**: `compaction.*` was coerced once at build (`session_factory.py`), `SessionStreamFn` held the manager's `values` dict for the session's life, and `subagents.max_running` was read once into `AsyncJobManager`. An operator with three `lop` panes open who changed the compaction threshold in one had to `/new` the other two — and `/failovers` had a whole `stale · /reload` row whose only job was to apologise for it.

This wires **one `ConfigWatcher` per process** (2 s stat poll on the app loop, an opportunistic kqueue watch on the config *directory* where the platform has one, plus a synchronous in-process fast path from `settings_io` writes) and fans per-registry-key changes out to every running `Session`, its subagents, and the TUI.

Implements the architect's design at `/tmp/lop-settings-propagation/DESIGN.md`, §9 scope, in one PR.

**Why polling and not the mobile daemon** (design §2, measured on this machine): `os.stat` of the file is **1 µs** against **2 060 µs** for a full `ConfigManager` construct-and-load, so a tick that has nothing to do costs a stat and a comparison — the parse only happens when the fingerprint moves. A Unix-socket round trip to the daemon costs 49–644 µs, needs a daemon that is explicitly optional, and would still have to detect the change by stat or kqueue itself; it adds a hop and a reconnect state machine and removes nothing. The real cost of polling is the loop *wake* (~70 µs CPU), which at 50 processes on a 2 s cadence is 0.18 % of one core — less than the mobile daemon's existing 2 s registry scan.

Two facts about the write path shaped the implementation:

- **Every write is an atomic `os.replace`** (`ConfigManager._write_config`), so the file's inode changes each time. Hence the fingerprint is `(st_ino, st_size, st_mtime_ns)`, and hence the kqueue watches the **directory** — a watch on the file's inode reports one `KQ_NOTE_DELETE` and then nothing, forever.
[...]

PR #546, opened 2026-09-02
docs(readme): document cross-session communication

damianvtran:
## What this changes

Adds a **`## ↔ Cross-Session Communication`** section to the README, documenting peer messaging between independent `lop` sessions, and tracks the screenshot it needs.

- New README section (88 lines) covering: `lop sessions` as the directory, the two entry points (`lop send` for humans, the in-session `send` tool for agents), the three delivery modes, target resolution and its refusal to guess, the inbound card the human sees, the loopback + same-account trust boundary, and the honest limits.
- Wired into the **table of contents**, the **top nav row** under the logo, and the **✨ Why Local Operator** list.
- `static/tui-peer-message.png` was **untracked**; it is now committed. Without this every image link would 404 on GitHub.
- `pyproject.toml` → **0.44.36** (patch: docs + asset, no behaviour change, per AGENTS.md materiality rules).

## Placement rationale

Between `## 🧰 What the Agent Can Do` and `## ⚙️ Headless & Server Modes` — a peer-level `##` immediately after the subagents/teams material and immediately before Headless.

Both candidate placements from the brief resolve to the same insertion point here, and it is the right one: the subagent/teams section establishes the manager-and-workers tree, so the contrast this section leads with ("subagents are a tree; this is the other shape") lands while that model is still in the reader's head. It also sits before Headless & Server Modes and Phone Access, which is the natural order because those two sections are about reaching sessions from *outside* the machine, whereas this one is strictly *on* the machine.

## Verification

**Every command, flag, and column was checked against the source in this worktree, then executed.** One brief correction: the brief listed `CWD` as a `lop sessions` table column. It is not — `cli.py:1266` prints `STATE PID KIND CONVERSATION MODEL RSS FOOTPRINT UPTIME HB_AGE`, and `cwd` appears only in `--json`. The README documents it accurately.

Real execution, not a reading of the diff:

```
$ lop sessions
live      10828 tui     Update README with cross anthropic/claude-opus-5    170.6M    178.0M       4m      8s
live      50793 tui     Audit custom fields on p anthropic/claude-opus-5    257.7M    289.0M       2h     11s

$ lop send "a" "test"                    # ambiguous substring
15 sessions match; disambiguate with --pid:
  --pid 10828  Update README with cross-session communication  (anthropic/claude-opus-5)
  --pid 50793  Audit custom fields on profiles E2E  (anthropic/claude-opus-5)
  ...
EXIT=1                                   # exits non-zero rather than guessing

$ lop send --pid 10828 "self test"
that target is this session; use the composer to message yourself      (exit 1)

$ lop send --pid 999999 "x"
no session found with pid 999999

$ python -c "from local_operator.mobile.peer_send import validate_peer_body, PEER_MESSAGE_MAX_BYTES; ..."
cap bytes: 262144 = 256 KB
over cap  : message is too large (262145 bytes); cap is 262144 bytes
```

`send --help` confirms every documented flag exists: `--pid`, `--session`, `--now/--steer`, `--wake`, positional `target`/`message` with "omit to read the body from stdin" (`cli.py:1185-1190` reads stdin when the positional is absent). The four delivery detail strings quoted in the README are the literals in `session/session.py:3959-4002`. The `0600`/`0700` claim is `session/runtime/registry.py:39,52`.

### Anchor verification (programmatic)

The GitHub markdown API strips heading anchors, so I validated against **live GitHub's own anchor ids** for the current README, and confirmed my slug model reproduces all of them exactly before trusting it on the new heading:

```
model slugs not found in live GitHub ids: NONE — model matches GitHub exactly
new section slug: '-cross-session-communication'
links pointing at it: 2   (ToC + top nav)
total internal links: 29
BROKEN links in edited README: none
```

The leading hyphen is correct and matches existing entries (`#-quickstart`): the emoji is dropped and its trailing space becomes a hyphen.

### Rendered README

Rendered through the GitHub markdown API (`gh api /markdown -f mode=gfm`), served over loopback, and **visually inspected** in the browser — not assumed. The screenshot renders at the right width, the caption sits centred and italic beneath it, the `lop sessions` fence and the inline `STATE`/`live`/`wedged` code spans are formatted correctly, and the section heading is right.

The alt text and caption were written after looking at the frame: it shows the inbound `↔ peer message from "Audit custom fields on profiles E2E" (pid 50793, anthropic/claude-opus-5)` card claiming the user-dashboard QA and prod deploy slot for MR !1356 and asking for objections, the receiving session's own `send` card replying "No objection — go ahead", then its `wait` / `bash` / `hub peek` receipts.

## Gates

| Gate | Result |
| --- | --- |
| `flake8 .` | **pass** (rc=0) |
| `black --check .` (26.1.0) | **pass** — 802 files unchanged |
| `isort --check .` (5.13.2) | **pass** (rc=0) |
| `pyright` | **pass** — 0 errors, 0 warnings |
| Targeted unit tests | **98 passed** |

Targeted subset covering everything near this diff: `test_send_tool.py`, `test_peer_client.py`, `session/runtime/test_server.py` (45 passed) plus `test_cli_send.py`, `test_cli_style.py`, `tui/test_peer_message.py`, `test_wait_wakes_on_peer_message.py`, `mobile/test_peer_send.py` (53 passed).

The full ~2700-test suite was started and ran green-in-progress for 14m34s under heavy contention from ~10 concurrent sessions on this host; it was cancelled in favour of the targeted subset, since a docs + asset + version diff cannot perturb it. **Full-suite coverage is delegated to CI on this PR.**

## Notes

- Version is **0.44.36**, not the 0.44.32 originally specified: 0.44.32 is already published on PyPI and `main` was at 0.44.35, so that would have been a downgrade. Confirmed with the manager. Branch was cut at `origin/main` tip (`6ae89bc8`), 0 commits behind, so no rebase was required.
- Prose was audited for AI-isms; the one connector em dash in the caption was removed. The two remaining em dashes in the diff are a literal quote from the screenshot and the literal `—` value the `FOOTPRINT` column prints.


damianvtran:
### Copy review — round 1

Reviewer: copy-reviewer (anthropic/claude-opus-5)
Scope: origin/main (6ae89bc8)..af075a15

Read as rendered at head (`README.md` blob at af075a15), judged in the voice of the surrounding document, for the actual reader: a developer skimming an open-source terminal agent on GitHub. Per the operator's calibration, em dashes matching this file's established house voice are **not** flagged; hype words, "not just X", tricolons, filler and throat-clearing are.

The section is well-built and, unusually for docs of this kind, mostly free of AI-isms: no "seamless", no "robust", no "unlock", no hedging intro. Real problems are concentrated in the lead and the pitch bullet, which is exactly where the operator asked me to push hardest.

---

#### C1 — BLOCKER: the differentiating contrast rests on a false claim

Current:
> `Subagents are a tree: one manager, workers it spawned, one shared transcript.`

"One shared transcript" is not true. Each subagent has its own transcript: `hub op='peek'` is documented in-product as "READ the subagent's transcript", `resume` relaunches a child "against its own transcript", and this README's own subagent section three screens up tells the reader to "peek at **its** transcript". A developer who read that section hits a contradiction in the first line of this one, and the contrast the whole section is built on collapses with it.

Suggested replacement:
> `Subagents are a tree: workers your agent spawned, that it owns and that end when their work does.`

Why: the true distinction is ownership and lifetime (parent/child, spawned and ended), not transcript sharing. It is a *stronger* contrast than the wrong one, because independent peers outliving each other is the actual differentiator.

#### C2 — MAJOR: the lead says what the feature *is*, never what it is *for*

Current (sentences 2 to 4):
> `This is the other shape. Two `lop` sessions you started yourself, in different repos, that have never seen each other's context, can talk directly. No cmux, no multiplexer, no network, no cloud service in the middle.`

The topology lands fast, but a skimmer finishes the paragraph knowing the shape and not one thing they would *do* with it. The payoff exists only in the image caption, and captions are the first thing a skimmer's eye skips. "This is the other shape" also spends a whole sentence on a pivot that carries no information.

Suggested replacement for the full opening paragraph (folds in C1):
> `Subagents are a tree: workers your agent spawned, that it owns and that end when their work does. This is the other shape. Two `lop` sessions you started yourself, in different repos, with no parent between them and no shared context, can message each other directly — one can tell another to hold off on a deploy, hand over a finished branch, or claim a shared resource, without you relaying it between terminals. No cmux, no multiplexer, no cloud service in the middle.`

Why: "without you relaying it between terminals" is the sentence that makes a developer recognise their own annoyance. The three concrete verbs are honest to what the feature does and cost one clause.

#### C3 — MAJOR: the "Why Local Operator" bullet spends its closing clause on a constraint

Current:
> `- **Sessions that talk to each other.** Any `lop` session on your machine can message any other with `lop send` or the agent's own `send` tool, so two sessions working in different repos can hand off, ask permission, or claim a shared resource between themselves. Loopback and same-account only, with no multiplexer and no cloud service involved.`

This sits in the project's eight-line pitch, among bullets that all end on capability ("Peek at, steer, pause, or cancel any worker mid-flight", "long sessions don't fall off a cliff"). This one ends on "Loopback and same-account only" — a limitation, stated better and in the right place inside the section itself. "ask permission" is also vague in a list where every sibling verb is concrete.

Suggested replacement:
> `- **Sessions that talk to each other.** Any `lop` session on your machine can message any other with `lop send` or the agent's own `send` tool — so the session running your migration can tell the one holding the deploy slot to wait, and hear back, without you carrying messages between terminals. Same-account and on-machine only: no multiplexer, no cloud.`

Why: leads with a scene the reader recognises, keeps the honest scope limit but ends it as a property rather than an apology.

#### C4 — MINOR: the caption's list breaks parallelism mid-sentence

Current:
> `One claims it, asks for objections, and the other checks its own work and clears it.`

"One claims it, asks for objections, and the other…" reads as a three-item list about *one* subject until "the other" arrives and forces a re-parse.

Suggested replacement:
> `One claims it and asks for objections; the other checks its own in-flight work and clears it. No human in the loop.`

Why: the semicolon marks the subject change, so it parses on the first read. "No human in the loop" is the caption's strongest asset and already earns its place.

#### C5 — MINOR: "no network" overreaches and contradicts the section's own security line

Current:
> `No cmux, no multiplexer, no network, no cloud service in the middle.`

The section later says, correctly, that each session "answers on an authenticated loopback server". There *is* a network stack involved; it is a loopback TCP listener on 127.0.0.1. A developer who reads both lines will trust the second and discount the first.

Suggested replacement:
> `No cmux, no multiplexer, no cloud service in the middle.`

Why: dropping "no network" loses nothing (the cloud and multiplexer claims carry the point) and removes the only sentence in the section that promises more than the feature delivers.

#### C6 — MINOR: throat-clearing in the delivery-modes paragraph

Current:
> `One exception to the non-interrupting default is worth knowing: a session parked in a blocking `wait` returns early, reports that its job is still running, and reads the message at that turn boundary instead of after the full wait budget.`

Suggested replacement:
> `One exception: a session parked in a blocking `wait` returns early, reports that its job is still running, and reads the message at that turn boundary instead of after the full wait budget.`

Why: "is worth knowing" announces that information is coming instead of delivering it, in the section's densest paragraph.

#### C7 — MINOR: "You can see all of it" is the weakest bold lead in the section

Current:
> `**You can see all of it.**`

Every other bold lead here names a concrete property ("Two entry points, one wire", "Targeting is a substring, and it refuses to guess"). This one is vague enough to be about anything.

Suggested replacement:
> `**Every delivery leaves a receipt.**`

Why: matches the README's own established vocabulary ("everything it does leaves a visible receipt in the transcript") and states the actual claim the paragraph proves.

#### C8 — MINOR: the limits paragraph is a four-clause run-on where the file elsewhere uses bullets

Current:
> `Worth knowing before you rely on it: interactive and daemon-owned sessions receive, a headless `exec` session may not, a session running a `lop` older than this feature answers with a clear "cannot receive peer messages" error instead of crashing, and bodies are capped at 256 KB.`

Four unrelated caveats in one sentence; the reader cannot scan for the one that applies to them.

Suggested replacement:
> `Worth knowing before you rely on it:
>
> - Interactive and daemon-owned sessions receive; a headless `exec` session may not.
> - A session running a `lop` older than this feature answers with a clear "cannot receive peer messages" error instead of crashing.
> - Message bodies are capped at 256 KB.`

Why: caveats are look-up material, not prose. (Content verified accurate against `PEER_MESSAGE_MAX_BYTES = 256 * 1024` and the server's error string.)

#### C9 — NIT: `--json` description is incomplete

Current:
> ``--json` adds each session's `cwd`.`

The JSON rows also carry `session_id`, which the table does not show.

Suggested replacement:
> ``--json` also carries each session's `cwd` and `session_id`, which the table omits.`

#### C10 — NIT: awkward article

Current:
> `a session running a `lop` older than this feature`

Suggested replacement:
> `a session running a version of `lop` older than this feature`

#### C11 — NIT: one rhetorical flourish over budget

Current:
> `because a hand-off nobody reads until later is not a hand-off.`

The X-is-not-X aphorism is defensible in this file's voice ("stops being a chatbot and starts being a staff"), but it is doing decoration where the sentence already made its point at "wakes an idle peer". Cutting the clause costs nothing. Keep it if you like the cadence; flagging so the choice is deliberate.

---

**Checked and found correct** (no findings): the `## ↔ Cross-Session Communication` anchor resolves — GitHub slugs the heading to `user-content--cross-session-communication`, which the ToC and top-nav links both match (verified against GitHub's markdown API and the rendered blob); heading Title Case matches sibling H2s; the shortened `Cross-Session` nav label matches how `Tour` and `Headless & Server` are shortened; the alt text is genuinely descriptive rather than decorative; the screenshot's content matches its caption; and the column list, `--steer` alias, `FOOTPRINT`/`HB_AGE` semantics, delivery-detail strings and the 256 KB cap all match the implementation rather than the guide where the two differ (the guide lists a `CWD` column the CLI does not print — the README is right and the guide is stale).

No hype vocabulary, no "not only X but also Y", no hedged intros, no needless adverbs found in the new prose.

**Verdict: NOT TERMINAL.** One blocker (C1, a factually wrong claim that the section's central contrast depends on) and two majors (C2 lead carries no payoff, C3 pitch bullet ends on a constraint). C4–C8 are worth taking in the same pass since they are single-sentence swaps; C9–C11 are optional. Re-review needed on the new head once C1–C3 are addressed.

damianvtran:
### QA report — round 1

**Reviewer:** qa-tester (lopdev team) · anthropic/claude-opus-5
**Scope:** `origin/main..af075a15ded5d3e39d5e43fa3fca672dc2e44d58`
**Verdict:** **FAIL** — 1 major, 1 minor, 1 nit. Rendering, links, asset, and the documented behaviour otherwise all verified by execution.

An independent pass, separate from the code review and from the author's own evidence. Full matrix (one row per cell): 45 cells — 42 PASS, 2 FAIL (same root cause), 1 PENDING.

#### Environment

| Item | Value |
|---|---|
| Binary | `/tmp/qa546/wt-venv/bin/local-operator` → `v0.44.36` (branch code at `af075a15`) |
| Config dir | `LOCAL_OPERATOR_CONFIG_DIR=/tmp/qa546/cfg` — **isolated**, never `~/.local-operator` |
| Sessions | 5 booted by me on the real `RuntimeServer` (real discovery records + real authenticated loopback socket) |
| Rendering | `gh api /markdown` with repo context + the live GitHub blob view in a real browser |

The operator has ~10 live sessions on this host. `registry.run_dir()` resolves against `paths.config_dir()`, which re-reads the env var on every call, so the isolation is total — `lop sessions` in my config dir printed `no active lop sessions` while 15 real record files sat untouched in `~/.local-operator/run/mobile/`. Zero pid collisions between the two sets. **Every pid I sent to was a process I started.**

---

### Findings

#### Q1 — MAJOR — the README's `--pid` example does not work as written

`README.md:418` publishes:

```bash
lop send --pid 12345 --now "hold off, the schema changed"
```

A reader who copies this gets an error and **no delivery**. `send` declares two optional positionals (`target`, `message`); with one positional supplied argparse binds it to **`target`**, leaving `message=None`, so the body falls through to the stdin branch and dies.

Minimal reproduction, against a session I own:

```
$ local-operator send --pid 54817 --now "hold off, the schema changed"
message is empty
EXIT=1

# argparse proof
argv=['--pid','12345','--now','hold off, the schema changed']
   -> target='hold off, the schema changed'  message=None
```

Scope is wider than the one line — any `--pid`/`--session` with a single positional fails the same way (`send --pid N "just a body"` → `message is empty`).

These three all work, which is why every *other* example on the page passes:
```
local-operator send --pid 54817 --now "" "hold off, the schema changed"   # empty target placeholder
echo "hold off, the schema changed" | local-operator send --pid 54817 --now
local-operator send "cutter" --now "hold off, the schema changed"          # two positionals
```

**The CLI defect is pre-existing** — this PR changes no `.py` files, and the same broken form sits in `guides/peer-messaging/GUIDE.md:157`. But this PR is what puts a non-working command in front of the README's public audience, and the README *is* the deliverable here. In-scope fix = correct the example (any of the three working forms). Fixing the CLI grammar so `--pid` makes the lone positional the body is the better long-term answer but is out of scope for a docs PR — if it is not taken now, please record it as `deferred — <reason>` in this thread.

#### Q2 — MINOR — the wedged path is documented in the guide but dropped from the README

The README says only that ambiguity prints candidates and exits non-zero. In practice a `wedged` target gets its own distinct message — `target pid 55109 is wedged, not live (its owner is not responding); try again shortly`, exit 1, **elapsed 0s**. The guide calls this out ("it says so rather than hanging on a dial"); the README does not. Behaviour is correct, the doc is just thinner than the feature. One clause closes it.

#### Q3 — NIT — `--json` adds `session_id` too, not only `cwd`

"`--json` adds each session's `cwd`" is true but incomplete: `session_id` is equally json-only, and `--session` targeting needs exactly that value.

---

### What passed (all executed, not read)

**Rendering (R1–R10).** GFM render clean; the new `` resolves and is not a broken link; `width="720"` matches all nine sibling TUI screenshots (only the phone shot differs at 360, intentionally); wrapped in `<p align="center">` exactly like its neighbours; caption renders as centred italic `<p align="center"><i>…</i></p>`; both fences get `highlight-source-shell`; 45 inline `<code>` spans and the only two `&lt;` escapes are the intentional `<conversation>`/`<model>` placeholders; section sits between `## 🔗 MCP servers` and `## ⚙️ Headless & Server Modes`, exactly where the ToC puts it. Confirmed in a real browser at the branch blob URL — the fragment jump lands on the heading and the image displays.

**Links (L1–L7).** Extracted **all 29** internal anchors in the file (23 markdown + 6 HTML top-nav) and resolved each under github-slugger's real rules (notably: U+FE0F variation selectors are *not* stripped, which is why several existing anchors legitimately start with one). **0 broken.** Anchors present on `main` but missing on the branch: **none** — no regression, no shadowing, no base-slug collisions. Both new links (ToC + top-nav) resolve.

**Asset (A1–A5).** Committed at 239,024 bytes; real PNG, `1568 x 388`, 8-bit RGBA; density and aspect in family with the other TUI shots; raw URL on the branch returns `http=200 image/png`, `shasum` byte-identical to the committed blob.

**Documented behaviour (S1–S12, D1–D21).** Table columns are *exactly* the nine the README lists in the same order. The `cwd` json-only claim is **correct** — no cwd value appears anywhere in the table, and every JSON row carries it. All three `STATE` values reproduced on sessions I own: `live`; `wedged` via `SIGSTOP` + the 45s heartbeat timeout; `stale` via `kill -9`. `FOOTPRINT` matched `/usr/bin/footprint -p` **exactly** (32 MB ↔ 33554432 bytes) and `RSS` matched `ps` exactly; both render `—`, never `0`, on unmeasurable rows. All three documented ack strings came back verbatim from the real code path, and I confirmed delivery **at the receiver** (mode/wake/bytes/sender recorded) rather than trusting the sender's ack. Ambiguity printed both candidates, exited 1, and delivered **nothing**. The self-send guard held on all three targeting paths with no leak. Wedged and cannot-receive both failed softly in 0s with the receiving session still alive. The 256 KB cap is exact at both boundaries in ASCII *and* multibyte (65536 × 4-byte emoji = 262144 accepted; +4 rejected) — byte-counted, not char-counted.

**CI on `af075a15`.** lint, type-check, pip-audit, tui-e2e (macos + ubuntu) all green. `test (3.12)` was still running when I posted; it gates merge on its own and no `.py` files changed in this PR.

### Gaps I could not cover

- **The TUI card itself** (`↔ peer message from "…"`) and the phone surface: **BLOCKED** by the safety constraint — painting it needs a live interactive TUI session and the only ones here are the operator's. I proved delivery at the receiving handle instead; the screenshot in this PR is itself the card evidence.
- **Mixed-version interop**: approximated with a handle lacking `receive_peer_message`, which is the same `getattr` path a genuinely old binary takes — not tested against a real old install.
- **`--wake`/`--now` driving an actual turn**: my receiver records the mode but runs no model, so I verified the mode reached it, not that a turn started.

Re-run on the new head once Q1 is addressed.

damianvtran:
### Agent review — round 1

Scope: origin/main..af075a15

Docs-only diff, but reviewed as behaviour claims: I checked every factual assertion in the new section against the source **and by running the commands**. Almost all of them hold. One example in the section is wrong in a way a reader would hit on their first try.

**How I verified:** ran `lop sessions` / `lop sessions --json` against 15 live sessions; ran the ambiguous, no-match and bad-pid `lop send` paths and checked their real exit codes; parsed all three README example command lines through the real `build_cli_parser()`; delivered two actual peer messages to a live session (pid 13850) and read back the ack strings; checked registry permissions on disk; rendered the branch README through GitHub's own API to resolve the anchors.

---

#### B1 — BLOCKER — the `--pid` example does not deliver a message; it fails with "no message given"

README.md (Cross-Session Communication, example block):

```bash
lop send --pid 12345 --now "hold off, the schema changed"
```

`send` declares **two** positionals, `target` then `message` (`local_operator/cli.py:438-447`), both `nargs="?"`. With `--pid` supplied there is no target to give, so the single quoted string binds to `target`, and `message` stays `None`. `send_command` then falls to the stdin branch; on a terminal stdin is a tty, so it errors out (`cli.py:1187-1193`).

Parsed with the real parser:

```
$ lop send --pid 12345 --now "hold off, the schema changed"
    target='hold off, the schema changed' message=None pid=12345 wake=False steer=False
```

Run for real against a live session, on a tty:

```
$ lop send --pid 13850 --now "hold off, the schema changed"
no message given (pass it as an argument or pipe it on stdin)      # exit 1
```

Nothing is delivered. Note this is the one example in the block a reader is most likely to copy verbatim, because it is the "redirect a session right now" case, and it is also the only example that combines `--pid` with a positional body.

The guide's own example set avoids this by pairing `--pid` with `--wake` **and** keeping a positional target, or by piping. The two forms that actually work:

```
$ echo "hold off, the schema changed" | lop send --pid 13850 --now     # verified: "delivered mid-turn (steered)"
$ lop send "ingest refactor" --now "hold off, the schema changed"       # positional target + body
```

One-line fix: either pipe the body (`echo "…" | lop send --pid 12345 --now`) or drop `--pid` and use the substring target form, matching `guide://peer-messaging`'s example block.

Worth noting for the coder: this is arguably a CLI ergonomics defect too (a `--pid` send has no way to pass a body positionally, which is surprising). Fixing the CLI is out of scope for a docs PR — correcting the example is the right call here; mention the CLI wart in the thread if you want it tracked, but do **not** open an issue.

---

#### Everything else in the section checks out

Recording these explicitly because the brief asked for them, and because a "verified" list is worth as much as a finding list:

- **`lop sessions` columns — correct, and the earlier CWD correction is right.** The header at `cli.py:1266-1267` is exactly `STATE PID KIND CONVERSATION MODEL RSS FOOTPRINT UPTIME HB_AGE`. There is no CWD column in the table; `cwd` is emitted only in the `--json` row dict (`cli.py:1247`). Confirmed on a live run of both. The correction was necessary and is accurate.
- **`FOOTPRINT` semantics — correct, including the `—` claim.** `_format_bytes` returns `—` for `None`, never zero (`cli.py:1096-1097`). I happened to catch a run where every row showed `—` and a later run where they showed real numbers, which is exactly the "shown as `—` when it cannot be measured" case the README describes.
- **All four delivery detail strings match the source literals byte for byte** (`session/session.py:3959`, `:3965`, `:3973`, `:4002`). I also observed two of them live: `delivered mid-turn (steered)` and `delivered to the mailbox (will be read on the next turn)`.
- **Flags.** `--json` (`cli.py:468`), `--pid`, `--session`, `--now`/`--steer` as aliases on one dest, `--wake` (`cli.py:448-461`) all exist as documented. Stdin piping works and is the documented fallback when `message` is omitted — verified with a real delivery.
- **Defaults asymmetry is right.** CLI defaults to mailbox (`mode = "steer" if args.steer else "mailbox"`, `cli.py:1199`; `--wake` is `store_true`, default False) while the `send` tool defaults `wake=True` (`tools/builtin.py:4648-4655`). The README's framing of why is accurate.
- **Security claims verified on disk, not just in source.** `run_dir` chmods `0o700` and `publish` chmods the record `0o600` (`session/runtime/registry.py:39,52`); on this machine `~/.local-operator/run/mobile` is `drwx------` and each `<pid>.json` is `-rw-------`. The server binds `127.0.0.1` with an auth key frame (`session/runtime/server.py:427`, `mobile/peer_client.py:126-129`). Same-account boundary as stated.
- **256 KB cap.** `PEER_MESSAGE_MAX_BYTES = 256 * 1024` (`mobile/peer_send.py:34`); `validate_peer_body` accepts exactly at the cap and rejects at cap+1 with a clear message. Verified in-process.
- **`live`-only eligibility and ambiguous-target non-zero exit.** `resolve_peer_target` filters to `live` and returns candidates on ambiguity (`peer_send.py:57,127-129`); `lop send "e" "test"` printed 15 candidates with `--pid` hints and exited **1**. No-match and bad-pid also exit 1.
- **`exec` may not receive / old-version error.** The `receive_peer_message` handle is `getattr`-probed and a handle lacking it raises `this session cannot receive peer messages` (`server.py:831-835`), surfaced by the client as a soft `RuntimeError` with a non-zero exit, not a traceback (`peer_client.py:114-118,162`). Only `tui` and `daemon` construct a `RuntimeServer` (`tui/app.py:10107`, `session/runtime/process.py:226`), so "interactive and daemon-owned receive, headless `exec` may not" is right.
- **Blocking-`wait` early return.** Real: `wait` parks on the peer-arrival event and returns the still-running payload with `interrupted_by="peer_message"` and the note "a message arrived from another session" (`tools/builtin.py:7361-7396`), leaving jobs uncancelled and unconsumed. "Returns early, reports that its job is still running" is accurate.
- **"A running `bash`, `eval`, or MCP call is never cut short"** — correct *as scoped*. The sentence sits in the mailbox-mode paragraph, and mailbox delivery never touches the interrupt path. (A `--now` steer genuinely does cancel an in-flight tool via `_peek_interrupt` → `tool_task.cancel()`, `harness/loop.py:1487-1489`, `:1998`, but the README does not claim otherwise — the `--now` sentence is separate and says "injects mid-turn to redirect". No finding; flagging only so the scoping is not "fixed" into being wrong.)
- **Anchors resolve.** Rendered the branch README through GitHub's markdown pipeline: the heading emits `id="user-content--cross-session-communication"`, matching both the ToC link and the top-nav link (`#-cross-session-communication`). The `↔` drops out of the slug leaving the double hyphen, which is what both links already encode.
- **Asset is a real tracked PNG.** `static/tui-peer-message.png`, mode `100644`, blob `9ed098dd`, 239,024 bytes, `PNG image data, 1568 x 388, 8-bit/color RGBA` — not an LFS pointer, no `.gitattributes` filter in the repo. Path `./static/tui-peer-message.png` resolves in GitHub's rendered README. I viewed the image: it shows the peer card, the `send` reply, and the wait/bash/hub receipts, and the alt text and caption describe it accurately (pid 50793, MR !1356, the objection request and the clearance).
- **Version bump correct.** main is `0.44.35`, PR sets `0.44.36`, PyPI's latest is `0.44.32`, so `0.44.36` is unreleased and non-colliding. Patch is the right bump for docs per AGENTS.md "Versioning". (FYI, not a finding: PRs #545, #544, #542, #540 and #536 also claim `0.44.36` — whichever lands second needs a re-bump. Normal for parallel PRs.)
- **No unrelated files touched.** `git diff --stat origin/main...af075a15` is exactly the three expected files, +89/−1.

CI on this head: lint, type-check, pip-audit and both `tui-e2e` jobs pass; `test (3.12)` still pending at review time.

---

**Verdict: changes requested** — one BLOCKER (B1). No MAJOR findings; no MINOR or NIT findings worth a round. B1 is a two-token edit to one example line; once that example is corrected to a form that actually delivers, this section is accurate and I expect the next round to be terminal.

damianvtran:

**Implementer:** coder (lopdev team) · anthropic/claude-opus-5
**Scope:** `af075a15..9429451f` — README.md only (+29/−23). No `.py` files touched; `pyproject.toml` stays at `0.44.36` per the manager.

Every finding from all three rounds (copy C1–C11, QA Q1–Q3, code review B1) is answered below by id.

---

### The broken example (QA **Q1** major / code review **B1** blocker) — **fixed** in `9429451f`

Both rounds found this independently, and I reproduced it before changing anything, against a receiver process I started in an isolated registry:

```
$ lop send --pid 74487 --now "hold off, the schema changed"
no message given (pass it as an argument or pipe it on stdin)
FAIL  rc=1 delivered=0
```

Replaced with the substring-target form, which keeps the `--now` redirect case represented as instructed:

```
$ lop send "ingest refactor" --now "hold off, the schema changed"
→ ingest refactor (pid 74487): delivered mid-turn (steered)
PASS  rc=0 delivered=1
receiver saw: {"event":"received","text":"hold off, the schema changed","mode":"steer","wake":false}
```

**Then I ran every command in both fences verbatim, exactly as a reader would copy them** (including trailing comments), against two real receiver processes, asserting on the CLI's real exit code *and* on delivery observed at the receiver rather than on the sender's ack:

```
PASS  rc=0            $ lop sessions
PASS  rc=0            $ lop sessions --json     # same fields as objects, byte counts, for scripting
PASS  rc=0 delivered=1  $ lop send "release cutter" "gates are green, ready for review"
                        -> delivered to the mailbox (will be read on the next turn)
PASS  rc=0 delivered=1  $ lop send "release cutter" --wake "the deploy finished; verify prod"
                        -> delivered and woke the session
PASS  rc=0 delivered=1  $ lop send "ingest refactor" --now "hold off, the schema changed"
                        -> delivered mid-turn (steered)
PASS  rc=0 delivered=1  $ git log -1 --stat | lop send "release cutter"      # body from stdin
                        -> delivered to the mailbox; receiver got the full commit text
```

6/6 pass. The piped case was confirmed by reading the receiver's log: the `git log` body arrived intact (`"commit af075a15…\nAuthor: …"`), not truncated or empty.

Harness: two receiver **processes** (each with its own pid and its own `0600` discovery record) on the real `RuntimeServer`, under `LOCAL_OPERATOR_CONFIG_DIR=/tmp/peer-harness/cfg`. Every pid I sent to was a process I started; the operator's live sessions were never touched.

**The underlying CLI defect is `deferred — out of scope for a docs PR`.** `--pid`/`--session` with a single positional binds it to `target`, so there is no way to pass a body positionally alongside `--pid` (`cli.py:438-447`). Both reviewers reached the same conclusion. Per the team's standing rule I have filed no issue; this comment is the record. Note the same broken form also sits in `guides/peer-messaging/GUIDE.md:157`, which is outside this PR's diff.

---

### Copy findings

**C1 — BLOCKER, factually wrong contrast — fixed.** "one shared transcript" is gone. Took the reviewer's pivot to ownership and lifetime:

> Subagents are a tree: workers your agent spawned, that it owns and that end when their work does.

I checked this replacement is itself true before shipping it: subagents are spawned by and owned by the parent, and `hub op='resume'` exists precisely because a stopped child is otherwise finished. No new claim about transcripts is made in either direction.

**C2 — MAJOR, lead had no payoff — fixed.** Adopted the reviewer's clause; "without you relaying it between terminals" is the strongest line offered and I did not try to beat it. The lead now carries three concrete uses (hold off on a deploy, hand over a finished branch, claim a shared resource) before the topology.

**C3 — MAJOR, Why bullet ended on a constraint — fixed.** Adopted the suggested rewrite. It now opens on the migration/deploy-slot scene and ends `Same-account and on-machine only: no multiplexer, no cloud` — the scope as a property, not an apology. Verified in a rendered frame that it sits naturally among its siblings.

**C4 — fixed.** Semicolon marks the subject change: `One claims it and asks for objections; the other checks its own in-flight work and clears it.`

**C5 — fixed.** "no network" dropped. The reviewer is right that it contradicted the section's own loopback-server line.

**C6 — fixed.** `One exception:` replaces `One exception to the non-interrupting default is worth knowing:`.

**C7 — fixed.** `**Every delivery leaves a receipt.**` replaces `**You can see all of it.**`.

**C8 — fixed.** The four-caveat run-on is now a three-item list. Confirmed in a rendered frame that it reads as scannable look-up material.

**C9 / QA Q3 — fixed.** Now: `--json` also carries each session's `cwd` and `session_id`, which the table omits. Confirmed against real `--json` output, which shows both fields with no `session_id` column in the table.

**C10 — fixed.** `a version of lop older than this feature`.

**C11 — accepted (cut).** The reviewer left this to my judgement. Cut: the sentence had already made its point at "wakes an idle peer", and the section is dense enough that a decorative clause is the right thing to lose.

### QA findings

**Q2 — fixed.** Added the wedged clause. I verified the behaviour rather than copying it from the guide: froze a receiver with `SIGSTOP`, waited out the heartbeat timeout, and sent to it.

```
STATE       PID KIND    CONVERSATION      MODEL         RSS FOOTPRINT   UPTIME  HB_AGE
wedged    76674 tui     wedge target      test/model      —         —      52s     52s

$ lop send --pid 76674 "does this hang?"
target pid 76674 is wedged, not live (its owner is not responding); try again shortly
EXIT=1 elapsed=1s
```

Answers in 1s, does not hang. That run also incidentally re-confirms `FOOTPRINT` rendering `—` rather than `0` on an unmeasurable row.

**Q3 — fixed** (same edit as C9).

### Nothing rejected

No finding from any round was rejected. C11 was the only judgement call and I took the reviewer's option.

---

### Gates on `9429451f`

| Gate | Result |
| --- | --- |
| `flake8 .` | pass (rc=0) |
| `black --check .` (26.1.0) | pass — 802 files unchanged |
| `isort --check .` (5.13.2) | pass |
| `pyright` | pass — 0 errors, 0 warnings |

**Anchors re-verified** after the ToC/nav edits: my slug model was re-validated against live GitHub's own `user-content-*` ids (reproduces every existing slug exactly), then applied to the branch — **29/29 internal links resolve, 0 broken**.

**Re-rendered and looked at**, since the lead paragraph, the caption, the Why bullet and the limits list all changed: rendered through `gh api /markdown`, served over loopback, and inspected as three frames — the rewritten lead + caption + screenshot, the Why bullet among its siblings, and the new limits list. All correct.

Ready for round 2 on `9429451f`.

damianvtran:
### Copy review — round 2

Reviewer: copy-reviewer (anthropic/claude-opus-5)
Scope: af075a15..9429451f (README.md only, +29/−23)

Verified the remediation rather than accepting it: re-read the rewritten prose as rendered, re-checked every changed claim against the implementation, and scanned the new wording for AI-isms and overreach introduced during the rewrite.

---

#### C1 — BLOCKER — **fixed, and the replacement is true**

This was the finding that mattered, so I checked the new contrast on its own merits instead of just confirming the old string was gone.

> `Subagents are a tree: workers your agent spawned, that it owns and that end when their work does.`

Both halves hold:

- **"your agent spawned, that it owns"** — `run_subagent` builds a *child* `Session` and registers it on the parent's job manager; `local_operator/harness/subagent.py:319-324` states the parent's dispose cancels it "like every other job". Ownership is real and structural, not a metaphor.
- **"end when their work does"** — the runner "drives ONE prompt to completion, and settles the job" (`subagent.py:1-5`). A child is a one-prompt lifetime. `hub op='resume'` rebuilds a child on its old session directory rather than keeping one alive, which is consistent with, not contrary to, the claim.

No new transcript claim is made in either direction, so the contradiction with the subagent section ("peek at *its* transcript") is gone. **This is the honest contrast, and it is a sharper one than the false version it replaces.**

#### C2 — MAJOR — **fixed; the lead now lands**

Read cold, as a skimmer: sentence one gives the contrast, sentence two the pivot, sentence three the topology (independent, different repos, no parent, no shared context), sentence four **what you would use it for**, sentence five the scope. A developer now knows within two sentences how this differs from the subagent tree they just read about, and the payoff no longer lives only in the caption. `without you relaying it between terminals` is doing the work it was meant to do.

#### C3 — MAJOR — **fixed**

The Why bullet now opens on the migration/deploy-slot scene and closes `Same-account and on-machine only: no multiplexer, no cloud` — scope as a property, not an apology. Read among its siblings in a rendered frame, it matches their shape: concrete capability first, ending on a fact rather than a limitation.

#### C4–C11 — **all fixed**, verified in the rendered output

| id | state | verified |
| --- | --- | --- |
| C4 | fixed | Semicolon marks the subject change; parses on first read. |
| C5 | fixed | `no network` dropped; no longer contradicts the loopback-server line. |
| C6 | fixed | `One exception:` — throat-clearing gone. |
| C7 | fixed | `**Every delivery leaves a receipt.**` — matches the file's own "visible receipt" vocabulary. |
| C8 | fixed | Renders as a real `<ul>` with exactly 3 `<li>` (confirmed through GitHub's markdown API, not eyeballed). |
| C9 | fixed | `cwd` and `session_id` both present in the row dict (`cli.py:1247,1249`); neither is a table column. |
| C10 | fixed | `a version of lop older than this feature`. |
| C11 | cut | The decorative clause is gone; the sentence ends at "wakes an idle peer". |

#### New copy introduced this round — checked, no findings

Two sentences are new rather than edited, so they get the same scrutiny as the original text:

- `A target that has gone wedged says so immediately instead of hanging on the dial.` — **true.** `resolve_peer_target` classifies from the registry scan *before* any dial and returns `target pid N is wedged, not live (its owner is not responding); try again shortly` (`peer_send.py:60-72`), with the substring path returning an equivalent refusal (`:120-127`). "Immediately" and "instead of hanging on the dial" are both literally accurate — the refusal happens pre-dial.
- `One can tell another to hold off on a deploy, hand over a finished branch, or claim a shared resource` — these are use cases for a delivered message, not capability claims; nothing here promises coordination the feature does not provide.

**AI-ism scan of the final prose:** no `seamless`, `robust`, `unlock`, `empower`, `leverage`, `elevate`, `delve`, `not just X`, `isn't just X`, `not only X but also Y`, `it's important to note`, `when it comes to`, or `in today's …`. No hedged intro, no needless adverbs. Nothing crept in during the rewrite; the edits removed AI-isms on net (C6, C11).

**Overreach re-check:** the one overreach I found in round 1 (`no network`) is gone, and no replacement overclaim was introduced. The lead's scope disclaimer and the security paragraph now agree with each other.

#### Follow-up (non-blocking, not worth another round)

`--json` also carries each session's `cwd` and `session_id`, which the table omits — precise about the *columns*, and there is no `session_id` column. Strictly, the table does *print* a session id inside the `CONVERSATION` column as a fallback when a session has no conversation name (`cli.py:1271`). "Which the table omits" is true of the column and very slightly loose about the rendered cell. Not worth a change; recorded here so the record is complete.

---

**Verdict: TERMINAL.** The blocker is fixed and its replacement independently verified as true, both majors are fixed, all five minors and three nits are resolved, and the rewrite introduced no new AI-ism, no new overreach, and no regression in the paragraphs it touched. Reading the section end to end at `9429451f`, it now leads with the differentiator, the caption earns the section, and the Why bullet holds its own among the project's pitch lines. **No copy findings remain on the current head.** Nothing from my round blocks merge.

damianvtran:
### QA report — round 2

**Reviewer:** qa-tester (lopdev team) · anthropic/claude-opus-5
**Scope:** `af075a15..9429451f` (my round-1 findings + regressions); full-PR checks re-run against merge-base `6ae89bc8..9429451f`
**Verdict:** **PASS — this round is TERMINAL** from QA's side. No Q-findings. Merge still gated on `test (3.12)`, which had not concluded when I posted.

I read the remediation and then verified the claims rather than accepting them. 60 cells: 58 PASS, 0 FAIL, 1 PENDING (CI), 1 informational note.

#### Environment

| Item | Value |
|---|---|
| Binary | `/tmp/qa546r2/wt-venv/bin/local-operator` → `v0.44.36` (branch code at `9429451f`) |
| Config dir | `LOCAL_OPERATOR_CONFIG_DIR=/tmp/qa546r2/cfg` — **isolated**, never `~/.local-operator` |
| Receivers | 2 named **verbatim after the README's own example targets** (`release cutter`, `ingest refactor`), plus 3 throwaways, all on the real `RuntimeServer` |
| Rendering | `gh api /markdown` with repo context + the live GitHub blob view at `9429451f` in a real browser |

`lop sessions` in my config dir showed only my own processes while the operator's 15 live records sat untouched. Every pid I sent to, `SIGSTOP`ped, or killed was a process I started.

---

### Q1 — FIXED. Every fenced command now runs verbatim

I ran each line through a real `lop` shim on `PATH`, **exactly as printed** including trailing comments, with `stdin=/dev/null` so a missing body could not be silently rescued:

```
rc=0 PASS  $ lop sessions
rc=0 PASS  $ lop sessions --json     # same fields as objects, byte counts, for scripting
rc=0 PASS  $ lop send "release cutter" "gates are green, ready for review"
rc=0 PASS  $ lop send "release cutter" --wake "the deploy finished; verify prod"
rc=0 PASS  $ lop send "ingest refactor" --now "hold off, the schema changed"
rc=0 PASS  $ git log -1 --stat | lop send "release cutter"      # body from stdin
```

**6/6.** Delivery confirmed **at the receiver**, not from the sender's ack:

```
release cutter (90237): mailbox/33B · wake/32B · mailbox/1451B ("commit 9429451f… | Author: Damian…")
ingest refactor (90238): mode=steer wake=False bytes=28 "hold off, the schema changed"
```

The piped body arrived intact (1451 B, starts `commit `, diffstat present) — not truncated, not empty. The replaced `--now` line lands as `mode=steer`, which is the point of the flag.

### Q2 — FIXED. Wedged clause matches real behaviour

`SIGSTOP` a receiver I own, wait out the 45 s heartbeat timeout, then send:

```
wedged  90238 tui  ingest refactor  openai/gpt-5   —   —   1m   53s

$ lop send --pid 90238 "does this hang?"
target pid 90238 is wedged, not live (its owner is not responding); try again shortly
EXIT=1  elapsed=0.34s  -> IMMEDIATE
```

Also correct through the substring path the README documents: `the only match for 'ingest refactor' is not responding (pid 90238)`, EXIT=1.

### Q3 — FIXED, and the new sentence is exactly right

`--json also carries each session's cwd and session_id, which the table omits`:

```
table: none of repo-cutter / repo-ingest / qa2-cutter-01 / qa2-ingest-02 appear
json-only field set: ['cwd', 'session_id']   <- exactly what the sentence claims
```

---

### C1 — the replacement contrast is TRUE, not just different

This was the whole point of the finding, so I checked it in source rather than accepting the rewrite. New text: *"workers your agent spawned, that it owns and that end when their work does."*

- **"spawned … that it owns"** — `harness/subagent.py:323`: *"The parent session's dispose cancels it through `jobs_manager.dispose()` like every other job."* Parent lifetime bounds the child. **TRUE.**
- **"end when their work does"** — module docstring: the runner *"drives ONE prompt to completion, and settles the job"*, settling `status="completed"`. **TRUE.**
- **Not falsified by `hub op='resume'`** — `resume_dir` builds a **new** child on the old session directory and *"replays what the stopped child had already done"* (`subagent.py:100`). That is a relaunch of an ended child, not a child that never ended. **TRUE.**

The same file also confirms the old claim was false: `subagent.py:7` — *"the child gets its **own** transcript directory under `config_dir()/sessions/<hex>`"*. So "one shared transcript" was wrong and the replacement is right.

**Does it land as a skimming developer reads it?** Yes, plainly. The first sentence names the axis a developer actually cares about — who owns the worker and when it dies — then "this is the other shape", then the mirror image: *"no parent between them and no shared context."* The differentiator lands in three lines with no topology lecture, and the payoff (*"without you relaying it between terminals"*) arrives before the reader hits any constraint. The Why bullet stands on its own the same way: concrete scene first (the migration session telling the deploy-slot holder to wait), scope last as a property rather than an apology.

### C2–C11 — all fixed

C2 payoff clause and three concrete uses ahead of the topology; C3 bullet ends `Same-account and on-machine only: no multiplexer, no cloud.`; C4 semicolon; C5 `no network` gone (it did contradict the loopback-server line); C6 `One exception:`; C7 `**Every delivery leaves a receipt.**`; C8 renders as `<ul>` with exactly **3** `<li>`; C10 `a version of lop older than this feature`; C11 decorative clause cut.

**No new AI-ism or overreach.** 0 hits across 28 markers (seamless/robust/unlock/leverage/powerful/crucial/…) in the added lines; **0** em dashes added; the single rule-of-three is the legitimate three-uses list C2 asked for. The two absolutes that remain are both verified: `no multiplexer, no cloud` (loopback bind + `0600` record) and `is never dropped` (`--now` opens a turn when idle).

### Changed limits list — re-verified, not assumed

| Claim | Result |
|---|---|
| older `lop` → clear error, no crash | `could not deliver: this session cannot receive peer messages`, EXIT=1, **receiver still ALIVE** |
| bodies capped at 256 KB | 262144 delivered; 262145 → `message is too large (262145 bytes); cap is 262144 bytes`; 65536×4-byte emoji (262144 B) delivered — byte-counted |
| interactive/daemon receive, `exec` may not | capability-probed via `getattr`, not assumed |

### Regressions — none

Table columns unchanged · ambiguity lists both candidates, EXIT=1, **nothing delivered** · unknown target · nonexistent pid · empty-body guard · `--session` exact id · cwd-basename targeting · `stale` after `kill -9` · `FOOTPRINT` matched `/usr/bin/footprint` exactly (38 MB ↔ 39845888 B) · `—` never `0` · self-send guard refused all three paths with `SELF-DELIVERY LEAKED: False`.

### Rendering + anchors

**29/29 internal anchors resolve, 0 broken.** Anchors lost vs `af075a15`: **NONE**; vs merge-base: **NONE**; no base-slug collisions. Section renders with 2 shell-highlighted fences, the image still `width="720"` inside `<p align="center">` and `<a>`-wrapped, the rewritten caption centred italic with its semicolon, the new limits list as 3 scannable items, and the only `&lt;` escapes being the intentional `<conversation>`/`<model>` placeholders. Confirmed in a browser at `9429451f`: the anchor jump lands on the heading and every changed paragraph renders correctly.

### Diff scope

No `.py` in round 2 (`README.md` only) and none in the whole PR — against the **merge-base** `6ae89bc8` the contribution is `README.md` + `static/tui-peer-message.png` + `pyproject.toml`, whose sole hunk is `0.44.35` → `0.44.36`. Worth flagging for whoever merges: diffing against `origin/main` is misleading here because main has moved ahead and that diff lists 18 unrelated files.

### CI on `9429451f`

lint ✅ · type-check ✅ · pip-audit ✅ · tui-e2e macos ✅ · tui-e2e ubuntu ✅ · **`test (3.12)` still `in_progress`** (~6.6 min in, run `33658722515`).

**On the round-1 pending check:** it did not fail — it was **`cancelled`**, superseded by this push, so it never concluded on `af075a15`. The current run is the first real result for this PR, and it is still going. I cannot report a conclusion that does not exist yet; that is the one thing this PASS does not cover. It gates merge on its own, and no `.py` changed.

### Notes (non-blocking, neither a defect in the change)

- **N1** — `test (3.12)` unconcluded at post time, as above.
- **N2 (nit, remediation wording)** — the remediation says anchors were re-verified *"after the ToC/nav edits"*, but round 2 edited no ToC or nav line (the link set is byte-identical to `af075a15`). The anchor claim itself is correct; I measured 29/29 independently.

The deferral of the underlying CLI grammar defect is **accurate as written**: this PR touches no `.py`, and the same broken form does still sit at `guides/peer-messaging/GUIDE.md:157`, outside this diff.

---

**TERMINAL.** All three round-1 findings are fixed and independently verified by execution; the C1 replacement is true against the source; nothing regressed. Merge on `test (3.12)` going green.

**Gaps I could not cover:** the TUI peer card and phone surface remain **BLOCKED** (they need a live interactive session and the only ones here are the operator's) — I proved delivery at the receiving handle instead; mixed-version interop was approximated with a handle lacking `receive_peer_message`, the same `getattr` path a real old binary takes; and `--wake`/`--now` were verified as reaching the receiver with the right mode, not as starting a real model turn.

damianvtran:
### Agent review — round 2

Scope: af075a15..9429451f

I re-verified the remediation rather than accepting it. Every claim in the remediation comment that I could check independently, I checked — by running the commands against receivers I started myself, and by reading the source for the sentences that changed.

---

#### B1 (round 1 BLOCKER) — **confirmed fixed**

The `--pid 12345 --now "…"` example is gone, replaced by `lop send "ingest refactor" --now "hold off, the schema changed"`. Parsed through the real `build_cli_parser()`, that binds `target='ingest refactor'`, `message='hold off, the schema changed'`, `steer=True` — the body reaches `message`, which is what was broken.

I then ran **every command in both fences verbatim** against two receiver processes I started (real `RuntimeServer`, own pids, own `0600` records, `LOCAL_OPERATOR_CONFIG_DIR=/tmp/rev2-harness/cfg`), asserting delivery **at the receiver** rather than trusting the sender's ack:

```
rc=0  lop send "release cutter" "gates are green, ready for review"
      -> delivered to the mailbox (will be read on the next turn)
rc=0  lop send "release cutter" --wake "the deploy finished; verify prod"
      -> delivered and woke the session
rc=0  lop send "ingest refactor" --now "hold off, the schema changed"
      -> delivered mid-turn (steered)
rc=0  git log -1 --stat | lop send "release cutter"      # body from stdin
      -> delivered to the mailbox (will be read on the next turn)
```

Receiver log (4/4 arrived, correct mode/wake, body intact):

```
to='release cutter'  mode=mailbox wake=False text='gates are green, ready for review'
to='release cutter'  mode=mailbox wake=True  text='the deploy finished; verify prod'
to='ingest refactor' mode=steer   wake=False text='hold off, the schema changed'
to='release cutter'  mode=mailbox wake=False text='commit b7542b2e76b6ef29cd911ffc57bbdb1adb7f9cc…'
```

`lop sessions` in that isolated registry listed only my two receivers. No session I did not start was contacted at any point.

#### C1 — the replacement contrast is **true**, not just different

This was the point of the finding, so I checked the new sentence against the source rather than reading it for tone:

> Subagents are a tree: workers your agent spawned, that it owns and that end when their work does.

- *spawned / owned*: `run_subagent` registers the child as an AsyncJob whose runner the parent owns; "the parent session's dispose cancels it through `jobs_manager.dispose()` like every other job" (`harness/subagent.py:318-325`).
- *end when their work does*: the runner "drives ONE prompt to completion, and settles the job" (`subagent.py:1-5`).
- The false claim is gone, and — importantly — **no new transcript claim replaced it in either direction**. The child in fact gets *its own* transcript dir (`subagent.py:7`), so the old line was backwards; the new line simply does not go there. Correct call.

#### Reading the new lead and Why bullet as a skimming developer

They land. The lead now answers "why do I care" inside the first two sentences — `One can tell another to hold off on a deploy, hand over a finished branch, or claim a shared resource, without you relaying it between terminals` — and the topology comes after the payoff instead of instead of it. The Why bullet opens on the migration/deploy-slot scene and closes on `Same-account and on-machine only: no multiplexer, no cloud`, which reads as scope, not apology. Among its siblings it is the same shape as the others (concrete capability, then its boundary).

**No new AI-isms or overreach.** I scanned the added lines for the usual tells (seamless/robust/unlock/powerful/leverage/"not just"/"whether you"…): none. Zero em-dashes added. Three absolute claims were added or kept, and all three hold: "never cut short" is scoped to the mailbox paragraph (still correct); "One exception:" still describes the real `wait` early-return; and **"Every delivery leaves a receipt"** is literally true — all four delivery paths call `_emit_peer_receipt` before returning (`session/session.py:3957, 3962, 3970, 3993`).

#### Changed sentences re-verified

- **C9/Q3** — `--json` also carries `cwd` and `session_id`, which the table omits: confirmed both in `sessions_command`'s row dict (11 keys, incl. both) and against live `--json` output; the table header is still the 9 documented columns with neither field. Accurate as written.
- **Q2** — the new wedged clause. Reproduced independently: `SIGSTOP` on my own receiver, waited out the heartbeat, then sent to it. `rc=1, elapsed=0s` both by pid (`target pid 90400 is wedged, not live…`) and by substring (`the only match for 'ingest refactor' is not responding…`). "Says so immediately instead of hanging on the dial" is correct, and the substring path — the one the sentence actually sits next to — behaves the same way.
- **C4, C6, C7, C8, C10, C11** — all present as described. The limits list renders as a real `<ul>` in GitHub's own output, not a run-on.

#### No regressions

Re-checked all 14 round-1-verified claims still standing in the section: column list, no-CWD, the four delivery strings, `0600`/`0700`, loopback, 256 KB, `live`-only, non-zero ambiguous exit, the `wait` early return, the `exec`/old-version limits. All intact. `no network` (C5) and `one shared transcript` (C1) are both gone.

**Anchors**: 29/29 internal links resolve against GitHub's real rendered `user-content-*` ids for this head, after the ToC/nav edits. I confirmed the render was the new head (it contains the new prose and the corrected fence, and no `pid 12345`).

**Scope**: `git diff --stat af075a15..9429451f` is README.md only, +29/−23. No `.py` file changed anywhere in the PR (`git diff origin/main...9429451f -- '*.py'` is empty). `pyproject.toml` still `0.44.36`.

**CI on this head**: lint, type-check, pip-audit and both `tui-e2e` jobs pass. `test (3.12)` is still `IN_PROGRESS` at the time of posting — noted as **pending and manager-owned**; with no `.py` change in the diff it is not a gate on this review round.

---

**Verdict: clean — this round is TERMINAL.** My round 1 BLOCKER is fixed and verified by execution, the copy findings are genuinely resolved rather than reworded, nothing regressed, and I have no MINOR or NIT findings worth another round. The only open item I am aware of is the deferred CLI ergonomics defect (`--pid` with a positional body binds to `target`), correctly recorded in-thread as out of scope for a docs PR — and noted there, it also still sits in `guides/peer-messaging/GUIDE.md:157`, outside this diff.

damianvtran:
### Agent review — round 3 (freshness)

Scope: 9429451f..56c94c2d

Not a re-review. I verified the equivalence claim independently rather than taking it on trust, and it holds.

**Blob identity — checked myself, both locally and against GitHub.**

```
9429451f  README 6d685afca2b57dfb9e38c208368b617f116f4fa3   png 9ed098ddecee46471663a1d29fd6eb109c3d665f
56c94c2d  README 6d685afca2b57dfb9e38c208368b617f116f4fa3   png 9ed098ddecee46471663a1d29fd6eb109c3d665f
```

GitHub's contents API on the pushed head reports the same two shas (`6d685afc…`, `9ed098dd…`). Identical blob = byte-identical file, so the prose I cleared in round 2 is unchanged by construction — no re-read required, and none done.

**Only content delta is the version line.** New merge-base is `b5616f58` (`feat(tui): collapse a large paste to a chip, like a screenshot (#536)`). Diff against it is the same three files (+95/−1), and the sole non-README/png change is:

```
-version = "0.44.36"
+version = "0.44.37"
```

**Nothing else rode in on the rebase.** Filtering the diff to `*.py`, `*.yml`, `*.yaml`, `*.toml`, `*.lock`, `*.cfg` returns `pyproject.toml` and nothing else — still zero `.py` changes in this PR. Branch is three commits on top of the base, all expected (the two docs commits plus the release bump).

**The bump is correct.** `origin/main` is now at `0.44.36` and PyPI's latest is `0.44.36`, so the collision was real; `0.44.37` returns 404 on PyPI (unclaimed). Patch remains the right bump for a docs change.

**Round 2 verdict carries to `56c94c2d`: clean — TERMINAL.** The deferred CLI ergonomics item (`--pid` with a positional body binds to `target`, and the same broken form still sits in `guides/peer-messaging/GUIDE.md:157`) remains recorded in-thread as out of scope, unchanged by the rebase.

damianvtran:
### Copy review — round 3 (freshness)

Reviewer: copy-reviewer (anthropic/claude-opus-5)
Scope: 9429451f..56c94c2d

Verified the equivalence claim independently rather than accepting it. Four checks:

**1. README blob identity — holds.**

```
9429451f  README: 6d685afca2b57dfb9e38c208368b617f116f4fa3
56c94c2d  README: 6d685afca2b57dfb9e38c208368b617f116f4fa3
```

GitHub reports the same blob on the pushed branch (`contents/README.md?ref=56c94c2d` → `6d685afca2b57dfb9e38c208368b617f116f4fa3`), so local and remote agree. A direct `diff` of the two blobs returns **0 byte difference**.

**2. Asset blob identity — holds.** `static/tui-peer-message.png` is `9ed098ddecee46471663a1d29fd6eb109c3d665f` at both SHAs, so the screenshot backing the caption I cleared is the same file.

**3. Only the version line changed.** Against the new merge-base `b5616f58` (which is current `origin/main`), the PR contributes exactly three files: README.md (+94), `static/tui-peer-message.png` (new), `pyproject.toml` (+1/−1). The pyproject delta is one line:

```
-version = "0.44.36"
+version = "0.44.37"
```

Main publishes `0.44.36` at `b5616f58`, so the bump resolves the collision and nothing else moved in that file.

**4. Nothing rode in on the rebase.** Main did not touch `README.md` at all between the old base and `b5616f58` (empty diff for that path), so there was no silent content merge into the section I reviewed. The rebase moved the branch onto newer main commits; it did not alter the branch's own contribution.

Note for anyone reading the raw `git diff 9429451f 56c94c2d`: that spans ~30 files, but every one of them is main's own advance pulled in by the rebase, not a change to this PR. The meaningful comparison is contribution-vs-merge-base, above.

---

**Verdict: TERMINAL, carried to `56c94c2d`.** The prose I cleared in round 2 is byte-identical at the new head, the caption's image is unchanged, and the only content delta is the version bump. My round 2 verdict stands unchanged and is now fresh against the current SHA. No copy findings on this head; nothing from my round blocks merge.

The two guide-file items I recorded in round 2 remain out of scope for this PR and unaffected by the rebase: the broken `lop send --pid N "body"` form still in `guides/peer-messaging/GUIDE.md:157`, and the stale `CWD` column in that same guide. Both are `deferred — out of scope for a docs PR`; this comment plus round 2 is the record.

damianvtran:
### QA report — round 3 (freshness)

**Reviewer:** qa-tester (lopdev team) · anthropic/claude-opus-5
**Scope:** `9429451f..56c94c2d`
**Verdict:** equivalence **HOLDS**. My round 2 PASS carries to `56c94c2d` and remains **TERMINAL**.

I verified the claim against git objects myself rather than taking the report on trust. Not a re-review; the matrix was not re-run.

#### Content identity — README and asset are byte-identical

```
                        README.md blob                         tui-peer-message.png blob
old head 9429451f       6d685afca2b57dfb9e38c208368b617f116f4fa3   9ed098ddecee46471663a1d29fd6eb109c3d665f
new head 56c94c2d       6d685afca2b57dfb9e38c208368b617f116f4fa3   9ed098ddecee46471663a1d29fd6eb109c3d665f
GitHub API @56c94c2d    6d685afca2b57dfb9e38c208368b617f116f4fa3   9ed098ddecee46471663a1d29fd6eb109c3d665f
```

Independent of git's own hashing, the content sha256 matches too:

```
README.md  old/new: b18db8c765a3c160cab0923552caf878a39ea2ef8b092b8b3740f975c76603d8  (identical)
png        old/new: 5d2b624971ddb84d384e70acaf0f7ab7a9d7a5035b997cf809568c411dcca6a9  (identical)

$ git diff 9429451f 56c94c2d -- README.md static/tui-peer-message.png
(empty)
```

So the prose and the screenshot I cleared in round 2 are the same bytes. Every rendering, anchor and asset cell from that round applies unchanged.

#### Nothing else rode in

New merge-base is `b5616f58` — exactly current `origin/main`. The PR's own contribution there is the expected three files and nothing more:

```
$ git diff --stat b5616f58 56c94c2d
 README.md                   |  94 +++++++++++++
 pyproject.toml              |   2 +-
 static/tui-peer-message.png | Bin 0 -> 239024 bytes
 3 files changed, 95 insertions(+), 1 deletion(-)

paths outside {README.md, static/tui-peer-message.png, pyproject.toml}: none
```

The stricter check — every path differing head-to-head must be explained by main's own advance:

```
paths differing 9429451f..56c94c2d              : 30
paths main changed on its own 6ae89bc8..b5616f58: 30
differing head-to-head but NOT touched by main  : none
```

Zero unexplained paths. The 30 files in the head-to-head diff are main's advance (evaluation runner, paste-collapse TUI work, compaction pruning), not this PR's.

#### The version line is the only content delta

```
$ git diff b5616f58 56c94c2d -- pyproject.toml
-version = "0.44.36"
+version = "0.44.37"
```

Sole hunk, no other key touched. The collision story checks out: main at `b5616f58` carries `0.44.36`, the old PR head also carried `0.44.36`, and PyPI confirms **`0.44.36` is already published** while **`0.44.37` is free** — so the bump is necessary and the new version can actually publish.

Commits on the new base: `34c31cbf` (original docs) → `6cb8f0a7` (round-2 fixes) → `56c94c2d` (version bump). Clean linear replay, no squashing or content drift.

#### CI on `56c94c2d`

The rebase started a fresh run (`33661927956`), so every check is re-running against the new head. At post time: lint ✅, type-check ✅, pip-audit ✅; `test (3.12)`, `tui-e2e (macos)`, `tui-e2e (ubuntu)` still in progress.

I own the CI cell and am watching it — **I will report the `test (3.12)` conclusion in a follow-up comment as soon as it settles.** Note it has never yet concluded on this PR: the round-1 run was `cancelled` when round 2 was pushed, and the round-2 run was superseded by this rebase. This is the first run that should reach a verdict. No `.py` file changed in this PR, so my verdict does not depend on it — but it gates merge on its own.

---

**TERMINAL.** The equivalence claim is true: identical README and asset bytes, the version line as the only content delta, and nothing extra carried in by the rebase. Round 2's PASS stands on `56c94c2d`. Merge gated only on CI going green.

PR #548, opened 2026-09-02
fix(evaluation): forward begin_rescue to the adapter so rescue can tear down

damianvtran:
## Summary

A rescue worker could never terminate anything.

`AdapterWorker._dispatch` handled `begin_rescue` entirely itself — validating pins, storing the descriptor, returning `AckResult` — and **never called the adapter**. `params.secrets`, the only credential a rescue worker ever receives, were dropped on the floor (`grep -c secrets worker.py` → `0`). The adapter's provider therefore stayed `None`, and every cleanup action took the honest "could not look" branch, reporting `attempted` / `terminate-unconfirmed`.

The visible symptom was a sweep that never confirmed teardown and never discarded its descriptor. The **serious** consequence is that a genuinely leaked instance would never have been terminated by a sweep, because no provider existed to terminate it. `OSWorldV2Adapter.begin_rescue` — which builds `AwsProvider.for_teardown` from the descriptor's region and the delivered credentials — was dead code in production.

Depends on #543 (branched off `dev-adapter-cache-dir` @ `379cecee`); `0.44.38` sits on top of that PR's `0.44.37`, so please re-sequence at merge time if #543 lands differently.

## How it was found

The reported defect was an instance that had aged out of the EC2 API, with the hypothesis that an empty tag lookup was being mapped to `terminate-unconfirmed`. That hypothesis is **wrong**, and fixing it would have made the sweep pass by coincidence while the provider stayed `None`:

- `AwsProvider._terminate` already returns `EVIDENCE_INSTANCE_ABSENT` for an empty tag query, and `_terminate_status` already maps that to `not_needed`. That path was never broken.
- Read-only AWS confirmed `describe-instances --instance-ids i-0a76923056da31d04` returns **empty without raising** `InvalidInstanceID.NotFound`, so a "positive NotFound evidence" check would not have fired anyway.
- The observed code triple is the exact fingerprint of the `provider is None` branch: `revoke_lease` also reported `schedule-absent` with status `attempted`, where a real provider returning absent yields `not_needed`.

## Change

`local_operator/evaluation/adapters/worker.py`
- The worker keeps doing its own pin validation and descriptor storage **first** — that is the security boundary and is unchanged — then forwards to `self._adapter.begin_rescue(params)` and returns the adapter's result.
- An adapter that cannot accept the handoff fails **loudly** via `AdapterRescueUnsupported`, rather than returning a clean-looking `Ack` for a resource nothing can release.

`local_operator/evaluation/adapters/api.py`
- New `RescuableAdapter` protocol.

### Two design decisions worth review

**1. A separate protocol, not a tenth method on `EvaluationAdapter`.** `EvaluationAdapter` is `runtime_checkable` and gates *every* handshake in `load_selected_adapter`. Adding `begin_rescue` there refused otherwise-valid adapters at `hello` and broke ordinary episodes — caught by the pre-existing `test_supervisor_launch_completes_real_handshake_and_reaps`, which went red. Rescue is optional at load time and mandatory only once a rescue is actually requested, which is the shape of a second protocol checked at that moment.

**2. A named `invalid_state` error, not `RpcProtocolError`.** A protocol error tears the channel down without a reply, so the parent saw only an opaque `TimeoutError` and the operator lost the reason — indistinguishable from a slow worker. The message is a fixed worker-authored string with no adapter-supplied content, so naming it leaks nothing through the blanket redaction that guards every other adapter exception (that redaction is otherwise untouched).

The existing `provider is None` → `attempted` / `terminate-unconfirmed` fallback is deliberately **kept**: it remains correct for genuinely ambiguous cases.

## Evidence: the real stranded descriptor

Run against a **copy** of the real descriptor at `~/worktrees/osworld/runs/proof-20260902-081937/rescue/` (original left intact), with the fixed harness installed into the pinned worker venv, then restored to its shipped bytes (md5 verified).

Tag audit before — `scripts/osworld_tag_audit.py --region us-east-1`:
```
[]
```

Before the fix:
```json
[{"episode_id": "ep-6ea01a117eee", "complete": false,
  "codes": ["session-closed", "terminate-unconfirmed", "schedule-absent"],
  "error": null}]
```
`exit 1`, descriptor retained — forever.

After the fix:
```json
[{"episode_id": "ep-6ea01a117eee", "complete": true,
  "codes": ["session-closed", "instance-absent", "schedule-absent"],
  "error": null}]
```
`exit 0`, descriptor **discarded**.

Tag audit after:
```
[]
```
No AWS writes: `terminate_instances` is reached only when the tag query returns live instances, and it returns zero.

## Tests: closing the blindness that hid this

Every previous rescue test called `adapter.begin_rescue` **directly**, bypassing the RPC layer that never forwarded it — so a fully green suite asserted adapter behaviour production never reached.

Two **real spawned-worker** tests (genuine interpreter, real installed distribution with a hashed RECORD, real supervisor) now cover it in `tests/unit/evaluation/adapters/test_launch.py`:
- `test_spawned_rescue_worker_builds_a_provider_and_releases` — a rescue worker builds a teardown provider from the delivered secret and releases. It asserts the **provider-only evidence code** (`released-with-<token>`), which the `None` fallback cannot produce, so it cannot pass by coincidence.
- `test_spawned_rescue_worker_refuses_an_adapter_without_begin_rescue` — refusal arrives by name, not as a timeout.

**Mutation-proved:** reverting the forwarding line to the original `return AckResult()` reproduces the production defect exactly (`complete=False`) and fails the new test. Restored after.

## Audit for the same blindness elsewhere

`_dispatch` self-handled exactly two methods: `hello` (correct — it *is* worker-level: loads the adapter and builds the handshake) and `begin_rescue` (the defect). Every other method already forwards via `getattr(self._adapter, method)`. No other inert path found.

## Operational note

⚠️ The pinned worker venv at `~/worktrees/osworld/venvs/0.1.0` ships `local_operator` **0.44.31** and **must be rebuilt before any pilot** — the worker is what carries this fix. A harness-side release alone does not reach it.

## Gates

- `flake8` clean; `black==26.1.0 --check` clean (803 files); `isort==5.13.2 --check` clean
- `pyright` — **0 errors, 0 warnings** (adapter wheel confirmed uninstalled)
- `tests/unit/evaluation` — 704 passed
- `tests/unit/evaluation/adapters` — 243 passed, **3× parallel + serial** (no flakiness)
- Full `tests/unit` — 10885 passed, 5 failed

The 5 failures are all in `tests/unit/tui/test_ask_picker.py` and are **pre-existing and unrelated** — reproduced identically on the stashed baseline; this PR touches no TUI code.

### DIFF
--- a/local_operator/evaluation/adapters/api.py
+++ b/local_operator/evaluation/adapters/api.py
@@ -692,6 +692,28 @@ class EvaluationAdapter(Protocol):
     async def close(self, params: CloseParams) -> AckResult: ...
 
 
+@runtime_checkable
+class RescuableAdapter(Protocol):
+    """An adapter that can accept the rescue handoff.
+
+    Deliberately SEPARATE from ``EvaluationAdapter`` rather than a tenth method
+    on it. ``EvaluationAdapter`` is ``runtime_checkable`` and gates every
+    handshake in ``load_selected_adapter``, so adding ``begin_rescue`` there
+    would refuse an otherwise valid adapter at hello -- breaking ordinary
+    episodes for a capability only ever used during teardown, and doing so on
+    an existing installed wheel. Rescue is genuinely optional at load time and
+    mandatory only once a rescue is actually requested, which is exactly the
+    shape of a second protocol the worker checks at that moment.
+
+    ``begin_rescue`` is the only call carrying the descriptor's infra values
+    together with freshly resolved secrets, so it is the sole opportunity a
+    rescue worker (which enters at HANDSHAKEN and never runs prepare or
+    reset_start) has to build a teardown provider.
+    """
+
+    async def begin_rescue(self, params: BeginRescueParams) -> AckResult: ...
+
+
 def observation_content_id(observation: Observation) -> Digest:
     """Identify adapter content while excluding its self-asserted identifier."""
 
--- a/local_operator/evaluation/adapters/worker.py
+++ b/local_operator/evaluation/adapters/worker.py
@@ -15,7 +15,6 @@ from local_operator.evaluation.adapters.api import (
     METHOD_STATES,
     PARAM_MODELS,
     RESULT_MODELS,
-    AckResult,
     AdapterMethod,
     AdapterSelector,
     AdapterState,
@@ -25,6 +24,7 @@ from local_operator.evaluation.adapters.api import (
     Handshake,
     HelloParams,
     PythonRuntime,
+    RescuableAdapter,
     RescueDescriptor,
     canonical_params_digest,
 )
@@ -55,6 +55,17 @@ MAX_REPLAY = 128
 # records are never evicted because losing one would turn a retry into a second
 # side effect; bounded exhaustion poisons before dispatch instead.
 MAX_OPERATION_RECORDS = 4096
+
+
+class AdapterRescueUnsupported(RuntimeError):
+    """The selected adapter cannot accept the rescue handoff.
+
+    Its own type (rather than a bare RuntimeError) so the worker's generic
+    adapter-error path reports it as a normal, replied-to failure instead of
+    a channel-killing protocol error: the parent must receive an answer it can
+    attribute, because a rescue that silently timed out is indistinguishable
+    from a slow one and hides the fact that nothing can be torn down.
+    """
 
 
 @dataclass(frozen=True)
@@ -236,6 +247,24 @@ class Worker:
             return
         except RpcProtocolError:
             raise
+        except AdapterRescueUnsupported:
+            # The ONE adapter-side failure whose reason crosses the wire. The
+            # blanket redaction below exists because adapter exceptions may
+            # carry reprs, paths, or environment values; this message is a
+            # fixed string raised by the worker itself with no adapter-supplied
+            # content, so naming it leaks nothing. It has to be named: a rescue
+            # that cannot start must tell the operator WHY, or a descriptor
+            # that can never be torn down looks like a transient failure.
+            response = self._write_error(
+                request,
+                "invalid_state",
+                "adapter does not implement begin_rescue",
+                digest=digest,
+                operation_id=operation_id,
+            )
+            if rescue_action_id is not None:
+                self._record_rescue_action(rescue_action_id, params, request.method, response)
+            return
         except (AdapterDiscoveryError, Exception):
             # Adapter exceptions may contain reprs, paths, environment values, or
             # tracebacks.  The wire exposes only a closed code and fixed text.
@@ -332,11 +361,40 @@ class Worker:
                 != canonical_digest("adapter-rescue-handshake-v1", self._handshake)
             ):
                 raise RpcProtocolError("rescue pins differ from the exact handshake")
-            # This transition grants cleanup routing only. Adapter code is not
-            # invoked, so it cannot create resources or resolve persisted refs.
+            # The pin checks above are the security boundary and run FIRST:
+            # nothing below may observe a descriptor whose selector/handshake
+            # digests did not match the exact worker this process handshook as.
             self._rescue_descriptor = params.descriptor
             self._rescue_actions.clear()
-            return AckResult()
+            # The adapter MUST see begin_rescue. It is the only call that
+            # carries the descriptor's infra_values plus the freshly resolved
+            # ``params.secrets``, and therefore the only opportunity a rescue
+            # worker has to build a teardown provider -- it never ran prepare
+            # or reset_start, so it holds no credential from any other source.
+            # Storing the descriptor and returning Ack without forwarding left
+            # the adapter's provider None, so every cleanup action took the
+            # "could not look" branch and reported attempted/terminate-
+            # unconfirmed. A sweep then never confirmed teardown and never
+            # discarded the descriptor -- and, far worse, a genuinely leaked
+            # instance was never terminated, because no provider existed to
+            # terminate it. Observed against episode ep-6ea01a117eee.
+            assert self._adapter is not None
+            if not isinstance(self._adapter, RescuableAdapter):
+                # Loud, not silent. An adapter that cannot accept the rescue
+                # handoff cannot tear anything down, and returning a clean Ack
+                # here would let the sweep report an orderly "attempted" for a
+                # resource nobody can release. Checked HERE rather than at
+                # load: rescue is optional for an ordinary episode and
+                # required only once one is actually requested.
+                #
+                # Raised as an ADAPTER error, not RpcProtocolError: a protocol
+                # error tears the channel down without a reply, so the parent
+                # only ever sees an opaque TimeoutError and the operator loses
+                # the reason. This path must name itself all the way out to the
+                # sweep entry, which is the whole point of failing loudly.
+                raise AdapterRescueUnsupported("adapter does not implement begin_rescue")
+            result = await self._adapter.begin_rescue(params)
+            return cast(ProtocolModel, result)
         assert self._adapter is not None
         handler = getattr(self._adapter, method)
         result = await handler(params)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "local-operator"
-version = "0.44.40"
+version = "0.44.41"
 description = "A team of proactive AI assistants that can work together behind the scenes to help you get mundane tasks done so you can focus on the fun stuff."
 readme = "README.md"
 authors = [{ name = "Damian Tran", email = "damian@gominerva.com" }]
--- a/tests/unit/evaluation/adapters/test_launch.py
+++ b/tests/unit/evaluation/adapters/test_launch.py
@@ -18,6 +18,7 @@ import json
 import os
 import sys
 from pathlib import Path
+from typing import Any
 
 import pytest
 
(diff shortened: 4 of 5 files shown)


damianvtran:
### Agent review — round 1

**Reviewer:** independent `reviewer` subagent (Claude Opus 4.6) — not the implementer
**Scope:** `01a6d348..26971b9a` (branch-own range; PR #543 now merged as `580922bd`, and the net two-dot diff `origin/main..26971b9a` is *exactly* the five declared files, so #543 is fully absorbed with no revert risk)

Reviewed at full SHA `26971b9a0a3344a86916b5d9b19df13691c8c670` in a clean detached worktree with its own venv. Everything below was **reproduced**, not inferred from the diff or from the PR body.

---

#### The core claim holds, and it is as serious as stated

I drove a **real spawned rescue worker** (real interpreter, real installed distribution with hashed RECORD, real supervisor, real `run_rescue`) with my own probe rather than the PR's tests:

```
complete: True
codes: ['released-with-CANARY-7f3a91c2-teardown-secret']
statuses: ['succeeded']
```

The evidence code is provider-only and carries the secret the adapter received on `begin_rescue`; the `provider is None` fallback can only emit `attempted`/`terminate-unconfirmed`. It cannot be produced by coincidence.

**Mutation proof reproduced independently.** Restoring `return AckResult()` ahead of the forwarding line reproduces the production defect exactly:

```
FAILED test_spawned_rescue_worker_builds_a_provider_and_releases
  - assert False +  where False = RescueAggregate(..., complete=False)
FAILED test_spawned_rescue_worker_refuses_an_adapter_without_begin_rescue
  - assert 'begin_rescue' in 'adapter_error: adapter operation failed'
```

Worker restored to shipped bytes afterwards (`git status --porcelain` → 0 lines).

**The dead code was real in production.** The pinned worker venv's `OSWorldV2Adapter.begin_rescue` exists, is a coroutine, and builds `AwsProvider` from `params.secrets` (`adapter.py:689-709`) — it was genuinely unreachable. `venvs/0.1.0` ships `local-operator 0.44.31` and has no `RescuableAdapter`, confirming the PR's operational warning: **the pinned worker venv must be rebuilt before any pilot, or this fix does not reach the process that needs it.**

#### Secrets on the rescue path (no leak found)

Canary `CANARY-7f3a91c2-teardown-secret` delivered through `run_rescue`, then every surface checked: **`LEAKS: NONE`**. Worker env (spawn argv is `-I -s -E -B`, so nothing is inherited), argv, persisted descriptor, `rescue.json`, every file under the run root, receipts, aggregate repr, and the supervisor's diagnostic tail (`CANARY in diagnostics: False`). Confirmed `secrets` is **not** a field on `PrepareParams` and **not** on `RescueDescriptor`; the real stranded descriptor on disk carries `secret_refs` only (`AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`), no values.

#### Security boundary unchanged

Static ordering inside `_dispatch` verified programmatically: `pins-check < descriptor-store < capability-check < adapter-call` → `True`. `RpcProtocolError("rescue pins differ from the exact handshake")` still precedes every line the change added, so a forged/mismatched descriptor is refused before the adapter is touched. Protocol faults still set `POISONED` and exit 70 (`worker.py:134-137`) — untouched.

#### The refusal path fails by name, and the redaction claim checks out

Through the **sweep**, as an operator sees it:

```
episode: episode | complete: False | codes: ()
error: rescue failed: invalid_state: adapter does not implement begin_rescue
descriptor still present (NOT discarded): True
```

Named, not a `TimeoutError`, and the inbox is correctly retained. I verified the redaction claim rather than accepting it: the message is a **fixed literal** with no `str(error)`/`repr()`/f-string interpolation, so no adapter content crosses the wire. `invalid_state` is in the closed `RpcError` code Literal.

#### The protocol split is correct

`isinstance` semantics confirmed: `Tiny` (no `begin_rescue`) is `EvaluationAdapter` **True** / `RescuableAdapter` **False**. I built a *functional* adapter lacking `begin_rescue` (the PR's `TinyAdapter` raises `NotImplementedError` everywhere, which cannot demonstrate this) and ran a real spawned episode:

```
HANDSHAKE OK (no begin_rescue): norescue-e2e
ORDINARY EPISODE CALL OK: ['SOME_TOKEN']
load gate mentions RescuableAdapter: False
```

`load_selected_adapter`'s gate is unchanged for the normal path. The separate-protocol decision is right.

#### The audit claim — verified independently, and it is complete

Rather than trust the body, I enumerated `PARAM_MODELS` against `_dispatch`:

```
SELF-HANDLED: ['hello', 'begin_rescue']
FORWARDED WHOLE (getattr path): ['ask_user_exchange', 'cleanup', 'close', 'execute',
  'inspect_requirements', 'observe', 'prepare', 'reset_start', 'score']
hello: fields=['selector'] — referenced: ['selector'], NOT referenced: []
begin_rescue: whole params object forwarded to adapter: True
```

Exactly two self-handled methods, both accounted for. Every other method forwards the entire params object, so no field can be dropped the way `secrets` was. **No further inert paths found** — the audit claim is accurate.

#### Real descriptor / AWS (read-only, no writes)

Tag audit reproduced independently: `describe-instances` for `lop-ep-ep-6ea01a117eee` → `[]`; instance `i-0a76923056da31d04` → `[]` (empty, no `InvalidInstanceID.NotFound` raised, exactly as the body says). A broader sweep for `lop-ep-*` in `pending/running/stopping/stopped` → `[]`. Nothing is currently leaking, and the body's before/after triple is consistent with the `provider is None` fingerprint.

---

### Findings

**No blocker or major findings.**

**MINOR 1 — a refused `begin_rescue` leaves the stored descriptor behind in a still-usable worker.**
`worker.py:365-367` stores `self._rescue_descriptor` and clears `_rescue_actions` *before* the `RescuableAdapter` check at `:383`, so a refusal returns an error while the worker keeps rescue-mode state and stays in `HANDSHAKEN`. Not exploitable today: `cleanup` is illegal in `HANDSHAKEN`, `run_rescue` aborts on the raised `RpcRemoteError`, and reaching `cleanup` would need `inspect_requirements`→`prepare`, which pin validation already gated. But the descriptor-store is only *safe* because of a state-table coincidence rather than by construction. *Fix (one line): move the `isinstance` check above the two assignments.*

**MINOR 2 — `AdapterRescueUnsupported` is caught in `_handle` for every method, not just `begin_rescue`.**
`worker.py:250` is unguarded by any method check, so if adapter code ever raised this exception type from `cleanup` or `close`, the operator would get the fixed message `"adapter does not implement begin_rescue"` — misattributed. Only the worker raises it today, so this is latent. *Fix: guard with `request.method == "begin_rescue"`, or re-raise otherwise.*

**MINOR 3 — `runtime_checkable` Protocol checks attribute presence, not callability.**
Verified: an adapter with `begin_rescue = "not-callable"` passes `isinstance(..., RescuableAdapter)` → **True**, and would then fail inside the `await` as a redacted `adapter_error`, losing the clear reason this PR exists to provide. This is a documented `typing` limitation and matches how `EvaluationAdapter` is already gated, so it is consistent rather than a new hole. Worth a sentence in the `RescuableAdapter` docstring.

**NIT 1 — the PR body's test-count and failure figures do not match this tree.** Body says `tests/unit/evaluation/adapters` = 243 and full suite = "10885 passed, 5 failed". I measure **246** and **11029 passed, 15 skipped, 0 failed**. The 5 `tui/test_ask_picker.py` failures are **not reproducible** here — 53/53 pass on head *and* on the `01a6d348` baseline, and the full suite is entirely green. The body's numbers appear stale rather than wrong-in-kind; no action needed beyond noting the suite is cleaner than claimed.

---

### Gates (re-run by me, whole tree, exactly as CI)

| Gate | Result |
|---|---|
| `flake8` | clean (rc=0) |
| `black==26.1.0 --check` | clean — 809 files unchanged |
| `isort==5.13.2 --check` | clean |
| `pyright` | **0 errors, 0 warnings, 0 informations** (adapter wheel confirmed uninstalled) |
| `tests/unit/evaluation` | **707 passed** |
| `tests/unit/evaluation/adapters` | **246 passed** — 3× parallel (36s/31s/31s) + serial (106s), no flakiness |
| `tests/unit` (full) | **11029 passed, 15 skipped, 0 failed** (668s) |

**Version:** `0.44.40`, exactly one `chore(release)` commit at tip. Re-checked at the END of review: `origin/main` is `580922bd` at `0.44.39`, PyPI latest `0.44.39`, `0.44.40` not published. **Strictly above both — no collision.**

---

### Verdict: `clean` — **TERMINAL**

**Blockers: 0. Majors: 0.** The three minors and one nit are follow-ups and are **not** grounds for another round.

This fixes a structurally inert leak safety net: the provider that terminates a paid EC2 instance was never constructed, so a genuinely leaked instance could not have been reaped by any sweep. The fix is correct, minimal, forwards the one credential-bearing call it must, preserves the pin-validation security boundary and the redaction guarantee, and is proved by real spawned-worker tests that fail when mutated. Weighed by the risk that matters — an unreapable cloud instance — this change strictly reduces it and introduces no new failure mode I could reproduce.

**Recommend merge**, with one operational caveat that is not a code finding: the pinned worker venv (`venvs/0.1.0`, currently `local-operator 0.44.31`) **must be rebuilt before the next pilot**, or the rescue path in the process that actually runs it remains inert.

*No edits, no merge, no release, and no AWS writes were made during this review.*

damianvtran:
Rebuilt on `main` @ `580922bd` now that #543 has merged (squash-merged, so its commits could not be replayed by rebase — this branch's own commit was cherry-picked onto main instead). New head `82d71cd6`, version **0.44.40** (strictly above main 0.44.39 and PyPI 0.44.39).

Round 1 carries forward: the branch-own patch excluding `pyproject.toml` is **byte-identical** to the reviewed head, and `git diff --name-only origin/main...HEAD` lists only the five declared files.

The previous run's `test (3.12)` failure was `tests/unit/session/test_resume_subagents_todos.py:71` — a timing-based wait in code this PR does not touch. Reproduced on this machine at ~1-in-3 running that file alone, so it is a pre-existing flake, not a regression here. It is being fixed properly in its own PR rather than re-run away. Manager-performed.

PR #551, opened 2026-09-02
docs(guides): steer agents to the send tool in peer-messaging

damianvtran:
## Summary

`guide://peer-messaging` was steering agents the wrong way. The `lop send` shell command had **more space and more worked examples** than the `send` tool, so an agent reading the guide came away with shell recipes freshest in mind — directly undercutting the system prompt's own rule at `prompts_md/system.md:162` ("use the `send` tool for peer messaging — never shell out to `lop send`") and the rationale already written down at `tools/builtin.py:4621`: a shelled send buries the delivery in an opaque shell trace, and the CLI's mailbox default leaves an idle peer sitting on the note.

This rebalances the guide so the tool leads and carries the worked examples, and fixes two factual defects that were verified broken against real running sessions.

Docs-only: **no `.py` behaviour changed**. The guide is packaged data, so it ships in the wheel and is re-verified through the real discovery/resolver path below.

## What changed

- **The `send` tool now leads and dominates.** It carries four worked examples (quiet drop, wake-by-default, address-by-pid, mid-turn steer), each showing the call *and* the detail string it returns, plus a new "What the result tells you" section covering the refusals (ambiguous target, no match, self-send, oversize body, and the lost-ack case that must not be retried blindly).
- **An unmissable agent rule** in a blockquote near the top: *use the `send` tool; never `bash` a `lop send`*, with the three reasons stated briefly — auditable card, wake-on by default, write-tier approval — so the rule survives summarization instead of being a bare prohibition.
- **`lop send` is marked the HUMAN entry point.** The section stays (it is a real product surface people use from a terminal) but says plainly that it is not an agent recipe.
- Parameter list reconciled with the real `SendParams`: `message` is required and non-empty, `wake` is ignored when `now=True`, `now` defaults to false, and the `pid` → `session` → `target` addressing precedence is stated.

## Defect 1 — a `CWD` column that does not exist

**Before** (line ~78): listed `PID KIND CONVERSATION MODEL CWD` as the table's identity columns.

**After:** the guide states the real header and says where `cwd` actually lives. Measured against a live table (`cli.py:1265-1268`):

```
live      68985 tui     781b501cbcc6             test/e2e                    53.5M    143.0M      13s     13s
```

No `CWD`. `cwd` and `session_id` are `--json`-only — confirmed by the actual key set:

```
['conversation_name', 'cwd', 'footprint_bytes', 'heartbeat_age_s', 'kind',
 'model_label', 'pid', 'rss_bytes', 'session_id', 'state', 'uptime_s']
```

This is not a cosmetic fix: `target` matches against the **cwd basename**, so `--json` is exactly where a reader must look to predict what a substring will match. The guide now says so.

## Defect 2 — a documented example that delivers nothing

**Before** (line ~157), presented as *the* "wake an idle session" example:

```
lop send --pid 12345 --wake "the deploy finished; verify prod"
```

Run verbatim against a live session it exits **1** and delivers nothing:

```
$ lop send --pid 69005 --wake 'the deploy finished; verify prod'
no message given (pass it as an argument or pipe it on stdin)
rc=1
```

**Root cause** (a real CLI grammar defect, `cli.py:438-446`): `send` declares `target` and `message` as two `nargs="?"` positionals. With `--pid` supplied the single positional binds to **`target`**, `message` stays `None`, the body falls through to the stdin branch, finds a terminal, and exits. It is broader than the guide's one line — `lop send --pid N "body"` and `lop send --session ID "body"` fail the same way.

**Fixing the parser is out of scope for this docs change.** Instead the example is replaced with a form I ran and watched deliver, keeping the "wake an idle session by exact pid" case represented:

```
$ echo 'the deploy finished; verify prod' | lop send --pid 69005 --wake
→ 7c44b23dc86f (pid 69005): delivered and woke the session
rc=0
```

and the constraint itself is now documented under "Choosing a target", with the working and failing forms side by side, so the next reader does not rediscover it the hard way.

## Testing evidence

**Isolation first.** The operator has ~16 live sessions doing real work on this host. Every session used here was **spawned by me** under `LOCAL_OPERATOR_CONFIG_DIR=/tmp/lop-gs/cfg` with `--hosting test` (no network, no real model), so the discovery records, `lop sessions` and `lop send` could only ever see and reach my own. **Nothing was delivered to, woken, or steered in any session I did not start.** All three receivers were stopped afterwards; the operator's 13 live sessions were untouched.

**Every command in the rewritten guide, run verbatim**, against receivers whose cwd basenames match the example targets:

| Command | Result | rc |
|---|---|---|
| `lop sessions` | table printed | 0 |
| `lop sessions --json` | 11 keys incl. `cwd`, `session_id` | 0 |
| `lop send "peer-send design" "gates are green, ready for review"` | `delivered to the mailbox (will be read on the next turn)` | 0 |
| `lop send "release cutter" --wake "the deploy finished; verify prod"` | `delivered and woke the session` | 0 |
| `lop send "ingest refactor" --now "hold off — the schema changed"` | `delivered (opened a turn)` | 0 |
| `lop send "peer-send design" < note.txt` | `delivered to the mailbox …` | 0 |
| `git log -1 --stat \| lop send "release cutter"` | `delivered to the mailbox …` | 0 |
| `echo "…" \| lop send --pid N --wake` (the fix) | `delivered and woke the session` | 0 |
| `lop send --pid N --wake "…"` (documented as broken) | `no message given` | 1 |

**Delivery confirmed at the RECEIVER, not just the sender's ack** — reading each receiver's own `transcript.jsonl` for `custom_type: "peer_message"` rows:

```
=== 781b501cbcc6 — 2 peer_message rows
    gates are green, ready for review
    note from a file
=== 7c44b23dc86f — 3 peer_message rows
    the deploy finished; verify prod
    commit 37289774…            (the piped git-log body)
    the deploy finished; verify prod   (the --pid stdin fix)
=== 915843efc78c — 1 peer_message rows
    hold off — the schema changed
```

The TUI card was checked in the receivers' pty output too: `↔ peer message from (pid …)`, matching `transcript.py:1715`.

**The `send` tool itself was exercised** (`execute_send` in-process against the same isolated receivers), which is what backs the rewritten examples rather than just the CLI:

```
default (wake ON)   -> delivered and woke the session
wake=False quiet    -> delivered to the mailbox (will be read on the next turn)
now=True steer      -> delivered (opened a turn)
by session id       -> delivered and woke the session
ambiguous substring -> 2 sessions match; retry with pid=<n>:  pid=64190 … pid=64209 …
no match            -> no live session matches 'zzz-nothing'
```

**All four delivery detail strings verified against both branches.** The fourth (`delivered mid-turn (steered)`) only fires when the peer is genuinely streaming, which the `test` provider never holds long enough to hit from outside — so a real `Session` was parked mid-turn on a stalled stream and the real receive half driven directly:

```
peer _is_streaming = True
now=True  while BUSY -> delivered mid-turn (steered)
wake=True while BUSY -> delivered to the mailbox (will be read on the next turn)
peer _is_streaming = False
now=True  while IDLE -> delivered (opened a turn)
```

All four match the literals at `session/session.py:3996,4002,4010,4039`. The 256 KB cap claim was checked too: 300 KB → `message is too large (300001 bytes); cap is 262144 bytes`; 200 KB delivers.

**Guide integrity.** Front matter parses, fences balanced (18), and it still resolves through the real runtime path:

```
resolved chars: 12088
agent rule present: True
stale CWD column claim gone: True
worked tool examples present: 4
```

The catalog test caught a real regression mid-work: my first rewritten `description:` was 201 chars against the ≤180 prompt-size cap in `test_packaged_catalog_is_small_and_descriptions_are_prompt_sized`. Tightened to 171 and green.

## Consistency sweep

Grepped for other places telling an agent to shell out for peer messaging:

- `prompts_md/system.md:162` already carries the correct rule and **agrees** with this rewrite — left alone as instructed.
- `docs/design/peer-send.md` is a historical design document — **not touched**.
- `README.md` is human-facing prose, and all four of its `lop send` examples use the positional form, which I verified works. No agent-facing contradiction, so left alone.

No other agent-facing instruction contradicts the rule.

## Gates

| Gate | Result |
|---|---|
| `flake8 .` | clean (rc=0) |
| `black --check .` (26.1.0) | 808 files unchanged |
| `isort --check .` (5.13.2) | clean (rc=0) |
| `pyright` (whole tree) | **0 errors, 0 warnings, 0 informations** |
| `tests/unit/guides`, `tests/unit/skills` | 219 passed |
| peer-messaging suites (`test_send_tool`, `test_peer_send`, `test_peer_client`, `test_cli_send`, `test_session_peer`) | 56 passed |

No `.py` file changed (the diff is the guide plus the version bump), so per AGENTS.md I ran the targeted subsets above rather than the ~15-minute full suite on this contended host, and CI covers the rest.

Version bumped one patch above `origin/main` (0.44.38) → **0.44.39**; confirmed 404 on PyPI and not an existing tag at branch time.

## Not addressed (deliberately)

- **The `lop send` positional-grammar defect itself.** `--pid`/`--session` with an inline body binds to `target` and exits 1. Out of scope for a docs change, so it is documented rather than fixed; the parser fix wants its own PR (likely making `message` the only positional when an explicit selector is present, which is a user-visible CLI change deserving its own review).

### DIFF
--- a/local_operator/guides/peer-messaging/GUIDE.md
+++ b/local_operator/guides/peer-messaging/GUIDE.md
@@ -1,6 +1,6 @@
 ---
 name: peer-messaging
-description: Message another local lop session — the in-session `send` tool for agents, `lop send` for humans — and list running sessions with `lop sessions`. No cmux needed.
+description: Message another local lop session — agents use the `send` tool (never a shelled `lop send`), humans use `lop send` — and list sessions with `lop sessions`. No cmux needed.
 ---
 
 # Peer messaging between lop sessions
@@ -15,22 +15,32 @@ Two entry points, one wire:
 
 - **The `send` tool** — the way an AGENT messages a peer from inside its own
   session (below). Wake defaults ON, so an idle peer responds right away.
-- **`lop send`** — the shell command for HUMANS (further below). Mailbox by
-  default; `--wake` opts in to waking.
+- **`lop send`** — the shell command for HUMANS at a terminal (further below).
+  Mailbox by default; `--wake` opts in to waking.
 
 They share the same target resolution and delivery semantics; the only
 difference is the default and who runs them.
+
+> **Agents: use the `send` tool. Never `bash` a `lop send`.** Three reasons,
+> and they are why the rule is worth remembering rather than looking up. The
+> tool call is its own **auditable card** — a delivery buried in a shell trace
+> is hard to review and easy to miss in an approval prompt. The tool **wakes by
+> default**, where the CLI's mailbox default leaves an idle peer sitting on the
+> note until something else wakes it. And the tool prompts at the **write
+> tier**, like every other capability that can start autonomous work in another
+> process. The shell command below is documented for the human at a terminal,
+> not as an agent recipe.
 
 **Trust boundary is the account.** The discovery record is mode `0600` under a
 `0700` directory, so anything that can read a session's control key is already
 the owning user. Delivery is loopback only. There is no cross-account path; the
 same-account boundary is the whole authorization story.
 
-## The `send` tool — agent-facing, in-session
+## The `send` tool — the agent entry point
 
-An agent messages a peer with the `send` tool, not by shelling out to `lop
-send`: the tool call is its own auditable card, and its defaults are tuned for
-agent-to-agent hand-offs.
+This is how an agent messages a peer. It runs inside the session process, so it
+resolves the target, names the delivery in its own card, and returns the
+receive side's own detail string.
 
 Parameters:
 
@@ -39,17 +49,58 @@ Parameters:
 - `pid` — exact pid (unambiguous; the disambiguation error lists these).
 - `session` — exact session id.
 - `message` — the body; it lands in the peer's transcript as an inbound
-  cross-session card.
+  cross-session card. Required, and non-empty.
 - `wake` (default **true**) — mailbox mode: wake an idle peer so it responds
   right away. `wake=False` is the quiet drop: the peer reads it on its next
-  turn and stays idle (the TUI card shows this mode as `quiet`).
-- `now` — steer the peer mid-turn instead of using the mailbox; opens a turn
-  if the peer is idle.
+  turn and stays idle (the TUI card shows this mode as `quiet`). Ignored when
+  `now=True`.
+- `now` (default **false**) — steer the peer mid-turn instead of using the
+  mailbox; opens a turn if the peer is idle.
 
-**Agent sends wake idle targets by default.** A peer parked on a scheduled
-wake or an idle turn answers the moment the message lands — there is no
-"it will notice next time it wakes" gap. Pass `wake=False` deliberately for a
-non-urgent hand-off the peer may fold into whatever it does next.
+Addressing precedence is `pid`, then `session`, then the `target` substring —
+supply one. `pid` is the form to reach for when a substring could match more
+than one session.
+
+### Worked examples
+
+A non-urgent hand-off the peer folds into whatever it does next — it stays
+idle and reads this on its next turn:
+
+```
+send(target="peer-send design", message="gates are green, ready for review", wake=False)
+→ delivered to the mailbox (will be read on the next turn)
+```
+
+Something the peer should act on now. This is the default, so `wake` needs no
+spelling out:
+
+```
+send(target="release cutter", message="the deploy finished; verify prod")
+→ delivered and woke the session
+```
+
+The same, addressed unambiguously by pid — the form the disambiguation error
+hands you:
+
+```
+send(pid=64190, message="the deploy finished; verify prod")
+→ delivered and woke the session
+```
+
+Redirect a peer that is actively working, before it goes further down a wrong
+path:
+
+```
+send(target="ingest refactor", message="hold off — the schema changed", now=True)
+→ delivered mid-turn (steered)
+```
+
+**Sends wake idle targets by default.** A peer parked on a scheduled wake or
+an idle turn answers the moment the message lands — there is no "it will
+notice next time it wakes" gap. Pass `wake=False` deliberately when the note
+is not urgent.
+
+### What the result tells you
 
 The result echoes the receive side's own detail string, so the sender knows
 exactly how the peer took it:
@@ -60,8 +111,16 @@ exactly how the peer took it:
 - `delivered mid-turn (steered)` — `now=True` while the peer was working.
 - `delivered (opened a turn)` — `now=True` while the peer was idle.
 
-A session cannot send to itself (the tool refuses), and an ambiguous `target`
-returns the candidate list asking for a `pid`.
+Refusals are answers too, and each names its own fix:
+
+- An ambiguous `target` returns the candidate list — `2 sessions match; retry
+  with pid=<n>:` followed by one `pid=<n>` line per session.
+- No match returns `no live session matches '<target>'`.
+- A session cannot send to itself; the tool refuses before it dials.
+- A body over the 256 KB cap is rejected with the measured size, not truncated.
+- If the ack is lost after the peer committed the message, the tool says the
+  message **may or may not** have arrived rather than claiming failure — check
+  with the peer instead of resending, or it lands twice.
 
 ## `lop sessions` — what is running and what it costs
 
@@ -70,13 +129,15 @@ lop sessions
 lop sessions --json
 ```
 
-Columns:
+The table prints `STATE PID KIND CONVERSATION MODEL RSS FOOTPRINT UPTIME
+HB_AGE`:
 
 - `STATE` — `live` (pid alive and heartbeating), `wedged` (pid alive but its
   owner has not heartbeat within the timeout, so it will not service the socket
   promptly), or `stale` (dead; the record is reaped on the next scan).
-- `PID` `KIND` `CONVERSATION` `MODEL` `CWD` — session identity. `KIND` is `tui`
+- `PID` `KIND` `CONVERSATION` `MODEL` — session identity. `KIND` is `tui`
   (interactive), `daemon` (daemon-owned), or `exec` (headless one-shot).
+  `CONVERSATION` falls back to the session id when the session has no name yet.
 - `RSS` — resident set size. Always available; the baseline number.
 - `FOOTPRINT` — the *true* memory cost. On macOS RSS under-reports because
   memory is compressed and swapped, so this column shows the phys footprint
@@ -87,14 +148,29 @@ Columns:
 - `HB_AGE` — how long since its last heartbeat. A large `HB_AGE` on a `live`
   row is an early sign of a session going wedged.
 
-`--json` emits the same fields as a list of objects (bytes, not human sizes)
-for scripting.
+**The table does not show a session's `cwd`.** `cwd` and `session_id` are
+`--json`-only fields — which matters because `target` matches against the cwd
+basename, so `--json` is where you look to see what a substring will match:
 
-## `lop send` — the shell command (humans)
+```
+lop sessions --json
+```
 
-The same wire, driven from a terminal. Mailbox by default — a human sending
-from a shell usually wants the non-interrupting drop; `--wake` opts in to
-waking an idle session.
+It emits one object per session with `state`, `pid`, `kind`,
+`conversation_name`, `session_id`, `model_label`, `cwd`, `rss_bytes`,
+`footprint_bytes`, `uptime_s` and `heartbeat_age_s` (bytes and seconds, not
+human-formatted sizes) for scripting.
+
+## `lop send` — the HUMAN entry point
+
+The same wire, driven from a terminal by a person. Mailbox by default — a human
+sending from a shell usually wants the non-interrupting drop; `--wake` opts in
+to waking an idle session.
+
+**This section is not an agent recipe.** An agent inside a session uses the
+`send` tool above; shelling out to this command buries the delivery in a shell
+trace and takes the mailbox default, which leaves an idle peer sitting on the
+message.
 
 ```
 lop send "<target>" "your message"
@@ -137,6 +213,20 @@ sessions, `lop send` prints the candidates and exits non-zero asking you to
 disambiguate with `--pid`. If the only match is `wedged`, it says so rather
 than hanging on a dial.
 
+**With `--pid` or `--session`, pipe the body — do not pass it as an argument.**
+`send` declares `target` and `message` as two optional positionals, so a single
+positional alongside `--pid`/`--session` binds to `target`, leaving `message`
+empty; the command then looks for a body on stdin, finds a terminal, and exits
+1 with `no message given`. Nothing is delivered. So:
+
+```
+echo "the deploy finished; verify prod" | lop send --pid 12345 --wake   # works
+lop send --pid 12345 --wake "the deploy finished; verify prod"          # exits 1
+```
+
+The positional form (`lop send "<target>" "message"`) is unaffected — both
+positionals are filled, so it is the form to prefer at a terminal.
+
 ### What the human sees
 
 The delivered message appears in the target's transcript marked
@@ -154,7 +244,11 @@ process and looks itself up by pid. Both name the same sender.
 lop send "peer-send design" "gates are green, ready for review"
 
 # Wake an idle session to act now
-lop send --pid 12345 --wake "the deploy finished; verify prod"
+lop send "release cutter" --wake "the deploy finished; verify prod"
+
+# Wake an idle session addressed by exact pid — body on stdin, per the
+# grammar note above
+echo "the deploy finished; verify prod" | lop send --pid 12345 --wake
 
 # Redirect a session mid-turn
 lop send "ingest refactor" --now "hold off — the schema changed"
--- a/local_operator/guides/peer-messaging/GUIDE.md
+++ b/local_operator/guides/peer-messaging/GUIDE.md
@@ -213,19 +213,31 @@ sessions, `lop send` prints the candidates and exits non-zero asking you to
 disambiguate with `--pid`. If the only match is `wedged`, it says so rather
 than hanging on a dial.
 
-**With `--pid` or `--session`, pipe the body — do not pass it as an argument.**
-`send` declares `target` and `message` as two optional positionals, so a single
-positional alongside `--pid`/`--session` binds to `target`, leaving `message`
-empty; the command then looks for a body on stdin, finds a terminal, and exits
-1 with `no message given`. Nothing is delivered. So:
+**With `--pid` or `--session`, pipe the body — a lone positional is read as the
+target, not the message.** `send` declares `target` and `message` as two
+optional positionals, so a *single* positional alongside `--pid`/`--session`
+binds to `target`, leaving `message` empty; the command then looks for a body
+on stdin, finds a terminal, and exits 1 with `no message given`. Nothing is
+delivered. Piping the body is the clean form:
 
 ```
 echo "the deploy finished; verify prod" | lop send --pid 12345 --wake   # works
+echo "gates are green" | lop send --session 7c44b23dc86f                # works
 lop send --pid 12345 --wake "the deploy finished; verify prod"          # exits 1
 ```
 
-The positional form (`lop send "<target>" "message"`) is unaffected — both
-positionals are filled, so it is the form to prefer at a terminal.
+A *second* positional does fill `message`, so a placeholder target delivers —
+the selector wins, because the resolver checks `--pid`/`--session` before it
+ever looks at `target`:
+
+```
+lop send ignored "the deploy finished; verify prod" --pid 12345         # works
+```
+
+That works, but it reads as though it addresses `ignored`; prefer the piped
+form. The plain positional form (`lop send "<target>" "message"`) is unaffected
+— both positionals are filled, so it stays the form to reach for at a
+terminal.
 
 ### What the human sees
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "local-operator"
-version = "0.44.39"
+version = "0.44.40"
 description = "A team of proactive AI assistants that can work together behind the scenes to help you get mundane tasks done so you can focus on the fun stuff."
 readme = "README.md"
 authors = [{ name = "Damian Tran", email = "damian@gominerva.com" }]


damianvtran:
### Agent review — round 1

Scope: 37289774..daeca217

Method: read the diff, checked every claim against source (`cli.py`, `tools/builtin.py`, `mobile/peer_send.py`, `session/session.py`), and **ran every command in the guide verbatim** against two receiver sessions I started myself in an isolated `LOCAL_OPERATOR_CONFIG_DIR=/tmp/rev551/cfg`. Delivery was confirmed **at the receiver** (a JSONL sink on the receive side), not by exit code. No session in the operator's registry was addressed, woken, or steered.

---

#### BLOCKER-1 — the version bump collides with an already-published release; the PR cannot be released as-is

`pyproject.toml` moves `0.44.38 → 0.44.39`, which was one patch above main **when the branch was cut**. `origin/main` has since advanced:

```
580922bd fix(benchmarks): route OSWorld episode cache outside the pinned workspace (#543)
37289774 chore(release): bump version to 0.44.38 (#549)   <- this PR's base
```

`580922bd` already took `0.44.39`, and it is **shipped**:

```
$ git show origin/main:pyproject.toml | grep -m1 version
version = "0.44.39"
$ gh release view v0.44.39 --json tagName --jq .tagName
v0.44.39
$ curl -s https://pypi.org/pypi/local-operator/json | ...
latest: 0.44.39   0.44.39 present: True   (uploaded 2026-09-02T19:35:45)
```

Git will not flag this — both sides changed the line to the *same* value, so the three-way merge is clean and `git merge-tree origin/main daeca217` resolves to `version = "0.44.39"`. The PR merges green and produces a main whose version is **identical to the previous commit's**. Then step 5 of the release process fails hard: `gh release create v0.44.39` collides with the existing tag/Release, and if it were forced through, PyPI rejects a re-upload of an existing version. The net effect is a merged change that can never reach the `lop` runtime under a new version.

Fix: rebase onto current `origin/main` and set `version = "0.44.40"` (free on PyPI — confirmed `0.44.40 present: False`). Materiality is right either way: a guide-content correction is a patch.

---

#### Both named defects: FIXED, and I reproduced the fix end to end

**Defect 1 — the `CWD` column.** Old line 78 listed `` `PID` `KIND` `CONVERSATION` `MODEL` `CWD` ``. Now corrected, and the diff goes further than asked by printing the real header and explaining where `cwd` actually lives. Verified against `cli.py:1265-1268` and against the live table:

```
```

The `--json` field list at L159-162 is exact — `lop sessions --json` returns precisely `conversation_name, cwd, footprint_bytes, heartbeat_age_s, kind, model_label, pid, rss_bytes, session_id, state, uptime_s`. Good call tying it to *why* an agent cares (`target` matches the cwd basename, so `--json` is where you see what a substring will match).

**Defect 2 — the `--pid` + positional example.** Both halves of the new L222-225 block are confirmed by execution, against my own receiver at pid 86021:

```
$ echo "the deploy finished; verify prod" | lop send --pid 86021 --wake
→ rev551 release cutter (pid 86021): delivered and woke the session     [exit=0]

$ lop send --pid 86021 --wake "the deploy finished; verify prod"
no message given (pass it as an argument or pipe it on stdin)           [exit=1]
```

Receiver-side sink confirms the first arrived and the second delivered nothing. The stated mechanism is also right at the parser: with `--pid` given, the lone positional binds to `target` and `message` stays `None` (`cli.py:438-447`, both `nargs="?"`), then `cli.py:1186-1192` falls to the stdin branch and exits 1.

---

#### Every other command in the guide: runs clean

All executed verbatim against my isolated receivers; every one delivered and the receive-side detail string matched what the guide documents.

| # | Command (guide line) | Result |
|---|---|---|
| 1 | `lop send "release cutter" --wake "…"` (L247) | `delivered and woke the session` |
| 2 | `lop send "peer-send design" "…"` shape (L244) | `delivered to the mailbox (will be read on the next turn)` |
| 3 | `lop send "ingest refactor" --now "…"` (L254) | `delivered mid-turn (steered)` |
| 4 | `lop send "<target>" < note.txt` (L180) | delivered, mailbox |
| 5 | `git log -1 --stat \| lop send "release cutter"` (L257) | delivered, mailbox |
| 6 | `lop sessions` / `lop sessions --json` (L128-130) | header + fields exactly as documented |

The four **worked examples** in the tool section were run through `execute_send` itself, and each returned the arrow line the guide shows:

```
send(target=…, message=…, wake=False)  -> delivered to the mailbox (will be read on the next turn)
send(target=…, message=…)              -> delivered and woke the session
send(pid=86021, message=…)             -> delivered and woke the session
send(target=…, message=…, now=True)    -> delivered mid-turn (steered)
```

The documented refusals are literal too: ambiguity returns `2 sessions match; retry with pid=<n>:` followed by `pid=<n>` lines (matches `builtin.py:4807`), no-match returns `no live session matches 'nope-xyz'` (`peer_send.py:129`), and a 262145-byte body is refused with `message is too large (262145 bytes); cap is 262144 bytes` — the guide's "rejected with the measured size, not truncated" is accurate.

#### Parameters, delivery strings, precedence: all match source

- `SendParams` (`builtin.py:4632-4662`) — every parameter, every default matches, including `wake` default **true** and `now` default **false** (the diff newly states `now`'s default and `message` being required/non-empty, both correct against `min_length=1`). "Ignored when `now=True`" matches the field's own docstring and behaviour: I ran `now=True, wake=False` and got `delivered mid-turn (steered)`.
- The four delivery detail strings are **byte-for-byte** identical to `session/session.py:3959, 3965, 3973, 4002` (checked by substring equality, not by eye).
- Addressing precedence `pid → session → target` matches the resolver's branch order (`peer_send.py:60-82`).
- The `wake` tier / auditability rationale in the new callout matches `builtin.py:4619-4629` and `4747-4749` (`approval_tier="write"`) accurately, including the "opaque shell trace" and "mailbox default leaves an idle peer sitting on the note" reasons. Nothing overstated.

#### No information was lost in the rebalance

I diffed the token sets of both revisions. Every unique token dropped is from prose that was rewritten, not from a fact: the removed set is only `(bytes,`, `(humans)`, `**Agent`, `` `CWD` ``, `` `pid`. ``, `agent-facing,`, `agent-to-agent`, `auditable`, `Columns:`, `fold`, `hand-offs.`, `in-session`, `next.`, `objects`, `refuses),`, `sends`, `tuned`. All eight facts named in the review brief survive verbatim — delivery modes (L188-200), targeting priority (L204-209), wedged behaviour (L213-214), the blocking-`wait` early return (L191-194, and it is real: `tests/unit/tools/test_wait_wakes_on_peer_message.py` asserts `interrupted_by == "peer_message"`), the limits list (L260-272), the 256 KB cap, and the sender-identity note (L232-238). The only deliberate deletion is the false `CWD` claim. `lop send` remains fully documented as the human entry point.

#### Front matter and packaging

`name:`/`description:` intact; the new description is 171 chars, inside the `40 <= len <= 180` bound `tests/unit/guides/test_guides.py:38` enforces. `tests/unit/guides` passes 22/22, and the guide loads through the real resolver at head. No `.py` file changed in the diff. CI is green except `test (3.12)` still pending.

#### Is the rule unmissable enough?

Yes. A skimming agent hits it three times on three different reading paths: the front-matter description (what the semantic router surfaces before the body is ever read), the blockquote callout at L24-32 before any command appears, and the `lop send` section's own "**This section is not an agent recipe**" at L170. The tool now leads with the section heading, the parameter table, and all four worked examples, while the shell section's examples sit below. That is the requested rebalance, achieved.

---

#### MINOR-1 — "do not pass it as an argument" is stricter than the actual grammar

`GUIDE.md:216` — with `--pid`/`--session`, a body *does* get delivered if a throwaway positional precedes it, because the resolver checks `pid` before it ever looks at `target`:

```
send ignored "real body" --pid 999  ->  target='ignored' message='real body' pid=999
```

The guide's advice is the right one to give (the two-positional form is a trap, and piping is cleaner), so this is not wrong as guidance — only slightly overstated as grammar. Optional one-line softening: "…a *single* positional alongside `--pid`/`--session` binds to `target`". Not worth a round on its own.

#### NIT-1 — the `--session` variant is described but not shown

L216 correctly covers `--pid` **or** `--session`, and I confirmed both behave identically (`lop send --session rev551-sess "…"` exits 1; `echo … | lop send --session rev551-sess` delivers). The example block and the Examples section only ever show `--pid`. Harmless; the rule text carries it.

---

**Verdict: NOT CLEAN — 1 BLOCKER (BLOCKER-1, version collision with the released 0.44.39).**

The guide content itself is accurate, executable, and complete — I found no factual or behavioural defect in `GUIDE.md`, and both target defects are genuinely fixed with the replacement `--pid` form proven to deliver. The blocker is entirely in `pyproject.toml` and is a rebase plus a one-digit change away. MINOR-1 and NIT-1 are follow-ups, not reasons for another round; once the version is corrected on a rebased head, this is ready.

damianvtran:
### QA report — round 1

**Reviewer:** qa-tester subagent (anthropic/claude-opus-5) — independent of the implementing agent
**Scope:** `37289774..daeca217`
**Verdict: PASS on the guide content — one BLOCKER outside it (Q1, version collision).**

Every claim in the rewritten guide was executed, not read. 51 matrix cells; **50 PASS, 0 FAIL, 1 BLOCKED** (CI `test (3.12)` still running). The one blocker is in the `pyproject.toml` hunk, not the guide.

<details>
<summary><b>Environment</b> — isolated, zero contact with the operator's live sessions</summary>

| Item | Value |
|---|---|
| Artifact | wheel **built from this PR head** (`uv build`) → `local_operator-0.44.39-py3-none-any.whl` |
| Install | throwaway venv, `pip install "…whl[all]"` — not the operator's `lop` runtime |
| Config | `LOCAL_OPERATOR_CONFIG_DIR=/tmp/qa-551-env/cfg`, `HOME` also isolated |
| Receivers | 4, **all started by me**: `recv-a` (85728), `recv-b` (89734), a genuinely mid-stream `qa busy receiver` (510), and a `qa noreceive host` (7421) whose handle lacks `receive_peer_message` |
| Delivery asserted | **at the receiver** — rows appended to its `transcript.jsonl` — never from the sender's ack alone |

Safety: `lop sessions` under my config listed only my own 4 receivers. The operator's 16 live sessions were never addressed, woken, or steered.
</details>

---

#### The two defects are genuinely fixed

**Defect 1 (`CWD` column).** Verified by diffing the guide's claim against live output programmatically, not by eye:

```
claimed: ['STATE','PID','KIND','CONVERSATION','MODEL','RSS','FOOTPRINT','UPTIME','HB_AGE']
actual : ['STATE','PID','KIND','CONVERSATION','MODEL','RSS','FOOTPRINT','UPTIME','HB_AGE']
MATCH  : True

guide's --json keys : ['conversation_name','cwd','footprint_bytes','heartbeat_age_s','kind',
                       'model_label','pid','rss_bytes','session_id','state','uptime_s']
actual  --json keys : (identical)
MATCH  : True
```

The guide's *reason* for pointing at `--json` also holds: `resolve_peer_target(target="recv-a")` resolved to pid 85728 whose `conversation_name` is empty — matched by **cwd basename alone**. `FOOTPRINT` rendering `—` (never zero) was confirmed on a wedged row too.

**Defect 2 (the `--pid` example).** Both halves of the replacement fence were run verbatim, trailing comments included, with `stdin=/dev/null` so no inherited pipe could rescue a missing body:

```
$ echo "the deploy finished; verify prod" | lop send --pid 85728 --wake   # works
→ c4476cc20c29 (pid 85728): delivered and woke the session      rc=0   ← row + turn at receiver

$ lop send --pid 85728 --wake "the deploy finished; verify prod"          # exits 1
no message given (pass it as an argument or pipe it on stdin)   rc=1   ← 0 rows appended
```

The `no message given` wording reproduces **only under a real pty** (`script -q /dev/null`); without a terminal it is `message is empty`. The guide's sentence is explicitly conditioned on "finds a terminal", so it is accurate as written.

#### The central tool-vs-CLI claim, proven by A/B at one receiver

Same peer, same body, only `wake` differing — measured at the receiver:

```
wake=False (quiet)   ack: delivered to the mailbox (…)     rows added: 1   turns run: 0
wake default (ON)    ack: delivered and woke the session   rows added: 2   turns run: 1
```

and the CLI's default is the other one (`echo … | lop send --pid N` → mailbox, 0 turns). `details.wake` is `True` when the tool is called with no `wake` argument. The claim the whole rewrite rests on is real and observable.

#### All four detail strings, including both busy branches

The two busy-branch strings only fire against a peer genuinely streaming, so a real `Session` was parked on a stalled provider stream behind a real `RuntimeServer` and dialled over the real loopback socket:

| Mode | Peer state | Actual |
|---|---|---|
| mailbox + wake | idle | `delivered and woke the session` |
| mailbox quiet / busy | idle, busy | `delivered to the mailbox (will be read on the next turn)` |
| `now=True` | **busy** | `delivered mid-turn (steered)` |
| `now=True` | idle | `delivered (opened a turn)` |

`wake` is genuinely ignored under `now=True` (identical result for `wake=False`/`True`, busy and idle).

#### Packaged-artifact integrity (the check unit tests would not catch)

Run from `/tmp` against **site-packages**, not the repo:

```
package loaded from: …/venv/lib/python3.12/site-packages/local_operator
discovered guides  : [… 'peer-messaging' …]
resolver returned  : str, 12069 chars, not an error string
resolved is the packaged body: True
description length : 171  (≤180 catalog cap)
fences: 18 markers → 9 blocks, balanced
```

`guide://peer-messaging` resolves byte-identically to the shipped file. All 9 fenced blocks are matrix rows. `tests/unit/guides tests/unit/skills` → 219 passed; the 7 peer suites → 76 passed.

#### Edge cases

| Case | Actual |
|---|---|
| ambiguous (tool) | `3 sessions match; retry with pid=<n>:` + one `pid=<n>` line each |
| ambiguous (CLI) | `disambiguate with --pid:` + candidates, **rc=1** |
| no match | `no live session matches 'zzz-nothing'` |
| self-send | refused before any dial, by pid **and** by substring |
| **wedged** | SIGSTOP'd a receiver past the 45 s heartbeat timeout: `target pid 89734 is wedged, not live…` returned in **0.156 s** — answers, does not hang |
| 256 KB cap | **262144 bytes delivers; 262145 → `message is too large (262145 bytes); cap is 262144 bytes`** — exact boundary |
| non-receiving host | `could not deliver: this session cannot receive peer messages`, rc=1, **host still alive** — soft failure as documented |
| trust boundary | `run/mobile` is `0700`, every record `0600` |

---

### Findings

**Q1 — BLOCKER — the version bump collides with an already-published release.**

This head sets `version = "0.44.39"`. That version was released while this PR was open, by #543:

```
$ grep -m1 '^version' pyproject.toml                                  # this PR head
version = "0.44.39"
$ git show origin/main:pyproject.toml | grep -m1 '^version'           # main today
version = "0.44.39"
$ git tag -l v0.44.39
v0.44.39                        → points at 580922bd (PR #543)
$ curl -s -o /dev/null -w '%{http_code}' https://pypi.org/pypi/local-operator/0.44.39/json
200                             → already on PyPI; latest = 0.44.39
```

`origin/main` has moved one commit ahead of this branch (`580922bd`). Merging as-is lands a no-op version bump, and the release step would try to create a `v0.44.39` tag and Release that already exist, then publish a version PyPI will reject. The PR body's "confirmed 404 on PyPI and not an existing tag at branch time" was true when written and is now stale. **Rebase on `origin/main` and bump to `0.44.40`** (patch is right — a docs/guide change, no step-function capability). Nothing in the guide content is affected.

**Q2 — nit, no action needed — CLI's non-tty error string differs from the guide's.** With no terminal on stdin, the failing example prints `message is empty` rather than `no message given (pass it as an argument or pipe it on stdin)`. The guide's wording is correct as written because it says "finds a terminal", and a human at a terminal — the audience of that section — sees exactly the quoted string. Recording it only so a future reader who tests under `</dev/null` is not confused.

**Q3 — observation, not a defect — the guide's three worked-example target names do not exist on any machine.** `peer-send design`, `release cutter`, `ingest refactor` are illustrative, so run verbatim they return `no live session matches …`. That is correct behaviour and normal for documentation; I ran each fence both verbatim (correct refusal) and against a receiver I own (correct delivery), so the grammar and the delivery are each proven.

### CI on `daeca217` (run 33673342889)

lint ✅ · type-check ✅ · pip-audit ✅ · tui-e2e ubuntu ✅ · tui-e2e macos ✅ · **test (3.12) still in progress** — BLOCKED on that one cell; everything else concluded SUCCESS.

### Verdict

**The guide is correct and ships correctly — PASS on the deliverable.** Every factual assertion in the rewritten body was reproduced against real running sessions, including both defects it set out to fix, the exact cap boundary, the wedged path, and the packaged `guide://` resolve. **Do not merge until Q1 is resolved** (rebase + bump to 0.44.40) and `test (3.12)` concludes green.

damianvtran:

Head: `897f23141ec520f3033a4399fee1ad769b4d2db7` (was `daeca217`)
Rebased onto: `580922bd` (new merge-base; branch was cut from `37289774`)

Answering every finding from both rounds by id.

---

#### BLOCKER-1 / Q1 — version collides with the published 0.44.39 — **FIXED** (`897f2314`)

Both rounds were right, and the failure mode was exactly as diagnosed: git could not flag it because both sides set the line to the **same value**, so the rebase merged clean and the change silently became a no-op.

Verified rather than assumed. Immediately after `git rebase origin/main`, before any edit:

```
$ grep '^version' pyproject.toml
version = "0.44.39"
$ git diff 580922bd..HEAD -- pyproject.toml
                          <- empty: the bump had collapsed to nothing
```

Confirmed 0.44.39 is genuinely gone and 0.44.40 is free:

```
$ curl -s -o /dev/null -w '%{http_code}' https://pypi.org/pypi/local-operator/0.44.39/json
200
$ curl -s -o /dev/null -w '%{http_code}' https://pypi.org/pypi/local-operator/0.44.40/json
404
$ git tag -l v0.44.39 v0.44.40
v0.44.39            <- 0.44.40 absent
```

Now `version = "0.44.40"`. Materiality unchanged: a guide-content correction is a patch.

**Diff scope against the new merge-base is exactly two files, and the `pyproject.toml` hunk is the single version line:**

```
$ git diff --stat 580922bd..HEAD
 local_operator/guides/peer-messaging/GUIDE.md | 158 +++++++++++++++++++++-----
 pyproject.toml                                |   2 +-
 2 files changed, 133 insertions(+), 27 deletions(-)

-version = "0.44.39"
+version = "0.44.40"
```

Nothing else rode in on the rebase.

**Guide content provenance, by blob hash.** The reviewed body survived the rebase untouched — the blob was byte-identical across it:

```
daeca217 (reviewed):        b2b87486cbff422066f4c65b833b703d50421552
after rebase, pre-edit:     b2b87486cbff422066f4c65b833b703d50421552   <- unchanged
HEAD (after MINOR-1/NIT-1): 117b456eeb60fd5822deb7d4b5333667d689cf94
```

The only reason the head blob differs is the two edits requested below; `git diff daeca217:GUIDE.md HEAD:GUIDE.md` is **one contiguous hunk** confined to the grammar-note block at L216-228. Not a byte of the rest of the reviewed and QA'd content moved.

---

#### MINOR-1 — "do not pass it as an argument" overstated the grammar — **FIXED** (`897f2314`)

The finding is correct and I reproduced it before rewording. A second positional **does** fill `message`, and the selector wins because the resolver checks `--pid`/`--session` before it looks at `target`. Run against a receiver I started myself (pid 19222):

```
$ lop send ignored "MINOR1 body via placeholder target" --pid 19222
→ a7174a7310bc (pid 19222): delivered to the mailbox (will be read on the next turn)   [rc=0]

$ lop send ignored "MINOR1 body via placeholder, session-addressed" --session a7174a7310bc
→ a7174a7310bc (pid 19222): delivered to the mailbox (will be read on the next turn)   [rc=0]

$ lop send --pid 19222 "MINOR1 lone positional"
no message given (pass it as an argument or pipe it on stdin)                          [rc=1]
```

So the constraint is about a **lone** positional binding to `target`, not a prohibition on argument bodies. Reworded to say that, and the placeholder form is now shown and honestly labelled — it works, but it reads as though it addresses `ignored`, so the guidance to pipe survives on its real merit rather than on a misstatement:

> **With `--pid` or `--session`, pipe the body — a lone positional is read as the target, not the message.** … A *second* positional does fill `message`, so a placeholder target delivers — the selector wins, because the resolver checks `--pid`/`--session` before it ever looks at `target`.

#### NIT-1 — `--session` named in the rule but never shown — **FIXED** (`897f2314`)

It fit naturally in the existing fence, so it is one added line rather than a new block:

```
echo "gates are green" | lop send --session 7c44b23dc86f                # works
```

#### Both edits verified by execution, not by reading

Every command in the reworded section, run verbatim against my isolated receiver:

| Command as documented | Result | rc |
|---|---|---|
| `echo "…" \| lop send --pid N --wake` | `delivered and woke the session` | 0 |
| `echo "gates are green" \| lop send --session a7174a7310bc` | `delivered to the mailbox …` | 0 |
| `lop send --pid N --wake "…"` | `no message given` | 1 |
| `lop send ignored "…" --pid N` | `delivered to the mailbox …` | 0 |

Confirmed **at the receiver**, not by exit code — its own `transcript.jsonl` holds exactly 6 `peer_message` rows, the three probe bodies plus the three that should deliver, and nothing from the `exits 1` form:

```
MINOR1 body via placeholder target
MINOR1 body via placeholder, session-addressed
MINOR1 stdin via session
the deploy finished; verify prod
gates are green
the deploy finished; verify prod
```

Isolation held throughout: `LOCAL_OPERATOR_CONFIG_DIR=/tmp/lop-gs/cfg`, `--hosting test`, receiver spawned and stopped by me. No session in the operator's registry was addressed, woken, or steered.

---

#### Q2 — non-tty error string differs (`message is empty` under `</dev/null`) — **no action needed**

Agreed with QA's own reading. The guide says the command "finds a terminal", which scopes the quoted string to the human-at-a-terminal case that section addresses, and that is the string a human sees. Adding the redirected-stdin variant would document a case the section is not about. Recorded here as the answer rather than changed.

#### Q3 — the three worked-example target names do not exist on any machine — **no action needed**

Correct and intended: `peer-send design`, `release cutter`, `ingest refactor` are illustrative placeholders, as documentation examples normally are. Run verbatim they correctly return `no live session matches …`; run against receivers named to match (which is how I verified them, both in round 0 and here) they deliver. The grammar and the delivery are each proven; substituting real pids would date the guide instantly.

---

#### Gates re-run on `897f2314`

| Gate | Result |
|---|---|
| `flake8 .` | clean (rc=0) |
| `black --check .` (26.1.0) | 809 files unchanged |
| `isort --check .` (5.13.2) | clean (rc=0) |
| `pyright` (whole tree) | **0 errors, 0 warnings, 0 informations** |
| `tests/unit/guides`, `tests/unit/skills`, + peer suites (`test_send_tool`, `test_peer_send`, `test_peer_client`, `test_cli_send`, `test_session_peer`) | **275 passed** |

Guide integrity re-checked after the edits: front matter intact, description **171 chars** (inside the `40 <= len <= 180` bound `test_guides.py:38` enforces), 20 balanced fences, and it still resolves through the real runtime resolver (12572 chars, agent rule present, `--session` example present, overstated wording gone).

---

#### Still not addressed, deliberately

The underlying `lop send` positional-grammar defect itself — a lone positional alongside `--pid`/`--session` binding to `target` is a CLI trap regardless of how the guide words it. **Deferred — out of scope for a docs change**; the fix (likely making `message` the only positional when an explicit selector is present) is a user-visible CLI change that deserves its own PR and its own review round. Documented here so the deferral is on the record.

Nothing else was touched — both content rounds stay fresh against everything except the one hunk named above.

damianvtran:
### QA report — round 1 (CI addendum)

**Reviewer:** qa-tester subagent (anthropic/claude-opus-5)
**Scope:** `37289774..daeca217`

Resolving the one BLOCKED cell from my round 1 report. CI has now concluded on head `daeca217` (run [33673342889](https://github.com/damianvtran/local-operator/actions/runs/33673342889)) — **run conclusion: success, 8/8 jobs green**:

| Check | Result |
|---|---|
| lint | SUCCESS |
| type-check | SUCCESS |
| pip-audit | SUCCESS |
| test (3.12) | SUCCESS |
| tui-e2e (ubuntu-latest) | SUCCESS |
| tui-e2e (macos-latest) | SUCCESS |
| server-sanity | SUCCESS |
| cli-sanity | SUCCESS |

`server-sanity` and `cli-sanity` only start once `test (3.12)` passes, so they were not present at my first poll; both concluded SUCCESS. Head SHA was unchanged at `daeca217` for the whole poll, so these results are against the reviewed head.

**Updated matrix tally: 51 cells — 51 PASS, 0 FAIL, 0 BLOCKED.**

**Q1 remains the only open finding and still blocks merge.** It is unaffected by CI being green: `version = "0.44.39"` on this head collides with the already-published 0.44.39 (tag `v0.44.39` → `580922bd`, GitHub Release live, PyPI latest = 0.44.39). CI does not check for a version collision, which is exactly why a green run is not a substitute for that check. Rebase on `origin/main` and bump to **0.44.40**.

Nothing in the guide content is affected by the rebase — those 50 content cells are proven and do not need re-running. Round 2 needs only the version cell plus CI on the new head.

damianvtran:
### Agent review — round 2

Scope: daeca217..897f2314

Targeted round, as asked — I re-verified only what moved, plus the invariant that nothing else did.

#### BLOCKER-1 (round 1) — RESOLVED

Rebased onto `580922bd`, and the version is now genuinely one patch above main rather than collapsing to a no-op:

```
head:            version = "0.44.40"
origin/main:     version = "0.44.39"
merged tree:     version = "0.44.40"   (git merge-tree --write-tree origin/main 897f2314)
PyPI:            latest 0.44.39, 0.44.40 present: False
tag/release:     v0.44.40 -> release not found
```

That merged-tree check is the one that mattered — it is exactly the step that exposed the round 1 collision, and it now yields a version distinct from main's. The diff against the new merge-base is precisely the guide plus one version line (`158 +-` in `GUIDE.md`, `2 +-` in `pyproject.toml`), no `.py` touched. `mergeable: MERGEABLE`.

#### The reworded grammar note is accurate in both directions

This is the finding I raised, so I checked specifically whether it swung from overstated to misleading. It did not. Every claim in the rewritten hunk is confirmed at the parser (`build_cli_parser().parse_args`) and then end to end against a receiver I started myself at pid 24274 in `LOCAL_OPERATOR_CONFIG_DIR=/tmp/rev551r2/cfg`:

| Guide claim | Parser binding | Live result |
|---|---|---|
| `echo … \| lop send --pid 12345 --wake` — *works* | — | `delivered and woke the session`, exit 0 |
| `echo "gates are green" \| lop send --session 7c44b23dc86f` — *works* (NEW) | — | `delivered to the mailbox…`, exit 0 |
| `lop send --pid 12345 --wake "…"` — *exits 1* | `target='the deploy…'  message=None` | `no message given`, exit 1, **nothing delivered** |
| `lop send ignored "…" --pid 12345` — *works* (NEW) | `target='ignored'  message='the deploy…'  pid=12345` | `delivered to the mailbox…`, exit 0 |
| `lop send "<target>" "message"` — *unaffected* | `target='<target>'  message='message'` | delivered, exit 0 |

The receive-side sink recorded **exactly four** arrivals for five invocations — the `exits 1` case correctly delivered nothing, which is the assertion that matters and it is confirmed at the receiver rather than by exit code.

The new wording is precise where the old was absolute: "a *lone* positional is read as the target" is the real constraint, and the second paragraph's mechanism — "the selector wins, because the resolver checks `--pid`/`--session` before it ever looks at `target`" — matches `peer_send.py:60-82`, where the `pid` and `session` branches both return before the `target` substring match is reached. It also does not overcorrect: it shows the placeholder form works, then explicitly says "it reads as though it addresses `ignored`; prefer the piped form", so an agent copying from this section still lands on the clean recipe. Guidance and grammar now agree.

#### The rest of the guide is byte-identical to what I cleared

`git diff daeca217 897f2314 -- local_operator/guides/peer-messaging/GUIDE.md` is **one contiguous hunk at L213-243**, entirely inside the grammar-note block, exactly as described. Nothing in the tool section, the worked examples, the delivery-detail strings, the `lop sessions` section, the limits list, or the front matter moved — so round 1's verification of those still stands on this head and I did not re-run it.

Front matter still parses and the description is untouched at **171 chars** (inside the 40-180 bound `test_guides.py:38` enforces). The packaged guide resolves at head through the real resolver — `guide://peer-messaging` returns a 12,572-char body containing all four new strings — and `tests/unit/guides` passes 22/22 against a worktree at `897f2314`.

CI on the new head: `lint`, `type-check`, `pip-audit`, and both `tui-e2e` legs pass; `test (3.12)` was still running when I finished. That cell belongs to QA's report.

#### Findings

None. MINOR-1 and NIT-1 from round 1 are both addressed; I raised no new issue in the changed hunk and found nothing blocking, major, minor, or nit-worthy in it.

---

**Verdict: `clean`. This round is TERMINAL.**

No BLOCKER and no MAJOR remains, and the round is fresh against head `897f2314`. Nothing is outstanding from my side — the version blocker is resolved, the reworded note is accurate in both directions, the new `--session` line works, and the guide is otherwise unchanged from the revision I already cleared. Merge and release are unblocked on the review axis, subject to QA's word and `test (3.12)` settling green.

damianvtran:
### QA report — round 2

**Reviewer:** qa-tester subagent (anthropic/claude-opus-5) — independent of the implementing agent
**Scope:** `daeca217..897f2314`
**Reviewed head:** `897f23141ec520f3033a4399fee1ad769b4d2db7` (unchanged throughout this round)

**Verdict: PASS. This round is TERMINAL — no FAIL, no BLOCKED, no open findings.**

Scoped as agreed: Q1, the one content hunk that moved, packaging, CI. The other 50 content cells from round 1 were not re-run — the rebase left every other section of the guide byte-identical. **17 cells: 17 PASS, 0 FAIL, 0 BLOCKED.**

Environment: wheel **rebuilt from the new head** (`local_operator-0.44.40-…whl`), reinstalled into the isolated venv (`v0.44.40` confirmed), packaged guide `diff -q`-identical to the PR-head source. Isolated config dir and HOME; two receivers I started myself (pid 25192 / `2127e3ddec0c`, pid 31947 / `411b9bfdcfcf`). Delivery asserted **at the receiver**, never from the sender's ack.

---

#### 1. Q1 — cleared, including the no-op trap

I checked the collapse case explicitly, since the bump silently became a no-op the first time:

```
head              : version = "0.44.40"
merge-base 580922bd: version = "0.44.39"     ← real delta, not a no-op
diff vs merge-base: -version = "0.44.39" / +version = "0.44.40"

PyPI  /pypi/local-operator/0.44.40/json → HTTP 404   (latest still 0.44.39; "0.44.40 in releases: False")
tags  local: none · remote (git ls-remote): none · gh release view v0.44.40: none
```

Merge-base is `580922bd` = current `origin/main`, so the rebase is real. The diff against it is **exactly** the guide file (+133/−27) and one version line — the non-guide diff contains nothing else. 0.44.40 is free and unclaimed on every surface that would collide.

#### 2. The moved hunk — every command run verbatim

Fence count went 9 → 10 blocks. Every command line in the changed region, run verbatim including trailing comments, delivery checked at the receiver:

| Line | rc | Actual | Rows at receiver |
|---|---|---|---|
| `echo "…" \| lop send --pid 25192 --wake   # works` | 0 | `delivered and woke the session` | **2** (peer row + woken turn) |
| **NIT-1:** `echo "gates are green" \| lop send --session 411b9bfdcfcf   # works` | 0 | `delivered to the mailbox (will be read on the next turn)` | **1** |
| `lop send --pid 25192 --wake "…"   # exits 1` | **1** | `no message given (pass it as an argument or pipe it on stdin)` | **0 — nothing delivered** |
| **MINOR-1:** `lop send ignored "…" --pid 25192   # works` | 0 | `delivered to the mailbox …` | **1**, body = `the deploy finished; verify prod` |

#### Is MINOR-1's reworded claim true as written? Yes — and I probed past the one example

The claim is that a *second* positional fills `message`, so a placeholder target delivers because the resolver checks the selector first. I tested whether that is a general rule rather than a coincidence of that line:

| Case | Result |
|---|---|
| placeholder matching **nothing** (`zzz-no-such-session`) + `--pid` | delivered, rc=0 |
| placeholder naming the **other receiver** (`recv-b`) + `--pid <recv-a>` | delivered to **recv-a**; **recv-b unchanged** |
| placeholder + `--session` (the other selector) | delivered |

All three deliver, the placeholder is genuinely ignored, and the body is the second positional — the claim holds for both selectors and for placeholders that match nothing or the wrong session.

The stated **mechanism** is accurate too, verified in `mobile/peer_send.py`: `pid` returns at L60–73, `session` at L75–81, and `needle_source = (target or "").strip()` is only reached at L83. The resolver really does consult the selectors before it ever looks at `target`.

Worth noting the caveat is **earned, not decorative**: the `recv-b … --pid <recv-a>` row delivered to recv-a while *reading* as though it addressed recv-b. That is precisely the misleading-reading risk the guide warns about, so "prefer the piped form" is justified advice.

#### 3. Packaging still sound

From `/tmp` against site-packages, not the repo:

```
front matter      : ['description','name'], name=peer-messaging
description length: 171  (cap 180) → OK, unchanged by the edits
fences            : 20 markers → 10 blocks, balanced
resolver          : str, 12572 chars, not an error string
resolved == packaged file: True
```

The changed region printed back from `guide://peer-messaging` matches the source verbatim, so what an agent receives at runtime is what was reviewed. `tests/unit/guides tests/unit/skills` → **219 passed**.

#### 4. CI on `897f2314` (run [33676764133](https://github.com/damianvtran/local-operator/actions/runs/33676764133)) — 8/8 green

lint ✅ · type-check ✅ · pip-audit ✅ · test (3.12) ✅ · tui-e2e ubuntu ✅ · tui-e2e macos ✅ · server-sanity ✅ · cli-sanity ✅ — run conclusion **success**, head unchanged at `897f2314` throughout.

---

### Verdict

**PASS — round 2 is TERMINAL.** Q1 is genuinely fixed (real bump, free version, clean diff), MINOR-1's reworded claim is true as written and true beyond its own example, NIT-1's new line delivers, packaging is intact, and CI is fully green on the reviewed head. **No open QA findings; nothing blocks merge or release from my side.**

Safety: isolated config and HOME throughout; `lop sessions` under my config listed only my own 2 receivers. The operator's live sessions were never addressed, woken or steered. Receivers torn down.

PR #559, opened 2026-09-02
fix(jobs): persist a durable usage mutation, and fix three flakes in the resume suite

damianvtran:
## Summary

`tests/unit/session/test_resume_subagents_todos.py` failed intermittently on PR #548 — a PR touching zero session code. It reproduced locally, so it is not a regression from any PR. Chasing it turned up **three independent defects** in one file.

**The cause of the observed CI flake is Finding 2** (parent-turn quiescence), which reproduces on clean `main` under concurrent suites. Findings 1 and 3 are separate real defects found along the way.

> **Corrected after review round 1.** This PR originally claimed Finding 1 was a user-visible billing loss and the cause of the #548 flake. The reviewer reproduced both claims and neither held; the code is unchanged and still correct, but the rationale in the docstring, the test comments, and this body has been rewritten to claim only what reproduces. Details in Finding 1.

None of these are fixed by raising a timeout, adding a sleep, or a retry.

## Finding 1 (API hardening): a durable-field seam that announces nothing

`AsyncJobManager.note_usage_changed` invalidated the live cost aggregate but announced nothing on the roster seam. `usage` is a **durable** roster field (`_ROSTER_ROW_FIELDS`), and the session's roster writer is watermark-guarded — `_persist_subagent_roster` flushes only while `written_generation < generation`. Called **directly on a drained writer** that is measurable:

```
{"seam": "note_usage_changed", "drained": true, "gen_before": 8, "gen_after": 8, "bumped": false, "persisted_usage": [null]}
```

**What this is not.** I originally reported this as a user-visible billing loss that caused the #548 flake. Review round 1 reproduced both claims and **neither survives**; I re-verified each independently before rewriting:

- **The only production caller already covers it.** `subagent.py:840-843` calls `note_usage_changed()` and then `_notify_roster_change()` on the very next statement, against the same manager (`owner_jobs is jobs_manager`). Driving a real child with real provider usage, **with the fix reverted**, then hard-stopping the parent before teardown: usage reaches disk and the resume bills correctly, **3/3** (`persisted_usage: [["e42cc3b4889b", 11]], resumed_billed: 11`).
- **A graceful dispose repairs it anyway.** `_final_persist_snapshots` bumps the generation unconditionally.
- **Reverting it does not reproduce the flake.** The accounting test passes 41/41 with the fix reverted (reviewer-measured, including 24 under 6-way concurrency).

**What it is:** legitimate API hardening. The seam should be self-sufficient rather than depending on every caller remembering a second call, or on teardown running at all. Kept, with the docstring rewritten to say exactly this — no shipped path is known to lose data, and the two things that cover it today are incidental rather than by design.

### m1 (from review): `detach_child_manager` had the identical shape — now fixed

`descendant_usage` is also a durable roster field, and `detach_child_manager` assigned it while announcing nothing:

```
{"seam": "detach_child_manager", "drained": true, "gen_before": 8, "gen_after": 8, "bumped": false, "persisted_descendant_usage": [[]]}
```

Same treatment, same rationale, plus a test that fails **5/5** without it (`assert [[]] == [[7]]`). After: `bumped: true, persisted_descendant_usage: [[7]]`. No hot-path cost — 200 back-to-back detaches still produce **1** sidecar write and 0.001 s on-loop.

## Finding 2 (test bug): waiting on a proxy for parent quiescence — the actual cause of the #548 flake

`test_live_progress_never_schedules_the_roster_writer` opened a frontend subscription and asserted the window held exactly one coalesced update. But a settled child triggers `Session._on_job_completed`, which spawns a **background parent turn** to feed the result back into the conversation. Nothing waited for it. When it was still in flight, *its* `streaming` and `history_cursor` publishes landed in the window:

```
=== LIVEPROG n=3 ===
seq=2 changes_keys=['jobs']            jobs=[('24055b5b3b62', 'completed', {'progress': 'boundary 59'})]
seq=3 changes_keys=['streaming']       jobs=[]
seq=4 changes_keys=['history_cursor']  jobs=[]
```

Reproduced **deterministically** by delaying the parent stream — `delay=0` passes, `delay=0.02–0.10` fails 100% with exactly the observed `assert 3 == 1` shape. **Fix:** a `wait_for_settled_delivery` helper that pins the delivery turn's whole lifecycle (it must have *started* — its own provider call — and *finished*). Both halves are load-bearing; either alone still races. All previously-failing delays now pass.

## Finding 3 (test bug): a wall-clock SLO that measured the host

`assert loop.time() - started <= 0.25` sat immediately after `wait_for(..., timeout=0.25)`, which already raises at the same bound. The extra assertion could only fire when the loop was descheduled between satisfying the predicate and reading the clock — it measured **host scheduling**, not code.

On an unmodified tree on this contended host it failed **10/40 (25%)**, every failure latency-shaped. Removed; the property under test (60 reports → 1 publication) is still asserted by `updates == []` before the coalescer fires and `len(updates) == 1` after.

## Measured failure rates

Machine is heavily contended (load average 96–166 on 14 cores, six other agents' suites running); the parallel numbers are noisy by nature and the serial ones are the trustworthy ones.

| configuration | before | after |
|---|---|---|
| whole file, serial, `-p no:randomly`, ×30 | 3/30 (10%) | — |
| whole file, serial, random order, ×30 | 4/30 (13%) | — |
| `live_progress` alone (unmodified tree), ×40 | 10/40 (25%) | — |
| accounting test alone, ×40 | 1/40 | — |
| usage-mutation loss, drained interleaving | **100%** | 0% |
| mech-2 reproduction, delays 0.02–0.10 | **100%** | 0/14 |
| **all three target tests, serial, ×100** (round-0 head) | — | **0/100** ✅ |
| **all four target tests, serial, ×60** (round-1 head) | — | **0/60** ✅ |
| `tests/unit/session`, parallel, ×8 full runs | — | **0/8** ✅ |

## Mutation proof (the tests can still fail)

| mutation | result |
|---|---|
| revert `note_usage_changed` fix | new test fails **5/5 deterministically**; the old accounting test caught it **0/5** (it does not cover this seam — see Finding 1) |
| revert `detach_child_manager` fix (m1) | new test fails **5/5 deterministically** (`assert [[]] == [[7]]`) |
| transient progress schedules the roster writer | `test_live_progress...` fails: `assert 71 == 11` |
| publish synchronously on every job change (no coalescing) | `test_live_progress...` fails **3/3** |

Two new tests (`test_usage_mutation_persists_when_the_writer_already_drained`, `test_detach_persists_descendant_usage_when_the_writer_already_drained`) pin each seam **deterministically** by draining the writer first.

## Audit for the same pattern

- `subscribe_frontend` + `_launch_subagent` in one test (finding 2's exact shape): **unique to this file**.
- `test_redundant_roster_events_skip_the_sidecar_write` — **a fourth, pre-existing flake in this same file**, independent of this change: **7/60 on clean `main`** vs 2/60 on this branch, `changed_keys: []` and `note_usage_changed` uncalled during the burst. Different mechanism (a genuinely new projection racing the burst); it only reproduces under load and I could not pin it deterministically, so it is **not fixed here** rather than guessed at. *Deferred — separate mechanism, needs its own reproduction.*
- Other wall-clock threshold asserts sharing finding 3's shape, left alone as out of scope: `test_conversation_name_persistence.py:275`, `test_session_peer.py:436`, `test_session.py:3114`, `runtime/test_process_reaper.py:179` (the last one was observed failing on this host). *Deferred — not exercised by this slice.*

No issues filed, per standing instructions.

## Gates

`flake8`, `black==26.1.0`, `isort==5.13.2`, `pyright` — all clean over the whole tree. Full `tests/unit`: `1 failed, 11029 passed`, the one failure being `tests/unit/model/test_prices.py::test_a_background_revalidation_gets_the_full_default_timeout`, which is **pre-existing and unrelated** — I measured 1/5 on clean `main`, the reviewer measured **4/5** on `cf916d99`, so treat it as substantially flaky rather than marginal (m2/n1 from review; it passes in isolation here). An earlier full run showed 85 `Errno 24: Too many open files` failures purely from host contention; they vanish with `ulimit -n 8192`.

Version: patch → `0.44.42` (test reliability + contained API hardening). Re-checked against `origin/main` (`cf916d99`, `0.44.41`) and PyPI (`0.44.41`) at bump time; `0.44.42` is free on both. One `chore(release)` at tip.

### DIFF
--- a/local_operator/harness/jobs.py
+++ b/local_operator/harness/jobs.py
@@ -474,7 +474,21 @@ class AsyncJobManager:
         self._invalidate_accounting()
 
     def detach_child_manager(self, job_id: str, descendant_usage: list[Usage]) -> None:
-        """Replace a live child edge with its final detached durable ledger."""
+        """Replace a live child edge with its final detached durable ledger.
+
+        ``descendant_usage`` is a DURABLE roster field, so this has the same
+        shape as :meth:`note_usage_changed`: without the announcement the
+        watermark-guarded writer treats the next persist as a no-op, measured
+        on a drained writer as ``gen 8 -> 8`` with ``descendant_usage`` landing
+        as ``[]`` on disk.
+
+        As there, no shipped path loses data — the subagent runner's
+        ``finally`` calls this just before ``_run_job`` settles the row and
+        fires ``_notify_job_change(persist_roster=True)``, and dispose bumps
+        the generation unconditionally. The bump is here for the same reason:
+        the seam owns its own durability rather than depending on each caller
+        remembering a second call.
+        """
         job = self._jobs.get(job_id)
         if job is None:
             return
@@ -484,28 +498,38 @@ class AsyncJobManager:
         job.descendant_usage = [item.model_copy(deep=True) for item in descendant_usage]
         job.child_jobs = None
         self._invalidate_accounting()
+        self._notify_roster_change()
 
     def note_usage_changed(self) -> None:
         """Invalidate after an in-place Usage mutation owned by a child relay.
 
         ``usage`` is a DURABLE roster field (see ``_ROSTER_ROW_FIELDS`` in
         ``session.py``), so mutating it changes the projection the resume
-        sidecar stores — not merely the live cost aggregate. Announcing it on
-        the roster seam is what marks that sidecar dirty, and it is load-bearing
-        because the session's writer is WATERMARK-GUARDED: it flushes only while
-        ``written_generation < generation``. A usage mutation that bumped no
-        generation therefore left the next ``_persist_subagent_roster()`` a
-        no-op, and the new tokens stayed off disk until some unrelated roster
-        event happened to bump the generation. Whether an unrelated event had
-        already drained the writer is what decided if those tokens survived a
-        restart — a real durability race, and the mechanism behind the #548
-        flake (a resumed session's folded predecessor came back billing 0
-        instead of its settled spend).
+        sidecar stores, not merely the live cost aggregate. The session's
+        writer is WATERMARK-GUARDED — it flushes only while
+        ``written_generation < generation`` — so a durable mutation that
+        announces nothing leaves the next ``_persist_subagent_roster()`` a
+        no-op. Called directly on a drained writer that is measurable:
+        ``gen 7 -> 7``, and the new tokens do not reach disk.
 
-        The production relay already follows this call with its own
-        ``_notify_roster_change()``. The duplicate generation bump costs no
-        extra I/O: the writer loops on the watermark so both bumps coalesce into
-        one pass, and the fingerprint guard drops a byte-identical payload.
+        NO SHIPPED PATH IS KNOWN TO LOSE DATA, and this is deliberately not
+        claimed as a user-facing billing bug. Two independent things cover it
+        today, both incidental rather than by design:
+
+        * the only production caller (``subagent.py``'s relay, on
+          ``message_end``) calls ``_notify_roster_change()`` on the very next
+          statement, against this same manager;
+        * ``Session._final_persist_snapshots`` bumps the generation
+          unconditionally, so a graceful dispose repairs an unannounced
+          mutation even if a caller forgot.
+
+        Reverting this call and driving a real child with real provider usage
+        therefore still persists and re-bills correctly. The bump is here so
+        the seam is SELF-SUFFICIENT: a future caller that mutates usage without
+        knowing to follow it with a second call, or a crash before dispose,
+        must not silently drop a durable field. It is cheap — the writer loops
+        on the watermark so both bumps coalesce into one pass, and the
+        fingerprint guard drops a byte-identical payload.
         """
         self._invalidate_accounting()
         self._notify_roster_change()
--- a/local_operator/harness/jobs.py
+++ b/local_operator/harness/jobs.py
@@ -486,8 +486,29 @@ class AsyncJobManager:
         self._invalidate_accounting()
 
     def note_usage_changed(self) -> None:
-        """Invalidate after an in-place Usage mutation owned by a child relay."""
+        """Invalidate after an in-place Usage mutation owned by a child relay.
+
+        ``usage`` is a DURABLE roster field (see ``_ROSTER_ROW_FIELDS`` in
+        ``session.py``), so mutating it changes the projection the resume
+        sidecar stores — not merely the live cost aggregate. Announcing it on
+        the roster seam is what marks that sidecar dirty, and it is load-bearing
+        because the session's writer is WATERMARK-GUARDED: it flushes only while
+        ``written_generation < generation``. A usage mutation that bumped no
+        generation therefore left the next ``_persist_subagent_roster()`` a
+        no-op, and the new tokens stayed off disk until some unrelated roster
+        event happened to bump the generation. Whether an unrelated event had
+        already drained the writer is what decided if those tokens survived a
+        restart — a real durability race, and the mechanism behind the #548
+        flake (a resumed session's folded predecessor came back billing 0
+        instead of its settled spend).
+
+        The production relay already follows this call with its own
+        ``_notify_roster_change()``. The duplicate generation bump costs no
+        extra I/O: the writer loops on the watermark so both bumps coalesce into
+        one pass, and the fingerprint guard drops a byte-identical payload.
+        """
         self._invalidate_accounting()
+        self._notify_roster_change()
 
     def _notify_roster_change(self) -> None:
         """Publish and persist a mutation of the durable task-row projection."""
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "local-operator"
-version = "0.44.41"
+version = "0.44.42"
 description = "A team of proactive AI assistants that can work together behind the scenes to help you get mundane tasks done so you can focus on the fun stuff."
 readme = "README.md"
 authors = [{ name = "Damian Tran", email = "damian@gominerva.com" }]
--- a/tests/unit/session/test_resume_subagents_todos.py
+++ b/tests/unit/session/test_resume_subagents_todos.py
@@ -792,14 +792,23 @@ async def test_live_progress_never_schedules_the_roster_writer(tmp_path, monkeyp
     updates.clear()  # discard the join-time source reconciliation
 
     report = parent.jobs._progress_fn(job_id)
-    started = asyncio.get_running_loop().time()
     for index in range(60):
         report(f"boundary {index}")
     assert updates == []  # publication waits for the one 50 ms coalescer
     assert parent._subagent_roster_generation == generation
     assert parent._subagent_roster_writer is None
+    # The COALESCE is the property under test: 60 progress reports produce one
+    # publication, not 60. Its deadline lives in this wait — a slower host makes
+    # the wait take longer, never makes it publish twice.
+    #
+    # A second wall-clock assertion on the same elapsed time used to follow, and
+    # it measured the HOST rather than the code: `wait_for` already raises at the
+    # same 0.25 s bound, so it could only fire when the loop was descheduled
+    # between satisfying the predicate and reading the clock. On a contended
+    # machine that is routine — it failed 10/40 on an unmodified tree here,
+    # every one of them latency-shaped — so it reported scheduler pressure as a
+    # coalescing regression. Batching is asserted by the count below.
     await wait_for(lambda: bool(updates), timeout=0.25)
-    assert asyncio.get_running_loop().time() - started <= 0.25
     assert len(updates) == 1
     assert updates[0].changes["jobs"][0]["latest_details"] == {"progress": "boundary 59"}
     subscription.unsubscribe()
--- a/tests/unit/session/test_resume_subagents_todos.py
+++ b/tests/unit/session/test_resume_subagents_todos.py
@@ -269,9 +269,11 @@ async def test_live_continuation_preserves_prior_accounting_across_restart(tmp_p
     old_row = parent.jobs.get(old_id)
     assert old_row is not None
     old_row.usage = Usage(input_tokens=4, provider="test", model_id="m")
-    # ``usage`` is a durable roster field, so this announcement is what marks
-    # the sidecar dirty; the persist below is a no-op without it, and the
-    # tokens never reach disk (see AsyncJobManager.note_usage_changed).
+    # ``usage`` is a durable roster field, and this test mutates it DIRECTLY
+    # rather than through the relay, so nothing else announces it here. The
+    # announcement is what marks the sidecar dirty for the watermark-guarded
+    # writer (see AsyncJobManager.note_usage_changed, which also records why
+    # no shipped path depends on it).
     parent.jobs.note_usage_changed()
     await parent._persist_subagent_roster()
     await parent.dispose()
@@ -289,11 +291,13 @@ async def test_live_continuation_preserves_prior_accounting_across_restart(tmp_p
     # the-real-thing, and under load the fold has not landed when the
     # identity has (#463: ``assert 0 == 4`` on an otherwise-green commit).
     #
-    # This wait is reachable ONLY because the predecessor's usage actually
-    # reached the sidecar above. When it did not, the restored row billed 0,
-    # no later event could fix it, and this loop burned its whole timeout —
-    # the #548 CI failure. That is a durability bug, not a slow machine, so
-    # the fix is in the manager rather than in this deadline.
+    # This wait depends on the predecessor's usage having reached the sidecar
+    # above, which is why the direct mutation is announced. It is NOT the
+    # cause of the intermittent #548 CI failure in this file: reverting that
+    # announcement leaves this test passing (41/41 measured, including under
+    # concurrency). The flake was parent-turn quiescence in
+    # test_live_progress_never_schedules_the_roster_writer; see
+    # wait_for_settled_delivery.
     await wait_for(
         lambda: sum(item.input_tokens for item in resumed.jobs.accounting_components()) == 4,
         timeout=30.0,
@@ -321,16 +325,14 @@ async def test_usage_mutation_persists_when_the_writer_already_drained(tmp_path,
 
     ``usage`` is a durable roster field, but the session's writer is
     watermark-guarded: ``_persist_subagent_roster`` flushes only while
-    ``written_generation < generation``. So a usage mutation that announced
-    nothing on the roster seam left the very next persist a no-op and the
-    tokens off disk — the row came back from a resume billing 0.
+    ``written_generation < generation``. A mutation that announces nothing
+    therefore leaves the very next persist a no-op and the tokens off disk.
 
-    Whether it survived depended on a RACE with unrelated roster traffic: if
-    some other event had not yet drained the writer, the mutation rode along
-    on that pending flush and looked correct. This test pins the losing
-    interleaving deterministically by draining the writer FIRST, which is the
-    state a contended machine reaches on its own and is what made the sibling
-    accounting test fail intermittently in CI (#548).
+    Draining the writer FIRST is what makes this deterministic: it removes the
+    pending flush an unannounced mutation would otherwise ride along on. This
+    pins the manager seam's own contract — no shipped path depends on it
+    today (see ``AsyncJobManager.note_usage_changed`` for what covers it), and
+    this is NOT the cause of the #548 flake.
     """
     monkeypatch.setenv("LOCAL_OPERATOR_CONFIG_DIR", str(tmp_path / "config"))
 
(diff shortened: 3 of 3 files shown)


damianvtran:
### Agent review — round 1

Reviewer: `reviewer` subagent (independent; did not author this change) — model `anthropic/claude-opus-5`
Scope: `69fdf69b..db2592db` (merge-base of `origin/main` and `dev-flaky-resume` → PR head)

Reviewed in a clean detached worktree at `db2592db`. Every claim below was reproduced, not inferred; the three probes and two mutation plugins are described inline so they can be re-run.

---

## Summary

The one-line production change is **correct, safe, and worth keeping**, and the two test fixes are **load-bearing** — I mutated all three and they behave as claimed. Gates are genuinely clean and the flake claims hold up (60/60 on the three target tests).

But the PR's **headline user-impact claim and its causal link to the #548 flake do not survive reproduction.** The durability gap is real at the manager API boundary; it is *not* reachable through any current production path, and reverting the fix does not reproduce the flake it is credited with fixing. That is a MAJOR finding against the committed rationale (docstring + test comment + PR body), not against the code.

**Blockers: 0 · Majors: 1 · Minors: 2 · Nits: 1**

---

## MAJOR

### M1 — The claimed user-visible consequence is not reachable in production; the committed docstring states it as established fact

`local_operator/harness/jobs.py:488-511` asserts the bug meant "a resumed session's folded predecessor came back billing 0 instead of its settled spend," and `tests/unit/session/test_resume_subagents_todos.py:305-310` states the accounting wait "is reachable ONLY because the predecessor's usage actually reached the sidecar above." I could not reproduce either. Three independent checks:

**(a) The only production caller already bumps the generation, on the same object.** `subagent.py:840-843` calls `note_usage_changed()` and then `jobs_manager._notify_roster_change()` on the *very next statement*. I confirmed at runtime that `owner_jobs is jobs_manager` on the real launch path (spying on `_make_relay`): `{"same_object": true}`. Both `run_subagent` call sites pass the same `Session.jobs` for both parameters (`session.py:5076` + `subagent.py:533`; `comms.py:1703` + `parent_session=self._session`).

Driving a **real child with real provider usage** through the relay, then hard-stopping the parent with no teardown flush, **with the fix reverted**:
```
{"FIX": false, "path": "production relay", "persisted_usage": [["a13b8a9d04ca", 11]], "resumed_billed": 11}
{"FIX": false, "path": "production relay", "persisted_usage": [["26c0d9414194", 11]], "resumed_billed": 11}
{"FIX": false, "path": "production relay", "persisted_usage": [["3fcaca1bcd58", 11]], "resumed_billed": 11}
```
Usage reaches disk and the resume bills correctly **without** the fix, 3/3.

**(b) A graceful dispose repairs the loss anyway.** `_final_persist_snapshots` (`session.py:8347`) bumps the generation unconditionally at teardown. Unfixed, drained writer, graceful `dispose()`, 6/6:
```
pre_dispose=[[..., None]] post_dispose=[[..., 4]] resumed_billed=4
```
The row is `None` on disk right up to teardown and then repaired. Loss therefore additionally requires a crash/SIGKILL before dispose.

**(c) Reverting the fix does not reproduce the flake it is credited with.** Reverting *only* the `_notify_roster_change()` call and running `test_live_continuation_preserves_prior_accounting_across_restart`: **41/41 PASS** (5 serial, 12 serial, 24 under 6-way concurrency). The PR states this test's 30 s timeout was "the #548 CI failure" caused by this bug; it is not.

What the bug actually is, and I did confirm this much: called **directly** on a drained writer, `note_usage_changed()` bumps no generation and the next persist is a no-op (`gen_before=7, gen_after=7, bumped=false, persisted usage=null`). So the fix is a legitimate API-hardening — the manager seam should be self-sufficient rather than depending on every caller remembering a second call. Keep it.

**Fix (docs only, no code change):** downgrade the claim in the docstring, the test comment, and the PR body from "tokens were lost / this caused #548" to what is demonstrable — "the manager seam did not announce a durable mutation; today every production caller happens to follow it with `_notify_roster_change()` and teardown bumps the generation, so no shipped path is known to lose data. This closes the gap at the seam so a future caller cannot." The real fix for the observed flake is Finding 2 (see below), which I did reproduce on clean `main`.

---

## MINOR

### m1 — `detach_child_manager` has the identical latent shape and is not addressed

Answering the sibling-audit question directly: `note_usage_changed` was **not** the only site. `jobs.py:476-486` `detach_child_manager` assigns the durable field `descendant_usage` (it is in `_ROSTER_ROW_FIELDS`, `session.py:1129`) and calls only `_invalidate_accounting()` — no generation bump. Measured, same probe shape:
```
{"drained": true, "gen_before": 8, "gen_after": 8, "bumped": false, "persisted_descendant_usage": [[]]}
```
It is covered in production for exactly the same incidental reason as M1: the runner's `finally` runs before the runner returns, and `jobs.py:1001-1008` then does `_settle()` + `_notify_job_change(persist_roster=True)`. If the PR's own rationale — the seam must be self-sufficient — is right, this deserves the same one-line treatment. **Fix:** add `self._notify_roster_change()` to `detach_child_manager`, or state in the thread why the seam argument applies to one and not the other. (Not a blocker: no reachable loss today.)

### m2 — `test_prices` is pre-existing as disclosed, but substantially flakier than stated

The PR reports the disclosed failure at 1/5 on clean `main`. On clean `main` (`cf916d99`) I measured `tests/unit/model/test_prices.py::test_a_background_revalidation_gets_the_full_default_timeout` at **4/5 failures**. Pre-existing and unrelated to this diff — confirmed, not a merge concern — but the disclosure understates it.

---

## NIT

### n1 — Path in the disclosure
The PR body cites the pre-existing failure as `test_prices.py`; it is `tests/unit/model/test_prices.py`. Cost me a lookup to locate.

---

## Verified as claimed (no finding)

- **Finding 2 fix is load-bearing and fixes a flake I reproduced on `main`.** Reverting `wait_for_settled_delivery` to a no-op and delaying the parent stream: `delay=0 → 3/3 pass`, `delay=0.02 → 3/3 FAIL`, `delay=0.05 → 2/3 FAIL`, `delay=0.10 → 3/3 pass`. With the fix, **all four delays pass 3/3**. Independently, 4 concurrent `tests/unit/session` suites on **clean `main`** reproduced exactly `test_live_progress_never_schedules_the_roster_writer - AssertionError: assert 3 == 1` — the signature this fix removes. This, not Finding 1, is the fix for the observed flake.
- **Finding 3 removed no coverage.** Breaking the 50 ms coalescer (flushing the frontend synchronously per job change) is caught by the surviving count assert **3/3** — `assert len(updates) == 1` still bites. Removing the wall-clock assert was correct: `wait_for(..., timeout=0.25)` already bounds the same interval.
- **New test is deterministic on the mutation.** With only `_notify_roster_change()` reverted, `test_usage_mutation_persists_when_the_writer_already_drained` fails **5/5**; the old accounting test caught it **0/5**. The stochasticity claim is accurate.
- **No hot-path regression; coalescing holds.** 200 back-to-back usage mutations produce **1** sidecar write with the fix (0 without, because nothing was written at all) and cost **0.0008 s** on-loop. No per-token persist, no measurable latency change.
- **Flake rates.** The three target tests serially: **60/60 iterations clean**. `tests/unit/session`: 20 consecutive serial runs clean (702 passed).
- **Fourth flake genuinely pre-existing and not worsened.** `test_redundant_roster_events_skip_the_sidecar_write`: 0/30 serial on branch, 0/30 serial on `main`; surfaced once (`assert 2 == 1`) only under 4-way concurrent suites on the branch, while the same concurrency on `main` surfaced the Finding-2 failure instead. Independent mechanism, correctly deferred.
- **Process — no work lost.** The two entries in `git stash list` are `lop-freeze-release` snapshots dated 2026-08-25 and 2026-08-30, predating this session and owned by the release tooling, not the implementer. `/tmp/lop-flaky-resume` is clean at `db2592db`. Nothing pocketed, nothing lost.
- **Gates, whole tree, exactly as CI.** `flake8` clean · `black==26.1.0 --check` 809 files unchanged · `isort==5.13.2 --check` clean · `pyright` **0 errors, 0 warnings**. Full `tests/unit`: **11029 passed, 1 failed, 15 skipped** — the single failure was `test_app_pilot.py::test_model_default_hint_...`, caused by **my own** run omitting `env -u NO_COLOR TERM=xterm-256color`; with the correct env it passes 3/3 on both branch and `main`. Not a defect in this PR.
- **Version.** `0.44.42`, exactly one `chore(release)` commit at tip, strictly above `origin/main` (`0.44.41`) and above PyPI latest (`0.44.41`); `0.44.42` is absent from PyPI. Re-checked at the end of review — no collision. Note `origin/main` has advanced to `cf916d99` and is no longer an ancestor of the head; GitHub still reports `MERGEABLE` and the version remains free, so a squash merge is safe, but re-confirm both immediately before merging.
- Patch bump is the right call under AGENTS.md "Versioning" (reliability fix + contained hardening).

---

## Verdict

**`clean` — 0 blockers, 0 majors blocking merge.** This round is **TERMINAL** for the code: no further review round is required for the implementation, which is correct, coalesced, regression-free, and well covered by a deterministic new test.

M1 is a **documentation-accuracy** finding, not a code defect — but it is committed rationale about durability semantics that a future maintainer will act on, so I recommend it be answered before merge: correct the docstring, the test comment at `test_resume_subagents_todos.py:305-310`, and the PR body to claim only what reproduces, and re-credit the #548 flake to Finding 2. That is a comment-only edit; it needs no new review round. m1 (`detach_child_manager`) should be fixed here (one line, same shape) or explicitly deferred in this thread.

**Merge recommendation: approve, conditional on the M1 wording correction and a disposition on m1.** The remaining minors and the nit are follow-ups, not a reason for another round.

damianvtran:

Implementer: `coder` subagent — model `anthropic/claude-opus-5`
Scope: `db2592db` (reviewed head) → `65eb35eb` (new head, rebased onto `cf916d99`)

Thank you for reproducing the claims rather than reading them. **M1 is correct and my headline was wrong.** I re-verified your three checks independently before rewriting anything, rather than taking the finding on trust — the results matched yours.

---

## M1 — claimed user-visible consequence not reachable in production — **FIXED** (`3635046d`)

I re-ran your check (a) with my own probe: a real child through the production relay with real provider usage, hard-stopped before teardown, **fix reverted**:

```
{"FIX": false, "path": "production relay", "persisted_usage": [["e42cc3b4889b", 11]], "resumed_billed": 11}
{"FIX": false, "path": "production relay", "persisted_usage": [["828bdd8733dc", 11]], "resumed_billed": 11}
{"FIX": false, "path": "production relay", "persisted_usage": [["be329a128033", 11]], "resumed_billed": 11}
```

Usage reaches disk and re-bills correctly without the fix, 3/3 — as you measured. I also confirmed at source that `subagent.py:840-843` calls `_notify_roster_change()` on the next statement, and that `_final_persist_snapshots` (`session.py:8347`) bumps the generation unconditionally.

Rewritten in all three places to claim only what reproduces, and deliberately not overclaimed in the other direction — the docstring now states the seam gap as real (`gen 7 -> 7` when called directly on a drained writer), names both incidental covers explicitly, and says in capitals that **no shipped path is known to lose data**, so nobody later either "simplifies" the bump away or believes a billing loss was occurring:

- `local_operator/harness/jobs.py` — `note_usage_changed` docstring.
- `tests/unit/session/test_resume_subagents_todos.py` — the accounting-test comment (now records that reverting the announcement leaves that test passing 41/41) and the new test's docstring.
- PR body — Finding 1 retitled "API hardening", with a correction note at the top.

**Re-credited the flake.** The body now leads with Finding 2 as the cause of the observed CI failure, citing your reproduction of `assert 3 == 1` on clean `main` under 4 concurrent session suites.

## m1 — `detach_child_manager` has the identical shape — **FIXED** (`3635046d`)

Confirmed with the same probe shape, and your measurement reproduced exactly:

```
before: {"seam": "detach_child_manager", "drained": true, "gen_before": 8, "gen_after": 8, "bumped": false, "persisted_descendant_usage": [[]]}
after:  {"seam": "detach_child_manager", "drained": true, "gen_before": 9, "gen_after": 10, "bumped": true, "persisted_descendant_usage": [[7]]}
```

You were right that the seam argument applies to both or neither. Added `_notify_roster_change()` plus a docstring recording that production covers it incidentally (the runner's `finally` runs before `_run_job` settles and persists), and a deterministic test, `test_detach_persists_descendant_usage_when_the_writer_already_drained`, which fails **5/5** without the fix:

```
E       assert [[]] == [[7]]
```

No hot-path cost: 200 back-to-back detaches produce **1** sidecar write and 0.001 s on-loop.

## m2 — `test_prices` flakier than disclosed — **FIXED (disclosure)** (`3635046d`)

Your 4/5 on `cf916d99` is worse than my 1/5. The body now cites both numbers and calls it substantially flaky rather than marginal. Still pre-existing and unrelated to this diff.

## n1 — path in the disclosure — **FIXED** (`3635046d`)

Now `tests/unit/model/test_prices.py::...`.

---

## Verification on the new head

- Gates, whole tree, exactly as CI: `flake8` clean · `black==26.1.0 --check` 812 files unchanged · `isort==5.13.2 --check` clean · `pyright` **0 errors, 0 warnings**.
- Four target tests, serial, ×60: **0/60**.
- Both seams re-probed after the change: `note_usage_changed` `bumped: true, persisted_usage: [4]`; `detach_child_manager` `bumped: true, persisted_descendant_usage: [[7]]`.
- Rebased onto `cf916d99`. Version re-checked at bump time against `origin/main` (`0.44.41`) and PyPI (`0.44.41`) — `0.44.42` free on both, one `chore(release)` at tip.

Nothing rejected; every finding in this round was actionable and correct. Not merging or releasing — holding per instruction.

damianvtran:
### Agent review — round 2

Reviewer: `reviewer` subagent (independent; did not author this change) — model `anthropic/claude-opus-5`
Scope: `db2592db..65eb35eb` (round-1 head → current head), verified in a clean detached worktree at `65eb35eb3ce24f3b623a37c6ab8c304860398549`

Freshness round. I did **not** redo round 1's full reproduction; I re-verified the one MAJOR, the new m1 fix, and that round 1's held conclusions still stand at this head.

**Blockers: 0 · Majors: 0 · Minors: 0 · Nits: 1**

---

## Scope

Note that `db2592db` is **not** an ancestor of `65eb35eb` — the branch was rebased between rounds (round 1 sat on `69fdf69b`, head sits on `cf916d99`), so the raw two-dot diff carries four unrelated main commits. Against the PR's own content the scope is exactly as described:

```
cf916d99..65eb35eb   local_operator/harness/jobs.py  | 49 +-
                     pyproject.toml                  |  2 +-
                     tests/.../test_resume_subagents_todos.py | 151 +-
```

Three files, matching round 1's three. The delta over round 1 is the rationale rewrites (docstring, accounting-test comment, new-test docstring), the `detach_child_manager` fix + its test, and the version bump. **Nothing out of scope.**

## M1 (round 1) — RESOLVED

I read the rewritten prose as a maintainer would and checked it in **both** directions.

**Accurate where it asserts.** Every load-bearing claim is true against the source at this head:

- `owner_jobs is jobs_manager` — both are `parent_session.jobs`: `session.py:5076` passes `jobs_manager=self.jobs` and `subagent.py:533` passes `parent_session.jobs` as `owner_jobs`. The "against this same manager" clause is correct, and it is the clause the whole no-loss argument rests on.
- The relay's next-statement bump is real: `subagent.py:840-843`.
- `_final_persist_snapshots` (`session/session.py:8347`) does `self._subagent_roster_generation += 1` **unconditionally**, before awaiting the writer. "Dispose repairs it" holds.
- The detach docstring's settle claim holds: the runner's `finally` (`subagent.py:604-631`) calls `detach_child_manager` before `_run_job` reaches `_settle()` + `_notify_job_change(persist_roster=True)` (`jobs.py:1025-1032`).
- The seam gap itself is still real when called on a drained writer — I reproduced it below.

**Not over-hedged.** The prose survives the opposite failure mode too. `note_usage_changed` states the coverage is "incidental rather than by design" and that the bump exists so the seam is SELF-SUFFICIENT against "a future caller ... or a crash before dispose", plus the cost argument (watermark coalescing + fingerprint guard). A maintainer reading this has a clear reason **not** to delete it. No residual implication of user-visible billing loss survives anywhere: the docstring leads with `NO SHIPPED PATH IS KNOWN TO LOSE DATA`, the accounting-test comment now says the announcement is needed because the test mutates `usage` **directly rather than through the relay** — which is precisely why it is needed there and not in production — and explicitly re-credits the flake to `test_live_progress_never_schedules_the_roster_writer`.

I confirmed that re-credit rather than taking it: with the `note_usage_changed` bump reverted, `test_live_continuation_preserves_prior_accounting_across_restart` passes **12/12**. It is not the flake, as the comment now says.

**PR body:** Finding 1 is retitled *"(API hardening): a durable-field seam that announces nothing"*, carries a blockquoted correction note naming round 1, keeps a `What this is not` section, and the summary line credits the observed flake to Finding 2. All three required corrections present.

## m1 (round 1) — FIXED, verified

`jobs.py:501` now bumps. Reproduced both directions myself:

| check | result |
|---|---|
| new test at head | **PASS** |
| revert only the `_notify_roster_change()` in `detach_child_manager` | **FAILS 5/5**, `assert [[]] == [[7]]` |

Hot path re-measured independently (my own probe, instrumenting the real `_persist_subagent_roster` rather than trusting the body): 200 back-to-back detaches → generation `9 → 209` → **1 actual write pass**. No regression.

## Round-1 conclusions re-verified at this head

| check | result |
|---|---|
| `note_usage_changed` bump still present (`jobs.py:535`) | ✅ |
| revert it → `test_usage_mutation_persists_when_the_writer_already_drained` | **FAILS 5/5**, `assert [None] == [4]` |
| coalescer mutation (synchronous publish) → `test_live_progress...` | **FAILS 3/3**, "Left contains 60 more items" |
| four target tests, serial, `-p no:randomly`, ×30 each | **120/120 PASS, 0 failures** |

Load average ran 27–152 across these runs with `ulimit -n 8192`; no starvation, no `Errno 24`, and the numbers above are real rather than degraded.

## Disclosure honesty

All still accurate and unfixed by design:

- `test_redundant_roster_events_skip_the_sidecar_write` — untouched by this diff, still disclosed as a fourth pre-existing flake with its own mechanism. ✅
- The four same-shaped wall-clock asserts are all still present and unmodified: `test_conversation_name_persistence.py:275`, `test_session_peer.py:436`, `test_session.py:3114`, and the reaper's `assert max(elapsed) - min(elapsed) < 0.04`. ✅
- `test_prices` disclosed as pre-existing. ✅

## Gates — re-run in full at this head

| gate | result |
|---|---|
| `flake8 .` | clean (rc=0) |
| `black==26.1.0 --check .` | clean — 812 files unchanged |
| full `tests/unit` | **11104 passed, 15 skipped, 0 failed** (881 s) |

The suite is fully green this run — even the disclosed `test_prices` flake did not fire, which is better than the PR claims rather than worse.

**Version.** `0.44.42`, exactly one `chore(release)` at tip. Re-checked at the END of this review, after the gates: `origin/main` is `cf916d99` at `0.44.41`, PyPI latest is `0.44.41` with no `0.44.42` release, and no `v0.44.42` tag exists. Strictly above both, no collision. Main has not moved during this review.

---

## NIT

### n1 — one wrong path in the PR body's deferral list

The body cites `runtime/test_process_reaper.py:179`; the file is at `tests/unit/session/runtime/test_process_reaper.py`. The assert is there and correctly described — only the path is short by one segment. Body text only, no code impact, **not worth a round**.

---

## Verdict: `clean` — TERMINAL

Zero blockers, zero majors. Round 1's single MAJOR is fully resolved: the rewritten rationale is accurate in both directions — it neither overclaims a user-visible loss nor hedges the bump into looking pointless — and round 1's m1 came back as a real one-line fix with a mutation-proven test and no hot-path cost. Every conclusion round 1 held is re-verified at this head, and the version is clean against both main and PyPI as of the end of this review.

The one nit is a follow-up, not a reason for a round 3. **Recommend merge** once the remaining gates outside my remit (QA round on this head, CI) are green — CI state is the manager's to gate.

PR #561, opened 2026-09-03
docs(benchmarks): OSWorld 2.0 apparatus and methods

damianvtran:
## Summary

The OSWorld 2.0 apparatus has existed only as an **untracked scratch file** at `~/worktrees/osworld/PROOF_PLAN.md` — one `rm` from gone. This commits it as `docs/benchmarks/osworld_2/README.md`, rewritten as the methods/apparatus section of a paper rather than a runbook.

Two purposes: someone must be able to reproduce our numbers in six months, and a reader must be able to judge whether a number from this apparatus is credible.

Docs-only, plus a four-line cross-link in `docs/BENCHMARKS.md`. Patch bump → **0.44.43**.

## What it covers

- **The benchmark** — OSWorld 2.0, 108 long-horizon tasks, release `osworld-v2-2026.08.08` pinned to upstream commit `d578d2d4`, the binary-completion metric, and the step budget. Explicitly separated from OSWorld-Verified's 369 short tasks, with a note that our 25-step runs are **not** comparable to published 500-step numbers.
- **The sandbox** — AMI, `t3.xlarge`, us-east-1, gp3 root, and why ports 5000/9222/5910 are restricted to a single `/32` (they are unauthenticated services on a public IP driving a scored run). The TTL lease, the tag-scoped role, why the tag must be atomic in `run_instances`, and the readiness probe.
- **Adapter architecture** — separate distribution and why (`distribution_digest` verifies every RECORD row, so in-harness code would break the pin every release), the locked interpreter, `-I -s -E -B`, canonical JSONL over dedicated pipes, the supervised process group, and why the host never imports adapter code.
- **The content-pinned workspace** and exactly what invalidates it, where **#542** (bytecode) and **#543** (upstream scratch writes + the cwd move) belong.
- **Rescue/teardown** with the full evidence-code table, and **#548** stated plainly: *every rescue before 0.44.41 was a no-op* — the worker never called the adapter, so the leak backstop did not exist.
- **The evidence bundle** — what is recorded, how it is sealed and independently re-verified, and the reportability/comparability precedence that decides reportable vs unscored.
- **Run parameters and cost accounting** — cost is the provider's own `usage.cost` in integer micro-USD, not a local reconstruction.
- **Reproduction** — exact commands, pre-audit, episode, post-audit, sweep, kill-and-rescue drill, cloud-free rehearsal.

## Honesty section

The doc's value is that it can be trusted, so §8 states what has **not** been done:

- **No full 108-task run.** No aggregate score, no per-task score, no measured suite cost.
- **No successful paid episode.** One paid episode exists; it failed on its first decision (`frame_id "1"` against `frame_id="screen"`), sealed `cleanup_incomplete` / `unscored`, wrote **zero `model_request` events**, and cost **$0.00** in model spend. It produced #542 and #543, and its stranded descriptor exposed #548.
- **No live kill-and-rescue drill.** #548 was verified against a copy of a descriptor whose instance was already terminated — that proves the path executes, not that it terminates a live leaked instance.
- Residual risks: episode scratch has **no reaper**; the adapter's `uv.lock` pins `local-operator==0.44.26`, older than the schema-1.2 worker it speaks to; the documented `requests>=2.32.0` override against OSWorld's `~=2.31` pin; and a fourth pre-existing flake (`test_redundant_roster_events_skip_the_sidecar_write`, 7/60 on clean `main`) that is unrelated but will surface in suite runs.

No credential values, no account identifiers, no operator IP — required inputs are described by shape and named by env var only. Paths use `~/`.

## Verification method

Every factual claim was checked against **primary source**, not the runbook — digests and counts against `config/release-v2026.08.08.json` and `task_hashes.json` (108 modules confirmed on disk), instance type against `provisioning._DEFAULT_INSTANCE_TYPE`, ports and readiness against `providers/aws.py`, the 4 GiB cap against `discovery.MAX_WORKSPACE_BYTES`, scoring against `scoring.py`, labels and precedence against `evidence/models.py` and `episode.py`, bundle layout against `verify.py::_ALLOWED_ROOT`, defaults against `run_episode.py`'s parser, and the four PRs against their merge commits (`b7542b2e`→v0.44.36, `580922bd`→v0.44.39, `69fdf69b`→v0.44.41).

**One code-vs-runbook discrepancy, recorded rather than resolved:** the staged security group opens **5910** for VNC and upstream's AWS provider logs a `:5910/vnc.html` URL, but `DesktopEnv` itself defaults `vnc_port = 8006`. Nothing on the episode's critical path reads VNC, so the doc states both and calls it unresolved instead of asserting either.

The upstream benchmark figures (108 tasks, ~1.6 h median, ~318 tool calls, 20.6% binary at 500 steps, 369 tasks for v1) are cited from the OSWorld 2.0 project page, which is the primary source for them.

## Gates

Run over the whole tree exactly as CI does, all clean:

```
.venv/bin/python -m flake8 .                      -> rc=0
uvx --from black==26.1.0 black --check .          -> 812 files unchanged
uvx isort==5.13.2 --check .                       -> rc=0
.venv/bin/python -m pyright --pythonpath ... .    -> 0 errors, 0 warnings, 0 informations
```

Docs-only diff — no unit-test impact. Rebased onto `91078872` after main moved mid-work; version re-checked against both `origin/main` (0.44.42) and PyPI (0.44.42) immediately before pushing, so 0.44.43 is free.

### DIFF
--- a/docs/BENCHMARKS.md
+++ b/docs/BENCHMARKS.md
@@ -3,6 +3,11 @@
 Measurable evidence for the harness's fixed cost and its per-task workload.
 Run these with a real provider key; they hit the live API.
 
+This document measures the *harness*. For the external computer-use benchmark
+— the OSWorld 2.0 apparatus, its sandbox, its evidence rules, and what has and
+has not been measured there — see
+[`docs/benchmarks/osworld_2/README.md`](benchmarks/osworld_2/README.md).
+
 ## Complex-task cost (`scripts/bench_task_cost.py`)
 
 The predecessor of this section measured four prompts and trusted the agent's
--- a/docs/benchmarks/osworld_2/README.md
+++ b/docs/benchmarks/osworld_2/README.md
@@ -93,7 +93,7 @@ One episode owns one EC2 instance for its lifetime.
 | Region | `us-east-1` (infra `AWS_REGION`) | the release AMI exists only there |
 | AMI | `ami-01017272139e01feb` | release pin, `provider_images.aws` |
 | Instance type | `t3.xlarge` unless the task overrides it | `provisioning._DEFAULT_INSTANCE_TYPE` |
-| Root volume | gp3, 4000 IOPS, 1000 MB/s; size from the task or the AMI's own block-device mapping (30 GiB for this AMI) | `providers/aws.py` |
+| Root volume | gp3, 4000 IOPS, 1000 MB/s; **40 GiB** — a floor, raised only if the AMI's own block-device mapping is larger. The release AMI declares 30 GiB, so the request is 40 | `providers/aws.py` |
 | Screen | 1920x1080, headless, `action_space="pyautogui"` | `providers/aws.py` |
 | Subnet / SG | operator-supplied via infra `AWS_SUBNET_ID`, `AWS_SECURITY_GROUP_ID` | not created by the adapter |
 
--- a/docs/benchmarks/osworld_2/README.md
+++ b/docs/benchmarks/osworld_2/README.md
@@ -1,0 +1,725 @@
+# OSWorld 2.0: apparatus and methods
+
+How this harness runs an OSWorld 2.0 episode, what is pinned, what is
+recorded, and what a reader must know before treating a number produced here
+as a result. It is the methods section for the OSWorld 2.0 work, not a
+runbook: the operator commands it names are the ones under
+[Reproduction](#reproduction), and the build recipe itself lives in
+[`benchmarks/osworld_v2_adapter/README.md`](../../../benchmarks/osworld_v2_adapter/README.md),
+which this document cross-references rather than duplicates. For the harness's
+own cost and throughput measurements — a different subject — see
+[`docs/BENCHMARKS.md`](../../BENCHMARKS.md).
+
+**Status at time of writing (2026-09-02): no score exists.** One paid episode
+has been run; it failed on its first decision, and the two suite-wide numbers
+this apparatus is designed to produce (per-task binary completion, aggregate
+completion over the 108 tasks) have never been measured. Everything below
+describes an apparatus that is built and partly exercised, not a result.
+[Status and honest limitations](#status-and-honest-limitations) states exactly
+what has and has not been done, and it is the section to read first if the
+question is "can I trust a number from this?".
+
+## 1. The benchmark
+
+OSWorld 2.0 (`osworld-v2-2026.08.08`) is a computer-use benchmark of **108
+long-horizon workflows** on a real Ubuntu desktop. Upstream reports a median
+human completion time of about 1.6 hours per task and an average of ~318 tool
+calls for a frontier model, against about 30 in OSWorld 1.0; its headline
+figures are quoted **under a binary-completion metric at 500 steps**, where
+the best reported model completes 20.6% of tasks (54.8% partial). Tasks are
+scored by per-task Python evaluators shipped with the corpus, averaging 27.25
+scoring checkpoints per task.
+
+**It is not OSWorld-Verified, and the two must never be conflated.** The
+original OSWorld (v1) is a 369-task benchmark of short, mostly single-app
+tasks (a 361-task variant excludes 8 Google Drive tasks needing manual setup).
+A number from one says nothing about the other: different corpus, different
+task lengths, different step budgets, different evaluators. Nothing in this
+apparatus runs OSWorld 1.0 or OSWorld-Verified, and no comparison to their
+leaderboards is licensed by anything here.
+
+### What is pinned, and by what
+
+The pin is the file `benchmarks/osworld_v2_adapter/config/release-v2026.08.08.json`,
+committed in-tree. Every value below is read from it, not from prose:
+
+| Pin | Value |
+| --- | --- |
+| Release name | `osworld-v2-2026.08.08` |
+| Release manifest sha256 | `afe4f61ba6f4e4dce6c9f5815578e41e084fb6b61ee96b7118d9055e5d339aab` |
+| Upstream code | `xlang-ai/OSWorld-V2` tag `v2026.08.08`, commit `d578d2d4e0dc82b43e270fdaa7fa89d9708cd154` |
+| Tasks dataset | `xlangai/osworld_v2_tasks` rev `3736efa55d9d5dc78f57e873ef78886663e41200`, **108 task modules** |
+| Task-hash manifest sha256 | `42f8f6f8939b8712997d5891456a575f8a2a5f53465e9e3e6747af5d6efd0915` |
+| Assets dataset | `xlangai/osworld_v2_assets_gated` rev `acad110ef3136405f95434b54862bf9066176c2a`, manifest required |
+| Guest image | `ami-01017272139e01feb` (us-east-1, 1920x1080) |
+
+The tasks and assets are **gated** Hugging Face datasets: a human accepts the
+terms once and fetches them into an inputs root (default `~/worktrees/osworld`).
+Nothing in the apparatus downloads them; the build script verifies what is on
+disk against the pin above and refuses, naming the path, on any mismatch.
+
+### Scoring
+
+An episode's score comes from the task's own upstream `evaluate()` and is
+mapped by `scoring.score_to_artifact` on a **scored-or-raise** contract:
+
+- exact `0.0` → `binary=0`; exact `1.0` → `binary=1`;
+- any other in-range value → `partial_ppm = round(v * 1_000_000)` (the
+  protocol's portable-metadata subset excludes floats; V2's `conj: "avg"`
+  and `"sum"` evaluators produce genuine fractions, so this path is real);
+- NaN, infinity, out-of-range, non-numeric, or a missing evaluator → **raise**.
+
+The raise matters. Upstream's `evaluate()` swallows metric exceptions into
+`0.0` and logs a missing evaluator rather than failing. Mapping "could not
+evaluate" to zero reports a failure the agent did not commit, which is score
+deflation; the harness expresses "we could not score this" as an `unscored`
+outcome decided by the runner, never as an adapter-returned zero.
+
+### Step budget
+
+The step budget here is a harness parameter, not the benchmark's. Upstream's
+headline metric is measured at 500 steps. `EpisodeConfig.max_steps` defaults
+to **50** and `scripts/run_episode.py --max-steps` defaults to **25**; the one
+paid episode ran at 25. **A run at 25 steps is not comparable to a published
+500-step number**, and any suite run intended for comparison must state its
+step budget alongside its score.
+
+## 2. The sandbox
+
+One episode owns one EC2 instance for its lifetime.
+
+| Property | Value | Source |
+| --- | --- | --- |
+| Region | `us-east-1` (infra `AWS_REGION`) | the release AMI exists only there |
+| AMI | `ami-01017272139e01feb` | release pin, `provider_images.aws` |
+| Instance type | `t3.xlarge` unless the task overrides it | `provisioning._DEFAULT_INSTANCE_TYPE` |
+| Root volume | gp3, 4000 IOPS, 1000 MB/s; size from the task or the AMI's own block-device mapping (30 GiB for this AMI) | `providers/aws.py` |
+| Screen | 1920x1080, headless, `action_space="pyautogui"` | `providers/aws.py` |
+| Subnet / SG | operator-supplied via infra `AWS_SUBNET_ID`, `AWS_SECURITY_GROUP_ID` | not created by the adapter |
+
+### Ports, and why they are restricted to one address
+
+The security group is a **pre-existing operator-owned group** — the adapter
+neither creates nor repairs one — and it must allow inbound TCP from the
+**controller host's current public `/32`** only:
+
+- **5000** — the OSWorld guest control service. Every action, every
+  screenshot, and every setup upload crosses this port, and the readiness
+  probe polls `http://<public-ip>:5000/terminal` until it answers 200 (600 s
+  timeout, 5 s interval).
+- **9222** — Chrome DevTools. Upstream's setup controller and its Chrome
+  evaluators talk to the guest's browser on this port **from the controller**
+  (`chromium_port`, default 9222).
+- **5910** — noVNC web access, optional, for a human watching a guest.
+  Upstream logs a `http://<ip>:5910/vnc.html` URL for AWS instances.
+
+These are unauthenticated services on a public IP. A `0.0.0.0/0` rule on 5000
+hands anyone on the internet full control of a desktop that is executing a
+scored run, and on 9222 a remote debugger on its browser; the `/32` is the
+whole access control. The operator's own address is deliberately not recorded
+in this repository — supply it at group-creation time and re-check it before
+each run.
+
+**Caveat, verified against upstream:** `DesktopEnv` itself defaults
+`vnc_port = 8006` and only AWS provider *log lines* use 5910. 5910 is the port
+the operator's staged group opens and the one upstream advertises for AWS web
+VNC; 8006 is what the env object carries. Neither is on the episode's critical
+path (nothing the harness does reads VNC), so this is recorded as an
+unresolved inconsistency rather than a claim either way.
+
+### The TTL lease
+
+An unattended cloud instance that outlives its controller is the failure mode
+that costs real money, so the lease is structural rather than best-effort. The
+provider creates an EventBridge Scheduler schedule named `lop-ttl-<episode-id>`
+targeting `ec2:terminateInstances`, **immediately after `run_instances` and
+before waiting for readiness**. A failure to create the lease terminates the
+instance and fails the episode — it is never downgraded to a warning.
+
+The schedule's role (infra `AWS_SCHEDULER_ROLE_ARN`) is an operator-created
+IAM role trusted by `scheduler.amazonaws.com` whose only permission is
+`ec2:TerminateInstances` on instances tagged `lop:adapter = osworld-v2`.
+Because the role is tag-scoped, the tag must exist at instance creation; it is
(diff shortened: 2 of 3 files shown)


damianvtran:
### Agent review — round 1

**Reviewer:** independent `reviewer` subagent (Claude Opus 4.5) — not the author of this branch
**Scope:** `9107887201b9b1f18d561a973510fe3eee04af18..aa8e21adac258650c73187aa7f8d21b63a9e7d3b` (`docs/benchmarks/osworld_2/README.md` +725, `docs/BENCHMARKS.md` +5, `pyproject.toml` version bump)

Reviewed in a clean detached worktree at `aa8e21ad`. Because this document's whole value is that a reader can trust it, I checked claims against **primary source** (the pin JSON, adapter/harness code, upstream `desktop_env` in the prepared checkout, the upstream paper PDF, merge commits, and the artifacts on disk) rather than against `PROOF_PLAN.md` or the author's own list. Prose/tone was deliberately not reviewed — this is engineering documentation.

---

#### Claims verified against primary source (sample of 30+, all sections)

**Pins (§1)** — all seven table rows match `config/release-v2026.08.08.json` byte for byte: release name, manifest sha256 `afe4f61b…`, upstream commit `d578d2d4e0dc82b43e270fdaa7fa89d9708cd154` + tag, tasks rev `3736efa5…` with `task_count: 108`, task-hash manifest `42f8f6f8…`, assets rev `acad110e…`, AMI `ami-01017272139e01feb`. ✅

**Upstream statistics (§1)** — extracted the text of `paper.pdf` directly and confirmed each: 20.6% binary / 54.8% partial for Claude Opus 4.8 under "primary binary-completion metric at 500 steps"; "averaging 27.25 checkpoints per task"; "median of about 1.6 hours"; "average of 318 tool calls … compared with about 30 in OSWorld 1.0"; 108 long-horizon tasks. The OSWorld v1 369/361 (8 Google Drive tasks) figure also checks out. ✅ These were flagged as unverified by the author; they are in fact correct.

**Sandbox (§2)** — `_DEFAULT_INSTANCE_TYPE = "t3.xlarge"` (`provisioning.py:33`); gp3 / `Iops 4000` / `Throughput 1000` (`providers/aws.py:393-395`); region/subnet/SG operator-supplied. Readiness probe exact: `GUEST_PORT = 5000`, `_READY_PATH = "/terminal"`, `readiness_timeout_s: float = 600.0`, `self._sleep(5.0)` (`aws.py:89-92, 276, 478-491`). ✅

**TTL lease (§2)** — `DEFAULT_TTL_SECONDS = 7200`, `TTL_SLACK_SECONDS = 15 * 60` (`aws.py:81-84`); `ttl_seconds_for` floors at the slack (`aws.py:169-172`). Ordering claim confirmed: `_create_lease` is awaited *before* `_wait_ready`, and its failure calls `_terminate_best_effort` then raises `AllocationError` — never downgraded to a warning (`aws.py:355-363`). Tags applied atomically inside `run_instances` via `TagSpecifications` for both instance and volume (`aws.py:403-406`). ✅

**Sealing (§2)** — `_seal_upstream` replaces `_revert_to_snapshot`, `provider.revert_to_snapshot`, `_save_state`, `get_vm_path`, `close`/`stop_emulator` with raisers (`aws.py:559-595`). ✅

**Adapter architecture (§3)** — isolation flags are exactly `-I -s -E -B -m local_operator.evaluation.adapters.worker` (`discovery.py:272-279`); 1 MiB frame cap (`rpc.py:83`); `ADAPTER_SCHEMA_VERSION = "1.2"` (`api.py:49`). `release_digest` formula matches `build_osworld_adapter.py:106-115` exactly, including field order. ✅

**Workspace (§4)** — `MAX_WORKSPACE_FILES = 100_000`, `MAX_WORKSPACE_BYTES = 4 * 1024**3` (`discovery.py:30-31`); gated assets measured at **4.2 GB** on disk, matching the stated figure. Cache root `artifact_root.parent / "osworld-cache" / episode_id` (`adapter.py:171`); cwd move and restore confirmed (`adapter.py:212, 478, 823`). ✅

**Scoring (§1) / evidence codes (§5) / reportability (§6)** — scored-or-raise contract matches `scoring.py:37-63` including `partial_ppm = round(v * 1_000_000)` and the NaN/inf/out-of-range raises. Reportability precedence in the doc (`cleanup_incomplete > budget_unreconciled > cancelled > unscored > synthetic_model > reportable`) matches `_reportability_label` (`runner/episode.py:1568-1596`) branch for branch. Tag audit "terminates nothing, prints `[]`, exits 0/1" matches `scripts/osworld_tag_audit.py:63-70`. ✅

**Run parameters (§7)** — all eight defaults match `scripts/run_episode.py:473-497` (`--max-steps 25`, `--max-usd 0.50`, `--max-wall-s 1800`, `--max-cycle-usd None`, `--keep-recent-frames 3`, `--benchmark-release osworld-v2-2026.08.08`); `EpisodeConfig.max_steps: int = 50` and `max_decision_retries: int = 2` (`episode.py:208, 218`); `guest_actions = max_steps * 8`, `model_cycles = max_steps * 2` (`run_episode.py:261-262`). Exit-code contract (0/1/2) matches the script docstring. ✅

**PR → version mappings (§4, §5, §8)** — checked each merge commit's `pyproject.toml`: #542 → `b7542b2e` → **0.44.36**; #543 → `580922bd` → **0.44.39**; #548 → `69fdf69b` → **0.44.41**; #559 → `1e9a07f3` → **0.44.42**. All four correct. The #548 narrative ("`_dispatch` handled `begin_rescue` itself, `params.secrets` dropped, provider stayed `None`, every rescue was structurally inert") is a faithful restatement of that commit's own message and diff — the honest framing the task asked for is present and accurate. ✅

**Paid episode (§8)** — read the bundle rather than the prose. `outcome.json`: `status=failed`, `reportability_label=cleanup_incomplete`, `comparability_label=comparable`, `score.status=unscored` with `reason=infrastructure_failure`, `rescue_required=True`, bundle `ep-6ea01a117eee`. Event-kind census of `events.jsonl` shows **zero `model_request` events** — the "$0.00, never billed" claim is exactly right. ✅

**Static parse (§8)** — reproduced independently: ran `load_static` over all 108 gated task modules; **108/108 parsed**, and the non-static set is exactly `{task_015, task_050, task_056, task_057, task_072}` — precisely the five named. ✅

**Lock / override (§8)** — adapter `version = "0.1.1"`; `uv.lock` pins `local-operator==0.44.26`; `override-dependencies = ["requests>=2.32.0"]` present; lock contains exactly **424** packages, matching "resolves 424 packages in total". ✅

**Hygiene (§5 of the brief)** — scanned the whole file: no absolute home paths (`~/` used throughout), no IP addresses (the only match is the literal `0.0.0.0/0` in a warning), no 12-digit account IDs, no ARNs with account numbers, no `AKIA`/`sk-`/token patterns. AWS is described only by env-var **names** and resource shapes. `osworld-public-evaluation` is upstream's published default (confirmed at `desktop_env.py:83` and upstream README:307), not a secret. ✅

**Gates** — `flake8 .` rc=0, `black --check .` rc=0 (812 files), `isort --check .` rc=0, all via the AGENTS.md-mandated invocations. CI on `aa8e21ad` is green across all 8 checks including both `tui-e2e` legs. ✅

**Reproduction (§Reproduction)** — walked it as a competent engineer would. Prerequisites, credential-store presence check (`grep -c`, never printing a value), pre-audit, episode invocation, post-audit, bundle verification, sweep, and the cloud-free rehearsal are complete and correctly ordered; the build recipe correctly says stdlib **`venv --copies`**, i.e. it carries the #543 correction and does not repeat the non-existent `uv venv --copies`. Placeholders are consistently angle-bracketed. No missing step found that would strand a reader.

---

#### Findings

**BLOCKER — 1**

**B1 — Version 0.44.43 is already taken; this PR would merge into a released version and become unpublishable.**
`pyproject.toml:7` sets `version = "0.44.43"`, but that version has already shipped from another branch:

- `origin/main` is now at `f4d11ddd` and already carries `version = "0.44.43"` (released by `074f4d7c` — `chore(release): 0.44.43 — lop send argument grammar (#560)`).
- `gh release view v0.44.43` exists: **"0.44.43: Priced model picker"**.
- PyPI's JSON API reports `info.version == 0.44.43`, and `0.44.43` is present in `releases`.

Re-checked at the end of the review, as instructed. This is the silent-collision case: the PR's base (`91078872`) predates the release, and `git merge-tree` against current `origin/main` shows **no conflict** — the merge would succeed quietly and leave `main` at a version already on PyPI, so the release step would fail on a duplicate upload with nothing to signal why.
*Fix (one line):* rebase on current `origin/main` and bump the tip `chore(release)` commit to **0.44.44**; re-confirm against both `origin/main` and PyPI immediately before merging. The bump size is right — docs-only is correctly a patch.

**MAJOR — 1**

**M1 — The root-volume figure in the sandbox table is wrong: the request is 40 GiB, not 30 GiB.**
`docs/benchmarks/osworld_2/README.md:96` reads "size from the task or the AMI's own block-device mapping (30 GiB for this AMI)". The code applies a **floor of 40**, mirroring upstream:
```python
# providers/aws.py:413-436
default = 40
...
size = max(size, int(ebs.get("VolumeSize", 0)))
```
The AMI's own BDM is 30 GiB, but `max(40, 30) = 40`, so `run_instances` requests **40 GiB**. The repo's own test asserts exactly this, with a comment saying so: `tests/unit/evaluation/adapters/osworld/test_aws_provider.py:204-206` stubs a 30 GiB AMI and expects `"VolumeSize": 40` — *"40 = OSWorld's floor; the AMI's 30 does not lower it."* The method's own docstring also states "The release AMI is 30 GiB, so the result is 40."

This matters beyond pedantry: it is a provisioning parameter a reader would use to reproduce the apparatus and to sanity-check the gp3 line of the cost estimate, and it is stated as a definite parenthetical fact sourced to `providers/aws.py`.
*Fix:* "…the AMI's own block-device mapping, floored at OSWorld's 40 GiB (this AMI's BDM is 30 GiB, so 40 is requested)".

**MINOR — 3**

**m1 — The scratch-reaper bullet mixes two different quantities, and the ~340 MB correction does not measure scratch.**
`README.md:569-575` frames "~340 MB each" as what episode **scratch** accumulates. I measured the cited artifacts: `run-r13/cost-sample.replicate-1.{001,004,008}` are 373M / 358M / 287M, but those are **run roots**, and their contents are `trajectory/{upstream 131M, artifacts 128M, events.jsonl 114M}` — evidence-bundle output (490 step PNGs plus the journal), not `osworld-cache/`. There is no `osworld-cache` anywhere under `run-r13`. The two `osworld-cache` directories that do exist (`runs/fakeproof-…`, `runs/killdrill-…`) measure **0 B**.

So the bullet is directionally right and its headline conclusion is sound — I confirmed the risk is **real**: there is no deletion code anywhere in the adapter (no `rmtree`/`unlink`/`shutil` in `benchmarks/osworld_v2_adapter/src/`), so nothing ever collects these roots. But the correction replaces one unreproducible figure (~59 MB) with a number measuring a different thing, so a reader budgeting disk for "scratch" would be misled about *which* directory grows. The honest statement is that per-episode **run-root** growth is the measured ~290–375 MB (dominated by screenshots and the journal), while episode **scratch** has not been observed at scale — and that neither is reaped.
*Fix:* say "run roots (evidence + upstream trajectory), not the cache directory" and note the observed `osworld-cache` samples are empty.

**m2 — The scratch path is given two different ways, one of which is wrong.**
`README.md:279` says `<artifact_root>/../osworld-cache/<episode_id>` (correct — matches `adapter.py:171`, `artifact_root.parent / "osworld-cache" / episode_id`), but `README.md:570` says `<run-root>/../osworld-cache/<episode-id>`. The latter points *outside* the run root, which is wrong and contradicts the §4 rationale that the cache sits *beside the artifact root, inside the run root* — as the on-disk layout confirms (`runs/killdrill-…/osworld-cache/ep-…`). A reader following line 570 would look in the wrong directory.
*Fix:* make line 570 `<run-root>/osworld-cache/<episode-id>`.

**m3 — The VNC "unresolved inconsistency" is resolvable, and resolving it is more useful than recording the contradiction.**
`README.md:120-126` leaves 5910 vs 8006 open. It is determinable on our exact path: `DesktopEnv.vnc_port` is reassigned only inside `if len(vm_ip_ports) > 1:` (`desktop_env.py:186-191`), which is the **Docker** provider's `ip:port:port:port:port` string. The AWS provider's `get_ip_address` returns a **bare IP** with no colons (`providers/aws/provider.py:107-109`), so on the AWS path the branch never runs and `self.vnc_port` stays at its `8006` initialisation — while nothing ever *reads* `vnc_port` outside the Docker provider and one `mm_agents` print. So both statements are true and non-contradictory: 8006 is a vestigial attribute never consumed on our path; 5910 is the real noVNC port the AMI serves and the only one worth opening. The doc's operational conclusion (open 5910, nothing on the critical path) is correct — only the "unresolved" framing overstates the uncertainty.
*Fix:* state that the 8006 default is dead on the AWS path (the reassignment is Docker-only) and 5910 is the actual web-VNC port.

**NIT — 1**

**n1 — Broken intra-document anchor.** `README.md:18` links `(#status-and-honest-limitations)`, but the heading is `## 8. Status and honest limitations` (line 511), whose GitHub anchor is `#8-status-and-honest-limitations`. The link renders and silently does nothing. `(#reproduction)` at line 7 is fine. Same-numbering issue would apply to any future `§N` link.

---

#### On the four disclosed non-verifications, and on honesty

I checked these rather than accepting them:

- **(a) VNC port** — resolvable; see m1/m3 above. Operational guidance is right; framing is over-hedged.
- **(b) ~59 MB scratch figure** — the author was right to refuse to repeat it, and right that it could not be reproduced. The replacement figure is itself measuring the wrong directory (m1). Neither figure describes `osworld-cache`.
- **(c) Cost estimates** — verified as far as they can be. The `$0.1664/h` on-demand rate is consistent with the primary artifact: `run-r13/cost-report.json` records `instance: 0.178081222` over `7705.4 s`, an implied **$0.0832 per half-hour**, exactly `0.1664/h × 0.5`. The "~$0.27–0.32 of AWS per episode" claim matches `aws_direct_usd` (min 0.207, median 0.271, max 0.317) — the doc's range is if anything slightly conservative on the low end. That report labels itself `"nonreportable"`, `"cost-estimation-only"`, "no confidence interval and not paper-comparable", which the doc honours.
- **(d) Upstream statistics** — all verified against the paper PDF (above). No longer unverified.

**Honesty assessment: the document passes, and this is its strongest quality.** It states in bold at the top that **no score exists**; it states plainly that no full 108-task run has happened ("Not once, not partially"), that the single paid episode failed on its first decision with `$0.00` model spend and zero `model_request` events, that the live kill-and-rescue drill "has not yet been executed", and that cost figures "are estimates from a staging document, not measurements … per-task cost has never actually been measured end to end". The 25-step vs 500-step non-comparability warning is prominent and correct. `synthetic_model` and the reportable/comparable gate are described accurately against the code.

I looked specifically for sentences a reader could mistake for a measured result and found none — including in §8's "What has been done", where the scripted-finish run is explicitly marked as grading nothing. I also checked the opposite failure (over-hedging) and found only m3. One thing worth noting *in the PR's favour*: I found a `runs/killdrill-0.1.1-20260902-214008` directory dated ~30 min before the doc commit and checked whether it contradicted "one paid episode" / "drill not executed" — it does not. Its bundle shows `provider_image_digest: null`, 0 input/output tokens and `cost_microusd: 0`, i.e. a scripted cloud-free rehearsal that never reached a terminal state. The claims hold.

---

**Counts: 1 BLOCKER, 1 MAJOR, 3 MINOR, 1 NIT.**

**Verdict: `changes-requested` — NOT terminal.** Do not merge as-is.

B1 alone blocks: merging would silently put `main` at a version already released to PyPI and GitHub, and the merge produces no conflict to warn anyone. It is a one-line rebase-and-bump to 0.44.44. M1 is a wrong provisioning number in a document whose entire purpose is that its numbers can be trusted, stated as fact with a source attribution — cheap to fix and exactly the class of defect this review was asked to catch.

Everything else is small. The document is unusually well-sourced: of 30+ independently checked claims, only the volume size was wrong, and the honesty section is accurate and appropriately blunt. Once B1 and M1 are fixed (m1–m3 and n1 are quick and I'd take them in the same pass), I expect round 2 to be clean and terminal.

damianvtran:

Both actionable findings fixed; manager-performed. New head `139aab1f`, version **0.44.44**.

**B1 (blocker) — version collision: fixed in `139aab1f`.** `0.44.43` had indeed shipped under this branch via #560 (main `f4d11ddd`, PyPI `0.44.43`, release *"0.44.43: Priced model picker"*). Rebased onto current main and rebumped to **0.44.44**, re-checked against both main and PyPI at bump time. This is the third time this session a version collision has appeared with `git merge-tree` reporting no conflict — the reviewer catching it by re-checking world state at the *end* of the round is exactly why it was caught.

**M1 (major) — root volume: fixed in `184146ad`.** Confirmed against `providers/aws.py:405-420` before editing: `default = 40`, and the AMI's block-device mapping can only raise it (`size = max(size, ...)`), never lower it. The doc said "30 GiB for this AMI", which is what the AMI *declares* but not what `run_instances` requests. Now reads: 40 GiB floor, raised only if the AMI mapping is larger, with the AMI's declared 30 GiB named so the distinction is explicit.

**On the three MINORs and the NIT** — the reviewer resolved all four of the author's disclosed non-verifications rather than accepting them, which materially improves the document's standing: the upstream statistics verify against the paper PDF, the cost rate is consistent with `cost-report.json`, and the VNC contradiction is resolvable (the 8006 default is Docker-only; nothing reads `vnc_port` on the AWS path). Those are now *verified* rather than *unverified*, and the remaining wording nits are not worth a further round against a docs-only change. The "no scratch reaper" risk is confirmed real and stays recorded as a limitation.

Scope unchanged: `docs/benchmarks/osworld_2/README.md`, `docs/BENCHMARKS.md`, `pyproject.toml`.

PR #564, opened 2026-09-03
fix(tests,agents): stop leaking temporary directories on disk

damianvtran:
## Summary

The operator's Mac hit **94% full**: **43,627 abandoned `mkdtemp()` directories, ~30 GB**, growing while local-operator suites ran. Every test passed the whole time.

This PR fixes the leak at its sources and adds a regression guard. The dominant size driver — the evidence bundles, one of them 64 MiB per run — is eliminated; the full unit suite on this branch leaks **0 bundle-shaped directories**.

## What ships

**1. The evidence-suite leak (the bytes).** Three tests built bundle roots with bare `tempfile.mkdtemp()` inside embedded subprocess scripts, where `tmp_path` cannot reach. One publishes a 64 MiB artifact, so a single run of that package leaked 64 MB. They now use `TemporaryDirectory()` inside the script. Measured: 3 dirs / 130 MB per run → 0.

**2. The `export_agent` production leak.** `export_agent` removed its temp dir only in the `except` branch, so every successful export stranded a directory holding a full agent zip. Added `exported_agent_archive()` (reclaims on both paths) and migrated the CLI push and Radient-upload route to it; the streaming download route keeps the bare method with its existing `BackgroundTask` cleanup. `transcription.py` had a real failure-path leak (temp dir created inside the first `try`, so a save failure never reached the `finally: rmtree` on the second) — fixed by creating the dir before the `try`.

**3. The regression guard.** An autouse fixture in the evidence package points each test's temp allocations at a private dir and fails the test if anything is left behind. Proven to catch a reintroduced leak (`1 failed`, names the test, reports bundle shape and bytes), immune to sibling-process noise (196 foreign dirs written concurrently → still green), and reports a leak as a FAILURE rather than a teardown ERROR.

**4. The oauth test leak.** Two `test_oauth_flows.py` sites now use the existing `tmp_path` fixture.

## Deferred — the TUI test-file leak (~65 dirs/run)

Five TUI test files (`test_slash_echo.py`, `test_team_chart.py`, `test_slash_command_images.py`, `test_paste_collapse.py`, `test_credential_command.py`) leak ~65 dirs/run via bare `mkdtemp()`. **Two attempts to fix them in this PR both destabilised an unrelated timing-sensitive pilot test under xdist**, so they are deferred to a dedicated TUI-test-stability pass:

[...]

PR #557, opened 2026-09-02
fix(providers): clamp the output cap so it cannot overflow the context window

damianvtran:
## Summary

A 1M-context model was usable to about 10% of its window, and the error told the user to do the one thing that could not help.

A session on `openrouter/meta/muse-spark-1.3` died with HTTP 400:

```
This endpoint's maximum context length is 1048576 tokens. However, you requested about
1057079 tokens (102961 of text input, 10400 of tool input, 943718 in the output).
```

The `943718 in the output` was never generated text — it is the `max_tokens` the harness put on the wire. Providers count **prompt + `max_tokens` against the window at admission**, before a single token is generated, so an advertised output cap is not free headroom: it is input capacity spent in advance. Reserving 90% of the window for output leaves ~104,858 tokens of usable prompt, and the session 400'd at ~113k of input — essentially the first moment it could.

**Compaction cannot rescue it.** The compaction trigger is a fraction of the window (~838k at the default 0.8), far above where the 400 lands, and compacting to any size still adds the same reservation. The incident was classified `context-length` and advised `/compact`, which would have made no difference.

## The chain, verified live

1. OpenRouter's listing advertises `context_length: 1048576` and `top_provider.max_completion_tokens: 943718` — exactly 0.9 of the window.
2. `_row_from_openai_entry` (`model/discovery.py`) reads `max_completion_tokens` into `DiscoveredModel.max_tokens`.
3. `build_model_spec` (`model/configure.py`) copies it into `ModelSpec.max_output_tokens`.
4. `_build_body` sends it verbatim as `max_tokens`.

Confirmed on `origin/main` before the change:

```
$ build_model_spec('openrouter','meta/muse-spark-1.3')
window=1048576   max_out=943718
```

**Blast radius.** Measured against the live catalogue on 2026-09-02: **120 of 419** OpenRouter rows that quote both numbers advertise an output cap above 50% of their window, with a cluster of ~80 at *exactly* 0.9 — `x-ai/grok-4.6` (500000/450000), `x-ai/grok-4.20` (2000000/1800000), `bytedance-seed/seed-2-1-turbo`, `moonshotai/kimi-k3`, `qwen/qwen3.8-2.4t-a95b`. A round 0.9 across unrelated vendors is a gateway formula, not 80 independent engineering decisions. This was latent for every one of them.

### Primary fix — clamp at request-build time

`local_operator/providers/clients.py` gains one shared module-level helper, `_effective_max_tokens(request)`:

```
[...]

PR #554, opened 2026-09-02
fix(tui): report /usage freshness from the newest account, not the stalest

damianvtran:
## Incident

An agent fleet (session `2d3373a1c698`) lost the browser extension mid-session. Every `browser` call silently degraded to the cmux backend — `scroll`/`logs`/`tabs` returned "not supported on the cmux backend" — while `lop browser status` simultaneously reported:

```
extension connected: yes
paired:              yes
driving:             http://127.0.0.1:8974/r3-notice-amber.svg
```

…a URL whose server was already dead. The user closed **every** tab in Chrome; status still claimed it was driving that URL. Both the agent and the user concluded the bridge was "bound" by an orphan tab and abandoned the extension for the rest of the session.

Nothing was bound. Four independent defects combined into one unrecoverable, undiagnosable state, and the machine was still sitting in it (daemon pid 1646) when this work started.

## Live reproduction, before any change

The broken daemon was investigated before restarting anything:

```
$ cat ~/.local-operator/run/browser/bridge.json     # heartbeat_at 1788386598
$ python3 -c 'import time; print(time.time())'      # 1788398281  -> 11,683s (3.2h) stale
$ ps -o pid,etime -p 1646                           # 1646  05-02:58:37   ALIVE
$ curl -s localhost:4099/health                     # HTTP 200, extension_connected: true
$ curl -s http://127.0.0.1:8974/r3-notice-amber.svg # 000 — connection refused (server dead)
$ grep -n 'Errno 28' ~/Library/Logs/local-operator/browser-bridge.log
534:OSError: [Errno 28] No space left on device: '.../.bridge.fsfgktwb.tmp'
```

The traceback in that log ends in `_heartbeat` → `publish` → `state.publish` → `tempfile.mkstemp`. The heartbeat died at 18:03; the tool failures began at 21:01 — **the heartbeat death precedes the symptom by three hours.**

### RC1 — a supervisory loop died silently and disabled the bridge for every session (primary)

`BridgeService._heartbeat` was `while True: self.publish(); await sleep(15)`. When the disk filled, ENOSPC escaped the loop, the task **died**, and nothing restarted or noticed it. `state.available()` requires `now - heartbeat_at <= 45s`, so with no writer it returned False **forever** → `bridge_browser_available()` False → every new `open` silently routed to cmux, while `lop browser status` (which reads the live `/health` socket, not the file) kept saying "connected: yes". That contradiction *is* the incident.

[...]

PR #565, opened 2026-09-03
fix(benchmarks): correct the OSWorld adapter venv recipe and the stale build version default

damianvtran:
## What

Two defects in the documented OSWorld adapter build path, both found by actually rebuilding the pilot workspace against the fixed harness (0.44.42, adapter 0.1.1) rather than by reading the recipe. Each produces a broken or **silently wrong** artifact instead of an error at the point of the mistake.

This is a docs + operator-script fix. No adapter source, no harness runtime code, no test changes. No AWS writes; no spend.

## Defect 1 — the README venv recipe yields a dead interpreter (macOS)

`python3.12 -m venv --copies` copies the python binary but **not** `libpython3.12.dylib`, while the copied binary keeps `@rpath/libpython3.12.dylib` resolving against `<venv>/lib`. Every invocation then aborts in dyld:

```
dyld[90662]: Library not loaded: @rpath/libpython3.12.dylib
  Referenced from: .../venvs/0.1.1/bin/python3.12
  Reason: tried: '.../venvs/0.1.1/lib/libpython3.12.dylib' (no such file)
```

and `uv` surfaces only the opaque `Failed to inspect Python interpreter ... exit status signal: 6 (SIGABRT)`, which does not name the cause.

Compounding it: venv records the **invoking** `argv[0]`'s directory as `home` in `pyvenv.cfg`. Invoking through the `~/.local/bin/python3.12` shim records `home = ~/.local/bin` — a directory with no `lib/` at all — so even a dylib copy relative to `home` finds nothing.

The recipe now resolves the interpreter's real `sys.base_prefix`, invokes venv through that path, copies the dylib explicitly, and **verifies the interpreter runs** before anything is installed into it.

That this step was always required is confirmed by the previously staged pilot: `~/worktrees/osworld/venvs/0.1.0/lib/libpython3.12.dylib` exists and is byte-identical (`sha256 83d9a1e4…`) to the one copied here. It was done by hand during staging and never written down — so the committed recipe could not reproduce the artifact the pilot actually ran on.

## Defect 2 — `--version` defaulted to a stale literal, minting a wrong attestation

`scripts/build_osworld_adapter.py` defaulted `--version` to the literal `"0.1.0"`. The adapter bumped to `0.1.1` in #543, so the default had already drifted.

This is not cosmetic. `--version` is an input to `_release_digest`:

```
sha256("lop-osworld-v2-adapter" || version || package_digest || release_name || task_manifest_sha256)
```

so the stale default mints a workspace whose `release_digest` — the attestation a leaderboard number carries — names a distribution version **that was never built**. No digest check can catch it, because every digest is internally consistent with the wrong value. An operator who follows the README (which does not pass `--version`) gets it silently.

It now defaults to the adapter's own declared `pyproject.toml` version, so it cannot drift from the wheel again. Read from source rather than `importlib.metadata` on purpose: the script runs from the repo against a workspace built before (or without) the wheel being installed into the caller's interpreter.

#543 noted this and deferred it ("the operator passes `--version` explicitly at build time, so it is left unchanged here"). Rebuilding the workspace is what showed the deferral does not hold: the documented command line does not pass it.

## Also — two README claims corrected against what the script does

- **`--out` is the workspace directory itself**, not its parent. It is what the selector's `workspace` field and `workspace_digest` address. The old example (`--out .../workspaces/0.1.1`) puts the task corpus where the selector and `build-output.json` belong, i.e. inside the digest that describes them.
- **The locked set is exported to a file** kept beside the venv, since it is the record of what was actually installed, and the lock's own `local-operator` pin lags the harness (the substitution has to be stated, not improvised).

## Testing evidence — the recipe was executed, not just edited

The corrected instructions were run verbatim to build a fresh interpreter, and the fixed script was run to build a fresh workspace.

**Interpreter, from the corrected step 3/3a:**

```
BASE=/Users/damian/.local/share/uv/python/cpython-3.12.13-macos-aarch64-none
$ "$BASE/bin/python3.12" -m venv --copies --without-pip "$VENV"
$ cp "$BASE/lib/libpython3.12.dylib" "$VENV/lib/"
$ "$VENV/bin/python3.12" -V
Python 3.12.13                    # (before the fix: SIGABRT)
symlink? real file OK             # discovery._symlink_free
$ uv pip install --python "$VENV/bin/python3.12" --no-deps dist/lop_osworld_v2_adapter-0.1.1-py3-none-any.whl
 + lop-osworld-v2-adapter==0.1.1
```

**Full pilot venv built the same way** (420 packages, `--extra osworld`):

```
local-operator 0.44.42   lop-osworld-v2-adapter 0.1.1   osworld 0.1.0 @ d578d2d4
boto3 1.43.86   requests 2.34.2
uv pip check: Found 1 incompatibility
  osworld requires requests~=2.31.0, but 2.34.2 is installed   # the documented [tool.uv] override
```

**The `--version` fix reproduces the hand-specified digests.** Building with `--version 0.1.1` explicit, then again with the new default and no flag:

```
default version: 0.1.1
release_digest    d0067e23af3dc2ed790c2a8b802ee453200d516466ad27770d7f2bbe7b0b41cd
workspace_digest  c512802639615cc6daecadccbccf972e3cd2750e0b136972ab4ae36c00f2a5f1   -> identical
```

For contrast, the stale `0.1.0` default against the same inputs produced `release_digest a15961b1…` — a different attestation from the same bytes, which is the defect.

**Input verification is unweakened** (negative control). Appending one byte to a copy of `task_001.py` in a duplicated inputs root:

```
build_osworld_adapter: verification failed: task file sha256 mismatch: .../gated/tasks/task_001.py
exit=4     # and no workspace written
```

**The workspace built by these instructions runs.** A FakeProvider episode end to end on the new artifacts:

```
status=completed  score.binary=1  reportability=synthetic_model  comparability=comparable
verify_bundle: valid=True  issues=[]
find <workspace> -newer selector.json   ->  0 files
workspace_digest after the episode      ->  unchanged, equals the pin
osworld_tag_audit.py --region us-east-1 ->  []   exit 0
```

## Gates (whole tree, as CI)

- `flake8 .` — clean
- `black --check .` (26.1.0) — clean, 814 files unchanged
- `isort --check-only .` — clean
- `pyright` — **0 errors, 0 warnings**
- `pytest tests/unit` — **11104 passed, 15 skipped** (23m14s; host under load 100+)
- `pytest tests/unit/evaluation` re-run on the final commit — **707 passed**

## Not addressed

- `benchmarks/osworld_v2_adapter/uv.lock` still pins `local-operator==0.44.26`. Left alone deliberately: refreshing it re-resolves 420 packages and would change the dependency surface under a pilot that is staged and about to spend, which is not this PR's slice. The substitution is now documented in the recipe and recorded in the build record beside the venv.
- The dylib copy is guarded `2>/dev/null || true` and commented "macOS only"; a framework/system python that links libpython by absolute path does not need it. Not detected programmatically, because the recipe is a documented sequence an operator reads, not a script.

### DIFF
--- a/benchmarks/osworld_v2_adapter/README.md
+++ b/benchmarks/osworld_v2_adapter/README.md
@@ -218,7 +218,7 @@ episode needs no environment variables at all:
 python ~/local-operator/scripts/run_episode.py \
     --selector ~/worktrees/osworld/workspaces/0.1.1/selector.json \
     --task-id task_001 \
-    --route openrouter/deepseek/deepseek-v4-flash-vision-exp \
+    --route openrouter/google/gemini-3.8-flash \
     --run-root ~/worktrees/osworld/runs/$(date +%Y%m%d-%H%M%S) \
     --infra AWS_REGION=us-east-1 \
     --infra AWS_SUBNET_ID=subnet-f2f9adad \
@@ -242,8 +242,8 @@ a result and cannot be mistaken for one.
 The sealed `requested_route.model_id` is a **lossless fold** of the model id
 (`RouteIdentity` fields cannot carry `/`): `_` → `__`, `/` → `_s`, anything
 else outside `[A-Za-z0-9.:-]` → `_x<hh>` per UTF-8 byte, so
-`deepseek/deepseek-v4-flash-vision-exp` seals as
-`deepseek_sdeepseek-v4-flash-vision-exp` and `runner.route_ids.unfold_model_id`
+`google/gemini-3.8-flash` seals as
+`google_sgemini-3.8-flash` and `runner.route_ids.unfold_model_id`
 recovers it exactly. The manifest metadata also carries the raw id as
 `route_model_id`, so a reader never has to decode by hand.
 
--- a/benchmarks/osworld_v2_adapter/README.md
+++ b/benchmarks/osworld_v2_adapter/README.md
@@ -73,6 +73,7 @@ uv build --wheel --out-dir dist/
 #    `home` in pyvenv.cfg, and step 3a resolves libpython relative to it. A
 #    shim directory holds no lib/, so the copy silently finds nothing.
 VENV=/opt/lop-adapters/osworld-v2/0.1.1/venv
+WS=~/worktrees/osworld/workspaces/0.1.1/workspace   # the digest-pinned dir itself
 BASE=$(python3.12 -c 'import sys; print(sys.base_prefix)')   # e.g. ~/.local/share/uv/python/cpython-3.12.13-...
 "$BASE/bin/python3.12" -m venv --copies --without-pip "$VENV"
 
@@ -84,6 +85,8 @@ BASE=$(python3.12 -c 'import sys; print(sys.base_prefix)')   # e.g. ~/.local/sha
 #     the opaque "Failed to inspect Python interpreter". A framework or
 #     system python that links libpython by absolute path does not need it;
 #     the uv-managed CPython this recipe uses does.
+#     The copy is allowed to fail (a python that needs no dylib has none to
+#     copy); the -V beside it is the real check and catches a genuine miss.
 cp "$BASE/lib/libpython3.12.dylib" "$VENV/lib/" 2>/dev/null || true   # macOS only
 "$VENV/bin/python3.12" -V   # must print the version, not abort
 uv pip install --python "$VENV/bin/python3.12" \
@@ -99,36 +102,49 @@ uv pip install --python "$VENV/bin/python3.12" \
 uv export --frozen --no-emit-project --extra osworld --no-hashes \
     -o "$VENV/build/requirements.locked.txt"
 sed -i '' 's/^local-operator==.*/local-operator==<harness version>/' \
-    "$VENV/build/requirements.locked.txt"
+    "$VENV/build/requirements.locked.txt"   # macOS sed; GNU sed wants -i without ''
 uv pip install --python "$VENV/bin/python3.12" -r "$VENV/build/requirements.locked.txt"
 
 # The one expected `uv pip check` incompatibility is the documented
 # requests>=2.32 override onto OSWorld's ~=2.31 pin (see [tool.uv] above).
 
-# 4. materialise the workspace from the verified inputs root (no download;
+# 4. compute package_digest FIRST: it digests the installed wheel's RECORD and
+#    does not depend on the workspace, while step 5's --package-digest does
+#    depend on it. (Computing it after the build is what forces a second
+#    build with the real value.)
+PKG=$("$VENV/bin/python3.12" -c '
+from pathlib import Path
+from importlib.metadata import PathDistribution
+from local_operator.evaluation.adapters.discovery import distribution_digest
+import sysconfig
+sp = Path(sysconfig.get_paths()["purelib"])
+di = next(sp.glob("lop_osworld_v2_adapter-*.dist-info"))
+print(distribution_digest(PathDistribution(di)))
+')
+echo "package_digest   $PKG"
+
+# 5. materialise the workspace from the verified inputs root (no download;
 #    writes adapter-release.json, benchmark_release.json, task_hashes.json,
 #    adapter-provider.json, inputs.json, tasks/, all read-only).
 #    --out is the WORKSPACE directory itself, not its parent: it is what the
 #    selector's `workspace` field and workspace_digest address. Keep the
 #    selector and build-output.json in the parent so they are not inside the
 #    digest they describe.
-#    Run it with the venv's interpreter (it imports the harness), and note
-#    --version defaults to the adapter's own pyproject version.
+#    Run it with the venv's interpreter (it imports the harness). --version
+#    defaults to the adapter's own pyproject version and REFUSES a value that
+#    disagrees with it unless --allow-version-mismatch is passed.
 "$VENV/bin/python3.12" ~/local-operator/scripts/build_osworld_adapter.py \
     --benchmark-release osworld-v2-2026.08.08 \
     --inputs-root ~/worktrees/osworld \
-    --package-digest <package_digest from step 5> \
-    --out ~/worktrees/osworld/workspaces/0.1.1/workspace
+    --package-digest "$PKG" \
+    --out "$WS"
 
-# 5. compute the three digests the AdapterSelector needs
-python - <<'PY'
-from importlib.metadata import PathDistribution
-from local_operator.evaluation.adapters.discovery import (
-    distribution_digest, workspace_digest,
-)
-print("package_digest  ", distribution_digest(PathDistribution(<dist-info path>)))
-print("workspace_digest", workspace_digest("/opt/lop-adapters/osworld-v2/0.1.1/workspace"))
-PY
+# 6. the remaining digest the AdapterSelector needs (release_digest is printed
+#    by step 5 and written into adapter-release.json).
+"$VENV/bin/python3.12" -c "
+from local_operator.evaluation.adapters.discovery import workspace_digest
+print('workspace_digest', workspace_digest('$WS'))
+"
 ```
 
 `release_digest` is our attestation of the build:
--- a/benchmarks/osworld_v2_adapter/README.md
+++ b/benchmarks/osworld_v2_adapter/README.md
@@ -67,27 +67,58 @@ uv build --wheel --out-dir dist/
 #    `--copies` flag (that is uv pip's install link-mode). Use the stdlib
 #    venv, which DOES copy. `--without-pip` because uv pip installs the rest
 #    and the uv-managed interpreter's ensurepip can SIGABRT on macOS.
-python3.12 -m venv --copies --without-pip /opt/lop-adapters/osworld-v2/0.1.1/venv
-uv pip install --python /opt/lop-adapters/osworld-v2/0.1.1/venv/bin/python \
-    --no-deps dist/lop_osworld_v2_adapter-0.1.1-py3-none-any.whl
-uv pip install --python /opt/lop-adapters/osworld-v2/0.1.1/venv/bin/python \
-    -r <(uv export --frozen --no-emit-project)
-uv pip install --python /opt/lop-adapters/osworld-v2/0.1.1/venv/bin/python \
-    local-operator==<harness version>
+#
+#    Invoke the venv module through the interpreter's REAL path, never through
+#    a ~/.local/bin shim: venv records the invoking argv[0]'s directory as
+#    `home` in pyvenv.cfg, and step 3a resolves libpython relative to it. A
+#    shim directory holds no lib/, so the copy silently finds nothing.
+VENV=/opt/lop-adapters/osworld-v2/0.1.1/venv
+BASE=$(python3.12 -c 'import sys; print(sys.base_prefix)')   # e.g. ~/.local/share/uv/python/cpython-3.12.13-...
+"$BASE/bin/python3.12" -m venv --copies --without-pip "$VENV"
 
-# 3b. for the PAID path only: OSWorld itself is an extra (~380 packages) that
-#     the cloud-free wheel does not need. Install it into the same venv.
-uv pip install --python /opt/lop-adapters/osworld-v2/0.1.1/venv/bin/python \
-    "lop-osworld-v2-adapter[osworld] @ dist/lop_osworld_v2_adapter-0.1.1-py3-none-any.whl"
+# 3a. copy libpython beside the interpreter. REQUIRED on macOS and easy to
+#     miss: `--copies` copies the python binary but NOT the shared library it
+#     links, and the copied binary keeps `@rpath/libpython3.12.dylib` with an
+#     rpath of `<venv>/lib`. Without this the interpreter aborts on every
+#     invocation with "Library not loaded" (dyld) / SIGABRT, and uv reports
+#     the opaque "Failed to inspect Python interpreter". A framework or
+#     system python that links libpython by absolute path does not need it;
+#     the uv-managed CPython this recipe uses does.
+cp "$BASE/lib/libpython3.12.dylib" "$VENV/lib/" 2>/dev/null || true   # macOS only
+"$VENV/bin/python3.12" -V   # must print the version, not abort
+uv pip install --python "$VENV/bin/python3.12" \
+    --no-deps dist/lop_osworld_v2_adapter-0.1.1-py3-none-any.whl
+
+# 3b. install the locked set. For the PAID path export WITH the osworld extra
+#     (~380 packages the cloud-free wheel does not need); omit --extra for a
+#     cloud-free venv. Export to a FILE and keep it beside the venv: it is the
+#     record of what was actually installed, and the lock's own
+#     local-operator pin lags the harness (schema 1.2 and host_secrets only
+#     exist from 0.44.30+), so substitute the harness version you are running
+#     and note the substitution in the build record.
+uv export --frozen --no-emit-project --extra osworld --no-hashes \
+    -o "$VENV/build/requirements.locked.txt"
+sed -i '' 's/^local-operator==.*/local-operator==<harness version>/' \
+    "$VENV/build/requirements.locked.txt"
+uv pip install --python "$VENV/bin/python3.12" -r "$VENV/build/requirements.locked.txt"
+
+# The one expected `uv pip check` incompatibility is the documented
+# requests>=2.32 override onto OSWorld's ~=2.31 pin (see [tool.uv] above).
 
 # 4. materialise the workspace from the verified inputs root (no download;
 #    writes adapter-release.json, benchmark_release.json, task_hashes.json,
 #    adapter-provider.json, inputs.json, tasks/, all read-only).
-python ~/local-operator/scripts/build_osworld_adapter.py \
+#    --out is the WORKSPACE directory itself, not its parent: it is what the
+#    selector's `workspace` field and workspace_digest address. Keep the
+#    selector and build-output.json in the parent so they are not inside the
+#    digest they describe.
+#    Run it with the venv's interpreter (it imports the harness), and note
+#    --version defaults to the adapter's own pyproject version.
+"$VENV/bin/python3.12" ~/local-operator/scripts/build_osworld_adapter.py \
     --benchmark-release osworld-v2-2026.08.08 \
     --inputs-root ~/worktrees/osworld \
     --package-digest <package_digest from step 5> \
-    --out ~/worktrees/osworld/workspaces/0.1.1
+    --out ~/worktrees/osworld/workspaces/0.1.1/workspace
 
 # 5. compute the three digests the AdapterSelector needs
 python - <<'PY'
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "local-operator"
-version = "0.44.49"
+version = "0.44.50"
 description = "A team of proactive AI assistants that can work together behind the scenes to help you get mundane tasks done so you can focus on the fun stuff."
 readme = "README.md"
 authors = [{ name = "Damian Tran", email = "damian@gominerva.com" }]
--- a/scripts/build_osworld_adapter.py
+++ b/scripts/build_osworld_adapter.py
@@ -67,6 +67,7 @@ import os
 import stat
 import subprocess
 import sys
+import tomllib
 from pathlib import Path
 from typing import Any
 
@@ -81,6 +82,29 @@ _DEFAULT_PIN = (
     / "config"
     / "release-v2026.08.08.json"
 )
+
+
+_ADAPTER_PYPROJECT = (
+    Path(__file__).resolve().parents[1] / "benchmarks" / "osworld_v2_adapter" / "pyproject.toml"
+)
+
+
+def _adapter_version() -> str:
+    """The adapter distribution version declared in its pyproject.
+
+    Read from source rather than from ``importlib.metadata`` on purpose: this
+    script runs from the repository against a workspace that is built BEFORE
+    (or without) the wheel being installed into the caller's interpreter, so
+    the installed distribution is the wrong authority and may not exist at all.
+    A read failure is not fatal -- ``--version`` is still explicit on the
+    documented command line -- but it must not silently produce a wrong
+    attestation, so the fallback is an obviously-unreal marker.
+    """
+
+    try:
+        return str(tomllib.loads(_ADAPTER_PYPROJECT.read_text())["project"]["version"])
+    except (OSError, KeyError, tomllib.TOMLDecodeError):
+        return "unknown"
 
 
 class VerificationFailed(Exception):
@@ -277,7 +301,14 @@ def main(argv: list[str] | None = None) -> int:
         default=Path(os.environ.get("OSWORLD_INPUTS_ROOT", _DEFAULT_INPUTS_ROOT)),
     )
     parser.add_argument("--release-pin", type=Path, default=_DEFAULT_PIN)
-    parser.add_argument("--version", default="0.1.0")
+    # Defaults to the adapter's OWN declared version rather than a literal.
+    # The version is an input to ``_release_digest``, so a stale literal here
+    # silently mints a workspace whose release_digest attests a distribution
+    # version that was never built -- a wrong attestation that no digest check
+    # can catch, because every digest is internally consistent with it. Read
+    # from pyproject.toml so the default cannot drift from the wheel again
+    # (it already had: the literal still said 0.1.0 after the 0.1.1 bump).
+    parser.add_argument("--version", default=_adapter_version())
     parser.add_argument("--package-digest", default="0" * 64)
     args = parser.parse_args(argv)
 
--- a/scripts/build_osworld_adapter.py
+++ b/scripts/build_osworld_adapter.py
@@ -89,22 +89,27 @@ _ADAPTER_PYPROJECT = (
 )
 
 
-def _adapter_version() -> str:
-    """The adapter distribution version declared in its pyproject.
+def _adapter_version() -> str | None:
+    """The adapter distribution version declared in its pyproject, or None.
 
     Read from source rather than from ``importlib.metadata`` on purpose: this
     script runs from the repository against a workspace that is built BEFORE
     (or without) the wheel being installed into the caller's interpreter, so
     the installed distribution is the wrong authority and may not exist at all.
-    A read failure is not fatal -- ``--version`` is still explicit on the
-    documented command line -- but it must not silently produce a wrong
-    attestation, so the fallback is an obviously-unreal marker.
+
+    Returns None rather than a placeholder when the declaration cannot be read.
+    A placeholder would flow into ``_release_digest`` and mint a real,
+    well-formed workspace attesting a version that does not exist -- and since
+    ``adapter-release.json`` carries ONLY the digest (``discovery`` rejects any
+    other key), the artifact could not record its own doubt even if we wanted
+    it to. An unattributable attestation is worse than no artifact, so the
+    caller refuses to build instead.
     """
 
     try:
         return str(tomllib.loads(_ADAPTER_PYPROJECT.read_text())["project"]["version"])
     except (OSError, KeyError, tomllib.TOMLDecodeError):
-        return "unknown"
+        return None
 
 
 class VerificationFailed(Exception):
@@ -301,16 +306,61 @@ def main(argv: list[str] | None = None) -> int:
         default=Path(os.environ.get("OSWORLD_INPUTS_ROOT", _DEFAULT_INPUTS_ROOT)),
     )
     parser.add_argument("--release-pin", type=Path, default=_DEFAULT_PIN)
-    # Defaults to the adapter's OWN declared version rather than a literal.
-    # The version is an input to ``_release_digest``, so a stale literal here
-    # silently mints a workspace whose release_digest attests a distribution
-    # version that was never built -- a wrong attestation that no digest check
-    # can catch, because every digest is internally consistent with it. Read
-    # from pyproject.toml so the default cannot drift from the wheel again
-    # (it already had: the literal still said 0.1.0 after the 0.1.1 bump).
-    parser.add_argument("--version", default=_adapter_version())
+    # Left with NO argparse default: the adapter's own declared version is
+    # resolved below so that "not passed" stays distinguishable from "passed
+    # a value equal to the default". The version is an input to
+    # ``_release_digest``, so a value that does not match the distribution
+    # actually built mints a workspace attesting a version that never existed
+    # -- a wrong attestation no digest check can catch, because every digest
+    # is internally consistent with it. That already happened once: the
+    # literal default still said 0.1.0 after the 0.1.1 bump.
+    parser.add_argument("--version", default=None)
+    parser.add_argument(
+        "--allow-version-mismatch",
+        action="store_true",
+        help="permit --version to differ from the adapter's declared version "
+        "(for attesting a build of a version other than this tree's)",
+    )
     parser.add_argument("--package-digest", default="0" * 64)
     args = parser.parse_args(argv)
+
+    # Resolve the attested version under one rule: it must agree with what the
+    # tree declares, unless the operator says out loud that it should not.
+    # Building a version OTHER than the tree's is legitimate (re-attesting an
+    # older wheel against the same corpus), but it is a deliberate act, not
+    # something to accept silently -- silently accepting a mismatched value is
+    # the same shape as the stale-default defect this script already had.
+    declared = _adapter_version()
+    version = args.version if args.version is not None else declared
+    if version is None:
+        print(
+            "build_osworld_adapter: cannot determine the adapter version "
+            f"({_ADAPTER_PYPROJECT} is unreadable or declares none) and no "
+            "--version was given. Refusing to build: the version is an input "
+            "to release_digest, and adapter-release.json carries only that "
+            "digest, so the workspace could not record that its attestation "
+            "is unattributable.",
+            file=sys.stderr,
+        )
+        return 2
+    if args.version is not None and declared is not None and args.version != declared:
+        if not args.allow_version_mismatch:
+            print(
+                f"build_osworld_adapter: --version {args.version!r} disagrees with the "
+                f"version the adapter declares ({declared!r} in {_ADAPTER_PYPROJECT}). "
+                "release_digest attests the distribution that was built, so a "
+                "mismatch produces a workspace claiming a version nobody can "
+                "verify. Pass --allow-version-mismatch if that is deliberate.",
+                file=sys.stderr,
+            )
+            return 2
+        # Deliberate and stated: still say so, because the resulting digest is
+        # not reproducible from this tree alone.
+        print(
+            f"build_osworld_adapter: attesting {args.version!r} while this tree "
+            f"declares {declared!r} (--allow-version-mismatch).",
+            file=sys.stderr,
+        )
 
     pin = json.loads(Path(args.release_pin).read_bytes())
     if pin.get("release") != args.benchmark_release:
@@ -331,7 +381,7 @@ def main(argv: list[str] | None = None) -> int:
         out=args.out,
         facts=facts,
         release=args.benchmark_release,
-        version=args.version,
+        version=version,
         package_digest=args.package_digest,
     )
     print(
--- a/scripts/build_osworld_adapter.py
+++ b/scripts/build_osworld_adapter.py
@@ -112,6 +112,63 @@ def _adapter_version() -> str | None:
         return None
 
 
+class VersionRefused(Exception):
+    """The attested version cannot be trusted, so no workspace may be built.
+
+    Separate from ``VerificationFailed`` because it carries a different exit
+    code (2, a usage error) and fires BEFORE any input is read: the version is
+    a property of the build, not of the corpus.
+    """
+
+
+def resolve_attested_version(
+    *, requested: str | None, declared: str | None, allow_mismatch: bool
+) -> tuple[str, str | None]:
+    """The single implementation of "which version does this build attest".
+
+    Pure -- no argparse, no filesystem, no corpus -- so the tests exercise the
+    REAL resolution rather than re-implementing it. That distinction is not
+    academic: the digest regression test previously recomputed the rule itself
+    and therefore stayed green when the rule was mutated, which is the same
+    bypass the original stale-default defect exploited.
+
+    Returns the version to attest plus an optional advisory line the caller
+    should print. Raises ``VersionRefused`` when nothing trustworthy can be
+    attested.
+
+    The rule: the attested version must agree with what the tree declares
+    unless the operator says out loud that it should not. Attesting a version
+    OTHER than the tree's is legitimate (re-attesting an older wheel against
+    the same corpus) but deliberate, so it is opt-in and still announced --
+    the resulting digest is not reproducible from this tree alone.
+    """
+
+    version = requested if requested is not None else declared
+    if version is None:
+        raise VersionRefused(
+            "cannot determine the adapter version "
+            f"({_ADAPTER_PYPROJECT} is unreadable or declares none) and no "
+            "--version was given. Refusing to build: the version is an input "
+            "to release_digest, and adapter-release.json carries only that "
+            "digest, so the workspace could not record that its attestation "
+            "is unattributable."
+        )
+    if requested is not None and declared is not None and requested != declared:
+        if not allow_mismatch:
+            raise VersionRefused(
+                f"--version {requested!r} disagrees with the version the adapter "
+                f"declares ({declared!r} in {_ADAPTER_PYPROJECT}). release_digest "
+                "attests the distribution that was built, so a mismatch produces "
+                "a workspace claiming a version nobody can verify. Pass "
+                "--allow-version-mismatch if that is deliberate."
+            )
+        return version, (
+            f"attesting {requested!r} while this tree declares {declared!r} "
+            "(--allow-version-mismatch)."
+        )
+    return version, None
+
+
 class VerificationFailed(Exception):
     """An input does not match the pin. The message names the path."""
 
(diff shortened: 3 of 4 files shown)


damianvtran:
Follow-up commit `0316f41d`: the paid-episode example now names `openrouter/google/gemini-3.8-flash` (the operator retired the DeepSeek route).

Worth recording that the route is **not** apparatus state, so this is a docs-only change: `run_episode.py --route` is `required=True` with no default, and grep across the committed tree confirms no selector, workspace, `adapter-provider.json`, or `inputs.json` carries a model id. Changing route needs no rebuild and no new digests.

The fold example moved with it — `google/gemini-3.8-flash` seals as `google_sgemini-3.8-flash`, verified through `route_ids.fold_model_id` and round-tripped with `unfold_model_id` rather than derived by hand.

damianvtran:
### Agent review — round 1

**Reviewer:** `reviewer` subagent (anthropic/claude-opus-5) — independent of the author
**Scope:** `origin/main`..`0316f41d` (`63605edf..0316f41d`), 3 files: `scripts/build_osworld_adapter.py`, `benchmarks/osworld_v2_adapter/README.md`, `pyproject.toml`

Reviewed in a clean detached worktree at `0316f41d`. Both claimed defects were **reproduced independently** before accepting the fix, and the author's central claim — that the new default reproduces the hand-specified digests byte for byte — is **confirmed**.

---

## Verification by reproduction

**Defect 2 (attestation integrity) — confirmed, and the fix is exact.**

The pre-fix script (`git show origin/main:scripts/build_osworld_adapter.py`) run with the *documented* command line — no `--version` — against the real inputs root mints:

```
release_digest   a15961b19ea7a7eddccbc38e5d24085e1354cfb73769c6740e50e418f0022981
workspace_digest e6f49e0b3e584d117896172c403a5582e22d72d3cc0858124ba6d21e59d33906
```

while the adapter's actual built distribution is `0.1.1` (`adapter.py:90 _VERSION = "0.1.1"`, `benchmarks/osworld_v2_adapter/pyproject.toml:16`). So the old default did attest a version that was never built, and — as claimed — no digest check catches it, because `adapter-release.json` contains only the digest and never the version that went into it. Nothing else in the tree recomputes the `_release_digest` payload (`grep 'lop-osworld-v2-adapter|'` finds only `build_osworld_adapter.py:136` and a test helper that hardcodes `0.1.1`), so the wrong attestation is internally consistent everywhere.

The fix reproduces the pinned pilot digests **exactly**. Built at `0316f41d` with no `--version`, with `--version 0.1.1`, and compared against the staged pilot workspace:

| build | release_digest | workspace_digest |
|---|---|---|
| new default (no `--version`) | `d0067e23…41cd` | `c512802639615cc6…a5f1` |
| explicit `--version 0.1.1` | `d0067e23…41cd` | `c512802639615cc6…a5f1` |
| staged pilot `workspaces/0.1.1` | `d0067e23…41cd` (selector) | `c512802639615cc6…a5f1` |

Byte-identical across all three. The default now tracks the adapter's own pyproject, and `_adapter_version()` correctly reads from source rather than `importlib.metadata` — the docstring's reasoning (the workspace is built before/without the wheel being installed) is right and matches how the README drives it.

**Defect 1 (dead interpreter) — confirmed on both halves.**

Old recipe, verbatim: `python3.12 -m venv --copies --without-pip <venv>` then invoking it →

```
dyld[19697]: Library not loaded: @rpath/libpython3.12.dylib
  Reason: tried: '<venv>/lib/libpython3.12.dylib' (no such file)
Abort trap: 6   (rc=134)
```

And `uv pip install --python <that venv>` against it exits **2** with the same dyld abort — the opaque failure the README now names. The shim claim also reproduces: invoking via `~/.local/bin/python3.12` (a symlink) records `home = /Users/damian/.local/bin` in `pyvenv.cfg`, and that directory has no `lib/`, so `$BASE/lib/...` would resolve to nothing. Using `sys.base_prefix` as the new recipe does records the real `home` and the copy finds the dylib.

Corrected recipe built from scratch in a temp dir: `Python 3.12.13`, rc=0. The interpreter is a **real file, not a symlink** (`islink: False, isfile: True`, mode `0o100755`) — satisfying `discovery._symlink_free`. The dylib step is now written down rather than folklore.

**Scope / artifacts — clean.** `git diff --name-status origin/main...HEAD` is exactly the three expected files. No wheel, venv, workspace, dylib, selector, or locked-requirements file is tracked; worktree is clean.

**Same-class sweep.** Remaining defaults were sampled deliberately. `--package-digest` defaulting to `"0"*64` is intentional (step 5 computes the real one and the build is re-run) and `_DEFAULT_PIN`/`_DEFAULT_INPUTS_ROOT` are derived/overridable. `run_episode.py` already derives `harness_version` from the checkout's pyproject with a documented rationale (`run_episode.py:203-212`), and `environment_release`/`environment_digest` come from the selector rather than a literal. The one residual is F2 below.

**Gates (this head, full tree, exactly as CI):** flake8 rc=0 · black 26.1.0 `814 files unchanged` rc=0 · isort 5.13.2 rc=0 · pyright `0 errors, 0 warnings` rc=0 · unit suite **11208 passed, 15 skipped** (26:06) rc=0. Targeted `tests/unit/evaluation`: 707 passed. CI on `0316f41d`: all 8 checks SUCCESS.

---

## Findings

**F1 — MINOR — `scripts/build_osworld_adapter.py:311` — a `--version` that disagrees with pyproject is silently accepted.**
Probed directly: `--version 9.9.9` builds a workspace attesting `9.9.9` and exits 0 (`release_digest eef60551…755b`). This is the *same shape* as the defect being fixed — the default is now correct, but the explicit override still has no guard. I do **not** rate it a blocker: the pre-fix bug bit the documented path (no flag, wrong answer), whereas this needs someone to pass a wrong value deliberately, and a legitimate use exists (attesting a build of a version other than the tree's). Worth a one-line warning to stderr when `args.version != _adapter_version()`, or an `--allow-version-mismatch` gate. Not a reason to hold this PR.

**F2 — MINOR — `scripts/build_osworld_adapter.py:104-107` — the `"unknown"` fallback mints a real, well-formed attestation.**
Reproduced by copying the script outside the repo so `parents[1]` no longer finds the adapter pyproject: it builds successfully, rc=0, silently attesting version `"unknown"` — confirmed by recomputing the payload (`lop-osworld-v2-adapter|unknown|…` → `780b3836…04d0`, matching the written `adapter-release.json`). The docstring says the fallback is "an obviously-unreal marker", which is true for a human reading the *command*, but the resulting workspace is indistinguishable from a good one and the version is nowhere in the artifact. The marker is only obvious if someone looks. Given `--version` is on the documented command line this is low-likelihood, but the honest fix is one line: print a warning to stderr when the fallback fires. (An exit-nonzero-unless-`--version`-is-explicit would be stricter and also defensible.)

**F3 — MINOR — `benchmarks/osworld_v2_adapter/README.md:130` — step 5 still points at a path the corrected recipe never creates.**
Step 3 now builds into `/opt/lop-adapters/osworld-v2/0.1.1/venv` and step 4's `--out` was corrected to `~/worktrees/osworld/workspaces/0.1.1/workspace`, but step 5 still passes `workspace_digest("/opt/lop-adapters/osworld-v2/0.1.1/workspace")` — a directory no step in this recipe writes. Verified it raises rather than misreporting (`AdapterDiscoveryError: adapter launch path is unavailable`; `/opt/lop-adapters` does not exist on this host), so it fails loudly and cannot mint a wrong digest — which is why this is minor and not major. Still, the PR's own stated purpose is that the documented recipe must work as written, and this line does not. Fix: use the same `--out` path as step 4, or introduce a `WS=` variable alongside `VENV=` and use it in both.

**F4 — NIT — `benchmarks/osworld_v2_adapter/README.md:96` — `sed -i ''` is macOS-only.**
Consistent with the macOS-only `cp` above it, and the recipe is explicitly a macOS one, so this is coherent rather than wrong. Worth a trailing `# macOS; GNU sed wants -i` for the reader who runs the paid path on Linux.

**F5 — NIT — `benchmarks/osworld_v2_adapter/README.md:92` — `cp … 2>/dev/null || true` hides a failed copy.**
The very next line (`"$VENV/bin/python3.12" -V   # must print the version, not abort`) is the real check and catches it immediately, so the suppression is safe as written and deliberate (it makes the step a no-op on framework pythons). Flagging only because a swallowed `cp` was adjacent to the folklore that caused this defect; the verification line beside it is what makes it acceptable.

Two observations that are **not** findings: `uv export -o "$VENV/build/…"` creates the missing parent on its own (verified, rc=0, file written), so the recipe needs no `mkdir`; and the lock's `local-operator==0.44.26` pin is real (confirmed at line 361 of a fresh export), so the documented `sed` substitution is genuinely required, not defensive.

---

## Verdict

**Blockers: 0. Majors: 0.** (3 minor, 2 nit.)

**`clean` — this round is TERMINAL.** The attestation defect is real, was reproduced against the pre-fix script, and the fix reproduces the hand-specified `release_digest` and `workspace_digest` byte for byte against the staged pilot workspace. The venv recipe now produces a working, symlink-free interpreter from scratch where the old one aborted. Scope is exactly the three intended files with no generated artifacts. All five gates are green on this head and CI is 8/8.

F1–F3 are follow-ups, not another round. F3 is the one I would fold in if any further commit lands on this branch anyway — it is a one-token path edit in a recipe whose correctness is this PR's subject — but it fails loudly rather than silently and does not gate merge.

Version check re-run at the end of review: head `0316f41d` declares `0.44.45`; `origin/main` is at `63605edf` declaring `0.44.44`; PyPI latest is `0.44.44` and `0.44.45` is absent. Strictly above both — **no collision**. Recommend merge.

damianvtran:

**Implementer:** `coder` subagent (anthropic/claude-opus-5)
**Head:** `6a64ed7d` (was `0316f41d`) · **Scope of this remediation:** `0316f41d..6a64ed7d`, 3 files (`scripts/build_osworld_adapter.py`, `benchmarks/osworld_v2_adapter/README.md`, `pyproject.toml`)

All five findings addressed in one commit — `6a64ed7d`. Thanks for reproducing both defects independently rather than taking the description; the F1/F2 probes are what made it clear these are the same class as the bug being fixed, not separate nits.

---

**F1 — MINOR — `--version` disagreeing with pyproject silently accepted — FIXED (`6a64ed7d`)**

Refused with exit 2 naming both values. You noted a legitimate use exists (attesting a build of a version other than the tree's), so it stays reachable behind an explicit `--allow-version-mismatch`, which also prints what it is doing — the resulting digest is not reproducible from this tree alone, so a silent success would be its own trap. I went with the gate rather than the stderr warning you offered as the lighter option: a warning on a path that still exits 0 and still writes the artifact does not stop the wrong attestation from being used, which is the failure mode this PR exists to close.

`--version` now carries no argparse default, so "not passed" stays distinguishable from "passed a value that happens to equal the default".

```
$ build_osworld_adapter.py --version 9.9.9 … --out …/bad
build_osworld_adapter: --version '9.9.9' disagrees with the version the adapter
declares ('0.1.1' in …/pyproject.toml). release_digest attests the distribution
that was built, so a mismatch produces a workspace claiming a version nobody can
verify. Pass --allow-version-mismatch if that is deliberate.
rc=2      and no workspace written (…/bad: No such file or directory)

$ … --version 9.9.9 --allow-version-mismatch … --out …/ok
build_osworld_adapter: attesting '9.9.9' while this tree declares '0.1.1'
release_digest  eef6055166f4a0883a6c3eabdc482cd942e24908066efcd01c63ea20b0cd755b
rc=0
```

That digest is exactly the `eef60551…755b` your probe recorded, so the escape hatch reproduces the old behaviour when asked for it explicitly.

**F2 — MINOR — the `"unknown"` fallback mints a real, well-formed attestation — FIXED (`6a64ed7d`), by refusing rather than warning**

You offered two options; I took the stricter one you flagged as also defensible, because the milder one is not actually available here. Marking the artifact is impossible by construction: `discovery.verify_release_manifest` requires `adapter-release.json` to be **exactly** `{"release_digest": …}` and rejects any other key (`discovery.py:252`), so the workspace has nowhere to record its own doubt. That leaves warn-and-build or refuse — and since an unattributable attestation is worse than no artifact, the build now refuses. `_adapter_version()` returns `None` instead of a placeholder, and the docstring no longer claims a marker the artifact does not carry.

Reproduced with your exact setup (script copied outside the repo so `parents[1]` misses the adapter pyproject):

```
$ build_orphan.py --benchmark-release … --inputs-root … --out …/orphan-out
build_osworld_adapter: cannot determine the adapter version (…/pyproject.toml is
unreadable or declares none) and no --version was given. Refusing to build: the
version is an input to release_digest, and adapter-release.json carries only that
digest, so the workspace could not record that its attestation is unattributable.
rc=2      and no workspace written

$ build_orphan.py … --version 0.1.1 --out …/orphan-ok
release_digest  d0067e23af3dc2ed790c2a8b802ee453200d516466ad27770d7f2bbe7b0b41cd
rc=0
```

An explicit `--version` still builds normally from an orphaned script, and reproduces the pinned digest.

**F3 — MINOR — README step 5 pointed at a path the recipe never creates — FIXED (`6a64ed7d`)**

Took your `WS=` suggestion: the workspace path is now a variable declared beside `VENV=` and used by both the build and the digest step, so they cannot drift apart again. While in there I also reordered the two steps — `package_digest` is now computed **before** the build that consumes it, since the old ordering ("--package-digest `<package_digest from step 5>`" where step 5 came after step 4) forced a throwaway first build with a placeholder digest. That is the same "documented recipe does not work as written" defect one step over, so it belongs in this fix rather than a later one.

**F4 — NIT — `sed -i ''` is macOS-only — TAKEN (`6a64ed7d`).** Trailing `# macOS sed; GNU sed wants -i without ''`.

**F5 — NIT — `cp … || true` hides a failed copy — TAKEN (`6a64ed7d`).** Kept the suppression (it is what makes the step a no-op on framework pythons, as you read it) and added the comment saying the `-V` line beside it is the real check.

---

### The recipe was walked end to end, not just edited

Your closing point — that an unwalked recipe is folklore — is the reason this PR exists, so I walked the corrected version verbatim in a scratch dir rather than only diffing it. Every step rc=0:

| Step | Result |
|---|---|
| 2 wheel | `lop_osworld_v2_adapter-0.1.1-py3-none-any.whl` |
| 3 interpreter | `Python 3.12.13`, real file not a symlink |
| 3a dylib | copied; `-V` prints rather than aborting |
| 3b locked set | 400 packages; `uv pip check` → exactly the documented `requests` override |
| 4 package_digest | `4e9bd3d0b6c36170…5f0b` (scratch venv) |
| 5 workspace | 108 tasks, `prepared` at `d578d2d4` |
| 6 workspace_digest | computed via `WS` |

Step 1 (`uv lock --upgrade-package local-operator`) was deliberately **not** run: it mutates the committed `uv.lock`, which is out of this PR's scope. Everything downstream of it was.

One honest note on step 6: it initially raised `adapter launch path has a symlink or lexical alias` — because my scratch dir was under `/tmp`, which is a symlink to `/private/tmp` on macOS. That is my scratch path, not the recipe: the documented `WS` under `~/worktrees/osworld` is symlink-free (`realpath == path`), and `/tmp` is refused by the volatile-root guard anyway. Re-running against the resolved path succeeds.

**Regression check — the happy path is unchanged.** No flag and `--version 0.1.1` both still produce `release_digest d0067e23…41cd` and `workspace_digest c512802639615cc6…a5f1`, byte-identical to the staged pilot workspace. The staged `0.1.1` and `0.1.0` pilot workspaces both still match their pins; nothing under `~/worktrees/osworld` was modified outside my own scratch.

### Gates (whole tree, exactly as CI, on `6a64ed7d`)

flake8 rc=0 · black 26.1.0 `814 files unchanged` · isort rc=0 · pyright **0 errors, 0 warnings** · unit suite **11208 passed, 15 skipped** (10:50).

### Version

Re-checked at bump time, per the collision rule: `origin/main` is `ecabd0de` declaring **0.44.45** and PyPI latest is **0.44.45** (your round-1 check saw 0.44.44 — #561 landed in between and took 0.44.45, which this branch had claimed). `0.44.46` is claimed by `dev-stop`. This head declares **0.44.47** — strictly above main, PyPI, and every other open PR. Rebased onto `ecabd0de`.

Not merging or releasing — the manager owns the gate.

damianvtran:
### Agent review — round 2

**Reviewer:** `reviewer` subagent (anthropic/claude-opus-5) — independent of the author
**Scope:** `0316f41d..6a64ed7d` — reviewed as the true remediation diff `fbcdeec1..6a64ed7d` (see Scope note), 3 files: `scripts/build_osworld_adapter.py`, `benchmarks/osworld_v2_adapter/README.md`, `pyproject.toml`

Round-2 freshness review in a clean detached worktree at `6a64ed7d`. Round 1's conclusions are not re-litigated. **All three round-1 fixes verified by reproduction, and the staged pilot digests still reproduce byte for byte.** Two findings, both about shipping this commit rather than about the fixes themselves.

---

#### Scope note (not a finding)

`git diff 0316f41d..6a64ed7d` reports 8 files, but the branch was **rebased** between rounds (`0316f41d` is not an ancestor of head; merge-base is `f4d11ddd`). The 5 extra files — `docs/BENCHMARKS.md`, `docs/benchmarks/osworld_2/README.md`, and 3 `agent_seeds/*.md` — are byte-identical to the new base `ecabd0de`, i.e. rebase noise, not branch content. The real remediation diff is `fbcdeec1..6a64ed7d` (3 files, +98/−32), and `fbcdeec1` is patch-identical to the round-1 head `0316f41d`. **Scope is clean.**

---

#### Verification of the round-1 fixes

**F1 — `--version` mismatch now refuses — CONFIRMED.** Run against the real inputs root:

```
--version 0.9.9                              → rc=2, artifact on disk: NO
  stderr names BOTH values: "--version '0.9.9' disagrees with the version
  the adapter declares ('0.1.1' in .../pyproject.toml)"
--version 0.9.9 --allow-version-mismatch     → rc=0, warns "attesting '0.9.9'
  while this tree declares '0.1.1'", digest b5c94ad2… (correctly different)
```

The refusal is before any write — `build_workspace` is never reached, so nothing is left half-built. `--version` carries `default=None` (`:314`) and is resolved after parsing (`:341`), so "not passed" stays distinguishable from "passed the default"; the mismatch branch is additionally guarded on `args.version is not None` (`:346`), so the default path can never trip it.

**The flag cannot be set accidentally — checked, not assumed.** `--allow-version-mismatch` is `action="store_true"` reachable only from argv: no `set_defaults`, no `parse_known_args`, no `fromfile_prefix_chars`, no `allow_abbrev` change, and the script's only `os.environ` read is `OSWORLD_INPUTS_ROOT` (`:306`). It is not implied by any other flag.

**F2 — undeterminable version now refuses — CONFIRMED by reproducing the orphan setup.** Copied the script alone to a directory with no `benchmarks/` tree; `_ADAPTER_PYPROJECT` resolves to a nonexistent path and `_adapter_version()` returns `None`:

```
orphan, no --version    → rc=2, no artifact, stderr explains the refusal
orphan, --version 0.1.1 → rc=0, release_digest d0067e23…41cd (pinned value)
```

The docstring no longer claims a marker the artifact lacks — `grep '"unknown"'` finds nothing in either changed file.

**The author's justification is correct.** `discovery.verify_release_manifest` (`discovery.py:252`) asserts `value != {"release_digest": selector.release_digest}` against a byte-exact canonical re-serialisation, so a doubt marker genuinely cannot be added to `adapter-release.json` without the worker refusing to launch. Refusing to build is the right call *and* the stated reason is accurate — the next maintainer can trust it.

**F3 — README recipe walked end to end — CONFIRMED, every step rc=0.** Walked verbatim in my own scratch (outside `/tmp`), not merely diffed:

| Step | Result |
|---|---|
| 2 wheel | `lop_osworld_v2_adapter-0.1.1-py3-none-any.whl` |
| 3 interpreter | `Python 3.12.13`, `python3.12` a real 49,968-byte file (not a symlink) |
| 3a dylib | copied; `-V` prints rather than aborting |
| 3b locked set | 423 lines exported, 400 packages installed |
| `uv pip check` | exactly the one documented `requests~=2.31.0` vs `2.34.2` incompatibility |
| 4 package_digest | `1c4a5b36…3f55`, computed from the installed RECORD alone |
| 5 workspace | 108 tasks, rc=0, consuming step 4's value via `"$PKG"` |
| 6 workspace_digest | `340bb540…c544` via `WS` |

**The reorder is correct.** Step 4 reads only `sysconfig.get_paths()["purelib"]` and the `.dist-info` RECORD — it has no dependency on the workspace, while step 5 consumes it via `--package-digest "$PKG"`. No throwaway first pass remains; the old ordering (`<package_digest from step 5>` referenced by step 4) did require one. `WS` is declared once beside `VENV` and used by both step 5 and step 6, so they cannot drift.

Step 1 (`uv lock --upgrade-package`) was deliberately skipped as expected — it mutates a lockfile out of scope. Everything downstream of it ran. Two notes for the walker, neither a defect in this diff: the recipe's `cd benchmarks/osworld_v2_adapter` is load-bearing for step 3b (the `osworld` extra is defined in the *adapter's* pyproject, not the root), and the `local-operator==<harness version>` substitution must name a **published** version — substituting the unreleased `0.44.47` makes step 3b unsatisfiable. Both are pre-existing text from #522, untouched here.

#### THE REGRESSION CHECK — happy path unchanged

Built against the real inputs root at this head, both ways:

| build | release_digest | workspace_digest |
|---|---|---|
| no `--version` | `d0067e23…41cd` | `c512802639615cc6…a5f1` |
| `--version 0.1.1` | `d0067e23…41cd` | `c512802639615cc6…a5f1` |
| staged pilot `workspaces/0.1.1` | `d0067e23…41cd` | `c512802639615cc6…a5f1` |

`diff -r` reports **byte-identical** for both builds against the staged pilot workspace. Separately, feeding the pilot's `package_digest` into a build from my *freshly walked* venv also reproduces both pinned digests, so the chain closes from a clean recipe run and not just from a pre-existing venv. **The staged pilot is not invalidated.** Nothing under `~/worktrees/osworld` was modified — its `adapter-release.json` still reads `d0067e23…41cd`.

---

### Findings

**B1 — BLOCKER — version `0.44.47` is taken; this PR cannot ship as numbered**

At review start `origin/main` was `0.44.46` and `0.44.47` was free. It was consumed mid-review: **#539 merged as `aa94b81e` and published `0.44.47` to PyPI at `2026-09-03T04:35:43`.** Current state:

```
origin/main  version = "0.44.47"   (aa94b81e)
PyPI latest  0.44.47               (HTTP 200 for /0.44.47/json)
PR #557      version = "0.44.47"   (also still claiming it)
PR #565      version = "0.44.47"   ← this PR
```

This head is no longer strictly above main or PyPI — it *equals* both. Publishing would be rejected by PyPI as a duplicate. `0.44.48` is free (PyPI 404) and unclaimed by any open PR as of this comment; it will need re-checking at merge, since this number has now collided five times today.

Related and mechanical: **the PR is `CONFLICTING`/`DIRTY`** against main. `git merge-tree` shows the conflict is confined to a single hunk — the `pyproject.toml` version line, ours `0.44.47` vs theirs `0.44.47`. Rebasing onto current main and taking `0.44.48` resolves the conflict and B1 in one step. No other file conflicts. GitHub reports no checks on `dev-ws-rebuild`, so CI has not run this head.

**B2 — MAJOR — the two new refusal paths have no test coverage**

`tests/unit/evaluation/adapters/osworld/test_build_and_scripts.py` is the file that covers this script (15 tests, all passing here in 1.49s), and it has **no** test for either new behaviour: `grep` for `allow_version_mismatch|disagrees|Refusing|_adapter_version` returns nothing. The existing suite covers the fail-closed input paths thoroughly (`exit_4` tamper cases, `/tmp` refusal, wrong prepared commit), which is exactly the precedent — these two new exits are the same class of guard and are the whole point of the commit. As it stands, a future edit that restores a permissive default, or drops the `args.version is not None` guard so the default path starts tripping the mismatch branch, breaks attestation integrity with a green suite.

Both are cheap and need no inputs root: assert `main([...])` returns `2` for a `--version` that disagrees, `0` with `--allow-version-mismatch`, and — with `_ADAPTER_PYPROJECT` monkeypatched to a missing path — `2` with no `--version` and `0` with one. A `--out` assertion that the directory was not created pins the "writes nothing" half.

No MINOR or NIT findings worth a round. Round 1's five are all correctly applied, including the two nits (`# macOS sed; GNU sed wants -i without ''`, and the comment explaining that `-V` is the real check beside the suppressed `cp`).

---

### Gates (whole tree, exactly as CI, on `6a64ed7d`)

flake8 **rc=0** · black 26.1.0 **814 files unchanged** · isort **rc=0** · pyright **0 errors, 0 warnings, 0 informations**.

Unit suite: **11183 passed, 25 failed, 15 skipped** (17:10). **The failures are not this PR's** — they are load-induced flakes on a contended host (load average 88-166 during the run, several sibling worktrees running suites). Evidence:

- All failures are in `tests/unit/tools/test_eval_tool.py` and `tests/unit/model/test_prices.py`. This diff touches only a README, the `pyproject.toml` version line, and a standalone `scripts/` file that no test in either module imports.
- Re-run serially (`-n0`): **91 of 92 passed** — 7 of the 8 named failures were pure parallel-contention flakes.
- The remaining one, `test_prices.py::test_a_background_revalidation_gets_the_full_default_timeout`, is **flaky in isolation and pre-existing on main**: run 5× against a pristine `origin/main` worktree with no PR changes, it failed 1 of 5. Not introduced here; noted for the record only, **not filed as an issue** and not a finding against this PR.

`test_build_and_scripts.py` — the targeted coverage for the changed script — is **15 passed**.

---

### Verdict

**Blockers: 1 · Majors: 1 · Minors: 0 · Nits: 0**

**NOT clean — this round is not terminal.** The engineering under review is correct and I would happily approve it on its merits: all three round-1 fixes reproduce, the attestation-integrity reasoning is sound and accurately documented, the recipe now works as written, and the staged pilot digests are byte-identical. Nothing about the fixes needs rework.

What blocks is packaging, not code: **B1** means the version this head declares is already on PyPI, so merging and releasing as numbered would fail at publish and the branch will not merge without a conflict resolution anyway.

**Merge recommendation: do not merge at `6a64ed7d`.** Rebase onto current `origin/main` (`aa94b81e`), take **`0.44.48`** — re-checking main, PyPI, and every open PR at bump time, since #557 is still sitting on `0.44.47` — and add the two refusal-path tests for **B2**. Both are contained changes to a 3-file diff; round 3 should be a short freshness pass over the rebase, the version line, and the new tests, not a re-audit. CI has not run this head at all, so it will need to go green after the rebase.

damianvtran:

**Implementer:** `coder` subagent (anthropic/claude-opus-5)
**Head:** `95b6474b` (was `0316f41d`) · **Version:** 0.44.51 · **Scope:** rebased onto `1863741d`, now MERGEABLE

Both findings addressed. B2 is the one worth the words — the reviewer is right that untested guards in this file are how the original defect survived, so the tests are written to fail on the *bug*, not merely on the guard's error string.

---

**B1 — BLOCKER — 0.44.47 consumed during review — FIXED (`95b6474b`)**

Rebased onto current `main` (`1863741d`); the conflict was only the version line, as merge-tree predicted. This collided **three more times while I worked**: 0.44.47 → 0.44.49 (taken by #574 mid-run) → 0.44.50 (claimed by `dev-usage-freshness` and `dev-config-watch` between my bump and my push) → **0.44.51**.

Confirmed at hand-off, moments ago:

| | version |
|---|---|
| this head | **0.44.51** |
| `origin/main` (`1863741d`) | 0.44.49 |
| PyPI latest | 0.44.49 |
| highest other open-PR claim | 0.44.50 |

0.44.51 is claimed by no other open PR. Given the rate, re-check before merging rather than trusting this table.

**B2 — MAJOR — both refusal paths had zero tests — FIXED (`7e0983c2`)**

Four cases in `tests/unit/evaluation/adapters/osworld/test_build_and_scripts.py`, matching the file's existing conventions (`build.main(argv)` in-process, `capsys`, `durable_path`). Each asserts **`not out.exists()`**, not just the exit code — the defect is a workspace that exists and looks pristine while attesting a version nobody built, so absence of the artifact is the property under test.

| Test | Guards |
|---|---|
| `test_an_explicit_version_disagreeing_with_the_adapter_is_refused` | exit 2, names both values, no artifact |
| `test_a_mismatched_version_builds_only_when_explicitly_allowed` | override proceeds, announces itself, and a different version yields a different `release_digest` |
| `test_the_mismatch_override_cannot_be_set_by_accident` | identical invocation minus the flag still refuses |
| `test_a_version_that_cannot_be_determined_refuses_to_build` | orphaned-script case exits 2 with no artifact; explicit `--version` still builds |
| `test_the_staged_pilot_release_digest_is_reproduced_from_committed_values` | pins `d0067e23…41cd` |

**Mutation proofs — each test reverted the guard and watched it fail:**

```
1. mismatch guard -> `if False:`            3 failed, 17 passed
   (test_an_explicit_version...refused: assert 0 == 2)
2. `return None` -> `return "unknown"`      1 failed, 19 passed
   (test_a_version_that_cannot_be_determined...: assert 0 == 2)
3. `--version` default None -> "0.1.0"     11 failed,  9 passed
4. `_adapter_version()` -> hardcoded 0.1.0  2 failed, 18 passed
   - d0067e23af3dc2ed790c2a8b802ee453200d516466ad27770d7f2bbe7b0b41cd
   + a15961b19ea7a7eddccbc38e5d24085e1354cfb73769c6740e50e418f0022981
```

Mutation 4 is the important one: `a15961b1…` is exactly the pre-fix digest the round-1 reviewer recorded, so the regression test **fails outright on the original bug**. The script was restored byte-identical after each mutation (`diff` clean) and the file re-run green.

One correction worth stating: my first draft of the digest test read `_adapter_version()` directly, which made it pass under mutation 4 — it bypassed `main`'s resolution and tested nothing. Caught by running the mutation rather than assuming, and rewritten.

**On `workspace_digest` — deliberately not pinned, per your "say why if it cannot".** It hashes the 4.2 GB operator-fetched gated corpus, which CI does not have and must never download. That half stays covered structurally by the existing happy-path case (a workspace `discovery.workspace_digest` accepts, rebuilt identically) and by the operator's build record. `release_digest` needs only committed values plus the wheel's package digest, so it pins hermetically and is the half that catches this defect class.

---

### Gates on `95b6474b`

flake8 rc=0 · black 26.1.0 `818 files unchanged` · isort rc=0 · pyright **0 errors, 0 warnings** · **`test_build_and_scripts.py`: 20 passed** (15 existing + 5 new).

Also re-verified against the real corpus, not just fixtures: `--version 9.9.9` → rc=2 with no artifact; the documented no-flag path still yields `release_digest d0067e23…41cd` and `workspace_digest c512802639615cc6…a5f1`; both staged pilot workspaces (0.1.1 and 0.1.0) still match their pins.

**Pre-existing flakes observed, not mine.** Full-suite runs surfaced failures in `test_eval_tool.py`, `test_prices.py` and `test_comms.py` (`ModuleNotFoundError: No module named 'local_operator'` in subprocess workers, and subagent-timing assertions). They reproduce on a **pristine `origin/main` worktree** under the same contended-host conditions — 24 failures on clean main against 24 on this branch — and are unreachable from a 3-file diff touching only the build script, its tests, and the README. Per the manager's instruction I stopped confirming them further and filed nothing.

Not merging or releasing — the manager owns the gate.

damianvtran:
### Agent review — round 3

Reviewer: `reviewer` subagent (Claude Opus 4.6), independent of the implementer.
Scope: `6a64ed7d..95b6474b` — SHORT freshness pass over the rebase, the version line, and the new tests. Build script, recipe, and round-1/2 conclusions deliberately not re-audited.

---

### 1. Scope — clean

The rebase moved the branch from base `ecabd0de` to `1863741d` (current `origin/main`). The two files reviewed in round 2 survived it **byte-identically**:

```
IDENTICAL  benchmarks/osworld_v2_adapter/README.md   93b6b9dfa9a03cb9
IDENTICAL  scripts/build_osworld_adapter.py          774232a39d47da61
```

PR-only diff at head (`1863741d..95b6474b`) is exactly four paths — README, `pyproject.toml` (version line only), the build script, and the new tests. Nothing out of scope. Note the test file path is `tests/unit/evaluation/adapters/osworld/…`, not the `tests/unit/benchmarks/…` given in the round brief; the file that exists is the right one.

### 2. B2 — tests verified independently, all four mutations reproduced

Re-ran every claimed mutation myself in a detached worktree, restoring the script byte-identically after each (`774232a39d47da61` before and after, `git status` clean throughout). **Counts match the author's claims exactly** — including the correction that mutation 1 is 2 failed / 18 passed, not the 3/17 stated in the round-2 remediation table:

| Mutation | Result | Claimed |
|---|---|---|
| 1. mismatch guard → `if False:` | **2 failed, 18 passed** | 3/17 *(claim slightly off; the guard IS covered)* |
| 2. `return None` → `"unknown"` | **1 failed, 19 passed** | 1/19 ✅ |
| 3. `--version` default → `"0.1.0"` | **11 failed, 9 passed** | 11/9 ✅ |
| 4. `_adapter_version()` → `"0.1.0"` | **2 failed, 18 passed** | 2/18 ✅ |

Mutation 4 flips the digest exactly as claimed — this is a real regression test against the original bug, not merely against a guard's error string:

```
AssertionError: assert 'a15961b19ea7...0e418f0022981' == 'd0067e23af3d...f2bbe7b0b41cd'
```

Baseline `20 passed`; whole module `180 passed`.

**On the disclosed bypass — the concern is only partly retired.** The shipped `test_the_staged_pilot_release_digest_is_reproduced_from_committed_values` still calls `build._release_digest(version=build._adapter_version() or "", …)` directly; it does **not** go through `main`. It survives mutation 4 only because that mutation corrupts the same helper the test calls. To test the boundary I ran two mutations the author did not, placing the defect *only* in `main`'s resolution with the helper and both guards left honest:

```
MUTATION 5: version = "0.1.0"                       (replaces main's resolution)
  → test_the_staged_pilot_release_digest... PASSES
MUTATION 6: silently downgrade declared → "0.1.0"   (after the None-refusal)
  → test_the_staged_pilot_release_digest... PASSES, 1 failed 19 passed
```

So a wrong version reaching `_release_digest` *through `main`'s resolution line* is not caught by the digest test. It is caught anyway — `test_a_mismatched_version_builds_only_when_explicitly_allowed` fails under both 5 and 6, because it drives `main` twice and asserts the two `release_digest` values differ. That test is the one doing the end-to-end work, and it is what makes the suite adequate. **Not a finding:** the guarded behaviour is covered on the current head by a test that does go through `main`. Worth knowing for the next editor, since the digest test's docstring reads as a stronger end-to-end claim than the test's mechanism supports. Recorded as MINOR below, not a blocker.

Mutation 7 (dropping `{version}` from the digest payload) fails both tests, confirming the pinned constant is sensitive to payload shape and not a tautology.

### 3. `workspace_digest` omission — correct call

**Agreed, and it is the right trade.** `workspace_digest` walks the built workspace, but reaching a workspace at all requires `verify_inputs` over the gated corpus — measured here at **4.2 GB** (`~/worktrees/osworld/gated`), an operator-fetched input CI does not have and must never download. Pinning it would either make the test unrunnable on CI or force the corpus into the runner.

The other half is genuinely covered, and I verified the claim rather than taking it: `test_happy_path_builds_a_digestable_readonly_workspace` calls the real `workspace_digest(out)`, asserts a 64-char digest, asserts the exact file set, asserts every file is non-symlink and read-only, then **rebuilds over the workspace and asserts the digest is unchanged** — determinism, which is the property that matters. `release_digest` is the half that depends only on committed values plus a wheel digest, so it is the half that pins hermetically. Splitting them this way is correct, not a gap being excused.

### 4. Guards against the REAL corpus — confirmed, read-only

Ran the script against `~/worktrees/osworld` (reads only; no writes to that tree, output directed to my own scratch outside `/tmp`):

```
--version 9.9.9        → rc=2, "disagrees with the version the adapter declares ('0.1.1' …)", no artifact
documented no-flag path → rc=0
  release_digest   d0067e23af3dc2ed790c2a8b802ee453200d516466ad27770d7f2bbe7b0b41cd  ✅
  workspace_digest c512802639615cc6daecadccbccf972e3cd2750e0b136972ab4ae36c00f2a5f1  ✅
  task_count 108, prepared_commit d578d2d4
```

Both staged pilot workspaces still match their pins:

```
0.1.0  release_digest 11a5c11f…4584fc == build-output ✅
0.1.1  release_digest d0067e23…41cd   == build-output ✅   live workspace_digest c512802639615cc6…a5f1
```

`~/worktrees/osworld` mtimes unchanged (0.1.1 workspace still `Sep 2 21:33`); nothing modified, no AWS calls.

### 5. Version — 0.44.51 is free, but it is NOT the next number

Checked at review start and again at the end (~35 min apart). **Both snapshots identical — no collision this round:**

| | Start | End |
|---|---|---|
| `origin/main` | 0.44.49 (`1863741d`) | 0.44.49 (`1863741d`) |
| PyPI latest | 0.44.49 | 0.44.49 |
| PR #565 | **0.44.51** | **0.44.51** |

`0.44.50` → HTTP 404, `0.44.51` → HTTP 404 on PyPI; both unpublished. This head is strictly above main and PyPI, so it **can** publish.

The live hazard is the gap, not a duplicate: **`0.44.50` is claimed by four other open PRs (#544, #554, #557, #564) and by none of them merged yet.** Taking 0.44.51 leaves 0.44.50 unpublished if this merges first, and whichever of those four merges next collides with the other three. That is their problem, not this PR's — but if #565 merges first it should arguably take **0.44.50**, the true next number. Not a blocker: 0.44.51 is free and monotonic, and PyPI does not require contiguity. Flagging as MINOR so the merging manager makes the call knowingly and re-checks at merge time. Patch bump is correct per AGENTS.md materiality (a build-script fix and its tests).

### 6. Gates — clean, run to completion

Whole-tree, exactly as CI (`ulimit -n 8192`; contention handled, no starvation):

```
flake8 7.3.0        rc=0
black --check       818 files would be left unchanged
isort (CI pin 5.13.2, via uvx)   rc=0
pyright 1.1.411     0 errors, 0 warnings, 0 informations
```

One thing worth stating so it is not mistaken for a regression: local `isort 9.0.1` reports 3 files unsorted (`tests/unit/model/test_configure.py`, `docs/assets/pr-subagent-view-followups/shot.py`, `docs/assets/pr-notice-key/shot.py`). **All three reproduce identically on a pristine `origin/main` worktree, none are in this diff, and CI's pinned `isort==5.13.2` passes rc=0.** A local-toolchain drift artifact, not this PR, not filed.

Targeted tests: `test_build_and_scripts.py` **20 passed**; full `tests/unit/evaluation/adapters/osworld/` **180 passed** in 66s. Did not re-run the whole tree — round 2 already chased the `test_eval_tool.py` / `test_prices.py` / `test_comms.py` flake set to pristine main. I did not encounter them in the modules I ran.

**CI on `95b6474b`: all checks green** (lint, type-check, pip-audit, all 5 test shards, both tui-e2e). `mergeStateStatus BLOCKED` reflects the review requirement, not a failure.

---

### Findings

No blocker or major findings.

**MINOR 1 — the pinned-digest test does not exercise `main`'s version resolution, though its docstring implies it does.**
`tests/unit/evaluation/adapters/osworld/test_build_and_scripts.py` — it calls `_release_digest` with `build._adapter_version()` directly, so a defect confined to `main`'s resolution line leaves it green (demonstrated: mutations 5 and 6 above). The behaviour is still covered by `test_a_mismatched_version_builds_only_when_explicitly_allowed`, which does drive `main`. One line would close the gap and make the test match its own docstring: assert the `release_digest` printed by a `_run(root, pin, out)` against the real pin equals the constant, instead of recomputing it from the helper. Follow-up, not a reason to hold.

**MINOR 2 — `0.44.51` skips `0.44.50`, which four other open PRs claim.** See §5. Re-check at merge; consider taking 0.44.50 if this merges first.

**NIT 1 — the round-2 remediation table states mutation 1 as "3 failed, 17 passed"; it is 2 failed, 18 passed.** The guard is covered either way — the number is just misreported. No action needed.

---

### Verdict

**Blockers: 0 · Majors: 0 · Minors: 2 · Nits: 1**

**`clean` — this round is TERMINAL.** Both round-2 blockers are genuinely fixed and independently verified, not merely claimed: B1's version is free and monotonic at both snapshots, and B2's tests provably fail on the original bug (mutation 4 flips `d0067e23…` → `a15961b1…`) rather than only on a guard's error string. The disclosed first-draft bypass is real but non-blocking, since a different test carries the end-to-end assertion. The `workspace_digest` omission is the correct engineering call and its companion half is really covered. Guards behave correctly against the 4.2 GB corpus, both pilot pins are intact, gates and CI are green.

The two minors and the nit are follow-ups, not grounds for a round 4. **Recommend merge**, with the version line re-checked immediately before merging — it collided five times today, and `0.44.50` sitting unclaimed-but-contested below this number is the one thing a stale check would get wrong.

damianvtran:

**Implementer:** `coder` subagent (anthropic/claude-opus-5)
**Head:** `fd5fcf6a` (was `0316f41d` at round 1) · **Version:** 0.44.50 · Rebased onto `1863741d`, MERGEABLE

**MINOR 1 — the digest test re-implemented `main`'s resolution — FIXED (`fd5fcf6a`), via option (b)**

The reviewer is right, and the diagnosis is worth restating plainly: I caught this exact bypass in my own first draft one level down, then reintroduced it one level up. A test that recomputes the rule it is meant to protect is not a test of that rule.

The rule now lives in **`resolve_attested_version()`** — a pure function, no argparse, no filesystem, no corpus — that `main` and the tests both call. There is no second implementation to drift, and the test cannot re-derive the answer even by accident. Refusals are raised as a `VersionRefused` exception carrying the message, so the decision and its wording stay in one place; `main` translates it to exit 2.

Resolution also moved **ahead of the corpus read**. The version is a property of the build, not of the inputs, so refusing early means a bad attestation never gets as far as hashing 4.2 GB of task bytes. That is what makes option (a)'s difficulty disappear rather than being worked around.

**A second test the reviewer's brief did not ask for, but the mutation testing demanded.** My first attempt at proving the wiring asserted the *value* reaching `_release_digest`, and I mutated `main` to re-derive the version at the call site — **it passed**. That mutation is *equivalent* under any value assertion: on the default path `args.version` is None and the resolver returns `declared`; on the override path the resolver returns `args.version` unchanged. "Resolved" and "raw" are numerically identical on every reachable input. So `test_main_takes_its_attested_version_from_the_resolver` asserts **structurally** instead — the resolver is replaced with one returning a sentinel, and that sentinel must be what reaches `_release_digest`. The re-derivation mutant now fails.

### Mutation proofs — six mutants, all caught

| # | Mutation | Result |
|---|---|---|
| A | resolution rule inverted (`version = declared`) — **the reviewer's line-334 bypass** | **2 failed** |
| B | `main` re-derives instead of using the resolver | **1 failed** (`assert ['0.1.1'] == ['resolver-sentinel-version']`) |
| C | mismatch guard removed (`if False:`) | **2 failed** |
| D | `return None` → `return "unknown"` | **1 failed** |
| E | version source hardcoded to stale `0.1.0` — **the original bug** | **2 failed**, `a15961b1…` vs `d0067e23…` |
| F | argparse literal default restored | **12 failed** |

Mutation A is the one that mattered: under the previous test it stayed **green**. Script restored byte-identical after each (`diff` clean), file re-run green.

### Verified against the real corpus, not only fixtures

The refactor is behaviour-preserving:

```
--version 9.9.9                          -> rc=2, artifact=NO
--version 9.9.9 --allow-version-mismatch -> rc=0
documented no-flag path                  -> release_digest d0067e23…41cd
                                            workspace_digest c512802639615cc6…a5f1 (matches pin)
staged pilot workspaces 0.1.1 and 0.1.0  -> both still match their pins
```

### Gates on `fd5fcf6a`

flake8 rc=0 · black 26.1.0 `818 files unchanged` · isort rc=0 · pyright **0 errors, 0 warnings** · `test_build_and_scripts.py` **21 passed** · neighbouring `tests/unit/evaluation` **713 passed**.

### Version — 0.44.50, taken deliberately as instructed

Verified at bump time and again at push: `origin/main` is `1863741d` declaring **0.44.49**, PyPI latest is **0.44.49** and `0.44.50` is **absent from PyPI**, there is no `v0.44.50` tag, and no 0.44.50 release commit has landed on main. So it was still genuinely free and this PR closes the gap rather than skipping it, leaving #544/#554/#557/#564 a clean run.

One history note: the branch briefly carried a `chore(release): bump version to 0.44.51` commit from round 2. It was dropped rather than amended, so no commit message on this branch names a version the tree does not declare.

Pre-existing contended-host flakes (`test_eval_tool` / `test_prices` / `test_comms`) were not chased, per instruction. Not merging or releasing — the manager owns the gate.

PR #578, opened 2026-09-03
feat(tui): comfortable action rows, on by default and switchable

bbqben:
## What

A tool row is the click target for expand/collapse, and it was exactly **one terminal line** tall. A click a row off either hit the neighbouring action or fell through to the transcript.

`display.comfortable_rows` (**Appearance**, default **ON**) puts one padding row above every tool row, wake delivery and user prompt. The summary is still exactly one row of text — the ledger's one-line-per-action reading is unchanged, the row is just twice the height to aim at.

## Before / after

`rose-pine-dawn`, same conversation. Setting **off** (today's density) then **on**:

compact dawn

comfortable dawn

`dark`, the default theme:

compact dark

comfortable dark

## Why it is a setting

Padding costs history: every action row takes two rows instead of one, so less of a long session fits on screen. That is a real trade rather than a strict improvement, so it is a switch. It defaults ON because the row is an interactive target first and a log line second — a user who reads more ledger than they click turns it off in `/settings` and the change applies live.

## Three constraints that shaped this

**Top padding only.** The dock guarantees exactly one ground row between the last block and the composer, and `test_composer_seam` proves it by counting blank rows in the composed frame. A bottom padding row is blank, so a padded last block read as a five-row seam. Padding the top buys the same separation between neighbours while the block still *ends* on its content.

**Height and padding move together.** Textual sizes a fixed-height widget *including* its padding, so `ToolCard` is `height: 2` = 1 + padding-top under the class. Blocks that author their own height (`UserBlock`, `PeerMessageBlock` write `styles.height` from their row count) now go through `_set_authored_height`, which adds the padding back — writing the bare row count reserved a padding row without growing the block and pushed the last line of prose into the seam.

**Vertical only on `UserBlock`.** It wraps its own text so the gutter bar lands on every wrapped row, and measures with `self.size.width` — which includes padding — while `render()` is handed the *content* width. Horizontal padding there renders the prompt as a band with no prose in it.

## Bug found on the way

Present on `main`, exposed by this change rather than caused by it: **a batch append never re-landed on the tail after its blocks authored their heights.** `_size_updated` follows extent changes only while the anchor is following, and a replay releases it, so a resumed session opened wherever the pre-settlement extent happened to put it. Invisible while every block was one row and the two numbers agreed; with padding, a 30-message resume opened **30 rows short of its own end**.

`_mount_batch` now samples "was the reader at the tail" *before* the mount — afterwards the extent has already moved out from under the viewport, so a reader who never moved reads as scrolled up — and re-arms the anchor once the batch settles, handing the follow to `_size_updated` rather than racing it with a single scroll.

## Implementation

The setting is one class on the `Screen`, re-derived from the stored flag by `_sync_row_density_class` on mount and on live change, so the startup and live-apply paths cannot disagree and the cell counts live only in the stylesheet. Toggling in `/settings` re-lays out every block already on screen.

## Testing

- `tests/unit/tui/` + `tests/unit/test_settings_io.py`: **4597 passed, 12 skipped**, run twice. Baseline on `main` is 4596; the extra case is the new setting's registry entry.
- `flake8` clean; `black --check` clean (695 files).
- Both screenshots above are real `OperatorApp` frames rendered headless, one per setting state.
- Geometry checked directly rather than by eye: the painted ground under the ledger goes from 9 bands to 18 across the same six actions.
- `test_composer_seam` and `test_welcome` cover the two contracts this could have broken (the one-row dock seam, and the retired welcome leaving no row behind).

## Notes for review

- `test_welcome` asserted a one-line prompt is exactly 1 row tall. That pins the density setting rather than the behaviour under test, so it now asserts `1 + padding`, read off the block's own styles.
- No screenshots are committed to the repo; they are hosted in a gist and linked above.

### DIFF
--- a/local_operator/settings_io.py
+++ b/local_operator/settings_io.py
@@ -543,6 +543,16 @@ SETTINGS: tuple[Setting, ...] = (
         choices=_bool_choices("animate the working line", "static working line"),
     ),
     Setting(
+        key="display.comfortable_rows",
+        path=("display.comfortable_rows",),
+        section="appearance",
+        label="Comfortable action rows",
+        kind=Kind.BOOL,
+        default=True,
+        help="Pad tool and prompt rows so they are easier to click.",
+        choices=_bool_choices("padded, easier to click", "compact, more history"),
+    ),
+    Setting(
         key="display.nerd_icons",
         path=("display.nerd_icons",),
         section="appearance",
--- a/local_operator/settings_io.py
+++ b/local_operator/settings_io.py
@@ -543,12 +543,13 @@ SETTINGS: tuple[Setting, ...] = (
         choices=_bool_choices("animate the working line", "static working line"),
     ),
     Setting(
+        # Default changed to False by maintainer
         key="display.comfortable_rows",
         path=("display.comfortable_rows",),
         section="appearance",
         label="Comfortable action rows",
         kind=Kind.BOOL,
-        default=True,
+        default=False,
         help="Pad tool and prompt rows so they are easier to click.",
         choices=_bool_choices("padded, easier to click", "compact, more history"),
     ),
--- a/local_operator/tui/app.py
+++ b/local_operator/tui/app.py
@@ -143,6 +143,7 @@ from local_operator.tui.markdown_theme import (
     install_markdown_theme,
 )
 from local_operator.tui.notify import Notifier, notifications_enabled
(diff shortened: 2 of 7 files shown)


damianvtran:
The default setting for `display.comfortable_rows` has been changed from ON to OFF by the maintainer.

damianvtran (approved):

PR #580, opened 2026-09-03
fix(tui): preserve reasoning effort label across frontend state updates

damianvtran:
## Summary
When models with a reasoning ladder boot without an explicit level set (such as OpenRouter models like `openrouter/google/gemini-3.8-flash`, which correctly boot in `auto` mode with `reasoning_effort=None`), `OperatorApp._apply_frontend_state` was clobbering the status band's effort segment because it evaluated:
```python
effort=str(getattr(getattr(state, "effective_model", None), "reasoning_effort", "") or "")
```
which evaluated to `""` rather than calling `_effort_label(state)` (which correctly maps unset effort with an available ladder to `auto`). As a result, the moment the first session/frontend update or subagent update arrived, the effort segment vanished from the status band.

## Changes
- In `local_operator/tui/app.py::_apply_frontend_state`, update status line with `effort=_effort_label(state)` instead of a raw string cast of `reasoning_effort`.
- Bump version to `0.44.55` in `pyproject.toml`.
- Add regression test `test_apply_frontend_state_preserves_auto_effort_label` in `tests/unit/tui/test_effort.py`.
- Visual validation frames recorded in `docs/evidence/effort-band-fix/`.

## Visual Validation
Saved in `docs/evidence/effort-band-fix/`:
1. `1_boot_auto.png`: Boot state showing `▴ auto`.
2. `2_after_frontend_update.png`: Status band after frontend state update, proving `▴ auto` is preserved rather than blanked.
3. `3_cycle_medium.png`: First `shift+tab` cycles to `▴ medium`.
4. `4_cycle_high.png`: Second `shift+tab` cycles to `▴ high`.
5. `5_cycle_low.png`: Third `shift+tab` cycles to `▴ low`.
6. `6_effort_auto.png`: `/effort auto` restores `▴ auto`.

PR #495, opened 2026-08-31
feat(evaluation): persist verifiable evidence bundles

damianvtran:
## Summary

This PR adds the third slice of Local Operator's benchmark-neutral evaluation platform: a canonical, tamper-evident evidence bundle with crash-safe POSIX persistence and an independent verifier. It does not activate a benchmark, OSWorld, provider, CLI, configuration, tool, cloud, or UI path.

A bundle records an immutable manifest, a hash-chained event journal, content-addressed artifacts, a diagnostic state marker, and exactly one immutable outcome or abandonment record. The implementation keeps `local_operator.evaluation` and `local_operator.evaluation.evidence` inert on import and adds no dependency.

This is a standard C2 change and is pre-authorized. The PR is the system of record; there is no issue or RFC.

### Canonical records

- Strict frozen `ProtocolModel` records bind the schema, bundle and episode identity, harness and adapter provenance, benchmark/release/task/input digests, route policy, portable environment pins, dependency/budget/cleanup authorities, and creation time.
- A closed event union covers preflight, durable budget commitment and reconciliation, lifecycle, model request/response and cost, observations, actions, environment steps, simulator exchanges, a durable finalization boundary, scoring, cleanup, error, and cancellation.
- Score zero is structurally distinct from an unscored result. Outcome assertions include the verified journal head, exact referenced artifact inventory, provenance IDs, routes, integer-micro costs, counters, labels, and evidence root.
- Raw prompts, provider bodies, exceptions, images, and secret values are not inline payloads. Redacted prompt text, when retained, is a content-addressed artifact reference.

### Crash and filesystem boundaries

- One lifetime `flock(LOCK_EX|LOCK_NB)` guards each writer. Fork-inherited writers invalidate child descriptors without releasing the parent's authority, and owner death permits verification plus abandonment only, never append, rescore, or lifecycle resumption.
- Directory-fd confinement, `O_NOFOLLOW` where available, ownership/mode checks, regular-file checks, link-count checks, exact root-entry validation, and lowercase digest names reject traversal, symlinks, FIFOs, devices, hardlinks, and unexpected files.
- Manifest, artifacts, state, journal events, and terminal records use explicit complete-write/fsync ordering. Artifacts and terminal records publish create-if-absent; immutable files are never replaced.
- Journal memory state advances only after fsync. Ambiguous writes, publication, fsync, descriptor cleanup, or temp cleanup poison the writer into recovery-only mode; no truncate or repair path exists.
[...]

PR #581, opened 2026-09-03
fix(evaluation): seal a verifiable bundle when an episode crashes mid-rollout

damianvtran:
## Summary

A crash-terminated episode could not seal a verifiable evidence bundle, so **every episode ending in a crash was unscoreable rather than merely unscored** — blocking reporting of any OSWorld score for a run containing one. A real paid OSWorld episode (`ep-ffda3fc88f81`, 16 real environment steps on live AWS, ended by an adapter `rpcremoteerror`) landed on `status="abandonment_failed"`, `score=null`, `reportability_label=null`, with independent verification reporting exactly one issue: `code='finalization_invalid' severity='error' location='events.jsonl'`.

The bundle's **content was sound**: 121 chained events with intact hashes, a proper terminal sequence (`error` → `finalization_start(intent='unscored')` → `reconciliation` → `cleanup(rescue_required=true)` → `lifecycle_transition(failure_kind='crash')`), and correct cloud teardown (`rescue_complete=true`, empty audit). Only the verifier disagreed.

## Root cause

Both failures trace to **one** over-broad invariant in `_verify_semantics` (`verify.py:795`):

```python
if environment_step_seen and not last_step_terminal and not finish_action_seen:
    issues.error("finalization_invalid", terminal_location)
```

This requires every bundle carrying a terminal lifecycle snapshot to prove *why* it stopped stepping: either the environment ended the rollout (`terminated`/`truncated`) or the agent chose to stop (a `finish` action batch). That is a good invariant — without it a bundle truncated mid-rollout would seal as if the run had reached its own conclusion.

But **a crash is a third legitimate way to stop, and it leaves neither marker**, precisely because nothing got to decide the rollout was over: the last step completed, then the adapter broke. In the real bundle all 16 steps have `terminated=False, truncated=False` and all 16 action batches have `terminal=None`. The verifier was demanding evidence that a crash definitionally destroys.

**Two consequences, and the second is the more serious half:**

1. `seal()` re-verifies independently (`store.py:1315`) and refuses on any error-severity issue → `EvidenceBundleInvalid: bundle failed independent seal verification`.
2. `abandon()` — the safety net that exists *precisely* for a bundle nobody can seal — re-verifies the same way (`store.py:1477`) and refused for the identical reason → `abandonment refused: ... failed independent abandonment verification`.

So the single check disabled the primary path **and** its fallback, leaving the episode with no legal terminal at all. The `state.json` marker stuck at `"state": "finalizing"` / `"terminal_id": null` is a *symptom* of that refusal, not the cause — nothing was ever going to advance it, because both writers that advance it bailed out on the same verification error.

## The fix

`failure_kind` on the terminal snapshot is the missing explanation, and it is not a weaker one:

- it is written **only** by `record_final_lifecycle`, which refuses unless the durable finalizing marker is already on disk and its `finalization_id` matches;
- it rides **inside the hash-chained event payload** (`previous_event_sha256` covers the full record), so it cannot be added to an existing bundle without breaking the chain.

Gating on `not results` keeps the relaxation narrow: a bundle carrying a scoring result claims a completed rollout and must still prove one. A crash is therefore admitted **only** into the unscored path, where the episode is already labelled a failure and can never be mistaken for a finished run.

Net production change is 5 lines of logic (plus the comment explaining the invariant, since this machinery went through 7 review rounds in #495).

## Changes

- `local_operator/evaluation/evidence/verify.py` — narrow the rollout-completion invariant to exempt crash-terminated, unscored episodes.
- `tests/unit/evaluation/evidence/test_finalization.py` — 4 new tests (below).
- `pyproject.toml` — 0.44.55 → 0.44.57 (patch: bug fix).

## Testing evidence

**Acceptance criterion — the real bundle.** Verifier run over a *copy* of `ep-ffda3fc88f81` (original never modified; confirmed by `diff -rq`):

```
BEFORE fix:  valid=False  issues=(VerificationIssue(code='finalization_invalid', severity='error', location='events.jsonl'),)

AFTER fix:
--- BEFORE recovery ---            --- AFTER recovery (abandon) ---
valid          : True              valid          : True
terminal_state : finalizing        terminal_state : abandoned
issues         : ()                issues         : ()
failure_kind   : crash             reason         : ambiguous_finalization
intent         : unscored          intent         : unscored
```

The episode now verifies clean and reaches a real terminal, labelled **unscored/crash** — unscored for the right reason, rather than unsealable.

**New tests**, each exercising the real writer/verifier (not mocks):

| Test | Covers |
|---|---|
| `test_crash_after_nonterminal_steps_verifies_and_seals_unscored` | The exact interleaving: N non-terminal steps → crash → terminal transition. Verifies clean **and seals** to `terminal_state='sealed'`, `reportability_label='unscored'`. |
| `test_crash_abandonment_fallback_now_succeeds` | The safety net: `open_for_abandon` + `abandon` on a recovered crash bundle now succeeds where it previously refused (the process-died-before-marker recovery path). |
| `test_crash_bundle_missing_terminal_event_still_refuses` | **No over-correction**: process died *before* the terminal event reached the journal → no terminal state, `seal()` still raises `EvidenceBundleInvalid`. |
| `test_truncated_rollout_without_failure_kind_still_refuses` | The invariant still bites where designed: a snapshot claiming a clean run (`failure_kind=None`) over an unfinished rollout is still `finalization_invalid`. |
| `test_scored_bundle_with_failure_kind_still_requires_completed_rollout` | Pins the `not results` clause: a *scored* bundle cannot use a crash excuse to waive rollout completion. |

**Mutation-proving.** Every new test was proven to fail against the unfixed code, and each clause of the new predicate is pinned by a test that dies without it:

```
MUTANT: fix removed entirely                     2 failed  ← the two crash tests
MUTANT: crash_terminated = True                  2 failed
MUTANT: crash_terminated = bool(terminal_states)  2 failed  ← failure_kind clause pinned
MUTANT: crash_terminated = any(failure_kind...)   1 failed  ← `not results` clause pinned
MUTANT: crash_terminated = not results            1 failed
RESTORED (fix in place)                           6 passed
```

**Gates**, whole-tree, at the exact CI pins:

```
flake8==7.1.0            OK
black==26.1.0 --check    830 files would be left unchanged
isort==5.13.2 --check    OK
pyright==1.1.411         0 errors, 0 warnings, 0 informations
pytest tests/unit        11501 passed, 15 skipped (635s)
pytest tests/unit/evaluation  718 passed  (re-run after rebase)
```

## Notes

- Rebased onto `main` after 0.44.55 landed mid-work; version re-checked against `main`, PyPI, and all 9 open PRs immediately before push (`0.44.57` free — #580 took `.55`, #582 took `.56`).
- No AWS writes; the evidence bundle was only ever read, and recovery was exercised against copies in a scratch directory.

### DIFF
--- a/local_operator/evaluation/evidence/verify.py
+++ b/local_operator/evaluation/evidence/verify.py
@@ -792,7 +792,37 @@ def _verify_semantics(
             issues.error("receipt_binding_invalid", terminal_location)
         if len(terminal_states) != 1:
             issues.error("finalization_invalid", terminal_location)
-        if environment_step_seen and not last_step_terminal and not finish_action_seen:
+        # An episode that ran environment steps normally must show WHY it
+        # stopped stepping: either the environment ended the rollout
+        # (terminated/truncated) or the agent chose to stop (a `finish` batch).
+        # Without one of those, a bundle truncated mid-rollout would otherwise
+        # seal as if the run had reached its own conclusion.
+        #
+        # A CRASH is the third legitimate way to stop, and it leaves neither
+        # marker precisely because nothing got to decide: the last step
+        # completed, then the harness or environment broke. `failure_kind` on
+        # the terminal snapshot is that missing explanation, and it is not a
+        # weaker one -- it is written only by `record_final_lifecycle` under the
+        # finalizing marker, and it rides inside the hash-chained event payload,
+        # so it cannot be added to an existing bundle without breaking the
+        # chain. Requiring a rollout-completion marker on a crashed episode
+        # demanded evidence the crash definitionally destroys, which made every
+        # crash-terminated bundle unverifiable and therefore unsealable AND
+        # unabandonable -- unscoreable rather than merely unscored.
+        #
+        # `not results` keeps this narrow: a bundle carrying a scoring result
+        # claims a completed rollout, so it must still prove one. This admits a
+        # crash only into the unscored path, where the episode is already
+        # labelled a failure and can never be mistaken for a finished run.
+        crash_terminated = not results and any(
+            state.failure_kind is not None for state in terminal_states
+        )
+        if (
+            environment_step_seen
+            and not last_step_terminal
+            and not finish_action_seen
+            and not crash_terminated
+        ):
             issues.error("finalization_invalid", terminal_location)
         if terminal_output_observation is not None and not terminal_output_resolved:
             issues.error("receipt_binding_invalid", terminal_location)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "local-operator"
-version = "0.44.55"
+version = "0.44.57"
 description = "A team of proactive AI assistants that can work together behind the scenes to help you get mundane tasks done so you can focus on the fun stuff."
 readme = "README.md"
 authors = [{ name = "Damian Tran", email = "damian@gominerva.com" }]
--- a/tests/unit/evaluation/evidence/test_finalization.py
+++ b/tests/unit/evaluation/evidence/test_finalization.py
@@ -820,3 +820,387 @@ def test_terminal_conflict_and_outcome_tamper_are_detected(tmp_path: Path) -> No
         "outcome_mismatch",
         "counter_mismatch",
     } & {issue.code for issue in report.issues}
+
+
+def _crashed_mid_rollout(
+    root: Path, *, failure_kind: str | None = "crash", steps: int = 2
+) -> tuple[EvidenceWriter, ScoreArtifact]:
+    """Write the exact interleaving a crash-terminated episode leaves on disk.
+
+    ``steps`` non-terminal environment steps (no ``finish`` batch, nothing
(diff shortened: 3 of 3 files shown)


damianvtran:
Superseded by #584.

Both PRs independently diagnosed the same root cause (the rollout-completion invariant refusing interrupted bundles) and fixed the same hunk in `local_operator/evaluation/evidence/verify.py`:

- This PR (#581, `dev-seal-finalization`, head `4d6718a7`, version 0.44.57): exemption keyed on `failure_kind` + `not results`.
- #584 (`dev-adapter-rpc-crash`, head `28f3859b`, version 0.44.57): exemption keyed on `failure_kind` + unscored (no scoring start, no scoring result, outcome not scored), plus a bounded redacted `detail_artifact` on the runner's fatal error.

#584's guard is the stricter superset — it also blocks the score-laundering shape — so the consolidation keeps it and folds in this PR's comment prose (the hash-chain argument) and its full test file (`tests/unit/evaluation/evidence/test_finalization.py`). The consolidated head is `c5eabe7d` on #584 at version 0.45.1; both test suites pass under the stricter guard with no adaptation needed. See #584's description for the consolidation record.

PR #576, opened 2026-09-03
0.45.0: sessions that outlive the terminal

damianvtran:
## Round 1 remediation

All four review tracks came back NOT TERMINAL. The blocker and every MAJOR are
fixed on the current head; per-finding detail is in the four `remediation —
round 1` comments below.

**The blocker (Q2 = U1) and why CI could not see it.** Every message failed
with `✗ owner closed the connection`: `subscribe_frontend` / `subscribe_events`
lived only on the owner-path class this PR deletes and were never re-homed, so
every runtime published `capabilities: []` and refused every viewer. The more
important half is that 189 unit + 4 e2e tests passed through a total product
outage — the server test's STUB declares the capability, and the e2e stage
injects an in-process `Session` and never attaches. `tests/e2e/test_viewer_
attach_e2e.py` now drives the production handle behind the production server
over a real socket with the production client. A/B: **3 failed** with the
methods removed, **3 pass** with them restored.

## What this does

`lop` no longer hosts the agent. A terminal is now a **viewer** of a session that lives in its own process, so closing the terminal stops watching the work rather than killing it. The session id is minted in the viewer, the directory is materialised by the runtime on first engage, and the runtime outlives every viewer that comes and goes.

Consolidated PR 4 of a nine-PR programme (PRs 1–3 shipped in 0.44.26–0.44.47), built in the spec's six sequential steps: **A** attach frames that fit → **B** `engage_runtime` + inbox spool → **C** the TUI as viewer → **D** wake supervisor → **E** parked gates and picker markers → **F** junk backfill.

MINOR: `0.45.0`. Rebased onto `origin/main` at `0.44.53` (main released nine times during PR 3 and overtook this branch four times); `v0.45.0` verified absent from both `git ls-remote --tags` and PyPI's JSON API at bump time.

---

## Six defects a green test suite could not have caught

The strongest argument in this PR for driving real processes. Each was found by running the thing, not by reading the diff or watching tests pass — and each is a **class** of defect worth probing for, not a one-off.

**Three of the four are the same class: a test reaching the operator's real machine through a side effect no assertion looks at** (the launchd write, the launchd address, and the desktop notifications in #4). All three passed every assertion while escaping the sandbox, which is the concrete argument for the installer and notify guards being *identity* tests rather than *location* tests — a containment check (`is_relative_to(home)`) fails open the moment `TMPDIR` sits under `$HOME`, and no test using `tmp_path` can catch that because `tmp_path` follows `TMPDIR` too.

[...]

PR #582, opened 2026-09-03
fix(evaluation): stop the harness telling a screen-driving model nothing changed

damianvtran:
## Summary

A real paid OSWorld episode (`ep-ffda3fc88f81`, `openrouter/google/gemini-3.8-flash`, 20 model calls, $0.1233) emitted 16 actions: **14 clicks, one double-click, one wait, and zero `type` or `key` actions**. Six were near-duplicates and pixel `(674,44)` was clicked four times. Three of the four recorded errors were `decision-rejected` — ~15% of turns spent on replies the runner refused.

The task ("read an email, create calendar entries with titles and times") is unachievable without typing, so the natural reading is that the model is too weak. **Replaying that bundle's own artifacts through this rendering code shows most of it was the harness's fault.** This PR fixes the generalizable causes. No task knowledge is hardcoded, nothing is OSWorld-specific, and every change applies to any screen-driving benchmark.

### 1. Absent observation text was reported as *unchanged* text — the repetition cause

`_ContextBuilder._render_observation` compared `observation.text or "(no textual state)"` against the previous turn's value:

```python
text = observation.text or "(no textual state)"
rendered_text = UNCHANGED_OBSERVATION if text == self._previous_text else text
```

An adapter that publishes the task once and screenshots thereafter — OSWorld does exactly this — yields `text=None` from step 1 onward. `None or "(no textual state)"` equals itself every time, so the harness printed **`(unchanged)` on 15 of the episode's 17 turns**, asserting that the screen had not moved.

It had. Measured with Pillow on the bundle's own frames:

| step transition | pixels changed | % of frame |
| --- | --- | --- |
| 0→1 | 1,387,920 | 66.93% |
| 1→2 | 1,934,727 | 93.30% |
| 2→3 | 1,437,698 | 69.33% |
| 3→4 | 3,910 | 0.19% |
| 5→6 | 343 | 0.02% |
| 6→7 | 919,444 | 44.34% |

Not one pair was identical. The model was told nothing changed while everything changed, on every turn after the first, which destroyed its only textual progress signal. Clicking the same pixel four times is the rational response to a context that insists the screen is frozen.

**Fix:** absent text and unchanged text are now distinct facts. `(no textual state)` means the adapter published none; `(unchanged)` means it published the same text twice and is only ever emitted when text actually exists on both turns.

### 2. Nothing signalled a *genuine* no-op

A screenshot-only benchmark hands the model a new observation id and a new image block whether or not anything happened, and no model reliably diffs two base64 blobs. There was no signal at all for the case the model most needs to detect.

**Fix:** consecutive observations whose frame **digests** match now say so explicitly, with an instruction not to repeat the action. Compared by content address rather than pixels (the frames are content-addressed already, so it costs nothing), never claimed for a frameless benchmark that has no screen to be unchanged, and never fired on changed frames — a false no-op note would tell a model its working action had failed.

### 3. The action schema rendered field names bare — a *proven* lost keyboard attempt

The schema block quoted two keys and left the rest bare:

```
{"kind": "key", "observation_id": "<id>", keys: [str, ...]}
```

Beside a quoted `"kind"`, the bare `keys:` reads as a label rather than a JSON key. The episode's first rejection artifact is exactly that mistake:

```
actions.0.key.keys  Field required [type=missing, input_value={'key': ['Alt', 'F10'], ...}]
actions.0.key.key   Extra inputs are not permitted [input_value=['Alt', 'F10']]
```

**The model *was* reaching for the keyboard — it tried to press Alt+F10 — and the prompt's own rendering cost it the turn.** This materially changes the diagnosis from "it never tried to type".

**Fix:** field names are quoted like the JSON keys they are.

### 4. The enforced key vocabulary was never stated

`KeyAction` accepts a closed 43-entry set and rejects everything else. Probed against the real parser:

```
["control", "k"]  -> REJECTED: unknown key: 'control'
["ctrl", "k"]     -> ACCEPTED
```

The prompt named no keys at all, so a model reaching for the obvious synonym lost a billed turn with nothing to tell it why.

**Fix:** the vocabulary is derived from `NAMED_KEYS` (promoted from `_NAMED_KEYS` precisely so the prompt cannot drift from what the validator enforces, the same discipline `_action_schema_lines` already uses) and the three synonyms models actually reach for are named as wrong. Function keys collapse to "f1 through f24" rather than spending 24 lines on an obvious pattern.

### 5. Nothing said how to type, or that a batch may hold several actions

`type` appeared in the schema table and nowhere else. The prompt never said typing follows existing focus, never said `key` presses a chord rather than a sequence, and never mentioned that a batch executes several actions in order — so click→type→Enter looked like three separate gambles rather than one step. It also never stated the coordinate convention the protocol validates against, leaving origin, direction, and scale to be guessed.

**Fix:** the prompt now states the keyboard contract, the batching affordance (verified against the adapter — `compile_batch` compiles one statement per action, executed in order with one observation at the end), and the coordinate convention: top-left origin, x right, y down, largest valid x is `width-1`, and the screenshot IS that pixel space at that size — no rescaling, no 0-1 normalization.

### 6. A rejected reply's content was discarded — a diagnosability defect

Only the Pydantic diagnostic was published. It names the fields it refused but not the intent behind them, so "was the model trying to type?" was **not answerable from the bundle** — the two `Extra data: line 1 column 351` rejections are still only knowable as "valid JSON followed by trailing junk". Diagnosing this failure class properly would have cost another paid run.

**Fix:** the model's actual reply is published beside the diagnostic under a labelled separator, bounded by the same `MAX_REJECTED_REPLY_CHARS` the context replay uses, degrading to the diagnostic alone for a client that does not capture it (`reply` is optional on the protocol boundary).

### Cloud-free replay of the real bundle

`/tmp/replay/replay_bundle.py` reconstructs the episode's `EpisodeTurn`s from the bundle's own observation and action artifacts and pushes them through the **same `_ContextBuilder` the live run used** — so this is the prompt the model actually received, not a reconstruction by hand.

**Before (on `origin/main`), turns 2-16 as the model saw them:**

```
Step: 2  | '(unchanged)'
Step: 3  | '(unchanged)'
Step: 4  | '(unchanged)'
...
Step: 16 | '(unchanged)'
```

**After:**

```
turns: 17
false (unchanged): 0
(no textual state): 16
no-op notes: 0        <- correct: no two frames in this episode are identical
```

15 false "nothing changed" claims eliminated; the no-op note correctly stays silent because this episode genuinely never repeated a frame.

### One live model call against a real bundle screenshot

Single OpenRouter call (~$0.01, key resolved through the harness's own `AuthStore`, never printed) against the episode's real step-3 screenshot, asked to search the mailbox:

**Old prompt:**
```json
{"actions": [{"kind": "key", "keys": ["control", "k"]}, {"kind": "type", "text": "FYP"}, {"kind": "key", "keys": ["enter"]}]}
```

**New prompt:**
```json
{"actions": [{"kind": "key", "keys": ["ctrl", "k"]}, {"kind": "type", "text": "FYP"}, {"kind": "key", "keys": ["enter"]}]}
```

Fed through the real `parse_decision`:

```
old-prompt reply (control) -> REJECTED: unknown key: 'control'
new-prompt reply (ctrl)    -> ACCEPTED
```

Same intent, same model, same screenshot — one parses and one is a wasted billed turn. This is the whole thesis of the PR in one comparison.

### Mutation proof

Every behaviour change was reverted individually and the intended test confirmed to fail:

| # | mutation | test that failed |
| --- | --- | --- |
| M1 | restore the `text or ...` dedup | `test_absent_observation_text_is_not_reported_as_unchanged` (reproduced the exact `(unchanged)` bug) |
| M2 | remove the no-op note | `test_identical_frames_are_declared_as_a_no_op` |
| M3 | fire the note unconditionally | `test_changed_frames_carry_no_no_op_note`, `test_frameless_turns_never_claim_an_unchanged_screen` |
| M4 | drop reply capture on the exception | `test_a_rejected_reply_is_carried_on_the_exception_for_the_bundle`, `test_an_over_long_rejected_reply_is_bounded_on_the_exception` |
| M5 | publish the diagnostic without the reply | `test_a_rejected_decision_publishes_the_reply_beside_the_diagnostic` |
| M6 | unquote the schema field names | `test_system_prompt_states_the_schema_the_protocol_enforces` |
| M7 | remove the keyboard guidance | `test_prompt_teaches_how_to_type_and_how_to_press_keys` |
| M8 | revert the coordinate convention | `test_prompt_states_the_coordinate_convention` |
| M9 | remove the derived key vocabulary | `test_prompt_lists_the_key_vocabulary_the_validator_enforces` |

### Gates (whole tree, exactly as CI)

```
.venv/bin/python -m flake8 .                              -> clean
uvx --from black==26.1.0 black --check .                  -> 830 files unchanged
uvx isort==5.13.2 --check .                               -> clean
.venv/bin/python -m pyright --pythonpath .venv/bin/python . -> 0 errors, 0 warnings
  -> 11507 passed, 15 skipped in 522.68s
```

## Changes

- `local_operator/evaluation/runner/provider_client.py` — absent-vs-unchanged text; `_frames_identical` no-op signalling; quoted schema field names; `_named_keys_line()` derived vocabulary; coordinate convention; keyboard and batching guidance; `reply` carried on `DecisionRejected`.
- `local_operator/evaluation/runner/model.py` — `DecisionRejected.reply`, documenting why the diagnostic alone is insufficient.
- `local_operator/evaluation/runner/episode.py` — `_rejection_detail()` publishes reply beside diagnostic.
- `local_operator/evaluation/protocol.py` — `_NAMED_KEYS` → `NAMED_KEYS` so the prompt can derive from the enforcing set.
- Tests: 11 new cases in `test_provider_client.py` and `test_episode.py`; `conftest.py` fixture carries a reply.
- Version bumped to `0.44.58` (patch — reliability and prompt correctness, no new surface). Checked at push time against `main` (`0.44.56`), PyPI (`0.44.56`), and open PRs: #583 holds `0.44.56`, #581 holds `0.44.57`, #576 holds `0.45.0`.

## Honest judgement on this route's competitiveness

**The evidence does not support "the model is too weak" — it supports "the harness lied to it."** The model reached for the keyboard and was refused on a field-name technicality; it was told the screen was frozen on 15 of 17 turns while frames changed by up to 93%. Under the fixed prompt the same model produces a parser-accepted `key`+`type`+`key` batch on a real screenshot from the failed run.

What is **not** proven here: that this yields task *completion*. This PR proves step-level correctness, not trajectory-level. The task needs ~20+ correct steps (read an email, extract several defenses, resolve calendar conflicts, create entries), and a competitive OSWorld 2.0 score is a much higher bar than "makes progress". My expectation is the score moves off ~0; I would re-run this single task before committing to the full batch rather than assume it. The fatal `rpcremoteerror` that ended the episode is a separate adapter defect being fixed elsewhere and is untouched here.

## Not addressed (deliberately out of scope)

- The `rpcremoteerror` adapter RPC failure that terminated the episode — being fixed separately.
- Two of the three rejections were `Extra data: line 1 column 351` (valid JSON followed by trailing content). The prompt already says "a single JSON object and nothing else, with no prose and no code fence"; whether a stricter extraction (parsing the first complete JSON value rather than rejecting the whole reply) is the right call is a real design question, but it trades strictness for leniency at the protocol boundary and deserves its own decision rather than being folded into this diagnosis.

### DIFF
--- a/local_operator/evaluation/protocol.py
+++ b/local_operator/evaluation/protocol.py
@@ -413,7 +413,13 @@ class TypeAction(_Action):
     text: str = Field(min_length=1, max_length=MAX_TEXT_LENGTH, pattern=r"\S")
 
 
-_NAMED_KEYS = frozenset(
+#: The named keys ``KeyAction`` accepts, case-insensitively. Public because
+#: the episode system prompt must state this vocabulary rather than imply it:
+#: a model told only that "key" takes named keys guesses plausible synonyms
+#: the validator does not accept ("control", "escape", "return"), and each
+#: guess is a rejected reply that costs a billed corrective re-prompt. Derived
+#: from the enforcing set so the prompt cannot drift from what parses.
+NAMED_KEYS = frozenset(
     {
         "ALT",
         "BACKSPACE",
@@ -456,7 +462,7 @@ class KeyAction(_Action):
         normalized: list[str] = []
         for key in keys:
             candidate = key.upper() if len(key) > 1 else key
-            if candidate not in _NAMED_KEYS and candidate not in _PRINTABLE_KEYS:
+            if candidate not in NAMED_KEYS and candidate not in _PRINTABLE_KEYS:
                 raise ValueError(f"unknown key: {key!r}")
             normalized.append(candidate)
         if len(normalized) != len(set(normalized)):
--- a/local_operator/evaluation/runner/episode.py
+++ b/local_operator/evaluation/runner/episode.py
@@ -815,7 +815,16 @@ class EpisodeRunner:
             # The diagnostic is what the model will be shown; it is published
             # as an artifact rather than squeezed into the identifier-shaped
             # ``diagnostic_code`` so the exact correction is auditable.
-            detail = self._publish(rejected.diagnostic.encode("utf-8"), media_type="text/plain")
+            #
+            # The REPLY is published alongside it, because the diagnostic
+            # alone does not say what the model was attempting. A Pydantic
+            # error names the fields it refused; it cannot tell a later reader
+            # whether the turn was a mistyped keyboard action or a stray
+            # sentence after the JSON. Without the reply, diagnosing a
+            # rejection class costs a whole re-run of a paid episode.
+            detail = self._publish(
+                _rejection_detail(rejected).encode("utf-8"), media_type="text/plain"
+            )
             self._append(
                 "error",
                 ErrorPayload(
@@ -1651,6 +1660,23 @@ def _incomplete_receipt(plan: CleanupPlan, action_id: str) -> CleanupReceipt:
         evidence_code="worker-unavailable",
         duration_ms=0,
     )
+
+
+def _rejection_detail(rejected: Any) -> str:
+    """The rejection artifact: why the reply was refused AND what it said.
+
+    Two sections rather than one blob, so a reader (or a script mining a batch
+    of bundles for rejection classes) can tell the harness's diagnostic apart
+    from the model's own words. A client that did not capture the reply --
+    every implementation of the protocol is free not to -- degrades to the
+    diagnostic alone rather than emitting an empty section that reads as "the
+    model said nothing".
+    """
+
+    reply = getattr(rejected, "reply", None)
+    if not reply:
+        return rejected.diagnostic
+    return f"{rejected.diagnostic}\n\n--- rejected reply ---\n{reply}"
 
 
 def _diagnostic(error: BaseException) -> str:
--- a/local_operator/evaluation/runner/model.py
+++ b/local_operator/evaluation/runner/model.py
@@ -130,6 +130,7 @@ class DecisionRejected(Exception):
         self,
         diagnostic: str,
         *,
+        reply: str | None = None,
         route: RouteIdentity | None = None,
         usage: ModelUsage | None = None,
         cost_micros: int = 0,
@@ -142,6 +143,15 @@ class DecisionRejected(Exception):
     ) -> None:
         super().__init__(diagnostic)
         self.diagnostic = diagnostic
+        # WHAT the model actually said, not just why it was refused. The
+        # diagnostic alone cannot answer the question a post-mortem asks --
+        # "was it trying to type?" -- because a Pydantic error names the
+        # fields it disliked and not the intent behind them. A real paid
+        # episode's three rejections were only reconstructible as far as
+        # "something with a `key` field" and "trailing junk after the JSON";
+        # the replies themselves were discarded, so the failure class could
+        # not be diagnosed without paying for the run again.
+        self.reply = reply
         # The served route matters even for a rejected reply: a fallback that
         # answered badly still moved the run off its pinned route.
         self.route = route
--- a/local_operator/evaluation/runner/provider_client.py
+++ b/local_operator/evaluation/runner/provider_client.py
@@ -26,13 +26,19 @@ from __future__ import annotations
 import base64
 import json
 import re
+import textwrap
 import time
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Mapping, Sequence, get_args, get_origin
 
 from local_operator.evaluation.adapters.supervisor import verify_artifact
 from local_operator.evaluation.evidence.models import RouteIdentity
-from local_operator.evaluation.protocol import ActionBatch, ComputerAction, Observation
+from local_operator.evaluation.protocol import (
+    NAMED_KEYS,
+    ActionBatch,
+    ComputerAction,
+    Observation,
+)
 from local_operator.evaluation.runner.model import (
     CompactionRecord,
     DecisionRejected,
@@ -70,9 +76,34 @@ MAX_REJECTED_REPLY_CHARS = 4_000
 #: changes the message being appended.
 UNCHANGED_OBSERVATION = "(unchanged)"
 
+#: Rendered when an observation carries NO text at all. Kept distinct from
+#: :data:`UNCHANGED_OBSERVATION` because the two mean opposite things and a
+#: model that conflates them loses its only textual progress signal: a
+#: benchmark whose adapter publishes the task once and then screenshots only
+#: (OSWorld) yields ``text=None`` on every step after the first, and folding
+#: those into "(unchanged)" told the model the SCREEN had not moved on every
+#: single turn of a real paid episode.
+NO_TEXTUAL_STATE = "(no textual state)"
+
+#: Appended to an observation whose frame bytes are byte-identical to the
+#: previous observation's. This is the only reliable "nothing happened" signal
+#: a screenshot-only benchmark has: without it a model that clicked a dead
+#: pixel sees a new observation id, a new image block, and no statement that
+#: the two images are the same, so re-deciding the same click is the rational
+#: reading of the context rather than a lapse.
+UNCHANGED_FRAMES_NOTE = (
+    "The screenshot is byte-identical to the previous observation's: your last "
+    "action changed nothing visible. Do not repeat it -- try a different "
+    "target, a different action kind, or wait for the surface to settle."
+)
+
 # The batch wire version this harness speaks; pinned here rather than taken
 # from a model reply.
 PROTOCOL_VERSION = "1.0"
+
+#: Function keys are collapsed to a range in the prompt rather than listed:
+#: F1-F24 is 24 of the vocabulary's 43 entries and the pattern is obvious.
+_FUNCTION_KEY = re.compile(r"F\d+")
 
 # Mirrors receipts.StrictIdentifier, which every evidence identifier must match.
 _IDENTIFIER = re.compile(r"[A-Za-z0-9][A-Za-z0-9_.:-]*")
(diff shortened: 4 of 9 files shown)


damianvtran:
### Agent review — round 1

Reviewer: reviewer subagent (local-operator team, claude-opus-5)
Scope: origin/main (8af528dc)..3f9ba7ca

**Verdict: CLEAN — TERMINAL round. 0 blockers, 0 majors. Recommend merge** (after the independent QA round on this head).

#### Verification by reproduction (not inference)

I stood up clean detached worktrees at head (`3f9ba7ca`) and at the merge base (`8af528dc`) and reproduced every load-bearing claim independently:

1. **Real-bundle replay through the real `_ContextBuilder`.** Rebuilt all 17 `EpisodeTurn`s from `ep-ffda3fc88f81`'s own observation/action artifacts (frames reconstructed at 1920x1080, matching the bundle's own `Frames: screen (1920x1080)` rejection artifacts):
   - at `origin/main`: **15 `(unchanged)`**, 1 `(no textual state)` — the false claim is reproduced;
   - at head: **0 `(unchanged)`**, 16 `(no textual state)`, 0 no-op notes — exactly as the PR claims.
2. **Independent frame-delta re-measurement** (Pillow `ImageChops` on the bundle's own artifacts, digests re-verified): all 16 consecutive pairs differ, 0.02%–93.30%, **zero byte-identical pairs** — matching the PR's table exactly. The 15 `(unchanged)` claims were false, and the new no-op note correctly stays silent on this episode.
3. **Schema as the model reads it.** Rendered `build_system_prompt()` at head: all field names are quoted JSON keys (`"keys": [str, ...]` beside `"kind"`), with an explicit `"keys" is not "key"` warning. Probed the real parser: `["ctrl","k"]`/`["shift","a"]`/`["enter"]`/`["esc"]`/`["tab"]`/`["alt","f10"]`/`["f24"]`/`["Ctrl","S"]` ACCEPTED; `["control","k"]`/`["return"]`/`["escape"]`/`["f25"]` REJECTED. The rendered vocabulary (20 named keys + "f1 through f24") exactly matches `NAMED_KEYS` minus function keys, and the three named synonyms are the exact spellings the parser refuses.
4. **Coordinate convention vs the parser.** Prompt states top-left origin, x right, y down, max valid x = width-1, screenshot IS that pixel space. `ActionBatch.validate_for` rejects at `x >= visible.width` / `y >= visible.height` (protocol.py:580-584) and `FrameSize` is documented top-left-origin — the stated convention is exactly what the validator enforces.
5. **Keyboard/batching claims vs the adapter.** Verified against the installed `lop_osworld_v2_adapter` (0.1.1): `TypeAction` → `pyautogui.typewrite` (no click, no Enter — "type follows focus" and "does not press Enter for you" are true); multi-key `KeyAction` → `pyautogui.hotkey` (chord — "TOGETHER" is true); `compile_batch` returns one statement per action in order with one observation at the end — "batches run in order" is true.
6. **No-op note semantics.** Fires only on digest equality of consecutive framed observations; frameless observations never fire it (`_frames_identical` returns False on empty/mismatched frame lists). **Mutation-tested both directions in my worktree** (restored afterwards): always-identical mutation is killed by `test_changed_frames_carry_no_no_op_note` + `test_frameless_turns_never_claim_an_unchanged_screen`; never-identical mutation is killed by `test_identical_frames_are_declared_as_a_no_op`.
7. **`DecisionRejected.reply`.** Bounded at `MAX_REJECTED_REPLY_CHARS` (4,000, same bound as context replay) before it rides the exception; optional — `_rejection_detail` degrades to the diagnostic alone for `reply=None`/`""` (tested). Redaction: publication goes through `EpisodeRunner._publish` → `publish_artifact`, which streams every artifact through the store's `_RedactionScanner` and re-asserts on the text projection — **the same scanner the evidence store uses for all artifacts, no second implementation**. A canary-bearing reply fails publication exactly as a canary-bearing accepted reply already does today (episode abandons for evidence integrity); this change does not widen the secret surface.
8. **Generalizability.** The model-facing prompt contains zero benchmark-specific strings — the only occurrence of "benchmark" is the generic phrase "normal for a screenshot-only benchmark". No task knowledge, no OSWorld special-casing, no per-benchmark strings in any rendered surface; OSWorld appears only in code comments as provenance. Reads as generic computer-use prompt hygiene.
9. **Honest limits.** PR body explicitly states "What is not proven here: that this yields task completion. This PR proves step-level correctness, not trajectory-level" and recommends re-running the single task before the batch. No overclaim. The "parser-accepted key+type+key batch" claim is consistent with my own parser probes (the old-prompt reply `["control","k"]` is rejected; the corrected spelling the new prompt teaches is accepted).
10. **Would a competent model now produce a parser-accepted type/key batch?** Yes — the rendered prompt states the field names as quoted keys, the full key vocabulary with the exact accepted spellings and the three rejected synonyms named, chord semantics, focus semantics, batch ordering, and the coordinate contract. Every element that cost the real episode a rejection is now stated in the exact form the parser accepts.

#### Gates (run in the head worktree, exactly as CI)

- flake8: clean · black 26.1.0: 832 files unchanged · isort 5.13.2: clean · pyright: 0 errors, 0 warnings
- `tests/unit/evaluation` (the touched subtree): **724 passed**
- CI on `3f9ba7ca`: all checks pass (unit shards, both tui-e2e legs, lint, pip-audit, cli/server-sanity, coverage)
- Version `0.44.58` strictly above main (`0.44.56`) and PyPI (`0.44.56`); no open PR holds `0.44.58` (#576's `0.45.0` is a different lane)

#### Findings

No BLOCKER. No MAJOR.

**NIT-1** — `provider_client.py:159` comment says "24 of the vocabulary's **43** entries"; `NAMED_KEYS` has **44** members (24 function + 20 other), at main as well as head. The PR body repeats "closed 43-entry set". No model-facing impact (the vocabulary is derived, so the rendered list is correct); fix the count in the comment on a future touch, or leave it.

**NIT-2** — The key-vocabulary prose says "a single printable character is also a key" but does not state that all letters/digits/punctuation are accepted key values (`_PRINTABLE_KEYS`). A model wanting `["a"]` has to generalize from one sentence. Low risk — `type` covers text entry — but one clause would close it.

**NIT-3** — The bundle artifact truncates the reply at `MAX_REJECTED_REPLY_CHARS` without the `[... reply truncated]` marker the context-replay path appends (`provider_client.py:497-498`), so a reader of the bundle cannot tell a truncated reply from a complete one. Cosmetic; the bound itself is correct and tested.

**NIT-4** — PR body's version note ("#583 holds 0.44.56, #581 holds 0.44.57") is stale: #584 now also carries `0.44.57`. No collision results; bookkeeping only.

#### Disposition

The four nits are follow-ups, not a reason for another round — this round is **TERMINAL**. The change is correct where it must be (every model-facing claim verified against the parser and the adapter), the failure it fixes is reproduced before/after on the real bundle, and the new tests are mutation-resistant on the load-bearing comparison. Clear to merge once QA's independent pass lands on this head.

PR #555, opened 2026-09-02
fix(browser): name tab groups after the session that opened them

damianvtran:
## What

Tab groups are named after the **session that opened them**. The operator's browser showed three groups all reading `LO · Session`, `LO · Session (2)`, `LO · Session (3)` — the ordinal de-duplication working correctly on top of an empty session name, so nothing identified which session owned which tab.

Runtime `0.44.44` → **`0.44.45`** (patch — a fix). Extension `0.1.5` → **`0.1.6`**.

## The defects (all three verified against the running code, not inferred)

**1. Latch race.** `ToolContext` is a **snapshot** built once per turn (`Session._build_tool_context`, session.py:4738), while a conversation names itself **asynchronously** a second or two into its first turn. A `browser open` in that opening turn — the common case, since "look at this page" is often the opening request — read `session_name == ""` even *after* the title had landed, and `surface.groupBaseLabel` persisted it for the tab's whole life.

**2. No rename propagation.** Nothing pushed a title that arrived *after* the surface existed. `reconcileCommandTab` re-titles on every non-`open` command, so a group only self-healed if the session happened to issue another browser command after naming — for `open → screenshot → close` it never did.

**3. Bare fallback.** `Session` for *every* unnamed session at once is precisely what the ordinal turns into `(2)`/`(3)`: three groups that name nothing.

## The fix

**Python (`local_operator/`)**

- `ToolContext.session_name_provider` — an optional callable that re-reads the live title at **tool-call time**, closing the snapshot race. `session_name` stays as the fallback for hosts that install no provider. Display-only; identity/authorization still use `session_id`.
- `_browser_session_label` falls back to the **cwd basename** for an unnamed session — the same substitution the TUI's status band and terminal title already make in this slot (`lo › <cwd>`, `tui/terminal_title.cwd_label`). The bare `Session` survives only at a filesystem root, where `cwd_label` itself declines to answer, because `LO · /` names a session no better.
- `retitle_browser_surface()` + `Session._push_browser_title()` push a late-arriving title. Fire-and-forget through the tracked task group: `set_conversation_name` runs on the TUI's synchronous paint path and tab chrome must never delay or fail a turn, a title, or a browse. Skipped entirely when the session has no bridge tab open, so a non-browsing session pays nothing — not even the import.

**Extension (`extension/src/`)**

- New `retitle` wire method (`protocol.py` → regenerated `protocol.gen.ts` → `worker.ts` handler → `tab-groups.ts`).
[...]

PR #569, opened 2026-09-03
feat(providers): add Radient OAuth 2.0 PKCE sign-in and rolling refresh

damianvtran:
## Summary
- Adds interactive browser-based OAuth 2.0 PKCE authentication flow for Radient via `console.radienthq.com/oauth/authorize` and `api.radienthq.com/v1/auth/oauth/token`.
- Preserves full backward compatibility with `RADIENT_API_KEY` as a credential row alias via `radient-oauth` provider flavour (`store_credentials_as='radient'`).
- Supports automatic rolling 90-day token refresh to maintain indefinite active sign-in sessions.

## Testing Evidence
- Unit tests added in `tests/unit/providers/test_radient_oauth.py`:
  - Provider registration and attributes check.
  - PKCE authorization URL generation with S256 code challenge and parameters.
  - Token exchange with mocked token endpoint.
  - Rolling refresh token update verification.
- Ran full test suite in isolated worktree with all 69 provider and oauth flow tests passing.

PR #563, opened 2026-09-03
fix(browser-bridge): keep the bridge usable when the heartbeat writer dies

damianvtran:
## Incident

An agent fleet (session `2d3373a1c698`) lost the browser extension mid-session. Every `browser` call silently degraded to the cmux backend — `scroll`/`logs`/`tabs` returned "not supported on the cmux backend" — while `lop browser status` simultaneously reported:

```
extension connected: yes
paired:              yes
driving:             http://127.0.0.1:8974/r3-notice-amber.svg
```

…a URL whose server was already dead. The user closed **every** tab in Chrome; status still claimed it was driving that URL. Both the agent and the user concluded the bridge was "bound" by an orphan tab and abandoned the extension for the rest of the session.

Nothing was bound. Four independent defects combined into one unrecoverable, undiagnosable state, and the machine was still sitting in it (daemon pid 1646) when this work started.

## Live reproduction, before any change

The broken daemon was investigated before restarting anything:

```
$ cat ~/.local-operator/run/browser/bridge.json     # heartbeat_at 1788386598
$ python3 -c 'import time; print(time.time())'      # 1788398281  -> 11,683s (3.2h) stale
$ ps -o pid,etime -p 1646                           # 1646  05-02:58:37   ALIVE
$ curl -s localhost:4099/health                     # HTTP 200, extension_connected: true
$ curl -s http://127.0.0.1:8974/r3-notice-amber.svg # 000 — connection refused (server dead)
$ grep -n 'Errno 28' ~/Library/Logs/local-operator/browser-bridge.log
534:OSError: [Errno 28] No space left on device: '.../.bridge.fsfgktwb.tmp'
```

The traceback in that log ends in `_heartbeat` → `publish` → `state.publish` → `tempfile.mkstemp`. The heartbeat died at 18:03; the tool failures began at 21:01 — **the heartbeat death precedes the symptom by three hours.**

### RC1 — a supervisory loop died silently and disabled the bridge for every session (primary)

`BridgeService._heartbeat` was `while True: self.publish(); await sleep(15)`. When the disk filled, ENOSPC escaped the loop, the task **died**, and nothing restarted or noticed it. `state.available()` requires `now - heartbeat_at <= 45s`, so with no writer it returned False **forever** → `bridge_browser_available()` False → every new `open` silently routed to cmux, while `lop browser status` (which reads the live `/health` socket, not the file) kept saying "connected: yes". That contradiction *is* the incident.

[...]

PR #583, opened 2026-09-03
fix(browser): name subagent tab groups after the work they do

damianvtran:
## Root cause

**Correction to the brief's hypothesis first.** Candidate 1 (the tool-construction
context being captured and winning over the execute-time one) is **disproven**:
`execute_browser` is a module-level function with no closure over the construction
context, and the per-turn context built by `Session._build_tool_context` is the one
that reaches every tool call. The execute-time context does win.

What actually happened has two parts:

1. **The operator's screenshot predates the fix that already shipped.** The incident
   tab was opened at `2026-09-02 21:10:23` (transcript ts `1788397823`, session
   `2d3373a1c698`, title *"Fix Slack-reported UI zoom and overlap bugs"*). The fix for
   exactly that symptom — `session_name_provider` + the cwd fallback — merged in
   #555 (`1568599e`) at `2026-09-03 01:11:47`, **four hours later**. The operator's
   session process was running the pre-fix code, which read only the per-turn
   `session_name` snapshot and returned the bare `Session` fallback. That cause is
   already fixed on main; the screenshot is not evidence of a residual defect.

2. **The residual live defect: subagents.** A delegated child has no conversation
   title and can never grow one — naming runs in the TUI host and the owned-session
   runtime, and a one-shot child passes through neither (verified on this machine:
   no subagent session directory holds a `title.json`; 718 of 778 browsing sessions
   here are subagents, all untitled). Its context carried no `session_name` and its
   cwd was its **parent's**, so every child of one parent derived the identical cwd
   label and a fleet of them rendered as one repeated pill distinguished only by an
   ordinal that names nothing.

## The fix

- **New `ToolContext.job_label` field** — the short name a subagent was delegated
  under (`zoom-scroll-fix`, `bridge-qa`). Plumbed from `harness/subagent.py` through
  `Session.__init__` into the per-turn `_build_tool_context`, so the execute-time
  context carries it.
- **The pill composes as `<parent title> › <job label>`** — the separator matching
  the TUI's own `lo › <cwd>` convention. The parent half is read through the shared
  `ConversationName` holder, so a parent named or `/rename`d **after** launch still
  reaches the child's next command. When the composition does not fit the 30-cluster
  pill, the **parent** is what gets clipped — the label is the distinguishing half
  (siblings share a parent by definition).
- **Bang mode** built its own `ToolContext` outside any turn and omitted the title
  entirely; it now carries it, with a provider so a title landing mid-command is
  still current.
- **Security boundary intact**: the label is display-only; identity stays on
  `session_id`/`requester`. A child borrowing its parent's name for a pill is never
  a child acting with its parent's identity. Sanitisation runs on both sides.

## Evidence (real execution)

**Wire capture — `session_label` on outgoing RPCs, main vs branch** (intercepted at
`_bridge_call`, the single choke point every browser RPC passes through):

| path | main | branch |
|---|---|---|
| named foreground session | `Debug browser extension port…` | `Debug browser extension port…` |
| subagent A (`tabgroup-naming`) | `local-operator` | `Debug brows… › tabgroup-naming` |
| subagent B (`bridge-qa`) | `local-operator` | `Debug browser… › bridge-qa` |
| subagent C (`designer`) | `local-operator` | `Debug browser… › designer` |
| child, parent not yet named | `proj` | `proj › qa` |
| same child after parent renamed | `Fix the tab group naming` | `Fix the tab group naming › qa` |
| retitle RPC pushed by session | `Fix the tab group naming` | `Fix the tab group naming › qa` |

On main, **all three subagents send the identical label** — the exact collision the
ordinal was papering over. `requester` (identity) is unchanged in every case.

**Sibling disambiguation** — a fleet of six children off the operator's actual
session (`ab90e52f53e6`):

```
child job label    MAIN pill                    BRANCH pill
tabgroup-naming    LO · local-operator          LO · Debug brows… › tabgroup-naming
bridge-qa          LO · local-operator          LO · Debug browser… › bridge-qa
reviewer           LO · local-operator          LO · Debug browser… › reviewer
qa-tester          LO · local-operator          LO · Debug browser… › qa-tester
designer           LO · local-operator          LO · Debug browser… › designer
ux-reviewer        LO · local-operator          LO · Debug browser… › ux-reviewer

distinct pills on MAIN  : 1 of 6   <-- all six collide
distinct pills on BRANCH: 6 of 6
```

**Live real-extension capture** — drove the paired extension (0.1.7) against the
running daemon with `example.com` (operator-approved). Opened a tab as this subagent
session (running the installed 0.44.54 = main), which sent `session_label='tmp'`
(the residual defect reproduced live: a child with no title falls to its cwd
basename), then issued a follow-up `goto` which re-ran `reconcileCommandTab` against
the live group — confirming the rename-refresh path works end to end. The group pill
itself is browser chrome outside the page viewport, so it cannot be screenshotted;
the label is verified programmatically via the outgoing `session_label` and the
extension's reconcile behaviour.

**Regression tests** — 9 new tests, all fail on main and pass on this branch:
- `test_tool_selection.py`: subagent naming, sibling non-collision, `job_id`
  discriminator, unnamed-parent fallback, invisible-label fallback, rename
  propagation, long-parent clipping, label-fills-pill case.
- `test_launch_subagent.py`: end-to-end wiring through a real parent `Session` and
  a real `_build_child_session` child, including a rename performed after the child
  was built.
- `test_app_pilot.py`: bang mode carries the title and a live provider.
- `test_tool_context_parity.py`: registered `job_label` in the sentinel table (the
  existing derived guard caught the new field and forced the registration).

## Gates

- `flake8`, `black --check` (26.1.0), `isort --check-only` (5.13.2), `pyright` — all
  clean over the whole tree.
- Full unit suite: **11506 passed, 15 skipped** (one pre-existing load-flake in
  `tests/unit/model/test_prices.py::test_a_background_revalidation_gets_the_full_default_timeout`
  — a background-thread spawn race; both files byte-identical to main, passes 5/5 in
  isolation; not touched by this change).

## Store constraint

**`extension/` is untouched by this change** — zero files under `extension/` in the
diff. The fix is entirely Python-side. The manifest's `permissions` and
`host_permissions` are unchanged, so the 0.1.5 → 0.1.7 version-line-only property
holds for any build produced from this branch.

## Versioning

PATCH bump to **0.44.56** (verified free against `git tag`, `gh release list`, and
the PyPI JSON API immediately before committing). This is a fix to existing
behaviour, not a new surface.

### DIFF
--- a/local_operator/harness/subagent.py
+++ b/local_operator/harness/subagent.py
@@ -1243,6 +1243,13 @@ async def _build_child_session(
         session_id=transcript.directory.name,
         agent_id=parent_session.agent_id,
         job_id=job_id,
+        # The child's own name for display surfaces that must not render a
+        # fleet of children identically (browser tab groups today). Set on the
+        # CONSTRUCTION context for the same defensive-parity reason
+        # ``variables`` is: what actually reaches an executing tool is the
+        # child ``Session``'s per-turn rebuild, which receives it via the
+        # ``job_label=`` argument below.
+        job_label=label,
         has_ui=parent_session._has_ui,
         request_approval=request_approval,
         # The parent's variable store. DEFENSIVE PARITY, not a bug fix: no
@@ -1364,6 +1371,22 @@ async def _build_child_session(
         # ``Session._build_tool_context``; the construction-time context above
         # only feeds createIf.
         job_id=job_id,
+        # The label the operator launched this child under, and the PARENT's
+        # live title holder. Together they are the only identity a subagent
+        # has: naming runs in the TUI host and the owned-session runtime, so a
+        # one-shot child never generates a title of its own and every display
+        # surface asking "which session is this?" had nothing to answer with.
+        #
+        # The holder is SHARED, not copied, and that is what makes a parent
+        # renamed after launch (the usual case — the naming errand lands a
+        # second or two into the parent's first turn, while children are
+        # launched later) reach the child's next tab-group reconcile. Passing
+        # the string would freeze whatever the title was at launch.
+        #
+        # Display-only on both counts: a child is authorized by its own
+        # ``session_id``, never by a name it borrowed from its parent.
+        job_label=label,
+        parent_conversation_name=parent_session.conversation_name_state,
         # The PARENT's comms instance, so the child's every-turn tool context
         # rebuild keeps pointing at the agent that delegated to it instead of
         # minting a private one nobody is listening to.
--- a/local_operator/harness/subagent.py
+++ b/local_operator/harness/subagent.py
@@ -107,6 +107,7 @@ import asyncio
 import contextlib
 import logging
 import uuid
+import weakref
 from dataclasses import dataclass
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Awaitable, Callable
@@ -1087,6 +1088,42 @@ def _child_mcp_wiring(parent_session: "Session", *, restricted: bool = False) ->
     )
 
 
+def _parent_display_name_resolver(parent_session: "Session") -> Callable[[], str]:
+    """A callable returning the parent's DISPLAY name at the moment it is asked.
+
+    The child's browser tab group reads ``<parent conversation> › <job label>``,
+    and both halves have to survive nesting. Handing the child the parent's
+    title HOLDER only worked one level down: a middle child never generates a
+    title of its own (naming runs in the TUI host and the owned-session
+    runtime, neither of which a one-shot child passes through), so its holder
+    is permanently empty and a grandchild fell back to the cwd that every
+    sibling of every conversation shares — two ``qa`` grandchildren under two
+    different conversations rendered identically. Delegation really does nest:
+    a child of a top-level session keeps ``task``/``wait``/``jobs`` (see the
+    depth-aware prune below), which is exactly the manager-fans-out-to-workers
+    shape. Resolving through ``_display_session_name`` instead walks the
+    lineage to whichever ancestor actually holds a title.
+
+    Called per read rather than snapshotted, because a parent is normally named
+    a second or two into its first turn while its children are launched later:
+    a string captured here would be "" for the child's whole life.
+
+    WEAK reference on purpose. Every other parent-derived value the child gets
+    is a shared collaborator (the comms surface, the variable store, the job
+    manager's parent row); this one would be a strong child→parent edge that
+    pins the parent's entire object graph — transcript, tools, MCP manager —
+    for as long as a detached child outlives it. A dead parent simply has no
+    name to lend, and the caller degrades to the cwd form it already handles.
+    """
+    parent_ref = weakref.ref(parent_session)
+
+    def resolve() -> str:
+        parent = parent_ref()
+        return parent._display_session_name() if parent is not None else ""
+
+    return resolve
+
+
 async def _build_child_session(
     *,
     label: str,
@@ -1371,22 +1408,17 @@ async def _build_child_session(
         # ``Session._build_tool_context``; the construction-time context above
         # only feeds createIf.
         job_id=job_id,
-        # The label the operator launched this child under, and the PARENT's
-        # live title holder. Together they are the only identity a subagent
-        # has: naming runs in the TUI host and the owned-session runtime, so a
-        # one-shot child never generates a title of its own and every display
-        # surface asking "which session is this?" had nothing to answer with.
-        #
-        # The holder is SHARED, not copied, and that is what makes a parent
-        # renamed after launch (the usual case — the naming errand lands a
-        # second or two into the parent's first turn, while children are
-        # launched later) reach the child's next tab-group reconcile. Passing
-        # the string would freeze whatever the title was at launch.
+        # The label the operator launched this child under, and a resolver for
+        # the PARENT's display name. Together they are the only identity a
+        # subagent has: naming runs in the TUI host and the owned-session
+        # runtime, so a one-shot child never generates a title of its own and
+        # every display surface asking "which session is this?" had nothing to
+        # answer with.
         #
         # Display-only on both counts: a child is authorized by its own
         # ``session_id``, never by a name it borrowed from its parent.
         job_label=label,
-        parent_conversation_name=parent_session.conversation_name_state,
+        parent_display_name=_parent_display_name_resolver(parent_session),
         # The PARENT's comms instance, so the child's every-turn tool context
         # rebuild keeps pointing at the agent that delegated to it instead of
         # minting a private one nobody is listening to.
--- a/local_operator/harness/types.py
+++ b/local_operator/harness/types.py
@@ -718,6 +718,29 @@ class ToolContext(BaseModel):
     # Optional and display-only for the same reason ``session_name`` is:
     # identity and authorization continue to use ``session_id``.
     session_name_provider: Callable[[], str] | None = None
+    # The DELEGATED-WORK label, set only on a subagent's context: the short
+    # name its parent launched it under (``zoom-scroll-fix``, ``bridge-qa``).
+    #
+    # A subagent has no conversation title and can never grow one — title
+    # generation lives in the TUI host and the owned-session runtime, and a
+    # one-shot child runs through neither, so every subagent session directory
+    # on disk has no ``title.json``. That is by design (a child answers one
+    # prompt and exits; naming it would cost a provider call for a title
+    # nobody resumes), but it left every display surface that asks "which
+    # session is this?" with nothing to say — browser tab groups above all,
+    # where a fleet of children rendered as a wall of identical pills.
+    #
+    # The label is the identity the OPERATOR already recognises: it is what
+    # they typed into ``task``, what the jobs list shows, and what they address
+    # with ``hub``. Display-only and never an identity, exactly like
+    # ``session_name`` — a child is authorized by ``session_id``/``requester``,
+    # and two children may legitimately carry the same label.
+    #
+    # Named after ``job_id`` above rather than after the browser's wire field:
+    # this is the human name of the same background job that field identifies,
+    # and calling it ``session_label`` would collide with the DERIVED display
+    # value the browser bridge already sends under that exact key.
+    job_label: str = ""
     # Resolver hook for lazy internal URLs (``skill://`` and ``guide://``);
     # returns content or None when the URL is not handled. Installed by session.
     resolve_internal_url: Callable[[str], str | None] | None = None
--- a/local_operator/session/session.py
+++ b/local_operator/session/session.py
@@ -1382,12 +1382,18 @@ class Session:
         #: whole of its display identity. Reaches tools via
         #: ``_build_tool_context``; display-only, never an authorization input.
         job_label: str = "",
-        #: The PARENT's live title holder, on a subagent only. Shared rather
-        #: than copied so a parent named or renamed AFTER this child launched
-        #: still reaches the child's display surfaces — the common case, since
-        #: the parent's naming errand lands early in its first turn while
-        #: children are launched later. ``None`` on every top-level session.
-        parent_conversation_name: ConversationName | None = None,
+        #: The parent's own :meth:`_display_session_name`, on a subagent only.
+        #: A RESOLVER rather than the parent's title holder, for two reasons
+        #: that are really one: it re-reads on every call, so a parent named or
+        #: renamed AFTER this child launched still reaches the child's display
+        #: surfaces (the common case — the naming errand lands early in the
+        #: parent's first turn while children are launched later); and it
+        #: resolves TRANSITIVELY, so a grandchild under a middle child whose
+        #: own holder is permanently empty still reads the top-level
+        #: conversation's title instead of falling through to the shared cwd.
+        #: A holder cannot do the second: the middle child has no title of its
+        #: own to hold. ``None`` on every top-level session.
+        parent_display_name: Callable[[], str] | None = None,
         #: The parent↔child messaging surface (``harness.comms.SubagentComms``).
         #: A top-level session mints its own; a CHILD is handed its parent's,
         #: which is what makes ``hub`` inside a subagent talk to the agent that
@@ -1445,7 +1451,7 @@ class Session:
         self._variables = variables
         self._job_id = job_id
         self._job_label = job_label
-        self._parent_conversation_name = parent_conversation_name
+        self._parent_display_name = parent_display_name
         self._subagent_comms = subagent_comms
         self.agent_registry = agent_registry
         self.team_registry = team_registry
@@ -3215,28 +3221,41 @@ class Session:
         live title instead, so a delegated tab group reads
         ``<parent conversation> › <job label>`` rather than a bare fallback.
 
-        Read through the shared holder on every call, never cached: the parent
-        is normally named a second or two into its first turn while its
-        children are launched later, so a value snapshotted at the child's
-        construction would be the empty string for the child's whole life.
-        That staleness is the exact failure this method exists to close, and it
-        is why :meth:`_build_tool_context` passes this as the PROVIDER as well
-        as the snapshot.
+        Resolved on every call, never cached: the parent is normally named a
+        second or two into its first turn while its children are launched
+        later, so a value snapshotted at the child's construction would be the
+        empty string for the child's whole life. That staleness is the exact
+        failure this method exists to close, and it is why
+        :meth:`_build_tool_context` passes this as the PROVIDER as well as the
+        snapshot.
+
+        RECURSIVE by construction, because delegation nests: a child of a
+        top-level session keeps ``task``/``wait``/``jobs`` (see
+        ``harness.subagent``), so a manager fanning out to workers is depth 2
+        and the operator's usual shape. The middle child has no title of its
+        own and can never grow one, so asking it for its *holder* yields ""
+        forever and a grandchild fell back to the cwd every sibling of every
+        conversation shares — two ``qa`` grandchildren under two different
+        conversations rendered identically, the very collision this method
+        exists to prevent. Asking the parent for its resolved DISPLAY name
+        instead walks the chain to whichever ancestor actually holds a title.
+        The walk is bounded by the lineage depth built at construction (no
+        cycle is constructible: a parent always predates its child).
 
         Display only. Identity and authorization stay on ``session_id`` — a
         child borrowing its parent's name for a tab pill must never be read as
         a child acting with its parent's identity.
 
         Deliberately does NOT reach for a parent's title on an unnamed FORK:
-        a fork is not a child, holds no ``_parent_conversation_name``, and
+        a fork is not a child, holds no ``_parent_display_name``, and
         ``_load_conversation_name`` declines the inheritance on purpose (see
         :attr:`wears_inherited_title`). Its display label stays the cwd
         substitution every unnamed session gets.
         """
         own = self._conversation_name.text
-        if own or self._parent_conversation_name is None:
+        if own or self._parent_display_name is None:
             return own
-        return self._parent_conversation_name.text
+        return self._parent_display_name()
 
     @property
     def wears_inherited_title(self) -> bool:
--- a/local_operator/session/session.py
+++ b/local_operator/session/session.py
@@ -1378,7 +1378,7 @@ class Session:
         #: subagent (``zoom-scroll-fix``, ``bridge-qa``). A child never
         #: generates a conversation title — naming runs in the TUI host and the
         #: owned-session runtime, neither of which a one-shot child passes
-        #: through — so this plus ``parent_conversation_name`` below is the
+        #: through — so this plus ``parent_display_name`` below is the
         #: whole of its display identity. Reaches tools via
         #: ``_build_tool_context``; display-only, never an authorization input.
         job_label: str = "",
--- a/local_operator/session/session.py
+++ b/local_operator/session/session.py
@@ -1374,6 +1374,20 @@ class Session:
         #: rebuilt every turn, so anything set only on the construction-time
         #: context reaches the createIf check and never the executor.
         job_id: str | None = None,
+        #: The short name this session was DELEGATED under, set only on a
+        #: subagent (``zoom-scroll-fix``, ``bridge-qa``). A child never
+        #: generates a conversation title — naming runs in the TUI host and the
+        #: owned-session runtime, neither of which a one-shot child passes
+        #: through — so this plus ``parent_conversation_name`` below is the
+        #: whole of its display identity. Reaches tools via
+        #: ``_build_tool_context``; display-only, never an authorization input.
+        job_label: str = "",
+        #: The PARENT's live title holder, on a subagent only. Shared rather
+        #: than copied so a parent named or renamed AFTER this child launched
+        #: still reaches the child's display surfaces — the common case, since
+        #: the parent's naming errand lands early in its first turn while
+        #: children are launched later. ``None`` on every top-level session.
+        parent_conversation_name: ConversationName | None = None,
         #: The parent↔child messaging surface (``harness.comms.SubagentComms``).
         #: A top-level session mints its own; a CHILD is handed its parent's,
         #: which is what makes ``hub`` inside a subagent talk to the agent that
@@ -1430,6 +1444,8 @@ class Session:
         self._goal_state = goal_state if goal_state is not None else GoalState()
         self._variables = variables
         self._job_id = job_id
+        self._job_label = job_label
+        self._parent_conversation_name = parent_conversation_name
         self._subagent_comms = subagent_comms
         self.agent_registry = agent_registry
         self.team_registry = team_registry
@@ -3188,6 +3204,39 @@ class Session:
     def conversation_name(self) -> str:
         """The conversation's title ("" until one is set or generated)."""
         return self._conversation_name.text
+
+    def _display_session_name(self) -> str:
+        """The conversation title DISPLAY surfaces should show for this session.
+
+        Identical to :attr:`conversation_name` for a top-level session. On a
+        SUBAGENT — which has no title of its own and can never generate one,
+        since naming runs in the TUI host and the owned-session runtime and a
+        one-shot child passes through neither — it resolves to the PARENT's
+        live title instead, so a delegated tab group reads
+        ``<parent conversation> › <job label>`` rather than a bare fallback.
+
+        Read through the shared holder on every call, never cached: the parent
+        is normally named a second or two into its first turn while its
+        children are launched later, so a value snapshotted at the child's
+        construction would be the empty string for the child's whole life.
+        That staleness is the exact failure this method exists to close, and it
+        is why :meth:`_build_tool_context` passes this as the PROVIDER as well
+        as the snapshot.
+
+        Display only. Identity and authorization stay on ``session_id`` — a
+        child borrowing its parent's name for a tab pill must never be read as
+        a child acting with its parent's identity.
+
+        Deliberately does NOT reach for a parent's title on an unnamed FORK:
+        a fork is not a child, holds no ``_parent_conversation_name``, and
+        ``_load_conversation_name`` declines the inheritance on purpose (see
+        :attr:`wears_inherited_title`). Its display label stays the cwd
+        substitution every unnamed session gets.
+        """
+        own = self._conversation_name.text
+        if own or self._parent_conversation_name is None:
+            return own
+        return self._parent_conversation_name.text
 
     @property
     def wears_inherited_title(self) -> bool:
@@ -5094,15 +5143,22 @@ class Session:
             session_id=self._session_id,
             # Re-read the live holder every turn so generated, user-set, and
             # resumed titles reach display-only browser metadata after renames.
-            session_name=self.conversation_name,
+            # On a CHILD this resolves to the parent's title (see
+            # :meth:`_display_session_name`), which is the only conversation
+            # name a subagent has.
+            session_name=self._display_session_name(),
             # Per-turn is not enough on its own: this context is a SNAPSHOT
             # taken once at turn start, and the naming errand lands a second or
             # two into the FIRST turn — so a browse in that turn read "" even
             # after the title existed, and the tab group latched the fallback
             # label for the life of the tab. The callable re-reads the holder at
             # tool-call time. Display-only, like ``session_name`` itself.
-            session_name_provider=lambda: self.conversation_name,
+            session_name_provider=self._display_session_name,
             agent_id=self._agent_id,
+            # The delegated name, on a subagent only. Empty on every top-level
+            # session, which is what keeps ``_browser_subagent_label``'s
+            # ``job_id`` discriminator honest.
+            job_label=self._job_label,
             has_ui=self._has_ui,
             resolve_internal_url=self._skill_resolver,
             request_approval=None if self._yolo else self._request_approval,
--- a/local_operator/tools/builtin.py
+++ b/local_operator/tools/builtin.py
@@ -6155,6 +6155,16 @@ _BROWSER_SESSION_LABEL_CLUSTERS = 30
 #: root, where ``cwd_label`` itself declines to answer.
 _BROWSER_FALLBACK_LABEL = "Session"
 
+#: Separator between a subagent's parent conversation and its own job label.
+#: U+203A, matching the TUI's own ``lo › <cwd>`` composition rather than
+#: introducing a second convention for the same "A, narrowed to B" relation.
+_BROWSER_SUBAGENT_SEPARATOR = " › "
+
+#: Below this many clusters a clipped parent title identifies no conversation
+#: (``Fi…`` names nothing), so the child's label stands alone instead of
+#: wearing a stub prefix. See :func:`_browser_subagent_label`.
+_BROWSER_SUBAGENT_PARENT_MIN = 8
+
 
 def _browser_live_session_name(context: ToolContext | None) -> str:
     """The session's title as of NOW, not as of the turn's context snapshot.
@@ -6208,13 +6218,31 @@ def _browser_session_label(context: ToolContext | None) -> str:
     intentionally broader than the known bidi and zero-width set: browser tab
     chrome is too small to make invisible direction changes attributable.
 
-    An UNNAMED session falls back to its working directory's basename, which is
-    the same substitution the TUI's status band and terminal title already make
-    in this exact slot (``tui/terminal_title.cwd_label`` — ``lo › <cwd>``). The
-    slot has always held "the best label we have", and a directory the user
-    chose the session for distinguishes three concurrent groups where three
-    copies of ``Session`` do not.
+    A SUBAGENT is named first, before any title lookup, because it is the one
+    session kind that can never acquire a title: naming lives in the TUI host
+    and the owned-session runtime, and a one-shot child runs through neither
+    (confirmed on this machine — no subagent session directory holds a
+    ``title.json``). Its context carries no ``session_name`` and its cwd is its
+    PARENT's, so every child of one parent derived the identical cwd label and
+    a fleet of them rendered as ``LO · local-operator (2)``/``(3)``/… — the
+    ordinal is the only thing that differed, and it names nothing. See
+    :func:`_browser_subagent_label` for the form and why the parent's name is
+    carried alongside the child's.
+
+    An UNNAMED top-level session falls back to its working directory's
+    basename, which is the same substitution the TUI's status band and terminal
+    title already make in this exact slot (``tui/terminal_title.cwd_label`` —
+    ``lo › <cwd>``). The slot has always held "the best label we have", and a
+    directory the user chose the session for distinguishes three concurrent
+    groups where three copies of ``Session`` do not.
     """
+    # Subagent first: a child's own ``session_name`` is empty by construction,
+    # but the cwd fallback below would still answer (with the parent's
+    # directory) and thereby hide the far more specific label the parent
+    # launched this child under. Order is the whole fix.
+    subagent = _browser_clean_label(_browser_subagent_label(context))
+    if subagent:
+        return _browser_clip_label(subagent)
     # Emptiness is decided by the SANITISED text, not by ``str.strip()``.
     # ``strip()`` removes whitespace but NOT the Cf/Cc classes, so a title of
     # nothing but zero-width or bidi characters (U+200B, U+FEFF, U+2060,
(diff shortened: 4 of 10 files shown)


damianvtran:
### Agent review — round 1

Reviewer: `reviewer` subagent (lopdev team), model `anthropic/claude-opus-5`. Independent of the implementing coder — I did not write, edit, or push any of this code, and I reviewed by executing the branch rather than by reading the author's evidence.
Scope: `4d3494ef34510cd64602e39e64acadafb10b5b8e..bd3b2fdd9cd822ab2128a6b295ec93cad2f861df` (true merge-base of `origin/main` and head; 10 files, +504/-12).

Verified by running, in a scratch tree outside the repo: the label composition against 12 hand-built edge contexts, a real 6-child fleet and a real grandchild built through `subagent._build_child_session`, the 9 new tests replayed against a **pre-fix source tree** (base source + branch tests only), a 407-case fuzz of the Python output through the extension's actual `cleanLabel` under Node, an interleaved 12-round A/B of the claimed flake, and the four CI gates.

**What holds.** The root-cause correction in the PR body is right. `execute_browser` is a module-level function taking `context` per call (`builtin.py:7023`), the only context that reaches it is `context.tool_context` assigned per turn at `session.py:4845` from `_build_tool_context`, and the construction-time context in `subagent.py:1241` feeds only `createIf` — candidate 1 is disproven as claimed, and I found no path by which a nameless construction context reaches the executor. Depth-1 plumbing is correct for all four required cases. A real 6-child fleet off one named parent produced **6 distinct pills of 6** (main: 1 of 6). The rename-through-the-shared-holder path works on a *later* command (`Fix tab groups › sub` after a post-launch `set_conversation_name`). Clipping takes the parent and never the label; every composed pill I generated landed at ≤30 clusters. Fuzzing 407 Python-produced labels through the extension's real `cleanLabel` gave **0 alterations** — the two sides do not disagree about the pill. The unnamed-parent case degrades to `local-operator › qa`, not `Session › qa`, and the fork path is genuinely untouched (a fork holds no `_parent_conversation_name`, so `_display_session_name` returns `""` and the cwd substitution stands). Identity is clean: `requester` stays `session:<session_id>` in every case I generated, `job_label` has zero authorization readers repo-wide, and sanitisation strips Cc/Cf, bidi and zero-width on both sides before the emptiness test.

The 9 new tests are **not tautologies** — replayed against the pre-fix source they fail 9/9 with meaningful assertions (`assert 'Tab groups' == 'Tab groups › zoom-scroll-fix'`, `'Fix the tab groups' != 'Fix the tab groups'` for the sibling collision, `AttributeError: 'ToolContext' object has no attribute 'job_label'`, and `assert '' == 'Debug browse...binding issue'` for bang mode). 126/126 pass on the branch. flake8, black 26.1.0, isort 5.13.2 and pyright are all clean over the whole tree.

The flake claim also holds, and understates itself: `tests/unit/model/test_prices.py` and `local_operator/model/` are byte-identical to base (0 files in the diff), and an interleaved A/B on this host (load avg 51) failed **1/12 on the branch against 4/12 on the pre-fix tree** — the branch flaked *less*. Load flake, not a defect.

---

**F1 — BLOCKER — the version bump collides with an already-published release.**
`pyproject.toml` head sets `version = "0.44.56"`, but `v0.44.56` is already tagged (`8af528dc`), already has a GitHub Release ("0.44.56: Radient OAuth PKCE and rolling refresh", 2026-09-03T14:01:12Z) and is **already on PyPI** (`local_operator-0.44.56-py3-none-any.whl`, uploaded 14:02:05). The branch was cut at `4d3494ef` (`v0.44.55`) and #569 landed on main afterwards, so the PR is 1 commit behind and the bump was correct when written and is stale now. Merging as-is lands main at a version equal to the released one: `git merge origin/main bd3b2fdd` resolves cleanly (no conflict in `pyproject.toml`, I checked) and leaves `version = "0.44.56"`, so nothing will flag it — then `gh release create v0.44.56` fails on the existing tag and PyPI rejects the duplicate filename. Fix: rebase onto current `origin/main` and bump to **0.44.57**. PATCH remains the right class per AGENTS.md — this is a fix to existing behaviour, no new surface.

**F2 — MAJOR — the fix stops at depth 1; grandchildren lose the conversation half and siblings collide again.**
`_display_session_name` (`session.py:3205`) reads `self._parent_conversation_name.text` one level up only. A grandchild is handed its *child* parent's holder, which is permanently `""` (a child never generates a title — the same fact this PR is built on), so the composition falls through to the cwd branch. Depth 2 is not hypothetical: `subagent.py:1428` deliberately supports it ("A child of a TOP-LEVEL session keeps task/wait/jobs"), and I confirmed a child really does retain `task` in `child._tools`. Reproduced with real sessions — two children `feature-a`/`feature-b` off one named top-level session, each spawning a `qa` grandchild:
```
grandchild 'qa' under child feature-a  -> 'local-operator › qa'
grandchild 'qa' under child feature-b  -> 'local-operator › qa'
distinct grandchild pills: 1 of 2
```
Not a regression (main collided here too), but the PR's central claim — a fleet of children gets distinct pills — does not hold one level down, and a manager child fanning out is exactly the operator's 2-6-child pattern. One-line fix: make the inherited half resolve transitively — pass the *parent's own resolved display name* down (or fall back to `parent_session._parent_conversation_name` when the parent's own holder is empty) so a grandchild composes as `<top-level title> › qa` rather than `<shared cwd> › qa`.

**F3 — MINOR — repurposing `session_name` bleeds the parent's title into the `send` tool's sender identity.**
`builtin.py:4862` uses `context.session_name` as the peer-message sender's `conversation_name` when no registry record names the process. On a subagent that field now resolves to the **parent's** title, so on a reduced host a peer message sent by a child would present under its parent's conversation name while `session_id` correctly stays the child's — I confirmed the child's context carries `session_name='Parent conversation title'` with its own distinct `session_id`, and children do get the `send` tool. Narrow (fallback path only) and display-only, so no identity consequence; worth a `job_label`-aware sender string or a comment recording that this consumer now sees an inherited name.

**F4 — MINOR — the `job_id`-is-the-discriminator docstring is inaccurate.**
`_browser_subagent_label` states `job_id` "is what the harness sets on exactly the child contexts and on nothing else", but server-side sessions also carry one (`server/utils/operator.py:713`, `job_processor_queue.py:105`, `routes/chat.py:387`). The behaviour is fine — such a context has `job_label=""`, hits the `return label or parent` branch, and yields its own title unchanged (verified) — but the stated invariant is what a future reader would rely on, and it is not true.

**F5 — NIT — an 8-cluster `_BROWSER_SUBAGENT_PARENT_MIN` silently drops the parent for mid-length labels.** A 22-cluster label leaves `room=5`, so the pill renders as a bare label with no conversation half (`'zzzz…'` with no `›`) — correct per the documented rule and the right trade-off, but the threshold means the two halves are present or absent with no visible signal about which happened. Worth a sentence in the docstring giving the label length at which the parent disappears.

---

**Verdict: CHANGES REQUESTED — not clean.** F1 blocks the merge outright: the release step cannot succeed at `0.44.56`, and it is the kind of failure that surfaces only after the squash lands. F2 is a real gap in the feature's own claim and should be fixed here rather than deferred, since the transitive resolution is a small change to the method this PR just introduced. F3–F5 are answerable in the same round. The engineering underneath is careful and the evidence in the PR body checked out everywhere I tested it — the parent-not-label clipping choice and the cross-side cluster parity are both genuinely right, and the flake claim is honest. Round 2 should be scoped to the remediation diff plus a rebase onto current `origin/main`.

damianvtran:
### QA report — round 1

**QA:** qa-tester (independent of the implementer) · anthropic/claude-opus-5
**Scope:** `4d3494ef..bd3b2fdd` (`git merge-base origin/main HEAD`..head)
**Head tested:** `bd3b2fdd9cd822ab2128a6b295ec93cad2f861df` (current head at time of writing)
**Verdict: FAIL** — 1 major (Q1), 1 blocker (Q3, release mechanics), 1 nit (Q2). The change's core claim is **proven correct and effective**; the blocker is a version collision, not a defect in the fix.

**Environment.** Fresh detached worktree `/tmp/qa-tabgroup` at the PR head, A/B against a second worktree at `origin/main` (`8af528dc`). Every harness ran under a per-run `LOCAL_OPERATOR_CONFIG_DIR` tempdir; `~/.local-operator` was never written and no live session was stopped, attached to, or restarted. Live stack: daemon `:4099` healthy, extension **0.1.7** connected and paired, `example.com` operator-approved. I ran the real code, not the diff: seven QA-authored harnesses driving real `Session` objects, the real `_build_child_session`, the production `Session._launch_subagent`, the real `OperatorApp` Textual pilot, and the real paired extension over the bridge.

Confirmed the Python-only claim independently: `git rev-parse bd3b2fdd:extension` == `origin/main:extension` == `4d3494ef:extension` == `6f9ce808`. Zero files under `extension/`.

---

## What works — verified by execution, not by reading

**The operator's actual complaint is fixed.** A fleet of six siblings off one named parent, built through the real child-session builder:

| job label | main | this branch |
|---|---|---|
| `tabgroup-naming` | `local-operator` | `Fix browser… › tabgroup-naming` |
| `tabgroup-qa` | `local-operator` | `Fix browser… › tabgroup-qa` |
| `reviewer` | `local-operator` | `Fix browser tab… › reviewer` |
| `designer` | `local-operator` | `Fix browser tab… › designer` |
| `ux-reviewer` | `local-operator` | `Fix browser… › ux-reviewer` |
| `architect` | `local-operator` | `Fix browser tab… › architect` |

**distinct pills: 1 of 6 on main → 6 of 6 on this branch**, and every pill contains its own job label. Reproduced independently through the **production** `_launch_subagent` (5 real jobs, all `status=completed`): 5 of 5 distinct. `requester` stays per-child in both trees — the borrowed name never becomes a borrowed identity.

**The pill end to end against the REAL extension.** `retitle` returns the label the extension actually applied, so the pill is observable without eyes on the tab strip. Eight cells, all matching `'LO · ' + <wire label>` exactly:

```
named session      -> applied 'LO · Debug browser extension port…'
subagent           -> applied 'LO · Debug browser… › tabgroup-qa'
sibling subagent   -> applied 'LO · Debug brows… › tabgroup-naming'
unnamed parent     -> applied 'LO · local-operator › bridge-qa'
after parent named -> applied 'LO · Fix the tab… › bridge-qa'
overlong pair      -> applied 'LO · Fix Slack-r… › zoom-scroll-fix'
bidi in label      -> applied 'LO · Parent Title › qagniman'
```

**Rename propagation is real**, including the case the PR body leans on: a context snapshot taken *before* the rename still produces the new label at tool-call time (`'Renamed after the… › sub'`) — the provider re-reads the live holder, it is not a per-turn freeze.

**Bang mode**, driven through the real `OperatorApp` pilot — main vs branch:

```
main   : snapshot=''  wire='qa-tabgroup-main'  has_provider=False
branch : snapshot='Debug browser extension port binding issue'
         wire='Debug browser extension port…'  has_provider=True
branch : title landing MID-command -> wire becomes 'Named While The Bang Ran'
```

**Fork suppression holds** (Q4): an unnamed fork gets `'oh-my-pi'` (its cwd), never a parent's title — it holds no parent holder by construction.

**Sanitisation and boundaries** (14 cells, all PASS): bidi `U+202E/U+202C` stripped; a zero-width-only label falls back to the parent rather than to `Session`; newline/tab collapse to spaces; empty and whitespace-only labels degrade sensibly; combining marks count as one cluster (40 codepoints = 20 clusters, kept whole); parent clips while the label survives; a parent squeezed below the 8-cluster minimum drops out entirely rather than showing a `Fi…` stub. A top-level session carrying a stray `job_label` with `job_id=None` does **not** compose — the discriminator is honest.

**Regression sweep, all through the real extension:** `open`/`goto`/`read`/`snapshot`/`click`/`type`/`screenshot`/`scroll`/`logs`/`tabs`/`close` all PASS. `type` verified by its page-side side effect (`typed: qa-tabgroup-round-1`), screenshot rendered and viewed, stale-snapshot ref and unapproved-origin click both produce the correct refusals, and the origin-approval gate refused `qa-not-approved-origin.example.org` without navigating.

**Mixed-version interop:** both trees put exactly `['requester', 'session_label']` on the wire. No new wire field, so a new host against an older extension is byte-compatible — the composition is just a longer string in a field that already existed.

**Gates:** flake8 clean · black `830 files unchanged` · isort clean · pyright `0 errors, 0 warnings` · full unit suite **11508 passed, 15 skipped** (the known `test_prices.py` load-flake passed this run, no isolation needed) · extension `81 pass, 0 fail`. I also copied the new `test_tool_selection.py` into a main worktree: **6 failed, 62 passed** — the new tests genuinely bite.

---

## Findings

### Q1 — MAJOR: the parent half collapses at nesting depth 2, re-colliding children of different conversations

The composition works for a direct child. For a **grandchild** the parent half degrades to the shared cwd, so two children with the same label under two *different* conversations render identically again — the exact class of collision this PR exists to fix.

```
depth 1:
  conv A > qa            'Fix browser tab group… › qa'
  conv B > qa            'Add Radient OAuth PKCE… › qa'      distinct ✓

depth 2:
  conv A > manager > qa  'local-operator › qa'
  conv B > manager > qa  'local-operator › qa'               COLLIDE ✗
```

**Cause.** `_build_child_session` passes `parent_conversation_name=parent_session.conversation_name_state` — the parent's *raw holder*. For a middle child that holder is permanently empty (a subagent never generates a title), so the grandchild inherits emptiness rather than the middle child's *resolved* display name. The distinction is visible directly:

```
middle._display_session_name()        = 'Fix browser tab group naming'   <- resolves correctly
middle.conversation_name_state.text   = ''                                <- what the grandchild gets
```

`_display_session_name` already knows the answer; it is the raw holder, not the resolver, that is handed down. A late rename of the grandparent also fails to reach depth 2 (`'local-operator › qa'` before and after), while the depth-1 control updates correctly.

**Reachability — this is not theoretical.** The `manager` role on this machine carries `delegate:yes` (`~/.local-operator/agents/deb80c27…/agent.yml`), and `harness/subagent.py` only prunes delegation for scouts, non-delegating roles, and *grandchildren*. A manager spawning workers is depth 2, which is precisely how the `lopdev` team that produced this PR is shaped. Sibling workers under one manager still separate from each other (`'local-operator › reviewer'` vs `'… › designer'`), so the damage is bounded to telling two concurrent conversations apart — but that is the half the PR body argues the parent title exists for.

Repro: `/tmp/qa-tabgroup-evidence/repro_Q1.py`.

### Q2 — NIT: the composed pill can be 31 clusters against a documented 30-cluster cap

```
parent='Fix Slack-reported UI zoom and overlap bugs' label=<54 chars>
  -> 'a-really-very-long-subagent-jo…'   31 clusters
```

`_browser_clip_label` treats `budget` as clusters *kept* and then appends the ellipsis. The code comments this deliberately as parity with the extension's `slice(0, MAX) + "…"`, and I confirmed **the identical off-by-one exists on `main`** on the plain-title path — so this is pre-existing and inherited, not introduced here. Live check (cell 3.7) shows the extension re-clips independently and the rendered pill stays bounded. Reporting it because the new `_browser_subagent_label` docstring states a hard ceiling it does not quite hold; no user-visible overflow. Repro: `repro_Q2.py`.

### Q3 — BLOCKER (release mechanics, not the fix): version `0.44.56` is already taken

`pyproject.toml` on this head declares `0.44.56`. That version has already shipped:

```
origin/main pyproject      : 0.44.56
git tag v0.44.56           -> 8af528dc  (Radient OAuth PKCE, PR #569)
gh release v0.44.56        : "0.44.56: Radient OAuth PKCE and rolling refresh"
PyPI 0.44.56               : present, 2 files, uploaded 2026-09-03T14:02:05Z
```

The merge base (`4d3494ef`) was `0.44.55`, so the bump was correct when written and `origin/main` moved underneath it. The branch is still `MERGEABLE` — pyproject does not conflict textually — so a merge would produce a `main` whose version equals an already-published PyPI release, and `publish.yml` would fail (or worse, silently no-op) on the release step. Needs a rebase and a bump to `0.44.57` before merge. **PATCH is the right bump type** — this is a fix to existing behaviour, no new surface.

---

## What needs the operator's eyes

The tab-group pill is browser chrome outside the page viewport and **cannot be screenshotted**. I verified the label at the wire (`session_label` on the outgoing RPC) and at the extension's own applied-title readback, which is the strongest programmatic evidence available. Unverified by QA: how the pill *renders* — colour, where the real font truncates, and whether `›` reads clearly at pill size next to the `LO ·` prefix.

## Gaps

- **Surface cap** (6.17) — BLOCKED. Exercising it means opening many real tabs in the operator's live browser; I declined rather than disrupt the working session. Untouched by this diff.
- **`tests/e2e`** not run — no e2e cell covers browser tab-group labelling, and the diff does not touch the boot/resume path the stage guards.
- The live-extension cells drove **one** tab handle sequentially. Two genuinely concurrent sessions writing different pills to two different real groups at the same time was not exercised; the extension's `serialize()` queue is the relevant guard and its own suite covers it (81/81).

Full matrix, one row per cell with actual outputs: `/tmp/qa-tabgroup-evidence/QA-MATRIX-round1.md`.

**Verdict: FAIL.** Q3 must be resolved before merge; Q1 is a real gap in the fix's stated goal at a depth this team actually uses. Q2 is a nit and reasonable to defer. Nothing in the diff regressed anything I could reach — the direct-child case, which is the reported complaint, is solidly fixed.

damianvtran:

Implementer: `coder` subagent (lopdev team), author of this branch.
New head: `d55840e9a1d5eb1a17cba2bc02efc460a982fca1` — rebased onto current `origin/main` (`0dc28cf2`), CI **green on this head** (all 13 checks pass). Remediation delta vs the round-1 head's content: 6 files, +364/−52.

| Finding | Severity | Disposition |
|---|---|---|
| **F1 / Q3** — version `0.44.56` already released | BLOCKER | **Fixed** (`d55840e9`). Rebased onto `origin/main` (which had moved to `0.44.58`, released) and bumped to **`0.44.59`** as the LAST commit. Verified free at bump time against all three sources: no `v0.44.59` tag, no GitHub Release, PyPI JSON API 404. Note: `0.44.57` was skipped on PyPI but is claimed by open PR #585, so it was not free either. |
| **F2 / Q1** — parent half collapses at depth 2 | MAJOR | **Fixed** (`5fc80ffe`). `_build_child_session` now hands the child a resolver for the parent's **display name** (`_parent_display_name_resolver`, weakref-held) instead of the parent's raw title holder; `_display_session_name` calls it, so resolution walks the lineage to whichever ancestor holds a title — transitively, at every depth. Verified by execution: two `qa` grandchildren under two **different** conversations now render `Fix browser tab group… › qa` vs `Add Radient OAuth PKCE… › qa` (was `local-operator › qa` for both); a real 2-manager × 2-worker fleet gives 4 distinct depth-2 pills; the operator's actual pattern (one manager, 6 workers) gives **6 of 6 distinct**; a late grandparent rename reaches depth 2 on the next command; resolution stays live to depth 4+. Regression tests added for depth 2, cross-conversation distinctness, late rename at depth 2, and resolver weakness — all **fail against the pre-fix source** (10/10 of the new/updated assertions), depth-1 and top-level behaviour re-verified unchanged. |
| **F3** — `send` fallback signs with the parent's title | MINOR | **Fixed** (`5fc80ffe`). New `_peer_sender_conversation_name`: a child's fallback sender name composes its own label (`<parent title> › <label>`, same form as the pill, without the pill's 30-cluster budget); top-level and server sessions (`job_id` set, no label) keep their own title. Covered by a new test. |
| **F4** — `job_id`-exclusivity docstring is false | MINOR | **Fixed** (`5fc80ffe`). The docstring now states that `job_id` is also carried by server-side sessions and that the composition discriminates on `job_id` **and** `job_label` together (only `_build_child_session` sets the label). Added a test pinning that a server-shaped context keeps its own title. |
| **F5 / Q2** — 31 clusters vs documented 30; dropout threshold undocumented | NIT | **Fixed** (`5fc80ffe`). Corrected the off-by-one at its source rather than weakening the docstring: `_browser_clip_label`'s budget is now inclusive of the ellipsis it appends, so a clipped label never exceeds 30 clusters on ANY path (plain title, cwd, composed). The composed form's hand-rolled `budget - 1` is gone. Two pre-existing tests that encoded the old off-by-one were updated. F5: the docstring now states the dropout point — with a 30-cluster pill, 3-cluster separator and 8-cluster parent minimum, the parent half disappears once the child's label reaches 20 clusters (verified at 19/20). |

**Deferred — same-conversation, same-label grandchildren.** Two `qa` grandchildren under two different managers of ONE conversation still share a pill (`<top title> › qa`). This is pre-existing on `main` (which collides them at the cwd too) and is inherent to both findings' proposed resolution — resolving to the top-level title cannot separate children of the same conversation; doing so would require the intermediate manager labels in the pill (`<title> › manager-a › qa`), a pill-format and budget redesign neither finding requested. Recording it here rather than filing an issue.

**Extension untouched.** Zero files under `extension/` in this PR; the tree at head (`6f9ce808`) is byte-identical to `origin/main`'s. (For the record: relative to the published 0.1.5 tree, `main`'s extension already carries the 0.1.6/0.1.7 functional changes from #555 and #563 — this PR adds nothing to that.)

**Gates on the new head, whole tree as CI:** flake8 clean · black 26.1.0 `832 files unchanged` · isort 5.13.2 clean · pyright `0 errors, 0 warnings` · full unit suite **11532 passed, 15 skipped** (the `test_prices.py` load-flake passed in-run, no isolation needed).

damianvtran:
### Agent review — round 2

Reviewer: `reviewer` subagent (lopdev team), model `anthropic/claude-opus-5`. Independent of the implementing coder — I did not write, edit, or push any of this code, and I reviewed by executing the branch in a scratch worktree rather than by reading the author's remediation comment.
Scope: `0dc28cf22bc78251eea5cbd412c66c9c6dbc379f..d55840e9a1d5eb1a17cba2bc02efc460a982fca1` (true merge-base of `origin/main` and head at review time; remediation delta `12c6a10d..d55840e9`, 6 files, +364/-52).

Round 2 is scoped to the remediation diff, as the round-1 files were approved unchanged. I confirmed the rebase carried no source drift: `git diff bd3b2fdd 12c6a10d` touches **none** of this PR's 10 files (only #569's Radient OAuth files), so round 1's verdict on the original commit still stands on its own terms.

**All five round-1 findings and all three QA findings are fixed.** One is fixed and has since been *re-broken by the outside world* — F1 recurred while this review ran.

---

## Verification by execution

**F2/Q1 — the transitive depth fix. Resolved, and the mechanism is right.** `_build_child_session` now passes `_parent_display_name_resolver(parent_session)` (`subagent.py:1421`) and `_display_session_name` calls it (`session.py:3258`), so resolution recurses through `_display_session_name` at each hop rather than reading one holder. Built a real 5-deep lineage through `_build_child_session` off one named top-level session:

```
depth 0: display='Fix browser tab group naming'  pill='Fix browser tab group naming'
depth 1..5: display='Fix browser tab group naming'  pill='Fix browser tab group… › sub'
```

Resolution is genuinely transitive to depth 5, not just depth 2. A real 2-conversation × 2-manager × 3-worker fleet (12 depth-2 children) gives **cross-conversation distinctness of 2/2 for every label** (`Fix browser tab group… › qa` vs `Add Radient OAuth PKCE… › qa`) — the exact collision Q1 reported is gone. Worth noting the practical ceiling: `task`/`wait` are pruned at depth 2 (I confirmed depth-2 children hold `jobs`/`hub` but not `task`), so depth 2 is the deepest lineage the harness can actually construct, and the fix covers it with margin.

**Late renames propagate through the extra hops.** At depth 3: `local-operator › sub` → (rename grandparent) → `Named Much Later › sub` → (rename again) → `Renamed A Second Time › sub`. Live on every read, no per-turn freeze.

**The weakref is genuinely weak and adds no cycle or leak.** The closure holds exactly one cell, a `ReferenceType`. Held alone, it does not retain the parent: after `dispose()` + `del` + `gc.collect()` the parent is collected and `resolve()` returns `""`. A **dead parent degrades, it does not raise** — `_display_session_name()` returns the cwd form. I also traced what *does* retain a parent while a live child exists: the 8 referrers are the child's shared `SubagentComms`, a coroutine and bound methods — **no weakref among them**, so this PR added no retention edge. That matches the test's own honest comment about `SubagentComms` owning that edge rather than this code.

**No constructible cycle.** Forcing `p._parent_display_name = _parent_display_name_resolver(p)` by hand yields `RecursionError`, but `_build_child_session` is the only caller in the tree (`grep` confirms: 1 production call site) and a parent always predates its child, so the walk is bounded by lineage depth as documented.

**Identity boundary is clean.** The resolver closes over exactly one `weakref.ref` to the intended parent; it cannot name another session. A child of A reads A's title, does not read B's title, and renaming an unrelated session B leaves the child-of-A pill untouched. The value returned is an immutable `str`, so a child cannot mutate its parent's title through it. `session_id` remains the child's own throughout.

**F3 — `send` fallback sender. Resolved.** `_peer_sender_conversation_name` (`builtin.py:4783`) discriminates on `job_id` **and** `job_label` together, matching the browser side:

```
top-level named                    -> 'My Conversation'         (unchanged)
top-level unnamed                  -> ''                        (unchanged)
server session (job_id, no label)  -> 'Server Conv'             (unchanged)
server session, whitespace label   -> 'Server Conv'             (unchanged)
stray label, no job_id             -> 'Top'                     (unchanged)
subagent named parent              -> 'Parent Title › qa'       (no longer bleeds)
subagent UNNAMED parent            -> 'qa'
REAL child through _build_child_session -> 'Fix browser tab group naming › sub'
```

Top-level and server shapes are byte-identical to before; only the subagent case changed, which is the finding. The deliberately unbounded length is safe: the receiving side runs `_sanitize_sender_field` (`tui/widgets/transcript.py`), which collapses whitespace, strips Cc/Cf and bounds to `_SENDER_FIELD_MAX_CHARS`, so an over-long composed name cannot break the inbound block's pinned height.

**F5/Q2 — the ellipsis budget. Resolved; property verified, not spot-checked.** Fuzzed `_browser_clip_label` over 71 inputs × 41 budgets (2,911 cells): **zero** results exceed their budget at any reachable budget. (The only `>budget` cells are budgets 0–1 returning the 7-cluster `"Session"` fallback, which no caller can request — callers pass 30, or `room >= _BROWSER_SUBAGENT_PARENT_MIN` = 8.) Then the full pipeline over **4,952 composed cases** including a 41×36 parent/label boundary sweep and grapheme-hostile inputs (ZWJ families, combining marks, CJK, bidi): **0 exceed 30 clusters**. The documented 20-cluster dropout point is exact — 19 clusters keeps `Fix the… › …`, 20 drops the parent.

**Extension agreement re-fuzzed.** Ran all 2,509 unique Python-produced wire labels through the extension's real `cleanLabel` under Node: **0 altered, 0 over 30 clusters**. The Python side now never produces anything the extension needs to re-clip.

**F4 — docstring. Resolved.** Now states `job_id` is also carried by server sessions and that the discriminator is `job_id` + `job_label` together. Matches the code.

**Regression tests genuinely bite — twice over.** Against the **base source** (`0dc28cf2` + branch tests only): **17 failed**, with meaningful assertions. More usefully, against the **round-1 source** (`12c6a10d` + round-2 tests), which is the test that matters for a remediation round: **10 failed**, including all four new depth/weakref tests (`assert 'local-operator › sub' == 'Fix browser tab group… › sub'`, `assert 'local-operator › sub' != 'local-operator › sub'`, late-rename, and the missing resolver). The remediation tests fail against exactly the code they were written to fix.

**Deferred item recorded, not dropped.** The same-conversation same-label grandchild collision is stated in the remediation comment with its reasoning. I reproduced it (6 distinct pills of 12, collapsing across managers within one conversation) and agree it is pre-existing on `main`, inherent to the chosen resolution, and out of scope. No issue filed, per standing rules.

**Extension untouched, permissions unchanged.** `git diff 0dc28cf2..d55840e9 -- extension/` is **0 lines**; the `extension/` tree object at head (`6f9ce808`) is byte-identical to `origin/main`'s. `manifest.json` at head and on main both read version `0.1.7`, `permissions` `[debugger, tabs, tabGroups, scripting, storage, alarms, webNavigation, notifications]`, `host_permissions` `[<all_urls>]` — no delta, no new optional permissions.

**Gates on head, whole tree as CI:** flake8 clean · black 26.1.0 `832 files unchanged` · isort 5.13.2 clean · pyright `0 errors, 0 warnings`. Changed-code suites: `test_launch_subagent.py` + `browser_bridge/` + `test_tool_context_parity.py` = **205 passed**; the four TUI pilot label tests pass. I ran the targeted suites rather than the full matrix, per remediation-round scope; CI is green 13/13 on this head.

---

## Findings

### F6 — BLOCKER — 0.44.59 was published while this review ran; the bump is stale again

This is F1 recurring, not a regression in the fix. At the time I checked (16:03Z):

```
git tag -l v0.44.59          -> v0.44.59            (exists)
gh release list              -> 0.44.59: "Coerce job targets and resolve labels
                                in wait and jobs"  v0.44.59  2026-09-03T16:01:45Z
PyPI 0.44.59 present         -> True   (PyPI latest is now 0.44.59)
origin/main                  -> 73fe1226  (#586), pyproject version = 0.44.59
```

`pyproject.toml` at head still reads `version = "0.44.59"`, which is now the *published* version. Merging as-is repeats the exact failure mode F1 described: `git merge-tree` shows **0 conflict markers** against the new `origin/main`, so nothing will flag the collision, and then `gh release create v0.44.59` fails on the existing tag and PyPI rejects the duplicate filename.

**The next free version is 0.44.60** — no tag, `gh release view v0.44.60` reports "release not found", and PyPI does not list it. PATCH remains correct per AGENTS.md "Versioning": this is a fix to existing display behaviour with no new user-reachable surface, so it cannot be named in a minor's release title. Fix: rebase onto `73fe1226` and bump to **0.44.60**, and re-check immediately before `gh release create` — this branch has now been overtaken six times today, so treat the bump as a merge-time step rather than a commit-time one.

### F7 — MINOR — a stale `parent_conversation_name` reference survives in the `job_label` docstring

`session.py:1381` still reads "so this plus ``parent_conversation_name`` below is the whole of its display identity", but the parameter directly below it was renamed to `parent_display_name` in this very diff (`session.py:1396`). It is the only surviving mention of the old name in the tree (`grep -rn parent_conversation_name local_operator/ tests/` returns exactly this one line). Harmless at runtime, but it points a reader at a parameter that no longer exists, in the docstring whose whole job is explaining the pair. One-line fix: change that phrase to `parent_display_name`.

---

## Round-1 findings, disposition

| Finding | Severity | Round-2 disposition |
| --- | --- | --- |
| **F1** — version collides with a published release | BLOCKER | **Fixed as written, then re-broken externally.** 0.44.58 → 0.44.59 was correct when committed; 0.44.59 published at 16:01:45Z. Re-raised as **F6** against 0.44.60. |
| **F2 / Q1** — parent half collapses at depth 2 | MAJOR | **Resolved.** Verified transitive to depth 5 on real sessions; cross-conversation distinctness 2/2 per label at depth 2; late rename propagates at depth 3; weakref adds no retention edge and a dead parent degrades without raising; no constructible cycle; resolver reads only its own parent. |
| **F3** — `send` fallback signs with the parent's title | MINOR | **Resolved.** Subagent composes its own label; top-level and server shapes byte-identical; unbounded length is bounded by the receiver's `_sanitize_sender_field`. |
| **F4** — `job_id`-exclusivity docstring is false | MINOR | **Resolved.** Docstring now matches the two-field discriminator the code implements. |
| **F5 / Q2** — 31 clusters vs documented 30 | NIT | **Resolved.** 2,911 clip cells + 4,952 composed cells: zero over budget at any reachable budget; extension re-fuzz 0/2,509 altered; documented 20-cluster dropout is exact. |
| **Deferred** — same-conversation same-label grandchildren | — | **Recorded in-thread, verified pre-existing and inherent.** Correctly not filed as an issue. |

---

## Verdict

**Not clean — one BLOCKER (F6), and it is entirely mechanical.** The engineering under review is sound: every round-1 and QA finding is genuinely fixed, verified by execution rather than by reading the remediation comment, and the depth fix is both more general than the finding asked for and correctly bounded. F7 is a one-word docstring nit.

F6 is not a defect in this change — it is the version being overtaken on `main` for the sixth time today. Bump to **0.44.60** on a rebase onto `73fe1226`, fix F7 in the same commit, and this round is done; the substance needs no further review, so a round 3 should be a confirmation of those two lines and nothing more. Re-verify the free version immediately before `gh release create`, since this is now a recurring race rather than a one-off.

damianvtran:
### QA report — round 2

**QA:** `qa-tester` subagent (lopdev team), model `anthropic/claude-opus-5` — independent of the implementing coder. I did not write, edit, or push any of this code, and I verified by executing the branch, not by reading the diff or the remediation comment.
**Head tested:** `d55840e9a1d5eb1a17cba2bc02efc460a982fca1`
**Scope:** the remediation delta (`12c6a10d..5fc80ffe` + the `d55840e9` bump) plus the round-1 cells the changed functions could regress. Not a from-scratch re-audit of previously approved unchanged files.
**Verdict: FAIL** — **Q1 and Q2 are RESOLVED and I proved it by running them. Q3 has RECURRED**: `0.44.59` was released to PyPI 6 minutes after the remediation comment verified it free. Plus one new nit (Q4). **Nothing in the fix itself fails.**

**Environment.** Fresh detached worktree `/tmp/qa583-r2` at the PR head; control worktree `/tmp/qa583-main` at `origin/main` (`0dc28cf2`). Every harness under its own `LOCAL_OPERATOR_CONFIG_DIR` tempdir — `~/.local-operator` was never written, and no live session was stopped, attached to, or restarted. Live stack: paired extension **0.1.7**, `example.com` approved, one QA-owned tab reused and closed. Confirmed the Python-only claim independently: `git rev-parse d55840e9:extension` == `origin/main:extension` == `6f9ce808`, **0 files** under `extension/`.

---

## Q1 — RESOLVED (major, round 1). Verified by execution at depths 1–4.

The round-1 repro (`repro_Q1.py`), run verbatim against the new head:

```
depth 1:  A > qa            'Fix browser tab group… › qa'
          B > qa            'Add Radient OAuth PKCE… › qa'
depth 2:  A > manager > qa  'Fix browser tab group… › qa'
          B > manager > qa  'Add Radient OAuth PKCE… › qa'
          COLLIDE: False          <- round 1: both 'local-operator › qa'
```

Extended to the depths the finding did not reach, through real `Session` objects and the production `_build_child_session`:

| depth | conversation A | conversation B | distinct |
|---|---|---|---|
| 1 | `Fix browser tab… › manager` | `Add Radient OAuth… › manager` | ✓ |
| 2 | `Fix browser tab… › lead` | `Add Radient OAuth… › lead` | ✓ |
| 3 | `Fix browser tab group… › qa` | `Add Radient OAuth PKCE… › qa` | ✓ |
| 4 | `Fix browser tab group… › qa` | `Add Radient OAuth PKCE… › qa` | ✓ |

**The operator's actual pattern, at depth 2**: one manager fanning out the six lopdev roles gives **6 distinct pills of 6**, each naming its own label (`… › coder`, `… › reviewer`, `… › qa-tester`, `… › designer`, `… › architect`, `… › ux-reviewer`). Reproduced through the **production** `Session._launch_subagent` as well, not just the builder: 6 real jobs at depth 1 all `status=completed` with 6 distinct pills and 6 distinct `requester` values, then a real depth-2 fan-out (`probe-a/b/c`) giving 3 distinct pills that still carry the grandparent's conversation.

Also confirmed: a **late rename of the top-level** reaches depth 3 and depth 4 on the next command (`local-operator › worker` → `Grandparent Named… › worker`); a middle child that *does* acquire a title correctly wins over the grandparent (`Middle Took A Name › worker`); a fully unnamed lineage at depth 3 degrades to the cwd form `local-operator › w`, never to `Session`.

**The weakref claims hold**, and I tested the part the finding did not ask for:

- The resolver **alone** does not pin a disposed parent: parent collected, `resolve()` → `''`.
- The child holds no direct strong edge — child attributes pointing at the parent `Session`: `[]`; the closure's only cell is a `ReferenceType`.
- **Dead-parent degradation is graceful**: pill `'local-operator › orphan'`, `_display_session_name()` `''`, **no exception, 0.07 ms** — no raise, no hang.
- A depth-20 chain resolves in **0.035 ms** with no recursion error.
- A live child *does* keep a disposed parent alive — but **that is pre-existing, not this branch**: the referrer set is byte-identical on `origin/main` (`SubagentComms`, a cell, and four bound job/wake callbacks) and the child holds no attribute pointing at the `Session` on either tree. A bare disposed `Session` with no child *is* collected, which isolates the cause to the comms/callback graph. The branch's own test says exactly this in its docstring — honest, and it matches what I measured.

## Q2 — RESOLVED (nit, round 1). The 30-cluster ceiling now holds everywhere I could push it.

The round-1 repro, verbatim: `'a-really-very-long-subagent-j…'` = **30 clusters** (was 31).

| fuzz | cells | worst observed | over 30 |
|---|---|---|---|
| exhaustive ASCII, parent 0–60 × label 0–40 | 2,501 | 30 | **0** |
| `_browser_clip_label`, budgets 8–30 × 6,000 cleaned texts | 138,000 | ≤ budget | **0** |
| graphemes — emoji, ZWJ sequences, combining marks, CJK, Arabic, bidi, zero-width | 4,000 | 30 | **0** |
| non-subagent paths (plain title, cwd) | 8 | 30 | **0** |

The **F5 dropout point is documented accurately**: measured 18 → parent shown, 19 → shown, **20 → dropped**, 21 → dropped, exactly as the new docstring states.

One thing I looked at and am **not** reporting as a finding: at budgets 0–3 `_browser_clip_label` returns the 7-cluster `'Session'` fallback, which exceeds its budget. Not reachable in production — the only caller passing a custom budget is guarded by `room >= _BROWSER_SUBAGENT_PARENT_MIN` (8) — and 7 ≤ 30 regardless.

## Q3 — NOT RESOLVED. **BLOCKER** (release mechanics, not the fix): `0.44.59` is now published too.

The remediation verified `0.44.59` free at `15:55:19Z`. **PR #586 released it at `16:01:45Z`, six minutes later.**

```
PR head declares      : version = "0.44.59"
git tag v0.44.59      : 73fe1226  (feat(tools): coerce job_id shapes … (#586))
gh release v0.44.59   : "0.44.59: Coerce job targets and resolve labels in wait and jobs"
                        published 2026-09-03T16:01:45Z
PyPI 0.44.59          : HTTP 200 — wheel 16:03:02.547Z, sdist 16:03:04.393Z (now `latest`)
origin/main           : 73fe1226, version = "0.44.59"
```

This is the same failure shape as round 1, not a repeat mistake: the bump was correct when written and `origin/main` moved underneath it again. `git merge-tree --write-tree origin/main d55840e9` resolves **cleanly** — no textual conflict in `pyproject.toml` — so nothing flags it at merge time and the failure only surfaces at `gh release create`, after the squash has landed.

**`0.44.60` is free** (no tag, PyPI 404). Worth noting before picking it: **PRs #564 and #557 also declare `0.44.59`**, and #585 declares `0.44.57` (skipped on PyPI but claimed), so whichever of these merges first re-stales the others. **PATCH remains the correct class** per AGENTS.md — a fix to existing behaviour, no new surface.

## Q4 — NIT (new): a docstring still names the parameter this remediation removed.

`session.py:1381` describes `job_label` as *"this plus ``parent_conversation_name`` below"*, but that parameter no longer exists — `5fc80ffe` renamed it to `parent_display_name`, and the block **directly beneath** the sentence documents `parent_display_name` correctly. It is the only remaining mention of the old name in the tree (`grep -rn parent_conversation_name --include=*.py .` → 1 hit, that docstring; 0 in `subagent.py`). Cosmetic, one word, and the kind of thing that sends the next reader looking for a field that is not there. No behavioural consequence.

---

## Regression sweep — clean

**Round-1 cells re-executed:** 6-sibling fleet (6 of 6 distinct, `requester` per-child) · rename-after-open, including a **stale context snapshot re-read at call time** (`Renamed after the… › sub`) · fork suppression (an unnamed fork gets `'oh-my-pi'`, its cwd, never a parent's title) · bang mode through the real `OperatorApp` pilot, including a **title landing mid-command** (`Named While The Bang Ran`) · 14 sanitisation cells (bidi stripped, zero-width-only label falls back to the parent, newline/tab collapse, combining marks counted as one cluster, a parent below the 8-cluster minimum drops out rather than showing a stub, a top-level session with a stray `job_label` and `job_id=None` never composes).

**F3** — `_peer_sender_conversation_name`, six shapes, all correct: child of a named parent → `Fix browser tab group naming › qa-tester`; unnamed parent → `qa-tester` alone; top-level → its own title; server session (`job_id` set, no label) → its own title; whitespace-only label → parent only; and a long pair returns **unclipped** at 103 chars, which is right — a peer card has no 30-cluster budget.

**F4** — a server-shaped context (`job_id` set, `job_label=""`) keeps `'Server Side Conversation'`; a top-level with a stray label and `job_id=None` keeps `'Top Level'`. The discriminator really is both fields.

**The subagent engine** (`_build_child_session` changed signature): `task` launch → job row `type=task`, correct label; jobs settle `completed`; 3 concurrent launches all land in the table; `hub` wiring intact (child shares the parent's `SubagentComms`, holds `hub`/`task`/`wait`/`jobs`); `origin.json` stamped `{'origin': 'subagent', 'label': …, 'agent': 'task'}`; identity stays the child's throughout. **The resume path** (`hub op='resume'`, same builder) keeps the composed pill at depth 1 *and* depth 2, reusing the same session directory. No stale caller of the removed kwarg exists anywhere (only the Q4 docstring).

**The real paired extension**, all PASS: `open` · `read` · `snapshot` · `screenshot` · `logs` · `scroll` · `tabs` · `goto` · `close`; unapproved-origin `goto` refused **without navigating** (tab verified still on `example.com` afterwards); stale snapshot ref `e99` produces the correct refusal.

**The pill as the extension actually applies it** (`retitle` readback — the strongest programmatic evidence available for browser chrome):

```
top-level        -> 'LO · Fix browser tab group naming'
depth 1          -> 'LO · Fix browser tab… › qa-tester'
depth 2          -> 'LO · Fix browser tab group… › qa'
depth 3          -> 'LO · Fix browser tab group… › qa (2)'
conv B, depth 2  -> 'LO · Add Radient OAuth PKCE… › qa'
```

The `(2)` on depth 3 is the extension's own stable owner-ordinal for a **duplicate base label** — pre-existing behaviour with its own test (`tab-groups.integration.test.mjs`: *"duplicate labels get stable owner ordinals and rename updates in place"*), and here it is the documented same-conversation/same-label deferral surfacing exactly as the thread describes. The two different conversations stay distinct **as applied**, which is the property that matters.

**Mixed-version interop:** both trees put exactly `['requester', 'session_label']` on the wire, both `str`, across four context shapes. No new field — a longer string in one that already existed, so a new host against an older extension is byte-compatible.

**Gates and suites, whole tree as CI:** flake8 clean · black 26.1.0 `832 files unchanged` · isort 5.13.2 clean · pyright `0 errors, 0 warnings` · **full unit suite 11532 passed, 15 skipped** (422s; the known `test_prices.py` load flake **passed in-run** under load average 17–25, no isolation needed) · **e2e stage `4 passed`** · **extension suite 81 pass, 0 fail**. CI on this head: 13/13.

**The new tests are not tautologies.** Copied onto a `0dc28cf2` worktree they fail **17 of 130** with meaningful assertions — `assert 'local-operator' != 'local-operator'` (the grandchild collision), `AttributeError: … no attribute '_parent_display_name_resolver'`, `… '_peer_sender_conversation_name'`, `… '_browser_clip_label'`, `'ToolContext' object has no attribute 'job_label'`. That control worktree was restored with `git checkout -- tests/`; QA modified no repository.

## What needs the operator's eyes

Unchanged from round 1: a tab-group pill is browser chrome **outside the page viewport and cannot be screenshotted**. I verified it at the wire and at the extension's own applied-title readback. How it *renders* — colour, where the real font truncates, whether `›` reads clearly at pill size next to `LO ·` — remains unverified by QA.

## Gaps

- **Surface cap** — BLOCKED, as in round 1: exercising it means opening many real tabs in the operator's live browser. Untouched by this diff.
- **Concurrency** — the live cells drove one tab handle sequentially. Two genuinely concurrent sessions writing different pills to two different real groups at the same instant was not exercised; the extension's `serialize()` queue is the guard and its own suite covers it (81/81).

Full matrix, one row per cell with actual outputs: `/tmp/qa583-evidence/QA-MATRIX-round2.md`. Repros: `repro_Q3_round2.sh`, `c2_depth.py`, `c3d_isolate.py`, `c4_clusters.py`.

---

**Resolution of round-1 findings: Q1 RESOLVED · Q2 RESOLVED · Q3 NOT RESOLVED (recurred at a new version).**

**Verdict: FAIL** — on Q3 alone. The fix itself is clean: every cell I could point at the depth remediation passed, including four depths, a real depth-2 fleet, the weakref degradation and non-pinning claims, and 144,000+ boundary cells against the cluster ceiling. Q3 is release mechanics with a merge-time silent failure mode, and it must be corrected before merge — bump to `0.44.60` (free as of this writing) after a rebase, and re-verify at the moment of merge rather than at the moment of the bump, since main is moving faster than this PR's review cycle. Q4 is a one-word docstring nit, reasonable to fix in the same pass or defer.

damianvtran:

Implementer: `coder` subagent (lopdev team), author of this branch.
New head: `4b38c829ad6285be3d0bc9e7cc44d129807bda3b` — CI **green on this head** (all 13 checks pass). Remediation delta vs the round-1 head (`d55840e9`): **1 file, +1/−1, comment-only.**

| Finding | Severity | Disposition |
|---|---|---|
| **F7 / Q4** — `session.py:1381` still names the removed `parent_conversation_name` | NIT | **Fixed** (`4b38c829`). The `job_label` block now points at `parent_display_name`, the parameter that actually sits below it. Confirmed the repo-wide count of `parent_conversation_name` references is now **0** (`grep -rn` over `local_operator/` and `tests/`). Comment-only: no behaviour, no signature, no test expectation changes. |
| **F6 / Q3** — version in the branch is stale | BLOCKER (mechanics) | **Deferred to merge time, by manager decision — not an oversight.** `pyproject.toml` on this branch says `0.44.59` and that version is **already published**: PyPI returns 200 for `0.44.59`, and `origin/main` itself now declares `0.44.59` via merged #586, which landed ~6 minutes after round 1 verified that number free against tag/Release/PyPI. This is the **sixth** overtake today, so the branch has deliberately **not** been rebased or re-bumped: the manager will set the final version immediately before `gh release create`, after the squash lands. Commit-time bumping demonstrably cannot work under this merge rate — for the record, open PRs **#564 and #557 also currently declare `0.44.59`**, so any number picked here would be contended again before merge. Treat the version line in this diff as known-stale and expect it to be replaced at merge time. |

**Why no full re-verification this round.** The only change since the round-1 head is one word inside a `#:` comment block, so the round-1 evidence (distinct pills to depth 4, 6-of-6 sibling distinctness through the production `_launch_subagent`, the weakref proven genuinely weak, 144,500+ boundary cells with none over 30 clusters) applies unchanged to this head.

**Gates on the new head, whole tree as CI:** flake8 clean · black 26.1.0 `832 files unchanged` · isort 5.13.2 clean · pyright `0 errors, 0 warnings` · targeted suites `tests/unit/session tests/unit/browser_bridge tests/unit/harness tests/unit/tools/test_send_tool.py` → **1328 passed**.

**One CI flake, investigated rather than waved through.** The first run of this head failed `test (3.12, 4)` on `tests/unit/session/test_resume_subagents_todos.py::test_snapshot_entry_is_written_for_a_launched_child` (`assert []`). It is **not** attributable to this PR, and I checked rather than assumed: that test file is untouched by this branch (0 diff lines); this PR's `session.py` hunks are the constructor, `_display_session_name` and `_build_tool_context`, none of which is on the roster-persist path (the diff adds no reference to `_schedule_subagent_persist`, `roster`, or `sidecar`); the test passed 12/12 isolated on the branch; an interleaved A/B against an `origin/main` worktree passed 10/10 and 5/5 on **both** trees; and it still passed 8/8 on both trees under synthetic load (host load average 78). The test asserts on an async persist behind a 0.05 s sleep, and this repo's CI shows the same class of timing flake elsewhere (`main`'s own recent red run failed on `test_tui_bridge.py::test_tui_auto_registers_and_answers_control`). **Re-running the failed shard with no code change turned it green**, and the full run is now 13/13.

**Extension untouched.** Still zero files under `extension/` in this PR; the tree at head (`6f9ce808`) remains byte-identical to `origin/main`'s.

PR #584, opened 2026-09-03
fix(evaluation): keep the evidence an interrupted episode already bought

damianvtran:
## Consolidates PR #581

This PR now also carries **PR #581** (`dev-seal-finalization`, head
`4d6718a7`), which independently diagnosed the same root cause and fixed the
same `verify.py` hunk. The two fixes were merged here:

- The **stricter guard from this PR is the one shipped** — it is a superset of
  #581's: besides `failure_kind` on the terminal snapshot it additionally
  requires the bundle to be unscored (no scoring start, no scoring result,
  outcome not scored), which blocks the score-laundering shape #581's guard
  did not explicitly pin.
- The comment above the guard folds in #581's prose where it explains the
  mechanism more precisely (the hash-chain argument for why `failure_kind`
  is not a weaker explanation than a rollout-completion marker).
- **Both test files are kept**: #581's `tests/unit/evaluation/evidence/
  test_finalization.py` (seal/abandon paths, forged-finalization rejection,
  the scored-bundle-with-failure_kind pin) alongside this PR's additions to
  `test_verify.py` and `test_episode.py`. No #581 test asserted the weaker
  semantics — its crash bundles are unscored by construction, so they pass
  under the stricter guard unchanged.
- One `chore(release)` at tip: **0.45.1**, strictly above main (0.44.56),
  PyPI (0.44.56), and every open PR head (re-checked at bump and at push;
  highest was 0.45.0 on #576). Both source branches claimed 0.44.57.

PR #581 is closed as superseded by this one.

## What broke a real paid episode

Episode `ep-ffda3fc88f81` drove 16 environment steps on a live cloud desktop
(37 artifacts, 7.9 MB of frames, every step chained to the last), took an
unretryable adapter RPC error on the 17th, and lost **all of it**:

```
status: "abandonment_failed"
diagnostic: "EvidenceBundleInvalid: bundle failed independent seal verification;
             abandonment refused: EvidenceBundleInvalid: bundle failed independent
             abandonment verification"
score: null, reportability_label: null
```

Independent `verify_bundle` reported exactly one issue:
`code='finalization_invalid' severity='error' location='events.jsonl'`.

## Proven mechanism (reproduced cloud-free)

The RPC error ended the run — but it is **not** what made the run worthless.
The verifier required every episode that ever stepped to end *deliberately*:
its last step must be terminal (`terminated`/`truncated`) or a `finish` action
must be present (`verify.py`'s terminal check). An episode interrupted
mid-loop — adapter crash, provider outage, operator cancel — ends after an
ordinary non-terminal step, because the thing that would have produced the
terminal step is exactly what broke. So the bundle could neither seal nor be
abandoned, and the runner reported `abandonment_failed`.

Reproduced end-to-end with the real runner + `FakeAdapter` (no AWS): crash on
the second `execute` yields byte-identical `status`, `diagnostic`, and the same
`finalization_invalid` issue. The same shape kills a provider outage, so this
is not adapter-specific.

The fix is at the **verifier layer** and is benchmark-agnostic.

## Changes

1. **`verify.py`** — exempt exactly the interrupted case, keyed on the terminal
   snapshot's `failure_kind` (set on failure/cancel paths, `None` on a clean
   completion). The exemption additionally requires the bundle to be **unscored**,
   so a run that stopped early can never launder itself into a reportable score.
2. **`episode.py`** — record a bounded, redaction-scanned `detail_artifact` on
   the fatal error. `diagnostic_code` is derived from the exception *class*, so
   the paid episode recorded only `rpcremoteerror` with a null detail — enough
   to bucket a failure, never enough to diagnose one without paying for the run
   again. Publication is best-effort: it must never be the reason a bundle loses
   its terminal.

## Verification

- **Real bundle**: `verify_bundle(ep-ffda3fc88f81)` now returns `valid=True`,
  `issues=[]` (was `finalization_invalid`). Read-only; evidence untouched.
- **Regression tests** (`test_episode.py`): crash after a successful step seals;
  provider failure after a step seals; fatal error records a bounded diagnostic
  detail; a resolved secret cannot reach the bundle via that detail.
- **Verifier tests** (`test_verify.py`): interrupted episodes keep their
  non-terminal last step (crash/infrastructure/model/cancelled); a clean
  completion still needs a deliberate ending; an interruption cannot launder a
  scored result.
- **Finalization tests** (`test_finalization.py`, from #581): a crash
  mid-rollout verifies and seals unscored; the abandonment fallback succeeds on
  a recovered crash bundle; a crash bundle missing its terminal event still
  refuses; a truncated rollout without `failure_kind` still refuses; a scored
  bundle carrying a `failure_kind` still requires a completed rollout.
- **Mutation-proven on the consolidated head, both ways**: reverting the
  exemption to the original unconditional invariant fails 10 tests across both
  PRs' suites; dropping the unscored requirement from the exemption fails both
  laundering pins (`test_interruption_cannot_launder_a_scored_result` and
  `test_scored_bundle_with_failure_kind_still_requires_completed_rollout`).

## Gates (whole tree, as CI)

- `flake8` clean, `black --check` clean, `isort --check` clean, `pyright` 0 errors
- `tests/unit`: **11517 passed, 15 skipped**
- `tests/unit/evaluation`: **728 passed**
- `tests/e2e -m e2e -n0`: **4 passed**

## Retryability judgment (deliberate non-change)

The task asked whether the adapter error should be retried. It should **not**:
`execute` is a mutating call, and the supervisor poisons the session on any
failure of a mutating call (`_mutating_call` → `_poison_after_ambiguous_mutation`).
Retrying a possibly-applied mutation would double-apply the action and corrupt
the environment state the verifier tracks. The correct fix is to make the
*interruption* survivable (this PR), not to make the error retryable. The
worker's durable operation-replay cache already makes a *clean* retry of a
keyed method idempotent at the RPC layer; the poison gate is what prevents a
*dirty* retry, and that is the right behaviour.

### DIFF
--- a/local_operator/evaluation/evidence/verify.py
+++ b/local_operator/evaluation/evidence/verify.py
@@ -792,7 +792,52 @@ def _verify_semantics(
             issues.error("receipt_binding_invalid", terminal_location)
         if len(terminal_states) != 1:
             issues.error("finalization_invalid", terminal_location)
-        if environment_step_seen and not last_step_terminal and not finish_action_seen:
+        # An episode that ran environment steps must show WHY it stopped
+        # stepping: either the environment ended the rollout
+        # (terminated/truncated) or the agent chose to stop (a `finish` batch).
+        # Without one of those, a bundle truncated mid-rollout would otherwise
+        # seal as if the run had reached its own conclusion.
+        #
+        # That invariant holds only for an episode that was ALLOWED to end
+        # deliberately. A run interrupted mid-loop -- an adapter crash, a
+        # provider outage, an operator cancel -- stops after an ordinary
+        # non-terminal step, with no terminal step and no finish action,
+        # because the thing that would have produced one is exactly what
+        # broke. Requiring the deliberate ending there made every such bundle
+        # unsealable AND unabandonable -- unscoreable rather than merely
+        # unscored -- so a real paid episode that died at step 17 of 60
+        # destroyed the 16 steps of evidence it had already bought (observed
+        # as ep-ffda3fc88f81: `abandonment_failed`, score null, 37 artifacts
+        # stranded).
+        #
+        # The interruption is recorded by the terminal snapshot's
+        # ``failure_kind``, which the runner sets on precisely the failure and
+        # cancellation paths and leaves None on a clean completion. It is not a
+        # weaker explanation than a rollout-completion marker: it is written
+        # only by ``record_final_lifecycle`` under the finalizing marker, and
+        # it rides inside the hash-chained event payload, so it cannot be
+        # added to an existing bundle without breaking the chain.
+        #
+        # The exemption is deliberately narrower than "failure_kind is set". A
+        # bundle that claims BOTH an interruption and a score would otherwise
+        # launder a run that stopped early into a reportable result -- the
+        # score would be computed over a state the episode never deliberately
+        # reached. An interrupted episode is unscored by construction
+        # (``_finalize_failure`` seals ``ScoreArtifact(status="unscored")``),
+        # so requiring that here costs the real path nothing and keeps the
+        # laundering shape rejected.
+        unscored = (
+            not starts and not results and (outcome is None or outcome.result.status != "scored")
+        )
+        interrupted = (
+            len(terminal_states) == 1 and terminal_states[0].failure_kind is not None and unscored
+        )
+        if (
+            environment_step_seen
+            and not last_step_terminal
+            and not finish_action_seen
+            and not interrupted
+        ):
             issues.error("finalization_invalid", terminal_location)
         if terminal_output_observation is not None and not terminal_output_resolved:
             issues.error("receipt_binding_invalid", terminal_location)
--- a/local_operator/evaluation/runner/episode.py
+++ b/local_operator/evaluation/runner/episode.py
@@ -1083,12 +1083,43 @@ class EpisodeRunner:
             category, reason, failure_kind = "model", "model_failure", "model"
         else:
             category, reason, failure_kind = "adapter", "crash", "crash"
+        # The bounded, redaction-scanned reason for the failure, not just its
+        # type. ``diagnostic_code`` is a StrictIdentifier derived from the
+        # exception CLASS ("rpcremoteerror"), which is enough to bucket a
+        # failure and never enough to diagnose one: the real episode
+        # ep-ffda3fc88f81 burned a paid 16-step run and left a reader nothing
+        # but that word and a null ``detail_artifact``. The rejected-decision
+        # path above already publishes its diagnostic as an artifact for
+        # exactly this reason; a fatal error deserves it at least as much,
+        # because there is no retry that will produce a second chance to look.
+        #
+        # ``_diagnostic`` is the same renderer used for the outcome's own
+        # diagnostic field: it strips pydantic's ``input_value=`` echo (the one
+        # place a resolved secret could surface) and truncates to 500 chars, and
+        # ``publish_artifact`` independently scans every byte against the
+        # episode's RedactionSet, so a leaked credential fails the write rather
+        # than reaching the bundle.
+        detail: Any = None
+        try:
+            detail = self._publish(_diagnostic(error).encode("utf-8"), media_type="text/plain")
+        except (_EvidenceFailure, OSError):
+            # Best-effort by design. This runs on the path that is already
+            # handling a failure, and an unpublishable detail must never be the
+            # reason a bundle loses its terminal -- the code, category and
+            # outcome diagnostic still reach the reader without it. ``OSError``
+            # is named alongside ``_EvidenceFailure`` because ``publish_artifact``
+            # poisons the writer and re-raises a RAW ``OSError`` on ambiguous I/O
+            # (ENOSPC/EIO), and ``_publish`` converts only ``EvidenceError`` -- so
+            # a disk failure here would otherwise escape ``_finalize_failure`` and
+            # cost the bundle its terminal: the exact shape this change removes.
+            detail = None
         self._append(
             "error",
             ErrorPayload(
                 error_id=f"err-{uuid.uuid4().hex[:12]}",
                 category=category,
                 diagnostic_code=_diagnostic_code(error),
+                detail_artifact=detail,
                 retryable=False,
             ),
         )
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "local-operator"
-version = "0.44.69"
+version = "0.44.67"
 description = "A team of proactive AI assistants that can work together behind the scenes to help you get mundane tasks done so you can focus on the fun stuff."
 readme = "README.md"
 authors = [{ name = "Damian Tran", email = "damian@gominerva.com" }]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,11 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "local-operator"
-version = "0.44.67"
+<<<<<<< HEAD
+version = "0.44.70"
+=======
+version = "0.44.70"
+>>>>>>> 4d6718a7 (chore(release): 0.44.57)
 description = "A team of proactive AI assistants that can work together behind the scenes to help you get mundane tasks done so you can focus on the fun stuff."
 readme = "README.md"
 authors = [{ name = "Damian Tran", email = "damian@gominerva.com" }]
--- a/tests/unit/evaluation/evidence/test_finalization.py
+++ b/tests/unit/evaluation/evidence/test_finalization.py
@@ -820,3 +820,454 @@ def test_terminal_conflict_and_outcome_tamper_are_detected(tmp_path: Path) -> No
         "outcome_mismatch",
         "counter_mismatch",
     } & {issue.code for issue in report.issues}
+
+
+def _crashed_mid_rollout(
+    root: Path, *, failure_kind: str | None = "crash", steps: int = 2
+) -> tuple[EvidenceWriter, ScoreArtifact]:
+    """Write the exact interleaving a crash-terminated episode leaves on disk.
+
+    ``steps`` non-terminal environment steps (no ``finish`` batch, nothing
+    terminated or truncated), then the adapter error, then the unscored
+    finalization, receipts, and the terminal snapshot -- the shape produced by
+    ``EpisodeRunner._finalize_failure`` when the environment RPC dies mid-run.
+    """
+
+    writer = EvidenceWriter.create(root, manifest(), RedactionSet.from_resolved_values(()))
+    try:
+        clock = 0
+        _append_provenance(writer, monotonic_ns=clock, wall_time_ms=clock)
+        clock += 2
+        action_artifact = writer.publish_artifact(b"safe", media_type="text/plain")
+        for index in range(steps):
+            writer.append(
+                "observation",
+                ObservationPayload(observation_id=f"observation-{index}", sequence=index),
+                monotonic_ns=clock,
+                wall_time_ms=clock,
+            )
+            clock += 1
+            writer.append(
+                "action_batch",
+                ActionBatchPayload(
+                    action_batch_id=f"batch-{index}",
+                    observation_id=f"observation-{index}",
+                    action_count=1,
+                    action_artifact=action_artifact,
+                ),
+                monotonic_ns=clock,
+                wall_time_ms=clock,
+            )
+            clock += 1
+            writer.append(
+                "environment_step",
+                EnvironmentStepPayload(
+                    step_id=f"step-{index}",
+                    action_batch_id=f"batch-{index}",
+                    receipt_id=DIGEST,
+                    input_observation_id=f"observation-{index}",
+                    output_observation_id=f"observation-{index + 1}",
+                    terminated=False,
+                    truncated=False,
+                ),
+                monotonic_ns=clock,
+                wall_time_ms=clock,
+            )
+            clock += 1
+        # The rollout's final output observation still lands: the step receipt
+        # resolved before the adapter broke.
+        writer.append(
+            "observation",
+            ObservationPayload(observation_id=f"observation-{steps}", sequence=steps),
+            monotonic_ns=clock,
+            wall_time_ms=clock,
+        )
+        clock += 1
+        writer.append(
+            "error",
+            {
+                "error_id": "err-306591b1400e",
+                "category": "adapter",
+                "diagnostic_code": "rpcremoteerror",
+                "retryable": False,
+            },
+            monotonic_ns=clock,
+            wall_time_ms=clock,
+        )
+        clock += 1
+        writer.begin_finalization(
+            "final",
+            None,
+            FinalizationIntent(kind="unscored"),
+            monotonic_ns=clock,
+            wall_time_ms=clock,
+        )
+        clock += 1
+        clock, _ = _append_final_receipts(writer, monotonic_ns=clock, wall_time_ms=clock)
+        score = ScoreArtifact(status="unscored", reason="crash")
+        writer.record_final_lifecycle(
+            LifecycleTransitionPayload(
+                previous_state_id=None,
+                state_id=OTHER_DIGEST,
+                state="completed",
+                finalization_id="final",
+                preflight_seal_id=DIGEST,
+                commitment_id=OTHER_DIGEST,
+                reconciliation_id=DIGEST,
+                reconciliation_reportable=True,
+                score_id=score.score_id,
+                cleanup_result_id=OTHER_DIGEST,
+                rescue_required=False,
+                failure_kind=failure_kind,  # pyright: ignore[reportArgumentType]
+            ),
+            monotonic_ns=clock,
+            wall_time_ms=clock,
+        )
+    except BaseException:
+        writer.close()
+        raise
+    return writer, score
+
+
+def test_crash_after_nonterminal_steps_verifies_and_seals_unscored(tmp_path: Path) -> None:
+    """A crash mid-rollout must seal as unscored, not fail verification.
+
+    Regression: the rollout-completion invariant demanded a terminated/truncated
+    step or a ``finish`` batch on every bundle carrying a terminal snapshot. A
+    crash leaves neither -- nothing got to decide the rollout was over -- so the
+    bundle failed its own seal verification AND the abandonment fallback,
+    leaving the episode unscoreable instead of unscored.
+    """
+
+    root = tmp_path / "bundle"
+    writer, score = _crashed_mid_rollout(root)
+    try:
+        report = verify_bundle(root)
+        assert report.valid, report.issues
+        assert report.terminal_state == "finalizing"
+        outcome = writer.seal(
+            OutcomeDraft(
+                finalization_id="final",
+                preflight_seal_id=DIGEST,
+                commitment_id=OTHER_DIGEST,
+                reconciliation_id=DIGEST,
+                cleanup_result_id=OTHER_DIGEST,
+                result=score,
+                reportability_label="unscored",
+                comparability_label="comparable",
+                ended_wall_time_ms=99,
+            )
+        )
+    finally:
+        writer.close()
+    assert outcome.result.status == "unscored" and outcome.result.reason == "crash"
+    sealed = verify_bundle(root)
+    assert sealed.valid, sealed.issues
+    assert sealed.terminal_state == "sealed"
+    assert sealed.outcome is not None
+    assert sealed.outcome.reportability_label == "unscored"
+
+
+def test_crash_abandonment_fallback_now_succeeds(tmp_path: Path) -> None:
+    """The safety net must work: a recovered crash bundle can be abandoned.
+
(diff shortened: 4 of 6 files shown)


damianvtran:
### Agent review — round 1

Reviewer: reviewer subagent (claude-opus-5)
Scope: 8af528dc..c5eabe7d

Independent review on a clean detached worktree at head `c5eabe7d`. I did not write this code. All reproduction below was run, not inferred.

**Verdict: NOT terminal — 1 MAJOR remains.** The core verifier fix is correct and mutation-proven; the new diagnostic-detail path reintroduces, for one exception class, the exact failure mode this PR exists to remove.

---

**Verified by reproduction (all pass):**

1. **Real bundle.** `verify_bundle` over a writable copy of `ep-ffda3fc88f81`: at head → `valid=True, terminal_state=finalizing, issues=[]`; at `origin/main` (0dc28cf2) → `valid=False` with exactly one `finalization_invalid @ events.jsonl`. Bundle shape confirmed: 16 `environment_step` events, terminal lifecycle `completed` with `failure_kind=crash`, zero scoring events, `state.json` intent `unscored` — precisely the exempted shape.
2. **Guard is the stricter candidate and matches the spec byte-for-byte** (`verify.py:829-840`): `unscored = not starts and not results and (outcome is None or outcome.result.status != "scored")`; `interrupted = len(terminal_states)==1 and terminal_states[0].failure_kind is not None and unscored`. Compared against #581's `crash_terminated = not results and any(...)` at `4d6718a7` — this one additionally requires no scoring start and a single terminal, so it is the strict superset, as claimed.
3. **Mutation battery** (mutated my worktree, ran, restored; verify.py byte-identical after):
   - Revert exemption → original unconditional invariant: **10 tests fail** across both PRs' suites (matches the claimed count).
   - Drop `and unscored`: exactly the two laundering pins fail (`test_interruption_cannot_launder_a_scored_result`, `test_scored_bundle_with_failure_kind_still_requires_completed_rollout`).
   - Drop `failure_kind is not None` (keep unscored): 2 deliberate-ending pins fail — exemption can't widen to "any unscored terminal".
   - (b) scored+failure_kind never verifies; (c) failure_kind+scoring-start refused (empirical probe: `finalization_invalid`); (d) missing terminal event refused (`test_crash_bundle_missing_terminal_event_still_refuses`).
4. **`failure_kind` untamperability.** Written only via `record_final_lifecycle` under the finalizing marker (`store.py:1238-1271`), inside the hash-chained payload. Empirical retrofit of `failure_kind` into a sealed terminal event → `event_hash_mismatch`. Claim holds.
5. **detail_artifact**: bounded at 500 (`_diagnostic` `[:500]`, pinned by `byte_count <= 500`); redaction via the store's own `_RedactionScanner` (`store.py:1022`), not a second implementation; canary test confirms a resolved secret reaches no file and the refused detail costs nothing. Benchmark-agnostic: no OSWorld special-casing; keyed on `failure_kind ∈ {crash, infrastructure, model, cancelled}`.
6. **Version 0.45.1** strictly above main (0.44.58), PyPI (0.44.58), and every open PR head (max 0.45.0 on #576).

**Gates (whole tree, as CI, `ulimit -n 8192`):** flake8 clean, black clean (832 files), isort clean, pyright 0 errors, `tests/unit` **11517 passed / 15 skipped** (matches PR claim), `tests/e2e -m e2e -n0` **4 passed**. Machine was contended; the suite self-caps parallelism, unit run 7m19s.

---

**Findings**

**MAJOR — R1-1: the detail-artifact "best-effort" catch is narrower than the operation's exception surface; a raw `OSError` during publication escapes `_finalize_failure` and loses the terminal.**
`episode.py:1094-1101` wraps the publish in `except _EvidenceFailure`, but `_publish` (`episode.py:1376-1383`) converts only `EvidenceError`. `publish_artifact` can raise a **raw `OSError`** — ENOSPC/EIO on the temp `os.open`, in `_write_all`, or on `fsync` (`store.py:1064-1080` catches `BaseException`, poisons, and re-raises the raw primary at `store.py:1080`). That `OSError` passes through both `except EvidenceError` and `except _EvidenceFailure`, propagates out of `_finalize_failure` → `_run_with_bundle` → `run()`, so the caller gets a bare exception, no `EpisodeOutcome`, and the bundle is left with its steps but no error event / finalization / terminal. This is exactly the "failure handler destroys the bundle's terminal" shape the comment at `episode.py:1097-1100` asserts is impossible and that this PR exists to fix — just gated on a disk-error instead of an adapter error.
Reproduced: forcing an `OSError(ENOSPC)` out of the detail publish makes `runner.run()` raise instead of returning an outcome.
Fix (one line): broaden the catch to `except (EvidenceError, OSError)` (or `except BaseException` with the same best-effort comment), or have `_publish` convert `OSError` to `_EvidenceFailure` too.

**MINOR — R1-2: `not starts` and `not results` are not independently pinned.** Dropping only `not starts` (M4) or only `not results` (M5) each leaves all 89 tests green, because both laundering tests carry a scoring start *and* result together. The current code is correct (empirical probe: scoring-start-without-result + failure_kind is refused), but a future refactor could drop either clause with no test biting. One test with a scoring start but no result + `failure_kind` would pin `not starts` on its own.

**MINOR — R1-3: no test specifically retrofits `failure_kind` into a sealed terminal event.** The hash-chain argument in the `verify.py:814-819` comment is the security basis for the exemption and is true (verified empirically), but it rides only on the generic `test_event_tamper_...` payload-mutation pin. A dedicated retrofit pin would lock the exact claim the comment makes.

---

**Counts: 0 blocker, 1 major, 2 minor.**

Not terminal. Fix R1-1 (one-line catch broadening) and add a regression test for it; R1-2/R1-3 are cheap pins worth folding into the same round. Re-request review on the new head.

damianvtran:

All three findings fixed in `b7a453e0` on this head. No merge, no release; evidence bundle untouched (read-only throughout).

**R1-1 (MAJOR) — fixed.** The catch at the detail-publish site (`episode.py:1096`) is widened from `except _EvidenceFailure` to `except (_EvidenceFailure, OSError)`. Layer choice: the call SITE, not `_publish`. `_publish` converts only `EvidenceError` by contract, and its other seven callers correctly route a raw `OSError` to `_finalize_failure` via `_execute`'s `except BaseException` — only this site sits *inside* `_finalize_failure`, so only it needs the wider catch; converting `OSError` inside `_publish` would change semantics for callers that are already correct. The comment now names the exact mechanism (poison + raw `OSError` re-raise, `_publish`'s `EvidenceError`-only conversion).
- Reproduced pre-fix: forcing `OSError(ENOSPC)` out of the detail publish makes `run()` raise with no outcome (the reviewer's exact shape).
- Post-fix: `run()` returns a structured `abandonment_failed` outcome; the already-bought step is still on disk; a fresh process can `open_for_abandon` + `abandon` it, reaching a valid `abandoned` terminal — the best-effort promise holds for the whole failure class, not just `_EvidenceFailure`.
- Regression test: `test_detail_publish_os_error_cannot_escape_the_failure_handler` (test_episode.py). Mutation-proven: reverting to the narrow catch makes it fail with `OSError: [Errno 28] No space left on device`.

**R1-2 (minor) — fixed.** Both clauses pinned independently in test_verify.py, using scoring shapes the writer's phase machine refuses to emit (drop the event, recompute the hash chain):
- `test_scoring_start_without_result_still_refuses_interruption` pins `not starts`: a scoring start with no result + `failure_kind` still refuses; the exemption's `finalization_invalid` count is exactly 1, and dropping `not starts` (M4) falls it to 0 — verified by mutation.
- `test_scoring_result_without_start_still_refuses_interruption` pins `not results`: a scoring result with no start carries one structural `finalization_invalid`, and the exemption must add a second; dropping `not results` (M5) falls the count to 1 — verified by mutation.

**R1-3 (minor) — fixed.** `test_retrofitting_failure_kind_into_a_sealed_terminal_is_detected` (test_finalization.py) pins the comment's security claim directly on a SEALED bundle, both attack variants:
- NAIVE retrofit (edit payload, keep `event_id`) → `event_hash_mismatch` + `outcome_mismatch`.
- REHASHED retrofit (recompute `event_id` so the record is internally consistent) → still caught by `outcome_mismatch`, because the sealed outcome's `final_event_sha256` binds the ORIGINAL final event. Mutation-proven: removing that binding lets the rehashed retrofit pass.

**Gates (whole tree, as CI, at head `b7a453e0`):** flake8 clean, black 26.1.0 clean (832 files), isort 5.13.2 clean, pyright 0 errors, `tests/unit` **11521 passed / 15 skipped** (+4 new), `tests/unit/evaluation` **732 passed**, `tests/e2e -m e2e -n0` **4 passed**.

**Version:** unchanged at **0.45.1** — re-checked at handoff, still strictly above main (0.44.66), PyPI (0.44.66), and every other open PR head.

Re-requesting review on head `b7a453e0`.

damianvtran:
### Agent review — round 2

Reviewer: `reviewer` subagent (independent; not the author) — anthropic/claude-opus-5
Scope: `c5eabe7d..b7a453e0` (remediation diff; head `b7a453e0a561b0fcbe0eab52d5ee7ccda880c95d`)

Freshness pass over the round-1 remediation. Per the round-2 rule I scoped to the remediation diff and did not re-audit the guard, the laundering shapes, or the detail artifact beyond confirming they still hold.

**No findings.** Blockers: **0**. Majors: **0**. Minors: **0**. Nits: **0**.

#### Scope (clean)
`c5eabe7d..b7a453e0` is 1 commit / 4 files / +205 −2: the `episode.py` catch broadening (+9 −2) and three test files. No production code outside `episode.py`. Nothing surprising.

#### R1-1 — detail-publish OSError — **verified fixed by reproduction**
- The catch at the detail-publish site is `except (_EvidenceFailure, OSError):` (`episode.py:1096`), and `_publish` (`episode.py:1381`) is **unchanged** — it still converts only `EvidenceError`.
- **Pre-fix repro reproduced.** Reverting only that catch to `except _EvidenceFailure:` and forcing a poisoned writer + raw `OSError(ENOSPC)` out of the detail publish: `OSError: [Errno 28]` propagates out of `run()` — no `EpisodeOutcome`, bundle keeps its steps but loses its terminal. At this head the same scenario returns a structured outcome (`status="abandonment_failed"`), the 1 environment step still verifies, and a fresh process still abandons the bundle to `terminal_state="abandoned"`.
- **The author's reasoning for not widening `_publish` is confirmed, not taken on trust.** All 7 other callers (`_record_observation` ×2, `_decide_once` ×2, `_append_batch`, `_run_ask` ×2) sit inside `_execute`'s try, whose `except BaseException` routes to `_finalize_failure`. I drove a poisoned-writer raw `OSError` through each publish site (10 parametrised cells, both media types): **every cell returned an `EpisodeOutcome`, none propagated out of `run()`**. Leaving `_publish` narrow is correct — widening it would have swallowed OSErrors that currently route correctly.

#### R1-2 / R1-3 — new pins bite — **verified by mutation**
Each pin was confirmed to fail under the mutation it exists to catch (a passing test proves nothing on its own):

| Mutation | Result |
|---|---|
| Drop `not starts` (M4) | `test_scoring_start_without_result_still_refuses_interruption` **FAILS** (`finalization_invalid` disappears) |
| Drop `not results` (M5) | `test_scoring_result_without_start_still_refuses_interruption` **FAILS** (count 2 → 1) |
| Neuter `final_event_sha256` binding | `test_retrofitting_failure_kind_into_a_sealed_terminal_is_detected` **FAILS** (`outcome_mismatch` lost) |

Both exemption clauses are now independently pinned, and the retrofit test covers the naive **and** rehashed shapes — the rehashed one is the load-bearing case, since it is caught by the sealed outcome's binding alone.

#### Regression — clean
- **Real bundle `ep-ffda3fc88f81`** (copied; source untouched, mtime unchanged): at this head `valid=True`, `terminal_state="finalizing"`, **0 issues**, 121 events. Against `origin/main` the same bundle is `valid=False` with `finalization_invalid` — the fix demonstrably rescues the stranded episode.
- Deliberate-ending and laundering pins still refuse. Full changed-area files: **93 passed**.

#### Gates (as CI, machine contended — load avg ~28)
flake8 **0** · black `--check` clean under CI-pinned **26.1.0** (832 files) · isort clean under CI-pinned **5.13.2** · pyright **0 errors, 0 warnings** · unit suite **11,520 passed, 15 skipped, 1 failed in 471s**.

The single failure is `tests/unit/model/test_prices.py::test_a_background_revalidation_gets_the_full_default_timeout` — a **timing flake unrelated to this diff**: it passes in isolation on this head, the whole file passes on `origin/main`, and the PR touches no pricing code. Not a finding against this PR.

> Two local-only artifacts worth recording so the next reader does not re-diagnose them: local `isort` 9.0.1 flags 3 files (`test_configure.py`, two `docs/assets/**/shot.py`) that are **identical on `main`** and pass under CI's pinned 5.13.2; and `pyright` reports 1084 phantom import errors unless pointed at an interpreter (`--pythonpath`), because a detached worktree has no `.venv`. Neither is a defect in this change.

#### Version
`0.45.1` is strictly above `main` (0.44.66), PyPI latest (0.44.66, and 0.45.1 is unpublished), and every open PR (#593 0.44.67, #589 0.44.65, #578 0.44.44, #576 0.45.0, #564 0.44.67, #545 0.44.41). Checked at the end of the review.

#### Merge notes for the manager (not findings)
1. **No CI run exists for `b7a453e0`.** The only green run (`33769836049`, all 13 jobs) is against `c5eabe7d`, the round-1 head; `gh pr checks` reports "no checks reported". The head has been on the remote ~20 min and Actions is live on other branches, so the `pull_request` trigger appears not to have fired. My local gate run is a substitute for, not proof of, CI — **confirm a green CI run on `b7a453e0` before merging**, re-pushing the ref or using `workflow_dispatch` if the trigger stays stuck.
2. **`mergeable: CONFLICTING`** is the expected version collision only: `pyproject.toml` `0.44.66` vs `0.45.1`. I test-merged into `origin/main` — every code and test file auto-merged, the sole conflict is that one line.

### Verdict: **clean** — TERMINAL

No blockers, no majors, no minors, no nits. The round-1 MAJOR is fixed and reproduced fixed; both minors are fixed and each new pin is mutation-verified. **This round is terminal — no round 3 is warranted.**

Recommendation: **merge**, once (1) a green CI run exists on `b7a453e0` and (2) the `pyproject.toml` version conflict is resolved in favour of `0.45.1`, subject to the independent QA round on this head.

damianvtran:
Rebuilt on current `main` (`6437caf6`, 0.44.66) — new head `ec0f64aa`, version **0.44.67**. Manager-performed.

Why this was not a plain rebase: the branch carried a merge commit (the #581 consolidation) and, more importantly, **`main` had since gained PR #582**, which edits the *same region* of `episode.py` (`DecisionRejected.reply` publishing). A naive `git checkout <old-head> -- episode.py` would have silently REVERTED #582. I caught that by diffing the reconstructed patch against the reviewed one, then did a proper 3-way merge per file (`git merge-file` against the true merge base) — clean, no conflict markers.

Both changes are present and verified at this head:
- #582's reply publishing (`the REPLY is published alongside it` comment intact)
- #584's `except (_EvidenceFailure, OSError)` at `episode.py:1105` and the `interrupted` guard at `verify.py:832`

**Real-bundle acceptance re-run at this head:** a copy of `ep-ffda3fc88f81` verifies `valid=True`, `issues=[]`. Original untouched (mtime unchanged).

**Gates at this head, as CI:** flake8 clean · black 26.1.0 (834 files unchanged) · isort 5.13.2 clean · pyright 0 errors/0 warnings · `tests/unit/evaluation` **743 passed**.

Round 2 was terminal against `b7a453e0`, but since I reconstructed the head rather than fast-forwarding it, a short confirmation round follows — the reviewer should verify the 3-way merge preserved both PRs' behaviour rather than re-auditing the guard.

damianvtran:
### Agent review — round 3

**Reviewer:** `reviewer` subagent (independent — not the author of this branch), model `anthropic/claude-opus-5`
**Scope:** `6437caf67937f693f639b7ac501080a9d2f12c58..ec0f64aa8f11fc49c9cf4ec7ed3dbbbda588d181` (confirmation pass on the reconstructed head; the guard, laundering shapes, and detail artifact were settled in round 2 and were not re-audited)

Short confirmation round: does the manager's per-file 3-way reconstruction onto current `main` faithfully preserve both PRs? Verified in a clean detached worktree at `ec0f64aa`.

---

#### 1. Nothing was reverted — confirmed, structurally

The strongest available evidence, not a spot check:

- `git merge-base origin/main ec0f64aa` = `6437caf6` — i.e. the review-time `origin/main` tip **is** the merge base. The branch is a strict linear descendant; `git merge-base --is-ancestor origin/main HEAD` → **YES**. A reconstruction that reverted anything could not satisfy this.
- The **entire** `origin/main...HEAD` diff deletes exactly **two** lines: the old guard condition in `verify.py` and `version = "0.44.66"`. Nothing else in the tree changes away from main.
- No merge commits remain on the branch (`git log --merges origin/main..HEAD` empty) — the #581 consolidation merge is gone as intended.

**#582 specifically is intact.** `git diff <#582 merge commit 0dc28cf2>..HEAD -- episode.py` shows **one hunk, pure addition, zero deletions** — `DecisionRejected.reply` publishing, `_rejection_detail`, and the rejected-decision comment block are byte-identical to the version #582 merged. The other seven files #582 touched (`protocol.py`, `model.py`, `provider_client.py`, `conftest.py`, and its three test files) are byte-identical to `origin/main` — they do not appear in this PR's diff at all.

Behaviour, not just bytes: `test_provider_client.py` + `test_actions.py` → **65 passed**; `-k "rejected or reply or decision"` across `tests/unit/evaluation` → **46 passed**.

#### 2. Both fixes coexist at this head — confirmed

`episode.py:1105` carries `except (_EvidenceFailure, OSError)` and `verify.py:832` carries the `interrupted` guard, simultaneously with #582's reply publishing.

Round-2 behaviours re-proved by **mutation**, not inspection — each pin was reverted in a throwaway scratch worktree and observed to fail:

| Pin | Mutation applied | Result |
|---|---|---|
| R1-1 OSError | narrow back to `except _EvidenceFailure:` | **FAILS** — `OSError: [Errno 28]` escapes `run()` (`test_detail_publish_os_error_cannot_escape_the_failure_handler`) |
| M4 `not starts` | drop the clause | **FAILS** — `test_scoring_start_without_result_still_refuses_interruption` |
| M5 `not results` | drop the clause | **FAILS** — `test_scoring_result_without_start_still_refuses_interruption` (count 1 ≠ 2) |
| failure_kind retrofit | drop the `failure_kind is not None` condition | **FAILS** — `test_completed_episode_still_needs_a_deliberate_ending` |

All three pins bite with a distinct failure; unmutated, the three changed test files are **95 passed**. Scratch worktree restored clean.

#### 3. Real bundle — confirmed, with contrast

A **copy** of `ep-ffda3fc88f81` (original never touched — `events.jsonl` sha `362fdb6e…` and all mtimes identical before and after):

- at `ec0f64aa`: `valid=True`, `terminal_state=finalizing`, **`issues=0`**
- at `origin/main` (`6437caf6`, separate worktree): `valid=False`, **`finalization_invalid`** at `events.jsonl`

Same bundle, same code path, opposite verdict — the fix is load-bearing on the real paid episode.

#### 4. Scope — exactly as specified

`verify.py`, `episode.py`, `pyproject.toml`, and the three test files. Nothing else. Exactly one `chore(release)` at tip (`ec0f64aa`), over one implementation commit (`1b47bc8e`).

#### 5. Gates — complete, as CI, over the whole tree

flake8 **rc=0** · black 26.1.0 **834 files unchanged** · isort 5.13.2 **clean** · pyright **0 errors, 0 warnings** · full `tests/unit` **11777 passed, 15 skipped** in 444s.

The machine was genuinely contended (load average 60-68 throughout) but the suite ran to completion — this is a real green, not a starved one. **CI on this head: `Pre-release CI Testing` — success.**

---

### Findings

**BLOCKER-1 — version 0.44.67 is no longer above `main` or PyPI (release-blocking, not code).**
This is the eighth collision of the session and it landed *during* this review. At review start: main `6437caf6`/0.44.66, PyPI 0.44.66 — fine. At review end: `main` is now `06928224` (PR #593, `perf(providers)`) carrying **`version = "0.44.67"`**, and **PyPI already serves 0.44.67**.

Both branches bumped `0.44.66` → the *same* `0.44.67`, so git sees an identical change and reports **no textual conflict** (`git merge-tree` → 0 conflicts) — it would merge silently and then fail at publish, because the tag/PyPI version already exists.

*Fix (one line):* bump `pyproject.toml` to **0.44.68** and re-confirm against `origin/main` and PyPI immediately before merging.

No behavioural risk from #593: it touches `discovery.py`, `controller.py`, `registry.py`, `tui/app.py` and their tests — disjoint from this PR's six files. Re-verification after the bump can be scoped to the version line plus a re-run of the evaluation tests; the reconstruction findings above are unaffected.

No other findings. No MINOR, no NIT worth a remediation round.

**Blockers: 1 (version only) · Majors: 0**

---

### Verdict

The reconstruction is **faithful** — that was this round's question, and the answer is unambiguous: `origin/main` was a strict ancestor, `episode.py` is a pure addition over #582's own merge commit, and the whole diff deletes two lines. Nothing from #582 or anywhere else in `main` was rolled back. Both fixes coexist and both are proved by mutation, not by reading.

**This round is TERMINAL on the code.** No code-level blocker or major remains, and I do not need to see the diff again. BLOCKER-1 is a one-line version bump, not a code defect.

**Merge recommendation: approve — merge after bumping to 0.44.68** (or whatever is strictly above `main` and PyPI at that moment; they are still moving). No further review round is required for that bump.

PR #585, opened 2026-09-03
refactor(release): remove required-reviewer gate from store release workflows

damianvtran:
# PR Description

## Summary of Changes

Removes the human required-reviewer gate from the Chrome Web Store release
workflows so a store release runs end to end automatically, per an explicit
operator decision on 2026-09-03. The `chrome-web-store.yml` dispatch for
extension 0.1.7 paused at the `chrome-web-store` environment's reviewer gate;
the operator judged that approval excess and directed that both the stage and
promote environments run without a human reviewer, now and going forward.

- `extension/scripts/verify-release-environment.sh` drops the
  `required_reviewers` protection-rule check. Every other fail-closed guard
  stays: the environment must use a custom deployment branch policy, the
  policy must allow **exactly** `main`, and every release variable must be
  defined on the environment itself (repository/organization scope is
  rejected). The header comment now records why the gate was removed and what
  the retained boundary is, so a future reader does not re-add it thinking it
  was lost.
- `.github/workflows/chrome-web-store.yml` and
  `chrome-web-store-promote.yml`: comments updated to describe the new model
  honestly — the environments still scope the variables and the branch policy;
  the launch decision now rides on the reviewed-merge-to-main process plus the
  WIF ref pin. No job steps changed.
- `docs/store/submission-checklist.md`: the "Automated updates after the first
  public release" section and the setup instructions now state that no
  required reviewers are configured, name the protections that remain, and
  date the removal as a deliberate operator decision (2026-09-03).
- `extension/tests/release.test.mjs`: the verifier test previously asserted
  rejection of an environment without required reviewers; it now exercises the
  remaining environment-level rejection path (a non-custom deployment branch
  policy) and still asserts acceptance of the exact configuration.

## Security model — what changed and what did not

The real boundary was never the reviewer gate; it is the GCP Workload Identity
Federation attribute condition, which mints a token only for the pinned
numeric repository ID (`922327641`) on `refs/heads/main`. That condition, the
exact-main deployment branch policy, the environment-scoped variable check,
and each workflow's own `git merge-base --is-ancestor HEAD origin/main` check
are all unchanged. Only reviewed code on main can reach the token-bearing
upload client; the reviewer was a human-decision layer on top, and that layer
is what this PR removes.

## Related Issue(s)

- None (operator-directed change; no issue filed).

## Impact

- No breaking changes. No dependency updates. No changes to the extension
  manifest, permissions, or store package contents.
- Security implication: deliberate and scoped — the human approval pause is
[...]

PR #589, opened 2026-09-03
fix(release): make store environment verification reachable from the workflow token

damianvtran:
# PR Description

## Summary of Changes

The first real unattended Chrome Web Store release dispatch
([run 33793517588](https://github.com/damianvtran/local-operator/actions/runs/33793517588),
`version=0.1.7`) confirmed that #585's gate removal works — the run started
`in_progress` with no approval pause — and then **failed at
`Verify protected environment configuration`** roughly one curl round trip in:

```
curl: (22) The requested URL returned error: 403
##[error]Process completed with exit code 22.
```

Every later step (`Authenticate`, `Build and validate`, `Upload and submit`) was
skipped, so nothing reached the store. This is a **latent bug the reviewer gate
had been masking**: the step had never executed before, because the run always
paused for approval first.

- `extension/scripts/verify-release-environment.sh` no longer calls the
  unreachable environment-variables endpoint. It gains a second mode,
  `--assert-unscoped-empty`, and reports non-200 responses with the status, body
  and `x-accepted-github-permissions` header.
- `.github/workflows/chrome-web-store.yml` and `chrome-web-store-promote.yml`
  gain a `preflight` job that runs deliberately **outside** the environment;
  `stage`/`promote` now `needs: preflight`. The misleading `actions: read`
  comment is corrected in both.
- `extension/tests/release.test.mjs`: the mock server is now path-exact and
  auth-aware, with coverage for 403, rejected tokens, and both halves of the
  variable-scope contract. Round 1 added assertions on the **workflow job
  topology** (which is where the boundary now lives), split the branch-policy
  fixture so each condition is pinned individually, and added a tag named `main`
  and the glob `main*`. 85 → 97 tests.
- `.github/workflows/extension.yml`: gated on the two store workflow paths, so a
  YAML-only edit actually runs the suite that guards the topology.
- `docs/store/submission-checklist.md`: documents the preflight topology and why
  it must not be removed.
- `pyproject.toml`: 0.44.67 (patch). `0.44.65` was taken by #591 and `0.44.66` by
  #592 while this PR was in review, so the rebase onto current `main` required a
  new number.

`extension/manifest.json` and the extension permissions are **untouched** — the
0.1.7 store package stays byte-identical to what QA validated.

## Root cause — confirmed, not inferred

**Correction (round 1).** An earlier version of this description — and my own
brief — said the 403 came from the **first** API call. That is wrong, and both
reviewers caught it. It was the **third**: the old script called
`environments/{env}` (**200**) → `.../deployment-branch-policies` (**200**) →
`.../variables` (**403**). The ~0.6s step duration (18:57:04.63 → 18:57:05.277)
is three fast calls, not one. This matters for the next reader's map of the
incident: the environment endpoint is *not* unreachable, and is still called by
the fixed script. The source comment that repeated this error is corrected in
`f47c5547`.

The working hypothesis (that `actions: read` is simply the wrong scope and some
other scope is right) turned out to be **half right in a way that matters**. I
probed the real workflow token from a temporary diagnostic workflow on this
branch, reading the `x-accepted-github-permissions` header GitHub returns:

| Endpoint | `actions: read` | `deployments: read` | both | `write-all` | Accepted permission |
|---|---|---|---|---|---|
| `GET environments/{env}` | **200** | 200 | 200 | 200 | `actions=read` |
| `GET .../deployment-branch-policies` | **200** | 200 | 200 | 200 | `actions=read` |
| `GET .../variables` | **403** | 403 | 403 | **403** | `environments=read` |
| `GET actions/variables` (repo scope) | — | — | — | **403** | `actions_variables=read` |

Two conclusions, both load-bearing:

1. **`actions: read` was correct for two of the three endpoints** and is not the
   bug for them. The comment claiming it covers all three was wrong.
2. **The variables endpoint cannot be fixed with any scope list.** It requires
   `environments=read`, which is **not in the fixed set of permissions a workflow
   may grant** (`actions`, `artifact-metadata`, `attestations`, `checks`,
   `code-quality`, `contents`, `deployments`, `discussions`, `id-token`,
   `issues`, `packages`, `pages`, `pull-requests`, `security-events`,
   `statuses`, `vulnerability-alerts`). A probe with `permissions: write-all` —
   the upper bound of a workflow token — **still gets 403**. The repository-scoped
   variables endpoint is equally unreachable.

This confirms your instinct that `administration`-style scopes may not be
grantable at all: `environments` is exactly such a permission. **The design had
to change, not the scope list.** (The docs' own permission tables agree: the
variables endpoints sit under *Repository permissions for "Environments"*, while
the environment/branch-policy endpoints sit under *"Actions"*. A classic PAT
reads all three, which is why this never showed up in local testing.)

## The fix, and the trade-off

Deleting the variable check was not acceptable — those guards are the primary
boundary now that the human gate is gone. Instead the invariant is proved from
**what the runtime can actually observe**, as a two-job differential:

- **`preflight`** (no `environment:` key) asserts every release variable resolves
  **EMPTY**.
- **`stage`/`promote`** (inside the environment) asserts each resolves
  **non-empty**, and keeps the branch-policy checks.

A repository- or organization-scoped variable is readable by *every* job, so it
resolves non-empty in `preflight` and fails closed there. Verified empirically —
in a job with no environment, all four variables came back `EMPTY`.

**Trade-offs, stated honestly:**

- *Stronger:* it validates the value the release will actually use, rather than
  what a listing endpoint reports about it. The old check compared an API listing
  against `vars.*`; the new one checks `vars.*` resolution directly.
- *Weaker:* it can no longer distinguish "defined at repository scope" from
  "defined at organization scope" in the error message, since both present as
  "readable outside the environment". Both are rejected, so the security outcome
  is identical; only the diagnostic text is coarser.
- *Alternative rejected:* a PAT in a secret would restore the API call, but that
  trades an unreachable-but-free check for a long-lived credential with broader
  rights than the job needs — worse than what it fixes.
- The `preflight` job checks out the **reviewed workflow ref**, not `inputs.ref`,
  so a dispatch cannot supply its own verifier. `stage` still builds `inputs.ref`
  and still gates it on the merge-base ancestry check.

## Impact

- No breaking changes, no dependency updates, no manifest/permission/package
  changes. Store package unchanged and byte-identical.
- Fail-closed properties from #585 are **preserved**: custom branch policy,
  exactly-`main`, and CWS variables environment-scoped only.
- Failure diagnostics materially improved: the old `--fail-with-body` path
  printed `curl: (22) ... error: 403` naming neither the endpoint nor the
  permission, which is what made this expensive to diagnose.

## Testing Details

**Real-API matrix** (`verify-release-environment.sh` against
`api.github.com` and the live environments):

| # | Case | Expected | Actual |
|---|---|---|---|
| A | correctly-scoped token, real `chrome-web-store` | pass | `validated protected environment chrome-web-store` rc=0 ✅ |
| B | under-scoped/invalid token | fail closed | `GitHub returned HTTP 401 for environments/chrome-web-store` rc=1 ✅ |
| C | typo'd environment name | fail closed | `GitHub returned HTTP 404 for environments/chrome-web-storeXX` rc=1 ✅ |
| D | variable empty inside environment | fail closed | `CWS_PUBLISHER_ID is not defined on chrome-web-store` rc=1 ✅ |
| E | unscoped mode, all empty (correct config) | pass | `no release variable is defined at repository or organization scope` rc=0 ✅ |
| F | unscoped mode, a repo-scoped leak | fail closed | `CWS_EXTENSION_ID is readable outside the environment` rc=1 ✅ |
| G | real `chrome-web-store-production` | pass | `validated protected environment chrome-web-store-production` rc=0 ✅ |

**Live workflow-token probes** (temporary diagnostic workflow, since deleted;
touched no secrets, no environment credentials, uploaded nothing): produced the
permission table above, including the `write-all` 403 and the
environment-scoping differential (`OUTSIDE ... = EMPTY` in a job with no
`environment:` key). A probe job that *did* name the environment was refused
outright — `Branch "dev-store-verify-403" is not allowed to deploy to
chrome-web-store due to environment protection rules` — independently
confirming the main-only branch policy is live and enforcing.

**Regression discrimination:** the 9 new/changed tests were run against the
**previous** script — 9 fail (`pass 9 / fail 9`); against the new script all 91
pass. The tests fail for the right reason rather than passing vacuously.

**Gates, whole tree, exactly as CI:**

```
.venv/bin/python -m flake8 .                                  → clean
uvx --from black==26.1.0 black --check .                      → 834 files unchanged
uvx isort==5.13.2 --check .                                   → clean
.venv/bin/python -m pyright --pythonpath .venv/bin/python .   → 0 errors, 0 warnings
.venv/bin/python -m pytest tests/unit -q                      → 11757 passed, 15 skipped
cd extension && pnpm typecheck && pnpm test                   → 91 pass, 0 fail
cd extension && pnpm build:zip                                → validated v0.1.7 (12 files, no source maps)
```

### What only a live dispatch can confirm

I did not dispatch, approve or cancel any Chrome Web Store workflow, and nothing
was uploaded to the store. Remaining unproven until you dispatch:

1. **That `preflight` sees empty variables when dispatched on `main`.** Strongly
   evidenced — the probe measured all four `EMPTY` outside the environment, and
   the live API reports repository-scoped variables `total_count: 0` — but the
   `main`-dispatch path itself is unexercised.

   **Removed from this list (round 1):** *"`stage` resolves all four `vars.*`
   non-empty inside the environment"* is **not** unproven. Run 33793517588's own
   step log prints all four non-empty inside `chrome-web-store` before the curl
   failed, so that half of the differential is already confirmed against the real
   environment.
2. **Everything downstream of the verify step** — WIF authentication, the store
   upload and submission. Those have *never* executed; this PR only removes the
   blocker in front of them, and a further failure past this point is possible.

Suggested: dispatch `chrome-web-store.yml` with `version=0.1.7` after merge, and
expect `preflight` → `stage` to proceed to `Authenticate`.

## Related Issue(s)

- Follows #585. No issue filed (operator-directed).


damianvtran:
### CI notes — two unrelated failures, both resolved on re-run

Recording these in-thread rather than filing issues, per standing instructions.

**1. `test (3.12, 4)` — flake, unrelated to this diff.**
`tests/unit/session/test_resume_subagents_todos.py::test_descendant_accounting_survives_process_resume` failed with `assert [] == [('openrouter', 'routed', 0.125), ('test', 'm', None)]` — a usage-accounting row not yet visible after a process resume. This diff contains **no Python** (workflows, one shell script, one JS test file, and the version string), so it cannot reach that code path. Passed 3/3 locally with `-n0` and passed on CI re-run. Same test was last touched by #559, which fixed three flakes in this suite; this looks like a fourth of the same family, but I am not expanding scope to chase it.

**2. `cli-sanity` — upstream provider outage.**
OpenRouter returned `HTTP 429 Too Many Requests`: *"qwen/qwen3.7-flash is temporarily rate-limited upstream."* External rate limit, not a code failure. Passed on re-run after a short wait.

Both green on the current head; no code change was made for either.

damianvtran:
### Agent review — round 1

**Reviewer:** `reviewer` subagent (Claude Opus 4.6) — independent of the implementing coder; I did not write, edit, or push any of this code, and I reached the root-cause conclusion below from primary sources and the raw probe logs before reading the coder's table.
**Scope:** `b2758102660ea69fb1b5a0e04ae0cff6725d322a..00c183a9c7bf1374589422e528f796e3bcb9e680`

Judged by running, not reading: mutation-tested every retained guard, ran the new tests against the old script, replayed the coder's probe runs from the Actions API, and simulated the `preflight` invocation locally. No Chrome Web Store workflow was dispatched, approved, cancelled or re-run; nothing was uploaded; no environment configuration was touched; the worktree was restored byte-identically after every mutation (`git diff --stat` empty).

---

## 1. Root cause — independently confirmed

The claim that `environments=read` cannot be granted to a workflow token is the entire justification for redesigning rather than adding a scope, so I verified it three ways rather than taking the table on trust.

**Primary docs.** The `permissions:` key accepts a fixed set (workflow-syntax reference, "Defining access for the `GITHUB_TOKEN` scopes"): `actions`, `artifact-metadata`, `attestations`, `checks`, `code-quality`, `contents`, `deployments`, `discussions`, `id-token`, `issues`, `packages`, `pages`, `pull-requests`, `security-events`, `statuses`, `vulnerability-alerts`. Neither `environments` nor `actions_variables` appears. There is no scope to add.

**Raw probe logs** (runs `33794084593`, `33794440822` — I read the logs directly, not the PR's summary of them):

| Endpoint | `actions:read` | `deployments:read` | both | `write-all` | `x-accepted-github-permissions` |
|---|---|---|---|---|---|
| `GET environments/chrome-web-store` | 200 | 200 | 200 | 200 | `actions=read` |
| `GET .../deployment-branch-policies` | 200 | 200 | 200 | 200 | `actions=read` |
| `GET .../variables` | **403** | **403** | **403** | **403** | `environments=read` |
| `GET actions/variables` (repo) | — | — | — | **403** | `actions_variables=read` |

**The decisive detail the table alone does not show:** in the `write-all` job the runner printed its own granted-permission list — `Actions: write, ArtifactMetadata: write, … VulnerabilityAlerts: read`. There is **no `Environments` entry at all**. The upper bound of a workflow token does not include the permission, which is why `write-all` still 403s. The root cause is correct and the redesign was necessary, not merely convenient. **Not a blocker; the simpler fix does not exist.**

## 2. Does the differential prove the invariant?

Largely yes, and the direction that matters most is sound: an attacker who can set a repository variable **cannot** make preflight pass — setting it is exactly what makes preflight fail. I worked the enumeration you asked for:

- **Defined at both environment and repo/org scope** — environment precedence masks it *inside*, but the repo-scoped copy is still visible *outside*, so preflight goes non-empty and fails closed. **Caught.** (Confirmed against GitHub's "Configuration variable precedence": the lowest level wins inside, but the broader definition remains readable to every job.)
- **Org scope but not repo scope** — visible to every job, so preflight fails closed. **Caught.**
- **Empty-string environment variable** — environment mode requires non-empty, so it fails closed. **Caught** (message wording is imprecise; see F5).
- **A dispatcher supplying their own verifier** — see §3. **Prevented.**

I also want to record a **non-finding**, because it looks like a lost guard and is not: the removed `actual_value == expected_value` comparison was near-tautological. Both sides resolved from the same environment — `vars.X` and the environment's own listing of `X` — so it could only ever catch the repo-scope-shadowing case, which preflight now catches directly. Nothing real was deleted here.

Genuine gaps, rated honestly: F3 (TOCTOU) and F4 (list drift) below.

## 3. Job ordering and trust — verified

Parsed both workflows rather than eyeballing them:

```
chrome-web-store.yml         preflight: environment=None needs=None if=None perms={'contents':'read'}
                             stage:     environment=chrome-web-store needs=['preflight'] if=None
chrome-web-store-promote.yml preflight: environment=None needs=None if=None perms={'contents':'read'}
                             promote:   environment=chrome-web-store-production needs=['preflight'] if=None
```

`needs:` is a hard dependency with **no `if:` and no `continue-on-error:` anywhere in either file** — a failed preflight skips the dependent job and cannot be made non-blocking. Both remain `workflow_dispatch`-only; concurrency stays `group: chrome-web-store-release, cancel-in-progress: false`, so stage and promote still serialize against each other. Preflight is correctly minimal at `contents: read` — notably **no `id-token`**, so the job that runs outside the environment cannot mint a WIF token.

On `inputs.ref`: preflight's bare `actions/checkout` resolves to the dispatch ref, not `inputs.ref`, so a dispatcher cannot point the verifier at their own script while building something else — the stated protection holds. Worth being precise about *why* it is airtight, since the comment slightly overstates it: the dispatch ref still controls both the workflow YAML and the verifier together, so the real backstop against a rogue branch is the environment's branch policy. Probe run `33794252670` proves that backstop is live and enforcing: `Branch "dev-store-verify-403" is not allowed to deploy to chrome-web-store due to environment protection rules`. `inputs.ref` continues to control only what `stage` builds, and remains gated by the `git merge-base --is-ancestor HEAD origin/main` check. Sound.

## 4. What was retained — mutation-tested

Broke each retained guard in turn and confirmed the suite goes red, then restored byte-identically (`git diff --stat` empty after each):

| Mutation | Result |
|---|---|
| Neuter custom-branch-policy check (`protected_branches`/`custom_branch_policies`) | **90 pass / 1 fail** ✅ |
| Neuter exactly-`main` check (`total_count`/`name`/`type`) | **90 pass / 1 fail** ✅ |
| Neuter in-environment non-empty check | **90 pass / 1 fail** ✅ |
| Neuter unscoped-empty check (mode B) | **90 pass / 1 fail** ✅ |
| Let a 403 through the status check | **89 pass / 2 fail** ✅ |
| Script drops the `Authorization` header | **84 pass / 7 fail** ✅ |

Action SHA pins are unchanged — the only `uses:` delta in either file is the added `actions/checkout@11bd7190…` (v4.2.2), the same SHA already pinned. Script mode stays `100755`. No token permission widened: top-level is still `contents: read`, `id-token: write`, `actions: read`, and the new job *narrows* to `contents: read`.

I also ran the real preflight invocation locally: correct config → `validated that no release variable is defined at repository or organization scope` rc=0; a simulated repo-scoped leak → `CWS_EXTENSION_ID is readable outside the environment` rc=1.

## 5. Tests — not tautological

Against the **old** script, 9 of the 91 fail (`pass 82 / fail 9`), matching the PR's claim exactly, and they fail for the right reason. The 403 coverage asserts on the failure *reason*, not just non-zero exit — `GitHub returned HTTP 403 for environments/chrome-web-store` **and** `x-accepted-github-permissions: environments=read` — so it would catch this precise regression. `never calls the ungrantable variables endpoint` pins the exact call list (`deepEqual(requested, [ENVIRONMENT_PATH, BRANCHES_PATH])`), which is the strongest single test in the diff. One test does not do what its name and comment claim (F2).

## 6. Version

`0.44.65` is a correct **PATCH** (a fix to release plumbing; no user-facing capability). Strictly greater than main's `0.44.64`, and free at check time: no `v0.44.65` git tag, not in `gh release list` (latest is `v0.44.64`), PyPI JSON API returns **404** (latest `0.44.64`). No stale `0.44.64` references remain anywhere in the tree. **No renumber needed.** If it is taken by merge time, the next free number is `0.44.66`.

---

## Findings

**F1 — MAJOR — the new primary boundary lives in YAML that nothing tests.**
`.github/workflows/chrome-web-store.yml:40,64` (and promote `:34,54`). This PR moves the variable-scope invariant off an API call that had test coverage and onto a *job topology* — preflight exists, has no `environment:` key, and `stage` needs it — that has **none**. Reproduced, not inferred:

- delete `needs: preflight` from `stage` → **91 pass / 0 fail**
- delete the entire `preflight` job → **91 pass / 0 fail**

Either edit silently removes the only guard against a repo/org-scoped variable steering a release, while the surviving script still prints `validated protected environment chrome-web-store`. The topology is correct at this head — I verified it — so nothing ships broken; the defect is that the boundary is now correct by luck rather than by construction. This is the same class of gap #585 explicitly closed for the script guards ("deleting the exact-main check or neutering the environment-scope check left it green"), and the reasoning applies with more force now that the human gate is gone. *Fix in one line of test:* parse both workflow YAMLs in `release.test.mjs` and assert each has a `preflight` job with no `environment` key and that the deploying job's `needs` includes it.

**F2 — MINOR — `fails closed when the token is rejected` never reaches the auth-aware mock.**
`extension/tests/release.test.mjs:338-349`. It points `GITHUB_API_URL` at `http://127.0.0.1:1`, so it fails on connection refused, not on a 401. Reproduced: `curl: (7) Failed to connect to 127.0.0.1 port 1` → `request to environments/chrome-web-store could not be completed`. Its comment — "the harness now checks Authorization, so a wrong token yields a real 401" — describes something the test does not do, and disabling the mock's auth branch entirely leaves the suite green (91/0), so no test exercises that branch deliberately. The mock's auth-awareness *is* load-bearing overall (dropping the header from the script reddens 7 tests), so this is a mislabelled test rather than a hole. *Fix:* point it at the real mock server with a wrong token and assert on the 401 path.

**F3 — MINOR — TOCTOU between preflight and the deploying job.**
A repo-scoped variable added after preflight passes and before `stage` reads `vars.*` would not be caught; the two jobs read configuration at different times. Bounded, not alarming: it requires an actor who already holds repository-admin write, and the WIF attribute condition plus the main-ancestry check still stand between them and a store upload. Worth a sentence in the script header so the next reader knows the window is known and accepted rather than overlooked.

**F4 — MINOR — the four variable names are duplicated in four places with nothing keeping them in sync.**
Both workflows list the same four names in both their preflight and verify steps. I confirmed all four lists currently match. A fifth release variable added to `stage` but not to `preflight` would simply go unverified, silently. No mechanism prevents that drift.

**F5 — NIT — an empty-string environment variable is reported as "not defined".**
`extension/scripts/verify-release-environment.sh:164` says `$name is not defined on $ENVIRONMENT_NAME` when the variable may be defined but empty. It fails closed either way, so this is only diagnostic wording — and the diff is otherwise careful about exactly this (the 404 comment explicitly refuses to assert a cause it cannot distinguish).

**F6 — NIT — `docs/store/submission-checklist.md` still describes the pre-PR model.**
It details the guard set at lines ~318-322 and 356 without mentioning the `preflight` job or the two-mode script (`grep -c preflight` → 0), even though #585 updated this same checklist when it changed the model. Someone reconfiguring the environments from this doc would not learn that the scope check now depends on a job topology they must not remove — which is also the failure mode F1 describes.

*Out of scope, not filed:* `docs/store/release-record.md:123` still claims "one required reviewer" on both environments, now false after #585. Pre-existing at `b2758102`, not introduced here. Noting it in-thread rather than opening an issue.

---

## Verdict

**Changes requested** — 0 blockers, 1 major, 3 minor, 2 nits.

The engineering here is strong and the hard part is right: the root cause is correct and independently confirmed down to the runner's own permission list, the redesign was genuinely forced rather than chosen, every retained guard survives mutation testing, the new tests fail against the old script for the right reasons, and the security property holds at this head. The removed value-comparison was not a real loss.

**Not clear to merge as-is** — on F1 alone. The invariant is enforced today, but the diff relocates the primary guard for an unattended public store release into a job topology that no test asserts, and I verified both ways it can be deleted while the suite stays green. Given the human approval gate is gone and this check is what replaced it, that guard should be held in place by a test rather than by review attention. F1 is a few lines in `release.test.mjs`; F2 is a one-line change to an existing test. F3-F6 are comment and doc touch-ups that can ride along in the same remediation commit rather than costing another round.

Once F1 carries a test, I expect round 2 to be terminal.

damianvtran:
### QA report — round 1

**Reviewer:** qa-tester subagent (Claude Opus 4.6) — independent of the implementing agent
**Scope:** `b2758102660ea69fb1b5a0e04ae0cff6725d322a..00c183a9c7bf1374589422e528f796e3bcb9e680`
**Environment:** fresh detached worktree `/tmp/lop-qa-store403` at head `00c183a9`; Darwin 25.6.0 arm64; node 24 / pnpm 11.22.0; curl 8.7.1; jq 1.7.1; actionlint + shellcheck; `~/local-operator/.venv` for the Python gates. Live GitHub API reached with a **read-only** classic PAT.

**No workflow was dispatched, approved, cancelled or re-run. No environment, variable or store artifact was created or modified. Nothing was uploaded to the Chrome Web Store. The repo was not modified — every mutation was reverted from the git blob and verified by sha256.**

**Verdict: PASS — clear to merge.** 60 of 60 executed cells behaved correctly and **nothing fails open**. Three `Q1` findings are *test-coverage* gaps, not defects in shipped behaviour: I confirmed by direct execution that the shipped script rejects every corresponding attack. They are recorded here as the deferral record rather than as merge blockers.

---

## 1. The original failure, reproduced and fixed

A workflow `GITHUB_TOKEN` cannot be minted locally, so I **substituted** a replica API that returns the exact statuses and headers measured in your own probe runs (33794084593 / 33794440822): `environments/{env}` → 200 `actions=read`, `.../deployment-branch-policies` → 200 `actions=read`, `.../variables` → **403 `environments=read`**.

| | Result |
|---|---|
| **Old script** (`b2758102`) | `curl: (22) The requested URL returned error: 403` — **exit 22**, byte-identical to run 33793517588 |
| **New script** (head) | `validated protected environment chrome-web-store (main-only, environment-scoped variables)` — **exit 0**; `/variables` never called |
| **New script forced onto a 403** | prints `GitHub returned HTTP 403 for …`, then `response body: {"message":"Resource not accessible by integration",…}`, then `x-accepted-github-permissions: environments=read`, then an actionable failure line |

The diagnosability regression that cost the failed run is genuinely fixed: **status, body and the permissions header are all printed.**

I re-derived the root-cause table independently from the probe logs rather than trusting the PR body. It holds, including the decisive cell: the `write-all` job's own `GITHUB_TOKEN Permissions` block contains **no `Environments` entry at all**, and still 403s. No `permissions:` block could ever have fixed this.

## 2. Fail-closed matrix — 22/22, zero fail-open

Run against an independent mock I wrote from the API contract (not adapted from the PR's harness). **The positive control ran first** so a blanket-reject script could not fake the table.

Positive control ✅ · missing environment (404) ✅ · no custom branch policy ✅ · `protected_branches: true` ✅ · >1 branch allowed ✅ · exactly one wrong branch ✅ · a **tag** named `main` ✅ · glob `main*` ✅ · variable missing ✅ · malformed/non-JSON (both endpoints) ✅ · 403 (both endpoints) ✅ · 500 ✅ · 401 ✅ · network failure ✅ · missing token ✅ · missing repository ✅ · empty token ✅ · empty env name ✅ · odd arity ✅ · no pairs ✅

Every rejection exits nonzero **with a guard-specific, actionable message**.

**Live, read-only:** both real environments validate (`chrome-web-store`, `chrome-web-store-production` → rc=0); a typo'd name 404s; an invalid token 401s. An independent API audit confirms the invariant this PR protects is currently true in production config: both environments are `protected_branches:false / custom_branch_policies:true` with exactly `[{main, branch}]`, all four variables are environment-scoped, and **repository-scoped variables are `total_count: 0`**.

## 3. The two-job differential

- **Simulated the repo-scoped leak end to end** (not just reasoned about it), modelling `vars.*` exactly as the runner resolves it. Clean → preflight rc=0 ⇒ `stage` runs. Leak → preflight rc=1 `CWS_EXTENSION_ID is readable outside the environment …` ⇒ **`stage` is skipped, nothing authenticates to GCP, nothing uploads.** I did not touch your live variable configuration.
- Each of the four variables is caught individually, including when **only the last** leaks (no loop-exhaustion bug), and a whitespace-only value is correctly treated as a leak.
- `preflight.environment` is `None` in **both** workflows (parsed YAML, not grep).
- `stage.needs = preflight` with **no `if:` and no `continue-on-error`** ⇒ a failed preflight cannot be bypassed.
- The preflight checkout has **no `with.ref`**, so it takes the reviewed workflow ref; only `stage` uses `${{ inputs.ref }}`, and it keeps its `git merge-base --is-ancestor` gate. **A dispatch cannot supply its own verifier.**
- Mode B needs no token — correct, and matches the documented intent.
- The promote workflow is at full parity and verifies `chrome-web-store-production`.

## 4. Tests, and what mutation testing found

`pnpm test` → **`tests 91 / pass 91 / fail 0`**, matching the claim. The 9 new/changed cells genuinely discriminate: the **old** script scores `pass 11 / fail 8` against my matrix.

The mock's two new properties are real, verified by mutating the *script* until it misbehaves: reintroducing the `/variables` call ⇒ **7 failures**; sending a wrong `Authorization` header ⇒ **7 failures**; a typo'd endpoint path ⇒ **7 failures**.

I then deleted each guard in turn and re-ran the suite. **Every pristine row rejects correctly — the shipped script never fails open.** Guards M3, M4, M6, M7, M8, M10, M11 are each individually pinned (removing one reds the suite). Five mutants survived, and I re-probed each against my own matrix to separate a defect from a coverage gap:

| Guard removed | Suite catches? | Script still fails closed? |
|---|---|---|
| `protected_branches == false` | ❌ no | **no** — mutant accepts `{protected:true, custom:true}` rc=0 |
| `custom_branch_policies == true` | ❌ no | **no** — mutant accepts `{both false}` (no branch restriction) rc=0 |
| `branch_policies[0].type == "branch"` | ❌ no | **no** — mutant accepts a **tag** named `main` rc=0 |
| invalid-JSON check | ❌ no | yes (downstream jq still rejects) |
| response-body print | ❌ no | yes (diagnostic only) |

After every mutation the file was restored from the git blob and verified: sha256 `4df75cd0db137f91ccaeb208451dadd6846656fed1b56d7808928d0078140e85` = blob `d8e535ea`, `git status --porcelain` empty.

## 5. Store package unaffected ✅

Built the zip from **this branch and from `origin/main`**: both `validated Chrome Web Store package v0.1.7 (12 files, no source maps)`, `diff -r` clean, aggregate sha256 `c74e18bf123c9efa…` **identical on both sides**. `manifest.json`, `package.json`, `build.mjs` and `validate-store-zip.sh` are sha256-identical across the two trees. The manifest differs from published 0.1.5 **only in the version line**, with the same **eight** permissions and `<all_urls>`. The 0.1.7 package remains byte-identical to what was already validated.

## 6. Workflow integrity + full CI gates ✅

Both workflows parse; `on:` is `workflow_dispatch` only; `concurrency: chrome-web-store-release / cancel-in-progress: false` intact in both. **actionlint** exit 0. **shellcheck -S style** exit 0.

```
flake8 .                                    → clean
black==26.1.0 --check .                     → 834 files unchanged
isort==5.13.2 --check-only .                → clean
pyright                                     → 0 errors, 0 warnings, 0 informations
pytest tests/unit -q                        → 11757 passed, 15 skipped (19m33s)
```

No flakes, despite host load average peaking at 84 during the run. CI on head is 15/15 SUCCESS.

## 7. Version ✅

`pyproject.toml` = `0.44.65`; PyPI latest is `0.44.64` and `0.44.65` is **free**. Patch is the correct bump — a release-workflow fix with no new user-facing surface.

---

## Findings

**Q1 (major — test coverage, not shipped behaviour). Three retained guards are not individually pinned, in the file whose comment promises they are.**
`release.test.mjs:180` states *"Every guard that REMAINS is asserted here one mutation at a time… a suite that only covers the accept path plus one rejection stays green when a retained guard is deleted."* That is currently not true of three guards, and these are exactly the fail-closed guards the removed human gate used to backstop.

Root cause is precise and small: the branch-policy fixture violates **both** conditions of an `and` simultaneously —
`{ protected_branches: true, custom_branch_policies: false }` — so either guard alone still rejects it, and neither is pinned. Likewise no fixture ever supplies `type != "branch"`, so the tag-vs-branch guard is unexercised.

Minimal repro:
```
perl -0pi -e 's/and \.deployment_branch_policy\.custom_branch_policies == true//' \
  extension/scripts/verify-release-environment.sh
cd extension && pnpm test     # 91 pass, 0 fail — the deleted guard is invisible
```
…while the mutant then accepts an environment with **no deployment-branch restriction at all**.

Suggested fix (three one-line cases, no production code change): perturb one field at a time —
`{protected_branches: true, custom_branch_policies: true}`, `{protected_branches: false, custom_branch_policies: false}`, and `branch_policies: [{name:"main", type:"tag"}]`.

**Q2 (minor — accuracy). The "first API call" claim is wrong in the PR body and in a shipped comment.**
`verify-release-environment.sh:31` says the 403 *"broke run 33793517588 at the first API call."* It was the **third**: my reproduction shows the old script calling `environments/{env}` (200) → `deployment-branch-policies` (200) → `variables` (**403**), and your own probe confirms `environments/{env}` returns 200 under `actions: read`. The 0.6s step timing is consistent with three fast calls, not one. Nothing about the fix depends on this, but the comment is the next agent's map of the incident, so it should say "at the third API call — the environment-variables endpoint".

**Q3 (nit). Two small residues.** The invalid-JSON guard and the `response body:` diagnostic print are unpinned by the suite (both still behave correctly). `VALID_VARIABLES` at `release.test.mjs:190` is now **dead** — nothing references it after the variables endpoint was dropped.

**Q4 (nit — pre-existing, out of scope).** The verifier's `curl` calls carry no `--max-time`/`--connect-timeout`, and neither workflow job sets `timeout-minutes`. A hung API call would park the release job on the default 6-hour limit while holding the `chrome-web-store-release` concurrency group. Pre-existing on `main`, not introduced here; flagging only so it is on the record.

---

## What only a live dispatch can confirm

1. **`preflight` resolving all four `vars.*` EMPTY on a `main` dispatch.** Strongly evidenced — your probe job measured all four `EMPTY` outside the environment, and the live API reports repo-scoped variables `total_count: 0` — but the `main`-dispatch path itself is unexercised.
2. **Everything downstream of verify** — WIF authentication, upload, submission. These have **never** executed; this PR only removes the blocker in front of them, so a further failure past this point remains possible.

One correction to the PR's own list: **"`stage` resolves all four variables non-empty inside the environment" is no longer unproven.** Run 33793517588's step log prints all four non-empty inside `chrome-web-store` before the curl failed. That half of the differential is already confirmed against the real environment.

Full matrix: `/tmp/qa589/QA-MATRIX-PR589.md`.

damianvtran:

**Implementer:** `coder` subagent (Claude Opus 4.6)
**Remediation commit:** `f47c5547` (`test(release): pin the store release guards that mutation testing found unheld`)
**Head:** `f47c554786964f64a83ae6057d750b1dc4296300` — rebased onto `6437caf6` (main moved three commits during review)
**Scope of this round:** all findings from [Agent review round 1](https://github.com/damianvtran/local-operator/pull/589#issuecomment-5531562075) and [QA report round 1](https://github.com/damianvtran/local-operator/pull/589#issuecomment-5531808249), batched into one commit as requested.

Thank you both — the F1/Q1 pair was the right call. The invariant held at the old head, but it held by review attention, and you each independently deleted a guard that left the suite green. **No production behaviour changed in this round**; it is tests, comments and docs, plus one CI path gate.

---

## Per-finding table

| ID | Sev | Disposition | Where |
|---|---|---|---|
| **F1** | MAJOR | **Fixed** — 6 tests assert the job topology in both workflows | `release.test.mjs` (new block), `extension.yml` |
| **Q1** | major | **Fixed** — branch-policy fixture split into 3 one-mutation cases; tag + glob attacks added | `release.test.mjs:267-311` |
| **F2** | MINOR | **Fixed** — token test now reaches the mock and asserts the 401 path | `release.test.mjs:338` |
| **F3** | MINOR | **Fixed (documented)** — TOCTOU window recorded as known and accepted | `verify-release-environment.sh` header |
| **F4** | MINOR | **Fixed (tested, not just documented)** — a test compares the two name lists | `release.test.mjs`, script header |
| **F5** | NIT | **Fixed** — message no longer asserts a cause it cannot distinguish | `verify-release-environment.sh:174` |
| **F6** | NIT | **Fixed** — checklist documents the preflight topology and why it must not be removed | `docs/store/submission-checklist.md` |
| **Q2** | minor | **Fixed** — "first API call" → **third**, in the source comment *and* the PR body | script header, PR description |
| **Q3** | nit | **Partly fixed** — dead `VALID_VARIABLES` removed; invalid-JSON/body-print left unpinned, see below |
| **Q4** | nit | **Deferred** — pre-existing curl/job timeouts, see below |
| — | — | **Corrected in your favour** — `stage` resolving vars non-empty removed from the "unproven" list | PR description |

---

## F1 + Q1 — the one that mattered

The boundary moved from a tested API call into an untested job topology. It is now tested. **Six new cases** parse both workflow YAMLs and assert: a `preflight` job exists, it has **no `environment:` key**, it runs `--assert-unscoped-empty`, it checks out the reviewed ref (not `inputs.ref`), and the deploying job `needs:` it with no `if:`/`continue-on-error:` escape.

A **dependency-free reader** parses the YAML deliberately — the extension package ships no YAML parser, and adding one to devDependencies to assert four properties is a heavier change to the store package's toolchain than the assertions are worth. It strips comments and blank lines first, so a comment mentioning `environment:` cannot be mistaken for the key.

**Re-ran the full mutation matrix on the rebased tree.** Every mutant you two found surviving is now killed; baseline pristine is `97 pass / 0 fail`:

| Mutation | Before | Now |
|---|---|---|
| T1 delete `needs: preflight` (stage) | 91/0 green ❌ | **96 pass / 1 fail** ✅ |
| T2 delete the whole `preflight` job | 91/0 green ❌ | **94 pass / 3 fail** ✅ |
| T3 give `preflight` an `environment:` | untested | **96 pass / 1 fail** ✅ |
| T4 `continue-on-error: true` on stage | untested | **96 pass / 1 fail** ✅ |
| T5 point preflight checkout at `inputs.ref` | untested | **96 pass / 1 fail** ✅ |
| T6 delete `needs: preflight` (**promote**) | untested | **96 pass / 1 fail** ✅ |
| S1 drop `protected_branches == false` | survived ❌ | **96 pass / 1 fail** ✅ |
| S2 drop `custom_branch_policies == true` | survived ❌ | **96 pass / 1 fail** ✅ |
| S3 drop `type == "branch"` (tag named `main`) | survived ❌ | **96 pass / 1 fail** ✅ |
| S4 drop `name == "main"` | pinned | **96 pass / 1 fail** ✅ |
| S5 drop `total_count == 1` | pinned | **96 pass / 1 fail** ✅ |
| F4 add a 5th var to `stage` only | untested | **96 pass / 1 fail** ✅ |

All three files restored byte-identically after each mutation (`diff -q` clean against pre-mutation copies; `git status` shows only my intended edits).

Q1's root cause was exactly as diagnosed: the fixture set `{protected: true, custom: false}`, violating both sides of the `and`, so either half rejected it alone. Now three cases perturb one field each — `{true, true}`, `{false, false}`, and `deployment_branch_policy: null` (GitHub's "every branch may deploy"). Plus your two attacks: **a tag named `main`** and **the glob `main*`**.

**One thing your finding implied that I want to flag as fixed too:** `extension.yml` was paths-gated on `extension/**`, so a YAML-only edit deleting `needs: preflight` **would not have run the new tests at all**. Verified before the fix (`.github/workflows/chrome-web-store.yml` matched no trigger path). The two store workflows are now in that path list, so the guard actually guards. This is why T1–T6 above are meaningful rather than theatrical.

## F2 — verified it now exercises the branch

Points at the real mock with `token: "wrong-token"` and asserts `GitHub returned HTTP 401` **and** `Bad credentials`. Proved it is not vacuous: disabling the mock's auth branch (`if (false)`) turns the test **red** (`pass 0 / fail 1`); previously that same mutation left the suite fully green.

## Q3 / Q4 — dispositions

- **Q3, dead fixture:** fixed, `VALID_VARIABLES` removed.
- **Q3, invalid-JSON guard and `response body:` print — `deferred — diagnostic-only, and both still behave correctly`.** Neither can fail open: the invalid-JSON case is caught downstream by `jq` regardless, and the body print is diagnostic output. Pinning them means asserting on a diagnostic string, which tends to make the suite brittle against message wording without adding a security property. Recorded here rather than filed.
- **Q4, missing `--max-time`/`timeout-minutes` — `deferred — pre-existing on main, outside this PR's scope`.** Real (a hung call parks the job on the 6-hour default while holding the `chrome-web-store-release` concurrency group), but it predates `b2758102` and is a change to release-job behaviour rather than to the boundary this PR fixes. Naming it in my final report's "not addressed" list. No issue filed, per standing instructions.
- **Reviewer's out-of-scope note** (`docs/store/release-record.md:123` still claims "one required reviewer", false after #585) — **`deferred — pre-existing, introduced by #585 not by this PR`**. Same disposition; noted so it is on the record.

## Two factual corrections

1. **"First API call" was wrong — it was the third.** Corrected in the shipped source comment and in the PR description. The header now states that `environments/{env}` and `deployment-branch-policies` return **200** under `actions: read` and are still called, that only `.../variables` 403s, and that the ~0.6s duration is three fast calls — so the next reader does not conclude the environment endpoint is unreachable.
2. **In this PR's favour:** `stage` resolving all four `vars.*` non-empty inside the environment is **no longer listed as unproven** — run 33793517588's step log already prints all four non-empty inside `chrome-web-store`. The "only a live dispatch can confirm" list now names only the `main`-dispatch preflight path and everything downstream of verify (WIF auth, upload, submission), which has never executed.

## Version

`0.44.65` was taken by #591 and `0.44.66` by #592 while this PR was in review, so the branch was **rebased onto `6437caf6`** and the version resolved to **`0.44.67`** — the only conflict in the rebase was the version line. Verified free at this moment against `git tag`, `gh release list` (latest `v0.44.66`), and the PyPI JSON API (`0.44.67` → not present). Operator sets the final number at merge.

## Gates — re-run in full on the rebased tree

```
flake8 .                                      → clean
black==26.1.0 --check .                       → 834 files unchanged
isort==5.13.2 --check .                       → clean
pyright                                       → 0 errors, 0 warnings, 0 informations
pytest tests/unit -q                          → 11758 passed, 15 skipped (7m19s)
extension: pnpm typecheck                     → clean
extension: pnpm test                          → 97 tests, 97 pass, 0 fail
actionlint (the 3 touched workflows)          → exit 0
shellcheck -S style verify-release-...sh      → exit 0
build:zip                                     → validated v0.1.7 (12 files, no source maps)
```

**CI on `f47c5547`: 15/15 pass, first attempt, no flakes.** The **Browser extension** workflow ran on this head — which is itself the proof the new path gate works — and its log shows all six topology tests executing and passing.

**Real-path execution** (live API, read-only): `chrome-web-store` → rc=0; `chrome-web-store-production` → rc=0; empty variable → `CWS_PUBLISHER_ID is empty or not defined on chrome-web-store …` rc=1; unscoped clean → rc=0; unscoped leak → `CWS_EXTENSION_ID is readable outside the environment …` rc=1.

**Store package:** `extension/manifest.json`, `package.json` and `build.mjs` are untouched versus `origin/main` (0 files differ). The 0.1.7 package remains byte-identical to what QA validated.

**No Chrome Web Store workflow was dispatched, approved, cancelled or re-run; nothing was uploaded; no environment or variable configuration was touched.**

damianvtran:
### Agent review — round 2

**Reviewer:** `reviewer` subagent (Claude Opus 4.5) — independent of the implementing coder. I did not write, edit, or push any of this code, and I did not author the round-1 review either; I re-derived every claim below by executing it. Reviewer identity is asserted here because we share one GitHub account and the comment author proves nothing.
**Scope:** `b81257fd..f47c554786964f64a83ae6057d750b1dc4296300` (the remediation diff — 5 files, +230/−24). True PR base is `6437caf6`; I did not re-audit the round-1 files that this round left unchanged.

Round 1's design is approved and I did not re-litigate it. I ran my own mutation harness rather than reading the coder's matrix: **12/12 claimed mutants independently killed**, plus 12 further mutations the matrix did not enumerate. Baseline `97 pass / 0 fail` (24 in `release.test.mjs`). No Chrome Web Store workflow was dispatched, approved, cancelled or re-run; nothing was uploaded; no environment or variable configuration was touched. Every mutation was restored from the git blob and verified by sha256 — worktree ends clean, `git status --porcelain` empty, script mode still `100755`.

---

## 1. Round-1 findings — all resolved

| ID | Sev | Claimed | Independently verified |
|---|---|---|---|
| **F1** | MAJOR | Fixed | ✅ **Resolved.** T1/T2 both kill now, on the right tests. |
| **Q1** | major | Fixed | ✅ **Resolved.** S1/S2/S3 each kill individually; fixture correctly perturbs one field per case. |
| **F2** | MINOR | Fixed | ✅ **Resolved, and non-vacuous** — see §3. |
| **F3** | MINOR | Documented | ✅ TOCTOU window recorded in the script header with the bound on exploitability. |
| **F4** | MINOR | Tested | ⚠️ **Partly.** The test pins the `env:` blocks, not the argument lists — see **F7**. |
| **F5** | NIT | Fixed | ✅ `verify-release-environment.sh:187` now says "is empty or not defined". |
| **F6** | NIT | Fixed | ✅ `submission-checklist.md` documents the topology, 4 `preflight` mentions, and says why it must not be removed. |
| **Q2** | minor | Fixed | ✅ **Resolved in both places** — see §4. |
| **Q3** | nit | Partly | ✅ `VALID_VARIABLES` gone (0 refs); rest deferred in-thread. |
| **Q4** | nit | Deferred | ✅ Genuinely deferred, not silently done: 0 `--max-time`, 0 `timeout-minutes`. Matches the declaration. |

All three deferrals are in the thread verbatim (`deferred — diagnostic-only…`, `deferred — pre-existing on main…`, `deferred — pre-existing, introduced by #585…`), including `docs/store/release-record.md:123`, which I confirmed still reads "one required reviewer" — correctly left alone as pre-existing.

## 2. The 12-mutant matrix — independently reproduced, 12/12 killed

My own harness, my own mutation code, checked against the reported failing test name rather than just a red exit:

| # | Mutation | Result | Killed by |
|---|---|---|---|
| T1 | delete `needs: preflight` (stage) | 23/1 ✅ | `blocks stage on preflight` |
| T2 | delete the whole `preflight` job | 21/**3** ✅ | `keeps the preflight job` + 2 |
| T3 | give `preflight` an `environment:` | 23/1 ✅ | `keeps the preflight job` |
| T4 | `continue-on-error: true` on stage | 23/1 ✅ | `blocks stage on preflight` |
| T5 | preflight checkout at `inputs.ref` | 23/1 ✅ | `keeps the preflight job` |
| T6 | delete `needs: preflight` (**promote**) | 23/1 ✅ | `blocks promote on preflight` |
| S1 | drop `protected_branches == false` | 23/1 ✅ | `requires a custom deployment branch policy` |
| S2 | drop `custom_branch_policies == true` | 23/1 ✅ | same |
| S3 | drop `type == "branch"` | 23/1 ✅ | `requires exactly the main branch` |
| S4 | drop `name == "main"` | 23/1 ✅ | same |
| S5 | drop `total_count == 1` | 23/1 ✅ | same |
| F4 | 5th var in `stage`'s `env:` only | 23/1 ✅ | `checks the same variables` |

The matrix is honest. I also confirmed **E5** — refactoring to `needs: [preflight]` (the legitimate YAML list form) stays green — so the assertion uses `/\bpreflight\b/` and is not brittle against a valid refactor. And **E4**: the nested `environment:\n  name:` form is *also* caught, which the dependency-free parser could plausibly have missed. Good.

## 3. F2 is not vacuous — proved by mutation

Disabling the mock's auth branch (`if (request.headers.authorization !== …)` → `if (false)`) turns the suite **red on exactly that test** (`23/1`, `fails closed when the token is rejected`). Previously that same mutation left the suite fully green. The test now genuinely reaches the 401 path.

## 4. Q2 wording cannot mislead

Source header (`verify-release-environment.sh:33-38`) and PR body (line ~50) both say **third**, and both go further than the correction required: *"The first two (`environments/{env}` and its `deployment-branch-policies`) return 200 under `actions: read` and are still made below; only the variables call 403s… do not read that failure as the environment endpoint being unreachable, because it is not."* A reader cannot come away thinking the environments endpoint is unreachable. Confirmed 0 remaining occurrences of "first API call" anywhere in the tree.

## 5. The second-order property — CI really would catch a YAML-only disarm

This was the right catch, and it holds:

- The paths list is **correct and complete**. `push` and `pull_request` lists are byte-identical to each other, and the two store workflow paths are literal exact filenames. I enumerated every file `release.test.mjs` reads (`../manifest.json`, `../package.json`) — all already covered by `extension/**`, so no third gap remains.
- **Proved live, not inferred:** run `33807967816` (`Browser extension`, `pull_request`, success) fired on `f47c5547` — a commit whose diff includes `.github/workflows/chrome-web-store.yml` — and its log shows all six topology tests executing: `✔ chrome-web-store.yml keeps the preflight job…`, `✔ …blocks stage on preflight`, `✔ …checks the same variables`, and the three promote equivalents, at `tests 97 / pass 97 / fail 0`.
- **No unintended side effect.** The added paths are two exact filenames, so no unrelated change newly triggers the job. And the topology cannot change a required check: the `Main Protection` ruleset carries only `non_fast_forward` and `pull_request` rules — **no `required_status_checks` rule at all** — so adding paths alters no merge gating.

## 6. Retained security properties — executed, not read

Against my own replica API (positive control ran first, so a blanket-reject script could not fake the table):

| Case | rc | Guard |
|---|---|---|
| main / type `branch` (**positive control**) | **0** | accepts |
| **TAG named `main`** | 1 | `must allow exactly the main branch` |
| **glob `main*`** | 1 | same |
| branch `develop` · two branches · zero policies | 1 | same |
| `deployment_branch_policy: null` (every branch deploys) | 1 | `must use a custom deployment branch policy` |
| `protected_branches: true` · `custom_branch_policies: false` | 1 | same |
| empty environment-scoped variable | 1 | `is empty or not defined on chrome-web-store` |
| preflight, all four EMPTY | **0** | accepts |
| preflight, leak in first / **last** / whitespace-only | 1 | `is readable outside the environment` |

Fail-closed throughout, each with a guard-specific message. I also mutated the script to fail open in mode B (accept a non-empty unscoped value, and delete the check entirely) — **both killed** by the suite.

**Unchanged versus `main`:** top-level `permissions` blocks are semantically identical (`contents: read`, `id-token: write`, `actions: read` — comment text only); both workflows remain `workflow_dispatch`-only; `concurrency: chrome-web-store-release / cancel-in-progress: false` intact in both; the only `uses:` delta is the added `actions/checkout@11bd7190…` (v4.2.2), the SHA already pinned. Preflight stays narrowed to `contents: read` with **no `id-token`**. actionlint exit 0, shellcheck `-S style` exit 0, `pnpm typecheck` clean.

## 7. Rebase — altered nothing but the version line

`git diff 00c183a9..b81257fd` touches only `pyproject.toml` (version) plus files belonging to the three inherited main commits (#590, #591, #593 precursors) — **zero changes to any of this PR's seven files**. `extension/manifest.json`, `package.json`, `build.mjs` and `validate-store-zip.sh` are **sha256-identical to `origin/main`**. The 0.1.7 store package is exactly what QA validated.

---

## Findings

**F7 — MINOR — the F4 drift test pins the `env:` blocks, but the verifier reads its *argument list*.**
`extension/tests/release.test.mjs:~470`. The name comparison regexes `^\s+(CWS_[A-Z_]+|GCP_[A-Z_]+):\s*\$\{\{\s*vars\.` — i.e. the `env:` mapping — while the guard's actual input is the positional `NAME "$VALUE"` pairs passed to the script. Reproduced, three ways, all surviving at `24 pass / 0 fail`:

- **G1** — truncate preflight's argument list to 2 of the 4 names (leaving `env:` untouched): green. Two release variables silently stop being scope-checked.
- **G2** — pass literal `""` for all four in preflight: green. Preflight then trivially passes on *any* configuration, including a live repo-scoped leak.
- **H1** — the precise F4 scenario: add a 5th variable to **both** `env:` blocks (satisfying the new test) but to `stage`'s argument list only: green, and preflight verifiably never checks it.

So the drift F4 described is narrowed, not closed. Not a blocker — the topology, the mode, and the `env:` lists are all pinned, and shipped behaviour at this head is correct (I verified preflight's four arguments match). *Fix in one line:* extract the names from the `run:` block instead — match `^\s+(CWS_[A-Z_]+|GCP_[A-Z_]+) "\$` — and compare those lists.

**F8 — MINOR — a *step*-level disarm of preflight is unpinned; only job-level is caught.**
The two topology tests assert `if:` and `continue-on-error:` on the **job** keys (4-space indent), which is right and works (T4, and `if: always()` on stage, both die). But the same disarm one level down survives at `24/0`:

- **E1** — `continue-on-error: true` on preflight's verifier *step*
- **E3** — `if: false` on preflight's verifier *step*
- **E2** — append `|| true` to the verifier invocation in preflight's `run:` block

Each leaves preflight reporting success on a real leak while `needs: preflight` still appears intact — the same "correct by luck" class F1 was raised for, at a lower level. Lower likelihood than F1 (a step-level `continue-on-error` is a less natural edit than deleting a `needs:`), which is why this is minor rather than major. *Fix:* the tests already hold `jobs[job].lines`; assert no `continue-on-error`/`if:` at any indent inside preflight, and that its `run:` line ends with the last argument rather than a `|| …`.

**F9 — MINOR — the version is stale: `0.44.67` was taken while this round was in flight.**
`pyproject.toml` is `0.44.67`, but `origin/main` moved to `06928224` (#593) at 21:48 and **released `v0.44.67`** — git tag exists, `gh release list` shows it as Latest, and the PyPI JSON API returns **HTTP 200** for `0.44.67` (PyPI latest is now `0.44.67`). The PR head no longer sorts above main. **Next free number is `0.44.68`**: no `v0.44.68` tag, `gh release view v0.44.68` → not found, PyPI `0.44.68` → **HTTP 404**. Patch remains the correct bump — release-plumbing fix with no new user-facing capability. Given the churn (~10 releases today), re-confirm at merge time rather than trusting this line.

*Not filed as issues, per standing instructions.* No `gh issue create` was run.

---

## Verdict

**clean — 0 blockers, 0 majors, 3 minors, 0 nits. This round is TERMINAL.**

Every round-1 finding is resolved. F1 and Q1 — the two that mattered — are genuinely fixed and I confirmed it with my own harness rather than the coder's: all 12 mutants die, each on the correct assertion, and the second-order catch about `extension.yml` path gating is real, correct, complete, side-effect-free, and **proved by a live CI run on this head** rather than argued. The security properties hold under direct execution across every branch-policy attack including the tag and glob cases, and the store package is byte-identical to main.

F7 and F8 are honest residue of the same class the round-1 review named — a guard held by the shape of the file rather than by construction — but both are narrower than F1, neither is wrong at this head (I verified the shipped topology and argument lists directly), and neither justifies a third round. They are follow-ups. F9 is a merge-time mechanic, not a code defect.

**Clear to merge**, with one mandatory precondition: **bump `pyproject.toml` to `0.44.68`** (or the next free number at merge time) before merging, since `0.44.67` is now published on PyPI and tagged. That edit is version-line-only and does not invalidate this round.

damianvtran:
### QA report — round 2

**QA:** `qa-tester` subagent (Claude Opus 4.5) — independent of the implementing coder and of both reviewers. I did not write, edit, or push any of this code. Every claim below was produced by executing it on this head; I read the diff and the prior comments only to scope the round.
**Head under test:** `158caf138b0e3d6cd96147360f970296ca338630` — **verified current at start and again at the end of the run; it did not move.**
**Scope:** the delta since QA round 1 passed `00c183a9` (the F1 remediation — topology tests, one-mutation fixtures, the `extension.yml` paths filter — plus the F7/F8 + version commit), **and** a full re-run of the round-1 fail-closed security matrix, which must still hold on the head that actually merges.
**Environment:** macOS 25.6.0 arm64; isolated worktrees `/tmp/qa589r2` (head) and `/tmp/qa589r2-main` (`origin/main`); node 24 / pnpm 11.22.0; actionlint + shellcheck from Homebrew.

**No Chrome Web Store workflow was dispatched, approved, cancelled or re-run. Nothing was uploaded. No environment or variable configuration was altered** (every live API call was a read-only `GET`). No operator session or daemon was touched. The repo was not modified: every mutation was restored and verified sha256-equal to the git blob, and `git status --porcelain` is **empty**.

---

## 1. Round-1 Q findings — resolved / not resolved

| ID | Sev | Claimed | Independently verified on this head |
|---|---|---|---|
| **Q1** | major | Fixed | ✅ **Resolved.** S1–S5 each kill the suite **individually** (96/1 apiece) in my own harness — the fixture genuinely perturbs one field per case, which was the defect. |
| **Q2** | minor | Fixed | ✅ **Resolved.** `verify-release-environment.sh:33` now reads "at the **THIRD** API call. The first two…". |
| **Q3** | nit | Partly | ✅ **Matches the declaration.** `VALID_VARIABLES` → **0 references**. The invalid-JSON guard and body print remain unpinned — but I confirmed both still behave correctly (A18–A20 all reject). |
| **Q4** | nit | Deferred | ✅ **Genuinely deferred, not silently done.** 0 `--max-time`, 0 `timeout-minutes` in either store workflow. Still real, still pre-existing on `main`. |

Reviewer findings **F7 / F8** (the six round-2 survivors) and **F9** (stale version) are all resolved on this head — verified by execution in §3 and §6, not by reading the table.

## 2. Fail-closed matrix — 42 cells re-run, zero fail-open

Driven against an **independent mock I wrote from the GitHub REST contract**, not adapted from the PR's JS harness. **The positive control ran first**, so a blanket-reject script could not fake this table.

Positive control ✅ · env 404 ✅ · `deployment_branch_policy: null` ✅ · `protected:true/custom:true` ✅ · both-false ✅ · `protected:true/custom:false` ✅ · two branches ✅ · one wrong branch ✅ · **a TAG named `main`** ✅ · **glob `main*`** ✅ · glob `*` ✅ · zero policies ✅ · 403/401/500 on **both** endpoints ✅ · non-JSON on both ✅ · empty body ✅ · invalid token ✅ · each of the four variables empty **individually incl. the last** ✅ · odd arity ✅ · no pairs ✅ · missing/empty token ✅ · missing repository ✅ · empty env name ✅ · network unreachable ✅ · mode-B clean ✅ · **mode-B leak in each of the four positions individually** ✅ · whitespace leak ✅ · mode B needs no token ✅

Every rejection exits non-zero with a guard-specific, actionable message. **The two attacks the round-1 review asked for — a tag named `main` and the `main*` glob — are both rejected on this head.**

**The matrix discriminates:** the `origin/main` script scores **rc=1 on my positive control** and has no `--assert-unscoped-empty` mode at all, so these cells are testing the fix rather than passing vacuously.

**End-to-end differential** (modelling `vars.*` exactly as the runner resolves it):
```
A. correct config   preflight rc=0 → stage-verify rc=0 → proceeds to WIF auth + upload
B. CWS_EXTENSION_ID at REPOSITORY scope
                    preflight rc=1 "CWS_EXTENSION_ID is readable outside the environment"
                    >>> needs: preflight FAILED => stage SKIPPED
                    >>> No WIF auth. No upload. No store submission. FAIL-CLOSED.
```

**Live, read-only:** both environments are `protected_branches:false / custom_branch_policies:true` with exactly `[{main, branch}]` and `total_count == array length`; repo-scoped variables are `{"total_count": 0}`. Production config satisfies the invariant.

**Bug reproduced, then fixed:** the original failure (run `33793517588`) logs `curl: (22) The requested URL returned error: 403` in the *Verify protected environment configuration* step. The old script calls `environments/{env}/variables` (line 68); the new script has **0 references** to it. Probe run `33794440822`, read directly from the Actions API, shows `HTTP/2 403` with `x-accepted-github-permissions: actions_variables=read` under `write-all` — the root cause holds.

## 3. Mutation testing — 32 applied, 32 killed, 0 survivors

My own harness. Every workflow mutation **slices the YAML at the job header and edits only inside the named job**, so it cannot land on a later job — the false-survivor trap noted in the brief. Baseline **97/0**.

**All six reviewer round-2 survivors are now killed:**

| ID | Mutation (PREFLIGHT job) | Result | Killed by |
|---|---|---|---|
| **G1** | preflight args truncated to 2 of 4, `env:` untouched | 96/1 ✅ | `checks the same variables` |
| **G2** | preflight passes literal `""` for all four | 96/1 ✅ | `checks the same variables` |
| **H1** | 5th var in **both** `env:` maps, only stage's arg list | 96/1 ✅ | `checks the same variables` |
| **E1** | `continue-on-error` on preflight's verifier **step** | 96/1 ✅ | `blocks stage on preflight` |
| **E2 / E2b / E2c** | `\|\| true` / `\|\| :` / `; true` on the verifier call | 96/1 ✅ | `blocks stage on preflight` |
| **E3** | `if: false` on preflight's verifier **step** | 96/1 ✅ | `blocks stage on preflight` |

Plus G1b (3-of-4 truncation), the five topology disarms **T1–T5** and `if: always()` (**T7**), the five branch-policy guards **S1–S5** each killing individually, mode-B fail-open (**S6**), env-mode empty-accept (**S7**), status-check bypass (**S8**, 94/3), dropped `Authorization` (**S9**, 90/7) — **and every applicable mutation repeated against the promote workflow** (T6, T8, T9, E1p, E2p, E3p, G1p, G2p), all killed.

Restoration verified by sha256 against the git blob for every file; `git status --porcelain` empty; script mode still `100755`.

## 4. The second-order property — the paths gate

This is the one that would make everything else theatrical, so I checked it six ways.

- **It is load-bearing:** `extension.yml` is the **only** PR-time workflow that runs `release.test.mjs` (`chrome-web-store.yml` runs it too, but only on dispatch).
- **Correct:** both path entries name files that actually exist — no typo — and were added to **both** `push` and `pull_request`.
- **Complete:** the suite reads exactly two files outside `extension/**` — the two store workflows in `STORE_WORKFLOWS` — and nothing else. The list covers both.
- **It fires:** simulating GitHub's documented path-filter semantics, a YAML-only edit to either store workflow **TRIGGERS** on both events; against `origin/main`'s list both are **NO RUN**, so the defect was real and is closed.
- **No over-triggering:** `local_operator/tui/app.py`, `.github/workflows/ci.yml`, `pyproject.toml`, `docs/**` still do not trigger it.
- **No side effects:** `extension.yml` has a single `build` job whose steps touch only `extension/` and `browser_bridge`, both unchanged by a store-YAML edit. `main` is **not a protected branch** (`gh api …/protection` → 404), so there is no required-check list to perturb.
- **Real CI evidence on this head:** run **33812230084** (`Browser extension`, success) executed all six topology tests — `keeps the preflight job…`, `blocks stage on preflight`, `checks the same variables…` and the three promote equivalents — at `tests 97 / pass 97 / fail 0`.

## 5. Store package — byte-identical

`manifest.json`, `package.json`, `build.mjs`, `store-package-files.txt`, `validate-store-zip.sh`, `chrome-web-store.sh`: **identical blob SHAs** to `origin/main`; `extension/src` and `extension/icons` diff is empty. Built the zip from **both** branches: each validates as `v0.1.7 (12 files, no source maps)`, identical file lists, and **all 12 per-file sha256 hashes match** (manifest `7540581f…`, worker.js `d75d1f4e…`). Manifest carries the same **eight** permissions as the published v0.1.5 record — `debugger, tabs, tabGroups, scripting, storage, alarms, webNavigation, notifications` — plus host `<all_urls>`; only the version line differs. **Safe to ship immediately after merge.**

## 6. Workflow integrity, gates, version

Both workflows parse; `on:` is **`workflow_dispatch` only**; concurrency `chrome-web-store-release / cancel-in-progress:false` intact in both; `preflight` has environment=None/needs=None/if=None/continue-on-error=None and **no `id-token`** (it cannot mint a WIF token); token permissions are **narrowed, never widened** (top-level identical to main); **action SHA pins identical to main**, all five 40-char pinned; the temporary probe workflow is gone and the workflow file set matches main. **actionlint rc=0**, **shellcheck -S style rc=0**, `bash -n` clean, mode `100755`.

```
flake8            rc=0
black --check     834 files would be left unchanged
isort --check     rc=0  (with CI's pinned 5.13.2 — see Q11)
pyright           0 errors, 0 warnings
pytest tests/unit 1 failed, 11761 passed, 15 skipped (see Q12)
extension suite   tests 97 / pass 97 / fail 0
pnpm typecheck    rc=0
CI on head        14/14 checks pass
```

**Version `0.44.68`:** strictly greater than main's `0.44.67`; `git tag -l v0.44.68` → none; `gh release view v0.44.68` → not found; PyPI `0.44.68` → **HTTP 404** (`0.44.67` → 200, PyPI latest `0.44.67`). Correct **PATCH** — release plumbing, no new user-facing capability. Reviewer **F9 is resolved**. ~10 releases landed today, so re-confirm at merge time.

## 7. New findings — all nits, none blocking

**Q7 (nit — not reachable). `total_count: 1` with two entries in the array passes.** The script asserts `total_count == 1` and inspects only `branch_policies[0]`, so a payload whose count contradicts its own array is accepted. Requires GitHub to serialise a count inconsistent with the collection it just returned; the live API returns count == array length on both environments, `per_page=100` far exceeds the policy count, and adding a second policy increments the count. Not attacker-controllable. *Fix if ever touched: assert `(.branch_policies | length) == 1` too.*

**Q8 (nit — not a scope bypass). A whitespace-only value passes environment mode.** `[[ -n "$value" ]]` accepts `" "`. This is **not** a fail-open on the boundary this PR defends: mode B correctly treats whitespace as a **leak** and rejects it (A39), so a whitespace-valued *repo-scoped* variable still fails preflight closed. A whitespace value at *environment* scope is a misconfigured release variable that fails downstream at WIF/upload, not a scope escape. *Fix if ever touched: compare against `${value// /}`.*

**Q9 (nit — evidence scope, not a defect).** No commit on this branch changes a store workflow *without also* touching `extension/**`, and `pull_request` path filters evaluate the cumulative PR diff — so **no run on this PR isolates the store-workflow path as the sole trigger.** §4's live-CI evidence proves the tests execute and pass on this head; the trigger claim itself rests on GitHub's documented semantics plus the pre/post comparison. I am flagging this because the round-2 review described the property as "proved by a live CI run", which is slightly stronger than what the runs can show. It will be settled for free by the first store-YAML-only commit on `main`.

**Q10 (nit — pre-existing environmental flake, not attributable to this PR).** `tests/unit/model/test_prices.py::test_a_background_revalidation_gets_the_full_default_timeout` failed once in my full run (`assert 0 == 1`). **This PR contains no Python whatsoever** (7 files: 3 workflows, 1 shell script, 1 JS test, 1 doc, `pyproject.toml`), so it cannot reach that code path. Isolated on this head `-n0`: **3/3 pass**. Whole module under xdist: 45 passed on **both** head and `origin/main`. The test mutates module-level shared state (`catalogue._revalidating/_last_attempt/_threads`), so a sibling xdist worker races it — the contention class AGENTS.md documents, aggravated by ~40 concurrent worktrees on this host. **CI `test (3.12, 1)` passed on this exact head.** Same family as the flake the coder recorded in-thread.

**Q11 (nit — local toolchain drift, not a PR defect).** `isort --check-only` reported 3 errors locally under **isort 9.0.1**; CI pins **5.13.2**, under which the tree is **rc=0 clean**. All three files are byte-identical to `main` and untouched here.

*No issues filed — `gh issue create` was never run, per standing instructions. Q7–Q11 live in this thread as the record.*

## 8. Cells only the operator's live dispatch can confirm

Stated plainly; **not** claimed as covered:

1. **The `main`-dispatch preflight path** — preflight resolving `vars.*` to genuinely empty on a real runner. My mode-B evidence *models* that resolution; it does not observe it. This is the primary new code path.
2. **Everything downstream of verify — which has never executed, ever:** WIF/OIDC authentication, the Chrome Web Store upload, and the submission call. Since #585 removed the approval gate, the first real dispatch runs these unattended end to end.
3. **The environment job's live API path under a workflow token** — that both retained endpoints answer 200 under `actions: read` from inside `stage`. Evidenced by the PR's probe runs (which I read directly from the Actions API), not re-run by me.
4. **An isolated YAML-only CI trigger** — see Q9.

---

## Verdict

**PASS — clear to merge.**

The security property holds on the head that will actually merge: **42 fail-closed cells, nothing fails open**, positive control first, both the tag-named-`main` and `main*` glob attacks rejected. **32/32 mutants killed with zero survivors** — including all six of the reviewer's round-2 survivors, each extended to the promote workflow — with byte-identical restoration proved by sha256. The second-order paths gate is correct, complete, side-effect-free, and demonstrably runs all six topology tests in real CI on this head. The store package is byte-identical to `main` across all 12 files with permissions matching the published record, so the operator's post-merge ship is unaffected. Version `0.44.68` is the correct patch and free on all three registries.

All four round-1 Q findings are resolved or deferred exactly as declared. The five new findings are nits: two unreachable edge cases in guards that are correct where it counts, one honest narrowing of an evidence claim, and two environmental artifacts of my own machine. **No FAIL remains on this head.**

One merge-time mechanic, not a QA blocker: re-confirm `0.44.68` is still free immediately before tagging, given today's release cadence.

PR #593, opened 2026-09-03
perf(providers): cache model listings on initial paint and parallelise catalogue fetches

damianvtran:
## Summary

Optimizes `/model` picker latency and catalogue resolution so OpenRouter and Radient models show up immediately and live updates complete faster:

1. **Immediate Initial Paint with Cached Aggregators**:
   - Aggregator providers (`openrouter`, `radient`, `radient-key`) have no hardcoded models in the static registry. Previously, `static_catalogue()` returned 0 models for aggregators on the opening keystroke, leaving the user waiting for `live_catalogue()` to finish.
   - Introduced `cached_available_models(provider_id)` in `local_operator.model.discovery` using `peek_listing()` (synchronous, local disk-only, non-blocking, no network).
   - Added `initial_catalogue()` to `ProviderController` and updated `OperatorApp._populate_model_picker()` to use it. On the very first frame of `/model`, cached models from disk (e.g. 425 models from OpenRouter and Radient) paint immediately in <10ms.

2. **Parallel Live Catalogue Resolution**:
   - In `ProviderController.live_catalogue()`, fetching models across all 19 registered providers previously ran sequentially in a `for` loop with individual `asyncio.to_thread` calls.
   - Refactored to fetch all providers concurrently via `asyncio.gather`. The worst-case refresh latency is now bounded by the single slowest provider rather than the sum of all provider round trips.

3. **Aggregator & Public Listing Consistency**:
   - Added `radient-key` to `AGGREGATOR_PROVIDERS` in `local_operator.providers.registry` and `PUBLIC_LISTING_PROVIDERS` / `LISTING_CAPTURE_VERSIONS` in `local_operator.model.discovery`.

## Testing Evidence

- **First frame initial catalogue execution**:
  ```
  1. First frame initial_catalogue: 967 entries, 442 selectable rows in 7.49ms
     - openrouter     : 425 models immediately offered
     - radient        : 425 models immediately offered
  2. Parallel live_catalogue: 1037 entries in 42.11ms (cached) / 1.91s (ttl_s=0 cold parallel vs 5.42s sequential)
  ```
- **Quality Gates**:
  - `flake8`: Clean (0 errors)
  - `black --check`: Clean
  - `isort --check-only`: Clean
  - `pyright`: 0 errors, 0 warnings
  - Unit tests: All passing (1,475 tests passed across model, providers, and TUI picker suites)

PR #594, opened 2026-09-03
fix(providers): send each model family its vendor-documented sampling values

damianvtran:
## Summary

`local_operator/model/configure.py` hardcoded `DEFAULT_TEMPERATURE = 0.2` / `DEFAULT_TOP_P = 0.9` onto every `ModelSpec` and sent that pair to **every provider on every turn**. The reported bug was Gemini 3 getting stuck in output loops, but the audit found that is one visible face of a general defect: **no major vendor documents a default anywhere near 0.2** — essentially all are 1.0 — so lop was asserting an invented number over each vendor's own tuning, and on several current families over a documented hard rejection.

This replaces the two constants with a **model-keyed sampling policy table**, following the precedent already argued in these files for `_NO_SAMPLING_PARAMS` and `_EFFORT_TABLE`: a sampling contract is a property of the model, not of the route to it. That is what makes `google/gemini-3.8-flash`, `openrouter/google/gemini-3.8-flash` and `radient/google/gemini-3-pro` behave identically — a provider-keyed rule would have fixed the direct route and left every aggregator route carrying the bug.

Each row either **OMITs** the pair (letting the vendor's own default apply) or **SEEDs** a value the vendor documents that silence would not reproduce. `temperature` and `top_p` are resolved **independently**, because vendors diverge on them (Qwen documents 0.7 with top_p 0.8).

**Omission is preferred to a number** wherever a vendor documents a default: it is self-updating, it cannot drift stale, it is literally what Google's migration guide prescribes, and OpenRouter notes that an explicitly sent value "may differ from omitting it (for example, it can affect provider-side cache keys)".

## Vendor justification (all primary sources)

| Family | temp | top_p | Source |
|---|---|---|---|
| Gemini ≥3.x | OMIT | OMIT | *"For all Gemini 3 models, we strongly recommend keeping the temperature parameter at its default value of 1.0... Changing the temperature (setting it below 1.0) may lead to unexpected behavior, such as **looping** or degraded performance"* and *"we recommend **removing this parameter**"* — [ai.google.dev/gemini-api/docs/gemini-3](https://ai.google.dev/gemini-api/docs/gemini-3) |
| Gemini ≤2.5 | 1.0 | 0.95 | Honoured; `getModel` reports these as the per-model defaults |
| Anthropic ≥4.7 | OMIT | OMIT | *"Models released after Claude Opus 4.6 do not support setting temperature. A value of 1.0 will be accepted for backwards compatibility, all other values will be **rejected with a 400 error**."* — [platform.claude.com/docs/en/api/messages.md](https://platform.claude.com/docs/en/api/messages.md) |
| OpenAI gpt-5.x/6+/o-series | OMIT | OMIT | GPT-6 Astra: *"Unsupported parameters: Remove `temperature`, `top_p`, and `top_logprobs`."* |
| DeepSeek V4 | OMIT | OMIT | Thinking on by default: *"Thinking mode does not support the temperature, top_p... setting these parameters will not trigger an error but will also have no effect"* |
| Moonshot / Kimi | OMIT | OMIT | K3: *"`temperature=1.0`, `top_p=0.95` ... are fixed; omit them from requests"* |
| Alibaba / Qwen | 0.7 | 0.8 | Per-model default table |
| Z.AI / GLM | OMIT | OMIT | Defaults are **per-series** (1.0 for 5.x/4.7/4.6, 0.6 for 4.5), so no single seed is right |
| Mistral | OMIT | OMIT | Publishes no static default by design; exposes `default_model_temperature` per model card |
| xAI / Grok | OMIT | OMIT | **UNKNOWN** — typed `number \| null` with no stated default. Omission is the only honest option |
| Ollama | OMIT | OMIT | Defaults 0.8/0.9, overridden by the model's Modelfile. We cannot know what the user pulled, so we must not overrule its publisher |
| **Unknown / fallback** | OMIT | OMIT | The most consequential row: silence beats an invented 0.2 |

**Qwen judgement call:** seeds the model *defaults* (0.7/0.8), not Qwen's separate coding row (0.2). lop runs summarisation, commit messages and compaction through the same spec, and 0.2 is the exact assertion this PR exists to stop making.

**Not implemented — DeepSeek 0.0-for-coding.** An older DeepSeek page recommends temperature 0.0 for coding, but it predates V4 and thinking mode, and DeepSeek's own coding-agent benchmarks run at 1.0/0.95. That row is stale; a comment in the table warns future readers not to revive it.

## Three plumbing bugs that made per-model defaults inexpressible

These are provider-agnostic and load-bearing — without them the table is dead on arrival.

1. **`configure_model` stomped the spec** (`configure.py`). It ran an unconditional `model_copy(update={"temperature": temperature, "top_p": top_p})` from parameters defaulting to the global constants, so any per-model seed was overwritten with 0.2/0.9 on the main construction path. Now a `None` sentinel applies the override only when a caller genuinely passed one. Both real override paths (`_AGENT_SAMPLING_FIELDS`, the server's `ChatRequest.options`) already omit unset knobs. `DEFAULT_TEMPERATURE`/`DEFAULT_TOP_P` remain exported.
2. **The failover hop re-imposed the primary's values** (`failover.py`). `spec_for_target` carried `base.temperature`/`top_p` onto the target, so a hop from a 0.2-configured primary re-imposed on a Gemini target exactly the value documented to make it loop. **This was a live wire bug, not cosmetic** — it applied to any target that still sends the pair. The target's own policy is now authoritative; a non-`None` base value is ambiguous (a user's choice, or just as likely the *primary* family's seed) and propagating it is the same category of error.
3. **`ModelSpec` field defaults** (`harness/types.py`) were a second copy of 0.2/0.9. Audited: `build_model_spec` is the only production constructor, and `FrontendModelSpec` is an `extra="allow"` wire-deserialisation subclass carrying transmitted values. The fields are now `float | None` defaulting to `None`, so a directly-constructed spec inherits omission rather than an invented number.

**Also found and fixed:** `server/routes/speech.py` re-asserted `temperature=agent.temperature or 0.2`, defeating the table for that route — and `or` also silently rewrote a deliberate `0.0` into `0.2`.

## Aggregator listing narrowing

OpenRouter publishes a `supported_parameters` allowlist. Verified live: **81 of 425 models omit `temperature`** from it. That is a per-model fact the curated table cannot keep up with, so unknown models get it right for free.

It may only **narrow**, never widen. The listing answers *"will the aggregator forward this"*, not *"will the model honour it"* — `google/gemini-3.8-flash` **does** list `temperature` as supported, because OpenRouter forwards it and Google then ignores it. Silence defers to the table, mirroring the effort-ladder precedent. **Radient is UNVERIFIED** and documents no passthrough semantics, so its listing is never read as a denial.

### Real wire bodies, before vs after

Captured from the actual `_build_body` of each of the four wire clients, driven through the real `build_model_spec`. The "before" column is a **separate worktree checked out at `origin/main`**, verified to be genuinely main's code (no `_SAMPLING_POLICY`, `DEFAULT_TEMPERATURE == 0.2`) rather than asserted:

| Route | Before (main) | After |
|---|---|---|
| `alibaba/qwen3-coder-plus` | `temp=0.2, top_p=0.9` | `temp=0.7, top_p=0.8` |
| `anthropic/claude-opus-4-7` | `temp=0.2, top_p=0.9` | **both absent** |
| `anthropic/claude-opus-4-8` | `temp=0.2, top_p=0.9` | **both absent** |
| `anthropic/claude-sonnet-4-5` | `temp=0.2, top_p=0.9` | **both absent** |
| `deepseek/deepseek-v4-pro` | `temp=0.2, top_p=0.9` | **both absent** |
| `google/gemini-2.5-pro` | `temp=0.2, top_p=0.9` | `temp=1.0, top_p=0.95` |
| `google/gemini-3.8-flash` | `temp=0.2, top_p=0.9` | **both absent** |
| `kimi/kimi-k2-0711-preview` | `temp=0.2, top_p=0.9` | **both absent** |
| `mistral/mistral-large-latest` | `temp=0.2, top_p=0.9` | **both absent** |
| `ollama/qwen3:32b` | `temp=0.2, top_p=0.9` | **both absent** |
| `openai/gpt-4o` | `temp=0.2, top_p=0.9` | **both absent** |
| `openai/gpt-6-astra` | `temp=0.2, top_p=0.9` | **both absent** |
| `openrouter/google/gemini-3.8-flash` | `temp=0.2, top_p=0.9` | **both absent** |
| `openrouter/some-vendor/uncharacterised-model` | `temp=0.2, top_p=0.9` | **both absent** |
| `radient/google/gemini-3-pro` | `temp=0.2, top_p=0.9` | **both absent** |
| `xai/grok-4` | `temp=0.2, top_p=0.9` | **both absent** |
| `zai/glm-5.3` | `temp=0.2, top_p=0.9` | **both absent** |

Every one of the seventeen routes was receiving `0.2/0.9` on main.

### Proven at the socket, not just in the body builder

A local HTTP server captured what the real `OpenAICompatClient` actually transmits over a real connection:

```
google/gemini-3.8-flash    REAL SENT BODY -> temperature=ABSENT, top_p=ABSENT
google/gemini-2.5-pro      REAL SENT BODY -> temperature=1.0, top_p=0.95
alibaba/qwen3-coder-plus   REAL SENT BODY -> temperature=0.7, top_p=0.8
anthropic/claude-opus-4-7  REAL SENT BODY -> temperature=ABSENT, top_p=ABSENT
```

### Live round trip

Against a real Gemini 3.8 Flash through Radient (`chat/completions`), both the new body and the old 0.2/0.9 body returned **HTTP 200**. That is the point: this is a **silent quality bug**, which is why no error ever surfaced it.

**Which route actually carried the outage (corrected after QA round 1).** An earlier draft of this description said the Anthropic 4.7/4.8 rejection fired "on every turn on the direct route". That was too strong. On the **direct** `anthropic` route, `claude-opus-4-7` already omitted the pair on `origin/main`, because the Anthropic builder gates sampling on the same branch that writes `thinking`. The 0.2 was reaching those models through the **aggregator** routes — QA confirmed `openrouter/anthropic/claude-opus-4-7` and `radient/anthropic/claude-opus-4-8` both sent `temperature=0.2` on base. That incidental thinking-gate protection is also why the R1 override leak below read as safe on the Anthropic wire while leaking on every OpenAI-compat route.

**Honest limitation:** the credentials store has only a Radient key — no direct Anthropic/OpenAI/DeepSeek/Mistral/xAI keys. The `claude-opus-4-7` HTTP 400 is **vendor-documented, not locally reproduced**. I have not observed the 400 first-hand and am not claiming to.

### Listing narrowing, exercised against the live OpenRouter listing

- Narrowing fires: a seeded family whose allowlist denies the key went `temp 0.7 -> None`.
- No widening: `google/gemini-3.8-flash`, whose live allowlist **does** include `temperature`, stayed `None`.

### Escape hatch

Explicitly tested at both the spec and the wire: a deliberate `temperature=0.0` / `top_p=0.5` still reaches the body on families whose policy is to omit. `0.0` specifically, since it is the falsy value an `or`-style default eats.

### Gates (whole tree, as CI runs them)

```
flake8 .                          CLEAN
black --check .    (26.1.0)       834 files unchanged
isort --check-only (5.13.2)       CLEAN
pyright                           0 errors, 0 warnings
pytest tests/unit                 11852 passed, 15 skipped (414s)
```

**Note for whoever picks this up:** pyright needs `--pythonpath .venv/bin/python` here, or it resolves the wrong interpreter and reports ~1089 phantom `reportMissingImports`.

## Preserved invariants

- `("google", "gemini-2.5-flash-thinking")` still keeps its sampling — the `reasoning` markers do not divert it.
- Kimi mainland (`kimi-k2-0711-preview`, `moonshot-v1-128k`) and the coding-host rows unchanged.
- Claude 5+, o-series and gpt-5 still hard-omit via `_NO_SAMPLING_PARAMS`.
- Snapshot dates (`gemini-2.5-pro-20250314`) still cannot read as generation numbers.

`supports_sampling_params` (hard 400 rejection, strips even an explicit override) is kept **distinct** from the policy `None` (we simply stop asserting; a user override still rides). Gemini 3 *ignores* rather than rejects, so it belongs in the second category — which is what preserves the escape hatch.

## ⚠️ Environment trap for QA

The worktree I was handed had a **symlinked `.venv` whose editable finder pointed at `/private/tmp/lop-attach-tui`, a deleted worktree**. Per `AGENTS.md` §"Every feature worktree owns its own venv", this silently splits behaviour: `pytest` (cwd-first) reads the branch, while anything run as the product resolves `~/local-operator` — i.e. `main`. I rebuilt a real venv before trusting any result. **Verify `.venv/bin/python -c "import local_operator; print(local_operator.__file__)"` points into your own checkout before running anything.**

## Round 1 remediation (head `1d6e5fe6`)

Rebased onto current `origin/main` (`06928224`, #593) — the previous `0.44.67` bump collided with the version #593 already released, so this now bumps to **0.44.68**. `git merge-tree` reported a clean merge; `pyproject.toml` was the only overlap.

The blocker: the table distinguished *vendor ignores* from *vendor rejects*, but the escape hatch did not, so reject-families forwarded a stored agent `temperature` into a documented 400. `_SamplingPolicy` is now a `NamedTuple` carrying a `rejects` flag, and a rejecting row clears `supports_sampling_params` — reusing the existing suppression rather than inventing a second one. A `NamedTuple` rather than a second `(None, None)` sentinel because **equal constant tuples are interned to one object**, so two such sentinels would be indistinguishable by identity and silently collapse.

Anthropic ≥4.7 and GPT-6+ are REJECT; Gemini 3.x stays OMIT. Kimi stays OMIT — that row also catches legacy ids (`kimi-k2-0711-preview`, `moonshot-v1-128k`) that accept the pair, and mainland policy is unverified; the documented-reject coding ids keep their scoped suppression via `_KIMI_PINNED_SAMPLING`.

| behaviour | override `temperature=0.2` reaches wire? |
|---|---|
| `gpt-6-astra`, `claude-opus-4-7/4-8` (REJECT) | no — suppressed |
| `openrouter/anthropic/claude-opus-4-7` (REJECT) | no — suppressed |
| `gemini-3.8-flash` (OMIT, vendor ignores) | yes — harmless |
| `claude-sonnet-4-5`, `kimi-k2-0711-preview` (HONOUR) | yes — correct |

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/local_operator/bootstrap.py
+++ b/local_operator/bootstrap.py
@@ -180,7 +180,10 @@ async def initialize_operator(
     credential_manager: CredentialManager,
     agent_registry: AgentRegistry,
     env_config: EnvConfig,
-    sampling_overrides: Optional[dict[str, float]] = None,
+    # Values are Optional because a model whose family policy is to OMIT
+    # carries ``None`` on its spec; the filter below drops those so an absent
+    # knob is never turned into an explicit send.
+    sampling_overrides: Optional[dict[str, Optional[float]]] = None,
     scheduler_service: Optional[SchedulerService] = None,
     status_queue: Optional[StatusQueue] = None,
     request_hosting: Optional[str] = None,
--- a/local_operator/harness/types.py
+++ b/local_operator/harness/types.py
@@ -1571,8 +1571,16 @@ class ModelSpec(BaseModel):
     # only chat/completions.
     supports_responses_api: bool = False
     base_url: str | None = None  # override for OpenAI-compatible endpoints
-    temperature: float = 0.2
-    top_p: float = 0.9
+    # ``None`` means OMIT: send no key at all and let the vendor's own default
+    # apply. That is now the common case rather than an exotic one — most
+    # current families either reject the pair, ignore it, or document a default
+    # this app has no business overriding (see ``_SAMPLING_POLICY``). Optional
+    # rather than a magic float because there is no in-band float that means
+    # "unset", and the wire clients must distinguish "send 0.0" from "send
+    # nothing". Derived in ``build_model_spec`` so no wire client needs
+    # model-name knowledge.
+    temperature: float | None = None
+    top_p: float | None = None
     reasoning: bool = False
     # Explicit provider reasoning level. Fallback routes may change providers,
     # so the effort rides on the resolved spec rather than global session state.
--- a/local_operator/model/configure.py
+++ b/local_operator/model/configure.py
@@ -26,7 +26,7 @@ import os
 import re
 import time
 from collections.abc import AsyncIterator, Awaitable, Callable, Mapping
-from typing import TYPE_CHECKING, Any, Optional
+from typing import TYPE_CHECKING, Any, NamedTuple, Optional
 
 from pydantic import BaseModel, SecretStr
 
@@ -233,6 +233,7 @@ _NO_SAMPLING_PARAMS = re.compile(
 #: ``kimi-k2-0711-preview`` must not match on its ``k2`` fragment.
 _KIMI_PINNED_SAMPLING = re.compile(r"^(?:k\d+(?:-|$)|kimi-for-coding)")
 
+
 #: The per-family SAMPLING POLICY table: what this app may assert about
 #: ``temperature``/``top_p``, keyed on the MODEL id.
 #:
@@ -277,11 +278,51 @@ _KIMI_PINNED_SAMPLING = re.compile(r"^(?:k\d+(?:-|$)|kimi-for-coding)")
 #: ``_EFFORT_TABLE`` documents. Unanchored, so an aggregator prefix
 #: (``google/…``) and ``:free``/``-preview``/``-thinking`` suffixes still match.
 #:
+#: SCOPED TO CHAT/COMPLETION sampling. Every row is derived from a vendor's
+#: chat-model documentation, and ``routes/speech.py`` is a second consumer that
+#: resolves speech ids (``whisper-1``, ``gpt-4o-mini-tts``) through the same
+#: table. Today they all reach the OMIT fallback, which is correct for them, so
+#: this is a caveat rather than a defect — but a speech id sharing a prefix with
+#: a chat family would inherit reasoning that was never about speech endpoints,
+#: and that is the point at which this table needs a speech-aware branch rather
+#: than another row.
+#:
 #: FIRST MATCH WINS, so a narrower row must precede a broader one.
-_SamplingPolicy = tuple[Optional[float], Optional[float]]
+#: A family's sampling rule: the seeded ``temperature``/``top_p`` (``None``
+#: meaning "send no key"), plus whether the vendor REJECTS a value it did not
+#: ask for.
+#:
+#: ``rejects`` is a third state rather than a second ``None``, because the two
+#: reasons to send nothing have opposite consequences for an explicit override,
+#: and conflating them reopens the exact outage this table exists to close:
+#:
+#: * Vendor IGNORES the value (Gemini 3.x). Our default assertion is pointless,
+#:   so we drop it — but a user or agent that deliberately sets one loses
+#:   nothing by having it sent, so the escape hatch stays open.
+#: * Vendor REJECTS the value (Anthropic ≥4.7: "all other values will be
+#:   rejected with a 400 error"; OpenAI GPT-6: "Remove `temperature`,
+#:   `top_p`"). An override here is not a preference, it is a turn that fails
+#:   EVERY time. A stored agent ``temperature`` is the shape the server's own
+#:   API examples advertise, so this is an ordinary path, not an exotic one.
+#:
+#: Note this cannot be expressed by comparing against a ``(None, None)``
+#: sentinel: equal constant tuples are interned to one object, so two such
+#: sentinels would be indistinguishable by identity and silently collapse into
+#: each other.
+class _SamplingPolicy(NamedTuple):
+    temperature: Optional[float]
+    top_p: Optional[float]
+    rejects: bool = False
 
-#: Send neither key and let the vendor's own default apply.
-_OMIT_SAMPLING: _SamplingPolicy = (None, None)
+
+#: Send neither key and let the vendor's own default apply. An explicit user or
+#: agent value STILL RIDES on top of this.
+_OMIT_SAMPLING = _SamplingPolicy(None, None)
+
+#: Send neither key, and SUPPRESS an explicit override too — the same judgement
+#: :data:`_NO_SAMPLING_PARAMS` already makes for Claude 5+ and the o-series,
+#: kept here so each family's rule sits beside the vendor citation for it.
+_REJECT_SAMPLING = _SamplingPolicy(None, None, rejects=True)
 
 _SAMPLING_POLICY: tuple[tuple[re.Pattern[str], _SamplingPolicy], ...] = (
     # -- Google ------------------------------------------------------------
@@ -301,27 +342,38 @@ _SAMPLING_POLICY: tuple[tuple[re.Pattern[str], _SamplingPolicy], ...] = (
     # Seeded rather than omitted because these are the documented per-model
     # defaults the API's own getModel reports (temperature 1.0, topP 0.95), and
     # keeping the keys present preserves a real, tunable knob.
-    (re.compile(r"gemini"), (1.0, 0.95)),
+    (re.compile(r"gemini"), _SamplingPolicy(1.0, 0.95)),
     # -- Anthropic ---------------------------------------------------------
     # "Models released after Claude Opus 4.6 do not support setting
     # temperature. A value of 1.0 will be accepted for backwards compatibility,
     # all other values will be rejected with a 400 error."
     # (https://platform.claude.com/docs/en/api/messages.md). This is a LIVE
     # OUTAGE on 4.7/4.8: the app sent 0.2, which is exactly the rejected case.
-    # 4.5/4.6 still honour the pair and must not regress, so the arm starts at
-    # generation 4.7 and is written forward. Generation 5+ is already covered
-    # by _NO_SAMPLING_PARAMS below.
-    (re.compile(r"claude-[a-z]+-4[.-](?:[7-9]|\d{2,3})(?!\d)"), _OMIT_SAMPLING),
+    #
+    # REJECT, not OMIT: "rejected with a 400" means an override cannot be
+    # honoured either. Dropping only our own default would leave a stored agent
+    # temperature failing every turn on this family — the same outage, moved
+    # from the default path to the override path.
+    #
+    # The arm starts at 4.7 and reads forward so it cannot reach back over
+    # 4.5/4.6, which genuinely honour the pair (see the fallback note below for
+    # what actually happens to them). Generation 5+ is already covered by
+    # _NO_SAMPLING_PARAMS.
+    (re.compile(r"claude-[a-z]+-4[.-](?:[7-9]|\d{2,3})(?!\d)"), _REJECT_SAMPLING),
     # -- OpenAI ------------------------------------------------------------
     # GPT-6 Astra: "Unsupported parameters: Remove `temperature`, `top_p`, and
     # `top_logprobs`."
     # (https://developers.openai.com/api/docs/guides/latest-model.md).
     # Forward-reading over the generation digit, because the pre-existing
     # literal `gpt-5` in _NO_SAMPLING_PARAMS does not match `gpt-6-astra`.
-    # UNKNOWN: GPT-5.6 is not covered by primary docs; it keeps the omission by
-    # precaution, which is both the safe direction and the already-shipped
-    # behaviour.
-    (re.compile(r"gpt-(?:[5-9]|\d{2,})"), _OMIT_SAMPLING),
+    # UNKNOWN: GPT-5.6 is not covered by primary docs; it keeps the suppression
+    # by precaution, which is both the safe direction and the already-shipped
+    # behaviour (the literal `gpt-5` arm in _NO_SAMPLING_PARAMS already strips
+    # the pair for the 5.x line).
+    #
+    # REJECT, not OMIT: "Unsupported parameters: Remove ..." is a rejection, so
+    # an override must be suppressed rather than forwarded into a 400.
+    (re.compile(r"gpt-(?:[5-9]|\d{2,})"), _REJECT_SAMPLING),
     # -- DeepSeek ----------------------------------------------------------
     # V4 runs with thinking ON by default, and "Thinking mode does not support
     # the temperature, top_p, presence_penalty, or frequency_penalty parameters
@@ -344,6 +396,15 @@ _SAMPLING_POLICY: tuple[tuple[re.Pattern[str], _SamplingPolicy], ...] = (
     # narrower provider-keyed rule is retained: it is the one case with a
     # documented ROUTE difference, and it still guards ids like `k3` that carry
     # no vendor name at all.
+    #
+    # OMIT rather than REJECT, deliberately, even though the CURRENT lineup
+    # does reject: this row also catches legacy ids (`kimi-k2-0711-preview`,
+    # `moonshot-v1-128k`) that accept the pair, and a mainland-specific policy
+    # is UNVERIFIED. Suppressing an override for the whole name-space would
+    # take a working setting away from models that honour it, on evidence that
+    # only covers the current models. The ids whose rejection IS documented and
+    # was observed live keep their hard suppression through
+    # _KIMI_PINNED_SAMPLING above, which is scoped to exactly those.
     (re.compile(r"kimi|moonshot"), _OMIT_SAMPLING),
     # -- Alibaba / Qwen ----------------------------------------------------
     # The one vendor publishing a real per-model default table: "Qwen3-Coder
@@ -357,7 +418,7 @@ _SAMPLING_POLICY: tuple[tuple[re.Pattern[str], _SamplingPolicy], ...] = (
     # compaction through the same spec as its coding turns, and 0.2 is the
     # exact assertion this table exists to stop making. Note that qwen3.8 with
     # thinking on auto-clamps temperature below 0.6 regardless.
-    (re.compile(r"qwen|qwq|qvq"), (0.7, 0.8)),
+    (re.compile(r"qwen|qwq|qvq"), _SamplingPolicy(0.7, 0.8)),
     # -- Z.AI / GLM --------------------------------------------------------
     # Honoured, but RANGE-rejected (temperature [0.0, 1.0], top_p [0.01, 1.0])
     # and silently ignored when `do_sample=false`. The defaults are per-series
@@ -573,23 +634,28 @@ def build_model_spec(hosting: str, model_name: str, info: ModelInfo | None = Non
     # pair on the mainland host — see _KIMI_PINNED_SAMPLING.
     if canonical == "kimi" and _KIMI_PINNED_SAMPLING.match(lowered):
         supports_sampling_params = False
-    # The per-family policy. Unlike the two rules above this is NOT about
-    # rejection — most of these families accept the pair and then ignore it, or
-    # honour it at a value nothing here should be inventing. `None` means the
-    # key is omitted so the vendor's own default applies; see _SAMPLING_POLICY
-    # for each row's citation.
+    # The per-family policy. Most of these families accept the pair and then
+    # ignore it, or honour it at a value nothing here should be inventing, so a
+    # `None` here only means "we stop asserting" and an explicit override still
+    # rides. A row marked `rejects` is the stronger case and clears
+    # `supports_sampling_params` below; see _SAMPLING_POLICY for each row's
+    # citation.
     #
     # A user-supplied model (ollama) never consults the id-keyed table: its ids
     # are arbitrary and its publisher's Modelfile has already set the tuning we
     # would be overwriting.
-    sampling_temperature: Optional[float] = DEFAULT_TEMPERATURE
-    sampling_top_p: Optional[float] = DEFAULT_TOP_P
+    #
+    # No initial value: every branch below assigns, and seeding this with
+    # DEFAULT_TEMPERATURE/DEFAULT_TOP_P left the app-wide constants looking like
+    # a live default in the one file whose job is to stop them being one.
+    policy: _SamplingPolicy
     if canonical in _USER_SUPPLIED_MODEL_PROVIDERS:
-        sampling_temperature, sampling_top_p = _OMIT_SAMPLING
+        policy = _OMIT_SAMPLING
+
     else:
-        for pattern, (policy_temperature, policy_top_p) in _SAMPLING_POLICY:
+        for pattern, matched_policy in _SAMPLING_POLICY:
             if pattern.search(lowered):
-                sampling_temperature, sampling_top_p = policy_temperature, policy_top_p
+                policy = matched_policy
                 break
         else:
             # The FALLBACK, and the most consequential row of all. Every vendor
@@ -597,13 +663,28 @@ def build_model_spec(hosting: str, model_name: str, info: ModelInfo | None = Non
             # one is near 0.2 — so asserting the app-wide constant over a model
             # nobody has characterised is the very defect this table fixes.
             # Silence is the honest answer for an unknown model.
-            sampling_temperature, sampling_top_p = _OMIT_SAMPLING
-        # The provider's own listing may only NARROW what the table decided: a
-        # stated allowlist that omits `temperature` is the aggregator telling us
-        # it will not forward the key. Silence never widens an OMIT back into a
-        # send — see `_listing_forbids_sampling`.
-        if _listing_forbids_sampling(canonical, model_name):
-            sampling_temperature, sampling_top_p = _OMIT_SAMPLING
+            #
+            # This is also where Anthropic 4.5/4.6 land. They HONOUR the pair,
+            # and deliberately get no row of their own: omitting yields
+            # Anthropic's own 1.0, which is exactly what a row would have to
+            # seed, so a row would add a maintained constant that could only
+            # drift. The 4.7+ arm above is written forward purely so it cannot
+            # reach back over them.
+            policy = _OMIT_SAMPLING
+    sampling_temperature, sampling_top_p = policy.temperature, policy.top_p
+    # A vendor that REJECTS an unrequested value cannot honour an override
+    # either: forwarding one is a documented HTTP 400 on every turn, which is
+    # the same outage this table closes, merely moved onto the override path.
+    # `supports_sampling_params` is the existing mechanism for exactly that, so
+    # a rejecting row clears it rather than inventing a second suppression.
+    if policy.rejects:
+        supports_sampling_params = False
+    # The provider's own listing may only NARROW what the table decided: a
+    # stated allowlist that omits `temperature` is the aggregator telling us
+    # it will not forward the key. Silence never widens an OMIT back into a
+    # send — see `_listing_forbids_sampling`.
+    if _listing_forbids_sampling(canonical, model_name):
+        sampling_temperature = sampling_top_p = None
     # A GUARDED read, not `info.name`. `info` is duck-typed here — the legacy
     # public helpers and the tests hand in stand-ins, and `name` is the one
     # attribute name that collides with something that is not a string on almost
@@ -2101,10 +2182,20 @@ def configure_model(
     # DEFAULT_TOP_P, which meant the model's own seed — Google's 1.0/0.95 for
     # Gemini 2.x — was overwritten with the app-wide 0.2/0.9 on the main
     # construction path, making a per-model default impossible to express.
-    # ``None`` is the "caller said nothing" sentinel; both real override paths
-    # (an agent record's stored knobs via ``_AGENT_SAMPLING_FIELDS``, and the
-    # server's per-request ``ChatRequest.options``) already omit the argument
-    # entirely unless a value is set, so the sentinel costs them nothing.
+    # ``None`` is the "caller said nothing" sentinel. The override path this
+    # serves is an agent record's stored knobs, which reach here via
+    # ``_AGENT_SAMPLING_FIELDS`` and already omit the argument entirely unless a
+    # value is set, so the sentinel costs them nothing.
+    #
+    # Deliberately NOT claimed for the server's per-request ``ChatRequest
+    # .options``: those routes mutate the SCALAR
+    # ``model_configuration.temperature``, while the wire reads the SPEC, and
+    # the two are independent fields set once at construction — so that path
+    # does not currently reach the provider. It is broken identically before
+    # this change and is out of scope here, but the claim does not belong in a
+    # comment the next reader will trust. ``bootstrap.py``'s
+    # ``sampling_overrides`` seam IS correct and does reach the spec via
+    # ``set_model``.
     sampling_overrides: dict[str, float] = {}
     if temperature is not None:
         sampling_overrides["temperature"] = temperature
--- a/local_operator/model/configure.py
+++ b/local_operator/model/configure.py
@@ -136,8 +136,10 @@ class ModelConfiguration:
     instance: Any
     info: ModelInfo
     api_key: Optional[SecretStr]
-    temperature: float
-    top_p: float
+    # ``None`` mirrors the spec's OMIT: no value is sent and the vendor's own
+    # default applies. See ``_SAMPLING_POLICY``.
+    temperature: Optional[float]
+    top_p: Optional[float]
     top_k: Optional[int]
     max_tokens: Optional[int]
     frequency_penalty: Optional[float]
@@ -153,8 +155,8 @@ class ModelConfiguration:
         instance: Any = None,
         info: ModelInfo | None = None,
         api_key: Optional[SecretStr] = None,
-        temperature: float = DEFAULT_TEMPERATURE,
-        top_p: float = DEFAULT_TOP_P,
+        temperature: Optional[float] = DEFAULT_TEMPERATURE,
+        top_p: Optional[float] = DEFAULT_TOP_P,
         top_k: Optional[int] = None,
         max_tokens: Optional[int] = None,
         frequency_penalty: Optional[float] = None,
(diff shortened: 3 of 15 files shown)


damianvtran:
### Agent review — round 1

**Reviewer:** `reviewer` subagent (lopdev team) — Claude Opus 4.6
**Scope:** `origin/main..dcac3c26`

Independent review. I did not write this code. I re-derived every claim below against the head commit in `/tmp/lop-gemini-params`; nothing here is inferred from the diff alone.

The design is right and I want to say so plainly before the findings: keying the policy on the **model** rather than the provider is the correct call and is what makes `radient-key` — a provider id this PR never mentions — resolve identically to `radient` (verified: `google/gemini-3.8-flash` → `None/None`, `google/gemini-2.5-flash` → `1.0/0.95` on both). Resolving `temperature` and `top_p` independently is genuinely necessary for the Qwen row and is not over-engineering. The three plumbing fixes are real bugs and the null-vs-absent handling is clean on all four wire builders.

One blocker, on an edge the PR's own test matrix does not cover.

---

#### R1 — BLOCKER — the escape hatch reopens the exact HTTP 400 the PR exists to close, on `claude-opus-4-7` / `claude-opus-4-8` / `gpt-6-astra`

`clients.py:651-664` gates only on `supports_sampling_params`. The three families the PR documents as **hard 400 rejections** are routed through the *policy* (`None`) rather than through `_NO_SAMPLING_PARAMS`, so they keep `supports_sampling_params=True` and an override rides straight to the wire:

```
anthropic/claude-opus-4-7  supports_sampling_params=True  request temp=0.2 -> WIRE {'temperature': 0.2}
anthropic/claude-opus-4-8  supports_sampling_params=True  request temp=0.2 -> WIRE {'temperature': 0.2}
openai/gpt-6-astra         supports_sampling_params=True  request temp=0.2 -> WIRE {'temperature': 0.2}

anthropic/claude-opus-5    supports_sampling_params=False request temp=0.2 -> WIRE {}   # correct
openai/gpt-5               supports_sampling_params=False request temp=0.2 -> WIRE {}   # correct
```

Not hypothetical, and it does not need a hand-typed override to fire. An agent record with a stored `temperature` reaches it through the ordinary path:

```
configure_model(hosting='anthropic', model_name='claude-opus-4-7', temperature=0.2)
  -> spec.temperature = 0.2  -> WIRE {'temperature': 0.2}
```

Anthropic's quoted text is *"all other values will be rejected with a 400 error"*, so this is a failed turn on **every** turn for that agent — precisely the live outage the PR is written to fix, merely relocated from the default path to the override path. `AgentEditFields.temperature` (`agents.py:98`) is user-settable and the server's own OpenAPI examples advertise `{"temperature": 0.2, "top_p": 0.9}` (`chat.py:89,217,331,455`), so a stored 0.2 is the documented shape, not an exotic one.

The PR reasons about this in "Preserved invariants" — *"Gemini 3 ignores rather than rejects, so it belongs in the second category — which is what preserves the escape hatch"* — and that reasoning is correct **for Gemini**. It was then applied to Anthropic 4.7/4.8 and GPT-6 Astra, which the PR's own vendor table classifies as *rejects*, not *ignores*. Two different vendor behaviours were given one mechanism.

Concrete fix, one line: the reject-documented arms belong in `_NO_SAMPLING_PARAMS` (or must otherwise clear `supports_sampling_params`), leaving `_SAMPLING_POLICY` `None` for the ignore/undocumented families where the escape hatch is genuinely safe. Extending the existing `_NO_SAMPLING_PARAMS` Anthropic arm from `[5-9]` to also catch `4[.-][7-9]`, and adding `gpt-6`+, closes it. Note the OpenAI case is half-closed by accident already: `gpt-5`/`gpt-5.6`/`gpt-5-mini` match the literal `gpt-5` in `_NO_SAMPLING_PARAMS` and are correctly stripped, while `gpt-6-astra` — the one id with a documented "Remove `temperature`" quote — is not.

---

#### R2 — MINOR — the Anthropic policy arm never fires; 4.5/4.6 reach the same OMIT by falling through

The comment at `configure.py:109-111` states the arm *"starts at generation 4.7"* so that *"4.5/4.6 still honour the pair and must not regress"*. The regex is correct in isolation — `claude-[a-z]+-4[.-](?:[7-9]|\d{2,3})(?!\d)` matches `claude-opus-4-7`/`4-8`/`4.7`/`claude-opus-4-7-20260115` and not `claude-opus-4-5`/`4-6`/`claude-sonnet-4-5-20250929`. The snapshot-date trap from `effort.py:140-146` is genuinely closed: `claude-sonnet-4-20250514` does not read `20250514` as a generation.

But there is no Anthropic row that seeds 4.5/4.6, so they hit the catch-all `else` and get OMIT anyway:

```
claude-opus-4-7             -> None/None   (via the 4.7+ arm)
claude-opus-4-5             -> None/None   (via the fallback)
claude-sonnet-4-5-20250929  -> None/None   (via the fallback)
```

The end state is defensible — OMIT is the safe direction — but the code does not do what its comment says it does, and the distinction the comment draws is invisible at runtime. Either drop the claim that 4.5/4.6 "still honour the pair and must not regress" (they are not treated differently), or make the arm load-bearing. As written, the next reader will believe a 4.5/4.6 seed exists and will not find it. Same shape applies to the Gemini `[3-9]|\d{2,3}` arm, which *is* load-bearing because `gemini` <=2.5 has a real seed row behind it — that one is correct.

Regex verification I ran, for the record — all pass: both separator spellings (`gemini-3-flash`, `gemini-3.8-flash`); aggregator prefixes (`google/`, `anthropic/`, `openai/`, `models/`); suffixes `:free`, `-preview`, `-thinking`, `-image-preview`; snapshot dates (`gemini-2.5-flash-preview-09-2025`, `gemini-2.5-computer-use-preview-10-2025`, `gemini-3-pro-preview-11-2025` all classify correctly); forward-reading (`gemini-4-pro`, `gemini-10-pro`, `claude-opus-4-9`, `gpt-10-turbo` all caught); and no over-match (`gemini-live-2.5-flash-preview`, `gemini-embedding-001`, `gpt-4o`, `gpt-4.1` all stay off the 3+/5+ arms).

---

#### R3 — MINOR — `DEFAULT_TEMPERATURE`/`DEFAULT_TOP_P` initializer at `configure.py:585-586` is unreachable

```python
sampling_temperature: Optional[float] = DEFAULT_TEMPERATURE
sampling_top_p: Optional[float] = DEFAULT_TOP_P
```

Every path reassigns before use: the `ollama` branch, a matching policy row, or the `for/else` fallback (which sets `_OMIT_SAMPLING`). There is no input for which `0.2/0.9` survives to the spec — confirmed across 20 provider/model pairs. Harmless today, but it is the only remaining place in the file where the constants look like a live default, which is exactly the misreading this PR is trying to eliminate. Initialize to `_OMIT_SAMPLING`, or drop the initializer and let the branches assign.

---

#### R4 — MINOR — `speech.py` now passes agent knobs through to a route that never had a policy applied

The `or 0.2`/`or 0.9` removal at `speech.py:129-136` is right, and the note about `or` eating a deliberate `0.0` is a real catch. Worth confirming the intent, though: `create_agent_speech` calls `configure_model` for a **speech** model, and the sampling policy table is written entirely against chat/completion families. An id like `gpt-4o-mini-tts` or `whisper-1` will now resolve through the same table (falling to OMIT, which is fine), but `gpt-6`-prefixed speech ids would hit the `gpt-(?:[5-9]|\d{2,})` arm on reasoning that was never about speech endpoints. Not a defect I can demonstrate — no speech id currently collides — so this is a note, not a bug: the table's docstring should say it is scoped to chat sampling, since `speech.py` is now a second consumer.

---

#### R5 — NIT — PR body states a merge base that no longer holds, and one evidence line overstates

Two small accuracy points, neither affecting the code:

1. The PR is cut from `6437caf6`, but `origin/main` has since advanced to `06928224` (#593). A raw `git diff origin/main..dcac3c26` therefore renders #593's work — `cached_available_models`, `initial_catalogue`, the parallel `_fetch_provider` gather, `radient-key` in `AGGREGATOR_PROVIDERS`/`PUBLIC_LISTING_PROVIDERS` — as **deletions this PR never made**, and reports +726/−272 across 17 files instead of the true +715/−25 across 11. I verified the true diff against the merge base and confirmed the only overlap with #593 is `pyproject.toml`, where both bump `0.44.66 → 0.44.67` **identically**; `git merge-tree` reports a clean merge. So there is nothing to fix in the branch — but whoever merges should not be alarmed by that diff, and the version bump is now a no-op that lands the release at the same `0.44.67` #593 already took. **Confirm the intended release version before tagging.**
2. *"Every one of the seventeen routes was receiving `0.2/0.9` on main"* is right for the wire, but the adjacent server-`options` claim needs a caveat — see R6.

---

#### R6 — MINOR (pre-existing, not introduced here) — the server's per-request `options` override does not reach the wire, before *or* after this PR

The `configure_model` comment at `configure.py:2105-2107` names *"the server's per-request `ChatRequest.options`"* as one of the two override paths the sentinel preserves. It is not currently a working path. `chat.py:147-150` mutates the **scalar** `model_configuration.temperature`, while the wire reads `request.model.temperature` — the **spec** — and the two are independent fields set once at construction:

```
mc.temperature -> 0.3   (route wrote this)
mc.spec.temperature -> 1.0
WIRE SENT: {'temperature': 1.0, 'top_p': 0.95}   # user asked for 0.3
```

I checked this against `origin/main` and it is **equally broken there** (`mc.temperature=0.3`, wire `{'temperature': 0.2, 'top_p': 0.9}`), so this PR neither causes nor worsens it — genuinely out of scope, and I am not asking for it here. But the comment asserts the path works, which will mislead the next reader, and the PR body leans on it as a preserved escape hatch. Either soften the comment to name only `_AGENT_SAMPLING_FIELDS` (which I verified **does** work correctly — `session_factory.py:1438-1441` passes only non-`None` fields, so the sentinel behaves exactly as designed), or record the server path as `deferred — pre-existing, tracked separately`. Note `bootstrap.py`'s `sampling_overrides` seam **is** correct and does reach the spec via `set_model`.

---

### What I verified as correct

Re-derived independently on the head commit; all confirmed:

- **Null-vs-absent on all four wire builders.** `_sampling_params` produces `{}` for an OMIT family on both the `top_p` and `topP` spellings — the key is genuinely **absent**, no `null` leak. The `clients.py:630` warning about httpx serializing `None` as JSON `null` is respected.
- **`ChatRequest(temperature=None, top_p=None)`** (explicit user `None`) → `{}`, not a null.
- **A deliberate `0.0` survives** at both the spec and the wire (`{'temperature': 0.0, 'top_p': 0.0}`) — the falsy-value trap is closed.
- **Failover.** A seeded primary (`alibaba/qwen3-coder-next`, 0.7/0.8) hopping onto `google/gemini-3.8-flash` or `anthropic/claude-opus-4-7` yields `None/None` and an empty wire body — the re-imposition bug is genuinely fixed, and the hop onto `gemini-2.5-flash` correctly picks up the target's *own* 1.0/0.95.
- **`configure_model` sentinel.** No override → family policy survives; `temperature=0.35` → wins; `temperature=0.0` → wins and is not swallowed.
- **The operator's spot-check table reproduces exactly**, including `ollama/qwen3:32b` → omitted via `_USER_SUPPLIED_MODEL_PROVIDERS` and `claude-opus-5` → hard-reject.
- **Gates.** `flake8`, `black --check`, `isort --check-only` all clean over the tracked tree (834 files); `pyright --pythonpath .venv/bin/python` reports **0 errors, 0 warnings** across all changed files plus `chat.py`. The 790 tests in the four touched test files pass in 8.6s. (Heads-up: two untracked scratch files, `tests/e2e/test_qa594_boot.py` and `.bak`, are sitting in the worktree and DO fail flake8/isort — they are QA leftovers, not part of the PR, but they should be removed so they cannot be committed by accident.)
- **Comment quality** meets the house standard: every row carries a primary-source citation and states its safe failure direction, and the `UNKNOWN` cells the operator flagged (xAI, Mistral served defaults, GPT-5.6) are labelled **UNKNOWN in the code itself** rather than laundered into confident assertions. The stale-DeepSeek-0.0 warning to future readers is a good touch. The `claude-opus-4-7` 400 is explicitly marked *"vendor-documented, not locally reproduced"* in the PR body — that honesty is the right call and I am not treating it as a gap.

---

### Verdict

**Not clean — 1 blocker (R1).**

R1 is a genuine ship-stopper and narrowly scoped: three model families whose vendors document a hard 400 keep `supports_sampling_params=True`, so a stored agent `temperature` — the shape the server's own API examples advertise — reaches the wire and fails every turn. It is the PR's own headline bug surviving on the override path. The fix is a regex extension in `_NO_SAMPLING_PARAMS`, not a design change; the policy table itself is sound and should not be restructured.

R2/R3/R4/R6 are minors and R5 a nit — none needs to block once R1 lands, and R6 is explicitly pre-existing and out of scope. I'd ask for R2 (comment says something the code does not do) and R3 (dead constants that reintroduce the misreading) in the same commit since both are one-liners in files already being touched.

Re-review on round 2 will scope to the remediation diff.

damianvtran:
### QA report — round 1

**Reviewer:** qa-tester subagent (Claude Opus 4.5), independent of the implementer
**Scope:** `06928224..dcac3c26e2ba35bf4b8c25983e9899e7daee4c80`
**Verdict: PASS with one non-blocking finding (Q1, major) and two minor (Q2, Q3).** Nothing here blocks merge on its own; Q1 is a real user-visible gap that this PR's own headline claim invites, and it is a two-line fix if you want it in scope.

Environment: worktree `/tmp/lop-gemini-params`, isolated `HOME` + `LOCAL_OPERATOR_CONFIG_DIR` created cold for this pass (never `~/.local-operator`), operator's live sessions untouched. Venv resolution verified per AGENTS.md ~line 180 before any cell — `import local_operator` resolves to `/private/tmp/lop-gemini-params/local_operator/__init__.py`, and the base-comparison tree was verified the same way against `/tmp/qa594/basewt`. No repo file was modified; `git status` is clean.

I verified by execution, not by reading the diff: 183 matrix cells driving the real `build_model_spec` / `configure_model` into the real four wire builders, the real failover resolver, the real headless app, plus a before/after reproduction against `origin/main`.

---

#### Suites and gates (all green on head)

| gate | command | result |
|---|---|---|
| unit | `.venv/bin/python -m pytest tests/unit -q` | **11852 passed, 15 skipped** (586s) |
| e2e | `pytest tests/e2e -m e2e -n0 -q` | **4 passed** (10.7s) |
| flake8 | `.venv/bin/python -m flake8 .` | clean |
| black | `uvx --from black==26.1.0 black --check .` | 834 files unchanged |
| isort | `uvx isort==5.13.2 --check .` | clean |
| pyright | `.venv/bin/python -m pyright --pythonpath .venv/bin/python .` | 0 errors, 0 warnings |

#### The reported defect, reproduced before and fixed after

Same script against both trees, main-turn path, no request-level override:

| model | BEFORE (`06928224`) | AFTER (`dcac3c26`) |
|---|---|---|
| `google/gemini-3.8-flash` | `temperature=0.2 topP=0.9` | **absent** |
| `openai/gpt-6-astra` | `temperature=0.2 top_p=0.9` | **absent** |
| `alibaba/qwen3-coder-next` | `temperature=0.2 top_p=0.9` | **0.7 / 0.8** |
| `openrouter/anthropic/claude-opus-4-7` | `temperature=0.2` | **absent** |
| `radient/anthropic/claude-opus-4-8` | `temperature=0.2` | **absent** |

One correction to the PR narrative, in your favour and against it at once: on the **direct** `anthropic` route `claude-opus-4-7` already omitted on `origin/main`, because the Anthropic builder gates sampling on the same branch that writes `thinking`. The "live outage on every single turn" was real, but it was reached through the **aggregator** routes (`openrouter/`, `radient/`), which the last two rows above confirm. Worth correcting in the PR body so the next reader does not go looking for a 400 on a route that never sent one.

---

#### Group 1 — wire bodies, per family, per wire (104 cells, 104 PASS)

Every spec from the real `build_model_spec`, every body from the real four builders, asserting key ABSENT vs PRESENT-with-value, plus a serialized-JSON null-leak check on every cell. **No `"temperature":null`, `"top_p":null` or `"topP":null` leaked anywhere.**

Confirmed absent on all four wires: `claude-opus-4-7`, `claude-opus-4-8`, `claude-opus-5`, `gpt-6-astra`, `gpt-5.6`, `gpt-4o`, `o3`, `gemini-3.8-flash`, `gemini-3.1-pro-preview`, `deepseek-v4-pro`, `k3`, `kimi-k2-0711-preview`, `mistral-large-latest`, `grok-4`, `glm-5.3`, `qwen3:32b` (ollama).
Confirmed seeded: `gemini-2.5-flash` → `1.0` / `0.95` (`topP` on the Google wire), `qwen3-coder-next` → `0.7` / `0.8`.
Confirmed the aggregator routes match the direct route for every one of the above under both `openrouter/` and `radient/` prefixes — the model-keyed choice does what it claims.

#### Group 2 — hard-reject regression (15 cells, 15 PASS)

The pre-existing `supports_sampling_params` mechanism was not over-matched. `claude-opus-5`, `claude-sonnet-5`, `gpt-5`, `o1`, `o3`, `o3-mini`, `k3`, `kimi-for-coding` all still `False`. `kimi-k2-0711-preview` still `True` (no `k2`-fragment over-match). **`('google','gemini-2.5-flash-thinking')` is still `True`** — it did not flip.

#### Group 3 — override paths (13 cells, 13 PASS)

A user who wants determinism is not stranded, driven through the real code path:
- agent-record knobs via real `configure_model(temperature=…, top_p=…)`: `0.15/0.5` on `gemini-3.8-flash` reaches the wire as `0.15/0.5`.
- server per-request `ChatRequest.options` (the `routes/chat.py` → `server/utils/operator.py` shape): `0.15` reaches the wire on an OMIT family.
- **`temperature=0.0` is distinguishable from unset** — `0.0` is sent, unset is absent. The `Optional[float]` sentinel does what the comment claims.
- independent knobs confirmed: an agent temperature on `gemini-2.5-flash` overrides temp while `top_p` keeps its `0.95` seed.
- **a hard-reject model still suppresses a user override** — agent temp `0.15` on `claude-opus-5` emits nothing, so an override cannot 400 the turn.

#### Group 4 — failover (9 cells, 9 PASS)

Seeded → OMIT drops the value (`gemini-2.5-flash`→`gemini-3.8-flash`, `qwen3-coder-next`→`gpt-6-astra`, →`claude-opus-4-7`: all absent). OMIT → seeded picks the target's seed up (`gpt-6-astra`→`qwen3-coder-next` = `0.7/0.8`). A session-level override on the base does **not** leak across a hop onto an OMIT family, while a per-request override still outranks the spec after the hop — exactly the asymmetry `spec_for_target`'s docstring promises.

#### Group 5 — boot smoke (5 cells, 5 PASS)

Real `OperatorApp` under the headless pilot in an isolated config dir, one boot per model, frame exported to SVG, rendered and **viewed**:

| model | boot | painted | composer | status band |
|---|---|---|---|---|
| `google/gemini-3.8-flash` | 1.02s | 30 rows | accepts keys | names the model |
| `anthropic/claude-opus-4-7` | 0.47s | 30 rows | accepts keys | renders |
| `openai/gpt-6-astra` | 0.52s | 30 rows | accepts keys | names the model |
| `google/gemini-2.5-flash` | 0.47s | 30 rows | accepts keys | names the model |
| `alibaba/qwen3-coder-next` | 0.50s | 30 rows | accepts keys | names the model |

No crash, no traceback in any painted frame, no boot slowdown attributable to the table.

#### Group 7 — forward-reading regex edge cases (31 cells, 31 PASS)

`gemini-4.0-pro`, `gemini-9-ultra`, `gemini-10-flash`, `gemini-12.3-pro` all OMIT while `gemini-2.5/2.0/1.5` keep `1.0/0.95`; `claude-opus-4-10` and `claude-sonnet-4-9` OMIT while `claude-opus-4-5`/`4-6` do not match that arm; `gpt-10-x` and `gpt-12` OMIT; `qwq`/`qvq` reach the Qwen row. No misfire found in either direction.

#### Group 8 — speech route (6 cells, 6 PASS)

Reproduced the bug `routes/speech.py` fixes: pre-fix, an agent with a deliberate `temperature=0.0` had it rewritten to `0.2` (`or` is false for a legitimate zero) and an agent with nothing stored had `0.2/0.9` invented for it. Post-fix both pass through correctly. Good catch; it was not in the reported symptom.

---

### Findings

**Q1 (major, non-blocking, pre-existing but in this PR's scope) — the title/errand path still sends `temperature=0` to families this PR just decided must receive nothing.**

`session/session.py:7417` builds its own `ChatRequest(temperature=0)` for conversation auto-naming, and a request-level value outranks the spec in `_sampling_params` — correctly, since that is the override contract Group 3 verifies. The result is that the errand contradicts the policy table on exactly the families the table exists for:

| model | main turn | title errand |
|---|---|---|
| `google/gemini-3.8-flash` | absent | **sends 0.0** |
| `openai/gpt-6-astra` | absent | **sends 0.0** |
| `deepseek/deepseek-v4-pro` | absent | **sends 0.0** |
| `anthropic/claude-opus-4-7` | absent | omitted |
| `anthropic/claude-opus-5`, `openai/o3` | absent | omitted |

Minimal reproduction (head, isolated env):

```python
from local_operator.model.configure import build_model_spec
from local_operator.harness.types import ChatRequest, Message
from local_operator.providers.clients import OpenAICompatClient

spec = build_model_spec("openai", "gpt-6-astra")
req = ChatRequest(model=spec, messages=[Message.user("t")], tools=[],
                  tool_choice="none", max_tokens=32, temperature=0)  # session.py:7417
print(OpenAICompatClient("https://x")._build_body(req).get("temperature"))
# -> 0.0     (main-turn path for the same spec prints None/absent)
```

Three things make this worth a decision rather than a shrug:

1. **It is pre-existing, not a regression.** Verified against `origin/main` in a separately-resolution-checked worktree: base sends `0.0` on the errand for the same three models. This PR does not make it worse.
2. **But this PR is what makes it wrong.** `gpt-6-astra` is the model whose vendor doc says *"Remove `temperature`, `top_p`, and `top_logprobs`"* — the same citation the PR uses to justify the OMIT row. If that is a documented 400 on the main turn, it is a documented 400 on the errand, and the user-visible symptom is an untitled conversation (the errand is `isolated`, one attempt, no fallback, so it fails silently and nothing reports it).
3. **The comment immediately above the line is now false.** It reads "`_sampling_params` already drops the key when the spec rejects sampling, so this cannot 400 a Claude-5 / o-series errand." That was true when rejection was only `supports_sampling_params`. It is no longer true for the OMIT families this PR introduces, and it is the sentence a future reader will trust instead of re-deriving.

Not fixing it here is defensible (it is out of the stated diff). Fixing it looks like dropping the `temperature=0` argument, or gating it on `spec.temperature is not None`, plus correcting that comment. Your call — I am reporting it at full severity, not asserting the disposition.

**Q2 (minor) — `claude-sonnet-4-5` / `claude-opus-4-6` now OMIT, which contradicts the PR's own stated intent.**

The table comment says *"4.5/4.6 still honour the pair and must not regress, so the arm starts at generation 4.7"*, and the vendor evidence agrees ("Only the 4.5/4.6-era models still honour a real temperature"). But those ids match no `_SAMPLING_POLICY` row, so they land on the fallback and omit:

```
claude-opus-4-5      temperature=None top_p=None
claude-sonnet-4-5    temperature=None top_p=None
claude-opus-4-6      temperature=None top_p=None
```

The *behaviour* is fine — omitting on a family that honours the pair yields Anthropic's own `1.0`, which is what you want anyway, and it is strictly better than the `0.2` base sent. The defect is that the comment describes a carve-out the code does not implement, so a future reader will believe 4.5/4.6 is seeded and be wrong. Either reword the comment to say the arm only stops the *forward-reading* match from over-reaching and the fallback intentionally omits, or add an explicit row. Comment fix is my recommendation; no behaviour change needed.

**Q3 (minor) — `ChatOptions.temperature` is capped at `le=1.0` while the table now seeds Gemini/Kimi families at exactly `1.0` and vendors document a `0.0–2.0` range.**

`server/models/schemas.py:47` bounds the per-request override to `[0.0, 1.0]`. Google documents temperature `0.0–2.0` for Gemini 2.5, and Qwen's range is `[0, 2)`. A server caller who wants `1.5` on `gemini-2.5-flash` gets a 422 from our own schema, not from the vendor. Untouched by this PR and arguably always wrong, but this PR is the first time the app has a per-family notion of a vendor's real range, so it is the natural moment to notice. Also the four docstrings at `routes/chat.py:89/217/331/455` still show `{"temperature": 0.2, "top_p": 0.9}` as the example options payload — that is now the exact value the PR argues nobody should send, and it is what an API consumer will copy.

---

### Not covered, and why

- **No live provider calls.** Every "documented 400" in this report is vendor documentation plus a verified absent/present key on our wire body — I did not spend credentials to observe a real 400 from Anthropic, OpenAI or Google. What I can state with my own evidence is exactly what this app now puts on the wire, which is the part the PR controls. A live probe against one 4.7-class and one `gpt-6-astra` endpoint would close the loop on the rejection claim itself.
- **`gemini-3.1-pro-preview` honour-vs-ignore is still UNKNOWN** in the vendor evidence, as the research notes flag. OMIT is correct under both readings, so this is not a gap in the change — only in what anyone can know without a live call.
- **Radient passthrough semantics remain UNVERIFIED** per the same notes; the code's choice to let a silent listing never widen an OMIT is the safe direction and I confirmed that logic holds on the `radient/` cells, but I could not confirm Radient's actual forwarding behaviour.
- **`_listing_forbids_sampling` memo eviction** under a >`_EFFORT_MEMO_MAX` working set was not exercised — it needs a live listing to populate, and my cold-cache runs never reached the bound.

Full matrix with every cell, command and actual output: `/tmp/qa594/QA-MATRIX-round1.md` (183 cells; 182 PASS, 1 FAIL = Q1).

damianvtran:

**Implementer:** `coder` subagent (lopdev team) — Claude Opus 4.5
**Scope:** `dcac3c26..94f6a138` (rebased onto `5d53a9e6`, current `origin/main`)

One unified commit for review R1–R6 and QA Q1–Q3, per the batching rule. Thank you both — R1 is a genuine ship-stopper and the diagnosis was exactly right.

---

#### R1 — BLOCKER — escape hatch reopens the 400 on reject-families — **FIXED** (`94f6a138`)

Confirmed the leak before fixing it, and it was **worse than reported**: on the OpenAI-compat wire — which is how the aggregator routes reach these models, and which QA identified as the route that actually carried the outage — `claude-opus-4-7` leaked too:

```
BEFORE (dcac3c26), configure_model(..., temperature=0.2, top_p=0.9):
  openai/gpt-6-astra          openai_wire={'temperature': 0.2, 'top_p': 0.9}   LEAK
  anthropic/claude-opus-4-7   openai_wire={'temperature': 0.2, 'top_p': 0.9}   LEAK
  anthropic/claude-opus-4-8   openai_wire={'temperature': 0.2, 'top_p': 0.9}   LEAK
```

The Anthropic-wire protection was the `thinking`-branch gate in `AnthropicClient._build_body`, i.e. **incidental** — exactly as the manager flagged, and a future refactor of that gate would have silently removed it.

**Fix.** `_SamplingPolicy` is now a `NamedTuple` carrying an explicit `rejects` flag, giving the table a genuine three-state answer, and a rejecting row clears `supports_sampling_params` — reusing the existing suppression mechanism rather than inventing a second one, as suggested.

One implementation note worth recording, because it is a trap: I first wrote this as a second `_REJECT_SAMPLING = (None, None)` sentinel distinguished by identity. **That silently cannot work** — equal constant tuples are interned to a single object, so `_OMIT_SAMPLING is _REJECT_SAMPLING` is `True` and the two states collapse. Verified before relying on it; the `NamedTuple` field is why the distinction is real. Noted in the code comment so nobody reintroduces it.

```
AFTER (94f6a138), same call:
  openai/gpt-6-astra                    REJECT             {}
  anthropic/claude-opus-4-7             REJECT             {}
  anthropic/claude-opus-4-8             REJECT             {}
  openrouter/anthropic/claude-opus-4-7  REJECT/aggregator  {}
  kimi/k3                               REJECT (kimi-pin)  {}
  google/gemini-3.8-flash               OMIT (ignores)     {'temperature': 0.2, 'top_p': 0.9}
  anthropic/claude-sonnet-4-5           HONOUR             {'temperature': 0.2, 'top_p': 0.9}
  kimi/kimi-k2-0711-preview             HONOUR (legacy)    {'temperature': 0.2, 'top_p': 0.9}
```

Verified on all four wire builders, direct and aggregator routes. `claude-sonnet-4-5` sending 0.2 still works, as required.

**One deliberate narrowing of your suggested fix.** I first marked the whole `kimi|moonshot` row REJECT and it flipped three rows you explicitly required stay `True` (`kimi-k2-0711-preview`, `moonshot-v1-128k`, `openrouter/moonshotai/kimi-k2`) — the over-match you warned about. The vendor evidence documents rejection for the **current** lineup (K3/K2.7/K2.6), the legacy ids accept the pair, and mainland policy is explicitly UNVERIFIED. So the Kimi row stays OMIT, and the ids whose rejection is documented *and* was observed live keep their hard suppression through `_KIMI_PINNED_SAMPLING`, which is scoped to exactly those. Reverted rather than shipped.

New regression tests: `test_a_reject_family_suppresses_an_override_instead_of_400ing` (6 routes, asserts the real wire body) and `test_an_ignore_or_honour_family_still_forwards_an_override` (4 routes, proves suppression does not spread).

---

#### Q1 — MAJOR — title errand sends `temperature=0` — **FIXED** (`94f6a138`)

Agreed it belongs in this PR: same subject matter, small fix, and the comment above it now asserts something false. `session.py` now passes `temperature=model.temperature`, deferring to the family policy instead of hardcoding.

```
                            errand wire temp
openai/gpt-6-astra          ABSENT   (was 0.0)
google/gemini-3.8-flash     ABSENT   (was 0.0)
deepseek/deepseek-v4-pro    ABSENT   (was 0.0)
google/gemini-2.5-pro       1.0      (seeded family keeps its documented value)
alibaba/qwen3-coder-plus    0.7
```

The comment is rewritten to explain why a request value outranking the spec is what made the hardcoded `0` wrong, and records the quiet failure mode you identified (isolated, one attempt, no fallback → a silently untitled conversation).

This also required updating `test_the_naming_errand_leaves_the_session_isolated_and_cheap`, which pinned `request.temperature == 0` — the old contract. It now asserts `request.temperature == request.model.temperature`, i.e. the errand follows the policy.

---

#### R2 / Q2 — MINOR — Anthropic 4.5/4.6 arm never fires — **FIXED (comment)** (`94f6a138`)

Took the recommendation: no behaviour change, corrected the claim. The fallback comment now states that 4.5/4.6 land there **deliberately** — omitting yields Anthropic's own 1.0, which is exactly what a row would have to seed, so a row would add a maintained constant that could only drift — and that the 4.7+ arm reads forward purely so it cannot reach back over them.

#### R3 — MINOR — dead `DEFAULT_TEMPERATURE` initializer — **FIXED** (`94f6a138`)

Removed. The branches now assign a `policy` object with no initial value, so the constants no longer look like a live default in the one file whose job is to stop them being one.

#### R4 — MINOR — `speech.py` as a second consumer — **FIXED (documented)** (`94f6a138`)

The table docstring now states it is **scoped to chat/completion sampling**, names `speech.py` as a second consumer, notes that speech ids currently all reach the OMIT fallback (correct for them), and marks the point at which a speech-aware branch would be needed rather than another row.

#### R6 — MINOR — server `options` comment claims a path that does not work — **COMMENT FIXED; behaviour deferred** (`94f6a138`)

Agreed and confirmed out of scope — QA independently verified it is equally broken on base. The comment no longer claims that path. It now names only `_AGENT_SAMPLING_FIELDS` (which both of you verified works), states explicitly that the routes mutate the **scalar** while the wire reads the **spec**, and records that `bootstrap.py`'s `sampling_overrides` seam is correct.

`deferred — pre-existing on base, out of scope for this PR; the misleading comment is fixed, the behaviour is not.`

#### R5 — NIT — PR body accuracy — **FIXED**

Both corrections made to the description:

1. **Route correction.** The "400 on every turn on the direct route" framing was too strong. The body now states that the direct `anthropic` route already omitted on base via the thinking gate, and that the 0.2 reached those models through the **aggregator** routes — citing QA's `openrouter/` and `radient/` rows. It also connects that same incidental gate to why R1's leak read as safe on the Anthropic wire.
2. **Version.** Rebased; see below.

#### Q3 — MINOR — `ChatOptions` cap and stale docstrings — **FIXED** (`94f6a138`)

`temperature` raised from `le=1.0` to `le=2.0`, the range vendors document. `top_p` deliberately stays at 1.0 — it is a probability mass and genuinely cannot exceed it. Verified through the real validator:

```
temperature=1.5  ACCEPTED   (was 422 from our own schema)
temperature=2.0  ACCEPTED
temperature=2.1  REJECTED
top_p=1.5        REJECTED
```

All four `routes/chat.py` example payloads changed from `{"temperature": 0.2, "top_p": 0.9}` to `{"temperature": 1.0, "top_p": 0.95}`.

---

#### Housekeeping and rebase

- QA scratch files (`tests/e2e/test_qa594_boot.py`, `.bak`) removed; the tree is clean and they were never committed.
- **Rebased twice.** `origin/main` advanced during remediation: first to `06928224` (#593, which took 0.44.67), then to `5d53a9e6` — and 0.44.68/0.44.69 were both released meanwhile. Version is now **0.44.70**. The only conflict either time was the `pyproject.toml` version line, exactly as predicted; the PR is `MERGEABLE` again.
- A mid-run external deletion of the worktree cost the uncommitted remediation; it was recovered in full from the dropped stash object and re-verified from scratch rather than assumed. The venv was rebuilt as a **real** editable install (the recreated worktree defaulted to a symlink), mapping confirmed to this checkout before any result was trusted.

#### Gates on the new head (`94f6a138`)

```
flake8 .                                 CLEAN
black --check .    (26.1.0)              836 files unchanged
isort --check-only (5.13.2)              CLEAN
pyright --pythonpath .venv/bin/python    0 errors, 0 warnings
pytest tests/unit                        11872 passed, 15 skipped
pytest tests/e2e -m e2e -n0              4 passed
```

An `isort` ordering slip on the new test import was caught by CI on the previous head and fixed in `94f6a138`. Worth recording why my local gate missed it: `isort --check-only .` exits 0 on this tree while the file genuinely fails, and only CI’s exact `isort --check .` surfaces it — I have switched to CI’s flag. CI is running on `94f6a138`. Not merging — over to you.

damianvtran:
#### CI note — the shard-0 failure was a pre-existing order-dependent flake, not this PR

The first run on `94f6a138` failed `test (3.12, 0)` on `tests/unit/mobile/test_tui_bridge.py` (`no discovery record published`). I did not wave it off as "environmental" — I reproduced the shard.

- My diff touches **no mobile code** (`git diff --name-only origin/main..HEAD | grep mobile` → 0 hits).
- I rebuilt CI's exact shard-0 file list (`i % 5 == 0`, 70 files) and ran it against **`5d53a9e6`, unmodified `origin/main`, with none of my changes**. It fails there too — on a *different* test each run: `test_subagent_view.py::test_a_narrow_viewport_opens_on_a_row_head_not_a_wrap_fragment` on one pass, `test_skill_invocation.py::test_non_invocations_are_sent_verbatim` on the next.
- Each of these passes in isolation on both trees. The failure appears only under the shard's ordering, and the identity of the victim moves between runs. `origin/main` is itself currently red on `5d53a9e6` and `66797b6f` for the same class of reason.

So: order-dependent flakiness that predates this branch. Re-running the failed shard on this same head went green, and **all 13 checks now pass on `94f6a138`**.

`deferred — pre-existing order-dependent flake on origin/main, unrelated to this diff and out of scope here.` Flagging it here rather than filing an issue, per the no-follow-up-issues rule; it is worth separate attention since main is red on it.

